PR #2410, opened 2023-04-14
Port Voodoo support with multi-threaded software backend

dreamer:
To test this:

- Linux / macOS / Windows MSYS builds: `meson setup -Dexperimental_voodoo_support=true build`
- Windows MSVC build: *manually* switch `C_VOODOO` define in `src/platform/visualc/config.h` - I expect the build will fail, but can't test it myself ATM.

--

image
*Tomb Raider I using 3dfx patch, with viewport set to 1280x960 and `crt/geom.tweaked` shader.*

image
*[Screamer Rally Demo](https://www.dosgamesarchive.com/download/screamer-rally) seems to be working fine.*

--
This is another attempt at #339, this time using @kekko-dos patch with major revision by @schellingb (ec0cd9d880287440d028d9ae03c5aa2308ca384f) and published [here](https://www.vogons.org/viewtopic.php?p=1099414#p1099414). The patch size went from ~10k LOC to ~7k LOC (with lots of ifdefed-out code, but I think it might be ok for the initial merge). It's definitely easier to understand now and it's *almost-reviewable* ;).

We kept 3dfx support out of main branch for so long due to [multiple serious issues](https://github.com/dosbox-staging/dosbox-staging/issues/339#issuecomment-623049560), and this version improves the situation considerably:

1. [x] Avoid duplicated OpenGL initialization -- *No longer an issue, this version supports software backend only.*
2. [x] Provide buildsystem switch for the feature -- *Patch did it, I made some revisions to make it work correctly in Meson.* 
3. [x] Clear-out legal and licensing issues with MAME project - *Not solved yet, but not a huge blocker any more - license text embedded in the original patch was actually verbatim copy of BSD-3-Clause. We still have some work to do here though, as it turns out we're not distributing BSD license text correctly :disappointed: (AFAIK no DOSBox version does it correctly). But this should be an easy fix. (solved)*
4. [x] Window being re-created when Vodoo renders stuff -- *Solved; software renderer does not touch emulator window code at all.*
5. [x] Bad performance in software mode -- *Partly solved - for me code works smoothly in window mode and then struggles in fullscreen. This is good enough for experimental support IMHO, we can fix it later.*
[...]

PR #2432, opened 2023-04-27
Fix regression in Build Engine games (3Dfx mode)

dreamer:
Fixes this issue: https://github.com/dosbox-staging/dosbox-staging/pull/2410#issuecomment-1521827139 /  https://www.youtube.com/watch?v=Wug_1j9gmsw

Turns out the regression was caused by performance improvement (avoiding slow `std::floor` function).

This is the first attempt at fixing it; it seems to work (tested so far in Blood, Shadow Warrior, TR1, Pył), but more testing is needed. The current rounding algorithm is not mathematically correct (gives wrong results below 0), but it seems like it makes no difference to the game visuals (and we don't know what was the actual behaviour on real Voodoo card).

An idea for mid-term future improvement: this function is always called twice, on x and y floats; it would be beneficial to change it into a function accepting 2 floats and utilising SIMD to round both integers at once.

Shoutout to @schellingb for finding the root cause of regression!

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -24,11 +24,11 @@ jobs:
             reconfigure_build: false
           - name: MSVC 64-bit
             arch: x64
-            max_warnings: 898
+            max_warnings: 902
             reconfigure_build: false
           - name: MSVC, +voodoo
             arch: x64
-            max_warnings: 910
+            max_warnings: 914
             reconfigure_build: true
 
     steps:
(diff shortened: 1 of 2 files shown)


kcgen:
Fantastic quick fix, @schellingb and @dreamer

kcgen (approved):

ThomasEricB:
I'm going to test on my end, I'll send my review shortly

ThomasEricB (approved):
Looks good on both Shadow Warrior and on Blood (even though Blood runs slower at this mode). Apologies for the lack of video, tonight is not being kind with my recording attempts :/

PR #2434, opened 2023-04-28
Make triangle_worker members threadsafe

kklobe:
Both `threads_active` and `done` are written to and read from different threads, which is technically a race condition, and `volatile` is not a thread synchronization primitive.

### DIFF
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -941,43 +941,43 @@ struct triangle_worker
 
 struct voodoo_state
 {
-	UINT8				chipmask;				/* mask for which chips are available */
+	UINT8				chipmask = {};				/* mask for which chips are available */
 
-	voodoo_reg			reg[0x400];				/* raw registers */
-	const UINT8 *		regaccess;				/* register access array */
-	bool				alt_regmap;				/* enable alternate register map? */
+	voodoo_reg			reg[0x400] = {};				/* raw registers */
+	const UINT8 *		regaccess = {};				/* register access array */
+	bool				alt_regmap = {};				/* enable alternate register map? */
 
-	pci_state			pci;					/* PCI state */
-	dac_state			dac;					/* DAC state */
+	pci_state			pci = {};					/* PCI state */
+	dac_state			dac = {};					/* DAC state */
 
-	fbi_state			fbi;					/* FBI states */
-	tmu_state			tmu[MAX_TMU];			/* TMU states */
-	tmu_shared_state	tmushare;				/* TMU shared state */
-	UINT32				tmu_config;
+	fbi_state			fbi = {};					/* FBI states */
+	tmu_state			tmu[MAX_TMU] = {};			/* TMU states */
+	tmu_shared_state	tmushare = {};				/* TMU shared state */
+	UINT32				tmu_config = {};
 
 #ifdef C_ENABLE_VOODOO_OPENGL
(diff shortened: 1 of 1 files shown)


src/hardware/voodoo.cpp
```diff
@@ -929,12 +930,13 @@ struct draw_state
...
 {
-	bool threads_active, use_threads, disable_bilinear_filter;
+	std::atomic_bool threads_active;
+	bool use_threads, disable_bilinear_filter;
 	UINT16 *drawbuf;
 	poly_vertex v1, v2, v3;
 	INT32 v1y, v3y, totalpix;
 	SDL_sem* sembegin[TRIANGLE_THREADS];
-	volatile bool done[TRIANGLE_THREADS];
+	std::atomic_bool done[TRIANGLE_THREADS];
```
dreamer, line 939: An unfortunate side-effect of this change: struct is no longer POD, thus C++ stopped handling copy-assignment for us (*thank you, C++! /s*).

Should be an easy fix though; compiler provides some more info:

> ../src/hardware/voodoo.cpp:7188:43: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct triangle_worker’ with no trivial copy-assignment; use value-initialization instead [-Wclass-memaccess]

dreamer (changes requested):

dreamer:
> (…) `volatile` is not a thread synchronization primitive.

Well said.

ThomasEricB:
I intend on reviewing this later as well @kklobe if you don't mind.

src/hardware/voodoo.cpp
kklobe: right-o - fixed

kklobe:
> I intend on reviewing this later as well @kklobe if you don't mind.

Please do!

kcgen:
All working well, without any functional or performance regressions.

kcgen (approved):
Working well without performance or functional regressions.

src/hardware/voodoo.cpp
```diff
@@ -929,53 +930,54 @@ struct draw_state
...
 };
 
 struct voodoo_state
 {
-	UINT8				chipmask;				/* mask for which chips are available */
+	UINT8				chipmask = {};				/* mask for which chips are available */
```
dreamer, line 944: Since we retouched these lines, please do replacements like `UINT8` -> `uint8_t` (int retouched lines only, please!). Once finished run `./scripts/format-commit.sh` to apply our coding standards to retouched lines (I think it will nicely align all the comments).

src/hardware/voodoo.cpp
```diff
@@ -3013,7 +3015,7 @@ iterated W    = 18.32 [48 bits]
 
 **************************************************************************/
 
-static voodoo_state *v;
+static voodoo_state *v = {};
```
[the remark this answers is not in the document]
kklobe: that is the default value.

```cpp
#include <cstdio>

char *v = {};
char *vv = nullptr;

int main()
{
    printf("a = %p, b = %p\n", v, vv);
}
```
```log
a = 0x0, b = 0x0
```
but happy to change it if we don't like that style.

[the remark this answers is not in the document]
dreamer: Please do, thanks! :) I prefer to keep the "normal" style and initialize the pointer with something that is clearly a pointer.

dreamer (approved):
Great, thanks! :smile:

PR #2441, opened 2023-04-29
Fix regression in VGA draw's wrapped line case

### LINKED ISSUE #2439: Hard crash on Windows when using QPEG with the latest dev build

dosbox-staging-windows-msys2-x86_64-v0.81.0-alpha-674-g3b349

Windows 10

_No response_

### Relevant hardware info

_No response_

With the latest dev build, QPEG consistently hard-crashes the DOSBox process when attempting to display an image.

I can reproduce it 100% of the time on Windows using the MSYS2 dev build. The issue does not happen on macOS with a self-built build off `main`; QPEG works flawlessly there.

v0.80.1 also works flawlessly on Windows. So this is a weird regression, most likely memory corruption that manifests more easily on my Windows box. I suspect it has something to do with the double-scanning VGA changes.

Explain how to reproduce

1. Use my attached repro pack
2. Start with `dosbox -noprimaryconf`
3. After QPEG started, press Enter on `GODS.BMP` to view it---bamm, crash! It crashes in spectacular ways, every time a bit differently, but the process always exits after a few seconds of hanging...

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
After incorporating `vgaonly`'s features as well as using a 32-bit DAC palette, the `draw_linear_line_from_dac_palette`'s wrapped portion of the function was not updated in the same way as the unwrapped (or "whole-line") approach.

This commit moves the line-writing chunk into a lambda (to avoid code-repetition), which also ensures they all use the same writing logic.

Fixes #2439.

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -353,42 +353,55 @@ static uint8_t * VGA_Draw_Linear_Line(Bitu vidstart, Bitu /*line*/) {
 
 static uint8_t* draw_linear_line_from_dac_palette(Bitu vidstart, Bitu /*line*/)
 {
-	const auto offset = vidstart & vga.draw.linear_mask;
-	const uint8_t *ret = &vga.draw.linear_base[offset];
-	const auto palette_map = vga.dac.palette_map;
+	const auto bytes_per_vga_pixel = check_cast<uint8_t>(vga.draw.bpp / 8);
+	assert(bytes_per_vga_pixel > 0);
+	const auto pixels_per_line = check_cast<uint16_t>(vga.draw.line_length /
+	                                                  bytes_per_vga_pixel);
+
+	// Quick reference
+	constexpr auto palette_map = vga.dac.palette_map;
+
+	// Ensure the size of the palettized VGA line will fit
+	constexpr uint8_t bytes_per_palette_pixel = sizeof(palette_map[0]);
+	assert(pixels_per_line * bytes_per_palette_pixel <=
+	       static_cast<int>(templine_buffer.size()));
+
+	// The running write position for the palettized VGA pixels
+	auto target_addr = TempLine;
+
+	auto write_pixels_from = [&](const uint8_t* vga_line_start,
+	                             const uint16_t num_pixels) {
+		auto vga_pixel          = vga_line_start;
+		const auto vga_line_end = vga_line_start + num_pixels;
+		while (vga_pixel < vga_line_end) {
+			const auto source_addr = palette_map + *vga_pixel++;
+			memcpy(target_addr, source_addr, bytes_per_palette_pixel);
+			target_addr += bytes_per_palette_pixel;
+		}
+	};
 
 	// see VGA_Draw_Linear_Line
+	const auto offset = vidstart & vga.draw.linear_mask;
 	if (GCC_UNLIKELY((vga.draw.line_length + offset)& ~vga.draw.linear_mask)) {
-		const auto end = (offset + vga.draw.line_length) &
-		                 vga.draw.linear_mask;
+		const auto wrapped_end_byte = (offset + vga.draw.line_length) &
+		                              vga.draw.linear_mask;
 
-		// assuming lines not longer than 4096 pixels
-		const auto wrapped_len   = static_cast<uint16_t>(end & 0xFFF);
-		const auto unwrapped_len = check_cast<uint16_t>(
-		        vga.draw.line_length - wrapped_len);
+		const auto num_wrapped_pixels = check_cast<uint16_t>(
+		        wrapped_end_byte / bytes_per_vga_pixel);
 
-		// unwrapped chunk: to top of memory block
-		for (uint16_t i = 0; i < unwrapped_len; ++i) {
-			write_unaligned_uint32_at(TempLine, i, palette_map[ret[i]]);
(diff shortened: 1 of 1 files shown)


src/hardware/vga_draw.cpp
```diff
@@ -353,42 +353,56 @@ static uint8_t * VGA_Draw_Linear_Line(Bitu vidstart, Bitu /*line*/) {
...
+
+	// Ensure the size of the palettized VGA line will fit
+	constexpr uint8_t bytes_per_palette_pixel = sizeof(palette_map[0]);
+	assert(pixels_per_line * bytes_per_palette_pixel <=
+	       static_cast<int>(templine_buffer.size()));
+
+	// Running write position for the palettized VGA pixels
+	auto target_addr = TempLine;
+
+	auto write_pixels_from = [&](const uint8_t* vga_line_start,
```
dreamer, line 372: I know it will be a bit worse on the eyes, but in this case I strongly suggest capturing references to `palette_map` and `target_addr` explicitly instead of all variables in context:

Replace `[&]` with `[&palette_map, &target_addr]`.

In this case I think it's reasonable because `target_addr` is de-facto in-out parameter for this lambda.

dreamer:
Tested our QPEG testcase in debug build - works fine. Thanks @kcgen!

src/hardware/vga_draw.cpp
kcgen: Looks good @dreamer; much clearer what's being scoped-in.

dreamer:
Oh, compilers complain about `unused-lambda-capture`? Hmm, maybe `constexpr` variable does not need to be captured in this case?

kcgen:
> Oh, compilers complain about `unused-lambda-capture`? Hmm, maybe `constexpr` variable does not need to be captured in this case?

Interesting; yes - looks like it's using the `map` by value without needing to be explicitly captured. Oh, this is also true for the `bytes_per_palette_pixel`, which is also a `constexpr` value.

(updated)

kcgen:
> Oh, compilers complain about `unused-lambda-capture`? Hmm, maybe `constexpr` variable does not need to be captured in this case?

Tug-o'-war between compilers :-) clang is happy now, but MSVC is not:

D:\a\dosbox-staging\dosbox-staging\src\hardware\vga_draw.cpp(377,1): error C3493: 'palette_map' cannot be implicitly captured because no default capture mode has been specified [D:\a\dosbox-staging\dosbox-staging\vs\dosbox.vcxproj]

D:\a\dosbox-staging\dosbox-staging\src\hardware\vga_draw.cpp(378,1): error C3493: 'bytes_per_palette_pixel' cannot be implicitly captured because no default capture mode has been specified [D:\a\dosbox-staging\dosbox-staging

kcgen:
Ok: trued using a reference and two values:

```c++
... = [&target_addr, palette_map, bytes_per_palette_pixel](...)
```

But now clang is complaining:

```
../../src/hardware/vga_draw.cpp:372:42: warning: lambda capture 'palette_map' is not required to be captured for this use [-Wunused-lambda-capture]
        auto write_pixels_from = [&target_addr, palette_map, bytes_per_palette_pixel](const uint8_t* vga_line_start,
                                              ~~^~~~~~~~~~~
../../src/hardware/vga_draw.cpp:372:55: warning: lambda capture 'bytes_per_palette_pixel' is not required to be captured for this use [-Wunused-lambda-capture]
        auto write_pixels_from = [&target_addr, palette_map, bytes_per_palette_pixel](const uint8_t* vga_line_start,
```

Going back to the single reference.

kcgen:
Investigated the additional clang static analysis issue:



These turned out to be all voodoo-related:



I've added a commit to accommodate them so we can carry on.

Will merge once this is passing CI.

.github/workflows/analysis.yml
```diff
@@ -125,7 +125,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 12
+          MAX_BUGS: 16
```
dreamer, line 128: Hmm, it was already accounted for in PR introducing Voodoo support in d2d2154413003d7a2be897d796aa3d26b3d48202. Maybe we get new warnings with new clang static analyser or something?

dreamer:
> Investigated the additional clang static analysis issue:
> 
> These turned out to be all voodoo-related:

Yeah, this is weird... we already had 12 warnings with voodoo enabled and now it jumped back up…

Good call, let's keep the higher limit for now and limit it by fixing issues in voodoo code.

dreamer (approved):

kcgen:
the warning bump was my fault in this PR.

I was testing some SIMD acceleration (both SSE and NEON) in the VGA draw function - and forgot to remove that GCC flag from Meson. 

This exposed 4 new warnings in libFLAC that I incorrectly believed were there all along.. I've snipped out that Meson SSE flag and we're back to 12 warnings.

GranMinigun:
MSVC cannot into standards-conforming lambdas by default, I had it spew out a ton of errors on my branch because of that. You can just cherry-pick 1453a5b657ecee5879880dc09c55a4da87230a53 to make everyone happy.

johnnovak:
@kcgen Just tested this on my end and now QPEG works fine without any crashes! Great quick fix! 🎉

PR #2451, opened 2023-05-04
Capture modularisation/cleanup

johnnovak:
Hacking'n'slashing the entangled spaghetti mess responsible for capturing stuff into more manageable pieces.
No functional changes, just reorganising stuff & cleanup. AFAIK I haven't broken anything, only upset the linters a bit... (to be fixed).

This all lived in `hardware.cpp` previously in a big cosy ball of mud 😩 Sure, that's a good place... we're all dealing with "hardware" after all... computers are hardware too, and so on 🤦🏻 🥶 

This is in preparation for the screenshot improvements.


kcgen:
> This all lived in hardware.cpp previously in a big cosy ball of mud weary Sure, that's a good place... we're all dealing with "hardware" after all... computers are hardware too, and so on 🤦🏻 :cold_face: 

:rofl:

ThomasEricB:
Nice work @johnnovak ! You really showed `hardware.cpp` who is boss!

johnnovak:
> You really showed `hardware.cpp` who is boss!

It had it coming. 😎

johnnovak:
So, for the record, this builds fine on macOS, even after a full clean & rebuild. I'll screw around on Windows today or tomorrow to eradicate those build failures—hopefully that will get rid of most of the Linux failures too.

include/hardware.h
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2002-2021  The DOSBox Team
+ *  Copyright (C) 2002-2023  The DOSBox Team
```
dreamer, line 2: We are keeping "DOSBox Team" copyrights intact; in cases like this one, add another copyright line for "DOSBox Staging Team" (grep through the code will give plenty of examples).

src/capture/capture.cpp
```diff
@@ -0,0 +1,268 @@
...
+		*test = 0;
+		if (strncasecmp(tempname, file_start, strlen(file_start)) != 0)
+			continue;
+		const int num = atoi(&tempname[strlen(file_start)]);
+		if (num >= last)
+			last = num + 1;
+	}
+	close_directory(dir);
+	char file_name[CROSS_LEN];
+	sprintf(file_name,
```
dreamer, line 91: Let's update it to `safe_sprintf` while we're at it.

include/hardware.h
johnnovak: Yeah okay. TBH I did not even read it, just updated the date 😄

src/capture/capture.cpp
johnnovak: Sure, good idea.

src/capture/capture.cpp
```diff
@@ -0,0 +1,268 @@
...
+	return file_name;
+}
+
+FILE* CAPTURE_CreateFile(const char* type, const char* ext)
+{
+	const auto file_name = capture_generate_filename(type, ext);
+
+	FILE* handle = fopen(file_name.c_str(), "wb");
+	if (handle) {
+		LOG_MSG("Capturing %s to %s", type, file_name.c_str());
```
dreamer, line 107: Let's surround paths we print to log with quotes: `'%s'`.

johnnovak: Good idea.

johnnovak:
Moving this back to draft as I'm going to clean it up more substantially in addition to the fairly mechanical splitting/refactoring/etc. I've already done.

The end goal is to make all this easily modifiable for future enhancements... because currently it isn't 😩

johnnovak:
Okay, this is ready for final review @kcgen & @dreamer. I've regressions tested everything and all capture functionality works 100% fine.

I think the whole thing should be condensed into a single commit; the commit history doesn't add much value as files have been split, moved around, and refactored in a major way. Plus I did break a few things in the first phase because the state handling was still a mess... This has been rectified in my last big refactor commit. Now you can mostly follow what's going on by reading the code, finally 😅

include/hardware.h
johnnovak: Done in all files.

src/capture/capture.cpp
johnnovak: Done.

src/capture/capture.cpp
johnnovak: Done.

johnnovak:
Someone give me a tick please? @kcgen @dreamer @FeralChild64 Feel free to test, I could not spot any regressions/problems after the refactor.

kcgen:
The "Move parts of generate rendered screenshot to capture_image" commit fails to compile (bad for bisecting):

``
In file included from ../../include/sdlmain.h:33,
                 from ../../src/capture/capture_image.cpp:24:
../../src/capture/../libs/ppscale/ppscale.h: In constructor ‘PPScale::PPScale(int, int, double, bool, int, int)’:
../../src/capture/../libs/ppscale/ppscale.h:51:17: error: ‘assert’ was not declared in this scope
   51 |                 assert(source_w > 0);
      |                 ^~~~~~
```

kcgen (commented):
Tests are running well - I tested wav, avi, and both screenshots all individually and then all combinations:
wav
avi
screenshots
wav + avi
wav + screenshots
wav + avi + screenshots
avi + screenshots

So if this is cleanup only without functional changes intended, then I think it's OK.  Just need all commits building and it'll be set!

johnnovak:
Thanks for testing it @kcgen. I've squashed the offending commits, I'm pretty sure I was able to compile every commit on macOS, but they I was relying on incremental commits.

kcgen:
> Thanks for testing it @kcgen. I've squashed the offending commits, I'm pretty sure I was able to compile every commit on macOS, but then I was relying on incremental compilation.

:+1: , your good commit messages indicate unbuildable commits still exist in the latest push:



Checking prior-to-fixed macOS commit -- indeed, it doesn't build:

kcgen (approved):
approving - feel free to merge @johnnovak once those last compile commits are handled :rocket:

kcgen:
> I think the whole thing should be condensed into a single commit

Sorry I missed that earlier! Given it all checks out, I'll go-ahead and squash and merge.

johnnovak:
> > I think the whole thing should be condensed into a single commit
> 
> Sorry I missed that earlier! Given it all checks out, I'll go-ahead and squash and merge.

Yeah, by the final commit everything ended up being substantially refactored anyway, so hence my suggestion to just squash the whole thing. It would be possible to do such big refactors and keep all builds green on the different plstforms, but that's much more time consuming... Thanks for merging.

PR #2452, opened 2023-05-04
Update macos.md

Burrito78:


johnnovak (commented):
Thanks for trying @Burrito78 , but something seems wrong here... Nevermind, as @kcgen has already raised the PRs and approved them.

PR #2453, opened 2023-05-04
docs (80.x): Fix SHA256 checksum reported for macOS package

kcgen:


johnnovak (approved):
@kcgen Approved both PRs. I'll let you do the manual deploys—just to spread the knowledge about doing the deploys (until we get auto-deploys working).

PR #2456, opened 2023-05-05
Implement integer viewport scaling

### LINKED ISSUE #2448: Kill "pixel-perfect" mode with fire 🔥🔥🔥🔥🔥
> _It is evil, so it must die._
>
> _If we let evil flourish, what makes that of us?_

## Summary

Pixel-perfect modes (`output = openglpp` and `output = texturepp`) prioritise 100% sharp pixels without any interpolation whatsoever over aspect ratio correctness. If the pixel-perfect scaler cannot hit the necessary integer scaling ratios to achieve an aspect ratio correct image, it will pick another pair of integer scaling factors that are _close_ (resulting in an aspect ratio _incorrect_ image).

This is based on two fundamentally misguided assumptions:

1. The most important goal when upscaling low resolution graphics is to achieve the sharpest upscaled pixels possible on a flat-screen.
2. Achieving the sharpest pixels possible is more important than aspect ratio correctness.

Both assumptions are quite wrong. Images on VGA CRT were never as sharp as on LCDs (but still quite sharp), so some very mild interpolation artifacts are more than permissible. The "bilinear sharp" shader used by default in DOSBox Staging achieves very good results with the interpolation "band" being no more than 1px wide around the "upscaled pixels", making the pixel-perfect shader not only redundant, but actually harmful (more on that below).

Make no mistake, integer scaling is still a good solution and is actually desirable for high-res content on a 1080p screen (e.g., 640x480 looks best when upscaled with a 2x scaling ratio, otherwise the interpolation artifacts will become a bit jarring even with the "sharp bilinear" shader). The big problem here is the "variable, semi-random aspect ratio" that the so-called "pixel-perfect" scaler generates (we should rename it to "aspect ratio imperfect"...)

## What's wrong with it

The good part is that resolutions with square pixels, such as 640x480 or 320x240, will get upscaled to 1280x960 (assuming a typical 1080p monitor, which is still the most common type). That's nice, no doubt.

But let’s consider 320×200, the most common resolution used by DOS games. The smallest integer scaling factors for this are 5x horizontally and 6x vertically. Problem is, that won’t fit into 1920×1080… So the pixel-perfect shader will cheerfully pick the “closest” integer scaling factors that would result in an image that fits. That will be 4x horizontally and 5x vertically, resulting in a 1280×1000 image. That's bad; now the image is stretched by 25% vertically instead of the correct 20% stretch (because 6/5 = 1.2, but 5/4 = 1.25).

Now switch to windowed mode and start resizing it. As we keep changing the window’s dimensions, we could end up with a whole variety of different PARs for our 320x200 image, e.g.:

- 960x800 (1:1.333 PAR)
- 640x400 (1:1 PAR)
- 1920x1400 (1:1.167 PAR)
- 2560x2000 (1:1.25 PAR)

Imagine random people posting screenshots of 320x200 games upscaled with semi-random PARs, then entering endless debates about which is the correct one...

## Solution

The solution is to completely eradicate every trace of the "pixel-perfect" mode with _extreme prejudice_. This means the total removal of the `openglpp` and `texturepp` output modes.

However, "snapping" the image to an integer scaling ratio in _one_ of the dimensions (so either horizontally, or vertically, but never both) is actually desirable, and that's something that the "pixel-perfect" shader actually gets right for screen modes that have 1:1 PAR (e.g., 640x480 and 320x240). 

Moreover, even for screen modes with non-square pixels, snapping the image size to the closest integer factor _only vertically_ is often desirable and even necessary to avoid interference (moire) artifacts when using CRT shaders that produce scanlines (e.g., snapping to integer vertical scaling ratios is a _must_ even on 4K, especially with 350 and 480-line resolutions to avoid moire).

Therefore, this useful subset of the "pixel-perfect" functionality will be retained under a different setting:

```ini
[render]
scaling_constraint = none | vertical-integer | horizontal-integer
```

Should be self-explanatory. `none` is the default—that's what you get currently with the default config. For example, when using `vertical-integer` on a 1080p screen in fullscreen with 320x200 content, the resulting image will be 1333x1000 with the default `sharp` scaler (5x vertical scaling factor, then whatever fractional horizontal ratio we need to produce an aspect ratio correct image). We'll get "pixel-perfect" scaling vertically, and horizontally we'll get the occasional 1px wide interpolation band which is hardly noticeable. Also, in general horizontal sharpness is a lot less important than vertical sharpness for computer generated images (e.g., think about text modes).

GranMinigun:
This PR removes pixel-perfect scaling modes and introduces simple integer scaling per selected axis. Help message needs some love, please provide it. And maybe setting name should be shortened by a single word. And I may have missed some weird corner-case with vertical 4D L-shaped pixels display modes which I have no idea where to test.

Closes #2448.

P.S. What's the purpose of `sdl.draw.scalex/y` when `sdl.draw.pixel_aspect` exists and has far better precision?


include/sdlmain.h
```diff
@@ -71,7 +69,9 @@ enum class HOST_RATE_MODE {
 	CUSTOM,
 };
 
-enum class SCALING_MODE { NONE, NEAREST, PERFECT };
+enum class SCALING_MODE { NONE, NEAREST };
+
+enum class INTEGER_SCALE_MODE { OFF, HORIZONTAL, VERTICAL };
```
johnnovak, line 74: UpperCamelCase please for new enum names and values.

johnnovak: IntegerScalingMode (not Scale)

include/sdlmain.h
```diff
@@ -108,6 +108,7 @@ struct SDL_Block {
 	bool resizing_window = false;
 	bool wait_on_error = false;
 	SCALING_MODE scaling_mode = SCALING_MODE::NONE;
+	INTEGER_SCALE_MODE integer_scale_mode = INTEGER_SCALE_MODE::OFF;
```
johnnovak, line 111: integer_scaling_mode

src/gui/sdlmain.cpp
```diff
@@ -3307,16 +3196,44 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
 	const auto draw_aspect = (sdl.draw.width * sdl.draw.scalex) /
 	                         (sdl.draw.height * sdl.draw.scaley);
-	const auto lwin_aspect = static_cast<double>(lwin_w) / lwin_h;
 
-	// calculate the viewport contingent on the aspect ratio of the limited
-	// window versus draw buffer
-	const auto lwin_is_wider = lwin_aspect > draw_aspect;
-	const auto view_w = lwin_is_wider ? iround(lwin_h * draw_aspect) : lwin_w;
-	const auto view_h = lwin_is_wider ? lwin_h : iround(lwin_w / draw_aspect);
+	int view_w = 0;
```
johnnovak, line 3203: Could we turn this function into something that can be easily comprehended by an average person unfamiliar with the code in a single pass?

I find the variable names confusing and inconsistent. The problem starts here:

```
const auto [lwin_w, lwin_h] = restrict_to_viewport_resolution(win_w, win_h);
```

Why `lwin`? Can we find a better name the describes this is the maximum bounds of the viewport we want to fit our image into? (hint: `bound_w`, `bound_h`).

src/gui/sdlmain.cpp
```diff
@@ -3307,16 +3196,44 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+	int view_h = 0;
+	switch (sdl.integer_scale_mode) {
+	case INTEGER_SCALE_MODE::OFF: {
+		// calculate the viewport contingent on the aspect ratio of the
+		// limited window versus draw buffer
+		const auto lwin_aspect   = static_cast<double>(lwin_w) / lwin_h;
```
johnnovak, line 3209: I'd find this a lot more readable:

```
const auto bounds_aspect   = static_cast<double>(bounds_w) / bounds_h
```

src/gui/sdlmain.cpp
johnnovak: Similarly, `draw_aspect` is very ambiguous to me... "Draw" can be anything; we're "drawing" here, right? 😅 

Can we be more precise and call this what it is? `source_aspect` would be a lot better I think. But technically, this is the _storage aspect ratio_, maybe call it that?

src/gui/sdlmain.cpp
```diff
@@ -3307,16 +3196,44 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+		const auto lwin_is_wider = lwin_aspect > draw_aspect;
```
johnnovak, line 3210: From this I can guess that `draw_aspect` must be aspect ratio corrected... right?

src/gui/sdlmain.cpp
```diff
@@ -3307,16 +3196,44 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+		view_w = lwin_is_wider ? iround(lwin_h * draw_aspect) : lwin_w;
+		view_h = lwin_is_wider ? lwin_h : iround(lwin_w / draw_aspect);
```
johnnovak, line 3212: I'd find this more readable and easier to parse:

```cpp
if (is_bounds_aspect_wider) {
	view_w = iround(bounds_h * output_aspect);
	view_h = bounds_h;
} else {
	view_w = bounds_w;
	view_h = iround(bounds_w / output_aspect);
}
```

src/gui/sdlmain.cpp
```diff
@@ -3307,16 +3196,44 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+		break;
+	}
+	case INTEGER_SCALE_MODE::VERTICAL: {
+		const auto integer_scale_factor = std::max(
+		        std::min(lwin_h / sdl.draw.height,
+		                 static_cast<int>(lwin_w / (sdl.draw.height *
+		                                            draw_aspect))),
+		        1);
+		view_w = iround(sdl.draw.height * integer_scale_factor * draw_aspect);
+		view_h = sdl.draw.height * integer_scale_factor;
```
johnnovak, line 3233: These are quite hard to follow, I spent like 2-3 minutes on it and gave up... I could figure it out eventually, but we really should optimise for readability by default. This is not performance intensive code, and even if it was, the compiler can usually optimise it well enough...

Can break this up into multiple statements with individual variable assignments with descriptive names instead of giving us the "final closed form"? Like in maths class; the teacher might see the final "closed form solution" already, but you need to explicitly tell the kids all the 10 individual derivation steps to see the solution. Well, we're the kids here 😄 

As an added benefit, that serves as documentation as well, instead of just giving people the final formula, then wishing them a nice day 😎

src/gui/sdlmain.cpp
```diff
@@ -3399,6 +3287,18 @@ static void set_output(Section *sec, bool should_stretch_pixels)
 		                                        // anymore
 	}
 
+	const std::string integer_scale_mode = section->Get_string("integer_scale_mode");
```
johnnovak, line 3290: I suggest `integer_scaling` as a better name.

src/gui/sdlmain.cpp
```diff
@@ -3399,6 +3287,18 @@ static void set_output(Section *sec, bool should_stretch_pixels)
...
+	if (integer_scale_mode == "off") {
+		sdl.integer_scale_mode = INTEGER_SCALE_MODE::OFF;
+	} else if (integer_scale_mode == "horizontal") {
+		sdl.integer_scale_mode = INTEGER_SCALE_MODE::HORIZONTAL;
+	} else if (integer_scale_mode == "vertical") {
+		sdl.integer_scale_mode = INTEGER_SCALE_MODE::VERTICAL;
+	} else {
+		LOG_WARNING("Unknown integer scale mode: %s",
```
johnnovak, line 3298: `"Unknown integer scaling mode '%s', defaulting to 'off'"`

johnnovak: Also set the integer scaling mode explicitly to "Off" here please.

src/gui/sdlmain.cpp
```diff
@@ -4483,6 +4383,10 @@ void config_add_sdl() {
 	        "  <custom>:  Limit the viewport within to a custom resolution or percentage of\n"
 	        "             the desktop. Specified in WxH, N%, N.M%. Examples: 960x720 or 50%");
 
+	pstring = sdl_sec->Add_string("integer_scale_mode", always, "off");
+	pstring->Set_help(
+	        "Scale output by integer factor along set axis: off, horizontal, vertical.");
+
```
johnnovak, line 4389: I suggest to rename this to `integer_scaling` and the following description that follows our new, normalised description style:

```
Constrain the horizontal or vertical scaling factor to integer values. The
correct aspect ratio is always maintained, which may result in a non-integer
scaling factor to be applied in the other dimension.
  off:         No integer scaling constraint; the aspect ratio correct image
               fills the viewport (default).
  horizontal:  Constrain the horizontal scaling factor to integer values
               within the viewport.
  vertical:    Constrain the vertical scaling factor to integer values
               within the viewport.

```

johnnovak (changes requested):
This works very well @GranMinigun; I gave it a good testing and it works splendidly! 🎉 

This is great because I can offer simplified instructions in my Getting Started guide in addition to explaining it to people how to calculate [aspect ratio correct viewport resolutions](https://dosbox-staging.github.io/preview/dev/getting-started/enhancing-prince-of-persia/#custom-viewport-resolution). E.g. now I can just give them something like this:

```ini
[sdl]
viewport_resolution = 80%
integer_scaling = vertical
```

So this is flawless from a functional point of view, but the code could be a bit more self-documenting. Also see my comments regarding the description text and the name of the new setting.

Great job, overall! 😎 🎉

src/gui/sdlmain.cpp
johnnovak: Also, `aspect` is in `[render]`, so I think this belongs to the `[render]` section too. I don't think we should add anything new to `[sdl]`, and in fact `viewport_resolution` should be moved from `[sdl]` to `[render]` as well. 

The whole `[sdl`] section is quite ill-named in itself. Why are implementation details leaking into the config? No one cares if we use SDL, GLFW, Unity, framework XYZ, or no framework at all—the configuration should be completely agnostic of that _implementation_ detail... We should move things _out_ from there into more appropriately named sections, like @FeralChild64 did with the `[mouse]` section—definitely we should not add to the mess.

include/sdlmain.h
GranMinigun: Are these styling nuances cemented anywhere? Maybe I missed some doc.

johnnovak: Not sure, maybe we should start a doco. But we went over this a number of times in other PRs already in the past 1-2 years. Maybe you check out other PRs a bit more often 😄

johnnovak:
> P.S. What's the purpose of `sdl.draw.scalex/y` when `sdl.draw.pixel_aspect` exists and has far better precision?

The creation of chaos and confusion? I have no idea, man 😄

include/sdlmain.h
ThomasEricB: @johnnovak @kcgen @dreamer Perhaps it's time to update [CONTRIBUTING.md](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md)? It does establish the use of `snake_case` and `CamelCase` for class and method names , but not `UpperCamelCase` for enums.

src/gui/sdlmain.cpp
GranMinigun: Replaced `std.max()` with a separate expression and added comments. Formatter has... interesting views on how these lines should look, so a bit more refinement might be needed if you find that exceptionally ugly.

GranMinigun:
Addressed comments (I think?), yell if something's missing.

src/gui/sdlmain.cpp
```diff
@@ -3305,19 +3194,61 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+	case IntegerScalingMode::Horizontal: {
+		// Calculate scaling multiplier that will allow to fit the whole
+		// image into the window or viewport bounds.
+		auto integer_scale_factor = std::min(
+		        bounds_w / sdl.draw.width,
+		        static_cast<int>(bounds_h / (sdl.draw.height *
+		                                     sdl.draw.pixel_aspect)));
+		// Safeguard to not let the image implode into a black hole.
+		if (integer_scale_factor < 1)
+			integer_scale_factor = 1;
```
johnnovak, line 3229: Please enclose single statements in curly braces too (reason: it's very error prone when you add more lines to it later).

src/gui/sdlmain.cpp
```diff
@@ -3305,19 +3194,61 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+		                sdl.draw.pixel_aspect);
+		break;
+	}
+	case IntegerScalingMode::Vertical: {
+		auto integer_scale_factor =
+		        std::min(bounds_h / sdl.draw.height,
+		                 static_cast<int>(bounds_w / (sdl.draw.height *
+		                                              output_aspect)));
+		if (integer_scale_factor < 1)
+			integer_scale_factor = 1;
```
johnnovak, line 3243: Please enclose single statements in curly braces too.

kcgen:
Huge cleanup, @GranMinigun - almost 500 LOC, gone :fire: 

No code comments from me.  

If anyone has done testing, please share!

I will plan to do light regression testing on Linux and macOS of the existing modes (given these code paths are intact, I don't expect to see changes), then will test the new integer-scaling settings, with and without viewport, resizing, and switching between full-and-windowed.

src/dosbox.cpp
```diff
@@ -622,6 +622,18 @@ void DOSBOX_Init()
 	        "320x200 or 640x400; whereas square-pixel modes, such as 640x480\n"
 	        "and 800x600, are displayed as-is.");
 
+	pstring = secprop->Add_string("integer_scaling", always, "off");
+	pstring->Set_help(
```
kcgen, line 626: (minor)
`integer_scaling` is added to the `[render]` section here, but looks like it's being read from the `[sdl]` section (in sdlmain).

kcgen:
Testing all good so far (`surface`, `texture`, and `opengl`).

src/dosbox.cpp
johnnovak: Yeah it was in `[sdl]` originally, then I asked @GranMinigun to move it to `[render]`.

src/gui/sdlmain.cpp
```diff
@@ -3305,19 +3194,61 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
+		break;
+	}
+	case IntegerScalingMode::Horizontal: {
+		// Calculate scaling multiplier that will allow to fit the whole
+		// image into the window or viewport bounds.
+		auto integer_scale_factor = std::min(
+		        bounds_w / sdl.draw.width,
+		        static_cast<int>(bounds_h / (sdl.draw.height *
+		                                     sdl.draw.pixel_aspect)));
+		// Safeguard to not let the image implode into a black hole.
```
johnnovak, line 3227: 😄

src/gui/sdlmain.cpp
GranMinigun:
```
# Commenting out for now as VisualStudio 2019 doesn't recognise it
#InsertBraces: true
```

...oh. That's why.

johnnovak:
We're almost there, @GranMinigun. In addition to our last round of comments, can you please also address the following few outstanding issues?

1. Remove `openglpp` and `texturepp` from the list of possible outputs in `sdlmain.cpp`:

    ```cpp
    	const char *outputs[] =
    	{ "surface",
    	  "texture",
    	  "texturenb",
    	  "texturepp",
    #if C_OPENGL
    	  "opengl",
    	  "openglnb",
    	  "openglpp",
    #endif
    	  0 };
    ```

2. Display a deprecation warning when `output = openglpp | texturepp` is selected. Just search for "deprecated" in `sdlmain.cpp` to see how it's done and test it to get the message right. Something like this:

    ```
    'openglpp' output mode has been deprecated, defaulting to 'opengl'. Consider using the 'integer_scaling' setting instead.
    ```
    Similar message for `texturepp`.

3. Remove/update references to `texturepp` and `openglpp` in README and README.md.

src/dosbox.cpp
kcgen: makes sense, thanks @johnnovak 

Once this conf match-up is ready I'll be able to check `horizontal` and `vertical`.

GranMinigun: I don't want to touch configuration ever again.

Should work now, but setting it at runtime requires manual triggering of viewport resize. Not quite sure yet how to do it at least remotely right.

GranMinigun:
I don't think it's possible to set deprecation message in such way.

kcgen:
Non-viewport testing results are great.
Vertical is especially nice to avoid moire!

---

## Off

mm2_000

mm2_001

mm2_002

vserver_000

---

## Horizontal

mm2_000
mm2_001
mm2_002
mm2_003
vgalemmi_000
vgalemmi_001
vgalemmi_002

---

## Vertical

mm2_000
mm2_001
mm2_002
mm2_003
scorch_000
scorch_001
scorch_002
vgalemmi_000
vgalemmi_001

johnnovak:
> I don't think it's possible to set deprecation message in such way.

Come on, man! 😄 If there is a will, there is a way 😎  Something like this will do:

```cpp
	if (output == "openglpp") {
		LOG_WARNING("SDL: Output mode 'openglpp' has been removed, defaulting to 'opengl'; "
		            "consider using the 'integer_scaling' setting instead");
	} else if (output == "texturepp") {
		LOG_WARNING("SDL: Output mode 'texturepp' has been removed, defaulting to 'texture'; "
		            "consider using the 'integer_scaling' setting instead");
	}
```

In action:

image

README.md
```diff
@@ -51,7 +51,6 @@ support today's systems.
 | **Feature**                 | **Status**
 |-                            |-
 | **CD-DA file codecs**       | Yes: Opus, OGG/Vorbus, MP3, FLAC, and WAV
-| **Pixel-perfect mode**      | Yes: `output=openglpp` or `output=texturepp`
```
johnnovak, line 54: I was secretly hoping you would mention `integer_scaling` here, but we can do that later... 😄

src/dosbox.cpp
johnnovak: > I don't want to touch configuration ever again.
> 
> Should work now, but setting it at runtime requires manual triggering of viewport resize. Not quite sure yet how to do it at least remotely right.

You'll get used to it. If you think this is hard, imagine what I went through when I added the chorus, reverb, crossfeed & filters to the mixer AND the mixer commands to manipulate these all in realtime... 😅

src/gui/sdlmain.cpp
johnnovak: Minor thing, but bump 😄

src/gui/sdlmain.cpp
johnnovak: Bump, bump 😄

src/gui/sdlmain.cpp
```diff
@@ -3349,25 +3277,19 @@ static SDL_Rect calc_viewport_fit(const int win_w, const int win_h)
...
-	else
-		return calc_viewport_fit(width, height);
+	if (new_mode == "off") {
+		sdl.integer_scaling_mode = IntegerScalingMode::Off;
+	} else if (new_mode == "horizontal") {
+		sdl.integer_scaling_mode = IntegerScalingMode::Horizontal;
+	} else if (new_mode == "vertical") {
+		sdl.integer_scaling_mode = IntegerScalingMode::Vertical;
+	} else {
+		LOG_WARNING("Unknown integer scaling mode '%s', defaulting to 'off'",
```
johnnovak, line 3289: Prefix log message with `RENDER: ` please.

kcgen:
Use `git grep vgaonly` to see how SetDeprecatedWithAlternateValue is used. 

Basically, eliminate "openglpp" as an option everywhere, and then use pstring->SetDeprecatedWithAlternateValue(...)  to register its alternative value, exactly the remains of "vgaonly".

If you set vgaonly, this warning is logged:

```
2023-05-10 22:28:21.469 | CONFIG:
'machine = vgaonly' is deprecated,
falling back to the alternate:
 'machine = svga_paradise'
```

src/dosbox.cpp
kcgen: I don't think we have on-the-fly adjustable windowresolution or full screen toggling (via conf settings), either.  

Although they can be set at runtime, their values only kick in when the context is replaced (by changing output = ..)

johnnovak:
> `git grep vgaonly` to see how SetDeprecatedWithAlternateValue is used.
> 
> Basically, eliminate "openglpp" as an option everywhere, and then use pstring->SetDeprecatedWithAlternateValue(...) to register its alternative value, exactly like the remains of "vgaonly".
> 
> If you try `machine = vgaonly`, this warning is logged:
> 
> ```
> 2023-05-10 22:28:21.469 | CONFIG:
> 'machine = vgaonly' is deprecated,
> falling back to the alternate:
>  'machine = svga_paradise'
> ```

That's cool; I guess I like the extra helpful message I included in the custom message in my example. Either way, we should log a deprecation warning _somehow_; we've been pretty good at logging deprecation warnings not just dropping features without warning out of nowhere, so we should continue this trend.

README.md
ThomasEricB, line 54: @GranMinigun , I don't think should commit a change to readme.md since it still applies to the latest release. We should wait to change the entire documentation in bulk for 0.81.

ThomasEricB (changes requested):

README.md
johnnovak: Fair point, probably best to leave it for now.

ThomasEricB: > Remove/update references to texturepp and openglpp in README and [README.md](http://readme.md/).

@johnnovak , I think we should wait before changing any of the documentation in the readme.md because it still applies for the latest release ... That's why for instance I held off adding Voodoo to the [README.md](http://readme.md/) and I believe @kcgen did not add IMFC (IBM Music Feature Card) there as well.

johnnovak: Yeah but you get my point about updating everything else, right? All build instructions, etc. Those get updated whenever we change main.

ThomasEricB: > Well, on a second thought, we're updating all the .md files in the main branch to reflect the "current dev" state (e.g. build instructions), so maybe it's ok to change it, after all? @kcgen? But the [README.md](http://readme.md/) everybody sees when goes to our repo is a bit special, admittedly.

@johnnovak Okay, so I had to look into my private messages with @kcgen just be sure I wasn't remembering things wrong. A couple of weeks ago, I asked @kcgen about updating the User Section of the README.md to add some of the new features, and he told me that updating the README.md was part of the release steps. And looking back, that makes 100% sense.

In my humble opinion, the README.md specifically should represent the state of the project at the latest stable release. Besides, it is better to review and discuss the changes bulk Pull Request with all the additions and removals to the sections than if we were to do it by small commits inside multiple PRs. In doing that, we may overload the front page with a lot of smaller features, or we may keep outdated information by mistake.

ThomasEricB: At the end of the day, I hope I'm not being too pedantic about this. I'll side with whatever @kcgen and @dreamer decides to do about this. However, I feel like this should be another addition to the [CONTRIBUTING.md](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md), just to prevent this happening again.

ThomasEricB: Made the PR https://github.com/dosbox-staging/dosbox-staging/pull/2470 so we can easily prevent stuff like this happening in the future.

dreamer: > > Well, on a second thought, we're updating all the .md files in the main branch to reflect the "current dev" state (e.g. build instructions), so maybe it's ok to change it, after all? @kcgen? But the [README.md](http://readme.md/) everybody sees when goes to our repo is a bit special, admittedly.
> 
> @johnnovak Okay, so I had to look into my private messages with @kcgen just be sure I wasn't remembering things wrong. A couple of weeks ago, I asked @kcgen about updating the User Section of the README.md to add some of the new features, and he told me that updating the README.md was part of the release steps. And looking back, that makes 100% sense.

Yes, it makes sense… Updating README.md *on the release branch*. That's also an opportunity to reassess if we haven't missed anything (as often happens). We might later cherry-pick that commit on the main branch, but it's much better to be proactive about it.

> In my humble opinion, the README.md specifically should represent the state of the project at the latest stable release.

Nope. This makes it harder for all developers. Documentation evolves with the code; if you see documentation not matching the code on the same branch, that means there is a bug. Inventing some additional rules about documentation on branch A representing project state on branch B makes everything unnecessarily complicated.

> Besides, it is better to review and discuss the changes in bulk, with all the additions and removals to the sections, than if we were to do it by small commits inside multiple PRs. (…)

Again, I do not agree - it's much better to have incremental changes (also in documentation), and then have separate PR to reedit them as necessary.

> In doing that, we may overload the front page with a lot of smaller features, or we may keep outdated information by mistake.

You have a point here - our README.md file again creeped out to be overloaded.

I think we no longer need "summary of features" table (at least not in README) - we have "Feature highlights" on https://dosbox-staging.github.io/ already. Perhaps we should replace this table in README.md with a new page on webpage listing *all* the features? @kcgen @johnnovak @ThomasEricB what do you think?

dreamer: In this context: I think "Pixel-perfect mode" line in table should not be removed, but rather edited to say something like:

> Pixel-perfect mode: Replaced by integer viewport scaling (link to discussions).

ThomasEricB: > I think we no longer need "summary of features" table (at least not in README) - we have "Feature highlights" on https://dosbox-staging.github.io/ already. Perhaps we should replace this table in README.md with a new page on webpage listing all the features? @kcgen @johnnovak @ThomasEricB what do you think?

While we slightly disagree (as you think the "summary of features" should reflect main while I think it should reflect the stable release), I agree with your solution. I think we should purge that section and replace it with a new page on webpage listing all the features.

kcgen: Yes, with the documentation site, we can  soon separate out the  notable changes from the readme (which was becoming almost like a project changelog) and make it more succinct (project description and quick build notes). 

My hunch is that the documentation that describes all the exposed interface will capture our features.

Every line in the conf file : they come from SVN or we changed them (and will be documented). 

All the new z: programs. 

New features that may not land in the above, but need documenting (like the virtual-host mouse integration). 

That's going to be a lot of stuff, however, that table-of-contents will almost act like the readme table we have now, where users can click and read more about the feature. 

Except with the documentation site, the TOC will take them to real documentation instead of PRs.

ThomasEricB: > But in the meantime - I agree we might as well add the current features to readme and make it reflect main. Sorry for the confusion I introduced, @ThomasEricB !

No problem. Relevant to @GranMinigun 's PR however, this still need changing before merging into main. To solve this, I think dreamer's solution is sensible: https://github.com/dosbox-staging/dosbox-staging/pull/2456#discussion_r1191266291

johnnovak: Pretty much agree with all you wrote @kcgen and @dreamer.

TL;DR: 

- README.md should evolve with the current dev version and should reflect the state of main.
- Until the manual is not published, the feature table in READMe.md could be updated. Once the manual is out, the table is no longer needed and should be removed.
- The feature highlights on the front page of the website is not a _comprehensive_ list of all features, just a list of most notable features to "entice" users. It should and will not include everything (it should still look good in the design); that role will be filled by the documentation that will be comprehensive (e.g. a full expanded TOC, or perhaphs an index will effectively list all we have, as @kcgen rightly pointed out).

johnnovak: So @GranMinigun , please update README.md with info about this feature.

johnnovak:
Hi @GranMinigun, how about finishing this? It's very close to the finish line and I can't wait to use it and update the guide 😄

johnnovak:
> P.S. What's the purpose of `sdl.draw.scalex/y` when `sdl.draw.pixel_aspect` exists and has far better precision?

Ok @GranMinigun... spent half the afternoon on it and figured it out 😅 `sdl.draw.pixel_aspect` was added specifically for the pixel-perfect feature when it was first introduced in this commit: d1be65b105de71 @kcgen you might wanna check this out too.

The problem with it is that it always calculates the pixel aspect assuming a 4:3 AR target rectangle. That might be okay 70-80% of the time, but in some more interesting cases (e.g. most if not all games that tweak standard resolutions into something else) this assumption yields incorrect results compared to `render.src.ratio` and the final calculated PAR that gets logged. That's because apparently the target render rectangle does _not_ always necessarily cover the whole 4:3 AR screen area, but could be smaller and not necessarily 4:3 ratio anymore (depending on how the VGA CRTC registers are set up in custom tweaked modes). A couple of examples and some explanations:

- `pixel_aspect` is the `sdl.draw.pixel_aspect`
- `render.src.ratio` (calculated based on the CRTC register states) is _always_ equal to our final calculated PAR with some minor error, as you can see (hence these two are always correct)
- Most notably, in the 1280x1024 standard VESA mode (which has 5:4 _image_ aspect ratio) pixels are supposed to be _square_, so the image should appear with slight black bars on the two sides. `render.src.ratio` and our final calculated PAR are correct in this case, but `pixel_aspect` gets calculated incorrectly assuming a 4:3 target rectangle, making the image squashed a bit...
- Same story for all the other modes... Quake is another good example where for example the 320x240 mode does _not_ have square pixels as you would think (and as the standard 320x240 VESA mode has), but the game messes around with the timings, resulting in an 1:1.022495 PAR... Weird and interesting!

Of course, in real life people can stretch the image on their CRT monitors with the hsize/vsize controls to pretty much anything manually, so these ratios can be slightly (or more) off anyway... But still, we're emulating an ideal, correctly set up monitor here.

Ok, concrete logs of values in a few interesting scenarios. `pixel_aspect` is underlined with red where it gets things wrong:

image

image

image

**Quake example**

Note the black bars on the two side of the image in 320x480 mode; it is _not_ stretched to the 4:3 AR window! (other custom modes are similar; almost none of them are perfect 4:3)

image

On the other hand, in the bog standard 320x200 (13h) mode, the image is 4:3 and fills the screen!

image

---

So that's for the explanation...

I've attached a patch that removes the `pixel_aspect` and the related wrong calculations. But we could actually keep it, put `render.src.ratio` there, and then simply use that instead of calculating the final PAR again from derived values, which would be simpler and remove the errors. If I'm right (and I think I am 😎 ), the final calculated PAR always equals `render.src.ratio`.

If you don't want to fuck around with this because you're sick of this PR I understand. I just thought you might be interested to learn about my findings because you asked the question originally. I'm happy to do these changes/fixes myself later once your PR has been merged, so if you don't want this to drag on, that's completely fine!

The patch:

[remove-pixel-aspect.patch](https://github.com/dosbox-staging/dosbox-staging/files/11471419/remove-pixel-aspect.patch)

src/gui/sdlmain.cpp
GranMinigun: GitHub doesn't update these previews, better check commits directly.

src/dosbox.cpp
GranMinigun: > You'll get used to it.

No.

GranMinigun:
> I saw you tried doing it with `Pstring->SetDeprecatedWithAlternateValue("openglpp", "opengl");`, but that doesn't log anything, apparently, and I'm not even sure it's working correctly...

Cannot confirm, it does work on my side. Please double-check that you've pulled in the latest push. If it still doesn't work, file a complain to the ARB.

My comment was specifically about hinting at the new setting, not setting deprecation warning at all. With the removal of pixel-perfect modes from the list of available values, there's no chance for the said values to pass configuration step, and so no trivial way to set a custom deprecation message.

> The problem with it is that it always calculates the pixel aspect assuming a 4:3 AR target rectangle.

Oh great... Alright, will address that. Thanks for looking into it.

GranMinigun:
> If I'm right (and I think I am sunglasses ), the final calculated PAR always equals `render.src.ratio`.

It's actually slightly off as well due to this ancient chunk:

https://github.com/dosbox-staging/dosbox-staging/blob/b982791d093bbb48951ecc81523bdd361f21220e/src/gui/render.cpp#LL484C2-L486C35

kcgen:
> ```c++
> height * ratio + 0.025
> ```

Interesting find (dug up background: 37fac15aea596c42ed2ae4184243b42db83a4be3 and 75da889c3edb66224bebc921d293c4ef85d6a6b1)

README.md
GranMinigun: Will the new line do?

johnnovak:
> Hmm.. substituting `height * ratio + 0.025` into `target`:
> 
> ```
> ratio = (height * ratio + 0.025) / height;
> ```
> 
> then
> 
> ```
> ratio = (height * ratio) / height + 0.025 / height;
> ```
> 
> then
> 
> ```
> ratio = ratio + 0.025 / height;
> ```
> 
> then
> 
> ```
> ratio += 0.025 / height;
> ```
> 
> .. or one 40th of height:
> 
> ```
> ratio += height / 40.0;
> ```
> 
> (background: [37fac15](https://github.com/dosbox-staging/dosbox-staging/commit/37fac15aea596c42ed2ae4184243b42db83a4be3) and [75da889](https://github.com/dosbox-staging/dosbox-staging/commit/75da889c3edb66224bebc921d293c4ef85d6a6b1))

Please kill this with fire to... 😩

kcgen:
@GranMinigun : I reviewed commit-by-commit of the latest, and all comments above appear to be addressed.

Approving, and will do some quick testing.

kcgen (approved):

README.md
johnnovak: You mean CR/LF? 😄 We can update this later, nevermind.

johnnovak: Yes, nice, thanks.

include/sdlmain.h
```diff
@@ -100,12 +100,12 @@ struct SDL_Block {
 	bool resizing_window = false;
 	bool wait_on_error = false;
 	SCALING_MODE scaling_mode = SCALING_MODE::NONE;
+	IntegerScalingMode integer_scaling_mode = IntegerScalingMode::Off;
 	struct {
 		int width = 0;
 		int height = 0;
 		double scalex = 1.0;
 		double scaley = 1.0;
-		double pixel_aspect = 1.0;
```
johnnovak, line 108: Good riddance!

johnnovak:
Works very well, @GranMinigun and the last few cleanup commits look good!

Double-scanned EGA on VGA with my tweaked version of the super nice Hyllian CRT shader looks _glorious_! 🎉 
Integer-upscaled vertically to 2000 on 4k.

```
2023-05-16 09:37:53.387 | DISPLAY: VESA 640x400 32-bit (mode 213h) at 69.755 Hz CFR, scaled to 2667x2000 with 1.2 pixel aspect ratio
```

image

johnnovak (approved):
Thanks @GranMinigun; once CI is passing please merge away 🚀

PR #2457, opened 2023-05-05
Adds full file path logging to the internal debugger commands that lo…

LowLevelMahn:
…ad/create files


src/debug/debug.cpp
```diff
@@ -73,6 +78,23 @@ static void DrawVariables(void);
 char* AnalyzeInstruction(char* inst, bool saveSelector);
 uint32_t GetHexValue(char* str, char*& hex);
 
+string get_full_path(const string& rel_or_abs_path)
```
dreamer, line 81: I think this whole implementation can be replaced by [`std::filesystem::absolute`](https://en.cppreference.com/w/cpp/filesystem/absolute)?

Aside from that, please mark any function intended to be static (like this one) as `static` - it helps a lot.

src/debug/debug.cpp
```diff
@@ -1226,11 +1248,14 @@ bool ParseCommand(char* str) {
 
 	if (command == "logcode") { //Shared code between all logs
 		DEBUG_ShowMsg("DEBUG: Starting log\n");
-		cpuLogFile.open("LOGCPU.TXT");
+		const string log_cpu_txt = "LOGCPU.TXT";
```
dreamer, line 1251: Use `std_fs::path`. Ditto for all other relative paths in this patch.

src/debug/debug.cpp
```diff
@@ -2308,7 +2333,13 @@ bool CDebugVar::SaveVars(char *name)
 	if (varList.size() > 65535) return false;
 
 	FILE* f = fopen(name,"wb+");
-	if (!f) return false;
+	if (!f) {
+		DEBUG_ShowMsg("DEBUG: Output of vars failed.\n");
+		return false;
+	}
+	DEBUG_ShowMsg("DEBUG: vars file %s created.\n",
```
dreamer, line 2340: When showin a path to any user-readable location, I think we should surround it with a single quotes, like this: `'%s'`. This makes it more readable if the path happens to include whitespace characters.

We should be more strict about this, I see some paths not surrounded by quotes in our log.

dreamer (changes requested):
Looks good, but can be improved by switching to C++17 `std::filesystem`.

Please note, not all platforms we support have full C++17 support yet, that's why use it this way:

```C++
#include "std_filesystem"

const std_fs::path memdump_txt = "MEMDUMP.TXT";
```

Also, please amend commit message, refer to https://cbea.ms/git-commit/ and look specifically at rules (2) and (5).

Issue #1986, opened 2022-10-02
Missing mapping of FreeDOS code page characters to Unicode


FeralChild64:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.0-alpha

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The main branch has just switched to UTF-8 locale. For this to work, the DOSBox Staging Unicode engine has to know how to map the DOS code page characters to Unicode graphemes. Currently the mappings are provided for all the FreeDOS code pages (as we are using FreeDOS CPX files for screen fonts), but there are few missing points:

- **864** (Arabic) - unidentified characters: **0x9b**, **0x9c**, **0x9f**, **0xa6**, **0xa7**
- ~~30020 (Low Saxon and Frisian) - possibly misidentified characters: 0xa4, 0xa5, 0xc6, 0xc7~~ EDIT: charactters identified dosbox-staging/dosbox-staging/pull/1988
- **59829** (Georgian) - unidentified character: **0xa6** (most likely a SHIN letter with some accent - but which one?)

Once the characters are identified, mappings should be updated in file:
https://github.com/dosbox-staging/dosbox-staging/blob/main/contrib/resources/mapping/MAIN.TXT
and, additionally, fallback mapping to a 7-bit ASCII character set shall be added to:
https://github.com/dosbox-staging/dosbox-staging/blob/main/contrib/resources/mapping/ASCII.TXT

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

FeralChild64:
Unidentified characters isolated on screenshots:

![HelpNeeded_864](https://user-images.githubusercontent.com/48332137/193514808-a36b0609-c3cf-45e6-8e22-35ea2bee6451.png)

![HelpNeeded_59829](https://user-images.githubusercontent.com/48332137/193514817-28bcdc20-42e2-4947-a4cb-4fa8b4602ba0.png)

FeralChild64:
4 arabic characters identified, see https://github.com/dosbox-staging/dosbox-staging/pull/2458

Still to identify:
- character `0x9f` from code page 864 (Arabic)
- character `0xa6` from code page 59829 (Georgian)

johnnovak:
Going into extreme niche territory here, but okay :)

Burrito78:
[...]

PR #2458, opened 2023-05-05
Improve Arabic support in Unicode engine

FeralChild64:
- identified 4 more arabic characters; 1 still unidentified
- adjusted range of recognized combining marks to prevent errors when switching to code page 864

Needed for https://github.com/dosbox-staging/dosbox-staging/issues/1986 (the fix here is partial).


kcgen (approved):
Four more unidentified characters down 👍.
Will merge when passing CI.

PR #2464, opened 2023-05-07
Implement SETVER.EXE command

FeralChild64:
This PR implements a `SETVER.EXE` utility, with nearly all extensions found in DR-DOS (this is the only extended version known to me; Microsoft has ditched this tool in modern Windows, so no more extensions should be expected from them).

The `ver set VERSION` command is "deprecated", but still mentioned in the documentation, I do not plan to remove it, ever. Unfortunately, this `ver` extended syntax isn't similar to other DOS commands at all, and it goes against the overall DOS design - but I guess we have to live with it - for sake of compatibility.

Technical reading regarding DOS version reporting and SETVER utility:
[https://www.geoffchappell.com/notes/dos/interrupts/21h/30h/index.htm](https://www.geoffchappell.com/notes/dos/interrupts/21h/30h/index.htm)

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -29,12 +29,12 @@
 #include "callback.h"
 #include "drives.h"
 #include "mem.h"
+#include "program_mount_common.h"
 #include "regs.h"
 #include "serialport.h"
 #include "setup.h"
 #include "string_utils.h"
 #include "support.h"
-#include "program_mount_common.h"
 
 #if defined(WIN32)
 #include <winsock2.h> // for gethostname
@@ -659,16 +659,24 @@ static Bitu DOS_21Handler(void) {
 		SegSet16(es,RealSegment(dos.dta()));
 		reg_bx=RealOffset(dos.dta());
 		break;
-	case 0x30:		/* Get DOS Version */
-		if (reg_al==0) reg_bh=0xFF;		/* Fake Microsoft DOS */
-		if (reg_al==1) reg_bh=0x10;		/* DOS is in HMA */
-		reg_al=dos.version.major;
-		reg_ah=dos.version.minor;
-		/* Serialnumber */
-		reg_bl=0x00;
-		reg_cx=0x0000;
-		break;
-	case 0x31:		/* Terminate and stay resident */
+	case 0x30: /* Get DOS Version */
+	{
+		if (reg_al == 0) {
+			reg_bh = 0xff; // 0xff for MS-DOS, 0x00 for PC-DOS
+		}
+		if (reg_al == 1) {
+			reg_bh = 0x10; // DOS is in HMA
+		}
+		DOS_PSP psp(dos.psp());
+		// Prior to MS-DOS 5.0 version number was hardcoded, later DOS
+		// releases used values from PSP, for SETVER functionality
+		reg_al = psp.GetVersionMajor();
+		reg_ah = psp.GetVersionMinor();
+		// Serial number
+		reg_bl = 0x00;
+		reg_cx = 0x0000;
+	} break;
+	case 0x31: /* Terminate and stay resident */
 		// Important: This service does not set the carry flag!
 		DOS_ResizeMemory(dos.psp(),&reg_dx);
 		DOS_Terminate(dos.psp(),true,reg_al);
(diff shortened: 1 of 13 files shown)


src/dosbox.cpp
```diff
@@ -1169,6 +1169,12 @@ void DOSBOX_Init()
 	pstring->Set_help(
 	        "Language code of the keyboard layout, or 'auto' ('auto' by default).");
 
+	pstring = secprop->Add_path("setver_table_file", only_at_start, "");
```
dreamer, line 1172: Maybe we should just pick a name for the database file, without giving users option for changing the path?

src/dos/program_setver.cpp
```diff
@@ -0,0 +1,721 @@
...
+	if (has_arg_batch || has_arg_paged) {
+		WriteOut(MSG_Get("SHELL_WRONG_SYNTAX"));
+	} else if (params.size() > 2) {
+		WriteOut(MSG_Get("SHELL_TOO_MANY_PARAMETERS"));
+	} else {
+		CommandSet(params[0], params[1], has_arg_quiet);
+	}
+}
+
+bool SETVER::ParseVersion(const std::string& version_str, FakeVersion& version)
```
dreamer, line 188: I strongly suggest rewriting this function using https://en.cppreference.com/w/cpp/regex

src/dos/program_setver.cpp
```diff
@@ -0,0 +1,721 @@
...
+	// Do nothing if no file name specified in the configuration
+	const auto file_path = GetTableFilePath();
+	if (file_path.empty()) {
+		return;
+	}
+
+	// Write table into the file
+	std::ofstream file(file_path.c_str(), std::ios::trunc);
+	if (setver_table.is_global_version_set) {
+		file << "\t" << setver_table.version_global << std::endl;
```
dreamer, line 656: Ugh, using `std::endl` will make this file format non-portable; use Unix line endings for all OSes, please.

dreamer (changes requested):

src/dosbox.cpp
FeralChild64: I would prefer to have such option, as I am storing the database inside my virtual drive C: - this makes archiving my guest system slightly easier, and I can check version override using DOS Navigator.

src/dos/program_setver.cpp
FeralChild64: Done.

src/dos/program_setver.cpp
FeralChild64: Done. The last time I tried to use regular expressions in DOSBox Staging reviewer asked me to migrate away from them. We'll see...

src/dos/dos.cpp
```diff
@@ -700,29 +708,32 @@ static Bitu DOS_21Handler(void) {
...
+			        break;
+		        }
+		        break;
+	case 0x34: /* Get INDos Flag */
+		        SegSet16(es, DOS_SDA_SEG);
+		        reg_bx = DOS_SDA_OFS + 0x01;
+		        break;
+	case 0x35: /* Get interrupt vector */
+		        reg_bx = real_readw(0, ((uint16_t)reg_al) * 4);
+		        SegSet16(es, real_readw(0, ((uint16_t)reg_al) * 4 + 2));
```
kcgen, line 732: I'm having trouble seeing context here - is the indenting on line 725 and prior due to an inner switch statement inside of an outer case?

(Just mentioning it - sometimes the clang formatter can incorrectly format switches)

src/dos/dos.cpp
```diff
@@ -700,29 +708,32 @@ static Bitu DOS_21Handler(void) {
...
+		        break;
+	case 0x36: /* Get Free Disk Space */
+	{
+		        uint16_t bytes,clusters,free;
 			uint8_t sectors;
 			if (DOS_GetFreeDiskSpace(reg_dl,&bytes,&sectors,&clusters,&free)) {
```
kcgen, line 738: (clang formatter might be off here)

src/dos/program_setver.cpp
```diff
@@ -0,0 +1,744 @@
...
+static struct {
+	SETVER::FakeVersion version_global = {};
+	bool is_global_version_set         = false;
+
+	NameVersionTable by_file_name = {
+	        // Since MS-DOS 6.22 has some default version table,
+	        // we can provide some sane defaults too
+	        {"WIN100.BIN", {3, 40}}, // fixes Microsoft Windows 1.x
+	        {"WIN200.BIN", {3, 40}}, // fixes Microsoft Windows 2.x
+	};
```
kcgen, line 49: excellent!

src/dos/program_setver.cpp
```diff
@@ -0,0 +1,744 @@
...
+}
+
+bool SETVER::PreprocessName(std::string& name, const bool permissive)
+{
+	// Preprocess file names and relative paths
+	trim(name);
+	if (IsNameWithPath(name)) {
+		char buffer[DOS_PATHLENGTH];
+		if (!DOS_Canonicalize(name.c_str(), buffer)) {
+			return permissive;
```
kcgen, line 241: Can "permissive" be named better? (Even in the function body, it doesn't permit or deny any code paths, and instead only controls the return values if we failed to canoicalize).

It seems like three return types are desired? fully processed, couldn't canonicalize, and failed/illegal. (maybe an enum return is better than the taking in the permissive bool?) -- just throwing out ideas if they can help.

MasterO2:
Whenever `setver` is eventually merged into Dosbox Staging, the game below will be a good test for this newly supported command: 

https://github.com/dosbox-staging/dosbox-staging/wiki/Game-issues#jetfighter-the-adventure

I assume `setvar jf.exe 3.10` would be the syntax.  (`jf.exe` is JetFighter's executable).

FeralChild64:
@kcgen True, we definitely need some protections against accidental wiping of the whole database.

@MasterO2 I would rather suggest something like `setver c:\games\jf\jf.exe 3.10` - this way we won't report old DOS version to some other software with the same executable name.

BTW, MS-DOS didn't support the path in `setver` command, but DR-DOS (which I was using as a teenager) did.

src/dos/program_setver.cpp
FeralChild64: Changed variable name to `allow_non_existing_files`. Enum return requires a definition + more complicated result checking.

FeralChild64:
@kcgen I have changed the code to require `/all` switch. Also added information about `/delete` and `/quiet` being same as `/d` and `/q` to the command help (longer switch versions are for DR-DOS compatibility only).

kcgen:
PVS is flagging a couple new hits in `setver`:
```
src/dos/program_setver.cpp : 2
```

zip is at the bottom here: https://github.com/dosbox-staging/dosbox-staging/actions/runs/4929529750?pr=2464

FeralChild64:
@kcgen Should be fixed now.

src/dos/dos.cpp
FeralChild64: I have reformatted the code manually, in a separate commit for better bisectability.

src/dos/dos.cpp
FeralChild64: Reformatted manually, as above.

kcgen:
Can't see anything left; passing tests, CI, and sanitizers. 
Thanks @FeralChild64 !

PR #2471, opened 2023-05-11
Replace invalid vector.end dereference with raw pointer access

LowLevelMahn:
the MSVC STL asserts on end dereference in Debug mode, that happens if there is no data left to copy (size == 0)
the assert happens under Linux only with -D_GLIBCXX_DEBUG

### DIFF
--- a/src/dos/drive_virtual.cpp
+++ b/src/dos/drive_virtual.cpp
@@ -346,15 +346,16 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
 	open = true;
 }
 
-bool Virtual_File::Read(uint8_t * data,uint16_t * size) {
-	uint32_t left = file_data->size() - file_pos;
-	if (left <= *size) {
-		memcpy(data, &((*file_data)[file_pos]), left);
-		*size = (uint16_t)left;
-	} else {
-		memcpy(data, &((*file_data)[file_pos]), *size);
-	}
-	file_pos += *size;
+bool Virtual_File::Read(uint8_t* data, uint16_t* bytes_requested)
+{
+	assert(file_pos <= file_data->size());
+	size_t bytes_to_read = *bytes_requested;
+	const size_t bytes_left = file_data->size() - file_pos;
+	const uint8_t* src = file_data->data() + file_pos;
+	bytes_to_read = std::min(bytes_left, bytes_to_read);
+	memcpy(data, src, bytes_to_read);
+	file_pos += bytes_to_read;
+	*bytes_requested = static_cast<uint16_t>(bytes_to_read);
 	return true;
 }
 


src/dos/drive_virtual.cpp
```diff
@@ -348,11 +348,12 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
 
 bool Virtual_File::Read(uint8_t * data,uint16_t * size) {
 	uint32_t left = file_data->size() - file_pos;
+	uint8_t* src  = file_data->data() + file_pos;
```
dreamer, line 351: Nice :smile:. Add an assert for verification if `file_pos` is small enough to not not trigger out of bounds memory access. I guess it does not, but if someone will change code around then we want this to blow up as soon as possible.

dreamer:
@kcgen maybe we should add `_GLIBCXX_DEBUG` to on of our builds to see if we can trigger this assert in CI (maybe clang static analysis build will manage to trigger this?).

dreamer:
@LowLevelMahn update the commit message according to [rule 5](https://cbea.ms/git-commit/#imperative), please.

LowLevelMahn:
> @LowLevelMahn update the commit message according to [rule 5](https://cbea.ms/git-commit/#imperative), please.

so "replaces" becomes "Replace"?

LowLevelMahn:
> @kcgen maybe we should add `_GLIBCXX_DEBUG` to on of our builds to see if we can trigger this assert in CI (maybe clang static analysis build will manage to trigger this?).

does the CI also start dosbox (its only triggered at runtime)?

LowLevelMahn:
there is also _GLIBCXX_SANITIZE_VECTOR (see: https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html)

> When defined, std::vector operations will be annotated so that AddressSanitizer can detect invalid accesses to the unused capacity of a std::vector.

and the libc++ from the LLVM project also got sanitizing for std::string, std::vector, std::deque and some others

src/dos/drive_virtual.cpp
```diff
@@ -348,11 +348,13 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
 
 bool Virtual_File::Read(uint8_t * data,uint16_t * size) {
 	uint32_t left = file_data->size() - file_pos;
+	assert(file_pos <= file_data->size());
+	uint8_t* src  = file_data->data() + file_pos;
```
kcgen, line 352: `const`

src/dos/drive_virtual.cpp
```diff
@@ -348,11 +348,13 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
...
 	if (left <= *size) {
-		memcpy(data, &((*file_data)[file_pos]), left);
+		memcpy(data, src, left);
 		*size = (uint16_t)left;
 	} else {
-		memcpy(data, &((*file_data)[file_pos]), *size);
+		memcpy(data, src, *size);
 	}
 	file_pos += *size;
 	return true;
```
kcgen, line 360: The if/else branches are confusing; maybe another commit refactoring to something like:

```c++
bool Virtual_File::Read(uint8_t * data, uint16_t * bytes_requested) {
  assert(file_pos <= file_data->size());
  const uint32_t bytes_left = file_data->size() - file_pos;
 
  const auto num_bytes = std::min(bytes_left, static_cast<uint32_t>(bytes_requested));

  const auto source = file_data->data() + file_pos;

  memcpy(data, source, num_bytes);
}

File: src/dos/drive_virtual.cpp [hunk 4]
```diff
@@ -348,11 +348,13 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
 
 bool Virtual_File::Read(uint8_t * data,uint16_t * size) {
 	uint32_t left = file_data->size() - file_pos;
+	assert(file_pos <= file_data->size());
```
kcgen, line 351: Suggest asserting before using.

src/dos/drive_virtual.cpp
```diff
@@ -346,15 +346,14 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
...
-	} else {
-		memcpy(data, &((*file_data)[file_pos]), *size);
-	}
-	file_pos += *size;
+bool Virtual_File::Read(uint8_t* data, uint16_t* bytes_requested)
+{
+	assert(file_pos <= file_data->size());
+	const uint32_t bytes_left = file_data->size() - file_pos;
+	const uint8_t* src  = file_data->data() + file_pos;
+	*bytes_requested = std::min((uint16_t)bytes_left, *bytes_requested);
```
kcgen, line 354: Are these files only 16bit in length?

If not, casting down to 16-bit will wrap bytes left back to 0 at "max-uint16 + 1".

I think we want file sizes to be 64bit (HDD images, being the main use of 64bit sizes).

So maybe store the bytes left in an explicit unint64_t, and also cast up the incoming required bytes up to match.

LowLevelMahn: `Virtual_File::Read` gets called by `bool DOS_ReadFile(uint16_t entry,uint8_t * data,uint16_t * amount,bool fcb)` which i think is the implementation for DOS int 21h/AH=3Fh Read file API which does not support larger reads then 16bit values

PR #2473, opened 2023-05-11
Add package download URL to issue template

kcgen:
### The issue being addressed

Many issues have been raised where the maintenance team was unable to reproduce the issue because the software being tested differed slightly from that used by the user reporting the issue.

### What this PR changes

If the issue affects a single game or program, the user is asked to input a download URL to the binary-exact package to ensure we are working with the same files and version(s).

Archive.org is given as the default example host, and given it's cleared to host all DOS preservation suites, there should be no reason why this cannot be provided.

If a user does not provide a download link for a game-specific issue, then maintainers should simply close out the ticket straight away.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -93,6 +93,16 @@ body:
     validations:
       required: false
   - type: textarea
+    id: package_url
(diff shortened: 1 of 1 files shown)


johnnovak (approved):
Great addition @kcgen!

PR #2476, opened 2023-05-13

Kappa971:


kcgen (approved):
That's a thorough overhaul, @Kappa971 🎉

Kappa971:
I don't know if it's thorough, but I hope it's good 🙂
I don't know how large the Italian community that uses DOSBox Staging is.

Also I see many sites/videos that still recommend DOSBox 0.74-3 and, with all due respect to DOSBox, I don't understand why 😅.

PR #2477, opened 2023-05-13
Disable framebuffer acceleration for debugger

### LINKED ISSUE #1824: SDL-backed PDCurses debugger is not compatible with OpenGL
Reproduce with steps:

1. **Setup a debugger build with OpenGL forced on:**

    ``` shell
    meson setup -Duse_opengl=true -Denable_debugger=normal \
        -Dunit_tests=disabled -Db_ndebug=true build/gl_debugger
    ```

2. **Confirm the debugger and OpenGL are both enabled:**

    ```
      Build Summary
        OpenGL support      : True
        ...
      User defined options
        b_ndebug            : true
        enable_debugger     : normal
        use_opengl          : true
    ```

3. **Build:**
   
    ``` shell
    meson compile -C build/gl_debugger
    ```

4. Launch the debugger build inside this package: [shamus.zip](https://github.com/dosbox-staging/dosbox-staging/files/9226836/shamus.zip)

On startup, the debugger window appears OK:

Tap F10 to step ahead by some instructions which triggers a GL video-mode change that impacts the debugger window:

Here's the state of the debugger after the mode change:

Screenshot from 2022-07-30 08-52-05

What's interesting is that the debugger is still alive.

GranMinigun:
Single-line change, some details in the commit message. Only affects outputs that directly access window surface (so surface output, mapper, and debugger).

Closes #1824.

### DIFF
--- a/src/libs/PDCurses/sdl2_queue/pdcscrn.cpp
+++ b/src/libs/PDCurses/sdl2_queue/pdcscrn.cpp
@@ -277,6 +277,9 @@ int PDC_scr_open(void)
         env = getenv("PDC_COLS");
         pdc_swidth = (env ? atoi(env) : 80) * pdc_fwidth;
 
+        /* Workaround to not disrupt OpenGL context state in other windows. */
+        SDL_SetHint(SDL_HINT_FRAMEBUFFER_ACCELERATION, "0");
+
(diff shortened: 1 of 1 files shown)


kcgen (approved):
Confirmed working; nice point-fix, @GranMinigun !

PR #2480, opened 2023-05-14
Don't auto-disable OpenGL for debugger-enabled builds on Linux

kcgen:
Now that @GranMinigun has fixed OpenGL when used with the debugger, this PR reverts the exclusion of OpenGL in debugger builds.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -549,29 +549,10 @@ summary('SDL_image 2.0 support', sdl2_image_dep.found())
 
 # OpenGL
 opengl_dep = optional_dep
-opengl_summary_msg = 'Disabled'
-if get_option('use_opengl') != 'false'
+if get_option('use_opengl')
     opengl_dep = dependency('gl', not_found_message: msg.format('use_opengl'))
 endif
-if (
-    opengl_dep.found()
-    and host_machine.system() == 'linux'
-    and get_option('use_opengl') == 'auto'
-    and get_option('enable_debugger') != 'none'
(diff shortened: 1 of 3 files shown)


kcgen:
Allowing OpenGL in debugger-enabled builds opened up more code paths (which in turn resulted in more warnings), so this PR had to squash those.

@GranMinigun , I tried to be as minimal as possible to resude the rebase-pain for both your active sdlmain PRs.

meson.build
```diff
@@ -549,29 +549,9 @@ summary('SDL_image 2.0 support', sdl2_image_dep.found())
...
-
-    warning(
-        '''OpenGL is disabled because it's not compatible with the debugger.
-            To use OpenGL with the debugger, force it with "-Duse_opengl=true"''',
-    )
-
-    conf_data.set10('C_OPENGL', false)
-    opengl_summary_msg = 'Found, but disabled due to conflict with debugger'
-else
-    conf_data.set10('C_OPENGL', opengl_dep.found())
```
GranMinigun, line 571: This cuts out OpenGL completely.

kcgen: In that case, we can get rid of a lot more code :laughing: j/k

(thanks for this big catch.. re-adding)

GranMinigun (approved):

kcgen:
@GranMinigun , thanks for the checks.  Will merge this after the integer PR goes in.

PR #2482, opened 2023-05-15
Refactor shell_misc.cpp

MeAreJeenius:


src/dos/dos_files.cpp
```diff
@@ -935,8 +935,8 @@ static bool isvalid(const char in){
 #define PARSE_RET_BADDRIVE      0xff
 
 // TODO: Refactor and document this function until it's understandable
-uint8_t FCB_Parsename(uint16_t seg,uint16_t offset,uint8_t parser ,char *string, uint8_t *change) {
-	char * string_begin=string;
+uint8_t FCB_Parsename(uint16_t seg,uint16_t offset,uint8_t parser ,const char *string, uint8_t *change) {
+	const char * string_begin=string;
 	uint8_t ret=0;
```
kcgen, line 940: If you're able spit into more commits, running `./scripts/format-commit.sh` on each commit will get these lines formatted a bit better too, given we're touching them (it will add spaces between `=`)

kcgen:
Thanks for the finer grained commits, @MeAreJeenius :+1:  Code looks great. Will test this as much as possible today and prepare for merge.

src/shell/shell_misc.cpp
```diff
@@ -72,331 +74,255 @@ void DOS_Shell::ShowPrompt(void) {
...
+			constexpr uint8_t ShiftPlusTab = 0x0F;
+			constexpr uint8_t F3           = 0x3D;
+			constexpr uint8_t Home         = 0x47;
+			constexpr uint8_t Up           = 0x48;
+			constexpr uint8_t Left         = 0x4B;
+			constexpr uint8_t Right        = 0x4D;
+			constexpr uint8_t End          = 0x4F;
+			constexpr uint8_t Down         = 0x50;
+			constexpr uint8_t Delete       = 0x53;
+			switch (data) {
```
johnnovak, line 146: Don't we already have all DOS keycodes defined somewhere? If not, could you perhaphs start a file with these common scancodes?

src/shell/shell_misc.cpp
```diff
@@ -72,331 +74,255 @@ void DOS_Shell::ShowPrompt(void) {
...
-					len = (uint16_t)it_history->length();
-					str_len = str_index = len;
-					size = CMD_MAXLINE - str_index - 2;
-					++it_history;
-					break;
+			case Down:
+				if (history_index < history_clone.size() - 1) {
+					history_clone[history_index] = command;
+					history_index++;
+					command = history_clone[history_index];
```
johnnovak, line 188: Minor: we prefer preincrement/predecrement where possible.

src/shell/shell_misc.cpp
```diff
@@ -32,6 +32,31 @@
 #include "string_utils.h"
 #include "../ints/int10.h"
 
+namespace {
+class Int10Manager {
```
johnnovak, line 34: Can you pick a better name that more accurately reflects what this class is doing? Just by looking at the name, it's reasonable to assume it "manages" all BIOS interrupt 10h functionality. Clearly, that's not the case here; it only deals with a rather small subset related to console I/O.

johnnovak:
Very nice cleanup / refactorisation job @MeAreJeenius! 🎉 If I'm not mistaken, this almost entirely deals with the interactive shell, so there's not much chance to cause regressions in batch file execution and command execution that could break existings configs.

Have you thought about how this change interacts with the AUTOTYPE command and non-English codepages?

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+	std::string command = ReadCommand();
+	history.emplace_back(command);
+
+	const auto* const dos_section = dynamic_cast<Section_prop*>(
+	        control->GetSection("dos"));
+	assert(dos_section != nullptr);
+	const std::string_view expand_shell_variable_pref = dos_section->Get_string(
+	        "expand_shell_variable");
+	if ((expand_shell_variable_pref == "auto" && dos.version.major >= 7) ||
+	    expand_shell_variable_pref == "true") {
```
kcgen, line 90: @johnnovak , I remember the improved config `bool` handling (where "1", "on", "true", and "yes" all are treated as positive and the opposites for negative).

I'm looking at:

```c++
bool Value::SetBool(const std::string& in)
{
	istringstream input(in);
	string result;
	input >> result;
	lowcase(result);
	_bool = false;

	if (!result.size()) {
		return false;
	}

	if (result == "0" || result == "disabled" || result == "false" ||
	    result == "off" || result == "no") {
		_bool = false;
	} else if (result == "1" || result == "enabled" || result == "true" ||
	           result == "on" || result == "yes") {
		_bool = true;
	} else {
		return false;
	}

	return true;
}
```

Ideally we could leverage that more robust bool handling here too (even though this is a conf string type instead of conf bool.

johnnovak: Yeah I guess finding & extracting my function that parses the config bools is in order then into string utils or something 😄 Good that you remembered this! 🎉

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+				if (history_index > 0) {
+					history_clone[history_index] = command;
+					history_index--;
+					command = history_clone[history_index];
+					cursor_position = command.size();
+				}
+				break;
 
+			case Down:
+				if (history_index < history_clone.size() - 1) {
```
kcgen, line 185: If `history_clone.size()` is empty, then subtracting 1 from the `size_t` will wrap it to the max value, in which case the `if` check would succeed.

Maybe need a `history_clone.empty()` pre-check?

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
-					/* Add executable list to front of completion list. */
-					std::copy(executable.begin(),executable.end(),std::front_inserter(l_completion));
-					it_completion = l_completion.begin();
-					dos.dta(save_dta);
+		}
+
+		case '\t': {
+			if (!completion.empty()) {
+				completion_index++;
+				if (completion_index == completion.size()) {
```
kcgen, line 221: Do we expect that the incremented `completion_index` will never exceed completion.size()?

If it's a valid risk, maybe: 
`if (completion_index >= completion.size())`

If we can be more certain, suggest asserting that prior to the if-check.

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+			viewing_tab_completions = true;
 			break;
-		case 0x1b: /* Esc */
-			while (str_index < str_len) {
-				outc(line[str_index++]);
+		}
+
+		case '\b':
+			if (cursor_position > 0) {
+				command.erase(cursor_position - 1, 1);
```
kcgen, line 248: Do we expect that the cursor_position not exceed the command's size()? Suggest asserting this prior to the erase. (This helps document reasonable expectations for readers too).

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+		}
 	}
 
-	if (!str_len) return;
-	str_len++;
+	return "";
+}
+
+namespace {
+std::vector<std::string> get_completions(const std::string_view command)
```
kcgen, line 284: We're not using this nameless `namespace` style in other refactored code.  Maybe `static` is sufficient to scope the function locally?

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+	std::vector<std::string> non_executables = {};
+	while (res) {
+		dta.GetResult(result);
+
+		if (!result.IsDummyDirectory() &&
+		    (!dir_only || result.IsDirectory())) {
+			if (is_executable_filename(result.name)) {
+				files.emplace_back(result.name);
+			} else {
+				non_executables.emplace_back(result.name);
```
kcgen, line 315: Maybe move the string's in: `.push_back(std::move(result.name))`.

I guess we'd have to also place the `DOS_DTA::Result result = {};` inside the while look prior to getting the result, just to start with a fresh object each time.

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+			}
+		}
+
+		res = DOS_FindNext();
 	}
+
+	dos.dta(save_dta);
+
+	files.insert(files.end(), non_executables.begin(), non_executables.end());
```
kcgen, line 324: suggest moving the non_executables content:

```c++
   files.insert(
       files.end(), 
       std::make_move_iterator(non_executables.begin()),
       std::make_move_iterator(non_executables.end()));
```

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+					completion_index = completion.size();
 				}
-				outc(' ');
-				move_cursor_back_one();
-				// moves the cursor left
-				while (str_remain--)
-					move_cursor_back_one();
+				completion_index--;
+				command.erase(completion_start);
+				command += completion[completion_index];
```
kcgen, line 208: Are we guarantee that `completion_index` isn't already 0 when decrementing?  Also, is it guaranteed that the decremented completion_index is in-bounds of the `completion` vector?

Maybe some assert or use `.at(..)`?

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+				auto delimiter = command.find_last_of(" \\");
+				if (delimiter == std::string::npos) {
+					completion_start = 0;
+				} else {
+					completion_start = delimiter + 1;
 				}
 			}
+
+			command.erase(completion_start);
+			command += completion[completion_index];
```
kcgen, line 240: Maybe assert completion_index is in-bound?

src/shell/shell_misc.cpp
```diff
@@ -501,234 +443,153 @@ bool DOS_Shell::Execute(char * name,char * args) {
...
-		reg_sp+=0x200;
-#if 0
-		reg_eip=oldeip;
-		SegSet16(cs,oldcs);
-#endif
+		bf   = std::make_shared<BatchFile>(this,
+                                                 fullname.c_str(),
+                                                 std::string(name).c_str(),
+                                                 std::string(args).c_str());
+		echo = temp_echo;
```
kcgen, line 466: (indented whitespace needs reformatting)

Maybe can use `name.data()` and `args.data()` directly from the string_views?

kcgen:
Testing is so far so good, exercising as much length, whitespace, history, and launching as possible.

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+
+	std::string search = {};
+	if (!args.empty() && command.back() != ' ') {
+		search += args.back();
 	}
+	search += "*.*";
+
+	auto save_dta = dos.dta();
+	dos.dta(dos.tables.tempdta);
+	DOS_DTA dta(dos.dta());
```
kcgen, line 299: (can be const)

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,421 @@
...
+		auto closing_percent = command.substr(1).find_first_of('%');
+
+		if (closing_percent == std::string::npos) {
+			break;
 		}
-		else {
-			text.replace(percent_pos, percent_len, "");
+		closing_percent += 1;
+
+		std::string env_key(command.substr(1, closing_percent - 1));
```
kcgen, line 344: (can be const)

src/shell/shell_misc.cpp
```diff
@@ -501,234 +443,153 @@ bool DOS_Shell::Execute(char * name,char * args) {
...
-	if (strcasecmp(extension, ".bat") == 0) 
-	{	/* Run the .bat file */
+	auto extension = fullname.substr(fullname.size() - extension_size);
+
+	if (iequals(extension, ".BAT")) {
 		/* delete old batch file if call is not active*/
-		bool temp_echo=echo; /*keep the current echostate (as delete bf might change it )*/
-		if (bf && !call)
+		/*keep the current echostate (as delete bf might change it )*/
+		bool temp_echo = echo;
```
kcgen, line 458: (can be const)

src/shell/shell_misc.cpp
kcgen: I'm totally neutral and just sticking to the existing pattern ; so unless we want to make a mass-switch, might as well carry on w/ `static` until there's a big improvement to be had.

Once C++20 or 23 gives us modules, we'll likely make whatever changes are needed en-mass to get the best bang for the buck.

src/shell/shell_misc.cpp
kcgen: > using c_str() is more correct, since data() doesn't guarantee null termination, which the called function needs.

roger that :+1:

kcgen:
> The bug I mentioned earlier is that an assertion fails when capturing an alt keycode in the terminal where the code > 255. For example, when you're in a debug build on the console, typing alt+1234 causes an assertion to fail

Great find!

> This is due to a call to check_cast casting a number greater than the max of the casted-to type. It probably wasn't found because the asserts aren't there in release builds, 

Yes, and most of us aren't using alt+codes, even in debug builds. 

> The desired behaviour for alt codes is alt_code == typed_code % 256 or alt_code == typed_code & 0xff, which a static cast to uint8_t effectively does for free

antizers include overflow checks (for both signed and unsigned), so if it's possible to include the % 256, that would be more illustrative for readers (and hopefully optimizing compilers elide `% 256` when casting unsigned to `uint8_t`)

johnnovak:
Thanks for the explanation @MeAreJeenius, great work! 🥳 The changes seem quite safe overall; I think we just to merge this and live with it for a while to let potential problems surface.

include/keyboard.h
```diff
@@ -99,6 +99,151 @@ enum KBD_KEYS {
...
+	SC_AltPageUp = 0x99,
+	SC_AltLeft   = 0x9B,
+	SC_AltRight  = 0x9D,
+	SC_AltEnd    = 0x9F,
+	SC_AltDown,
+	SC_AltPageDown,
+	SC_AltInsert,
+	SC_AltDelete,
+	SC_AltTab = 0xA5,
+};
```
johnnovak, line 245: Very nice! Is it possible to turn it into an enum class, though, then drop the `SC_` prefix? Then we can use proper enum scoping, e.g. `ScanCode::AltTab`.

src/shell/shell_misc.cpp
johnnovak: Are you planning to do this @MeAreJeenius ?

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,416 @@
...
-	uint16_t len=0;
-	bool current_hist=false; // current command stored in history?
+std::string DOS_Shell::ReadCommand()
+{
+	std::vector<std::string> history_clone = history;
+	history_clone.emplace_back("");
+	auto history_index = history_clone.size() - 1;
+
+	std::vector<std::string> completion = {};
+	auto completion_index               = completion.size();
```
johnnovak, line 106: Minor: insert newline above `completion_index`.

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,416 @@
...
-					}
-					break;
 
-				case 0x4B: /* Left */
-					if (str_index) {
-						move_cursor_back_one();
-						str_index --;
-					}
+		constexpr uint8_t ExtendedKey = 0x00;
+		constexpr uint8_t Escape      = 0x1B;
```
johnnovak, line 138: Minor: constexpr auto?

src/shell/shell_misc.cpp
```diff
@@ -20,479 +20,416 @@
...
-			strncat(line, args, CMD_MAXLINE - 2);
-			terminate_str_at(line, CMD_MAXLINE - 1);
-		} else {
-			safe_strcpy(line, args);
+CommandPrompt::CommandPrompt()
+{
+	SetCurrentPositionToZero();
+}
+
+uint32_t CommandPrompt::MaxCommandSize()
```
johnnovak, line 361: picky: uint8_t? uint16_t?

src/shell/shell_misc.cpp
```diff
@@ -501,234 +438,155 @@ bool DOS_Shell::Execute(char * name,char * args) {
...
-			if((name_len + len + 1) >= DOS_PATHLENGTH) continue;
+	return "";
+}
 
-			safe_strcat(path, name);
-			safe_strcpy(which_ret, path);
-			if (DOS_FileExists(which_ret))
-				return which_ret;
+std::string full_arguments = "";
+// TODO De-mystify magic numbers and verify logical correctness
```
johnnovak, line 504: That would be good, yes 😅

src/shell/shell_misc.cpp
johnnovak: Yeah, good idea to make it robust then 👍🏻

src/shell/shell_misc.cpp
johnnovak: Just to make the equals look nicer, yeah 😄 No big deal 😄

include/keyboard.h
johnnovak: Having said that, sometimes it's just more convenient to use namespaces instead (although we avoid namespaces in general, but I think in this very particular case they add value).

I've done this in `midi.h` where I just wanted nicely scoped constants (e.g. `MidiStatus::NoteOff`) but without the hassle of having to cast them to `uint8_t` all the time. Plus arguably these are not _really_ enums, in the sense that exhaustiveness-checking is basically meaningless in pretty much all use-cases.

TL;DR consider `namespace` too and use your judgement 😎 🤘🏻 

```cpp
namespace MidiStatus {
// Channel Voice Messages -- lower 4-bits (nibble) specify one of the 16 MIDI
// Channels (channel 1 = 0x0, channel 16 = 0xf)
constexpr uint8_t NoteOff         = 0x80;
constexpr uint8_t NoteOn          = 0x90;
constexpr uint8_t PolyKeyPressure = 0xa0;
constexpr uint8_t ControlChange   = 0xb0;
constexpr uint8_t ProgramChange   = 0xc0;
constexpr uint8_t ChannelPressure = 0xd0;
constexpr uint8_t PitchBend       = 0xe0;
...
```

src/shell/shell_misc.cpp
johnnovak: Yeah, you've done more than enough good work in this PR already 👍🏻 Gotta leave some of tasty bits for others too to clean up 😅  🍖

kcgen: [MeAreJeenius](https://github.com/MeAreJeenius), write access to the repository is inbound (check email).

You can now open branches in the repository under your initials, for example: `ab/feature-branch-[revision]`.  Your branch prefix (`ab/*`) will be write-protected for just you. 

Thank you for all the improvements!

kcgen (approved):
Local interactive tests are all still passing, and the remaining changes look good. Set to merge when they're ready. Thanks for this another big code quality upgrade, @MeAreJeenius :rocket:

src/shell/shell_misc.cpp
kcgen: Let's go w/ a separate refactoring and get this PR inbound!

kcgen:
@MeAreJeenius , starting at commit 28c2b083b (`Clean up and modernize DOS_Shell::Which()`), I get `double free or corruption (fasttop)` when launching `autotype 1`.  Unfortunately I didn't catch this in testing :sob: 

`git checkout 28c2b083b^` works, and `git checkout 28c2b083b` repros the crash.

johnnovak:
> @MeAreJeenius , starting at commit [28c2b08](https://github.com/dosbox-staging/dosbox-staging/commit/28c2b083bad9907a70a6844b2292dba492ed0cff) (`Clean up and modernize DOS_Shell::Which()`), I get `double free or corruption (fasttop)` when launching `autotype 1`. Unfortunately wasn't caught in testing.
> 
>     * `git checkout 28c2b083b` repros the crash.
> 
>     * `git checkout 28c2b083b^` works.

Related: I just learned that if you use `std::free`, freeing null pointers becomes a no-op. Maybe we should default to using that.

PR #2489, opened 2023-05-16
Drop Visual Studio 2019-specific accomodations in `.clang-format` 

kcgen:
These were added to accommodate VS2019, with the cost of now missing out on automatic formatting improvements, like adding brackets around `for`, `while`, and `if/else` bodies.

Because the majority of contributors are using Linux or macOS, and that Windows users can still run the formatter from MSYS2 or us VS2022 instead of VS2019, the cost of this change (loss of very valuable formatting) outweighs the benefit to Windows VS2019 users (who are few in number and still have other options to run the formatter).


johnnovak (approved):
Yeah it's for the best @kcgen.

PR #2494, opened 2023-05-17
Let the VESA 8-bit mode draw lines from the DAC palette

### LINKED ISSUE #2486: 7th Guest FMV flickering

DEV

Mac OS

Windows 10 and macOS Ventura

### Relevant hardware info

X86 and ARM

I have an issue with 7th Guest, large white/colored flickering splotches while moving.
They have the shape of the video content so fmv decoder related.
Don't show up on captures. Already tried surface and texture output, same.

Tested versions of Staging:
0.80.1 
0.81.0 alpha from today

Tested OS:
macOS Ventura M2
MS Windows 10 x86

Testend versions:
eXoDOS v5.0
Redump "7th Guest, The (Europe) (v1.2)"

IMG_2173

_No response_

### Your configuration

```ini
[sdl]
fullscreen          = false
display             = 0
fullresolution      = desktop
windowresolution    = default
viewport_resolution = fit
window_position     = auto
window_decorations  = true
transparency        = 0
host_rate           = auto
vsync               = true
vsync_skip          = 7000
presentation_mode   = auto
output              = opengl
texture_renderer    = auto
waitonerror         = true
priority            = auto auto
mute_when_inactive  = false
pause_when_inactive = false
mapperfile          = mapper-sdl2-0.81.0-alpha.map
screensaver         = auto

[dosbox]
language                    = 
machine                     = svga_s3
captures                    = capture
memsize                     = 32
mcb_fault_strategy          = deny
vmemsize                    = auto
dos_rate                    = default
vesa_modes                  = compatible
vga_8dot_font               = false
speed_mods                  = true
autoexec_section            = join
automount                   = true
startup_verbosity           = auto
allow_write_protected_files = true

[render]
aspect             = true
monochrome_palette = white
cga_colors         = default
glshader           = default

[composite]
composite   = auto
era         = auto
hue         = 0
saturation  = 100
contrast    = 100
brightness  = 0
convergence = 0

[cpu]
core      = auto
cputype   = auto
cycles    = 15000
cycleup   = 10
cycledown = 20

[mouse]
mouse_capture             = onclick
mouse_middle_release      = true
mouse_multi_display_aware = true
mouse_sensitivity         = 100
mouse_raw_input           = true
dos_mouse_driver          = true
dos_mouse_immediate       = false
ps2_mouse_model           = explorer
com_mouse_model           = wheel+msm

[mixer]
nosound    = false
rate       = 48000
blocksize  = 512
prebuffer  = 20
negotiate  = true
compressor = true
crossfeed  = off
reverb     = off
chorus     = off

[midi]
mididevice      = fluidsynth
midiconfig      = 
mpu401          = intelligent
raw_midi_output = false

[fluidsynth]
soundfont     = FluidR3_GM.sf2
fsynth_chorus = auto
fsynth_reverb = auto
fsynth_filter = off

[mt32]
model       = auto
romdir      = 
mt32_filter = off

[sblaster]
sbtype              = sb16
sbbase              = 220
irq                 = 7
dma                 = 1
hdma                = 5
sbmixer             = false
sbwarmup            = 100
oplmode             = auto
sb_filter           = modern
sb_filter_always_on = false
opl_filter          = auto
cms_filter          = on

[gus]
gus        = false
gusbase    = 240
gusirq     = 5
gusdma     = 3
ultradir   = C:\ULTRASND
gus_filter = off

[imfc]
imfc        = false
imfc_base   = 2a20
imfc_irq    = 3
imfc_filter = off

[innovation]
sidmodel          = none
sidclock          = default
sidport           = 280
6581filter        = 50
8580filter        = 50
innovation_filter = off

[speaker]
pcspeaker           = discrete
pcspeaker_filter    = on
tandy               = auto
tandy_filter        = on
tandy_dac_filter    = on
lpt_dac             = none
lpt_dac_filter      = on
ps1audio            = false
ps1audio_filter     = on
ps1audio_dac_filter = on

[reelmagic]
reelmagic       = off
reelmagic_key   = auto
reelmagic_fcode = 0

[joystick]
joysticktype                = auto
timed                       = true
autofire                    = false
swap34                      = false
buttonwrap                  = false
circularinput               = false
deadzone                    = 10
use_joy_calibration_hotkeys = false
joy_x_calibration           = auto
joy_y_calibration           = auto

[serial]
serial1       = dummy
serial2       = dummy
serial3       = disabled
serial4       = disabled
phonebookfile = phonebook.txt

[dos]
xms                   = true
ems                   = true
umb                   = true
ver                   = 5.0
country               = 0
expand_shell_variable = auto
keyboardlayout        = auto
setver_table_file     = 

[ipx]
ipx = false

[ethernet]
ne2000            = true
nicbase           = 300
nicirq            = 3
macaddr           = AC:DE:48:88:99:AA
tcp_port_forwards = 
udp_port_forwards = 

[autoexec]
@echo off
mount c /users/alex/games
PATH=Z:\;Y:\DOS;C:\DN;C:\DOS
c:
```

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
Previously the 8-bit VESA mode's line drawing routine wasn't doing DAC palette lookups.  This adds an equivalent linear lookup routine, but adds palette lookup.

Fixes #2486 

For VGA 640x480+ modes that 8-bit VESA is typically used in, rendering is still performed in bulk at the end of the frame instead of paced out per-line because so far we haven't seen games or demos perform DAC's palette effects mid-frame in 640x480+ modes.

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1303,26 +1303,49 @@ void VGA_CheckScanLength(void) {
 	}
 }
 
-void VGA_ActivateHardwareCursor(void) {
-	bool hwcursor_active=false;
-	if (svga.hardware_cursor_active) {
-		if (svga.hardware_cursor_active()) hwcursor_active=true;
+// If the hardware mouse cursor is activated, this function changes the VGA line
+// drawing function-pointers to call the more complicated hardware cusror
+// routines (for the given color depth).
+
+// If the hardware cursor isn't activated, the simply fallback to the normal
+// line-drawing routines for a given bit-depth.
+
+// Finally, return the current mode's bits per line buffer value.
+uint8_t VGA_ActivateHardwareCursor()
+{
+	uint8_t bit_per_line_pixel = 8;
+
+	auto should_draw_with_hw_mouse = [&]() {
+		return svga.hardware_cursor_active && svga.hardware_cursor_active();
+	};
+
+	switch (vga.mode) {
+	case M_LIN32: // True-colour VESA
+		VGA_DrawLine = should_draw_with_hw_mouse()
+		                     ? VGA_Draw_LIN32_Line_HWMouse
+		                     : VGA_Draw_Linear_Line;
+		//
+		// Use the "VGA_Draw_Linear_Line" routine that skips the DAC
+		// 8-bit palette LUT and prepares the true-colour pixels for
+		// rendering.
+		break;
+	case M_LIN16:
+	case M_LIN15: // High-colour VESA
+		VGA_DrawLine = should_draw_with_hw_mouse()
+		                     ? VGA_Draw_LIN16_Line_HWMouse
+		                     : VGA_Draw_Linear_Line;
+		//
+		// Use the "VGA_Draw_Linear_Line" routine that skips the DAC
+		// 8-bit palette LUT and prepares the high-colour pixels for
+		// rendering.
+		break;
+	default:
+		VGA_DrawLine = should_draw_with_hw_mouse()
+		                     ? VGA_Draw_VGA_Line_HWMouse
+		                     : VGA_Draw_Linear_Line;
+		break;
 	}
-	if (hwcursor_active) {
-		switch(vga.mode) {
-		case M_LIN32:
-			VGA_DrawLine=VGA_Draw_LIN32_Line_HWMouse;
-			break;
-		case M_LIN15:
-		case M_LIN16:
-			VGA_DrawLine=VGA_Draw_LIN16_Line_HWMouse;
-			break;
-		default:
-			VGA_DrawLine=VGA_Draw_VGA_Line_HWMouse;
-		}
-	} else {
-		VGA_DrawLine=VGA_Draw_Linear_Line;
-	}
+	return bit_per_line_pixel;
 }
 
 // Simple helper to check for SVGA+ resolutions
@@ -1797,7 +1820,7 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 			}
 		}
 		/* Use HW mouse cursor drawer if enabled */
-		VGA_ActivateHardwareCursor();
+		bpp = VGA_ActivateHardwareCursor();
 		break;
 	case M_LIN15:
  	case M_LIN16:
@@ -1810,7 +1833,7 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 		else
 			aspect_ratio /= 2.0;
 		/* Use HW mouse cursor drawer if enabled */
-		VGA_ActivateHardwareCursor();
+		bpp = VGA_ActivateHardwareCursor();
 		break;
 	case M_LIN4:
 		doublewidth = vga.seq.clocking_mode.is_pixel_doubling;
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -51,7 +51,8 @@ static_assert(SCALER_MAXWIDTH >= SCALER_MAX_MUL_WIDTH * max_scan_doubled_width);
 constexpr auto max_pixel_bytes = sizeof(uint32_t);
 constexpr auto max_line_bytes  = SCALER_MAXWIDTH * max_pixel_bytes;
 
-static std::array<uint8_t, max_line_bytes> templine_buffer;
+// The line buffer can be written in units up to RGB888 pixels (32-bit) size
+alignas(uint32_t) static std::array<uint8_t, max_line_bytes> templine_buffer;
 static auto TempLine = templine_buffer.data();
 
 static uint8_t * VGA_Draw_1BPP_Line(Bitu vidstart, Bitu line) {
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -371,23 +371,19 @@ static uint8_t* draw_unwrapped_line_from_dac_palette(Bitu vidstart, Bitu)
 
 	auto linear_pos = vidstart;
 
+	// Draw in batches of four to let the host pipeline deeper.
+	constexpr auto num_repeats = 4;
+	assert(pixels_in_line % num_repeats == 0);
+
 	// This function typically runs on 640+-wide lines and is a rendering
-	// bottleneck. Draw in batches of four to let the host pipeline deeper.
+	// bottleneck.
 	while (linear_pos < video_end) {
-		const auto masked_pos_0 = linear_pos++ & linear_mask;
-		const auto masked_pos_1 = linear_pos++ & linear_mask;
-		const auto masked_pos_2 = linear_pos++ & linear_mask;
-		const auto masked_pos_3 = linear_pos++ & linear_mask;
-
-		const auto palette_index_0 = *(linear_addr + masked_pos_0);
-		const auto palette_index_1 = *(linear_addr + masked_pos_1);
-		const auto palette_index_2 = *(linear_addr + masked_pos_2);
-		const auto palette_index_3 = *(linear_addr + masked_pos_3);
-
-		*line_addr++ = *(palette_map + palette_index_0);
-		*line_addr++ = *(palette_map + palette_index_1);
-		*line_addr++ = *(palette_map + palette_index_2);
-		*line_addr++ = *(palette_map + palette_index_3);
+		auto repeats = num_repeats;
+		while (repeats--) {
+			const auto masked_pos    = linear_pos++ & linear_mask;
+			const auto palette_index = *(linear_addr + masked_pos);
+			*line_addr++ = *(palette_map + palette_index);
+		}
 	}
 
 	return TempLine;
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1339,6 +1339,20 @@ uint8_t VGA_ActivateHardwareCursor()
 		// 8-bit palette LUT and prepares the high-colour pixels for
 		// rendering.
 		break;
+	case M_LIN8: // 8-bit VESA
+		if (should_draw_with_hw_mouse()) {
+			VGA_DrawLine = VGA_Draw_VGA_Line_HWMouse;
+		} else {
+			bit_per_line_pixel = 32;
+			VGA_DrawLine = draw_unwrapped_line_from_dac_palette;
+			//
+			// Use a routine that treats the 8-bit pixel values as
+			// indexes into the DAC's palette LUT. The palette LUT
+			// is populated with 32-bit RGB's (XRGB888) pre-scaled
+			// from 18-bit RGB666 values that have been written by
+			// the user-space software via DAC IO write registers.
+		}
+		break;
 	default:
 		VGA_DrawLine = should_draw_with_hw_mouse()
 		                     ? VGA_Draw_VGA_Line_HWMouse
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -352,6 +352,47 @@ static uint8_t * VGA_Draw_Linear_Line(Bitu vidstart, Bitu /*line*/) {
 	return ret;
 }
 
+static uint8_t* draw_unwrapped_line_from_dac_palette(Bitu vidstart, Bitu)
+{
+	// Quick references
+	static constexpr auto palette_map        = vga.dac.palette_map;
+	static constexpr uint8_t bytes_per_pixel = sizeof(palette_map[0]);
+	const auto linear_mask                   = vga.draw.linear_mask;
+	const auto linear_addr                   = vga.draw.linear_base;
+
+	// Video mode-specific line variables
+	const auto pixels_in_line = static_cast<uint16_t>(vga.draw.line_length /
+	                                                  bytes_per_pixel);
+	const auto video_end = vidstart + pixels_in_line;
+
+	// The line address is where the RGB888 palettized pixel is written.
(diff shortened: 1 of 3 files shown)


Grounded0:
Jane's ATF fixed in DOSBox Staging 8-bit SVGA LUT PR:

https://www.youtube.com/watch?v=ON58W2GMgC8

src/hardware/vga_draw.cpp
```diff
@@ -1265,26 +1307,44 @@ void VGA_CheckScanLength(void) {
...
+
+	auto should_draw_with_hw_mouse = [&]() {
+		return svga.hardware_cursor_active && svga.hardware_cursor_active();
+	};
+	switch (vga.mode) {
+	case M_LIN8:
+		if (should_draw_with_hw_mouse()) {
+			VGA_DrawLine = VGA_Draw_VGA_Line_HWMouse;
+		} else {
+			// Otherwise let 8-bit VESA modes draw from the DAC palette
```
johnnovak, line 1325: Maybe it's me, but I had to read this a few times and it's still ambiguous a bit. Ultimately the code made it clear what we're doing. "Drawing from the DAC palette"--we always do that, the question is how 😄 

Maybe say something like this: "Otherwise we render 8-bit VESA modes as 32-bit RGB (XRGB8888) instead of paletted." Then a further explanation what this fixes perhaphs.

johnnovak:
> For VGA 640x480+ modes that 8-bit VESA is typically used in, rendering is still performed in bulk at the end of the frame instead of paced out per-line because so far we haven't seen games or demos perform DAC's palette effects mid-frame in 640x480+ modes.

How does this fix the 7th Guest problem then? Do we have a clear understanding of why and how the flicker happens and how this fixes it? Specifically, is this a proper fix for a well-understood emulation problem, or a workaround that fixes one specific game, and maybe by fluke?

Also, how can we test the VESA hardware cursor stuff for regressions? Does Win3.x use that feature?

Grounded0:
We're basically emulating SVGA hardware more accurately now because we're actually using lookup table to pick colours in the 8-bit palette.

joncampbell123:
> > For VGA 640x480+ modes that 8-bit VESA is typically used in, rendering is still performed in bulk at the end of the frame instead of paced out per-line because so far we haven't seen games or demos perform DAC's palette effects mid-frame in 640x480+ modes.
> 
> How does this fix the 7th Guest problem then? Do we have a clear understanding of why and how the flicker happens and how this fixes it? Specifically, is this a proper fix for a well-understood emulation problem, or a workaround that fixes one specific game, and maybe by fluke?
> 
> Also, how can we test the VESA hardware cursor stuff for regressions? Does Win3.x use that feature?

Windows 3.1, 95, 98, and ME S3 drivers use the hardware cursor.

johnnovak:
> We're basically emulating S3 Trio64 hardware more accurately now because we're actually using a lookup table to pick colours in the 8-bit palette:

Huh? All VGA cards, or more broadly all devices with analog RGB output use some RAMDAC, usually with some LUT. That's not what the fix is about.

src/hardware/vga_draw.cpp
```diff
@@ -351,6 +352,47 @@ static uint8_t * VGA_Draw_Linear_Line(Bitu vidstart, Bitu /*line*/) {
...
+		const auto palette_index_0 = *(linear_addr + masked_pos_0);
+		const auto palette_index_1 = *(linear_addr + masked_pos_1);
+		const auto palette_index_2 = *(linear_addr + masked_pos_2);
+		const auto palette_index_3 = *(linear_addr + masked_pos_3);
+
+		*line_addr++ = *(palette_map + palette_index_0);
+		*line_addr++ = *(palette_map + palette_index_1);
+		*line_addr++ = *(palette_map + palette_index_2);
+		*line_addr++ = *(palette_map + palette_index_3);
+	}
```
johnnovak, line 391: No need to manually unroll; GCC is more than clever enough to do that if needed. Just do something like this:

```cpp
auto num_repeats = 4;
while (num_repeats--) {
   ...
}
```

johnnovak: Plus perhaps also assert that line length MOD 4 equals zero (mostly for documentation purposes).

src/hardware/vga_draw.cpp
kcgen: Added more comments in this switch statement to contrast how the high and true-colour modes are handled versus the 8-bit mode.

In @Grounded0 's diagram, it shows true and high-colour modes' buffers bypassing the 8-bit->18-bit RGB666 palette LUT, and instead their buffers go straight out to the 8/6 bit DAC for rendering.

Recall `machine = vgadraw`: for the standard EGA and VGA modes, it takes each 8-bit pixel value and looks up the 18-bit RGB666 palette LUT value (well, previously it was a 16-bit RGB565 LUT) on a per-line basis.

After we combined `vgadraw`, it turns out 8-bit VESA mode wasn't using the 18-bit palette LUT (instead, it was using `VGA_Draw_Linear_Line`, which is the same routine used by the true and high colour modes that bypass the 16-bit palette LUT). 

So this PR just adds case to handle do the 8-bit index -> 18-bit RGB666 palette LUT for 8-bit VESA:



And the route that the palette gets adjusted (Send Color is called from DAC's IO write ports functions):



> How does this fix the 7th Guest problem then

It uses 8-bit VESA and makes small adjustments to the DAC palette (not mid-frame, but either prior or after drawing), and only every now and then.  So the fix is to actually lookup values from there.

> Specifically, is this a proper fix for a well-understood emulation problem

Yes - it should handle any game that's using 8-bit VESA and making DAC palette tweaks.

> how can we test the VESA hardware cursor stuff for regressions?

That code path is the same: if hardware cursor is needed, it's using `VGA_Draw_VGA_Line_HWMouse` which is what it previous used. 

But this function /doesn't/ perform palette lookups. However, I'd rather wait for evidence of games needing this before figuring out how to do it.. because it's quite an involved function.

src/hardware/vga_draw.cpp
kcgen: The inner loop still unfortunately isn't being vectorized :sob: 

### GCC



### Clang



---

I included the while-loop in the last commit, however comparing it to the explicit version, I get about 2 to 3 fps more in QTG 640x480.

### While-loop




### Without

johnnovak: That's... disappointing. Stupid GCC 😢 Maybe leave the explicit unrolling then 🤷🏻

src/hardware/vga_draw.cpp
```diff
@@ -1343,6 +1343,20 @@ uint8_t VGA_ActivateHardwareCursor()
...
+			VGA_DrawLine = VGA_Draw_VGA_Line_HWMouse;
+		} else {
+			bit_per_line_pixel = 32;
+			VGA_DrawLine = draw_unwrapped_line_from_dac_palette;
+			//
+			// Use a routine that treats the 8-bit pixel values as
+			// indexes into the DAC's palette LUT. The palette LUT
+			// is populated with 32-bit RGB's (XRGB888) pre-scaled
+			// from 18-bit RGB666's writted by the user-space
+			// software via DAC IO write registers.
```
johnnovak, line 1353: Nice 👍🏻

src/hardware/vga_draw.cpp
kcgen: Yup - will do a comparison on the pi where it's the quietest environment and any improvement has the most benefit.

johnnovak: On a second thought... There's nothing to vectorise here, we're not doing any calculations, just 4 random table lookups. Of course, four memory reads from random memory locations cannot be vectorised!

Also, I _highly_ doubt the explicit unrolling vs the loop approach would make such a huge FPS difference (huge compared to how computationally infinitesimally, microscopically small incrementing a counter in a loop is...) Also, I doubt any unrolling would make any difference here at all. Have you measured that doing lookups 4 times here improves anything, or was it a guess?

kcgen: The `while` loop won all three rounds on the Pi:

| round | neither | unrolled | while |
|:---:|:---:|:---:|:---:|
1 | 6.4 | 6.8 | 7.0 
2 | 6.5 | 6.9 | 7.2 
3 | 6.6 | 6.8 | 7.1
 

Ran in round-robin all back-to-back. Planning to leave the PR as-is.

> Of course, four memory reads from random memory locations cannot be vectorised!

I think a scatter gather approach would work - the initial masked pos values are (usually) adjacent, as are the writes at the end (so those could maybe be SIMD'd) - with a scatter/gather in the middle.

But yeah.. pretty bandwidth-sensitive area .. these frame-rates are s slow, you can watch the pi struggle to grind through these lookups :sweat_smile:

joncampbell123:
> > We're basically emulating S3 Trio64 hardware more accurately now because we're actually using a lookup table to pick colours in the 8-bit palette:
> 
> Huh? All VGA cards, or more broadly all devices with analog RGB output use some RAMDAC, usually with some LUT. That's not what the fix is about.
> 
> If you're referring to the "high and true colour bypass" path, that's different. Here we're essentially "pretending" for 8-bit modes that they're 24-bit true colour (stored as 32-bit XRGB8888 in memory, but that's an implementation detail).
> 
> Anyway, this becomes a bit theoretical because we're not emulating any display adapter 1:1 accurately anyway... I'm just curious how this fixes the problem, and what the actual problem is smile
> 
> (An interesting curiosity: in earlier Amigas before they started using DACs, a resistor network was used to convert 12-bit RGB (4096 colours) into analog signals ([source](http://theamigamuseum.com/the-hardware/vidiot/)))

You'll find that the RAMDAC is still in effect in highcolor/truecolor modes. It still remaps values. The difference is that in highcolor/truecolor modes it remaps each channel independently (R/G/B). This is how SVGA cards from the late 1990s on are able to offer those nifty "gamma correction" settings in all supported desktop modes. As a weird example, the Radeon framebuffer driver in Linux used to have a strange quirk where it relied on that remapping to render the framebuffer console using the first two "steps" of the gamma ramp. A side effect of this is that if you were writing code to try to draw images on the framebuffer, they came out all messed up because of that mapping, unless you were willing to fiddle with the colormap ioctls() to program a linear ramp, at the cost of making your framebuffer console text very dim.

src/hardware/vga_draw.cpp
johnnovak: Nice measurements @kcgen , the `while` loop variant it is then 🚀 Thanks for testing this!

johnnovak:
> You'll find that the RAMDAC is still in effect in highcolor/truecolor modes. It still remaps values.

Oh yeah, that's pretty much a given, regardless of whether we're dealing with paletted or high/true-colour; _something_ must still generate analog RGB signals from the digital values. There's just one extra lookup in the case of the paletted modes. There must have been such RAMDACs on all modern boards too until they completely dropped the VGA output; now it's all digital.

Related to the gamma lookup thing you mentioned, S3 cards also have this "black pedestal" bit; I guess they switch between output signal ranges similar to the full/limited HDMI ranges. Some S3 cards default to "black pedestal = on", which leads to washed out blacks on regular VGA monitors. None of my S3 cards do that, but I found some simple utils on Vogons that can switch the pedestal bit on and off. I can turn it _on_ on some of my S3 cards, then I get washed out blacks 😎 

I guess you can only do trickery with analog circuitry to convert from digital values to analog RGB purely with analog components like in the case of the Amiga if you need very limited precision (e.g. 4-bit per RGB component in the case of the Amiga).

Grounded0:
@johnnovak can you sign off this PR already. That's what i'm mostly interested about. 😁

johnnovak (approved):
Great work @kcgen, merge away 🚀

joncampbell123:
> > You'll find that the RAMDAC is still in effect in highcolor/truecolor modes. It still remaps values.
> 
> Oh yeah, that's pretty much a given, regardless of whether we're dealing with paletted or high/true-colour; _something_ must still generate analog RGB signals from the digital values. There's just one extra lookup in the case of the paletted modes. There must have been such RAMDACs on all modern boards too until they completely dropped the VGA output; now it's all digital.
> 
> Related to the gamma lookup thing you mentioned, S3 cards also have this "black pedestal" bit; I guess they switch between output signal ranges similar to the full/limited HDMI ranges. Some S3 cards default to "black pedestal = on", which leads to washed out blacks on regular VGA monitors. None of my S3 cards do that, but I found some simple utils on Vogons that can switch the pedestal bit on and off. I can turn it _on_ on some of my S3 cards, then I get washed out blacks sunglasses
> 
> I guess you can only do trickery with analog circuitry to convert from digital values to analog RGB purely with analog components like in the case of the Amiga if you need very limited precision (e.g. 4-bit per RGB component in the case of the Amiga).

As far as I am aware, the "palette as gamma ramp" behavior remains standard even on chipsets that can do both VGA, DVI, and HDMI even well into the 2010s. My current laptop I purchased in 2019 with an Intel CPU (laptop display and HDMI port) clearly *still* has that feature because I can use the "xgamma" command on my Linux desktop to play with it.

joncampbell123:
The funny thing about the Radeon case is that I ran into it long ago when I installed Yellow Dog Linux on an old PowerPC Apple PowerBook (you can dual boot Mac OS X and Linux on those!). The ones in the early to mid 2000s used ATI Radeon chipsets. We used to do video production in the early 2010s and we used it all the time to capture MiniDV tapes in Linux through all those Firewire ports Apple laptops used to have back then. It was an odd problem when I wrote my own code to draw on the framebuffer, though obviously programs like X-Windows and MPlayer were aware of it and did the colormap controls to compensate when you ran them.

PR #2101, opened 2022-12-05
Use std::vector to manage most of the emulated memory blocks

kcgen:
Also fixes some latent presence-check bugs in the imageDiskList code and drops a unused object from a memory structure.

PR #2496, opened 2023-05-18
Further debloat of paging arrays

kklobe:
Following on #2101, @kcgen was there a performance or compatibility reason `read` and `write` weren't `std::vector`-ized?

### DIFF
--- a/include/paging.h
+++ b/include/paging.h
@@ -162,8 +162,8 @@ struct PagingBlock {
 	} base = {};
 #if defined(USE_FULL_TLB)
 	struct {
-		HostPt read[TLB_SIZE]  = {};
-		HostPt write[TLB_SIZE] = {};
+		std::vector<HostPt> read = std::vector<HostPt>(TLB_SIZE);
+		std::vector<HostPt> write = std::vector<HostPt>(TLB_SIZE);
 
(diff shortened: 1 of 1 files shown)


kcgen:
The only reason I can remember for why some arrays weren't converted was due to that ARM dynrec core issue that in Elite (2?) where the intro loops.  Although maybe that happens regardless?

kklobe:
OK - let me also test this on my Windows box, since `tlb.read` and `tlb.write` are accessed in `core_dyn_x86\decoder.h`

kklobe:
completely crashes on x64, back to the drawing board for now.

kcgen:
Darn. Yeah; GDB doesn't have much visibility into the dynamic core. This is as much as I get:

```
#9  0x00000ffffe6b2280 in  ()
#10 0x0000000000002d00 in  ()
#11 0x00007fffffff8c70 in  ()
#12 0x0000555559429023 in sync_normal_fpu_and_run_dyn_code(uint8_t const*) (code=0x0)
    at ../../src/cpu/core_dyn_x86.cpp:288
```

The only difference I see is order of operations:
 - the literal arrays are backed in and would be available to even the earliest running static and global code.
 - the `std::vector`s content will only exist after it's constructed at some  part of the startup sequence.

PR #2498, opened 2023-05-19
Fix the scan code sequence for the `pause` key

kcgen:
@FeralChild64 , see the first commit which has the fix.

The second commit updates the other scan code sequences to match the same style.

### DIFF
--- a/src/hardware/input/keyboard_scancodes.cpp
+++ b/src/hardware/input/keyboard_scancodes.cpp
@@ -248,23 +248,26 @@ std::vector<uint8_t> KEYBOARD_GetScanCode1(const KBD_KEYS key_type,
 		return scan_code;
 
 	case KBD_printscreen:
-		scan_code.resize(4);
-		scan_code[0] = 0xe0;
-		scan_code[1] = 0x2a | (is_pressed ? 0 : 0x80);
-		scan_code[2] = 0xe0;
-		scan_code[3] = 0x37 | (is_pressed ? 0 : 0x80);
+		scan_code = {
+		        0xe0,
+		        static_cast<uint8_t>(0x2a | (is_pressed ? 0 : 0x80)),
+		        0xe0,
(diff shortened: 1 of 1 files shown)


src/hardware/input/keyboard_scancodes.cpp
```diff
@@ -236,34 +236,38 @@ std::vector<uint8_t> KEYBOARD_GetScanCode1(const KBD_KEYS key_type,
...
-			scan_code[1] = 0x1d | 0x80;
-			scan_code[2] = 0x45 | 0x80;
+			scan_code = {
+			        0xe1,
+			        0x1d,
+			        0x45,
+			        0xe1,
+			        0x1d | 0x80,
+			        0x45 | 0x80,
+			};
```
kcgen, line 246: (copy & paste typo in the original that overwrites indexes 0, 1, and 2)

src/hardware/input/keyboard_scancodes.cpp
```diff
@@ -444,27 +448,29 @@ std::vector<uint8_t> KEYBOARD_GetScanCode2(const KBD_KEYS key_type,
...
 		if (!is_pressed) {
 			// Print Screen gets reported only when released
-			scan_code.resize(6);
-			scan_code[0] = 0xe0;
-			scan_code[1] = 0xf0;
-			scan_code[2] = 0x7c;
-			scan_code[3] = 0xe0;
-			scan_code[4] = 0xf0;
-			scan_code[5] = 0x12;
+			scan_code = {
```
FeralChild64, line 451: How about getting rid of the local `scan_code` variable from all 3 `KEYBOARD_GetScanCode` routines and just doing something like
```c
return { 0xe0, 0xf0, 0x7c, 0xe0, 0xf0, 0x12 };
```
?

FeralChild64 (approved):

PR #2506, opened 2023-05-20
Update Windows 32-bit checksum

johnnovak:


kcgen:
Thanks for catching this!

PR #2508, opened 2023-05-21
Implement DOS mouse driver function 0x32

FeralChild64:
- implement one more informational function of the DOS mouse driver - might help applications to determine which functions our driver implements
- manually clean-up some indentation

### DIFF
--- a/src/hardware/input/mouseif_dos_driver.cpp
+++ b/src/hardware/input/mouseif_dos_driver.cpp
@@ -1356,8 +1356,8 @@ static Bitu int33_handler()
 		                   static_cast<float>(state.maxpos_x));
 		// Or alternatively this:
 		// pos_x = (state.maxpos_x - state.minpos_x + 1) / 2;
-		LOG(LOG_MOUSE, LOG_NORMAL)
-		("Define Hortizontal range min:%d max:%d", state.minpos_x, state.maxpos_x);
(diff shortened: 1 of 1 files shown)


johnnovak (approved):
Looks okay, although I have no idea what this is doing 😄 Did you test it with something that uses this?

kcgen (approved):
Looks good, @FeralChild64 - and nice logging refinements.

FeralChild64:
> Looks okay, although I have no idea what this is doing smile Did you test it with something that uses this?

It returns a bitfield telling which functions the mouse driver supports. Unfortunately, I haven't found anything that uses it - but the implementation itself is rather trivial.
In fact, some software SHOULD call this function, but it doesn't - for example, MSD.EXE calls function 0x34 to get the MOUSE.INI file path information, not checking first whether the driver supports function 0x34 or not :/

johnnovak:
> > Looks okay, although I have no idea what this is doing smile Did you test it with something that uses this?
> 
> It returns a bitfield telling which functions the mouse driver supports. Unfortunately, I haven't found anything that uses it - but the implementation itself is rather trivial. In fact, some software SHOULD call this function, but it doesn't - for example, MSD.EXE calls function 0x34 to get the MOUSE.INI file path information, not checking first whether the driver supports function 0x34 or not :/

Thanks for the explanation @FeralChild64.

PR #2522, opened 2023-05-23
Make the RWQueue more usable for complex types

kcgen:
To support the upcoming overhauled capturing PR, this adds to `RWQueue`:
 - `Stop()`, as a way to stop and the queue from accepting more items and freeing the need for these controls on the user-side.
 - `IsRunning()`, to check if the queue is still accepting new items.
 - `IsRunning() || !IsEmpty()`, can be used tell when no remaining items are left in the queue
 - `<optional>` return value from `Dequeue()`, to allow the using-side to test if the value is available.

### DIFF
--- a/include/rwqueue.h
+++ b/include/rwqueue.h
@@ -40,6 +40,7 @@
 #include <condition_variable>
 #include <deque>
 #include <mutex>
+#include <optional>
 #include <vector>
 
 template <typename T>
@@ -96,8 +97,9 @@ public:
 	// in the queue to dequeue.
 
 	// If queuing has stopped, this will continue to return item(s) until
-	// none remain in the queue, at which point it immediately returns T{}.
-	T Dequeue();
+	// none remain in the queue, at which point it returns empty results
+	// as indicated by the <optional> wrapper evaluating as "false".
+	std::optional<T> Dequeue();
 
 	// Bulk operations move multiple items from/to the given vector, which
 	// signficantly reduces the number of mutex lock state changes. It also
--- a/include/rwqueue.h
+++ b/include/rwqueue.h
@@ -25,20 +25,18 @@
 
 /*  RW (Read/Write) Queue
  *  ---------------------
- *  A fixed-size thread-safe queue that blocks both the
- *  producer until space is available and the consumer until
- *  items are available.
+ *  A fixed-size thread-safe queue that blocks both the producer until space is
+ *  available, and the consumer until items are available.
  *
- *  For some background, this class was authored to replace
- *  the one that resulted from this discussion:
- *  https://github.com/cameron314/readerwriterqueue/issues/112
- *  because the MoodyCamel implementation:
- *   - Is roughly 5-fold larger (and more latent)
- *   - Consumes more CPU by spinning (instead of locking)
- *   - Lacks bulk queue/dequeue methods (request was rejected
- *     https://github.com/cameron314/readerwriterqueue/issues/130)
+ *  For optimal performance inside the rwqueue, blocking is accomplished by
+ *  putting the thread into the waiting state, then waking it up via notify when
+ *  the required conditions are met.
+ *
+ *  Producer and consumer thread(s) are expected to simply call the enqueue and
+ *  dequeue methods directly without any thread state management.
  */
 
+#include <atomic>
 #include <condition_variable>
 #include <deque>
 #include <mutex>
@@ -52,6 +50,7 @@ private:
 	std::condition_variable has_room  = {};
 	std::condition_variable has_items = {};
 	size_t capacity                   = 0;
+	std::atomic<bool> is_running      = true;
 	using difference_t = typename std::vector<T>::difference_type;
 
 public:
@@ -63,8 +62,20 @@ public:
 	void Resize(size_t queue_capacity);
 
 	bool IsEmpty();
+
+	// non-blocking call
+	bool IsRunning() const;
+
+	// non-blocking call
 	size_t Size();
+
+	// non-blocking calls
+	void Stop();
+
+	// non-blocking call
 	size_t MaxCapacity() const;
+
+	// non-blocking call
 	float GetPercentFull();
 
 	// Discourage copying into the queue. Instead, use std::move into the
@@ -73,9 +84,19 @@ public:
 	//
 	// void Enqueue(const T& item);
 
-	// items will be empty (moved-out) after call
-	void Enqueue(T&& item);
+	// Items will be empty (moved-out) after call.
+	// The method potentially blocks until the queue has enough capacity to
+	// queue a single item.
 
+	// If queuing has stopped prior to enqueing, then this will immediately
+	//  return false and the item will not be queued.
+	bool Enqueue(T&& item);
+
+	// The method potentially blocks until there is at least a single item
+	// in the queue to dequeue.
+
+	// If queuing has stopped, this will continue to return item(s) until
+	// none remain in the queue, at which point it immediately returns T{}.
 	T Dequeue();
 
 	// Bulk operations move multiple items from/to the given vector, which
@@ -90,11 +111,28 @@ public:
 
 	// Items are std::move'd out of the source vector into the queue. This
 	// function clear()s the vector such that it's in a defined state on
-	// return (and can be reused).
-	void BulkEnqueue(std::vector<T>& from_source, const size_t num_requested);
+	// return (and can be reused). The method potentially blocks until there
+	// is enough capacity in the queue for the new items.
 
-	// The target vector will be resized to accomodate, if needed.
-	void BulkDequeue(std::vector<T>& into_target, const size_t num_requested);
+	// If queuing has stopped prior to bulk enqueing, then this will
+	// immediately return false and no items will be queued.
+
+	// If queuing stops in the middle of enqueing prior to completion, then
+	// this will immediately return false. The items queued /before/
+	// stopping will be available in the queue however the items that came
+	// after stopping will not be queued.
+	bool BulkEnqueue(std::vector<T>& from_source, const size_t num_requested);
+
+	// The target vector will be resized to accomodate, if needed. The
+	// method potentially blocks until the requested number of items have
+	// been dequeued.
+
+	// If queuing has stopped:
+	//  - Returns true when  one or more item(s) have been dequeued.
+	//  - Returns false when no items can be dequeued.
+	//
+	// The vector is always sized to match the number of items returned.
+	bool BulkDequeue(std::vector<T>& into_target, const size_t num_requested);
 };
 
 #endif
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -544,9 +544,8 @@ void MidiHandlerFluidsynth::Close()
 
 	// Stop rendering and drain the fifo
 	keep_rendering = false;
-	while (audio_frame_fifo.Size()) {
-		(void)audio_frame_fifo.Dequeue();
-	}
+	work_fifo.Stop();
+	audio_frame_fifo.Stop();
 
 	// Wait for the rendering thread to finish
 	if (renderer.joinable())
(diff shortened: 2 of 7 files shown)


include/rwqueue.h
```diff
@@ -90,11 +114,28 @@ class RWQueue {
...
+	// after stopping will not be queued.
+	bool BulkEnqueue(std::vector<T>& from_source, const size_t num_requested);
+
+	// The target vector will be resized to accomodate, if needed. The
+	// method potentially blocks until the requested number of items have
+	// been dequeued.
+
+	// If queuing has stopped:
+	//  - Returns "true" if item(s) have been dequeued.
+	//  - Returns "false" when no items can be dequeued.
```
johnnovak, line 137: minor: no need for double quotes

include/rwqueue.h
```diff
@@ -90,11 +114,28 @@ class RWQueue {
...
+	//
+	// The vector is always sized to match the number of items returned.
+	bool BulkDequeue(std::vector<T>& into_target, const size_t num_requested);
```
johnnovak, line 140: Very nice API documentation 🎉

src/misc/rwqueue.cpp
```diff
@@ -65,16 +78,40 @@ bool RWQueue<T>::IsEmpty()
...
+template <typename T>
+bool RWQueue<T>::StoppedDequeuing()
+{
+	if (is_stopped) {
+		// We're only done when there's nothing left
+		std::lock_guard<std::mutex> lock(mutex);
+		return queue.empty();
+	}
+	return false;
+}
```
johnnovak, line 95: Isn't this API a little bit redundant? We already have `IsEmpty`, would only adding `IsStopped` suffice instead of `StoppedEnqueuing` and `StoppedDequeuing`? Especially `StoppedDequeuing` combined multiple checks into one call.

src/misc/rwqueue.cpp
```diff
@@ -65,16 +78,40 @@ bool RWQueue<T>::IsEmpty()
...
+
+template <typename T>
+bool RWQueue<T>::Enqueue(T&& item)
 {
-	// wait until the queue has room to accept the item
+	// wait until we're stopped or the queue has room to accept the item
 	std::unique_lock<std::mutex> lock(mutex);
-	has_room.wait(lock, [this] { return queue.size() < capacity; });
+	has_room.wait(lock,
+	              [this] { return is_stopped || queue.size() < capacity; });
```
johnnovak, line 103: Nice trick to emulate multiple notify conditions 👍🏻

src/misc/rwqueue.cpp
```diff
@@ -105,42 +142,58 @@ void RWQueue<T>::BulkEnqueue(std::vector<T>& from_source, const size_t num_reque
...
+			queue.insert(queue.end(),
+			             std::move_iterator(source_start),
+			             std::move_iterator(source_end));
+			source_start = source_end;
+			num_remaining -= num_items;
+		} else {
+			// If we stopped while bulk enqueing, then stop here.
+			// Anything that was enqueued prior to being stopped is
+			// safely in the queue.
+			num_remaining = 0;
```
johnnovak, line 163: I think that's a sensible choice. It should be the client's responsibility to defer the queue being stopped if it wants to ensure some (bulk)enqueue operations are fully processed. This makes for a more flexible API too.

src/misc/rwqueue.cpp
```diff
@@ -105,42 +142,58 @@ void RWQueue<T>::BulkEnqueue(std::vector<T>& from_source, const size_t num_reque
...
-
-	// get it, and notify the first waiting thread that the queue has room
-	T item = std::move(queue.front());
-	queue.pop_front();
+	has_items.wait(lock, [this] { return is_stopped || !queue.empty(); });
+
+	T item = {};
+	// Even if the queue has stopped, we need to drain the (previously)
+	// queued items before we're done.
+	if (!is_stopped || !queue.empty()) {
```
johnnovak, line 184: Subjectively, I'd use `is_running` because "not running" (=stopped) reads a lot better to me than "not stopped" (=running). The latter is almost like a double negative to me, and double negatives always give me a pause 😄 It's a very minor thing, though.

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
 
+TEST(RWQueue, StopImmediately)
+{
+	RWQueue<int> q(65);
+
+	q.Stop();
+	EXPECT_TRUE(q.StoppedEnqueuing());
+	EXPECT_TRUE(q.StoppedDequeuing());
+
+	q.Enqueue(1);                   // shouldn't block
```
johnnovak: Expect false retval?

johnnovak: Expect IsEmpty to be true?

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+	const auto value = q.Dequeue(); // shouldn't block
+	EXPECT_EQ(value, 0);            // once stopped, dequeues defaults
+}
+
+TEST(RWQueue, StopMidway)
+{
+	RWQueue<int> q(2);
+
+	q.Enqueue(1);
```
johnnovak, line 457: Expect Size() == 1
Expect IsEmpty() == false

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+
+	EXPECT_TRUE(q.StoppedEnqueuing());
+	EXPECT_TRUE(q.StoppedDequeuing());
+
+	std::vector<int> items = {1, 2, 3};
+	const auto num_items   = items.size();
+
+	// Bulk enqueuing fails after being stopped
+	const auto bulk_enqueue_result = q.BulkEnqueue(items, num_items);
+	EXPECT_FALSE(bulk_enqueue_result);
```
johnnovak, line 492: Expect Size() == 3
Expect IsEmpty() == false

johnnovak, line 492: Expect q.IsEmpty() == false

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+TEST(RWQueue, StopBulkMidway)
+{
+	RWQueue<int> q(8);
+
+	// Bulk enque a couple before stopping
+	std::vector<int> items   = {1, 2, 3, 4, 5};
+	auto bulk_enqueue_result = q.BulkEnqueue(items, items.size());
+	EXPECT_TRUE(bulk_enqueue_result);
+	EXPECT_FALSE(q.StoppedEnqueuing());
+	EXPECT_FALSE(q.StoppedDequeuing());
```
johnnovak, line 509: Expect q.Size() == 5

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+
+	q.Stop();
+
+	// Bulking enqueuing fails after being stopped
+	items               = {6, 7};
+	bulk_enqueue_result = q.BulkEnqueue(items, items.size());
+	EXPECT_FALSE(bulk_enqueue_result);
+	EXPECT_TRUE(q.StoppedEnqueuing());
```
johnnovak, line 517: Expect q.Size() == 5

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+
+	// But we still have a handful of items queued
+	EXPECT_FALSE(q.StoppedDequeuing());
+
+	// Bulk dequeue the first couple
+	auto num_items           = 2u;
+	auto bulk_dequeue_result = q.BulkDequeue(items, num_items);
+	EXPECT_TRUE(bulk_dequeue_result);
+	std::vector<int> expected_items = {1, 2};
+	EXPECT_EQ(items, expected_items);
```
johnnovak, line 527: Expect q.Size() == 3

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+
+	// Dequeue the middle value
+	auto value = q.Dequeue();
+	EXPECT_EQ(value, 3);
```
johnnovak: Expect q.Size() == 2

tests/rwqueue_tests.cpp
```diff
@@ -439,4 +439,118 @@ TEST(RWQueue,ContainerMoveAsync)
...
+
+	// Bulk dequeue the last couple, but over-request
+	num_items           = 3u;
+	bulk_dequeue_result = q.BulkDequeue(items, num_items);
+	EXPECT_TRUE(bulk_dequeue_result);
+	expected_items = {4, 5};
+	EXPECT_EQ(items, expected_items);
```
johnnovak, line 538: Expect q.Size() == 0
Expect q.IsEmpty()

src/midi/midi_fluidsynth.cpp
```diff
@@ -682,33 +680,36 @@ void MidiHandlerFluidsynth::RenderAudioFramesToFifo(const uint16_t num_audio_fra
...
+		assert(work->message_type == MessageType::SysEx);
+		ApplySysexMessage(work->message);
 	}
 }
 
 // Keep the fifo populated with freshly rendered buffers
 void MidiHandlerFluidsynth::Render()
 {
-	while (keep_rendering.load()) {
+	while (!work_fifo.StoppedEnqueuing()) {
```
johnnovak, line 712: Yeah, so (to me) `while (work_fifo.IsRunning())` would be perhaps a bit more natural than the "double negation".

src/misc/rwqueue.cpp
```diff
@@ -172,24 +172,24 @@ bool RWQueue<T>::BulkEnqueue(std::vector<T>& from_source, const size_t num_reque
 }
 
 template <typename T>
-T RWQueue<T>::Dequeue()
+std::optional<T> RWQueue<T>::Dequeue()
```
johnnovak, line 175: 💯 much better than using the default value!

johnnovak (approved):
Very nice addition @kcgen 🎉 Just a couple of relatively minor suggestions

Nice to see you've added tests too; next level would be a threaded multi-consumer test, but those are quite hard to write in a non-flaky way... This is good enough for now, I'd say 😄

include/rwqueue.h
kcgen: +1

include/rwqueue.h
kcgen: :+1

src/midi/midi_fluidsynth.cpp
kcgen: definitely agree; will do.

kcgen: The queue represents a form of defered work.
 - Stopping it immediate prevents new work from entering.
 - But a choice needs (or can) be made about the current work in the queue.
 - Do you ignore that work or complete it?

In the case of audio samples - we don't care and just throw the queue away.. the audio stream just comes to a close sooner.

When it comes to important stuff in the queue, it makes sense to finish that up.

So that's why we've got two indicators.

Thinking we can flip them to the positive.
 - IsEnqueuing() -> when false, queuing is futile and discarded.
 - IsDequeuing() -> when false, the queue has run dry /and/ there won't be more.
 -

kcgen: A general option might be more readable- IsRunning(), could be IsEnqueuing() || IsDequeuing() under the hood.

johnnovak: > * IsEnqueuing() -> when false, queuing is futile and discarded.
> 
> * IsDequeuing() -> when false, the queue has run dry /and/ there won't be more.

Yeah I like that. Better than inferring it from IsRunning() and IsEmpty(), on a second thought.

src/misc/rwqueue.cpp
kcgen: `is_running` -> switched; thanks @johnnovak :+1:

tests/rwqueue_tests.cpp
kcgen: added

src/misc/rwqueue.cpp
kcgen: Knocked out `IsDequeuing()` in favour of the letting the user combine calls. Now down to `Stop()` and `IsRunning()`.

kcgen:
Ok, think it's ready. Test both commits, they build and run (MT32 and MIDI).

johnnovak:
> Ok, think it's ready. Test both commits, they build and run (MT32 and MIDI).

Nice one! 🚀

PR #2528, opened 2023-05-24
Resize VGA within half the current frame-time (#2510)

### LINKED ISSUE #2510: Earthworm Jim graphical errors on left screen border

DEV

Mac OS

_No response_

### Relevant hardware info

_No response_

At the start of the first level, when you move the character there are parts of the backround that flicker on the left screen border.

Error does happen using v0.80.1 and latest dev build.
Error also happens in vanilla DOSBox v0.74-3.

Couldn't reproduce the error using DOSBox-X (May release). 

Explain how to reproduce
1. Start Earthworm Jim (i used the eXoDOS version) after mounting the CD image.
2. Select option '1' for mixed resolutions (320x224 and 320x240 (recommended)).
3. You can skip the intro by pressing ESC once.
4. Start the game which brings you to the first level.

-The error doesn't always happen, try starting only the game again from the DOS CLI is enough to reproduce the problem.
-Performance also takes a huge hit when the error happens.

Log output about resolution switching with latest main:
Bildschirmfoto 2023-05-22 um 11 02 58

Video capture of the problem:
[ewj1_000.avi.zip](https://github.com/dosbox-staging/dosbox-staging/files/11529729/ewj1_000.avi.zip)

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
Fixes #2510

### DIFF
--- a/include/vga.h
+++ b/include/vga.h
@@ -545,7 +545,10 @@ void VGA_SetMonoPalette(const char *colour);
 void VGA_SetMode(VGAModes mode);
 void VGA_DetermineMode(void);
 void VGA_SetupHandlers(void);
-void VGA_StartResize(Bitu delay=50);
+
+void VGA_StartResize();
+void VGA_StartResizeAfter(const uint16_t delay_ms);
+
 void VGA_SetupDrawing(uint32_t val);
 void VGA_CheckScanLength(void);
 void VGA_ChangedBank(void);
--- a/src/hardware/vga.cpp
+++ b/src/hardware/vga.cpp
@@ -139,7 +139,7 @@ void VGA_SetModeNow(VGAModes mode) {
 	if (vga.mode == mode) return;
 	vga.mode=mode;
 	VGA_SetupHandlers();
-	VGA_StartResize(0);
+	VGA_StartResizeAfter(0);
 }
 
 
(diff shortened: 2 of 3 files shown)


Burrito78 (approved):
The issue is fixed using this build:

https://github.com/dosbox-staging/dosbox-staging/actions/runs/5072287747

Thanks @kcgen , awesome work as always!

kcgen:
Prehistorik 2's skill-selection scroller (might?) be improved, too -- although it seems pretty messy on real hardware.

### Hardware (lower drawn regions 'phases out' during vertical sweep)

https://www.youtube.com/watch?v=839fGMjX9qc

### Baseline

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/7b0e847f-7b87-40e8-a9fb-c7d44b9929f1

### PR

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/c2ebd546-a9a5-4fc5-8e63-9727c3c56ce0

johnnovak (approved):
Interesting 😄 You got this from DOSBox-X @kcgen ?

johnnovak:
> Prehistorik 2's skill-selection scroller (might?) be improved, too -- although it seems pretty messy on real hardware.

That's horrible, it's not supposed to look like that at all... The guy must have used some late era VGA card with bad backwards compatibility.

The text is supposed to be completely stationary, and the background image should be panning 100% smoothly. It actually used to be like that with DOSBox about a year ago; it was perfect.

It has some artifacts on my S3 (late VGA card) but nowhere near the horrible mess like on that youtube video.

Wonder if using tseng3000/4000 fixes anything.

kcgen:
> Interesting :smile: You got this from DOSBox-X @kcgen ?

Not this one; I could change the "occurrence" rate by change the `dos_rate` so then started looking at other timings. Unfortunately it's a "stochastic" fix, in that I don't have a technical reference - and only using game correctness as a guide. So who knows.. we're probably in for another round of regressions :sweat_smile:

johnnovak:
> Unfortunately it's a "stochastic" fix, in that I don't have a technical reference - and only using game correctness as a guide. So who knows.. we're probably in for another round of regressions 😅

Uhhuh... Well, the thing is, if games relied on such timings, that was a recipe for disaster. And that can explain why on some later VGA cards you got glitches. Like many demos requiring some specific brand/type VGA cards to work 100% correctly.

I'm wondering if it's worth messing around with stochastic fixes for a non-breaking glitch for a _single_ game if it can cause regressions and it's not mimicing measured or documented hardware behaviour (too risky).

kcgen:
I think it's worth it - these two games were early 90s and normally look perfect on good hardware -- and both actually get better by tightening up this timing.

DOSBox's default was to lag by 50ms of tick time from when the request came in before it takes effect and to set the mode (and pushes the signal out to the monitor).  

50 ms is over 3 frames of lag where the prior mode is still ticking along.. meanwhile, most games have prepared themselves for the new mode at the time of the request.  

The fact that both these games' rendering glitches improve when we tighten this up says we're "pulling" it in a more correct direction. 

(I've also regression tested the suite 'oddball video mode' games and programs, and all are still OK)

johnnovak:
> 



> I think it's worth it - these two games were early 90s and normally look perfect on good hardware -- and both actually get better by tightening up this timing.
> 
> DOSBox's default was to lag by 50ms of tick time from when the request came in before it takes effect and to set the mode (and pushes the signal out to the monitor).
> 
> 50 ms is over 3 frames of lag where the prior mode is still ticking along.. meanwhile, most games have prepared themselves for the new mode at the time of the request.
> 
> The fact that both these games' rendering glitches improve when we tighten this up says we're "pulling" it in a more correct direction.
> 
> (I've also regression tested the suite 'oddball video mode' games and programs, and all are still OK)

Yeah, we can give it a go, it's easily bisectable 😄 Actually, 99% of games just set a standard mode via BIOS anyway and then forget about it, which narrows down the regression testing to our "special" set of games that use weird custome resolutions, etc. This should make no difference to most programs, which is good.

kcgen:
Ok, let's give it a go! Thanks @Burrito78 for the testing, and @Grounded0  and @johnnovak for all the technical help :rocket:

PR #2529, opened 2023-05-24
Prefer evdev mouse API over xinput on Linux

FeralChild64:
Changed ManyMouse default API on Linux to `evdev`. It is a more modern approach, the `xinput` is more like legacy today - moreover, in my test `evdev` driver is better at identifying which logical USB device is actually a mouse and which, is not -  for example in case of keyboard advertising it's configuration interface as a device.


kcgen:
Nicely documented, @FeralChild64.

PR #2534, opened 2023-05-26
Track and free allocations in MEM program

### LINKED ISSUE #2532: Internal command Z:\MEM.COM eats away all the UMB memory

0.80.1, also reproducible on current GIT

Linux x86_64

_No response_

### Relevant hardware info

_No response_

It seems the internal command `mem` eats away all the UMB memory. I have used the `mem` command from DR-DOS to confirm that it is not a display bug, but indeed UMB is not available any more.

1. Start DOSBox Staging with option `-noprimaryconf`
2. Execute `mem` command twice
3. The second time it reports no UMB memory

bug

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
Fixes #2532

### DIFF
--- a/src/dos/program_mem.cpp
+++ b/src/dos/program_mem.cpp
@@ -25,74 +25,92 @@
 #include "program_more_output.h"
 #include "regs.h"
 
-void MEM::Run(void) {
+void MEM::Run(void)
+{
 	if (HelpRequested()) {
 		MoreOutputStrings output(*this);
 		output.AddString(MSG_Get("PROGRAM_MEM_HELP_LONG"));
 		output.Display();
 		return;
 	}
-    /* Show conventional Memory */
-    WriteOut("\n");
+	/* Show conventional Memory */
+	WriteOut("\n");
 
-    uint16_t umb_start=dos_infoblock.GetStartOfUMBChain();
-    uint8_t umb_flag=dos_infoblock.GetUMBChainState();
-    uint8_t old_memstrat=DOS_GetMemAllocStrategy()&0xff;
-    if (umb_start!=0xffff) {
-        if ((umb_flag&1)==1) DOS_LinkUMBsToMemChain(0);
-        DOS_SetMemAllocStrategy(0);
-    }
+	uint16_t umb_start   = dos_infoblock.GetStartOfUMBChain();
+	uint8_t umb_flag     = dos_infoblock.GetUMBChainState();
+	uint8_t old_memstrat = DOS_GetMemAllocStrategy() & 0xff;
+	if (umb_start != 0xffff) {
+		if ((umb_flag & 1) == 1) {
+			DOS_LinkUMBsToMemChain(0);
+		}
+		DOS_SetMemAllocStrategy(0);
+	}
 
-    uint16_t seg,blocks;blocks=0xffff;
-    DOS_AllocateMemory(&seg,&blocks);
-    WriteOut(MSG_Get("PROGRAM_MEM_CONVEN"),blocks*16/1024);
+	uint16_t seg, blocks;
+	blocks = 0xffff;
+	DOS_AllocateMemory(&seg, &blocks);
+	WriteOut(MSG_Get("PROGRAM_MEM_CONVEN"), blocks * 16 / 1024);
 
-    if (umb_start!=0xffff) {
-        DOS_LinkUMBsToMemChain(1);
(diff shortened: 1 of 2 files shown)


src/dos/program_mem.cpp
```diff
@@ -83,6 +94,12 @@ void MEM::Run(void)
 		}
 	}
 
+	// Free the base and UMB allocations made above
+	while (!allocated_segs.empty()) {
+		DOS_FreeMemory(allocated_segs.top());
+		allocated_segs.pop();
+	}
+
```
johnnovak, line 102: Huh? They _allocated_ memory to check how much memory is free, then didn't free it? 🤦🏻 Can't we figure out how much free memory there is without _allocating_ it? 🤣

kcgen: It's probably easiest way to gather MEM's stats; like total number of blocks, largest chunk, and combined amount. 

MEM sits outside the "kernel" so also isn't privy to internals.. and without calls that give back these values, MEM would have to scan the memory for the memory chain block structure and so on (which would re-implement a lot of what DOS internals do).

johnnovak: Kinda makes sense, okay... But... man... You think you've seen everything, then this... 😐

johnnovak (approved):
👍🏻

Grounded0 (approved):

kcgen:
> shouldn’t the DOS free all the allocated memory once the executable quits

Critical point, @FeralChild64 .  Indeed - and the fault was my own on the freeing side: a typo when looking for corruption in the UMB chain (part of the `mcb_fault_strategy`).

Fixed in the updated push.

src/dos/program_mem.cpp
kcgen: Yup.. FreeDOS's MEM sources inspect the data structures instead allocating, which weighs in just under 4000 LOC across two files:
- https://gitlab.com/FreeDOS/base/mem/-/blob/master/SOURCE/MEM/MEM.C
- https://gitlab.com/FreeDOS/base/mem/-/blob/master/SOURCE/MEM/MEM2.C

johnnovak: I think this could screw up TSRs perhaphs that do try to allocate memory while MEM is doing it's weird alloc the world the free it dance... Probably low priority, but just saying. You could probably cause interesting failures in those DOS multitasking environments too by running some memory intensive program while issuing a few MEM commands simultaneously. Ugh...

johnnovak:
> One question: shouldn’t the DOS free all the allocated memory once the executable quits (and doesn’t turn into TSR)?

Is that so? I thought only modern OSes do that, Windows NT 4.0 and Windows 2000 and later.

Although DOS probably has some memory protection? Don't really know, TBH. On AmigaOS there's zero protection and anything goes, one rogue wtmrtie by one program can crash the whole system and probably no memory gets freed automatically when something just crashes (and the OS survived the crash 😄)

FeralChild64:
@johnnovak AmigaOS does not remember which process allocated the memory. DOS does, see here: https://stanislavs.org/helppc/memory_control_block.html

Once the program ends, it’s PSP is destroyed and DOS knows that it can free memory blocks owned by this PSP.

johnnovak:
Yeah, [Microsoft's MS-DOS 3.3 Programmer's Reference](https://www.pcjs.org/documents/books/mspl13/msdos/dosref33/) says this:

> When a program exits, MS-DOS automatically frees its initial memory-
allocation block before returning control to the calling program
(command.com is usually the calling program for application programs). The
DOS frees any memory owned by the exiting process.

What made me unsure what these sentences from another document:

https://www.plantation-productions.com/Webster/www.artofasm.com/DOS/ch13/CH13-5.html

> MS-DOS Memory Management Functions [...]
> MS-DOS provides three memory management functions- allocate, deallocate, and resize (modify). For most programs, these three memory allocation calls are not used. When DOS executes a program, it gives all of the available memory, from the start of that program to the end of RAM, to the executing process. Any attempt to allocate memory without first giving unused memory back to the system will produce an "insufficient memory" error.

Then later:

>Misusing these commands may result in loss of system memory that can be reclaimed only by rebooting the system. 

But the info about TSRs implies the automatic memory freeing behaviour, it's just not explicitly stated anywhere like in the M$ doco:

> This function also terminates program execution, but upon returning to DOS, the memory in use by the process is not returned to the DOS free memory pool.

PR #2541, opened 2023-05-27
Fix a couple regressions in `main` relating to MIDI refactoring

### LINKED ISSUE #1807: More consistent handling of enabling/disabling features
## Consistent handling of enabling/disabling features

Currently, boolean flags in the config accept the following values:
- true, on, enabled, 1
- false, off, disabled, 0

This is a really nice way of handling flags, I fully approve. But when "toggle states" are included in a list of enum values (e.g. for `reverb` the values are `off, on, tiny, small, medium, large, huge`), we only accept whatever we included in the enums for "enabling" or "disabling" the feature (so only `on` and `off` in this case, but not `enabled` or `disabled`, for example).

Some examples to highlight the numerous inconsistencies with toggle flags in enum values:

- `sidmodel`: auto, 6581, 8580, **none**
- `pcspeaker`: discrete, impulse, **none**, **off**
- `mpu401`: intelligent, uart, **none**
- `cms_filter`: **on**, **off**
- `composite`: auto, **on**, **off**

The problem is that the user has to remember whether to use `none` or `off` for every single config param. For config params that accept enum values but also `on` and `off` (or `none`), it's confusing why you cannot just use `disabled` or `false` instead of say `none`, like you can with simple boolean toggles.

I'd say it would be better if we did the following cleanup across the board:

- Extend list of accepted "disabled" boolean values with `none`
- Introduce two helper functions to parse disabled and enabled values, respectively, e.g. `parseEnabledValue`, `parseDisabledValue`. These would use the same set of allowed boolean values for parsing, *minus* the numeric `0` and `1` values, because those could interfere with more complex config params that allow either string enums or numbers (e.g. `crossfeed`).
- Using the above two methods for parsing enums, the enabled/disabled values would be automatically aliased to the full set of allowed values.
- To keep the noise levels down in the descriptions, we should always just mentione *one* enum value that is the "most natural fit" for the feature (e.g. `off` for a filter setting, but `none` for the SID model.) 

## Invalid boolean values shouldn't enable features

Try the following in the DOS prompt:

```
gus = asdf
ps1audio = qwer
```

You'll get the "not a valid value" error in both cases, yet the audio interfaces will be enabled. Haven't tried all boolean flags, but probably many would exhibit the same behaviour. I'd say a no-op would be the best course of action, and it that's not possible, just using the default (which is "disabled" in these cases).

### LINKED ISSUE #2397: Supaplex locks up in `main` unless MPU+MIDI is enabled
Thanks @rarioj for this submission. It's been bisected to 9a42619 and reproduced with the following:

Use the attached self-contained zip that has all the audio devices off: [2397.tar.gz :file_folder: ](https://github.com/dosbox-staging/dosbox-staging/files/11216912/2397.tar.gz)

1. `git checkout 9a42619` should hang when clicking and moving the mouse randomly inside the controls menu.

2. `git checkout 9a42619^` will not hang when clicking and moving the mouse randomly inside the controls menu.

### LINKED ISSUE #2460: Gauntlet locks up in `main` unless MPU+MIDI is enabled (but isn't a MIDI game)
Gauntlet's EGA 256x200 mode doesn't work in `main` but did in prior releases:

### 0.80.1 release:

```ini
[sdl]
windowresolution = 1280x960
output = texture
presentation_mode = vfr
vsync = false
vsync_skip = 0

[dosbox]
machine=ega
memsize=1
automount = false
startup_verbosity = quiet

[cpu]
core = normal
cycles = 2000

[sblaster]
sbtype = none

[midi]
mididevice = none

[speaker]
pcspeaker = none

[joystick]
joysticktype = disabled

[ethernet]
ne2000 = false

[autoexec]
mount c .
c:
cd Gauntlet
GAUNTLET
```

Repro pack: [2460.zip](https://github.com/dosbox-staging/dosbox-staging/files/11413252/2460.zip)

Steps:
 1. Press `Enter` to get to the player selection menu
 2. Use left arrow and slight up arrow to hover over the **'One'** player, press `Enter`
 3. Press `Enter` again to select the warrior.
 4. The new 256x200 resolution is set and gameplay either starts or freezes

### LINKED ISSUE #2547: Indiana Jones and the Fate of Atlantis no MT-32 music, log shows "MT32: Audio buffer underrun"

0.81.0-alpha-785

Mac OS

_No response_

### Relevant hardware info

M2 mini

When launching the above game using MT-32 for music, no music is heard and the log shows the above error messages.

Repro pack:
https://mega.nz/file/0hRVQR4K#yYbjOlLsONWqoE7neWlarxGzztZwwmMw8QekKLHZdBA

**The problem seems to arise because `config -set mididevice=mt32` is called twice in the configuration i uploaded.**

Use above repro pack plz.

_No response_

### Your configuration

`-noprimaryconf`

```
imgmount d cd\*.cue -t cdrom
config -set memsize=63
config -set mididevice=mt32
```

`atlantis r`

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
Objective is to these, which worked in 0.80.1 and prior:
- Fix #2397
- Fix #2460 

Tangentially:
- Fixes #1807
- Fixes #2547 

(suggest reading commit-by-commit)

### DIFF
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -91,19 +91,9 @@ size_t safe_strlen(char (&str)[N]) noexcept
 	return strnlen(str, N - 1);
 }
 
-template <size_t N>
-bool starts_with(const char (&pfx)[N], const char *str) noexcept
-{
-	return (strncmp(pfx, str, N - 1) == 0);
-}
+bool starts_with(const std::string_view prefix, const std::string_view str) noexcept;
 
-template <size_t N>
-bool starts_with(const char (&pfx)[N], const std::string &str) noexcept
-{
-	return (strncmp(pfx, str.c_str(), N - 1) == 0);
-}
-
-bool ends_with(const std::string &str, const std::string &suffix) noexcept;
+bool ends_with(const std::string_view str, const std::string_view suffix) noexcept;
 
 bool find_in_case_insensitive(const std::string &needle, const std::string &haystack);
 
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -91,7 +91,7 @@ size_t safe_strlen(char (&str)[N]) noexcept
 	return strnlen(str, N - 1);
 }
 
-bool starts_with(const std::string_view prefix, const std::string_view str) noexcept;
+bool starts_with(const std::string_view str, const std::string_view prefix) noexcept;
 
 bool ends_with(const std::string_view str, const std::string_view suffix) noexcept;
 
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -324,11 +324,11 @@ void dos_to_utf8(const std::string& str_in, std::string& str_out,
 //   else
 //       log_warning("%s was invalid", s.c_str());
 //
-std::optional<float> parse_value(const std::string &s, const float min_value,
-                                 const float max_value);
+std::optional<float> parse_value(const std::string_view s,
+                                 const float min_value, const float max_value);
 
 // parse_value clamped between 0 and 100
-std::optional<float> parse_percentage(const std::string &s);
+std::optional<float> parse_percentage(const std::string_view s);
 
 // Parse a value from a character-prefixed string, clamp the result within the
 // given min and max values, and return it as a float. This API should give us
--- a/src/hardware/innovation.cpp
+++ b/src/hardware/innovation.cpp
@@ -28,18 +28,21 @@
 
 CHECK_NARROWING();
 
-void Innovation::Open(const std::string &model_choice,
-                      const std::string &clock_choice, const int filter_strength_6581,
+void Innovation::Open(const std::string_view model_choice,
+                      const std::string_view clock_choice,
+                      const int filter_strength_6581,
                       const int filter_strength_8580, const int port_choice,
-                      const std::string &channel_filter_choice)
+                      const std::string_view channel_filter_choice)
 {
 	Close();
 
 	// Sentinel
-	if (model_choice == "none")
+	const auto model_choice_has_bool = parse_bool_setting(model_choice);
+	if (model_choice_has_bool && *model_choice_has_bool == false) {
 		return;
+	}
 
-	std::string model_name;
+	std::string_view model_name = "";
 	int filter_strength = 0;
 	auto sid_service    = std::make_unique<reSIDfp::SID>();
 
@@ -87,9 +90,13 @@ void Innovation::Open(const std::string &model_choice,
 	                                       ChannelFeature::Synthesizer});
 
 	if (!mixer_channel->TryParseAndSetCustomFilter(channel_filter_choice)) {
-		if (channel_filter_choice != "off")
+		const auto filter_choice_has_bool = parse_bool_setting(
+		        channel_filter_choice);
+
+		if (!filter_choice_has_bool) {
 			LOG_WARNING("INNOVATION: Invalid 'innovation_filter' value: '%s', using 'off'",
-			            channel_filter_choice.c_str());
+			            channel_filter_choice.data());
+		}
 
 		mixer_channel->SetHighPassFilter(FilterState::Off);
 		mixer_channel->SetLowPassFilter(FilterState::Off);
@@ -124,12 +131,12 @@ void Innovation::Open(const std::string &model_choice,
 	if (filter_strength == 0)
 		LOG_MSG("INNOVATION: Running on port %xh with a SID %s at %0.3f MHz",
 		        base_port,
-		        model_name.c_str(),
+		        model_name.data(),
 		        chip_clock / us_per_s);
 	else
 		LOG_MSG("INNOVATION: Running on port %xh with a SID %s at %0.3f MHz filtering at %d%%",
 		        base_port,
-		        model_name.c_str(),
+		        model_name.data(),
 		        chip_clock / us_per_s,
 		        filter_strength);
 
--- a/src/hardware/lpt_dac.cpp
+++ b/src/hardware/lpt_dac.cpp
@@ -31,10 +31,11 @@
 #include "disney.h"
 #include "ston1_dac.h"
 
-LptDac::LptDac(const std::string &name, const uint16_t channel_rate_hz,
+LptDac::LptDac(const std::string_view name, const uint16_t channel_rate_hz,
                channel_features_t extra_features)
         : dac_name(name)
 {
+	assert(!dac_name.empty());
 	using namespace std::placeholders;
 	const auto audio_callback = std::bind(&LptDac::AudioCallback, this, _1);
 
@@ -48,7 +49,7 @@ LptDac::LptDac(const std::string &name, const uint16_t channel_rate_hz,
 	// Setup the mixer callback
 	channel = MIXER_AddChannel(audio_callback,
 	                           channel_rate_hz,
-	                           dac_name.c_str(),
+	                           dac_name.data(),
 	                           features);
 
 	ms_per_frame = millis_in_second / channel->GetSampleRate();
@@ -58,7 +59,7 @@ LptDac::LptDac(const std::string &name, const uint16_t channel_rate_hz,
 	status_reg.busy  = false;
 }
 
-bool LptDac::TryParseAndSetCustomFilter(const std::string filter_choice)
+bool LptDac::TryParseAndSetCustomFilter(const std::string_view filter_choice)
 {
 	assert(channel);
 	return channel->TryParseAndSetCustomFilter(filter_choice);
@@ -118,7 +119,7 @@ void LptDac::AudioCallback(const uint16_t requested_frames)
 
 LptDac::~LptDac()
 {
-	LOG_MSG("%s: Shutting down DAC", dac_name.c_str());
+	LOG_MSG("%s: Shutting down DAC", dac_name.data());
 
 	// Update our status to indicate we're no longer ready
 	status_reg.error = true;
@@ -153,8 +154,8 @@ void LPT_DAC_Init(Section *section)
 	// Get the user's LPT DAC choices
 	assert(section);
 	const auto prop = static_cast<Section_prop *>(section);
-	const std::string dac_choice    = prop->Get_string("lpt_dac");
-	const std::string filter_choice = prop->Get_string("lpt_dac_filter");
+
+	const std::string_view dac_choice = prop->Get_string("lpt_dac");
 
 	if (dac_choice == "disney")
 		lpt_dac = std::make_unique<Disney>();
@@ -162,17 +163,30 @@ void LPT_DAC_Init(Section *section)
 		lpt_dac = std::make_unique<Covox>();
 	else if (dac_choice == "ston1")
 		lpt_dac = std::make_unique<StereoOn1>();
-	else if (dac_choice == "none" || dac_choice == "off")
+	else {
+		const auto dac_choice_has_bool = parse_bool_setting(dac_choice);
+		if (!(dac_choice_has_bool && *dac_choice_has_bool)) {
+			LOG_WARNING("LPT_DAC: Invalid 'lpt_dac' choice: '%s', using 'none'",
+			            dac_choice.data());
+		}
 		return;
+	}
 
+	// Let the DAC apply its own filter type
+	const std::string_view filter_choice = prop->Get_string("lpt_dac_filter");
 	assert(lpt_dac);
-
 	if (!lpt_dac->TryParseAndSetCustomFilter(filter_choice)) {
-		const auto filter_state = filter_choice == "on" ? FilterState::On
-														: FilterState::Off;
-		if (filter_state == FilterState::Off && filter_choice != "off")
+		auto filter_state = FilterState::Off;
+		const auto filter_choice_has_bool = parse_bool_setting(filter_choice);
+
+		if (filter_choice_has_bool) {
+			filter_state = *filter_choice_has_bool ? FilterState::On
+			                                       : FilterState::Off;
+		} else {
 			LOG_WARNING("LPT_DAC: Invalid 'lpt_dac_filter' value: '%s', using 'off'",
-						filter_choice.data());
+			            filter_choice.data());
+			assert(filter_state == FilterState::Off);
+		}
 
 		lpt_dac->ConfigureFilters(filter_state);
 	}
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -993,8 +993,9 @@ bool MixerChannel::TryParseAndSetCustomFilter(const std::string_view filter_pref
 	SetLowPassFilter(FilterState::Off);
 	SetHighPassFilter(FilterState::Off);
 
-	if (!(starts_with("lpf", filter_prefs) || starts_with("hpf", filter_prefs)))
+	if (!(starts_with(filter_prefs, "lpf") || starts_with(filter_prefs, "hpf"))) {
 		return false;
+	}
 
 	const auto parts = split(filter_prefs, ' ');
 
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -396,12 +396,16 @@ static void configure_reverb(const std::string_view reverb_pref)
 
 static ChorusPreset chorus_pref_to_preset(const std::string_view pref)
 {
-	if (pref == "off")
-		return ChorusPreset::None;
-	if (pref == "light")
+	const auto pref_has_bool = parse_bool_setting(pref);
+	if (pref_has_bool) {
+		return *pref_has_bool ? ChorusPreset::Normal : ChorusPreset::None;
+	}
+	if (pref == "light") {
 		return ChorusPreset::Light;
-	if (pref == "normal" || pref == "on")
+	}
+	if (pref == "normal") {
 		return ChorusPreset::Normal;
+	}
 	if (pref == "strong")
 		return ChorusPreset::Strong;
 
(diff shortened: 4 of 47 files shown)


src/hardware/pic.cpp
```diff
@@ -642,7 +642,16 @@ class PIC_8259A final : public Module_base {
...
 
-		if (machine==MCH_PCJR) {
+		// On AT systems, Both IRQ 2 and 9 were available.
+		// Ref: https://dosdays.co.uk/topics/io_addresses_irq_dma.php
+		// If present, the MPU-401 activates and uses IRQ 9.
+		//
+		if (machine >= MCH_EGA) {
+			PIC_SetIRQMask(9, false);
+			PIC_DeActivateIRQ(9);
+		}
```
johnnovak, line 652: Nice 👍🏻

src/hardware/mpu401.cpp
```diff
@@ -780,16 +780,17 @@ class MPU401 final : public Module_base {
...
+		assert(section);
 
-		Section_prop *section = static_cast<Section_prop *>(configuration);
-		const auto mpu_type = std::string(section->Get_string("mpu401"));
-		if (mpu_type == "none" || mpu_type == "off" || mpu_type == "false")
-			return;
+		const std::string_view mpu_type = section->Get_string("mpu401");
+
+		installed = (mpu_type != "none" && mpu_type != "off" &&
+		             mpu_type != "false");
```
johnnovak, line 789: This misses "disabled" and "no".

I suggest factoring out the boolean string checks into helper functions from `Value::SetBool` in `setup.cpp`, then use that here (e.g. `parse_bool_setting(mpu_type)`).

`std::optional<bool> parse_bool_setting()` would be probably the best; then it's the client's call whether to raise  and error on invalid values, or just default to whatever the default is (could be either "enabled" or "disabled"; it would be a mistake to always default to "disabled", so "false").

kcgen:
(still need to do Innovation)

include/string_utils.h
```diff
@@ -91,19 +91,9 @@ size_t safe_strlen(char (&str)[N]) noexcept
...
-{
-	return (strncmp(pfx, str, N - 1) == 0);
-}
-
-template <size_t N>
-bool starts_with(const char (&pfx)[N], const std::string &str) noexcept
-{
-	return (strncmp(pfx, str.c_str(), N - 1) == 0);
-}
+bool starts_with(const std::string_view prefix, const std::string_view str) noexcept;
```
johnnovak, line 94: While you're at it, could you please switch the order of arguments so it's consistent with `starts_with` and general conventions where the first arg is the receiver/main operand?

I've done this change in my capture PR already, btw.

kcgen: Will do!

src/hardware/innovation.cpp
```diff
@@ -28,18 +28,21 @@
...
                       const int filter_strength_8580, const int port_choice,
-                      const std::string &channel_filter_choice)
+                      const std::string_view channel_filter_choice)
 {
 	Close();
 
 	// Sentinel
-	if (model_choice == "none")
+	const auto model_choice_has_bool = parse_bool_setting(model_choice);
+	if (model_choice_has_bool && *model_choice_has_bool == false) {
```
johnnovak, line 41: Nice usage pattern 🎉

src/hardware/lpt_dac.cpp
```diff
@@ -153,26 +154,39 @@ void LPT_DAC_Init(Section *section)
...
 	if (dac_choice == "disney")
 		lpt_dac = std::make_unique<Disney>();
 	else if (dac_choice == "covox")
 		lpt_dac = std::make_unique<Covox>();
 	else if (dac_choice == "ston1")
 		lpt_dac = std::make_unique<StereoOn1>();
-	else if (dac_choice == "none" || dac_choice == "off")
+	else {
+		const auto dac_choice_has_bool = parse_bool_setting(dac_choice);
+		if (!dac_choice_has_bool || *dac_choice_has_bool != false) {
```
johnnovak, line 168: Personally, I find this easier to parse mentally:

```cpp
if (!(dac_choice_has_bool && *dac_choice_has_bool == true))
```

src/hardware/mixer.cpp
```diff
@@ -326,10 +336,13 @@ static ReverbPreset reverb_pref_to_preset(const std::string_view pref)
...
 
-static void configure_reverb(std::string reverb_pref)
+static void configure_reverb(const std::string_view reverb_pref)
 {
 	auto was_reverb_on = mixer.do_reverb;
-	mixer.do_reverb = (reverb_pref != "off");
+
+	const auto reverb_pref_has_bool = parse_bool_setting(reverb_pref);
+
+	mixer.do_reverb = (!reverb_pref_has_bool || *reverb_pref_has_bool == true);
```
johnnovak, line 345: I think this criteria is wrong? Should be:

```cpp
mixer.do_reverb = (reverb_pref_has_bool && *reverb_pref_has_bool == false) ? false : true;
``` 

Because reverb can be `off`, `disabled`, `no` etc. for false, but if it `on`, `enabled`, etc. _or_ if it's set to a preset name like `small`, `medium`, etc, then it should be enabled.

johnnovak: Better yet, `reverb_pref_to_preset` handles this already. So maybe call that, and do:

```
mixer.do_reverb = (reverb_pref_to_preset(reverb_pref) != ReverbPreset::None);
```

src/hardware/mixer.cpp
```diff
@@ -401,10 +414,13 @@ static ChorusPreset chorus_pref_to_preset(const std::string_view pref)
...
 
-static void configure_chorus(std::string chorus_pref)
+static void configure_chorus(const std::string_view chorus_pref)
 {
 	auto was_chorus_on = mixer.do_chorus;
-	mixer.do_chorus = (chorus_pref != "off");
+
+	const auto chorus_pref_has_bool = parse_bool_setting(chorus_pref);
+
+	mixer.do_chorus = (!chorus_pref_has_bool || *chorus_pref_has_bool == true);
```
johnnovak, line 423: How about

```
mixer.do_chorus = (chorus_pref_to_preset(chorus_pref) != ChorusPreset::None);
```

src/hardware/mpu401.cpp
```diff
@@ -57,50 +57,88 @@ constexpr uint8_t MSG_MPU_CLOCK = 0xfd;
...
+	uint8_t queue_pos           = 0;
+	uint8_t queue_used          = 0;
+
+	MpuTrack playbuf[8] = {};
+	MpuTrack condbuf    = {};
+
+	MpuState state = {};
+
+	MpuClock clock = {};
+};
```
johnnovak, line 131: Nice cleanup 👍🏻

src/hardware/mpu401.cpp
```diff
@@ -742,25 +781,24 @@ static void MPU401_Reset()
...
-		if (mpu_type == "none" || mpu_type == "off" || mpu_type == "false")
-			return;
+		const std::string_view mpu_choice = section->Get_string("mpu401");
 
-		// Enabled and there is a Midi
-		installed = true;
+		if (const auto has_bool = parse_bool_setting(mpu_choice);
+		    has_bool && *has_bool == false) {
+			return;
+		}
```
johnnovak, line 801: Nice consistency improvements 👍🏻

src/hardware/pcspeaker.cpp
```diff
@@ -42,28 +42,33 @@ void PCSPEAKER_Init(Section *section)
...
 	if (!pc_speaker->TryParseAndSetCustomFilter(filter_choice)) {
-		if (filter_choice == "on") {
+		const auto filter_choice_has_bool = parse_bool_setting(filter_choice);
+
+		if (filter_choice_has_bool && *filter_choice_has_bool == true) {
 			pc_speaker->SetFilterState(FilterState::On);
 		} else {
-			if (filter_choice != "off") {
+			if (!filter_choice_has_bool) {
 				LOG_WARNING("PCSPEAKER: Invalid 'pcspeaker_filter' value: '%s', using 'off'",
```
johnnovak, line 72: Consider slightly reorganising this:

```cpp
if (filter_choise_has_bool) {
    if (*filter_filterchoice_has_bool) {
        ....
    }
} else {
    ....
}
```

src/hardware/reelmagic/driver.cpp
```diff
@@ -1396,11 +1396,18 @@ void ReelMagic_Init(Section* sec)
...
 
 	// Does the user want ReelMagic emulation?
-	const auto reelmagic_choice = std::string_view(section->Get_string("reelmagic"));
-	const auto wants_reelmagic  = (reelmagic_choice == "on" || reelmagic_choice == "cardonly");
+	const auto reelmagic_choice = std::string_view(
+	        section->Get_string("reelmagic"));
 
-	if (!wants_reelmagic) {
-		if (reelmagic_choice != "off") {
+	const auto wants_card_only = reelmagic_choice == "cardonly";
```
johnnovak, line 1402: Enclose in brackets.

src/misc/setup.cpp
```diff
@@ -415,6 +402,21 @@ bool Prop_string::IsValidValue(const Value& in)
...
+		const auto is_bool = parse_bool_setting(in.ToString());
+		if (is_bool) {
+			if (*is_bool == true && is_positive_bool_valid) {
+				return true;
+			}
+			if (*is_bool == false && is_negative_bool_valid) {
+				return true;
+			}
+		}
+	}
```
johnnovak, line 418: This is a bit "inside out" 😅 Could "invert" it and make it a lot simpler by using a `valid_bool_value` lambda, something like this:

```cpp
if (is_positive_bool_valid && valid_bool_value(in, true)) {
   return true;
} else if (is_negative_bool_valid && valid_bool_value(in, false)) {
   return false;
}
```

src/misc/setup.cpp
```diff
@@ -674,13 +676,36 @@ const std::vector<Value>& PropMultiVal::GetValues() const
...
+// boolean string (ie: "on" or "off"). If so, this records if a boolean is valid
+// and its direction (either positive or negative) so we can accept all of those
+// corresponding boolean strings from the user (ie: composite = disabled")
+//
+void Property::MaybeSetBoolValid(const std::string_view valid_value)
+{
+	// Deliberately ignore the numerical bools to avoid false-positive
+	if (valid_value == "0" || valid_value == "1") {
+		return;
+	}
```
johnnovak, line 690: Nice 👍🏻

johnnovak: Actually, I'm wondering if this approach is a bit backwards and error prone. How about _only_ "whitelisting" `0` and `1` as valid values in the "pure bool" config param handler in `setup.cpp`, but not in `parse_bool_setting`? So to be clear, `parse_bool_setting` would only handle the textual descriptions and _not_ `0` and `1`.

The reason I'd prefer to do it this way is because this could result in subtle bugs that might go unnoticed for multi-value settings that accept either `off|none|disabled|no` for "disabling" the feature, and an integer for the "enabled" state.

src/misc/setup.cpp
```diff
@@ -674,13 +676,36 @@ const std::vector<Value>& PropMultiVal::GetValues() const
...
+	const auto is_valid_value_bool = parse_bool_setting(valid_value);
+	if (is_valid_value_bool) {
+		if (*is_valid_value_bool == true) {
+			is_positive_bool_valid = true;
+		} else {
+			is_negative_bool_valid = true;
+		}
```
johnnovak, line 697: Set the other to false, as a precaution?

src/misc/setup.cpp
```diff
@@ -1434,6 +1460,27 @@ parse_environ_result_t parse_environ(const char* const* envp) noexcept
...
+	        {"0", false},
+	        {"disabled", false},
+	        {"false", false},
+	        {"off", false},
+	        {"no", false},
+	        {"none", false},
+	};
+
+	const auto it = mapping.find(setting);
+	return it != mapping.end() ? std::optional<bool>(it->second) : std::nullopt;
```
johnnovak, line 1481: minor: can use {} instead of std::nullopt

src/misc/string_utils.cpp
```diff
@@ -268,27 +268,41 @@ void strip_punctuation(std::string &str)
 	          str.end());
 }
 
-bool ends_with(const std::string &str, const std::string &suffix) noexcept
+// TODO in C++20: replace with str.starts_with(prefix)
```
johnnovak, line 271: One can dream 😎

johnnovak (approved):
Great cleanup work @kcgen in addition to fixing the original issue 😄 I love it that you've made the handling of "enabled" and "disabled" settings consistent across the board; that's been annoying me for quite some while.

Made a few suggestions for improving this a little bit further, but approving anyway!

src/hardware/mixer.cpp
```diff
@@ -267,17 +267,23 @@ static void set_global_crossfeed(mixer_channel_t channel)
...
+	const auto crossfeed_pref_has_bool = parse_bool_setting(crossfeed_pref);
+
+	if (crossfeed_pref_has_bool) {
+		if (*crossfeed_pref_has_bool == true) {
+			constexpr auto default_crossfeed_strength = 0.40f;
+			crossfeed = default_crossfeed_strength;
+		} else { // false
+			crossfeed = 0.0f;
+		}
+	}
```
johnnovak, line 280: I think because a value of "1" is treated as "enabled", `crossfeed = 1` would set crossfeed to 40, actually. Then `crossfeed = 0` would work fine by fluke because that means disabled anyway 😄 

I have a proposed solution for this in a later comment below (basically excluding `0` and `0` from `parse_bool_setting`, and only handling them as extra special values when parsing "pure bool" config settings.

johnnovak:
This also addresses https://github.com/dosbox-staging/dosbox-staging/issues/1807

The second issue in the referenced ticket ("Invalid boolean values shouldn't enable features") seems to have been addressed by someone (maybe me? 😅) since I raised it.

johnnovak:
Not necessarily asking you to do this now, but there appear to be further freeform multi-value settings that accept `none`, `off` etc., e.g `ps2_mouse_model`, `mididevice`, `sidmodel`,  `composite`. Writing the default config and searching for `none`, `on`, `off`, etc. should round them all up 😄

kcgen:
> ps2_mouse_model, mididevice, sidmodel, composite

Yup - they're all handled here :-)

> All *_filter settings accept off too, which could be generalised to none

Same with these (and more! :laughing: , like `ems = `). Give it a go!

src/misc/setup.cpp
```diff
@@ -1434,6 +1460,27 @@ parse_environ_result_t parse_environ(const char* const* envp) noexcept
...
+	        {"true", true},
+	        {"on", true},
+	        {"yes", true},
+
+	        {"0", false},
+	        {"disabled", false},
+	        {"false", false},
+	        {"off", false},
+	        {"no", false},
+	        {"none", false},
```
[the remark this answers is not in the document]
kcgen: Yup, I like this because I don't want to start encouraging users typing in 1's and 0's throughout their confs because it can look ambiguous too.

src/misc/setup.cpp
kcgen: They're separate (and not connected); they record if the property access positive bools and negative bools. For example:
 - `sbtype` only allows negatives but not positives, but
 - `composite` allows both positives and negatives

src/misc/setup.cpp
kcgen: Yup - agree that we should snip the numeric aliasing to bool.

src/midi/midi.cpp
```diff
@@ -68,24 +69,6 @@ uint8_t MIDI_message_len_by_status[256] = {
...
-// the end of the chain.
-//
-// This is hacky as hell and needs to be rewritten in a sane way at some
-// point...
-MidiHandler::MidiHandler() : next(handler_list)
-{
-	handler_list = this;
-}
-
 /* Include different midi drivers, lowest ones get checked first for default.
```
johnnovak, line 72: 🥇 for that! 🎊  I hate those ad-hoc linked lists sprinkled through various random places of the codebase... Especially the way how they link them together, as explained in my above comment. So bad!

kcgen: It bit me when I was testing the mpu-401's on - off - on handling at runtime ; there was no way to clear or invalidate prior midi handlers (without adding more complexity into it). 

Ps -- thanks for all the other comments. Will get 'em fixed up tomorrow.

src/hardware/mixer.cpp
kcgen: > I think this criteria is wrong?

`mixer.do_reverb = (!reverb_pref_has_bool || *reverb_pref_has_bool == true);`

- The first criteria (`!reverb_pref_has_bool`) checks if a custom value is provided (so I think that's OK).

- The second criteria only gets tested if the first isn't true. So if we're not using a non-custom value (in other words, if a bool was provided), then test if the bool was positive (so I think that's OK, too).

(given the <optional> itself is testable as boolean combined with the content itself being bool, I find it a bit hard on the eyes!)

Definitely like the re-use of `reverb_pref_to_preset` better.

src/hardware/pcspeaker.cpp
kcgen: Much improved!

src/misc/setup.cpp
kcgen: Unfortunately in ternaries, it's less flexible.. :sob: 

```c++
error: initializer list cannot be used on the right hand side of operator ':'
        return it != mapping.end() ? std::optional<bool>(it->second) : {};
```

src/hardware/pcspeaker.cpp
johnnovak: The same improved pattern applies to a lot of other files you touched I just didn't comment on all of them 😄

src/misc/setup.cpp
johnnovak: Ah, bummer.

kcgen:
Going to merge this - all commits build and run, and confirmed fixing affected issues.

This also options up using the `parse_bool_setting()` for use with new conf settings in other PRs.

PR #2542, opened 2023-05-27
Copy only the available command line chars

LowLevelMahn:
Prevent buffer overflow or MSVC STL assert on overflow

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -40,7 +40,8 @@
 #endif
 struct CommandTail{
 	uint8_t count = 0;     /* number of bytes returned */
-	char buffer[127] = {}; /* the buffer itself */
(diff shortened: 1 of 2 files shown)


include/dos_inc.h
```diff
@@ -40,7 +40,8 @@
 #endif
 struct CommandTail{
 	uint8_t count = 0;     /* number of bytes returned */
-	char buffer[127] = {}; /* the buffer itself */
+	static constexpr size_t max_cmdtail_buffer_size = 126;
```
kcgen, line 43: Use CamelCase for constants.

kcgen (approved):
Thanks @LowLevelMahn :rocket: 

Also, thank you for continuing your tests with the checked STL under MSVC.  I will have a follow up PR that turns on some level of equivalent checks for GCC and Clang shortly.

PR #2555, opened 2023-05-30
Tab completion fixes for / and .

### LINKED ISSUE #2554: Loss of shell `/<tab>` expansion
Reproduce:
```shell
meson setup -Dbuildtype=debug build
meson compile -C build
./build/dosbox .
```

At the DOS prompt, type:
1. `y:`
2. `cd dos`
3. `cd /<tab>`:
  - `main` correctly expands 'DOS' however it deletes the forward slash, prevent the `cd ...` command from being able to actually change to the requested directory.
  - `0.80.1` correctly expands to '/DOS', allowing the user to press enter and `cd ...` into the expanded path.

MeAreJeenius:
Fixes #2554 

Also found an issue where completions wouldn't be found if the original searched command had a period.

### DIFF
--- a/src/shell/shell_misc.cpp
+++ b/src/shell/shell_misc.cpp
@@ -229,7 +229,7 @@ std::string DOS_Shell::ReadCommand()
 				}
 				completion_index = 0;
 
-				const auto delimiter = command.find_last_of(" \\");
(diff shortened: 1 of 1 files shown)


kcgen:
Thanks for another quick fix, @MeAreJeenius.

PR #2563, opened 2023-05-31
 Adapt to setup function additions

kcgen:
The addition of the checked STL flags (on Linux) caused crashes inside the Setup class' `ExecuteInit()`. 

The crash happens because the container that ExecuteInit is looping over gets modified (appended to) mid-iteration, which invalidates the loop iterator.

A prior attempt to avoid this moved the appended chunk to a separate container (EarlyInit), however this broke the startup sequence for the mapper.

A prior attempt to solve this also failed:
- #2557

This PR is now a third attempt to handle the modified container (and avoid crashing with the checked STL) while also keeping the original mapper placement so the startup sequence/functionality is retained.

### DIFF
--- a/include/setup.h
+++ b/include/setup.h
@@ -307,7 +307,7 @@ private:
 	};
 
 	std::deque<Function_wrapper> early_init_functions = {};
-	std::deque<Function_wrapper> initfunctions        = {};
+	std::deque<Function_wrapper> init_functions       = {};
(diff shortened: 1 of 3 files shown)


GranMinigun (approved):

kcgen:
Thanks for the tests, @GranMinigun. Also working here and passing regression tests w/ the mapper. Given this is impacting most builds, let's get this merged ASAP.

Issue #2546, opened 2023-05-28
kkowboy (demo) crashes with FPU stack overflow


johnnovak:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-alpha

### What Operating System are you using?

Mac OS

### Relevant hardware info

M1 mac

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The demo crashes with "ABORT: FPU stack overflow" in the logs after the sine scroller part about 2 minutes in.

If I change `DB_FPU_STACK_CHECK_PUSH` to `DB_FPU_STACK_CHECK_NONE` for the release build, it runs flawlessly. I'm inclined to think we should just do that to at least give a chance for erratical programs to run instead of a hard exit. Then if they mess up the stack so badly then will crash anyway somehow later, so be it...

### Steps to reproduce the behaviour.

Use repro pack, start with `-noprimaryconf`.

### Your configuration

```ini
[dosbox]
memsize = 48
vmemsize = 8

[render]
glshader = glshader/hyllian-updated-new.tweaked
integer_scaling = vertical

[cpu]
cycles=max 95% limit 400000

[sblaster]
sbtype = none

[gus]
gus = on

[autoexec]
c:
cd demos
cd kkowboy
kkowboy
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Repro pack:

[kkowboy.zip](https://github.com/dosbox-staging/dosbox-staging/files/11584082/kkowboy.zip)

Grounded0:
Real DOS didn't mollycoddle user over what type of program it would allow to execute. This design choice makes since its period correct and increases compatibility.

johnnovak:
> Real DOS didn't mollycoddle the user over what type of code it would allow to execute. This design choice makes sense as its period correct and increases software compatibility.

Yeah I think so; it's the difference between being able to watch the demo from start to finish vs a hard crash in the middle. It's unlikely people will go back and "fix" their old 30+ year old EXEs to play nicely with DOSBox...

PR #2564, opened 2023-06-01
Never hard-exit DOSBox when an x87 FPU stack overflow is detected

johnnovak:
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2546

There's no good reason to take such a drastic measure as to hard exit DOSBox when an FPU stack overflow is detected. Real hardware does not do that; the FPU stack over/underflow just happens and sets the overflow bit that the application can optionally inspect and recover from it, etc.

This can prevent programs that cause x87 stack over/underflows but recover from them later from running. I've run into one such example, see linked ticket with repro pack.

### DIFF
--- a/include/fpu.h
+++ b/include/fpu.h
@@ -187,7 +187,7 @@ static inline void FPU_LOG_WARN(unsigned tree, bool ea, uintptr_t group, uintptr
 #define DB_FPU_STACK_CHECK_PUSH DB_FPU_STACK_CHECK_EXIT
 #else
 #define DB_FPU_STACK_CHECK_POP DB_FPU_STACK_CHECK_NONE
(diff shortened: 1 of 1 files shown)


johnnovak:
@joncampbell123 Maybe you're interested in this.

joncampbell123:
:+1: I made a similar change in DOSBox-X a long time ago when some demos would trigger the hard exit despite running perfectly fine without such a hard exit. Though it was done differently.

johnnovak:
> 👍 I made a similar change in DOSBox-X a long time ago when some demos would trigger the hard exit despite running perfectly fine without such a hard exit. Though it was done differently.

Nice, looks like a safe change then.

PR #2568, opened 2023-06-02
De-SDL voodoo threading

kklobe:
Migrate all Voodoo SDL threading-related code to C++17. Add a simple semaphore implementation to replace SDL's semaphore. 

Mostly this is just an excuse for me to get a bit more experience with C++ threading; I didn't notice any tangible performance benefits.

### DIFF
--- a/include/semaphore.h
+++ b/include/semaphore.h
@@ -1,0 +1,66 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2021-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


include/semaphore.h
```diff
@@ -0,0 +1,52 @@
+#include <condition_variable>
```
kcgen, line 1: Add header block (copyright + year + GPL + preprocessor one-time-include check and definds).  rwqueue's header should be a good example.

kklobe: Hah, just did that 2 minutes before you posted, thanks!

kcgen (commented):
Suggest adding a couple unit tests to exercise it.

The rwqueue has a couple test cases that access a queue using multiple threads, I think this would be a good template to use here too.

kklobe:
> Suggest adding a couple unit tests to exercise it.
> 
> The rwqueue has a couple test cases that access a queue using multiple threads, I think this would be a good template to use here too.

I put a very basic unit test in there, will try to get more creative if desired.

kklobe:
P.S. I consider the semaphore class a short-term hack until we can use C++20's `std::counting_semaphore`

kcgen:
Nice to have an all C++17 implementation, @kklobe !
I'm not seeing any regression in testing, so let's get this merged :rocket:

PR #2569, opened 2023-06-02
Fix VESA 256 / 8-bit in Windows 3.1

kcgen:
The move the `vgaonly` broke Windows 3.1 in 256 colour-mode; screenshots of the damage, courtesy @johnnovak :

#### Windows 3.11 desktop

image

#### Myst

image

The `vgamode` transition included lookup of RGB666 (18-bit) palette values for VESA 256-colour modes, however drawing the hardware cursor was still done using the original non-palette lookup routine. This adapts the current HW cursor draw routine to use palette as well.

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -651,7 +651,7 @@ static uint8_t * VGA_Draw_LIN32_Line_HWMouse(Bitu vidstart, Bitu /*line*/) {
 }
 
 static const uint8_t* VGA_Text_Memwrap(Bitu vidstart) {
-	vidstart &= vga.draw.linear_mask;
+	vidstart = vidstart & vga.draw.linear_mask;
 	Bitu line_end = 2 * vga.draw.blocks;
 	if (GCC_UNLIKELY((vidstart + line_end) > vga.draw.linear_mask)) {
 		// wrapping in this line
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -523,6 +523,92 @@ static uint8_t * VGA_Draw_VGA_Line_HWMouse( Bitu vidstart, Bitu /*line*/) {
 	}
 }
 
+enum CursorOp : uint8_t {
+	Foreground  = 0b10,
+	Background  = 0b00,
+	Transparent = 0b01,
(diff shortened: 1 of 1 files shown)


src/hardware/vga_draw.cpp
```diff
@@ -469,57 +470,87 @@ static uint8_t* draw_linear_line_from_dac_palette(Bitu vidstart, Bitu /*line*/)
...
+	// ~~~~~~~~~~~~~~~~~
+	// The cursor is a 64x64 pattern which is shifted (inside the 64x64 mouse
+	// cursor space) to the right by posx pixels and up by posy pixels.
+	//
+	// This is used when the mouse cursor partially leaves the screen. It is
+	// arranged as bitmap of 16bits of bit A followed by 16bits of bit B, each AB
+	// bits corresponding to a cursor pixel. The whole map is 8kB in size.
+
+	// the index of the bit inside the cursor bitmap we start at:
+	const auto source_start_bit = ((line_at_y - cursor.originy) + cursor.posy) * 64 +
```
johnnovak, line 506: Extract 64 to `pitch`?

johnnovak:
Nice one @kcgen, works great, I just tested it with Myst!

Just for my own interest, so this had nothing to do with the commit I bisected the issue to? I tried it a few times, but maybe something went wrong; I had to bisect it manually because Win3.x could not be started on a range of commits (it just hard exited to the DOS prompt).

johnnovak (approved):

src/hardware/vga_draw.cpp
kcgen: Sounds like a reasonable name for that constant. Will also see if it relates to the 63 (magic number) in here to, which are carried in from the original HW cursor function.

kcgen:
> Just for my own interest, so this had nothing to do with the commit I bisected the issue to? 

Yes, the bisect went to the VESA RGB666 commit (which lacked the HW cursor additional)-- but I can't remember details if I had to skip or fix some of the bisect steps, so that's probably where our efforts differed.

(In main just now, reverting the Earthworm Jim commit doesn't fix win 3.1 in 256c mode -- but maybe checking out prior to that is before the breakage was introduced, so that probably makes sense from a chronological position).

PR #2583, opened 2023-06-08
(S)VGA/VESA square pixel mode fix/cleanup

johnnovak:
Improve & simplify the aspect ratio calculation logic for (S)VGA/VESA modes intended to be displayed with square pixels.

(S)VGA and VESA modes that have 4:3 storage aspect ratio (e.g., 320x240, 400x300, 640x480, 1024x768, 1152x864, etc.) are clearly intended to be displayed with square pixels on a 4:3 aspect ratio CRT monitor. The VGA CRTC register based pixel aspect ratio calculation logic yields values around 1.0 for these modes, but not exactly 1.0.

To remedy this, a simple hack was in place that just set the PAR of all 640x480+ modes to 1:1. But this approach had several problems:

- High-resolution VESA text modes that require non-square pixels (132-column 1056x344/400/480/688 resolution modes) were forced to incorrectly use square pixels.

- Did not correctly set square pixels for VESA modes below the 640x480 cutoff point that are clearly intended to be displayed with square pixels (e.g., 400x300, 512x384).

- Incorrectly forced square pixels for the 1280x1024 mode.

This change solves all the above issues while removing some further obsolete hacks. The change only affects standard, non-tweaked BIOS modes; tweaked modes are left completely alone.

---

Below is an overview of all standard BIOS resolutions supported by DOSBox and how this change affects them. Rows highlighted with yellow are the square pixel modes (where the storage aspect ratio (SAR) equals 4/3 = 1.3333), then 1280x1024 is a special outlier that needs special handling to force the 4:3 stretch.

I've tested all (S)VGA and VESA modes in QPV (`vesa_modes = all` & `vmemsize = 8`), so all 15/16-bit hi-colour and 24/32-bit true colour modes on `svga_s3`. This test should be good enough as the fix is very generic as you can see.

image

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1412,26 +1412,6 @@ static bool is_high_resolution(const video_mode_block_iterator_t mode_block)
 	return is_high_resolution(mode_block->swidth, mode_block->sheight);
 }
 
-static void maybe_aspect_correct_tall_modes(double &current_ratio)
-{
-	// If we're in a mode that's wider than it is tall, then do nothing
-	if (CurMode->swidth >= CurMode->sheight || current_ratio > 0.9)
-		return;
-
-	// We're in a tall mode
-
-	if (!render.aspect) {
-		LOG_INFO("VGA: Tall resolution (%ux%u) may appear stretched wide, per 'aspect = false'",
-	         CurMode->swidth, CurMode->sheight);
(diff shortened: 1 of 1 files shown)


kcgen (approved):
Tested on my side too - and all is good!

Excellent to see more specific handling eliminated :partying_face: 

This should be the basis for the other aspect regressions and the screenshot PR - so getting this merged.

PR #2584, opened 2023-06-08

Kappa971:
As already said (I don't remember where), `'cputype'` should also be redone, both how it works and the description.
For example I would expect that by setting `cputype = pentium_slow` (what cpu would that be? 😆) and `cycles = auto,` the cycles would be automatically set based on the speed of the chosen cpu, but DOSBox does none of this (I don't even know what `'cputype'` is for 😅).


kcgen:
Re: CPU model and speed settings ("486 33 Mhz");  - I can't find the ticket for that, but yes, it's come up a couple times. Please open a new one if you also can't find a match!

Thanks for the translation @Kappa971 ; merging.

Issue #2586, opened 2023-06-09
'integer_scaling=horizontal' text is cut-off with minimal window size


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-alpha-879

### What Operating System are you using?

Mac OS

### Relevant hardware info

M2 mini

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When setting 'config -set integer_scaling=horizontal' and resizing the window to the smallest size possible, the text is cut off, see screenshot.

![Bildschirmfoto 2023-06-09 um 18 16 02](https://github.com/dosbox-staging/dosbox-staging/assets/61995128/6a830fc5-b172-4525-872c-ca3b899ec8e9)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. 
2.
3.

### Your configuration

```ini
-noprimaryconf
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
I guess the only 100% fix for this is to set the min window bounds dynamically based on the actual output dimensions and/or resize the window too automatically.

E.g. some VESA 132-column text modes appear almost widescreen and they have 1056×352/400/480 resolutions. 

So imagine displaying 640x480 gfx with min window size which is 640x480, then switching to 1056x480 text mode with aspect ratio correction turned off... Major clipping!

kcgen:
feels like a bug; the SDL/OpenGL canvas size and shape are not controlled or "pressed outward" by large DOS draw sizes. 

For example, here's a 1600x1200 DOS dimension rendered into the users desired (and set) window size:

 
![2023-06-09_16-16](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/d48726cc-7e00-405c-b271-27b7b2e4378f)

> resize the window too automatically.

We really want to avoid that - earlier versions adjusted the window size based on the DOS size, and it was a big source of problems (user feedback *"If I sent a window size, that's what I want, please don't pull the rug out from me"*), and worse, SDL doesn't differentiate between programmatically driven window changes versus "user GUI" changes -> and so all the usual SDL event callbacks fire, and this can cascade.  This lead to lots of code to disengage callbacks to prevent "on-change" feedback loops.

johnnovak:
[...]

PR #2588, opened 2023-06-10
Auto-disable the integer scaling feature if the image is smaller than…

johnnovak:
… the viewport

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2586

@Burrito78 ☝🏻

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -649,9 +649,11 @@ void DOSBOX_Init()
 
 	pstring = secprop->Add_string("integer_scaling", always, "off");
 	pstring->Set_help(
-	        "Constrain the horizontal or vertical scaling factor to integer values. The\n"
-	        "correct aspect ratio is always maintained, which may result in a non-integer\n"
-	        "scaling factor to be applied in the other dimension.\n"
+	        "Constrain the horizontal or vertical scaling factor to integer values.\n"
+	        "The correct aspect ratio is always maintained, which may result in a\n"
+	        "non-integer scaling factor to be applied in the other dimension. If the\n"
+	        "image is larger than the viewport, the integer scaling constraint is\n"
+	        "auto-disabled (same as 'off').\n"
 	        "  off:         No integer scaling constraint; the aspect ratio correct image\n"
 	        "               fills the viewport (default).\n"
 	        "  horizontal:  Constrain the horizontal scaling factor to integer values\n"
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3203,28 +3203,35 @@ static SDL_Rect calc_viewport(const int win_w, const int win_h)
 	assert(std::isfinite(sdl.draw.scalex));
 	assert(std::isfinite(sdl.draw.scaley));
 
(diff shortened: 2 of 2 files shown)


Burrito78:
Does it make sense to log it that integer scaling is off because window size is too small?

johnnovak:
> Does it make sense to log it that integer scaling is off because window size is too small?

I don't think so.

kcgen (approved):
Very nice update, @johnnovak!

kcgen:
Will hold for @GranMinigun to checkout as well.

johnnovak:
> Will hold for @GranMinigun to checkout as well.

Sure, makes sense.

src/dosbox.cpp
```diff
@@ -649,9 +649,11 @@ void DOSBOX_Init()
...
 	pstring = secprop->Add_string("integer_scaling", always, "off");
 	pstring->Set_help(
-	        "Constrain the horizontal or vertical scaling factor to integer values. The\n"
-	        "correct aspect ratio is always maintained, which may result in a non-integer\n"
-	        "scaling factor to be applied in the other dimension.\n"
+	        "Constrain the horizontal or vertical scaling factor to integer values.\n"
+	        "The correct aspect ratio is always maintained, which may result in a\n"
+	        "non-integer scaling factor to be applied in the other dimension. If the\n"
+	        "image is larger than the viewport, the integer scaling constraint is\n"
+	        "auto-disabled (same as'off').\n"
```
GranMinigun, line 656: Missed space in "same as 'off'".

src/gui/sdlmain.cpp
```diff
@@ -3243,6 +3254,10 @@ static SDL_Rect calc_viewport(const int win_w, const int win_h)
 		// Calculate the final viewport.
 		view_w = sdl.draw.width * integer_scale_factor;
 		view_h = iround(sdl.draw.height * integer_scale_factor * pixel_aspect);
+
+		if (!is_within_bounds(view_w, view_h)) {
+			std::tie(view_w, view_h) = calculate_bounded_dims();
+		}
```
GranMinigun, line 3260: `integer_scale_factor < 1` right above would indicate that the image doesn't fit into window, so this can be moved there instead.

src/gui/sdlmain.cpp
```diff
@@ -3256,6 +3271,10 @@ static SDL_Rect calc_viewport(const int win_w, const int win_h)
 
 		view_w = iround(sdl.draw.height * integer_scale_factor * output_aspect);
 		view_h = sdl.draw.height * integer_scale_factor;
+
+		if (!is_within_bounds(view_w, view_h)) {
+			std::tie(view_w, view_h) = calculate_bounded_dims();
+		}
```
GranMinigun, line 3277: Ditto.

src/gui/sdlmain.cpp
johnnovak: Yeah ok, I can do that, plus add some comments.

GranMinigun (approved):

johnnovak:
Thanks for the review @GranMinigun

Issue #2484, opened 2023-05-16
Small improvements to 'mixer /listmidi'


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Mac OS

### Is your feature request related to a problem? Please describe.

When MT-32 is active, the current active ROM isn't displayed in green and so it's not immediately clear whats going on.
It's functional but could be better.

Also the paths to the SF2 soundfonts are too long and therefore cut off.

![Bildschirmfoto 2023-05-15 um 14 50 19](https://github.com/dosbox-staging/dosbox-staging/assets/61995128/1ef8c251-b781-431a-b077-4b4d0f21e976)

Grounded0:
Highlighting active MT-32/CM-32 ROM makes sense and behaves similarly how we highlight active Fluidsynth soundfont.

johnnovak:
This has been bothering me too. I have a branch that highlights the active MT-32 romset in green, like it used to earlier. I just need to finish it sometime...

Grounded0:
@Burrito78 You can write your Fluidsynth directory as ~/Library/Preferences/DOSBox... so the whole path name + file name fits into the text field.

Burrito78:
#3011 fixed these issues, closing.

PR #2591, opened 2023-06-11
Load MT-32 ROMs from `romdir` using hash matches 

weirddan455:
Hi, first PR to this project.  I ran into an issue getting MT32 emulation working on my system and realized it was due to case sensitivity.  The names of the ROMs are hardcoded here:

https://github.com/dosbox-staging/dosbox-staging/blob/b8f72a08a2c87541314ef1e02ec2ff7ad16fce70/src/midi/midi_mt32.cpp#L74-L98

I figured it might be useful to search for these files in a case insensitive manner instead so that's what this patch does.  It would have saved me, as a user, a bit of time diagnosing this problem in any case.

### DIFF
--- a/src/midi/midi_lasynth_model.cpp
+++ b/src/midi/midi_lasynth_model.cpp
@@ -23,9 +23,9 @@
 #if C_MT32EMU
 
 #include <cassert>
+#include <unordered_set>
 
 #include "fs_utils.h"
-#include "string_utils.h"
 
 // Construct a new model and ensure both PCM and control ROM(s) are provided
 LASynthModel::LASynthModel(const std::string &rom_name,
@@ -54,6 +54,7 @@ std::optional<std_fs::path> LASynthModel::find_rom(const service_t& service,
                                                    const std_fs::path& dir,
                                                    const Rom* rom)
 {
+	static std::unordered_set<std::string> unknown_files;
 	if (!rom) {
 		return {};
 	}
@@ -64,12 +65,18 @@ std::optional<std_fs::path> LASynthModel::find_rom(const service_t& service,
 			continue;
 		}
 
+		const std::string filename = std_fs::canonical(entry.path(), ec).string();
+		if (ec) {
+			continue;
+		}
 		mt32emu_rom_info info;
-		const std::string filename = entry.path().string();
 		if (service->identifyROMFile(&info,
 		                             filename.c_str(),
 		                             nullptr) != MT32EMU_RC_OK) {
-			LOG_WARNING("MT32: Unknown file in ROM folder: %s", filename.c_str());
+			// Only log unknwon files one time (if not already in the unknown_files set).
+			if (unknown_files.insert(filename).second) {
+				LOG_WARNING("MT32: Unknown file in ROM folder: %s", filename.c_str());
+			}
 			continue;
 		}
 
--- a/src/midi/midi_lasynth_model.cpp
+++ b/src/midi/midi_lasynth_model.cpp
@@ -25,6 +25,7 @@
 #include <cassert>
 
 #include "fs_utils.h"
+#include "string_utils.h"
 
 // Construct a new model and ensure both PCM and control ROM(s) are provided
 LASynthModel::LASynthModel(const std::string &rom_name,
@@ -49,65 +50,88 @@ LASynthModel::LASynthModel(const std::string &rom_name,
 	assert(ctrl_full || (ctrl_a && ctrl_b));
 }
 
+std::optional<std_fs::path> LASynthModel::find_rom(const service_t& service,
+                                                   const std_fs::path& dir,
+                                                   const Rom* rom)
+{
+	if (!rom) {
+		return {};
+	}
+
+	std::error_code ec;
+	for (const auto& entry : std_fs::directory_iterator(dir, ec)) {
+		if (ec) {
+			continue;
+		}
+
+		mt32emu_rom_info info;
+		const std::string filename = entry.path().string();
+		if (service->identifyROMFile(&info,
+		                             filename.c_str(),
+		                             nullptr) != MT32EMU_RC_OK) {
+			LOG_WARNING("MT32: Unknown file in ROM folder: %s", filename.c_str());
+			continue;
+		}
+
+		const char* rom_id = NULL;
+		if (rom->type == ROM_TYPE::PCM) {
+			rom_id = info.pcm_rom_id;
+		} else if (rom->type == ROM_TYPE::CONTROL) {
+			rom_id = info.control_rom_id;
+		}
+
+		if (rom_id && rom->id == rom_id) {
+			return entry.path();
+		}
+	}
+	return {};
+}
+
 // Checks if its ROMs can be positively found in the provided directory
 bool LASynthModel::InDir(const service_t& service, const std_fs::path& dir) const
 {
 	assert(service);
 
-	auto find_rom = [&service, &dir](const Rom *rom) -> bool {
-		if (!rom)
-			return false;
-
-		const auto rom_path = dir / rom->filename;
-		if (!std_fs::exists(rom_path))
-			return false;
-
-		mt32emu_rom_info info;
-		if (service->identifyROMFile(&info, rom_path.string().c_str(), nullptr) !=
-		    MT32EMU_RC_OK)
-			return false;
-
-		if (rom->type == ROM_TYPE::UNVERSIONED)
-			return true;
-
-		const bool found_pcm = info.pcm_rom_id &&
-		                       (rom->id == info.pcm_rom_id);
-		const bool found_ctrl = info.control_rom_id &&
-		                        (rom->id == info.control_rom_id);
-		return found_pcm || found_ctrl;
-	};
-
-	const bool have_pcm = find_rom(pcm_full) ||
-	                      (find_rom(pcm_l) && find_rom(pcm_h));
-	const bool have_ctrl = find_rom(ctrl_full) ||
-	                       (find_rom(ctrl_a) && find_rom(ctrl_b));
+	const bool have_pcm = find_rom(service, dir, pcm_full) ||
+	                      (find_rom(service, dir, pcm_l) &&
+	                       find_rom(service, dir, pcm_h));
+	const bool have_ctrl = find_rom(service, dir, ctrl_full) ||
+	                       (find_rom(service, dir, ctrl_a) &&
+	                        find_rom(service, dir, ctrl_b));
 	return have_pcm && have_ctrl;
 }
 
 // If present, loads either the full or partial ROMs from the provided directory
 bool LASynthModel::Load(const service_t& service, const std_fs::path& dir) const
 {
-	if (!service || !InDir(service, dir))
+	if (!service) {
 		return false;
+	}
 
-	auto load_rom = [&service, &dir](const Rom *rom_full,
+	auto load_rom = [&service, &dir](const Rom* rom_full,
 	                                 mt32emu_return_code expected_code) -> bool {
-		if (!rom_full)
+		const auto rom_path = find_rom(service, dir, rom_full);
+		if (!rom_path) {
 			return false;
-		const auto rom_path = dir / rom_full->filename;
-		const auto rcode = service->addROMFile(rom_path.string().c_str());
+		}
+		const auto rcode = service->addROMFile(rom_path->string().c_str());
 		return rcode == expected_code;
 	};
 
-	auto load_both = [&service, &dir](const Rom *rom_1, const Rom *rom_2,
+	auto load_both = [&service, &dir](const Rom* rom_1,
+	                                  const Rom* rom_2,
 	                                  mt32emu_return_code expected_code) -> bool {
-		if (!rom_1 || !rom_2)
+		const auto rom_1_path = find_rom(service, dir, rom_1);
+		if (!rom_1_path) {
 			return false;
-		const auto rom_1_path = dir / rom_1->filename;
-		const auto rom_2_path = dir / rom_2->filename;
+		}
+		const auto rom_2_path = find_rom(service, dir, rom_2);
+		if (!rom_2_path) {
+			return false;
+		}
 
 		const auto rcode = service->mergeAndAddROMFiles(
-		        rom_1_path.string().c_str(), rom_2_path.string().c_str());
+		        rom_1_path->string().c_str(), rom_2_path->string().c_str());
 		return rcode == expected_code;
 	};
 
--- a/src/midi/midi_lasynth_model.h
+++ b/src/midi/midi_lasynth_model.h
@@ -26,6 +26,7 @@
 #if C_MT32EMU
 
 #include <memory>
+#include <optional>
 #include <string>
 
 #include "std_filesystem.h"
@@ -37,11 +38,10 @@
 // form.
 class LASynthModel {
 public:
-	enum class ROM_TYPE { UNVERSIONED, VERSIONED };
+	enum class ROM_TYPE { PCM, CONTROL };
 
 	struct Rom {
 		const std::string id;
-		const std::string filename;
 		const ROM_TYPE type;
 	};
 
@@ -75,6 +75,9 @@ public:
 
 private:
 	size_t SetVersion();
+	static std::optional<std_fs::path> find_rom(const service_t& service,
+	                                            const std_fs::path& dir,
+	                                            const Rom* rom);
 
 	const std::string name = {};
 	const size_t version_pos = std::string::npos;
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -68,71 +68,81 @@ constexpr bool USE_NICE_PANNING = true;
 constexpr bool USE_NICE_PARTIAL_MIXING = false;
 
 using Rom = LASynthModel::Rom;
-constexpr auto versioned = LASynthModel::ROM_TYPE::VERSIONED;
-constexpr auto unversioned = LASynthModel::ROM_TYPE::UNVERSIONED;
 
-// Traditional ROMs
-const Rom mt32_pcm_any_f = {"pcm_mt32", "MT32_PCM.ROM", unversioned};
-const Rom mt32_ctrl_any_f = {"ctrl_mt32", "MT32_CONTROL.ROM", unversioned};
-const Rom cm32l_pcm_any_f = {"pcm_cm32l", "CM32L_PCM.ROM", unversioned};
-const Rom cm32l_ctrl_any_f = {"ctrl_cm32l", "CM32L_CONTROL.ROM", unversioned};
-
-// MAME ROMs (versioned)
-const Rom mt32_pcm_100_f = {"pcm_mt32", "r15449121.ic37.bin", versioned};
-const Rom mt32_pcm_100_l = {"pcm_mt32_l", "r15179844.ic21.bin", versioned};
-const Rom mt32_pcm_100_h = {"pcm_mt32_h", "r15179845.ic22.bin", versioned};
-const Rom mt32_ctrl_104_a = {"ctrl_mt32_1_04_a", "mt32_1.0.4.ic27.bin", versioned};
-const Rom mt32_ctrl_104_b = {"ctrl_mt32_1_04_b", "mt32_1.0.4.ic26.bin", versioned};
-const Rom mt32_ctrl_105_a = {"ctrl_mt32_1_05_a", "mt32_1.0.5.ic27.bin", versioned};
-const Rom mt32_ctrl_105_b = {"ctrl_mt32_1_05_b", "mt32_1.0.5.ic26.bin", versioned};
-const Rom mt32_ctrl_106_a = {"ctrl_mt32_1_06_a", "mt32_1.0.6.ic27.bin", versioned};
-const Rom mt32_ctrl_106_b = {"ctrl_mt32_1_06_b", "mt32_1.0.6.ic26.bin", versioned};
-const Rom mt32_ctrl_107_a = {"ctrl_mt32_1_07_a", "mt32_1.0.7.ic27.bin", versioned};
-const Rom mt32_ctrl_107_b = {"ctrl_mt32_1_07_b", "mt32_1.0.7.ic26.bin", versioned};
-const Rom mt32_ctrl_bluer_a = {"ctrl_mt32_bluer_a", "blue_ridge__mt32a.bin", versioned};
-const Rom mt32_ctrl_bluer_b = {"ctrl_mt32_bluer_b", "blue_ridge__mt32b.bin", versioned};
-const Rom mt32_ctrl_204_f = {"ctrl_mt32_2_04", "mt32_2.0.4.ic28.bin", versioned};
-const Rom cm32l_ctrl_100_f = {"ctrl_cm32l_1_00", "lapc-i.v1.0.0.ic3.bin", versioned};
-const Rom cm32l_ctrl_102_f = {"ctrl_cm32l_1_02", "cm32l_control.rom", versioned};
-const Rom cm32l_pcm_100_h = {"pcm_cm32l_h", "r15179945.ic8.bin", versioned};
-const Rom &cm32l_pcm_100_l = mt32_pcm_100_f; // Lower half of samples comes from MT-32
+const Rom mt32_pcm_100_f  = {"pcm_mt32", LASynthModel::ROM_TYPE::PCM};
+const Rom mt32_pcm_100_l  = {"pcm_mt32_l", LASynthModel::ROM_TYPE::PCM};
+const Rom mt32_pcm_100_h  = {"pcm_mt32_h", LASynthModel::ROM_TYPE::PCM};
+const Rom mt32_ctrl_104_f = {"ctrl_mt32_1_04", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_104_a = {"ctrl_mt32_1_04_a", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_104_b = {"ctrl_mt32_1_04_b", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_105_f = {"ctrl_mt32_1_05", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_105_a = {"ctrl_mt32_1_05_a", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_105_b = {"ctrl_mt32_1_05_b", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_106_f = {"ctrl_mt32_1_06", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_106_a = {"ctrl_mt32_1_06_a", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_106_b = {"ctrl_mt32_1_06_b", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_107_f = {"ctrl_mt32_1_07", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_107_a = {"ctrl_mt32_1_07_a", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_107_b = {"ctrl_mt32_1_07_b", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_bluer_f = {"ctrl_mt32_bluer", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_bluer_a = {"ctrl_mt32_bluer_a", LASynthModel::ROM_TYPE::CONTROL};
+const Rom mt32_ctrl_bluer_b = {"ctrl_mt32_bluer_b", LASynthModel::ROM_TYPE::CONTROL};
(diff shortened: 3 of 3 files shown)


src/midi/midi_lasynth_model.cpp
```diff
@@ -94,20 +112,26 @@ bool LASynthModel::Load(const service_t& service, const std_fs::path& dir) const
...
 		if (!rom_1 || !rom_2)
 			return false;
-		const auto rom_1_path = dir / rom_1->filename;
-		const auto rom_2_path = dir / rom_2->filename;
+		const auto rom_1_path = GetROMPath(dir, rom_1->filename);
+		if (!rom_1_path)
+			return false;
+		const auto rom_2_path = GetROMPath(dir, rom_2->filename);
+		if (!rom_2_path)
+			return false;
```
kcgen, line 131: Thanks!

Do you have `clang-format` in your path? 

If you can run `./scripts/format-commit.sh`, it will update the formatting of just the changed lines.

src/midi/midi_lasynth_model.cpp
```diff
@@ -49,6 +50,23 @@ LASynthModel::LASynthModel(const std::string &rom_name,
 	assert(ctrl_full || (ctrl_a && ctrl_b));
 }
 
+// Checks the directory for the ROM name in a case insensitive manner.
+static std::optional<std_fs::path> GetROMPath(const std_fs::path& dir, const std::string& filename)
+{
+	std::string filename_lower = filename;
+	lowcase(filename_lower);
+	try {
+		for (const auto& entry : std_fs::directory_iterator(dir)) {
```
kcgen, line 59: The try will bail out on the first  problem (maybe a broken symlink), and we won't be able to drill in further to get the files. 
Suggest using the error-checking versions of all the filesystem calls here to continue past problems (and avoid exception handling entirely).;

kcgen: Also, the function is very generic and useful for other file lookups (perhaps SF2 and CDROM image files). 

Suggest adding it to fs_utils

src/midi/midi_lasynth_model.cpp
weirddan455: Yes, I  just ran that script and pushed a new commit with fixed formatting.

src/midi/midi_lasynth_model.cpp
weirddan455: I just moved the function to fs_utils as you suggested.  I switched to the error code version of `directory_iterator` which I believe is the only function that can throw an exception.  Let me know if that looks right to you.  I don't quite know all the ins and outs of this C++ 17 filesystem API.

src/misc/fs_utils.cpp
```diff
@@ -143,3 +144,21 @@ std::time_t to_time_t(const std_fs::file_time_type &fs_time)
...
+// Searches the directory for the filename. Returns the first file that matches
+// the filename, regardless of casing.
+std::optional<std_fs::path> case_insensitive_search(const std_fs::path& dir,
+                                                    const std::string& filename)
+{
+	std::string filename_lower = filename;
+	lowcase(filename_lower);
+	std::error_code ec;
+	for (const auto& entry : std_fs::directory_iterator(dir, ec)) {
+		std::string itr_lower = entry.path().filename().string();
```
kcgen, line 157: The error code is set true if the iterator has a problem; suggest add a check and continue first thing in the for loop:

``` c++
if (ec) // Skip invalid or broken content {
    continue;
}
```

(all std::filesystem calls clear the error_code first thing, so we don't have to reset/clear it)

src/misc/fs_utils.cpp
```diff
@@ -143,3 +144,21 @@ std::time_t to_time_t(const std_fs::file_time_type &fs_time)
...
+		lowcase(itr_lower);
+		if (filename_lower.compare(itr_lower) == 0) {
```
kcgen, line 159: suggest direct comparison `(filename_lower == itr_lower)`

kcgen:
looks good, @weirddan455.  Just a couple suggestions, and then if you can squash on your side to reduce the intermediate commits, then that should be it!  Thanks again :rocket:

src/misc/fs_utils.cpp
weirddan455: Ah, OK.  I suppose I was mistaken in thinking the iterator could only error on construction rather than every iteration.  Added that check.

src/misc/fs_utils.cpp
weirddan455: Done

weirddan455:
Just did a force push with your new suggestions + squashing everything into one commit.  Thanks for the fast review :)

FeralChild64 (commented):
Looks OK to me.

kcgen:
Looks great, @weirddan455 ! 

It just hit me that having unit tests for this new insensitive directory search would be helpful, too.

We use very basic tests; if you can look into the tests/ directory, checkout `tests_fs_utils.cpp`. You can see how other if it's functions are tested; they lookup placeholder files in the `tests/files/` directory, which you can also use (or add to).

If you're on linux:

``` shell
meson setup -Dbuildtype=debug build/debug
meson compile -C build/debug
```

Run just the `fs_utils` tests:

``` shell
./build/debug/tests/fs_utils
```


Or run all the tests (for all modules):

``` shell
meson test -C build/debug
```

src/misc/fs_utils.cpp
kcgen: indeed; that aspect isn't very intuitive (that it re-applies `ec` for each iteration).

weirddan455:
@kcgen Sure I'll take a look at that in a bit.  I'm thinking if we need something reproducible like a unit test, the function might need to sort the directory contents alphabetically and return the first result.  Right now it's arbitrary what gets returned if the directory contains files with the same names but different casing.

I see this function does something similar (except with the POSIX glob function):

https://github.com/dosbox-staging/dosbox-staging/blob/b8f72a08a2c87541314ef1e02ec2ff7ad16fce70/include/fs_utils.h#L53-L70

kcgen:
I think it's OK to keep this new function std::filesystem 'clean', which I think it is, right now.

The filesystem::path are very robust, and so combining a path + filename results in a logical object that can be tested, as opposed to just string manipulation (where things like 'testfile' != './testfile').

Perhaps for this function, we'll need to use filesystem's canonical handling, that can bake down nonsense like ./././testfile just to testfile. (So perhaps in your unit tests, it will require the filename portion to be canonicalized too). 

Regarding alphabetic sorting; it sounds reasonable, but when I think it through, for example:

MyFiLE
mYfILE
MYfiLE

... Well, we're already in random land where case doesn't matter. So sorting is trying to apply casing rules into a space where casing has no meaning.

User A might complain that it should have picked his lowercase file first, but user B might want her uppercase file instead.  Both are invalid.. these users need to cleanup their ambiguous messes. 
---

Back to the mt32 use case, we do know that if users have files 'as-is' (with casing straight from the zip files), then we want to always search for those expected filenames first, and then fallback to case insensitive. 

So perhaps we can keep the direct-hit first (apologies if that's already in place but I just didn't notice😅)

weirddan455:
@johnnovak I really like that solution.  I just modified the PR to do away with hard coded filenames.  The mt32emu library we are using has a method `identifyROMFile` that internally does a SHA1 hash + compare to a known list of ROMs.

https://github.com/munt/munt/blob/813e8f0eafbfc1771a9d7704e7025915af89643e/mt32emu/src/ROMInfo.cpp#L56-L87

I also removed the idea of "unversioned ROMs" and moved everything so it has a 1 to 1 link with what mt32emu is doing.

Could still use caching in persistent storage, as you suggested.  I put that a TODO comment in there for now.

johnnovak:
This is great @weirddan455! I'm glad you're tackling this; I like your enthusiasm! 🎉 

> @johnnovak I really like that solution. I just modified the PR to do away with hard coded filenames. The mt32emu library we are using has a method `identifyROMFile` that internally does a SHA1 hash + compare to a known list of ROMs.

💯 🥇 

> I also removed the idea of "unversioned ROMs" and moved everything so it has a 1 to 1 link with what mt32emu is doing.

I think we need to leave that in for compatibility reasons. Other DOSBox forks use the "unversioned" ROMs if they are in a set location or in the local game folder (I think, but don't quote me on the exact lookup behaviour). So basically we only want to keep this for compatibility with existing gamepacks that have whatever MT-32 ROM versions bundled with the game.

But yes, this is just for compatibility; when configuring a game from scratch, we discourage the use of non-versioned ROMs, absolutely.

> Could still use caching in persistent storage, as you suggested. I put that a TODO comment in there for now.

Actually, thinking about this more deeply, it's hard to detect if the contents of a folder have changed in a cross-platform manner. So if we can get away with no caching, that's always the better and more robust solution.

Could you perhaps measure (e.g. log the time taken) how long it takes to scan the full MT-32 ROM set at startup? I'm pretty sure the limiting factor here is disk I/O; even old hardware can calculate SHA1 hashes at faster rates than reading from disk. Ideally, if this only adds a few hundred milliseconds at most, then we most definitely don't need caching.

kcgen:
For the scanning, the catch will be if a user has a couple SF2 files in there as well (which might be 500+ MB).  On a desktop w/ SSD, their startup time might still be quick :-)  But we could _probably_ glob `*.rom`, case-insensitive, as a very coarse filter.

johnnovak:
> For the scanning, the catch will be if a user has a couple SF2 files in there as well (which might be 500+ MB). On a desktop w/ SSD, their startup time might still be quick :-) But we could _probably_ glob `*.rom`, case-insensitive, as a very coarse filter.
> 
> We could also log a warning for files it scans that have nothing to do w/ mt32emu (when mt32emu return a failure core). This would encourage the user to maintain a very clean `mt32emu-roms` directory, and nothing more.

Yeah, I was gonna say, people shouldn't put all sorts of crap into the `mt32emu-roms` folder 😄

I propose a slightly different coarse filtering: we know for certain the expected filesizes of the valid ROM files. So let's add a whitelist of valid filesizes and only check those. The side-effect is that even if some random other file has the same filesize (extremely unlikely), calculating the hash for it won't be a problem as the ROM files are all rather small.

We could still log a warning for all the various random crap we find in the MT32 ROM folder, sure.

weirddan455:
>  I think we need to leave that in for compatibility reasons. Other DOSBox forks use the "unversioned" ROMs if they are in a set location or in the local game folder

I believe this should still work with my patch.  "unversioned" ROMs were really just versioned ROMs with specific filenames which our code excluded from version checks.  However, they still went through the exact same path with regards to mt32emu so I believe they will still work now (and further more now be recognized as the real version that they are).  Of course, testing would be welcome to know for sure.

>  Could you perhaps measure (e.g. log the time taken) how long it takes to scan the full MT-32 ROM set at startup?

Sure.  Is there a particular benchmarking or profiling tool you would recommend for this?

>  I propose a slightly different coarse filtering: we know for certain the expected filesizes of the valid ROM files. So let's add a whitelist of valid filesizes and only calculate the SHA1 hash for those.

I had a look through the mt32emu source code as I was writing this and this is already done.  Filesize gets matched before the SHA gets calculated:

Whitelist:
https://github.com/munt/munt/blob/813e8f0eafbfc1771a9d7704e7025915af89643e/mt32emu/src/ROMInfo.cpp#L56-L87

Filesize + SHA compare
https://github.com/munt/munt/blob/813e8f0eafbfc1771a9d7704e7025915af89643e/mt32emu/src/ROMInfo.cpp#L176-L185

johnnovak:
> However, they still went through the exact same path with regards to mt32emu so I believe they will still work now (and further more now be recognized as the real version that they are). Of course, testing would be welcome to know for sure.

You can easily test this yourself by getting hold of some of the "unversioned" ROMs. I'm a bit busy, but I can send you those files later if you're on Discord.

> > Could you perhaps measure (e.g. log the time taken) how long it takes to scan the full MT-32 ROM set at startup?
> 
> Sure. Is there a particular benchmarking or profiling tool you would recommend for this?

No, just measure delta time in the code and log the results.

> > I propose a slightly different coarse filtering: we know for certain the expected filesizes of the valid ROM files. So let's add a whitelist of valid filesizes and only calculate the SHA1 hash for those.
> 
> I had a look through the mt32emu source code as I was writing this and this is already done. Filesize gets matched before the SHA gets calculated:
> 
> Whitelist: https://github.com/munt/munt/blob/813e8f0eafbfc1771a9d7704e7025915af89643e/mt32emu/src/ROMInfo.cpp#L56-L87
> 
> Filesize + SHA compare https://github.com/munt/munt/blob/813e8f0eafbfc1771a9d7704e7025915af89643e/mt32emu/src/ROMInfo.cpp#L176-L18

Nice, these guys have thought of everything! We only need to use these functions then! 🚀

weirddan455:
So it's pretty fast on my system.  I wrapped this `load_model` function with `clock_gettime(CLOCK_MONOTONIC)` to get  a quick benchmark with my full ROM set.  Here's some numbers:

67ms - Debug build pre-patch
143ms - Debug build post-patch

11ms - Release build pre-patch
22ms - Release build post-patch

As far as the "unversioned" ROMs, I can confirm the following 2 sets are working (after the latest commit where I found one was missing).

CM32L "unversioned"
a439fbb390da38cada95a7cbb1d6ca199cd66ef8  cm32l_control.rom
289cc298ad532b702461bfc738009d9ebe8025ea  cm32l_pcm.rom

MT32 "unversioned"
b083518fffb7f66b03c23b7eb4f868e62dc5a987  mt32_control.rom 
f6b1eebc4b2d200ec6d3d21d51325d5b48c60252  mt32_pcm.rom

Lastly, the code is already calling the `identifyROMFile` function in mt32emu which in turn calls the functions with the optimizations I mentioned above if that wasn't clear.

kcgen:
@sergm , nice to see @weirddan455 leveraging your features even more directly in this PR! 

This will open up access to all the models currently supported by mt32emu:

| Model | Version | Staging support 
| --- | --- | ---
| MT-32 'old' | 1.04 | 0.80.1
| MT-32 'old' | 1.05 | 0.80.1
| MT-32 'old' | 1.06 | 0.80.1
| MT-32 'old' | 1.07 | 0.80.1
| MT-32 'new' | 2.03 | 0.80.1
| MT-32 'new' | 2.04 | 0.80.1
| MT-32 'new' | 2.06 | PR (0.81-alpha)
| MT-32 'new' | 2.07 | PR (0.81-alpha)
| CM-32L | 1.00 | 0.80.1
| CM-32L | 1.02 | 0.80.1
| [CM-32LN](https://www.vogons.org/viewtopic.php?t=55345) | 1.00 | PR (0.81-alpha) |

weirddan455:
@kcgen Yes, that is the idea.  I haven’t added all of those yet.  I believe 2.06 and 2.07 are missing.  Those probably need new config options as well since they were never supported by us.

kcgen:
> I believe 2.06 and 2.07 are missing. Those probably need new config options as well since they were never supported by us.

Contacted @sergm, who might be able to help in this regard. I will follow up via email.

for general ref, I believe this is where active MT-32/CM32 version and ROM status is being held: https://en.wikipedia.org/wiki/User:Lord_Nightmare

johnnovak:
Those are great results @weirddan455 regarding the timings. I suspect the detection of the full ROM-set won't take too long even on underpowered devices then such as the Raspberry Pi4.

I have uploaded a while ago the full MT-32 ROM set to archieve.org along with instructions how to create the "unversioned" files out of them, if that helps with testing. I will add 2.06 and 2.07 to it once we get hold of them.

https://archive.org/details/Roland-MT-32-ROMs

We also need to update the output of the MIXER command for the new ROMs, but let's treat that separately:

image

We can simply move the path to the current MT32 ROM below the little table that lists the found MT32 versions. That would be an improvement for other reasons too because then the ROM path could use the full width of the screen.

johnnovak:
By the they way, the output of `MIXER /LISTMIDI` is not hardcoded to 80-columns; it takes full advantage of the full available horizontal area. Whoever wrote it did a good job! 🏅 This is how it looks like in 132-column VESA modes:

src/midi/midi_lasynth_model.cpp
```diff
@@ -49,65 +50,91 @@ LASynthModel::LASynthModel(const std::string &rom_name,
...
 }
 
-// Checks if its ROMs can be positively found in the provided directory
-bool LASynthModel::InDir(const service_t& service, const std_fs::path& dir) const
+// TODO: Possibly add caching to this function in persistant storage (perhaps a
+// file in ~/.config/dosbox) The identifyROMFile method in the mt32emu library
+// compares with known SHA1 sums so it may be slow for users with lots of files
+// to search. This is mitaged by checking the file size first at least, only
+// running a checksum on files of the correct size.
+std::optional<std_fs::path> LASynthModel::find_rom(const service_t& service,
```
johnnovak, line 58: Given the timings are really good, I think we can remove this TODO comment as it's fast enough to detect all the ROMs.

src/midi/midi_lasynth_model.cpp
```diff
@@ -49,65 +50,91 @@ LASynthModel::LASynthModel(const std::string &rom_name,
...
-		if (service->identifyROMFile(&info, rom_path.string().c_str(), nullptr) !=
-		    MT32EMU_RC_OK)
-			return false;
-
-		if (rom->type == ROM_TYPE::UNVERSIONED)
-			return true;
+		if (service->identifyROMFile(&info,
+		                             entry.path().string().c_str(),
+		                             nullptr) != MT32EMU_RC_OK) {
+			continue;
```
johnnovak, line 76: I think we should log a warning for every single file that could not be identified, so then the user can remove them. If they dumped all sorts of random files into the ROM folder, this sudden influx of warnings will be very educational and will hopefully motivate them to keep their ROM folders clean 😎 🤘🏻

kcgen:
> @kcgen Yes, that is the idea. I haven’t added all of those yet. I believe 2.06 and 2.07 are missing. Those probably need new config options as well since they were never supported by us.

Partially resolved, messaged in Discord.

weirddan455:
I pushed a commit yesterday for the items @johnnovak pointed out (added logging of unknown files and removed the TODO comment).

I just added support for the rest of the ROMs supported by mt32emu.  Thanks to @kcgen for providing the 2.07 ROM to test with.

The mixer command now also lists the additional models.  I believe @kcgen wrote the mt32 `ListAll` interface for that.  I just added the new models to an array and it worked :smile:

kcgen:
Tested it and it's working great! The PR is also surprisingly small modification for such a big change in the loading functionality!

For the `/listmidi` output, it was previously:



Notice how the unversioned model is followed by it's versioned values.

So users can type:
- `model = cml32l`, and they get the first available CM-32L from left-to-right (priority-order), or
- `model = cm32l_102`, and get a specific version.

That's actually how the PR works too :-) However, visually we've lost the model names from the columns, so now the versions are ambiguous:



So I think we need to put back the model prefix names (with cm32ln, now too), this is going to make the table quite wide, so after this, we probably need to rethink a new way to present the information (can be a follow on PR!).

For the error logging, I get a lot of repetition when changing the model.
try:

```shell
Z:\> model mt32_207
Z:\> mididevice mt32
```

```txt
2023-06-15 06:03:10.458 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.462 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.463 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.466 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.468 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.472 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.473 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.475 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.478 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.484 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.488 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.489 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.492 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.495 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.499 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.500 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.502 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.505 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.509 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.513 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.514 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.518 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.521 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.525 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.526 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.529 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.531 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.536 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.540 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.540 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.543 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.550 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.555 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.559 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.564 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.564 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.567 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.571 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.575 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.579 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.580 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.584 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.587 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.591 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.596 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.599 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.603 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.607 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.608 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.610 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
2023-06-15 06:03:10.614 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.619 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/sed1200-a.bin
2023-06-15 06:03:10.623 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/cm32l_pcm.rom
2023-06-15 06:03:10.624 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/a__m-9.27c256.ic27.bin
2023-06-15 06:03:10.627 | MT32: Unknown file in ROM folder: /usr/share/mt32-rom-data/b__m-9.27c256.ic26.bin
```

Some users are going to point `romdir` at a common area used by Munt and ScummVM, so I think we should just print one warning per file, without repeats.

(so probably need a new member `std::set` to hold std_fs::path objects that we've already warned about)

johnnovak:
> I don't want to bloat this PR too much but can you throw this in while you're at it. Also can be included in a follow-up PR that categorises the different ROM sets as indicated above.
> 
> Mainly the green highlight colouring of the active munt ROM like it was before and is with fluidsynth. It also acts as a great way to ascertain are you on `mididevice` auto, fluidsynth or mt32:
> 
>     * [Small improvements to 'mixer /listmidi' #2484](https://github.com/dosbox-staging/dosbox-staging/issues/2484)

Good point @Grounded0 but let's treat that separately as restoring the highlighting is already half-implemented on a branch of mine. I'll finish it after this has been merged.

johnnovak:
Before you guys go into redesigning the `MIXER /LISTMIDI` output @kcgen @weirddan455, I have some ideas too how it could be improved. I'll post a mockup in a hour or two, you might wanna hold off any extensive changes until then.

But probably it's best to merge this as is, then the visual refactoring can be done in a separate ticket (I'm happy to do it).

weirddan455:
> Before you guys go into redesigning the `MIXER /LISTMIDI` output @kcgen @weirddan455, I have some ideas too how it could be improved. I'll post a mockup in a hour or two, you might wanna hold off any extensive changes until then.
> 
> But probably it's best to merge this as is, then the visual refactoring can be done in a separate ticket (I'm happy to do it).

Cool.  I just pushed commits fixing the duplicate logging and re-ordering the ROMs as @kcgen suggested.  I tried setting the mixer output to show the full name but it failed an assertion and crashed because there was not enough room on the terminal to display so I didn't do any changes there.

Agreed that the mixer output should probably be on a separate ticket.

johnnovak:
> > Before you guys go into redesigning the `MIXER /LISTMIDI` output @kcgen @weirddan455, I have some ideas too how it could be improved. I'll post a mockup in a hour or two, you might wanna hold off any extensive changes until then.
> > But probably it's best to merge this as is, then the visual refactoring can be done in a separate ticket (I'm happy to do it).
> 
> Cool. I just pushed commits fixing the duplicate logging and re-ordering the ROMs as @kcgen suggested. I tried setting the mixer output to show the full name but it failed an assertion and crashed because there was not enough room on the terminal to display so I didn't do any changes there.
> 
> Agreed that the mixer output should probably be on a separate ticket.

If @kcgen agrees, I think now that the order and the logging is fixed, this is good to go. Thanks for your help, much appreciated! ❤️ 

Then @kcgen and myself can brainstorm about the UI improvements and treat it separately.

kcgen:
Building issue on macOS:

https://github.com/dosbox-staging/dosbox-staging/actions/runs/5284347668/jobs/9562708411

kcgen:
@weirddan455 - I think we're set to merge!

One small suggestion, if you can reorder and combine (squash) a couple commits, I think there are three groups of changes:



Also suggest naming the commits using the active voice like you've done the first couple (without the `MT32: ` prefix).

kcgen:
Looks great, @weirddan455. All commits build and run clean. 
Inbound as soon as CI goes green.

johnnovak:
Thanks @weirddan455! 😎

Issue #2544, opened 2023-05-27
Switching Tandy on/off uses SBPRO2 instead of SB16 when using 'sbtype=sb16' afterwards


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-alpha-785

### What Operating System are you using?

Mac OS

### Relevant hardware info

M2 mini

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

<img width="1174" alt="Bildschirmfoto 2023-05-27 um 22 51 29" src="https://github.com/dosbox-staging/dosbox-staging/assets/61995128/92ac8e17-effc-4855-9ab6-5ff2b6f9e327">

I don't know if the log message is wrong or if the wrong card settings are actually loaded.

### Steps to reproduce the behaviour.

See screenshot how to reproduce.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Did you start with `-noprimaryconf` and the stock default config? It's definitely a bug according to the screenshot, just wondering where it got the `sbpro2` from, so my hunch is either your local `dosbox.conf` or the global config. Could you please check?

Burrito78:
Yes, '-noprimaryconf', see attached full log:

```
Last login: Sun May 28 00:18:51 on ttys000
/Users/Alex/Games/eXoDOS/eXoDOS.command ; exit;
Alex@MacBook-Alex ~ % /Users/Alex/Games/eXoDOS/eXoDOS.command ; exit;
date       time         | 
2023-05-28 00:19:57.991 | arguments: /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox -noprimaryconf --working-dir /Users/Alex/Games/eXoDOS
2023-05-28 00:19:57.991 | Current dir: /Users/Alex
2023-05-28 00:19:57.991 | stderr verbosity: 0
2023-05-28 00:19:57.991 | -----------------------------------
2023-05-28 00:19:57.992 | dosbox-staging version 0.81.0-alpha-785-g4b9bfc94c
2023-05-28 00:19:57.992 | ---
2023-05-28 00:19:57.992 | LOG: Loguru version 2.1.0 initialized
2023-05-28 00:19:58.125 | SDL: version 2.26.2 initialized (cocoa video and coreaudio audio)
[...]

PR #2593, opened 2023-06-12
Let only one device reserve a DMA channel at a time

kcgen:
Previously we could have three active audio devices all thinking they had access to DMA channel 1. For example:

```ini
[dosbox]
machine = tandy

[sblaster]
sbtype = sb1
dma = 1

[gus]
gus = on
gusdma = 1

[speaker]
tandy = on (the DAC runs on DMA 1)
```

Unfortunately only one device will be functional and the others will be alive, but in a broken unrecoverable zombie state.

This PR adds the ability to reserve a DMA channel: the reservation consists of an owner and a callback. Only one reservation can be held, so if the user wants to start a subsequent (conflicting) audio device, then the prior device is gracefully shut down (through and eviction callback).

The PR then updates the three DMA-capable devices (Sound Blaster, Tandy, and GUS) to request DMA channel reservation.

- Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2544

### DIFF
--- a/include/dma.h
+++ b/include/dma.h
@@ -44,15 +44,15 @@ using DMA_Callback = std::function<void(DmaChannel *chan, DMAEvent event)>;
 class DmaChannel {
 public:
 	// Defaults at the time of initialization
-	uint32_t pagebase = 0;
+	uint32_t page_base = 0;
 	uint32_t curr_addr = 0;
 
 	uint16_t base_addr = 0;
 	uint16_t base_count  = 0;
 	uint16_t curr_count  = 0;
 
-	const uint8_t channum = 0;
-	uint8_t pagenum = 0;
+	const uint8_t chan_num = 0;
+	uint8_t page_num = 0;
 	const uint8_t DMA16 = 0;
 
 	bool increment = true;
--- a/include/dma.h
+++ b/include/dma.h
@@ -39,7 +39,7 @@ class Section;
 using DMA_ReservationCallback = std::function<void(Section*)>;
 
 class DmaChannel;
-using DMA_Callback = std::function<void(DmaChannel* chan, DMAEvent event)>;
+using DMA_Callback = std::function<void(const DmaChannel* chan, DMAEvent event)>;
 
 class DmaChannel {
 public:
@@ -66,7 +66,7 @@ public:
 	DmaChannel(uint8_t num, bool dma16);
 	~DmaChannel();
 
-	void DoCallback(DMAEvent event);
+	void DoCallback(DMAEvent event) const;
 	void SetMask(bool _mask);
 	void RegisterCallback(const DMA_Callback cb);
 	void ReachedTerminalCount();
--- a/include/dma.h
+++ b/include/dma.h
@@ -122,7 +122,7 @@ public:
 	void ResetChannel(const uint8_t channel_num) const;
 };
 
-DmaChannel * GetDMAChannel(uint8_t chan);
+DmaChannel * DMA_GetChannel(uint8_t chan);
 
 void CloseSecondDMAController();
 void DMA_ResetChannel(const uint8_t channel_num);
--- a/include/dma.h
+++ b/include/dma.h
@@ -68,7 +68,7 @@ public:
 
 	void DoCallback(DMAEvent event);
 	void SetMask(bool _mask);
-	void Register_Callback(const DMA_Callback cb);
+	void RegisterCallback(const DMA_Callback cb);
 	void ReachedTC();
 	void SetPage(uint8_t val);
 	void Raise_Request();
--- a/include/dma.h
+++ b/include/dma.h
@@ -69,7 +69,7 @@ public:
 	void DoCallback(DMAEvent event);
 	void SetMask(bool _mask);
 	void RegisterCallback(const DMA_Callback cb);
-	void ReachedTC();
+	void ReachedTerminalCount();
 	void SetPage(uint8_t val);
 	void RaiseRequest();
 	void ClearRequest();
--- a/include/dma.h
+++ b/include/dma.h
@@ -35,6 +35,9 @@ enum DMAEvent {
 	DMA_UNMASKED,
 };
 
+class Section;
+using DMA_ReservationCallback = std::function<void(Section*)>;
+
 class DmaChannel;
 using DMA_Callback = std::function<void(DmaChannel *chan, DMAEvent event)>;
 
@@ -61,6 +64,7 @@ public:
 	DMA_Callback callback = {};
 
 	DmaChannel(uint8_t num, bool dma16);
+	~DmaChannel();
 
 	void DoCallback(DMAEvent event);
 	void SetMask(bool _mask);
@@ -72,8 +76,22 @@ public:
 	size_t Read(size_t words, uint8_t* dest_buffer);
 	size_t Write(size_t words, uint8_t* src_buffer);
 
+	// Reset the channel back to defaults, without callbacks or reservations.
+	void Reset();
+
+	// Reserves the channel for the owner. If a subsequent reservation is
+	// made then the previously held reservation callback is run to
+	// cleanup/remove that reserver (see the EvictReserver call below).
+	void ReserveFor(const std::string_view new_owner,
+	                const DMA_ReservationCallback new_cb);
+
 private:
+	void EvictReserver();
+	bool HasReservation();
 	size_t ReadOrWrite(DMA_DIRECTION direction, size_t words, uint8_t *buffer);
+
+	DMA_ReservationCallback reservation_callback = {};
+	std::string_view reservation_owner = {};
 };
 
 class DmaController {
@@ -101,6 +119,7 @@ public:
 
 	void WriteControllerReg(io_port_t reg, io_val_t value, io_width_t width);
 	uint16_t ReadControllerReg(io_port_t reg, io_width_t width);
+	void ResetChannel(const uint8_t channel_num) const;
 };
 
 DmaChannel * GetDMAChannel(uint8_t chan);
@@ -109,5 +128,6 @@ void CloseSecondDMAController(void);
 bool SecondDMAControllerAvailable(void);
 
 void DMA_SetWrapping(const uint32_t wrap);
+void DMA_ResetChannel(const uint8_t channel_num);
 
 #endif
--- a/include/dma.h
+++ b/include/dma.h
@@ -47,7 +47,7 @@ public:
 	uint32_t pagebase = 0;
 	uint32_t curr_addr = 0;
 
-	uint16_t baseaddr = 0;
+	uint16_t base_addr = 0;
 	uint16_t basecnt  = 0;
 	uint16_t currcnt  = 0;
 
--- a/include/dma.h
+++ b/include/dma.h
@@ -124,10 +124,8 @@ public:
 
 DmaChannel * GetDMAChannel(uint8_t chan);
 
-void CloseSecondDMAController(void);
-bool SecondDMAControllerAvailable(void);
-
-void DMA_SetWrapping(const uint32_t wrap);
+void CloseSecondDMAController();
 void DMA_ResetChannel(const uint8_t channel_num);
+void DMA_SetWrapping(const uint32_t wrap);
 
 #endif
--- a/include/dma.h
+++ b/include/dma.h
@@ -100,8 +100,8 @@ private:
 
 	std::unique_ptr<DmaChannel> dma_channels[4] = {};
 
-	IO_ReadHandleObject DMA_ReadHandler[0x12]   = {};
-	IO_WriteHandleObject DMA_WriteHandler[0x12] = {};
+	IO_ReadHandleObject io_read_handlers[0x12]   = {};
+	IO_WriteHandleObject io_write_handlers[0x12] = {};
 
 	const uint8_t index = 0;
 
--- a/include/dma.h
+++ b/include/dma.h
@@ -53,7 +53,7 @@ public:
 
 	const uint8_t chan_num = 0;
 	uint8_t page_num = 0;
-	const uint8_t DMA16 = 0;
+	const uint8_t is_16bit = 0;
 
 	bool increment = true;
 	bool autoinit  = false;
--- a/include/dma.h
+++ b/include/dma.h
@@ -71,8 +71,8 @@ public:
 	void RegisterCallback(const DMA_Callback cb);
 	void ReachedTC();
 	void SetPage(uint8_t val);
-	void Raise_Request();
-	void Clear_Request();
+	void RaiseRequest();
+	void ClearRequest();
 	size_t Read(size_t words, uint8_t* const dest_buffer);
 	size_t Write(size_t words, uint8_t* const src_buffer);
 
--- a/include/dma.h
+++ b/include/dma.h
@@ -45,7 +45,7 @@ class DmaChannel {
 public:
 	// Defaults at the time of initialization
 	uint32_t pagebase = 0;
-	uint32_t curraddr = 0;
+	uint32_t curr_addr = 0;
 
 	uint16_t baseaddr = 0;
 	uint16_t basecnt  = 0;
--- a/include/dma.h
+++ b/include/dma.h
@@ -1,4 +1,5 @@
 /*
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
  *  Copyright (C) 2002-2021  The DOSBox Team
  *
  *  This program is free software; you can redistribute it and/or modify
--- a/include/dma.h
+++ b/include/dma.h
@@ -55,11 +55,11 @@ public:
 	uint8_t page_num = 0;
 	const uint8_t is_16bit = 0;
 
-	bool increment = true;
-	bool autoinit  = false;
-	bool masked    = true;
-	bool tcount    = false;
-	bool request   = false;
+	bool is_incremented = true;
+	bool is_autoiniting  = false;
+	bool is_masked    = true;
+	bool has_reached_terminal_count    = false;
+	bool has_raised_request = false;
 
 	DMA_Callback callback = {};
 
--- a/include/dma.h
+++ b/include/dma.h
@@ -124,7 +124,7 @@ public:
 
 DmaChannel * DMA_GetChannel(uint8_t chan);
 
-void CloseSecondDMAController();
+void DMA_ShutdownSecondaryController();
 void DMA_ResetChannel(const uint8_t channel_num);
 void DMA_SetWrapping(const uint32_t wrap);
 
--- a/include/dma.h
+++ b/include/dma.h
@@ -49,7 +49,7 @@ public:
 
 	uint16_t base_addr = 0;
 	uint16_t base_count  = 0;
-	uint16_t currcnt  = 0;
+	uint16_t curr_count  = 0;
 
 	const uint8_t channum = 0;
 	uint8_t pagenum = 0;
--- a/include/dma.h
+++ b/include/dma.h
@@ -36,64 +36,41 @@ enum DMAEvent {
 };
 
 class DmaChannel;
-using DMA_CallBack = std::function<void(DmaChannel *chan, DMAEvent event)>;
+using DMA_Callback = std::function<void(DmaChannel *chan, DMAEvent event)>;
 
 class DmaChannel {
 public:
-	uint32_t pagebase;
-	uint16_t baseaddr;
-	uint32_t curraddr;
-	uint16_t basecnt;
-	uint16_t currcnt;
-	uint8_t channum;
-	uint8_t pagenum;
-	uint8_t DMA16;
-	bool increment;
-	bool autoinit;
-//	uint8_t trantype; //Not used at the moment
-	bool masked;
-	bool tcount;
-	bool request;
-	DMA_CallBack callback;
+	// Defaults at the time of initialization
+	uint32_t pagebase = 0;
+	uint32_t curraddr = 0;
+
+	uint16_t baseaddr = 0;
+	uint16_t basecnt  = 0;
+	uint16_t currcnt  = 0;
+
+	const uint8_t channum = 0;
+	uint8_t pagenum = 0;
+	const uint8_t DMA16 = 0;
+
+	bool increment = true;
+	bool autoinit  = false;
+	bool masked    = true;
+	bool tcount    = false;
+	bool request   = false;
+
+	DMA_Callback callback = {};
 
 	DmaChannel(uint8_t num, bool dma16);
 
-	void DoCallBack(DMAEvent event) {
-		if (callback)
-			callback(this, event);
-	}
-	void SetMask(bool _mask) {
-		masked=_mask;
-		DoCallBack(masked ? DMA_MASKED : DMA_UNMASKED);
-	}
-	void Register_Callback(DMA_CallBack _cb) { 
-		callback = _cb; 
-		SetMask(masked);
-		if (callback) Raise_Request();
-		else Clear_Request();
-	}
-	void ReachedTC(void) {
-		tcount=true;
-		DoCallBack(DMA_REACHED_TC);
-	}
-	void SetPage(uint8_t val) {
-		pagenum=val;
-		pagebase=(pagenum >> DMA16) << (16+DMA16);
-	}
-	void Raise_Request(void) {
-		request=true;
-	}
-	void Clear_Request(void) {
-		request=false;
-	}
-	size_t Read(size_t words, uint8_t *dest_buffer)
-	{
-		return ReadOrWrite(DMA_DIRECTION::READ, words, dest_buffer);
-	}
-	size_t Write(size_t words, uint8_t *src_buffer)
-	{
-		return ReadOrWrite(DMA_DIRECTION::WRITE, words, src_buffer);
-	}
+	void DoCallback(DMAEvent event);
+	void SetMask(bool _mask);
+	void Register_Callback(const DMA_Callback cb);
+	void ReachedTC();
+	void SetPage(uint8_t val);
+	void Raise_Request();
+	void Clear_Request();
+	size_t Read(size_t words, uint8_t* dest_buffer);
+	size_t Write(size_t words, uint8_t* src_buffer);
 
 private:
 	size_t ReadOrWrite(DMA_DIRECTION direction, size_t words, uint8_t *buffer);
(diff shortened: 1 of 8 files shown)


johnnovak:
I get the following warning then activating any SB model on a Tandy machine:

```
2023-06-12 16:36:08.886 | DMA: Cannot activate the secondary (high) DMA controller because the machine needs to be at least EGA
```

AFAIK, high DMA is needed for two things: 16-bit playback, and full-duplex operation. Full-duplex means that you can simultaneously use the card for recording and playback, and as we're not emulating recording, I'm wondering why we even need the high DMA for any pre SB16 models? At the very least, we should definitely not use high DMA for `sb1` which was 8-bit and non-full-duplex.

Sound Blaster 2.0 — first full-duplex SB card
Sound Blaster 16 — first SB card that supports 16-bit audio

johnnovak:
Setting SB to DMA 3 replaced the GUS, but the logs are wrong:

image

image

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
-	{
-		return ReadOrWrite(DMA_DIRECTION::READ, words, dest_buffer);
-	}
-	size_t Write(size_t words, uint8_t *src_buffer)
-	{
-		return ReadOrWrite(DMA_DIRECTION::WRITE, words, src_buffer);
-	}
+	DMA_Callback callback = {};
+
+	DmaChannel(uint8_t num, bool dma16);
```
johnnovak, line 65: const?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
+	~DmaChannel();
+
+	void DoCallback(DMAEvent event);
+	void SetMask(bool _mask);
+	void Register_Callback(const DMA_Callback cb);
```
johnnovak, line 70: Lose underscore?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
+	void ReachedTC();
+	void SetPage(uint8_t val);
+	void Raise_Request();
+	void Clear_Request();
```
johnnovak, line 74: Lose underscore?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
+	DMA_Callback callback = {};
+
+	DmaChannel(uint8_t num, bool dma16);
+	~DmaChannel();
+
+	void DoCallback(DMAEvent event);
+	void SetMask(bool _mask);
+	void Register_Callback(const DMA_Callback cb);
+	void ReachedTC();
+	void SetPage(uint8_t val);
```
johnnovak, line 72: const?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
-	}
+	DMA_Callback callback = {};
+
+	DmaChannel(uint8_t num, bool dma16);
+	~DmaChannel();
+
+	void DoCallback(DMAEvent event);
+	void SetMask(bool _mask);
+	void Register_Callback(const DMA_Callback cb);
+	void ReachedTC();
```
johnnovak, line 71: What's TC?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
-	size_t Write(size_t words, uint8_t *src_buffer)
-	{
-		return ReadOrWrite(DMA_DIRECTION::WRITE, words, src_buffer);
-	}
+	DMA_Callback callback = {};
+
+	DmaChannel(uint8_t num, bool dma16);
+	~DmaChannel();
+
+	void DoCallback(DMAEvent event);
```
johnnovak, line 68: const?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
+	void SetMask(bool _mask);
```
johnnovak, line 69: const?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
+	void Register_Callback(const DMA_Callback cb);
+	void ReachedTC();
+	void SetPage(uint8_t val);
+	void Raise_Request();
+	void Clear_Request();
+	size_t Read(size_t words, uint8_t* dest_buffer);
+	size_t Write(size_t words, uint8_t* src_buffer);
```
johnnovak, line 76: const?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
+	~DmaChannel();
+
+	void DoCallback(DMAEvent event);
+	void SetMask(bool _mask);
+	void Register_Callback(const DMA_Callback cb);
+	void ReachedTC();
+	void SetPage(uint8_t val);
+	void Raise_Request();
+	void Clear_Request();
+	size_t Read(size_t words, uint8_t* dest_buffer);
```
johnnovak, line 75: Mark method as const?

include/dma.h
```diff
@@ -35,114 +35,96 @@ enum DMAEvent {
...
 class DmaController {
 private:
-	bool flipflop;
-	DmaChannel *dma_channels[4];
+	bool flipflop = false;
+
+	std::unique_ptr<DmaChannel> dma_channels[4] = {};
+
+	IO_ReadHandleObject DMA_ReadHandler[0x12]   = {};
+	IO_WriteHandleObject DMA_WriteHandler[0x12] = {};
```
johnnovak, line 103: DmaReadHandler / DmaWriteHandler?

src/hardware/gus.cpp
```diff
@@ -594,6 +596,14 @@ Gus::Gus(const io_port_t port_pref, const uint8_t dma_pref, const uint8_t irq_pr
           irq1(irq_pref),
           irq2(irq_pref)
 {
+	// Bail out if we can't set the DMA address, then . This will happen if the
```
johnnovak, line 599: First sentence seems unfinished?

src/hardware/sblaster.cpp
```diff
@@ -2057,17 +2057,24 @@ SB_TYPES find_sbtype()
...
+
+		const auto dma_pref = static_cast<uint8_t>(sect->Get_int("dma"));
+		const auto hdma_pref = static_cast<uint8_t>(sect->Get_int("hdma"));
+
+		const auto have_both_channels = (GetDMAChannel(dma_pref) &&
+		                                 GetDMAChannel(hdma_pref));
+		if (!have_both_channels) {
+			LOG_WARNING("SB: Cannot initialize %s because it requires the secondary "
+			            "(high) DMA controller, falling back to 'sbpro2'",
+			            pref.data());
```
johnnovak, line 2075: Nice 👍🏻

src/hardware/dma.cpp
```diff
@@ -100,29 +106,71 @@ static void perform_dma_io(const DMA_DIRECTION direction,
...
+	primary = std::make_unique<DmaController>(primary_index);
+	return true;
+}
+
+static bool activate_secondary()
+{
+	// We need an AT architecture to support the secondary controller
+	if (!IS_EGAVGA_ARCH) {
+		LOG_WARNING("DMA: Cannot activate the secondary (high) DMA controller "
+		            "because the machine needs to be at least EGA");
```
johnnovak, line 124: Would it be better to be a bit more precise here and say "because the machine type needs to be at least 'ega'"?

This is something new to me that we default to `sbpro2` on the `hercules` and `cga` machine types and only allow SB16 on `ega` or better. Is there a technical reason for this limitation? In real life, people were using Hercules as a secondary display adapter right until the end of the 90s, and you can use CGA and Hercules adapters in 386 and 486 systems without problems. Maybe Pentium too, but not sure about that.

I guess we might want to keep the defaulting this way due to backward compatibility reasons, but is it not possible to simply just allows any SB type on any machine type, excluding `pcjr` and `tandy` as those emulate some very specific hardware?

Also, starting DOSBox simply with `-noprimaryconf -machine hercules` now suddenly generates lots of warnings.

image

This seems a little silly as `sbpro2` is what has been auto-selected based on the machine type, at least conceptually. What's really happening I think is that we default to `sb16` which then gets "demoted" to SB Pro2, and with the new logging we're informed about this in excruciating detail 😅 
 
If we really want to keep this behaviour, I think introducing `sbtype = auto` is in order perhaps?

src/hardware/dma.cpp
```diff
@@ -100,29 +106,71 @@ static void perform_dma_io(const DMA_DIRECTION direction,
...
+}
+
+constexpr uint8_t to_secondary_index(const uint8_t channel_num)
+{
+	assert(channel_num >= SecondaryMin);
+	assert(channel_num <= SecondaryMax);
+	return (static_cast<uint8_t>(channel_num - SecondaryMin));
+}
+
+DmaChannel* GetDMAChannel(const uint8_t channel_num)
```
johnnovak, line 160: Shouldn't this be called `DMA_GetChannel` for consistency? It's not using in many places, so consider renaming it.

src/hardware/dma.cpp
```diff
@@ -100,29 +106,71 @@ static void perform_dma_io(const DMA_DIRECTION direction,
...
+		return secondary->GetChannel(to_secondary_index(channel_num));
+	}
+	return nullptr;
 }
 
-/* check availability of second DMA controller, needed for SB16 */
-bool SecondDMAControllerAvailable(void) {
-	if (DmaControllers[1]) return true;
-	else return false;
+void CloseSecondDMAController()
```
johnnovak, line 171: DMA_CloseSecondController? (and same comment for all functions part of the public API)

src/hardware/dma.cpp
```diff
@@ -356,74 +446,184 @@ size_t DmaChannel::ReadOrWrite(DMA_DIRECTION direction, size_t words, uint8_t *b
...
+}
+
+void DmaChannel::Reset()
+{
+	pagebase = 0;
+	curraddr = 0;
+
+	baseaddr = 0;
+	basecnt  = 0;
+	currcnt  = 0;
```
johnnovak, line 497: Time to fix these up somewhat? `curr_count`, `base_count`, etc.

Burrito78 (approved):
Works like expected, sb16 activates sb16 now always.

Good job as always, @kcgen!

kcgen:
@johnnovak , @Burrito78 - thanks for the testing and feedback!

Yup - the logging is excessive for the PR; I wanted to show the reservation and eviction process in play.  Will knock it out in the next rev, and will only show logging when a prior audio card is shut down to make room for the requested card.

src/hardware/dma.cpp
kcgen: > This is something new to me that we default to sbpro2 on the hercules and cga machine types and only allow SB16 on ega or better. Is there a technical reason for this limitation? In real life, people were using Hercules as a secondary display adapter right until the end of the 90s, and you can use CGA and Hercules adapters in 386 and 486 systems without problems. Maybe Pentium too, but not sure about that.

Yup, the `machine` setting sets up the BIOS and motherboard memory, registers, and IO handlers to provide a period-correct environment for the give video device. 

`hercules`, `cga`, `tandy`, and `pcjr` were released when XT-class machines were only available, and software that was written in this pre-AT era could make assumptions (or confirm) that the machine meets the XT specifications.

Where as software written in the AT-era had to provide backward compatibility for XT systems and hardware; exactly like you said, people could plug in a CGA card to their pentium-class AT motherboard and BIOS.

For the best CGA experience, when a given game was written in the XT timeframe (and EGA wasn't even a thing then), using `machine = cga` gives the game the correct XT environment, and yeah - those motherboards on had a single 8-bit DMA controller.

kcgen: > This seems a little silly as sbpro2 is what has been auto-selected based on the machine type, at least conceptually. What's really happening I think is that we default to sb16 which then gets "demoted" to SB Pro2, and with the new logging we're informed about this in excruciating detail :sweat_smile:

Yes :-)  I'll knock down that logging :laughing: , and just mention that we're falling back to SB Pro2.

kcgen: > I think introducing `sbtype = auto` is in order perhaps? Then sbpro2 would be selected for anything lesser than ega and we can omit this downgrading dance...

That's a great suggestion!

- Perhaps older SB's might be preferred in the `cga`, `hercules` and `pcjr` machine types, so CMS is also available?

- Also, if a user is explicitly setting `machine = ega`, we /might/ want to give them `sbpro2`, because that's probably a better fit for the era of software they're using.  As an example: were there ever EGA-authored games that used 16-bit audio samples? EGA/<=286-era faded out as the primary target platform for games by ~1991 with devs now targeting VGA/386/486, and the SB16 wouldn't be released until 1992.

Also, the SB16 CT mixer's volume ramp is not backward compatible with pre-SB16 cards' volume ramp.  If you take a pre-SB16-era game that happens to adjust voice volumes and give it an SB16, then those volumes will be /way/ too quiet. They require the older volume ramp.  You'll see users working around this with the `sbmixer = false` setting.. but really, that's an indication the game wasn't authored for the SB16.

kcgen: > But I guess I'd prefer removing the limitation altogether except for Tandy/PCjr if possible (but we might still wanna keep the defaulting via the auto type).

The constraint here is that the XT machines didn't have a 16-bit DMA controller - and for those machine types, we really should keep it that way.

However, because this PR directly tests if the SB16's configured  `"hdma"` DMA channel is available, it is now possible to use an SB16 with these XT machine types.  You simply have to set `"hdma = 1"` or `"hdma = 3"`, which will use the 8-bit DMA controller to move 16-bit audio samples.

Edit: turns out this was possible with hardware too using the same technique (configuring the SB16's jumpers to use DMA address 1 for both the low and high channels):

### Running a Sound Blaster 16 on an XT

https://dosdays.co.uk/topics/sb16_on_old_systems.php

rderooy: > > But I guess I'd prefer removing the limitation altogether except for Tandy/PCjr if possible (but we might still wanna keep the defaulting via the auto type).
> 
> The constraint here is that the XT machines didn't have a 16-bit DMA controller - and for those machine types, we really should keep it that way.

Isn't that kind of moot, since there is no support for XT (or even AT) machine types (e.g. 8086/8088 or 286 CPUs) at this point? By default DOSBox emulates a 486 type CPU which will always have two PICs.

kcgen: I guess both be satisfied:

- Out of the box, `sbtype = auto` can respect the XT machine-types by picking an early SB model, similar to the "you asked for sb16 but we gave  you sbpro2 approach).

- But if the user demands a 16-bit DMA channel on an XT machine type ( `sbtype = sb16` or `gusdma = 6`), then we can still allow it.

johnnovak: Very good points @kcgen, I like these ideas and would like them to be implemented. 🎉  Rather than replying to the individual points here, I started writing a proposal I'm going to post as a "discussion" soon because we're touching on some overarching concepts here that goes beyond this single enhancement.

johnnovak:
> @johnnovak , @Burrito78 - thanks for the testing and feedback!
> 
> Yup - the logging is excessive for the PR; I wanted to show the reservation and eviction process in play. Will knock it out in the next rev, and will only show logging when a prior audio card is shut down to make room for the requested card.

That's cool, but note that I believe the logging is actually wrong in some cases as mentioned here: https://github.com/dosbox-staging/dosbox-staging/pull/2593#issuecomment-1586688782 (the log always says DMA 1; I think that's wrong).

src/hardware/dma.cpp
johnnovak: Okay, so I was quite enthusiastic about this auto-configuration mechanism and started writing a proposal about how we could configure all sorts of settings based on the `machine` type. Long story short, the more I thought about this, the more I disliked it. Something rubs me very much the wrong way with making all sorts of choices based on the *type of graphics adapter* (!) in the system.

_**Clippy, the DOSBox Assitant:** Alrighty, so because you've configured a `machine` with a Hercules graphics adapter, I'll set the `cycles` setting for you to XT speeds, moreover, I'll give you an `sb1` card, and maybe also disable `xms`, `ems`, and `umb`... Isn't that just W–O–N–D–E–R–F–U–L?_

Um, no, it's not. Just nope. Get lost.

The problem with such extensive implicit, invisible auto-config behaviour is that it can become very confusing and very hard to figure out what really is going on. For some people even the layered config approach is just too much, and frankly, I often confuse myself when I have my global config, local config, plus I also inject some further configs with the `-conf` option. You really need to know what you're doing if you're going down this route, and it's just too much for ordinary users.

Instead, I'm quite convinced providing some machine templates in the new user manual is the way forward that users can _copy_ into their local config and adjust further. That way everything is kept explicit, not some weird cascading mechanism where you have your global config, your local config, your "machine template config", maybe DOSBox also auto-configures a few things here and there... argh, my brain hurts just thinking about it (and the massive influx of confused people raising support tickets left, right, and center).

So, after so much tirade, what is the _least_ confusing, most explicit option we can do here? Do we _have_ to downgrade the SB16 or GUS to something else when a `machine` lesser than `ega` is configured? I'd much rather do the `hdma` configuration change you mentioned @kcgen than changing the `sbtype` by magic. If the config says `sb16`, we should do our absolute best to give the use `sb16` by whatever means necessary—especially because historically we default to `sb16` and SB is enabled by default. And yeah, I _really_ dislike the "shapeshifting Sound Blaster model" behaviour based on the *graphics adapter* type, let's not go down this path at all.

rderooy: I do think that having some sensible default "template" machine options for people to select between would be a good option, but that could come with the GUI, whenever that is. Something like:

- XT machine type with Hercules and SB 1.5
- XT machine type with CGA and SB 1.5
- AT machine type with EGA and SB Pro 2
- 386 machine type with VGA and SB Pro 2
- 486 machine type with SVGA and SB16
- Pentium machine type with SVGA and SB16
- Tandy with DAC machine type (286?)
- PCjr machine type

johnnovak: Yeah we can revisit it later. Some GUI for this would be a viable option, because then you select a template which just sets various settings, but all settings are laid out before you on the GUI. You see exactly what settings the template set for you. Whereas if things get configured "magically" or by setting something like `configuration = 486` in the config, then how do you know what that actually configured for you? You have no idea, which makes further tweaking of those settings very difficult and confusing. Yes, we could document it, but if you have to look things up constantly, you're better off copying a template example from the manual into your local config, then it's all laid out before you for further tweaking... I don't think it gets better than that.

But yeah, a GUI approach takes all uncertainty out of that. You instantiate the template, you see exactly what the template configured for you, tweak further, then save. But the GUI approach would not play well at all with layered configs. They're almost totally mutually exclusive, at least to maintain them in a sane way. If a GUI was involved, I'd made sure that all layering and global config usage is turned off, and the GUI fully manages all the defaulting, templates, and whatnot, and always writes a _full_ local config, and even switches off global config loading with `-noprimaryconf`. How I see it, you either manually configure things, then you get layering, or you use a GUI, which then fully takes over the config management. Attempting to reconcile the two is a route to madness, I think.

For the record, the planned OSD is not really meant to be a full-blown DOSBox configurer/frontend sort of utility, at least in my mind. It would only allow changing settings which are useful to change in realtime while a game is running. Reconfiguring your whole machine that at the end requires a full emualtor restart anyway is out of scope, I think. A good guiding principle would be that if something cannot be tweaked at runtime while the game is running, then it won't belong to the OSD.

kcgen: We're in the DOSBox world, so we need to remember that `machine` is [supposed](url) be the entire system. 

What does this mean in practical terms?

- If you plugged a real TGA graphics card and TandSound card into a 286 DOS PC, then it still wouldn't run most Tandy-specific games. `machine = tandy` on the other hand will, [because it fakes out the BIOS chips, too](https://github.com/dosbox-staging/dosbox-staging/blob/main/src/ints/bios.cpp#L775:L808), and perhaps other details that I'm not aware of.

- `machine = tandy` is supposed to mimic an entire Tandy 1000 system so that all "APIs" (BIOS, IO devices, video, audio, etc) look like that machine, well, enough to coax the Tandy games to run. 

The point is that `machine = ` is not just swapping the graphics adapter in a bog-standard "PC clone".  It's meant to be the entire 'holistic' machine enough to satisfy picky games that really do need the `machine = ` specification.  (yes: the CPU model and speed doesn't follow suite: I suspect the DOSBox devs left that as-is because the _majority_ of games aren't CPU-sensitive and thus are unaffected by this)

> what is the least confusing, most explicit option we can do here? If the config says sb16, we should do our absolute best to give the use sb16 by whatever means necessary.

We should respect that - and give it to them, even if it means turning on the 16-bit DMA controller in those old machine types.

This PR also prevent DMA conflicts producing broken zombie sound cards.

So if SB wants DMA 1, and GUS wants DMA 1, and TandySound wants DMA 1 -> then only the last request will get it and prior users will be shudown explicitly in order, like musical chairs.  But we'll inform the user at each change.

> Do we have to downgrade the SB16 or GUS to something else when a machine lesser than ega is configured? 

I also think the magic downgrade "You set `sbtype = sb16` but we gave you `sbtype = sbpro2` -> that's asking for problems, especially if a user has configured their game explicitly for sb16.

Same with any magic changes to GUS *"You asked for `gusdma = 6`, but we sneakily configured `gusdma = 3`"* -> that's evil and also will break user's games that have hard-coded DMA values.

But! We do have a better opportunity with `sbtype = auto` to provide reasonable period "sane" SB types (especially critical to get the volume ramp correct), and for that, @rderooy's suggestion makes a lot of sense.

johnnovak: Ok, we both think the magic downgrade is a bad idea, so far so good.

About the auto-config, I get it that the `machine` type is a fuzzy concept and it does not just set the graphics adapter. Arguably, this whole machine concept is strongest with `pcjr`and `tandy`--with the rest it's much looser.

One issue with `sbtype = auto` is that the choice of SB type is pretty arbitrary, there's no clear best setting.

- SB1 came out in 1989. With what SB adapter do you configure the "Hercules XT"? I'd say `none`. But maybe `gb`, `sb1`, or even `sbtype = none` and ` oplmode = opl2` are reasonable and realistic options.

- Exact same issue with CGA.

- PCjr and Tandy should get `none`, that's clear.

- With EGA it's clearer that at least OPL2 is reasonable, but probably SB1 or SB2 are best. EGA machines are Sierra machines 😎 

- With SVGA types do we differentiate? Or set a single SB model for all of them? 

At the end of the day, regardless that emulating a full Tandy and PCjr machine can requested via the `machine` setting, otherwise when I request `cga` or `ega` or °svga_s3`, I'm still thinking about graphics. My choice of graphics affecting what SoundBlaster model I get (including perhaphs none at all) is very counterintuitive and unexpected, and in many cases the auto-selected choice is pretty arbitrary. 

My other problem is, where do you draw the line? Also auto-default `cycles` based on `machine`, and `memsize` too? I started writing it down what to default these other settings to per each `machine`, I can share it with you. But then I realised, I won't remember what the auto-defaults per machine would be; and there won't be any form of good feedback on the auto-configured behaviour.

We can continue to play with these ideas for sure (I abandoned them in favour of explicitness), but what do you think about the two above issues? My ideal solution is user education: give them period accurate config examples in the manual, mentioning the possible variations for each, then they should use those as starting points and tweak further.

johnnovak: > But! We do have a better opportunity with sbtype = auto to provide reasonable period "sane" SB types (especially critical to get the volume ramp correct), 

I don't think you can fix that with defaulting. Games that were specifically designed for SB16 and not for the lowest common denominator SB Pro 2 started appearing from 1995 onwards only. In 1993 the SB16 was a flop at release it was so buggy, plus most people had their SB Pros already, so adoption was slow. 

All in all, we can default to SB Pro 2 for sure, I think that's a better option. But that means we never auto select the SB16. We can't make that decision based on the configured graphics adapter/machine type (we cannot infer that "this is a 1995+ game").

johnnovak: Also, if we had total freedom, I'd do this:

```ini
machine = pc | pcjr | tandy
video_adapter = hercules | cga | cga_mono | ega | svga_*
```

Or even split the `pc` type further if we want to emulate the lack of high DMA for `xt`:

```ini
machine = xt | at | pcjr | tandy
```

Certain `video_adapter` types would not be allowed for `pcjr` and `tandy`. Actually, we *could* make this scheme even work in a backwards-compatible manner if we wanted to!

Then technically you could put an EGA or VGA adapter in a Tandy. But maybe not that interesting/useful from a DOSBox perspective. But with the current conflated way of things you'd need something silly such as `tandy_ega` and `tandy_vga`...

I guess my point is the conflation of machines types with video adapters is bad enough already; I don't really want to pour oil on the fire... And _really_ dislike the conflation of vaguely (un)related things 😄

include/dma.h
kcgen: Renamed to `ReachedTerminalCount`

include/dma.h
kcgen: Marked the buffer as a const-pointer, but the Channel itself is modified during the read (it keeps track of how much data is left before reaching terminal count).

include/dma.h
kcgen: pointers marked const.

src/hardware/dma.cpp
kcgen: The new functionality does what the user requests:

- If TandySound is running followed by a request for the SB16 using `'hdma' > 4` or GUS using `'gusdma' > 4, TandySound is shutdown to avoid IO port conflicts with the (now required) secondary DMA controller.

- Likewise, if SB16 or GUS are using the secondary controller followed by a request for TandySound, then the secondary controller is shutdown (hich cascades to shutting down devices on its channels).

- The same conflict resolution happens for DMA channels: The prior device is evicted to make way for the new resevation. For example, TandySound uses DMA channel 1 for its DAC, just like the Sound Blaster. Therefore, these two cannot co-exist on channel 1.

kcgen:
Ready for review and testing!

Please try all combinations of enabling:
- `tandy`, plus
- `sbtype` + `dma` values, plus
- `gus` + `gusdma` values

Followed by in-game checks to make sure the audio device is working OK.

src/hardware/dma.cpp
johnnovak: Very nice @kcgen.

johnnovak:
> Ready for review and testing!
> 
> Please try all combinations of enabling:
> 
>     * `tandy`, plus
> 
>     * `sbtype` + `dma` values, plus
> 
>     * `gus` + `gusdma` values
> 
> 
> Followed by in-game checks to make sure the audio device is working OK.

For the record , Indiana Jones and the Last Crusade, original EGA version is a very good test subject for this as it has Tandy support too and it can use the PC Speaker, AdLib, GameBlaster or Tandy 3-Voice the music. You can force the audio device via command line flags, or let it autodetect the best device (e.g. if AdLib is present, it prefers that).

Although it does not use the SB DAC.

kcgen:
> > Ready for review and testing!
> > Please try all combinations of enabling:
> > ```
> > * `tandy`, plus
> > 
> > * `sbtype` + `dma` values, plus
> > 
> > * `gus` + `gusdma` values
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Followed by in-game checks to make sure the audio device is working OK.
> 
> For the record, Indiana Jones and the Last Crusade, original EGA version, is a very good test subject for this as it has Tandy support too and it can use the PC Speaker, AdLib, GameBlaster, or Tandy 3-Voice for the music. You can force the audio device via command line flags, or let it auto-detect the best device (e.g. if AdLib is present, it will prefer that).
> 
> Although it does not use the SB DAC.

Perfect! Will do some rounds of testing with it. 
4D Boxing uses the DAC, so can test that too.
Will report back.

src/hardware/sblaster.cpp
```diff
@@ -2215,6 +2206,24 @@ class SBLASTER final {
...
+		// Only Sound Blaster 16 uses a 16-bit DMA channel.
+		if (sb.type == SB_TYPES::SBT_16) {
+			sb.hw.dma16 = static_cast<uint8_t>(section->Get_int("hdma"));
+
+			// Reserve the second DMA channel only if it's unique.
+			if (sb.hw.dma16 != sb.hw.dma8) {
+				dma_channel = DMA_GetChannel(sb.hw.dma16);
+				assert(dma_channel);
+				dma_channel->ReserveFor(CardType(),
+										SBLASTER_ShutDown);
```
johnnovak, line 2223: formatting

kcgen: I can't see the problem (maybe I'm missing something?)

johnnovak: That's what I see on my screen:

image

kcgen: Oh, thanks - got it now 👍

kcgen:
Tests passing for me when manually rotating on deliberately conflicting DMA channels. For example, setting Tandy on, followed by sbtype sbpro2, (`dma = 1`), and GUS (with `gusdma = 1`). 

Likewise, using other conflicting DMAs (like 5 or 6, between sb16 and GUS) also results in one bumping the prior out.

johnnovak:
I gave this some testing and it looks fine from my side. I suggest merging it if you think it fixes the original problems, then I'll give it a good workaround anyway in my demo pack where I change GUS/SB settings at runtime.

PR #2594, opened 2023-06-12
Improve Unicode character fallback rules

FeralChild64:
Made some changes requested by @Torinde:
- `BYELORUSSIAN-UKRAINIAN I`, if not available, is replaced with `LATIN I` (much more similar), not with `CYRILLIC I`
- `EURO SIGN` might be replaced with `GREEK SMALL LETTER EPSILON` or `ELEMENT OF`, if one of these is available in a given code page


kcgen:
Nice refinement, @FeralChild64, and great suggestions, @Torinde !

kcgen (approved):

Issue #2599, opened 2023-06-14
The default mouse driver only supports up to 80-column text modes


johnnovak:
When using the default mouse settings, the mouse pointer cannot be moved past the 80th column in 132-column VESA text modes.

However, if you load CuteMouse, you can use the full available area (you can move past the 80th column).

## Reproduction steps

All these steps assume stock settings.

1. Go to **Options / Configuration / System Setup** in DOS Navigator 1.51

<img width="990" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/9bc6b97c-c8b7-4204-a62f-f26662a24091">

2. Set **Video Mode 2** to 266

<img width="986" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/c2db5543-bfc1-4296-b3ba-bd21533e1c51">

3. Switch to **Custom video mode 2** either by using the menu option or by pressing **Ctrl+F10**

<img width="987" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/c728fbca-2db4-49f2-a242-19b0dd67a735">

4. In 132-column mode, try to move the mouse pointer to the right edge of the screen. You won't be able to; you cannot move it past the 80th column. If you load CTMOUSE, however, then you can move it to the right edge of the screen.

<img width="1130" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/b18ac560-a59f-4b4d-83e4-a413f91842f1">

FeralChild64:
Most likely a problem within `‎MOUSEDOS_AfterNewVideoMode‎` function - I can see it reads video mode to `uint8_t`, and additionally has a hardcoded list of modes.

johnnovak:
> Most likely a problem within `‎MOUSEDOS_AfterNewVideoMode‎` function - I can see it reads video mode to `uint8_t`, and additionally has a hardcoded list of modes.

Mind you, this is _very_ low priority, I just thought I'd report it for completeness. And probably it's an easy fix.

PR #2601, opened 2023-06-14
DOS mouse driver fixes

FeralChild64:
- small code cleanups
- fix problem with mouse cursor in text modes being limited to 80 columns - for both VESA (which fixes https://github.com/dosbox-staging/dosbox-staging/issues/2599) and non-VESA modes (fixes modes like 94x30)
- set proper screen parameters on DOS mouse driver reset / video mode set hook (partially based on DOSBox-X code)

-----

The easiest way to test mouse driver behavior in various text modes is to use _Necromancer's DOS Navigator_ - open the _burger_ menu in the top-left screen corner, go to _Video modes_.

### DIFF
--- a/include/mouse.h
+++ b/include/mouse.h
@@ -134,7 +134,7 @@ bool MOUSEPS2_SendPacket();
 // ***************************************************************************
 
 void MOUSEDOS_BeforeNewVideoMode();
-void MOUSEDOS_AfterNewVideoMode(const bool setmode);
+void MOUSEDOS_AfterNewVideoMode(const bool is_mode_changing);
 
 // ***************************************************************************
 // MOUSECTL.COM / GUI configurator interface
--- a/src/hardware/input/mouseif_dos_driver.cpp
+++ b/src/hardware/input/mouseif_dos_driver.cpp
@@ -171,7 +171,7 @@ static struct { // DOS driver state
 	int16_t update_region_y[2] = {0};
 
 	uint16_t language = 0; // language for driver messages, unused
-	uint8_t mode      = 0;
+	uint8_t bios_screen_mode = 0;
 
 	// sensitivity
 	uint8_t sensitivity_x = 0;
@@ -385,8 +385,9 @@ static void draw_cursor_text()
 	// Save Background
 	state.background.pos_x = static_cast<uint16_t>(x / 8);
 	state.background.pos_y = static_cast<uint16_t>(y / 8);
-	if (state.mode < 2)
+	if (state.bios_screen_mode < 2) {
 		state.background.pos_x = state.background.pos_x / 2;
+	}
 
 	// use current page (CV program)
 	const uint8_t page = real_readb(BIOSMEM_SEG, BIOSMEM_CURRENT_PAGE);
@@ -512,8 +513,9 @@ static void clip_cursor_area(int16_t &x1, int16_t &x2, int16_t &y1, int16_t &y2,
 
 static void restore_cursor_background()
 {
-	if (state.hidden || state.inhibit_draw || !state.background.enabled)
+	if (state.hidden || state.inhibit_draw || !state.background.enabled) {
 		return;
+	}
 
 	save_vga_registers();
 
@@ -545,9 +547,12 @@ static void restore_cursor_background()
 
 static void draw_cursor()
 {
-	if (state.hidden || state.inhibit_draw)
+	if (state.hidden || state.inhibit_draw) {
 		return;
+	}
+
 	INT10_SetCurMode();
+
 	// In Textmode ?
 	if (CurMode->type == M_TEXT) {
 		draw_cursor_text();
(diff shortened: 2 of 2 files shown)


include/mouse.h
```diff
@@ -134,7 +134,7 @@ bool MOUSEPS2_SendPacket();
 // ***************************************************************************
 
 void MOUSEDOS_BeforeNewVideoMode();
-void MOUSEDOS_AfterNewVideoMode(const bool setmode);
+void MOUSEDOS_AfterNewVideoMode(const bool is_mode_set);
```
kcgen, line 137: Very helpful rename! The original `setmode` a bool appears to _ask_ or _tell_ the function to set some mode.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -385,8 +385,9 @@ static void draw_cursor_text()
 	// Save Background
 	state.background.pos_x = static_cast<uint16_t>(x / 8);
 	state.background.pos_y = static_cast<uint16_t>(y / 8);
-	if (state.mode < 2)
+	if (state.bios_screen_mode < 2) {
```
kcgen, line 388: another small change, but very big quality boost. I thought `state.mode` was a mouse mode :laughing:

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -920,8 +889,120 @@ void MOUSEDOS_AfterNewVideoMode(const bool setmode)
...
 
-	clear_pending_events();
+	// Some software (like 'Down by the Laituri' game) is known to first set
+	// the min/max mouse cursor position and then set VESA mode, therefore
+	// (unless this is a driver reset) skip setting min/max position and
+	// granularity for SVGA graphic modes
+
+	if (is_mode_set && is_svga_mode && !is_svga_text) {
+		return;
+	}
```
kcgen, line 901: Regarding min/max positions, I have a suspicion that the Dostodon client that crashes when moving the mouse out-of-bounds is also having some min/max issues. Are there reasonable or sane min/max values that can be set on a video mode change, so even if the software doesn't set these up right, they still get some protection based on the video mode's bounds?

include/mouse.h
FeralChild64: Actually, I think it is still misleading. The parameter is set to true is the screen mode is changed and to false if this is driver reset. I'll have to think about better name - just not today, it's rather late here.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -920,8 +889,120 @@ void MOUSEDOS_AfterNewVideoMode(const bool setmode)
...
+
+		if (rows < threshold_lo || rows > threshold_hi) {
+			rows = default_rows;
+		}
+		if (columns < threshold_lo || columns > threshold_hi) {
+			columns = default_columns;
+		}
+
+		state.maxpos_x = static_cast<int16_t>(8 * columns - 1);
+		state.maxpos_y = static_cast<int16_t>(8 * rows - 1);
```
kcgen, line 923: Are these `8`'s related to "8-dot-mode text" vs "9-dot-mode text"?

kcgen: If so, the 8 vs. 9 dot mode sequence register bit is held in:

`vga.seq.clocking_mode.is_eight_dot_mode`, a new `bool VGA_GetEightDotMode()` function could be added to `vga.h` to get that state.

src/hardware/input/mouseif_dos_driver.cpp
FeralChild64: If I remember correctly, Dostodon was crashing even when using CtMouse driver. At this concrete point we shouldn't change anything; SVGA text/graphic modes are handled at the end of the function.

src/hardware/input/mouseif_dos_driver.cpp
FeralChild64: I'll check. If so, also `move_cursor_seamless` might need to be adapted.

include/mouse.h
kcgen: if it relates to the `bios_screen_mode` variable below, then perhaps `bios_screen_changed` would be a consistent match.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -920,8 +889,120 @@ void MOUSEDOS_AfterNewVideoMode(const bool setmode)
...
+			// SVGA text mode
+			state.granularity_x = 0xfff8;
+			state.granularity_y = 0xfff8;
+			set_maxpos_text();
+		} else {
+			// SVGA graphic mode
+			state.maxpos_x = static_cast<int16_t>(CurMode->swidth - 1);
+			state.maxpos_y = static_cast<int16_t>(CurMode->sheight - 1);
+		}
+		break;
```
johnnovak, line 1004: So just checking, how would this work with programs that set up a BIOS video mode first, then tweak it further via the CRTC registers to come up with some custom resolution? I'm guessing if DOSBox also sets the `swidth` and `sheight` params to reflect the tweaked dimensions, this will work perfectly. I'll have to double check that, I guess I'm just wondering if you've tested this at all.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -1444,7 +1522,7 @@ static Bitu int33_handler()
 		// Both buttons down = pen pressed, otherwise pen considered
 		// off-screen
 		// TODO: maybe implement light pen using SDL touch events?
-		LOG(LOG_MOUSE, LOG_ERROR)("Mouse light pen emulation not implemented");
+		LOG_WARNING("MOUSE (DOS): Light pen emulation not implemented");
```
johnnovak, line 1525: Nice normalisation of log messages. 🎉  Me approves 😎 🤘🏻

src/hardware/input/mouseif_dos_driver.cpp
kcgen: > if DOSBox also sets the `swidth` and `sheight` params to reflect the tweaked dimensions

Those come from the video mode 'info block' data structure. They're fixed and represent the official mode 'screen' width and height, among other settings, like the text-mode width and height (`tw` and `th`). They can and can be queried in the VESA 2.x case by the DOS program:



The tweak timing values fall out of vga_draw and then hit the `RENDER_` and `GFX_` functions, and I think there's a call in SDL main that updates the `MOUSE_` side when the mode changes (maybe it could be passed back through there?). 

I'm not sure if the tweaked mode resolution is stuffed back into a VGA register values that we can query (unfortunately).

johnnovak: I guess if this newly introduced behaviour is not _worse_ than what we had before, then it's all good and can maybe progressively refine it as needed.

I just wanna make sure we're not introducing regressions for tweaked modes by making decisions based on `swidth` and `sheight`. Like you said, probably this is fine for the defaulting, though, then tweaked modes somehow propagate the changes screen dimensions to the mouse code... "Somehow" being the keyword here, I don't know how it works, neither have the time/inclination now to find out 😄

So I guess it falls to @FeralChild64 to give this some minimal testing in a few tweaked-mode VGA apps to make sure it still works as before.

kcgen: The mouse boundaries are working in Scorched Earth's various tweaked modes.

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/e65b6b56-91ad-42ca-afef-82e6d459c5ab

include/mouse.h
FeralChild64: I have changed to `is_mode_changing` - this will indicate that we are a part of the screen mode change flow.

src/hardware/input/mouseif_dos_driver.cpp
FeralChild64: Actually, I did a small change here - there is no warning log if the application just disables the light pen emulation; _The Settlers_ does this during the startup.

kcgen (approved):
All tests are passing, including tweaked resolution modes. Nice fix, @FeralChild64 !

src/hardware/input/mouseif_dos_driver.cpp
FeralChild64: @kcgen

> If these 8s related to 8-dot-mode vs 9-dot-mode text height, this bit is held in the `vga.seq.clocking_mode.is_eight_dot_mode `register.

Unlikely. I have checked that seamless mouse integration works properly in the text mode regardless if the character is 8 or 9 pixels wide. And setting it to 9 would affect the position we report to guest application. I don't think real mouse driver were that smart.

Issue #2607, opened 2023-06-16
autoexec commands are executed after a given program exits.


amano-kenji:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Linux x86_64

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

autoexec commands are executed after a program given to dosbox command exits.

Thus, if I put mixer command in autoexec section, the audio volume is not changed in a program.

As far as I know, there is no other way to configure audio volume.

### Steps to reproduce the behaviour.

1. Put `mixer master 30 \noshow` in `[autoexec]` config section.
2. Execute `dosbox /path/to/program`
3. The master volume is not 30, but 100.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

FeralChild64:
I can confirm. The problem is within the `src/shell/autoexec.cpp` - some `AddCommandBefore` calls should be changed to `AddCommandAfter`, also secure mode switch might need adaptations.

PR #2609, opened 2023-06-16
 Cleanup AUTOEXEC.BAT generation

FeralChild64:
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2607

Changes (edited, previous idea was corrected):

1. Commands in Z:\AUTOEXEC.BAT are created in the following order:

- set `BLASTER` variable, mount `Y:`, etc.
- mount drive C: (if directory supplied as command line argument)
- mount CD-ROM images as drive D: (if ISO/CUE image supplied as command line argument)
- execute commands supplied after `-c` arguments
- enable secure mode (`Z:\CONFIG.COM -securemode` - if `-securemode` argument and no boot image supplied)
- content of `[autoexec]` sections (unless `-noautoexec` option is supplied)
- `BOOT` command (if floppy image is supplied as argument), or execute program (or call batch file) supplied without argument `-c`
- `EXIT` command if `-exit` or `-c exit` argument is provided

@kcgen Unfortunately, mixing `-c` with `-conf` and combining this into one autoexec section would be complicated and require further rework; I consider this out of scope of this PR. I have checked behavior in commit e2af38fc7 (the last one before my previous autoexec rework) and tried to implement behavior as it was back then.

2. If `-securemode` argument is provided in the command line, keyboard input is blocked until (`Z:\CONFIG.COM -securemode` is executed; I'm afraid that malicious user might attempt to terminate `Z:\AUTOEXEC.BAT` execution by sending CTRL+C early enough (it is not currently handled, but I expect at some point will be), to gain access to unrestricted mode.


kcgen:
The `[autoexec]` sections join or overwrite each other (depending on the `autoexec_section` setting), in order, starting the with primary conf to  the local conf and finally any number of `-conf custom.conf` specified on the CLI. 

For example, given three local confs as follows:

- `dosbox.conf`:

    ```ini
    [autoexec]
    @echo 1
    ```

- `3.conf`:

    ```ini
    [autoexec]
    @echo 3
    ```

- `5.conf`:

    ```ini
    [autoexec]
    @echo 5
    ```

The command: **`dosbox -conf 3.conf -conf 5.conf`** produces:

```
1
3
5
```

When it comes to `-c command`s they can be thought of as line-by-line [autoexec] sections as well.  

For several releases, the CLI commands have come first in the above order, however I think it makes sense to treat them in the same order, with the first `-c command` coming after (or joining onto) the end of the local `dosbox.conf`.

So interspersing `-c command`s into the above:

**`dosbox -c "@echo 2" -conf 3.conf -c "@echo 4" -conf 5.conf`**

Should produce:

```
1
2
3
4
5
```

The secure flag is meant to prevent commands inside the DOS environment from mucking with the mounts (and I suppose, doing harmful things.. like remounting the host's C:\WINDOWS\SYSTEM32\ and deleting everything in there). 

For groups like `archive.org` remotely or via VNS, it's meant to prevent random public users from trying to change the mounts as well (given they can probably get themselves to the C:\ or A:\ prompt).

I think as soon as we're done processing all arguments on the CLI, then `-secure` should go into effect.

johnnovak:
But you must enable securemode right at the start, otherwise one could still pass in a mont command on the CLI, no?

FeralChild64:
This `-securemode` switch dos not look like a reliable way to protect against malicious user - if he somehow manages to send CTRl+C to the emulator before we execute `Z:\CONFIG.COM -securemode`... the only way that prevents such an exploit from working is that we do not have a proper CTRL+C support :)
OK, I'll move the secure mode enabling before the `[autoexec]` content, additionally I'm going to make the `-securemode` switch to block any keyboard input until we execute `Z:\CONFIG.COM -securemode` - this should improve the security.

Regarding the commands and `[autoexec]` - OK, makes sense, I'll implement as @kcgen suggests.

kcgen (approved):
looks good, @FeralChild64.

To help test, do you have a zip of conf and bat files to experiment with these different settings?

kcgen:
> @kcgen Configuration is not really that relevant here - the most important are the command line parameters. Try different combinations of:
> 
>     * `dosbox -c command1 -c command2`
> 
>     * `dosbox -securemode`
> 
> 
> with:
> 
>     * `dosbox directory`
> 
>     * `dosbox command`
> 
>     * `dosbox file.img`
> 
>     * `dosbox file.iso`
> 
> 
> for example, `dosbox -c command1 -securemode file.iso /directory/`. When you see the command prompt, type `more Z:\AUTOEXEC.BAT` to check content of the generated `AUTOEXEC.BAT`.

Wow @FeralChild64 .. this is brilliant! It "just works" :-), and I cannot confuse it.

kcgen:
Will restart the CI tasks and will merge once the remainder are confirmed passing.

kcgen:
CI is :green_book: - time to merge :rocket:

Issue #2610, opened 2023-06-17
Center mouse cursor before releasing it doesn't work with windows scale of 200%


Max833:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

Newest dev build

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Center mouse cursor before releasing it (https://github.com/dosbox-staging/dosbox-staging/commit/bb294a0ed1a837b8763f25308c0f784e55d8732a) is a great idea and it works with 100% DPI scaling but not with 200%. The cursor isn't in the middle of the screen when DPI scaling is enabled and you release the mouse.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Select Start > Settings > System > Display and set the DPI scaling to 200%.
2. start dosbox and release the mouse

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

FeralChild64:
@Max883 Thanks for reporting this. Are you able to test whether https://github.com/dosbox-staging/dosbox-staging/tree/fc/mouse-centering-fix-1 fixes the issue for you? Unfortunately, I'm a Linux user, and I can't reproduce the problem using Wayland.

Max833:
It is working perfectly now! Thanks a lot!! =)

FeralChild64:
Not so fast with closing - this still has to go through review and get merged to the main branch :)

Issue #2603, opened 2023-06-15
Viewport sizing weirdness


johnnovak:
M1 MacBook, 3456 × 2234 Retina screen (so High DPI applies)

All tests performed with `-noprimaryconf`.

## Case 1 – Correct

```ini
[sdl]
windowresolution = desktop
```

<img width="1602" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/aa890078-580f-4997-a163-12e4f6eef426">

## Case 2 – Wrong

```ini
[sdl]
viewport_resolution = 100%
windowresolution = desktop
```

Switching to fullscreen and back resolves it. When first switching to fullscreen, I'm getting the same wrong results—the second time it's fine.

<img width="1602" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/b05a82d9-dc99-4ce4-92d8-fe28f8987dfc">

## Case 3 — Wrong

Wait, it gets worse!

```ini
[sdl]
viewport_resolution = 50%
windowresolution = desktop
```

<img width="1602" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/4c85b6ca-b692-405e-a201-307b8c1f2b44">

## Case 4 — Wrong

I could continue this all day, there's so many variations. In short, if `windowresolution` is larger than the available size (full desktop minus menubar, taskbar, etc.), then a different path gets triggered somehow that miscalculates the dimensions. `windowresultion = desktop` trigger this bad path too.

```ini
[sdl]
viewport_resolution = 1440x1080
windowresolution = 1440x1080
```

<img width="1552" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/e8904611-f604-4243-bb57-043a64456b8b">

---

I started investigating it with trace logging but gave up after an hour... I hope I'm wrong, but it appears to me that `calc_viewport` sometimes get passed physical dimensions, sometimes logical dimensions... If that's the case, this is madness and it's impressive it works as well as it does.

It's tempting to look the other way and ignore this and just merge @GranMinigun's SDL cleanup stuff after the release is out, but I don't think it's great to just leave this regression in...

Note how `calc_viewport: win_w: 1480` turns into `calc_viewport: win_w: 2980` after `calc_viewport` gets called repeatedly a few times... I'd bet good money on it that we're semi-randomly calling it with either physical or logical dimensions—well, both can't be right, that's for sure!
```
[...]

PR #2611, opened 2023-06-17
Fix mouse centering on high DPI displays with scaling enabled

FeralChild64:
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2610

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2338,12 +2338,21 @@ void GFX_SetMouseHint(const MouseHint hint_id)
 
 void GFX_CenterMouse()
 {
-	int current_width  = 0;
-	int current_height = 0;
-
 	assert(sdl.window);
-	SDL_GetWindowSize(sdl.window, &current_width, &current_height);
-	SDL_WarpMouseInWindow(sdl.window, current_width / 2, current_height / 2);
+
+	int width  = 0;
+	int height = 0;
+
+#if defined(WIN32)
+	const auto window_canvas_size = get_canvas_size(sdl.desktop.type);
+
+	width  = window_canvas_size.w;
+	height = window_canvas_size.h;
+#else
+	SDL_GetWindowSize(sdl.window, &width, &height);
+#endif
+
+	SDL_WarpMouseInWindow(sdl.window, width / 2, height / 2);
 }
 
 void GFX_SetMouseRawInput(const bool requested_raw_input)


src/gui/sdlmain.cpp
```diff
@@ -2341,9 +2341,15 @@ void GFX_CenterMouse()
...
+	auto get_middle = [](const int value) {
+		return lround((value * sdl.desktop.dpi_scale) / 2);
+	};
+
 	assert(sdl.window);
 	SDL_GetWindowSize(sdl.window, &current_width, &current_height);
-	SDL_WarpMouseInWindow(sdl.window, current_width / 2, current_height / 2);
+	SDL_WarpMouseInWindow(sdl.window,
+	                      get_middle(current_width),
+	                      get_middle(current_height));
```
kcgen, line 2352: `get_canvas_size()` might be a more robust way to get the logical size (and the divide its results by 2).

It's meant to let us avoid having to handle DPI in all the sdlmain code (I think; is that right, @GranMinigun ?).

GranMinigun: `get_canvas_size()` returns physical pixels, so no, `SDL_GetWindowSize()` is the way to go. But DPI scaling should not be applied here - mouse cursor coordinates are in client points, not physical pixels. This will likely break on macOS, and if what is done here works on Windows, then the correct way to proceed would be reporting the behaviour to the SDL team.

johnnovak: My 2c, I don't think you can ever be totally oblivious to DPI in the code, but that's not a problem. We just need to make that always explicit whether an argument/variable is a physical (actual physical pixel) coordinate or a logical (client) coordinate. And generally you should default to client coordinates, so if we are consistent about that, marking physical coords explicitly in the namings should de sufficient.

Otherwise we get into a mess like the weird issue I stumbled into the other day. I was really exhausted then so maybe I drew the wrong conclusions, but I was dead certain then that some functions get repeatedly called with both logical and physical coordinates, and that's how the bug manifests...
https://github.com/dosbox-staging/dosbox-staging/issues/2603

johnnovak: > This will likely break on macOS, and if what is done here works on Windows, then the correct way to proceed would be reporting the behaviour to the SDL team.

I agree with that, just pointing out we can ifdef it to not apply the division on Apple as a temp workaround. I'll test it soon.

johnnovak:
Yeah, so this needs more work. Results on macOS:

image


So the pointer appears in the bottom-right corner, which makes it turn into the resize pointer shape.

I've done some debugging:

image

image

SDL bug perhaps?

kcgen:
Running a debug build, an assertion is catching an issue at `capture.cpp:165`, when I hit: `Ctrl+F5` -> `Alt+F5` -> **2x** `Ctrl+F5`

```text
2023-06-17 20:15:36.404 | CAPTURE: Capturing upscaled image to 'capture/image0001.png'
2023-06-17 20:15:38.140 | CAPTURE: Capturing rendered image to 'capture/image0002-rendered.png'
2023-06-17 20:15:41.424 | CAPTURE: Capturing upscaled image to 'capture/image0003.png'

src/capture/capture.cpp:165: const char *capture_type_to_postfix(const CaptureType): Assertion `((void)"Unknown CaptureType", false)' failed.
```

src/gui/sdlmain.cpp
kcgen: Relaying from @GranMinigun :

> for now this workaround can be `#ifdef`ed (to Windows) to use `get_canvas_size()` on Windows.

FeralChild64: Added `#ifdef` to use `get_canvas_size` on Windows.

FeralChild64: Or, maybe the best solution would be to use SDL_GetRendererOutputSize? https://wiki.libsdl.org/SDL2/SDL_GetRendererOutputSize

Sorry, this is all very confusing to me. Why does anyone assume I need window size in some strange units? I just want size in pixels, sometimes a DPI (horizontal and vertical - I assume they might be different).

GranMinigun: Because one may have displays with different scaling factors applied, and the window is supposed to keep its relative size the same on both in such cases. Cursor should also preserve its speed, so that there's no sudden jump in sensitivity by merely moving it between displays. One display may be 4K, another Full HD, both may have similar physical size. Hence the need for DPI scaled coordinates (which Apple calls "client points"). HiDPI shenanigans, yes.

`SDL_GetRendererOutputSize()`, as hinted by its name, works only with SDL's renderer (`texture` output in DOSBox). It's wrapped by `get_canvas_size()` function.

src/gui/sdlmain.cpp
```diff
@@ -2338,12 +2338,20 @@ void GFX_SetMouseHint(const MouseHint hint_id)
...
+#if defined(WIN32)
+	const auto window_canvas_size = get_canvas_size(sdl.desktop.type);
+
+	width  = window_canvas_size.w;
+	height = window_canvas_size.h;
+#else
+	SDL_GetWindowSize(sdl.window, &width, &height);
+#endif
 
 	assert(sdl.window);
```
GranMinigun, line 2353: Please move the assert before ifdef, so that it is triggered before calling `SDL_GetWindowSize()`.

kcgen:
Confirmed working OK here on Linux w/ a slightly raised DPI, using both `output` opengl and texture.

src/gui/sdlmain.cpp
FeralChild64: OK, changed.

src/gui/sdlmain.cpp
FeralChild64: Cursor speed can depend on resolution and DPI. For the GUI scaling - when I was young, the system font size was used for this, it worked. I find the desktop scaling useless, it makes everything blurry.

Unfortunately, I don't know whether it is the `SDL_GetWindowSize` which is not consistent, or `SDL_WarpMouseInWindow`, or is it due to some capability flag in Windows EXE file - I only use Linux at home.

kcgen:
> @GranMinigun: You have to switch to Wayland session, yes, as DPI scaling concept exists only there.

PR is working great on Wayland with DPI scaling:

Screenshot_20230618_100331

kcgen:
Using `SDL_VIDEODRIVER=wayland`, the released cursor appears in the original place it was captured (so perhaps, mouse warp doesn't work there):

Screenshot_20230618_121914

This is perfect fine; if that's how Wayland does things, then it's probably true for anything that captures the mouse for exclusive control as well.

FeralChild64:
> Using `SDL_VIDEODRIVER=wayland`, the released cursor appears in the original place it was captured (so perhaps, mouse warp doesn't work there):

Really strange - works for me. SDL2 v2.26.5, Plasma 5.27.5, KDE frameworks 5.106.0, Wayland 1.22.0

kcgen:
> > Using `SDL_VIDEODRIVER=wayland`, the released cursor appears in the original place it was captured (so perhaps, mouse warp doesn't work there):
> 
> Really strange - works for me. SDL2 v2.26.5, Plasma 5.27.5, KDE frameworks 5.106.0, Wayland 1.22.0

SDL2 2.26.3, Plasma 5.27.4, KDE frameworks 5.104.0, and Wayland 1.21.0

All these packages are still improving and I guess there have been important mouse position improvements made in some of these between our versions. 

So long as the code on our side has the correct meaning, then we'll just let the user-space catch up eventually.

FeralChild64:
@kcgen Sorry, I was wrong - indeed, it gets released at the place where it was captured. I must have misspelled something during my previous test.

kcgen:
> @kcgen Sorry, I was wrong - indeed, it gets released at the place where it was captured. I must have misspelled something during my previous test.

Oh, interesting! Maybe this is going to be the intended behavior of SDL 2.x on Wayland (or it'll change in 3.x; looks like other open mouse position tickets are now scheduled to be addressed in the 3.x timeframe like: https://github.com/libsdl-org/SDL/issues/1836)

johnnovak:
So ifdef this out for Apple and possibly raise an SDL bug report?

kcgen:
> So ifdef this out for Apple and possibly raise an SDL bug report?

yup - @FeralChild64's got the change isolated to `WIN32` now.

I was checking in SDL2's repo and noticed this one with a possible solution, https://github.com/libsdl-org/SDL/issues/7245, however that's for SDL3 :-)!


The 2.28 release (already in RC1) is their [final release for the 2.x series](https://github.com/libsdl-org/SDL/releases), so I suspect we've missed the boat.  This is probably a OK work-around until we migrate to SD3.

johnnovak:
> > So ifdef this out for Apple and possibly raise an SDL bug report?
> 
> yup - @FeralChild64's got the change isolated to `WIN32` now.
> 
> I was checking in SDL2's repo and noticed this one with a possible solution, [libsdl-org/SDL#7245](https://github.com/libsdl-org/SDL/issues/7245), however that's for SDL3 :-)!
> 
> The 2.28 release (already in RC1) is their [final release for the 2.x series](https://github.com/libsdl-org/SDL/releases), so I suspect we've missed the boat. Given it's a one-liner, not a big deal to carry until we can migrate to SDL3.
> 
> (But maybe it's worth reporting anyway; I don't feel like it though.. given they are migrating to SDL3 and have punted many DPI issues into 3.x already).

So we're planning to migrate SDL3 soon?

kcgen:
> So we're planning to migrate SDL3 soon?

Not us.. (we still have to merge @GranMinigun 's branch post 0.81, then wait for things to stabilize in SDL 3, which might be some years). 

But SDL devs are migrating! They've finalized the 2.x feature set and 2.28 will be their last release in the 2mx series. 

Also, they have punted very old 2.x DPI-related issues (ones that have been open since 2016) into the 3.x series.. never to be fixed in 2.x. 

It's a bit line someone reporting an issue against Staging 0.79. we won't be releasing 0.79.2..  likewise, I don't think we can expect fresh 2.x bugs to be fixed. Instead, we'll just wait for 3.x, which reworks DPI handling from the ground up.

kcgen (approved):

kcgen:
Nice quick fix @FeralChild64.

weirddan455:
For what it's worth, there's going to be an SDL2 -> SDL3 compatibility layer (similar to what they did for SDL1 -> SDL2 except this time it's going to available from day 1) so we're not necessarily "stuck" with any SDL bugs.  When SDL 3 releases, we should just be able to link to the compat library and benefit from bug fixes (or have end users link dynamically, I believe it even works with the same binary).

GranMinigun:
@kcgen, bugs are meant to be reported, and SDL 2.28 will be supported for some time and will receive patch releases (2.28.x). Whether that specific behaviour will be addressed in SDL 2 or 3 is up to them, of course, but unless there's an open bug report already, we ought to let them know about inconsistency in behaviour between different platforms. We all want SDL to be better, don't we?

johnnovak:
> @kcgen, bugs are meant to be reported, and SDL 2.28 will be supported for some time and will receive patch releases (2.28.x). Whether that specific behaviour will be addressed in SDL 2 or 3 is up to them, of course, but unless there's an open bug report already, we ought to let them know about inconsistency in behaviour between different platforms. We all want SDL to be better, don't we?

I agree, this should be reported... by someone else than me as I'm busy with other stuff :ducks:

kcgen:
Yup, I agree too.

( but I also won't be opening it because I'm not well versed in this area, and there's too much going on here to break out a side project and make a stand alone repro case, check if  latest SDL 2.28 rc1 on a Win VM is affected, etc).

I'm not about to ask anyone to do it, but if it gets done, great and thank you.

FeralChild64:
@GranMinigun Thank you. It would be hard for me to provide support for the SDL team, as I’m a Linux-only user.

kcgen:
@GranMinigun , ticket and repro code are so good, slouken was able to fix it immediately. :exploding_head: :1st_place_medal:

PR #2612, opened 2023-06-18
Fix a couple aspect ratios

### LINKED ISSUE #2589: Zak McKracken Tandy composite mode has wrong aspect ratio
image

### LINKED ISSUE #2605: Emulator crash after selecting 200-column text mode in Necromancer's DOS Navigator

0.81.0-alpha, GIT 7f4f4e48b3b1a8cb3a1fa8547c1f821f32cd943b (current main branch)

Linux x86_64

Gentoo, GCC 13.1.1 20230527

### Relevant hardware info

AMD64

DOSBox Staging crashes when selecting 200-column text mode in Necromancer's DOS Navigator. This is a regression - the problem does not happen in 0.80.1.

1. Start _Necromancer's DOS Navigator_ (I have version v3.00.0007v, DPMI32)
2. Select _"burger" menu_ -> _Video modes_ -> _Choose extended video mode..._ -> _Q200x25x8_ (but it crashes with other 200 column modes too)
3. DOSBox Staging crashes

http://ndn.muxe.com/download/file/ndn_3_00_0007_d32.zip

### Your configuration

```ini
[sdl]
fullscreen          = false
display             = 0
fullresolution      = desktop
windowresolution    = 1280x960
viewport_resolution = fit
window_position     = auto
window_decorations  = true
transparency        = 0
host_rate           = auto
vsync               = true
vsync_skip          = 7000
presentation_mode   = auto
output              = opengl
texture_renderer    = auto
waitonerror         = true
priority            = auto auto
mute_when_inactive  = false
pause_when_inactive = false
mapperfile          = mapper-sdl2-0.80.0-alpha.map
screensaver         = auto

[dosbox]
language                    = en
machine                     = svga_s3
captures                    = capture
memsize                     = 31
mcb_fault_strategy          = repair
vmemsize                    = 4
dos_rate                    = default
vesa_modes                  = compatible
vga_8dot_font               = true
speed_mods                  = true
autoexec_section            = join
automount                   = true
startup_verbosity           = quiet
allow_write_protected_files = true

[render]
aspect             = true
integer_scaling    = off
monochrome_palette = green
cga_colors         = default
glshader           = 

[cpu]
core      = dynamic
cputype   = pentium_slow
cycles    = 100000
cycleup   = 1
cycledown = 1
```

### Provide a Log

```text
2023-06-15 16:37:20.158 | DISPLAY: Text 640x480 16-colour (mode 03h) at 67.433 Hz VFR, scaled to 1280x960 with 1:1 pixel aspect ratio
2023-06-15 16:37:27.053 | DISPLAY: Text 1600x400 16-colour (mode 03h) at 17.406 Hz VFR, scaled to 1280x879 with 1:2.75 pixel aspect ratio

```

- [X] Yes, I agree.

kcgen:
Commits are relatively stand-alone, suggest reviewing one-by-one.
- fixes #2605 
- fixes #2589

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -2161,10 +2161,19 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 	    (vga.draw.bpp != bpp) || fps_changed) {
 		VGA_KillDrawing();
 
-		vga.draw.width = width;
-		vga.draw.height = height;
-		vga.draw.doublewidth = doublewidth;
+		if (width > SCALER_MAXWIDTH || height > SCALER_MAXHEIGHT) {
+			LOG_ERR("VGA: The calculated video resolution %ux%u will be limited to the maximum of %ux%u",
+			        width,
+			        height,
+			        SCALER_MAXWIDTH,
+			        SCALER_MAXHEIGHT);
+		}
+		vga.draw.width  = std::min(width, static_cast<uint32_t>(SCALER_MAXWIDTH));
+		vga.draw.height = std::min(height, static_cast<uint32_t>(SCALER_MAXHEIGHT));
+
+		vga.draw.doublewidth  = doublewidth;
 		vga.draw.doubleheight = doubleheight;
+
 		vga.draw.aspect_ratio = aspect_ratio;
 		vga.draw.bpp = bpp;
 		if (doubleheight) vga.draw.lines_scaled=2;
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -2194,10 +2194,10 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 			               doublewidth, doubleheight);
 	} else {
 		// Always log mode changes at a minimum
-		static auto previous_mode = CurMode->mode;
-		if (CurMode->mode != previous_mode) {
+		static decltype(CurMode) previous_mode = {};
+		if (CurMode != previous_mode) {
 			GFX_LogDisplayProperties();
-			previous_mode = CurMode->mode;
+			previous_mode = CurMode;
 		}
 	}
 }
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -2042,7 +2042,6 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 		}
 		break;
 	case M_TANDY4:
-		aspect_ratio = 1.2;
 		doubleheight = true;
 		if (machine == MCH_TANDY)
 			doublewidth = (vga.tandy.mode_control & 0b10000) == 0b00000;
@@ -2051,6 +2050,8 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 
 		vga.draw.blocks = width * 2;
 		width = vga.draw.blocks * 4;
+		aspect_ratio    = width * 3.0 / (height * 4.0);
+
 		if ((machine == MCH_TANDY && (vga.tandy.gfx_control & 0b01000)) ||
 		    (machine == MCH_PCJR && (vga.tandy.mode_control == 0b01011)))
 			VGA_DrawLine = VGA_Draw_2BPPHiRes_Line;
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1978,8 +1978,14 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 		vga.draw.blocks = width;
 		doublewidth = vga.seq.clocking_mode.is_pixel_doubling;
 		width *= vga.draw.pixels_per_character;
+
+		aspect_ratio = (static_cast<double>(width) /
+		                static_cast<double>(height)) *
+		               (3.0 / 4.0);
+
 		aspect_ratio *= vga.draw.pixels_per_character /
 		                static_cast<double>(PixelsPerChar::Eight);
+
 		break;
 	case M_HERC_GFX:
 		vga.draw.blocks=width*2;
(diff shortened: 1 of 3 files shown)


src/hardware/vga_draw.cpp
```diff
@@ -2161,6 +2164,13 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 	    (vga.draw.bpp != bpp) || fps_changed) {
 		VGA_KillDrawing();
 
+		if (width > SCALER_MAXWIDTH || height > SCALER_MAXHEIGHT) {
```
FeralChild64, line 2167: Maybe it would be better to print a warning log and simply clip the values to the maximum? Terminating the emulation seems rather drastic to me.

johnnovak: I agree. I think asserts are fine, but in general we should be as lenient as possible, and err on the side of "not crashing" whenever we can. Logging a warning and clipping would be better, I think.

johnnovak:
Looks okay at first sight, I'll test tomorrow!

src/hardware/vga_draw.cpp
kcgen: Yup - this is really the equivalent to "out of range" error:

images

That's better - Will clamp it and log strongly.

But if I honestly expected even bigger resolutions beyond this - then absolutely: let's just accommodate them here too; but I want the data to lead me to that value instead of adding random contingency.

I'll try to work backward through the tweaked resolution calculations and see if there's a limit.

kcgen: Updated to log an error and limit the width to 1630 adapter pixels wide.

Going back through the registers, setting the overscan bit (b1) combined with setting all eight bits in the Horizontal Display End Register will produce a horizontal resolution of 4600 pixels (`1 + (255 | (0b10 << 7)) * 9`), which in turn will produce a very slow refresh rate of ~5 Hz (if the vertical is still 1200 lines). 

Well, I guess we'll wait for reports of needing more, if such software exists!

johnnovak: Yeah, log a warning and let it crash/misbehave in the emulation but prevent DOSBox from crashing should be our philosophy 👍🏻 We'll still get report, just not random crash report 😄

johnnovak:
I've tested Zak, and when switching from the Tandy TGA mode to composite, the CGA composite aspect ratio is correct. But when pressing the composite hotkey again, the game switches to the 640x200 hi-res Tandy mode, and the aspect ratio is still incorrect there. 

Plus the logging is incorrect in different ways across the board (I thought we got this right earlier?)

## Tandy mode (at startup)

This looks correct, but gets logged wrong (should log 320x200 for TGA mode 09h):

```
DISPLAY: Tandy 160x100 16-colour (mode 09h) at 59.923 Hz VFR, scaled to 2205x1654 with 1:1.2 pixel aspect ratio
```

image

## CGA composite (after pressing the composite hotkey once)

This looks correct but gets logged incorrectly (640x200 is fine, but that's the TGA mode 0Ah — check `int10_modes.cpp` of [this reference](https://www.minuszerodegrees.net/video/bios_video_modes.txt)):

```
DISPLAY: CGA 640x200 composite (mode 09h) at 59.923 Hz VFR, scaled to 2205x1654 with 1:2.4 pixel aspect ratio
```

image


## Tandy 640x200 hi-res (after pressing the composite hotkey once more)

This looks wrong and gets logged incorrectly (that's the 0Ah CGA-equivalent hi-res mode)

```
DISPLAY: Tandy 640x200 4-colour (mode 09h) at 59.923 Hz VFR, scaled to 2206x827 with 1:1.2 pixel aspect ratio
```

image

kcgen:
Can you double check that you've got the latest pushes? I removed the logging changes earlier.

For Tandy high res, didn't realize I could keep stepping through them with the composite hotkey. Will add a fix tomorrow. Thanks for catching it!

kcgen:
@johnnovak , pushed and update - aspect is now OK in all the composite tandy, PC Jr, and CGA games I could find.

the logging comes from the `VideoModeBlock`'s mode and types. There are no overrides for the the actual value (like 9h); so not sure.  It could be that some of the labels are wrong? See the int10_modes.cpp `**ModeList_OTHER**` block at line 313.

johnnovak:
> @johnnovak , pushed and update - aspect is now OK in all the composite tandy, PC Jr, and CGA games I could find.

Yep, the aspect is fixed now in Zak! Great 🎉 

> the logging comes from the `VideoModeBlock`'s mode and types. There are no overrides for the the actual value (like 9h); so not sure. It could be that some of the labels are wrong? See the int10_modes.cpp **`ModeList_OTHER`** block at line 313.

Yeah, don't worry about that. I'll give that `src.pixel.width/height` improvement a go after this went in, which should fix it.

src/hardware/vga_draw.cpp
```diff
@@ -2161,10 +2165,19 @@ void VGA_SetupDrawing(uint32_t /*val*/)
...
-		vga.draw.doublewidth = doublewidth;
+		if (width > SCALER_MAXWIDTH || height > SCALER_MAXHEIGHT) {
+			LOG_ERR("VGA: The calculated video resolution %ux%u will be limited to the maximum of %ux%u",
+			        width,
+			        height,
+			        SCALER_MAXWIDTH,
+			        SCALER_MAXHEIGHT);
+		}
+		vga.draw.width  = std::min(width, static_cast<uint32_t>(SCALER_MAXWIDTH));
+		vga.draw.height = std::min(height, static_cast<uint32_t>(SCALER_MAXHEIGHT));
```
johnnovak, line 2176: Nice 👍🏻

johnnovak (approved):
Nice work @kcgen. 🚢 it! 😄

src/hardware/vga_draw.cpp
kcgen: Thanks for all the testing @johnnovak !

kcgen:
> Yeah, don't worry about that. I'll give that src.pixel.width/height improvement a go after this went in, which should fix it (logging the original source res, not the mode... that's a bit of a mystery).

Right on! Thanks going to get to the bottom of it! 🚀

Issue #233, opened 2020-03-21
Distorted sound in Worms


numberZero:
In Worms sound is distorted, while it is OK in 0.74. Also in 0.74 there are two “clicks” on Worms start, which are not there in Staging. Maybe the signedness is wrong somewhere?

dreamer:
You need to be more precise - is distorted sound only Worms voices or also music and video cutscenes?

I know about single audio distortion bug in Worms we inherited from SVN: [SF #510](https://sourceforge.net/p/dosbox/bugs/510/) - do you experience this one?

> while it is OK in 0.74

Oh! This is important. Do you mean 0.74, 0.74-2 or 0.74-3? When testing 0.74-3 I *think* I hear this distortion, but I'm not sure; I haven't tested older versions yet.

> Also in 0.74 there are two “clicks” on Worms start

IIRC this was some kind of harmless problem with CD-DA implementation in upstream project; this issue disappeared when we merged-in new CD-DA emulation stack by @krcroft.

numberZero:
The issue is with voices. It is actually mush like SF#510 you referenced. Changing `sbtype` from `sb16` to `sb1` fixes the distortion. But on svn-0.74-3, it works with `sb16` as well, in stereo (position-dependent).

My system has:
 * JACK as the sound server, running at 48kHz
 * Its ALSA plugin as the default output
 * DOSBox Staging d0da4341 and SVN 0.74-3

Note that the sound is chopped in capture files as well, not only in real output.

Here is relevant part of the config:
```
[mixer]
nosound=false
rate=48000
#blocksize=1024
blocksize=4096
prebuffer=20

[sblaster]
sbtype=sb16
sbbase=220
irq=7
dma=1
hdma=5
sbmixer=true
oplmode=auto
oplemu=default
oplrate=48000
```
and output:
```
MIXER: Got different values from SDL: freq 48000, blocksize 4096
ALSA:Can't subscribe to MIDI port (65:0) nor (17:0)
No working midi device found/selected! Please check your settings and/or compilation environment.
MIDI: Opened device:none
```

P.S. CD-DA is perfect in Staging (and totally broken in SVN).

krcroft:
[...]

PR #2614, opened 2023-06-18
Refill the SB audio channel when DMA reaches terminal count (#233)

kcgen:
When the active DMA channel signals its reached terminal count, there is a balance of now transferred (from the game into the memory location) samples waiting to be consumed, but that haven't been read yet by the SB.

This commit asks the audio channel to get caught up to the current time: that is, to the same state when the DMA transfer completed. This fill up step will read a balance of remaining DMA samples, and queue them for playback.

In 'Worms', logging shows the pending tail of DMA data is anywhere from 4 to 30 bytes. Filling up the channel to the current time draws in enough DMA samples to complete the transfer on the SB side.

- Fixes https://github.com/dosbox-staging/dosbox-staging/issues/233

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -588,8 +588,12 @@ static void DSP_FlushData()
 static double last_dma_callback = 0.0;
 
 static void DSP_DMA_CallBack(DmaChannel * chan, DMAEvent event) {
-	if (chan!=sb.dma.chan || event==DMA_REACHED_TC) return;
-	else if (event==DMA_MASKED) {
+	assert(chan && chan == sb.dma.chan);
+
+	if (event == DMA_REACHED_TC) {
+		assert(sb.mode == MODE_DMA);
+		sb.chan->FillUp();
+	} else if (event == DMA_MASKED) {
 		if (sb.mode==MODE_DMA) {
 			//Catch up to current time, but don't generate an IRQ!
 			//Fixes problems with later sci games.


src/hardware/sblaster.cpp
```diff
@@ -588,8 +588,12 @@ static void DSP_FlushData()
 static double last_dma_callback = 0.0;
 
 static void DSP_DMA_CallBack(DmaChannel * chan, DMAEvent event) {
-	if (chan!=sb.dma.chan || event==DMA_REACHED_TC) return;
-	else if (event==DMA_MASKED) {
+	assert(chan && chan == sb.dma.chan);
+
```
johnnovak, line 592: If this fails that can only mean a DOSBox side programming error, right? The more I think about it, invalid behaviour that can happen on real hardware too should never be asserted but handled.

kcgen: Yup, this is just coding sanity check and to document expectations (which should be 100% guaranteed), to make sure order-of-operations are being handled properly.

Right; asserts should never be used to validate user-side data, and instead, they should represent expected truths about the state of the code (and can act to catch broken program flow before it derails further into something harder to debug).

kcgen: Yeah, the original devs didn't use asserts, so when reading/refactoring code it's hard to separate expectations versus nominal pathway handling.

In this case, DMA callbacks are called by DMA channel objects, and pass themselves into the function as the `* chan` pointer.  And there's no way a `nullptr` DMA channel will making this callback.

johnnovak:
Before this fix, was there any problem with the audio in Worms? Does this mean those samples were effectively dropped and never played?

I don't know if the current sample dropping is an attempt to accurately reproduce what the SB hardware does, or if it's a problem introduced by emulation. If this not fully reading the data was put there for a reason, we might break play routines that rely on this bug?

I don't know the answer, just saying to exercise caution 😄

kcgen:
> Before this fix, was there any problem with the audio in Worms? Does this mean those samples were effectively dropped and never played?

Yes, it looks like worms has audible samples right up to the last frame in a given PCM sequence, and depending on the card type, there were just a couple audio frames periodically left behind (in the way Worms uses it). Where as when SB1 is used, all of the transferred frames make it through. 

> I don't know if the current sample dropping is an attempt to accurately reproduce what the SB hardware does, or if it's a problem introduced by emulation. 

> If this not fully reading the data was put there for a reason, we might break play routines that rely on this bug?

> I don't know the answer, just saying to exercise caution smile

Yes to all - but it seems like a sane approach (famous last words!).

The DMA controller signals to the card that it's reached terminal count, but at the moment of that signalling, there might still be a balance of data left to consume. 

That's why we really need to get this in for testing - if this causes any regressions, it'll be reverted straight away.

johnnovak:
> That's why we really need to get this in for testing - if this causes any regressions, it'll be reverted straight away.

Yeah, based on what I read about it, DMA handling on various SB models was a sure mess. The 38824 different (and differently buggy and incompatible with earlier models) SB16 variants did not help either... I guess many games have just broken SB support in all sorts of different ways because SB support was a constantly moving target... what a nightmare for those poor people 😅 I guess we can try to emulate "idealised" and bug-free SB models that work as they're supposed to, after all, and hope for the best 😄

johnnovak (approved):

PR #2615, opened 2023-06-18
Implement persistent command line history

### LINKED ISSUE #2559: Persistent command line history
It would be cool to have a persistent command line history that survives DOSBox restarts.

weirddan455:
Fixes #2559 

I mimicked the behavior of what I observed from bash_history on my system.  I set a cap of 500 lines on the history file so it won't grow endlessly (same as what bash appears to use).  The shell simply reads from the file on startup and writes back out to it on close.

I thought about adding this to constructor and destructor of the DOS_Shell class but in testing I noticed that when COMMAND.COM gets created, it uses an instance of DOS_Shell and was clobbering the file so I implemented this only for `first_shell` instead, which from testing works fine.

### DIFF
--- a/include/shell.h
+++ b/include/shell.h
@@ -111,6 +111,8 @@ public:
 	bool CheckConfig(char* cmd_in,char*line);
 	/* Internal utilities for testing */
 	virtual bool execute_shell_cmd(char *name, char *arguments);
+	void ReadCommandHistory();
+	void WriteCommandHistory();
 
 	/* Commands */
 	void CMD_HELP(char * args);
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1222,6 +1222,10 @@ void DOSBOX_Init()
 	        "tab-separated format, used by SETVER.EXE as a persistent storage\n"
 	        "(empty by default).");
 
+	pstring = secprop->Add_path("command_history_file", only_at_start, "cmd_history.txt");
+	pstring->Set_help(
+		"File containing persistent command line history ('cmd_history.txt' by default).");
+
 	// Mscdex
 	secprop->AddInitFunction(&MSCDEX_Init);
 	secprop->AddInitFunction(&DRIVES_Init);
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -463,6 +463,9 @@ static std_fs::path get_cmd_history_path()
 
 void DOS_Shell::ReadCommandHistory()
 {
+	if (control->SecureMode()) {
+		return;
+	}
 	std::ifstream history_file(get_cmd_history_path());
 	if (history_file) {
 		std::string line;
@@ -478,6 +481,9 @@ void DOS_Shell::ReadCommandHistory()
 
 void DOS_Shell::WriteCommandHistory()
 {
+	if (control->SecureMode()) {
+		return;
+	}
 	std_fs::path history_path = get_cmd_history_path();
 	std::ofstream history_file(history_path);
 	if (!history_file) {
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -19,6 +19,7 @@
 
 #include "shell.h"
 
+#include <fstream>
 #include <list>
 #include <memory>
 #include <stdarg.h>
@@ -36,6 +37,9 @@
 #include "string_utils.h"
 #include "support.h"
 #include "timer.h"
+
+constexpr int HistoryMaxLineSize = 256;
+constexpr int HistoryMaxNumLines = 500;
 
 callback_number_t call_shellstop = 0;
 /* Larger scope so shell_del autoexec can use it to
@@ -450,6 +454,58 @@ void DOS_Shell::Run()
 void DOS_Shell::SyntaxError()
 {
 	WriteOut(MSG_Get("SHELL_SYNTAX_ERROR"));
+}
+
+static std_fs::path get_cmd_history_path()
+{
+	return get_platform_config_dir() / "cmd_history.txt";
+}
+
+void DOS_Shell::ReadCommandHistory()
+{
+	std::ifstream history_file(get_cmd_history_path());
+	if (history_file) {
+		std::string line;
+		while (getline(history_file, line)) {
+			trim(line);
+			auto len = line.length();
+			if (len > 0 && len <= HistoryMaxLineSize) {
+				history.emplace_back(std::move(line));
+			}
+		}
+	}
+}
+
+void DOS_Shell::WriteCommandHistory()
+{
+	std_fs::path history_path = get_cmd_history_path();
+	std::ofstream history_file(history_path);
+	if (!history_file) {
+		LOG_WARNING("SHELL: Unable to write command history to file: '%s'", history_path.string().c_str());
+		return;
+	}
+	std::vector<std::string> trimmed_history;
+	trimmed_history.reserve(history.size());
+	for (std::string str : history) {
+		trim(str);
+		auto len = str.length();
+		if (len > 0 && len <= HistoryMaxLineSize) {
+			trimmed_history.emplace_back(std::move(str));
+		}
+	}
+	// Remove "exit" from the history if it is the last command entered
+	if (!trimmed_history.empty()) {
+		std::string last = trimmed_history.back();
+		lowcase(last);
+		if (last == "exit") {
+			trimmed_history.pop_back();
+		}
+	}
+	int size = static_cast<int>(trimmed_history.size());
+	int start = std::max(0, size - HistoryMaxNumLines);
+	for (int i = start; i < size; ++i) {
+		history_file << trimmed_history[i] << std::endl;
+	}
 }
 
 extern int64_t ticks_at_program_launch;
@@ -1268,7 +1324,9 @@ void SHELL_Init() {
 	// first_shell is only setup here, so may as well invoke
 	// it's constructor directly
 	first_shell = new DOS_Shell;
+	first_shell->ReadCommandHistory();
 	first_shell->Run();
+	first_shell->WriteCommandHistory();
 	delete first_shell;
 	first_shell = nullptr; // Make clear that it shouldn't be used anymore
 }
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -456,9 +456,16 @@ void DOS_Shell::SyntaxError()
 	WriteOut(MSG_Get("SHELL_SYNTAX_ERROR"));
 }
 
-static std_fs::path get_cmd_history_path()
+static std::string get_command_history_path()
 {
-	return get_platform_config_dir() / "cmd_history.txt";
+	const auto section = static_cast<Section_prop *>(control->GetSection("dos"));
+	if (section) {
+		const auto path = section->Get_path("command_history_file");
+		if (path) {
+			return path->realpath;
+		}
+	}
+	return {};
 }
 
 void DOS_Shell::ReadCommandHistory()
@@ -466,7 +473,11 @@ void DOS_Shell::ReadCommandHistory()
 	if (control->SecureMode()) {
 		return;
 	}
-	std::ifstream history_file(get_cmd_history_path());
+	std::string history_path = get_command_history_path();
+	if (history_path.empty()) {
+		return;
+	}
+	std::ifstream history_file(history_path);
 	if (history_file) {
 		std::string line;
 		while (getline(history_file, line)) {
@@ -484,10 +495,13 @@ void DOS_Shell::WriteCommandHistory()
 	if (control->SecureMode()) {
 		return;
 	}
-	std_fs::path history_path = get_cmd_history_path();
+	std::string history_path = get_command_history_path();
+	if (history_path.empty()) {
+		return;
+	}
 	std::ofstream history_file(history_path);
 	if (!history_file) {
-		LOG_WARNING("SHELL: Unable to write command history to file: '%s'", history_path.string().c_str());
+		LOG_WARNING("SHELL: Unable to write command history to file: '%s'", history_path.c_str());
 		return;
 	}
 	std::vector<std::string> trimmed_history;


johnnovak:
Nice to see you picking up tickets from the backlog @weirddan455 ! 🎉 I'll test this a bit later today. The code looks nice and clean! 👍🏻

include/shell.h
```diff
@@ -37,6 +37,7 @@
 #define CMD_MAXLINE 4096
 #define CMD_MAXCMDS 20
 #define CMD_OLDSIZE 4096
+#define CMD_MAX_HISTORY_FILE_LINES 500
```
johnnovak, line 40: Use `constexpr` instead and UpperCamelCase for constants is the preferred standard for new stuff.

src/shell/shell.cpp
```diff
@@ -452,6 +453,49 @@ void DOS_Shell::SyntaxError()
...
+		}
+	}
+}
+
+void DOS_Shell::WriteCommandHistory()
+{
+	std_fs::path history_path = get_cmd_history_path();
+	std::ofstream history_file(history_path);
+	if (!history_file) {
+		LOG_WARNING("SHELL: Unable to write command history to file: %s", history_path.string().c_str());
```
johnnovak, line 479: Enclose filenames in single quotes in log messages.

src/shell/shell.cpp
```diff
@@ -452,6 +453,49 @@ void DOS_Shell::SyntaxError()
...
+	for (std::string& str : history) {
+		if (!str.empty()) {
+			trimmed_history.push_back(str);
+		}
+	}
+	size_t start = 0;
+	size_t size = trimmed_history.size();
+	if (size > CMD_MAX_HISTORY_FILE_LINES) {
+		start = size - CMD_MAX_HISTORY_FILE_LINES;
+	}
```
johnnovak, line 493: minor: Just use `std::max` or `clamp` from `math_utils.h`?

src/shell/shell.cpp
```diff
@@ -452,6 +453,49 @@ void DOS_Shell::SyntaxError()
...
+	for (size_t i = start; i < size; i++) {
```
johnnovak, line 494: We do `++i` by convention.

src/shell/shell.cpp
```diff
@@ -452,6 +453,49 @@ void DOS_Shell::SyntaxError()
...
+		history_file << trimmed_history[i] << std::endl;
+	}
```
johnnovak, line 496: Not sure about the usage of C++ streams. @kcgen , @dreamer ?

johnnovak:
Shall we have an option perhaps that enables/disables the feature? Maybe it's fine to have it always enabled, but I'm not 100% sure. @kcgen ? @FeralChild64 ?

johnnovak:
I've just tested it, works very well @weirddan455 ! 🚀

FeralChild64:
@johnnovak IMHO we definitely need a configuration option to disable this. I propose to introduce similar configuration option to `setver_table_file` - empty file = no persistent history.

I also wonder whether history should be stored in external file when secure mode is enabled (`control->SecureMode()` call to check for this condition).

include/shell.h
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Sure.  I had to switch to signed int for that to work but I don't think this vector will ever be big enough to cause an overflow.

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: I saw existing code in `src/misc/setup.cpp` was using C++ streams to parse the config file so that along with the fact that this is already using `std::string`, I thought C++ streams would be the simplest solution.

The output would be simple enough to switch to C-style file I/O.  Input is a little trickier since there's not a good "getline" function in C.  "getline" itself is POSIX only so it would need to use `fgets` alongside a temporary `char[] buffer` or something.

Just let me know what you prefer.

MasterO2:
Can the command line history also be cleared by the end user as well?

johnnovak:
> @johnnovak IMHO we definitely need a configuration option to disable this. I propose to introduce similar configuration option to `setver_table_file` - empty file = no persistent history.
> 
> I also wonder whether history should be stored in external file when secure mode is enabled (`control->SecureMode()` call to check for this condition).

Yeah I was thinking of the security aspect too. Wouldn't it be best to fully hard disable the feature in secure mode?

weirddan455:
It's disabled now for secure mode and there's now a config option.  I set it to enabled by default but if you think it should be disabled by default let me know.

src/dosbox.cpp
```diff
@@ -1222,6 +1222,10 @@ void DOSBOX_Init()
 	        "tab-separated format, used by SETVER.EXE as a persistent storage\n"
 	        "(empty by default).");
 
+	pstring = secprop->Add_path("cmd_history_file", only_at_start, "cmd_history.txt");
+	pstring->Set_help(
+		"File containing persistent command line history ('cmd_history.txt' by default)");
```
johnnovak, line 1227: minor: missing dot at the end of the sentence.

johnnovak: "File where the persistent command

johnnovak: Maybe call it `command_line_history_file`?

src/shell/shell.cpp
```diff
@@ -452,6 +453,67 @@ void DOS_Shell::SyntaxError()
 	WriteOut(MSG_Get("SHELL_SYNTAX_ERROR"));
 }
 
+static std::string get_cmd_history_path()
```
johnnovak, line 456: `get_command_history_path` would be better.

src/shell/shell.cpp
```diff
@@ -452,6 +453,67 @@ void DOS_Shell::SyntaxError()
...
+		}
+	}
+	return {};
+}
+
+void DOS_Shell::ReadCommandHistory()
+{
+	if (control->SecureMode()) {
+		return;
+	}
```
johnnovak, line 472: Nice 🎉

src/shell/shell.cpp
```diff
@@ -452,6 +453,67 @@ void DOS_Shell::SyntaxError()
...
+		}
+	}
+}
+
+void DOS_Shell::WriteCommandHistory()
+{
+	constexpr int MaxHistoryFileLines = 500;
+	if (control->SecureMode()) {
+		return;
+	}
```
johnnovak, line 493: 👍🏻

src/shell/shell.cpp
johnnovak: @kcgen @dreamer ? ☝🏻 I'm 50/50.

johnnovak:
> It's disabled now for secure mode and there's now a config option. I set it to enabled by default but if you think it should be disabled by default let me know.

Great! I think enabled by default is the user-friendly option.

src/dosbox.cpp
weirddan455: Yeah that name looks better for the config setting.  Fixed.

src/shell/shell.cpp
weirddan455: Changed

weirddan455:
Did a force push with the above changes plus rebasing to latest master branch.  Hopefully that fixes the build error of the last commit.  Windows build was showing too many warnings (even though no warnings from the files changed in this PR).

johnnovak:
> Can the command line history also be cleared by the end user as well?

You can delete the external file 🤷🏻 I guess we could add some built-in command to delete the history. But let's not get carried away and overload the PR; those should be treated as future enhancements after the basic, minimally useful feature is in.

johnnovak (approved):
This looks fine from my POV. @kcgen @dreamer can complain about the use of streams if they dislike it, I don't mind it too much and it's not used excessively.

Great enhancement @weirddan455, thank you 🎉

src/shell/shell.cpp
```diff
@@ -452,6 +453,67 @@ void DOS_Shell::SyntaxError()
...
+	std::string history_path = get_command_history_path();
+	if (history_path.empty()) {
+		return;
+	}
+	std::ifstream history_file(history_path);
+	if (history_file) {
+		std::string line;
+		while (getline(history_file, line)) {
+			if (!line.empty()) {
+				history.push_back(line);
```
kcgen, line 482: can avoid copying the string:
```c++
history.emplace_back(std::move(line));
```

src/shell/shell.cpp
kcgen: Right; we're not gaining any benefit from having streaming operators processing disparate types (the 'if/ofstream implementation come with a lot of bloat where they are used).

The history is just a list of strings, so hopefully can avoid using streams.

kcgen: I wonder if the copy of the history (trimmed_history) can be avoided - I /think/ it programmatically guaranteed that no empty lines exist as part of the prompt parsing.

However, we can't trust the incoming file (users might be tampering with it or it might be corrupt), so when reading, we should strip off empty lines and same with line exceeding our prompt length (I'm not sure of that.. might have to poke into the shell code).

For the writing, portion (after the check-and-returns), maybe this would :

```c++
constexpr size_t MaxHistoryFileLines = 500;
const auto num_lines = std::min(history.size(), MaxHistoryFileLines);

// the first line we'll write, iterated in the loop
const auto line = strings.end() - num_lines;

while (line != strings.end()) {
    history_file.write(line->c_str(), line->length());
    history_file.write("\n", 1);
    ++line;
}

history_file.close();
```

(I'm not sure if the close() is needed.. perhaps that's done on scope closure anyway)

src/shell/shell.cpp
```diff
@@ -452,6 +453,67 @@ void DOS_Shell::SyntaxError()
...
+	}
+	std::string history_path = get_command_history_path();
+	if (history_path.empty()) {
+		return;
+	}
+	std::ifstream history_file(history_path);
+	if (history_file) {
+		std::string line;
+		while (getline(history_file, line)) {
+			if (!line.empty()) {
```
kcgen, line 481: maybe we should `trim` and then check if it's empty, just to blow away useless lines too.

`void trim(std::string &str, const char trim_chars[] = " \r\t\f\n");` (from string_utils.h)

I think the prompt has a length limit (should match DOS's own length limit), but I'm not sure what that is - might have to dig in the shell code.  We should ignore lines that exceed that.

src/shell/shell.cpp
```diff
@@ -1268,7 +1330,9 @@ void SHELL_Init() {
 	// first_shell is only setup here, so may as well invoke
 	// it's constructor directly
 	first_shell = new DOS_Shell;
+	first_shell->ReadCommandHistory();
 	first_shell->Run();
+	first_shell->WriteCommandHistory();
```
kcgen, line 1335: nice and tidy!

src/shell/shell.cpp
weirddan455: I only added `trimmed_history` in there after I tested and saw blank lines in the file.  It looks like items get added to the history before any parsing is done.

kcgen: Good catch, @weirddan455. I now understand the need for the trimmed history :-) (sorry!)

I'm thinking we only need the _indexes_ of the populated lines.

Here's one approach that can be dropped in after the ofstream check. 

```c++
auto get_populated_indices = [](const auto& lines, const auto max_indices) {
	std::stack<size_t> indices = {};
	auto reverse_i = lines.size();
	while (reverse_i-- != 0 && indices.size() < max_indices) {
		if (!lines[reverse_i].empty()) {
			indices.push(reverse_i);
		}
	}
	return indices;
};

constexpr size_t MaxHistoryFileLines = 500;
auto indices = get_populated_indices(history, MaxHistoryFileLines);

while (!indices.empty()) {
	const auto i = indices.top();
	indices.pop();

	history_file.write(history[i].c_str(), history[i].length());
	history_file.write("\n", 1);
}
```

It needs `#include <stack>` (for last-in-first-out behavior).

kcgen:
Also a general usability thought: I noticed that when I `Z:\> exit` cleanly, I'm always greeted to `exit` being the first in my history (which makes sense, but now I always need to "double-tap" up to get past the first exit.

So the questio is: if 'exit' is the last (more recently entered) command, should we skip writing it (or opposite: skip reading it)? To avoid this annoyance to skip past the first exit?

src/shell/shell.cpp
weirddan455: I mean, yeah that works.  I'm thinking the strings will be small enough that it probably won't matter much in practice.  If you have mostly strings less than 8 characters, storing a `size_t` would actually be more memory.

I think I could actually re-write it to avoid a 2nd data structure at all if you think that's something worth doing.  It would just need 2 iterations checking `str.empty()` through `history`.  One to find the starting index and then one that actually writes it out to the file.

kcgen: Yup; good point @weirddan455 - let's go w/ your existing implementation. Nice and clear, and zero concern about speed.

kcgen: Although, suggest using the `.write()` calls, if you can.

Torinde:
Just sharing an observation:
A related historisation item is the console output - modern computers are so fast that text which originally would've been visible to the user now goes by too quickly (e.g. when starting a game before it reaches the graphical part or batch installer, etc.)

Would be nice if DOSbox has a scroll-bar like CMD.EXE, but I assume that's much more difficult than the command history.

src/shell/shell.cpp
weirddan455: It's a variable length thing based on how many lines and columns the user's display is.  Probably larger than anything the original DOS could do based on that description (a single command can fill up most of the screen).

https://github.com/dosbox-staging/dosbox-staging/blob/58a10bd49eeba87ed356fffd04310ff7ab163542/src/shell/shell_misc.cpp#L375-L385

Anyway, agree with checking length and trimming away useless whitespace.  I'll take a look at that tomorrow probably.

johnnovak:
> Just sharing an observation: A related historisation item is the console output - modern computers are so fast that text which originally would've been visible to the user now goes by too quickly (e.g. when starting a game before it reaches the graphical part or batch installer, etc.)
> 
> Would be nice if DOSbox has a scroll-bar like CMD.EXE, but I assume that's much more difficult than the command history.

It's a bit more difficult, but not impossible. Although not sure how compatible that would be with existing DOS programs. Also, we emulate that hardware relatively accurately, so what's in the graphics cards framebuffer, that gets displayed on the screen. Although probably such scrollback solution TSRs must have existed on DOS too. Maybe in 4DOS? I'm not too familiar with it.

weirddan455:
> Just sharing an observation: A related historisation item is the console output - modern computers are so fast that text which originally would've been visible to the user now goes by too quickly (e.g. when starting a game before it reaches the graphical part or batch installer, etc.)
> 
> Would be nice if DOSbox has a scroll-bar like CMD.EXE, but I assume that's much more difficult than the command history.

Regarding the first point, I believe there's some config settings that can be used to slow the emulation down though I haven't messed with that much myself.

A scroll bar is easier said than done since dosbox doesn't have a real GUI at the moment (though I believe I've heard some talk of maybe adding one someday).  Could maybe implement scrolling with hotkeys like page up/page down, though it's possible that also ends up being a lot of work.  It's certainly out of scope of this PR but maybe put an issue in for feature request if that's something you want to see.

johnnovak:
> It's certainly out of scope of this PR but maybe put an issue in for feature request if that's something you want to see.

Requires further thought and investigation, and I'm quite sure it's not solvable generically anyway (many TUI programs writes bytes directly to the video mem). Definitely out of scope for this PR :)

weirddan455:
> Also a general usability thought: I noticed that when I `Z:\> exit` cleanly, I'm always greeted to `exit` being the first in my history (which makes sense, but now I always need to "double-tap" up to get past the first exit.
> 
> So the questio is: if 'exit' is the last (more recently entered) command, should we skip writing it (or opposite: skip reading it)? To avoid this annoyance to skip past the first exit?

I don't have a strong opinion on this.  I don't might writing a quick if statement to special case that.  @johnnovak what do you think?

johnnovak:
> Also a general usability thought: I noticed that when I `Z:\> exit` cleanly, I'm always greeted to `exit` being the first in my history (which makes sense, but now I always need to "double-tap" up to get past the first exit.
> 
> So the questio is: if 'exit' is the last (more recently entered) command, should we skip writing it (or opposite: skip reading it)? To avoid this annoyance to skip past the first exit?

Good point, and I'd filter it out as it serves no useful purpose.

johnnovak:
> @johnnovak what do you think?

Go for it! 😎

src/shell/shell.cpp
kcgen: > Probably larger than anything the original DOS could do

Interesting! Maybe we can set a reasonable line limit just for the history, (simple hard-coded limit) like the 500-item limit?

Reason is - I honestly don't want a fullscreen worth of text landing in my command history.. it would be highly suspect and likely garbage data.

I think it's easier for users to get accidental junk on their prompt (say some combination of `type`ing, `|`ing, and/or `echo`ing a binary file) versus the odds of someone manually typing a valid command sequences that sprawls for the entire screen.

(Even doing manual linux prompt entry w/ sed, awk, etc.. it's rare to need more than a couple lines).

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Sounds good.  I set a limit to 256 characters per line.

weirddan455:
I believe I've addressed all the concerns raised so far.  Exit is now filtered out before being written to the file.  Just did a force push to clean up the history a bit and rebase to main.  Let me know if you see any other issues.

kcgen:
Looks great, @weirddan455 - let's get this merged. Thanks!

PR #2618, opened 2023-06-19
Make BatchFile Testable

MeAreJeenius:
There shouldn't be any functional changes except for a fix for a crash when attempting to read a batch file without read permissions on that file.


kcgen:
Sorry for the delay @MeAreJeenius .. so heavy weight PRs moving through the pipes :-)

If you can look into the 'Checks' tab above:

https://github.com/dosbox-staging/dosbox-staging/pull/2618/checks

MacOS and Linux show a warning - which I think it causing link problems on Windows.  Once that fixed it, CI should go green.

The refactoring and new tests look great!

Will get the review underway soon.

kcgen:
> @kcgen It looks like I don't have the needed permissions on the feature branch to push a rebase with my fixes. Can that be fixed?

Thanks for the heads up. Fixing.

kcgen:
@MeAreJeenius - ok, you should have force-push rights to the `mj/*` branches (sorry, normally I add that too but missed it!)

kcgen:
If the git client is pushing through an gpg-agent process or the ssh client is reusing an open socket (which is a comment feature these days), those prior credentials will be pinned in place until a new connection is established. Try to killall those processes and started a fresh connection. Hopefully it does the trick!

src/shell/shell_misc.cpp
```diff
@@ -462,17 +459,18 @@ bool DOS_Shell::Execute(std::string_view name, std::string_view args)
...
+			batchfiles.pop();
+		}
+		try {
+			batchfiles.emplace(*this,
+			                   std::make_unique<FileReader>(fullname),
+			                   name,
+			                   args,
+			                   echo);
+		} catch (const std::exception&) {
+			WriteOut("Could not open %s", fullname.c_str());
```
kcgen, line 472: please avoid adding new exceptions, and when possible use non-throwing calls and check return codes (or use `std::optional<>` to return unavailable objects). 

For example, a file that can't be opened is a routine scenario and should be handled. It's typical and normal.  

Some C++ calls _do_ throw and we deliberately don't catch those when they're truly exceptional.  For example, if the host can't allocate memory, `new` will throw.  We let that exception go all the way to the top and take down the process, because there's no way to proceed from that.

kcgen: Yup; throwing and catching doesn't exist in DOSBox, preferring function return codes or member heath checks. (It's a good fit for DOS too, which was all return-value based).

Some of our most user-data-ladden object are the MIDI devices classes like FluidSynth and MT32: we need to load various ROMs, geting mt32emu running, get threading going, ensure connections to MPU-401, and so on.. 

If any of those fail, the MIDI device's `bool is_open` member never gets set positive, and the outside code can discard the object and inform the users.

Would an similar health-check work here too? `IsOpen()` or whatever makes sense for a new FileReader object. Then you can throw it away if didn't fully construct.

kcgen: You could instantiate it first as a bare object, and only std::move() it into the `batchfiles` container if it's healthy.

kcgen: Yeah, those are single device objects (the entire midi device, and aren't being blasted out by the hundreds) -- yeah; let's see how the wrapper function works! Maybe it'll be a better approach, and we can start to move towards it.

include/shell.h
```diff
@@ -20,134 +20,164 @@
...
+	[[nodiscard]] bool Echo() const;
 
 private:
 	[[nodiscard]] std::string ExpandedBatchLine(std::string_view line) const;
 	[[nodiscard]] std::string GetLine();
+
+	const HostShell& shell;
+	CommandLine cmd;
+	std::unique_ptr<ByteReader> reader;
+	bool echo;
```
kcgen, line 90: Please use defaults initializers.

include/shell.h
```diff
@@ -20,134 +20,164 @@
...
-	DOS_Shell *shell = nullptr;
-	std::shared_ptr<BatchFile> prev = {}; // shared with Shell.bf
-	std::unique_ptr<CommandLine> cmd = {};
-	std::string filename{};
+	void SetEcho(bool echo_on);
+	[[nodiscard]] bool Echo() const;
 
 private:
 	[[nodiscard]] std::string ExpandedBatchLine(std::string_view line) const;
 	[[nodiscard]] std::string GetLine();
```
kcgen, line 85: If we're getting a line, can the function be `const`?

src/hardware/serialport/serialport.cpp
```diff
@@ -1306,7 +1306,7 @@ class SERIALPORTS final : public Module_base {
 			s_property[6] = '1' + static_cast<char>(i);
 			PropMultiVal* p = section->GetMultiVal(s_property);
 			std::string type = p->GetSection()->Get_string("type");
-			CommandLine cmd(0,p->GetSection()->Get_string("parameters"));
+			CommandLine cmd("", p->GetSection()->Get_string("parameters"));
```
kcgen, line 1309: Awesome catch in cleaning up these "zero passed in place of a pointer" instances!

These "I'll just pass a zero in" are extremely misleading to readers.. Readers will see this as a measure of length (strcpy/snprintf) or a starting offset or absolute index.  Small but very impactful!

src/shell/command_line.cpp
```diff
@@ -322,49 +322,27 @@ uint16_t CommandLine::Get_arglength()
...
-
-	while ((c = *c_cmdline) != 0) {
-		if (inquote) {
-			if (c != '"') {
-				str += c;
-			} else {
-				inquote = false;
-				cmds.push_back(str);
-				str.erase();
+	decltype(cmds)::value_type argument;
```
kcgen, line 328: Suggest a `std::string argument = {};` because it's accruing character-by-character from the incoming std::string_view. 

`decltype(cmds)::value_type` is an eyeful and generally we want to avoid this (if we adopted it whole-sale, it would riddle the code base).

Normally reserve it when the source type (or an interfacing type) is even uglier - or there's ongoing risk that the source type itself is changing through ongoing and frequent development or perhaps even a template type.

src/shell/shell.cpp
```diff
@@ -354,8 +353,8 @@ void DOS_Shell::ParseLine(char *line)
 void DOS_Shell::RunInternal()
 {
 	char input_line[CMD_MAXLINE] = {0};
-	while (bf && !shutdown_requested) {
-		if (bf->ReadLine(input_line)) {
+	while (!batchfiles.empty() && !shutdown_requested) {
+		if (batchfiles.top().ReadLine(input_line)) {
```
kcgen: Nice improvement w/ the stack :-)

src/shell/shell_misc.cpp
```diff
@@ -467,16 +462,16 @@ bool DOS_Shell::Execute(std::string_view name, std::string_view args)
...
 		auto reader = FileReader::GetFileReader(fullname);
 		if (reader) {
-			bf = std::make_shared<BatchFile>(this,
-			                                 std::move(*reader),
-			                                 std::string(name).c_str(),
-			                                 std::string(args).c_str());
+			batchfiles.emplace(this,
+			                   std::move(*reader),
+			                   std::string(name).c_str(),
+			                   std::string(args).c_str());
```
kcgen: The `std::string(string_view).c_str()` looks pretty convoluted (Getting a C-string from a temporarily constructed `std::string` initialized with a `std::string_view`).

If the `BatchFile constructor takes in a std::string, then I think we should be able to construct with the sting_view straight-away (the std::string will take a copy of the string_view and then terminate them with `\0`, right?)

include/shell.h
```diff
@@ -66,14 +66,16 @@ class HostShell {
...
+	          bool echo_on);
 	BatchFile(const BatchFile&)            = delete; // prevent copying
 	BatchFile& operator=(const BatchFile&) = delete; // prevent assignment
-	virtual ~BatchFile();
+	virtual ~BatchFile()                   = default;
 	virtual bool ReadLine(char* line);
 	bool Goto(std::string_view label);
 	void Shift();
-	bool echo                        = false;
+	void SetEcho(bool echo_on);
```
kcgen, line 80: > Previously, BatchFiles contained their parent's echo value

Do we know how real DOS behaves wrt batch files inheriting the parent shell's echo state?

include/shell.h
kcgen: Makes sense; thanks for that explanation!

include/shell.h
kcgen: very good to know, thanks for digging into that!

include/shell.h
kcgen: > All of these members are completely dependent on what the constructor gives them.

OK; perfect. We have plenty of places where that's the case for the same reason.  (maybe mention that in a comment here though).

Usually class implementations (in general) are so sloppy that there are many ways to construct an object. Ideally I would like explicitly invalid defaults which get asserted as non-invalid. That's not needed here because of how well locked down you've made the class, which is great. 

 Can the CMD line and bool be const then?  (I realize the reader cannot because it updates per line read, like you explained before).

kcgen:
Just checked the CI results (see the 'checks' tab above).
- PVS Studio - Download, extract, and open the [PR's report](https://github.com/dosbox-staging/dosbox-staging/actions/runs/5373665069?pr=2618) versus [main's report](https://github.com/dosbox-staging/dosbox-staging/actions/runs/5370186145), and compare the `shell_cmds.cpp`; the PR adds a new one (I'm just looking at the counts) 

- MSVC [reports 40 new warnings](https://github.com/dosbox-staging/dosbox-staging/actions/runs/5373665067/jobs/9748254927?pr=2618) attributed to `shell_*.cpp` files:

    ```
      shell_cmds.cpp  : 38
      pdcscrn.cpp     : 5
      enet.h          : 5
      debug.c         : 3
      scr_dump.c      : 2
      slk.c           : 1
      shell_misc.cpp  : 1
      shell_batch.cpp : 1
      keyname.c       : 1
      instr.c         : 1
      initscr.c       : 1
      color.c         : 1
    
    Warnings grouped by type:
    
      C4996 : 55
      C4068 : 3
      C4244 : 2
    
    Total: 60 warnings (out of 20 allowed)
    ```

    I suspect it's just a header block issue; double check that all added files have front matter (license text, `#include dosbox.h` followed by language includes and then subproject includes). Copy off existing Staging additions, like `include/mouse.h` and `src/hardware/input/mouse.cpp`

kcgen:
All commits are building; bisect is healthy :+1: 

```text
Compiling 45e355721 Add BatchFile tests                     .. gcc clang
Compiling 138ef6d07 Use HostShell interface in BatchFile    .. gcc clang
Compiling 6c2006b58 Change BatchFile constructor args to str.. gcc clang
Compiling ec888bd42 Delete move constructor and assigment ex.. gcc clang
Compiling d8cdeef26 Make BatchFiles contain their echo value.. gcc clang
Compiling b57858e7c Make a DOS_Shell function to handle Batc.. gcc clang
Compiling 5d687314c Replace DOS_Shell BatchFile pointer with.. gcc clang
Compiling fc90a638f Add HostShell interface                 .. gcc clang
Compiling bd0727e8c Replace BatchFile CommandLine pointer wi.. gcc clang
Compiling 57b4a6a3f Refactor CommandLine constructor        .. gcc clang
Compiling f5fe41b6e Move CommandLine implementation to its o.. gcc clang
Compiling 2c80d5186 Use ByteReader in BatchFile             .. gcc clang
Compiling c5f8bff2b Add ByteReader interface and FileReader .. gcc clang
Compiling 4c3b7b909 Format shell.h                          .. gcc clang
Compiling 84e87eb37 Move env functions from Program to DOS_S.. gcc clang
```

tests/batch_file_tests.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+TEST(BatchFileGoto, SkipLines)
+{
+	const auto shell = MockShell({});
+	auto batchfile   = BatchFile(shell,
+                                   std::make_unique<MockReader>("before\n:label\nafter"),
+                                   "",
+                                   "",
+                                   true);
+	char line[CMD_MAXLINE];
```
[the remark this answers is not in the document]
kcgen: Ahh interesting.. maybe the Google test macros are confusing the formatter. Not a big deal (setting to resolve).

shermp:
> @kcgen Those warnings were caused by me removing `#include "dosbox.h"` from `shell.h`. Why does including that header suppress those warnings, which were about usage of cstring functions? Why do we even want those suppressed?

Because MSVC is overzealous in warning about using "unsafe" (but standard) c-string functions in favor of their own decidedly non-standard `_s` versions.

PR #2622, opened 2023-06-20
Let derived CDROM devices asynchronously lag their calls

kcgen:
This PR simulates the delay a physical CD-ROM drive took to respond to queries.

This fixes picky games that (expect/assume) that real hardware always had _some_ delay, and therefore back-to-back queries should report monotonically increasing Minute-Second-Frame (MSF) time values.

Fixes Chasm : the Rift's music, which previous failed to loop (resulting in the game eventually going silent), as ripsaw8080 nicely explains: https://www.vogons.org/viewtopic.php?p=225579#p225579

### DIFF
--- a/src/dos/cdrom.cpp
+++ b/src/dos/cdrom.cpp
@@ -28,7 +28,9 @@
 
 #include "SDL.h"
 
+#include "callback.h"
 #include "dosbox.h"
+#include "pic.h"
 #include "string_utils.h"
 
 CDROM_Interface_SDL::CDROM_Interface_SDL()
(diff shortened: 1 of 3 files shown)


src/dos/cdrom.cpp
```diff
@@ -222,17 +238,46 @@ bool CDROM_Interface_Fake :: GetAudioSub(unsigned char& attr, unsigned char& tra
...
 	trayOpen     = false;
+
+	LagDriveResponse();
 	return true;
 }
+
+// Simulate the delay a physical CD-ROM drive took to respond to queries. When
+// added to calls, this ensures that back-to-back queries report monotonically
+// increasing Minute-Second-Frame (MSF) time values.
+//
```
johnnovak, line 265: Very nice that you've explained it! 🎉

src/dos/cdrom.cpp
```diff
@@ -222,17 +238,46 @@ bool CDROM_Interface_Fake :: GetAudioSub(unsigned char& attr, unsigned char& tra
...
+void CDROM_Interface::LagDriveResponse() const
+{
+	// Always simulate a very small amount of drive response time
+	CALLBACK_Idle();
+
+	// Handle tick-rollover
+	static decltype(PIC_Ticks) prev_ticks = 0;
```
johnnovak, line 272: Why not just `static PIC_Ticks prev_ticks = 0;`? What does this give us?

src/dos/cdrom.cpp
```diff
@@ -222,17 +238,46 @@ bool CDROM_Interface_Fake :: GetAudioSub(unsigned char& attr, unsigned char& tra
...
+	prev_ticks = std::min(PIC_Ticks, prev_ticks);
+
+	// Ensure results a monotonically increasing
+	auto since_last_response_ms = [=]() { return PIC_Ticks - prev_ticks; };
```
johnnovak, line 276: I hope this is a proper `uint32_t` at least not just the vague `int`...

johnnovak (approved):
Great enhancement! 🎉

src/dos/cdrom.cpp
kcgen: (it looks like a type, but `PIC_Ticks` is the variable holds the tick count) - so in this case, I'm asking for whatever type `PIC_Ticks` is, that way if we change it, this follows suite).

PR #2624, opened 2023-06-21
Don't emulate CDROM lag for stationary calls

kcgen:
These are calls where time is not being measured or changing.

The remaining calls (left unchanged) involve the passage of time, such as getting the current playback position as relative and absolute MSF objects.

Thanks to @0mnicydle for quickly flagging this as a regression in Dark Seek II's intro FMV sequence.


kcgen:
Reproduced and fixed.

PR #2626, opened 2023-06-21
Fix mounting directory with space in its name, passed as an argument

FeralChild64:
Fix the following problem, reported by user Jow on Discord:
- prepare directory with space in it's name
- pass it as an argument to DOSBox executable (`dosbox "/path/to/directory with space/"` on Linux)
- DOSBox fails to mount the directory as drive C:

Reason: lack of quotation marks in the `MOUNT` command in generated `Z:\AUTOEXEC.BAT` file.


kcgen (approved):
Very nice fix!

PR #2630, opened 2023-06-22
Don't save command history to current working directory

### LINKED ISSUE #2629: DOSBox sometimes writes command line history to current directory

current git main (172ba3322acf5a686e6daa1df4dbf4a0bc18a71d)

Linux x86_64

Gentoo

### Relevant hardware info

AMD64

When DOSBox Staging is started:
- either with `-noprimaryconf` switch
- or with configuration file not containing `command_history_file =` line

it stores the command line history file in the current directory.

This shouldn't happen - with various desktop environments the current directory can be anywhere (documents? desktop? home directory? temp? you don't really know), we shouldn't write there anything unless user explicitly requests.

(see description above)

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
Fixes #2629 

Root problem is this function doesn't get run if the user's config does not have an entry at all:

https://github.com/dosbox-staging/dosbox-staging/blob/172ba3322acf5a686e6daa1df4dbf4a0bc18a71d/src/misc/setup.cpp#L435-L462

So it seems `realpath` cannot be relied on.  I added a workaround to command history's case to append the config directory manually if it hasn't already been done (user can still specify an absolute path to store it wherever they like). 

I think maybe the `Prop_path::SetValue` function should be re-visited but I'm not sure of the best solution.  Maybe it should store a `std_fs::path` rather than a string and find the real path at construction time rather than when it finds a matching line in the config file?

### DIFF
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -462,6 +462,11 @@ static std::string get_command_history_path()
 	if (section) {
 		const auto path = section->Get_path("command_history_file");
 		if (path) {
+			// Workaround: If command_history_file is not in the user's config, path->SetValue function never gets called which sets realpath.
+			// Append the config dir here to prevent this from getting written to current working directory in that case.
+			if (path->realpath.find(CROSS_FILESPLIT) == std::string::npos) {
+				return (get_platform_config_dir() / path->realpath).string();
+			}
 			return path->realpath;
 		}
 	}


kcgen:
When user's run without their primary conf (`-noprimaryconf`), in stand-alone, or "portable" modes (where they provide everything locally), we currently leave their home directory alone.

My hunch is we should avoid touching their directory in this case too.

johnnovak (approved):
I think this small workaround is rather non-invasive and you explained the necessity for it nicely in your comment. Well done @weirddan455 🎉 

I don't think there are too many path settings in the config, but still, a better generic config path solution would require understanding, modifying and regression testing all cases. So I'm inclined to treat that separately; that seems a much bigger refactoring/cleanup job to me.

Such a refactoring would give you a nice leisurely tour over the codebase and various features, though (MIDI, capturing features, shader loadig, etc.) (wink wink 😄 but if you would at least raise a backlog ticket for this, that would be much appreciated!)

weirddan455:
> When user's run without their primary conf (`-noprimaryconf`), in stand-alone, or "portable" modes (where they provide everything locally), we currently leave their home directory alone.
> 
> My hunch is we should avoid touching their directory in this case too.
> 
> (edit: perhaps if they have an existing file, we could update it. But I would avoid writing a new file or creating any directorieS).

There is some detection for the portable mode here:

https://github.com/dosbox-staging/dosbox-staging/blob/172ba3322acf5a686e6daa1df4dbf4a0bc18a71d/src/misc/cross.cpp#L175-L204

I believe that avoids writing to the home directory (instead writing to wherever the dosbox config file is located).

weirddan455:
Also, we're already creating directories in the path the above function returns:

https://github.com/dosbox-staging/dosbox-staging/blob/172ba3322acf5a686e6daa1df4dbf4a0bc18a71d/src/gui/sdlmain.cpp#L4888-L4908

kcgen:
> Also, we're already creating directories in the path the above function returns:
> 
> https://github.com/dosbox-staging/dosbox-staging/blob/172ba3322acf5a686e6daa1df4dbf4a0bc18a71d/src/gui/sdlmain.cpp#L4888-L4908

Good catch, @weirddan455.  Just tested all attempts as a portable launch (`-conf`, `-noprimaryconf`, and local executable) and indeed, the config tree is checked and created if it doesn't exist. (We had a user get quite upset years ago; we shouldn't dare write application records in a portable launch). But.. well, no one is complaining now - and certainly this PR isn't going to affect that anyway! :-)

Merging. Thank you for the quick update!

(the CI warning is a new MSYS-related tar issue, unrelated to anything on our side.. probably a bug introduced in the latest CI VMs).

PR #2635, opened 2023-06-23
Disable writing command line history if entry is empty

weirddan455:
Follow up to #2629 and #2630

Command history should be disabled if the config entry is empty.  This is a regression caused by my previous PR.  Added missing empty check.

Likely this just caused an erroneous warning message as it tried to create a file with the same name as the user's config directory, which will fail.


johnnovak (approved):
@weirddan455 I have given you write access to the `dosbox-staging` repo; now you should be able to create your own branches with the `wd/` prefix without having to fork. You should also be able to force push into it and delete your branches. Makes life so much easier! 🎉 

Please pay attention to the `wd/` prefix — that must be present in branch names created by you.

Congratulations! 🎉 🍻 

Merging this now, then you can try out your new powers in your next PR 😎

PR #2639, opened 2023-06-24
Cleanup a handul of new Coverity issues 

kcgen:
This is a handful of very minor fixes to get back to baseline w/ Coverity. Changes are 1:1 with the original code. Tests as passing, as is CI.

The beast is still delivering :-)

(Here's just one ticket)

### DIFF
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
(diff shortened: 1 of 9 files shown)


johnnovak (approved):
Thanks for the cleanup @kcgen 👍🏻

kcgen:
Finally at zero Coverity defects (the are still a handful left in 3rd party libs and the system's header files):



https://scan.coverity.com/projects/dosbox-staging

PR #2642, opened 2023-06-24
Add a script to compile all commits on the current branch

kcgen:
Meant to be no-fuss script to check the compile-ability of all commits on a given branch.

Ensuring that each commit compiles is important for future bisect-ability; because if a commit doesn't compile, then the commit's changes cannot be tested in isolation (and the person bisecting needs to skip it, in which case the changes get lumped into downstream commits that do eventually build.. making the bug-isolation task harder).

For PR reviewers interested in trying it before merge:
1. Copy the script somewhere outside the repo
2. Checkout a different branch that has a bunch of commits.
3. Launch the script

Sample output:

```
Realtime log: /dev/shm/tmp.v8bJ3c5k6n
Using clang native file: .github/meson/native-clang-17.ini
Compiling c07c848b1 Lower allowed PVS Studio and MSVC warnin.. gcc clang
Compiling 26db0bccf Fix a loss-of-precision warning converti.. gcc clang
Compiling 6a0cf76c6 Fix implified float to double conversion.. gcc clang
Compiling 4c68831f5 Zero-out delete[]'d members before destr.. gcc clang
Compiling a10d46732 Fix an implicit sign conversion in rende.. gcc clang
Compiling 336df9c9f Fix an implicit loss-of-precision warnin.. gcc clang
Compiling e2b58beee Fix a use-after-move in the string_forma..
```

In this case, my system has both compilers, so it builds with both.

Once `ccache` is primed, builds can go quite quickly in succession.

The long term plan is to include this in CI so we have this check taken care of for all future PRs.


kcgen:
(including all recent contributors just so you're aware of this script and can start putting it to use)

kcgen:
When ccache get hots, the builds can really cruise.. 

```
time ~/compile_commits.sh 
Realtime log: /var/folders/g7/3ntympz90_z9gr61_578k7wc0000gp/T/tmp.b5vvhME2
No supported gcc versions found, skipping
Compiling c07c848b1 Lower allowed PVS Studio and MSVC warnin.. clang
Compiling 26db0bccf Fix a loss-of-precision warning converti.. clang
Compiling 6a0cf76c6 Fix implified float to double conversion.. clang
Compiling 4c68831f5 Zero-out delete[]'d members before destr.. clang
Compiling a10d46732 Fix an implicit sign conversion in rende.. clang
Compiling 336df9c9f Fix an implicit loss-of-precision warnin.. clang
Compiling e2b58beee Fix a use-after-move in the string_forma.. clang
Compiling 33194913a Fix an uninitialized scalar variable war.. clang
Compiling a7982bab2 Avoid a sign extension when computing th.. clang
```

**`real	0m13.161s`**
**`user	0m25.490s`**
**`sys	0m10.097s`**

scripts/compile_commits.sh
```diff
@@ -0,0 +1,224 @@
...
+	                 "$SED" 's@^refs/remotes/origin/@@')
+
+	if [[ -z $default_branch ]]; then
+		echo "Failed to determine the default branch."
+		exit 1
+	fi
+	readarray -t commits < <(git log \
+	                         --format="%h" HEAD \
+	                         --not "$default_branch")
+
```
kklobe, line 132: `readarray` is not supported under Zsh on macOS nor, I believe, on older versions of Bash on Mac. Consider a loop for portability, e.g.:

```shell
commits=()
	while IFS= read -r line; do
		commits+=("$line")
	done < <(git log --format="%h" HEAD --not "$default_branch")
```

kcgen: Thanks for that fix up, @kklobe !

kcgen: Nice - you've got GCC 13.x building too? 
I need to install that and get it in my path on the M1 mini :-)

kklobe: yep, the GCC formulae on homebrew have come a long ways in the past year, I need it primarily for gfortan with R packages.

kcgen: GCC 13.0.0 working great here now on the M1, thanks @kklobe :+1: 

```text
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: arm64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

johnnovak:
Very nice & welcome addition @kcgen . I'll test it on macOS later.

PR #2645, opened 2023-06-25
Fix UTF8 to DOS conversion by mistake applied twice in autoexec code

FeralChild64:
There is a bug in `Z:\AUTOEXEC.BAT` generation code, which is only visible if translation is used - namely, the `MSG_Get` is used to get the translatable messages in DOS code page format, and afterwards `utf8_to_dos` is called to convert the entire `Z:\AUTOEXEC.BAT` (which at this point contains UTF8 content of `[autoexec]` section) - this is now fixed, code now requests translatable strings in an unprocessed format.


kcgen (approved):
Great catch on these corner cases, @FeralChild64!

PR #2648, opened 2023-06-26
Fix mouse movement limitation regression in some games

FeralChild64:
The recent mouse cursor range fix for text modes over 80 columns went too far - it seems some graphic modes below 640 pixels width really need maximum x set to 639, or mouse cursor in some games won’t work correctly.

### DIFF
--- a/src/hardware/input/mouseif_dos_driver.cpp
+++ b/src/hardware/input/mouseif_dos_driver.cpp
@@ -948,25 +948,21 @@ void MOUSEDOS_AfterNewVideoMode(const bool is_mode_changing)
 		state.granularity_y = 0xfff8;
 		set_maxpos_text();
 		break;
-	case 0x08: // 160x200, 16 colors    (PCjr)
-		state.maxpos_x = 159;
-		state.maxpos_y = 199;
-		break;
-	case 0x04: // 320x200, 4 colors     (CGA, EGA, MCGA, VGA)
-	case 0x05: // 320x200, 4 colors     (CGA, EGA, MCGA, VGA)
-	case 0x09: // 320x200, 16 colors    (PCjr)
-		state.maxpos_x = 319;
-		state.maxpos_y = 199;
-		break;
 	case 0x0d: // 320x200, 16 colors    (EGA, VGA)
 	case 0x13: // 320x200, 256 colors   (MCGA, VGA)
 		state.granularity_x = 0xfffe;
 		state.maxpos_x = 639;
 		state.maxpos_y = 199;
 		break;
+	case 0x04: // 320x200, 4 colors     (CGA, EGA, MCGA, VGA)
+	case 0x05: // 320x200, 4 colors     (CGA, EGA, MCGA, VGA)
 	case 0x06: // 640x200, black/white  (CGA, EGA, MCGA, VGA)
+	case 0x08: // 160x200, 16 colors    (PCjr)
+	case 0x09: // 320x200, 16 colors    (PCjr)
 	case 0x0a: // 640x200, 4 colors     (PCjr)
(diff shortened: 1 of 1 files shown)


kcgen:
I'm guessing VGA tweaks modes should be OK if they were previously OK  (Scorched Earth's various modes make good test-cases).

Approving; just waiting for feedback that this fixes it.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -948,25 +948,21 @@ void MOUSEDOS_AfterNewVideoMode(const bool is_mode_changing)
...
-		state.maxpos_y = 199;
-		break;
-	case 0x04: // 320x200, 4 colors     (CGA, EGA, MCGA, VGA)
-	case 0x05: // 320x200, 4 colors     (CGA, EGA, MCGA, VGA)
-	case 0x09: // 320x200, 16 colors    (PCjr)
-		state.maxpos_x = 319;
-		state.maxpos_y = 199;
-		break;
 	case 0x0d: // 320x200, 16 colors    (EGA, VGA)
 	case 0x13: // 320x200, 256 colors   (MCGA, VGA)
```
johnnovak, line 952: Note that mode 13h is the basis of most tweaked VGA modes used in many games. The program set up mode 13h first via BIOS, then tweaks it further by changing the VGA registers directly to something like 320x240, 320x400, 360x240, 360x400, 256x256, 384x304, and so on... Pretty wild cominations are possible, and the last set BIOS mode will always be reported as 13h.

Then a program can set up any mode circumventing the BIOS entirely.

How would all this affect your mouse code? On the flipside, I don't think any mouse drivers are "VGA tweak mode aware" but they still work somehow correctly.

johnnovak: Also, I think out get out of jail for free card here is that most games that use tweaked modes are action games and feature no mouse pointer.

kcgen: Here's Scorched Earth in a tweaked mode:

image0001

The mouse can be moved to each corner (without going beyond), without problems.

kcgen: For testing: 
[Scorched Earth.zip :file_folder: ](https://github.com/dosbox-staging/dosbox-staging/files/11884537/Scorched.Earth.zip)

FeralChild64: @johnnovak: The `maxpos_x` / `maxpos_y` can be manipulated by software, they can simply call INT 33 with AX equal to 0x06 or 0x07 - see `int33_handler` in the same file. These calls can make our driver work with just about any strange resolution, moreover bugs in `MOUSEDOS_AfterNewVideoMode` might not trigger wrong mouse behavior is the software sets the values by itself, "just in case".

I have recently ordered a "Microsoft Mouse Programmer's Reference" book, 2nd edition, maybe it will give me some more information - but it might take a few weeks before it arrives from the other side of The Pond. For now I have restored the old behavior for Hercules/CGA/EGA/VGA graphic modes, like it was here: https://github.com/dosbox-staging/dosbox-staging/blob/9f507e19ee9b3b5c5abd32cbc3f9041f332dd586/src/hardware/input/mouseif_dos_driver.cpp.

kcgen:
Comments addressed and fixed; merging. Thanks for the quick fix, @FeralChild64 !

src/hardware/input/mouseif_dos_driver.cpp
johnnovak: Thanks for the explanation @FeralChild64, probably that's how all these games with weird custom modes work then (they set up the right max values via INT33 calls).

PR #2650, opened 2023-06-28
Show backtraces in debug builds using Loguru's assert

kcgen:


johnnovak (approved):
Nice one @kcgen! Hopefully that will give us more context when a debug assert hits!

PR #2652, opened 2023-06-28
Enable loguru stack traces

kklobe:
Loguru's autodetect logic for enabling stack traces is not very robust, so add logic to meson.build that checks for the headers included by `loguru.cpp` if `LOGURU_STACKTRACES=1`.

Here is the relevant snippet from `loguru.cpp`

```cpp
#if LOGURU_STACKTRACES
	#include <cxxabi.h>    // for __cxa_demangle
	#include <dlfcn.h>     // for dladdr
	#include <execinfo.h>  // for backtrace
#endif // LOGURU_STACKTRACES
```

So  I just check for those headers in `meson.build` and force the flag, since `loguru.cpp` needs to be compiled with it.

Here's an example trace from an assert:

```log
Assertion failed: (1==0), function increaseticks, file dosbox.cpp, line 186.

Stack trace:
18         0x18d68bf28 start + 2236
17         0x1000b2624 main + 36
16         0x1003a3760 sdl_main(int, char**) + 5844
15         0x1001186bc Config::StartUp() + 28
14         0x100626408 SHELL_Init() + 3904
13         0x100623704 DOS_Shell::Run() + 1540
12         0x100647408 DOS_Shell::InputCommand(char*) + 56
11         0x100647984 DOS_Shell::ReadCommand() + 712
10         0x10025a8f0 DOS_ReadFile(unsigned short, unsigned char*, unsigned short*, bool) + 292
9          0x1002551cc DOS_Device::Read(unsigned char*, unsigned short*) + 64
8          0x100252c04 device_CON::Read(unsigned char*, unsigned short*) + 340
7          0x100184a90 CALLBACK_RunRealInt(unsigned char) + 92
6          0x1000b2ecc DOSBOX_RunMachine() + 28
5          0x1000b2e90 Normal_Loop() + 264
4          0x1000b2704 increaseticks() + 196
3          0x18d8f0e44 err + 0
2          0x18d8f1ae8 abort + 180
1          0x18d9e3c28 pthread_kill + 288
0          0x18da12a24 _sigtramp + 56
2023-06-28 10:03:55.285 | Signal: SIGABRT
```

### DIFF
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -246,16 +246,4 @@
 // XDG_DATA_HOME or XDG_DATA_DIRS to include the --datadir.
 #mesondefine CUSTOM_DATADIR
 
-/* Loguru-related defines
- */
-
-// Prevent loguru from parsing command-line arguments with
-// with the hosts's locale-applied, because this can foul up
-// ncurses. (DOSBox also doesn't have foreign-language arguments).
-//
-#define LOGURU_USE_LOCALE 0
-
-// Redefine “assert” to call Loguru version (!NDEBUG only).
-#define LOGURU_REDEFINE_ASSERT 1
-
 #endif
--- a/src/libs/loguru/meson.build
+++ b/src/libs/loguru/meson.build
@@ -1,3 +1,23 @@
+# Enable Loguru stack traces if supported
+stacktrace_headers = ['cxxabi.h', 'dlfcn.h', 'execinfo.h']
+
+all_stacktrace_headers_found = true
+
+foreach header : stacktrace_headers
+    if not cxx.has_header(header)
+        all_stacktrace_headers_found = false
+    endif
+endforeach
+
+if all_stacktrace_headers_found
+    add_project_arguments('-DLOGURU_STACKTRACES=1', language: 'cpp')
+endif
+
+# Prevent loguru from parsing command-line arguments with
+# the hosts's locale-applied, because this can foul up
+# ncurses. (DOSBox also doesn't have foreign-language arguments).
+add_project_arguments('-DLOGURU_USE_LOCALE=0', language: 'cpp')
+
 libloguru = static_library(
     'loguru',
     'loguru.cpp',
--- a/src/platform/visualc/config.h
+++ b/src/platform/visualc/config.h
@@ -122,15 +122,3 @@
 // On Windows, this path is not customizeable, so it's left blank here.
 //
 #define CUSTOM_DATADIR ""
-
-/* Loguru-related defines
- */
-
-// Prevent loguru from parsing command-line arguments with
-// with the hosts's locale-applied, because this can foul up
-// ncurses. (DOSBox also doesn't have foreign-language arguments).
-//
-#define LOGURU_USE_LOCALE 0
-
-// Redefine “assert” to call Loguru version (!NDEBUG only).
-#define LOGURU_REDEFINE_ASSERT 1


meson.build
```diff
@@ -150,6 +150,23 @@ if get_option('buildtype').startswith('debug')
...
     endif
+
+    # Enable Loguru stack traces if supported
+    stacktrace_headers = ['cxxabi.h', 'dlfcn.h', 'execinfo.h']
+
+    all_stacktrace_headers_found = true
+
+    foreach header : stacktrace_headers
+        if not cxx.has_header(header)
+            message('@0@ not found'.format(header))
```
kcgen, line 161: The message log probably isn't needed; Meson reports these checks on the fly:

```
Has header "pwd.h" : YES 
Has header "strings.h" : YES 
Has header "netinet/in.h" : YES 
Has header "sys/socket.h" : YES 
etc..
```

meson.build
```diff
@@ -150,6 +150,23 @@ if get_option('buildtype').startswith('debug')
...
+            all_stacktrace_headers_found = false
+        endif
+    endforeach
+
+    if all_stacktrace_headers_found
+        message('Enabling Loguru stack traces')
+        add_project_arguments('-DLOGURU_STACKTRACES=1', language: 'cpp')
+    endif
```
kcgen, line 169: Suggest Meson's conf header define approach (see `src/config.h`)

```c++
#mesondefine LOGURU_STACKTRACES
```

Then here in the meson.build:

```python
conf_data.set10('LOGURU_STACKTRACES', all_stacktrace_headers_found)
```

The `message` can summarized at the end of the `meson setup` phase. something like:

```python
summary('Loguru stack traces', all_stacktrace_headers_found)
```

kklobe: that didn't work, since loguru.cpp doesn't include config.h, so I put it in the build rather than hack up the loguru base.

meson.build
```diff
@@ -150,6 +150,21 @@ if get_option('buildtype').startswith('debug')
...
+
+    foreach header : stacktrace_headers
+        if not cxx.has_header(header)
+            all_stacktrace_headers_found = false
+        endif
+    endforeach
+
+    if all_stacktrace_headers_found
+        add_project_arguments('-DLOGURU_STACKTRACES=1', language: 'cpp')
+    endif
```
kcgen, line 167: Want to move the other two loguru settings over from src/config.h.in?

(For MSVC builds, maybe loguru's auto checks are good enough, or should defines be added in the solution file?)

kklobe: I wonder if we even need the assert redefine, since I'm getting a stack trace just fine without it, and it just does an undef, so `loguru.hpp` would have to be included after every include of `cassert`. I'm not sure it buys us much. From `loguru.hpp':

```cpp
#if LOGURU_REDEFINE_ASSERT
	#undef assert
	#ifndef NDEBUG
		// Debug:
		#define assert(test) CHECK_WITH_INFO_F(!!(test), #test) // HACK
	#else
		#define assert(test)
	#endif
#endif // LOGURU_REDEFINE_ASSERT
```

kklobe:
@kcgen I moved this into loguru's `meson.build` per your suggestions, makes much more sense there.

kcgen:
> @kcgen I moved this into loguru's `meson.build` per your suggestions, makes much more sense there.

Looks great, @kklobe 🚀 

@johnnovak , very curious what this reveals in your next hour+ run w/ a debug build :)

johnnovak:
Thanks @kklobe, this is great! 🚀 

> @johnnovak , very curious what this reveals in your next hour+ run w/ a debug build :)

Yeah now those bugs have nowhere to hide! 😆 🐛 🪲 🦟

PR #2653, opened 2023-06-28
Fix bug where DTA can be modified by DELETE command

weirddan455:
I was looking at code for these shell commands and I believe I found a bug here.  There's two error checking + early returns where the DTA can be modified and not restored.  In other commands, it gets restored on all paths.

DTA is not actually needed until `DOS_FindFirst` gets run so I moved the saving/modifying code until after the error checking is done.


FeralChild64:
Looks OK to me (BTW, thanks for noticing this bug!), but let's give others a chance to review.

johnnovak (approved):
Nice catch @weirddan455 👍🏻

FeralChild64 (approved):

PR #2655, opened 2023-06-29
Update increaseticks sleep strategy

kklobe:
This is an experiment to improve the accuracy of the engine that is `increaseticks`. `PIC_RunQueue` is at the mercy of the 1ms `sleep_for`, which can cause an entire millisecond tick to be missed. By lowering the sleep time, `PIC_RunQueue` can run more often. This allows it to get closer to the precise time specified when an event is added via `PIC_AddEvent`; many calls throughout the codebase provide sub-millisecond or fractional delays.

Also, it keeps track of the excess microseconds after each sleep and adjusts `ticksDone` for more accuracy during cycle guessing, since `sleep_for` isn't precise; for example, a 1ms sleep is actually ~1150-1200us on my M1 Mac, causing it to think it slept for a total of 10 ticks when it was really 11.

None of this should provide a massive difference, but it should give much more accurate responsiveness within ticks. I did notice a difference on the Future Crew Panic demo, on the first 'scrollers suck' portion, with fewer flickers compared to `main`. At best, it will help smooth out micro-glitches in audio and video here and there, and worst case it should be identical to `main`.

The one wild card is the new sleepDuration:

```cpp
constexpr auto sleepDuration = std::chrono::microseconds(100);
```

On my M1 Mac, it significantly reduces CPU usage compared to 250 or 500 (and even 1000), but that is completely up to the Host OS scheduler. I wanted to start with an aggressive value, and we can back it off if empirical testing warrants.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -180,18 +180,29 @@ void increaseticks() { //Make it return ticksRemain and set it in the function a
 		return;
 	}
 
-	const auto ticksNew = GetTicks();
+	const auto ticksNewUs = GetTicksUs();
+	const auto ticksNew = ticksNewUs / 1000;
+
 	ticksScheduled += ticksAdded;
 	if (ticksNew <= ticksLast) { //lower should not be possible, only equal.
 		ticksAdded = 0;
 
-		constexpr auto duration = std::chrono::milliseconds(1);
-		std::this_thread::sleep_for(duration);
+		static int64_t cumulativeTimeSlept = 0;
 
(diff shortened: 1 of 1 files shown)


johnnovak:
Nice one @kklobe! Definitely a welcome improvement, I'm only hoping the current inaccurate sleep behaviour doesn't mask some timing related emulation deficiencies. But we will see!

johnnovak (approved):
Looks ok, I guess we can just merge and live with it for a while and see what happens!

PR #2656, opened 2023-07-01
Fix image capture indices always starting from 1

johnnovak:
A last-minute improvement to the capture PR to create the capture directory on-demand (lazily) introduced a regression: the image capture indices now always start from 1, potentially overwriting existing images. This was a result of the image capturers running asynchronously in separate threads combined with the lazy/deferred capture directory creation and  highest index lookup approach.

This is the fix for that, plus also for a potential race condition when determining the current highest capture indices.

### DIFF
--- a/src/capture/capture.cpp
+++ b/src/capture/capture.cpp
@@ -24,6 +24,7 @@
 #include <cassert>
 #include <cerrno>
 #include <cstdio>
+#include <mutex>
 
 #include "capture_audio.h"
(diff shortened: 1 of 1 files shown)


FeralChild64 (approved):
Looks OK.

BTW - funny thing, I just wanted to report this bug, and the fix is already available :)

johnnovak:
Thanks for the review @FeralChild64

PR #2658, opened 2023-07-02
Improve text mode detection in DOS mouse driver

FeralChild64:
Fix for potential problem with text mode detection found by @johnnovak - although, according to my tests, the graphics-specific routines also work for text mode (checked with Norton Commander 5.5 started on Hercules, CGA, and Tandy).


johnnovak (commented):
Thanks @FeralChild64 🎉 Good to hear it worked anyway. The `CurMode` stuff is a bit of a mess as I've just discovered... Especially for the Tandy/PCjr/CGA modes and their composite variants.

johnnovak (approved):
Looks good @FeralChild64 👍🏻

PR #2659, opened 2023-07-02
Update RPM spec file

rderooy:
The RPM spec file was rather out of date, so I updated it to 0.80.1.

Being able to build the RPM package from a SPEC file with `rpmbuild` however does not guarantee that you can build it on other systems. As it uses your local environment and installed packages. 

As such, I also did a `mock` build (chroot environment, like copr uses) to ensure all the `BuildRequires` statements are actually pulling in everything that is needed to do a full build.

Lastly, I spotted a few typos in BUILD.md

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -23,7 +23,7 @@ example, the **debug** and **release** configurations can reside in
 `build/debug` and `build/release`, respectively.
 
 We recommend setting the following Ccache environment variables to maximize
-the use of storage and benefit from precompiled headers:
+the use of storage and benefit from pre-compiled headers:
 
 ```shell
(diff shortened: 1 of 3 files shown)


johnnovak (approved):
The doco fixes look good, but can't really comment on the rest as I'm not a Linux user. Maybe @FeralChild64? (Approving anyway, don't want to hold this up)

FeralChild64 (approved):

FeralChild64:
Unfortunately, I stopped using RPM-based distros some 20 years ago, so I don't have environment to test this. But I don't see anything suspicious, so approving too.

PR #2660, opened 2023-07-03
Implement MOVE shell command

### LINKED ISSUE #2539: Add the MOVE internal DOS command
Shouldn't be too hard:

https://www.computerhope.com/movehlp.htm

weirddan455:
Fixes #2539

I'm sure there's some stuff in here that needs cleaning up so feel free to be critical.  It's not the prettiest function (neither is CMD_COPY).  I've tested though and, as far as I can tell, it works identically to MS-DOS 6.22.

Error messages are an exception.  That can maybe get cleaned up.  MS-DOS does like `source => dest [error message]` and right now I'm just throwing errors as soon as they come up without formatting like that.

### DIFF
--- a/include/shell.h
+++ b/include/shell.h
@@ -178,6 +178,7 @@ public:
 	void CMD_VER(char* args);
 	void CMD_LS(char* args);
 	void CMD_VOL(char* args);
+	void CMD_MOVE(char* args);
 
 	/* The shell's variables */
 	uint16_t input_handle         = 0;
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -647,6 +647,8 @@ void SHELL_Init() {
 	MSG_Add("SHELL_FILE_NOT_FOUND", "File not found - '%s'\n");
 	MSG_Add("SHELL_FILE_EXISTS", "File '%s' already exists.\n");
 	MSG_Add("SHELL_DIRECTORY_NOT_FOUND", "Directory not found - '%s'\n");
+	MSG_Add("SHELL_READ_ERROR", "Error reading file - '%s'\n");
+	MSG_Add("SHELL_WRITE_ERROR", "Error writing file - '%s'\n");
 
 	// Command specific messages
 	MSG_Add("SHELL_CMD_HELP","If you want a list of all supported commands type [color=yellow]help /all[reset] .\nA short list of the most often used commands:\n");
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
 	        " Volume in drive %c is %s\n"
 	        " Volume Serial Number is %04X-%04X\n"
 	        "\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP",
+	        "Moves files and renames files and directories.\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP_LONG",
+	        "Usage:\n"
+	        "  [color=green]move[reset] [color=white]FILENAME1[,FILENAME2,...][reset] [color=cyan]DESTINATION[reset]\n"
+	        "  [color=green]move[reset] [color=white]DIRECTORY1[reset] [color=cyan]DIRECTORY2[reset]\n"
+	        "\n"
+	        "Where:\n"
+	        "  [color=white]FILENAME[reset]    Can be either an exact filename or an inexact filename with\n"
+	        "              wildcards, which are the asterisk (*) and the question mark (?).\n"
+	        "              Multiple, comma-seperated, filenames can be provided.\n"
+	        "  [color=white]DIRECTORY[reset]   An exact directory name, not containing any wildcards.\n"
+	        "  [color=cyan]DESTINATION[reset] An exact filename or directory, not containing any wildcards.\n"
+	        "\n"
+	        "Notes:\n"
+	        "  If multiple source files are specified, destination must be a directory.\n"
+	        "  If not, one will be created for you.\n"
+	        "\n"
+	        "  If a single source file is specified, it will overwrite destination.\n"
+	        "\n"
+	        "Examples:\n"
+	        "  [color=green]move[reset] [color=white]source.bat[reset] [color=cyan]new.bat[reset]\n"
+	        "  [color=green]move[reset] [color=white]file1.txt,file2.txt[reset] [color=cyan]mydir[reset]\n");
+	MSG_Add("SHELL_CMD_MOVE_MULTIPLE_TO_SINGLE",
+	        "Cannot move multiple files to a single file.\n");
 
 	/* Ensure help categories are loaded into the message vector */
 	HELP_AddMessages();
--- a/src/shell/shell_cmds.cpp
+++ b/src/shell/shell_cmds.cpp
@@ -68,6 +68,7 @@ static const std::map<std::string, SHELL_Cmd> shell_cmds = {
 	{ "LOADHIGH", {&DOS_Shell::CMD_LOADHIGH, "LOADHIGH", HELP_Filter::All,    HELP_Category::Misc } },
 	{ "MD",       {&DOS_Shell::CMD_MKDIR,    "MKDIR",    HELP_Filter::Common, HELP_Category::File } },
 	{ "MKDIR",    {&DOS_Shell::CMD_MKDIR,    "MKDIR",    HELP_Filter::All,    HELP_Category::File } },
+	{ "MOVE",     {&DOS_Shell::CMD_MOVE,     "MOVE",     HELP_Filter::All,    HELP_Category::File } },
 	{ "PATH",     {&DOS_Shell::CMD_PATH,     "PATH",     HELP_Filter::All,    HELP_Category::Misc} },
 	{ "PAUSE",    {&DOS_Shell::CMD_PAUSE,    "PAUSE",    HELP_Filter::All,    HELP_Category::Batch } },
 	{ "RD",       {&DOS_Shell::CMD_RMDIR,    "RMDIR",    HELP_Filter::Common, HELP_Category::File } },
@@ -2241,4 +2242,303 @@ void DOS_Shell::CMD_VOL(char* args)
 	const auto low  = check_cast<uint16_t>(serial & 0xFFFF);
 	const auto trimmed_label = trim(label);
 	WriteOut(MSG_Get("SHELL_CMD_VOL_OUTPUT"), drive, trimmed_label, high, low);
+}
+
+static std::vector<std::string> cmd_move_parse_sources(const char*& args)
+{
+	bool parsing             = true;
+	bool inside_quote        = false;
+	bool stop_next_letter    = false;
+	bool continue_after_stop = false;
+	std::string source;
+	std::vector<std::string> sources;
+
+	// Sources are seperated by commas and terminated by a trailing space
+	while (parsing) {
+		const auto ch = *args;
+		switch (ch) {
+		case '"':
+			inside_quote = !inside_quote;
+			++args;
+			break;
+		case ' ':
+			if (inside_quote) {
+				source.push_back(ch);
+				++args;
+			} else {
+				if (!source.empty()) {
+					sources.emplace_back(std::move(source));
+					source              = {};
+					stop_next_letter    = true;
+					continue_after_stop = false;
+				}
+				++args;
+			}
+			break;
+		case ',':
+			stop_next_letter    = true;
+			continue_after_stop = true;
+			if (!source.empty()) {
+				sources.emplace_back(std::move(source));
+				source = {};
+			}
+			++args;
+			break;
+		case 0:
+			if (!source.empty()) {
+				sources.emplace_back(std::move(source));
+			}
+			parsing = false;
+			break;
+		default:
+			if (stop_next_letter) {
+				parsing             = continue_after_stop;
+				continue_after_stop = false;
+				stop_next_letter    = false;
+			} else {
+				source.push_back(ch);
+				++args;
+			}
+		}
+	}
+	return sources;
+}
+
+static std::string cmd_move_parse_destination(const char*& args)
+{
+	std::string destination;
+	bool parsing      = true;
+	bool inside_quote = false;
+
+	while (parsing) {
+		const auto ch = *args;
+		switch (ch) {
+		case '"':
+			inside_quote = !inside_quote;
+			++args;
+			break;
+		case ' ':
+			if (inside_quote) {
+				destination.push_back(ch);
+				++args;
+			} else {
+				++args;
+				parsing = false;
+			}
+			break;
+		case 0: parsing = false; break;
+		default: destination.push_back(ch); ++args;
+		}
+	}
+	return destination;
+}
+
+struct cmd_move_arguments {
+	std::vector<std::string> sources = {};
+	std::string destination          = {};
+	const char* error                = nullptr;
+};
+
+cmd_move_arguments cmd_move_parse_arguments(const char* args)
+{
+	cmd_move_arguments ret;
+	// args pointer gets advanced by the two helper functions
+	ret.sources = cmd_move_parse_sources(args);
+	if (ret.sources.empty()) {
+		ret.error = "SHELL_MISSING_PARAMETER";
+		return ret;
+	}
+	ret.destination = cmd_move_parse_destination(args);
+	if (ret.destination.empty()) {
+		ret.error = "SHELL_MISSING_PARAMETER";
+		return ret;
+	}
+	// Check for extra arguments
+	while (*args) {
+		if (*args++ != ' ') {
+			ret.error = "SHELL_TOO_MANY_PARAMETERS";
+			return ret;
+		}
+	}
+	return ret;
+}
+
+void DOS_Shell::CMD_MOVE(char* args)
+{
+	HELP("MOVE");
+
+	// TODO: Add support for these flags along with the COPYCMD environment
+	// variable (overwrite prompts) Also currently ignored by CMD_COPY
+	(void)ScanCMDBool(args, "Y");
+	(void)ScanCMDBool(args, "-Y");
+
+	char* rem = ScanCMDRemain(args);
+	if (rem) {
+		WriteOut(MSG_Get("SHELL_ILLEGAL_SWITCH"), rem);
+		return;
+	}
+
+	cmd_move_arguments parsed = cmd_move_parse_arguments(args);
+	if (parsed.error) {
+		WriteOut(MSG_Get(parsed.error));
+		return;
+	}
+
+	char temp[DOS_PATHLENGTH];
+	if (!DOS_Canonicalize(parsed.destination.c_str(), temp)) {
+		WriteOut(MSG_Get("SHELL_FILE_NOT_FOUND"),
+		         parsed.destination.c_str());
+		return;
+	}
+	std::string canonical_destination = temp;
+
+	// Command uses dta so set it to our internal dta
+	const RealPt save_dta = dos.dta();
+	dos.dta(dos.tables.tempdta);
+	DOS_DTA dta(dos.dta());
+
+	// Each raw source can be a wildcard (*.*) containing multiple real
+	// sources.
+	std::vector<std::string> final_sources;
+	for (const std::string& source : parsed.sources) {
+		FatAttributeFlags attr = {0xff};
+		attr.volume.flip();
+		if (!DOS_FindFirst(source.c_str(), attr._data)) {
+			WriteOut(MSG_Get("SHELL_FILE_NOT_FOUND"), source.c_str());
+			continue;
+		}
+		if (!DOS_Canonicalize(source.c_str(), temp)) {
+			WriteOut(MSG_Get("SHELL_FILE_NOT_FOUND"), source.c_str());
+			continue;
+		}
+		std::string source_dir = temp;
+		source_dir = source_dir.substr(0, source_dir.rfind(DosSeparator) + 1);
+		do {
+			DOS_DTA::Result result = {};
+			dta.GetResult(result);
+			if (result.name != CurrentDirectory &&
+			    result.name != ParentDirectory) {
+				final_sources.emplace_back(source_dir + result.name);
+			}
+		} while (DOS_FindNext());
+	}
+
+	// Done with DTA. Restore it.
+	dos.dta(save_dta);
+
+	uint16_t destination_attr = 0;
+	bool destination_exists = DOS_GetFileAttr(canonical_destination.c_str(),
+	                                          &destination_attr);
+	bool dest_is_dir        = false;
+
+	if (destination_exists) {
+		if (destination_attr & DOS_ATTR_DIRECTORY) {
+			dest_is_dir = true;
+		} else if (final_sources.size() > 1) {
+			WriteOut(MSG_Get("SHELL_CMD_MOVE_MULTIPLE_TO_SINGLE"));
+			return;
+		}
+	} else if (final_sources.size() > 1) {
+		dest_is_dir = true;
+		if (!DOS_MakeDir(canonical_destination.c_str())) {
+			WriteOut(MSG_Get("SHELL_CMD_MKDIR_ERROR"),
+			         canonical_destination.c_str());
+			return;
+		}
+	}
+
+	if (dest_is_dir && canonical_destination.back() != DosSeparator) {
+		canonical_destination.push_back(DosSeparator);
+	}
+
+	for (const std::string& source : final_sources) {
+		std::string final_destination = canonical_destination;
+		if (dest_is_dir) {
+			final_destination += source.substr(
+			        source.rfind(DosSeparator) + 1);
+		}
+		if (source == final_destination) {
+			WriteOut(MSG_Get("SHELL_FILE_EXISTS"),
+			         final_destination.c_str());
+			continue;
+		}
+		// Overwrite existing file
+		if (destination_exists && !dest_is_dir) {
+			// Condition should be checked for previously
+			assert(final_sources.size() == 1);
+			if (!DOS_UnlinkFile(final_destination.c_str())) {
+				WriteOut(MSG_Get("SHELL_CMD_DEL_ERROR"),
+				         final_destination.c_str());
+				return;
+			}
+		}
+		// If same drive, do a rename. Otherwise it needs a copy + delete.
+		const auto source_drive_letter = source[0];
+		const auto dest_drive_letter   = final_destination[0];
+		if (source_drive_letter == dest_drive_letter) {
+			if (DOS_Rename(source.c_str(), final_destination.c_str())) {
+				WriteOut("%s => %s\n",
+				         source.c_str(),
+				         final_destination.c_str());
+			} else {
+				WriteOut(MSG_Get("SHELL_FILE_CREATE_ERROR"),
+				         final_destination.c_str());
+			}
+		} else {
+			uint16_t source_attr = 0;
+			if (!DOS_GetFileAttr(source.c_str(), &source_attr)) {
+				WriteOut("SHELL_FILE_NOT_FOUND", source.c_str());
+				continue;
+			}
+			uint16_t source_handle = 0;
+			if (!DOS_OpenFile(source.c_str(), OPEN_READ, &source_handle)) {
+				WriteOut("SHELL_FILE_OPEN_ERROR", source.c_str());
+				continue;
+			}
+			uint16_t dest_handle = 0;
+			if (!DOS_CreateFile(final_destination.c_str(),
+			                    source_attr,
+			                    &dest_handle)) {
+				WriteOut(MSG_Get("SHELL_FILE_CREATE_ERROR"),
+				         final_destination.c_str());
+				DOS_CloseFile(source_handle);
+				continue;
+			}
+			constexpr uint16_t buffer_capacity = 4096;
+			uint8_t buffer[buffer_capacity];
+			uint16_t bytes_requested = buffer_capacity;
+			bool success             = true;
+			do {
+				if (!DOS_ReadFile(source_handle, buffer, &bytes_requested)) {
+					WriteOut(MSG_Get("SHELL_READ_ERROR"),
+					         source.c_str());
+					success = false;
+					break;
+				}
+				uint16_t bytes_read = bytes_requested;
+				if (!DOS_WriteFile(dest_handle, buffer, &bytes_requested) ||
+				    bytes_requested != bytes_read) {
+					WriteOut(MSG_Get("SHELL_WRITE_ERROR"),
+					         final_destination.c_str());
+					success = false;
+					break;
+				}
+			} while (bytes_requested == buffer_capacity);
+
+			if (success) {
+				WriteOut("%s => %s\n",
+				         source.c_str(),
+				         final_destination.c_str());
+			}
+
+			DOS_CloseFile(source_handle);
+			DOS_CloseFile(dest_handle);
+			if (success) {
+				if (!DOS_UnlinkFile(source.c_str())) {
+					WriteOut(MSG_Get("SHELL_CMD_DEL_ERROR"),
+					         source.c_str());
+				}
+			}
+		}
+	}
 }
--- a/tests/cmd_move_tests.cpp
+++ b/tests/cmd_move_tests.cpp
@@ -1,0 +1,49 @@
+#include <gtest/gtest.h>
+
+struct cmd_move_arguments {
+	std::vector<std::string> sources = {};
+	std::string destination          = {};
+	const char* error                = nullptr;
+};
+
+cmd_move_arguments cmd_move_parse_arguments(const char* args);
+
+TEST(cmd_move_parse_arguments, multiple_sources)
+{
+	cmd_move_arguments ret = cmd_move_parse_arguments(
+	        "   a.bat   ,   b.bat        ,c.bat,d.bat  destdir");
+	EXPECT_EQ(ret.error, nullptr);
+	EXPECT_EQ(ret.destination, "destdir");
+	ASSERT_EQ(ret.sources.size(), 4);
+	EXPECT_EQ(ret.sources[0], "a.bat");
+	EXPECT_EQ(ret.sources[1], "b.bat");
+	EXPECT_EQ(ret.sources[2], "c.bat");
+	EXPECT_EQ(ret.sources[3], "d.bat");
+}
+
+TEST(cmd_move_parse_arguments, too_few_arguments)
+{
+	cmd_move_arguments ret = cmd_move_parse_arguments("a");
+	ASSERT_NE(ret.error, nullptr);
+	EXPECT_EQ(strcmp(ret.error, "SHELL_MISSING_PARAMETER"), 0);
+}
+
+TEST(cmd_move_parse_arguments, too_many_arguments)
+{
+	cmd_move_arguments ret = cmd_move_parse_arguments("a b c");
+	ASSERT_NE(ret.error, nullptr);
+	EXPECT_EQ(strcmp(ret.error, "SHELL_TOO_MANY_PARAMETERS"), 0);
+}
+
+TEST(cmd_move_parse_arguments, weird_quotes)
+{
+	// This actually works in MS-DOS 6.22
+	cmd_move_arguments ret = cmd_move_parse_arguments(
+	        "\"a.bat,\"  b.\"ba\"t  \",\"  c.bat   d\"estdi\"r");
+	EXPECT_EQ(ret.error, nullptr);
+	EXPECT_EQ(ret.destination, "destdir");
+	ASSERT_EQ(ret.sources.size(), 3);
+	EXPECT_EQ(ret.sources[0], "a.bat");
+	EXPECT_EQ(ret.sources[1], "b.bat");
+	EXPECT_EQ(ret.sources[2], "c.bat");
+}
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -41,6 +41,10 @@
 
 #define LFN_NAMELENGTH 255
 
+constexpr auto CurrentDirectory = ".";
+constexpr auto ParentDirectory  = "..";
+constexpr auto DosSeparator     = '\\';
+
 // obsolete - TODO: remove
 enum FatAttributeFlagsValues : uint8_t { // 7-bit
 	DOS_ATTR_READ_ONLY = 0b000'0001,
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -70,6 +70,7 @@ unit_tests = [
     {'name': 'batch_file', 'deps': [libmisc_stubs_dep, libshell_dep]},
     {'name': 'bit_view', 'deps': []},
     {'name': 'bitops', 'deps': []},
+    {'name': 'cmd_move', 'deps': [dosbox_dep], 'extra_cpp': []},
     {'name': 'dos_files', 'deps': [dosbox_dep], 'extra_cpp': []},
     {'name': 'drives', 'deps': [dosbox_dep], 'extra_cpp': []},
     {'name': 'fraction', 'deps': []},


src/shell/shell.cpp
```diff
@@ -647,6 +647,8 @@ void SHELL_Init() {
 	MSG_Add("SHELL_FILE_NOT_FOUND", "File not found - '%s'\n");
 	MSG_Add("SHELL_FILE_EXISTS", "File '%s' already exists.\n");
 	MSG_Add("SHELL_DIRECTORY_NOT_FOUND", "Directory not found - '%s'\n");
+	MSG_Add("SHELL_READ_ERROR", "Error reading from file - %s\n");
```
FeralChild64, line 650: 1. Change `%s` to `'%s'`, like in the strings above.
2. Wouldn't it sound better to write _Error reading file_ instead of _Error reading from file_, and similarly _Error writing file_?

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
 	        " Volume in drive %c is %s\n"
 	        " Volume Serial Number is %04X-%04X\n"
 	        "\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP",
+	        "Moves files and renames files and directories\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP_LONG",
+	        "Usage:\n"
+	        "To move one or more files:\n"
+	        "  [color=green]move[reset] [color=cyan][drive:][path]filename1[,...] destination[reset]\n"
```
FeralChild64, line 1268: `[color=cyan][drive:][path]filename1[,...]` -> `[color=cyan]FILE[reset] [[color=cyan]FILE[reset] ...]` - for uniformity with other commands
`destination` -> `[color=cyan]DESTINATION[reset]` - compare with COPY command help string

You can explain in the description that a file can contain a path.

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
...
+	        "\n"
+	        "To rename a directory:\n"
```
FeralChild64, line 1270: Remove `To move one or more files:\n` and `To rename a directory:\n` parts - if explanation is needed, put it in `Notes:` section.

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
...
+	        "  [color=green]move[reset] [color=cyan][drive:][path]dirname1 dirname2[reset]\n"
```
FeralChild64, line 1271: `[color=green]move[reset] [color=cyan]DIRECTORY1[reset] [color=cyan]DIRECTORY2[reset]\n` - for uniformity with `mkdir` command help.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,271 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	// Parse sources (seperated by commas)
+	while (parsing) {
+		if (num_sources >= ARRAY_LEN(raw_sources)) {
+			WriteOut(MSG_Get("SHELL_TOO_MANY_PARAMETERS"));
+			return;
+		}
+		switch (*args) {
+		case '"':
+			quote = !quote;
+			memmove(args, args + 1, strlen(args));
```
FeralChild64, line 2279: `memmove` and pointer arithmetics are dangerous, try to rework the code using `std::string` type (BTW, I know there is a lot of legacy code uses these techniques).

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,271 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	(void)ScanCMDBool(args, "-Y");
+
+	char* rem = ScanCMDRemain(args);
+	if (rem) {
+		WriteOut(MSG_Get("SHELL_ILLEGAL_SWITCH"), rem);
+		return;
+	}
+
+	size_t num_sources = 0;
+	char* raw_sources[32];
```
FeralChild64, line 2263: Use `std::vector`, make the table size dynamic

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,271 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		}
+	}
+
+	if (!*raw_destination) {
+		WriteOut(MSG_Get("SHELL_MISSING_PARAMETER"));
+		return;
+	}
+
+	// Check for extra arguments
+	while (*args) {
```
FeralChild64, line 2351: C-style argument array is prone to command line parser bugs. If you want to parse remaining arguments manually, I suggest to use vector of strings:

```c
std::vector<std::string> params;
cmd->FillVector(params);
...
```

include/shell.h
```diff
@@ -178,6 +178,7 @@ class DOS_Shell : public Program, public HostShell {
 	void CMD_VER(char* args);
 	void CMD_LS(char* args);
 	void CMD_VOL(char* args);
+	void CMD_MOVE(char* args);
```
johnnovak, line 181: Make it a pointer to a const string? (The signatures of the other commands should be the same but we gotta start somewhere)

src/shell/shell.cpp
johnnovak: Agree on both points.

src/shell/shell.cpp
johnnovak: Consistency is important.

src/shell/shell_cmds.cpp
johnnovak: Also please use the highest level functions thst operate on std::strings. This is not performance intensive code, we should absolutely optimise for simplicity, succinctness, and ease of understanding.

Look into the `split` string util function, that would be the best fit here for splitting a string by some fixed token (commas in your case).

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,271 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		}
+	} else if (final_sources.size() > 1) {
+		dest_is_dir = true;
+		if (!DOS_MakeDir(full_destination)) {
+			WriteOut(MSG_Get("SHELL_CMD_MKDIR_ERROR"), full_destination);
+			return;
+		}
+	}
+
+	if (dest_is_dir && full_destination[destination_len - 1] != '\\') {
```
johnnovak, line 2415: This could be generalised into an `is_dos_path` helper function, or at least extract the check into a boolean var so it's self-documenting. Maybe there is such a helper function already?

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,271 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		if (dest_is_dir) {
+			const char* filename = strrchr(source_c, '\\') + 1;
+			size_t filename_len  = strlen(filename);
+			if (filename_len + destination_len >=
+			    ARRAY_LEN(full_destination)) {
+				WriteOut(MSG_Get("SHELL_ILLEGAL_PATH"));
+				continue;
+			}
+			memcpy(full_destination + destination_len,
+			       filename,
```
johnnovak, line 2435: The ideal number of C string functions in new code is exactly zero 😄 Especially for non-performance intensive code. `const char*` is fine for string literals though that don't get manipulated.

Could you rewrite this using std::string only?

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
...
+	        "\n"
+	        "Notes:\n"
+	        "  If multiple source files are specified, destination must be a directory (or one will be created for you).\n"
+	        "  If a single source file is specified, it will overwrite destination, if it exists.\n"
+	        "\n"
+	        "Examples:\n"
+	        "  [color=green]move[reset] [color=white]source.bat[reset] [color=cyan]new.bat[reset]\n"
+	        "  [color=green]move[reset] [color=white]file1.txt,file2.txt[reset] [color=cyan]mydir[reset]\n");
+	MSG_Add("SHELL_CMD_MOVE_MULTIPLE_TO_SINGLE",
+	        "Cannot move multiple files to a single file\n");
```
johnnovak, line 1287: - Indentation after `filename` is off
- Lines in the notes section are too long
- Insert blank line above `Where:`

image

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
...
+	        "Where:\n"
+	        "  [color=white]filename[reset]      Can be either an exact filename or an inexact filename with\n"
+	        "              wildcards, which are the asterisk (*) and the question mark (?).\n"
+	        "              Multiple tokens can be provided and must be seperated by commas.\n"
+	        "  [color=cyan]destination[reset] An exact filename or directory, not containing any wildcards.\n"
+	        "  [color=cyan]dirname[reset] An exact directory name, not containing any wildcards.\n"
+	        "\n"
+	        "Notes:\n"
+	        "  If multiple source files are specified, destination must be a directory (or one will be created for you).\n"
+	        "  If a single source file is specified, it will overwrite destination, if it exists.\n"
```
johnnovak, line 1281: No comma needed before "if it exists".

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
...
+	        "Usage:\n"
+	        "To move one or more files:\n"
+	        "  [color=green]move[reset] [color=cyan][drive:][path]filename1[,...] destination[reset]\n"
+	        "\n"
+	        "To rename a directory:\n"
+	        "  [color=green]move[reset] [color=cyan][drive:][path]dirname1 dirname2[reset]\n"
+	        "Where:\n"
+	        "  [color=white]filename[reset]      Can be either an exact filename or an inexact filename with\n"
+	        "              wildcards, which are the asterisk (*) and the question mark (?).\n"
+	        "              Multiple tokens can be provided and must be seperated by commas.\n"
```
johnnovak, line 1275: "Multiple filenames can be provided"

A regular user wouldn't have a clue what "token" means.

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
 	        " Volume in drive %c is %s\n"
 	        " Volume Serial Number is %04X-%04X\n"
 	        "\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP",
+	        "Moves files and renames files and directories\n");
```
johnnovak, line 1264: Missing dot at the end.

src/shell/shell.cpp
johnnovak: Yes, and I think we already have enough explanation and examples in the provided help (🎉 for that one! 😄), so this is redundant and just uses up valuable vertical space.

src/shell/shell_cmds.cpp
johnnovak: 💯

johnnovak:
Very nice first cut implementation @weirddan455! The biggest thing that needs changing is moving to `std::string` wholesale, otherwise there are just fairly minor issues, mostly related to the help text. You're getting more comfortable with the codebase by the day! 🎉

weirddan455:
I'll give this another once over before it gets merged.  I still need to fix the help text.  But I've got rid of the mem* and str* functions and switched to `std::string` and friends.

include/shell.h
johnnovak: Ah okay, let's leave it then.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,259 @@ void DOS_Shell::CMD_VOL(char* args)
...
+			uint16_t dest_handle = 0;
+			if (!DOS_CreateFile(final_destination.c_str(),
+			                    source_attr,
+			                    &dest_handle)) {
+				WriteOut(MSG_Get("SHELL_FILE_CREATE_ERROR"),
+				         final_destination.c_str());
+				DOS_CloseFile(source_handle);
+				continue;
+			}
+			constexpr uint16_t buffer_capacity = 512;
```
johnnovak, line 2464: Probably better to use a 4k buffer?

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,259 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		default:
+			if (stop_next_letter) {
+				parsing          = comma;
+				comma            = false;
+				stop_next_letter = false;
+			} else {
+				word.push_back(*args++);
+			}
+		}
+	}
```
johnnovak, line 2308: Suggest to turn this into a proper state machine so it's easier to see the parsing logic. I've done something similar with `parse_cga_colors` in `int10_modes.cpp`.

Also, the parser part would be a good candidate for unit testing; you could extract the argument parsing into a separate pure function, then write a comprehensive test suite for it, just like I've done for `parse_cga_colors`.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,259 @@ void DOS_Shell::CMD_VOL(char* args)
...
+
+	if (raw_sources.empty()) {
+		WriteOut(MSG_Get("SHELL_MISSING_PARAMETER"));
+		return;
+	}
+
+	std::string raw_destination;
+	parsing = true;
+
+	// Parse detination
```
johnnovak, line 2318: typo: destination

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,259 @@ void DOS_Shell::CMD_VOL(char* args)
...
+				raw_destination.push_back(*args++);
+			} else {
+				args++;
+				parsing = false;
+			}
+			break;
+		case '\0': parsing = false; break;
+		default: raw_destination.push_back(*args++);
+		}
+	}
```
johnnovak, line 2336: Same comment as above about the parsing; at the very least this could go into a separate function to split this big function up a little bit into more logical chunks.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,259 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	(void)ScanCMDBool(args, "-Y");
+
+	char* rem = ScanCMDRemain(args);
+	if (rem) {
+		WriteOut(MSG_Get("SHELL_ILLEGAL_SWITCH"), rem);
+		return;
+	}
+
+	bool parsing          = true;
+	bool quote            = false;
```
johnnovak, line 2263: `inside_quote` would be better.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,259 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	bool stop_next_letter = false;
+	bool comma            = false;
```
johnnovak, line 2265: `curr_token_is_comma` or something, but with the proposed better parsing logic (see below) this would disappear.

johnnovak:
> I'll give this another once over before it gets merged. I still need to fix the help text. But I've got rid of the mem* and str* functions and switched to `std::string` and friends.

Nice one. The parsing stuff could be extracted from the big function to split it up into logical chunks a little (as per my new comments), and consider turning the parser into a proper state machine and adding unit tests for it. Might be overkill a bit, I'm 50/50 on that. But it wouldn't hurt and it would be good practice.

src/shell/shell_cmds.cpp
weirddan455: Sure, I can change it.  I wasn't sure what size was optimal.  I just looked at `CMD_COPY` and it uses a 32k buffer (statically allocated because they were worried about stack overflow):

https://github.com/dosbox-staging/dosbox-staging/blob/46c10f073091f088315d7b1dafbab1758e96caad/src/shell/shell_cmds.cpp#L1303

johnnovak: Well, that copy command is a horrible mess, let's not use it as an example for anything 😅

weirddan455:
I pushed a commit to fix the help + error text.  For some reason, it shows on the branch:  https://github.com/dosbox-staging/dosbox-staging/tree/wd/move

But has not yet shown up on this PR.  So... idk what's going on there.

I'll take a look at that parsing example and try to refactor mine to be a bit nicer like you suggested.

move help

johnnovak:
> I pushed a commit to fix the help + error text. For some reason, it shows on the branch: https://github.com/dosbox-staging/dosbox-staging/tree/wd/move

The help text looks great now! 🎉 It even fits nicely into a single page! 🥇 

> I'll take a look at that parsing example and try to refactor mine to be a bit nicer like you suggested.

Sounds good!

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
 	        " Volume in drive %c is %s\n"
 	        " Volume Serial Number is %04X-%04X\n"
 	        "\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP",
+	        "Moves files and renames files and directories.\n");
+	MSG_Add("SHELL_CMD_MOVE_HELP_LONG",
+	        "Usage:\n"
+	        "  [color=green]move[reset] [color=white]FILENAME1,FILENAME2[,...][reset] [color=cyan]DESTINATION[reset]\n"
```
FeralChild64, line 1267: `[color=white]FILENAME1,FILENAME2[,...][reset]` - shouldn't it rather be `[color=white]FILENAME1[,FILENAME2,...][reset]`?

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: I just removed "if it exists" so it fits in a single line (kind of obvious you can't overwrite something that doesn't exist)

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Changed to 4k buffer

src/shell/shell.cpp
weirddan455: Sure, just changed it.

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: I renamed it to `continue_after_stop` since that's really what it's doing.  It's a flag for whether the parser should continue parsing the next word or if it should finish.  If it sees a comma in between the words, it continues.  If not, it stops.

weirddan455:
@johnnovak I refactored the parsing functions out into separate functions and added a few unit tests.  I'm not really sure how to make it a real state machine.  The state is really just 3 boolean variables.

Compared to your example `parse_cga_colors`, mine is mostly just doing the "tokenize" step.  The actual parsing of DOS paths, filenames, and wildcards is done by a helper function `DOS_Canonicalize` later on.

tests/cmd_move_tests.cpp
```diff
@@ -0,0 +1,49 @@
...
+	// This actually works in MS-DOS 6.22
+	cmd_move_arguments ret = cmd_move_parse_arguments(
+	        "\"a.bat,\"  b.bat  \",\"  c.bat   destdir");
+	EXPECT_EQ(ret.error, nullptr);
+	EXPECT_EQ(ret.destination, "destdir");
+	ASSERT_EQ(ret.sources.size(), 3);
+	EXPECT_EQ(ret.sources[0], "a.bat");
+	EXPECT_EQ(ret.sources[1], "b.bat");
+	EXPECT_EQ(ret.sources[2], "c.bat");
+}
```
weirddan455, line 49: Btw this is why the parser might look kind of janky.  I tested against MS-DOS 6.22 and it works with things like this.

tests/cmd_move_tests.cpp
```diff
@@ -0,0 +1,49 @@
...
+	// This actually works in MS-DOS 6.22
+	cmd_move_arguments ret = cmd_move_parse_arguments(
+	        "\"a.bat,\"  b.\"ba\"t  \",\"  c.bat   d\"estdi\"r");
+	EXPECT_EQ(ret.error, nullptr);
+	EXPECT_EQ(ret.destination, "destdir");
+	ASSERT_EQ(ret.sources.size(), 3);
+	EXPECT_EQ(ret.sources[0], "a.bat");
+	EXPECT_EQ(ret.sources[1], "b.bat");
+	EXPECT_EQ(ret.sources[2], "c.bat");
+}
```
weirddan455, line 49: Btw this is why the parser might look kind of janky. I tested against MS-DOS 6.22 and it works with things like this.

johnnovak: Huh, that's surprising! But excellent job matching the behaviour of the quirky MS-DOS parser! 💯

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	bool parsing             = true;
+	bool inside_quote        = false;
+	bool stop_next_letter    = false;
+	bool continue_after_stop = false;
+	std::string word;
+	std::vector<std::string> sources;
+
+	// Sources are seperated by commas and terminated by a trailing space
+	while (parsing) {
+		switch (*args) {
```
johnnovak, line 2258: Would be more descriptive to set up a simple `ch` reference to args and use that in the parser.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
 	WriteOut(MSG_Get("SHELL_CMD_VOL_OUTPUT"), drive, trimmed_label, high, low);
 }
+
+static std::vector<std::string> cmd_move_parse_sources(const char*& args)
+{
+	bool parsing             = true;
+	bool inside_quote        = false;
+	bool stop_next_letter    = false;
+	bool continue_after_stop = false;
+	std::string word;
```
johnnovak, line 2253: Rename to `source`?

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+			if (inside_quote) {
+				word.push_back(*args++);
+			} else {
+				if (!word.empty()) {
+					sources.emplace_back(std::move(word));
+					word             = {};
+					stop_next_letter = true;
+					continue_after_stop = false;
+				}
+				args++;
```
johnnovak, line 2273: Minor: prefer pre-increment by convention.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		switch (*args) {
+		case '"':
+			inside_quote = !inside_quote;
+			args++;
+			break;
+		case ' ':
+			if (inside_quote) {
+				destination.push_back(*args++);
+			} else {
+				args++;
```
johnnovak, line 2320: Pre-increment

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+}
+
+static std::string cmd_move_parse_destination(const char*& args)
+{
+	std::string destination;
+	bool parsing      = true;
+	bool inside_quote = false;
+
+	while (parsing) {
+		switch (*args) {
```
johnnovak, line 2311: Introduce `ch` alias

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		}
+	}
+	return destination;
+}
+
+struct cmd_move_arguments {
+	std::vector<std::string> sources = {};
+	std::string destination          = {};
+	const char* error                = nullptr;
+};
```
johnnovak, line 2335: 🎖️ for using a struct for multiple return types (I see it as the "poor man's data class" in C++ 😅 )

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	// Command uses dta so set it to our internal dta
+	const RealPt save_dta = dos.dta();
+	dos.dta(dos.tables.tempdta);
+	DOS_DTA dta(dos.dta());
+
+	// Each raw source can be a wildcard (*.*) containing multiple real
+	// sources.
+	std::vector<std::string> final_sources;
+	for (const std::string& source : parsed.sources) {
+		if (!DOS_FindFirst(source.c_str(), 0xffff & ~DOS_ATTR_VOLUME)) {
```
johnnovak, line 2398: Could we do better here with the bit manipulations? Look into `bitops.h` and perhaphs `bit_view.h`.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		if (!DOS_Canonicalize(source.c_str(), temp)) {
+			WriteOut(MSG_Get("SHELL_FILE_NOT_FOUND"), source.c_str());
+			continue;
+		}
+		std::string source_dir = temp;
+		source_dir = source_dir.substr(0, source_dir.rfind('\\') + 1);
+		do {
+			DOS_DTA::Result result = {};
+			dta.GetResult(result);
+			if (result.name != "." && result.name != "..") {
```
johnnovak, line 2411: Aren't these defined as common constants somewhere? If not, should we introduce constants for them?

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+	} else if (final_sources.size() > 1) {
+		dest_is_dir = true;
+		if (!DOS_MakeDir(canonical_destination.c_str())) {
+			WriteOut(MSG_Get("SHELL_CMD_MKDIR_ERROR"),
+			         canonical_destination.c_str());
+			return;
+		}
+	}
+
+	if (dest_is_dir && canonical_destination.back() != '\\') {
```
johnnovak, line 2441: Don't we have DOS filename/path helper functions already that identify directory-paths by looking at the last char? If not, probably time to introduce one.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		canonical_destination.push_back('\\');
```
johnnovak, line 2442: Don't we have an MS-DOS path separator constant defined somewhere already?

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+			// Condition should be checked for previously
+			assert(final_sources.size() == 1);
+			if (!DOS_UnlinkFile(final_destination.c_str())) {
+				WriteOut(MSG_Get("SHELL_CMD_DEL_ERROR"),
+				         final_destination.c_str());
+				return;
+			}
+		}
+		// If same drive, do a rename. Otherwise it needs a copy + delete.
+		if (source[0] == final_destination[0]) {
```
johnnovak, line 2466: Continuing the above line of thought: don't we have a set of MS-DOS path manipulation functions already? If not, I'd prefer adding them instead of always doing low-level string manipulations. I'll look into this later (rewieving now on my phone 😄 ). Ideally, we'd have a DosPath class or something, with similar methods to std::fs_path.

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+			do {
+				if (!DOS_ReadFile(source_handle, buffer, &bytes_requested)) {
+					WriteOut(MSG_Get("SHELL_READ_ERROR"),
+					         source.c_str());
+					success = false;
+					break;
+				}
+				uint16_t bytes_read = bytes_requested;
+				if (!DOS_WriteFile(dest_handle, buffer, &bytes_requested) ||
+				    bytes_requested != bytes_read) {
```
johnnovak, line 2508: Does this work when moving 0 byte files?

tests/cmd_move_tests.cpp
johnnovak: Also, very nice set of tests 🎉

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,290 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		return ret;
+	}
+	ret.destination = cmd_move_parse_destination(args);
+	if (ret.destination.empty()) {
+		ret.error = "SHELL_MISSING_PARAMETER";
+		return ret;
+	}
+	// Check for extra arguments
+	while (*args) {
+		if (*args++ != ' ') {
```
johnnovak, line 2352: At least we should document that the parse methods have a side-effect of mutating the passed in char pointer, and subsequent parsing relies on this behaviour. At the very least I'd still not destroy the args pointer passed to the main parse function.

weirddan455: Just to be clear, it the actual string behind the `char *` never gets mutated.  The main function takes a `const char *`.  The 2 helper functions take `const char*&` so that they can advance the pointer past what's already been parsed.

I can certainly add a comment stating that.  Or if you have a better idea in mind I'm all ears.

johnnovak: > Just to be clear, it the actual string behind the `char *` never gets mutated. The main function takes a `const char *`. The 2 helper functions take `const char*&` so that they can advance the pointer past what's already been parsed.
> 
> I can certainly add a comment stating that. Or if you have a better idea in mind I'm all ears.

Yeah I get that; it's a mutable pointer to a const char string. In general, I dislike mutating passed in arguments. This might seem a petty thing in this case, and you can figure out what's happening in a minute or two, so I'm not asking you to change the whole thing, just document that the passed-in args pointer get advanced, and further parse functions rely on that behavoiur.

The cleanest solution would be to pass back a `string_view` of the remaining part. If this was a complicated parses with tens of parse functions, I'd definitely say do that. But it would be an overkill in this case, so just document it please.

src/shell/shell_cmds.cpp
weirddan455: Sorry, I'm not sure what you mean by that exactly.

src/shell/shell_cmds.cpp
weirddan455: Well, it's marked obsolete here and there's a newer bit_view implementation:  https://github.com/dosbox-staging/dosbox-staging/blob/d64ad72581e08e0c48b9dcc9d5da7703f03fb8b8/include/dos_system.h#L44-L75

But from grepping the callers of this specific function, all except one (in `program_tree)` use this old way.  It also takes in a `uint16_t` even though it's supposed to be a 6 bit field?  I'm not really sure what's going on here.  The function probably needs  a refactor at some point.

src/shell/shell_cmds.cpp
johnnovak: I was just thinking `const auto curr_char = *args` or simply just `ch` is more descriptive. Then you can just advance the pointer with `++args` somewhere still.

src/shell/shell_cmds.cpp
weirddan455: All I see is this `DOS_MakeName` function that returns a drive number.  That gets used by `DOS_Canonicalize` so by the time we get here, it's got a drive letter as the path's first letter.  I'm not seeing anything like what you're suggesting, though it's possible I'm missing it.

https://github.com/dosbox-staging/dosbox-staging/blob/d64ad72581e08e0c48b9dcc9d5da7703f03fb8b8/src/dos/dos_files.cpp#L797-L809

src/shell/shell_cmds.cpp
weirddan455: I'm not seeing a helper function but maybe I don't know where to look.  The issue is you can't really tell if something is a directory just by looking at the last char.  From testing, after running `DOS_Canonicalize`, a path to a sub-directory does not end with a slash (ex. would look like `C:\mydir`)  but a path to the root drive does (ex. `C:\`).  That's why I have the flag variable to check if it's a directory first and then append a trailing slash if it's not there already.

src/shell/shell_cmds.cpp
weirddan455: Not that I can find.

src/shell/shell_cmds.cpp
weirddan455: I just tested.  It works.  There's a `DOS_CreateFile` call above that makes the new empty file and then 0 bytes get copied here without throwing any errors.

src/shell/shell_cmds.cpp
weirddan455: Defined a constant for it at the top of this file (wasn't sure where to put it)

src/shell/shell_cmds.cpp
weirddan455: Got it.  Changed.

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell_cmds.cpp
weirddan455: I defined some constants at the top of the file (couldn't find them and didn't know where else to put them)

src/shell/shell_cmds.cpp
weirddan455: Documented

src/shell/shell_cmds.cpp
johnnovak: Good enough for now 😄

src/shell/shell_cmds.cpp
weirddan455: Comment removed since it's in its own function now

src/shell/shell_cmds.cpp
johnnovak: Can you move them into `dos_system.h`, please? The casing style of the constants might seem chaotic, but we arrived at a convention of using UpperCamelCase for new global constants.
```cpp

constexpr auto DosCurrentDirectory = ".";
constexpr auto DosParentDirectory  = "..";
constexpr auto DosSeparator        = '\\';
```

src/shell/shell_cmds.cpp
johnnovak: Yeah fair enough, it was just an idea. Using `DosSeparator` will be an improvement anyway.

src/shell/shell_cmds.cpp
johnnovak: Yeah, it's a mess 😩 In `DOS_FindFirst` the `uint16_t attr` only gets compared to `DOS_ATTR_VOLUME` (which is `uint8_t`) and then it's cast to `uint8_t` anyway (without `check_cast`) here:

```cpp
dta.SetupSearch(drive,(uint8_t)attr,pattern);
```

So it is safe to use the `FatAttributeFlags` bit_view; please do that and then just cast the result to `uint16_t` on the way in... It's a start 🤷🏻 

Btw, I've tried changing `attr` to `uint8_t` in `DOS_FindFirst`, but it results in too many warnings. I agree that cleaning this up should be a separate piece of work. If you could create a ticket for it to migrate to `FatAttributeFlags`, that would be much appreciated!

image

src/shell/shell_cmds.cpp
weirddan455: Sure, I just moved them.

src/shell/shell_cmds.cpp
johnnovak: Okay, probably job for a later refactoring step. It's fine as-is then.

src/shell/shell.cpp
```diff
@@ -1258,6 +1260,31 @@ void SHELL_Init() {
...
+	        "              wildcards, which are the asterisk (*) and the question mark (?).\n"
+	        "              Multiple, comma-seperated, filenames can be provided.\n"
+			"  [color=white]DIRECTORY[reset]   An exact directory name, not containing any wildcards.\n"
+	        "  [color=cyan]DESTINATION[reset] An exact filename or directory, not containing any wildcards.\n"
+	        "\n"
+	        "Notes:\n"
+	        "  If multiple source files are specified, destination must be a directory.\n"
+			"  If not, one will be created for you.\n"
+			"\n"
+	        "  If a single source file is specified, it will overwrite destination.\n"
```
johnnovak, line 1281: Indentation is off on a few lines (just run it through clang format).

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,298 @@ void DOS_Shell::CMD_VOL(char* args)
...
+struct cmd_move_arguments {
+	std::vector<std::string> sources = {};
+	std::string destination          = {};
+	const char* error                = nullptr;
+};
+
+cmd_move_arguments cmd_move_parse_arguments(const char* args)
+{
+	cmd_move_arguments ret;
+	// args pointer gets advanced by the two helper functions
```
johnnovak, line 2345: 👍🏻

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,298 @@ void DOS_Shell::CMD_VOL(char* args)
...
+		} while (DOS_FindNext());
+	}
+
+	// Done with DTA. Restore it.
+	dos.dta(save_dta);
+
+	uint16_t destination_attr = 0;
+	bool destination_exists = DOS_GetFileAttr(canonical_destination.c_str(),
+	                                          &destination_attr);
+	bool dest_is_dir       = false;
```
johnnovak, line 2430: Minor: remove extra spaces (just run the whole function through clang format, it should fix everything up).

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,298 @@ void DOS_Shell::CMD_VOL(char* args)
...
+			// Condition should be checked for previously
+			assert(final_sources.size() == 1);
+			if (!DOS_UnlinkFile(final_destination.c_str())) {
+				WriteOut(MSG_Get("SHELL_CMD_DEL_ERROR"),
+				         final_destination.c_str());
+				return;
+			}
+		}
+		// If same drive, do a rename. Otherwise it needs a copy + delete.
+		if (source[0] == final_destination[0]) {
```
johnnovak, line 2474: This would describe the intent better:

```cpp
const auto src_drive_letter = source[0];
const auto dest_drive_letter = final_destination[0];

if (src_drive_letter == dest_drive_letter) {
...
```

src/shell/shell_cmds.cpp
johnnovak: Excellent 👍🏻

src/shell/shell_cmds.cpp
```diff
@@ -2242,3 +2243,298 @@ void DOS_Shell::CMD_VOL(char* args)
...
+			constexpr uint16_t buffer_capacity = 4096;
+			uint8_t buffer[buffer_capacity];
+			uint16_t bytes_requested = buffer_capacity;
+			bool success             = true;
+			do {
+				if (!DOS_ReadFile(source_handle, buffer, &bytes_requested)) {
+					WriteOut(MSG_Get("SHELL_READ_ERROR"),
+					         source.c_str());
+					success = false;
+					break;
```
johnnovak, line 2512: As there's nothing more to do in this function after this, I'd prefer a `return` here.

src/shell/shell_cmds.cpp
weirddan455: How do you suggest using it exactly?  I'm not really familiar with this type of interface.  In this case, it wants every bit set except for volume.  So do you have to do:

``` cpp
FatAttributeFlags attr;
attr.read_only = true;
attr.hidden = true;
attr.system = true;
attr.volume = false;
attr.directory = true;
attr.archive = true;
attr.device = true;
attr.unused = idk... it's unsued?
```
Seems a bit overly verbose to me?

src/shell/shell_cmds.cpp
weirddan455: I think clang format is what put those spaces there, trying to get the equal signs to line up.

src/shell/shell_cmds.cpp
johnnovak: Yes, that's too verbose. These two variants both work:

```cpp
FatAttributeFlags attr = {0};
bit::flip_all(attr._data);
attr.volume.flip();
```
```cpp
FatAttributeFlags attr = {0xff};
attr.volume.flip();
```

I'd say use the second form for now, then we can look into improving `bit_view` for such operations later (to init with all bits set without having to specify `0xff`).

src/shell/shell_cmds.cpp
weirddan455: Fixed

src/shell/shell.cpp
weirddan455: Just did the big squash + re-run scripts/format-commit.sh.  Looks like it changed these lines.

src/shell/shell_cmds.cpp
weirddan455: Thanks.  Yeah, that looks nicer.  Fixed.

johnnovak: No problem! Are you going to create a ticket for refactoring everything to move to 'FatAttributeFlags'? If not that's fine, I just need to know then I'll do it myself later 😄

johnnovak (approved):
Top notch work @weirddan455 ! 🎉 Everything looks good and I saw you've squashed your changes into a single commit. Feel free to merge away 👍🏻

src/shell/shell_cmds.cpp
weirddan455: There you go:  https://github.com/dosbox-staging/dosbox-staging/issues/2663

weirddan455:
@johnnovak Well, I don't have merge permissions so you have to do it :smile:

src/shell/shell_cmds.cpp
johnnovak: Cheers @weirddan455 !

johnnovak:
> @johnnovak Well, I don't have merge permissions so you have to do it 😄

Ah yes, I've merged it. Again, well done! 🚀

PR #2661, opened 2023-07-04
Add several warnings to the GCC build

MeAreJeenius:
Not an exhaustive list by any means. Probably lots of low-hanging fruit left. Definitely lots of other warnings that should be added, but will require a bunch of work.

All of the warnings requiring fixes were pretty trivial fixes, so no refactoring or changes in logic were done here.


src/cpu/core_normal/table_ea.h
```diff
@@ -136,10 +136,10 @@ static GetEAHandler EATable[512]={
...
 /* 11 These are illegal so make em 0 */
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
 /* 00 */
```
johnnovak, line 143: Maybe just use single spaces as separators?

src/cpu/core_normal/table_ea.h
```diff
@@ -168,10 +168,10 @@ static GetEAHandler EATable[512]={
...
 	EA_32_80_n,EA_32_81_n,EA_32_82_n,EA_32_83_n,EA_32_84_n,EA_32_85_n,EA_32_86_n,EA_32_87_n,
 /* 11 These are illegal so make em 0 */
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,
-	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0,	0
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr
```
johnnovak, line 174: Maybe just use single spaces as separators?

src/cpu/modrm.cpp
```diff
@@ -137,18 +137,18 @@ uint32_t * lookupRMregd[256]={
...
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
+	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,
```
johnnovak, line 151: Maybe just use single spaces as separators?

src/cpu/modrm.cpp
```diff
@@ -161,18 +161,18 @@ uint8_t * lookupRMEAregb[256]={
...
```
johnnovak, line 175: Maybe just use single spaces as separators? (same comment for the rest 😄 )

src/debug/debug_disasm.cpp
```diff
@@ -311,28 +311,28 @@ static char const *second[] = {
   "setl %Eb",         "setge %Eb",       "setle %Eb",      "setg %Eb",
 /* a */
   "push fs",          "pop fs",          "cpuid",          "bt %Ev,%Gv",
-  "shld %Ev,%Gv,%Ib", "shld %Ev,%Gv,cl", 0,                0,
-  "push gs",          "pop gs",          0,                "bts %Ev,%Gv",
-  "shrd %Ev,%Gv,%Ib", "shrd %Ev,%Gv,cl", 0,                "imul %Gv,%Ev",
+  "shld %Ev,%Gv,%Ib", "shld %Ev,%Gv,cl", nullptr,                nullptr,
+  "push gs",          "pop gs",          nullptr,                "bts %Ev,%Gv",
+  "shrd %Ev,%Gv,%Ib", "shrd %Ev,%Gv,cl", nullptr,                "imul %Gv,%Ev",
```
johnnovak, line 316: Indentation seems to be off with the rest.

src/debug/debug_disasm.cpp
```diff
@@ -311,28 +311,28 @@ static char const *second[] = {
...
   "cmpxchg %Eb,%Gb",  "cmpxchg %Ev,%Gv", "lss %Mp",        "btr %Ev,%Gv",
   "lfs %Mp",          "lgs %Mp",         "movzx %Gv,%Eb",  "movzx %Gv,%Ew",
-  0,                  0,                 "%g7 %Ev,%Ib",    "btc %Ev,%Gv",
+  nullptr,                  nullptr,                 "%g7 %Ev,%Ib",    "btc %Ev,%Gv",
   "bsf %Gv,%Ev",      "bsr %Gv,%Ev",     "movsx %Gv,%Eb",  "movsx %Gv,%Ew",
 /* c */
-  "xadd %Eb,%Gb",     "xadd %Ev,%Gv",    0,                0,
-  0,                  0,                 0,                0,
+  "xadd %Eb,%Gb",     "xadd %Ev,%Gv",    nullptr,                nullptr,
+  nullptr,                  nullptr,                 nullptr,                nullptr,
```
johnnovak, line 324: Indentation seems to be off with the rest.

src/debug/debug_disasm.cpp
```diff
@@ -346,38 +346,38 @@ static char const *groups[][8] = {   /* group 0 is group 3 for %Ev set */
...
   { "sldt %Ew",       "str %Ew",         "lldt %Ew",       "ltr %Ew",
-    "verr %Ew",       "verw %Ew",        0,                0               },
+    "verr %Ew",       "verw %Ew",        nullptr,                nullptr               },
 /* 6 */
   { "sgdt %Ms",       "sidt %Ms",        "lgdt %Ms",       "lidt %Ms",
-    "smsw %Ew",       0,                 "lmsw %Ew",       "invlpg"        },
+    "smsw %Ew",       nullptr,                 "lmsw %Ew",       "invlpg"        },
 /* 7 */
-  { 0,                0,                 0,                0,
+  { nullptr,                nullptr,                 nullptr,                nullptr,
```
johnnovak, line 361: Indentation seems to be off with the rest.

src/gui/render_scalers.cpp
```diff
@@ -179,17 +179,17 @@ ScalerSimpleBlock_t ScaleNormal1x = {
...
 },{
 {	Normal1x_8_8_R,		Normal1x_8_15_R ,	Normal1x_8_16_R ,	Normal1x_8_32_R },
-{	             0,		Normal1x_15_15_R,	Normal1x_15_16_R,	Normal1x_15_32_R},
-{	             0,		Normal1x_16_15_R,	Normal1x_16_16_R,	Normal1x_16_32_R},
-{	             0,		Normal1x_24_15_R,	Normal1x_24_16_R,	Normal1x_24_32_R},
-{	             0,		Normal1x_32_15_R,	Normal1x_32_16_R,	Normal1x_32_32_R},
+{	             nullptr,		Normal1x_15_15_R,	Normal1x_15_16_R,	Normal1x_15_32_R},
+{	             nullptr,		Normal1x_16_15_R,	Normal1x_16_16_R,	Normal1x_16_32_R},
+{	             nullptr,		Normal1x_24_15_R,	Normal1x_24_16_R,	Normal1x_24_32_R},
+{	             nullptr,		Normal1x_32_15_R,	Normal1x_32_16_R,	Normal1x_32_32_R},
```
johnnovak, line 192: Indentation seems to be off with the rest?

src/gui/render_scalers.cpp
```diff
@@ -219,16 +219,16 @@ ScalerSimpleBlock_t ScaleNormalDh = {
...
 },{
 {	NormalDh_8_8_R,		NormalDh_8_15_R ,	NormalDh_8_16_R ,	NormalDh_8_32_R },
-{	             0,		NormalDh_15_15_R,	NormalDh_15_16_R,	NormalDh_15_32_R},
-{	             0,		NormalDh_16_15_R,	NormalDh_16_16_R,	NormalDh_16_32_R},
-{	             0,		NormalDh_24_15_R,	NormalDh_24_16_R,	NormalDh_24_32_R},
-{	             0,		NormalDh_32_15_R,	NormalDh_32_16_R,	NormalDh_32_32_R},
+{	             nullptr,		NormalDh_15_15_R,	NormalDh_15_16_R,	NormalDh_15_32_R},
+{	             nullptr,		NormalDh_16_15_R,	NormalDh_16_16_R,	NormalDh_16_32_R},
+{	             nullptr,		NormalDh_24_15_R,	NormalDh_24_16_R,	NormalDh_24_32_R},
+{	             nullptr,		NormalDh_32_15_R,	NormalDh_32_16_R,	NormalDh_32_32_R},
```
johnnovak, line 232: Indentation seems to be off with the rest?

include/reelmagic.h
```diff
@@ -36,17 +36,6 @@ struct ReelMagic_VideoMixerMPEGProvider {
...
-
-void ReelMagic_RENDER_SetSize(const uint16_t width, const uint16_t height,
-                              const bool double_width, const bool double_height,
-                              const Fraction& pixel_aspect_ratio,
-                              const uint8_t bits_per_pixel,
-                              const double frames_per_second);
-
-bool ReelMagic_RENDER_StartUpdate(void);
-// void ReelMagic_RENDER_EndUpdate(bool abort);
-// void ReelMagic_RENDER_DrawLine(const void *src);
```
johnnovak, line 49: Are these unused? If they're internal-only, ideally we should change their names (by convention, static functions are in `lower_snake_case` and they're not prefixed by the module name; e.g. `RENDER_DoStuff` would become `do_stuff`).

johnnovak:
Very nice @MeAreJeenius, that's massive! 🎉 Just a few comments/questions/remarks, otherwise should be good to go once we sort those few things out!

include/reelmagic.h
johnnovak: Are you sure removing these is ok? Just removed them in `main` and I'm getting compilation errors:

```
FAILED: src/hardware/libhardware.a.p/vga_other.cpp.o
ccache c++ -Isrc/hardware/libhardware.a.p -Isrc/hardware -I../../src/hardware -I../../include -I. -I../.. -Isrc/libs/ghc -I../../src/libs/ghc -Isubprojects/iir1-1.9.3 -I../../subprojects/iir1-1.9.3 -Isrc/libs/loguru -I../../src/libs/loguru -I/opt/local/include/slirp -I/opt/local/include/glib-2.0 -I/opt/local/lib/glib-2.0/include -I/opt/local/include -I/opt/local/include/libpng16 -I/opt/local/include/SDL2 -fcolor-diagnostics -Wall -Winvalid-pch -Wextra -Wpedantic -std=c++17 -O0 -g -Wno-unknown-pragmas -fpch-preprocess -Weffc++ -Wextra-semi -Wno-deprecated-declarations -fPIC -DLOGURU_STACKTRACES=1 -DLOGURU_USE_LOCALE=0 -D_THREAD_SAFE -MD -MQ src/hardware/libhardware.a.p/vga_other.cpp.o -MF src/hardware/libhardware.a.p/vga_other.cpp.o.d -o src/hardware/libhardware.a.p/vga_other.cpp.o -c ../../src/hardware/vga_other.cpp
../../src/hardware/vga_other.cpp:585:4: error: use of undeclared identifier 'ReelMagic_RENDER_SetPal'
                        RENDER_SetPal(index, r, g, b);
                        ^
../../src/hardware/reelmagic/vga_passthrough.h:26:30: note: expanded from macro 'RENDER_SetPal'
#define RENDER_SetPal        ReelMagic_RENDER_SetPal
                             ^
1 error generated.
[11/19] Compiling C++ object src/hardware/libhardware.a.p/reelmagic_video_mixer.cpp.o
```

src/cpu/core_normal/table_ea.h
johnnovak: > Is there any reason I shouldn't just apply clang-format to the whole project except the libs directory in a separate commit?

Yes, don't do that. It can make porting changes over from DOSBox SVN harder than it needs to be.

The "informal" rule of thumb is to reformat only files you're touching, and even then to restrict it to the parts you're touching. And of course ideally always in separate commits.

So we're doing the clang reformatting "lazily", on demand.

include/reelmagic.h
johnnovak: Hmm, the branch itself compiles fine locally here. Not quite sure what's going on...

src/gui/render_scalers.cpp
johnnovak: This is how the indentation looks in current `main`:

image

After the reformat: 

image

Can you please fix this up so the nice tabular layout is kept? Then enclose the tabular block with `// clang-format off` and `// clang-format on` to exclude it from the Clang formatter? Check out `int10_modes.cpp` for a concrete example.

src/cpu/modrm.cpp
johnnovak: Ok these are fine actually.

src/debug/debug_disasm.cpp
johnnovak: This looked like this before:

image

Now it looks like this; the tabular formatting is lost:

image

johnnovak:
@MeAreJeenius I went through some of the formatting issues I commented on you marked as resolved, and posted screenshot examples of how the clang formatter ruined the nice tabular format.

I appreciate your proactive work on getting rid of the warnings, but removing these well-crafted tabular layouts and leaving restoring the layout to the next person who's gonna touch these files in the future is a no-go. We should not leave the files in a worse state after touching them, and the tabular layout is important, it makes things more apparent and easier to parse visually.

Can you please fix up the layouts and restore them to their former glory? 😄  After that, please fence them off with `// clang-format off` and `// clang-format on` (see `int10_modes.cpp` for a concrete example). Thank you!

johnnovak:
> @johnnovak I marked them resolved to make the page easier to read through. It's a lot of repetition, and I've already fixed them locally.

Ah, right, thanks for clarifying that 😅 Yeah we don't really have an agreed upon standard way of resolving comments, so I kinda thought you were just dismissing them. Apologies 😄

src/cpu/core_normal/table_ea.h
LowLevelMahn: there is a script that formats only the "changed lines" of the last commit - see: script/format-commit.sh

works for me on Windows only in a MSYS2 environment :/

1. make you're changes
2. commit
3. use the script/format-commit.sh - every changed line gets formatted
(you still need to keep out these tables johnnovak is talking about)
4. rebase or squash the formatted changes back into in one commit

johnnovak: That's cool, I never tried it myself. I use a Vim plugin for Clang Format and pretty much reformat what I've changed manually 100% of the time.

include/reelmagic.h
johnnovak: I think @kcgen did not want to change the code much so it's easier to merge further improvements. But my understanding is that the original author will update the version in our repo (if he'll ever get to it). I vaguely remember some correspondence between @kcgen and the author about that. So probably it's okay to make a few judicious improvements to it (and I had to touch it too a bit because of my recent image capture / video related refactorings & improvements).

src/cpu/core_normal/table_ea.h
LowLevelMahn: i was forced to use it - for my micro pulls :) - nothing compared to your output

johnnovak:
> I changed how the warnings are set in meson. Now they are in extra_args, which filters out unavailable warnings for the current compiler, so different compilers are no longer an issue.
> 
> To compensate, I had to disable warnings altogether on some third-party code in src/libs. Not ideal, but meson really doesn't want to make this easy. Definitely open to suggestions if there is a nicer way.

I'm really a beginner with Meson, so I defer to the others' opinion on this. @shermp if you think it's fine, then it's fine by me too 😄

johnnovak:
This is a heroic cleanup job @MeAreJeenius , congrats 🥇 Formatting of tabular stuff looks now good, thanks for addressing that!

Would these changes allow us to lower our allowed warning limits? We've done that periodically, e.g. check out these commits 49a80c3b373c0e2,  9a374d88bc5b

If not, that's fine too, just checking!

johnnovak:
How are you going with this one @MeAreJeenius ?

.github/workflows/linux.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           - name: GCC, Debian 10, ARMv7
             os: ubuntu-20.04
             build_flags: -Dbuildtype=debug -Dexperimental_voodoo_support=true --cross-file dosbox-cross
-            max_warnings: 2
+            max_warnings: 21
```
johnnovak, line 99: That's a bit unfortunate to have to increase the max warnings. Going from 2 to 21 is quite significant. Ideally, we want to go from 2 to 0 😄 Can't you find a better workaround so we can keep it at 2?

src/hardware/gameblaster.h
```diff
@@ -36,7 +36,7 @@
 
 #include "mame/emu.h"
 #include "mame/saa1099.h"
-#include "../libs/residfp/resample/TwoPassSincResampler.h"
+#include <residfp/resample/TwoPassSincResampler.h>
```
johnnovak, line 39: I'm unsure about the value and necessity of this change. Can you explain why this is an improvement in your view?

This blurs the distinction between system-provided and bundled libraries; I actually _prefer_ to make that distinction explicit.

Sure, if you move files around, you need to readjust the relative paths. But how often do we do that? Almost never, and even then it's a 2 min job to adjust the paths.

include/reelmagic.h
```diff
@@ -38,16 +38,15 @@ struct ReelMagic_VideoMixerMPEGProvider {
 
 void ReelMagic_RENDER_SetPal(uint8_t entry, uint8_t red, uint8_t green, uint8_t blue);
 
-void ReelMagic_RENDER_SetSize(const uint16_t width, const uint16_t height,
+void ReelMagic_RENDER_SetSize(const uint32_t width, const uint32_t height,
```
johnnovak, line 41: You're changing the types here. Would be better to separate the type changes out into their own commit?

include/reelmagic.h
```diff
@@ -38,16 +38,15 @@ struct ReelMagic_VideoMixerMPEGProvider {
...
                               const bool double_width, const bool double_height,
                               const Fraction& pixel_aspect_ratio,
-                              const uint8_t bits_per_pixel,
+                              const uint32_t bits_per_pixel,
```
johnnovak, line 44: Same here. Not sure why you needed to change it back from uint8_t. I have changed bpp to uint8_t in most places and I'm finishing the job in my big VGA related PR currently under review.

src/hardware/vga_dac.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include "inout.h"
 #include "mem.h"
-#include "reelmagic/vga_passthrough.h"
+#include "reelmagic.h"
```
johnnovak, line 25: Ok so this commit is doing a lot more than just removing a few unused declarations. You could either call it "Various cleanup changes" but better yet, you could split the unrelated changes, which should be easy to do.

src/hardware/vga_draw.cpp
```diff
@@ -970,10 +970,10 @@ static void VGA_DrawSingleLine(uint32_t /*blah*/)
 				write_unaligned_uint32_at(TempLine, i++, background_color);
 			}
 		}
-		RENDER_DrawLine(TempLine);
+		ReelMagic_RENDER_DrawLine(TempLine);
 	} else {
-		uint8_t * data=VGA_DrawLine( vga.draw.address, vga.draw.address_line );	
-		RENDER_DrawLine(data);
+		uint8_t * data=VGA_DrawLine( vga.draw.address, vga.draw.address_line );
+		ReelMagic_RENDER_DrawLine(data);
```
johnnovak, line 976: The ReelMagic related cleanups/renames most definitely warrant their own commit.

tests/rgb_tests.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include <gtest/gtest.h>
 
+#include <cmath>
```
johnnovak, line 25: Making changes to code should be in a separate commit after this. This commit should only do what the commit message says.

.github/workflows/linux.yml
johnnovak: I did read the commit description, but I'm wondering if there's a way to keep `max_warnings` at 2 while keeping your change 😄 

E.g. if you can use some pragma or "magic comment" to make these problematic compilers happy to ignore these 20 warnings, that's a much better workaround than globally increasing the number of allowed warnings for all compilers from 2 to 21 wholesale. That's a bit too heavy-handed approach, and I'd actually prefer not to do this change at all and keep the `max_warnings` level at 2.

src/hardware/gameblaster.h
johnnovak: Good points, I agree that not having to use relative paths is nicer. My problem was mainly about the angled bracket vs double-quote convention. Ultimately, I want to keep this convention:

- Angle brackets are used for standard C and C++ libraries and 3rd party dependencies not embedded into the project (e.g. `#include <SDL.h>`) 
- Double-quotes are used for everything else, basically files that live inside the repo. That includes 3rd party dependencies embedded into the repository.

If switching to double-quotes would still work, I'm all for this!

include/reelmagic.h
johnnovak: Sure, but separate them out into their own commit please? They don't belong the the commit they're under now.

include/reelmagic.h
johnnovak: Separate commit please. I'm gonna change these back to uint8_t when my PR gets merged, but whatever.

src/hardware/vga_dac.cpp
johnnovak: Well, the commit's title must reflect what the commit does 😄 So either change the commit message so it's a lot more broad (worse solution), or split the commit and make the commit messages of the resulting commits more specifi (better solution) 😄

.github/workflows/linux.yml
johnnovak: Ok kinda missed that, my bad. I thought it's for all Linux builds. If others are happy with this, I think it's acceptable compromise perhaps. @FeralChild64 @shermp ?

tests/rgb_tests.cpp
```diff
@@ -118,9 +118,9 @@ TEST(rgb, linear_to_srgb8_lut)
 	// This is good enough accuracy with a (16 * 1024) element LUT
 	constexpr auto NumIter = 500;
 	for (auto i = 0; i < NumIter; ++i) {
-		const float lin        = i * (1.0f / NumIter);
-		const uint8_t expected = round(linear_to_srgb(lin) * Rgb8Max);
+		const float lin     = i * (1.0f / NumIter);
+		const auto expected = static_cast<std::uint8_t>(
+		        std::round(linear_to_srgb(lin) * Rgb8Max));
```
johnnovak, line 123: I saw you've addressed most of my comments, thanks for that. Not sure if you were going to address this later or have just missed it. So updating `rgb_tests.cpp` should split out into its own commit too from the "Add several warnings to the GCC build" commit. Apologies if this is already on your TODO list 😅

johnnovak:
@MeAreJeenius is this ready for final review now then? Can you please also rebase to `main`?

johnnovak (approved):
> Should be good to go @johnnovak .

Cheers @MeAreJeenius, that was a massive amount of work! 🎉 Thanks for bearing with me and addressing all the comments!

PR #1627, opened 2022-04-25
Fix gamma-incorrect interpolation in the 'sharp' OpenGL shader

johnnovak:
# Problem description

The built-in 'sharp' OpenGL rescale filter is great, but it suffers from gamma-incorrect interpolation (it always uses hardware-accelerated linear interpolation on the sRGB encoded pixel values; this is mathemathically incorrect and yields wrong results). The adverse effects are most noticeable on high-contrast checkboard dither patterns, especially in EGA games (so half of the Sierra catalogue and early LucasArts games are affected), and on high-resolution text rendering (e.g. most Legend Entertainment adventure games).

Note that unless pixel-perfect mode is used, some interpolation will almost always be needed for screen modes that need aspect ratio correction, so for 320x200, 640x350 and 640x400, to name the most common ones. On 1080p displays the adverse interpolation artefacts can be quite visible, especially if the viewport is restricted to smaller than fullscreen, e.g. 960x720.

The following image illustrates the problem. Look for differences in text rendering quality and weird interference-like artefacts in checkerboard dither patterns (the red arrows show where to look). Make sure to check out the comparison images further down as well; they make the problem more apparent when switching back and forth between the good and bad images.

comparison

# Solution

The solution is to not use hardware interpolation, but gamma-decode (linearise) the colour values on the way in, perform the interpolation correctly in linear RGB space, then gamma-encode the results on the way out. This is pretty much what all good quality CRT shaders are doing.

There *might* be a way to get this done in hardware when specifying the textures as sRGB, but even the OpenGL specs are wishy-washy about this (it says the hardware _might_ perform the interpolation in linear space in that case... or not... The Vulkan specs seem to be much more strict and well-defined in this matter). Seems like it's another OpenGL feature that's highly driver and hardware dependent; no wonder virtually all high-quality CRT shaders bypass hardware interpolation completely and do their own linearisation & interpolation.

This makes the shader slower, but I doubt this would be a problem for any GPU released in the last 8-10 years. I have strong feelings about making the 'sharp' shader do the correct thing by default (as DOSBox-Staging generally tries to do), but perhaps we could also include the previous gamma-incorrect version as an option under the name 'sharp-fast'.

# Test images

_**You cannot properly see the problem by just looking at the images as they appear on this page because the browser scales them!**_

Please open the images pairwise in new tabs and switch between them, or download and view them with 1:1 pixel mapping in an image viewer; the differences are really easy to spot that way. 

[...]

PR #2665, opened 2023-07-06
Fix sRGB GLSL pragmas

johnnovak:
This is embarrassing... It doesn't even warrant the "regression" label because it couldn't have ever worked properly! :lolsob:

I guess I never noticed because I started using CRT shaders that do their own gamma decoding/encoding "manually" and thus interpolate in a  gamma-correct fashion.

Anyway, it should be good now, and it can be verified by uncommenting the debug logging. These two flags are enabled in the default `interpolation/sharp` shader and disabled in most CRT shaders, such as `crt/hyllian-updated`.

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2184,6 +2184,13 @@ dosurface:
 		const auto texformat = RENDER_UseSrgbTexture() && sdl.opengl.framebuffer_is_srgb_encoded
 		                             ? GL_SRGB8_ALPHA8
 		                             : GL_RGB8;
+
+#if 0
+		if (texformat == GL_SRGB8_ALPHA8) {
+			LOG_MSG("OPENGL: Using sRGB texture");
+		}
+#endif
+
 		glTexImage2D(GL_TEXTURE_2D,
 		             0,
 		             texformat,
@@ -2197,6 +2204,9 @@ dosurface:
 
 		if (sdl.opengl.framebuffer_is_srgb_encoded) {
 			glEnable(GL_FRAMEBUFFER_SRGB);
+#if 0
+			LOG_MSG("OPENGL: Using sRGB framebuffer");
+#endif
 		}
 
 		glClearColor (0.0f, 0.0f, 0.0f, 1.0f);
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2165,17 +2165,23 @@ dosurface:
 
 		int is_framebuffer_srgb_capable = 0;
 		if (SDL_GL_GetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE,
-		                        &is_framebuffer_srgb_capable))
+		                        &is_framebuffer_srgb_capable)) {
 			LOG_WARNING("OPENGL: Failed getting the framebuffer's sRGB status: %s",
 			            SDL_GetError());
+		}
 
-		sdl.opengl.framebuffer_is_srgb_encoded = RENDER_UseSRGBFramebuffer() && is_framebuffer_srgb_capable > 0;
+		sdl.opengl.framebuffer_is_srgb_encoded =
+		        RENDER_UseSrgbFramebuffer() &&
+		        (is_framebuffer_srgb_capable > 0);
 
-		if (RENDER_UseSRGBFramebuffer() && !sdl.opengl.framebuffer_is_srgb_encoded)
+		if (RENDER_UseSrgbFramebuffer() &&
+		    !sdl.opengl.framebuffer_is_srgb_encoded) {
 			LOG_WARNING("OPENGL: sRGB framebuffer not supported");
+		}
 
-		// Using GL_SRGB8_ALPHA8 because GL_SRGB8 doesn't work properly with Mesa drivers on certain integrated Intel GPUs
-		const auto texformat = RENDER_UseSRGBTexture() && sdl.opengl.framebuffer_is_srgb_encoded
+		// Using GL_SRGB8_ALPHA8 because GL_SRGB8 doesn't work properly
+		// with Mesa drivers on certain integrated Intel GPUs
+		const auto texformat = RENDER_UseSrgbTexture() && sdl.opengl.framebuffer_is_srgb_encoded
 		                             ? GL_SRGB8_ALPHA8
 		                             : GL_RGB8;
 		glTexImage2D(GL_TEXTURE_2D,
--- a/contrib/resources/glshaders/none.glsl
+++ b/contrib/resources/glshaders/none.glsl
@@ -1,9 +1,10 @@
 /*
  *  SPDX-License-Identifier: GPL-2.0-or-later
  *
+ *  Copyright (C) 2020-2023  The DOSBox Staging Team
  *  Copyright (C) 2020  The DOSBox Team
  *
- * Contributors:
+ *  Contributors:
  *   - 2020, jmarsh <jmarsh@vogons.org>: authored.
  *           https://svn.code.sf.net/p/dosbox/code-0/dosbox/trunk@4319
  *
--- a/contrib/resources/glshaders/none.glsl
+++ b/contrib/resources/glshaders/none.glsl
@@ -22,9 +22,6 @@
  *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-#pragma use_srgb_texture
-#pragma use_srgb_framebuffer
-
 varying vec2 v_texCoord;
 #if defined(VERTEX)
 uniform vec2 rubyTextureSize;
--- a/contrib/resources/glshaders/none.glsl
+++ b/contrib/resources/glshaders/none.glsl
@@ -24,17 +24,26 @@
  */
 
 varying vec2 v_texCoord;
+
 #if defined(VERTEX)
+
 uniform vec2 rubyTextureSize;
 uniform vec2 rubyInputSize;
 attribute vec4 a_position;
-void main() {
+
+void main()
+{
 	gl_Position = a_position;
-	v_texCoord = vec2(a_position.x+1.0,1.0-a_position.y)/2.0*rubyInputSize/rubyTextureSize;
+	v_texCoord  = vec2(a_position.x + 1.0, 1.0 - a_position.y) / 2.0 *
+	             rubyInputSize / rubyTextureSize;
 }
+
 #elif defined(FRAGMENT)
+
 uniform sampler2D rubyTexture;
-void main() {
+void main()
+{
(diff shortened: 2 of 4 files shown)


FeralChild64:
After the change the video output started looking strange with my configuration - font now looks thicker, and by a lot.

```init
[sdl]
windowresolution    = 1280x960
output              = opengl
texture_renderer    = auto

[dosbox]
vga_8dot_font               = true

[render]
aspect             = true
integer_scaling    = off
glshader           = 
```
Before:

old

After:

new

FeralChild64:
Windows 3.1 also works better on main (zoom to 100% size to see the difference):

old-win

than after the patch:

new-win

src/gui/sdlmain.cpp
```diff
@@ -2165,19 +2165,30 @@ Bitu GFX_SetSize(int width, int height, const Bitu flags, double scalex,
...
-		const auto texformat = RENDER_UseSRGBTexture() && sdl.opengl.framebuffer_is_srgb_encoded
+		// Using GL_SRGB8_ALPHA8 because GL_SRGB8 doesn't work properly
+		// with Mesa drivers on certain integrated Intel GPUs
+		const auto texformat = RENDER_UseSrgbTexture() && sdl.opengl.framebuffer_is_srgb_encoded
 		                             ? GL_SRGB8_ALPHA8
 		                             : GL_RGB8;
+
+		// if (texformat == GL_SRGB8_ALPHA8) {
+		// 	LOG_MSG("OPENGL: Using sRGB texture");
+		// }
```
weirddan455, line 2190: I've never looked at the render code in DosBox so I'm probably not the best person for review.  But I don't like commented out code like this.  I'd say either wrap this in an `#ifdef` for debug builds or just remove it.

src/gui/sdlmain.cpp
```diff
@@ -2191,6 +2202,7 @@ Bitu GFX_SetSize(int width, int height, const Bitu flags, double scalex,
 
 		if (sdl.opengl.framebuffer_is_srgb_encoded) {
 			glEnable(GL_FRAMEBUFFER_SRGB);
+		//	LOG_MSG("OPENGL: Using sRGB framebuffer");
```
weirddan455, line 2205: More commented out code here.

johnnovak:
> After the change the video output started looking strange with my configuration - font now looks thicker, and by a lot.

Thanks for reporting this, something clearly went wrong there. Can't reproduce it on my Mac, trying Windows now. It's a bit perpexling because you've explicitly disabled shaders, and this fix should only make a difference with shaders enabled (and not even all shaders, just certain ones that "opt-in" to the use of sRGB textures & framebuffer via the pragmas).

Investigating...

src/gui/sdlmain.cpp
johnnovak: They definitely won't be removed as they are super useful for focused debugging sessions. Writing them over and over again is a pain. Also , I don't like using "debug" log levels because then you get flooded with all sorts of debug messages you don't care about when enabling debug (or "trace") logging.

Another non-solution is logging categories -- too complicated to maintain, and at the end of the day, often I only want one of two specific debug messages and that's it...

What could be done, perhaps, is surrounding them with `#if 0` / `#endif`, so you can toggle them on by changing 0 to 1. But... dunno, I don't have problems with the current comment based solution. It does the job with minimal hassle.

@FeralChild64 @shermp @kklobe your views?

johnnovak: Btw, these were trivial one-liners. What I wrote is more about "heavy-duty" logging like this:

https://github.com/dosbox-staging/dosbox-staging/blob/main/src/capture/image/image_scaler.cpp#L139-L189

(Although, maybe a bit of a non-example as I wrapped this in a function, so I can easily only toggle calling this function... 😄)

johnnovak:
Good news @FeralChild64, could reproduce on my Nvidia card on Windows!

src/gui/sdlmain.cpp
weirddan455: Just an idea that I've seen SDL code do is wrap these in (for example) `#ifdef DEBUG_RENDERER `

Then when you're working on this specific part of the code, you can just type `#define DEBUG_RENDERER` at the top of the file (or use the build system to pass the flag in) and you get everything turned on.

But if you just want to stick to commented out code, I won't make too big of a fuss over it.  I just think it looks kind of ugly.

johnnovak: I'm personally happy to switch to `#if 0` highlighting; like I said, I want to be able to toggle comments on an individual basis. Toggling them wholesale with `DEBUG_RENDERER` won't cut it as I explained. Too noisy, they most often need to be toggled individually in most debugging tasks.

So for example this

image

Would become this

image

So unless objections from others, I'm happy to start using this style going forward. I was concerned that I won't see these blocks highlighted as comments, but the Vim C++ syntax plugin does support `#if 0`/`#if 1` highlighting out of the box, so that's great 😄

johnnovak: And as I said, for multi-line debug log statements changing `0` to `1` and back is actually a lot easier than toggling both ends of the comment block, or per line (if you editor doesn't support commenting/uncommenting whole blocks with one shortcut), so it was a good suggestion @weirddan455 👍🏻

johnnovak:
Okay, I managed to de-confuse myself and now I understand what's going on. This is happening on all platforms, I was confused when I said this cannot be reproduced on macOS...

1. The "no shader" path with `output = opengl` and `glshader = ` (explicitly set to empty string) actually uses the `none.glsl` shader. We _always_ use a shader with OpenGL, no matter what (we have to; this is modern OpenGL); and this is the so-called "no shader" path.

2. When I introduced gamma-correct interpolation last year, I've updated all shaders with the `use_*` pragmas to ensure gamma-correct bilinear interpolation. It's just that feature never worked 😩 so you never saw the gamma-correct results with the `none` shader that you dislike, so you couldn't complain about it back then 😆 

3. My view: both gamma-correct and gamma-incorrect bilinear interpolation look like shit—but it's a slightly different shit😎 You got used to _one_ particular brand of crappiness, but in the "extreme edge cases" of white-text-on-black-background, or black-text-on-white-background, the letters look either thicker or thinner, depending on whether correct or incorrect interpolation is used. It's easier to understand this by looking at the zoom-in image example below. The TL;DR is, you can't "win" with bilinear interpolation; either black-on-white or white-on-black will look slightly off, depending on the interpolation method. Solution: use the default "sharp" shader, or better yet, use a nice CRT shader.

(As a side-note, this kinda explains why you were insistent on saying that forcing 8-dot VGA fonts looked more authentic to you. Bright grey or white characters on a black background definitely look _thinner_ than they should with gamma-incorrect bilinear interpolation you have been using for a long time, presumably, no doubt about that. So by forcing 8-dot VGA fonts, you make the vertical lines a bit thicker compared to the more standard 9-dot VGA fonts, so you're effectively adding a "fudge factor" to "bring back" the correct-appearing width of the vertical lines. Or you indeed had some really early VGA card with 8-dot fonts, that's the other explanation.)

# Zoomed-in examples

## Default `interpolation/sharp` shader (gamma-correct)

This is the "idealised" look when we treat as little sharp rectangles.

image

## `none` shader (gamma-correct bilinear interpolation)

Mathematically correct; white-on-black looks a bit thick, black-on-white looks a bit thin.

image

## `none` shader (gamma-incorrect bilinear interpolation)

Mathematically incorrect; basically some random nonsensical fudging of the numbers because that's what gamma-incorrect interpolation is (sadly this has been the default for GPUs for a long time because it used to be computationally less expensive before hardware-assisted sRGB handling via lookups tables became standard). White-on-black looks a bit thing, black-on-white looks quite thick.

image

## CRT shader

A custom-tweaked version of `crt/hyllian-updated` I'm working on, just for reference how a good CRT shader fixes these problems. That ghastly bilinear interpolation blur is gone and the letters have optimal width in both edge cases, and the results are very similar to what you get on an actual CRT (I know because I keep comparing the results of my tweaked shader to my real CRTs).

image

# Solution

Well, not quite sure... Neither gamma-correct and gamma-incorrect straight bilinear interpolation look good at all, they're both crap in their unique and different ways 😎 Maybe we can just revert to the old "incorrect" way as that would make @FeralChild64 happy and all those people who for some reason like that look...

However, gamma-correctness is absolutely the one true and correct way for the "sharp" shader, as explained in my original ticket with numerous examples, that's non-negotiable. See that ticket for details: https://github.com/dosbox-staging/dosbox-staging/pull/1627

The new image capture feature also uses the same gamma-correct upscaling method as the "sharp" shader I implemented in software.

Gamma correctness is one of my pet peeves, and it's no exaggeration to say I'm probably a bit of an expert on the topic. The article I wrote about gamma was quite popular at HackerNews, so if you're even just a bit interested in computer graphics, then you might wanna to check it out. And if you're doing _any_ sort of image processing and you're just a bit unsure about gamma-correctness, I would say it's a _mandatory_ read.

https://news.ycombinator.com/item?id=12552094
https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/

---

# Full image examples

Just for reference; open them in multiple browser tabs and switch between them.

## Start screen

### `interpolation/sharp` (gamma-correct)

image

### `none` (gamma-correct, current PR behaviour)

image

### `none` (random gamma-incorrect result by fluke, `main` behaviour)

image


## Spellcasting 201

### `interpolation/sharp` (gamma-correct)

image

### `none` (gamma-correct, current PR behaviour)

image

### `none` (random gamma-incorrect result by fluke, `main` behaviour)

image

johnnovak:
Can you re-test @FeralChild64 that we're back to the previous behaviour? Please make sure you're using the updated version of `none.glsl` with the `use_srgb_*` pragmas removed.

FeralChild64:
Works nicely now, great!

FeralChild64 (approved):

GranMinigun:
Nitpick: commented-out pragmas are still parsed. And I would've preferred to leave default bilinear interpolation in the correct colour space for rare cases when I may use it, but oh well. Maybe I'll just port a couple more cubic shaders instead.

johnnovak:
> Nitpick: commented-out pragmas are still parsed.

Ah, ok. I'll try to shoehorn comment handling into that the regex. Did I say I hate regexes?

> And I would've preferred to leave default bilinear interpolation in the correct colour space for rare cases when I may use it, but oh well.

I may have a _slight_ preference toward the "correct" bilinear interpolation by default myself too, but I don't care much... 
As I see it, it's there mainly for people used to how bilineear with gamma-incorrect interpolation looks, otherwise everybody should use proper CRT shaders or even the default sharp shader.

GranMinigun:
Eh, can always drop a custom shader anyway. No biggie.

PR #2670, opened 2023-07-09
Store envelope channel_name as std::string

weirddan455:
Found a bug in this envelope class while working on cdrom stuff (almost done btw).  It doesn't do a deep copy of its name string so if it gets created with anything besides a string literal, you get random garbage printed out to the debug log.

Pretty simple change to just std::string.  It's only used for printing debug text.

### DIFF
--- a/src/hardware/envelope.cpp
+++ b/src/hardware/envelope.cpp
@@ -59,12 +59,14 @@ void Envelope::Update(const int frame_rate, const int peak_amplitude,
 	assert(expansion_phase_frames);
 	edge_increment = static_cast<float>(ceil_sdivide(peak_amplitude, expansion_phase_frames));
 
-//	DEBUG_LOG_MSG("ENVELOPE: %s grows by %-3f to %-5f across %-3u frames (%u ms)",
-//	              channel_name,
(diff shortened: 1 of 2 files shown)


src/hardware/envelope.cpp
```diff
@@ -60,7 +60,7 @@ void Envelope::Update(const int frame_rate, const int peak_amplitude,
 	edge_increment = static_cast<float>(ceil_sdivide(peak_amplitude, expansion_phase_frames));
 
 //	DEBUG_LOG_MSG("ENVELOPE: %s grows by %-3f to %-5f across %-3u frames (%u ms)",
-//	              channel_name,
+//	              channel_name.c_str(),
```
johnnovak, line 63: Feel free to turn this comment into and `#if 0` / `#endif` pair. I started using it and I actually like it better now.

johnnovak (approved):
Nice catch @weirddan455, thanks for fixing!

src/hardware/envelope.cpp
weirddan455: Fixed

PR #2673, opened 2023-07-10
Omit blank lines from command history

johnnovak:
Should be self-explanatory. Type `dir` in the current version, then press Enter a couple of times, then execute say `mixer`. You can't just press up-arrow twice to get back to the `dir` command in the history; you need to go through _every single blank line_ as well...

That's quite bad; none of the normal shells work like that!

### DIFF
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -445,14 +445,19 @@ void DOS_Shell::SyntaxError()
 
 static std::string get_shell_history_path()
 {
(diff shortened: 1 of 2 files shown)


weirddan455 (approved):
This is a good change.  I wrote some checks to avoid reading and writing blank lines to the history file but that didn't affect the in memory history for the current session.

FeralChild64 (approved):

Issue #1720, opened 2022-06-12
Add samples as floats to the mixer where possible


johnnovak:
Now that the mixer buffer uses floats and a float variant of `AddSamples` has been introduced, quite a few audio device emulations that use floats internally could be simplified to add samples as floats to the mix buffer, thus avoiding an unnecessary and potentially costly "float -> integer -> float" roundtrip. In certain cases this could also help with potential clipping issues (e.g. in the case of AdLib Gold).

The simplification could be made in these cases:

- Decoding compressed CD audio tracks in Vorbis (further details: https://github.com/dosbox-staging/dosbox-staging/pull/1715#discussion_r895035045)
- FluidSynth (needs SoftLimiter float support for a full float path)
- GUS (needs SoftLimiter float support for a full float path)
- GameBlaster: in mame/saa1099, the output samples are divided by 6 and cast to 16bit, then passed to gameblaster.cpp where they're summed into a float accumulator (stereo frames from both saa1099 chips), which is then passed into the SoftLimiter.
- mt32emu: we configure it with internal float rendering (using MT32Emu::RendererType_FLOAT), and call service->renderFloat() to get float samples out, which are then passed into the SoftLimiter.

@kcgen, are you perhaps aware of any others?

kcgen:
GamrBlaster: in mame/saa1099, the output samples are divided by 6 and cast to 16bit, then passed to the gameblaster where they're summed into a float accumulator (stereo frames from both saa1099 chips), which is then soft-limited.

kcgen:
mt32emu: we use it's `service->renderFloat()` API to get floats out, which we SoftLimit.

johnnovak:
Cheers, I've added these to the top description.

johnnovak:
I've already started working on this, @kcgen, as part of removing the soft-limiter and adding the master compressor.

johnnovak:
Okay, this is halfway done. Outstanding items:

```
src/hardware/sblaster.cpp:              sb.chan->AddSamples_m8(frames, maybe_silence(samples, MixTemp));
src/hardware/sblaster.cpp:              sb.chan->AddSamples_m8(frames,maybe_silence(samples, MixTemp));
src/hardware/sblaster.cpp:              sb.chan->AddSamples_m8(frames, maybe_silence(samples, MixTemp));
src/hardware/sblaster.cpp:                                      sb.chan->AddSamples_s8s(
[...]

PR #2675, opened 2023-07-11
Implement CD Audio using Linux ioctls

### LINKED ISSUE #2664: Document how to mount a physical CD-ROM

_No response_

Linux x86_64

_No response_

Because I can't figure it out.  I think other users might have problems too.  `mount -cd` lists my cd drive as `/dev/cdrom` which is good.  But then every other command I type either says `/dev/cdrom is not a directory` or it just prints the help page.

I even had a quick look at the code in `program_mount.cpp` but it's a bit hard to read what arguments are what.

And no, I'm not trying to mount the CD first.  That would work for data access but not for CD audio I believe (it would just be a normal directory mount).

An example of how to mount a physical cdrom in the `mount /?` page.

_No response_

mount cd

- [X] Yes, I agree.

weirddan455:
Fixes #2664

I've mostly de-coupled the Linux ioctl interface from SDL.  The only thing it still uses SDL for is getting device names (the UI commands use SDL device numbers so they need to match up).  This removed a kind of messy inheritance thing.

Most of the CD interface functions DOSBox uses mapped pretty cleanly to the Linux ioctls.  The hardest part was getting the mixer stuff right.  I've tested and it sounds correct to me now though.

I also tested that this works with CDEmu (https://cdemu.sourceforge.io/).  That creates a virtual CD drive and can maybe be useful if someone has an image that DOSBox won't mount.

Tested with Secret of Monkey Island (image through CDEmu) and Quake (physical CD).

### DIFF
--- a/src/dos/cdrom.h
+++ b/src/dos/cdrom.h
@@ -127,47 +127,42 @@ public:
 	CDROM_Interface_SDL();
 	CDROM_Interface_SDL(const CDROM_Interface_SDL&);
 	CDROM_Interface_SDL& operator = (const CDROM_Interface_SDL&);
-	virtual ~CDROM_Interface_SDL();
-	virtual bool SetDevice(const char *path, int cd_number);
-	virtual bool GetUPC(unsigned char &attr, char *upc)
+	~CDROM_Interface_SDL();
+	bool SetDevice(const char* path, int cd_number);
+	bool GetUPC(unsigned char& attr, char* upc)
 	{
 		attr = '\0';
 		strcpy(upc, "UPC");
 		return true;
 	}
-	virtual bool GetAudioTracks(uint8_t &stTrack, uint8_t &end, TMSF &leadOut);
-	virtual bool GetAudioTrackInfo(uint8_t track, TMSF &start, unsigned char &attr);
-	virtual bool GetAudioSub(unsigned char &attr,
-	                         unsigned char &track,
-	                         unsigned char &index,
-	                         TMSF &relPos,
-	                         TMSF &absPos);
-	virtual bool GetAudioStatus(bool &playing, bool &pause);
-	virtual bool GetMediaTrayStatus(bool &mediaPresent,
-	                                bool &mediaChanged,
-	                                bool &trayOpen);
-	virtual bool PlayAudioSector(const uint32_t start, uint32_t len);
-	virtual bool PauseAudio(bool resume);
-	virtual bool StopAudio();
-	virtual void ChannelControl([[maybe_unused]] TCtrl ctrl)
+	bool GetAudioTracks(uint8_t& stTrack, uint8_t& end, TMSF& leadOut);
+	bool GetAudioTrackInfo(uint8_t track, TMSF& start, unsigned char& attr);
+	bool GetAudioSub(unsigned char& attr, unsigned char& track,
+	                 unsigned char& index, TMSF& relPos, TMSF& absPos);
+	bool GetAudioStatus(bool& playing, bool& pause);
+	bool GetMediaTrayStatus(bool& mediaPresent, bool& mediaChanged,
+	                        bool& trayOpen);
+	bool PlayAudioSector(const uint32_t start, uint32_t len);
+	bool PauseAudio(bool resume);
+	bool StopAudio();
+	void ChannelControl([[maybe_unused]] TCtrl ctrl)
 	{
 		return;
 	}
-	virtual bool ReadSectors([[maybe_unused]] PhysPt buffer,
-	                         [[maybe_unused]] const bool raw,
-	                         [[maybe_unused]] const uint32_t sector,
-	                         [[maybe_unused]] const uint16_t num)
+	bool ReadSectors([[maybe_unused]] PhysPt buffer,
+	                 [[maybe_unused]] const bool raw,
+	                 [[maybe_unused]] const uint32_t sector,
+	                 [[maybe_unused]] const uint16_t num)
 	{
 		return false;
 	}
-	virtual bool ReadSectorsHost([[maybe_unused]] void *buffer,
-	                             [[maybe_unused]] bool raw,
-	                             [[maybe_unused]] unsigned long sector,
-	                             [[maybe_unused]] unsigned long num)
+	bool ReadSectorsHost([[maybe_unused]] void* buffer, [[maybe_unused]] bool raw,
+	                     [[maybe_unused]] unsigned long sector,
+	                     [[maybe_unused]] unsigned long num)
 	{
 		return true;
 	}
-	virtual bool LoadUnloadMedia(bool unload);
+	bool LoadUnloadMedia(bool unload);
 
 private:
 	bool Open();
@@ -354,18 +349,45 @@ private:
 };
 
 #if defined (LINUX)
-class CDROM_Interface_Ioctl : public CDROM_Interface_SDL
-{
+class CDROM_Interface_Ioctl : public CDROM_Interface {
 public:
-	CDROM_Interface_Ioctl		();
+	~CDROM_Interface_Ioctl();
 
-	bool	SetDevice		(const char* path, const int cd_number);
-	bool	GetUPC			(unsigned char& attr, char* upc);
-	bool	ReadSectors		(PhysPt buffer, const bool raw, const uint32_t sector, const uint16_t num);
-	bool	ReadSectorsHost	(void* buffer, bool raw, unsigned long sector, unsigned long num);
+	bool SetDevice(const char* path, const int cd_number);
+	bool GetUPC(unsigned char& attr, char* upc);
+	bool GetAudioTracks(uint8_t& stTrack, uint8_t& end, TMSF& leadOut);
+	bool GetAudioTrackInfo(uint8_t track, TMSF& start, unsigned char& attr);
+	bool GetAudioSub(unsigned char& attr, unsigned char& track,
+	                 unsigned char& index, TMSF& relPos, TMSF& absPos);
+	bool GetAudioStatus(bool& playing, bool& pause);
+	bool GetMediaTrayStatus(bool& mediaPresent, bool& mediaChanged,
+	                        bool& trayOpen);
+	bool ReadSectors(PhysPt buffer, const bool raw, const uint32_t sector,
+	                 const uint16_t num);
+	bool ReadSectorsHost(void* buffer, bool raw, unsigned long sector,
+	                     unsigned long num);
+	bool PlayAudioSector(const uint32_t start, uint32_t len);
+	bool PauseAudio(bool resume);
+	bool StopAudio();
+	void ChannelControl(TCtrl ctrl);
+	bool LoadUnloadMedia(bool unload);
 
 private:
-	char	device_name[512];
+	void CdAudioCallback(const uint16_t requested_frames);
+	void InitAudio(const int device_number);
+	bool IsOpen() const;
+	bool Open(const char* device_name);
+
+	int cdrom_fd                          = -1;
+	mixer_channel_t mixer_channel         = nullptr;
+	std::vector<int16_t> input_buffer     = {};
+	std::vector<AudioFrame> output_buffer = {};
+	size_t input_buffer_position          = 0;
+	size_t input_buffer_samples           = 0;
+	int current_sector                    = 0;
+	int sectors_remaining                 = 0;
+	bool is_playing                       = false;
+	bool is_paused                        = false;
 };
 
 #endif /* LINUX */
--- a/src/dos/cdrom_ioctl_linux.cpp
+++ b/src/dos/cdrom_ioctl_linux.cpp
@@ -23,27 +23,40 @@
 
 #if defined(LINUX)
 #include <fcntl.h>
-#include <unistd.h>
+#include <limits.h>
 #include <linux/cdrom.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <unistd.h>
 
-CDROM_Interface_Ioctl::CDROM_Interface_Ioctl() : CDROM_Interface_SDL()
+constexpr auto MixerChannelNamePrefix = "CDAUDIO_";
+// ioctl cannot be read more than 75 redbook frames at a time (one second of audio)
+constexpr int InputBufferMaxRedbookFrames = 25;
+constexpr size_t PcmSamplesPerRedbookFrame = BYTES_PER_RAW_REDBOOK_FRAME / REDBOOK_BPS;
+constexpr size_t InputBufferMaxSamples = InputBufferMaxRedbookFrames *
+                                         PcmSamplesPerRedbookFrame;
+
+CDROM_Interface_Ioctl::~CDROM_Interface_Ioctl()
 {
-	*device_name = '\0';
+	if (mixer_channel) {
+		MIXER_DeregisterChannel(mixer_channel);
+	}
+}
+
+bool CDROM_Interface_Ioctl::IsOpen() const
+{
+	return cdrom_fd != -1;
 }
 
 bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
 {
-	int cdrom_fd = open(device_name, O_RDONLY | O_NONBLOCK);
-	if (cdrom_fd == -1)
+	if (!IsOpen()) {
 		return false;
+	}
 
 	struct cdrom_mcn cdrom_mcn;
 	int ret = ioctl(cdrom_fd, CDROM_GET_MCN, &cdrom_mcn);
-
-	close(cdrom_fd);
 
 	if (ret > 0) {
 		attr = 0;
@@ -53,14 +66,192 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
 	return (ret > 0);
 }
 
-bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
-                                        const bool raw,
-                                        const uint32_t sector,
-                                        const uint16_t num)
+bool CDROM_Interface_Ioctl::GetAudioTracks(uint8_t& stTrack, uint8_t& end, TMSF& leadOut)
 {
-	int cdrom_fd = open(device_name, O_RDONLY | O_NONBLOCK);
-	if (cdrom_fd == -1)
+	if (!IsOpen()) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioTracks: cdrom_fd not open");
+#endif
 		return false;
+	}
+
+	cdrom_tochdr toc = {};
+	if (ioctl(cdrom_fd, CDROMREADTOCHDR, &toc) != 0) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioTracks: ioctl CDROMREADTOCHDR failed");
+#endif
+		return false;
+	}
+
+	cdrom_tocentry entry = {};
+	entry.cdte_track     = CDROM_LEADOUT;
+	entry.cdte_format    = CDROM_MSF;
+	if (ioctl(cdrom_fd, CDROMREADTOCENTRY, &entry) != 0) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioTracks: ioctl CDROMREADTOCENTRY failed");
+#endif
+		return false;
+	}
+
+	stTrack     = toc.cdth_trk0;
+	end         = toc.cdth_trk1;
+	leadOut.min = entry.cdte_addr.msf.minute;
+	leadOut.sec = entry.cdte_addr.msf.second;
+	leadOut.fr  = entry.cdte_addr.msf.frame;
+
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: GetAudioTracks => start track is %2d, last playable track is %2d, "
+	         "and lead-out MSF is %02d:%02d:%02d",
+	         stTrack,
+	         end,
+	         leadOut.min,
+	         leadOut.sec,
+	         leadOut.fr);
+#endif
+
+	return true;
+}
+
+bool CDROM_Interface_Ioctl::GetAudioTrackInfo(uint8_t track, TMSF& start,
+                                              unsigned char& attr)
+{
+	if (!IsOpen()) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioTrackInfo: cdrom_fd not open");
+#endif
+		return false;
+	}
+
+	cdrom_tocentry entry = {};
+	entry.cdte_track     = track;
+	entry.cdte_format    = CDROM_MSF;
+	if (ioctl(cdrom_fd, CDROMREADTOCENTRY, &entry) != 0) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioTrackInfo: ioctl CDROMREADTOCENTRY failed");
+#endif
+		return false;
+	}
+
+	start.min = entry.cdte_addr.msf.minute;
+	start.sec = entry.cdte_addr.msf.second;
+	start.fr  = entry.cdte_addr.msf.frame;
+	attr      = (entry.cdte_ctrl << 4) | entry.cdte_adr;
+
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: GetAudioTrackInfo for track %u => "
+	         "MSF %02d:%02d:%02d, which is sector %d",
+	         track,
+	         start.min,
+	         start.sec,
+	         start.fr,
+	         msf_to_frames(start));
+#endif
+
+	return true;
+}
+
+bool CDROM_Interface_Ioctl::GetAudioSub(unsigned char& attr, unsigned char& track,
+                                        unsigned char& index, TMSF& relPos,
+                                        TMSF& absPos)
+{
+	if (!IsOpen()) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioSub: cdrom_fd not open");
+#endif
+		return false;
+	}
+
+	cdrom_subchnl sub = {};
+	sub.cdsc_format   = CDROM_MSF;
+	if (ioctl(cdrom_fd, CDROMSUBCHNL, &sub) != 0) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetAudioSub: ioctl CDROMSUBCHNL failed");
+#endif
+		return false;
+	}
+
+	attr       = (sub.cdsc_ctrl << 4) | sub.cdsc_adr;
+	track      = sub.cdsc_trk;
+	index      = sub.cdsc_ind;
+	relPos.min = sub.cdsc_reladdr.msf.minute;
+	relPos.sec = sub.cdsc_reladdr.msf.second;
+	relPos.fr  = sub.cdsc_reladdr.msf.frame;
+	absPos.min = sub.cdsc_absaddr.msf.minute;
+	absPos.sec = sub.cdsc_absaddr.msf.second;
+	absPos.fr  = sub.cdsc_absaddr.msf.frame;
+
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: GetAudioSub => position at %02d:%02d:%02d (on sector %u) "
+	         "within track %u at %02d:%02d:%02d (at its sector %u)",
+	         absPos.min,
+	         absPos.sec,
+	         absPos.fr,
+	         msf_to_frames(absPos),
+	         track,
+	         relPos.min,
+	         relPos.sec,
+	         relPos.fr,
+	         msf_to_frames(relPos));
+#endif
+
+	return true;
+}
+
+bool CDROM_Interface_Ioctl::GetAudioStatus(bool& playing, bool& pause)
+{
+	playing = is_playing;
+	pause   = is_paused;
+
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: GetAudioStatus => %s and %s",
+	         playing ? "is playing" : "stopped",
+	         pause ? "paused" : "not paused");
+#endif
+
+	return true;
+}
+
+// Called from CMscdex::GetDeviceStatus - dos_mscdex.cpp line 821
+// Function doesn't check return value or initialize variables so just set some
+// defaults and never fail.
+bool CDROM_Interface_Ioctl::GetMediaTrayStatus(bool& mediaPresent,
+                                               bool& mediaChanged, bool& trayOpen)
+{
+	mediaPresent = false;
+	mediaChanged = false;
+	trayOpen     = false;
+
+	if (!IsOpen()) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetMediaTrayStatus: cdrom_fd not open");
+#endif
+		return true;
+	}
+
+	switch (ioctl(cdrom_fd, CDROM_DRIVE_STATUS, CDSL_CURRENT)) {
+	case CDS_TRAY_OPEN: trayOpen = true; break;
+	case CDS_DISC_OK: mediaPresent = true; break;
+	}
+
+	auto ret     = ioctl(cdrom_fd, CDROM_MEDIA_CHANGED, CDSL_CURRENT);
+	mediaChanged = (ret > 0) && (ret & 1);
+
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: GetMediaTrayStatus => media is %s, %s, and the tray is %s",
+	         mediaPresent ? "present" : "not present",
+	         mediaChanged ? "was changed" : "hasn't been changed",
+	         trayOpen ? "open" : "closed");
+#endif
+
+	return true;
+}
+
+bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer, const bool raw,
+                                        const uint32_t sector, const uint16_t num)
+{
+	if (!IsOpen()) {
+		return false;
+	}
 
 	const auto buflen = raw ? num * (unsigned int)CD_FRAMESIZE_RAW
 	                        : num * (unsigned int)CD_FRAMESIZE;
@@ -83,26 +274,53 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
 		if ((Bitu)ret != buflen)
 			ret = -1;
 	}
-	close(cdrom_fd);
 
 	MEM_BlockWrite(buffer, buf.data(), buflen);
 
 	return (ret > 0);
 }
 
-bool CDROM_Interface_Ioctl::SetDevice(const char *path, const int cd_number)
+// Opens the given device name, replacing any currently opened device
+bool CDROM_Interface_Ioctl::Open(const char* device_name)
 {
-	bool success = CDROM_Interface_SDL::SetDevice(path, cd_number);
+	int fd = open(device_name, O_RDONLY | O_NONBLOCK);
+	if (fd == -1) {
+		return false;
+	}
+	if (IsOpen()) {
+		close(cdrom_fd);
+	}
+	cdrom_fd = fd;
+	return true;
+}
 
-	if (success) {
-		const char *tmp = SDL_CDName(cd_number);
-		if (tmp)
-			safe_strncpy(device_name, tmp, 512);
-		else
-			success = false;
+bool CDROM_Interface_Ioctl::SetDevice(const char* path, const int cd_number)
+{
+	assert(path != nullptr);
+	assert(*path);
+
+	int num = SDL_CDNumDrives();
+	if ((cd_number >= 0) && (cd_number < num)) {
+		auto cd_name = SDL_CDName(cd_number);
+		if (cd_name && Open(cd_name)) {
+			InitAudio(cd_number);
+			return true;
+		}
 	}
 
-	return success;
+	for (auto i = 0; i < num; ++i) {
+		auto cd_name = SDL_CDName(i);
+		if (cd_name && strcmp(cd_name, path) == 0 && Open(cd_name)) {
+			InitAudio(i);
+			return true;
+		}
+	}
+
+	LOG_WARNING("CDROM: SetDevice failed to find device for path '%s' and device number %d",
+	            path,
+	            cd_number);
+
+	return false;
 }
 
 bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
@@ -111,6 +329,162 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
                                             [[maybe_unused]] unsigned long num)
 {
 	return false; /*TODO*/
+}
+
+void CDROM_Interface_Ioctl::InitAudio(const int device_number)
+{
+	if (mixer_channel) {
+		return;
+	}
+
+	std::string name = std::string(MixerChannelNamePrefix) +
+	                   std::to_string(device_number);
+
+	// Input buffer is used as a C-style buffer passed to an ioctl.
+	// Its size must always be 1 second of data.
+	// Stored in int16_t samples.
+	input_buffer.resize(InputBufferMaxSamples);
+
+	auto callback = std::bind(&CDROM_Interface_Ioctl::CdAudioCallback,
+	                          this,
+	                          std::placeholders::_1);
+	mixer_channel = MIXER_AddChannel(callback,
+	                                 REDBOOK_PCM_FRAMES_PER_SECOND,
+	                                 name.c_str(),
+	                                 {ChannelFeature::Stereo,
+	                                  ChannelFeature::DigitalAudio});
+}
+
+void CDROM_Interface_Ioctl::CdAudioCallback(const uint16_t requested_frames)
+{
+	if (input_buffer_position == 0) {
+		if (sectors_remaining < 1) {
+			StopAudio();
+			return;
+		}
+		cdrom_read_audio cd = {};
+		cd.addr.lba         = current_sector;
+		cd.addr_format      = CDROM_LBA;
+		cd.buf = reinterpret_cast<uint8_t*>(input_buffer.data());
+
+		// Maximum this ioctl can read in a single call is 1 second (75
+		// redbook frames)
+		cd.nframes = std::min(sectors_remaining, InputBufferMaxRedbookFrames);
+		input_buffer_samples = cd.nframes * PcmSamplesPerRedbookFrame;
+		current_sector += cd.nframes;
+		sectors_remaining -= cd.nframes;
+
+		if (ioctl(cdrom_fd, CDROMREADAUDIO, &cd) != 0) {
+			// Write out silence if something goes wrong
+			std::fill(input_buffer.begin(), input_buffer.end(), 0);
+
+#ifdef DEBUG_IOCTL
+			LOG_WARNING("CDROM_IOCTL: CdAudioCallback: CDROMREADAUDIO ioctl failed");
+#endif
+		}
+	}
+
+	size_t end = input_buffer_position + (requested_frames * REDBOOK_CHANNELS);
+	end = std::min(end, input_buffer_samples);
+
+	output_buffer.clear();
+	while (input_buffer_position < end) {
+		const int16_t left = host_to_le16(
+		        input_buffer[input_buffer_position++]);
+		const int16_t right = host_to_le16(
+		        input_buffer[input_buffer_position++]);
+		output_buffer.emplace_back(left, right);
+	}
+
+	mixer_channel->AddSamples_sfloat(output_buffer.size(), &output_buffer[0][0]);
+
+	if (input_buffer_position >= input_buffer_samples) {
+		input_buffer_position = 0;
+	}
+}
+
+bool CDROM_Interface_Ioctl::PlayAudioSector(const uint32_t start, uint32_t len)
+{
+	if (!mixer_channel) {
+		return false;
+	}
+	input_buffer_position = 0;
+	current_sector        = start;
+	sectors_remaining     = len;
+	is_playing            = true;
+	is_paused             = false;
+	mixer_channel->Enable(true);
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: PlayAudioSector: start: %u len: %u", start, len);
+#endif
+	return true;
+}
+
+bool CDROM_Interface_Ioctl::PauseAudio(bool resume)
+{
+	if (mixer_channel) {
+		mixer_channel->Enable(resume);
+	}
+	is_paused = !resume;
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM: PauseAudio => audio is now %s",
+	         resume ? "unpaused" : "paused");
+#endif
+	return true;
+}
+
+bool CDROM_Interface_Ioctl::StopAudio()
+{
+	if (mixer_channel) {
+		mixer_channel->Enable(false);
+	}
+	is_playing            = false;
+	is_paused             = false;
+	current_sector        = 0;
+	sectors_remaining     = 0;
+	input_buffer_position = 0;
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: StopAudio => stopped playback and halted the mixer");
+#endif
+	return true;
+}
+
+// Taken from CDROM_Interface_Image
+void CDROM_Interface_Ioctl::ChannelControl(TCtrl ctrl)
+{
+	if (!mixer_channel) {
+		return;
+	}
+
+	constexpr float MaxVolume = 255.0f;
+	// Adjust the volume of our mixer channel as defined by the application
+	mixer_channel->SetAppVolume(ctrl.vol[0] / MaxVolume, ctrl.vol[1] / MaxVolume);
+
+	// Map the audio channels in our mixer channel as defined by the
+	// application
+	const auto left_mapped  = static_cast<LINE_INDEX>(ctrl.out[0]);
+	const auto right_mapped = static_cast<LINE_INDEX>(ctrl.out[1]);
+	mixer_channel->ChangeChannelMap(left_mapped, right_mapped);
+#ifdef DEBUG_IOCTL
+	LOG_INFO("CDROM_IOCTL: ChannelControl => volumes %d/255 and %d/255, "
+	         "and left-right map %d, %d",
+	         ctrl.vol[0],
+	         ctrl.vol[1],
+	         ctrl.out[0],
+	         ctrl.out[1]);
+#endif
+}
+
+bool CDROM_Interface_Ioctl::LoadUnloadMedia(bool unload)
+{
+	if (!IsOpen()) {
+		return false;
+	}
+	if (unload) {
+		return ioctl(cdrom_fd, CDROMEJECT) == 0;
+	} else {
+		return ioctl(cdrom_fd, CDROMCLOSETRAY) == 0;
+	}
 }
 
 #endif
--- a/src/dos/program_mount.cpp
+++ b/src/dos/program_mount.cpp
@@ -462,20 +462,23 @@ void MOUNT::AddMessages() {
 	        "\n"
 	        "Notes:\n"
 	        "  - '-t overlay' redirects writes for mounted drive to another directory.\n"
+	        "  - '-usecd ID' gives direct access to a CD-ROM drive.\n"
+	        "    This is needed for CD audio (only supported on Windows and Linux).\n"
+	        "    Run 'mount -cd' to find out the list of valid IDs.\n"
 	        "  - Additional options are described in the manual (README file, chapter 4).\n"
 	        "\n"
 	        "Examples:\n"
 #if defined(WIN32)
 	        "  [color=green]mount[reset] [color=white]C[reset] [color=cyan]C:\\dosgames[reset]\n"
-	        "  [color=green]mount[reset] [color=white]D[reset] [color=cyan]D:\\[reset] -t cdrom\n"
+	        "  [color=green]mount[reset] [color=white]D[reset] [color=cyan]D:\\[reset] -t cdrom -usecd 0\n"
 	        "  [color=green]mount[reset] [color=white]C[reset] [color=cyan]my_savegame_files[reset] -t overlay\n"
 #elif defined(MACOSX)
 	        "  [color=green]mount[reset] [color=white]C[reset] [color=cyan]~/dosgames[reset]\n"
-	        "  [color=green]mount[reset] [color=white]D[reset] [color=cyan]\"/Volumes/Game CD\"[reset] -t cdrom\n"
+	        "  [color=green]mount[reset] [color=white]D[reset] [color=cyan]\"/Volumes/Game CD\"[reset] -t cdrom -usecd 0\n"
 	        "  [color=green]mount[reset] [color=white]C[reset] [color=cyan]my_savegame_files[reset] -t overlay\n"
 #else
 	        "  [color=green]mount[reset] [color=white]C[reset] [color=cyan]~/dosgames[reset]\n"
-	        "  [color=green]mount[reset] [color=white]D[reset] [color=cyan]\"/media/USERNAME/Game CD\"[reset] -t cdrom\n"
+	        "  [color=green]mount[reset] [color=white]D[reset] [color=cyan]\"/media/USERNAME/Game CD\"[reset] -t cdrom -usecd 0\n"
 	        "  [color=green]mount[reset] [color=white]C[reset] [color=cyan]my_savegame_files[reset] -t overlay\n"
 #endif
 	);


src/dos/cdrom.h
```diff
@@ -354,18 +349,42 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	bool LoadUnloadMedia(bool unload);
 
 private:
-	char	device_name[512];
+	void CDAudioCallback(uint16_t requested_frames);
+	void InitAudio();
+
+	int cdrom_fd                            = -1;
+	mixer_channel_t mixer_channel           = nullptr;
+	std::unique_ptr<int16_t[]> audio_buffer = nullptr;
```
johnnovak, line 381: It would be better to use unsigned unless we really need signed, so `uint16_t`.

src/dos/cdrom.h
```diff
@@ -354,18 +349,42 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	size_t audio_position                   = 0;
+	size_t frames_in_buffer                 = 0;
+	int current_sector                      = 0;
+	int sectors_remaining                   = 0;
```
johnnovak, line 385: `uint16_t` for both (or the smallest integer we can use).

src/dos/cdrom.h
```diff
@@ -354,18 +349,42 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	bool playing                            = false;
+	bool paused                             = false;
```
johnnovak, line 387: minor: is_playing and is_paused?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+	LOG_INFO("CDROM_IOCTL: GetAudioTracks => start track is %2d, last playable track is %2d, "
+	         "and lead-out MSF is %02d:%02d:%02d",
+	         stTrack,
+	         end,
+	         leadOut.min,
+	         leadOut.sec,
+	         leadOut.fr);
+#endif
+	return true;
+}
```
johnnovak, line 99: A few blank lines here and there would help readability; now this function is just one big block of text.

src/dos/cdrom.h
```diff
@@ -354,18 +349,42 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	bool StopAudio();
+	void ChannelControl(TCtrl ctrl);
+	bool LoadUnloadMedia(bool unload);
 
 private:
-	char	device_name[512];
+	void CDAudioCallback(uint16_t requested_frames);
+	void InitAudio();
+
+	int cdrom_fd                            = -1;
```
johnnovak, line 379: Can be unsigned?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+
+bool CDROM_Interface_Ioctl::GetAudioTrackInfo(uint8_t track, TMSF& start,
+                                              unsigned char& attr)
+{
+	if (cdrom_fd == -1) {
```
johnnovak, line 104: It would be better to have an `is_opened` flag instead of deriving that information from magic values in multiple places. Then this becomes `if (is_opened)`.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+	         absPos.fr,
+	         msf_to_frames(absPos),
+	         track,
+	         relPos.min,
+	         relPos.sec,
+	         relPos.fr,
+	         msf_to_frames(relPos));
+#endif
+	return true;
+}
```
johnnovak, line 176: Same general comment: please insert some blank lines to group things logically and break up the wall-of-text look. This is hard to parse visually. I'd insert blanks around the `if` and before the last debug log statement.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+// Called from CMscdex::GetDeviceStatus - dos_mscdex.cpp line 821
+// Function doesn't check return value or initialize variables so just set some
+// defaults and never fail.
+bool CDROM_Interface_Ioctl::GetMediaTrayStatus(bool& mediaPresent,
+                                               bool& mediaChanged, bool& trayOpen)
+{
+	mediaPresent = false;
+	mediaChanged = false;
+	trayOpen     = false;
+	if (cdrom_fd == -1) {
```
johnnovak, line 199: Blank above.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetMediaTrayStatus: cdrom_fd not open");
+#endif
+		return true;
+	}
+	switch (ioctl(cdrom_fd, CDROM_DRIVE_STATUS, CDSL_CURRENT)) {
+	case CDS_TRAY_OPEN: trayOpen = true; break;
+	case CDS_DISC_OK: mediaPresent = true; break;
+	}
+	int ret = ioctl(cdrom_fd, CDROM_MEDIA_CHANGED, CDSL_CURRENT);
```
johnnovak, line 209: Blank above.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+	mediaPresent = false;
+	mediaChanged = false;
+	trayOpen     = false;
+	if (cdrom_fd == -1) {
+#ifdef DEBUG_IOCTL
+		LOG_WARNING("CDROM_IOCTL: GetMediaTrayStatus: cdrom_fd not open");
+#endif
+		return true;
+	}
+	switch (ioctl(cdrom_fd, CDROM_DRIVE_STATUS, CDSL_CURRENT)) {
```
johnnovak, line 205: Blank above.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+	case CDS_TRAY_OPEN: trayOpen = true; break;
+	case CDS_DISC_OK: mediaPresent = true; break;
+	}
+	int ret = ioctl(cdrom_fd, CDROM_MEDIA_CHANGED, CDSL_CURRENT);
+	mediaChanged = (ret > 0) && (ret & 1);
+#ifdef DEBUG_IOCTL
```
johnnovak, line 211: Blank above.
(Stopping now... you get the picture 😄 )

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -53,12 +57,169 @@ bool CDROM_Interface_Ioctl::GetUPC(unsigned char &attr, char *upc)
...
+	LOG_INFO("CDROM_IOCTL: GetMediaTrayStatus => media is %s, %s, and the tray is %s",
+	         mediaPresent ? "present" : "not present",
+	         mediaChanged ? "was changed" : "hasn't been changed",
+	         trayOpen ? "open" : "closed");
+#endif
```
johnnovak, line 216: Very nice debug logging 💯

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -92,17 +252,35 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
+		if (cd_name && strcmp(cd_name, path) == 0) {
+			cdrom_fd = open(cd_name, O_RDONLY | O_NONBLOCK);
+			if (cdrom_fd != -1) {
+				return true;
+			}
+		}
 	}
 
-	return success;
+	return false;
```
johnnovak, line 283: Ok, so the vertical spacing/grouping is nicely done in this function 🎉  I'd expect the same in all the others.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
 	return false; /*TODO*/
 }
 
+void CDROM_Interface_Ioctl::InitAudio()
+{
+	// Find a a channel name that doesn't exist (channel names must be
+	// unique or they get overwritten)
+	mixer_channel_t exists;
+	int num                  = 0;
+	const std::string prefix = MixerChannelNamePrefix;
```
johnnovak, line 300: Why not just use `MixerChannelNamePrefix` directly below?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
 	return false; /*TODO*/
 }
 
+void CDROM_Interface_Ioctl::InitAudio()
+{
+	// Find a a channel name that doesn't exist (channel names must be
+	// unique or they get overwritten)
+	mixer_channel_t exists;
+	int num                  = 0;
```
johnnovak, line 299: `auto`

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -92,17 +252,35 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
+	}
 
-	if (success) {
-		const char *tmp = SDL_CDName(cd_number);
-		if (tmp)
-			safe_strncpy(device_name, tmp, 512);
-		else
-			success = false;
+	for (int i = 0; i < num; i++) {
+		const char* cd_name = SDL_CDName(i);
```
johnnovak, line 274: Use auto?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -92,17 +252,35 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
+	assert(path != nullptr);
+
+	if (cdrom_fd != -1) {
+		close(cdrom_fd);
+		cdrom_fd = -1;
+	}
+
+	int num = SDL_CDNumDrives();
+	if ((cd_number >= 0) && (cd_number < num)) {
+		const char* cd_name = SDL_CDName(cd_number);
```
johnnovak, line 264: Use auto?

src/dos/cdrom_ioctl_linux.cpp
johnnovak, line 209: Use auto?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -92,17 +252,35 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
 
 bool CDROM_Interface_Ioctl::SetDevice(const char *path, const int cd_number)
 {
-	bool success = CDROM_Interface_SDL::SetDevice(path, cd_number);
+	assert(path != nullptr);
+
+	if (cdrom_fd != -1) {
+		close(cdrom_fd);
+		cdrom_fd = -1;
```
johnnovak, line 259: Yeah, so please introduce that `is_opened` flag.

src/dos/cdrom.h
weirddan455: Well, the data is actually 16 bit signed integers because that's how CD audio is stored.  As of now, I'm not directly manipulating it so it could technically be anything (though it does need to be explicitly cast to `int16_t` before passing the pointer off to `AddSamples_s16`)

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+	int num                  = 0;
+	const std::string prefix = MixerChannelNamePrefix;
+	std::string full_name;
+	do {
+		full_name = prefix + std::to_string(num++);
+		exists    = MIXER_FindChannel(full_name.c_str());
+	} while (exists);
+
+	audio_buffer = std::make_unique<int16_t[]>(AudioBufferBytes / 2);
+	mixer_channel = MIXER_AddChannel(std::bind(&CDROM_Interface_Ioctl::CDAudioCallback,
```
johnnovak, line 308: Maybe break the `std::bind` stuff out into a temp var.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+		current_sector += cd.nframes;
+		sectors_remaining -= cd.nframes;
+		if (cdrom_fd == -1 || ioctl(cdrom_fd, CDROMREADAUDIO, &cd) != 0) {
+			// Write out silence if something goes wrong
+			memset(audio_buffer.get(), 0, AudioBufferBytes);
+#ifdef DEBUG_IOCTL
+			LOG_WARNING("CDROM_IOCTL: CDAudioCallback: CDROMREADAUDIO ioctl failed");
+#endif
+		}
+	}
```
johnnovak, line 343: Vertical spacing please...

src/dos/cdrom.h
johnnovak, line 381: It would be better to use `std::vector`.

src/dos/cdrom.h
weirddan455: It definitely can't be `uint16_t` since we have this constant `#define MAX_REDBOOK_SECTOR 1826090u`.  Could change it to be unsigned since negative values don't make sense.  I just left it `int` because that's ultimately what the `ioctl` returns and takes for sector values.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+			return;
+		}
+		cdrom_read_audio cd = {};
+		cd.addr.lba         = current_sector;
+		cd.addr_format      = CDROM_LBA;
+		cd.buf = reinterpret_cast<uint8_t*>(audio_buffer.get());
+		// Maximum this ioctl can read in a single call is 1 second (75
+		// redbook frames)
+		cd.nframes = std::min(sectors_remaining, CD_FRAMES);
+		constexpr int PCMFramesPerRedbookFrame = BYTES_PER_RAW_REDBOOK_FRAME /
```
johnnovak, line 331: Minor: Pcm instead of PCM (only capitalise the first letter of acronyms... yes, I know it's all over the place in the existing code...)

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+		                                         BYTES_PER_REDBOOK_PCM_FRAME;
+		frames_in_buffer = cd.nframes * PCMFramesPerRedbookFrame;
+		current_sector += cd.nframes;
+		sectors_remaining -= cd.nframes;
+		if (cdrom_fd == -1 || ioctl(cdrom_fd, CDROMREADAUDIO, &cd) != 0) {
```
johnnovak, line 336: Yeah, so this is too much... `is_opened` would be much better (or `is_open`).

johnnovak: I wonder why you're not checking for `is_open` at the top of the function though?

src/dos/cdrom.h
weirddan455: Nope.  Linux file descriptors explicitly return -1 for error so I'm using that as the default "invalid/initialize" value.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+			// Write out silence if something goes wrong
+			memset(audio_buffer.get(), 0, AudioBufferBytes);
```
johnnovak, line 338: Use vector then you can use `std::fill` 😄

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+		}
+	}
+
+	const size_t frames_available = frames_in_buffer - audio_position;
+	const size_t frames_to_write = std::min(static_cast<size_t>(requested_frames),
+	                                        frames_available);
+
+	// audio_position must be doubled since a frame is two int16_t's for stereo
+	// sound Does this need to be AddSamples_s16_nonnative on big endian?
+	mixer_channel->AddSamples_s16(frames_to_write,
```
johnnovak, line 351: Ideally, we'd want a full float mixer path; see this ticket for further details:
https://github.com/dosbox-staging/dosbox-staging/issues/1720

But this is kind of a special case as you can only retrieve the raw CD Audio data as a stream of integers. Anyway, just something to keep in mind for the future.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Could do that I guess.  I think it just makes things more complex.  When you do an `open`, it's always -1 if error.  Anything else is valid.  0 is even valid (it's stdin).

I'd rather do a macro or something if readability is a concern.  Those checks can maybe even be omitted entirely since the `ioctl` will fail and get handled later anyway (though it does avoid potentially expensive syscalls if we know they're going to fail ahead of time).

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +291,154 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+#endif
+		}
+	}
+
+	const size_t frames_available = frames_in_buffer - audio_position;
+	const size_t frames_to_write = std::min(static_cast<size_t>(requested_frames),
+	                                        frames_available);
+
+	// audio_position must be doubled since a frame is two int16_t's for stereo
+	// sound Does this need to be AddSamples_s16_nonnative on big endian?
```
johnnovak, line 350: Yes, you need `AddSamples_s16_nonnative`. Just do the same thing as `sblaster.cpp` for now.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Yeah, I stole those from `cdrom_image.cpp` :smile:

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Because it needs to be converted to `std::string`

src/dos/program_mount.cpp
```diff
@@ -462,20 +462,21 @@ void MOUNT::AddMessages() {
 	        "\n"
 	        "Notes:\n"
 	        "  - '-t overlay' redirects writes for mounted drive to another directory.\n"
+	        "  - '-usecd #' gives direct access to a CD-ROM drive (needed for CD audio).\n"
```
johnnovak, line 465: `-usecd ID` would be better, then explain in a few words what ID is supposed to be.

src/dos/cdrom.h
weirddan455: I was considering that.  I'm never using any vector operations though and it's always a fixed size.  The buffer is only used to read in one second's worth of data from a read `ioctl` and then pass it on (a millisecond or so at a time) to `AddSamples_s16`.  Both of which take raw pointers.

I'll change it if you want.  I don't have strong feeling about it.  But all I would be doing is calling `.data()` on it to get to the raw buffer so I'm not really sure what the benefit is.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: lol ok i'm seeing your point on this

src/dos/program_mount.cpp
weirddan455: Is it broken on Mac?  I don't have one.  I do dual boot Windows but haven't tested there.  Anyway, that section is already `#ifdef`'d for platform specific notes so I can put a note about that.

johnnovak:
Great work @weirddan455 🎉 

An overarching question/comment: it seems you're creating `CDROM_IOCTL_0`, `CDROM_IOCTL_1`, etc. mixer channels for every single mounted physical CD-ROM drive. That's problematic for a couple of reasons:

- Currently, we use the `CDAUDIO` mixer channel for mounted images that have CD audio, and I'm pretty sure it's the same deal with [FLAC/WAV/MP3/Opus tracks](https://dosbox-staging.github.io/downloads/release-notes/0.75.0/#support-flac-opus-and-mp3-cd-da-tracks) (but I've never personally used this feature).
- So at the very least the channel names should start with `CDAUDIO`, not `CDROM`. `IOCTL` seems also very OS specific; we don't want that. What if next month someone will add similar mixer-level CD Audio support for Windows? We don't want `CDAUDIO_IOCTL_0` on Linux, `CDAUDIO_WINDOWS_0`, then `CDAUDIO_MAC_0` or something some day... At the very least, I suggest using the names `CDAUDIO_0`, `CDAUDIO_1`, and so on. And then `CDAUDIO` is always dedicated to the mixer channel used by mounted CD images.
- But then, I'm not 100% sure it's a good thing to suddenly have 2-3-4+ CDAUDIO channels... In actual practice, how likely is it that CD Audio would be played simultaneously from multiple mounted CDs, and maybe from a mounted image too? Probably zero? I don't think this would ever happen in real life. Maybe technically possible, but it's more of a curiosity than something people would actually do. Therefore, unless you or someone else can convince me with some very strong arguments, I think we should just mix everything into the single `CDAUDIO` channel and job done. In 99.9999% of the cases (and I would place a good bet on this), only a single CD drive would stream its audio output into this channel. But in the extreme edge case scenario when we'd end up with output from multiple drives being mixed together, the user can simply just lower the volume of the `CDAUDIO` channel to avoid clipping (and yes, we can do without clipping because of the float mixing path!).

Okay, so let's just do that: single `CDAUDIO` mixer channel, and every kind of CD Audio output gets mixed into that. You can even test that when mounting a physical CD and a BIN/CUE image with audio tracks (well, maybe not the mixing of the two sources together as that would be pretty hard to trigger, but just that the audio gets mixed into the same mixer channel when both are mounted).

src/dos/program_mount.cpp
johnnovak: Based on what others have reported (I never tried it), it only works on Windows, so now we have Windows support but via the SDL mixer, not the DOSBox mixer, and with this feature Linux support via the DOSBox mixer, and that's it. No support on macOS.

src/dos/cdrom.h
johnnovak: Yeah, fair points but vector is just nicer and safer, and that would remove the need for the unique_ptr cruft and memset. Not that important for this specific use case, but it contributes to the overall consistency of the codebase. And in most cases there is no reason to really _prefer_ raw arrays over vector or even std:array.

weirddan455:
I don't think I can share a single mixer channel with the CDROM image class.  Whenever you create a mixer, it gets tied to a specific callback function (and if you use `std::bind` like I did and the image class does, it gets tied to a specific object as well).

I agree it's not ideal to be creating multiple channels.  I am creating mine lazily at least so the channel only gets created when music actually starts playing, rather when the device gets mounted.

src/dos/cdrom.h
johnnovak: Ah right, then it makes sense to use `int16_t.

src/dos/cdrom.h
johnnovak: Ok, it's been a while I used C 😄

johnnovak:
> I agree it's not ideal to be creating multiple channels. I am creating mine lazily at least so the channel only gets created when music actually starts playing, rather when the device gets mounted.

Yep, that's no good either because then the user cannot adjust the volume level of the channel before launching the game. It should be done when the drive is mounted.

Okay, I will look into sharing the same mixer channel across multiple devices. We should do this once and do it well.

src/dos/cdrom.h
johnnovak: Yeah, we try to do better than the loose C way of doing things, but probably ok for this specific case.

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Ah ok.

src/dos/program_mount.cpp
weirddan455: From what I've seen, SDL CD support doesn't use **any** mixer.  It doesn't actually read digital audio.  It just issues a command to the CD drive to tell it to start playing.  On Linux, this method is just broken on modern systems which is why I wrote this.  Maybe this method still works on Windows (this is how DOS games worked anyway) but there's no sound code here.  If anything gets mixed, it's up to the OS.

SDL does have code for Mac OSX, though no idea how well it works.  I'll test on Windows later.

weirddan455:
>  Yep, that's no good either because then the user cannot adjust the volume level of the CD Audio channel before launching the game. It should be done when the drive is mounted.

How would the user adjust the volume before loading the game?  I didn't know that was a feature.  Quake has a volume slider in-game and I tested that worked.

src/dos/cdrom.h
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

johnnovak:
> > Yep, that's no good either because then the user cannot adjust the volume level of the CD Audio channel before launching the game. It should be done when the drive is mounted.
> 
> How would the user adjust the volume before loading the game? I didn't know that was a feature. Quake has a volume slider in-game and I tested that worked.

Oh man, you're in for a treat! 😎 

image

image

Make sure to read these too (but you can read the whole audio section in the release notes; it's full of useful info):
https://dosbox-staging.github.io/downloads/release-notes/0.79.0/#general-mixer-improvements
https://dosbox-staging.github.io/downloads/release-notes/0.79.0/#reverb-and-chorus
https://dosbox-staging.github.io/downloads/release-notes/0.79.0/#crossfeed

A bunch of game specific audio configs by me:
https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-configuration-recommendations

Mixer diagram (I need to update it with some more recent additions):
https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-mixer-signal-flow-diagram

johnnovak:
Okay, I've done some digging and indeed, it would complicate things a lot to share mixer channels. `CDAUDIO_#` will do, where `#` is the ID of the device. That should be reasonably cross-platform too, should we add mixer-level CD Audio support on Windows and macOS in the future.

At the end of the day, this is a very niche feature that will eventually die out (as will the lasers in all CD-ROMs in existence in the not so distant future), so I don't want to make you do a lot of unnecessary work.

But the creation of the mixer channels must be done at the time the drive gets mounted.

src/dos/program_mount.cpp
johnnovak: >  It just issues a command to the CD drive to tell it to start playing.

Right, it's even worse then than I thought 😅 

> SDL does have code for Mac OSX, though no idea how well it works. I'll test on Windows later.

Thanks for that. I have a rather low interest in this feature, and the guy who (re-)introduced it disappeared 🤷🏻

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Nice one 🎉

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -347,9 +347,13 @@ void CDROM_Interface_Ioctl::CDAudioCallback(uint16_t requested_frames)
 	                                        frames_available);
 
 	// audio_position must be doubled since a frame is two int16_t's for stereo
-	// sound Does this need to be AddSamples_s16_nonnative on big endian?
```
johnnovak: minor: add back "sound" at the end of the sentence 😄

src/dos/program_mount.cpp
weirddan455: Yeah, that SDL code is pretty janky.  It's ripped out of version 1 which is just unmaintained nowadays.  If I was doing it, I would have at least tried to use the sdl12-compat library (re-implementation of the SDL 1 API using SDL 2 as a backend):  https://github.com/libsdl-org/sdl12-compat

That one looks like it tries to do digital audio properly.  I also have a low interest in messing with library code for other platforms though.

johnnovak: Fair enough. I wasn't involved, but I guess nobody was super intersted in this and went "yeah, whatever" 🤷🏻

src/dos/cdrom.h
weirddan455: I changed this to use `std::vector`

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

weirddan455:
> Okay, I've done some digging and indeed, it would complicate things a lot to share mixer channels. `CDAUDIO_#` will do, where `#` is the ID of the device. That should be reasonably cross-platform too, should we add mixer-level CD Audio support on Windows and macOS in the future.
> 
> At the end of the day, this is a very niche feature that will eventually die out (as will the lasers in all CD-ROMs in existence in the not so distant future), so I don't want to make you do a lot of unnecessary work.
> 
> But the creation of the mixer channels must be done at the time the drive gets mounted.

@johnnovak I think I've addressed all your concerns now.  I changed this to do as you said.  Good idea on using device ID for the name.  That gets rid of the loop checking for available names since we can just assume only 1 channel per device.

Thanks again for the fast review :smile:

FeralChild64:
I confirm that it works - on my at least the DOS Navigator can now play audio CDs on my Linux system :)

src/dos/cdrom.h
```diff
@@ -354,18 +349,44 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	                     unsigned long num);
+	bool PlayAudioSector(const uint32_t start, uint32_t len);
+	bool PauseAudio(bool resume);
+	bool StopAudio();
+	void ChannelControl(TCtrl ctrl);
+	bool LoadUnloadMedia(bool unload);
 
 private:
-	char	device_name[512];
+	void CDAudioCallback(uint16_t requested_frames);
```
johnnovak, line 376: CdAudio (lowercase `d`)

johnnovak: `const uint16_t`

src/dos/cdrom.h
```diff
@@ -354,18 +349,44 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	void InitAudio(int device_number);
```
johnnovak, line 377: `const int`

src/dos/cdrom.h
```diff
@@ -354,18 +349,44 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	bool IsOpen();
```
johnnovak, line 378: Can be `bool IsOpen() const;`?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -83,26 +269,47 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
-	bool success = CDROM_Interface_SDL::SetDevice(path, cd_number);
+	int fd = open(device_name, O_RDONLY | O_NONBLOCK);
+	if (fd == -1) {
+		return false;
+	}
+	if (IsOpen()) {
+		close(cdrom_fd);
+	}
+	cdrom_fd = fd;
+	return true;
```
johnnovak, line 288: Maybe put a comment there that `Open` also closes the currently open device (so effectively it "re-opens").

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -83,26 +269,47 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
+}
+
+bool CDROM_Interface_Ioctl::SetDevice(const char* path, const int cd_number)
+{
+	assert(path != nullptr);
```
johnnovak, line 293: minor: you could also assert for a non-blank string, I guess.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -83,26 +269,47 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
+		}
+	}
 
-	if (success) {
-		const char *tmp = SDL_CDName(cd_number);
-		if (tmp)
-			safe_strncpy(device_name, tmp, 512);
-		else
-			success = false;
+	for (int i = 0; i < num; i++) {
```
johnnovak, line 304: minor:
`auto i`
`++i`

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -83,26 +269,47 @@ bool CDROM_Interface_Ioctl::ReadSectors(PhysPt buffer,
...
+		auto cd_name = SDL_CDName(i);
+		if (cd_name && strcmp(cd_name, path) == 0 && Open(cd_name)) {
+			InitAudio(i);
+			return true;
+		}
 	}
```
johnnovak, line 311: Maybe log a warning if we couldn't find it?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +320,159 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
 
+void CDROM_Interface_Ioctl::InitAudio(int device_number)
+{
+	if (mixer_channel) {
+		return;
+	}
+
+	std::string name = std::string(MixerChannelNamePrefix) +
+	                   std::to_string(device_number);
+	audio_buffer.resize(AudioBufferBytes / 2);
```
johnnovak, line 331: I don't quite like this; can you derive the number of _samples_ instead from the constants and call this `AudioBufferSamples`? Then you don't need to divide by two.

To be clear:

1 sample frame = 2 samples (always stereo)
1 sample = int16_t

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +320,159 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+#endif
+		}
+	}
+
+	const size_t frames_available = frames_in_buffer - audio_position;
+	const size_t frames_to_write = std::min(static_cast<size_t>(requested_frames),
+	                                        frames_available);
+
+	// audio_position must be doubled since a frame is two int16_t's for stereo sound
+#ifdef WORDS_BIGENDIAN
```
johnnovak, line 378: Ok, forgot to mention this before... It gets all quite murky and confusing with the samples vs frames business. Some people also use the two interchangeably (not you, luckily 😅), which doesn't help either...

We introduced `AudioFrame` (`include/audio_frame.h`) for this very reason to simplify audio stream handling. I know exactly the distinction between frames and samples, but I managed to still confuse myself sometimes when doing buffer length manipulations, etc.; it's very easy to introduce bugs this way.

We have standardised on the use of `AudioFrames` in all audio rendering code that feeds the mixer. There are a few audio renderers that haven't been converted yet, but that's the ultimate goal. When that goal is accomplished, we will only keep `AddSamples_mfloat` and `AddSamples_sfloat`; every other `AddSample_*` variant will be nuked. That will simply `mixer.cpp` a lot too.

Some example on the use of `AudioFrame`:

#### gus.cpp
```cpp
using pan_scalars_array_t = std::array<AudioFrame, PAN_POSITIONS>;
```
#### gameblaster.cpp
```cpp
std::queue<AudioFrame> fifo = {};
```

####  opl.h
```cpp
std::queue<AudioFrame> fifo = {};
```

####  Misc signatures
```cpp
AudioFrame Process(const AudioFrame &in);
AudioFrame RenderFrame();
bool MaybeRenderFrame(AudioFrame &frame);
void Envelope::Process(const bool is_stereo, AudioFrame &frame)
```

The we can consistently use "frames" in our namings, and everything is crystal clear! For example:

```cpp
	while (frames_remaining) {
		if (float frame = 0.0f; MaybeRenderFrame(frame)) {
			channel->AddSamples_mfloat(1, &frame);
		}
		--frames_remaining;
	}
```

I should have said this earlier... Given our goals stated above, can I ask you to please migrate to using `AudioFrame` and use the term "frame" everywhere consistently? Thanks 😄

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +320,159 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+	mixer_channel->AddSamples_s16_nonnative(frames_to_write,
+	                                        audio_buffer.data() +
+	                                                (audio_position * 2));
+#else
+	mixer_channel->AddSamples_s16(frames_to_write,
+	                              audio_buffer.data() + (audio_position * 2));
```
johnnovak, line 384: Yeah, so `audio_position * 2` would simply become `frame_pos` as your buffer would be a vector (or `std::array`, up to you) of `AudioFrame`s.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +320,159 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+#endif
+
+	audio_position += frames_to_write;
+	if (audio_position >= frames_in_buffer) {
```
johnnovak, line 388: Yeah so this gets confusing... `audio_position` seems to be a "frame position"... but it's not called that.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +320,159 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+
+// Taken from CDROM_Interface_Image
+void CDROM_Interface_Ioctl::ChannelControl(TCtrl ctrl)
+{
+	if (!mixer_channel) {
+		return;
+	}
+
+	// Adjust the volume of our mixer channel as defined by the application
+	mixer_channel->SetAppVolume(ctrl.vol[0] / 255.0f, ctrl.vol[1] / 255.0f);
```
johnnovak, line 449: Extract `255.0f` to `MaxVolume`.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +320,159 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+
+	// Map the audio channels in our mixer channel as defined by the
+	// application
+	const auto left_mapped  = static_cast<LINE_INDEX>(ctrl.out[0]);
+	const auto right_mapped = static_cast<LINE_INDEX>(ctrl.out[1]);
+	mixer_channel->ChangeChannelMap(left_mapped, right_mapped);
```
johnnovak, line 455: 🥇 for getting this right. I assume you tested it that reversing channels via the mixer works?

src/dos/program_mount.cpp
johnnovak: So can you change `-usecd` to this (just format it properly, and feel free to reword it better and/or more succinctly):
```
'-usecd ID' gives direct access to a CD-ROM drive. This is needed for CD audio (only supported on Windows and Linux). Run 'mount -cd' to find out the list of valid IDs.
```

src/dos/cdrom_ioctl_linux.cpp
johnnovak: > I wrote an `IsOpen()` function that checks whether the file descriptor is in error state. I'm a little afraid if I make a separate boolean that there could be bugs if one of the variables gets changed but not the other.

It's fine, it's not that complex. We do similar things in a number of places.

> Hopefully this improves readability while keeping the data in one place.

Yes, I think it does. Reduced duplication and gets rid of "deriving" the state from magic values in file descriptors, which although not a huge deal, it's not ideal either.

johnnovak:
> @johnnovak I think I've addressed all your concerns now. I changed this to do as you said. Good idea on using device ID for the name. That gets rid of the loop checking for available names since we can just assume only 1 channel per device.
> 
> Thanks again for the fast review 😄

My pleasure 😄 You're making great progress and adding new enhancements to Staging at a quick pace! 🎉 

Thanks for addressing the changes, they all look good. I've made another round of comments (I missed some stuff yesterday), but we should be pretty close now!

src/dos/cdrom_ioctl_linux.cpp
weirddan455: That `AudioFrame` struct only stores floats (well it has a constructor for `int16_t` but it implicitly converts to float).

I'm a little confused how you want me to use this.  I've done a bit of audio coding in the past so I understand the value of wanting to do all your mixing in floats.  However, it's also my understanding you usually want to normalize your floats from -1.0f to 1.0f so I don't know if this cast will cut it.

Like you said earlier, CD audio is different from a SoundBlaster emulator or what have you in that I'm literally just reading `int16_t`'s off a disk.

src/dos/program_mount.cpp
weirddan455: I changed it to the requested text.  Not sure how best to format it.  You gave me 3 lines of text there and now the help command is paging.

mount

src/dos/cdrom.h
weirddan455: Fixed

src/dos/cdrom.h
weirddan455: Fixed

src/dos/cdrom.h
weirddan455: Fixed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Asserted

src/dos/cdrom_ioctl_linux.cpp
weirddan455: auto'd

src/dos/cdrom_ioctl_linux.cpp
weirddan455: warned

src/dos/cdrom_ioctl_linux.cpp
weirddan455: I renamed `audio_position` to `frame_position` but it still has to be multiplied by 2 because the vector is in samples.  See my above comment about `AudioFrame`.  Need some clarification there.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: renamed to `frame_position`

src/dos/cdrom_ioctl_linux.cpp
weirddan455: extracted

src/dos/cdrom_ioctl_linux.cpp
weirddan455: I'm don't know what the mixer command is for reversing.  It's not in the help page.  I just copied that entire function from `cdrom_image.cpp` so I assume it works.  I know the volume setting does by using Quake's volume slider and seeing the debug logs.

src/dos/cdrom_ioctl_linux.cpp
johnnovak: All good questions, and I'm here to help! Feel free to ask anything about audio if something is unclear.

Yes, using a `[-1.0f, 1.0f]` value range is common in audio processing, but that's just a convention—you can use any range you want, of course.

Currently, the "native" range of the mixer is `[INT16_MIN, INT16_MAX]` internally. You can see evidence of that here; the master compressor I introduced assumes `INT16_MIN` for the [0 dBFS](https://en.wikipedia.org/wiki/DBFS) level (more or less equivalent to "peak amplitude" in common language).
https://github.com/dosbox-staging/dosbox-staging/blob/main/src/hardware/mixer.cpp#L456

So eventually everything has to be normalised to `[INT16_MIN, INT16_MAX]` in our mixer, but you can configure the _expected_ sample value ranges per mixer channels individually with `SetPeakAmplitude()` and `Set0dbScalar()` (expected in the sense of what you pass into the channel with `AddSamples_*`):
https://github.com/dosbox-staging/dosbox-staging/blob/main/src/hardware/mixer.cpp#L840-L845
https://github.com/dosbox-staging/dosbox-staging/blob/main/src/hardware/mixer.cpp#L604-L614

You can read about the whole volume/peak scaling business here:
https://github.com/dosbox-staging/dosbox-staging/blob/main/include/mixer.h#L261-L295

By default, the peak of a mixer channel is `MAX_AUDIO`
https://github.com/dosbox-staging/dosbox-staging/blob/main/include/mixer.h#L291-L295

which is basically `INT16_MAX`
https://github.com/dosbox-staging/dosbox-staging/blob/main/include/mixer.h#L68

So the TL;DR is that yes, you can just cast the raw `int16_t` samples to `float` and simply use `Add_Samples_sfloat` and things will work. That way you will push samples into the mixer channel in our "native" `[INT16_MIN, INT16_MAX]` range. That's a lot of words for confiming "yes, just case to float", but I guess I wanted to explain it to you how it all fits together 😎 

Maybe it's a bit confusing and abstract, but check out how different audio devices set these mixer channel parameters, that should clear things up. However, you only need to cast to floats here.

johnnovak: (And yes, probably I will refactor this all at some point so we use the more common `[-1.0f, 1.0f]` range. There are evolutionary reasons for normalising to the `int16_t` range, but I will rework that when I add 24-bit and 32-bit float audio capture support; then only the `[-1.0f, 1.0f]` internal range will make sense which can be written straight to disk for the 32-bit float capture).

src/dos/program_mount.cpp
```diff
@@ -462,7 +462,9 @@ void MOUNT::AddMessages() {
 	        "\n"
 	        "Notes:\n"
 	        "  - '-t overlay' redirects writes for mounted drive to another directory.\n"
-	        "  - '-usecd #' gives direct access to a CD-ROM drive (needed for CD audio).\n"
+	        "  - '-usecd ID' gives direct access to a CD-ROM drive.\n"
+	        "This is needed for CD audio (only supported on Windows and Linux).\n"
+	        "Run 'mount -cd' to find out the list of valid IDs.\n"
```
johnnovak, line 467: Nice, thank you 👍🏻

johnnovak: One thing, can you please indent these two new lines with the first line of the bullet point? So just add 4 spaces in front, if I'm counting them right 😄

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Interesting.  Yeah, if you get it normalized to -1.0f to 1.0f, you can also feed that directly to SDL (had a quick look and we're currently doing a conversion back to `int16_t` first).  It can accept floating point format but I believe it expects it to be normalized in that manner.

Most modern sound systems can accept float (I know PulseAudio in Linux will) but if not SDL will automatically convert it to what the platform needs.  I vaguely remember reading that Mac OSX **only** accepts floats so if that's true you end up going `int16_t->float->int16_t` then back to float again once SDL does the final conversion internally.

src/dos/cdrom_ioctl_linux.cpp
johnnovak: image

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Yeah, maybe 🤷🏻 0.00001% performance drop, I guess 😄 But it will be [-1.0f, 1.0f] anyway for the above outlined reasons in the future.

src/dos/cdrom_ioctl_linux.cpp
johnnovak: As per my long explanation. You'll have this instead:

```
std::vector<AudioFrame> audio_buffer = {};
```

Then you simply pass yout `int16_t` samples to this contructor:

```
constexpr AudioFrame(const int16_t l, const int16_t r)
```

And push the AudioFrames into the vector. That's it. Make sure to remove every trace of referring to "samples"; the code should only refer to "frames" after this.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Anyway, back to this PR.  I'm still not sure how you want me to use this `AudioFrame` struct.  I can do the conversion from `int16_t` to `float` in my callback function, no problem.  I think that would require two buffers (an `int_16` "read_buffer" to pass to the `ioctl` and then an `output_buffer` with float data to pass to the mixer.  But that's fine if that's what you want.

But then you said you wanted my buffer to be a vector of `AudioFrame`.  Problem is... the mixer doesn't accept that type.  It needs to be a vector of float.  I could `reinterpret_cast` and hope it works out (and it probably will until someone makes changes since the data is just 2 floats) but I think it's bordering on undefined behavior.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Not sure how I missed that.  Anyway, my ears really can't tell a difference.  Also, it looks like the mixer code never calls into that function anyway.  That function gets called when a game tries to modify something.  When the user uses the "mixer" command, it just modifies the mixer channel directly without going through the drive.

https://github.com/dosbox-staging/dosbox-staging/blob/40f2a0ac547a53bf4ccf95906a82cb16ec9fed4d/src/hardware/mixer.cpp#L2360-L2361

https://github.com/dosbox-staging/dosbox-staging/blob/40f2a0ac547a53bf4ccf95906a82cb16ec9fed4d/src/hardware/mixer.cpp#L1813-L1828

src/dos/cdrom_ioctl_linux.cpp
weirddan455: >  And push the AudioFrames into the vector. Add the results to the channel with AddSamples_sfloat. That's it.

`AddSamples_sfloat` won't take a vector of `AudioFrame` is the problem.  I'd be creating a vector that I can't really do anything with.  The add samples function just wants a raw pointer to some floats.

src/dos/cdrom_ioctl_linux.cpp
johnnovak: It will be fine :) If you check out the mixe code in more detail, you'll see those left/right mappings are indices; swapping is accomplished by swapping the 0 and 1 indices.

johnnovak: > Anyway, my ears really can't tell a difference. 

Yes, you won't hear much of a difference unless the audio has some strong stereo content with instruments panned hard left or right. Cymbals and hihats are usually either on the left or right; maybe try listening to that (Quake OST is quite industrial, from memory, with lots of drums).

weirddan455: I mean it's really hard to tell.  My point was that typing `mixer reverse` never calls into the class I wrote so that test isn't really relevant anyway.  That's all done in the mixer code.  It's a different variable even.  `channel_map` is for when a DOS application wants to change its mapping.  `output_map` is for when the user changes mixer settings (as far as I can tell).

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Just do this like every other audio device that uses `AddSamples_sfloat`. Look up the usages, please. `audio_frames` is a `std::vector<AudioFrame>` in this example:

```cpp
channel->AddSamples_sfloat(requested_audio_frames, &audio_frames[0][0]);
```

Again, look at the bigger picture. When everything in the codebase will only push floats to the mixer via `AddSamples_sfloat` and `AddSamples_mfloat`, we can remove that huge templated `AddSamples` junk and introduce two much simpler `AddSamplesStereo` and `AddSamplesMono` functions or something that just take `std::vector<AudioFrame>` directly.

One step at a time...

src/dos/cdrom_ioctl_linux.cpp
johnnovak: It's a bit more complex, but I don't have time to explain this in detail now... You're doing the same thing as `cdrom_image.cpp`, so it's fine. Moving on 😄

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Okay, I'm kinda busy now, this is surprising. There shouldn't be _that_ much of a performance difference... but then, debug builds are naturally 4-5x slower, or a lot more in some cases. Consider switching to `std::array` perhaps? But probably you'd experience the same relative slowness in debug builds because of the bounds checks or something.

At the end of the day, only the performance of the release build matters.

johnnovak: Btw, only use emplace, that's the potentially more performant version as it skips creating the temporary.

Don't use debug builds for judging the performance of any code, it's meaningless.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -356,45 +363,51 @@ void CDROM_Interface_Ioctl::CdAudioCallback(const uint16_t requested_frames)
...
+		for (size_t i = 0; i < samples; i += 2) {
+			const int16_t left  = host_to_le16(input_buffer[i]);
+			const int16_t right = host_to_le16(input_buffer[i + 1]);
+
+			// output_buffer.emplace_back(left, right);
+
+			// For some reason, emplace_back is much slower
+			// in debug build and causes noticible popping
+			// This makes the loop 4ms instead of 10ms on my system.
+			AudioFrame frame(left, right);
```
johnnovak, line 396: It's fine, please use emplace and remove this comment. This happens because filling up 1s worth of data takes just a bit too long, which results on starving the audio buffer occasionally in debug builds. You can increase the buffer size in your config to work around this somewhat in debug builds, but it's normal. Debug builds are dog slow with certain settings and can crackle like hell... Also, probably you won't get crackling if you used a standalone audio player program with low cycles settings. Quake with max cycles setting will cause crackles in debug builds, it's not a surprise. Maybe try Monkey Island CD edition with around 3000-5000 cycles. But ultimately it's ok. We could put comments like these into every single file about debug builds 😄

src/dos/cdrom.h
```diff
@@ -354,18 +349,44 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+
+	int cdrom_fd                      = -1;
+	mixer_channel_t mixer_channel     = nullptr;
+	std::vector<int16_t> input_buffer     = {};
+	std::vector<AudioFrame> output_buffer = {};
+	size_t frame_position                 = 0;
+	int current_sector                = 0;
+	int sectors_remaining             = 0;
+	bool is_playing                   = false;
+	bool is_paused                    = false;
```
johnnovak, line 389: Please reformat.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +324,175 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+
+	// Input buffer is used as a C-style buffer passed to an ioctl.
+	// Its size must always be 1 second of data.
+	// Stored in int16_t samples.
+	input_buffer.resize(REDBOOK_PCM_FRAMES_PER_SECOND * REDBOOK_CHANNELS);
+
+	// Output buffer uses C++ vector operations.
+	// Its max capacity will be 1 second so go ahead and reserve that.
+	// Stored in AudioFrame structs.
+	output_buffer.reserve(REDBOOK_PCM_FRAMES_PER_SECOND);
```
johnnovak, line 344: Yep, as explained, better not to do this. Only convert as many `int16` samples from the `input_buffer` that the audio callback requested to even out performance spikes. And even then let the `output_buffer` resize itself lazily—the number of requested samples will hover around the same value.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: So I was definitely measuring the wrong thing when I said "release builds are mostly the same".  That was wrong.  I did a micro-benchmark of just the loop that was converting 1 second of audio from int to float.

Debug with push_back - 3ms
Debug with emplace_back - 10ms
Release build with either (times are about the same) - 150 **microseconds**

So yeah... almost two orders of magnitude slower in the worst case.  I see your point about not trusting debug performance now.

I think this won't ever be a problem with release builds based on that but I will go ahead and take your suggestion to do the conversion as needed.  If anything, it saves a bit of memory.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: I've done the on demand float conversations and that seems to have solved the performance issues on debug builds.

src/dos/cdrom.h
weirddan455: It's been squashed and ran through the format script again.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Reserve has been removed.  Output buffer will be much smaller now (45 samples or so instead of a full second).

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -113,4 +324,175 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
+	const size_t frames_to_write = std::min(static_cast<size_t>(requested_frames),
+	                                        frames_available);
+
+	mixer_channel->AddSamples_sfloat(frames_to_write,
+	                                 &output_buffer[frame_position][0]);
+
+	frame_position += frames_to_write;
+	if (frame_position >= output_buffer.size()) {
+		frame_position = 0;
+	}
```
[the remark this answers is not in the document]
weirddan455: Great explanation.  I did actually do that logging while I was initially writing this and was curious why it was being called so frequently and asking for such a low amount of samples.  Now I know why.

I am a little bit concerned now about the `ioctl` call in the callback.  It was taking 3ms or so to complete (which seems bad if the entire tick is supposed to take 1ms, though the spike only happens once per second) but I haven't noticed any problems in practice.  The `input_buffer` size could be reduced.  It can be anywhere from 1 to 75 redbook frames (75 being 1 second).  I was just using the max to reduce the number of  total syscalls and reducing it probably just means smaller but more frequent latency spikes.

Ultimately though, it's reading from a slow optical drive so I don't know if there's much that can be done short of doing reads on a separate thread.  So maybe it's fine how it is if it doesn't appear to cause any problems?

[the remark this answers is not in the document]
johnnovak: Finding the optimal number of redbook frames to read is easy enough to do, so I suggest to play around with it a bit (in release builds), and see if you can lower the call latency.

Ultimately, if this works well enough, then I'm happy for you to merge it. Yes, threading  is the best way to go about it, but I don't want to test your patience 😅 ; you've done a stellar job on this feature already! 🚀 🎉 

However, if you're _really_ interested in adding threading (entirely optional!), I'm here to help. Even then, I suggest merging this and adding threading in a secon PR. The patterns are already laid out, so it's a matter of following the `midi_mt32.cpp` or `midi_fluidsynth.cpp` implementations that use `std::thread` and our nice `RWQueue` (in `rwqueue.h`). But more on that if you ask me 😄

johnnovak:
This would be a better commit name (to stress this is for Linux only):

Implement CD Audio on Linux using ioctls

weirddan455:
I've renamed the commit.  I also did some quick performance testing in a release build and reduced the maximum number of redbook frames to read from 75 to 50.  This reduces the spikes from 4.1ms average to 2.5ms and slightly lowers the total time spent in ioctls over 10 seconds (38ms down to 37ms).

It's never going to be very fast though without threading.  Just to test, I changed it to read max of 1 redbook frame.  This takes 1ms and increases total time spent in ioctls over 10 seconds to 670ms.  Definitely seeing diminishing returns here.  Although even with this (very sub-optimal) configuration, I did not notice any pops in the audio.

I also created some constants at the top of the file so this value can be tweaked in the future by simply modifying the `InputBufferMaxRedbookFrames` constant.

With that said, I think this is now good to merge if everything looks OK to you.

johnnovak:
> also did some quick performance testing in a release build and reduced the maximum number of redbook frames to read from 75 to 50. This reduces the spikes from 4.1ms average to 2.5ms and slightly lowers the total time spent in ioctls over 10 seconds (38ms down to 37ms).

That's an improvement. I get it that the overall time spent when reading a singles frames per call would be way too high, but what if you say read 25 frames at a time instead of 50? If the total time increases even by 30-50% but the spikes are cut in half again, that would make it well worth it.

Perhaps you can try that quickly (if you want), but approving otherwise. Nice work, and thanks for taking my comments on board 👍🏻

johnnovak (approved):
Great job @weirddan455 🎉

weirddan455:
At 25 frames, it's about 1.7ms average and total time of 50ms.  I'm not sure if this is an improvement or not.  This is where the diminishing returns start.

johnnovak:
> At 25 frames, it's about 1.7ms average and total time of 50ms. I'm not sure if this is an improvement or not. This is where the diminishing returns start.

Yeah, I think it's well worth it. Let's do 25 frames then, and then let's merge this. Stutter-free audio is the most important—slightly higher CPU usage, who cares 😄

weirddan455:
>   Yeah, I think it's well worth it. Let's do 25 frames then, and then let's merge this. Stutter-free audio is the most important—slightly higher CPU usage, who cares smile

Right, I wasn't concerned about CPU usage.  This reads twice as often so you hit that potential stutter twice as often was my line of thinking.  But like I said, I don't really have a way to gauge which is better (both sound find on my system).

So changed to 25.

Issue #2616, opened 2023-06-19
Reinstate the SoundFont level volume setting


johnnovak:
It was a mistake to remove that but it's easy to restore it. Mea culpa.

PR #2677, opened 2023-07-12
Un-deprecate the per-SoundFont volume scaling

johnnovak:
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2616 plus other small cleanups, fixes, enhancements—as usual.

Review per commit.

### DIFF
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -373,7 +373,7 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 			                                0.0,
 			                                21.0);
 
-			if (chorus[4] == "triange") {
+			if (chorus[4] == "triangle") {
 				chorus_mod_wave = fluid_chorus_mod::FLUID_CHORUS_MOD_TRIANGLE;
 			} else if (chorus[4] != "sine") { // default is sine
 				LOG_WARNING("FSYNTH: Invalid chorus modulation wave type ('%s'), "
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -42,13 +42,15 @@
 
 MidiHandlerFluidsynth instance;
 
-static void init_fluid_dosbox_settings(Section_prop &secprop)
+constexpr auto SoundFontExtension = ".sf2";
+
+static void init_fluid_dosbox_settings(Section_prop& secprop)
 {
 	constexpr auto when_idle = Property::Changeable::WhenIdle;
 
-	// Name 'default.sf2' picks the default soundfont if it's installed
+	// Name 'default.sf2' picks the default SoundFont if it's installed
 	// in the OS. Usually it's Fluid_R3.
-	auto *str_prop = secprop.Add_string("soundfont", when_idle, "default.sf2");
+	auto* str_prop = secprop.Add_string("soundfont", when_idle, "default.sf2");
 	str_prop->Set_help(
 	        "Path to a SoundFont file in .sf2 format ('default.sf2' by default).\n"
 	        "You can use an absolute or relative path, or the name of an .sf2 inside the\n"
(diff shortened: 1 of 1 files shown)


weirddan455 (commented):
I'm not really familiar with the FluidSynth code but this all looks fine as far as I can tell.

johnnovak:
Someone please give me a tick? @FeralChild64 ?

src/midi/midi_fluidsynth.cpp
```diff
@@ -139,7 +148,8 @@ static std::deque<std_fs::path> get_data_dirs()
 	return {
 	        get_platform_config_dir() / "soundfonts",
 
-	        // C:\soundfonts is the default place where FluidSynth places default.sf2
+	        // C:\soundfonts is the default place where FluidSynth places
+	        // default.sf2
 	        // https://www.fluidsynth.org/api/fluidsettings.xml#synth.default-soundfont
 	        "C:\\soundfonts\\",
```
FeralChild64, line 154: This value is good for Windows, but for Linux I suggest to use `/usr/share/soundfonts/default.sf2`; not sure about macOS.

FeralChild64:
I have just one remark (actually a pre-existing problem).

src/midi/midi_fluidsynth.cpp
FeralChild64: According to the website mentioned in the comment, it should be `${CMAKE_INSTALL_PREFIX}/share/soundfonts/default.sf2` by default - it seems distros decided otherwise, on Gentoo the `media-sound/fluid-soundfont` installs into `/usr/share/sounds/sf2/` too.

johnnovak: Sure, happy to change to whatever. In practice it makes little difference as I doubt Linux distros include a GS compatible SoundFont by default, no? If "GeneralUser GM GS" was the default, that would be cool, though.

johnnovak: Okay @FeralChild64, I'm far from being an expert on Linux distro directory conventions, but I vaguely remember kcgen mentioned all these path lookups on Linux are for distro packagers, and looking at this commit e1aa0551b3252c it seems to be fine as it is based on @dreamer 's commit message?

image

I can change it to whatever, but then you need to tell me exactly what to do. Better yet, if you still think the Linux SF2 lookup logic needs changing, and it's not a simple one-liner change, let's merge this and you can address that in another PR (and as a Linux user, you can test it on Linux).

weirddan455: It sounds like most distros put them somewhere in `/usr/share` (which is usually what is in the `XDG_DATA_DIRS` variable) but this is descrepency in what subfolder (`soundfonts` vs `sounds`) and I don't know of an environment variable for that.  On Arch Linux, I get this:

```
$ echo $XDG_DATA_DIRS
/usr/local/share:/usr/share
```
Two of the popular soundfonts on Arch are [freepats](https://archlinux.org/packages/extra/any/freepats-general-midi/) and [fluid](https://archlinux.org/packages/extra/any/soundfont-fluid/).  They both get stored in `/usr/share/soundfonts/`

However, I have no `default.sf2` (I don't think any package in Arch contains that file) so it's going to fail to find regardless.

weirddan455: Although reading this code more fully now, I'm a bit confused why we're discussing this.  The line that we're reviewing here is inside an `if defined (WIN32)`

The Linux version of this function already checks both `/usr/share/soundfonts` as well as `/usr/share/sounds/sf2`

https://github.com/dosbox-staging/dosbox-staging/blob/21ea3c587fcca1088dc31ddca954ee587f9d597f/src/midi/midi_fluidsynth.cpp#L160-L181

I tested on my system and it correctly finds my soundfonts if I type only the name of the file.  It automatically finds the correct path.

johnnovak: @FeralChild64 please share your thoughts 😄 I still think we should merge this, then I let you guys address any Linux specific issues in separate PRs, if needed. Looks like it's not needed, but again, I'm not the best person for this job as I don't use Linux.

FeralChild64: @johnnovak I didn't notice this is a Windows-specific code path; sorry for confusion. Let's leave it as is.

FeralChild64 (approved):

src/midi/midi_fluidsynth.cpp
johnnovak: > @johnnovak I didn't notice this is a Windows-specific code path; sorry for confusion. Let's leave it as is.

No problem 😄

PR #2685, opened 2023-07-15
Close file descriptor in cdrom ioctl class

weirddan455:
I forgot to clean up after myself and leaked a file descriptor.  Just need to add a `close` in the destructor.


johnnovak (approved):
Good catch @weirddan455 !

Issue #2671, opened 2023-07-10
Fullscreen doesn't work in SDL Wayland mode


shmerl:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.1

### What Operating System are you using?

Linux x86_64

### If Other OS, please describe

Debian testing

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When trying to run some games (Dune, Dune 2) using `-fullscreen` with SDL video driver set to Wayland, the window isn't scaling to screen size, unlike when using X11 SDL mode.

### Steps to reproduce the behaviour.

The variable I'm setting:

```
SDL_VIDEODRIVER='wayland'
```

### Your configuration

```ini
[sdl]
fullscreen          = false
display             = 0
fullresolution      = desktop
windowresolution    = default
viewport_resolution = fit
window_position     = auto
window_decorations  = true
transparency        = 0
host_rate           = auto
vsync               = false
vsync_skip          = 7000
presentation_mode   = auto
output              = opengl
texture_renderer    = auto
waitonerror         = true
priority            = auto auto
mute_when_inactive  = false
pause_when_inactive = false
mapperfile          = mapper-sdl2-0.80.1.map
screensaver         = auto

[dosbox]
language           = 
machine            = svga_s3
captures           = capture
memsize            = 16
mcb_fault_strategy = deny
vmemsize           = auto
dos_rate           = default
vesa_modes         = compatible
speed_mods         = true
autoexec_section   = join
automount          = true
startup_verbosity  = auto

[render]
aspect             = true
monochrome_palette = white
cga_colors         = default
glshader           = default

[composite]
composite   = auto
era         = auto
hue         = 0
saturation  = 100
contrast    = 100
brightness  = 0
convergence = 0

[cpu]
core      = auto
cputype   = auto
cycles    = auto
cycleup   = 10
cycledown = 20

[mouse]
mouse_capture        = onclick
mouse_middle_release = true
mouse_sensitivity    = 100
mouse_raw_input      = true
dos_mouse_driver     = true
dos_mouse_immediate  = false
ps2_mouse_model      = intellimouse
com_mouse_model      = wheel+msm

[mixer]
nosound    = false
rate       = 48000
blocksize  = 512
prebuffer  = 20
negotiate  = true
compressor = true
crossfeed  = off
reverb     = off
[...]

PR #2689, opened 2023-07-16
Use dynamic SDL version checking for logging

weirddan455:
This was using a compile time constant, rather than showing what the user is actually running.  Brought to my attention in #2671 and then confirmed by running the pre-compiled 0.81.0 binary from the DOSBox Staging website.  Apparently it was compiled against an ancient SDL 2.0.8 version :roll_eyes: 

There's also some sketchy checks against this compile time constant version:

https://github.com/dosbox-staging/dosbox-staging/blob/56a52cfceb1de0f5621228675789dd4b29653648/src/gui/sdlmain.cpp#L4154

https://github.com/dosbox-staging/dosbox-staging/blob/56a52cfceb1de0f5621228675789dd4b29653648/src/gui/sdlmain.cpp#L4870

I don't know enough about what the code is doing so I left it out for now.  However, I suspect the code that is inside these macros never made it out to our users since we are building with an old SDL version and this would have just gotten `#ifdef`'d out.

If we want this to be checking against what the user is actually running, we need to be using the dynamic `SDL_GetVersion()` function.

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4886,8 +4886,10 @@ int sdl_main(int argc, char *argv[])
 	// Once initialised, ensure we clean up SDL for all exit conditions
 	atexit(QuitSDL);
 
+	SDL_version sdl_version;
+	SDL_GetVersion(&sdl_version);
 	LOG_MSG("SDL: version %d.%d.%d initialised (%s video and %s audio)",
-		SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_PATCHLEVEL,
+		sdl_version.major, sdl_version.minor, sdl_version.patch,
 		SDL_GetCurrentVideoDriver(), SDL_GetCurrentAudioDriver());
 
 	const auto config_path = get_platform_config_dir();


johnnovak:
Nice improvement @weirddan455! 🎉 Yes, it's much better to do these checks at runtime and to make them dynamic. Note that we're always including the SDL library in our Windows and macOS distributions, so you can only get the SDL version mismatches you pointed out on certain Linux builds (but I'm unsure about the Linux packaging situation, but I assume by default we use the dynamic libraries available at the OS level).

For reference, this is what I'm getting currently on macOS (same results with `main` and your branch, unsurprisingly):

```
2023-07-17 13:08:07.783 | SDL: version 2.26.4 initialised (cocoa video and coreaudio audio)
```

Those SDL version checks in `sdlmain.cpp` should likely be dynamic, but we should be careful with changing that stuff. In any case, that will be @GranMinigun's territory in his `sdlmain.cpp` rework, so I'll just mention the versioning issue in that PR.

johnnovak (approved):
Nice one! 🚀

weirddan455:
>  Note that we're always including the SDL library in our Windows and macOS distributions, so you can only get the SDL version mismatches you pointed out on certain Linux builds

@johnnovak Because of a cool feature SDL has, you can actually get version mismatches on any platform, even if SDL is statically linked.  See Dynamic API:  https://sdl-mirror.readthedocs.io/en/latest/README-dynapi.html

The tl;dr is you can do `export SDL_DYNAMIC_API=/usr/lib/my-updated/sdl.so` and SDL will, at runtime, do a `dlload` or `LoadLibrary` call and load up your updated version of SDL, patching an internal jump table it maintains.

johnnovak:
> > Note that we're always including the SDL library in our Windows and macOS distributions, so you can only get the SDL version mismatches you pointed out on certain Linux builds
> 
> @johnnovak Because of a cool feature SDL has, you can actually get version mismatches on any platform, even if SDL is statically linked. See Dynamic API: https://sdl-mirror.readthedocs.io/en/latest/README-dynapi.html
> 
> The tl;dr is you can do `export SDL_DYNAMIC_API=/usr/lib/my-updated/sdl.so` and SDL will, at runtime, do a `dlload` or `LoadLibrary` call and load up your updated version of SDL, patching an internal jump table it maintains.

Uh 😬 As long as it's not mandatory... I like static linking.

But I guess it can help with games that never get updated to a newer SDL... maybe.

rderooy:
I'm wondering it it would not be a good thing to do a compare on startup between compiled and runtime SDL versions, and print a warning message to the logs in case of a mismatch...

johnnovak:
> > I'm wondering it it would not be a good thing to do a compare on startup between compiled and runtime SDL versions, and print a warning message to the logs in case of a mismatch...
> 
> Well, a mismatch is going to be the normal case on Linux since we don't ship SDL with our binaries and distros that ship DOSBox Staging aren't necessarily going to re-build it every time they update SDL.
> 
> Also, in the case of a mismatch, I think it will nearly always be that the user has a newer version, which shouldn't cause problems. The Dynamic API will refuse to load an older version and in the case of dynamic linking, it's likely to fail to execute due to missing symbols.

Can't we disable the dynamic linking? Personally, I don't like getting random dependencies... that's makes troubleshooting hell. I would force SDL (if possible) to _not_ do the dynamic lookup, then we have _one_ recommended SDL version per release we've tested with. But give the users the ability to enable dynamic loading, with the caveat that it's untested so they're on their own... "Pinning" a known-good SDL version per release is pretty important, I think. I really don't like this "loose" dependency handling anywhere... certainly not at work, I pin everything.

weirddan455:
So, there's two things at play here.  Just to clarify so we don't get them mixed up:

**1. The OS's normal dynamic loader**

On Windows, we ship an installer with a crap-load of .dll files that get put in the DOSBox Staging directory, including one for SDL.  Pretty typical for Windows software because Windows has no `/usr/lib` equivalent.  The user could, of course, replace those .dlls with a newer version.

On Linux, we don't ship any dynamic libraries.  [We just provide instructions to tell the user to install it with their package manager.](https://dosbox-staging.github.io/downloads/linux/)  SDL is a very common dependency and there's a good chance it's installed right out of the box.  I believe some of the less common dependencies like FluidSynth and Munt might be statically linked (not 100% sure, I'd have to check).  The Linux dynamic loader won't even look in current directory for libraries unless you explicitly tell it to with an environment variable.  It expects everything to be `/usr/lib`

You could statically link to SDL in the binary on the website but that doesn't really get us off the hook for making sure it works with the newest SDL version. [For example, we're in Fedora's repository now.](https://src.fedoraproject.org/rpms/dosbox-staging) and they are almost certainly dynamically linking to whatever version of SDL they ship, which can be updated at any time by them.

**2.  SDL's Dynamic API**

This happens after the OS's dynamic loader.  I believe it only does its thing if the user explicitly sets the `SDL_DYNAMIC_API` environment variable so I don't think it really hurts anything to leave it on but it can be turned off if you really want by a build flag.  It's mostly useful so users can patch in an updated version of SDL to fix bugs for closed source games that are statically linked (and may be abandoned by their developers).

**My opinion**

I think everything is fine how it is personally.  SDL's recommendation is to dynamically link.  Shipping a .dll is very common in the Windows-land and using common system libraries is very common in Linux-land.  I have no idea how you Mac users do things so do whatever you think is best there :smile:

johnnovak:
Thanks for clearing that up @weirddan455. Yeah I was kinda talking about the link statically scenario against our pinned version, then people can fool around with `SDL_DYNAMIC_API` if they feel adventurous 😝 

The situation on Windows and macOS is good and manageable; our library versions are effectively pinned on those platforms. MacOS apps have a similar model to Windows and Flatpak on Linux; they bundle copies of the 3rd party 
dependencies, and generally the only dynamic dependency is for fundamental dynamic libraries ("frameworks") that are part of the OS.

If I suddenly I became a Linux user, I guess I'd advocate strongly for linking everything statically (especially SDL) and/or providing a Flatpak distribution... but I'm not, so I'm happy with whatever you guys are happy with 😄 Distro maintainers are not in the position to make library upgrade decisions; they surely won't put as many hours into testing a specific DOSBox Staging version as we do (and perhaps wouldn't even know where to start to test... "application starts up" is not testing).

weirddan455:
>  and/or providing a Flatpak distribution

Well, there is a Flatpack:  https://flathub.org/apps/io.github.dosbox-staging

And a Snap:  https://snapcraft.io/dosbox-staging

I don't use either of those personally but that would be "pinned version" solution.

>  Distro maintainers are not in the position to make library upgrade decisions; they surely won't put as many hours into testing a specific DOSBox Staging version as we do

Right, so that's the main reason I want to dynamically link, at least for development builds.  That way we get to test with different SDL versions.  I'm on Arch which is a more "bleeding edge" distro so I get to test with the new SDL versions first.

Dynamic linking also makes it easier on distro maintainers because they have rules that say they **are required** to dynamically link everything.

If you want to statically link the Linux binary blob release on the website though, that's probably fine.  My suspicion is that most of our Linux users probably get it elsewhere.  Either directly from their distro, a community maintained thing like the  Arch AUR or Ubuntu PPA, or from Flatpack or Snap.

johnnovak:
> Right, so that's the main reason I want to dynamically link, at least for development builds. That way we get to test with different SDL versions. I'm on Arch which is a more "bleeding edge" distro so I get to test with the new SDL versions first.

Yeah, sure you can test with different SDL versions, and I don't want to prevent you, but that multiplies testing & maintenance effort. In my mind: 1 version of our software -> 1 set of library dependency versions. Not some moving targets.

My view of the world is that every app should have its own dependency versions and absolutely upgrade everything at their own pace, not mandated by the OS... unless it's something very fundamental that's built into the OS. I don't think SDL qualifies to be treated as an "OS-level component", it's just a wrapper around lower level APIs.

> Dynamic linking also makes it easier on distro maintainers because they have rules that say they are required to dynamically link everything.

Well, easier for them, but their rules suck then 😁 Here's a way how to suck less: http://web.archive.org/web/20090525150626/http://blog.garbe.us/2008/02/08/01_Static_linking/

(I actually quite like the suckless.org guys and their attitude towards things 😄 )

> My suspicion is that most of our Linux users probably get it elsewhere. Either directly from their distro, a community maintained thing like the Arch AUR or Ubuntu PPA, or from Flatpack or Snap.

Oh, the fragmentation! 😅 Don't even get me started... 😆 

Anyway, carry on, it's just me poking fun at desktop Linux 😄

rderooy:
From a distro maintainability perspective, dynamic linking makes perfect sense. Static linking is just simply unmaintainable. They have many thousands of packages, consider the mess if even 10% would use static linking, and there is a security update, as has happened with packages like openssl where packages bundled their own private version...

Issue #2667, opened 2023-07-08
SUBST must be an external command


NicknineTheEagle:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

None

### Is your feature request related to a problem? Please describe.

We have SUBST command from Dosbox SVN, however, it was implemented as an internal command which does not match MS-DOS where SUBST is an external command (subst.exe). For reference, internal commands are commands built into command interpreter while external commands are standalone programs in PATH. This potentially poses a problem because:

1. Internal commands do not have return codes.
2. Internal commands cannot be called from other applications.

### Describe the solution you'd like

Re-implement SUBST as a Z: drive program.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

PR #2691, opened 2023-07-18
Make MOVE and SUBST external commands

### LINKED ISSUE #2666: MOVE must be an external command

_No response_

None

_No response_

https://github.com/dosbox-staging/dosbox-staging/commit/98ae30850938d5b91dbeaa1cd52bdcfb7941a64d added MOVE command, however, it was implemented as an internal command which does not match MS-DOS 6.x where MOVE is an external command (move.exe). For reference, internal commands are commands built into command interpreter while external commands are standalone programs in PATH. This potentially poses a problem because:

1. Internal commands do not have return codes.
2. Internal commands cannot be called from other applications.

There was actually a real case where 2 caused an issue, there's a game that calls CHOICE which is bugged on Dosbox SVN because it implements CHOICE as an internal command, Staging has fixed that.

Re-implement MOVE as a Z: drive program.

_No response_

_No response_

- [X] Yes, I agree.

weirddan455:
Fixes #2666 #2667

As discussed, this is just re-using the existing code from `shell_cmds.cpp`

### DIFF
--- a/src/dos/meson.build
+++ b/src/dos/meson.build
@@ -44,6 +44,7 @@ libdos_sources = files(
     'program_rescan.cpp',
     'program_serial.cpp',
     'program_setver.cpp',
+    'program_subst.cpp',
     'program_tree.cpp',
 )
 
--- a/src/dos/meson.build
+++ b/src/dos/meson.build
@@ -39,6 +39,7 @@ libdos_sources = files(
     'program_mount.cpp',
     'program_mount_common.cpp',
     'program_mousectl.cpp',
+    'program_move.cpp',
     'program_placeholder.cpp',
     'program_rescan.cpp',
     'program_serial.cpp',
--- a/src/dos/program_move.cpp
+++ b/src/dos/program_move.cpp
@@ -1,0 +1,35 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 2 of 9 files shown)


src/dos/program_move.cpp
```diff
@@ -0,0 +1,35 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2020-2022  The DOSBox Staging Team
```
johnnovak, line 4: Update to 2023-2023

src/dos/program_move.h
```diff
@@ -0,0 +1,40 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2020-2022  The DOSBox Staging Team
```
johnnovak, line 4: Update to 2023-2023

src/dos/meson.build
```diff
@@ -39,10 +39,12 @@ libdos_sources = files(
     'program_mount.cpp',
     'program_mount_common.cpp',
     'program_mousectl.cpp',
+    'program_move.cpp',
     'program_placeholder.cpp',
     'program_rescan.cpp',
     'program_serial.cpp',
     'program_setver.cpp',
+    'program_subst.cpp',
```
johnnovak, line 47: You need to update the Visual Studio project files too:

https://github.com/dosbox-staging/dosbox-staging/blob/main/vs/dosbox.vcxproj
https://github.com/dosbox-staging/dosbox-staging/blob/main/vs/dosbox.vcxproj.filters

Once you've done that, the VS builds should start passing. The MSYS2 failure is a different issue, unrelated to your change, so please ignore that.

weirddan455: Those files are a mess.  I think I added it to the right place but lets see if the builds pass.  Is there a reason we can't just use Meson to generate Visual Studio files so we don't have to maintain these?

src/dos/program_move.cpp
weirddan455: Fixed

src/dos/program_move.h
weirddan455: Fixed

src/dos/meson.build
johnnovak: Dunno. It's not too hard to update them, and you don't need to do that very often.

If you can come up with a better solution I'm not against it, but you'll need to fully test it, make the CI loops pass, etc.

weirddan455: Fair enough.  I was just curious.  I'm probably not the right person to come up with a better solution.  I hate fussing with build systems... especially Windows build systems.

johnnovak: I'm the same, man 😄

johnnovak (approved):
Thanks @weirddan455 , approved & merging. As discussed with @shermp , the MSYS2 fiasco should not hold up merging PRs.

PR #2692, opened 2023-07-20
Hopefully make MSYS2 build pass

weirddan455:
I played around with this on my Windows partition for a while.  The problem seems to be this flag in fluidsynth that tries to link statically against Glib (dependency of Fluidsynth).

However, I just could not for the life of me get Glib to statically link in MSYS2.  I even tried a "hello world" Glib program, directly calling the compiler so as to bypass Meson and everything else in our build system.  It links just fine dynamically but as soon as I add the `-static` flag, it throws up undefined symbol errors.

It assume it's a regression in MSYS2 with some package updates (since none of the commits we've made recently should have broken this) although I wasn't able to find it.  I tried the previous version of Glib and the previous version of Meson but it still fails.

Anyway, hopefully this workaround is fine.  Did I mention I hate build systems?  :roll_eyes:

### DIFF
--- a/.github/workflows/windows-msys2.yml
+++ b/.github/workflows/windows-msys2.yml
@@ -104,6 +104,7 @@ jobs:
             mingw-w64-${{matrix.conf.toolchain}}${{matrix.conf.arch}}-zlib
             mingw-w64-${{matrix.conf.toolchain}}${{matrix.conf.arch}}-speexdsp
             mingw-w64-${{matrix.conf.toolchain}}${{matrix.conf.arch}}-python-pip
+            mingw-w64-${{matrix.conf.toolchain}}${{matrix.conf.arch}}-fluidsynth
 
       - name:  Prepare compiler cache
         id:    prep-ccache
@@ -229,6 +230,7 @@ jobs:
             mingw-w64-clang-${{matrix.conf.arch}}-zlib
             mingw-w64-clang-${{matrix.conf.arch}}-speexdsp
             mingw-w64-clang-${{matrix.conf.arch}}-python-pip
+            mingw-w64-clang-${{matrix.conf.arch}}-fluidsynth
 
       - name:  Prepare compiler cache
         id:    prep-ccache
@@ -294,10 +296,10 @@ jobs:
 
       - name: Setup release build
         run: |
-          meson setup -Ddefault_library=static -Db_lto=true -Db_lto_threads=$(nproc) $BUILD_RELEASE_DIR
+          meson setup -Db_lto=true -Db_lto_threads=$(nproc) $BUILD_RELEASE_DIR
 
       - name: Setup debugger build
-        run: meson setup -Ddefault_library=static -Db_lto=true -Db_lto_threads=$(nproc) -Denable_debugger=normal $BUILD_DEBUGGER_DIR
+        run: meson setup -Db_lto=true -Db_lto_threads=$(nproc) -Denable_debugger=normal $BUILD_DEBUGGER_DIR
 
       - name: Build release
         run: |


shermp:
I was looking at this too, and came to similar conclusions. I think MSYS2 never used to have a static glib2 available, but it does now.

I think the root cause is that somewhere along the line, the correct defines are not being set to disable the `__declspec(dllimport)` rubbish.

weirddan455:
Yeah, on my local system all the glib symbols were getting `__imp_` prefixed.

Anyway, I guess this isn't the fix.  One of the MSYS2 build is still failing and this broke on of the Linux and Mac builds too.  Sigh...

shermp:
Instead of changing the meson build file to do this, I would look at the msys2 CI file to make changes there on the command line.

shermp:
I ran out of time before going to work, but I was going to test what happens if `glib2` and `fluidsynth` packages are not installed in MSYS2, which would force the fluidsynth wrap to use it's own glib2 wrap.

shermp:
Sigh... looks like another msys2 package depends on glib2 (probably libslirp).

shermp:
Actually, it looks like the fluidsynth wrap doesn't declare a glib2 wrap anyway.

We may just have to switch MSYS2 to use shared libs for now :(

weirddan455:
libslirp does depend on glib.  I guess that one is being installed as an MSYS2 pacakge so my change didn't actually do anything (glib just gets pulled in as a libslirp anyway).

We have a glib.wrap file though in the subprojects folder.  Maybe we can tell Meson to use the glib wrap instead of looking for it on the system if we're on msys?  Not sure if that will actually solve the static linking problem but maybe worth a try.

johnnovak:
Thank you very much for looking into this @shermp and @weirddan455 🎉 Hopefully you're close to a solution!

> Anyway, hopefully this workaround is fine. Did I mention I hate build systems? 🙄

You're most definitely not alone 😏

weirddan455:
It's green finally!  Final solution was to first add fluidsynth to the MSYS2 install.  This solved the glib problem.  Then I just removed the `-Ddefault_library=static` flag.

I tried to see if I could get the static build to stay but it was just constant linker errors.  The first problem I ran into was the `__declspec(dllimport)` nonsense that fluidsynth was spitting out like @shermp pointed out.  That was a simple enough fix, just add a define of `FLUIDSYNTH_NOT_A_DLL` if it's being statically linked.

However, it then became apparent that the `default_library=static` flag is flawed, at least our implementation of it.  It doesn't affect all of our libraries.  Specifically, it still tries to dynamically link SDL... which is a dependency of fluidsynth... which tries to statically link SDL.  So you get multiple definition linker errors for SDL functions because it's being both statically and dynamically linked in the final executable.

Also, the Windows binaries we actually ship are dynamically linked and we just ship the .dlls along with it.  So I'm not sure why we need the CI to test a static build.

IMO if we really want a full static build, we need to build all dependencies from source and statically link everything.  MSYS2 is the one platform that actually ships .a libraries for statically linking.  On a real Linux distro, `-Ddefault_library=static` falls back to dynamic linking for most deps because they mostly don't ship static libraries.

johnnovak:
> It's green finally! Final solution was to first add fluidsynth to the MSYS2 install. This solved the glib problem. Then I just removed the `-Ddefault_library=static` flag.

Outstaind work @weirddan455 and much appreciated! Thanks for spending time on fixing this annoying and non-trivial issue! 🎉 

> Also, the Windows binaries we actually ship are dynamically linked and we just ship the .dlls along with it. So I'm not sure why we need the CI to test a static build.

I see that a quite pointless and unnecessary complication. We should only build and care for what we actually ship — life is complicated enough with those targets only, no need to _add more_! If you have the appetite for this, please feel free to remove the static build from the pipeline, unless @shermp vehemently disagrees 😅 

> IMO if we really want a full static build, we need to build all dependencies from source and statically link everything. MSYS2 is the one platform that actually ships .a libraries for statically linking. On a real Linux distro, `-Ddefault_library=static` falls back to dynamic linking for most deps because they mostly don't ship static libraries.

Yeah maybe let's not complicate life with all that just yet 😅 The Windows and macOS situation just works(tm), no need to fix what's not broken. And I'm not the best person to comment on the Linux situation anyway, but it has worked so far, so let's leave it as it is.

johnnovak (approved):

shermp:
I think the reason I was trying for static fluidsynth/slirp was so we didn't have to ship the bloated glib2 dll. But that was really the only reason.

PR #2693, opened 2023-07-21
Implement creating file with attributes in local FS

NicknineTheEagle:
This implements creating files with specific attributes in local FS. Known to be used by some games (e.g. Lemmings, floppy version, creates a hidden file named RUSSELL.DAT). I had to move away from fopen towards platform-specific functions.
I've admittedly not tested Unix version because I don't have the required environment so I'd appreciate if someone could take PR Linux/Mac build and see if it works.

### DIFF
--- a/include/cross.h
+++ b/include/cross.h
@@ -161,4 +161,13 @@ bool get_expanded_files(const std::string &path,
 
 std::string get_language_from_os();
 
+#ifndef WIN32
+#define XATTR_FATTR_NAME "user.dosbox_fattr"
+#define XATTR_FATTR_LEN 9 // 8 bits + null terminator
+
+uint16_t get_fat_attribs_from_xattr(const char *path);
+int set_xattr_from_fat_attribs(const char* path, uint16_t fattr);
+int set_xattr_from_fat_attribs(int fd, uint16_t fattr);
+#endif
+
 #endif
--- a/src/dos/drive_fat.cpp
+++ b/src/dos/drive_fat.cpp
@@ -1075,8 +1075,7 @@ bool fatDrive::FileCreate(DOS_File **file, char *name, uint16_t attributes) {
                                     fileEntry.loFirstClust,
                                     fileEntry.entrysize,
                                     this);
-	bool is_readonly     = fileEntry.attrib & DOS_ATTR_READ_ONLY;
-	fat_file->flags      = is_readonly ? OPEN_READ : OPEN_READWRITE;
+	fat_file->flags      = OPEN_READWRITE;
 	fat_file->dirCluster = dirClust;
 	fat_file->dirIndex   = subEntry;
 	fat_file->time       = fileEntry.modTime;
--- a/src/dos/drive_local.cpp
+++ b/src/dos/drive_local.cpp
@@ -28,6 +28,7 @@
 #include <climits>
 #include <cstdio>
 #include <ctime>
+#include <fcntl.h>
 #include <limits>
 #include <stdlib.h>
 #include <string.h>
@@ -46,32 +47,53 @@
 #include "cross.h"
 #include "inout.h"
 
-bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t /*attributes*/) {
-//TODO Maybe care for attributes but not likely
+bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t attributes) {
+	// Don't allow overwriting read-only files.
+	uint16_t test_attr = 0;
+	if (GetFileAttr(name, &test_attr) && (test_attr & DOS_ATTR_READ_ONLY)) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
+		return false;
+	}
+
+	bool existing_file = FileExists(name);
+
 	char newname[CROSS_LEN];
 	safe_strcpy(newname, basedir);
 	safe_strcat(newname, name);
 	CROSS_FILENAME(newname);
-	char* temp_name = dirCache.GetExpandName(newname); //Can only be used in till a new drive_cache action is preformed */
-	/* Test if file exists (so we need to truncate it). don't add to dirCache then */
-	bool existing_file = false;
-	
-	FILE * test = fopen_wrap(temp_name,"rb+");
-	if (test) {
-		fclose(test);
-		existing_file=true;
+	FILE* fp = nullptr;
 
+#ifdef WIN32
+	DWORD win_attribs = attributes ? (DWORD)attributes : FILE_ATTRIBUTE_NORMAL;
+	HANDLE win_fhandle = CreateFile(newname,
+	                                GENERIC_READ | GENERIC_WRITE,
+	                                0,
+	                                nullptr,
+	                                CREATE_ALWAYS,
+	                                win_attribs,
+	                                nullptr);
+	if (win_fhandle != INVALID_HANDLE_VALUE) {
+		int fd = _open_osfhandle((intptr_t)win_fhandle, _O_RDWR | _O_BINARY);
+		fp = _fdopen(fd, "wb+");
 	}
-	
-	FILE * hand = fopen_wrap(temp_name,"wb+");
-	if (!hand) {
+#else
+	constexpr int rw_perms = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+	int fd = open(newname, O_CREAT | O_RDWR | O_TRUNC, rw_perms);
+	if (fd != -1) {
+		set_xattr_from_fat_attribs(fd, attributes);
+		fp = fdopen(fd, "wb+");
+	}
+#endif
+
+	if (!fp) {
 		LOG_MSG("Warning: file creation failed: %s",newname);
+		DOS_SetError(DOSERR_ACCESS_DENIED);
 		return false;
 	}
    
 	if (!existing_file) dirCache.AddEntry(newname, true);
 	/* Make the 16 bit device information */
-	*file = new localFile(name, hand, basedir);
+	*file = new localFile(name, fp, basedir);
 	(*file)->flags=OPEN_READWRITE;
 
 	return true;
@@ -121,6 +143,15 @@ bool localDrive::FileOpen(DOS_File **file, char *name, uint32_t flags)
 		DOS_SetError(DOSERR_ACCESS_CODE_INVALID);
 		return false;
 	}
+
+	// Don't allow opening read-only files in write mode.
+	uint16_t test_attr = 0;
+	if (((flags & 0xf) == OPEN_WRITE || (flags & 0xf) == OPEN_READWRITE) &&
+	    (GetFileAttr(name, &test_attr) && (test_attr & DOS_ATTR_READ_ONLY))) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
+		return false;
+	}
+
 	char newname[CROSS_LEN];
 	safe_strcpy(newname, basedir);
 	safe_strcat(newname, name);
@@ -236,6 +267,13 @@ bool localDrive::FileUnlink(char * name) {
 		DEBUG_LOG_MSG("FS: Skipping removal of %s because it doesn't exist",
 		              name);
 		DOS_SetError(DOSERR_FILE_NOT_FOUND);
+		return false;
+	}
+
+	// Don't allow deleting read-only files.
+	uint16_t test_attr = 0;
+	if (GetFileAttr(name, &test_attr) && (test_attr & DOS_ATTR_READ_ONLY)) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
 		return false;
 	}
 
@@ -384,9 +422,8 @@ bool localDrive::FindNext(DOS_DTA & dta) {
 		if (!(find_attr & DOS_ATTR_DIRECTORY)) {
 			find_attr |= DOS_ATTR_ARCHIVE;
 		}
-		if (!(stat_block.st_mode & S_IWUSR)) {
-			find_attr |= DOS_ATTR_READ_ONLY;
-		}
+		
+		find_attr |= get_fat_attribs_from_xattr(temp_name);
 #endif
 		if (~srch_attr & find_attr &
 		    (DOS_ATTR_DIRECTORY | DOS_ATTR_HIDDEN | DOS_ATTR_SYSTEM)) {
@@ -438,11 +475,12 @@ bool localDrive::GetFileAttr(char *name, uint16_t *attr)
 #else
 	struct stat status;
 	if (stat(newname, &status) == 0) {
-		*attr = status.st_mode & S_IFDIR ? 0 : DOS_ATTR_ARCHIVE;
-		if (status.st_mode & S_IFDIR)
-			*attr |= DOS_ATTR_DIRECTORY;
-		if (!(status.st_mode & S_IWUSR))
-			*attr |= DOS_ATTR_READ_ONLY;
+		// HACK: Always report Archive attribute on Unix since managing it
+		// manually would be a pain.
+		*attr = status.st_mode & S_IFDIR ? DOS_ATTR_DIRECTORY
+		                                 : DOS_ATTR_ARCHIVE;
+
+		*attr |= get_fat_attribs_from_xattr(newname);
 		return true;
 	}
 	*attr = 0;
@@ -471,14 +509,7 @@ bool localDrive::SetFileAttr(const char *name, const uint16_t attr)
 		return false;
 	}
 
-	if (attr & (DOS_ATTR_SYSTEM | DOS_ATTR_HIDDEN))
-		LOG_WARNING("FILESYSTEM: Application attempted to set system or hidden"
-		            " attributes for '%s', which is ignored for local drives",
-		            newname);
-
-	const auto result = attr & DOS_ATTR_READ_ONLY ? make_readonly(f)
-	                                              : make_writable(f);
-	if (!result) {
+	if (set_xattr_from_fat_attribs(newname, attr) == -1) {
 		DOS_SetError(DOSERR_ACCESS_DENIED);
 		return false;
 	}
--- a/src/misc/cross.cpp
+++ b/src/misc/cross.cpp
@@ -44,6 +44,7 @@
 #	include <shlobj.h>
 #else
 #	include <libgen.h>
+#	include <sys/xattr.h>
 #endif
 
 #if defined HAVE_PWD_H
@@ -781,6 +782,91 @@ std::string cfstr_to_string(CFStringRef source)
 	assert(lang.empty());
 	return lang;
 }
+
+#ifndef WIN32
+uint16_t xattr_string_to_fattr(const char* xattr)
+{
+	uint16_t fattr = 0;
+
+	if (strchr(xattr, 'R')) {
+		fattr |= DOS_ATTR_READ_ONLY;
+	}
+	if (strchr(xattr, 'H')) {
+		fattr |= DOS_ATTR_HIDDEN;
+	}
+	if (strchr(xattr, 'S')) {
+		fattr |= DOS_ATTR_SYSTEM;
+	}
+	if (strchr(xattr, 'A')) {
+		fattr |= DOS_ATTR_ARCHIVE;
+	}
+
+	return fattr;
+}
+
+void fattr_to_xattr_string(uint16_t fattr, char* outbuf, size_t outsize)
+{
+	memset(outbuf, 0, outsize);
+
+	if (fattr & DOS_ATTR_READ_ONLY) {
+		strcat(outbuf, "R");
+	}
+	if (fattr & DOS_ATTR_HIDDEN) {
+		strcat(outbuf, "H");
+	}
+	if (fattr & DOS_ATTR_SYSTEM) {
+		strcat(outbuf, "S");
+	}
+	if (fattr & DOS_ATTR_ARCHIVE) {
+		strcat(outbuf, "A");
+	}
+}
+
+uint16_t get_fat_attribs_from_xattr(const char *path)
+{
+	char xattr[XATTR_FATTR_LEN];
+#ifdef MACOSX
+	int read_len = getxattr(path, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN - 1, 0, 0);
+#else
+	int read_len = getxattr(path, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN - 1);
+#endif
+	if (read_len > 0) {
+		// xattr strings are not null terminated.
+		xattr[read_len] = '\0';
+		return xattr_string_to_fattr(xattr);
+	}
+
+	return 0;
+}
+
+int set_xattr_from_fat_attribs(const char *path, uint16_t fattr)
+{
+	char xattr[XATTR_FATTR_LEN];
+	fattr_to_xattr_string(fattr, xattr, XATTR_FATTR_LEN);
+	size_t xattr_len = strlen(xattr);
+#ifdef MACOSX
+	int result = setxattr(path, XATTR_FATTR_NAME, xattr, xattr_len, 0, 0);
+#else
+	int result = setxattr(path, XATTR_FATTR_NAME, xattr, xattr_len, 0);
+#endif
+
+	return result;
+}
+
+int set_xattr_from_fat_attribs(int fd, uint16_t fattr)
+{
+	char xattr[XATTR_FATTR_LEN];
+	fattr_to_xattr_string(fattr, xattr, XATTR_FATTR_LEN);
+	size_t xattr_len = strlen(xattr);
+#ifdef MACOSX
+	int result = fsetxattr(fd, XATTR_FATTR_NAME, xattr, xattr_len, 0, 0);
+#else
+	int result = fsetxattr(fd, XATTR_FATTR_NAME, xattr, xattr_len, 0);
+#endif
+
+	return result;
+}
+#endif
 
 namespace cross {
 


src/dos/drive_local.cpp
```diff
@@ -56,22 +56,43 @@ bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t /*attributes*
...
+	                                nullptr);
+	if (win_fhandle != INVALID_HANDLE_VALUE) {
+		int fd = _open_osfhandle((intptr_t)win_fhandle, _O_RDWR | _O_BINARY);
+		fp = _fdopen(fd, "wb+");
 	}
-	
-	FILE * hand = fopen_wrap(temp_name,"wb+");
-	if (!hand) {
+#else
+	int unix_perms = (attributes & DOS_ATTR_READ_ONLY) ? 0444 : 0666;
```
NicknineTheEagle, line 81: PVS Analyzer shows warning V536 here even though it's not supposed to do that for multiple octal constants in close proximity. Don't know what's up with that.

src/dos/drive_local.cpp
```diff
@@ -56,22 +56,50 @@ bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t /*attributes*
...
+	int fd         = open(newname, O_CREAT | O_RDWR | O_TRUNC, unix_perms);
+	if (fd != -1) {
+		fp = fdopen(fd, "wb+");
+	}
+#endif
+
+	if (!fp) {
 		LOG_MSG("Warning: file creation failed: %s",newname);
+		DOS_SetError(ERROR_ACCESS_DENIED);
```
weirddan455, line 96: You've got a compile error here:  `../src/dos/drive_local.cpp:96:16: error: ‘ERROR_ACCESS_DENIED’ was not declared in this scope; did you mean ‘DOSERR_ACCESS_DENIED’?`

NicknineTheEagle: Whoops.

FeralChild64:
I confirm this works under Linux - test application from the Discord created file with the following attributes:
attributes

Without the patch the file was not read-only.

src/dos/drive_local.cpp
FeralChild64: There are constants for this: https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html

But I'm not convinced omitting `fopen_wrap` here is safe - wouldn't it be better to continue using it, but to add a parameter specifying DOS attributes and handle them inside? AFAIK that's what the `cross.cpp` was created for.

NicknineTheEagle:
> I confirm this works under Linux - test application from the Discord created file with the following attributes: attributes
> 
> Without the patch the file was not read-only.

Good, looks like writing to the file before closing the handle succeeds as well, same as on Windows.

src/dos/drive_local.cpp
NicknineTheEagle: fopen_wrap is used in a whole bunch of places, not sure about moving all this stuff there. Plus it takes standard fopen mode parameter (rb, wb, etc) which would somehow need to be translated into custom parameters used by each platform.

FeralChild64: Yes, but it seems on some system `fopen_wrap` offers protection against accessing `/proc` filesystem - I'm not sure, why, but think this should better be preserved.
Moreover, since `cross.cpp` contains various filesystem abstraction routines, this code should go to `cross.cpp` too.

NicknineTheEagle: Again, moving this into fopen_wrap is major scope change that would affect a lot more than this PR intends to cover. And there's still an issue of different open modes, translating them would be a headache.

johnnovak:
Ok, so I see no difference in file attribute handling behaviour between this branch and `main` on macOS. Your comment confuses me @FeralChild64 

> Without the patch the file was not read-only.
 
Toggling the read-only attribute has been working fine for me on `main` for along time, at least on macOS. I started setting the read-only flag on some files months ago and the flag was correctly saved; I even remarked that it's nice that this just works.

@NicknineTheEagle 's description implies that this change is needed for Windows only. We should make that clear in the ticket's title and the commit messages. Like I said, I'm quite sure this has been working on Linux & macOS for a long time.

For the record, I'm getting these results with current `main` on macOS, and the results are exactly the same with this branch:

image

image

---

Then when I remove the read-only flag:

image

image


---

If I try to set the hidden or system flags, I'm getting this log warning, which is fine and expected:

image

src/dos/drive_local.cpp
```diff
@@ -56,22 +56,50 @@ bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t /*attributes*
...
+	HANDLE win_fhandle = CreateFile(newname,
+	                                GENERIC_READ | GENERIC_WRITE,
+	                                0,
+	                                nullptr,
+	                                CREATE_ALWAYS,
+	                                win_attribs,
+	                                nullptr);
+	if (win_fhandle != INVALID_HANDLE_VALUE) {
+		int fd = _open_osfhandle((intptr_t)win_fhandle, _O_RDWR | _O_BINARY);
+		fp = _fdopen(fd, "wb+");
```
johnnovak, line 78: Does this mean the Hidden, System and Archive attributes are supported now on Windows using NTFS? If so, I actually have quite mixed feelings about supporting all attributes on Windows, but only a subset on *nixes. 

I'd much prefer a consistent cross-platform behaviour and go with the least common denominator, otherwise we'll end up with scenarios where some programs works on Windows only, or has subtle bugs on other OSes than Windows.

FeralChild64:
@johnnovak 

> Ok, so I see no difference in file attribute handling behaviour between this branch and main on macOS

Get the TEST.EXE file shared by @NicknineTheEagle on Discord, dosbox-staging-dev channel. It creates a file - after the PR it has same attributes as it would have on MS-DOS, without the PR it misses the read-only attribute.

johnnovak:
> @johnnovak
> 
> > Ok, so I see no difference in file attribute handling behaviour between this branch and main on macOS
> 
> Get the TEST.EXE file shared by @NicknineTheEagle on Discord, dosbox-staging-dev channel. It creates a file - after the PR it has same attributes as it would have on MS-DOS, without the PR it misses the read-only attribute.

So how come it works when I toggle the attributes in DOS Navigator then? It should make no difference.

FeralChild64:
Toggling attributes in DOS Na igator shouldn’t change. It’s the process of creating the file which is affected - setting the attributes during file creation didn’t work properly.

johnnovak:
> Toggling attributes in DOS Na igator shouldn’t change. It’s the process of creating the file which is affected - setting the attributes during file creation didn’t work properly.

Thanks, kinda missed that part that this in only about creation.

Ok @NicknineTheEagle, so this is fine then, but make the feature behave consistently across platforms, meaning don't allow setting Archive, Hidden and System on Windows/NTFS only but not on *nixes as per my earlier comment (if I understood that logic right).

Another minor thing about the wording of the commit messages. Instead of this:

- Implement creating file with attributes in local FS
- Fix r/o attribute handling in FAT FS

I would prefer this:

- Implement creating files with attributes on the local filesystem
- Fix read-only attribute handling on FAT filesystem

Mainly changing to "FS" to "filesystem" and "r/o" to "read-only" would make these commits searchable in the Git history. I would never think of searching for "r/o" and "FS"; I would search for "readonly" or "read-only" and "filesystem". Thanks 😄

src/dos/drive_local.cpp
NicknineTheEagle: The current behavior is consistent with `localDrive::SetFileAttr`. The reason Hidden and System are not supported on Linux is simply because they _do not exist_ there. Linux does have a hidden files feature but it's used by prefixing filename with a dot.

rderooy:
I think it should be possible on Linux to support the DOS attributes, if the backing filesystem supports them (e.g. FAT or NTFS). Which to be honest, would probably not often be the case. And could perhaps even cause more confusion for users. e.g., why does it work on this Linux system and not this other....

johnnovak:
> And could perhaps even cause more confusion for users. e.g., why does it work on this Linux system and not this other....

Yes, exactly my point.

DOSBox Staging should provide a _consistent_ user experience on all three supported platforms: Windows, macOS, and Linux. Consistent cross-platform behaviour is our default and a _must_;  we can only deviate from that for non-essential enhancements, and even that very sparingly. This came up a few times before and kcgen has made our position on this very clear, and I 100% agree with that. 

One such niche enhancement is the support for CD Audio when using physical media. We previously only supported this on Windows, then @weirddan455 has recently added Linux support. That's fine, it's a niche feature not used by many people (most people use CD images anyway), and having this _enhancement_ on Windows on Linux does not cause serious maintenance or troubleshooting problems, and doesn't fundamentally alter the consistent cross-platform experience of core features (it's pretty obvious that if CD Audio doesn't work with physical media on macOS that is because it's only supported on Windows and Linux).

However, the DOS-to-host-filesystem mapping is a fundamental core feature where we do _not_ want to introduce _any_ cross-platform differences. The filesystem mapping must be 100% consistent across platforms, full stop. That's what we have currently, and this change would start us on a path diverging from this model. We do not want that. This would create the possibility of situations where you ZIP up your C drive folder that works with Staging on Windows, but it would fail on Linux or macOS because of some file attribute differences. 

Addin support for the read-only attribute on file creation is a welcome improvement @NicknineTheEagle , thanks for that. But the support for the Archive, Hidden and System attributes on Windows has to go from the current PR in the light of the project's goals. However, if you can find a way to add support for these flags across all three platforms in a maintainable way, by all means, do so in a future PR!

johnnovak (changes requested):
As per my previous comments:

- Remove non-cross-platform features
- Improve commit messages

NicknineTheEagle:
@johnnovak
> This would create the possibility of situations where you ZIP up your C drive folder that works with Staging on Windows, but it would fail on Linux or macOS because of some file attribute differences.

Such situation would occur either way because Read Only attribute technically does not exist on Linux either. Staging implements it with a hack using user permissions.

NicknineTheEagle:
I frankly do not like this line of though, you're basically saying "we cannot fix it for this minority of users so let's break it for everyone just for the sake of consistency".

weirddan455:
You could technically implement setting these extra attributes on Linux and Mac with "Extended file attributes".  They allow setting any arbitrary string.  It won't do anything in the host OS but it would allow these attributes to persist in DOSBox.  https://en.wikipedia.org/wiki/Extended_file_attributes

NicknineTheEagle:
> You could technically implement setting these extra attributes on Linux and Mac with "Extended file attributes". They allow setting any arbitrary string. It won't do anything in the host OS but it would allow these attributes to persist in DOSBox. https://en.wikipedia.org/wiki/Extended_file_attributes

Ooh, I will look into it ASAP.

johnnovak:
> @johnnovak
> 
> > This would create the possibility of situations where you ZIP up your C drive folder that works with Staging on Windows, but it would fail on Linux or macOS because of some file attribute differences.
> 
> Such situation would occur with Read Only as well because Read Only attribute technically does not exist on Linux either. Staging implements it on Linux with a hack using user permissions.

johnnovak:
Sorry, closed by mistake! Wasn't intentional 😅

johnnovak:
> I frankly do not like this line of thought, you're basically saying "we cannot fix it for this minority of users so let's break it for everyone just for the sake of consistency".

In a way, yes, that's what consistent cross-platform behaviour means if there is no other solution. I don't like it either but that's life 🤷🏻

johnnovak:
> > You could technically implement setting these extra attributes on Linux and Mac with "Extended file attributes". They allow setting any arbitrary string. It won't do anything in the host OS but it would allow these attributes to persist in DOSBox. https://en.wikipedia.org/wiki/Extended_file_attributes
> 
> Ooh, I will look into it ASAP.

Yes that would be a good solution if you can make it work.

Indeed, my ZIP example was bad as attributes won't carry over between Windows and *nix. So let's soften that a bit so if the software behaves the same way on all three platforms when installed from scratch when it comes to attributes and when executing the same steps, then that's OK.

johnnovak:
Btw, you brought up an interesting issue with ZIP. I think if we used PKZIP inside DOSBox to create and extract the ZIP files, then the attribs would carry over different platforms. That's actually quite nice!

NicknineTheEagle:
> Btw, you brought up an interesting issue with ZIP. I think if we used PKZIP inside DOSBox to create and extract the ZIP files, then the attribs would carry over different platforms if you got extended attribs working on Linux and macOS. That's actually quite nice!

Yeah, I thought about, too, quite convenient.

NicknineTheEagle:
@FeralChild64 @johnnovak Alright, try this. If I did everything correctly, ATTRIB should now be fully functional on Linux and Mac, making use of xattr to store FAT attributes.

src/misc/cross.cpp
```diff
@@ -782,6 +782,46 @@ std::string cfstr_to_string(CFStringRef source)
 	return lang;
 }
 
+void fat_attributes_to_xattr(uint16_t fattr, char* outbuf, size_t outsize)
+{
+	memset(outbuf, 0, outsize);
+
+	strcat(outbuf, "\"");
```
weirddan455, line 789: Quotes aren't needed.  Sorry about the confusion on Discord.  The CLI tool will show it as a string if the value contains only ASCII data.

src/misc/cross.cpp
```diff
@@ -782,6 +782,46 @@ std::string cfstr_to_string(CFStringRef source)
...
+	if (fattr & DOS_ATTR_HIDDEN) {
+		strcat(outbuf, "H");
+	}
+	if (fattr & DOS_ATTR_SYSTEM) {
+		strcat(outbuf, "S");
+	}
+	if (fattr & DOS_ATTR_ARCHIVE) {
+		strcat(outbuf, "A");
+	}
+	strcat(outbuf, "\"");
```
weirddan455, line 802: Remove quotes

src/dos/drive_local.cpp
```diff
@@ -443,6 +480,18 @@ bool localDrive::GetFileAttr(char *name, uint16_t *attr)
...
 			*attr |= DOS_ATTR_READ_ONLY;
+
+		char xattr[XATTR_FATTR_LEN] = "";
+#ifdef MACOSX
+		int read_len = getxattr(newname, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN, 0, 0);
+#else
+		int read_len = getxattr(newname, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN);
+#endif
+		if (read_len > 0) {
+			xattr[read_len - 1] = '\0';
```
weirddan455, line 491: This is off by one, bug hidden because of the quotes.  Should be `xattr[read_len] = '\0` or you over-write the last character.

src/dos/drive_local.cpp
```diff
@@ -443,6 +480,18 @@ bool localDrive::GetFileAttr(char *name, uint16_t *attr)
 			*attr |= DOS_ATTR_DIRECTORY;
 		if (!(status.st_mode & S_IWUSR))
 			*attr |= DOS_ATTR_READ_ONLY;
+
+		char xattr[XATTR_FATTR_LEN] = "";
+#ifdef MACOSX
+		int read_len = getxattr(newname, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN, 0, 0);
+#else
+		int read_len = getxattr(newname, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN);
```
weirddan455, line 488: This should be `XATTR_FATTR_LEN - 1` so you leave room to add a null terminator.

src/dos/drive_local.cpp
```diff
@@ -443,6 +480,18 @@ bool localDrive::GetFileAttr(char *name, uint16_t *attr)
 			*attr |= DOS_ATTR_DIRECTORY;
 		if (!(status.st_mode & S_IWUSR))
 			*attr |= DOS_ATTR_READ_ONLY;
+
+		char xattr[XATTR_FATTR_LEN] = "";
+#ifdef MACOSX
+		int read_len = getxattr(newname, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN, 0, 0);
```
weirddan455, line 486: This should be XATTR_FATTR_LEN - 1 so you leave room to add a null terminator.

weirddan455 (changes requested):

src/dos/drive_local.cpp
```diff
@@ -46,32 +47,53 @@
...
+	                                nullptr);
+	if (win_fhandle != INVALID_HANDLE_VALUE) {
+		int fd = _open_osfhandle((intptr_t)win_fhandle, _O_RDWR | _O_BINARY);
+		fp = _fdopen(fd, "wb+");
 	}
-	
-	FILE * hand = fopen_wrap(temp_name,"wb+");
-	if (!hand) {
+#else
+	constexpr int rw_perms = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
```
weirddan455, line 80: 0644 is the default for new files on Linux.  So no write permissions for group or other.  I would leave it at that unless you have a good reason not to.

FeralChild64:
@NicknineTheEagle I hope I’ll have time to look at it tomorrow - but already one remark: there might be some standard for such extended attributes, as more application might want to store them (DOSBox-X, Wine, Samba); if so, it might be a good idea to follow it. See for example here: https://www.winehq.org/pipermail/wine-patches/2009-September/078227.html

src/dos/drive_local.cpp
NicknineTheEagle: Actually, when you create a file on Linux, the process's current umask is subtracted from permission bits and umask is usually 022 by default. So it's perfectly fine to pass 0666.
https://man7.org/linux/man-pages/man2/umask.2.html

weirddan455: I stand corrected.  This is fine then.

src/dos/drive_fat.cpp
```diff
@@ -1075,8 +1075,7 @@ bool fatDrive::FileCreate(DOS_File **file, char *name, uint16_t attributes) {
                                     fileEntry.loFirstClust,
                                     fileEntry.entrysize,
                                     this);
-	bool is_readonly     = fileEntry.attrib & DOS_ATTR_READ_ONLY;
-	fat_file->flags      = is_readonly ? OPEN_READ : OPEN_READWRITE;
+	fat_file->flags      = OPEN_READWRITE;
```
weirddan455, line 1078: What is the purpose of this change?

NicknineTheEagle: I was under false impression that creating a read-only file on DOS returns a handle that you can't write to but testing showed that I was wrong. You *can* write to a freshly created r/o file until you close the handle. I.e. r/o attribute merely blocks you from opening new r/w streams, it doesn't block existing r/w streams from working.

NicknineTheEagle:
> @NicknineTheEagle I hope I’ll have time to look at it tomorrow - but already one remark: there might be some standard for such extended attributes, as more application might want to store them (DOSBox-X, Wine, Samba); if so, it might be a good idea to follow it. See for example here: https://www.winehq.org/pipermail/wine-patches/2009-September/078227.html

Good thinking, I'll look into that as well.

weirddan455 (approved):
This all looks fine to me.  I tested on Linux with the ATTRIB command and everything seemed to work.  I assume you've tested the WIN32 stuff.

src/dos/drive_fat.cpp
weirddan455: Ok, makes sense.

NicknineTheEagle:
According to that Wine link, Wine uses "user.DOSATTRIB" xattrib with hex number as the value which also matches the standard used by Samba 3.

johnnovak:
Great stuff, I'll give this a good testing today on macOS and Win, and maybe even RPi4 running Linux. I'll try using ZIP/RAR/ARJ inside DOSBox too to see if the attribs carry over between platforms (they should, but I wanna see 😄).

johnnovak:
Ok, no luck on macOS. This needs further work.

image

NicknineTheEagle:
> Ok, no luck on macOS. This needs further work. I won't comment on the PR until we get into a working state on all 3 platforms.
> image

Hmm, can you try xattr CLI utility on your Mac?
https://ss64.com/osx/xattr.html

johnnovak:
Ok, with another file it works:

image

The problem was that I tried first on your `TEST.EXE` that I downloaded via Firefox, and apparently that set some "quarantine" xattr flags on it by default, making all subsequent xattr operations fail... some weird macOS security stuff 🤷🏻 

image

I will make comments on the PR later today as some things need changing at the code level, but functionality wise seems fine so far 👍🏻

johnnovak:
Okay, found one bug. If you made a file read-only prior to this change, it's impossible to change its attributes anymore (for example, to remove the read-only flag) because that changed the file permissions on the host OS level to non-writeable. Then you just get a log message like that:

image

The only way to resolve the situation is to make the file writeable again outside of DOSBox.

We want a clear, seamless upgrade path for existing installations, so when setting attributes of existing files, we need to check if the file is read-only on the host OS level, and if so, we must make it writeable again. 

Plus if we fail to make the file writeable (it's possible), it should be logged as an error, not just an info level log message.

include/cross.h
```diff
@@ -161,4 +161,13 @@ bool get_expanded_files(const std::string &path,
 
 std::string get_language_from_os();
 
+#ifndef WIN32
+#define XATTR_FATTR_NAME "user.dosbox_fattr"
```
johnnovak, line 165: - Please only use `const` or `constexpr`.
- `const char*` is fine for string constants, otherwise default to `std::string`.
- Please use UpperCamelCase for new constants and convert old constants if you touch related code.
- This name is too cryptic, this would be better:

```cpp
constexrp char DosAttrExtendedAttributeName[] = "user.dosbox_fattr"
```

include/cross.h
```diff
@@ -161,4 +161,13 @@ bool get_expanded_files(const std::string &path,
...
+#define XATTR_FATTR_LEN 9 // 8 bits + null terminator
```
johnnovak, line 166:
```cpp
constexpr auto DosAttrExtendedAttributeLength = 9
```

But it would be better to use `std::string` (more comments on that below).

include/cross.h
```diff
@@ -161,4 +161,13 @@ bool get_expanded_files(const std::string &path,
...
+
+uint16_t get_fat_attribs_from_xattr(const char *path);
+int set_xattr_from_fat_attribs(const char* path, uint16_t fattr);
```
johnnovak, line 169: Let's use `dos_attrib`; that's the least ambiguous name (better than `fattr` which is very cryptic or `fat_attrib` even).

Make args const whenever possible:

`const uint16_r dos_attrib`

johnnovak: I'd suggest these name:

` get_dos_attribs_from_extended_attrib`
` set_extended_attrib_from_dos_attribs`

But at the very least rename `fat_attrib` to `dos_attrib`.

src/dos/drive_local.cpp
```diff
@@ -46,32 +47,53 @@
 #include "cross.h"
 #include "inout.h"
 
-bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t /*attributes*/) {
-//TODO Maybe care for attributes but not likely
+bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t attributes) {
```
johnnovak, line 50: Formatting issues; run all your new code through Clang format, please.

johnnovak: `const uint16_t attributes`

src/dos/drive_local.cpp
```diff
@@ -46,32 +47,53 @@
...
+	// Don't allow overwriting read-only files.
+	uint16_t test_attr = 0;
+	if (GetFileAttr(name, &test_attr) && (test_attr & DOS_ATTR_READ_ONLY)) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
+		return false;
+	}
+
+	bool existing_file = FileExists(name);
```
johnnovak, line 58: `bool file_exists` reads better

include/cross.h
```diff
@@ -161,4 +161,13 @@ bool get_expanded_files(const std::string &path,
...
+int set_xattr_from_fat_attribs(int fd, uint16_t fattr);
```
johnnovak, line 170: const uint16_t fat_attribs

src/dos/drive_local.cpp
johnnovak: Going forward, we should use the nice `FatAttributeFlags` union that kcgen introduced. @weirddan455 has made good use of it in his MOVE command implementation, you can refer that for example usages.

src/dos/drive_local.cpp
```diff
@@ -46,32 +47,53 @@
 #include "cross.h"
 #include "inout.h"
 
-bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t /*attributes*/) {
-//TODO Maybe care for attributes but not likely
+bool localDrive::FileCreate(DOS_File * * file,char * name,uint16_t attributes) {
+	// Don't allow overwriting read-only files.
+	uint16_t test_attr = 0;
+	if (GetFileAttr(name, &test_attr) && (test_attr & DOS_ATTR_READ_ONLY)) {
```
johnnovak, line 53: Going forward, we should use the nice `FatAttributeFlags` kcgen introduced. @weirddan455 has made good use of it in his MOVE command implementation, you can refer to that for example usages.

src/dos/drive_local.cpp
```diff
@@ -121,6 +143,15 @@ bool localDrive::FileOpen(DOS_File **file, char *name, uint32_t flags)
 		DOS_SetError(DOSERR_ACCESS_CODE_INVALID);
 		return false;
 	}
+
+	// Don't allow opening read-only files in write mode.
+	uint16_t test_attr = 0;
+	if (((flags & 0xf) == OPEN_WRITE || (flags & 0xf) == OPEN_READWRITE) &&
+	    (GetFileAttr(name, &test_attr) && (test_attr & DOS_ATTR_READ_ONLY))) {
```
johnnovak, line 150: Sometimes low-level bit-twiddling is necessary, but whenever we can, we should use the much nicer `bit_view` unions that lead to more readable code, and thus hopefully less obscure bugs. This is a prime cadidate to be restructured to use the `FatAttributeFlags` union instead.

johnnovak: At the very least extract `flags & 0xf` into a temp var with a descriptive name. I have no idea what `0xf` does just by looking at this.

src/dos/drive_local.cpp
```diff
@@ -438,11 +475,12 @@ bool localDrive::GetFileAttr(char *name, uint16_t *attr)
 #else
 	struct stat status;
 	if (stat(newname, &status) == 0) {
-		*attr = status.st_mode & S_IFDIR ? 0 : DOS_ATTR_ARCHIVE;
-		if (status.st_mode & S_IFDIR)
-			*attr |= DOS_ATTR_DIRECTORY;
-		if (!(status.st_mode & S_IWUSR))
-			*attr |= DOS_ATTR_READ_ONLY;
+		// HACK: Always report Archive attribute on Unix since managing it
+		// manually would be a pain.
```
johnnovak, line 479: How so? We should add support for all attributes, including archive. How is this more difficult than supporting the rest of the FAT attributes?

src/dos/drive_local.cpp
```diff
@@ -438,11 +475,12 @@ bool localDrive::GetFileAttr(char *name, uint16_t *attr)
...
+		*attr = status.st_mode & S_IFDIR ? DOS_ATTR_DIRECTORY
+		                                 : DOS_ATTR_ARCHIVE;
+
+		*attr |= get_fat_attribs_from_xattr(newname);
```
johnnovak, line 483: So all `DOS_ATTR_*` stuff should use the `FatAttributeFlags` union instead.

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+	if (fattr & DOS_ATTR_HIDDEN) {
+		strcat(outbuf, "H");
+	}
+	if (fattr & DOS_ATTR_SYSTEM) {
+		strcat(outbuf, "S");
+	}
+	if (fattr & DOS_ATTR_ARCHIVE) {
+		strcat(outbuf, "A");
+	}
+}
```
johnnovak, line 823: I don't see much value in doing things this way when a much simpler solution would suffice. The FAT16 filesystem is a fixed entity, we don't need to design this feature for extendability. I would suggest using a 4-character long zero-terminated string, so 5 bytes in total with the following fixed layout:

- no flags set: `----`
- all flags set: `ARHS`
- read-only and system set: `-R-S`

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+
+uint16_t get_fat_attribs_from_xattr(const char *path)
+{
+	char xattr[XATTR_FATTR_LEN];
+#ifdef MACOSX
+	int read_len = getxattr(path, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN - 1, 0, 0);
+#else
+	int read_len = getxattr(path, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN - 1);
```
johnnovak, line 831: So this is confusing. With my proposed fixed 4-char schema, `FatAttribsExtendedAttributeLength` would be 4, as this is a 4-char long string. Then you would _add_ space for the extra terminating zero by adding `+ 1`; that's much clearer.

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+	if (fattr & DOS_ATTR_ARCHIVE) {
+		strcat(outbuf, "A");
+	}
+}
+
+uint16_t get_fat_attribs_from_xattr(const char *path)
+{
+	char xattr[XATTR_FATTR_LEN];
+#ifdef MACOSX
+	int read_len = getxattr(path, XATTR_FATTR_NAME, xattr, XATTR_FATTR_LEN - 1, 0, 0);
```
johnnovak, line 829: What are the two zero args at the end? Highly suggest extracting them to meaningfully named constants. We started doing this and it greatly enhances the readability of the code. In this case something like this would be better:

```cpp
constexpr uint32_t position = 0;
constexpr int options       = 0;

const auto num_bytes_read = getxattr(path, FatAttribsExtendedAttributeName,
                                     xattr, FatAttribsExtendedAttributeLength,
                                     position, options);
```

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+		strcat(outbuf, "S");
+	}
+	if (fattr & DOS_ATTR_ARCHIVE) {
+		strcat(outbuf, "A");
+	}
+}
+
+uint16_t get_fat_attribs_from_xattr(const char *path)
+{
+	char xattr[XATTR_FATTR_LEN];
```
johnnovak, line 827: Initialise buffer with `{}` then you don't need to manually set the null byte at the end of the read value. We should initialise literally _everything_; not initialising variables/members due to performance reasons should be a very rare exception rather than the norm.

```cpp
char xattr[XATTR_FATTR_LEN] = {};
```

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+		xattr[read_len] = '\0';
+		return xattr_string_to_fattr(xattr);
+	}
+
+	return 0;
+}
+
+int set_xattr_from_fat_attribs(const char *path, uint16_t fattr)
+{
+	char xattr[XATTR_FATTR_LEN];
```
johnnovak, line 844: Init with `{}`.

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+	if (read_len > 0) {
+		// xattr strings are not null terminated.
+		xattr[read_len] = '\0';
+		return xattr_string_to_fattr(xattr);
+	}
+
+	return 0;
+}
+
+int set_xattr_from_fat_attribs(const char *path, uint16_t fattr)
```
johnnovak, line 842: `fat_attribs`

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+		fattr |= DOS_ATTR_SYSTEM;
+	}
+	if (strchr(xattr, 'A')) {
+		fattr |= DOS_ATTR_ARCHIVE;
+	}
+
+	return fattr;
+}
+
+void fattr_to_xattr_string(uint16_t fattr, char* outbuf, size_t outsize)
```
johnnovak, line 807: Requiring the user of the function to pass in `outsize` is very oldschool; we don't do that. Please use `std::string`. This is not performance intensive code, and even if it was, we can still fully take advantage of the safe and easy to use `std::string` interface and use string views.

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+	}
+	if (fattr & DOS_ATTR_SYSTEM) {
+		strcat(outbuf, "S");
+	}
+	if (fattr & DOS_ATTR_ARCHIVE) {
+		strcat(outbuf, "A");
+	}
+}
+
+uint16_t get_fat_attribs_from_xattr(const char *path)
```
johnnovak, line 825: Ideally, you'd use `std_fs::path` and only covert at the "edges", i.e. when you must convert it to a C string so you can pass it to an OS-level C function.

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+#ifdef MACOSX
+	int result = setxattr(path, XATTR_FATTR_NAME, xattr, xattr_len, 0, 0);
+#else
+	int result = setxattr(path, XATTR_FATTR_NAME, xattr, xattr_len, 0);
+#endif
+
+	return result;
+}
+
+int set_xattr_from_fat_attribs(int fd, uint16_t fattr)
```
johnnovak, line 856: `const uint16_t fat_attribs`

johnnovak: Returning a `bool` for success would be much better, especially if this a platform-agnostic abstraction, so error code would not even be consistent across platforms, I assume.

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+{
+	char xattr[XATTR_FATTR_LEN];
```
johnnovak, line 858: Init with `{}`

src/misc/cross.cpp
```diff
@@ -782,6 +783,91 @@ std::string cfstr_to_string(CFStringRef source)
...
+	fattr_to_xattr_string(fattr, xattr, XATTR_FATTR_LEN);
+	size_t xattr_len = strlen(xattr);
+#ifdef MACOSX
+	int result = fsetxattr(fd, XATTR_FATTR_NAME, xattr, xattr_len, 0, 0);
```
johnnovak, line 862: Extract `0` args no meaningful vars, etc.

johnnovak:
> > @NicknineTheEagle I hope I’ll have time to look at it tomorrow - but already one remark: there might be some standard for such extended attributes, as more application might want to store them (DOSBox-X, Wine, Samba); if so, it might be a good idea to follow it. See for example here: https://www.winehq.org/pipermail/wine-patches/2009-September/078227.html
> 
> Good thinking, I'll look into that as well.

I've had a look [at this implementation](https://github.com/samba-team/samba/blob/master/librpc/idl/xattr.idl), and I find it too complicated. Interoperability with other software that use this DOSATTRIB xattr format is not among our requirements. I suggest do the simple thing now, then in the future there's nothing preventing us from adding support for DOSATTRIB later if that's really needed (we could easily write both extended attributes, so there's a clear path forward without breaking anything). For now, let's shelve it.

NicknineTheEagle:
> > > @NicknineTheEagle I hope I’ll have time to look at it tomorrow - but already one remark: there might be some standard for such extended attributes, as more application might want to store them (DOSBox-X, Wine, Samba); if so, it might be a good idea to follow it. See for example here: https://www.winehq.org/pipermail/wine-patches/2009-September/078227.html
> > 
> > 
> > Good thinking, I'll look into that as well.
> 
> I've had a look [at this implementation](https://github.com/samba-team/samba/blob/master/librpc/idl/xattr.idl), and I find it too complicated. Interoperability with other software that use this DOSATTRIB xattr format is not among our requirements.
> 
> I'd suggest to do the simple thing now, then in the future, there's nothing preventing us from adding support for DOSATTRIB too if that's really needed (we could easily write both extended attributes, so there's a clear path forward without breaking anything). But as this is not really needed and would just complicate things for no good reason (as I see it), let's shelve it for now.

I looked at Wine patched linked above and it looks like it just uses hex number to store FAT attribute bits. I really don't see what's so complicated about it. Wine does mention that Samba 4 has some extended attribute functionality that they didn't bother bringing in, though.

src/dos/drive_local.cpp
NicknineTheEagle: Because Archive isn't just an attribute, it's automatically added by OS whenever the file is modified.
https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#attributes

src/misc/cross.cpp
NicknineTheEagle: What's the point of that if it's immediately converted to C string anyway for writing xattr?

include/cross.h
NicknineTheEagle: This is 31 flipping characters, it doesn't need to be that long. FatAttribsXAttrName is fine, xattr is an official name for that feature used in a bunch of places.

NicknineTheEagle:
> Okay, found one bug. If you made a file read-only prior to this change, it's impossible to change its attributes anymore (for example, to remove the read-only flag) because that changed the file permissions on the host filesystem level to non-writeable. Then you just get a log message like that:
> image
> 
> The only way to resolve the situation is to make the file writeable again outside of DOSBox.
> 
> We want a clear, seamless upgrade path for existing installations, so when setting attributes of existing files, we need to check if the file is read-only on the host filesystem level, and if so, we must make it writeable again (even when adding say "Hidden" and keeping the file "ReadOnly").
> 
> Plus if we fail to make the file writeable (it's possible), it should be logged as an error, not just an info level log message.

This might be a bad idea because the user may have manually removed write permissions for his own reasons (e.g. to protect the game files) so I don't think we should mess with permissions under the assumption that we were the ones that edited them before.

johnnovak:
> > > > @NicknineTheEagle I hope I’ll have time to look at it tomorrow - but already one remark: there might be some standard for such extended attributes, as more application might want to store them (DOSBox-X, Wine, Samba); if so, it might be a good idea to follow it. See for example here: https://www.winehq.org/pipermail/wine-patches/2009-September/078227.html
> > > 
> > > 
> > > Good thinking, I'll look into that as well.
> > 
> > 
> > I've had a look [at this implementation](https://github.com/samba-team/samba/blob/master/librpc/idl/xattr.idl), and I find it too complicated. Interoperability with other software that use this DOSATTRIB xattr format is not among our requirements.
> > I'd suggest to do the simple thing now, then in the future, there's nothing preventing us from adding support for DOSATTRIB too if that's really needed (we could easily write both extended attributes, so there's a clear path forward without breaking anything). But as this is not really needed and would just complicate things for no good reason (as I see it), let's shelve it for now.
> 
> I looked at Wine patch linked above and it looks like it just uses hex number to store FAT attribute bits which is brought over from Samba 3. I really don't see what's so complicated about it. Wine does mention that Samba 4 has some extended attribute functionality that they didn't bother bringing in, though.

It's extra complexity. They store a lot more, not just the attribs. Once you bring this in, we need to support it. Extra complexity => more room for things to break & extra maintenance effort. It needs to be justified, not just be brought in because yeah why not 😄 

This full implementation is a lot more complex than storing a 4-char string like we have now:

https://github.com/samba-team/samba/blob/master/librpc/idl/xattr.idl

Hex values make thing more hard to debug too. Overall, why would we need this full implementation, what's the use case here? If there's no solid use case and clear benefits over the current simple implementation in this PR that would warrant the extra complexity, then we shouldn't go there.

NicknineTheEagle:
> > > > > @NicknineTheEagle I hope I’ll have time to look at it tomorrow - but already one remark: there might be some standard for such extended attributes, as more application might want to store them (DOSBox-X, Wine, Samba); if so, it might be a good idea to follow it. See for example here: https://www.winehq.org/pipermail/wine-patches/2009-September/078227.html
> > > > 
> > > > 
> > > > Good thinking, I'll look into that as well.
> > > 
> > > 
> > > I've had a look [at this implementation](https://github.com/samba-team/samba/blob/master/librpc/idl/xattr.idl), and I find it too complicated. Interoperability with other software that use this DOSATTRIB xattr format is not among our requirements.
> > > I'd suggest to do the simple thing now, then in the future, there's nothing preventing us from adding support for DOSATTRIB too if that's really needed (we could easily write both extended attributes, so there's a clear path forward without breaking anything). But as this is not really needed and would just complicate things for no good reason (as I see it), let's shelve it for now.
> > 
> > 
> > I looked at Wine patch linked above and it looks like it just uses hex number to store FAT attribute bits which is brought over from Samba 3. I really don't see what's so complicated about it. Wine does mention that Samba 4 has some extended attribute functionality that they didn't bother bringing in, though.
> 
> It's extra complexity. They store a lot more, not just the attribs. Once you bring this in, we need to support it. Extra complexity => more room for things to break & extra maintenance effort. It needs to be justified, not just be brought in because yeah why not 😄
> 
> This full implementation is a lot more complex than storing a 4-char string like we have now:
> 
> https://github.com/samba-team/samba/blob/master/librpc/idl/xattr.idl
> 
> Hex values make thing more hard to debug too. Overall, why would we need this full implementation, what's the use case here? If there's no solid use case and clear benefits over the current simple implementation in this PR that would warrant the extra complexity, then we shouldn't go there.

Why should we bring in the full implementation in the first place? Wine has only brought in "user.DOSATTRIB" with hex value.

johnnovak:
> Why should we bring in the full implementation in the first place? Wine has only brought in "user.DOSATTRIB" with hex value and they're doing okay.

Ok, perhaps I misunderstood then. Maybe link to the Wine implementation then. If it's just a single byte or something stored in "user.DOSATTRIB", and it's just about the 4 FAT flags, then OK.

NicknineTheEagle:
> > Why should we bring in the full implementation in the first place? Wine has only brought in "user.DOSATTRIB" with hex value and they're doing okay.
> 
> Ok, perhaps I misunderstood then. Maybe link to the Wine implementation then. If it's just a single byte or something stored in "user.DOSATTRIB", and it's just about the 4 FAT flags, then OK.

That's literally what FeralChild did in OP, lol.

johnnovak:
In other news, I've tested PKZIP and RAR, and couldn't get PKZIP to store DOS attributes of the archived files. There's no mentioning of attribs in their doco either.

RAR, on the other hand, stores FAT DOS attribs by default (it's configurable). 🚀 Tried archiving & restoring a few files with various attribs, works fine.

johnnovak:
> > > Why should we bring in the full implementation in the first place? Wine has only brought in "user.DOSATTRIB" with hex value and they're doing okay.
> > 
> > 
> > Ok, perhaps I misunderstood then. Maybe link to the Wine implementation then. If it's just a single byte or something stored in "user.DOSATTRIB", and it's just about the 4 FAT flags, then OK.
> 
> That's literally what FeralChild did in OP, lol, it's a link to Wine patch.

Ok, where's the link to the patch? 😄 I click on the link, I see a message that mentions a patch... No time to find the patch, link me to the code in their repo 😄 But OK, if it's just about the hex flags, I can't imagine it's complicated.

NicknineTheEagle:
> > > > Why should we bring in the full implementation in the first place? Wine has only brought in "user.DOSATTRIB" with hex value and they're doing okay.
> > > 
> > > 
> > > Ok, perhaps I misunderstood then. Maybe link to the Wine implementation then. If it's just a single byte or something stored in "user.DOSATTRIB", and it's just about the 4 FAT flags, then OK.
> > 
> > 
> > That's literally what FeralChild did in OP, lol, it's a link to Wine patch.
> 
> Ok, where's the link to the patch? 😄 I click on the link, I see a message that mentions a patch... No time to find the patch, link me to the code in their repo 😄 But OK, if it's just about the hex flags, I can't imagine it's complicated.

There's a link to .bin file at the very bottom. Yeah, it took me a while to find it as well.

johnnovak:
> There's a link to .bin file at the very bottom. Yeah, it took me a while to find it as well.
> http://www.winehq.org/pipermail/wine-patches/attachments/20090908/db865417/attachment-0001.bin

Cheers. Yeah ok, it's basically what you've done, just a hex value instead of the strings. Sure, that will do, I'm fine with either the 4-byte string or this. Flip a coin 😄 Maybe this would give us better compatibility with Wine, might be useful to some.

Will be a nice opportunity for you to play around with `bit_view` if you go the hex value route 😎

FeralChild64:
Just checked - it seems both _Wine_ (https://github.com/wine-mirror/wine/blob/master/dlls/ntdll/unix/file.c, see `parse_samba_dos_attrib_data` routine) and _Dosemu2_ (https://github.com/dosemu2/dosemu2/blob/devel/src/dosext/mfs/xattr.c) support the simple "one hex value" variant of `user.DOSATTRIB`, so I would vote to reuse this format in _DOSBox Staging_ too.
Especially interoperability between _DOSBox_ and _Dosemu2_ might be useful for testing purposes, not to mention that if _KDE_/_Gnome_/etc. decides to support DOS/Windows emulators, they will have an easier job to do.

johnnovak:
> Just checked - it seems both _Wine_ (https://github.com/wine-mirror/wine/blob/master/dlls/ntdll/unix/file.c, see `parse_samba_dos_attrib_data` routine) and _Dosemu2_ (https://github.com/dosemu2/dosemu2/blob/devel/src/dosext/mfs/xattr.c) support the simple "one hex value" variant of `user.DOSATTRIB`, so I would vote to reuse this format in _DOSBox Staging_ too. Especially interoperability between _DOSBox_ and _Dosemu2_ might be useful for testing purposes, not to mention that if _KDE_/_Gnome_/etc. decides to support DOS/Windows emulators, they will have an easier job to do.

Sounds reasonable @FeralChild64 , let's do that!

src/misc/cross.cpp
johnnovak: Consistency. Safety. But moot point if you're going the hex value route.

src/dos/drive_local.cpp
johnnovak: That Wikipedia page states this:

> [Archive](https://en.wikipedia.org/wiki/Archive_bit). (Since DOS 2.0) Typically set by the operating system as soon as the file is created or modified to mark the file as "dirty", and reset by backup software once the file has been backed up to indicate "pure" state.

So you simply set the flag in the create file and write file calls in the DOSBox code, I don't see the difficulty. Surely we don't have hundreds of functions that needs this adjustment, but only a few. Certainly doable with minimal extra effort 😄 

Then the modify attribute call, and by extension the ATTRIB command could clear (or set) this flag.

If you go to the trouble of adding attribute support, then let's do a complete job and do it properly.

johnnovak:
I've resolved my string handling related comments as they're no longer relevant (although they contained some useful tips 😄).

johnnovak:
Could you please give us a status update on this @NicknineTheEagle ? It would be nice improvement for sure to add proper attribute handling.

NicknineTheEagle:
> Could you please give us a status update on this @NicknineTheEagle ? It would be nice improvement for sure to add proper attribute handling.
> 
> If you're busy and going to work on this later, that's completely fine of course. I'll just remove it from the 0.81.0 project then.

I'll go through the review suggestions tomorrow.

johnnovak:
Closing & removing from the 0.81.0 scope due to 1+ month of inactivity.

I've created an issue ticket if someone wants to pick this up in the future:
https://github.com/dosbox-staging/dosbox-staging/issues/new

PR #2694, opened 2023-07-21
Fix joysticks breaking after a config -set

### LINKED ISSUE #2687: Changing some configuration options at runtime disables joysticks

_No response_

Linux x86_64

Also affects macOS ARM

### Relevant hardware info

Thrustmaster T.Flight Stick X and VKB Gladiator NXT Evo are both affected

If certain configuration parameters are changed using `config -set`, joysticks stop responding in DOS. They still function in the keymapper, but neither axes nor buttons respond in DOS programs. The `joytest` utility I found [on VOGONS](https://www.vogons.org/viewtopic.php?p=187168#p187168) exhibits the problem, as do multiple games I tried.

The `sdl mapperfile` configuration parameter is a simple one to test, but I believe changing any parameters under `sdl` will cause the problem.

Explain how to reproduce
1. Run joytest utility; confirm it works
2. Execute `config -set sdl mapperfile=test.map`
3. Run joytest utility again; confirm it does not work

_No response_

### Your configuration

```ini
[sdl]
fullscreen = true
```

### Provide a Log

```text
https://gist.github.com/JKingweb/5ee19fc724a4ba9fca9c7d5afef076a6
```

- [X] Yes, I agree.

weirddan455:
Fixes #2687

`MAPPER_BindKeys` gets called when an SDL config entry gets changed.  This resets this global `num_groups` variable.  On first run, `CreateBindGroups()` gets called which increments this variable.  On subsequent runs, the bindgroups do not get re-created and resetting this variables makes it think there are no joysticks connected.

This whole file is kind of hard to follow... globals and raw pointers everywhere.  This fixes the problem as far as I can tell though.

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -2802,10 +2802,11 @@ void BIND_MappingEvents() {
 // results. If no joysticks are valid then joytype is set to JOY_NONE_FOUND.
 // This also resets mapper.sticks.num_groups to 0 and mapper.sticks.num to the
 // number of found SDL joysticks.
+
+// 7-21-2023: No longer resetting mapper.sticks.num_groups due to https://github.com/dosbox-staging/dosbox-staging/issues/2687
 static void QueryJoysticks()
 {
 	// Reset our joystick status
-	mapper.sticks.num_groups = 0;
 	mapper.sticks.num = 0;
 
 	JOYSTICK_ParseConfiguredType();


johnnovak:
Great stuff @weirddan455 🎉 Would you be able to test this on Windows too?

> This whole file is kind of hard to follow... globals and raw pointers everywhere. This fixes the problem as far as I can tell though.

Don't even get me started 😅  The whole mapper spaghetti monster, including controller handling, will be burned to the ground eventually, the land will be sowed with salt, then we'll rebuild it from the ground, based on sane principles.

Thanks for venturing into this perilous realm and coming out alive 😎 

If someone can confirm this on Windows, I'm happy to merge (I just don't have the time right now).

johnnovak:
> Confirmed the bug happens in main on Windows and this PR fixes it.

Thanks @weirddan455 !

PR #2697, opened 2023-07-22
website: Remove 0.79.0 release from downloads as the files are missing

johnnovak:
I just noticed that kcgen had apparently removed the 0.79.0 release binaries from the GitHub releases with this comment:

> Instead of 0.79.0, we recommend downloading the 0.79.1 maintenance release because it fixes several issues.
> 
> (The 0.79.0 release can always be compiled from source or via the v079.0 commit tag)

See here: https://github.com/dosbox-staging/dosbox-staging/releases/tag/v0.79.0

I don't quite like this; I think we shouldn't remove `.0` releases retrospectively from the downloads section just because we released a patch point release later. We _want_ all of our published releases archived for easier regression testing at the very least! (e.g. I can't rebuild any commit before about 2022 May here on macOS; I'm getting linker errors... surely fixable, but time-consuming)

Anyway, I'm not aware of an easy way to rebuild the missing artifacts, so I'm removing the broken links instead as that reeks of amateur hour... I'd prefer to rebuild those artifacts and restore the download links, so if anyone knows how to do that instead, I'm all ears!

### DIFF
--- a/website/docs/downloads/linux.md
+++ b/website/docs/downloads/linux.md
@@ -132,12 +132,6 @@ are designed with developers and testers in mind.
   sha256: aebf8619bb44934f18d0e219d50c4e2c<wbr>03b179c37daa67a9b800e7bd3aefc262
   </small>
 
-- [DOSBox Staging 0.79.0 (tar.xz)][0_79_0]
-  <br>
-  <small>
-  sha256: 804adb294096ab651490b1664570203f<wbr>24e460048d7e6e2e388d210a8380016a
-  </small>
-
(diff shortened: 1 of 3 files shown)


FeralChild64:
I’m curious - how is it at all possible that the file is missing?

FeralChild64 (approved):

johnnovak:
> I’m curious - how is it at all possible that the file is missing?

You can edit releases with the pencil icon and remove any of the files that are part of the release. I'm guessing kcgen did just that.

PR #2698, opened 2023-07-22
Accumulate GUS frames accross each voice instead of per-sample (#2696)

### LINKED ISSUE #2696: Gravis Ultrasound sound effects fuzzy/crackly in DUKE3D when set to 44kHz

dosbox-staging-v0.81.0-alpha-1149-g09ee

Windows 10

_No response_

### Relevant hardware info

i5-4590/r9 290, i7-8650u/UHDgraphics620

Sound effects exhibit significant distortion in DUKE3D when the game setup options are set to use the GUS at 44kHz. 22kHz and below sounds normal. GUS music sounds correct.

The current release version of DosBox-Staging exhibits similar distortions in several additional games (A Final Unity, Heretic, Doom) at any sample rate setting, but they all sound correct in the latest Dev build, so the latest patches must have fixed them.

The latest release build of DOSBox-X using the original GUS implementation isn't exhibiting the distortion at 44kHz.

[DUKE3D-44kHz-GUS.zip](https://github.com/dosbox-staging/dosbox-staging/files/12135197/DUKE3D-44kHz-GUS.zip)

Explain how to reproduce
1. with fresh config file using known working mounted C:\ with \ULTRASND\ and \DUKE3D\ directory, disable sound blaster (this actually didn't seem to affect it, just removing a variable) and enable GUS
2.  set sound device in duke3d setup to GUS, 44kHz mixing rate, disable music to hear SFX more clearly (number of voices doesn't seem to affect it, music sounds correct either way when enabled)
3. start duke3d, the initial radiation logo animation and main menu demo will show obvious sound distortion

Atomic Edition of DUKE3D from GOG

### Your configuration

```ini
https://gist.github.com/x73rmin8r/36c2d6569e31d5df5f617231c7d7c275
```

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
Previously when rendering frames in the GUS, we iterated one sample at a time, gathering the next panned sample from each voice into the resulting frame before moving onto the next frame (and again, gathering the next sample from all the voices).

```
         start
          i=0,     i=1,    i=2, .. for N frames
           |        |       |
           V        V       V
voice 1  [v1s0]   [v1s1]   ...
voice 2 +[v2s0]  +[v2s1]   ...
voice 3 +[v3s0]  +[v3s1]   ...
...
output = {summed-s0, summed-s1, summed-s2 ... }
```

Rendering samples from the GUS changes its state, and in the case of Duke 3D, the above approach doesn't traverse all of the available samples in all the populated voices before a state change halts one or more of the voices, resulting in essentially a truncated sequence.

This change switches back to how the original SVN and X GUS rendering loops work, which is to gather samples per-voice first, summing all of the samples across an array. This ensures that we gather all the requested samples from each voice before moving onto the next voice.

```
start
i=1 ->  voice 1   [v1s0]   [v1s1]   ... for N frames
i=2,    voice 2  +[v2s0]  +[v2s1]   ... for N frames
i=3,    voice 3  +[v3s0]  +[v3s1]   ... for N frames
...
output = {summed-s0, summed-s1, summed-s2 ... }
```

The sub-millisecond rendering prior to port and register writes is still in place, as is the use of the mixer's floating-point compressor.

Fixes #2696 ; thanks for reporting @x73rmin8r!

### DIFF
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -24,9 +24,10 @@
 #include <array>
 #include <iomanip>
 #include <memory>
+#include <queue>
 #include <string>
 #include <unistd.h>
-#include <queue>
+#include <vector>
 
 #include "autoexec.h"
 #include "control.h"
@@ -134,9 +135,10 @@ class Voice {
 public:
 	Voice(uint8_t num, VoiceIrq &irq) noexcept;
 
-	AudioFrame RenderFrame(const ram_array_t &ram,
-	                       const vol_scalars_array_t &vol_scalars,
-	                       const pan_scalars_array_t &pan_scalars);
+	void RenderFrames(const ram_array_t& ram,
+	                  const vol_scalars_array_t& vol_scalars,
+	                  const pan_scalars_array_t& pan_scalars,
+	                  std::vector<AudioFrame>& frames);
 
 	uint8_t ReadVolState() const noexcept;
 	uint8_t ReadWaveState() const noexcept;
@@ -253,7 +255,9 @@ private:
 
 	void RegisterIoHandlers();
 	void Reset(uint8_t state);
-	AudioFrame RenderFrame();
+
+	const std::vector<AudioFrame>& RenderFrames(const int num_requested_frames);
+
 	void RenderUpToNow();
 	void StopPlayback();
 	void UpdateDmaAddress(uint8_t new_address);
@@ -271,6 +275,7 @@ private:
 	read_io_array_t read_handlers   = {};
 	write_io_array_t write_handlers = {};
 	voice_array_t voices            = {{nullptr}};
+	std::vector<AudioFrame> rendered_frames = {};
 
 	const address_array_t dma_addresses = {
 	        {MIN_DMA_ADDRESS, 1, 3, 5, 6, MAX_IRQ_ADDRESS, 0, 0}};
@@ -441,21 +446,25 @@ float Voice::GetSample(const ram_array_t &ram) noexcept
 	return sample;
 }
 
-AudioFrame Voice::RenderFrame(const ram_array_t &ram,
-                              const vol_scalars_array_t &vol_scalars,
-                              const pan_scalars_array_t &pan_scalars)
+void Voice::RenderFrames(const ram_array_t& ram,
+                         const vol_scalars_array_t& vol_scalars,
+                         const pan_scalars_array_t& pan_scalars,
+                         std::vector<AudioFrame>& frames)
 {
 	if (vol_ctrl.state & wave_ctrl.state & CTRL::DISABLED)
-		return {0.0f, 0.0f};
-
-	// Keep track of how many ms this voice has generated
-	Is16Bit() ? ++generated_16bit_ms : ++generated_8bit_ms;
-
-	const auto sample = GetSample(ram) * PopVolScalar(vol_scalars);
+		return;
 
 	const auto pan_scalar = pan_scalars.at(pan_position);
 
-	return {sample * pan_scalar.left, sample * pan_scalar.right};
+	// Sum the voice's samples into the exising frames, angled in L-R space
+	for (auto& frame : frames) {
+		float sample = GetSample(ram);
+		sample *= PopVolScalar(vol_scalars);
+		frame.left += sample * pan_scalar.left;
+		frame.right += sample * pan_scalar.right;
+	}
+	// Keep track of how many ms this voice has generated
+	Is16Bit() ? generated_16bit_ms++ : generated_8bit_ms++;
 }
 
 // Returns the current wave position and increments the position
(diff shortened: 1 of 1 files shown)


src/hardware/gus.cpp
```diff
@@ -441,21 +445,25 @@ float Voice::GetSample(const ram_array_t &ram) noexcept
 	return sample;
 }
 
-AudioFrame Voice::RenderFrame(const ram_array_t &ram,
-                              const vol_scalars_array_t &vol_scalars,
-                              const pan_scalars_array_t &pan_scalars)
+void Voice::RenderVoiceToFrames(const ram_array_t& ram,
+                                const vol_scalars_array_t& vol_scalars,
+                                const pan_scalars_array_t& pan_scalars,
+                                std::vector<AudioFrame>& frames)
```
johnnovak, line 451: It's in the `Voice` class, so RenderVoice is redundant. Call it `RenderFrames` or just `Render`?

src/hardware/gus.cpp
```diff
@@ -665,27 +673,37 @@ void Gus::ActivateVoices(uint8_t requested_voices)
 	}
 }
 
-AudioFrame Gus::RenderFrame()
+const std::vector<AudioFrame>& Gus::GetFrames(const int num_requested_frames)
```
johnnovak, line 676: Perhaps keep this `RenderFrames` so it's consistent with all the other audio device implementations (well, MT-32 and FluidSynth had only `Render`, but the rest `RenderFrames`). The fact that this function forwards the rendering job to the `Voice` classes is an implementation details, it should not be reflected in the function's name.

src/hardware/gus.cpp
```diff
@@ -665,27 +673,37 @@ void Gus::ActivateVoices(uint8_t requested_voices)
...
 {
-	AudioFrame accumulator = {};
+	// use a static vector to avoid re-allocating each millisecond
+	static std::vector<AudioFrame> frames = {};
```
johnnovak, line 679: How about storing this in the `Gus` class instead? Would be a bit cleaner that way, no?

src/hardware/gus.cpp
kcgen: Yup.. before I was called 'RenderFrame' (in the voice class), and we also had 'RenderFrame' in the Gus class; I was worried they were confusing or ambiguous, so renamed them.. but with your comments here I think they originals were fine :-)  Will rename them back.

src/hardware/gus.cpp
kcgen: Yup.. will make it a member.

src/hardware/gus.cpp
```diff
@@ -665,27 +673,37 @@ void Gus::ActivateVoices(uint8_t requested_voices)
...
-
-		while (voice < voice_end && *voice) {
-			const auto voice_frame = voice->get()->RenderFrame(
-			        ram, vol_scalars, pan_scalars);
-
-			accumulator.left += voice_frame.left;
-			accumulator.right += voice_frame.right;
-
+		const auto last_voice = voice + active_voices;
+		while (voice < last_voice && *voice) {
```
johnnovak, line 690: Related: can't we move this into `Gus` and remove the `unique_ptr` cruft?

```cpp
using voice_array_t = std::array<std::unique_ptr<Voice>, MAX_VOICES>;
```

src/hardware/gus.cpp
```diff
@@ -665,27 +673,37 @@ void Gus::ActivateVoices(uint8_t requested_voices)
...
+			// Render all of the requested frames from each voice
+			// before moving onto the next voice. This ensures each
+			// voice can deliver all its samples without being
+			// affected by state changes that (might) occur when
+			// rendering subsequence voices.
```
johnnovak, line 695: subsequent

src/hardware/gus.cpp
```diff
@@ -698,33 +716,45 @@ void Gus::RenderUpToNow()
...
-	// Keep rendering until we're current
-	while (last_rendered_ms < now) {
-		last_rendered_ms += ms_per_render;
-		fifo.emplace(RenderFrame());
+
+	const auto elapsed_ms = now - last_rendered_ms;
+	if (elapsed_ms > ms_per_render) {
+		// How many frames have elapsed since we last rendered?
+		const auto num_elapsed_frames = iround(
+		        std::floor(elapsed_ms / ms_per_render));
```
johnnovak, line 724: Are you sure we aren't swallowing/skipping sample frames sometimes because of the rounding?

src/hardware/gus.cpp
johnnovak: Yeah I think that's better 😄 The scope is still very clear, e.g. `voice->RenderFrames(...)`.

johnnovak:
Nice quick fix @kcgen , will test shortly with a few demos and mod players as well 👍🏻

src/hardware/gus.cpp
kcgen: willsnip the `unique_ptr` and moves to full objects.

johnnovak (approved):
The issue in Duke3D is now gone. I've also tried FastTracker II, CapaMod and a few demos—all good :tada: Feel free to merge @kcgen once you've addressed those minor issues.

src/hardware/gus.cpp
kcgen: Yes; this errs on the side of whole frames only for the sub-millisecond chunk of time, because the game or program at a minimum needs to make sure at least that many samples have been DMA'd into the card - if we round up, then we might go beyond what's available.

But to to sure - no matter what, during the audio callback where the mixer pulls frames, it gets all the frames at that point.  So no frames are dropped or lost.

kcgen:
> The issue in Duke3D is now gone. I've also tried FastTracker II, CapaMod and a few demos—all good tada Feel free to merge @kcgen once you've addressed those minor issues.

Thanks for the bisect, review, and test feedback @johnnovak !

johnnovak:
> > The issue in Duke3D is now gone. I've also tried FastTracker II, CapaMod and a few demos—all good tada Feel free to merge @kcgen once you've addressed those minor issues.
> 
> Thanks for the bisect, review, and test feedback @johnnovak !

Nice one! You're back and you're solving issues in a swift manner on day 1! 🚀 😎 🤘🏻

Issue #1692, opened 2022-05-22
Better path resolution in config files


johnnovak:
I think it would be better if path resolution was always relative to the config file that's being processed, not the directory where DOSBox was launched from.

Example:

I have this in my global config in `D:\Emulators\DOS\DOSBox-Staging\dosbox-staging.conf`. 

```ini
[fluidsynth]
soundfont = ..\SoundFonts\GeneralUser GS 1.471\GeneralUser GS v1.471.sf2

[mt32]
model = cm32l
romdir = ..\ROMs\Roland\

[autoexec]
mount U "..\Utils"
```

This will only work if I launch DOSBox from its home directory. But I use cascading configs and batch files to launch games from their own directories, so I need to use absolute paths as a workaround. This is also a problem for portable setups, so probably makes the life of eXoDOS people a bit harder.

To illustrate the problem better, this is my setup:

```
DOSBox-Staging
    dosbox.exe
    dosbox-staging.conf
Games
    Adventure
        Alone in the Dark
            C
            CD
            Extras
            Alone in the Dark.bat
            dosbox.conf
        Azrael's Tear
            C
            CD
            Extras
            Azrael's Tear.bat
            dosbox.conf
        ...
    RPG
        Akalabeth - World of Doom
            C
            Extras
            Akalabeth - World of Doom.bat
            dosbox.conf
        Albion
            ...
```

Thoughts?

kcgen:
I think that makes a lot of sense, and might even simplify some of the logic. 

The layered configs have a lot of benefits, but it's a concept that lots of users miss out on. I think this approach would make it even clearer and have more predictable behavior.

johnnovak:
Upon further consideration, config-location relative path resolution would cause great confusion, and after all it wouldn't be that useful. Adding DOSBox executable path priority to some of these settings would solve the problem in a less surprising and more robust way.

kcgen:
> Adding DOSBox executable path priority to some of these settings would solve the problem in a less surprising and more robust way.

I think so too. We probably need a PR that systematically goes through each loadable conf settings, and bases them off the resource loader. 

Using the exe-path-first gets us as close as possible to having pinned or versionable file assets, tied to their releases. 

So users could have 0.81 with its tree of sf2, mt32 ROMs, shaders, etc, and some future release build in a separate directory also with it's own data files. There would be no overlap of confusion.

PR #2700, opened 2023-07-25
Refactor Prop_path to use std_fs::path

### LINKED ISSUE #2636: Refactor misc/setup.cpp Prop_path

_No response_

Linux x86_64

_No response_

Prop_path has caused some previous issues.  See #2629 and #2630

The main issue being that if a user's config is missing an entry, `Prop_path::SetValue` never gets executed.  This function is supposed to set the variable `realpath` to an absolute path.  Instead, you can get left with just a filename.

`Prop_path::SetValue` is using a static global variable `current_config_dir` which gets set while the config file is being parsed and then cleared immediately after.  This means the function is unable to be called during construction time.

I think ideally, that global should be done away with and replaced with a call to `get_platform_config_dir()`.  This would let us set the `realpath` variable during construction of the object.

As an additional improvement, it may be worth looking to have `Prop_path` store a C++ 17 `std_fs::path` (alias for `std::filesystem::path`) rather than a string.

_No response_

_No response_

- [X] Yes, I agree.

### LINKED ISSUE #2695: config -set sdl mapperfile requires full path

_No response_

Linux x86_64

_No response_

### Relevant hardware info

_No response_

Changing the mapperfile using `config -set sdl mapperfile=mymap.map` does not work.  It instead requires typing out the entire path.  ex.  `config -set sdl mapperfile=~/.config/dosbox/mymap.map`

This is in contrast to setting the mapperfile in the config file, where it automatically prepends the config directory `~/.config/dosbox`.

I believe this is going to be related to some refactoring work that needs to be done in #2636

Explain how to reproduce
1.   `dosbox -startmapper` + click "Save" to create a mapperfile
2.   Rename the mapper file to test.map
3.   `config -set sdl mapperfile=test.map`
4.    It fails to load.

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
Fixes #2695
Fixes #2636

This was a lot simpler than I thought it would be since it's not used in that many places.  I made Prop_path store a std_fs::path instead of a string.  I also made sure it gets initialized properly to an absolute path on object construction which lets me get rid of the workaround I wrote for the shell history.

### DIFF
--- a/src/hardware/serialport/softmodem.cpp
+++ b/src/hardware/serialport/softmodem.cpp
@@ -67,13 +67,13 @@ static bool MODEM_IsPhoneValid(const std::string &input) {
 	return true;
 }
 
-bool MODEM_ReadPhonebook(const std_fs::path &filename) {
-	std::ifstream loadfile(filename);
+bool MODEM_ReadPhonebook(const std_fs::path &path) {
+	std::ifstream loadfile(path);
 	if (!loadfile)
 		return false;
 
-	const std::string file_string = filename.string();
-	LOG_MSG("SERIAL: Phonebook loading from %s.", file_string.c_str());
+	const auto path_string = path.string();
+	LOG_MSG("SERIAL: Phonebook loading from %s.", path_string.c_str());
 
 	std::string linein;
 	while (std::getline(loadfile, linein)) {
@@ -81,8 +81,8 @@ bool MODEM_ReadPhonebook(const std_fs::path &filename) {
 		std::string phone, address;
 
 		if (!(iss >> phone >> address)) {
-			LOG_MSG("SERIAL: Phonebook skipped a bad line in %s.",
-			        file_string.c_str());
+			LOG_MSG("SERIAL: Phonebook skipped a bad line in '%s'",
+			        path_string.c_str());
 			continue;
 		}
 
--- a/src/hardware/serialport/softmodem.cpp
+++ b/src/hardware/serialport/softmodem.cpp
@@ -67,12 +67,13 @@ static bool MODEM_IsPhoneValid(const std::string &input) {
 	return true;
 }
 
-bool MODEM_ReadPhonebook(const std::string &filename) {
+bool MODEM_ReadPhonebook(const std_fs::path &filename) {
 	std::ifstream loadfile(filename);
 	if (!loadfile)
 		return false;
 
-	LOG_MSG("SERIAL: Phonebook loading from %s.", filename.c_str());
+	const std::string file_string = filename.string();
+	LOG_MSG("SERIAL: Phonebook loading from %s.", file_string.c_str());
 
 	std::string linein;
 	while (std::getline(loadfile, linein)) {
@@ -81,7 +82,7 @@ bool MODEM_ReadPhonebook(const std::string &filename) {
 
 		if (!(iss >> phone >> address)) {
 			LOG_MSG("SERIAL: Phonebook skipped a bad line in %s.",
-			        filename.c_str());
+			        file_string.c_str());
 			continue;
 		}
 
--- a/src/hardware/serialport/softmodem.h
+++ b/src/hardware/serialport/softmodem.h
@@ -63,7 +63,7 @@ enum ResTypes {
 #define TEL_CLIENT 0
 #define TEL_SERVER 1
 
-bool MODEM_ReadPhonebook(const std::string &filename);
+bool MODEM_ReadPhonebook(const std_fs::path &filename);
 void MODEM_ClearPhonebook();
 
 class CFifo {
--- a/src/hardware/serialport/softmodem.h
+++ b/src/hardware/serialport/softmodem.h
@@ -63,7 +63,7 @@ enum ResTypes {
 #define TEL_CLIENT 0
 #define TEL_SERVER 1
 
-bool MODEM_ReadPhonebook(const std_fs::path &filename);
+bool MODEM_ReadPhonebook(const std_fs::path &path);
 void MODEM_ClearPhonebook();
 
 class CFifo {
--- a/src/misc/setup.cpp
+++ b/src/misc/setup.cpp
@@ -444,17 +444,17 @@ bool Prop_path::SetValue(const std::string& input)
 		return false;
 	}
 
-	const std_fs::path workcopy = resolve_home(input);
+	const auto temp_path = resolve_home(input);
 
-	if (workcopy.is_absolute()) {
-		realpath = workcopy;
+	if (temp_path.is_absolute()) {
+		realpath = temp_path;
 		return is_valid;
 	}
 
 	if (current_config_dir.empty()) {
-		realpath = get_platform_config_dir() / workcopy;
+		realpath = get_platform_config_dir() / temp_path;
 	} else {
-		realpath = current_config_dir / workcopy;
+		realpath = current_config_dir / temp_path;
 	}
 
 	return is_valid;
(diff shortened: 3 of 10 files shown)


src/hardware/serialport/softmodem.cpp
```diff
@@ -67,12 +67,13 @@ static bool MODEM_IsPhoneValid(const std::string &input) {
 	return true;
 }
 
-bool MODEM_ReadPhonebook(const std::string &filename) {
+bool MODEM_ReadPhonebook(const std_fs::path &filename) {
```
johnnovak, line 70: Renaming the arg to `path` would be better.

src/hardware/serialport/softmodem.cpp
```diff
@@ -67,12 +67,13 @@ static bool MODEM_IsPhoneValid(const std::string &input) {
...
 	std::ifstream loadfile(filename);
 	if (!loadfile)
 		return false;
 
-	LOG_MSG("SERIAL: Phonebook loading from %s.", filename.c_str());
+	const std::string file_string = filename.string();
```
johnnovak, line 75:
```cpp
const auto path_string = path.string();
```

src/hardware/serialport/softmodem.cpp
```diff
@@ -81,7 +82,7 @@ bool MODEM_ReadPhonebook(const std::string &filename) {
 
 		if (!(iss >> phone >> address)) {
 			LOG_MSG("SERIAL: Phonebook skipped a bad line in %s.",
```
johnnovak, line 84: Pre-existing issue; can you please enclose filenames in log messages with single-quotes? (so this will become `'%s'`, plus in this case I'd get rid of the period at the end because it could be confusing when the messages ends with a filename or a parameter)

src/hardware/serialport/softmodem.h
```diff
@@ -63,7 +63,7 @@ enum ResTypes {
 #define TEL_CLIENT 0
 #define TEL_SERVER 1
 
-bool MODEM_ReadPhonebook(const std::string &filename);
+bool MODEM_ReadPhonebook(const std_fs::path &filename);
```
johnnovak, line 66: `path`

src/shell/shell.cpp
```diff
@@ -470,7 +460,7 @@ void DOS_Shell::ReadShellHistory()
 	if (control->SecureMode()) {
 		return;
 	}
-	std::string history_path = get_shell_history_path();
+	const auto history_path = get_shell_history_path();
```
johnnovak, line 463: 👍🏻

src/shell/shell.cpp
```diff
@@ -492,14 +482,14 @@ void DOS_Shell::WriteShellHistory()
...
```
johnnovak, line 485: 👍🏻

src/shell/shell.cpp
```diff
@@ -443,22 +443,12 @@ void DOS_Shell::SyntaxError()
...
-			// Workaround: If shell_history_file is not in the
-			// user's config, path->SetValue function never gets
-			// called which sets realpath. Append the config dir
-			// here to prevent this from getting written to current
-			// working directory in that case.
-			if (!path->realpath.empty() &&
-			    path->realpath.find(CROSS_FILESPLIT) == std::string::npos) {
-				return (get_platform_config_dir() / path->realpath)
-				        .string();
-			}
```
johnnovak, line 461: Good riddance! 🎉

src/misc/setup.cpp
```diff
@@ -444,18 +444,17 @@ bool Prop_path::SetValue(const std::string& input)
 		return false;
 	}
 
-	const std::string workcopy = resolve_home(input).string();
+	const std_fs::path workcopy = resolve_home(input);
```
johnnovak, line 447: Uh, can we use a better name for this? `workcopy` — that can be literally anything... Like `data`, `temp`, `stuff` etc. 😅

johnnovak:
Nice one @weirddan455 and it's great to see your enthusiasm in improving our codebase! 🎉 This looks good apart from some fairly minor comments, however, I noticed a subtle regression during my testing. Admittedly it's an edge case that's not easy to see from the code alone.

Here's how to reproduce:

### Test 1 – main

1. Delete all mapperfiles and the global config from your config directory.

1. Start DOSBox with `dosbox -noprimaryconf` in an empty directory that has no local `dosbox.conf` .

1. Note `MAPPER: Loaded default key bindings` gets logged as expected.

1. Enter the keymapper with Ctrl+F1 then press Save. This writes a new mapperfile to the _local working directory_ named `mapper-sdl2-0.81.0-alpha.map`.


### Test 2 – this PR

1. Repeat the same steps from Test 1.

2. Note that the mapperfile gets written to the _config directory_ instead of the _local working directory_ DOSBox was started from — **this is a regression.**


### Test 3 – main

1. At this two copies of `mapper-sdl2-0.81.0-alpha.map` exist: one in your config directory, and one in your local directory you start DOSBox from.

1. Start DOSBox with `dosbox -noprimaryconf`.

1. Note the mapperfile from the _local working directory_ gets loaded.


### Test 4 – this PR

1. Start DOSBox with `dosbox -noprimaryconf`.

1. Note the mapperfile from the _config directory_ gets loaded — **this is a regression.**

----

For completeness, this is the full list of config settings that take path values. Luckily, there aren't too many of them. Could you please check that your fix hasn't introduced similar regressions to the rest of them?

Better to test thoroughly now than to let our users run into them—troubleshooting these subtle path resolution issues can be confusing even for us devs. The main concern here is that these subtle changes can break existing setups and gamepacks like eXoDOS. `soundfont`, `romdir` and `mapperfile` are the most important ones to get 100% regression free.

Thank you for your help on this!

```ini
[sdl]
mapperfile

[render]
glshader

[capture]
capture_dir

[fluidsynth]
soundfont

[mt32]
romdir

[serial]
phonebookfile

[dos]
shell_history_file
```

weirddan455:
@johnnovak Good catch on the regressions.  I fixed the style issues you pointed out and I'll see if I can get the path issue sorted out shortly.

weirddan455:
@johnnovak Your test 1 + 2 sounds very familiar to the bug report that inspired me to write the workaround for `shell_history_file` #2629

@FeralChild64 wrote that bug report and made the argument that we should not be writing files to the current working directory, which I am inclined to agree with.  I think we need to come to an agreement to what paths we should be searching, because as of now this is just inconsistent.

IMO, if some gamepack wants to ship their own mapperfile, they should also ship a `dosbox.conf` pointing to that file.  I'm not familiar with how those are packaged to know if this is even a practical issue or not.

johnnovak:
> @FeralChild64 wrote that bug report and made the argument that we should not be writing files to the current working directory, which I am inclined to agree with. I think we need to come to an agreement to what paths we should be searching, because as of now this is just inconsistent.
 
Yeah, good point, and I absolutely agree--we need to be consistent. Let me think a bit about that, maybe we don't actually need to change anything here. With relative paths, resolving them relative to the config they're defined in is the most logical approach, and looks like that's the behaviour you introduced. I'll play around with your branch a bit more and get back to you today.

> IMO, if some gamepack wants to ship their own mapperfile, they should also ship a `dosbox.conf` pointing to that file. I'm not familiar with how those are packaged to know if this is even a practical issue or not.

Yeah and on a second thought, it's pretty unlikely they would run into this edge case, and even then there are various solutions and workarounds available.

Let's do the logical and correct thing!

johnnovak (approved):
Ok, so after further consideration @weirddan455 I think this behaviour is better overall, regression or not.

If the user specifies `mapperfile` in the local config as a filename, then it will be looked up in the local directory, otherwise the config directory will  used. That sounds reasonable.

About the normalisation of the path lookup behaviour, I did create a ticket for that a while ago, just remembered. Let's continue the discussion there as I think it serves as a good starting point.

https://github.com/dosbox-staging/dosbox-staging/issues/1692

PR #2552, opened 2023-05-29
Simplify and improve `presentation_mode` handling (#2551)

kcgen:
With more feedback and data to guide the presentation_mode's `auto` behavior, this PR make safer choices to try to avoid frame dropping and rendering stalls.

1. The `vsync` setting now accepts `auto` (default) and `yield` settings,
   in addition to the existing on off settings.

    **Auto** only enables enable vsync when fullscreen and using a VRR display that performs frame-interpolation, because these displays absolutely need vsync enabled to lock onto the content. (see code comments for more detail).

    **Yield** simply lets the host OS and display drivers control the vsync state.

2. VFR (Variable Frame Rate) modes now always back-fill with the last presented frame if the DOS side hasn't sent a new frame in some time. This prevents the frame rate from falling below low-single digit thresholds common on modern VRR displays. It also fills in any gaps caused by dropped or torn frames on the host side.

3. The special handling for text modes is now unnecessary, as the risk of missing the last-presented frame in both CFR (Constant Frame Rate) and VFR modes has been eliminated with both re-sending the last presented frame.

4. The logic for determining the vsync state has been simplified. 

    Previously, three pieces of information were considered:
    -    The user's preference (on/off/...).
    -    The reported vsync state obtained from SDL or OpenGL.
    -    The measured vsync state determined through presentation timing.

    For instance, in OpenGL, you could set vsync to "off" and receive a successful return code. Querying the vsync state would also report it as "off" with a successful return code. However, testing it by sending 10 frames back-to-back and measuring the duration would reveal that vsync is actually on and enforced, limiting the frame rate to 60 fps - exactly matching the host's display refresh rate.

---

- Fixes #2551

PR #2702, opened 2023-07-28

Kappa971:
Could someone please tell me what this means?
`:CONFIG_VSYNC`
```
  auto:      Limit vsync to beneficial cases, such as when using an
             interpolating VRR display in fullscreen (default).
```
I assume that by `interpolating VRR display` you mean a display with interpolation capabilities (like the `auto motion plus` feature on Samsung TVs) and variable refresh rate, but I don't understand the connection with vsync.
I'm sorry but I would like to understand.

### DIFF
--- a/contrib/resources/translations/it.lng
+++ b/contrib/resources/translations/it.lng
@@ -76,7 +76,7 @@ Abilita le decorazioni delle finestre in modalità finestra
 .
 :CONFIG_TRANSPARENCY
 Imposta la trasparenza della finestra di DOSBox Staging (predefinito 0).
-Da 0 (nessuna trasparenza) a 90 (alta trasparenza).
+Da 0 (nessuna trasparenza) a 90 (trasparenza elevata).
 .
 :CONFIG_MAX_RESOLUTION
 Questa impostazione è stata rinominata in viewport_resolution.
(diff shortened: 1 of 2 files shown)


johnnovak:
> Could someone please tell me what this means? `:CONFIG_VSYNC`
> 
> ```
>   auto:      Limit vsync to beneficial cases, such as when using an
>              interpolating VRR display in fullscreen (default).
> ```
> 
> I assume that by `interpolating VRR display` you mean a display with interpolation capabilities (like the `auto motion plus` feature on Samsung TVs) and variable refresh rate, but I don't understand the connection with vsync. I'm sorry but I would like to understand.

Look here @Kappa971:
https://github.com/dosbox-staging/dosbox-staging/pull/2552

Read the top description and maybe the very long comments thread if interested.

But ultimately you can translate it pretty much as it is, that's a succinct wording with enough detail. We don't want to make these descriptions too long; that will be the role of the manual (when it's ready).

Let me know when this is ready for merging.

Kappa971:
> Look here @Kappa971: #2552
> 
> Read the top description and maybe the very long comments thread if interested.

Thanks @johnnovak, here is some sort of explanation:
```
Auto only enables enable vsync when fullscreen and using a VRR display that performs frame-interpolation, because these displays absolutely need vsync enabled to lock onto the content. (see code comments for more detail).
```

> But ultimately you can translate it pretty much as it is, that's a succinct wording with enough detail. We don't want to make these descriptions too long; that will be the role of the manual (when it's ready).

Sure, that's right. The problem is that sometimes (often) short sentences in English don't translate well into Italian and need a few more words, otherwise the sentence doesn't make sense or is "ugly". Probably the reason is that they are extremely different languages even though English has many words similar to Italian or even the same words (probably taken directly from Italian or Latin).
P.S. I'm not a literate 😆

> Let me know when this is ready for merging.

You can merge this into the main branch.

johnnovak:
> The problem is that sometimes (often) short sentences in English don't translate well into Italian and need a few more words, otherwise the sentence doesn't make sense or is "ugly". Probably the reason is that they are extremely different languages even though English has many words similar to Italian or even the same words (probably taken directly from Italian or Latin).

I can certainly understand that as Hungarian is very similar in this regard (my native language). English tends to be a lot more compact; usually, you need a lot more words and more explicitness to translate the same sentence to Hungarian.

I'm glad the linked PR helped to clear things up, merging then! 🚀

johnnovak (approved):

PR #2703, opened 2023-07-29
Refactor command line arguments

### LINKED ISSUE #2121: Allow both `-` and `--` prefixes for all command line options
Currently, some command line options only work with a single dash (e.g. `-opencaptures`, `-noconsole`), while some others work with either a single or a double dash (e.g. `--editconf`/`-editconf`, `--list-glshaders`/`-list-glshaders`).

While there's good reason to keep the single dash variants working for backwards compatibility, the most consistent approach would be to simply allow both single and double dashes for all command line options (that's for example what the Nim compiler and the CLI parser in the standard library does; maybe that's not strictly GNU Command Line Interface standard compliant, but I don't think that really matters...)

Also, the current handling of both single and double dash variants for the same option is very manual; the command line parser needs to be refactored to handle the two variants transparently (but be careful because the parser is also used by some of the included DOS commands, so maybe the DOSBox CLI parser needs to be separated out).

weirddan455:
Fixes #2121

I didn't end up using `getopt`.  It seemed more trouble than it was worth and didn't fit in well with our existing `CommandLine` class.  It also has some oddities like behaving differently if certain environment variables are set such as `POSIXLY_CORRECT` and it mutates the `argv` argument that it takes in.  There's also the possibility that it behaves differently on different platforms.

Anyway, I think I've cleaned the mess up quite a bit.  All the argument get parsed into a struct in the global Config class on program start.  Then, whenever someone wants to check if an argument is set, they can just directly read the already parsed value.  That way we're not doing parsing stuff all over the code and it should be more maintainable.  If someone wants to add a new argument, it's just adding a variable to the struct and then add a single line to the `ParseArguments()` function.

This mostly removes all the `control->cmdline` calls.  The one place I left that in was Autoexec.  There's a bunch of custom parsing logic in there that I didn't want to touch.

### DIFF
--- a/include/control.h
+++ b/include/control.h
@@ -40,15 +40,44 @@ enum class Verbosity : int8_t {
 	High,          //   yes   |    yes       |
 };
 
+struct CommandLineArguments {
+	bool printconf;
+	bool noprimaryconf;
+	bool nolocalconf;
+	bool fullscreen;
+	bool list_glshaders;
+	bool version;
+	bool help;
+	bool eraseconf;
+	bool erasemapper;
+	bool noconsole;
+	bool startmapper;
+	bool exit;
+	bool securemode;
+	bool noautoexec;
+	std::string opencaptures;
+	std::string working_dir;
+	std::string lang;
+	std::string machine;
+	std::vector<std::string> conf;
+	std::vector<std::string> set;
+	std::optional<std::vector<std::string>> editconf;
+	std::optional<int> socket;
+};
+
 class Config {
 public:
 	CommandLine * cmdline = nullptr;
+	CommandLineArguments arguments = {};
+
 private:
 	std::deque<Section*> sectionlist = {};
 	Section_line overwritten_autoexec_section = {};
 	std::string overwritten_autoexec_conf = {};
 	void (*_start_function)(void) = nullptr;
 	bool secure_mode = false; // Sandbox mode
+	void ParseArguments();
+
 public:
 	std::vector<std::string> startup_params = {};
 	std::vector<std::string> configfiles = {};
@@ -61,6 +90,7 @@ public:
 		assert(cmdline);
 		startup_params.push_back(cmdline->GetFileName());
 		cmdline->FillVector(startup_params);
+		ParseArguments();
 	}
 
 	Config() = default;
--- a/include/programs.h
+++ b/include/programs.h
@@ -22,11 +22,12 @@
 
 #include "dosbox.h"
 
+#include "std_filesystem.h"
 #include <functional>
 #include <list>
 #include <memory>
+#include <optional>
 #include <string>
-#include "std_filesystem.h"
 
 #include "dos_inc.h"
 #include "help_util.h"
@@ -36,20 +37,23 @@
 
 class CommandLine {
 public:
-	CommandLine(int argc, char const *const argv[]);
+	CommandLine(int argc, const char* const argv[]);
 	CommandLine(std::string_view name, std::string_view cmdline);
 
 	const char *GetFileName() const { return file_name.c_str(); }
 
-	bool FindExist(char const * const name,bool remove=false);
-	bool FindInt(char const * const name,int & value,bool remove=false);
-	bool FindString(char const * const name,std::string & value,bool remove=false);
-	bool FindCommand(unsigned int which,std::string & value) const;
-	bool FindStringBegin(char const * const begin,std::string & value, bool remove=false);
-	bool FindStringRemain(char const * const name,std::string & value);
-	bool FindStringRemainBegin(char const *const name, std::string &value);
-	bool GetStringRemain(std::string & value);
-	int GetParameterFromList(const char* const params[], std::vector<std::string> & output);
+	bool FindExist(const char* const name, bool remove = false);
+	bool FindInt(const char* const name, int& value, bool remove = false);
+	bool FindString(const char* const name, std::string& value,
+	                bool remove = false);
+	bool FindCommand(unsigned int which, std::string& value) const;
+	bool FindStringBegin(const char* const begin, std::string& value,
+	                     bool remove = false);
+	bool FindStringRemain(const char* const name, std::string& value);
+	bool FindStringRemainBegin(const char* const name, std::string& value);
+	bool GetStringRemain(std::string& value);
+	int GetParameterFromList(const char* const params[],
+	                         std::vector<std::string>& output);
 	void FillVector(std::vector<std::string> & vector);
 	bool HasDirectory() const;
 	bool HasExecutableName() const;
@@ -58,6 +62,12 @@ public:
 	                   const std::list<std::string_view>& post_args) const;
 	void Shift(unsigned int amount = 1);
 	uint16_t Get_arglength();
+	bool FindRemoveBoolArgument(const std::string& name, char short_letter = 0);
+	std::string FindRemoveStringArgument(const std::string& name);
+	std::vector<std::string> FindRemoveVectorArgument(const std::string& name);
+	std::optional<std::vector<std::string>> FindRemoveOptionalArgument(
+	        const std::string& name);
+	std::optional<int> FindRemoveIntArgument(const std::string& name);
 
 private:
 	using cmd_it = std::list<std::string>::iterator;
@@ -65,7 +75,10 @@ private:
 	std::list<std::string> cmds = {};
 	std::string file_name = "";
 
-	bool FindEntry(char const * const name,cmd_it & it,bool neednext=false);
+	bool FindEntry(const char* const name, cmd_it& it, bool neednext = false);
+	std::string FindRemoveSingleString(const char* name);
+	bool FindBoolArgument(const std::string& name, bool remove,
+	                      char short_letter = 0);
 };
 
 class Program {
@@ -104,7 +117,7 @@ protected:
 
 using PROGRAMS_Creator = std::function<std::unique_ptr<Program>()>;
 void PROGRAMS_Destroy([[maybe_unused]] Section* sec);
-void PROGRAMS_MakeFile(char const * const name, PROGRAMS_Creator creator);
+void PROGRAMS_MakeFile(const char* const name, PROGRAMS_Creator creator);
 
 template<class P>
 std::unique_ptr<Program> ProgramCreate() {
--- a/src/gui/gui_msgs.h
+++ b/src/gui/gui_msgs.h
@@ -35,50 +35,69 @@ There is NO WARRANTY, to the extent permitted by law.
 constexpr char help_msg[] =
         R"(Usage: dosbox [OPTION]... [FILE]
 
-List of common options:
+List of available options:
 
-  --printconf          Print the location of the default configuration file.
+  --conf <configfile>      Start with the options specified in <configfile>.
+                           Multiple configfiles can be specified.
+                           Example: --conf conf1.conf --conf conf2.conf
 
-  --editconf           Open the default configuration file in a text editor.
+  --printconf              Print the location of the default configuration file.
+                           If it does not exist, create it.
 
-  -c <command>         Run the specified DOS command before running FILE.
-                       Multiple commands can be specified.
+  --editconf               Open the default configuration file in a text editor.
 
-  -noautoexec          Don't perform any [autoexec] actions.
+  --eraseconf              Delete the default configuration file.
 
-  -noprimaryconf       Don't read settings from the primary configuration file
-                       located in your user folder.
+  --noprimaryconf          Don't read settings from the default configuration file
+                           located in your user folder.
 
-  -nolocalconf         Don't read settings from "dosbox.conf" if present in
-                       the current working directory.
+  --nolocalconf            Don't read settings from "dosbox.conf" if present in
+                           the current working directory.
 
-  -conf <configfile>   Start with the options specified in <configfile>.
-                       Multiple configfiles can be specified.
+  --set <property>=<value> Set a configuration property.
+                           Example: --set mididevice=fluidsynth --set soundfont=mysoundfont.sf2
 
-  --working-dir <path> Set working directory to <path>. DOSBox will act as if
-                       started from this directory.
+  --working-dir <path>     Set working directory to <path>. DOSBox will act as if
+                           started from this directory.
 
-  -fullscreen          Start in fullscreen mode.
+  --list-glshaders         List available GLSL shaders and their directories.
+                           Results are useable in the "glshader = " config setting.
 
-  -lang <langfile>     Start with the language specified in <langfile>.
+  --fullscreen             Start in fullscreen mode.
 
-  --list-glshaders     List available GLSL shaders and their directories.
-                       Results are useable in the "glshader = " config setting.
+  --lang <langfile>        Start with the language specified in <langfile>.
 
-  -machine <type>      Emulate a specific type of machine. The machine type has
-                       influence on both the emulated video and sound cards.
-                       Valid choices are: hercules, cga, cga_mono, tandy,
-                       pcjr, ega, svga_s3 (default), svga_et3000, svga_et4000,
-                       svga_paradise, vesa_nolfb, vesa_oldvbe.
+  --machine <type>         Emulate a specific type of machine. The machine type has
+                           influence on both the emulated video and sound cards.
+                           Valid choices are: hercules, cga, cga_mono, tandy,
+                           pcjr, ega, svga_s3 (default), svga_et3000, svga_et4000,
+                           svga_paradise, vesa_nolfb, vesa_oldvbe.
 
-  -exit                Exit after the DOS program specified by FILE has ended.
+  -c <command>             Run the specified DOS command before running FILE.
+                           Multiple commands can be specified.
 
-  -h, --help           Print this help message and exit.
+  --noautoexec             Don't execute any [autoexec] actions.
 
-  -v, --version        Print version information and exit.
+  --exit                   Exit after the DOS program specified by FILE has ended.
+                           If no FILE has been specified, execute [autoexcec] and
+                           terminate.
 
-You can find full list of options in the man page: dosbox(1)
-And in the file: /usr/share/doc/dosbox-staging/README
+  --startmapper            Run the mapper GUI.
+
+  --erasemapper            Delete the default mapper file.
+
+  --securemode             Enable secure mode. The [config] and [autoexec] sections
+                           of the loaded configuration file will be skipped,
+                           and commands such as MOUNT and IMGMOUNT are disabled.
+
+  --socket <num>           Run nullmodem on the specified socket number.
+
+  --opencaptures           Call program with as first parameter the location of the
+                           captures folder.
+
+  -h, --help               Print this help message and exit.
+
+  -v, --version            Print version information and exit.
 )";
 
 #endif
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3729,7 +3729,8 @@ static void GUI_StartUp(Section *sec)
 	sdl.resizing_window = false;
 	sdl.wait_on_error = section->Get_bool("waitonerror");
 
-	sdl.desktop.fullscreen = control->cmdline->FindExist("-fullscreen", true) || section->Get_bool("fullscreen");
+	sdl.desktop.fullscreen = control->arguments.fullscreen ||
+	                         section->Get_bool("fullscreen");
 
 	auto priority_conf = section->GetMultiVal("priority")->GetSection();
 	SetPriorityLevels(priority_conf->Get_string("active"),
@@ -4613,14 +4614,13 @@ static int LaunchEditor()
 		execlp(prog.c_str(), prog.c_str(), path.c_str(), (char *)nullptr);
 	};
 
-	std::string editor;
-	constexpr bool remove_arg = true;
 	// Loop until one succeeds
-	while (control->cmdline->FindString("--editconf", editor, remove_arg))
-		replace_with_process(editor);
-
-	while (control->cmdline->FindString("-editconf", editor, remove_arg))
-		replace_with_process(editor);
+	const auto arguments = &control->arguments;
+	if (arguments->editconf) {
+		for (const auto& editor : *arguments->editconf) {
+			replace_with_process(editor);
+		}
+	}
 
 	const char *env_editor = getenv("EDITOR");
 	if (env_editor)
@@ -4785,17 +4785,14 @@ int sdl_main(int argc, char *argv[])
 {
 	CommandLine com_line(argc, argv);
 	control = std::make_unique<Config>(&com_line);
+	const auto arguments = &control->arguments;
 
-	if (control->cmdline->FindExist("--version") ||
-	    control->cmdline->FindExist("-version") ||
-	    control->cmdline->FindExist("-v")) {
+	if (arguments->version) {
 		printf(version_msg, CANONICAL_PROJECT_NAME, DOSBOX_GetDetailedVersion());
 		return 0;
 	}
 
-	if (control->cmdline->FindExist("--help") ||
-	    control->cmdline->FindExist("-help") ||
-	    control->cmdline->FindExist("-h")) {
+	if (arguments->help) {
 		printf(help_msg); // -V618
 		return 0;
 	}
@@ -4819,15 +4816,12 @@ int sdl_main(int argc, char *argv[])
 
 	int rcode = 0; // assume good until proven otherwise
 	try {
-		std::string working_dir;
-		constexpr bool remove_arg = true;
-		if (control->cmdline->FindString("--working-dir", working_dir, remove_arg) ||
-		    control->cmdline->FindString("-working-dir", working_dir, remove_arg)) {
+		if (!arguments->working_dir.empty()) {
 			std::error_code ec;
-			std_fs::current_path(working_dir, ec);
+			std_fs::current_path(arguments->working_dir, ec);
 			if (ec) {
 				LOG_ERR("Cannot set working directory to %s",
-				        working_dir.c_str());
+				        arguments->working_dir.c_str());
 			}
 		}
 
@@ -4840,22 +4834,24 @@ int sdl_main(int argc, char *argv[])
 		config_add_sdl();
 		DOSBOX_Init();
 
-		if (control->cmdline->FindExist("--editconf") ||
-		    control->cmdline->FindExist("-editconf")) {
+		if (arguments->editconf) {
 			const int err = LaunchEditor();
 			return err;
 		}
 
-		std::string editor;
-		if(control->cmdline->FindString("-opencaptures",editor,true)) launchcaptures(editor);
-		if(control->cmdline->FindExist("-eraseconf")) eraseconfigfile();
-		if(control->cmdline->FindExist("-resetconf")) eraseconfigfile();
-		if(control->cmdline->FindExist("-erasemapper")) erasemapperfile();
-		if(control->cmdline->FindExist("-resetmapper")) erasemapperfile();
+		if (!arguments->opencaptures.empty()) {
+			launchcaptures(arguments->opencaptures);
+		}
+		if (arguments->eraseconf) {
+			eraseconfigfile();
+		}
+		if (arguments->erasemapper) {
+			erasemapperfile();
+		}
 
 		/* Can't disable the console with debugger enabled */
 #if defined(WIN32) && !(C_DEBUG)
-		if (control->cmdline->FindExist("-noconsole")) {
+		if (arguments->noconsole) {
 			FreeConsole();
 			/* Redirect standard input and standard output */
 			if(freopen(STDOUT_FILE, "w", stdout) == NULL)
@@ -4876,14 +4872,12 @@ int sdl_main(int argc, char *argv[])
 		}
 #endif  //defined(WIN32) && !(C_DEBUG)
 
-		if (control->cmdline->FindExist("--printconf") ||
-		    control->cmdline->FindExist("-printconf")) {
+		if (arguments->printconf) {
 			const int err = PrintConfigLocation();
 			return err;
 		}
 
-		if (control->cmdline->FindExist("--list-glshaders") ||
-		    control->cmdline->FindExist("-list-glshaders")) {
+		if (arguments->list_glshaders) {
 			ListGlShaders();
 			return 0;
 		}
@@ -4926,8 +4920,7 @@ int sdl_main(int argc, char *argv[])
 		"Usage: [color=green]config [/reset]-set [color=cyan][SECTION][/reset] "
 		"[color=white]PROPERTY[/reset][=][color=white]VALUE[/reset]\n");
 
-	std::string line;
-	while (control->cmdline->FindString("-set", line, true)) {
+	for (auto line : arguments->set) {
 		trim(line);
 		if (line.empty()
 		    || line[0] == '%' || line[0] == '\0'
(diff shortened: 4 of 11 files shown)


johnnovak:
Exciting stuff and long overdue! I'm so glad you've done this @weirddan455, it will be a very nice enhancement for the upcoming release! 🏅 I will review this in detail later today.

> I didn't end up using `getopt`. It seemed more trouble than it was worth and didn't fit in well with our existing `CommandLine` class. It also has some oddities like behaving differently if certain environment variables are set such as `POSIXLY_CORRECT` and it mutates the `argv` argument that it takes in. There's also the possibility that it behaves differently on different platforms.

I think you've done the right thing. I did not research `getopt` in great detail, but definitely the simplest solution that does the job is the correct choice ✅ 

> Anyway, I think I've cleaned the mess up quite a bit. All the argument get parsed into a struct in the global Config class on program start. Then, whenever someone wants to check if an argument is set, they can just directly read the already parsed value. That way we're not doing parsing stuff all over the code and it should be more maintainable. If someone wants to add a new argument, it's just adding a variable to the struct and then add a single line to the `ParseArguments()` function.

Sounds awesome! Parsing stuff in random places all over the codebase was utter crap, yeah... So glad that has been cleaned up!

> This mostly removes all the `control->cmdline` calls. The one place I left that in was Autoexec. There's a bunch of custom parsing logic in there that I didn't want to touch.

Good stuff!

include/control.h
```diff
@@ -40,15 +40,42 @@ enum class Verbosity : int8_t {
...
+	bool startmapper;
+	bool exit;
+	std::string opencaptures;
+	std::string working_dir;
+	std::string lang;
+	std::string machine;
+	std::vector<std::string> conf;
+	std::vector<std::string> set;
+	std::optional<std::vector<std::string>> editconf;
+	std::optional<int> socket;
```
johnnovak, line 63: It's a good idea to keep using the command line switch names here for clarity 👍🏻

src/gui/sdlmain.cpp
```diff
@@ -5004,7 +4995,7 @@ int sdl_main(int argc, char *argv[])
 			Sleep(5000);
 #endif
 		}
-	} catch (const std::exception &e) {
+	} catch (const std::exception& e) {
 		// catch all exceptions that derive from the standard library
 		LOG_ERR("EXCEPTION: Standard library exception: %s", e.what());
 		rcode = 1;
```
johnnovak, line 5001: Everything reads better and is better organise this way, well done 🎉

include/programs.h
```diff
@@ -58,14 +61,22 @@ class CommandLine {
 	                   const std::list<std::string_view>& post_args) const;
 	void Shift(unsigned int amount = 1);
 	uint16_t Get_arglength();
+	bool FindBoolArgument(const std::string& name, bool remove,
```
johnnovak, line 64: I'd introduce two version of this for consistency with the other functions, and extra boolean args are best to be avoided:

- `FindRemoveBoolArgument(const std::string& name, char short_letter = 0)`
- `FindBoolArgument(const std::string& name, char short_letter = 0)`

src/misc/setup.cpp
```diff
@@ -1789,3 +1790,35 @@ const char* SetProp(std::vector<std::string>& pvars)
...
+	arguments.startmapper = cmdline->FindBoolArgument("startmapper", true);
+	arguments.exit        = cmdline->FindBoolArgument("exit", true);
+
+	arguments.eraseconf = cmdline->FindBoolArgument("eraseconf", true) ||
+	                      cmdline->FindBoolArgument("resetconf", true);
+	arguments.erasemapper = cmdline->FindBoolArgument("erasemapper", true) ||
+	                        cmdline->FindBoolArgument("resetmapper", true);
+
+	arguments.version = cmdline->FindBoolArgument("version", true, 'v');
+	arguments.help    = cmdline->FindBoolArgument("help", true, 'h');
```
johnnovak, line 1811: As I mentioned above, `FindRemoveBoolArgument("help", 'h')` etc. would be much nicer. Second best would be:

```cpp
constexpr auto remove_arg = true;
...
FindBoolArgument("help", remove_arg, 'h');
...
```

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+                                   char short_letter)
+{
+	const std::string double_dash = "--" + name;
+	const std::string dash        = '-' + name;
+	char short_name[3]            = {};
+	short_name[0]                 = '-';
+	short_name[1]                 = short_letter;
+	return FindExist(double_dash.c_str(), remove) ||
+	       FindExist(dash.c_str(), remove) ||
+	       (short_letter && FindExist(short_name, true));
```
johnnovak, line 388: `FindExist(short_name, true)` — what is `true`? Suggest extracting it to a const with descriptive name.

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+}
+
+std::string CommandLine::StringArgHelper(const char* name)
+{
+	cmd_it it;
+	while (FindEntry(name, it, true)) {
```
johnnovak, line 394: Extract `true`.

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+	const std::string dash        = '-' + name;
+	char short_name[3]            = {};
+	short_name[0]                 = '-';
+	short_name[1]                 = short_letter;
+	return FindExist(double_dash.c_str(), remove) ||
+	       FindExist(dash.c_str(), remove) ||
+	       (short_letter && FindExist(short_name, true));
+}
+
+std::string CommandLine::StringArgHelper(const char* name)
```
johnnovak, line 391: How does it help exactly? 😄 Can you try to make the name reflect what the function is actually doing? "Helper", "Util", etc. can be literally anything.

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+	}
+	return {};
+}
+
+std::string CommandLine::FindRemoveStringArgument(const std::string& name)
+{
+	const std::string double_dash = "--" + name;
+	const std::string dash        = '-' + name;
+
+	std::string ret = StringArgHelper(double_dash.c_str());
```
johnnovak, line 415: Yeah, I'm not sure what the help is 😆

johnnovak: `const some_better_name = ...`

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+	if (ret.empty()) {
+		ret = StringArgHelper(dash.c_str());
```
johnnovak, line 417: Just a minor style comment: return from here, then you don't need to mutate `ret`.

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+	}
+	return ret;
+}
+
+std::vector<std::string> CommandLine::FindRemoveVectorArgument(const std::string& name)
+{
+	std::vector<std::string> ret = {};
+	while (1) {
```
johnnovak, line 425: Minor: i `for (;;)` is pretty standard, or even `while (true)`. Kinda dislike there's a number there... weird, i know 😄

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+	std::string ret = StringArgHelper(double_dash.c_str());
+	if (ret.empty()) {
+		ret = StringArgHelper(dash.c_str());
+	}
+	return ret;
+}
+
+std::vector<std::string> CommandLine::FindRemoveVectorArgument(const std::string& name)
+{
+	std::vector<std::string> ret = {};
```
johnnovak, line 424: Maybe use it `arg`. I don't really like the generic `ret` name in functions.

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+		}
+		ret.emplace_back(std::move(str));
+	}
+	return ret;
+}
+
+std::optional<std::vector<std::string>> CommandLine::FindRemoveOptionalArgument(
+        const std::string& name)
+{
+	if (!FindBoolArgument(name, false)) {
```
johnnovak, line 438: What's false?

johnnovak:
Very nice @weirddan455. Gave it some quick testing, works well.

Next thing would be updating the actual help text in `gui_msgs.h` so it's consistent, lists all the supported command line options (not just the "common options"...), and it's not a weird mismash of single and double-dashes 🤦🏻

All you up for the task? 😎

src/shell/command_line.cpp
```diff
@@ -374,3 +374,74 @@ void CommandLine::Shift(unsigned int amount)
...
+	short_name[0]                 = '-';
+	short_name[1]                 = short_letter;
+	return FindExist(double_dash.c_str(), remove) ||
+	       FindExist(dash.c_str(), remove) ||
+	       (short_letter && FindExist(short_name, true));
+}
+
+std::string CommandLine::StringArgHelper(const char* name)
+{
+	cmd_it it;
```
johnnovak, line 393: Best to init with `{}`, just to be consistent.

include/programs.h
weirddan455: Fixed

src/misc/setup.cpp
weirddan455: I did that 2nd approach just to avoid duplicated code.  I made `FindBoolArgument` a private method because it was only used in once place inside the `CommandLine` class.

src/shell/command_line.cpp
weirddan455: `true` was a bug in this case.  It was supposed to be the `remove` flag.  Fixed.

src/shell/command_line.cpp
weirddan455: Fixed

src/shell/command_line.cpp
weirddan455: Renamed to `FindRemoveSingleString`

src/shell/command_line.cpp
weirddan455: Renamed to `FindRemoveSingleString`

src/shell/command_line.cpp
weirddan455: Fixed

src/shell/command_line.cpp
weirddan455: Fixed

src/shell/command_line.cpp
weirddan455: Extracted the value to `remove_arg`

src/shell/command_line.cpp
johnnovak: Yeah in multi-arg functions it's almost never apparent what something like this means: `foo(blah, 4, 0, true, 233, false, false)` 🥶

johnnovak:
This is looking good now @weirddan455 . So how about updating the actual help text in `gui_msgs.h`?

weirddan455:
> This is looking good now @weirddan455 . So how about updating the actual help text in `gui_msgs.h`?

Sure, I'll get to that in a little bit.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
-  -noautoexec          Don't perform any [autoexec] actions.
+  --erasemapper            Delete the mapper file.
 
-  -noprimaryconf       Don't read settings from the primary configuration file
-                       located in your user folder.
+  --startmapper            Run the mapper GUI.
 
-  -nolocalconf         Don't read settings from "dosbox.conf" if present in
-                       the current working directory.
+  --securemode             Start in secure mode.
```
weirddan455, line 50: I don't know what secure mode does exactly.  Can you help with a description here?

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
 
-  -conf <configfile>   Start with the options specified in <configfile>.
-                       Multiple configfiles can be specified.
+  --noautoexec             Ignore autoexec section in configuration file.
 
-  --working-dir <path> Set working directory to <path>. DOSBox will act as if
-                       started from this directory.
+  --opencaptures           Launches some external program (I don't even know...)
```
weirddan455, line 54: This runs the following function (note the TODO comment).  It looks like the user is supposed to specify an external program they want to run against the captures directory.  Not sure what the purpose of this is.

https://github.com/dosbox-staging/dosbox-staging/blob/1930f7f3cd92d2615b5e65059c2ac7350ca85475/src/gui/sdlmain.cpp#L4683-L4706

weirddan455:
I updated the help with what I could.  Need some help on descriptions for the two above arguments.  I set them all to have double-dashes for long arguments and single dashes for single letter arguments, which is kind of standard.  We support both variants for either type of argument but I think duplicating that in help text is wasted space IMO (could change if you have strong feelings on it).

I also extracted a couple of arguments from `autoexec.cpp`

johnnovak:
> We support both variants for either type of argument but I think duplicating that in help text is wasted space IMO (could change if you have strong feelings on it).

Oh yeah, single dashes for long options are only for backward compatibility, but their usage should not be encouraged going forward.

I'll help with those two options later.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
 
-  --printconf          Print the location of the default configuration file.
+  --printconf              Print the location of the default configuration file.
 
-  --editconf           Open the default configuration file in a text editor.
+  --editconf               Open the default configuration file in a text editor.
 
-  -c <command>         Run the specified DOS command before running FILE.
-                       Multiple commands can be specified.
+  --eraseconf              Delete the configuration file.
```
johnnovak, line 44: We need to specify which one (default or local).

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
 
-  -noautoexec          Don't perform any [autoexec] actions.
+  --erasemapper            Delete the mapper file.
```
johnnovak, line 46: I assume the one defined in the default config? Or whatever `mapperfile` points to after loading the config? Can we make this clearer, I'd be afraid to use it for the first time going by this description.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
-  --working-dir <path> Set working directory to <path>. DOSBox will act as if
-                       started from this directory.
+  --opencaptures           Launches some external program (I don't even know...)
 
-  -fullscreen          Start in fullscreen mode.
+  --set <property>=<value> Sets a configuration property.
+                           Example: --set mididevice=fluidsynth
 
-  -lang <langfile>     Start with the language specified in <langfile>.
+  --socket <num>           Runs nullmodem on the specified socket number
```
johnnovak, line 59: You're using imperative verbs consistently everywhere else (which is good), so this should be "run".

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
-  -conf <configfile>   Start with the options specified in <configfile>.
-                       Multiple configfiles can be specified.
+  --noautoexec             Ignore autoexec section in configuration file.
 
-  --working-dir <path> Set working directory to <path>. DOSBox will act as if
-                       started from this directory.
+  --opencaptures           Launches some external program (I don't even know...)
 
-  -fullscreen          Start in fullscreen mode.
+  --set <property>=<value> Sets a configuration property.
```
johnnovak, line 56: "Set" (not "sets").

johnnovak: We should mention multiple `--set` options can be provided.

src/gui/gui_msgs.h
johnnovak: Please put a period at the end of every config description.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
-                       Results are useable in the "glshader = " config setting.
+  -c <command>             Run the specified DOS command before running FILE.
+                           Multiple commands can be specified.
 
-  -machine <type>      Emulate a specific type of machine. The machine type has
-                       influence on both the emulated video and sound cards.
-                       Valid choices are: hercules, cga, cga_mono, tandy,
-                       pcjr, ega, svga_s3 (default), svga_et3000, svga_et4000,
-                       svga_paradise, vesa_nolfb, vesa_oldvbe.
+  --noautoexec             Don't perform any [autoexec] actions.
```
johnnovak, line 64: "Don't execute" would be better.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
+                           influence on both the emulated video and sound cards.
+                           Valid choices are: hercules, cga, cga_mono, tandy,
+                           pcjr, ega, svga_s3 (default), svga_et3000, svga_et4000,
+                           svga_paradise, vesa_nolfb, vesa_oldvbe.
+
+  --exit                   Exit after the DOS program specified by FILE has ended.
+
+  -h, --help               Print this help message and exit.
+
+  -v, --version            Print version information and exit.
```
johnnovak, line 95: Can we remove the two lines below? The man page is either going away or it will contain the same info, same for the README, plus that path is distro/installations specific. Better to just rid of them.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
 
 List of common options:
 
-  --printconf          Print the location of the default configuration file.
+  --printconf              Print the location of the default configuration file.
```
johnnovak, line 40: Change the above to "Available options" or something similar (we are aiming to list everything here).

src/gui/gui_msgs.h
weirddan455: Well, each one needs to be prefixed with `--set`.  Although it kind of works by accident if you don't.  For example, `--set mididevice=fluidsynth soundfont=font.sf2`.  `soundfont=font.sf2` gets sent to DOS which can also set these options.

src/gui/gui_msgs.h
johnnovak: From the DOSBox X manual:

> Enables DOSBox-X’s secure mode. The [config] and [autoexec] sections of the loaded configuration file will be skipped, and commands such as MOUNT and IMGMOUNT are disabled.

https://github.com/joncampbell123/dosbox-x/wiki/DOSBox%E2%80%90X%E2%80%99s-Command%E2%80%90Line-Options

You can use this with a few changes.

src/gui/gui_msgs.h
weirddan455: It does `const auto path = (get_platform_config_dir() / file).string();` so the default one.  I'll specify.

src/gui/gui_msgs.h
johnnovak: It' a bizarre feature... Let's just say the same thing as the X manual:

> Calls program with as first parameter the location of the captures folder.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
+  --list-glshaders         List available GLSL shaders and their directories.
+                           Results are useable in the "glshader = " config setting.
+
+  --machine <type>         Emulate a specific type of machine. The machine type has
+                           influence on both the emulated video and sound cards.
+                           Valid choices are: hercules, cga, cga_mono, tandy,
+                           pcjr, ega, svga_s3 (default), svga_et3000, svga_et4000,
+                           svga_paradise, vesa_nolfb, vesa_oldvbe.
+
+  --exit                   Exit after the DOS program specified by FILE has ended.
```
johnnovak, line 91: Doesn't this work without specifying FILE, in which case DOSBox exits after executing the autoexec sections? Can you test it out and modify accordingly?

src/gui/gui_msgs.h
johnnovak: Also, you've listed this twice; delete the previous one.

src/gui/gui_msgs.h
```diff
@@ -37,45 +37,62 @@ constexpr char help_msg[] =
...
-  -exit                Exit after the DOS program specified by FILE has ended.
+  --noprimaryconf          Don't read settings from the primary configuration file
+                           located in your user folder.
 
-  -h, --help           Print this help message and exit.
+  --nolocalconf            Don't read settings from "dosbox.conf" if present in
+                           the current working directory.
 
-  -v, --version        Print version information and exit.
+  --conf <configfile>      Start with the options specified in <configfile>.
```
johnnovak, line 72: Would better to say multiple configs can be specified; by multiple args to a single `--conf` option, or by multiple `--conf` options? Can you test it out and clarify?

weirddan455: It needs a `--conf` before each file.

src/gui/gui_msgs.h
weirddan455: This one is also the default `const auto path = (get_platform_config_dir() / MAPPERFILE).string();`

weirddan455:
Does that look good?  I'll squash it down before you merge.  Maybe 2 commits, one for the refactoring and one for the help changes?

I usually like to do the squash at the end to make reviewing easier.

src/gui/gui_msgs.h
weirddan455: Specified "default"

src/gui/gui_msgs.h
weirddan455: Good stuff.  I removed the DOSBox-X name and used that.

src/gui/gui_msgs.h
weirddan455: Sounds good.

src/gui/gui_msgs.h
johnnovak: About `--printconf`, hold on to your hat! This also _writes_ the default config if it does not exist 😐 Well, that's how it works, so we need to document it...

src/gui/gui_msgs.h
weirddan455: Specified "default"

src/gui/gui_msgs.h
weirddan455: Fixed

src/gui/gui_msgs.h
weirddan455: I added multiples to the example to clarify that.

src/gui/gui_msgs.h
weirddan455: Fixed

src/gui/gui_msgs.h
weirddan455: Yeah, I was eyeing those lines suspiciously.  Removed.

src/gui/gui_msgs.h
weirddan455: Changed to "available"

src/gui/gui_msgs.h
weirddan455: That is correct.  Added a clarifying line.

src/gui/gui_msgs.h
weirddan455: Added an example to clarify how to set multiple conf files.

src/gui/gui_msgs.h
weirddan455: > About `--printconf`, hold on to your hat! This also _writes_ the default config if it does not exist neutral_face Well, that's how it works, so we need to document it...

So it does.  That's unexpected.  Documented.

johnnovak:
> Does that look good? I'll squash it down before you merge. Maybe 2 commits, one for the refactoring and one for the help changes?
>
> I usually like to do the squash at the end to make reviewing easier.

Thanks for addressing the comments so quickly @weirddan455, it all looks excellent now. 🎉 Great job! 🚀 

Suggest squashing this into 3 commits: refactoring, actual code changes, and updating the help text.

---

And one last suggestion, it would be better to group the options loosely by function and list them in rough order of importance. It's somewhat subjective, but more frequently used options should come earlier.

This is my stab at it; I think it's an improvement. if you agree with it, you can basically copy/paste it (it contains no tabs, only spaces). I've also wrapped a few long lines to make them a bit shorter.

```
Usage: dosbox [OPTION]... [FILE]

List of available options:

  --conf <configfile>      Start with the options specified in <configfile>.
                           Multiple configfiles can be specified.
                           Example: --conf conf1.conf --conf conf2.conf

  --printconf              Print the location of the default configuration file.
                           If it does not exist, create it.

  --editconf               Open the default configuration file in a text editor.

  --eraseconf              Delete the default configuration file.

  --noprimaryconf          Don't read settings from the primary configuration file
                           located in your user folder.

  --nolocalconf            Don't read settings from "dosbox.conf" if present in
                           the current working directory.

  --set <property>=<value> Set a configuration property.
                           Example: --set mididevice=fluidsynth --set soundfont=mysoundfont.sf2

  --working-dir <path>     Set working directory to <path>. DOSBox will act as if
                           started from this directory.

  --list-glshaders         List available GLSL shaders and their directories.
                           Results are useable in the "glshader = " config setting.

  --fullscreen             Start in fullscreen mode.

  --lang <langfile>        Start with the language specified in <langfile>.

  --machine <type>         Emulate a specific type of machine. The machine type has
                           influence on both the emulated video and sound cards.
                           Valid choices are: hercules, cga, cga_mono, tandy,
                           pcjr, ega, svga_s3 (default), svga_et3000, svga_et4000,
                           svga_paradise, vesa_nolfb, vesa_oldvbe.

  -c <command>             Run the specified DOS command before running FILE.
                           Multiple commands can be specified.

  --noautoexec             Don't execute any [autoexec] actions.

  --exit                   Exit after the DOS program specified by FILE has ended.
                           If no FILE has been specified, execute [autoexcec] and
                           terminate.

  --startmapper            Run the mapper GUI.

  --erasemapper            Delete the default mapper file.

  --securemode             Enable secure mode. The [config] and [autoexec] sections
                           of the loaded configuration file will be skipped,
                           and commands such as MOUNT and IMGMOUNT are disabled.

  --socket <num>           Run nullmodem on the specified socket number.

  --opencaptures           Call program with as first parameter the location of the
                           captures folder.

  -h, --help               Print this help message and exit.

  -v, --version            Print version information and exit.
```

weirddan455:
That help text looks good to me.  Thanks for helping out with that :smile: 

>  Suggest squashing this into 3 commits: refactoring, actual code changes, and updating the help text.

I wasn't sure where to draw the line between refactoring and code changes.  I originally just had that as one commit.  I don't believe I changed any behavior except for accepting both single and double dash arguments everywhere and fixing a bug where some arguments were not removed leading to them getting passed to DOS.  So "actual code changes" should be minimal?

Anyway, I squashed it down to 2 commits and rebased to latest main to make sure there weren't any merge conflicts (there weren't :smile: ).  If you can point to where in that first commit should be split up, I can try and do that if you like.

weirddan455:
PVS is failing on main.  That was passing on this PR before I rebased.

FeralChild64:
> PVS is failing on main. That was passing on this PR before I rebased.

PR https://github.com/dosbox-staging/dosbox-staging/pull/2715 will fix it.

kcgen:
> PVS is failing on main. That was passing on this PR before I rebased.

:+1: - @johnnovak and @FeralChild64 are on it! (cleanup PR: https://github.com/dosbox-staging/dosbox-staging/pull/2715 will get back down [from 275 to 270 warnings](https://github.com/dosbox-staging/dosbox-staging/actions/workflows/pvs-studio.yml?query=branch%3Amain))

It looks like this branch's count here is 276, so I _think_ there's just one new issue lurking in the changes! if you can ferret it out (or wait and rebase once #2715 goes in)

weirddan455:
>  It looks like this branch's count here is 276, so I think there's just one new issue lurking in the changes! 

How can you tell what the issues actually are?  I see the breakdown of number of issues per file but I don't see what the warnings/errors actually are.

FeralChild64:
Click _Summary_ (left side), download `pvs-analysis-report` file, unpack it, open the report with a web browser.
Indeed, getting the actual warnings is not straightforward.

weirddan455:
> Click _Summary_ (left side), download `pvs-analysis-report` file, unpack it, open the report with a web browser. Indeed, getting the actual warnings is not straightforward.

Thanks.  I found this in a couple places in changes I made in this PR.  It's labeled micro-optimization.  `Decreased performance. Consider creating a reference to avoid using the 'control->arguments' expression repeatedly.`

weirddan455:
Looks like the Mac builds are having a fit now (unrelated to this PR):

```
 ==> Pouring openssl@3--3.1.2.ventura.bottle.tar.gz
Error: Could not symlink bin/openssl
Target /usr/local/bin/openssl
is a symlink belonging to openssl@1.1. You can unlink it:
  brew unlink openssl@1.1

  brew link --overwrite openssl@3

  brew link --overwrite --dry-run openssl@3
Error: Process completed with exit code 1.
```

OpenSSL 3.1.2 was just released yesterday.  These CI machines constantly updating packages and breaking is getting annoying...

weirddan455:
Anyway, I pushed a commit to address the PVS warnings.  It's down to 273 now so maybe I inadvertently fixed some other warnings along the way.  I can rebase once the other PR gets merged if you want the green check.

kcgen:
Nice @weirddan455 -> yup, suggest letting the other PR reconcile first then rebase to get the new totals.

(I'm guessing you've reduced the total even more, and you can tighten down the limit lower than 270 :partying_face: !).

johnnovak:
> That help text looks good to me. Thanks for helping out with that 😄

Teamwork, man 😎 

> Anyway, I squashed it down to 2 commits and rebased to latest main to make sure there weren't any merge conflicts (there weren't 😄 ). If you can point to where in that first commit should be split up, I can try and do that if you like.

Yeah, I haven't double-checked if separating the cleanup/reformatting makes sense or not in this. Sometimes in bigger or more sensitive jobs it helps to do the cleanup/reformat upfront (especially the reformat), so we can see in subsequent commits the actual changes. Not an issue in this case, so the two commits you create are fine (you can rebase the PVS studio stuff into the first one). Just use your judgment in the future.

johnnovak:
Main is supposed to be fixed, rebase away @weirddan455 🚀

weirddan455:
I just pushed a rebase.  These Mac builds are still failing with that OpenSSL error.  I definitely rebased to latest main that includes the fix @kcgen made so I'm not sure what's happening :cry:

kcgen:
> I just pushed a rebase. These Mac builds are still failing with that OpenSSL error. I definitely rebased to latest main that includes the fix @kcgen made so I'm not sure what's happening cry

This could be GitHub feeding us different VMs; this sometimes that happens when they migrate configurations. Trying a re-run right now; worst-case we'll forcefully remove openssl 1.x prior to the install.

kcgen:
Yup; still a problem. Unlink added. can you rebase against `main`?
(sorry for these CI troubles, @weirddan455 !)

weirddan455:
@kcgen I just rebased again... Mac builds still failing :sob:

kcgen:
Ok this is really confusing.

`main` is OK but this branch is still haven't problems even after rebasing.

kcgen:
@weirddan455 , ok, let's try even harder.

In your branch, try:

```
git cherry-pick d0accd729
```

Then push.  If it works, then we're good.  If not, you can drop/delete that commit.

weirddan455:
Everything is green now!  Thanks @kcgen for emerging victorious over these strange build system issues.

johnnovak:
Thanks for persevering @weirddan455 👍🏻

Grounded0:
Connected issues and PRs:

- #2837

Issue #2701, opened 2023-07-27
Fluidsynth sounds incorrect


x73rmin8r:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

DOSBOX-staging-v0.81.0-alpha-1152-g5d881

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

I think there may be a problem with the Fluidsynth implementation in Staging. I ran across a song (level 8 of Descent) that sounds noticeably wrong using Fluidsynth, but sounds correct when using the same soundfont file using the external midi player CoolSoft VirtualMIDISynth. Also DOSBOX-X using Fluidsynth sounds correct, so it might not be a fault with Fluidsynth itself.

Same behavior in alpha 1152 and release version of Staging. Tried with the Fluidsynth reverb and chorus settings off and on.

[descent-fluidsynth-error.zip](https://github.com/dosbox-staging/dosbox-staging/files/12179354/descent-fluidsynth-error.zip)

The sound recorder didn't capture the music using the external midi player, but to my ear the output I got when using VirtualMIDISynth and Fluidsynth in DOSBOX-X sounded essentially the same as an AWE32 card (1mgm.sf2 is from a dump of the AWE32 ROM) (https://www.youtube.com/watch?v=avrvEVQgdaM). There's possibly some differences in the reverb/chorus effects, but the actual melody starting at about 0:16 in my recording was what I noticed being incorrect.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. set fluidsynth as the midi device with the 1mgm.sf2 soundfont
2. set Descent to use general midi 
3. start level 8 (use cheats to levelskip, during gameplay type "gabbagabbahey" to enable cheats, then "farmerjoe" to open level select menu

### Download URL of affected game or software

gog descent (level 8 isn't in shareware version unfortunately)

### Your configuration

```ini
https://gist.github.com/x73rmin8r/b4a9900ad3196cba7f27bc47823ddb53
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Thanks @x73rmin8r, I can reproduce the problem in DOSBox Staging with the Level 8 music.

My findings so far:

- I can reproduce it in the latest dev version of Staging with _any_ SoundFont on macOS and Windows
[...]

PR #2705, opened 2023-07-30
Do not send portamento messages (MIDI CC 5, 65 & 84) to FluidSynth

johnnovak:
## Overview

Roland SC-55 style Portamento cannot be emulated on FluidSynth, and out-of-tune sounding portamento is a much bigger problem than the lack of subtle portamento effects in a handful of synth-oriented game soundtracks.

This fixes the weirdly out of tune synth parts in the Level 8 music of Descent (see https://github.com/dosbox-staging/dosbox-staging/issues/2701).

## Detailed explanation

The Roland SC-55 and its clones (Yamaha MU80 or Roland's own later modules that emulate the SC-55) handle portamento (pitch glides between consecutive notes on the same channel) in a very specific and unique way, just like most synthesisers.

The SC-55 accepts only 7-bit Portamento Time values via MIDI CC5, where the min value of 0 sets the fastest portamento time (effectively turns it off), and the max value of 127 the slowest (up to 8 minutes!). There is an exponential mapping between the CC values and the duration of the portamento (pitch slides/glides); this custom curve is apparently approximated by multiple linear segments. Moreover, the distance between the source and destination notes also affect the portamento time, making portamento dynamic and highly dependent on the notes being played.

FluidSynth, on the other hand, implements a very different portamento model. Portament Time values are set via 14-bit CC messages (via MIDI CC5 (coarse) and CC37 (fine)), and there is a linear mapping between CC values and the portamento time as per the following formula:

```
  (CC5 * 127 ms) + (CC37 ms)
```

Because of these fundamental differences, emulating Roland SC-55 style portamento on FluidSynth is practically not possible. Music written for the SC-55 that use portamento sounds weirdly out of tune on FluidSynth (e.g. the Level 8 music of Descent), and "mapping" SC-55 portamento behaviour to the FluidSynth range is not possible due to dynamic nature of the SC-55 portamento handling. All in all, it's for the best to ignore portamento altogether. This is not a great loss as it's used rarely and usually only to add some subtle flair to the start of the notes in synth-oriented soundtracks.

This table and diagram illustrate the exponential mapping between the CC values and the Portamento Time. Note the below figures apply only for the two specific MIDI notes I used for my reverse engineering exercise; as the Portamento Time is dynamic, you'll get different exponential curves based on the distance between the two notes:

```
CC 5 value  Portamento time
----------  ---------------
     0          0.000 s
     1          0.006 s
     2          0.023 s
     4          0.050 s
     8          0.110 s
    16          0.250 s
    32          0.500 s
    64          2.060 s
    80          4.200 s
    96          8.400 s
   112         19.500 s
   116         26.700 s
   120         40.000 s
   124         80.000 s
   127        480.000 s
```

image

## Sound examples

These are MP3 files—download them then change the extension from ZIP to MP3.

### Roland Virtual Sound Canvas

This is the isolated synth part in question of the Level 8 music of Descent. This is how portamento sounds on the Roland Virtual Sound Canvas:
[scva-portamento-8.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208658/scva-portamento-8.zip)

Now with the portamento disabled. The difference is certainly subtle and it's a feel thing; note how there is less attack and "vigour" at the start of the notes. You can't really hear a pitch glide when portamento is on because it's fast, but it makes the notes "hit harder" and sound more dynamic and "dangerous" than with portamento off.
[scva-no-portamento.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208657/scva-no-portamento.zip)

Now increasing portamento substantially to 48 to mimic the FluidSynth "bendy notes" behaviour:
[scva-portamento-48.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208659/scva-portamento-48.zip)

And a simpler example where it's more obvious to hear the difference between portamento off and portamento 8. Pay attention to the start of the notes:
[scva-portamento-example-no-porta.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208661/scva-portamento-example-no-porta.zip)
[scva-portamento-example-8.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208660/scva-portamento-example-8.zip)

Another example, portamento 80 on a saw wave with notes spaced several octaves apart (that's a looooong 4.2s portamento!) 
[saw-portamento-80.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208662/saw-portamento-80.zip)

### FluidSynth

Isolated synth part as played by FluidSynth using the GeneralUser GS SoundFont when we let the portamento CC messages through. Yeah, quite horrible due to the very different portamento behaviour between the SC-55 and FluidSynth (not FluidSynth's "fault" though; these are two different synths and FS never aimed to be SC-55 compatible):
[fs-porta-coarse-8.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208675/fs-porta-coarse-8.zip)

With the portamento CC messages completely filtered out, this sounds very similar to the "no portamento" SCVA recording — this is the final solution:
[fs-no-porta.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208674/fs-no-porta.zip)

An attempt to mimic the SC-55 portamento behaviour by setting portamento coarse to 0 and fine to 60, which should achieve a 60ms portamento. This was done by ear as it's impossible to come up with a single figure as the portamento times are different for every note-pair on the SC-55. Still, it doesn't sound anything like the the SC-55; even manually it's impossible to come up with a single number that replicates the feel of the SC-55 behaviour _on this single synth line_. Coming up with a general algorithm that would decide the correct number on the fly for every musical score is practically impossible... and we're already verging into accurate SC-55 emulation territory 😅 Which is not possible on FluidSynth, as we well know—we just have to be content with a ~70-80% approximation of the real thing, at best.

[fs-porta-fine-60.zip](https://github.com/dosbox-staging/dosbox-staging/files/12208676/fs-porta-fine-60.zip)

### DIFF
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -663,10 +663,57 @@ void MidiHandlerFluidsynth::ApplyChannelMessage(const std::vector<uint8_t>& msg)
 
 	// clang-format off
 	switch (status) {
-	case MidiStatus::NoteOff:         fluid_synth_noteoff(         synth.get(), channel, msg[1]);                 break;
-	case MidiStatus::NoteOn:          fluid_synth_noteon(          synth.get(), channel, msg[1], msg[2]);         break;
-	case MidiStatus::PolyKeyPressure: fluid_synth_key_pressure(    synth.get(), channel, msg[1], msg[2]);         break;
-	case MidiStatus::ControlChange:   fluid_synth_cc(              synth.get(), channel, msg[1], msg[2]);         break;
+	case MidiStatus::NoteOff:         fluid_synth_noteoff(     synth.get(), channel, msg[1]);         break;
+	case MidiStatus::NoteOn:          fluid_synth_noteon(      synth.get(), channel, msg[1], msg[2]); break;
+	case MidiStatus::PolyKeyPressure: fluid_synth_key_pressure(synth.get(), channel, msg[1], msg[2]); break;
+
+	case MidiStatus::ControlChange: {
+		const auto controller = msg[1];
+		const auto value = msg[2];
+
+		if (controller == MidiController::Portamento ||
+			controller == MidiController::PortamentoTime ||
+			controller == MidiController::PortamentoControl) {
+
+			// The Roland SC-55 and its clones (Yamaha MU80 or Roland's own
+			// later modules that emulate the SC-55) handle portamento (pitch
+			// glides between consecutive notes on the same channel) in a very
+			// specific and unique way, just like most synthesisers.
+			//
+			// The SC-55 accepts only 7-bit Portamento Time values via MIDI
+			// CC5, where the min value of 0 sets the fastest portamento time
+			// (effectively turns it off), and the max value of 127 the
+			// slowest (up to 8 minutes!). There is an exponential mapping
+			// between the CC values and the duration of the portamento (pitch
+			// slides/glides); this custom curve is apparently approximated by
+			// multiple linear segments. Moreover, the distance between the
+			// source and destination notes also affect the portamento time,
+			// making portamento dynamic and highly dependent on the notes
+			// being played.
+			//
+			// FluidSynth, on the other hand, implements a very different
+			// portamento model. Portament Time values are set via 14-bit CC
+			// messages (via MIDI CC5 (coarse) and CC37 (fine)), and there is
+			// a linear mapping between CC values and the portamento time as
+			// per the following formula:
+			//
+			//   (CC5 * 127 ms) + (CC37 ms)
+			//
+			// Because of these fundamental differences, emulating Roland
+			// SC-55 style portamento on FluidSynth is practically not
+			// possible. Music written for the SC-55 that use portamento
+			// sounds weirdly out of tune on FluidSynth (e.g. the Level 8
+			// music of Descent), and "mapping" SC-55 portamento behaviour to
+			// the FluidSynth range is not possible due to dynamic nature of
+			// the SC-55 portamento handling. All in all, it's for the best to
+			// ignore portamento altogether. This is not a great loss as it's
+			// used rarely and usually only to add some subtle flair to the
+			// start of the notes in synth-oriented soundtracks.
+		} else {
+			fluid_synth_cc(synth.get(), channel, controller, value);
+		}
+	} break;
+
 	case MidiStatus::ProgramChange:   fluid_synth_program_change(  synth.get(), channel, msg[1]);                 break;
 	case MidiStatus::ChannelPressure: fluid_synth_channel_pressure(synth.get(), channel, msg[1]);                 break;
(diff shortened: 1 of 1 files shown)


johnnovak:
@dwhinham this might be relevant to your interests 😄

src/midi/midi_fluidsynth.cpp
```diff
@@ -663,10 +663,57 @@ void MidiHandlerFluidsynth::ApplyChannelMessage(const std::vector<uint8_t>& msg)
...
-	case MidiStatus::ControlChange:   fluid_synth_cc(              synth.get(), channel, msg[1], msg[2]);         break;
+	case MidiStatus::NoteOff:         fluid_synth_noteoff(     synth.get(), channel, msg[1]);         break;
+	case MidiStatus::NoteOn:          fluid_synth_noteon(      synth.get(), channel, msg[1], msg[2]); break;
+	case MidiStatus::PolyKeyPressure: fluid_synth_key_pressure(synth.get(), channel, msg[1], msg[2]); break;
+
+	case MidiStatus::ControlChange: {
+		const auto controller = msg[1];
+		const auto value = msg[2];
+
+		if (controller == MidiController::Portamento ||
```
kklobe, line 674: maybe avoid the empty block? 

```cpp
if (controller != MidiController::Portamento &&
	controller != MidiController::PortamentoTime &&
	controller != MidiController::PortamentoControl) {
        //...
	fluid_synth_cc(synth.get(), channel, controller, value);
}

Comment (kklobe) [inline src/midi/midi_fluidsynth.cpp:674]: maybe avoid the empty block? 

```cpp
if (controller != MidiController::Portamento &&
	controller != MidiController::PortamentoTime &&
	controller != MidiController::PortamentoControl) {
        //...
	fluid_synth_cc(synth.get(), channel, controller, value);
}

File: src/midi/midi_fluidsynth.cpp [hunk 2]
```diff
@@ -663,10 +663,57 @@ void MidiHandlerFluidsynth::ApplyChannelMessage(const std::vector<uint8_t>& msg)
...
+			controller == MidiController::PortamentoTime ||
+			controller == MidiController::PortamentoControl) {
+
```
kklobe, line 677: maybe avoid the empty block?

```cpp
if (controller != MidiController::Portamento &&
	controller != MidiController::PortamentoTime &&
	controller != MidiController::PortamentoControl) {
        // ...
	fluid_synth_cc(synth.get(), channel, controller, value);
}
```

johnnovak: Yeah, I guess I like it better this way as it more resembles a case or when block.

If blah, that's a NOP
Else do stuff

I have a feeling we'll need to add further special cases in the future too.

kklobe: definitely disagree not enough to force the issue!

kklobe (commented):
gave a good listen to all the samples and fully agree with this change.

johnnovak:
> gave a good listen to all the samples and fully agree with this change.

Thanks for the review, man!

PR #2708, opened 2023-07-30
Fix mapper screen for high dpi

### LINKED ISSUE #2706: The mapper is broken
## Problem

This is how the mapper looks like on macOS here right now (High DPI screen) on SDL 2.28.1. Was working yesterday, it's totally broken today... which is perplexing as SDL 2.28.1 came out like a month ago?

It's not just small, it's unusable... The graphics are not scaled, but the mouse input is... 🤦🏻 Bluddy on our Discord has reported the same issue on Windows a few days ago. To the SDL folks' credit, at least they broke things on all platforms at the same time! 🥶 

image

For reference this is how it should be like. That's what I get with v0.80.1 that uses SDL 2.26.1.

image

## Solution

Obviously, we just cannot release a new DOSBox version with a completely broken mapper on High DPI. That's non-negotiable.

We'll need to fix this for 2.28.1 _somehow_, then we need to rethink our relationship with 3rd party dependencies and reproducible builds. My very strong preferences are:

- Pin _all_ 3rd party library versions
- Everything should come from Meson, absolutely _no_ dependency on system-wide libraries
- Build should be 100% reproducible:
  - If I rebuild the project tomorrow, it should never fail just because someone has published a new version of their library. But that's the better outcome; the worse is when it _does_ build but things break in a subtle way...
  - Another use case: if I build a commit from 6 months ago, I want to get the _exact same build_ I got on that day, not something "similar" using the currently installed libraries on my system.
- Ultimately, we do not want library updates being forced on us whenever the 3rd party authors feel like it's time for releasing a new version. We want controlled dependency upgrades, when we want, at our own pace (and yes, we should definitely upgrade... but in this case, I would have much rather stayed on an earlier stable SDL2 version, then only upgrade when the guys have fixed their regressions. And who knows, maybe the current behaviour _is_ correct... but it's different behaviour, and perhaps our workarounds we put in earlier for a good reasons in the past no longer work. The point being: we can't just take new library versions and expect them to work flawlessly just because we can still compile the project; everything needs to be thoroughly tested before we say we're ready to take the new version. High DPI support in SDL2 seems to be an especially fragile and flaky thing).

kklobe:
This is a very quick and dirty hotfix for the high DPI mapper issue until we can do it properly.

Fixes #2706

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -1984,7 +1984,8 @@ static void SetActiveEvent(CEvent * event) {
 	}
 }
 
-extern SDL_Window * GFX_SetSDLSurfaceWindow(uint16_t width, uint16_t height);
+extern SDL_Window* GFX_SetSDLSurfaceWindow(uint16_t width, uint16_t height,
+                                           bool allow_highdpi = true);
 extern SDL_Rect GFX_GetSDLSurfaceSubwindowDims(uint16_t width, uint16_t height);
 extern void GFX_UpdateDisplayDimensions(int width, int height);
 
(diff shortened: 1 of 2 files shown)


johnnovak:
Whoa, that was quick! Thanks a lot for this @kklobe. I will test it shortly on my Mac.

johnnovak (approved):
Works like a charm @kklobe, merging! 🚀

PR #2710, opened 2023-07-31
Give Voodoo a gentle push to get it passing CI

kcgen:
This PR gets Voodoo passing CI, enables it by default, and makes the conf section consistent.

```ini
[voodoo]
voodoo_memsize = ...
voodoo_perf = ...
```

(This PR target CI fixes only - it's not a refactoring, renaming, or rewrite PR.  For anyone interested in a more full-scale overhaul, it's best to rebase against MAME's latest Voodoo 1 and 2 revision as it's been updated with more modern C++ styles).

---

push-voodoo-into-ci

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -122,7 +122,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 12
+          MAX_BUGS: 3
         run: |
           # summary
           echo "Full report is included in build Artifacts"
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 333
+          MAX_BUGS: 270
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -781,19 +781,24 @@ void DOSBOX_Init()
 	secprop->AddInitFunction(&DMA_Init);
 	secprop->AddInitFunction(&VGA_Init);
 	secprop->AddInitFunction(&KEYBOARD_Init);
+	secprop->AddInitFunction(&PCI_Init); // PCI bus
 
-	secprop=control->AddSection_prop("pci", &PCI_Init, false); // PCI bus
+	secprop = control->AddSection_prop("voodoo", &VOODOO_Init, false);
 
-	secprop->AddInitFunction(&VOODOO_Init, false);
-
-	const char* voodootypes[] = { "12mb", "4mb", "disabled", nullptr };
-	pstring = secprop->Add_string("voodoo", only_at_start, "12mb");
+	const char* voodootypes[] = {"disabled", "4", "12", nullptr};
+	pstring = secprop->Add_string("voodoo_memsize", only_at_start, "12mb");
 	pstring->Set_values(voodootypes);
-	pstring->Set_help("RAM amount of emulated Vodooo 3dfx card.");
+	pstring->Set_help(
+	        "Memory size (in MB) of the 3dfx Vodooo card (12 MB as default).");
 
-	pint = secprop->Add_int("voodoo_perf", only_at_start, 0);
-	pint->SetMinMax(0, 4);
-	pint->Set_help("Toggle performance optimizations for Vodooo 3dfx emulation (0 = none, 1 = use multi-threading, 2 = disable bilinear filter, 3 = both).");
+	pint = secprop->Add_int("voodoo_perf", only_at_start, 1);
+	pint->SetMinMax(0, 3);
+	pint->Set_help("Performance optimisations to use when emulating the 3dfx Voodoo card:\n"
+	               "Note: emulation is software-based and does not use host-level OpenGL calls.\n"
+	               "   0:  No optimizations.\n"
+	               "   1:  Multi-threading (default).\n"
+	               "   2:  Disable bilinear filtering.\n"
+	               "   3:  All optimizations (both 1 and 2).");
 
 	// Configure capture
 	CAPTURE_AddConfigSection(control);
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -6952,13 +6952,18 @@ static double Voodoo_GetVRetracePosition() {
 
 static double Voodoo_GetHRetracePosition() {
 	// TODO proper implementation
-	double time_in_frame = PIC_FullIndex() - v->draw.frame_start;
-	double hfreq = v->draw.vfreq*100.0;
-	if (hfreq <= 0.0) return 0.0;
+	const auto time_in_frame = PIC_FullIndex() - v->draw.frame_start;
+
+	const auto hfreq = static_cast<double>(v->draw.vfreq) * 100;
+
+	if (hfreq <= 0) {
+		return 0.0;
+	}
 	if (v->clock_enabled && v->output_on) {
 		return time_in_frame/hfreq;
-	} else if (v->output_on) {
-		double rtime = time_in_frame/hfreq;
+	}
+	if (v->output_on) {
+		auto rtime = time_in_frame / hfreq;
 		rtime = fmod(rtime, 1.0);
 		return rtime;
 	}
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -6912,7 +6912,8 @@ static void voodoo_shutdown() {
 	}
 }
 
-static void voodoo_vblank_flush(void) {
+static void voodoo_vblank_flush()
+{
 #ifdef C_ENABLE_VOODOO_OPENGL
 	if (v->ogl)
 		voodoo_ogl_vblank_flush();
@@ -6928,7 +6929,8 @@ static void voodoo_set_window(void) {
 }
 #endif
 
-static void voodoo_leave(void) {
+static void voodoo_leave()
+{
 #ifdef C_ENABLE_VOODOO_OPENGL
 	if (v->ogl) {
 		voodoo_ogl_leave(true);
@@ -6937,7 +6939,8 @@ static void voodoo_leave(void) {
 	v->active = false;
 }
 
-static void voodoo_activate(void) {
+static void voodoo_activate()
+{
 	v->active = true;
 
 #ifdef C_ENABLE_VOODOO_OPENGL
@@ -7061,7 +7064,8 @@ static double Voodoo_GetHRetracePosition() {
 	return 0.0;
 }
 
-static void Voodoo_UpdateScreen(void) {
+static void Voodoo_UpdateScreen()
+{
 	// abort drawing
 	RENDER_EndUpdate(true);
 
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -7146,19 +7146,6 @@ static void voodoo_init() {
 	recompute_video_memory(v);
 }
 
-static void voodoo_shutdown() {
-	if (v!=nullptr) {
-#ifdef C_ENABLE_VOODOO_OPENGL
-		if (v->ogl)
-			voodoo_ogl_shutdown(v);
-#endif
-		v->active=false;
-		triangle_worker_shutdown(v->tworker);
-		delete v;
-		v = nullptr;
-	}
-}
-
 static void voodoo_vblank_flush()
 {
 #ifdef C_ENABLE_VOODOO_OPENGL
@@ -7530,7 +7517,7 @@ static uint32_t voodoo_current_lfb;
 static PageHandler* voodoo_pagehandler;
 
 struct PCI_SSTDevice : public PCI_Device {
-	enum {
+	enum : uint16_t {
 		vendor          = 0x121a,
 		device_voodoo_1 = 0x0001,
 		device_voodoo_2 = 0x0002
@@ -7681,6 +7668,28 @@ struct PCI_SSTDevice : public PCI_Device {
 		return true;
 	}
 };
+
+static void voodoo_shutdown()
+{
+	if (!v) {
+		return;
+	}
+	LOG_MSG("VOODOO: Shutting down");
+
+#ifdef C_ENABLE_VOODOO_OPENGL
+	if (v->ogl) {
+		voodoo_ogl_shutdown(v);
+	}
+#endif
+
+	v->active = false;
+	triangle_worker_shutdown(v->tworker);
+
+	delete v;
+	v = nullptr;
+
+	PCI_RemoveDevice(PCI_SSTDevice::vendor, PCI_SSTDevice::device_voodoo_1);
+}
 
 static void Voodoo_Startup() {
 	// This function is called delayed after booting only once a game actually requests Voodoo support
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -6336,19 +6336,19 @@ static int32_t texture_w(uint32_t offset, uint32_t data) {
 
 		[[maybe_unused]] bool changed = false;
 		if (dest[BYTE4_XOR_LE(tbaseaddr + 0)] != ((data >> 0) & 0xff)) {
-			dest[BYTE4_XOR_LE(tbaseaddr + 0)] = (data >> 0) & 0xff;
+			dest[BYTE4_XOR_LE(tbaseaddr + 0)] = static_cast<uint8_t>((data >> 0) & 0xff);
 			changed = true;
 		}
 		if (dest[BYTE4_XOR_LE(tbaseaddr + 1)] != ((data >> 8) & 0xff)) {
-			dest[BYTE4_XOR_LE(tbaseaddr + 1)] = (data >> 8) & 0xff;
+			dest[BYTE4_XOR_LE(tbaseaddr + 1)] = static_cast<uint8_t>((data >> 8) & 0xff);
 			changed = true;
 		}
 		if (dest[BYTE4_XOR_LE(tbaseaddr + 2)] != ((data >> 16) & 0xff)) {
-			dest[BYTE4_XOR_LE(tbaseaddr + 2)] = (data >> 16) & 0xff;
+			dest[BYTE4_XOR_LE(tbaseaddr + 2)] = static_cast<uint8_t>((data >> 16) & 0xff);
 			changed = true;
 		}
 		if (dest[BYTE4_XOR_LE(tbaseaddr + 3)] != ((data >> 24) & 0xff)) {
-			dest[BYTE4_XOR_LE(tbaseaddr + 3)] = (data >> 24) & 0xff;
+			dest[BYTE4_XOR_LE(tbaseaddr + 3)] = static_cast<uint8_t>((data >> 24) & 0xff);
 			changed = true;
 		}
 
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -7390,10 +7390,12 @@ void VOODOO_Destroy(Section* /*sec*/) {
 }
 
 void VOODOO_Init(Section* sec) {
-	// Only active on SVGA machines
-	if (machine != MCH_VGA || svgaCard == SVGA_None) return;
+	Section_prop* section = dynamic_cast<Section_prop*>(sec);
 
-	Section_prop * section = static_cast<Section_prop *>(sec);
+	// Only activate on SVGA machines
+	if (machine != MCH_VGA || svgaCard == SVGA_None || !section) {
+		return;
+	}
 
 	switch (section->Get_string("voodoo")[0])
 	{
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -1000,7 +1000,7 @@ inline uint8_t count_leading_zeros(uint32_t value)
 	DWORD idx = 0;
 	return (_BitScanReverse(&idx, value) ? (uint8_t)(31 - idx) : (uint8_t)32);
 #else
-	return (value ? (uint8_t)__builtin_clz(value) : (uint8_t)32);
+	return (value != 0u ? (uint8_t)__builtin_clz(value) : (uint8_t)32);
 #endif
 	//int32_t result;
 	//
@@ -1062,7 +1062,7 @@ inline int64_t fast_reciplog(int64_t value, int32_t* log_2)
 	}
 
 	/* if we've spilled out of 32 bits, push it down under 32 */
-	if (value & LONGTYPE(0xffff00000000))
+	if ((value & LONGTYPE(0xffff00000000)) != 0)
 	{
 		temp = (uint32_t)(value >> 16);
 		exponent -= 16;
@@ -1142,7 +1142,7 @@ inline int32_t float_to_int32(const uint32_t data, const int fixedbits)
 	} else {
 		result = clamp_to_int32(static_cast<int64_t>(result) << exponent);
 	}
-	if (data & 0x80000000) {
+	if ((data & 0x80000000) != 0u) {
 		result = -result;
 	}
 	return result;
@@ -1166,7 +1166,7 @@ inline int64_t float_to_int64(uint32_t data, int fixedbits)
 		else
 			result = LONGTYPE(0x7fffffffffffffff);
 	}
-	if (data & 0x80000000)
+	if ((data & 0x80000000) != 0u)
 		result = -result;
 	return result;
 }
@@ -3241,7 +3241,7 @@ static inline void raster_generic(const voodoo_state* vs, uint32_t TMUS, uint32_
 		}
 		else
 		{
-			if (!(texel.rgb.a & 0x80))					/* iterated RGB */
+			if ((texel.rgb.a & 0x80) == 0)					/* iterated RGB */
 				c_local.u = iterargb.u;
 			else											/* color0 RGB */
 				c_local.u = regs[color0].u;
@@ -3640,7 +3640,7 @@ static void init_fbi(fbi_state* f, int fbmem)
 	//v->fbi.clut[32] = MAKE_ARGB(32,0xff,0xff,0xff);
 
 	/* allocate a VBLANK timer */
-	f->vblank = false;
+	f->vblank = 0u;
 
 	/* initialize the memory FIFO */
 	f->fifo.size = 0;
@@ -3920,7 +3920,7 @@ static void recompute_video_memory(voodoo_state *vs)
 static void ncc_table_write(ncc_table *n, uint32_t regnum, uint32_t data)
 {
 	/* I/Q entries reference the palette if the high bit is set */
-	if (regnum >= 4 && (data & 0x80000000) && n->palette)
+	if (regnum >= 4 && ((data & 0x80000000) != 0u) && (n->palette != nullptr))
 	{
 		const uint32_t index = ((data >> 23) & 0xfe) | (regnum & 1);
 
@@ -3935,7 +3935,7 @@ static void ncc_table_write(ncc_table *n, uint32_t regnum, uint32_t data)
 		}
 
 		/* if we have an ARGB palette as well, compute its value */
-		if (n->palettea)
+		if (n->palettea != nullptr)
 		{
 			const uint32_t a = ((data >> 16) & 0xfc) |
 			                   ((data >> 22) & 0x03);
@@ -4104,7 +4104,7 @@ static void recompute_texture_params(tmu_state *t)
 	bppscale = TEXMODE_FORMAT(t->reg[textureMode].u) >> 3;
 
 	/* start with the base of LOD 0 */
-	if (t->texaddr_shift == 0 && (t->reg[texBaseAddr].u & 1))
+	if (t->texaddr_shift == 0 && ((t->reg[texBaseAddr].u & 1) != 0u))
 		LOG(LOG_VOODOO,LOG_WARN)("Tiled texture\n");
 	base = (t->reg[texBaseAddr].u & t->texaddr_mask) << t->texaddr_shift;
 	t->lodoffset[0] = base & t->mask;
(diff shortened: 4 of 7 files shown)


kcgen:
Commits are building ...



```text
Compiling 776a8da6d Reduce allowed Clang analysis warnings b.. gcc clang
Compiling 4d23a0d32 Reduce allowed PVS Studio warnings by 66.. gcc clang
Compiling d851e6a6d De-Bitu the BIOS find-device-by-class qu.. gcc clang
Compiling 540c8044a De-Bitu the BIOS find-device-by-tag quer.. gcc clang
Compiling 71146023e Drop unused word and dword PCI read size.. gcc clang
Compiling 9a04e40a4 Drop unused word and dword PCI write siz.. gcc clang
Compiling 1027dfc87 Update voodoo header with changes and to.. gcc clang
Compiling 812aaa45f Apply parentheses to voodoo boolean bloc.. gcc clang
Compiling e413acf3e Clarify use of static members versus ins.. gcc clang
Compiling 25d6d045d Drop redundant declaration in voodoo    .. gcc clang
Compiling 4e4f306e5 Replace implicit boolean statements with.. gcc clang
Compiling a6edfe1c9 Replace typedefs with using in voodoo   .. gcc clang
Compiling e46971d52 Use a literal bool instead of integer in.. gcc clang
Compiling 2e66102d0 Use auto for a couple variables in voodo.. gcc clang
Compiling bab8223e7 Drop redundant void args in voodoo      .. gcc clang
Compiling e752a7242 Bundle voodoo fog-related macro defines .. gcc clang
Compiling 9ef64496a Loop directly over thread variables in v.. gcc clang
Compiling 5d541dd34 Apply const to all applicable voodoo var.. gcc clang
Compiling eadac0b42 Drop else after return statements in voo.. gcc clang
Compiling 856652cb9 Move initialization of members to defaul.. gcc clang
Compiling 16195f1c8 Dynamic cast to derived section class in.. gcc clang
Compiling b176738e0 Apply bugprone macro parentheses cleanup.. gcc clang
Compiling a2bac6aa8 Apply readability isolate declaration cl.. gcc clang
Compiling 4d04184b6 Use double in voodoo PIC_AddEvent calls .. gcc clang
Compiling fa6f76d12 Move voodoo draw_state vfreq to double t.. gcc clang
Compiling 4e5ffa83d Initialize voodoo draw_state members wit.. gcc clang
Compiling d90deda81 Align the voodoo dithermatrix to the lar.. gcc clang
Compiling 248135dd3 Retain the voodoo dithermatrix across ca.. gcc clang
Compiling 7780746fd Allocate and align voodoo FBI memory    .. gcc clang
Compiling 590e06ad3 Allocate and align voodoo TMU memory    .. gcc clang
Compiling 6eceb5869 Clamp the exponent and left-shift value .. gcc clang
Compiling de61fa386 Use safer signed left-shift in voodoo fa.. gcc clang
Compiling 37437198d Limit voodoo override value to destinati.. gcc clang
Compiling eb095512a Replace E_Exit with assertions in voodoo.. gcc clang
Compiling bab06cfe3 Replace E_Exit with assertions in voodoo.. gcc clang
Compiling 478c84c7b Replace E_Exit with assertions in voodoo.. gcc clang
Compiling 8bb661460 Replace E_Exit with assertions in voodoo.. gcc clang
Compiling 61b0dd635 Replace E_Exit with assertion in voodoo .. gcc clang
Compiling 2a3c52746 Replace E_Exit with assertion in voodoo .. gcc clang
Compiling b3530c912 Replace E_Exit with assertions in voodoo.. gcc clang
Compiling eb8929a65 Limit voodoo word values to match destin.. gcc clang
Compiling e7849a243 Limit voodoo byte values to match destin.. gcc clang
Compiling 72a8a404e Use const int in alpha blend macro      .. gcc clang
Compiling 4067d5192 Reformat voodoo alpha blend macro       .. gcc clang
Compiling e106708da Disambiguate local s-values from previou.. gcc clang
Compiling d37012cd6 Use const auto and simplify voodoo retra.. gcc clang
Compiling c2b8459c1 Limit voodoo rgb values to destination t.. gcc clang
Compiling 25692091a Limit voodoo fog values to destination t.. gcc clang
Compiling 40c1b332f Move unused voodoo vtotal into debug blo.. gcc clang
Compiling 6c598f4da Use float literal in voodoo FBI calculat.. gcc clang
Compiling 31ac4a63f Explicitly cast to float in voodoo round.. gcc clang
Compiling 45e8d74b5 Limit voodoo FBI frontbuf value to desti.. gcc clang
Compiling 04d98ec27 Drop unused voodoo state in FBI initiali.. gcc clang
Compiling a72ef08fc Expand voodoo clamp local variables to a.. gcc clang
Compiling bab237484 Drop useless casts in voodoo memory base.. gcc clang
Compiling 5221b6b37 Drop useless casts in voodoo memory writ.. gcc clang
Compiling 41e2bccc9 Drop useless casts in voodoo port read f.. gcc clang
Compiling b95885e26 Drop useless casts in voodoo port write .. gcc clang
Compiling d736ef96a Drop useless casts in voodoo buffer offs.. gcc clang
Compiling 4e19eca12 Drop useless cast in voodoo log2 macro  .. gcc clang
Compiling 7d583ce29 Replace voodoo 32-bit byte swap macro wi.. gcc clang
Compiling 40f99f79b Drop useless casts in voodoo prepare_tmu.. gcc clang
Compiling 35ae69065 Drop useless casts in voodoo color handl.. gcc clang
Compiling 3be94ccef Drop useless casts in voodoo texture pip.. gcc clang
Compiling 2a062e0c1 Drop useless casts in voodoo clamp funct.. gcc clang
Compiling b865cb4e1 Disambiguate the semaphore's initial cou.. gcc clang
Compiling f2c36d0e3 Instantiate voodoo setup_and_draw_triang.. gcc clang
Compiling 2e74a491c Reference the voodoo setup vertex struct.. gcc clang
Compiling c4cfe80a6 Reference the voodoo texture mapping uni.. gcc clang
Compiling 8ba82648e Expand voodoo FBI and TMU acronyms      .. gcc clang
Compiling 4be5eb201 Reference the voodoo frame buffer interf.. gcc clang
Compiling a94a0f9e3 Reference the voodoo state registers loc.. gcc clang
Compiling 3db190ccb Don't shadow the global voodoo state in .. gcc clang
Compiling 8edc98f70 Don't shadow the global voodoo state in .. gcc clang
Compiling 1f146bcce Don't shadow the global voodoo state in .. gcc clang
Compiling 80a61bc23 Don't shadow the global voodoo state in .. gcc clang
Compiling e3e4874be Don't shadow the global voodoo state in .. gcc clang
Compiling d5d2b33ff Don't shadow the global voodoo state in .. gcc clang
Compiling f5022239d Don't shadow the global voodoo state in .. gcc clang
Compiling 1fba29e0a Don't shadow the global voodoo state in .. gcc clang
Compiling 19da1e292 Don't shadow the global voodoo state in .. gcc clang
Compiling d40f79c36 Don't shadow the global voodoo state in .. gcc clang
Compiling 5a671e468 Don't shadow the global voodoo state in .. gcc clang
Compiling 844e2d8c3 Don't shadow the global voodoo state in .. gcc clang
Compiling 39aeb3322 Don't shadow the global voodoo state in .. gcc clang
Compiling 245d3220f Don't shadow the global voodoo state in .. gcc clang
Compiling 3d5783dea Don't shadow the global voodoo state in .. gcc clang
Compiling a2f213250 Use const auto and bit-literal determini.. gcc clang
Compiling f20eb4ece Drop unused voodoo fifo-space management.. gcc clang
Compiling 3985a7155 Collapse duplicate voodoo alpha-blend ca.. gcc clang
Compiling 3551e1d3f Use unaligned 64-bit IO in voodoo fastfi.. gcc clang
Compiling 2f0731686 Reduce size of voodoo chips type to actu.. gcc clang
Compiling ac2d789cb Expand voodoo debug block to log-only va.. gcc clang
Compiling 420c134ff Collapse duplicate voodoo triangle cases.. gcc clang
Compiling 1a4624427 Expand voodoo variable prior to shifting.. gcc clang
Compiling 5393e901d Comment out voodoo pending swaps as its .. gcc clang
Compiling c67b90c9a Collapse duplicate voodoo criteria into .. gcc clang
Compiling 9c5f4c24d Refactor voodoo reciprocal log LUT gener.. gcc clang
Compiling 6b0818fb5 Refactor voodoo dither LUT generator mac.. gcc clang
Compiling 63340ae1c Make voodoo dither matrices constexpr   .. gcc clang
Compiling a00d34af6 Avoid re-assigning the same vtype in voo.. gcc clang
Compiling c25601bdd Avoid right-shifting by a negative amoun.. gcc clang
Compiling c3cfbc2f8 Localize assignment of regnum in voodoo .. gcc clang
Compiling 6b1fbcb45 Swap the voodoo extent start and end as .. gcc clang
Compiling 78be5645e Exclude OpenGL-specific variable assignm.. gcc clang
Compiling 8a74bf316 Initialize voodoo lfb_w locals with defa.. gcc clang
Compiling 4d53bf361 Eliminate redundant type-assigning in vo.. gcc clang
Compiling 8d55d46b0 Localize conversion of addr to offset in.. gcc clang
Compiling c3a7874ce Notify thread semaphores directly in voo.. gcc clang
Compiling 4bbbef914 Initialize and iterate threads directly .. gcc clang
Compiling 78db0d46b Move array length to constexpr in voodoo.. gcc clang
Compiling 837e2d099 Check voodoo's pagehandler readd return .. gcc clang
Compiling bdf6da8c5 Use const in voodoo register_r function .. gcc clang
Compiling 045337341 Localize variable init in voodoo registe.. gcc clang
Compiling 4924abdda Use const auto in voodoo register_r     .. gcc clang
Compiling a4bf21e71 Use const auto and defaults in voodoo lf.. gcc clang
Compiling 22f6eab2e Remove optional voodoo support          .. gcc clang
```

src/hardware/voodoo.cpp
```diff
@@ -383,7 +383,7 @@ static constexpr uint8_t dither_matrix_2x2[16] =
 #define CLAMP(val,min,max)		do { if ((val) < (min)) { (val) = (min); } else if ((val) > (max)) { (val) = (max); } } while (0)
 
 /* macro to compute the base 2 log for LOD calculations */
-#define LOGB2(x)				(log((double)(x)) / log(2.0))
+#define LOGB2(x)				(log((x)) / log(2.0))
```
kklobe, line 397: This macro isn't even necessary, as `math.h` contains `log2()`

src/hardware/voodoo.cpp
```diff
@@ -383,7 +383,7 @@ static constexpr uint8_t dither_matrix_2x2[16] =
 #define CLAMP(val,min,max)		do { if ((val) < (min)) { (val) = (min); } else if ((val) > (max)) { (val) = (max); } } while (0)
```
kklobe, line 394: could we clean this up too, e.g.

```cpp
#define CLAMP(val, min, max)		(val = std::clamp(val, min, max))
```

src/hardware/voodoo.cpp
```diff
@@ -4824,15 +4824,11 @@ static void fastfill(voodoo_state *v)
...
+			auto extent = &extents[curscan - startscanline];
 
 			/* force start < stop */
-			if (istartx > istopx)
-			{
-				int32_t temp = istartx;
-				istartx = istopx;
-				istopx = temp;
+			if (extent->startx > extent->stopx) {
+				std::swap(extent->startx, extent->stopx);
```
johnnovak, line 5097: I'm a bit worried about fixing i ntended stuff like this that effectively did nothing. Maybe doing nothing is the correct behaviour here and the code is misleading (the person just forgot to remove the effectively dead code that's no longer needed).

Have ypu compared the output after "fixing" this? Is there any difference?

src/hardware/voodoo.cpp
```diff
@@ -2316,8 +2317,9 @@ do																				\
 				depthval = 0xffff;												\
 			else																\
 			{																	\
-				int exp = count_leading_zeros(temp);							\
-				depthval = ((exp << 12) | ((~temp >> (19 - exp)) & 0xfff));		\
+				const auto exp = count_leading_zeros(temp);                     \
+				const auto right_shift = std::max(0, 19 - exp);	                \
+				depthval = ((exp << 12) | ((~temp >> right_shift) & 0xfff));    \
```
johnnovak, line 2334: Does this fix an observable, obvious rendering bug? We can't make any assumptions on how the Voodoo hardware works, this might be intentional and exact recreation of the behaviour of the real thing.

src/hardware/voodoo.cpp
```diff
@@ -1286,8 +1286,39 @@ inline uint32_t compute_raster_hash(const raster_info* info)
...
+
+		const auto color = static_cast<uint8_t>((i >> 3) & 0xff);
+		const auto amount = dither_amounts[y * 4 + x];
+		const auto use_rb = (i & 1) == 0;
+
+		dither_lut[i] = use_rb ? dither_rb(color, amount)
+		                       : dither_g(color, amount);
+	}
+	return dither_lut;
+}
```
johnnovak, line 1315: Very nice 🎉

src/hardware/voodoo.cpp
```diff
@@ -5920,7 +5920,7 @@ static void lfb_w(uint32_t offset, uint32_t data, uint32_t mem_mask) {
 			/* make sure we care about this pixel */
 			if (mask & 0x0f)
 			{
-				int64_t iterw = sw[pix] << (30-16);
+				int64_t iterw = static_cast<int64_t>(sw[pix]) << (30-16);
```
johnnovak: Another risky change; are we 100% the "fix" doesn't break anything?

src/hardware/voodoo.cpp
kcgen: got it :+1:

src/hardware/voodoo.cpp
kcgen: There are a couple areas that process face-forward triangles - a slower path and a fastfill path. 

Both have the same logic, and they also have the same comments. Here's the other area (existing code, not from this PR, see line 4234):

```
/* force start < stop */
if (extent.startx >= extent.stopx) {
	if (extent.startx == extent.stopx) continue;
	std::swap(extent.startx, extent.stopx);
}
```

If the triangles are flipped, then the face is transparent, which is the same is not even processing the triangle:



Unfortunately during testing I wasn't able to find anything that uses the fastfill code path.

But there is evidence in `raster_fastfill` that it expects face-forward triangles:

1. There is no initial sanity check or bail-out "if start > stop"
2. Loops go from `start` to `stop`
3. a pixel-counter would run in reverse if start > stop: `stats.pixels_out += stopx - startx`
4. the most telling: the function compute a new stop value (line 3633) and at that point, if `stop < start` then it immediately bails out.

So in absense of being able to test it, there are three points of evidence that point to the code actually expecting the triangle extent to be "start <= stop".
 - the comment
 - the code itself (try) to do the swap
 - the raster_fastfill function expecting start <= stop plus actual bail out when stop is recomputed.

src/hardware/voodoo.cpp
kcgen: It expects to shift the `~temp` value to the right by up to 19 bits (at the most).

At the least, we can right-shift by zero places (and not change anything).

However if the `temp` value has more than 19 leading zero, then the (original) code actually tries to right-shift `~temp` by /negative/ bits.  For example: `~temp >> -3` : that's garbage land and triggers a UBSAN crash.

So I'm just stopping it from driving into negative territory and triggering undefined behavior.

But in terms of valid-shift ranges, that's left intact. The shift range from 19 at the most all the way down to 0. 

(If the code needs to trigger UBSAN errors to work correctly, then it's not passing CI and it goes back into the expeterminal-flag jail :-)!)

src/hardware/voodoo.cpp
kcgen: The original author used a 64-bit LHS, expecting the RHS to step up to full 64-bit range due to 14-bit left-shift.

So the code just explicitly lets that happen (and fixes a warning about it).

I'm confident that if the author really wanted to cut-off at 32-bit mark, they would have `& 0xffffffff` the result and stored it in an `int32` (because this &-mask at type-sizes is prevalent throughout the voodoo code).

src/hardware/voodoo.cpp
johnnovak: Thanks, given that you've extensively researched the situation, that make this definitely a safe change. I was just lacking all that context only by looking at this commit 😄

src/hardware/voodoo.cpp
johnnovak: Thanks for the explanation. I'm actually not going into great detail with the review, just flagging things that go beyond simple refactoring and look suspcious 😄 I can stop that if it's bothersome and you've done extensive testing.

kcgen: Please keep flagging!
I did test each commit as I went - but there's always a chance of typos and my own mistakes! Need all the help I can get ❤️👍

src/hardware/voodoo.cpp
```diff
@@ -4644,79 +4663,139 @@ static void setup_and_draw_triangle(voodoo_state *vs)
...
+		vs->tmu[1].dsdy = (int64_t)(((fbi.svert[0].s1 - fbi.svert[2].s1) * dy1 -
+		                             (fbi.svert[0].s1 - fbi.svert[1].s1) * dy2) *
+		                            tdiv);
+		vs->tmu[1].startt = (int64_t)(fbi.svert[0].t1 * 65536.0f * 65536.0f);
+		vs->tmu[1].dtdx = (int64_t)(((fbi.svert[0].t1 - fbi.svert[1].t1) * dx1 -
+		                             (fbi.svert[0].t1 - fbi.svert[2].t1) * dx2) *
+		                            tdiv);
+		vs->tmu[1].dtdy = (int64_t)(((fbi.svert[0].t1 - fbi.svert[2].t1) * dy1 -
+		                             (fbi.svert[0].t1 - fbi.svert[1].t1) * dy2) *
+		                            tdiv);
```
johnnovak: Nice readability improvements 👍🏻

src/hardware/voodoo.cpp
```diff
@@ -4515,8 +4518,10 @@ static void triangle(voodoo_state *vs)
 	extra.r_alphaMode    = regs[alphaMode].u;
 	extra.r_fogMode      = regs[fogMode].u;
 
-	extra.r_textureMode0 = vs->tmu[0].reg[textureMode].u;
-	if (vs->tmu[1].ram != NULL) extra.r_textureMode1 = vs->tmu[1].reg[textureMode].u;
+	extra.r_textureMode0 = tmu0.reg[textureMode].u;
+	if (tmu1.ram != NULL) {
```
johnnovak, line 4650: `nullptr` while you're at it?

src/hardware/voodoo.cpp
```diff
@@ -4640,19 +4640,22 @@ static void setup_and_draw_triangle(voodoo_state *vs)
...
+	const auto& vertex1 = fbi.svert[1];
+	const auto& vertex2 = fbi.svert[2];
+
+	/* grab the X/Ys at least */
+	fbi.ax = (int16_t)(vertex0.x * 16.0);
+	fbi.ay = (int16_t)(vertex0.y * 16.0);
+	fbi.bx = (int16_t)(vertex1.x * 16.0);
+	fbi.by = (int16_t)(vertex1.y * 16.0);
+	fbi.cx = (int16_t)(vertex2.x * 16.0);
+	fbi.cy = (int16_t)(vertex2.y * 16.0);
```
johnnovak: I like how you're "rolling" these nested references up iteratively, commit by commit.

src/hardware/voodoo.cpp
```diff
@@ -166,22 +167,6 @@ struct rectangle
...
-		(((((uint64_t) (x)) >> 48) & ((uint64_t) 0xFF)) <<  8)	|	\
-		(((((uint64_t) (x)) >> 40) & ((uint64_t) 0xFF)) << 16)	|	\
-		(((((uint64_t) (x)) >> 32) & ((uint64_t) 0xFF)) << 24)	|	\
-		(((((uint64_t) (x)) >> 24) & ((uint64_t) 0xFF)) << 32)	|	\
-		(((((uint64_t) (x)) >> 16) & ((uint64_t) 0xFF)) << 40)	|	\
-		(((((uint64_t) (x)) >>  8) & ((uint64_t) 0xFF)) << 48)	|	\
-		(((((uint64_t) (x)) >>  0) & ((uint64_t) 0xFF)) << 56)		\
-	)
-
 #define ACCESSING_BITS_0_15				((mem_mask & 0x0000ffff) != 0)
```
johnnovak, line 180: Good riddance! 🥇

src/hardware/voodoo.cpp
```diff
@@ -2430,16 +2430,16 @@ do																				\
...
 	/* write to aux buffer */													\
-	if (depth && FBZMODE_AUX_BUFFER_MASK(FBZMODE))								\
+	if ((depth) && FBZMODE_AUX_BUFFER_MASK(FBZMODE))								\
 	{																			\
 		if (FBZMODE_ENABLE_ALPHA_PLANES(FBZMODE) == 0)							\
-			depth[XX] = (uint16_t)depthval;										\
+			(depth)[XX] = (uint16_t)depthval;										\
 		else																	\
-			depth[XX] = (uint16_t)a;												\
+			(depth)[XX] = (uint16_t)a;												\
```
johnnovak, line 2441: These macros 🥶 Oh man...

src/hardware/voodoo.cpp
```diff
@@ -7351,10 +7395,10 @@ struct PCI_SSTDevice : public PCI_Device {
 
 		registers[0x3c] = 0xff;	// no irq
 
-		// memBaseAddr: size is 16MB
-		uint32_t address_space = (VOODOO_INITIAL_LFB & 0xfffffff0) |
-		                         0x08; // memory space, within first
-		                               // 4GB, prefetchable
+		// 16MB within within the first 4GB is prefetchable
+		const uint32_t address_space = (VOODOO_INITIAL_LFB & 0xfffffff0) |
```
johnnovak, line 7673: Const 'em all! 😎 🤘🏻

src/hardware/voodoo.cpp
```diff
@@ -3007,11 +3007,13 @@ static voodoo_state* v = nullptr;
...
-#define LOG_REGISTERS		(0)
-#define LOG_LFB				(0)
-#define LOG_TEXTURE_RAM		(0)
-#define LOG_RASTERIZERS		(0)
+enum {
+	LOG_VBLANK_SWAP = (0),
+	LOG_REGISTERS   = (0),
+	LOG_LFB         = (0),
+	LOG_TEXTURE_RAM = (0),
+	LOG_RASTERIZERS = (0)
```
johnnovak, line 3040: Minor: drop the parens?

src/hardware/voodoo.cpp
```diff
@@ -5672,12 +5672,12 @@ static void register_w(uint32_t offset, uint32_t data)
 		case texBaseAddr_1:
 		case texBaseAddr_2:
 		case texBaseAddr_3_8:
-			if (chips & 2)
+			if ((chips & 2) != 0)
```
johnnovak, line 5895: I'm not against these changes, I just had the impression we were okay with implicit true/false checks when using integer expressions in `if` statements. But happy to do these checks explicitly in the future.

src/dosbox.cpp
```diff
@@ -781,19 +781,22 @@ void DOSBOX_Init()
...
-
-	secprop->AddInitFunction(&VOODOO_Init, false);
+	secprop = control->AddSection_prop("voodoo", &VOODOO_Init, false);
 
 	const char* voodootypes[] = { "12mb", "4mb", "disabled", nullptr };
-	pstring = secprop->Add_string("voodoo", only_at_start, "12mb");
+	pstring = secprop->Add_string("voodoo_memsize", only_at_start, "12mb");
 	pstring->Set_values(voodootypes);
-	pstring->Set_help("RAM amount of emulated Vodooo 3dfx card.");
+	pstring->Set_help("Memory size of the emulated Vodooo 3dfx card ('12mb' by default).");
```
johnnovak, line 791: I think we should refrain from the word "emulated" in the config setting descriptions and in the documentation in general as it's very redundant (this an emulator... 😅). I think we should only use "emulated" when we must disambiguate a concept that can refer to both something in the emulated and the host machine (e.g. emulated CPU vs host CPU), but that's not the case here.

src/dosbox.cpp
```diff
@@ -781,19 +781,22 @@ void DOSBOX_Init()
...
+	secprop->AddInitFunction(&PCI_Init); // PCI bus
 
-	secprop=control->AddSection_prop("pci", &PCI_Init, false); // PCI bus
-
-	secprop->AddInitFunction(&VOODOO_Init, false);
+	secprop = control->AddSection_prop("voodoo", &VOODOO_Init, false);
 
 	const char* voodootypes[] = { "12mb", "4mb", "disabled", nullptr };
-	pstring = secprop->Add_string("voodoo", only_at_start, "12mb");
+	pstring = secprop->Add_string("voodoo_memsize", only_at_start, "12mb");
```
johnnovak, line 789: I'd uppercase `MB` (megabyte is not `mb`) in the config description, but parse it in a case-insensitive manner.

src/dosbox.cpp
```diff
@@ -781,19 +781,22 @@ void DOSBOX_Init()
...
 	pstring->Set_values(voodootypes);
-	pstring->Set_help("RAM amount of emulated Vodooo 3dfx card.");
+	pstring->Set_help("Memory size of the emulated Vodooo 3dfx card ('12mb' by default).");
 
 	pint = secprop->Add_int("voodoo_perf", only_at_start, 0);
-	pint->SetMinMax(0, 4);
-	pint->Set_help("Toggle performance optimizations for Vodooo 3dfx emulation (0 = none, 1 = use multi-threading, 2 = disable bilinear filter, 3 = both).");
+	pint->SetMinMax(0, 3);
+	pint->Set_help("Performance optimizations avaiable when emulating the Vodooo 3dfx card:\n"
```
johnnovak, line 795: typo: available

I think the word "emulate" is warranted here 😄

johnnovak: But this should be reworded: "Performance optimisations to use when emulating the 3dfx Voodoo card" (not Voodoo 3dfx; 3dfx is the name of the company).

src/dosbox.cpp
johnnovak: 3dfx Voodoo

src/dosbox.cpp
```diff
@@ -781,19 +781,22 @@ void DOSBOX_Init()
...
+	               "   0:  No optimizations (default).\n"
+	               "   1:  Multi-threading.\n"
+	               "   2:  Disable bilinear filtering.\n"
+	               "   3:  All optimizations (both 1 and 2).");
```
johnnovak, line 799: So this is a software-only renderer, right? Absolutely zero OpenGL is involved? Should we make that clear somewhere in the descriptions that is is a pure software-based emulation of the Voodoo card that requires no OpenGL?

src/hardware/voodoo.cpp
```diff
@@ -3008,7 +3008,33 @@ iterated W    = 18.32 [48 bits]
...
+		return " and no optimizations";
+	case PerformanceFlags::MultiThreading:
+		return " and multi-threading";
+	case PerformanceFlags::NoBilinearFiltering:
+		return " and no bilinear filtering";
+	case PerformanceFlags::All:
+		return ", multi-threading, and no biliear filtering";
+	}
+	assert(false);
+	return "unknown performance flags (bug)";
```
johnnovak, line 3029: Maybe drop "(bug)" 😄 (we don't do that anywhere else in this "safety" error messages)

.github/workflows/pvs-studio.yml
```diff
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 333
+          MAX_BUGS: 270
```
johnnovak, line 114: Rejoice! 🎉

.github/workflows/analysis.yml
```diff
@@ -122,7 +122,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 12
+          MAX_BUGS: 3
```
johnnovak, line 125: Oh yeah!

johnnovak:
What a comeback @kcgen! You were being coy about this, not telling anyone, then when we had our guards down, you drop _THIS_ for **_M A X I M U M - D R A M A T I C - E F F E C T_**! 😆 Something that I would've done! 😎🤘🏻Congrats, well done! 🎉 

Looks very good, just some minor comments!

src/hardware/voodoo.cpp
kcgen: yup.. the preprocessor expands the macros as one super long line at the point where the macro is called (combined with macros-in-macros further expanding the line). Made it really tricky to understand where static analysis issues were in the code!

src/hardware/voodoo.cpp
kcgen: The voodoo code mixes bools and ints both ways: like storing an int as a bool and vice-versa.  So there were probably some PVS or MSVC warnings about it and I just mass-fixed it all to make it all explicit.

I original started to switch based on one of your original code review comments (where the code was assessing integers in an if/else statement) :-)

If the code is obvious or self-documenting, then I don't really mind either way (and it's really up to the reviewer what they want to allow or not allow).  So.. your call!

Perhaps if this were new code (looking at the code snippet above), we'd be using a bitview.. and have something like:

```c++
   if (chips_register.has_texture) {
       ...
   }
```

src/dosbox.cpp
```diff
@@ -781,19 +781,22 @@ void DOSBOX_Init()
...
 	secprop->AddInitFunction(&VGA_Init);
 	secprop->AddInitFunction(&KEYBOARD_Init);
+	secprop->AddInitFunction(&PCI_Init); // PCI bus
 
-	secprop=control->AddSection_prop("pci", &PCI_Init, false); // PCI bus
-
-	secprop->AddInitFunction(&VOODOO_Init, false);
+	secprop = control->AddSection_prop("voodoo", &VOODOO_Init, false);
 
 	const char* voodootypes[] = { "12mb", "4mb", "disabled", nullptr };
```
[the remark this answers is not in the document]
kcgen: I'm going to switch them to integers too (simply 4 and 12), to match the other `memsize` conf settings.

src/dosbox.cpp
kcgen: fully agree! (and dropped)

src/dosbox.cpp
kcgen: Dropping the 'mb' (or 'MB') postfix from these values, and going with integers to match the other `memsize` conf values (which take in integers).

src/dosbox.cpp
kcgen: > I think the word "emulate" is warranted here smile

yes :-) and replaced w/ your improved wording. Much better!

src/dosbox.cpp
kcgen: Added:
```c++
pint->Set_help("Performance optimisations to use when emulating the 3dfx Voodoo card:\n"
               "Note: emulation is software-based and does not use host-level OpenGL calls.\n"
               "   0:  No optimizations.\n"
               "   1:  Multi-threading (default).\n"
               "   2:  Disable bilinear filtering.\n"
               "   3:  All optimizations (both 1 and 2).");
```

@kklobe : I've tentatively enabled threading as the default optimization; can't remember which we decided was best overall.

src/hardware/voodoo.cpp
kcgen: yes :-)
Dropped!

kcgen:
Thanks for the review, @johnnovak!

PR #2712, opened 2023-08-01
Change VESA BIOS branding to 'DOSBox Staging'

FeralChild64:
I have noticed our VESA BIOS branding is still a vanilla-DOSBox one, time to change it!

Old branding:

old-branding

New branding:

new-branding

Example original BIOS branding (screenshot from 86Box) - for reference:

example-original


johnnovak (approved):
Nice one @FeralChild64 🎉

PR #2713, opened 2023-08-02
Fix deactivation of queued joystick axis movement

kcgen:
When mapping a host keyboard button to a virtual joystick axis, the prior logic would allow a queue of the same event (ie: holding a given direction) to ignore deactivation of the event (ie: release of the direction).

This change now zeroes the queued action and deactivates when requested.

[joytest.zip](https://github.com/dosbox-staging/dosbox-staging/files/12235225/joytest.zip) was used with:
 - Host analog stick driving the axes (native control)
 - Host D-pad mapped to the analog axes
 - Keypad up/down/left/right mapped to the analog axes

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -197,19 +197,16 @@ public:
 			if (!GetActivityCount()) Active(true);
 		}
 	}
-	void DeActivateEvent(bool ev_trigger) override {
-		if (ev_trigger) {
-			if (activity>0) activity--;
-			if (activity==0) {
-				/* test if still some trigger-activity is present,
-				   adjust the state in this case accordingly */
-				if (GetActivityCount()) RepostActivity();
-				else Active(false);
-			}
-		} else {
-			if (!GetActivityCount()) Active(false);
+	void DeActivateEvent(const bool ev_trigger) override
+	{
+		if (ev_trigger || GetActivityCount() == 0) {
+			// Zero-out this event's pending activity if triggered
+			// or we have no opposite-direction events
+			activity = 0;
+			Active(false);
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


src/gui/sdl_mapper.cpp
```diff
@@ -197,19 +197,16 @@ class CContinuousEvent : public CEvent {
...
-				/* test if still some trigger-activity is present,
-				   adjust the state in this case accordingly */
-				if (GetActivityCount()) RepostActivity();
-				else Active(false);
-			}
-		} else {
-			if (!GetActivityCount()) Active(false);
+	void DeActivateEvent(const bool ev_trigger) override
+	{
+		if (ev_trigger || GetActivityCount() == 0) {
```
johnnovak, line 202: Not familiar with this whole code, but I'm wondering why just `if (ev_trigger)` is not enough?

kcgen: Yup; that second path exists in the original code, so I've retained it.

If the event is deactived indirectly (when ev_trigger is false), then this activity-count-zero code path will still deactivate the 'held' event if the opposite axis also has no events remaining (the get activity count function includes the opposite direction too). 

In this second code path, if the opposite axis has events remaining then what happens is the joystick can move to the opposite side in the next step (imagine pulling right to left in one go). 

Where as if there are no activities left, then we reactive back to zero (or center). 

This appears to work that way too: I can snap right and left quickly, but also settle it back to center

johnnovak: Ok, that's some weird ass logic 😅 I guess this will do until we rebuild the mapper from scratch if it works well enough in the interim 🤷🏻

johnnovak (approved):
Nice quick fix @kcgen 🎉 

image

PR #2716, opened 2023-08-02
Always let MacOS homebrew overwrite conflicting junk files

kcgen:
Thanks @weirddan455:

> Looks like the Mac builds are having a fit now (unrelated to this PR):

```text
 ==> Pouring openssl@3--3.1.2.ventura.bottle.tar.gz
Error: Could not symlink bin/openssl
Target /usr/local/bin/openssl
is a symlink belonging to openssl@1.1. You can unlink it:
  brew unlink openssl@1.1

  brew link --overwrite openssl@3

  brew link --overwrite --dry-run openssl@3
Error: Process completed with exit code 1.
```

> OpenSSL 3.1.2 was just released yesterday. These CI machines constantly updating packages and breaking is getting annoying...

This should fix it once and for all.  Will merge when passing.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Install C++ compiler and libraries
         if:   matrix.conf.needs_deps
         run: |
-          arch -arch=${{ matrix.conf.arch }} brew install \
+          arch -arch=${{ matrix.conf.arch }} brew install --overwrite \
             ${{ matrix.conf.packages }} \
             $(cat packages/macos-12-brew.txt)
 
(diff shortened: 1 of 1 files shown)


weirddan455:
Wow, that was fast.  Nice job!  I don't have a Mac and I've never used Brew so I don't know what `--overwrite` does exactly.  I see you updated documentation.  It seems not ideal to force users to add an extra flag but I don't know enough about it really.

weirddan455:
Well, I didn't mean to say drop it from the documentation.  If that flag is required for the install to work, maybe it should be documented.

This just seems like a workaround, which is maybe fine for CI but if it's affecting users trying to install DOSBox Staging from Brew, maybe we should find the root cause?

johnnovak:
So glad I didn't have to deal with this crap. Thanks @kcgen 😅

kcgen:
> **`Pouring openssl@3--3.1.2.ventura.bottle.tar.gz`**
> **`**Error**`**

Beer bros tried to pour us an openssl brew from the bottle, but it overflowed after splashing on a prior brew.  No worries bros `--over-pour` is permitted now :beers:!

PR #2722, opened 2023-08-03
Turn down loguru verbosity when started with following arugments

### LINKED ISSUE #2714: Remove extra noise from the console output when DOSBox is invoked with certain options
# Problem

When the user executes `dosbox --printconf` to get the path of the default config, there's a lot of noise around the actually useful output, making it hard to find what he's really looking for.

In this case, there are 9 extra lines of stuff that can be considered just noise vs the _one single line_ of info the user asked for:

image

The following command line options have the same issue:

- `--editconf`
- `--eraseconf`
- `--list-glshaders`
- `--opencaptures`

# Solution

Logging should be turned off for these command line options. We should only print the useful info to the console without all extra the noise.

weirddan455:
Fixes #2714

These arguments do fairly trival tasks and then end the program. Many also print information to the console.
Turn down log verbosity so the user can more easily see the output.

--printconf --editconf --eraseconf --list-glshaders --erasemapper --opencaptures

It turns out these arguments require some of the initialization functions to be run first.  They also use Loguru macros to log warnings and errors if something goes wrong.  So I couldn't just move them to the top of `sdlmain` like I was originally thinking.

However, it's pretty easy to turn down the verbosity of Loguru and silence the noise.

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4808,6 +4808,12 @@ int sdl_main(int argc, char *argv[])
 	loguru::g_preamble_verbose = false; // The verbosity field
 	loguru::g_preamble_pipe    = true; // The pipe symbol right before the message
 
+	if (arguments->printconf || arguments->editconf ||
+	    arguments->eraseconf || arguments->list_glshaders ||
+	    arguments->erasemapper || !arguments->opencaptures.empty()) {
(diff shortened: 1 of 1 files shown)


johnnovak (approved):
Nice usability improvement @weirddan455, just tested it and works great! 🎉

PR #2723, opened 2023-08-03
String fixes in CONFIG command

FeralChild64:
Colors of several colons in CONFIG help output was inconsistent:

strings1


kcgen (approved):
Good eye catching these inconsistencies, @FeralChild64.

PR #2728, opened 2023-08-05
Remove scan and pixel doubling config settings

johnnovak:
...with FIRE! 🔥🔥🔥🔥 😎 (unfortunately, there is no welding safety glasses emoji available 😆)

Before anyone would start accusing me of indiscriminately removing features added by other people, now one of my own creations meets the chopping board! 🪓 🔪  I've spent quite some time making sure these config settings work correctly at runtime, but given they will be unnecessary in light of what's coming (to be announced later), these features need to go to simplify and streamline the user experience!

---

Okay, so this PR removes the `force_vga_single_scan` and `force_no_pixel_doubling` settings. _Only_ the config settings; the shaders can still set these flags via the  `force_single_scan` and `force_no_pixel_doubling` pragmas.

These flags are set automatically when no OpenGL shader is in use (they're on for nearest-neighbour interpolation for optimisation purposes, and off for bilinear to achieve better image quality), and shaders have full control over these flags via the two similarly named pragmas. Conceptually, these are rendering concerns and belong to the shaders, so having yet another mechanism to override the behaviour via config settings is an unnecessary complication.

If someone wishes to force single-scanning in VGA games, that can be accomplished by creating a custom "15kHz monitor" or "arcade monitor" shader that forces single-scanning via the pragma (and possibly disable pixel doubling too to emulate a low-TVL-count display).

We are actually going to bundle such a special shader with the next release.

### DIFF
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -786,7 +786,7 @@ static void ReloadShader(const bool pressed)
 	assert(glshader_prop);
 	const auto shader_path = std::string(glshader_prop->GetValue());
 	if (shader_path.empty()) {
-		LOG_WARNING("RENDER: Failed gettina path for 'glshader' setting; not reloading");
+		LOG_WARNING("RENDER: Failed getting path for 'glshader' setting; not reloading");
 		return;
 	}
 
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -803,28 +803,23 @@ static bool force_no_pixel_doubling = false;
 
 // We double-scan VGA modes and pixel-double all video modes by default unless:
 //
-//  1) Single-scanning or no pixel-doubling is forced in the
-//     configuration
-//  2) Single-scanning or no pixel-doubling is forced by the OpenGL
-//     shader
-//  3) If the interpolation mode is nearest-neighbour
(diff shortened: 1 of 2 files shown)


kcgen (approved):
Nice improvement to lighten the conf load :+1: 

(skimmed the changes on my phone; all looks good!)

src/gui/render.cpp
```diff
@@ -803,28 +803,23 @@ static bool force_no_pixel_doubling = false;
 
 // We double-scan VGA modes and pixel-double all video modes by default unless:
 //
-//  1) Single-scanning or no pixel-doubling is forced in the
-//     configuration
-//  2) Single-scanning or no pixel-doubling is forced by the OpenGL
-//     shader
-//  3) If the interpolation mode is nearest-neighbour
+//  1) Single-scanning or no pixel-doubling is forced by the OpenGL shader
+//  2) If the interpolation mode is nearest-neighbour
```
shermp, line 807: While you're at it, maybe drop the `If`? Otherwise it doesn't quite read right.

shermp (approved):
A _very_ minor comment. Otherwise looks fine.

src/gui/render.cpp
johnnovak: Good point, fixed.

PR #2730, opened 2023-08-05
Reload shader fixes

johnnovak:
In current main, reloading/changing shaders at runtime is plagued with several serious problems. Sometimes you get a black screen, the sRGB framebuffer cannot be disabled via pragmas once enabled, and there are memory leaks that will invariably result in a hard crash after a certain number of shader changes or reloads.

This PR fixes all that 😎 You can switch shaders and/or reload them as many times as you want—it just works 🤘🏻 Passed the UBSAN tests too.

It should be easy to follow what I've done commit by commit; otherwise not so much 😅

### DIFF
--- a/include/reelmagic.h
+++ b/include/reelmagic.h
@@ -37,7 +37,8 @@ struct ReelMagic_VideoMixerMPEGProvider {
 	virtual const ReelMagic_PlayerAttributes& GetAttrs() const     = 0;
 };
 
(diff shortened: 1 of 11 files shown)


FeralChild64 (approved):
Looks OK to me.

kcgen (approved):
Looks good to me, too. 
That was quick!

johnnovak:
> Looks good to me, too. That was quick!

I'm excited! This is gonna be a great feature! 😎

Issue #2734, opened 2023-08-06
Fix video capture regressions


johnnovak:
I broke video capturing a little bit... Now we're capturing the rendered frames, which might be pixel and scan-doubled. We need to "divide that back" to the video mode's dimensions, just like I'm doing in the `raw` image capture mode.

This is an oversight and I know what to do — this is just a placeholder ticket for myself.

Will fix https://github.com/dosbox-staging/dosbox-staging/issues/2813 and https://github.com/dosbox-staging/dosbox-staging/issues/2812

PR #2735, opened 2023-08-06
Introduce PixelFormat enum

johnnovak:
Ok, so learning from my recent experience with clinically obese PRs 😅, from now on I'm trying to raise multiple smaller ones for refactoring jobs that can be well-separated.

The `bpp` (bits per pixel) thing has been bothering me for quite some time now — it really should be an enum, and it should be a pixel format descriptor rather than just a number. Well, now it is 😎 

Tested with a few games (CGA/EGA/VGA/Tandy), demos, and Return to Zork (to make sure I did not break ReelMagic that had some weird `bpp == 0` init hack that luckily wasn't doing anything useful, apparently), plus checked the image & video capture features for regressions. Everything is hunky-dory 😄 (except for https://github.com/dosbox-staging/dosbox-staging/issues/2734, but I broke that one a while ago; that's unrelated to this change).

### DIFF
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -51,6 +51,18 @@
 Render_t render;
 ScalerLineHandler_t RENDER_DrawLine;
 
+const char* to_string(const PixelFormat pf)
+{
+	switch (pf) {
+	case PixelFormat::Indexed8: return "Indexed8";
+	case PixelFormat::BGR555: return "BGR555";
+	case PixelFormat::BGR565: return "BGR565";
+	case PixelFormat::BGR888: return "BGR888";
+	case PixelFormat::BGRX8888: return "BGRX8888";
+	default: assertm(false, "Invalid pixel format"); return {};
+	}
+}
+
 static void render_callback(GFX_CallBackFunctions_t function);
 
 static void check_palette(void)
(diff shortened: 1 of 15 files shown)


src/gui/render.cpp
```diff
@@ -51,6 +51,22 @@
...
+	case PixelFormat::BGR555: return "BGR555";
+	case PixelFormat::BGR565: return "BGR565";
+	case PixelFormat::BGR888: return "BGR888";
+	case PixelFormat::BGRX8888: return "BGRX8888";
+	default: assertm(false, "Invalid pixel format"); return {};
+	}
+}
+
+uint8_t get_bits_per_pixel(const PixelFormat pf) {
+	return enum_val(pf);
```
FeralChild64, line 67: I'm not sure if essentially casting the enum this is a good idea. What if we have to introduce `RGB888` in addition to `BGR888` in the future?

johnnovak: Then we'll rethink the approach 😄 Unlikely, though.

But yeah I could remove the value assignments from the enums and just do another switch/case block here. I'm like 50/50, it's doesn't matter much for now.

FeralChild64 (approved):

PR #2739, opened 2023-08-08
Rendering related refactorings

johnnovak:
A couple of tactical chess moves so I can fit in the shader auto-switching stuff in a clean way. I thought I'd raise these separately as they stand on their own and improve the code in general.

The logging of the current video mode has also been improved, now we have stuff like these:

- "CGA 640x200 16-colour text mode 03h"
- "EGA 640x350 16-colour graphics mode 10h"
- "VGA 720x400 16-colour text mode 03h"
- "VGA 320x200 256-colour graphics mode 13h"
- "VGA 360x240 256-colour graphics mode"
- "VESA 800x600 256-colour graphics mode 103h"

No functional changes apart from that. Definitely review by commit.

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3358,9 +3358,9 @@ static void setup_window_sizes_from_conf(const char* windowresolution_val,
 	        sdl.display_number);
 }
 
-static SDL_Rect calc_viewport(const int canvas_width, const int canvas_height,
-                              const int draw_width, const int draw_height,
-                              const double draw_scalex, const double draw_scaley)
+SDL_Rect GFX_CalcViewport(const int canvas_width, const int canvas_height,
+                          const int draw_width, const int draw_height,
+                          const double draw_scalex, const double draw_scaley)
 {
 	assert(draw_width > 0);
 	assert(draw_height > 0);
@@ -3446,12 +3446,12 @@ static SDL_Rect calc_viewport(const int canvas_width, const int canvas_height,
 
 static SDL_Rect calc_viewport(const int canvas_width, const int canvas_height)
 {
-	return calc_viewport(canvas_width,
-	                     canvas_height,
-	                     sdl.draw.width,
-	                     sdl.draw.height,
-	                     sdl.draw.scalex,
-	                     sdl.draw.scaley);
+	return GFX_CalcViewport(canvas_width,
+	                        canvas_height,
+	                        sdl.draw.width,
+	                        sdl.draw.height,
+	                        sdl.draw.scalex,
+	                        sdl.draw.scaley);
 }
 
 void GFX_SetIntegerScalingMode(const std::string& new_mode)
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3358,23 +3358,26 @@ static void setup_window_sizes_from_conf(const char* windowresolution_val,
 	        sdl.display_number);
 }
 
-static SDL_Rect calc_viewport(const int win_w, const int win_h)
+static SDL_Rect calc_viewport(const int canvas_width, const int canvas_height,
+                              const int draw_width, const int draw_height,
+                              const double draw_scalex, const double draw_scaley)
 {
-	assert(sdl.draw.width > 0);
-	assert(sdl.draw.height > 0);
-	assert(sdl.draw.scalex > 0.0);
-	assert(sdl.draw.scaley > 0.0);
-	assert(std::isfinite(sdl.draw.scalex));
-	assert(std::isfinite(sdl.draw.scaley));
+	assert(draw_width > 0);
+	assert(draw_height > 0);
+	assert(draw_scalex > 0.0);
+	assert(draw_scaley > 0.0);
+	assert(std::isfinite(draw_scalex));
+	assert(std::isfinite(draw_scaley));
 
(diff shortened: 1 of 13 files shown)


src/gui/sdlmain.cpp
```diff
@@ -3401,51 +3404,56 @@ static SDL_Rect calc_viewport(const int win_w, const int win_h)
...
 		if (integer_scale_factor < 1) {
 			// Revert to fit to viewport
 			std::tie(view_w, view_h) = calculate_bounded_dims();
 		} else {
 			// Calculate the final viewport.
-			view_w = sdl.draw.width * integer_scale_factor;
-			view_h = iround(sdl.draw.height * integer_scale_factor * pixel_aspect);
+			view_w = draw_width * integer_scale_factor;
+			view_h = iround(draw_height * integer_scale_factor *
+			                pixel_aspect);
```
kcgen, line 3422: Just trying to understand the nuances of the various roundings (we've got `round()`, `static_cast<int>(...)`, and `iround()` all used here).

For the `static_cast<int>(bounds_h / (draw_height * pixel_aspect)` case, is this where we really need the results like `0.999` to be cast back down to a `0` integer instead of rounding up to `1` (if say, `iround` were used?).

src/hardware/vga.cpp
```diff
@@ -141,10 +141,10 @@ void VGA_LogInitialization(const char *adapter_name,
                            const char *ram_type,
                            const size_t num_modes)
 {
-	const auto mem_in_kib = vga.vmemsize / 1024;
+	const auto mem_in_kb = vga.vmemsize / 1024;
 	LOG_INFO("VIDEO: Initialised %s with %d %s of %s supporting %d modes",
-	         adapter_name, mem_in_kib < 1024 ? mem_in_kib : mem_in_kib / 1024,
-	         mem_in_kib < 1024 ? "KB" : "MB", ram_type,
+	         adapter_name, mem_in_kb < 1024 ? mem_in_kb : mem_in_kb / 1024,
+	         mem_in_kb < 1024 ? "KB" : "MB", ram_type,
```
kcgen, line 66: No more kibby bytes for us :bowl_with_spoon: :laughing:

src/ints/int10.h
```diff
@@ -48,8 +48,8 @@
 #define BIOSMEM_CRTCPU_PAGE   0x8A
 #define BIOSMEM_VS_POINTER    0xA8
 
-constexpr uint16_t MinVesaMode = 0x100;
-constexpr uint16_t MaxVesaMode = 0x7ff;
+constexpr uint16_t MinVesaBiosModeNumber = 0x100;
+constexpr uint16_t MaxVesaBiosModeNumber = 0x7ff;
```
kcgen, line 52: Nice clarification :+1:

src/hardware/vga.cpp
```diff
@@ -196,7 +115,64 @@ void VGA_DetermineMode(void) {
...
+
+// Return a human-readable description of the video mode, e.g.:
+//   - "CGA 640x200 16-colour text mode 03h"
+//   - "EGA 640x350 16-colour graphics mode 10h"
+//   - "VGA 720x400 16-colour text mode 03h"
+//   - "VGA 320x200 256-colour graphics mode 13h"
+//   - "VGA 360x240 256-colour graphics mode"
+//   - "VESA 800x600 256-colour graphics mode 103h"
+std::string to_string(const VideoMode& video_mode)
+{
```
kcgen, line 157: Very nice overhaul here, @johnnovak !

kcgen (approved):
Looks great!
Easy to review; and just one comment.

src/hardware/vga.cpp
johnnovak: Yeah it never really caught on :)

src/gui/sdlmain.cpp
johnnovak: In essence, we're dealing with two concepts here:

- rounding numbers, 
- and converting a number to its immediately smaller integer, which is the floor operation.

Both are needed here.

`iround` is just a shortcut so you don't need to cast to an integer type after `round` or `roundf`. I'd say we should introduce a similar `ifloor` helper function as that's a lot more descriptive than `static_cast<int>(...)`.

Now, there's a difference between casting a floating-point number to an integer type vs using the `floor` function:

```c
#include <math.h>
#include <stdio.h>

int main() {
  printf(" (int) 1.6  = %2d\n", (int) 1.6);
  printf(" (int)-1.6  = %2d\n", (int)-1.6);
  printf("floor( 1.6) = %2g\n", floorf( 1.6));
  printf("floor(-1.6) = %2g\n", floor(-1.6));
}
```

Output:

```
% gcc test.c && ./a.out                                                                        [cmd]
 (int) 1.6  =  1
 (int)-1.6  = -1
floor( 1.6) =  1
floor(-1.6) = -2
```

So casting to `int` takes the closest integer closer to zero, while `floor` the closest smaller integer. If we only dealing with positive numbers, this does not matter, which is the case here. 

I will introduce an `iround` helper function that adheres to the `floor` contract.

kcgen: Thanks for the explanation - all makes sense!
That `ifloor` is going to be a big self-documenting improvement about the intent :+1:

johnnovak: Done 😎

PR #2740, opened 2023-08-08
Fix const issue writing a png with png12

sulix:
When building against libpng v1.2 (png12), compilation fails due to png_write_row accepting a 'png_bytep' (unsigned char*), where we provide data from a const_iterator:
```
../src/capture/image/png_writer.cpp: In member function ‘void PngWriter::WriteRow(std::vector<unsigned char>::const_iterator)’:
../src/capture/image/png_writer.cpp:245:32: error: invalid conversion from ‘const unsigned char*’ to ‘png_bytep’ {aka ‘unsigned char*’} [-fpermissive]
  245 |         png_write_row(png_ptr, &*row);
      |                                ^~~~~
      |                                |
      |                                const unsigned char*
In file included from /usr/include/zlib.h:34,
                 from /usr/include/libpng12/png.h:317,
                 from ../src/capture/image/png_writer.h:33,
                 from ../src/capture/image/png_writer.cpp:24:
/usr/include/libpng12/png.h:1646:39: note:   initializing argument 2 of ‘void png_write_row(png_structp, png_bytep)’
 1646 | extern PNG_EXPORT(void,png_write_row) PNGARG((png_structp png_ptr,
      |                                       ^~~~~~
```

It's safe to cast away the const issue, as png_write_row() doesn't write to the row pointer passed in.

Even though png12 is quite old, it's used by default if installed (at least on openSUSE), so failing to support it is breaking the default build.

### DIFF
--- a/src/capture/image/png_writer.cpp
+++ b/src/capture/image/png_writer.cpp
@@ -242,7 +242,7 @@ void PngWriter::WritePngInfo(const uint16_t width, const uint16_t height,
 void PngWriter::WriteRow(std::vector<uint8_t>::const_iterator row)
 {
 	assert(png_ptr);
-	png_write_row(png_ptr, &*row);
+	png_write_row(png_ptr, const_cast<png_bytep>(&*row));
 }
 
 void PngWriter::FinalisePng()


src/capture/image/png_writer.cpp
```diff
@@ -242,7 +242,7 @@ void PngWriter::WritePngInfo(const uint16_t width, const uint16_t height,
 void PngWriter::WriteRow(std::vector<uint8_t>::const_iterator row)
 {
 	assert(png_ptr);
-	png_write_row(png_ptr, &*row);
+	png_write_row(png_ptr, const_cast<png_bytep>(&*row));
```
kcgen, line 245: Thanks for this fix up, @sulix.

It would be nice to operate on PNG's desired type directly, given this is a PNG-specific class ("PngWrite"), which also makes good use of other PNG-data types in its members.

In other words - because it's PNG-specific, we don't have try to be 'generic' so to speak on its interfaces (well, at least just at this point).

weirddan455:
>  Even though png12 is quite old, it's used by default if installed (at least on openSUSE), so failing to support it is breaking the default build.

I think this should be fixed if possible.  Do you also have a newer version of libpng installed?  If so, we should probably be linking against that by default.  I tried to reproduce on Arch Linux.  I have libpng 1.6.40 installed which is required by lots of packages.  `libpng12` (version 1.2.59) also exists in Arch repos (was not installed on my system).  I installed `libpng12` to test and re-built but it still linked against 1.6.  I'm not sure why the situation is different on OpenSUSE.

Anyway, this patch is probably fine so we can support v1.2 but maybe we should look into why it's linking to that if there's a newer version also on the system.

kcgen:
@sulix , curious if this branch builds for you? `kc/png-writer-interface-1`

src/capture/image/png_writer.cpp
kcgen: (I've addressed my comment in a PR https://github.com/dosbox-staging/dosbox-staging/pull/2741)

kcgen:
@weirddan455 , yes - 1.2.x is still in wide use and its interfaces are 1.x-compatible (well, I guess /mostly/ compatible besides casting like this). 

We do have a libpng as a wrap, so one work-around would be to crank up the minimum allowed version in meson.build:

```python
 png_dep = dependency(
    'libpng',
    version: ['>= 1.2', '< 2'],
    static: ('png' in static_libs_list or prefers_static_libs),
    include_type: 'system',
)
```

To say, 1.6, and then (most) users would be compiling and linking statically with the wrap.  That's a bit heavy handed though given there's nothing in 1.6 we strictly need.. 1.2 is functionally correct and reliable.  So I'd rather get our usage of the PNG API even tighter to the spec.

kcgen:
Yup - also agree @weirddan455! Meson should use the newest system lib available that meets the version band (newer being safer/faster/bug-fixed, etc..).

Also - sorry, didn't mean to imply you wanted to force 1.6.x (I was thinking about ways we could do it in Meson).

src/capture/image/png_writer.cpp
johnnovak: Going with this original solution then, as discussed.

johnnovak (approved):
Nice fix @sulix , thanks for that 🎉

sulix:
> @sulix Can you confirm if you have multiple libpng versions installed?

Yeah: I have both libpng 1.6.50  and 1.2.59 installed, but libpng12 is configured as the default, i.e.:
```
david@sparky:~> pkg-config --libs --cflags libpng
-I/usr/include/libpng12 -lpng12 
david@sparky:~> pkg-config --libs --cflags libpng12
-I/usr/include/libpng12 -lpng12 
david@sparky:~> pkg-config --libs --cflags libpng16
-I/usr/include/libpng16 -lpng16 
```
And ``/usr/lib64/libpng.so`` is a symlink to ``libpng12.so``

Since libpng12 is the system default, I think it makes sense for dosbox to use it rather than override it with a newer version, if that doesn't cause any further problems.

CMake has had issues with this in the past, too:
https://gitlab.kitware.com/cmake/cmake/-/issues/24348

weirddan455:
Ok, I guess that explains it.  Arch just has libpng16 as the default.  Maybe it doesn't make sense to try to override the defaults in our build system then?  Not really sure.  Regardless, good job getting the build fixed on OpenSUSE.

```
[daniel@arch ~]$ pkg-config --libs --cflags libpng
-I/usr/include/libpng16 -lpng16 
[daniel@arch ~]$ ls -lh /usr/lib/libpng.so
lrwxrwxrwx 1 root root 11 Jun 29 04:29 /usr/lib/libpng.so -> libpng16.so
```

kcgen:
Excellent - this looks perfectly reasonable @weirddan455 and @sulix (that the default is being adhered to in both cases).  Thanks for the fix, @sulix (and not worries on other PR, @johnnovak!)

PR #2741, opened 2023-08-08
Use the PNG native row type in png-specific write functions

kcgen:
This PR lets the `PngWriter` row-write function takes in PNG's row type directly ([`png_bytep`](https://refspecs.linuxbase.org/LSB_3.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.write.row.1.html)), which avoids dereferencing and casting from iterators as @sulix [reported](https://github.com/dosbox-staging/dosbox-staging/pull/2740).

The `PngWriter`'s job is to interface w/ the PNG library, and using PNG's API-defined `png_bytep` type puts PNG in charge of the type we need accept (making the interface directly compliant with whatever the PNG header needs).

From the looks of it, `png_bytep` can be a `const uint8_t*` or non-const, both of which are supported by vector's `.data()` output.

It adjusts `ImageScaler::GetNextOutputRow`'s interface to use a pointer to the vector's contiguous data (and vector's `.data()` interface), without having to inject any PNG-specific details into the scaler.

---

It would be nice if there was a C++ binding and we could use iterators more naturally with PNG, but one draw back is iterators aren't guaranteed to be contiguous (just in general) - where as memory blocks are.  So this is probably a better natural fit in the future too if we eventually add another lossless format, say, JPEG-XL (if/when it's widely supported).

### DIFF
--- a/src/capture/image/image_saver.cpp
+++ b/src/capture/image/image_saver.cpp
@@ -149,9 +149,9 @@ static void write_upscaled_png(FILE* outfile, PngWriter& png_writer,
 		break;
 	};
 
-	auto rows_to_write = image_scaler.GetOutputHeight();
-	while (rows_to_write--) {
-		auto row = image_scaler.GetNextOutputRow();
+	// Write the rows
+	png_bytep row = {};
+	while ((row = image_scaler.GetNextOutputRow())) {
 		png_writer.WriteRow(row);
 	}
 }
@@ -215,7 +215,7 @@ void ImageSaver::SaveRawImage(const RenderedImage& image, const VideoMode& video
 				*out++ = pixel.blue;
 			}
 		}
-		png_writer.WriteRow(row_buf.begin());
+		png_writer.WriteRow(row_buf.data());
 		image_decoder.AdvanceRow();
 	}
 }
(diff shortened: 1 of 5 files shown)


johnnovak:
This works and it is not the worst thing to do, but I think I'd prefer to do leave the interface of the image saver and scaler alone and only cast the iterator to a pointer at the last minute in the PNG writer before calling the row write function (so that would accept `png_bytep`, but only that function).

Iterators doesn't guaranteeing contiguousness is only a problem if you write very generic algorithms that can operate on, say, both vectors and deques. But here `std::vector<uint8_t>::const_iterator ` stares you in the face, so I see zero harm casting that with whatever force necessary to a pointet within the PNG writer. The chance that we'll ever switch to non-contiguous containers to store image row data in the future is effectively zero, making that approach very safe 😄 

So yeah, I'd just like to minimise pointer usage, but it's a small thing. Btw, in C++20 we could use `contiguou_iterator` to avoid the forced cast.

kcgen:
Ok; let's toss it! :laughing:

johnnovak:
Happy to do this myself later @kcgen 😄

kcgen:
> Happy to do this myself later @kcgen 😄

Sulix's PR has it done :-) https://github.com/dosbox-staging/dosbox-staging/pull/2740

johnnovak:
> > Happy to do this myself later @kcgen 😄
> 
> Sulix's PR has it done :-) #2740

Ah, cool. I'm sorry that you have wasted time on doing this, I just prefer to keep the interfaces clean.
This actually looks very straightforward and 100% safe: we're guaranteeing the iterator points to the contents of a vector, so the data if 100% guaranteed to be contiguous:

image

Merging Sulix's PR then 😄

PR #2742, opened 2023-08-09
Fix compatibility problems with VBADOS (VBMOUSE.EXE) mouse driver

FeralChild64:
- fixed PS/2 packet data loss problems when BIOS was receiving the next packet before the previous one was fully retrieved (due to this problem VBMOUSE.EXE was completely broken if emulated PS/2 mouse supported wheel)
- prevented unnecessary dummy mouse events in VMware mouse protocol
- added support for PS/2 mouse packet size 1 to the BIOS - VBADOS seems to prefer it
- silenced warning in our internal DOS driver, which was a result of VBADOS checking if it is already installed

Driver link: https://git.javispedro.com/cgit/vbados.git/about/ (changes are related to `VBMOUSE.EXE` DOS driver, not to the Windows driver)

### DIFF
--- a/src/hardware/input/mouseif_ps2_bios.cpp
+++ b/src/hardware/input/mouseif_ps2_bios.cpp
@@ -883,24 +883,37 @@ bool MOUSEBIOS_CheckCallback()
 	return true;
 }
 
+static bool is_bios_frame_size_supported(const uint8_t frame_size)
+{
+	return (frame_size == 1) || // VBADOS (VBMOUSE.EXE) preferred size
+	       (frame_size == 3) || // standard size, has to be supported
+	       (frame_size == 4);   // CuteMouse 2.1 uses this for wheel mice
+}
+
 void MOUSEBIOS_DoCallback()
 {
-	assert(bios_frame_size == 1 || bios_frame_size == 3 || bios_frame_size == 4);
+	assert(is_bios_frame_size_supported(bios_frame_size));
 	assert(bios_buffer.size() >= bios_frame_size);
 
-	if (bios_frame_size == 3) {
+	switch (bios_frame_size) {
+	case 1:
+		CPU_Push16(bios_buffer[0]);
+		CPU_Push16(0);
+		CPU_Push16(0);
+		break;
+	case 3:
 		CPU_Push16(bios_buffer[0]);
 		CPU_Push16(bios_buffer[1]);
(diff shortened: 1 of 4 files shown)


src/hardware/input/mouseif_ps2_bios.cpp
```diff
@@ -926,7 +986,10 @@ void MOUSEBIOS_Subfunction_C2() // INT 15h, AH = 0xc2
 		set_return_value(BiosRetVal::Success);
 		break;
 	case 0x05: // initialize
-		if (reg_bh == 3 || reg_bh == 4) {
+		if (reg_bh == 1 || reg_bh == 3 || reg_bh == 4) {
```
johnnovak, line 989: Would it be better to extract the supported frame sizes of 1, 3, and 4 to constants?

johnnovak (approved):
Just one minor comment @FeralChild64 , nice work as always 👍🏻

src/hardware/input/mouseif_ps2_bios.cpp
FeralChild64: I don't think constants would improve readability - but I have extracted the check to a dedicated function, with comments explaining why we support each and every one of the frame sizes.

Grounded0:
Hint for anyone who wants to use this but doesn't want to deal with compatibility problems that might occur by keeping this loaded outside of Windows 3.1x:

```
$ cat WIN.BAT
config -set cputype=pentium_slow
c:\vbados\vbmouse.exe
win.com
c:\vbados\vbmouse.exe uninstall
config -set cputype=auto
```

FeralChild64:
This driver is mainly useful if booting real MS-DOS. We already provide what is needed for the Windows INT33 driver.

Grounded0:
I have a little bit of problems when working with MS-DOS Prompt inside Windows 3.1x. Mainly tracking and mouse click registration problems. When this driver is loaded its smooth sailing even with MS-DOS Prompt running.

Am i right stating that you can disregard `VBMOUSE.EXE` if you don't use MS-DOS prompt inside Windows 3.1x? It could be that i'm doing something wrong or does loading the FreeDOS `COMMAND.COM` cause problems.

Can you personally notice this or should i file a ticket about it?

FeralChild64:
@Grounded0 I can see the mouse does nit behave correctly within MS-DOS Prompt window, but I think it would be a good idea to have an issue for this.

PR #2744, opened 2023-08-09
Fix invalid default config value for voodoo_memsize

weirddan455:
The default value was "12mb" but valid values do not include "mb"

This fixes a warning: CONFIG: '12mb' is an invalid value for
'voodoo_memsize', using the default: '12mb'

Changed default from "12mb" to "12" to correct this.


kcgen:
nice quick fix, @weirddan455 !

Issue #2468, opened 2023-05-10
Preserve mixer settings per device when changing 'mididevice' between fluidsynth/mt-32/...


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Mac OS

### Is your feature request related to a problem? Please describe.

Currently if you have fluidsynth set as the active mididevice and adjusted the fluidsynth volume using the mixer and a 'config -set mididevice=fluidsynth' command restores the volume back to 100 because default values are loaded.

Also loading e.g. mt-32 resets volume settings of fluidsynth and mt-32.

The problem is that eXoDOS for example always does switch the mididevice before a game is started, therefore volume is always 100 currently.

### Describe the solution you'd like

'config -set mididevice=XXX' should remember last volume set per device.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Thanks for raising this, @Burrito78. I'm aware of this issue but couldn't be bothered to fix it as it can get messy. But the eXoDOS use case turns this into a more pressing issue, agreed.

Could you please generalise the description that the whole mixer state (all channels) must be preserved when switching MIDI devices, including all mixer channel settings, so volume, crossfeed, chorus & reverb levels, and stereo reverse? From memory, all those settings are reset too.

Once we have a generic mechanism that enables us to save/restore the whole mixer state, we only need to identify other scenarios where the mixer params are reset, and simply just plug in the same mechanism to fix them.

Thanks!

Burrito78:
Set everything to 99 to see changes easily.
Ran Dark Forces (the batchfile runs 'CONFIG -set "mididevice=fluidsynth"'.

See mixer before and after:
![Bildschirmfoto 2023-05-13 um 13 01 51](https://github.com/dosbox-staging/dosbox-staging/assets/61995128/42382568-4e65-44d9-9dbf-77e04c21e49a)
![Bildschirmfoto 2023-05-13 um 13 02 21](https://github.com/dosbox-staging/dosbox-staging/assets/61995128/c361b264-0c0a-4782-ae1a-193a690507e6)

Burrito78:
Why r54 and c54 though??

johnnovak:
Because you must have set some chorus and reverb preset, and on mixer reset the values are reset to the preset defaults.

Burrito78:
[...]

PR #2746, opened 2023-08-09
Control the startup and shutdown order of our global and file-scoped objects

### LINKED ISSUE #2726: Defender of the Crown Tandy version crashes in 16-colour Tandy mode

0.81.0-alpha (current main)

Mac OS

_No response_

### Relevant hardware info

_No response_

Defender of the Crown Tandy version crashes when pressing "any key" on startup to run in 16-colour Tandy mode. CGA mode work by pressing the F1 key.

When attempting to run in 16-colour Tandy mode, you either get graphical glitches and DOSBox hangs, or it hard exits with a stack trace after a few seconds.

Here's how the graphical glitch looks like:

image

This is not a regression; SVN behaves the same way and 0.79.1 too.

1. Use Tandy version of Defender of the Crown and press any key at startup (or use the attached repro pack).
2. The hard crash or graphical corruption will always happen.

_No response_

### Your configuration

```ini
[dosbox]
machine = tandy

[autoexec]
mount C "C"
c:
cd EGA
DEFENDER
exit
```

### Provide a Log

```text
The stack trace dump is always a bit different, here are two variations:


Stack trace:
15         0x1048dd08c 20  dyld                                0x00000001048dd08c start + 520
14         0x100644b50 main + 36
13         0x100a05b90 sdl_main(int, char**) + 4668
12         0x1006d3498 Config::StartUp() + 28
11         0x100d907fc SHELL_Init() + 3992
10         0x100d8db04 DOS_Shell::Run() + 1472
9          0x100d8d488 DOS_Shell::RunBatchFile() + 300
8          0x100d8d03c DOS_Shell::ParseLine(char*) + 2008
7          0x100da3bfc DOS_Shell::DoCommand(char*) + 440
6          0x100dd20d0 DOS_Shell::Execute(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 996
5          0x100dd2fb8 run_binary_executable(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 1556
4          0x10078c480 CALLBACK_RunRealInt(unsigned char) + 92
3          0x100645548 DOSBOX_RunMachine() + 28
2          0x100645448 Normal_Loop() + 44
1          0x1007c619c CPU_Core_Normal_Run() + 65288
0          0x100706bbc unsigned long static_if_array_then_zero<char [1024]>() + 0
2023-08-04 20:52:36.914 | ABORT: Illegal GRP4 Call 2
```

```
Stack trace:
15         0x1067e50d8 19  dyld                                0x00000001067e50d8 start + 596
14         0x19bde8ec4 dyld4::LibSystemHelpers::getenv(char const*) const + 0
13         0x19bcc7b74 exit + 44
12         0x19bcc7dd0 __cxa_finalize_ranges + 464
11         0x102e0c1d8 std::unique_ptr<MPU401, std::default_delete<MPU401> >::~unique_ptr() + 28
10         0x102e0c3cc std::unique_ptr<MPU401, std::default_delete<MPU401> >::~unique_ptr() + 32
9          0x102e0c43c std::unique_ptr<MPU401, std::default_delete<MPU401> >::reset(MPU401*) + 96
8          0x102e0c4c0 std::default_delete<MPU401>::operator()(MPU401*) const + 44
7          0x102e0c51c MPU401::~MPU401() + 28
6          0x102e0c650 MPU401::~MPU401() + 292
5          0x102db1944 IO_WriteHandleObject::Uninstall() + 60
4          0x102db19a0 IO_FreeWriteHandler(unsigned short, io_width_t, unsigned short) + 68
3          0x102db2864 std::unordered_map<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)>, std::hash<unsigned short>, std::equal_to<unsigned short>, std::allocator<std::pair<unsigned short const, std::function<void (unsigned short, unsigned int, io_width_t)> > > >::erase(unsigned short const&) + 32
2          0x102dbd388 unsigned long std::__hash_table<std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, std::__unordered_map_hasher<unsigned short, std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, std::hash<unsigned short>, std::equal_to<unsigned short>, true>, std::__unordered_map_equal<unsigned short, std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, std::equal_to<unsigned short>, std::hash<unsigned short>, true>, std::allocator<std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> > > >::__erase_unique<unsigned short>(unsigned short const&) + 36
1          0x102dbb4e0 std::__hash_iterator<std::__hash_node<std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, void*>*> std::__hash_table<std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, std::__unordered_map_hasher<unsigned short, std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, std::hash<unsigned short>, std::equal_to<unsigned short>, true>, std::__unordered_map_equal<unsigned short, std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> >, std::equal_to<unsigned short>, std::hash<unsigned short>, true>, std::allocator<std::__hash_value_type<unsigned short, std::function<void (unsigned short, unsigned int, io_width_t)> > > >::find<unsigned short>(unsigned short const&) + 104
0          0x19bdf34a4 _sigtramp + 56
2023-08-04 20:53:49.266 | Signal: SIGSEGV
```
```


- [X] Yes, I agree.

### PR DESCRIPTION
**Pull request by kcgen**
- Fixes #2726, specifically the scenario where we segfault with memory errors.

Turns out the reason for this was because we (currently) have no control over the shutdown order of global and file-scoped variables during abnormal exits. Suggest reading this: https://isocpp.org/wiki/faq/ctors#static-init-order

This PR brings that under control by controlling the startup order of our module instances, and with that brings order to their shutdown (which happens in reverse). 

More explanation below!

---

During startup in `dosbox.cpp`, each module's `_Init()` function is registered with the Config system in order of dependence. Things like `Memory_Init()`, `Paging_Init()`, etc.. are registered before the modules that rely on them.

- **At launch**, the Config object executes the `_Init()`s in the same order.

- **During runtime**, modified conf settings have their respective `_Destroy()` and `_Init()` functions run, essentially re-init'ing them.

- **At shutdown**, there are two scenarios:

    1. **Nominal or sunny-day**: the Config object runs any registered `_Destroy()` functions in reverse order, after which the config object itself goes out of scope followed by C++'s at-exit registry destructing all static objects in reverse startup order followed again by non-static globals.

    2. **Abnormal or immediate termination** (`E_Exit`): we jump to the at-exit point where all static objects are destructed in reverse startup order followed again by non-static globals.

There are some problems with this:

- We have no control over the startup order of globals and file-scoped objects, which means we have no control over their shutdown order. See the C++ FAQ's "static initialization order fiasco problem": https://isocpp.org/wiki/faq/ctors#static-init-order

- The `_Destroy()` functions give the appearance of orderliness when in fact they are helpless to control the shutdown order of global and file-scoped objects atexit time, leading to rugs being pulled during the abnormal shutdown sequence like we see when running the known-corrupt Defender of the Crown game.

**The solution** is to move globals and file-scoped objects into their respective modules' `_Init()` functions and therefore enforce a startup order on-demand by the `_Init()` functions. This means they're destroyed in reverse-`Init()` order, which is as good as it gets. Second: we can eliminate our reliance on the `_Destroy()` functions, which means that both the nominal and abnormal shutdown pathways effectively collapse into into one (or very close to the same).

This is the first in a series of commits that perform this change, module-by-module, until both the the shutdown pathways are unwound in correct `_Init()` order.

### CODE REVIEW

**Comment by johnnovak** [comment]
Amazing cleanup job @kcgen; global  initialisation order is evil (and static, which are really just globals in disguise).

In my own apps generally I have a _single_ global `App` object, then that contains all other stuff as members that would be "normally" be other global variables. That's one way to regain full control over the initialisation order.

Will review in detail later today.

**Comment by johnnovak** [comment]
I'm also thinking we should start a "for devs only" wiki section that describes certain architectural aspects that are not immediately obvious, i.e. how the configuration system and the (re)init on setting changes works (basically we could start with a slightly re-worked version of your above description of the current situation).

**Comment by kcgen** [comment]
Great suggestion! we could even have an "example module" that tries to do everything "right", with some example code and comments. It could also have some examples for constants, enums, classes, IO-port registration, etc..

**Comment by kcgen** [comment]
> global initialisation order is evil (and static, which are really just globals in disguise).

It sure is ugly..  this wrestles (most!) of the major modules' globals out of file-space and into the their Init() functions; it's a start, anyway!

There are still hundreds (thousands?) more "lesser" variables littered in file-scope space, but the majority of them are just scalars or containers that don't have cross-module dependencies. Many of these are more C-based, and ideally those modules would be bundled up more as a class and those variables moved inside as members.  Oh well.

Working through the PR did reveal a couple corner cases where we had some use-after-free and double-free, and perhaps with testing we'll find more. I've tried to extensively test audio, video, serial, network, and runtime config-changes, and so far it's OK.

File: src/misc/setup.cpp [hunk 1]
```diff
@@ -1795,3 +1796,11 @@ void Config::ParseArguments()
 
 	arguments.editconf = cmdline->FindRemoveOptionalArgument("editconf");
 }
+
+Config* SETUP_CreateConfig(const int argc, char* argv[])
+{
+	static auto command_line = CommandLine(argc, argv);
+
+	static auto config = Config(&command_line);
+	return &config;
```
weirddan455, line 1805: Aren't you returning a pointer to an object that's about to go out of scope here?

weirddan455: Oh, I somehow missed the static keyword.  Ignore me.

src/midi/midi.cpp
```diff
@@ -88,43 +88,49 @@ uint8_t MIDI_message_len_by_status[256] = {
 #include "midi_alsa.h"
 #endif
 
-static std::list<std::unique_ptr<MidiHandler>> handlers = {};
+// Pointer to the static handlers list instance
+using handlers_list_t = std::list<std::unique_ptr<MidiHandler>>;
```
weirddan455, line 92: Minor, but is there a reason for `std::list` over `std::vector`?  I've seen it in other places in the code as well.  I get it if you don't want to mess with this in this PR but usually vector is faster and more memory efficient unless there's a good reason not to use it.

src/misc/setup.cpp
kcgen: All good 😎

weirddan455 (commented):
Just the one comment for now.  I'm not all that familiar with a lot of the code you touched.  I like your solution though:  Just return pointers to static local objects so the init and de-init is controlled.  Great cleanup job :+1:

johnnovak:
> There are still hundreds (thousands?) more "lesser" variables littered in file-scope space, but the majority of them are just scalars or containers that don't have cross-module dependencies. Many of these are more C-based, and ideally those modules would be bundled up more as a class and those variables moved inside as members. Oh well.

Yeah, many of those are benign, but the class-based approach would make things cleaner for sure.

> Working through the PR did reveal a couple corner cases where we had some use-after-free and double-free, and perhaps with testing we'll find more.

`std::free` handles double-frees fine in the sense that it ignores nulls (or hides these issues, which might be a problem, depending on your perspective 😅 ).

> I've tried to extensively test audio, video, serial, network, and runtime config-changes, and so far it's OK.

Nice one 🚀

src/midi/midi.cpp
johnnovak: Yes, I would recommend defaulting to vector and only deviating from that if there's a very good reason. Linked lists are very slow because contiguousness is not guaranteed and that trumps the need reallocate once in a while by _far_ (by a very very large margin). Realloc and copying memory is very fast and a non-issue, while non-local random memory access is a huge performance killer on modern processors.

I'm not even sure what a good use case for lists are these days. Maybe huge linked lists where you must do a lot of re-linking and re-ordering all the time, i.e. when your algorithm is really designed around unlinking/relinking, etc. But even in that case vectors under a few thousand elements (or maybe even a lot more) probably wouldn't even improve anything performance-wise.

But for non-performance intensive code I would definitely choose a list or dequeue or something similar if that makes the code more _readable_, just to make this clear 😄

weirddan455: >  I'm not even sure what a good use case for lists are these days.

I've used linked lists for implementing a queue (in C where I don't have the STL) because removing from the front is worst-case for a resizable array.  Even then, there's probably a better way (circular buffer or something?) because, like you said non-contiguous memory is a huge bottleneck because of cache misses.  But if you're using STL, you just have `std::queue` and `std::deque`.

I always do a double-take whenever I see `std::list` because there's usually a better data structure.  I don't think this is performance intensive code so, minor but I also don't think there's a difference in readability between `std::vector` and `std::list`.

johnnovak: To put it simply: random memory access is crap for utilising your cache effectively. That's an instant performance killer, like you also mentioned.

Whereas reallocating and recopying a vector is only done once in a while, and it's a _fast_ operations (going through memory linearly both at source and destination).

Net effect: the copying on realloc is negligible compared to paying the penalty of cache misses on _every single_ operation.

src/shell/shell_batch.cpp
```diff
@@ -86,9 +86,12 @@ std::string BatchFile::GetLine()
 		 */
 		if (data <= UnitSeparator && data != '\t' && data != '\b' &&
 		    data != Esc && data != '\n' && data != '\r') {
-			LOG_WARNING("BATCH: Encountered non-standard character: Dec %03u and Hex %#04x",
+			LOG_WARNING("BATCH: line '%s' encountered a non-standard character: "
```
johnnovak, line 88: Minor: How about "Non-standard character found in line '%s': decimal %03u (hex %04x);", then the rest.

src/shell/shell_batch.cpp
```diff
@@ -92,9 +91,21 @@ std::string BatchFile::GetLine()
...
 		}
+
+		// The maximum batch line length in MS-DOS was 1024 characters,
+		// which is a reasonable maximum.
+		// Ref: https://groups.google.com/g/alt.msdos.batch/c/z_H4JWPih-A/m/C499Mq4eJpAJ
+		//
+		constexpr size_t max_line_length = 1024;
+
+		if (line.length() >= max_line_length) {
+			LOG_WARNING("BATCH: line '%.40s (...)' exceeds the maximum length of 1024 characters; truncating",
```
johnnovak, line 103: Minor: "Line" (uppercase L)

`'%.40s (...)'` is a nice touch 👍🏻

include/dosbox.h
```diff
@@ -55,8 +55,7 @@ void DOSBOX_SetNormalLoop();
 void DOSBOX_Init(void);
 
 class Config;
-using config_ptr_t = std::unique_ptr<Config>;
-extern config_ptr_t control;
+extern Config* control;
```
johnnovak, line 58: Um, maybe out of scope, but could we call this `config` instead?

include/programs.h
```diff
@@ -37,7 +37,7 @@
 
 class CommandLine {
 public:
-	CommandLine(int argc, const char* const argv[]);
+	CommandLine(int argc, char* argv[]);
```
johnnovak, line 40: Why remove the consts?

src/misc/setup.cpp
```diff
@@ -45,7 +46,7 @@ extern char** environ;
 using namespace std;
 
 // Commonly accessed global that holds configuration records
-std::unique_ptr<Config> control = {};
+Config* control = {};
```
johnnovak, line 49: `config`?

src/hardware/memory.cpp
```diff
@@ -699,18 +700,10 @@ class MEMORY final : public Module_base {
 	}
 };
 
-static MEMORY* test;
```
johnnovak, line 702: "test" 🤦🏻 😐

src/cpu/cpu.cpp
```diff
@@ -2425,25 +2433,12 @@ class CPU final : public Module_base {
 	}
 };
 
-static CPU * test;
```
johnnovak, line 2428: "test"... Why not "apricot"? Or "truckDriver"? 🤣

src/hardware/sblaster.cpp
```diff
@@ -2358,18 +2358,21 @@ class SBLASTER final {
...
+	if (sec) {
+		sblaster_instance = std::make_unique<SBLASTER>(sec);
+	}
+}
+
+void SBLASTER_ShutDown(Section*)
+{
+	// shutdown with an empty section
+	constexpr Section* shutdown = nullptr;
+	SBLASTER_Init(shutdown);
```
johnnovak, line 2377: Uh, that's a bit weird. Can't we avoid that? Maybe extracting the common logic out from `SBLASTER_Init` then calling it from both Init and Shutdown?

src/hardware/tandy_sound.cpp
```diff
@@ -631,26 +630,66 @@ static void set_tandy_sound_flag_in_bios(const bool is_enabled)
...
-		tandy_psg.reset();
+		tandy_psg_instance.reset();
+	}
+
+	// Also reset the Tandy DAC with an empty section
+	constexpr Section* empty_section = nullptr;
+	toggle_dac(empty_section);
+
+	// If the section's empty, then there's nothing to configure
+	if (!section) {
```
johnnovak, line 689: I'm a bit confuse why this is needed. To guard against programming errors? Unless I'm wrong, `section` can only be null if someone removed the whole section from the config, so this should rather be an assert on top of the function, no?

src/hardware/tandy_sound.cpp
```diff
@@ -659,30 +698,28 @@ void TANDYSOUND_Init(Section *section)
...
 	set_tandy_sound_flag_in_bios(true);
+}
 
-	constexpr auto changeable_at_runtime = true;
-	section->AddDestroyFunction(&TANDYSOUND_ShutDown, changeable_at_runtime);
+void TANDYSOUND_ShutDown(Section*)
+{
+	// shutdown with an empty section
+	constexpr Section* empty_section = nullptr;
+	TANDYSOUND_Init(empty_section);
```
johnnovak, line 724: Ok... so that's how `TANDYSOUND_Init` can receive a null section.

I'm not in love with this as it feels very much like a hack. I'd rather have some code duplication here and make the shutdown process explicit. Surely not everything from init is needed for the shutdown. Or extract the commonallities between init and shutdown into a function, but frankly, that's a bit overdoing it. Some duplication and more explicitness in this case is fine as it would improve readability a lot.

src/hardware/tandy_sound.cpp
```diff
@@ -631,26 +630,66 @@ static void set_tandy_sound_flag_in_bios(const bool is_enabled)
...
+static ConfigProfile get_config_profile()
 {
+	switch (machine) {
+	case MCH_PCJR: return ConfigProfile::PCjrSystem; break;
+	case MCH_TANDY: return ConfigProfile::TandySystem; break;
+	default: return ConfigProfile::SoundCardOnly; break;
+	}
+}
+
+static void toggle_dac(Section* section)
```
johnnovak, line 642: Yeah... I don't like "dual-purposing" `section` so it can either be a kind of a flag to just disable the DAC (when it's null), or an actual section that holds the config value, then based on the config it will be either disabled or enabled.

The function is doing a bit too much. Could it be broken down into separate parts so it can be composed in a less tangled way? E.g. one function to parse the config settings and come up with a decision whether to enable or disable the DAC, then a second one to disable the DAC, and a third one to enable it.

`toggle_dac` has two invocations: one always disables the DAC, the other invocation either enables or disables it based on the config. So with three-function approach, the intent would be clearer and a lot more explicit — no more passing of "magic null sections" that in fact acts as "disable DAC":

- 1st invocation always calls `disable_dac`
- 2nd invocation calls `is_dac_enabled`, then based on the result it calls either `enable_dac` or `disable_dac`.

The less things are "entwined", the better, IMO 😄

src/hardware/opl.cpp
```diff
@@ -967,16 +968,22 @@ OPL::~OPL()
 	MIXER_DeregisterChannel(channel);
 }
 
-void OPL_ShutDown([[maybe_unused]] Section* sec)
+void OPL_ShutDown()
 {
-	opl = {};
+	constexpr Section* empty_section = nullptr;
+	OPL_Init(empty_section, OplMode::None);
```
johnnovak, line 974: Yeah, still don't like this 😄

src/hardware/opl.h
```diff
@@ -76,7 +76,7 @@ class Chip {
 typedef uint8_t RegisterCache[512];
 
 // Internal class used for dro capturing
-class Capture;
+class DroCapture;
```
johnnovak, line 79: Nice 👍🏻

src/hardware/gus.cpp
```diff
@@ -1713,7 +1707,9 @@ void GUS_AddConfigSection(Config* conf)
 
 	constexpr auto changeable_at_runtime = true;
 
-	Section_prop* sec = conf->AddSection_prop("gus", &gus_init, changeable_at_runtime);
+	Section_prop* sec = conf->AddSection_prop("gus",
+	                                          &toggle_gus,
```
johnnovak, line 1711: Isn't the name `toggle_gus` a bit misleading? I think I preferred `init_gus`, that makes it clearer this is really just the init function for the GUS section.

I guess "toggling" to me implies only turning something on or off, but `toggle_gus` does more than that: it gets all the GUS settings from the config, configures the GUS, etc. Especially then `toggle_gus` is called for the first time, it initialised the GUS. Then if I only change the DMA and IRQ settings, it doesn't _really_ toggle it, does it? Rather it reconfigures the GUS (which technically involved turning it off then back on, but still 😄 ).

include/dos_system.h
```diff
@@ -171,13 +171,24 @@ class DOS_Device : public DOS_File {
 	virtual bool WriteToControlChannel(PhysPt bufptr, uint16_t size,
 	                                   uint16_t* retcode);
 	virtual uint8_t GetStatus(bool input_flag);
+
+	// Indicate or check if the device is managed (or not). Managed devices
+	// are externally owned and managed per RAII. Unmanaged devices will be
+	// destroyed by the device registry on close.
+	void SetManaged();
+	void SetUnmanaged();
+	bool IsUnmanaged() const;
```
johnnovak, line 180: Nitpicking: `IsManaged`? Just ask for the positive case?

include/dos_system.h
```diff
@@ -414,10 +425,14 @@ typedef bool (MultiplexHandler)(void);
...
 
-/* AddDevice stores the pointer to a created device */
-void DOS_AddDevice(DOS_Device * adddev);
-/* DelDevice destroys the device that is pointed to. */
-void DOS_DelDevice(DOS_Device * dev);
+// Add (or remove) the device in the DOS device registry. Managed devices are
+// externally owned and managed per RAII. Unmanaged devices will be destroyed by
+// the device registry on close. Note that when adding, duplicate-named devices
+// are removed from the registry prior to adding because devices must be
+// uniquely named.
```
johnnovak, line 432: Nice explanatory comments 💯

include/dos_system.h
```diff
@@ -414,10 +425,14 @@ typedef bool (MultiplexHandler)(void);
...
+void DOS_AddUnmanagedDevice(DOS_Device* unmanaged_dev);
+void DOS_AddManagedDevice(DOS_Device* managed_dev);
```
johnnovak, line 434: 🎖️ for not just doing ` void DOS_AddDevice(DOS_Device* dev, const bool is_managed);`

src/dos/dos_devices.cpp
```diff
@@ -364,7 +365,22 @@ uint8_t DOS_Device::GetStatus(bool input_flag)
...
+{
+	is_managed = true;
+}
+
+void DOS_Device::SetUnmanaged()
+{
+	is_managed = false;
+}
+
+bool DOS_Device::IsUnmanaged() const
```
johnnovak, line 378: Please make it `IsManaged` 😄 🙏🏻

src/dos/dos_devices.cpp
```diff
@@ -436,28 +452,78 @@ uint8_t DOS_FindDevice(const char* name)
...
 }
 
-void DOS_DelDevice(DOS_Device * dev) {
-// We will destroy the device if we find it in our list.
-// TODO:The file table is not checked to see the device is opened somewhere!
-	for (Bitu i = 0; i <DOS_DEVICES;i++) {
-		if (Devices[i] && !strcasecmp(Devices[i]->GetName(), dev->GetName())) {
-			delete Devices[i];
-			Devices[i] = nullptr;
+void DOS_AddManagedDevice(DOS_Device* managed_dev)
```
johnnovak, line 482: `DOS_Device* const managed_dev`? Or would that be overdoing it? 😄

src/midi/midi.cpp
kklobe: Yes, violent agreement: start with `vector` and only change when profiling proves it's a real bottleneck.

The only two cases I can pop off the top of my head are some algorithm that has incredibly frequent updates in the middle of the list along with infrequent reads, or very large elements in the list that are expensive to move.

Similar with sorting. I see people arguing over quick sort or shell sort or heap sort... for an array with 35 elements.

kcgen: Yup, some of these `std::list`s have been the result of refactoring the older hand-rolled structs that held "pointer-to-next". and it was up to the calling code to manage them (usually in non-trivial ways):



Prior refactorings have migrated that to a std::list of MIDI handlers as `unique_ptr`s, where we now register and deregister the handlers, and have a get_handler:



On a given system there are only about four handlers, so the list is very small :)

But looking at this specifically, using a `std::map` would let us lookup the handler directly by `name`, giving a perfect excuse to eliminate another std::list!

The mapper would be a good place to replace the list's with vectors beacuse that code does walk the list looking for matches on every event.. making vector a perfect fit.

kcgen: Will set this to resolved - but fully agree that a list-to-vector search 'n replace PR would be 100% welcome!

weirddan455: >  Will set this to resolved - but fully agree that a list-to-vector search 'n replace PR would be 100% welcome!

Sounds good.  I think I'll grep for `std::list` later and have a look.  This code probably doesn't matter at all for performance since the list is so small but maybe there's some wins to be had elsewhere in the code.

src/misc/setup.cpp
kcgen: Done in a rename-separate commit (touched a lot of code)

src/cpu/cpu.cpp
kcgen: better than a single underscore! `CPU* _`

include/dos_system.h
kcgen: I did have `IsManaged()`, but the only use-case (two of them) involves checking  `if(!IsManaged())`, because we both delete *and* null the unmanaged devices, where as we only null the managed devices. 

And I thought `if(IsUnmanaged())` was easier on the eyes, given we have no `IsManaged()` use-cases.

But if we had both use cases - then absolutely, the positive scenario wins.  Will wait for your vote in either direction, now that you can see that there's no usecase for the positive. (I'm happy to switch if you still prefer).

kcgen:
@johnnovak , totally agree that the `_Init(nullptr)` approach feels like a hack to destruct the device. Let's figure out a nice approach, and I'll put it use across the board here.

Some aspect that make this tricky:

We have to do all the manipulation through the single function that holds the static managed object because that's where it lives; there's no way for other functions (like a separate `_Destroy()` call) to directly manipulate it or get access to the object.

(But we can definitely make a self-documenting Destroy() wrapper that calls the owning function with the "please-destroy-your-object" API of our choice :-)!

The other thing to keep in mind is that the destruction sequence:

```text
2023-08-10 19:14:43.943 | SDL: Exit requested
2023-08-10 19:14:43.968 | PCSPEAKER: Shutting down discrete model
2023-08-10 19:14:43.968 | DMA: Shutting down secondary controller
2023-08-10 19:14:43.968 | DMA: Shutting down SB16 on 16-bit DMA channel 5
2023-08-10 19:14:43.968 | OPL: Shutting down OPL3
2023-08-10 19:14:43.968 | SB16: Shutting down
2023-08-10 19:14:43.968 | DMA: Shutting down primary controller
2023-08-10 19:14:43.968 | MPU-401: Shutting down
2023-08-10 19:14:43.970 | IOBUS: Releasing 65 read and 56 write 8-bit port handlers
2023-08-10 19:14:43.970 | IOBUS: Releasing 34 read and 35 write 16-bit port handlers
2023-08-10 19:14:43.970 | IOBUS: Releasing 34 read and 35 write 32-bit port handlers
2023-08-10 19:14:43.970 | IOBUS: Handlers consumed 8864 total bytes
```

Are our static objects being "de-scoped" in reverse-creation order entirely managed by C++, without Setup() calling any `_Destroy()` calls.  The huge benefit is that both the nominal and "E-Exit()" rapid-death now have the exact same orderly destroy sequence.

The key point for us in writing this API is that having a bunch of "Destroy()" functions calls might seem confusing. New developers might think that the Destroy() calls are being run, and wonder why code inside them actually isn't executing in our shutdown sequence (well, it's because the static object is simply descoped instead of any Destroy() functions being run).

kcgen:
First cut for us to bash around:

```c++
enum ModuleAction {
  Create,
  // Reconfigure, (maybe a future addition?)
  Destroy,
}

MODULENAME_Perform(const ModuleAction action, const Section* section == nullptr) {
 ... 
}
```

So in the above comments, we could do things like:

`TANDYDAC_Perform(ModuleAction::Destroy)` instead of `TANDYDAC_Init(nullptr)`, if we wanted to destroy the DAC.

`_Perform()` and `ModuleAction` are a tad clunky though.. but it's at least a bit more self-documenting, and 'Perform()' is meant to be flexible in its action (instead of `_Init()`, which very much is the opposite of destroying :laughing: )

include/dos_system.h
johnnovak: > And I thought `if(IsUnmanaged())` was easier on the eyes, given we have no `IsManaged()` use-cases.

Yeah, I thought that's the reason why you did it like this 😄 

I personally prefer to always add the positive check only (`IsActive`, `IsEnabled`, etc.) then I don't need to think about it. Sooner or later we might need an `IsManaged()` check then we need to refactor... which is not the end of the world, of course; I guess I just like to use the "active check" for these types of functions across the board for consistency and "future proofing".

But don't wanna blow it out of proportion, it's a minor thing 😄 Up to you, really.

kcgen: Ah yeah - let's go the positive route 100% of the time; knowing that, it's even easier to read (and have even clearer expectations). Will do!

johnnovak: > knowing that, it's even easier to read (and have even clearer expectations)

Yeah, I think so. It's a bit weird to me to read stuff like `!IsActive() && IsUnmanaged()`. Contrived example in this case, but if we mix positive and negative cases, this can sooner or later happen.

include/dosbox.h
kcgen: Got it!

include/dos_system.h
kcgen:
```
DOS_AddDevice(com1, true);
DOS_AddDevice(ems, false);
DOS_AddDevice(nul, true);
```

Just lookup the API, bro :trollface:

johnnovak:
> First cut for us to bash around:
> 
> ```c++
> enum ModuleAction {
>   Create,
>   // Reconfigure, (maybe a future addition?)
>   Destroy,
> }
> 
> MODULENAME_Perform(const ModuleAction action, const Section* section == nullptr) {
>  ... 
> }
> ```
> 
> So in the above comments, we could do things like:
> 
> `TANDYDAC_Perform(ModuleAction::Destroy)` instead of `TANDYDAC_Init(nullptr)`, if we wanted to destroy the DAC during a runtime re-configure event.
> 
> `_Perform()` and `ModuleAction` are a tad clunky though.. but it's at least a bit more self-documenting, and 'Perform()' is meant to be flexible in its action (instead of `_Init()`, which very much is the opposite of destroying 😆 )
> 
> If you have tighter names - please suggest!

Thanks for the above explanation and I like this approach with the enum; that makes it very explicit, clean, and readable! 🎉 Let's see if we can up with slightly better names!

I think "lifecycle" is a pretty standard name for this type of thing, and "configure" is perhaps better than "perform"?

```cpp
enum ModuleLifecycleAction {
  Create,
  Reconfigure,
  Destroy
}

 MODULENAME_Configure(const ModuleLifecycleAction action, const Section* section == nullptr) {
```

I'm not 100% happy with these names though, and `ModuleLifecycleAction` is a bit too long... So I think we need a few more ideas.

kcgen:
+1 for Lifecycle!

> `enum ModuleLifecycleAction`

@FeralChild64 , @GranMinigun, @kklobe , @weirddan455 : any suggestions you can think of to tighten the above enum class name? 

Usage example:

`MIDI_Configure(ModuleLifecycleAction::Create, ...)`

johnnovak:
Yeah, I don't think it gets much better than `ModuleLifecycleAction`. E.g. `ModuleAction` is too generic, and `LifecycleAction` is the same. At the end of the day, we're not in the business of saving a few characters 😎 

It would be nice if `LifecycleAction::Reconfigure` would be sent when any of the section's settings have been changed _after_ init. That makes certain optimisations possible in the `MODULE_Configure` function, and I'm pretty sure that will come in handy when I get to fixing this: https://github.com/dosbox-staging/dosbox-staging/issues/2468#issuecomment-1596526490

kcgen:
@johnnovak , got the scafolding in place to handling the new `ModuleLifecyleAction` enum and the new `*_Configure()` functions instead of `Init()` and `Destroy()`.

Here's a preview of the first switch-over:


```diff
diff --git a/src/hardware/iohandler.cpp b/src/hardware/iohandler.cpp
index f47201593..0d8e39dec 100644
--- a/src/hardware/iohandler.cpp
+++ b/src/hardware/iohandler.cpp
@@ -402,13 +402,19 @@ public:
        }
 };
 
-static IO* test;
+void IO_Configure(const ModuleLifecycleAction lifecycle_action, Section* sec)
+{
+       static std::unique_ptr<IO> io_instance = {};
 
-void IO_Destroy(Section*) {
-       delete test;
-}
+       switch (lifecycle_action) {
+       case ModuleLifecycleAction::Create:
+               if (!io_instance) {
+                       io_instance = std::make_unique<IO>(sec);
+               }
+               break;
 
-void IO_Init(Section * sect) {
-       test = new IO(sect);
-       sect->AddDestroyFunction(&IO_Destroy);
+       // This module doesn't support runtime reconfiguration or destruction
+       case ModuleLifecycleAction::Reconfigure:
+       case ModuleLifecycleAction::Destroy: break;
+       }
 }
```

One of the design simplifications is that the module itself will be in-charge of what it supports at runtime, so that "bool changeable_at_runtime" can be localized here and not "spread out" into the conf system.

In this case, the IO module is (currently) can only be created once, after which it's torn down when the program closes and the static variable goes out of scope.

My one comment is the `ModuleLifecycleAction` is visually pretty heavy. I'd rather it just be `ModuleLifecycle`, which represent both the action and the state, which would lighten this up a bit.

Yeah, `Create` is an action, but it becomes a state in the past-tense (`Created`). 

Let me know what you think; will proceed once we're happy w/ it.

johnnovak:
> One the design simplifications is that the module itself will be in-charge of what it supports at runtime, so that "bool changeable_at_runtime" can be localized here and not "spread out" into the conf system.

Yeah, that's the sane way to do it 😄 We've already done this in a few modules, but it's not applied consistently. A logical continuation of this is that no other module should ever query config values they're not responsible for; they should ask the module who owns them, and even then no actual config values should leak out of the module, but the state/intent should be queried via an abstract interface (e.g. `RENDER_GetOutputMode` that returns an enum for OpenGL or Texture instead of returning the value of the `output` setting as a string directly. Gotta chop off those tentacles probing where they don't belong! 🐙 🪓 😎 )

> In this case, the IO module (currently) can only be created one time, after which it's torn down when the static variable goes out of scope on both program exit pathways (normal or E-Exit).

Nice 👍🏻 

> My one comment is the `ModuleLifecycleAction` is visually pretty heavy. It could be lightened up to just be `ModuleLifecycle`, which represent both the action and the state.

Sounds good!

> Yeah, `Create` is an action, but it becomes a state in the past-tense (`Created`).

I'd stay with present tense, personally. Past tense is slightly confusing when `Created` gets passed in to the configure function about to create the object.

> Let me know what you think; will proceed once we're happy w/ it!

The example looks very good @kcgen, so please proceed full steam ahead! 🚂 😄  This will go a long way towards normalising the wacky init, reconfigure, and shutdown sequences that in many cases only work by accident, I swear.

I have further ideas on improving the init sequence and managing intermodule dependencies (basically by turning all modules into proper classes as you touched on), but that doesn't conflict at all with this refactoring but is a continuation of it, plus it's more of a long-term goal. The work you're planning to do should definitely happen first and is an important stepping stone to achieve that endgoal. Will create a discussion ticket for it sometime.

GranMinigun:
Erm, I barely understand what's going on, but why not think about ditching sections altogether and e.g. do `MIDI::Create()` `GUS::Reconfigure()` `Video::Destroy()`?

johnnovak:
> Erm, I barely understand what's going on, but why not think about ditching sections altogether and e.g. do `MIDI::Create()` `GUS::Reconfigure()` `Video::Destroy()`?

Yes, next step would be just constructor, destructor, and `ModuleName::Reconfigure()` in my mind. But that would be a bigger job. Then everything would be instantiated from a single `App` or `Configuration` class with the dependencies injected, so zero global methods and structs. I'll write a discussion ticket about it some time plus I'll do some proof-of-concepts. But this is kinda the stepping stone to that full rework.

kcgen:
Just a status that I'm about half-way done on this; editing module to convert over to the lifecycle approach. 
```
[dosbox-staging] (kc/static-destruct-sequence-1 *+$|REBASE 20/48)
```

---

FeralChild64:
@kcgen After your rework, will it finally be possible to „reboot” the simulated machine?

kcgen:
> @kcgen After your rework, will it finally be possible to „reboot” the simulated machine?

It'll be slightly closer in that modules can be asked to `_Configure(ModuleLifeCycle::Reconfigure)` themselves, and this will take place for all modules in a given section.

For example, the `[dosbox]` section has `machine=` key, which multiple modules need be aware of: CPU, MEMORY, PAGING, BIOS, etc.

The problem is, even if those are rolled-over and reconfigured with a fresh machine type, then we need all downstream dependencies that are users of those services to be rolled-over as well. 

So it'll require another design change where modules can register themselves as dependencies of other modules, and then flow-through of lifecycle changes to all dependent child modules (recursively).

johnnovak:
Very nice progress, @kcgen 🎉 

About resetting, I believe my proposed explicit dependency injection approach will solve most of the recursive child subgraph restart problem. At the very minimum, that will _force_ us to "materialise" the object graph as there will be no ways to grow tentacles into other modules via global calls as all global module functions will be _gone_; the modules will all be object singletons (well, it's a dependency tree in reality, starting from the top App object; if it's a cyclic graph, it must be broken up into a tree by extracting common parts, etc).

But yeah, won't just happen by magic, and my proposal is a bigger work. I wanna finish up all my 0.81.0 stuff before writing the discussion ticket; we won't do that for the next release anyway.

johnnovak:
> Latest push is just to sync up intermediate work. About 25% of the way through this, so moving to draft until it's ready.

Seems like you're a magnet for monstruous PRs, no matter what you do -- just like me 😎 🤘🏻

johnnovak:
Just an idea @kcgen: when reinit happens, that's always due to changing a single config setting at runtime. Would it be possible to pass in a reference to that config property that contains the new value, and also a second "virtual" property that has the old value?

Some of the reinit stuff could be made more fine-grained and less "heavy handed" with such info being available. E.g. I'm working on the mixer now, and changing the reverb and chorus setting always triggers a full audio subsytem reinit, which is not ideal. Each module maintaining a copy of the whole config section state and do the diffing manually to figure out which single setting has been changed would be a rather clunky solution.

If you don't want to do this as part of this mega PR, I completely understand, though 😄 And then you might have done it already -- I haven't looked at the code yet 😄

johnnovak:
> Just an idea @kcgen: when reinit happens, that's always due to changing a single config setting at runtime. Would it be possible to pass in a reference to that config property that contains the new value, and also a second "virtual" property that has the old value?
> 
> Some of the reinit stuff could be made more fine-grained and less "heavy handed" with such info being available. E.g. I'm working on the mixer now, and changing the reverb and chorus setting always triggers a full audio subsytem reinit, which is not ideal. Each module maintaining a copy of the whole config section state and do the diffing manually to figure out which single setting has been changed would be a rather clunky solution.
> 
> If you don't want to do this as part of this mega PR, I completely understand, though 😄 And then you might have done it already -- I haven't looked at the code yet 😄

Not sure if you've seen this earlier comment of mine @kcgen? If it's out of scope for this, that's completely fine, of course (this is big enough already). I'm just curious what you think of the idea.

src/hardware/gus.cpp
kcgen: Yup! this will be replaced (or has been, maybe in the latest push) with the new `MODULE_Configure(ModuleLifecycle::Create / Reconfigure / Destroy)` approach.

kcgen:
> Not sure if you've seen this earlier comment

Thanks for bumping that up, @johnnovak! - it indeed slipped off my radar. Will try to write up some thoughts tomorrow.

kcgen:
> when reinit happens, that's always due to changing a single config setting at runtime. Would it be possible to pass in a reference to that config property that contains the new value, and also a second "virtual" property that has the old value?

Currently the Setup system is quite bloaty having to leap-frog from strings, over to a `Section_prop` object, where we get values from (like ``Get_bool("sbmixer")`), and even then it runs a search of all the possible key names, before returning the typed-value:

```c++
bool Section_prop::Get_bool(const std::string& _propname) const
{
    for (const_it tel = properties.begin(); tel != properties.end(); ++tel) {
        if ((*tel)->propname == _propname) {
            return ((*tel)->GetValue());
        }
    }
    return false;
}
```

There are lots of these functions:

```c++

int Get_int(const std::string& _propname) const
bool Get_bool(const std::string& _propname) const
double Get_double(const std::string& _propname) const
Prop_path* Get_path(const std::string& _propname) const
PropMultiVal* GetMultiVal(const std::string& _propname) const
PropMultiValRemain* GetMultiValRemain(const std::string& _propname) const
Property* Get_prop(int index)
const char* Get_string(const std::string& _propname) const
Prop_string* GetStringProp(const std::string& propname) const
Hex Get_hex(const std::string& _propname) const
std::string GetPropValue(const std::string&) const
```

It's certainly do-able carry a second "previous_properties" object and then write similar `GetPrevious<Type>`, and from there more code to lift out the current and changed property into another API.

But I don't like that Setup's output (the `Section_prop` object) is too bloated for Modules to use directly. For example, you would never have audio emulation code looping, sameple by sample, using the  `sec->Get_int("rate")`; instead we save the configured rate as `int sample_rate`.

So adding more bloaty, leap-froggy A -> to B -> to C etc.. (with string conversion, loops, and searches), feels like writing throw-away code.

I want the setup system's API trafficking in "direct-drive" scalar structs:
For example, the setup system would pass back a `MixerSettings settings` struct populated with strongly classed enum choices, integers, and bool that the modules can use straight-away in tight loops.


> Some of the reinit stuff could be made more fine-grained and less "heavy handed" with such info being available. 

I know it feels heavy handed (and lazy) to recreate a device, but this is very fast and the creation-based-on-conf follows a single code path.

The benefits with that single creation code path making testing and maintenance eaasier.

Imagine if every module had a separate code path to change each of its configurable settings: the testing would be daunting, there would be a lot of more state-management code.

If the code gets simpler and easier to maintain, then that's sounds good - like the mixer that you mentioned.

> Each module maintaining a copy of the whole config section state and do the diffing manually to figure out which single setting has been changed would be a rather clunky solution.

This is still tricky to think about even if we moved to returning an efficient `MixerSettings` struct of scalars and enum values.

Setup would still likely have to return a char*/std::string of the key-name in text form (or maybe an enum of the keys), and then the module would need something like a `std::map<key-type, std::function>` or `switch` that keys off the changed item calling separate reconfiguration blocks of code.

> If you don't want to do this as part of this mega PR, I completely understand, though 😄

Yup; I think overhauling the setup system's internals (its types, and how it conveys change) are best done in a separate PR - and then rippled into all the modules!

johnnovak:
Thanks for the thoughtful and detailed reply @kcgen; admittedly, I haven't thought about the details and flow-on effects too much, plus all the potential complications. I pretty much agree with all what you've written. Some specific comments:

> I want the setup system's API trafficking in "direct-drive" scalar structs: For example, the setup system would pass back a `MixerSettings settings` struct populated with strongly classed enum choices, integers, and bool that the modules can use straight-away in tight loops.

Yep, that's the best approach.

> > Some of the reinit stuff could be made more fine-grained and less "heavy handed" with such info being available.
> 
> I know it feels heavy handed (and lazy) to recreate a device, but this is very fast and the creation-based-on-conf follows a single code path.
> 
> The benefits with a single creation code path means one block of code to maintain and one block of code to test.
> 
> Imagine if every module had a separate code path to change each of its configurable settings: the testing would be daunting, there would be a lot of more state-management code.
> 
> If the code gets simpler and easier to maintain, then that's sounds good - like the mixer that you mentioned.

Yeah, the mixer is probably the most complex use case: channel settings can be altered in a variety of ways (the `MIXER` command, the `reverb`, `chorus`, `crossfeed` config params, etc.) Maybe me writing "heavy handed" led the discussion a bit astray; what I wanted to really say is that fine-grained control would actually make things easier to reason about and with the current approach some thiings are impossible to solve.

It gets complex and I won't write a book now 😄, I've encountered a lot of these during my almost full mixer rework. But when the mixer will be exposed on the OSD too, that will be _another_ layer of complication. What I'm saying, it's actually best to wait with more substantial state management / re-init refactors until the OSD work, and let the OSD work "guide" that rework, unless we might build something that's not really fit for GUI usage, only for the config file & command line approach.
 
> Yup; I think overhauling the setup system's internals (its types, and how it conveys change) are best done in a separate PR - and then rippled into all the modules!

Agreed, and like I said, I highly recommend to postpone it to when the OSD mixer stuff will be implemented. That will be one of the most complex (or _the_ most complex) UI components, so it makes sense to start with that one. If the requirements of the OSD mixer are fullfilled config handling wise, the rest of the OSD config stuff will be an order of magnitude easier... except for the mapper, of course, but that's still quite a bit simpler than the mixer UI.

So the TL;DR is, I suggest postponing all substantial config handling rework until the OSD mixer & mapper work, then let that OSD work "drive" the config handling redesign.

johnnovak:
@kcgen If it's not clear, what I was trying to say above is to postpone substantial _further_ rework after this PR to the time we start the OSD work. The scope of this PR is of course completely fine for now—I was only referring to what comes _after_ this 😄

kcgen:
> @kcgen If it's not clear, what I was trying to say above is to postpone substantial _further_ rework after this PR to the time we start the OSD work. The scope of this PR is of course completely fine for now—I was only referring to what comes _after_ this 😄

:laughing: Indeed :-)  this PR only works to package up the modules in a bit tighter form that can survive earlier destruction (in reverse-setup order) instead of at the less defined more sprawling file-level scale. 

It also doesn't touch the Setup class itself (or how it's values are passed or parsed);  also doesn't touch the reconfiguration lifecycles - all of that will be good stuff to fill in (and really, best guided by OSD-need).

The CPU module does have a reconfiguration ability (already there), so that's put to use.

Just gotta wrap this thing up!

kcgen:
Rebase conflicts, combined with this PR's large scope, have become too much for me to deal with, so I'm canning this pursuit.

This PR de-global'd almost all the modules by eliminating their static and dynamically constructed objects that sit at the modules' file-scope, which have indeterminate startup and shutdown ordering.

These fixes are valuable regardless of the setup and initialization approach. 

The lesson learned here is to open tickets for incremental features/fixes/improvements, on a per-module basis.  That way, even if just a single module is improved, then that's something we want - even if the others aren't.

PR #2750, opened 2023-08-11
Fix several static analysis warnings

FeralChild64:
- fixed few CppCheck warnings
- fixed 28 PVS Studio warnings

These are low-risk changes, I have tried to avoid files where active development is currently taking place.

### DIFF
--- a/src/dos/drive_cache.cpp
+++ b/src/dos/drive_cache.cpp
@@ -31,21 +31,29 @@
 
 int fileInfoCounter = 0;
 
-bool SortByName(DOS_Drive_Cache::CFileInfo* const &a, DOS_Drive_Cache::CFileInfo* const &b) {
-	return strcmp(a->shortname,b->shortname)<0;
+bool SortByName(DOS_Drive_Cache::CFileInfo* const a,
+                DOS_Drive_Cache::CFileInfo* const b)
+{
+	return strcmp(a->shortname, b->shortname) < 0;
 }
 
-bool SortByNameRev(DOS_Drive_Cache::CFileInfo* const &a, DOS_Drive_Cache::CFileInfo* const &b) {
-	return strcmp(a->shortname,b->shortname)>0;
+bool SortByNameRev(DOS_Drive_Cache::CFileInfo* const a,
+                   DOS_Drive_Cache::CFileInfo* const b)
+{
+	return strcmp(a->shortname, b->shortname) > 0;
 }
 
-bool SortByDirName(DOS_Drive_Cache::CFileInfo* const &a, DOS_Drive_Cache::CFileInfo* const &b) {
+bool SortByDirName(DOS_Drive_Cache::CFileInfo* const a,
+                   DOS_Drive_Cache::CFileInfo* const b)
+{
 	// Directories first...
 	if (a->isDir!=b->isDir) return (a->isDir>b->isDir);	
 	return strcmp(a->shortname,b->shortname)<0;
 }
 
-bool SortByDirNameRev(DOS_Drive_Cache::CFileInfo* const &a, DOS_Drive_Cache::CFileInfo* const &b) {
+bool SortByDirNameRev(DOS_Drive_Cache::CFileInfo* const a,
+                      DOS_Drive_Cache::CFileInfo* const b)
+{
 	// Directories first...
 	if (a->isDir!=b->isDir) return (a->isDir>b->isDir);	
 	return strcmp(a->shortname,b->shortname)>0;
@@ -394,7 +402,7 @@ bool DOS_Drive_Cache::GetShortName(const char* fullname, char* shortname) {
 }
 
 int DOS_Drive_Cache::CompareShortname(const char* compareName, const char* shortName) {
-	char const* cpos = strchr(shortName,'~');
+	const char* cpos = strchr(shortName, '~');
 	if (cpos) {
 /* the following code is replaced as it's not safe when char* is 64 bits */
 /*		Bits compareCount1	= (int)cpos - (int)shortName;
--- a/src/misc/unicode.cpp
+++ b/src/misc/unicode.cpp
@@ -1889,8 +1889,8 @@ static bool construct_mapping(const uint16_t code_page)
 
 	auto& mappings = per_code_page_mappings[code_page];
 
-	mappings.dos_to_grapheme_normalized = new_mapping;
-	mappings.grapheme_to_dos = new_mapping_reverse;
+	mappings.dos_to_grapheme_normalized = std::move(new_mapping);
+	mappings.grapheme_to_dos = std::move(new_mapping_reverse);
 
 	// Construct decomposed mapping
 	construct_decomposed(mappings.dos_to_grapheme_normalized,
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 265
+          MAX_BUGS: 237
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -412,7 +412,7 @@ enum SeekType : uint8_t {
 
 typedef bool (MultiplexHandler)(void);
 void DOS_AddMultiplexHandler(MultiplexHandler* handler);
-void DOS_DeleteMultiplexHandler(MultiplexHandler* handler);
+void DOS_DeleteMultiplexHandler(MultiplexHandler* const handler);
 
 /* AddDevice stores the pointer to a created device */
 void DOS_AddDevice(DOS_Device * adddev);
--- a/src/cpu/core_dynrec/decoder.h
+++ b/src/cpu/core_dynrec/decoder.h
@@ -414,7 +414,7 @@ restart_prefix:
 			dyn_mov_byte_imm(opcode&3,(opcode>>2)&1,decode_fetchb());
 			break;
 		case 0xb8:case 0xb9:case 0xba:case 0xbb:case 0xbc:case 0xbd:case 0xbe:case 0xbf:	
-			dyn_mov_word_imm(opcode&7);break;
+			dyn_mov_word_imm(opcode&7);
 			break;
 
 		// 'shiftop []/reg8,imm8/1/cl'
--- a/src/debug/debug_gui.cpp
+++ b/src/debug/debug_gui.cpp
@@ -39,7 +39,7 @@
 #endif
 
 struct _LogGroup {
-	char const *front = nullptr;
+	const char* front = nullptr;
 	bool enabled = false;
 };
 #include <list>
@@ -55,9 +55,8 @@ static FILE *debuglog = nullptr;
 
 extern int old_cursor_state;
 
-
-
-void DEBUG_ShowMsg(char const* format,...) {
+void DEBUG_ShowMsg(const char* format, ...)
+{
 	// Quit early if the window hasn't been created yet
 	if (!dbg.win_out)
 		return;
@@ -99,9 +98,9 @@ void DEBUG_RefreshPage(int scroll) {
 		return;
 
 	if (scroll == -1 && logBuffPos != logBuff.begin())
-		logBuffPos--;
+		--logBuffPos;
 	else if (scroll == 1 && logBuffPos != logBuff.end())
-		logBuffPos++;
+		++logBuffPos;
 
 	list<string>::iterator i = logBuffPos;
 	int maxy, maxx; getmaxyx(dbg.win_out,maxy,maxx);
@@ -121,7 +120,8 @@ void DEBUG_RefreshPage(int scroll) {
 	wrefresh(dbg.win_out);
 }
 
-void LOG::operator() (char const* format, ...){
+void LOG::operator()(const char* format, ...)
+{
 	char buf[512];
 	va_list msg;
 	va_start(msg,format);
@@ -132,7 +132,6 @@ void LOG::operator() (char const* format, ...){
 	if ((d_severity!=LOG_ERROR) && (!loggrp[d_type].enabled)) return;
 	DEBUG_ShowMsg("%10u: %s:%s\n",static_cast<uint32_t>(cycle_count),loggrp[d_type].front,buf);
 }
-
 
 static void Draw_RegisterLayout(void) {
 	// Quit early if the window hasn't been created yet
--- a/src/dos/dos_misc.cpp
+++ b/src/dos/dos_misc.cpp
@@ -31,11 +31,11 @@ static callback_number_t call_int2a = 0;
 static std::list<MultiplexHandler*> Multiplex;
 typedef std::list<MultiplexHandler*>::iterator Multiplex_it;
 
-void DOS_AddMultiplexHandler(MultiplexHandler * handler) {
+void DOS_AddMultiplexHandler(MultiplexHandler* handler) {
 	Multiplex.push_front(handler);
 }
 
-void DOS_DeleteMultiplexHandler(MultiplexHandler* handler)
+void DOS_DeleteMultiplexHandler(MultiplexHandler* const handler)
 {
 	for (Multiplex_it it = Multiplex.begin(); it != Multiplex.end(); ++it) {
 		if(*it == handler) {
--- a/src/dos/dos_misc.cpp
+++ b/src/dos/dos_misc.cpp
@@ -37,7 +37,7 @@ void DOS_AddMultiplexHandler(MultiplexHandler * handler) {
 
 void DOS_DeleteMultiplexHandler(MultiplexHandler* handler)
 {
-	for(Multiplex_it it =Multiplex.begin();it != Multiplex.end();it++) {
+	for (Multiplex_it it = Multiplex.begin(); it != Multiplex.end(); ++it) {
 		if(*it == handler) {
 			Multiplex.erase(it);
 			return;
@@ -46,9 +46,12 @@ void DOS_DeleteMultiplexHandler(MultiplexHandler* handler)
 }
 
 static Bitu INT2F_Handler(void) {
-	for(Multiplex_it it = Multiplex.begin();it != Multiplex.end();it++)
-		if( (*it)() ) return CBRET_NONE;
-   
+	for (Multiplex_it it = Multiplex.begin(); it != Multiplex.end(); ++it) {
+		if ((*it)()) {
+			return CBRET_NONE;
+		}
+	}
+
 	LOG(LOG_DOSMISC,LOG_ERROR)("DOS:Multiplex Unhandled call %4X",reg_ax);
 	return CBRET_NONE;
 }
--- a/src/hardware/ipx.cpp
+++ b/src/hardware/ipx.cpp
@@ -131,7 +131,7 @@ ECBClass::ECBClass(uint16_t segment, uint16_t offset)
 }
 
 void ECBClass::writeDataBuffer(uint8_t* buffer, uint16_t length) {
-	if(databuffer!=nullptr) delete [] databuffer;
+	delete[] databuffer;
 	databuffer = new uint8_t[length];
 	memcpy(databuffer,buffer,length);
 	buflen=length;
@@ -262,7 +262,7 @@ ECBClass::~ECBClass() {
 			if(nextECB != nullptr) nextECB->prevECB = prevECB;
 		}
 	}
-	if(databuffer!=nullptr) delete [] databuffer;
+	delete[] databuffer;
 }
 
 
--- a/src/hardware/ne2000.cpp
+++ b/src/hardware/ne2000.cpp
@@ -1526,11 +1526,11 @@ public:
 	}
 };
 
-static NE2K* test;
-void NE2K_ShutDown(Section* sec) {
-    (void)sec;//UNUSED
-	if(test) delete test;
-	test=nullptr;
+static NE2K* instance;
+void NE2K_ShutDown(Section* /* sec */)
+{
+	delete instance;
+	instance = nullptr;
 }
 
 void NE2K_Init(Section* sec)
@@ -1538,14 +1538,14 @@ void NE2K_Init(Section* sec)
 	assert(sec);
 	// LOG(LOG_MISC,LOG_DEBUG)("Initializing NE2000 network card emulation");
 
-	test = new NE2K(sec);
+	instance = new NE2K(sec);
 
 	constexpr auto changeable_at_runtime = true;
 	sec->AddDestroyFunction(&NE2K_ShutDown, changeable_at_runtime);
 
-	if (!test->load_success) {
-		delete test;
-		test = nullptr;
+	if (!instance->load_success) {
+		delete instance;
+		instance = nullptr;
 	}
 }
 
--- a/src/hardware/serialport/nullmodem.cpp
+++ b/src/hardware/serialport/nullmodem.cpp
@@ -154,8 +154,8 @@ CNullModem::CNullModem(const uint8_t port_idx, CommandLine *cmd)
 }
 
 CNullModem::~CNullModem() {
-	if (serversocket) delete serversocket;
-	if (clientsocket) delete clientsocket;
+	delete serversocket;
+	delete clientsocket;
 	// remove events
 	for (uint16_t i = SERIAL_BASE_EVENT_COUNT + 1;
 	     i <= SERIAL_NULLMODEM_EVENT_COUNT; i++) {
--- a/src/misc/fs_utils.cpp
+++ b/src/misc/fs_utils.cpp
@@ -52,7 +52,7 @@ bool is_hidden_by_host(const std_fs::path& pathname)
 
 	// Filenames that don't start with dot or are the two directory entries
 	// are not hidden by the host
-	if (filename.find(".") != 0 || filename == "." || filename == "..") {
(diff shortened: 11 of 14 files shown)


weirddan455 (approved):
This all looks good to me.  Nice job getting the warnings lowered.

FeralChild64:
Fixed 8 more PVS Studio warnings by simply removing unnecessary `const_cast`'s.

src/misc/unicode.cpp
```diff
@@ -1889,8 +1889,8 @@ static bool construct_mapping(const uint16_t code_page)
 
 	auto& mappings = per_code_page_mappings[code_page];
 
-	mappings.dos_to_grapheme_normalized = new_mapping;
-	mappings.grapheme_to_dos = new_mapping_reverse;
+	mappings.dos_to_grapheme_normalized = std::move(new_mapping);
+	mappings.grapheme_to_dos = std::move(new_mapping_reverse);
```
johnnovak, line 1893: I'm just questioning whether we should go all-in on `std::move` as it's complicated as hell and the benefits are marginal at best. Where performance really matters you can still pass around references to containers, job done. I've encountered a few weird cases where `std::move` resulted in memory usages after free and couldn't figure it out for the life of me what and how went wrong. Straight copy just worked and it was just non-performance intensive code that gets called once in a blue moon, so who cares.

I'm just very wary of complex features like this; prefer to use a sweet-spot subset of C++ where you get most of the benefits without your brain exploding 😅

src/dos/drive_cache.cpp
```diff
@@ -31,21 +31,29 @@
 
 int fileInfoCounter = 0;
 
-bool SortByName(DOS_Drive_Cache::CFileInfo* const &a, DOS_Drive_Cache::CFileInfo* const &b) {
-	return strcmp(a->shortname,b->shortname)<0;
+bool SortByName(DOS_Drive_Cache::CFileInfo* const a,
```
johnnovak, line 34: Pointers by reference 🥶

FeralChild64: Exactly...

src/misc/unicode.cpp
johnnovak: So in this case, this gets executed rarely, thus falls squarely in the "who cares" category. We're also not dealing with gigabytes of data, it does not matter if a few or even a few hundred kilobytes worth of data exists twice momentarily.

FeralChild64: There are very few few warnings like this - I've fixed 2, but I see there are 2 more, so in my opinion adding `std::move` won't complicate the code in any significant way.

@kcgen What's your opinion? Maybe we should just suppress all _v820_ warnings?

johnnovak (approved):
Nice job @FeralChild64, you've nicely broken the changes up into commits.

I'm just not sold on `std::move`; looks benign in this case, but I don't want to see it everyhere in the codebase. Too much extra mental overhead for little payoff. We have enough of that with pointers, references, smart pointers, iterators, etc. and can do everything we want with them already.

src/misc/unicode.cpp
kcgen: Yup, if we're doing things right, we shouldn't need lots of moves so I think careful use makes sense. 

For examples, v820 https://pvs-studio.com/en/docs/warnings/v820/ warns about an object being copied and the original still being "alive" but not used anymore. 

This indicates that maybe we can even collapse down to one object.


Temp_string = "stuff"
Temp_string += " more stuff"

Final_string = Temp_string

Return Final_string


PVS would warn about Temp_string being an unused duplicate.. so we could use std::move or even better, maybe we could just use one string. 

Another good thing with this duplicate-content warnings is it reveals a "split brain" scenario where we now have two objects for some timeframe. 

I think the benefit is not really about speed or space usage (which are inconsequential), but more about helping improve the code quality and make it more direct or more readable. 

Another nice thing with move is it can make the intention clearer and reveal bugs, if the moved-out object does happen to be used after being moved from, then it reveals a bug. 

Where as if we don't use move (or ignore the v820 bugs), then we can have multiple 'split brain' objects both operating with the same content, and hide bugs.

johnnovak: Yeah, pretty benign in this case, so if we only have a couple of `std::move`'s in our codebase, I honestly don't care.

My problem is I _thought_ I understood how `std::move` works, but then I hit a weird bug that I could not figure out in 30+ mins. So either I'm not seeing the obvious, or `std::move` is really counterintuitive in some cases, or maybe it's a compiler bug (unlikely, though). See here:

https://github.com/dosbox-staging/dosbox-staging/pull/2748/commits/57c5d99eb216c1592104d94c03491936c9713f7a

This is WIP, but it will be part of the final PR as with `std::move` I'm getting 100% reproducible crashes... and I don't understand why!

I tried to read up on `std::move`, then I found these 15-20 page long blogs that go into a lot of detail and there are lots of subtleties about its usage... So that's when I though, screw this, C++ is hard and complicated enough without `std::move`.

Having said that, if you guys can point me to some _simple_ explanation on its usage, I will thank you for it and will read it. Just so far my impression about it isn't great as I could not figure it out why it's misbehaving in a relatively simple case...

weirddan455: @johnnovak That's weird.  That usage of `std::move` looks safe to me.

As I understand it, `std::move` is just a cast to an R value reference and nothing more.  It's up to the class you're moving to provide a move constructor like `basic_string& operator=(basic_string&& __str)` for string (note the `&&` to mean R value reference).

weirddan455: @johnnovak  Oh, actually I found the bug in that commit.  Was looking at main and didn't see it but in your commit you're using `loaded_shader_name` after you moved it.

```cpp
render.shader.name   = std::move(loaded_shader_name);
render.shader.source = std::move(source);

render.shader.settings = parse_shader_settings(loaded_shader_name,
												source);
```

kcgen: yup, `std::move` is awesome for clarifying the intent:

_"I'm transplanting this object from A into B, and I don't plan for A to ever be used again, even though the variable might still exist "in-scope"._

It communicates a whole lot to readers and becomes "hard tested" at runtime, especially with UBSAN that will catch the any follow on usage of the zombie "hollowed-out" objects.

Looking at your example, you nailed the intent perfectly (and as a reader, it says a whole lot to me):



The only gotcha was that the zombies were used further down:



Replacing that with the left-hand-size objects should do the trick:
```c++
render.shader.settings = parse_shader_settings(render.shader.name,
                                               render.shader.source);
```

kcgen: (looks like our two posts crossed in the ether, @weirddan455 !)

johnnovak: Thanks for the explanation guys, and that made perfect sense, so I just tried it:

```cpp
// Move the temporary name and source into the members
render.shader.name   = std::move(loaded_shader_name);
render.shader.source = std::move(source);

render.shader.settings = parse_shader_settings(
		render.shader.name, render.shader.source);

// Pass the shader source up to the GFX engine
GFX_SetShader(render.shader.source);
```

Based on your explanation this should now be perfectly safe, right? Except it isn't 🥶 

Start with `dosbox -noprimaryconf -nolocalconf`, then issue `glshader = asdf` twice in the console. UBSAN crash, every single time 🤷🏻 So I thought I got it... but I still don't get it, and it's still a mystery to me 😐 

Any further suggestions are welcome. I'm thinking maybe it matters what happens to these moved values downstream... But, logically (and correct me if my thinking is wrong), it _should not matter one bit_ what happens downstream, right? Except it does. So... still not getting what the problem is here.

```
==53612==ERROR: AddressSanitizer: heap-use-after-free on address 0x00010aa3a380 at pc 0x00010808270c bp 0x00016f8ebc20 sp 0x00016f8eb3b8
READ of size 1450 at 0x00010aa3a380 thread T0
    #0 0x108082708 in MemcmpInterceptorCommon(void*, int (*)(void const*, void const*, unsigned long), void const*, void const*, unsigned long)+0x118 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x16708) (BuildId: d1f24f37c92c38e8a3d1320c68b56ea932000000200000000100000000000b00)
    #1 0x108082b2c in wrap_memcmp+0x80 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x16b2c) (BuildId: d1f24f37c92c38e8a3d1320c68b56ea932000000200000000100000000000b00)
    #2 0x100536c5c in std::__1::char_traits<char>::compare(char const*, char const*, unsigned long) char_traits.h:262
    #3 0x1005369b0 in std::__1::basic_string_view<char, std::__1::char_traits<char> >::compare(std::__1::basic_string_view<char, std::__1::char_traits<char> >) const string_view:462
    #4 0x101990138 in bool std::__1::operator!=[abi:v15007]<char, std::__1::char_traits<char>, 1>(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::common_type<std::__1::basic_string_view<char, std::__1::char_traits<char> > >::type) string_view:800
    #5 0x10198faa4 in GFX_SetShader(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) sdlmain.cpp:2300
    #6 0x10177bad4 in load_shader(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) render.cpp:813
    #7 0x10177d210 in RENDER_Init(Section*) render.cpp:943
    #8 0x10078efec in Section::ExecuteInit(bool) setup.cpp:1228
    #9 0x100600ed0 in CONFIG::Run() programs.cpp:814
    #10 0x100605488 in PROGRAMS_Handler() programs.cpp:149
    #11 0x100513510 in Normal_Loop() dosbox.cpp:148
    #12 0x100513628 in DOSBOX_RunMachine() dosbox.cpp:320
    #13 0x100a306b4 in CALLBACK_RunRealInt(unsigned char) callback.cpp:114
    #14 0x102c4abec in run_binary_executable(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >) shell_misc.cpp:607
    #15 0x102c46dfc in DOS_Shell::Execute(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >) shell_misc.cpp:480
    #16 0x102bc85b8 in DOS_Shell::DoCommand(char*) shell_cmds.cpp:185
    #17 0x102bc7794 in DOS_Shell::CheckConfig(char*, char*) shell_cmds.cpp:146
    #18 0x102bc8658 in DOS_Shell::DoCommand(char*) shell_cmds.cpp:186
    #19 0x102b5d6ac in DOS_Shell::ParseLine(char*) shell.cpp:326
    #20 0x102b61290 in DOS_Shell::Run() shell.cpp:436
    #21 0x102b6a200 in SHELL_Init() shell.cpp:1431
    #22 0x10079a2a4 in Config::StartUp() setup.cpp:1525
    #23 0x1019b6b78 in sdl_main(int, char**) sdlmain.cpp:5017
    #24 0x100510f18 in main main.cpp:30
    #25 0x107a69088 in start+0x204 (dyld:arm64+0x5088) (BuildId: 75627683a78032adae34cf86dd23a26b32000000200000000100000000050c00)
    #26 0x1e7b7ffffffffffc  (<unknown module>)

0x00010aa3a380 is located 0 bytes inside of 1456-byte region [0x00010aa3a380,0x00010aa3a930)
freed by thread T0 here:
    #0 0x1080b759c in wrap__ZdlPv+0x68 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x4b59c) (BuildId: d1f24f37c92c38e8a3d1320c68b56ea932000000200000000100000000000b00)
    #1 0x10053977c in void std::__1::__libcpp_operator_delete[abi:v15007]<void*>(void*) new:256
    #2 0x10053972c in void std::__1::__do_deallocate_handle_size[abi:v15007]<>(void*, unsigned long) new:280
    #3 0x1005396b0 in std::__1::__libcpp_deallocate[abi:v15007](void*, unsigned long, unsigned long) new:296
    #4 0x1019f2f34 in std::__1::allocator<char>::deallocate[abi:v15007](char*, unsigned long) allocator.h:128
    #5 0x1019f2e14 in std::__1::allocator_traits<std::__1::allocator<char> >::deallocate[abi:v15007](std::__1::allocator<char>&, char*, unsigned long) allocator_traits.h:282
    #6 0x1019f2b34 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__move_assign[abi:v15007](std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&, std::__1::integral_constant<bool, true>) string:2537
    #7 0x1019771b0 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::operator=[abi:v15007](std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&&) string:2562
    #8 0x10177b8e4 in load_shader(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) render.cpp:807
    #9 0x10177d210 in RENDER_Init(Section*) render.cpp:943
    #10 0x10078efec in Section::ExecuteInit(bool) setup.cpp:1228
    #11 0x100600ed0 in CONFIG::Run() programs.cpp:814
    #12 0x100605488 in PROGRAMS_Handler() programs.cpp:149
    #13 0x100513510 in Normal_Loop() dosbox.cpp:148
    #14 0x100513628 in DOSBOX_RunMachine() dosbox.cpp:320
    #15 0x100a306b4 in CALLBACK_RunRealInt(unsigned char) callback.cpp:114
    #16 0x102c4abec in run_binary_executable(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >) shell_misc.cpp:607
    #17 0x102c46dfc in DOS_Shell::Execute(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >) shell_misc.cpp:480
    #18 0x102bc85b8 in DOS_Shell::DoCommand(char*) shell_cmds.cpp:185
    #19 0x102bc7794 in DOS_Shell::CheckConfig(char*, char*) shell_cmds.cpp:146
    #20 0x102bc8658 in DOS_Shell::DoCommand(char*) shell_cmds.cpp:186
    #21 0x102b5d6ac in DOS_Shell::ParseLine(char*) shell.cpp:326
    #22 0x102b61290 in DOS_Shell::Run() shell.cpp:436
    #23 0x102b6a200 in SHELL_Init() shell.cpp:1431
    #24 0x10079a2a4 in Config::StartUp() setup.cpp:1525
    #25 0x1019b6b78 in sdl_main(int, char**) sdlmain.cpp:5017
    #26 0x100510f18 in main main.cpp:30
    #27 0x107a69088 in start+0x204 (dyld:arm64+0x5088) (BuildId: 75627683a78032adae34cf86dd23a26b32000000200000000100000000050c00)
    #28 0x1e7b7ffffffffffc  (<unknown module>)

previously allocated by thread T0 here:
    #0 0x1080b718c in wrap__Znwm+0x68 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x4b18c) (BuildId: d1f24f37c92c38e8a3d1320c68b56ea932000000200000000100000000000b00)
    #1 0x19bd47c54 in std::__1::basic_stringbuf<char, std::__1::char_traits<char>, std::__1::allocator<char> >::str() const+0x9c (libc++.1.dylib:arm64+0x29c54) (BuildId: ca366de6c395319ab3abe22126d2839432000000200000000100000000050c00)
    #2 0x485d800101a2b980  (<unknown module>)
    #3 0x101785e6c in read_shader_source(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&)::$_0::operator()(std::__1::__fs::filesystem::path const&) const render.cpp:606
    #4 0x101783a7c in read_shader_source(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) render.cpp:622
    #5 0x10177b4b4 in load_shader(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) render.cpp:790
    #6 0x10177d210 in RENDER_Init(Section*) render.cpp:943
    #7 0x10078efec in Section::ExecuteInit(bool) setup.cpp:1228
    #8 0x100600ed0 in CONFIG::Run() programs.cpp:814
    #9 0x100605488 in PROGRAMS_Handler() programs.cpp:149
    #10 0x100513510 in Normal_Loop() dosbox.cpp:148
    #11 0x100513628 in DOSBOX_RunMachine() dosbox.cpp:320
    #12 0x100a306b4 in CALLBACK_RunRealInt(unsigned char) callback.cpp:114
    #13 0x102c4abec in run_binary_executable(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >) shell_misc.cpp:607
    #14 0x102c46dfc in DOS_Shell::Execute(std::__1::basic_string_view<char, std::__1::char_traits<char> >, std::__1::basic_string_view<char, std::__1::char_traits<char> >) shell_misc.cpp:480
    #15 0x102bc85b8 in DOS_Shell::DoCommand(char*) shell_cmds.cpp:185
    #16 0x102bc7794 in DOS_Shell::CheckConfig(char*, char*) shell_cmds.cpp:146
    #17 0x102bc8658 in DOS_Shell::DoCommand(char*) shell_cmds.cpp:186
    #18 0x102b5d6ac in DOS_Shell::ParseLine(char*) shell.cpp:326
    #19 0x102b61290 in DOS_Shell::Run() shell.cpp:436
    #20 0x102b6a200 in SHELL_Init() shell.cpp:1431
    #21 0x10079a2a4 in Config::StartUp() setup.cpp:1525
    #22 0x1019b6b78 in sdl_main(int, char**) sdlmain.cpp:5017
    #23 0x100510f18 in main main.cpp:30
    #24 0x107a69088 in start+0x204 (dyld:arm64+0x5088) (BuildId: 75627683a78032adae34cf86dd23a26b32000000200000000100000000050c00)
    #25 0x1e7b7ffffffffffc  (<unknown module>)

SUMMARY: AddressSanitizer: heap-use-after-free (libclang_rt.asan_osx_dynamic.dylib:arm64+0x16708) (BuildId: d1f24f37c92c38e8a3d1320c68b56ea932000000200000000100000000000b00) in MemcmpInterceptorCommon(void*, int (*)(void const*, void const*, unsigned long), void const*, void const*, unsigned long)+0x118
Shadow bytes around the buggy address:
  0x007021567420: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x007021567430: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x007021567440: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x007021567450: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x007021567460: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x007021567470:[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x007021567480: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x007021567490: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0070215674a0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0070215674b0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0070215674c0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==53612==ABORTING

Stack trace:
32         0x107a6908c 36  dyld                                0x0000000107a6908c start + 520
31         0x100510f1c main + 36
30         0x1019b6b7c sdl_main(int, char**) + 18024
29         0x10079a2a8 Config::StartUp() + 168
28         0x102b6a204 SHELL_Init() + 7152
27         0x102b61294 DOS_Shell::Run() + 8808
26         0x102b5d6b0 DOS_Shell::ParseLine(char*) + 6488
25         0x102bc865c DOS_Shell::DoCommand(char*) + 3508
24         0x102bc7798 DOS_Shell::CheckConfig(char*, char*) + 2260
23         0x102bc85bc DOS_Shell::DoCommand(char*) + 3348
22         0x102c46e00 DOS_Shell::Execute(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 4680
21         0x102c4abf0 run_binary_executable(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 7376
20         0x100a306b8 CALLBACK_RunRealInt(unsigned char) + 412
19         0x10051362c DOSBOX_RunMachine() + 28
18         0x100513514 Normal_Loop() + 596
17         0x10060548c PROGRAMS_Handler() + 1344
16         0x100600ed4 CONFIG::Run() + 41944
15         0x10078eff0 Section::ExecuteInit(bool) + 2064
14         0x10177d214 RENDER_Init(Section*) + 3052
13         0x10177bad8 load_shader(std::string const&) + 3440
12         0x10198faa8 GFX_SetShader(std::string const&) + 676
11         0x10199013c bool std::operator!=[abi:v15007]<char, std::char_traits<char>, 1>(std::basic_string_view<char, std::char_traits<char> >, std::common_type<std::basic_string_view<char, std::char_traits<char> > >::type) + 616
10         0x1005369b4 std::basic_string_view<char, std::char_traits<char> >::compare(std::basic_string_view<char, std::char_traits<char> >) const + 848
9          0x100536c60 std::char_traits<char>::compare(char const*, char const*, unsigned long) + 68
8          0x108082b30 wrap_memcmp + 132
7          0x10808272c MemcmpInterceptorCommon(void*, int (*)(void const*, void const*, unsigned long), void const*, void const*, unsigned long) + 316
6          0x1080b04ec __asan::ReportGenericError(unsigned long, unsigned long, unsigned long, unsigned long, bool, unsigned long, unsigned int, bool) + 1456
5          0x1080b11a8 __asan::ErrorDescription::Print() + 0
4          0x1080cadc0 __sanitizer::SetCheckUnwindCallback(void (*)()) + 0
3          0x1080cb654 __sanitizer::Atexit(void (*)()) + 0
2          0x19bd16340 abort + 168
1          0x19bddbee0 pthread_kill + 288
0          0x19bdf34a4 _sigtramp + 56
2023-08-12 13:32:26.958 | Signal: SIGABRT
```

johnnovak: Some related reading material. Even the C++ compiler is quite good and omitting copying on say returning a `std::string`. 99% of the time you return a `std::string` from a function, it's not like it will 1) construct the string in the function, 2) pass a copy to the caller, 3) destruct the copy in the function, but it's actually clever enough to build the string _once_ right at the call site, and the final destination.

So manual `std::move` jugglery _seems_ like a bit of a code smell to me; let the compiler just do it's job, we should not normally need this.

Like I said, Nim copying literally everything by default raised a few eyebrows, and Nim basically generates C sources then compiles it via gcc/clang. Whenever people brought it up, the creator of Nim basically always just went like "yeah, don't worry about it; C compilers are very good at preventing any unnecessary copies".

So I guess that's where I'm coming from; we should not really need this 🤷🏻 Now I'll shut up 😄 

What are copy elision and return value optimization?
https://stackoverflow.com/questions/12953127/what-are-copy-elision-and-return-value-optimization

Copy elision
https://en.cppreference.com/w/cpp/language/copy_elision

kcgen: Yup, the good news is the analyzers aren't recommending we ladden the code with harmful moves, so I don't think there's any risk of us abusing  move 😅 🚬 :booze:

(I suspect PVS would be smart enough to catch the harmful cases, too!).

johnnovak: > Yup, the good news is the analyzers aren't recommending we ladden the code with harmful moves, so I don't think there's any risk of all of us suddenly going on a `std::move` bender! 😅 🚬 :booze:
> 
> (I suspect PVS would be smart enough to catch the harmful cases, too!).

Yeah, and I'm a bit stressed, so perhaps I've been overracting. In that case, apologies 😊

weirddan455: I wish C++ had a destructive move.  Do the cheap move operations (copy a pointer to heap memory or whatever) and then run the destructor on the moved object in one go.  Any further uses become a compile time error.

That Google article sounds like it’s talking about implementing a move constructor for custom classes being complex and error prone (which is true).  But move on standard library classes should be safe.

I may take a look at that example tomorrow because you’ve got me curious now.  I don’t know why it should still be crashing.

FeralChild64: OK, so I guess we are keeping the `std::move` and I can close this remark. It actually makes sense in this case - despite I initially write this function with maintainability and robustness in mind, not with performance.

PR #2753, opened 2023-08-12
Change sdl.opengl.shader_source from std::string_view to std::string

weirddan455:
Fixes a potential use after free bug if the source string becomes invalidated.

@johnnovak I was unable to reproduce the crash on Linux.  Tried with `-Db_sanitize=address,undefined` on both Clang and GCC on your "Do not review" branch.  However, based on the stack trace you provided, I believe this is the root cause. 

We were storing a `std::string_view` in this SDL struct which causes undefined behavior if the backing string becomes invalidated.  I think this should really just be a `std::string`.

It doesn't matter much to me if you want to do the `std::move` stuff or not but this is liable to cause problems down the road regardless.  Quote from https://en.cppreference.com/w/cpp/string/basic_string_view

>  It is the programmer's responsibility to ensure that std::string_view does not outlive the pointed-to character array: 

Note how it says "character array" rather than "string".  I believe this can cause problems even with a copy instead of a move.  If the copy causes a heap reallocation to a different place in memory, that would also invalidate the character array.

I'd be interested to know if this fixes the crash for you.

### DIFF
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -191,7 +191,7 @@ struct SDL_Block {
 		bool use_shader;
 		bool framebuffer_is_srgb_encoded;
 		GLuint program_object;
-		std::string_view shader_source_sv = {};
+		std::string shader_source = {};
 		struct {
 			GLint texture_size;
 			GLint input_size;
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2294,8 +2294,7 @@ dosurface:
 void GFX_SetShader([[maybe_unused]] const std::string &source)
 {
 #if C_OPENGL
-	if (sdl.opengl.shader_source != source)
-		sdl.opengl.shader_source = source;
+	sdl.opengl.shader_source = source;
 
 	if (!sdl.opengl.use_shader)
 		return;
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2039,7 +2039,7 @@ dosurface:
 				if (sdl.opengl.program_object == 0) {
 					GLuint vertexShader, fragmentShader;
 
-					if (!LoadGLShaders(sdl.opengl.shader_source_sv,
+					if (!LoadGLShaders(sdl.opengl.shader_source,
 					                   &vertexShader,
 					                   &fragmentShader)) {
 						LOG_ERR("SDL:OPENGL: Failed to compile shader!");
@@ -2294,8 +2294,8 @@ dosurface:
 void GFX_SetShader([[maybe_unused]] const std::string &source)
 {
 #if C_OPENGL
-	if (sdl.opengl.shader_source_sv != source)
-		sdl.opengl.shader_source_sv = source;
+	if (sdl.opengl.shader_source != source)
+		sdl.opengl.shader_source = source;
 
 	if (!sdl.opengl.use_shader)
 		return;


johnnovak:
And the cigar goes to @weirddan455! 🎉 Congratulations! 😎

That was it, the `string_view` threw a spanner in the works. If I change the `string_view` to `std::string`, the moves work as expected. Zero crashes!

What was really unfortunate in this case is that the _combination_ of using `std::string_view` and `std::move` in a very particular way manifested the bug. In normal debug/release builds reloading/changing shaders would sooner or later result in crashes, or shader compilation errors, or something similarly weird. Sometimes after the 3rd attempt, sometimes after the 100th... It was completely random runtime behaviour depending most likely on the memory layout at any given point in time.

Luckily, it always tripped the UBSAN build for me at the second reload attempt of the same shader.

> @johnnovak I was unable to reproduce the crash on Linux. Tried with -Db_sanitize=address,undefined on both Clang and GCC on your "Do not review" branch. However, based on the stack trace you provided, I believe this is the root cause.

Interesting, if I'm using a sanitizer build as described here, I'm getting the use-after-free crash with 100% reliability on macOS by just executing `glshader = blah` twice in a row.

https://github.com/dosbox-staging/dosbox-staging/blob/main/BUILD.md#make-a-sanitizer-build

---

Ok, so some "retrospective code review" on the use of string views, if I may, because I think it's important to have some agreement of how we're using this quite useful tool. Paging @kcgen , @kklobe , @FeralChild64 , @shermp and @weirddan455 😎 

String views are basically read-only pointer + length pairs to a subrange of a string. All important details about them are described here:
https://en.cppreference.com/w/cpp/string/basic_string_view

So basically they're string subranges; you can pass around read-only "pointer-ranges" to a string without having to pass around copies. Why would you want to do that? For performance optimisation reasons. E.g. in a parser you could pass around string views internally in your parser object that all point to a _single_ big string that is your source code string. So you don't need to create millions of small temporary strings when processing your 2 gigabyte XML file 😎  If the source code string only gets destroyed when the parser object itself is destroyed, and no string views _ever_ "escape" the scope of the parser object, then that's a 100% safe use of string views.

Now, if we don't have such lifecycle guarantees, we need to practice some restraint. E.g. if module/class A owns a string, then passes string views to say module/class B, then B _must_ create a copy of the string if it wants to "hold on to it" indefinitely because it has probably no control over the behaviour of module/class A. Doing read-only operations on string views in the scope of a single function call should be safe most of the time (unless threading is involved, I guess...), but if B wants to hold on to a portion of a string passed in as a string view "for later", in general it must create a copy for safety (otherwise we're back to manual memory management very quickly where you constantly have to think about lifecycles and who creates/frees/owns what, _and_ you need to invent methods how to manage these lifecycle-dependencies manually in more complex cases).

So in this concrete example, that's a variation on the second case as `sdlmain` held on to the passed-in string reference in the`sdl.opengl.shader_source_sv` string view:

```cpp
void GFX_SetShader(const std::string& source)
{
	if (sdl.opengl.shader_source_sv != source) {
		sdl.opengl.shader_source_sv = source;
	}
	...
}
```

This feels like a misuse of string views for optimisation purposes, just to avoid a copy. First, it's generally unsafe for the above outlined reasons, second, we're storing the _whole_ string as a string view, which makes it clear this solution was optimisation-motivated. I'd say using string views for anything else than subranges is a code smell; you can just pass `std::string` references around if you're dealing with the entire string but don't want to copy it.

If we really wanted to avoid a copy (which is kinda premature micro-optimisation, but OK), this would be actually a perfect place to use `std::move`, then `sdl.opengl.shader_source` can become a `std::string`:

```cpp
void GFX_SetShader(const std::string& source)
{
	if (sdl.opengl.shader_source != source) {
		sdl.opengl.shader_source = std::move(source);
	}
	...
}
```

Now things make perfect sense and are 100% safe. I've tested it, it works fine.

Why I'm saying this is premature optimisation? Please look at the file sizes:

image

We can safely conclude even the biggest shaders will stay under 100K. Copying 100K once in a blue moon when you change shaders is practically nothing. Even in the early 90s PC democoders stopped giving a shit optimising every single routine in asm and only went for the 10% that mattered, then the rest was C calls for all those execute-once or twice init functions, data loading, etc... 😄

Anyway, in this case I can leave `std::move` in because I've already done it, and it's really clear what it's doing. All I'm suggesting is to let's try to default to "fact and measurement based" optimisation practices versus "premature micro-optimisation". Not to beat a dead horse, but if we just used `std::string` copies all through the shader pipeline, the bug and this conversation would never have happened, and nobody would've noticed any degraded performance.

Okay, please let me know your thoughts 😄

johnnovak:
@weirddan455 Ok, just realised you were trying to reproduce the bug using the last commit of my branch. I "fixed" it by that point by removing the `std::move`. You should be able to reproduce the crash by using this earlier commit: 44a1c98f8de6a1e3150 (that's right before getting rid of the `std::move`).

johnnovak:
Btw @weirddan455 , we can merge this in if you like, but I've reworked the `string_view` stuff in sdlmain based on your findings in this commit anyway:

https://github.com/dosbox-staging/dosbox-staging/pull/2748/commits/81857e3c4f3401aaa7d5dba770011d56561e78f2

weirddan455:
>  Ok, just realised you were trying to reproduce the bug using the last commit of my branch. I "fixed" it by that point by removing the std::move. You should be able to reproduce the crash by using this earlier commit: https://github.com/dosbox-staging/dosbox-staging/commit/44a1c98f8de6a1e3150a25387ad536f1a39e44ec (that's right before getting rid of the std::move).

I had checked out your latest commit and then manually put the `std::move` in.  I tested against that specific commit just now and still no crashes.  That's the thing about undefined behavior.  Sometimes it just accidentally works :shrug:  You just happened to find a configuration on your system that triggered the bug.  Different OS's are going to have different ways of managing memory.

```cpp
void GFX_SetShader(const std::string& source)
{
	if (sdl.opengl.shader_source != source) {
		sdl.opengl.shader_source = std::move(source);
	}
	...
}
```

^^ The above would be bad, IMO.  It depends on the caller not re-using the string it sent to this function (actually, not even sure it would compile since the string is `cosnt`).  This is where I wish we had destructive move semantics for arguments of the function.  Rather than just passing a reference, the function signature could say "I'm taking ownership of this now.  Don't use it anymore".  But, no real way to do that as far as I know.

Agree with everything else you said.  The cost of copying 20KB or on shader change is negligible compared to compiling the shader, sending it to the GPU, and everything else it needs to do.

weirddan455:
> Btw @weirddan455 , we can merge this in if you like, but I've reworked the `string_view` stuff in sdlmain based on your findings in this commit anyway:
> 
> [1cab3fa](https://github.com/dosbox-staging/dosbox-staging/commit/1cab3fa23a676d3bbc0a101cb25bd354fec536b7)

It's up to you.  I think the current code in `main` is bugged.  Just a rare, hard to reproduce memory bug.  Luckily with your PR, you were able to reproduce it.

I see you removed `string_view` in a couple other places in that commit.  I was eyeing those helper functions and thinking about changing to `cosnt std::string&` myself.  I'm fine with not merging this if you want to just do it all in your PR.

kcgen:
Pointers, references, (and string_views) let you have one point of truth for a given object or piece of data, and let you avoid synchronizing N full copies. 

`std::string&` is very clear in that regard, so yeah, fully agree with the suggestion @johnnovak !

And like here, where the shader source is used in both the render.cpp and sdlmain.cpp sources, we can programmatically express who is the "owner" by using a reference or std::move plus a clear() to wipe out the old object (not about saving 100k of RAM and more about describing who has the ball).

johnnovak:
> ^^ The above would be bad, IMO. It depends on the caller not re-using the string it sent to this function (actually, not even sure it would compile since the string is `cosnt`). This is where I wish we had destructive move semantics for arguments of the function. Rather than just passing a reference, the function signature could say "I'm taking ownership of this now. Don't use it anymore". But, no real way to do that as far as I know.

That's not my understanding how moving this works and I'm convinced this is 100% safe. Please read the following SO:
https://stackoverflow.com/questions/47841596/where-to-use-stdmove-for-strings

I'm not trying to "prove" who's right or wrong here, just wanna make sure I haven't misunderstood the semantics 😅 I think it's important that we come to the same understanding regarding at least the basic usage of `std::move`, so yeah, please read the SO guys and let me know if you think what I've done in the above snippet is dangerous.

weirddan455:
>  Please read the following SO:
https://stackoverflow.com/questions/47841596/where-to-use-stdmove-for-strings

In the example code on that post, the string was passed by value.  That's the reason it's safe there.  It's not safe when you're passing by reference (unless you guarantee the caller doesn't try to use it again, but that type of code leads to hard to find bugs).

See https://en.cppreference.com/w/cpp/utility/move

>  Unless otherwise specified, all standard library objects that have been moved from are placed in a "valid but unspecified state", meaning the object's class invariants hold (so functions without preconditions, such as the assignment operator, can be safely used on the object after it was moved from): 

And https://en.cppreference.com/w/cpp/language/move_assignment

>  Move assignment operators typically "steal" the resources held by the argument (e.g. pointers to dynamically-allocated objects, file descriptors, TCP sockets, I/O streams, running threads, etc.), rather than make copies of them, and leave the argument in some valid but otherwise indeterminate state. For example, move-assigning from a [std::string](https://en.cppreference.com/w/cpp/string/basic_string) or from a [std::vector](https://en.cppreference.com/w/cpp/container/vector) may result in the argument being left empty. This is not, however, a guarantee. A move assignment is less, not more restrictively defined than ordinary assignment; where ordinary assignment must leave two copies of data at completion, move assignment is required to leave only one.

I'd recommend reading the whole article but gist of it is that after a move, the state of the object is "valid but undefined".  You can do things like use the assignment operator on it after a move because that doesn't care about current state but even calling `.empty()` on the moved string is undefined.

weirddan455:
Added one quick fix to this PR.  It doesn't really make sense to do a string compare to check if we need to copy... just copy.  String compare is probably slower than a copy anyway.

You can merge this when the CI passes if you like.

johnnovak:
> In the example code on that post, the string was passed by value. That's the reason it's safe there. It's not safe when you're passing by reference (unless you guarantee the caller doesn't try to use it again, but that type of code leads to hard to find bugs).

Oh man... 😩 Okay... I'm getting the feeling again that I'm really dealing with a super overcomplicated assembler here instead of a proper high-level language, having to care about all that low-level details myself... 🥶 

Okay, I'll just use `std::move` with non-reference types and job done. I really don't wanna waste mental cycles on this stuff. Looks at the `(2)` example here. Just nope.

Thanks, btw 😄 

> I'd recommend reading the whole article but gist of it is that after a move, the state of the object is "valid but undefined". You can do things like use the assignment operator on it after a move because that doesn't care about current state but even calling `.empty()` on the moved string is undefined.

Yeah, I'll read it detail later, but in general I'll steer clear of it.

johnnovak (approved):
Merge away man, copying `std::string` it is 😎

johnnovak:
@weirddan455 @kcgen 

Sorry... C++ made me do it 😎 

image

kcgen:
Only in C++ ...

@weirddan455 @kcgen: *"bro, you gonna risk moving an rvalue ref?"*
@johnnovak: *"nah, imma raw dog it with a const ref copy."*

🤘 🤣

weirddan455:
`std::move` is just a cast to R value reference is the easiest way to think about it.  cppreference even says "It is exactly equivalent to a static_cast to an rvalue reference type."

It's like saying "pretend this named variable is an un-named temporary".  Then when you assign it to something else, it assumes you don't care about the state afterwards because it's the same as if you provided a temporary.  It just has to be in some valid state so it can be deconstructed later.

If you do that to a passed by reference variable, your function signature is saying "I'm just gonna borrow this for the duration of my function (and I won't modify it if `const`)" and then you just go ahead and give it to someone else.

I checked and it actually does compile, even with `const` but that's because `std::move` just casts that away.  In a lot of situations, casts aren't safe either so it helps me to think about it in that way.

weirddan455:
Anyway, tests are passed now but I can't merge so one of you has to hit the button.

merge

kklobe:
If anyone's interested in a very, very thorough treatment: http://www.cppmove.com

johnnovak:
> If anyone's interested in a very, very thorough treatment: http://www.cppmove.com

Thanks, @kklobe. A testimonial from that linked page:

> Sometimes I think I have a better grasp on entanglement and quantum teleportation than I do in some weird C++ move semantics. To paraphrase Feynman: If you think you understand C++ move semantics, you do not understand C++ move semantics. Read this book.

Which was my point all along. @kcgen assured me we won't suddenly go on a std::move-everything-spree, so I'm not _really_ worried, but my general concern is that liberal use of std::move would raise the barrier of entry too much for new people (apart from adding too much mental overhead... well, it's too much for me personally, I can say as much).

It's no wonder virtually all C++ teams have style guides that _forbid_ using large chunks of the C++ language, restricting it to a most useful non-esoteric subset. Otherwise you get a team where everybody understands and uses a different non-overlapping 20% of this huge monster of a language.

Staging has been quite good at restricting C++ use to a sane and mostly uncomplicated subset and things have been working well. I guess I like to keep it that way 😄

PR #2756, opened 2023-08-13
Another batch of PVS Studio fixes

FeralChild64:
- fixed 30 more warnings
- regarding loguru changes - it seems the project is still being developed (like few commits per year), so once this gets merged I intend to create a GitHub Issue in the loguru project

### DIFF
--- a/src/dos/program_autotype.cpp
+++ b/src/dos/program_autotype.cpp
@@ -78,9 +78,9 @@ void AUTOTYPE::PrintKeys()
  */
 bool AUTOTYPE::ReadDoubleArg(const std::string &name,
                              const char *flag,
-                             const double &def_value,
-                             const double &min_value,
-                             const double &max_value,
+                             const double def_value,
+                             const double min_value,
+                             const double max_value,
                              double &value)
 {
 	bool result = false;
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 237
+          MAX_BUGS: 207
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -298,7 +298,7 @@ private:
 	struct StereoLine {
 		LINE_INDEX left = LEFT;
 		LINE_INDEX right = RIGHT;
-		bool operator==(const StereoLine &other) const;
+		bool operator==(const StereoLine other) const;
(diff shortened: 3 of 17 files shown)


src/dos/program_autotype.cpp
```diff
@@ -78,9 +78,9 @@ void AUTOTYPE::PrintKeys()
  */
 bool AUTOTYPE::ReadDoubleArg(const std::string &name,
                              const char *flag,
-                             const double &def_value,
-                             const double &min_value,
-                             const double &max_value,
+                             const double def_value,
+                             const double min_value,
+                             const double max_value,
```
weirddan455, line 83: const ref to a double?  Yeah, get that outta here.  :+1:

kcgen:
The loguru author hasn't given enough people rights to keep the project actively maintained, so it's gotten in a bad state with lots of forks.  

Their own `master` branch fails in their own CI, and they haven't even accepted PRs to get it passing (the only green PRs in their queue include the fix to get their CI working again).

I've got queued PRs that fix real issues, like:
- https://github.com/emilk/loguru/pull/188

And sanitizer cleanups:
- https://github.com/emilk/loguru/pull/179

Back when we adopted it, I forked the project here: https://github.com/dosbox-staging/loguru, and then applied the above fixes to both our fork and the upstream (via PRs, which are stills stagnant).

To then get these changes into Staging, I simply copies the latest source file in a one-shot commit "Sync loguru from upstream" (just like all our other ported-in sources), but in this case, the upstream is our own DOSBox-organization fork of loguru.

The seems convoluted, but the benefit is that we can easily keep the fork of loguru rebased against upstream, if/when it gets more active.

Currently our fork is still ahead by ~35 commits (and you can see 245 people have launched new forks from ours!):

https://github.com/emilk/loguru/compare/master...dosbox-staging:loguru:master

The delta are commits that we don't want or cause more warnings and analysis issues. 

----

So I would suggest PRing your good Loguru fixes against both Staging loguru and also upstream's Loguru.  I've added you as admin to the Loguru project here, so you can simply merge.

Once Staging org's Loguru has your fixed commits, then simply copy that new source .cpp/.hpp file into Staging's repo as a single commit `"Sync loguru with upstream"`.

If the upstream loguru does accept your PR at some point, then we just have to rebase against upstream here, and we can continue to keep our loguru project in-sync with little effort.

weirddan455 (approved):
These all look like good low risk changes.  I like seeing those warnings continue to get lower.  Nice job.

kcgen (approved):
Very nice cleanup, @FeralChild64. 
Was able to review all the commits; looks good to me.

PR #2760, opened 2023-08-15
Improve compile commits script

johnnovak:
- Compile commits in forward order, starting from the oldest one
- Check out the branch itself when done instead of the head commit by hash

☝🏻 Should be pretty straighforward. Creating this script was a great idea @kcgen and it's super handy—I just wished it worked a bit differently... now it does 😎 

Feel feel to disagree with any of this; I just find this slightly tweaked mode of operation more useful. Especially checking out the head by commit hash vs the actual branch name at the end trips me up every single time 😅 I would expect the script to restore the "Git state" exactly to where it was started from.

### DIFF
--- a/scripts/compile_commits.sh
+++ b/scripts/compile_commits.sh
@@ -24,6 +24,7 @@ set -euo pipefail
 # Common variables that are setup and used below
 declare temp_file
 declare head_commit
+declare current_branch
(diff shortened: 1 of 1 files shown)


kcgen (approved):
Looks good, and nice improvement!

PR #2450, opened 2023-05-03
Cleanup SDL `output` handling

GranMinigun:
~~This PR fulfils the prophecy of surface output meeting its demise in cleansing flames of Purgatory. They will not be missed.~~ Extracted into #2780.

As a side bonus, keymapper was ported to SDL renderer and now reuses the window instead of creating a new one (extracted into #2763). Some functionality was moved into separate routines to make initialization and handling of video mode changes less painful to read.

Some (a lot of) testing and feedback is required. Almost all of `[sdl]` configuration block is potentially affected. While testing all possible combinations is an overkill, ensuring that each setting works with both remaining outputs on every platform would be nice: Windows, macOS, X11, Wayland, KMS/DRM. Also please verify whether HiDPI and multi-monitor still work as expected. One interesting case is to check how `fullresolution = original` behaves on a real CRT, if at all.

If you feel I've missed some comments (or punctuation in them), please let me know. Same for moving more stuff around, checks, etc. I'd like to get feedback, else this work could easily drag for six months more without much progress.

Part of work outlined in #2131.

PR #2763, opened 2023-08-15
Port keymapper to SDL renderer

GranMinigun:
Does what it says: moves mapper from software surface to SDL's rendering API. The emulator window is now reused instead of creating a new one.

If you're not happy with the mapper's pseudo-lambdas and how they're formatted, you're welcome to steal this branch. These changes would go far beyond the scope of the PR though, and the whole thing will be eventually rewritten anyway, hence why I'm leaving it as is.

While it seems to work with the surface output, this is not intentional, and it will be removed in a subsequent PR anyway. (Damn, that sounds broken...)

Extracted from - and is a prerequisite to - #2450.

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -60,15 +60,11 @@ constexpr Rgb888 marginal_color(255, 103, 0); // Amber for marginal conditions
 constexpr Rgb888 on_color(0, 1, 0);           // Green for on/ready/in-use
 constexpr Rgb888 off_color(0, 0, 0);          // Black for off/stopped/not-in-use
 
-enum {
-	CLR_BLACK=0,
-	CLR_GREY=1,
-	CLR_WHITE=2,
-	CLR_RED=3,
-	CLR_BLUE=4,
-	CLR_GREEN=5,
-	CLR_LAST=6
-};
+constexpr Rgb888 color_black(0, 0, 0);
+constexpr Rgb888 color_grey(127, 127, 127);
+constexpr Rgb888 color_white(255, 255, 255);
+constexpr Rgb888 color_red(255, 0, 0);
+constexpr Rgb888 color_green(0, 255, 0);
 
 enum BB_Types {
 	BB_Next,BB_Add,BB_Del,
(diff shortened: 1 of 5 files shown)


kcgen:
Awesome overhaul and function improvement.
 - De-Bitu
 - `#define constants` to constexpr; now we can see variables in debuggers and not literals
 - `#define functions` to lamdas: now we can get valid backtraces

The functional chunks of code are made more readable.

johnnovak (approved):
Good stuff @GranMinigun! These changes improve readability a lot too. 🎉 

No comments from me as the days of the old mapper are numbered, like you mentioned. We only need to keep it limping along for a bit longer. In a way it got a lot more love from you than it deserves 😎 

(On a related note, glad to see you returning to our circles, man. I've been missing you, your graphics pipeline related insights, and the snarky remarks 😎🤘🏻 )

kcgen (approved):

kcgen:
> @GranMinigun "Sanitizers seem to be happy here"

:+1: 

Scaling perfectly to various DPIs on Linux and macOS:

mapper

johnnovak:
@GranMinigun @kcgen Ok, that merging was a bit premature, so it seems.

I can open the mapper fine on macOS and the coordinate mappings work fine, so it's fully functional. But the Exit buttond does _not_ hide it, so once opened, I'm stuck with it.

johnnovak:
Shall I create a regression ticket for this @GranMinigun or are looking into it? This broke the mapper in main at least on macOS.

GranMinigun:
I'm not sure how you're unable to exit from the mapper. Definitely cannot reproduce here. No need for a separate ticket, it can be discussed here.

johnnovak:
> I'm not sure how you're unable to exit from the mapper. Definitely cannot reproduce here. No need for a separate ticket, it can be discussed here.

In OpenGL `output` modes if I click on Exit, the mapper never disappears. That's it, plain and simple! 😄 Switching to fullscreen then back to windowed doesn't fix it, changing the windows size doesn't fix it, nothing fixes it. The mapper screen never disappears, and the Exit button stays higlighted:

image

The emulator keeps working though. So I'm a sneaky bastard and typing `output = texture` in the console, completely blind, the lo and behold:

image

So somehow the Exit keypress never triggers the "hide mapper" event or something. 

If I use `output = texture`, it works fine; Exit closes the mapper.

johnnovak:
@GranMinigun Works fine on Windows 10 🥶 SDL... 😐 

This is a bit of a problem because you don't have a Mac... I guess it will be @kklobe, @kcgen or myself who fixes it.

kcgen:
Sorry for merging too soon. Been offline, but will check it out tomorrow w/ @GranMinigun's help.

GranMinigun:
Two things to check: whether other renderer backends work (e.g. `texture_renderer=opengl` and others); and whether `context` variable gets populated properly and is still valid by the time it is made current.

kcgen:
> Two things to check: whether other renderer backends work (e.g. `texture_renderer=opengl` and others)

That's it @GranMinigun; results from macos:

This exits the mapper properly:

```ini
output = opengl
texture_renderer = opengl
```

These also exit properly:

```ini
output = texture
texture_renderer = (all: metal, software, opengles2, opengl)
```

However these don't:

```ini
output = opengl
texture_renderer = metal, opengles2, or software
```

The `context` is always getting populated.

GranMinigun:
Alright, I see software renderer outright crashes on my side. Will think about what to do and raise a follow-up PR.

PR #2766, opened 2023-08-17
Fix comment describing BGRX8888 PixelFormat

weirddan455:
This briefly tripped me up working on ffmpeg stuff.  I had to map this format to what ffmpeg calls `AV_PIX_FMT_BGR0`, implying that it's the low bits that are unused, not the high bits.  I had tried `AV_PIX_FMT_0BGR` first and the colors were all wrong.

It's also what the enum name suggests.  If it were really the high bits that were unused, I would expect it to be called `XBGR8888` instead.

### DIFF
--- a/include/render.h
+++ b/include/render.h
@@ -64,7 +64,7 @@ enum class PixelFormat : uint8_t {
 	BGR888 = 24,
 	//
 	// 16.7M (32-bit) true colour; 8 bits per red/blue/green component;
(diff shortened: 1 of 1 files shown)


johnnovak:
Sorry for the confusion; I remember first I used `XBGR8888` then changed it later when I realised it's wrong, but apparently forgot to update the comment.

PR #2769, opened 2023-08-18
Monochrome mode fixes & improvements

johnnovak:
This fixes a couple of issues with the Hercules and CGA Monochrome modes when reloading shaders. The dynamic shader stuff keeps uncovering these state management issues, which is good in a way.

I've taken the liberty of changing the default monochrome mode to `amber` from `white` because amber displays are the coolest mono CRTs (change my mind! well, you can't 😎), while `white` is a bit... meh. Also changed the cycling order to `amber`, `green`, `white` and `paperwhite` because I'd say amber and green are the most iconic phosphor types.

Another important thing: the config settings should be always kept in sync with the actual settings in use. For example, cycling the mono palette with hotkeys did not keep the `monochrome_palette` setting in sync with what you actually see on the screen—now it does. This lack of syncing will become a big issue with the OSD and when adding support for saving settings configured via the OSD into the config, so better start fixing stuff up now. Well, I needed this for the shader reloading stuff as well (the way I implemented it) so it won't reset the monochrome palette on shader reloads anymore.

For the record, almost none of the settings are kept in sync, so all of them need to be revisited at some point (e.g., try `cycles = 4000`, change it via hotkeys, then query it again with `cycles`—it will still print out 4000... umm 😐).

---

I've also simplified the CGA monochrome mode, according to the intent of the original author. The "bright" mode was brightness/contrast tweaked version intended for 16-colour CGA text modes, while the "non-bright" variant was meant for 4-colour CGA graphics modes. This "bright" setting was really confusing; I never understood why enabling "bright mode" actually results in a _flatter_ and _darker_ looking image in graphics modes... well, now I know—because it was never intended for graphics modes!

So I've removed the manual "bright mode" toggle; now the correct mono palette variant is set based on whether we're in a text or graphics mode.

Further details in the commit, plus this is the image from VileR, the author of the CGA mono patch, that clearly illustrates the differences and the intended usage:

image

---

And now some obligatory screenshots because I love how the amber mode looks 😎🤘🏻 Did I say it's the greatest mono mode? 😆 

### Hercules

image

image

image

### Monochrome CGA

image

image


src/misc/setup.cpp
```diff
@@ -873,6 +873,21 @@ const char* Section_prop::Get_string(const std::string& _propname) const
...
+	for (const_it tel = properties.begin(); tel != properties.end(); ++tel) {
+		if ((*tel)->propname == propname) {
+			Prop_string* val = dynamic_cast<Prop_string*>((*tel));
+			if (val) {
+				return val;
+			} else {
+				return nullptr;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
kcgen: (I think the `properties` are a container of pointers, so that's why I'm just looping by const-value instead of reference; but I haven't tried it)

src/hardware/vga_other.cpp
```diff
@@ -1218,84 +1122,99 @@ static void write_pcjr(io_port_t port, io_val_t value, io_width_t)
...
 }
 
-static void CycleMonoCGAPal(bool pressed) {
-	if (!pressed) return;
-	if (++mono_cga_pal>3) mono_cga_pal=0;
-	Mono_CGA_Palette();
-}
+void VGA_SetMonochromeCgaPalette()
+{
+	constexpr auto num_cga_colors = 16;
```
kcgen, line 1168: Might be able to programmatically get it with:

```c++
constexpr auto num_cga_colors = ARRAY_LEN(mono_cga_palettes[0]);
```

src/hardware/vga_other.cpp
```diff
@@ -256,163 +256,63 @@ enum class COMPOSITE_STATE : uint8_t {
...
-                {0x2b, 0x2c, 0x2b},
-                {0x2d, 0x2e, 0x2d},
-                {0x34, 0x35, 0x33},
-                {0x37, 0x37, 0x34},
-                {0x3e, 0x3e, 0x3a},
-                {0x3f, 0x3f, 0x3b},
-        }};
+	{
+		// 0 - Amber, 4-colour-optimised contrast
+		{0x00, 0x00, 0x00}, {0x15, 0x05, 0x00}, {0x20, 0x0b, 0x00}, {0x24, 0x0d, 0x00},
```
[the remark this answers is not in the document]
johnnovak: Very nice suggestion; it just worked! 👍🏻

src/misc/setup.cpp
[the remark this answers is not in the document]
johnnovak: Thanks, that just worked!

src/hardware/vga_other.cpp
johnnovak: I've introduced a constant for it.

johnnovak:
I've addressed the review comments @kcgen, plus tinkering with the code a bit more led to a few further improvements. See the commit messages for details, plus I've extended the top comment as well (in a nutshell, I've removed the "bright" CGA mono mode—now it's set automatically).

kcgen:
> The dynamic shader stuff keeps uncovering these state management issues, which is good in a way.

nice!

> Another important thing: the config settings should be always kept in sync with the actual settings in use.

Great catch. Yes, hotkeys actions should change conf/setup values in a given section. Then the section can pass the updated settings to its supporting modules as a `ModuleLifecycle::Reconfigure` request.

The other benefit is that users writing their conf after hotkey tweaks will write the updates.

>  never understood why enabling "bright mode" actually results in a flatter and darker looking image in graphics modes... well, now I know—because it was never intended for graphics modes!

> So I've removed the manual "bright mode" toggle; now the correct mono palette variant is set based on whether we're in a text or graphics mode.

looks just how I remember it.. this is bringing these games back in their original glory :rocket: :metal: ; awesome stuff, @johnnovak.

include/vga.h
```diff
@@ -135,6 +135,8 @@ enum class ColorDepth {
 
 const char* to_string(const ColorDepth c);
 
+constexpr auto NumCgaColors = 16;
```
kcgen, line 138: :+1:

include/vga.h
```diff
@@ -957,13 +957,18 @@ struct VGA_Type {
...
-/* CGA Mono Palette function */
-void Mono_CGA_Palette(void);
+// Hercules & CGA monochrome palette
+enum class MonochromePalette : uint8_t {
+	Amber      = 0,
+	Green      = 1,
+	White      = 2,
+	Paperwhite = 3
+};
+constexpr auto MaxMonochromePalette = MonochromePalette::Paperwhite;
```
kcgen, line 967: Maybe replace the Max with the Num:
```c++
constexpr NumMonochromePalettes = auto static_cast<int>(MonochromePalette::Paperwhite) + 1
```

And could be used directly in:

```c++
auto value = (enum_val(palette) + 1) % (enum_val(MaxMonochromePalette) + 1);
```

(I didn't see other uses for the `MaxMonochromePalette`, so figured it would be a useful switch)

src/ints/int10_modes.cpp
```diff
@@ -679,6 +679,17 @@ void INT10_SetCurMode(void) {
...
 
+bool INT10_IsTextMode(const VideoModeBlock& mode_block)
+{
+	switch (mode_block.type) {
+	case M_TEXT:
+	case M_HERC_TEXT:
+	case M_TANDY_TEXT:
+	case M_CGA_TEXT_COMPOSITE: return true;
+	default: return false;
+	}
```
kcgen, line 690: :+1:

kcgen (approved):
Just a couple minor suggestions, and approved :-)
Nothing more from me :sweat_smile: , whenever you're ready.

include/vga.h
johnnovak: Yeah, I was hesitating which one is preferable. Num is better in this case and generally more useful, I agree.

GranMinigun (approved):
Can't give any comments code-wise. Quickly checked in Rogue (Hercules and CGA text), Arcticfox (Hercules and CGA), King's Bounty (it allows switching CGA palette on the fly, how cool is that!), Dangerous Dave (CGA), Might and Magic (Hercules and CGA). Everything seems to be in order.

johnnovak:
> Can't give any comments code-wise. Quickly checked in Rogue (Hercules and CGA text), Arcticfox (Hercules and CGA), King's Bounty (it allows switching CGA palette on the fly, how cool is that!), Dangerous Dave (CGA), Might and Magic (Hercules and CGA). Everything seems to be in order.

Thanks for testing @GranMinigun. Did not know that about King's Bounty in CGA mode! I'm very fond of that game, I played it a lot on the C64 but could never complete it because I had a bad crack... I've been meaning to replay all ports; the B/W Mac version looks glorious.

Took me a while to find how to change the CGA palette (`C` key followed by `6`, then arrow keys):

image

Roadwar 2000 looks kinda cool in green mono mode too:

image

johnnovak:
> The other benefit is that users writing their conf after hotkey tweaks will write the updates.

Yeah, that's true, so there's definitely benefit from the config setting syncing even before the OSD stuff.

PR #2773, opened 2023-08-19
Add missing mutex lock to RWQueue

weirddan455:
From https://en.cppreference.com/w/cpp/thread/condition_variable

>  Even if the shared variable is atomic, it must be modified while owning the mutex to correctly publish the modification to the waiting thread.


kcgen (approved):
Nice improvement, @weirddan455!

PR #2775, opened 2023-08-20
Notify all when RWQueue stops

weirddan455:
Found another bug in RWQueue.  We need to notify all threads when the queue stops or we can have a deadlock.

I'm pretty sure I'll need to add a couple more methods to this class as well once I get FFmpeg threading sorted out but  that'll go in my big PR alongside the use case.  At a minimum, I think I'll need a `Drain` method and a `Start` method so that I can `Stop` the queue, `Drain` all remaining elements from it in a non-blocking manner, and then `Start` it back up again.

For now though, I'm just pushing these bug fixes in small PRs as I see them since this could cause a stall in existing code.

### DIFF
--- a/src/misc/rwqueue.cpp
+++ b/src/misc/rwqueue.cpp
@@ -56,8 +56,8 @@ void RWQueue<T>::Stop()
 	mutex.unlock();
 
 	// notify the conditions
-	has_items.notify_one();
-	has_room.notify_one();
+	has_items.notify_all();
+	has_room.notify_all();
 }
 
 template <typename T>


kcgen:
Looks good 👍
For draining, check out the other uses in the midi code (how we drain the queue on shutdown), which hopefully can be used.

kcgen (approved):

kcgen:
In the main thread, call `.Stop()` on the queue, after which wait for the thread to become joinable.

Then inside the threaded processing function, always check the `.IsRunning()` state on the queue after each unit of work. We usually have it as a while-condition that bails out once it's true. 

`git grep IsRunning` can find the chunks where we do this in the capture and midi code. 

Hope it helps!

PR #2776, opened 2023-08-20
Another batch of PVS Studio fixes

FeralChild64:
Fixed 23 more PVS Studio warnings. I suggest to review commit-by-commit.

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -89,18 +89,7 @@ class DOS_DTA;
 
 class DOS_File {
 public:
-	DOS_File()
-	        : flags(0),
-	          time(0),
-	          date(0),
-	          attr(0),
-	          refCtr(0),
-	          open(false),
-	          name(""),
-	          newtime(false),
-	          hdrive(0xff)
-	{}
-
+	DOS_File() = default;
 	DOS_File(const DOS_File &orig) = default;
 	DOS_File &operator=(const DOS_File &orig);
 
@@ -129,17 +118,17 @@ public:
 
 	void SetDrive(uint8_t drv) { hdrive=drv;}
 	uint8_t GetDrive(void) { return hdrive;}
-	uint32_t flags;
-	uint16_t time;
-	uint16_t date;
-	uint16_t attr;
-	Bits refCtr;
-	bool open;
-	std::string name;
-	bool newtime;
+	uint32_t flags   = 0;
+	uint16_t time    = 0;
+	uint16_t date    = 0;
+	uint16_t attr    = 0;
+	Bits refCtr      = 0;
+	bool open        = false;
+	std::string name = {};
+	bool newtime     = false;
 	/* Some Device Specific Stuff */
 private:
-	uint8_t hdrive;
+	uint8_t hdrive = 0xff;
 };
 
 class DOS_Device : public DOS_File {
--- a/src/dos/program_imgmount.cpp
+++ b/src/dos/program_imgmount.cpp
@@ -291,11 +291,12 @@ void IMGMOUNT::Run(void)
 		}
 
 		// Test if input is file on virtual DOS drive.
+		constexpr int posix_success = 0;
 		struct stat test;
-		if (stat(temp_line.c_str(), &test)) {
+		if (posix_success != stat(temp_line.c_str(), &test)) {
 			// See if it works if the ~ are written out
 			const auto home_dir = resolve_home(temp_line).string();
-			if (!stat(home_dir.c_str(), &test)) {
+			if (posix_success == stat(home_dir.c_str(), &test)) {
 				temp_line = home_dir;
 			} else {
 				// convert dosbox filename to system filename
@@ -326,7 +327,7 @@ void IMGMOUNT::Run(void)
 				ldp->GetSystemFilename(tmp, fullname);
 				temp_line = tmp;
 
-				if (stat(temp_line.c_str(), &test)) {
+				if (posix_success != stat(temp_line.c_str(), &test)) {
 					if (add_wildcard_paths(temp_line, paths)) {
 						continue;
 					} else {
--- a/src/dos/program_imgmount.cpp
+++ b/src/dos/program_imgmount.cpp
@@ -278,8 +278,8 @@ void IMGMOUNT::Run(void)
 				        temp_line.c_str());
 			}
 		} else {
-			const auto home_resolve = resolve_home(temp_line).string();
-			if (home_resolve == real_path) {
+			const auto home_dir = resolve_home(temp_line).string();
+			if (home_dir == real_path) {
 				LOG_MSG("IMGMOUNT: Path '%s' found",
 				        temp_line.c_str());
 			} else {
@@ -294,9 +294,9 @@ void IMGMOUNT::Run(void)
 		struct stat test;
 		if (stat(temp_line.c_str(), &test)) {
 			// See if it works if the ~ are written out
-			const auto homedir = resolve_home(temp_line).string();
-			if (!stat(homedir.c_str(), &test)) {
-				temp_line = std::move(homedir);
+			const auto home_dir = resolve_home(temp_line).string();
+			if (!stat(home_dir.c_str(), &test)) {
+				temp_line = home_dir;
 			} else {
 				// convert dosbox filename to system filename
 				char fullname[CROSS_LEN];
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -140,9 +140,8 @@ static void empty_line_handler(const void*) {}
 static void start_line_handler(const void* s)
 {
 	if (s) {
-		auto src   = static_cast<const Bitu*>(s);
-		auto cache = static_cast<Bitu*>(
-		        static_cast<void*>(render.scale.cacheRead));
+		auto src   = static_cast<const uintptr_t*>(s);
+		auto cache = reinterpret_cast<uintptr_t*>(render.scale.cacheRead);
 		for (Bits x = render.src.start; x > 0;) {
 			const auto src_ptr = reinterpret_cast<const uint8_t*>(src);
 			const auto src_val = read_unaligned_size_t(src_ptr);
@@ -172,9 +171,8 @@ static void start_line_handler(const void* s)
 static void finish_line_handler(const void* s)
 {
 	if (s) {
-		auto src   = static_cast<const Bitu*>(s);
-		auto cache = static_cast<Bitu*>(
-		        static_cast<void*>(render.scale.cacheRead));
+		auto src   = static_cast<const uintptr_t*>(s);
+		auto cache = reinterpret_cast<uintptr_t*>(render.scale.cacheRead);
 		for (Bits x = render.src.start; x > 0;) {
 			cache[0] = src[0];
 			x--;
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -140,8 +140,9 @@ static void empty_line_handler(const void*) {}
 static void start_line_handler(const void* s)
 {
 	if (s) {
-		const Bitu* src = (Bitu*)s;
-		Bitu* cache     = (Bitu*)(render.scale.cacheRead);
+		auto src   = static_cast<const Bitu*>(s);
+		auto cache = static_cast<Bitu*>(
+		        static_cast<void*>(render.scale.cacheRead));
 		for (Bits x = render.src.start; x > 0;) {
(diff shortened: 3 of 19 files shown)


src/hardware/serialport/misc_util.cpp
```diff
@@ -31,6 +31,15 @@
...
 constexpr int connection_timeout_ms = 5000;
 
+std::string to_string(const SocketType socket_type)
+{
+	switch (socket_type) {
+	case SocketType::Tcp: return "TCP";
+	case SocketType::Enet: return "ENet";
+	default: assert(false); return "(unknown)";
+	}
+}
```
kcgen, line 41: suggest returning a `const char* const`, because that's what they are (all `"string_literal"s` are permanent and pinned into the executable, their memory and accessibility is permanent)

src/hardware/serialport/misc_util.h
```diff
@@ -60,7 +60,13 @@
 
 #include "enet/include/enet.h"
 
-enum SocketTypesE { SOCKET_TYPE_TCP = 0, SOCKET_TYPE_ENET, SOCKET_TYPE_COUNT };
+enum class SocketType {
+	Tcp  = 0,
+	Enet = 1,
+	Invalid,
+};
+
```
kcgen, line 68: nice cleanup and error handling with this!

include/dos_system.h
```diff
@@ -89,18 +89,7 @@ class DOS_DTA;
...
-	          date(0),
-	          attr(0),
-	          refCtr(0),
-	          open(false),
-	          name(""),
-	          newtime(false),
-	          hdrive(0xff)
-	{}
-
+	DOS_File() = default;
```
kcgen, line 92: nice to get these moved over to the default initializers :+1:

src/gui/render_simple.h
```diff
@@ -45,8 +45,8 @@ static void conc4d(SCALERNAME,SBPP,DBPP,R)(const void *s) {
 #endif
 	/* Clear the complete line marker */
 	Bitu hadChange = 0;
-	const SRCTYPE *src = (SRCTYPE*)s;
-	SRCTYPE *cache = (SRCTYPE*)(render.scale.cacheRead);
+	auto src   = static_cast<const SRCTYPE*>(s);
+	auto cache = static_cast<SRCTYPE*>(static_cast<void*>(render.scale.cacheRead));
```
kcgen, line 49: Can `reinterpret_cast<SRCTYPE*>(render.scale.cacheRead);` avoid the double cast?

src/hardware/serialport/nullmodem.h
```diff
@@ -51,7 +51,7 @@ class CNullModem final : public CSerial {
 	void setDTR(bool val) override;
 	void handleUpperEvent(uint16_t type) override;
 
-	SocketTypesE socketType = SOCKET_TYPE_TCP;
+	SocketType socketType = SocketType::Tcp;
```
kcgen, line 54: nice! Can the `SOCKET_TYPE_TCP` and `SOCKET_TYPE_ENET` definitions be eliminated (where ever they are?)

src/hardware/serialport/softmodem.cpp
```diff
@@ -250,7 +250,7 @@ void CSerialModem::SendNumber(uint32_t val)
 }
 
 void CSerialModem::SendRes(const ResTypes response) {
-	char const * string = nullptr;
+	const char* string = nullptr;
```
kcgen, line 253: (this `string` variable name is pretty bad.. like calling a variable `int` :grimacing: :laughing: !)

This function sends either a number-code or text response, based on the incoming response-type enum. This `string` variable is only used to hold the text response (in that code path). 

Maybe rename the variable (or eliminate the need for it using a lamda that returns the literal char*)?

src/hardware/serialport/misc_util.cpp
kcgen: Will also cleanup the need for re-converting back to the `char*`s in a couple places:

to_string(socketType).c_str());

kcgen (changes requested):
Just a handful of minor suggestions. Looks great, @FeralChild64 !

src/dos/program_imgmount.cpp
```diff
@@ -294,9 +294,9 @@ void IMGMOUNT::Run(void)
 		struct stat test;
 		if (stat(temp_line.c_str(), &test)) {
 			// See if it works if the ~ are written out
-			const auto homedir = resolve_home(temp_line).string();
-			if (!stat(homedir.c_str(), &test)) {
-				temp_line = std::move(homedir);
+			const auto home_dir = resolve_home(temp_line).string();
+			if (!stat(home_dir.c_str(), &test)) {
+				temp_line = home_dir;
```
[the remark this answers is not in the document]
weirddan455: As a sidenote if you're using POSIX functions like `stat`, I find it hard to read when you check for error conditions like this.  POSIX functions generally return 0 on success so saying `if (stat(...))` makes it look at first glance like you're checking for success when really you're checking for error.  And similarly `if (!stat(...))` is actually checking for success.

You could extract into a constant `constexpr int posix_success = 0` and compare against that or even just `if (stat(...) == 0`  `if (stat(...) < 0)` is better.

I know this is just a cleanup commit and you didn't write this but boolean expressions being inverted like this is confusing.

src/hardware/serialport/nullmodem.h
FeralChild64: They are now gone,transformed into `SocketType::Tcp` and `SocketType::Enet`.

src/gui/render_simple.h
FeralChild64: OK, used `reinterpret_cast`, here and in 2 more places.

src/gui/render.cpp
```diff
@@ -140,8 +140,9 @@ static void empty_line_handler(const void*) {}
 static void start_line_handler(const void* s)
 {
 	if (s) {
-		const Bitu* src = (Bitu*)s;
-		Bitu* cache     = (Bitu*)(render.scale.cacheRead);
+		auto src   = static_cast<const Bitu*>(s);
+		auto cache = static_cast<Bitu*>(
```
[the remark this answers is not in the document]
FeralChild64: Done.

src/hardware/serialport/softmodem.cpp
FeralChild64: Renamed the variable. Refactoring would be more complicated (switch/case sets more than just the string variable), I prefer not to introduce such a risk.

kcgen: Makes perfect sense @FeralChild64 :+1:

src/hardware/mixer.cpp
```diff
@@ -1296,29 +1296,30 @@ AudioFrame MixerChannel::ConvertNextFrame(const Type *data, const work_index_t p
 					frame[0] = static_cast<float>(data[pos * 2 + 0]);
 					frame[1] = static_cast<float>(data[pos * 2 + 1]);
 				} else {
+					auto host_pt0 =
+					        (const uint8_t* const)&data[pos * 2 + 0];
+					auto host_pt1 =
+					        (const uint8_t* const)&data[pos * 2 + 1];
```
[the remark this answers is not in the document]
FeralChild64: OK, changed.

src/hardware/mixer.cpp
```diff
@@ -1333,23 +1334,27 @@ AudioFrame MixerChannel::ConvertNextFrame(const Type *data, const work_index_t p
 					        static_cast<int>(data[pos * 2 + 1]) -
 					        offs);
 				} else {
+					auto host_pt0 =
+					        (const uint8_t* const)&data[pos * 2 + 0];
+					auto host_pt1 =
+					        (const uint8_t* const)&data[pos * 2 + 1];
```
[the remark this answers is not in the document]
FeralChild64: OK, changed.

src/hardware/serialport/misc_util.cpp
```diff
@@ -31,6 +31,15 @@
 // Constants
 constexpr int connection_timeout_ms = 5000;
 
+std::string to_string(const SocketType socket_type)
+{
+	switch (socket_type) {
+	case SocketType::Tcp: return "TCP";
+	case SocketType::Enet: return "ENet";
+	default: assert(false); return "(unknown)";
```
johnnovak, line 39: Minor: I usually put "Invalid SocketType value"

src/hardware/serialport/misc_util.h
```diff
@@ -60,7 +60,13 @@
 
 #include "enet/include/enet.h"
 
-enum SocketTypesE { SOCKET_TYPE_TCP = 0, SOCKET_TYPE_ENET, SOCKET_TYPE_COUNT };
+enum class SocketType {
+	Tcp  = 0,
+	Enet = 1,
+	Invalid,
```
johnnovak, line 66: Should we not specify 2 as the value?

src/hardware/serialport/softmodem.cpp
```diff
@@ -494,16 +494,20 @@ void CSerialModem::DoCommand()
 			if (is_next_token("SOCK", scanbuf)) {
 				scanbuf += 4;
 				const uint32_t requested_mode = ScanNumber(scanbuf);
-				if (requested_mode >= SOCKET_TYPE_COUNT) {
+				const auto requestet_type = static_cast<SocketType>(
```
johnnovak, line 500: Typo: requested

src/dos/program_imgmount.cpp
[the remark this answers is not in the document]
johnnovak: Much clearer that way, yes. Zero == success would trip me up every time...

src/hardware/mixer.cpp
```diff
@@ -1328,16 +1332,16 @@ AudioFrame MixerChannel::ConvertNextFrame(const Type *data, const work_index_t p
...
 					        offs);
 				} else {
-					auto host_pt0 =
-					        (const uint8_t* const)&data[pos * 2 + 0];
-					auto host_pt1 =
-					        (const uint8_t* const)&data[pos * 2 + 1];
+					auto host_pt0 = reinterpret_cast<const uint8_t* const>(
+					        data + left_pos);
+					auto host_pt1 = reinterpret_cast<const uint8_t* const>(
+					        data + right_pos);
```
johnnovak, line 1344: Much nicer @FeralChild64 👍🏻

johnnovak (approved):
Great improvement @FeralChild64 , you separated the changes nicely into bisectable commits 🎖️

kcgen (approved):

src/hardware/serialport/misc_util.h
kcgen: Might be OK, but will let @FeralChild64 comment:

An enum _without_ any values tells me we're using the names in a "pure enum" sense, without ever needing to cast-down to numbers.

So when I read this, the first `= 0` and `= 1` tell me we're casting them down to numbers and (maybe) passing them into a C function.

But the bare 'Invalid' tells me we only use it as an enum, perhaps in an assert.

FeralChild64: I've fixed the values of `Tcp` and `Enet` due to the following code in `softmodem.cpp` (if I understand correctly, this is a modem emulation code):
```C++
const uint32_t requested_mode = ScanNumber(scanbuf);
const auto requested_type = static_cast<SocketType>(requested_mode);
```
The `Invalid` - well, I just want it to be the first invalid value. I'll add a comment there.

src/hardware/serialport/softmodem.cpp
FeralChild64: Fixed.

src/hardware/serialport/misc_util.cpp
FeralChild64: OK, changed.

src/hardware/serialport/misc_util.h
johnnovak: > While the bare 'Invalid' tells me we only use it as an enum, perhaps in an assert (so it communicates to me that we never expect to cast `Invalid` down to a number).

Yeah, that's fine and a good way to think about it. But we're comparing values to Invalid in the code, so that's I thought we might want to specify the next value explicitly. But the maybe it's part of the standard that auto-assigned enum values have no holes.

johnnovak:
@FeralChild64 I think you'll need to manually do a `git rebase --autosquash` on your branch before merging because GitHub cannot do that when rebasing and merging AFAIK. The squash option on the GUI just squashes all commits into a single one.

FeralChild64:
@johnnovak I've manually added changes to previous network-related commit.

PR #2777, opened 2023-08-21
Authentic adaptive CRT shaders

johnnovak:
I've decided to write the release notes for this one now and paste it here. I intend to continue this practice, at least for major features.

A bit long, but an important feature like this must be properly explained, plus most of it will be eventually lifted into the manual. Definitely not wasted effort.

---

## Authentic adaptive CRT shaders

We are happy to announce that we've removed all CRT shaders previously bundled with our releases.

Every single one of them.

_What the @^$*%#!!!?? Have these guys gone mad??!_

The reason for removing them is that they got replaced with something far better: a set of no less than **21 CRT shaders** tweaked to perfection to emulate the glorious PC CRT monitors of the past, including authentic recreations of Hercules, CGA, monochrome CGA, composite CGA, EGA, and VGA monitors.

21 new CRT shaders sound a lot—how will someone not intimately familiar with PC CRT monitor technology know which is the best for a particular game? Well, the good news is you don't need to know—DOSBox Staging automatically picks the best shader variant based on the current video mode and viewport resolution. High-DPI and displays from 720p to 4k resolutions are fully supported and absolutely no manual configuration is required!

The following "magic" `glshader` options have been introduced:

- `crt-auto` — This is the new default—an adaptive CRT shader that prioritises developer intent and how people experienced the game at the time of release. The appropriate shader variant is automatically selected based on the graphics standard of the current video mode and the viewport resolution, irrespective of the `machine` setting. This means that even on an emulated VGA card you'll get authentic single-scanned EGA monitor emulation with visible "thick scanlines" in EGA games.

- `crt-auto-machine` — Similar to `crt-auto`, but this picks a fixed CRT monitor appropriate for the video adapter configured via the `machine` setting. So CGA and EGA games on an emulated VGA adapter will appear double-scanned with big chunky pixels, just like on a real VGA monitor.

- `crt-auto-arcade` — This fantasy option does not exist in real life, but it can be a lot of fun! It emulates a 15 kHz arcade or home computer monitor less sharp than a typical PC CRT with thick scanlines in low-resolution modes. Playing VGA DOS ports of Amiga games with this setting will give you that authentic single-scanned Commodore monitor look—now you can experience Amiga games with MT-32 or SC-55 sound from a strange parallel universe! 😎

Adaptive CRT shaders work in tandem with the new integer scaling feature to achieve the most optimal results. The default `auto` setting for `integer_scaling` auto-enables vertical integer scaling for adaptive CRT shaders only. This is to avoid unwanted interference patterns that can result from non-integer vertical scaling ratios.

If you don't like CRT shaders, no problem—you can simply set `glshader` to `sharp` in your config to revert to the previous behaviour. Naturally, you can continue to use regular GLSL shaders like before. And if you'd like to experiment with the deprecated CRT shaders, they're still available as a separate download.

_(This will be a sidenote section)_
> One problem with the previously bundled shaders was that they lacked proper documentation, so it wasn't clear which shader was meant for what monitor resolution (some were tuned for 1080p, some for 4k, etc.) Moreover, most were designed to emulate arcade monitors or TV sets from the 80s and early 90s; these displays had very different characteristics from PC monitors. Because of all this, many people were not getting the results they were hoping for from them and gave up on CRT shaders altogether.

---

## Technical notes

Some further comments (these are not part of the release notes):

- I've spent a lot of time tweaking and optimising these 21 presets. 19 of them use a custom hacked version of `crt-hyllian` (I grafted some other stuff on top of it from other shaders), and two are my own creations that I cobbled together from various bits and pieces (`vga-1080p` and `vga-1080p-fake-double-scan`). I've tweaked all of them for equal brightness across the board, meaning if you switch between the `sharp` shader and any of these CRT shaders, in most cases you will get the same perceptual brightness (or very close). I'm gonna do some further testing and tweaking over the coming week or so, but these are 90% already there.

- The actual shader-switching implementation is as sophisticated and elegant as a sledgehammer. But—just like a sledgehammer—it does its job very effectively. Whenever we're switching to a new shader, 1) the shader gets loaded from disk, 2) it gets compiled, 3) then we reinit the whole renderer—all from the _main thread!_ Yeah, fuck it 😎 It all takes only about 15-20ms though, so this works out fine. This definitely results in a small audio pop about 50% of the time, but very few games do actually switch between resolutions, and even if they do, they don't do it all the time, and even for these handful of games, the mode switching will not necessarily result in shader switching (it depends on the modes in question and the resolution of the viewport). Then for entering/exiting fullscreen or changing the window size, it's a moot point as those operations already introduce a small pause and thus a very audible gap in the audio emulation anyway.

- If someone wants to optimise this and perhaps cache the shader sources, or maybe even store the compiled shaders in a cache, or perhaps do things in a more clever way without doing a full `RENDER_Init` on switching, be my guest. I guarantee you it won't be easy, though, and it most certainly won't be me as I consider this fit for purpose—for now. Also be aware that this will be throw-away work... The next evolution of the adaptive shaders will be implemented as a single mega-shader that we only compile once and then dynamically set various shader parameters at runtime. This mega uber-shader approach is what most modern games do and that will result in effectively zero lag. But a few other things will need to be in place before we can do that, so more on that later (proposal coming at some point, etc.)

## Some screenies

At 4k only (open in new tab, zoom in, etc.)

## VGA

image0008-rendered

image0009-rendered

image0003-rendered

image0006-rendered

## EGA

image0005-rendered

image0007-rendered

## CGA
image0039-rendered

image0038-rendered

## Composite CGA

image0040-rendered

image0043-rendered

image0044-rendered

## Arcade (15 kHz monitor)

Some oldsk00l Amiga pixel gfx 😎 

image

image

image

image

image

Last one...

image

(I just love how it looks 😎 🤘🏻 )


weirddan455:
I just gave this branch a test.  I like the way it looks and it was very fast in switching shaders on window size change for me.  I heard a single audio pop when I first started resizing (maybe it was loading a fresh shader from disk) but then after that no problems even with constant resizing.

The one gripe I have is the display doesn't fill the entire window anymore, like it does on `main`.  Even if I set `glshader` to `none` and even if I set sdl's  `output` to `texture`.  At full screen 1080p I'm left with black bars at the top and bottom of my screen and at some window sizes, it's even more noticable.

crt

src/gui/render.cpp
```diff
@@ -1007,19 +819,37 @@ static void init_render_settings(Section_prop& secprop)
...
+            "                     (default). The appropriate shader variant is automatically\n"
+            "                     selected based the graphics standard of the current video\n"
+            "                     mode and the viewport resolution, irrespective of the\n"
+            "                     'machine' setting. This means that even on an emulated VGA\n"
+            "                     card you'll get authentic single-scanned EGA monitor\n"
+            "                     emulation with visible \"thick scanlines\" in EGA games.\n"
+            "  crt-auto-machine:  Similar to 'crt-auto', but this picks a fixed CRT monitor\n"
+            "                     appropriate for the video adapter configured via the\n"
+            "                     'machine' setting. E.g., CGA and EGA games will appear\n"
+			"                     double-scanned on an emulated VGA adapter.\n"
```
FeralChild64, line 844: Indentation needs to be fixed here.

src/gui/render.cpp
```diff
@@ -1007,19 +819,37 @@ static void init_render_settings(Section_prop& secprop)
...
-	        "  - If you used the normal2x/3x scalers, set a desired 'windowresolution'\n"
-	        "    instead.\n"
+	        "  - If you used the normal2x/3x scalers, set the desired 'windowresolution'\n"
+	        "    or 'viewport_resolution' instead, or consider using 'integer_scaling'.\n"
 	        "  - If you used an advanced scaler, consider one of the 'glshader'\n"
 	        "    options instead.");
 
 #if C_OPENGL
-	string_prop = secprop.Add_path("glshader", always, "default");
+	string_prop = secprop.Add_path("glshader", always, "crt-auto");
```
FeralChild64, line 828: Previously an empty setting:

```ini
glshader =
```

was behaving like:

```ini
glshader = none
```

Now it prints out a log like:

```
2023-08-23 17:32:46.409 | SDL:OPENGL: Failed to compile shader!
```

and doesn't scale the image at all. I suggest to bring back the previous behavior.

FeralChild64:
@weirddan455 You have to set `integer_scaling = off` for an old behavior. And, frankly, I don't think enabling integer scaling is a good idea (at least not until 8K monitors are common) - it can create a really large black bars around the screen.

kcgen:
Finally correctly-applied CRT-experience for all the video modes, that "does the right thing" without requiring detailed fiddling behind the scene by the user to dial in perfect window resolution, view-port resolution, etc.

Amazing work, @johnnovak !

I haven't been able to get enough time these days though, and probably won't have review cycles till the weekend. Will try then; but don't hold up for me if it's all reviewed before hand!

FeralChild64:
I've played a little with it and I'm really impressed. I'll probably start using shaders - but for now I'll probably only enable them per-game, as IMHO some games look best with _15kHz_ shaders, other are better with _machine_ or full _auto_ one, and for text mode or 1024x768 resolution I prefer no shaders at all.

I'll look more into the code, but not today - this PR is huge, but really important to recreate the original feeling of many games.

src/gui/render.cpp
johnnovak: Good catch, that wasn't intentional.

johnnovak:
Thanks for the appreciative words guys, I'm really glad you like it 😎

Yeah, the reason why we thought with @kcgen this should be enabled by default was to give people a chance to experience DOS games how they were seen by the developers and the gamers back in the day, and in our view that's how they are meant to be experienced. Many people (including past me) had bad experiences with heavy-handed, inappropriate, or plain bad shaders, therefore perhaps they wouldn't even consider using them and would just scroll through the shader announcement in the release notes. @FeralChild64 is a good example; as far as I'm aware he wasn't too keen on shaders before but now seems to be a "convert" 😄 One cannot hope for a better endorsement! So yeah, we just don't want people rob themselves of the experience out of prejudice; we want to give them the best chance to give themselves the chance to join in on the fun... Uh, this gets complicated fast but you get the picture 😅 (and I should really cut back on the puns...)

Now, in general, CRT shaders need integer scaling ratios in the vertical direction. The denser the scanlines are, the more crucial this is, otherwise you'll get moire (interference) patterns that are _really_ ugly. The more pixels per "emulated scanline", the more leeway you have to diverge from this. You can use non-integer ratios to stretch 200-line content to fullscreen even on 1080p, and you're generally fine to stretch 400-line content to fullscreen on 4k, but you might see a _little_ moire on solidly filled areas with certain colour intensities.

Try streching  640x480, 640x350, or 640x400 content to fullscreen on 1080p--it will look like shit. You might get lucky with some busy darker images, but keep testing with a variety of content and you'll hit the issue soon with images that contain larger solidly filled areas, or any image that uses lighter shades.

And remember, 320x200 is 640x400, 320x240 is 640x480 on VGA, and so on. Low-res content is hardware 2x scaled by VGA, we're emulating that correctly now. Scaling 320x200 to just 800 vertical lines would be a really hard sell for most people, hence I invented my fake overscan shader specifically for low-res VGA games on 1080p, and with that we can 5x vertical scale to 1000 pixels and the results are sorta ok.

I've manually sized the viewport with `viewport_resolution` for years now for each game to get the correct vertical integer scaling ratios, but it's so fiddly, and in some games that switch resolutions there is usually no single good value to use, and the text mode will most likely look like shit when launching the game, etc.

So, I really don't see other good general, set-and-forget solutions for CRT shaders than to default to vertical integer scaling when using the adaptive shader modes.

However, that's a fair point about the `sharp` and `none` shaders. Especially for the `none` shader, integer scaling doesn't make too much sense. I was thinking of introducing an `auto` setting for `integer_scaling`; this would be the new default and it would only enable vertical integer scaling for the adaptive shaders.

How about that?

kcgen:
Really like the ideal of an `auto` setting that applies integer scaling when the shader needs it. 

Because DOS (and win 3.1) games were designed (and played) on CRTs, I think it's the truest representation of the game and really brings the nostalgia factor. Because of this, I think `crt-auto` and `integer-scaling=auto` should be the out of the box default (of course, with users being able to adjust if they prefer after first launch). 

I also 100% agree that the earlier scalers (SuperEagle-2x-Sai, et.al) were basically anti-nostalgia and sent people running to "bare pixels, please and thank you"; and finally these auto-crt shaders do a 180 and massively _enhance_ the nostalgia factor.

johnnovak:
> or now I'll probably only enable them per-game, as IMHO some games look best with _15kHz_ shaders, other are better with _machine_ or full _auto_ one, and for text mode or 1024x768 resolution I prefer no shaders at all.

The best thing, you can switch between them at runtime with `glshader`—it's been extensively tested 😎  (I know you like to use DOSBox as a "machine emulator", running it like a real system and switching configs in batch files when necessary).

GranMinigun:
Switching from 320x400 to 320x200 seems to break scaling calculations.

These alternating scanlines initially tripped me over. Looked like uneven pixel scaling. After closer inspection - nope, exactly five pixels per line. 1080p screen.

Screenshot_20230824_073859

EGA:

Screenshot_20230824_080846

And here's a (rough) comparison between regular and arcade CRT shaders:

Screenshot_20230824_082147

Screenshot_20230824_082403

kcgen:
That Mechwarrior intro screen. Now see how the artist saw it on their own circa-80s EGA CRT (and it's spectacular; 😍 ) wow.

johnnovak:
> Switching from 320x400 to 320x200 seems to break scaling calculations.
> 
> These alternating scanlines initially tripped me over. Looked like uneven pixel scaling. After closer inspection - nope, exactly five pixels per line. 1080p screen.

No, it's working _exactly_ as designed and as I expect it to work. It's all on purpose.

You're using a 1080p monitor in fullscreen. So:

Case 1 — 320x400 content gets 2x vertical upscaled to 800 lines with the `crt/vga-1080p` shader. We absolutely cannot do better in this case on 1080p with vertical integer scaling as those 400 lines could all contain different content for all we know. That's a "true 400 line" mode, end of story.

Case 2 -  320x200 content gets 5x (!) vertical upscaled to 1333x1000 with the `crt/vga-1080-fake-double-scan` shader. How so? That's a concession I'm making for people on 1080p screens. 320x200 is technically double scanned to 640x400 on VGA adapters, so if we treated it as such, then we could only upscale to 800 lines, just like in case 1. But, every vertical line is actually line-doubled in 320x200 VGA mode, so we're cheating a bit. We're just applying 5x vertical scaling to the raw 320x200 non-double-scanned content, then we're cheerfully dimming every second scanline a bit, hoping no one will notice... Except @GranMinigun , of course 😎 That will give you the perception of that typical "double-scanned texture" for 320x200 content, but yeah, technically inaccurate as some pixels will contain 2 "scanlines", some 3. But that's the best we can do! The other alternative is to skip the CRT shader altogether and just use the sharp scaler, but I like the fake scanlined look.

Zoomed in portion of your screenshot; notice some pixels have 2, and some have 3 emulated scanlines:

image

I'm explaining this actually in the code too, so maybe check out the PR :P

GranMinigun:
You didn't get it. If I go fullscreen on my machine with 1080p display to fix the window size, then in Quake select 320x400 mode, then switch back to 320x200, the output image is smaller than it initially was, and it doesn't scale up to full 5 pixels per line until I resize the window to trigger scaling recalculation.

johnnovak:
> You didn't get it.

You did not describe this situation in detail, mate 😄 So there was not much to get 😄 

> If I go fullscreen on my machine with 1080p display to fix the window size, then in Quake select 320x400 mode, then switch back to 320x200, the output image is smaller than it initially was, and it doesn't scale up to full 5 pixels per line until I resize the window to trigger scaling recalculation.

That's a better description but still somewhat confusing. "fix the window size" — what does that mean? Are you now in fullscreen or in windowed mode, it's unclear.

Anyway, couldn't reproduce on macOS using a 1080 pixel high window and switching between 320x200 and 320x400. Will try on my Windows box later that has a 1080p screen attached and switch between windowed/fullscreen etc. In the meantime, maybe you could provide the _exact_ step by step instructions for reproducing it if you believe it's an issue.

GranMinigun:
1. Set the window size to 1334x1000, or go fullscreen on a 1080p display;
2. Launch Quake;
3. Change the video mode to 320x200, note how much space the image takes up;
4. Change the video mode to 320x400;
5. Change the video mode back to 320x200, the image will be scaled only 4 times vertically.

src/gui/render.cpp
johnnovak: Fixed.

src/gui/render.cpp
johnnovak: Fixed.

johnnovak:
I've addressed all comments, fixed the issue reported by @FeralChild64, and have added `auto` integer scaling mode. I'll look into the issue reported by @GranMinigun tomorrow. I've also updated the release notes draft:

> Adaptive CRT shaders work in tandem with the new integer scaling feature to achieve the most optimal results. The default auto setting for integer_scaling auto-enables vertical integer scaling for adaptive CRT shaders only. This is to avoid unwanted interference patterns that can result from non-integer vertical scaling ratios.

FeralChild64:
@kcgen Well, I don’t like _sharp_ at all - gives too ”rough” looking picture for my liking.

kcgen:
> @kcgen Well, I don’t like _sharp_ at all - gives too ”rough” looking picture for my liking.

Yeah: given this is `crt-auto`, we should still try to achieve the slight softness that CRTs really did - especially when viewed at a typical working distance.

So the fallback (could?) be `none`, especially when the DOS resolutions are too high (like 800x600+).  I also use `none` in 800x600+ modes, and it looks very close to how I remember (like your first win 3.1 screenshot, as I'm not inspecting pixels with a magnifying glass).

Maybe `sharp` still makes sense as a fallback when DOS resolutions are low (like 320x200 on an embedded 800x600 screen).

weirddan455:
>  In 800x600+ modes, none still looks pretty close to a CRT (like your first win 3.1 screenshot, @FeralChild64 ), when sitting at a comfortable distance (as I'm not inspecting pixels with a magnifying glass).

I don't know about that.  It looks more like an LCD screen to me, which is probably best for that use case.  I remember the first thing I noticed when I switched from a CRT to an LCD monitor was how much better text looked.

CRT may be better for low res DOS games but if you're looking at a Windows GUI, I'd prefer no shaders.

kcgen:
> CRT may be better for low res DOS games but if you're looking at a Windows GUI, I'd prefer no shaders.

@weirddan455 , @FeralChild64 : yes - I think we're all on the same page once resolutions get high enough (>= 800x600), especially-so on sub-4k screens. Will see what @johnnovak sees based on testing with his own CRTs vs. modern dosbox hosts.

weirddan455:
@kcgen  I think the text still looks bad on those screenshots personally (compared to the no shader screenshot @FeralChild64 showed).

Not to say anything bad about this PR though.  I'm assuming most people are playing DOS games and we should optimize our defaults for that case (and I think these sharers look great in-game).  If you're spending most of your time looking at a Windows GUI, you can change to no shader if you like.

johnnovak:
Whoa, I wake up this morning and I see a detailed discussion forming about the new shaders already! I'm really glad you guys are giving this thing a good spin, and I'm especially pleased that @FeralChild64 thinks the results are great with many classic DOS games.

Here are my extended thoughts about the points you brought up (block of text incoming, grab some drinks  🍵 ☕🍻 😄  ):

- First of all, all Windows 3.11 screenshots that use the shaders posted by both @FeralChild64 and @kcgen look completely fine to me. The comparison of the CRT shaded output to the `none` output is unfair and misleading—detailed explanation of that below.

- Secondly, using the `none` "shader" anywhere in the adaptive CRT shaders (also known as "gamma-incorrect bilinear interpolation"; see here)—well, I'm quite vehemently against that as bilinear interpolation is just terrible. What we _might_ consider is falling back to @GranMiniGun's [Catmull-Rom interpolation shader](https://github.com/dosbox-staging/dosbox-staging/blob/main/contrib/resources/glshaders/interpolation/catmull-rom.glsl) instead of the `sharp` shader, that would be a much better option. I could do that with good conscience, but this will need further testing & experimentation (which I'm intending to do).

Before tackling the details, let me set one thing straight: 99% of our users use DOSBox to play games (and that's conservative; I think you could slap a few more decimal 9s at the end). Even if there was a problem with the way the Win 3.11 desktop looks with the default 16-colour palette (there isn't), most would use Win 3.11 as a game launcher, at best.

The good thing is for the remaining <1% of us who like to use DOSBox as a "whole computer emulator" to re-live the 80s/90s, you can still set any `glshader` setting manually. That markedly belongs to the advanced power user category, though. 99.9%+ of people will just enjoy their 320x200 to 640x480 DOS and Win 3.11 games with the adaptive CRT settings.


## Shader selection criteria

Shader selection is based on the number of pixels per "emulated scanline"; in other words, the viewport height / DOS video mode height ratio. In general, you need 3.0 native pixels per emulated scanline for good results, 5.0 for very good, and 8.0+ for absolutely stellar outcomes.

That's the only practical way for a fully dynamic setup, and the best way if you think about it. The pixels per scanline metric is all that matters. E.g., windowed mode on a 1440p screen with a window a bit smaller than fullscreen will call for a 1080p preset, and so on. If you start thinking about actual native monitor resolutions (at fullscreen) and DOS resolutions—yeah, don't even go there. It will get complicated very fast and it's unmanageable to handle it like that.

Please read the second half of `shader_manager.cpp` for details. You'll find that I'm doing some trickery to get the most out of the very restricted 1080p viewport size scenario where we fall below 3.0 ratios. But generally, when we can't maintain a 3.0 vertical ratio (or 2.0 for the special 1080p VGA stuff), we fall back to `sharp`. With the `integer_scaling = auto` change implemented, if that happens we also turn integer scaling off, meaning that the result will be scaled to the full extents of the viewport.


## Windows 3.11 desktop

Okay, this "issue" is coming back to haunt me as a restless spirit... @FeralChild64 and I have [some history on the matter](https://github.com/dosbox-staging/dosbox-staging/pull/2665#issuecomment-1624502438).

First of all, the Win 3.11 screenshot with the `none` shader that @FeralChild64 posted is not a correct or authentic representation of how this would look on a CRT monitor. It uses bilinear interpolation with gamma-incorrect blending, which makes 100% black text on a 100% white background look fatter than it would on a real CRT. Compare the `sharp` and `none` output of a 640x480 or 800x600 Windows 3.11 desktop (these are square pixel aspect ratio modes, you'll get 100% sharp pixels with the `sharp` shader). The hairline letters will look fatter with the `none` shader.

Now, I did correct the `none` shader to use gamma-correct interpolation, but @FeralChild64 protested because that made the _other_ extreme edge case, namely white text on black background fatter. Due to his objection, I conceded to revert the `none` shader to gamma-incorrect interpolation because I consider it a legacy thing. I thought, okay, let people used to the old gamma-incorrect bilinear shader keep using it... until they see the light and move on to something better 😎 

So, to recap, this is how bilinear interpolation behaves in the extreme edge case scenarios; as you see, you just can't win with this thing:

- **With gamma-correct bilinear interpolation:**
  - White text on black background looks a bit thin
  - Black text on white background looks a bit fat

- **With gamma-incorrect bilinear interpolation (flipped results):**
  - White text on black background looks a bit far
  - Black text on white background looks a bit thin

Neither are "accurate" representations of how a cathode-ray tube monitor that fires particles at high velocities at a thin phosphor layer to momentarily excite them creates its picture... Bilinear interpolation is a very simple _mathematical process_ with no parameters to tweak, and this is what you get, end of story. It has no basis in reality as far as the workings of a CRT monitor are concerned.

CRT shaders, on the other hand, while still mathematical processes, expose a large number of parameters to tweak to replicate a quite convincing emulated CRT experience. Because of these parameters, the possible solution space is very large; tweaking and fine-tuning these parameters to achieve pleasing results is more art than science. Maths alone won't get you there—unless you'd emulate the working of a CRT on the atomic level, which we're clearly not doing...

If we're starting to discuss what a CRT does and doesn't do, let's talk about the "dilation" of the black thin lines in the letters in the bilinear output @FeralChild64 posted compared to the `sharp` output. How could that happen in real life? How could the black lines "dilate"? Well, they couldn't. In real life, the brighter a "pixel" is, the larger it becomes (only slightly on high-resolution (S)VGA monitors, but a lot more on CGA/EGA). This brightness-variable dilation is called "bloom" in CRT shader parlance. The result is that bright pixels "eat into" neighbouring black pixels a little bit. In non-edge case scenarios this creates a subtle, pleasant blending of neighbouring pixels; in the extreme case of white hairline text on black, it makes the text look a bit fatter; in the opposite extreme scenario of black text on white, black hairline text will appear a bit thinner because the surrounding white areas will bloom into it.

100% black letters blooming into neighbouring 100% white areas, that's a physical impossibility, I hope that's clear to everybody at this point. That cannot happen on any CRT, ever. The closest you'd get on very high-quality late 2000 CRTs is a very close approximation of the `sharp` shader look, but never the `none` look with incorrect gamma that has the "dilated black pixels" effect.

I'm only explaining all this in painstaking detail to drive it home that the look of bilinear interpolation is by no means a standard we should measure any CRT shader against. It's a non-standard; it's a simple mathematical process that blurs the image in a rather bad way.

With all due respect @FeralChild64, and imagine I'm telling you this in a calm, relaxed, and friendly voice: you have been staring at the bilinear interpolated output with incorrect gamma for so long that it has become your "normal", expected experience. But this is no different from people being used to playing all 320x200 games with square pixels and being attached to it. It has no basis in actual reality.

See my earlier comment for further explanation with images: https://github.com/dosbox-staging/dosbox-staging/pull/2665#issuecomment-1624502438

For more "proof" you'll need to do further experiments yourself—I've exhausted what I can say on this topic. But I'm 100% convinced what I wrote is correct, and I have no less than 9 CRT monitors in my possession to back my claims up.


## Detailed A/B comparisons

> With CRT shader we are getting too low contrast and washed-out colors in icons like PIF Editor or Control Panel.

Rest assured, this pack of 21 shaders is a result of more than a year's worth of _obsessive_ shader tweaking. The thing is, you can always single out specific details on specific images and certain edge cases, then make claims about the CRT shader being this or that. This exercise is, however, rather unfruitful.

The objective here was to come up with a limited number of shaders that handle a wide variety of content equally well (and I still ended up with 21 shaders, all manually fine-tuned; a lot more than I originally anticipated). I could always take a single shader and create a more optimised version of it for a given static screenshot, displayed at a specific resolution on a specific flat-screen monitor that uses specific brightness & contrast settings. That's the easy part. The _hard_ part (and I mean, _really hard_) is creating a _single_ VGA preset that works equally well with all VGA games, another single preset for all EGA, and so on. Frankly, I wouldn't start doing "optimised" per-game preset even for myself... that's a road to madness. I don't want to be dismissive here, but you might be vastly underestimating the difficulty of achieving such universal shaders. It required 100+ hours of fine-tuning and tweaking, and my honest opinion is I've pretty much maxed out the sweet spot settings for each one.

So, if you guys notice _systemic_ issues that you can back up with screenshots of a variety of content (say, at least 5 screenshots), I'm very happy to take a look and investigate. But, to be blunt, I would be extremely surprised if you found could. Single one-off cases, especially when they're about extreme edge cases; I'm not going to waste my time on those. I've been there, and I've been going around in circles for many iterations until I've arrived at these "sweet-spot" presets that work great for 90%+ of all scenarios.

---

Okay, that was long, and make no mistake, I'm always open to constructive criticism. Like how I introduced the `auto` integer scaling mode yesterday based on your feedback. That was a great idea, thanks for it! So was trying Catmull-Rom interpolation as a fallback for the adaptive shaders. Just the reasoning for the Win 3.11 stuff went a bit sideways because it rested on incorrect assumptions—given that I have significant experience in this whole CRT shading topic, I had to steer the discussion back to the correct path 🙂

johnnovak:
> 1. Set the window size to 1334x1000, or go fullscreen on a 1080p display;
> 
>     2. Launch Quake;
> 
>     3. Change the video mode to 320x200, note how much space the image takes up;
> 
>     4. Change the video mode to 320x400;
> 
>     5. Change the video mode back to 320x200, the image will be scaled only 4 times vertically.

I am able to reproduce, that's great feedback, cheers @GranMinigun 👍🏻

johnnovak:
Okay, I did give the Catmull-Rom shader a good workout, and it almost passed the test... but in the end, it just didn't. It was worth a shot.

Here's why: Quake bench results using my 1080p monitor and the default window size. Open images in separate tabs, then compare them at 100% zoom level.

## 320x200

The fake double-scan shader gets picked up, life is good 😎 

image


## 360x240 -- sharp

Not enough vertical res to use the fake double-scan shader, to we're reverting to `sharp`. This is accurate and as good as it gets; low-res VGA pixels should look like little rectangles.

image


## 360x240 -- catmull-rom

Uh.... blur city. This is _most definitely_ not how low-res VGA should look like. Where are the blocky pixels? You won't see this level of blurriness on any VGA monitor in existence, not even the shittiest ones.

image

GranMinigun:
It's intended to be a slightly better bilinear, not slightly worse sharp, for SVGA content.

johnnovak:
> It's intended to be a slightly better bilinear, not slightly worse sharp, for SVGA content.

Yeah, I guess my problem is even for say 640x480 or 800x600, if your viewport is only say 1.5x larger, then you just get the blurriness. Then if you keep enlarging your viewport and suddenly when the 2x integer scaling kicks in (or 3x if we're not on 1080p), the output is markedly sharper again. This transitioning between sharp/blurry is very confusing at the least, so I'm not in favour.

But it's good we have it, so people can always use it as a fixed shader as an alternative option.

johnnovak:
Ok, I think I fixed the 320x200 <-> 320x400 switching issue reported by @GranMinigun, but will only test on Windows / 1080p tomorrow. It was an interesting bug, nothing to do with shader switching, actually, but going much deeper (related to the double-scanning & pixel-doubling enhancements).

FeralChild64:
@johnnovak 

I'm fully aware that bi-linear filtering is not a realistic CRT simulation, and that it makes some fonts to appear bolder than they look in reality. The problem is, that the "natural anti-aliasing" effect here is almost completely absent, the font is, iMHO, too sharp.

> 100% black letters blooming into neighbouring 100% white areas, that's a physical impossibility, I hope that's clear to everybody at this point

And this is something I don't really agree. White pixel does not emit infinitely-bright light - so if a black pixel gets a "boost" from neighbor white pixels due to blooming effect, same should happen to any other pixel, including white ones! Let's do some pixel-peeping:

Example

Pixel I have marked with green color is far from real black - that's OK, as it is almost completely surrounded by white pixels. But if the pixel is already white, this effect is almost completely gone; the one I have marked in red is almost surrounded by black pixels, it gets a luminosity boost via blooming from just one neighbor, yet it is still #FFFFFF in the center.

johnnovak:
> Let's do some pixel-peeping:

How about let's not do that? 😅 

@FeralChild64, please read the "Detailed A/B comparisons" of my attached booklet again carefully—I'm not going there, man 😎 

These settings have been optimised for people playing _games_ on DOSBox. That's 99.9(9)% of our users. 

It's a game of compromise. Start creating your own presets and you'll realise how much compromise is involved... after you've sunk 5-10-20-50+ hours into it! You can't please _absolutely everybody_ with such universal presets, but we can please _most people_ who play _games_ in DOSBox.

Pleasing 99% of our userbase versus 3 guys who stare at hairline 100% black fonts on 100% white background (the absolutely **_worst_** possible edge case there is) at 800% magnification and the "PIF Editor" icon... I'm sorry, are we even having this discussion? 😅 

Special people with special requirements will always need special configs, there's no way around it. That's the Staging philosophy with which I 100% agree. Let's create a great out-of-the-box experience for 90%+ of our users, and provide people with special needs all the levers, buttons, and switches to customise the behaviour (to a reasonable degree).

As I see it, you have all you need to customise your Win 3.11 experience to your liking. I _might_ even create a _"Special FeralChild Edition Window 3.11 VGA Preset"_ just for you if you ask me nicely that has less bloom 😎... but that will make every other type of content darker and/or will result in a loss of contrast. So don't come back complaining that now all games are 20-30% darker... 😅 Like I said, compromises.

johnnovak:
@GranMinigun The 320x200 <-> 320x400 issue is fixed, confirmed.

This is all ready for code review, guys.

johnnovak:
So, I've been playing a bit more with the Catmull Rom interpolator for SVGA content only as I really _want_ to like it, but I think it's just too blurry. 

First one is Catmull Rom, second is `sharp`, and the third one the CRT shader once we're hitting the 3x verticaling scaling ratio.

My observations:

- Catmull Rom looks just too blurry to me, that's not how CRTs look (unless they're broken).

- `sharp` is a bit too sharp, no doubt. Ideally we'd need something between Catmull Rom and sharp, just in the middle, because I like it how it rounds off the edges, but as it is now, it's overdoing it and there's too much blur and no pixel definition.

- Then the proper CRT shaded results are the best, just the right balance between a bit of blur and sharpness and pixel definition, just like on a real CRT.

So, overall I'm still leaning towards just falling back to sharp 🤷🏻 

image

image


image

GranMinigun:
Can always play with sinc or jinc interpolation, and tune parameters to your liking. But none of these filters are intended to replicate characteristics of a CRT display, of course (that includes Catmull-Rom spline). You'll also need to deal with ringing artefacts, and convert to and back from so-called Sigmoid space (tuned logistic function) for better results. Bye-bye performance on integrated graphics.

Maybe a different cubic spline will satisfy your needs, like Hermite. Or maybe you could pick a bilinear interpolation shader with controlled sharpness, I think I saw that in Slang shaders repo.

Sticking to sharp is completely fine though.

johnnovak:
> Can always play with sinc or jinc interpolation, and tune parameters to your liking. But none of these filters are intended to replicate characteristics of a CRT display, of course (that includes Catmull-Rom spline). You'll also need to deal with ringing artefacts, and convert to and back from so-called Sigmoid space (tuned logistic function) for better results. Bye-bye performance on integrated graphics.
> 
> Maybe a different cubic spline will satisfy your needs, like Hermite. Or maybe you could pick a bilinear interpolation shader with controlled sharpness, I think I saw that in Slang shaders repo.
> 
> Sticking to sharp is completely fine though.

Yeah, all our options are limited by practical reality, that's for sure.

I'm currently replaying Syberia on an actual 17" CRT, it's an 800x600 rendered fixed-background game, and I can clearly see the pixels if I lean closer. None of this level of blurriness is present. I just fired up Gothic and Morrowing too 5 minutes ago and checked them out in 640x480, 800x600 and 1024x768; all are crisp, zero blur.

So I'll just stick with sharp for now, while not completely ideal, that's closer to the actual high-quality SVGA CRT experience than any interpolating filter.

kcgen:
> So I'll just stick with sharp for now, while not completely ideal, that's closer to the actual high-quality SVGA CRT experience than any interpolating filter.

I also remember Win 3.1 at 800x600+ being sharper than our bilinear smoothed results:

2023-08-25_07-20-none,bilinear

---

I guess w/ `sharp`, we can disable integer scaling too (alternatively, if we keep integer scaling, `sharp` performs like switching OpenGL to nearest neighbour (lower screenshot):

2023-08-25_07-21-opengl-sharp

2023-08-25_07-22-openglnb-none

---

> Maybe a different cubic spline will satisfy your needs, like Hermite. Or maybe you could pick a bilinear interpolation shader with controlled sharpness, I think I saw that in Slang shaders repo.

That's the best thing about the `auto` settings: we can keep fine tuning corner cases like this later on.

johnnovak:
> I guess w/ `sharp`, we can disable integer scaling too (alternatively, if we keep integer scaling, `sharp` performs like switching OpenGL to nearest neighbour (lower screenshot):

Yep, I'm already doing that, I implemented thst on Friday, check out the latest version.

If the minimum vertical scaling factor requirement isn't met for any adaptive CRT shader variant, in `auto` integer scaling mode (the new default) we're not doing integer scaling anymore when falling back to `sharp`.

So at 1080p fullscreen, 640x480 qualifies for `crt/vga-1080p`, but 800x600 and above gets stretched to fullscreen with `sharp`.

That's preferable as 800x60p without scaling is already too small on 1080p.

FeralChild64:
@johnnovak 

> As I see it, you have all you need to customise your Win 3.11 experience to your liking. I _might_ even create a _"Special FeralChild Edition Window 3.11 VGA Preset"_ just for you if you ask me nicely that has less bloom 😎... but that will make every other type of content darker and/or will result in a loss of contrast. So don't come back complaining that now all games are 20-30% darker... 😅 Like I said, compromises.

Damn - black-on-white text is not only Windows 3.11 use case, many DOS games displays something like this from time to time (Civilization, Transport Tycoon, Alone in The Dark series if I remember correctly, ...). I guess it will be safer for me to stick to billinear for now.

johnnovak:
> Damn - black-on-white text is not only Windows 3.11 use case, many DOS games displays something like this from time to time (Civilization, Transport Tycoon, Alone in The Dark series if I remember correctly, ...). I guess it will be safer for me to stick to billinear for now.

Rest assured, I've tested this with many DOS games that feature black text on white background. Those are all early Sierra adventures and the Legend Entertainment adventure games, among others.

Black text on white background looks fine in these games; I know because I did spend quite some time coming up the sweet spot between maintaining equal brightness but now blowing out the highlights too much. I love these games and I want the to look good.

Your problem with the hairline fonts in Win 3.11 is a bit special; I admit a _bit_ less bloom would look subjectively better there, but that could only be done to compromise brightness and contrast everywhere else. Hence I opted to treat the Win 3.x desktop situation with lower priority, to make games look bettet.

kcgen:
@FeralChild64 ,

> Civilization, Transport Tycoon, Alone in The Dark series if I remember correctly

With these all being 320x200, they will easily qualify for full-up CRT shaders on 1080p and greater host screens; so the `sharp` fallback won't kick in.

johnnovak:
Good tip on checking out the RA shader repo @GranMinigun; I found TheMaister's original sharp bilinear code where the sharpness factor is a parameter (in our version, it's always auto-determined to the input/ouput size at a baked-in sharpness).

I'll port this back to our sharp shader and play around with it.
https://github.com/libretro/slang-shaders/blob/master/pixel-art-scaling/shaders/sharp-bilinear.slang

This is as lightweight as it gets, even for Pi users. There are other adjustable sharpnes pixel scaler shaders there, but most rely on explicit sampling and linearisation done in the code, so those are a lot slower.

johnnovak:
> Damn - black-on-white text is not only Windows 3.11 use case, many DOS games displays something like this from time to time (Civilization, Transport Tycoon, Alone in The Dark series if I remember correctly, ...). I guess it will be safer for me to stick to billinear for now.

@FeralChild64 It would be regrettable if our discussion left you with a bad taste in your mouth—so much that it would dissuade you from even using these shaders with games. It's just I found it very frustrating that we got stuck on one minor edge-case for two three days straight as I've spent a long time on these shaders to make them look as good as I can.

But when we're talking about _games_, I take your concerns very seriously. Like I said, I've spent considerable time tweaking these so even 100% dark text on pure white background looks good and legible in actual games. To ease your concerns, and to document things, here are a couple of screenshots at every resolution that exercise these _edge cases_. Again, these are worst-case scenarios and my opinion is that the shaders handle them admirably well. This does not mean I won't tweak them further in the future (maybe even next week), but they're already quite good I think.

So let's see the game stress tests!

# Space Quest III (1) [320x200]

All EGA Sierra adventures have black text on white background, so this has to look good.

## 4k

sc3_1-4k

## 1440p

sc3_1-1440p

## 1080p

sc3_1-1080p

## 720p

sc3_1-720p


# Space Quest III (2) [320x200]

1-pixel wide font.

## 4k

sc3_2-4k

## 1440p

sc3_2-1440p

## 1080p

sc3_2-1080p

## 720p

sc3_2-720p


# Spellcasting 201 (640x350)

## 4k

sc201-4k

## 1440p

sc201-1440p

## 1080p

sc201-1080p


# Timequest

Hairline font stress test — looks fine to me, even at 720p the results are very legible. These black hairlines get a little bit think even on a real 14-15" monitor!

## 4k

tq-4k

## 1440p

tq-1400p

## 1080p

tq-1080p


# Transport Tycoon Deluxe [640x400]

Another hairline font test. Legible and good-looking across the board.

## 4k fullscreen

tt-4k-fullscreen

## 4k windowed

tt-4k-windowed

## 1440p

tt-1440p

## 1080p

tt-1080p


# Sim City [640x480]

Black and white stress test. Passing with flying colours (well...) as I see it 😎 

## 4k

sim-4k

## 1440p

sim-1440p

## 1080p

sim-1080p

johnnovak:
...aaand by doing this comparison I've realised that the brightness of the 1440p VGA one could be taken back a little back; it's a tad overcooked 😎 Check out the Transport Tycoon screenshots. Maybe that's the one you complained about. Like I said, there is always room for improvement.

johnnovak:
Ok, pulled back the bloom/brightness of the `vga-1440p` preset a little bit. It's not an earth-shattering difference, but it's an improvement. 

## Before

vga-1440p-before

## After 

vga-1440p-after

FeralChild64:
@johnnovak Thanks, this change is indeed an improvement. I'm under the impression that it helps some text mode apps (like DOS Navigator) too - but I don't have an old screenshot to compare.

johnnovak:
> @johnnovak Thanks, this change is indeed an improvement. I'm under the impression that it helps some text mode apps (like DOS Navigator) too - but I don't have an old screenshot to compare.

Yeah, I think so. Hardcore shader enthusiasts buy some extremely bright monitor that they use at 100% brightness, then you can use copious amounts of phosphor mask and scanline effects and still get the full contrast range in the resulting CRT shaded image. But you can only do that in fullscreen, otherwise the rest of the desktop would look like a nuclear explosion...

These are settings that everybody can use on any display, without having to fiddle with their monitor controls just for DOSBox, either windowed or fullscreen, so it becomes a game of how far can you push things to achieve equal brightness compared to the raw output while not blowing out highlights and not compressing the dynamic range too much, and so on. It's tricky to get it right, and I found that test grayscale ramps and such don't help much; best way is to continuously retest using real-world game captures. 

Like I said, ultimately we'll have emulated monitor controls, so you could even dial the brightness back a bit for Win 3.x usage, then increase it for games (in reality, I remember having to fiddle with analog monitor controls a bit per-game on the Amiga and the PC).

Ok, so can we focus on the code now? @FeralChild64 @GranMinigun @weirddan455 @kcgen @kklobe ? I'd like to get this merged this week and show it to eXo.

weirddan455:
I took a look at the code but I just feel really out of my depth trying to review something like this.  You know a lot more than I do about the intricacies of CRT monitors and shaders.  I haven't even really looked at the DOSBox's renderer before.

It checks all my boxes in any case:

Builds: :heavy_check_mark: 
Doesn't crash: :heavy_check_mark: 
Looks good: :heavy_check_mark: 
Ship it! :rocket:

johnnovak:
> I took a look at the code but I just feel really out of my depth trying to review something like this. You know a lot more than I do about the intricacies of CRT monitors and shaders. I haven't even really looked at the DOSBox's renderer before.
> 
> It checks all my boxes in any case:
> 
> Builds: ✔️ Doesn't crash: ✔️ Looks good: ✔️ Ship it! 🚀

Cheers. If you have purely code-related remarks (at the micro level), feel free to leave comments 😄

GranMinigun (approved):
Testing is all green. No comments code-wise.

GranMinigun:
Actually, no, I do have one comment. If the shaders do not rely on `GL_CLAMP_TO_BORDER` vs `GL_CLAMP_TO_EDGE` difference, and behave correctly with NPOT texture size, then please add them to the list in `is_shader_flexible()`.

johnnovak:
> Actually, no, I do have one comment. If the shaders do not rely on `GL_CLAMP_TO_BORDER` vs `GL_CLAMP_TO_EDGE` difference, and behave correctly with NPOT texture size, then please add them to the list in `is_shader_flexible()`.

I don't think they care about such things, but I thought `is_shader_flexible` has something to do with window resizing?  Or maybe that was historical and we removed that restriction.

Anyway, I guess I can test by adding them to the flexible shader list.

By the way, why is NPOT desirable at all in our case? I'm very far from being an expert on such matters, I just would like to understand I guess. This is the first answer I found, and according to this there are only drawbacks in our case?

https://gamedev.stackexchange.com/a/7949

GranMinigun:
Less VRAM usage, and no black halo on the edges of the output image. No drawbacks, because we don't need mipmaps.

johnnovak:
> Less VRAM usage, and no black halo on the edges of the output image. No drawbacks, because we don't need mipmaps.

Ok, I guess the 20-30% speed penalty when using NPOT put me off the most. But maybe that was only historically the case?

The Hyllian shader I'm using just samples a 2x4 kernel around the current pixel and that's it. I guess we'd need to oversize our texture by a slight border to make it technically correct... but I'll just test it how it behaves.

GranMinigun:
Yeah, if it was the case, then that's history. Nowadays GPUs have to deal with NPOT textures all the time (GUIs and such). There are also rectangular textures which do not support mipmaps, so can always have arbitrary dimensions, but they're likely implemented the same way as regular textures behind the drivers.

kcgen:
With default settings on Linux, GCC thread sanitizer catches a data read race on startup and when switching shaders while resizing up and down - but I think this has been a long-standing issue and out of our control (reported in Intel's IRIS driver).



On macOS, clang's thread sanitizer runs clean doing the same actions, so that bodes well.

kcgen:
Minor heads up - some of eXo's testers on Windows 10 reported that, when Window desktop scaling feature is set above 100% (say, 125% and 150%), Staging's hidpi handling honors this and in turn scales the window up accordingly.

However, because DOSBox 0.74, SVN, and other forks _**don't**_ handle high DPI,  the common exoDOS window resolution dials in exact window resolution values, however when those are used w/ Staging, we (additionally) balloon up by the configured Windows desktop scaling percentage.

To prevent Staging from scaling and to make it behave like the others, the eXoDOS v6 common config disables Windows highDPI scaling by setting the `SDL_WINDOWS_DPI_SCALING=0` environment variable before launch (as an example):

### launch.bat

```
SET SDL_WINDOWS_DPI_SCALING=0
.\emulators\dosbox\staging0.80.1\dosbox.exe -conf .\eXoDOS\!dos\acrhine\dosbox.conf
```

So if someone has a high-DPI Windows machine and can configure their desktop to scale to 125% or 150%, suggest we test this PR using eXo's above configuration (`SDL_WINDOWS_DPI_SCALING=0` prior to launch), and see how it behaves.

Maybe if high DPI handling is disabled, we just fall back to sharp (or something like that). Just trying to avoid misunderstandings that might derail things.

GranMinigun:
There's no need for any special handling, HiDPI acts like a window resize from DOSBox's point of view - viewport gets reconfigured according to the available physical pixels space. Shaders receive physical pixel dimensions, not logical. So, if the user moves a 720p window from a display with 100% scaling configured to a 150% scaled display, then 1080p versions of shaders would kick in and correctly see all the pixels.

Not that I ever got a proof of it actually working on multi-monitor configurations outside of macOS, so testing and confirmation is still quite welcome.

kcgen:
@GranMinigun , I found the report details. 

1. Windows 10 is configured w/ 150% window scaling.
1. A common `dosbox.conf` is set to `windowresolution = 2560x1920`
1. Both screenshots are of the entire desktop, when in Window-mode:

    ECE:
    image

    Staging (without `SET SDL_WINDOWS_DPI_SCALING=0`):
    image-2

When `SET SDL_WINDOWS_DPI_SCALING=0` is set, the Staging window size matches ECE.

GranMinigun:
Because all the settings in the configuration assume logical, unscaled units.

johnnovak:
> With default settings on Linux, GCC thread sanitizer catches a data read race on startup and when switching shaders while resizing up and down - but I think this has been a long-standing issue and out of our control (reported in Intel's IRIS driver).

Not much to do there then 😄 

> On macOS, clang's thread sanitizer runs clean doing the same actions, so that bodes well.

Yep, same experience here with the sanitizer build. Plus I've been resizing windowsz and reloading shaders hundreds of times on both macOS and Windows over the past weeks during tweaking, and no crashes.

> My hunch is that we need proper hiDPI handling for the CRT shaders to behave correctly; so we could consider just falling back to sharp (or something like that) if we detect that hiDPI handling has been disabled.

No, the `SDL_WINDOWS_DPI_SCALING` setting only affects window sizing, effectively, as @GranMinigun has explained. To be clear, the shaders are oblivious to all DPI scaling stuff behind the scenes, they operate on pixels, that's it. Then the auto-switching logic is oblivious to the DPI stuff too; it only takes viewport dimensions (in pixels) and the dimensions of the DOS video mode into account. So both are actually _should_ be oblivious to any DPI scaling, so that's all good.

As per the description of the flag ([source](https://discourse.libsdl.org/t/sdl-windows-dpi-scaling-highdpi-support/37010)), it only affects the handling of logical coordinates and thus window dimension calculations. The shaded output should still map to exact physical pixels without any extra scaling whatsoever with `SDL_WINDOWS_DPI_SCALING` either enabled or disabled (it needs testing, but that's the theory). So in theory it all just works as it should.

> Adds hint "SDL_WINDOWS_DPI_SCALING" which can be set to "1" to
> change the SDL coordinate system units to be DPI-scaled points, rather
> than pixels everywhere.
>
> **This means windows will be appropriately sized, even when created on
> high-DPI displays with scaling.**
>
> e.g. requesting a 640x480 window from SDL, on a display with 125%
> scaling in Windows display settings, will create a window with an
> 800x600 client area (in pixels).
>
> Setting this to "1" implicitly requests process DPI awareness
> (setting SDL_WINDOWS_DPI_AWARENESS is unnecessary),
> and forces SDL_WINDOW_ALLOW_HIGHDPI on all windows.

johnnovak:
Also note this last paragraphs in `SDL_WINDOWS_DPI_SCALING`s description:

> Setting this to "1" implicitly requests process DPI awareness
> (setting SDL_WINDOWS_DPI_AWARENESS is unnecessary),
> and forces SDL_WINDOW_ALLOW_HIGHDPI on all windows.

So if we set `SDL_WINDOWS_DPI_SCALING` to 0, we _might_ need to manually set `SDL_WINDOWS_DPI_AWARENESS` to 1, and perhaps even `SDL_WINDOW_ALLOW_HIGHDPI`. Ultimately, it needs some experimentation on eXo's part (or whoever has a high DPI monitor in our team who uses Windows—I don't have one).

src/gui/render.cpp
```diff
@@ -140,9 +144,9 @@ static void empty_line_handler(const void*) {}
 static void start_line_handler(const void* s)
 {
 	if (s) {
-		auto src   = static_cast<const uintptr_t*>(s);
+		auto src = static_cast<const uintptr_t*>(s);
 		auto cache = reinterpret_cast<uintptr_t*>(render.scale.cacheRead);
-		for (Bits x = render.src.start; x > 0;) {
+		for (Bits x = render.src_start; x > 0;) {
```
kcgen, line 149: maybe similar to your capture code, and we can replace the `Bits x` with:

```c++
auto pixels_to_render = render.src_start;
while (pixels_to_render--) {
  ...
}
```

src/gui/render.cpp
```diff
@@ -373,22 +390,19 @@ static void render_reset(void)
...
 		gfx_flags        = (gfx_flags & ~GFX_CAN_8);
 		break;
 	case PixelFormat::BGR888:
-		render.src.start = (render.src.width * 3) / sizeof(Bitu);
+		render.src_start = (render.src.width * 3) / sizeof(Bitu);
 		gfx_flags        = (gfx_flags & ~GFX_CAN_8);
 		break;
 	case PixelFormat::BGRX8888:
-		render.src.start = (render.src.width * 4) / sizeof(Bitu);
+		render.src_start = (render.src.width * 4) / sizeof(Bitu);
```
kcgen, line 405: maybe hoist the `sizeof(Bitu)`s out of the switch to a common `constexpr auto src_pixel_bytes = sizeof(uintptr_t)` (or a name like that?)

src/gui/render.cpp
```diff
@@ -527,364 +542,144 @@ void RENDER_SetSize(const uint16_t width, const uint16_t height,
...
+	                                        InterpolationMode::NearestNeighbour);
 
-	std::stringstream buf;
-	buf << fshader.rdbuf();
-	fshader.close();
-	if (buf.str().empty()) {
-		return false;
-	}
+	force_vga_single_scan   = nearest_neighbour_enabled;
+	force_no_pixel_doubling = nearest_neighbour_enabled;
```
kcgen, line 573: Very thorough comment! Not much code, but non-trivial concepts :-)

src/gui/render.cpp
```diff
@@ -527,364 +542,144 @@ void RENDER_SetSize(const uint16_t width, const uint16_t height,
...
-	if (s.empty()) {
-		source.clear();
-		LOG_ERR("RENDER: Failed to read shader source");
-		return false;
-	}
-	source = std::move(s);
-	assert(source.length());
-	return true;
-}
+	const auto changed_shader = (new_shader_name != render.current_shader_name);
```
kcgen, line 608: The `new_shader_name` is a bit confusing because we're asking the shader manager to get the "CurrentShaderInfo"'s name.

Should `CurrentShaderInfo` be `NewShaderInfo` instead? 

Or maybe `CurrentShaderInfo` is simply _more current_ (or newer) than the `render.current_shader_name`?

src/gui/render.cpp
```diff
@@ -527,364 +542,144 @@ void RENDER_SetSize(const uint16_t width, const uint16_t height,
...
-	return true;
-#endif
+	return using_opengl;
 }
 
-void RENDER_InitShader(Section* sec)
+std::deque<std::string> RENDER_InventoryShaders()
 {
-	init_shader(sec);
+	return get_shader_manager().InventoryShaders();
```
kcgen, line 663: Maybe `RENDER_GetShadersNames()`? (is that what it's returning?)

`InventoryShaders()` tells me it's doing the inventorying work, and a return value might not be needed. "CollateTheBooks", "InventoryTheStuff"... just do some internal book keeping.

src/gui/render.cpp
```diff
@@ -940,19 +735,27 @@ static void init_render_settings(Section_prop& secprop)
...
+	        "  auto:        'vertical' mode auto-enabled for adaptive CRT shaders only,\n"
+	        "               otherwise 'off' (default).\n"
+	        "  vertical:    Constrain the vertical scaling factor to integer values within\n"
+	        "               the viewport. This is the recommended setting when using CRT\n"
+	        "               shaders to avoid uneven scanlines and unwanted interference\n"
+	        "               artifacts.\n"
+	        "  horizontal:  Constrain the horizontal scaling factor to integer values within\n"
+	        "               the viewport.\n"
+	        "  off:         No integer scaling constraint is applied; the image fills the\n"
+	        "               viewport according to the 'aspect' setting.");
```
kcgen, line 754: +1, great descriptions!

src/gui/render.cpp
```diff
@@ -1007,19 +810,37 @@ static void init_render_settings(Section_prop& secprop)
...
+	        "                     'machine' setting. E.g., CGA and EGA games will appear\n"
+	        "                     double-scanned on an emulated VGA adapter.\n"
+	        "  crt-auto-arcade:   Emulation of an arcade or home computer monitor less sharp\n"
+	        "                     than a typical PC monitor with thick scanlines in low-\n"
+	        "                     resolution modes. This fantasy option does not exist in\n"
+	        "                     real life, but it can be a lot of fun, especially with DOS\n"
+	        "                     ports of Amiga games.\n"
+	        "Other options include 'sharp', 'none', a shader listed using the\n"
+	        "--list-glshaders command-line argument, or an absolute or relative path\n"
+	        "to a file. In all cases, you may omit the shader's '.glsl' file extension.");
```
kcgen, line 843: +1, awesome descriptions.

src/gui/shader_manager.h
```diff
@@ -0,0 +1,180 @@
...
+
+// forward references
+class Fraction;
+
+const auto FallbackShaderName = "none";
+const auto SharpShaderName    = "interpolation/sharp";
+
+const auto AutoGraphicsStandardShaderName = "crt-auto";
+const auto AutoMachineShaderName          = "crt-auto-machine";
+const auto AutoArcadeShaderName           = "crt-auto-arcade";
```
kcgen, line 37: `constexpr`

src/gui/shader_manager.h
```diff
@@ -0,0 +1,180 @@
...
+	std::string MapShaderName(const std::string& name) const;
+
+	void NotifyGlshaderSettingChanged(const std::string& shader_name);
+
+	void NotifyRenderParametersChanged(const uint16_t canvas_width,
+	                                   const uint16_t canvas_height,
+	                                   const VideoMode& video_mode);
+
+	ShaderInfo GetCurrentShaderInfo() const;
+	std::string GetCurrentShaderSource() const;
```
kcgen, line 138: Suggest returning `const std::string&`, the caller can store a value  if they want to work with it:

```c++
std::string shader_copy = GetCurrentShaderSource();
```

src/gui/shader_manager.h
```diff
@@ -0,0 +1,180 @@
...
+
+	std::string MapShaderName(const std::string& name) const;
+
+	void NotifyGlshaderSettingChanged(const std::string& shader_name);
+
+	void NotifyRenderParametersChanged(const uint16_t canvas_width,
+	                                   const uint16_t canvas_height,
+	                                   const VideoMode& video_mode);
+
+	ShaderInfo GetCurrentShaderInfo() const;
```
kcgen, line 137: Suggest returning `const ShaderInfo&`, the caller can store it by value if they want to work with it.

src/hardware/vga_draw.cpp
```diff
@@ -1844,15 +1846,20 @@ static Fraction calc_pixel_aspect_from_timings(const VgaTimings& timings)
...
+//   vga.draw.delay.hdend
+//   vga.draw.dos_refresh_hz
+//   vga.draw.resizing
+//   vga.draw.vret_triggered
+//   vga.draw.linear_base
+//   vga.draw.linear_mask
+//   vga.draw.blocks
+//   vga.draw.vblank_skip
+//   vga.draw.line_length
+RenderParams setup_drawing()
```
kcgen, line 1861: very helpful comment list!

kcgen (approved):
Quick review and just some minor suggestions. Looking forward to this addition, @johnnovak !

kcgen:
@GranMinigun , @johnnovak -> thanks for your explanations. (I was worried that the integer-scaling wouldn't "line up" with host pixels if the host magnification wasn't accounted for, but makes sense that the integer multiple only applies to the source image relative to the logical host units)

src/gui/render.cpp
```diff
@@ -22,35 +22,39 @@
...
+static ShaderManager& get_shader_manager()
+{
+	static std::unique_ptr<ShaderManager> shader_manager = nullptr;
+
+	if (!shader_manager) {
+		shader_manager = std::make_unique<ShaderManager>();
+	}
+	return *shader_manager;
+}
+#endif
```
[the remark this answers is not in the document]
johnnovak: Sure, that's simpler, I'll do it like that.

src/gui/render.cpp
johnnovak: Yep, I agree that's better; I just kept the old name. I think it was you who come up with this name, btw 😄

src/gui/shader_manager.h
johnnovak: Sure, that's better.

src/gui/shader_manager.h
johnnovak: Sure.

src/gui/render.cpp
johnnovak: Open to a better name, I'm not 100% happy with "current", but this is how the shader manager works; you feed it events that could potentially change the "current" (to be used) shader, then you query the "current" shader (that should be used now, in reflection of the incoming events). It's described here:

https://github.com/dosbox-staging/dosbox-staging/pull/2777/files#diff-506240744b4aec8b55f275b514845b582ae81f49c727bc4d0e3e54453a352f09R105-R121

So I think that comment described its function accurately, but yeah, open to suggestions. I don't think "GetCurrentShaderInfo" would be an improvement, maybe something else?

src/gui/render.cpp
johnnovak: I can do as much, but otherwise I don't wanna touch this existing old code too much now.

src/gui/render.cpp
johnnovak: Existing code, don't wanna refactor it now. I just had to touch it because I had to after introducing the RenderParams struct.

johnnovak:
@GranMinigun I've tested the new shaders with NPOT textures & `GL_CLAMP_TO_EDGE`; they all work fine. I started adding a `use_npot_texture` pragma so shaders can reques NPOT textures... but then realised that texture creation always happens just _once_ at startup. That's right, NPOT creation only works correctly in main if you set `glshader` in the config. Any runtime changes, it will never recreate the texture based on the shader name whitelist.

So, I have zero appetite to restructure/rewrite the shader init pipeline so it can recreate the textures at runtime... That's a much bigger work, not for now. I left my `use_npot_texture` pragma commit for now, if you want (entirely optional), have a look, then if you have a quick'n'dirty idea to recreate the texture at runtime, please send me a diff 😄 

Otherwise, I'm gonna revert those NPOT commits and this is a job for another day.

GranMinigun:
Texture creation cannot happen just once, else it'd crash when trying to squeeze 1600x1200 pixels image into a 512x512 texture which was intended for a 320x200 emulated framebuffer. The texture gets recreated every time `GFX_SetSize()` is called - which happens literally on every video mode change and window resize (which is far from optimal, and yes, it is addressed in my renderer rework).

johnnovak:
> Texture creation cannot happen just once, else it'd crash when trying to squeeze 1600x1200 pixels image into a 512x512 texture which was intended for a 320x200 emulated framebuffer. The texture gets recreated every time `GFX_SetSize()` is called - which happens literally on every video mode change and window resize (which is far from optimal, and yes, it is addressed in my renderer rework).

You're right, I'm wrong.

Well, you're fully right, but I'm only half-wrong 😎 NPOT support at the OpenGL is checked _once_ at startup, and it's conflated with also checking whether the shader supports it.

I just need to untangle that criteria, then this will work I think.

johnnovak:
@GranMinigun Done, tested, working, thanks for the fish 🐠 😎

src/gui/render.cpp
johnnovak: Leaving as is for now.

johnnovak:
@GranMinigun If you could take a look at the NPOT stuff that is now working fine, that would be great, then merge time 😎 🤘🏻

src/gui/render.cpp
kcgen: Understandable :+1:

src/gui/render.cpp
kcgen: Perfect; thanks @johnnovak. It's been a while so I forgot I wrote that. The function name is much clearer now, thanks for that! :rocket:

johnnovak:
Merging; @GranMinigun can complain later if he doesn't like something about the NPOT support 😄

GranMinigun:
I don't like having over 9000 pragmas, but whatever. Here, I complained, and later.

johnnovak:
> I don't like having over 9000 pragmas, but whatever. Here, I complained, and later.

The good news is we have 5, so long way until we hit 9000 😄 We can use either pragmas or an extra metadata file per shader. Hardcoding info about shaders is the worst possible solution, bye bye extendability by the user.

Custom pragmas are supported, even encouraged by the standard to store metadata, look it up. They are used by RA shaders too to add metadata to tweakable shader param uniforms, such as labels and parameter ranges. Which is nice.

So, I guess liking or disliking them is immaterial as they're the only practical solution 😄 (unless you add an extra metadata file for each shader, which has pros and cons).

PR #2779, opened 2023-08-22
Downgrade Visual Studio Platform Toolset to v142

johnnovak:
v142 is the last support version on VS 2019; v143 requires VS 2022.

### DIFF
--- a/src/libs/zmbv/zmbv.vcxproj
+++ b/src/libs/zmbv/zmbv.vcxproj
@@ -34,7 +34,7 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Tracy|Win32'" Label="Configuration">
@@ -44,7 +44,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Tracy|x64'" Label="Configuration">
--- a/tests/vs/tests.vcxproj
+++ b/tests/vs/tests.vcxproj
@@ -43,7 +43,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Tracy|Win32'" Label="Configuration">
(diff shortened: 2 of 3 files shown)


johnnovak:
> I don't have any objections, but can't users retarget as necessary? Or that upsets pulls from the repository? And CI runs VS2022, if I remember correctly.

I was under the impression we want DOSBox Staging to compile on VS2019 for... whatever reason. The README states Visual Studio 2019 or newer:

https://github.com/dosbox-staging/dosbox-staging#windows---visual-studio-2019-or-newer

So we need to either:

- Update the README and state that VS2022 is the new minimum requirement
- Or merge this, because otherwise people couldn't compile the project out-of-the-box

@kcgen @shermp any specific reason to stay on VS2019? I can install VS2022 on my machine if needed.

kcgen:
We should try (or need) to support the latest compilers available on still-in support OSes.

If windows 10 is still in-support and vs2019 is the latest compiler that's allowed to be installed (that is: vs2022 requires windows 11+), then yes: that's the criteria to hold us back to 2019 until windows 10 goes EOL. 

But if win10 users can install vs2022: then there's no reason to keep 2019 compatibility.

kcgen:
(same rules apply on Linux: just because users can still install GCC 4.8 on Debian 10, doesn't mean we're supporting it or defaulting to that. We want users on the latest compilers their in-support OS supports)

johnnovak:
Makes sense @kcgen. Turns out VS2022 can run on Win10:
https://learn.microsoft.com/en-us/visualstudio/releases/2022/system-requirements

This means I'll just upgrade and update the project readme instead.

Btw, I have slight concerns about perhaphs losing the ability to build for Win7 anymore, but I guess the MinGW build would continue to satisfy that need. Just in case I won't be able to Win10 anymore because MS will make it so hard on my current PC, and I won't be able to run Win11 on it either because it has a (very powerful) 4th gen Intel.

So in that eventuality I can easily see myself just using Win7 on it and treat it as an offline gaming / audio production box that will run all sorts of emulators and my audio apps fine until the hardware dies... maybe for the next 20 years?

But if we started to use C++ features only VS2022 supports, building for Win7 will become increasingly difficult. Clearly, I'm not saying let's stay on VS2019 / C++17 forever because one person on the dev team _might_ want to downgrade to Win7 one day, but it's something to consider. I know guys who use XP to this day as offline audio production boxes -- with audio stuff and tons of VST plugins people are reluctant to upgrade as shit tends to break left right center, so many people just have a working OS with stable drivers "frozen in time". I'm guessing there might be an overlap here with retro gamers who don't care about online, Steam, etc. but just wanna run emulators and old games.

GranMinigun:
When people still consider WinXP above Linux-based systems, clearly something went wrong.

Visual Studio 2022 should still allow to build for WinVista, and Visual C++ 2015-2022 redistributale claims to support it as well. This should cover project's needs even for C++20 features, but I'm not so sure about C++23. MinGW/Cygwin/MSYS2 do not care, with statically linked STL and stuff the resulting binaries can run on anything that have the required instruction sets, - as long as nothing depends on specific Win32 API functionality (GLib does, which is a dependency of FluidSynth, which already cut us from Win7).

johnnovak:
> When people still consider WinXP above Linux-based systems, clearly something went wrong.

Haha, I tend to agree. Anyway, that's a whole different discussion, and I'm doing audio production as a hobby, and 99% of VST plugins I use only exist on Windows, the FireWire audio interface I have since ~2008 only has Windows and macOS drivers (which is true for all semi-pro audio interfaces; the RME pro interfaces had Linux drivers for a while, but even they don't bother anymore... there is no market for it, people just use Mac/Windows for audio work).

Anyway, long story and largely irrelevant, but Linux is not suitable for my needs, to put it simply.

> Visual Studio 2022 should still allow to build for WinVista, and Visual C++ 2015-2022 redistributale claims to support it as well. This should cover project's needs even for C++20 features, but I'm not so sure about C++23. MinGW/Cygwin/MSYS2 do not care, with statically linked STL and stuff the resulting binaries can run on anything that have the required instruction sets, - as long as nothing depends on specific Win32 API functionality (GLib does, which is a dependency of FluidSynth, which already cut us from Win7).

That's great info, thanks for that. The situation actually sounds really good then; I can easily just not compile in FluidSynth support and Bob's your uncle (I have zero need for it; I'd much rather use my hardware, Roland's SCVA plugin, or the free S-YXG50 VST plugin, in that order).

GranMinigun:
Correction: VS2022 seems to have [dropped](https://learn.microsoft.com/en-us/visualstudio/releases/2022/compatibility#build-apps-that-run-on-windows-clients) support for building for Vista, and C++ runtime covers [only 2015 and 2017](https://learn.microsoft.com/en-us/visualstudio/releases/2022/system-requirements#microsoft-visual-c-2015-2022-redistributable-system-requirements) tools for Win7 (which, I guess, are available within VS2022, but might require retargeting).

johnnovak:
> Correction: VS2022 seems to have [dropped](https://learn.microsoft.com/en-us/visualstudio/releases/2022/compatibility#build-apps-that-run-on-windows-clients) support for building for Vista, and C++ runtime covers [only 2015 and 2017](https://learn.microsoft.com/en-us/visualstudio/releases/2022/system-requirements#microsoft-visual-c-2015-2022-redistributable-system-requirements) tools for Win7 (which, I guess, are available within VS2022, but might require retargeting).

Yeah, I'm not even sure what that means, but ok, let's go with VS2022 for now.

PR #2780, opened 2023-08-22
Retire surface output

### LINKED ISSUE #433: Deprecate `output = surface`
Hi, 
- My CRT Display support hardware resolution 640x400 (added it with Custom Resolution Utility to windows registry) and <= DOSBOX 0.74 versions use it as fullscreen without BSODs
Display also support hardware 640x480, no need to add it manualy, it already report this resolution in EDID

- Appllication: Dungeon Keeper Gold, dkeeper.exe
This game uses three resolution and dynamicaly switch between:
640x480 - first load screen, menu screens
320x200 - cinematics
640x400 - in-game resolution

- no BSOD if game started in Window mode

- BSOD occurs before or when game try to show first load screen, game started from 640x400 fullscreen DOSBOX command prompt:
_MAIN: Draw resolution: 640x400, pixel aspect ratio: 1.20
MAIN: Emulator resolution: 640x480_

- Settings:
fullresolution   = original
output           = surface
texture_renderer = auto
scaler    = none

EDIT: details formerly provided in the title moved there for lisibility:
BSOD at .text:014903C0 (mov [eax], ecx), Windows 7 x64, v0.75.0, FullScreen 640x400

GranMinigun:
This is a flammenwerfer, it werfs flammen.

I don't like that `goto` inside `GFX_SetSize()`. `GFX_GetBestMode()` seems to be redundant. PVS didn't like `get_rendered_output_from_backbuffer()`, but perhaps a switch would fit there better. Smart ideas are welcome.

Oh, and it's completely untested.

Closes #433.


johnnovak:
Nice @GranMinigun 👍🏻 

> Oh, and it's completely untested.

So who will test it? 🤔

GranMinigun:
Dunno. It doesn't let me. Says surface is deprecated or something.

Maybe users should? I heard Microsoft successfully utilizes that strategy in their products.

src/gui/render.cpp
```diff
@@ -338,15 +338,15 @@ static void render_reset(void)
 	// driver operating in a different thread or process.
 	std::lock_guard<std::mutex> guard(render_reset_mutex);
 
-	Bitu width         = render.src.width;
+	uint32_t width     = render.src.width;
```
johnnovak, line 341: uint16_t?

src/gui/render.cpp
```diff
@@ -338,15 +338,15 @@ static void render_reset(void)
...
 	bool double_width  = render.src.double_width;
 	bool double_height = render.src.double_height;
 
-	Bitu gfx_flags, xscale, yscale;
+	uint8_t gfx_flags, xscale, yscale;
 	ScalerSimpleBlock_t* simpleBlock = &ScaleNormal1x;
 
 	// Don't do software scaler sizes larger than 4k
-	Bitu maxsize_current_input = SCALER_MAXWIDTH / width;
+	uint32_t maxsize_current_input = SCALER_MAXWIDTH / width;
```
johnnovak, line 349: uint16_t?

GranMinigun:
Reproduced: "The specified window isn't an OpenGL window"

Just to be sure we're having the same error:
```diff
diff --git a/src/gui/sdlmain.cpp b/src/gui/sdlmain.cpp
index 576370596..fe87b8d5a 100644
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3284,7 +3284,7 @@ static void set_output(Section* sec, const bool wants_aspect_ratio_correction)
                } else {
                        sdl.opengl.context = SDL_GL_CreateContext(sdl.window);
                        if (sdl.opengl.context == nullptr) {
-                               LOG_WARNING("Could not create OpenGL context, switching back to texture");
+                               LOG_WARNING("Could not create OpenGL context, switching back to texture; error message: %s", SDL_GetError());
                                sdl.desktop.want_type = SCREEN_TEXTURE;
                        }
                }
```

johnnovak:
> Reproduced: "The specified window isn't an OpenGL window"
> 
> Just to be sure we're having the same error:

Yeah, I'm getting the same error:

image

GranMinigun:
Moving to draft for now. Too many things broke for visibly no reason:
- Toggling fullscreen mode hangs;
- Resizing window to smaller than initial size blanks it;
- Resizing mapper blanks it (but `--startmapper` isn't affected);
- Aforementioned output switching, which is an oversight on my side - the window isn't recreated.

GranMinigun:
I can't read. Ready for additional testing and review.

johnnovak:
> I can't read. Ready for additional testing and review.
> 
> (Just realized output switching is still borked. Sorry, ~will fix it tomorrow~ fixed it now. I can't read.)

Will test & review tomorrow then, when you think you're done 😄

src/gui/sdlmain.cpp
```diff
@@ -655,42 +651,10 @@ void GFX_RequestExit(const bool pressed)
...
-			goto check_surface;
-		}
-		flags&=~(GFX_CAN_8|GFX_CAN_15|GFX_CAN_16);
-		break;
-	default:
-		goto check_surface;
-		break;
-	}
+	flags &= GFX_CAN_32;
+	flags &= ~(GFX_CAN_8 | GFX_CAN_15 | GFX_CAN_16);
```
kcgen, line 657: Huge cleanup!
Given this functions always does the same thing now (turns on "can-do-32-bit colour" and shuts off the other bits), this need for it is getting quite tenuous.. so next refactoring it could probably just be thrown out (and we simply assume `GFX_CAN_32`).. making the need for all those GFX_CAN flags also unnecessary too.

GranMinigun: In theory, we could check for 16-bit texture format availability for OpenGL, or other formats for SDL renderers (its software backend supports even indexed formats). But probably I'll be doing that in a different manner, in a different place, if at all.

johnnovak: Isn't that micro-optimisation? The most demanding late (S)VGA games are almost all rendered as 32-bit because of the 18-bit RGB palette, no? And it's rather pointless to optimise anything for the very undemanding CGA/EGA/etc. games.

I would actually do the opposite, which is the current behaviour; just treat everything as 32-bit, job done.

I'm a ruthless simplifier, I recommend that to everyone 😄  I only optimise if the gains are significant (more than 10%, in general).

GranMinigun: Currently that conversion to 32-bits is performed by the CPU. Considering we're constantly short on cycles, feeding raw image data to the GPU whenever possible is far from being a "micro"-optimization (unless you're a rich boi with 13900K + RTX 4090 setup). I'm not talking about the final rendered image though, it's almost always 8- or 10-bits per channel (or float, in case of HDR displays).

johnnovak: i7 4790k + RTX 3060 here, bought in 2015 😎 That's was middle of the road hardware even then, 8 years ago, and every DOS game is blazing fast already on my machine. So was on my previous card, a GTX 750 Ti with all emulators, WinUAE, PCSX2, VICE, etc.

So I'm not really buying that it matters, but benchmark it, and then we're talking 😄 

But that's a pretty sane way of working: do the simple thing first, benchmark, then complicate it if it really matters (after benchmarking, not before based on assumptions).

johnnovak: Plus, if you're thinking about just passing the palette indexes for 256 colour graphics plus the palette to the GPU (you mentioned this before) as byte values, beware that some games and demos change the palette mid-screen. That's the reason why currently we're rendering all 256-colour modes as 32-bit because that's fiddly to detect whether the palette was changed mid-screen or not.

Most programs don't do this, of course, but you still need to handle this special case... somehow if you want to try this.

johnnovak:
Let me know when this is ready fro review and another round of testig @GranMinigun.

GranMinigun:
Unless you can comment on the things I pointed out in the first post, it was ready for a review for about a week. Just needs a rebase.

include/video.h
```diff
@@ -60,8 +60,8 @@ bool GFX_Events();
 // Useful during output initialization or transitions.
 void GFX_DisengageRendering();
 
-Bitu GFX_GetBestMode(Bitu flags);
-Bitu GFX_GetRGB(uint8_t red,uint8_t green,uint8_t blue);
+uint8_t GFX_GetBestMode(uint8_t flags);
+uint32_t GFX_GetRGB(uint8_t red, uint8_t green, uint8_t blue);
```
johnnovak, line 64: Nice 👍🏻

johnnovak: Can we make the args const?

src/gui/sdlmain.cpp
```diff
@@ -1316,29 +1280,19 @@ static void check_and_handle_dpi_change(SDL_Window* sdl_window,
 }
 
 static SDL_Window* SetWindowMode(SCREEN_TYPES screen_type, int width,
-                                 int height, bool fullscreen, bool resizable,
-                                 bool allow_highdpi = true)
+                                 int height, bool fullscreen)
```
johnnovak, line 1283: Make them const while we're at it?

src/gui/sdlmain.cpp
```diff
@@ -2774,44 +2606,9 @@ static bool present_frame_gl()
 }
 #endif
 
-// Surface update & presentation
-// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-static void update_frame_surface([[maybe_unused]] const uint16_t *changedLines)
+uint32_t GFX_GetRGB(uint8_t red, uint8_t green, uint8_t blue)
```
johnnovak, line 2609: const args?

johnnovak:
Ok, `opengl` <-> `texture` switching works now, but every time I switch to `texture` I'm getting this in the logs:

```
2023-08-28 21:36:24.647 | DISPLAY: Initialised 1103x827 windowed mode using bilinear scaling on display-0
2023-08-28 21:36:24.657 | SDL: Failed setting the OpenGL vsync state to off (0): No OpenGL context has been made current
2023-08-28 21:36:24.665 | SDL: Using 'metal' driver for 32-bit texture rendering
2023-08-28 21:36:24.665 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz VFR, scaled to 2205x1654 with 1:1.35 (20:27) pixel aspect ratio
```

Doesn't seem right that I'm getting an OpenGL related error in `texture` mode.

johnnovak:
Another one. Start up with default config, switch to fullscreen, execute `output = texture`. Hard crash on macOS.

image

GranMinigun:
Interesting. I can't reproduce the crash, but it seems like the root cause is the same as in #2603.

johnnovak:
> Interesting. I can't reproduce the crash, but it seems like the root cause is the same as in #2603.

SDL's behaviour is _not_ normalised across OSes. It really sucks, I don't get it why they don't make it their no.1 goal.

src/gui/sdlmain.cpp
```diff
@@ -4476,8 +4221,7 @@ void config_add_sdl() {
 	pstring->Set_values(presentation_modes);
 
 	const char* outputs[] =
-	{ "surface",
-	  "texture",
+	{ "texture",
```
FeralChild64, line 4224: Please update `windowresolution` help string, it mentions `output=surface`:

```C++
	pstring = sdl_sec->Add_string("windowresolution", on_start, "default");
	pstring->Set_help(
	        "Set window size when running in windowed mode:\n"
	        "  default:   Select the best option based on your environment and\n"
	        "             other settings.\n"
	        "  small, medium, large (s, m, l):\n"
	        "             Size the window relative to the desktop.\n"
	        "  <custom>:  Scale the window to the given dimensions in WxH format.\n"
	        "             For example: 1024x768.\n"
	        "             Scaling is not performed for output=surface.");
```

kcgen: `git grep -i surface`

johnnovak:
> Interesting. I can't reproduce the crash, but it seems like the root cause is the same as in #2603.

Kinda related, maybe.

I'm getting random crashes at startup with the following config on macOS. Happens maybe 30-40% of the time, not always.

I'm using this simple config and start with `--noprimaryconf`. Whether I comment out `viewport_resolution` or not, that seems to make little difference, the crashes can happen in either case randomly.

```
[sdl]
#viewport_resolution = 100%
windowresolution = desktop
```

```
Assertion failed: (canvas.w > 0 && canvas.h > 0), function get_canvas_size, file sdlmain.cpp, line 1434.

Stack trace:
45         0x106ed108c 49  dyld                                0x0000000106ed108c start + 520
44         0x102ce67b8 main + 36
43         0x1030a862c sdl_main(int, char**) + 4796
42         0x102d75840 Config::StartUp() + 28
41         0x10343bc58 SHELL_Init() + 3992
40         0x103438f94 DOS_Shell::Run() + 1524
39         0x10347b0e0 DOS_Shell::InputCommand(char*) + 56
38         0x10347b664 DOS_Shell::ReadCommand() + 708
37         0x102f045d0 DOS_ReadFile(unsigned short, unsigned char*, unsigned short*, bool) + 280
36         0x102efeaa8 DOS_Device::Read(unsigned char*, unsigned short*) + 64
35         0x102efc684 device_CON::Read(unsigned char*, unsigned short*) + 312
34         0x102e2ebb0 CALLBACK_RunRealInt(unsigned char) + 92
33         0x102ce71b0 DOSBOX_RunMachine() + 28
32         0x102ce7098 Normal_Loop() + 20
31         0x103289644 PIC_RunQueue() + 292
30         0x10331f234 VGA_SetupDrawing(unsigned int) + 8060
29         0x1032c66a0 ReelMagic_RENDER_SetSize(unsigned short, unsigned short, bool, bool, Fraction const&, PixelFormat, double, VideoMode const&) + 204
28         0x1032c6734 SetupVideoMixer(bool) + 136
27         0x103041668 RENDER_SetSize(unsigned short, unsigned short, bool, bool, Fraction const&, PixelFormat, double, VideoMode const&) + 244
26         0x103041aa8 render_reset() + 1000
25         0x10309ea74 GFX_SetSize(int, int, Fraction const&, unsigned char, VideoMode const&, void (*)(GFX_CallBackFunctions_t)) + 2440
24         0x10309fe08 SetupWindowScaled(SCREEN_TYPES) + 364
23         0x1030aaf58 SetWindowMode(SCREEN_TYPES, int, int, bool) + 348
22         0x107019698 SDL_CreateWindow_REAL + 1160
21         0x10701b458 SDL_ShowWindow_REAL + 104
20         0x107034254 Cocoa_ShowWindow + 160
19         0x1bc99cba0 -[NSWindow makeKeyAndOrderFront:] + 72
18         0x1bc9a850c -[NSWindow _doOrderWindow:relativeTo:findKey:forCounter:force:isModal:] + 324
17         0x1bc9a9520 -[NSWindow _reallyDoOrderWindow:relativeTo:findKey:forCounter:force:isModal:] + 132
16         0x1bc9a9d24 -[NSWindow _reallyDoOrderWindowAboveOrBelow:relativeTo:findKey:forCounter:force:isModal:] + 1000
15         0x1bc8c5208 -[NSWindow _setFrameCommon:display:fromServer:] + 3284
14         0x1bab95c98 -[NSNotificationCenter postNotificationName:object:userInfo:] + 96
13         0x1b9cd01c0 _CFXNotificationPost + 732
12         0x1b9da5b84 _CFXRegistrationPost + 456
11         0x1b9da5c18 ___CFXRegistrationPost_block_invoke + 52
10         0x1b9d0151c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 28
9          0x107033284 -[Cocoa_WindowListener windowDidResize:] + 396
8          0x106faa558 SDL_SendWindowEvent + 688
7          0x106fa50d0 SDL_PushEvent_REAL + 172
6          0x1030ab688 watch_sdl_events(void*, SDL_Event*) + 132
5          0x1030ab730 HandleVideoResize(int, int) + 104
4          0x10309ffe0 get_canvas_size(SCREEN_TYPES) + 140
3          0x1b9b79754 err + 0
2          0x1b9b7a340 abort + 168
1          0x1b9c3fee0 pthread_kill + 288
2023-08-29 14:03:54.194 | Signal: SIGABRT
zsh: abort      ~/jn/dosbox-staging/build/debug/dosbox --noprimaryconf

```

kcgen:
Repro'd 100% on my side on macOS. 
Exactly as you said, @johnnovak, our SDL resource cleanup step wasn't  cleaning up the callabck watch (tested and pushed the one-liner to main).
Should be good to go if you rebase, @GranMinigun :rocket:

kcgen:
Hmm to delete a previously added watch, `SDL_DelEventWatch` needs both the function pointer and the originally added userdata to match _at the time of adding_ (otherwise the watch won't be deleted).

So we need to keep track of the originally-added userdata. Will make one or push, and that should always knock out the watch (in all scenarios).

johnnovak:
> So we need to keep track of the originally-added userdata. Will make one or push, and that should always knock out the watch (in all scenarios).

That's good if we make it work, but how about just using standard SDL window resized events instead? Is this workarlund still necessary?

kcgen:
> > So we need to keep track of the originally-added userdata. Will make one or push, and that should always knock out the watch (in all scenarios).
> 
> That's good if we can make it work, but how about just using standard SDL window resized events instead? Is this workarlund still necessary?
> 
> I will test this with your cleanup later @kcgen, but I don't think that fix can help with the random startup crash? Am I missing something?

yup, it should fix these crashes too.

I've done a temporary rebase of @GranMinigun's branch here on top of the improved function for you to try, @johnnovak:

It's: **`kc/gm/surface-purge-2`** - hopefully it's uncrashable on macOS!

kcgen:
> Tried the branch and restarted 20-30 times, then switched output/texture in fullscreen 20-30 times.

The `kc/gm/surface-purge-2` branch includes @GranMinigun's PR plus the fix for the long-standing incorrectly managed SDL event watch bug.

Suggest we consider merging this `kc/gm/surface-purge-2`, just to avoid re-work and cherry-picking, etc.

I've also included a couple separate commits to fix the OpenGL vsync issue (now moved to the end of the function) and address @FeralChild64's comment.

One remaining issue is that `output = texture`, on macOS, defaults to software texture_renderer (not sure why):

```text
2023-08-29 16:27:33.284 | SDL: Using standard SDI (auto) display refresh rate of 60 Hz
2023-08-29 16:27:33.284 | SDL: Using 'software' driver for 32-bit texture rendering
2023-08-29 16:27:33.352 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz throttled VFR, scaled to 1920x1440 with 1:1.35 (20:27) pixel aspect ratio
2023-08-29 16:27:37.040 | DISPLAY: VGA 320x200 256-colour graphics mode 13h at 70.086 Hz throttled VFR, scaled to 1920x1440 with 1:1.2 (5:6) pixel aspect ratio
2023-08-29 16:27:41.025 | SDL: We can only render 40 FPS, which is well below the host's reported refresh rate of 60 Hz.
2023-08-29 16:27:41.025 | SDL: You will experience rendering lag and stuttering. Consider updating your video drivers and try disabling vsync in your host and drivers, set [sdl] vsync = false
2023-08-29 16:27:41.025 | DISPLAY: VGA 320x200 256-colour graphics mode 13h at 70.086 Hz throttled VFR, scaled to 3840x2880 with 1:1.2 (5:6) pixel aspect ratio
```

Where as the `main` branch will use `metal` by default.

johnnovak:
> > Tried the branch and restarted 20-30 times, then switched output/texture in fullscreen 20-30 times.
> 
> The `kc/gm/surface-purge-2` branch includes:
> 
>     * @GranMinigun's PR,
> 
>     * fixes the long-standing incorrectly managed SDL event watch bug fix,
> 
>     * fixes the OpenGL vsync warning (separate commit, so blame rests w/ me :-)
> 
>     * addresses the open comments here (separate commit as well)
> 
> 
> Suggest we consider merging `kc/gm/surface-purge-2` when testing is complete, just to avoid the hassle of cherry-picking and branch rework.
> 
> Let me know what you think, @GranMinigun, @johnnovak, and @FeralChild64.

Sounds all very reasonable to me; let's do that! 🚀

Grounded0:
@kcgen seems to be working fine after 20 launches on macOS.

kcgen:
Thanks for the help testing, @Grounded0 !

kcgen:
@GranMinigun: testing is OK so far on `kc/gm/surface-purge-2`.
Holding for your final go-ahead, if it's OK from your side.

src/gui/sdlmain.cpp
GranMinigun: Backported from @kcgen's branch.

GranMinigun:
> By they way, the `#else` branch of `SDL_VERSION_ATLEAST(2, 26, 0)` is completely non-functional; if I force using that branch, that results in a crash at startup 100% of the time. So might wanna remove that whole branch altogether?

No. It won't compile on Ubuntu 20.xx (and readme mentions support for SDL 2.0.5), and `SDL_GetWindowSizeInPixels()` does the exact same thing under the hood, so you're hitting the exact same behaviour here. 

Wiped event filter completely, it only fights with SDL's window handling. macOS freezes output during window resize anyway. Should be ready for another round of testing.

johnnovak:
> Wiped event filter completely, it only fights with SDL's window handling. macOS freezes output during window resize anyway. Should be ready for another round of testing.

Thanks that was my point; it seemed completely unnecessary.

kcgen:
Testing is producing good results, @GranMinigun.

I'm seeing a minor but inconsequential functional difference versus `main` on macOS, which are probably acceptable/expected given the new (and more proper) SDL window re-use approach instead of the old hamfisted "rebuild everything from scratch" approach.

Here's what I'm seeing:

On _macOS_ (only) and in _fullscreen_ mode, when manually flipping between `output` types, the `main` branch reconstructs and stays in fullscreen mode, where as SDL in the branch will fallback to a maximized window-mode at flip time.

It seems this behavior is entirely driven by SDL and not through any new logic in the PR.

Here are the comparison videos to see how this plays out:

 - `main` branch - https://kcgen.duckdns.org/main-fs-window-toggle.mp4
 - `PR` branch - https://kcgen.duckdns.org/surface-pr-fs-window-toggle.mp4

Warning - the videos are 4k, h264, and in super-bloated QuickTime bitrates :joy:.  Also, my 5-button PC mouse cursor tends to "go behind the scene" in all branches regularly, so please disregard that aspect; it's nothing to w/ this branch.

kcgen:
> Ok, opengl <-> texture switching works now, but every time I switch to texture I'm getting this in the logs:

```
SDL: Failed setting the OpenGL vsync state to off (0):
     No OpenGL context has been made current
```

@johnnovak -> thanks for the heads up on this. I also get this in the `main` branch, and is a bug I introduced w/ the VFR PR. I'll follow up after this with a commit for it (out of scope of this PR).

kcgen (approved):

johnnovak:
> @johnnovak -> thanks for the heads up on this. I also get this in the `main` branch (see log output in above `main` video). It's a bug I added w/ the VFR PR, so we can consider it out of scope from this PR. I'll follow up after this with a commit for it.

Sounds good @kcgen. Talking about VFR, man, that thing keeps giving 😅 On my fixed 60 Hz monitor the default VFR mode with vsync off now works reliably in the sense that no frames are lost when the video output is only infrequently updated, but because the duping now happens at 1/10th of the host rate or something, you get visible tearing that "resolves itself" after around 60-80 ms. That very visible and it's not an authentic experience, to put it mildly.

As we have established, this can happen in any graphics or text mode; just text mode apps are the ones that tend to update the screen only sporadically, but I'm getting it all the time in the Quake video options menu, for example.

So you can have a look at some pont if you have capacity, otherwise I will because I find this pretty jarring. It even manifests at a simple command prompt; the blinking of the cursor and a normal typing rate is not enough to trigger new frames at a sufficiently high FPS to prevent this from happening 100% of the time.

I don't quite get it why the previous "dupe at the rate of the DOS mode" thing had to be removed.

GranMinigun:
> This window behavior appears to entirely be driven by SDL re-using its window but getting a new renderer

No, it's still recreated, that's the only way to ensure the window has all necessary flags for whatever renderer is chosen. Why it goes maximized instead of fullscreen, I can't tell right now. I think Burrito noticed that behaviour in my sdlpain cleanup PR as well. Something's fishy.

kcgen:
> > This window behavior appears to entirely be driven by SDL re-using its window but getting a new renderer
> 
> No, it's still recreated, that's the only way to ensure the window has all necessary flags for whatever renderer is chosen. Why it goes maximized instead of fullscreen, I can't tell right now. I think Burrito noticed that behaviour in my sdlpain cleanup PR as well. Something's fishy.

I'm not seeing the PR changing or resetting the fullscreen boolean value, and combined with the fact that it switches to a /maximized/ window points to some state being partially lost or maybe it's unable to fully re-engag full screen the moment after the renderer is created (maybe there's some kind of host-side "settle-out" delay after the renderer is re-initialized before the window manager lets us Kickin full screen, or something like that).

kcgen:
> > @johnnovak -> thanks for the heads up on this. I also get this in the `main` branch (see log output in above `main` video). It's a bug I added w/ the VFR PR, so we can consider it out of scope from this PR. I'll follow up after this with a commit for it.
> 
> Sounds good @kcgen. Talking about VFR, man, that thing keeps giving 😅 On my fixed 60 Hz monitor the default VFR mode with vsync off now works reliably in the sense that no frames are lost when the video output is only infrequently updated, but because the duping now happens at 1/10th of the host rate or something, you get visible tearing that "resolves itself" after around 60-80 ms. That very visible and it's not an authentic experience, to put it mildly.
> 
> As we have established, this can happen in any graphics or text mode; just text mode apps are the ones that tend to update the screen only sporadically, but I'm getting it all the time in the Quake video options menu, for example.
> 
> So you can have a look at some pont if you have capacity, otherwise I will because I find this pretty jarring. It even manifests at a simple command prompt; the blinking of the cursor and a normal typing rate is not enough to trigger new frames at a sufficiently high FPS to prevent this from happening 100% of the time.
> 
> I don't quite get it why the previous "dupe at the rate of the DOS mode" thing had to be removed.

Right. It should follow up at least with the next dupe frame as fast as possible (that is, at the negotiated rate: if throttled, that's the host rate ie: 60 Hz, or if full VFR at the DOS rate). 
I suspect we can likely do away with the constant dupe rate, because VRR monitors are now being properly satisfied when they need CFR rates (when vsync and fullscreen).
More to add to the queue.

kcgen:
@GranMinigun, should we merge this and then punt this very corner-case _"full screen falls back to max-window on output change"_ for the future?

To me, the cleanup and code stream lining is more valuable than this non-showstopping small functional difference.

johnnovak:
> Right. It should follow up at least with the next dupe frame as fast as possible (that is, at the negotiated rate: if throttled, that's the host rate ie: 60 Hz, or if full VFR at the DOS rate).

Well, it definitely doesn't do that 😄 

> I suspect we can likely do away with the constant dupe rate, because VRR monitors are now being properly satisfied when they need CFR rates (when vsync and fullscreen).

Yeah, I just use CFR myself, usually with vsync disabled because it doesn't make much of a difference for the stuff I'm interested in, or with vsync enabled with demos but then I need to set my custom tweaked 70 Hz mode. That works like a dream, though; butter-smooth scrolling in Pinball Dreams and similar 100% synced to 70 Hz games.

> More to add to the queue!

🚀

kcgen:
> Well, it definitely doesn't do that 

Yup! I meant, it's the functionality we want.  (what it currently does is follow up with dupe frames at about 10 fps)

johnnovak:
> > Well, it definitely doesn't do that
> 
> Yup! I meant, it's the functionality we want. (what it currently does is follow up with dupe frames at about 10 fps)

Right okay 😄 I read it as: "this is how it should work, but I'm actually surprised to learn it's buggy" 😅 All clear now!

kcgen:
@GranMinigun - "Properly retaining fullscreen state, I'd say that's out of scope. Can't think immediately which change broke it, but will definitely be addressed in the subsequent PR(s)"

Agree here as well. Looking at the PR's purpose, you've nailed it then some (by eliminating the crashy callback). Let's get this down the chute. :rocket:

PR #2783, opened 2023-08-23
Suppress ALSA API macros PVS Studio warnings

FeralChild64:
The last PR of my PVS Studio crusade - at least for now :)

This time I would like to suppress some warnings, shown on the screenshot below. These are all originating from ALSA headers - the `snd_*` are macros, which PVS Studio does not like. I believe there is nothing sane we can do about them (we probably don't want to alter ALSA headers after all), other than ignoring these.

PVS

### DIFF
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 177
+          MAX_BUGS: 170
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -23,7 +23,7 @@ jobs:
             max_warnings: 20
           - name: MSVC 64-bit
             arch: x64
-            max_warnings: 886
+            max_warnings: 893
 
     steps:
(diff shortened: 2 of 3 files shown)


kcgen (approved):

FeralChild64:
Why the MSVC 64-bit build suddenly started exceeding the warning limit? Did the GitHub silently alter the pipeline, or what?

kcgen:
> Why the MSVC 64-bit build suddenly started exceeding the warning limit? Did the GitHub silently alter the pipeline, or what?

They did. From from VS revision 14.35.xyz to 14.36.abc; looks like patch update. 

Suggest bumping the count to match the new detection quantity, and mention the VS rev number.

FeralChild64:
OK. I still don't like such changes being made without clear warning and explicit consent from the project.

kcgen:
It was an excellent crusade; almost the half of peeveestalia has been conquered 📯 🏴

johnnovak:
> OK. I still don't like such changes being made without clear warning and explicit consent from the project.

Yeah that's a bit dodgy, isn't it...

FeralChild64:
@kcgen Actually, this was just 1/3 (when I started we had 265 warnings, we are down to 170 now; but few warnings were fixed in the meantime by others), and mostly just the low-hanging fruits.

PR #2786, opened 2023-08-25
Update Catmull-Rom shader

GranMinigun:
Now it samples in the correct colour space. Also added it to the list of well-behaving shaders, so no more black halo on the edges of the output. Comparison, before and after:

Screenshot_20230825_114850
Screenshot_20230825_115128


johnnovak (approved):
Great enhancement @GranMinigun, now it's gamma-correct! 😎 

I will experiment with this to use it as the fallback shader for the adaptive CRT shaders instead of `sharp`. Should only be picked up when we have less than 3 pixels per emulated scanline, so it just give us just the right amount of blur, in theory.

PR #2792, opened 2023-08-26
Revert increaseticks sleep to 1ms

### LINKED ISSUE #2790: MSYS2 Windows Build with GCC is very slow

_No response_

Windows 10

_No response_

### Relevant hardware info

_No response_

Performance is so bad it's unplayable in a Windows MSYS2 GCC build.  Clang is fine.  I don't know why.  Reproduced on `main` so it's not my ffmpeg PR.  Maybe I'm missing some important package or something.  I'll investigate later but I'll put this ticket in in case anyone else can reproduce.

Should note this affects debug and release builds.  It's not just a slight FPS drop either.  It runs Doom like a slideshow.  It even hangs on a black screen for 5-10 seconds before showing the title screen.

Task Manager shows a constant 25% CPU (which is a full core of my quad core CPU).  I think something is stalling the main thread.

MSVC builds are also fine on my system.  Only affecting GCC for some reason.

Explain how to reproduce
1. Install MSYS 2.
2. Load the MSYS2 MinGW64 shortcut
3. Compile through Meson and run
4. Start a game (tested with Doom, gave up on Quake and Duke3D as they took too long to load).
5. Very slow 😢 

Compiling with the MSYS2 Clang64 shortcut to get the Clang compiler runs fine.

_No response_

### Your configuration

```ini
I deleted the conf folder and tested with a default config.
```

### Provide a Log

_No response_

- [X] Yes, I agree.

kklobe:
This reverts the aggressive 100us sleep time in `increaseticks()` from #2655. There was no real measurable benefit from it, so let's go back to the old reliable 1ms, but keep the new fractional `ticksDone` logic.

Fixes #2790

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -184,7 +184,7 @@ void increaseticks() { //Make it return ticksRemain and set it in the function a
 
 		static int64_t cumulativeTimeSlept = 0;
 
-		constexpr auto sleepDuration = std::chrono::microseconds(100);
+		constexpr auto sleepDuration = std::chrono::microseconds(1000);
 		std::this_thread::sleep_for(sleepDuration);
 
 		const auto timeslept = GetTicksUsSince(ticksNewUs);


weirddan455 (approved):
Looks good as far as I'm concerned.  This works well on all the builds on my machine across Windows and Linux.

shermp:
Confirmed, Quake working properly with `cycles = max` now.

johnnovak:
Do we _really_ wanna sleep here, as in suspend the thread to preserve CPU power, or are we after accuracy? If the latter, this approach seems to be fine on macOS and Linux as discussed in the related thread, but on recent versions of Windows `sleep(1)` will sleep for a random value from 1ms to 15.625msas that is global scheduling tick rate.

So you cannot rely on `sleep` for timing on Windows, which is reasonably, given it does not aim to be a real-time OS.

Whay you can do is to poll the `QueryPerformanceCounter` system call in a busy loop as th ast guarantees microsecond accuracy.
https://learn.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancecounter?source=recommendations

Maybe the C++ query chrono stuff already maps to that, but it's rather easy to use it, and there are lots of examples of its usage online.

Letting windows sleep for up 15ms when we really just want to wait for 1ms doesn't seem like a desirable solution, plus I'd still ifdef in the 100 nano sleep stuff for non-Windows platforms; if it works well on those, why drop it?

kklobe:
thanks @weirddan455 and @shermp, thanks for taking your time to track this down.

weirddan455:
Technically Linux's sleep can't be relied on if you need absolute accuracy either.  There is a real time fork of Linux but almost no one uses it on desktop PCs.  In practice, it seems to be "accurate enough" for this use case but as I understand it, the thread will sleep until whenever the scheduler decides to run.

There's also the issue of how precise is your hardware's clock exactly.  On x86, there's at least tsc and hpet and Linux decides which one of those to use at boot time.

Anyway, I got confused trying to figure out what this DOSBox function is doing exactly.  It still seems very strange to me that _not sleeping_ makes performance tank.

kklobe:
> Do we _really_ wanna sleep here, as in suspend the thread to preserve CPU power, or are we after accuracy? If the latter, this approach seems to be fine on macOS and Linux as discussed in the related thread, but on recent versions of Windows `sleep(1)` will sleep for a random value from 1ms to 15.625ms as that is global scheduling tick rate. See explanation here: https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/
> 
> So you absolutely cannot rely on `sleep` for timing on Windows, which is reasonable, given it does not aim to be a real-time OS. You need to actively measure time and keep the thread active. I'd even go as far as saying that using sleep for _timing_ is a hack everywhere.

I've gone back and forth on this for the past couple of years - I still have a local branch with a precise busy-wait that I try out from time to time, but do we want to peg out a core while sitting at `C:>`?

I'm personally in favor of a busy-wait loop with `std::this_thread::yield()`.

> 
> What you can do is to poll the `QueryPerformanceCounter` system call in a busy loop as that guarantees microsecond accuracy. https://learn.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancecounter?source=recommendations
> 
> Maybe the C++ query chrono stuff already maps to that, but it's rather easy to use it, and there are lots of examples of its usage online. I'd suggest a custom wait for microsecondsfunctions that wraps the platform specific ifdefs
> 
> Letting Windows sleep for up 15ms when we really just want to wait for 1ms doesn't seem like a desirable solution, plus I'd still ifdef in the 100 nano sleep stuff for non-Windows platforms; if it works well on those, why drop it?

Because any of the other platforms could change their scheduling algorithm at any time, and I don't think a bunch of empirically-derived `#ifdef`'d magic sleep numbers is any more attractive.

kcgen:
> Letting Windows sleep for up 15ms when we really just want to wait for 1ms doesn't seem like a desirable solution

That blog post is interesting (and I sure hope chrono isn't sleeping for ~15ms!). 

[The code](https://github.com/randomascii/blogstuff/blob/main/timer_interval/measure_interval.cpp) barely looks like C++, it's making almost 100% Windows-specific calls :nauseated_face: without any use of `std::chrono`.

When we're running at `cycles = max`, DOSBox's auto cycle algorithm tries to back off cycles such that we can still fit all of the emulated 1ms tick's worth of work inside 1ms worth of "host time", leading to that comparison `if (ticksNew <= ticksLast)`, inside of which we do the `sleep`, being true for some percent of passes through the normal loop.

The longer we sleep, the lower our "duty cycle", or the less time we get to spend "running flat out".

If we're sleeping 15ms, then benchmark scores should take a big hit versus the prior 100us or even 1ms sleep.

If you guys can compare some quake bench timedemo scores: https://anttipeltola.eu/misc/files/DOSBENCH.ZIP, it should be pretty obvious if we're affected by 15ms sleep inside our ticks.

weirddan455:
>  That blog post is interesting (and I sure hope chrono isn't sleeping for ~15ms!).

I measured GCC on Windows sleeping for up to 15ms (when requested time was 1ms).  GCC's implementation of `std::thread::sleep_for` is using a different sleep function (`nanosleep`) than Clang or MSVC and sleep times varied anywhere from 3ms to 15ms.

On Clang or MSVC, it was sleeping pretty consistently between 1 and 2ms.

kklobe:
I am testing out my old precise busy-wait branch, and am noticing almost-perfect consistency on my QTD as well as Future Crew's Panic, the first "SCROLLERS SUCK" sequence (I get zero glitching after 10 fresh Staging invocations).

Are we interested in going down the near-perfect-accuracy-while-pegging-a-CPU-core path?

kcgen:
> I am testing out my old precise busy-wait branch, and am noticing almost-perfect consistency on my QTD as well as Future Crew's Panic, the first "SCROLLERS SUCK" sequence (I get zero glitching after 10 fresh Staging invocations).
> 
> Are we interested in going down the near-perfect-accuracy-while-pegging-a-CPU-core path?

Any hope that `yield` can save us?

If we're emulating some old mid 80s EGA Sierra game that only needs `cycles = 600`, it should would be nice to avoid spinning up the host CPU and blowing hot air :-)

But if we're playing `Screamer 2` or some other protected mode game and running `cycles = max` or some other "give me everything you've got", then absolutely: busy-wait me and give me as much duty cycle as possible!

kklobe:
> > I am testing out my old precise busy-wait branch, and am noticing almost-perfect consistency on my QTD as well as Future Crew's Panic, the first "SCROLLERS SUCK" sequence (I get zero glitching after 10 fresh Staging invocations).
> > Are we interested in going down the near-perfect-accuracy-while-pegging-a-CPU-core path?
> 
> Any hope that `yield` can save us?

Unfortunately it seems to have no effect on CPU usage, it just lets the scheduler rest a little bit.

> 
> If we're emulating some old mid 80s EGA Sierra game that only needs `cycles = 600`, it would be nice to avoid spinning up the host CPU and blowing hot air :-)

Exactly my hesitation around doing this :)

johnnovak:
Yeah, if timing really matters, unattractive ifdefs hidden in a function is the way. Code has to do its job, first and foremost.

I'm not that close to this code, but at a conceptual level, having accurate time measurement functionality do wildly fluctuating random shit depending on OS _and_ compiler variant sounds like really bad design and asking for trouble. Like auto max cycles behaving differently on different machines.

Ultimately, I don' t think accurate waiting and sleep meshes well. So we either a) rework stuff so it doesn't rely on sleep/busy-waiting, but does the timing adaptively somehow by querying performance timers,  b) take the busy wait hit where it really matters and query performance counters, c) rely on events/notifications/signals to wake things up from sleep at precise intervals.

Even that blog post I linked started with this: sleep for timing is never a good idea... but you have to, here are the issues on Windows.

kcgen:
@weirddan455 , curious if swapping chrono::sleep for `SDL_Delay(1)` is also sleeping for 15ms on your Windows system?

kklobe:
> Edit: if it turns our we can't trust `std::chrono`, the first thing to try would be going back to `SDL_Delay(1)`, which is a bunch of old C-calls in very unattractive empirically-derived OS-specific #ifdefs 😅

That's supposed to be what `std::this_thread::sleep_for()` does! We can easily inspect the code to see what's it doing.

But, I thought everyone said things are working back to normal after this revert?

kklobe:
> It would be interesting to see how other cycle-accurate emulators like WinUAE and VICE deal with this, or audio apps, where timing is crucial. I'd bet good money on it it's whacky ifdef'd syscall hell that works splendidly. 😅

I guess what I'm trying to say is that the gist of this revert is that sleeping for 100us offered almost zero benefit (it drops CPU usage on my M1 Mac), so let's go back to what we know works, the good old 1ms, for the time being, and then step back and assess a better path forward, rather than continuing to throw hacky whack-a-mole sleep durations at the problem.

weirddan455:
> @weirddan455 , curious if swapping chrono::sleep for `SDL_Delay(1)` is also sleeping for 15ms on your Windows system?

I can test a bit later (I'm in Linux now) but all that function does is call `Sleep(1)`  Sleep with a capital S being the Windows specific sleep that takes milliseconds.

https://github.com/libsdl-org/SDL/blob/031912c4b6c5db80b443f04aa56fec3e4e645153/src/timer/windows/SDL_systimer.c#L139-L166

kcgen:
> > It would be interesting to see how other cycle-accurate emulators like WinUAE and VICE deal with this, or audio apps, where timing is crucial. I'd bet good money on it it's whacky ifdef'd syscall hell that works splendidly. 😅
> 
> I guess what I'm trying to say is that the gist of this revert is that sleeping for 100us offered almost zero benefit (it drops CPU usage on my M1 Mac), so let's go back to what we know works, the good old 1ms, for the time being

With @johnnovak finding and @weirddan455 measuring that Win10+ is actually sleeping for 15ms (well, with GCC), It might make sense to revert all the way back to SDL_Delay if it proves to "fix" GCC, too. 

> and then step back and assess a better path forward, rather than continuing to throw hacky whack-a-mole sleep durations at the problem.

100% agree, @kklobe 👍

johnnovak:
what I'm trying to say is that the gist of this revert is that sleeping for 100us offered almost zero benefit (it drops CPU usage on my M1 Mac), so let's go back to what we know works, the good old 1ms, for the time being
> 
> With @johnnovak finding and @weirddan455 measuring that Win10+ is actually sleeping for 15ms (well, with GCC), It might make sense to revert all the way back to SDL_Delay if it proves to "fix" GCC, too.
> 
> > and then step back and assess a better path forward, rather than continuing to throw hacky whack-a-mole sleep durations at the problem.
> 
> 100% agree, @kklobe 👍



> > It would be interesting to see how other cycle-accurate emulators like WinUAE and VICE deal with this, or audio apps, where timing is crucial. I'd bet good money on it it's whacky ifdef'd syscall hell that works splendidly. 😅
> 
> I guess what I'm trying to say is that the gist of this revert is that sleeping for 100us offered almost zero benefit (it drops CPU usage on my M1 Mac), so let's go back to what we know works, the good old 1ms, for the time being, and then step back and assess a better path forward, rather than continuing to throw hacky whack-a-mole sleep durations at the problem.

Oh yeah, absolutely for this case. I was a bit jumping forward and started discussing our general approach to accurate timing.

kklobe:
> With @johnnovak finding and @weirddan455 measuring that Win10+ is actually sleeping for 15ms (well, with GCC), It might make sense to revert all the way back to SDL_Delay if it proves to "fix" GCC, too.

I would argue for finding out what GCC and/or their implementation of `sleep_for` is doing so horribly wrong, if it indeed is significantly worse than `SDL_Delay`. There are only so many ways to "sleep" on a given platform, SDL_Delay isn't doing anything magic.

shermp (approved):

weirddan455:
> I would argue for finding out what GCC and/or their implementation of `sleep_for` is doing so horribly wrong, if it indeed is significantly worse than `SDL_Delay`. There are only so many ways to "sleep" on a given platform, SDL_Delay isn't doing anything magic.

As I wrote in the bug report, GCC's implementation of `sleep_for` calls the `nanosleep` function.  If you call `nanosleep` directly in Clang, you get the same behavior. `nanosleep` isn't available in MSVC.  It's an MSYS2 implementation of a POSIX function.

I haven't tested `Sleep` or `SDL_Delay` (which just calls `Sleep`) but I will shortly.

johnnovak:
> > With @johnnovak finding and @weirddan455 measuring that Win10+ is actually sleeping for 15ms (well, with GCC), It might make sense to revert all the way back to SDL_Delay if it proves to "fix" GCC, too.
> 
> I would argue for finding out what GCC and/or their implementation of `sleep_for` is doing so horribly wrong, if it indeed is significantly worse than `SDL_Delay`. There are only so many ways to "sleep" on a given platform, SDL_Delay isn't doing anything magic.

Yeah, I don't really like using `SDL_Sleep` for the same reason I don't like using  the SDL threading stuff.

I'd ideally like to see SDL only being used to help us with the murky and completely uninteresting and fiddly stuff of normalising cross-platform windows management, context creation, window system related event handling and keyboard handling. Nothing more.

I'm not in love with the audio side of things in SDL. It does the job for now, but in the future I'd really like to move to RtAudio, just like I'm planning to move to RtMidi for MIDI stuff. With RtAudio, people like me could use their ASIO supported pro audio cards for ultra-low latencies (sub 5-8ms rage) on Mac and Windows.

kklobe:
> As I wrote in the bug report, GCC's implementation of `sleep_for` calls the `nanosleep` function. If you call `nanosleep` directly in Clang, you get the same behavior. `nanosleep` isn't available in MSVC. It's an MSYS2 implementation of a POSIX function.

Right, and what is that `nanosleep` implementation doing, is all I was getting at.

kcgen:
Fully agree we should use C++17 language and only carve out exceptions when needed. The more specific the carve-our, the better (Windows && GCC, hopefully in this case). 

```c
int nanosleep(const struct timespec *req, struct timespec *rem)
{
	int ret = syscall2(int, SYS_NANOSLEEP, req, rem);
	if (ret < 0) {
		errno = -ret;
		return -1;
	}

	return ret;
}
```c
#define SYS_NANOSLEEP  6

#define syscall2(rettype, no, a1, a2)				(rettype) _syscall5(no, (void *) a1, (void *) a2, NULL, NULL, NULL)
```

```c
#define syscall5(rettype, no, a1, a2, a3, a4, a5)	(rettype) _syscall5(no, (void *) a1, (void *) a2, (void *) a3, (void *) a4, (void *) a5)
```

And finally, into ASM land:

```c
#define _syscall5(type,name,atype,a,btype,b,ctype,c,dtype,d,etype,e) \
type name (atype a,btype b,ctype c,dtype d,etype e) \
{ \
long __res; \
__asm__ __volatile__ ("int $0x80" \
	: "=a" (__res) \
	: "0" (SYS_##name),"b" ((long)(a)),"c" ((long)(b)), \
	  "d" ((long)(c)),"S" ((long)(d)),"D" ((long)(e))); \
	return (type) __res; \
}
```

https://github.com/Maximus5/Cygwin-MSYS2/blob/master/newlib/libc/sys/phoenix/nanosleep.c

weirddan455:
I've tested now in GCC for Windows.  `Sleep(1)` and `SDL_Delay(1)` both sleep for between 1 and 2ms.  About the same as using the C++ `sleep_for` in Cland and MSVC.

You could probably just throw an `#ifdef _WIN32 Sleep(1)` in there and be done with it for now, though I'm curious what the actual performance implications are.  I'll run that Quake benchmark next.

>  Right, and what is that nanosleep implementation doing, is all I was getting at.

It looks like it does a bunch of nonsense and then calls either `Sleep` or `WaitSingleObject`.  It first has to convert from nanoseconds to milliseconds, which can be 0 if the value is too small.  In that case, it calls `Sleep(0)` which explains why it was immediately returning at 100 microseconds.

It's part of the pthread library and I don't really feel like looking through all the logic but here's links to the code.  The bottom function is where it calls `Sleep`.

https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-libraries/winpthreads/src/nanosleep.c

https://github.com/Alexpux/mingw-w64/blob/d0d7f784833bbb0b2d279310ddc6afb52fe47a46/mingw-w64-libraries/winpthreads/src/thread.c#L640-L662

weirddan455:
@kcgen I don't think it's making a real Windows syscall.  If you look at the syscall.h it's all Unix type functions.  Like SYS_FORK for example, which I am pretty sure does not have a direct Windows mapping.  On Linux, nanosleep is indeed just a syscall and the work is done in the kernel.

https://github.com/Maximus5/Cygwin-MSYS2/blob/master/newlib/libc/sys/phoenix/syscall.h

If I had to guess, this is part of some "virtual syscall" type system that gets mapped to Windows equivalents elsewhere in the code.

kcgen:
>  It first has to convert from nanoseconds to milliseconds, which can be 0 if the value is too small. In that case, it calls Sleep(0) which explains why it was immediately returning at 100 microseconds.

👌 nice find, @weirddan455!

kklobe:
> > It first has to convert from nanoseconds to milliseconds, which can be 0 if the value is too small. In that case, it calls Sleep(0) which explains why it was immediately returning at 100 microseconds.
> 
> 👌 nice find, @weirddan455!

Yes, very nice find!. On Windows, I would argue that should always round up to Sleep(1).

johnnovak:
> > > It first has to convert from nanoseconds to milliseconds, which can be 0 if the value is too small. In that case, it calls Sleep(0) which explains why it was immediately returning at 100 microseconds.
> > 
> > 
> > 👌 nice find, @weirddan455!
> 
> Yes, very nice find!. On Windows, I would argue that should always round up to Sleep(1).

Yeah, if you read a few more blog posts of the guy who wrote the article I linked, he explains in paintstaking detail why you should never ever call `Sleep(0)` on Windows. It's evil 🥶

weirddan455:
I've run the Quake benchmark on Linux with both C++ sleep and busy wait.  Here are some numbers.  Windows performance was very similar, at least around the 1 millisecond range (didn't do the full suite of tests on Windows).  There's not a whole lot of performance difference here.  Anything from 100 microseconds to 15 milliseconds seems fine.  But don't you dare wait 16 milliseconds!

no sleep/wait
gave up waiting for it to load (very slow)

busy wait 1 microsecond
78fps - 320 X 200
20fps - 800 X 600

sleep for 1 microsecond
131fps - 320 X 200
42fps - 800 X 600

sleep 100 microseconds
133fps - 320X200
43fps - 800X600

busy wait 100 microseconds
135fps - 320X200
44fps - 800X600

sleep 1 millisecond
136fps - 320X200
44fps - 800X600

busy wait 1 millisecond
135 fps - 320X200
44 fps - 800X600

sleep 15 milliseconds
135fps - 320 X 200
43fps - 800 X 600

busy wait 15 milliseconds
136 fps - 320 X 200
42 fps - 800X600

sleep/wait anything 16milliseconds or above
gave up waiting on it to load(very slow)

johnnovak:
> sleep/wait anything 16milliseconds or above
> gave up waiting on it to load(very slow)

Uh... ok. So it works in practice. But also seems fragile... I guess in real life those longer than 16ms delays only happen sporadically if at all. But technically a sleep of 1ms gives you no guarantees; sometime it might wait 50ms, who knows. But maybe it gets averaged out or something by the algo.

kklobe:
The original motivation for #2655 was not to increase raw performance, but to (hopefully) allow higher accuracy by allowing `PIC_RunQueue` to execute more often (see `Normal_Loop()` above `increaseticks()`, reducing the chance that a PIC event scheduled for a given tick would be late and cause an audio or video glitch. In many places throughout the codebase, `PIC_AddEvent()` is called with a precise non-whole-number delay, sometimes less than a millisecond. Given the OS overhead, sleeping for 1ms could completely run over a whole emulator tick.

Realistically, this probably never makes enough difference to be perceptible in a game.

kklobe:
> > sleep/wait anything 16milliseconds or above
> > gave up waiting on it to load(very slow)
> 
> Uh... ok. So it works in practice. But also seems fragile... I guess in real life those longer than 16ms delays only happen sporadically if at all. But technically a sleep of 1ms gives you no guarantees; sometime it might wait 50ms, who knows. But maybe it gets averaged out or something by the algo.

It might be interesting to add a `LOG_WARN` if the actual slept time is, oh, > 5ms, and see how often this really happens when requesting 1ms.

johnnovak:
> > > sleep/wait anything 16milliseconds or above
> > > gave up waiting on it to load(very slow)
> > 
> > 
> > Uh... ok. So it works in practice. But also seems fragile... I guess in real life those longer than 16ms delays only happen sporadically if at all. But technically a sleep of 1ms gives you no guarantees; sometime it might wait 50ms, who knows. But maybe it gets averaged out or something by the algo.
> 
> It might be interesting to add a `LOG_WARN` if the actual slept time is, oh, > 5ms, and see how often this really happens when requesting 1ms.

Yeah, that would be interesting.

johnnovak:
Merging for now; further exploration of the sleep topic can be done separately.

kcgen:
> further exploration of the sleep topic

PR #2794, opened 2023-08-27
Minor shell cleanup

MeAreJeenius:
These changes broadly deal with 2 things: shell history and the environment functions.

I moved the shell history setup functions to its own class. At first this meant that it would be called upon construction of every shell, including the ones made as a part of the registration of COMMAND.COM. Looking at the original PR it seems that this was the reason for having them in function in the first place. However making the class a singleton avoids this problem entirely, since it will now all be referring to the same instance. Note that this slightly changes behavior by recording the commands from all sub-shells instead of only the first.

Most of the changes have to do with the environment functions. I changed the result of an access to be the return values instead of out-parameters. I also made it so that when accessing an environment variable, it just gives back the value instead of the full string, which had necessitated each caller to manually parse the string to extract the value themselves. In addition I replaced the function that returned the number of variables, and the one that returned the variable at an "index". It's now a single function that just returns the strings for all of the variables. 

I also moved the functions to the program segment prefix since they're functions that operate on the PSP. In the process, I changed the programs which interact with the environment to set their environment variables to their parent's PSP instead of always the first_shell.

I originally wanted to get rid of the global first_shell instance altogether, but there are a couple of places that still use the first_shell environment in ways they probably shouldn't, but that would require some more work to change. These are in the autoexec code and the render code.

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -27,6 +27,7 @@
 #include <algorithm>
 #include <array>
 #include <cstddef>
+#include <optional>
 #include <string>
 #include <type_traits>
 
@@ -351,17 +352,34 @@ public:
 	MemStruct(uint16_t seg, uint16_t off) : pt(PhysicalMake(seg, off)) {}
 	MemStruct(RealPt addr) : pt(RealToPhysical(addr)) {}
 
-	void SetPt(uint16_t seg) { pt = PhysicalMake(seg, 0); }
+	void SetPt(uint16_t seg)
+	{
+		pt = PhysicalMake(seg, 0);
+	}
 
 protected:
-	PhysPt pt = 0;
+	PhysPt pt    = 0;
+	~MemStruct() = default;
+};
+
+class Environment {
+public:
+	virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
+
+	virtual ~Environment() = default;
+
+protected:
+	Environment()                              = default;
+	Environment(const Environment&)            = default;
+	Environment& operator=(const Environment&) = default;
+	Environment(Environment&&)                 = default;
+	Environment& operator=(Environment&&)      = default;
 };
 
 /* Program Segment Prefix */
-
-class DOS_PSP final : public MemStruct {
+class DOS_PSP final : public MemStruct, public Environment {
 public:
-	DOS_PSP(uint16_t segment) : seg(segment) { SetPt(seg); }
+	DOS_PSP(uint16_t segment) : seg(segment) { SetPt(seg); }	
 
 	void MakeNew(uint16_t mem_size);
 
@@ -413,6 +431,10 @@ public:
 	void SetFCB1(RealPt src);
 	void SetFCB2(RealPt src);
 	void SetCommandTail(RealPt src);
+
+	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
+	std::vector<std::string> GetAllEnvVars() const;
+	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 private:
 	#ifdef _MSC_VER
--- a/include/shell.h
+++ b/include/shell.h
@@ -53,21 +53,9 @@ public:
 	virtual ~ByteReader()                    = default;
 };
 
-class HostShell {
-public:
-	virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
-
-	HostShell()                            = default;
-	HostShell(const HostShell&)            = delete;
-	HostShell& operator=(const HostShell&) = delete;
-	HostShell(HostShell&&)                 = delete;
-	HostShell& operator=(HostShell&&)      = delete;
-	virtual ~HostShell()                   = default;
-};
-
 class BatchFile {
 public:
-	BatchFile(const HostShell& host, std::unique_ptr<ByteReader> input_reader,
+	BatchFile(const Environment& host, std::unique_ptr<ByteReader> input_reader,
 	          std::string_view entered_name, std::string_view cmd_line,
 	          bool echo_on);
 	BatchFile(const BatchFile&)            = delete;
@@ -86,7 +74,7 @@ private:
 	[[nodiscard]] std::string ExpandedBatchLine(std::string_view line) const;
 	[[nodiscard]] std::string GetLine();
 
-	const HostShell& shell;
+	const Environment& shell;
 	CommandLine cmd;
 	std::unique_ptr<ByteReader> reader;
 	bool echo;
@@ -122,7 +110,7 @@ private:
 	bool secure_mode;
 };
 
-class DOS_Shell : public Program, public HostShell {
+class DOS_Shell : public Program {
 private:
 	void PrintHelpForCommands(MoreOutputStrings& output, HELP_Filter req_filter);
 	void AddShellCmdsToHelpList();
@@ -162,7 +150,7 @@ public:
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
+	std::optional<std::string> GetEnvStr(std::string_view entry) const;
 	std::vector<std::string> GetAllEnvVars() const;
 	bool SetEnv(std::string_view entry, std::string_view new_string);
 
--- a/include/shell.h
+++ b/include/shell.h
@@ -55,7 +55,7 @@ public:
 
 class HostShell {
 public:
-	virtual bool GetEnvStr(const char* entry, std::string& result) const = 0;
+	virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
 
 	HostShell()                            = default;
 	HostShell(const HostShell&)            = delete;
@@ -162,7 +162,7 @@ public:
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	bool GetEnvStr(const char* entry, std::string& result) const override;
+	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
 	[[nodiscard]] Bitu GetEnvCount() const;
 	bool SetEnv(std::string_view entry, std::string_view new_string);
--- a/include/shell.h
+++ b/include/shell.h
@@ -163,8 +163,7 @@ public:
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
 	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
-	bool GetEnvNum(Bitu num, std::string& result) const;
-	[[nodiscard]] Bitu GetEnvCount() const;
+	std::vector<std::string> GetAllEnvVars() const;
 	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 	/* Commands */
--- a/include/shell.h
+++ b/include/shell.h
@@ -165,7 +165,7 @@ public:
 	bool GetEnvStr(const char* entry, std::string& result) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
 	[[nodiscard]] Bitu GetEnvCount() const;
-	bool SetEnv(const char* entry, const char* new_string);
+	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 	/* Commands */
 	void CMD_HELP(char* args);
--- a/include/shell.h
+++ b/include/shell.h
@@ -150,8 +150,8 @@ public:
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	std::optional<std::string> GetEnvStr(std::string_view entry) const;
-	std::vector<std::string> GetAllEnvVars() const;
+	// HACK: Don't use in new code
+	// TODO: Remove the call to this function from autoexec
 	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 	/* Commands */
--- a/include/shell.h
+++ b/include/shell.h
@@ -103,6 +103,25 @@ struct SHELL_Cmd {
 	HELP_Category category = HELP_Category::Misc;
 };
 
+class ShellHistory {
+public:
+	static ShellHistory& Instance();
+	const std::vector<std::string>& GetCommands() const;
+	void Append(std::string);
+
+	~ShellHistory();
+	ShellHistory(const ShellHistory&)            = delete;
+	ShellHistory& operator=(const ShellHistory&) = delete;
+	ShellHistory(ShellHistory&&)                 = delete;
+	ShellHistory& operator=(ShellHistory&&)      = delete;
+
+private:
+	ShellHistory();
+	std::vector<std::string> commands{};
+	std_fs::path path;
+	bool secure_mode;
+};
+
 class DOS_Shell : public Program, public HostShell {
 private:
 	void PrintHelpForCommands(MoreOutputStrings& output, HELP_Filter req_filter);
@@ -113,15 +132,17 @@ private:
 	[[nodiscard]] std::string Which(std::string_view name) const;
 
 	friend class AutoexecEditor;
-	std::vector<std::string> history{};
-	std::stack<BatchFile> batchfiles{};
 
+	ShellHistory& history = ShellHistory::Instance();
+	std::stack<BatchFile> batchfiles{};
+	uint16_t input_handle                  = STDIN;
+	bool call                              = false;
 	bool exit_cmd_called                   = false;
 	static inline bool help_list_populated = false;
 
 public:
 	DOS_Shell();
-	~DOS_Shell() override = default;
+	~DOS_Shell() override                  = default;
 	DOS_Shell(const DOS_Shell&)            = delete; // prevent copy
 	DOS_Shell& operator=(const DOS_Shell&) = delete; // prevent assignment
 	void Run() override;
@@ -140,9 +161,6 @@ public:
 	virtual bool ExecuteShellCommand(const char* const name, char* arguments);
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
-
-	void ReadShellHistory();
-	void WriteShellHistory();
 
 	bool GetEnvStr(const char* entry, std::string& result) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
@@ -182,10 +200,7 @@ public:
 	void CMD_VOL(char* args);
 	void CMD_MOVE(char* args);
 
-	/* The shell's variables */
-	uint16_t input_handle         = 0;
-	bool echo                     = false;
-	bool call                     = false;
+	bool echo = true;
 };
 
 std::tuple<std::string, std::string, std::string> parse_drive_conf(
--- a/src/dos/dos_classes.cpp
+++ b/src/dos/dos_classes.cpp
@@ -26,6 +26,7 @@
 #include <cstring>
 
 #include "mem.h"
+#include "string_utils.h"
 #include "support.h"
 
 static void dos_memset(PhysPt addr, uint8_t val, size_t n)
@@ -305,6 +306,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
 		SSET_DWORD(sPSP, file_table, data);
 	}
 	SSET_WORD(sPSP, max_files, file_num);
+	return true;
+}
+
+static constexpr auto bytes_to_read = 1024;
+
+std::optional<std::string> DOS_PSP::GetEnvStr(const std::string_view entry) const
+{
+	/* Walk through the internal environment and see for a match */
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+
+	char env_string[bytes_to_read + 1];
+	if (entry.empty()) {
+		return {};
+	}
+
+	for (;;) {
+		MEM_StrCopy(env_read, env_string, bytes_to_read);
+		if (!env_string[0]) {
+			return {};
+		}
+		env_read += (PhysPt)(safe_strlen(env_string) + 1);
+		char* equal = strchr(env_string, '=');
+		if (!equal) {
+			continue;
+		}
+		/* replace the = with \0 to get the length */
+		*equal = '\0';
+		if (strlen(env_string) != entry.size()) {
+			continue;
+		}
+		if (!iequals(entry, env_string)) {
+			continue;
+		}
+
+		return env_string + entry.size() + sizeof('=');
+	}
+}
+
+std::vector<std::string> DOS_PSP::GetAllEnvVars() const
+{
+	auto all_env_vars = std::vector<std::string>{};
+
+	char env_string[bytes_to_read + 1];
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+	for (;;) {
+		MEM_StrCopy(env_read, env_string, bytes_to_read);
+		if (!env_string[0]) {
+			return all_env_vars;
+		}
+		all_env_vars.emplace_back(env_string);
+		env_read += (PhysPt)(safe_strlen(env_string) + 1);
+	}
+}
+
+bool DOS_PSP::SetEnv(std::string_view entry, std::string_view new_string)
+{
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+
+	// Get size of environment.
+	DOS_MCB mcb(GetEnvironment() - 1);
+	uint16_t envsize = mcb.GetSize() * 16;
+
+	PhysPt env_write                   = env_read;
+	PhysPt env_write_start             = env_read;
+	char env_string[bytes_to_read + 1] = {0};
+	const auto entry_length            = entry.size();
+	do {
+		MEM_StrCopy(env_read, env_string, bytes_to_read);
+		if (!env_string[0]) {
+			break;
+		}
+		env_read += (PhysPt)(safe_strlen(env_string) + 1);
+		if (!strchr(env_string, '=')) {
+			continue; /* Remove corrupt entry? */
+		}
+		if (iequals(entry,
+		            std::string_view(env_string).substr(0, entry_length)) &&
+		    env_string[entry_length] == '=') {
+			continue;
+		}
+		MEM_BlockWrite(env_write,
+		               env_string,
+		               (Bitu)(safe_strlen(env_string) + 1));
+		env_write += (PhysPt)(safe_strlen(env_string) + 1);
+	} while (true);
+	/* TODO Maybe save the program name sometime. not really needed though */
+	/* Save the new entry */
+
+	// ensure room
+	if (envsize <= (env_write - env_write_start) + entry.size() + 1 +
+	                       new_string.size() + 2) {
+		return false;
+	}
+
+	if (!new_string.empty()) {
+		std::string bigentry(entry);
+		for (std::string::iterator it = bigentry.begin();
+		     it != bigentry.end();
+		     ++it) {
+			*it = toupper(*it);
+		}
+		snprintf(env_string,
+		         bytes_to_read + 1,
+		         "%s=%s",
+		         bigentry.c_str(),
+		         std::string(new_string).c_str());
+		MEM_BlockWrite(env_write,
+		               env_string,
+		               (Bitu)(safe_strlen(env_string) + 1));
+		env_write += (PhysPt)(safe_strlen(env_string) + 1);
+	}
+	/* Clear out the final piece of the environment */
+	mem_writeb(env_write, 0);
 	return true;
 }
 
--- a/src/dos/program_mount.cpp
+++ b/src/dos/program_mount.cpp
@@ -59,7 +59,8 @@ void MOUNT::Move_Z(char new_z)
 		/* Update environment */
 		std::string line = "";
 		std::string tempenv = {new_drive_z, ':', '\\'};
-		if (first_shell->GetEnvStr("PATH",line)) {
+		if (const auto result = first_shell->GetEnvStr("PATH")) {
+			line = *result;
 			std::string::size_type idx = line.find('=');
 			std::string value = line.substr(idx +1 , std::string::npos);
 			while ( (idx = value.find("Z:\\")) != std::string::npos ||
--- a/src/dos/program_mount.cpp
+++ b/src/dos/program_mount.cpp
@@ -69,9 +69,15 @@ void MOUNT::Move_Z(char new_z)
 		} else {
 			value = tempenv;
 		}
-		first_shell->SetEnv("PATH", value);
+
 		tempenv += "COMMAND.COM";
-		first_shell->SetEnv("COMSPEC", tempenv);
+
+		auto psp_copy = DOS_PSP(psp->GetSegment());
+		while (psp_copy.GetSegment() != DOS_PSP::rootpsp) {
+			psp_copy = DOS_PSP(psp_copy.GetParent());
+			psp_copy.SetEnv("PATH", value);
+			psp_copy.SetEnv("COMSPEC", tempenv);
+		}
 
 		/* Change the active drive */
 		if (DOS_GetDefaultDrive() == 25)
--- a/src/dos/program_mount.cpp
+++ b/src/dos/program_mount.cpp
@@ -59,7 +59,7 @@ void MOUNT::Move_Z(char new_z)
 		/* Update environment */
 		std::string value   = "";
 		std::string tempenv = {new_drive_z, ':', '\\'};
-		if (const auto result = first_shell->GetEnvStr("PATH")) {
+		if (const auto result = psp->GetEnvStr("PATH")) {
 			value = *result;
 			std::string::size_type idx;
 			while ((idx = value.find("Z:\\")) != std::string::npos ||
--- a/src/dos/program_mount.cpp
+++ b/src/dos/program_mount.cpp
@@ -57,21 +57,21 @@ void MOUNT::Move_Z(char new_z)
 		Drives[25]      = nullptr;
 		if (!first_shell) return; //Should not be possible
 		/* Update environment */
-		std::string line = "";
+		std::string value   = "";
 		std::string tempenv = {new_drive_z, ':', '\\'};
 		if (const auto result = first_shell->GetEnvStr("PATH")) {
-			line = *result;
-			std::string::size_type idx = line.find('=');
-			std::string value = line.substr(idx +1 , std::string::npos);
-			while ( (idx = value.find("Z:\\")) != std::string::npos ||
-				(idx = value.find("z:\\")) != std::string::npos  )
-				value.replace(idx,3,tempenv);
-			line = std::move(value);
+			value = *result;
+			std::string::size_type idx;
+			while ((idx = value.find("Z:\\")) != std::string::npos ||
+			       (idx = value.find("z:\\")) != std::string::npos) {
+				value.replace(idx, 3, tempenv);
+			}
+		} else {
+			value = tempenv;
 		}
-		if (!line.size()) line = tempenv;
-		first_shell->SetEnv("PATH",line.c_str());
+		first_shell->SetEnv("PATH", value);
 		tempenv += "COMMAND.COM";
-		first_shell->SetEnv("COMSPEC",tempenv.c_str());
+		first_shell->SetEnv("COMSPEC", tempenv);
 
 		/* Change the active drive */
 		if (DOS_GetDefaultDrive() == 25)
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -281,13 +281,10 @@ void DOS_Shell::ParseLine(char *line)
 			             "pipe%d.tmp",
 			             get_tick_random_number());
 		} else {
-			const auto idx   = env_temp_path.find('=');
-			std::string temp = env_temp_path.substr(idx + 1,
-			                                        std::string::npos);
-			if (DOS_GetFileAttr(temp.c_str(), &fattr) &&
+			if (DOS_GetFileAttr(env_temp_path.c_str(), &fattr) &&
 			    fattr & DOS_ATTR_DIRECTORY)
 				safe_sprintf(pipe_tempfile, "%s\\pipe%d.tmp",
-				             temp.c_str(),
+				             env_temp_path.c_str(),
 				             get_tick_random_number());
 			else
 				safe_sprintf(pipe_tempfile, "pipe%d.tmp",
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -268,9 +268,15 @@ void DOS_Shell::ParseLine(char *line)
 	char pipe_tempfile[270]; // Piping requires the use of a temporary file
 	uint16_t fattr;
 	if (pipe_file.length()) {
+		auto result = GetEnvStr("TEMP");
+		if (!result) {
+			result = GetEnvStr("TMP");
+		}
 		std::string env_temp_path = {};
-		if (!GetEnvStr("TEMP", env_temp_path) &&
-		    !GetEnvStr("TMP", env_temp_path)) {
+		if (result) {
+			env_temp_path = *result;
+		}
+		if (env_temp_path.empty()) {
 			safe_sprintf(pipe_tempfile,
 			             "pipe%d.tmp",
 			             get_tick_random_number());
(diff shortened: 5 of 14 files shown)


johnnovak:
Thanks for this @MeAreJeenius, this is great! I'll review in detail, just some quick thoughts:

> Note that this slightly changes behavior by recording the commands from all sub-shells instead of only the first.

Actually, that's the preferred behaviour IMO, so this is an improvement for sure.

> I changed the result of an access to be the return values instead of out-parameters. I also made it so that when accessing an environment variable, it just gives back the value instead of the full string, which had necessitated each caller to manually parse the string to extract the value themselves.

That's a great improvement too, nice one!

> In the process, I changed the programs which interact with the environment to set their environment variables to their parent's PSP instead of always the first_shell.

I assume this is how real DOS works too, in which case this is another good fix. Have you been able to confirm this reflects now actual DOS accurately, btw?

> I originally wanted to get rid of the global first_shell instance altogether, but there are a couple of places that still use the first_shell environment in ways they probably shouldn't, but that would require some more work to change. These are in the autoexec code and the render code.

Yeah, one step at a time 😅

include/shell.h
```diff
@@ -103,6 +103,25 @@ struct SHELL_Cmd {
 	HELP_Category category = HELP_Category::Misc;
 };
 
+class ShellHistory {
+public:
+	[[nodiscard]] static ShellHistory& Instance();
+	[[nodiscard]] const std::vector<std::string>& Commands() const;
```
johnnovak, line 97: Honestly, I'd drop the `[[nodiscards]]`. The bugs that result from not using return values are very obvious, I don't think this add any extra benefit that would warrant the extra noise. I guess I don't wanna see them popping up everywhere; `const`s are enough... 😅

johnnovak: `GetInstance()` and `GetCommands()`?

include/shell.h
```diff
@@ -180,10 +199,7 @@ class DOS_Shell : public Program, public HostShell {
 	void CMD_VOL(char* args);
 	void CMD_MOVE(char* args);
 
-	/* The shell's variables */
-	uint16_t input_handle         = 0;
-	bool echo                     = false;
-	bool call                     = false;
+	bool echo = true;
```
johnnovak, line 188: Does this mean we're changing the default behaviour? If so, why? Was our previous behaviour incorrect?

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include "shell.h"
+
+#include <fstream>
+
+#include "control.h"
+#include "dosbox.h"
+#include "fs_utils.h"
+#include "string_utils.h"
+
```
johnnovak, line 27: I suggest enabling strict narrowing checks with `CHECK_NARROWING();` for new files.

```cpp
#include "checks.h"

CHECK_NARROWING();
```

See `adlib_gold.cpp` for a concrete examples.

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,109 @@
...
+static constexpr int HistoryMaxLineSize = 256;
+static constexpr int HistoryMaxNumLines = 500;
+
+[[nodiscard]] static std_fs::path get_shell_history_path();
```
johnnovak, line 31: Can we lose the `[[nodiscard]]`s?

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+ShellHistory& ShellHistory::Instance()
+{
+	static auto history = ShellHistory();
```
johnnovak, line 35: Hmm, `new ShellHistory()` perhaps? It's static, but still, this gave me a pause.

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,109 @@
...
+	return history;
+}
+
+void ShellHistory::Append(std::string command)
+{
+	if (!iequals(command, "exit")) {
```
johnnovak, line 41: We should do left/right whitespace trimming and lowercasing on command, no? Or has it been already done at this point?

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,109 @@
...
+		commands.emplace_back(std::move(command));
+	}
+}
+
+const std::vector<std::string>& ShellHistory::Commands() const
```
johnnovak, line 46: `GetCommands()`

src/shell/shell_misc.cpp
```diff
@@ -683,7 +683,7 @@ Bitu DOS_Shell::GetEnvCount() const
 	return num;
 }
 
-bool DOS_Shell::SetEnv(const char* entry, const char* new_string)
+bool DOS_Shell::SetEnv(std::string_view entry, std::string_view new_string)
```
johnnovak, line 621: I guess string views are kinda OK in this case as you're treating them as a "universal type" that can accept both `const char*` and `std::string`, right? Like a unified interface.

include/shell.h
```diff
@@ -55,7 +55,7 @@ class ByteReader {
 
 class HostShell {
 public:
-	virtual bool GetEnvStr(const char* entry, std::string& result) const = 0;
+	[[nodiscard]] virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
```
johnnovak: Let's discard the `[[nodiscard]]`s with prejudice 😎

src/dos/program_mount.cpp
```diff
@@ -59,7 +59,8 @@ void MOUNT::Move_Z(char new_z)
 		/* Update environment */
 		std::string line = "";
 		std::string tempenv = {new_drive_z, ':', '\\'};
-		if (first_shell->GetEnvStr("PATH",line)) {
+		if (auto result{first_shell->GetEnvStr("PATH")}) {
```
johnnovak: That syntax is a bit too funky... just to get rid of a temporary in non performance intensive code that the compiler will optimise away anyway?

Can we use standard non-surprising syntax for regular use cases?

src/dos/program_mount.cpp
```diff
@@ -59,7 +59,8 @@ void MOUNT::Move_Z(char new_z)
...
+			line = std::move(*result);
```
johnnovak: Oh, man, just copy the string with `=`, let the compiler deal with it 😄 `std::move` should be reserved for performance intensive code when there is tangible difference.

src/shell/shell.cpp
```diff
@@ -268,9 +268,15 @@ void DOS_Shell::ParseLine(char *line)
...
 	if (pipe_file.length()) {
+		auto result = GetEnvStr("TEMP");
+		if (!result) {
+			result = GetEnvStr("TMP");
+		}
 		std::string env_temp_path = {};
-		if (!GetEnvStr("TEMP", env_temp_path) &&
-		    !GetEnvStr("TMP", env_temp_path)) {
+		if (result) {
+			env_temp_path = std::move(*result);
```
johnnovak, line 277: copy

src/shell/shell_cmds.cpp
```diff
@@ -1566,8 +1567,9 @@ void DOS_Shell::CMD_SET(char * args) {
 				char * second = strchr(++p,'%');
 				if (!second) continue;
 				*second++ = 0;
-				std::string temp;
-				if (GetEnvStr(p,temp)) {
+				auto envvar = GetEnvStr(p);
+				if (envvar) {
+					auto temp = std::move(*envvar);
```
johnnovak: copy

src/shell/shell_misc.cpp
```diff
@@ -615,20 +614,19 @@ static void run_binary_executable(const std::string_view fullname,
...
-	if (!entry[0]) {
-		return false;
+	if (entry.empty()) {
+		return std::nullopt;
 	}
 	do {
 		MEM_StrCopy(env_read, env_string, 1024);
 		if (!env_string[0]) {
-			return false;
+			return std::nullopt;
```
johnnovak: `return {}` is simpler, please do that.

src/shell/shell_misc.cpp
```diff
@@ -637,18 +635,17 @@ bool DOS_Shell::GetEnvStr(const char* entry, std::string& result) const
...
 			continue;
 		}
 		/* restore the = to get the original result */
 		*equal = '=';
-		result = env_string;
-		return true;
+		return env_string;
 	} while (1);
-	return false;
+	return std::nullopt;
```
johnnovak: `return {}` is simpler, please do that.

tests/batch_file_tests.cpp
```diff
@@ -55,13 +55,13 @@ class MockReader final : public ByteReader {
...
 class MockShell final : public HostShell {
 public:
-	bool GetEnvStr(const char* entry, std::string& result) const override
+	std::optional<std::string> GetEnvStr(std::string_view entry) const override
 	{
-		if (env.find(entry) == std::end(env)) {
-			return false;
+		auto key = std::string(entry);
+		if (env.find(key) == std::end(env)) {
+			return std::nullopt;
```
johnnovak, line 62: `return {}` is simpler, please do that.

src/dos/program_mount.cpp
```diff
@@ -57,21 +57,21 @@ void MOUNT::Move_Z(char new_z)
 		Drives[25]      = nullptr;
 		if (!first_shell) return; //Should not be possible
 		/* Update environment */
-		std::string line = "";
+		std::string value   = "";
 		std::string tempenv = {new_drive_z, ':', '\\'};
 		if (auto result{first_shell->GetEnvStr("PATH")}) {
```
johnnovak: Yeah, don't like this syntax here either...

src/dos/program_mount.cpp
```diff
@@ -57,21 +57,21 @@ void MOUNT::Move_Z(char new_z)
...
-			line = std::move(*result);
-			std::string::size_type idx = line.find('=');
-			std::string value = line.substr(idx +1 , std::string::npos);
-			while ( (idx = value.find("Z:\\")) != std::string::npos ||
-				(idx = value.find("z:\\")) != std::string::npos  )
-				value.replace(idx,3,tempenv);
-			line = std::move(value);
+			value = std::move(*result);
```
johnnovak, line 63: Looks like we were moving already... okay, hopefully this won't result in hard to debug bugs down the road 🤞🏻

src/shell/shell_batch.cpp
```diff
@@ -122,13 +122,10 @@ std::string BatchFile::ExpandedBatchLine(std::string_view line) const
...
 				break;
 			}
-			std::string env_key(line.substr(0, closing_percent));
 
-			// Get the key's corresponding value from the environment
-			if (const auto env_val = shell.GetEnvStr(env_key)) {
-				// append just the trailing value portion
-				expanded += env_val->substr(env_key.length() +
-				                            sizeof('='));
+			if (const auto env_val{shell.GetEnvStr(
```
johnnovak, line 126: `const auto env_val = shell.GetEnvStr(env_key)` was better.

src/shell/shell_cmds.cpp
```diff
@@ -810,11 +810,9 @@ void DOS_Shell::CMD_DIR(char* args)
 {
 	HELP("DIR");
 
-	auto envvar = GetEnvStr("DIRCMD");
-	if (envvar){
-		auto line = *envvar;
-		std::string::size_type idx = line.find('=');
-		std::string value=line.substr(idx +1 , std::string::npos);
+	auto line = std::string{};
```
johnnovak, line 813: Yeah, or `std::string line = ""` looks less funky and is a less roundabout way ("I ask for `auto`... then change my mind and specify the type anyway 😎")

src/shell/shell_cmds.cpp
```diff
@@ -810,11 +810,9 @@ void DOS_Shell::CMD_DIR(char* args)
...
+	if (const auto envvar{GetEnvStr("DIRCMD")}){
```
johnnovak: Just use that `=` sign 😄

src/shell/shell_cmds.cpp
```diff
@@ -1567,17 +1565,14 @@ void DOS_Shell::CMD_SET(char * args) {
...
 				if (!second) continue;
 				*second++ = 0;
-				auto envvar = GetEnvStr(p);
-				if (envvar) {
-					auto temp = std::move(*envvar);
-					std::string::size_type equals = temp.find('=');
-					if (equals == std::string::npos)
-						continue;
+
+				if (const auto envvar{GetEnvStr(p)}) {
```
johnnovak: `=`

src/shell/shell_cmds.cpp
```diff
@@ -2160,7 +2155,7 @@ void DOS_Shell::CMD_PATH(char *args){
 		this->ParseLine(set_path);
 		return;
 	} else {
-		if (const auto envvar = GetEnvStr("PATH"))
+		if (const auto envvar{GetEnvStr("PATH")})
```
johnnovak: Previous version was better.

src/shell/shell_misc.cpp
```diff
@@ -355,9 +355,9 @@ std::string DOS_Shell::SubstituteEnvironmentVariables(std::string_view command)
 		}
 		closing_percent += 1;
 
-		const std::string env_key(command.substr(1, closing_percent - 1));
-		if (const auto env_val = GetEnvStr(env_key)) {
-			expanded += env_val->substr(env_key.length() + sizeof('='));
+		if (const auto env_val{
+		            GetEnvStr(command.substr(1, closing_percent - 1))}) {
```
johnnovak: `=`

Please let's not micro-optimise and obfuscate the syntax for no good reason.

src/shell/shell_misc.cpp
```diff
@@ -494,12 +494,9 @@ std::string DOS_Shell::Which(const std::string_view name) const
 	static constexpr auto extensions = {"", ".COM", ".EXE", ".BAT"};
 
 	std::vector<std::string> prefixes = {""};
-	auto path_environment             = GetEnvStr("PATH").value_or("");
-	const auto path_equals            = path_environment.find_first_of('=');
 
-	if (path_equals != std::string::npos) {
-		path_environment = path_environment.substr(path_equals + 1);
-		auto path_directories = split(path_environment, ';');
+	if (const auto path{GetEnvStr("PATH")}) {
```
johnnovak: `=`

src/shell/shell_misc.cpp
```diff
@@ -623,7 +620,8 @@ std::optional<std::string> DOS_Shell::GetEnvStr(std::string_view entry) const
 	if (entry.empty()) {
 		return std::nullopt;
 	}
-	do {
+
+	while (true) {
```
johnnovak: Yea, or `for (;;)`.

src/shell/shell_misc.cpp
```diff
@@ -634,18 +632,16 @@ std::optional<std::string> DOS_Shell::GetEnvStr(std::string_view entry) const
...
 			continue;
 		}
 		if (!iequals(entry, env_string)) {
 			continue;
 		}
-		/* restore the = to get the original result */
-		*equal = '=';
-		return env_string;
-	} while (1);
-	return std::nullopt;
```
johnnovak: `return {};`

include/shell.h
```diff
@@ -162,8 +162,7 @@ class DOS_Shell : public Program, public HostShell {
 	virtual bool execute_shell_cmd(char* name, char* arguments);
 
 	[[nodiscard]] std::optional<std::string> GetEnvStr(std::string_view entry) const override;
-	bool GetEnvNum(Bitu num, std::string& result) const;
-	[[nodiscard]] Bitu GetEnvCount() const;
+	[[nodiscard]] std::vector<std::string> FullEnvironment() const;
```
johnnovak, line 152: Drop it please.

src/gui/render.cpp
```diff
@@ -619,12 +619,8 @@ static bool read_shader_source(const std::string& shader_path, std::string& sour
 
 	if (first_shell) {
 		std::string pre_defs;
-		const size_t count = first_shell->GetEnvCount();
-		for (size_t i = 0; i < count; ++i) {
-			std::string env;
-			if (!first_shell->GetEnvNum(i, env)) {
-				continue;
-			}
+		for (auto& env: first_shell->FullEnvironment())
```
johnnovak, line 622: Space before `:`.

src/shell/shell_cmds.cpp
```diff
@@ -1537,9 +1537,8 @@ void DOS_Shell::CMD_SET(char * args) {
 	std::string line;
 	if (!*args) {
 		/* No command line show all environment lines */
-		Bitu count=GetEnvCount();
-		for (Bitu a=0;a<count;a++) {
-			if (GetEnvNum(a,line)) WriteOut("%s\n",line.c_str());
+		for (const auto& entry : FullEnvironment()) {
```
johnnovak: `GetFullEnvironment` or something?

src/shell/shell_misc.cpp
```diff
@@ -644,36 +644,20 @@ std::optional<std::string> DOS_Shell::GetEnvStr(const std::string_view entry) co
 	}
 }
 
-bool DOS_Shell::GetEnvNum(Bitu num, std::string& result) const
+std::vector<std::string> DOS_Shell::FullEnvironment() const
```
johnnovak, line 615: `GetFullEnvironment`, or something better like `GetAllEnvVars`.

src/shell/shell_misc.cpp
```diff
@@ -644,36 +644,20 @@ std::optional<std::string> DOS_Shell::GetEnvStr(const std::string_view entry) co
...
 {
+	auto full_env = std::vector<std::string>{};
+
 	char env_string[1024 + 1];
 	PhysPt env_read = PhysicalMake(psp->GetEnvironment(), 0);
-	do {
+	while(true) {
```
johnnovak: `for (;;)`

src/shell/shell_misc.cpp
```diff
@@ -644,36 +644,20 @@ std::optional<std::string> DOS_Shell::GetEnvStr(const std::string_view entry) co
 	}
 }
 
-bool DOS_Shell::GetEnvNum(Bitu num, std::string& result) const
+std::vector<std::string> DOS_Shell::FullEnvironment() const
 {
+	auto full_env = std::vector<std::string>{};
```
johnnovak: `all_env_vars` would be slightly better.

include/dos_inc.h
```diff
@@ -351,15 +352,31 @@ class MemStruct {
...
 protected:
-	PhysPt pt = 0;
+	PhysPt pt    = 0;
+	~MemStruct() = default;
 };
 
-/* Program Segment Prefix */
+class Environment {
+public:
+	[[nodiscard]] virtual std::optional<std::string> GetEnvStr(
```
johnnovak, line 367: Drop nodiscard please.

include/dos_inc.h
```diff
@@ -414,6 +431,10 @@ class DOS_PSP final : public MemStruct {
 	void SetFCB2(RealPt src);
 	void SetCommandTail(RealPt src);
 
+	[[nodiscard]] std::optional<std::string> GetEnvStr(std::string_view entry) const override;
+	[[nodiscard]] std::vector<std::string> FullEnvironment() const;
```
johnnovak, line 435: Drop nodiscard please.

johnnovak: `GetAllEnvVars`?

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
 }
 
+std::optional<std::string> DOS_PSP::GetEnvStr(const std::string_view entry) const
+{
+	/* Walk through the internal environment and see for a match */
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+
+	char env_string[1024 + 1];
+	if (entry.empty()) {
+		return std::nullopt;
```
johnnovak, line 319: `return {}`

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+	}
+
+	while (true) {
```
johnnovak, line 322: `for (;;)`

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		MEM_StrCopy(env_read, env_string, 1024);
```
johnnovak, line 323: Extract max size to constant (you used 1024 above once already).

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		if (!env_string[0]) {
+			return std::nullopt;
```
johnnovak, line 325: `return {}`

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+}
+
+std::vector<std::string> DOS_PSP::FullEnvironment() const
+{
+	auto full_env = std::vector<std::string>{};
+
+	char env_string[1024 + 1];
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+	while(true) {
+		MEM_StrCopy(env_read, env_string, 1024);
```
johnnovak, line 352: Extract 1024 to constant or use vector or array.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+{
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+
+	// Get size of environment.
+	DOS_MCB mcb(GetEnvironment() - 1);
+	uint16_t envsize = mcb.GetSize() * 16;
+
+	PhysPt env_write          = env_read;
+	PhysPt env_write_start    = env_read;
+	char env_string[1024 + 1] = {0};
```
johnnovak, line 371: Extract 1024 to constant.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,117 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+
+	if (new_string[0]) {
+		std::string bigentry(entry);
+		for (std::string::iterator it = bigentry.begin();
+		     it != bigentry.end();
+		     ++it) {
+			*it = toupper(*it);
+		}
+		snprintf(env_string,
+		         1024 + 1,
```
johnnovak, line 409: Another use of 1024...

src/dos/program_mount.cpp
```diff
@@ -59,7 +59,7 @@ void MOUNT::Move_Z(char new_z)
 		/* Update environment */
 		std::string value   = "";
 		std::string tempenv = {new_drive_z, ':', '\\'};
-		if (auto result{first_shell->GetEnvStr("PATH")}) {
+		if (auto result{psp->GetEnvStr("PATH")}) {
```
johnnovak, line 62: Still don't like this funky syntax 😄

src/shell/shell_cmds.cpp
```diff
@@ -811,7 +811,7 @@ void DOS_Shell::CMD_DIR(char* args)
 	HELP("DIR");
 
 	auto line = std::string{};
-	if (const auto envvar{GetEnvStr("DIRCMD")}){
+	if (const auto envvar{psp->GetEnvStr("DIRCMD")}){
```
johnnovak, line 814: You know what I'm gonna say 😄

src/shell/shell_cmds.cpp
```diff
@@ -1565,7 +1565,7 @@ void DOS_Shell::CMD_SET(char * args) {
 				if (!second) continue;
 				*second++ = 0;
 
-				if (const auto envvar{GetEnvStr(p)}) {
+				if (const auto envvar{psp->GetEnvStr(p)}) {
```
johnnovak, line 1568: `=`

src/shell/shell_cmds.cpp
```diff
@@ -2154,7 +2154,7 @@ void DOS_Shell::CMD_PATH(char *args){
 		this->ParseLine(set_path);
 		return;
 	} else {
-		if (const auto envvar{GetEnvStr("PATH")})
+		if (const auto envvar{psp->GetEnvStr("PATH")})
```
johnnovak, line 2157: `=`

src/shell/shell_misc.cpp
```diff
@@ -355,8 +355,8 @@ std::string DOS_Shell::SubstituteEnvironmentVariables(std::string_view command)
 		}
 		closing_percent += 1;
 
-		if (const auto env_val{
-		            GetEnvStr(command.substr(1, closing_percent - 1))}) {
+		if (const auto env_val{psp->GetEnvStr(
```
johnnovak, line 358: `=`

src/shell/shell_misc.cpp
```diff
@@ -495,7 +495,7 @@ std::string DOS_Shell::Which(const std::string_view name) const
 
 	std::vector<std::string> prefixes = {""};
 
-	if (const auto path{GetEnvStr("PATH")}) {
+	if (const auto path{psp->GetEnvStr("PATH")}) {
```
johnnovak, line 498: `=`

johnnovak (commented):
Looks good overall @MeAreJeenius, but can we tone down the micro-optimisation attempts please 😅

I'd say let's use the simplest syntax that works for non-performance intensive code, and even for performance intensive code, only complicate things when there's good reason (i.e., it's proven by benchmarking that it matters... and not just micro-benchmarks).

Plus I'm against introducing `[[nodiscard]]`.

johnnovak:
I very much recommend reading this excellent piece @MeAreJeenius 
https://quuxplusone.github.io/blog/2019/02/18/knightmare-of-initialization/

I'm reading the most important conclusions and recommendations (but read the whole thing; it's quite amusing).

> If you’re wondering whether to use curly braces or parentheses for construction, your first question should always be: “Do I really need either one?”
>
> Simple guidelines for variable initialization in C++:
>
>    - Use `=` whenever you can.
>
>    - Use initializer-list syntax `{}` only for element initializers (of containers and aggregates).
>
>    - Use function-call syntax `()` to call a constructor, viewed as an [object-factory](https://quuxplusone.github.io/blog/2018/06/21/factory-vs-conversion/).

This way we'll end up with a codebase that everyone will understand; no need to become a language-lawyer and keep the byzantine, baroque, and dizzying complexity of all slightly incompatible and tacked-on C++ features in your head 😅

include/shell.h
kcgen: `discard the [[nodiscard]]`,  :rofl: 

The thing it tells me is that I need to handle and do something with the return value, otherwise calling the function does nothing of value.

```c++
(void) GetEnvStr("testing");
```

is 100% equivalent to doing nothing, and therefore the compiler will give you an error if you try the above, where as with other functions, it'll happily carry on.

src/shell/shell_misc.cpp
kcgen: hopefully the most common or logical break-criteria can be put in the while loop and we can avoid both of them (`while(true)` and `for (;;)`.

include/shell.h
johnnovak: But we need to do something with the return value in all functions that return a value? So where is the value here? We could add `[[nodiscard]]` to every function and I don't see what would improve. It's basically noise.

johnnovak: Ok, I guess you're saying @kcgen that for pure functions that have no side effects to start using this... Well, I still find it very much superfluous; it adds visual noise, but if I don't do anything with the return value, surely my code that uses the function won't work... Dead easy to spot those "errors", but I don't even think I ever made such an error to use a pure function that returns a value then not do anything with the value.

Seems like a solution to a non-issue to me 🤷🏻

johnnovak: To elaborate further, if I look at `GetEnvStr`, it's obvious it returns a value, and most people would assume it does not mutate anything. I don't see how `[[nodiscard]]` makes anyones life better by chucking before "get" functions, adding visual noise. Now, if it _did_ mutate anything in this "get" function, probably it should be renamed and/or documented. Like `GetEnvStrAndCheckEmail` or dunno... you get my drift 😎

johnnovak: Okay, so what's your long-term plan here; add `[[nodiscard]]` to _every single function_ that returns a value? Because surely no function just returns something "just for the fun of it", right? So why not annotate them all, "just in case" it helps?


Real world scenario:

1. When I write some code that calls some function because **I want a value**, I want to do something with that value one or two lines below—that's why I called that function to get the value in the first place, right?
2. What's the likelihood of me _not realising_ one or two lines below that hey, I forgot to assign that value to a var, a value **that I absolutely need** to do what I want to do? Like... zero? 😄 

Let me simplify it:

1. I want to drink a glass of water. Ok, so now I call `GetGlass()` that gives me a glass (I hold the glass in my right hand at this point).
2. Second, I call `PourWaterIntoGlass()`, and I need to pass in _something_ to that function I want to pour the water into (I extend my right hand with the glass in it and open the tap).

What's the likelihood of me sitting there with a dumb expression on my face "not being able to find the bug" because instead of holding the glass in my hand, I let it fall to the floor, and then I open the tap... and where's my glass of water? 😆  How do I even fix this bug... I don't know, it's just too hard... 😝 

Another example:

```c
FILE* fp = fopen("file.txt", "w+");
fseek(fp, 0, SEEK_SET);
```

How can this feature "help" me if I do this?

```c
fopen("file.txt", "w+");
fseek(fp, 0, SEEK_SET);
```

I get a compilation error and it doesn't take a genius to find the "bug" that I forgot to store `fp`...

I guess if our static analysers flag such discarded return value issues then sure, let's do that. But polluting the codebase with `[[nodiscard]]` on every single function, just because it _might_ help someone is a bit of a crazy idea.

Just because people chuck all sorts of things into newer C++ standards it doesn't mean we all must use them without thinking about it a bit first, or that they are even well-thought-out features.

So yeah it wouldn't be a _completely_ worthless warning, but the benefits are extremely minor as I see it. And we'd pollute our codebase significantly. A case of much ado about... well, very slim _potential_ benefits. So yeah, please feel free to configure the static analysers to flag these cases where a return value is discarded if we don't have it configured already.

src/shell/shell_misc.cpp
johnnovak: Yeah, it's just a double-edged swords, because if these string views are stored somewhere else, you can pull the rug from underneath them by freeing the original string they refer to. So I'd use them very, very cautiously. I guess in this case it's fine, though—but micro-optimisation.

johnnovak:
> Neither of the two programs in question are dos programs. They're both dosbox programs, so no actual DOS behavior to compare to (`mount` and `config`).

Okay, but in a broader sense we can check what gets inherited under real MS-DOS if you launch another COMMAND.COM. E.g. are the env vars propagated, what about the SUBST command that maps a directory to a drive letter, etc. We should mimic that so that our non-DOS commands follow the same principles.
 
> With mount, I noticed that updating first_shell while executing in a subshell didn't propagate the environment to the subshell, since the environment is merely copied on initialization. This means that when `mount -z` changes the Z drive to another letter, your path is broken.
> 
> My current change has the opposite problem (breaks parent shell's path), but I've got a fix in the works.

Okay, sounds good, and sounds like we have more freedom here, but I'd still investigate what does get inherited in subshells in a broader sense and mimic that. But doing that child & parent shell syncing sounds like a good idea, and we ideally don't want to introduce regressions in behaviour (like breaking the parent shell's path suddenly seems an arbitrary change).

> Will be making changes and applying your feedback throughout the week.

Sounds good man, thanks 👍🏻 

> See my response in the relevant comment. But yeah, I get that this is a `=` initialization codebase, which is 100% cool with me. I was just way overthinking the (nonexistent) implications of using it in the if statement.

Yeah, I like to keep it simple and approachable for most people 👍🏻 Like the pragmatic approach of the guy in the linked article; do the simplest and non-surprising think, diverge from the simplest thing only if there are very good reasons.

include/shell.h
johnnovak: Thanks for your input on this @MeAreJeenius. I think we've already said everything we can about the subject to each other on this one 😄

Ultimately, the maintainers of the project decide on the overall style and what goes in and what doesn't. In this case, I see this a low-value thing that just adds noise, therefore I don't want these annotations to go in, so please remove them. If there was one universal C++ style to follow, there wouldn't be so many team-specific styles per every project 😄 

However, like I said, you can look into turning on a compiler warning that would issue warnings without having to add these annotations, or maybe extending the static analyser configs to flag unused return values (if we're not already flagging them; I have no idea).

Regardless of this _small_ thing or disagreement, I very much value your functional contributions in this ticket. Great stuff 👍🏻

src/shell/shell_history.cpp
johnnovak: Yeah I agree. I don't like functions that assume things in general. Sometimes necessary, but best to be avoided.

src/shell/shell_cmds.cpp
johnnovak: Yeah ok, it's a minor thing, my biggest problem was the braces 😄

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
 	return true;
 }
 
+static constexpr auto bytes_to_read = 1024;
+
+std::optional<std::string> DOS_PSP::GetEnvStr(const std::string_view entry) const
```
johnnovak, line 314: Suggest `GetEnvString`, `SetEnvString` and `GetAllEnvStrings` for consistency.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		}
+		if (!iequals(entry, env_string)) {
+			continue;
+		}
+
+		return env_string + entry.size() + sizeof('=');
+	}
+}
+
+std::vector<std::string> DOS_PSP::GetAllEnvVars() const
```
johnnovak, line 347: Does this list the names of all env vars? So just the key names (variable names), but not the values, right? Could we use unambiguous and consistent naming across the board? By reading these method names and the names of their arguments, and having no familiarity with this code, I'm a bit confused.

```
std::optional<std::string> DOS_PSP::GetEnvStr(const std::string_view entry) const
```
So `entry` is the name of the variable? Can we use `var_name` instead of `entry` then?

I guess my point is, we should pick one single name for these things and then stick with it everywhere, method names, arg names, variable names in the code, etc. 

I know some of this is pre-existing, I guess I'm just saying let's try to get it right now if you went to the trouble of touching all this code 😄

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		MEM_StrCopy(env_read, env_string, bytes_to_read);
+		if (!env_string[0]) {
+			return all_env_vars;
+		}
+		all_env_vars.emplace_back(env_string);
+		env_read += (PhysPt)(safe_strlen(env_string) + 1);
+	}
+}
+
+bool DOS_PSP::SetEnv(std::string_view entry, std::string_view new_string)
```
johnnovak, line 363: So following the above line of thought, `entry`, should become `var_name` and `new_string` will be `value`.

johnnovak:
Looks fine @LowLevelMahn, thanks for addressing the comments. Gave it a quick testing and works ok. I think launching subshells is quite a niche use in general, so should be relatively safe to merge.

kcgen:
@MeAreJeenius, can you build with UBSAN and ASAN and try to exercise as many of the changed code pathes here (with good, bad, and empty input, too - if possible)?

johnnovak:
> @MeAreJeenius, can you build with UBSAN and ASAN and try to exercise as many of the changed code pathes here (with good, bad, and empty input, too - if possible)?

+1 on this one. What I've been in my tickets is to post my manual test plan along with test evidence, especially for changes that require "specialist knowledge". Like all my recent VGA changes that deal with edge cases and niche areas of VGA emulation and shaders.

So listing out in a bulleted list what you tested and how it the behaviour fixed or changed would be sufficient, it doesn't need to be big. Then we can all re-test manually before merging or perhaphs identify gaps, etc.

You've done something along those lines in the first comment but that's a bit general.

kcgen:
>  What I've been doing in my PRs is to post my manual test plan along with test evidence, especially for changes that require "specialist knowledge". Like all my recent VGA changes that deal with edge cases and niche areas of VGA emulation and shaders.

This has been a huge help @johnnovak.

I've been making extremely slow progress with my destruction sequence PR..  I'm constantly re-stashing my pending changes, checkout other builds, commenting, compiling-all-commits, then running my own UB+ASAN tests, and so on.  Then get back to my own PR and regroup :sweat_smile: (of course, a very good problem to have!)

The more testing we can share (especially when running UB+ASAN builds), the more confidence we can offload from reviewers.

kcgen:
@MeAreJeenius - latest push looks excellent and all feedback addressed :rocket: 

> I'm not personally familiar with the intricacies of the shell/subshell handling

Same; I'm not sure how to exercise all/most of the changed code paths. I tried with the previous shell refactoring, but I missed a bunch that [thankful surfaced with more routine test jobs](https://github.com/dosbox-staging/dosbox-staging/issues?q=is%3Aissue+assignee%3AMeAreJeenius+is%3Aclosed+label%3Aregression) in the weeks after merging.

johnnovak:
> @MeAreJeenius - latest push looks excellent with all of @johnnovak's feedback addressed 🚀
> 
> > I'm not personally familiar with the intricacies of the shell/subshell handling
> 
> Same; I'm not sure how to exercise all/most of the changed code paths. I tried with the previous shell refactoring, but I missed a bunch that [thankfully surfaced with more routine test jobs](https://github.com/dosbox-staging/dosbox-staging/issues?q=shell+is%3Aissue+assignee%3AMeAreJeenius+is%3Aclosed+label%3Aregression+) in the weeks after merging.

Exactly. The goal is to catch as many of those potential regressions now, before merge time.

johnnovak:
> > can you build with UBSAN and ASAN and try to exercise as many of the changed code pathes here (with good, bad, and empty input, too - if possible)?
> 
> That's the plan. I wanted to wait until most of the review changes were addressed before doing tests. From what I remember, a couple of the regressions from last time were from review changes that weren't well tested. Right now, it's not well tested beyond the happy path, and I've found a couple of bugs, so it's definitely not ready for a merge yet.
> 
> Once I implement the remaining requests, I'll do a bunch of testing, document it, and add context that I think is important sometime in the next week or so.

All sounds good @LowLevelMahn 👍🏻 

> I'll probably end up renaming the functions to `GetEnvironmentValue`, `SetEnvironmentValue`, and `GetAllRawEnvironmentStrings`.

That would be a welcome improvement 💯 Some of these methods are named so badly it's crazy...

kcgen:
@johnnovak , I might not be connecting some dots: is @LowLevelMahn == @MeAreJeenius (author of this PR).
(You mentioned the former a couple times above.. just double checking!)

johnnovak:
Closing due to inactivity.

kcgen:
@MeAreJeenius , thank;s for your prior committed PRs and the intent you had with this PR, even though this one resulted in more discussion than all of us wanted.

Hopefully you can return one day and plow through the remaining feedback items (even the feedback items that you feel are inferior).

For now I'll be removed your account's write-access as you mentioned this would be your last PR.  The PR can still be re-opened or your branch pushed fresh from your local repo if your plans change. And of course, your account can be re-activated.  Regards.

kcgen:
A bit of a follow up - this is an excellent reference:

https://www.jeffgeerling.com/blog/2022/burden-open-source-maintainer

> _"My coping strategy for the past five years has been somewhat [ruthlessly closing PRs and issues](https://www.jeffgeerling.com/blog/2016/why-i-close-prs-oss-project-maintainer-notes). I also wrote about [enabling the stale bot](https://www.jeffgeerling.com/blog/2020/enabling-stale-issue-bot-on-my-github-repositories) two years ago, and let me tell you:"_

> _"Despite how much some people detest the stale bot, it—along with a more laissez-faire attitude towards my projects—is the best burnout prevention measure I've taken."_

johnnovak:
Another good article from the same guy:
https://www.jeffgeerling.com/blog/2020/saying-no-burnout-open-source-maintainer

> Be liberal with your 'no', be judicious with your 'yes'.

Ultimately, this is a spare-time hobby for all of us—there's no point in forcing things if we cannot get along and have different goals/visions. That's just... life; nobody is compatible with every other person on the planet, and that's fine.

Nevertheless, we wish you all the best @MeAreJeenius! 😄

johnnovak:
Man, this guy is a goldmine of good articles:
https://www.jeffgeerling.com/blog/2016/why-i-close-prs-oss-project-maintainer-notes

> 'No' is the default

He's right! 😄

johnnovak:
> @johnnovak @kcgen
> 
> Really sorry about the huge delay. Life stuff ate up a lot more time than I would have liked.
> 
> Earlier this week I started work on code tweaks + unit tests. Later this week I plan on going through the non-automatable tests and writing those down. If no significant bugs are found both issues should be ready to be reopened, though I'll ping you both when stuff is pushed to the PR.

Thanks for getting back to us with an update @MeAreJeenius. We are looking forward to your PR. Adding unit test coverage sounds very good indeed! 🎉

Issue #2797, opened 2023-08-28
Shell: file size column in "dir" output formatted wrongly


darkstar:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

git HEAD 8270aad5b443b6deb10dd03331721419a3730378

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The output of the "dir" command is broken:
![image](https://github.com/dosbox-staging/dosbox-staging/assets/43795/e1ed5257-4c93-40f4-a4c2-5c0838b4d284)

it should look more like this:
![image](https://github.com/dosbox-staging/dosbox-staging/assets/43795/f168fd61-76ed-4ee9-8ff6-f6a7db930334)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. launch dosbox-staging
2. mount anything as drive C:
3. type "dir c:"

### Your configuration

```ini
I don't think it is relevant but I changed these settings:

[dos]
country = 49

[ipx]
ipx = true

[autoexec]
keyb gr 437
```

### Provide a Log

```text
n/a
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

darkstar:
This is a fix (sorry it's late and I'm too lazy to clone the repo to make a pull request for just a single-character fix ;-) 

````diff
diff --git a/src/shell/shell_cmds.cpp b/src/shell/shell_cmds.cpp
index b6ce9ed0b..a5dd2e2fc 100644
--- a/src/shell/shell_cmds.cpp
+++ b/src/shell/shell_cmds.cpp
@@ -1009,7 +1009,7 @@ void DOS_Shell::CMD_DIR(char* args)
                const auto hour   = static_cast<uint8_t>((entry.time >> 5) >> 6);
                const auto minute = static_cast<uint8_t>((entry.time >> 5) & 0x003f);

-               output.AddString("%-8s %-3s   %-21s %s %s\n",
+               output.AddString("%-8s %-3s   %21s %s %s\n",
                                 entry.GetBareName().c_str(),
                                 entry.GetExtension().c_str(),
                                 entry.IsDirectory()
````

johnnovak:
> This is a fix (sorry it's late and I'm too lazy to clone the repo to make a pull request for just a single-character fix ;-)

[...]

PR #2799, opened 2023-08-29
Fix alignment of file size column in the DIR command's output

johnnovak:
Reported & fixed by @darkstar — I just simply applied his fix 😄 
Thanks @darkstar 🎉 

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2797


Torinde:
Fixes #2797 (self-referencing typo in OP Issue number)

PR #2802, opened 2023-08-29
Track previously added userdata when deleting the SDL resize watch

kcgen:
In order for SDL to delete a previously added EventWatch, the original arguments need to be submitted otherwise the delete request won't be performed.

This PR makes the window resize callback management into into a self-contained function where we can track both of these EventWatch arguments, which guarantees the delete happens.

The variables are:
 - The function pointer (to the callback)
 - The `userdata`, in our case, a pointer to the SDL `window` object

This fixes random crashes on startup on macOS as well as fixes crashes fullscreen and switching to `output = texture` in @GranMinigun's surface-removal PR.

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -530,36 +530,50 @@ static void populate_requested_vsync_settings()
 	}
 }
 
-/* This function is SDL_EventFilter which is being called when event is
- * pushed into the SDL event queue.
- *
- * WARNING: Be very careful of what you do in this function, as it may run in
- * a different thread!
- *
- * Read documentation for SDL_AddEventWatch for more details.
- */
-static int watch_sdl_events(void *userdata, SDL_Event *e)
+// Enable or disable the resize handler in SDL's event system. It should be
+// disabled prior to making any significant window or rendering modifications.
+//
+enum class HandlerState { Enable, Disable };
+static void set_window_resize_handler(const HandlerState handler_state)
 {
-	/* There's a significant difference in handling of window resize
-	 * events in different OSes. When handling resize in main event loop
-	 * we receive continuous stream of events (as expected) on Linux,
-	 * but only single event after user stopped dragging cursor on Windows
-	 * and macOS.
-	 *
-	 * Watching resize events here gives us continuous stream on
-	 * every OS.
-	 */
-	if (e->type == SDL_WINDOWEVENT && e->window.event == SDL_WINDOWEVENT_RESIZED) {
-		SDL_Window *win = SDL_GetWindowFromID(e->window.windowID);
-		if (win == (SDL_Window *)userdata) {
-			const int w = e->window.data1;
-			const int h = e->window.data2;
-			// const int id = e->window.windowID;
-			// DEBUG_LOG_MSG("SDL: Resizing window %d to %dx%d", id, w, h);
-			HandleVideoResize(w, h);
+	// The resize_handler is an 'SDL_EventFilter' (see SDL_AddEventWatch)
+	auto resize_handler = [](void* userdata, SDL_Event* e) -> int {
+		assert(userdata);
+		assert(e);
+
+		// Is it a window resize event?
+		if (e->type == SDL_WINDOWEVENT &&
+		    e->window.event == SDL_WINDOWEVENT_RESIZED) {
+			// Does the event's window match our SDL block's window?
+			const auto window = SDL_GetWindowFromID(e->window.windowID);
+			auto sdl_window = reinterpret_cast<SDL_Window*>(userdata);
+			if (window == sdl_window) {
+				const int w = e->window.data1;
+				const int h = e->window.data2;
+				HandleVideoResize(w, h);
+			}
 		}
+		return 0;
+	};
+
+	static void* userdata = nullptr;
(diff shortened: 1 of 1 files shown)


johnnovak (approved):
Very nice quick fix @kcgen! 🎉 Merge away 🚀 

While this fixes the macOS crash and transitions us into using the SDL even watching functionality properly, I still don't get it why this extra custom stuff is needed.

The old comment says SDL issues a single resize event once the user stopped resizing the window on Windows and macOS. And we're effectively pausing the emulation during window resizing anyway, at least on macOS and Windows.

So why do we even need a flurry of _continuous_ window resize events when a single one would suffice once the resizing has stopped? I guess just that we can display the current content being strecthed to the new window size while resizing? Or the emulation itself isn't stopped on Linux while resizing? 

I guess I'd like to understand as it seems like unnecessary complexity.

kcgen:
hmm.. even though it fixes the crash, the resizing isn't working on Linux now. Argh. Reverting.

kcgen:
Sorry @johnnovak - I thought had a solid understanding of this SDL event watch, but I guess not :sob:

johnnovak:
> Sorry @johnnovak - I thought had a solid understanding of this SDL event watch, but I guess not 😭

No prob, but my original question still stands: what if we got rid of this and only handled `SDL_WINDOWEVENT_SIZE_CHANGED`? What would we lose, if anything? That would simplify things a lot.

kcgen:
Hmm.. I'm testing `kc/gm/surface-purge-2`, and it's working great on Linux _and_ macOS; so maybe all is not lost, after all.

I'm able to resize up and down on Linux on macOS, switch from opengl to texture and back. Full screen or windowed.

Is `kc/gm/surface-purge-2` working OK for you too, on macOS @johnnovak ?

johnnovak:
> Is `kc/gm/surface-purge-2` working OK for you too, on macOS @johnnovak ?

Will test after breakfast and playing with the cats. On my phone while still laying in bed 😎

kcgen:
> what if we got rid of this and only handled `SDL_WINDOWEVENT_SIZE_CHANGED`?

I'm not sure - it might be functionally the same.

One of the biggest dilemmas the we go into with the SDL callback code was creating feedback loops:
 - User does something, triggering an SDL event
 - the SDL event is "handled" by our callback, but in doing so, trigger a new SDL event
 - wash, rinse, repeat

That was the motivation for the `safe_set_window_size` function that disengages our callbacks across `SDL_SetWindowSize` events; otherwise we triggered various UBSAN and TSAN issues.

One of the nice things w/ SDL performing the callback (using their Watch system) is the call is performed just prior to the event, so all the objects are still in a stable/reliable state.

johnnovak:
> > what if we got rid of this and only handled `SDL_WINDOWEVENT_SIZE_CHANGED`?
> 
> I'm not sure - it might be functionally the same.
> 
> One of the biggest dilemmas the we go into with the SDL callback code was creating feedback loops:
> 
>     * User does something, triggering an SDL event
> 
>     * the SDL event is "handled" by our callback, but in doing so, trigger a new SDL event
> 
>     * wash, rinse, repeat
> 
> 
> That was the motivation for the `safe_set_window_size` function that disengages our callbacks across `SDL_SetWindowSize` events; otherwise we triggered various UBSAN and TSAN issues.
> 
> One of the nice things w/ SDL performing the callback (using their Watch system) is the call is performed just prior to the event, so all the objects are still in a stable/reliable state.

Yeah I guess I'm saying all this cruft was necessary at some time, but maybe not because we've changed stuff and/or SDL has improved, etc. So I'd always look at dropping the cruft and try the simplest thing. Might be worthwhile.

Tried the branch and restarted 20-30 times, then switched output/texture in fullscreen 20-30 times. All stable, but the cat is getting now very impatient 😄

johnnovak:
On a second thought, there might be more complex uses cases that I'm not aware, so yeah, maybe safer to just move on with this fix for now if it solves the immediate problem, then let @GranMinigun think about potential simplification as part of his SDL rework.

Torinde:
> Will test after breakfast and playing with the cats. On my phone while still laying in bed 😎

How do you test macOS (or any, actually) Staging on a phone?
I assume it's via Remote desktop, but have to ask... just in case there is an Android or iOS version I somehow missed...

johnnovak:
> > Will test after breakfast and playing with the cats. On my phone while still laying in bed 😎
> 
> How do you test macOS (or any, actually) Staging on a phone? I assume it's via Remote desktop, but have to ask... just in case there is an Android or iOS version I somehow missed...

I don't. I'm just replying to messages sometime in the morning while still in bed on my phone, but I don't run Staging on my phone 😄

GranMinigun:
> there might be more complex uses cases that I'm not aware of

There aren't. It's just a desire of the original author to have unified behaviour, as far as I can tell, which actively fought SDL's own window handling. Sorry for your wasted time and headaches. I've burned that event watch to ashes.

kcgen:
Even better, @GranMinigun! The less code that tries to circumvent or augment SDL's own window handling, the better.

johnnovak:
> I've burned that event watch to ashes.

Praise be! 😎 🤘🏻

PR #2785, opened 2023-08-25
Implement FFmpeg Video Enocder

weirddan455:
Fixes #2764

I figure it's about time I show my work on this.  If nothing else, I need to see if this builds on CI.  I need to add a meson wrap and figure out WTF to do with the Visual Studio files to add a new dependency.  I dual boot Windows 10 with Visual Studio 2022 so I'll tackle the build stuff next.  FFmpeg is going to be installed out of the box on nearly all Linux distros so it should work there as is.

There's still a few question marks.  Mainly, it needs testing on pixel formats other than `BGRX8888`, because that's all DOSBox ever outputs on my computer.  `Indexed8` is a todo (Does anyone use this?).  The others have mappings to FFmpeg and should theoretically work but needs testing.

It's fully threaded (using 3 threads actually, one for video encoding, one for audio encoding, and one for the muxer).  I've spent the last couple days ironing out threading bugs and I've got it into a state I'm fairly happy with now overall.

Just needs the last finishing touches mentioned above.  Also, some config options to control quality.  If anyone has suggestions to better default quality settings, I'm all ears.  Video doesn't look that great in fullscreen, though I suspect part of that may be due to VLC doing some poor scaling.  It looks a lot better when it's the size of a postage stamp :laughing:   If that integer scaling @johnnovak was working on is in `main`, I may try to implement that.  Right now it's just outputting the same resolution that it gets out.

Feel free to review if you like.  I'm still going to be working on the above mentioned things but any comments and suggestions are welcome :smile: 

Also, here's about a minute and a half of Doom I just recorded as a sample:

https://github.com/dosbox-staging/dosbox-staging/assets/8128047/95288338-8504-4b79-b127-4ea6c9ec6860

PR #2805, opened 2023-08-30
Update packages/*.txt files to include ffmpeg and conan

weirddan455:
See discussion on #2785

Docker images seem to be getting clobbered.  This should update them to include the packages I need so CI can build.


kcgen:
Nice, @weirddan455. Hopefully the docket images get populated 👍

Issue #2804, opened 2023-08-29
Catacomb 3-D video regression


johnnovak:
The video rendering of Catacomb 3-D (1992) (from eXoDOS v5) has regressed compared to 0.80.1 and latest SVN.

Just start the game with the default config, then press Enter a couple of times to start the game.

The HUD at the bottom of the screen is constantly and rapidly flickering. In 0.80.1 and SVN the image is 100% stable—zero flicker. Tried with different `svga_*` machine types; there's no improvement.

![image](https://github.com/dosbox-staging/dosbox-staging/assets/698770/b41452f4-9328-48c0-96c0-1b7a28f22e2f)

kcgen:
great catch, @johnnovak.
Bisecting.

johnnovak:
> great catch, @johnnovak. Bisecting.

I'm going through a huge pile of EGA games because I love the EGA era, plus I have some tricks up my sleeve that I'm gonna unleash soon (some enhancements to the adaptive CRT shader stuff for EGA modes).

kcgen:
bisects to: 2b26d90dc0cac2e17538836c95932b4b17e667bf
Prior commit is good: `git checkout 2b26d90~1`
Nothing jumping out at me that's obvious - looks like mostly rendering changes as opposed to VGA internals (which is more what I was suspecting!)

Hopefully you can discover what triggered it :+1:

johnnovak:
> bisects to: [2b26d90](https://github.com/dosbox-staging/dosbox-staging/commit/2b26d90dc0cac2e17538836c95932b4b17e667bf) Prior commit is good: `git checkout 2b26d90~1` Nothing jumping out at me that's obvious - looks like mostly rendering changes as opposed to VGA internals (which is more what I was suspecting!)
> 
> Hopefully you can discover what triggered it 👍

Thanks, that helps. I think it's something related to VGA internals; it's just this change changes the double-scan rendering behaviour, that triggers some earlier mistake that I probably made around the double-scanning.

I'll investigate, this helps!

PR #2806, opened 2023-08-30
Fix the calculation of vga.draw.lines_scaled when double scanning

johnnovak:
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2804

It was a very interesting regression; spent a good part of today getting to the bottom of it, comparing `vga.draw.*` dumps, etc.

In a nutshell, proper double-scanning has been broken for about a month now, but this is very hard to spot as ~98% of games and demos would happily continue to work. You can test 1000+ DOS games and never notice... until you hit that 2% that manifest some weird glitch if double scanning is not quite right. Like Catacomb 3D—it goes into my VGA regression testing folder right away 😎🤘🏻 

## Problem description

- Before adding support for 'force_single_scan' via shader pragmas, all double-scanned VGA modes were either rendered as double-scanned ("true" double scanning), or were doubled vertically post-render ("fake" double scanning). Because of this, the 'double_height' flag effectively indicated whether we're in fake or true double-scan mode.

- The introduction of 'force_single_scan' has changed this; now we're only enabling 'double_height' when fake double scanning CGA modes, and for all double-scanned VESA modes. But when forcing single scanning of any other double-scanned VGA modes, 'double_height' is no longer set. Therefore, 'double_height' is no longer an indicator whether we're in fake or true double-scan mode.

- Downstream VGA rendering relies on the `vga.draw.lines_scaled` set to 2 when in fake double scan VGA modes, and to 1 in any other case.

- This value was derived from 'double_height' before. That won't work anymore for the above reasons, so this change introduces a 'force_single_scan' flag that is explicitly calculated during video mode setup, then 'vga.draw.lines_scaled' is derived from that.

I've created a spreadsheet for this to make sure I'm getting it right. The red column in the "NEW" block caused the problem; with the fix we're back to the old `lines_scaled` values:

image

## How to test

Catacomb 3D is a great way to test. So is flicking through all video modes in DPaint II and logging `lines_scaled`. Test with both double-scanning and forced single scanning, so:

1. `glshader = default` (forced single scan case)
2. Any CRT shader that does _not_ use `#pragma force_single_scan` (double scan case).
3. Any CRT shader that _does_ use `#pragma force_single_scan` (forced single scan case, like point 1).

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1926,18 +1926,28 @@ void VGA_SetupDrawing(uint32_t /*val*/)
 	const auto horiz_end = updated_timings.horiz_display_end;
 	const auto vert_end  = updated_timings.vert_display_end;
 
+	// Determine video mode info, render dimensions, and source & render
+	// pixel aspect ratios
+
 	VideoMode video_mode = {};
 
-	// Calculate render dimensions and source & render pixel aspect ratios
 	uint32_t render_width  = 0;
 	uint32_t render_height = 0;
 
-	// If true, the rendered image will be doubled vertically post-render
-	// via a scaler.
+	// True only if we're rendering a double scanned VGA mode as single
+	// scanned (so rendering half the lines only, e.g., only 200 lines for
+	// the double-scanned 320x200 13h VGA mode).
+	bool force_single_scan = false;
+
+	// If true, the rendered image will be doubled horizontally post-render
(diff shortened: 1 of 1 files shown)


kcgen (approved):
Nice quick fix, @johnnovak  -and very helpful comment additions.
That spreadsheet looks very helpful!
No code comments; all looks good :+1:

Issue #2795, opened 2023-08-28
MT32 does not reset instruments when changing game


Scattershor:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.1 Maintenance release

### What Operating System are you using?

None

### Relevant hardware info

Ryzen 7 5800X, RTX 3080

### Have you checked that no other similar issue already exists?

- [x] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The MT32 emulation does not reset instruments upon game quit or switch. Running SQ3 then switching to SOMI makes the theme of SOMI all stuffed up. This occurs when switching games from one MT32 game to another.

### Steps to reproduce the behaviour.

Explain how to reproduce
To easily replicate: 
1. Run Space Quest 3 with built in MT32 emulation. 
2. Quit to DOS prompt
3. Run Secrets of Monkey Island

Observe the theme music be all whack. 

### Download URL of affected game or software

https://archive.org/download/20220917_20220917_0541/Secret%20of%20Monkey%20Island%2C%20The%20%28VGA%29%20%281992%29%20%28v1.0%29%20%28Rerelease%29%20%281.44M%29%20%5B%23%5D.zip ; https://archive.org/download/spacequestcollectorseditionusa/Space%20Quest%20-%20Collector%27s%20Edition%20%28USA%29.zip

### Your configuration

```ini
https://gist.github.com/Scattershor/b8c7ae7820557b0652c2ae03e5fffc8f
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

rderooy:
I remember the same thing would happen on my real MT-32 that I had back in the day. The easy solution back then was to power-cycle the MT-32 in between games, which was quite easy with a real MT-32. But if you had the ISA card version (LAPC-I) you had to power cycle the PC. But I think there was also a MIDI command that could cause the MT-32 to reset to default settings.

Scattershor:
> I remember the same thing would happen on my real MT-32 that I had back in the day. The easy solution back then was to power-cycle the MT-32 in between games, which was quite easy with a real MT-32. But if you had the ISA card version (LAPC-I) you had to power cycle the PC. But I think there was also a MIDI command that could cause the MT-32 to reset to default settings.

Is there any way to do that on DOSBox Staging without quitting the program outright? (This is referring to the internal MT32 emulator built in).

kcgen:
[...]

PR #2807, opened 2023-08-30
Make shell configuration shortcuts optional/configurable

FeralChild64:
Our `COMMAND.COM` replacement allows to use shortcuts for changing the configuration, for example instead of

`config -set dma=7`

it is enough to execute command `dma 7`. Some like it, some don't (see a slightly off-topic discussion in issue https://github.com/dosbox-staging/dosbox-staging/issues/2795), as the shell can go the 'configuration view/change' route in unexpected moments - like when we thing some executable in in the search `PATH`, while in reality it is not.

This PR makes this functionality configurable (but enabled by default). The small argument 'const' cleanup didn't cover the whole `DOS_Shell::Execute*` functions, as I wanted to limit the scope of the changes to a reasonable size and not introduce additional risks.

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1135,6 +1135,11 @@ void DOSBOX_Init()
 	        "(auto by default, enabled if DOS version >= 7.0).\n"
 	        "FreeDOS and MS-DOS 7/8 COMMAND.COM supports this behavior.");
 
+	pbool = secprop->Add_bool("shell_config_shortcuts", when_idle, true);
+	pbool->Set_help("Allow shortcuts for direct configuration management (enabled by default), i.e.\n"
+	                "instead of 'config -set sbtype sb16' it is enough to execute 'sbtype sb16', \n"
+	                "and instead of 'config -get sbtype' it is enough to execute 'sbtype' command.\n");
+
 	pstring = secprop->Add_path("shell_history_file", only_at_start, "shell_history.txt");
 	pstring->Set_help(
 		"File containing persistent command line history ('shell_history.txt'\n"
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -631,6 +631,11 @@ void DOSBOX_Init()
 		"copy-on-write or network-based filesystem, this setting avoids triggering\n"
 		"write-operations for these write-protected files.");
 
+	pbool = secprop->Add_bool("shell_config_shortcuts", when_idle, true);
+	pbool->Set_help("Allow shortcuts for direct configuration management (enabled by default), i.e.\n"
+	                "instead of 'config -set sbtype sb16' it is enough to execute 'sbtype sb16', \n"
+	                "and instead of 'config -get sbtype' it is enough to execute 'sbtype' command.");
+
 	// Configure render settings
 	RENDER_AddConfigSection(control);
 
@@ -1134,11 +1139,6 @@ void DOSBOX_Init()
 	        "Enable expanding environment variables such as %PATH% in the DOS command shell\n"
 	        "(auto by default, enabled if DOS version >= 7.0).\n"
 	        "FreeDOS and MS-DOS 7/8 COMMAND.COM supports this behavior.");
-
-	pbool = secprop->Add_bool("shell_config_shortcuts", when_idle, true);
-	pbool->Set_help("Allow shortcuts for direct configuration management (enabled by default), i.e.\n"
-	                "instead of 'config -set sbtype sb16' it is enough to execute 'sbtype sb16', \n"
-	                "and instead of 'config -get sbtype' it is enough to execute 'sbtype' command.\n");
 
 	pstring = secprop->Add_path("shell_history_file", only_at_start, "shell_history.txt");
 	pstring->Set_help(
--- a/src/shell/shell_cmds.cpp
+++ b/src/shell/shell_cmds.cpp
@@ -129,7 +129,7 @@ bool lookup_shell_cmd(std::string name, SHELL_Cmd &shell_cmd)
 	return true;
 }
 
-bool DOS_Shell::CheckConfig(const char* const cmd_in, const char* const line)
+bool DOS_Shell::ExecuteConfigChange(const char* const cmd_in, const char* const line)
 {
 	assert(control);
 	const auto section_dos = static_cast<Section_prop*>(
@@ -160,7 +160,8 @@ bool DOS_Shell::CheckConfig(const char* const cmd_in, const char* const line)
 	return true;
 }
 
-bool DOS_Shell::execute_shell_cmd(char *name, char *arguments) {
+bool DOS_Shell::ExecuteShellCommand(const char* const name, char* arguments)
+{
 	SHELL_Cmd shell_cmd = {};
 	if (!lookup_shell_cmd(name, shell_cmd))
 		return false; // name isn't a shell command!
@@ -182,22 +183,31 @@ void DOS_Shell::DoCommand(char * line) {
 //		if (*line == ':') break; //This breaks drive switching as that is handled at a later stage.
 		if ((*line == '.') ||(*line == '\\')) {  //allow stuff like cd.. and dir.exe cd\kees
 			*cmd_write=0;
-			if (execute_shell_cmd(cmd_buffer, line)) {
+			if (ExecuteShellCommand(cmd_buffer, line)) {
 				return;
 			}
 		}
 		*cmd_write++=*line++;
 	}
 	*cmd_write=0;
-	if (is_empty(cmd_buffer))
+	if (is_empty(cmd_buffer)) {
 		return;
-	/* Check the internal list */
-	if (execute_shell_cmd(cmd_buffer, line))
+	}
+
+	// First try to execute the line as internal shell command
+	if (ExecuteShellCommand(cmd_buffer, line)) {
 		return;
-/* This isn't an internal command execute it */
-	if (Execute(cmd_buffer,line)) return;
-	if (CheckConfig(cmd_buffer,line)) return;
-	WriteOut(MSG_Get("SHELL_EXECUTE_ILLEGAL_COMMAND"),cmd_buffer);
+	}
+	// Try to execute the line as external program
+	if (ExecuteProgram(cmd_buffer, line)) {
+		return;
+	}
+	// Last resort - try to handle the line as configuration change request
+	if (ExecuteConfigChange(cmd_buffer, line)) {
+		return;
+	}
+
+	WriteOut(MSG_Get("SHELL_EXECUTE_ILLEGAL_COMMAND"), cmd_buffer);
 }
 
 bool DOS_Shell::WriteHelp(const std::string &command, char *args) {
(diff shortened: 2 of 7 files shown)


src/dosbox.cpp
```diff
@@ -1135,6 +1135,11 @@ void DOSBOX_Init()
 	        "(auto by default, enabled if DOS version >= 7.0).\n"
 	        "FreeDOS and MS-DOS 7/8 COMMAND.COM supports this behavior.");
 
+	pbool = secprop->Add_bool("shell_config_shortcut", when_idle, true);
+	pbool->Set_help("Enable shell shortcuts for easy configuration changing, i.e. instead of\n"
+	                "'config -set property value' it is enough to execute 'config value' command\n"
+	                "(enabled by default).\n");
+
```
johnnovak, line 1142: The description is good, but need to be extended because querying config settings is also possible by executing their name only.

johnnovak: The default value usually goes on top, at the end of the short summary, so I'd mention it before the "i.e." part, then from "i.e." I'd turn that into a new sentence (that's the detailed description).

src/shell/shell_cmds.cpp
```diff
@@ -323,7 +346,7 @@ void DOS_Shell::CMD_HELP(char * args){
 	char help_arg[] = "/?";
 	const auto& hl = HELP_GetHelpList();
 	if (contains(hl, args) && hl.at(args).type == HELP_CmdType::Program) {
-		Execute(args, help_arg);
+		ExecuteProgram(args, help_arg);
```
johnnovak, line 349: Nice explanations and the method names are much better 🎉

src/dosbox.cpp
rderooy: Don't you mean "it is enough to execute 'property value'"?

johnnovak: Yeah, "setting" instead of "property" or "property value" should be used, actually. That's the name I'm consistently using in the manual ("setting", or "config setting").

Property is too ambiguous and/or too technical; setting or config setting for disambiguation is a good term understood by everybody, so let's use that for consistency (I've already normalised all config descriptions to only use this term).

src/dosbox.cpp
```diff
@@ -1135,6 +1135,11 @@ void DOSBOX_Init()
 	        "(auto by default, enabled if DOS version >= 7.0).\n"
 	        "FreeDOS and MS-DOS 7/8 COMMAND.COM supports this behavior.");
 
+	pbool = secprop->Add_bool("shell_config_shortcut", when_idle, true);
```
kcgen, line 1138: Suggest plural: `allow_shell_config_shortcuts = yes/no`

src/dosbox.cpp
kcgen: Suggest a common example instead of placeholders:

The shell config shortcuts allow directly making conf changes, i.e.: `C:\>sbtype sb16` without using the CONFIG tool: `C:\> config -set sblaster sbtype = sb16`

johnnovak: > I think using "config" can be confused with the "config" command.

No, because you talk about "settings", "config settings", or "the `CONFIG` command". I will set up a glossary at some point for the manual so everybody follows the same terms consistently. Examples:

"You can change the `glshader` config setting to <blah blah>. What this setting does is <more blah>"
"Run the `CONFIG` command to <...>"

It's 100% unambigous; the key is consistency 😄 Property may be familiar to us devs, but it's a super weird term to use in a manual meant for regular people of power users.

src/dosbox.cpp
johnnovak: Actually, that's a lot better, yes.

src/dosbox.cpp
johnnovak: > Suggest a common example instead of placeholders:
> 
> The shortcuts allow conf settings to be changed directly, for example: `C:\>sbtype sb16`, instead of using the more combersome CONFIG tool, such as: `C:\>config -set sblaster sbtype = sb16`

Even better! 🚀 Writing concisely and clearly is really hard; usually my 4th or 5th iteration is that I keep for config descriptions, and let's not even get into the user manual situation 😄

kcgen: Thanks @johnnovak :-)

At least everyone mucking with conf settings knows about `sbtype`; it's been there since the first release and will be there forever more. Another commonly used one is `cycles` as well; whatever you like @FeralChild64.

And just for fun, this setting itself can be set via shortcut, so it could be described as: 

> Shortcuts allow conf settings to be changed directly, for example: `C:\>allow_shell_config_shortcuts no` will disable shell config shortcuts, after which the more cumbersome CONFIG tool would be needed re-enable them: `C:\>config -set dosbox allow_shell_config_shortcuts = yes`.

But that makes my head spin :face_with_spiral_eyes: :laughing:

johnnovak: Uhuhh, that sentence is too complicated 😅 A bit baroque 😎 Your first attempt is much better and conveys all the user needs to know perfectly and succinctly:

> The shortcuts allow conf settings to be changed directly, for example: `C:\>sbtype sb16`, instead of using the more cumbersome CONFIG tool, such as: `C:\>config -set sblaster sbtype = sb16`

(And nice touch including `C:\>` at the start of the commands to make it clear it's about executing stuff on the command line. These little touches go a long way towards removing ambiguity and confusion).

kcgen: > Uhuhh, that sentence is too complicated

(lol.. yeah; I was just joking around w/ that one! :rofl: )

Torinde:
[@rderooy ](https://github.com/dosbox-staging/dosbox-staging/issues/2795#issuecomment-1696821602)
> I really dislike the "shorthand" notation. I consider it namespace pollution. In DOSBox-X it is disabled by default for that very reason.

Disabled by default in the source code or in a .conf setting? If there is the same setting in DOSBox-X I suggest to synchronize the labels.

johnnovak:
> If there is the same setting in DOSBox-X I suggest to synchronize the labels.

We can _consider_ it, but if the DOSBox-X setting name is not to our liking, then we won't do it and just use our own more descriptive name. Also, X setting names contain spaces, so I don't think this is a practical idea anyway.

kcgen:
> If there is the same setting in DOSBox-X I suggest to synchronize the labels.

Most of our overlap is in upstream settings (and I agree, if there's a reasonable chance to have stuff in common: then absolutely!). 

X has allowed expanded conf notation (or relaxation), introduced from patches and PRs without forcing them back into original compliance, like allowing dots as a pseudo-category notation, introduced by the MT32 patch (mt32.key = val), as well a spaces in keys.

We'd like to adopt a standard INI parser one day (there's a handful of nice header-only implementations), and similarly hold to these tighter standards.

I also think this specific item fits inside the [dosbox] category, because the whole conf system and how it works is dosbox-related (both 'config' and the shorthand are 100% dosbox'isms), and less about DOS itself (which is where things like long filename support or tweaks to how we comply with DOS behavior sit). 

So there's not much chance for overlap here. 
But it's good to see that IMFC has recent 100% overlap 👍 (awesome!)

johnnovak:
> So with us not adopting spaces (or putting this in the DOS section), there's probably not much chance to overlap here.
> But on a positive note - it's good to see that IMFC has recent 100% overlap :high_five: (awesome!)

Yeah, how I see it coming up with good descriptive config names for Staging is the primary concern, cross-compatibility of enhancements with any other DOSBox fork/variant less so. In other words, I would rather introduce our own good name than to just use any other forks so-so name (not saying this X name is particularly bad, just in general).

Torinde:
Yes, I understand your reasoning. Pity there will be a divergence, but it's a minor one.

I don't think it's worth the effort for _this setting_, but maybe useful for some future case ( - sharing an idea for a middle ground: support a X label "silently" as an alias:
- in the .conf file you put only `allow_shell_config_shortcuts`
- when processing the file (or using config -set, etc.), Staging accepts both `allow_shell_config_shortcuts` and `shell configuration as commands`

johnnovak:
> Yes, I understand your reasoning. Pity there will be a divergence, but it's a minor one.
> 
> I don't think it's worth the effort for _this setting_, but maybe useful for some future case - sharing an idea for a middle ground: support a X label "silently" as an alias:
> 
>     * in the .conf file you put only `allow_shell_config_shortcuts`
> 
>     * when processing the file (or using config -set, etc.), Staging accepts both `allow_shell_config_shortcuts` and `shell configuration as commands`

My view is that when it comes to base features, yes, compatibility is important with "legacy" core DOSBox settings, mainly for eXoDOS.

But for enhanced features, even if both Staging and fork XYZ have _vaguely_ similar features, it does not mean all details about the implementation are the same. So vaguely similar features sharing the same setting name, but actually doing slightly (or very) different things implementation wise—I don't see how this a good thing. If anything, that's misleading.

Setting names are just the "tip of the iceberg". We're not aiming to achieve parity when it comes to advanced features; we do our own thing, and so do other forks. So this "let's copy the setting names" approaches the whole thing in a backwards fashion, that's not a good way to approach the enhanced features, IMO.

Ultimately, we don't want to tie ourselves to any other implementation; we want to do our best for any given enhanced feature X, what we think is best for our users. Even if today feature XYZ does the same thing as in X or some other fork, and we share the setting name, what if we want to to enhance it further, thus diverging in functionality? Would the share setting name still make sense?

So, that's a lot of words about the subject 😄, but overall, I think we should do our own thing when it comes to enhanced features, and do what makes sense for us. Our current list of enhanced features, and DOSBox-X list of extra settings (which are numeours) already make interoperability basically next to impossible, or at the very least extremely impractical, barring perhaps legacy "core" settings.

So, IMO, we should drop this idea 😄

src/dosbox.cpp
FeralChild64: After reading the whole discussion I have rewritten the description. Feel free to reopen if you have more remarks.

FeralChild64:
Regarding spaces in key names - I'm strongly against them for two reasons:
- spaces in command parameters are always tricky in most OS'es, and we want to keep the `config -set`
- INI syntax highlighting engines often get confused with this syntax, GitHub being a prominent example:

```ini
[snafucation]
enabled = true
snafucation value = 0xDEADBEEF
```

src/shell/shell_cmds.cpp
```diff
@@ -129,25 +129,39 @@ bool lookup_shell_cmd(std::string name, SHELL_Cmd &shell_cmd)
...
-bool DOS_Shell::CheckConfig(char *cmd_in, char *line) {
+bool DOS_Shell::ExecuteConfigChange(const char* const cmd_in, const char* const line)
+{
+	assert(control);
+	const auto section_dos = static_cast<Section_prop*>(
+	        control->GetSection("dos"));
+	assert(section_dos);
+	if (!section_dos->Get_bool("shell_config_shortcuts")) {
+		return false;
+	}
```
kcgen, line 140: Suggest putting this in the `[dosbox]` section because it's not DOS emulation related (yes: these are commands in the shell, but it's not DOS internal behavior - like memory / timing / interrupts / file handling / etc.).

kcgen (approved):
Nice quick and tidy addition, @FeralChild64!

kcgen:
(All comments addressed; will merge shortly)

PR #2808, opened 2023-08-31
Video related refactorings

johnnovak:
Just some non-functional refactorings that I want to merge separately (part of some upcoming adaptive CRT shader enhancement I've been working on for EGA modes on VGA; stay tuned).

Less noise in my actual functional PR this way. Plus reviewing multiple refactoring PRs followed by one much smaller functional PR is less taxing psychologically on reviewers 😅 

Suggest to review by commit.


include/vga.h
```diff
@@ -923,75 +869,93 @@ struct VGA_Dac {
...
+	VgaChanges changes = {};
 #endif
-	VGA_LFB lfb = {};
+
+	VgaLfb lfb = {};
+
 	// Composite video mode parameters
-	int ri = 0, rq = 0, gi = 0, gq = 0, bi = 0, bq = 0;
+	// TODO enclose in struct
+	int ri        = 0;
```
FeralChild64, line 953: Would it be feasible to change `int`s to some type of fixed size, like `int32_t`?

src/hardware/vga_draw.cpp
```diff
@@ -2876,7 +2864,7 @@ void VGA_SetOverride(bool vga_override) {
 			vga.draw.vga_override=true;
 		} else {
 			vga.draw.vga_override=false;
-			vga.draw.width=0; // change it so the output window gets updated
+			vga.draw.render.width=0; // change it so the output window gets updated
```
FeralChild64, line 2867: Style: missing spaces around `=`

src/hardware/vga_memory.cpp
```diff
@@ -260,15 +263,15 @@ class VGA_UnchainedEGA_Handler : public VGA_UnchainedRead_Handler {
...
+		VgaLatch pixels;
 		pixels.d=((uint32_t*)vga.mem.linear)[start];
 		pixels.d&=vga.config.full_not_map_mask;
 		pixels.d|=(data & vga.config.full_map_mask);
 		((uint32_t*)vga.mem.linear)[start]=pixels.d;
 		uint8_t * write_pixels=&vga.fastmem[start<<3];
 
 		uint32_t colors0_3, colors4_7;
-		VGA_Latch temp;temp.d=(pixels.d>>4) & 0x0f0f0f0f;
+		VgaLatch temp;temp.d=(pixels.d>>4) & 0x0f0f0f0f;
```
FeralChild64, line 274: Style: missing spaces around `=` and `>>`, assignment should go to the next line.

FeralChild64:
Nice cleanup - just a few very minor remarks from my side.

src/hardware/vga.cpp
```diff
@@ -483,3 +483,7 @@ void SVGA_Setup_Driver(void) {
 	}
 }
 
+VideoMode VGA_GetCurrentVideoMode() {
+	return vga.draw.render.video_mode;
```
weirddan455, line 487: Not a big fan of getter functions that don't actually do anything.

include/video.h
```diff
@@ -200,6 +200,21 @@ struct RenderParams {
...
+
+	constexpr bool operator==(const RenderParams& that) const
+	{
+		return (width != that.width || height != that.height ||
+		        double_width != that.double_width ||
+		        double_height != that.double_height ||
+		        pixel_aspect_ratio != that.pixel_aspect_ratio ||
+		        pixel_format != that.pixel_format ||
+		        video_mode != that.video_mode);
+	}
```
weirddan455, line 214: Nice improvement.  There's some code in video capture that can use this.

weirddan455: Although wait... aren't these checks inverted?  Should be == and &&.

src/hardware/vga.cpp
kcgen: The value here the API firewall between the VGA module's struct and external modules that want the `VideoMode` (maybe like the rendering code).

We definitely don't want non-VGA module code having to understand the massive `vga.*.*.*` struct layout and poking into it. 

Similar to sdlmain's `GFX_<action>` module calls. We similar don't want the MOUSE_ module to go fetching values deep out of the `sdl.desktop.*.*` structs.

It's a de-spaghettifier :-)

weirddan455: That would be fair enough reasoning to add a getter function.  I don't see any usage code in this PR though so there's no way to tell if that reasoning applies here or not.  It's just adding a getter function for no apparent reason.

include/video.h
kcgen: If you're confident that each struct members can be compared with it's own `==` operator, you can use the compiler to generate the comparison operator:

```c++
struct MyStruct {
    int x;
    double y;

    bool operator==(const MyStruct& that) const = default;
};
```

This will create a body that runs `==` on each member against the other.

src/hardware/vga_memory.cpp
```diff
@@ -167,15 +170,15 @@ class VGA_ChainedEGA_Handler final : public PageHandler {
...
+		VgaLatch pixels;
 		vga.mem.linear[start] = val;
 		start >>= 2;
 		pixels.d=((uint32_t*)vga.mem.linear)[start];
 
 		uint8_t * write_pixels=&vga.fastmem[start<<3];
 
 		uint32_t colors0_3, colors4_7;
-		VGA_Latch temp;temp.d=(pixels.d>>4) & 0x0f0f0f0f;
+		VgaLatch temp;temp.d=(pixels.d>>4) & 0x0f0f0f0f;
```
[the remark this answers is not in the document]
johnnovak: Yeah, when I do renames, I only do the rename if I don't touch the other file for anything else.

I guess we can reformat that single line in such cases, but seems to be moree trouble than worth? (Too much noise when bisecting, etc.)

include/video.h
```diff
@@ -40,6 +41,168 @@ enum class IntegerScalingMode {
...
+// Details about the rendered image.
+// E.g. for the 320x200 256-colour 13h VGA mode with double-scanning
+// and pixel-doubling enabled:
+//   - width = 320 (will be pixel-doubled post-render via double_width)
+//   - height = 400 (2*200 lines because we're rendering scan-doubled)
+//   - pixel_aspect_ratio = 5/6 (1:1.2) (because the PAR is meant for the
+//     final image, post the optional width & height doubling)
+//   - double_width = true (pixel-doubling)
+//   - double_height = false (we're rendering scan-doubled)
+struct RenderParams {
```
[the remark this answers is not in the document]
johnnovak: No can't do because of circular reference problems, plus it gets propagated and stored in other parts.

It is a public structure that stores information about the rendered image, `render.cpp` "gives out" this info to other parts of the system.

But on a technical level, it needs to go into some higher level include file to break the circular reference problem. C/C++ includes and preprocessor magic... so 70s 🥶

src/hardware/vga.cpp
johnnovak: Exactly what @kcgen said. The usefulness of functions are not measured by their line count or their verb prefix 😄 Clear module boundaries is what matters.

I'm not using it here, and I might have ended up not using it later, but inclined to leave it anyway as sooner or later it will be needed. Previously other modules started mucking around in the VGA structs..

include/video.h
johnnovak: Uhhuhh, good pickup. I guess things keep working but do potentially a lot more work after this change 😅 (it's used to only recreate/re-render stuff when there's a change)

That defaulting thing seems handy; I was looking for it but couldn't find it. 🚀

src/hardware/vga.cpp
johnnovak: (But, in general I quite hate getters on objects @weirddan455 so I think I understand where you're coming from. Like these IDEs that auto-magically generate getters/setter for _every single member of a class. That's very questionable. I'm of the objects communicate via _messages_ which are high level and loose, not directly poking into each othernals via getters/setters... That's crap.)

include/video.h
johnnovak: Maybe another day @kcgen 😅 

```
warning: defaulted comparison operators are a C++20 extension [-Wc++20-extensions]
```

src/hardware/vga.cpp
johnnovak: Dropped due to popular demand (unused even in my next upcoming PR). Will reinstate when it's needed.

include/video.h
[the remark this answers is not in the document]
johnnovak: (`vga.h` and `render.h` both include each other; so common stuff needs to go to a third file)

include/vga.h
johnnovak: I've done that plus some other minor cleanup. But note I'm only touching these files accidentally because of the renames—I'm not actually working on these other files.

src/hardware/vga_draw.cpp
johnnovak: This is just a rename; not touching this file otherwise.

src/hardware/vga_memory.cpp
johnnovak: Just a rename.

include/video.h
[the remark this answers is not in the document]
johnnovak: > I would say move it back or rename it.

Actually, renaming it to `ImageParams` is the best solution, you're right. It's generic enough and it's used for slightly different purposes. Done.

include/video.h
kcgen: Argh; darn. I was able to use a default `=` operator in:
```c++
MidiWork& operator=(MidiWork&&) = default;
Section& operator=(Section&& other) = default;
Section_line& operator=(Section_line&& other) = default
```

But I guess comparisons weren't ready for C++17 at the time :-)

src/hardware/vga_draw.cpp
```diff
@@ -148,15 +148,15 @@ static uint8_t byte_clamp(int v)
...
+static uint8_t *Composite_Process(uint8_t border, uint32_t blocks, bool double_width)
 {
 	static int temp[SCALER_MAXWIDTH + 10] = {0};
 	static int atemp[SCALER_MAXWIDTH + 2] = {0};
 	static int btemp[SCALER_MAXWIDTH + 2] = {0};
 
 	int w = blocks * 4;
 
-	if (doublewidth) {
+	if (double_width) {
```
kcgen, line 162: Not comment on the code change - just the commit message, suggest being a bit more descriptive, `"Rename a VGA function variable"` instead of `Cleanup`, cause it can get a bit ambiguous reading history:

include/video.h
weirddan455: @kcgen Well, it goes back to C I think.  You could always use `=` to copy values from one struct to another but `==` was a compile error in plain C.  So I think default assignment operator would have been around long before C++ 17.

src/hardware/vga_draw.cpp
johnnovak: Sure, I always struggle with those, especially if lots of cleanup stuff gets bundled together 😄 Changing this particular one.

include/video.h
kcgen: Ahh indeed - Thanks for clarifying, @weirddan455 :+1:

include/vga.h
```diff
@@ -838,79 +768,95 @@ struct VGA_Seq {
...
 
-	uint8_t index = 0;
+	uint8_t index  = 0;
 	bool read_only = false;
 };
 
-struct VGA_Gfx {
-	uint8_t index = 0;
-	uint8_t set_reset = 0;
+struct VgaGfx {
```
kcgen, line 833: very nice standarding of the struct names :+1:

include/vga.h
```diff
@@ -923,76 +869,96 @@ struct VGA_Dac {
...
+	VgaCrtc crtc   = {};
+	VgaGfx gfx     = {};
+	VgaDac dac     = {};
+	VgaLatch latch = {};
+	VgaS3 s3       = {};
+	VgaSvga svga   = {};
+	VgaHerc herc   = {};
+	VgaTandy tandy = {};
+	VgaOther other = {};
+	VgaMemory mem  = {};
```
kcgen, line 934: wow.. looks great!

src/hardware/vga_other.cpp
```diff
@@ -628,12 +631,15 @@ static void update_cga16_color()
...
-	vga.bi = static_cast<int>(bi * iq_adjust_i + bq * iq_adjust_q);
-	vga.bq = static_cast<int>(-bi * iq_adjust_q + bq * iq_adjust_i);
+	// clang-format off
+	vga.composite.ri = static_cast<int32_t>( ri * iq_adjust_i + rq * iq_adjust_q);
+	vga.composite.rq = static_cast<int32_t>(-ri * iq_adjust_q + rq * iq_adjust_i);
+	vga.composite.gi = static_cast<int32_t>( gi * iq_adjust_i + gq * iq_adjust_q);
+	vga.composite.gq = static_cast<int32_t>(-gi * iq_adjust_q + gq * iq_adjust_i);
+	vga.composite.bi = static_cast<int32_t>( bi * iq_adjust_i + bq * iq_adjust_q);
+	vga.composite.bq = static_cast<int32_t>(-bi * iq_adjust_q + bq * iq_adjust_i);
+	// clang-format on
```
kcgen, line 641: :+1: very nice and a great use of the format off to get vertical alignment, when all the types and meanings are the same.

kcgen (approved):
That's a lot of cleanup to get names standardized across all these file. Looks great, @johnnovak!

johnnovak:
Thanks for the review @kcgen @FeralChild64 @kcgen 👍🏻

PR #2815, opened 2023-09-01

Kappa971:
I would like to point out:
In `PROGRAM_IMGMOUNT_HELP_LONG`:
```
  [color=cyan]BOOTIMAGE[reset] is a bootable disk image with specified -size GEOMETRY:
            bytes-per-sector,sectors-per-head,heads,cylinders
```
And in `PROGRAM_IMGMOUNT_SPECIFY_GEOMETRY`:
```
bytes_per_sector, sectors_per_cylinder, heads_per_cylinder, cylinder_count.
```

Shouldn't they be indicated the same way?


kcgen:
> Shouldn't they be indicated the same way?

Yes, they are both text / documentation, so should use the same notation; I've pushed a fixed for that.

Kappa971:
> Yes, they are both text / documentation, so should use the same notation; I've pushed a fixed for that.

Also here:
[color=blue]IMGMOUNT drive-letter location-of-image -size **bps,spc,hpc,cyl]**[reset]

kcgen:
@Kappa971 , can you fix the conflicts? (The 'Resolve conflicts' button below will give you a web editor to update the lines)

kcgen (approved):
Looks good @Kappa971 !

Issue #2812, opened 2023-08-31
ZMBV Video Encoder Performance Regression


weirddan455:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Linux x86_64

### If Other OS, please describe

Arch Linux

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

I had complained previously about bad performance of the current ZMBV video encoder.  Today I grabbed the old stable version v0.80.1 and tried there and performance was fine.  I then did a git bisect and found the first bad commit 0d01f906ed9baec649146d62f46dedcf165229f6

This was tested in Release builds recording Doom.

```
0d01f906ed9baec649146d62f46dedcf165229f6 is the first bad commit
commit 0d01f906ed9baec649146d62f46dedcf165229f6
Author: kcgen <kcgen@users.noreply.github.com>
Date:   Mon Apr 10 17:04:27 2023 -0700

    Eliminate vgaonly-specific handling when checking for line-doubling

 src/hardware/vga_draw.cpp | 18 ++++++++++--------
 1 file changed, 10 insertions(+), 8 deletions(-)
```

It's related to this line-doubling thing which I don't really understand.  I just checked the output videos as I was typing this and it looks the bad performance may simply be due to encoding a higher res video.  The ones that made the game all laggy were 640X400 vs 320X200 for the ones that performed well in game.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Start Doom (or probably any somewhat demanding VGA game)
2. Ctrl + F7 to start video capturing
3. Game starts stuttering badly

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Yes, it's normal as it stands now. My capture fixes will take care of this by dividing it back to 320x200 prior to passing the frame to the vide capturer (I'm doing that for upscaled image captures already).

https://github.com/dosbox-staging/dosbox-staging/issues/2734 will fix it and restore the exact v0.80.1 performance.

weirddan455:
@johnnovak The video encoder is already being passed a 320X200 frame.  It looks for these `double_width` and `double_height` flags and does the doubling inside the video capture code here:

[...]

PR #2817, opened 2023-09-01
Offload ZMBV video encoder to worker thread

weirddan455:
Not ready for merge or review.  I banged this out in like an hour and a half and just wanted to post this as a starting point.

The threading logic is fine I think but it is just a band-aid for poor encoder performance and DOSBox reacts violently when you rip the band-aid off.

On my machine it goes like this.  Start Doom, start recording.  If you stand still, everything is good.  In single-threaded, this was a laggy mess so it's kind of better.

If you start moving around a lot, the queue quickly reaches max capacity at which point the main thread stalls while it waits for the worker thread to clear out the queue.  This is a stall for a few seconds.  Not good :sob: 

Sometimes if the main thread gets stalled for long enough,  it doesn't recover.  I think something weird may be happening in the OPL sound code because when I turn OPL off, the main thread recovers fine after the stall.

So, like I thought might happen.  Pretty straight forward to offload to a worker thread but it doesn't solve the real problem.

### DIFF
--- a/include/rwqueue.h
+++ b/include/rwqueue.h
@@ -70,7 +70,10 @@ public:
 	// non-blocking call
 	size_t Size();
 
-	// non-blocking calls
+	// non-blocking call
+	void Start();
+
+	// non-blocking call
 	void Stop();
 
 	// non-blocking call
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -19,7 +19,9 @@
  *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+#include "capture_video.h"
 #include "capture.h"
+#include "rwqueue.h"
 
 #include <cassert>
 
@@ -30,10 +32,10 @@
 
 #include "zmbv/zmbv.h"
 
-static constexpr auto NumSampleFramesInBuffer = 16 * 1024;
-
 static constexpr auto SampleFrameSize  = 4;
 static constexpr auto NumAudioChannels = 2;
+
+static constexpr auto SampleByteSize = sizeof(int16_t);
 
 static constexpr auto AviHeaderSize = 500;
 
@@ -45,6 +47,7 @@ static struct {
 	int width                = 0;
 	int height               = 0;
 	PixelFormat pixel_format = {};
+	ZMBV_FORMAT zmbv_format  = {};
 	float frames_per_second  = 0.0f;
 
 	uint32_t written           = 0;
@@ -54,13 +57,15 @@ static struct {
 	uint32_t index_used        = 0;
 
 	struct {
-		int16_t buf[NumSampleFramesInBuffer][NumAudioChannels] = {};
-
 		uint32_t sample_rate     = 0;
 		uint32_t buf_frames_used = 0;
 		uint32_t bytes_written   = 0;
 	} audio = {};
 } video = {};
+
+static ZmbvWorkUnit main_thread_work = {};
+static std::thread worker_thread     = {};
+static RWQueue<ZmbvWorkUnit> worker_queue{256};
 
 static void add_avi_chunk(const char* tag, const uint32_t size,
                           const void* data, const uint32_t flags)
@@ -100,11 +105,8 @@ static void add_avi_chunk(const char* tag, const uint32_t size,
 	host_writed(index + 12, size);
 }
 
-void capture_video_finalise()
+static void finish_encode_job()
 {
-	if (!video.handle) {
-		return;
-	}
 	if (video.codec) {
 		video.codec->FinishVideo();
 	}
@@ -265,6 +267,17 @@ void capture_video_finalise()
 	video.handle = nullptr;
 }
 
+void capture_video_finalise()
+{
+	if (!video.handle) {
+		return;
+	}
+	worker_queue.Stop();
+	if (worker_thread.joinable()) {
+		worker_thread.join();
+	}
+}
+
 void capture_video_add_audio_data(const uint32_t sample_rate,
                                   const uint32_t num_sample_frames,
                                   const int16_t* sample_frames)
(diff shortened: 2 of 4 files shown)


johnnovak:
Seems like an easy low-hanging fruit to claim @weirddan455 , great job. I fully agree, it needs an eventual replacement, but until then it's a low effort way improve the situation, so why not (if it's solid, or not _less solid_ than how it was before 😅).

weirddan455:
> Seems like an easy low-hanging fruit to claim @weirddan455 , great job. I fully agree, it needs an eventual replacement, but until then it's a low effort way improve the situation, so why not (if it's solid, or not _less solid_ than how it was before 😅).

Well, no it is less solid.  In single threaded mode it's a laggy mess (at least until you lower the resolution) but at least things work.

In multithreaded mode, sometimes the game just locks up.  In Doom, you can't move and it just loops the same half-second of sound over and over.  Video capture keeps working though and records the freeze.  It's weird.  I'm thinking it's timing sensitive code (probably sound code but who knows) that doesn't like stalling for seconds at a time.

johnnovak:
Oh right @weirddan455 . Reading comprehension fail on my part—haven't had my morning coffee yet 😄 

Yeah, you might wanna timebox it and not waste too much time on it, but of course up to you. That's what I'd do. Ultimately, we should analyse a bit what the drawback of depending on FFmpeg really is. Maybe we can make it mandatory for video captures; no FFmpeg, no capture, and then we can offload the whole threaded capture to FFmpeg.

I don't want to polish a dodgy implementation or carry it around and waste valuable time on it when there's a clearly superior solution. What would be the drawback of mandating FFmpeg for video captures? @kcgen @FeralChild64 ? It's a pretty standard tool in Linux land, right? And on macOS and Windows it's 10 MB extra fat to the ZIP file, who cares.

weirddan455:
Well ZMBV performs a lot better when you lower the resolution to how it is in 0.80.1.  I also haven't tested FFmpeg at any higher resolutions so you can't really replace working code with something that's not finished yet.

Btw, I was able to reproduce the crash with this simple patch on main with no video capture going.

```
diff --git a/src/gui/render.cpp b/src/gui/render.cpp
index 5e69f8c07..db1fbf02d 100644
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -296,6 +296,13 @@ void RENDER_EndUpdate(bool abort)
                CAPTURE_AddFrame(image, frames_per_second);
        }
 
+       const Uint8 *keyboard_state = SDL_GetKeyboardState(nullptr);
+       if (keyboard_state[SDL_SCANCODE_Q]) {
+               struct timespec sleep_time {};
+               sleep_time.tv_nsec = 50000000; // 50 ms
+               nanosleep(&sleep_time, nullptr);
+       }
+
        if (render.scale.outWrite) {
                GFX_EndUpdate(abort ? nullptr : Scaler_ChangedLines);
        } else {
```

Just hold Q for a couple seconds to simulate something holding up the main thread.  Game locks up just the same.

I guess it's more of a soft-lock technically.  Doom stops responding but DOSBox itself is not locked up.

johnnovak:
> Well ZMBV performs a lot better when you lower the resolution to how it is in 0.80.1. I also haven't tested FFmpeg at any higher resolutions so you can't really replace working code with something that's not finished yet.

Of course, I was talking about what to do going forward.

1. We do nothing for ZMBV, stays as it is and only works on raw, non-upscaled content. Legacy behaviour, should continue to work indefinitely.
2. Try later ZMBV with FFmpeg, with upscaled content as well. See how it goes, but potentially we might the same issues as you identified. In which case we'll decide which path we'll take (mandate FFmpeg or not)
3. Making the current ZMBV threaded but flaky is clearly not an option.

Ultimately, we'll see how this goes but can you do things in a way to not alter existing functionality too much? E.g. even if merge experimental FFmpeg support or threaded ZMBV support, we should be able to disable it with a compile-time switch or some hardcoded value somewhere so if this thing drags on, it won't delay 0.81.0 any further.

So yeah, please just keep that in mind; make things so that they can be merged but disabled. Worst case scenario: all this new capture code is in main, we ship with it, but it's hard disabled until we stabilise it for 0.82.0.

johnnovak:
I will also give you a pack of some demoscene productions @weirddan455 that use software rendering at 640x480 in 15/16-bit true colour or 24-bit true colour and require 300k+ cycles to run at full FPS. Those will stress test your capture code quite nicely, and capturing demoscene productions is a valid use case. Will do that today or tomorrow (they're freeware).

kcgen:
This is brilliant work @weirddan455!

An alternative might be to offload ZMBV ... into the bucket. 

ZMBV is /the/ dosbox format itself.. so old timers and DOS forensic pixel analysis might 😱 scream (don't take my lossless codec away!), but we're now 20 years beyond when ZMBV filled a necessary role. 

These days, h.264, h.265, and AV1 all have lossless modes. 

Likewise, these codecs support proper source and display aspect handling, unlike ZMBV+ AVI, which presents the results as always broken. 

There's also efficiency: 
 - hardware-playback decode and encode are now common (which we get for free with ffmpeg) and continues to be standard even on embdded systems too weak to effectively run dosbox itself!

 - Nvidia, Intel, and the Raspberry Pi foundation are unfortunately not making progress in ZMBV hardware decode and encode; that ship has sailed.

 - even at the software level, x264 lossless (-qp 0 or -crf 0) runs at > 4x realtime rate for me (270 fps), but ZMBV often is sub-realtime, just as you're seeing here.

weirddan455:
>  even if merge experimental FFmpeg support or threaded ZMBV support, we should be able to disable it with a compile-time switch or some hardcoded value somewhere so if this thing drags on, it won't delay 0.81.0 any further.

This is already done for FFmpeg.  There is both a compile-time and a run-time switch.  Run time is set to off by default (changeable in the config file).

> 1. We do nothing for ZMBV, stays as it is and only works on raw, non-upscaled content. Legacy behaviour, should continue to work indefinitely.
> 2. Try later ZMBV with FFmpeg, with upscaled content as well. See how it goes, but potentially we might hit the same issues as you identified. In which case we'll decide which path we'll take (mandate FFmpeg or not)
> 3. Making the current ZMBV threaded but flaky is clearly not an option.

I say probably Option 1.  Do nothing with ZMBV for 0.81 and put threading on the backburner for after release.  FFmpeg can either go in as an optional feature or wait for 0.82 as well.

johnnovak:
> > even if merge experimental FFmpeg support or threaded ZMBV support, we should be able to disable it with a compile-time switch or some hardcoded value somewhere so if this thing drags on, it won't delay 0.81.0 any further.
> 
> This is already done for FFmpeg. There is both a compile-time and a run-time switch. Run time is set to off by default (changeable in the config file).

Like a pro 😎 It's great working with you, man 🚀 

> > 1. We do nothing for ZMBV, stays as it is and only works on raw, non-upscaled content. Legacy behaviour, should continue to work indefinitely.
> > 2. Try later ZMBV with FFmpeg, with upscaled content as well. See how it goes, but potentially we might hit the same issues as you identified. In which case we'll decide which path we'll take (mandate FFmpeg or not)
> > 3. Making the current ZMBV threaded but flaky is clearly not an option.
> 
> I say probably Option 1. Do nothing with ZMBV for 0.81 and put threading on the backburner for after release. FFmpeg can either go in as an optional feature or wait for 0.82 as well.

I'd vote on Option 1 too.

I'm preparing the demoscene pack for you now because I got excited. I love demos! 😎 🤘🏻 Will post a download link here in 30 mins, I just gotta watch them all first 😎

kcgen:
Also onboard with option 1.

Ffmpeg itself supports zmbv encode: so we can still dump it from our codebase  and shunt it through weirddan455's sweet threaded ffmpeg route.

weirddan455:
@kcgen Yep, I agree with all of that.  We could probably dump ZMBV and say FFmpeg is the future.  Hell FFmpeg even has a ZMBV encoder built in that is probably more performant than this in-tree code.

I think it should stay as is for next release though.  For 0.81.0, if FFmpeg gets merged, it's an extra (maybe an expeirmental off-by-default extra).  Then once all the kinks get worked out, maybe we drop ZMBV.

johnnovak:
> An alternative might be to offload ZMBV ... into the bucket. ♻️

Quite so, and I agree with rest of your very informative assesment of the year 2023 situation 😄

johnnovak:
Demoscene pack @weirddan455 
https://drive.google.com/file/d/1BLAEJ1mjbtHWGDfQKGc0n30oArVFZWxH/view?usp=sharing

Use the included configs per demo, run with `dosbox --noprimaryconf` ideally, and definitely use the release builds as these are quite demanding and performance is dogshit for debug builds 😅 

image

Some notes:

- Don't enable vsync unless you have a VRR monitor; half of these demos switch between weird refresh rates, so no single fixed refresh rate can work with vsync.

- **Blasphemy – Kkowboy** is the most demanding; needs 450k+ cycles and renders 512x384 24-bit true-colour output @ 70 Hz. It's also a stress test for audio encoders because of the hard-panned synthetic instruments with lots of sharp transients a good does of high-freq content. You can also start the demo with `kkowboy.exe hicolor` to run in 16-bit high-colour mode instead.

- **TPOLM - TE-2RB** is similar; it's 640x480 / 256-colour, but greyscale, and the noise overlays really stress lossy video encoders. Music is the same category, quite stressful for lossy encoders 😅 

- **Sunflower – Zilog** is 320x200 24-bit true-colour @ 80 Hz by default. You can launch it with `-bpp32`, `-bpp24`, `-bpp16`, and `-bpp15` to test those hi/true colour VESA modes. Very handy! You can also try `-fake` and `-fuck` for fun for that fake ~12bpp mode emulated via a 256-colour palette popular in demoscene circles before hi/true colour VESA cards became widespread (it should stress lossy encoders like crazy; zoom in and you'll see 😎)

- **Majic 12 – Show** switches weird custom VGA modes like crazy (that my old group's award-winning production 😎 🤘🏻) and also refresh rates (60 Hz, 63 Hz, 70 Hz, and maybe something else too). This actually stresses my auto-switcher too and the glitches are noticeable on screen mode changes. But it's a small price to pay to get zero-config automatic CRT shaders for games that very infrequently switch modes (well, most of them), and we'll improve it in the future.

I guess the gold standard would be to be able to encode Kkowboy at 4k with no performance drop, taking advantage of the idle cores in the system (well, one should aim high, right? 😅).

This should keep you busy for a while, enjoy the fake electronic lightshow 😎 🤘🏻

weirddan455:
Thanks @johnnovak  That gave me tests for all the different pixel formats I needed.  Sunflower – Zilog crashed about a minute into the demo at -bpp15 though.

image0001

johnnovak:
> Thanks @johnnovak That gave me tests for all the different pixel formats I needed. Sunflower – Zilog crashed about a minute into the demo at -bpp15 though.
> 
> image0001

Yeah, don't worry about that. It also crashes with defaults at any cycles value above 130k.

kcgen:
@weirddan455, without ZMBV being able to keep up with realtime on typical hardware, I was originally worried that aspect might be a showstopper for this PR - despite all the improvements that it brings.

But one option is tightening the RW's queue to a single item, which essentially makes the encoding serial again.

So we could make this a conf option - where, if threaded, the RW's queues are sized generously, otherwise they're just one element.

Or we could (put more effort) into making the decision automatic. If the queue's fill up, warn: `"ZMBV: Encoding can't keep up with realtime; switching to serial encoding", at which points we drain the queues and re-instantiate them with one item.

I also see ZMBV as a dead-end on the deprecation pathway, just like `output = surface`, given h.264 et.al has a lossless mode *and* hardware decode and encode support *and* proper aspect handling. So a third option is simply stepping away and putting all effort into the FFMPEG route.

weirddan455:
>  without ZMBV being able to keep up with realtime on typical hardware, I was originally worried that aspect might be a showstopper for this PR - despite all the improvements that it brings.

You mean aspect ratio correction?  ZMBV has never supported that as far as I know.  From what you guys have told me, AVI doesn't support setting the SAR like MP4 does (this is all it took for FFmpeg).  We'd have to either switch container or resize the video ourselves, the latter of which probably comes with a performance hit.

>  But one option is tightening the RW's queue to a single item, which essentially makes the encoding serial again.

>  So we could make this a conf option - where, if threaded, the RW's queues are sized generously, otherwise they're just one element.

>  Or we could (put more effort) into making the decision automatic. If the queues fill up, we warn, `"ZMBV: Encoding can't keep up with realtime; switching to serial encoding", at which points we drain the queues and re-instantiate them with one item queues, and then carry on.

And what would be the point of this?  If the encoder can keep up in real time, it works just fine single threaded.  If it can't, it's laggy no matter what you do.  I played around with different queue sizes just now.  Queue size of 1 is constant lagginess.  It seems to be a slight perf improvement over single threaded but not much.  Queue size of 16 and it's smooth for about a second then stutter.  A conf option is just asking the user "How would you like your lag today sir?"

The only way to get real performance gains by threading is to have multiple worker threads.  Video encoding by nature is an easily to parallelize problem.  ZMBV seems to work line-by-line so theoretically you could just throw more CPUs at the problem until it goes away.  However, that's more work than I'm willing to put into it right now and has the potential to introduce some tricky threading bugs.

See #2812 for some discussion on performance.  @johnnovak said he was going to work on getting us back to v0.80.1 levels of performance which runs fine single threaded on my machine.  Ultimately, this just means dropping resolution so the encoder has less work to do.

>  I also see ZMBV as a dead-end on the deprecation pathway, just like output = surface, given h.264 et.al has a lossless mode and hardware decode and encode support and proper aspect handling. So a third option is simply stepping away and putting all effort into the FFMPEG route.

That is kind of where I'm at.  Unless someone else wants to take this code and run with it, I'm probably just going to close this PR.

johnnovak:
> That is kind of where I'm at. Unless someone else wants to take this code and run with it, I'm probably just going to close this PR.

Yeah, like we already said, and it was one of @kcgen 's options above: let's bring back the default ZMBV capture to parity with 0.80.1 and that's it. Then it will be replaced by something lossless by FFmpeg, or ZMBV by FFmpeg, but whatever. But it's wasted effort to improve it in a meaningful way at this point. Multiple worker threads are a must, yes; let's FFmpeg handle all that for us... later 😄

kcgen:
Sorry for the late reply on your question, @weirddan455: 

> You mean aspect ratio correction?

'aspect' was a bad adjective choice on my part. I should have said, "zmbv's less than real-time performance might be a showstopper"). 

But yeah, fully agree with all points you and @johnnovak made.

weirddan455:
I think this has been discussed enough and I'm not planning on trying to optimize ZMBV more so I'm just going to close this.

merge

PR #2818, opened 2023-09-02
Refine when `presentation mode = auto` uses throttled mode

kcgen:
A handful of small adjustments, each separate.

Suggest reviewing commit-by-commit, and running various text and GUI benchmarks with `--nodefaultconf` and with `output = openglnb`, to drive frame-rates as fast as possible.

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -485,40 +485,43 @@ static void populate_requested_vsync_settings()
 		sdl.vsync.when_fullscreen.requested = VsyncState::Yield;
 	} else {
 		assert(user_pref == "auto");
-		// In windowed-mode, assume the window manager has exclusive
-		// access to the GPU so leave vsync off
+
+		// In 'vsync = auto' windowed-mode, we try to disable vsync for
+		// our Window because enabling vsync both at the program and
+		// compositor levels (which might be enforced) may add latency
+		// (i.e,: extra time blocked inside rendering or presentation
+		// calls), which stalls DOSBox.
+		//
 		sdl.vsync.when_windowed.requested = VsyncState::Off;
 
-		// For fullscreen mode, VRR displays that perform
-		// frame interpolation need vsync enabled to lock onto the
-		// content. So we select "vsync on" whenever a display is
-		// /potentially/ this type. This has little downside because at
-		// >140+ Hz, all DOS refresh rates are going to be presented.
-		//
+		// In fullscreen-mode, the above still applies however we add
+		// handling for VRR displays that perform frame interpolation.
+		// as tehy need vsync enabled to lock onto the content:
 		const bool prefers_vsync_when_fullscreen =
 		        (get_host_refresh_rate() >= InterpolatingVrrMinRateHz);
 
-		// For fullscreen mode with non-interpolating VRR displays, we
-		// try to disable vsync because this:
-		//
-		// 1) has no downside when the DOS rate is below the host rate.
-		//    For example, in DOOM a DOS rate of 35 fps on a 60 Hz host
-		//    will never tear because it won't exceed the host rate.
-		//
-		// 2) is "less worse" when the *unique* DOS frame rate actually
-		//    exceeds the host rate. Why? When disabled, frames might
-		//    tear, but there is no slowdown or full frames dropped.
-		//    When enabled, entire frames will either be dropped or
-		//    the host will jam up with a rendering stall, drain the
-		//    audio buffer, and possible stutter or pop.
-		//
-		// 'auto' lets us make a judgement call for average user, so
-		// we accept the risk of tearing with the guarantee of no
-		// rendering stalls or pops (versus the opposite).
-		//
 		sdl.vsync.when_fullscreen.requested = prefers_vsync_when_fullscreen
 		                                            ? VsyncState::On
 		                                            : VsyncState::Off;
+
+		// In 'vsync = auto', we also /assume/ vsync is enabled
+		// (regardless how the above requests actually played out) by
+		// overriding the measured state, as follows:
+		//
+		sdl.vsync.when_windowed.measured   = VsyncState::On;
+		sdl.vsync.when_fullscreen.measured = VsyncState::On;
+		//
+		// A 60 Hz display can only show 60 complete frames per second.
+		// To achieve a higher frame rate, the host must drop or tear
+		// frames. This creates two layers of tearing when combined with
+		// DOS's (potentially) torn frames, which is common in games
+		// that don't use vblank. In "auto" mode, we aim for the best
+		// user experience, and dual-tearing isn't it. However, users
+		// can always set 'vsync = off'.
+		//
+		// There is no downside to making this assumption when the host
+		// display is faster than the DOS rate because all frames will
+		// be presented.
 	}
 }
 
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -465,12 +465,14 @@ static double get_host_refresh_rate()
 	return rate;
 }
 
-// Populates the vsync preferences from the user's conf setting. This is called
-// on-demand after startup because we need SDL running to assess the host's
-// frame rate in the "auto" vsync case.
+// Reset and populate the vsync settings from the user's conf setting. This is
+// called on-demand after startup and on output-mode changes (ie: switching from
+// an SDL-based drawing context to an OpenGL-based context).
 //
-static void populate_requested_vsync_settings()
+static void initialize_vsync_settings()
 {
+	sdl.vsync = {};
+
 	const auto section = dynamic_cast<Section_prop*>(control->GetSection("sdl"));
 	const std::string user_pref = (section ? section->Get_string("vsync") : "auto");
 
@@ -786,7 +788,7 @@ static VsyncSettings& get_vsync_settings()
 {
 	if (sdl.vsync.when_fullscreen.requested == VsyncState::Unset ||
 	    sdl.vsync.when_windowed.requested == VsyncState::Unset) {
-		populate_requested_vsync_settings();
+		initialize_vsync_settings();
 	}
 	return sdl.desktop.fullscreen ? sdl.vsync.when_fullscreen
 	                              : sdl.vsync.when_windowed;
@@ -1669,12 +1671,13 @@ uint8_t GFX_SetSize(const int width, const int height,
 
 	sdl.draw.callback = callback;
 
-	// If we're changing the rendering type, clear any previously measure
-	// vsync data because the rendering system (OpenGL, Metal, GLES, etc..)
-	// can have different vsync behavior and performance.
-	//
+	// If we're changing the SDL output type (ie: going from 'output =
+	// texture' to 'output = OpenGL'), then re-initialize our vsync settings
+	// because how the OS's compositor (if one exists) handles (or doesn't
+	// handle) this new type may be different, and thus warrants new
+	// measurements.
 	if (sdl.desktop.want_type != sdl.desktop.type) {
-		get_vsync_settings().measured = VsyncState::Unset;
+		initialize_vsync_settings();
 	}
 
 	switch (sdl.desktop.want_type) {
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -980,29 +980,14 @@ static void remove_window()
 	}
 }
 
-// After the last new frame, present a duplicate frame every N calls of this
-// function. This is an insurance policy to ensure the last frame is shown even
-// if the host previously dropped it. This is a temporary hack; it should be
-// removed in the future when everyone's PC display technology can support the >
-// 70 Hz refresh rates available in 1990s PC CRT monitors.
-static void present_new_or_maybe_dupe(const bool frame_is_new)
-{
-	static auto dupe_countdown = sdl.frame.vfr_dupe_countdown;
-	assert(dupe_countdown > 0);
-
-	if (frame_is_new || --dupe_countdown <= 0) {
-		sdl.frame.present();
-		dupe_countdown = sdl.frame.vfr_dupe_countdown;
-	}
-}
-
-// The throttled presenter skips frames that have inter-frame spaces narrower
-// than the allowed frame period, but also back-fills with duplicate frames as
-// an insurance policy to ensure the last frame is shown even if the host
-// previously dropped it.
-static void maybe_present_throttled_or_dupe(const bool frame_is_new)
+// The throttled presenter skips frames that have inter-frame spacing narrower
+// than the allowed frame period (sdl.frame.period_us). When a frame is skipped,
+// the presenter still tries to present it at its next oppourtunity.
+//
+static void maybe_present_throttled(const bool frame_is_new)
 {
 	static int64_t last_present_time = 0;
+	static auto was_new_and_throttled = false;
 
 	const auto now     = GetTicksUs();
 	const auto elapsed = now - last_present_time;
@@ -1012,7 +997,15 @@ static void maybe_present_throttled_or_dupe(const bool frame_is_new)
 		// this extra wait back by deducting it from the recorded time.
 		const auto wait_overage = elapsed % sdl.frame.period_us;
 		last_present_time = now - (9 * wait_overage / 10);
-		present_new_or_maybe_dupe(frame_is_new);
+
+		if (frame_is_new || was_new_and_throttled) {
+			sdl.frame.present();
+		}
+	}
+	// Otherwise we've had to throttle the frame, however if the frame was
+	// new, we'll record it as such and try to present it next time.
+	else {
+		was_new_and_throttled = frame_is_new;
 	}
 }
 
@@ -2389,9 +2382,13 @@ void GFX_EndUpdate(const uint16_t *changedLines)
 		case FrameMode::Cfr:
 			maybe_present_synced(frame_is_new);
 			break;
-		case FrameMode::Vfr: present_new_or_maybe_dupe(frame_is_new); break;
+		case FrameMode::Vfr:
+			if (frame_is_new) {
+				sdl.frame.present();
+			}
+			break;
 		case FrameMode::ThrottledVfr:
-			maybe_present_throttled_or_dupe(frame_is_new);
+			maybe_present_throttled(frame_is_new);
 			break;
 		case FrameMode::Unset:
 			break;


src/gui/sdlmain.cpp
```diff
@@ -1109,7 +1109,15 @@ static void setup_presentation_mode(FrameMode &previous_mode)
...
-		if (supported_rate >= dos_rate) {
+		// Text-based programs (including DOSBox itself) often re-draw
+		// text elements in VGA memory immediately on-demand without
+		// waiting for vblank. This means that throttled VFR's follow up
+		// frame has often already been modified (or is new) So for text
+		// modes, we avoid throttling.
+		//
+		const auto in_text_mode = VGA_InTextMode();
+
+		if (supported_rate >= dos_rate || in_text_mode) {
```
johnnovak, line 1114: Please let's not special case text modes in any way; we've discussed this before a few times and ultimately you agreed with me that video output is video output 😄 So let's not bring any text-mode specific special handling back.

An application that uses a graphics mode and updates the screen sporadically without waiting for vblank can manifest the _exact same issue_—there is zero difference, it's 100% immaterial whether we're in a text mode or a graphics mode, the "update patterns" matter only. Image viewers fall into this category, but also the in-game menu of many games where the screen is completely static (if they don't do vsync).

src/hardware/vga.cpp
```diff
@@ -70,7 +70,20 @@ void VGA_SetMode(VGAModes mode) {
...
+{
+	switch (vga.mode) {
+	case M_TEXT:
+	case M_HERC_TEXT:
+	case M_TANDY_TEXT:
+	case M_CGA_TEXT_COMPOSITE: return true;
+
+	default: return false;
+	}
+}
```
johnnovak, line 83: This is not needed as I introduced `INT10_IsTextMode` for this a while ago. But let's not do any text-mode only special casing, as per my other comment.

src/gui/sdlmain.cpp
```diff
@@ -485,9 +485,10 @@ static void populate_requested_vsync_settings()
 		sdl.vsync.when_fullscreen.requested = VsyncState::Yield;
 	} else {
 		assert(user_pref == "auto");
-		// In windowed-mode, assume the window manager has exclusive
-		// access to the GPU so leave vsync off
-		sdl.vsync.when_windowed.requested = VsyncState::Off;
+		// In windowed-mode, assume modern compositors are using some
+		// form of vsync or frame limiting to prevent visual tearing of
+		// the desktop itself.
+		sdl.vsync.when_windowed.requested = VsyncState::On;
```
johnnovak, line 491: Yes, the comment is correct and that's the case, but that means we don't need to enable vsync ourselves in windowed mode.

We just send frames whenever in windowed mode, and it will be vsync'ed by the compositor of the OS, as if "by magic".

That's my experience on Windows with other emulators; you absolutely don't _need_ to manually enable vsync in windowed mode. Well, probably it's a no-op anyway, but I'll leave this off for clarity.

src/gui/sdlmain.cpp
```diff
@@ -4228,8 +4229,8 @@ void config_add_sdl() {
 
 	pstring->Set_help(
 	        "Set the host video driver's vertical synchronization (vsync) mode:\n"
-	        "  auto:      Limit vsync to beneficial cases, such as when using an\n"
-	        "             interpolating VRR display in fullscreen (default).\n"
+	        "  auto:      Limit vsync to beneficial cases, such as when in Windowed-mode\n"
```
johnnovak, line 4229: No, vsync in windowed mode just happens on all modern OSes. It's impossible to get tearing in windowed mode in any modern OS, no matter how hard you try. So the previous comment and behaviour was correct.

src/gui/sdlmain.cpp
```diff
@@ -2392,9 +2385,13 @@ void GFX_EndUpdate(const uint16_t *changedLines)
...
 			break;
-		case FrameMode::Vfr: present_new_or_maybe_dupe(frame_is_new); break;
+		case FrameMode::Vfr:
+			if (frame_is_new) {
+				sdl.frame.present();
+			}
+			break;
 		case FrameMode::ThrottledVfr:
-			maybe_present_throttled_or_dupe(frame_is_new);
+			maybe_present_throttled(frame_is_new);
```
johnnovak, line 2394: Yes, so this commit is only what we need from this PR, I think. And we _absolutely_ want this for _any_ screen mode, not just text modes!

johnnovak:
Nice quick fix @kcgen 🎉 , but we only need the last commit 😄 We've been back and forth on text mode special handling for a while and both of us came to the same conclusion that video output is video output — type of mode does not matter one bit. 

I'll test this later on my Windows box, but just the last commit 😄 I'm quite certain the rest are not necessary, but we can always discuss for sure.

src/gui/sdlmain.cpp
johnnovak: For the record, compositor-level vsync was introduced in Windows Vista, so in 2007. You could disable compositor level vsync in Win 8 (somehow), but it's that's no longer an option in Win 10+.

kcgen: That's why I'm defaulting to assuming vsync is enabled in windowed mode (only when `vsync = auto`) instead of possibly getting it wrong through analyzing benchmark results if the timing or frameates just happen to run a bit faster than normal. 
In other words, it's the most reasonable default in windowed mode (because compositors are already enforcing vsync). 

The consequence of getting it wrong (that is, the drivers and SDL telling you that vsync is off when it's actually enforced) is that 70 Hz frames rates will stutter and "breathe" in waves that go above and below the 60 Hz range.

kcgen: The presentation mode needs to know if we can present frames faster or slower than the DOS rate. 

On a 60hz display in windowed mode, the answer, 99% of the time, we'll be orcefully blocked if we try to present faster more 60 fps. 

And yet, querying both SDL and the OpenGL might tell you "vsync is off, the sky is the limit". But yet, our micro benchmark will actually prove that we're clamped at 60 fps (which is our last chance to detect I'd we're actually vsync'd).

src/gui/sdlmain.cpp
kcgen: We can leave vsync off in windowed mode, but I still need (or strongly prefer) being able to make the reasonable assumption that we're "effectively-vsyncd", or rate-bound no faster than the reported display's refresh rate when both in windowed and when the vsync setting is auto.

src/gui/sdlmain.cpp
kcgen: Ok, I'll check how that looks tomorrow. Maybe the other change is sufficient.

johnnovak: Well, just forget about differentiating text vs graphics modes altogether. You're thinking how the problem can manifest is 100% correct; the error is only when assuming this can only happen in text modes. No, it can happen in _any_ video mode!

src/gui/sdlmain.cpp
johnnovak: Ok, in that case the error is mine 😄 I thought this means you're trying to _enforce_ vsync through SDL, or by setting `glSwapInterval(1)` or something. If this is merely an assumption about what is _already happening_ in windowed mode, then yes, this is actually correct and I take it all back 😄 

Haven't read the whole code, maybe I should've...

src/gui/sdlmain.cpp
johnnovak: You're absolutely right! I take it back, as per my other comment. I thought you setting this flag would mean we in DOSBox are actively enforcing vsync in windowed mode by asking SDL, or asking the OpenGL driver directly.

If it's just passing this implicit info about windowed mode to your benchmarker, then yeah, we're in 100% agreement, and this is most definitely what we should do! Apologies 😄

johnnovak: Although... it's okay that we're making this assumption for the _benchmarker_, but this help setting talks about what we enforce on the driver level:

> Set the host video driver's vertical synchronization (vsync) mode:

So actually no, we're not setting vsync ourselves in windowed mode; we _assume_ vsync in on in windowed mode, right? Based on what you wrote above, I think that is what's happening, so the help text is actually incorrect now, no?

kcgen:
> I'll test this later on my Windows box, but just the last commit

Excellent. Just double check that you're running in **_throttled_** VFR mode to exercise that code path.

Then try feeding it changed frames with back-to-back timing tighter that 60 Hz to ensure that it's forced to drop frames (and the commits "follow up present logic" is ususe.

Some tricks to ensure you're in throttled mode is to set vsync true in your conf, or have your video drivers enforced vaync and prevent/override applications from changing it. 

Quake's time demo, especially using openglnb output, definitely runs > 70 fps (and it has a nice flashing graphical text cursor at the end when it reports the results).

src/gui/sdlmain.cpp
weirddan455: >  No, vsync in windowed mode just happens on all modern OSes.

I mean, that is just false on a Linux system.  I've written enough simple games in SDL to know you get thousands of FPS and a very warm computer if you don't enable VSync, even in Windowed mode.

Maybe the compositioner helps tearing but it does not cap the FPS in the way that real VSync does.  You're also not even guaranteed to have a compositioner running.  KDE, for example, has a feature that turns off its compositioner on demand when it detects an SDL application or game running in order to boost performance.

kcgen: > If it's just passing this implicit info about windowed mode to your benchmarker, 

Need to double-check 😅! I'm positive the info is passed through, however need to check if we're trying to enable it as well.

The dilemma, at least on Linux and the Pi, is that everything can lie. So basically it's all just best effort. 
I've had cases where SDL and OpenGL say vsync is enabled, get 0==success return codes from those C-calls.. but nope: the compositor or drive can override you. (And vice-versa). 
This is what drove the need for the microbenchmark. 

But yeah. . Will double check it tomorrow and worstcase, will just make auto the assumed state.

johnnovak:
> Quake's time demo, especially using openglnb output, definitely runs > 70 fps (and it has a nice flashing graphical text cursor at the end when it reports the results).

Yeah, that's a good way. Running my monitor in fixed 50 Hz mode and simply using the DOS prompt in VGA mode at 70 Hz and just executing DIR or something is also a good way. The issue reliably happens then. Or using QPV and just moving around on the screen.

src/gui/sdlmain.cpp
johnnovak: > > No, vsync in windowed mode just happens on all modern OSes.
> 
> I mean, that is just false on a Linux system. I've written enough simple games in SDL to know you get thousands of FPS and a very warm computer if you don't enable VSync, even in Windowed mode.
> Maybe the compositioner helps tearing but it does not cap the FPS in the way that real VSync does.

Actually, that's true. Maybe I should read the whole code again sometime... But ok, we kinda need to enable OpenGL level vsync level anyway to act as back-pressure to avoid generating too many frames. That's actually too. Mea culpa. I'm tired, didn't have much sleep... 😩 

> You're also not even guaranteed to have a compositioner running. KDE, for example, has a feature that turns off its compositioner on demand when it detects an SDL application or game running in order to boost performance.

Yeah, Linux might be special, but it's been enforced on macOS and Windows for the last 10+ years.

> EDIT: And sorry, I will not accept your Reddit thread with 3 upvotes as proof 😆

Haha, fair enough 😄 "Bbbb... but I found it on the internet!...." 😅

GranMinigun: 1) X compositors can be suspended or not be present at all (which will lead to tearing), and compositor-side vsync can be disabled (which, again, will lead to tearing). In either case, actual framerate is not limited, it's up to the program how many frames it renders (rendering is separate to presenting).
2) Wayland compositors recently (yes, it took years) got a protocol to enable tearing. But that only allows tearing presentation, it does not affect rendering speed in any way. Even without that protocol, the framerate is not limited.
3) Vulkan has mailbox presentation mode. Unlimited framerate, but vsynced presentation. Works the same on any platform that supports it, - that includes both Windows and macOS. While Staging currently does not implement Vulkan-based renderer, I brought it up for a better understanding of what compositors actually do.

Now, enabling vsync both on program and compositor side might actually lead to additional latency (remember, we're talking frames, not milliseconds) and even dropped frames. But OpenGL has absolutely no way of integrating with any platform's compositor, it isn't aware of such things. On Windows, we'd have to directly talk to either DWM or DXGI to get properly timed vsync, for example. (Same goes for Vulkan, really.) Only exclusive fullscreen allows the program to properly drive presentation, and that's discouraged on Windows and outright does not exist on Linux.

johnnovak: Thanks @GranMinigun. So what default do you suggest?

(Btw, I started thinking about how the hell does lagless vsync even work in WinUAE. Well I know it needs fullscreen mode, it cannot be used in windowed mode (the option is even greyed out). Then in fullscreen I'm not even using exclusive fullscreen with the DirectX 12 renderer, well because it does not exist on Windows 10. But it does brute-force buffer swapping _with intentional tearing_ to achieve 1/N frame latencies where N is the frame slice number. It works splendidly... I just don't understand how is that possible... Anyway, side remark.)

johnnovak: Another remark, on Windows 10 I can achieve perfect butter-smooth scrolling in games like Pinball Dreams only one way: to use my custom tweaked 70 Hz mode on my monitor, then CFR with vsync enabled. Vsync is crucial otherwise it tends to get out of sync a couple of times per second and jumps a bit. So it does make a difference, but only if I can hit the exact DOS rate (or very close).

GranMinigun: Hard to say. Windowed mode might behave differently in Wayland compositors that allow direct scan-out (so presentation might tear if vsync is disabled on program's side). In all other cases, it doesn't really matter, as presentation will be driven by the compositor (or lack of one) anyway. Probably leave it off, if we're limited by the emulated monitor's refresh rate anyway? We shouldn't care how often display memory contents change, only how often emulated monitor hits vblank.

WinUAE's sync method works because DX's presentation is handled by DXGI, so it's completely integrated with the compositor. That was the case ever since DX10, I believe. And it's possible to ask for the current scanline.

johnnovak: > We shouldn't care how often display memory contents change, only how often emulated monitor hits vblank.

That's what I suggested last year and that would solve this whole frame duping and missed frame business. But @kcgen explained VFR is necessary for some use cases still and it can conserve power for example when you're using a DOS IDE that rarely updates the screen (although, there is usually a blinking cursor anyway...) But pretty much all other emulators just do present on emulated vblank like you said and then you either enable vsync or you don't. Simple and works so that's still my preference, for the record 😄 Just present at 70 Hz in VGA modes, even if my monitor is in 50 Hz mode, then let the OS deal with it. Then for the vsync on case you really need a VRR monitor for DOS, or you can sort of of use fixed monitor refresh rates for specific modes like me, but that's a bit crap... (especially for demos that switch between custom refresh rates in tweaked modes every 10 seconds...)

kcgen: > Just present at 70 Hz in VGA modes, even if my monitor is in 50 Hz mode, then let the OS deal with it

If vsync is enforced either by the user at the driver level or frame rates are being limited by OS (or compositor), even if vsync is reported as off, our present-on-vblank calls will block for a lot more than 1ms.

This is actually what lead to @kklobe originally writing the frame-pacer. Apple M1-based laptops running external displasys hard-block our presentation calls for > 10ms, caused brutal stalls and lag.

The frame-pacer was then rejoiced for boosting performance, because the same stalling was being experienced under Linux on the Pi; and other users also reported performance boosts (as did I, only Linux + 60 Hz system).

The downside of the frame-pacer was that it dropped the /next/ frame, which might have been important (which lead to more changes, like the presentation-mode throttling-and re-present approach on the front-side), which the pacer itself is still  a valuable failsafe on the backside.

kcgen: The root cause is that the presentation and emulation are running serially. If we can split them into separate threads, then the suggestion:

> Just present at 70 Hz in VGA modes, even if my monitor is in 50 Hz mode, then let the OS deal with it

Would be perfect - because the presentation layer can jam up against against any vsync blocks, while the emalator carry on :-)

kcgen:
Round two ready:
 - Removed text-mode handling.
 - `vsync = auto` now only _assumes_ it's enabled (as discussed above), but we still try to disable vsync (per original behavior).

src/gui/sdlmain.cpp
kcgen: Removed!👍

src/gui/sdlmain.cpp
kcgen: Updated autoto the assume vsync, but not set it.

src/gui/sdlmain.cpp
kcgen: (updated in above comment)

src/hardware/vga.cpp
kcgen: Removed.

src/gui/sdlmain.cpp
johnnovak: > If vsync is enforced either by the user at the driver level or frame rates are being limited by OS (or compositor), even if vsync is reported as off, our present-on-vblank calls will block for a lot more than 1ms.

Ok @kcgen but that's how other emulators work. If I run VICE to emulate a PAL 50 Hz C64 or WinUAE to emulate a PAL Amiga on a fixed 60 Hz display with vsync on, things will go comically wrong. Audio stutters, general lagginess and jerkiness, etc.; it will become 100% unusable. Solution? Don't do that unless you can provide a 50 Hz host rate to vsync to 😄 

In those emulators vsync on is a power user thing for people who know what they're doing. Everybody else; please use the vsync off default 😅 And in windowed mode the vsync option is greyed out; they must just present the frame at timed intervals without any vsync enforcement on the _emulator_ side. There will be zero tearing that way due to the compositor level syncing, or Linux with compositor level vsync off there will be tearing, but that's how it is. So to be clear they _never_ block or time anything to host rates in the vsync on mode; they continuously present frames at 50 Hz in a non-blocking fashion then it's the driver's business.

You're saying that we got blocking with this approach when _not_ enforcing vsync in our program, so even if we do `glSwapInterval(0)` and present frames at precisely timed intervals only, running on our own timers? That's very surprising if that's the case, to put it mildly. There should be no blocking in our app in that scenario. Am I completely confused here @GranMinigun ?

Can't we just do the same?

GranMinigun: It will block for the amount of time required to render the image and send it to the presentation engine, but it won't wait for vblank. We don't do threaded/async rendering/presentation.

johnnovak:
Nice @kcgen , I'm gonna test this in 30 mins.

src/gui/sdlmain.cpp
kcgen: There's lots of systems that have 60 Hz vsync enforced at the driver or compositor-levels, and these user suffer badly when running dosbox playing fast 70 Hz games. Lots of slight micro jamups! I know.. I'm one of those people 😂

It also happens on kklobe's Apple laptop + external display.

I'm running a standard Linux + KDE, have it configured for the lowest latency, and OG DOSBox (as well as pure VFR mode) will give me only 45 fps in Quake 320x200 because of these jamups.

It happens to SBC users too running on the framebuffer, like in retropie. 

I believe the default Windows Intel settings are vsynced as well, as I had to dig around to figure out how to disable it. But some Intel driver vartianta are even worse and don't provide that option.

johnnovak: Thanks for the explanation @kcgen (again and again 😅) Don't you think it would be worthwhile to start a documentation page on these presentation mode details? I can create a blank page template for you, then you can put in some rough info in there and I can help with rewording and stripping it down to easily digestible information, etc. if you just wanna do it quickly. Frankly, I could do it myself and use your above comments as a starting point, just wanted to ask you first. If busy or don't have time, I can give it a shot and you'll review it eventually anyway.

It's just all this info is super useful and necessary to explain the various presentation mode options to our users and I'll forget 50%+ of it by next week 😅 JustWorks(tm) by default is the best, no doubt, but we still need to explain things in the manual, and power users need to know. The release notes on presentation modes don't mention any of this background context, so going by that very few people could correctly guess why this is needed, why is it useful, and so on (well, hence I was super confused by the beginning; the release notes was the only info I was going by).

Btw, yeah, when emulating pre SVGA cards, fixed host rate setting could work 100% well. But starting with VGA, 640x480 is already 60 Hz, then with SVGA and VESA anything goes because by that time multisync monitors became the standard. So for perfect SVGA/VESA emulation, especially to run all those demoscene prod without throttling, a full VRR setup is non-negotiable. Just making this observation, there's nothing we can do about this 😄 But I guess it underlines the importance of throttled VFR, if we have to do VRF because of the above reasons.

johnnovak:
First thing, this fixed the problem @kcgen 🎉 Great job! 😎 

---

Now on to the details:

### Reproduction

I found a good way to spot the "too low dupe rate" problem that this PR 100% fixes. Starting at the cursor blink rate is not really reliable.

1. Use lower than 70 Hz host rate. Fixed 60 Hz is fine, 50 Hz is even better. You **must** use lower than 70 Hz rate to test this!
2. Launch DOSBox with the defaults (`--noprimaryconf`). Confirm that "throttled VFR" is picked.
3. Start DOS Navigator.
4. Open some directory that has a bunch of files in it
5. Keep pressing the up arrow five times in a row, then the down arrow five times, and keep repeating this, so the cursor bar moves and up down.
6. Establish an _even rythm_ of keypresses. This is important!
7. The latency should be ideally roughtly _constant_, like on a real computer. But when the lower-than-host-rate frame rate dupe logic kicks in, the FPS effectively drops to 7-10 FPS or something, causing a _very_ noticeable delay. This always throws me off my rhythm!

So, with current main, you get the variable latency problem, which manifests as "I pressed a key, but the screen got updated MUCH later, like 50+ ms later". Super jarring. With the current PR, this problem is finally gone! 🎉 

Like it said, but to drive it home 100% this time 😅: _not only text modes are affected!_. But it's the _easiest_ to spot this problem in text modes. You can do the "rhythm test" by typing the letter "A" or something in the DOS prompt at an even rhythm. Every few seconds, your even rhythm will be "broken" by an unexpected latency spike. Try it!

## Results

Tried the above test, then just using thec command prompt randomly, DOS Navigator, QPV (with image preview turned off), a bunch of demos, switching between modes, windowed <-> fullscreen, etc. Throttled VFR was consistently picked in these scenarios and things just worked as they should.

Specifically:

- macOS, ProMotion built-in display (120 Hz capable). Tried running in 60 Hz, 50 Hz, 48 Hz, and 47.95 Hz modes, all work as expected.
- Windows 10, tested with fixed 50 Hz and 60 Hz. Same results, works splendidly, all weird dropped FPS problems and missing frame problems are gone.
- Win 10, custom 70.089 Hz mode. Now this is interesting, the default settings still pick throttled VFR. The scrolling in Pinball Dreams is jerky (both at the credits screen and in-game). If I set `vsync = on`, there is exactly _zero_ difference or improvement. The _only way_ to get 100% smooth scrolling in VGA games that sync to 70 Hz is to use `presentation_mode = cfr`. The `vsync` setting makes zero difference. Actually, `vsync` made zero difference in all my testing.
- For the record, on my 120 Hz (fixed) ProMotion laptop screen on macOS, the best result for Pinball Dreams is to leave everything on auto. That picks VFR, and the scrolling is not smooth, but it's not horribly jerky either. It's like something in-between, but that's the best we can do because 120 Hz is not an even multiple of 70 Hz. 
- `vsync = on` makes some difference on macOS, but only in 50/60Hz modes, and not for the better. In 120Hz mode it makes no discernible difference.

So, to sum up:

- The defaults are good on my systems
- Throttled VFR now works as expected, no weird uneven frame pacing now and then when there are no constant screen updates
- Throttled VFR just can't do proper smooth 70 Hz syncing, apparently not even when I'm in my custom 70.089 Hz mode. Only `presentation_mode = cfr` can. Not _really_ a major issue, but wanted to call it out.
- `vsync` doesn't make anything better on my systems; it's a completely useless setting (on _my_ systems).

kcgen:
Phew, thanks for raising this issue @johnnovak and running so many extensive tests. I feel like the approach is pretty robust heading into 0.81.  
One day we'll all be running displays fast enough for DOS (and this code can all be tossed in favour of CFR all the time :-)

src/gui/sdlmain.cpp
kcgen: Yes, I can try to summarize it for the manual and power users; happy to fill in those details 👍

johnnovak:
> Phew, thanks for raising this issue @johnnovak and running so many extensive tests. I feel like the approach is pretty robust heading into 0.81. One day we'll all be running displays fast enough for DOS (and this code can all be tossed in favour of CFR all the time :-)

image

weirddan455:
>  One day we'll all be running displays fast enough for DOS 

Didn't know I needed some high tech display to run software from the 80s :laughing:

johnnovak:
> > One day we'll all be running displays fast enough for DOS
> 
> Didn't know I needed some high tech display to run software from the 80s 😆

The first affordable 8k wide-gamut HDR 500 Hz+ VRR OLED will be a good candidate to replace my CRTs. Before that happens, it's a matter of different levels of compromises 😎

Issue #2757, opened 2023-08-14
Integrate the header files from the `include` directory into `src`


johnnovak:
Having separate `include` and `src` directories makes a lot of sense for libraries where the public header files from `include` and the object files are your published artifacts.

It makes very little sense for _applications_, though, where nothing gets published and your final deliverable is a binary (plus resource files and 3rd party DLLs, of course, but no header files).

See question and top comments here:
https://stackoverflow.com/questions/21881410/best-practice-src-include-or-include-folder-next-to-src-folder

The separate `include` dir of DOSBox serves no useful purpose; it's pretty arbitrary whether something goes into `src` or `include`. It's all over the place currently, and yeah, header-only stuff mostly lives in include, but that's no good reason for its existence.

As this separation only complicates and confuses things but solves nothing, it has to go. So I'm proposing to integrate all header files from `include` into `src`, then get rid of the `include` dir altogether.

Note I said _integrate_, not _dump_ 😄 Simply moving all header files into `src` and calling it a day won't cut it; every individual file must be moved to their most appropriate place.

FeralChild64:
I have some doubts. Our _application_ in reality consists of several different _modules_, and it makes sense to have a separate directory for inter-module communication headers.
If you look into `src/hardware/input` - there are quite a few headers there, but they are used for communication inside mouse/keyboard emulation only. `include/keyboard.h` or `include/mouse.h`  are to be used by external modules. Sure, this separation is not 100% clean and it could be improved - but there is clearly an idea behind it.

johnnovak:
Yeah, I'm agreeing with you on the importance of making a distinction between "module-internal public" and proper "module public" exports. But that further solidifies

johnnovak:
(The cat closed the issue, not me 😹 ⌨️ )

johnnovak:
Addressed in https://github.com/dosbox-staging/dosbox-staging/pull/4451

Issue #2051, opened 2022-10-30
Externalise hard cga_colors palettes


johnnovak:
The currently hardcoded `cga_colors` palettes should be externalised into resources, except for the two parametric palettes, `tandy` and `ibm5153` — these need to remain hardcoded.

See these comments for further details:

https://github.com/dosbox-staging/dosbox-staging/issues/2014#issuecomment-1280251029
https://github.com/dosbox-staging/dosbox-staging/issues/2014#issuecomment-1288014867

MashrafeeAryan:
Hi @johnnovak! I am Mashrafee. I am an undergrad student and I wanted to learn more about open source contribution. I believe I can work on this issue. Is this issue still open?

johnnovak:
> Hi [@johnnovak](https://github.com/johnnovak)! I am Mashrafee. I am an undergrad student and I wanted to learn more about open source contribution. I believe I can work on this issue. Is this issue still open?

Sure. Read our contributor's guide first:
https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/CONTRIBUTING.md

MashrafeeAryan:
> > Hi [@johnnovak](https://github.com/johnnovak)! I am Mashrafee. I am an undergrad student and I wanted to learn more about open source contribution. I believe I can work on this issue. Is this issue still open?
> 
> Sure. Read our contributor's guide first: https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/CONTRIBUTING.md

Thank you so much! I read through the guide and will start working on the issue.

MashrafeeAryan:
Hi @johnnovak, 
I opened a pull request for this issue #2051. I was struggling to think of a proper way to test it and tried testing it to the best of my ability.  I compiled it in Windows with debug-windows preset and launched the emulator with various cga-colors to verify if the program crashes. If you have suggestions on how test it in other ways or if anyone would help me test it on macOS  or Linux, I would really appreciate it!

Thank you so much for your time!

PR #2819, opened 2023-09-02
`crt-auto` should auto-switch to VGA shaders in EGA modes with 18-bit VGA palettes

johnnovak:
_Note: I will use parts of this writing in my upcoming mega-article about PC monitors. Hence I put quite some effort into it, but you can safely skip the details in the second post and only focus on this first post._

---

# Overview

Demos and many DOS ports of Amiga action/platformer games "repurpose" the 320x200 0Dh EGA mode: instead of the default 16-colour palette, they reprogram the palette to use 18-bit VGA colours. The planar 320x200 16-colour 0Dh EGA mode is preferred by these programs because certain effects/operations are more efficient in planar modes than in the "chunky" (1 byte per pixel) 320x200 256-colour 13h mode (e.g., smooth-scrolling on low-end hardware, such as 286 and slow 386 machines equipped with VGA adapters; a popular low-cost option in the early 90s in Europe where most of these Amiga ports originated from).

These programs need a VGA card; in most cases, they don't even work on an EGA card, or they do but display wrong colours. Therefore, the authentic choice in these cases for the `crt-auto` adaptive CRT shader is to use a VGA shader.

## 200-line EGA modes

The most important difference between using EGA vs VGA shaders for these 200-line games is that the VGA shaders are double-scanned, resulting in a very different-looking picture (big chunky pixels, as is expected from low-res VGA games). The single-scanned EGA look is like a _really_ sharp Commodore monitor (which never existed 😅) and thus it's completely inauthentic. Can be a cool effect, though, but that's what the "fantasy" `crt-arcade` shader is for.

In 99% of cases, such programs that repurpose EGA modes with VGA colours use the 320x200 0Dh low-res EGA mode. I know of a _single_ obscure game called Rusty that does it with the 640x200 0Eh EGA mode, and only in-game. This game is handled correctly by the detection logic and gets double-scanned with a VGA shader too.

## 350-line EGA modes

I'm not bothering with 640x350 10h EGA modes as those are single-scanned both on EGA and VGA, so there's minimal difference between the EGA and VGA CRT shaded output. For 640x350 EGA modes we continue to always pick EGA shaders.

I know only of _two_ games that use the 640x350 mode with VGA colours: Darkseed (the whole game) and Daughter of Serpents (only the character creation screen). Of these two, Darkseed is the really important one, so we're really talking about a single game, and it looks completely fine with EGA shaders.

The main reason is I couldn't work out the VGA palette detection for 350-line modes, but as it turns out, luckily this affects only Darkseed and would make very little difference.

## `cga_colors`

Naturally, the palette detection is compatible with any custom CGA palette configured via `cga_colors` (built-in or user-provided). JustWorks(tm), just like you'd expect! 😎 

# Example

## Gods – double-scanned VGA shader, correct

gods-vga

## Gods – single-scanned EGA shader, incorrect

gods-ega


johnnovak:
# Test cases

Testing graphics is always fun, especially edge-case-y stuff like this for a person with mild OCD tendencies. Like me 😎 (I'd say I'm a solid 5 on a 1 to 10 OCD scale 😅)

Note that some games _look_ like EGA games, but many of them are these weird low-effort "VGA conversions". They slapped on a 256-colour start screen maybe for the VGA release, then reused the exact same 16-colour graphics in-game in the 320x200 VGA 13h mode so they can proudly put the "now with VGA support!" badge on the box... 🤦🏻 Windwalker is a prime example; the VGA intro doesn't look any better than the EGA version (I'd say it's worse with those ghastly gradients), and the in-game graphics is 100% the same EGA art. Yeah, there are many games like that.

Naturally, we cannot and _should_ not treat those as EGA games, but it's quite confusing. Some other games tweak the default EGA palette in quite insignificant way when you run them on a VGA card, e.g., just change the purple colour to a skin tone (Might and Magic 2 does that, and only in-game; people in the start screen and hand in the intro are still purple... talk about low-effort!). These "technically VGA games" pretty much still look like EGA games to the untrained eye, and in fact, the graphics are 90%+ still _effectively_ EGA. Some of them do some really minor, almost unnoticeable palette tweaking when run on VGA (It Came From the Desert is like that...) 

You won't believe how many games do this low-effort "VGA conversion" thing, and technically that makes them "VGA games" from the point of view of the algo. And people who bought them back in the day, thinking they'll get some nice VGA game, well, at least they got double-scanning on VGA, so we're emulating their experience accurately—including the confusion and the disappointment 😎  Anyway, these are the card we're dealt with :shrug:


## 320x200 0Dh EGA mode with default colours

These games use the default 16-colour CGA palette, so single-scanned EGA shaders are selected for them. This is basically regression testing to make sure true EGA games are not misdetected as VGA games.

100% success rate ✅ 

- 2400 A.D.
- Arachnophobia
- Bard's Tale 1, The - Tales Of The Unknown
- Bard's Tale 2, The - The Destiny Knight
- Bard's Tale 3, The - Thief Of Fate
- Bard's Tale I 
- California Games
- Castle Master
- Catacomb 3D
- Champions of Krynn
- Chip's Challenge
- Commander Keen - Keen Dreams
- Commander Keen 1 - Marooned on Mars
- Commander Keen 2 - The Earth Explodes
- Commander Keen 3 - Keen Must Die!
- Commander Keen 4 - Secret of the Oracle
- Cosmo's Cosmic Adventure - Forbidden Planet
- Death Knights of Krynn
- Defender of the Crown
- Dragon Wars
- Drakkhen
- Duke Nukem - Episode 1 - Shrapnel City
- Eye of the Beholder I
- Fountain of Dreams
- It Came From The Desert 
- King's Bounty
- Leisure Suit Larry 2
- Out of This World (EGA mode)
- Pool of Radiance
- Quest for Glory I
- Quest for Glory II
- Rings of Medusa
- Space Quest 3
- Street Rod
- Total Eclipse
- Ultima IV - Quest of the Avatar
- Ultima V - Warriors of Destiny
- Wasteland
- Wizardry VI - Bane of the Cosmic Forge
- Zak McKracken and the Alien Mindbenders (Enhanced)



## 320x200 0Dh EGA mode with VGA colours

These games set up 16 custom 18-bit VGA colours.

100% success rate ✅ 

- Blues Brothers
- Cadaver
- Cadaver - The Payoff
- Duke Nukem II
- Gods
- Heimdall
- Immortal, The
- Knightmare
- Magic Pockets
- Metal Mutant
- Out of this World (Another World)
- Prehistorik 2
- Rastan
- Might and Magic II (on VGA it redefines the skin tone colour, in-game only)
- Rick Dangerous 2
- Speedball 2
- Starblade
- Xenon 2 - Megablast
- Zool

## 640x200 0Eh EGA mode with VGA colours

These games set up 16 custom 18-bit VGA colours. They require VGA cards, therefore double-scanned VGA shaders are in order.

100% success rate ✅  (1 out of 1! 😎 )

- Rusty (in-game only)

## 640x350 10h EGA mode with default colours

These games pick 16 colours from the standard 64-colour EGA palette, hence EGA shaders (single-scanned, but it makes no difference as 350-line modes are always single-scanned on VGA anyway).

100% success rate ✅ 

- Chuck Yeager's Advanced Flight Trainer 2.0
- Corruption
- EGA Trek
- Fish!
- Gateway
- Jet
- Jinxter
- Pawn, The
- Sierra's 3-D Helicopter Simulator
- SimAnt
- SimCity
- SimFarm
- Spellcasting 101, 201, 301
- Sub Battle Simulator The Masters Collection
- Timequest
- UFO

## 640x350 EGA mode with custom VGA palette

These games set up 16 custom 18-bit VGA colours. We just use EGA for these as the differences are small compared to the VGA shader (because of single scanning) and I can't be bothered to figure out the detection for these two (one and a half...) games.

100% success rate ✅  (all 2!)

- Darkseed
- Daughter of Serpents (character creation screen only)


# Screenshots

## Arachnophobia (320x200 EGA)

image

## Champions of Krynn (320x200 EGA)

image

## Dragon Wars (320x200 EGA)

image



## Gods (320x200 EGA with VGA colours)

### Double-scanned VGA shader (new behaviour)

image

### Single-scanned EGA shader (old behaviour)

image


## Rusty (640x200 EGA with VGA colours)

### Double-scanned VGA shader (new behaviour)

image

### Single-scanned EGA shader (old behaviour)

image

johnnovak:
# Confusing "enhanced" VGA versions

## Might and Magic 2 - Gates to Another World

On VGA adapters, you can run the game in VGA mode with `loadfix mm2 M` to get EGA graphics with a _minimally_ enhanced palette but in the 320x200 256-colour 13h VGA mode, or with `loadfix mm2 E` to run it in pure EGA mode.

### Character creation – EGA

Zero difference between the two modes in terms of colours; skin tones are always purple.

mm2-char-ega

### Character creation – VGA

mm2-char-vga

### In-game – EGA

The purple colour used for skin tones in the EGA version has a proper skin tone colour in VGA mode. That's the only difference! I hope the team did not have to work overtime to create the "enhanced VGA version" 😛 

mm2-ingame-ega

### In-game – VGA

mm2-ingame-vga


# Windwalker

## EGA

This is how the game looks with `machine = ega`:

windwalker-ega

## VGA

On a VGA adapter, the intro screens are a bit different because they use a 256-colour palette, but the actual game uses the same EGA graphics.

But there's lots of weirdness here:

- Running the game with `WIND.EXE` or `WIND.EXE EGA` results in identical looking results, but the former uses mode 13h (VGA), and the latter 0dh (EGA).
- Both seem to use the standard EGA palette, but there must be some slight change in EGA mode on VGA as it triggers double-scanning.
- Note that the brown colour is mapped to pink in the `machine = ega` output. But that seems to be the standard EGA brown! So something super weird is going on here... You _don't_ get the canonical palette with `machine = ega` mode because the brown is mapped to pink, but you get it on VGA... when you run it in EGA more... but it's slightly different than the EGA palette still. Wat???!?

windwalker-vga

johnnovak:
PVS Studio wants me to pass `unions` that wrap `uint8_t`s by reference 🥶 I refuse to do that...

image

FeralChild64:
> PVS Studio wants me to pass `unions` that wrap `uint8_t`s by reference 🥶 I refuse to do that, that's crazy...
> 

Always happens with the bit_view style unions. I’m afraid at some point we will get overhelmed with such warnings…

johnnovak:
> > PVS Studio wants me to pass `unions` that wrap `uint8_t`s by reference 🥶 I refuse to do that, that's crazy...
> 
> Always happens with the bit_view style unions. I’m afraid at some point we will get overhelmed with such warnings…

Yeah I'd turn this warning off them. Bitviews are super useful, and I can see them being used more and more going forward, and yes, also as function arguments. No way I'm gonna pass a 64-bit pointer (reference) instead of a byte...

kcgen:
Yup, no harm in disabling this specific V-number.

@johnnovak , in the pvs-studio CI yaml, add V801 to the list:

```
disable_warnings="V002,V1042,V826,V802,V2008,V1071"
```

Should bring down the count by a significant number.

My guess is it's seeing the `enum class` objects as real classes at the object/symbol level (and with must stronger type guarantees), and so probably has a hard time differentiating. 

Where as the C-enums are just integers so those appear as scalars.

Our coding style passes real class objects as const-ref: `Function(const ClassType& class_instance)`, so just have to catch pass-by-value at review time (although I think we're all pretty good about that).

include/vga.h
```diff
@@ -119,6 +120,9 @@ enum VGAModes {
 	M_ERROR = 1 << 31,
 };
 
+constexpr auto NumCgaColors = 16;
+constexpr auto NumVgaColors = 256;
```
weirddan455, line 124: Are we doing British or American spelling of color?  You've got a mix of both in this PR.  I'm not going to comment on each one but do a find on "color" and "colour" and be consistent one way or the other.

include/vga.h
```diff
@@ -755,6 +755,84 @@ union CrtcModeControlRegister {
...
+	//     1        1        100%
+	bit_view<0, 1> b;
+	bit_view<1, 1> g;
+	bit_view<2, 1> r;
+	bit_view<3, 1> sb;
+	bit_view<4, 1> sg;
+	bit_view<5, 1> sr;
+
+	// On VGA, this is treated as 6-bit index that addresses the first 64
+	// Color Registers that store the actual 18-bit RGB colours.
```
weirddan455, line 832: Ok, I guess most of the actual code is using the "Color" spelling but this comment where you used both spellings in the same sentence is what caught my eye.

weirddan455:
Looks fine as far as I can tell but those vga* files look like black magic to me :laughing: 

The bulk of the changes look like refacors which seem to seem help readability.  Other than that, nice job improving shader auto-detection for people like me who wouldn't know the right one to choose :+1:

johnnovak:
> Looks fine as far as I can tell but those vga* files look like black magic to me 😆

To all of us 😄 I only ventured into those areas after @kcgen bravely paved the way with his VGA double-scanning improvements earlier this year 😅 But yeah, you could easily spend years understanding and cleaning up that code...

> The bulk of the changes look like refacors which seem to seem help readability. 

Yeah, and most of those refactors are not strictly needed for this PR, but doing them armed with a good book on VGA helped me understand what the code is doing. They named arguments and variables very sloppily, making it impossible to understand the palette vs color register distinction and mapping, unless you're super familiar with VGA at the lowest levels... Well, I was somewhat familiar in the 90s, but this confused the hell out of me.

I'm chipping away at the VGA code with such small improvements whenever I'm touching it, which will improve the situation over time.

> Other than that, nice job improving shader auto-detection for people like me who wouldn't know the right one to choose 👍

It does get confusing, doesn't it 😄 Some of those low-effort "VGA conversions" confused me too; first I thought my detection method was buggy, but no 😄

kcgen (commented):
More great changes, @johnnovak.
Just a couple minor comments.

include/rgb666.h
```diff
@@ -0,0 +1,51 @@
...
+	uint8_t green = 0;
+	uint8_t blue  = 0;
+
+	constexpr Rgb666() = default;
+
+	constexpr Rgb666(const uint8_t r, const uint8_t g, const uint8_t b)
+	        : red(r),
+	          green(g),
+	          blue(b)
+	{}
```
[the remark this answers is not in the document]
johnnovak: Agree on both points 😄

src/gui/render.cpp
```diff
@@ -626,6 +632,21 @@ bool RENDER_MaybeAutoSwitchShader([[maybe_unused]] const uint16_t canvas_width,
 #endif // C_OPENGL
 }
 
+void RENDER_NotifyEgaModeWithVgaPalette()
+{
+	auto video_mode = VGA_GetCurrentVideoMode();
```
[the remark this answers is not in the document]
johnnovak: Yeah, good idea. It's kinda protected already, but on my first attempt it wasn't so i was getting random crashes in Hercules mode because it started doing _something_ in that mode too 😅

include/vga.h
johnnovak: This is pretty annoying... We use British spelling on our website and in our manual. But in the code we use American spelling, so "color" and letter "Z" everywhere because it's too much of a hassle to change it, and for not much benefit. In code the American spelling is "standard" anyway, kind of; all graphics APIs use "color" AFAIK.

So my approach is to suck up to Americans and use American spelling in the code 😄 But in all text meant for human consumption, so READMEs, commit logs, documentation, even log messages to some extent, I use International English spelling.

Important to note that English is the official language in many countries, and they in all these countries they use British spelling, basically. American spelling is for Americans only, so we're using Internal English spelling, which is effectively British spelling, yes 😄

include/vga.h
johnnovak: Take it up with Noah Webster 😆 (I would have a few word for him myself...)

johnnovak:
Addressed all comments & pushed some further fixups @kcgen, so you might wanna have another look at the commits from "Disable PVS studio warning V801" onwards.

tests/rgb_tests.cpp
```diff
@@ -34,6 +35,12 @@ constexpr uint8_t rgb6_to_8_reference(const uint8_t c)
 
 constexpr auto AbsError = 0.000001f;
 
+TEST(rgb, Rgb888_FromRgb444)
+{
+	EXPECT_EQ(Rgb888().FromRgb444(0x0, 0x1, 0x2), Rgb888(0x00, 0x11, 0x22));
+	EXPECT_EQ(Rgb888().FromRgb444(0x8, 0xe, 0xf), Rgb888(0x88, 0xee, 0xff));
```
kcgen, line 41: nice tests :-)

include/rgb888.h
```diff
@@ -42,6 +43,26 @@ class Rgb888 {
...
+
+	constexpr bool operator!=(const Rgb888& that) const
+	{
+		return !operator==(that);
+	}
+
+	// Scoped conversion helpers
+	static constexpr Rgb888 FromRgb444(const uint8_t r4, const uint8_t g4, const uint8_t b4)
+	{
+		const auto r8 = r4 | (r4 << 4);
```
kcgen, line 60: Suggest asserting the incoming 4-bit values are `< 16`.

include/rgb888.h
```diff
@@ -32,6 +32,7 @@ class Rgb888 {
 #pragma pack(pop)
 
 	constexpr Rgb888() = default;
+
```
kcgen, line 35: Can we get rid of the #pragma stuff in this header?

tests/rgb_tests.cpp
kcgen, line 41: Static class functions can be called more like a scoped function or an enum class function (if those existed):

```Rgb888::FromRgb444(args)```

File: include/rgb666.h [hunk 8]
```diff
@@ -53,6 +55,16 @@ class Rgb666 {
...
 	}
+
+	// Scoped conversion helpers
+	static constexpr Rgb666 FromRgb888(const Rgb888 rgb888)
+	{
+		const auto r6 = rgb888.red >> 2;
+		const auto g6 = rgb888.green >> 2;
+		const auto b6 = rgb888.blue >> 2;
+
+		return Rgb666(r6, g6, b6);
```
kcgen, line 66: because of the bit manipulation, the `const auto` are /probably/ `int` types, so then they're being implicity cast back into the `uint8_t` in the Rgb666 constructor.

Suggest static_cast<uint8_t>(..) the RHS.

include/rgb565.h
```diff
@@ -28,34 +28,26 @@
...
 	constexpr Rgb565(const uint16_t _pixel)
 	{
 		pixel = _pixel;
 	}
 
-	// Construct from separate RGB 8-bit values
 	constexpr Rgb565(const uint8_t r8, const uint8_t g8, const uint8_t b8)
 	{
-		FromRgb888(r8, g8, b8);
+		pixel = Rgb565::PixelFromRgb888(r8, g8, b8);
```
kcgen, line 40: Maybe a constructor that takes in an real `Rgb888` object?

src/ints/int10_modes.cpp
```diff
@@ -2162,23 +2160,23 @@ std::optional<Rgb666> parse_color_token(const std::string& token,
...
 
-		const auto r = parse_component(r_string);
-		const auto g = parse_component(g_string);
-		const auto b = parse_component(b_string);
+		const auto r8 = parse_component(r_string);
+		const auto g8 = parse_component(g_string);
+		const auto b8 = parse_component(b_string);
 
-		if (r < 0 || g < 0 || b < 0) {
+		if (r8 < 0 || g8 < 0 || b8 < 0) {
```
kcgen, line 2174: I guess this is user-data coming (custom CGA palette stuff), so we're not certain of these integer values. Suggest adding upper-bound checks to make sure `values are < 256` so they're not wrapped in the convert to Rgb888.

include/rgb888.h
johnnovak: I'd rather not. It's a pack pragma, so it's risky, and I don't see what we would gain by removing it (when you need packing, you need packing).

src/ints/int10_modes.cpp
johnnovak: It' handled in the parser lambda a few lines above. This is the error path because the one or all the components are set to negative to signal an error. But I'll change it to optional 😄

src/ints/int10_modes.cpp
```diff
@@ -2119,21 +2120,18 @@ std::optional<Rgb666> parse_color_token(const std::string& token,
...
 
-			return Rgb666(r, g, b);
-		} else {
-			auto r = static_cast<uint8_t>(value >> 16 & 0xff);
-			auto g = static_cast<uint8_t>(value >>  8 & 0xff);
-			auto b = static_cast<uint8_t>(value       & 0xff);
+		} else { // hex6 token
+			const auto r8 = static_cast<uint8_t>(value >> 16 & 0xff);
+			const auto g8 = static_cast<uint8_t>(value >>  8 & 0xff);
+			const auto b8 = static_cast<uint8_t>(value       & 0xff);
```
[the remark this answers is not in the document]
johnnovak: Oops, resolved by mistake 😄 Yeah some of that colour parsing stuff does not belong here and should be extracted. The parsing of `cga_colors` will eventually go away; there's a ticket to turn those into resource files. So I'm leaving the parser bits that will go away anyway here for now.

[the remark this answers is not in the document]
johnnovak: Actually, it's too tied to this whole `cga_colors` thing so I'll leave it be for now. It will be addressed in https://github.com/dosbox-staging/dosbox-staging/issues/2051 (made extra comment there about potentially extracting the hex colour parsing).

kcgen:
I was compiling with increased warnings enabled, I noticed that  
When I compile very non-video-related things, like `src/midi/midi_mt32.cpp`, I was getting RGB code-conversion warnings:

```
../../include/rgb888.h:71:31: warning: conversion from 'int' to 'uint8_t' {aka 'unsigned char'} may change value [-Wconversion]
   71 |                 return Rgb888(r8, g8, b8);
      |                               ^~
../../include/rgb888.h:71:35: warning: conversion from 'int' to 'uint8_t' {aka 'unsigned char'} may change value [-Wconversion]
   71 |                 return Rgb888(r8, g8, b8);
      |                                   ^~
../../include/rgb888.h:71:39: warning: conversion from 'int' to 'uint8_t' {aka 'unsigned char'} may change value [-Wconversion]
   71 |                 return Rgb888(r8, g8, b8);
      |                                       ^~
../../src/midi/midi_mt32.cpp: In static member function 'static void get_report_handler_interface()::ReportHandler::printDebug(void*, const char*, __va_list_tag*)':
../../:390:37: warning: format not a string literal, argument types not checked [-Wformat-nonliteral]
  390 |                         safe_sprintf(msg, fmt, list);

```

This is because the `int10.h` has the rgb666.h and rgb888.h included, which in turn is included because 

```
struct palette_t {
	// 64 entries
	std::vector<Rgb666> mono_text = {};

	// 64 entries
	std::vector<Rgb666> mono_text_s3 = {};

	// 16 entries. This is default canonical 16-colour CGA palette as emulated
	// by VGA cards.
	std::vector<Rgb666> cga16 = {};

	// 64 entries. This is the default 64-colour 6-bit RGB EGA palette as
	// emulated by VGA cards. The BIOS sets up these colours in the first 64
	// of the 256 VGA colour registers in EGA modes.
	std::vector<Rgb666> cga64 = {};

	// 64 entries
	std::vector<Rgb666> ega = {};

	// 256 entries. This is the default 256-colour VGA palette.
	std::vector<Rgb666> vga = {};
};
```

Which in turn adds a global `palette` variable inside the scope of (most) of our cpp, include the MIDI code.

```
extern palette_t palette;
```

The RGB headers include a fair bit of inline code itself - as opposed to just being structs.

I would be nice to have the `palette` object have reduced scope - more at the video-only level (vga.h?), at least not in such a top-level interrupt file like `int10.h`.

johnnovak:
> I was compiling with increased warnings enabled, I noticed that When I compile very non-video-related things, like `src/midi/midi_mt32.cpp`, and the `src/shell/.cpp` (and many others), I was getting RGB conversion warnings repeating over and over, from every .cpp compiled:

Isn't that true for all the inlined stuff? It's the nature of C++. The solution is to get rid of those warnings, which I can do.

> ```
> ../../include/rgb888.h:71:31: warning: conversion from 'int' to 'uint8_t' {aka 'unsigned char'} may change value [-Wconversion]
>    71 |                 return Rgb888(r8, g8, b8);
>       |                               ^~
> ../../include/rgb888.h:71:35: warning: conversion from 'int' to 'uint8_t' {aka 'unsigned char'} may change value [-Wconversion]
>    71 |                 return Rgb888(r8, g8, b8);
>       |                                   ^~
> ../../include/rgb888.h:71:39: warning: conversion from 'int' to 'uint8_t' {aka 'unsigned char'} may change value [-Wconversion]
>    71 |                 return Rgb888(r8, g8, b8);
>       |                                       ^~
> ../../src/midi/midi_mt32.cpp: In static member function 'static void get_report_handler_interface()::ReportHandler::printDebug(void*, const char*, __va_list_tag*)':
> ../../:390:37: warning: format not a string literal, argument types not checked [-Wformat-nonliteral]
>   390 |                         safe_sprintf(msg, fmt, list);
> ```

Yes, so that's easy to get rid of. Unfortunately you cannot put the `CHECK_NARROWING` thing into header files, so all header only, inlined stuff is prone to this.

> This is because `rgb666.h` and `rgb888.h` are included in `int10.h`:
> Which in turn adds a global `palette`:
> In (most) of our cpps, include the MIDI and shell code.
> The RGB headers include a fair bit of inline code itself - as opposed to just being structs.


> It would be nice to have the `palette` and rgb headers have reduced scope - more at the video-only level (vga.h?), at least not in such a top-level interrupt file like `int10.h`.

I agree, in general, that would be nice, but sometimes it's just unavoidable with our current tangled structure. Check out how much of the VGA codes messes around with the palette globals directly. 

> I'm also not sure if the palette can be access at the module level.

As above, doable, but time consuming, and I'd like to merge this and move on 😄 

> Another suggestion, with the palettes being fixed-size, is to use `std::array` instead of `std::vector`.

Can't do that because they vary in size, and many functions that work with them are generic and can accept any of them. I know these exact reasons because I tried a few days ago 😄 

Your suggestion of making a distinction between module private and module public stuff is good. That would reduce the scope of such inlined functions being included in half the codebase. However, that's how the crappy C include system works by munging files together, plus it's not an actual problem if those inlined functions don't generate warnings. So from a practical POV, this is ok.

But yes, going forward it would be nice to move towards the so-called canonical project structure that has clear delineation between module private and module public stuff. Please see this comment of mine which strangely did not receive much feedback, but I think it's a great idea. The strange standalone `include` dir would also go. See here: https://github.com/dosbox-staging/dosbox-staging/issues/2757#issuecomment-1679704560

But all these things are bigger refactoring jobs and I have a strong preference to do them completely separately. Then I have the mindset that ok, now I will spend a week or two on turning int10h into a well-separated module that does not expose its private parts 😏 I don't want to add one or two weeks of work to this PR 😅 I'm constantly chipping away at cleanup stuff in all my PR, much more than strictly necessary, but at some point I just wanna move on. This is that point, otherwise every 2 hour job can turn into a 2 month refactoring marathon 😄 The codebase we inherited makes that very much possible no matter what you're changing 😄 

I'll see if there are some quick wins to be had about the issues you raised. But I'll timebox it to 30 mins. Worst case scenario, I'll fix up the warning in the rgb includes and nothing else.

kcgen:
I agree @johnnovak but figured I'd ask. I think the only way to do it would be to use forward declarations, and I guess returning `palette_t*`. 

Both aspects (excessive globals that wind up in very top-level .h's like this int10.h) versus the struggle with trying to be more modular or fine-grained are both different fights with the language itself.

If some progress can be made -great, but yeah, I know the pain and it's not worth too much effort. Will be happy with what you decide!

johnnovak:
Related read @kcgen, written by a Clang / LLVM contributor 😄 

**C Is Not a Low-level Language**
Your computer is not a fast PDP-11.

https://queue.acm.org/detail.cfm?id=3212479

johnnovak:
Ok @kcgen a forward declaration in `int10.h` has obviated the need for including `rgb666.h`, but I've added some industrial-strength `check_cast`ing to `rgb666.h` anyway 😎 That will do for now.

kcgen:
> Related read @kcgen, written by a Clang / LLVM contributor 😄
> 
> **C Is Not a Low-level Language** Your computer is not a fast PDP-11.
> 
> https://queue.acm.org/detail.cfm?id=3212479

(not sure I need to read it.. C was meant to be a universal high-level language that finally let people stop writing actual machine code.. it was truly the "java" of the early mainframe era).  I worked with guys who wrote machine code on mainframes.. and trust me, they told me I basically was spoiled for have a C compiler that wrote the machine code for me.. write once, compile many :-)

johnnovak:
> > Related read @kcgen, written by a Clang / LLVM contributor 😄
> > **C Is Not a Low-level Language** Your computer is not a fast PDP-11.
> > https://queue.acm.org/detail.cfm?id=3212479
> 
> (not sure I need to read it 😅 I worked with guys who wrote machine code on mainframes.. and trust me, they told me I basically was spoiled for having a C compiler that wrote the machine code for me: write once, compile many! C and then C++ was a universal high-level language that finally let people stop writing actual machine code.. compilers were truly a marvel!)

Up to you, but it's not about that 😄 The TL;DR is that back in the day C was arguably a cross-platform assembler, but these days it's so far removed from the actual hardware with optimising compilers that it's basically a quite bad abstraction to the underlying machine. Kind of "worst of both worlds"; neither high-level enough, neither low-level enough, just archaic. Same goes for C++.

johnnovak:
Can I get the tick please @kcgen ?

kcgen:
> > > Related read @kcgen, written by a Clang / LLVM contributor 😄
> > > **C Is Not a Low-level Language** Your computer is not a fast PDP-11.
> > > https://queue.acm.org/detail.cfm?id=3212479
> > 
> > 
> > (not sure I need to read it 😅 I worked with guys who wrote machine code on mainframes.. and trust me, they told me I basically was spoiled for having a C compiler that wrote the machine code for me: write once, compile many! C and then C++ was a universal high-level language that finally let people stop writing actual machine code.. compilers were truly a marvel!)
> 
> Up to you, but it's not about that 😄 The TL;DR is that back in the day C was arguably a cross-platform assembler, but these days it's so far removed from the actual hardware with optimising compilers that it's basically a quite bad abstraction to the underlying machine. Kind of "worst of both worlds"; neither high-level enough, neither low-level enough, just archaic. Same goes for C++.
> 
> The important takeaway for me is that the "illusion of control" of C just gets in the way these days; it would be far better to just let describe "intent" using much higher-level abstractions then let the optimising compiler do its job properly. There is an interesting analogy with Fortran, and that's one of the reasons why C never reached Fortran speeds; with all that micro-management of "illusory low-levelness" you just make the job of the optimising compiler a lot harder.
> 
> Anyway, I found it interesting, and integer promotion in C makes me cry every single time I run into it...

Ok - read (most) of it. 

If we went back in time to when C was authored and those memory layout guarantees were put in place.. would the article still be true? I don't think so. Surely those guarantees were trivial and mapped 1:1 to mainframes and their non-parallelized CPU's inner workings. 

I would also lament that C's old memory state model isn't a good fit for instruction-hungry processors that are digesting (hundreds!) of adjacent instructions all in parallel and speculatively to try to maximize performance.

But is that's the circa 1970's C language spec's fault? It was a perfect fit at the time.  To me, it's the C working committee's fault for having too weak a backbone to strongly deprecate older features and also firmly mandate new ones - and provide a strong version firewall between language versions.

It should be like Python: there's 2.x and 3.x: you can't use the inferior 2.x language features in 3.x and just carry on. 

He might have been careful to only bash C and not C++ though.

C++ is really trying hard (and succeeding) at moving to code-by-intent, which is the purpose of  `<algorithm>`: it lets optimizing compilers  "achieve the requested action" using any means possible, in any order (or in parallel at the SIMD or thread level).

there have been lots of attempts at this, like Intel's SIMD and parallel pragmas, and with hybrid C languages like ISPC.

So this is super exciting that C++ is finally at this state - and will get even more of it; perhaps even GPU/compute offload.

Fully agree though - people just need to stop using it all together, especially the old C language constructs, and move to `<algorithm>` and the other nice new stuff!

.github/workflows/pvs-studio.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           general_criteria="GA:1,2;64:1;OP:1,2,3;CS:1"
           stamp="$(date +'%Y-%m-%d_T%H%M')-${GITHUB_SHA:0:8}"
           reportdir="pvs-report/pvs-report-${stamp}"
-          disable_warnings="V002,V1042,V826,V802,V2008,V1071"
+          disable_warnings="V002,V801,V802,V826,V1042,V1071,V2008"
```
kcgen, line 77: thanks for the re-order!

include/rgb888.h
```diff
@@ -32,16 +34,44 @@ class Rgb888 {
...
+	{
+		[[maybe_unused]] constexpr auto MaxValue = (1 << 4) - 1;
+
+		assert(r4 <= MaxValue);
+		assert(g4 <= MaxValue);
+		assert(b4 <= MaxValue);
+
+		const uint8_t r8 = check_cast<uint8_t>(r4 | (r4 << 4));
+		const uint8_t g8 = check_cast<uint8_t>(g4 | (g4 << 4));
+		const uint8_t b8 = check_cast<uint8_t>(b4 | (b4 << 4));
```
kcgen, line 71: `const auto`

johnnovak:
> But is that concern due to the circa 1970's C language spec? It was a perfect fit at the time. To me, it's the C language committee's fault for not evolving the language to suit our security needs and the evolving hardware architectures. The C C language committee has too weak a backbone to strongly deprecate older features and firmly mandate new ones through a language-spec "firewall" that guarantees one or the other.

Oh yeah, in a way it's a victim of its own success; machines kept evolving, but the C language not so much—to the point that CPU manufacturers actually started to need to design their CPUs around the vast amounts of existing compiled C code. Which is a bit comical, but I guess yeah, the practical choice.

I guess what I find frustrating is when mankind collectively gets stuck with something that's just "good enough". But that's my problem, I know 😄 

> It should be like Python: there's 2.x and 3.x: you can't use the inferior 2.x language features in 3.x, full stop. But if the C language committee was managing Python, it would be a big ball of backward-compatible mud!

I'm quite certain of that 😄 

Ultimately, we have the benefit of hindsight in as armchair-language-critiques (like me 😎)

include/vga.h
```diff
@@ -755,6 +755,84 @@ union CrtcModeControlRegister {
...
+	//     0        0          0%
+	//     0        1         33%
+	//     1        0         66%
+	//     1        1        100%
+	bit_view<0, 1> b;
+	bit_view<1, 1> g;
+	bit_view<2, 1> r;
+	bit_view<3, 1> sb;
+	bit_view<4, 1> sg;
+	bit_view<5, 1> sr;
```
kcgen, line 829: hey.. another failing of C (and C++, damnit! :face_exhaling:) - unreliable bitfields. 

At least we've got our own. :laughing:

johnnovak:
Hey, and thanks for the review @kcgen 😄

src/hardware/vga_attr.cpp
```diff
@@ -213,93 +241,119 @@ void write_p3c0(io_port_t, io_val_t value, io_width_t)
...
+			//      0-3	Indicates number of pixels to shift the
+			// display left Value  9bit textmode   256color mode
+			//  Other modes 0          1               0 0 1 2 n/a 1
+			//              2          3               1 2 3 4 n/a 3
+			//              4          5               2 4 5 6 n/a 5
+			//              6          7               3 6 7 8 n/a 7
+			//              8          0               n/a     n/a
 			break;
-		case 0x14:	/* Color Select Register */
+
```
kcgen, line 262: (just a question - is the comment block still OK in this new wrapped-form?)

src/hardware/vga_attr.cpp
```diff
@@ -213,93 +241,119 @@ void write_p3c0(io_port_t, io_val_t value, io_width_t)
...
+	case 0x12: // Color Plane Enable Register (EGA & VGA)
+		return vga.attr.color_plane_enable;
+
+	case 0x13: // Horizontal Pixel Panning Register (EGA & VGA)
+		return vga.attr.horizontal_pel_panning;
+
+	case 0x14: // Color Select Register (VGA only)
+		return vga.attr.color_select;
+
```
kcgen, line 329: +1, so nice easier to read

kcgen (approved):
These VGA overhauls are making marked difference in readability, @johnnovak! 

Just a couple minor comments, merge away when ready!

johnnovak:
> There's also the fact that the x86 assembly language is arguably not low level anymore. The "registers" don't map directly to hardware registers anymore. They're more like variables to the CPU. All the instructors get broken down into micro-ops and then executed in a completely different order than you typed them. Even if C does map neatly to assembly, it doesn't necessarily map to the hardware.
> 
> I do wish they would add better SIMD support to the language though.

100%, when even x86 instruction set is a facade and the underlying microcode is something completely different and alien to that, I guess it becomes a bit of joke when trying to "hand-optimise" code. Sure, sometime you can achieve modest benefits and out-smart the compiler, but the future is machine-driven optimisation. I guess that was my "point", if I had any 😄 

That's part of the charm of retro-coding for me on old 8-bit or 16/32-bit processors. Even an 486 is quite oldschool compared to what we have now; I was counting instruction latencies when writing x86 asm and it was an easy to understand model. Even more so for the Motorola 68k, MOS 6502 or the Z80—there is no pipelining, no caches, just serial execution. There's a beauty in that, a warm fuzzy feeling that you as a single person can hold the workings of the machine fully in your head. I doubt there are more than say 10 people alive in the world today who 100% understand what happens when execute a program on a modern CPU monster, down to the lowest levels...

...but try to encode H.265 on a Motorola 68k or view a JPEG on a c64. You can't win in everything. 😎

kcgen:
> I guess what I find frustrating is when mankind collectively gets stuck with something that's just "good enough". But that's my problem, I know 😄

Lol.. yeah.. the steering committees sponsors and corporate members have billions of LOC of old C and C++ between all of them, and have massive contracts, etc. A language firewall like Python would strand most of their products and require massive re-write efforts.. so they've (always) been 100% "backward-compatible friendly". :facepalm: To all of our detriment!

The good news is the NSA has declared C and C++ unsafe - and recommends people stop using it. 

https://www.nsa.gov/Press-Room/News-Highlights/Article/Article/3215760/nsa-releases-guidance-on-how-to-protect-against-software-memory-safety-issues/

There's no wiggle room here. New "big-gov" contracts for sensitive materiel are simply going to block C and C++.

The only way to get around that wording is to literally release C++ 2.0.. we'll finally get that hardware firewall.  It took the smoking man at the NSA to give it to us :-)

include/rgb666.h
```diff
@@ -0,0 +1,71 @@
...
+	{
+		return !operator==(that);
+	}
+
+	// Scoped conversion helpers
+	static constexpr Rgb666 FromRgb888(const Rgb888 rgb888)
+	{
+		const uint8_t r6 = check_cast<uint8_t>(rgb888.red   >> 2);
+		const uint8_t g6 = check_cast<uint8_t>(rgb888.green >> 2);
+		const uint8_t b6 = check_cast<uint8_t>(rgb888.blue  >> 2);
```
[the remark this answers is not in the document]
johnnovak: Ok, wasn't sure about this so I had to see it for myself. TL;DR I'm making the change 😄 

```cpp
#include <iostream>

using namespace std;

int main()
{
    uint8_t   u8 = 0;
    uint16_t u16 = 0;
    uint32_t u32 = 0;
    int8_t    i8 = 0;
    int16_t  i16 = 0;
    int32_t  i32 = 0;

    auto  a_u8 =  u8;
    auto a_u16 = u16;
    auto a_u32 = u32;
    auto  a_i8 =  i8;
    auto a_i16 = i16;
    auto a_i32 = i32;

    cout << " uint8_t: " << typeid(u8).name()  << endl;
    cout << "uint16_t: " << typeid(u16).name() << endl;
    cout << "uint32_t: " << typeid(u32).name() << endl;
    cout << "  int8_t: " << typeid(i8).name()  << endl;
    cout << " int16_t: " << typeid(i16).name() << endl;
    cout << " int32_t: " << typeid(i32).name() << endl << endl;

    cout << "auto uint8_t: " << typeid(a_u8).name()  << endl;
    cout << "auto int16_t: " << typeid(a_u16).name() << endl;
    cout << "auto int32_t: " << typeid(a_u32).name() << endl;
    cout << "auto  int8_t: " << typeid(a_i8).name()  << endl;
    cout << "auto int16_t: " << typeid(a_i16).name() << endl;
    cout << "auto int32_t: " << typeid(a_i32).name() << endl;

    return 0;
}
```

Output:

```
% g++ -std=c++17 auto.cpp && ./a.out                                                           [cmd]
 uint8_t: h
uint16_t: t
uint32_t: j
  int8_t: a
 int16_t: s
 int32_t: i

auto uint8_t: h
auto int16_t: t
auto int32_t: j
auto  int8_t: a
auto int16_t: s
auto int32_t: i
```

src/hardware/vga_attr.cpp
johnnovak: Probably not, but it was screwed up to begin with. To be cleared up using my VGA book during my next refactoring exercise 😄

PR #2830, opened 2023-09-05
Update templates

runlevel5:
## Changes

* Split macOS options into 2
* Add `linux-ppc64le`

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -41,7 +41,8 @@ body:
         - Windows 8
         - Windows 7
         - Windows Other
-        - Mac OS
+        - macOS (Intel)
+        - macOS (Apple Silicon)
         - Linux x86
         - Linux x86_64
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -41,9 +41,11 @@ body:
         - Windows 8
         - Windows 7
         - Windows Other
-        - Mac OS
+        - macOS (Intel)
+        - macOS (Apple Silicon)
         - Linux x86
         - Linux x86_64
+        - Linux ppc64le
```
johnnovak, line 48: Well, PPC builds are not officially supported, so I'm unsure about this. But I'll defer to @kcgen who is the only person here who cares about those builds 😄

kcgen:
Thanks for the addition, @runlevel5!

If you'd like to try to add logging of the host OS, OS version, and host architecture to our status log output, this would be a big help and we currently don't have that visibility from our log files.

PR #2832, opened 2023-09-05
Debug/trace logging improvements

johnnovak:
When it comes to debugging, I'm very much an old-school `printf` trace logger type of person (_Who needs 'em stinkin' debuggers? Not this guy!_ 😎)

So I've been prefixing my `LOG_MSG` statements with stuff like to `>>>>>>>>` and `!!!!!!!!!` and `*******` to make them stand out from the rest, then started repurposing `LOG_ERR` and `LOG_WARNING` because the coloured output makes my debug/trace messages stand out even more.

Then today it dawned on me: I can actually make `LOG_DEBUG` output stuff in a different colour with loguru! I've also seen the light and realised that @weirddan455 was right all along—module-specific `DEBUG_MODULNAME` `ifdef`s are the way to go. So I started converting a few modules to that convention as well.

Here's how it looks in action with `mixer.cpp` and `vga_draw.cpp` debug logging enabled. As you can see, my VGA and mixer debug crap doesn't get drowned in the rest of the crap anymore. Crap stands out, loud and clear—it doesn't get better than that! 😎 🤘🏻 

...except when it does! 😆 I've also added `LOG_TRACE` for good measure which uses a nice purple colour .

image

### DIFF
--- a/include/logging.h
+++ b/include/logging.h
@@ -91,13 +91,13 @@ void GFX_ShowMsg(const char* format, ...)
 #endif // C_DEBUG
 
 #ifdef NDEBUG
-// DEBUG_LOG_MSG exists only for messages useful during development, and not to
+// LOG_DEBUG exists only for messages useful during development, and not to
 // be redirected into internal DOSBox debugger for DOS programs (C_DEBUG feature).
-#define DEBUG_LOG_MSG(...)
+#define LOG_DEBUG(...)
 #else
 
 template <typename... Args>
-void DEBUG_LOG_MSG(const std::string& format, const Args&... args) noexcept
+void LOG_DEBUG(const std::string& format, const Args&... args) noexcept
 {
 	const auto format_green = std::string(loguru::terminal_green()) +
 	                          loguru::terminal_bold() + format +
--- a/include/logging.h
+++ b/include/logging.h
@@ -106,6 +106,17 @@ void LOG_DEBUG(const std::string& format, const Args&... args) noexcept
 	DLOG_F(INFO, format_green.c_str(), args...);
 }
 
+template <typename... Args>
+void LOG_TRACE(const std::string& format, const Args&... args) noexcept
+{
+	const auto format_purple = std::string(loguru::terminal_purple()) +
+	                           loguru::terminal_bold() + format +
+	                           loguru::terminal_reset();
+
+	DLOG_F(INFO, format_purple.c_str(), args...);
+}
+
+
 #endif // NDEBUG
 
 #endif
--- a/include/logging.h
+++ b/include/logging.h
@@ -20,6 +20,7 @@
 #define DOSBOX_LOGGING_H
 
 #include <cstdio>
+#include <string>
 
 #include "compiler.h"
 
@@ -56,11 +57,12 @@ public:
 		d_type(type),
 		d_severity(severity)
 		{}
-	void operator() (char const* buf, ...) GCC_ATTRIBUTE(__format__(__printf__, 2, 3));  //../src/debug/debug_gui.cpp
-
+	        void operator()(const char* buf, ...)
+	                GCC_ATTRIBUTE(__format__(__printf__, 2, 3)); //../src/debug/debug_gui.cpp
 };
 
-void DEBUG_ShowMsg(char const* format,...) GCC_ATTRIBUTE(__format__(__printf__, 1, 2));
+void DEBUG_ShowMsg(const char* format, ...)
+        GCC_ATTRIBUTE(__format__(__printf__, 1, 2));
 #define LOG_MSG DEBUG_ShowMsg
 
 #define LOG_INFO(...)    LOG(LOG_ALL, LOG_NORMAL)(__VA_ARGS__)
@@ -72,11 +74,12 @@ void DEBUG_ShowMsg(char const* format,...) GCC_ATTRIBUTE(__format__(__printf__, 
 struct LOG
 {
 	inline LOG(LOG_TYPES, LOG_SEVERITIES){ }
-	inline void operator()(char const* , ...){ }
+	inline void operator()(const char*, ...) {}
 }; //add missing operators to here
 
 	//try to avoid anything smaller than bit32...
-void GFX_ShowMsg(char const* format,...) GCC_ATTRIBUTE(__format__(__printf__, 1, 2));
+void GFX_ShowMsg(const char* format, ...)
+        GCC_ATTRIBUTE(__format__(__printf__, 1, 2));
 
 // Keep for compatibility
 #define LOG_MSG(...)	LOG_F(INFO, __VA_ARGS__)
@@ -92,7 +95,17 @@ void GFX_ShowMsg(char const* format,...) GCC_ATTRIBUTE(__format__(__printf__, 1,
 // be redirected into internal DOSBox debugger for DOS programs (C_DEBUG feature).
 #define DEBUG_LOG_MSG(...)
 #else
-#define DEBUG_LOG_MSG(...) DLOG_F(INFO, __VA_ARGS__)
+
+template <typename... Args>
+void DEBUG_LOG_MSG(const std::string& format, const Args&... args) noexcept
+{
+	const auto format_green = std::string(loguru::terminal_green()) +
+	                          loguru::terminal_bold() + format +
+	                          loguru::terminal_reset();
+
+	DLOG_F(INFO, format_green.c_str(), args...);
+}
+
 #endif // NDEBUG
 
 #endif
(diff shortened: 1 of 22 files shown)


include/logging.h
```diff
@@ -88,11 +91,36 @@ void GFX_ShowMsg(char const* format,...) GCC_ATTRIBUTE(__format__(__printf__, 1,
...
-// DEBUG_LOG_MSG exists only for messages useful during development, and not to
+// LOG_DEBUG exists only for messages useful during development, and not to
 // be redirected into internal DOSBox debugger for DOS programs (C_DEBUG feature).
-#define DEBUG_LOG_MSG(...)
+#define LOG_DEBUG(...)
 #else
-#define DEBUG_LOG_MSG(...) DLOG_F(INFO, __VA_ARGS__)
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-security"
```
johnnovak, line 100: I tried using `format_string` from `string_utils.h`, but that's a no-go because of circular dependency include hell or something.

Ultimately, I'm completely fine with shutting up those security warnings here because we're in 100% control over the format strings. That's a generic warnings that is not applicable for this specific use case at all (plus we're not OpenSSL 😅).

If someone has a "better" solution that I can just paste in then I will, but I won't spend more time on this as it works and is fit for purpose I think.

kcgen: Suggest moving `-Wformat-security` to meson.build flags, and avoid adding a new pragma that (mostly) touches all the codebase.

Currently the only `#pragma` used is when we need packed structs to match the DOS-side struct layout (and only for on Microsoft Visual Studio):

**`git grep '^#pragma' src/ *.cpp *.c | grep -v /libs/`**

```c++
src/dos/dos_execute.cpp:#pragma pack(1)
src/dos/dos_execute.cpp:#pragma pack()
src/dos/dos_files.cpp:#pragma pack (1)
src/dos/dos_files.cpp:#pragma pack()
src/dos/dos_mscdex.cpp:	#pragma pack(1)
src/dos/dos_mscdex.cpp:	#pragma pack()
src/dos/dos_tables.cpp:#pragma pack(1)
src/dos/dos_tables.cpp:#pragma pack ()
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/hardware/imfc.cpp:#pragma pack(push, 1)
src/hardware/imfc.cpp:#pragma pack(pop)
src/ints/int10_misc.cpp:#pragma pack(1)
src/ints/int10_misc.cpp:#pragma pack()
src/ints/int10_vesa.cpp:#pragma pack (1)
src/ints/int10_vesa.cpp:#pragma pack()
src/ints/xms.cpp:#pragma pack (1)
src/ints/xms.cpp:#pragma pack ()
```

johnnovak: > Suggest moving `-Wformat-security` to meson.build flags, and avoid adding a new pragma that (mostly) touches all the codebase.
> 
> Currently the only `#pragma` used is when we need packed structs to match the DOS-side struct layout (and only for on Microsoft Visual Studio):

But why is that approach preferable or better? This way the warning is localised to the `LOG_DEBUG` and `LOG_TRACE` definitions with laser focus.

Ignoring these warnings at the meson.build level would turn them off wholesale? Are you saying we don't care about these security warnings in general anyway because DOSBox does not have high security requirements? (Which I can agree with.)

In more general, judicious use of pragmas is fine I think, ahen you need them you need them. I wouldn't avoid them as a matter of principle.

johnnovak: Btw, there's more in `include`. And even more in `scr/libs`. We're already disabling `-Wformat-security` in `string_utils.h`.

image

johnnovak: > avoid adding a new pragma that (mostly) touches all the codebase.

To be clear, I don't think that's true. That warnings is only turned off `inside` those debug and trace log statements, effectively. It does not matter that they're included say 20k times in other files; outside of the debut and trace statements, those are completely unaffected.

kcgen: > But why is that approach preferable or better? 

I'm not a fan of pragma because it's a pre-processor directive that "branches" at the compiler level, inside the code itself.  The branch being those compilers that recognize the pragma versus those that don't (I'm just speaking in general).

Given any equivalent alternative that's more universal (and not in the source files), I'd rather go that route. So moving this into meson with other C-flags is one way.

> Ignoring these warnings at the meson.build level would turn them off wholesale? Are you saying we don't care about these security warnings in general anyway

Yup.  We're already knocking them in string_utils.h and now in logging.h: both of these headers get included quite high up to the point where (almost all the time), this flag is going into be in-effect.

Plus, if it was historically already such a problem we had to knock it out from our string_util's dealings, then I think it's give us enough headaches already. Time for it to be tossed entirely.

Microsoft's similar "use snprintf" warnings were in a similar boat and we had to shut them off too. Thankfully with static analyzers plus ASAN builds, we have no problems catching any nasty C-string memory hickups!

> In more general, judicious use of pragmas is fine I think; when you need them you need them. I wouldn't avoid them as a matter of principle.

My principle is to avoid them as strongly as other preprocessor directives: move as much as you can into agnostics standard C++. But yeah: if you /need/ to use 'em, then so be it!

Some teams buy heavily into #pragma-based coding styles.
For example:

```c++
#pragma startup my_function 3
#pragma exit other_function 1
```

Where you can control which function(s) run at startup (before the main function is called) and program exit (just before the program terminates through _exit). 

The numbers even lets you control the order (the numbers)! So these become band-aids for badly managed flow-control. 

Whats worse.. they're only honored by Clang and GCC - so MSVC will break :-)

kcgen: > even more in scr/libs

Yup, that's (in theory) 3rd party code that we're trying to avoid maintaining and owning.  So we just copy in new updates from upstream.. and they're free to use pragmas.

The downside is that sloppy pragma's in their headers can affect our own .cpp when we include them. But thankfully they wrap their pragma's inside even more pre-processor protection layers:

enet.h:

```c++
#if defined(_WIN32)
    #if defined(_MSC_VER) && defined(ENET_IMPLEMENTATION)
        #pragma warning (disable: 4267) // size_t to int conversion
        #pragma warning (disable: 4244) // 64bit to 32bit int
        #pragma warning (disable: 4018) // signed/unsigned mismatch
        #pragma warning (disable: 4146) // unary minus operator applied ```

Those thankful only kick in inside the `ENET_IMPLEMENTATION` chunk, which is their .cpp file. But yeah.. messy stuff. Just fix the code guys!

johnnovak: Yeah I'm fine with turning them off wholesale. Just wasn't sure about that, that's all, as these pragmas only affect the log statements.

Yeah, I agree pragmas are not great when overused. But not all projects are dealing with server-grade stuff, plus users doing snprintf based attacks is such a moot point in DOSBox when you can make them run arbitrary code inside the emulator anyway 😄 

Ok, removing the warning 😄

kcgen: Ok, removing the warning 😄

(:rocket: ! blow it away from string_utils.h, too)

johnnovak:
Done @kcgen , let me know if there's a better way to disabled the warning in `meson.build`.

kcgen (approved):
Looks good; nice to see more loguru benefits being put to use!

PR #2839, opened 2023-09-06
Extract MIXER command into its own file

johnnovak:
Prequel to https://github.com/dosbox-staging/dosbox-staging/issues/2468

`mixer.cpp` has become quite large, so the more things we can move out of it, the better.

This incidentally also forced better modularisation and API barriers; the MIXER command can no longer poke in the mixer's internals. 🎉 

No functional changes; the MIXER command will be extensively improved as part of https://github.com/dosbox-staging/dosbox-staging/issues/2468, and then it will be properly unit tested (now it's a bit of a mess; half of the bugs are caused by the very loose validation of the CLI input).

Suggest to review commit by commit.


kcgen:
Very nice change! Put that mixer.cpp on a diet :running_man:  :-)

kcgen (approved):
All looks good from my side. Looking forward to pure float pass-through!

johnnovak:
Thanks for the review @kcgen 😄

Kappa971:
Isn't the mixer command using MORE?
scr1

With `mixer /?|more` it works correctly:
scr2

johnnovak:
> Isn't the mixer command using MORE?

Looks like it isn't. I'll see what I can do, probably it's not hard (the English text fits on a single page, btw).

Kappa971:
> Looks like it isn't. I'll see what I can do, probably it's not hard (the English text fits on a single page, btw).

I think all commands use MORE (right?), so MIXER would be the only one not using it.
The English text would also benefit (the `Where` and `Note` sections have no space).

johnnovak:
> > Looks like it isn't. I'll see what I can do, probably it's not hard (the English text fits on a single page, btw).
> 
> I think all commands use MORE (right?), so MIXER would be the only one not using it. The English text would also benefit from it (the `Where` and `Note` sections have no space and crossfeed, reverb and chorus on the same line).

Don't know if all commands use MORE, I think you need to address each command individually. Yeah when I rewrote the MIXER help I was shuffling things around until it fit onto a single page. The English version 😄 With MORE support, I can add back those blank lines, etc. Thanks for flagging this @Kappa971.

FeralChild64:
All the commands should have used the MORE engine for displaying help now - unfortunately, it seems I’ve missed MIXER when doing the rework in commit https://github.com/dosbox-staging/dosbox-staging/commit/c82467945f16be3b79f93a7cd3140ea626d7e793, sorry for this.

It’s actually trivial to switch the command to use MORE for help display - see changes to `program_keyb.cpp` in the commit I’ve linked.

johnnovak:
> All the commands should have used the MORE engine for displaying help now - unfortunately, it seems I’ve missed MIXER when doing the rework in commit [c824679](https://github.com/dosbox-staging/dosbox-staging/commit/c82467945f16be3b79f93a7cd3140ea626d7e793), sorry for this.
> 
> It’s actually trivial to switch the command to use MORE for help display - see changes to `program_keyb.cpp` in the commit I’ve linked.

Thanks for the tip @FeralChild64, and no worries—everybody misses a few things from time to time 😄 I'm rewriting the MIXER now, so I'll address this as well.

PR #2840, opened 2023-09-06
Fix fallback shader scaling for adaptive CRT shaders

johnnovak:
Always disable integer scaling in `integer_scaling = auto` mode for adaptive CRT shaders when the fallback `sharp` shader is active.

This fixes a rather obscure bug that manifests when:

- An adaptive CRT shader is selected (`glshader = crt-auto-*`)
- The default `integer_scaling = auto` mode is selected
- The fallback `sharp` shader is active because the minimum vertical scaling ratio requirement hasn't been met
- When the fallback shader activation is triggered first by a screen mode change or a window resize event, the integer scaling is turned off **_— this is correct_** ✅ 
- But if you then press the "Reload Shader" shortcut, or enter/exit fullscreen, the vertical integer scaling will be activated again **_— this is a bug_** 🛑

### DIFF
--- a/src/gui/shader_manager.cpp
+++ b/src/gui/shader_manager.cpp
@@ -157,8 +157,18 @@ void ShaderManager::LoadShader(const std::string& shader_name)
 	const auto settings = ParseShaderSettings(new_shader_name,
 	                                          current_shader.source);
 
-	const bool is_adaptive = (mode != ShaderMode::Single);
-	current_shader.info    = {new_shader_name, settings, is_adaptive};
+	const bool is_adaptive = [&] {
+		if (mode == ShaderMode::Single) {
+			return false;
+
+		} else {
+			// This will turn off vertical integer scaling for the
+			// 'sharp' shader in 'integer_scaling = auto' mode
+			return (new_shader_name != SharpShaderName);
+		}
+	}();
+
+	current_shader.info = {new_shader_name, settings, is_adaptive};
 }
 
 const ShaderInfo& ShaderManager::GetCurrentShaderInfo() const
@@ -358,7 +368,6 @@ void ShaderManager::MaybeAutoSwitchShader()
 			LOG_MSG("RENDER: Using shader '%s'",
 			        current_shader.info.name.c_str());
 		}
-		current_shader.info.is_adaptive = false;
 
 	} else {
 		auto shader_changed = false;
@@ -384,11 +393,6 @@ void ShaderManager::MaybeAutoSwitchShader()
 			LOG_MSG("RENDER: Auto-switched to shader '%s'",
 			        current_shader.info.name.c_str());
 		}
-
-		// This will turn off vertical integer scaling for the 'sharp'
-		// shader in 'integer_scaling = auto' mode
-		current_shader.info.is_adaptive = (current_shader.info.name !=
-		                                   SharpShaderName);
 	}
 }
 


src/gui/shader_manager.cpp
```diff
@@ -157,8 +157,18 @@ void ShaderManager::LoadShader(const std::string& shader_name)
...
+	const bool is_adaptive = [&] {
+		if (mode == ShaderMode::Single) {
+			return false;
+
+		} else {
+			// This will turn off vertical integer scaling for the
+			// 'sharp' shader in 'integer_scaling = auto' mode
+			return (new_shader_name != SharpShaderName);
+		}
+	}();
```
weirddan455, line 169: Not seeing a need for this lambda.  This can just be:

```cpp
const bool is_adaptive = mode == ShaderMode:Single && new_shader_name != SharpShaderName;
```

johnnovak: Yeah, I prefer the lambda because it makes the intent clearer, plus I can include the comment in the second branch. I don't like too "clever" compound if conditions; I like things to be spelled out and a bit verbose because it helps understanding

E.g. I'm reading your condensed version now and I'm finding it confusing because of the `&&`—even though I wrote it! 😅  The lambda version is very clear to me.

johnnovak: Also, your version doesn't do the same thing at all... so yeah, confusing, as I said it 😄

weirddan455: Oh right, should be a != first:

```cpp
const bool is_adaptive = mode != ShaderMode:Single && new_shader_name != SharpShaderName;
```

Yeah, fair enough.  I'm just not a huge fan of lambda but it's not a big deal.  Leave it if you want.

johnnovak: By default, I write code for readability, first and foremost. If I can understand the verbose version in 2 seconds, and the clever version in 10 seconds, I'll always pick the the verbose version, all things being equal 😄

weirddan455 (approved):
Seems like a pretty straight forward and small bug fix.  Nice job.

src/gui/shader_manager.cpp
johnnovak: But I get what you're saying about disliking the lambdas @weirddan455 . It's the poor man's substitute for nicer language constructs like the `when` block in Kotlin. The fact that C/C++ makes a distinction between "statements" and "expressions" doesn't help either. In nicer languages, expressions can occur pretty much anywhere which allows for expressing yourself better in code.

For example, some Kotlin'ish pseudo-code how this would ideally look like in C++2050 😎 I was kind "simulating" this.

```kotlin
val is_adaptive = when {
  mode == ShaderMode::Single -> false
  else ->
    // This will turn off vertical integer scaling for the
    // 'sharp' shader in 'integer_scaling = auto' mode
    new_shader_name != SharpShaderName
}

```

johnnovak: Well, some guy with Stroustrup in his name is already considering adding pattern matching and when blocks at some point, so maybe we can use it by 2030-2040 or something 😄 

https://www.stroustrup.com/pattern-matching-November-2014.pdf

johnnovak:
> Seems like a pretty straight forward and small bug fix. Nice job.

Thanks for the review @weirddan455 👍🏻

src/gui/shader_manager.cpp
weirddan455: Yeah, it's mostly that I just don't like the C++ syntax for lambdas with needing to pass it parameters that are already local to the function.

I also watched a Cppcon talk with Chandler Carruth (works on LLVM/Clang at Google) a while back where he said lambdas can be hard for the compiler to optimize well.  Hopefully that's not applicable to something this simple but still, something to keep in mind.

johnnovak: > I also watched a Cppcon talk with Chandler Carruth (works on LLVM/Clang at Google) a while back where he said lambdas can be hard for the compiler to optimize well. Hopefully that's not applicable to something this simple but still, something to keep in mind.

Good to know and to keep in mind @weirddan455!

Generally, we're using lambdas to make non-performance intensive code nicer and/or read better by removing redundancies, so that's fine I think. This particular code gets executed once in a blue moon on screen mode, window size, and fullscreen/windowed changes, so we can do pretty much whatever as it has zero performance implications. When we actuall trigger a shader change, that does about 10000x more work, so... 😄

PR #2842, opened 2023-09-06
Add support for disabling screen via the clocking mode register, fixes #2838

### LINKED ISSUE #2838: Clocking mode screen disable bit is mishandled

_No response_

Linux ppc64le

_No response_

### Relevant hardware info

_No response_

This is apparently a very old bug and doesn't affect many programs, but it does affect Alien Carnage/Halloween Harry. I like this game, and my wife likes this game because Australians wrote it. See earlier report (from 2007!) here: https://www.dosbox.com/comp_list.php?showID=950&letter=H

What happens is that the VGA display never blanks even when the program turns it off. This manifests in screens painting, fading out and fading in, when they should simply fade in from black.

This is because DOSBox improperly aliases the bit in the clocking register which disables screen display with other sources; although the `ClockingModeRegister` struct has `bit_view<5, 1> is_screen_disabled;`, it never actually gets used by anything. Instead, `vga_seq.cpp` sets `vga.attr.disabled` in `write_p3c5()`, which is the only thing checked in `vga_draw.cpp`. Unfortunately, in a debug build, it can be shown that `vga.attr.disabled` gets frequently reset by other writes to port `0x3c0` in `write_p3c0()` in `vga_attr.cpp`, so the value of the clocking mode register is thus dropped and the screen never blanks.

The problem affects upstream as well; it doesn't look like anyone bothered to sit down and fix it. Does the project allow local fixes?

Explain how to reproduce
1. Run Alien Carnage. It happens right from the Apogee logo and everywhere else like the main menu.

https://www.myabandonware.com/game/alien-carnage-1pe

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

classilla:
Fixes #2838 by painting black for screen lines when the screen disable bit is set in the clocking mode register. This wasn't ever checked before.

Fixes: Alien Carnage (a/k/a Halloween Harry)

Regression tested against: Descent (shareware), Dark Forces (full), Ultimate Doom, Quake, Blake Stone AoG, MDK, Pinball Illusions, CyberMage Dark Awakening, Epic Pinball, Extreme Pinball (yes, I mostly play pinball and shooters), Death Rally, GeoWorks Ensemble 2.0

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -427,6 +427,13 @@ static uint8_t* draw_linear_line_from_dac_palette(Bitu vidstart, Bitu /*line*/)
 	auto pixels_remaining = check_cast<uint16_t>(vga.draw.line_length /
 	                                             bytes_per_pixel);
 
+	// If the screen is disabled, just paint black. This fixes screen
+	// fades in titles like Alien Carnage.
+	if (GCC_UNLIKELY(vga.seq.clocking_mode.is_screen_disabled)) {
+		memset(line_addr, 0, vga.draw.line_length);
+		return TempLine;
+	}
+
 	// see VGA_Draw_Linear_Line
 	if (GCC_UNLIKELY((vga.draw.line_length + offset) & ~vga.draw.linear_mask)) {
 		// Note: To exercise these wrapped scenarios, run:


johnnovak:
Thanks @classilla, I will give this some further testing today before approving & merging.

johnnovak:
Works splendidly @classilla 🎉 Couldn't find any issues with other games. Very few programs seem to use this bit, which makes this a very safe change anyway.

johnnovak (approved):

PR #2854, opened 2023-09-11
Add FOR Command

### LINKED ISSUE #448: FOR loops support in batch files
I just discovered that DosBox doesn't recognize FOR loops (and probably other common batch functionalities). I'm pretty sure this is already preventing many programs to properly install and run.

Could that be added?

Regards and thanks for all the cleanup and improvements you're making to DosBox.

MeAreJeenius:
# Description

Added the FOR Command.

To the best of my knowledge, this behaves exactly like the FOR command on DOS.

Resolves #448

Also added a string_view delimiters parameter to the split function, allowing splitting on multiple characters of the caller's choosing.

# Manual testing
TODO

# Checklist

TODO

I have:

### DIFF
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -262,7 +262,8 @@ void strip_punctuation(std::string& str);
 //   split("::", ':') returns {"", "", ""}
 std::vector<std::string> split(std::string_view seq, char delim);
 
-// Split a string on whitespace, where whitespace can be any of the following:
+// Split a string on any character found in delim.
+// Delim defaults to all whitespace characters:
 // ' '    (0x20)  space (SPC)
 // '\t'   (0x09)  horizontal tab (TAB)
 // '\n'   (0x0a)  newline (LF)
@@ -276,7 +277,8 @@ std::vector<std::string> split(std::string_view seq, char delim);
 //   split("a\tb\nc\vd e\rf") returns {"a", "b", "c", "d", "e", "f"}
 //   split("  ") returns {}
 //   split(" ") returns {}
-std::vector<std::string> split(std::string_view seq);
+std::vector<std::string> split(std::string_view seq,
+                               std::string_view delims = " \f\n\r\t\v");
 
 std::string join_with_commas(const std::vector<std::string>& items,
                              const std::string_view and_conjunction = "and",
--- a/src/shell/shell_cmds.cpp
+++ b/src/shell/shell_cmds.cpp
@@ -50,7 +50,7 @@
 
 // clang-format off
 static const std::map<std::string, SHELL_Cmd> shell_cmds = {
-	{ "CALL",     {&DOS_Shell::CMD_CALL,     "CALL",     HELP_Filter::All, HELP_Category::Batch } },
+	{ "CALL",     {&DOS_Shell::CMD_CALL,     "CALL",     HELP_Filter::All,    HELP_Category::Batch } },
 	{ "CD",       {&DOS_Shell::CMD_CHDIR,    "CHDIR",    HELP_Filter::Common, HELP_Category::File } },
 	{ "CHDIR",    {&DOS_Shell::CMD_CHDIR,    "CHDIR",    HELP_Filter::All,    HELP_Category::File } },
 	{ "CLS",      {&DOS_Shell::CMD_CLS,      "CLS",      HELP_Filter::Common, HELP_Category::Misc} },
@@ -62,6 +62,7 @@ static const std::map<std::string, SHELL_Cmd> shell_cmds = {
 	{ "ECHO",     {&DOS_Shell::CMD_ECHO,     "ECHO",     HELP_Filter::All,    HELP_Category::Batch } },
 	{ "ERASE",    {&DOS_Shell::CMD_DELETE,   "DELETE",   HELP_Filter::All,    HELP_Category::File } },
 	{ "EXIT",     {&DOS_Shell::CMD_EXIT,     "EXIT",     HELP_Filter::Common, HELP_Category::Misc } },
+	{ "FOR",      {&DOS_Shell::CMD_FOR,      "FOR",      HELP_Filter::All,    HELP_Category::Batch } },
 	{ "GOTO",     {&DOS_Shell::CMD_GOTO,     "GOTO",     HELP_Filter::All,    HELP_Category::Batch } },
 	{ "IF",       {&DOS_Shell::CMD_IF,       "IF",       HELP_Filter::All,    HELP_Category::Batch } },
 	{ "LH",       {&DOS_Shell::CMD_LOADHIGH, "LOADHIGH", HELP_Filter::All,    HELP_Category::Misc } },
@@ -81,7 +82,7 @@ static const std::map<std::string, SHELL_Cmd> shell_cmds = {
 	{ "TYPE",     {&DOS_Shell::CMD_TYPE,     "TYPE",     HELP_Filter::Common, HELP_Category::Misc } },
 	{ "VER",      {&DOS_Shell::CMD_VER,      "VER",      HELP_Filter::All,    HELP_Category::Misc } },
 	{ "VOL",      {&DOS_Shell::CMD_VOL,      "VOL",      HELP_Filter::All,    HELP_Category::Misc } }
-	};
+};
 // clang-format on
 
 // support functions
@@ -2570,5 +2571,123 @@ void DOS_Shell::CMD_MOVE(char* args)
 				}
 			}
 		}
+	}
+}
+
+static std::vector<std::string> search_files(std::string_view query)
+{
+	const auto save_dta = dos.dta();
+	dos.dta(dos.tables.tempdta);
+	const auto dta = DOS_DTA(dos.dta());
+
+	bool found = DOS_FindFirst(std::string(query).c_str(),
+	                           ~(DOS_ATTR_VOLUME | DOS_ATTR_DIRECTORY));
+
+	std::vector<std::string> files = {};
+	while (found) {
+		DOS_DTA::Result result = {};
+		dta.GetResult(result);
+		files.emplace_back(std::move(result.name));
+		found = DOS_FindNext();
+	}
+
+	dos.dta(save_dta);
+	
+	return files;
+}
+
+void DOS_Shell::CMD_FOR(char* args)
+{
+	HELP("FOR");
+	
+	static const auto delimiters = std::string_view(",;= \t");
+
+	auto argsview = std::string_view(args);
+	auto consume_next_token = [&argsview]() -> std::optional<std::string_view> {
+
+		const auto start = argsview.find_first_not_of(delimiters);
+		if (start >= argsview.size()) {
+			return {};
+		}
+
+		const auto end = argsview.substr(start).find_first_of(delimiters);
+		if (end >= argsview.size()) {
+			return {};
+		}
+
+		const auto token = argsview.substr(start, end);
+		argsview         = argsview.substr(end + 1);
+		return token;
+	};
+
+	const auto variable = consume_next_token();
+	if (!variable || variable->size() != sizeof('%') + 1 ||
+	    variable->front() != '%') {
+		return;
+	}
+
+	const auto in_keyword = consume_next_token();
+	if (!in_keyword || !iequals(*in_keyword, "IN")) {
+		return;
+	}
+
+	const auto parameters =
+	        [&argsview]() -> std::optional<std::vector<std::string>> {
+		const auto list_start = argsview.find_first_of('(');
+		if (list_start >= argsview.size()) {
+			return {};
+		}
+		const auto list_end = argsview.substr(list_start).find_first_of(')');
+		if (list_end >= argsview.size()) {
+			return {};
+		}
+		auto string_list = split(argsview.substr(list_start + sizeof('('),
+		                                         list_end - sizeof(')')),
+		                         delimiters);
+		argsview = argsview.substr(list_end + sizeof(')') + 1);
+
+		auto parameters = std::vector<std::string>();
+		for (auto& str : string_list) {
+			if (str.find('*') == std::string::npos &&
+			    str.find('?') == std::string::npos) {
+				parameters.emplace_back(std::move(str));
+				continue;
+			}
+			auto files = search_files(str);
+			parameters.insert(parameters.end(),
+			                  std::make_move_iterator(files.begin()),
+			                  std::make_move_iterator(files.end()));
+		}
+		return parameters;
+	}();
+
+	if (!parameters) {
+		return;
+	}
+
+	const auto do_keyword = consume_next_token();
+	if (!do_keyword || !iequals(*do_keyword, "DO")) {
+		return;
+	}
+
+	const auto format_command = std::string(argsview);
+
+	for (const auto& parameter : *parameters) {
+		
+		// TODO: C++20: Use std::vformat instead
+		auto command  = format_command;
+		auto position = std::string::size_type();
+		while ((position = command.find(*variable, position)) !=
+		       std::string::npos) {
+			command.replace(position, variable->size(), parameter);
+			position += parameter.size();
+		}
+		
+		// TODO: Pass command directly to ParseLine when it takes
+		// a std::string_view instead of a char*
+		char cmd_array[CMD_MAXLINE];
+		std::strncpy(cmd_array, command.c_str(), CMD_MAXLINE);
+		cmd_array[CMD_MAXLINE - 1] = '\0';
+		ParseLine(cmd_array);
 	}
 }
--- a/include/shell.h
+++ b/include/shell.h
@@ -159,6 +159,7 @@ public:
 	void CMD_DELETE(char* args);
 	void CMD_ECHO(char* args);
 	void CMD_EXIT(char* args);
+	void CMD_FOR(char* args);
 	void CMD_MKDIR(char* args);
 	void CMD_CHDIR(char* args);
 	void CMD_RMDIR(char* args);
--- a/src/misc/string_utils.cpp
+++ b/src/misc/string_utils.cpp
@@ -148,30 +148,29 @@ std::vector<std::string> split(const std::string_view seq, const char delim)
 	return words;
 }
 
-std::vector<std::string> split(const std::string_view seq)
+std::vector<std::string> split(const std::string_view seq, const std::string_view delims)
 {
 	std::vector<std::string> words;
-	if (seq.empty())
+	if (seq.empty()) {
 		return words;
-
-	constexpr auto whitespace = " \f\n\r\t\v";
+	}
 
 	// count words to reserve space in our vector
 	size_t n  = 0;
-	auto head = seq.find_first_not_of(whitespace, 0);
+	auto head = seq.find_first_not_of(delims, 0);
 	while (head != std::string::npos) {
-		const auto tail = seq.find_first_of(whitespace, head);
-		head            = seq.find_first_not_of(whitespace, tail);
+		const auto tail = seq.find_first_of(delims, head);
+		head            = seq.find_first_not_of(delims, tail);
 		++n;
 	}
 	words.reserve(n);
 
 	// populate the vector with the words
-	head = seq.find_first_not_of(whitespace, 0);
+	head = seq.find_first_not_of(delims, 0);
 	while (head != std::string::npos) {
-		const auto tail = seq.find_first_of(whitespace, head);
+		const auto tail = seq.find_first_of(delims, head);
 		words.emplace_back(seq.substr(head, tail - head));
-		head = seq.find_first_not_of(whitespace, tail);
+		head = seq.find_first_not_of(delims, tail);
 	}
 
 	// did we reserve the exact space needed?
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -627,6 +627,9 @@ static const char* const full_name      = "Z:\\COMMAND.COM";
 static const char* const init_line      = "/INIT AUTOEXEC.BAT";
 
 void SHELL_Init() {
+	
+	//clang-format off
+	
 	// Generic messages, to be used by any command or DOS program
 	MSG_Add("SHELL_ILLEGAL_PATH", "Illegal path.\n");
 	MSG_Add("SHELL_ILLEGAL_FILE_NAME", "Illegal file name.\n");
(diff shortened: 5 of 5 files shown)


include/string_utils.h
```diff
@@ -276,7 +277,8 @@ std::vector<std::string> split(std::string_view seq, char delim);
 //   split("a\tb\nc\vd e\rf") returns {"a", "b", "c", "d", "e", "f"}
 //   split("  ") returns {}
 //   split(" ") returns {}
-std::vector<std::string> split(std::string_view seq);
+std::vector<std::string> split(std::string_view seq,
+                               std::string_view delims = " \f\n\r\t\v");
```
kcgen, line 281: This will mean we have two "split" routines with virtually identical interfaces (char* vs. string_view, which itself can accept a char*) but with different behavior:




---



---

We should only have one API:
 - `split(haystack, delims)`
 - `split(haystack)` <- simply calls `split(haystack, " \f\n\r\t\v")`

Can you check the existing use-cases of the `split(..)`. and see if you can boil it down to one implementation? (I really don't want to have two split rountines with different output)

johnnovak: Yes, just keep this new one and get rid of the old one that accepts a single char delimiter (and refactor accordingly, please).

src/shell/shell_cmds.cpp
```diff
@@ -2572,3 +2573,121 @@ void DOS_Shell::CMD_MOVE(char* args)
 		}
 	}
 }
+
+static std::vector<std::string> search_files(std::string_view query)
```
johnnovak, line 2577: const

src/shell/shell_cmds.cpp
```diff
@@ -2572,3 +2573,121 @@ void DOS_Shell::CMD_MOVE(char* args)
...
+	dos.dta(save_dta);
+	
+	return files;
+}
+
+void DOS_Shell::CMD_FOR(char* args)
+{
+	HELP("FOR");
+	
+	static const auto delimiters = std::string_view(",;= \t");
```
johnnovak, line 2603: Just use `std::string` or`const char *` everywhere, please. This is not performance intensive code, let's do the simplest thing.

src/shell/shell_cmds.cpp
```diff
@@ -2572,3 +2573,121 @@ void DOS_Shell::CMD_MOVE(char* args)
...
+			    str.find('?') == std::string::npos) {
+				parameters.emplace_back(std::move(str));
+				continue;
+			}
+			auto files = search_files(str);
+			parameters.insert(parameters.end(),
+			                  std::make_move_iterator(files.begin()),
+			                  std::make_move_iterator(files.end()));
+		}
+		return parameters;
```
johnnovak, line 2661: Can you drop this whole `string_view` and  move business, please; we're parsing an up to 256 char max string here, not a 100 MB XML 😅 All this is unnecessary complication at the code level.

Please reimplement this in the most basic and primitive way that does the job. If we're creating temporary strings left, right, center, who cares — this is not performance-intensive code.

johnnovak: (Otherwise, the code is nice and clearly structured, for 🥇 on that. But let's keep it simple 😅)

src/shell/shell_cmds.cpp
```diff
@@ -2572,3 +2573,121 @@ void DOS_Shell::CMD_MOVE(char* args)
...
+	for (const auto& parameter : *parameters) {
+		
+		// TODO: C++20: Use std::vformat instead
+		auto command  = format_command;
+		auto position = std::string::size_type();
+		while ((position = command.find(*variable, position)) !=
+		       std::string::npos) {
+			command.replace(position, variable->size(), parameter);
+			position += parameter.size();
+		}
```
johnnovak, line 2684: Can't you just use `format_string` from our string utils?

src/shell/shell_cmds.cpp
johnnovak: 1.6x slowdown, 10x slowdown, or 100x slowdown is meaningless without context.

I'm advocating for using a simple style and only complicating things when they make a tangible benefit. Getting better results in micro-benchmarks is not a tangible benefit. Overall, observable _whole-application performance_ and behaviour is.

Even in the DOS days, well, even before that, in the 80s, people realised such exercises are fairly meaningless in isolation. People wrote their games in C, then rewrote the hot loops in asm after benchmarking, and that's it. Config parsing, disk I/O—just use C, performance is a non-issue there.

I get what you're saying, though; when everything is slow, nothing really jumps out when benchmarking the top 100 slowest functions. But as I see it, we could write the shell commands in Python, and even them being 1000x slower than the "worst C++ code" you can imagine, no one would notice a difference even on the slowest hardware.

> Again, it's not about this one function in particular, it's about how it spreads to the program as a whole.

Yes, so it does not spread, because this is a FOR loop you execute from the DOS prompt... If you were poking around in the CPU emulation or something similar, yeah, then we need to be careful about such things. Not for DOS commands though that run once in a blue moon.

src/shell/shell_cmds.cpp
kcgen: Yup, normally the char* string literals are fine as:

`constexpr auto str = "....";`, but in this case you want the class benefits of std::string so it can upgraded:

```c++
constexpr std::string_view delimiters = "....";
```

The nice thing with strong literals is that they're permanently pinned I to the program, so this is a very simple and elegant way to get the stringview benefit without any risks.

src/shell/shell_cmds.cpp
```diff
@@ -2572,3 +2573,121 @@ void DOS_Shell::CMD_MOVE(char* args)
...
+		files.emplace_back(std::move(result.name));
+		found = DOS_FindNext();
+	}
+
+	dos.dta(save_dta);
+	
+	return files;
+}
+
+void DOS_Shell::CMD_FOR(char* args)
```
kcgen, line 2599: @johnnovak , @MeAreJeenius : I feel bad that my own buggy use of passing the compiled rendered shader as a string_view across modules (which I shouldn't have done) has kind of poisoned the use of string_view. 

I think when it's used in a self-contained manner like this, it's nothing more than a read-only std::string class wrapper around a char array.

But that said: maybe we can simplify this more at the functional level and take in `FOR(const std::string_view args)`, and just work from it directly?

As a reader, seeing that we're taking a `(char* args)` tells me that we intend to modify both the pointer *and* the char values themselves (is that really the intent?)

src/shell/shell_cmds.cpp
kcgen: Can you add a top level comment describing the  syntax that this function digests - including a real life example string that hits all the functional bits in the body?

Then include more comments below describing which chunk of the syntax each bit of logic is digesting, one by one, to make it easier to follow along. 

(Right now, I'm trying to reverse-engineer it in my head from the delimiter and search strings, and getting too overwhelmed to hold it all. My contest space is like ChatGPT 0.5 beta level 😅)

kcgen: My other major comment is that the entire body is using totally generic variable names, often simply named after the type itself. 

- args, variable, token, string_list, str, list_start, list_end, and parameters. Can these be named for the chunk or bits of the FOR loop components themselves? 

Or maybe they are named after the FOR components, but I just need help connecting the dots.

src/shell/shell_cmds.cpp
kcgen: I'm not sure I'm seeing where or how the code is being harmed in exchange for the micro optimizations.

In other words: can it be written in much fewer and easier to read lines, if we're willing to forfeit the measured 1.6x performance boost of string_view vs string copies? 

I don't see that a trade like that is possible (if I did, I would offer it as a code replacement suggestion). 

Yeah: I do see move being used, but it tells me wer're just taking a rock from pile A and placing it on pile B. 

Could the emplace, insert, move-iterator-makers being replaced with a for loop? Yup, it would be about the samr number of lines and would probably be a bit less busy to look at.

src/shell/shell_cmds.cpp
johnnovak: Yeah, all I'm saying, let's optimise for the _correct_ things. As one of my former team leads used to say: people always optimise for _something_, either consciously or unconsiously, but not always for the _correct_ thing.

Making something that takes 1ms to execute 10x faster but also making the code harder to understand by 10% is a net-negative contribution if that something is non-performance intensive and could run for 10ms or even 100ms and it wouldn't matter. That's optimising for the wrong thing. And like _meaningful_ performance improvements, these extra complexities that don't really solve anything but make the code less readable and harder to understand also add up, making the codebase overall a bit harder to maintain and understand.

All in all, the DOS programs and config file parsing are two areas where performance does not matter the _slightest_ compared to the rest of the emulator. All "inefficiencies" there are masked by the rest of the codebase by several orders of magnitude. So I would by all means make these "glue code" type of parts as readable as possible. I actually had an "outlandish" idea a while back that it would be nice if only the performance intensive parts were in C++, then all config parsing and the rest of the glue code in Python or something (no need to panic, I won't do it... too late for that, but that would have been a nice sweet-spot solution). LIke I said, you could write these "native" DOS programs in Python and no one would complain or notice the difference 😄

So yeah, the less of that iterator noise, the less mental overhead of thinking in terms of string views, pointers, copies, etc. the better in my book. Looking at this code I'm actually thinking of introducing more simple string utility functions, e.g. `std::find` is so noisy with all those iterator arguments. Ideally, I'd like non-performance intensive string mangling code like this to read like Python pseudocode.

But I won't die for it... 😄 Overall, this is no big deal, but yeah... it could be a lot simpler I think.

src/shell/shell_cmds.cpp
johnnovak: Yeah wasn't entirely sure, just thought it's worthwhile to check 😄

johnnovak:
Are you planning to add unit tests for this @MeAreJeenius ? That would be a welcome addition for somewhat complex commands like this.

src/shell/shell_cmds.cpp
kcgen: Fully agree @johnnovak, and as you say: the dosbox codebase continues to deliver 👍

Looking at the original C shell-parsing code, it was a nightmare maze of single-letter pointer arithmetic without any safeguards. ⛷️ 

I'm also not a fan of all the functions that take in .begin() and .end() iterators.. like c'mon C++ committee: give us a default argument where we can  pass in the container itself (and operate on all the content).

src/shell/shell_cmds.cpp
johnnovak: > How exactly is string_view not the simplest thing here?

Well, the simplest thing for me is this:

```ini
static const char* blah delimiters = ",;= \t";
```

or (if you need `std::string` methods):

```ini
static const std::string delimiters = ",;= \t";
```

Seeing `string_view` anywhere in the code kinda signals to me: "ok, this is performance intensive code, we're probably dealing with large strings, maybe many thousands of string manipulations in tight loops, so we want to avoid copies because it makes a big difference, etc." None of that happening here? Also it signals me to be careful not to let the `string_view` "leak out" from the function/module/etc because a `string_view` is basically a pointer-pair.

So I wouldn't call it the simplest thing to do 😄

shermp: > `static const std::string delimiters = ",;= \t";`

This cannot be a compile time constant however, unlike `string_view`.

src/shell/shell_cmds.cpp
johnnovak: > Could the emplace, insert, move-iterator-makers be replaced with a for loop? Yup, it would be about the same number of lines and would probably be a bit less busy. So maybe that's a small but minor suggestion.

That's a big part of it, I think; move and string_view just add insult to injury 😅 Maybe it would be quite alright by introducing a few helpers to string utils to remove all that iterator noise.

johnnovak: Also, can't we just take the incoming string, split it by the delimiters, that gives us a vector of strings. Then just operate on that vector, so need for that tokeniser that returns string views, etc. Just make it simpler. We're only dealing with less than 1000 bytes of input here... You could copy that 10 times and it still wouldn't matter one bit. Whatever is more readable and hopefully all that iterator business would go away.

johnnovak: > As a contributor, I hope you can at least appreciate how frustrating it can be to just be coding using completely run-of-the-mill C++, and to be told it's micro-optimization and that it complicates the code, without any further elaboration.

I don't think "run-of-the-mill C++" exists. For example, many people don't use the std lib at all—that's still C++. Many people still use it as "C with classes" and none of the C++11 and later features–that's still C++.

I'm hearing that you're frustrated, but then these discussions happen regularly only when reviewing your PRs @MeAreJeenius, not with other contributor's PRs. In a way, it's frustrating for me too, because if I contribute to another project, then I try to follow their established standards not change their project. I _might_ bring up my ideas for improvement once or twice, but if they don't want, I will not push it further.

Because you've asked about it, here's my personal perspective:

Many people are not that familiar with `string_view` and `move`, I'm actually one of them. I don't like C++, I don't use it daily at work (and that's the last thing I want to do 😅), I just want to use a minimal subset here that does the job. To put it differently, my love for DOSBox is greater than my dislike for C++ 😄

It's judgement call, but I personally think there are lots of gotchas and complications with `std::move` and `string_view`, you need to be careful to use them. E.g. if you store a string view somewhere, then the string it points to gets freed, well you're in trouble—that's extra complexity. Then I wanted to learn more about `std::move` and found blog posts that go on for 5-10-20 pages, discussing all the intricate details, edge case, gotchas, etc. No thanks, we can solve everything with pointers and references. There are already too many ways to shoot yourself in the foot in C++, these just give you more rope to hang yourself, metaphorically speaking.

From my experience, the problem with languages that have too many features is that people start using _all_ those features. Then if someone new want to contribute, he/she has to be familiar with all the 10 different ways of doing things, when in reality just restricting ourselves to 2 or 3 ways would have sufficed.

So basically that's where I'm coming from. I like to do the simplest solution that does the job.

src/shell/shell_cmds.cpp
kcgen: > Seeing string_view anywhere in the code kinda signals to me: "ok, this is performance intensive code".

But it shouldn't; well, the purpose of using it isn't solely for higher performance code. It's just a functional wrapper to strap on top of raw C-string. 

It says to me, "I only intent to perform read only operations to find / compare / check" using nicer C++ string calls instead of more dangerous.

It tells me were inspecting as opposed to authoring new strings.

The RHS is a const char[], so string_view is a perfect fit to make it a read

kcgen: (You can imagine a `std::vector_view<T>`, that takes in a C-array and lets you run .size(), .find(Val), .any_of(Val), etc..)

johnnovak: > > Seeing string_view anywhere in the code kinda signals to me: "ok, this is performance intensive code".
> 
> It's just a functional wrapper to strap on top of a read-only C-string.
> 
> It says to me, _"I only intent to perform read only operations to find / compare / check"_ using nicer C++ string calls instead of more dangerous C calls.
> 
> Basically: _inspecting_ as opposed to _authoring_ new strings.
> 
> We can inspect with std::string, too: but string_view makes that use-case guaranteed and clearer.

Yeah, and actually using string views as function args is one of the nicer use cases where it acts as a sort of a "wildcard" type that can accept both `std::string` and `const char *`.

I guess if you only use them internally and never store them in class members, etc., or only where the lifecycles of the actual strings it references are very clear, then they're ok, and no doubt can be helpful.

My problem with them is that once you use string views, you also need to consider the lifecycle of the underlying strings, and that's actually a source for bugs and/or crashes as we've seen. So to be 100% clear, _that_ is my only (and biggest at the same time) problem with them that it kinda forces us back into the "manual memory management" mindset. Unless I'm missing something obvious 😄 With references or straight copies that's never a problem.

src/shell/shell_cmds.cpp
kcgen: @MeAreJeenius , the way you've described these use cases is clear to me, and when re-reading your code, as you've explained, actually makes these aspects more self-documenting:

#### string_view

- I want to see a char buffer -> I use string_view. Simple. No need to think about why to use it or why it's there.
- Can't accidentally change characters if you forget const

#### move

- I want this object to go over there -> I use move. Simple. No need to think about why to use it or why it's there.
- Accidental use after move is caught by CI / SA. Accidental use after copy is not.

So I see your use as justified, especially when the intent is simple and clear as you've described. 

I think the dillema as reviewers is we haven't had a lot of exposure to these (so John and I might wonder.. is this just overly faffy syntax bloat?). But perhaps with fresh eyes, the intent becomes much clearer.

Things do get muddy when  lamdas are put inline into  algorithms, combined with iterator generators and others. I find sometimes going back to for loops easier to see. 

Would you agree with that take, @johnnovak?

include/string_utils.h
kcgen: I like your new rename_with_empties name. Let's go with the rename, and yours becomes the better default split() 👍

src/shell/shell_cmds.cpp
johnnovak: Yeah pretty much @kcgen, that's a good summary. I think there's a lot going on there that makes me wonder, is this really the simplest way to do this relatively simple job? But the C++ is very syntax heavy still, we could improve that with some helper functions but it's still far from pseudo-code level cleanness.

src/shell/shell_cmds.cpp
johnnovak: > The ranges library in C++20 has overloads of many of these functions. So `std::coolthing(v.begin(), v.end())` becomes `std::ranges::coolthing(v)`. The nested namespace is for some back-compat reasons that I can't recall, but at least it's an improvement, despite the verbose namespace.

Good to know. Not quite sure why we can't migrate to C++20 yet, maybe MSVC limitation or something?

src/shell/shell_cmds.cpp
kcgen: Good point about string_view introducting new concerns about lifetime of the content versus the object itself. The two are no longer coupled. 

There are probably some good rules of thumb or use-cases that can make this easy to review and ensure won't bite us: like the string_view objects only live (or have scope) as long as their content is in-scope. 

So self-contained use inside a function means that the string_view cannot outlive its content.

@MeAreJeenius , are there some rules like that?

johnnovak: To be fair, using string views only within a single function is as safe as it gets, no doubt. I guess I was reacting more to the _sum_ of all these things that are going on here, as @kcgen put it 😅 I'm a simple man, and I'm getting old, I just want it simple, hey 😎

src/shell/shell_cmds.cpp
kcgen: It's debian 10 (which includes retropie): unfortunately they don't have a c++20 compiler.

But debian 10 goes EOL in 9 months, so After 0.81 we can move to c++20 (knowing that debian 10 will be toast prior to the 0.82 release).

johnnovak: > Many of them mutate the args, so it can't be changed to const char* either, which is a massive pain.

Yeah, that's quite horrible (the mutation)...

src/shell/shell_cmds.cpp
johnnovak: My other problem with string views is that `.data()` is not guaranteed to return null-terminated substrings, right? That's actually hassle and another source for bugs (e.g. if you pass in std::strings as string views to some function, things will work fine when just requesting `data()` then passing that to C-style APIs... until you start passing in _real_ substrings, not full strings, then things will start breaking).

johnnovak: For the record, I did not say to stop using lambdas. I use them all the time, I too like to keep one-off helpers within the scope (unless when they grow too big).

> I'm not enjoying contributing anymore, since it's clear what I want out of it, and what the maintainers want are at odds.

That's a personal choice and entirely up to you, but I agree that reviewing relatively simple changes like this should not feel like a constant fight. Ultimately, contributors to any project are kind of expected to adapt/adjust to the style of the project. Improvements over time can be made, of course, but usually over time.

kcgen:
(overcome by events in PR #2863)

src/shell/shell_cmds.cpp
kcgen: @MeAreJeenius , I'm not sure if you've been following and reading all the other team's PRs that have been going into the code base. 

As an example, you'll see that we never use `type var{};` initializers, so we were off to a rough start.
 - One one hand, the C++ committee claimed this is the more efficient initialzer, so please migrate from ` = value` :facepalm: 
 - On the other hand, compilers are free to eliminate unnecessary copies: so we see counter-arguments recommending readability of `= value` and still obtaining efficiency under the hood.

I'm bringing that up more as the battles that projects face with the langauge itself, and how the C++ committee shoots itself in the foot by making these constructs that introduce divisiveness in projects, and how we're all a victim to it here. 

So yeah: here were are claiming to using modern languages and practices.. and yet we're not drinking _all_ of the C++17 and soon-to-be C++20 coolaid.

I'm sure we're in for more syntax A vs. syntax B vs. syntax C battles.. and yeah, if we don't keep pace, maybe someone will fork Staging itself _"A fork that welcomes all C++20 (and beyond) syntax to the maximum." 

So we need to keep pace.

I feel like this last PR, he really tested the waters by using that lame var{}; initializer nonsense..  he mentioned how he had previously decided it was going to be his last PR, so I feel like this was kind of deliberate.

I still feel like there's no problem with using `std::move` and `string_view` within your PR: however, it simply needs to be justified with some comments, like "Move the item from this vector to that vector" -> and it would be self-documenting, and to address the comment that this is more about object ownership and not about performance (to address @johnnovak's concern).

There's also the big picture: this might be the only C++ implementation of DOS `FOR` handling out there.  That old patch on vogons was using lots of C string manipulations, so given the balance: I would prefer yours straight away.

Likewise: dig up FreeDOS's `FOR` implementation (which I can't find), but I guarantee isn't highly non-trivial and very low level.  Similarly if you have access to the original MS DOS implementation: that's going to be some extremely involved code. 

So yeah: in the end, `FOR` is an addon that doesn't tough the normal flow, so it's an incremental improvement we want.

weirddan455: To be fair, just about every project has rules about which parts of C++ to use because the language is huge and some features were designed by the drunk and/or mentally deranged.  "C++ to the max" sounds like "every little thing has to be a templated class that fits into our system-wide multiple inheritance hierarchy and if we ever catch you with a raw pointer we ban you from the project".

You kind of have to pick some subset of C++ in order to stay sane and I've seen it range from "use almost exclusively C style code" to "classes are fine but don't use the STL" to "use modern features but stay away from the crazy bits" to "go nuts and good luck figuring this shit out" and anywhere in between.

Which type of initializer to use is almost strictly stylistic btw.  I agree it's important to stay consistent though or code gets hard to read.  Same with indentation style, snake_case vs MicrosoftCase, etc.  I don't care that much what the style is for a given project, so long as it's consistent.

>  On the other hand, compilers are free to eliminate unnecessary temporaries: so we see counter-arguments recommending readability of = value and still obtaining equivalent efficiency under the hood.

Prior to C++ 17, this is true.  Now, compilers are _required_ to eliminate the temporary so like I said strictly stylistic (and also annoying that there's two forms of syntax that do exactly the same thing).

https://en.cppreference.com/w/cpp/language/copy_elision

>  Since C++17, a prvalue is not materialized until needed, and then it is constructed directly into the storage of its final destination. This sometimes means that even when the language syntax visually suggests a copy/move (e.g. [copy initialization](https://en.cppreference.com/w/cpp/language/copy_initialization)), no copy/move is performed — which means the type need not have an accessible copy/move constructor at all.

johnnovak: I like STL containers, they do the job. We also started using `optional` which is great, and I'm also experimenting with `variant` and `visit` in my upcoming DOS command PR as we speak. Overall, I like these two features more than I dislike they slight verboseness because they simplify the code (I think) and they could establish a common pattern for writing unit testable DOS programs that have lots of command line switches.

But I will ask the opinion of the team about their use: if everybody hates them, I will try to find an alternative solution. Ironically, I've also started experimenting with the inline creation of structs with braces _sparingly_, only where it increases readability and cuts down on the temporaries and the number of lines.

I'm just saying this because it's not like I'm completely against the usage of _any_ new C++ features, but I definitely want to do it sparingly, with approval from the team, and considering every single new feature individually and carefully.

I can be very easy to work with, or very annoying to work with (I guess) because I have to be convinced about the overarching, global benefits of introducing something new before I'm happy with it. That includes readability, maintainability, comfort level of the team, my own comfort level, is this feature _fun_ to use, performance, how will this increase or decrease the approachability of the project for others, and the list goes on. If overall I think the sum of all these parts is not positive, I don't care if the thing in question is the most trendy thing to do, or the "best practice", or was blessed by committe X, or is the favourite pet feature of the person I'm talking with, or brings 10-20% benefit in a few items on the list of considerations, but has otherwise a net negative effect.

This FOR implementation is surely an improvement over existing alternative implementations @MeAreJeenius. If you would still like to finish the job, I think the next highest value thing would be to put a comprehensive unit test suite around it. We can keep the string views and the move, or remove them later; all in all, and big picture again, I care a lot more whether the feature is functionally sound and compatible with DOS 6.22. So if you would do that next, that would be much appreciated and.

johnnovak:
Closing due to inactivity.

johnnovak:
@MeAreJeenius I'm on vacation (kinda, as you can see), will be back in a few weeks. Sure, please feel free to raise a new PR for these when you're ready.

Issue #2861, opened 2023-09-12
Crash on development build (v0.81.0-alpha-1668-g4c3db) when core is switched to dynamic


mckellenjaundice122:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

v0.81.0-alpha-1668-g4c3db

### What Operating System are you using?

Windows 10

### Relevant hardware info

i5-7200U @ 2.50 GHz, 8 GB Ram, Intel HD 630

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When dosbox switches the core either automatically or manually to dynamic, dosbox crashes. The log itself doesn't seem to point to much, other than it only seems to happen in the latest development build. Nothing at all in the config was changed besides core. This happens in both the MSYS2 and MSVC builds.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Launch either a protected mode game or switch core to dynamic via command line or in the config file

### Provide a Log

```text
https://gist.github.com/mckellenjaundice122/9400531e294b88048678c31c8f1199a8
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Kappa971:
I confirm that setting `core = dynamic` DOSBox crashes. Build 0.81.0-alpha-1659

kcgen:
![2319](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/33febcd4-f30e-4c26-9ff3-4fb33a9e14b1)

kcgen:
@Kappa971 and @mckellenjaundice122, and anyone else on Win32/64: Can you try to isolate which PR caused this regression?

Here are the successful `main` builds starting with the current (which we know is bad), going back in time one PR at a time:

https://github.com/dosbox-staging/dosbox-staging/actions/workflows/windows-msys2.yml?query=branch%3Amain+event%3Apush+is%3Asuccess

Click back a couple pages in the list (say, 2 weeks ago) and find one that works.  If it works OK, then pick another build half-way in between.

Keep picking builds half-way in between your bad one and your good one to cut the list in half each time.

It should only take about 7 attempts to isolate it down (if using this half-way approach)

weirddan455:
Or you can use `git bisect` to automate that process if you're building from source.

Kappa971:
[...]

PR #2862, opened 2023-09-12
Revert dynamic core changes

johnnovak:
This reverts @classilla's dynamic core related changes as it wasn't sufficiently regression tested and it turns out it breaks the dynamic core on Windows (at least).

We're not against these improvements, but any changes related to such fundamental parts of DOSBox should be thoroughly regression tested before merging on all supported platforms, so macOS, Windows, and Linux.

See related discussion:
https://github.com/dosbox-staging/dosbox-staging/issues/2861

### DIFF
--- a/include/support.h
+++ b/include/support.h
@@ -52,8 +52,7 @@
 #endif
 
 #ifdef PAGESIZE
-// Some platforms like ppc64 have page sizes of 64K, so uint16_t isn't enough.
-constexpr uint32_t host_pagesize = { PAGESIZE };
+constexpr uint16_t host_pagesize = { PAGESIZE };
 #else
 constexpr uint16_t host_pagesize = 4096;
 #endif
--- a/meson.build
+++ b/meson.build
@@ -277,9 +277,6 @@ conf_data.set10('C_TRACY', get_option('tracy'))
 conf_data.set10('C_FPU', true)
 conf_data.set10('C_FPU_X86', host_machine.cpu_family() in ['x86', 'x86_64'])
 
-# Default page size is 4K, but not on all platforms.
-pagesize = 4096
-
 if get_option('enable_debugger') != 'none'
     conf_data.set10('C_DEBUG', true)
 endif
@@ -384,24 +381,12 @@ if host_machine.endian() == 'big'
     conf_data.set10('WORDS_BIGENDIAN', true)
 endif
 
-# 64-bit Power ISA can run with either 4K or 64K page size.
-# (32-bit Power ISA and PowerPC are always 4K.)
-if host_machine.cpu_family() in ['ppc64', 'ppc64le']
-    pagesize_cmd = run_command('getconf', 'PAGESIZE', check: true)
-    if pagesize_cmd.returncode() != 0
-        error('''error executing getconf: unable to determine host architecture page size for ppc64 dynamic core''')
-    else
-        pagesize = pagesize_cmd.stdout().strip().to_int()
-        if pagesize < 4096
-            # unexpected; did getconf return an empty or bogus string?
-            error('''need at least 4096-byte pages, getconf PAGESIZE returned '''+pagesize_cmd.stdout())
-        else
-            if pagesize != 4096
-                conf_data.set('PAGESIZE', pagesize)
-            endif
-        endif
-    endif
-endif
+# Non-4K memory page size is supported only for ppc64 at the moment.
+# TODO re-enable ppc dynrec while working on W^X stuff
+#      disabled because SVN r4424 broke compilation of ppc backends
+#if host_machine.cpu_family() in ['ppc64', 'ppc64le']
+#  conf_data.set('PAGESIZE', 65536)
+#endif
 
 set_prio_code = '''
 #include <sys/resource.h>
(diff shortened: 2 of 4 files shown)


weirddan455:
I'm gonna reboot to Windows and test this real quick.  It looks like 90% of this PR only affects the PPC build so it should be fairly straightforward to find the root cause of this.  I have not seen any crashes on Linux.

johnnovak:
> I'm gonna reboot to Windows and test this real quick. It looks like 90% of this PR only affects the PPC build so it should be fairly straightforward to find the root cause of this. I have not seen any crashes on Linux.

Yeah sure, but if it drags on, people will download broken dev builds. Windows users are 90% of our users, and that's conservative. So sure, have a look, but I think we should revert first, then regression test and/or fix in a non-stressful manner over the next hours, days, whatever. Then re-add when it's ready.

weirddan455:
Yes, our Windows builds shouldn't be crashing.  I will also say that probably 90% of our users aren't running dev builds though 😉

kcgen:
We can avoid doing further merges to main until this is fixed, one way or another (quick fix or revert).  

The most important thing is keeping the swath start-to-end commits to a minimum, and we're all on board with that. 

It's possible @weirddan455 will get it fixed or @classilla will in the next 24 hours or so. 

My guess is perhaps setting the default 4k pagesize has slipped through the cracks (and maybe we need some `static_assert` to check if it's 4k, 16k, or 64k just as a sanity check). 

Thanks for putting on the 2319 Hazmat suit, @weirddan455 !

kcgen:
> I'm a bit confused, so you're agreeing with me merging this rollback right now? 

Not yet, however, we need to make sure the next commit to main fixes the regression (one way or another), and we don't let other commits land in main till it's fixed. 

Best option: @weirddan455 solves it, if not, maybe @classilla solves it (say, in the next 24hrs), but failing those (if there's going to too much time or maybe @classilla doesn't even have a Windows system), then we revert.

In all paths, the repo's next commit will get rid of the regression. We just have to endure the period in between 😅

weirddan455:
I finally got it compiled in Visual Studio.  Made the mistake of doing a clean `git clone` and rebuild and those vcpkg deps take forever to compile.

Anyway, it's a null pointer de-reference that I can easily reproduce.  Looking for a fix now.

johnnovak:
> Best option: @weirddan455 solves it, if not, maybe @classilla solves it (say, in the next 24hrs), but failing those (if there's going to be too much time or maybe @classilla doesn't even have a Windows system), then we merge your revert.
> 
> In all paths, the repo's next `main` commit will get rid of the regression. We just have to endure the period in between 😅

Sounds good 👍🏻 I just have my default work mindset where even 1 hour of outage would be unacceptable, but yeah, we can afford to wait 24 hours here. I guess old habits die hard ("Error in the logs!!! Hit that rollback button like crazy"—then have a 2-hour retrospective meeting about the 10-minute outage 😅)

weirddan455:
@johnnovak lol it's not like we're running a live service here.  That commit has been in the tree for over and week and no one has noticed until now 😆 

I think the problem is just an `endif` got moved to the wrong place.  Should be a one line fix I'll post shortly.

kcgen:
>  2-hour retrospective meeting about the 10-minute outage 

😂 

Like Duke says, it's time for @weirddan455 to kick ass and attend the retrospective all-hands meeting, and @weirddan455 is not in attendance.

johnnovak:
> @johnnovak lol it's not like we're running a live service here. That commit has been in the tree for over and week and no one has noticed until now 😆

Yeah, I know 😅 It's just that's my mindset, it's hard for me to break from that. Similarly, I find it hard to "just hack shit together"; I find it exceedingly hard to _stop_ being methodical, thorough, etc.

> I think the problem is just an `endif` got moved to the wrong place. Should be a one line fix I'll post shortly.

Thank you very much for looking into this @weirddan455 , much appreciated 🎉

weirddan455:
Fix is up #2863

>  Like Duke says, it's time for @weirddan455 to kick ass and attend the retrospective all-hands meeting, and @weirddan455 is not in attendance.

It was broke.  Now it's not.  We can all go home and this really should have been an email 😆

johnnovak:
> I like to think of the projects' fuel in the tank being the maintenance teams' spirits (and not dollars in the bank/happy customers), so if we can afford some leeway (i.e., a window of time, in this case) it helps keeps the tank full.

I 100% agree with that. 💯 

I think the lesson for us all is to be a bit more thorough with testing fundamental changes. I think the dynarec PR was merged in a bit too hastily, that's all 😄 We're not bad with keeping `main` fairly functional, but I personally would like to put more emphasis on testing on all platforms before merging (and I started doing so with my own PRs).

kcgen:
> I think the lesson for us all is to be a bit more thorough with testing fundamental changes. I think the dynarec PR was merged in a bit too hastily, that's all

💯 agreed there too; even though I tested dynrec and normal cores on Linux and MacOS, should have checked Windows too!

johnnovak:
> > I think the lesson for us all is to be a bit more thorough with testing fundamental changes. I think the dynarec PR was merged in a bit too hastily, that's all
> 
> 💯 agreed there too; even though I tested dynrec and normal cores on Linux and MacOS, should have checked Windows too!

I admit, sometimes I don't always test on Windows when it's something that couldn't reasonably break on different platforms. But the auto-switching shader stuff surprised me; I got hard crashes on Windows because the SDL events are not 100% the same across platforms, so they exercised the code a bit differently (things got initialised in a different order on Windows than on macOS). It was a one-liner fix to support "lazy initialisation" in one spot so it works on Windows too, but yeah, that's one such example.

Grounded0:
Close:

- #2863

PR #2863, opened 2023-09-13
Fix null pointer de-reference in dyncache on Windows

### LINKED ISSUE #2861: Crash on development build (v0.81.0-alpha-1668-g4c3db) when core is switched to dynamic

weirddan455:
# Description
Reverts part of commit 765bcc2

This is the only part of that commit that should affect non PPC processors.  The `#endif` got moved so that the Windows build never sets the `cache_code` variable leading to a null pointer de-reference.

I also removed the "aligned by definition" on mmap part.  I don't see the purpose of this change.  I believe VirtualAlloc on Windows which we're using a couple lines above is also "aligned by definition" to page size boundary.

In any case, if the memory is already aligned, the alignment check doesn't do anything (you still get the same address back out as far as I can tell).  Doesn't make sense as a micro-optimization either as the heap allocation is far more expensive than an add and a couple bitwise instructions.

Better to just set this back to the way it was so this variable gets set in the same way in all paths IMO.

Sidenote:  This is also why I dislike lots of inline `#ifdefs`.  Makes it very easy to introduce bugs like this.  I'd prefer to see it separated out into an `allocate_cache` function or something that has multiple implementations.  But this is legacy code and I don't feel like touching it 😄 

## Related issues
Fixes #2861

# Manual testing
I tested in MSVC and doesn't crash anymore.

# Checklist

I have:

### DIFF
--- a/src/cpu/dyn_cache.h
+++ b/src/cpu/dyn_cache.h
@@ -939,18 +939,16 @@ static void cache_init(bool enable) {
 			if (cache_code_start_ptr == MAP_FAILED) {
 				E_Exit("DYNCACHE: Failed memory-mapping cache memory because: %s", strerror(errno));
 			}
-			// aligned by definition
-			cache_code = reinterpret_cast<uint8_t *>(cache_code_start_ptr);
 #else
 			cache_code_start_ptr=static_cast<uint8_t *>(malloc(cache_code_size));
 			if (!cache_code_start_ptr) {
 				E_Exit("DYNCACHE: Failed allocating cache memory because: %s", strerror(errno));
 			}
+#endif
 			// align the cache at a page boundary
 			cache_code = reinterpret_cast<uint8_t *>(
 			    (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
 			    host_pagesize - 1) & ~(host_pagesize - 1));
-#endif
 
 			cache_code_link_blocks=cache_code;
 			cache_code=cache_code+host_pagesize;
--- a/src/cpu/dyn_cache.h
+++ b/src/cpu/dyn_cache.h
@@ -948,7 +948,7 @@ static void cache_init(bool enable) {
 			// align the cache at a page boundary
 			cache_code = reinterpret_cast<uint8_t *>(
 			    (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
-			    host_pagesize - 1) & ~(host_pagesize - 1));
+			    static_cast<size_t>(host_pagesize) - 1) & ~(static_cast<size_t>(host_pagesize) - 1));
 
 			cache_code_link_blocks=cache_code;
 			cache_code=cache_code+host_pagesize;


kcgen:
@Kappa971 , can you check the MSYS2 build generated from this CI run?

kcgen (approved):

johnnovak:
I'll test on macOS shortly. Once we've all tested the builds on all platforms with some protected mode programs, I think this is safe to merge.

johnnovak:
And great job @weirddan455, super quick turnaround too! 🎉

classilla:
I'll have to find another fix for PPC, because this breaks us again.

classilla:
For the record, the PPC build asserts with reverting that.

```
dosbox: ../../src/cpu/dyn_cache.h:840: void dyn_mem_set_access(void*, size_t, bool): Assertion `mp_res == 0' failed.

Thread 1 "dosbox" received signal SIGABRT, Aborted.
0x00007ffff6eba7fc in __pthread_kill_implementation () from /lib64/libc.so.6
```

The math is wrong for how `cache_code` is computed, which is why I made the change in the first place. If you get a pointer for `cache_code_start_ptr` that has the top 64 bits populated, it will get stripped when computing `cache_code`. Proof:

```
(gdb) bt
#0  0x00007ffff6eba7fc in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff6e57d9c in raise () at /lib64/libc.so.6
#2  0x00007ffff6e30980 in abort () at /lib64/libc.so.6
#3  0x00007ffff6e4c220 in __assert_fail_base () at /lib64/libc.so.6
#4  0x00007ffff6e4c2c4 in __assert_fail () at /lib64/libc.so.6
#5  0x000000001080c0cc in dyn_mem_set_access(void*, size_t, bool)
    (ptr=0xb1fb0000, size=65536, execute=false)
    at ../../src/cpu/dyn_cache.h:840
#6  0x000000001080c180 in dyn_mem_write(void*, size_t)
    (ptr=0xb1fb0000, size=8192) at ../../src/cpu/dyn_cache.h:868
#7  0x000000001080c5e8 in cache_init(bool) (enable=true)
    at ../../src/cpu/dyn_cache.h:966
(gdb) p cache_code_start_ptr
$4 = (uint8_t *) 0x7fffb1fa0000 ""
(gdb) p cache_code
$2 = (uint8_t *) 0xb1fb0000 <error: Cannot access memory at address 0xb1fb0000>
```

This is unmapped memory. If we get a page from `mprotect` it's aligned by definition, which is why I made the comment - it *must* use the system page size - so there was no need to munge the pointer. However, since Windows seems to rely on that behaviour, I will just make a PPC specific change.

But I want to point out for the record that the code has a demonstrable bug.

johnnovak:
Thanks for the info @classilla. Hopefully, someone will fix that one day. But then, even though theoretically buggy, the dynarec stuff has been working just fine on Windows. Potentially buggy vs outright broken is a pretty big difference 😄

weirddan455:
>  However, since Windows seems to rely on that behaviour, I will just make a PPC specific change.

Well, I don't know if that is true.  I could test.  The problem was that by moving the `#endif`, Windows did not get the pointer set _at all_ (it got left null and crashed).

@classilla Just out of curiosity, do you know what is wrong with that code exactly?  It looks correct to me, assuming the page size is always a power of 2.

```
$ python
Python 3.11.5 (main, Sep  2 2023, 14:16:33) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> page_size = 65536
>>> pointer  = 0x7fffb1fa0000
>>> aligned_pointer = pointer + (page_size - 1) & ~(page_size - 1)
>>> print(aligned_pointer)
140736179339264
>>> print(pointer)
140736179339264
```

From Microsoft's documentation of `VirtualAlloc`:

>  The starting address of the region to allocate. If the memory is being reserved, the specified address is rounded down to the nearest multiple of the allocation granularity. If the memory is already reserved and is being committed, the address is rounded down to the next page boundary. To determine the size of a page and the allocation granularity on the host computer, use the [GetSystemInfo](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsysteminfo) function. If this parameter is NULL, the system determines where to allocate the region.

We're doing commit and reserve in one step from the looks of it so... idk.

classilla:
Assuming your patch, plus some debugging `printf`s,

```
diff --git a/src/cpu/dyn_cache.h b/src/cpu/dyn_cache.h
index cfb82ae33..309012487 100644
--- a/src/cpu/dyn_cache.h
+++ b/src/cpu/dyn_cache.h
@@ -934,28 +934,32 @@ static void cache_init(bool enable) {
                        int prot_flags = PROT_READ | PROT_WRITE | PROT_EXEC;
 #if defined(HAVE_MAP_JIT)
                        map_flags |= MAP_JIT;
 #endif
                        cache_code_start_ptr=static_cast<uint8_t *>(mmap(nullptr, cache_code_size, prot_flags, map_flags, -1, 0));
                        if (cache_code_start_ptr == MAP_FAILED) {
                                E_Exit("DYNCACHE: Failed memory-mapping cache memory because: %s", strerror(errno));
                        }
-                       // aligned by definition
-                       cache_code = reinterpret_cast<uint8_t *>(cache_code_start_ptr);
 #else
                        cache_code_start_ptr=static_cast<uint8_t *>(malloc(cache_code_size));
                        if (!cache_code_start_ptr) {
                                E_Exit("DYNCACHE: Failed allocating cache memory because: %s", strerror(errno));
                        }
+#endif
                        // align the cache at a page boundary
+fprintf(stderr, "%0lx\n", reinterpret_cast<uintptr_t>(cache_code_start_ptr));
+fprintf(stderr, "%0lx\n", (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
+                            host_pagesize - 1));
+fprintf(stderr, "%0lx\n",  (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
+                            host_pagesize - 1) & ~(host_pagesize - 1));
                        cache_code = reinterpret_cast<uint8_t *>(
                            (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
                            host_pagesize - 1) & ~(host_pagesize - 1));
-#endif
+fprintf(stderr, "%0lx\n", cache_code);
 
                        cache_code_link_blocks=cache_code;
                        cache_code=cache_code+host_pagesize;
                        CacheBlock *block = cache_getblock();
                        cache.block.first=block;
                        cache.block.active=block;
                        block->cache.start=&cache_code[0];
                        block->cache.size=CACHE_TOTAL;
```

I get this:

```
2023-09-12 22:26:33.704 | DOSBox has switched to max cycles, because of the setting: cycles=auto.
If the game runs too fast, try a fixed cycles amount in DOSBox's options.
7fffb1790000
7fffb179ffff
b1790000
b1790000
dosbox: ../../src/cpu/dyn_cache.h:840: void dyn_mem_set_access(void*, size_t, bool): Assertion `mp_res == 0' failed.
```

It seems to be at the point it tacks on the `& ~(host_pagesize - 1)`. Python is probably handling it with its own arithmetic, so I'd be interested to see what you get out of your C compiler. This is vanilla `gcc`/`g++`.

Also, from your description it sounds like Windows `VirtualAlloc` is not an exact substitute for `mmap`. But it sounds like it wouldn't need this code either if it automatically aligns to the integral page.

weirddan455:
Something very strange is happening on your machine @classilla 

I ran it in actual Dosbox code on my machine in C++

```cpp
			uint8_t* test = reinterpret_cast<uint8_t*>(0x7fffb1790000);
			fprintf(stderr, "%0lx\n", reinterpret_cast<uintptr_t>(test));
			fprintf(stderr, "%0lx\n", (reinterpret_cast<uintptr_t>(test) +
                            host_pagesize - 1));
			fprintf(stderr, "%0lx\n",  (reinterpret_cast<uintptr_t>(test) +
                            host_pagesize - 1) & ~(host_pagesize - 1));

			fprintf(stderr, "%0lx\n", reinterpret_cast<uintptr_t>(test));
			fprintf(stderr, "%0lx\n", (reinterpret_cast<uintptr_t>(test) +
                            65536 - 1));
			fprintf(stderr, "%0lx\n",  (reinterpret_cast<uintptr_t>(test) +
                            65536 - 1) & ~(65536 - 1));


			fprintf(stderr, "%0lx\n", reinterpret_cast<uintptr_t>(cache_code_start_ptr));
			fprintf(stderr, "%0lx\n", (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
                            host_pagesize - 1));
			fprintf(stderr, "%0lx\n",  (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
                            host_pagesize - 1) & ~(host_pagesize - 1));
			cache_code = reinterpret_cast<uint8_t *>(
			    (reinterpret_cast<uintptr_t>(cache_code_start_ptr) +
			    host_pagesize - 1) & ~(host_pagesize - 1));
```

Top test is your hardcoded address.  Middle test is your hardcoded address with a hardcoded 65536 page size.  Bottom test is the real address mmap gave me on x86_64 Linux.  Results:

```
7fffb1790000
7fffb1790fff
7fffb1790000

7fffb1790000
7fffb179ffff
7fffb1790000

7f119c5f1000
7f119c5f1fff
7f119c5f1000
```

weirddan455:
It looks like we're dealing with some weird integer promotion (or demotion?) that only affected PowerPC for some reason.  I don't know how it would ever try and promote a pointer which should be an unsigned 64 bit type.

Is `host_pagesize` a signed type on your machine?  If the pointer gets converted to 32 bit while doing arithimitic for whatever reason, that would explain what we're seeing (the top 32 bits are being removed).

classilla:
Thanks. I'm going to chalk this up to a compiler bug, then. Here's the assembly, fwiw.

```
[Switching to thread 1 (Thread 0x7ffff5a961c0 (LWP 3734914))]
#0  cache_init (enable=true) at ../../src/cpu/dyn_cache.h:954
954     __asm__("trap\n");
Dump of assembler code from 0x1080c59c to 0x1080c5ec:
   0x000000001080c59c <cache_init(bool)+628>:   addi    r4,r4,8152
   0x000000001080c5a0 <cache_init(bool)+632>:   mr      r3,r10
   0x000000001080c5a4 <cache_init(bool)+636>:   bl      0x10623220 <0000001b.plt_call.__fprintfieee128@@GLIBC_2.32>
   0x000000001080c5a8 <cache_init(bool)+640>:   ld      r2,24(r1)
=> 0x000000001080c5ac <cache_init(bool)+644>:   trap
   0x000000001080c5b0 <cache_init(bool)+648>:   addis   r9,r2,355
   0x000000001080c5b4 <cache_init(bool)+652>:   addi    r9,r9,-31088
   0x000000001080c5b8 <cache_init(bool)+656>:   ld      r9,0(r9)
   0x000000001080c5bc <cache_init(bool)+660>:   addis   r9,r9,1
   0x000000001080c5c0 <cache_init(bool)+664>:   addi    r9,r9,-1
   0x000000001080c5c4 <cache_init(bool)+668>:   clrrwi  r9,r9,16
   0x000000001080c5c8 <cache_init(bool)+672>:   mr      r10,r9
   0x000000001080c5cc <cache_init(bool)+676>:   addis   r9,r2,355
   0x000000001080c5d0 <cache_init(bool)+680>:   addi    r9,r9,-31080
   0x000000001080c5d4 <cache_init(bool)+684>:   std     r10,0(r9)
   0x000000001080c5d8 <cache_init(bool)+688>:   trap
   0x000000001080c5dc <cache_init(bool)+692>:   ld      r9,-31736(r2)
   0x000000001080c5e0 <cache_init(bool)+696>:   ld      r10,0(r9)
   0x000000001080c5e4 <cache_init(bool)+700>:   addis   r9,r2,355
   0x000000001080c5e8 <cache_init(bool)+704>:   addi    r9,r9,-31080
End of assembler dump.
(gdb) set $pc+=4
(gdb) disp/x $r9
1: /x $r9 = 0x0
(gdb) disp/i $pc
2: x/i $pc
=> 0x1080c5b0 <cache_init(bool)+648>:   addis   r9,r2,355
(gdb) si
0x000000001080c5b4      957                                 host_pagesize - 1) & ~(host_pagesize - 1));
1: /x $r9 = 0x12947b00
2: x/i $pc
=> 0x1080c5b4 <cache_init(bool)+652>:   addi    r9,r9,-31088
(gdb) 
0x000000001080c5b8      957                                 host_pagesize - 1) & ~(host_pagesize - 1));
1: /x $r9 = 0x12940190
2: x/i $pc
=> 0x1080c5b8 <cache_init(bool)+656>:   ld      r9,0(r9)
(gdb) 
0x000000001080c5bc      957                                 host_pagesize - 1) & ~(host_pagesize - 1));
1: /x $r9 = 0x7fffb5790000
2: x/i $pc
=> 0x1080c5bc <cache_init(bool)+660>:   addis   r9,r9,1
(gdb) 
0x000000001080c5c0      957                                 host_pagesize - 1) & ~(host_pagesize - 1));
1: /x $r9 = 0x7fffb57a0000
2: x/i $pc
=> 0x1080c5c0 <cache_init(bool)+664>:   addi    r9,r9,-1
(gdb) 
957                                 host_pagesize - 1) & ~(host_pagesize - 1));
1: /x $r9 = 0x7fffb579ffff
2: x/i $pc
=> 0x1080c5c4 <cache_init(bool)+668>:   clrrwi  r9,r9,16
(gdb) 
955                             cache_code = reinterpret_cast<uint8_t *>(
1: /x $r9 = 0xb5790000
2: x/i $pc
=> 0x1080c5c8 <cache_init(bool)+672>:   mr      r10,r9
(gdb) 
```

`gcc` should have used a `clrrdi` instead. Still, even if this is a compiler bug, we would still need something like this as a bandaid since others would have the same compiler and issue. `gcc` 13.2.1 is the most current version in the Fedora 38 repository for `ppc64le`.

classilla:
Alternatively:

`host_pagesize` is indeed `uint32_t` (it was `uint16_t` before, but needed to be bigger for 64K pages). If I make it a `uint64_t` then this code works. I still read it as a compiler bug, but the only platform this would even possibly affect would be Apple silicon where the page size is 16K (and it should work fine there). I can make that PR separately if people are willing.

```
diff --git a/include/support.h b/include/support.h
index aa70480e4..b250ebba6 100644
--- a/include/support.h
+++ b/include/support.h
@@ -48,17 +48,17 @@
 
 #ifdef _MSC_VER
 #define strcasecmp(a, b) _stricmp(a, b)
 #define strncasecmp(a, b, n) _strnicmp(a, b, n)
 #endif
 
 #ifdef PAGESIZE
 // Some platforms like ppc64 have page sizes of 64K, so uint16_t isn't enough.
-constexpr uint32_t host_pagesize = { PAGESIZE };
+constexpr uint64_t host_pagesize = { PAGESIZE };
 #else
 constexpr uint16_t host_pagesize = 4096;
 #endif
 
 constexpr uint16_t dos_pagesize = 4096;
 
 // Some C functions operate on characters but return integers,
 // such as 'toupper'. This function asserts that a given int
```

classilla:
That works!

johnnovak:
> I hate dealing with C's often very unintuitive integer promotion.

My current no.1 thing I hate about C. The second prize goes to the include file "system".

If you understand C integer promotion, you can easily see why those asm coders on 16/32-bit machines back in the day saw C as a complete joke (especially Amiga asm coders had very low opinions of C in general).

weirddan455:
> That works!

Cool.  Pushed a commit to this PR to that.  I think it's a safe change.  Just a couple `static_cast`'s to tell C to not mangle our integers :roll_eyes:

classilla:
Thanks for working with me on a all-around solution. Appreciated.

Kappa971:
> @Kappa971 , can you check the MSYS2 build generated from this CI run?

I tried this https://github.com/dosbox-staging/dosbox-staging/actions/runs/6166531743 and DOSBox no longer crashes.

johnnovak:
Regression tested on macOS again for good measure and all seems fine. Thanks again @weirddan455 @classilla @kcgen  and @Kappa971 🚀

johnnovak (approved):

Grounded0:
Runs fine on macOS 12.6.5/x86-64.

Issue #2865, opened 2023-09-13
Auto integer scaling affects texture output


GranMinigun:
1) Launch DOSBox Staging with default configuration;
2) Switch `output` to `texture`;
3) Resize the window.

It appears that automatic integer scaling retains whatever it calculated for the last used shader instead of resetting for the new output. @johnnovak, as you've been tinkering around that area lately, think you could have a look?

kcgen:
> It appears that automatic integer scaling retains whatever it calculated for the last used shader 

Additionally, it looks like even when in texture-mode, shaders are still active (so many some state isn't being passed through?)

![2023-09-13_11-27](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/d59efec9-fc73-4d87-8ebe-1c61cd3d243e)

johnnovak:
I'll have a look. I've done some testing with `texture` output now and then, but 90% of the time I'm testing with `opengl`. In any case, this is easy to fix, thanks for raising it @GranMinigun.

KainXVIII:
For education purposes - how does texture output option differs from other like surface or opengl?

johnnovak:
> For education purposes - how does texture output option differs from other like surface or opengl?

Surface has been removed; we only have the texture and OpenGL modes now.

Texture is basically a fallback mode we get cheaply (almost free) from SDL. It's only meant for people who don't have a somewhat functional OpenGL card 😄 

The huge difference is that `texture` mode does not support shaders, only `opengl` does. There's no reason to use texture mode, unless you're on some really underpowered or old hardware. It's just a fallback option.

KainXVIII:
Thanks for the enlightenment!

PR #2866, opened 2023-09-14
Adaptive shader fixes

johnnovak:
# Description

This fixes two adaptive CRT shader related bugs:

- The adaptive shader-switching mechanism was erroneously also activated in non-OpenGL output modes when `glshader` was set to an adaptive shader (`crt-auto`, `crt-auto-machine`, or `crt-auto-arcade`).
- The `integer_scaling` setting in `auto` mode wasn't restricted to OpenGL output modes only. `auto` mode should effectively work as `off` (no integers scaling) in `texture` mode as no adaptive CRT shader can be active in `texture` mode.

@GranMinigun I've also done some sdlmain cleanup as I needed to extract the badly named `SCREEN_TYPES` enum, so I took the chance to refactor it while I was at it. Let me know if this conflicts too much with your sdlmain rework, then I can revert that part.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2865

Note that starting DOSBox with `output = texture` set in the config, then switching to `opengl` at runtime does _not_ work. That's broken in main too. I'll need @GranMinigun's help on this because I'm not too sure what's going on. However, if we never allowed rendering backend changes at runtime, I don't think that would be a great loss, so I don't think this is a high priority issue (but, yeah, technically buggy currently).

# Manual testing

### Test 1

1. Start Staging with `output = texture` in the config and `--noprimaryconf`. No integer scaling should be applied; the image should fill the window.
2. Resize the window to its minimum size. Then start resizing it in multiple steps until the window occupies the whole screen. You should not see any auto-shader switching related messages in the logs.
3. Switch to fullscreen and then back to windowed. You should not see any auto-shader switching related messages in the logs.
4. Enable integer scaling with `integer_scaling = vertical` — should work as expected.
5. Disable integer scaling with `integer_scaling = off` — should work as expected.

### Test 2

1. Start Staging with the defaults and no local config (`--noprimaryconf --nolocalconf`).
2. Note that the auto CRT shader and vertical integer scaling are actived (resize the window a few times to make sure).
3. Switch to `texture` mode by running `output = texture`.
4. Note that integer scaling is disabled (the image fills the window).

### Test 3 (failing)

1. Start Staging with `output = texture` in the config and `glshader = sharp`.
2. Switch to `opengl` output (`output = opengl`).
3. Switching to OpenGL will fail and you'll see the following in the logs:

```
2023-09-14 18:41:31.995 | SDL:OPENGL: No support for texture size of 1024x512, falling back to texture
2023-09-14 18:41:32.010 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz VFR, scaled to 2205x1654 with 1:1.35 (20:27) pixel aspect ratio
2023-09-14 18:41:32.010 | Could not create OpenGL context, switching back to texture
```

@GranMinigun I need your help on this. This also happens in main, but with a slightly different error because the renderer is not even being reinitialised there as in this branch, thereforce the shader source will contain an empty string (so the thing would fail earlier). 

# Checklist

I have:

### DIFF
--- a/include/render.h
+++ b/include/render.h
@@ -152,6 +152,7 @@ struct RenderedImage {
 extern Render_t render;
 extern ScalerLineHandler_t RENDER_DrawLine;
 
+void RENDER_Init(Section*);
 void RENDER_Reinit();
 
 void RENDER_AddConfigSection(const config_ptr_t& conf);
--- a/include/render.h
+++ b/include/render.h
@@ -152,6 +152,8 @@ struct RenderedImage {
 extern Render_t render;
 extern ScalerLineHandler_t RENDER_DrawLine;
 
+void RENDER_Reinit();
+
 void RENDER_AddConfigSection(const config_ptr_t& conf);
 
 bool RENDER_IsAspectRatioCorrectionEnabled();
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -882,6 +882,39 @@ void RENDER_SyncMonochromePaletteSetting(const enum MonochromePalette palette)
 	string_prop->SetValue(to_string(palette));
 }
 
+static bool handle_shader_changes()
+{
+	if (GFX_GetRenderingBackend() != RenderingBackend::OpenGl) {
+		return false;
+	}
+
+	auto& shader_manager = get_shader_manager();
+
+	if (GFX_GetRenderingBackend() == RenderingBackend::OpenGl) {
+		const auto section     = get_render_section();
+		const auto shader_name = shader_manager.MapShaderName(
+		        section->Get_string("glshader"));
+
+		shader_manager.NotifyGlshaderSettingChanged(shader_name);
+
+		const auto string_prop = section->GetStringProp("glshader");
+		string_prop->SetValue(shader_name);
+	}
+	const auto new_shader_name = shader_manager.GetCurrentShaderInfo().name;
+
+	const auto shader_changed = render.force_reload_shader ||
+	                            (new_shader_name != render.current_shader_name);
+
+	if (render.force_reload_shader) {
+		shader_manager.ReloadCurrentShader();
+	}
+
+	render.force_reload_shader = false;
+	render.current_shader_name = new_shader_name;
+
+	return shader_changed;
+}
+
 void RENDER_Init(Section* sec)
 {
 	Section_prop* section = static_cast<Section_prop*>(sec);
@@ -911,33 +944,7 @@ void RENDER_Init(Section* sec)
 
 	GFX_SetIntegerScalingMode(get_integer_scaling_mode_setting());
 
-	auto shader_changed = false;
-
-	if (GFX_GetRenderingBackend() == RenderingBackend::OpenGl) {
-		auto& shader_manager = get_shader_manager();
-
-		if (GFX_GetRenderingBackend() == RenderingBackend::OpenGl) {
-			const auto shader_name = shader_manager.MapShaderName(
-			        section->Get_string("glshader"));
-
-			shader_manager.NotifyGlshaderSettingChanged(shader_name);
-
-			const auto string_prop = section->GetStringProp("glshader");
-			string_prop->SetValue(shader_name);
-		}
-		const auto new_shader_name =
-		        shader_manager.GetCurrentShaderInfo().name;
-
-		shader_changed = render.force_reload_shader ||
-		                 (new_shader_name != render.current_shader_name);
-
-		if (render.force_reload_shader) {
-			shader_manager.ReloadCurrentShader();
-		}
-
-		render.force_reload_shader = false;
-		render.current_shader_name = new_shader_name;
-	}
+	auto shader_changed = handle_shader_changes();
 
 	setup_scan_and_pixel_doubling();
 
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -591,6 +591,10 @@ bool RENDER_MaybeAutoSwitchShader([[maybe_unused]] const uint16_t canvas_width,
                                   [[maybe_unused]] const bool reinit_render)
 {
 #if C_OPENGL
+	if (GFX_GetRenderingBackend() != RenderingBackend::OpenGl) {
+		return false;
+	}
+
 	// Currently, the init sequence is slightly different on Windows, macOS,
 	// and Linux. The first call to this function on Windows receives an
 	// uninitialised VideoMode param with width and height set to 0 which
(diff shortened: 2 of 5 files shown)


GranMinigun:
> Let me know if this conflicts too much with your sdlmain rework, then I can revert that part.

No, that's fine.

> However, if we never allowed rendering backend changes at runtime, I don't think that would be a great loss, so I don't think this is a high priority issue (but, yeah, technically buggy currently).

It was allowed for quite some time (several releases?). Not necessary among the most demanded features, but helps in catching context/renderer management problems.

Unholstering debugger.

src/gui/sdlmain.cpp
```diff
@@ -1355,7 +1355,19 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
...
+#if C_OPENGL
+	const auto rendering_backend_changed = (sdl.rendering_backend !=
+	                                        rendering_backend);
+#endif
+
 	sdl.rendering_backend = rendering_backend;
+
+#if C_OPENGL
+	if (rendering_backend_changed) {
+		RENDER_Reinit();
```
GranMinigun, line 1367: So it trips over here, because deep inside `GFX_SetSize()` is called which expects some values to be already populated (like maximum texture size). But moving it to the end of `set_output()`, when everything is setup and should be mostly ready to go doesn't help, as now it trips inside shader building... because the context gets destroyed along the way.

This place is cursed, I'm telling you. Digging further.

GranMinigun: Scratch that one, I was just overzealously cleaning up resources to right the wrongs. Moving `RENDER_Reinit()` to the end of `set_output()` seems to be enough to make it functional for now.

src/gui/sdlmain.cpp
```diff
@@ -2302,15 +2324,15 @@ bool GFX_StartUpdate(uint8_t * &pixels, int &pitch)
 	if (!sdl.active || sdl.updating)
 		return false;
 
-	switch (sdl.desktop.type) {
-	case SCREEN_TEXTURE:
+	switch (sdl.rendering_backend) {
+		case RenderingBackend::Texture:
```
GranMinigun, line 2328: Indentation.

src/gui/sdlmain.cpp
```diff
@@ -2563,12 +2585,12 @@ static bool present_frame_gl()
 
 uint32_t GFX_GetRGB(const uint8_t red, const uint8_t green, const uint8_t blue)
 {
-	switch (sdl.desktop.type) {
-	case SCREEN_TEXTURE:
+	switch (sdl.rendering_backend) {
+		case RenderingBackend::Texture:
```
GranMinigun, line 2589: Ditto.

src/gui/sdlmain.cpp
johnnovak: Thanks @GranMinigun , I messed around with it for like 15 minutes before going to bed but got frustrated with the sdlmain mess. Glad you found a workable solution and I'm really looking forward to your SDL rework hitting main!

johnnovak:
Addressed comments @GranMinigun & thanks for the fix! Now the opengl/texture switching is restored.

include/render.h
```diff
@@ -152,6 +152,8 @@ struct RenderedImage {
 extern Render_t render;
 extern ScalerLineHandler_t RENDER_DrawLine;
 
+void RENDER_Reinit();
```
GranMinigun, line 156: The function is declared unconditionally, but only implemented when OpenGL is enabled, which trips minimal build. Probably needs more `#if C_OPENGL` sorcery?

johnnovak: Ah yeah, pushed the changes, the forgot to check the CI status later. Definitely needs more ifdefs 😅

johnnovak: Actually... it needed _less_ 😄

src/gui/render.cpp
```diff
@@ -591,6 +591,10 @@ bool RENDER_MaybeAutoSwitchShader([[maybe_unused]] const uint16_t canvas_width,
                                   [[maybe_unused]] const bool reinit_render)
 {
 #if C_OPENGL
+	if (GFX_GetRenderingBackend() != RenderingBackend::OpenGl) {
+		return false;
+	}
+
```
kcgen, line 595: Good fix here!

Looking at the pre-processor span, I think the prior return false can be moved up here now too, and have the two texture sentinels together.



So we can move the check for non-OpenGL usage or availability together:

```c++
#if C_OPENGL
	// We have OpenGL, but are we using it?
	if (GFX_GetRenderingBackend() != RenderingBackend::OpenGl)
# else
	// or we don't have OpenGL at all, so can't use it.
#endif
	// In both scenarios:
	{
		return false;
	}
```

There's an even cleaner alternative to the above that doesn't involve pre-processing stuff:

```c++
	if (GFX_GetRenderingBackend() == RenderingBackend::Texture) {
		return false;
	}
```

Because we only have two backends now.. but the prior approach is more robust if we add more backends.

johnnovak: I agree with your above assessment and ideas @kcgen; when I originally implemented this, I was just carrying forward the ifdefs without giving them much thought, then yesterday I was in "easy in / easy out / surgical fix" mode 😄 It's great when someone else who is less close to the code makes these observations — that's where I see the value of code reviews! 🎉 

That's a good insight about removing the ifdefs from the enum itself, and with that, it turns out I can get rid of _all_ the ifdefs in `render.cpp` and convert them to runtime checks, except for a single one for conditionally compiling in the `glshader` setting. Will give it a good testing then push it as an additional commit soon.

Similar ifdef pruning could be done in `sdlmain.cpp`, but I'll leave that to @GranMinigun.

johnnovak:
@kcgen @GranMinigun  Now it's done done. All compiling and the ifdef cruft has been removed from `render.cpp`.

src/gui/render.cpp
kcgen: > and with that, it turns out I can get rid of all the ifdefs in render.cpp and convert them to runtime checks

Oh wow - I didn't realize it ran that deep! Very nice taking out the werffen flammen :fire:  device to all those preprocessor branches!

kcgen: > I was just carrying forward the ifdefs without giving them much thought, then yesterday I was in "easy in / easy out / surgical fix" mode 😄 

The surgical point fixes are all good too! :man_health_worker: :scissors:... 

But I'm glad you were up for a bigger mission and took out the entire asylum .. turned into a surgical strike mission instead :laughing:

kcgen (approved):
Fix is nice and straight forward; plus :+1: for the  name improvement and de-preprocessing!

ThomasEricB:
I know a certain @exoscoriae is going to be smiling about these changes! Thanks @johnnovak

PR #2868, opened 2023-09-15
MIXER command improvements & general mixer enhancements + fixes

johnnovak:
# Description

This is a significant rework of the command parsing, error reporting, and state handling of the `MIXER` command. The following bug fixes and improvements have been made:

- Mixer channel settings now "survive" deleting then later re-creating the channel instead of simply getting the default settings when the channel is recreated. For example, if you customise the settings of the `SB` and `OPL` channels, then change `sbtype` to something else, the `SB` and `OPL` channels will retain their settings. Similar situation for the `FSYNTH` channel when setting `mididevice` to something else then back to `fluidsynth`.

- The `MIXER` command no longer accepts slightly incorrect input; command parsing is now strict and we're displaying error messages for invalid commands. These messages also try to guide the user to correct their mistakes.

- This _might_ cause backward compatibility issues if people had slightly (or very!) incorrect `MIXER` command invocations in their configs, so we're also logging a warning when the MIXER command fails. That should give people a chance to correct these mistakes. Examples of invalid input that is no longer accepted:
    - `MIXER ASDF R1` – Previously, the `ASDF` param was ignored because there is no channel with that name, so the scope stayed at "global scope" (all channels), then a global reverb setting of 1 was applied to all channels, which is super confusing.
    - `MIXER SB 10 GUS` – This set the `SB` volume to 10; now it fails because every channel name must be followed by one or more commands to remove ambiguity, and there's nothing after `GUS`.
    - `MIXER SB 10 BLAH 4` – This set the `SB` channel's volume to 10 and ignored the non-existing `BLAH` channel; now an error is raised that `BLAH` is an invalid command.
    - `MIXER SB ASDF` – This just printed the mixer status but did not raise any errors; now it does.
    - `MIXER SB R150` – This silently clamped the reverb level to 100 and set it for the `SB` channel; now we raise an error.
    -  `MIXER GUS 50` – If the GUS was not enabled, this just displayed the current mixer settings; now it prints and error saying that the `GUS` channel is not active.
    - There are millions of other cases that were confusing, wrong, or "technically correct" (according to the weird and overly relaxed parsing rules) but plain counterintuitive. Strict parsing and error reporting take care of all that. See the unit tests for further details and test cases.
   
- `MIXER` command handling is now "transactional"—if there is a single error in a list of multiple commands, no command will be executed but an error for the first problem will be raised. This is just common sense and leads to less confusing behaviour overall.

- The `crossfeed` config setting has been revised; now it works the same way as the `chorus` setting (it has `light`, `normal` and `strong` presets, and `on` defaults to `normal`). The previous behaviour where you could set a global crossfeed value (e.g., `crossfeed = 30`) was unsupportable even with a lot of special casing. It was simply impossible to make that behaviour consistent with the `reverb` and `chorus` settings—I literally rewrote it in 4-5 different ways before I settled on this solution. It just can't be done.

## Other bundled changes

The number parsing string util functions have been revised to make them more generic, and the parsing is now strict. Previously, non-digit characters after digits were silently ignored; this now results in a parse error. Strict parsing is desirable as lenient parsing is a fertile breeding ground for subtle bugs and surprising behaviour (if affected the mixer commands, for example).

The flow-on effect is that the argument parsing of the `MORE` command is now stricter: `MORE /t4x fname` worked before (it set the tab size to 4 chars), but now that generates an "illegal switch" error (`/t4` is the correct invocation). Other DOS commands implemented by us might be affected too.

## Technical notes

I started using the `std::visit` and `std::variant` std lib features which are ideally suited for command line parsers and any kind of scenario when dealing with polymorphic command execution. I'm very cautious when adopting new C++ features, but these are so ideally suited to the task that I found them irresistible 😎  They're quite sane and low overhead, both mentally and syntax-wise.

`std::variant` is quite nice even in isolation; it can be used to return an "Either" or "Result" type common in functional languages that encapsulates either a "success" or an "error" result. IMO, this is the preferred alternative to the hacky old C/C++ way of providing "out" pointer arguments to return "secondary" results, such as errors.

A few nice introductory articles about using `std::visit` and `std::variant` together to implement command parsers & executors:

- https://cpp-corner.com/visit-them-all-with-stdvariant/
- https://www.cppstories.com/2020/04/variant-virtual-polymorphism.html/
- https://trussel.ch/cpp/design%20patterns/2020/10/18/command-pattern.html

If everybody hates them, we can discuss finding an alternative way, but it would be a lot more complex and there would be a lot of boilerplate involved, so I'm strongly in favour of using `std::visit` and `std::variant` judiciously going forward (at least in non-performance intensive code).

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2468
List of various issues fixed: https://github.com/dosbox-staging/dosbox-staging/issues/2468#issuecomment-1596526490

# Manual testing

Proof of manual testing. These are end-to-end tests that exercise the whole stack; it would be difficult to create an automated test fixture/harness that wires up )just enough components_ to make this possible and would yield the same level of confidence.

These are _highest-value tests_, the absolutely highest value there is. Therefore, I'd like to keep them around by all means, preferably in an automated form. They are invaluable for regression testing when anyone touches the mixer the next time (some situations are complex and subtle; most people wouldn't even know what regressions to look out for... that includes me 6 months from now :sunglasses:).

Btw, one of the reasons why this PR was dragging on for so long was that redoing all this testing manually takes like 20 minutes (in a relaxed manner) and I was getting sick of it after re-doing it for the 10th time...

My low-cost solution is to write a batch file containing these commands that pipes the results to a file, then have an `expected.txt` that contains the expected outcome with the ANSI colour codes stripped. Then write a bash script that takes the actual output, strips the ANSI stuff out, then diffs it against `expected.txt`. Would do the job, but better ideas are welcome (hint: suggesting to write an actual end-to-end test harness that would take days or weeks is _not_ a better idea :sunglasses:)

## Error messages

error-messages-1

error-messages-2

error-messages-3

error-messages-4

## Setting persistence

setting-persistence-1

setting-persistence-2

setting-persistence-3

## MIDI stuff

midi-1

midi-2

midi-3

midi-4

# Checklist

I have:

### DIFF
--- a/include/channel_names.h
+++ b/include/channel_names.h
@@ -1,0 +1,83 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOSBOX_CHANNEL_NAMES_H
+#define DOSBOX_CHANNEL_NAMES_H
+
+#include <vector>
+
+namespace ChannelName {
+
+constexpr auto CdAudio             = "CDAUDIO";
+constexpr auto Cms                 = "CMS";
+constexpr auto CovoxDac            = "COVOX";
+constexpr auto DisneySoundSource   = "DISNEY";
+constexpr auto FluidSynth          = "FSYNTH";
+constexpr auto GravisUltrasound    = "GUS";
+constexpr auto IbmMusicFeatureCard = "IMFC";
+constexpr auto InnovationSsi2001   = "INNOVATION";
+constexpr auto Master              = "MASTER";
+constexpr auto Opl                 = "OPL";
+constexpr auto PcSpeaker           = "PCSPEAKER";
+constexpr auto Ps1AudioDac         = "PS1DAC";
+constexpr auto Ps1AudioPsg         = "PS1";
+constexpr auto ReelMagic           = "REELMAGIC";
+constexpr auto RolandMt32          = "MT32";
+constexpr auto SoundBlasterDac     = "SB";
+constexpr auto StereOn1Dac         = "STON1";
+constexpr auto TandyDac            = "TANDYDAC";
+constexpr auto TandyPsg            = "TANDY";
+
+} // namespace ChannelName
+
+// A list of all possible mixer channel names (for enhancements only;
+// core functionality should still work even if this list is completely
+// empty).
+//
+// Currently, this is only used by the MIXER command to report errors more
+// accurately. The worst that can happen if a new channel is not added to the
+// list is that certain error messages will be a little less helpful. Because of
+// how this list is used, inventing a mechanism to build this list automatically
+// at startup would be overkill.
+//
+// clang-format off
+static std::vector<std::string> AllChannelNames = {
+        ChannelName::CdAudio,
+        ChannelName::Cms,
+        ChannelName::CovoxDac,
+        ChannelName::DisneySoundSource,
+        ChannelName::FluidSynth,
+        ChannelName::GravisUltrasound,
+        ChannelName::IbmMusicFeatureCard,
+        ChannelName::InnovationSsi2001,
+        ChannelName::Master,
+        ChannelName::RolandMt32,
+        ChannelName::PcSpeaker,
+        ChannelName::Ps1AudioPsg,
+        ChannelName::Ps1AudioDac,
+        ChannelName::ReelMagic,
+        ChannelName::SoundBlasterDac,
+        ChannelName::StereOn1Dac,
+        ChannelName::TandyPsg,
+        ChannelName::TandyDac,
+};
+// clang-format on
+
+#endif // DOSBOX_CHANNEL_NAMES_H
--- a/include/channel_names.h
+++ b/include/channel_names.h
@@ -25,25 +25,25 @@
 
 namespace ChannelName {
 
-constexpr auto CdAudio             = "CDAUDIO";
-constexpr auto Cms                 = "CMS";
-constexpr auto CovoxDac            = "COVOX";
-constexpr auto DisneySoundSource   = "DISNEY";
-constexpr auto FluidSynth          = "FSYNTH";
-constexpr auto GravisUltrasound    = "GUS";
-constexpr auto IbmMusicFeatureCard = "IMFC";
-constexpr auto InnovationSsi2001   = "INNOVATION";
-constexpr auto Master              = "MASTER";
-constexpr auto Opl                 = "OPL";
-constexpr auto PcSpeaker           = "PCSPEAKER";
-constexpr auto Ps1AudioDac         = "PS1DAC";
-constexpr auto Ps1AudioPsg         = "PS1";
-constexpr auto ReelMagic           = "REELMAGIC";
-constexpr auto RolandMt32          = "MT32";
-constexpr auto SoundBlasterDac     = "SB";
-constexpr auto StereOn1Dac         = "STON1";
-constexpr auto TandyDac            = "TANDYDAC";
-constexpr auto TandyPsg            = "TANDY";
+constexpr auto CdAudio              = "CDAUDIO";
+constexpr auto Cms                  = "CMS";
+constexpr auto CovoxDac             = "COVOX";
+constexpr auto DisneySoundSourceDac = "DISNEY";
+constexpr auto FluidSynth           = "FSYNTH";
+constexpr auto GravisUltrasound     = "GUS";
+constexpr auto IbmMusicFeatureCard  = "IMFC";
+constexpr auto InnovationSsi2001    = "INNOVATION";
+constexpr auto Master               = "MASTER";
+constexpr auto Opl                  = "OPL";
+constexpr auto PcSpeaker            = "PCSPEAKER";
+constexpr auto Ps1AudioCardDac      = "PS1DAC";
+constexpr auto Ps1AudioCardPsg      = "PS1";
+constexpr auto ReelMagic            = "REELMAGIC";
+constexpr auto RolandMt32           = "MT32";
+constexpr auto SoundBlasterDac      = "SB";
+constexpr auto StereoOn1Dac         = "STON1";
+constexpr auto TandyDac             = "TANDYDAC";
+constexpr auto TandyPsg             = "TANDY";
 
 } // namespace ChannelName
 
@@ -62,21 +62,21 @@ static std::vector<std::string> AllChannelNames = {
         ChannelName::CdAudio,
         ChannelName::Cms,
         ChannelName::CovoxDac,
-        ChannelName::DisneySoundSource,
+        ChannelName::DisneySoundSourceDac,
         ChannelName::FluidSynth,
         ChannelName::GravisUltrasound,
         ChannelName::IbmMusicFeatureCard,
         ChannelName::InnovationSsi2001,
         ChannelName::Master,
-        ChannelName::RolandMt32,
         ChannelName::PcSpeaker,
-        ChannelName::Ps1AudioPsg,
-        ChannelName::Ps1AudioDac,
+        ChannelName::Ps1AudioCardDac,
+        ChannelName::Ps1AudioCardPsg,
         ChannelName::ReelMagic,
+        ChannelName::RolandMt32,
         ChannelName::SoundBlasterDac,
-        ChannelName::StereOn1Dac,
-        ChannelName::TandyPsg,
+        ChannelName::StereoOn1Dac,
         ChannelName::TandyDac,
+        ChannelName::TandyPsg,
 };
 // clang-format on
 
--- a/include/channel_names.h
+++ b/include/channel_names.h
@@ -68,6 +68,7 @@ static std::vector<std::string> AllChannelNames = {
         ChannelName::IbmMusicFeatureCard,
         ChannelName::InnovationSsi2001,
         ChannelName::Master,
+        ChannelName::Opl,
         ChannelName::PcSpeaker,
         ChannelName::Ps1AudioCardDac,
         ChannelName::Ps1AudioCardPsg,
--- a/src/dos/program_mixer.cpp
+++ b/src/dos/program_mixer.cpp
@@ -104,8 +104,7 @@ void MIXER::Run()
 
 	auto showStatus = !cmd->FindExist("/NOSHOW", true);
 
-	std::vector<std::string> args = {};
-	cmd->FillVector(args);
+	auto args = cmd->GetArguments();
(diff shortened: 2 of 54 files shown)


weirddan455:
I was curious about the `std::variant` stuff as I had never really used it.  It looks like it's just a fancy tagged union and `std::visit` is just a fancy switch statement over the type tag.

>  If everybody hates them, we can discuss finding an alternative way, but it would be a lot more complex and there would be a lot of boilerplate involved, so I'm strongly in favour of using std::visit and std::variant judiciously going forward (at least in non-performance intensive code).

IMO for your use case I think it's syntactically a bit nicer to have `Command` be a pure virtual interface class and the different types of commands inherit from that with a virtual `Execute` function.  That gets rid of the `Executor` class, and having to type out the template type of `std::variant` everywhere.

However, `std::variant` avoids having to heap allocate every `Command` and also avoids the virtual function overhead.  So my initial assessment is the opposite.  It seems a bit worse syntactically but probably has better performance.

Overall, I don't hate it.  I think you could do exactly the same thing with a tagged union and a switch statement.  But if you like the syntax of `std::variant` better then go for it.

johnnovak:
Thanks for the feedback @weirddan455. Just a few further comments / thoughts why I did things this particular way. I'm not "arguing" as you're already fine with this, just offering further explanations, maybe for the benefit of others too.

And my comment still stands; if everybody else than me really doesn't like this, then we're back to tagged unions / virtual functions / out vars 😄 

> IMO for your use case I think it's syntactically a bit nicer to have `Command` be a pure virtual interface class and the different types of commands inherit from that with a virtual `Execute` function. That gets rid of the `Executor` class, and having to type out the template type of `std::variant` everywhere.

I like the `Executor` class because we also need to maintain a state when executing a list of commands. Look at the private vars. Also, on a pure aesthetic level, I like the total decoupling of command data from actions in this case. In more complex cases, you could even have different execution strategies implemented in different executors (not needed here, of course; just noting it as an advantage of this approach versus virtual functions).

```cpp
struct Executor {
	void operator()(const SelectChannel cmd);
	void operator()(const SetVolume cmd);
	void operator()(const SetStereoMode cmd);
	void operator()(const SetCrossfeedStrength cmd);
	void operator()(const SetReverbLevel cmd);
	void operator()(const SetChorusLevel cmd);

private:
	bool global_command = false;

	// If 'master_channel' is true, then the MASTER channel is selected,
	// otherwise 'channel' points to the selected non-master channel.
	bool master_channel = false;

	std::shared_ptr<MixerChannel> channel = {};
};
```

> However, `std::variant` avoids having to heap allocate every `Command` and also avoids the virtual function overhead. So my initial assessment is the opposite. It seems a bit worse syntactically but probably has better performance.

Yeah, performance is absolutely a non-issue here, that wasn't even a consideration. We're only dealing with a handful (less than say 20) mixer commands on each MIXER command invocation. There are only a few (or zero) MIXER command invocations typically in the autoexec section of a config, or users execute MIXER commands manually several tens of seconds apart. No tight loops, no tens or hunders of thousands of commands being continuously evaluated, so maintainability trumps everything in this case.

However, for my OSD work, I will also need a similar command pattern, but that will need to be efficient as thousands of draw commands will be evaluated per-frame at the FPS rate. But even in that case, I'd measure the performance impact of this, but you're saying `std::variant` is actually preferable for performance reasons, so that's something I should remember then.
 
> Overall, I don't hate it. I think you could do exactly the same thing with a tagged union and a switch statement. 

A nice property of the `Executor` is that if you forget to define an `operator()` for a variant, then you'll get compilation errors. Even if you use a tagged union with proper enum classes, you'll only get "soft" exhaustiveness check warnings if you miss a few cases in your switch blocks I believe? Not such a big issue in this case where the variants and their executor are defined in the same file, but in bigger projects where the commands are executed in multiple ways by multiple executors in multiple places throughout the codebase, the tagged union approach can become rather brittle I think. Overall, I like hard compile-time checks whenever I can take them.

> But if you like the syntax of `std::variant` better then go for it.

There's one more thing as well which I mentioned, namely that I like to use `std::variant` as a poor man's Either (or Result) type. Look at these:

```cpp
std::variant<Error, std::queue<Command>> ParseCommands(...)

static std::variant<Error, Command> parse_volume_command(const std::string& s,
                                                         const std::string& channel_name)

static std::variant<Error, Command> parse_reverb_command(const std::string& s,
                                                         const std::string& channel_name)
```

What I'm "simulating" here is Scala's [Either type](https://www.baeldung.com/scala/either-type) which is super handy for propagating either a result of type R, or an error type E as a single self-contained unit from functions. So you define it as `Either[E, R]`, then upstream you can make decisions whether it's a "left" or "right" outcome. Sometimes people use a renamed variant of this called Result, where the "left" side is the error type, and the "right side is the success type.

Without this, we're back to the old clunky way of using an "out type" for the actual error, something like this:

```cpp
static std::option<Command> parse_volume_command(const std::string& s,
                                                         const std::string& channel_name, Error& out_error)
```

Sure, that works, but I find the `std::variant` return types nicer and more descriptive. It's a _bit_  boilerplat-y at the call site, but not overly so:

```cpp
const auto result = parse_crossfeed_command(arg, curr_channel_name);

if (auto cmd = std::get_if<Command>(&result); cmd) {
	commands.emplace(*cmd);
	++curr_channel_command_count;
} else {
	return std::get<Error>(result);
}
```

The clunky option with the out param would look like this. You don't need to do the `std::get_if<T>` and `std::get<E>` parts, true, but you need a mutable var for the error out var... which gets initialised with... something 😅  I kinda borderline prefer the `std::variant` version because it's more "correct", but like I said, if enough people hate the extra verbosity then I'm gonna revert back to the clunky out var way 😄 

```cpp
Error err = {};
if (auto result = parse_crossfeed_command(arg, curr_channel_name, err); result) {
	commands.emplace(*result);
	++curr_channel_command_count;
} else {
	return err;
}
```

weirddan455:
>  However, for my OSD work, I will also need a similar command pattern, but that will need to be efficient as thousands of draw commands will be evaluated per-frame at the FPS rate. But even in that case, I'd measure the performance impact of this, but you're saying std::variant is actually preferable for performance reasons, so that's something I should remember then.

Yeah, measuring is the only way to be sure when it comes to performance.  I could be wrong here but I think you would want to avoid doing thousands of heap allocations per frame if you can.  Also, virtual functions in C++ are a bit worse than a normal function pointer because you have a take a double indirection through the vtable.

>  There's one more thing as well which I mentioned, namely that I like to use std::variant as a poor man's Either (or Result) type. Look at these:

I saw those for error handling.  I don't love it but there's also not really a better option.  I'd prefer `std::optional` if you only need to know success or fail but if you need to return a specific error code, it gets ugly.  If you can't add a field to the return struct (which you can't really because `Command` is itself a typedef to `std::variant`), it's either out parameters or doing something like returning `nullptr` and setting a global `errno` type variable.

So `std::variant` might actually be the least bad option for functions that can fail with specific error codes.

johnnovak:
> Yeah, measuring is the only way to be sure when it comes to performance. I could be wrong here but I think you would want to avoid doing thousands of heap allocations per frame if you can. Also, virtual functions in C++ are a bit worse than a normal function pointer because you have to take a double indirection through the vtable.

Absolutely, I have a big pre-alloced `std::vector` in mind that has `std::variant` elements that need no heap allocation. That's as lightweight as it gets. Technically, you wouldn't even need to prealloc because the vector doesn't give memory "back" easily unless prodded a little bit, but yeah, why not prealloc to 5k elements or something.

Yeah, virtual functions involve an extra indirection... but then, many languages have nothing else than the equivalent of the C++ virtual functions. Most of the C performance discussion at some point devolve into saying things like, "you know, but the fastest car is one without brakes, windshield, etc". Yeah, suuuuure, if you turn all checks, indirection, all the niceties off, it's "theoretically the fastest possible"—but do I want that? 😄 Maybe some very low-level number crunching, but not much else... Not criticising what you said, needs to be performance measured and that's it, just reflecting in general to the "max bare-metal speed or go home" C coder attitude 😎 

> I saw those for error handling. I don't love it but there's also not really a better option. 
> So `std::variant` might actually be the least bad option for functions that can fail with specific error codes.

Yeah, I'm exactly the same—don't really _love_ it, but it's the least bad option, and overall it ticks all the boxes, except for the (slight) verbosity. At the end of the day, C++ just lacks the general "infrastructure" to make these things nice and lean and minimal syntax like in better languages, but it is what it is.

johnnovak:
its-happening

My desperate mud-wrestling match with the `MIXER` has finally ended! 😎 🤘🏻 

Brace yourselves @kcgen @FeralChild64 @weirddan455 @kklobe 

I've tested _the shit_ out of it—the person who finds a bug gets a present!

include/channel_names.h
```diff
@@ -0,0 +1,84 @@
...
+        ChannelName::PcSpeaker,
+        ChannelName::Ps1AudioCardDac,
+        ChannelName::Ps1AudioCardPsg,
+        ChannelName::ReelMagic,
+        ChannelName::RolandMt32,
+        ChannelName::SoundBlasterDac,
+        ChannelName::StereoOn1Dac,
+        ChannelName::TandyDac,
+        ChannelName::TandyPsg,
+};
```
kcgen, line 81: These would be really nice to make them an enum class, and eliminate the strings entirely.  For logging, we could have a std::map or the usual enum-to-char* switch function.

The nice this is, these constexpr's already look like they're enums :), for example: ChannelName::GravisUltrasound, so I think the conversion could be pretty straight forward.

(This could also be done in a follow up PR)

include/channel_names.h
```diff
@@ -0,0 +1,84 @@
...
+// clang-format on
```
kcgen, line 82: Out of scope out of scope from this PR, but this (by appearances) gets us closer to dropping string-based identification of mixer channels in favour of logical enum-class-based IDs, because the new `ChannelName::FluidSynth` identifiers (currently assigned to strings) will make the enum class refactoring a drop in replacement :-)

Of course, we'd still need an enum-to-string switch function for logging names and parsing `MIXER` CLI channels, but that's expected :)

kcgen (approved):
Did extensive mixer CLI and conf testing and couldn't cause crashes or unexpected behavior.

The new error handling is excellent, @johnnovak !

Reviewed all the code and nothing is jumping out, and appreciate your discussion, @weirddan455, regarding std::variant and alternatives.

Also, :medal_sports: for the extensive unit tests for all the CLI parsing, conversions (volume percent and dB), and channel features/handling.

---

@FeralChild64 , @weirddan455 , @kklobe : over to you for more checks :-)

FeralChild64:
It would be nice to have it rebased and conflicts solved - yesterday another batch of MIXER fixes were merged, it would be best to test both changes together.

johnnovak:
> It would be nice to have it rebased and conflicts solved - yesterday another batch of MIXER fixes were merged, it would be best to test both changes together.

Done.

src/dos/program_mixer.cpp
```diff
@@ -260,23 +801,29 @@ void MIXER::AddMessages()
 	        "  [color=light-green]mixer[reset] [/listmidi]\n"
```
FeralChild64, line 801: `[color=light-cyan][CHANNEL][reset]` suggests the channel is optional, where in fact is is required. It should be changed to `[color=light-cyan]CHANNEL[reset]`.

FeralChild64:
I played a little with the command, just one minor remark from my side.

src/dos/program_mixer.cpp
johnnovak: It is optional, e.g., `MIXER x30 r10 c50` is entirely valid. See the unit tests for more examples.

johnnovak:
Thanks @kcgen and @FeralChild64 for the review. Merge time 😎

kcgen:
@johnnovak -> can you check:

`lpt_dac = disney`, then see if `mixer disney 50` works?

Getting an 'invalid command' on my side:

johnnovak:
> @johnnovak -> can you check:
> 
> `lpt_dac = disney`, then see if `mixer disney 50` works?
> 
> Getting an 'invalid command' on my side:
> 
> 

Confimed @kcgen, it's a bug. Started debugging it, maybe I'll have time to finish it tomorrow and raise a fix PR.

PR #2947, opened 2023-09-20
Improve the online help of the MORE command

johnnovak:
# Description

This is a minor usability enhancement for the online help text of the `MORE` command. I personally use Vim keys when I use Unix `more` and `less`, so I'm constantly confused about what `Space` and `Enter` do. Other people who are new to the `MORE` command might face similar issues. Also, if they open a long text file by mistake, they might not know how to quit the program with the `Q` key.

# Manual testing

Tested displaying a single file with `MORE`, multiple files, and built-in commands with long help texts

image1876470530-rendered

image1876470529

# Checklist

I have:

### DIFF
--- a/src/dos/program_more.cpp
+++ b/src/dos/program_more.cpp
@@ -226,11 +226,11 @@ void MORE::AddMessages()
 	MSG_Add("PROGRAM_MORE_NEW_DEVICE",
 	        "[reset][color=light-yellow]--- device %s ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_SINGLE",
-	        "[reset][color=light-yellow]--- press SPACE or ENTER for more ---[reset]");
+	        "[reset][color=light-yellow]--- press SPACE for next page, ENTER for next line, Q to quit ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_PERCENT",
-	        "[reset][color=light-yellow]--- (%d%%) press SPACE or ENTER for more ---[reset]");
+	        "[reset][color=light-yellow]--- (%d%%) press SPACE for next page, ENTER for next line, Q to quit ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_MULTI",
-	        "[reset][color=light-yellow]--- press SPACE or ENTER for more, N for next file ---[reset]");
+	        "[reset][color=light-yellow]--- press SPACE or ENTER for more, N for next file, Q to quit ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_LINE",
 	        "[reset][color=light-yellow]--- line %u ---[reset]");
 	MSG_Add("PROGRAM_MORE_OPEN_ERROR",


src/dos/program_more.cpp
```diff
@@ -226,11 +226,11 @@ void MORE::AddMessages()
...
 	        "[reset][color=light-yellow]--- device %s ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_SINGLE",
-	        "[reset][color=light-yellow]--- press SPACE or ENTER for more ---[reset]");
+	        "[reset][color=light-yellow]--- press SPACE for next page, ENTER for next line, Q to quit ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_PERCENT",
-	        "[reset][color=light-yellow]--- (%d%%) press SPACE or ENTER for more ---[reset]");
+	        "[reset][color=light-yellow]--- (%d%%) press SPACE for next page, ENTER for next line, Q to quit ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_MULTI",
-	        "[reset][color=light-yellow]--- press SPACE or ENTER for more, N for next file ---[reset]");
+	        "[reset][color=light-yellow]--- press SPACE or ENTER for more, N for next file, Q to quit ---[reset]");
```
johnnovak, line 233: It would make the line too long to explain what SPACE and ENTER do here as well.

Kappa971: Too long for Italian 😣:
**Old**
[reset][color=light-yellow]--- premere SPAZIO o INVIO per mostrare di più, N per il file successivo ---[reset]
(77 characters) _OK_
**New**
[reset][color=light-yellow]--- premere SPAZIO o INVIO per mostrare di più, N per il file successivo, Q per uscire ---[reset] 
(91 characters) _too long_

Also:
**Old**
[reset][color=light-yellow]--- (%d%%) premere SPAZIO o INVIO per mostrare di più ---[reset]
(58 characters) _OK_
**New**
[reset][color=light-yellow]--- (%d%%) premere SPAZIO per la pagina successiva, INVIO per la prossima riga, Q per uscire ---[reset]
(97 characters) _too long_

johnnovak: Too bad? 😄 Try to abbreviate the Italian version, I guess.

Kappa971: > Too bad? 😄 Try to abbreviate the Italian version, I guess.
> 
> I guess English is the most important, the rest are "nice to haves". You can also just leave the Italian text as is.

II think I have found an acceptable solution as stated above. Thanks

johnnovak: I guess my point is, I don't really wanna _not_ improve the English experience because of the translations. The translation can stay exactly as they are now as a last resort, but I see no point why not to make the most used English option better.

FeralChild64: English is typically more "compact" than Slavic or Romance (neo-Latin) languages, we have to live with it somehow. Since this string is just a hint to the user, I see no problem with skipping some information in certain translations.

Windows 11 only prints something like "-- More --", if I remember correctly :)

FeralChild64:
Just a single minor (optional) remark.

src/dos/program_more.cpp
Kappa971: > Windows 11 only prints something like "-- More --", if I remember correctly :)

Maybe DOSBox should do the same thing? MORE /? indicates all keys...

FeralChild64: I don’t like to be a copycat, especially if I can keep the spirit of the system, but at the same time making it better.

Also, Windows console users typically know what `MORE` is for - and I would prefer to be ready for users who didn’t use DOS-style prompt since 25 years, or maybe even are using it for the first time in their life.

Kappa971: > I don’t like to be a copycat, especially if I can keep the spirit of the system, but at the same time making it better.
> 
> Also, Windows console users typically know what `MORE` is for - and I would prefer to be ready for users who didn’t use DOS-style prompt since 25 years, or maybe even are using it for the first time in their life.

That's right, but that's why help commands exist. 
Anyway I will wait for the definitive solution before making any changes to the translation.

src/dos/program_more.cpp
```diff
@@ -226,11 +226,11 @@ void MORE::AddMessages()
 	MSG_Add("PROGRAM_MORE_NEW_DEVICE",
 	        "[reset][color=light-yellow]--- device %s ---[reset]");
 	MSG_Add("PROGRAM_MORE_PROMPT_SINGLE",
-	        "[reset][color=light-yellow]--- press SPACE or ENTER for more ---[reset]");
```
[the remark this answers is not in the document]
johnnovak: I think that's unnecessary because nobody uses the 40-column modes willingly 😄 Then all the help texts and config setting descriptions assume 80 columns at least, the output of the mixer command too, and probably a few other things. I except most things to break in 40-column modes, in other words, and I'm fine with that 😄

src/dos/program_more.cpp
johnnovak: > English is typically more "compact" than Slavic or Romance (neo-Latin) languages, we have to live with it

I think most compact than any other language that uses the Latin alphabet 😏 I'm just thinking of German, Finnish, and Hungarian here. MS-DOS would've defaulted to 132-column text modes if it was invented by those people 😅 Whenever I translate English to Hungarian, there's a minimum 1.5x-2x length increase 😄

Kappa971: > I think more compact than any other language that uses the Latin alphabet 😏 I'm just thinking of German, Finnish, and Hungarian here. MS-DOS would've defaulted to 132-column text modes if it was invented by those people 😅 

"Contemporaneamente"
Which is probably (probabilmente 😆) the longest commonly used word.

>Whenever I translate English to Hungarian, there's a minimum 1.5x-2x length increase 😄

You should create a Hungarian translation...

johnnovak: > You should create a Hungarian translation...

Nah, I dislike translated OSes. I don't even like using the Hungarian keyboard layout; I just use English letters without accents if I can avoid it 😄 The special symbols are at different wacky places too, ugh...

Every time I had to fix something on some relative's computer that had Hungarian localisation, I was completely lost plus the translations seemed like an absurdist comedy to me, so I'm not the best person for the job 😆

Kappa971: >Nah, I dislike translated OSes. I don't even like using the Hungarian keyboard layout; I just use English letters without accents if I can avoid it 😄 The special symbols are at different wacky places too, ugh...

I have no idea why there are so many keyboard layouts... For example, most European languages could use the same layout and instead there are Italian, French, German, etc. layouts. Obviously a Greek can't use the same keyboard as an American.

>Every time I had to fix something on some relative's computer that had Hungarian localisation, I was completely lost plus the translations seemed like an absurdist comedy to me, so I'm not the best person for the job 😆 

You need to understand that people generally use things that make them feel comfortable. Using your native language is always easier, and this also applies to an operating system that must interface with people. 
The magic word is "accessibility". I'm in favor of everything that can improve accessibility, in any area and I'm against, for example, an English-only world, the result of a world where we all must be homologated.

johnnovak: > You need to understand that people generally use things that make them feel comfortable.

Exactly—hence I'm using English 😄

FeralChild64 (approved):

src/dos/program_more.cpp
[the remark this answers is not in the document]
johnnovak: > OK, so feel free to merge it as is.
> 
> (offtopic) The second screenshot looks very intriguing :)

Yeah, I've started the "preparations" 😄

Issue #2793, opened 2023-08-27
Store FAT file attributes as extended attributes on the host filesystem


johnnovak:
Work has started on this, but was abandoned. We collectively came to the conclusion to simply mirror the Wine and Dosemu2 implementation of the `user.DOSATTRIB` extended attribute:
https://github.com/dosbox-staging/dosbox-staging/pull/2693#issuecomment-1646815262

See abandoned implementation attempt and discussion here:
https://github.com/dosbox-staging/dosbox-staging/pull/2693

Torinde:
A pity, it seemed that @NicknineTheEagle went through the hard part and just a few coding remarks were pending.

FeralChild64:
Started working on this, as I would really like to have this feature - https://github.com/dosbox-staging/dosbox-staging/tree/fc/fs-attrib-1

- refactored the code a bit, most of the original PR review remarks are now handled
- changed the attribute format to be (in general) WINE-compatible

Still to be fixed/improved/done:
- more refactoring / cleanup / comments in the code
- for the read-only attribute WINE takes away the owner-write permission (like DOSBox Staging before this patch - this is something I have to fix
- if I'm not mistaken, DOS sets the Archive bit every time the file is created/modified - I haven't really tested this yet

There is one limitation regarding WINE compatibility I intend to keep: WINE does not allow to clear the Archive attribute on files, we do :) Files without `user.DOSATTRIB` extended attribute are considered to have Archive attribute set.

----

Unfortunately, I won't have much time in the near future (like a week or two) - if anyone wants to take over, feel free. If not, I'll finish when I have free time again.

johnnovak:
That's great @FeralChild64; thanks for taking this over. Please make sure to credit @NicknineTheEagle as well in the commit logs somehow for the initial implementation.

There's no rush, it will take a while until I finish my outstanding tickets for the release candidate, so you'll have time to finish this at a leisurely pace once you're back. The auto-switching shader stuff caused a 5-6 week delay, but it was well worth it. But no more major scope creeps 😅

johnnovak:
Would you have the appetite to tackle this as well @FeralChild64 ? It seems to be related and it might make your job even easier overall.

https://github.com/dosbox-staging/dosbox-staging/issues/2663

FeralChild64:
We'll see, I'll have a look at https://github.com/dosbox-staging/dosbox-staging/issues/2663 - but it might be best to do it in a separate PR.

[...]

PR #2951, opened 2023-09-21
Implement DOS file/directory attributes support

FeralChild64:
# Description

This is a next generation version of the patch originally prepared by @NicknineTheEagle https://github.com/dosbox-staging/dosbox-staging/pull/2693

First short explanation: Archive bit in MS-DOS means _file is newly created or modified, and archivizer software should make a backup copy during next run and clean the bit afterwards_ (at least this was the idea, in reality most of the files in MS-DOS system have Archive bit set).

This PR implements the following:
- all DOS file/directory attributes are now supported (not just Read-Only, as before)
- newly created or modified files get the Archive bit set (on both local drive and FAT images), I have tried to duplicate MS-DOS behavior as closely as possible
- on Windows host attributes are stored normally in the filesystem
- on Linux / macOS hosts Read-Only is denoted by clearing write permissions (as before), moreover all the attributes are stored in _xattr_ extended attributes, in a format compatible with WINE and DOSEmu 2 (they both use Samba 3 format); lack of extended attributes is considered as Archive bit set and Hidden / System not set (as before this implementation)

Limitations:
- overlay drives (`-t overlay` option in `MOUNT` command) are partially supported; attributes can be retrieved (always), but can be only set for files in the overlay - before anyone complains: current implementation does not even allow to set file in the overlay as Read-Only (request is mostly ignored), so the situation is actually improved now :)
- interoperability with WINE does not cover Archive bit; WINE always assumes it is set for the files and does not allow to change it
- interoperability with DOSEmu 2 wasn't tested, as I am unable to set up the emulator (crashes for me instantly after startup)

## Related issues
Implements https://github.com/dosbox-staging/dosbox-staging/issues/2793

# Manual testing

1. Just play with `ATTRIB` command or other tools to change attributes; all newly created or modified files (for example using `EDIT.COM` from MS-DOS) should get the Archive attribute set. Tested on Linux and Windows (yes, despite I'm a Linux user, I have launched it, waited till it installs 6 months of updates, and did some tests!). I was unable to test under macOS, I don't have access to Mac.

2. To test interoperability with WINE, the easiest way is to symlink some directory from DOSBox local drive to WINE virtual drive, execute `wine cmd` command from console, and play with `ATTRIB` commands of both emulators; note that WINE ignores Archive attribute (always has it set for files) and does not allow to change it.

3. I wasn't sure how MS-DOS precisely handles Archive attribute (does it set it after you open file for writing? after you write to the file? when you are closing the modified file?), so I have written a test program (attached below). To try it:
- rename file to `ATEST.C`
- install OpenWatcom inside DOSBox
- execute `OWSETENV.BAT` from the compiler directory
- compile the test program using `WCL.EXE -l=dos ATEST.C` command
- run the executable; for convenience, it shows reference values - attributes in hex format, retrieved while running the program under real MS-DOS 6.22

[ATEST.C.TXT](https://github.com/dosbox-staging/dosbox-staging/files/12688978/ATEST.C.TXT)

# Checklist

I have:

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -171,7 +171,8 @@ private:
 
 class localFile : public DOS_File {
 public:
-	localFile(const char* name, FILE* handle, const char* basedir);
+	localFile(const char* name, const std_fs::path& path, FILE* handle,
+	          const char* basedir);
 	localFile(const localFile&)            = delete; // prevent copying
 	localFile& operator=(const localFile&) = delete; // prevent assignment
 	bool Read(uint8_t* data, uint16_t* size) override;
@@ -189,15 +190,25 @@ public:
 	{
 		return basedir;
 	}
+	std_fs::path GetPath() const
+	{
+		return path;
+	}
 	FILE* fhandle = nullptr; // todo handle this properly
 private:
-	const char *basedir;
-	long stream_pos = 0;
+	const std_fs::path path = {};
+	const char* basedir     = nullptr;
+	long stream_pos         = 0;
+
 	bool ftell_and_check();
 	void fseek_and_check(int whence);
 	bool fseek_to_and_check(long pos, int whence);
-	bool read_only_medium;
-	enum { NONE,READ,WRITE } last_action;
+
+	bool read_only_medium     = false;
+	bool set_archive_on_close = false;
+
+	enum class LastAction : uint8_t { None, Read, Write };
+	LastAction last_action = LastAction::None;
 };
 
 /* The following variable can be lowered to free up some memory.
--- a/src/dos/drive_overlay.cpp
+++ b/src/dos/drive_overlay.cpp
@@ -47,12 +47,13 @@ using namespace std;
 #define	CROSS_DOSFILENAME(blah) strreplace(blah,'/','\\')
 #endif
 
-/* 
- * design principles/limitations/requirements:
- * 1) All filenames inside the overlay directories are UPPERCASE and conform to the 8.3 standard except for the special DBOVERLAY files.
+/*
+ * Design principles/limitations/requirements:
+ * 1) All filenames inside the overlay directories are UPPERCASE and conform to
+ *    the 8.3 standard except for the special DBOVERLAY files.
  * 2) Renaming directories is currently not supported.
- *
- * Point 2 is still being worked on.
+ * 3) It is only possible to change file attributes for files present in the
+ *    overlay.
  */
 
 /* New rename for base directories:
@@ -197,8 +198,9 @@ bool Overlay_Drive::TestDir(char * dir) {
 
 class OverlayFile final : public localFile {
 public:
-	OverlayFile(const char *name, FILE *handle, const char *basedir)
-	        : localFile(name, handle, basedir),
+	OverlayFile(const char* name, const std_fs::path& path, FILE* handle,
+	            const char* basedir)
+	        : localFile(name, path, handle, basedir),
 	          overlay_active(false)
 	{
 		if (logoverlay)
@@ -232,7 +234,8 @@ public:
 //Create leading directories of a file being overlayed if they exist in the original (localDrive).
 //This function is used to create copies of existing files, so all leading directories exist in the original.
 
-FILE* Overlay_Drive::create_file_in_overlay(const char* dos_filename, const char* mode)
+std::pair<FILE*, std_fs::path> Overlay_Drive::create_file_in_overlay(
+        const char* dos_filename, const char* mode)
 {
 	if (logoverlay) LOG_MSG("create_file_in_overlay called %s %s",dos_filename,mode);
 	char newname[CROSS_LEN];
@@ -254,7 +257,7 @@ FILE* Overlay_Drive::create_file_in_overlay(const char* dos_filename, const char
 		f = fopen_wrap(newname,mode);
 	}
 
-	return f;
+	return {f, newname};
 }
 
 #ifndef BUFSIZ
@@ -302,7 +305,9 @@ bool OverlayFile::create_copy() {
 	if (drive_set != 0xff && drive_set < DOS_DRIVES && Drives[drive_set]){
 		const auto od = dynamic_cast<Overlay_Drive*>(Drives[drive_set]);
 		if (od) {
-			newhandle = od->create_file_in_overlay(GetName(),"wb+"); //todo check wb+
+			std_fs::path path = {};
+			// TODO: check wb+
+			std::tie(newhandle, path) = od->create_file_in_overlay(GetName(), "wb+");
 		}
 	}
  
@@ -331,7 +336,9 @@ static OverlayFile* ccc(DOS_File* file) {
 	localFile* l = dynamic_cast<localFile*>(file);
 	if (!l) E_Exit("overlay input file is not a localFile");
 	//Create an overlayFile
-	OverlayFile *ret = new OverlayFile(l->GetName(), l->fhandle,
+	OverlayFile* ret = new OverlayFile(l->GetName(),
+	                                   l->GetPath(),
+	                                   l->fhandle,
 	                                   l->GetBaseDir());
 	ret->flags = l->flags;
 	ret->refCtr = l->refCtr;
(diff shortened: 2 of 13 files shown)


include/dos_system.h
```diff
@@ -189,15 +190,21 @@ class localFile : public DOS_File {
...
+	const std_fs::path path = {};
+	const char* basedir     = nullptr;
+	long stream_pos         = 0;
 	bool ftell_and_check();
 	void fseek_and_check(int whence);
 	bool fseek_to_and_check(long pos, int whence);
-	bool read_only_medium;
-	enum { NONE,READ,WRITE } last_action;
+	bool read_only_medium                  = false;
+	enum { NONE, READ, WRITE } last_action = NONE;
```
johnnovak, line 206: Could you change the enum values to CamelCase?

src/dos/drive_overlay.cpp
```diff
@@ -254,7 +257,7 @@ FILE* Overlay_Drive::create_file_in_overlay(const char* dos_filename, const char
 		f = fopen_wrap(newname,mode);
 	}
 
-	return f;
+	return std::make_pair(f, newname);
```
johnnovak, line 260: You should be able to just write `return {f, newname};`.

src/dos/drive_overlay.cpp
```diff
@@ -302,7 +305,9 @@ bool OverlayFile::create_copy() {
 	if (drive_set != 0xff && drive_set < DOS_DRIVES && Drives[drive_set]){
 		const auto od = dynamic_cast<Overlay_Drive*>(Drives[drive_set]);
 		if (od) {
-			newhandle = od->create_file_in_overlay(GetName(),"wb+"); //todo check wb+
+			std_fs::path path = {};
+			// TODO: check wb+
+			std::tie(newhandle, path) = od->create_file_in_overlay(GetName(), "wb+");
```
johnnovak, line 310: Can't rewrite to `const auto [newhandle, path] = ...`?

src/dos/drive_overlay.cpp
```diff
@@ -998,34 +1009,46 @@ bool Overlay_Drive::GetFileAttr(char *name, uint16_t *attr)
...
+	if (result == DOSERR_NONE) {
 		return true;
 	}
-	DOS_SetError(DOSERR_FILE_NOT_FOUND);
+
+	if ((result == DOSERR_FILE_NOT_FOUND || result == DOSERR_PATH_NOT_FOUND) &&
+	    !is_deleted_file(name) && localDrive::FileExists(name)) {
+		// We do not allow changing file attributes outside of overlay,
+		// instead we fail silently - limitation by design
+		return true;
```
johnnovak, line 1048: Nice comment 👍🏻

johnnovak:
Immediate assertation failure when trying to start DOS Navigator on macOS.

```
Assertion failed: (static_cast<size_t>(length) < xattr_max_length), function get_xattr, file fs_utils_posix.cpp, line 216.

Stack trace:
22         0x1086bd08c 26  dyld                                0x00000001086bd08c start + 520
21         0x1047340cc main + 36
20         0x104ae56ec sdl_main(int, char**) + 4796
19         0x1047c4a8c Config::StartUp() + 28
18         0x104e9251c SHELL_Init() + 3992
17         0x104e8f824 DOS_Shell::Run() + 1472
16         0x104e8f1a8 DOS_Shell::RunBatchFile() + 300
15         0x104e8ed5c DOS_Shell::ParseLine(char*) + 2008
14         0x104ea5d14 DOS_Shell::DoCommand(char*) + 440
13         0x104ed41e4 DOS_Shell::ExecuteProgram(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 996
12         0x104ed50cc run_binary_executable(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 1556
11         0x1048877a8 CALLBACK_RunRealInt(unsigned char) + 92
10         0x104734ac4 DOSBOX_RunMachine() + 28
9          0x104734a20 Normal_Loop() + 136
8          0x10494eba0 DOS_21Handler() + 7644
7          0x10495da88 DOS_CloseFile(unsigned short, bool, unsigned char*) + 256
6          0x1049a792c localFile::Close() + 112
5          0x104777594 local_drive_get_attributes(std::__fs::filesystem::path const&, FatAttributeFlags&) + 160
4          0x10477771c get_xattr(std::__fs::filesystem::path const&) + 172
3          0x1b9b79754 err + 0
2          0x1b9b7a340 abort + 168
1          0x1b9c3fee0 pthread_kill + 288
2023-09-22 10:32:34.318 | Signal: SIGABRT
zsh: abort      ~/jn/dosbox-staging/build/debug/dosbox --noprimaryconf

~/Documents/DOS/DOS Navigator
%
```

johnnovak:
If I comment out the assert, it works fine on macOS. But the `COPY` command does not carry over the attributes—is that intentional? Maybe that's how real MS-DOS works?

image

If I copy files in DOS Navigator, the attributes carry over (including the archive bit), but that's different because I think it has some special logic for that.

src/misc/fs_utils_posix.cpp
```diff
@@ -147,4 +151,182 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept
 
 #endif
 
+// ***************************************************************************
+// Local drive file/directory attribute handling, WINE-compatible
+// ***************************************************************************
+
+constexpr int ro_permissions = S_IRUSR | S_IRGRP | S_IROTH;
```
johnnovak, line 158: Constants are UpperCamelCase by convention. There are lots of new constant definitions here, so same comment for all of them. So `RoPermissions`, `XattrReadMask`, etc.

include/dos_system.h
FeralChild64: Done. Also changed type to `enum class`.

src/dos/drive_overlay.cpp
FeralChild64: OK, changed.

src/misc/fs_utils_posix.cpp
FeralChild64: OK, changed - I think all of them.

src/dos/drive_overlay.cpp
FeralChild64: Wouldn't be practical. `path` is not needed in this method (will be ignored), and `newhandle` is defined earlier (outside of this block) and it's scope cannot be reduced.

FeralChild64:
@johnnovak This assertion actually had wrong condition - it is now fixed, and assertion is replaced with more graceful code.

I have tested the `copy` command under real MS-DOS:
- it ignores files with Hidden/System attributes (our implementation is wrong here, not only for `copy`)
- the destination file gets Archive attribute (always) and loses ReadOnly attribute (here our implementation works correctly)

I'll create a separate issue in a moment for the attribute handling in our DOS commands; this is a bigger problem (IMHO out of scope of this PR), can already be reproduced (even without this PR) on mounted FAT partitions or floppy images, and might need a while to fully fix all the commands.

NicknineTheEagle:
Good to see that someone had motivation to finish this (unlike me).
Is updating permissions on Linux/Mac really necessary to keep, though? It feels hacky (I mean more hacky than using xattr) and user might have their own reasons for removing write permissions so having Dosbox mess with them may be undesirable. Plus it forces that awkward workaround where you have to set write permissions and then unset them to change other attributes.

src/misc/fs_utils_posix.cpp
```diff
@@ -147,4 +151,190 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept
...
+constexpr size_t XattrMaxLength = 4;
+
+static std::string fat_attribs_to_xattr(const FatAttributeFlags fat_attribs)
+{
+	return format_string("0x%x", fat_attribs._data & XattrWriteMask);
+}
+
+static std::optional<FatAttributeFlags> xattr_to_fat_attribs(const std::string& xattr)
+{
+	constexpr uint8_t hex_base = 16;
```
johnnovak, line 182: HexBase

FeralChild64: Changed.

FeralChild64:
> Is updating permissions on Linux/Mac really necessary to keep, though?

IMHO, yes:
- this improves interoperability between us, DOSBox SVN, and other forks (also between Linux and Windows builds of DOSBox Staging) 
- WINE does the same
- DOSEmu 2 does the same (I couldn't test it, but I see this in their source code: https://github.com/dosemu2/dosemu2/blob/devel/src/dosext/mfs/xattr.c, `set_dos_xattr` function)

johnnovak (approved):
Very nice work @FeralChild64, merge when you're ready 🚀

PR #2960, opened 2023-09-24
Remove the SSE 4.2 instruction set from the hardware requirements (`80.x` release branch PR)

### LINKED ISSUE #2955: SSE 4.2 listed as hardware requirement but does not appear to be used

_No response_

Linux x86_64

_No response_

### Relevant hardware info

_No response_

On the website, SSE 4.2 is listed as a requirement on x86:  https://dosbox-staging.github.io/downloads/linux/#hardware-requirements

However, I don't think we're setting the correct compiler flags to even utilize this.  We need to be setting `-march` to whatever the minimum supported CPU is.  https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

Doing a `git grep _m128` (to look for SIMD intrinsics) shows Voodoo emulation doing some SSE2 stuff.  SSE2 should be fine with the default `march` as I believe every x86_64 CPU supports SSE2 at a minimum.

The only other place I'm seeing intrinsics is inside `src/libs`.  `dr_flac.h` (part of SDL Sound from the looks) has a lot of it in particular, including SSE4.1.  However, it's all behind `#ifdefs` that I don't think we're defining anywhere.  I wouldn't be surprised if this is getting built without SIMD at all.  It looks like these files just got copy-pasted and we're compiling without running the build system they were meant to be built with.  The `meson.build` file in that directory is very bare bones.

I noticed this because I was trying some SIMD for the video scaler and the SSE4 intrinsics wouldn't compile without adding `-march=native` to the meson build flags manually.  I'm getting close to 10x the performance with this so I think it's worth getting these compile flags right :smile: 

_No response_

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
# Description

Remove the SSE 4.2 instruction set from the 80.1-release documentation's hardware requirements.

## Related issues

Fixes #2955 

# Manual testing

# Checklist

I have:

### DIFF
--- a/website/docs/downloads/linux.md
+++ b/website/docs/downloads/linux.md
@@ -110,7 +110,7 @@ You can easily configure your DOS games on
 
 ## Hardware requirements
 
-For x86 CPUs the SSE 4.2 instruction set is required.
+From the x86 family of processors, all x86-64 processors are supported.
 
 ## Development snapshot builds
 
--- a/website/docs/downloads/macos.md
+++ b/website/docs/downloads/macos.md
@@ -13,7 +13,7 @@ sha256: 46a256645255e8345981ea357f1416b8<wbr>ce4bc60a2aba9a86b5122d5075aa7fa9
 </small>
 
 This package is compatible with macOS 10.15 (Catalina) or newer and supports both
-Intel and M1 Macs.
+Intel and Apple silicon Macs.
 
 Check out the [0.80.1 release notes](release-notes/0.80.1.md) to learn about
 the changes and improvements introduced by this release.
@@ -52,10 +52,8 @@ to reproduce some issues).
 
 ## Hardware requirements
 
-For x86 CPUs the SSE 4.2 instruction set is required. For Intel Mac models
-that don't support it, you can try using the partial SSE4.2 emulator
-[MouSSE](https://forums.macrumors.com/threads/mp3-1-others-sse-4-2-emulation-to-enable-amd-metal-driver.2206682/).
-
+This package is compatible with macOS 10.15 (Catalina) or newer and supports
+both 64-bit Intel and Apple silicon Macs.
 
 ## Development snapshot builds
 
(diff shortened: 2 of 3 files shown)


Grounded0 (approved):

johnnovak:
Looks good @kcgen & nice test evidence! 🥇 

I'm wondering if it would be better to rephrase the sentence for macOS a bit. At the top of the Mac downloads page we're stating this:

> This package is compatible with macOS 10.15 (Catalina) or newer and supports both Intel and M1 Macs.

So saying "From the x86 family of processors, all x86-64 processors are supported" is slightly confusing for Mac users because:

- generally, they're not this technical; they understand Intel and M1 only because Apple marketing materials use these names,
- macOS is only supported on Intel (officially), not AMD,
- we also state that the earliest OS version you _might be able_ to run Staging on is Snow Leopard (I doubt it would actually work, or even be usable, but whatever...). Snow Leopard was the last macOS that still supported 32-bit Intel processors.

Given all that, probably this would be better to say:

> Both 64-bit Intel and M1 processors are supported.

Grounded0:
@johnnovak 

Sounds good.

kcgen:
Very succinctly worded; thanks @johnnovak and @Grounded0 .

Will push the revision tomorrow am..

johnnovak:
> Sounds about right. I would rephrase as:
> 
> Both 64-bit Intel and Apple silicon processors are supported.

Right, I'm behind the times 😄 Let's go with this as you suggested:

> Both 64-bit Intel and Apple silicon processors are supported.

@kcgen ☝🏻

kcgen:
all set for round two:

johnnovak (approved):

Grounded0:
I have a remove snap link from linux.md PRs coming once this is deployed.

kcgen:
Thanks for the double-check, @johnnovak.
Will merge.

PR #2966, opened 2023-09-28
Always write the primary config if it doesn't exist

johnnovak:
# Description

Previously, the primary config was only written if it did not exist and no local config was present. This was super confusing and inconsistent behaviour that served no useful purpose whatsoever. It tripped me up several times and I never understood why the default config was written only _sometimes_—until now, but I had to read the code for that...

Additionally, when writing the primary config was triggered, it disregarded the secure mode option which was clearly a bug.

The new behaviour is clear, consistent, and logical:

- If NOT in secure mode (`--securemode` option) and the `--noprimaryconf` option is NOT specified, the primary config always gets written with the default settings if it does not exist (and only then).

- For consistency, whenever DOSBox Staging is started, that can trigger writing the default config (so even `dosbox --version` or `--help` will write the primary config!)

- The presence of a local `dosbox.conf` file in the working directory, or specifying extra config files with the `--conf` option have zero effect on the above behaviour.

Other changes:

- A lot of non-functional refactorings and removal of a bunch of legacy cruft.
- The weird `--opencaptures` option has been removed.
- The `--help` text has been improved.
- The help text of the `CONFIG` command has been improved.
- The `--printconf`, `--editconf`, `--eraseconf`,  and `--erasemapper` commands now perform a lot more validation and provide useful feedback to the user.

# Manual testing

Tested the following on Windows 10 and macOS:

- Confirmed that starting DOSBox Staging normally creates the primary config if `--noprimaryconf` or `--securemode` is NOT specified.

- Deleted the primary config, then confirmed that the following commands return an error message:

	- `dosbox --printconf --noprimaryconf`
	- `dosbox --printconf --securemode`
	- `dosbox --editconf --noprimaryconf`
	- `dosbox --editconf --securemode`
	- `dosbox --eraseconf --noprimaryconf`
	- `dosbox --eraseconf --securemode`

- Deleted the primary config, then confirmed that the following CLI commands create the primary config if `--noprimaryconf` and `--securemode` is NOT specified:

	- `--editconf`,
	- `--erasemapper` 
	- `--help`
	- `--list_glshaders`
	- `--printconf`
	- `--version`

- Confirmed that `--eraseconf` and `--erasemapper` print the path of the deleted file on success.

- Confirmed that `--editconf` still works.

- Verified that portable mode works as expected (e.g., resource subdirs get created in the exe folder, but not in the user's home folder; default mapper file location is in the exe directory `config -wcd` writes the config to the exe directory, etc.)

- Verified that portable mode works the same way when `--securemode` is also specified.

# Checklist

I have:


src/misc/setup.cpp
```diff
@@ -988,7 +988,7 @@ std::string Section_line::GetPropValue(const std::string&) const
 	return NO_SUCH_PROPERTY;
 }
 
-bool Config::PrintConfig(const std::string& filename) const
+bool Config::WriteConfig(const std::string& filename) const
```
weirddan455, line 991: I would rename `filename` to `path` since it wants a full path name.

Maybe it should be a `std_fs::path` too but it just gets passed to `fopen` so you could also just leave it as a string since it doesn't matter much in this use case.

src/misc/setup.cpp
```diff
@@ -1629,32 +1632,20 @@ void SETUP_ParseConfigFiles(const std_fs::path& config_dir)
...
-			control->ParseConfigFile("new primary", config_combined);
-		} else {
-			LOG_WARNING("CONFIG: Unable to write a new primary config file '%s'",
-			            config_combined.c_str());
-		}
-	}
+std_fs::path SETUP_GetPrimaryConfigPath()
+{
+	std::string name;
+	Cross::GetPlatformConfigName(name);
```
weirddan455, line 1647: This is a very dumb function.  Maybe out of scope of this PR but it doesn't do anything besides call another function and use an out parameter.

```
static std::string GetConfigName()
{
	return CANONICAL_PROJECT_NAME ".conf";
}

void Cross::GetPlatformConfigName(std::string &in)
{
	in = GetConfigName();
}
```

Ideally, there would only be one function here rather than 2 and it would return a string rather than using an out parameter.

weirddan455 (commented):
Just a couple comments that were really pre-existing issues.  This change looks fine to me and you tested the things I would have been concerned about.  :+1:

src/misc/setup.cpp
johnnovak: Yeah, those were terrible... 🤦🏻 Nuked the whole thing.

`Cross::CreatePlatformConfigDir` is another thing of beauty... but I have to leave some fun for the next guy 😆

src/misc/setup.cpp
johnnovak: Good point, done.

GranMinigun:
I sense approaching of "I do not want no freaking configs in my `%APPDATA%` in my portable setup!!" comments...

johnnovak:
> I sense approaching of "I do not want no freaking configs in my `%APPDATA%` in my portable setup!!" comments...

In portable setups, you need place the primary `dosbox-staging.conf` config in the same folder where the DOSBox executable resides and that's it. No configs will be created in your home/user folder in that case.

That's one extra step for installing DOSBox for the _first_ time, but people running portable setups are considered power users and requiring them to do a single extra step when creating their portable setup for the first time is a small ask.

johnnovak:
...aaaaaaaand it creates the config in the user folder in portable mode too 😩 Okay, needs more work. It should work as I explained above in portable mode. Good that you raised this @GranMinigun.

weirddan455:
> ...aaaaaaaand it creates the config in the user folder in portable mode too 😩 Okay, needs more work. It should work as I explained above in portable mode. Good that you raised this @GranMinigun.

This is a pre-existing issue and it's more than just the config file.  The shell history file and the capture folder/files also get created all the time I believe.

src/gui/sdlmain.cpp
```diff
@@ -4622,6 +4636,25 @@ int sdl_main(int argc, char *argv[])
...
+	// Write the default primary config if it doesn't exist when not in secure
+	// mode.
+	if (!arguments->securemode) {
+		const auto primary_config_path = SETUP_GetPrimaryConfigPath();
+
+		if (!path_exists(primary_config_path)) {
+			// No config is loaded at this point, so we're writing the default
+			// settings to the primary config.
+			const auto path_string = primary_config_path.string();
+			if (control->WriteConfig(path_string)) {
```
weirddan455, line 4648: WriteConfig should take in `primary_config_path` now that you've changed it to take a `std_fs::path`.  Otherwise it gets implicitly converted from the string.

src/misc/programs.cpp
```diff
@@ -432,13 +432,7 @@ void CONFIG::Run(void)
...
 				return;
 			}
-			std::string name;
-			Cross::GetPlatformConfigName(name);
-
-			// write file to the default config directory
-			const auto config_path = get_platform_config_dir() / name;
-			name = config_path.string();
-			WriteConfig(name);
+			WriteConfig(SETUP_GetPrimaryConfigPath().string());
```
weirddan455, line 435: Same here.  No need for `.string()`

johnnovak:
> > ...aaaaaaaand it creates the config in the user folder in portable mode too 😩 Okay, needs more work. It should work as I explained above in portable mode. Good that you raised this @GranMinigun.
> 
> This is a pre-existing issue and it's more than just the config file. The shell history file and the capture folder/files also get created all the time I believe.

Nah, it works, I'm just tried (move the conf to the wrong folder...) When it's in the exe folder, it even informs you in the logs we're running in portable mode. Which means the `capture`, `glshader`, `mt32-roms`, `soundfonts` folders and the `shell_history.txt` will be all created in the exe folder. Yay! 😄 Will give it tomorrow another round of testing, just to be sure... after sleep 😄 

```
2023-09-28 23:10:09.726 | arguments: /Users/jnovak/jn/dosbox-staging/build/debug/dosbox
2023-09-28 23:10:09.726 | Current dir: /Users/jnovak/Documents
2023-09-28 23:10:09.726 | stderr verbosity: 0
2023-09-28 23:10:09.726 | -----------------------------------
2023-09-28 23:10:09.738 | dosbox-staging version 0.81.0-alpha-1702-g8b5b72483
2023-09-28 23:10:09.738 | ---
2023-09-28 23:10:09.738 | LOG: Loguru version 2.1.0 initialised
2023-09-28 23:10:09.738 | CONFIG: Using portable configuration layout in /Users/jnovak/jn/dosbox-staging/build/debug
2023-09-28 23:10:09.816 | SDL: version 2.28.2 initialised (cocoa video and coreaudio audio)
2023-09-28 23:10:09.818 | CONFIG: Loaded primary config file '/Users/jnovak/jn/dosbox-staging/build/debug/dosbox-staging.conf'
2023-09-28 23:10:09.818 | LANG: Using internal English language messages
```

src/gui/sdlmain.cpp
johnnovak: Not on Windows on MSVC & co. I had to add it back for those because I got all Windows targets failing...

src/misc/programs.cpp
johnnovak: Yeah. Windows 🤷🏻

src/gui/sdlmain.cpp
johnnovak: Oh... ok. I think you're right... really need some sleep...

weirddan455: On Windows, it's a problem when you have a `std_fs::path` and want a C string.  Doing `path.c_str()` can cause compile errors because it returns a UTF16 string.  In that case, you need to do `path.string().c_str()` to make it UTF8.  I've ran into that same thing before in Dosbox and I see you did just that in the call to `fopen`

Here, it's fine to just pass along the path.  Otherwise you're converting to string then back to path (implicitly) for no reason.

johnnovak: Yeah, thanks, I realised eventually, I was just really tired 😄 I'll fix it shortly.

johnnovak: Done.

johnnovak:
Ok, I'm gonna work a bit more on this because there are appalling things going on with the config folder & primary config initialisation when I started digging a bit deeper...

Every single time I want to fix something small, it turns into this refactoring exercise to remove all the insane cruft...

johnnovak:
Okay, I've addressed the comments @weirddan455 plus I've done a bunch of other cleanups and refactorings. I'll stop here—I thought this would be an easy one hour job, but there's so much crazy stuff going on that I could keep cleaning things up for weeks... It's definitely in a better shape now, but there's still a lot to be done.

kcgen:
Looking at this snippet of the criteria:

> Primary config is NOT created when:
> - `--noprimaryconf` or extra configs specified with `--conf` doesn't change the behaviour

If I separate the clauses, does this still have the same meaning?

> Primary config is NOT created when:
- Started with `--noprimaryconf` in a way that doesn't change the behavior
- Started with `--conf` configs that don't change the behavior

---

And if I simplify the wording in this criteria:

> Primary config IS created when:
> - Start with `--noprimaryconf` in a folder with NO local config
> - Start with `--noprimaryconf` in a folder that HAS a local config

To this:

Primary config IS created when:
- Started with `--noprimaryconf`, regardless of the presence of a local config

---

Then I'm finding these statements side-by-side hard to reconcile:

1. Primary config **_is NOT_** created when started with `--noprimaryconf` in a way that doesn't change the behavior

2. Primary config **_IS_** created when started with `--noprimaryconf`, regardless of the presence of a local config

---

If I launch `dosbox --noprimaryconf` in an empty directory, then it satisfies both conditions (all defaults are used, so behavior isn't changed) and there is no local config present. So now the result in conflict or ambiguous (is the primary written or not?).

I can only imagine users having more complex scenarios, and how could we easily explain or figure it out in a logical explanation?

I suggest that the presence of `--noprimaryconf` simply do nothing with it:
 - It never reads it
 - It never writes it

If we do this, the criteria simply becomes: "A primary config is written unless the `--securemode` or `--noprimaryconf` flags are provided".

To me, that's simple and unambiguous - and hopefully would simplify the logic further?

kcgen (commented):
Lots of nice cleanup, @johnnovak! Replacing some of the pre-C++17 filesystem checks with std::FS is very nice, and will let the code easily handle filesystems that we previously didn't support, like network-mount syntax and stranger syntax like morphOS👍

Just one comment (previously added), to try to simplify the criteria.

johnnovak:
> I suggest that the presence of `--noprimaryconf` simply do nothing with it:
> 
>     * It never reads it
> 
>     * It never writes it
> 
> 
> If we do this, all of criteria can be simplified down to this:
> 
> "DOSBox Staging will automatically create your primary config unless the `--securemode` or `--noprimaryconf` flags are provided".
> 
> To me, that's simple and unambiguous.

I was considering this @kcgen, and it's probably for the best. I guess at some point I decided that `nolocalconf` and `noprimaryconf` should only control loading, but not writing, then moved on. But yeah, I like your proposal better, after all.

For the record, when I said "no change in behaviour", I was referring to the behaviour of the previous test cases, not observable DOSBox behaviour. That was ambiguous, in retrospect. Communication is hard 😅  But you came to the right conclusion at the end anyway that we agree on 🎉 😄 

What's your view on these?

- Starting DOSBox Staging with flags that only print stuff to the command line (e.g. `--help`, or `--list-glshaders`) always creates the primary config if it doesn't exist, unless `--noprimaryconf` or `--securemode` is specified.
- Consequently, `--editconf` and `--printconf` both create the primary config, unless `--noprimaryconf` or `--securemode` is specified.

If we agree on the above, I'll add some edge case handling to `--editconf` and `--printconf` (e.g., `--editconf --noprimaryconf` should just exit if the primary config does not exist already).

kcgen:
> .. at the end anyway that we agree on

Yes! 


> What's your view on these? <snip>

Probably we should still auto-create, just to avoid more exceptions.

I like the PR title as you've written, that we *"always write the primary conf"*, unless either of those two flags are given. 

The nice thing with --noprimaryconf is that the user can still manually write the conf themselves with "config -wc" inside dosbox. 

While the secure mode take this a step further and should also block 'config -wc'.  (because in secure mode, we only trust our confs as written on the host, and we seal off the interactive session from making mount changing or writing new conf files and such; secure mode is essentially a kiosk mode where we assume the user at the Z:\> prompt is hostile and trying to gain access to the host or damage the configuration.

johnnovak:
> Yes; I think even for all those flags, we should do as the PR suggests: _"always write the primary conf"_, (and stick to the two exceptions: secure and/or noprimaryconf).

Yeah, I'm approaching it from this angle too (not just this feature, but everything in dev work) — the less exceptions and the more consistent the behaviour, the better 🚀 

Ok, I'll do that plus some further cleanup. So hard to stop, this thing is a mess... 😅

johnnovak:
@kcgen @weirddan455 Addressed feedback plus fixed/improved a couple of other things. I've also substantially rewritten the top description and the manual tests section, so please read it again. From my perspective, this is ready to go.

src/gui/gui_msgs.h
```diff
@@ -37,35 +37,35 @@ constexpr char help_msg[] =
...
-  --editconf               Open the default configuration file in a text editor.
+  --editconf               Open the primary configuration file in a text editor.
 
-  --eraseconf              Delete the default configuration file.
+  --eraseconf              Delete the primary configuration file.
 
-  --noprimaryconf          Don't read settings from the default configuration file
-                           located in your user folder.
+  --noprimaryconf          Don't load the primary configuration file, and don't
+                           write it if it doesn't exist.
```
kcgen, line 51: Suggest:

`Don't load or create the the primary configuration file.`

kcgen (approved):
Another big improvement; looks good @johnnovak.
Just one small suggestion.

src/gui/gui_msgs.h
johnnovak: Haha, generally I'm good at not overcomplicating sentences, but yeah, this is a lot simpler 😄

johnnovak:
> Another big improvement; looks good @johnnovak. Just one small suggestion.

Thanks for the review @kcgen 🎉 Made the help text improvement, just double-checking that all commits still compile, then merge time!

kcgen:
That's a lot of good care you've put into standarding the terms and wording across all the CLI args. Lots of ❤️ going into this. Awesome stuff, @johnnovak.

Torinde:
> "I do not want no freaking configs in my %APPDATA% in my portable setup!!"

OK, let me be that user. I fully agree with that notion - if I run an .exe of a program and NOT of its installer - I expect (hope/wish) "clean" run - preferably no files/folders get created, but if some have to be created - that should be where the .exe is (or its "Start in:" path) and definitely not left over in some far away (and unknown to me) place (like %APPDATA%\Local\Dosbox\)

From what's written further up in this PR - it seems that's the case for glshaders, mt32, soundfonts, shell-history and I even get my wish fully IF there is a .conf file together with the .exe... but on first run there isn't, so a .conf is created in %APPDATA%. Why isn't that created in the .exe location instead?

On the other hand - as johnnovak writes above - "power users" can handle that, so although I don't like it, I also don't think it's a priority. Just sharing my opinion:
- portable mode should not touch %APPDATA%, ever
- on first run .conf/other files and folders should be created where the .exe is
- if user wants stuff present in %APPDATA% they should use the installer (maybe I'm missing something here and there is a reason to have such quasi-installation on first run, as convenience for _other_ types of new users... or for Linux/macOS or for something else...)

johnnovak:
Portable mode is turned on by `dosbox-staging.conf` existing in the exe directory, that's all there is to it. You can make make an "installed" DOSBox "portable" by putting the primary config in the exe dir (and vice versa)

One way to make the portable ZIP truly portable is to include the primary config in the ZIP. But then people upgrading by unzipping the new release into their existing installation would overwrite their existing primary config (unless they're careful not to).

We could introduce a `--portable` command line option, but thar's easy to forget about so I don't like that idea.

We could implement an _additional_ trigger to enter portable mode if a file called `.portable-mode` exists in the exe folder. So if that file _or_ the primary config exists in the exe folder, we're in portable mode!

That would solve the start first without a primary conf in portable mode problem in an unobtrusive way. We'd only need to start including `.portable-mode` in the portable release ZIP.

Thoughts? @kcgen @weirddan455 @Kappa971

Kappa971:
> Portable mode is turned on by `dosbox-staging.conf` existing in the exe directory, that's all there is to it. You can make make an "installed" DOSBox "portable" by putting the primary config in the exe dir (and vice versa).
> 
> One way to make the portable ZIP _truly_ portable is to include the primary config in the ZIP. But then people upgrading by unzipping the new release into their existing installation would overwrite their existing primary config (unless they're careful not to).
> 
> We could introduce a `--portable` command line option, but that's easy to forget about so I don't like that idea.
> 
> We could implement an _additional_ trigger to enter portable mode if a file called `.portable-mode` exists in the exe folder. So if that file _or_ the primary config exists in the exe folder, we're in portable mode!
> 
> That would solve the "start first without a primary conf in portable mode" problem in an unobtrusive way. We'd only need to start including `.portable-mode` in the portable release ZIP.
> 
> Thoughts? @kcgen @weirddan455 @Kappa971

In Duckstation (PS1 emulator), to switch to portable mode you can simply place a "portable.txt" file in the executable folder.
This moves all paths (covers, bios, etc.) from Documents into the executable folder.
Regarding DOSBox, this mode should recreate all the folders that DOSBox creates in APPDATA (capture, glshaders, mt32-roms, soundfonts) and the files (dosbox-staging.conf, shell_history.txt) in the DOSBox executable folder.

weirddan455:
I really don't care about a "portable mode" feature personally.  `%APPDATA%` is the correct place for modern applications to create and modify config/log/cache files they need at runtime.  Dumping files wherever the .exe happens to lie is legacy behavior from a time before Windows cared about security where all users and programs had full read/write to `C:\`.

On Linux, where normal users were always restricted by permissions to the root filesystem, pretty much every program you run creates files in `~/.config` and/or `~/.cache`.  It's rare to find a program supporting "portable mode" because the normal case is that the executable lives in a directory the user doesn't have write access to.

I'm also not sure what the use case is.  Are people storing Dosbox on a USB or network drive and running it from multiple computers (and also wanting to use the same config)?  If so, we kind of already support that use case by placing `dosbox-staging.conf` in the same folder like @johnnovak pointed out.

johnnovak:
Yeah, I care a _little bit_ more about portable mode than @weirddan455, but I'm 90% aligned with his views.

So we have 2 options:

- Do nothing – portable people put `dosbox-staging.conf` into the the exe folder on the very first run. Frankly, the best solution.
- We introduce the `.portable-mode` or `portable.txt` additional portable mode trigger. Requires some work to support it, plus to mess around with CI to put the file into the packaged up ZIP file. It's not gonna be, but if someone's absolutely dying to get this done, I won't stop him.

Regarding moving files in and out between the APP_DATA and the exe folder—that's a categorical NO, we won't do that shit ever. I couldn't care less what DuckStation does.

My vote is "do nothing", just document it. The second option is up for grabs, won't be me.

Kappa971:
> I really don't care about a "portable mode" feature personally. `%APPDATA%` is the correct place for modern applications to create and modify config/log/cache files they need at runtime. Dumping files wherever the .exe happens to lie is legacy behavior from a time before Windows cared about security where all users and programs had full read/write to `C:\`.
> 
> On Linux, where normal users were always restricted by permissions to the root filesystem, pretty much every program you run creates files in `~/.config` and/or `~/.cache`. It's rare to find a program supporting "portable mode" because the normal case is that the executable lives in a directory the user doesn't have write access to.
> 
> I'm also not sure what the use case is. Are people storing Dosbox on a USB or network drive and running it from multiple computers (and also wanting to use the same config)? If so, we kind of already support that use case by placing `dosbox-staging.conf` in the same folder like @johnnovak pointed out.

I don't agree with this. Linux is Linux and Windows is Windows. Having a portable mode in Windows is very useful because you can install the emulator wherever you want and have everything in one folder, this is my opinion.

> Regarding moving files in and out between the APP_DATA and the exe folder—that's a categorical no, we won't do that shit ever. I couldn't care less what DuckStation does.

Duckstation doesn't move any files, it just changes all the paths and recreates the files/folders in the executable folder instead of APPDATA, so you have everything in the same folder (portable mode, right? 😅)

johnnovak:
> Duckstation doesn't move any files, it just changes all the paths and recreates the files/folders in the executable folder instead of APPDATA, so you have everything in the same folder (portable mode, right? 😅)

Ok, you should improve your wording a bit then. Then I'm fine with it as long as somebody's up for the job.

johnnovak:
If you figure out the CI stuff @Kappa971, I'm doing the code change. Deal? 😄

Kappa971:
> Ok, you should improve your wording a bit then. Then I'm fine with it as long as somebody's up for the job. Updated my summary above.

I make some mistakes, English is not my language so I apologize (if you want I write in Italian 😄).

> If you figure out the CI stuff @Kappa971, I'm doing the code change. Deal? 😄

What should I do? Other than creating branches and editing files with the GitHub GUI, I don't currently know how to do anything else 😅

weirddan455:
>  I don't agree with this. Linux is Linux and Windows is Windows. Having a portable mode in Windows is very useful because you can install the emulator wherever you want and have everything in one folder, this is my opinion.

"install the emulator wherever you want" - No, it would have to be installed somewhere you have write access to (so not in `C:\Program Files` where the majority of programs live).  Normal use case is you must be Administrator to install the program to a system wide location like that and then any user with lesser permissions can run it.  When Microsoft rolled out real user permissions (mirroring what Unix had long been doing), they created the `%APPDATA%` folders (also mirroring what Unix had long been doing).

I also don't think this is a platform specific issue.  If we want to have a portable mode, it should function the same no matter what OS you're on.  The problem with doing so is also the same on every OS:  You might not have write access to the place where the .exe lives.

johnnovak:
> > Ok, you should improve your wording a bit then. Then I'm fine with it as long as somebody's up for the job. Updated my summary above.
> 
> I make some mistakes, English is not my language so I apologize (if you want I write in Italian 😄).
> 
> > If you figure out the CI stuff @Kappa971, I'm doing the code change. Deal? 😄
> 
> What should I do? Other than creating branches and editing files with the GitHub GUI, I don't currently know how to do anything else 😅

Ok, let's leave it at that then 😄 Explaining things would create extra work for me, and in fact I wanted help. I might look into this sometime with a low priority once all other release related tasks are completed. Can't be that hard, to be honest...

johnnovak:
> I also don't think this is a platform specific issue. If we want to have a portable mode, it should function the same no matter what OS you're on. The problem with doing so is also the same on every OS: You might not have write access to the place where the .exe lives.

Yeah, but that's not really a problem. E.g., I'm running NeoVim from `~/bin/nvim` and look at them permissions 😎 

image

It's expert mode, if attempt to run portable mode from a directory you have no full write access to, you lose your warranty, pal 😎 

Otherwise yeah, we're in agreement about the backstory...

Kappa971:
> "install the emulator wherever you want" - No, it would have to be installed somewhere you have write access to (so not in `C:\Program Files` where the majority of programs live). Normal use case is you must be Administrator to install the program to a system wide location like that and then any user with lesser permissions can run it. When Microsoft rolled out real user permissions (mirroring what Unix had long been doing), they created the `%APPDATA%` folders (also mirroring what Unix had long been doing).
> 
> I also don't think this is a platform specific issue. If we want to have a portable mode, it should function the same no matter what OS you're on. The problem with doing so is also the same on every OS: You might not have write access to the place where the .exe lives.

The thing is simple, let's assume that DOSBox Staging already has portable mode, there are:
1. The installer (or the Linux package manager) that installs the emulator in `C:\Program Files` (or I don't know where in Linux), and DOSBox works in normal mode;
2. Zip package with DOSBox in portable mode.

If someone decides to put the DOSBox files (in portable mode) in a folder with write permissions, well... in Italian in a "friendly" sense we say that that person is "ciuccio" o "asino" (donkey) 😆

johnnovak:
> > "install the emulator wherever you want" - No, it would have to be installed somewhere you have write access to (so not in `C:\Program Files` where the majority of programs live). Normal use case is you must be Administrator to install the program to a system wide location like that and then any user with lesser permissions can run it. When Microsoft rolled out real user permissions (mirroring what Unix had long been doing), they created the `%APPDATA%` folders (also mirroring what Unix had long been doing).
> > I also don't think this is a platform specific issue. If we want to have a portable mode, it should function the same no matter what OS you're on. The problem with doing so is also the same on every OS: You might not have write access to the place where the .exe lives.
> 
> The thing is simple, let's assume that DOSBox Staging already has portable mode, there are:
> 
>     1. The installer (or the Linux package manager) that installs the emulator in `C:\Program Files` (or I don't know where in Linux), and DOSBox works in normal mode;
> 
>     2. Zip package with DOSBox in portable mode.
> 
> 
> If someone decides to put the DOSBox files (in portable mode) in a folder without write permissions, well... in Italian in a friendly sense we say that that person is "ciuccio" o "asino" (donkey) 😆, so I don't see what the problem is.
> 
> Obviously take everything I wrote as my opinion. If there is no portable mode, I will continue to live anyway 😄

Ok... this gave me pause... Do you understand do we _do_ have portable mode, right?

1. Download ZIP, unpack zip to `D:\DOSBox`
2. Run dosbox.exe -> stuff gets created in APP_DATA
3. Move `dosbox-staging.conf` from APP_DATA into `D:\DOSBox`
4. Bamm, portable mode! If `dosbox-staging.conf` exists in the exe folder, portable mode is turned on.

We're only talking about the primary config and the resource dirs being created in APP_DATE for the very first launch.

kcgen:
There are some game developers who bundle DOSBox Staging directly inside the same directory as their game files, very similar to GoG (yes, they're typically Windows-only packages).

I think that's how 'Zoom Platform' is also packaging their for-pay DOS games (which all include Staging).

Users just want their games to run and don't care about lots of duplicate copies of DOSBox inside each directory.  Likewise.. these commercial platforms packagers (like GoG and Zoom) like to set stuff up once, and then walk away from it. So for them, having a known-working version + conf is perfect for them.

Not that I care about commercial platforms like Zoom and GoG or want anyone wasting time "supporting" them for free; but at least their use cases are good examples of how portable mode is used.

johnnovak:
> There are some game developers who bundle DOSBox Staging directly inside the same directory as their game files, very similar to GoG (yes, they're typically Windows-only packages).
> 
> I think that's how 'Zoom Platform' is also packaging their for-pay DOS games (which all include Staging).
> 
> Users just want their games to run and don't care about lots of duplicate copies of DOSBox inside each directory. Likewise.. these commercial platforms packagers (like GoG and Zoom) like to set stuff up once, and then walk away from it. So for them, having a known-working version + conf is perfect for them.
> 
> Not that I care about commercial platforms like Zoom and GoG or want anyone wasting time "supporting" them for free; but at least their use cases are good examples of how portable mode is used.

Absolutely. Plus I wanna keep using my workflow of having everything selfcontained in `D:\Emulators`. All emulators I use can run in "self-contained" ("portable") mode, and I like it that way. I'm not running a server OS, I wanna play games, m'kay... :trollface:

Kappa971:
> Ok... this gave me pause... Do you understand we _do_ have portable mode, right?
> 
>     1. Download ZIP, unpack zip to `D:\DOSBox`
> 
>     2. Run dosbox.exe -> stuff gets created in APP_DATA
> 
>     3. Move `dosbox-staging.conf` from APP_DATA into `D:\DOSBox`
> 
>     4. Bamm, portable mode! If `dosbox-staging.conf` exists in the exe folder, portable mode is turned on.
> 
> 
> We're only talking about the primary config and the resource dirs being created in APP_DATA for the _very first launch_! If you already had `dosbox-staging.conf` in your exe folder _before the first launch_, nothing ever would get created in APP_DATA, do you understand this part?

So, for example, the mt-32 roms folder is recreated in the executable folder? If yes then we can talk about portable mode, otherwise no because part of the resources remain in APPDATA, I have to try.
I was thinking, in addition to the existing zip package, you could create a second zip package in portable mode (instead of replacing the existing one) with a portable.txt file or something to avoid the dosbox-staging.conf file being overwritten with a new version (if portable package doesn't use portable.txt).

johnnovak:
> EDIT
> I moved `dosbox-staging.conf` to the executable folder and all the folders are created too, so yes, portable mode is already there. However it is a bit inconvenient because you have to start DOSBox (the first time) so that it can create the files in APPDATA and move the configuration file manually. Also, as mentioned, if you will create a portable zip package, the included `dosbox-staging.conf` file may replace an existing one (if you are upgrading).
> With a separate file like `portable.txt`, you would start DOSBox for the first time and you would already have everything in the executable folder.

Man... What I told you this morning already, yes? 😄 How about trying things first before writing your walls of text? 😅

Kappa971:
> Man... What I told you this morning already, yes? 😄 How about trying things first before writing your walls of text? 😅
> 
> Plus we _told you_ a few times that we _have_ portable mode... It just requires this minor optional tweak to make it a little better for first time users. I made that clear a few posts ago. Don't you read what we write here? 😄
> 
> What is this in my original manual testing steps? References to the **_portable mode_**! 😅
> image

I knew it was possible to have the DOSBox configuration file in the same folder as the executable (I thought that's what you meant in the message above) but I didn't realize it was 100% portable mode.

> You could have spared yourself (and us!) a lot of effort if you read things carefully first before acting...

Oh no please, this seems to have forced you to look for gold in a mine 😆
My final opinion is to add a file that makes DOSBox start in portable mode from the first boot, and document it everywhere (GitHub page, main site page, etc. in upper case! 😄) case closed. You choose whether to do it or not.

johnnovak:
> My final opinion is to add a file that makes DOSBox start in portable mode from the first boot, and document it everywhere (GitHub page, main site page, etc. in upper case! 😄) case closed. You choose whether to do it or not.

Stop this "case closed", "you do what you want, I'm out", "end of discussion" bullshit or I'm afraid I'll ignore your comment from now on.

Kappa971:
> Stop this "case closed", "you do what you want, I'm out", "end of discussion" bullshit or I'm afraid I'll ignore your comments from now on.
> 
> And learn to read.

Case closed for me, I have nothing else to say. You asked me for an opinion and I answered.
You should learn to respond in a less hostile way, because it seems like I insulted you. Maybe you're used to writing this way with other people but it bothers me a lot and if so far I haven't told you to go to hell it's only out of good manners. If you want to ignore my comments, do as you want (you're the one who asked for my opinion).

Torinde:
I'm sorry to see people annoyed like that above. I think both of you meant well to each other (but unfortunately misunderstood each other).

From the above I see the following scenarios in Windows:
- NEW user downloads INSTALLER, runs it - no controversy here, GUI to choose installation location, etc.
- EXISTING user downloads INSTALLER, runs it - no controversy here, GUI to choose installation location to be upgraded, existing .conf is kept (actually I haven't tested if Installer "upgrades" it via `config -wc` equivalent - as I would expect. Or overwrites %APPDATA% .conf - which I think isn't good), etc.
- NEW user downloads ZIP, unzips it where they want/can, runs the .exe, .conf (and more) will be created in %APPDATA% - that I find strange/objectionable. **Why isn't that created in the .exe location instead?**
- EXISTING user downloads ZIP, unzips it in the EXISTING folder, thus overwriting the .exe and having a leftover existing .conf (and more), runs the .exe and then `config -wc`

I can't comment on how it looks for Linux/macOS users... is it because they don't have write access to the .exe location that by default, if missing, .conf gets created in %APPDATA% (and its Linux/macOS equivalents)? If so, I suggest adding a check - on first run, if .conf (or shell_history or mt32, etc.) is missing, **try creating it in the .exe location and if not possible, then use %APPDATA% (and its Linux/macOS equivalents).**

Adding portable.txt file to the ZIP seems to me like overcomplicating and wasted effort. Also, existing non-portable users who upgrade via unzipping in their folder would have to delete that file or logic has to be created to ignore it or maybe no need to ignore it, but it'll be confusing (e.g. "why do I have a _portable.txt_ in my Program files _non portable installation_?").

My personal workflow is often to use ZIP file, but unzip it in NEW folder (to keep multiple versions), while usually preferring to use default config for each new version - so I don't like when I have a forgotten/unknown .conf in %APPDATA%, because I forgot to take care of that auto-creation at some moment. But as I wrote above - topic is minor and if there is a reason to keep the current (strange for me) operation, then fine, I'm just trying to understand what's the logic/use-case for it.

Kappa971:
> I'm sorry to see people annoyed like that above. I think both of you meant well to each other (but unfortunately misunderstood each other).
> 
> From the above I see the following scenarios in Windows:
> 
>     * NEW user downloads INSTALLER, runs it - no controversy here, GUI to choose installation location, etc.
> 
>     * EXISTING user downloads INSTALLER, runs it - no controversy here, GUI to choose installation location to be upgraded, existing .conf is kept (actually I haven't tested if Installer "upgrades" it via `config -wc` equivalent - as I would expect. Or overwrites %APPDATA% .conf - which I think isn't good), etc.
> 
>     * NEW user downloads ZIP, unzips it where they want/can, runs the .exe, .conf (and more) will be created in %APPDATA% - that I find strange/objectionable. **Why isn't that created in the .exe location instead?**
> 
>     * EXISTING user downloads ZIP, unzips it in the EXISTING folder, thus overwriting the .exe and having a leftover existing .conf (and more), runs the .exe and then `config -wc`
> 
> 
> I can't comment on how it looks for Linux/macOS users... is it because they don't have write access to the .exe location that by default, if missing, .conf gets created in %APPDATA% (and its Linux/macOS equivalents)? If so, I suggest adding a check - on first run, if .conf (or shell_history or mt32, etc.) is missing, **try creating it in the .exe location and if not possible, then use %APPDATA% (and its Linux/macOS equivalents).**
> 
> Adding portable.txt file to the ZIP seems to me like overcomplicating and wasted effort. Also, existing non-portable users who upgrade via unzipping in their folder would have to delete that file or logic has to be created to ignore it or maybe no need to ignore it, but it'll be confusing (e.g. "why do I have a _portable.txt_ in my Program files _non portable installation_?").
> 
> My personal workflow is often to use ZIP file, but unzip it in NEW folder (to keep multiple versions), while usually preferring to use default config for each new version - so I don't like when I have a forgotten/unknown .conf in %APPDATA%, because I forgot to take care of that auto-creation at some moment. But as I wrote above - topic is minor and if there is a reason to keep the current (strange for me) operation, then fine, I'm just trying to understand what's the logic/use-case for it.

- Installer - DOSBox in normal mode (with files in APPDATA)
- Zip package - DOSBox in normal mode (with files in APPDATA)
- Zip package and you create `portable.txt` file - DOSBox in portable mode
I think this would be a good compromise.

weirddan455:
@Torinde There's not much of a difference in terms of functionality between the installer and the .zip.  I'm not booted into Windows at the moment to test but as I understand it, the Windows installer basically just unzips the files plus creates a shortcut and adds an entry to "Add/Remove Programs" so the user can run the un-install later.  It shouldn't create or modify anything in `%APPDATA&`.  That's done at first run.

>  try creating it in the .exe location and if not possible, then use %APPDATA% (and its Linux/macOS equivalents)

I don't think we should be making portable mode the default behavior.

The way to enable portable mode is to create a `dosbox-staging.conf` file in the same directory as the .exe.  I tested and this works.  No files get created in `~/.config` if this happens.  So rather than fussing with some separate `portable.txt` file, you can simply create an empty `dosbox-staging.conf`.

The only issue is, the config file remains empty after first run.  I propose detecting if the config file is empty.  If so, regenerate it.  We can later expand this to do something sensible if the config file is corrupted or outdated but for this use case, handling the empty case is enough.

johnnovak:
> You should learn to respond in a less hostile way, because it seems like I insulted you.

_Absolutely zero_ idea why that might be case? Like nothing? As if you were not harassing us on Discord for weeks after _we told you to stop_... 🤦🏻‍♂️ Don't play the innocent. If I were you, I'd try _super hard_ from now not to annoy the staff here...

---

That's not a bad idea @weirddan455 , I like it better than the `portable.txt` solution.

Issue #2663, opened 2023-07-06
Refactor DOS file attributes to use FatAttributeFlags


weirddan455:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Linux x86_64

### Is your feature request related to a problem? Please describe.

There are these bit-fields defined here for DOS file attributes:  https://github.com/dosbox-staging/dosbox-staging/blob/d64ad72581e08e0c48b9dcc9d5da7703f03fb8b8/include/dos_system.h#L44-L75

The top one (marked obsolete) used plain old C style bitwise operations to make a bitfield.  It is still being used lots of places but we should probably start refactoring these to use the fancy new C++ bit_view API.

### Describe the solution you'd like

Grep for uses of the enum marked obsolete and replace them.  Or just delete them from the header file and check for compile errors.  Some of the functions that use these functions are a little cludgy so be warned.  `DOS_FindFirst` for example takes in a `uint16_t` while it should really be a `uint8_t` (or even better, refactor it to take in a `FatAttributeFlags` union directly.)

Similarly, `DOS_GetFileAttr` should probably return one of these instead (also currently returning a `uint16_t` for some reason).

There was some discussion about this in #2660

### Describe alternatives you've considered

Haven't considered alternatives but if anyone has any ideas to make these interfaces less cludgy, feel free to discuss.

### Add any other context or screenshots about the feature request here.

@johnnovak made the compiler very unhappy when he tried touching these things

https://user-images.githubusercontent.com/698770/251313955-2c4014e3-9da2-4a6a-bcb7-02fcd16d268e.png

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Thanks for raising @weirddan455 and including all relevant context! 🎉

johnnovak:
Changed the label to "plumbing" @weirddan455 because we use "enhancement" for actual user-observable feature enhancements/improvements. Code-only improvements only get the unglamorous "plumbing" label. Hope that doesn't cause disappointment! 😆

PR #2976, opened 2023-10-07
[WIP] Remove obsolete APIs from 'dos_system.h'

FeralChild64:
# Description

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2663

# Manual testing

# Checklist

I have:


include/dos_system.h
```diff
@@ -45,17 +45,6 @@ constexpr auto CurrentDirectory = ".";
...
-	DOS_ATTR_HIDDEN    = 0b000'0010,
-	DOS_ATTR_SYSTEM    = 0b000'0100,
-	DOS_ATTR_VOLUME    = 0b000'1000,
-	DOS_ATTR_DIRECTORY = 0b001'0000,
-	DOS_ATTR_ARCHIVE   = 0b010'0000,
-	DOS_ATTR_DEVICE    = 0b100'0000,
-};
-
 union FatAttributeFlags {
 	uint8_t _data = 0;
```
kcgen, line 49: (just a minor suggestion, if you think it's better)

One of the nice things with enums is that the definition is a "spaceless" logical construct, without being real objects.

So replacing this `enum` with the global objects is kind of trade off - given (I think) we would still /like/ to use them as enums, but unfortunately C++ doesn't allow non-scalar enums.

Ideally we'd have:

```c++
enum class FatAttributes : FatAttributeFlags {
    ReadOnly = { 0b000'0001 },
    Hidden   = { 0b000'0010 },
    System   = { 0b000'0100 },
    Archive  = { 0b000'1000 },
    ...
};
```

One alternative (that get's us a bit closer) is to make a scoped enum inside the union:

```c++
union FatAttributeFlags {
	enum : uint8_t {
		ReadOnly  = bit::literals::b0,
		Hidden    = bit::literals::b1,
		System    = bit::literals::b2,
		Volume    = bit::literals::b3,
		Directory = bit::literals::b4,
		Archive   = bit::literals::b5,
		Device    = bit::literals::b6,
		NotVolume = bit::mask_flip_all(Volume),
	};
    ...
};
```

This eliminates the global objects and it's strongly scope  just like a class-enum.

The tradeoff is we still need to use scalars (uint8_t) in enum, which matches the `_data` type.

This allows for a bit more natural comparisons when dealing with the uint8_t and uint16_t `"attr"` and `"attrib"` variables, and we can still assign and pass them into functions that take `FatAttributeFlags` objects.

kcgen: @FeralChild64 , I've pushed a branch off your branch to show it in action, if you'd like to cherry-pick:

Here's the change for you to review: 1570dee6d

The branch `kc/fc-attr-improvements-scoped-enum-1` has two commits, if you want to git cherry-pick them:

`git cherry-pick ac60fb05d` // white-space only
`git cherry-pick 1570dee6d` // scoped-enum

kcgen (approved):
Another big cleanup; looks great @FeralChild64 !

Just one minor suggestion.

src/dos/drive_fat.cpp
```diff
@@ -1284,23 +1306,31 @@ bool fatDrive::FindNextInternal(uint32_t dirClustNumber, DOS_DTA &dta, direntry
...
+		if (!(sectbuf[entryoffset].attrib & FatAttributeVolume._data)) {
+			goto nextfile;
+		}
 		dirCache.SetLabel(find_name, false, true);
 	} else {
-		if (~attrs & sectbuf[entryoffset].attrib & (DOS_ATTR_DIRECTORY | DOS_ATTR_VOLUME | DOS_ATTR_SYSTEM | DOS_ATTR_HIDDEN) ) goto nextfile;
+		if (~(attrs._data) & sectbuf[entryoffset].attrib & attr_mask._data) {
+			goto nextfile;
+		}
 	}
```
kcgen, line 1332: I think we can levrage FatAttributeFlags to more directly test these volume, directory, etc.. bits instead of using bit-wise operations.

Push an example here: a17542b37

johnnovak: That would be preferable, yes. We should try to get as much use out of named bitfields as we can; it greatly enhances readability and reduces the likelihood of bit-twiddling bugs.

src/dos/dos_files.cpp
```diff
@@ -385,16 +385,17 @@ bool DOS_Rename(const char* const oldname, const char* const newname)
...
+	LOG(LOG_FILES, LOG_NORMAL)
+	("file search attributes %X name %s", attr._data, search);
 	DOS_DTA dta(dos.dta());
 	uint8_t drive;char fullsearch[DOS_PATHLENGTH];
 	char dir[DOS_PATHLENGTH];char pattern[DOS_PATHLENGTH];
 	size_t len = strlen(search);
 
 	const bool is_root = (len > 2) && (search[len - 2] == ':') &&
-	                     (attr == DOS_ATTR_VOLUME);
+	                     (attr == FatAttributeVolume);
```
johnnovak, line 398: Why can't we just check for `attr.volume`?

src/dos/drive_fat.cpp
```diff
@@ -1174,15 +1194,15 @@ bool fatDrive::FindFirst(char *_dir, DOS_DTA &dta,bool /*fcb_findfirst*/) {
...
 		if (strcmp(GetLabel(), "") == 0 ) {
 			DOS_SetError(DOSERR_NO_MORE_FILES);
 			return false;
 		}
-		dta.SetResult(GetLabel(),0,0,0,DOS_ATTR_VOLUME);
+		dta.SetResult(GetLabel(),0,0,0,FatAttributeVolume);
 		return true;
 	}
-	if(attr & DOS_ATTR_VOLUME) //check for root dir or fcb_findfirst
+	if(attr & FatAttributeVolume) //check for root dir or fcb_findfirst
```
johnnovak, line 1205: `attr.volume`?

src/dos/drive_local.cpp
```diff
@@ -331,32 +331,33 @@ bool localDrive::FindFirst(char* _dir, DOS_DTA& dta, bool fcb_findfirst)
 	}
 	safe_strcpy(srchInfo[id].srch_dir, tempDir);
 	dta.SetDirID(id);
-	
-	uint8_t sAttr;
+
+	FatAttributeFlags sAttr = {};
```
johnnovak, line 335: Maybe rename to `attrs` or something while you're at it.

src/dos/drive_overlay.cpp
```diff
@@ -814,8 +820,8 @@ bool Overlay_Drive::FindNext(DOS_DTA & dta) {
 	char full_name[CROSS_LEN];
 	char dir_entcopy[CROSS_LEN];
 
-	uint8_t srch_attr;char srch_pattern[DOS_NAMELENGTH_ASCII];
-	uint8_t find_attr;
+	char srch_pattern[DOS_NAMELENGTH_ASCII];
+	FatAttributeFlags srch_attr = {};
```
johnnovak, line 824: `search_attrs`?

include/dos_system.h
johnnovak: Very nice @kcgen, I think that's an improvement over the const approach, and it's nicely namescaped. Good creative use of namespace nesting! 🏅

johnnovak (approved):
Another monumental work @FeralChild64 , very nice! Just minor comments, plus I have a preference for @kcgen's scoped enum improvement, so consider adding that in. Otherwise, feel free to merge whenever you're ready.

I don't have time to test this, but I believe you gave it a thorough testing already anyway.

include/dos_system.h
FeralChild64: OK, makes sense. Cherry-picket the commit.

src/dos/drive_fat.cpp
FeralChild64: OK, cherry-picked your change.

src/dos/drive_fat.cpp
FeralChild64: Actually, `FatAttributeFlags(attr).volume` - already changed by @kcgen

src/dos/drive_overlay.cpp
FeralChild64: OK, changed.

src/dos/drive_local.cpp
```diff
@@ -371,7 +372,7 @@ bool localDrive::FindNext(DOS_DTA& dta)
 	char full_name[CROSS_LEN];
 	char dir_entcopy[CROSS_LEN];
 
-	uint8_t srch_attr;
+	FatAttributeFlags srch_attr = {};
```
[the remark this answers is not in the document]
FeralChild64: Renamed to `search_attr`, as used everywhere else.

src/dos/drive_fat.cpp
```diff
@@ -1306,27 +1306,31 @@ bool fatDrive::FindNextInternal(uint32_t dirClustNumber, DOS_DTA& dta,
...
 	if (attrs == FatAttributeFlags::Volume) {
-		if (!(sectbuf[entryoffset].attrib & FatAttributeFlags::Volume)) {
+		if (!entry_attributes.volume) {
 			goto nextfile;
 		}
 		dirCache.SetLabel(find_name, false, true);
 	} else {
-		if (~(attrs._data) & sectbuf[entryoffset].attrib & attr_mask._data) {
+		// Compare attributes to search attributes
+		// Compare attributes to search attributes
```
kcgen, line 1326: (repeating comment; Sorry for this - probably a typo on my side)

src/dos/dos_files.cpp
FeralChild64: Seems to be safe. Changed, but in a separate commit, for better bisectability in case of problems.

src/dos/drive_fat.cpp
FeralChild64: OK, fixed.

src/dos/dos_files.cpp
```diff
@@ -394,8 +394,7 @@ bool DOS_FindFirst(const char* search, FatAttributeFlags attr, bool fcb_findfirs
 	char dir[DOS_PATHLENGTH];char pattern[DOS_PATHLENGTH];
 	size_t len = strlen(search);
 
-	const bool is_root = (len > 2) && (search[len - 2] == ':') &&
-	                     (attr == FatAttributeFlags::Volume);
+	const bool is_root = attr.volume && (len > 2) && (search[len - 2] == ':');
```
kcgen: Just a question:

`attr == FatAttributeFlags::Volume`, is true if the all the other bits are false.

Where as the new `attr.volume` doesn't care about the other (non-volume) bits. 

So for example, A "volume" device that's marked system and hidden would fail in the prior equality check, but would pass in the replacement.

I /think/ the second is more correct - but just mentioning it.

kcgen: @FeralChild64, there are more of these comparisons that could be fixed (by making the only check the desired bit): `git grep "== FatAttributeFlags"` will reveal them.

FeralChild64: The problem is, the remaining code looks like this:

```C++
	if (attr == FatAttributeFlags::Volume) {
		dta.SetResult(discLabel, 0, 0, 0, FatAttributeFlags::Volume);
		return true;
	} else if (attr.volume && isRoot && !fcb_findfirst) {
		// something else
```

so I suppose the `attr == FatAttributeFlags::Volume` check was done this way on purpose. I start having doubts about change in `dos_files.cpp`- but I can't find any way to make the code misbehave.

kcgen: > I start having doubts about change in dos_files.cpp- but I can't find any way to make the code misbehave.

Good call; I guess if we knew how the original code operated, we could adjust these bit checks a bit closer.. but then, things are working very solid in DOSBox, so likely the original authors nailed it.

FeralChild64: I have reverted the change to `is_root` calculation - `attr` is actually a parameter received from the guest application, and the whole thing is some kind of hack. This should be cleaned up some day, but for now I consider it too risky.

Issue #2984, opened 2023-10-10
Alone in the Dark - broken text rendering


FeralChild64:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

current GIT main, e755b9e53042b2fdb0bf0917b128422c15216c28

### What Operating System are you using?

Linux x86_64

### If Other OS, please describe

Gentoo

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When using dynamic CPU core, the main menu of the game gets corrupted once you move the cursor:

![bug](https://github.com/dosbox-staging/dosbox-staging/assets/48332137/20cdb22a-a776-4c3b-83d6-6779e2cd4450)

The problem only occurs on current GIT main; on 0.80.1 the game works correctly.

### Steps to reproduce the behaviour.

1. Set `core = dynamic` (with `core = normal` the problem does not occur)
2. Start _Alone in the Dark_ game (I have used the CD version)
3. In the main menu press cursor up/down few times to move the highlight
4. The menu text gets broken; if you start the game all the in-game texts are broken in a similar way

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

kcgen:
Great find, @FeralChild64.

Interesting.. `core = auto` is OK, but explicitly setting `dynamic` triggers it.

Also confirmed in the ~5.5 MB floppy edition.  Bisecting with:

```ini
[cpu]
core = dynamic
cycles = 10000

[autoexec]
mount c .
c:
autotype -w 0.2 enter , esc , down up down
TATOU.COM
```

kcgen:
Can't repro on ARM64.

weirddan455:
Bisected to fa4c535efe99e60a004c1053e6b1be4c75549f55 as the first bad commit.

Note for future bisections:  I had to bisect with release builds for this because many of the commits were crashing with a SIGABRT (some iterator compare error) when attempting to start the game preventing me from testing this bug.  Release builds were unaffected.

johnnovak:
Can reproduce reliably on Windows with this config and Soundblaster Pro sound using the eXoDOS v5 version:

```ini
[cpu]
cycles = 6000
core = dynamic
```

[...]

PR #2988, opened 2023-10-11
Fix the cache's write mask grow call with a `std::vector` (#2984)

kcgen:
# Description

Fix the regression in #2984 using a `std::vector` instead of manually managing of the cache's mask (which has a buggy grow implementation), which also eliminates the need for the mask's length member variable.

## Related issues

#2984

# Manual testing

Tested Alone in the Dark using on x86-64 with the dynamic core using sanitizer builds:

```ini
[cpu]
core = dynamic
cycles = 10000

[autoexec]
mount c .
c:
autotype enter , esc , down up down
TATOU.COM
```
# Checklist

I have:

### DIFF
--- a/src/cpu/dyn_cache.h
+++ b/src/cpu/dyn_cache.h
@@ -23,6 +23,7 @@
 #include <cerrno>
 #include <new>
 #include <type_traits>
+#include <vector>
 
 #include "mem_unaligned.h"
 #include "paging.h"
@@ -89,9 +90,8 @@ public:
 		CacheBlock* next = {};
 		// writemap masking maskpointer/start/length to allow holes in
 		// the writemap
-		uint8_t* wmapmask = {};
+		std::vector<uint8_t>* wmapmask = nullptr;
 		uint16_t maskstart = 0;
-		uint16_t masklen   = 0;
 	} cache = {};
 
 	struct Hash {
@@ -450,8 +450,9 @@ public:
 			     i++, maskct++) {
 				if (write_map[i]) {
 					// only adjust writemap if it isn't masked
-					if ((maskct >= block->cache.masklen) ||
-					    (!block->cache.wmapmask[maskct])) {
+					assert(block->cache.wmapmask);
+					if ((maskct >= block->cache.wmapmask->size()) ||
+					    (!(*block->cache.wmapmask)[maskct])) {
 						write_map[i]--;
 					}
 				}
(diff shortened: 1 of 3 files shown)


kcgen:
Rebased; this moves the write map to a `std::vector`.

weirddan455:
Is this performance sensitive code?  If so you're adding a double indirection here by storing a pointer to a vector (allocated by `new`) and then the vector itself contains a 2nd pointer to the actual data.

The code looks fine to me other than that.  I don't know how hot this code is so maybe it doesn't matter.

kcgen:
Yup; it's in the area of sensitivity ... (and two dereferences certainly isn't cache-friendly, for being cache objects!)

Will add some logging and poke around; I remember some games are pretty hard on the dynrec cache (resulting in constantly recreating new cache blocks). 

If it turns out every cache block needs a write map aray of the same size, then it would really nice to eliminate the dynamic allocation, too.

johnnovak:
What's the upper-bound for the size of the mask? Surely, it can't be large, so `std::vector` seems a bit heavy-handed solution for such performance-intensive code. Can it ever grow beyond say 4 or 8 bytes? If not, definitely don't use a vector, just store the length manually and use a a 8-byte long uint8_t buffer.

src/cpu/dyn_cache.h
```diff
@@ -566,28 +567,22 @@ CacheBlock::~CacheBlock() {
...
-	std::copy(cache.wmapmask, cache.wmapmask + cache.masklen, new_mask);
-
-	// Update the current
-	delete[] cache.wmapmask;
-	cache.wmapmask = new_mask;
-
-	cache.masklen = new_mask_len;
+	if (cache.wmapmask) {
+		assert(new_mask_len > cache.wmapmask->size());
+		cache.wmapmask->resize(new_mask_len);
```
FeralChild64, line 581: Do we know what are the expected / maximum sizes of `wmapmask`? We could put a `wmapmask->reserve(…)` call somewhere to ensure the C++ standard library won’t move the vector content in memory (maybe it doesn’t already - implementations are smart enough to not always return the extra memory if vector is being shrink - but I would prefer to be sure).

FeralChild64:
@kcgen I have created an issue late in the evening - I woke up the next day and the fix is aleady available for the review. Nice :D

src/cpu/dyn_cache.h
kcgen: Yup; it's capped at 65k pages - the cache of dynamically-generated pages caps out around 256 MB.

> If not, definitely don't use a vector, just store the length manually and use a a 8-byte long uint8_t buffer.

The good news is the `std::vector` is only has three members: the pointer, capacity, and current size (it avoids re-allocating when when `resize(n)`ing < `capacity`).

kcgen:
> @kcgen I have created an issue late in the evening - I woke up the next day and the fix is aleady available for the review. Nice :D

Even better; @weirddan455's already fixed the existing manually-managed code! :rocket:

Time to add some logging!

kcgen:
Hmm.. it's even lighter used than I expected; most games don't use it (Doom: never, Quake: never, ..). Only a couple of Phil's benchmark make use  of just a couple 64-byte write masks.

Windows 3.1 was the largest user - opening a new "folder" of icons allocates six, 64-bit write-masks, but only one-time. 

The authors of the dyn_cache (for all cores) actually "codified" it's frequency with the expectation that it doesn't exist:

```c++
if (GCC_UNLIKELY(block->cache.wmapmask)) {
    ...
```

PR #2994, opened 2023-10-12
De-unionise X86PageEntry

kklobe:
# Description
`paging.cpp` does quite a bit of type punning against the `X86PageEntry` union, e.g. setting a value through `.load` then immediately reading through `.block`. I've added `get()` and `set()` functions to the struct to replace the punning.

# Manual testing
I tested Quake and Doom, both of which heavily use paging, on the assumption that if something fundamental were wrong with the logic, they would crash almost immediately.

This approach seems to have no impact on performance, as noted by this [microbenchmark](https://gist.github.com/kklobe/e9ff3cb0d455c3ff659c5d3d5678be54). Here are three consecutive runs on my M1 Mac:

```log
Old struct/union duration: 31662 microseconds
New struct duration: 31545 microseconds

Old struct/union duration: 31656 microseconds
New struct duration: 31417 microseconds

Old struct/union duration: 32409 microseconds
New struct duration: 31611 microseconds
```
# Checklist
I have:

### DIFF
--- a/include/paging.h
+++ b/include/paging.h
@@ -100,45 +100,92 @@ void MEM_SetLFB(Bitu page, Bitu pages, PageHandler *handler, PageHandler *mmioha
 void MEM_SetPageHandler(Bitu phys_page, Bitu pages, PageHandler * handler);
 void MEM_ResetPageHandler(Bitu phys_page, Bitu pages);
 
-
-#ifdef _MSC_VER
-#pragma pack (1)
-#endif
-struct X86_PageEntryBlock{
+struct X86PageEntry {
 #ifdef WORDS_BIGENDIAN
-	uint32_t		base:20;
-	uint32_t		avl:3;
-	uint32_t		g:1;
-	uint32_t		pat:1;
-	uint32_t		d:1;
-	uint32_t		a:1;
-	uint32_t		pcd:1;
-	uint32_t		pwt:1;
-	uint32_t		us:1;
-	uint32_t		wr:1;
-	uint32_t		p:1;
+	uint32_t base : 20;
+	uint32_t avl : 3;
+	uint32_t g : 1;
+	uint32_t pat : 1;
+	uint32_t d : 1;
+	uint32_t a : 1;
+	uint32_t pcd : 1;
+	uint32_t pwt : 1;
+	uint32_t us : 1;
+	uint32_t wr : 1;
+	uint32_t p : 1;
 #else
-	uint32_t		p:1;
-	uint32_t		wr:1;
-	uint32_t		us:1;
-	uint32_t		pwt:1;
-	uint32_t		pcd:1;
-	uint32_t		a:1;
-	uint32_t		d:1;
-	uint32_t		pat:1;
-	uint32_t		g:1;
-	uint32_t		avl:3;
-	uint32_t		base:20;
-#endif
-} GCC_ATTRIBUTE(packed);
-#ifdef _MSC_VER
-#pragma pack ()
+	uint32_t p : 1;
+	uint32_t wr : 1;
+	uint32_t us : 1;
+	uint32_t pwt : 1;
+	uint32_t pcd : 1;
+	uint32_t a : 1;
+	uint32_t d : 1;
+	uint32_t pat : 1;
+	uint32_t g : 1;
+	uint32_t avl : 3;
+	uint32_t base : 20;
 #endif
 
+	constexpr void set(const uint32_t value)
+	{
+#ifdef WORDS_BIGENDIAN
+		base = (value >> 12) & 0xFFFFF;
+		avl  = (value >> 9) & 0x7;
+		g    = (value >> 8) & 0x1;
+		pat  = (value >> 7) & 0x1;
+		d    = (value >> 6) & 0x1;
+		a    = (value >> 5) & 0x1;
+		pcd  = (value >> 4) & 0x1;
+		pwt  = (value >> 3) & 0x1;
+		us   = (value >> 2) & 0x1;
+		wr   = (value >> 1) & 0x1;
+		p    = value & 0x1;
+#else
+		p    = value & 0x1;
+		wr   = (value >> 1) & 0x1;
+		us   = (value >> 2) & 0x1;
+		pwt  = (value >> 3) & 0x1;
+		pcd  = (value >> 4) & 0x1;
+		a    = (value >> 5) & 0x1;
+		d    = (value >> 6) & 0x1;
+		pat  = (value >> 7) & 0x1;
+		g    = (value >> 8) & 0x1;
+		avl  = (value >> 9) & 0x7;
+		base = (value >> 12) & 0xFFFFF;
+#endif
+	}
 
-union X86PageEntry {
-	uint32_t load = 0;
-	X86_PageEntryBlock block;
+	constexpr uint32_t get() const
+	{
+		uint32_t value = 0;
+#ifdef WORDS_BIGENDIAN
+		value |= (base << 12);
+		value |= (avl << 9);
+		value |= (g << 8);
+		value |= (pat << 7);
+		value |= (d << 6);
+		value |= (a << 5);
+		value |= (pcd << 4);
+		value |= (pwt << 3);
+		value |= (us << 2);
+		value |= (wr << 1);
+		value |= p;
+#else
+		value |= p;
+		value |= (wr << 1);
+		value |= (us << 2);
+		value |= (pwt << 3);
+		value |= (pcd << 4);
+		value |= (a << 5);
+		value |= (d << 6);
+		value |= (pat << 7);
+		value |= (g << 8);
+		value |= (avl << 9);
+		value |= (base << 12);
+#endif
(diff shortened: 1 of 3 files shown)


include/paging.h
```diff
@@ -101,44 +101,90 @@ void MEM_SetPageHandler(Bitu phys_page, Bitu pages, PageHandler * handler);
...
+    uint32_t    pcd:1;
+    uint32_t    a:1;
+    uint32_t    d:1;
+    uint32_t    pat:1;
+    uint32_t    g:1;
+    uint32_t    avl:3;
+    uint32_t    base:20;
 #endif
 
+    void set(const uint32_t value) {
```
kcgen, line 131: Maybe can be `constexpr`, too?

include/paging.h
```diff
@@ -101,44 +101,90 @@ void MEM_SetPageHandler(Bitu phys_page, Bitu pages, PageHandler * handler);
...
+        g = (value >> 8) & 0x1;
+        avl = (value >> 9) & 0x7;
+        base = (value >> 12) & 0xFFFFF;
+#endif
+    }
 
-union X86PageEntry {
-	uint32_t load = 0;
-	X86_PageEntryBlock block;
+    uint32_t get() const {
```
kcgen, line 159: Hopefully also `constexpr`-able!

src/cpu/paging.cpp
```diff
@@ -214,16 +214,16 @@ static inline bool InitPageCheckPresence_CheckOnly(PhysPt lin_addr,bool writing,
...
+	if (!table.p) {
 		paging.cr2=lin_addr;
 		cpu.exception.which=EXCEPTION_PF;
 		cpu.exception.error=(writing?0x02:0x00) | (((cpu.cpl&cpu.mpl)==0)?0x00:0x04);
 		return false;
 	}
-	const auto entry_addr=(table.block.base<<12)+t_index*4;
-	entry.load=phys_readd(entry_addr);
-	if (!entry.block.p) {
+	const auto entry_addr=(table.base<<12)+t_index*4;
```
kcgen, line 224: Suggest a follow-up `./scripts/format-commit.sh` commit that formats these changes.

src/cpu/paging.cpp
```diff
@@ -387,26 +387,26 @@ class InitPageHandler final : public PageHandler {
...
-			if (!table.block.a) {
-				table.block.a=1;		// set page table accessed
-				phys_writed((paging.base.page<<12)+(lin_page >> 10)*4,table.load);
+			if (!table.a) {
+				table.a=1;		// set page table accessed
+				phys_writed((paging.base.page<<12)+(lin_page >> 10)*4,table.get());
 			}
-			if ((!entry.block.a) || (!entry.block.d)) {
-				entry.block.a=1;		// set page accessed
+			if ((!entry.a) || (!entry.d)) {
```
kcgen, line 399: Suggest dropping the extra `()`, this is a bit easier on the eyes:

```c++
if (!entry.a || !entry.d)
```

src/cpu/paging.cpp
```diff
@@ -387,26 +387,26 @@ class InitPageHandler final : public PageHandler {
...
+				entry.a=1;		// set page accessed
 
 				// page is dirty if we're writing to it, or if we're reading but the
 				// page will be fully linked so we can't track later writes
-				if (writing || (priv_check==0)) entry.block.d=1;		// mark page as dirty
+				if (writing || (priv_check==0)) entry.d=1;		// mark page as dirty
```
kcgen, line 404: The formtter will tidy this up.

kklobe:
HOLD ON THIS, I see Windows builds are crashing.

kcgen (approved):
Looks great @kklobe !
Nice to have the extra `load` and `block` layers removed from the code itself (regardless if the optimizer eliminates them or not).

Will takes some measurements on the linux side too.

Reviwed in detail - and cant' see any mistakes; all 1:1 replacements.

Only suggestion is to apply a formatting pass and add as a whitespace-only commit.

kklobe:
I did run `format-commit.sh`, will do another pass.

kcgen:
Measured 30s of `Blasphemy-MoralHardCandy` using:

```ini
[cpu]
cputype = pentium_slow
cycles = 250000
```

### `operf`,  30 seconds of sampling

Branch:

```text
   602065 100.000 dosbox
	cpu_clk_unhalt...|
	  samples|      %|
	------------------
	   388467 64.5224 anon (tgid:1447506 range:0x7fb48f7f3000-0x7fb48fff7fff)
	   185958 30.8867 dosbox
	    14475  2.4042 no-vmlinux
	     4417  0.7336 libc.so.6
	     2413  0.4008 [vdso] (tgid:1447506 range:0x7fffa5da0000-0x7fffa5da1fff)
	     1604  0.2664 iris_dri.so
	     1238  0.2056 libSDL2-2.0.so.0.2600.3
```

Main:


```text
cpu_clk_unhalt...|
  samples|      %|
------------------
   626970 100.000 dosbox
	cpu_clk_unhalt...|
	  samples|      %|
	------------------
	   395469 63.0762 anon (tgid:1443719 range:0x7f4978ff6000-0x7f49797fafff)
	   204914 32.6832 dosbox
	    14164  2.2591 no-vmlinux
	     4254  0.6785 libc.so.6
	     2338  0.3729 [vdso] (tgid:1443719 range:0x7ffd16ddd000-0x7ffd16ddefff)
	     1437  0.2292 iris_dri.so
	     1187  0.1893 libSDL2-2.0.so.0.2600.3
```

### `time`, 30 seconds of runtime

Branch:

```
real	0m30.046s
user	0m15.151s
sys	0m0.436s
```
Main:

```
real	0m30.060s
user	0m15.787s
sys	0m0.384s
```

### Results

Both runs measured Blasphemy-MoralHardCandy at fixed 250k cycles for 30 seconds before being killed.

1. Operf recorded `626970` samples from `main` and `602065` samples from the branch over 30s of runtime. Although the sampling time-frame isn't *exact* (leading to back-to-back runs having slightly different event counts), this is showing a 4% event-count reduction, which is more than can be account for in just timing errors (notice the `time` counts are within (~0.05% of each other).

2. `time` recorded `main` consumed `15.787s` of user-time across `30.060s` of wall-time, or 52.5% of CPU usage.  The branch consumed `15.151s` of user-time across `30.046s` of wall-time, or ~50.4% CPU usage.

So the results show a little reduction in CPU load to achieve the same results.  Awesome, @kklobe!

---

Script:

```bash
#!/bin/bash

# Check if the command is provided
if [[ "$#" -eq 0 ]]; then
    echo "Usage: $0 <command> [args...]"
    exit 1
fi

if [[ -d oprofile_data ]]; then
  rm -rf oprofile_data
fi

# Launch the command
time "$@" &
command_pid=$(pidof $1)

# Start profiling via the PID
operf --pid "$command_pid" &
operf_pid=$!

# Profile for 30 seconds
sleep 30

# Stop the command and profiler
kill -15 $command_pid
kill -15 $operf_pid
killall operf

opreport
```

kklobe:
Ok, Windows builds are good again, I ripped out the vectorizing in `paging.tlb.read/write`.

kcgen (approved):

kcgen:
Will hold on merging till others can review and approve.
Passing everything I can throw at it.

kklobe:
Interesting side note: looks like compilers just collapse `get()` and `set()` to a single register load and store.

https://godbolt.org/z/hcT95v3xf

kcgen:
> Interesting side note: looks like compilers just collapse `get()` and `set()` to a single register load and store.
> 
> https://godbolt.org/z/hcT95v3xf

Looks great! I like how making the code more explicit lets the compiler very efficiently manage the registers how it wishes.

johnnovak:
Is this related to Elite II @kklobe ? What to look out for when testing this?

kklobe:
> Is this related to Elite II @kklobe ? What to look out for when testing this?

There is no specific app issue this addresses. The undefined behavior of type punning in paging has always bugged me and I wanted to see if we could get rid of it. Just due to the nature of the paging block logic, I would expect immediate catastrophic crashes on any program that uses paging if there are any bugs; I can't imagine any subtle edge cases that would arise from such a straightforward change.

kcgen:
> Is this related to Elite II @kklobe ? What to look out for when testing this?

@kklobe is taking away the Elite II paging bug's musical chairs one by one.. :chair: :chair: :bug: !

johnnovak:
> > Is this related to Elite II @kklobe ? What to look out for when testing this?
> 
> There is no specific app issue this addresses. The undefined behavior of type punning in paging has always bugged me and I wanted to see if we could get rid of it. Just due to the nature of the paging block logic, I would expect immediate catastrophic crashes on any program that uses paging if there are any bugs; I can't imagine any subtle edge cases that would arise from such a straightforward change.

Thanks for the explanation @kklobe.

johnnovak:
I'm not gonna review this in detail, so if stuff still runs, that's good enough for me 😄

kcgen:
Another improvement, down the chute :rocket: !

PR #2997, opened 2023-10-13
Rework shader loading

GranMinigun:
# Description

This PR changes the way shader programs and related settings are stored. Instead of keeping just one shader program and recompiling it each time a shader change is requested, every shader is now loaded and compiled once on demand, then stored in a `std::map`. This allows for much quicker transitions between shaders in `crt-auto` sets, and shader loading process in general was reworked a bit to be slightly more sane.

Partially extracted from #2450. The first two commits are self-explanatory high-precision chainsaw surgery, mostly unrelated.

Unfortunately, the work is unfinished (see the next section), and I don't really have a feeling for where it should move. The real reason behind the changes was to have some sane way of loading shaders outside of OpenGL rendering initialization in #2450, and initially the changes were meant to be a part of that PR. The entanglement of `sdlmain.cpp`, `render.cpp`, and `shader_manager.cpp` did me. Couldn't sit on it, so pushing as is. It's up for grabs, if anyone wants it. Sorry.

# Manual testing

- Running with any shader set in the configuration - works;
- Triggering automatic shader switch by resizing the window - works, the storage is populated as expected, but there's still a noticeable hiccup whenever the target shader changes pixel doubling setting;
- Triggering manual shader reload (Ctrl+F2) - fails, not implemented;
- Attempting to load a non-existing shader source file - works, shader manager catches that and instead passes a fallback source (note that it will still attempt to load other sources from the auto set and will retain integer scaling setting when automatic switch is triggered);
- Attempting to load a malformed shader source - fails, either keeps the last successfully compiled shader or outputs nothing.

# Checklist

I have:

### DIFF
--- a/include/render.h
+++ b/include/render.h
@@ -180,6 +180,7 @@ bool RENDER_MaybeAutoSwitchShader([[maybe_unused]] const uint16_t canvas_width,
                                   [[maybe_unused]] const uint16_t canvas_height,
                                   [[maybe_unused]] const VideoMode& video_mode,
                                   [[maybe_unused]] const bool reinit_render);
+const std::string RENDER_GetShaderSource();
 
 void RENDER_NotifyEgaModeWithVgaPalette();
 
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -173,11 +173,9 @@ struct SDL_Block {
 		int pitch = 0;
 		void* framebuf = nullptr;
 		GLuint texture;
-		GLuint displaylist;
 		GLint max_texsize;
 		bool bilinear;
 		bool npot_textures_supported = false;
-		bool use_shader;
 		bool framebuffer_is_srgb_encoded;
 		GLuint program_object;
 
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -177,17 +177,7 @@ struct SDL_Block {
 		bool bilinear;
 		bool npot_textures_supported = false;
 		bool framebuffer_is_srgb_encoded;
-		GLuint program_object;
 
-		ShaderInfo shader_info    = {};
-		std::string shader_source = {};
-
-		struct {
-			GLint texture_size;
-			GLint input_size;
-			GLint output_size;
-			GLint frame_count;
-		} ruby = {};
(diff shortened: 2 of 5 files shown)


johnnovak:
> This PR changes the way shader programs and related settings are stored. Instead of keeping just one shader program and recompiling it each time a shader change is requested, every shader is now loaded and compiled once on demand, then stored in a `std::map`. This allows for much quicker transitions between shaders in `crt-auto` sets, and shader loading process in general was reworked a bit to be slightly more sane.

That would be a nice improvement, yeah. Thanks for giving this a go @GranMinigun.

> The entanglement of `sdlmain.cpp`, `render.cpp`, and `shader_manager.cpp` did me.

Welcome to my world 😎 Took me at least a week to figure out where to insert the shader manager hooks so things work reliably... 😅 

> Couldn't sit on it, so pushing as is. It's up for grabs, if anyone wants it. Sorry.

I'm interested in this, and I'll try to finish it but only after 0.81.0 is out—everything else takes priority, and auto-shader loading works ok as it is now. Who know, you might get a second wind until then and finish it yourself 😉 

Some feedback on your points:

>     * Triggering automatic shader switch by resizing the window - works, the storage is populated as expected, but there's still a noticeable hiccup whenever the target shader changes pixel doubling setting;

Yes, and I guess that will be always like that because we need to re-render the frame with different dimensions if the pixel doubling or "true" double-scanning settings change. Although that should be rather fast... Hmmm.

>     * Triggering manual shader reload (Ctrl+F2) - fails, not implemented;

That's a _must_ for shader tweaking, so I want it back 😎 (either I'll do it, or you if you pick this up again).

>     * Attempting to load a non-existing shader source file - works, shader manager catches that and instead passes a fallback source (note that it will still attempt to load other sources from the auto set and will retain integer scaling setting when automatic switch is triggered);

Nice! 🎉 

>     * Attempting to load a malformed shader source - fails, either keeps the last successfully compiled shader or outputs nothing.

Keeping the last sounds fine, but it should default to `none.glsl` otherwise. And manual shader loading should re-attempt loading it (e.g. if you made a typo during shader tweaking, fixing the typo then a manual reload should work).

Issue #2996, opened 2023-10-12
Assess and fix Coverity issues prior to 0.81 release


kcgen:
These are all very minor and likely need just a line or two to fix up.

Feel free to grab any of them and push single fix up PRs.

1. ![2023-10-12_14-46](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/bd01d03b-bae8-480d-965a-7f35d5e224b2)

1. ![2023-10-12_14-45](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/694cb8af-a8ea-45e9-8174-4fa797b61fd8)

1. ![2023-10-12_14-44](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/0bc591db-a476-4b6b-8a0e-d49ea08d30ac)

1. ![2023-10-12_14-43](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/87aa56a2-9699-4ea4-9138-dad634e48749)

1. ![2023-10-12_14-45-1](https://github.com/dosbox-staging/dosbox-staging/assets/1557255/8ac0ca57-e749-48a7-ae61-4fa0d30d85a5)

weirddan455:
Some of these are from third party libraries.  I saw at least SDL Sound and stb_vorbis in there.  Do we really want to be diverging from upstream for some minor warning fixes?  Or should be ignoring these warnings from the `libs` directory?

kcgen:
Yup, PR or report as much as possible upstream (just like you did w/ libmt32emu). For STB Vorbis specifically, we've:

- reported 19 issues: https://github.com/nothings/stb/issues?q=author%3Akcgen+is%3Aclosed - 
- PR'd fixed for two Coverity issues: https://github.com/nothings/stb/pull/854
- PR'd fixes for file handling issues: https://github.com/nothings/stb/pull/678

Suggest a similar approach here; and no problem carrying this issue in Coverity as long as needed till upstream fixes it. 

We've had a similar approach for other ported-in sources too (like Enet, dr_flac, dr_wav, dr_mp3, PD curses, slirp, .. and maybe others).

kcgen:
For `SDL_Sound.c`, this is carried over from SDL1 (when SDL2 had abandoned it).

It's highly cut down just for our needs for compressed CDDA audio (but still has the same original API).  

All that to say, it's own own file to mangle and maintain for now (it's  received lots of prior PVS and Coverity fixes, as well as warning cleanups). 

[...]

PR #3000, opened 2023-10-14
Fix a Coverity warning in SDL Sound

weirddan455:
# Description

First commit fixes the assert macro.  The downside of how we're copy + pasting third party libraries into our tree is that we're not using the build system it was written for.  `HAVE_ASSERT_H` is something SDL's build system would have defined.  I just removed that macro and always do a `#include assert.h` since we rely on that header in our other code regardless.

Second commit adds an assert that hopefully Coverity is smart enough to follow.  It thinks there is a use after free bug if we're removing from the middle of the linked list in `Sound_Quit` but that never happens.  It's freeing the entire list always removing from the head.

## Related issues

#2996

# Manual testing

Tested with ripped .ogg files of the Quake sound track.  Added the assert and confirmed it does not trigger on quit.

# Checklist

I have:

### DIFF
--- a/src/libs/decoders/SDL_sound.c
+++ b/src/libs/decoders/SDL_sound.c
@@ -570,6 +570,11 @@ void Sound_FreeSample(Sound_Sample *sample)
     /* update the sample_list... */
     if (internal->prev != NULL)
     {
+        // We're not removing from the head.
+        // Assert that we're not touching sample_list (which should be the head of the linked list).
+        // This assert fixes a Coverity warning.
+        assert(sample_list != sample);
+
         Sound_SampleInternal *prevInternal;
         prevInternal = (Sound_SampleInternal *) internal->prev->opaque;
         prevInternal->next = internal->next;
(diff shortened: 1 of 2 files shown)


weirddan455:
I don't see Coverity in the list of CI checks for this PR.  @kcgen Does this only happen in the main branch?  Is there a way I can test this before merging to see if it actually fixes the warning?

kcgen (approved):
Nice fix, @weirddan455 !

kcgen:
> I don't see Coverity in the list of CI checks for this PR. @kcgen Does this only happen in the main branch? Is there a way I can test this before merging to see if it actually fixes the warning?

Yes - we have only 7 scans per week; once we use them, PRs will pass through unscanned.

I'll merge your Coverity fixes in quick succession (CI will quickly cancel prior merges), so this is one way to manually "combine" a handful of small merges and have them only "cost" us a single Coverity scan.

A scan takes about an hour to complete and it does the entire code base.

PR #3005, opened 2023-10-15
Try again to fix Coverity warning in SDL Sound

weirddan455:
# Description
The previous PR did not solve the issue.  Coverity saw my assert and still thinks it's freeing from the middle of the list.

coverity

It's possible it is seeing a data race.  I've removed the use of the volatile keyword and used a mutex lock around the entire loop.  `Sound_FreeSample` also takes a lock but SDL's mutex is recursive so this should be fine.

This also removes a temporary variable that was maybe confusing Coverity.  Now `Sound_FreeSample` is explicitly being called with `sample_list` (head of the linked list).

## Related issues

#2996

# Manual testing

Started Quake with ripped .ogg sound files.  Closed Dosbox and confirmed there was no crashes or deadlocks.

# Checklist

I have:

### DIFF
--- a/src/libs/decoders/SDL_sound.c
+++ b/src/libs/decoders/SDL_sound.c
@@ -570,11 +570,6 @@ void Sound_FreeSample(Sound_Sample *sample)
     /* update the sample_list... */
     if (internal->prev != NULL)
     {
-        // We're not removing from the head.
-        // Assert that we're not touching sample_list (which should be the head of the linked list).
-        // This assert fixes a Coverity warning.
-        assert(sample_list != sample);
-
         Sound_SampleInternal *prevInternal;
         prevInternal = (Sound_SampleInternal *) internal->prev->opaque;
         prevInternal->next = internal->next;
--- a/src/libs/decoders/SDL_sound.c
+++ b/src/libs/decoders/SDL_sound.c
@@ -148,12 +148,12 @@ int Sound_Quit(void)
 
     BAIL_IF_MACRO(!initialized, ERR_NOT_INITIALIZED, 0);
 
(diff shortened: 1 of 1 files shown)


weirddan455:
I'm not sure what's wrong with the macOS build.  I've seen it fail like this before randomly.  I think this is the machine you run locally @kcgen ?

>  Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/Volumes/RamDrive/ccache/tmp/_actions/actions/checkout/v4'. Did you forget to run actions/checkout before running your local action?

kcgen (approved):
Looks good, @weirddan455.
Let's see what Coverity thinks.

kcgen:
> I'm not sure what's wrong with the macOS build. I've seen it fail like this before randomly. I think this is the machine you run locally @kcgen ?
> 
> > Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/Volumes/RamDrive/ccache/tmp/_actions/actions/checkout/v4'. Did you forget to run actions/checkout before running your local action?

Yes - this has eluded me. The machine runs a single GitHub's "runner" process; it's in charge of managing its workspace where it checks out the project and run the CI commands.

The runs happen serially, usually with a couple seconds between back-to-back runs.

When it goes off the rails, Meson reports that a bunch of critical project files are missing - like in this case, `docs/dosbox.1`, which should be impossible because the runner checks out the project (and the branch), sets up dependencies, etc.. before setting up the build.

My only hunch is that there's some kind of race-condition or improper conditions where the runner purges the content from a running job, and it impacts the current job, because the runner does re-use the same directory tree for subsequent jobs.

I've stopped the runner, purged all the directories, and restarted it.

Issue #2987, opened 2023-10-11
`dosbox FOOBAR.EXE` no longer mounts and launches the executable


es20490446e:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.1

### What Operating System are you using?

Linux x86_64

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

In `dosbox-staging.conf`, the `autoexec` section seems not to work when launching a game from the terminal with `dosbox [game]`. Only when launching `dosbox` first, then entering the appropriate commands to run the game.

Oddly it will work if you launch the game directly with `dosbox [game]`, then exit it without leaving the emulator, then launching it again by typing commands. On that second launch it will work.

For example `mixer fsynth 20` won't work.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

FeralChild64:
@es20490446e The `[autoexec]` section implementation was rewritten few months ago - if you can, try the latest GIT main, it now executes the supplied command after the content of `[autoexec]`.

es20490446e:
Let me check...

es20490446e:
I can't test because git just [crashes](https://github.com/dosbox-staging/dosbox-staging/files/12876425/dosbox.log) with any program. Also by deleting the dosbox configuration folder.

johnnovak:
> I can't test because git just [crashes](https://github.com/dosbox-staging/dosbox-staging/files/12876425/dosbox.log) with any program. Also by deleting the dosbox configuration folder.

That seems like a clean exit according to the posted logs.

kcgen:
@es20490446e , just to confirm, you're not using any [autoexec] sections because you're starting from scratch (without a primary config,which gets created).

Without an autoexec section, passing an executable on the CLI like:

`/Path/to/dosbox-staging INSTALL.EXE` 

Should result in:
1. The file's path (`./` in this case) should be mounted as`C:`
2. The `C:` drive should be switched into.
3. The executable, now present and available in DOS's current working directory, should be launched.
4. When the executable exits back to the DOS prompt, DOSBox Staging itself should exit back to the host (I believe with error-code passthru, too). 

In this case though, the executable doesn't even get off the ground for a split second, so step 4 is aborted and the shell is kept running so you can investigate the problem:

`SHELL: Exit blocked because program quit after only 0.2 seconds`

[...]

PR #3007, opened 2023-10-15
Autoexec fixes

FeralChild64:
# Description

- restored behavior of `dosbox FOOBAR.EXE` automounting current directory as drive C:
- restored behavior of skipping `[autoexec]` sections if directory, file, or boot image was provided on the command line, and  `autoexec_section` from `[dosbox]` has value different from `join`
- emulator won't try to automount  Z: drive from a user provided directory
- all the mount commands generated are now preceded by unmount attempts
- commented out the `SHELL: Redirecting output to ...` log, it is printed out much too often, also when executing some autoexecs

---

**Note:** The `Z:\AUTOEXEC.BAT` generator tries to be smart and friendly - but IMHO it often works in a way which is strange and counterintuitive:

- imagine we fave `FOOBAR.BAT` in the current directory; `dosbox FOOBAR.BAT` will generate `CALL FOOBAR.BAT`, but `dosbox FOOBAR` or `dosbox "FOOBAR.BAT /switch=y"` won't prefix the batch file invocation with `CALL`
- one of the conditions to determine whether `AUTOEXEC.BAT` should get an exit command at the end is the startup verbosity parameter, and it is the value `auto` that is handled differently
- if we have a `DOOM.BAT` file and `DOOM` directory, `dosbox DOOM` will try to mount the directory, but won't launch the game

I'm not using this functionality, I don't know what was the rationale behind them, so it is hard for me to propose a solution that is logical and won't upset users.

---

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/2987

# Manual testing

I recomment to uncomment `#define DEBUG_AUTOEXEC` in `src/shell/autoexec.cpp` and recompile before playing with various options - this will print out the generated `Z:\AUTOEXEC.BAT` to the logs, making it easier to see what is actually happening.

# Checklist

I have:


johnnovak:
Thanks, @FeralChild64.

> I'm not using this functionality, I don't know what was the rationale behind them, so it is hard for me to propose a solution that is logical and won't upset users.

I think it's probably good overall that we restored this functionality, but if it's ultimately something no one in the dev team cares about and it's weird, counterintuitive, obscure, complicated, and edge case-y, maybe it's chopping block time... 😄 

We'll see how this goes and much trouble it causes.

src/shell/autoexec.cpp
```diff
@@ -34,11 +34,19 @@
...
 // ***************************************************************************
 // Constants
 // ***************************************************************************
 
 static const std::string autoexec_file_name = "AUTOEXEC.BAT";
+static const std::string cmd_boot           = "@Z:\\BOOT.COM ";
+static const std::string cmd_config         = "@Z:\\CONFIG.COM ";
+static const std::string cmd_mount          = "@Z:\\MOUNT.COM ";
+static const std::string cmd_imgmount       = "@Z:\\IMGMOUNT.COM ";
+static const std::string to_nul             = " >NUL";
```
johnnovak, line 49: UpperCamelCase constants?

johnnovak (approved):
Code looks fine @FeralChild64 and I have really low interest in this feature so I trust your manual testing 😄

weirddan455:
Did some testing.

```
cd /home/daniel/dos/DOOM_SE
/home/daniel/code/dosbox-staging/build/debug/dosbox DOOM.EXE
```
This works :+1: 

```
cd /home/daniel/code/dosbox-staging
build/debug/dosbox /home/daniel/dos/DOOM_SE/DOOM.EXE
```
This does not work and is a regression from 0.80.1

>  restored behavior of dosbox FOOBAR.EXE automounting current directory as drive C:

I believe this is the problem.  Behavior should not be to mount current directory but instead mount the directory the .exe is located in.

I'm going to stay out of the argument of whether we should keep this feature in.  I think @kcgen made some good points about why we should have this.

src/shell/autoexec.cpp
FeralChild64: Ok, changed.

FeralChild64:
> ```
> cd /home/daniel/code/dosbox-staging
> build/debug/dosbox /home/daniel/dos/DOOM_SE/DOOM.EXE
> ```
> 
> This does not work and is a regression from 0.80.1
> 
> > restored behavior of dosbox FOOBAR.EXE automounting current directory as drive C:
> 
> I believe this is the problem. Behavior should not be to mount current directory but instead mount the directory the .exe is located in.

OK, I'll investigate it. Switching back to draft for now.

src/shell/autoexec.cpp
```diff
@@ -509,9 +519,21 @@ void AutoExecModule::AutoMountDrive(const std::string& dir_letter)
...
+	AddLine(placement, "@C:");
+}
+
+void AutoExecModule::AutoMountDriveD(const std::string& cdrom_images)
+{
+	constexpr auto placement = Placement::InitialAutogeneratedCommands;
+	AddLine(placement, CmdMount + std::string("-u D") + ToNul);
+	AddLine(placement,
+	        CmdImgMount + std::string("D ") + cdrom_images +
+	                std::string(" -t iso") + ToNul);
```
kcgen, line 549: The `std::string`'s can add literals, so this can be maybe be simplified:

```c++
CmdImgMount + "D " + cdrom_images + " -t iso" + ToNul
```

src/shell/autoexec.cpp
```diff
@@ -509,9 +519,21 @@ void AutoExecModule::AutoMountDrive(const std::string& dir_letter)
...
+	constexpr auto placement = Placement::InitialAutogeneratedCommands;
+	AddLine(placement, CmdMount + std::string("-u C") + ToNul);
+	AddLine(placement, CmdMount + std::string("C ") + directory + ToNul);
+	AddLine(placement, "@C:");
+}
+
+void AutoExecModule::AutoMountDriveD(const std::string& cdrom_images)
+{
+	constexpr auto placement = Placement::InitialAutogeneratedCommands;
+	AddLine(placement, CmdMount + std::string("-u D") + ToNul);
```
kcgen, line 546: `std::string` can add literals, so maybe a simplification: to `+ "-u D" + `

src/shell/autoexec.cpp
```diff
@@ -509,9 +519,21 @@ void AutoExecModule::AutoMountDrive(const std::string& dir_letter)
 	}
 }
 
-void AutoExecModule::AddLine(const Placement placement, const std::string& line)
+void AutoExecModule::AutoMountDriveC(const std::string& directory)
 {
-	autoexec_lines[placement].push_back(line);
+	constexpr auto placement = Placement::InitialAutogeneratedCommands;
+	AddLine(placement, CmdMount + std::string("-u C") + ToNul);
+	AddLine(placement, CmdMount + std::string("C ") + directory + ToNul);
```
kcgen, line 539: std::string can add literals, so maybe a simplification to avoid the `std::string(...)` cast/temporary.

src/shell/autoexec.cpp
```diff
@@ -406,7 +410,7 @@ AutoExecModule::AutoExecModule(Section* configuration)
 			LOG_WARNING("AUTOEXEC: Secure mode skipped as it does not allow booting");
 		} else {
 			AddLine(Placement::SecureModeCommand,
-			        "@Z:\\CONFIG.COM -securemode");
+			        CmdConfig + std::string("-securemode"));
```
kcgen, line 426: std::string can add literals, so maybe a simplification to avoid the `std::string(...)`

src/shell/shell.cpp
```diff
@@ -309,8 +309,8 @@ void DOS_Shell::ParseLine(char *line)
 		if (out_file.length() && pipe_file.length())
 			WriteOut(MSG_Get("SHELL_CMD_DUPLICATE_REDIRECTION"),
 			         out_file.c_str());
-		LOG_MSG("SHELL: Redirecting output to %s",
-		        pipe_file.length() ? pipe_tempfile : out_file.c_str());
+		// LOG_MSG("SHELL: Redirecting output to %s",
+		//         pipe_file.length() ? pipe_tempfile : out_file.c_str());
```
kcgen, line 313: :+1:

src/shell/autoexec.cpp
```diff
@@ -502,9 +532,29 @@ void AutoExecModule::AutoMountDrive(const std::string& dir_letter)
...
+	        CmdImgMount + std::string("D ") + cdrom_images +
+	                std::string(" -t iso") + ToNul);
+}
+
+void AutoExecModule::ReMountCurrentDirAsDriveC()
+{
+	constexpr auto placement = Placement::CommandsAfterAutoexecSection;
+	AddLine(placement, CmdMount + std::string("-u C") + ToNul);
+	AddLine(placement, CmdMount + std::string("C .") + ToNul);
+	AddLine(placement, "@C:");
```
kcgen, line 557: very tidy :-)

src/shell/autoexec.cpp
FeralChild64: Ok, changed.

src/shell/autoexec.cpp
FeralChild64: Ok, changed.

src/shell/autoexec.cpp
FeralChild64: Ok, changed.

src/shell/autoexec.cpp
FeralChild64: OK, changed.

FeralChild64:
@weirddan455 Should be fixed now.
@kcgen You can continue the review.

kcgen (approved):
Working great, @FeralChild64.
Executables with relative paths, absolute paths, and those without any path (local files) all mount and launch without issues. :+1: 

Baseline tests (`-c`, `-conf`, and without args) are running good too.

kcgen:
I think everyone's reviewed the code, and with the regression fixed, it's time to merge.

es20490446e:
Thanks for the fix, number one 🥇

PR #3009, opened 2023-10-15
Refactor Property->Get_string to return a std::string

weirddan455:
# Description
This was previously returning a const char* pointer to a std::string.c_str()
This has been a source of bugs in the past if the backing std::string goes out of scope or the caller tries to modify it.

Many of the callers of this function were constructing a `std::string` or doing a C style `strcpy` anyway.  In one case, the caller was `const_cast`'ing the returned `const char*` and modifying it in place (yikes).

I just think returning a pointer to a `std::string` object outside of API bounds is a bad idea.  Performance really shouldn't matter here.  The strings are short and it's all initialization code.  I highly doubt this would even show up in a profiler.

# Manual testing
I started DosBox up and changed a few of the config settings (particularly the resolution as I removed some C-style string manipulation).  Did not notice any regressions.  Behavior should not change as this is a pretty straight forward refactoring.

I will note that Clang caught some bugs that GCC did not.  There were a few places with something like `std::string_view foo = prop->Get_string("bar")` which were fine when it returned a `char*`.  With a `std::string`, it creates a temporary object that gets immediately destroyed, leaving behind a dangling `std::string_view` pointing to an invalid object.

Clang caught this citing `-Wdangling-gsl` (which was already enabled by our build system somewhere, maybe part of `-Wall` or something).  Could not find a way to make GCC report this warning.  This may make me switch to Clang as my preferred compiler.

I also did a `git grep Get_string` to manually check everywhere this was called.

# Checklist

I have:

### DIFF
--- a/include/setup.h
+++ b/include/setup.h
@@ -107,7 +107,7 @@ public:
 	operator Hex() const;
 	operator int() const;
 	operator double() const;
-	operator const char*() const;
+	operator std::string() const;
 
 	bool SetValue(const std::string& in, Etype _type = V_CURRENT);
 
@@ -399,7 +399,7 @@ public:
 
 	int Get_int(const std::string& _propname) const;
 
-	const char* Get_string(const std::string& _propname) const;
+	std::string Get_string(const std::string& _propname) const;
 
 	Prop_string* GetStringProp(const std::string& propname) const;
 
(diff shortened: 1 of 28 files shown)


johnnovak (approved):
Big thanks for addressing this @weirddan455 🏅 

This is what tends to happen when someone values "shaving off" 0.00001 nanosecond more than causing hours of headaches when hunting bugs caused by such "optimisations" 🤦🏻‍♂️ 

For the record, I hit this a few times and got crashes but then just copied the same existing pattern which worked and moved on with life. The config handling in general is pretty bad, so I always thought we should rebuild that anyway, eventually.

But until then, this a nice incremental improvement! 🎉

Issue #2969, opened 2023-10-01
Various problems with the `model` setting for the MT-32 emulation


johnnovak:
A couple of problems with the `model` setting for the MT-32:

- The `mt32` and `cm32l` symbolic model settings stopped working (probably a regression).
- Changing `model` only takes effect when resetting the MIDI device with `mididevice = mt32` (very confusing).
- The full list of accepted values should be documented in the config description.
- Add support for `cm32ln`.

weirddan455:
I'll take a look at this.

>  The mt32 and cm32l symbolic model settings stopped working (probably a regression).

This one looks to be my regression from the first PR I made to this project:  https://github.com/dosbox-staging/dosbox-staging/pull/2591

weirddan455:
PR for the first point is up (was a very simple fix).

>  Changing model only takes effect when resetting the MIDI device with mididevice = mt32 (very confusing).

I looked at this.  The way we have the MIDI class set up makes this a bit annoying.  Usually, we have Init and Destroy functions for each section.  However, the MIDI class "manages' each of the MIDI handlers.  The MT32 section doesn't self-init (it has an empty init function that does nothing).  It instead gets initialized from the MIDI section.  I have an idea of how I might fix it though.

johnnovak:
Sorry @weirddan455 , but I'm already in the middle of implementing this. I did self-assign it a while ago, so in such cases when the ticket is already assigned to someone it's best to check with the person to avoid wasted effort.

The auto reinit on ROM model change is tricky, most likely I won't do that for now, though. You can continue to look into that if you're interested, but PR incoming for the rest in a day or two.

johnnovak:
Fixed.

weirddan455:
>  Changing model only takes effect when resetting the MIDI device with mididevice = mt32 (very confusing).

Did you fix this bit or are we just not worrying about that for now?

johnnovak:
> > Changing model only takes effect when resetting the MIDI device with mididevice = mt32 (very confusing).
> 
> Did you fix this bit or are we just not worrying about that for now?

I knew I forgot something! Changed the top description.

Issue #2617, opened 2023-06-19
Refactor MIXER /LISTMIDI


johnnovak:
I know what to do, just a reminder so I won't forget...

Mockups by _yours truly_ we more or less agree on with @kcgen (incl. a few typos 🤦🏻 👀 )

![image](https://github.com/dosbox-staging/dosbox-staging/assets/698770/2cb5e4c5-d8a1-4748-9a8b-d22b89b1cf3c)

![image](https://github.com/dosbox-staging/dosbox-staging/assets/698770/a0052c91-81d4-4bcf-abb6-958ca7a39d9e)

Burrito78:
The asterisk stands for "active", additionally to the green color?

johnnovak:
> The asterisk stands for "active", additionally to the green color?

Yes. So it's consistent with marking the current SoundFont with an asterisk in addition the green colour. I think the asterisk is a good idea, especially for the monochrome Hercules mode (edge case, I know, but still...)

Burrito78:
@johnnovak This is done, right?

johnnovak:
> @johnnovak This is done, right?

Yep.

PR #3010, opened 2023-10-15
Re-add "mt32" and "cm32l" alias models

weirddan455:
# Description
Regression from PR #2591. The logic for these still works. It's a simple reverse string search. They were just accidentally removed from the list of valid config values.

## Related issues
#2969

# Manual testing
Tested model=mt32 and model=cm32l and confirmed they work.

# Checklist

I have:

### DIFF
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -217,6 +217,8 @@ static void init_mt32_dosbox_settings(Section_prop &sec_prop)
 	constexpr auto when_idle = Property::Changeable::WhenIdle;
 
 	const char* models[] = {"auto",
+	                        "mt32",
+	                        "cm32l",
 	                        cm32ln_100_model.GetName(),
 	                        cm32l_102_model.GetName(),
 	                        cm32l_100_model.GetName(),


johnnovak:
Maybe I wasn't clear, but my PR for https://github.com/dosbox-staging/dosbox-staging/issues/2617 will also address this issue from the ticket you referenced.

Thanks for trying to fix this, but let's not merge it as I already have a different alternative implementation in progress. More importantly, the symbolic non-versioned names will pick the best available device for the device class in question, which is now a possibility in all scenarios given we don't care about the filenames.

weirddan455:
Oh, I didn't know you were already working on this @johnnovak Have you pushed a PR for this yet?  I didn't see this fix in #2868

>  More importantly, the symbolic non-versioned names will pick the best available device for the device class in question, which is now a possibility in all scenarios given we don't care about the filenames.

That's what this does.  It iterates in the same order that `auto` does and chooses the first one that matches either MT32 or CM32L (matching on ROM ID, not filename).

But anyway, this was a very quick fix (2 lines of code) so not a big deal if you want to close this and push your own solution.

johnnovak:
> Oh, I didn't know you were already working on this @johnnovak Have you pushed a PR for this yet? I didn't see this fix in #2868

Not yet, PR coming in a day or two.

> > More importantly, the symbolic non-versioned names will pick the best available device for the device class in question, which is now a possibility in all scenarios given we don't care about the filenames.
> 
> That's what this does. It iterates in the same order that `auto` does and chooses the first one that matches either MT32 or CM32L (matching on ROM ID, not filename).

I don't think it's quite the same as my implementation, but you'll see. I will also pick the best version for `mt32_new` and `mt32_old`, not just link them to a sine fixed version, etc.

> But anyway, this was a very quick fix (2 lines of code) so not a big deal if you want to close this and push your own solution.

👍🏻

PR #3011, opened 2023-10-16
Rework the MT-32/CM-32L ROM lookup and `MIXER /LISTMIDI` improvements

johnnovak:
# Description

Now that we have moved from filename-based MT-32/CM-32L ROM lookup to [hash-based](https://github.com/dosbox-staging/dosbox-staging/pull/2591), the unversioned vs versioned ROM differentiation doesn't make much sense.

`mt32_old` and `mt32_new` linking to specific v1.x and v2.x MT-32 versions, respectively, was also a pretty arbitrary decision and was in conflict with how the `auto` model worked.

This change makes the lookup behaviour of "symbolic" models uniform and consistent (`auto`, `mt32`, `cm32l`, `mt32_old`, and `mt32_new`). The lookup mechanism of models that specify the version number (e.g., `mt32_107`, `cm32l_102`) is unchanged.

The output of `MIXER /LISTMIDI` is also improved.

## Previous behaviour

- **Unversioned ROMs** were expected to exist with the `MT32_CONTROL.ROM` & `MT32_PCM.ROM` filenames for MT-32, and `MT-32 CM32L_CONTROL.ROM` & `CM32L_PCM.ROM` for the CM-32L.
- **Versioned ROMs** were expected to be named according to the MAME MT-32/CM-32L file naming convention.
- `model = auto` loaded the first existing ROM in priority order as specified in the config description of `model`.
- `model = mt32` attempted to load the **unversioned** MT-32 ROM and fail the ROM file did not exist.
- `model = cm32l` attempted to load the **unversioned** CM-32L ROM and fail the ROM file did not exist.
- `model = mt32_new` attempted to load the **versioned** v2.04 MT-32 ROM and fail if that exact ROM file did not exist.
- `model = mt32_old` attempted to load the **versioned** v1.07 MT-32 ROM and fail if that exact ROM file did not exist.

## New revised behaviour

- We drop the concept of **versioned** and **unversioned** ROMs completely. Valid MT-32/CM-32L ROMs are identified by their checksums (hashes) at startup, the filenames do not matter. Every identified ROM file is tied to a concrete version number, naturally—it can't work any other way.
- `model = auto` loads the first existing ROM in priority order as specified in the config description of `model`. **This is the same behaviour as before.**
- `model = mt32` loads the first MT-32 ROM using the same priority order (it fails if only CM-32L ROMs are present or no ROMs at all). 
- `model = cm32l` loads the first CM-32L ROM in priority order (it fails if only MT-32 ROMs are present or no ROMs at all). 
- `model = mt32_new` loads the first v2.x MT-32 ROM in priority order (it fails only if no v2.x MT-32 ROMs are present).  
- `model = mt32_old` loads the first v1.x MT-32 ROM in priority order (it fails only if no v1.x MT-32 ROMs are present).   

## New `model` setting description

Basically, a succinct version of the above:

image

## Backward compatibility

In some weird edge cases (point 3), the behaviour _can_ be different than before (meaning a different ROM model might be picked up while the contents of the `romdir` is _exactly_ the same and the user's config is _exactly_ the same too). However, I think this is a complete non-issue as most people would fall into the sane 1) and 2) use cases. People doing weird shit as described in 3) will need to fix up their configs 😎 

1. `romdir` has the "unversioned" ROM files _only_. In this legacy use case, `model = mt32` will pick the MT-32 ROM and `model = cm32l` the CM-32L ROM. Easy!

2. `romdir` has all versioned ROM files (recommended setup). In this case, everything works as expected, even a. bit better (e.g., for `mt32_new` we'll still pick the "best" available "new" MT-32 ROM, even if v2.04 is missing).

3. Mixed case when `romdir` contains _both_ versioned and unversioned ROM files, *and* the person uses `mt32` or `cm32l` which were previously referring to the "unversioned" files—well, depending on what version the "unversioned" ROMs are, and what "versioned" ROMs the person has, the behaviour can be... pretty much anything 😄 However, this use case was always a bit of a clusterfuck and normally people who still use "legacy unversioned ROMs" fall into the first category. The behaviour with mixed versioned and unversioned ROMs was pretty confusing previously and it wasn't a good idea to do that anyway.

## Related issues

Fixes:

- https://github.com/dosbox-staging/dosbox-staging/issues/2969 (except for actually changing the MT-32 when setting a different `model` value)
- https://github.com/dosbox-staging/dosbox-staging/issues/2484
- https://github.com/dosbox-staging/dosbox-staging/issues/2617

# Manual testing

Tested for regressions when using exact model versions and the symbolic models. The symbolic model resolution is more interesting, so here's the test evidence for that:

image

image

image

### Hercules output

image

# Checklist

I have:

### DIFF
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -230,9 +230,9 @@ const LASynthModel mt32_old_model = {"mt32_old", // "mt32_old" is 1.07
                                      &mt32_ctrl_107_b};
 
 // In order that "model = auto" will load
-const LASynthModel* all_models[] = {&cm32ln_100_model,
-                                    &cm32l_102_model,
+const LASynthModel* all_models[] = {&cm32l_102_model,
                                     &cm32l_100_model,
+                                    &cm32ln_100_model,
                                     &mt32_old_model,
                                     &mt32_107_model,
                                     &mt32_106_model,
@@ -252,9 +252,9 @@ static void init_mt32_dosbox_settings(Section_prop& sec_prop)
 	constexpr auto when_idle = Property::Changeable::WhenIdle;
 
 	const char* models[] = {"auto",
-	                        cm32ln_100_model.GetName(),
 	                        cm32l_102_model.GetName(),
 	                        cm32l_100_model.GetName(),
+	                        cm32ln_100_model.GetName(),
 	                        mt32_old_model.GetName(),
 	                        mt32_107_model.GetName(),
 	                        mt32_106_model.GetName(),
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -47,31 +47,39 @@
 // ----------------
 
 // Analogue circuit modes: DIGITAL_ONLY, COARSE, ACCURATE, OVERSAMPLED
-constexpr auto ANALOG_MODE = MT32Emu::AnalogOutputMode_ACCURATE;
+constexpr auto AnalogMode = MT32Emu::AnalogOutputMode_ACCURATE;
 
 // DAC Emulation modes: NICE, PURE, GENERATION1, and GENERATION2
-constexpr auto DAC_MODE = MT32Emu::DACInputMode_NICE;
+//
+// Produce samples at double the volume, without tricks.
+// Nicer overdrive characteristics than the DAC hacks (it simply clips samples
+// within range) Higher quality than the real devices
+constexpr auto DacEmulationMode = MT32Emu::DACInputMode_NICE;
 
 // Analog rendering types: int16_t, FLOAT
-constexpr auto RENDERING_TYPE = MT32Emu::RendererType_FLOAT;
+// Use float samples in the renderer and simplified wave generator model.
+// Maximum output quality and minimum noise.
+constexpr auto RenderingType = MT32Emu::RendererType_FLOAT;
 
 // Sample rate conversion quality: FASTEST, FAST, GOOD, BEST
-constexpr auto RATE_CONVERSION_QUALITY = MT32Emu::SamplerateConversionQuality_BEST;
+constexpr auto ResamplingQuality = MT32Emu::SamplerateConversionQuality_BEST;
 
-// Prefer higher ramp resolution over the coarser volume steps used by the hardware
-constexpr bool USE_NICE_RAMP = true;
+// Prefer amp ramp interpolation to avoid clicks (the hardware doesn't
+// always interpolate).
+constexpr bool UseNiceRamp = true;
 
-// Prefer higher panning resolution over the coarser positions used by the hardware
-constexpr bool USE_NICE_PANNING = true;
+// Prefer higher panning resolution over the coarser positions used by the
+// hardware (this allos for "true center" pan positions which is not
+// possible on the real hardwre).
+constexpr bool UseNicePanning = true;
 
-// Prefer the rich sound offered by the hardware's existing partial mixer
-constexpr bool USE_NICE_PARTIAL_MIXING = false;
+// Prefer not forcing always in-phase partial mixing (this is more authentic
+// sounding).
+constexpr bool UseNicePartialMixing = false;
 
 using Rom = LASynthModel::Rom;
 
-const Rom mt32_pcm_100_f  = {"pcm_mt32", LASynthModel::ROM_TYPE::PCM};
-const Rom mt32_pcm_100_l  = {"pcm_mt32_l", LASynthModel::ROM_TYPE::PCM};
-const Rom mt32_pcm_100_h  = {"pcm_mt32_h", LASynthModel::ROM_TYPE::PCM};
+// MT-32
 const Rom mt32_ctrl_104_f = {"ctrl_mt32_1_04", LASynthModel::ROM_TYPE::CONTROL};
 const Rom mt32_ctrl_104_a = {"ctrl_mt32_1_04_a", LASynthModel::ROM_TYPE::CONTROL};
 const Rom mt32_ctrl_104_b = {"ctrl_mt32_1_04_b", LASynthModel::ROM_TYPE::CONTROL};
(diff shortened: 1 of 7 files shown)


johnnovak:
TODO: We'll need to update the MIDI section of the wiki that details the MT-32/CM-32L ROM lookup mechanism.

src/midi/midi_mt32.cpp
```diff
@@ -346,33 +444,63 @@ static std::deque<std_fs::path> get_selected_dirs()
...
 {
-	const bool is_auto = (selected_model == "auto");
-	for (const auto &model : all_models)
-		if (is_auto || model->Matches(selected_model))
-			for (const auto &dir : rom_dirs)
-				if (model->Load(service, dir))
+	auto [load_first_available, candidate_models] =
+	        [&]() -> std::pair<bool, const std::vector<const LASynthModel*>&> {
+		// Symbolic model names resolve the first available model in the
+		// given category.
```
kcgen, line 473: this is pretty confusing; it's an anonymous lambda function that populates a bool and a reference to a vector of models, and then we use those in the search loop below.

I don't have suggestions to simplifying this closer to the prior code version, but it would be handy to document it just briefly because the initial block is an eyeful! :slightly_smiling_face: 

```c++
	auto [load_first_available, candidate_models] =
	        [&]() -> std::pair<bool, const std::vector<const LASynthModel*>&>
```

src/midi/midi_mt32.cpp
```diff
@@ -483,111 +611,89 @@ static std::set<const LASynthModel*> populate_available_models(
...
+	                                                      &mt32_bluer_model,
+	                                                      &mt32_203_model,
+	                                                      &mt32_204_model,
+	                                                      &mt32_206_model,
+	                                                      &mt32_207_model};
+
+	const std::vector<const LASynthModel*> cm32_models = {&cm32l_100_model,
+	                                                      &cm32l_102_model,
+	                                                      &cm32ln_100_model};
+
```
kcgen, line 633: These local `mt32_models` and `cm32_models` vectors have a name conflict with the file-scope variables.

kcgen (commented):
That's a big improvement, @johnnovak - very nice!
Just a couple suggestions.

src/midi/midi_mt32.cpp
johnnovak: Yeah, lack of proper type inference of the return value... 😩 I can simplify to get rid of the lambda and use two mutable vars for the flag & list of candidate models if that helps.

I mean if you find things confusing at the mechanical code level. At the conceptual level, I find the logic clear with the "candidate models". But maybe I could break this into two functions with some code duplication to get rid of the boolean altogether: one function for loading versioned models, and another one for looking up the "best" symbolic model. Probably that's for the best.

src/midi/midi_mt32.cpp
johnnovak: They just shadow them 😄 , but probably best to avoid that when we're in full control over namings.

src/midi/midi_mt32.cpp
kcgen: The mechanical code (inside the lands) and the candidate loop all look good (and very similar to the original code); but I just had to paused at the lamda and think about what it's doing / why it's needed. 

So yeah, no need to change the code, but just a one-liner comment to say what we're trying to get or setup with the lamda. 

(This, being an anonymous Landa, means there isn't a function name that can act as self-documenting code.. so the reader only see the variable names and return type).

Just a minor comment though 👍

johnnovak:
Had a bit of time @kcgen  so I've addressed your comments & have rebased to latest main.

src/midi/midi_mt32.cpp
```diff
@@ -181,77 +215,141 @@ const LASynthModel cm32ln_100_model = {"cm32ln_100",
...
+	        "The Roland MT-32/CM-32ML model to use.\n"
+	        "You must have the ROM files for the selected model available (see 'romdir').\n"
+			"The lookup order for models that don't specify a version in their name is\n"
+			"performed in order as listed.\n"
+	        "  auto:       Pick the first available model (default).\n"
+	        "  mt32:       Pick the first available MT-32 model.\n"
+	        "  cm32l:      Pick the first available CM-32L model.\n"
+	        "  mt32_old:   Pick the first available \"old\" MT-32 model (v1.0x).\n"
+	        "  mt32_new:   Pick the first available \"new\" MT-32 model (v2.0x).\n"
+	        "  <version>:  Use the exact specified model version (e.g., 'mt32_204').");
```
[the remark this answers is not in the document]
johnnovak: Fixed. I've also reordered the `mt32*` model names, lemme know what you think.

src/midi/midi_mt32.cpp
```diff
@@ -483,111 +623,87 @@ static std::set<const LASynthModel*> populate_available_models(
...
-	const std::string dirs_padding(max_dir_width, ' ');
-	caller->WriteOut("%s%s", indent, dirs_padding.c_str());
-	for (const auto &model : models_without_aliases) {
-		const bool is_missing = (available_models.find(model) == available_models.end());
-		const auto color = get_highlight(is_missing);
-		caller->WriteOut("%s%s%s%s", color, model->GetVersion(),
-		                 nocolor, column_delim);
-	}
-	caller->WriteOut("\n");
+		const auto is_active = (model_and_dir->first == model);
```
kcgen, line 666: If `MIXER /LISTMIDI` is called before MT32 is initialized, the std::optional `model_and_dir` will still be empty (this generates an assert on Linux).

Tested this locally:

```c++
const auto is_active = (model_and_dir && model_and_dir->first == model);
````

Comment (johnnovak) [reply src/midi/midi_mt32.cpp:666]: I don't see how that can happen; for me `MIXER /LISTMIDI` prints nothing when `mididevice` is set to `none` in the config, and it works as expected when it's set to `fluidsynth`. That's on macOS. Not sure how you can trigger the assert?

But I can add what you proposed to make things generally more robust, I guess.

Comment (johnnovak) [reply src/midi/midi_mt32.cpp]: Done.

Comment (kcgen) [reply src/midi/midi_mt32.cpp]: > Not sure how you can trigger the assert?

On Linux, it at least comes from deep in the STL, but we can just check it before dereferencing (in the original version):

```diff
+ assert(model_and_dir);
+ const auto is_active = (model_and_dir->first == model);
```

#### Before calling `mixer /listmidi`

before-mixer-listmidi

#### After calling `mixer /listmidi`

after-mixer-listmidi

It seems like clang (even debug builds!) might be missing those internal checks, so it barrels on.. I really don't know how it's giving back a valid but empty `std::pair` value 😆

johnnovak: @kcgen Rebased to main again. So are we good now then? 😄

FeralChild64 (approved):

Issue #2998, opened 2023-10-13
Sync 3rd party ported-in dependencies prior to 0.81 testing


kcgen:
```
src/libs
├── decoders
│   └── docs
├── enet
│   └── include
├── fontstash
├── ghc
│   └── ghc
├── loguru
├── manymouse
├── mverb
├── nuked
├── PDCurses
│   ├── common
│   ├── pdcurses
│   └── sdl2_queue
├── residfp
│   └── resample
├── stb_truetype
├── tal-chorus
├── whereami
└── YM7128B_emu
```

Snipped from the above list:

```
├── sdlcd     -> Internally maintained transplant from SDL 1.x
│   ├── dummy
│   ├── linux
│   ├── macosx
│   └── win32
└── zmbv      -> Not 3rd party code
```

kcgen:
Examples of prior sync PRs:

https://github.com/dosbox-staging/dosbox-staging/pulls?q=is%3Apr+sync+label%3A%22upstream+sync%22

weirddan455:
sdl_truetype is up to date.  Hasn't had an update upstream in 2 years.  I ran a shasum and ours matches exactly with the upstream file:  https://github.com/nothings/stb/blob/master/stb_truetype.h

I believe the version of SDL CD we have is unmaintained upstream.  It's from SDL 1.  There is an implementation of SDL CD in the SDL1->2 comparability layer that is maintained:  https://github.com/libsdl-org/sdl12-compat

It would take more work than a simple sync to update though.  Maybe something to do after 0.81?

kcgen:
Thanks for poking through the list, @weirddan455.

> I believe the version of SDL CD we have is unmaintained upstream. It's from SDL 1.

Yup; SDL dropped its host-level CDROM drive interface in the transition to SDL 2.x; so when Staging moved from SDL 1.x to 2.x, this "functional regression" became a sore point that some users were quite unhappy with.

Through a [mega-PR effort](https://github.com/dosbox-staging/dosbox-staging/pull/1540), Wengier transplanted the cross-platform functionality from SDL1. I think you've helped fix and maintain this, too :blush: :+1: !

I think SDL2 is missing this functionality (and has no plans to restore it, which is fine.. CDROM drives are now retro hardware themselves.. but I understand the desire for folks to play their DOS games via original media).

Looking at the compat12 code, they've stubbed it out with MP3 playback:

[...]

PR #3013, opened 2023-10-16
Remove fontstash library

weirddan455:
# Description
This library is unused and unmaintained by upstream.

## Related issues
#2998

# Checklist

I have:


weirddan455:
It just occurred to me maybe @johnnovak was planning on using this for his OSD.  It was brought in with commit b204e392685b5f3bb2a0b7bfd9f680a12c3f10ab (committed by @kcgen) around the same time as stb_truetype

I'm just a little confused as I can't tell where the code actually came from.  It's significantly changed from upstream and upstream is also unmaintained since 2019.

If you really want to use the library, I'll close this PR but I'd suggest looking into alternatives.  I believe stb_truetype itself has some support for font atlases.  There's also rtext from the raylib library which is based on stb_truetype.  It's modular so we don't need to bring in the entire raylib dependency:  https://github.com/raysan5/raylib/blob/master/src/rtext.c

kcgen (approved):
I agree; even if this package is still desirable, it might make more sense to re-assess the state of maintained forks just prior to putting it to use in our repo.

weirddan455:
I'll wait for @johnnovak to see this before merging in case he has any objections.

johnnovak:
Yeah I was planning to use it for my OSD stuff. NanoVG + stb_truetype + fontstash is my "stack" of choice for drawing custom GUIs in my other projects.

Yes, all three are effectively "unmaintained" but it doesn't matter much because they're all very mature and self-contained libraries with no further dependencies. Are they 100% bug-free? Probably not. Have I run into a single bug during my heavy use of these libs? No, not one time.

These are not fast-moving targets like SDL, fonstash in particular doesn't depend on anything, even conceptually only on the concept of UTF-8 glyphs which is set in stone. Personally, I'm fine with it, but also happy to look into alternative implementations. In short, I don't care much either way, I don't think what we choose matters a lot. Probably you'll be able to use the current version of fonstash fine in 2050 too 😎 

That other implementation you linked is 2.1 KLOC, fonstash is 1.7 KLOC. So if you're proposed implementation can be simply plugged in, why not 😄 I'll need to think about this more, but I think in either case my plan was to rewrite fonstash in C++... or maybe just use it from C++. The reasons are a bit complicated, we'll see.

So I guess you can remove it for now if you want, that will force me to re-evaluate things when the time comes.

weirddan455:
@johnnovak  I don't particularly care if this library sits in the tree not doing anything.  At least I know why it's there now.  My main issue is that the code we've pulled in does not match the upstream source (https://github.com/memononen/fontstash).  I assume this came from one of the forks but I don't know which one.

I'm just trying to go through the list in #2998 and verify everything is up to date.  The advantage of removing it would be we don't have to worry about keeping it in sync.  If you want to use it, just pull in a  fresh copy from upstream when it's actually going to be used.

johnnovak:
> @johnnovak I don't particularly care if this library sits in the tree not doing anything. At least I know why it's there now. My main issue is that the code we've pulled in does not match the upstream source (https://github.com/memononen/fontstash). I assume this came from one of the forks but I don't know which one.
> 
> I'm just trying to go through the list in #2998 and verify everything is up to date. The advantage of removing it would be we don't have to worry about keeping it in sync. If you want to use it, just pull in a fresh copy from upstream when it's actually going to be used.

That's fine, I'll re-add this or something else if I need it. Well, you don't need to update it anyway as it's unused 😄 I'm pretty sure I copied it from the VCVRack repo.

PR #3016, opened 2023-10-16
Add game mode support for macOS Sonoma or newer

### LINKED ISSUE #3014: Enable macOS "Game Mode" for Staging

_No response_

macOS (Apple Silicon)

_No response_

Info about this is hard to find but this should be the way to go:
https://developer.apple.com/forums/thread/739387

General info about "Game Mode":
https://support.apple.com/en-us/HT213658

_No response_

_No response_

_No response_

- [X] Yes, I agree.

kcgen:
# Description

Add a field in the macOS package manifest to let macOS Sonoma users enable Game-Mode when running full screen.

## Related issues

Fixes #3014

# Manual testing

I'm unable to test it because I'm not running Sonoma.

@Burrito78 - can you check?
1. Ensure you're running macOS Sonoma on Apple silicon.
1. Move your pointer over the green button in the upper-left corner of the game window.
2. Choose Enter Full Screen from the menu that appears.

# Checklist

I have:

### DIFF
--- a/contrib/macos/Info.plist.template
+++ b/contrib/macos/Info.plist.template
@@ -44,6 +44,9 @@
 	<key>CGDisableCoalescedUpdates</key>
 	<true/>
 
+	<key>LSApplicationCategoryType</key>
+	<string>public.app-category.games</string>
+
 	<key>LSMinimumSystemVersionByArchitecture</key>
 	<dict>
 	<key>x86_64</key>
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -479,6 +479,9 @@ static void initialize_vsync_settings()
 	if (has_true(user_pref)) {
 		sdl.vsync.when_windowed.requested   = VsyncState::On;
 		sdl.vsync.when_fullscreen.requested = VsyncState::On;
+	} else if (user_pref == "adaptive") {
+		sdl.vsync.when_windowed.requested   = VsyncState::Adaptive;
+		sdl.vsync.when_fullscreen.requested = VsyncState::Adaptive;
 	} else if (has_false(user_pref)) {
 		sdl.vsync.when_windowed.requested   = VsyncState::Off;
 		sdl.vsync.when_fullscreen.requested = VsyncState::Off;
@@ -4183,7 +4186,7 @@ static void config_add_sdl() {
 	        "  <custom>:  Specify a custom rate as an integer or decimal Hz value\n"
 	        "             (23.000 is the allowed minimum).");
 
-	const char* vsync_prefs[] = {"auto", "on", "off", "yield", nullptr};
+	const char* vsync_prefs[] = {"auto", "on", "adaptive", "off", "yield", nullptr};
 	pstring = sdl_sec->Add_string("vsync", always, "auto");
 
 	pstring->Set_help(
@@ -4193,6 +4196,10 @@ static void config_add_sdl() {
 	        "  on:        Enable vsync. This can prevent tearing in some games but will\n"
 	        "             impact performance or drop frames when the DOS rate exceeds the\n"
 	        "             host rate (e.g. 70 Hz DOS rate vs 60 Hz host rate).\n"
+	        "  adaptive:  Enables vsync when the frame rate is higher than the host rate,\n"
+	        "             but disables it when the frame rate drops below the host rate.\n"
+	        "             This is a reasonable alternative on macOS instead of 'on'.\n"
+	        "             Note: only valid in OpenGL output modes; otherwise treated as 'on'.\n"
 	        "  off:       Attempt to disable vsync to allow quicker frame presentation at\n"
 	        "             the risk of tearing in some games.\n"
 	        "  yield:     Let the host's video driver control video synchronization.");
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1069,6 +1069,24 @@ static void setup_presentation_mode(FrameMode &previous_mode)
 		const auto display_might_be_interpolating = (host_rate >=
 		                                             InterpolatingVrrMinRateHz);
 
+		const auto conditions_demand_constant_host_rate = (
+#if defined(MACOSX)
+		        // Normally OpenGL drivers with a swap-interval of 1
+		        // only block for the gap in time after the application
+		        // requests a buffer swap through to presentation of the
+		        // current frame.
+		        //
+		        // However, circa-2023 macOS OpenGL drivers impose
+		        // additional per-frame block penalities if frames
+		        // /aren't/ presented. Therefore, this condition demands
+		        // we present at a constant host rate.
+		        //
+		        sdl.rendering_backend == RenderingBackend::OpenGl &&
+		        get_vsync_settings().requested == VsyncState::On);
+#else
+		        false);
+#endif
+
 		// If we're fullscreen, vsynced, and using a VRR display that
 		// performs frame interpolation, then we prefer to use a
 		// constant rate.
@@ -1087,15 +1105,22 @@ static void setup_presentation_mode(FrameMode &previous_mode)
 		        supported_rate >= dos_rate
 		                ? "Host can handle the full DOS rate"
 		                : "Host cannot handle the DOS rate");
-		LOG_MSG("SDL:   - %s",
-		        conditions_prefer_constant_rate
-		                ? "CFR selected because we're fullscreen, "
-		                  "vsync'd, and display is 140+Hz"
-		                : "VFR selected because we're not "
-		                  "fullscreen, nor vsync'd, nor < 140Hz");
+		if (conditions_demand_constant_host_rate) {
+			LOG_MSG("SDL:   - CFR selected because we're on macOS using"
+			        " OpenGL with vsync on");
+		} else {
+			LOG_MSG("SDL:   - %s",
+					conditions_prefer_constant_rate
+							? "CFR selected because we're fullscreen, "
+							"vsync'd, and display is 140+Hz"
+							: "VFR selected because we're not "
+							"fullscreen, nor vsync'd, nor < 140Hz");
+		}
 #endif
-
-		if (supported_rate >= dos_rate) {
+		if (conditions_demand_constant_host_rate) {
+			mode = FrameMode::Cfr;
+			save_rate_to_frame_period(host_rate);
+		} else if (supported_rate >= dos_rate) {
 			mode = conditions_prefer_constant_rate ? FrameMode::Cfr
 			                                       : FrameMode::Vfr;
 			save_rate_to_frame_period(dos_rate);
@@ -4198,7 +4223,6 @@ static void config_add_sdl() {
 	        "             host rate (e.g. 70 Hz DOS rate vs 60 Hz host rate).\n"
 	        "  adaptive:  Enables vsync when the frame rate is higher than the host rate,\n"
 	        "             but disables it when the frame rate drops below the host rate.\n"
-	        "             This is a reasonable alternative on macOS instead of 'on'.\n"
 	        "             Note: only valid in OpenGL output modes; otherwise treated as 'on'.\n"
 	        "  off:       Attempt to disable vsync to allow quicker frame presentation at\n"
 	        "             the risk of tearing in some games.\n"


Grounded0:
@kcgen I would rename this PR as "Add game mode support for macOS Sonoma or newer" or similar.

kcgen:
@Burrito78

> _@Grounded0 > would be nice to see PERFDOS score with that enabled_

:+1:

kcgen:
The updated manifest was accepted as part of the .dmg packaging.

(As this is a recommended tag to categorize an application, it seems useful regardless of the test results)

Grounded0:
This should be also be documented in 0.81 release notes so tagging this with documentation.

Burrito78:
PERF_DOS.EXE gives 370 with latest DEV version and 225 using this build.

Game Mode works though...mmm.

Bildschirmfoto 2023-10-16 um 22 08 52

Grounded0:
😂

Grounded0:
@Burrito78 

Can you plug your computer to AC adapter and run again.

Burrito78:
Quake

DEV
Run 1: 181 fps
Run 2: 151 fps
Run 3: 149 fps

This PR
Run 1: 128 fps
Run 2: 115 fps
Run 3: 114 fps

It would be interesting to find out what causes it to drop that much exactly.

I think we can scrap that feature currently. :(

Burrito78:
> @Burrito78
> 
> Can you plug your computer to charger and try again.

Yes, will rerun both benches tomorrow on the M2 mini.

Grounded0:
I always warn people not to run macOS releases until they hit xx.6.

kcgen:
Amazing, @Burrito78.
This is literally just a change in the MacOS manifest.

_"Of course frames rates are lower, but we deliver them at tighter latencies"_ :sweat_smile: 

Keep the test results coming :-)

Grounded0:
We can leave this PR open until Apple fixes game mode on macOS Sonoma. Still want more data points so calling all macOS 14.x/Apple silicon users to test this out.

johnnovak:
For the record, I was forced to upgrade to Sonoma yesterday too (company laptop).

Burrito78:
It gets weirder and weirder.

Machine: 	Mac mini 2023 / Apple M2
OS:			macOS Sonoma 14.0

All benchmarks runs in fullscreen mode with default config except vsync.

Shiny PERF_DOS.EXE
				vsync = auto		vsync = off		vsync = on 
Main DEV		223				221				10
PR Branch		223				225				30

Quake SW 1.06
				vsync = auto		vsync = off		vsync = on 
Main DEV		175/148/148		117/96/96		4.3/-/-
PR Branch		119/99/98		119/97/98		9.3/-/-


Shiny Benchmark:	https://daryldixonretro.111mb.de/Sonstiges.html
Quake SW 1.06:	https://archive.org/details/Quake_802

Grounded0:
@johnnovak 

Can you run game mode benchmarks on your shiny Sonoma when you have time. Need more data points.

johnnovak:
> @johnnovak
> 
> Can you run game mode benchmarks on your Sonoma when you have time. Need more data points. Do note that game mode requires Apple silicon hardware.

At some point yes, but I'm quite busy now, then offline for 3 weeks from Wednesday.

kcgen:
@Burrito78 : looks like thermal throttling.

After five back-to-back runs, all versions were getting the low ~220 score.

Game mode will unfortunately will accentuate this by running the CPU and GPU in their most power-hungry modes.

If you search for m2 throttling, there are reports about people getting defective systems that run really hot and slow. They replaced them, and the problem is fixed.

It looks like the m2 studio has a better cooling system; perhaps it wouldn't throttle?

Other benefits of game-mode would be that the performance and power are fully dedicated to the game, and not to other wasteful power hungry applications.  

So you could try running something like the multi-core Geekbench test in the background, and then simultaneously run some dosbox benchmarks; maybe this is where game-mode will pull out some wins?

Grounded0:
Shiny has no time limit so it can be left running for 30 minutes and then you get 30 minute average of the performance. That's a good way to test thermal throttling.

Grounded0:
If Mac Mini posts weak results like those after 30 minutes of running it should be reported to Apple for potential hardware defect. Because the thermal dissipation system would be fully up and running then with no issues.

Burrito78:
Running for 30+ Minutes, no throttling happening.

Bildschirmfoto 2023-10-17 um 14 06 58

Grounded0:
Btw, you should benchmark with max 105 % for maximum performance but yeah we'll wait for @johnnovak datapoint and if its similar this PR should be closed.

kcgen:
Those graphs help, thanks @Burrito78.

Is it possible to get the same charts across a 5 min run of the PR while in game mode the entire time?

I'm very curious if or how game mode influences the clock frequency ramping behavior. 

What we're seeing in the existing screenshots is Apple's CPU scheduler drip-feeding slightly increased CPU frequency (over and over, a bit more each time) combined with DOSBox's auto-cycler detecting and using those new, slightly higher CPU frequencies by ratcheting up the auto cycle count.

When the Shiny benchmark score finally grinds its way to the peak of 370, that's when Apple's scheduler is also giving it the peak CPU frequency.

My guess is that there's some slightly different dynamic at play in game-mode.

Burrito78:
> Is it possible to get the same charts across a 5 min run of the PR while in game mode the entire time?

This is the PR running at full 370 for 5 minutes.

At the first dip i switched to windowed to see how much time is left on the graph an switched back to fullscreen.
On the second dip i switched to windowed mode permanently (game mode switches off).

Bildschirmfoto 2023-10-17 um 16 04 38

kcgen:
Thanks for that last screenshot, @Burrito78 !

> This is the PR running at full 370 for 5 minutes.

Do you know what changed? (maybe, `cycles = 105`?)

In the previous DEV screenshot, the CPU speed only briefly crests 3 Ghz for short periods:



Here, the PR + game-mode ramps faster to ~3.2 Ghz, and it now scores a full 370 (and probably gets there faster):

2023-10-17_08-11-gm

I think the solution is to try to make DOSBox's auto-cycler, specifically in `cycles = max`, more demanding and clearer to the host CPU scheduler. By demanding, I mean trying to consume 100% duty cycle from the host faster and with fewer calls to `sleep()`.

Burrito78:
> Do you know what changed? (maybe, `cycles = 105`?)

Nothing changed, it's random if it will get high or low fps.
It's just so that the PR build -for whatever reason- gets high fps less often than DEV currently.

kcgen:
> > Do you know what changed? (maybe, `cycles = 105`?)
> 
> Nothing changed, it's random if it will get high or low fps. It's just so that the PR build -for whatever reason- gets high fps less often than DEV currently.

Yeah; that's definitely worse than without it.

Just trying to see if it's helpful in any capacity.. 

Can you open a terminal and paste this in to simulate a  competing load on all cores.

```shell
while true; do /usr/bin/openssl speed -multi $(($(nproc) * 3)) &> /dev/null; done
```

Then see if the PR game-mode fullscreen can produce better scores than DEV fullscreen.

Burrito78:
@kcgen macOS doesn't have 'nproc'. Can you hardwire the command line to 4 or 8?

kcgen:
Oh, I'm using `nproc` from HomeBrew. 

Sorry! Yup, no worries with hard-coding the number; use three-fold num-cores; so 30 in your case:

`while true; do /usr/bin/openssl speed -multi 30 &> /dev/null; done`

This will give DOSBox just a fraction of any given core.

If game-mode is a competent butler, it'll adjust priority and give openssl only table scraps while DOSBox feasts like a king.

Burrito78:
macBook again, but plugged-in this time. :)


DEV with stresstest, started benchmark after a short while after being sure that I was getting "high" numbers:
Bildschirmfoto 2023-10-17 um 22 05 05

PR with stresstest, 5 Minutes, only minimized for screenshot:
Bildschirmfoto 2023-10-17 um 21 52 38

The numbers of this benchmark are not very stable, btw. Can't be an average over the whole time for sure.
So take the numbers in the screenshot not as an absolute result.

Burrito78:
Apple says this about game mode:

> Game Mode optimizes your gaming experience by giving your game the highest priority access to your CPU and GPU, lowering usage for background tasks. And it doubles the Bluetooth sampling rate, which reduces input latency and audio latency for wireless accessories like game controllers and AirPods.

I don't see any negative impact and we can't test every advertised aspect of it.

 Currently I'm seeing two issues here for which I think we should have separate tickets:

-Apple M Performance-Cores randomly don't scale to full frequencies on macOS.
-'sync = on' is currently broken on macOS (Sonoma?), see the table in https://github.com/dosbox-staging/dosbox-staging/pull/3016#issuecomment-1765888544

kcgen:
Thanks @Burrito78 - will have a follow up commit switching to adaptive vsync, to solve the second point.

Burrito78:
Updates with newest PR build including 'adaptive sync'.

Machine: Mac mini 2023 / Apple M2 OS: macOS Sonoma 14.0

All benchmarks run in fullscreen mode with default config except vsync.
I started the benchmarks multiple times to make sure i’m always in the „high fps“ mode (this is currently a Staging or macOS bug).

```
Shiny PERF_DOS.EXE v1.4

			vsync = auto	vsync = off	vsync = on 
Main DEV		377		375		7
PR Branch		379		378		232
```

Burrito78 (approved):
Results show only positive impact and changes follow Apples guidelines.

Grounded0:
Still one benchmark with latest changes and i'm happy to sign this off if we make sure release notes bring up that we use game mode by default on Sonoma or newer as per Apple guidelines.

Grounded0:
Anyone who doesn't like this behaviour can comment off that line from their Info.plist.

Grounded0:
Pinging @kklobe if he's happy with these changes.

kcgen:
Thanks @Burrito78.

---

Before going ahead, let's put adaptive vsync to the test to see if has down-sides versus the baseline. My understanding from the documentation is that there are none, regardless of the display type (fixed Hz vs variable rate).

The testing will take about 5 minutes.

---

## Vsync Test Pack

[:file_folder: **`vsync-pack.zip`**](https://github.com/dosbox-staging/dosbox-staging/files/13030462/vsync-pack.zip)

Each game comes with an included `dosbox.conf`.

Open your terminal, enter the game's directory, and launch your DOSBox Staging build with the `--noprimaryconf` flag.

For example:

``` shell
cd dangerous_dave-60Hz-EGA-VFR
/path/to/dosbox-staging/dosbox --noprimaryconf
cd .. 
```

### Dangerous Dave (EGA, 60 Hz, VFR)

- The intro sequence has a huge static image that pans left to right, which should be free from mid-frame tearing or displacement.


### Keen (VGA, 70 Hz, VFR)

- The intro sequence has huge scrolling graphical text plus vertical scrolling credits.

- The vertical lines should be free from mid-frame tearing displacement, and the credits should be smooth without juddering.


### Pinball Dreams (VGA, 60 Hz, CFR)

- Let the intro splash screens go by until credits scroll vertically. Watch for signs of vertical judder.
- At the game selection screen, press F4 for the Nightmare board.
- When the board is panned vertically, watch for tearing and judder.

### Jazz Jackrabbit (VGA Mode X, 60 Hz, VFR)

- In-game frames are rendered at variable rates depending on what's happening.   
  
- The background should be tear-free when running and jumping.

Burrito78:
@kcgen How about automatically setting 'vsync = adaptive' on macOS if 'vsync = on' is used? I'm thinking about existing conf files here. This should be indicated in logging of course to prevent confusion.

Grounded0:
I personally would like game mode to be enabled without any vsync fixes for now.

Burrito78:
Will test Windows on Monday if nobody else can do it by then.

kcgen:
> are there cases where 'vsync = on' actually makes sense on macOS?

Yes, when it comes to tearing prevention, `vsync = on` is useful and behaves correctly on all platforms.

It simply passes that request through to the OpenGL drivers, and users can expect real vsync behavior. There some caveats out of our control though:

- The OpenGL drivers might not honor the vsync request (some users disable vsync universally. Well, if that's what they want, then it won't work here either).

- DOS-level tearing might still be happening if the game writes to the video RAM as fast as possible, like Hexen does. So `vsync = on` will simple draw these pre-torn frames as they would have been torn on a real DOS machine.

- DOOM, on the other hand, runs at 35 Hz and draws on vblank. @GranMinigun has a handy suggestion to detect if your host is tearing: get close to the pillars on the first level and rotate side-to-side and look for tearing along the verical column edge.  With `vsync = on`, there should be none.

> How about automatically setting 'vsync = adaptive' on macOS if 'vsync = on' is used?

This PR lets users set `vsync = adaptive` for OpenGL (on all platforms), but I'd rather not move the chairs under the hood. If a user sets `vsync = on`, then that's what they get - and same  with `vsync = adaptive`.

kcgen:
@Burrito78 , thanks for all the testing and showing that this new game-mode feature is actually no worse than baseline.

Also, both you and I confirmed on macOS that:

```ini
[sdl]
vsync = on
presentation_mode = auto
```

Comes with a large performance penalty because of excessive vsync block times if frames aren't provided at the host-rate.

So this PR has `auto` presentation mode select CFR on macOS when vsync is on.  It's a tiny refinement to avoid this pitfall.  Without it, things are a lot worse.

Based on this, I'll go ahead and merge.

The only change this PR does in when `presentation_mode = on

PR #3023, opened 2023-10-17
Updated modem CONNECT response behavior

NicknineTheEagle:
# Description

- Renamed modem "baudrate" parameter to "bps" (bits per second). This is done because CONNECT response reports established connection's _speed_ not baud rate. Modem speed and modem baud rate are two different things as described here: https://www.linuxjournal.com/files/linuxjournal.com/linuxjournal/articles/010/1097/1097s2.html
- Reported modem speed now matches port baud rate by default. According to this article, later modems would usually synchronize their speed to port baud rate: https://tldp.org/HOWTO/Remote-Serial-Console-HOWTO/preparation-setspeed.html This should make things easier for the user in cases like Doom 1.2 which mandates modem speed to be 9600.
- You can now use AT+BPS command to change reported modem speed at runtime.

# Manual testing

Tested with Doom 1.2 and Doom 1.9. Doom 1.2 sets port baud rate to 9600 so modem returns CONNECT 9600 response. Doom 1.9 sets port baud rate to 14400 by default to modem returns CONNECT 14400 response.

# Checklist

I have:

### DIFF
--- a/include/serialport.h
+++ b/include/serialport.h
@@ -49,6 +49,9 @@
 #include "hardware.h"
 #endif
 
+constexpr uint32_t SerialMinBaudRate = 300u;
+constexpr uint32_t SerialMaxBaudRate = 115200u;
+
 // Serial port interface
 #define SERIAL_IO_HANDLERS 8
 #define SERIAL_MAX_FIFO_SIZE 256
--- a/include/serialport.h
+++ b/include/serialport.h
@@ -291,6 +291,8 @@ public:
 	bool Getchar(uint8_t *data, uint8_t *lsr, bool wait_dsr, uint32_t timeout);
 	uint8_t GetPortNumber() const { return port_index + 1; }
 
+	uint32_t GetPortBaudRate() const;
+
 	// What type of port is this?
 	SERIAL_PORT_TYPE serialType = SERIAL_PORT_TYPE::DISABLED;
 
--- a/src/hardware/serialport/directserial.cpp
+++ b/src/hardware/serialport/directserial.cpp
@@ -247,7 +247,7 @@ bool CDirectSerial::doReceive() {
 
 // updatePortConfig is called when emulated app changes the serial port
 // parameters baudrate, stopbits, number of databits, parity.
-void CDirectSerial::updatePortConfig(uint16_t divider, uint8_t lcr)
+void CDirectSerial::updatePortConfig(uint16_t, uint8_t lcr)
 {
 	uint8_t parity = 0;
 
@@ -262,7 +262,7 @@ void CDirectSerial::updatePortConfig(uint16_t divider, uint8_t lcr)
 	uint8_t bytelength = (lcr & 0x3) + 5;
 
 	// baudrate
-	const uint32_t baudrate = divider ? 115200 / divider : 115200;
+	const uint32_t baudrate = GetPortBaudRate();
 
 	// stopbits
 	uint8_t stopbits;
--- a/src/hardware/serialport/serialport.cpp
+++ b/src/hardware/serialport/serialport.cpp
@@ -1289,6 +1289,14 @@ bool CSerial::Putchar(uint8_t data, bool wait_dsr, bool wait_cts, uint32_t timeo
 	return true;
 }
 
+uint32_t CSerial::GetPortBaudRate() const {
+	if (baud_divider == 0) {
+		return 115200;
+	}
+
+	return 115200 / baud_divider;
+}
+
 class SERIALPORTS final : public Module_base {
 public:
 	SERIALPORTS (Section * configuration):Module_base (configuration) {
--- a/src/hardware/serialport/serialport.cpp
+++ b/src/hardware/serialport/serialport.cpp
@@ -1018,10 +1018,9 @@ static constexpr std::tuple<uint8_t, uint8_t> baud_to_regs(uint32_t baud_rate)
 	// Cap the lower-bound to 300 baud. Although the first 1950s modem
 	// offered 110 baud, by the time DOS was available 8-bit ISA modems
 	// offered at least 300 and even 1200 baud.
-	baud_rate = std::max(300u, baud_rate);
+	baud_rate = std::max(SerialMinBaudRate, baud_rate);
 
-	constexpr auto max_baud = 115200u;
-	const auto delay_ratio = static_cast<uint16_t>(max_baud / baud_rate);
+	const auto delay_ratio = static_cast<uint16_t>(SerialMaxBaudRate / baud_rate);
 
 	const uint8_t transmit_reg = delay_ratio & 0xff;  // bottom byte
 	const uint8_t interrupt_reg = delay_ratio >> 8;   // top byte
@@ -1291,10 +1290,10 @@ bool CSerial::Putchar(uint8_t data, bool wait_dsr, bool wait_cts, uint32_t timeo
 
 uint32_t CSerial::GetPortBaudRate() const {
 	if (baud_divider == 0) {
-		return 115200;
+		return SerialMaxBaudRate;
 	}
 
-	return 115200 / baud_divider;
+	return SerialMaxBaudRate / baud_divider;
 }
 
 class SERIALPORTS final : public Module_base {
--- a/src/hardware/serialport/softmodem.cpp
+++ b/src/hardware/serialport/softmodem.cpp
@@ -153,21 +153,18 @@ CSerialModem::CSerialModem(const uint8_t port_idx, CommandLine *cmd)
 	}
 
 	// Get the connect speed to report
-	constexpr auto min_bps = 300u;
-	constexpr auto max_bps = 57600u;
 	if (getUintFromString("bps:", val, cmd)) {
-		val = clamp(val, min_bps, max_bps);
+		modem_bps_config = val;
 	} else {
-		val = max_bps;
+		modem_bps_config = 57600;
 	}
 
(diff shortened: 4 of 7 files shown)


src/hardware/serialport/serialport.cpp
```diff
@@ -1289,6 +1289,14 @@ bool CSerial::Putchar(uint8_t data, bool wait_dsr, bool wait_cts, uint32_t timeo
 	return true;
 }
 
+uint32_t CSerial::GetPortBaudRate() const {
+	if (baud_divider == 0) {
+		return 115200;
```
johnnovak, line 1294: This is the max baud rate, right? Maybe extract it to some common header as a named constant as it's being used in the other files you touched too.

src/hardware/serialport/softmodem.cpp
```diff
@@ -355,6 +347,37 @@ char CSerialModem::GetChar(char * & scan) const {
 	return ch;
 }
 
+void CSerialModem::SetModemSpeed(uint32_t cfg_val) {
+	modem_bps_config   = cfg_val;
+	uint32_t reported_bps = UpdateConnectString();
```
johnnovak, line 352: Could be `const auto`?

src/hardware/serialport/softmodem.cpp
```diff
@@ -516,6 +539,13 @@ void CSerialModem::DoCommand()
 				}
 				break;
 			}
+			// Set reported connection speed.
+			if (is_next_token("BPS", scanbuf)) {
+				scanbuf += 3;
+				const uint32_t requested_bps = ScanNumber(scanbuf);
```
johnnovak, line 545: Could be `auto`?

src/hardware/serialport/softmodem.h
```diff
@@ -191,6 +190,8 @@ class CSerialModem final : public CSerial {
 	void AcceptIncomingCall();
 	uint32_t ScanNumber(char *&scan) const;
 	char GetChar(char * & scan) const;
+	void SetModemSpeed(uint32_t cfg_val);
```
johnnovak, line 193: The argument could be `const`.

src/hardware/serialport/softmodem.cpp
```diff
@@ -355,6 +347,37 @@ char CSerialModem::GetChar(char * & scan) const {
...
+		        GetPortNumber());
+	}
+}
+
+uint32_t CSerialModem::UpdateConnectString() {
+	constexpr auto min_bps = 300u;
+	constexpr auto max_bps = 115200u;
+	uint32_t connect_val;
+	if (modem_bps_config != 0) {
+		connect_val = clamp(modem_bps_config, min_bps, max_bps);
```
johnnovak, line 371: Question: shouldn't we log a warning about using out of bounds rates and informing the user about what value we ended up using?

johnnovak:
Nice clean enhancement @NicknineTheEagle 🎉 I've learned today something about baud vs bps rates, interesting! Just some minor comments / questions, and as I never use the modem stuff in DOSBox myself, I'll rely on your experience and testing here.

src/hardware/serialport/softmodem.cpp
```diff
@@ -355,6 +347,44 @@ char CSerialModem::GetChar(char * & scan) const {
...
+			            connect_val);
+		}
+	} else {
+		connect_val = GetPortBaudRate();
+	}
+
+	assert(connect_val >= SERIAL_MIN_BAUDRATE &&
+	       connect_val <= SERIAL_MAX_BAUDRATE);
+	safe_sprintf(connect_string, "CONNECT %d", connect_val);
+	return connect_val;
```
kcgen, line 385: These two functions serve to set `connect_string`. Because of this, suggest they be collapsed into a single simplified `UpdateConnectString()` that reads the `modem_bps_config` member variable.

This will also let you log the port rate in the case where the `modem_bps_config` is zero.

It will also let you assess if the `modem_bps_config` and port baud rate are compatible - for example, the `modem_bps_config` shouldn't be able to exceed the port rate (I think?).

include/serialport.h
```diff
@@ -49,6 +49,9 @@
 #include "hardware.h"
 #endif
 
+#define SERIAL_MIN_BAUDRATE 300u
+#define SERIAL_MAX_BAUDRATE 115200u
```
kcgen, line 53: Suggest `constexpr uint32_t` and CamelCase, like other newly added constants.

Making them real variables allow their type to be assessed and also show up in debuggers by name (instead of literal numbers).

src/hardware/serialport/directserial.cpp
```diff
@@ -262,7 +262,8 @@ void CDirectSerial::updatePortConfig(uint16_t divider, uint8_t lcr)
 	uint8_t bytelength = (lcr & 0x3) + 5;
 
 	// baudrate
-	const uint32_t baudrate = divider ? 115200 / divider : 115200;
+	const uint32_t baudrate = divider ? SERIAL_MAX_BAUDRATE / divider
+	                                  : SERIAL_MAX_BAUDRATE;
```
kcgen, line 266: This code here is a dupe of the `GetPortBaudRate()` function.

Suggest de-duping them using a common static function with this logic, and use it in both/all places.

kcgen (commented):
Overall looks good, @NicknineTheEagle.
Just a couple comments.

src/hardware/serialport/serialport.cpp
```diff
@@ -1018,10 +1018,9 @@ static constexpr std::tuple<uint8_t, uint8_t> baud_to_regs(uint32_t baud_rate)
 	// Cap the lower-bound to 300 baud. Although the first 1950s modem
 	// offered 110 baud, by the time DOS was available 8-bit ISA modems
 	// offered at least 300 and even 1200 baud.
-	baud_rate = std::max(300u, baud_rate);
+	baud_rate = std::max(SERIAL_MIN_BAUDRATE, baud_rate);
 
-	constexpr auto max_baud = 115200u;
-	const auto delay_ratio = static_cast<uint16_t>(max_baud / baud_rate);
+	const auto delay_ratio = static_cast<uint16_t>(SERIAL_MAX_BAUDRATE / baud_rate);
```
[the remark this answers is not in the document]
NicknineTheEagle: I don't know, I didn't write this function.

src/hardware/serialport/softmodem.cpp
NicknineTheEagle: If you read the code closer you'll see that the former function is called specifically when the user updates modem speed whereas the latter is called individually whenever serial port config is updated. I've split them specifically so that logging is only done when user updates modem speed because updatePortConfig is called a lot and would cause log spam.

kcgen: I read close enough that they could be merged, but wasn't aware of the logging frequency. Sounds good.

src/hardware/serialport/softmodem.cpp
```diff
@@ -355,6 +347,44 @@ char CSerialModem::GetChar(char * & scan) const {
...
+}
+
+uint32_t CSerialModem::UpdateConnectString() {
+	uint32_t connect_val;
+	if (modem_bps_config != 0) {
+		connect_val = clamp(modem_bps_config,
+		                    SERIAL_MIN_BAUDRATE,
+		                    SERIAL_MAX_BAUDRATE);
+		if (connect_val != modem_bps_config) {
+			LOG_WARNING("SERIAL: Port %" PRIu8 " modem's requested speed out of "
```
kcgen, line 373: This says it's clamping, but the modem_bps_config is never updated, so the warning will spam the log on repeated `updatePortConfig(...)` calls.

src/hardware/serialport/softmodem.cpp
```diff
@@ -355,6 +347,44 @@ char CSerialModem::GetChar(char * & scan) const {
...
+		        reported_bps);
+	} else {
+		LOG_MSG("SERIAL: Port %" PRIu8 " modem will synchronize reported speed "
+		        "to port baud rate",
+		        GetPortNumber());
+	}
+}
+
+uint32_t CSerialModem::UpdateConnectString() {
+	uint32_t connect_val;
```
kcgen, line 367: Initialize it to zero (or rework the body to make it a const and only set it once).

src/hardware/serialport/softmodem.cpp
```diff
@@ -355,6 +347,44 @@ char CSerialModem::GetChar(char * & scan) const {
...
+	if (modem_bps_config != 0) {
+		connect_val = clamp(modem_bps_config,
```
kcgen, line 369: Suggest bounding it by `GetPortBaudRate()`, because that's the underlying port's actual baud limit. 

Then the asserts will still sanity checks bounds at the end.

src/hardware/serialport/serialport.cpp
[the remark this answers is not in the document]
kcgen: Ok, let's consider it out of scope.

src/hardware/serialport/softmodem.cpp
NicknineTheEagle: Then there's no point in default mode (where modem_bps_config is 0). I can just default modem_bps_config to 57600 and clamp it to port speed.

kcgen: Nice simplification!

kcgen (approved):
Nice improvement, @NicknineTheEagle 👍

PR #3028, opened 2023-10-19
Fix extended ASCII codes not being rendered in batch files

weirddan455:
# Description
Regression from https://github.com/dosbox-staging/dosbox-staging/commit/039532b95b04da2538a9ff806d9bef7193c1c14f

It was doing a `static_cast<char>` plus a signed comparison of characters.  Technically this is implementation defined and `char` can be unsigned but at least on my machine (Linux x86_64), it is signed.

This resulted in the extended ASCII codes (greater than 127) being cast to negative numbers and failing the inclusion check.

The fix is very simple:

meme

## Related issues
#3027

This fix does not help with the non-UTF-8 .conf files.  Those still render incorrectly.  However, copying the contents of autoexec in those same problematic .conf files to a .bat file now does render correctly.

# Manual testing
- Copied the .conf file from above issue to .bat.
- Removed everything except the echo commands.
- Execute the .bat
- It now renders correctly.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #3029, opened 2023-10-20
Add legacy codepage to UTF-8 script

rderooy:
# Description

As requested in #3028, the simple script I created to convert files in legacy codepage format to UTF-8.

## Related issues

#3028

# Manual testing

- Converted cp437 input file to UTF-8
- Attempted to convert binary file to UTF-8
- Attempted to convert file with incorrect codepage

# Checklist

I have:

### DIFF
--- a/contrib/scripts/codepage_to_utf8.py
+++ b/contrib/scripts/codepage_to_utf8.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: MIT
+#
+# Copyright (C) 2023 rderooy <rderooy@users.noreply.github.com>
+
+"""
+Convert file from legacy codepage to Unicode UTF-8
+
+For example to convert a legacy GOG dosbox config file such that menus are correctly drawn.
+"""
+import argparse
+import os
+import sys
+
+parser = argparse.ArgumentParser(
+    description="Convert file from legacy codepage to Unicode UTF-8",
+    formatter_class=argparse.RawTextHelpFormatter,
+)
+parser.add_argument("filename", type=str, help="filename to convert")
+parser.add_argument(
+    "--cp", type=int, default=437, help="codepage to convert from (default: 437)"
+)
+parser.add_argument(
+    "-o", "--output", type=str, help="output filename (default: overwrite input file)"
+)
+args = parser.parse_args()
+
+if not args.output:
+    args.output = args.filename
+
+try:
+    with open(
+        os.path.expanduser(args.filename), "r", encoding="cp" + str(args.cp)
+    ) as input_file:
+        lines = input_file.read()
+except FileNotFoundError:
+    print("Error: could not locate the input file")
+    sys.exit(1)
+except LookupError:
+    print(f"Error: Unknown encoding '{args.cp}'")
+    sys.exit(1)
+except UnicodeDecodeError:
+    print("Error: input file does not appear to be a text file")
+    sys.exit(1)
+except PermissionError:
+    print("Error: permission denied trying to read the file")
+    sys.exit(1)
+
+try:
+    with open(os.path.expanduser(args.output), "w", encoding="utf-8") as output_file:
+        output_file.write(lines)
+except PermissionError:
+    print("Error: permission denied trying to write to the file")
+    sys.exit(1)
+
+print(f"File '{args.filename}' converted to UTF-8")


contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,36 @@
...
+if not args.output:
+    args.output = args.filename
+
+try:
+    with open(os.path.join
+                  (path, './' + args.filename), 'r', encoding='cp' + str(args.cp)) as input_file:
+        lines = input_file.read()
+        with open(os.path.join(path, './' + args.output), 'w', encoding='utf-8') as output_file:
+            output_file.write(lines)
+            output_file.close()
```
weirddan455, line 28: Correct me if I'm wrong with my limited Python knowledge.  It seems you're reading the entire file into memory with `lines = input_file.read()`?  That's fine, and probably needed if you want to overwrite the existing file.

I would suggest closing the input file before writing the output.  I believe files get automatically closed at the end of the `with` block.  If my assumptions are correct:

1. Remove a layer of indention from the 2nd `width` block.
2. Remove `output_file.close()` as that should be handled automatically.

contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,36 @@
...
+except LookupError:
+    print(f"Error: Unknown encoding '{args.cp}'")
+    sys.exit(1)
+except UnicodeDecodeError:
```
weirddan455, line 32: Is this the correct exception if an error happens when decoding from codepage 437?  It's obviously not Unicode but maybe Python throws this regardless?

contrib/scripts/codepage_to_utf8.py
rderooy: Yeah, it gets closed automatically, the close() is indeed not strictly needed.

But the indentation from the 2nd with block is required.
The first 'with' line had to be broken into two lines it exceeded 100 characters, which caused the linter to fail.

contrib/scripts/codepage_to_utf8.py
rderooy: The LookupError exception will occur if python does not find the encoding. So for instance if you pass `--cp 1`, it will fail with that error.

The UnicodeDecodeError will occur if it cannot decode the input file as the codepage specified. This is typically because the file is a binary file.

weirddan455: Yeah, the LookupError makes sense.  I was just wondering if it actually throws a Unicode error when attempting to decode from code page 437 (not Unicode).  Seems like a badly named exception if that's the case.  But if it works, then sure no problem.

contrib/scripts/codepage_to_utf8.py
weirddan455: Significant white-space problems... :roll_eyes: 

I was just worried about potential issues caused by holding open a file handle while modifying the same file under a different handle.

rderooy:
I made a few updates.
- it now handles things like tilde for the home directory
- it now handles a few more error cases, like not being able to locate the input file, or some kind of permission error.

contrib/scripts/codepage_to_utf8.py
rderooy: Indeed it does throw that exception.

contrib/scripts/codepage_to_utf8.py
rderooy: Not sure what those significant whitespace problems are. An a empty line here and there to separate code blocks? If you are talking indentation, that is required by Python.

weirddan455: I was just poking fun at Python for having significant whitespace and it apparently causing problems with the linter.  You can ignore that comment.

I see you've made some changes so now the first `open` fits on one line.  Can you just remove the indention from the second `with` block that outputs (putting on the same level on indentation as the input block)?  That way it happens after the input file is closed.  That's all I'm asking.

contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,38 @@
+#!/bin/python3
```
kcgen, line 1: Suggest: `#!/usr/bin/env python3`

This an improvement over our previous python scripts in the repo, which are still using `#!/usr/bin/python3`, which isn't as compatible for virtual environments. 

More info here: https://discuss.python.org/t/pep-394-clarify-bullet-for-usr-bin-env-python/7953/2

> *such a clarification would also benefit PEP 397 as the Python Launcher admits “virtual” shebangs, so it will provide further impetus for Windows users to add `#!/usr/bin/env python3` to their scripts to ensure the greatest degree of cross-platform compatability between Linux, MacOS, and Windows.*

contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+Convert file from legacy codepage to Unicode UTF-8
+"""
```
kcgen, line 4: Add an SPDX and copyright line to clarify the file's license. 

Use `scripts/count-pvs-bugs.py` as an example. 

Suggest a bit more explanation about the tool, and can mentioning that the primary use of this is to convert GOG conf files, as an example use case.

contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,38 @@
...
+import argparse
+import os
+import sys
+
+parser = argparse.ArgumentParser(description="Convert file from legacy codepage to Unicode UTF-8",
```
kcgen, line 9: Use pip to install 'black', which is a very robust (and opinionated! 😅) python formatter.

You can then execute 'black my script.py' and reformat it. 

This is what I've been using, and it works great.

contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,38 @@
...
+                                 formatter_class=argparse.RawTextHelpFormatter)
+parser.add_argument("filename", help="filename to convert")
```
kcgen, line 11: The `filename` argument's type can be upgraded from mere string to a read-only file, added after the help key, like: `help="...", type=argparse.FileType('r'))`.
                      
Also suggest adding the `required=True` key, because this is a necessary arg.

contrib/scripts/codepage_to_utf8.py
```diff
@@ -0,0 +1,38 @@
...
+parser.add_argument("--cp", type=int, default=437, help="codepage to convert from (default: 437)")
+parser.add_argument("-o", "--output", type=str,
```
kcgen, line 13: Same as above, but the file type can be for writing and you can even set the encoding:

`FileType('w', encoding='UTF-8')`

contrib/scripts/codepage_to_utf8.py
kcgen: The parser will take care of checking that the file exists and is readable, and you'll get the file object from the parser, so this eliminates some code further down.

rderooy: I will have a look at it the various comments over the weekend. Regarding the `required=True` key, that is only needed for non-positional arguments. positional arguments, which `filename` is, are automatically required.

OpenRift412:
Sorry, don't wanna be a bother, but would there perhaps be a way to have it so Staging will call this script during launch to automatically convert any conf files being used by the current instance?

kcgen:
> Sorry, don't wanna be a bother, but would there perhaps be a way to have it so Staging will call this script during launch to automatically convert any conf files being used by the current instance?

No; the executable won't be making system-calls or launching scripts from inside the binary.  If you want to always run this script on every launch, then you can make a wrapper-script that does two things:
1) Call this conversion script for the given .conf arguments
2) Launches DOSBox Staging with the converted conf files

contrib/scripts/codepage_to_utf8.py
rderooy: Done, let me know if you think this is sufficient.

contrib/scripts/codepage_to_utf8.py
rderooy: Done, although I'm not a fan of the way it split lines even further.

contrib/scripts/codepage_to_utf8.py
rderooy: I had thought of it originally, but did not go down the path of using `argparse.FileType()`. I gave it a quick try and ran into various hurdles.

One problem is that the output file argument is currently optional, so while I can use it for the output file, I still need to handle the case where I need to open the input file as write. This just adds more code with no benefit.

Another problem is the input file, because at that point, I don't yet know the encoding. Unless I want to hard code the script to only handle cp437, which I don't want to do.

contrib/scripts/codepage_to_utf8.py
rderooy: see above

kcgen (approved):

kcgen:
Looks good, @rderooy 👍

Issue #2956, opened 2023-09-22
Wrong file / directory attribute handling by DOSBox commands


FeralChild64:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

GIT main

### What Operating System are you using?

Linux x86_64

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

At least some DOSBox commands do not handle file / directory attributes properly - for example:
- `copy` command does not ignore files with Hidden / System attributes set (contrary to MS-DOS 6.22)
- `dir` command displays Hidden / System files and directories, always (contrary to MS-DOS 6.22)

These should be fixed, as it might influence some software installers, also after https://github.com/dosbox-staging/dosbox-staging/pull/2951 the bugs are visible not only on mounted FAT partitions or floppy / CD-ROM images, but also on local filesystem.

The following commands and programs should be tested and fixed if necessary:

- `cd` / `chdir`
- `copy`
- `del` / `delete`
- `dir`
- `erase`
- `rd`
- `ren` / `rename`
- `rmdir`
- `lh` / `loadhigh`
- `COMMAND.COM` (launching `.EXE`, `.COM` or `.BAT` files with Hidden / System attributes)
- `ATTRIB.COM`
- `MORE.COM`
- `TREE.COM`
- `MOVE.EXE`

The following points should be checked against MS-DOS 6.22:
- Does the command ignore input files (check both single file and wildcard match) with Hidden / System attribute?
- What are the attributes of files produced by the command (they might depend on the input file attributes)?
- Does the original command have switches related to file attributes, which are not implemented in DOSBox?

I would recommend to implement https://github.com/dosbox-staging/dosbox-staging/issues/2663 before fixing the commands, though.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Nice plan and overview of the problem @FeralChild64. I think the first step without touching a single line of code would be to create a big spreadsheet and document the MS-DOS 6.22 behaviour as per your description. Without that, this change cannot be reviewed or tested meaningfully and I doubt anybody can hold all these details in their heads.

The behaviour could be tested in a VM running real MS-DOS 6.22. @Torinde, I unironically think you are good at this type of detail oriented work, would you be interested in testing all these things and documentig them?

Torinde:
I can do that, thanks!

[...]

Issue #1336, opened 2021-10-19
Suggested Improvements to Staging's Implementation of the LS Command


MasterO2:
Looking at the LS command that is now included with Dosbox-Staging, the following switches would also be useful to have included as well if that's possible:

```
-d, --directory  list directories themselves, not their contents
-h, --human-readable
-l, use a long listing format
-R, --recursive  list subdirectories recursively
-X, sort alphabetically by entry extension

https://www.man7.org/linux/man-pages/man1/ls.1.html
```
Also, for large directories that take more than one page to display, LS should be pipe-able to the `more` and `less` commands, since Staging's LS itself does not have the ability to scroll up or down.

Grounded0:
> Also, for large directories that take more than one page to display, LS should be pipe-able to the `more` and `less` commands, since Staging's LS itself does not have the ability to scroll up or down.

It already pipes just fine to more and less last time i checked.

brandonrobertz:
> As i said i have non-standard installation so if someone has energy to check against bog-standard DOSBox Staging install to confirm it actually doesn't work then thank you.

I've confirmed what  MasterO2 is saying on a default dosbox-staging environment. Running FreeDOS is probably getting you a pipe that works more robustly.

MasterO2:
@Wengier In addition to adding the ability to pipe commands to Staging, would you also be willing to add the above improvements to the LS command?

Wengier:
@MasterO2 As a command originally from *nix systems, LS command has a lot of *nix-like options. I hope to clean up and convert some particularly useful ones to DOS style (e.g. ``/d`` instead of ``-d`` or ``--directory``) to work in DOSBox (which is a DOS emulator). But the details can be discussed.

Wengier:
Piping is supported via PR #1517.

kcgen:
Thanks for these additions @Wengier .

Regarding LS, it's only purpose is to keep muscle-memory intact for users primarily coming from Linux / macOS / BSD / MSYS2. They can continue to rapidly type `ls` (and common flags) without having to think twice. 

So we don't want to take away it's handful of existing GNU flags: doing so would confuse that muscle-memory: _"Wait, does this one need -d or /d?"_). 

That said, certainly no harm in adding DOS-style slash handling for the existing GNU-shorthand (single-dash) flags!

Wengier:
[...]

PR #3034, opened 2023-10-21
Rework and improve `LS.COM` command

FeralChild64:
# Description

- Moved `LS.COM` command implementation to `program_ls.cpp` file, cleaned up the code a little.
- Added support for multiple directories / patterns in the command line.
- Command does not display files and directories with Hidden or System attributes anymore, unless `/a` switch is given. 
Rationale: GNU implementation also skips hidden files by default; there are no system files on Unix, and DOS commands which skips hidden files typically skip system files too.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/2956 (partial fix only)

# Manual testing

Played with the command in  various directories, also containing hidden/system files. Sanitizer build did not reveal any issues.

# Checklist

I have:

### DIFF
--- a/src/dos/program_ls.cpp
+++ b/src/dos/program_ls.cpp
@@ -1,7 +1,7 @@
 /*
  *  SPDX-License-Identifier: GPL-2.0-or-later
  *
- *  Copyright (C) 2020-2022  The DOSBox Staging Team
+ *  Copyright (C) 2020-2023  The DOSBox Staging Team
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


MasterO2:
@FeralChild64 Are you familiar with the issue below as well?

https://github.com/dosbox-staging/dosbox-staging/issues/1336

FeralChild64:
@MasterO2 A lot of GNU ls command functionality is missing, I didn't intend to add everything - I primarily want to make our commands handle file attributes properly, support for multiple patterns and directories was actually very easy to add.

src/dos/program_ls.cpp
```diff
@@ -22,14 +22,240 @@
...
+	const size_t num_columns = column_widths.size();
+
+	constexpr int ansi_blue  = 34;
+	constexpr int ansi_green = 32;
+
+	auto write_color = [&](const int ansi_color,
+	                       const std::string& txt,
+	                       const uint8_t width) {
+		const auto margin_size = static_cast<int>(width - txt.size());
+		WriteOut("\033[%d;1m%s\033[0m%-*s", ansi_color, txt.c_str(), margin_size, "");
```
kcgen, line 147: > ```c++
> constexpr int ansi_blue  = 34;
> constexpr int ansi_green = 32;
> WriteOut("\033[%d;1m%s\033[0m% ...);
> ```

We can make use of the `convert_ansi_markup()` routine:


```c++
auto write_color = [&](const std::string& colour,
                       const std::string& txt,
                       const uint8_t width) {
	const auto ansi_txt = convert_ansi_markup(colour + txt + "[reset]");
	const auto margin_size = static_cast<int>(width - txt.size());
	WriteOut("%s%-*s", ansi_txt.c_str(), margin_size, "");
};
```

To replace the  to replace the ANSI numbers and escape sequences with the colour tags:

```c++
write_color("[color=light-blue]", name, column_width);
write_color("[color=light-green]", name, column_width);
```

kcgen: > `const auto margin_size = static_cast<int>(width - txt.size());`

Is there any risk of this going negative?

kcgen (commented):
Just a couple suggestions.

It played with some of the patterns and they work selectively:



Hopefully the patterns can be more flexible, so people can use it for bulk discovery of files.. like:

`ls */setup.exe`, and find all the games with a setup tool.

src/dos/program_ls.cpp
```diff
@@ -22,14 +22,240 @@
...
+		return;
+	}
+
+	// Sort directory contents
+
+	constexpr bool reverse_order = false;
+	DOS_Sort(dir_contents,
+	         ResultSorting::ByName,
+	         reverse_order,
+	         ResultGrouping::NonFilesFirst);
```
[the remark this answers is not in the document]
FeralChild64: Actually, I intend to rethink this API one day, so that we can support ordering in `DIR` command like in latest Windows: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/dir

src/dos/program_ls.cpp
FeralChild64: Changed the code to use `convert_ansi_markup` - but I'm calling them before the loop starts, to generate `ansi_blue`, `ansi_green` and `ansi_reset` as strings just once.

We shouldn't go negative (or even 0), unless there is a bug in `LS::GetFileNameLengths` or `LS::GetColumnWidths`. Added an assertion there.

FeralChild64:
> Hopefully the patterns can be more flexible so it can be used for bulk filtering and discovery.. like:
> 
> `ls */setup.exe`, to find all the games with a setup tool.

This would mean quite a serious change to the implementation - and my main goal was to handle Hidden / System attributes in the command.

kcgen:
Testing with this directory tree:
```
.
├── EMPTYDIR
├── FILENAME
└── SUBDIR
    └── SUBFILE
```

### No wildcard tests

- [X] `ls` Full name (file)

    

- [x] `ls` Full name (directory)

    

- [x] `ls` Full name (directory) with slash/

    

- [x] `ls` Full name (directory/file)

    

### Wildcard tests

- [ ] `ls` prefix wildcard (file)

    Extra matches on DOS side.

    

- [x] `ls` postfix wildcard (file)

    

- [x] `ls` prefix wildcard (directory)

    

- [x] `ls` single wildcard (all)

    

- [ ] `ls` single wildcard and slash (all) 

    

- [ ] `ls` wildcard postfix and slash

    

- [ ] `ls` wildcard prefix and slash

    


The handling is very inconsistent, to the point where I probably won't know if it's working or not.

> This would mean quite a serious change to the implementation - and my main goal was to handle Hidden / System attributes in the command.

Understood, and this improvement is good.

Just thinking about the unhandled wildcard cases: would it be easy to constraints/block them?

I would be a lot more comfortable using this if it it printed a color warning: **`"Unhandled wildcard pattern"`**, instead of confidently telling me lies, like: `"No files or subdirectories"` :sweat_smile:

FeralChild64:
@kcgen

Latest changes:
- added detection of unsupported wildcards
- changed switch `/a` to `-a` (since I guess you want Unix syntax for this command)

As for `ls *NAME` behavior - this is the result of reusing DOS wildcard support code within `ls` implementation. To fix this, we would need to implement brand new Unix-style wildcard matching.

kcgen:
> * changed switch `/a` to `-a` (since I guess you want Unix syntax for this command)

The slash was good (sorry, when I tested: `ls -d` I just typed the command into the DOS side too, without thinking.. and it ignored it).

I don't think we should change the `/arg` syntax on the DOS side.

kcgen:
> Latest changes:
> 
>     * added detection of unsupported wildcards
>
> As for `ls *NAME` behavior - this is the result of reusing DOS wildcard support code within `ls` implementation. To fix this, we would need to implement brand new Unix-style wildcard matching.

The new functionality is a lot clearer:



I have no doubt about using it now; just read the message and carry on.

If you want to revert/drop the `/a` to `-a` change, I think it's OK to merge.

kcgen (approved):

FeralChild64:
I'll revert the `/a` change, just wait a few minutes.

FeralChild64:
OK, change reverted.

PR #3048, opened 2023-10-23
Tune the IBM Music Feature card's 0-dB level and default filter 

kcgen:
# Description

@Python-Exoproject found a nice set of IMFC/FB01 line-out recordings. 

Comparing them against our existing IMFC settings showed we're delivering a little under 3 dB too much default volume and that our default low-pass filter is twice as permissive as it should be (both in dB order and the pass frequency).

This PR tunes these defaults to match the recordings.

## Testing: King's Quest I SCI (1990)

**Line-out recording:**

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/8e9487b4-a86f-43ce-be24-478b88208383

Source: Pierre, YouTube https://www.youtube.com/watch?v=WHVWDi15AIw

**PR:**

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/caa55b17-d4c1-4251-b18c-af157e7da4e5

```ini
[dosbox]
machine = ega
memsize = 1

[cpu]
cycles = 3000

[imfc]
imfc = on
imfc_filter = on

[dos]
xms = false
ems = false
umb = false

[autoexec]
@mount c .
c:
copy iesource.cfg resource.cfg
sierra
exit
```

**Waveform and spectrum comparison:**

Line-out is first, PR is second. Both roll off at 3.5 KHz and taper down to `-85 dB` just under 16 KHz.

## Testing: Laura Bow 1 - The Colonel's Bequest (1989)

**Line-out recording:**

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/ea374a99-9490-4c21-9ce3-7f738c8637b7

Source: Pierre, Youtube https://www.youtube.com/watch?v=VZs3VZm61Z8

**PR:**

Similar to the original recording, a chorus value of 80 `mixer imfc c80` is applied to the IMFC output in the `[autoexec]` section. 

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/6d7d0cc5-4f82-4796-85c5-c82f782c06cc

```ini
[dosbox]
machine = ega
memsize = 1

[cpu]
cycles = 3000

[imfc]
imfc = on
imfc_filter = on

[dos]
xms = false
ems = false
umb = false

[autoexec]
@mount c .
c:
copy iesource.cfg resource.cfg
mixer imfc c80
sierra
exit
```

**Waveform and spectrum comparison:**

Line-out is first, PR is second. Both roll off at 3.5 KHz and taper down to `-85 dB` just under 9 KHz.

If you inspect the low frequency side, the IMFC recordings have more amplitude retained going down from 50 Hz and below, which makes sense because we have our 50 Hz high-pass filter in place to scrub off mostly inaudible ultra low frequencies.

# Checklist

I have:

### DIFF
--- a/src/hardware/imfc.cpp
+++ b/src/hardware/imfc.cpp
@@ -13393,15 +13393,17 @@ static void imfc_init(Section* sec)
 	constexpr auto volume_scalar = 2.1f;
 	channel->Set0dbScalar(volume_scalar);
 
-	// The filter parameters have been tweaked by analysing real hardware
-	// recordings. The results are virtually indistinguishable from the
-	// real thing by ear only.
+	// The filter parameters have been tweaked by analysing hardware
+	// line-out recordings by Pierre: Ref:
+	// https://www.youtube.com/watch?v=WHVWDi15AIw. The results are
+	// virtually indistinguishable from the real thing by ear and spectrum
+	// analysis.
 	const std::string filter_choice = conf->Get_string("imfc_filter");
 	const auto filter_choice_has_bool = parse_bool_setting(filter_choice);
 
 	if (filter_choice_has_bool && *filter_choice_has_bool == true) {
-		constexpr auto order       = 1;
-		constexpr auto cutoff_freq = 8000;
+		constexpr auto order       = 2;
+		constexpr auto cutoff_freq = 3500;
 		channel->ConfigureLowPassFilter(order, cutoff_freq);
 		channel->SetLowPassFilter(FilterState::On);
 
--- a/src/hardware/imfc.cpp
+++ b/src/hardware/imfc.cpp
@@ -13450,7 +13450,7 @@ void init_imfc_dosbox_settings(Section_prop& secprop)
 	int_prop->Set_help(
 	        "The IRQ number of the IBM Music Feature Card (3 by default).");
 
-	const auto str_prop = secprop.Add_string("imfc_filter", when_idle, "off");
+	const auto str_prop = secprop.Add_string("imfc_filter", when_idle, "on");
 	assert(str_prop);
 	str_prop->Set_help(
 	        "Filter for the IBM Music Feature Card output:\n"
--- a/src/hardware/imfc.cpp
+++ b/src/hardware/imfc.cpp
@@ -13375,9 +13375,13 @@ static void imfc_init(Section* sec)
 	m_loggerMutex = SDL_CreateMutex();
 #endif
 
+	// The emulation requires playback at 44.1 KHz to match the pitch of
+	// original hardware.
+	constexpr uint16_t imfc_sampling_rate_hz = 44100;
+
 	// Register the Audio channel
 	auto channel = MIXER_AddChannel(IMFC_Mixer_Callback,
-	                                use_mixer_rate,
+	                                imfc_sampling_rate_hz,
 	                                "IMFC",
 	                                {ChannelFeature::Stereo,
 	                                 ChannelFeature::ReverbSend,
--- a/src/hardware/imfc.cpp
+++ b/src/hardware/imfc.cpp
@@ -13384,8 +13384,9 @@ static void imfc_init(Section* sec)
 	                                 ChannelFeature::ChorusSend,
 	                                 ChannelFeature::Synthesizer});
 
-	// Bring up the volume to get it on-par with line recordings
-	constexpr auto volume_scalar = 4.0f;
+	// Default volume scalar adjusted to match hardware line-out levels
+	// recorded by Pierre: Ref: https://www.youtube.com/watch?v=WHVWDi15AIw
+	constexpr auto volume_scalar = 2.1f;
 	channel->Set0dbScalar(volume_scalar);
 
 	// The filter parameters have been tweaked by analysing real hardware


src/hardware/imfc.cpp
```diff
@@ -13376,8 +13376,10 @@ static void imfc_init(Section* sec)
 #endif
 
 	// Register the Audio channel
+	constexpr uint16_t imfc_sampling_rate_hz = 44100;
```
johnnovak, line 13379: So when this is not set to 44.1kHz, the generated pitch does not match that of real-world FB-01 / IMFC recordings?

Having to use 44.1kHz render or "emulation" resolution seems fishy. For one, I could not find any info on the internal sample rate of tge Yamaha PSGs used in the FB-01 / IMFC. I doubt it's 44.1kHz; that is very much tied to CD Audio. But this is an FM sound generator, probably it has some weird internal rate like OPL3.

If you really must set 44.1k, the only explanation I can think of is that the author of the IMFC emulation hardcoded the 44.1k requirement, probably because had a soundcard thst could only do 44.1k or something.

So yeah, I'd like to understand this a bit better.

src/hardware/imfc.cpp
```diff
@@ -13384,8 +13384,9 @@ static void imfc_init(Section* sec)
 	                                 ChannelFeature::ChorusSend,
 	                                 ChannelFeature::Synthesizer});
 
-	// Bring up the volume to get it on-par with line recordings
-	constexpr auto volume_scalar = 4.0f;
+	// Default volume scalar adjusted to match hardware line-out levels
+	// recorded by Pierre: Ref: https://www.youtube.com/watch?v=WHVWDi15AIw
+	constexpr auto volume_scalar = 2.1f;
```
johnnovak, line 13391: Can we get the emulation to generate a full-scale test signal? Like a 1kHz sine wave? It would be nice to also mention what the peak value of such a signal should be on the DBFS scale (e.g., -3.1dB).

https://en.m.wikipedia.org/wiki/DBFS

src/hardware/imfc.cpp
```diff
@@ -13448,7 +13448,7 @@ void init_imfc_dosbox_settings(Section_prop& secprop)
 	int_prop->Set_help(
 	        "The IRQ number of the IBM Music Feature Card (3 by default).");
 
-	const auto str_prop = secprop.Add_string("imfc_filter", when_idle, "off");
+	const auto str_prop = secprop.Add_string("imfc_filter", when_idle, "on");
```
johnnovak, line 13451: I agree with the commit message 😄

johnnovak:
> If you inspect the low frequency side, the IMFC recordings have more amplitude retained going down from 50 Hz and below, which makes sense because we have our 50 Hz high-pass filter in place to scrub off mostly inaudible ultra low frequencies.

No, the master HPF is set to 20Hz, from memory, and it's quite steep. The extra 50Hz low-end bump on the hardware recordings is probably caused by various colorations of the analog output stages of the card. We emulate the lowpass filters of our emulated sound cards, but AFAIK only the MT-32 emulation goes to the next level to emulate the _whole_ frequency response spectrum that includes such colorations (subtle peaks and valleys) across the spectrum.

src/hardware/imfc.cpp
kcgen: yes; the emulation itself assumes a 44.1 KHz playback sampling rate (or that we "tick over" the emulation at this pace, otherwise the pitch is affected.

There are no obvious ratios / constants / or configurable variables  that set this; so I can only assume this was an assumption i IBM's firmware itself (the emulation was 1:1 reverse engineering of that firmware at the ASM level).

johnnovak:
Very nice and welcome enhancement @kcgen. I wanted of research the filter behaviour myself because I suspected it's not accurate, but got distracted by the shader work. Will give them a good listen today.

src/hardware/imfc.cpp
kcgen: That would be nice; although we'd probably want to run that on real hardware, too?  (I don't know how to do this though.. well, without having to learn the custom MIDI-like calls and going down a rabbit hole with it).

These games do generate tones across most of the frequency range (including percussion), and the 2.1 scalar was a very close approximation versus the hardware; and I checked quite a few. 

```
cap-1.90vol-441khz-nofilter-nomixer.wav
cap-2.00vol-48khz-nofilter-nomixer.wav
cap-2.10vol-44100-hz-filter-lpf-2-3500Hz-mixer-effects-off.wav
cap-2.13vol-48khz-nofilter-nomixer.wav
cap-2.15vol-441khz-nofilter-nomixer.wav
cap-2.17vol-48khz-nofilter-nomixer.wav
cap-2.20vol-48khz-filter-on-nomixer.wav
cap-2.25vol-48khz-filter-lpf-2-3500Hz-mixer-effects-off.wav
cap-2.27vol-441khz-nofilter-nomixer.wav
cap-2.30vol-48khz-nofilter-nomixer.wav
cap-2.42vol-48khz-filter-on-nomixer.wav
```

kcgen:
> No, the master HPF is set to 20Hz, from memory, and it's quite steep @kcgen.

:+1: ; ahh yes: `constexpr auto highpass_cutoff_freq = 20.0;`

The hardware recordings show IMFC's filter tracks logarithmically down (appears as a very linear down-trend in the spectrum log charts), where as our 2nd order filter appear to be just slightly less logarithmic with a slight visual convex curve to the trend line.

But not a problem, and I certainly am not able to tell the difference between these side-by-side recordings.

src/hardware/imfc.cpp
johnnovak: Makes sense, maybe a comment would help there to clarify this (saying that the emulation _must_ operate at 44.1kHz).

kcgen: Good call; otherwise it's asking for someone to make it run at whatever the mixer-rate is :-)

kcgen: Added.

src/hardware/imfc.cpp
johnnovak: > That would be nice; although we'd probably want to run that on real hardware, too? (I don't know how to do this though.. well, without having to learn the custom MIDI-like calls and going down a rabbit hole with it).

That's right, and even measuring the peak DBFS value of a test signal on the emulation side would require figuring out how to generate that signal in the first place. Maybe just leave it as is, then 😄

johnnovak (approved):
Very nice @kcgen , these sound very close to the real hardware recordings; the nice warm character of the original IMFC/FB-01 is retained 🎉

src/hardware/imfc.cpp
johnnovak: > Good call; otherwise it's asking for someone to make it run at whatever the mixer-rate is :-)

Yeah, I'd do that next time I touch this file if the comment was not there 😅

PR #3058, opened 2023-10-26
Add a fade-to-sleep-after-last-port-write feature to the mixer

kcgen:
# Description

Adds a linear fade-to-sleep after the last port-write feature to the mixer's existing sleeper class:

1. This feature is **disabled by default**. It needs to be opted-in via device-specific conf setting (`tandy_fadeout = ...` or `opl_fadeout = ...`). 

1. This is a quality of life feature for users who want protection against "phone off the hook" behavior that some games exhibit _when using tone-generating audio devices_.

1. This is not an emulation improvement or bug fix. This feature is desirable regardless if the game is buggy or coded correctly. Who knows.. maybe the developers intended to blast the user with a 10-second fixed-waveform  during the intermission!.. OK: but I don't want to hear it :-)

1. It will only be made available for **fixed-waveform generating devices** because these devices will continue to play the last-requested waveform indefinitely until requested otherwise.

1. The `WAIT`-before-fade period and the `FADE` down period are both configurable. Conservative defaults will be used, so users who turn it on don't have to worry about "cutting the music short", so to speak.

    For example, here are two inn visit sequences in Bard's Tale (Staging and 86Box):

    https://github.com/dosbox-staging/dosbox-staging/assets/1557255/5e1796b3-c7d9-4c57-81f3-6a1a9bb362b8

    https://github.com/dosbox-staging/dosbox-staging/assets/1557255/c46e65c1-1416-4367-95ba-177b140d15b2

## Related issues

None.

# Manual testing

Tested many OPL and Tandy games to checks for regressions, both with it enabled and disabled.

Tested Bard's Tale Construction Set and California games II. Confirmed their output is correctly faded down to silence after the configured WAIT and FADE durations, after their last IO port write.

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/741cd2ba-ae14-4c9a-80ad-c51ab5310827

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/c5ca9425-b5ad-47d5-a016-4b371cdfb6bc

Tested values ranges:
- `WAIT` values: too high, too low, garbage text, and missing
- `FADE` values: too high, too low, garbage text, and missing

- `WAIT` min, while `FADE` min
- `WAIT` min, while `FADE` mid
- `WAIT` min, while `FADE` max

- `WAIT` mid, while `FADE` min
- `WAIT` mid, while `FADE` mid
- `WAIT` mid, while `FADE` max

- `WAIT` max, while `FADE` min
- `WAIT` max, while `FADE` mid
- `WAIT` max, while `FADE` max

# Checklist

I have:

### DIFF
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -116,6 +116,7 @@ static constexpr StereoLine Reverse = {Right, Left};
 enum class ChannelFeature {
 	ChorusSend,
 	DigitalAudio,
+	FadeOut,
 	ReverbSend,
 	Sleep,
 	Stereo,
@@ -209,6 +210,8 @@ public:
 	void ConfigureHighPassFilter(const uint8_t order, const uint16_t cutoff_freq);
 	void ConfigureLowPassFilter(const uint8_t order, const uint16_t cutoff_freq);
 	bool TryParseAndSetCustomFilter(const std::string_view filter_prefs);
+
+	bool ConfigureFadeOut(const std::string_view fadeout_prefs);
 
 	void SetResampleMethod(const ResampleMethod method);
 	void SetZeroOrderHoldUpsamplerTargetFreq(const uint16_t target_freq);
@@ -402,15 +405,30 @@ private:
 	class Sleeper {
 	public:
 		Sleeper() = delete;
-		Sleeper(MixerChannel& c);
-		void Listen(const AudioFrame frame);
+		Sleeper(MixerChannel& c,
+		        const uint16_t sleep_after_ms = default_wait_ms);
+		bool ConfigureFadeOut(const std::string_view prefs);
+		AudioFrame MaybeFadeOrListen(const AudioFrame& frame);
 		void MaybeSleep();
 		bool WakeUp();
 
 	private:
+		void DecrementFadeLevel(const int64_t awake_for_ms);
+
 		MixerChannel& channel;
-		int64_t woken_at_ms = 0;
-		bool had_noise      = false;
+
+		// The wait before fading or sleeping is bound between:
+		static constexpr auto min_wait_ms     = 100;
+		static constexpr auto default_wait_ms = 500;
+		static constexpr auto max_wait_ms     = 5000;
+
+		int64_t woken_at_ms                = {};
+		float fadeout_level                = {};
+		float fadeout_decrement_per_ms     = {};
+		uint16_t fadeout_or_sleep_after_ms = {};
+
+		bool wants_fadeout = false;
+		bool had_signal    = false;
 	};
 	Sleeper sleeper;
 	const bool do_sleep = false;
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -796,6 +796,19 @@ void DOSBOX_Init()
 	                  "On 'auto' the mode is determined by 'sbtype'.\n"
 	                  "All OPL modes are AdLib-compatible, except for 'cms'.");
 
+	pstring = secprop->Add_string("opl_fadeout", when_idle, "off");
+	pstring->Set_help(
+	        "Fade out the OPL synth output after the last IO port write:\n"
+	        "  off:       Don't fade out; residual output will play forever (default).\n"
+	        "  on:        Wait 0.5s before fading out over a 0.5s period.\n"
+	        "  <custom>:  A custom fade-out definition in the following format:\n"
+	        "               WAIT FADE\n"
+	        "             Where WAIT is how long after the last IO port write fading begins,\n"
+	        "             ranging between 100 and 5000 milliseconds; and FADE is the fade-out\n"
+	        "             period, ranging between 10 and 3000 milliseconds. Examples:\n"
+	        "                300 200 (Wait 300ms before fading out over a 200ms period)\n"
+	        "                1000 3000 (Wait 1s before fading out over a 3s period)");
+
 	pstring = secprop->Add_string("oplemu", deprecated, "");
 	pstring->Set_help("Only 'nuked' OPL emulation is supported now.");
 
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -905,6 +905,13 @@ void DOSBOX_Init()
 	        "         without DAC to avoid conflicts with other cards using DMA 1.\n"
 	        "  off:   Disable Tandy/PCjr sound.");
 
+	pstring = secprop->Add_string("tandy_fadeout", when_idle, "off");
+	pstring->Set_help(
+	        "Fade out the Tandy synth output after the last IO port write:\n"
+	        "  off:       Don't fade out; residual output will play forever (default).\n"
+	        "  on:        Wait 0.5s before fading out over a 0.5s period.\n"
+	        "  <custom>:  Custom fade out definition; see 'opl_fadeout' for details.");
+
 	pstring = secprop->Add_string("tandy_filter", when_idle, "on");
 	pstring->Set_help(
 	        "Filter for the Tandy synth output:\n"
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -1055,6 +1055,7 @@ void MixerChannel::Mix(const uint16_t frames_requested)
 
 		handler(static_cast<work_index_t>(frames_remaining));
 	}
+
 	if (do_sleep) {
 		sleeper.MaybeSleep();
 	}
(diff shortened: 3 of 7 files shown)


johnnovak:
My general high-level thoughts about this @kcgen:

1. Are we 100% certain the "stuck notes" are not a byproduct of inaccurate emulation but also happen on real hardware? I am planning to verify the OPL3 stuck note cases on real OPL3 (on an AWE32 card) but that has to wait a few weeks. My point: we should address the root cause (inaccurate emulation) rather than apply a bandaid solution (fade outs), if that's the case. Not necessarily saying that's the case, just want to verify/confirm that because at this point I personally don't know for certain 😄 

2. If we have determined that the stuck notes _do_ happen on real hardware too, optionally fading the stuck notes to silence makes sense. However, I'm not completely sure about the current implementation; does the fade timeout start if there are no more port writes for a set number of milliseconds? If that's the case, that's a rather risky approach. What if the music contains long held notes, potentially lasting for seconds? In that case, using a proper [gate audio processor](https://en.wikipedia.org/wiki/Noise_gate) that operates only on the audio signal would be much preferable. In particular, the hysteresis is super important to avoid rapid "re-triggers" around the gate threshold. There are several high-quality free and open-source gate implementations available, e.g. [this JSFX gate plugin](https://github.com/chkhld/jsfx/blob/master/plugins/gate_expander.jsfx) is a serious contender. I very much recommend reading the excellent top comment of this linked effect source code as it explains the workings of a proper audio gate very well.

3. If we have determined that we need the auto-silencer and we have a proper audio gate implementation, I would consider _not_ introducing a separate `FadeOut`/`Gate` feature but merging it with the `Sleep` feature instead. My reasoning is that the "sleep" and "silence the signal once below a set threshold" ("gate") feature are very closely tied together, and we're doing an extremely simplistic "gate" implementation for the "sleep" feature already. 

I guess that's probably enough to consider for now 😄 The biggest question for me is the first point, then the second and third points are refinements.

Grounded0:
> Are we 100% certain the "stuck notes" are not a byproduct of inaccurate emulation but also happen on real hardware? I am planning to verify the OPL3 stuck note cases on real OPL3 (on an AWE32 card) but that has to wait a few weeks. My point: we should address the root cause (inaccurate emulation) rather than apply a bandaid solution (fade outs), if that's the case. Not necessarily saying that's the case, just want to verify/confirm that because at this point I personally don't know for certain 😄

💯

kcgen:
@johnnovak , thanks for all the feedback and great points.

I've updated the original PR wording to hopefully address all the points. As we'll probably reference this in the release minutes, I figured updating the original wording would cover both needs :-)

I haven't touched the code yet, but will adapt the feature into the existing sleper class, like you've suggested.

johnnovak:
> @johnnovak , thanks for all the feedback and great points.
> 
> I've updated the original PR wording to hopefully address all the points. As we'll probably reference this in the release minutes, I figured updating the original wording would cover both needs :-)
> 
> I haven't touched the code yet, but will adapt the feature into the existing sleper class, like you've suggested.

Thanks, that updated description puts the PR into a different light 😄 Yeah, I think merging this optional fade feature into the auto-sleep makes sense.

Grounded0:
If you quit Legend's Gateway and Gateway II with OPL music both games leave a hanging note in the background.

Grounded0:
If this OPL3 hanging note on quit bug is pervasive hardware bug then one option would be to allow what was normal on my Sound Blaster 32 and included as a command: `AWEUTIL /S´.

This would "zap" or reinitialise the EMU8000 and YMF262 on the card and would of course eliminate any hanging notes too. This could be implemented as a `CONFIG` option that could be added to the last line of `.BAT` files of prolematic games to "zap" the emulated OPL3 back to normal state.

johnnovak:
Good idea @Grounded0, we'll see how the hardware investigation pans out.

rderooy:
> If this OPL3 hanging note on quit is pervasive hardware or games behaving badly issue then one option would be to implement functionality that was on my Sound Blaster 32 and included as a command: `AWEUTIL /S`.
> 
> This would "zap" or reinitialise the EMU8000 (MIDI) and YMF262 (FM synthesis) on the card and would of course eliminate any hanging notes too. This could be implemented as a `CONFIG` option that could be added to the last line of `.BAT` files for problematic games to "zap" the emulated OPL3 back to normal state.

I think something like `mixer /reset` may be better in that case to silence any hanging notes.

kcgen:
Yank and re-inserting the card:

```shell
sbtype none
sbtype sb16
```

(It's a full state reset in the DOSBox SB and OPL modules)

johnnovak:
> Yank and re-insert the card:
> 
> ```shell
> sbtype none
> sbtype sb16
> ```
> 
> (It's a full state reset in the DOSBox SB and OPL modules)

The more I think about it, I think we don't need special facilities for this but should just use @kcgen's suggestion. I will mention this in the manual.

kcgen:
Merged the fadeout feature with the sleeper class.

Overhead is minimal: there is only a single `bool` check per mixer callback.
- If the fadeout feature is disabled (which it is, by default), there there is nothing more.
- If the feature is enabled _but_ the `WAIT` period hasn't happened yet, then no additional work is done.
- IF the feature is enabled _and_ the `WAIT` period has been crossed, then:
    - Two floating-point multiplications (one per left and right audio samples) are performed on each audio frame
    - One floating point subtraction (of a pre-computed decrement amount) is performed per millisecond callback event.

kcgen:
Added a screenshot showing the fade out in action to the PR's wording.

kcgen:
Suggest reviewing all at once (in final-form), because I've kept both the separate implementation and the merged implementation on their own commits, just to ease separating them if needed in the future.

include/mixer.h
```diff
@@ -210,6 +211,9 @@ class MixerChannel {
 	void ConfigureLowPassFilter(const uint8_t order, const uint16_t cutoff_freq);
 	bool TryParseAndSetCustomFilter(const std::string_view filter_prefs);
 
+	void ConfigureFadeout(const std::string_view fadeout_prefs);
```
johnnovak, line 214: FadeOut (uppercase O)

src/dosbox.cpp
```diff
@@ -796,6 +796,20 @@ void DOSBOX_Init()
 	                  "On 'auto' the mode is determined by 'sbtype'.\n"
 	                  "All OPL modes are AdLib-compatible, except for 'cms'.");
 
+	pstring = secprop->Add_string("opl_fadeout", when_idle, "off");
+	pstring->Set_help(
+	        "Fadeout the OPL synth output after the last IO port write:\n"
```
johnnovak, line 801: "Fade out the..."

Should be "fade out" (two words) everywhere.

src/dosbox.cpp
```diff
@@ -796,6 +796,20 @@ void DOSBOX_Init()
...
+	        "  off:       Don't fadeout; residual output will play forever (default).\n"
+	        "  on:        Wait 350 ms before fading out over a 350 ms period.\n"
```
johnnovak, line 803: Dunno, isn't the wait a bit short? Maybe 500 would be safer?

src/dosbox.cpp
```diff
@@ -796,6 +796,20 @@ void DOSBOX_Init()
...
+	        "  <custom>:  A custom fadeout definition in the following format:\n"
+	        "               WAIT FADE\n"
+	        "             Where WAIT is how long (in milliseconds) before fading begins,\n"
+	        "             and FADE is the duration (in milliseconds) of the fade out period.\n"
+	        "             Both values can range from 100 to 10000 (milliseconds).\n"
```
johnnovak, line 808: 10s seems excessive, especially for fade outs.

I think I'd use 10 ms to 3s for fades (even 3s is very long, but let's give them a generous range, while also allowing almost instant 10ms fades).

Then for wait I don't think anyone would want to listen for a static tone for 10 seconds before fading it out. Maybe 100ms to 5s?

src/dosbox.cpp
```diff
@@ -796,6 +796,20 @@ void DOSBOX_Init()
...
+	        "             Examples:\n"
+	        "                 1000 3000  (Wait 1 s before fading-out over a 3 s period)\n"
```
johnnovak, line 810: 1s, 3s, 10ms etc. would look better IMO (no space).

src/hardware/mixer.cpp
```diff
@@ -1693,45 +1694,130 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
-void MixerChannel::Sleeper::Listen(const AudioFrame frame)
+void MixerChannel::Sleeper::ConfigureFadeout(const std::string_view prefs)
 {
-	if (had_noise) {
+	auto set_wait_and_fade = [&](const int wait_ms, const int fade_ms) {
+		fadeout_or_sleep_after_ms = check_cast<uint16_t>(wait_ms);
+
+		fadeout_decrement_per_ms = 1.0f / static_cast<uint16_t>(fade_ms);
+
+		LOG_MSG("%s: Fadeout enabled (wait %d ms then fade for %d ms)",
```
johnnovak, line 1709: "Fade out"

src/hardware/mixer.cpp
```diff
@@ -1693,45 +1694,130 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
-// Records if samples had a magnitude great than 1. This is a one-way street;
-// once "had_noise" is tripped, it can only be reset after waking up.
-void MixerChannel::Sleeper::Listen(const AudioFrame frame)
+void MixerChannel::Sleeper::ConfigureFadeout(const std::string_view prefs)
 {
-	if (had_noise) {
+	auto set_wait_and_fade = [&](const int wait_ms, const int fade_ms) {
+		fadeout_or_sleep_after_ms = check_cast<uint16_t>(wait_ms);
+
+		fadeout_decrement_per_ms = 1.0f / static_cast<uint16_t>(fade_ms);
```
johnnovak, line 1707: Successively decrementing can leave you with a close to zero but not quite zero number. I know there were problems with such denormal floats on earlier CPUs, so make sure to set the level explicitly to 0.0f once the fade is over. The sometime an extra MUL might be more performant than a SUB and an extra IF, but in this case we're doing so little work it doesn't matter either way...

Btw, I don't think a float SUB is really faster than a MUL in actual algos when pipelining is involved, etc. A MUL is 1 cycle usually.

src/hardware/mixer.cpp
```diff
@@ -1693,45 +1694,130 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
+	// The constructed sleep period is programmatically controlled (so assert)
+	assert(fadeout_or_sleep_after_ms >= min_duration_ms);
+	assert(fadeout_or_sleep_after_ms <= max_duration_ms);
+}
 
-	// Not enough time has passed.. try again later
-	if (GetTicksSince(woken_at_ms) < consider_sleeping_after_ms) {
-		return;
+// Either fades the frame by the current fade-level, or records if the frame had
+// a magnitude great than 1. Cnce "had_noise" is tripped, it can only be reset
```
johnnovak, line 1777: "greater than"
"Once"

src/hardware/mixer.cpp
```diff
@@ -1693,45 +1694,130 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
+	if (awake_for_ms < fadeout_or_sleep_after_ms) {
+		return;
+	}
+	if (wants_fadeout) {
+		// The channel is still fading.. try to sleep later
+		if (fadeout_level > 0.0f) {
+			DecrementFadeLevel(awake_for_ms);
+			return;
+		}
+	} else if (had_noise) {
```
johnnovak, line 1806: Existing code but `had_noise` and the use of the term "noise" gave me a pause. I know this is the "sleeper" class but still, we're in the context of OPL and Tandy here which feature noise waveforms and dedicated noise channels, so I find the use of the word "noise" instead of "signal" confusing.

I recommend using a less confusing term such as "level" or "signal" or something.

src/dosbox.cpp
kcgen: Sure; still reasonable and not too onerous if you need to hear the fixed tone.

src/dosbox.cpp
kcgen: Ok; but the International System of Units (SI) always has a space between values and their units - https://physics.nist.gov/cuu/Units/checklist.html

src/dosbox.cpp
kcgen: If the last values isn't a tone by stuck in a DC offset, a 10 ms ramp down produces a 100 Hz triangle waveform, which is higher pitch than our 20 Hz high-pass filter. (I've heard these digital ramps before; they kind of sound like a soft plastic zippering sound!)

So that's where the 100 ms lower bound came from; but if the user really wants a near-instant cut-off, then lets let them have it :-)

src/hardware/mixer.cpp
kcgen: > Successively decrementing can leave you with a close to zero but not quite zero number. 

I think you mean successively multiplying (by a percent)?

This is a decrement that's subtracted each millisecond, and just keeps decrementing until it dips below 0 and is clamped to zero.

So for example, a 10ms fade out will produce a decrement of 0.1/ms.

The multiplier values each millisecond pass become:

1.0 - 0.1 * (1ms into fade) = 0.9
1.0 - 0.1 * (2ms into fade) = 0.8
...
1.0 - 0.1 * (9ms into fade) = 0.1
1.0 - 0.1 * (10ms into fade) = 0.0
...
next step goes negative, and is clamped to zero.

src/hardware/mixer.cpp
kcgen: good eye :-)

src/hardware/mixer.cpp
kcgen: `had_signal` looks better; switched!

src/dosbox.cpp
johnnovak: Interesting, from memory I've been using short 10-20ms fades all the time in DAW work, but I might misremember 😄 I'll pay attention to the zipper noises next time 😄

src/dosbox.cpp
johnnovak: Yeah, OK, you're right. Just "over a 3 s" seemed funky, but that's what you get with monospaced fonts 😅 

We use "70 Hz" in other places, but "256K" and "8MB" in other places, so it's a bit all over the place. Sticking "K" and "MB" to the numbers is kind of a "convention" I guess 😄

src/hardware/mixer.cpp
johnnovak: Yeah, that's fine, maybe I was a bit overthinking it 😄 Glad to see the clamp to zero there too 😄

johnnovak:
I'm wondering if these checks could be converted to some simple unit tests @kcgen? Ideally you'd only need to test a single parser function that returns the parsed values or an error, similarly to how I approached error handling and unit testing in my MIXER command rework.

```
Tested values ranges:

    WAIT values: too high, too low, garbage text, and missing

    FADE values: too high, too low, garbage text, and missing

    WAIT min, while FADE min

    WAIT min, while FADE mid

    WAIT min, while FADE max

    WAIT mid, while FADE min

    WAIT mid, while FADE mid

    WAIT mid, while FADE max

    WAIT max, while FADE min

    WAIT max, while FADE mid

    WAIT max, while FADE max
```

src/dosbox.cpp
```diff
@@ -796,6 +796,18 @@ void DOSBOX_Init()
...
 	                  "All OPL modes are AdLib-compatible, except for 'cms'.");
 
+	pstring = secprop->Add_string("opl_fadeout", when_idle, "off");
+	pstring->Set_help(
+	        "Fade out the OPL synth output after the last IO port write:\n"
+	        "  off:       Don't fade out; residual output will play forever (default).\n"
+	        "  on:        Wait 0.5s before fading out over a 0.5s period.\n"
+	        "  <custom>:  A custom fade out definition in the following format:\n"
+	        "               WAIT FADE\n"
+	        "             Where WAIT is how long before fading begins from 100 to 5000 in ms,\n"
```
johnnovak, line 806: I'd spell out "milliseconds" instead of "ms".

src/dosbox.cpp
```diff
@@ -796,6 +796,18 @@ void DOSBOX_Init()
...
+	        "             and FADE is the fade out period from 10 and 3000 in ms. Examples:\n"
+	        "                 300 200 (Wait 300ms before fading out over a 200ms period)\n"
+	        "                 1000 3000 (Wait 1s before fading out over a 3s period)");
```
johnnovak, line 809: I'm not sure this is a good idea, just wondering if having to write `wait 300 fade 200` would be better or not. Whenever I have to specify a bunch of values I always forget what is what, so this would make it self-documenting at the setting level. But maybe it would unnecessary complication.

src/hardware/mixer.cpp
```diff
@@ -1693,46 +1694,140 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
+
+			if (wait_is_valid && fade_is_valid) {
+				set_wait_and_fade(*wait_ms, *fade_ms);
+				wants_fadeout = true;
+				return;
+			}
+		}
+	}
+	// Otherwise inform the user and disable the fade
+	LOG_WARNING("%s: Invalid custom fadeout definition: '%s'. Must be "
```
johnnovak, line 1749: Ok, checked out the Oxford dictionary, and I think this is the correct usage:

- "fade-out" — that's the noun, so for example "Invalid custom fade-out definition:"
- "fade out" — that's the verb form, e.g., "To fade out stuck notes, use the new `opl_fadeout` setting"
- "fadeout" — not a work 😎 

Could you just grep "fadeout" and "fade out" and update the comments and especially user-facing messages and descriptions accordingly?

src/hardware/mixer.cpp
```diff
@@ -1693,46 +1694,140 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
+	            "specified in \"WAIT FADE\" format where WAIT is between "
+	            "%d and %d (in milliseconds) and FADE is between %d and "
+	            "%d (in milliseconds). Using 'off'",
```
johnnovak, line 1752: Excellent warning message otherwise 🎖️

src/hardware/mixer.cpp
```diff
@@ -1693,46 +1694,140 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
+MixerChannel::Sleeper::Sleeper(MixerChannel& c, const uint16_t sleep_after_ms)
+        : channel(c),
+          fadeout_or_sleep_after_ms(sleep_after_ms)
+{
+	// The constructed sleep period is programmatically controlled (so assert)
+	assert(fadeout_or_sleep_after_ms >= min_wait_ms);
+	assert(fadeout_or_sleep_after_ms <= max_wait_ms);
+}
+
+// Either fades the frame by the current fade-level, or records if the frame had
```
johnnovak, line 1785: Records what? 😄

src/hardware/mixer.cpp
```diff
@@ -1693,46 +1694,140 @@ AudioFrame MixerChannel::ApplyCrossfeed(const AudioFrame frame) const
...
+// a magnitude greater than 1. Once "had_signal" is tripped, it can only be
+// reset after waking up.
+void MixerChannel::Sleeper::MaybeFadeOrListen(AudioFrame& frame)
```
johnnovak, line 1788: You don't need to change this specific instance if you don't wanna, but instead of `foo(Type& a)` we can do `a foo(const Type& a)` and return the new value instead of mutating it. Because of the mandatory optimisations of new C++ compilers that has the same performance and the calculation would be performed in place (I'm pretty sure of that), but it's a nicer signature and then it's a pure function.

src/dosbox.cpp
kcgen: We're not prefixing multi-values with their type.
For example:
- `sb_filter = lpf 2 12000` isn't: `lfp order 2 freq 12000`
- `sensitivity = 50 100` isn't `x 50 y 100`
- (and a couple others)

So I think using that approach here would be inconsistent with the other areas.

src/hardware/mixer.cpp
kcgen: ok, got 'em :-)

src/hardware/mixer.cpp
kcgen: Ahh yes; clarified the comment now :-)

src/hardware/mixer.cpp
kcgen: Switch to a return value.
In the case where we're just sleeping, the frame is simply passed through.

kcgen:
> I'm wondering if these checks could be converted to some simple unit tests 

Nice suggestion @johnnovak - they were perfect for unit tests.

src/dosbox.cpp
johnnovak: Yeah, that's cool.

src/dosbox.cpp
```diff
@@ -796,6 +796,19 @@ void DOSBOX_Init()
...
+	pstring = secprop->Add_string("opl_fadeout", when_idle, "off");
+	pstring->Set_help(
+	        "Fade out the OPL synth output after the last IO port write:\n"
+	        "  off:       Don't fade out; residual output will play forever (default).\n"
+	        "  on:        Wait 0.5s before fading out over a 0.5s period.\n"
+	        "  <custom>:  A custom fade out definition in the following format:\n"
+	        "               WAIT FADE\n"
+	        "             Where WAIT is how long before fading begins from 100 to 5000 in\n"
+	        "             milliseconds, and FADE is the fade out period from 10 and 3000 in\n"
+			"             milliseconds. Examples:\n"
```
johnnovak, line 808: Formatting

johnnovak (approved):
Nice job @kcgen 🎉

kcgen:
Thanks for the detailed review, @johnnovak!

src/dosbox.cpp
kcgen: (I squished 'em anyway because it's more consistent w/ the conf :grinning: )

PR #3052, opened 2023-10-24
Let the command line `--help` text be translatable

kcgen:
# Description

Adds the main executable CLI help message to the messaging system.

Excellent suggestion, @Kappa971.

## Related issues

Fixes #3051 

# Manual testing

Tested with CLI `--help`.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [x] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [x] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

Issue #3051, opened 2023-10-24
The DOSBox Staging help screen should be translatable


kcgen:
Originally posted by @Kappa971 in https://github.com/dosbox-staging/dosbox-staging/issues/884#issuecomment-1777376888

PR #3064, opened 2023-10-27
Handle host code pages when running `dosbox.exe --help` on Windows

FeralChild64:
# Description

This is my follow-up of the PR https://github.com/dosbox-staging/dosbox-staging/pull/3052

1. On Windows the 'dosbox.exe --help' command will ask the host OS about the console code page and (if it's not UTF-8) will try to render the help text as nicely as possible using internal Unicode engine.
**[edit]** Changed implementation to temporarily switch the console to UTF-8.

3. Default (fallback) code page is no longer 437, instead we are using a 7-bit ASCII. Code page 437 was a sane choice when not many code pages were supported; with current selection it is better to assume that unknown codepage is something completely exotic.

4. I have no idea who fixed one PVS Studio warning, but I have decremented the allowance.

Some screenshots - default console settings, before my changes:

Screenshot-Before-Changes

Same settings, after my changes:

Screenshot-After-Changes

And here is how the text is rendered when console is UTF-8:

Screenshot-UTF-8

## Related issues

Improves https://github.com/dosbox-staging/dosbox-staging/issues/3051

# Manual testing

Test file: [fc.lng.gz](https://github.com/dosbox-staging/dosbox-staging/files/13192332/fc.lng.gz)

0. Use the Microsoft Windows build

1. Set DOSBox to use dummy language provided above (decompress the file, put it into resources/translations) by setting

```ini
[dosbox]
language = fc
```

3. From the Windows console (`cmd.exe`) execute `dosbox --help` and enjoy the selection of multi-language song lyrics :D 

4. Repeat previous with different code pages set in the console - you can switch between them using `chcp <codepage>` command, where `<codepage>` is one from https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers

Notes:
- not all the code pages are supported - unsupported ones will cause DOSBox to drop all the national accents
- https://github.com/dosbox-staging/dosbox-staging/blob/main/contrib/resources/mapping/MAIN.TXT - contains a list of code pages currently supported by our Unicode engine (not all of them are supported by Windows 11)
- UTF-8 console (`chcp 65001`) is supported too

# Checklist

I have:

### DIFF
--- a/contrib/resources/mapping/MAIN.TXT
+++ b/contrib/resources/mapping/MAIN.TXT
@@ -1,4 +1,10 @@
 #
+# Note: when adding new code pages, avoid conflicts with currently supported
+# Microsoft code pages, the list can be found here:
+# - https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
+# Otherwise please adapt 'utf8_to_host_console' function accordingly.
+#
+#
 # Fully supported code pages (92 code pages):
 #
 # 113   - Yugoslavian, Latin
--- a/contrib/resources/mapping/MAIN.TXT
+++ b/contrib/resources/mapping/MAIN.TXT
@@ -2,7 +2,6 @@
 # Note: when adding new code pages, avoid conflicts with currently supported
 # Microsoft code pages, the list can be found here:
 # - https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
-# Otherwise please adapt 'utf8_to_host_console' function accordingly.
 #
 #
 # Fully supported code pages (92 code pages):
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 162
+          MAX_BUGS: 161
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -323,10 +323,6 @@ void dos_to_utf8(const std::string& in_str, std::string& out_str);
 void dos_to_utf8(const std::string& in_str, std::string& out_str,
                  const uint16_t code_page);
 
-// As above, but converts to encoding used by (or at least compatible to) host
-// OS command line console.
-bool utf8_to_host_console(const std::string& in_str, std::string& out_str);
-
 // Convert DOS code page string to lower/upper case; converters are aware of all
 // the national characters. Functions without 'code_page' parameter use current
 // DOS code page.
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -323,6 +323,10 @@ void dos_to_utf8(const std::string& in_str, std::string& out_str);
 void dos_to_utf8(const std::string& in_str, std::string& out_str,
                  const uint16_t code_page);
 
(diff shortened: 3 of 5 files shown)


contrib/resources/mapping/MAIN.TXT
```diff
@@ -1,4 +1,10 @@
 #
+# Note: when adding new code pages, avoid conflicts with currently supported
+# Microsoft code pages, the list can be found here:
+# - https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers
+# Otherwise please adapt 'utf8_to_host_console' function accordingly.
+#
```
kcgen, line 6: Very helpful comment!

kcgen (approved):
Test cases look good, and no comments on the code from me. Nice quick fix, @FeralChild64 !

shermp:
Quick question, would using `SetConsoleOutputCP` to change to 65001 at dosbox start (and revert to previous codepage at exit) be a good strategy instead? That way anything written to the console could be UTF-8.

Kappa971:
It works in both Windows Powershell and Command Prompt.

CMD uses CP850 (default) and special characters are now displayed correctly:
immagine

I still haven't figured out what code page Powershell uses but it works so it doesn't matter 😅

Grounded0:
I don't know enough about Windows but typing `CHCP` in PowerShell window should display what codepage is used. Anymore more knowledgeable about Windows correct if this is wrong.

Kappa971:
> I don't know enough about Windows but typing `CHCP` in PowerShell window should display what codepage is used. Anymore more knowledgeable about Windows correct if this is wrong.
> 
> Otherwise looking good.

Oh right, for some reason I thought it didn't work on Powershell 😅 
It also use CP850

FeralChild64:
@ shermp Dunno... I have no idea what consequences it might have to change the code page (like UAC warning on some configurations or some Windows versions), I'm not even sure how to find all the exit points from DOSBox.

FeralChild64:
@ Kappa971 I wouldn't be surprised if changing the system language (or installing the different language version) could change the console code page too.

shermp:
@FeralChild64 here's the documentation: https://learn.microsoft.com/en-us/windows/console/setconsoleoutputcp

I believe it is the equivalent of calling `chcp`. It shouldn't require any admin rights.

FeralChild64:
@shermp I'll try to test it today with the `--help` output. Changing it for the whole DOSBox would require finding all the exit points to do it correctly; I'm not sure if I can do this reliably.

Kappa971:
> @ Kappa971 I wouldn't be surprised if changing the system language (or installing the different language version) could change the console code page too.

Probably yes. CP850 should be the default of other European countries as well. Perhaps in Eastern European countries the code table is different. 
You're Polish, right? You could check it out 🙂

> @shermp I'll try to test it today with the --help output. Changing it for the whole DOSBox would require finding all the exit points to do it correctly; I'm not sure if I can do this reliably.
>
>For now switched the PR back to draft.

Why change something that works?

FeralChild64:
@Kappa971

> You're Polish, right? You could check it out 🙂

I have English version of Windows installed.

> Why change something that works?

Because maybe it will work better :)

FeralChild64:
I have changed the implementation to switch Windows console to UTF-8 for the duration of `dosbox.exe --help` output display.

The old implementation is still in the commit history, plus I want to keep some minor Unicode engine changes done there.

kcgen (approved):

PR #3066, opened 2023-10-27
Snip truism note from MIXER long help text

kcgen:
# Description

See title. This is a follow up fix that slipped through the prior PR. 

# Manual testing

Confirmed the mixer's long help does not include the snipped text. 

# Checklist

I have:

### DIFF
--- a/src/dos/program_mixer.cpp
+++ b/src/dos/program_mixer.cpp
@@ -817,8 +817,6 @@ void MIXER::AddMessages()
 	        "  - You may change the settings of more than one channel in a single command.\n"
 	        "  - If no channel is specified, you can set crossfeed, reverb, or chorus\n"
 	        "    of all channels globally.\n"
-	        "  - The reverb and chorus commands also enable the default reverb and chorus\n"
-	        "    presets, respectively, if those effects are not yet enabled.\n"
 	        "  - The /noshow option applies the changes without showing the mixer settings.\n"
 	        "\n"
 	        "Examples:\n"


Kappa971 (approved):
I won't miss you 🤣

Grounded0 (approved):

johnnovak:
I'm not entirely sure removing this was a good idea @kcgen, but I won't fight for it 😄 I will describe it in the manual in detail for sure.

kcgen:
yeah - I think it was just a bit over-documented, to the point where the user might double-guess their own intuitive understanding. 

If the user is running the `MIXER` command, then they're right there to see it action -- and can see (and hear!) that it behaves perfectly normal and as one would expect. 

But yeah.. totally agree that more space (and room for graphics and such) will be great to have in the manual.

johnnovak:
> f the user is running the `MIXER` command, then they're right there to see it action -- and can see (and hear!) that it behaves perfectly normal and as one would expect.

See my explanation above; not quite, but ok. The manual will clarify it. To be honest, I was 50/50 whether to include it or not in the help text. In the manual it must be and will be documented in detail, of course.

kcgen:
Yeah.. without:



Asking for `reverb level 30 on OPL` is honored, but you get default reverb enabled as well (because the reverb engine started up):

Grounded0:
Waiting the day i can finally replace this:

Screenshot 2023-10-28 at 4 12 54

johnnovak:
> Yeah.. without:
> 
> 
> 
> When you activate an effect using the mixer for the first time, default settings are applied to the other channels.
> 
> 
> 
> I think it's intuitive.. the absence of specific values for the other channels means they get defaults.
> 
> If you're ordering sushi, then you're getting wasabi and pickled ginger by default :-)
> 
> If a user wants non-defaults, they need to ask for them, too:
> 
> `mixer opl r30 pcspeaker r0 sb r0`

You're conflating reverb _presets_ with reverb send _levels_ @kcgen 😄 And yeah, all reverb presets have different _default send levels_. Anyway, all this will be explained in detail in the manual.

Kappa971:
I'm in favor of adding more details to make sure people are better informed, but that sentence didn't make much sense and was confusing so the solutions were:
1. reformulate it so that it was more understandable (and this does not mean making it longer or writing a poem, but reworking it).
2. remove it if it is not possible to do what was said previously, because it would only create confusion for the reader.

Now, @kcgen says it was removed because it was superfluous, maybe. He certainly didn't quite understand its meaning either (and I don't blame him).

In any case, that's fine.

PR #3072, opened 2023-10-29
Minor string fixes in 'MIXER' command

FeralChild64:
# Description

Fixes of two minor string issues I noticed during translation:
- there was a full stop instead of comma in one place
- if the help message uses 'x', 'c', and 'r' (lowercase), error messages should use the same capitalization

# Checklist

I have:

### DIFF
--- a/src/dos/program_mixer.cpp
+++ b/src/dos/program_mixer.cpp
@@ -808,7 +808,7 @@ void MIXER::AddMessages()
 	        "                 volumes of stereo channels separately (e.g. [color=white]10:20[reset], [color=white]150:d6[reset])\n"
 	        "    Stereo mode: [color=white]stereo[reset], or [color=white]reverse[reset] (stereo channels only)\n"
 	        "    Crossfeed:   [color=white]x0[reset] to [color=white]x100[reset], sets crossfeed strength (stereo channels only)\n"
-	        "    Reverb:      [color=white]r0[reset] to [color=white]r100[reset]. sets reverb level\n"
+	        "    Reverb:      [color=white]r0[reset] to [color=white]r100[reset], sets reverb level\n"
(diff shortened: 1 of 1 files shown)


src/dos/program_mixer.cpp
```diff
@@ -881,15 +881,15 @@ void MIXER::AddMessages()
 	        "(must be between 0 and 100)");
 
 	MSG_Add("SHELL_CMD_MIXER_MISSING_GLOBAL_CROSSFEED_STRENGTH",
-	        "Missing global crossfeed strength after [color=white]X[reset] "
+	        "Missing global crossfeed strength after [color=white]x[reset] "
 	        "(must be between 0 and 100)");
 
 	MSG_Add("SHELL_CMD_MIXER_MISSING_GLOBAL_CHORUS_LEVEL",
-	        "Missing global chorus level after [color=white]C[reset] "
+	        "Missing global chorus level after [color=white]c[reset] "
```
johnnovak, line 888: Hmm, I used uppercase so the singe-char command names stand out more in the error meesages.

I'm wondering if itwould be better to use uppercase everywhere instead?

FeralChild64: In my opinion it stands out well enough due to bright color. Uppercase is less convenient for the user, as he has to press SHIFT.

johnnovak: Yeah, and our examples use all lowercase anyway. I was just wondering, but you're right 😄 

image

Lowercase it is then!

johnnovak (approved):
Thanks @FeralChild64 👍🏻

src/dos/program_mixer.cpp
kcgen: I agree; easy on the eyes. 
(The height difference actually helps too, versus the height of the numbers).

PR #3077, opened 2023-10-31
Fix a crash when using glshader=crt-auto-machine

### LINKED ISSUE #3059: Crash at startup with `crt-auto-machine` shader

GIT main - f861d58f894ed3862c8b4524ef4b8d6696583d14

Linux x86_64

_No response_

### Relevant hardware info

_No response_

With `crt-auto-machine` emulator crashes with exception very early during the startup.

1. In the configuration file set `glshader = crt-auto-machine`
2. Start DOSBox

_No response_

### Your configuration

```ini
[sdl]
fullscreen          = false
fullresolution      = desktop
windowresolution    = 1067x800
viewport_resolution = fit
host_rate           = auto
vsync               = auto
vsync_skip          = 7000
presentation_mode   = vfr
output              = opengl
texture_renderer    = auto

[render]
aspect             = true
integer_scaling    = off
glshader           = crt-auto-machine
```

### Provide a Log

```text
2023-10-27 00:14:18.248 | DISPLAY: Initialised 1067x800 windowed mode using bilinear scaling on display-0
2023-10-27 00:14:18.314 | OPENGL: Vendor: Intel
2023-10-27 00:14:18.314 | OPENGL: Version: 4.6 (Compatibility Profile) Mesa 23.1.8
2023-10-27 00:14:18.314 | OPENGL: GLSL version: 4.60
2023-10-27 00:14:18.314 | OPENGL: Pixel buffer object available
2023-10-27 00:14:18.314 | OPENGL: NPOT textures supported
2023-10-27 00:14:18.314 | RENDER: Using adaptive CRT shader based on the configured graphics adapter
2023-10-27 00:14:18.314 | VGA: Pixel-doubling enabled
2023-10-27 00:14:18.330 | VIDEO: Using the DOS video mode's frame rate
2023-10-27 00:14:18.338 | MEMORY: Using 7936 DOS memory pages (31 MB) at address: 0x7f99340fb010
2023-10-27 00:14:18.338 | RENDER: Auto-switched to shader 'interpolation/sharp'
2023-10-27 00:14:18.338 | VGA: Forcing no pixel-doubling
2023-10-27 00:14:18.338 | VGA: Unhandled video mode 00h

Loguru caught a signal: SIGFPE
Stack trace:
9       0x56354a63c761 _start + 33
8       0x7f99c1573ac5 __libc_start_main + 133
7       0x7f99c1573a0a /lib64/libc.so.6(+0x23a0a) [0x7f99c1573a0a]
6       0x56354a8414d3 sdl_main(int, char**) + 7635
5       0x56354a64da36 Config::Init() const + 70
4       0x56354a64d8bb Section::ExecuteInit(bool) + 363
3       0x56354a801318 RENDER_Init(Section*) + 808
2       0x56354a90f5b9 VGA_SetupDrawing(unsigned int) + 57
1       0x56354a90d36f setup_drawing() + 2735
0       0x7f99c15889c0 /lib64/libc.so.6(+0x389c0) [0x7f99c15889c0]
2023-10-27 00:14:18.338 | Signal: SIGFPE
```

- [X] Yes, I agree.

weirddan455:
# Description
This was a weird one.  I got an assert failure in the shader code because the machine global was invalid.  Removing the assert gave the same divide by 0 crash that @FeralChild64 got so I assume he was just running a Release build.

My solution is admittedly kind of a hack.  I'm just reading the machine type from the config (extracted code from DOSBOX_RealInit into a separate function) because the shader code gets run before Dosbox get initialized.

As @johnnovak mentioned in the issue thread, init ordering is kind of a mess although in this case I am not seeing anything platform specific.  I haven't tested on Windows but I would wager a guess that at least the assertion failure happens there as well.

I had to bust out gdb to figure out what was really going on here.  If you're interested in the details, read on.  The first time the shader code gets run, it's from this stack trace:

```
Thread 1 "dosbox" hit Breakpoint 1, ShaderManager::FindShaderAutoMachine[abi:cxx11]() const (
    this=0x555557c68680 <get_shader_manager()::shader_manager>) at ../../src/gui/shader_manager.cpp:531
531             if (video_mode.color_depth == ColorDepth::Composite) {
(gdb) bt
#0  ShaderManager::FindShaderAutoMachine[abi:cxx11]() const (this=0x555557c68680 <get_shader_manager()::shader_manager>)
    at ../../src/gui/shader_manager.cpp:531
#1  0x0000555555f6189d in ShaderManager::MaybeAutoSwitchShader (this=0x555557c68680 <get_shader_manager()::shader_manager>)
    at ../../src/gui/shader_manager.cpp:380
#2  0x0000555555f5f66e in ShaderManager::NotifyGlshaderSettingChanged (
    this=0x555557c68680 <get_shader_manager()::shader_manager>, shader_name="crt-auto-machine")
    at ../../src/gui/shader_manager.cpp:72
#3  0x0000555555f164d2 in handle_shader_changes () at ../../src/gui/render.cpp:898
#4  0x0000555555f16905 in RENDER_Init (sec=0x555558a270c0) at ../../src/gui/render.cpp:947
#5  0x0000555555f14550 in RENDER_Reinit () at ../../src/gui/render.cpp:338
#6  0x0000555555f5504b in set_output (sec=0x555558a18de0, wants_aspect_ratio_correction=true)
    at ../../src/gui/sdlmain.cpp:3407
#7  0x0000555555f56046 in GUI_StartUp (sec=0x555558a18de0) at ../../src/gui/sdlmain.cpp:3555
#8  0x0000555555bfcf43 in Section::ExecuteInit (this=0x555558a18de0, init_all=true) at ../../src/misc/setup.cpp:1234
#9  0x0000555555bfcd58 in Config::Init (this=0x555558a14950) at ../../src/misc/setup.cpp:1206
#10 0x0000555555f5bda2 in sdl_main (argc=1, argv=0x7fffffffe4f8) at ../../src/gui/sdlmain.cpp:4858
#11 0x0000555555bd1ce6 in main (argc=1, argv=0x7fffffffe4f8) at ../../src/main.cpp:30
```

As far as I can tell, the SDL section is always the first to get initialized (its init function is `GUI_StartUp`).  This in turn calls `RENDER_Init` which calls the shader code.  This all happens before the Dosbox section gets initialized so the shader manager doesn't know what to do about the machine type (triggering the assert).

Later on, the Render section gets initialized (calling for a 2nd time `RENDER_Init` and repeating this process.)

```
Thread 1 "dosbox" hit Breakpoint 1, ShaderManager::FindShaderAutoMachine[abi:cxx11]() const (
    this=0x555557c68680 <get_shader_manager()::shader_manager>) at ../../src/gui/shader_manager.cpp:531
531             if (video_mode.color_depth == ColorDepth::Composite) {
(gdb) bt
#0  ShaderManager::FindShaderAutoMachine[abi:cxx11]() const (this=0x555557c68680 <get_shader_manager()::shader_manager>)
    at ../../src/gui/shader_manager.cpp:531
#1  0x0000555555f6189d in ShaderManager::MaybeAutoSwitchShader (this=0x555557c68680 <get_shader_manager()::shader_manager>)
    at ../../src/gui/shader_manager.cpp:380
#2  0x0000555555f5f66e in ShaderManager::NotifyGlshaderSettingChanged (
    this=0x555557c68680 <get_shader_manager()::shader_manager>, shader_name="crt-auto-machine")
    at ../../src/gui/shader_manager.cpp:72
#3  0x0000555555f164d2 in handle_shader_changes () at ../../src/gui/render.cpp:898
#4  0x0000555555f16905 in RENDER_Init (sec=0x555558a270c0) at ../../src/gui/render.cpp:947
#5  0x0000555555bfcf43 in Section::ExecuteInit (this=0x555558a270c0, init_all=true) at ../../src/misc/setup.cpp:1234
#6  0x0000555555bfcd58 in Config::Init (this=0x555558a14950) at ../../src/misc/setup.cpp:1206
#7  0x0000555555f5bda2 in sdl_main (argc=1, argv=0x7fffffffe4f8) at ../../src/gui/sdlmain.cpp:4858
#8  0x0000555555bd1ce6 in main (argc=1, argv=0x7fffffffe4f8) at ../../src/main.cpp:30
```

By this point (if you're running a release build and the assert didn't trigger), the Dosbox section has been initialized and the shader manager is happy.  However, the shader manager reports that there has been a shader change (from "empty string" to something valid).  This makes the 2nd run of `RENDER_Init` take a branch to call into `VGA_SetupDrawing` which runs into a divide by 0 bug due to more uninitialized global variables.

https://github.com/dosbox-staging/dosbox-staging/blob/55a6f98e0394623d48a73a4b62e1b587faaa2e5c/src/gui/render.cpp#L947-L962

This PR kind of sidesteps that 2nd issue because now the shader manager gets it right the first time and the `needs_reinit` variable will be false.

## Related issues
Fixes #3059

# Manual testing
- Followed repo steps from #3059
- Stepped through with GDB to figure out wtf was going on
- Assertion failure no longer happens
- Divide by 0 no longer happens

# Checklist

I have:

### DIFF
--- a/src/gui/shader_manager.cpp
+++ b/src/gui/shader_manager.cpp
@@ -532,6 +532,13 @@ std::string ShaderManager::FindShaderAutoMachine() const
 		return GetCompositeShader();
 	}
 
+	// DOSBOX_RealInit may have not been run yet.
+	// If not, go ahead and set the globals from the config.
+	if (machine == MCH_INVALID) {
+		DOSBOX_SetMachineTypeFromConfig(
+		        static_cast<Section_prop*>(control->GetSection("dosbox")));
+	}
+
 	switch (machine) {
 	case MCH_HERC: return GetHerculesShader();
 
--- a/include/dosbox.h
+++ b/include/dosbox.h
@@ -74,12 +74,13 @@ extern bool mono_cga;
 enum MachineType {
 	// In rough age-order: Hercules is the oldest and VGA is the newest
 	// (Tandy started out as a clone of the PCjr, so PCjr came first)
-	MCH_HERC  = 1 << 0,
-	MCH_CGA   = 1 << 1,
-	MCH_TANDY = 1 << 2,
-	MCH_PCJR  = 1 << 3,
-	MCH_EGA   = 1 << 4,
-	MCH_VGA   = 1 << 5,
+	MCH_INVALID = 0,
+	MCH_HERC    = 1 << 0,
+	MCH_CGA     = 1 << 1,
+	MCH_TANDY   = 1 << 2,
+	MCH_PCJR    = 1 << 3,
+	MCH_EGA     = 1 << 4,
+	MCH_VGA     = 1 << 5,
 };
 
 extern MachineType machine;
--- a/include/dosbox.h
+++ b/include/dosbox.h
@@ -44,6 +44,7 @@ const char* MSG_GetRaw(const char*); // get messages (in UTF-8, without ANSI
 bool MSG_Exists(const char*);
 
 class Section;
+class Section_prop;
 
 typedef Bitu (LoopHandler)(void);
 
@@ -55,6 +56,8 @@ void DOSBOX_SetLoop(LoopHandler * handler);
 void DOSBOX_SetNormalLoop();
 
 void DOSBOX_Init(void);
+
+void DOSBOX_SetMachineTypeFromConfig(Section_prop* section);
 
 class Config;
 using config_ptr_t = std::unique_ptr<Config>;
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -343,17 +343,8 @@ static void DOSBOX_UnlockSpeed( bool pressed ) {
 	}
 }
 
-static void DOSBOX_RealInit(Section * sec) {
-	Section_prop * section=static_cast<Section_prop *>(sec);
-	/* Initialize some dosbox internals */
-	ticksRemain=0;
-	ticksLast=GetTicks();
-	ticksLocked = false;
-	DOSBOX_SetLoop(&Normal_Loop);
-
-	MAPPER_AddHandler(DOSBOX_UnlockSpeed, SDL_SCANCODE_F12, MMOD2,
-	                  "speedlock", "Speedlock");
-
+void DOSBOX_SetMachineTypeFromConfig(Section_prop* section)
+{
 	const auto arguments = &control->arguments;
 	if (!arguments->machine.empty()) {
 		//update value in config (else no matching against suggested values
@@ -395,6 +386,20 @@ static void DOSBOX_RealInit(Section * sec) {
 	// VGA-type machine needs an valid SVGA card and vice-versa
 	assert((machine == MCH_VGA && svgaCard != SVGA_None) ||
 	       (machine != MCH_VGA && svgaCard == SVGA_None));
+}
+
+static void DOSBOX_RealInit(Section* sec)
+{
+	Section_prop* section = static_cast<Section_prop*>(sec);
+	/* Initialize some dosbox internals */
+	ticksRemain = 0;
+	ticksLast   = GetTicks();
+	ticksLocked = false;
+	DOSBOX_SetLoop(&Normal_Loop);
+
+	MAPPER_AddHandler(DOSBOX_UnlockSpeed, SDL_SCANCODE_F12, MMOD2, "speedlock", "Speedlock");
+
+	DOSBOX_SetMachineTypeFromConfig(section);
 
 	// Set the user's prefered MCB fault handling strategy
 	DOS_SetMcbFaultStrategy(section->Get_string("mcb_fault_strategy").c_str());
--- a/src/ints/int10_char.cpp
+++ b/src/ints/int10_char.cpp
@@ -641,6 +641,8 @@ void INT10_WriteChar(uint8_t chr, uint8_t attr, uint8_t page, uint16_t count, bo
 			case MCH_HERC:
 			case MCH_TANDY:
 				break;
+
+			default: assertm(false, "Invalid MachineType value");
 		}
 	}
 
--- a/src/ints/int10_modes.cpp
+++ b/src/ints/int10_modes.cpp
@@ -676,6 +676,8 @@ void INT10_SetCurMode(void) {
 			}
 			if (mode_changed && CurMode->type==M_TEXT) SetTextLines();
 			break;
+
+		default: assertm(false, "Invalid MachineType value");
 		}
 
 		if (mode_changed) {
@@ -820,6 +822,8 @@ static bool INT10_SetVideoMode_OTHER(uint16_t mode, bool clearmem)
 		// handled in function INT10_SetVideoMode.
 		assert(false);
 		break;
+
+	default: assertm(false, "Invalid MachineType value");
 	}
 
 	//  Setup the VGA to the correct mode
@@ -979,6 +983,8 @@ static bool INT10_SetVideoMode_OTHER(uint16_t mode, bool clearmem)
 		// handled in function INT10_SetVideoMode.
 		assert(false);
 		break;
+
+	default: assertm(false, "Invalid MachineType value");
 	}
 
 	RealPt vparams = RealGetVec(0x1d);
--- a/src/ints/int10_pal.cpp
+++ b/src/ints/int10_pal.cpp
@@ -91,6 +91,8 @@ void INT10_SetSinglePaletteRegister(uint8_t reg, uint8_t val)
 	case MCH_HERC:
 	case MCH_CGA:
 		break;
+
+	default: assertm(false, "Invalid MachineType value");
 	}
 }
 
@@ -115,6 +117,8 @@ void INT10_SetOverscanBorderColor(uint8_t val)
 	case MCH_HERC:
 	case MCH_CGA:
 		break;
+
+	default: assertm(false, "Invalid MachineType value");
 	}
 }
 
@@ -151,6 +155,8 @@ void INT10_SetAllPaletteRegisters(PhysPt data)
 	case MCH_HERC:
 	case MCH_CGA:
 		break;
+
+	default: assertm(false, "Invalid MachineType value");
 	}
 }
 
@@ -395,6 +401,8 @@ void INT10_SetBackgroundBorder(uint8_t val)
 
 	case MCH_HERC:
 		break;
+
+	default: assertm(false, "Invalid MachineType value");
 	}
 }
 


weirddan455:
Added an `MCH_INVALID = 0` type to the machine type enum to make PVS Studio happy.

src/gui/shader_manager.cpp
```diff
@@ -532,6 +532,13 @@ std::string ShaderManager::FindShaderAutoMachine() const
 		return GetCompositeShader();
 	}
 
+	// DOSBOX_RealInit may have not been run yet.
+	// If not, go ahead and set the globals from the config.
+	if (machine == MCH_INVALID) {
```
johnnovak, line 537: Maybe add a comment that this is a workaround we need to deal with the inconsistent/random/platform & config dependent init order.

johnnovak:
> As @johnnovak mentioned in the issue thread, init ordering is kind of a mess although in this case I am not seeing anything platform specific. I haven't tested on Windows but I would wager a guess that at least the assertion failure happens there as well.

Yeah, maybe this specific case is not platform-specific. My original adaptive shader stuff was fine on macOS, but when raising the final PR I tested it on Windows, and I got hard crashes in release builds. So I hit at least one platform-specific init order problem.

> My solution is admittedly kind of a hack. I'm just reading the machine type from the config (extracted code from DOSBOX_RealInit into a separate function) because the shader code gets run before Dosbox get initialized.

This is fine and it works, I'm just wondering if there's a less invasive / simpler solution. In my case, the fix for the problem I mentioned above was simply to return early from some functions when the render dimensions were still 0. That got me past the problem because that init function somehow got called twice, for the first time with zero render dimensions, then for the second time with the proper values... Yeah, it's a mess 🤷🏻

weirddan455:
>  This is fine and it works, I'm just wondering if there's a less invasive / simpler solution. In my case, the fix for the problem I mentioned above was simply to return early from some functions when the render dimensions were still 0. That got me past the problem because that init function somehow got called twice, for the first time with zero render dimensions, then for the second time with the proper values... Yeah, it's a mess 🤷🏻 E.g., you could try checking if the divisor in VGA_SetupDrawing is 0 that causes the problem, then return early if that's the case, and see if that fixes the problem...

That is basically what's going on.  I should have included more details about the divide by 0 crash, although my original description was already getting quite long.

```
Thread 1 "dosbox" received signal SIGFPE, Arithmetic exception.
0x00005555560afc71 in setup_drawing () at ../../src/hardware/vga_draw.cpp:2704
2704                             final_render_height / video_mode.height);
(gdb) bt
#0  0x00005555560afc71 in setup_drawing () at ../../src/hardware/vga_draw.cpp:2704
#1  0x00005555560aff3a in VGA_SetupDrawing () at ../../src/hardware/vga_draw.cpp:2792
#2  0x0000555555f16955 in RENDER_Init (sec=0x555558a270c0) at ../../src/gui/render.cpp:961
#3  0x0000555555bfceff in Section::ExecuteInit (this=0x555558a270c0, init_all=true) at ../../src/misc/setup.cpp:1234
#4  0x0000555555bfcd14 in Config::Init (this=0x555558a14950) at ../../src/misc/setup.cpp:1206
#5  0x0000555555f5bd5e in sdl_main (argc=1, argv=0x7fffffffe4f8) at ../../src/gui/sdlmain.cpp:4858
#6  0x0000555555bd1ce6 in main (argc=1, argv=0x7fffffffe4f8) at ../../src/main.cpp:30
```

https://github.com/dosbox-staging/dosbox-staging/blob/55a6f98e0394623d48a73a4b62e1b587faaa2e5c/src/hardware/vga_draw.cpp#L2702-L2704

This is from the 2nd time `RENDER_Init` gets called when the "render" section is actually initialized from the Config calling init on everything.

I dug a bit deeper and found the root cause is from this `calculate_vga_timings` function.  It's reading from a global `vga` struct which is plain old data with everything initialized to 0 at program start but valid data hasn't been filled out at the time this gets called.  I stepped through with a debugger to verify this.

The VGA timings get used for in some calculations and this is why `video_mode.height` is set to 0.

https://github.com/dosbox-staging/dosbox-staging/blob/55a6f98e0394623d48a73a4b62e1b587faaa2e5c/src/hardware/vga_draw.cpp#L1519-L1532

I wasn't sure how to fix this problem, and doing so won't fix the assertion failure anyway because that happens before this divide by 0 bug does.

I think I understand the problem you were saying about indeterminate/platform specific init and destroy if that's the same thing @kcgen was trying to fix in that PR was closed for getting too hard to maintain.  If I understand correctly, the problem there was global and file scoped objects with constructors and destructors getting automatically called by the C++ runtime library at program start and exit.

This I believe is a different problem because we're dealing with plain old data.  It's just a problem of various parts of the code reaching out to global variables before they get properly initialized by their respective modules.  Ideally we'd want to lock some of these things behind API boundaries or something so this can't happen (open to suggestions but this seems like a systematic issue with these global variables).

>  Can you also regression test it on Windows? Then I'll test on macOS at the end.

Yeah, I'll test on Windows here in a little bit.

src/gui/shader_manager.cpp
weirddan455: I believe this problem is going to be consistently reproducible across all platforms.  I'm going to test on Windows here in a little bit to confirm.

weirddan455:
Confirmed the same behavior happens in Windows.  Both the assertion failure and, if the assert is removed, the divide by 0 crash.

divide-error

kcgen:
> This I believe is a different problem because we're dealing with plain old data.

Yes; that's my understand too, @weirddan455.

The setup/control class is in charge of starting up each module in-order of their registration (all those _Init() functions along with their conf key-val descriptions in `dosbox.cpp`).

This includes the VGA and Render modules.

However, I think chunks of `sdlmain`'s code expects/uses the VGA or Render modules.

I'm all for as much re-ordering as needed of the big init chunks combined with asserts to catch unexpected empty/null values, to ensure the ordering is robust and correct.

I don't like the idea of quiet skips or repeated calls until eventually eveything is valid; the more we can get to the correct serialized startup sequence, the better!

weirddan455:
>  However, I think chunks of sdlmain's code expects/uses the VGA or Render modules even before they're init'd.

Yes, that's exactly what's happening.  sdlmain in `GUI_StartUp` is explicitly calling `RENDER_ReInit` before it gets initalized through the config system.  `RENDER_Init` is also (sometimes) calling into the VGA layer before it gets initialized as well.

>  I'm all for as much re-ordering as needed of startup pieces (combined with asserts to catch unexpected empty/null values), to ensure the ordering is robust and correct.

>  I don't like the idea of quiet skips or repeated calls until eventually everything is valid; the more we can get to the correct serialized startup sequence, the better!

Agree with all of that.  We've got modules calling into each other in their init code all willy-nilly and probably some circular dependencies as well.  This certainty needs to be cleaned up at some point.

What are you thoughts on the 0.81 feature freeze though?  I'm not sure what exactly we should be merging right now.  This PR (IMO) is a low risk, quick fix of a hard crash.  Ideally we'd fix the root cause by cleaning up the init code as you said but that is a larger, more risky change.

kcgen:
> What are you thoughts on the 0.81 feature freeze though? I'm not sure what exactly we should be merging right now. This PR (IMO) is a low risk, quick fix of a hard crash. Ideally we'd fix the root cause by cleaning up the init code as you said but that is a larger, more risky change.

I think your changes are better than the baseline and necessary - so absolutely should go in.

I also think that every possible code path this PR touches is going to be tested almost immediately, so there's little risk of a regression slipping through the cracks.

kcgen (approved):
Nice and small. Passing tests here on Linux and macOS.

kcgen:
I don't see why we would want to carry this known crash into 0.81; merging. Thanks for the fix, @weirddan455.

weirddan455:
Yeah, I figured we would need to fix the crash before release.  I was more talking about how the bigger job of cleaning up the init ordering and dependencies might have to wait until after 0.81 (unless you think it's important enough to try to solve now).

kcgen:
> Yeah, I figured we would need to fix the crash before release. I was more talking about how the bigger job of cleaning up the init ordering and dependencies might have to wait until after 0.81 (unless you think it's important enough to try to solve now).

Oh - yeah, I agree with you and @johnnovak to hold off on this larger re-ordering pursuit, especially if's not understood or what it will entail
(I was thinking that this PR itself was the item being considered to hold-off till post-0.81).

johnnovak:
> Yeah, I figured we would need to fix the crash before release. I was more talking about how the bigger job of cleaning up the init ordering and dependencies might have to wait until after 0.81 (unless you think it's important enough to try to solve now).

Absolutely, cleaning up the init order stuff is a huge tasks, now is not the time. One can work on that for many many months to get it right.

The global and static variables problem is well-understood and what you guys wrote above about the various issues that inject randomness into the init order is all 100% true @kcgen & @weirddan455.

I think there is _another_ source of indeterminism that is harder to see, although I haven't systematically verified it. I _think_ the different window managers on different OSes send UI events in different order in various circumstances. E.g., maybe on macOS you get these UI events through SDL on init:

- _Window Create, Resize, Show, Activate_

Then on Windows:

- _Windows Create, Activate, Show_ (so different order and no initial resize)

I'm completely making the example event streams up; I'm just trying to illustrate the theorethical problem here.

I'm pretty sure SDL just translates OS-issued events into SDL events (GLFW works this way that I know for sure), but does not remove events or inserts extra "synthetic" events to make the event stream _exactly the same_ across all supported platforms and platform versions on init and in response to various other actions (fullscreen toggling, minimize/maximize, etc.; for example, on Windows 7 and macOS you might get a resize event after a windows maximize, on Windows 10 and some Linux you might not get it... I did encounter such differences during UI dev work with GLFW at the very least).

Because different SDL events at startup trigger different call chains, this can also add a significant platform dependent "randomness" to the init call chain.

I brought this up several times, I think this is the 3rd or 4th time  but never got a reaction, maybe because the problem and how it can affect the init order is hard to see. In any case, if I'm right, this is important.

I thought "normalising" the events the app sees is the solution, i.e. remove events or insert synthetic SDL events so Staging "sees" the same events on the same order on all supported OSes. But now I'm starting to think that's too hard, a moving target with the millions of Linus window managers out there, not future proof, and ultimately brittle. The best approach would be that our init functions, at least for the graphics subsystem, assumed no fixed init order of incoming SDL events. It required a different thinking about tge problem and slightly different code that involves "lazy initing" many things (or most, or maybe all).

Hope that all makes some sense at last @kcgen and @weirddan455 😅

weirddan455:
>  I think the different window managers on different OSes send UI events in different order in various circumstances.

Yeah, almost certainly that is true.

>  I thought "normalising" the events the app sees is the solution, i.e. remove events or insert synthetic SDL events so Staging "sees" the same events in the same order on all supported OSes.

It's... complicated.  SDL does do some amount of normalizing but you're never going to guarantee exactly the same events in the same order across platforms.

>  But now I'm starting to think that's too hard, a moving target with the millions of Linux window managers out there, not future proof, and ultimately brittle.

Differences in Linux window managers are the least of the problems with this approach.  SDL also supports Android, iPhone, Web via emscripten, and even running directly from a Linux console (Dosbox Staging actually runs in this last mode).  These platforms have no concept of a "window" at all but SDL has to fake it.  You can't expect the same stream of events though.

>  The best approach would be that our init functions, at least for the graphics subsystem, assume no fixed init order of incoming SDL events. That requires a different thinking about the problem and slightly different code that involves "lazy initing" many things (or most, or maybe all).

From what I saw reading the code, our startup code does have a fixed ordering.  We have a data structure in the Config section that has function pointers to each module's Init + Destroy functions.  It looks like this is run before we even start polling for SDL events:

https://github.com/dosbox-staging/dosbox-staging/blob/e91cae410d1d279dfad2cde2b49314c55e42aebe/src/misc/setup.cpp#L1203-L1208

Perhaps what you're seeing though is certain modules re-initalizing themselves in response to certain events.  Ex. maybe the renderer has to do some re-initalizing when there's a change in window size.

I think the crux of the problem is that there's a lot of cross-talk between the different modules and some incorrect assumptions being made as to their state.  Ex. With this divide by 0 bug, the renderer is assuming based on its own state that VGA is ready to go which it turns out is an incorrect assumption.

>  I brought this up several times, I think this is the 3rd or 4th time, but never got a reaction, maybe because the problem and how it can affect the init order is hard to see. In any case, if I'm right, this is important.

This is the first I'm reading about this in detail.  I've probably just missed the other times you've brought it up.  I think it would be useful if you could raise a ticket containing "code smells" as a result of this problem so that we can go over these when it's time to clean some of this stuff up.  You mentioned you had to do some platform specific workarounds for the shader code.  Maybe add a reference to that code so that we can have something specific to investigate.

johnnovak:
> It's... complicated. SDL does do some amount of normalizing but you're never going to guarantee exactly the same events in the same order across platforms.
 
> Differences in Linux window managers are the least of the problems with this approach. SDL also supports Android, iPhone, Web via emscripten, and even running directly from a Linux console (Dosbox Staging actually runs in this last mode). These platforms have no concept of a "window" at all but SDL has to fake it. You can't expect the same stream of events though.

Interesting, and I'm not that familiar with SDL as I'm with GLFW. That console use-case is particularly interesting, I never considered that special scenario but it makes perfect sense SDL has to fake some events in that case.

> Perhaps what you're seeing though is certain modules re-initalizing themselves in response to certain events. Ex. maybe the renderer has to do some re-initalizing when there's a change in window size.

Yeah, most likely, and it's a confusing combination of things. I'm dead convinced there are many things during init that just work literally by fluke. In other words, no one understands _why_ certain things work because it's a complex interaction of DOSBox-side code triggered by SDL events, then if we change something on the code or something in SDL changes, things break... then we add more bandaids to patch things up because we don't even understand what's really happening and worse yet, we don't have _plan_ to make this more deterministic (maybe perfection is not achievable, but we could surely improve the current undefined-and-works-by-fluke-mostly mess).

> I think the crux of the problem is that there's a lot of cross-talk between the different modules and some incorrect assumptions being made as to their state. Ex. With this divide by 0 bug, the renderer is assuming based on its own state that VGA is ready to go which it turns out is an incorrect assumption.

100%, and it's not understood.
 
> This is the first I'm reading about this in detail. I've probably just missed the other times you've brought it up. I think it would be useful if you could raise a ticket containing "code smells" as a result of this problem so that we can go over these when it's time to clean some of this stuff up. You mentioned you had to do some platform specific workarounds for the shader code. Maybe add a reference to that code so that we can have something specific to investigate.

Yeah, I always just wrote my similar comments in response to some ticket. Raising a ticket for this is a good idea and I'll do that once I'm back in full capacity 🚀 

Appreciate your input @weirddan455 ❤️

PR #3081, opened 2023-11-02
Set the icon on window regeneration

weirddan455:
# Description
The icon currently gets lost when the window is regenerated.  To reproduce:

1. Type `fullscreen=true` into Dosbox (doesn't work but it still regenerates the window, changing any other setting in the [sdl] group should also suffice).
2. Icon goes missing in both the top left of the window and at the bottom where the open windows are shown.  On KDE, it gets replaced by a default Xorg icon.

Note, this is probably not reproducible on Mac or MSVC builds as those don't  have an empty `SetIcon` function:

https://github.com/dosbox-staging/dosbox-staging/blob/e91cae410d1d279dfad2cde2b49314c55e42aebe/src/gui/sdlmain.cpp#L534-L544

The fix is to move the `SetIcon()` call into the `set_output` function.  This gets called in both the inital startup and when the window gets regenerated.  Also moved setting the window title in there for good measure (this gets re-set later anyway to show cycles/ms so it probably doesn't matter much).

# Manual testing
Confirmed icon and window title are getting set properly and stay set after a window regeneration.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -316,9 +316,9 @@ void GFX_SetTitle(const int32_t new_num_cycles, const bool is_paused = false)
 	char title_buf[200] = {0};
 
 #if !defined(NDEBUG)
-	#define APP_NAME_STR "DOSBox Staging (debug build)"
+	#define APP_NAME_STR DOSBOX_NAME " (debug build)"
 #else
-	#define APP_NAME_STR "DOSBox Staging"
+	#define APP_NAME_STR DOSBOX_NAME
 #endif
 
 	auto &num_cycles      = sdl.title_bar.num_cycles;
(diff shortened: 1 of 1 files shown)


src/gui/sdlmain.cpp
```diff
@@ -3403,6 +3403,8 @@ static void set_output(Section* sec, const bool wants_aspect_ratio_correction)
 	const auto transparency = clamp(section->Get_int("transparency"), 0, 90);
 	const auto alpha = static_cast<float>(100 - transparency) / 100.0f;
 	SDL_SetWindowOpacity(sdl.window, alpha);
+	SDL_SetWindowTitle(sdl.window, "DOSBox Staging");
+	SetIcon();
```
kcgen, line 3407: Suggest using `DOSBOX_NAME`, which is a #define that comes in from the config system:

```
#define DOSBOX_NAME "DOSBox Staging"
```

(Used in a couple places)

kcgen (approved):
(just a small suggestion)
Nice catch, @weirddan455 !

src/gui/sdlmain.cpp
weirddan455: Pushed a new commit to address this.  There was another one called `APP_NAME_STR` that appends `(debug build)` for the debug build.

kcgen: Oh, another de-dupe - nice @weirddan455!

kcgen (approved):

PR #3089, opened 2023-11-04
Fix memory access in VirtualBox device, enable VirtualBox mouse protocol

FeralChild64:
# Description

Problem reason: VirtualBox device implementation used memory access function which were paging aware, while we are supposed to use functions which ignore paging. This is now fixed, plus several other improvements were implemented - back then I didn't test the support too much when I realized it didn't work with Windows.

#### DOS driver

Unfortunately, the unofficial VirtualBox DOS mouse driver works reliably only when real MS-DOS is booted, with our simulated DOS it is unusable. Hopefully, this is not a serious problem, as our driver already provides seamless mouse integration.

#### Mouse mapping

If we map a physical mouse A to our virtual DOS driver, and a physical mouse B to a PS/2 driver, and guest side is using mouse driver with virtual machine integration - we have a problem.
In case of VMware driver, we cheat to the driver: we pretend the integration is OK, but in reality the driver only gets events from the selected physical mouse.
With VirtualBox protocol... the driver can tell us that it won't be drawing the cursor by itself, and we should just show the host OS cursor. And we have no good way to reject the request from the driver (especially once it is already running...), and we have absolutely no way to tell the host OS to use only the particular selected mouse and ignore all the others.
If something like this happens, we are removing all the mouse mappings to prevent problems.

#### Desktop-less environment, without host mouse cursor

Same problem might happen on some exotic environments, when the guest side driver wants us to draw host side mouse cursor, and we have no possibility to do so.
Since there is no easy way to handle this situation gracefully (unless we want to emulate part of the desktop environment and draw our own mouse cursor), VirtualBox mouse interface is disabled in such environments - it's not useful in this case anyway.

# Manual testing

Played with the drivers:

- unofficial Windows 3.x driver: https://git.javispedro.com/cgit/vbmouse.git/about/
- unofficial DOS driver: https://git.javispedro.com/cgit/vbados.git/about/

The Windows 3.x driver seems to be very stable, plus if the VirtualBox mouse protocol is disabled, it reverts to classic PS/2.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -344,14 +344,22 @@ bool GFX_HaveDesktopEnvironment()
 	// https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys
 	// https://askubuntu.com/questions/72549/how-to-determine-which-window-manager-and-desktop-environment-is-running
 	// https://unix.stackexchange.com/questions/116539/how-to-detect-the-desktop-environment-in-a-bash-script
-	//
-	constexpr const char* vars[] = {"XDG_CURRENT_DESKTOP",
-	                                "XDG_SESSION_DESKTOP",
-	                                "DESKTOP_SESSION",
-	                                "GDMSESSION"};
 
-	return std::any_of(std::begin(vars), std::end(vars), std::getenv);
+	static bool already_checked          = false;
+	static bool have_desktop_environment = false;
+	if (!already_checked) {
+		constexpr const char* vars[] = {"XDG_CURRENT_DESKTOP",
+		                                "XDG_SESSION_DESKTOP",
+		                                "DESKTOP_SESSION",
+		                                "GDMSESSION"};
 
+		have_desktop_environment = std::any_of(std::begin(vars),
+		                                       std::end(vars),
+		                                       std::getenv);
+		already_checked = true;
+	}
+
+	return have_desktop_environment;
 #else
 	// Assume we have a desktop environment on all other systems
 	return true;
--- a/include/mouse.h
+++ b/include/mouse.h
@@ -26,12 +26,6 @@
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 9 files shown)


kcgen:
:butter: smooth mouse action, here we come!
Amazing find, @FeralChild64 .

johnnovak:
Make sure to test with a couple of Win 3.x adventure games such as Myst, Shivers, or any other adventure game, basically. Those used to crash with these drivers in no time.

FeralChild64:
@johnnovak I don't even have any, I think. I have played with Lotus SmartSuite, Total Commander (it used to crash VMWare unofficial driver rather quickly), also played Critical Mass for an hour.

johnnovak:
I'm still getting visual glitches (occasional tearing, see screenshot below) in Myst with the VMware driver, but the VBMouse driver seems to be fine. Well, at least there are no crashes. Will test with other games too later.

image

src/gui/sdlmain.cpp
```diff
@@ -381,14 +381,22 @@ bool GFX_HaveDesktopEnvironment()
...
 	// https://unix.stackexchange.com/questions/116539/how-to-detect-the-desktop-environment-in-a-bash-script
-	//
-	constexpr const char* vars[] = {"XDG_CURRENT_DESKTOP",
-	                                "XDG_SESSION_DESKTOP",
-	                                "DESKTOP_SESSION",
-	                                "GDMSESSION"};
 
-	return std::any_of(std::begin(vars), std::end(vars), std::getenv);
+	static bool already_checked          = false;
+	static bool have_desktop_environemnt = false;
```
kcgen, line 386: (minor: typo in `environment`)

kcgen (approved):
Looks great, @FeralChild64.
Just a minor comment.

src/gui/sdlmain.cpp
FeralChild64: Fixed.

kcgen:
Comments addressed and passing CI; merging! Thanks, @FeralChild64.

PR #3093, opened 2023-11-05
Move environment functions to psp (+ other small changes)

MeAreJeenius:
# Description

Most of the changes involve moving the functions querying and commanding the environment out of the shell class and into the psp class. These function only operate on the shell's psp anyways so it makes no sense to have them outside the psp. This also means that programs that query the environment no longer have to go through the hacky global `first_shell`. Removing `first_shell` was the primary motivation for these commits, though after these changes it is still being used in autoexec.

There are also a few single-commit changes in here:

Moved shell_history out of individual functions and into its own class.

Fixed a crash in the setver program when too few args are present.

Vastly improved the semantics of the FileReader class (the interface remains unchanged).

More details in a comment below.

# Manual testing

Ensure that read, write, and list environment functions work as expected with the `SET` command.

- `SET PATH`
- `SET VAR=VAL`
- `SET`

`MOUNT -Z` changes the letter of the Z drive. Note that you can't run this more than once per session. From what I can tell, this was a hack for WINE, since there was a conflict with the WINE Z drive. In addition to the environment interface change, I made the change propagate across subshells, whereas it would break your path before.

 - Run `MOUNT -Z E`. Ensure that your PATH variable has changed to include E: instead of Z:. Go into a subshell and ensure that the change is visible there too.
 - Restart and run the command again, this time from a subshell. Ensure the changes are reflected in the parent shell.

ParseLine is functionally unchanged. Only the name resolution for temp files was touched. Launch from terminal so you can see logging ouptut. Input any command with the vertical bar pipe. It can be a garbage command like `a | a`. In the logging output, you should see `Redirecting output to pipeXXXX.tmp`. Now try again while setting a `TMP` or `TEMP` environment variable to a directory. If you pipe again, the file will be prefixed by the chosen directory.

Environment variable resolution is functionally unchanged. Create a batch file with the line `ECHO %PATH%`, and ensure that your path is printed. Do the same directly on the DOS command line when the settings to get shell expansion there are set to on.

Executable resolution is functionally unchanged. Ensure all directories in your `PATH` work.

The `PATH` command is functionally unchanged. Ensure it matches the output of `SET PATH`.

The `DIR` command is functionally unchanged. Ensure that it works without the `DIRCMD` environment variable being set. Also ensure that it applies the expected switches when `DIRCMD` is set. For example, `SET DIRCMD=/b` should apply the `/b` switch automatically when using `DIR`.

The `CONFIG` command is functionally unchanged. The writes to the `CONFIG` environment variable should be the same as before.

- `CONFIG -GET VER` echoes current does version, and sets the `CONFIG` environment variable to that value
- `CONFIG -GET DOS VER` does the same

## Shell History Tests
Make sure the config file can be read from and written to.

When the sum of lines in the history file and the new commands are over 500 (or whatever the specified constant in the class is), ensure that the file contains only the most recent 500 lines upon exit.

Ensure that exit never written to the file.

Remove read permissions and ensure that dosbox doesn't explode. Then do that again, but make the file write-only.

# Checklist

I have:

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -27,6 +27,7 @@
 #include <algorithm>
 #include <array>
 #include <cstddef>
+#include <optional>
 #include <string>
 #include <type_traits>
 
@@ -354,15 +355,32 @@ public:
 	MemStruct(uint16_t seg, uint16_t off) : pt(PhysicalMake(seg, off)) {}
 	MemStruct(RealPt addr) : pt(RealToPhysical(addr)) {}
 
-	void SetPt(uint16_t seg) { pt = PhysicalMake(seg, 0); }
+	void SetPt(uint16_t seg)
+	{
+		pt = PhysicalMake(seg, 0);
+	}
 
 protected:
-	PhysPt pt = 0;
+	PhysPt pt    = 0;
+	~MemStruct() = default;
+};
+
+class Environment {
+public:
+	virtual std::optional<std::string> GetEnvironmentValue(std::string_view entry) const = 0;
+
+	virtual ~Environment() = default;
+
+protected:
+	Environment()                              = default;
+	Environment(const Environment&)            = default;
+	Environment& operator=(const Environment&) = default;
+	Environment(Environment&&)                 = default;
+	Environment& operator=(Environment&&)      = default;
 };
 
 /* Program Segment Prefix */
-
-class DOS_PSP final : public MemStruct {
+class DOS_PSP final : public MemStruct, public Environment {
 public:
 	DOS_PSP(uint16_t segment) : seg(segment) { SetPt(seg); }
 
@@ -416,6 +434,11 @@ public:
 	void SetFCB1(RealPt src);
 	void SetFCB2(RealPt src);
 	void SetCommandTail(RealPt src);
+
+	std::optional<std::string> GetEnvironmentValue(std::string_view variable) const override;
+	std::vector<std::string> GetAllRawEnvironmentStrings() const;
+	bool SetEnvironmentValue(std::string_view variable,
+	                         std::string_view new_string);
 
 private:
 	#ifdef _MSC_VER
--- a/include/shell.h
+++ b/include/shell.h
@@ -55,7 +55,7 @@ public:
 
 class HostShell {
 public:
-	virtual bool GetEnvStr(const char* entry, std::string& result) const = 0;
+	virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
 
 	HostShell()                            = default;
 	HostShell(const HostShell&)            = delete;
@@ -162,7 +162,7 @@ public:
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	bool GetEnvStr(const char* entry, std::string& result) const override;
+	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
 	[[nodiscard]] Bitu GetEnvCount() const;
 	bool SetEnv(std::string_view entry, std::string_view new_string);
--- a/include/shell.h
+++ b/include/shell.h
@@ -163,8 +163,7 @@ public:
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
 	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
-	bool GetEnvNum(Bitu num, std::string& result) const;
-	[[nodiscard]] Bitu GetEnvCount() const;
+	std::vector<std::string> GetAllEnvVars() const;
 	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 	/* Commands */
--- a/include/shell.h
+++ b/include/shell.h
@@ -165,7 +165,7 @@ public:
 	bool GetEnvStr(const char* entry, std::string& result) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
 	[[nodiscard]] Bitu GetEnvCount() const;
-	bool SetEnv(const char* entry, const char* new_string);
+	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 	/* Commands */
 	void CMD_HELP(char* args);
--- a/include/shell.h
+++ b/include/shell.h
@@ -150,8 +150,8 @@ public:
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	std::optional<std::string> GetEnvStr(std::string_view entry) const;
-	std::vector<std::string> GetAllEnvVars() const;
+	// HACK: Don't use in new code
+	// TODO: Remove the call to this function from autoexec
 	bool SetEnv(std::string_view entry, std::string_view new_string);
 
 	/* Commands */
--- a/include/shell.h
+++ b/include/shell.h
@@ -103,6 +103,25 @@ struct SHELL_Cmd {
 	HELP_Category category = HELP_Category::Misc;
 };
 
+class ShellHistory {
+public:
+	std::vector<std::string> GetCommands(uint16_t code_page) const;
+	void Append(std::string command, uint16_t code_page);
+
+	ShellHistory();
+	~ShellHistory();
+	ShellHistory(const ShellHistory&)            = delete;
+	ShellHistory& operator=(const ShellHistory&) = delete;
+	ShellHistory(ShellHistory&&)                 = delete;
+	ShellHistory& operator=(ShellHistory&&)      = delete;
+
+private:
+	std::vector<std::string> commands{};
+	std_fs::path path;
+};
+
+extern std::unique_ptr<ShellHistory> global_shell_history;
+
 class DOS_Shell : public Program, public HostShell {
 private:
 	void PrintHelpForCommands(MoreOutputStrings& output, HELP_Filter req_filter);
@@ -113,15 +132,17 @@ private:
 	[[nodiscard]] std::string Which(std::string_view name) const;
 
 	friend class AutoexecEditor;
-	std::vector<std::string> utf8_history{};
-	std::stack<BatchFile> batchfiles{};
 
+	ShellHistory& history                  = *global_shell_history;
+	std::stack<BatchFile> batchfiles       = {};
+	uint16_t input_handle                  = STDIN;
+	bool call                              = false;
 	bool exit_cmd_called                   = false;
 	static inline bool help_list_populated = false;
 
 public:
 	DOS_Shell();
-	~DOS_Shell() override = default;
+	~DOS_Shell() override                  = default;
 	DOS_Shell(const DOS_Shell&)            = delete; // prevent copy
 	DOS_Shell& operator=(const DOS_Shell&) = delete; // prevent assignment
 	void Run() override;
@@ -140,9 +161,6 @@ public:
 	virtual bool ExecuteShellCommand(const char* const name, char* arguments);
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
-
-	void ReadShellHistory();
-	void WriteShellHistory();
 
 	bool GetEnvStr(const char* entry, std::string& result) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
@@ -183,10 +201,7 @@ public:
 	void CMD_VOL(char* args);
 	void CMD_MOVE(char* args);
 
-	/* The shell's variables */
-	uint16_t input_handle         = 0;
-	bool echo                     = false;
-	bool call                     = false;
+	bool echo = true;
 };
 
 std::tuple<std::string, std::string, std::string, bool> parse_drive_conf(
--- a/include/shell.h
+++ b/include/shell.h
@@ -53,21 +53,9 @@ public:
 	virtual ~ByteReader()                    = default;
 };
 
-class HostShell {
-public:
-	virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
-
-	HostShell()                            = default;
-	HostShell(const HostShell&)            = delete;
-	HostShell& operator=(const HostShell&) = delete;
-	HostShell(HostShell&&)                 = delete;
-	HostShell& operator=(HostShell&&)      = delete;
-	virtual ~HostShell()                   = default;
-};
-
 class BatchFile {
 public:
-	BatchFile(const HostShell& host, std::unique_ptr<ByteReader> input_reader,
+	BatchFile(const Environment& host, std::unique_ptr<ByteReader> input_reader,
 	          std::string_view entered_name, std::string_view cmd_line,
 	          bool echo_on);
 	BatchFile(const BatchFile&)            = delete;
@@ -86,7 +74,7 @@ private:
 	[[nodiscard]] std::string ExpandedBatchLine(std::string_view line) const;
 	[[nodiscard]] std::string GetLine();
 
-	const HostShell& shell;
+	const Environment& shell;
 	CommandLine cmd;
 	std::unique_ptr<ByteReader> reader;
 	bool echo;
@@ -122,7 +110,7 @@ private:
 
 extern std::unique_ptr<ShellHistory> global_shell_history;
 
-class DOS_Shell : public Program, public HostShell {
+class DOS_Shell : public Program {
 private:
 	void PrintHelpForCommands(MoreOutputStrings& output, HELP_Filter req_filter);
 	void AddShellCmdsToHelpList();
@@ -162,7 +150,7 @@ public:
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
+	std::optional<std::string> GetEnvStr(std::string_view entry) const;
 	std::vector<std::string> GetAllEnvVars() const;
 	bool SetEnv(std::string_view entry, std::string_view new_string);
 
--- a/src/dos/dos_classes.cpp
+++ b/src/dos/dos_classes.cpp
@@ -26,6 +26,7 @@
 #include <cstring>
 
 #include "mem.h"
+#include "string_utils.h"
 #include "support.h"
 
 static void dos_memset(PhysPt addr, uint8_t val, size_t n)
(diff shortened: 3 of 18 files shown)


johnnovak:
I haven't forgotten about this @MeAreJeenius , but I'm busy with other stuff that I want to go into the 0.81.0 release. We'll try to fit this in too, but reviewing & testing this is the last item on my list as it's a rather large change. Stay tuned!

include/dos_inc.h
```diff
@@ -353,17 +354,34 @@ class MemStruct {
...
+	Environment& operator=(const Environment&) = default;
+	Environment(Environment&&)                 = default;
+	Environment& operator=(Environment&&)      = default;
+};
+
+/* Program Segment Prefix */
+class DOS_PSP final : public MemStruct, public Environment {
 public:
-	DOS_PSP(uint16_t segment) : seg(segment) { SetPt(seg); }
+	DOS_PSP(uint16_t segment) : seg(segment) { SetPt(seg); }
```
FeralChild64, line 384: Unnecessary tab at the end of line.

include/dos_inc.h
```diff
@@ -416,6 +434,10 @@ class DOS_PSP final : public MemStruct {
 	void SetFCB2(RealPt src);
 	void SetCommandTail(RealPt src);
 
+	std::optional<std::string> GetEnvironmentValue(std::string_view entry) const override;
```
FeralChild64, line 437: `EnvironmentVariable` sounds better to me than `EnvironmentValue`

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
 	return true;
 }
 
+static constexpr auto bytes_to_read = 1024;
```
FeralChild64, line 312: Why 1024? Is this some hard limit for our DOS, enforced in other place? If so, it would be nice to have some comment. If I remember correctly, MS-DOS had a limit of 127 characters, or so.

If there is no real limit, maybe it would be better to have some dynamic buffer?

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		return {};
+	}
+
+	for (;;) {
+		MEM_StrCopy(env_read, env_string, bytes_to_read);
+		if (!env_string[0]) {
+			return {};
+		}
+		env_read += (PhysPt)(safe_strlen(env_string) + 1);
+		char* equal = strchr(env_string, '=');
```
FeralChild64, line 330: Avoid C-style string operations and pointer arithmetic in new code.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+	PhysPt env_write                   = env_read;
+	PhysPt env_write_start             = env_read;
+	char env_string[bytes_to_read + 1] = {0};
+	const auto entry_length            = variable.size();
+	do {
+		MEM_StrCopy(env_read, env_string, bytes_to_read);
+		if (!env_string[0]) {
+			break;
+		}
+		env_read += (PhysPt)(safe_strlen(env_string) + 1);
```
FeralChild64, line 380: Use static_cast f necessary, we try to avoid C-style casting in new code.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		if (!strchr(env_string, '=')) {
+			continue; /* Remove corrupt entry? */
```
FeralChild64, line 382: I suggest to add a warning log about damaged environment. Preferably have some variable (like `bool already_warned`) to avoid flooding the output log.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+
+	// ensure room
+	if (envsize <= (env_write - env_write_start) + variable.size() + 1 +
+	                       new_string.size() + 2) {
+		return false;
+	}
+
+	if (!new_string.empty()) {
+		std::string bigentry(variable);
+		for (std::string::iterator it = bigentry.begin();
```
FeralChild64, line 405: We have` void upcase(std::string& str);` in `string_utils.h` for this.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		     it != bigentry.end();
+		     ++it) {
+			*it = toupper(*it);
+		}
+		snprintf(env_string,
```
FeralChild64, line 410: Use std::string here, and `format_string` from `string_utils.h`.

src/dos/dos_classes.cpp
kcgen: > Why 1024? Is this some hard limit for our DOS, enforced in other place? If so, it would be nice to have some comment. If I remember correctly, MS-DOS had a limit of 127 characters, or so.

MS-DOS's environment space can be dialed up to 1024 characters with the `/e` (environment) flag at boot-time (I think in `CONFIG.SYS`):

```
SHELL=COMMAND.COM /e:1024
```

Maybe that's the connection here?

johnnovak:
@MeAreJeenius Just updating you on the current state of affairs regarding the upcoming 0.81.0 release and how it relates to this PR of yours as we definitely don't want to give the impression that we're stringing you along! So just setting expectations here 😄 

We're aiming to release the 0.81.0 Release Candidate on 22 Dec, just before Christmas. This means we want everything in-flight and planned to be merged and tested by 15 Dec, then we'll have one more week to finish the release notes, do some internal testing, and fix any potential showstoppers.

That's a bit tight but doable. But we've decided this PR will have to wait until after 0.81.0 is out. It contains a large number of refactors/changes that are very welcome, no doubt, but these are risky changes that need comprehensive regression testing and there are few user-observable improvements to warrant the relatively large risk as a "last-minute merge". Simply put, we have no capacity to give this PR the attention it deserves before the release.

So when can we return to this to get it reviewed, tested, and merged? We've decided to have a code freeze once the release candidate is out, which will last at least a month, but no longer than two months. So realistically we can continue the work on this PR around February 2024 once the final 0.81.0 is out.

Thank you for your patience and see you then! Of course, feel free to improve or polish the PR in the meantime as you see fit!

johnnovak:
@MeAreJeenius Thanks for the patience. Now is the time to finalise this, then the team will review it in the coming weeks.

johnnovak:
I've noticed you keep improving this @MeAreJeenius, that's great.

Let us know when it's ready for final review.

What I'm missing from the description, though, is a _detailed_ test plan. The idea is that we do want to shift the effort to the author of the PR so then the reviewer's don't need to spend time to figure out how to test things. This is also important to remove any ambiguity and misunderstandings.

For example, instead of this sentence of yours:

>  Input any command with the vertical bar pipe. It can be a garbage command like a | a.

I'm expecting a detailed step-by-step test plan for testing the pipe command. You could come up with two or three (or more) test cases with exact commands to execute, and an exact description of the expected correct results. Again, the idea is that I can go through your test plan and mostly just execute the steps to verify them.

This is also important for catching regressions later on.

Speaking from personal experience, many times when I thought I'm "done" with the PR, once I started writing a detailed test plan, I realised I'm only maybe at 30-50% completion 😎 

Thank you.

johnnovak:
@MeAreJeenius Thanks for improving the test descriptions, this now looks a lot better, and it will make reviewers' job easier 👍🏻

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+static std_fs::path get_shell_history_path();
+static bool command_is_exit(std::string_view command);
+
+ShellHistory& ShellHistory::Instance()
+{
+	static auto history = ShellHistory();
+	return history;
+}
+
+void ShellHistory::Append(std::string command, std::uint16_t code_page)
```
FeralChild64, line 43: `std::uint16_t` -> `uint16_t`

We are not using the `std::` for integer types in the rest of the code.

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+#include "checks.h"
+#include "control.h"
+#include "dosbox.h"
+#include "fs_utils.h"
+#include "string_utils.h"
+
+CHECK_NARROWING();
+
+static constexpr int HistoryMaxLineSize = 256;
```
FeralChild64, line 31: `HistoryMaxLineLength` would IMHO look better

src/dos/dos_classes.cpp
FeralChild64: Was this set to `resolved` by mistake?

src/dos/dos_classes.cpp
FeralChild64: Was this set to `resolved` by mistake?

src/dos/dos_classes.cpp
FeralChild64: Was this set to `resolved` by mistake?

src/dos/dos_classes.cpp
FeralChild64: Was this set to `resolved` by mistake?

src/dos/dos_classes.cpp
FeralChild64: Was this set to `resolved` by mistake?

FeralChild64:
I have already played with the change the last year and I couldn't find any regression. Overall, I like the change - I only have few style-related commands.

src/shell/shell_history.cpp
johnnovak: Yep, that's right.

FeralChild64:
I know this is out of scope of this PR - but if you want to continue working on the shell, it would be nice to store environment variables like `BLASTER` or `ULTRASND` in a separate place, common for all the shell instances. Currently the sequence:

```
CONFIG -set sbtype=sb16
COMMAND
CONFIG -set sbtype=sbpro1
```

leads to the wrong value of BLASTER variable, which might cause issues with game launchers and BAT files changing the emulated sound card just before launching the game.

src/shell/shell_history.cpp
johnnovak: Yeah, it's length.

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+static constexpr int HistoryMaxNumLines = 500;
+
+static std_fs::path get_shell_history_path();
+static bool command_is_exit(std::string_view command);
+
+ShellHistory& ShellHistory::Instance()
+{
+	static auto history = ShellHistory();
+	return history;
+}
```
johnnovak, line 41: The problem with such static init stuff is that the init order becomes platform and compiler-dependent. I have plans how to refactor the whole init stuff eventually but we need to 100% eradicate all static and global init stuff. In the interim, just make this a regular class like for example `ImageCapturer` and just create an unique_ptr instance of it in the global scope that gets initialised by the init method (so not statically initialised).

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+void ShellHistory::Append(std::string command, std::uint16_t code_page)
+{
+	auto to_utf8_str = [code_page](const std::string& dos_str) {
+		std::string utf8_str{};
```
johnnovak, line 46: As discussed previously, please use the `= {}` initialiser format everywhere as the project conventions.

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+		dos_to_utf8(dos_str, utf8_str, code_page);
+		return utf8_str;
+	};
+	
+	trim(command);
+	
+	auto utf8_command = to_utf8_str(command);
```
johnnovak, line 53: `const`?

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+	// Must check arguments directly as control->SwitchToSecureMode()
+	// will not be called until the first shell is run
+	if (control->arguments.securemode) {
+		return;
+	}
+	if (path.empty()) {
+		return;
+	}
+	auto history_file = std::ifstream(path);
```
johnnovak, line 88: `const`? I won't make the same comment everywhere, but we try to maximise the use of `const`s where applicable 😎

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	for (const auto& command : commands) {
+		history_file << command << '\n';
+	}
+}
+
+static std_fs::path get_shell_history_path()
+{
+	const auto* section = dynamic_cast<Section_prop*>(control->GetSection("dos"));
```
johnnovak, line 132: `assert(section)` is more appropriate. If there is no `[dos]` section, that's a coding error and something went very wrong somewhere.

include/shell.h
```diff
@@ -55,7 +55,7 @@ class ByteReader {
 
 class HostShell {
 public:
-	virtual bool GetEnvStr(const char* entry, std::string& result) const = 0;
+	virtual std::optional<std::string> GetEnvStr(std::string_view entry) const = 0;
```
johnnovak: Much nicer signature🏅

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "fs_utils.h"
+#include "string_utils.h"
+
+CHECK_NARROWING();
+
+static constexpr int HistoryMaxLineSize = 256;
+static constexpr int HistoryMaxNumLines = 500;
+
+static std_fs::path get_shell_history_path();
+static bool command_is_exit(std::string_view command);
```
johnnovak, line 35: `const std::string_view`?

src/shell/shell_history.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+	const auto* path = section->Get_path("shell_history_file");
+	if (path == nullptr) {
+		return {};
+	}
+
+	return path->realpath;
+}
+
+static bool command_is_exit(std::string_view command)
```
johnnovak, line 145: `const std::string_view`? I dislike mutating input args, just introduce a temp var in the function.

include/shell.h
```diff
@@ -164,7 +164,7 @@ class DOS_Shell : public Program, public HostShell {
 	bool GetEnvStr(const char* entry, std::string& result) const override;
 	bool GetEnvNum(Bitu num, std::string& result) const;
 	[[nodiscard]] Bitu GetEnvCount() const;
-	bool SetEnv(const char* entry, const char* new_string);
+	bool SetEnv(std::string_view entry, std::string_view new_string);
```
johnnovak, line 154: `const std::string_view`?

include/shell.h
johnnovak: `const std::string_view`?

include/shell.h
```diff
@@ -161,7 +161,7 @@ class DOS_Shell : public Program, public HostShell {
 	bool ExecuteProgram(std::string_view name, std::string_view args);
 	bool ExecuteConfigChange(const char* const cmd_in, const char* const line);
 
-	bool GetEnvStr(const char* entry, std::string& result) const override;
+	std::optional<std::string> GetEnvStr(std::string_view entry) const override;
```
johnnovak: `const std::string_view`?

src/shell/shell_misc.cpp
```diff
@@ -644,36 +644,20 @@ std::optional<std::string> DOS_Shell::GetEnvStr(const std::string_view entry) co
 	}
 }
 
-bool DOS_Shell::GetEnvNum(Bitu num, std::string& result) const
+std::vector<std::string> DOS_Shell::GetAllEnvVars() const
 {
+	auto all_env_vars = std::vector<std::string>{};
```
johnnovak: `std::vector<std::string> all_env_vars = {};` is our style.

include/dos_inc.h
```diff
@@ -354,15 +355,32 @@ class MemStruct {
...
 protected:
-	PhysPt pt = 0;
+	PhysPt pt    = 0;
+	~MemStruct() = default;
 };
 
-/* Program Segment Prefix */
+class Environment {
+public:
+	virtual std::optional<std::string> GetEnvironmentValue(std::string_view entry) const = 0;
```
johnnovak, line 370: `const std::string_view`

include/dos_inc.h
```diff
@@ -417,6 +435,11 @@ class DOS_PSP final : public MemStruct {
 	void SetFCB2(RealPt src);
 	void SetCommandTail(RealPt src);
 
+	std::optional<std::string> GetEnvironmentValue(std::string_view variable) const override;
+	std::vector<std::string> GetAllRawEnvironmentStrings() const;
+	bool SetEnvironmentValue(std::string_view variable,
+	                         std::string_view new_string);
```
johnnovak, line 441: `const std::string_view`

src/dos/dos_classes.cpp
johnnovak: `BytesToRead` is the style for constants.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+	}
+}
+
+bool DOS_PSP::SetEnvironmentValue(std::string_view variable, std::string_view new_string)
+{
+	PhysPt env_read = PhysicalMake(GetEnvironment(), 0);
+
+	// Get size of environment.
+	DOS_MCB mcb(GetEnvironment() - 1);
+	uint16_t envsize = mcb.GetSize() * 16;
```
johnnovak, line 369: What's 16? minor: `env_size` would be better.

src/dos/dos_classes.cpp
```diff
@@ -308,6 +309,119 @@ bool DOS_PSP::SetNumFiles(uint16_t file_num)
...
+		         bytes_to_read + 1,
+		         "%s=%s",
+		         bigentry.c_str(),
+		         std::string(new_string).c_str());
+		MEM_BlockWrite(env_write,
+		               env_string,
+		               (Bitu)(safe_strlen(env_string) + 1));
+		env_write += (PhysPt)(safe_strlen(env_string) + 1);
+	}
+	/* Clear out the final piece of the environment */
```
johnnovak, line 420: minor: we generally use `//`.

src/dos/dos_classes.cpp
johnnovak: I think this code was just moved around, the commit message even says so. It's fine to leave it as this for now, then refactoring can happen in a later PR as this is already quite overloaded 😅 Let's not add more to it now.

src/shell/file_reader.cpp
```diff
@@ -18,42 +18,69 @@
...
+
 #include "file_reader.h"
 
-std::optional<std::unique_ptr<FileReader>> FileReader::GetFileReader(std::string_view file)
+std::optional<FileReader> FileReader::GetFileReader(const std::string& filename)
 {
-	auto reader = std::make_unique<FileReader>(file, PrivateOnly());
-	if (!reader->valid) {
-		return std::nullopt;
+	std::uint16_t handle = {};
```
johnnovak, line 28: Just `uint16_t`.

src/shell/file_reader.cpp
```diff
@@ -18,42 +18,69 @@
...
+		return {};
 	}
-	return reader;
+	return FileReader(handle);
 }
 
-FileReader::FileReader(std::string_view file, [[maybe_unused]] PrivateOnly key)
-        : filename(file),
-          valid(DOS_OpenFile(filename.c_str(), (DOS_NOT_INHERIT | OPEN_READ), &handle))
+FileReader::FileReader(std::uint16_t file_handle) : handle(file_handle) {}
```
johnnovak, line 35: `const uint16_t`

src/shell/file_reader.h
```diff
@@ -27,32 +27,23 @@
...
-	FileReader& operator=(FileReader&&)      = delete;
+
+	FileReader(FileReader&& other) noexcept;
+	FileReader& operator=(FileReader&& other) noexcept;
 
 private:
-	std::string filename = {};
-	uint16_t handle      = 0;
-	bool valid;
+	explicit FileReader(std::uint16_t file_handle);
```
johnnovak, line 45: `const uint16_t`

johnnovak:
Ok @MeAreJeenius I'm done reviewing this, please see my comments.

src/dos/dos_classes.cpp
johnnovak: Ok, wasn't sure if it was just moved 😄

include/shell.h
johnnovak: I'm far from being a C++ expert, but doesn't mismatching const specifier between function declarations and definitions cause problems? Is this a standard thing to do, do all compilers support such mismatching use of consts?

So far we've 100% matched the consts between the definitions and declarations in new code AFAIK. That's a model I can understand, at least 😅 Not saying I'm necessarily against this, but this is something unfamiliar to me.

johnnovak: Thanks for clarifying, that makes sense. I'm seeing now why is this preferable then, yeah. Cool trick 👍🏻 (and disregard my similar `const` comments)

src/dos/dos_classes.cpp
FeralChild64: Yes, but if changes are small, IMHO they are still worth cleaning up to match the current coding style.

johnnovak:
> Unrelated to the PR, but I noticed there are a couple of my stale branches still in the repo (prefixed with mj/). Those can be deleted.

Cool, I've just deleted them.

include/shell.h
```diff
@@ -113,15 +120,17 @@ class DOS_Shell : public Program, public HostShell {
 	[[nodiscard]] std::string Which(std::string_view name) const;
 
 	friend class AutoexecEditor;
-	std::vector<std::string> utf8_history{};
-	std::stack<BatchFile> batchfiles{};
 
+	ShellHistory& history = *global_shell_history;
+	std::stack<BatchFile> batchfiles{};
```
johnnovak, line 125: minor: `= {}`

johnnovak (approved):
This is looking very good now @MeAreJeenius, thanks for the patience and addressing the review comments.

I've made one last extremely minor comment, then it's only a matter of rebasing it and it's merge time! 🚀

PR #3099, opened 2023-11-06
Drop OpenGL PBO support

kcgen:
# Description

Thanks to @GranMinigun for this commit.

The pixel buffer object (PBO) harms performance on already slow-video systems: specifically those using shared-memory / integrated graphics, like SBCs, Intel HD, and AMD APUs.

Removing the PBO will give these systems a noticeable and valuable performance increase.

Removing OpenGL PBO has been discussed, extensively benchmarked on all platforms, and agreed by the team before:  https://github.com/dosbox-staging/dosbox-staging/pull/2450#issuecomment-1703884190

## Related issues

None.

# Manual testing

Testing has been done with the same commit here:

https://github.com/dosbox-staging/dosbox-staging/pull/2450#issuecomment-1703884190

# Checklist

I have:

### DIFF
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -171,13 +171,11 @@ struct SDL_Block {
 	struct {
 		SDL_GLContext context;
 		int pitch = 0;
-		void *framebuf = nullptr;
-		GLuint buffer;
+		void* framebuf = nullptr;
 		GLuint texture;
 		GLuint displaylist;
 		GLint max_texsize;
 		bool bilinear;
-		bool pixel_buffer_object = false;
 		bool npot_textures_supported = false;
 		bool use_shader;
 		bool framebuffer_is_srgb_encoded;
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -85,42 +85,6 @@
 #define APIENTRYP APIENTRY *
 #endif
 
-#ifndef GL_ARB_pixel_buffer_object
-#define GL_ARB_pixel_buffer_object 1
-#define GL_PIXEL_PACK_BUFFER_ARB           0x88EB
-#define GL_PIXEL_UNPACK_BUFFER_ARB         0x88EC
-#define GL_PIXEL_PACK_BUFFER_BINDING_ARB   0x88ED
-#define GL_PIXEL_UNPACK_BUFFER_BINDING_ARB 0x88EF
-#endif
-
-#ifndef GL_ARB_vertex_buffer_object
-#define GL_ARB_vertex_buffer_object 1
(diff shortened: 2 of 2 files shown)


FeralChild64 (approved):
Nice simplification!

kcgen:
Thanks for the review @FeralChild64 and the cleanup @GranMinigun. 🚀

johnnovak:
Integrated GPUs... man, why do they even exist 😐 Clearly, the boost with no-PBO is too large to ignore for those contraptions. A case of shit OpenGL drivers, most likely.

I'm very far from being an OpenGL expert, but the performance benefit of async texture uploads enabled by the use of PBOs would only manifest app-wide if the program can do some other tasks in the time gained by doing uploads the async way. If we can't use the "free time" gained, yeah, no difference.

So looks like it's a case of shit drivers slowing down the iGPUs when using PBOs combined with not using the gained "free time" for anything on dedicated GPUs. Well, at least that's my hunch. Plus what we're doing on the OpenGL side is extremely undemanding compared to a modern 3D game (just rendering to a single texture on a single quad), so I don't think we can ever take advantage of PBOs.

johnnovak:
PS: @kcgen we should mention the perf increase on lowend systems in the release notes.

Grounded0:
I don't want to shitpost on to PRs but only Intel iGPUs are junk. AMD and Apple makes good stuff.

kcgen:
Thanks for the detailed explanation, @johnnovak !

Fresh benchmarks of Quake 320x200 added with out-of-the-box setting (ie: `crt-auto`) to the body.

johnnovak:
> I don't want to shitpost on to PRs but only Intel iGPUs are shit. AMD and Apple makes good iGPUs.

I was wondering about that yeah. Thanks for the clarification 😄 But then, we have proven we can't improve anything by using PBOs with the current code anyway.

johnnovak:
Just a reference for myself and @GranMinigun if interested.

Duckstation optionally uses PBOs, but you must enable it conditionally depending on the driver vendor and/or the available extensions.

Reference:

https://github.com/stenzek/duckstation/blob/9cba1decad7aebf50cf058f8738640cdcc1f014c/src/util/opengl_device.cpp#L391

Also, we'd need to use a technique called buffer orphaning with PBOs to get the most performance out of them. I don't think our PBO path was doing that.

GranMinigun:
For future reference, DuckStation isn't compatible with any open-source licence anymore, so careful with reading its sources.

johnnovak:
> For future reference, DuckStation isn't compatible with any open-source licence anymore, so careful with reading its sources.

As you as you don't start copy pasting large blocks from it 1:1, you can read all you want 😄 Technically you could have the source open in one window, and reimplement it your way in a second window -- perfectly legal, it's your code.

PR #3102, opened 2023-11-07
Implement SCAS and CMPS opcodes for dynrec

kklobe:
# Description

This PR implements the SCASB/SCASW/SCASD and CMPSB/CMPSW/CMPSD opcodes natively in the dynamic core. These are common in many applications, especially `REPNZ SCASB`, which is often used to find the end of a null-terminated C string, and `REPZ CMPSB` to compare if two strings are equal.

Since these opcodes were not implemented in dynrec, that causes the normal core to spin up and decode, then go back to dynrec, so this should improve performance of these instructions.

# Manual testing

I have tested a handful of games, including Quake, Doom, PERFDOS, Blood, Black Cauldron, King's Quest 2, and Top Bench, forcing `core dynamic` where needed. 

I don't know that you'll visibly notice a performance improvement in most cases. There are some cases that hit it pretty hard, for example in Doom when you hit Quit and it's waiting for your input:

```log
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
2023-11-07 15:16:14.014 | dynrec_scasb_dword
```

Here is Blood, which runs at this pace (~150x per sec) continually:

```log
2023-11-07 18:21:08.314 | dynrec_cmpsb_dword
2023-11-07 18:21:08.326 | dynrec_cmpsb_dword
2023-11-07 18:21:08.354 | dynrec_scasb_dword
2023-11-07 18:21:08.354 | dynrec_scasb_dword
2023-11-07 18:21:08.354 | dynrec_scasb_dword
2023-11-07 18:21:08.354 | dynrec_scasb_dword
2023-11-07 18:21:08.355 | dynrec_cmpsb_dword
2023-11-07 18:21:08.369 | dynrec_cmpsb_dword
2023-11-07 18:21:08.369 | dynrec_cmpsb_dword
2023-11-07 18:21:08.369 | dynrec_cmpsb_dword
2023-11-07 18:21:08.369 | dynrec_cmpsb_dword
2023-11-07 18:21:08.369 | dynrec_cmpsb_dword
2023-11-07 18:21:08.397 | dynrec_scasb_dword
2023-11-07 18:21:08.397 | dynrec_scasb_dword
2023-11-07 18:21:08.397 | dynrec_scasb_dword
2023-11-07 18:21:08.398 | dynrec_scasb_dword
2023-11-07 18:21:08.398 | dynrec_cmpsb_dword
2023-11-07 18:21:08.412 | dynrec_cmpsb_dword
2023-11-07 18:21:08.442 | dynrec_scasb_dword
2023-11-07 18:21:08.442 | dynrec_scasb_dword
2023-11-07 18:21:08.442 | dynrec_scasb_dword
2023-11-07 18:21:08.442 | dynrec_scasb_dword
2023-11-07 18:21:08.442 | dynrec_cmpsb_dword
2023-11-07 18:21:08.454 | dynrec_cmpsb_dword
2023-11-07 18:21:08.482 | dynrec_scasb_dword

```

# Checklist

I have:

### DIFF
--- a/src/cpu/core_dynrec/operators.h
+++ b/src/cpu/core_dynrec/operators.h
@@ -1967,6 +1967,441 @@ static uint32_t DRC_CALL_CONV dynrec_stosd_dword(uint32_t count,int32_t add_inde
 	return count_left;
 }
 
+static uint16_t DRC_CALL_CONV dynrec_scasb_word(uint16_t count, int16_t add_index,
+                                                int32_t decode_rep,
+                                                PhysPt di_base) DRC_FC;
+static uint16_t DRC_CALL_CONV dynrec_scasb_word(uint16_t count, int16_t add_index,
+                                                int32_t decode_rep, PhysPt di_base)
+{
+	if (count == 0) {
+		return 0;
+	}
+
+	uint8_t val = 0;
+
+	while (count > 0) {
+		--count;
+		--CPU_Cycles;
+
+		val = mem_readb_inline(di_base + reg_di);
+
+		reg_di += add_index;
+
+		if ((decode_rep == REP_Z) != (val == reg_al)) {
+			break;
+		}
+	}
+
+	lf_var1b    = reg_al;
+	lf_var2b    = val;
+	lf_resb     = lf_var1b - lf_var2b;
+	lflags.type = t_CMPb;
+
+	return count;
+}
+
+static uint32_t DRC_CALL_CONV dynrec_scasb_dword(uint32_t count, int32_t add_index,
+                                                 int32_t decode_rep,
+                                                 PhysPt di_base) DRC_FC;
+static uint32_t DRC_CALL_CONV dynrec_scasb_dword(uint32_t count, int32_t add_index,
+                                                 int32_t decode_rep, PhysPt di_base)
+{
+	if (count == 0) {
+		return 0;
+	}
+
+	uint8_t val = 0;
+
+	while (count > 0) {
+		--count;
+		--CPU_Cycles;
+
+		val = mem_readb_inline(di_base + reg_edi);
+
+		reg_edi += add_index;
+
+		if ((decode_rep == REP_Z) != (val == reg_al)) {
+			break;
+		}
+	}
+
+	lf_var1b    = reg_al;
+	lf_var2b    = val;
+	lf_resb     = lf_var1b - lf_var2b;
+	lflags.type = t_CMPb;
+
+	return count;
+}
+
+static uint16_t DRC_CALL_CONV dynrec_scasw_word(uint16_t count, int16_t add_index,
+                                                int32_t decode_rep,
+                                                PhysPt di_base) DRC_FC;
+static uint16_t DRC_CALL_CONV dynrec_scasw_word(uint16_t count, int16_t add_index,
+                                                int32_t decode_rep, PhysPt di_base)
+{
+	if (count == 0) {
+		return 0;
+	}
+
+	add_index = left_shift_signed(add_index, 1);
+
+	uint16_t val = 0;
+
+	while (count > 0) {
+		--count;
+		--CPU_Cycles;
+
+		val = mem_readw_inline(di_base + reg_di);
+
+		reg_di += add_index;
+
+		if ((decode_rep == REP_Z) != (val == reg_ax)) {
+			break;
+		}
+	}
+
+	lf_var1w    = reg_ax;
+	lf_var2w    = val;
+	lf_resw     = lf_var1w - lf_var2w;
+	lflags.type = t_CMPw;
+
+	return count;
+}
+
+static uint32_t DRC_CALL_CONV dynrec_scasw_dword(uint32_t count, int32_t add_index,
+                                                 int32_t decode_rep,
+                                                 PhysPt di_base) DRC_FC;
+static uint32_t DRC_CALL_CONV dynrec_scasw_dword(uint32_t count, int32_t add_index,
+                                                 int32_t decode_rep, PhysPt di_base)
+{
+	if (count == 0) {
+		return 0;
+	}
+
+	add_index = left_shift_signed(add_index, 1);
+
+	uint16_t val = 0;
+
+	while (count > 0) {
+		--count;
+		--CPU_Cycles;
+
+		val = mem_readw_inline(di_base + reg_edi);
+
+		reg_edi += add_index;
+
+		if ((decode_rep == REP_Z) != (val == reg_ax)) {
+			break;
+		}
+	}
+
+	lf_var1w    = reg_ax;
+	lf_var2w    = val;
+	lf_resw     = lf_var1w - lf_var2w;
+	lflags.type = t_CMPw;
+
+	return count;
+}
+
+static uint16_t DRC_CALL_CONV dynrec_scasd_word(uint16_t count, int16_t add_index,
+                                                int32_t decode_rep,
+                                                PhysPt di_base) DRC_FC;
+static uint16_t DRC_CALL_CONV dynrec_scasd_word(uint16_t count, int16_t add_index,
+                                                int32_t decode_rep, PhysPt di_base)
+{
+	if (count == 0) {
+		return 0;
+	}
+
+	add_index = left_shift_signed(add_index, 2);
+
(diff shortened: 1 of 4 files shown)


src/cpu/core_dynrec/operators.h
```diff
@@ -1967,6 +1967,152 @@ static uint32_t DRC_CALL_CONV dynrec_stosd_dword(uint32_t count,int32_t add_inde
 	return count_left;
 }
 
+static uint16_t DRC_CALL_CONV dynrec_scasb_word(uint16_t count, int16_t add_index,
+                                                PhysPt di_base) DRC_FC;
+static uint16_t DRC_CALL_CONV dynrec_scasb_word(uint16_t count,
+                                                int16_t add_index, PhysPt di_base)
+{
+	for (; count > 0; count--) {
```
kcgen, line 1975: Maybe a while + an decrement could de-dupe the two `count--`s?


```c++
while(count > 0) {
    --count;

    ....

    if (lf_resb == 0) {
        break;
    }
}
```

src/cpu/core_dynrec/operators.h
```diff
@@ -1967,6 +1967,152 @@ static uint32_t DRC_CALL_CONV dynrec_stosd_dword(uint32_t count,int32_t add_inde
...
+		const auto addr = di_base + reg_di;
+		const auto val  = mem_readb(addr);
```
kcgen, line 1977: `addr` might be an `int` from the addition; the `mem_read` calls take in a `PhysPt` (`uint32_t`).  Maybe need a cast the addition?

src/cpu/core_dynrec/operators.h
```diff
@@ -1967,6 +1967,152 @@ static uint32_t DRC_CALL_CONV dynrec_stosd_dword(uint32_t count,int32_t add_inde
...
+
+		lf_var1b    = reg_al;
+		lf_var2b    = val;
+		lf_resb     = (uint8_t)(lf_var1b - lf_var2b);
```
kcgen, line 1981: Suggest a c++ cast.

src/cpu/core_dynrec/operators.h
```diff
@@ -1967,6 +1967,152 @@ static uint32_t DRC_CALL_CONV dynrec_stosd_dword(uint32_t count,int32_t add_inde
...
+		lflags.type = t_CMPb;
+
+		reg_di += add_index;
+		if (lf_resb == 0) {
+			count--;
+			break;
+		}
+	}
+
+	return count;
```
kcgen, line 1991: Just an idea; could `return count` in the `if` statement, and then `return 0` at the end because `count` should always be zero at that point.

kcgen (commented):
Very interesting addition, @kklobe ! Looks heavily used, too. 
Just a couple thoughts to consider :-)

src/cpu/core_dynrec/operators.h
kklobe: I myself am a bigger fan of single return statements in a function unless it causes major contortions.

kklobe:
@kcgen please hold on merging this, I'm seeing some things in testing I want to explore further

src/cpu/core_dynrec/operators.h
johnnovak: I'm generally the same, yeah. @kcgen has convinced me to prefer returning early vs nesting the whole function body, but in general I dislike multiple return statements too.

johnnovak:
Nice find @kklobe 🚀

kklobe:
> @kcgen please hold on merging this, I'm seeing some things in testing I want to explore further

Never mind, false alarm; carry on 😁

src/cpu/core_dynrec/operators.h
kcgen: yup; all good!
When we've got a waterfall of chunks of logic, the early returns can help flatten or act like a hard partition, which can be easy to juggle stuff in your head.
But yeah - in this case, it just collapsed down to a single return statement.

Was thinking the interesting aspect here might be to `return 0` at the end as a way to self-document/explain that zero is always expected if we get to the end of the function. (Or another way: the only time we return a non-zero `count` is when `lf_resb == 0`.

kcgen (approved):
All looks good, @kklobe.

kklobe:
> All looks good, @kklobe.

hey @kcgen, let's hold again; it seems adding CMPSx is going much quicker than I anticipated and I could squeeze that in this PR as well.

LowLevelMahn:
do you have a small assembler test to benchmark that in detail or only on game basis?

LowLevelMahn:
> > do you have a small assembler test to benchmark that in detail or only on game basis?
> > maybe you could create some sort of base benchmark based on this: https://github.com/lethal-guitar/EGA_benchmark/ its Borland C based with inline asm for the test-code - and there is already the timer implementation done for benchmarking
> 
> Great idea, and, yes, I am already putting together a Turbo C++ benchmark with inline asm, as you mention.

can you attach the benchmark tool, so others can use it for adding new commands to the dynamic core

johnnovak:
> can you attach the benchmark tool, so others can use it for adding new commands to the dynamic core

Yup it would be good to include the tool in `contrib`.

kklobe:
> > > do you have a small assembler test to benchmark that in detail or only on game basis?
> > > maybe you could create some sort of base benchmark based on this: https://github.com/lethal-guitar/EGA_benchmark/ its Borland C based with inline asm for the test-code - and there is already the timer implementation done for benchmarking
> > 
> > 
> > Great idea, and, yes, I am already putting together a Turbo C++ benchmark with inline asm, as you mention.
> 
> can you attach the benchmark tool, so others can use it for adding new commands to the dynamic core

It's not really any type of comprehensive "tool", more of a quick and dirty microbenchmark (which really doesn't prove much in the context of the dynamic core in a larger application), but I can include it in `contrib` as mentioned by @johnnovak

kklobe:
after quite a bit of micro benchmarking, it looks like there is no performance improvement from this approach. In the interest of not adding complexity for no benefit, I will close this out, even though it was very instructive for me - thank you all for your time reviewing and discussing this!

kcgen:
> after quite a bit of micro benchmarking, it looks like there is no performance improvement from this approach. In the interest of not adding complexity for no benefit, I will close this out, even though it was very instructive for me - thank you all for your time reviewing and discussing this!

The code is looking great, @kklobe. I can see you've even hoisted all of the repeat assignments outside of the function into a one-time step at the end.

If there's any value in benchmarks on other slower systems (Pi, x86 at 800 Hz, etc..) or if you can think of some whipsaw load patterns - just shout!

PR #3108, opened 2023-11-12
Draw the video frame in text row-sized chunks

kcgen:
# Description

When in text video modes, we were drawing video video frames with per-line timings.  Because DOSBox's own text console rasterizer just draws to the video memory as-is without doing fancy buffer-swapping or flipping, there could be instances where the updating frame "reveals" line-level differences during the frame update.

This PR adjusts the VGA draw part size to match the number of text rows (when in text mode), which granularizes the text updates at the row level to avoid revealing inter-row differences (at the pixel-line level).

Of course - differences can still occur mid-frame, but they will be at the text-row level.

Graphical modes are unaffected.

# Manual testing

Tested rapidly running `ls`, and also tested rapid text-based demos:

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/14202968-b326-491c-80a5-786a4b360c23

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/efa7e6d1-ea68-454d-a3a2-745b7e8ee576

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/af9a4413-16b8-4134-90e1-99ee8bebd0e6

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/e7f53fc6-45cd-40d6-a63d-dfd99769597f

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1425,7 +1425,21 @@ PixelFormat VGA_ActivateHardwareCursor()
 // A single point to set total drawn lines and update affected delay values
 static void setup_line_drawing_delays(const uint32_t total_lines)
 {
-	vga.draw.parts_total = total_lines > 480 ? 1 : total_lines;
+	if (INT10_IsTextMode(*CurMode)) {
+		// In text modes, draw the frame in row-sized chunks, but not
+		// finer. This granarlity avoids showing inter-row changes such
+		// as when the framebuffer is being scrolled or being changed at
+		// the time of vblank.
+		//
+		vga.draw.parts_total = INT10_GetTextRows();
+	} else {
+		// In graphical modes, either draw the entire frame in one-go
+		// for high resolutions, or per-line for low resolutions as
+		// games and demos using these modes were more likely to use
+		// per-line drawing tricks.
+		//
+		vga.draw.parts_total = total_lines > 480 ? 1 : total_lines;
+	}
 
 	vga.draw.delay.parts = vga.draw.delay.vdend / vga.draw.parts_total;
 


johnnovak:
I appreciate the effort @kcgen , but frankly, it was better before. Before this change only a single scanline was glitching, now it's an entire row, so it's worse. See attached video (just remove the second `.zip` extension).

If there's always gonna be glitching with the current DOSBox-side text output method, it's best to restrict the artifacts to a single scanline.

[glitch.avi.zip](https://github.com/dosbox-staging/dosbox-staging/files/13327310/glitch.avi.zip)

kcgen:
> Before this change only a single scanline was glitching, now it's an entire row, so it's worse. 

This was deliberate because the smallest "logical unit" in text modes is a text cell, so switching from per-line to per-row chunks (or "atomic row changes") means we'll only see row-level changes, and never slice through the middle of a text character (or strange effects like that).

But yeah - if we're just trying to shrink this region of change (as the frame is updated); then pixel-lines is the way to go.

> Cool text demos, but as we were saying, proper emulated DOS programs are different

I wasn't booted into real MS-DOS; so the text demo was making INT10 text calls into DOSBox's emulated DOS routines, and from there the drawn text is render to the VGA side where this per-line chunked behavior is seen.

(In the videos, you will also see there's no tearing or sub-row slicing; and the demo is using a real text mode).

But regardless, I'd rather the per-line updates anyway :sweat_smile:

johnnovak:
> I wasn't booted into real MS-DOS; so the text demo was making INT10 text calls into DOSBox's emulated DOS routines, and from there the drawn text is render to the VGA side where this per-line chunked behavior is seen.
> 
> (In the videos, you will also see there's no tearing or sub-row slicing; and the demo is using a real text mode).

Ok, but do we get any slicing/tearing artifacts with these text demos prior to this change, when rendering by scanline? I think not, we established that the probable cause for text tearing at the DOS prompt is that our emulated DOS prompt implementation does "weird things" and doesn't quite go through the normal "emulation pathways" like regular DOS programs (not our commands) do.

Can you post the names of these demos, links to them, or the demos themselves? Then I can also test if there's any tearing when rendering by line as before.

> But regardless, I'd rather the per-line updates anyway 😅

Let's take a step back 😄 I brought this up because the _single scanline_ glitches annoyed me in our DOS prompt _only_. I never encountered these glitches on any other text programs, just with our DOS prompt.

So, if running _actual DOS programs_ that use text mode never glitch anyway with per scanline rendering (to be tested if you send links to the demos), then this change only _magnifies_ the problem with the DOS prompt. So yeah, then it's worse because it's more noticeable now and more distracting 😄

kcgen:
Before I close it out, I did a deeper dive into what's happening as the frame level.

Notice that `DIR` isn't affected by this issue despite both it and `LS`just printing text using  `WriteOut(char*)`, so what's going on?

## `DIR`, stepping per-millisecond

1. `DIR` step 1

    

2. After tapping `Enter`, `DIR` prints all of its content within the same millisecond moment in time:

    

    There are no intermediate steps.

---

Now let's see how `LS` behaves:

1. `LS` step 1, just before pressing `Enter`

    

1. Ater pressing `Enter`, `LS` prints an initial new line causing the entire frame to scroll up one row:

    

1. There's still more time spent here, `LS`, the cursor at the bottom is just starting to blink in. We're now several milliseconds in:

    

1. Finally `LS` prints all its content, causing the frame to scroll up even more rows:

    

So the big difference is that `LS` has three visible drawing events, all of which can drive frame updates. These seems to appear (nearly) instantly, but it's really doing things in proper sequence: scrolling content up and then re-drawing in the new scrolled stated.

kcgen:
> Can you post the names of these demos, links to them, or the demos themselves? Then I can also test if there's any tearing when rendering by line as before.

Yup - that one is `textro`

> So, if running actual DOS programs that use text mode never glitch anyway with per scanline rendering (to be tested if you send links to the demos), then this change only magnifies the problem with the DOS prompt.

I'm still digging; what I've concluded so far is that the our DOSBox-side commands (like `LS` and `DIR`) are able to use `WriteOut(char*)`, which is a DOSBox API that interfaces with the real text frame content.

Basically, we can call `WriteOut("hello world")` and instantly inject text right into the DOS console; and this is very fast and independent of CPU cycle count.

Compare that with DOS-side programs - if they want to print text, they use INT10H interrupt, which DOSBox emulates.

Both text pathways ultimately terminate in the same BIOS text frame handling code.

> Am I missing some benefits of the per-row rendering here?

I don't think so!

johnnovak:
Couple of ideas:

- Don't instantly inject text into the DOS console, just go through the normal channels as other DOS programs (i.e., BIOS interrupts, or write characters to the raw video memory the same way actual programs do)
- Just render a single "slice" (part) for text modes. Or would that break any weird text mode demos? I don't think so, but would need testing.

The first option seems better to me.

Issue #3104, opened 2023-11-09
Ultima VIII – Pagan keyboard regression


johnnovak:
Paging @FeralChild64 😆

Commit e73aaa57436bf5ef4eeea16255c78bef8511787e ("Rework keyboard emulation") has introduced a keyboard input regression in **Ultima VIII – Pagan**. I used the eXoDOS v5 version for testing.

### Steps to reproduce

1. Start the game (I just used option 1 (SoundBlaster) with the eXoDOS launcher, but it doesn't matter).
2. Press Esc to skip the intro.
3. Try to type something in at the "Give thy name" prompt — you can't.

FeralChild64:
The game issues command 0xAD to the Intel 8042 controller, which disables keyboard input. But I don't see the game doing anything which should re-enable it. It seems I missed something in the reworked emulation. Commenting out the 

```
is_disabled_kbd = true;
```

line in the `intel8042.cpp` "fixes" the problem, but it is definitely not a proper fix. Investigating...

johnnovak:
Thank for the quick reply @FeralChild64.

> The game issues command 0xAD to the Intel 8042 controller, which disables keyboard input.

Channeling Toni Wilen (WinUAE maintainer): never code anything in emulators relying on specifications. Almost always there is a lot of undocumented behaviour 😏 Good luck, you'll probably need to test on hardware at some point and/or check out what other emulators do.

johnnovak:
> [edit] Before my rework disabling the keyboard did not really do anything, that's why it was working. Strange.

This issue got me interested, I can test the game on real hardware and can run test programs if that helps @FeralChild64. Just send me some test binaries I can simply run.

johnnovak:
You might be already aware of the information from this article, but here we go:
http://www.os2museum.com/wp/ibm-pcat-8042-keyboard-controller-commands/

Some interesting tidbits below, but make sure to read the whole thing (if you haven't already). The last one is the most interesting and maybe that's the answer to the mystery. Like I said, manufacturer's specs always lie 😎 Test programs or I don't believe it 😆 

> IBM documented a number of commands the host can send to the KBC. It should be understood that all those commands are a pure software construct, with nothing about the 8042 hardware dictating that the commands need to follow any specific format, function, or that they even need to be there at all. Therefore understanding the 8042 ROM code is the only way towards understanding exactly what the commands are and what they do, with the caveat that different controllers may and do have somewhat different code in their ROM.

[...]

PR #3109, opened 2023-11-12
Workaround for keyboard problem in Ultima VIII: Pagan

FeralChild64:
# Description

This is a workaround for _Ultima VIII: Pagan_ keyboard problem; it introduces timeout for disabling the keyboard (commands to disable keyboard used to have almost no effect before the rework which introduced the regression).

A proper fix would be too risky for now, as it would involve:
- firing keyboard interrupt for a key press/release even if keyboard is disabled (easy) - this is something the new 8042 emulation missed
- improving our BIOS to work with keyboard most of the time disabled with 0xad command (risky, might easily introduce regressions if not done properly) 

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3104

# Manual testing

1. Start Ultima VIII: Pagan game, press Esc to skip intro, type player name - should work noe.
2. Start Windows 3.11 for Workgroups, quit it - keyboard should still be operable (regression test).

# Checklist

I have:

### DIFF
--- a/src/hardware/input/intel8042.cpp
+++ b/src/hardware/input/intel8042.cpp
@@ -372,6 +372,63 @@ static uint8_t get_translated(const uint8_t byte)
 }
 
 // ***************************************************************************
+// Port disable workaround
+// ***************************************************************************
+
+// TODO: This part is a workaround for 'Ultima VIII: Pagan' problem with
+// non-working keyboard. The proper fix would require larger rework:
+// - it seems that even if keyboard is disabled, the 8042 should fire the
+//   interrupt
+// - testing with 86Box revealed that command 0xad (disable keyboard port) and
+//   0xae (enable port) are being sent by BIOS with each and every key press;
+//   our internal BIOS does not work this way
+//
+// Ignoring command 0xad totally might be risky, as Windows 3.11 for Workgroups
+// disables/reenables the keyboard port during startup and shutdown, issuing
+// controller commands in between; if something disturbs it, we might run into
+// a problem with keyboard/mouse not responding at all.
+//
+// This code re-enables the keyboard after certain time - unless there is
+// some command currently being issued to the controller.
+
+static bool kbd_disabled_timer_running = false;
+static bool kbd_disabled_timer_expired = true;
+
+// Make the delay a little longer than between two bytes of a scancode,
+// even lower values are enough
+constexpr double kbd_disabled_duration_ms = port_delay_ms * 1.5f;
+
+// Forward declaration
+static void restart_kbd_disabled_timer();
+
+static void maybe_reenable_kbd_port()
+{
+	if (is_disabled_kbd && kbd_disabled_timer_expired) {
+		is_disabled_kbd = false;
+	}
+}
+
+static void kbd_disabled_timer_handler(uint32_t /*val*/)
+{
+	kbd_disabled_timer_running = false;
+	kbd_disabled_timer_expired = true;
+
+	if (current_command != Command::None) {
+		restart_kbd_disabled_timer();
+	}
+}
+
+static void restart_kbd_disabled_timer()
+{
+	if (kbd_disabled_timer_running) {
+		PIC_RemoveEvents(kbd_disabled_timer_handler);
+	}
+	PIC_AddEvent(kbd_disabled_timer_handler, kbd_disabled_duration_ms);
+	kbd_disabled_timer_running = true;
+	kbd_disabled_timer_expired = false;
+}
+
+// ***************************************************************************
 // Controller buffer support
 // ***************************************************************************
 
@@ -713,6 +770,7 @@ static void execute_command(const Command command)
 		is_disabled_kbd      = true;
 		uses_kbd_translation = true;
 		passed_self_test     = true;
+		restart_kbd_disabled_timer();
 		flush_buffer();
 		buffer_add(0x55);
 		break;
@@ -721,6 +779,7 @@ static void execute_command(const Command command)
 		// (as with aux port test)
 		// Disables the keyboard port
 		is_disabled_kbd = true;
+		restart_kbd_disabled_timer();
 		flush_buffer();
 		buffer_add(0x00); // as with TestPortAux
 		break;
@@ -746,6 +805,7 @@ static void execute_command(const Command command)
 		// Disable keyboard port; any keyboard command
 		// reenables the port
 		is_disabled_kbd = true;
+		restart_kbd_disabled_timer();
 		break;
 	case Command::EnablePortKbd: // 0xae
 		// Enable the keyboard port
@@ -966,6 +1026,9 @@ static void write_data_port(io_port_t, io_val_t value, io_width_t) // port 0x60
 		// A controller command is waiting for a parameter
 		const auto command = current_command;
 		current_command    = Command::None;
+		if (is_disabled_kbd) {
+			restart_kbd_disabled_timer();
+		}
 
 		const bool should_notify_aux = !I8042_IsReadyForAuxFrame();
 		const bool should_notify_kbd = !I8042_IsReadyForKbdFrame();
@@ -1009,6 +1072,9 @@ static void write_command_port(io_port_t, io_val_t value, io_width_t) // port 0x
 	status_byte.was_last_write_cmd = true;
 
 	current_command = Command::None;
+	if (is_disabled_kbd) {
+		restart_kbd_disabled_timer();
+	}
 	if ((byte <= 0x1f) || (byte >= 0x40 && byte <= 0x5f)) {
 		// AMI BIOS systems command aliases
 		execute_command(static_cast<Command>(byte + 0x20));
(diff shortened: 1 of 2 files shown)


src/hardware/input/intel8042.cpp
```diff
@@ -371,6 +371,61 @@ static uint8_t get_translated(const uint8_t byte)
...
+// controller commands in between; if something disturbs it, we might run into
+// a problem with keyboard/mouse not responding at all.
+//
+// This code re-enables the keyboard after certain time - unless there is
+// some command currently being issued to the controller.
+
+static bool kbd_disabled_timer_running = false;
+static bool kbd_disabled_timer_expired = true;
+
+constexpr double kbd_disabled_duration_ms = port_delay_ms * 2;
```
johnnovak, line 397: UpperCamelCaseConstants

src/hardware/input/intel8042.cpp
```diff
@@ -371,6 +371,61 @@ static uint8_t get_translated(const uint8_t byte)
...
+//   0xae (enable port) are being sent by BIOS with each and every key press;
+//   our internal BIOS does not work this way
+//
+// Ignoring command 0xad totally might be risky, as Windows 3.11 for Workgroups
+// disables/reenables the keyboard port during startup and shutdown, issuing
+// controller commands in between; if something disturbs it, we might run into
+// a problem with keyboard/mouse not responding at all.
+//
+// This code re-enables the keyboard after certain time - unless there is
+// some command currently being issued to the controller.
```
johnnovak, line 392: Nice explanatory comment 👍🏻

src/hardware/input/intel8042.cpp
FeralChild64: Updated in a separate commit, cleans up many constant names.

src/hardware/input/intel8042.cpp
```diff
@@ -46,16 +46,16 @@ CHECK_NARROWING();
 // - http://www.os2museum.com/wp/ibm-pcat-8042-keyboard-controller-commands/
 // - http://www.os2museum.com/wp/ibm-ps2-model-50-keyboard-controller/
 
-static constexpr uint8_t irq_num_kbd_pcjr  = 6;
-static constexpr uint8_t irq_num_kbd_ibmpc = 1;
-static constexpr uint8_t irq_num_mouse     = 12;
+static constexpr uint8_t IrqNumKbdPCJr  = 6;
```
johnnovak, line 49: Nitpicking: Pcjr

src/hardware/input/intel8042.cpp
```diff
@@ -46,16 +46,16 @@ CHECK_NARROWING();
...
+static constexpr uint8_t IrqNumKbdIbmPC = 1;
```
johnnovak, line 50: Nitpicking: IbmPc

src/hardware/input/intel8042.cpp
```diff
@@ -46,16 +46,16 @@ CHECK_NARROWING();
...
-constexpr uint8_t        firmware_revision  = 0x00;
-static const std::string firmware_copyright = DOSBOX_COPYRIGHT;
+constexpr uint8_t        FirmwareRevision  = 0x00;
+static const std::string FirmwareCopyright = DOSBOX_COPYRIGHT;
 
-static constexpr uint8_t buffer_size = 64; // in bytes
+static constexpr uint8_t BufferSize = 64; // in bytes
 // delay appropriate for 20-30 kHz serial clock and 11 bits/byte
-static constexpr double port_delay_ms = 0.300;
+static constexpr double PortDelay_ms = 0.300;
```
johnnovak, line 58: PortDelayMs

src/hardware/input/intel8042.cpp
```diff
@@ -423,7 +423,7 @@ static void restart_kbd_disabled_timer()
 	if (kbd_disabled_timer_running) {
 		PIC_RemoveEvents(kbd_disabled_timer_handler);
 	}
-	PIC_AddEvent(kbd_disabled_timer_handler, kbd_disabled_duration_ms);
+	PIC_AddEvent(kbd_disabled_timer_handler, KbdDisabledDuration_ms);
```
johnnovak, line 426: I don't think we should mix camel and underscore cases like this. Existing example from `mixer.cpp`, I think this should be the style:

```
constexpr auto EnvelopeMaxExpansionOverMs = 15u;
constexpr auto MaxPrebufferMs = 100;
```

src/hardware/input/intel8042.cpp
johnnovak: If you remember, we had similar discussions before and decided not to mimic "marketing capitalisation", and thus just went with `VmWare`.

src/hardware/input/intel8042.cpp
```diff
@@ -371,6 +371,63 @@ static uint8_t get_translated(const uint8_t byte)
...
+// - it seems that even if keyboard is disabled, the 8042 should fire the
+//   interrupt
+// - testing with 86Box revealed that command 0xad (disable keyboard port) and
+//   0xae (enable port) are being sent by BIOS with each and every key press;
+//   our internal BIOS does not work this way
+//
+// Ignoring command 0xad totally might be risky, as Windows 3.11 for Workgroups
+// disables/reenables the keyboard port during startup and shutdown, issuing
+// controller commands in between; if something disturbs it, we might run into
+// a problem with keyboard/mouse not responding at all.
```
kcgen, line 389: just some naive comments / questions:

- Does original DOSBox just ignores this (and says _"Yes, it's disabled"_) but really does nothing?

- Are there benefits to really honoring the "shutdown"? Maybe the keyboard timers aren't updated when shutdown? Or some other benefits, like clearing "held" states (like capslock/numlock)?

I like that you're getting closer and closer to how the hardware behaved!

src/hardware/input/intel8042.cpp
```diff
@@ -371,6 +371,63 @@ static uint8_t get_translated(const uint8_t byte)
...
+//
+// This code re-enables the keyboard after certain time - unless there is
+// some command currently being issued to the controller.
```
johnnovak, line 392: Can you think of any cases where this re-enabling workaround would introduce unwanted behaviour? Is there anything we can help regression test?

kcgen (approved):

src/hardware/input/intel8042.cpp
FeralChild64: Ok, changed.

src/hardware/input/intel8042.cpp
FeralChild64: OK, changed.

src/hardware/input/intel8042.cpp
FeralChild64: OK, changed.

src/hardware/input/intel8042.cpp
FeralChild64: I can only think about starting/quitting different versions of Windows (from 1.0 to 9x) and see whether keyboard is still working.

src/hardware/input/intel8042.cpp
FeralChild64: > Does original DOSBox just ignores this (and says "Yes, it's disabled") but really does nothing?

Basically, yes. Search for `keyb.active` in their [keyboard.cpp](https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/src/hardware/keyboard.cpp) file.

> Are there benefits to really honoring the shutdown? Maybe the keyboard timers aren't updated when shutdown? Or clearing "held" states (like capslock/numlock)?

This is not exactly a shutdown, it does not directly affect the microcontroller in the keyboard - it just prevents the scancodes from being sent from the keyboard to the 8042 PS/2 controller; they are not even discarded, they are waiting in a queue inside the keyboard.
This way the guest software can issue commands to the mouse or to the controller without worrying that suddenly a key scancode appears on the input. Windows tends to do this during startup/shutdown.
If we ignore the disable request, there is a very small (but real) risk, that key pressed/released in the wrong moment disrupts whatever the guest software is doing.

johnnovak (approved):
Tested with Ultima VIII, Win 3.11 for Workgroups, and a couple of DOS programs — all work fine and couldn't find regressions. Nice quick fix @FeralChild64 🎉 

I noticed some performance degradation when setting the key repeat rate to "fast" with tools like `TURBOKEY`; on main with `TURBOKEY` and `cycles = 50000` I get no performance throttling behaviour (this manifest in the key repeat becoming much slower and jerky), but with this patch I can only go up to about `cycles = 30000` before throttling happens. So yeah, this degrades performance somewhat with fast key repeat settings, but release builds are fine (both main and this patch starts throttling somewhat around 130k cycles in QPV with `TURBOKEY`, but it's still quite usable at 200k. Heavy throttling starts at 300k cycles on both branches).

In any case, I think this is a fine workaround until the issues is addressed properly at the BIOS level.

src/hardware/input/intel8042.cpp
kcgen: > This way the guest software can issue commands to the mouse or to the controller without worrying that suddenly a key scancode appears on the input. Windows tends to do this during startup/shutdown.

Thanks for that explanation, @FeralChild64 - makes sense!

PR #3111, opened 2023-11-12
Clean up OpenGL attribute init

kklobe:
# Description

This is a pedantic cleanup of the `SDL_GL_SetAttribute()` calls in `sdlmain.cpp`. The [SDL docs](https://wiki.libsdl.org/SDL2/SDL_GLattr) specify that these should be set prior to `SDL_CreateWindow`:

```
While you can set most OpenGL attributes normally, the attributes listed above must be 
known before SDL creates the window that will be used with the OpenGL context.
```

The `SDL_GL_FRAMEBUFFER_SRGB_CAPABLE` attribute requires a bit of hack. The driver vendor cannot be queried without a context, so I create a temporary hidden OpenGL window and context, query the driver, then destroy them.

Also, `SDL_GL_DOUBLEBUFFER` is enabled by default, so I got rid of the call to set that, but, inspired by a Discord comment from @johnnovak, I query the value and display a warning if it's not enabled.

Things seem to be working fine as-is, but given the wide range of quality of OpenGL drivers, let's do it by the book.

# Manual testing

I ran the artifact builds for macOS, Windows, and Linux (via WSL2), verifying `output = opengl` and testing Quake, Doom, King's Quest 3, Future Crew's Panic, Unreal, and Second Reality.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -211,6 +211,8 @@ static bool present_frame_texture();
 #if C_OPENGL
 static void update_frame_gl(const uint16_t *changedLines);
 static bool present_frame_gl();
+static const char* safe_gl_get_string(const GLenum requested_name,
+                                      const char* default_result);
 #endif
 
 static const char* vsync_state_as_string(const VsyncState state)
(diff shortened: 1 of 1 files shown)


src/gui/sdlmain.cpp
```diff
@@ -1244,6 +1246,42 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 #if C_OPENGL
 		if (rendering_backend == RenderingBackend::OpenGl) {
 			flags |= SDL_WINDOW_OPENGL;
+
+			// We need a context to query the vendor string.
+			const auto tempWindow = SDL_CreateWindow(
```
johnnovak, line 1251: minor: temp_window

src/gui/sdlmain.cpp
```diff
@@ -1244,6 +1246,42 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
...
+
+#if WIN32
+			const auto is_vendors_srgb_unreliable = (gl_vendor == "Intel");
+#else
+			constexpr auto is_vendors_srgb_unreliable = false;
+#endif
+			if (is_vendors_srgb_unreliable) {
+				LOG_WARNING("SDL:OPENGL: Not requesting an sRGB framebuffer"
+				            " because %s's driver is unreliable",
+				            gl_vendor.data());
```
johnnovak, line 1279: Very nice 🏅

src/gui/sdlmain.cpp
```diff
@@ -1950,6 +1971,17 @@ uint8_t GFX_SetSize(const int width, const int height,
...
+		int is_double_buffering_enabled = 0;
+		if (SDL_GL_GetAttribute(SDL_GL_DOUBLEBUFFER,
+		                        &is_double_buffering_enabled)) {
+			LOG_WARNING("OPENGL: Failed getting double buffering status: %s",
+			            SDL_GetError());
+		} else {
+			if (!is_double_buffering_enabled) {
+				LOG_WARNING("OPENGL: Double buffering not enabled");
+			}
+		}
```
johnnovak, line 1983: 💯

johnnovak (approved):
Very nice clean up @kklobe. 🎉  One _extremely minor_ comment 😅

kklobe:
Thanks for the review @johnnovak - was it you who originally found the Intel sRGB issues and worked around them? I was just wondering if we could re-test that scenario and see if this change makes any difference.

johnnovak:
> Thanks for the review @johnnovak - was it you who originally found the Intel sRGB issues and worked around them? I was just wondering if we could re-test that scenario and see if this change makes any difference.

Sure I'll retest today or tomorrow.

I introduced sRGB texture & framebuffer support (which was broken for a year and nobody realised, it basically never worked in the stable release LOL), then I think @GranMinigun added the Intel-related checks.

Currently, shaders can request sRGB textures & framebuffers via pragmas (and now it works 😅), e.g. this is from the `sharp` shader:

image

johnnovak:
As discussed @kklobe, this should be merged after this goes in:
https://github.com/dosbox-staging/dosbox-staging/pull/3124

Will test after the rebase to my merged PR.

johnnovak:
@kklobe My SDL refactorings have been merged, which means your time has come 😎

src/gui/sdlmain.cpp
```diff
@@ -1307,6 +1309,40 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 		if (rendering_backend == RenderingBackend::OpenGl) {
 			flags |= SDL_WINDOW_OPENGL;
 		}
+
+		// We need a context to query the vendor string.
+		const auto temp_window = SDL_CreateWindow(
+		        "", 0, 0, 1, 1, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
```
johnnovak, line 1315: We're not showing the temp window anyway, so I think it would be safer to use a realistic window size, like 200 by 200. I can imagine requesting an 1x1 window to perhaphs fail on some OS variants / Window managers.

johnnovak:
I will test later today @kklobe

PR #3113, opened 2023-11-13
Perform console cursor and text calls via interrupt

kcgen:
# Description

Picking up from the end of the discussion [here](https://github.com/dosbox-staging/dosbox-staging/pull/3108#issuecomment-1807286992), this PR adjusts the the emulated console (the `CON` device) to no longer immediately inject text into the DOS console and instead uses the INT10 interrupt callback just as real DOS programs do.

The calls used are imported from DOSBox Daum, adapts to avoid duplicate code using `constexpr` toggle in the INT10 routines.

# Manual testing

Tested debug and sanitizer builds with various console programs at various speeds (XT in low-hundreds, early AT in 500 to 700, and faster 386 and 486 at 3000+ cycles).

### Slow XT Class (< 100 cycles)

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/f0df1c0f-44a8-4940-9747-081670e61638

### Slow AT Class through 386 and 486 (cycles >= 100 through 10000)

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/09c6dc50-a2a4-477d-ac0b-c51672476368

# Checklist

I have:


johnnovak:
Nice one @kcgen.

I tested how the various commands behave on my Pentium MMX 200 downclocked to 125Mhz (or 100 Mhz, can't remember), and the scrolling is much slower on real hardware in general. I need to set the following cycles values to get approximately the same "scrolling speed" when executing various commands:

- `DIR` –  `cycles 200`
- `MEM` –  `cycles 120`
- `COPY /?` –  `cycles 120`
- `TYPE <somefile` – `cycles 40` (yes 😅)

Interesting, the scrolling speed doesn't change on real hardware even if I disable all caches in the BIOS and use `SETMUL` to slow things down as much as possible to slow 386 speeds.

Running real MS-DOS inside Staging doesn't fare much better; the scrolling is a little bit slower in general, but still waaaaay fast compared to even my Pentium MMX box. This indicates that achieving the same text output rate is not really possible because of general emulation inaccuracies in DOSBox.

In any case, this is a step in the right direction and it brings us closer to the RealThing(tm) a _tiny_ bit, but it's still waaaay faster.

Does it matter? Not really, I think 😄 And maybe watching the help text of the various command visibly scroll is a hard sell for many, but yeah, that's what happens on a real PC.

@kklobe might chime in as well 😄 I've recorded a few videos, but there's not much point in sharing them. You can accurate replicate the real MS-DOS behaviour with the above low cycles settings from 40 to 200.

dosbox.conf
kklobe, line 1: Is this file intended to be checked in a merged?

kklobe:
This is much better - even though it's not "authentic" scrolling speed (didn't real DOS utilities and COMMAND.COM use INT 21 calls like 02 and 09 for writing strings? Maybe that incurred additional overhead?). On `main` if I do a `cls` then repeated `dir`, the screen doesn't change at all, it's like an image viewer. With this change, I at least see activity, so I give this a 🥇.

weirddan455 (approved):
Code looks good as far as I can tell.

The scrolling speed on a real 8088 is _much_ slower than that first video and your 386-486 video looks more like a fast Pentium.  Personally, I don't miss waiting 10 seconds for a `DIR` command to finish scrolling but you guys do you :stuck_out_tongue:

kklobe:
> Code looks good as far as I can tell.
> 
> The scrolling speed on a real 8088 is _much_ slower than that first video and your 386-486 video looks more like a fast Pentium. Personally, I don't miss waiting 10 seconds for a `DIR` command to finish scrolling but you guys do you 😛

:grin: indeed! If I really, really get nostalgic and just have to have that experience, I can always jump over to https://www.pcjs.org for a quick hit.

FeralChild64:
On my system scrolling speed is still very fast, unless I go very low with cycles. Even if we issue a real interrupt, most of the DOS/BIOS processing is done by the native code (with native speed), not emulated one.

But the whole processing (using the actual interrupts) is now definitely more realistic.

kcgen:
The next step would be emulate video-card memory read and write latency, like DOSBox-X's configurable `vmemdelay` delay that touches `src/hardwrae/vga_memory.cpp`.

I have a feeling DOSBox didn't emulate this because there was never a XT or AT "specification" that bound video memory to some threshold. 

The delay was always seen as undesirable: people writing games wanted faster video handling, and people coding 'printf' in C wanted content printed as fast as possible. Plus.. video card manufacturers constantly pursued faster chips (so the landscape was never fixed).

Game devs learned to draw on vblank.. which works with slow or really fast video cards, and the few that didn't either tear or need to be throttled at the CPU level to pace their game logic. 

I scanned all of eXoDOS's conf files (7300+ games), and none are using DOSBox-X's `vmemdelay` setting, which is at least is some comfort that we may not need this.

Would be very interesting to know if there are any games that rely sole on video card latency to "throttle" their game speed; that if latency goes down, the game breaks or runs too fast.

Will merge this for now - but maybe `vmemdelay` still makes sense, just for a more nostalgic feel on the CLI :-)

kcgen:
> Interestingly, the scrolling speed doesn't change on real hardware much even if I disable all caches in the BIOS and use SETMUL to slow things down as much as possible to slow 386 speeds.

Yup, the video card is bottleneck'ing the display rate, even when the CPU is throttled down.

dosbox.conf
kcgen: was my local tester; accidentally hoovered up w/ a `git commit -a` :-)

thanks @kklobe (snipping this)

rderooy:
The only game that I could quickly find that seems to benefit from `vmemdelay` is Super Zaxxon
https://github.com/joncampbell123/dosbox-x/issues/4197#issuecomment-1520990989

johnnovak:
I played around with `vmemdelay` in DOSBox-X because I hoped it would help with the Dungeon Keeper map menu and Quest for Glory II intro scroll speed issues. It had absolutely zero effect.

I'm really curious where these extra latencies are coming from that slow down the scrolling in DK and QfG. I think I'll create issue tickets with video recordings so we at least keep track of them.

In QfG II in the intro sequence sometimes there is full screen scrolling between the ground and the sky. On my _fast_ MMX box you can see the scrolling, but in DOSBox you can't see it it's so fast, unless you use comically low cycles settings that make the game unplayable. 

Haven't tried it PCem yet, but that might reveal a few things. Maybe these games are calling some BIOS routines that do specific things that our BIOS emulation don't emulate or too quickly. Well, you can run actual BIOS images in X: https://github.com/joncampbell123/dosbox-x/issues/193#issuecomment-417772231

But then it might be some quirks of the VGA/VESA BIOS, or maybe something completely different 🤷🏻‍♂️ 

It's worthwhile to note that I'm running CompactFlash cards in this MMX machine of mine with CF to IDE adapters. So disk I/O bottlenecks are effectively ruled out at the _drive_ level (but there's the IDE controllers, the BIOS, and the DIR command itself).

Anyway, accurate emulation is a rabbit whole of its own. Approving the PR in the meantime 😄

kklobe:
> I'm really curious where these extra latencies are coming from that slow down the scrolling in DK and QfG. I think I'll create issue tickets with video recordings so we at least keep track of them.

Sounds like a great excuse to play with Tracy. If someone can make a repro pack I'd be willing to take a chop at it.

src/ints/int10_char.cpp
```diff
@@ -612,7 +636,48 @@ void WriteChar(uint16_t col,uint16_t row,uint8_t page,uint8_t chr,uint8_t attr,b
...
-void INT10_WriteChar(uint8_t chr, uint8_t attr, uint8_t page, uint16_t count, bool showattr)
+static void write_char_via_interrupt(const uint8_t cur_col, const uint8_t cur_row,
+                                     const uint8_t page, const uint8_t char_value,
+                                     const uint8_t attribute, const bool use_attribute)
+{
+	constexpr uint8_t write_char_cmd        = 0x1;
+	constexpr uint8_t with_attribute_cmd    = 0x9;
+	constexpr uint8_t without_attribute_cmd = 0x0A;
+
+	// Position the cusror
```
johnnovak, line 647: cursor

johnnovak (approved):
Nice enhancement @kcgen 🥳

johnnovak:
> > I'm really curious where these extra latencies are coming from that slow down the scrolling in DK and QfG. I think I'll create issue tickets with video recordings so we at least keep track of them.
> 
> Sounds like a great excuse to play with Tracy. If someone can make a repro pack I'd be willing to take a chop at it.

Will do sometime this week and will tag you. Curious what you'll find out!

kcgen:
Thanks for the tests and reviews, @johnnovak, @FeralChild64, @kklobe, @weirddan455, and @rderooy. 

What started out as something we thought was a minor glitch turned out to reveal that DOSBox's emulation is realistic and robust (despite being unthrottled at the ISA video card level).. and that this might lead to introducing more video-side latency.

Issue #3114, opened 2023-11-13
Black screen in fullscreen mode on macOS Catalina


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-alpha

### What Operating System are you using?

macOS (Intel)

### If Other OS, please describe

Catalina 10.15.7 (latest)

### Relevant hardware info

Mac mini (End 2012)

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When using the current DEV build on macOS Catalina, the screen goes black when switching to fullscreen mode.

When using 'glshader = none', everything works as expected.

Config: Default.

Staging v0.80.1 works as expected.

Main windows after launch using 'glshader = crt-auto' (default).
![Bildschirmfoto 2023-11-10 um 01 17 16](https://github.com/dosbox-staging/dosbox-staging/assets/61995128/3d97c2d8-a21b-4b4f-ac47-6ce858a8013a)

Hardware info:
2,5GHz Dual-Core Intel Core i5
Intel HD Graphics 4000 1536 MB

### Steps to reproduce the behaviour.

Explain how to reproduce
1. 
2.
3.

### Your configuration

```ini
Default config
```

### Provide a Log

```text
Log output after launch and switching to fullscreen, then switching back to windowed mode:

Last login: Sun Nov 12 22:09:52 on ttys000
/Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox ; exit;
alex@Alexanders-Mac-mini ~ % /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox ; exit;
date       time         | 
2023-11-12 22:10:23.725 | arguments: /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox
2023-11-12 22:10:23.725 | Current dir: /Users/alex
2023-11-12 22:10:23.725 | stderr verbosity: 0
2023-11-12 22:10:23.725 | -----------------------------------
2023-11-12 22:10:23.725 | dosbox-staging version 0.81.0-alpha-1918-g40fb9b3dca
2023-11-12 22:10:23.725 | ---
2023-11-12 22:10:23.725 | LOG: Loguru version 2.1.0 initialised
2023-11-12 22:10:23.728 | CONFIG: Loaded primary config file '/Users/alex/Library/Preferences/DOSBox/dosbox-staging.conf'
[...]

PR #3121, opened 2023-11-14
Always log OpenGL shader and program info logs if they're non-empty

johnnovak:
# Description

After compiling an OpenGL shader and linking a shader program, the vendor-specific OpenGL info logs might contain warnings and helpful informational messages even if the compilation/linking status was successful, so we'll always log them if they're non-empty.

This might or might not help diagnose shader issues (like this one https://github.com/dosbox-staging/dosbox-staging/issues/3114), but it's good practice to do it anyway.

For the record, these info logs stay empty with all the new adaptive CRT shaders on my MacBook M1 Pro with an integrated GPU.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3114

# Manual testing

Started up Staging with the adaptive CRT shaders. Nothing extra got logged and the shaders were working fine with various window sizes and in fullscreen.

Trace logged that the info log sizes are zero, so the no extra logging checks out.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -925,7 +925,7 @@ static void set_vsync(const VsyncState state)
 									: "0";
 	if (SDL_SetHint(SDL_HINT_RENDER_VSYNC, hint_str) == SDL_TRUE)
 		return;
-	LOG_WARNING("SDL: Failed setting SDL's vsync state to to %s (%s): %s",
+	LOG_WARNING("SDL: Failed setting vsync state to to %s (%s): %s",
 				vsync_state_as_string(state), hint_str, SDL_GetError());
 }
 
@@ -1325,7 +1325,10 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 		displayMode.h = height;
 
 		if (SDL_SetWindowDisplayMode(sdl.window, &displayMode) != 0) {
-			LOG_WARNING("SDL: Failed setting fullscreen mode to %dx%d at %d Hz", displayMode.w, displayMode.h, displayMode.refresh_rate);
+			LOG_WARNING("SDL: Failed setting fullscreen mode to %dx%d at %d Hz",
+			            displayMode.w,
+			            displayMode.h,
+			            displayMode.refresh_rate);
 		}
 		SDL_SetWindowFullscreen(sdl.window,
 		                        sdl.desktop.full.display_res
@@ -1664,7 +1667,7 @@ uint8_t GFX_SetSize(const int width, const int height,
 
 		if (!SetupWindowScaled(RenderingBackend::Texture)) {
 			LOG_ERR("DISPLAY: Can't initialise 'texture' window");
-			E_Exit("SDL: Creating window failed");
+			E_Exit("SDL: Failed to create window");
 		}
 
 		/* Use renderer's default format */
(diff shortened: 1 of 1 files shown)


src/gui/sdlmain.cpp
```diff
@@ -1849,24 +1858,45 @@ uint8_t GFX_SetSize(const int width, const int height,
...
-					glGetProgramiv(sdl.opengl.program_object, GL_LINK_STATUS, &isProgramLinked);
-					if (!isProgramLinked) {
-						GLint info_len = 0;
-						glGetProgramiv(sdl.opengl.program_object, GL_INFO_LOG_LENGTH, &info_len);
-
-						if (info_len > 1) {
-							std::vector<GLchar> info_log(info_len);
-							glGetProgramInfoLog(sdl.opengl.program_object, info_len, nullptr, info_log.data());
-							LOG_ERR("SDL:OPENGL: Error link program:\n %s", info_log.data());
+					GLint is_program_linked;
```
kcgen, line 1870: Assign a default that assumes negative/failure (0, I guess?)

johnnovak: Yeah, I'll set it to `{}`.

src/gui/sdlmain.cpp
```diff
@@ -1849,24 +1858,45 @@ uint8_t GFX_SetSize(const int width, const int height,
...
+						std::vector<GLchar> info_log(info_len);
+
+						glGetProgramInfoLog(sdl.opengl.program_object,
+						                    info_len,
+						                    nullptr,
+						                    info_log.data());
+
+						if (is_program_linked) {
+							LOG_WARNING("SDL:OPENGL: Program info log:\n %s",
+							            info_log.data());
```
kcgen, line 1892: Just a question: if the program is linked, then the presence of the  message indicates some kind of warning?

(Or on other words: there are no messages when things are perfect?)

kcgen (approved):
Looks good! Just one minor question 👍

src/gui/sdlmain.cpp
johnnovak: We don't know, it's entirely driver dependent. But usually they are either warnings or helpful hints to improve things, so their presence indicates "warning like" scenarios.

src/gui/sdlmain.cpp
```diff
@@ -1509,23 +1509,32 @@ static GLuint BuildShader(GLenum type, const std::string& source)
...
+	if (info_len > 1) {
+		std::vector<GLchar> info_log(info_len);
+		glGetShaderInfoLog(shader, info_len, nullptr, info_log.data());
+
+		if (is_shader_compiled) {
+			LOG_WARNING("SDL:OPENGL: Shader info log: %s",
+			            info_log.data());
+		} else {
+			LOG_ERR("SDL:OPENGL: Error compiling shader: %s",
+			        info_log.data());
```
weirddan455, line 1528: I think these should just be prefixed with `OPENGL:` rather than `SDL:OPENGL:`.  I see in my logs where we use one or the other of these prefixes but never both in the same log line.

johnnovak: There are few more such cases for MIDI and VGA logging but just using `OPENGL` makes sense, so I'll do as you suggested.

Behold my crappy regex skills! 😆 

image

weirddan455: Yeah, it makes a bit more sense for MIDI and VGA I think.  A user might not know what ALSA is and I don't even know what those VGA sub-systems are.

`sdlmain.cpp` logging looks inconsistent though and it appears there are a few other places that use both prefixes.  I would prefer drawing the line at which function you last called when you got the error.  Here's an OpenGL function that's returned an error.  Although I guess you could make the argument that it's in our internal SDL "module".  I'll let you make the final judgement call on what you want to do.

src/gui/sdlmain.cpp
kcgen: Thanks - yup, I agree `LOG_WARNING` sounds like a good fit (because these aren't just bog-standard "sunny day" messages, if they exist, despite the program compiling).

Hopefully @Burrito78's GLSL 1.2 system has something to say at these spots!

src/gui/sdlmain.cpp
johnnovak: Agreed @weirddan455. I made a judgement call on when to use the `OPENGL:` and `SDL:` prefixes, to me it makes sense but feel free to disagree. Please have a look at the new logging improvement commits.

weirddan455: The line is kinda blurry on some of those where it's an SDL function and error returned by `SDL_GetError()` but it's clearly related to OpenGL.  I think these new logging changes are positive though.  It makes it more consistent.

weirddan455 (approved):
All looks good to me.  I built this and I am not showing any new error/warnings regarding shaders on my Nvidia + Linux system.  More logging is always good and hopefully will let us track down problems in the future.

johnnovak:
> All looks good to me. I built this and I am not showing any new error/warnings regarding shaders on my Nvidia + Linux system. More logging is always good and hopefully will let us track down problems in the future.

Thanks for testing this 👍🏻

src/gui/sdlmain.cpp
johnnovak: Yeah I said, judgement call, but my reasoning is SDL  acts as a very thin wrapper to OpenGL in those cases, and those are still legitimate OpenGL errors that SDL just forwards (e.g. not being able to request an sRGB framebuffer; that's an OpenGL issue, SDL just acts as a "messenger").

PR #3123, opened 2023-11-15
`sdlmain` improvements

johnnovak:
# Description

A bunch of mostly non-functional changes separated out of my upcoming viewport resolution/ aspect ratio enhancement PR.

Definitely review by commit.

Main changes:

- `SDL_Rect` is not leaking into the core emulation layers anymore; I've introduced a custom `Rectangle` type for this purpose.
- In 99% of cases there was no differentation between _logical coordinates_ (device independent coordinates) and _physical coordinates_ (device coordinates, physical pixels) in SDL main. This is very bad because you can never be sure what you're dealing with just by looking at the variable/function names, and it was making me quite upset after hacking sdlmain for a while... I've clarified the namings in the areas of sdlmain that I'm touching in my other PR directly or indirectly. About 80% of the code is still a mess, but it's a start...

For people who are not experienced with logical/physical coordinates, here's a quick rundown:

Using logical coordinates is the way forward for most use cases now that high DPI displays are increasingly the norm, especially for general GUI development and scalable vector graphics. You just use logical coordinates and things work fine on high DPI displays (if your framework is good), no need to think about it (that's what I'm doing in my other GLFW + OpenGL based programs that feature scalable vector graphics, and I can confirm, ThingsJustWork(tm)).

But when you're dealing with bitmap graphics and OpenGL, you _need_ to deal with physical pixels _as well_. Also, understanding how things to map to physical pixels is important because of the OpenGL shaders, for instance. So the code will be naturally a _mixture_ of logical and physical units, therefore we _must_ encode the units in the variable and function names, otherwise you can only guess things when reading the code.

Should I multiply this quantity by the DPI scale factor?
Should I not? 😕 
Should I divide it by the DPI factor? 🤔 

It's anyone's guess if the var is just called something like `sdl.clip.w`! 🥶 

That is bad, very bad, and it results in copious amounts of trace logging and intense hair loss during a compressed duration of time. Now, you don't wish that on your fellow devs, do you? 😎 

So the system I started adopting is as follows:

- In sdlmain we're dealing with a mixture of logical and physical coords. That's normal and it's never gonna change. As using logical coords is the preferred and recommended default these days (for the best possible reasons), names like `width` and `get_width` mean _logical coordinates_, while `width_px`, and `get_width_in_pixels` means—I'm sure you can figure that out 😎 SDL has also started adding the `*InPixels` postfix to function names in its API (although I'm not sure how consistent they are with that).
- In _all other parts of the emulation code_, we should not deal with logical units _at all_, this being a computer emulator that deals with emulated framebuffers, raw pixel arrays, and so on. So in other parts, there is no need for the `_px` and `_in_pixels` postfixes; it's implicit that `width`, `height`, etc. are always in pixels in the core layers. I stress it again: only sdlmain should deal with logical pixels! (Well, and the upcoming OSD layer as well, because it will draw the UI as scaleable vector graphics, but that's for later and it's another special case 😎)
- Another special case is passing windowing system related parameters, so for example canvas and viewport dimensions and similar quantities into the core emulation layer (e.g., the shader manager needs to know of the canvas dimensions in physical pixels to select the appropriate shader). So in order to avoid ambiguity, we still use the `_px` postfix for function arguments that have anything to do with window sizes, canvas sizes, viewport dimensions, etc. Can be a little bit redundant in some cases, but I'd rather be a little redundant than a little confusing 😄 
 
I'm hoping this is all not much controversial 😄 I don't think this is wasted effort because we don't know when we'll migrate to SDL3, plus clarifying the current sdlmain mess would help the transition effort too (I'm sure some of the current code could be (will be?) even lifted into the SDL3 branch).

# Manual testing

Tried the shaders and the various integer scaling and viewport resolution modes, fullscreen / windowed, etc., and there are no regressions.

# Checklist

I have:

### DIFF
--- a/include/rect.h
+++ b/include/rect.h
@@ -1,0 +1,67 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOSBOX_RECTANGLE_H
+#define DOSBOX_RECTANGLE_H
+
+#include "math_utils.h"
+
+namespace DosBox {
+
+// Struct to represent rectangles.
+struct Rect {
+	constexpr Rect() = default;
+
+	constexpr Rect(const float _w, const float _h)
+	        : x(0.0f),
+	          y(0.0f),
+	          w(_w),
+	          h(_h)
+	{}
+
+	constexpr Rect(const int _w, const int _h)
+	        : x(static_cast<float>(0.0f)),
+	          y(static_cast<float>(0.0f)),
+	          w(static_cast<float>(_w)),
+	          h(static_cast<float>(_h))
+	{}
+	constexpr Rect(const float _x, const float _y, const float _w, const float _h)
+	        : x(_x),
+	          y(_y),
+	          w(_w),
+	          h(_h)
+	{}
+
+	constexpr Rect(const int _x, const int _y, const int _w, const int _h)
+	        : x(static_cast<float>(_x)),
+	          y(static_cast<float>(_y)),
+	          w(static_cast<float>(_w)),
+	          h(static_cast<float>(_h))
+	{}
+
+	float x = 0;
+	float y = 0;
+	float w = 0;
+	float h = 0;
+};
+
+} // namespace DosBox
+
+#endif
--- a/include/render.h
+++ b/include/render.h
@@ -176,8 +176,8 @@ void RENDER_EndUpdate(bool abort);
 void RENDER_SetPalette(const uint8_t entry, const uint8_t red,
                        const uint8_t green, const uint8_t blue);
 
-bool RENDER_MaybeAutoSwitchShader([[maybe_unused]] const uint16_t canvas_width,
-                                  [[maybe_unused]] const uint16_t canvas_height,
+bool RENDER_MaybeAutoSwitchShader([[maybe_unused]] const uint16_t canvas_width_px,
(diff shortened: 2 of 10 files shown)


johnnovak:
Also, I need a recommendation for the new rectangle struct... First i called it `Rect`—that conflicted with some macOS core framework struct/class. Ok, then I renamed it to `Rectangle`—that conflicts with some Windows internals:

```
note: struct 'Rectangle' is hidden by a non-type declaration of 'Rectangle' here
    WINGDIAPI WINBOOL WINAPI Rectangle(HDC hdc,int left,int top,int right,int bottom);
```

So what should I call it? Other than `Rctngle` 😅 I'd hate to call it something like `DB_Rect` or `DosBoxRect` 🤦🏻 

Maybe `FloatRect`...

weirddan455:
Just call it `Rect` or `Rectangle` and wrap it in `namespace Dosbox` or something would be my suggestion.  Namespaces are the C++ feature meant to help with naming conflicts like that.

Grounded0:
It took years for most projects to migrate to SDL3 so i wouldn't stress about work being wasted with improving what we have now.

kcgen:
`WreckedAngle`

johnnovak:
> `WreckedAngle`

`Rectum`?

johnnovak:
Ok, went with the less controversial `DosBox::Rect` option 🤞🏻

weirddan455:
Wish I hadn’t said anything. I prefer `WreckedAngle` now.

include/rect.h
```diff
@@ -0,0 +1,67 @@
...
+	          h(static_cast<float>(_h))
+	{}
+	constexpr Rect(const float _x, const float _y, const float _w, const float _h)
+	        : x(_x),
+	          y(_y),
+	          w(_w),
+	          h(_h)
+	{}
+
+	constexpr Rect(const int _x, const int _y, const int _w, const int _h)
```
kcgen, line 52: For the argument names, just a minor suggestion to use verbose (or expanded) form because these are usually expanded by IDEs, so the expanded form is a bit more self-documenting for users of the class.

(for example: `x` and `y` often serve as width and height in lots of places, so spelling out the args as: `x_pos`, `y_pos`, `width`, `height` will help differentiate them)

This would then also avoids the need for all the underscore prefixes.

(probably unwarranted, but I try to always stay away of underscore prefixes just to avoid  inmplementation-specific reservations and name conflicts):

---

2003 C++ Standard:

    17.4.3.1.2 Global names [lib.global.names]

        Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.165

    7.1.3 Reserved identifiers

        All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.

https://stackoverflow.com/a/228797

kcgen (approved):
Just one minor suggestion.
This is a nice clarification in names, and more use of `optional<>` will help root out initialization order-of-operation problems.

include/rect.h
johnnovak: Good suggestion @kcgen; I wasn't in love with the underscores either.

kcgen: Right on, @johnnovak!

PR #3124, opened 2023-11-15
DO NOT REVIEW

johnnovak:
# Description

## Related issues

# Manual testing

# Checklist

I have:


src/gui/sdlmain.cpp
```diff
@@ -1411,27 +1424,84 @@ SDL_Window* GFX_GetWindow()
...
+			                sdl.desktop.dpi_scale);
+
+			return canvas_px.Contains(viewport_px)
+			             ? viewport_px
+			             : viewport_px.ScaleToFit(canvas_px);
+
+		} else if (sdl.viewport.fit.desktop_percent) {
+			const auto desktop = get_desktop_size();
+			LOG_TRACE("*** 2: desktop: %d x %d", desktop.w, desktop.h);
+			const auto p = *sdl.viewport.fit.desktop_percent;
```
kcgen, line 1478: If we programmatically expect the <optional> `desktop_percent` to be populated, suggest asserting before using it.

`assert(sdl.viewport.fit.desktop_percent)`

src/gui/sdlmain.cpp
```diff
@@ -1411,27 +1424,84 @@ SDL_Window* GFX_GetWindow()
...
+	// If SDL_GetWindowBordersSize fails, it populates the values with 0.
+	desktop.w -= (left + right);
+	desktop.h -= (top + bottom);
+
+	assert(desktop.w >= FallbackWindowSize.x);
+	assert(desktop.h >= FallbackWindowSize.y);
+	return desktop;
+}
+
+static DosBox::Rect calc_restricted_viewport_size_in_pixels(const DosBox::Rect canvas_px)
```
kcgen, line 1462: const ref

src/gui/sdlmain.cpp
```diff
@@ -3108,14 +3154,12 @@ static void setup_window_sizes_from_conf(const char* windowresolution_val,
 	        sdl.display_number);
 }
 
-DosBox::Rect GFX_CalcViewportInPixels(const int canvas_width_px,
-                                      const int canvas_height_px,
-                                      const int draw_width_px,
-                                      const int draw_height_px,
+DosBox::Rect GFX_CalcDrawSizeInPixels(const DosBox::Rect canvas_size_px,
+                                      const DosBox::Rect render_size_px,
```
kcgen, line 3158: const refs

include/video.h
```diff
@@ -352,13 +356,12 @@ bool GFX_HaveDesktopEnvironment();
 void MAPPER_UpdateJoysticks(void);
 #endif
 
-DosBox::Rect GFX_CalcViewportInPixels(const int canvas_width_px,
-                                      const int canvas_height_px,
-                                      const int draw_width_px,
-                                      const int draw_height_px,
+DosBox::Rect GFX_CalcDrawSizeInPixels(const DosBox::Rect canvas_size_px,
+                                      const DosBox::Rect render_size_px,
```
kcgen, line 360: const refs

src/gui/sdlmain.cpp
```diff
@@ -679,13 +688,12 @@ static uint32_t opengl_driver_crash_workaround(const RenderingBackend rendering_
 	return (default_driver_is_opengl ? SDL_WINDOW_OPENGL : 0);
 }
 
-static SDL_Rect calc_viewport_in_pixels(const int canvas_width_px,
-                                        const int canvas_height_px)
+static DosBox::Rect calc_draw_size_in_pixels(const DosBox::Rect canvas_size_px)
```
kcgen, line 691: const ref

src/hardware/vga_draw.cpp
```diff
@@ -2780,6 +2812,8 @@ ImageInfo setup_drawing()
 
 void VGA_SetupDrawing(uint32_t /*val*/)
 {
+	LOG_TRACE("VGA_SetupDrawing");
```
kcgen, line 2815: Left over from debugging?

src/gui/sdlmain.cpp
```diff
@@ -3126,47 +3170,46 @@ DosBox::Rect GFX_CalcViewportInPixels(const int canvas_width_px,
 	assert(std::isfinite(draw_scale_y));
 
 	// Limit the window to the user's desired viewport, if configured
-	const auto restricted_dims_px =
-	        restrict_to_viewport_resolution(canvas_width_px, canvas_height_px);
-
-	const auto bounds_w_px = restricted_dims_px.x;
-	const auto bounds_h_px = restricted_dims_px.y;
+	const auto bounds_px = calc_restricted_viewport_size_in_pixels(canvas_size_px);
```
kcgen, line 3173: this is a nice cleanup! finally collapsing all these free-floating dimensions into structs. :1st_place_medal:

include/sdlmain.h
```diff
@@ -89,6 +90,25 @@ enum PRIORITY_LEVELS {
...
 
+struct ViewportSettings {
+	ViewportMode mode = {};
+
+	// Either parameter can be set in Fit mode (but not both at the
+	// same time), or none
+	struct {
+		std::optional<DosBox::Rect> limit_rectangle = {};
+		std::optional<float> desktop_percent        = {};
+	} fit = {};
```
kcgen, line 101: Is the `limit_rectangle` just the relative dimensions (`width` and `height` only) or do its `x` and `y` values also position the viewport on the display, in an absolute sense?

The ``limit_rectangle` name implies the latter; just wanted to double check!

include/rect.h
```diff
@@ -29,33 +31,79 @@ namespace DosBox {
...
+	}
+
+	constexpr float y2() const
+	{
+		return y + h;
+	}
+
+	constexpr float cx() const
+	{
+		return x + w / 2;
```
kcgen, line 77: Suggest asserting that `w != 0`.

include/rect.h
```diff
@@ -29,33 +31,79 @@ namespace DosBox {
...
+	}
+
+	constexpr float cy() const
+	{
+		return y + h / 2;
```
kcgen, line 82: Suggest asserting that `h != 0`.

include/rect.h
```diff
@@ -29,33 +31,79 @@ namespace DosBox {
...
+	}
+
+	constexpr Rect Scale(const float s) const
+	{
+		return {x * s, y * s, w * s, h * s};
+	}
```
kcgen, line 88: I would think only the dimensions (or size) would be scaled, and not the position in space.

include/rect.h
```diff
@@ -29,33 +31,79 @@ namespace DosBox {
...
+
+	constexpr Rect Translate(const float dx, const float dy) const
+	{
+		return {x + dx, y + dy, w, h};
+	}
```
kcgen, line 93: These are member functions, so they really should modify the object itself. 

So I think these members should modify the object and then return a reference to itself, with a signature like:

```c++
Rect& Translate(const float dx, const float dy) {
   modify itself
   ...
   return *this;
}
```

This will allow for chaining: `my_rect.Scale(0.5).Translate(1, 1);`

To illustrate: imagine calling `my_vector.push_back(item)`, but it not actually adding it to `my_vector` and instead only the returned result had it added.

The pure functions have a place, but normally they'd take in the object being operated on in the args, and would be stand-alone functions.

kcgen (commented):
Mostly minor suggestions.
Lots of nice improvements in this, @johnnovak !

include/rect.h
```diff
@@ -21,6 +21,8 @@
 #ifndef DOSBOX_RECTANGLE_H
 #define DOSBOX_RECTANGLE_H
 
+#include <algorithm>
+
 #include "math_utils.h"
```
kcgen, line 26: Is `math_utils.h` needed?

johnnovak:
Thanks for the early review @kcgen , just checking that you're aware it's still a draft and work-in-progress? It's only about 70% finished yet.

include/sdlmain.h
johnnovak: Just width & height. I just don't wanna go overboard with having a different `Size` type in addition to `Rect`, I think it's overkill. From code it becomes obvious what we're doing, and there's a limit how much info you can encode into names & types (unless you're happy to introduce tons of types 😅).

src/gui/sdlmain.cpp
```diff
@@ -198,6 +198,11 @@ static DosBox::Rect to_rect(const SDL_Rect r)
 	return {r.x, r.y, r.w, r.h};
 }
 
+static SDL_Rect to_sdl_rect(const DosBox::Rect r)
```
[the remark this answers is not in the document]
johnnovak: For small things, references can actually make things worse. Small types can be passed in registers, there's no forced memory indirection, etc.

I looked this up a while ago and structs under about 16-byte in size are better to passed by value. Makes perfect sense with 128-bit wide registers being standard. E.g. check this answer:

https://stackoverflow.com/questions/3314010/should-small-simple-structs-be-passed-by-const-reference

SDL_Rect can be 256-bits, so if we're in micro optimising territory, passing 256-bits by copy is potentially a lot cheaper than memory indirection.

src/gui/render.cpp
```diff
@@ -725,23 +724,49 @@ static const char* to_string(const enum MonochromePalette palette)
...
+		return AspectRatioCorrectionMode::On;
+
+	} else if (mode == "viewport") {
+		return AspectRatioCorrectionMode::Viewport;
+
+	} else {
+		LOG_WARNING("RENDER: Invalid aspect ratio correction mode '%s', using 'on'",
+		            mode.c_str());
+		return AspectRatioCorrectionMode::On;
+	}
```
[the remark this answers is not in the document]
johnnovak: I prefer the else if, this structure make it clearer to me what's going on. I get the argument for exiting early / avoiding multiple nestings in big functions though.

include/rect.h
```diff
@@ -29,33 +31,79 @@ namespace DosBox {
...
-	          y(static_cast<float>(0.0f)),
-	          w(static_cast<float>(_w)),
-	          h(static_cast<float>(_h))
-	{}
 	constexpr Rect(const float _x, const float _y, const float _w, const float _h)
 	        : x(_x),
 	          y(_y),
 	          w(_w),
 	          h(_h)
+	{
```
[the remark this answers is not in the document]
johnnovak: Already done :)

include/rect.h
```diff
@@ -29,33 +31,79 @@ namespace DosBox {
...
+		return {x + dx, y + dy, w, h};
+	}
+
+	constexpr bool Contains(const Rect r) const
+	{
+		return (r.x1() >= x1() && r.x2() <= x2()) &&
+		       (r.y1() >= y1() && r.y2() <= y2());
+	}
+
+	constexpr Rect ScaleToFit(const Rect dest) const
```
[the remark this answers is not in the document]
johnnovak: See my other comment below; at least up to 16-bytes it's better to pass by value.

src/gui/sdlmain.cpp
```diff
@@ -694,63 +702,61 @@ static SDL_Rect calc_viewport_in_pixels(const int canvas_width_px,
...
+			return calc_draw_size_in_pixels(
+			        {canvas_size_override_px->w,
+			         canvas_size_override_px->h});
+		} else {
+			const auto canvas_px = get_canvas_size_in_pixels(
+			        rendering_backend);
 
-	const auto [viewport_w_px, viewport_h_px] = get_viewport_size_in_pixels();
+			return calc_draw_size_in_pixels(canvas_px);
+		}
```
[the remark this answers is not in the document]
johnnovak: Yeah, but I like it to see there's branching there.

src/gui/sdlmain.cpp
```diff
@@ -1411,27 +1424,84 @@ SDL_Window* GFX_GetWindow()
...
+			        sdl.viewport.fit.limit_rectangle->Scale(
+			                sdl.desktop.dpi_scale);
+
+			return canvas_px.Contains(viewport_px)
+			             ? viewport_px
+			             : viewport_px.ScaleToFit(canvas_px);
+
+		} else if (sdl.viewport.fit.desktop_percent) {
+			const auto desktop = get_desktop_size();
+			LOG_TRACE("*** 2: desktop: %d x %d", desktop.w, desktop.h);
```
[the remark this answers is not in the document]
johnnovak: Unfinished PR, as mentioned below 😄

src/gui/sdlmain.cpp
johnnovak: That's what the `if (sdl.viewport.fit.desktop_percent)` does.

include/rect.h
johnnovak: Yeah, it's a toss-up, really. Both are fine but needs to be documented. I guess scaling them all is a bit more flexible, though.

include/rect.h
johnnovak: > These are member functions, so they really should modify the object itself.

As a user of functional languages on a daily basis at work, I disagree 😄 Those have sizable standard libraries and almost nothing mutates; all class methods return copies. Then there are some mutating variants, but their use is discouraged—you have to actively ask for it, never mutates by default.

But ok, we're in C++. I think the compiler would optimise it anyway, but I don't care much, happy to mutate. Btw, these rects won't be used in any high-performance code.

include/sdlmain.h
kcgen: yup - was only asking just for the name of the variable (`limit_rectangle`) -- if it's just dimensions, maybe it could be named `limit_dims` or `limit_dimensions`)?  

Rectange at least to me says that position is probably involved too :-)

include/rect.h
kcgen: Maybe `ScaleSize(s)` and `ScalePosition(s)` could just do those components?

include/rect.h
kcgen: It's how all STL C++ objects work when you call their member functions.

```c++
string s = 'hello';
s.append(' world');
```

We expect that `s` contains `hello world`.

If it behaved like the above, we would need:

```c++
string s = 'hello';
const auto new_s = s.append(' world');
```

- `s` would still contain `hello`
- `new_s` would have `hello world`.

kcgen:
> Thanks for the early review @kcgen , just checking that you're aware it's still a draft and work-in-progress? It's only about 70% finished yet. There's gonna be lots of force pushes and re-jigging things still, plus not everything in the top comment is implemented yet (and that comment is not finished either; there will be screenshots coming, etc.)

Yup - might as well nip the easy stuff early. Was only commenting on superficial/rote things.

include/rect.h
johnnovak: Actually, not needed, we're not dividing by it.

src/gui/sdlmain.cpp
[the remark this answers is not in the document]
kcgen: Wasn't meaning to micro-optimize.. just the old rule of thumb to use `const ref` for compound types (more a consistency things).

I guess setting thresholds when to use references versus copies is more of a micro-optimization thing. Now that we've silenced PVS Studio on these, I guess it won't be complaining about them :-)

include/sdlmain.h
johnnovak: Yeah fair point, renaming.

include/rect.h
johnnovak: Nope.

include/rect.h
johnnovak: Well, that's STL 😄  Yep, that's exactly how strings work in "immutable by default" languages. All strings are immutable. You need to ask for special `MutableStrings` if you want to muck around with the contents of a strings in place. It's actually nice because if you use mutable objects in hash maps that can lead to obscure bugs if you mutate them later... (been there in Java 😅 )

Ok, I've rewritten it to mutate. I kinda liked the first form better but I can live with it 😄 

```cpp
const auto viewport_px = limit_rect.Scale(sdl.desktop.dpi_scale);

return canvas_px.Contains(viewport_px)
			 ? viewport_px
			 : viewport_px.Intersect(canvas_px);

```

```cpp
auto viewport_px = limit_rect;
viewport_px.Scale(sdl.desktop.dpi_scale);

if (!canvas_px.Contains(viewport_px)) {
	viewport_px.Intersect(canvas_px);
}

return viewport_px;
```

src/gui/sdlmain.cpp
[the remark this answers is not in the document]
johnnovak: Ok, I can do it for consistency as it doesn't actually matter much, yeah 😄 

Well, I'd prefer to always write `const Type` and never think about it... I actually don't get why can't the compiler figure that out... hey, it's a const, so you pass it by value or reference, whatever is better, but we're not gonna mutate it.

I guess it just annoys me to know I'm passing something by reference when I know it can fit into a register 😄

kcgen:
Oh - wanted to mention this early on - it would be really slick to combine the Fraction class with `aspect` to allow for  natural strings like 16:9, 4:3, 4:5 (apologies if it's already there and I just didn't catch it! 😅)

include/rect.h
johnnovak: Okay, I've done the always return self-reference trick, plus also introduced a `Copy()` method so we can copy & transform & assign one rect to another const rect. It's not so bad, although more error-prone than the pure version. But so is STL, so ThisIsFine(tm) 😆

johnnovak:
> Oh - wanted to mention this early on - it would be really slick to combine the Fraction class with `aspect` to allow for natural strings like 16:9, 4:3, 4:5 (apologies if it's already there and I just didn't catch it! 😅)

Cool idea, that's something to think about, although in its current iteration things will work a bit differently. You'll see when I'm done, then we can discuss again 👍🏻

Kappa971:
> Oh - wanted to mention this early on - it would be really slick to combine the Fraction class with `aspect` to allow for natural strings like 16:9, 4:3, 4:5 (apologies if it's already there and I just didn't catch it! 😅)

Speaking as a normal user, this is an interesting addition but the "overflow" function seems a bit too complex for an inexpert user (well maybe even for an expert user).
Wouldn't an option like `aspect = 16:9 <zoom%>` be simpler? (with `aspect = 4:3` as default).
I don't know...

johnnovak:
> > Oh - wanted to mention this early on - it would be really slick to combine the Fraction class with `aspect` to allow for natural strings like 16:9, 4:3, 4:5 (apologies if it's already there and I just didn't catch it! 😅)
> 
> Speaking as a normal user, this is an interesting addition but the `overflow` option seems a bit too complex for an inexpert user (well maybe even for an expert user). Wouldn't an option like `aspect = 16:9 <zoom%>` be simpler? (with `aspect = 4:3` as default). I don't know...

Good to see you're thinking about possible simplifications and taking user friendliness into account @kcgen @Kappa971!

To be absolutely crystal clear, I'd also like to make this as simple and intuitive as I can, so I gave your ideas very serious consideration, which means literally spending the last 3 hours (!) sketching out ideas on alternative config implementations based on your suggestions and trying out various things hacking the code. The TL;DR is, I've been going around in circles because the simplifications all break down in some scenarios and/or are unsupportable or rather meaningless. So I'm pretty sure I've hit a local maximum with my current solution that balances flexibility, simplicity, and ease of use. I just can't simplify it any further while catering for all current and foreseeable future needs, and the current solution is already a culmination of a week's thinking, carefully weighing of pros and cons of various possible solutions. This is already about my 4th or 5th iteration of the config options!

Believe me, alternative solutions would make things a lot more complex. I've spent about 100+ hours dealing with aspect ratio-related things earlier this year dealing with the VGA code. One side effect of that work is the following huge aspect ratio table of various DOS video modes:
https://github.com/dosbox-staging/dosbox-staging/wiki/Video-emulation-tests#summary-of-dos-video-modes

I don't want to appear condescending and I'm absolutely saying this in a friendly, amicable manner: understanding how the various rendering options work together with aspect ratio stuff requires a bit of an expert knowledge — it's far more complex than just "yeah, stretch to 4:3 or 16:9", that's not even scratching the surface of understanding the implications and what we're trying to solve here. Just to enter a meaningful discussion about the issue I'm trying to solve, a deep understanding of the below points is required:

- 100% understand the differences between storage, display, and pixel aspect ratios, 
- the reasons why graphics in some DOS games appear in the wrong aspect ratio (e.g., Amiga ports, Hercules games, etc.; these reasons are as much "societal" as technical),
- familiarity how _wildly_ different pixel aspect ratios all pre-640x480 VGA / square-pixel VESA/SVGA video modes have (see my table)
- how CRT monitors work, specifically the horiz and vert size adjustments (how early monitors did not have these controls at all, then later monitors had "global" analog controls, then even later 90s monitors had digital controls where you could dial in specific horiz/vert stretch settings per display timings, so different stretch factors _per video mode (!)_  in practice)
- what our host-side limitations on modern flat screens are,
- how integer scaling, shaders, existing viewport restriction etc. would work together with all this (including border support),
- how would all this be adjusted in the OSD in real-time in an intuitive manner,
- and I'm sure there's something else I forgot...

I could spend the whole day writing a dissertation about how some of the proposed simplifications would break various things (they are good ideas on the surface, but naive), but just a quick rundown:

- `aspect = on` does _not_ simply stretch whatever is rendered on the DOS side to a 4:3 output rectangle, that's a misconception. The actual pixel aspect ratio (PAR) is calculated based on VGA timings, then based on the resulting PAR and video mode timings we can end up with pretty much _any_ output image aspect ratio. It is _usually_ 4:3, or close to 4:3, but it can be widescreen(ish) in some games and VESA text modes. Stretching to 4:3 no matter what would break a small but significant number of games/programs.
- `aspect = 16:9` is not that helpful; the use case here would be for people who just wanna stretch the content to fullscreen, no matter what, to get rid of any black bars. They don't care if they have a 16:9 or a 16:10 screen, or whatever else; for them, `aspect = viewport` and the default `viewport_resolution = fit` option is actually best for the job.
- To make Hercules games displaying 320x200 assets in the 720x348 Hercules graphics to not appear squashed, coming up with the correct _fractional_ (!) aspect ratio would do your head in, trust me. I just gave up after literally 30 mins of trying and failing 😅 Surely it can be calculated, but that's not helpful to users. What is just allowing them set arbitraty stretch factors as percentages, just like on a real CRT with the horiz/vert adjustments. Then you adjust until it looks right, done. This could be exposed on the OSD too as continuous stretch controls. Calculating this _exactly_ then having to specify `aspect = 13390:10004` or `aspect = 0.898441` (just making shit up, but it would be something obscure like this) is not more user-friendly.
- There is the dilemma of what the `aspect` setting should actually do: correcting _pixel aspect ratios_, or stretching the output to an arbitrary _output image aspect ratio_ (display aspect ratio)? Currently, it's a bit vague, but it's best to keep it that way, actually, `on` now does the calculation based on VGA timings and in some cases heuristics, `off` just assumes square pixels. I will rename the `aspect = viewport` setting to `aspect = fit` that is quite intuitive; it will just massage the aspect ratios so the image _fits_ into the viewport rectangle. Specifying PARs would make things quite complex.

---

I could go on...

In the end:

- Complex issues can be simplified to some extent, but can't be made _simpler_ at a certain point without making the feature useless.
- To use complex features, you either need to have a full understanding of what's involved (I 100% agree that will be rare among users), _OR_ you need to rely on "config recipes" for the most common things you wanna do.
- These recipes will be provided in the manual—that's the easy option. Wanna do this, put this into the config. The manual testing section will contains such "recipes" in action.
- Advanced users who understand these things well can of course do whatever.

Now, after a 4 hours detour, finishing the feature as I had planned 😎

kcgen:
Thanks for the explanation and clarification @johnnovak!

I think the setting as you're suggesting (specifically the name and use of percentages) makes perfect sense (and makes the intent clearer than using ratios which users would no doubt confuse with existing TV display dimensions), given this is an end-stage stretch to the accurately sized resulting image.

We're handing over the correctly sized output image, just as would have happened on a CRT, and the end user can still apply a bit more h-size and v-size (stretch or squeeze) just to fix up those rare modes and games that still won't look ideal.

src/gui/sdlmain.cpp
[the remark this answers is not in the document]
johnnovak: @kcgen Related to this, PVS Studio started complaining about this:

V835. Passing cheap-to-copy argument by reference may lead to decreased performance.
https://pvs-studio.com/en/docs/warnings/v835/

So that's basically what I was talking about 😄 

This pass by reference triggered it

```cpp
static DosBox::Rect to_rect(const SDL_Rect& r)
```

I'm passing `DosBox::Rect` by ref everywhere now for consistency, and it doesn't complain, so I'll leave those. I guess I just wanted to show that it has some heuristics for raising warnings for inefficient pass by refs, which is cool.

Grounded0:
Hopefully no performance regressions either.

johnnovak:
> Hopefully no performance regressions either from this PR.
> 
> 0.81 release notes:
> 
> • Max 100% is now thread capacity 100% for optimum performance
> 
> Previously, using the max 100% setting resulted in only 90% of the thread capacity being used. Now DOSBox Staging uses the full 100% capacity of the thread, resulting in similar or better performance versus the previous 0.80.1, even with the new accuracy improvements. This also makes the main thread go into oversubscription at times (over 100% capacity), resulting in a faster and more aggressive boost clock ramp-up.

Nice one 🎖️

Grounded0:
I'm concerned is this going to bring us performance regressions when we're currently at 0.80.1 or better performance as stated in the release notes.

johnnovak:
> I'm concerned is this going to bring us performance regressions when we're currently at 0.80.1 or better performance as stated in the release notes.

No need to be concerned as it has exactly zero impact on performance. It affects viewport scaling calculations on window size changes and fullscreen/windowed changes, that's it. It literally has zero performance impact.

johnnovak:
Read for review, gentlemen! 😎  @kklobe @FeralChild64 @weirddan455 @kcgen

FeralChild64:
With the following settings:

```ini
[render]
aspect = stretch
viewport = relative 110% 160%
integer_scaling = off
```

the seamless mouse integration is completely broken.

FeralChild64:
Most likely the `MouseScreenParams` structure needs to be updated, together with calculations on `update_cursor_absolute_position` in `mouse.cpp`.

johnnovak:
> Most likely the `MouseScreenParams` structure needs to be updated, together with calculations on `update_cursor_absolute_position` in `mouse.cpp`.

Thanks for the pointer, I started digging around and I was just looking at that struct. Probably I'll need to allow negative clip coords or something...

Very nice job on interfacing with the GFX subsystem, and I like those `MOUSE_Notify*` functions. Very elegant, that's my planned approach for the OSD notifications too 🚀 

Please remind me, I know we had a few discusions about seamless mode, and wasn't the conclusion that it's flaky and cannot be supported 100% reliably in pure DOS games, only with the VMware drivers under Win 3.x because those can report absolute positions?

Regardless, I'll try to fix the regression, so this change should not make the situation much worse even for pure DOS games if I can avoid it.

FeralChild64:
Yes, seamless mouse integration works with same games/software, cannot work with other. DOS Navigator and Norton Commander are compatible, fortunately.

Windows 3.1x is not compatible and requires special mouse driver, currently one of:
- https://github.com/NattyNarwhal/vmwmouse - uses VMware interface, driver is known to be unstable
- https://git.javispedro.com/cgit/vbmouse.git/about/ - uses VirtualBox interface
- https://git.javispedro.com/cgit/vbados.git/about/ - uses slightly extended INT33 interface, has great mouse wheel support

johnnovak:
@FeralChild64 Just letting you know that I'm 80% done with a fix, so don't waste time on it. I'll push it as separate commits when I'm done, most likely sometime later today.

johnnovak:
@FeralChild64 seamless mouse mode now works seamlessly! 😎 See the last two commits (plus this triggered a bunch of further refactors/cleanups unrelated to this).

Updated the manual testing section with relevant info.

johnnovak:
> Check for performance regressions against `main`. I cannot since i'm preoccupied with the release notes.

@Grounded0 I explained to you yesterday that it's impossible to get any performance changes with this change... But okay, ran a couple of Quake Bench tests with main and this branch, and unsurprisingly, there's no difference.

Like I said, it can't be any difference 😄

johnnovak:
@kcgen @weirddan455 @kklobe @FeralChild64 Code review, please... anyone? 😄

weirddan455:
I don't have any in-flight PRs except for ffmpeg but that's on pause until after release.  I'll review it once I've got some time.  Going commit by commit, you've still got 44 commits here.

johnnovak:
> I don't have any in-flight PRs except for ffmpeg but that's on pause until after release. I'll review it once I've got some time. Going commit by commit, you've still got 44 commits here.

My bad, I was referring to this (not a PR, yeah):
https://github.com/dosbox-staging/dosbox-staging/issues/3132

Btw, that is really not important, so no need to try to rush that into the current release.

include/sdlmain.h
```diff
@@ -203,39 +288,41 @@ struct SDL_Block {
...
+		// in ms, for use with PIC timers
+		double period_ms      = 0.0;
 		float max_dupe_frames = 0.0f;
-		int period_us       = 0; // same but in us, for use with chrono
+
+		// same but in us, for use with chrono
+		int period_us       = 0;
 		int period_us_early = 0;
-		int period_us_late    = 0;
+		int period_us_late  = 0;
```
kklobe, line 316: I know this is just a formatting change, but should these be `int64_t` if they are microseconds, and used with chrono, since we also updated `GetTicksxx` to return int64_t's?

kklobe:
> @kcgen @weirddan455 @kklobe @FeralChild64 Code review, please... anyone? 😄
> 
> It's not that hard if you go commit by commit as many changes are intentionally kept small and are just renames and shuffling things around. I optimised things for easy reviewability (is that even a word? 😅) and bisecting.

Workin' on it!

> 
> @weirddan455 and @kklobe have two in-flight PRs that are dependent on this going in (but I guess we can not merge them for the RC if this drags on for long).

No worries, I've already rebased my PR off this, so won't be too much work.

include/sdlmain.h
kcgen: `int` is OK, because these are boundary values calculated from the video frame period.
- The largest late period is ~60,000 us if someone overrides the DOS rate to the 24 Hz min (producing the largest late frame period)
- The smallest early period is ~550 us if someone overrides the DOS rate all the way to the 1000 Hz max (producing the smallest early frame period).

johnnovak: Yeah and I'm reluctant to add more to this PR anyway 😅 So no "while you're at it, can you also..." type of comments now, please 😄

src/gui/render.cpp
```diff
@@ -1103,7 +1103,13 @@ static void init_render_settings(Section_prop& secprop)
 	        "                   beyond the window/screen. Useful to force arbitrary display\n"
 	        "                   aspect ratios with 'aspect = stretch' and to zoom into the\n"
 	        "                   image. This effectively emulates the horizontal and vertical\n"
-	        "                   stretch controls of CRT monitors.");
+	        "                   stretch controls of CRT monitors.\n"
+	        "\n"
+	        "Notes:\n"
+	        " - Using 'relative' mode with 'integer_scaling' enabled could lead to\n"
+	        "   surprising (but correct) results.\n"
+	        " - You can use the 'Strech Axis', 'Inc Stretch', and 'Dec Stretch' hotkey\n"
```
kklobe, line 1106: typo 'Stretch Axis'

include/sdlmain.h
```diff
@@ -203,7 +203,7 @@ struct SDL_Block {
 
 		PixelFormat pixel_format = {};
 
-		double dpi_scale = 1.0;
+		float dpi_scale = 1.0;
```
kklobe, line 223: 1.0f for consistency?

include/rect.h
```diff
@@ -21,41 +21,186 @@
...
+	constexpr Rect(const float x_pos, const float y_pos, const float width,
+	               const float height)
+	        : x(x_pos),
+	          y(y_pos),
+	          w(width),
+	          h(height)
+	{
+		assert(w >= 0.0f);
+		assert(h >= 0.0f);
+	}
```
kcgen, line 46: These asserts are good, but because it's a `struct`, the API encourages directly working with member variables. 

```
// some bit where we update frame size..
window_frame.w = (math based on some variables);
window_frame.h = (math based on some variables);
```

This on-the-fly adjustment is where we could accidentally violate the asserts (if something returned a negative, even if it's very small), leading to the polygon overlap / intersect / etc.. routines not being right.

One option is making them private and asserting on the only path that's allowed to change them.

src/gui/render.cpp
```diff
@@ -725,30 +722,299 @@ static const char* to_string(const enum MonochromePalette palette)
...
+}
+
+static std::optional<ViewportSettings> parse_fit_viewport_modes(const std::string& pref)
+{
+	if (pref == "fit") {
+		ViewportSettings viewport = {};
+		viewport.mode             = ViewportMode::Fit;
+		return viewport;
+
+	} else if (const auto width_and_height = parse_int_dimensions(pref)) {
```
kcgen, line 814: the 'if' body looks a bit strange without knowing that `parse_int_dimensions()` returns an optional.

For `<optional>` return values, I like to prefix the variables with `maybe_` because the variable might not exist, and then strip off the `maybe_` once we've confirmed it's available:

(this is a good use of the C++17's `if (initializer; conditional)` syntax where we can test the optional in the condition):

```c++
if (const auto maybe_w_and_h = parse_dims(pref); maybe_w_and_h) {
    const auto [w, h] = *maybe_w_and_h;
    ...
}
```

src/gui/render.cpp
```diff
@@ -896,6 +1283,25 @@ void RENDER_AddConfigSection(const config_ptr_t& conf)
...
+	                  0,
+	                  "incstretch",
+	                  "Inc Stretch");
+
+	MAPPER_AddHandler(decrease_viewport_stretch,
+	                  SDL_SCANCODE_UNKNOWN,
+	                  0,
+	                  "decstretch",
+	                  "Dec Stretch");
+
```
kcgen, line 1304: Suggest a single hotkey that activates the stretch controls.
The controls themselves can be fixed to something like:
 - left & right arrow keys: horizontal shrink and stretch.
 - down & up arrow keys: vertical shrink and stretch.

(this is how the joystick shift and scale controls work using a single "throw switch" that gives you the control, after which you user the arrow keys to very intuitively tweak the joystick's shift and scale wheels along all four axis)

The single activator button is just like flipping open the CRT control panel:



Once the panel is flipped open, you've got your left-right and up-down buttons for very intuitive realtime tweaking:



Then the user hits Ctrl+Alt+S again to disable the stretch settings.

And the arrow keys go back to normal.

The user would be able to change the Ctrl+Alt+S to something else, just in case this is an actual in-game key. But I think Ctrl+Alt+S is rarely used, so it's probably reasonable to have this available by default to access the stretch settings.

FeralChild64:
> > @kcgen @weirddan455 @kklobe @FeralChild64 Code review, please... anyone? 😄

Sorry, I won’t have time till the weekend. But I have already tested that seamless mouse breakage is fixed.

kcgen (commented):
Just a couple minor suggestions, and one notable one about the hotkeys and controls.

Also wanted to mention that mid '90s ViewSonic and NEC monitors could store settings like this per-video-mode.

So a nice addition would be to make the stretch conf setting and the LOG result handle multiple entries.

This got more important as games (and software) started to mix lower resolution modes with higher resolutions. Sometimes cutscene videos flipped back to lower resolution modes and then back up to higher resolution modes in game.

(But I'm not sure if we're affected by this; maybe we already disable aspect for the higher resolution VESA modes - so perhaps this is moot?)

src/gui/sdlmain.cpp
```diff
@@ -1460,21 +1460,65 @@ RenderingBackend GFX_GetRenderingBackend()
 	return sdl.rendering_backend;
 }
 
+static SDL_Rect get_desktop_size()
+{
+	SDL_Rect desktop;
+	assert(sdl.display_number >= 0);
+
+	SDL_GetDisplayBounds(sdl.display_number, &desktop);
```
weirddan455, line 1476: Possibly uninitialized variable `SDL_Rect desktop;`.  It's not initialized to 0 and the return value of this function is not checked.

Also, this entire function seems strange to me.  Why do we care about the size of window decorations when calculating desktop size?

weirddan455 (commented):
Went through this commit by commit.  Just the one thing stood out to me although to be honest my eyes kind of glaze over once PRs get above a certain size.

Also did a quick compile + test with the "stretch to full" settings you mentioned and it seemed to work.

src/gui/sdlmain.cpp
johnnovak: Dunno, but out of scope for now 😅 I'll add the defaulting, though.

include/rect.h
```diff
@@ -21,41 +21,186 @@
...
+			const auto iy1 = std::max(y1(), r.y1());
+			const auto iy2 = std::min(y2(), r.y2());
+
+			if (iy1 < iy2) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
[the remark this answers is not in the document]
johnnovak: Yep, I'll add tests.

src/gui/sdlmain.cpp
kcgen: > Why do we care about the size of window decorations when calculating desktop size?

In windowed mode, if you set `windowresolution = desktop`, SDL will avoid clobbering/overlapping your main control panel (similar to setting an application to "maximized" window state).  SDL is smart enough to be able to query the WM on Windows, macOS, Linux, (even BSD), and also works if the desktop OS supports using a vertical-side panel instead of at the top or bottom.

src/gui/render.cpp
johnnovak: Yeah, I guess I can, although it's a bit wordy. In `programs_mixer.cpp` I've just done something like this in many places:

```ini
if (const auto v = parse_volume(parts[0]); v) {
  const SetVolume cmd = {AudioFrame(*v, *v)};
```

Might seem strange a bit first, but people quickly get used to it that the `parse_*` functions all return options.

Maybe I'll do the "maybe" here 😄 , because there's only a few `parse` invocations. If there were many, I'd just say people need to learn how the `parse` functions work.

src/gui/render.cpp
johnnovak: That's a nifty idea, but the problem is I don't think you can do that at runtime. Meaning once you have added a key handler with `MAPPER_AddHandler`, you can't remove it again. I really don't wanna make mapper changes as part of this PR! 😅 (or touch the mapper at all except for basic stuff, as a matter of fact).

You kinda worked around this with the `use_joy_calibration_hotkeys` setting. I guess the idea is you add that to your config, calibrate the joystick, then remove it from your config once you're arrived to the correct settings. A bit clunky, but does the job 😄 

Also, because all those cursor + modifier combos need to be defined as separate hotkeys, there's a large-scale "hotkey pollution" going on with `use_joy_calibration_hotkeys` enabled:

image

So... dunno. I'm kinda inclined to leave it as is because I don't see the joy approach to be really superior. Plus the current approach is consistent with the CGA composite colour adjustments which works similarly to how I've done it.

Overall, It's one of those things that we'll be able to do a lot better with the new mapper. This does the job for now, and we can always iterate on it and refine it in the future.

johnnovak:
> Just a couple minor suggestions, and one notable one about the hotkeys and controls.
> 
> Also wanted to mention that mid '90s ViewSonic and NEC monitors could store settings like this per-video-mode.
> 
> So a nice addition would be to make the stretch conf setting and the LOG result handle multiple entries.
> 
> This got more important as games (and software) started to mix lower resolution modes with higher resolutions. Sometimes cutscene videos flipped back to lower resolution modes and then back up to higher resolution modes in game.

I considered this, but don't wanna overengineer it from the get-go 😅 

As mentioned in the "future work" section, separate text and graphics stretch factors would be a practical compromise later on:

```
viewport = relative text 100% 100% graphics 110% 160%
```

I was toying the idea of extending that so you can target specific video modes by their dimensions (that's the simplest practical approach), e.g.:

```
viewport = relative text 100% 100% mode-320x200 110% 160% mode-640x400 100% 120% 
```

Something like that. 

> (But I'm not sure if we're affected by this; maybe we already disable aspect for the higher resolution VESA modes - so perhaps this is moot?)

Yeah, I presently think it's massive overkill, but certainly an option for the future.

include/rect.h
johnnovak: Good points, and seem good suggestions on the surface, but the more I think about, I'm coming to the conclusion I want this `Rectangle` class to be flexible and support generic use cases. That includes rectangles with zero dimensions, and possibly even negative dimensions. These are just useful in generic algorithms, same way an "empty list/container" is useful. A rectangle can be similarly "empty" if its dimensions are zero, and negative dimensions are valid for certain algorithms and/or for intermediate results.

I'm considering turning this into something similar to the Rectangle class from the Java standard library which allows and handles zero and negative dimensions. It's worthwhile to read the intro paragraph, but some key points are below:
https://docs.oracle.com/javase/8/docs/api/java/awt/Rectangle.html

> A Rectangle whose width or height is exactly zero has location along those axes with zero dimension, but is otherwise considered empty. The [isEmpty()](https://docs.oracle.com/javase/8/docs/api/java/awt/Rectangle.html#isEmpty--) method will return true for such a Rectangle.

> Methods which test if an empty Rectangle contains or intersects a point or rectangle will always return false if either dimension is zero.

> Methods which combine such a Rectangle with a point or rectangle will include the location of the Rectangle on that axis in the result as if the [add(Point)](https://docs.oracle.com/javase/8/docs/api/java/awt/Rectangle.html#add-java.awt.Point-) method were being called.

> A Rectangle whose width or height is negative [...] is also empty with respect to containment calculations

So I'm thinking of allowing everything, keeping the direct property access which is desirable, and then asserting a non-empty rectangle is the caller's responsibility if such restriction is desired (varies on a case by case basis):

```cpp
assert(rect.IsNonEmpty());
```

That moves the assert up the chain to its correct place, and we keep our flexible rectangle implementation.

src/gui/sdlmain.cpp
johnnovak: That was my guess, thanks for clarifying @kcgen .

src/gui/render.cpp
kcgen: That's a lot of maybes!

Also no worries - having the test `(... ; v) {` is the self-documenting part -> so that works too!

include/rect.h
kcgen: All sounds good; and like that approach.
Just having a generic rectangle struct is very handy - even if you only use as a collection instead of a functional object.

Yeah - for the functions, like Intersect, that's where you could add the `assertm`'s and mention the intersection algorithm only handles positive WxH values.

src/gui/render.cpp
johnnovak: To be clear @kcgen, I like your idea a lot, and this would be the ultimate goal for keyboard-driven viewport adjustments, no disagreements. We could also throw the Shift modifier into the mix to allow 5% increments instead of just 1%, and ideally, we'd also have keyrepeat, just like on a real monitor that has digital width/height controls. Now you must press the adjustment keys many many times, but that's how the mapper works...

But we need some sane infrastructure to go in first, yeah 😅

johnnovak:
Thanks for the review @kcgen @kklobe @weirddan455 , I'll address the comments tomorrow!

src/gui/render.cpp
johnnovak: Adding `maybe_*` prefix in `parse_relative_viewport_modes()` because it clarifies things there. Leave the rest as-is when I'm checking for the optional value being set in `if` statements.

FeralChild64 (approved):
I'm not familiar with the SDL code, but mouse-related changes look OK to me.

PR #3131, opened 2023-11-17
Improve auto-cycle convergence rate with small adjustment

### LINKED ISSUE #3122: Apple Silicon Mac has random 40% performance loss on 80% of DOSBox Staging runs

DEV

macOS (Apple Silicon)

macOS Sonoma 14.1.1 (23B81)

### Relevant hardware info

Mac mini 2023

When running the latest DEV build on a Mac with an M2 (and probably M1 too) processor, there is a 40% performance drop on 80% of DOSBox Staging program starts.

Performance in the Shiny Benchmark using 10 program starts, after 1 Minute runs:

220
220
219
220
380
220
220
220
380
220

The issue happens:
-When DOSBox Staging is quit and restartet, not between restarts of the benchmark only.
-In DOSBox SVN, DOSBox-X, Staging 0.80.1
-Also when using 'glshader = none'.

The issue doesn't happen:
-On Windows using latest DEV build.
-On a Mac mini 2012.
-On the above Mac when using the Apple Rosetta compatibility layer.

Explain how to reproduce
1. 
2.
3.

https://archive.org/details/shiny-entertainment-gamers-performance-test

### Your configuration

```ini
Default config
```

### Provide a Log

```text
Last login: Tue Nov 14 17:32:02 on ttys000
/Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox ; exit;
alex@Mac-mini-von-Alexander ~ % /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox ; exit;
date       time         | 
2023-11-15 00:49:10.037 | arguments: /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox
2023-11-15 00:49:10.037 | Current dir: /Users/alex
2023-11-15 00:49:10.037 | stderr verbosity: 0
2023-11-15 00:49:10.037 | -----------------------------------
2023-11-15 00:49:10.037 | dosbox-staging version 0.81.0-alpha-1947-g3f67c91dd6
2023-11-15 00:49:10.037 | ---
2023-11-15 00:49:10.037 | LOG: Loguru version 2.1.0 initialised
2023-11-15 00:49:10.038 | CONFIG: Loaded primary config file '/Users/alex/Library/Preferences/DOSBox/dosbox-staging.conf'
2023-11-15 00:49:10.039 | LANG: Loaded language file: /Applications/DOSBox Staging.app/Contents/MacOS/../Resources/translations/de.lng
2023-11-15 00:49:10.124 | SDL: version 2.28.4 initialised (cocoa video and coreaudio audio)
2023-11-15 00:49:10.125 | DISPLAY: Initialised 1184x888 windowed mode using bilinear scaling on display-0
2023-11-15 00:49:10.145 dosbox[3298:273426] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.
2023-11-15 00:49:10.484 | OPENGL: Vendor: Apple
2023-11-15 00:49:10.484 | OPENGL: Version: 2.1 Metal - 88
2023-11-15 00:49:10.484 | OPENGL: GLSL version: 1.20
2023-11-15 00:49:10.484 | OPENGL: NPOT textures supported
2023-11-15 00:49:10.484 | RENDER: Using adaptive CRT shader based on the graphics standard of the video mode
2023-11-15 00:49:10.495 | RENDER: Auto-switched to shader 'crt/hercules'
2023-11-15 00:49:10.495 | VGA: Pixel-doubling enabled
2023-11-15 00:49:10.495 | VIDEO: Using the DOS video mode's frame rate
2023-11-15 00:49:10.498 | MEMORY: Using 4096 DOS memory pages (16 MB) at address: 0x115658000
2023-11-15 00:49:10.498 | VGA: Double-scanning VGA video modes enabled
2023-11-15 00:49:10.499 | VIDEO: Initialised S3 Trio64 VESA 2.0 compatible with 4 MB of FP DRAM supporting 86 modes
2023-11-15 00:49:10.500 | VOODOO: Initialized with 4 MB of RAM, multithreading, and no bilinear filtering
2023-11-15 00:49:10.500 | CAPTURE: Image capturer started
2023-11-15 00:49:10.524 | MIXER: Negotiated 2-channel 48000 Hz audio of 512-frame blocks
2023-11-15 00:49:10.524 | MIXER: Master compressor enabled
2023-11-15 00:49:11.114 | MIDI: Opened device: coremidi
2023-11-15 00:49:11.114 | MPU-401: Running in intelligent mode on ports 330h and 331h
2023-11-15 00:49:11.114 | OPL: Operating at 48000 Hz without resampling
2023-11-15 00:49:11.114 | OPL: Running OPL3 on ports 220h and 388h
2023-11-15 00:49:11.114 | SB16: Sound Blaster 16 OPL output filter enabled
2023-11-15 00:49:11.114 | DMA: Initialised primary controller
2023-11-15 00:49:11.114 | DMA: Initialised secondary controller
2023-11-15 00:49:11.114 | SB: Operating at 22050 Hz and upsampling to the output rate
2023-11-15 00:49:11.114 | SB16: Modern DAC output filter enabled
2023-11-15 00:49:11.114 | SB16: Setting 'BLASTER' environment variable to 'A220 I7 D1 H5 T6'
2023-11-15 00:49:11.114 | SB16: Running on port 220h, IRQ 7, DMA 1, and high DMA 5
2023-11-15 00:49:11.114 | PCSPEAKER: Operating at 48000 Hz without resampling
2023-11-15 00:49:11.114 | PCSPEAKER: Initialised discrete model
2023-11-15 00:49:11.114 | PCSPEAKER: Highpass filter enabled (18 dB/oct at 120 Hz)
2023-11-15 00:49:11.114 | PCSPEAKER: Lowpass filter enabled (12 dB/oct at 4800 Hz)
2023-11-15 00:49:11.115 | MOUSE: Will be captured after the first left or right button click
2023-11-15 00:49:11.115 | MOUSE: Middle button will capture/release the mouse (clicks not sent to the game/program)
2023-11-15 00:49:11.115 | MOUSE (PS/2): Standard, 3 buttons
2023-11-15 00:49:11.129 | RENDER: Auto-switched to shader 'crt/vga-1080p'
2023-11-15 00:49:11.130 | SDL: Using standard SDI (auto) display refresh rate of 60 Hz
2023-11-15 00:49:11.158 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz throttled VFR, scaled to 1067x800 with 1:1.35 (20:27) pixel aspect ratio
2023-11-15 00:49:11.282 | LAYOUT: Loaded codepage 850 for detected language de
2023-11-15 00:49:11.282 | SLIRP: Slirp version: 4.7.0
2023-11-15 00:49:11.282 | SLIRP: Successfully initialized
2023-11-15 00:49:11.282 | NE2000: Initialised on port 300h and IRQ 3
2023-11-15 00:49:11.283 | CONFIG: Loaded auto-mounted drive config file '/Applications/DOSBox Staging.app/Contents/MacOS/../Resources/drives/y.conf'
2023-11-15 00:49:11.283 | AUTOEXEC: Using autoexec from one or more joined sections
2023-11-15 00:49:11.290 | MAPPER: No joysticks found
2023-11-15 00:49:11.290 | MAPPER: Loaded default key bindings
2023-11-15 00:49:11.296 | MOUNT: Path '/Applications/DOSBox Staging.app/Contents/Resources/drives/y' found
2023-11-15 00:49:11.297 | LAYOUT: Loaded codepage 850 for detected language gr
2023-11-15 00:49:11.297 | MOUNT: Path '/Users/alex/Games' found, while looking for '~/games'
2023-11-15 00:49:14.500 | DOSBox has switched to max cycles, because of the setting: cycles=auto.
If the game runs too fast, try a fixed cycles amount in DOSBox's options.
2023-11-15 00:49:15.002 | RENDER: Auto-switched to shader 'interpolation/sharp'
2023-11-15 00:49:15.002 | VGA: Forcing single-scanning of double-scanned VGA video modes
2023-11-15 00:49:15.002 | VGA: Forcing no pixel-doubling
2023-11-15 00:49:15.009 | DISPLAY: VESA 640x480 256-colour graphics mode 101h at 70.007 Hz throttled VFR, scaled to 1184x888 with 1:1 (1:1) pixel aspect ratio
2023-11-15 00:50:15.682 | RENDER: Auto-switched to shader 'crt/vga-1080p'
2023-11-15 00:50:15.682 | VGA: Double-scanning VGA video modes enabled
2023-11-15 00:50:15.682 | VGA: Pixel-doubling enabled
2023-11-15 00:50:15.693 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz throttled VFR, scaled to 1067x800 with 1:1.35 (20:27) pixel aspect ratio
```

- [X] Yes, I agree.

kcgen:
# Description

Since the transition to SDL2, we've had a hard time matching upstream's `cycles = max` performance.  This PR makes a very small (3%) change in the auto-cycle metrics to improve the rate as which the auto-cycle behavior converges on the requested rate.

This only affects the automatic behavior (such as `cycles = max`) as opposed to fixed-cycle behavior (such as `cycles = 30000`).

Although I've tested it extensively, because we're close to the 0.81 release I decided to put it behind the existing `speed_mods` feature (default-enabled), which is used for performance-sensitive features that we still want to let users disable if in doubt. 

Note to reviewers: this isn't a cleanup or refactoring of the auto-cycle chunk of code, so all the existing types/casting/names/formatting is left as-is. I only wanted to touch the value itself without further changes.

This also doesn't improve the way that macOS and DOSBox interact, so there's still some time needed to quiesce prior to starting a benchmark. This seems OK though, because it's just a nuance of launching straight into a benchmark from cold-start, where as where as actual gaming sessions are unaffected.

## Related issues

Fixes #3122 

# Manual testing

Tested on:
- Raspberry Pi 400,
- x86-64 Linux at 800 MHz
- x86-64 Linux at4 GHz
- Apple M1 (Big Sur)

Specifically confirmed that this doesn't worsen auto cycle overshooting and undershooting versus the baseline behavior.

Please test with these two packs, especially if you're on macOS:

1. [:file_folder: **`perfdos.zip`**](https://github.com/dosbox-staging/dosbox-staging/files/13398108/perfdos.zip)

    - extract, `cd perfdos`, launch: `/path/to/pr/dosbox --noprimaryconf`

2. [:file_folder: **`quakes.zip`**](https://github.com/dosbox-staging/dosbox-staging/files/13398103/quakes.zip)

    - extract, `cd quakes`, launch: `/path/to/pr/dosbox --noprimaryconf`

These test packs deliberately use a small window, `outputnb`, and the `sharp` shader to minimize GL overhead and reveal the peak performance just for testing purposes.

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/fa78bd47-5a1c-47f4-8211-a05cee78e86f

For those on a Pi + RetroPie, as usual, the console framebuffer performs best when limited to 720p (this is a defect in Pi400 and below; maybe Pi5 has fixed it):

`/boo/config.txt`
```
hdmi_group=1
hdmi_mode=4
```

And because the framebuffer is vsync bound, set this in your primary conf:

```ini
[sdl]
fullscreen = true
fullresolution = original
vsync = on
```

---

@GranMinigun, @weirddan455, @FeralChild64 : curious how the above goes on your Linux systems. I've only been able to test on Ubuntu 23.04 and RetroPie.

@shermp , @kklobe, @Kappa971 : curious if you can test on any of your Windows machines.

@Grounded0 , @kklobe , @Burrito78 , @johnnovak : curious if you can test on your various Apple x86-64 and M1/M2 machines.

# Checklist

I have:

### DIFF
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -58,6 +58,7 @@ extern int32_t CPU_CycleLeft;
 extern int32_t CPU_CycleMax;
 extern int32_t CPU_OldCycleMax;
 extern int32_t CPU_CyclePercUsed;
+constexpr int32_t CPU_CycleProtectedModeDefault = 50000;
 extern int32_t CPU_CycleLimit;
 extern int64_t CPU_IODelayRemoved;
 extern bool CPU_CycleAutoAdjust;
--- a/meson.build
+++ b/meson.build
@@ -161,6 +161,16 @@ if get_option('buildtype') in ['release', 'minsize']
         '-fdata-sections',
     ]
     extra_link_flags += ['-Wl,--gc-sections']
+
+    # Apply Apple silicon-specific CPU tunings
+    if (os_family_name == 'MACOSX'
+        and host_machine.cpu_family() == 'aarch64')
+        extra_flags += [
+            '-mcpu=native',
+            '-mtune=native',
+        ]
+    endif
+
 endif
 
 # Let sanitizer builds recover and continue
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -1583,6 +1583,8 @@ void CPU_SET_CRX(Bitu cr,Bitu value) {
 				CPU_CycleLeft       = 0;
 				CPU_Cycles          = 0;
 				CPU_OldCycleMax     = CPU_CycleMax;
+				CPU_CycleMax        = std::max(CPU_CycleMax,
+				                               CPU_CycleProtectedModeDefault);
 
 				GFX_SetTitle(CPU_CyclePercUsed);
 				if (!printed_cycles_auto_info) {
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -1562,52 +1562,58 @@ Bitu CPU_SIDT_limit(void) {
 static bool printed_cycles_auto_info = false;
 void CPU_SET_CRX(Bitu cr,Bitu value) {
 	switch (cr) {
-	case 0:
-		{
-			value|=CR0_FPUPRESENT;
-			Bitu changed=cpu.cr0 ^ value;
-			if (!changed) return;
-			cpu.cr0=value;
-			if (value & CR0_PROTECTION) {
-				cpu.pmode=true;
-				LOG(LOG_CPU,LOG_NORMAL)("Protected mode");
-				PAGING_Enable((value & CR0_PAGING)>0);
-
-				if (!(CPU_AutoDetermineMode&CPU_AUTODETERMINE_MASK)) break;
-
-				if (CPU_AutoDetermineMode&CPU_AUTODETERMINE_CYCLES) {
-					CPU_CycleAutoAdjust=true;
-					CPU_CycleLeft=0;
-					CPU_Cycles=0;
-					CPU_OldCycleMax=CPU_CycleMax;
-				        GFX_SetTitle(CPU_CyclePercUsed);
-				        if(!printed_cycles_auto_info) {
-						printed_cycles_auto_info = true;
-						LOG_MSG("DOSBox has switched to max cycles, because of the setting: cycles=auto.\nIf the game runs too fast, try a fixed cycles amount in DOSBox's options.");
-					}
-				} else {
-				        GFX_RefreshTitle();
-			        }
-#if (C_DYNAMIC_X86)
-				if (CPU_AutoDetermineMode&CPU_AUTODETERMINE_CORE) {
-					CPU_Core_Dyn_X86_Cache_Init(true);
-					cpudecoder=&CPU_Core_Dyn_X86_Run;
-				}
-#elif (C_DYNREC)
-				if (CPU_AutoDetermineMode&CPU_AUTODETERMINE_CORE) {
-					CPU_Core_Dynrec_Cache_Init(true);
-					cpudecoder=&CPU_Core_Dynrec_Run;
-				}
-#endif
-				CPU_AutoDetermineMode<<=CPU_AUTODETERMINE_SHIFT;
-			} else {
-				cpu.pmode=false;
-				if (value & CR0_PAGING) LOG_MSG("Paging requested without PE=1");
-				PAGING_Enable(false);
-				LOG(LOG_CPU,LOG_NORMAL)("Real mode");
-			}
-			break;
+	case 0: {
+		value |= CR0_FPUPRESENT;
+		Bitu changed = cpu.cr0 ^ value;
+		if (!changed) {
+			return;
 		}
+		cpu.cr0 = value;
+		if (value & CR0_PROTECTION) {
+			cpu.pmode = true;
+			LOG(LOG_CPU, LOG_NORMAL)("Protected mode");
+			PAGING_Enable((value & CR0_PAGING) > 0);
+
+			if (!(CPU_AutoDetermineMode & CPU_AUTODETERMINE_MASK)) {
+				break;
+			}
+
+			if (CPU_AutoDetermineMode & CPU_AUTODETERMINE_CYCLES) {
+				CPU_CycleAutoAdjust = true;
+				CPU_CycleLeft       = 0;
+				CPU_Cycles          = 0;
+				CPU_OldCycleMax     = CPU_CycleMax;
+
+				GFX_SetTitle(CPU_CyclePercUsed);
+				if (!printed_cycles_auto_info) {
+					printed_cycles_auto_info = true;
+					LOG_MSG("DOSBox has switched to max cycles, because of the setting: cycles=auto.\nIf the game runs too fast, try a fixed cycles amount in DOSBox's options.");
+				}
+			} else {
+				GFX_RefreshTitle();
+			}
+#if (C_DYNAMIC_X86)
+			if (CPU_AutoDetermineMode & CPU_AUTODETERMINE_CORE) {
+				CPU_Core_Dyn_X86_Cache_Init(true);
+				cpudecoder = &CPU_Core_Dyn_X86_Run;
+			}
+#elif (C_DYNREC)
+			if (CPU_AutoDetermineMode & CPU_AUTODETERMINE_CORE) {
+				CPU_Core_Dynrec_Cache_Init(true);
+				cpudecoder = &CPU_Core_Dynrec_Run;
+			}
+#endif
+			CPU_AutoDetermineMode <<= CPU_AUTODETERMINE_SHIFT;
+		} else {
+			cpu.pmode = false;
+			if (value & CR0_PAGING) {
+				LOG_MSG("Paging requested without PE=1");
+			}
+			PAGING_Enable(false);
+			LOG(LOG_CPU, LOG_NORMAL)("Real mode");
+		}
+		break;
+	}
 	case 2:
 		paging.cr2=value;
 		break;
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -1584,7 +1584,7 @@ void CPU_SET_CRX(Bitu cr,Bitu value) {
 				CPU_Cycles          = 0;
 				CPU_OldCycleMax     = CPU_CycleMax;
 				CPU_CycleMax        = std::max(CPU_CycleMax,
-				                               CPU_CycleProtectedModeDefault);
+                                                        CPU_CycleProtectedModeDefault);
 
 				GFX_SetTitle(CPU_CyclePercUsed);
 				if (!printed_cycles_auto_info) {
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -414,6 +414,17 @@ static void DOSBOX_RealInit(Section* sec)
 	ticksRemain = 0;
 	ticksLast   = GetTicks();
 	ticksLocked = false;
+
+	// Our assumption is that the auto cycle adjustment will not cause
+	// regressions, however if it is suspected, users can disable speed_mods
+	// and see if it helps. The speed_mods description asks user to report
+	// cases to the team for further assessment. Like prior speed_mods,
+	// after a extended period without negative reports, the feature will be
+	// made permanent.
+	//
+	auto_cycle_adjusted_scalar = adjust_auto_cycle_scalar(
+	        section->Get_bool("speed_mods"));
+
 	DOSBOX_SetLoop(&Normal_Loop);
 
 	MAPPER_AddHandler(DOSBOX_UnlockSpeed, SDL_SCANCODE_F12, MMOD2, "speedlock", "Speedlock");
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -146,6 +146,9 @@ static int16_t auto_cycle_adjusted_scalar = AutoCycleRatioScalar;
 static constexpr int16_t AutoCycleMaxTicksScheduled = 250;
 static constexpr int8_t AutoCycleMinTicksScheduled  = 5;
 
+static constexpr int16_t AutoCycleMaxTicksDone = AutoCycleMaxTicksScheduled;
+static constexpr int8_t AutoCycleMinTicksDone  = 2 * AutoCycleMinTicksScheduled;
+
 // General CPU tick constants
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~
 static constexpr int8_t MaxScheduledTicks = 20;
@@ -254,9 +257,12 @@ void increaseticks() { //Make it return ticksRemain and set it in the function a
 	// Is the system in auto cycle mode guessing ? If not just exit. (It can be temporary disabled)
 	if (!CPU_CycleAutoAdjust) return;
 
-	if (ticksScheduled >= AutoCycleMaxTicksScheduled || ticksDone >= 250 ||
+	if (ticksScheduled >= AutoCycleMaxTicksScheduled ||
+	    ticksDone >= AutoCycleMaxTicksDone ||
 	    (ticksAdded > 15 && ticksScheduled >= AutoCycleMinTicksScheduled)) {
-		if(ticksDone < 1) ticksDone = 1; // Protect against div by zero
+		if (ticksDone < 1) {
+			ticksDone = 1; // Protect against div by zero
+		}
 		int32_t ratio = static_cast<int32_t>(
 		        (ticksScheduled * (CPU_CyclePercUsed * auto_cycle_adjusted_scalar / 100)) /
 		        ticksDone);
@@ -275,14 +281,15 @@ void increaseticks() { //Make it return ticksRemain and set it in the function a
 				/* Don't allow very high ratio which can cause us to lock as we don't scale down
 				 * for very low ratios. High ratio might result because of timing resolution */
 				if (ticksScheduled >= AutoCycleMaxTicksScheduled &&
-				    ticksDone < 10 && ratio > 16384) {
+				    ticksDone < AutoCycleMinTicksDone &&
+				    ratio > 16384) {
 					ratio = 16384;
 				}
 
 				// Limit the ratio even more when the cycles are
 				// already about the protected mode default
 				if (ticksScheduled >= AutoCycleMaxTicksScheduled &&
-				    ticksDone < 10 && ratio > 5120 &&
+				    ticksDone < AutoCycleMinTicksDone && ratio > 5120 &&
 				    CPU_CycleMax > CPU_CycleProtectedModeDefault) {
 					ratio = 5120;
 				}
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -143,17 +143,16 @@ static constexpr int8_t AutoCycleRatioBoost = 24;
 // Either the auto-cycle ratio scalar or with added boost
 static int16_t auto_cycle_adjusted_scalar = AutoCycleRatioScalar;
 
-static constexpr int16_t AutoCycleMaxTicksScheduled = 250;
-static constexpr int8_t AutoCycleMinTicksScheduled  = 5;
-
-static constexpr int16_t AutoCycleMaxTicksDone = AutoCycleMaxTicksScheduled;
+static constexpr int8_t AutoCycleMinTicksScheduled = 5;
 static constexpr int8_t AutoCycleMinTicksDone  = 2 * AutoCycleMinTicksScheduled;
-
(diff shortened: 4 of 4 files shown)


Burrito78:
Thanks @kcgen for packing up these testing packs, very nice!

The situation is similar for me though. The "high" runs are way more now but I still get the occasional "low" run from time to time. I ran both benchmarks using the PR (alpha-1975) 10 times, see results below:

perfdos
412
405
**238**
408
403
407
**230**
408
**242**
410

quakes
197
197
196
196
194
199
197
197
**137**
195

For comparison, here are the "quakes" scores using main (alpha-1952):
**127**
184
183
**128**
182
183
**129**
182
183
**129**

kklobe:
I'm still seeing about the same split of slow/fast, but the numbers have definitely moved north - 127/186 vs 99/156! Very similar to what I've seen with my own tinkering.

weirddan455:
Tested on x86_64 Arch Linux.  I don't trust the results I'm getting from perfdos.  The number keeps slowly increasing the longer I leave it up.

This branch is a solid perf improvement in Quake though.  Ran it several times and went from an average of 131 FPS in main to 152 FPS in this branch.

Note that I've never experienced the issue described on #3122.  Every time I've ran benchmarks in Dosbox, it's been fairly consistent from run to run.  In Quake, there was no more 5 FPS deviation from run to run when benchmarking the same branch.

FeralChild64:
@kcgen

> curious how the above goes on your Linux systems

Linux, AMD64, Quake 1, `timedemo demo1`, 320x200, with sound effects, no CD audio, FPS values (higher = better):

| | before  | after |
| ------------- | ------------- | ------------- |
|| | |
|| 70.7 | 83.4 |
|| 73.6 | 78.8 |
|| 72.6 | 83.5 |
|| 72.8 | 81.3 |
|| 71.0 | 78.9 |
|| 71.0 | 82.9 |
|| 74.0 | 76.9 |
|| 72.9 | 83.0 |
|| 75.1 | 76.5 |
|| 72.3 | 78.1 |
|| | |
|average| 72.6 | 80.3 |
|std. dev. | 1.42 | 2.79 |

Conclusion: slightly less stable FPS, but +10% performance. Nice!

Grounded0:
My PERFDOS score went from 350 to 370 (several minute average).

kcgen:
Thanks for all the testing! 

Good to see the results are in line with local testing; we can probably extrapolate them to most hardware out there.

Ready for round 2! This makes a couple small changes to the overage behavior.

kcgen:
> I have just noticed, that 360x240 gives about 4 times better results than 320x200

@FeralChild64 , suggest using the 'quakes.zip' attached in the PR body above, at least for these tests. (I would assume your Reyzen system should get near 200 fps in 320x200 Quake timedemo).

FeralChild64:
@kcgen I've switched from AMD Ryzen to Intel Core - but I have limited it's power consumption in BIOS to avoid fan noise, so my results might not be relevant to this CPU model in general.

I prefer to test games in environment I'm normally using. And I'm getting over 280 FPS in 360x240 mode.

weirddan455:
Tested on Windows 10.  I dual boot so same machine as my Linux test although this testing was done after you force-pushed new changes to the branch.

Quake:
Main -> 140 FPS average
This PR -> 155 FPS average

Quake seems to perform slightly better on Windows but the numbers are very similar to Linux.  Still not seeing any runs with large FPS drops.

kcgen:
@FeralChild64, sounds good - and wow, that must be a fast system to crank out 280 fps in 360x240!

@weirddan455 - thanks for confirming; as long as we're not seeing regressions then that's perfect.

Linux and Windows systems haven't been affected but this issue (well, maybe to a tiny degree, but usually Quake benchmarks vary to a much smaller degree than on macOS).

The most sensitive system is the Raspberry Pi, where aggressive settings  produce audio gaps in demanding games like Quake and Screamer 2, and so far this PRs is delivering a couple more FPS without introducing audio gaps.

@Burrito78 , @kklobe - the last update helps even more on macOS get up to speed from a cold start launching straight into the benchmarks.

I've repacked the two benchmarks, if you want to try:

1. [:file_folder: **`perfdos.zip`**](https://github.com/dosbox-staging/dosbox-staging/files/13398108/perfdos.zip)

    - extract, `cd perfdos`, launch: `/path/to/pr/dosbox --noprimaryconf`

2. [:file_folder: **`quakes.zip`**](https://github.com/dosbox-staging/dosbox-staging/files/13398103/quakes.zip)

    - extract, `cd quakes`, launch: `/path/to/pr/dosbox --noprimaryconf`

Grounded0:
I'm running Quake Time Demo now in my preferred setting of 640x480 to model real world performance requirements and i'm getting equivalent or better performance vs gold standard 0.80.1 so this is helluva improvement. 👍

Grounded0 (approved):

Burrito78:
Just tested Quake this time using your new bundle.
This is on the M2 MacBook Air, plugged in and no other apps open.

I did 30 runs because this benchmark is so fast to do and to have a better sample size.

3 runs between 210 and 211fps.
19 runs in the range of 196-199fps.
8 had between 141 and 149fps.

So I'm now seeing "high", "mid" and "low" fps runs.

Grounded0:
Download DOSBENCH:

https://anttipeltola.eu/misc/files/DOSBENCH.ZIP

And do Quake Time Demo 640x480 runs to get less variation between benchmark runs.

Burrito78:
> @Burrito78
> 
> Download DOSBENCH:
> 
> https://anttipeltola.eu/misc/files/DOSBENCH.ZIP
> 
> And do Quake Time Demo 640x480 runs to get less variation between benchmark runs as QTD runs too quickly at lower resolutions and thus introduces too much variation for my liking. Apple silicon GPU is powerful enough not to interfere with the results.

With DOSBench 640x480 and default config doing 10 runs I'm getting:

8x 47-50fps
2x 75fps

So the same situation as in main for me (80% of runs have 40% reduced performance).

Grounded0:
So this has similar level of variation but with better performance.

Grounded0:
Personally all my benchmarks are landing in the 70-75 fps zone. 👍

The best and most consistent performance yet.

2017 15" MacBook Pro with macOS 12.6.5.

kcgen:
@Burrito78 , I've turned on some extra logging. To capture:

/Path/to/pr/dosbox &> log.txt

Quit as soon you get the FPS results, then rename the log to indicate which bucket it falls into: fast-1, mid-1, slow-1, fast-2, etc..

Fully shutdown all other applications for this to avoid random cpu conflicts. 

Just need two or three logs from each. 

If you run 320 vs 640, please name those too just to help, as the number will look different.

Grounded0:
Yeah good call to check for thermal throttle. This rig can run with 70 W TDP all day. Big laptops FTW.

Grounded0:
Using max 105% might also tell the CPU scheduler that we're not messing around and start giving out those peak boost clocks.

kklobe:
I don't think throttling is the issue here, at least not enough to explain such a wide variance.

Here are the CPU frequencies of the P-Cores for slow and fast run on my M1:

slowQTD
fastQTD

Burrito78:
@kcgen Did the quake benchmark again using newest PR and your package (320x200) with --noprimaryconf and wrote the log files.

[quakes_logs.zip](https://github.com/dosbox-staging/dosbox-staging/files/13400627/quakes_logs.zip)

kklobe:
To expand a bit more on my hunch (after looking at this issue on and off for a while now), is that we're dealing with non-deterministic cache line/memory alignment and/or hit/miss ratio due to dynrec's JIT. Every time a code page gets JITted, I have to invalidate the cache for that memory range, since caches are not coherent on ARM like they are on Intel. I believe the cache line size on Apple Silicon is 128 bytes, but I have tried changing the 32-byte aligning in `dyn_cache.h` to 128 and that didn't make any difference. It seems like there are runs where the memory/cache alignment is just perfect and those are the high ones, although I don't suspect this magnitude of difference would be from memory alignment; cache misses seems more likely.

I haven't been able to find any utilities for Apple Silicon that will give processor cache hit/miss statistics; I know you used to be able to allow kernel extensions and run Intel's performance counter tools. If anyone knows of any, please jump in.

weirddan455:
Just want to confirm, it seems that everyone who is having this issue is on an ARM system?  @Grounded0 said he has no issues but I believe that 2017 Macbook is Intel correct?

If the fix is simply setting the number of cycles to a specific number, would it be worth hard-coding the value that works best on Mac M1/M2 and Raspberri Pi (detect if the user is on one of those systems and special case it for "max")?  I don't love that solution but maybe it ends up working best in practice (at least for 0.81).

kklobe:
> Just want to confirm, it seems that everyone who is having this issue is on an ARM system? @Grounded0 said he has no issues but I believe that 2017 Macbook is Intel correct?
> 
> If the fix is simply setting the number of cycles to a specific number, would it be worth hard-coding the value that works best on Mac M1/M2 and Raspberri Pi (detect if the user is on one of those systems and special case it for "max")? I don't love that solution but maybe it ends up working best in practice (at least for 0.81).

I don't think there's going to be a single magic number, as it varies so much between runs on ARM with dynamic core. As I mentioned above, when I use cycle guessing with normal core, I get almost no variation.

kklobe:
Also, remember to test on things with audio; when I made the auto-cycling too aggressive by just running QTD and PERFDOS, I'd notice audio pops and clicks in regular Quake/Doom/Blood.

Grounded0:
QTD duration marked between the red lines with result 70.5 FPS. Bear in mind that i have 7 virtual desktops of stuff running in the background and no. I'm not shutting them down for benchmarks. 😁

Screenshot 2023-11-18 at 19 40 25

Grounded0:
No audio skip with Blood with this and cpu max 105%. Really nice.

kcgen:
Trying for another set of changes that increase the tempo of the auto-cycle assessment. 

Will follow up when it's ready to test.

@kklobe, I wonder if there are compiler alignment flags we can pass the page size value (although, I suspect the compiler already know of these details). 

Another thought is using something like `-march=native` or `-mtune=native` specifically for macos arm64 builds, which should include all the processor-specific alignment settings?

kcgen:
@Grounded0 , thanks for those charts. Looks like your system isn't breaking a sweat 😅

kklobe:
> Trying for another set of changes that increase the tempo of the auto-cycle assessment.
> 
> Will follow up when it's ready to test.
> 
> @kklobe, I wonder if there are compiler alignment flags we can pass the page size value (although, I suspect the compiler already know of these details).
> 
> Another thought is using something like `-march=native` or `-mtune=native` specifically for macos arm64 builds, which should include all the processor-specific alignment settings?

I don't think that's the problem. We have a few variables here: Core, OS/Scheduler, CPU, and autocycle algorithm. From this ticket, and my experience as well, people are only seeing this on an ARM system with the dynamic core. It doesn't happen on Rosetta with an x86 build on Apple Silicon. I cannot get it to happen on normal core with an ARM build (don't know if anyone else can confirm). My graph above makes CPU frequency throttling unlikely, and the fact that autocycling works fine on dyn-x86 and normal on ARM leads me to believe it's not an OS/scheduler thing or an intrinsic ARM thing.

I'm trying to get hard data to get to the root of this and will keep plugging away - suggestions appreciated! It may be time to do deep with Tracy in the dynamic core and compare slow with fast runs.

Grounded0:
Did some Blood 105% running with

main: Zero audio skips

this branch few commits back: Extremely few skips probably due to ton of stuff running in the background.

kklobe:
I get a few very small audio glitches in Blood with main, and far more with this branch.

Grounded0:
I think there have been talks of taking all the DAC audio stuff into a separate thread since its the most sensitive. If people get more skips then maybe this aggressive setting can be used after that work is done?

weirddan455:
> I think there have been talks of taking all the PCM audio stuff into a separate thread since its the most sensitive. If people get more audio skips then maybe this aggressive setting can be used after that work is done?
> 
> There's your rationale for anyone taking this on: More perf.

I had mentioned that but it's more of a long term goal.  I'm not even going to look at it seriously until after 0.81.  CD audio should be fairly easy to multi-thread since it's basically "fire and forget" (game issues a command to play track 5 and you just keep on playing until it tells you otherwise).

OPL and MIDI are likely to be tricky since they need to be tightly synchronized to the emulation core.  Multi-threading isn't necessarily a magic fix for audio stuttering and it will mean writing more complex code.

Grounded0:
I think a better change would be to make cpu max 100% actually close to cpu max 100% to leave little performance on the table and still leave this scheduler on its fairly conservative settings until PCM audio is moved off the main thread.

This thread load setting belongs to year 2004 without multicore CPUs.

johnnovak:
> OPL and MIDI are likely to be tricky since they need to be tightly synchronized to the emulation core. Multi-threading isn't necessarily a magic fix for audio stuttering and it will mean writing more complex code.

MIDI synths already run in their own threads (FluidSynth and MT-32). If you mean core MIDI message processing, that's small change, doesn't matter.

But the rest could be moved to threads although they're not exactly doing heavy calculations. I think audio choking is not because these synth are particularly heavy, but because cycles=max just gobbles up _everything_. Anyway, needs experimentation; it's hard to reason about it beyond this. Also, I suspect disk I/O happening in the main thread can have some effect on the audio as well by causing micro-dropouts when the game loads large files at max host speeds (DOS era disk speeds were a lot slower)

johnnovak:
> My graph above makes CPU frequency throttling unlikely, and the fact that autocycling works fine on dyn-x86 and normal on ARM leads me to believe it's not an OS/scheduler thing or an intrinsic ARM thing.
> 
> I'm trying to get hard data to get to the root of this and will keep plugging away - suggestions appreciated! It may be time to do deep with Tracy in the dynamic core and compare slow with fast runs.

I think @kklobe is on the money. I think these problems are the result of an autocycler algo bug or edge case being hit with the dynamic core on ARM due to the pecularities of the ARM JIT and nothing else (well, maybe macOS specific syscall differences like subtly different timer behaviour could play a role too).

johnnovak:
Lemme know when this is ready for another round of testing @kcgen. Looks like I've missed the first round, but I'm here for the next one 😎 🤘🏻

kcgen:
Closing this one down. Despite frame rates being ~70% better on the Pi4, I can hear slightly more audio gaps in CPU-intensive games - so it's not the right approach.

Issue #3133, opened 2023-11-17
A few `[mouse]` settings need some love


johnnovak:
`mouse_sensitivity` in particular:

- Modifying the setting at runtime has no effect; it can be only set from the config at startup. This is something super important to configure at runtime in the upcoming OSD.
- Nothing is logged when using invalid settings (e.g., `100 150` instead of `100, 150`, or `asdf`).
- We should support the comma-less syntax which is consistent with other compound config values (the comma should be still optional for backward compatibility reasons).
- Percentage signs should be optionally accepted for consistency with other config settings (e.g., `150%` and `100% 150%` should be valid). (Related: https://github.com/dosbox-staging/dosbox-staging/issues/3132)

The rest of the`[mouse]` settings should also be reviewed and enhanced as needed to support changing values at runtime (`mouse_capture`, `mouse_middle_release`, and `mouse_multi_display_aware` in particular; the rest are probably fine to only set at startup as they are hardware related).

FeralChild64:
Regarding `mouse_sensitivity`:

> Modifying the setting at runtime has no effect; it can be only set from the config at startup.

I have made it only-at-start, because I have no idea how to define a sane behavior when it is being changed at the runtime.
What's more, the whole config API does not tell me which value was touched by the user, so if he executes `config -set mouse_sensitivity=120` when it already was 120, I won't be able to tell whether he touched `mouse_sensitivity` or some other setting (in fact, I'm not even sure if I'll get the notification at all if the supplied string is the same, I didn't test it).

If you can define some sane behavior, let me know.

> This is something super important to configure at runtime in the upcoming OSD.

Just use the same API the `MOUSECTL.COM` command uses - see `program_mousectl.cpp` file, `MOUSECTL::CmdSensitivity` function. Provide an empty `ListIDs &list_ids` parameter to change the setting for all the mouse interfaces, or specify which ones would you like to affect (like only DOS driver or only PS/2 mouse).

> Nothing is logged when using invalid settings

True - this is one of the drawbacks of the current configuration API. The code to retrieve the values is:

```C++
	PropMultiVal *prop_multi = conf->GetMultiVal("mouse_sensitivity");
	mouse_config.sensitivity_x = static_cast<int16_t>(
	        prop_multi->GetSection()->Get_int("xsens"));
	mouse_config.sensitivity_y = static_cast<int16_t>(
	        prop_multi->GetSection()->Get_int("ysens"));
```

and it does not return back any error information to me.

[...]

PR #3138, opened 2023-11-19
Improve `mouse_sensitivity` behavior

FeralChild64:
# Description

Implemented changes requested by @johnnovak:
- `mouse_sensitivity` can now be changed at runtime; to achieve this, sensitivity settings from `mousectl` tool (which can be set differently per mouse interface) now work independently, as a second layer of scaling
- improved `mouse_sensitivity` argument parsing for better error detection

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3133

# Manual testing

- test that `mousectl -s <value>` command and `mouse_sensitivity` setting still work
- try different legal and illegal values of `mouse_sensitivity` in config file; values out of range should be cropped, any other error should cause the defaults to be applied; all errors should be logged

# Checklist

I have:

### DIFF
--- a/src/dos/program_mousectl.cpp
+++ b/src/dos/program_mousectl.cpp
@@ -544,10 +544,12 @@ void MOUSECTL::AddMessages()
 	        "  -s -sx -sy     sets sensitivity / for x axis / for y axis, from -999 to +999\n"
 	        "  -r             sets minimum mouse sampling rate\n"
 	        "  -on -off       enables or disables mouse on the given interface\n"
-	        "  -reset         restores all mouse settings from the configuration file\n"
+	        "  -reset         restores default mouse settings\n"
 	        "\n"
 	        "Notes:\n"
 	        "  If sensitivity or rate is omitted, it is reset to default value.\n"
+	        "  Mouse sensitivity set in the configuration file acts as global scale factor,\n"
+	        "  per-interface sensitivity set by this commands works on top of that.\n"
 	        "\n"
 	        "Examples:\n"
 	        "  [color=light-green]mousectl[reset] [color=white]DOS[reset] [color=white]COM1[reset] -map    ; asks user to select mice for a two player game");
--- a/src/hardware/input/mouse_config.cpp
+++ b/src/hardware/input/mouse_config.cpp
@@ -105,78 +105,151 @@ static const std::vector<uint16_t> list_rates = {
         // issues.
 };
 
-bool MouseConfig::ParseCaptureType(const std::string_view capture_str,
-                                   MouseCapture& capture)
+const std::vector<uint16_t>& MouseConfig::GetValidMinRateList()
 {
-	if (capture_str == capture_type_seamless_str)
-		capture = MouseCapture::Seamless;
-	else if (capture_str == capture_type_onclick_str)
-		capture = MouseCapture::OnClick;
-	else if (capture_str == capture_type_onstart_str)
-		capture = MouseCapture::OnStart;
-	else if (capture_str == capture_type_nomouse_str)
-		capture = MouseCapture::NoMouse;
-	else
-		return false;
-	return true;
+	return list_rates;
 }
 
-bool MouseConfig::ParseCOMModel(const std::string_view model_str,
+bool MouseConfig::ParseComModel(const std::string_view model_str,
                                 MouseModelCOM& model, bool& auto_msm)
 {
 	if (model_str == model_com_2button_str) {
 		model    = MouseModelCOM::Microsoft;
 		auto_msm = false;
-		return true;
 	} else if (model_str == model_com_3button_str) {
 		model    = MouseModelCOM::Logitech;
 		auto_msm = false;
-		return true;
 	} else if (model_str == model_com_wheel_str) {
 		model    = MouseModelCOM::Wheel;
 		auto_msm = false;
-		return true;
 	} else if (model_str == model_com_msm_str) {
 		model    = MouseModelCOM::MouseSystems;
 		auto_msm = false;
-		return true;
 	} else if (model_str == model_com_2button_msm_str) {
 		model    = MouseModelCOM::Microsoft;
 		auto_msm = true;
-		return true;
 	} else if (model_str == model_com_3button_msm_str) {
 		model    = MouseModelCOM::Logitech;
 		auto_msm = true;
-		return true;
 	} else if (model_str == model_com_wheel_msm_str) {
 		model    = MouseModelCOM::Wheel;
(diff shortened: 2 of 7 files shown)


src/hardware/input/mouse_config.cpp
```diff
@@ -127,56 +116,132 @@ bool MouseConfig::ParseCOMModel(const std::string_view model_str,
...
+		mouse_config.model_ps2 = MouseModelPS2::NoMouse;
 	} else {
-		return false;
+		assert(false);
 	}
-	return true;
 }
 
-const std::vector<uint16_t> &MouseConfig::GetValidMinRateList()
+static void SetCOMModel(const std::string_view model_str)
```
johnnovak, line 174: Preexisting:
MouseModelPs2, SetComModel etc. as per our naming convention

src/hardware/input/mouse_config.cpp
```diff
@@ -127,56 +116,132 @@ bool MouseConfig::ParseCOMModel(const std::string_view model_str,
...
+	const int value_max = mouse_predefined.sensitivity_user_max;
+	for (auto& value_str : values_str) {
+		// Remove trailing '%' signs, if present
+		if (ends_with(value_str, "%")) {
+			value_str.pop_back();
+		}
+
+		const auto value = parse_int(value_str);
+		if (!value) {
+			LOG_WARNING("CONFIG: Syntax error in 'mouse_sensitivity', using defaults");
```
johnnovak, line 216: For consistency with other error messages, let's say:

Invalid 'mouse_sensitivity' setting, using defaults

src/hardware/input/mouse_config.cpp
```diff
@@ -127,56 +116,132 @@ bool MouseConfig::ParseCOMModel(const std::string_view model_str,
...
+
+		const auto value_clamped = clamp(*value, value_min, value_max);
+		if (value_clamped != *value) {
+			out_of_range = true;
+		}
+		values_int.push_back(value_clamped);
+	}
+
+	if (out_of_range) {
+		LOG_WARNING("CONFIG: 'mouse_sensitivity' adjusted to range %+d - %+d",
```
johnnovak, line 228: Let's use the `MOUSE: ` prefix in all warnings.

src/hardware/input/mouse_config.cpp
```diff
@@ -127,56 +116,132 @@ bool MouseConfig::ParseCOMModel(const std::string_view model_str,
...
+
+static void SetSensitivity(const std::string_view sensitivity_str)
+{
+	// Coefficient to convert percentage in integer to float
+	constexpr float coeff = 0.01f;
+
+	// Split input string into values
+	auto values_str = split(sensitivity_str, " \t,;");
+	if (values_str.size() > 2) {
+		LOG_WARNING("CONFIG: Too many values in 'mouse_sensitivity', using defaults");
```
johnnovak, line 189: `using '100,100'` for consistency

src/hardware/input/mouse_config.cpp
```diff
@@ -302,21 +353,11 @@ static void config_init(Section_prop &secprop)
 	                    "      using mirrored display mode or using an AV receiver's HDMI input for\n"
 	                    "      audio-only listening.");
 
-	prop_multi = secprop.AddMultiVal("mouse_sensitivity", only_at_start, ",");
-	prop_multi->Set_help(
+	prop_str = secprop.Add_string("mouse_sensitivity", always, "100,100");
+	prop_str->Set_help(
 	        "Mouse sensitivity for the horizontal and vertical axes, as a percentage\n"
```
johnnovak, line 358: I'd say "Global mouse sensitivity" at the start, then I'd suggesting appending this at the end:

"Mouse sensitivity can be further fine-tuned per mouse interface using the internal MOUSECTL.COM tool available on the Z drive."

johnnovak:
Great job @FeralChild64, it works really well! I left some comments mainly around improving the help text and the warnings, but the code itself looks fine.

src/dos/program_mousectl.cpp
```diff
@@ -544,10 +544,11 @@ void MOUSECTL::AddMessages()
 	        "  -s -sx -sy     sets sensitivity / for x axis / for y axis, from -999 to +999\n"
 	        "  -r             sets minimum mouse sampling rate\n"
 	        "  -on -off       enables or disables mouse on the given interface\n"
-	        "  -reset         restores all mouse settings from the configuration file\n"
+	        "  -reset         restores default mouse settings\n"
 	        "\n"
 	        "Notes:\n"
 	        "  If sensitivity or rate is omitted, it is reset to default value.\n"
+	        "  Sensitivity set in the configuration file works independently.\n"
```
johnnovak, line 551: I think this needs more clarification. We should make it clear here that the `mouse_sensitivity` settings acts as a global scale factor, and then per mouse interface sensitivity settings are applied on top of that.

FeralChild64: OK, added more information.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, added more information. Also changed default to '100', as it is currently.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, changed.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, changed - but the default in log is taken from the constant and actually applied.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, changed - but the default in log is taken from the constant and actually applied.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, changed.

src/hardware/input/mouse_config.cpp
johnnovak: That's fine, I'm approaching these messages from the perspective of the regular user who will only have access to the config descriptions and the user manual. So if the config description say "100,100 is the default", I think the most useful and least confusing thing to say in the logs is "using 100,100" (instead of "using defaults" or "using X" where X is the actual value). The idea is thst we inform the user that an invalid setting was provided, therefore we operate now as if the default "100,100" setting wasprovided manually. I find that very clear.

johnnovak (approved):
It's perfect now @FeralChild64 🎉 Big cheers for jumping on this so quickly! 🥇

PR #3139, opened 2023-11-19
Regenerate primary config file if it is empty

weirddan455:
# Description
This PR simply checks if the found config file is empty (or contains only whitespace).  If so, re-create the config file as if it were not there.

There was some discussion about how best to enable portable mode that happened on a closed PR #2966

Flagging everyone involved in that discussion as a reviewer.

@Torinde it would not let me add you as a reviewer for some reason but feel free to comment here.

## Related issues
#2966

# Manual testing
Portable mode can be enabled simply by `touch /path/to/dosbox/dosbox-staging.conf` (or similarly creating an empty file with Notepad or something).  This worked before but left  the user with an empty config file.  Now, the config file gets regenerated with default settings.

Tested that no behavior changes on finding a non-empty config file or non-existent config file (configfile if created if it doesn't exist and left alone if it is non-empty).  Tested various whitespace characters and confirmed an all white-space containing file is treated as empty.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4692,7 +4692,7 @@ int sdl_main(int argc, char* argv[])
 		if (!arguments->securemode && !arguments->noprimaryconf) {
 			const auto primary_config_path = GetPrimaryConfigPath();
 
-			if (!path_exists(primary_config_path)) {
+			if (!ConfigFileIsValid(primary_config_path)) {
 				// No config is loaded at this point, so we're
 				// writing the default settings to the primary
 				// config.
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4692,7 +4692,7 @@ int sdl_main(int argc, char* argv[])
 		if (!arguments->securemode && !arguments->noprimaryconf) {
 			const auto primary_config_path = GetPrimaryConfigPath();
 
-			if (!ConfigFileIsValid(primary_config_path)) {
+			if (!config_file_is_valid(primary_config_path)) {
 				// No config is loaded at this point, so we're
 				// writing the default settings to the primary
 				// config.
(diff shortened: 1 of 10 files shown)


src/gui/sdlmain.cpp
```diff
@@ -4616,6 +4616,31 @@ static void override_wm_class()
...
+			if (!isspace(buffer[i])) {
+				fclose(file);
+				return true;
+			}
+		}
+	} while (bytes_read == BufferSize);
+
+	fclose(file);
+	return false;
+}
```
johnnovak, line 4642: That's a nice optimal way of checking if the file is empty without rewriting the config parser (which will be refactored anyway) 🎉 

My only suggestion is to move it to `setup.cpp` (and upper camel case the name) as it doesn't really belong in sdlmain.

johnnovak (approved):
Looks good @weirddan455 , one minor comment.

src/gui/sdlmain.cpp
weirddan455: Moved and renamed as you suggested.  I am curious why non-member functions in that file use `MicrosoftCase` as opposed to `snake_case` though.  I thought `snake_case` was our coding style for stand alone functions.

johnnovak: To be honest I wasn't thinking about it too deeply, just following the conventions in `setup.cpp` 😄  I guess all functions in `setup.h` would need to be prefixed with `SETUP_`; maybe you do wanna make adjustment quickly too 😄 

```cpp
std_fs::path GetPrimaryConfigPath();
void ParseConfigFiles(const std_fs::path& config_path);

const std::string& GetLanguage();

const char* SetProp(std::vector<std::string>& pvars);
```

weirddan455: Actually, I think those should all be members of the `Config` class.  They're all referencing this `control` global (of type `Config`) when they should probably be member functions with the caller referencing the global.

`ConfigFileIsValid` is just a standalone function though.

weirddan455: I'm gonna do a small bit of refactoring here.

weirddan455: See second commit whenever you get a moment to see if the refactor looks good to you.  I saw several other places we're not following code conventions for stand-alone functions, especially in `cross.cpp` and `support.cpp` but I'm not going to refactor all that in this PR.

johnnovak: That's fine man; thanks for doing some additional cleanup!

johnnovak:
Ship it @weirddan455 whenever you're ready 😆

Issue #1913, opened 2022-09-09
CPU max 100% being CPU thread load 90% is outdated


Grounded0:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Mac OS

### Relevant hardware info

2017 15" MacBook Pro with Radeon Pro 560

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

DOSBox Staging still uses outdated model of issuing load to thread it runs on. Current model actually issues 90% load on the thread with CPU max 100% while 95% load is being issued with CPU max 105%.

This is based on decision made 20 years ago as single thread computers needed more leeway on top of their single thread for background operations to run smoothly. Way to fix this is by making CPU max 100% issue 95% load on the thread and CPU max 105% issue 98% load on the thread that is more suitable to multicore and multithread CPUs of this era.

You can test how flawlessly thread load 95% runs by setting CPU max at 105%. I flagged this as a bug because its misinforming the user that they're maxed out while they're not.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Burrito78:
Yes, agreed. If 105% has no downsides for the platforms we support then it should be the new 100%.

Should be a 0.80 change though. We should now focus on features freeze and bug hunting/translation in the next few days.

Grounded0:
Yeah we're in a code freeze for 0.79.

johnnovak:
I wonder if we could stop this 105% silliness altogether and just treat the percentage values literally (e.g. 100% would mean 100% thread utilisation, no matter what).

Burrito78:
@kcgen Maybe a candidate to implement for 0.81 RC and see how it goes?

johnnovak:
Related: how about not defaulting to max cycles in protected mode? It's a cause for way too many problems. I'd rather default to a roughly 486 DX2/66 equivalent speed which is probably the best overall fit for 90s DOS games when protected mode usage became prevalent. So say 20-30k cycles would be a good starting point instead of max. Maybe we could use your fine-tuned per game configs for the minimum viable cycles value as a guide @kcgen?

MasterO2:
https://dosbox-x.com/wiki/Guide%3ACPU-settings-in-DOSBox%E2%80%90X#_cycles

Scroll down that page for a chart of CPU cycles.

Grounded0:
[...]

PR #3141, opened 2023-11-20
Restore sane `max N%` cycles behaviour where 100% means 100%

johnnovak:
# Description

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/1913

Previously,  `cycles = max` or `cycles = max 100%` really meant 90% CPU core utilisation, and the rather hacky `cycles = max 105%` corresponded to about 95% actual utilisation. You could not go higher than that! This perhaps made _some_ sense in the early 2000s when single-core CPUs were the norm to protect inexperienced users, but as since about 2010 onwards at least dual-cores became common, this is nothing more than some confusing legacy cruft. Even the quite old and underpowered Raspberry Pi 3B released in 2016 has a 64-bit quad-core ARM Cortex...

Having said that, people are always free to use `max 90%`, `max 50%`, or whatever. I wanted to get this in for the 0.81.0 testing cycle so we can give `cycles = max` a good workout, but I expect no issues.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/1913

# Manual testing

Tried a few extremely demanding late 90s demos that use the VESA 512x384 24-bit true colour mode for CPU-based 3D rendering ([kkowboy](https://www.pouet.net/prod.php?which=87) and [moralhardcandy](https://www.pouet.net/prod.php?which=11)). They worked perfectly fine with `cycles = max`; 100% smooth 70 FPS rendering, zero sound glitches, and I was using Discord and a couple of browsers and tons of other stuff in the background on my 10-core M1 Mac.

# Checklist

I have:

### DIFF
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -2114,7 +2114,7 @@ static void CPU_CycleIncrease(bool pressed) {
 	if (!pressed) return;
 	if (CPU_CycleAutoAdjust) {
 		CPU_CyclePercUsed+=5;
-		if (CPU_CyclePercUsed>105) CPU_CyclePercUsed=105;
+		if (CPU_CyclePercUsed>100) CPU_CyclePercUsed=100;
 		LOG_MSG("CPU speed: max %d percent.",CPU_CyclePercUsed);
 		GFX_SetTitle(CPU_CyclePercUsed);
 	} else {
@@ -2265,7 +2265,7 @@ public:
 		CommandLine cmd("", p->GetSection()->Get_string("parameters"));
 
 		constexpr auto min_percent = 0;
-		constexpr auto max_percent = 105;
+		constexpr auto max_percent = 100;
 
 		if (type == "max") {
 			CPU_CycleMax = 0;
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -226,9 +226,9 @@ void increaseticks() { //Make it return ticksRemain and set it in the function a
 
 	if (ticksScheduled >= 250 || ticksDone >= 250 || (ticksAdded > 15 && ticksScheduled >= 5) ) {
 		if(ticksDone < 1) ticksDone = 1; // Protect against div by zero
-		/* ratio we are aiming for is around 90% usage*/
+		// Ratio we are aiming for is 100% usage
 		int32_t ratio = static_cast<int32_t>(
-		        (ticksScheduled * (CPU_CyclePercUsed * 90 * 1024 / 100 / 100)) /
+		        (ticksScheduled * (CPU_CyclePercUsed * 100 * 1024 / 100 / 100)) /
 		        ticksDone);
 
 		int32_t new_cmax = CPU_CycleMax;


Grounded0:
I have been testing this for years so i give it a pass.

Grounded0:
> // Ratio we are aiming for is 100% usage

We're actually targeting 95% usage if it's functional equivalent of 105%.

johnnovak:
> ```
> // Ratio we are aiming for is 100% usage
> ```
> 
> We're actually targeting 95% thread load if it's functional equivalent of 105%. That gives it a little bit of overhead so it should be safe.

I'm not sure about that. I'm not seeing that from the fixed calculations.

Grounded0:
Yeah i have to test this out myself a bit if i'm right or wrong about that. Here's Duke Nukem 3D on `main` looking like its targeting 95% at 105% setting:

Screenshot 2023-11-20 at 7 36 54

johnnovak:
> Yeah i have to test this out myself a bit first. Here's Duke Nukem 3D on `main` looking like its targeting 95% at 105% setting:
> Screenshot 2023-11-20 at 7 36 54

It's tricky because I think the "100% CPU utilisation" is a _potential_, so it's not guaranteed that you set cycles to `max 80%` or `90%` or `100%` and you see the CPU pinned exactly to 80%, 90% and 100% all the time. There are always fluctuations.

kcgen:
fully agree with knocking out the 105% and just making it 100%.

Regarding the 90% target, here's how it works:

The auto-cycler takes a ratio of completed 1ms ticks versus scheduled (or potential) 1ms ticks across a 250 ms period. It scales up cycles more and more as the system continues to prove that it can keep up.. until it can't, at which point it starts lowering cycles.

Benchmark applications are the easiest for this algorithm to get right, because the load is usually 100% and constant; including not waiting for vblank.

The most challenging scenario is when you have an extended period of light load (maybe not sound effect, simple scenery).. perhaps the the auto-cycler's dialed up to 750,000 cycles/ms.. and the host is actually able to complete them.

But then the scene changes quickly, perhaps you open a door and the scene requires a lot of rendering, and there's a bunch of NPC's all generating sound effects that the game needs to mix and then DMA. 

Right then, those cycles are going to be a lot more work.. and the host might actually need 10 ms or more to grind through 750,000 cycles to complete just one 1ms worth of DOS time.

The auto-cycler sees the ratio tank, and needs to brings cycles back to a point where the host can keep up... but even then, it brings them down in a step wise manner to avoid turning the game into a slide slow.

During this "in debt" phase, we're actually bleeding down our audio buffer and that 10% headroom (from the 90% target) will be used up. So it's there more like an internal buffer or "emergency fund", and less about controlling host CPU usage.

johnnovak:
Super informative explanation on how the auto cycler works, @kcgen 🎖️ Ok, so you're advising to change the percentage target back to 90% then? (I changed it to 100%, being not fully across all this background info.)

Grounded0:
@kcgen 

Are you saying that 105% maps to thread load 95%?

Grounded0:
I'm consistently seeing this 5% gap between `main` and `jn/max-cycles-fix` with Jane's Advanced Tactical Fighters in 800x600 with everything maxed out:

Main:

main



Test:

test

Grounded0:
If we're going for actual 100% is 100% then we should default to 95%. I have little problems running at full 100% so it would be a 10% speed boost against `main` 100% 😁

Grounded0:
Waiting until kcgen wakes up since he knows how this works. Thankfully he seems to agree its time to retire this 2004 era assumption.

Grounded0:
My understanding always was that it targeted thread load 90% at 100% and 105% would bump to thread load 95% but i'll wait until kcgen wakes up.

johnnovak:
> Waiting until kcgen wakes up since he knows how this works. Thankfully he seems to agree its time to retire this 2004 era assumption.

Well, now you're getting 100% utilisation with `cycles = max 100%`. That's what you wanted, no? And I think that's what we want, unless @kcgen disagrees.

Grounded0:
Pinging our Quality Assurance Team:

@Burrito78

Grounded0:
Interested to hear how this works for macOS/ARM64.

Grounded0:
My impressions of "100% is 100%" is that the emulation isn't crapping out even if the main thread is pushed at times to 102-103% so it seems to be working much better than i expected. Me advocating for 100% being thread load 95% was just being careful but in fact i start to agree that "100% is 100%".

Grounded0:
With this new model:

Old 105% is now 95%.

Old 100% is now 90%.

johnnovak:
> With this new model:
> 
> Old 105% is now 95%.
> 
> Old 100% is now 90%.

I think that's the good way: you get what you ask for.

Burrito78:
In PCPBench in Mode 122 (1600x1200x16bpp) i'm getting 90% in main with default settings and 99.x% using this PR.

Grounded0 (approved):

Grounded0:
Should be pretty safe merge since we're just working the main thread to full capacity which forces kernel to schedule tasks to other threads leaving main thread exclusively for DOSBox Staging if necessary. 👍

johnnovak:
> Should be pretty safe merge since we're just working the main thread up to full capacity which forces kernel to schedule tasks to other threads leaving the main thread exclusively for DOSBox Staging if necessary. 👍

Waiting for @kcgen to approve too, then yeah, I'll press the Big Green Button 😎 🤘🏻

Grounded0:
Let's merge as of now and fix CPU cycles MAX on another PR.

Grounded0:
Someone should check cycles AUTO and cycles MAX too:

https://www.dosbox.com/wiki/Performance

I'm super sleep deprived and need to sleep now.

kcgen:
Tested with Screamer 2 on the Pi 400, and the audio gaps occur slightly later than versus `main`. So cycles are being kept a kept higher for longer (before "going into debt"), however I'm not counting any additional gaps across the entire demo period - so the Pi seems fast enough to recover that it can fill in the audio buffer on the next ticks.

So I think this is  good :-)

Merging this on the heels of @kklobe's `Adjust ticksDone with render time` PR so the two are back-to-back given the common theme.

Grounded0:
We were chatting on other benefits of this bit of code that may be directly invisible. Mainly the CPU scheduler seeing we're "ringing the bell" by running sometimes at over 100% thread load and giving us those sweet boost clocks faster and more aggressively.

johnnovak:
> We were chatting on IRC on other benefits of this bit of code that may be directly invisible. Mainly the CPU scheduler seeing we're "ringing the bell" by running sometimes at over 100% thread load and giving us those sweet boost clocks faster and more aggressively.

That's a very interesting point; wasn't aware of that but makes sense!

Grounded0:
Tested all the code paths that i know of:

`cycles=auto` "ringing the bell" at 100%.

`cycles=max` "ringing the bell" at 100%.

PR #3148, opened 2023-11-23
Fix state handling of the BIOS's Tandy DAC callbacks

kcgen:
# Description

Since making the BIOS's Tandy DAC callback state configurable at runtime, it had a bug where it could only be shutdown once after which it stayed off until the Tandy card was re-initialized.

The state of the BIOS's Tandy DAC callbacks is now logged, so it's clearer what it's using or if the callbacks have been shutdown:

When the BIOS interrupt calls the actual Tandy DAC:
```
TANDYDAC: Operating at 48000 Hz without resampling
TANDYDAC: Highpass filter enabled (18 dB/oct at 120 Hz)
TANDYDAC: Lowpass filter enabled (12 dB/oct at 4800 Hz)
BIOS: Tandy DAC interrupt linked to Tandy Sound on IRQ 7
```

When the BIOS interrupt calls the Sound Blaster:
```
SBPRO2: Sound Blaster Pro 2 OPL output filter enabled
OPL: Lowpass filter enabled (6 dB/oct at 8000 Hz)
BIOS: Tandy DAC interrupt linked to Sound Blaster on IRQ 7
```

Also makes a couple small self-contained tweaks:
- Eliminates the logical Tandy DAC low sample rate (now that the resampler is in place, this isn't critical)
- Eliminates the SB hot warm up timeframe when the SB is being used as the Tandy DAC because the BIOS toggles the speaker on and off for every DAC sequence.

# Manual testing

- `tandy on` (BIOS DAC linked to Tandy) -> `sbtype sbpro2` (Tandy DAC shutdown, BIOS DAC linked to SB) -> `tandy on` (SB shutdown, BIOS DAC linked back to Tandy)

- `tandy off` (BIOS DAC doesn't exist) -> `sbtype sbpro2` (BIOS DAC doesn't exist)  -> `tandy psg` (BIOS DAC linked to SB)

- `tandy off` & `sbtype off` (BIOS DAC doesn't exist) -> `tandy psg` (BIOS Tandy requested, but neither DAC exists) -> `sbtype sbpro2` (BIOS DAC linked to SB) -> `sbtype off` (BIOS DAC shutdown because neither DAC exists, but BIOS Tandy requested) -> `tandy on` (BIOS DAC linked to Tandy)

Reproduce and tested with: 
[OutRun.zip](https://github.com/dosbox-staging/dosbox-staging/files/13447527/OutRun.zip)

DAC is used for intro voice, engine rumble, and crash effects:

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/bc912d15-a4b0-46e4-abfd-7b35149ed181

Also tested sanitizer builds.

# Checklist

I have:

### DIFF
--- a/src/hardware/tandy_sound.cpp
+++ b/src/hardware/tandy_sound.cpp
@@ -519,9 +519,8 @@ TandyPSG::TandyPSG(const ConfigProfile config_profile, const bool is_dac_enabled
 	base_device->device_start();
 	device->convert_samplerate(render_rate_hz);
 
-	LOG_MSG("TANDY: Initialised audio card with a TI %s PSG %s",
-	        base_device->shortName,
-	        is_dac_enabled ? "and 8-bit DAC" : "but without DAC");
+	LOG_MSG("TANDY: Initialised audio card with a TI %s PSG",
+	        base_device->shortName);
 }
 
 TandyPSG::~TandyPSG()
@@ -632,12 +631,6 @@ bool TS_Get_Address(Bitu &tsaddr, Bitu &tsirq, Bitu &tsdma)
 	return true;
 }
 
-static void set_tandy_sound_flag_in_bios(const bool is_enabled)
-{
-	real_writeb(0x40, 0xd4, is_enabled ? 0xff : 0x00);
-	BIOS_SetupTandySbDacCallbacks();
-}
-
 static void shutdown_dac(Section*)
 {
 	if (tandy_dac) {
@@ -649,8 +642,8 @@ static void shutdown_dac(Section*)
 void TANDYSOUND_ShutDown(Section*)
 {
 	if (tandy_psg || tandy_dac) {
+		BIOS_ConfigureTandyDacCallbacks(false);
 		LOG_MSG("TANDY: Shutting down");
-		set_tandy_sound_flag_in_bios(false);
 		tandy_dac.reset();
 		tandy_psg.reset();
 	}
@@ -662,7 +655,7 @@ void TANDYSOUND_Init(Section *section)
 	const auto prop = static_cast<Section_prop*>(section);
 	const auto pref = prop->Get_string("tandy");
 	if (has_false(pref) || (!IS_TANDY_ARCH && pref == "auto")) {
-		set_tandy_sound_flag_in_bios(false);
+		BIOS_ConfigureTandyDacCallbacks(false);
 		return;
 	}
 
@@ -684,12 +677,15 @@ void TANDYSOUND_Init(Section *section)
 		tandy_dac = std::make_unique<TandyDAC>(
 		        cfg, prop->Get_string("tandy_dac_filter"));
 	}
+
+	// Always request the DAC even if the card doesn't have one because the
+	// BIOS can be routed to the Sound Blaster's DAC if one exists.
+	BIOS_ConfigureTandyDacCallbacks(true);
+
 	tandy_psg = std::make_unique<TandyPSG>(cfg,
 	                                       wants_dac,
 	                                       prop->Get_string("tandy_fadeout"),
 	                                       prop->Get_string("tandy_filter"));
-
-	set_tandy_sound_flag_in_bios(true);
 
 	constexpr auto changeable_at_runtime = true;
 	section->AddDestroyFunction(&TANDYSOUND_ShutDown, changeable_at_runtime);
--- a/src/hardware/tandy_sound.cpp
+++ b/src/hardware/tandy_sound.cpp
@@ -126,7 +126,7 @@ public:
 		bool is_done = false;
 	};
 	struct Registers {
-		uint16_t frequency = 0;
+		uint16_t clock_divider = 0;
 		uint8_t mode = 0;
 		uint8_t control = 0;
 		uint8_t amplitude = 0;
@@ -315,10 +315,14 @@ void TandyDAC::DmaCallback([[maybe_unused]] const DmaChannel*, DMAEvent event)
 
 void TandyDAC::ChangeMode()
 {
-	// Avoid under or overruning the mixer with invalid frequencies
-	// Typically frequencies are in the 8 to 22 Khz range
-	constexpr auto dac_min_freq_hz = 4900;
-	constexpr auto dac_max_freq_hz = 49000;
+	// Typical sample rates are 1.7. 5.5, 11, and rarely 22 KHz. Although
+	// several games (one being OutRun) set instantaneous rates above
+	// 100,000 Hz, we throw these out as they can cause garbage high
+	// frequency harmonics and also cause problems for the Speex Resampler.
+	// For example, a clock divider value of 8 (which is valid) produces a
+	// 450 KHz sampling rate, which is way beyond what Speex can handle.
+	//
+	constexpr auto dac_max_sample_rate_hz = 49000;
 
 	// LOG_MSG("TANDYDAC: Mode changed to %d", regs.mode);
 	switch (regs.mode & 3) {
@@ -327,13 +331,15 @@ void TandyDAC::ChangeMode()
 	case 2: // recording
 		break;
 	case 3: // playback
-		if (!regs.frequency)
+		// Prevent divide-by-zero
+		if (regs.clock_divider == 0) {
 			return;
-		if (const auto freq = tandy_psg_clock_hz / regs.frequency;
-		    freq > dac_min_freq_hz && freq < dac_max_freq_hz) {
+		}
+		if (const auto sample_rate = tandy_psg_clock_hz / regs.clock_divider;
+		    sample_rate < dac_max_sample_rate_hz) {
 			assert(channel);
-			channel->FillUp(); // using the prior frequency
-			channel->SetSampleRate(freq);
+			channel->FillUp(); // using the prior sample rate
+			channel->SetSampleRate(check_cast<uint16_t>(sample_rate));
 			const auto vol = static_cast<float>(regs.amplitude) / 7.0f;
 			channel->SetAppVolume({vol, vol});
 			if ((regs.mode & 0x0c) == 0x0c) {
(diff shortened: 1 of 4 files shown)


src/hardware/tandy_sound.cpp
```diff
@@ -315,9 +315,7 @@ void TandyDAC::DmaCallback([[maybe_unused]] const DmaChannel*, DMAEvent event)
 
 void TandyDAC::ChangeMode()
 {
-	// Avoid under or overruning the mixer with invalid frequencies
-	// Typically frequencies are in the 8 to 22 Khz range
-	constexpr auto dac_min_freq_hz = 4900;
+	// Typical frequencies run up to 22 Khz range
 	constexpr auto dac_max_freq_hz = 49000;
```
johnnovak, line 319: Bit confused where 49 Khz is derived from based on the comment. Actually, the 4.9k min freq is weird too.

src/ints/bios.cpp
```diff
@@ -1069,14 +1070,31 @@ static void shutdown_tandy_sb_dac_callbacks()
...
+// BIOS-based interrupt callbacks using either a Sound Blaster or the 'actual'
+// Tandy DAC module, respectively. If neither are present then the BIOS
+// callbacks aren't setup.
+//
+// The BIOS callbacks are shutdown when the backing device is shutdown to avoid
+// advertizing the DAC's presence when none exists.
+//
+// Returns true if a DAC was actually setup.
+//
+bool BIOS_ConfigureTandyDacCallbacks(const std::optional<bool> maybe_request_dac)
```
johnnovak, line 1083: So just to be clear, this is some DOSBox-special code right? A real Tandy with an SB card in it wouldn't do such trickery, right?

src/hardware/tandy_sound.cpp
kcgen: Yup, these were based on logging all the 'in the wild' Tandy DAC games that I knew of at the time, plus modern-ish '90s mod players that could play through the DAC.

These bounds sat just outside the actual ranges I saw. The Tandy DAC rates are set by dividing the PSG's NTSC-derived clock rate (~3.58 GHz) by a 12-bit divider.

Turns out _OutRun (1989)_ was a DAC game I didn't know about, and it uses a very large divider to stretch (and lower) the engine rumble at lower car speeds! Pretty neat.

See the updated second commit - added more comments, renamed 'freq' to 'sample_rate', and renamed the 'frequency' register to 'clock_divider'.

src/ints/bios.cpp
kcgen: Yup, the Tandy predated the Sound Blaster, so this deconflicting approach woulnd't have been in the original Tandy BIOS:

- Circa 2020 discussion:

    https://forum.vcfed.org/index.php?threads/is-soundblaster-compatible-with-tandy-1000-tl-2-would-plug-in-but-wont-fit-in-case.77144/

    > [mystman12](https://forum.vcfed.org/index.php?members/mystman12.40047/):  _well not everything is working perfectly. For some reason DeskMate programs that use sampled audio, like the music player and the audio recorder, crash the moment they try playing sound. I tried Life & Death and the sampled sounds still play just fine in that. I guess it's an issue specific to DeskMate?_ 

    > [Trixter](https://forum.vcfed.org/index.php?members/trixter.2251/): _The Tandy TL/2 has a built-in DAC and Deskmate is likely using that. It shouldn't crash, not sure what's going on there, but that explains why it's not trying to use the Sound Blaster._

- Circa 1996 newsgroup chat in `comp.sys.tandy` with the DeskMate author about the Tandy DAC:

    https://groups.google.com/g/comp.sys.tandy/c/7VxlbVGughE/m/wekj31wJfZYJ

- Circa 2005 document about Tandy:

    http://www.oldskool.org/guides/tvdog/1kfaq.html

    > _you can install a SoundBlaster or SoundBlaster Pro in an expansion slot. If you do so, you will have to stop using the Tandy DAC, since the SoundBlaster drivers will be confused by the DAC BIOS routines._

But given the BIOS interrupt callback is a software construct, maybe some tools or drivers were able to re-program the BIOS as a fix-up.

Also, eventually newer Sound Blaster models got better at handling this conflict using IRQ sharing at the hardware level:

    https://www.youtube.com/watch?v=jfkLDZEb8YA

    > _I finally fixed the Sound Blaster and Tandy DAC Conflict in my Tandy 1000 TL/2 Computer, and I'm beyond thrilled! the Sound Blaster Card that I used that worked is the CT1350B because it has a DMA Sharing Jumper on it, and that made all the difference!_

So given all of the above, there would have been ~1990's users able to have these hybrid Tandy Sound + Sound Blaster machines, and I suspect the DOSBox devs wanted to provide that same level of maximum compatibility.

src/ints/bios.cpp
```diff
@@ -1146,10 +1148,12 @@ void BIOS_ConfigureTandyDacCallbacks(const std::optional<bool> maybe_request_dac
 			for (auto i = 0; i < 0x10; i++) {
 				phys_writeb(PhysicalMake(0xf000, 0xa084 + i), 0x80);
 			}
-		} else {
-			real_writeb(0x40, 0xd4, 0x00);
+			return true;
 		}
 	}
+	// Inidicate that the Tandy DAC callbacks are unavailable
```
johnnovak, line 1154: Indicate

src/hardware/tandy_sound.cpp
```diff
@@ -315,10 +315,13 @@ void TandyDAC::DmaCallback([[maybe_unused]] const DmaChannel*, DMAEvent event)
...
-	// Typically frequencies are in the 8 to 22 Khz range
-	constexpr auto dac_min_freq_hz = 4900;
-	constexpr auto dac_max_freq_hz = 49000;
+	// Typical sample rates are 1.7. 5.5, 11, and rarely 22 KHz. Although
+	// several games (one being OutRun) set instantaneous rates above
+	// 100,000 Hz, we throw these out as invalid by using the following
+	// reasonable and sane maximum. There are no known instances where we
+	// should instead try to play such audio.
+	//
+	constexpr auto dac_max_sample_rate_hz = 49000;
```
johnnovak, line 324: Nice explanation and I guess that max rate wouldn't really cause harm in a practical sense, but is it really necessary? What would happen if we just ditched it?

src/ints/bios.cpp
johnnovak: Clearly you've done a thorough research on this @kcgen 🎖️ Your extra comments in this PR describe the "how" well (the technicalities of this SB/Tandy DAC dance), but not the "why". I think a short top-level comment in `tandy.cpp` with an overview of this whole SB vs Tandy DAC handling business and the motivation for it would help future maintainers. Like in my case, I was just scratching my head reading your PR and the number of people aware of the background context is only going to decrease going forward 😅

src/hardware/tandy_sound.cpp
kcgen: I remember putting it in to eliminate to high frequency artifacts and harmonics; but forgot the exact cases.

There's another problem in that the Speex Resampler fails at really high rates. For example, a divisor of 1 means we'll request a sample rate of 3+ GHz; even a divisor of 8 is still in 450 KHz, which is still way beyond what Speex can handle.

I'm not willing to do this level of regression testing to make this change - but if anyone else wants to in a seperate PR; them the issue can bisect to them :-)

johnnovak: Thanks for the explanation, I was just curious 😄 It would be useful to mention we need to cap the max rate to avoid such riduculously high sample rates that will upset Speex.

src/ints/bios.cpp
kcgen: Will do :-) yes; without the logging, it was something that either worked or didn't behind the scenes!

johnnovak (approved):
Code looks good @kcgen, my comments are mainly about adding explanation to make life easier for the next guy maintaining this 😄

kcgen:
> Code looks good @kcgen, my comments are mainly about adding extra explanations to make life easier for the next guy maintaining this 😄

Absolutely! :grin: comment added in the last comment; see if you think it covers it.

src/hardware/tandy_sound.cpp
kcgen: Added - and much clearer (snippet):

```c++
// Although
// several games (one being OutRun) set instantaneous rates above
// 100,000 Hz, we throw these out as they can cause garbage high
// frequency harmonics and also cause problems for the Speex Resampler.
// For example, a clock divider value of 8 (which is valid) produces a
// 450 KHz sampling rate, which is way beyond what Speex can handle.
```

johnnovak: Yep, that covers it and now it makes perfect sense why we need that upper limit 🚀

kcgen: Thanks for bashing through this with me :-)

src/hardware/tandy_sound.cpp
```diff
@@ -23,57 +23,23 @@
...
-MCE2VGA to capture the output of a real Tandy 1000, and then tried to adjust the
-Tandy DAC code to match the artifact after each step."
-
-The implementation below prioritizes the game author's intended audio score
-ahead of deleterious artifacts caused by hardware defects or limitations.
-Because the sound caused by the Tandy's DAC is not part of the game's audio
-score, we deliberately omit this behavior and terminate sequences at the
-centerline.
-*/
+// Interaction between the Tandy DAC and the Sound Blaster:
```
johnnovak, line 26: Good riddance 🎉 I don't think including this super-detailed info about one very specific issue was warranted, so I'm glad it's gone.

src/hardware/tandy_sound.cpp
```diff
@@ -23,57 +23,23 @@
...
+// How does this work in DOSBox? DOSBox Staging always shuts down conflicting
+// DMA devices (and the Tandy DAC vs. SB is no exception), however the Tandy DAC
+// is unique in that the machine's BIOS (yes, on real hardware, too) is
+// programmed with a callback that points to the DAC device.  In the case of
+// DOSBox, that BIOS routine either points to the Sound Blaster's DAC or Tandy
+// DAC, whichever is running.
+//
+// So using this BIOS callback, DOSBox (and Staging) is able to support a
+// Tandy+SB combo configuration as well. Note that the Tandy DAC BIOS routine
+// only exists if the Tandy Card is enabled (either 'tandy=on' or 'tandy=psg').
```
johnnovak, line 42: Very nice explanation @kcgen 🎖️

Issue #2404, opened 2023-04-13
Can't change Quake viewsize parameter above "99" through in-game menu on ARM machines


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.1 and latest DEV

### What Operating System are you using?

Mac OS

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

In Quake SW I can't change the screen size above the default value.
The slider in the menu just won't move beyond that point.

Tested in vanilla, 0.80.1 and latest DEV on macOS Ventura.

On Windows using 0.80.1 it works as expected!

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Download latest DEV on macOS
2. Run Quake SW 1.06
3. Go into menu and try to increase screen size, doesn't work

See the screenshots to see what it should look like:
![IMG_2056](https://user-images.githubusercontent.com/61995128/231757327-8be7f4e5-54de-4c43-b69a-ac3126a32c5c.jpeg)
![IMG_2057](https://user-images.githubusercontent.com/61995128/231757278-f314923f-b622-4dbd-b732-5e5f0d7dc962.jpeg)
![IMG_2058](https://user-images.githubusercontent.com/61995128/231757313-2e348727-52e5-4a26-9830-649fccbc68bf.jpeg)

### Your configuration

```ini
default
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Burrito78:
OK, tried to reproduce on a second Mac (Macbook Pro mid 2014 running Big Sur latest update), there latest DEV release is working fine using Quake SW 1.06!

So for me it only happens on macMini M2 using macOS Ventura.

GranMinigun:
This is not related to window sizing at all.

@Burrito78, can you please back up your current Quake installation and test with a clean one, preferably patched to 1.08?

Burrito78:
Yes, of course. I've used Redumps "Quake (USA) (Rerelease) (19980921)", it is v1.08. Same issue. A very weird bug and no regression from our POV as it happens in vanilla too.

Burrito78:
Its a Apple silicon or Ventura bug affecting all DOSBox versions.

Burrito78:
This is it what happens when I delete config.cfg (Quakes internal config file) and try to move the slider to the right for the first time:

[...]

PR #3149, opened 2023-11-24
Fix rounding when very close to an integer

kklobe:
# Description

This is an experimental hack for #2404. When in `ROUND_Chop` mode on non-x64 platforms, if the input number is very close to, but less than, a whole integer, it gets chopped to the value below, e.g. 7.9999999994 will chop to 7 when it should be 8.

This is only done when the FPU is set to extended precision mode, the assumption being that the number to chop, if that close to an integer, is inaccurate and would likely be the integer under full 80-bit precision.

# Manual testing

Tested the options sliders with Quake SW 1.06, and also tested QTD for performance regressions (none noted).

# Checklist

I have:

### DIFF
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -20,6 +20,7 @@
 #include "fpu.h"
 #endif
 
+static constexpr uint16_t ExtendedPrecisionModeMask = 0x0300;
 
 static void FPU_FINIT(void) {
 	FPU_SetCW(0x37F);
@@ -105,17 +106,24 @@ static double FROUND(double in){
 	case ROUND_Down: return std::floor(in);
 	case ROUND_Up: return std::ceil(in);
 	case ROUND_Chop: 
-	{	
-		constexpr double tolerance = 1e-14;
-		const auto lower = std::floor(in);
-		const auto upper = std::ceil(in);
+	{
+		// This is a fix for rounding to a close integer in
+		// extended precision mode, e.g. 7.999999999999994; an example
+		// can be seen in the Quake options screen size slider.
+		// In this case, what is almost certainly wanted is 8.0,
+		// so return the closer integer within the tolerance instead
+		// of chopping to the lower value.
+		if (fpu.cw & ExtendedPrecisionModeMask) {
+			constexpr double tolerance = 1e-14;
 
-		// Check if the value is within tolerance of the lower or upper integer
-		if ((in - lower) < tolerance) {
-			return lower;
-		} else if ((upper - in) < tolerance) {
-			return upper;
-		} 
+			if (const auto lower = std::floor(in);
+			    (in - lower) < tolerance) {
+				return lower;
+			} else if (const auto upper = std::ceil(in);
+			           (upper - in) < tolerance) {
+				return upper;
+			}
+		}
 		return std::trunc(in);
 	}
 	default: return in;
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -104,7 +104,20 @@ static double FROUND(double in){
 	case ROUND_Nearest: return std::nearbyint(in);
 	case ROUND_Down: return (floor(in));
 	case ROUND_Up: return (ceil(in));
-	case ROUND_Chop: [[fallthrough]]; // cast by the caller chops
+	case ROUND_Chop: 
+	{	
+		constexpr double tolerance = 1e-15;
+		const auto lower = std::floor(in);
+		const auto upper = std::ceil(in);
+
+		// Check if the value is within tolerance of the lower or upper integer
+		if ((in - lower) < tolerance) {
+			return lower;
+		} else if ((upper - in) < tolerance) {
+			return upper;
+		} 
+		return in;
+	}
 	default: return in;
 	}
 }
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -107,6 +107,9 @@ static double FROUND(double in){
 	case ROUND_Up: return std::ceil(in);
 	case ROUND_Chop: 
 	{
+		// Most x32/x64 builds use fpu_instructions_x86.h.
+		// If we are here on x64, we don't need this fix.
+#if !defined(__x86_64__) && !defined(_M_X64)
 		// This is a fix for rounding to a close integer in
 		// extended precision mode, e.g. 7.999999999999994; an example
 		// can be seen in the Quake options screen size slider.
@@ -124,6 +127,7 @@ static double FROUND(double in){
 				return upper;
 			}
 		}
+#endif
 		return std::trunc(in);
 	}
 	default: return in;
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -102,8 +102,8 @@ static void FPU_FPOP(void){
 static double FROUND(double in){
 	switch(fpu.round){
 	case ROUND_Nearest: return std::nearbyint(in);
-	case ROUND_Down: return (floor(in));
-	case ROUND_Up: return (ceil(in));
+	case ROUND_Down: return std::floor(in);
+	case ROUND_Up: return std::ceil(in);
 	case ROUND_Chop: 
 	{	
 		constexpr double tolerance = 1e-15;
@@ -116,7 +116,7 @@ static double FROUND(double in){
 		} else if ((upper - in) < tolerance) {
 			return upper;
 		} 
-		return in;
+		return std::trunc(in);
 	}
 	default: return in;
 	}
@@ -432,13 +432,12 @@ static void FPU_FUCOM(Bitu st, Bitu other){
 }
 
 static void FPU_FRNDINT(void){
-	int64_t temp  = static_cast<int64_t>(FROUND(fpu.regs[TOP].d));
-	double tempd = static_cast<double>(temp);
+	const auto rounded  = FROUND(fpu.regs[TOP].d);
 	if (fpu.cw&0x20) { //As we don't generate exceptions; only do it when masked
-		if (tempd != fpu.regs[TOP].d)
+		if (rounded != fpu.regs[TOP].d)
 			fpu.sw |= 0x20; //Set Precision Exception
 	}
-	fpu.regs[TOP].d = tempd;
+	fpu.regs[TOP].d = rounded;
 }
 
 static void FPU_FPREM(void){
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -106,7 +106,7 @@ static double FROUND(double in){
 	case ROUND_Up: return std::ceil(in);
 	case ROUND_Chop: 
 	{	
-		constexpr double tolerance = 1e-15;
+		constexpr double tolerance = 1e-14;
 		const auto lower = std::floor(in);
 		const auto upper = std::ceil(in);
 
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -116,7 +116,7 @@ static double FROUND(double in){
 		// In this case, what is almost certainly wanted is 8.0,
 		// so return the closer integer within the tolerance instead
 		// of chopping to the lower value.
-		if (fpu.cw & ExtendedPrecisionModeMask) {
+		if ((fpu.cw & ExtendedPrecisionModeMask) == ExtendedPrecisionModeMask) {
 			constexpr double tolerance = 1e-14;
 
 			if (const auto lower = std::floor(in);


kcgen:
(Have my system switched over in FreeBSD / PPC land; will help test when I get through it :)

Here's Intel's FPU tester [mcpdiag.zip](https://github.com/dosbox-staging/dosbox-staging/files/13461505/mcpdiag.zip); repro-pack will autotype to launch the test.

kklobe:
> (Have my system switched over in FreeBSD+PPC land; will help test when I get through it :)
> 
> Here's Intel's FPU tester [mcpdiag.zip](https://github.com/dosbox-staging/dosbox-staging/files/13461505/mcpdiag.zip); repro-pack will autotype to launch the test.

Yep - 8 more "Round to Integer" failures on this branch, which would be expected in a strict compliance test.

kcgen:
> Yep - 8 more "Round to Integer" failures on this branch, which would be expected in a strict compliance test.

:+1: ; makes sense - but maybe we're still at-par (or better) on ARM64?

kklobe:
> > Yep - 8 more "Round to Integer" failures on this branch, which would be expected in a strict compliance test.
> 
> 👍 ; makes sense - but maybe we're still at-par (or better) on ARM64?

Yeah - main vs this branch on the M1 has the 8 add'l failures, but I think this hack provides the true intent of code hitting it, like the Quake scenario: if someone is calling FRNDINT with 7.9999999999999998 or 8.0000000000000003, in `ROUND_Chop` mode, they almost certainly were passing in 8.

Burrito78:
M2 MacBook Air

MCPDIAG: Only difference of this PR is that it passes 8 tests less than main in the category "Round to Integer" (as expected).

There is still an issue in Quake though. The slider still doesn't go to 11 err completely to the right. But it's getting better!

main
main

PR
PR

kklobe:
Thanks @Burrito78! Just a slight loosening of the tolerance took care of it, I think:

CleanShot 2023-11-25 at 09 52 16@2x
CleanShot 2023-11-25 at 09 51 31@2x

rderooy:
should this not be limited to affected architectures, or to non-x86_64?

src/fpu/fpu_instructions.h
```diff
@@ -102,9 +103,29 @@ static void FPU_FPOP(void){
...
+		if (fpu.cw & ExtendedPrecisionModeMask) {
+			constexpr double tolerance = 1e-14;
+			const auto lower           = std::floor(in);
+			const auto upper           = std::ceil(in);
+
+			if ((in - lower) < tolerance) {
+				return lower;
+			} else if ((upper - in) < tolerance) {
+				return upper;
+			}
```
kcgen, line 125: (just a minor micro-logic comment):
- `upper` isn't needed until we fail the first `lower` check,
- and both `upper` and `lower` aren't needed once the if/else fails

So could be a nice use of C++17's if/else initializer (moving `lower` and `upper` in as initializers).

src/fpu/fpu_instructions.h
```diff
@@ -419,13 +440,12 @@ static void FPU_FUCOM(Bitu st, Bitu other){
 }
 
 static void FPU_FRNDINT(void){
-	int64_t temp  = static_cast<int64_t>(FROUND(fpu.regs[TOP].d));
-	double tempd = static_cast<double>(temp);
+	const auto rounded  = FROUND(fpu.regs[TOP].d);
```
kcgen, line 443: Very nice discombobulating :-)

Burrito78:
Confirmed to work now as expected! Man, this is really great! Thank you so much for doing this! 🥳

Bildschirmfoto 2023-11-25 um 20 43 07

kklobe:
> should this not be limited to affected architectures, or to non-x86_64?

I think on x86_64 `fpu_instructions_x86.h` gets pulled in instead of this file, but I haven't done the full matrix testing on the different platforms yet - good reminder, thanks.

kcgen (approved):
Extremely nice work putting this together, @kklobe.
Looks good from my side!
Sooner we get this in, the sooner we can get wider test coverage.

kklobe:
> Extremely nice work putting this together, @kklobe. Looks good from my side! Sooner we get this in, the sooner we can get wider test coverage.

Thanks! Just want to get in a skip for x64 platforms, as mentioned by @rderooy, and test the artifacts on Rosetta and Windows 11, shouldn't take long.

src/fpu/fpu_instructions.h
```diff
@@ -102,9 +103,33 @@ static void FPU_FPOP(void){
...
-	case ROUND_Down: return (floor(in));
-	case ROUND_Up: return (ceil(in));
-	case ROUND_Chop: [[fallthrough]]; // cast by the caller chops
+	case ROUND_Down: return std::floor(in);
+	case ROUND_Up: return std::ceil(in);
+	case ROUND_Chop: 
+	{
+		// Most x32/x64 builds use fpu_instructions_x86.h.
+		// If we are here on x64, we don't need this fix.
+#if !defined(__x86_64__) && !defined(_M_X64)
```
kklobe, line 112: @kcgen let me know if you are ok with this here, or if you would prefer a new build system define get created and used instead.

johnnovak:
> > Extremely nice work putting this together, @kklobe. Looks good from my side! Sooner we get this in, the sooner we can get wider test coverage.
> 
> Thanks! Just want to get in a skip for x64 platforms, as mentioned by @rderooy, and test the artifacts on Rosetta and Windows 11, shouldn't take long.

Hang on a minute, so as I said yesterday on the chat @kklobe, our 32-bit build works fine (it doesn't manifest the vertical stripes bug) probably because somehow the 80-bit FPU regs get used by the C libraries. So x86_64 is fine, but I think it's still brittle... (what it the compiler stops using the 80-bit FPU regs one day).

But the 64-bit builds, so x86_64, however are broken and manifest the bug 100% as 80-bit FPU regs are not used for 64-bit x86 builds. We absolutely want this fir for the x86_64 builds!

The only thing you could exclude theoretically is the 32-bit x86 builds... but I wouldn't bother.

kklobe:
> > > Extremely nice work putting this together, @kklobe. Looks good from my side! Sooner we get this in, the sooner we can get wider test coverage.
> > 
> > 
> > Thanks! Just want to get in a skip for x64 platforms, as mentioned by @rderooy, and test the artifacts on Rosetta and Windows 11, shouldn't take long.
> 
> Hang on a minute, so as I said yesterday on the chat @kklobe, our 32-bit build works fine (it doesn't manifest the vertical stripes bug) probably because somehow the 80-bit FPU regs get used by the C libraries. So x86_64 is fine, but I think it's still brittle... (what it the compiler stops using the 80-bit FPU regs one day).
> 
> But the 64-bit builds, so x86_64, however are broken and manifest the bug 100% as 80-bit FPU regs are not used for 64-bit x86 builds. We absolutely want this fir for the x86_64 builds!
> 
> The only thing you could exclude theoretically is the 32-bit x86 builds... but I wouldn't bother.

Right, this is just for the rounding fix, which isn't needed on x64 builds, as shown by @Burrito78's screenshots for main on Rosetta above.

src/fpu/fpu_instructions.h
kcgen: Good point - if the CPU type isn't commonly defined, Meson could genericize them for us.  Seeing them used in the DR codecs, XXHash, and STB Vorbis libs.. and these are extremely universal headers compiled by countless compilers; so I think these are defacto enough.

johnnovak: No, this can't be correct. As I explained below, the bug 100% manifest in our 64-bit x86 builds—I tried it yesterday on Win10. I also tried the 0.80.1 32-bit build, and the 32-bit x86 build is "fine" because that uses 80-bit FPU regs internally by fluke.

johnnovak:
> Right, this is just for the rounding fix, which isn't needed on x64 builds, as shown by @Burrito78's screenshots for main on Rosetta above.

Ok... I slep like 5 hours 😬 Carry on, I was referring to the vertical stripes fix, so the other PR 😊

Grounded0:
@johnnovak 

You're not the only one mixing these two up. At least now the PRs are correctly linked to the correct issues.

kcgen:
@kklobe : maybe this rounding fix should be excluded for Intel x86 32bit machines, too? (__i386__, __i686__, ?) Hmm.. I'm not sure of what preprocessor macros are common for these.

src/fpu/fpu_instructions.h
```diff
@@ -102,9 +103,33 @@ static void FPU_FPOP(void){
...
+			constexpr double tolerance = 1e-14;
+
+			if (const auto lower = std::floor(in);
+			    (in - lower) < tolerance) {
+				return lower;
+			} else if (const auto upper = std::ceil(in);
+			           (upper - in) < tolerance) {
+				return upper;
+			}
+		}
```
johnnovak, line 129: *Maybe* this could cause issues with application software that use heavy numerical calculations, but who knows... I guess only testing will tell!

kklobe:
> @kklobe, should it be excluded for Intel x86 32bit machines, too? (`__i386__`, `__i686__`, ?) Hmm.. I'm not sure of what preprocessor macros are common for these.

Those will be skipped entirely, because the 32-bit x86 builds all use `fpu_instructions_x86.h` - MSVC hard-defines it (`C_FPU_X86` in `config.h`), and Meson defines it for both 32 and 64 bit x86 builds.

src/fpu/fpu_instructions.h
kklobe: Hopefully it should be more accurate for most cases! Like in Quake, there's that small difference when rounding using truncate (or "chop") mode, that what is calculated as 8.0 on x64, on ARM is e.g. 7.999999999999996, but the chop mode puts that down to 7.

johnnovak: A disclaimer section in our webpage is in order, btw. Basically we should state our emulator is only for entertainment purposes and for personal use, and we're not responsible for any loss or catastrophes when used for anything else... 😬

MasterO2: > A disclaimer section on our webpage is in order, btw. Basically we should state our emulator is only for entertainment purposes and for personal use, and we're not responsible for any loss or catastrophes when used for anything else... 😬

And a sub-disclaimer for those engineers who, for some strange reason, are using Dosbox Staging to engineer bridges.... [/half joking]

ThomasEricB: > A disclaimer section on our webpage is in order, btw. Basically we should state our emulator is only for entertainment purposes and for personal use, and we're not responsible for any loss or catastrophes when used for anything else... 😬

Yeah we can do something like that.

kcgen: > *Maybe* this could cause issues with application software that use heavy numerical calculations, but who knows... I guess only testing will tell!

With rounding pumped up a notch, we should see improvements in truss thicknesses and I-beam width calculations, so perhaps safer buildings and bridges: 👍 👷

I wish I had some of those old posts.. but yeah: there are some very old civil engineering packages that people are still using and teaching with in schools (and using dosbox!)

Grounded0: I think there should be a setting where we default to this ON and optionally allowing people to run with more precision if they so desire (on ARM platforms).

Grounded0: But that can be implemented as a follow-up PR.

kcgen:
> Those will be skipped entirely, because the 32-bit x86 builds all use fpu_instructions_x86.h

Thanks for clearly this up, @kklobe; perfect!

src/fpu/fpu_instructions.h
kklobe: This is an awfully big hammer for one feature in one game...

Grounded0: Its Quake. 🙂

kklobe:
Another FYI: Quake switches precision mode quite often:

```log
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: extended precision
2023-11-27 14:51:16.937 | FPU rounding mode: extended precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: extended precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: extended precision
2023-11-27 14:51:16.937 | FPU rounding mode: single precision
2023-11-27 14:51:16.937 | FPU rounding mode: extended precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: extended precision
2023-11-27 14:51:16.938 | FPU rounding mode: extended precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: extended precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: extended precision
2023-11-27 14:51:16.938 | FPU rounding mode: single precision
2023-11-27 14:51:16.938 | FPU rounding mode: extended precision
2023-11-27 14:51:16.939 | FPU rounding mode: single precision
2023-11-27 14:51:16.939 | FPU rounding mode: single precision
2023-11-27 14:51:16.939 | FPU rounding mode: single precision
2023-11-27 14:51:16.939 | FPU rounding mode: single precision
```

johnnovak (approved):
Looks good to me, let's merge then @kklobe 🚀

kklobe:
> Looks good to me, let's merge then @kklobe 🚀

Wait! Critical fix incoming.

Issue #3125, opened 2023-11-15
Allow extended precision floating point calculations to be done in software


Grounded0:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Is your feature request related to a problem? Please describe.

Currently any hardware that doesn't support IEEE 754 extended precision format (80-bit) is doing calculations in double precision (64-bit) format when it sees an x87 instruction. There should be a configuration option to enable extended precision formats to be calculated in software.

Possibly connected issues and PRs:

- ~~#610~~
- ~~#2404~~
- ~~#1231~~
- ~~#2900~~

And a lot of stuff we're not even aware of yet.

### Describe the solution you'd like

There should be a configuration option to enable extended precision formats to be calculated in software.

It could have smart analysis of incoming x87 instructions:

x87 extended precision format => calculate in software
x87 double precision and below format => calculate by hardware

Old x86 and 68k processors without FPUs did all floating point calculation in software when program used a floating point instruction. Now we would only need to calculate extended precision formats in software through smart analysis of the instructions.

### Describe alternatives you've considered

I can't think of anything except using architectures with quadruple precision floating point in hardware and doing extended precision in `binary128` or sticking with x86-64 with its register level x87 support.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

weirddan455:
>  Old x86 and 68k processors without FPUs did all floating point calculation in software when program used a floating point instruction.

I don't think this is true regarding old x86 processors (I don't know about 68k but that's not very relevant to Dosbox).  I believe if you don't have an FPU (either a separate x87 co-processor or one that's built-in to the main CPU), you simply cannot execute floating point instructions.  The executable should either crash or throw an error.

Games, on the other hand, could emulate this in software using integer instructions.  Usually this was fixed-point though, not floating-point at all.  There's nothing we as an emulator need to do for this case.  It's all handled in game code.

Also, note that this should only be a problem on non-x86 systems.  On x86, those x87 instructions are still available in hardware on modern CPUs (and I seem to recall someone tested that we are using them to 80-bit precision).

On ARM, we probably would need software emulation if we want 80-bit precision.  Doing so will help accuracy but is likely to cause a significant hit to performance (how much it will hurt is hard to say).

johnnovak:
[...]

Issue #1231, opened 2021-08-29
Games w/ vertical lines


nemo93:
**Describe the bug**
At least 2 games do have vertical lines making them unplayable: [Carmageddon](https://github.com/dosbox-staging/dosbox-staging/wiki/Games-issues#carmageddon) and [Motor Mash](https://github.com/dosbox-staging/dosbox-staging/wiki/Games-issues#motor-mash). Here's a picture which illustrates the problem (taken from that [thread](https://www.vogons.org/viewtopic.php?t=23521)):

![carma](https://user-images.githubusercontent.com/8100500/131256393-151d6015-7494-47fc-b082-3afba019b652.jpg)

This is a rather well-known issue present also on SVN. According to Qbix: "on non-x86 we emulate the fpu using doubles which are 64 bit". [[Reference](https://www.vogons.org/viewtopic.php?p=188733#p188733)]

**To Reproduce**
Launch the games on a non-x86 (eg. Pi) or 64-bit system.

**Expected behaviour**
No lines at all. Clear and crisp display :-)

**Environment:**
- Raspbian OS 32bit (Buster)
- Pi4, no overclock, fKMS (fake mode, default), stock kernel

**Additional context**
Both games do rely on the same game engine: [BRender](https://en.wikipedia.org/wiki/Argonaut_Games#BRender).

kcgen:
I wonder if @cimarronm 's [FPU improvements](https://github.com/joncampbell123/dosbox-x/pulls?q=is%3Apr+is%3Aclosed+author%3Acimarronm+fpu) correctly handle this?

arrowgent:
ive seen this before in carmageddon,  i figured it was the settings i was using or version from GOG (i had some old versions with different DRM "fixes")
doesnt seem to happen in all settings
`voodo2c -hires -novoodoo`
doesnt seem to have this issue

nemo93:
Thanks for this @arrowgent! Indeed the `-hires` parameter does improve the situation a bit as with it, vertical lines are not visible during gameplay. However there are still there at loading screen and on various elements in the menus (like on the "faces" from the drivers selection screen). 

I must add "hires" mode is too much for such a low-spec device like RPi 😢

arrowgent:
https://github.com/dosbox-staging/dosbox-staging/releases/tag/v0.77.1

tested on linux x86_64 again, no issues
cant test on RPI4 right now, will have to wait a day.

other files provided by the CD/ISO/GOG:

[...]

PR #3151, opened 2023-11-25
Fix vertical bars due to FILD/FIST memcpy trick

kklobe:
# Description

This is a fix for the vertical bars video corruption seen on a few games (Carmageddon, Motor Mash) and demos (Sunflower, Multikolor), due to a trick they use with FILD/FIST FPU instructions to do a fast memcpy, specifically with 64-bit integer values.

Obviously this needs a lot of testing - throw everything you can at it.

## Related issues

Several games and demos are mentioned in #1231 and #3125.

Ideas for this fix were inspired by https://github.com/joncampbell123/dosbox-x/issues/119.

# Manual testing

I tested the following:

- Intel Math CoProcessor Advanced Diagnostics to verify no new floating point issues were introduced.
- Quake SW 1.06 to verify non-memcpy-trick usage of FILD/FIST still works
- Carmageddon
- Motor Mash
- Sunflower
- Multikolor

# Checklist

I have:

### DIFF
--- a/include/fpu.h
+++ b/include/fpu.h
@@ -83,14 +83,16 @@ enum FPU_Round : uint8_t {
 };
 
 struct FPU_rec {
-	FPU_Reg regs[9]      = {};
-	FPU_P_Reg p_regs[9]  = {};
-	FPU_Tag tags[9]      = {};
-	uint16_t cw          = 0;
-	uint16_t cw_mask_all = 0;
-	uint16_t sw          = 0;
-	uint32_t top         = 0;
-	FPU_Round round      = {};
+	FPU_Reg regs[9]         = {};
+	int64_t regs_memcpy[9]  = {}; // for FILD/FIST 64-bit memcpy fix
+	bool use_regs_memcpy[9] = {};
+	FPU_P_Reg p_regs[9]     = {};
+	FPU_Tag tags[9]         = {};
+	uint16_t cw             = 0;
+	uint16_t cw_mask_all    = 0;
+	uint16_t sw             = 0;
+	uint32_t top            = 0;
+	FPU_Round round         = {};
 };
 
 extern FPU_rec fpu;
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -35,6 +35,10 @@ static void FPU_FINIT(void) {
 	fpu.tags[6] = TAG_Empty;
 	fpu.tags[7] = TAG_Empty;
 	fpu.tags[8] = TAG_Valid; // is only used by us
+
+	for (auto& use_regs_memcpy : fpu.use_regs_memcpy) {
+		use_regs_memcpy = false;
+	}
 }
 
 static void FPU_FCLEX(void){
@@ -65,6 +69,7 @@ static void FPU_PREP_PUSH(void){
 	}
 #endif
 	fpu.tags[TOP] = TAG_Valid;
+	fpu.use_regs_memcpy[TOP] = false;
 }
 
 static void FPU_PUSH(double in){
@@ -94,6 +99,7 @@ static void FPU_FPOP(void){
 	}
 #endif
 	fpu.tags[TOP]=TAG_Empty;
+	fpu.use_regs_memcpy[TOP] = false;
 	//maybe set zero in it as well
 	TOP = ((TOP+1)&7);
 //	LOG(LOG_FPU,LOG_ERROR)("popped from %d  %g off the stack",top,fpu.regs[top].d);
@@ -220,10 +226,26 @@ static void FPU_FLD_I32(PhysPt addr,Bitu store_to) {
 }
 
 static void FPU_FLD_I64(PhysPt addr,Bitu store_to) {
-	FPU_Reg blah;
-	blah.l.lower = mem_readd(addr);
-	blah.l.upper = mem_readd(addr+4);
-	fpu.regs[store_to].d = static_cast<Real64>(blah.ll);
+	constexpr int64_t int53_min = -(1LL << 53);
+	constexpr int64_t int53_max = (1LL << 53) - 1;
+
+	FPU_Reg temp_reg;
+	temp_reg.l.lower = mem_readd(addr);
+	temp_reg.l.upper = mem_readd(addr + 4);
+
+	fpu.regs[store_to].d = static_cast<Real64>(temp_reg.ll);
+	// This is a fix for vertical bars that appear in some games that
+	// use FILD/FIST as a fast 64-bit integer memcpy, such as Carmageddon,
+	// Motor Mash, and demos like Sunflower and Multikolor.
+
+	// If the value won't fit in the 53-bit mantissa of a double, save the
+	// value into the regs_memcpy register and set the corresponding bool in
+	// use_regs_memcpy to indicate that the integer value should be read out
+	// by the FPU_FST_I64 function instead.
+	if (temp_reg.ll > int53_max || temp_reg.ll < int53_min) {
+		fpu.regs_memcpy[store_to]     = temp_reg.ll;
+		fpu.use_regs_memcpy[store_to] = true;
+	}
 }
 
 static void FPU_FBLD(PhysPt addr,Bitu store_to) {
@@ -291,13 +313,23 @@ static void FPU_FST_I32(PhysPt addr) {
 	mem_writed(addr,(val < 2147483648.0 && val >= -2147483648.0)?static_cast<int32_t>(val):0x80000000);
 }
 
-static void FPU_FST_I64(PhysPt addr) {
-	double val = FROUND(fpu.regs[TOP].d);
-	FPU_Reg blah;
-	blah.ll = (val < 9223372036854775808.0 && val >= -9223372036854775808.0)?static_cast<int64_t>(val):LONGTYPE(0x8000000000000000);
+static void FPU_FST_I64(PhysPt addr)
+{
+	FPU_Reg temp_reg;
+	// If the value was loaded with the FILD/FIST 64-bit memcpy trick,
+	// read the 64-bit integer value instead.
+	if (fpu.use_regs_memcpy[TOP]) {
+		temp_reg.ll = fpu.regs_memcpy[TOP];
+	} else {
+		double val = FROUND(fpu.regs[TOP].d);
+		temp_reg.ll = (val <= static_cast<double>(INT64_MAX) &&
+		               val >= static_cast<double>(INT64_MIN))
+		                    ? static_cast<int64_t>(val)
+		                    : LONGTYPE(0x8000000000000000);
+	}
 
-	mem_writed(addr,blah.l.lower);
-	mem_writed(addr+4,blah.l.upper);
+	mem_writed(addr, temp_reg.l.lower);
+	mem_writed(addr + 4, temp_reg.l.upper);
 }
 
 static void FPU_FBST(PhysPt addr) {
@@ -409,19 +441,26 @@ static void FPU_FSUBR(Bitu st, Bitu other){
 }
 
 static void FPU_FXCH(Bitu st, Bitu other){
-	FPU_Tag tag = fpu.tags[other];
-	FPU_Reg reg = fpu.regs[other];
-	fpu.tags[other] = fpu.tags[st];
-	fpu.regs[other] = fpu.regs[st];
-	fpu.tags[st] = tag;
-	fpu.regs[st] = reg;
+	const auto tag             = fpu.tags[other];
+	const auto use_reg_memcpy  = fpu.use_regs_memcpy[other];
+	const auto reg             = fpu.regs[other];
+	const auto reg_memcpy      = fpu.regs_memcpy[other];
+	fpu.tags[other]            = fpu.tags[st];
+	fpu.use_regs_memcpy[other] = fpu.use_regs_memcpy[st];
+	fpu.regs[other]            = fpu.regs[st];
+	fpu.regs_memcpy[other]     = fpu.regs_memcpy[st];
+	fpu.tags[st]               = tag;
+	fpu.use_regs_memcpy[st]    = use_reg_memcpy;
+	fpu.regs[st]               = reg;
+	fpu.regs_memcpy[st]        = reg_memcpy;
 }
 
 static void FPU_FST(Bitu st, Bitu other){
-	fpu.tags[other] = fpu.tags[st];
-	fpu.regs[other] = fpu.regs[st];
+	fpu.tags[other]            = fpu.tags[st];
+	fpu.use_regs_memcpy[other] = fpu.use_regs_memcpy[st];
+	fpu.regs[other]            = fpu.regs[st];
+	fpu.regs_memcpy[other]     = fpu.regs_memcpy[st];
 }
-
 
 static void FPU_FCOM(Bitu st, Bitu other){
 	if(((fpu.tags[st] != TAG_Valid) && (fpu.tags[st] != TAG_Zero)) || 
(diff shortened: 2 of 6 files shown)


johnnovak:
Great stuff @kklobe. 🎖️  I guess we'll need to just merge it after some initial testing and live with it during the RC period.

Can you write a more detailed comment explaining how this hack is used, how it manifests in vertical bars (usually black), and mention a few games/demos that need it? Otherwise, people will be scratching their heads in the future... it's really not obvious from the code why this is necessary.

include/fpu.h
```diff
@@ -68,6 +68,9 @@ struct FPU_P_Reg {
 	uint32_t d2 = 0;
 };
 
+// Used for FILD/FIST 64-bit memcpy fix
+constexpr uint8_t TAG_MEMCPY = 4;
```
johnnovak, line 72: Minor: TagMemCpy or something.

src/fpu/fpu_instructions.h
```diff
@@ -20,6 +20,8 @@
 #include "fpu.h"
 #endif
 
+static constexpr int64_t int53_min = -(1LL << 53);
+static constexpr int64_t int53_max = (1LL << 53) - 1;
```
johnnovak, line 24: Int53Min, etc.

src/fpu/fpu_instructions.h
```diff
@@ -198,7 +200,18 @@ static void FPU_FLD_I64(PhysPt addr,Bitu store_to) {
 	FPU_Reg blah;
 	blah.l.lower = mem_readd(addr);
 	blah.l.upper = mem_readd(addr+4);
-	fpu.regs[store_to].d = static_cast<Real64>(blah.ll);
+
+	// Only assign values that fit in the 53-bit mantissa. If the
+	// value exceeds that, assume it's the FILD/FIST 64-bit memcpy trick
+	// and memcpy the value into the FPU register.
+	if (blah.ll <= int53_max && blah.ll >= int53_min) {
```
johnnovak, line 207: Can we take the opportunity to rename `blah` to something else? Please? 😅

src/fpu/fpu_instructions.h
```diff
@@ -266,10 +279,19 @@ static void FPU_FST_I32(PhysPt addr) {
 	mem_writed(addr,(val < 2147483648.0 && val >= -2147483648.0)?static_cast<int32_t>(val):0x80000000);
 }
 
-static void FPU_FST_I64(PhysPt addr) {
-	double val = FROUND(fpu.regs[TOP].d);
+static void FPU_FST_I64(PhysPt addr)
+{
 	FPU_Reg blah;
-	blah.ll = (val < 9223372036854775808.0 && val >= -9223372036854775808.0)?static_cast<int64_t>(val):LONGTYPE(0x8000000000000000);
+	// If the value was loaded with the FILD/FIST 64-bit memcpy trick,
```
johnnovak, line 285: `blah`... 🤦🏻

kklobe:
> Great stuff @kklobe. 🎖️ I guess we'll need to just merge it after some initial testing and live with it during the RC period.
> 
> Can you write a more detailed comment explaining how this hack is used, how it manifests in vertical bars (usually black), and mention a few games/demos that need it? Otherwise, people will be scratching their heads in the future... it's really not obvious from the code why this is necessary.

Absolutely, I plan on adding more context, I just wanted to get this thing up for testing ASAP.

johnnovak:
Tried a few demos and apps, and this works splendidly. Bear in mind that very very few DOS games and demos use the FPU, so 99% games & demos won't exercise the FPU paths _at all_. Also bear in mind that this trick only made sense in the Pentium era, so from 1995 onwards or so. You just won't see it in any earlier programs. Just saying before someone starts regression testing all pre-1990 CGA/EGA games with this fix in place 😎 

Otherwise, I'd say let's just merge it and see how it goes. All my testing has been positive so far! Outstanding work @kklobe 🚀

Grounded0:
@johnnovak 

You're not the only one mixing these two up. At least now the PRs are correctly linked to the correct issues.

Grounded0 (approved):

johnnovak (approved):

johnnovak:
Is there anything outstanding on this @kklobe? If not, let's press the Big Green Button 😎

kklobe:
> Is there anything outstanding on this @kklobe? If not, let's press the Big Green Button 😎

Actually if you could give me just a bit more time - I have an idea for a little more robust fix when the value to be copied is outside the 53-bit integer range, shouldn't take long.

kklobe:
> > Is there anything outstanding on this @kklobe? If not, let's press the Big Green Button 😎
> 
> Actually if you could give me just a bit more time - I have an idea for a little more robust fix when the value to be copied is outside the 53-bit integer range, shouldn't take long.

OK @johnnovak, I improved the out-of-range logic and fixed a couple other bugs. I ran a bunch of tests, but maybe give it another quick whirl on your preferred vertical bar apps.

And, if you were implying I should merge - I don't have (and don't want) "press the green button" permissions :grin:

src/fpu/fpu_instructions.h
```diff
@@ -34,6 +34,10 @@ static void FPU_FINIT(void) {
 	fpu.tags[6] = TAG_Empty;
 	fpu.tags[7] = TAG_Empty;
 	fpu.tags[8] = TAG_Valid; // is only used by us
+
+	for (auto& use_regs_memcpy : fpu.use_regs_memcpy) {
+		use_regs_memcpy = false;
+	}
```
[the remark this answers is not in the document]
kklobe: That doesn't compile, the loop shouldn't really be a big deal for a one-time init.

src/fpu/fpu_instructions.h
kcgen: `blah`, being a potential placeholder world in English, was recommended above less flexible interjectives like `argh`, `ugh`, `grrr`, and `doh` 🤦

kklobe:
this raises a philosophical question: since integers between INT53_MAX and INT64_MAX likely will not be the same after storing and loading, should we stick with the old behavior, which returns a slightly inaccurate integer, or should we return a NaN, as I was doing? Is it better to be inaccurate but close or have the app in question barf on a NaN?

For example, here's a quick test:

```cpp
#include <cstdio>

#define INT53_MAX ((1LL << 53) - 1)
#define INT53_MIN -(1LL << 53)

int main(int argc, char *argv[]) {
	for (int64_t i = INT53_MAX - 10; i<INT53_MAX + 10;i++) {
		double dbl = static_cast<double>(i);
		const int64_t casted = static_cast<int64_t>(dbl);
		printf("%lld, %f, %lld, %s\n", i, dbl, casted, i == casted ? "PASS" : "FAIL");
	}
}
```

and the output:

```log
9007199254740981, 9007199254740981.000000, 9007199254740981, PASS
9007199254740982, 9007199254740982.000000, 9007199254740982, PASS
9007199254740983, 9007199254740983.000000, 9007199254740983, PASS
9007199254740984, 9007199254740984.000000, 9007199254740984, PASS
9007199254740985, 9007199254740985.000000, 9007199254740985, PASS
9007199254740986, 9007199254740986.000000, 9007199254740986, PASS
9007199254740987, 9007199254740987.000000, 9007199254740987, PASS
9007199254740988, 9007199254740988.000000, 9007199254740988, PASS
9007199254740989, 9007199254740989.000000, 9007199254740989, PASS
9007199254740990, 9007199254740990.000000, 9007199254740990, PASS
9007199254740991, 9007199254740991.000000, 9007199254740991, PASS
9007199254740992, 9007199254740992.000000, 9007199254740992, PASS
9007199254740993, 9007199254740992.000000, 9007199254740992, FAIL
9007199254740994, 9007199254740994.000000, 9007199254740994, PASS
9007199254740995, 9007199254740996.000000, 9007199254740996, FAIL
9007199254740996, 9007199254740996.000000, 9007199254740996, PASS
9007199254740997, 9007199254740996.000000, 9007199254740996, FAIL
9007199254740998, 9007199254740998.000000, 9007199254740998, PASS
9007199254740999, 9007199254741000.000000, 9007199254741000, FAIL
9007199254741000, 9007199254741000.000000, 9007199254741000, PASS
```

EDIT: nevermind for now, we'll leave the logic the way it's always been and address any bugs resulting from legitimate use of large numbers for non-memcpy-trick scenarios when (and if) they arise.

kcgen:
> this raises a philosophical question: since integers between INT53_MAX and INT64_MAX likely will not be the same after storing and loading, should we stick with the old behavior, which returns a slightly inaccurate integer, or should we return a NaN, as I was doing? Is it better to be inaccurate but close or have the app in question barf on a NaN?
> 
> For example, here's a quick test:
> 
> ```c++
> #include <cstdio>
> 
> #define INT53_MAX ((1LL << 53) - 1)
> #define INT53_MIN -(1LL << 53)
> 
> int main(int argc, char *argv[]) {
> 	for (int64_t i = INT53_MAX - 10; i<INT53_MAX + 10;i++) {
> 		double dbl = static_cast<double>(i);
> 		const int64_t casted = static_cast<int64_t>(dbl);
> 		printf("%lld, %f, %lld, %s\n", i, dbl, casted, i == casted ? "PASS" : "FAIL");
> 	}
> }
> ```
> 
> and the output:
> 
> ```
> 9007199254740981, 9007199254740981.000000, 9007199254740981, PASS
> 9007199254740982, 9007199254740982.000000, 9007199254740982, PASS
> 9007199254740983, 9007199254740983.000000, 9007199254740983, PASS
> 9007199254740984, 9007199254740984.000000, 9007199254740984, PASS
> 9007199254740985, 9007199254740985.000000, 9007199254740985, PASS
> 9007199254740986, 9007199254740986.000000, 9007199254740986, PASS
> 9007199254740987, 9007199254740987.000000, 9007199254740987, PASS
> 9007199254740988, 9007199254740988.000000, 9007199254740988, PASS
> 9007199254740989, 9007199254740989.000000, 9007199254740989, PASS
> 9007199254740990, 9007199254740990.000000, 9007199254740990, PASS
> 9007199254740991, 9007199254740991.000000, 9007199254740991, PASS
> 9007199254740992, 9007199254740992.000000, 9007199254740992, PASS
> 9007199254740993, 9007199254740992.000000, 9007199254740992, FAIL
> 9007199254740994, 9007199254740994.000000, 9007199254740994, PASS
> 9007199254740995, 9007199254740996.000000, 9007199254740996, FAIL
> 9007199254740996, 9007199254740996.000000, 9007199254740996, PASS
> 9007199254740997, 9007199254740996.000000, 9007199254740996, FAIL
> 9007199254740998, 9007199254740998.000000, 9007199254740998, PASS
> 9007199254740999, 9007199254741000.000000, 9007199254741000, FAIL
> 9007199254741000, 9007199254741000.000000, 9007199254741000, PASS
> ```

When the int64 value goes beyond 53-bits in size, we could "spillover" those lost bits of precision into a new `uint16_t` variable and then  we  recover it on the backside when casting the double back to an `int64_t`.

Something like this:

```c++
#include <cassert>
#include <cmath>
#include <cstdint>
#include <cstdio>

#define INT53_MAX ((1LL << 53) - 1)
#define INT53_MIN -(1LL << 53)
#define SPILLBITS 12

using double_spillover_t = std::pair<double, uint16_t>;

static double_spillover_t
maybe_spillover(const int64_t value)
{
  const auto in_range = (std::abs(value) <= INT53_MAX);
  if (in_range) {
    return { static_cast<double>(value), 0 };
  }
  return { static_cast<double>(value >> SPILLBITS),
           static_cast<uint16_t>(value & 0b1111'1111'1111) };
}

static constexpr int64_t
to_int64_from_spillover(const double_spillover_t ds)
{
  return ds.second == 0
           ? static_cast<int64_t>(ds.first)
           : ((static_cast<int64_t>(ds.first) << SPILLBITS) | ds.second);
}

int
main(int argc, char* argv[])
{
  for (int64_t i = INT53_MAX + 1; i < INT64_MAX; i *= 10) {
    const auto ds = maybe_spillover(i);
    const int64_t casted = to_int64_from_spillover(ds);
    if (i != casted) {
      printf("FAIL: %lld, %f, %lld", i, ds.first, casted);
    }
  }
  return 0;
}
```

johnnovak:
> When the int64 value goes beyond 53-bits in size, we could "spillover" those lost bits of precision (when casting to `double`) into a new `uint16_t` variable and then we recover the bits on backside when casting the `double` back to an `int64_t`.

I think that would be the safest—the question is how "hot" this path would be and the performance impact compared to returning slighlty inaccurate results. I don't think returning NaN is a good idea; at the very least we want the old behaviour I think.

kklobe:
> > When the int64 value goes beyond 53-bits in size, we could "spillover" those lost bits of precision (when casting to `double`) into a new `uint16_t` variable and then we recover the bits on backside when casting the `double` back to an `int64_t`.
> 
> I think that would be the safest—the question is how "hot" this path would be and the performance impact compared to returning slighlty inaccurate results. I don't think returning NaN is a good idea; at the very least we want the old behaviour I think.

I'm not even sure this would help the scenario I'm thinking of: when someone isn't caliing FPU_FST_I64 as part of the memcpy FILD/FIST pair. Let's say they do a direct load of the FP register that ends up in fpu.regs[TOP] (the code in the `else` block below), and it's > int53_max but <= int64_max, but they really do want that large value. We've already lost the precision at that point.

```cpp
static void FPU_FST_I64(PhysPt addr)
{
	FPU_Reg temp_reg;
	// If the value was loaded with the FILD/FIST 64-bit memcpy trick,
	// read the 64-bit integer value instead.
	if (fpu.use_regs_memcpy[TOP]) {
		temp_reg.ll = fpu.regs_memcpy[TOP];
	} else {
		double val = FROUND(fpu.regs[TOP].d);
		temp_reg.ll = (val <= static_cast<double>(INT64_MAX) &&
		               val >= static_cast<double>(INT64_MIN))
		                    ? static_cast<int64_t>(val)
		                    : LONGTYPE(0x8000000000000000);
	}

	mem_writed(addr, temp_reg.l.lower);
	mem_writed(addr + 4, temp_reg.l.upper);
}

```
One thought I had was that we could set the floating point exception for Invalid Operation and/or Loss of Precision. From the Intel CPU ref for `FIST`:

```log
Floating-Point Exceptions
#IS Stack underflow occurred.
#IA Converted value is too large for the destination format.
Source operand is an SNaN, QNaN, ±∞, or unsupported format. #P Value cannot be represented exactly in destination format.
```

But yes, let's keep the old inaccurate behavior for now, until we have a test case in a real app 😁

johnnovak:
> But yes, let's keep the old inaccurate behavior for now, until we have a test case in a real app 😁

Not broken, don't fix it 😎 Sounds good to me.

kklobe:
Sorry guys - added quadword memory functions and used them in the FPU load/store functions, a bit more to review.

kklobe:
OK, fixed most of the embarrassing mistakes, ready for scrutiny.

kklobe:
As far as the quadword functions go, I saw that the `FPU_FST_F64` and `FPU_FLD_F64` are very heavily called, both in the games that need the memcpy hack, as well as those that don't.

Quake calls them several hundred times per millisecond:

```log
2023-11-27 22:31:14.386 | FPU_FST_F64
2023-11-27 22:31:14.386 | FPU_FST_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
2023-11-27 22:31:14.386 | FPU_FST_F64
2023-11-27 22:31:14.386 | FPU_FLD_F64
```

So I figured cutting the memory-related calls in half was a good thing, and should benefit many more games than just the memcpy-hack ones.

kklobe:
setting this back to DRAFT since I rebased on @johnnovak's aspect PR.

johnnovak:
Are we done? Merge time @kklobe ?

kklobe:
> Are we done? Merge time @kklobe ?

No further changes from me for now, let's do it.

Issue #3145, opened 2023-11-21
/src/cpu/meson.build core selection is incorrect for ppc64


casey-ac:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-alpha

### What Operating System are you using?

Other

### If Other OS, please describe

Linux PPC64 kernel, 32-bit userland (Debian derivative)

### Relevant hardware info

PowerPC 770MP x2, Radeon HD6450

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

DOSBox-Staging does not currently build on 64-bit PowerPC (PPC64) systems due to an error in /src/cpu/meson.build. This file assigns the target "PPC64LE" for PPC64; however, PPC64 is a big-endian architecture. If the target is changed to "POWERPC" as specified below the build will complete successfully. 

Line 26 currently:  
    [ ['ppc64',   'powerpc64'],   ['auto', 'dynrec'],  'C_DYNREC',      'PPC64LE', 1,         1 ], # 64 bit PPC processors

However, it should specify:  
    [ ['ppc64',   'powerpc64'],   ['auto', 'dynrec'],  'C_DYNREC',      'POWERPC', 1,         1 ], # 64 bit PPC processors

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Attempt to build DOSBox-Staging on a PPC64 based system
2.
3.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Grounded0:
I don't think its a good idea to start compiling 32-bit big endian binaries for `ppc64` target but pinging @classilla as our PowerPC person.

classilla:
No, reporter is correct, `ppc64` is big-endian and `ppc64le` is little. However, I have no idea if the dynrec JIT backend would work properly on big systems, so we'd need to check if it does (and disable it if not). There's not much that requires a particular endianness and I tried to write with that in mind but it is certainly not tested.

Torinde:
What is "PowerPC 770MP" and where is it used? I found 750 and 970, but no 770...

Another PowerPC build that fails for a long time is the Fedora ppc64le:
https://github.com/dosbox-staging/dosbox-staging/actions/runs/6945073363
![image](https://github.com/dosbox-staging/dosbox-staging/assets/97228894/0a47bdde-ba6b-4f47-9499-d3c2e4371421)

PR #3155, opened 2023-11-27
Fix compilation failures on *BSD and colour rendering on big-endian platforms

### LINKED ISSUE #3144: S3 Colors Incorrect on PowerPC big-endian platforms

0.81.0-alpha

Other

Linux PPC64 kernel, 32-bit userland (Debian derivative)

### Relevant hardware info

IBM PowerPC 970MP x2, Radeon HD6450

DOSBox Staging on PowerPC Linux displays incorrect colors for VGA modes. For example, red and green channels seem to be flipped. However, EGA does not have this issue. Additionally, if a screenshot is taken using the in-app Ctrl+F5 method, the colors will be correct, but using an out-of-app screenshot utility will show the incorrect colors. Changing renderer to texture/software also does not have an impact. This bug seems to be very similar to a bug reported for DOSBox-X:  [https://github.com/joncampbell123/dosbox-x/pull/3428](url)

I am happy to patch/test any potential solutions. 

VGA:  
dosbox-staging_VGA-S3

EGA:  
dosbox-staging_ega

Theme Hospital In-Game Example (note the grass is red):  
dosbox-staging_theme-hospital_colors

Explain how to reproduce
1. Launch DOSBox Staging on PowerPC Linux
2.
3.

_No response_

### Your configuration

```ini
https://gist.github.com/casey-ac/1c60eeb7c77c63ef592e937f0711c04f
```

### Provide a Log

```text
https://gist.github.com/casey-ac/75c4e1f56345530cf9682423d79a1a3c
```

- [X] Yes, I agree.

kcgen:
# Description

- Fix binary compilation failures on FreeBSD.

- Fix unit test compilation and link errors on FreeBSD.

- Fix non-palette colour channel rendering and image capture on big-endian systems.

Unfortunately I've only been able to test `output=texture` combined with `texturerenderer=software` because FreeBSD's PPC big-endian AMD OpenGL drivers aren't compatible with my chipset; so need more checks to verify the OpenGL code paths.

## Related issues

Fixes #3144

# Manual testing

2023-11-27-085638_1280x854_scrot

2023-11-27-085502_1280x854_scrot

2023-11-27-085403_1280x854_scrot

2023-11-27-085350_1280x854_scrot

2023-11-27-085255_1280x854_scrot

2023-11-27-085230_1280x854_scrot

2023-11-27-085204_1280x854_scrot

2023-11-27-085137_1280x854_scrot

# Checklist

I have:

### DIFF
--- a/.muon.ini
+++ b/.muon.ini
@@ -1,4 +1,3 @@
-max_line_len=80
 indent_by='    '
 space_array=false
 wide_colon=false
--- a/include/bgrx8888.h
+++ b/include/bgrx8888.h
@@ -1,0 +1,81 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOSBOX_BGRX8888_H
+#define DOSBOX_BGRX8888_H
+
+#include <cstdint>
+
+// A class that holds the colour channels as an array of four 8-bit values
+// in byte order (also known as memory order), regardless of host
+// endianness. The bytes in memory are always Blue, Green, Red, and finally
+// a placeholder.
+//
+class Bgrx8888 {
+public:
+	constexpr Bgrx8888() = default;
+
+	Bgrx8888(const uint8_t b8, const uint8_t g8, const uint8_t r8)
+	{
+		Set(b8, g8, r8);
+	}
+
+	void Set(const uint8_t b8, const uint8_t g8, const uint8_t r8)
+	{
+		colour.components = {b8, g8, r8, 0};
+	}
+
+	constexpr uint8_t Blue8() const
+	{
+		return colour.components.b8;
+	}
+
+	constexpr uint8_t Green8() const
+	{
+		return colour.components.g8;
+	}
+
+	constexpr uint8_t Red8() const
+	{
+		return colour.components.r8;
+	}
+
+	// Cast operator to read-only uint32_t
+	constexpr operator uint32_t() const
+	{
+		return colour.bgrx8888;
+	}
+
+private:
+	union {
+		struct {
+			// byte order
+			uint8_t b8 = 0;
+			uint8_t g8 = 0;
+			uint8_t r8 = 0;
+			uint8_t x8 = 0;
+		} components;
+		uint32_t bgrx8888 = 0;
+	} colour = {};
+};
+
+static_assert(sizeof(Bgrx8888) == sizeof(uint32_t));
+
+#endif
--- a/include/video.h
+++ b/include/video.h
@@ -144,7 +144,6 @@ struct VideoMode {
 
 std::string to_string(const VideoMode& video_mode);
 
-
 enum class PixelFormat : uint8_t {
 	// Up to 256 colours, paletted;
 	// stored as packed uint8 data
@@ -177,9 +176,9 @@ enum class PixelFormat : uint8_t {
 	RGB565_Packed16 = 16,
 
 	// 16.7M (24-bit) true colour, 8 bits per red/blue/green component;
-	// stored as packed 24-bit data
+	// stored as a sequence of three packed uint8_t values in BGR byte
+	// order, also known as memory order. This format is endian-agnostic.
 	//
-	// Stored as array of uint8_t in BGR memory order (endian agnostic)
 	// Example:
 	// uint8_t *pixels = image.image_data;
 	// pixels[0] = blue; pixels[1] = green; pixels[2] = red;
@@ -188,18 +187,18 @@ enum class PixelFormat : uint8_t {
 	// FFmpeg Equivalent: AV_PIX_FMT_BGR24
 	BGR24_ByteArray = 24,
 
-	// 16.7M (32-bit) true colour; 8 bits per red/blue/green component;
-	// stored as packed uint32 data with highest 8 bits unused
+	// Same as BGR24_ByteArray but padded to 32-bits. 16.7M true colour,
+	// 8 bits per red/blue/green/empty component; stored as a sequence of
+	// four packed uint8_t values in BGRX byte order, also known as
+	// memory order. This format is endian-agnostic.
 	//
-	// Stored as array of uint32_t in host native endianess.
-	// Each uint32_t is layed out as follows:
-	// (msb)8X 8R 8G 8B(lsb)
 	// Example:
-	// uint32_t pixel = (unused << 24) | (red << 16) | (green << 8) | (blue << 0)
+	// uint8_t *pixels = image.image_data;
+	// pixels[0] = blue; pixels[1] = green; pixels[2] = red; pixels[3] = empty;
 	//
-	// SDL Equivalent: SDL_PIXELFORMAT_XRGB8888
-	// FFmpeg Equivalent: AV_PIX_FMT_0RGB32
-	XRGB8888_Packed32 = 32
+	// SDL has has no equivalent.
+	// FFmpeg Equivalent: BGRX32_ByteArray
+	BGRX32_ByteArray = 32,
 };
 
 const char* to_string(const PixelFormat pf);
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -367,17 +367,35 @@ static void create_avi_file(const uint16_t width, const uint16_t height,
 static void compress_frame(const RenderedImage& image)
 {
 	const auto& src = image.params;
-
 	auto src_row = image.image_data;
-	uint8_t tgt_row[SCALER_MAXWIDTH * 4] = {};
 
 	const auto pixel_multiple = (src.double_width ? 2 : 1);
 	const auto row_multiple   = (src.double_height ? 2 : 1);
 
+	auto compress_row = [&](const uint8_t* row_buffer) {
+		for (auto i = 0; i < row_multiple; ++i) {
+			video.codec->CompressLines(1, &row_buffer);
+		}
+	};
+
 	const auto src_bpp = to_bytes_per_pixel(src.pixel_format);
 	const auto tgt_bpp = to_bytes_per_pixel(to_zmbv_format(src.pixel_format));
 
-	// Arrange the source bytes into a target row before compressing
+	// Maybe compress the source rows straight away without copying. Note
+	// that this is a shortcut scenario; hard-code it to false to exercise
+	// the rote version below.
+	const auto can_use_src_directly = (src_bpp == tgt_bpp) && !src.double_width;
+	if (can_use_src_directly) {
+		for (auto i = 0; i < src.height; ++i, src_row += image.pitch) {
+			compress_row(src_row);
+		}
+		return;
+	}
+
+	// Otherwise we need to arrange the source bytes before compressing
+	assert(!can_use_src_directly);
+	uint8_t tgt_row[SCALER_MAXWIDTH * 4] = {};
+
 	for (auto i = 0; i < src.height; ++i, src_row += image.pitch) {
 		auto src_pixel = src_row;
 		auto tgt_pixel = tgt_row;
@@ -386,10 +404,7 @@ static void compress_frame(const RenderedImage& image)
 				std::memcpy(tgt_pixel, src_pixel, src_bpp);
 			}
 		}
-		for (auto j = 0; j < row_multiple; ++j) {
-			const uint8_t* row_pointer = tgt_row;
-			video.codec->CompressLines(1, &row_pointer);
-		}
+		compress_row(tgt_row);
 	}
 }
 
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -22,6 +22,7 @@
 #include "capture.h"
 
 #include <cassert>
+#include <cmath>
 
 #include "mem.h"
 #include "render.h"
@@ -61,6 +62,44 @@ static struct {
 		uint32_t bytes_written   = 0;
 	} audio = {};
 } video = {};
+
+static ZMBV_FORMAT to_zmbv_format(const PixelFormat format)
+{
+	switch (format) {
+	case PixelFormat::Indexed8: return ZMBV_FORMAT::BPP_8;
+	case PixelFormat::RGB555_Packed16: return ZMBV_FORMAT::BPP_15;
+	case PixelFormat::RGB565_Packed16: return ZMBV_FORMAT::BPP_16;
+
+	// ZMBV is "the DOSBox capture format" supported by external tools such
+	// as VLC, MPV, and ffmpeg. Because DOSBox originally didn't have 24-bit
+	// colour, the format itself doesn't support it and treats 4-byte
+	// formats as byte-ordered BGR colour values.
+	//
+	case PixelFormat::BGR24_ByteArray: return ZMBV_FORMAT::BPP_32;
+	case PixelFormat::BGRX32_ByteArray: return ZMBV_FORMAT::BPP_32;
+	default: assertm(false, "Invalid pixel_format value"); break;
+	}
+	return ZMBV_FORMAT::NONE;
+}
+
+// Returns number of bytes needed to represent the given PixelFormat
+static uint8_t to_bytes_per_pixel(const PixelFormat format)
+{
+	// PixelFormat's underly value is colour depth in number of bits
+	const float num_bits = static_cast<uint8_t>(format);
+
+	// Use float-ceil to handle non-power-of-two bit quantities
+	constexpr uint8_t bits_per_byte = 8;
+	const auto num_bytes = iround(ceilf(num_bits / bits_per_byte));
+
+	assert(num_bytes >= 1 && num_bytes <= 4);
+	return static_cast<uint8_t>(num_bytes);
+}
+
+static uint8_t to_bytes_per_pixel(const ZMBV_FORMAT format)
+{
+	return ZMBV_ToBytesPerPixel(format);
+}
 
 static void add_avi_chunk(const char* tag, const uint32_t size,
                           const void* data, const uint32_t flags)
@@ -334,22 +373,7 @@ void capture_video_add_frame(const RenderedImage& image, const float frames_per_
 		capture_video_finalise();
 	}
 
-	ZMBV_FORMAT format;
-
-	switch (src.pixel_format) {
-	case PixelFormat::Indexed8: format = ZMBV_FORMAT::BPP_8; break;
-	case PixelFormat::RGB555_Packed16: format = ZMBV_FORMAT::BPP_15; break;
-	case PixelFormat::RGB565_Packed16: format = ZMBV_FORMAT::BPP_16; break;
-
-	// ZMBV is "the DOSBox capture format" supported by external
-	// tools such as VLC, MPV, and ffmpeg. Because DOSBox originally
-	// didn't have 24-bit color, the format itself doesn't support
-	// it. I this case we tell ZMBV the data is 32-bit and let the
-	// rgb24's int() cast operator up-convert.
-	case PixelFormat::BGR24_ByteArray: format = ZMBV_FORMAT::BPP_32; break;
-	case PixelFormat::BGRX32_ByteArray: format = ZMBV_FORMAT::BPP_32; break;
-	default: assertm(false, "Invalid pixel_format value"); return;
-	}
+	const auto format = to_zmbv_format(src.pixel_format);
 
 	if (!video.handle) {
 		create_avi_file(video_width,
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -394,17 +394,24 @@ static void compress_frame(const RenderedImage& image)
 
 	// Otherwise we need to arrange the source bytes before compressing
 	assert(!can_use_src_directly);
-	uint8_t tgt_row[SCALER_MAXWIDTH * 4] = {};
+
+	// Grow the target row buffer if needed
+	const auto tgt_row_bytes = check_cast<uint16_t>(src.width * tgt_bpp *
+	                                                pixel_multiple);
+	static std::vector<uint8_t> tgt_row = {};
+	if (tgt_row.size() < tgt_row_bytes) {
+		tgt_row.resize(tgt_row_bytes, 0);
+	}
 
 	for (auto i = 0; i < src.height; ++i, src_row += image.pitch) {
 		auto src_pixel = src_row;
-		auto tgt_pixel = tgt_row;
+		auto tgt_pixel = tgt_row.data();
 		for (auto j = 0; j < src.width; ++j, src_pixel += src_bpp) {
 			for (auto k = 0; k < pixel_multiple; ++k, tgt_pixel += tgt_bpp) {
 				std::memcpy(tgt_pixel, src_pixel, src_bpp);
 			}
 		}
-		compress_row(tgt_row);
+		compress_row(tgt_row.data());
 	}
 }
 
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -358,6 +358,42 @@ static void create_avi_file(const uint16_t width, const uint16_t height,
 	video.audio.bytes_written   = 0;
 }
 
+// Reads the image's pixels from the emulated DOS video RAM (which is always
+// little-endian) and passes them in the same byte-order to the encoder.
+//
+// Ideally this could just be one memcpy, but many DOS video modes need their
+// pixels either doubled horizontally and/or vertically to produce the correct
+// on-screen size. So extra copies are needed in both these cases.
+//
+static void compress_frame(const RenderedImage& image)
+{
+	const auto& src = image.params;
+
+	auto src_row = image.image_data;
+	uint8_t tgt_row[SCALER_MAXWIDTH * 4] = {};
+
+	const auto pixel_multiple = (src.double_width ? 2 : 1);
+	const auto row_multiple   = (src.double_height ? 2 : 1);
+
+	const auto src_bpp = to_bytes_per_pixel(src.pixel_format);
+	const auto tgt_bpp = to_bytes_per_pixel(to_zmbv_format(src.pixel_format));
+
+	// Arrange the source bytes into a target row before compressing
+	for (auto i = 0; i < src.height; ++i, src_row += image.pitch) {
+		auto src_pixel = src_row;
+		auto tgt_pixel = tgt_row;
+		for (auto j = 0; j < src.width; ++j, src_pixel += src_bpp) {
+			for (auto k = 0; k < pixel_multiple; ++k, tgt_pixel += tgt_bpp) {
+				std::memcpy(tgt_pixel, src_pixel, src_bpp);
+			}
+		}
+		for (auto j = 0; j < row_multiple; ++j) {
+			const uint8_t* row_pointer = tgt_row;
+			video.codec->CompressLines(1, &row_pointer);
+		}
+	}
+}
+
 void capture_video_add_frame(const RenderedImage& image, const float frames_per_second)
 {
 	const auto& src = image.params;
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -26,7 +26,6 @@
 
 #include "mem.h"
 #include "render.h"
-#include "rgb888.h"
 #include "support.h"
 
 #include "zmbv/zmbv.h"
@@ -432,80 +431,7 @@ void capture_video_add_frame(const RenderedImage& image, const float frames_per_
 		return;
 	}
 
-	alignas(uint32_t) uint8_t row_buffer[SCALER_MAXWIDTH * 4];
-
-	auto src_row = image.image_data;
-
-	for (auto i = 0; i < src.height; ++i) {
-		const uint8_t* row_pointer = row_buffer;
-
-		// TODO This all assumes little-endian byte order; should be
-		// made endianness-aware like capture_image.cpp
-		if (src.double_width) {
-			switch (src.pixel_format) {
-			case PixelFormat::Indexed8:
-				for (auto x = 0; x < src.width; ++x) {
-					const auto pixel      = src_row[x];
-					row_buffer[x * 2 + 0] = pixel;
-					row_buffer[x * 2 + 1] = pixel;
-				}
-				break;
-
-			case PixelFormat::RGB555_Packed16:
-			case PixelFormat::RGB565_Packed16:
-				for (auto x = 0; x < src.width; ++x) {
-					const auto pixel = ((uint16_t*)src_row)[x];
-
-					((uint16_t*)row_buffer)[x * 2 + 0] = pixel;
-					((uint16_t*)row_buffer)[x * 2 + 1] = pixel;
-				}
-				break;
-
-			case PixelFormat::BGR24_ByteArray:
-				for (auto x = 0; x < src.width; ++x) {
-					const auto pixel = reinterpret_cast<const Rgb888*>(
-					        src_row)[x];
-
-					reinterpret_cast<uint32_t*>(
-					        row_buffer)[x * 2 + 0] = pixel;
-					reinterpret_cast<uint32_t*>(
-					        row_buffer)[x * 2 + 1] = pixel;
-				}
-				break;
-
-			case PixelFormat::BGRX32_ByteArray:
-				for (auto x = 0; x < src.width; ++x) {
-					const auto pixel = ((uint32_t*)src_row)[x];
-
-					((uint32_t*)row_buffer)[x * 2 + 0] = pixel;
-					((uint32_t*)row_buffer)[x * 2 + 1] = pixel;
-				}
-				break;
-			}
-			row_pointer = row_buffer;
-
-		} else {
-			if (src.pixel_format == PixelFormat::BGR24_ByteArray) {
-				for (auto x = 0; x < src.width; ++x) {
-					const auto pixel = reinterpret_cast<const Rgb888*>(
-					        src_row)[x];
-
-					reinterpret_cast<uint32_t*>(
-					        row_buffer)[x] = pixel;
-				}
-				row_pointer = row_buffer;
-			} else {
-				row_pointer = src_row;
-			}
-		}
-
-		auto lines_to_write = src.double_height ? 2 : 1;
-		while (lines_to_write--) {
-			video.codec->CompressLines(1, &row_pointer);
-		}
-
-		src_row += image.pitch;
-	}
+	compress_frame(image);
 
 	const auto written = video.codec->FinishCompressFrame();
 	if (written < 0) {
--- a/src/capture/capture_video.cpp
+++ b/src/capture/capture_video.cpp
@@ -69,12 +69,9 @@ static ZMBV_FORMAT to_zmbv_format(const PixelFormat format)
 	case PixelFormat::RGB555_Packed16: return ZMBV_FORMAT::BPP_15;
 	case PixelFormat::RGB565_Packed16: return ZMBV_FORMAT::BPP_16;
 
-	// ZMBV is "the DOSBox capture format" supported by external tools such
-	// as VLC, MPV, and ffmpeg. Because DOSBox originally didn't have 24-bit
-	// colour, the format itself doesn't support it and treats 4-byte
-	// formats as byte-ordered BGR colour values.
-	//
-	case PixelFormat::BGR24_ByteArray: return ZMBV_FORMAT::BPP_32;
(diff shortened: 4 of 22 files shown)


kcgen:
@casey-ac, when you have a chance, can you test this branch on your big-endian system?

src/misc/fs_utils_posix.cpp
```diff
@@ -241,14 +248,18 @@ static bool set_xattr(const std_fs::path& path, const FatAttributeFlags attribut
...
 	constexpr int flags = 0;
 
 	const auto result = setxattr(path.c_str(),
                                      XattrName.c_str(),
                                      xattr.c_str(),
                                      xattr.size(),
                                      flags);
+#else
+	// Platform doesn't support extended attributes
+	// so we always have a failed result-code.
```
johnnovak, line 261: Nitpicking: result code

src/misc/fs_utils_posix.cpp
```diff
@@ -267,14 +278,18 @@ static bool set_xattr(const int file_descriptor, const FatAttributeFlags attribu
...
 	constexpr int flags = 0;
 
 	const auto result = fsetxattr(file_descriptor,
 	                              XattrName.c_str(),
 	                              xattr.c_str(),
 	                              xattr.size(),
 	                              flags);
+#else
+	// Platform doesn't support extended attributes
+	// so we always have a failed result-code.
```
johnnovak, line 291: Nitpicking: result code

src/shell/command_line_stubs.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+// This is a stubbed out implementation replacement for the actual CommandLine
+// class used by the unit test system for two reasons:
+//
+//  1. To avoid pulling in the DOS shell dependencies, which in turn
+//     depend on the greater DOS system.
+//
+//  2. Unit tests that aren't testing the CommandLine class itself
+//     don't need command line parsing functionality as their inputs
+//     are defined in the unit test function bodies.
```
johnnovak, line 29: Looks like a worthy addition! 🚀

include/video.h
```diff
@@ -188,18 +188,18 @@ enum class PixelFormat : uint8_t {
 	// FFmpeg Equivalent: AV_PIX_FMT_BGR24
 	BGR24_ByteArray = 24,
 
-	// 16.7M (32-bit) true colour; 8 bits per red/blue/green component;
-	// stored as packed uint32 data with highest 8 bits unused
+	// 16.7M (32-bit) true colour, 8 bits per red/blue/green component;
+	// stored as packed 32-bit data
```
johnnovak, line 192: "16.7M (32-bit) true colour, 8 bits per red/blue/green component **plus a padding byte**; stored as packed 32-bit data"

Isn't that more accurate and how 32-bit true colour on actual hardware works?

include/video.h
```diff
@@ -188,18 +188,18 @@ enum class PixelFormat : uint8_t {
...
 	//
-	// Stored as array of uint32_t in host native endianess.
-	// Each uint32_t is layed out as follows:
-	// (msb)8X 8R 8G 8B(lsb)
+	// Stored as array of uint8_t in BGRA memory order (endian agnostic)
```
johnnovak, line 194: BGRX no? VGA cards don't have the concept of an alpha channel.

include/video.h
```diff
@@ -188,18 +188,18 @@ enum class PixelFormat : uint8_t {
...
 	// Example:
-	// uint32_t pixel = (unused << 24) | (red << 16) | (green << 8) | (blue << 0)
+	// uint8_t *pixels = image.image_data;
+	// pixels[0] = blue; pixels[1] = green; pixels[2] = red; pixels[3] = alpha;
```
johnnovak, line 197: There is no alpha, the last byte is just padding. So remove `pixels[3] = alpha;`.

include/video.h
```diff
@@ -188,18 +188,18 @@ enum class PixelFormat : uint8_t {
...
 	//
-	// SDL Equivalent: SDL_PIXELFORMAT_XRGB8888
-	// FFmpeg Equivalent: AV_PIX_FMT_0RGB32
-	XRGB8888_Packed32 = 32
+	// SDL Equivalent: SDL_PIXELFORMAT_BGRA32 (alias for SDL_PIXELFORMAT_BGR24)
+	// FFmpeg Equivalent: AV_PIX_FMT_BGR0
+	// OpenGL Equivalent: GL_BGRA with data type GL_UNSIGNED_INT_8_8_8_8_REV
+	BGR32_ByteArray = 32,
```
johnnovak, line 202: So is this `BGRX32` or `BGRA32`? We must specify the meaning of fourth component as well, as per above.

include/bgra8888.h
```diff
@@ -0,0 +1,98 @@
...
+			uint8_t b8 = 0;
+			uint8_t g8 = 0;
+			uint8_t r8 = 0;
+			uint8_t a8 = 0;
+		} components;
+		uint32_t bgra8888 = 0;
+	} colour = {};
+};
+
+static_assert(sizeof(Bgra8888) == sizeof(uint32_t));
```
johnnovak, line 96: Nice 👍🏻 I guess we don't need the pack pragma then.

include/bgra8888.h
```diff
@@ -0,0 +1,98 @@
...
+
+	constexpr uint8_t GetRed8() const
+	{
+		return colour.components.r8;
+	}
+
+	constexpr uint8_t GetAlpha8() const
+	{
+		return colour.components.a8;
+	}
```
johnnovak, line 75: Suggest dropping the `Get` prefixes, it's just too noisy in actual use.

include/bgra8888.h
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+#ifndef DOSBOX_RGBA8888_H
+#define DOSBOX_RGBA8888_H
+
+#include <cstdint>
+
+// A class that holds the colour channels as an array of 8-bit values in memory
+// order (alsa known as byte order), regardless of endianness. The bytes in
+// memory are always Blue, Green, Red then Alpha.
```
johnnovak, line 28: "Class to represent a BGRA8888 pixel stored as an array of four consecutive bytes (so the representation is endianness-agnostic). The bytes in memory are always Blue, Green, Red, and then Alpha."

include/bgra8888.h
```diff
@@ -0,0 +1,98 @@
...
+//
+// This matches SDL's PIXELFORMAT_BGR24 and OpenGL's pixel array formats, both
```
johnnovak, line 30: Correction: `PIXELFORMAT_BGRA24`
Add: `GL_RGBA`

src/hardware/vga_draw.cpp
```diff
@@ -904,12 +905,12 @@ static void VGA_ProcessSplit()
...
+static constexpr uint16_t to_rgb_565(const Bgra8888 bgra_8888)
 {
-	const auto rgb565 = ((rgb888 & 0xf80000) >> 8) +
-	                    ((rgb888 & 0xfc00) >> 5) + ((rgb888 & 0xf8) >> 3);
-
-	return check_cast<uint16_t>(rgb565);
+	const auto rgb_565 = Rgb565(bgra_8888.GetRed8(),
+	                            bgra_8888.GetGreen8(),
+	                            bgra_8888.GetBlue8());
+	return rgb_565.pixel;
```
johnnovak, line 913: Can we move this into `Rgb565` or `Rgba8888`?

tests/rgb_tests.cpp
```diff
@@ -131,3 +134,103 @@ TEST(rgb, linear_to_srgb8_lut)
...
+	// Do the array elements match the BGRA values we put in?
+
+	// Bgra8888 represents the colour value as a byte sequence
+	// with endian-agnostic memory order or byte-order, so the
+	// following tests must pass on big and little endian hosts.
+	//
+	EXPECT_EQ(bgra_array[0], b8);
+	EXPECT_EQ(bgra_array[1], g8);
+	EXPECT_EQ(bgra_array[2], r8);
+	EXPECT_EQ(bgra_array[3], a8);
```
johnnovak, line 235: Nice tests 💯

include/video.h
kcgen: Will do!

include/video.h
kcgen: Will rename to `Bgrx8888` across the board.

include/video.h
```diff
@@ -188,18 +188,18 @@ enum class PixelFormat : uint8_t {
...
+	// Stored as array of uint8_t in BGRA memory order (endian agnostic)
 	// Example:
-	// uint32_t pixel = (unused << 24) | (red << 16) | (green << 8) | (blue << 0)
+	// uint8_t *pixels = image.image_data;
+	// pixels[0] = blue; pixels[1] = green; pixels[2] = red; pixels[3] = alpha;
 	//
-	// SDL Equivalent: SDL_PIXELFORMAT_XRGB8888
-	// FFmpeg Equivalent: AV_PIX_FMT_0RGB32
-	XRGB8888_Packed32 = 32
+	// SDL Equivalent: SDL_PIXELFORMAT_BGRA32 (alias for SDL_PIXELFORMAT_BGR24)
```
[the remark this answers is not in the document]
weirddan455: Yes, those two are not aliases for each other.  `SDL_PIXELFORMAT_BGRA32` aliases to either `SDL_PIXELFORMAT_BGRA8888` or `SDL_PIXELFORMAT_ARGB8888` depending on endianess.  These are 4 bytes per pixel,

`SDL_PIXELFORMAT_BGR24 ` is 3 bytes per pixel stored in a byte array.

`SDL_PIXELFORMAT_BGRX32` does not exist in SDL.  You would have to write your own marcro to alias to either `SDL_PIXELFORMAT_BGRX8888` or `SDL_PIXELFORMAT_XRGB8888`  depending on endianess.  I believe SDL internally always loads 32 bit formats as 32 bit packed words so an alias is needed if what you're describing is a byte array.

https://github.com/libsdl-org/SDL/blob/15ead9a40d09a1eb9972215cceac2bf29c9b77f6/include/SDL_pixels.h#L273-L284

I wrote these comments by the way.  @johnnovak and I went back and forth for quite a while on how to best describe these pixel formats.  It can be quite confusing and different libraries/vendors have different ways of describing them.

peteroupc:
> Unfortunately I've only been able to test `output=texture` combined with `texturerenderer=software` because FreeBSD's PPC big-endian AMD OpenGL drivers aren't compatible with my chipset[.]

This may be a known problem with the Mesa graphics library; see:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/?sort=closed_at_desc&state=opened&search=powerpc&first_page_size=20

See also:

- https://gitlab.freedesktop.org/mesa/mesa/-/issues/4190
- https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/8767

src/capture/capture_video.cpp
```diff
@@ -413,7 +413,7 @@ void capture_video_add_frame(const RenderedImage& image, const float frames_per_
 				}
 				break;
 
-			case PixelFormat::XRGB8888_Packed32:
+			case PixelFormat::BGR32_ByteArray:
```
weirddan455, line 416: Has video capture been tested?  If I recall correctly, ZMBV expects a packed 32 bit format in host native endianess.  Since you've changed this to a byte array, we might need to change some logic here.

src/capture/image/image_decoder.h
```diff
@@ -106,32 +106,24 @@ class ImageDecoder {
 
 		switch (image.params.pixel_format) {
 		case PixelFormat::RGB555_Packed16: {
-			const auto p = read_unaligned_uint16(pos);
+			const auto p = host_readw(pos);
```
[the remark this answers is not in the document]
kklobe: the `host_readx` functions pass through the functions to adjust for the host endianness :

```cpp
// Read a 16-bit word from 8-bit DOS/little-endian byte-ordered memory.
static inline uint16_t host_readw(const uint8_t* const arr) noexcept
{
	return le16_to_host(read_unaligned_uint16(arr));
}
```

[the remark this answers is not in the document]
weirddan455: Thanks @kklobe for the refresher.

I just checked git blame because I thought this code looked familiar.  I made this change a couple months ago citing an endinaess bug.  97755b29d2bc5ef81ccfd544c372431f2eae2629

My change is essentially being reverted here.  Has the image capturer been tested on big endian to confirm this is correct?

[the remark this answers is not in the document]
kcgen: Yup, I tested image capture (raw scaled and post-rendered) at all colour depths and video modes, including this one (using sdd553a), and it was broken on big-endian.

It was working backward from the image capture colour values (I believe from raw scaled code path) that eventually landed me here :-)

When I read your documentation about this the RGB555_Packed16 enum:

> "stored as packed uint16 data with highest bit unused in host native endianess."

This change was like a puzzle piece, given the change was flipping the DOS-written little-endian 16-bit value (from the emulated S3 VESA memory space) over to a host-native value, which is exactly as you've described.

include/bgra8888.h
johnnovak: On a second thought @kcgen , it's best to remove all SDL and OpenGL pixel format references. Just drop then entirely. It's more complicated than it seems on the surface, so mentioning them doesn't help, only muddies the waters as I'll explain below.

src/capture/capture_video.cpp
johnnovak: Yes, video capture in paletted and 15/16/24/32-bit hi/true colour modes need to be tested as well.

src/capture/image/image_decoder.h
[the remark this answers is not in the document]
johnnovak: > My change is essentially being reverted here. Has the image capturer been tested on big endian to confirm this is correct?

Just confirming, your commit message says "Fix endianness bug in image decoder". Do you remember, did you fix an actual observed problem, or you thought you're fixing an endianness bug, but were not able to actually test it on BE?

If the latter, that explains the situation. Just wanna get to the bottom of it to 100% understand the situation, that's all 😄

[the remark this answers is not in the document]
johnnovak: Btw, I find this function confusing as fuck, as _most_ BE/LE conversion functions in most libraries. I would need to draw to understand it, like always, but if this is what I think it is based on the comment ("Read a 16-bit word from 8-bit DOS/little-endian byte-ordered memory."), `read_from_16le` would be a far better name IMO. At least that's how my brain works 😄 

```ini
static inline uint16_t read_from_16le(const uint8_t* const arr) noexcept
``

File: include/video.h (hunk 14 above)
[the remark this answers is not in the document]
Comment (johnnovak) [reply include/video.h:199]: 1. Yeah, what @weirddan455 writes is true, and I remember this confused me quite a bit too about the SDL pixel formats @kcgen. They're simply a bit sloppy in their pixel format namings, that's the gist of it, and as @weirddan455 said, there's no out-of-the-box SDL pixel format definition for `BGRX32_ByteArray`. It would be best to drop the SDL equivalent line for this altogether.

2. As stated, `BGRX32_ByteArray` is the correct name to use here. I'm 99% sure we just have random garbage in the 4th byte, or maybe all zeroes. Based on your fix to the image decoder in this PR, I'm confident you got it right and this is indeed 4-components as a byte array 😄 

3. "OpenGL Equivalent: GL_BGRA with data type GL_UNSIGNED_INT_8_8_8_8_REV" — that's probably incorrect. `GL_BGRA` just specifies the "register layout", and `GL_UNSIGNED_INT_8_8_8_8_REV` just specifies the memory layout, plus tells the driver to reverse the bytes. You need to use the `*_REV` or non-REV version (`GL_UNSIGNED_INT_8_8_8_8`) depending on whether you're on a BE or LE architecture.

    What you really want is this ([source](https://afrantzis.com/pixel-format-guide/opengl.html)):

    ```
    Example: GL_RGBA with GL_UNSIGNED_BYTE

    This combination specifies a pixel format laid out in memory as the bytes R, G, B, A
    (R at the lowest address, A at the highest)
    ```

    If you really want to include the OpenGL "equivalent", make sure to mention this parenthesis: "(assuming the alpha component is set to 255; what we have here is really `GL_RGBX`, so the 4th byte is unused/undfined, but that has no OpenGL equivalent)".

File: include/video.h (hunk 7 above)
Comment (johnnovak) [reply include/video.h:202]: As per above, I've concluded this is in fact `BGRX32_ByteArray` 😄

File: src/capture/image/image_decoder.h (hunk 16 above)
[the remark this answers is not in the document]
Comment (johnnovak) [reply src/capture/image/image_decoder.h:109]: @weirddan455 All good man, the important thing is we get it right finally right 😅 This is a very confusing topic, indeed, and you've cleaned up & clarified the rest of the pixel formats, so your success rate is pretty high even with getting a single one wrong 😄

File: tests/rgb_tests.cpp [hunk 17]
```diff
@@ -131,3 +134,103 @@ TEST(rgb, linear_to_srgb8_lut)
...
 	}
 }
+
+TEST(rgb, rgb565_in_memory_order)
+{
+	//                                  | red | green| blue|
+	constexpr uint16_t rgb_as_uint16 = 0b00001'000011'00111;
+	constexpr uint8_t r8             = 0b00001'000;
+	constexpr uint8_t g8             = /* */ 0b000011'00;
+	constexpr uint8_t b8             = /*        */ 0b00111'000;
```
johnnovak, line 144: Very nice creative use of alignment 🎉

tests/rgb_tests.cpp
```diff
@@ -131,3 +134,103 @@ TEST(rgb, linear_to_srgb8_lut)
...
+	constexpr uint8_t g6 = 0b100'111;
+	constexpr uint8_t b6 = 0b101'111;
+
+	constexpr Rgb666 rgb_object(r6, g6, b6);
+
+	constexpr auto array_size = sizeof(rgb_object);
+
+	// Copy the object byte-by-byte into an array
+	uint8_t rgb_array[array_size] = {};
+	memcpy(rgb_array, reinterpret_cast<const void*>(&rgb_object), array_size);
```
johnnovak, line 164: I'm a bit uncomfortable about this. The RGB components are just defined as member vars in the class:

```cpp
public:
	uint8_t red   = 0;
	uint8_t green = 0;
	uint8_t blue  = 0;
```

I'd prefer if you converted the class to the `union` / `struct` trick you used in `bgra8888.h`, otherwise this seems a bit hacky to me to use `rgb666` this way.

Plus we'd need some explanatory class comments in `rgb666`, similarly to `bgra8888.h`.

But taking a step back, aren't we always using RGB666 and RGB565 packed to `uint16_t` in the actual code? If not, the proposed change isn't needed, and I'd get rid of the test testing the memory layout.

tests/rgb_tests.cpp
```diff
@@ -131,3 +134,103 @@ TEST(rgb, linear_to_srgb8_lut)
...
+	EXPECT_EQ(rgb_array[0], 0);
+	EXPECT_EQ(rgb_array[1], b8);
+	EXPECT_EQ(rgb_array[2], g8);
+	EXPECT_EQ(rgb_array[3], r8);
+#else
+	// On little endian, the byte-order ia RGBA
+	EXPECT_EQ(rgb_array[0], r8);
+	EXPECT_EQ(rgb_array[1], g8);
+	EXPECT_EQ(rgb_array[2], b8);
+#endif
```
johnnovak, line 205: Similarly, why do we need this? We already have the `BGR24_ByteArray` pixel format. Is it really true that some parts of the code in fact use the RGB24_ByteArray format without having an actual pixel format definition for it by mucking around with casting things to `Rgb888`? If so, that's rather unfortunate...

casey-ac:
Hello! Thank you so much for the work you've put into this fix. 

There was one build issue:  the same issue as is described in [https://github.com/dosbox-staging/dosbox-staging/issues/3145](url). Once "PPC64EL" was changed to "POWERPC", the build finished successfully. The dynrec core seems to function just fine, and with noticeable performance increase (using core=dynamic). 

The color issue appears to be fixed as tested with cga, ega, hercules, svga_s3, and tandy, in both opengl and texture output and captured as both scaled and rendered. 

I also captured svga_s3 opengl and svga_s3 texture. These both worked but the OpenGL version produced a 640x200 video and the texture version appears normal (640x400). 

cga opengl rendered
cga-opengl-rendered

cga opengl scaled
cga-opengl-scaled

cga texture rendered
cga-texture-rendered

cga texture scaled
cga-texture-scaled

ega opengl rendered
ega-opengl-rendered

ega opengl scaled
ega-opengl-scaled

ega texture rendered
ega-texture-rendered

ega texture scaled
ega-texture-scaled

hercules opengl rendered
hercules-opengl-rendered

hercules opengl scaled
hercules-opengl-scaled

hercules texture rendered
hercules-texture-rendered

hercules texture scaled
hercules-texture-scaled

svga_s3 opengl rendered
svga_s3-opengl-rendered

svga_s3 opengl scaled
svga_s3-opengl-scaled

svga_s3 texture rendered
svga_s3-texture-rendered

svga_s3 texture scaled
svga_s3-texture-scaled

tandy opengl rendered
tandy-opengl-rendered

tandy opengl scaled
tandy-opengl-scaled

tandy texture rendered
tandy-texture-rendered

tandy texture scaled
tandy-texture-scaled

Grounded0:
Going in to 0.81 RC and lets make sure 0.81.0 is a big-endian compatible release.

src/capture/image/image_decoder.h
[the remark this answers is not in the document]
kcgen: > Then these read functions get byte pointer args. Similarly for writes, `write_as_16le` starting at the byte-pointer, etc.

> Also, agree that something like `read_from_16le` would be a better name.

@johnnovak , @weirddan455 -> :+1: 

I added the _unaligned_ accessors early in 2020 - 505669046719430186a057ae3f7d5d0033c6a689:

- `read_unaligned_*`
- `read_unaligned_*_at`
- `write_unaligned_*`
- `write_unaligned_*_at`
- `inc_unaligned_*`
- `add_to_unaligned_*`

But these don't involve big/little endian swaps. The `host_read*` and `host_write*` calls have been in the code since 2003.

A bit of context for the `host_`, from my understanding:

DOSBox juggles two worlds, the native host platform space where the business logic and emulation happens, and the emulated x86 hardware and DOS-space, which is always LE.

If we need to pluck values from DOS space into host-space and actually work with them (ie: compare them, check if they're beyond 32000) then it's done with these `host_read` calls.  Think of them as _"the host is reading a value from DOS-land"_.

The `host_write` takes a value from the host and writes it back into DOS memory. Think of it like, _"The host is writing a value into DOS-land"_.

It's likely that the names were chosen through the eyes of DOS, so "the host" is the thing acting on it. 

If we flipped that around to be "through the eyes of the host", we could be the only acting on the DOS space, with calls like: `read_uint32_from_dos` and `write_uint16_to_dos`. 

Both angles-of-attack make the code a bit more self-documenting, and there's probably room to mass-rename them or simply add more aliases, because it might make it clearer.

The big endian fixups are a nice way to catch these boundaries between the host world and DOS world, where previously the code might be more ambiguously working with memory and values - and it becomes hard to tell which "world" we're in.

src/capture/capture_video.cpp
kcgen: Those are working with the changes so far, but I've only been able to hit the non-double-width code path for the high colour modes. 
 
@johnnovak, in capture_video.cpp's `capture_video_add_frame` function, do you know which games/program can get me into the double-width conditional?

```c++
for (auto i = 0; i < src.height; ++i) {
    const uint8_t* row_pointer = row_buffer;
    if (src.double_width) {
        switch (src.pixel_format) {
            ...
            case PixelFormat::RGB555_Packed16:
            case PixelFormat::RGB565_Packed16:
```

(So far I'm only getting the 8-bit and below modes into it)

src/capture/image/image_decoder.h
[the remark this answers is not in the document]
johnnovak: I kinda get the reasoning behind it, but I'd prefer if these were explicit and that's it. So `read_from_16le` and `write_as_16le`, etc. and none of this "host" concept. Then of course anyone messing around with a DOS emulator must 100% know that DOS PCs were a strictly little-endian platform, and job done 😎

include/bgra8888.h
kcgen: Will do.

src/capture/image/image_decoder.h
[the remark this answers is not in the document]
kcgen: Our of scope for this PR, but it's only 200+ touch points in the code for a mass-rename PR. 

But anyone porting features to/from Staging and other DOSBox's will have to be aware of this (most of the big community patches are in, so this risk is mostly moot anyway). 

```
[dosbox-staging] (main $) git grep -c 'host_read\|host_write'
include/mem.h:11
include/mem_host.h:30
include/paging.h:12
src/capture/capture_audio.cpp:4
src/capture/capture_video.cpp:7
src/cpu/core_dyn_x86/decoder.h:2
src/cpu/core_dynrec/decoder_basic.h:6
src/cpu/core_simple.cpp:3
src/cpu/dyn_cache.h:12
src/cpu/paging.cpp:6
src/dos/dos_execute.cpp:2
src/dos/dos_keyboard_layout.cpp:16
src/dos/dos_tables.cpp:1
src/dos/program_boot.cpp:2
src/hardware/gus.cpp:1
src/hardware/ide.cpp:61
src/hardware/mixer.cpp:12
src/hardware/vga_memory.cpp:24
```

include/video.h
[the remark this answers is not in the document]
kcgen: (Removed references to SDL, OpenGL, and ffmpeg equivalent names.)

tests/rgb_tests.cpp
kcgen: This chunk of code is only in the unit tests to see what colour values we get if the memory is copied as a byte-ordered array.  

Basically to check the underlying representation.

But I can avoid the cast by using an array, and will inspect some byte-offsets as another approach.

> I'd prefer if you converted the class to the union / struct trick you used in bgra8888.h

Unfortunately the Rgb888 object needs to be 3-bytes packed, so a union with a `uint32_t` would push the object's size up to 32-bit (and that `int()` conversion cast is used heavily in the render.h routines.. so going to avoid touching this header for now :-).

johnnovak: Cool, but maybe I wasn't entirely clear 😄 `Rgb666` is only used for the palettes, so I guess I'm questioning if need to _care_ about the actual representation at all. In other words, if we always read/write `Rgb666` entries by their component values, then it does not matter one bit what the underlying representation is.

My view is that we should avoid making assumption about the underlying representation and such casting unless we absolutely must.

It makes sense for `Bgra8888` you introduced because the purpose of that is really to serve as an interface/view to the underlying representation. Same for `Rgb555` and `Rgb565` as well that function as a "view" into an `uint16_t` with a specific bit-layout.

I'm just questioning if we need all that for `Rgb666`. Can't we just leave it abstract and make no assumptions on the underlying representation if it's only used for the palette stuff and the individual r6, g6, b6 values get mapped to 8-bit values on an individual component basis anyway (I'm pretty sure that's only how it's used; I'd be surprised if any code treated `Rgb666` values a "array of bytes" in the code).

tests/rgb_tests.cpp
kcgen: I've updated the unit tests to just sanity check byte-order and values (and exluded Rgb666).  

Although our DOS video-card memory is just uint8_t*, the tests confirm that our structs (Rgb888 and Bgrx8888) can be used to go to-or-from memory directly. More importantly, that the underlying representation is endian-agnostic, too.

kcgen:
Ready for review.

Unit tests are passing on big-endian, and it's passing everything I can throw at it, video-wise. Raw captures, rendered captures, and video modes in 4-bit, 8-bit, 15-bit, 16-bit, 24-bit, and 32-bit. 

Here's a walk through vbetest with both image and video captures from every colour depth:

[capture-from-powerpc-freebsd-bsd-bigendian-fixes-1.zip](https://github.com/dosbox-staging/dosbox-staging/files/13539917/capture-from-powerpc-freebsd-bsd-bigendian-fixes-1.zip)

Regression tested on x86-64 and all is passing.

RGB unit tests are are passing on big-endian:

```text
FreeBSD g4 13.2-RC2 FreeBSD 13.2-RC2 releng/13.2-n254580-5a905d8219bb GENERIC powerpc

[==========] Running 17 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 17 tests from rgb
[ RUN      ] rgb.Rgb888_FromRgb444
[       OK ] rgb.Rgb888_FromRgb444 (0 ms)
[ RUN      ] rgb.rgb5_to_8
[       OK ] rgb.rgb5_to_8 (0 ms)
[ RUN      ] rgb.rgb6_to_8
[       OK ] rgb.rgb6_to_8 (0 ms)
[ RUN      ] rgb.rgb8_to_5
[       OK ] rgb.rgb8_to_5 (0 ms)
[ RUN      ] rgb.rgb8_to_6
[       OK ] rgb.rgb8_to_6 (0 ms)
[ RUN      ] rgb.rgb5_to_8_lut
[       OK ] rgb.rgb5_to_8_lut (0 ms)
[ RUN      ] rgb.rgb6_to_8_lut
[       OK ] rgb.rgb6_to_8_lut (0 ms)
[ RUN      ] rgb.srgb_to_linear
[       OK ] rgb.srgb_to_linear (0 ms)
[ RUN      ] rgb.linear_to_srgb
[       OK ] rgb.linear_to_srgb (0 ms)
[ RUN      ] rgb.srgb_linear_roundtrip
[       OK ] rgb.srgb_linear_roundtrip (2 ms)
[ RUN      ] rgb.srgb8_to_linear_lut
[       OK ] rgb.srgb8_to_linear_lut (0 ms)
[ RUN      ] rgb.linear_to_srgb8_lut
[       OK ] rgb.linear_to_srgb8_lut (24 ms)
[ RUN      ] rgb.type_sizes
[       OK ] rgb.type_sizes (0 ms)
[ RUN      ] rgb.rgb565_pixel_components
[       OK ] rgb.rgb565_pixel_components (0 ms)
[ RUN      ] rgb.rgb888_byte_order
[       OK ] rgb.rgb888_byte_order (0 ms)
[ RUN      ] rgb.bgrx8888_byte_order
[       OK ] rgb.bgrx8888_byte_order (0 ms)
[ RUN      ] rgb.bgrx8888_cast_byte_order
[       OK ] rgb.bgrx8888_cast_byte_order (0 ms)
[----------] 17 tests from rgb (35 ms total)

[----------] Global test environment tear-down
[==========] 17 tests from 1 test suite ran. (37 ms total)
[  PASSED  ] 17 tests.
```

src/hardware/vga_draw.cpp
kcgen: I'd rather the Bgra8888 and Rgb565's classes not have other format code it, and vice-versa. We need some common bucket where these multi-format conversion calls can live.

But because this is the only place in the code we convert from Bgra8888 to Rgb565, I figured it was fine.

kcgen: I can eliminate the function and directly move between formats given because both classes can read and write from their r8, g8, and b8 channels (the designs are nicely lining up!)

kcgen:
@peteroupc, thanks for those links to issues about broken drivers. Indeed, I need to launch w/ `SDL_FRAMEBUFFER_ACCELERATION=0` and use `output = texture` and `texture_renderer = software` to prevent the broken OpenGL lib from being queried to avoid crashes.

@casey-ac, thanks for those extremely thorough tests. I've pushed the final review touches and I think we're close to final form.

It's been extremely valuable to catch code-smells  logic regressions that have creeped in since the last go-round back in March 2023.

Using BSD is also really handy because it"falls through" our `#if WIN32 / MACOS / LINUX` pre-processor criteria, so is really handy to catch these pre-processor and build-system breakages.

If you can continue to help build and test `main` periodically, it would be a great help; like an early-warning system :-)

include/video.h
```diff
@@ -171,35 +167,26 @@ enum class PixelFormat : uint8_t {
...
-	// (msb)8X 8R 8G 8B(lsb)
 	// Example:
-	// uint32_t pixel = (unused << 24) | (red << 16) | (green << 8) | (blue << 0)
-	//
-	// SDL Equivalent: SDL_PIXELFORMAT_XRGB8888
-	// FFmpeg Equivalent: AV_PIX_FMT_0RGB32
-	XRGB8888_Packed32 = 32
+	// uint8_t *pixels = image.image_data;
+	// pixels[0] = blue; pixels[1] = green; pixels[2] = red; pixels[3] = empty;
+	BGRX32_ByteArray = 32,
```
johnnovak, line 189: Maybe leave the SDL/FFmpeg equivalents for the rest, just get rid of the SDL equivalent for `BGRX32_ByteArray`? The rest appear to be correct and could be helpful.

src/capture/capture_video.cpp
```diff
@@ -68,13 +68,7 @@ static ZMBV_FORMAT to_zmbv_format(const PixelFormat format)
 	case PixelFormat::Indexed8: return ZMBV_FORMAT::BPP_8;
 	case PixelFormat::RGB555_Packed16: return ZMBV_FORMAT::BPP_15;
 	case PixelFormat::RGB565_Packed16: return ZMBV_FORMAT::BPP_16;
-
-	// ZMBV is "the DOSBox capture format" supported by external tools such
-	// as VLC, MPV, and ffmpeg. Because DOSBox originally didn't have 24-bit
-	// colour, the format itself doesn't support it and treats 4-byte
-	// formats as byte-ordered BGR colour values.
-	//
-	case PixelFormat::BGR24_ByteArray: return ZMBV_FORMAT::BPP_32;
```
johnnovak: So this is no longer true then? Maybe an explanatory comment is in order?

include/bgrx8888.h
```diff
@@ -0,0 +1,81 @@
...
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOSBOX_BGRX8888_H
+#define DOSBOX_BGRX8888_H
+
+#include <cstdint>
+
+// A class that holds the colour channels as an array of four 8-bit values
+// in byte order (alsa known as memory order), regardless of host
```
johnnovak, line 27: also

src/capture/capture_video.cpp
```diff
@@ -358,6 +358,42 @@ static void create_avi_file(const uint16_t width, const uint16_t height,
...
+				std::memcpy(tgt_pixel, src_pixel, src_bpp);
+			}
+		}
+		for (auto j = 0; j < row_multiple; ++j) {
+			const uint8_t* row_pointer = tgt_row;
+			video.codec->CompressLines(1, &row_pointer);
+		}
+	}
+}
+
```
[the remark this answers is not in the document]
kcgen: The function is identical in loop structure and quantities (num rows, num columns, optional doubling on width, optional doubling on height) as the prior one.

It produces the exact same results on little-endian as `main`.

The only difference is how it copies the values:
 - The prior code used pointer casts (buint16*, uint32*)
 - The new one copies byte-by-byte.

It's impossible to get different results in this branch on little endian.
 - All AVIs will be identical
 - All pngs (Alt+F5 and Ctrl+F5) will be identical

The loop is just a simplified and more generic form of the prior.

I'll compare them side-by-side and draw arrows to the elements if it helps :-)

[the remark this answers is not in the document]
kcgen: I'll add a comment in the commit to make it clearer that the loop structure and logic is identical to the existing function, with the only change being the way the pixels are copied from the source (using memcopy byte-by-byte instead of multi-byte pointers).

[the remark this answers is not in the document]
johnnovak: Yeah a comment in the git commit would be good. I admittedly did not spend the time to correlate the removed code with the added code 😄

src/capture/capture_video.cpp
kcgen: :+1: ; added a note in the code.

include/video.h
kcgen: ok! re-added those SDL/FFmpeg equivalents, but not mentioning any SDL/FFmpeg equivalents for `BGRX32_ByteArray`. It's kind of inconsistent, but I have no concerns either way! :sweat_smile:

johnnovak: Well, we're listing equivalents where there are equivalents, and not listing them where there aren't 😄 

The FFmpeg equivalent for `BGRX32_ByteArray` is valid, you might wanna put that back too. The only thing "missing" is the the SDL equivalent for `BGRX32_ByteArray`. You could even say this has no SDL equivalent.

kcgen: added :)

.muon.ini
```diff
@@ -1,4 +1,3 @@
-max_line_len=80
```
johnnovak, line 1: I don't think we need this file?

include/video.h
```diff
@@ -188,18 +187,18 @@ enum class PixelFormat : uint8_t {
...
-	// (msb)8X 8R 8G 8B(lsb)
 	// Example:
-	// uint32_t pixel = (unused << 24) | (red << 16) | (green << 8) | (blue << 0)
+	// uint8_t *pixels = image.image_data;
+	// pixels[0] = blue; pixels[1] = green; pixels[2] = red; pixels[3] = empty;
 	//
-	// SDL Equivalent: SDL_PIXELFORMAT_XRGB8888
-	// FFmpeg Equivalent: AV_PIX_FMT_0RGB32
-	XRGB8888_Packed32 = 32
+	// SDL has has no equivalent.
```
johnnovak, line 199: 🥇

kcgen:
Away it goes! :ship:

kcgen:
Just noticed that NetBSD 10 is RC2 is out, and it mentioned that 10.x supports booting the Raspberry Pi 0-3 in big-endian mode:

> **Hardware support**
> Added support for booting the Raspberry Pi 0-3 in big endian mode. 

https://www.netbsd.org/releases/formal-10/NetBSD-10.0.html

And I think these are the most relevant notes: https://wiki.netbsd.org/ports/evbarm/raspberry_pi/#index3h2

This sure would be more convenient to exercise big-endian on a fast and accelerated platform! @weirddan455 , I think you were curious about such a beast as well.  

If anyone pokes around with this, please share progress and any tips here or in the discussion items! :beers:

PR #3159, opened 2023-11-29
Fix single-precision FPU rounding

kklobe:
# Description

Quake does many floating point operations in single precision mode. This was causing issues in the `FROUND` FPU emulation function, since the `double in` parameter could easily exceed the precision of a `float`. After logging this function, I saw Quake often pass values like 5.9999999967 to be chopped in single precision mode, so was getting a value of 5.0 back. By casting to a `float` prior to the `trunc`, these cases get automatically handled correctly.

On my M1 MacBook Pro system, at least, this has the nice side effect of boosting QTD by 5-7%.

# Manual testing

Tested Quake extensively. I also tried a handful of other games/demos, but couldn't find anything else that flipped over to single precision mode.

MCPDIAG results have not changed.

# Checklist

I have:

### DIFF
--- a/src/fpu/fpu_instructions.h
+++ b/src/fpu/fpu_instructions.h
@@ -20,6 +20,8 @@
 #include "fpu.h"
 #endif
 
+#include "math_utils.h"
+
 static constexpr uint16_t ExtendedPrecisionModeMask = 0x0300;
 
 static void FPU_FINIT(void) {
@@ -100,8 +102,9 @@ static void FPU_FPOP(void){
 	return;
 }
 
-static double FROUND(double in){
-	switch(fpu.round){
+static double FROUND(double in)
+{
+	switch (fpu.round) {
 	case ROUND_Nearest: return std::nearbyint(in);
 	case ROUND_Down: return std::floor(in);
 	case ROUND_Up: return std::ceil(in);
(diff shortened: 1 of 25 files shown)


src/fpu/fpu_instructions.h
```diff
@@ -107,16 +111,24 @@ static double FROUND(double in){
 	case ROUND_Up: return std::ceil(in);
 	case ROUND_Chop: 
 	{
+		switch (fpu.cw & PrecisionModeMask) {
+		// This case properly handles "in" values beyond the precision
+		// of a float, e.g. 5.99999999367, by correctly rounding to 6.0
+		// via the cast, then performing the truncation,
+		// instead of returning 5.0. Quake, for example, does many
+		// calculations in single precision mode.
```
johnnovak, line 119: Minor: reformat comment block

johnnovak (approved):
Looks alright @kklobe, and it's hard to reason about this code too much, so I'd say let's merge it and the proof is in the pudding 😎

src/fpu/fpu_instructions.h
kklobe: ah, ok, had to put it back to a single-line comment and let the script do it over again.

kklobe:
> Looks alright @kklobe, and it's hard to reason about this code too much, so I'd say let's merge it and the proof is in the pudding 😎
> 
> If Quake is fine, then it's a low risk change as you need to find FPU using games/demos with a magnifying glass... (and they're only from the late-90s Pentium era)

Yeah, I tried to keep it as simple as possible - the main idea is, if the FPU is in single precision mode, then cast the value to be truncated to single precision first, to mimic what the FPU would be doing, and then do the chop.

johnnovak:
We'll wait a day of two, maybe someone else wants to comment too, then merge time @kklobe .

kklobe:
setting this back to DRAFT since I'm rebasing off @johnnovak's aspect PR

Issue #3169, opened 2023-12-02
100 USD bounty for fixing icons on Linux due to the deployment of PR #3157


Grounded0:
Icons are now a mishmash of legacy and new style icons on Linux due to deployment of PR #3157. Deliver a PR fixing this in a deployable state and collect 100 USD bounty from me. I can pay via SWIFT, cryptos available in Coinbase, or PayPal.

#3157 has all the technical aspects already detailed on what needs to be fixed and how it can be fixed. We just need a person with motivation to fix Linux desktop user experience now for everyone who uses Linux like me (though i only use it over SSH or Mosh).

Connected issues and PRs:

- #3157

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

weirddan455:
I just compiled latest main with the icon changes merged.  However, I can only see the small icon in the bottom taskbar and the even smaller icon at the top-left of the window.  I see your comment here:  https://github.com/dosbox-staging/dosbox-staging/pull/3157#issuecomment-1837000072

However, I don't know how you're seeing all those different icon sizes from inside KDE.  My alt+tab doesn't look like your "large icon switcher".  It opens a bar on the left side of the screen and shows an icon the same size as in the task bar.

Grounded0:
Cuttlefish is KDE's icon theme viewer and comes with plasma-sdk package.

Grounded0:
Enabling KDE Large Icon Application Switcher for testing purposes:

https://bbs.archlinux.org/viewtopic.php?id=218308

weirddan455:
Did some testing on this.  The .desktop file that we have in `contrib/linux/dosbox-staging.desktop` is good with regards to icons and matches what other applications are doing.  It's one single line in there `Icon=dosbox-staging` that tells it what icon to use.

The icon files themselves need to be placed in a specific location for the desktop environment to find them.  There are (at least) two directories that should always get searched: `~/.local/share/icons` (per user directory) and `/usr/share/applications/icons` (system-wide directory).  Inside that directory, they should be placed in `hicolor/256X256/apps/dosbox-staging.png`.  Icons of multiple resolutions can be placed in the appropriate folder (replace 256X256 with the real resolution of the icon).  One SVG icon can go in `hicolor/scalable/apps/dosbox-staging.svg`

>  Question is what we're going to do with [icon.c](https://github.com/dosbox-staging/dosbox-staging/tree/main/src/gui/icon.c).

[...]

PR #3171, opened 2023-12-03
Do not set icon at runtime

weirddan455:
# Description
Removes the embedded icon in icon.c.  Also removes the `SetIcon()` function that was decoding this embedded icon and programmatically setting it at runtime with `SDL_SetWindowIcon()`.  As mentioned in #3169 this is no longer the "proper" way to set an icon on any of the 3 major operating systems.  Windows prefers to have an icon embedded in its .exe by the build system which we are already doing.  On Linux, this method does not work in Wayland systems.  MacOS also does something else apparently (this code was already compiled out for Mac and MSVC).

Note for us Linux devs, we might fall back to a generic X.org icon if we're running from the build directory and don't have the .desktop file and icon files in the proper locations.  I'm considering writing a script to install these.  Once the desktop file is in place, the icon gets cached and shows up even if you run Dosbox from the terminal in a build directory.

The script could maybe also be used for end users.  Currently for Linux, we just ship a .tar.xz file.  Unzip and run.  A short script would maybe be helpful that basically just copies a desktop/start menu shortcut plus icon files.

End users that get Staging from Flatpack or from their distro's package manager would likely be unaffected.  Those packages should already be putting the .desktop and icon files in the correct place and nothing needs to be done by us.

## Related issues
#3169

# Manual testing
Copied the .desktop file to ~/.local/share/applications and the icon files to ~/.local/share/icons and confirmed the icon shows up when this is done.

# Checklist

I have:


johnnovak (approved):
Good riddance! 🗑️ 🧹 😎

johnnovak:
> > @weirddan455 Does the install target of the icons Makefile do a similar thing? Would it make sense to extend that Makefile instead of writing a script?
> 
> Yeah, that Makefile does pretty much what I was thinking about doing. I don't know how our end users are meant to run this though. Currently all they do is unzip a tarball if they download it from our website. I don't even see this Makefile in the the 0.80.1 release so it may get stripped out.

It's used here when creating the Linux package in our build process:

https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/create-package.sh#L142

I don't think it's meant to be packaged into the final package.

kcgen (approved):
:+1: ; looks good, @weirddan455 !

Grounded0:
Does deploying this close?

- #3169

weirddan455:
Probably not @Grounded0 but you haven't fully articulated what problem you are having exactly on that ticket.  The only bug this PR is likely to solve is the old icon appearing at the top left of the window instead of the new icons that @johnnovak made.

As far as I can tell, the only thing remaining to do for Linux icons is test the packaging and maybe write a script for users that download the tarball off the website.

This AUR package seems to depend on a `meson install` command so I'll test that later and make sure the new icons are getting installed properly:  https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=dosbox-staging

PR #3172, opened 2023-12-03
Adaptive CRT shader compatibility fixes for really old GPUs

johnnovak:
# Description

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3114

@Burrito78 kindly gave me remote desktop access to his problematic old MacMini with an Intel HD 4000 iGPU that displays only a black screen when using any of the new adaptive CRT shaders. I was able to pinpoint the problem by the time-tested method of trial and error, also known as "just dicking around" in some circles 😎 

The problem seems to be that there's just too much branching going on in one of the functions. GLSL is still a fairly high-level language at the end of the day, and the shader compilers try to do their best to optimise per-pixel branching away. This branching code is not a problem at all on more recent GPUs (7-8 years old), but on this ancient 10+ year old iGPU, the shader compiles without errors, but somehow, who knows how and why, it just displays a black screen at runtime... Commenting out some of the branches fixes it, so the solution I adopted is to comment every branch of the function out that is not used by the adaptive shaders. This is an attempt to pre-empt similar problems on other weird old GPUs.

For the record, the problem could not be reproduced on my $200 bottom-of-the-barrel mini laptop I bought in 2015 which also has an Intel iGPU, but a more recent model. So I have reason to believe the issue could only manifest on ancient 10+ year old GPUs...

## Disclaimer

Yes, I know that it's **_absolutely_** **_horrible_** that I have 22 copies of the _exact same_ (!) shader code, just with different settings... 🥶 But we have no preset support, so this was the only practical way to release my set of shaders this year (and to do this before eXoDOS v6 is out, so our shader work can be included) using our current shader pipeline which is very barebones.

I have plans for how to make this a lot nicer in the future; most likely these 22 copies will be replaced by a single shader, and I'll just send it different settings at runtime when there's a "shader change" (which is really just a "preset change"). But that's for the future... 😄 

# Manual testing

Tested a few fixed shaders on Burrito's box to make sure I understood the problem.

Regression tested all fixed shaders on my M1 MacBook to make sure they still work.

# Checklist

I have:

### DIFF
--- a/contrib/resources/glshaders/crt/arcade-1080p.glsl
+++ b/contrib/resources/glshaders/crt/arcade-1080p.glsl
@@ -169,10 +169,10 @@ uniform PRECISION float BLUE_SHIFT;
 
 #define WP_ADJUST           0.00
 #define TEMPERATURE      9300.00
-#define LUMA_PRESERVE 		1.00
-#define RED_SHIFT 			0.00
-#define GREEN_SHIFT 		0.00
-#define BLUE_SHIFT 			0.00
+#define LUMA_PRESERVE       1.00
+#define RED_SHIFT           0.00
+#define GREEN_SHIFT         0.00
+#define BLUE_SHIFT          0.00
 #endif
 // END PARAMETERS //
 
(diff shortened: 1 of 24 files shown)


Burrito78 (approved):
Works in windowed and fullscreen with various resolutions. Thanks!

Burrito78:
> Check at least a few popular shaders for performance regressions.
I checked QTD with crt-auto, sharp and none using default cfg and got got around 137-140fps which is the highest result you're going to get on that machine. So I think we are fine.

FeralChild64:
Tested on my Intel ARC, no problems found.

kcgen:
Tested CGA, Tandy, EGA, VGA (various), and VESA modes on the Pi4 (Broadcom OpenGL 4.x compatible) and found no problems.

kcgen (approved):

PR #3174, opened 2023-12-04
Let Meson use a system call to detect the page size

### LINKED ISSUE #3160: Crashing on raspberry pi 5 when auto switching to max 100% CPU cycles

_No response_

Linux ARM (Raspberry Pi)

_No response_

### Relevant hardware info

Raspberry PI 5

Dosbox-staging runs normally, but when a dos application is executed that leads to auto switching to max 100% cpu cycles (e.g. Windows 3.1) it crashes.

Explain how to reproduce
1. Launch a DOS program that requests max 100% cycles, e.g. Windows 3.1
2.
3.

_No response_

### Your configuration

```ini
https://gist.github.com/ekondis/42cdc9aaed40f91c3cd6572524d15054
```

### Provide a Log

```text
dosbox-staging logs:
https://gist.github.com/ekondis/c76d876ca8ccf0726760553c80afa796

gdb stack trace:
https://gist.github.com/ekondis/b18581a0217ec9dff6d3e6f8004c3b3f
```

- [X] Yes, I agree.

kcgen:
# Description
    
Previously we ran local binaries to get the page size, however some systems don't have them (such as RetroPie, which is missing the `pagesize` binary).
    
The good news is both Microsoft and POSIX both have long-lived extremely standardized calls to get the page size:
    
- https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsysteminfo
    Supported on Windows 2000 and greater.
    
- https://pubs.opengroup.org/onlinepubs/9699919799/
    Supported on POSIX systems since 2001
    
By using a C syscall, Meson will build and run it using the cross-compiler's runnable environment (if one is configured).
   
This is particularly important when cross compiling, as we want the destination host's page size and not the local build machine's page size.

(If a cross-compiler isn't configured Meson just runs it on the local build machine)

Finally, this PR also lets users (optionally) configure the page size via Meson's setup options, such as: `meson setup -Dpagesize=16384 ...`

Some machine support multiple page sizes - so there's no harm in letting them override it or experiment.

## Related issues

Fixes #3160.

# Manual testing

Tested on x86-64 Linux, the Pi4 w/ Linux, and Apple's M1.

# Checklist

I have:

### DIFF
--- a/contrib/pagesize/posix_pagesize.c
+++ b/contrib/pagesize/posix_pagesize.c
@@ -1,0 +1,51 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+// Get the memory page size on POSIX systems
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+//
+// Ref: https://pubs.opengroup.org/onlinepubs/9699919799/
+//
+// Usage: ./posix_pagesize
+//
+// On success, prints the page size in bytes and return 0 to the shell.
+// On failure, prints an error messages and return 1 to the shell.
+//
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+int main(void) {
+	// Try two queries
+	long pagesize = sysconf(_SC_PAGESIZE);
+	if (pagesize == -1) {
+		pagesize = sysconf(_SC_PAGE_SIZE);
+	}
+
+	// Print the results
+	if (pagesize > 0) {
+		printf("%ld\n", pagesize);
+	} else {
+		printf("Error getting page size: %s\n", strerror(errno));
+	}
+
+	return (pagesize > 0) ? 0 : 1;
+}
--- a/contrib/pagesize/windows_pagesize.c
+++ b/contrib/pagesize/windows_pagesize.c
@@ -1,0 +1,62 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2023  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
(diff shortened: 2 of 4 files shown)


weirddan455 (approved):
The code all looks good and this certainly looks like a win for native compilation.  I don't understand how this would work for cross compilation though.  If I'm compiling for ARM on my x86 machine, I compile that small program but then I can't execute the resulting binary unless I have qemu or something installed (which is not typically part of a cross compile environment).

The old method probably isn't any better though (running some command may just return host-side value instead of target value).  Plus you've introduced an option to manually set page size so someone can always use that if they're compiling for some esoteric architecture.

shermp:
As an addition to what @weirddan455 mentioned, if it is a syscall, isn't it also getting the info from the running kernel? Which might be the host kernel even with qemu-user?

kklobe:
I like this idea. How about we just go all the way and modify the code to compute and use this at runtime instead of compile-time? I think the need for this is largely (totally?) confined to `dyn_cache.h`, and we could just run your nifty utility functions as part of `cache_init()`.

weirddan455:
> I like this idea. How about we just go all the way and modify the code to compute and use this at runtime instead of compile-time? I think the need for this is largely (totally?) confined to `dyn_cache.h`, and we could just run your nifty utility functions as part of `cache_init()`, and optionally add a config parameter to let it be overridden.

Yeah, maybe this makes sense to be a runtime value.  It's probably out of scope for what @kcgen was planning here though (unless he feels like doing it).  IIRC there's some compile time `#ifdef`'s relating to page size that would need to be refactored.  And then we would need to test for bugs and performance regressions since this is in the core of the emulator.

kcgen:
> If I'm compiling for ARM on my x86 machine, I compile that small program but then I can't execute the resulting binary unless I have qemu or something

This is where Meson leaves the other build systems in dust; it has first-class handling for cross compiling, including three-way cross compiling (where you cross-compile your cross-compiler as an intermediary).

When setting it up, you populate a **_cross build definition file_**, and one of the fields is an `exe_wrapper` that's able to run cross-compiled output binaries.

So that's where you'd plug in qemu / wine / etc.. or even a script that does something like `ssh actual-machine $1`, and you get stdout, stderr, and even the rcode back from the SSH client.

Great walk through here: https://mesonbuild.com/Cross-compilation.html

kklobe:
> > I like this idea. How about we just go all the way and modify the code to compute and use this at runtime instead of compile-time? I think the need for this is largely (totally?) confined to `dyn_cache.h`, and we could just run your nifty utility functions as part of `cache_init()`, and optionally add a config parameter to let it be overridden.
> 
> Yeah, maybe this makes sense to be a runtime value. It's probably out of scope for what @kcgen was planning here though (unless he feels like doing it). IIRC there's some compile time `#ifdef`'s relating to page size that would need to be refactored. And then we would need to test for bugs and performance regressions since this is in the core of the emulator.

Certainly up to @kcgen. I've already got a version working, it's a very small change, and can't see any place else in the codebase that uses `PAGESIZE` or `host_pagesize`, but understandably makes me a bit nervous for 0.81 at this stage.

kcgen:
@kklobe: maybe we can merge this, and then follow up w/ yours? We've got plenty of time to hit regressions and compare performance (and it would be a small revert in the worst cast).

kcgen:
> Sounds good; preliminary QTD and PERFDOS benchmarks show no change (I don't see any array sizing using that value, only the DOS page-size values, which of course wouldn't be affected by this).

> I'll throw it in a PR and let the team decide!

Right on, @kklobe!

johnnovak:
@kklobe @kcgen Um, seems like massive overkill? Has anyone proven not just using 4k has any measurable benefit?

I think assuming 64 byte cache lines and 4k page sizes gets you 99% there on modern hardware, so why complicate life? 😄 

At least, I'd measure first, then if the gains are significant, then complicate.

kcgen:
Page size needs to be correct for the dynamic cores. 
(See my expanded comment here: https://github.com/dosbox-staging/dosbox-staging/issues/3160#issuecomment-1839584719)

We definitely want runtime detection to handle aarch64 processors that support setting either 4k or 16k page size at boot time.

kklobe:
> @kklobe @kcgen Um, seems like massive overkill?

Probably!

Now go away and leave us alone :smile:

kcgen:
Also, hardware and operating systems that support write-or-execute flagging ([W^X](https://en.wikipedia.org/wiki/W%5EX)) require flagging of whole pages.  That's probably what crashed the Pi5: we asked to flag only a 4k portion of a 16K page (and were denied).  So dynrec didn't even get off the ground to fill its first page.

PR #3180, opened 2023-12-06
Cleanup a handful of unchecked return-code issues flagged by Coverity

kcgen:
# Description

Fix a handful of unchecked return-code issues flagged by the updated Coverity scanner.

## Related issues

# Manual testing

Tested shell command line handling and booting floppy and HDD images.

Also tested with UBSAN and ASAN builds.

# Checklist

I have:

### DIFF
--- a/include/fs_utils.h
+++ b/include/fs_utils.h
@@ -89,6 +89,39 @@ constexpr uint32_t OK_IF_EXISTS = 0x1;
 
 int create_dir(const std_fs::path& path, uint32_t mode, uint32_t flags = 0x0) noexcept;
 
+// Behaves like fseek, but logs an error stating the module, byte offset, file
+// description, filename, and strerror on failure. Returns true on success and
+// false on failure. On failure, it closes the file as it's no longer in a good
+// state.
+//
+bool check_fseek(const char* module_name, const char* file_description,
+                 const char* filename, FILE*& stream, const long long offset,
+                 const int whence);
+
+// Returns a 'check_fseek' function object that behaves like the above. This can
+// be used when lots of sequential seeks are needed.
+//
+inline auto make_check_fseek_func(const std::string& module_name,
+                                  const std::string& file_description,
+                                  const std_fs::path& filepath)
+{
+	// Use the lambda copy-operator to keep copies of the arguments inside
+	// the lambda, as these arguments would normally go out of scope with
+	// respect to the lifetime of the lamda.
+	//
+	auto check_fseek_lambda = [=](FILE*& stream,
+	                              const long long offset,
+	                              const int whence) -> bool {
+		return check_fseek(module_name.c_str(),
+		                   file_description.c_str(),
+		                   filepath.string().c_str(),
+		                   stream,
+		                   offset,
+		                   whence);
+	};
+	return check_fseek_lambda;
+}
+
 // Convert a filesystem time to a raw time_t value
 std::time_t to_time_t(const std_fs::file_time_type &fs_time);
 
--- a/src/misc/fs_utils.cpp
+++ b/src/misc/fs_utils.cpp
@@ -22,7 +22,9 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cerrno>
 #include <chrono>
+#include <cstring>
 #include <fstream>
 
 #include "checks.h"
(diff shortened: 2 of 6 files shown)


src/misc/fs_utils.cpp
```diff
@@ -22,16 +22,39 @@
...
+	}
+
+	assert(module_name);
+	assert(filename);
+	LOG_ERR("%s: Failed seeking to byte %lld in file '%s': %s",
+	        module_name,
+	        offset,
+	        filename,
+	        strerror(errno));
+	fclose(stream);
```
johnnovak, line 53: Isn't this information only useful for devs? Seeks offset, error codes, etc. Would it be better to turn this into an ifdef'd out debug level message, and in the else branch of the ifdef we'd just log an error saying "Error reading file..." or something?

src/misc/fs_utils.cpp
```diff
@@ -22,16 +22,39 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cerrno>
 #include <chrono>
+#include <cstring>
 #include <fstream>
 
+// #include "cross.h"
```
johnnovak, line 30: Remove?

include/fs_utils.h
```diff
@@ -89,6 +89,34 @@ constexpr uint32_t OK_IF_EXISTS = 0x1;
...
+// failure. On failure, it close the file as it's no longer in a good state.
+//
+bool check_fseek(const char* module_name, const char* filename, FILE* stream,
+                 const long long offset, const int whence);
+
+// Returns a 'check_fseek' function object that behaves like the above, with
+// the same signature as fseek. This can be used in areas where lots of
+// sequential seeks are needed.
+//
+inline auto create_check_fseek(std::string module_name, std::string filename)
```
johnnovak, line 103: Suggest renaming to `make_check_fseek_func`. The word "create" gave me a pause as it has a common meaning in the context of file operations (_"Does this create a file then does some seeking? That's one weird function..._ 🤔 _"_  😄 )

include/fs_utils.h
```diff
@@ -89,6 +89,34 @@ constexpr uint32_t OK_IF_EXISTS = 0x1;
 
 int create_dir(const std_fs::path& path, uint32_t mode, uint32_t flags = 0x0) noexcept;
 
+// Behaves like fseek, but logs an error stating the module, byte offset,
+// filename, and strerror on failure. Returns true on success and false on
+// failure. On failure, it close the file as it's no longer in a good state.
```
johnnovak, line 94: closes

src/misc/fs_utils.cpp
weirddan455: `strerror` turns the C error code into something human readable like "permissions error" or "file not found".  As long as this isn't flooding the logs, more info is usually good IMO.  The user doesn't have to know what it means but it helps us when they provide detailed logs on a tricky bug we're trying to fix.

kcgen: Yup; what @weirddan455 said.

These seeks are coming from user-provided files, and they'll be able to see the path and issue, so definitely worth while. 

We tend to try to pass on strerror whenever C functions can fail on user data (so this is just keeping with that trend).

include/fs_utils.h
kcgen: Yup, that's a nicer name 👍

src/misc/fs_utils.cpp
johnnovak: Yeah okay, sounds reasonable then. I thought we're logging error codes or something... Less convinced having seek offsets in the logs is a good idea, though, but whatever 😄

include/fs_utils.h
```diff
@@ -89,6 +89,34 @@ constexpr uint32_t OK_IF_EXISTS = 0x1;
...
+
+		return check_fseek(local_module_name.c_str(),
+		                   local_filename.c_str(),
+		                   stream,
+		                   offset,
+		                   whence);
+	};
+	return check_fseek_lambda;
+}
+
```
FeralChild64, line 119: It would be nice if you could use this mechanism to replace `fseek` and `log_layout_read_error` calls in `dos_keyboard_layout.cpp`.

kcgen: @FeralChild64 , see the last commit.

It's not as tidy as I would have liked because I had to add a bit more pass-through of the filename (to log it) as well as moving to the simpler 'FILE*' pointers.

If it's not to your liking I can drop that commit (no problems!).

kcgen: Edit.. decided to knock it out; it was too messy and introduced more file management headaches because it operates on a `FILE*` instead of the `FILE_unique_ptr`.

kcgen:
(Might as well get this merged)

kcgen:
Confirmed!

PR #3183, opened 2023-12-07
Use zlib-ng, a faster drop-in replacement for zlib, when found

kcgen:
# Description

zlib-ng makes use of host SIMD calls and so can speed deflate times by about 50% or better based on my timings.

## Related issues

This is an optional mitigation for the slower ZMBV performance @weirddan455 investigated due to the move to 32-bit pixels #2812.

# Manual testing

Try it:
```shell
git clone --depth 1 https://github.com/zlib-ng/zlib-ng
cd zlib-ng
cmake -DZLIB_COMPAT=1 -DWITH_NATIVE_INSTRUCTIONS=1 .
cmake --build . --config Release
ctest --verbose -C Release && sudo cmake --build . --target install
```

Run a `meson setup ...`, and look for `zlib found: YES 1.3.0.zlib-ng`:

Tested and measured image capture and video frame compression times. The results are roughly a 50% (or better) speed up.

# Checklist

I have:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -528,6 +528,7 @@ sdl2_dep = dependency(
 
 zlib_dep = dependency(
     'zlib',
+    fallback: ['zlib-ng', 'zlib'],
     version: ['>= 1.2.11', '< 2'],
     default_options: default_wrap_options,
     static: ('zlib' in static_libs_list or prefers_static_libs),


johnnovak:
> Tested and measured image capture and video frame compression times. The results are roughly a 50% (or better) speed up.

Noice 😎

weirddan455:
Wow, this library completely solved the performance issues I was having.  No stutters at all when I use this.  I had to set a compile option to generate something that's ABI compatible with regular zlib `cmake -DZLIB_COMPAT=ON ..` then run `LD_LIBRARY_PATH=/path/to/zlib-ng dosbox` and it works without even having to recompile.

I wasn't thinking zlib would be the bottleneck here.  zlib sees such widespread use I figured it would be optimized already.  Guess not.

Would be nice to ship with this but yeah build systems are pain.

kcgen:
Thanks for testing it, @weirddan455 !

I added timer calls around the various chucks of ZMBV and isolated the big time hit to the last `deflate(...)` call:

```
	// Only tally bytes if the stream was OK
	if (deflate(&zstream, Z_SYNC_FLUSH) >= Z_OK) {
		bytes_processed = static_cast<int>(compress.writeDone + zstream.total_out);
	}
```

During large flushes, zlib on my system can waste up to 100 ms but -ng slashes that into a quarter. 

For small flushes, zlib will be some whole fraction of a millisecond (say, 0.3 ms… which itself is a non-trivial chunk of DOSBox's 1 ms tick), however -ng can often collapse by 1/10th! Down to 0.03 ms.

Amazing.  I see -ng is using SIMD calls on all the hardware, too (Not only Intel and ARM, but also AltiVec and others processor models).

PR #3186, opened 2023-12-07
Minor country and help fixes

FeralChild64:
# Description

1. Fix two comments in country data definitions (data is OK; it is the comments which are wrong) - noticed by @rderooy 
2. Fix problem I have discovered while updating Polish translation - to determine UTF-8 string length in characters we need something much more sophisticated than `strlen` or C++ counterparts (they give us length in bytes, and a single Unicode character might need many bytes to be encoded); it is much easier to rely on the translator to put a proper number of minuses (as a title underscore). 

Screenshot-Bug

# Manual testing

Execute `dosbox --list-countries` and `dosbox --list-glshaders` commands

# Checklist

I have:

### DIFF
--- a/contrib/resources/translations/nl.lng
+++ b/contrib/resources/translations/nl.lng
@@ -417,6 +417,7 @@ Oezbekistan
 .
 :DOSBOX_HELP_LIST_GLSHADERS_1
 Lijst met beschikbare GLSL-shaders
+----------------------------------
 .
 :DOSBOX_HELP_LIST_GLSHADERS_2
 Bovenstaande shaders kunnen precies worden gebruikt zoals vermeld in de
--- a/contrib/resources/translations/pl.lng
+++ b/contrib/resources/translations/pl.lng
@@ -67,6 +67,7 @@ Lista dostępnych opcji:
 .
 :DOSBOX_HELP_LIST_COUNTRIES_1
 Kody krajów, często takie same jak telefoniczny numer kierunkowy
+----------------------------------------------------------------
 .
 :DOSBOX_HELP_LIST_COUNTRIES_2
 Powyższych kodów numerycznych należy użyć w ustawieniu 'country'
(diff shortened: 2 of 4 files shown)


johnnovak:
> to determine UTF-8 string length in characters we need something much more sophisticated than strlen or C++ counterparts

Oh man... don't tell me this is still an unsolved problem in the C++ std lib... Like every lowly scripting language had this for 20 years now...😞

FeralChild64:
> Oh man... don't tell me this is still an unsolved problem in the C++ std lib...

Unfortunately, it is.
- There is `std::wstring`, but it is not portable, on Windows it is 16-bit only. AFAIK not much standard library functions can deal with it.
- We have `std::u32string`, which can be used to deal with UTF-32 encoding, not really useful for us. Moreover, `size()` will tell us the number of code points, while we need the number of graphemes here (and a grapheme can be made, for example, from one code point telling about a base Latin letter + two other telling about combining marks; IIRC Baltic languages are particularly nasty here).
- C++20 has `std::u8string` to deal with UTF-8. But the standard library offers nothing to handle it. Might be useful to us once we migrate to C++20, but only for type checking.

I could implement UTF-8 compliant `strlen`, by asking my Unicode engine to convert UTF-8 to ASCII and then run a regular string size check - but this is too risky, I might improve my Unicode engine in the future to cooperate with emulated DOS more closely, and at this point we don't have DOSBox fully initialized - I'm not sure if I'll remember to test this particular use case like 2 or 3 years in the future.

IMHO better ask the translator to do the job.

johnnovak:
> Moreover, size() will tell us the number of code points, while we need the number of graphemes here (and a grapheme can be made, for example, from one code point telling about a base Latin letter + two other telling about combining marks; IIRC Baltic languages are particularly nasty here).

Actually, I take my statement back that "every other stdlib" can do this 😅 I'm pretty sure most other languages offer the same basic support only, i.e., they only count "code points", or "runes" (they're called runes in Nim, for instance). But not that wacky situation you described.

Also some libs just shit themselves when dealing with 4-byte UTF-8 emojis which are super popular among Asian audiences (talking from real world experience here; 4-byte UTF-8 emojis caused so many headaches for me in the past...)

kcgen (approved):
Tha is for the thorough discussion and explanation, @FeralChild64 !

Makes sense.

PR #3188, opened 2023-12-07
Use zlib-ng in the Linux, macOS, and MSYS2 release builds 

kcgen:
# Description

Builds and links zlib-ng statically into our release binaries for faster AVI recording and image capturing.

Ideally we would have a Meson's wrap for zlib-ng, but one doesn't exist. 

As it's also a fair bit of work to create one, I opted for a simple Makefile in `contrib/` that CI invokes before setting up the release builds. 

CI also fully caches it, so it's only built once per platform (until we update its version).

As for the CI workflow files, this is a minimal-touch approach so we can easily discard it when we move to MP4 recording.

## Related issues

This fixes slow recordings for users of the release package.

@shermp : I noticed VCPKG does have zlib-ng, however I wasn't sure how to update our NuGet json list (which I think is held server-side somewhere?).

# Manual testing

Tested capturing video and images on Linux and macOS.

# Checklist

I have:


shermp:
This seems a bit... hacky to me? A more elegant solution would be to use a cmake wrap, but that's a very new meson feature.

With regards to VCPKG, if we require API compatibility with zlib, we would need to use an overlay port for it to work, which seems to be in a too-hard basket for something temporary.

johnnovak (approved):
Very nice improvement @kcgen, thanks for doing the build system changes so we don't need to 😎 👍🏻 And it's awesome you found this library in the first place.

You guys know I'm a fan of static linking, so I have zero problems with this approach. Solid and does the job. I don't even mind "amalgamating" library sources directly into our source tree (especially for header-only libraries), but yeah, downloading is maybe slightly better (as long as the links keep working 😅).

Stellar work! 🎉 🥇 🍻 🥳

johnnovak:
> This seems a bit... hacky to me? A more elegant solution would be to use a cmake wrap, but that's a very new meson feature.

@shermp  By all means work together with @kcgen on this if you think this can be done in a better / cleaner way. But this solves a very real problem (it restores video capturing performance, maybe even makes it faster), so telling people we opted _not_ to do this because it's not elegant enough is a hard sell 😄 But sure, it you build system gurus can improve the solution later, go for it 😎 

I'm not sure if it's temporary, btw; zlib is used by other things too, and why would we _not_ want to take the performance improvements offered by zlib-ng if we can.

kcgen:
@shermp -- it's absolutely a hack!🙉

I didn't realize Meson can officially wrap a cmake project and offer it as a dep.

Do you have some doco links for this or examples in the wild?

What I want to do is fully replace the traditional zlib dep discovery with the zlib-ng, without any wiggle room to use an old system zlib.a or zlib.so. 

So any docs on how I can connection zlib_dep to a cmake project would be perfect!

shermp:
> @shermp -- it's absolutely a hack!🙉
> 
> I didn't realize Meson can officially wrap a cmake project and offer it as a dep. Thanks for mentioning this!
> 
> Do you have some doco links or examples in the wild?
> 
> What I want to do is fully replace the traditional `zlib_dep` discovery with zlib-ng, without any wiggle room to use an old system zlib.a or zlib.so.
> 
> So any docs on how I can connection `zlib_dep` to a cmake project would be perfect!

https://mesonbuild.com/Wrap-dependency-system-manual.html#cmake-wraps

I believe it's very new though, and may only be available since the latest version of meson.

Another option is to include `zlib-ng` as a cmake subproject, bypassing the wrap process. That has been supported for years.

kcgen:
Will check those out; thanks!

shermp:
If you do go the cmake route though, it might be advisable to hide support behind a disabled by default option, so that cmake doesn't become mandatory for users.

shermp:
Still not sure how to handle VCPKG though...

johnnovak:
Can't we merge this as-is @kcgen @shermp and refine later? Just the deliver the enhancement to the users, then the rest is internal housekeeping.

kcgen:
> Can't we merge this as-is @kcgen @shermp and refine later? Just to deliver the enhancement to the users, then the rest is internal housekeeping.

Unfortunately this solution isn't perfect because our sub packages depend on zlib too, and they pull in zlib via the wrap (as we requests), which trumps zlib-ng, which is being provided as a system lib in this PR.

Will close this down and try for the cmake wrap approach like @shermp suggested.

PR #3082, opened 2023-11-02
Add a memory read breakpoint feature to the builtin debugger

LowLevelMahn:
# Description

adds a C_HEAVY_DEBUG feature for breaking on memory reads - memory write breakpoints were already available in dosbox svn long time

* will break on read of the memory using F5-run or just show accessing code point on F10 single-step
* only working so far for non-protected realmode, and as most of the debug features only reliable when using core=normal (or as i was told on vogons years ago)

bpmr

use in debugger with: BPMR seg:ofs

# Manual testing

tested the feature with a simple nasm DOS COM executable aka clean environment

assemble with recent nasm using `nasm -f bin bpmr.asm -o bpmr.com`

```
; nasm -f bin bpmr.asm -o bpmr.com
org 100h

mov si,test1
lodsb
lodsb
lodsb

mov byte [test2],'z'

mov al,byte [test1]

mov cx,3
mov si,test1
rep lodsb

mov si,test1
mov di,test2
movsb
movsb
movsb

mov cx,3 
mov si,test1
mov di,test2
rep movsb

mov ax,4C00h
int 21h

test1 db 'ABC'
test2 db 'abc'
```

# Checklist

I have:
 
* [x]  followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
* [x]  performed a self-review of my code.
* [ ]  commented on the particularly hard-to-understand areas of my code.
* [x]  split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
* [x]  applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
* [x]  [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
* [ ]  confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
* [ ]  added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
* [ ]  made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
[...]

PR #3191, opened 2023-12-09
Add a memory read breakpoint feature to the builtin debugger

kcgen:
# Description

(PR body below is from @LowLevelMahn in #3082)

adds a C_HEAVY_DEBUG feature for breaking on memory reads - memory write breakpoints were already available in dosbox svn long time

* will break on read of the memory using F5-run or just show accessing code point on F10 single-step
* only working so far for non-protected realmode, and as most of the debug features only reliable when using core=normal (or as i was told on vogons years ago)

bpmr

use in debugger with: BPMR seg:ofs

# Manual testing

tested the feature with a simple nasm DOS COM executable aka clean environment

assemble with recent nasm using `nasm -f bin bpmr.asm -o bpmr.com`

```
; nasm -f bin bpmr.asm -o bpmr.com
org 100h

mov si,test1
lodsb
lodsb
lodsb

mov byte [test2],'z'

mov al,byte [test1]

mov cx,3
mov si,test1
rep lodsb

mov si,test1
mov di,test2
movsb
movsb
movsb

mov cx,3 
mov si,test1
mov di,test2
rep movsb

mov ax,4C00h
int 21h

test1 db 'ABC'
test2 db 'abc'
```

# Checklist

I have:
 
* [x]  followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
* [x]  performed a self-review of my code.
* [x]  commented on the particularly hard-to-understand areas of my code.
* [x]  split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
* [x]  applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
* [x]  [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
* [x]  confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
* [ ]  added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
* [ ]  made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
* [ ]  [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -23,7 +23,7 @@ jobs:
             max_warnings: 17
           - name: MSVC 64-bit
             arch: x64
-            max_warnings: 882
+            max_warnings: 687
 
     steps:
       - name: Checkout repository
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -169,10 +169,12 @@ bool CPU_SetSegGeneral(SegNames seg,Bitu value);
 bool CPU_PopSeg(SegNames seg,bool use32);
 
 bool CPU_CPUID(void);
-Bitu CPU_Pop16(void);
-Bitu CPU_Pop32(void);
-void CPU_Push16(Bitu value);
-void CPU_Push32(Bitu value);
+
+uint16_t CPU_Pop16();
+uint32_t CPU_Pop32();
+
+void CPU_Push16(const uint16_t value);
+void CPU_Push32(const uint32_t value);
 
 #define EXCEPTION_DB			1
 #define EXCEPTION_UD			6
--- a/include/debug.h
+++ b/include/debug.h
@@ -20,22 +20,33 @@
 #define DOSBOX_DEBUG_H
 
 #include "dosbox.h"
+#include "mem.h"
 
-void DEBUG_DrawScreen(void);
-bool DEBUG_Breakpoint(void);
+#if C_DEBUG
+void DEBUG_DrawScreen();
+bool DEBUG_Breakpoint();
 bool DEBUG_IntBreakpoint(uint8_t intNum);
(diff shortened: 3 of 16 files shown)


kcgen:
The previous code has been reviewed and approved in #3082, leaving just the warning cleanup.

The warnings cleanup only uses (or casts) to the type of the variables being assigned into. 

Tested the normal and simple cores on arm64 and x86-64 without issues.

Will plan to merge later today - shout otherwise!

kcgen:
Merging. Tested on arm, x86_64, and unit tested on ppc64le.

gzip:
@LowLevelMahn Are these breakpoints supposed to be shown with `BPLIST`? It looks like they're taking space and can be deleted but they don't show up in the list. For example, if I do `BP es:1234`, `BPMR es:5678`, `BPLIST`, it only shows one entry: `01: 1000:1234`. If I then do `BPDEL 0` it gives a success message and the list changes to `00: 1000:1234`. This is on 0.81-rc2.

LowLevelMahn:
oh sorry, never used BPLIST and forget to test it - i think they should be visible and behave as every other breakpoint type, my fault, do you got an PR?

gzip:
I don't have a PR but I'll file a separate issue.

PR #3193, opened 2023-12-10
Packaging improvements for icons on Linux

weirddan455:
# Description
This should be the final PR for getting icon files up to snuff on Linux.

First commit was an oversight when I originally edited this icon Makefile.  The install section was missing the new high-res icons.  This is used by the `create-package.sh` script which creates the package that ends up for download on our website.

Second commit creates a script intended to be run by the end user.  This installs the icons and desktop shortcut in the correct place on their system, either system-wide or user-specific (default).

## Related issues
#3169

# Manual testing
Ran `create-package.sh` script.  Verified the high-res icons plus by new script appears in the created package directory.

Ran my new script `install-icons.sh` and verified the icons get installed.  Verified the `Exec=` line in the .desktop file gets replaced by the real dosbox binary location.  Verified icons get removed with the `--remove` flag.

# Checklist

I have:


contrib/linux/install-icons.sh
```diff
@@ -0,0 +1,47 @@
...
+
+strstr() {
+    [ "${1#*$2*}" = "$1" ] && return 1
+    return 0
+}
+
+usage() {
+    echo "Installs icon files and desktop shortcut."
+    echo "--system: Install to system wide directory (/usr/share)"
+    echo "--remove: Remove icons and desktop file"
```
johnnovak, line 15: desktop shortcut?

contrib/linux/install-icons.sh
```diff
@@ -0,0 +1,47 @@
...
+USAGE=false
+
+strstr() {
+    [ "${1#*$2*}" = "$1" ] && return 1
+    return 0
+}
+
+usage() {
+    echo "Installs icon files and desktop shortcut."
+    echo "--system: Install to system wide directory (/usr/share)"
```
johnnovak, line 14: minor: system-wide

contrib/linux/install-icons.sh
```diff
@@ -0,0 +1,47 @@
...
+    echo "--remove: Remove icons and desktop file"
+}
```
johnnovak, line 16: minor: Would be a bit nicer to indent the options by 2 spaces at least?

johnnovak (approved):
Nice one @weirddan455 🎉

contrib/linux/install-icons.sh
weirddan455: It's the .desktop file.  It creates a shortcut to Dosbox in the "start menu" (idk what the proper name is in Linux, I just call it that since Windows 95).  Actual location can vary based on desktop environment.  This is also what tells the desktop environment to use the icons.  Tested in KDE that the icon shows even if you launch Dosbox from the terminal (as long as the .desktop file is present).

contrib/linux/install-icons.sh
weirddan455: fixed

contrib/linux/install-icons.sh
weirddan455: Sure, that looks better.  Changed.

contrib/linux/install-icons.sh
johnnovak: Cool, I just noticed you're saying "desktop shortcut" in the first line, that's all 😄

Grounded0:
Isn't the "correct" way to declare Bash shell script:

`#!/usr/bin/env bash`

[https://www.baeldung.com/linux/bash-shebang-lines](The Difference Between #!/usr/bin/bash and #!/usr/bin/env bash)

contrib/linux/install-icons.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
```
kcgen, line 1: For the header, check the .sh files in the ./scripts area.
(They have an SPDX, short description, and usage).

kcgen: Suggest using `set -e` at the top to quit on the first error. 
Or if using bash, `set -euo pipefile` to quit on the first error, and undefined variables, and also any inter-pipe failures.

johnnovak:
> Isn't the "correct" POSIXy way to declare Bourne shell shell script:
> 
> `#!/usr/bin/env sh`
> 
> [The Difference Between #!/usr/bin/bash and #!/usr/bin/env bash](https://www.baeldung.com/linux/bash-shebang-lines)
> 
> [Make Linux/Unix Script Portable With #!/usr/bin/env As a Shebang](https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html)
> 
> This probably comes off as musings of a Unix portability nutter but then again i have used SGI IRIX at my home regularly since 2004. 🤷‍♂️

Yeah, I think that's highly recommended to increase the robustness of the script.

contrib/linux/install-icons.sh
weirddan455: Added the header and the `set -e`.  I suppose I'll try to stay POSIX compliant in order to support the odd BSD user.  I believe most of the desktop stuff is going to be the same on BSD as it is on Linux.

weirddan455:
> Isn't the "correct" POSIXy way to declare Bourne shell shell script:
> 
> `#!/usr/bin/env sh`
> 
> [The Difference Between #!/usr/bin/bash and #!/usr/bin/env bash](https://www.baeldung.com/linux/bash-shebang-lines)
> 
> [Make Linux/Unix Script Portable With #!/usr/bin/env As a Shebang](https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html)
> 
> This probably comes off as musings of a Unix portability nutter but then again i have used SGI IRIX at my home regularly since 2004. 🤷‍♂️

Sure.  I never researched the difference before.  Changed as you suggested.

johnnovak:
Is this ready to be merged @weirddan455 ?

weirddan455:
> Is this ready to be merged @weirddan455 ?

Yep.

PR #3194, opened 2023-12-10
Only capture the visible image in rendered image capture mode

johnnovak:
# Description

When using the `relative` viewport mode to stretch the image beyond the extents of the window or screen (e.g., when aspect ratio correcting Hercules games), rendered image captures contained the _whole_ stretched texture, not just the visible parts when using the OpenGL backend. This is wrong—rendered captures should always work in a WYSIWYG manner.

With the texture backend, the rendered capture action outright failed; at least on macOS I was getting the following error:

```
AGX: Texture read/write assertion failed: (xoffset + width) <= ALIGNGRAN_NPOT(getViewLevelWidth(mipmapLevel), block_width) && "Region width OOB"
```

So this PR fixes that and makes rendered captures fully WYSIWYG again. Upscaled captures still save the entire image. This is probably fine for now, but we might rethink it in the future. That would be a more involved change, though.

## Before

Prince of Persia, Hercules mode, image corrected to fill the whole 4:3 "emulated monitor" with the following config:

```ini
[render]
aspect = stretch
viewport = relative 112% 173%
```

This is what we see on the screen:

image

And this is the rendered capture; the whole stretched texture gets saved, it's not clipped to the extents of the viewport:

image0010-rendered

## After

The rendered capture is correctly clipped to the extents of the viewport and it works with the texture backend too:

### OpenGL

image0009-rendered

### Texture

image

image0002-rendered

# Manual testing

Only posting the texture results as that output mode is more "sensitive", but I've tested with OpenGL as well.

## Case 1

```ini
[render]
aspect = stretch
viewport = relative 50% 200%
```

image

image0003-rendered

## Case 2

```ini
[render]
aspect = stretch
viewport = relative 200% 50% 
```

image

image0005-rendered

## Case 3

```ini
[render]
aspect = stretch
viewport = relative 50% 50% 
```

image

image0008-rendered

# Checklist

I have:

### DIFF
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -93,6 +93,11 @@
 //   shaders, which is the image that is displayed on the host monitor with
 //   1:1 physical pixel mapping.
 //
+//   Because the viewport can be larger than the canvas, the draw area can be
+//   larger too. In other words, the draw rectangle can extend beyond the
+//   edges of the window or the screen in fullscreen mode, in which case the
+//   image is centered and the overhanging areas are clipped.
+//
 
 #define SDL_NOFRAME 0x00000020
 
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2543,9 +2543,24 @@ static std::optional<RenderedImage> get_rendered_output_from_backbuffer()
 {
 	RenderedImage image = {};
 
+	// The draw rect can extends beyond the bounds of the window or the screen
+	// in fullscreen when we're "zooming into" the DOS content in `relative`
+	// viewport mode. But rendered captures should always capture what we see
+	// on the screen, so only the visible part of the enlarged image.
+	// Therefore, we need to clip the draw rect to the bounds of the canvas
+	// (the total visible area of the window or screen), and only capture the
+	// resulting output rectangle.
+	
+	auto canvas_rect_px = get_canvas_size_in_pixels(sdl.rendering_backend);
+	canvas_rect_px.x = 0.0f;
+	canvas_rect_px.y = 0.0f;
+
+	const auto output_rect_px = canvas_rect_px.Copy().Intersect(
+	        to_rect(sdl.draw_rect_px));
+
 	auto allocate_image = [&]() {
-		image.params.width              = sdl.draw_rect_px.w;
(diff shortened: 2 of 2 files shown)


weirddan455 (approved):
Is it possible to have the rendered game image go "outside" the window?  I'm thinking back to how the CRT knobs worked and I seem to recall you could get a state where you're not seeing the entire image.  And old TVs did that by default called "overscan".

Anyway, I haven't used this feature but if this is possible, you should probably test an image capture with that.

This change looks to me though.  Just an edge case that popped into my head (which I'm not sure is even possible).

johnnovak:
Thanks for the review @weirddan455.

> Is it possible to have the rendered game image go "outside" the window? I'm thinking back to how the CRT knobs worked and I seem to recall you could get a state where you're not seeing the entire image. And old TVs did that by default called "overscan".

Yeah, definitely possible, that's what all this is about! 😄 Check out the problem description and the tests with the screenshots again—all tests exercise that exact scenario you described, for except for the last `relative 50% 50%` case).

weirddan455:
> Thanks for the review @weirddan455.
> 
> > Is it possible to have the rendered game image go "outside" the window? I'm thinking back to how the CRT knobs worked and I seem to recall you could get a state where you're not seeing the entire image. And old TVs did that by default called "overscan".
> 
> Yeah, definitely possible, that's what all this is about! 😄 Check out the problem description and the tests with the screenshots again—all tests exercise that exact scenario you described, except for the last `relative 50% 50%` case.

I see.  Your first couple of screenshots looked like you were testing the opposite case where the captured image had extra black space that shouldn't be there.

If both cases work now, all good.  This looks like a quick and easy fix.

johnnovak:
> I see. Your first couple of screenshots looked like you were testing the opposite case where the captured image had extra black space that shouldn't be there.

Yeah, so the key thing to look for is the `viewport` setting. If the percentage values are larger than 100%, that means the image will be stretched outside of the extents of the viewport (the window). For example, `viewport = relative 112% 173%`.

Then the manual tests are about interesting combinations where the image it stretched to extend beyond the window in _one_ dimension only, but not the other (e.g., `viewport = relative 50% 200%`).

PR #2825, opened 2023-09-03
[DRAFT] Website - Linux Untested builds, Development builds - architecture

Torinde:
Replacement for #2798
- development-builds.md - Linux and macOS architectures mentioned
- linux.md - add section for the Untested snapshot builds - Platforms.yml

PR #3202, opened 2023-12-13
Use zlib-ng in local builds and CI releases

### LINKED ISSUE #2812: ZMBV Video Encoder Performance Regression

kcgen:
# Description

zlib-ng is an actively maintained and optimized fork of zlib.
    
It includes SIMD optimizations for:
- Intel and AMD's SSE2/3/4, AVX (and more)
- ARM's NEON,
- Power ISA's Altivec, and
- RISC-V's "V" Vector (RVV) extensions
    
It's enabled by default for anyone compiling an optimized build
type provided they have both CMake and Meson version 1.3.0, as
this version of Meson added CMake subproject handling for wraps.
    
If those aren't provided or available, then the existing zlib system lib provides the deflate functionality.
    
Status is logged in the meson setup summary list:
```
Build Summary
  zlib provider: built-in
    
Subprojects
  zlib-ng: YES
```

### Benchmark
Threw in a benchmark to test your own system. Checkout this `kc/zlib-ng-5` branch and paste this in:

```shell
cd contrib/benchmark_zlibs/
meson setup build
meson compile -C build
```

Run a best of 3 comparison

```shell
./build/benchmark_zlib
./build/benchmark_zlib
./build/benchmark_zlib
```

```shell
./build/benchmark_zlib_ng
./build/benchmark_zlib_ng
./build/benchmark_zlib_ng
```

## Related issues

Fixes #2812 

# Manual testing

Tested ZMBV capture and frame capture.

# Checklist

I have:


FeralChild64:
Command meson setup `build/release` gives me the following summary:

```
dosbox-staging 0.81.0-alpha

  Build Summary
    Build type            : release

    Host page size (bytes): 4096
    zlib provider         : system library
    PNG provider          : system library
    SDL_net 2.0 support   : True
    slirp support         : True
    FluidSynth support    : True
    mt32emu support       : True
    ManyMouse support     : True
    ALSA support          : True
    Byte order            : little-endian
    Per-page W^X          : True
    CPU dynamic core      : optimized for x86/x86_64 (dyn-x86)
    CPU core target arch  : X86_64

Found ninja-1.11.1 at /usr/bin/ninja
```

despite I have a `sys-libs/zlib-ng-2.1.5` package installed in my Gentoo Linux system. The `meson compile -C build/release` command normally builds a binary, but `ldd ./build/release/dosbox | grep 'libz'` displays:
```
libz.so.1 => /lib64/libz.so.1 (0x00007f0491edc000)
libzstd.so.1 => /lib64/libzstd.so.1 (0x00007f048db6a000)
```
This is not expected, if zlib-ng is going to be used - `equery files sys-libs/zlib-ng` lists the following:
```
 * Searching for zlib-ng in sys-libs ...
 * Contents of sys-libs/zlib-ng-2.1.5:
/usr
/usr/include
/usr/include/zconf-ng.h
/usr/include/zlib-ng.h
/usr/include/zlib_name_mangling-ng.h
/usr/lib64
/usr/lib64/libz-ng.so -> libz-ng.so.2
/usr/lib64/libz-ng.so.2 -> libz-ng.so.2.1.5
/usr/lib64/libz-ng.so.2.1.5
/usr/lib64/pkgconfig
/usr/lib64/pkgconfig/zlib-ng.pc
/usr/share
/usr/share/doc
/usr/share/doc/zlib-ng-2.1.5
/usr/share/doc/zlib-ng-2.1.5/README.md.bz2
```

FeralChild64:
It works only if I use a _builtin_ library by adding `-Duse_builtin_zlib_ng=sse2,ssse3,sse42,avx2,pclmulqdq` to meson setup - there seems to be no way to use system-wide shared library.

kcgen:
Thanks @FeralChild64 ; I'll add support for a system-provided zlib-ng using name mangling.

contrib/benchmark_zlibs/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#include <array>
+#include <chrono>
+#include <cstdio>
+
+#include <zlib.h>
+
+constexpr auto one_megabyte = 1024 * 1024;
```
johnnovak, line 27: Minor: OneMegabyte

contrib/benchmark_zlibs/src/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+	const auto duration_us = duration_cast<microseconds>(end - start);
+	const auto duration_s  = duration_us.count() / 1000000.0;
+	const auto speed_mb_s  = static_cast<double>(data_in.size()) /
+	                        one_megabyte / duration_s;
+
+	printf("%.2f MB/s\n", speed_mb_s);
+
+	// OK!
+	return 0;
+}
```
johnnovak, line 82: Very cool that you've added a benchmark 🥳

johnnovak:
@weirddan455 Could you compare the performance of this branch with 0.80.1 on you aging hardware? 😅 Then if there is an improvement and it's significant, we could put a note about it into the release notes.

kcgen:
@FeralChild64, system support is now in.

## System scenario

1. Confirm zlib-ng is available in the system:

    `pkg-config --exists zlib-ng && echo true`

2. Setup with all defaults:

    `meson setup build/system`

3. Confirm both zlib-ng and zlib are provided by the system:

    ```
    zlib-ng provider: system library
    zlib provider   : system library
    PNG provider    : system library
    ```

When we detect zlib-ng, we mangle our zlib calls in ZMBV to make use of zlib-ng.  However, we're still at the mercy of PNG's own use of zlib, and therefore it still makes traditional `z_deflate()` calls, so that's why we need to provide it `zlib` as well.

## Requested built-in scenario

1. Confirm zlib-ng is available in the system:

    `pkg-config --exists zlib-ng && echo true`

2. Setup, but request built-in and native-optimized support

   `meson setup -Duse_zlib_ng=built-in,native build/req-built-in`

3. zlib is now provided entirely by zlib-ng (even PNG's calls to zlib will be serviced by zlib-ng :-).

    ```
    zlib provider: built-in (zlib-ng)
    PNG provider : built-in
    ```

4. If you compile, `ldd build/build/req-built-in/dosbox | grep z` should not have any zlib libraries present.

## Automatic built-in scenario

1. Uninstall zlib-ng at the system level

2. Confirm the library can't be found by `pkg-config`:

    `pkg-config --exists zlib-ng || echo not present`

3. Setup with all defaults:

    `meson setup build/auto-built-in`

4. The output and compiled binary should match the requested built-in scenario above.

weirddan455:
> @weirddan455 Could you compare the performance of this branch with 0.80.1 on you aging hardware? 😅 Then if there is an improvement and it's significant, we could put a note about it into the release notes.

I haven't looked at this branch yet but I already tested zlib-ng on main simply by using the `LD_LIBRARY_PATH` variable to make a drop-in replacement for normal zlib.  The performance gain is very significant.  There was constant lag when recording Doom with regular zlib and it was perfectly smooth with zlib-ng.

I can test this specific build later though and report some real numbers.

johnnovak:
> The performance gain is very significant. There was constant lag when recording Doom with regular zlib and it was perfectly smooth with zlib-ng.
> 
> I can test this specific build later though and report some real numbers.

That would be great, thanks. I just want to quantify the improvement; any rough number is better than say "it's now a lot faster" 😅

johnnovak:
I'm seeing a similar roughly 10x performance boost on my M1 Mac @kcgen. That's rocket fuel class improvement, indeed! 🥳 🚀 🔥

So glad you discovered this library and took the time to integrate it into our build process! 🥇 

### ./build/benchmark_zlib

```
224.42 MB/s
126.77 MB/s
128.57 MB/s
```

### ./build/benchmark_zlib_ng

```
1727.12 MB/s
2631.58 MB/s
1145.48 MB/s
```

johnnovak (approved):
Everything works as per the instructions on my M1 Mac. Excellent work @kcgen 🎉 🥳 🚀

kcgen:
Right now, CI will generate SSE2+SSSE3-enabled builds for x86-64 (Linux, macOS, and Windows MSYS2), and NEON-enabled builds for the Apple Silicon binaries.

I figured SSSE3 was pretty conservative as it's supported by the Core 2 Duo line. The risk here is that if someone only has a Pentium 4 and runs our release binaries, the zlib-ng code will generate an illegal instruction crash.

The solution for these users is building it themselves, which always provides native optimizations for whatever their processor supports (folks on newer machines will get AVX512 while someone building on a Pentium 4 will just get SSE2).

johnnovak:
> Right now, CI will generate SSE2+SSSE3-enabled builds for x86-64 (Linux, macOS, and Windows MSYS2), and NEON-enabled builds for the Apple Silicon binaries.
> 
> I figured SSSE3 was pretty conservative as it's supported by the Core 2 Duo line. The risk here is that if someone only has a Pentium 4 and runs our release binaries, the zlib-ng code will generate an illegal instruction crash.
> 
> The solution for these users is building it themselves, which always provides native optimizations for whatever their processor supports (folks on newer machines will get AVX512 while someone building on a Pentium 4 will just get SSE2).

I agree, assuming SSE3 is fine and it _IS_ conservative. Just found this interesting relevant piece of info (from 2015! 😎)

> **SSE and SSE2 are required for all processors capable of x64 native**, so they are good baseline assumptions for all code. Windows 8.0, Windows 8.1, and **Windows 10 explicitly require SSE and SSE2 support** even for x86 architectures so those instruction sets are pretty ubiquitous. In other words, if you fail a check for SSE or SSE2, just exit the app with a fatal error.

([source](https://stackoverflow.com/a/30356591))

johnnovak:
> Currently the Staging website mentions as supported for Windows/Linux any x86-64 CPU. Will you add a note there about zlib-ng usage increasing that to SSSE3? For macOS - website states any CPU supporting macOS 10.15 and those have AVX, so no impact.

What's the benefit of mentioning it, Torinde, when the SSE3 instruction set is standard starting from the Pentium 4 Prescott line? That was released in 2004...
https://en.wikipedia.org/wiki/SSE3

Do you honestly believe people will attempt to run Staging on a P4? 😄 How? When I don't even think Staging can run on earlier Windows versions than 7... We officially support Win 10. Can you even find a P4 motherboard that has Windows 10 or 7 drivers?...

weirddan455:
>  What's the benefit of mentioning it, Torinde, when the SSE3 instruction set is standard starting from the Pentium 4 Prescott line? That was released in 2004...
> https://en.wikipedia.org/wiki/SSE3

Small correction.  We're talking about SS**S**E3 which came after SSE3.

>  Supplemental Streaming SIMD Extensions 3 (SSSE3 or SSE3S) is a [SIMD](https://en.wikipedia.org/wiki/SIMD) instruction set created by [Intel](https://en.wikipedia.org/wiki/Intel) and is the fourth iteration of the [SSE](https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions) technology.

https://en.wikipedia.org/wiki/SSSE3

Naming things is hard but Intel probably could have done better than that.

I don't see a problem with making this a requirement for the binary.  zlib-ng doesn't require SSSE3 so the workaround would be simply compile without that compile flag if you want a build that runs on older hardware.

I also don't see a problem with mentioning it on the website.  SSSE3 required (anything made in the last 12 years should be fine).

kklobe:
> Right now, CI will generate SSE2+SSSE3-enabled builds for x86-64 (Linux, macOS, and Windows MSYS2), and NEON-enabled builds for the Apple Silicon binaries.
> 
> I figured SSSE3 was pretty conservative as it's supported by the Core 2 Duo line. The risk here is that if someone only has a Pentium 4 and runs our release binaries, the zlib-ng code will generate an illegal instruction crash.
> 
> The solution for these users is building it themselves, which always provides native optimizations for whatever their processor supports (folks on newer machines will get AVX512 while someone building on a Pentium 4 will just get SSE2).

I'll be the conservative dissenting voice - unless profiling shows a significant improvement of SSE2+SSSE3 vs. just plain SSE2, stick with the widest support.

FeralChild64:
I confirm, it links against system-wide `libz-ng` now.

kcgen:
Still left:
 - A couple small wording fixes
 - Let the benchmark's SIMD instructions be configurable, so we can answer your good question, @kklobe : is SSE2 good enough?

kcgen:
Ok, the benchmarks now support the `-Duse_zlib_ng=` options, so you can build some side-by-side binaries:

```shell
contrib/benchmark_zlibs

meson setup -Duse_zlib_ng=built-in,sse2       build/sse2
meson setup -Duse_zlib_ng=built-in,sse2,ssse3 build/ssse3
meson setup -Duse_zlib_ng=built-in,native     build/native

meson compile -C build/sse2
meson compile -C build/ssse3
meson compile -C build/native
```

Best of three...

**`./build/sse2/benchmark_zlib_ng`**
```
1329.79 MB/s
1364.26 MB/s
1345.90 MB/s
```

**`./build/ssse3/benchmark_zlib_ng`**
```
1976.28 MB/s
1953.12 MB/s
1949.32 MB/s
```

**`./build/native/benchmark_zlib_ng`**
```
2487.56 MB/s
2538.07 MB/s
2487.56 MB/s
```

kklobe:
Seems strange, it just keeps getting faster (using `native` on my M3 Pro):

```log
1273.89 MB/s
1666.67 MB/s
1633.99 MB/s
1901.14 MB/s
1893.94 MB/s
1937.98 MB/s
2222.22 MB/s
2079.00 MB/s
2331.00 MB/s
2222.22 MB/s
2217.29 MB/s
2352.94 MB/s
2506.27 MB/s
2538.07 MB/s
2512.56 MB/s
2638.52 MB/s
2557.54 MB/s
2808.99 MB/s
2717.39 MB/s
2624.67 MB/s
2710.03 MB/s
```

kcgen:
@kklobe , my guess is the workload is too small to get reliable timing numbers from it. 

You could edit the benchmark  and give it 1 GB of data instead of 1 MB, and also run it in a loop so the scheduler has no time to relax.

(I guess I could add an argument to dial up the data size, too).

kklobe:
I tried writing a test function that might generate more realistic data, and the numbers seem to be a bit more realistic and also top out a bit quicker:

```cpp
static data_array_t generate_screen_capture_data()
{
  data_array_t data = {};

  constexpr size_t block_size = 1024; // size of a block of same color
  constexpr uint8_t pattern_repeat = 10; // repeat a pattern every few bytes

  for (size_t i = 0; i < data.size(); ++i) {
    // Generate blocks of same color
    if ((i / block_size) % 2 == 0) {
      data[i] = 45;
    }
    // Generate repeating patterns
    else if ((i % pattern_repeat) == 0) {
      data[i] = 255;
    }
    // Fill remaining bytes with random data
    else {
      data[i] = static_cast<Bytef>(rand() % 256);
    }
  }

  return data;
}
```

```log
65.45 MB/s
107.77 MB/s
130.70 MB/s
145.39 MB/s
156.79 MB/s
166.69 MB/s
175.38 MB/s
177.68 MB/s
174.16 MB/s
185.39 MB/s
186.22 MB/s
```

I don't know if this means anything w.r.t. our decision to use SSE2 or SSE2+steroids :grin:

weirddan455:
I've pulled down this branch and done a quick benchmark with our real usage case (recording Doom).  Wrapped the `capture_video_add_frame` function with `GetTicksUs()` and logged the time per frame.  It only seems to do encoding work every other frame so I skipped ~100 microsecond frames.

System zlib as distributed by Arch Linux  (-Duse_zlib_ng=false) - 26ms
-Duse_zlib_ng=built-in - 5.3ms
-Duse_zlib_ng=built-in,sse2 - 4.5ms
-Duse_zlib_ng=built-in,sse2,ssse3 - 3.8ms
-Duse_zlib_ng=built-in,native - 3.8ms

Native on my system enables SSE4.2 and PCLMULQDQ but doesn't seem to have much effect on performance.  My CPU does not support AVX2 or above.

So whatever you guys decide for compile flags, I am happy with.  The big win is simply using this library to begin with.  Even without SSE2, zlib-ng is perfectly smooth on my system vs the laggy standard zlib.

johnnovak:
> -Duse_zlib_ng=built-in,sse2 - 4.5ms
> -Duse_zlib_ng=built-in,sse2,ssse3 - 3.8ms

Thanks @weirddan455. I honestly thought SSSE3 (with three S-es) was a typo! 😊 At least I learned something... 😝 

So, with SSSE3 it's still 20% faster. That's a significant difference. I don't like the idea of leaving performance on the table for people stuck on CPU released before 2012... I'd imagine that's like less than 1% of our users.

On the other hand, this really only matters for video capturing now, as image capturing was more than fast enough even with the standard zlib. @weirddan455 is planning to hardware-accelerated video capturing support, and in that scenario zlib won't even come into the picture...

...however, people on old hardware without hardware-accelerated encoding support might still benefit from the 20% SSSE3 bump 😎 

All in all, I think more people would benefit from SSSE3 than the very few people who'd be locked out of the awesome DOSBox Staging experience, so I vote for compiling with SSSE3 with confidence 😎 🤘🏻

johnnovak:
And yeah @Torinde, if this was SSE3 (two S-es) I wouldn't bother mentioning it, honestly. But ok, I think mentioning the SSSE3 requirement on the website makes sense.

I just like to keep things simple—that's where I'm coming from. 😄

kcgen:
> I tried writing a test function that might generate more realistic data, and the numbers seem to be a bit more realistic

@kklobe - thanks! That new data is so hard on the compressors that zlib-ng with the power of AVX2 only manages a 35% speed up against the bog-standard zlib :laughing: :hammer_and_pick: That'a tough nut to crack.

This is nice because this hard data might be closer to a lower-bound, while the easy data is more of an upper bound:

| data | base<br />zlib | SSE2<br />zlib-ng | SSSE3<br />zlib-ng | AVX2<br />zlib-ng |
| ---: | ---: | ---: | ---: | ---: |
| hard | 58.9 | 72.6 | 73.1 | 80.7 |
| easy  | 232.0 | 1369.8 | 2030.3 | 2442.0 |

Rates in MB/s

The updated benchmark runs through both datasets and generates very stable numbers.

Looking at @weirddan455's and my own timings added to ZMBV, I think video data is closer to “easy” data set numbers, which is great.

Fully agree @johnnovak - SSSE3 brings a large speedup over SSE2, 45% in my case (and possibly even more for newer chips); so that's a lot of performance all users will /not/ get if we fall back just to SSE2 for a vanishingly small number of users.

kcgen:
I don't think there's much left; reviewers, please shout if there's anything more left to change.

Will plan to merge in a couple hours.

kklobe:
I did the same timing test as @weirddan455 on the M3 Pro MacBook and it is a nice speedup there, too:

main:
```log
2023-12-14 20:42:37.308 | elapsed frame capture time = 9.597709ms
2023-12-14 20:42:37.337 | elapsed frame capture time = 9.579667ms
2023-12-14 20:42:37.365 | elapsed frame capture time = 9.846167ms
2023-12-14 20:42:37.395 | elapsed frame capture time = 10.460375ms
2023-12-14 20:42:37.424 | elapsed frame capture time = 10.492000ms
```

branch:
```log
2023-12-14 20:41:21.856 | elapsed frame capture time = 2.569583ms
2023-12-14 20:41:21.885 | elapsed frame capture time = 2.735250ms
2023-12-14 20:41:21.914 | elapsed frame capture time = 2.617542ms
2023-12-14 20:41:21.943 | elapsed frame capture time = 2.672625ms
2023-12-14 20:41:21.970 | elapsed frame capture time = 2.485250ms
```

No other comments from me, merge it :grin:

johnnovak:
> "The optimized zlib-ng library has been added to mitigate the performance hit when capturing 0.81's higher colour-depth AVIs and PNGs. Infact, zlib-ng not only makes up the difference, but speeds it up by another factor of three versus 0.80.1."

Thank for that man 🥳

Grounded0:
Website should be updated that running our Windows binaries has a SSSE3+ CPU feature level requirement now. That also lines up with cards offering a robust OpenGL 2.1 support which is somewhere around 2006 so matches perfectly.

Grounded0:
I think we should do performance comparisons between SSE3 and SSSE3 is the juice worth squeeze.

Grounded0:
Seems like the choices are SSE2 or SSSE3 from the documentation so we'll have to see how this is accepted by the users with the 0.81 RC.

johnnovak:
> For the Windows and Linux download pages - should I make a PR changing the text to:
> 
> Hardware requirements for the current release From the x86 family of processors, all processors with Supplemental Streaming SIMD Extensions 3 (SSSE3) are supported.

Yeah, but test it locally please.

johnnovak:
> OK, but then it'll take more time... I wanted to setup website local testing since #2825

I'll do it @Torinde, don't worry.

PR #3203, opened 2023-12-14
Use std math functions in `compressor.cpp`

johnnovak:
# Description

Standard C functions are best to be avoided if we don't specifically need exact conformance with what the C standard mandates. In particular, the `std::min` and `std::max` functions are potentially more performant as they can be compiled to branchless code as opposed to `fmin/fmax`.

Key point from this [extremely informative post](https://stackoverflow.com/a/40870480):

> In C++, use std::min and std::max, which are defined in terms of > or <, and don't have the same requirements on NaN behaviour that fmin and fmax do. Avoid [fmin and fmax](http://en.cppreference.com/w/c/numeric/math/fmin) for performance unless you need their NaN behaviour.

# Manual testing

Space Quest III still works and there is sound 😎 🤘🏻 

# Checklist

I have:

### DIFF
--- a/src/hardware/compressor.cpp
+++ b/src/hardware/compressor.cpp
@@ -24,7 +24,7 @@
 #include <array>
 #include <cassert>
 #include <cmath>
- 
+
 #include "checks.h"
 #include "mixer.h"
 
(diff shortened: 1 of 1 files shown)


kcgen (commented):
These are probably doing silent upgrades to `double`s; should be float-equivalent for each (std::expf, etc)

Edit: ignore that; these look like template calls in C++!

I tested with `meson setup -Dcpp_args='-Wdouble-promotion' build`

kcgen (approved):
:+1: nice cleanup!

johnnovak:
> These are probably doing silent upgrades to `double`s; should be float-equivalent for each (std::expf, etc)

I don't think that's the case. `std::min/std::max` are templated, and `std::exp`, `std::sqrt` etc. all have overloaded float, double, and long double variants. E.g.:

image


image

Source:

- https://en.cppreference.com/w/cpp/algorithm/min
- https://en.cppreference.com/w/cpp/numeric/math/sqrt
- https://en.cppreference.com/w/cpp/numeric/math/exp
- https://en.cppreference.com/w/cpp/numeric/math/pow

johnnovak:
> Edit: ignore that; these look like template calls in C++!
> I tested with meson setup -Dcpp_args='-Wdouble-promotion' build

Ah, thanks for that! Like I'm always saying: trust but verify 😏

PR #3209, opened 2023-12-15
Fix infinite recursion problem at startup on Pop OS 22.04 and SDL 2.0.20

johnnovak:
# Description

It turns out resetting the renderer is not needed on window size change events, and this somehow caused an infinite regression on Pop OS 22.04 running SDL 2.0.20 which manifested in high CPU load and a (mostly) black DOSBos Staging screen at startup.

# Manual testing

Regression tested on Windows 10 and macOS Sonoma 14.0:

- Confirmed that the shader auto-switching still works fine when changing the window size.
- Confirmed that the new pixel aspect ratios are still propagated downstream and the updated PAR gets logged when switching the `aspect` setting (e.g., `aspect = stretch` with integer scaling disabled).

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3984,10 +3984,7 @@ bool GFX_Events()
 
 				finalise_window_state();
 
-				const auto shader_switched = maybe_auto_switch_shader();
-				if (!shader_switched) {
-					GFX_ResetScreen();
-				}
+				maybe_auto_switch_shader();
 				continue;
 			}
 


weirddan455 (approved):
I was able to reproduce the bug in a VM on Pop OS 22.04.  This fixes the problem.

bmunger:
I also confirm it fixes the problem installed on bare metal.

johnnovak:
Thanks guys. I was super frustrated by this initially, and I'm glad the fix turned out to be rather simple 😅

PR #3214, opened 2023-12-15
Display 'max limit' cycles setting properly in the title bar

FeralChild64:
# Description

I have noticed that with `cycles = max limit <value>` the title bar only displays `max` value, without telling about the limit. This is very misleading and it might lead to mistake when running benchmark.

This PR fixes the problem. ~~I have set the project to _0.82_, as we are after code soft freeze - but if you think this should go to _0.81_, I have no objections.~~

Screenshot-NewTitleBar

# Manual testing

Try different `cycles = <value>` settings, use internal command `config -set cycles="<value>"`.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -312,22 +312,41 @@ void GFX_SetTitle(const int32_t new_num_cycles, const bool is_paused = false)
 	auto &hint_mouse_str  = sdl.title_bar.hint_mouse_str;
 	auto &hint_paused_str = sdl.title_bar.hint_paused_str;
 
-	if (new_num_cycles != -1)
+	if (new_num_cycles != -1) {
 		num_cycles = new_num_cycles;
+	}
 
 	if (cycles_ms_str.empty()) {
 		cycles_ms_str   = MSG_GetRaw("TITLEBAR_CYCLES_MS");
 		hint_paused_str = std::string(" ") + MSG_GetRaw("TITLEBAR_HINT_PAUSED");
 	}
 
-	if (CPU_CycleAutoAdjust)
-		safe_sprintf(title_buf, "%8s - max %d%% - " APP_NAME_STR "%s",
-		             RunningProgram, num_cycles,
-		             is_paused ? hint_paused_str.c_str() : hint_mouse_str.c_str());
-	else
-		safe_sprintf(title_buf, "%8s - %d %s - " APP_NAME_STR "%s",
-		             RunningProgram, num_cycles, cycles_ms_str.c_str(),
-		             is_paused ? hint_paused_str.c_str() : hint_mouse_str.c_str());
+	const auto& hint_str = is_paused ? hint_paused_str : hint_mouse_str;
+	if (CPU_CycleAutoAdjust) {
+		if (CPU_CycleLimit > 0) {
+			safe_sprintf(title_buf,
+			             "%8s - max %d%% limit %d %s - " APP_NAME_STR "%s",
+			             RunningProgram,
+			             num_cycles,
+			             CPU_CycleLimit,
+			             cycles_ms_str.c_str(),
+			             hint_str.c_str());
+		} else {
+			safe_sprintf(title_buf,
+			             "%8s - max %d%% %s - " APP_NAME_STR "%s",
+			             RunningProgram,
+			             num_cycles,
+			             cycles_ms_str.c_str(),
+			             hint_str.c_str());
+		}
+	} else {
+		safe_sprintf(title_buf,
+		             "%8s - %d %s - " APP_NAME_STR "%s",
+		             RunningProgram,
+		             num_cycles,
+		             cycles_ms_str.c_str(),
+		             hint_str.c_str());
+	}
 
 	SDL_SetWindowTitle(sdl.window, title_buf);
 }


src/gui/sdlmain.cpp
```diff
@@ -297,7 +297,7 @@ extern bool CPU_CycleAutoAdjust;
 bool startup_state_numlock=false;
 bool startup_state_capslock=false;
 
-void GFX_SetTitle(const int32_t new_num_cycles, const bool is_paused = false)
+void GFX_SetTitle(const int32_t new_num_cycles, const bool is_paused = false) // XXX
```
johnnovak, line 300: Remove XXX?

johnnovak:
I'm not against this going in @FeralChild64 as it can be considered a bug fix, it's trivial, and essentially zero-risk. Will test in detail a bit later.

The lack of the "limit" indicator has been bothering me too.

src/gui/sdlmain.cpp
FeralChild64: OK, sorry.

FeralChild64:
OK, moved to 0.81 project.

kcgen (approved):
Very helpful improvement, @FeralChild64 !

johnnovak:
@FeralChild64 I've tested it, it works, but I think the formatting could be improved.

E.g., for `cycles 20000` I get this:

image

But for `cycles max 80%` this:

image

And for `cycles max 80% limit 200000`

image

---

So, the problems:

- `cycles/ms` is omitted for anything else that's not a fixed cycles setting.
- I'd to see the same argument order in the titlebar as for the `cycles` setting; that's just the least confusing. So for say `max 90% limit 200000` I'd like to see the exact same string in the titlebar. In general, the comma implies that the percentage value refers to something else? Who knows what? 😅 
- For `max 50%` we get `max, 50%` — that's rather weird and confusing, as per above.

My suggestions:

- `cycles 20000` -> output: `20000 cycles/ms`
- `cycles max 80%` -> output: `max 80% cycles/ms`
- `cycles max 80% limit 200000` -> output: `max 80% limit 200000 cycles/ms`

You can argue that `max 80% cycles/ms` looks a bit strange. But I still prefer to always put `cycles/ms` at the end for consistency; that's simply the least surprising thing to do. Then user knows, _"ok, whatever is before "cycles/ms" refers to the emulated CPU cycles setting"._

FeralChild64:
@johnnovak OK, changed according to your idea.

My main concern is that `cycles/ms` is translatable, while `max` and `limit` are not (and we don't want them to be translatable, as they are keywords used in config file) - that's why I wanted to avoid mixing them.
Let's see if anyone complains :)

johnnovak:
> @johnnovak OK, changed according to your idea.
> 
> My main concern is that `cycles/ms` is translatable, while `max` and `limit` are not (and we don't want them to be translatable, as they are keywords used in config file) - that's why I wanted to avoid mixing them. Let's see if anyone complains :)

I think translating `cycles/ms` is a bad idea. `cycles` is the name of a config setting -- we don't want to translate config settings, do we? 😄 Plus if anyone doesn't understand _that much_ English, they will have lots of problems with using English DOS commands 😏 Or reading the logs, etc.

FeralChild64:
While `cycles` is a config setting, `cycles/ms` is not. I would prefer to leave it to the translator.

johnnovak:
> While `cycles` is a config setting, `cycles/ms` is not. I would prefer to leave it to the translator.

You can do that, but if you don't allow it, you'd avoid the mixing of max, limit and translated cycles/s that you did not like. I personally think that's over the top as everybody understands cycles/s. But up to you 🤷🏻

Grounded0:
Lets roll with the translation folks. I don't mind if no one is pressuring me to produce Finnish or Swedish translation.

FeralChild64:
At least in Polish the translated string looks OK. In a situation where it does not (_Reverse_ mode in the `MIXER` command), I'm leaving it in English.

----

I start to hate the pipeline compilation - fails again, like on the main branch :/

kcgen:
@FeralChild64 ; I'll look into this msys2 download issue.

johnnovak:
Tested again, thanks for addressing the changes @FeralChild64 . Now I like it 😄 

Vaguely related: not too sure about the value of chucking "DOSBox Staging (debug build)" or "DOSBox Staging" in the middle of the title bar. That's one busy title bar 😅 I'd put " (debug)" at the end because it helps with local dev, but that's it.  But job for later... 

image

johnnovak (approved):
Nice improvement @FeralChild64 🍻

Grounded0:
@johnnovak 

DOSBox Staging should be first followed by the running executable name (this is actually useful information) but thats another PR.

johnnovak:
> @johnnovak
> 
> DOSBox Staging should be first followed by the running executable name (this is actually useful information unlike on multitasking OSes) but thats another PR.

We can discuss later, sure. I see it as something pointless; what does these names tell anyone `GAME`, `SIERRA`, `SCIV`, `START`, `INSTALL`. DOS exe names are pretty much nonsensical. I personally never ever look at, I see it as noise.

Grounded0:
Good thing is that its never longer than 8 characters. 🙂

johnnovak:
> Good thing is that its never longer than 8 characters. 🙂

Yeah, at least it's short semi-random noise, haha 😎 🤘🏻

FeralChild64:
@Grounded0 But, together with spacing/separators, it can take 11 characters :)

@johnnovak I don't like this name in the title bar either; if there are no objections from the team, removal would be rather simple: https://github.com/dosbox-staging/dosbox-staging/tree/fc/titlebar-2

johnnovak:
> @Grounded0 But, together with spacing/separators, it can take 11 characters :)
> 
> @johnnovak I don't like this name in the title bar either; if there are no objections from the team, removal would be rather simple: https://github.com/dosbox-staging/dosbox-staging/tree/fc/titlebar-2

Please raise a PR @FeralChild64 & let's do it 😎

rderooy:
dosbox-x has a setting:

```ini
[dosbox]
title = MyGame
```

that I like, it allows me to set the title to the actual name of the game, and not just "GAME" or "SIERRA". Issue is of course that this title can be longer, so pushing other text further out to the right and possibly cut it of as a result.

FeralChild64:
@rderooy Good idea, but I suggest to do this after 0.81.0. This will require additional help strings, plus possibly some experimentation (maybe we should skip the cycles values display in such case?). IMHO no time for this anymore if we want to release RC soon.

PR #3217, opened 2023-12-16
Always clamp `frames_needed` to the allowed mix/max range in the mixer

johnnovak:
# Description

This fixes a regression by reverting commit a25af54. The clamping causes the audio to become out of sync when using fast-forward mode in _release_ mode (debug mode is not fast enough, but you can still get the audio out-of-sync by fast forwarding, but to a much lesser extent). Tested with OPL and MT-32 output, but probably all sound devices are affected.

The lack of clamping doesn't cause issues in release builds; it just helped not to trip an assert in debug builds when putting your laptop to sleep while running Staging, then waking the machine up a few seconds later. But this was never a proper fix anyway, just a workaround, so it's no great loss.

The problem should be properly addressed when the whole mixer gets refactored. I really don't have the time for that, and we definitely don't want to break the sound sync when fast-forwarding.

## Related issues

Reverts https://github.com/dosbox-staging/dosbox-staging/pull/2843

# Manual testing

## Before fix

1. Start Space Quest 3 with OPL or MT-32 sound (but any game will do).
2. Press fast-forward for a few seconds.
3. When you release the fast-forward hotkey, the audio will "rush" to catch up with the "current time". This sounds quite comical and quite wrong 😄 

## After fix

Repeat the same steps—the audio always stays in perfect sync, no matter for how long you've been fast-forwarding.

# Checklist

I have:

### DIFF
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -219,16 +219,6 @@ StereoLine MixerChannel::GetLineoutMap() const
 	return output_map;
 }
 
-// TODO Once the mixer code is thorougly refactored, revisit whether this is
-// still necessary (i.e., we might be able to be more precise with our
-// 'frames_needed' calculation so we never under or overshoot).
-static int clamp_frames_needed(const int frames_needed)
-{
-	return clamp(frames_needed,
-	             mixer.min_frames_needed.load(),
-	             mixer.max_frames_needed.load());
-}
-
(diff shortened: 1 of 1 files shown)


kcgen (approved):
Nice mitigation, @johnnovak !
This is perfect because if/when the mixer is refactored more, the clamps can be inspected and checked - until we figure out the root cause for overages (or at least, limit the overage right at the source).

PR #3219, opened 2023-12-16
Fix 'auto' presentation mode on macOS in fullscreen

johnnovak:
# Description

**_TL;DR for the impatient 😄 Read the whole discussion in the comments section to the get the full picture and reasoning._**

The special casing for macOS causes incorrect output in fullscreen at 120 Hz with the default settings (half-updated 
frames) when the screen is only updated sporadically. This is common in older games that update the screen only a few times per second, or not even update it at all for as long as there's no user input (which can be several seconds). That's a sizable chunk of the DOS gaming catalogue, and the macOS special performance optimisation path broke these games with the default config in fullscreen @ 120 Hz.

By reverting the macOS special path, we're defaulting to the *correct* (not the *fastest*    possible) behaviour out-of-the-box with everything left at default/auto for the broadest possible range of games. The whole auto presentation mode logic also becomes easier to maintain and reason about.

---

Trying to put this as simply as possible. Please bear with me 😅 

Some base truths first:

1. When `auto` mode selects CFR on macOS in fullscreen, it does **not** behave the same way as setting `presentation_mode = cfr` manually.
2. The "real CFR" mode via `presentation_mode = cfr` works as expected, no issues.
3. But "CFR via auto mode" **just on macOS** causes a problem, namely tearing/missed half-frame refreshes. **Only** on macOS, and **only** in "CFR via auto mode"... which is not quite CFR, as we'll see 😅 

This simple change fixes that and makes "CFR via auto mode" work exactly as manual CFR.

Also, I tested manual CFR and "CRT via auto mode" on Windows in windowed & fullscreen, and with 50/60/70 Hz host and 60/70 Hz DOS rate combinations. **All is fine on Windows.** Luckily, the changed code path only affects macOS and nothing else.

I **cannot** create a video capture of this because video capturing somehow forces full redraws. You'll have to trust me on this one 😎 (and I'm sensitive as fuck to weird screen tearing / missed refresh issues), but it's trivial from inspecting the code that real and "auto" CFR **didn't** work the same way prior to this fix.

## Repro steps

I used **Leisure Suit Larry 2** for testing on my 120 Hz adaptive sync MacBook. Of course, the issue is present in program that doesn't continuously produce new frames.

1. Start the game with `presentation_mode = auto` and everything on defaults, then switch to fullscreen. 
2. Stay on the first screen of the game and start moving the cursor around _slowly_ above Larry. Decreasing cycles to ~150 helps trigger the issue. About 50% of the time, the cursor will be only _half drawn_ (half of it is missing!).
3. Walk around with Larry sideways at ~150 cycles with the left/right cursor keys. You'll see the occasional tearing, e.g., his upper torso  still shows the _previous_ animation frame. By starting/stopping movement by presseing the left cursor key repeatedly, if you stop in the right moment in the middle of a "torn frame" and don't move the mouse to force another frame, the tearing will stay there indefinitely (until a new frame arrives). Again, this happens at faster cycles settings too (I noticed it at 3000 cycles), but it's easiest to see at low 100-150 cycles.

## Analysis

Turns out that code path that calls `save_rate_to_frame_period(host_rate)` can only happen on macOS in fullscreen when the host refresh rate is higher than the DOS rate. But obviously it yields incorrect results. Passing in `dos_rate` instead, like the "true CFR path" 100% fixes the issue.

I don't fully understand how this issue manifests, but this is the fix based on empirical evidence (aka dicking around 😎 🤘🏻).

# Manual testing

With this fix, the weird tearing issues are 100% gone on my 120 Hz MacBook screen in fullscreen.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1116,24 +1116,6 @@ static void setup_presentation_mode(FrameMode &previous_mode)
 		const auto display_might_be_interpolating = (host_rate >=
 		                                             InterpolatingVrrMinRateHz);
 
-		const auto conditions_demand_constant_host_rate = (
-#if defined(MACOSX)
-		        // Normally OpenGL drivers with a swap-interval of 1
-		        // only block for the gap in time after the application
-		        // requests a buffer swap through to presentation of the
-		        // current frame.
-		        //
-		        // However, circa-2023 macOS OpenGL drivers impose
-		        // additional per-frame block penalities if frames
-		        // /aren't/ presented. Therefore, this condition demands
-		        // we present at a constant host rate.
-		        //
-		        sdl.rendering_backend == RenderingBackend::OpenGl &&
-		        get_vsync_settings().requested == VsyncState::On);
-#else
-		        false);
-#endif
-
 		// If we're fullscreen, vsynced, and using a VRR display that
 		// performs frame interpolation, then we prefer to use a
 		// constant rate.
@@ -1152,22 +1134,15 @@ static void setup_presentation_mode(FrameMode &previous_mode)
 		        supported_rate >= dos_rate
 		                ? "Host can handle the full DOS rate"
 		                : "Host cannot handle the DOS rate");
-		if (conditions_demand_constant_host_rate) {
-			LOG_MSG("SDL:   - CFR selected because we're on macOS using"
-			        " OpenGL with vsync on");
-		} else {
-			LOG_MSG("SDL:   - %s",
-					conditions_prefer_constant_rate
-							? "CFR selected because we're fullscreen, "
-							"vsync'd, and display is 140+Hz"
-							: "VFR selected because we're not "
-							"fullscreen, nor vsync'd, nor < 140Hz");
-		}
+		LOG_MSG("SDL:   - %s",
+		        conditions_prefer_constant_rate
+		                ? "CFR selected because we're fullscreen, "
+		                  "vsync'd, and display is 140+Hz"
+		                : "VFR selected because we're not "
+		                  "fullscreen, nor vsync'd, nor < 140Hz");
 #endif
-		if (conditions_demand_constant_host_rate) {
-			mode = FrameMode::Cfr;
-			save_rate_to_frame_period(host_rate);
-		} else if (supported_rate >= dos_rate) {
+
+		if (supported_rate >= dos_rate) {
 			mode = conditions_prefer_constant_rate ? FrameMode::Cfr
 			                                       : FrameMode::Vfr;
 			save_rate_to_frame_period(dos_rate);
@@ -4353,6 +4328,7 @@ static void config_add_sdl() {
 	        "             host rate (e.g. 70 Hz DOS rate vs 60 Hz host rate).\n"
 	        "  adaptive:  Enables vsync when the frame rate is higher than the host rate,\n"
 	        "             but disables it when the frame rate drops below the host rate.\n"
+	        "             This is a reasonable alternative on macOS instead of 'on'.\n"
 	        "             Note: only valid in OpenGL output modes; otherwise treated as 'on'.\n"
 	        "  off:       Attempt to disable vsync to allow quicker frame presentation at\n"
 	        "             the risk of tearing in some games.\n"


kcgen:
This chunk of macOS-specific code resulted from a performance issue @Burrito78 reported when running the PERFDOS benchmark with OpenGL and vsync enabled.

When vsync is enabled, macOS's OpenGL drivers _need_ us to present at exactly the **_host rate_**, no matter what.

See the blue highlights:



The problem with using the DOS rate is that it's **_not the host rate_**:

1. When the DOS rate is _slower_ than the host, then we're under-presenting and frame rates and stalls will happen. I re-tested this and confirmed PERFDOS numbers collapse into 1/4th when we're not filling frames 1:1 at the host rate.

2. When the DOS rate is _faster_ than the host (ie: 70 Hz DOS game on a 60 Hz host), then we're presenting too fast, and we'll be vsync-blocked.
    Comparing this branch against main, vsync=on, everything else default, w/ Chris's 3D benchmark (320x200 mode). main gets a score of 490 and branch gets ~320 on my M1 mini. Those vsync blocks literally pause the emulator for up to 7 ms every couple frames.

Here's what I'd suggest inside the `conditions_depand_constant_host_rate` scope:

1. Use the `std::max(dos_rate, host_rate)` instead of just the `dos_rate`.
 
2. Add a warning when `dos_rate > host_rate`: "The DOS rate of %d exceeds the host's rate of %d Hz, but vsync is configured on. This will produce vsync stalls that will impact benchmarks and high FPS games; consider setting `vsync=off` or `vsync=adaptive` as alternatives."

The result is a bit better:
- If the DOS rate happens to be below the host rate, then we at least get the first issue solved by stepping up to run 1:1 w/ the host.
- If the DOS rate is faster than the host rate, then we at least warning them about performance problems and recommend some options.

johnnovak:
@kcgen Thanks for taking the time to explain this, I understand the problem better now, but your proposed fix does not fix the issue I'm trying to fix here 😄 Let's sidestep technicalities for a moment—I'm sure they're all true and you've done your research, and I don't doubt your results.

All I'm trying to fix here is the default behaviour on 120 Hz on macOS in fullscreen, which is a common scenario. That default experience with everything on auto, so stock config, is sadly broken for half of DOS catalogue for games that only update the screen sporadically.

I guess Burrito and me represent two category of users here:

1. Burrito would like to achieve max FPS and no stalls in fast-paced 3D games, so mid to late 90s DOS era.
2. I don't want to have any broken screen updates in pre-3D games that update the screen sporadically (maybe only once every few seconds—that low!). That's the entire 80s catalogue, and a good chunk of the early to mid 90s.

If optimising things for one category causes broken behaviour for the other category with default settings, I think we have a problem. We want to take a compromise then to at least make the out-of-the-box experience non-broken for both  game categories, maybe at the price of sacrificing some performance in some edge cases.

With the "macOS special fullscreen hack" I'm getting get weird half-frame updates out-of-the-box in fullscreen with "auto CFR... but not really" mode 😄, which is super jarring in gamee that updates the screen only rarely, sometimes not even updating for a second! I'll upload a video about this.

The only fix that works is `presentation_mode = cfr` manually. Also, it's quite confusing that "auto CFR" is different to "manual CFR".

My suggestion: get rid of this whole macOS special thing in auto mode. Auto CFR will be regular CFR, full stop. By the way, by removing _all_ that special macOS code, I'm getting VFR selected in 120 Hz in fullscreen which doesn't exhibit the problem either.

In a way, the full out-of-the-box auto mode now yields worse overall results on 120 Hz + macOS + fullscreen than setting either VFR or CFR manually—this is not ideal. Having to tell people "if you play older games that don't update the screen all the time, yeah, you can't use the stock settings on a 120 Hz macBoo, sorry", well, that's not great 😄 

I can upload a video I recorded with my phone later that shows the half frame updates clearly, and they are jarring. But I hope this is enough to help us reach a decision. To be clear, I'm advocating to remove _all_ that macOS special code, more than in my PR, and just default to some middle-of-the-road setting performance wise that has zero artifacts.

kcgen:
I don't like the special handling for macOS, either! Yeah, I say eliminate all of that stuff, and if there's still problems with vsync on my 60 Hz display, I'll PR a warning log message about.

kcgen:
> I guess Burrito and me represent two category of users here

Like having a Dad that starts tearing the seats and arm rests out of the family sedan.. for better power to weight ratios in the quarter mile. 
Comfort vs. speed; pick one!🔥

johnnovak:
> No worries about a recording; I've seen lots of those half frames before 😅.
> 
> I don't like the special handling for macOS, either! Yeah, I say eliminate all of that stuff, and if there's still problems with MacOS w/ OpenGL + vsync on my 60 Hz display, I'll PR a warning log mention potential performance impact with the suggestion to use adaptive or turn it off.

Will do that. To reiterate, I appreciate all the work, investigation, and testing you've put into the

> > I guess Burrito and me represent two category of users here
> 
> Like having a Dad that starts tearing the seats and arm rests out of the family sedan.. for better power to weight ratios in the quarter mile. Comfort vs. speed; pick one!🔥

Haha, exactly 😎 Good thing is we _do_ support such tuning options for performance freaks—but they'll need to roll up their sleeves and do some experimentation themselves. But that's power user _"you've just voided your warranty there, pal"_ territory 😎 🤘🏻 

The simpletons like me who only want to get from A to B can simply just get in the car and drive it safely to their destination... without the seat catapulting mid-journey or some weird shit 😅 

To reiterate, I appreciate all the work, research, and testing you put into this whole presentation mode business. I'm getting more familiar with it slowly, but man, there's is a lot of complexity there 😅

kcgen:
No worries 🍻 ! I do agree that simplifying this (by eliminating this macos chunk) is the right approach; it makes the presentation mode handling generic again, which is what I want, as well.

johnnovak:
I've reverted the macOS special path @kcgen and I've updated the commit message and the top description.

With this, the out-of-the-box experience @ 120 Hz in fullscreen on macOS has been restored — no more weird tearing and missed half-frame updates in any older games I tried! 🎉

kcgen (approved):
Looks good!

PR #3220, opened 2023-12-16
Log display mode correctly on fullscreen/windowed switch & window resize

johnnovak:
# Description

Before this change, if you switched to fullscreen, then back to windowed mode, the display mode params were logged "too early". Specifically, the scaled pixel dimensions were off; the _previous_ pixel dimensions you had in fullscreen mode were logged again after going back to windowed. This is because the logging happened too early in the event handling chain.

Now we're also logging the changed display dimensions on window size changes. The way I'm doing it shouldn't spam the logs _while_ resizing the window on any Linux window manager, just log _once_ when you stop resizing, but we'll see.

Windows and macOS are fine; they don't send continuous resize events (it's a challenge to get continuous resize events on these platforms, in fact, but Linux tends to do this differently).

One drawback of this approach is that on window resizes, the display settings get logged _twice_, at least on macOS, because there are _two_ `SDL_WINDOWEVENT_RESIZED` events in a row. Why, that's anyone's guess 🤷🏻 Maybe SDL bug. Personally, I'd rather have two log messages in those cases than zero and this is something we could perhaps improve later on but it's hardly a deal-breaker.

# Manual testing

## Before fix

Use the defaults by staring DOSBox with `--noprimaryconf`.

1. Switch to fullscreen. The `DISPLAY: ...` log message contains something like `scaled to 2667x2000 pixels` in the middle.
2. Switch back to windowed. The display log message still contains the same dimensions, but clearly that can't be true.

Changing the window size does _not_ log the new display settings.

## After fix

1. Switch to fullscreen. The `DISPLAY: ...` log message contains something like `scaled to 2667x2000 pixels` in the middle.
2. Switch back to windowed. The display log message contains smaller pixel dimensions such as `scaled to 1600x1200 pixels` in my case.

Changing the window size _does_ log the new display settings when you stop resizing.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -767,23 +767,13 @@ static DosBox::Rect get_canvas_size_in_pixels(
 static void log_display_properties(const int render_width_px,
                                    const int render_height_px,
                                    const VideoMode& video_mode,
-                                   const std::optional<SDL_Rect>& canvas_size_override_px,
                                    const RenderingBackend rendering_backend)
 {
-	const auto draw_size_px = [&]() -> DosBox::Rect {
-		if (canvas_size_override_px) {
-			return calc_draw_rect_in_pixels(
-			        {canvas_size_override_px->w,
-			         canvas_size_override_px->h});
-		} else {
-			const auto canvas_size_px = get_canvas_size_in_pixels(
-			        rendering_backend);
-
-			return calc_draw_rect_in_pixels(canvas_size_px);
-		}
-	}();
-
 	assert(render_width_px > 0 && render_height_px > 0);
+
+	const auto canvas_size_px = get_canvas_size_in_pixels(rendering_backend);
+	const auto draw_size_px = calc_draw_rect_in_pixels(canvas_size_px);
+
 	assert(draw_size_px.HasPositiveSize());
 
 	const auto scale_x = static_cast<double>(draw_size_px.w) / render_width_px;
@@ -2220,7 +2210,6 @@ uint8_t GFX_SetSize(const int render_width_px, const int render_height_px,
 		log_display_properties(sdl.draw.render_width_px,
 		                       sdl.draw.render_height_px,
 		                       sdl.video_mode,
-		                       {},
 		                       sdl.rendering_backend);
 	}
 
@@ -2391,9 +2380,8 @@ void GFX_SwitchFullScreen()
 	sdl.desktop.switching_fullscreen = true;
 
 	// Record the window's current canvas size if we're departing window-mode
-	auto& canvas_size_px = sdl.desktop.window.canvas_size;
 	if (!sdl.desktop.fullscreen) {
-		canvas_size_px = to_sdl_rect(
+		sdl.desktop.window.canvas_size = to_sdl_rect(
 		        get_canvas_size_in_pixels(sdl.rendering_backend));
 	}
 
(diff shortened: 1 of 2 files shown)


kcgen (approved):
Looks more straightforward; no comments 👍

johnnovak:
Thanks for the quick review @kcgen.

I've just added an additional commit that "debounced" the duplicate logging of the display params when changing the window size. Although this is a workaround, it's fairly bening, clean, and it only affects the logging. It's unfortunate we need to do this, but it cannot have unwanted consequences, so I think it's fine.

src/gui/sdlmain.cpp
```diff
@@ -3832,11 +3833,30 @@ bool GFX_Events()
...
+
+				if (elapsed_ms > log_debounce_threshold_ms) {
+					log_display_properties(sdl.draw.render_width_px,
+					                       sdl.draw.render_height_px,
+					                       sdl.video_mode,
+					                       sdl.rendering_backend);
+				}
+
+				last_resized = clock::now();
+			}	continue;
```
kcgen, line 3859: If there are duplicate resize events, can we get to the bottom of why that's happening?

Worst case: compare the existing size settings to the incoming request, and skip if it's asking for the same settings.

johnnovak: Not rapid resize. Just a single normal resize.

`SDL_WINDOWEVENT_RESIZED` is supposed to be fired _once_ once you've stopped resizing. Note `RESIZED` (paste tense) in the name.

...but on macOS (at least) it fires twice. It's either an SDL bug, or a macOS thing, or any number of other SDL/OS interactions, but it's firmly outside of our control.

johnnovak: We can't compare the incoming settings because at this point we've already adjusted all the `sdl.*` structs to the new settings.

johnnovak: So, I'm not getting into the whole complicated SDL maze now; I'd much rather put that effort into rewriting this from scratch. So, two options:

1) Leave the timer hack in; it does the job, and the user observable behaviour is what matter.
2) Remove the time hack, then we'll get the display settings logged twice when the window is resized 🤷🏻 

I'm fine with either 😄

kcgen: I don't like either 😅
I'd rather dedupe logically than make it time based. 
The logging function can hold static `last_`* args that are compared against the incoming args.
That way any duplicate logging will never happen, and likewise all unique sizes will always be logged.

johnnovak: Right, now I understood what you asked for 😄 Okay, I've rewritten it like that and it's still getting rid of the dupes.

src/gui/sdlmain.cpp
```diff
@@ -3842,20 +3817,42 @@ bool GFX_Events()
...
+				const auto new_width  = event.window.data1;
+				const auto new_height = event.window.data2;
+
+				if (prev_width != new_width ||
+				    prev_height != new_height) {
+					log_display_properties(sdl.draw.render_width_px,
+					                       sdl.draw.render_height_px,
+					                       sdl.video_mode,
+					                       sdl.rendering_backend);
+				}
```
kcgen, line 3851: If we switch VESA color depths, say 15-bit -> 16-bit -> 24-bit, then the mode will change, but the resolution will stay the same.

(And there are probably similar CGA -> EGA -> VGA switches where resolution can be similar?)

Probably safest just to hold and compare all four inputs.

johnnovak: No man, this is just the OS level width & height 😄 Just what the windowing system sends, so the new window dimensions 😄

kcgen: Ah! Thanks for clarifying :-)

kcgen (approved):
All good, @johnnovak !

src/gui/sdlmain.cpp
johnnovak: No problem :)

PR #3223, opened 2023-12-17
Remove PSP name from window titlebar

FeralChild64:
# Description

Follow-up of the discussion here: https://github.com/dosbox-staging/dosbox-staging/pull/3214

This removes the application executable name from the title bar; it is not useful and in many cases cryptic to the end user.

# Manual testing

Checked title bard with different values of `cycles` setting (`max`, `max 60%`, `max limit 10000`, `10000`), in debug and release builds.

# Checklist

I have:

### DIFF
--- a/src/dos/dos_execute.cpp
+++ b/src/dos/dos_execute.cpp
@@ -32,8 +32,6 @@
 #include "regs.h"
 #include "string_utils.h"
 
-const char * RunningProgram="DOSBOX";
-
 #ifdef _MSC_VER
 #pragma pack(1)
 #endif
@@ -65,23 +63,7 @@ struct EXE_Header {
 #define LOAD    1
 #define OVERLAY 3
 
-
-extern void GFX_RefreshTitle(const bool is_paused = false);
 extern void GFX_SetTitle(const int32_t cycles, const bool is_paused = false);
-
-void DOS_UpdatePSPName(void) {
-	DOS_MCB mcb(dos.psp()-1);
-	static char name[9];
-	mcb.GetFileName(name);
-	name[8] = 0;
-	if (!strlen(name)) strcpy(name,"DOSBOX");
-	for(Bitu i = 0;i < 8;i++) { //Don't put garbage in the title bar. Mac OS X doesn't like it
-		if (name[i] == 0) break;
-		if ( !isprint(*reinterpret_cast<unsigned char*>(&name[i])) ) name[i] = '?';
-	}
-	RunningProgram = name;
-	GFX_RefreshTitle();
-}
 
 void DOS_Terminate(uint16_t pspseg,bool tsr,uint8_t exitcode) {
 
@@ -123,7 +105,6 @@ void DOS_Terminate(uint16_t pspseg,bool tsr,uint8_t exitcode) {
 	real_writew(SegValue(ss),reg_sp+4,0x7202);
 	// Free memory owned by process
 	if (!tsr) DOS_FreeProcessMemory(pspseg);
-	DOS_UpdatePSPName();
 
 	if ((!(CPU_AutoDetermineMode>>CPU_AUTODETERMINE_SHIFT)) || (cpu.pmode)) return;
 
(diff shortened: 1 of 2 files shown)


johnnovak (approved):
Good riddance! 🎉 The title bar now looks a lot better, less crowded, and we have actually useful info there, not just some random useless characters that take up space 😎 Well done @FeralChild64 🚀

rderooy:
I have mixed feelings about it. Yes it is cryptic, but it has sometimes helped me to diagnose quickly some type of problems when there are multiple executables (that are sometimes not clearly named as such). An example is Thexder which has multiple executable for different machine options and you can see which is being run.

How about having the
```ini
[dosbox]
title = myGame
```
 option like dosbox-x has, such that you can at least set a custom title?

johnnovak:
> I have mixed feelings about it. Yes it is cryptic, but it has sometimes helped me to diagnose quickly some type of problems when there are multiple executables (that are sometimes not clearly named as such). An example is Thexder which has multiple executable for different machine options and you can see which is being run.
> 
> How about having the
> 
> ```ini
> [dosbox]
> title = myGame
> ```
> 
> option like dosbox-x has, such that you can at least set a custom title?

Maybe later we could add something like that. But in 99% of the scenarios, it's rather useless. I appreciate it can be useful in 1% of the cases though.

kcgen:
In every application that handles user data, I see the name of the data in the title bar:

### `IRC user@server` -> shows the users@server:


### `Firefox $1` -> shows the webpage name


`VSCode $1` -> shows the filename: 


### `Editor $1` -> shows the filename:


### `LibreOffice $1` -> shows the filename:


### `File manager $1` -> shows the directory:


### `Terminal $1` -> shows the binary name:


---

However now, `dosbox $1`doesn't show the executable name?

8.3 executable names /are/ shorter than modern long names, and yes, they /can/ be cryptic.

I don't see any of the above programs applying judgment calls to discard short or cryptic user data names arguments from their window titles though.

Likewise.. MS-DOS's twenty years of reign managed to describe every program and game executable in that 8.3 filename space, and yes: some are more meaningful than others.

I would suggest we show the executable just like every other program out there.

And if we want to let users provide a richer name, then OK - give them a CLI or conf flag to replace it.

`dosbox wc1.exe -richname "Wing Commander 1 v1.02"` (Or a conf setting).

FeralChild64:
Well, to be clear: I don't care which solution is chosen at this moment, as I have started working on making the title bar configurable to make everyone (mostly) happy.

If you wish to show program name, just revert the commit https://github.com/dosbox-staging/dosbox-staging/commit/f4f0d2ba4be95f135def2274442ec100bd1801c7

kcgen:
I don't want to cause more work - is it easier for you to work from the current state or reverted state?

FeralChild64:
> is it easier for you to work from the current state or reverted state?

Not much difference, I can rebase easily. And my rework won't be ready for 0.81.0, that's for sure.

kcgen:
> my rework won't be ready for 0.81.0, that's for sure.

Ok - that makes sense. I think we should match all the other applications out there until we have a more complete solution to provide a rich or user-replaced name. 

Will revert for now.

johnnovak:
Huh, never imagined this would cause such backslash, I thought everybody will be happy with the cleaner title bar 😄 




We are removing/changing lots of rubbish after being in there for 20 years, so that's not the best reasoning for our project especially 😆 

Well, probably better to revert then @FeralChild64 and do it in 0.82.0. In all honesty, it's a minor thing; it's just in my opinion polluting the title bar with actually useful info for most people. Plus an emulator is not really an editor or shell, but I get it that some power users can sometimes glean useful info from it... Anyway, not willing to die for it 😅 Make it configurable... then I can turn it off... or organise a deathmatch for the best defaults... or something :trollface:

johnnovak:
Another remark: showing the open file in the title bar is more of a norm in multi-tab programs, or in editors where you might have multiple files open, plus it's handy to know what you're overwriting, and they can display the "changed" asterisk after the name. 

Terminals are then just a bit weird; I never particularly liked the current process name flashing in the terminal's title bar like crazy, but as a tmux user, I normally don't see that 😄

kcgen:
So if it sounded like blacklash. 

Showing the currently passed-in user file is just how all applications work (and I agree: DOS names can be cryptic or useless :shrug: but not all of the time).

MIDI, MOD, MP3, and other emu's also show the open filename or ROM name in  their title bars too.

The passed-in executable can be as meaningful as the user likes (or maybe it's not), so it makes sense to default to this "standard behavior", but absolutely give them options to silence it or name it more fully, if they like.

johnnovak:
> it makes sense to default to this “standard behaviour”.

> But absolutely agree - give users the option to silence it or name it more richly, if they like.

Yeah, that's for the best 👍🏻 You know I'm just stirring the pot 😏 

Personally, I use fullscreen most of the time, plus I'll be able to set whatever I like with the new format string, so all good 😄

PR #3228, opened 2023-12-17
Fix CI hand-over of the MSVC binaries to MSYS2 packaging

kcgen:
# Description

A race condition in the MSVC CI artifact upload step meant we could upload the MSVC package with (or without) the debugger present.

If the MSVC debugger build job took longer than the non-debugger job, then the debugger would "miss the boat", leading to a warning-but-passing result on the MSYS2 side.

This PR splits out the debugger executable upload into its own step, so it's no longer dependent on the non-debugger package.

# Manual testing

I've inspected the packages and CI output, and it appears that everything is working - however we need to test setup.exe package to see if it has both the MSYS2 and MSVC executables.

@johnnovak, @shermp - can you check the installer?

# Checklist

I have:

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -182,7 +182,7 @@ jobs:
         id:    set_pkg_dir
         shell: bash
         run: |
-          echo "pkg_dir=dosbox-staging-windows-${{ matrix.conf.arch }}-${{ env.VERSION }}" >> $GITHUB_OUTPUT
+          echo "pkg_dir=dosbox-staging-windows-${{ matrix.conf.arch }}" >> $GITHUB_OUTPUT
 
       - name: Package standard build
         if: ${{ !matrix.conf.debugger }}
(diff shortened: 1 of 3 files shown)


johnnovak:
> @johnnovak, @shermp - can you check the installer?

I will later, yeah.

johnnovak:
@kcgen I've noticed you're still making changes to this. Let me know when it's final, then I'll test it.

kcgen:
Unfortunately it's rough going.

I've filled 10 pages worth of CI jobs throughout the day and am no closer to figuring out how to reliably making our msys2 packing job depend on the msvc workflow's outputs (and I've tried github's workflow_run, fountainhead's wait action, and lots of variations)😭

I'm going to close this and will have to revert the CI changes that dependapot and I added, and we'll go back to the v3 APIs.

johnnovak:
> Unfortunately it's rough going.
> 
> I've filled 10 pages worth of CI jobs throughout the day and am no closer to figuring out how to reliably making our msys2 packing job depend on the msvc workflow's outputs (and I've tried github's workflow_run, fountainhead's wait action, and lots of variations)😭
> 
> I'm going to close this and will have to revert the CI changes that dependapot added, and we'll go back to the v3 APIs.

That's the best. It's not like the v3 API will stop working soon. Then we can figure this out post 0.81.0 without stress. It's for the best not to mess with this before 22 Dec at least.

shermp:
I guess we could always just merge MSVC and MSYS2 into a single "Windows" workflow, which would allow easier dependency management.

kcgen:
> I guess we could always just merge MSVC and MSYS2 into a single "Windows" workflow, which would allow easier dependency management.

Yeah - that was my last option, but these workflows (the windows ones specifically, being more complex and larger) are already on the precipice of having runaway technical debt. 

Plus, being able to only debug them at a distance, in 20 minute increments, makes them very unwieldy - it's one of those tasks that just drains you, because there only outcome is just getting the damn thing back to a working baseline 😅

I think the next step for them is to go to shell scripts (or some 3rd party open-source python tool that can do this for us, and leave all this GitHub yaml behind.. well, except for the single hook to launch our system on the GitHub machine). 

It's the only way to move the development and testing locally and speed up the development iterations.

PR #3241, opened 2023-12-19
Log warning to console when not using full FPU 80-bit precision

kklobe:
# Description

Just a quick idea I had to print a warning to the console if we are running on a system with non-x87-conformant floating point:

CleanShot 2023-12-18 at 17 20 29@2x

# Manual testing

I've downloaded and tested the artifacts for macOS ARM, macOS x64, and MSYS2 Windows x64.

# Checklist

I have:

### DIFF
--- a/src/fpu/fpu.cpp
+++ b/src/fpu/fpu.cpp
@@ -651,6 +651,9 @@ void FPU_ESC7_Normal(Bitu rm) {
 
 
 void FPU_Init(Section*) {
+#if !C_FPU_X86
+	LOG_WARNING("FPU: Using reduced-precision floating-point emulation");
+#endif
 	FPU_FINIT();
 }
 


johnnovak (approved):
I like it!

kcgen (approved):
Love that tie into the build system flag; perfect :-)

src/fpu/fpu.cpp
```diff
@@ -651,6 +651,9 @@ void FPU_ESC7_Normal(Bitu rm) {
 
 
 void FPU_Init(Section*) {
+#if !C_FPU_X86
+	LOG_WARNING("FPU: Using reduced-precision floating-point emulation");
```
johnnovak, line 655: Would something like this be more informative to the user?

```
Using reduced 64-bit double precision floating-point emulation instead of full 80-bit extended precision
```

kcgen: Or `FPU using 64-bit instead of 80-bit precision`?

kcgen: If we want to convey that the  FPU shouldn't be trusted, then *"reduced precision"* is very nicely vague enough 😅!

If we say the exact number of bits to which we're accurate, then some "users" (ie: German airbus parts machine shops) might take it to the bank: *"64 bits is fine for our tolerance requirements; fire up the mill!"* ⚙️

johnnovak: Okay then 😄

kklobe: Yeah my main thought was, for people who know what the issue is, the vague description will be enough. For people who have no idea what the issue is, more descriptive text will probably be even more confusing. But I like the German indemnification protection explanation much better 😁

kcgen: I don't know if you guys remember, but someone from a massive German machine shop running a multi-stage, digital CNC system had contacted Wengier.

Their system used machining entry data generated from a DOS program, and (as expected), they were using DOSBox. Wengier was giving them urgent support because a recent bug impacted their production line).

kklobe: That is simultaneously joyous and terrifying.

MasterO2: > I don't know if you guys remember, but someone from a massive German machine shop running a multi-stage, digital CNC system had contacted Wengier.
> 
> Their system used machining entry data generated from a DOS program, and (as expected), they were using DOSBox. Wengier was giving them urgent support because a recent issue impacted their production line).

Wasn't Staging going to add a disclaimer about only using Staging for DOS games, and not DOS work productivity software?

kklobe: > Wasn't Staging going to add a disclaimer about only using Staging for DOS games, and not DOS work productivity software?

That should be covered by the NO WARRANTY clause in e.g. GPL 2.0 but maybe we should include it specifically in `./LICENSE`?

kcgen: > That should be covered by the NO WARRANTY clause in e.g. GPL 2.0

Yup, I think that covers it!

> but maybe we should include it specifically in `./LICENSE`?

No harm if you want to add it, but looking at history, DOSBox upstream and DOSBox-X has gone decade(s) with the same approach, and I think we're extra-clear with your FPU warning, too!

johnnovak: Yeal I'm planning to add a warning to the about page @MasterO2, and it will be in the manual too. Mainly about the numerical precision. Pentium class machines with ISA bus are still used widely in industrial applications, so sensible people use real hardware anyway.

PR #3242, opened 2023-12-19
Fix broken default build on macOS xcode

kcgen:
# Description

Default builds on macOS fail when xcode's Clang doesn't support the `-march=native` flag:

```
meson setup build
meson compile -C build
```

```text
[1/352] Compiling C object subprojects/zlib-ng/libzlib.a.p/compress.c.o
FAILED: subprojects/zlib-ng/libzlib.a.p/compress.c.o 
clang: error: the clang compiler does not support '-march=native'
```

Note that this doesn't affect the CI builds, which limit the SIMD instruction sets to NEON (on Arm) and SSE2+SSSE3 (on x86-64).

# Manual testing

Tested default builds on macOS, Linux, and Raspberry Pi (Debian Buster).

# Checklist

I have:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -243,6 +243,118 @@ if os_family_name == 'BSD'
     extra_link_flags += ['-Wl,-z,wxneeded']
 endif
 
+# SIMD assessment for optimized builds with zlib-ng
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# If the user asked for 'native' SIMD flags then we check if
+# their compiler supports the native arch type. If it doesn't,
+# we try to enable NEON (on Arm) and SSE2 and/or SSSE3 (on x86).
+#
+# Currently zlib-ng is the driving force behind these checks as
+# it provides a meaningful speedup in video and frame capture.
+#
+# Note that these build flags needs to be defined before adding
+# subpackages because the flags are used when compiling them.
+
+simd_summary = 'none'
+
+zlib_ng_options = get_option('use_zlib_ng')
+zlib_ng_wants_native = zlib_ng_options.contains('native')
+zlib_ng_is_native = zlib_ng_wants_native and cc.has_argument('-march=native')
+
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -243,6 +243,118 @@ if os_family_name == 'BSD'
...
+        ssse3_test = cc.run(
+            ssse3_test_code,
+            args: ssse3_cflag,
+            name: 'SSSE3 instruction set test',
+        )
+        if (ssse3_test.compiled() and ssse3_test.returncode() == 0)
+            zlib_ng_options += ['ssse3']
+            extra_flags += ssse3_cflag
+            simd_summary = 'SSE2 and SSSE3'
+            message('Enabling the SSSE3 instruction set')
```
johnnovak, line 352: Huh, this all seems rather complicated; isn't there a simpler way?

kcgen: Unfortunately not; the compiler will happily generate the instructions but execution on the target is the only way test them. These are tiny zero-dependency builds, so they fly by during the setup phase (and aren't tested again for all subsequent builds).

johnnovak: Ok, I'm far from being a build automation expert, so I'll trust you on this 😄 Seemed a bit like a maintenance burden (perhaps) but if you're saying it's necessary, I'm fine with it.

meson.build
```diff
@@ -243,6 +243,118 @@ if os_family_name == 'BSD'
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# If the user asked for 'native' SIMD flags then we check if
+# their compiler supports the native arch type. If it doesn't,
+# we try to enable NEON (on Arm) and SSE2 and/or SSSE3 (on x86).
+#
+# Currently zlib-ng is the driving force behind these checks as
+# it provides a meaningful speedup in video and frame capture.
+#
+# Note that these build flags needs to be defined before adding
+# subpackages because the flags are used when compiling them.
```
johnnovak, line 256: Nice explanatory comments 🏅

johnnovak (approved):
👍🏻

Issue #3226, opened 2023-12-17
Improved, more configurable title bar


FeralChild64:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

None

### Is your feature request related to a problem? Please describe.

This is a follow-up of the discussions from the PRs: https://github.com/dosbox-staging/dosbox-staging/pull/3214 and https://github.com/dosbox-staging/dosbox-staging/pull/3223. To please everyone (or at least just about everyone), we need a configurable title bar message. Ideas so far (both mine and from the discussion):

- there should be a runtime-changeable config file option (like `title =`) to override the default _DOSBox Staging_ or  _DOSBox Staging (debug build)_ string in the title bar with something arbitrary, for example a game name
- similarly to debug build, we could mention that the build is an alpha
- DOSBox version information should be optionally displayed
- ability to display running program name should be brought back - but should be optional (disabled by default)
- ability to display program path and extension (configurable)
- displaying CPU cycles configuration in the title bar should be optional (enabled by default) - it might look strange and cryptic to an ordinary user who just runs a pre-configured emulator
- title bar should indicate if audio/video recording is in progress
- DOS program names containing box drawing or national characters should be provided to host OS using UTF-8 encoding

Work-in-progress code: https://github.com/dosbox-staging/dosbox-staging/tree/fc/titlebar-3

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Nice idea. We could have something called `window_titlebar` set to `auto` by default, or to some simple format string.

Grounded0:
Just keep the old method available for the reasons pointed by @rderooy.

Windows 3.1x itself has 2 different operating modes: KRNL286 and KRNL386 protected modes and i need to know which one was picked due to operating factors.

FeralChild64:
@Grounded0 The change to remove program name was just reverted. For 0.82.0 I intend to make it configurable.

FeralChild64:
Draft implementation is ready in https://github.com/dosbox-staging/dosbox-staging/tree/fc/titlebar-3, not tested thoroughly yet (and there is a risk of regressions!).

Known problems:
- detection of alpha build not implemented yet
- changing any title bar setting at runtime causes the window to be destroyed and created again

johnnovak:
Will test it later, but this is for 0.82.0, to be clear.

A couple of things that I'd like to see addressed:

[...]

PR #3249, opened 2023-12-21
Improve title bar and make it configurable

FeralChild64:
# Description

The following title bar improvements have been implemented:

- title bar can display either DOSBox Staging application name, program name (segment name, file name with extension, or filename with extension and absolute path), or a custom string
- displays if the emulator is an alpha, optionally displays emulator version
- it is possible to disable the cycles setting part of the window title
- pause tag was moved to the beginning of the window title, in addition a recording tag was added, with optional animated indicator; tags are displayed at the beginning of the title to make sure it is visible on the task bar:
Frame1
Frame2
- removed the no-mouse mode title bar hint; it doesn't bring much useful information and look strangely in case of minimalist title bar configuration, for example just showing a full game name

All the new configuration options are changeable in runtime.

Example1

Example2

**Limitation:** if paging is enabled (like when Windows 3.1x is started in enhanced mode), the program name part only displays segment name (technical limitation).

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3226

# Manual testing

# Checklist

I have:


src/gui/title_bar.cpp
```diff
@@ -0,0 +1,586 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2020-2023  The DOSBox Staging Team
+ *  Copyright (C) 2002-2021  The DOSBox Team
+ *
```
kcgen, line 6: (Probably just a copy-and-paste)

If this is a new for this PR, suggest `2023-2023  The DOSBox Staging Team` only; the copyright lines are just facts about who authored the file, when it was created, and when it was last modified.

(same w/ the header file)

FeralChild64: OK, changed.

kcgen:
This is a really nice (and unexpected) enhancement @FeralChild64 !

FeralChild64:
@kcgen I only intended to fix the cycles display for `cycles = max limit <num_cycles>` - but then a discussion started...

johnnovak:
First run attempt. Needs more work 😎 

image

FeralChild64:
@johnnovak Debug build is fixed now, sorry.

FeralChild64:
Putting back to draft for now, as recent changes to `sdlmain.h` broke the implementation.

FeralChild64:
Breakage fixed, ready for review again.

johnnovak:
Tried this today, and it works well @FeralChild64. The `[PAUSE]` prefix is useful, and the blinking recording indicator looks cool for sure 😎

Before we'd get into the code-level specifics, I think a few high-level issues need addressing first:

- I think we have way too many options to customise the behaviour and the `[titlebar]` section is too specific and unwarranted. I would prefer if this went into a single `window_titlebar` setting in the `[sdl]` section and used simple tokens in the format string. E.g., "pause, rec+, dosprog, version, cycles, mouse" or something. You could invent some simple token names. Then you could freely select the order, and we wouldn't have all these new settings. In my example, `rec+` stands for the animated recording indicator, then `rec` could be the non-animated static variant, etc. Btw, I thought you were planning it this way; a single format string for the titlebar, so I was a bit surprised at this implementation...
- If you keep starting and then stopping the video or audio capture, the blinking recording indicator becomes erratic. The more start/stop cycles, the more erratic it gets. It starts blinking like crazy irregularly, at least on macOS. But the indicator always works correctly when pressing record for the first time after a restart.

FeralChild64:
@johnnovak Sure, I can change the setting style to keywords/token, but placing the options within `[sdl]` section might be tricky - I definitely want to be able to change the settings during runtime, so that a .BAT file which launches a game could also change the emulator title bar.

Putting this back to draft for now.

johnnovak:
> placing the options within `[sdl]` section might be tricky - I definitely want to be able to change the settings during runtime, so that a .BAT file which launches a game could also change the emulator title bar.

Yeah, it should be changeable at runtime.

I've done a fair bit of experimentation on how to achieve these runtime config changes (e.g. when I refactored the `MIXER` command and the `chorus`, `reverb`, and `crossfade` settings), so you can definitely put it in the `[sdl]` section and still achieve what you want.

The solutions relies on these two observations:

1) All settings in a config section are set during init, but only then.
2) After init, only a _single_ setting can be changed at a time. It's technically impossible to change more than a single setting at a time after init. In other words, even if a section contains 50 different settings, you can be certain that from the _second_ call of the init function onwards the diff is only a _single_ setting's value.

Based on this observation, the solution is this if you want to be super specific:

- Introduce a static boolean that defaults to false and gets set to true once the `[sdl]` section has been initialised for the first time after startup.
- Then you can safely assume that subsequent calls of the SDL_Init function (or whatever it's called) will be only triggered by changing a _single_ property. So you simply store `last_window_title` in another static var, then if the init flag is true (so we're past the initial init), you compare the current value of `window_title` to `last_window_title`. If they differ, `window_title` has been changed, so you do your thing, then exit the init function early.

I'm doing very similar things in `render.cpp` too because I need to quite specific which setting changes would trigger a full render re-init.

This approach will work fine even with the OSD stuff; you cannot change more than a single setting with the mouse / keyboard at a time 😄 Then if the situation changes, we can always refactor, but this solution will server you for a few more years, that's my estimation 😎

FeralChild64:
@johnnovak Yeah, this can always be handled in one way of another, I just don't like the spaghetti code which is already inside `sdlmain.cpp`, and which will now complicate even more. Config options moved to `[sdl]` section, now time to rethink the number of options and overall syntax.

johnnovak:
> I just don't like the spaghetti code which is already inside `sdlmain.cpp`, and which will now complicate even more.

Me neither, and I agree with you; we should not add more to the existing cruft unnecessarily. It's just in this specific case the logical place for `window_title` is in the `[sdl]` section, next to the three other `window_*` settings.

The way you've inserted this stuff into `sdlmain.cpp` is quite clean, so good job.

> Config options moved to `[sdl]` section, now time to rethink the number of options and overall syntax.

Great, looking forward for the new format.

FeralChild64:
@johnnovak Done. Ready for review again.

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+	        "                        name:     Program name, with file extension (default).\n"
+	        "                        path:     Name, extension, and full absolute path.\n"
+	        "                        segment:  Displays program memory segment name.\n"
+	        " 	                  Note: With some software (like Windows 3.1x in enhanced\n"
+	        "                      mode) it is impossible to recognize the full program name\n"
+	        "                      or path; in such case 'segment' value is used instead.\n"
+	        "  version[=<value>]:  Display DOSBox version information after emulator name.\n"
+	        "                      The <value> can be one of:\n"
+	        "                        simple:    Simple version format (default).\n"
+	        "                        detailed:  Include GIT hash, if available.");
```
johnnovak, line 577: These params should be listed in the order they appear in in the titlebar.

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+{
+	constexpr auto always = Property::Changeable::Always;
+
+	Prop_string* prop_str = nullptr;
+
+	prop_str = secprop.Add_string("titlebar_content", always,
+	                              "animation cycles dosbox program=name");
+	prop_str->Set_help(
+	        "Space separated list of information to be displayed on the window title bar\n"
+	        "('animation cycles dosbox program=name' by default). Possible information are:\n"
```
johnnovak, line 561: The params in the default setting should be listed in the order they appear in in the titlebar.

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+	        "                        segment:  Displays program memory segment name.\n"
+	        " 	                  Note: With some software (like Windows 3.1x in enhanced\n"
+	        "                      mode) it is impossible to recognize the full program name\n"
+	        "                      or path; in such case 'segment' value is used instead.\n"
+	        "  version[=<value>]:  Display DOSBox version information after emulator name.\n"
+	        "                      The <value> can be one of:\n"
+	        "                        simple:    Simple version format (default).\n"
+	        "                        detailed:  Include GIT hash, if available.");
+
+	prop_str = secprop.Add_string("titlebar_text", always, "");
```
johnnovak, line 579: Technically, this should be `titlebar_custom_program_name` or something. The current name is super confusing—"text" is like "data" or "value", it can mean literally anything 😄

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+			continue;
+		}
+
+		if (iequals(setting, "version=detailed")) {
+			check_not_set_version();
+			config.version = VersionDisplay::Detailed;
+			continue;
+		}
+
+		LOG_WARNING("SDL: Unknown setting '%s' in 'titlebar_content'",
```
johnnovak, line 543: The new consistent pattern I introduced a while ago (I reworked most if not all such error messages) would be this:

  `SDL: Invalid 'titlebar_content' setting '<setting-value>', using
  '<the-default-value'`

  Do a `git grep ', using ` for tons of examples.

johnnovak: When you change a config property programmatically (e.g., when defaulting because the user provided an invalid setting), you should also sync the config property to the changed setting. Otherwise `config -h <settingname>` would still display the invalid value!

This will be a major problem for the OSD too. I've started doing such syncing in my new code; check out `sync_crossfeed_setting` and `sync_reverb_setting` in `mixer.cpp`, or search for `SetValue` in `render.cpp`.

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+	trim(config.titlebar_text);
+}
+
+void TITLEBAR_AddConfig(Section_prop &secprop)
+{
+	constexpr auto always = Property::Changeable::Always;
+
+	Prop_string* prop_str = nullptr;
+
+	prop_str = secprop.Add_string("titlebar_content", always,
```
johnnovak, line 557: Can we include include `window` in the setting's name? Can this just be `window_titlebar`?

include/video.h
```diff
@@ -401,9 +401,17 @@ void GFX_EndUpdate( const uint16_t *changedLines );
...
 void GFX_RegenerateWindow(Section *sec);
 
+void GFX_RefreshTitle();
+void GFX_RefreshAnimatedTitle();
+void GFX_NotifyBooting();
+void GFX_SetAudioCaptureMark(const bool is_capturing);
+void GFX_SetVideoCaptureMark(const bool is_capturing);
+void GFX_SetProgramName(const std::string& segment_name,
+                        const std::string& canonical_name);
+void GFX_SetCycles(const int32_t cycles);
```
johnnovak, line 411: Can we introduce some light decoupling here? For now, it's just mostly about renaming these functions, e.g., from ` GFX_SetAudioCaptureMark` to `GFX_NotifyAudioCaptureStatus`. Then `GFX_NotifyAudioCaptureStatus` would call another function called `GFX_SetTitlebarAudioCaptureMark`.

The reason I'm proposing this is because with the OSD stuff `GFX_NotifyAudioCaptureStatus` will then call a second method called `OSD_SetAudioCaptureMark` or something, and if you do it this way now, I won't need to rework it later 😄

src/cpu/cpu.cpp
```diff
@@ -1596,7 +1594,7 @@ void CPU_SET_CRX(Bitu cr,Bitu value) {
 					CPU_CycleLeft=0;
 					CPU_Cycles=0;
 					CPU_OldCycleMax=CPU_CycleMax;
-				        GFX_SetTitle(CPU_CyclePercUsed);
+				        GFX_SetCycles(CPU_CyclePercUsed);
```
johnnovak, line 1597: Similarly, `GFX_NotifyCyclesChanged`.

src/dos/dos_execute.cpp
```diff
@@ -438,8 +478,16 @@ bool DOS_Execute(char * name,PhysPt block_pt,uint8_t flags) {
...
-		SETVER::OverrideVersion(name, newpsp);
+
+		char canonical_name[DOS_PATHLENGTH];
+		if (!DOS_Canonicalize(name, canonical_name)) {
+			assert(false);
+		} else {
+			/* If needed, override reported DOS version */
+			SETVER::OverrideVersion(canonical_name, newpsp);
+			/* Store canonical name for display/debug purposes */
+			add_canonical_name(dos.psp(), canonical_name);
```
johnnovak, line 489: Minor: line comments? (`//`)

src/dos/dos_keyboard_layout.cpp
```diff
@@ -38,6 +38,16 @@ using sv = std::string_view;
 #include "setup.h"
 #include "string_utils.h"
 
+extern void DOS_UpdateCurrentProgramName();
+
+static void notify_new_code_page(const bool reason_keyboard_layout = false)
```
johnnovak, line 43: I have no idea what `reason_keyboard_layout` means.

src/dos/dos_keyboard_layout.cpp
```diff
@@ -960,11 +970,7 @@ KeyboardErrorCode KeyboardLayout::ReadCodePageFile(const char *requested_cp_file
 			}
 			INT10_SetupRomMemoryChecksum();
 
-			// re-create country information and [autoexec] section
-			// to match new code page
-			DOS_RefreshCountryInfo();
-			AUTOEXEC_NotifyNewCodePage();
-
+			notify_new_code_page();
```
johnnovak, line 973: `notify_code_page_changed`?

src/gui/meson.build
```diff
@@ -4,6 +4,7 @@ libgui_sources = files(
     'sdl_mapper.cpp',
     'sdlmain.cpp',
     'shader_manager.cpp',
+    'title_bar.cpp',
```
johnnovak, line 7: I'm fine with "titlebar" (one word) which you seem to be using consistently everywhere. But then the file should be called `titlebar.cpp`.

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+#include <atomic>
+
+CHECK_NARROWING();
+
+// ***************************************************************************
+// Datat types and storage
+// ***************************************************************************
+
+enum class ProgramDisplay : uint8_t { None, Name, Path, Segment };
+enum class VersionDisplay : uint8_t { None, Simple, Detailed };
```
johnnovak, line 44: Minor: Personally, I'd drop the size specification unless we care about the size. Most of the time, why should we care? Let the compiler do whatever (probably use the widest type, but who cares).

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+	std::string title_frame_1 = {};
+	std::string title_frame_2 = {};
+
+	SDL_TimerID timer_id = {};
+
+	std::atomic<bool> is_animation_running      = false;
+	std::atomic<bool> animation_phase_alternate = false;
+} state = {};
+
+// Pre-calculated strings, to save some time when generating new title bar text
```
johnnovak, line 76: Haha, "save some time"... Watch me when the OSD will regenerate the UI at 60-70 FPS using Bezier curves etc 😎

src/gui/title_bar.cpp
```diff
@@ -0,0 +1,600 @@
...
+		state.hint_mouse = std::move(hint);
+		GFX_RefreshTitle();
+	}
+}
+
+// ***************************************************************************
+// Lifecycle
+// ***************************************************************************
+
+void TITLEBAR_ReadConfig(Section* section)
```
johnnovak, line 439: If you'd return a struct with the configured options (`std::optional<TitlebarConfig>`), then this function would become testable. I recommend doing so.

src/gui/title_bar.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#ifndef DOSBOX_TITLE_BAR_H
+#define DOSBOX_TITLE_BAR_H
+
+#include "setup.h"
+
+void TITLEBAR_AddMessages();
+void TITLEBAR_AddConfig(Section_prop &secprop);
+void TITLEBAR_ReadConfig(Section* section);
```
johnnovak, line 28: `std::optional<TitlebarConfig> TITLEBAR_ParseConfig(const std::string& config)` would be better, and then you could add unit tests (once you've condensed the two settings into a single format string, that is).

johnnovak:
This is looking quite good now @FeralChild64 and it works really well. Great stuff! 🚀 

In addition to my code-level comments, I have some suggestions for improving the user experience, plus I found a bug with the animations (at least on macOS; see the recording below).

Let's start with my high-level UX improvement suggestions:

- Can we introduce a simplified "mouse capture" option too? The default is good for beginners, but I'd prefer to have nothing displayed when uncaptured and only "mouse captured" in captured state (or something even shorter).

- The coupling between`titlebar_text` and the rest of the options is quite confusing; it took me a good 10 minutes of experimentation to figure it out how to use it, and even then I found it rather clunky. I get it that complex features are inherently complex, but this should be a lot easier. I'm not 100% sure why we need custom "program" strings either, so might as well remove this feature? If you wanna keep it, I think it's best to integrate it into the format string (see my examples below).

- Similarly, the coupling between `dosbox` and `version` is non-intuitive and a bit weird. I highly recommend to remove _all_ coupling between the format specifiers—keep it simple and stupid and easy to comprehend 😎 The current behaviour is a bit confusing and not that flexible; it would be better to completely separate `dosbox` and `version`.

  - With just `version`, I get neither the DOSBox string, neither the version
  - With just `dosbox`, I get `DOSBox Staging (alpha, debug build)`
  - With `dosbox version`, I get `DOSBox Staging 0.82.0-alpha (debug build)`
  - With `dosbox version=detailed`, I get `DOSBox Staging 0.82.0-alpha-242-g70816d534 (debug build)`

- Could we add an "audio muted" indicator as well?

--- 

Here are some ideas how the decoupled and simplified settings could work:

- `program version mouse=short` -> `DN.PRG - 0.82.0-alpha (debug build) - mouse captured`
- `dosbox version=detailed mouse=short` -> `DOSBox Staging 0.82.0-alpha-242-g70816d534 (debug build) - mouse captured`
- `program="custom Yikes! 1234" dosbox mouse=short` (if you want to support the weird custom prefix string feature) -> `Yikes! 1234 - DOSBox Staging - mouse captured`

---

And here's the flaky recording indicator. As you can see, it blinks as it should when pressing record for the first time. The subsequent recording stop & start actions will make the blinking it increasingly erratic. Maybe the old timers are still active, dunno.

https://github.com/dosbox-staging/dosbox-staging/assets/698770/ad05797b-c1d7-45c1-8704-b0930c137765

src/gui/title_bar.cpp
FeralChild64: > But I wonder why this setting is needed. What's the use case, who would want this and why?

Me, to make testing less confusing:) It is helpful to see whether I'm playing the pure game, or game with mods, or which test installation of Windows is currently running - I am putting all this information into the title bar, with .BAT scripts. Therefore, I definitely don't want the `"` character as a part of the syntax, as this becomes confusing when you write a batch file.

Also, with prepackaged games, it is much nicer to display the name of the game in the title bar, and not something cryptic.

I'll think about the syntax, for now I have no better idea than a separate config option.

FeralChild64:
> Can we introduce a simplified "mouse capture" option too?

Sure, no problem.

> Could we add an "audio muted" indicator as well?

I think I can use [U+1F507](https://www.compart.com/en/unicode/U+1F507) (Speaker with Cancellation Stroke) symbol for this, we'll see how it works.

> flaky recording indicator

I can see it too on my system, looks like SDL timer is doing something I don't really understand. I'll investigate this.

----

Putting this back to draft. I don't have that much time currently, it will take some time to handle the remarks.

Grounded0:
Audio muted indicator is a nice idea. 👍

src/gui/title_bar.cpp
FeralChild64: Changed.

src/gui/title_bar.cpp
FeralChild64: Changed.

FeralChild64:
@Grounded0 For sure only the "audio muted" status will be displayed. I didn't think about, as I use KDE mixer to mute the given application (or to switch it between headphones and speakers), and this results in KDE displaying the muted symbol in the taskbar.

src/gui/meson.build
FeralChild64: Ok, file name changed.

src/gui/title_bar.cpp
johnnovak: Thanks for the clarification. You're quite particular about the contents of your titlebar 😄

src/gui/title_bar.h
johnnovak: But if you simplify it and parsing becomes trivial, you might not need the tests that much.

FeralChild64:
Animation timer problem fixed. My mistake, the result of `SDL_AddTimer` wasn't stored in `state.timer_id`, as it should be.

src/gui/title_bar.cpp
FeralChild64: OK, changed.

johnnovak:
> Animation timer problem fixed. My mistake, the result of `SDL_AddTimer` wasn't stored in `state.timer_id`, as it should be.
> 
> [edit] Also decoupled version from DOSBox application name. [edit] Mouse hint is now configurable.

Cool, let me know when it's ready for final review as I can't see easily what has changed.

Btw, TIL `[skip ci]` 😆

include/video.h
FeralChild64: OK, changed.

FeralChild64:
I try to skip actions if I don't need them, to lower the chance that GitHub gets pissed off and starts charging us for the computing power :)

Mute mark implemented, moved the custom game title to `window_titlebar` - now it's time to clean up and fully test all this code. I'll let you know once I'm ready for the review.

johnnovak:
> I try to skip actions if I don't need them, to lower the chance that GitHub gets pissed off and starts charging us for the computing power :)

Citation needed, has that ever happened to any project you know of? I don't think they give a shit.

> Mute mark implemented, moved the custom game title to `window_titlebar` - now it's time to clean up and fully test all this code. I'll let you know once I'm ready for the review.

Kewl 😄 Nice one, looking forward to it.

src/cpu/cpu.cpp
FeralChild64: Changed.

src/dos/dos_execute.cpp
FeralChild64: Changed.

src/dos/dos_keyboard_layout.cpp
FeralChild64: Renamed.

src/dos/dos_keyboard_layout.cpp
FeralChild64: Renamed variable to `keyboard_layout_changed`, hopefully it is more clear now.

src/gui/title_bar.cpp
FeralChild64: OK, submodule interface changed to be testable.

src/gui/title_bar.h
FeralChild64: Unfortunately, unit tests are not easy to add. Including the `sdlmain.h` leads to a lot of undefined symbols - a lot of stubbing would be needed. But I have reworked the configuration parsing, so that it just takes a string.

src/gui/title_bar.cpp
FeralChild64: OK, changed warning messages and added the synchronization code.

FeralChild64:
@johnnovak Ready for the review again.

src/gui/sdl_mapper.cpp
```diff
@@ -3247,13 +3247,23 @@ static void MAPPER_Destroy(Section *sec) {
...
 	const auto property = section->Get_path("mapperfile");
+	
+	// Filter out unneeded calls
+	static bool first_time = true;
+	static std::string old_mapperfile_value = {};
+	if (!first_time && mapperfile_value == old_mapperfile_value) {
+		return;
+	}
+	first_time = false;
+	old_mapperfile_value = mapperfile_value;
```
johnnovak, line 3262: Is this some other unrelated fix?

src/gui/titlebar.h
```diff
@@ -0,0 +1,30 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOSBOX_TITLE_BAR_H
+#define DOSBOX_TITLE_BAR_H
```
johnnovak, line 22: TITLEBAR to be consistent with the file name.

src/misc/programs.cpp
```diff
@@ -775,6 +780,7 @@ void CONFIG::Run(void)
 					         sec_name);
 					return;
 				}
+				// XXX
```
johnnovak, line 783: ?

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+CHECK_NARROWING();
+
+// ***************************************************************************
+// Datat types and storage
+// ***************************************************************************
+
+static struct TitlebarConfig {
+	enum class ProgramDisplay { NotSet, Default, None, Name, Path, Segment, Custom };
+	enum class VersionDisplay { NotSet, Default, None, Simple, Detailed };
+	enum class MouseHintDisplay { NotSet, Default, None, Short, Full };
```
johnnovak, line 46: Do we really need to differentiate between NotSet and None? If yes, a better solution would be not to encode the unset/uninitialised status in the enum values but use `std::optional`.

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+
+	bool no_animated_rec = false;
```
johnnovak, line 48: No double negatives, please 😅 

```cpp
bool animated_rec = true
```

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	bool show_cycles     = false;
+	bool show_dosbox     = false;
+
+	ProgramDisplay program = ProgramDisplay::NotSet;
+	VersionDisplay version = VersionDisplay::NotSet;
+	MouseHintDisplay mouse = MouseHintDisplay::NotSet;
```
johnnovak, line 54: `std::optional<MouseHintDisplay> mouse = {}` etc.

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+// leads to unpleasant effect when it suddenly appears on the title bar.
+// Similarly, we do not use U+1F507 (Speaker With Cancellation Stroke) symbol.
+static const std::string MediumWhiteCircle = "\xe2\x9a\xaa"; // U+26AA
+static const std::string MediumBlackCircle = "\xe2\x9a\xab"; // U+26AB
+
+static const std::string Separator = " - ";
+static const std::string BeginTag  = "[";
+static const std::string EndTag    = "] ";
+
+static const std::string RecordingMarkGeneric = "REC";
```
johnnovak, line 103: Minor: RecordingMarkText ?

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+static const std::string RecordingMarkFrame1 = MediumBlackCircle + RecordingMarkGeneric;
+static const std::string RecordingMarkFrame2 = MediumWhiteCircle + RecordingMarkGeneric;
+
+// ***************************************************************************
+// Title bar rendering
```
johnnovak, line 108: Minor: titlebar for consistency

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+}
+
+static uint32_t animation_tick(uint32_t /* interval */, void* /* name */)
+{
+	SDL_Event event = {};
+	event.user.type = enum_val(SDL_DosBoxEvents::RefreshAnimatedTitle);
+	event.user.type += sdl.start_event_id;
+
+	// We are outside of the main thread; we can't update the window title
+	// here, SDL does not like it - we have to go through the event queue
```
johnnovak, line 126: Nice explanatory comment for the next guy 🏅

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	} else {
+		cycles = Separator + format_str("max %d%%", state.num_cycles);
+	}
+
+	return cycles + " " + MSG_GetRaw("TITLEBAR_CYCLES_MS");
+}
+
+static std::string get_mouse_hint_simple()
+{
+	// We are using 'MSG_GetRaw' here as we want messages to stay as UTF-8
```
johnnovak, line 237: 🎖️

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+static void maybe_add_muted_mark(std::string& title_str)
+{
+	// Do not add 'mute' tag if emulator is paused
+	if (state.is_audio_muted && !sdl.is_paused) {
+		title_str = BeginTag + MSG_GetRaw("TITLEBAR_MUTED") + EndTag +
+		            title_str;
+	}
+}
+
+static void maybe_add_rec_pause_mark(std::string& title_str)
```
johnnovak, line 304: Maybe spell out "record"?

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	}
+}
+
+void GFX_NotifyProgramName(const std::string& segment_name,
+                           const std::string& canonical_name)
+{
+	std::string segment_name_dos = segment_name;
+
+	// Segment name might contain just about any character - adapt it
+	for (size_t i = 0; i < segment_name_dos.size(); ++i) {
```
johnnovak, line 424: Nice 👍🏻

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	        "                      'titlebarbar_text'. The <value> can be one of:\n"
+	        "                        none:     Do not display program name.\n"
+	        "                        name:     Program name, with file extension (default).\n"
+	        "                        path:     Name, extension, and full absolute path.\n"
+	        "                        segment:  Displays program memory segment name.\n"
+	        "                      or you can provide a custom name in one of the forms:\n"
+	        "                        =\"Game\", ='Game', =(Game), =<Game> or =[Game]\n"
+	        "                      Note: With some software (like Windows 3.1x in enhanced\n"
+	        "                      mode) it is impossible to recognize the full program name\n"
+	        "                      or path; in such case 'segment' value is used instead.\n"
```
johnnovak, line 845: typo: `in such cases,` (plural, comma)

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	        "  dosbox:             DOSBox emulator name.\n"
```
johnnovak, line 846: DOSBox Staging

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	        "  version[=<value>]:  Display DOSBox version information after emulator name.\n"
+	        "                      The <value> can be one of:\n"
```
johnnovak, line 848: Drop `The`

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	        "                        none:      Do not display DOSBox version.\n"
+	        "                        simple:    Simple version format (default).\n"
+	        "                        detailed:  Include GIT hash, if available.\n"
```
johnnovak, line 851: Should be `Git`?

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	        "  cycles:             Show CPU cycles setting.\n"
+	        "  mouse[=<value>]:    Mouse capturing hint verbosity level:\n"
+	        "                        none:      Do not display any mouse hints.\n"
+	        "                        short:     Only display if mouse is captured.\n"
+	        "                        full:      Display verbose information how to capture or\n"
```
johnnovak, line 856: "on how to"

johnnovak:
This is looking very good now @FeralChild64, it works well and all major issues are gone. 🎉  We're almost at the finish line! 🥳  As people say, the first 90% is always the easy part, but finishing the remaining 90% can take a long time... 😎 

Great job improving the consistent handling of parameters, this is now a lot easier to use than before. The erratic flashing is gone now for the recording indicator as well, and the updated settings are always synced back to the config setting. Very nice! 🥇 

Apart from my rather minor inline comments, here are a few things that I think could be further improved with the settings.

---

## `dosbox` param

`dosbox` and `program` are still somewhat intertwined, which I think is okay. However, it needs to be better documented, and we could make things even clearer by changing `dosbox` slightly.

Here's how it behaves:

#### `dosbox` not specified

If not specified, "DOSBox Staging" is only displayed if there's no program running.

`window_titlebar = version=detailed mouse=none`

DOS prompt:
image

When running a program:
image

#### `dosbox` specified 

If specified, "DOSBox Staging" is always displayed.

`window_titlebar = dosbox version=detailed mouse=none`

DOS prompt:
image

When running a program:
image

### Suggestion

```
dosbox[=<value>]:  Displays 'DOSBox Staging' in the title bar.
                   <value> can be one of:
                     always:     Always display 'DOSBox Staging'.
                     auto:       Only display it if no program is running or
                                 'program=none' is set (default).
```

## `animation` & `cycles` params

I think we should accept `animation=on` too, which should be the default.

Then to make `cycles` consistent with the other settings, I think it should also accept `on` and `off`. It's just a bit weird that `animation` defaults to enabled, but you can only disable it with `animation=off`. Then `cycles` is only enabled by including `cycles`, so it defaults to `off`... but there are no `on` and `off` settings for it.

So allowing both `animation=on|off` and `cycles=on|off` would make this consistent.

## Edge case

`window_titlebar = version=none mouse=none`

DOS prompt:
image

When running a program:
image

Just lose the dash in this case.

## General

The wrapping is off for `animation`, and it would look better if `program`/`segment` were indented to the red vertical line.

image

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	constexpr auto always = Property::Changeable::Always;
+
+	Prop_string* prop_str = nullptr;
+
+	prop_str = secprop.Add_string("window_titlebar",
+	                              always,
+	                              "program=name dosbox cycles mouse=full");
+	prop_str->Set_help(
+	        "Space separated list of information to be displayed on the window title bar\n"
+	        "('program=name dosbox cycles mouse=full' by default).\n"
```
johnnovak, line 831: This is implied, but I think we should state it explicitly: 

`If a parameter is not specified, its default value is used.`

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,878 @@
...
+	        "                        name:     Program name, with file extension (default).\n"
+	        "                        path:     Name, extension, and full absolute path.\n"
+	        "                        segment:  Displays program memory segment name.\n"
+	        "                      or you can provide a custom name in one of the forms:\n"
+	        "                        =\"Game\", ='Game', =(Game), =<Game> or =[Game]\n"
+	        "                      Note: With some software (like Windows 3.1x in enhanced\n"
+	        "                      mode) it is impossible to recognize the full program name\n"
+	        "                      or path; in such case 'segment' value is used instead.\n"
+	        "  dosbox:             DOSBox emulator name.\n"
+	        "  version[=<value>]:  Display DOSBox version information after emulator name.\n"
```
johnnovak, line 847: Just `Display DOSBox version information.` (If `dosbox` is not set and `program` is something else than `none`, the version will come after the DOS program/path/segment name.)

Grounded0:
Looks great and i only have one test case to propose for this:

https://github.com/dosbox-staging/dosbox-staging/wiki/Frequently-asked-questions#how-to-run-your-command-line-interpreter-of-choice-exclusively

Or test couple third party shells in permanent mode and check does everything work as it should.

src/gui/titlebar.cpp
FeralChild64: Changed.

src/gui/titlebar.cpp
FeralChild64: OK, dropped.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.h
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: With `std::optional<ProgramDisplay> program`, we would need to have additional check before `switch (config.program)` - IMHO not worth it.

FeralChild64:
@Grounded0 3rd party shells do work. What just can't fully work is Windows 3.x in Enhanced mode or a custom OS booted.

src/gui/titlebar.cpp
FeralChild64: OK, changed. This was a remain of the old configuration system.

src/gui/titlebar.cpp
FeralChild64: OK, added.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/sdl_mapper.cpp
FeralChild64: I needed to prevent the mapper settings from being re-evaluated, as this was causing the DOSBox window to be recreated. Execute `config -set mapperfile=foobar` to see what I mean.

src/misc/programs.cpp
FeralChild64: Oops, missing implementation. The change above makes `config -get window_titlebar` display sane values if custom program name contains national characters. The missing implementation here should handle `config -get sdl window_titlebar` command. Fixed.

src/gui/titlebar.cpp
FeralChild64: There are two reasons:
- to warn about setting provided more than once
- in case of syntax error, to create a fixed string in `sync_config` which is similar to what the user originally provided (whether there was `mouse`, `mouse=full`, or nothing at all in the original string - I would like to preserve what the user submitted, despite every variant producing same results)

The second case can be handled in an alternative way, by storing the actual substring the user provided. Originally, I was considering this approach too - if you think it's better, I can change the implementation.
`std::optional` is not really suitable, as it would have turned a simple `switch (value)` into something more complex.

FeralChild64: Update: I'm going to store the original string. Having the animation set to enabled / set to disabled / not set at all / omitted makes some if's too complicated and the overall code would be too mistake-prone.

src/misc/programs.cpp
johnnovak: Glad to hear our code reviews serve their intended purpose 😄

src/gui/titlebar.cpp
johnnovak: Yeah I think storing the original string is the better solution @FeralChild64.

I did consider that using optional might overcomplicate the code. Now the switch block are exhaustive, so yeah, ifs would cause unnecesary complexity. As these enums are internal only, this is fine.

src/gui/titlebar.cpp
johnnovak: Yeah, on a second thought, you're right.

src/gui/titlebar.cpp
FeralChild64: Original settings are now stored in the strings.

FeralChild64:
@johnnovak OK, ready for the next review. I think I have handled all the remarks.

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+
+	Prop_string* prop_str = nullptr;
+
+	prop_str = secprop.Add_string("window_titlebar",
+	                              always,
+	                              "program=name dosbox=auto cycles=on mouse=full");
+	prop_str->Set_help(
+	        "Space separated list of information to be displayed on the window titlebar\n"
+	        "('program=name dosbox=auto cycles=on mouse=full' by default). If a parameter\n"
+	        "is not specified specified, its default value is used.\n"
```
johnnovak, line 755: You wrote "specified" twice.

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+	        "Possible information to display are:\n"
+	        "  animation=<value>:  If set to 'on' (default), animate the audio/video\n"
+	        "                      recording mark. Set to 'off' to disable animation; this\n"
+	        "                      is usefull if your screen font produces weird results.\n"
```
johnnovak, line 759: Typo: useful

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+	        "  program=<value>:    Display the name of a running program.\n"
```
johnnovak, line 760: "the running program"

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+	        "                      <value> can be one of:\n"
+	        "                        none:     Do not display program name.\n"
+	        "                        name:     Program name, with file extension (default).\n"
+	        "                        path:     Name, extension, and full absolute path.\n"
+	        "                        segment:  Displays program memory segment name.\n"
+	        "                        'Title':  Custom game title. Alternatively, you can use\n"
```
johnnovak, line 766: Maybe just say "Custom name."

Name because "program name" was used so far, not "title". And you can use it for any program, not just games 😄

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, fixed.

src/gui/titlebar.cpp
FeralChild64: OK, fixed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

weirddan455:
>  It seems one of the jobs, Linux builds / Release build (push), is not functioning anymore:

Just re-run it.  It's just a network error.  It passed on my PR that I pushed to a few minutes ago.

FeralChild64:
Already tried this, didn't work. Let's try again...

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+#if !defined(NDEBUG)
+	return true;
+#else
+	return false;
+#endif
+}
+
+// The U+23FA (Black Circle For Record) symbol would be more suitable, but with
+// some fonts it is larger than Latin alphabet symbols - this (at least on KDE)
+// leads to unpleasant effect when it suddenly appears on the titlebar.
```
johnnovak, line 115: "in the titlebar"

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+{
+	constexpr auto always = Property::Changeable::Always;
+
+	Prop_string* prop_str = nullptr;
+
+	prop_str = secprop.Add_string("window_titlebar",
+	                              always,
+	                              "program=name dosbox=auto cycles=on mouse=full");
+	prop_str->Set_help(
+	        "Space separated list of information to be displayed on the window titlebar\n"
```
johnnovak, line 753: "in the window's titlebar"

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+	        "Possible information to display are:\n"
+	        "  animation=<value>:  If set to 'on' (default), animate the audio/video\n"
+	        "                      recording mark. Set to 'off' to disable animation; this\n"
+	        "                      is useful if your screen font produces weird results.\n"
+	        "  program=<value>:    Display the name of the running program.\n"
+	        "                      <value> can be one of:\n"
+	        "                        none:     Do not display program name.\n"
+	        "                        name:     Program name, with file extension (default).\n"
+	        "                        path:     Name, extension, and full absolute path.\n"
+	        "                        segment:  Displays program memory segment name.\n"
```
johnnovak, line 765: "Display" (lose the "s")

src/gui/titlebar.cpp
```diff
@@ -0,0 +1,808 @@
...
+	        "                      name or path; in such cases 'segment' is used instead.\n"
+	        "  dosbox=<value>:     Display 'DOSBox Staging' in the title bar.\n"
+	        "                      <value> can be one of:\n"
+	        "                        always:   Always display 'DOSBox Staging'.\n"
+	        "                        auto:     Only display it if no program is running or\n"
+	        "                                  'program=none' is set (default).\n"
+	        "  version=<value>:    Display DOSBox version information.\n"
+	        "                      <value> can be one of:\n"
+	        "                         none:      Do not display DOSBox version (default).\n"
+	        "                         simple:    Simple version.\n"
```
johnnovak, line 779: minor: "Simple version information."

johnnovak:
This works perfectly now @FeralChild64, great job! All issues are fixed now, and top-notch code quality too, as always 🚀 

One last minor idea, but it's entirely optional: you might wanna also accept `off` instead of the `none` values, and then document them as `none/off`. It's a bit better because now for examples `cycles` wants `off`, but `program` wants `none`, etc. Anyway, it's really minor, up to you.

`pcspeaker` already supports such a thing:

```ini
#           pcspeaker: PC speaker emulation model:
#                        discrete:  Waveform is created using discrete steps (default).
#                                   Works well for games that use RealSound-type effects.
#                        impulse:   Waveform is created using sinc impulses.
#                                   Recommended for square-wave games, like Commander Keen.
#                                   While improving accuracy, it is more CPU intensive.
#                        none/off:  Don't use the PC Speaker.
```

johnnovak (approved):

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

FeralChild64:
@johnnovak OK, extended the syntax.

PR #3250, opened 2023-12-21
Render video lines one-by-one even in high resolution modes

### LINKED ISSUE #3248: Microsoft Space Simulator (1994)

81.0a

Windows 11

_No response_

### Relevant hardware info

_No response_

This game seems to suffer from two other issues that still haunt Staging and most other Dosbox forks.

1. Its a flickering mess from the moment you are in the cockpit. As per #3244, #2905 & #2950 the copy of Daum included in eXoDOS fixes it.
2. There is a speed issue as per #2751. At around 50k you can move around fine (maybe even a bit slowly) while the Earth is in view. The second Earth is of screen you flick around in the blink of an eye to have the Earth back on screen

Explain how to reproduce
1. Play the game in Staging 81.0a
2.
3.

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
# Description

Previously, high resolution (800×600 and up) video modes were rendered out from the emulated video card as whole frames. That is, we waited until vblank at which point the entire block of the emulated video card's memory (as indicated by the game) was rendered out by the emulated card.

The _rendering out_ process emulates the video card's DAC pipeline by reading a row from memory (written by the DOS application) and scanning it out to the CRT. This is where line and row doubling may happen, such as when VGA 320×200 modes are doubled at the analog signal level up to 640×400 which hits the CRT.

(To clarify for readers: the above is all still in hardware emulation - the vga_draw* and render* code, and comes before any involvement with the host presentation side such as SDL, OpenGL, shaders, etc.)

This commit changes that behaviour to render out at the row level (per row of digital pixel data in the video card's memory). Each row has a PIC timer setup to render the row as soon as the time comes when it would be drawn to the CRT. This is similar to how lower resolutions are handled.  It allows sophisticated games to start writing the next frame's line data “behind the beam”.

This fixes a regression in Microsoft's Space Simulator that was flickering at 800×600 with the previous whole frame approach.

There's a performance hit on memory-limited systems like the Pi 4: Quake at 800×600 drops from 11.5 fps to 7.8 fps, but this is simply the cost of doing things correctly (and it's critical for games that rely on this more accurate behaviour).

On faster systems, the performance hit is in the noise, so overall users won't be affected because Pi 4 users typically run games in their lowest, highest performance resolution anyway.

## Related issues

Fixes #3248 

# Manual testing

## Microsoft Space Simulator

Microsoft Space Simulator is an early 1994 real-mode game with only 386 CPU requirements. In addition to this rendering issue, its physics engine can run away at high cycle counts, so needs to be held back to reasonable period-correct rates.

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/ba07e066-ba2c-46ac-9d8c-765818255063

- **`dosbox.conf`**:
    
    ```ini
    [cpu]
    cycles = max 100% limit 60000
    ```

- **`SETUP.EXE` configured**:

    1. 800×600
    2. VESA 1.2

- **In-game settings**:
    Preferences -> precision -> max level 8

    [SPACESIM/SPACESIM.INI](https://github.com/dosbox-staging/dosbox-staging/files/13745230/SPACESIM.INI.ZIP)

    The precision can be directly set on line 34 of the `SPACESIM/SPACESIM.INI` text file. A host tool, like GNU `sed`, can be used to set it:

    ```shell
    sed -i '34s/.*/08/' SPACESIM/SPACESIM.INI
    ```

---

Regression tested Quake, VBETEST, and Screamer without issue.

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1442,7 +1442,7 @@ PixelFormat VGA_ActivateHardwareCursor()
 // A single point to set total drawn lines and update affected delay values
 static void setup_line_drawing_delays(const uint32_t total_lines)
 {
-	vga.draw.parts_total = total_lines > 480 ? 1 : total_lines;
+	vga.draw.parts_total = total_lines;
 
 	vga.draw.delay.parts = vga.draw.delay.vdend / vga.draw.parts_total;
 


kcgen:
(Suggest we cherry-pick this into 0.81.0-rc2 or final)

Grounded0:
Since performance is now better than in 0.80.1 we can be more accurate too.

Grounded0 (approved):

johnnovak:
Great quick fix @kcgen 🎉 The performance hit is not too bad either. Yeah best not to do it as a last minute RC1 change.

johnnovak (approved):

johnnovak:
One question, though. Do we understand why this was a regression, and why this is fixing it? In 0.80.1, this was not flickering even without the render-by-line approach.

kcgen:
> One question, though. Do we understand why this was a regression, and why this is fixing it? In 0.80.1, this was not flickering even without the render-by-line approach.

Yup, 0.80.1 and prior (through to upstream SVN and 0.74) emulates the rendering out to the CRT in chunks of rows, where the chunk was a quarter of the total number of vertical lines. This chunked approach is a coarser version of the per-line approach: it works so long as the game doesn't start drawing the next frame's content at the start of the chunk by the time you grab the memory at the end of the chunk (and no doubt: quarter sized chunked was probably the sweet spot).

johnnovak:
Thanks for the explanation @kcgen. With this change we're not just restoring previous correct behaviour then, but we're potentionally increasing compatibility for all other programs that use SVGA / VESA modes. Nice! 🎉

Grounded0:
Can someone run Quake 800x600 benchmarks with this against 0.81 RC1 and report the results on a half decent machine like Intel Skylake. Thanks.

johnnovak (approved):
Nice improvement 👍🏻

kcgen:
With 0.82 in place, merging.

PR #3260, opened 2023-12-24
Fix a couple more corner cases when using zlib-ng

kcgen:
# Description

- Adds `-mfpu=neon` flag to zlib-ng's CMake arguments, otherwise it fails to use Neon even when native optimizations are requested (affects gcc on arm Linux platforms).
- Drops the `auto / system / built-in` options for zlib-ng in favour of Meson's standardized wrap fallback controls (`--wrap-mode` and `--force-fallback-for=[list]`)
- Fixes a small `CFLAG` vs. `LDFLAG` typo.
- Fixes zlib-ng not being enabled for MSYS2 builds (now it is).

## Related issues

https://github.com/zlib-ng/zlib-ng/issues/1621

# Manual testing

Tested on the Pi 4, x86-64, Apple M1, and confirmed zlib-ng is being built and used under MSYS2.

# Checklist

I have:

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -123,7 +123,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 7
+          MAX_BUGS: 3
         run: |
           # summary
           echo "Full report is included in build Artifacts"
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -97,7 +97,7 @@ jobs:
 
           - name: GCC, Debian 10, ARMv7
             os: ubuntu-20.04
-            build_flags: -Dbuildtype=debug --cross-file dosbox-cross
+            build_flags: -Dbuildtype=debug -Duse_zlib_ng=false --cross-file dosbox-cross
             max_warnings: 21
             run_tests: true
             cross: true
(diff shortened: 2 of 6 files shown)


johnnovak (approved):
Noice 👍🏻

kcgen:
With 0.82 now in place; merging.

weirddan455:
It seems zlib-ng is disabled by default now.  My system zlib is plain old (slow) zlib.  Is that intentional?

```
  Build Summary
    Build type            : debug
    SIMD instruction set  : SSE2 and SSSE3
    Host page size (bytes): 4096

    zlib provider         : system library
```

I believe prior to this PR, it was installing zlib-ng from the meson wrap.  I realize I could use a flag to force wrap fallback soi f this is intended, all good.  Just wanted to confirm.

kcgen:
@weirddan455, yes, we only go through the extra work of setting up zlib-ng for optimized build.  I started to write more here explaining why and decided instead to codify it with feedback in Meson file (because, no doubt, others will wonder why, too).

The new PR will give this feedback in your case:

```
  Build Summary
    Build type            : debug
    SIMD instruction set  : SSE2 and SSSE3
    Host page size (bytes): 4096

    zlib-ng provider      : skipped (non-optimized build)
    zlib provider         : system library
    PNG provider          : system library
```

And I've added this comment in the meson.build file:

``` text
# The zlib-ng wrap, on the other hand, is a costly endeavour because it
# single-handledly quadruples the 'meson setup' time for the entire project (6.1s
# to 25.8s) due to the use of the much slower CMake build system, therefore we
# only expend this effort (to optimize zlib) for comensurate optimized builds.
# This nuance is also explained in meson_options.txt for packagers.
```

To benchmark your own builds:

```
git clean -ffdx; time meson setup build

git clean -ffdx; time meson setup -Dbuildtype=debug -Dunit_tests=disabled build
```

PR #3263, opened 2023-12-25
Fix macOS Info.plist template so the actual version is shown

johnnovak:
# Description

In the v0.81.0-rc1 macOS app bundle, this is what we have in the Info.plist file:

image

So we're only substituting one of the version strings at build time, and it turns out the least important internal one. If you check out the version of the app in Finder, it still shows the old version :

image

---

The simplest thing for us to do is just to set both to the same string at build time as this guy is doing it:

> Think of it this way: The "short version" (CFBundleShortVersionString) is the public version number. The "version" (CFBundleVersion) is more of an internal version number that could change far more frequently than the public "short version". Personally I use the same for both but many people update the "version" on every build. Either way you typically update the "short version" when you release to Apple. How often you update the "version" is up to you and your needs.

(source: https://stackoverflow.com/a/19728342)

# Manual testing

We'll find out next time an app bundle is built 😄 

# Checklist

I have:

### DIFF
--- a/contrib/macos/Info.plist.template
+++ b/contrib/macos/Info.plist.template
@@ -27,7 +27,7 @@
 	<string>APPL</string>
 
 	<key>CFBundleShortVersionString</key>
-	<string>0.82.0-alpha</string>
+	<string>%VERSION%</string>
 
 	<key>CFBundleVersion</key>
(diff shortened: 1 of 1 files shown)


kcgen (approved):
nice catch!

Issue #3268, opened 2023-12-26
Legend of Kyrandia buggy scroller in title screen in 0.81.0-rc1


BackSpace54:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-rc1

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Hi,
with 0.81.0-rc1 release and a first launch and unmodified dosbox-staging.conf file install Legend of Kyrandia game, launch setup and configure it with default settings (VGA, SB, Mouse).
At title screen the scroller is broken (see capture).
The scroller bug is not present in the 0.80.1 version.
![dosbox_Nwp16G9HcJ](https://github.com/dosbox-staging/dosbox-staging/assets/57332400/a2a4e737-9422-4765-b5f9-71d834135b80)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. install 0.81.0-rc1 and first run it
2. install Legend of Kyrandia and setup it with default settings (VGA, SB and mouse)
3. run Legend of Kyrandia, the island scroller is broken

### Your configuration

```ini
dosbox-staging configuration file is from the 1st run
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

rderooy:
I can confirm the same corruption on Linux.

weirddan455:
Confirmed and bisected.  Pinging @johnnovak 

```
97f73bf8c6737478837e273e38a5e5fcbe497c8a is the first bad commit
commit 97f73bf8c6737478837e273e38a5e5fcbe497c8a
Author: John Novak <john@johnnovak.net>
Date:   Sat Jun 24 22:40:34 2023 +1000

    Always recalculate the display dimensions when the Maximum Scan Line Register changes
    
    Because the timeout for recalculating the display dimensions was reduced
    from 50ms to half-frame time in commit b68b66f0 to fix some visual
    glitches in a few games, now other games don't always get the "full
    state" of the VGA registers when the recalculation hits on slower cycles
    settings.
    
    One example is the game Threat from 1995. The game sets up mode 13h
    first then tweaks it into 320x400 by sending a value of 40h to the
    Maximum Scanline Register to disable line-doubling. On the default 3000
    cycles setting, this value arrives invariably after the half-frame time,
[...]

PR #3279, opened 2023-12-28
VGA CRTC regression fixes

johnnovak:
# Description

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3268

I made a simplification earlier in https://github.com/dosbox-staging/dosbox-staging/commit/97f73bf8c6737478837e273e38a5e5fcbe497c8a which was plain wrong. Important functionality was removed, my bad. This PR reverts that, plus I've also done some cleanup and some proactive improvement based on the DOSBox X code (untested as I don't know how to trigger that rather obscure path, but the legacy DOSBox version seemed to be quite broken, so this can only be better, in theory 😅).

Definitely review commit by commit.

While the issue I described in the reverted https://github.com/dosbox-staging/dosbox-staging/commit/97f73bf8c6737478837e273e38a5e5fcbe497c8a commit's comment is true, I overestimated its importance. It can only affect logging, that's the short of it. The worst that can happen is that 320x400 custom modes in some games get logged as 320x200 _if and only if_ the cycles setting is set too low (so low that the games are not really playable anymore), but the graphics appear correctly otherwise. It stands to reason that similar logging issues could happen for other custom modes too that need messing around with the Maximum Scan Line VGA register, but I could not find any other games apart from these 320x400 titles.

Given these are 486/Pentium-era 90s games and the logging issue only happens below a ~5k threshold or much lower (depending on the game), getting the logging correct in those cases is more like a "nice to have" than a strict requirement. Maybe I or someone else will figure out later how to do the logging correctly even in these edge cases, but it's a very minor issue I think.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3268

https://github.com/dosbox-staging/dosbox-staging/pull/2528

https://github.com/dosbox-staging/dosbox-staging/commit/e7c596b26982e87f85cf3db8f4b4f392da3ac8e9

# Manual testing

The intro scroller in Legend of Kyriandia 1 is fixed.

Regression tested a few games I mentioned in https://github.com/dosbox-staging/dosbox-staging/commit/97f73bf8c6737478837e273e38a5e5fcbe497c8a. The outcome is that the 320x400 tweaked VGA mode indeed gets logged as 320x200 under a certain low cycles threshold (variable per game), but the graphics are otherwise displayed 100% correctly. **It's 100% a logging issue only.** Examples:

**Threat (1995)**
threshold for correct 320x400 logging: ~5200 cycles

**The Summoning (1992)**
threshold for correct 320x400 logging: ~4800 cycles

**Ravenloft: Strahd's Possession (1994)**
threshold for correct 320x400 logging: ~2500 cycles

Realistically, these games need well above 10k cycles, ideally ~25k which corresponds to the legendary 486DX2/66.

# Checklist

I have:

### DIFF
--- a/include/vga.h
+++ b/include/vga.h
@@ -257,6 +257,8 @@ struct VgaDraw {
 	Bitu lines_scaled           = 0;
 	Bitu split_line             = 0;
 
+	bool is_double_scanning = false;
+
 	// When drawing in parts, how many many 'chunks' should we draw at a
 	// time? A value of 1 is the entire frame where as a value of 2 will
 	// draw the top then the bottom, 4 will draw in quarters, and so on.
--- a/src/hardware/vga_crtc.cpp
+++ b/src/hardware/vga_crtc.cpp
@@ -19,15 +19,12 @@
  *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-#include <stdlib.h>
-#include "dosbox.h"
-#include "inout.h"
-#include "vga.h"
-#include "debug.h"
-#include "cpu.h"
-#include "video.h"
-#include "pic.h"
+#include <cstdlib>
 
+#include "cpu.h"
+#include "debug.h"
+#include "pic.h"
+#include "vga.h"
 
 void VGA_MapMMIO(void);
 void VGA_UnmapMMIO(void);
--- a/src/hardware/vga_crtc.cpp
+++ b/src/hardware/vga_crtc.cpp
@@ -28,8 +28,6 @@
 #include "video.h"
 #include "pic.h"
 
-#define crtc(blah) vga.crtc.blah
-
 
 void VGA_MapMMIO(void);
 void VGA_UnmapMMIO(void);
@@ -39,37 +37,37 @@ void vga_write_p3d5(io_port_t port, io_val_t value, io_width_t);
 void vga_write_p3d4(io_port_t, io_val_t value, io_width_t)
 {
 	const auto val = check_cast<uint8_t>(value);
-	crtc(index) = val;
+	vga.crtc.index = val;
 }
 
 uint8_t vga_read_p3d4(io_port_t, io_width_t)
 {
-	return crtc(index);
+	return vga.crtc.index;
 }
 
 void vga_write_p3d5(io_port_t, io_val_t value, io_width_t)
 {
 	const auto val = check_cast<uint8_t>(value);
-	// if (crtc(index) > 0x18) {
-	// 	LOG_MSG("VGA CRCT write %" sBitfs(X) " to reg %X", val, crtc(index))
+	// if (vga.crtc.index > 0x18) {
+	// 	LOG_MSG("VGA crtc write %" sBitfs(X) " to reg %X", val, vga.crtc.index)
 	// }
 
-	switch (crtc(index)) {
+	switch (vga.crtc.index) {
 	case 0x00: // Horizontal Total Register
-		if (crtc(read_only)) {
+		if (vga.crtc.read_only) {
 			break;
 		}
-		crtc(horizontal_total) = val;
+		vga.crtc.horizontal_total = val;
 
 		// 0-7  Horizontal Total Character Clocks minus 5.
 		break;
 
 	case 0x01: // Horizontal Display End Register
-		if (crtc(read_only)) {
+		if (vga.crtc.read_only) {
 			break;
 		}
-		if (val != crtc(horizontal_display_end)) {
-			crtc(horizontal_display_end) = val;
+		if (val != vga.crtc.horizontal_display_end) {
+			vga.crtc.horizontal_display_end = val;
 			VGA_StartResize();
 		}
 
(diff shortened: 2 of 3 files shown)


src/hardware/vga_crtc.cpp
```diff
@@ -178,14 +178,55 @@ void vga_write_p3d5(io_port_t, io_val_t value, io_width_t)
...
-		VGA_StartResize();
-		break;
+
+		if (vga.draw.is_double_scanning) {
+			auto flipped = new_val;
+			flipped.start_vertical_blanking_bit9.flip();
+
+			// Only true if only `start_vertical_blanking_bit9` has
+			// been changed and all other bits are unchanged.
+			if (old_val.data == flipped.data) {
```
johnnovak, line 198: This check was previously expressed as `if ((vga.crtc.maximum_scan_line ^ val) & 0x20)`. I think my way of doing the check is way more readable, but open to suggestions (suggestions that work _with_ `bit_view`, and not against it 😅).

src/hardware/vga_crtc.cpp
```diff
@@ -178,14 +178,55 @@ void vga_write_p3d5(io_port_t, io_val_t value, io_width_t)
...
+			//   - `start_vertical_blanking_bit9`
+			//   - `is_scan_doubling_enabled`
+			// and all other bits are unchanged.
+			//
+			// This revised criteria was taken from DOSBox X. The
+			// old DOSBox criteria was "if ALL bits have been
+			// changed EXCEPT for `line_compare_bit9`"; I can't see
+			// how could that possibly work.
+			//
+			if (old_val.data == flipped.data) {
```
johnnovak, line 219: Similarly, I think this is easy to understand if you read the comment. The alternative way via bit-twiddling is rather obscure (took me a good ten minutes to decipher it as I'm out of practice with that stuff 😎).

src/hardware/vga_crtc.cpp
kcgen: Very nice! After reading your comment, the code makes perfect sense, and is self documenting by flipping the vertical blank bit by name 🏅

src/hardware/vga_crtc.cpp
kcgen: So much easier to read and understand. Doesn't get easier than this!

src/hardware/vga_crtc.cpp
```diff
@@ -152,147 +192,173 @@ void vga_write_p3d5(io_port_t, io_val_t value, io_width_t)
...
-			5	Turns Cursor off if set
-		*/
+	case 0x0a: // Cursor Start Register
+		crtc(cursor_start)    = val;
+		vga.draw.cursor.sline = val & 0x1f;
+		if (IS_VGA_ARCH) {
+			vga.draw.cursor.enabled = !(val & 0x20);
+		} else {
+			vga.draw.cursor.enabled = true;
+		}
```
kcgen: This cleanup commit is so much nicer on the eyes with spaces and carriage returns. The packed-one-liners was just too dense to take in (for these old eyes, anyway!)

johnnovak: Cheers. Yeah, just the logic is obscure and complex enough; there's no need to complicate it even further with bad style 😅  We should take all the small readability improvements we can get IMO.

src/hardware/vga_crtc.cpp
johnnovak: Thanks man 😎

include/vga.h
```diff
@@ -257,6 +257,8 @@ struct VgaDraw {
 	Bitu lines_scaled           = 0;
 	Bitu split_line             = 0;
 
+	bool is_double_scanning = false;
+
```
kcgen, line 261: Also have `double_scanning_enabled = false` in the same `vga.draw.` struct on line 294.  There's probably a good reason for these (maybe one is existing and the other is the new/changing state?).

Also have `video_mode.is_double_scanned_mode`, in a different struct - but maybe it could share the same name (if the meaning is identical?)

johnnovak: These are three related but different things. All three are needed, they serve different purposes. In order:

- `VgaDraw.double_scanning_enabled` -- This existing flag tells us whether double-scanning is allowed or not. The regular VGA shaders allow double-scanning, in which case this is true. But we can force single-scanning in the arcade shaders with `#pragma force_single_scan`; then the flag is false. So this is basically a configuration option for the renderer.
- `VgaDraw.is_double_scanning` -- I introduced this now, this reflects actual reality. Are we in double-scanned rendering mode, yes or no? If `double_scanning_enabled` is false, this can only be false, of course, as double-scanning is disabled.
- `video_mode.is_double_scanned_mode` -- Regardless of whether double-scanning is currently enabled via `double_scanning_enabled` or not, this flag represents in the VideoMode struct whether the given mode is a double-scanned mode. This info is needed in the upscaler, and maybe in the shader switcher too.

Happy to change the names if you have improvement ideas. Maybe `double_scanning_enabled` could be called `double_scanning_allowed` (enabled can be misleading). But the other two I think are fine.

kcgen: Thanks for this explanation!

Yes, that rename to `_allowed` looking like a nice improvement.

I would also say your bullet comments here could be pasted straight into the structs just above each of these three variables (good IDEs will expand the preceding comment block, so would be nice to have your explanations always in the code :-)

kcgen (approved):
Very nice improvement and fix!

include/vga.h
johnnovak: Yeah I think I've documented some of these structs/flags already, but I'll double check and add the missing descriptions where needed. Using "allowed" is definitely an improvement, agree (just read my above explanation, it just "begs" for allowed in the name 😆).

PR #3281, opened 2023-12-28
Fix a couple of FPU emulation issues

kklobe:
# Description

Since the memcpy trick fix for vertical bars is only required when `C_FPU_X86` is false, don't compile in the extra arrays and associated logic when they aren't needed.

There were also a couple of places where I had missed setting the `use_regs_memcpy[]` array element to `false` when the corresponding`fpu.tags[]` element is set to `TAG_Empty`, specifically with the `FFREE` instruction, which so far I've only seen used by `PERF_DOS`.

# Manual testing

I tested Carmageddon, Sunflower, and MCPDIAG on:

- macOS arm64, normal and dynamic cores
- macos x86-64, normal and dynamic cores
- Windows MSYS2 x64-86, normal and dynamic core

If anyone can run through the Linux builds for normal and dynamic core, that would be appreciated.

# Checklist

I have:

### DIFF
--- a/include/fpu.h
+++ b/include/fpu.h
@@ -84,8 +84,10 @@ enum FPU_Round : uint8_t {
 
 struct FPU_rec {
 	FPU_Reg regs[9]         = {};
+#if !C_FPU_X86
 	int64_t regs_memcpy[9]  = {}; // for FILD/FIST 64-bit memcpy fix
 	bool use_regs_memcpy[9] = {};
+#endif
 	FPU_P_Reg p_regs[9]     = {};
 	FPU_Tag tags[9]         = {};
 	uint16_t cw             = 0;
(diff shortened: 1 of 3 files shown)


kcgen (approved):
Very nice cleanup, @kklobe !

kcgen:
> If anyone can run through the Linux builds for normal and dynamic core, that would be appreciated.

Calling in @FeralChild64 and @weirddan455 for help! (my repo is currently in bisect land; thanks guys!)

kcgen:
With a couple new fixes, we should probably get this backported to 0.81.0-rc2, I think?

johnnovak:
> With a couple new fixes, we should probably get this backported to 0.81.0-rc2, I think?

I think so, yeah.

johnnovak (approved):
Looks good man, but I'm not close to this code so I'll trust you on it 😄

FeralChild64:
Checked MCPDIAG, Quake, and GTA (both normal and 3dfx) - both dynamic and normal core, Linux build. No problems found.

Non-3dfx GTA on normal core is too slow to be playable, unfortunately.

PR #3283, opened 2023-12-28
Remove obsolete check for forbidden paths on POSIX systems

FeralChild64:
# Description

DOSBox contains code that prevents accessing files/directories within the `/proc` filesystem on certain Unix-like OS'es; it is realized by `fopen_wrap`, which does the checks and calls `fopen` if everything is OK. 

~~To support file attributes on POSIX systems, creating new files is now realized in a different way (so that we can use extended attributes), which causes the checks to be omitted. This PR restores the checks when creating new files.~~

It was decided to remove these checks, as they are considered incomplete (no check for `/sys`, for example) and not needed (we have a permission system, and DOSBox Staging is not supposed to be run as root).

# Checklist

I have:

### DIFF
--- a/src/misc/cross.cpp
+++ b/src/misc/cross.cpp
@@ -368,7 +368,8 @@ void close_directory(dir_information* dirp) {
 
 #endif
 
-FILE *fopen_wrap(const char *path, const char *mode) {
+static bool is_path_allowed([[maybe_unused]] const char* path)
+{
 #if defined(WIN32)
 	;
 #elif defined(MACOSX)
@@ -392,35 +393,34 @@ FILE *fopen_wrap(const char *path, const char *mode) {
 			if ( ( strlen(check) == 5 && strcmp(check,"/proc") == 0) || strncmp(check,"/proc/",6) == 0) {
 //				LOG_MSG("lst hit %s blocking!",path);
 				free(check);
-				return nullptr;
+				return false;
 			}
 			free(check);
 		}
 	}
-
-#if 0
-//Lightweight version, but then existing files can still be read, which is not ideal	
-	if (strpbrk(mode,"aw+") != NULL) {
-		LOG_MSG("pbrk ok");
-		char* check = realpath(path,NULL);
-		//Will be null if file doesn't exist.... ENOENT
-		//TODO What about unlink /proc/self/mem and then create it ?
-		//Should be safe for what we want..
-		if (check) {
-			if (strncmp(check,"/proc/",6) == 0) {
-				free(check);
-				return NULL;
-			}
-			free(check);
-		}
-	}
-*/
-#endif //0 
 
 #endif //HAVE_REALPATH
 #endif
+	return true;
+}
 
-	return fopen(path,mode);
+bool is_path_allowed(const std_fs::path& path)
+{
+#if defined(WIN32)
+	// path.c_str() would be UTF-16 encoded
+	return is_path_allowed(path.string().c_str());
+#else
+	return is_path_allowed(path.c_str());
+#endif
+}
+
+FILE* fopen_wrap(const char* path, const char* mode)
+{
+	if (!is_path_allowed(path)) {
+		return nullptr;
+	}
+
+	return fopen(path, mode);
 }
 
 // A helper for fopen_wrap that will fallback to read-only if read-write isn't possible.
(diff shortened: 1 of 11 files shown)


src/misc/cross.cpp
```diff
@@ -392,35 +393,34 @@ FILE *fopen_wrap(const char *path, const char *mode) {
...
+}
+
+bool is_path_allowed(const std_fs::path& path)
+{
+#if defined(WIN32)
+	// path.c_str() would be UTF-16 encoded
+	return is_path_allowed(path.string().c_str());
+#else
+	return is_path_allowed(path.c_str());
+#endif
```
weirddan455, line 414: I'd probably get rid of the `#ifdef` and just do `path.string().c_str()` everywhere.  That's what I've done elsewhere in the code in any case.

weirddan455 (approved):
Looks good to me.  Just one small suggestion.

src/misc/cross.cpp
FeralChild64: This would have triggered static analysis warning about unnecessary conversion. That's why you can find `path.string().c_str()` in f`s_util_win32.cpp`, but `path.c_str()` in `fs_util_posix.cpp`

johnnovak:
So I'm looking at this special Linux-only path exclusion and I'm scratching my head:

1. Is it normal for Linux GUI programs to do such exclusions manually per app, in the app's code? Surely most command line programs rely on file permissions only to prohibit access at the OS-level. 
2. This is an OS level concern, isn't it? What if someone mounts `/` as the C drive as root and starts deleting files in the DOS prompt? Why should we even prevent that? I probably can fuck up any system with VLC if I'm creative enough 😎 
4. What kind of "security" does this even achieve? Isn't it possible for a malevolent attacker to cause harm via DOSBox in 10000 other ways that we can't prevent anyway?
5. Why not protect 50+ other important system paths too, why just this?
6. macOS is a Unix. Doesn't it also deserve our "protection"? 😂 
7. While we're at it, why not "protect" `C:\Windows`? Throw in maybe the AppData folder in the user's home folder too?

I could go on all day. So why don't we just get rid of this cruft? 😄

FeralChild64:
> Is it normal for Linux GUI programs to do such exclusions manually per app, in the app's code?

I'm 99% sure it's not normal. This was probably done for some very specific use case.

> What kind of "security" does this even achieve?

Probably it can prevent from user mistakes in some very insecure systems - nothing more.

> So why don't we just get rid of this cruft?

If the team agrees, I'll happily get rid of this. I intend to keep the fix commit in the history in case there is some important reason to have this protection.

johnnovak:
> > Is it normal for Linux GUI programs to do such exclusions manually per app, in the app's code?
> 
> I'm 99% sure it's not normal. This was probably done for some very specific use case.
> 
> > What kind of "security" does this even achieve?
> 
> Probably it can prevent from user mistakes in some very insecure systems - nothing more.
> 
> > So why don't we just get rid of this cruft?
> 
> If the team agrees, I'll happily get rid of this. I intend to keep the fix commit in the history in case there is some important reason to have this protection.

Yeah, I hate such special stuff as you can tell, so nuke it by all means... unless someone vehemently disagrees and can cite _one_ good practical reason for keeping it 😄

kcgen:
>  I intend to keep the fix commit in the history in case there is some important reason to have this protection.

Sounds like a perfect approach, @FeralChild64.

FeralChild64:
OK, change is ready for re-review. The checks have been removed.

weirddan455 (approved):
Yeah I wasn't thinking about why we needed this check in the first place.  It doesn't make a lot of sense for the reasons @johnnovak pointed out.

I'm a little worried this might uncover a bug somewhere.  It almost smells like a hack.  I really can't think of a good reason why we need this check though so agree with getting rid of it.

Still looks fine to me from a code perspective.

johnnovak (approved):
Good riddance @FeralChild64. Less weird legacy cruft 🚀

MasterO2:
Dosbox-Staging excels at removing weird legacy cruft.

johnnovak:
> I'm a little worried this might uncover a bug somewhere. It almost smells like a hack that maybe someone put in to "fix" something (mostly because I can't think of any other reason why this would be there). Either way, I agree with getting rid of it. We'll be testing and debugging stuff heavily anyway.

The good news is it can absolutely not affect Windows and macOS 😎 And only some weirdo would try to mount `/proc` as a DOS drive 🤣

PR #3284, opened 2023-12-28
Use C++ system headers

kklobe:
# Description

A long-overdue chore, I went and replaced the old C library .h files with the C++ equivalents.

I left `src/libs` alone for now, and the reelmagic `mgeg_decoder.h` as well (open to discussion).

As an interesting note, the C++ headers provide additional functionality in some cases. For example, Apple's Xcode 15.1 `ctime` provides the following:

```cpp
// TODO: Remove once rdar://101515782 is resolved
#ifdef _LIBCPP_ON_SEP
inline size_t strftime(char *, size_t, const char *, const struct tm *) { return 0; }
#endif
```

`cstring` provides:

```cpp
inline _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX14 size_t __constexpr_strlen(const char* __str) {
  // GCC currently doesn't support __builtin_strlen for heap-allocated memory during constant evaluation.
  // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70816
#ifdef _LIBCPP_COMPILER_GCC
  if (__libcpp_is_constant_evaluated()) {
    size_t __i = 0;
    for (; __str[__i] != '\0'; ++__i)
      ;
    return __i;
  }
#endif
  return __builtin_strlen(__str);
}
```

# Manual testing

Ran the build artifacts for:

- macOS x86-64
- macOS arm64
- Windows MSYS2 x86-64

# Checklist

I have:

### DIFF
--- a/src/cpu/callback.cpp
+++ b/src/cpu/callback.cpp
@@ -17,9 +17,9 @@
  *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+#include <cstdlib>
+#include <cstring>
 #include <limits>
-#include <stdlib.h>
-#include <string.h>
 #include <string>
 
 #include "dosbox.h"
--- a/src/cpu/core_dyn_x86.cpp
+++ b/src/cpu/core_dyn_x86.cpp
@@ -36,7 +36,7 @@
 #if defined(HAVE_MPROTECT) || defined(HAVE_MMAP)
 #include <sys/mman.h>
 
-#include <limits.h>
+#include <climits>
 
 #endif // HAVE_MPROTECT
 
--- a/src/cpu/core_dyn_x86/dyn_fpu.h
+++ b/src/cpu/core_dyn_x86/dyn_fpu.h
@@ -23,13 +23,12 @@
 
 #if C_FPU
 
-#include <math.h>
-#include <float.h>
-#include "cross.h"
-#include "mem.h"
-#include "fpu.h"
 #include "cpu.h"
-
+#include "cross.h"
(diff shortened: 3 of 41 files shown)


kcgen:
Now that's some cleanup, and a nice runway to C++20 :airplane: 
(My guess is we'll queue this for 0.82, but I don't see anything risky either, so very well could be a trivial backport too. @johnnovak , your call)

kcgen (approved):

kklobe:
> Now that's some cleanup, and a nice runway to C++20 ✈️ (My guess is we'll queue this for 0.82, but I don't see anything risky either, so very well could be a trivial backport too. @johnnovak , your call)

That was my main motivation, can't see any really good reason to risk some weird compiler-library-specific regressions for 0.81, but up to you guys.

johnnovak:
> > Now that's some cleanup, and a nice runway to C++20 ✈️ (My guess is we'll queue this for 0.82, but I don't see anything risky either, so very well could be a trivial backport too. @johnnovak , your call)
> 
> That was my main motivation, can't see any really good reason to risk some weird compiler-library-specific regressions for 0.81, but up to you guys.

I see it as too noisy & too risky. It's a good long-term cleanup, no doubt, but very little to be gained from backporting it short-term. I'd say, 0.82 because I like my life simple 😎

johnnovak (approved):
Nice one @kklobe. I've done a similar cleanup in one of the files in my VGA CRTC commit, but I'm leaning towards holding this off until 0.82 as this is doing the same thing at a mass-scale.

PR #3285, opened 2023-12-29
Refactor logging to use loguru library.

naerbnic:
# Description

There is complex logic that is enabled/disabled when C_DEBUG is set concerning the log architecture. Some function and class definitions are enabled/disabled/replaced depending on its value. This makes it hard to reason about the behavior of the code both at runtime and compilation, as indicated by the fact that the tests in the current main do not compile when enable_debugger is not "none".

This CL is a refactor of the logging architecture that makes the following changes:

- Have a single set of logging macros and definitions that do not depend on C_DEBUG and are always compiled in.
- Route those logs through the already-included loguru library.
- Use loguru callbacks to route log messages to a log file and to the debugger window (through the Logger class).

This has some significant advantages compared to the previous version:

- Behavior of how messages are logged are changed less between debug and release builds. In debug builds, no messages were written to stderr, but in release builds, they were. Now, messages are always written to stderr.
- The "log/logfile" option now works in both debug and release builds.
- "logging.h" now only depends on code in "logging.cpp", making building and including it in other libraries simpler.
- The option for logging file and line numbers when using the LOG() macro are now available.
- The "Logger" class in "logging.h" allows for other parts of the code to respond to logged messages through loguru.

## Related issues

# Manual testing

- Ran quake using the built binary, testing with enable_debugger=non,norm
- Set the `log/logfile` option in the config file, and tested it was populated.
- Checked that messages were logged into the debugger console when open during execution by running "debug quake -nosound", then using the step-over key repeatedly.

# Checklist

I have:


johnnovak:
@weirddan455 I think you found some concurrency issues with loguru earlier? Do we even want this change? Thoughts @kklobe ?

weirddan455:
> @weirddan455 I think you found some concurrency issues with loguru earlier? Do we even want this change? Thoughts @kklobe ?

I don't have time to review this now but the issue I found was a hard to reproduce hang on exit bug on the MSVC Windows build.  I can't remember if it was before or after @kklobe changed it to use Clang.  I also couldn't confirm it was a Loguru issue.

johnnovak:
> > @weirddan455 I think you found some concurrency issues with loguru earlier? Do we even want this change? Thoughts @kklobe ?
> 
> I don't have time to review this now but the issue I found was a hard to reproduce hang on exit bug on the MSVC Windows build. I can't remember if it was before or after @kklobe changed it to use Clang. I also couldn't confirm it was a Loguru issue.

Maybe we'll just merge this then as work is already done as see how it goes.

kklobe:
> @weirddan455 I think you found some concurrency issues with loguru earlier? Do we even want this change? Thoughts @kklobe ?

I'm strongly considering a migration from loguru -> spdlog in the near future (given that dev on loguru is dead), so that would create a little extra work, but I like this change because, honestly, it's work I should have done when I added loguru.

@naerbnic, if you're still around, please rebase this on main, then I'd like to run some testing.

src/misc/meson.build
```diff
@@ -8,6 +8,7 @@ libmisc_nomsg_sources = [
     'fs_utils_posix.cpp',
     'fs_utils_win32.cpp',
     'help_util.cpp',
+    'logging.cpp',
```
kklobe, line 11: Changes in meson.build files need to propogate to the Visual Studio solution as well.

naerbnic:
> @naerbnic, if you're still around, please rebase this on main, then I'd like to run some testing.

I will do so by tomorrow. 

Loguru did seem to be a bit idiosyncratic, but I didn't want to make significant waves in the codebase

src/misc/logging.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+// A thread_local value that is used to pass along the message type (if any) to
+// the logger.
+//
+// This makes the assumption that the log callback will be called in the dynamic
+// scope of the log call itself. If this does not happen, then the value will be
+// lost.
+thread_local LOG_TYPES curr_msg_log_type = LOG_ALL;
+
+static void OnLogCallback(void* logger, const loguru::Message& message)
```
johnnovak, line 56: lower_came_case for static functions please

src/misc/logging.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+static void LOG_Destroy(Section*)
+{
+	Logger::RemoveLogger(LOGURU_FILE_CALLBACK_ID);
+}
+
+static void LOG_Init(Section* sec)
+{
+	Section_prop* sect = static_cast<Section_prop*>(sec);
+	std::string blah   = sect->Get_string("logfile");
```
johnnovak, line 127: Maybe call this `logfile_setting` 😄

johnnovak (approved):
Just a few minor comments about naming @naerbnic, great work. A very nice improvement indeed that brings more consistency to the logging.

PR #3286, opened 2023-12-29
Migrate to C++20

kklobe:
# Description

Makes the necessary changes to allow building under C++20. I mostly just did what was required to get everything green on CI, except for the technically non-necessary addition of #3284.

I found I had to nuke my Meson `build` folders (or do a `meson setup --wipe`) when switching standards.

[Clang 14 on the Ubuntu CI has a known C++20 issue](https://github.com/actions/runner-images/issues/8659) so I bumped it to 15.

The existing semaphore files conflicted with the C++20 native one, and unfortunately it wasn't a drop-in replacement, but soon enough it will be updated to use something native.

# Manual testing

The CI builds were all green, and I've compiled release and debug builds on my macOS system with clang 15 and GCC 13.
 
# Checklist

I have:


kklobe:
@kcgen I think a first big decision will be deciding the new minimum compiler versions. I implemented native `semaphore`  in `voodoo.cpp` and most of the CI builds blew up trying to compile it.

kcgen:
This is looking great @kklobe. 

Ratcheting forward the language standard `--cpp_std=c++20`  is exactly how we moved from C++11 to 14 to 17.

In terms of dealing with compilers supporting a gradient of C++20 features:
- [GCC 8.x has basic support](https://gcc.gnu.org/projects/cxx-status.html#cxx20), while 11.x has full support
- [Clang 6.x has basic support](https://clang.llvm.org/cxx_status.html#cxx20), while 17.x is mostly complete

One school of thought is to use those tables and test the compiler versions; another approach is using [Meson's feature tests](https://mesonbuild.com/Compiler-properties.html#does-a-header-exist) with `HAVE_FEATURE` pre-processor items, finally there's the ISO C++ [SD-6 feature tests](https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations), which would have us writing code like the following (some C++14 compilers lagged in their `<optional>` support):

```c++
#ifdef __has_include
#  if __has_include(<optional>)
#    include <optional>
#    if __cpp_lib_optional >= 201606
#      define have_optional 1
#    endif
#  elif __has_include(<experimental/optional>)
#    include <experimental/optional>
#    if __cpp_lib_experimental_optional >= 201400
#      define have_optional 1
#      define experimental_optional 1
#    endif
#endif
#ifndef have_optional
#    define have_optional 0
#endif
```

I don't like any of the above because those approach riddle the build system and/or sources with versions and language feature tests. It's so bad that you would wind up re-implmenting those GCC and Clang suport tables in Meson, and have a pile of `HAVE_<C++20 FEATURE>` defines.

The approach we used for C++11 -> 14 -> 17 was to let CI's oldest in-support compiler act as the lower-bound and the newest CI compiler act as the upper bound.

CI will either pass or fail when we use a particular feature, naturally gating our use of features.

The task for the maintenance team is dropping EOL systems (like Debian 10 in July 2024) and opening up new CI machines with their latest compiler.

With time, all the language features are simply available and we use them freely - like `<optional>` today.

The only sticking point is when Apple refuses to correctly support something in Xcode Clang that's actually supported in "normal" Clang (like the `std::filesystem` fiasco).  So in those cases.. we might need to back fill with 3rd part implementations (like GHC `filesystem`).

kcgen:
One of the biggest challenges isn't on the low side but throttling the use of features on the high side.

For example:

- Low-side developer: running Debian 10 with GCC 8.3, they can freely use all C++20 feature that their local compiler builds.
    When they push their PR, it will likely build across the board.

- High-side developer: running Arch Linux with GCC 13.x, they work on a PR that uses `modules`, `concepts`, and `coroutines`.
    Everything compiler and runs fine locally.
    When they push their PR, it fails to compile on almost every system. They then argue with the team about _"why do you say you support C++20 but not all the features"_. :sob:

johnnovak:
> The approach we used for C++11 -> 14 -> 17 was to let CI's oldest in-support compiler act as the lower-bound and the newest CI compiler act as the upper bound.

Agree 100%, that's the best, safest, and lowest-maintenance approach. Let natural selection sort what we can and cannot use 😄 



> With time, all the language features are simply available, and we use them freely - like `<optional>` today.

Exactly, what we can and cannot use from C++20 is naturally determined and evolves with what our "lowest-bound" target supports. 

That will need some experimentation, but we should start using new language features carefully and sparingly anyway. I see this as a feature, not a bug 😎

kklobe:
> The approach we used for C++11 -> 14 -> 17 was to let CI's oldest in-support compiler act as the lower-bound and the newest CI compiler act as the upper bound.

That is definitely the approach I agree with, so, the question at this point is, does that lowest common denominator compiler need to be revisited, or is what we're using now in CI that LCD?

kcgen:
> so, the question at this point is, does that lowest common denominator compiler need to be revisited

For GCC, we should only be using the in-support versions, which are 11.x, 12.x, and 13.x. They're EOL'd when a `.5` MINORVER is released: https://gcc.gnu.org/releases.html

For Clang, it looks like 16, 17, and 18 are supported, according to [[1](https://llvm.org/docs/DeveloperPolicy.html#updating-toolchain-requirements)] and [[2](https://apt.llvm.org/)]:
> "It is a general goal to support LLVM and GCC versions from the last 3 years at a minimum."
> "Packages are available for [snip]. This for both the stable, qualification and development branches (currently 16, 17 and 18)."

For the `main` CI workflows, we should try for maximum OS coverage across these compiler versions and drop any OS that can't give them to us.

include/string_utils.h
```diff
@@ -363,7 +363,7 @@ std::optional<float> parse_percentage_with_percent_sign(const std::string_view s
 std::optional<float> parse_percentage_with_optional_percent_sign(const std::string_view s);
 
 template <typename... Args>
-std::string format_string(const std::string& format, const Args&... args) noexcept
+std::string format_str(const std::string& format, const Args&... args) noexcept
```
FeralChild64, line 366: Could you explain what is the rationale behind this change?

kklobe: I put a short note in the commit message:

```log
C++20 provides std::format_string when <format> is included, triggering a compiler error on MSClang.
```

So the <format> header dumps [`format_string`](https://en.cppreference.com/w/cpp/utility/format/basic_format_string) into the global namespace, and MSClang cannot compile it. We shouldn't be having the same name as built-ins.

FeralChild64 (approved):

johnnovak (approved):
Looking great @kklobe, the commits are nicely separated. Merge away 🚀

PR #3293, opened 2023-12-30
Remove GCC_LIKELY/GCC_UNLIKELY

kklobe:
This is just to performance test the removal of the GCC_LIKELY/GCC_UNLIKELY macros. If no one sees any performance loss over this, I'll remove the noise from the codebase entirely.

EDIT: I lied, I removed it from the codebase entirely. I've noticed zero performance difference on macOS or Windows from cleaning these out entirely.

### DIFF
--- a/src/cpu/paging.cpp
+++ b/src/cpu/paging.cpp
@@ -217,7 +217,7 @@ static inline void InitPageCheckPresence(PhysPt lin_addr,bool writing,X86PageEnt
 		PAGING_PageFault(lin_addr,table_addr,
 			(writing?0x02:0x00) | (((cpu.cpl&cpu.mpl)==0)?0x00:0x04));
 		table.set(phys_readd(table_addr));
-		if (GCC_UNLIKELY(!table.p)) {
+		if (!table.p) {
 			E_Exit("Pagefault didn't correct table");
 		}
 	}
@@ -228,7 +228,7 @@ static inline void InitPageCheckPresence(PhysPt lin_addr,bool writing,X86PageEnt
 		PAGING_PageFault(lin_addr,entry_addr,
 			(writing?0x02:0x00) | (((cpu.cpl&cpu.mpl)==0)?0x00:0x04));
 		entry.set(phys_readd(entry_addr));
-		if (GCC_UNLIKELY(!entry.p)) {
+		if (!entry.p) {
 			E_Exit("Pagefault didn't correct page");
 		}
 	}
@@ -992,10 +992,11 @@ void PAGING_Enable(bool enabled) {
 	if (paging.enabled==enabled) return;
 	paging.enabled=enabled;
 	if (enabled) {
-		if (GCC_UNLIKELY(cpudecoder==CPU_Core_Simple_Run)) {
-//			LOG_MSG("CPU core simple won't run this game,switching to normal");
-			cpudecoder=CPU_Core_Normal_Run;
-			CPU_CycleLeft+=CPU_Cycles;
+		if (cpudecoder == CPU_Core_Simple_Run) {
+			// LOG_MSG("CPU core simple won't
+			// run this game,switching to normal");
+			cpudecoder = CPU_Core_Normal_Run;
+			CPU_CycleLeft += CPU_Cycles;
 			CPU_Cycles=0;
 		}
 //		LOG(LOG_PAGING,LOG_NORMAL)("Enabled");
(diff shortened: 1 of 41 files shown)


johnnovak:
Fuck me, it's a like a virus! It's everywhere! 😱 

(great job @kklobe 😎 )

kklobe:
> Definitely something for 0.82, I feel uneasy backporting it.

Yep, just getting a bunch of janitorial work done so it gets in 0.82 right away.

src/cpu/paging.cpp
```diff
@@ -992,10 +992,11 @@ void PAGING_Enable(bool enabled) {
 	if (paging.enabled==enabled) return;
 	paging.enabled=enabled;
 	if (enabled) {
-		if (GCC_UNLIKELY(cpudecoder==CPU_Core_Simple_Run)) {
-//			LOG_MSG("CPU core simple won't run this game,switching to normal");
-			cpudecoder=CPU_Core_Normal_Run;
-			CPU_CycleLeft+=CPU_Cycles;
+		if (cpudecoder == CPU_Core_Simple_Run) {
+			//			LOG_MSG("CPU core simple won't
+			// run this game,switching to normal");
```
japsmits, line 997:
```suggestion
			// LOG_MSG("CPU core simple won't run this game,switching to normal");
```

src/gui/render.cpp
```diff
@@ -262,13 +261,13 @@ extern uint32_t PIC_Ticks;
...
 {
-	if (GCC_UNLIKELY(!render.updating)) {
+	if (!render.updating) {
 		return;
 	}
 
 	RENDER_DrawLine = empty_line_handler;
 
-	if (GCC_UNLIKELY((CAPTURE_IsCapturingImage() || CAPTURE_IsCapturingVideo()))) {
+	if ((CAPTURE_IsCapturingImage() || CAPTURE_IsCapturingVideo())) {
```
japsmits, line 270:
```suggestion
	if (CAPTURE_IsCapturingImage() || CAPTURE_IsCapturingVideo()) {
```

japsmits (approved):
Only a few cosmetic things, most likely caused by reformatting.

kklobe:
ok @johnnovak, let me know if this is something we still want!

johnnovak:
> ok @johnnovak, let me know if this is something we still want!

We want it, but we should be careful. We need more data points to determine whether removing these hints doesn't have a negative performance impact.

> I've noticed zero performance difference on macOS or Windows from cleaning these out entirely.

That's a good start, but I'm planning to test this on the Raspberry Pi 4 box I specifically bought for Staging dev on Linux. I think that's a reasonable low-end test case. I will also test things on my i7 4790k, which is bought in 2015. If there is no measurable performance impact, then we can merge. 

Anyone else can test this on low-end / weak hardware, please step forward 😄

johnnovak:
Btw, probably these macros are just noise. Interesting reading material, quoting in full ([source](https://softwareengineering.stackexchange.com/a/375352))

> If you're writing for x86/x64 (and are not using 20-year-old CPUs), the performance gain from using __builtin_expect() will be negligible if any. The reason for it is that modern x86/x64 CPUs (not 100% sure about Atom though), have dynamic branch prediction, so essentially CPU "learns" about the branch which is taken more often. Sure, this information can be stored only for a limited number of branches, however, there are only two cases possible. If (a) it is a "frequently used" branch, then your program will benefit from that dynamic branch prediction, and if (b) it is "rare" branch, you won't really see any realistic performance hit due to mispredictions in such rare branches (20 CPU cycles of branch misprediction are not TOO bad if it happens once in a blue moon).
>
> NB: this does NOT imply that on modern x86/x64 importance of branch misprediction got lower: any branch with 50-50 chance of jump-nojump will still incur a penalty (IIRC 10-20 CPU cycles), so in inner loops branches may still need to be avoided. It is only importance of __builtin_expect() on x86/x64 which has diminished (IIRC, around 10-15 years ago or so) - mostly because of dynamic branch prediction.

kklobe:
> Btw, probably these macros are just noise. Interesting reading material, quoting in full ([source](https://softwareengineering.stackexchange.com/a/375352), comment made in 2018)

Yes, I will be shocked if there's any measurable difference. This is also a good read (basically C++20's standardized version of GCC_UN/LIKELY: https://blog.aaronballman.com/2020/08/dont-use-the-likely-or-unlikely-attributes/

johnnovak:
Yeah, probably it made sense to have these in the early 2000s, but not anymore. So yeah, I'm 90% sure it belongs the "useless legacy junk no one needs at this point" category :) It's a good excuse for me to tinker around a bit on the RPi4, though, so I'll do the tests 😄 Then we'll merge, I'm quite sure 😄

FeralChild64 (approved):

johnnovak:
Actually, I'll be busy putting together my new PC, so feel free to merge if you guys have done benchmarking and are happy.

kklobe:
> Actually, I'll be busy putting together my new PC, so feel free to merge if you guys have done benchmarking and are happy.

I can't see a bit of difference on three different PCs (Mac & Windows).

Merging.

PR #3300, opened 2024-01-02
Fix two overflows caught when testing Frontier - First Encounters

kcgen:
# Description

A couple overflows were flagged when testing Frontier - First Encounters with a debug and UBSAN build.

This first is a type casting issue in the `DIMULD` operation, and a second is a counter not being of sufficient size.

Suggest reviewing commits separately.

## Related issues

When compiling a debug sanitizer build with the recompiling core, this allows Frontier - First Encounters to generate the same DOS page fault-related crash on x86-64 hosts as it currently does on arm64 and PPC hosts (finally; at least we have a common behaviour now).

# Manual testing

```shell
meson setup -Dbuildtype=debug -Db_sanitize=undefined \
            -Ddynamic_core=dynrec build
meson compile -C build
```

Tested the above debug + UBSAN build with the entire benchmark pack, including LINPACK and Frontier - First Encounters.

If the same build is run with Frontier - First Encounters, it will report a UBSAN overflow and terminate with out-of-bounds `check_cast` assertion.

# Checklist

I have:

### DIFF
--- a/src/cpu/core_full/string.h
+++ b/src/cpu/core_full/string.h
@@ -27,7 +27,14 @@
 	uint32_t si_index       = is_32bit_addr ? reg_esi : reg_si;
 	uint32_t di_index       = is_32bit_addr ? reg_edi : reg_di;
 
-	int32_t count = check_cast<int32_t>(is_32bit_addr ? reg_ecx : reg_cx);
+	// Count has to be large enough to both hold an unsigned 32-bit
+	// value and also be used as a signed counter. To test the
+	// maximum, press enter during Frontier First Encounter's intro
+	// sequence on an arm64 or PPC platform, in which case all bits
+	// of ECX will be set.
+	//
+	int64_t count = is_32bit_addr ? reg_ecx : reg_cx;
+
 	int32_t count_left = 0;
 
 	if (!(inst.prefix & PREFIX_REP)) {
--- a/src/cpu/instructions.h
+++ b/src/cpu/instructions.h
@@ -787,18 +787,22 @@
 
 #define DIMULD(op1, op2, op3, load, save) \
 	{ \
-		const auto res = static_cast<int64_t>(op2) * \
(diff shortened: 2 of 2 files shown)


johnnovak:
Nice job @kcgen . Please help me understand this a bit better:

- Without the fix, the x86 release builds were just fine, but the ARM and PPC builds crashed when hitting this bug.
- The UBSAN build somehow allows us to trigger the crash on x86 too.
- This fixes the game for ARM and PPC. No user-observable changes for the x86 release build as the game has always been fine without this fix on x86.

Does this accurately reflect the situation?

kcgen:
> Please help me understand this a bit better:
> 
> - Without the fix, the x86 release builds were just fine, but the ARM and PPC builds crashed when hitting this bug.

Yes.

> - The UBSAN build somehow allows us to trigger the crash on x86 too.

Yes, the UBSAN build (and the assert) both are triggered when using the dynrec core on x86, so this only happens when building on x86 with the slower dynrec core (as opposed to the default x86-64 dynamic core, which has no problem running the game).


> - This fixes the game for ARM and PPC.

It unfortunately doesn't fix the game, but the benefit os the game now gets further (past the assert crash, when using a debug + UBSAN build) and triggers a DOS page fault, which ultimately produce a backtrack.

The good news is this DOS page fault crash is now also seen when using the dynrec core on 86 hosts, too. So it provides a common baseli6for anyone investigating this page fault issue (a fix should fix it across the board). 

But even if we ignored this normalizing benefit in First Encounter, these are both just bugs straight away.

> No user-observable changes for the x86 release build as the game has always been fine without this fix on x86.

Yup: under the default dynamic core, the game was fine and still is fine. 

> 
> Does this accurately reflect the situation?

Yup!

johnnovak (approved):
Nice one @kcgen . Thanks for the extra explanation 😄

PR #3301, opened 2024-01-03
Improve Sound Blaster DSP write buffer status IO (02x0Ch) logic

### LINKED ISSUE #3212: Digitized sounds do not play in SimLife

_No response_

Windows 10

### Relevant hardware info

_No response_

* [x]  I have searched and not found similar issues.

Digital sound will not play when running the game in Dosbox-Staging. The music plays fine, but the digital sound effects (easily heard when manually placing plants or animals, for instance) will not. The game's quick start guide mentions the following:
sounds
Changing the cputype and increasing cycles doesn't seem to make a difference. The game does not support Sound Blaster 16, but it is compatible with earlier Sound Blaster models. I've set the game up with Dosbox-X, and the digitized sounds work perfectly fine- the only configuration change I make is setting sbtype to sbpro2.

Explain how to reproduce

1. Configure sbtype to a model that the game supports, such as sbpro2 or earlier
2. Install and play the DOS version of the game and check if digital sound effects are working. The quickest and easiest way to check is to manually place animals or plants (the in-game tutorial explains how). There should be a sound effect every time you manually place one. 
3. Follow the above steps in Dosbox-X to hear the digital sound effects playing properly

_No response_

### Your configuration

```ini
https://gist.github.com/kadaj29/1a96b0de659816b00bf02f14624c1fbf
```

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
# Description

The Sound Blaster's 02x0Ch read IO call checks to see if the DSP can support more data writes.

In the baseline code, repeated reads from 02x0Ch use a counter that flip-flops in batches of eight between _"we're full, we can't accept data"_ and _"yes, we can accept data"_.

This behaviour causes SimLife and Scali's modified the Crystal Dream demo to eventually give up on setting up their DMA transfers.

This PR split this eight on/off logic:
1. The `off` part is kept; we now report available room every 8th call.
2. The `on` part is replaced with an actual check of the DMA buffer state.

So this replaces the _"we're busy, we're busy, we're busy, [x8]"_ with
an actual check of the DMA status.. and if DMA isn't running or
at the minimum (about to run dry), then we say _"we're available"_.

Suggest reviewing commit-by-commit. I've taken care to make sure each type of change is isolated one by one, so we can bisect to the specific issue.

## Related issues

Fixes #3212 

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/bd6f2db2-c2c3-4bf0-993e-f8b91932e3c0

# Manual testing

Regression tested:
- Alone in the Dark
- Albion
- Crystal Dream (Scali's DMA-improved version)
- DOOM
- DOOM II
- Duke Nukem 2
- Freddy Pharkas
- HoMM I
- Impulse Tracker
- Jet Pack
- Mortal Kombat 2
- OCP Player
- Overdrive
- Pinball Dreams
- Scream Tracker 3
- Sim City 2000
- SimLife
- Tale of Peter Rabbit
- Terminal Velocity
- The Rocketeer
- Prehistorik 2
- Blue Brothers 2

# Checklist

I have:

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -159,7 +159,7 @@ struct SB_INFO {
 			uint8_t used = 0; // number of entries in the fifo
 		} in = {}, out = {};
 		uint8_t test_register = 0;
-		uint32_t write_busy = 0;
+		uint8_t write_status_counter = 0;
 		uint32_t reset_tally = 0;
 		uint8_t cold_warmup_ms = 0;
 		uint8_t hot_warmup_ms = 0;
@@ -1225,7 +1225,7 @@ static void DSP_Reset() {
 	sb.dsp.cmd=DSP_NO_COMMAND;
 	sb.dsp.cmd_len=0;
 	sb.dsp.in.pos=0;
-	sb.dsp.write_busy=0;
+	sb.dsp.write_status_counter = 0;
 	sb.dsp.reset_tally++;
 	PIC_RemoveEvents(DSP_FinishReset);
 
@@ -1981,11 +1981,30 @@ static bool write_buffer_at_capacity()
 	if (sb.dsp.state != DSP_S_NORMAL) {
 		return true;
 	}
-	// Use a counter to report the buffer is unable to accept data for eight
-	// consecutive calls followed by reporting we *are* able receive data
-	// for another eight calls (flip-flop every 8 calls).
-	sb.dsp.write_busy++;
-	return (sb.dsp.write_busy & 8) != 0;
+	// Report the buffer as having some room every 8th call, as the buffer
+	// will have run down by some amount after sequential calls.
+	if ((++sb.dsp.write_status_counter % 8) == 0) {
+		return false;
+	}
+	// If DMA isn't running then the buffer's definitely not at capacity.
+	if (sb.dma.mode == DSP_DMA_NONE) {
+		return false;
+	}
+	// Finally, the DMA buffer is considered full until it's able to accept
+	// a full (64 KB) write; which is once we've hit the minimum threshold.
+	//
+	// Notes:
+	// 86Box and DOSBox-X both calculate the playback rate of the current
+	// DMA transfer and then steadily draining down that time until it nears
+	// completion. One of the nuances is that games can change the playback
+	// rate as well as switch from mono to stereo; so the drain down rate
+	// can vary: and indeed, 86Box does this extra bookkeeping.
+	//
+	// In our case, these rate changes are already taken care of by the
+	// existing code, and it just happens that this threshold (dma.left vs
+	// dma.min) will happen sooner if the rates are faster.
+	//
+	return (sb.dma.left > sb.dma.min);
 }
 
 static uint8_t read_sb(io_port_t port, io_width_t)
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -28,6 +28,7 @@
 
 #include "autoexec.h"
 #include "bios.h"
+#include "bit_view.h"
 #include "channel_names.h"
 #include "control.h"
 #include "dma.h"
@@ -2007,25 +2008,53 @@ static bool write_buffer_at_capacity()
 	return (sb.dma.left > sb.dma.min);
 }
 
+// Sound Blaster DSP status byte
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// Ref: http://archive.gamedev.net/archive/reference/articles/article443.html
+//
+// Read from 02x0Ch (DSP Write Buffer Status): Bit 7, when 0, indicates that the
+// DSP is ready to receive data through the DSP Write Data or Command port
+// (02x0Ch write).
+//
+// Read from 02x0Eh (DSP Data Available Status): Bit 7, when 1, indicates that
+// the DSP has pending data to be read through the DSP Read Data port (02x0Ah
+// read).
+//
+union BufferStatus {
+	// Default to all bits high
+	uint8_t data = 0b1111'1111;
+
+	// Unused bits. They appear to be set to 1 when reading the status
+	// register.
+	bit_view<0, 7> reserved;
+
+	// when 1, the buffer has data and is ready to send
+	// when 0, the buffer doesn't have data and is ready to receive
+	bit_view<7, 1> has_data;
+};
+
 static uint8_t read_sb(io_port_t port, io_width_t)
 {
 	switch (port - sb.hw.base) {
 	case MIXER_INDEX: return sb.mixer.index;
 	case MIXER_DATA: return CTMIXER_Read();
 	case DSP_READ_DATA: return DSP_ReadData();
-	case DSP_READ_STATUS:
+	case DSP_READ_STATUS: {
 		// TODO See for high speed dma :)
 		if (sb.irq.pending_8bit) {
 			sb.irq.pending_8bit = false;
 			PIC_DeActivateIRQ(sb.hw.irq);
 		}
-		if (sb.dsp.out.used) {
-			return 0xff;
-		} else {
-			return 0x7f;
-		}
+		BufferStatus read_status = {};
+		read_status.has_data = (sb.dsp.out.used != 0);
+		return read_status.data;
+	}
 	case DSP_ACK_16BIT: sb.irq.pending_16bit = false; break;
-	case DSP_WRITE_STATUS: return write_buffer_at_capacity() ? 0xff : 0x7f;
+	case DSP_WRITE_STATUS: {
+		BufferStatus write_status = {};
+		write_status.has_data = write_buffer_at_capacity();
+		return write_status.data;
+	}
 	case DSP_RESET: return 0xff;
 	default:
 		LOG(LOG_SB, LOG_NORMAL)("Unhandled read from SB Port %4X", port);
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -1973,6 +1973,21 @@ static uint8_t CTMIXER_Read() {
 	return ret;
 }
 
+// Called by DSP_WRITE_STATUS to check if the write buffer is at capacity.
+static bool write_buffer_at_capacity()
+{
+	// Is the DSP in an abnormal state and therefore we should consider the
+	// buffer at capacity and unable to receive data?
+	if (sb.dsp.state != DSP_S_NORMAL) {
+		return true;
+	}
+	// Use a counter to report the buffer is unable to accept data for eight
+	// consecutive calls followed by reporting we *are* able receive data
+	// for another eight calls (flip-flop every 8 calls).
+	sb.dsp.write_busy++;
+	return (sb.dsp.write_busy & 8) != 0;
+}
+
 static uint8_t read_sb(io_port_t port, io_width_t)
 {
 	switch (port - sb.hw.base) {
@@ -1991,18 +2006,7 @@ static uint8_t read_sb(io_port_t port, io_width_t)
 			return 0x7f;
 		}
 	case DSP_ACK_16BIT: sb.irq.pending_16bit = false; break;
-	case DSP_WRITE_STATUS:
-		switch (sb.dsp.state) {
-		case DSP_S_NORMAL:
-			sb.dsp.write_busy++;
-			if (sb.dsp.write_busy & 8) {
-				return 0xff;
-			}
-			return 0x7f;
-		case DSP_S_RESET:
-		case DSP_S_RESET_WAIT: return 0xff;
-		}
-		return 0xff;
+	case DSP_WRITE_STATUS: return write_buffer_at_capacity() ? 0xff : 0x7f;
 	case DSP_RESET: return 0xff;
 	default:
 		LOG(LOG_SB, LOG_NORMAL)("Unhandled read from SB Port %4X", port);
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -1980,31 +1980,32 @@ static uint8_t read_sb(io_port_t port, io_width_t)
 	case MIXER_DATA: return CTMIXER_Read();
 	case DSP_READ_DATA: return DSP_ReadData();
 	case DSP_READ_STATUS:
-		//TODO See for high speed dma :)
-		if (sb.irq.pending_8bit)  {
-			sb.irq.pending_8bit=false;
+		// TODO See for high speed dma :)
+		if (sb.irq.pending_8bit) {
+			sb.irq.pending_8bit = false;
 			PIC_DeActivateIRQ(sb.hw.irq);
 		}
-		if (sb.dsp.out.used) return 0xff;
-		else return 0x7f;
-	case DSP_ACK_16BIT:
-		sb.irq.pending_16bit=false;
-		break;
+		if (sb.dsp.out.used) {
+			return 0xff;
+		} else {
+			return 0x7f;
+		}
+	case DSP_ACK_16BIT: sb.irq.pending_16bit = false; break;
 	case DSP_WRITE_STATUS:
 		switch (sb.dsp.state) {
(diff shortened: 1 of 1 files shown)


Grounded0 (approved):
If it works and at least a dozen games have been tested not to cause regressions it should be good enough.

src/hardware/sblaster.cpp
```diff
@@ -1995,25 +1996,47 @@ static bool write_buffer_at_capacity()
...
 		}
-		if (sb.dsp.out.used) {
-			return 0xff;
-		} else {
-			return 0x7f;
-		}
+		BufferStatus read_status = {};
+		read_status.has_data = (sb.dsp.out.used != 0);
+		return read_status.data;
+	}
```
kklobe, line 2033: are we officially OK with type punning in cases like this?

Grounded0:
I would still push this into 0.82 since this is a long standing issue and we mainly want to fix regressions in the RC cooldown period.

src/hardware/sblaster.cpp
kcgen: (it's how all the `bit_view` unions work.. if we're not OK with then, there's a lot we're not OK with :-)

`bit_view` is just a vehicle to operate the data `.data` member:

```c++
// all-purpose assignment
template <class rhs_type>
constexpr bit_view& Assign(const rhs_type& rhs_value) noexcept
{
	if constexpr (std::is_integral_v<rhs_type>) {
		check_rhs(rhs_value);
	}
	const auto outer = data & ~mask::shifted;
	const auto inner = (rhs_value & mask::unshifted) << view_index;

	data = static_cast<data_type>(outer | inner);
	return *this;
}
```

So the write goes straight to `.data` and then we read `.data` (in this case).

But same with reading bits; they all boil down to reads on `.data`:

```c++
// read the view's value
constexpr operator data_type() const noexcept
{
return (data & mask::shifted) >> view_index;
}
```

kcgen: @kklobe - to eliminate the union, maybe `<variant>` can be used under the hood? I don't know enough about it though.

kklobe: well - `memcpy` is one way around it 😁. C++20 will give us `bit_cast`, so maybe we just let it ride for now.

kcgen: (I should also say that the `bit_view` unit tests pass GCC and Clang's undefined behaviour sanitizers on Linux and macOS; so I think this is pretty much solid until we want to swap it out :slightly_smiling_face: !)

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
+		return false;
+	}
+	// Finally, the DMA buffer is considered full until it's able to accept
+	// a full (64 KB) write; which is once we've hit the minimum threshold.
+	return (sb.dma.left > sb.dma.min);
+}
+
+// Sound Blaster DSP status byte
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// Ref: http://archive.gamedev.net/archive/reference/articles/article443.html
```
johnnovak, line 2001: Great find! We should really start collecting these text files, PDFs, website dumps, eBooks, etc. that contain important hardware related information in a "library". A kitchen-sink item on archive.org would suffice.

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
+//
+// Read from 02x0Eh (DSP Data Available Status): Bit 7, when 1, indicates that
+// the DSP has pending data to be read through the DSP Read Data port (02x0Ah
+// read).
+//
+union BufferStatus {
+	uint8_t data = 0;
+	bit_view<7, 1> has_data;
+	// when 1, the buffer has data and is ready to send
+	// when 0, the buffer doesn't have data and is ready to receive
```
johnnovak, line 2015: Can you put the explanation of the bit _above_ it? That's pretty standard I think, and I've been doing it like that for other register bitview definitions.

johnnovak: The linked doco says the low 7 bits are reserved. So I suggest doing this to make the definition complete (e.g. i was wondering, are the other bits unmapped because we're not using them? or are they reserved/undefined?)


```cpp
// Unused bits. They appear to be set to 1 when reading the status register.
bit_view<0, 7> reserved;
```

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
-		case DSP_S_RESET:
-		case DSP_S_RESET_WAIT:
-			return 0xff;
-		}
-		return 0xff;
-	case DSP_RESET:
-		return 0xff;
+		BufferStatus read_status = {};
+		read_status.has_data = (sb.dsp.out.used != 0);
+		return read_status.data;
```
johnnovak, line 2032: So in the previous implementation the reserved low 7 bits were always set to 1. I highly recommend to keep that behaviour. Maybe that's how the hardware behaves (or certain revisions of the hardware 😄 ), and maybe some programs simplistically check for equality with `0xff` and `0x7f`. This change would break those program.

It would be extremely hard to prove this, given we a) don't have access to all the historic SB hardware, b) there are no comments explaining this in the code, c) maybe only a few programs out of hundreds are affected.

Best to keep the existing behaviour, there's no harm in that. Just add a comment please explaining the above succinctly 😄

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
+	}
+	case DSP_ACK_16BIT: sb.irq.pending_16bit = false; break;
+	case DSP_WRITE_STATUS: {
+		BufferStatus write_status = {};
+		write_status.has_data = write_buffer_at_capacity();
+		return write_status.data;
+	}
+	case DSP_RESET: return 0xff;
```
johnnovak, line 2040: I'd do this here, more self-documenting:

```cpp
BufferStatus status = {};
status.flip();
return status;
```

(Also, without setting the low 7 bits to 1 as suggested, this is a discrepancy because the low 7 bits _are_ set in the reset status response.)

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
 	default:
-		LOG(LOG_SB,LOG_NORMAL)("Unhandled read from SB Port %4X",port);
+		LOG(LOG_SB, LOG_NORMAL)("Unhandled read from SB Port %4X", port);
 		break;
 	}
 	return 0xff;
```
johnnovak, line 2045: ``cpp
BufferStatus status = {};
status.flip();
return status;
```

Comment (johnnovak) [reply src/hardware/sblaster.cpp:2045]: Another thought. I find sporadic returns in the middle of the function confusing as it obscures the flow. Could you refactor this so this is at the very top:

```cpp
BufferStatus status = {};
status.flip();
```

Then the `has_data` bit gets flipped only where necessary, and `status` is simply returned at the end? That would make things cleaner.

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
+	if ((++sb.dsp.write_status_counter % 8) == 0) {
+		return false;
+	}
+	// If DMA isn't running then the buffer's definitely not at capacity.
+	if (sb.dma.mode == DSP_DMA_NONE) {
+		return false;
+	}
+	// Finally, the DMA buffer is considered full until it's able to accept
+	// a full (64 KB) write; which is once we've hit the minimum threshold.
+	return (sb.dma.left > sb.dma.min);
```
johnnovak, line 1996: Are these new checks and conditions based on someone's reverse engineering work, or have you tried to do what's logical here @kcgen? If the latter, it's probably wrong as the behaviour of most hardware can be hardly derived from first principles... 😅 The whole SB behaviour is a minefield too; the TL;DR is only trust actual measurements, the doco is lying, etc.

I'm not saying I'm outright _against_ these extra checks, but try not to introduce too many ad-hoc constraints based on logic alone, that's my suggestion. I'd remove anything that's not absolutely essential, even if that means the buffer gets sometime ovewritten, etc. We just don't know how hardware works until we measure it.

I'm also confused about that modulo 8 thing (hack?) Do we know where that's coming from? Why not 11th call? 😅 Or why not report the actual reality? So, lots of questions...

I'm just cautious because when I screwed up a few things in my VGA refactor (single bits set wrong in very specific cases), it takes many months and a lot of testing for problems to surface (e.g. the Kyrandia scrolling bug). So all I'm saying, I highly suggest to be extremely conservative with these improvements 😄

johnnovak:
Very nice improvement @kcgen, I'm glad you managed to get the sound working in those problematic games 🎉 I might appear extremely nitpicky in my comments, but mucking around with the VGA stuff has taught me to be very conservative with any improvements & changes that "make sense". Hardware usually don't abide by those rules of logic 😅 I guess people did all sorts of cheap tricks to save a few transistors here and there which resulted in... funky behaviour 😅

src/hardware/sblaster.cpp
kcgen: :+1: ; we can default to all bits high.

src/hardware/sblaster.cpp
kcgen: Good call and agree - I'm now defaulting this with all the bits high:

```c++
union BufferStatus {
	// Default to all bits high
	uint8_t data = 0b1111'1111;

	// Unused bits. They appear to be set to 1 when reading the status register.
	bit_view<0, 7> reserved;

	// when 1, the buffer has data and is ready to send
	// when 0, the buffer doesn't have data and is ready to receive
	bit_view<7, 1> has_data;
};
```

src/hardware/sblaster.cpp
```diff
@@ -1973,38 +1974,72 @@ static uint8_t CTMIXER_Read() {
...
-	case DSP_RESET:
-		return 0xff;
+		BufferStatus read_status = {};
+		read_status.has_data = (sb.dsp.out.used != 0);
+		return read_status.data;
+	}
+	case DSP_ACK_16BIT: sb.irq.pending_16bit = false; break;
+	case DSP_WRITE_STATUS: {
+		BufferStatus write_status = {};
+		write_status.has_data = write_buffer_at_capacity();
```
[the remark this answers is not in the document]
kcgen: (default high bits will take care of it)

src/hardware/sblaster.cpp
kcgen: (default high bits will take care of it)

src/hardware/sblaster.cpp
kcgen: This return value doesn't involve the buffer status (same with a couple of the other DSP reset IO reads.. they're just checking if the DSP was reset, not on the buffer state).

src/hardware/sblaster.cpp
johnnovak: Looking great! Good call on the defaulting, nice technique! 🎉

src/hardware/sblaster.cpp
kcgen: 100%!

Lots of great info out there that turns up whenever all of us start digging into something; I do also fear it might be lost!

src/hardware/sblaster.cpp
johnnovak: Yeah I realised later this function is a "generic" I/O port handler, my bad 😄

src/hardware/sblaster.cpp
kcgen: > I'm also confused about that modulo 8 thing (hack?) Do we know where that's coming from? Why not 11th call? 😅 Or why not report the actual reality? So, lots of questions...

Yes - I think it was a reasonable heuristic or perhaps someone measured it from their actual card. Unfortunately I'm not sure; but it wasn't perfect because games like SimLife simply give up if they're told they can't send data despite knowing the card is totally idle (probably 11 would be worse :laughing: )

We can see some of this in PaybackTime 2's pascal code: it does these door knocks 1000 times and then skips.

> re these new checks and conditions based on someone's reverse engineering work, or have you tried to do what's logical here @kcgen?

I created this based on logging what these games were doing (just a bunch of LOGs in SB) combined with reading that about what this check routine is meant to represent.

86Box and DOSBox-X both calculate the playback rate of the current DMA transfer and then steadily draining down that time until it nears completion.

In this implementation, fortunately we've already got the `dma.left` and `dma.min` values to see how close we are to the end of the transfer.

One of the nuances is that games can change the playback rate as well as switch from mono to stereo; so the drain down rates vary: and indeed, 86Box does this extra bookkeeping.

In our case, these rate changes are already taken care of by the existing code, and it just happens that this threshold (`dma.left` vs `dma.min`) will just happen sooner if the rates are faster.

Scali had a through discussion about Sound Blaster seamless playback using all the different DMA methods (autoinit, signle init, and hacked methods) https://www.vogons.org/viewtopic.php?t=52806. He wrote a nice seamless playback tool, too (tested this PR with it, and it's fine).

> I'm just cautious

Absolutely; given my track record with regression, there's probably a regression /somewhere/ with /some/ game! :laughing: and I've planned for that by committing one change at a time, so it can be isolated.  

But so far, it seems to be working quite well - and gets SimLife working, too. It's possible other games will be fixed.



I know there some efforts to disassemble a couple of the SB DSPs out there -

johnnovak: Thanks for the explanation, you've certainly done your homework then 😅 📓 🖊️  Would it be worthwhile to include some pointers about all that in a few sentences? At least a link to the Vogons thread would be very helpful for the next person deep-investigating the SB DMA situation.

johnnovak (approved):
Stellar work @kcgen! Merge when you're ready, man 😎

src/hardware/sblaster.cpp
johnnovak: > are we officially OK with type punning in cases like this?

Yes 😎 (short and sweet)

...what @kcgen said, keep the end goal in mind. Bit views are super useful and increase the clarity of the code a great deal. That's what we should optimise for, whatever it takes, so I'm 100% fine.

kklobe: Yes, but it's UB, so I'm clarifying we are ok with that.

kcgen: I don't think it's UB because under the hood, we always write to the same union member (the data member) as we read from.

Provided those are always done back to back, it's not UB.

@kklobe, are you getting any compiler or sanitizer warnings from it?

kklobe: Hmm, interesting, It's still a union, but given the slight of hand, I wonder if it would fall outside of https://en.cppreference.com/w/cpp/language/union#Explanation. I don't get compiler warnings, but none of the other non-bitview instances in the codebase throw warnings either, so it clearly must not be a very big deal.

kcgen: Could it be that it exclusives uses `uint8_t`'s? It never moves from one type to another (ie: write a 32-bit to a union `float` but then read those bits back as a union-punned `int32_t`… that's definitely UB, I think).

It's used throughout the code now, and passes Coverity and PVS Studio too (as well as sanitizer builds).

Undefined behaviour is unwanted behaviour; absolutely - I want Staging passing all the sanitizers. And certainly - if there's a better way to get the job done, let's do it!

src/hardware/sblaster.cpp
kcgen: Will add some of that, and add most of it to the commit comment block (so it's easy to "git log -S")!

kcgen: Ok - added!

src/hardware/sblaster.cpp
kcgen: Sounds like a plan; thanks @kklobe.

PR #3310, opened 2024-01-08
Add a simple DC bias remover for the OPL output

### LINKED ISSUE #963: Wizardry 6+7 sound fix?
Just found a commit in another fork that works around a supposed bug in Wizardry 6 and 7: https://github.com/unixnation/Boxer/commit/c4a0ba6e814a4158c2ccde3040368bfce1322233
I'm mainly curious if this issue is still present, and if the patch is worth integrating.

### LINKED ISSUE #3306: Golden Eagle (1991)

81.0a

Windows 11

_No response_

### Relevant hardware info

_No response_

When you select Adlib with this game during its startup and have sbtype=none with oplmode set in the conf the game gives some extra fx which is good but it does it with some hideous popping at the start of every effect. The best examples are during the intro when the men are walking across the screen, every single step is accompanied by this pop

Explain how to reproduce
1. Set oplmode and sbtpe=none in conf
2. Choose Adlib during game start
3. Watch intro

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

kcgen:
# Description

Some games like Golden Eagle (1991) play PCM music and effects using the Adlib OPL channels by rapidly changing the volume in very crude steps, similar to Disney or the Covox LPT DACs.

In this case, the game uses a large DC bias which produces a heavy pop around the effect sequences:

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/f86999c8-e4db-4b9f-aca0-304b844cd6f5

---

This PR adds an opt-in feature that eliminates the OPL DC bias, which eliminates the pops:

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/31b1317d-206e-4ff8-8d86-723ba2f1874a

We could parameterize the settings if more games can be served by this, and if they benefit from other settings. For now, let's not over-engineer it until we need to.

Vogons has a thread that lists some (all?) of these games:

https://www.vogons.org/viewtopic.php?f=7&t=50525

## Related issues

Fixes #3306.

Fixes #963.

Recommended settings for Wizardry 6:

```ini
[dosbox]
machine = ega

[cpu]
cycles = 1000

[sblaster]
sbtype = none
oplmode = opl2
opl_remove_dc_bias = true
opl_filter = lpf 2 5500

[speaker]
pcspeaker=off

[autoexec]
@mount c ..
c:
cd bane
mixer opl 500
```

# Manual testing

This is a default-off (opt-in) feature, and it works as expected in this one games.  Did a couple regression tests to confirm it's not active with default settings.

# Checklist

I have:

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -821,6 +821,14 @@ void DOSBOX_Init()
 	        "                300 200 (Wait 300ms before fading out over a 200ms period)\n"
 	        "                1000 3000 (Wait 1s before fading out over a 3s period)");
 
+	pbool = secprop->Add_bool("opl_remove_dc_bias", when_idle, false);
+	pbool->Set_help(
+	        "Remove DC bias from the OPL output. This should only be used as a last resort\n"
+	        "to fix popping in games that play PCM audio using the OPL synthesiser on a\n"
+	        "Sound Blaster or AdLib card, such as in: Golden Eagle (1991), Wizardry 6\n"
+	        "(1990), and Wizardry 7 (1992). Please open an issue ticket if you find other\n"
+	        "affected games.");
+
 	pstring = secprop->Add_string("oplemu", deprecated, "");
 	pstring->Set_help("Only 'nuked' OPL emulation is supported now.");
 
--- a/src/hardware/opl.cpp
+++ b/src/hardware/opl.cpp
@@ -22,6 +22,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <memory>
+#include <queue>
 #include <sys/types.h>
 
 #include "../capture/capture.h"
@@ -511,10 +512,39 @@ io_port_t OPL::WriteAddr(const io_port_t port, const uint8_t val)
 	return addr;
 }
 
+template <LineIndex line_index>
+int16_t remove_running_average(const int16_t sample)
+{
+	// Calculate the number of samples we need average across to maintain
+	// the lowest frequency given an assumed playback rate.
+	constexpr int16_t PcmPlaybackRateHz      = 16000;
+	constexpr int16_t LowestFreqToMaintainHz = 200;
+	constexpr int16_t NumToAverage = PcmPlaybackRateHz / LowestFreqToMaintainHz;
+
+	static int32_t sum = 0;
+	static std::queue<int16_t> samples = {};
+
+	sum += sample;
+	samples.push(sample);
+
+	int16_t average = 0;
+	if (samples.size() == NumToAverage) {
+		average = static_cast<int16_t>(sum / NumToAverage);
+		sum -= samples.front();
+		samples.pop();
+	}
+	return static_cast<int16_t>(sample - average);
+}
+
 AudioFrame OPL::RenderFrame()
 {
 	static int16_t buf[2] = {};
 	OPL3_GenerateStream(&oplchip, buf, 1);
+
+	if (ctrl.wants_dc_bias_removed) {
+		buf[0] = remove_running_average<Left>(buf[0]);
+		buf[1] = remove_running_average<Right>(buf[1]);
+	}
 
 	AudioFrame frame = {};
 	if (adlib_gold) {
@@ -922,6 +952,11 @@ OPL::OPL(Section *configuration, const OplMode oplmode)
 
 	// Setup fadeout
 	channel->ConfigureFadeOut(section->Get_string("opl_fadeout"));
+
+	ctrl.wants_dc_bias_removed = section->Get_bool("opl_remove_dc_bias");
+	if (ctrl.wants_dc_bias_removed) {
+		LOG_MSG("%s: DC bias removal enabled", channel->GetName().c_str());
+	}
 
 	Init(check_cast<uint16_t>(channel->GetSampleRate()));
 
--- a/src/hardware/opl.cpp
+++ b/src/hardware/opl.cpp
@@ -513,7 +513,7 @@ io_port_t OPL::WriteAddr(const io_port_t port, const uint8_t val)
 }
 
 template <LineIndex line_index>
-int16_t remove_running_average(const int16_t sample)
+int16_t remove_dc_bias(const int16_t back_sample)
 {
 	// Calculate the number of samples we need average across to maintain
 	// the lowest frequency given an assumed playback rate.
@@ -524,16 +524,28 @@ int16_t remove_running_average(const int16_t sample)
 	static int32_t sum = 0;
 	static std::queue<int16_t> samples = {};
 
-	sum += sample;
-	samples.push(sample);
+	// Clear the queue if the stream isn't biased
+	constexpr int16_t BiasThreshold = 5;
+	if (back_sample < BiasThreshold) {
+		sum = 0;
+		samples = {};
+		return back_sample;
+	}
 
-	int16_t average = 0;
+	// Keep a running sum and push the sample to the back of the queue
+	sum += back_sample;
+	samples.push(back_sample);
+
+	int16_t average      = 0;
+	int16_t front_sample = 0;
 	if (samples.size() == NumToAverage) {
+		// Compute the average and deduct it from the front sample
 		average = static_cast<int16_t>(sum / NumToAverage);
-		sum -= samples.front();
+		front_sample = samples.front();
+		sum -= front_sample;
 		samples.pop();
 	}
-	return static_cast<int16_t>(sample - average);
+	return static_cast<int16_t>(front_sample - average);
 }
 
 AudioFrame OPL::RenderFrame()
@@ -542,8 +554,8 @@ AudioFrame OPL::RenderFrame()
 	OPL3_GenerateStream(&oplchip, buf, 1);
 
 	if (ctrl.wants_dc_bias_removed) {
-		buf[0] = remove_running_average<Left>(buf[0]);
-		buf[1] = remove_running_average<Right>(buf[1]);
+		buf[0] = remove_dc_bias<Left>(buf[0]);
+		buf[1] = remove_dc_bias<Right>(buf[1]);
 	}
 
 	AudioFrame frame = {};
--- a/src/hardware/opl.h
+++ b/src/hardware/opl.h
@@ -130,6 +130,7 @@ private:
 
 		bool active = false;
 		bool mixer  = false;
+		bool wants_dc_bias_removed = false;
 	} ctrl = {};
 
 	void Init(const uint16_t sample_rate);


src/hardware/opl.cpp
```diff
@@ -511,11 +512,50 @@ io_port_t OPL::WriteAddr(const io_port_t port, const uint8_t val)
...
+	samples.push_back(in_sample);
+	if (samples.size() < BiasBlockSize) {
+		return {};
+	}
+
+	// Compute the running bias
+	int32_t sum = 0;
+	for (const auto s : samples) {
+		sum += s;
+	}
```
johnnovak, line 539: Probably wouldn't make a huge difference, but you could use a moving average approach instead. So you wouldn't need to average 2 * 16 samples on each invocation. Example implementtaion:

https://www.zrzahid.com/moving-average-of-last-n-numbers-in-a-stream/

johnnovak (approved):
That's a nice practical solution @kcgen. This is fine, or you could refine the calculation a bit to use moving averages—entirely up to you.

src/hardware/opl.cpp
kcgen: Thanks for the tip; will swap this in tomorrow 👍

johnnovak: I think you could also get rid of the conditionals completely and make it branchless. More for fun & simplicity than the actual performance benefits 😄 Just maintain a running average of the DC offset that starts from zero, taking the last N=16 samples into account.

kcgen: Replaced it with a simple running average:

```c++
int16_t remove_running_average(const int16_t sample)
{
	// Calculate the number of samples we need average across to maintain
	// the lowest frequency given an assumed playback rate.
	constexpr int16_t PcmPlaybackRateHz      = 16000;
	constexpr int16_t LowestFreqToMaintainHz = 200;
	constexpr int16_t NumToAverage = PcmPlaybackRateHz / LowestFreqToMaintainHz;

	static int32_t sum = 0;
	static std::queue<int16_t> samples = {};

	sum += sample;
	samples.push(sample);

	int16_t average = 0;
	if (samples.size() == NumToAverage) {
		average = static_cast<int16_t>(sum / NumToAverage);
		sum -= samples.front();
		samples.pop();
	}
	return static_cast<int16_t>(sample - average);
```

However, a bunch of the pops remain:

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/83e95e2e-91f1-4ed0-b8d7-4623558df640

I've added this running-average approach as the first commit (just for the record), and then modified it in the second commit with the original algorithm.

kcgen:
Queued for merging till 0.81.x is done with :)

src/hardware/opl.cpp
kcgen: (resolving; the running average approach is now in the original algorithm)

src/dosbox.cpp
```diff
@@ -821,6 +821,12 @@ void DOSBOX_Init()
 	        "                300 200 (Wait 300ms before fading out over a 200ms period)\n"
 	        "                1000 3000 (Wait 1s before fading out over a 3s period)");
 
+	pbool = secprop->Add_bool("opl_remove_dc_bias", when_idle, false);
+	pbool->Set_help(
+	        "Remove DC bias from the OPL output. This should only be used as a last resort\n"
+	        "to fix popping in games that play PCM using the Adlib card, such as is the case\n"
```
johnnovak, line 827: Minor clarification suggestion: "that play PCM audio using the OPL synthesiser on a Sound Blaster or AdLib card"

src/dosbox.cpp
```diff
@@ -821,6 +821,12 @@ void DOSBOX_Init()
...
+	        "in Golden Eagle (1991). Please open a bug report if you find other affected games.");
```
johnnovak, line 828: nitpicking: "open an issue ticket" ?

johnnovak:
Nice one @kcgen. I'm wondering if it would make sense to extract this into `dc_remover.cpp` and apply it to other audio devices as well that can generate unipolar signals? I think the PC speaker qualifies when it's used to output PCM streams via various tricks. Mean Streets, Crime Wave, Pinball Dreams and all early Legend Entertainment adventures can all use the PC speaker for PCM output.

It might be also worthwhile to apply this to the Sound Blaster output to get rid of bias-induced clicking in some games. I haven't verified it, but I'm pretty sure Prehistorik 2 falls into this category, the ingame sound effects and even the music clicks like crazy on real hardware too. If it's bias-related, we could improve the situation there too.

Given the above considerations and that this must be a very low overhead process, would it make sense to use a single `remove_dc_offset` flag in the `[mixer]` section that would toggle DC removal for the Tandy, OPL, and Sound Blaster? I'm inclined to leave it on all the time, just like the master compressor and the high-pass filter, unless we can prove it does cause audible artifacts.

src/dosbox.cpp
kcgen: got it

src/dosbox.cpp
kcgen: added

johnnovak:
> If we did want a general approach, I think we'd want to offer the continuous moving average deduction approach for the other cases - but that's if we can find actual problem scenario that it fixes.

Thanks for checking. Looks like it's best to only apply this to specific interfaces as an opt-in fix then.

> Definitely tons of games play PCM, but they're not too bad for popping. It's really the popping that's the issue. Lots of games setup the bias and hold it the entire time, so eliminates the risk of popping.

> The good news is the bias is held and the audio stream is contiunous, so there isn't any popping in-game:

Yeah, and the high-pass filter gets rid of the constant DC offset completely. Really, the problem is unipolar signals with sudden jumps to/from zero.

marsej:
Somewhat related question:

With 0.81.0rc2 In Star Control 2 CD ver started using starcon2.com , with settings...: 
gus      = true
gusbase  = 240
gusirq   = 5
gusdma   = 5
ultradir = C:\ULTRASND
gus_filter = off


...there's audible pops , in Talking pet music (attached save, save slot 25). The question is, could this kind of fix be attempted to apply to the GUS as well?

[SC2MUSICSAVES.zip](https://github.com/dosbox-staging/dosbox-staging/files/14007828/SC2MUSICSAVES.zip)

johnnovak:
> Somewhat related question:
> 
> With 0.81.0rc2 In Star Control 2 CD ver started using starcon2.com , with settings...: gus = true gusbase = 240 gusirq = 5 gusdma = 5 ultradir = C:\ULTRASND gus_filter = off
> 
> ...there's audible pops , in Talking pet music (attached save, save slot 25). The question is, could this kind of fix be attempted to apply to the GUS as well?
> 
> [SC2MUSICSAVES.zip](https://github.com/dosbox-staging/dosbox-staging/files/14007828/SC2MUSICSAVES.zip)

Can you attach a short WAV recording, please?

marsej:
Here's the recordings, I found popping in 7 SC2CD songs. We tried pop removal before but it affected the sound quality of other songs negatively and caused regression on some other game so perhaps this time make it optional for testing first?

[sc2popping.zip](https://github.com/dosbox-staging/dosbox-staging/files/14041571/sc2popping.zip)

marsej:
The sounds played are still correct but the audibility of the pops has increased significantly, hear the attachment.

almost inaudible pops: 
encounter windows-msys2-x86_64-v0.79.0-alpha-166-gb6083.wav

other versions, very audible repeat popping: 
encounter*.wav

I also included HPF and LPF filtered versions, they have more audible pops than the v0.79.0-alpha-166-gb6083.

[EncounterPops.zip](https://github.com/dosbox-staging/dosbox-staging/files/14042883/EncounterPops.zip)

kcgen:
Thanks @marsej.
Interesting; unfortunately I'm not able to investigate, but perhaps this can be bisected eventually and the cause isolated, as this seems like a regression.

johnnovak:
I've listened to your recordings @marsej. Yes, I can hear some popping there. However, there's a good chance you'd get the same popping on real hardware too. Unless we can compare it to real hardware, we simply don't know.

Just putting this out there as we might be emulating the GUS 100% correctly and we're chasing some mythical "clickless GUS" dragon that never existed 😉  I have an AWE32 and in many games that support it, it _clicks like crazy_, man. Like you wouldn't believe it. If this small level of clicking annoys you, you'd be absolutely _livid_ when you hear the real AWE32 experience in some games 😅 

I know for a fact you had to do a lot of tricky coding to achieve 100% click-less playback on the GUS. It doesn't just happen when coding "naively". You'd get a lot of clicks in MOD players that did not use these tricky "de-clicking" methods, for example (which involve starting new samples on a brand new channel instead of abruptly stopping the current sample on the current channel and starting a new one, so you could easily end up using 16 channels to play back a 4 channel MOD file in a "clickless" way!)

---

So, with all this reality check out of the way, we're not against offering an improved "better than hardware" experience if we can. I will also _try_ to minimise clicking in my upcoming AWE32 emulation, because honestly, that clicking is not exactly desirable. But if it's too hard or impossible to do so, we'll just live with it. Hey, that's accurate emulation for you 😎

@kcgen I noticed you implemented some declicking improvement (which was of course not present in real hardware, that's a DOSBox Staging special feature), but you backed it out shortly thereafter.

I assume this envelope trickery improved the situation for Star Control 2, but only for a short time as you can see from the dates 😄 I tried to look up the reason for reverting this out later, but couldn't find any info. Maybe I suck at searching for this... So can you guys remember why this "Use a 12% expansion envelope for the GUS" was backed out? @shermp @kcgen 

```
commit 6b34088b65a2d8a45ea24cf338154ba9f7953d93
Author: kcgen <kcgen@users.noreply.github.com>
Date:   Thu Feb 17 23:32:45 2022 -0800

    Revert "Use a 12% expansion envelope for the GUS"

    This reverts commit 1099006c643dc531e10cf98c4f709fb04111918a.

    Co-authored-by: shermp <14854761+shermp@users.noreply.github.com>
    Co-authored-by: Kappa971 <Kappa971@users.noreply.github.com>

commit 1099006c643dc531e10cf98c4f709fb04111918a
Author: kcgen <kcgen@users.noreply.github.com>
Date:   Fri Jan 28 16:55:29 2022 -0800

    Use a 12% expansion envelope for the GUS

    The means that for a given channel, the next
    sample will be allowed to differ from the prior
    by up to 12% of the peak total volume .

    This was found to be sufficient to reduce
    popping in some GUS sequences, like Star
    Control 2.
```

kcgen:
@johnnovak , yup - eventually that maximum-difference limiter was found to cause some audible artifacts, so I got rid of it.

Yeah, unfortunately star control 2's internal mod player introduces small gaps and glitches in its PCM output - and this shows up in both GUS and SB playback. 

Newest Pleasures recorded the intro SC2 music on his physical GUS and SB carda, and just after the first round of drum hits you can hear the micro glitches.

http://nerdlypleasures.blogspot.com/2016/10/reasons-for-owning-gravis-ultrasound.html

I actually think these glitches are not as severe in DOSBox (Staging and others), both with GUS and SB,  possibly because DMA is more instantaneous than real hardware (just a theory though).

johnnovak:
> I actually think these glitches are not as severe in DOSBox (Staging and others), both with GUS and SB, possibly because DMA is more instantaneous than real hardware (just a theory though).
> 
> That said, I do think the issues in Nerdy Pleasures output is different in nature than the very short and repeating pattern in @marsej's recordings; So I think we're dealing with different issues.

Yeah I guess my TL;DR comment and general message to  @marsej is that unless someone can prove it to me that the hardware play game X without clicks and the clicks are artifact of imperfect emulation, I'm quite reluctant to spend significant time on these things. Unless the clicks are extremely annoying, like in the case of OPL which you improved in the present PR @kcgen.

The clicks in the @marsej 's recording are audible, sure, but I started listening to the "bad" recordings first and honestly I was like "what is even wrong with this?". That's in my studio headphones after 20-30 years of doing audio production as a hobby. But yeah, if you A/B it with the "good" version, you can spot the difference.

So, all in all, an extremely low priority issue, and most likely not even an issue because that's just how the GUS and these playroutines work.

kcgen:
Took a recording here (sc2 w/ .com file, non-CD version).

Skip to 0:26s and you'll see the oscilloscope mid-line bump upward with a small pop (so that particular sound track/effect within the MOD has some DC bias):

https://github.com/dosbox-staging/dosbox-staging/assets/1557255/9712e7f7-254e-43bc-8ae7-ed93a11399b8

This also happens in the open source UQM recording at ~0.28s, and that's a native implementation directly writing PCM to the host sound card (non-DOS):

https://www.youtube.com/watch?app=desktop&v=MrvxgcFCyKw

So my hunch is the talking pet MOD file has problems. Maybe the UQM community has fixed this MOD?

johnnovak:
> So my hunch is the talking pet MOD file needs all it's tracks normalized/centered; maybe the UQM community has fixed it?(perhaps just for UQM, but ideally even hacked it back into the original DOS data files?)

Yeah, that's a rookie mistake. I use to DC-correct my samples and fade them in & out at the start and end. If indeed that's the root source for this, that is.

PR #3318, opened 2024-01-10
Improve English language strings

ftortoriello:
# Description

Some (mostly) minor improvements to the internal English messages.

Summary of changes:
* Fix the "Adding utilities" Wiki URL
* Fix default value of the "timed" joystick setting in the help description
* Fix some typos (such as "to fit width width")
* Split the "cycleup" and "cycledown" setting descriptions; it was strange when viewing the help with "CONFIG"
* Split lines when they have more than 80 characters (in overlay directory messages)
* Improve consistency in help and config messages:
  * Always write ISO, BIN, CUE and KB in uppercase
  * Spacing: Normalize margins, newlines and space between sentences
  * Fix punctuation (full stops)
  * Always use "-" for lists in the "Notes" sections of help messages when there are multiple paragraphs, except when that would cause more line breaks.
  * Use single quotes for configuration values (in the IMGMOUNT unsupported format message)
  * Add missing quotes in command line options
  * Replace instances of "CDROM" by "CD-ROM", "start-up" by "startup", and "file name" by "filename"
* Remove the question mark from the CHOICE example, as another one is added after the list of choices ("[Y,N]?")

# Checklist

I have:


johnnovak:
Great attention to detail @ftortoriello, excellent work! 🎉 Merging once CI passes.

src/dos/program_tree.cpp
```diff
@@ -424,12 +424,12 @@ void TREE::AddMessages()
...
-	        "  /p        display one page a time, shorten output to fit width width.\n"
-	        "  /da       display attributes.\n"
-	        "  /df       display size for files.\n"
-	        "  /dh       also display hidden and system files/directories.\n"
-	        "  /o[color=white]ORDER[reset]   orders the list (see above)\n"
+	        "  /f        also displays files.\n"
+	        "  /p        displays one page a time, shorten output to fit screen width.\n"
+	        "  /da       displays attributes.\n"
+	        "  /df       displays size for files.\n"
+	        "  /dh       also displays hidden and system files/directories.\n"
```
johnnovak, line 431: "Display" was fine for all of these; it's the "command form" or "imperative mood" (I'm not a grammar expert 😄). Can you change these back please, and rather fix up the ones you haven't touched? (e.g. "use only 7-bit" instead of "uses", "omit header" instead of "omits" etc.)

Generally, we want "imperative mood" for command line option descripitons.

src/dosbox.cpp
```diff
@@ -692,12 +692,13 @@ void DOSBOX_Init()
 
 	pint = secprop->Add_int("cycleup", always, 10);
 	pint->SetMinMax(1, 1000000);
-	pint->Set_help("Number of cycles added or subtracted with speed control hotkeys\n"
-	               "(10 by default).");
+	pint->Set_help("Number of cycles added with the speed up hotkey (10 by default).\n"
```
johnnovak, line 695: Technically, it's called "increase cycles", so we should say "with the increase cycles" hotkey.

src/dosbox.cpp
```diff
@@ -692,12 +692,13 @@ void DOSBOX_Init()
...
+	               "Setting it lower than 100 will be a percentage.");
 
 	pint = secprop->Add_int("cycledown", always, 20);
 	pint->SetMinMax(1, 1000000);
-	pint->Set_help("Setting it lower than 100 will be a percentage (20 by default).");
+	pint->Set_help("Number of cycles subtracted with the slow down hotkey (20 by default).\n"
```
johnnovak, line 700: "decrease cycles hotkey"

src/dosbox.cpp
```diff
@@ -990,7 +991,7 @@ void DOSBOX_Init()
 	        "ReelMagic (aka REALmagic) MPEG playback support:\n"
 	        "  off:       Disable support (default).\n"
 	        "  cardonly:  Initialize the card without loading the FMPDRV.EXE driver.\n"
-	        "  on:        Initialize the card and load the FMPDRV.EXE on start-up.");
+	        "  on:        Initialize the card and load the FMPDRV.EXE on startup.");
```
johnnovak, line 995: 🏅

src/gui/render.cpp
```diff
@@ -1145,7 +1145,7 @@ static void init_render_settings(Section_prop& secprop)
 	        "                 contrast.\n"
 	        "You can also set custom colours by specifying 16 space or comma separated\n"
 	        "colour values, either as 3 or 6-digit hex codes (e.g. #f00 or #ff0000 for full\n"
-	        "red), or decimal RGB triplets (e.g. (255, 0, 255) for magenta). The 16 colours\n"
+	        "red), or decimal RGB triplets (e.g. '(255, 0, 255)' for magenta). The 16 colours\n"
```
johnnovak: Hmm, maybe remove the quotes here? I did not use quotes for the hex values.

src/shell/shell.cpp
```diff
@@ -889,8 +891,8 @@ void SHELL_Init() {
 	        "  [color=light-green]echo[reset] [color=light-cyan][MESSAGE][reset]\n"
 	        "\n"
 	        "Where:\n"
-	        "  [color=light-cyan]on|off[reset]  Turns on/off command echoing.\n"
-	        "  [color=light-cyan]MESSAGE[reset] The message to display.\n"
+	        "  [color=light-cyan]on|off[reset]  turns on/off command echoing.\n"
```
johnnovak, line 894: "turn off"?

src/shell/shell.cpp
```diff
@@ -1219,19 +1221,19 @@ void SHELL_Init() {
 	        "\n"
 	        "Where:\n"
 	        "  [color=light-cyan]TEXT[reset]         is the text to display as a prompt, or empty.\n"
-	          "  /c[:][color=white]CHOICES[reset] Specifies allowable keys, which default to [color=white]yn[reset].\n"
-	          "  /n           Do not display the choices at end of prompt.\n"
-	          "  /s           Enables case-sensitive choices to be selected.\n"
-	          "  /t[:][color=white]c[reset],[color=light-magenta]nn[reset]    Default choice to [color=white]c[reset] after [color=light-magenta]nn[reset] seconds.\n"
+	          "  /c[:][color=white]CHOICES[reset] specifies allowable keys, which default to [color=white]yn[reset].\n"
+	          "  /n           does not display the choices at end of prompt.\n"
+	          "  /s           enables case-sensitive choices to be selected.\n"
```
johnnovak, line 1226: I have mixed feelings about these. How I see it, we either do one of the following:

- Complete sentences that start with uppercase letter, and then we put a period at the end
- Incomplete sentences that start with a lowercase letter, and there's NO period at the end

Also, please restore the use of the imperative mood. We should use imperative mood all the time (I know currently it's inconsistent, but we should strive towards that goal). So "do not display", "enable case-sensitive", etc.

Study the output of some GNU Linux commands, such as `ls --help`, `cp --help` for guidance.

src/shell/shell.cpp
```diff
@@ -1241,15 +1243,15 @@ void SHELL_Init() {
 	MSG_Add("SHELL_CMD_PATH_HELP_LONG",
 	        "Usage:\n"
 	        "  [color=light-green]path[reset]\n"
-	        "  [color=light-green]path[reset] [color=light-cyan][[drive:]path[;...][reset]\n"
+	        "  [color=light-green]path[reset] [color=light-cyan][[DRIVE:]PATH[;...][reset]\n"
 	        "\n"
 	        "Where:\n"
-	        "  [color=light-cyan][[drive:]path[;...][reset] is a path containing a drive and directory.\n"
+	        "  [color=light-cyan][[DRIVE:]PATH[;...][reset] is a path containing a drive and directory.\n"
```
johnnovak, line 1249: In all honestly, I'd get rid of these "is" words everywhere. My ideal version looks like this:

```
  [DRIVE:]PATH[;...]  path containing a drive and directory
```

So two spaces at least between the option name and the description for clear separation, and all lowercase, imperative mood, and no period at the end.

If there's a long block of text of multiple fully-formed sentences, we can use normal sentences (starts with uppercase, period at the end, etc.)

src/shell/shell.cpp
```diff
@@ -1301,17 +1304,16 @@ void SHELL_Init() {
 	        "  [color=light-green]move[reset] [color=white]DIRECTORY1[reset] [color=light-cyan]DIRECTORY2[reset]\n"
 	        "\n"
 	        "Where:\n"
-	        "  [color=white]FILENAME[reset]    Can be either an exact filename or an inexact filename with\n"
+	        "  [color=white]FILENAME[reset]    can be either an exact filename or an inexact filename with\n"
 	        "              wildcards, which are the asterisk (*) and the question mark (?).\n"
 	        "              Multiple, comma-seperated, filenames can be provided.\n"
-	        "  [color=white]DIRECTORY[reset]   An exact directory name, not containing any wildcards.\n"
-	        "  [color=light-cyan]DESTINATION[reset] An exact filename or directory, not containing any wildcards.\n"
+	        "  [color=white]DIRECTORY[reset]   is an exact directory name, not containing any wildcards.\n"
```
johnnovak, line 1310: "an exact directory name, not containing any wildcards"

src/shell/shell.cpp
```diff
@@ -1301,17 +1304,16 @@ void SHELL_Init() {
...
+	        "  [color=light-cyan]DESTINATION[reset] is an exact filename or directory, not containing any wildcards.\n"
```
johnnovak, line 1311: "an exact filename or directory, not containing any wildcards"

src/shell/shell.cpp
```diff
@@ -1325,11 +1327,11 @@ void SHELL_Init() {
...
-		"  [color=white]%VAR[reset]    A single character representing a variable, prefixed by a '%'.\n"
-		"  [color=light-cyan]in[reset]      A case-insensitive keyword.\n"
-		"  [color=white](SET)[reset]   The set of strings to replace [color=white]%VAR[reset] instances in [color=white]COMMAND[reset].\n"
-		"  [color=light-cyan]do[reset]      A case-insensitive keyword.\n"
-		"  [color=white]COMMAND[reset] The command to be repeated for each string in [color=white](SET)[reset].\n"
+		"  [color=white]%VAR[reset]    is a single character representing a variable, prefixed by a '%'.\n"
+		"  [color=light-cyan]in[reset]      is a case-insensitive keyword.\n"
+		"  [color=white](SET)[reset]   is the set of strings to replace [color=white]%VAR[reset] instances in [color=white]COMMAND[reset].\n"
+		"  [color=light-cyan]do[reset]      is a case-insensitive keyword.\n"
+		"  [color=white]COMMAND[reset] is the command to be repeated for each string in [color=white](SET)[reset].\n"
```
johnnovak, line 1334: Remove all the redundant "is" words, remove periods, as per above. Just use sentence fragments, which is perfectly fine in such cases (also for bulleted lists, that's perfectly legal in English).

johnnovak:
Okay @ftortoriello, I've reviewed the first commit in detail. You've done a lot of nice improvements and I appreciate your general attention to detail! However, I think I have a bit of a different idea on how we should standardise the help descriptions going forward.

Please see my inline comments. I haven't commented on every single occurrence that needs adjustment, just a few. You can then exatrapolate from that and adjust things accordingly. I'm not saying to refactor _all existing messages_ in the codebase according to my suggested style, just what you've touched. Thanks 😄

src/dos/program_tree.cpp
ftortoriello: I changed it to "displays" because it is part of the "Where:" section, so I thought it should be read "Where /p displays...". But I can change it everywhere if you want.

Regarding the first line of help messages, it is like that for every command (checking "help /a" I've just noticed that only the "more" description is different...). And it's also like that on MS-DOS and FreeDOS. Should I change it for every command anyways?

src/dosbox.cpp
ftortoriello: Agreed

src/shell/shell.cpp
ftortoriello: I'd remove the "is" words too...
Perhaps "Where:" could be changed to "Parameters:" or "Arguments:"?
I think as it is now it reads fine ("Where ARGUMENT is..." or "Where /p does..."), but with the changes you propose renaming "Where" would be much better.

ftortoriello:
Thanks for the review!
I've made those changes in the style that was predominant for consistency, but I agree that the GNU style may be better / simpler, I'll switch to that.

src/gui/render.cpp
ftortoriello: True. I added them because of the double parenthesis, but they are not necessary

johnnovak: People won't be confuse by the double parenthesis, but could be confused by the quotes. So I'd go with the least confusing option :)

src/shell/shell.cpp
johnnovak: Yeah, I agree "arguments" would be better, yes. I think we should break away from this concept that this should read as a whole sentence starting from "Where" -- there's not much benefit of that.

src/dos/program_tree.cpp
johnnovak: Yeah, it should be like the `MORE` command, so imperative everywhere (e.g., "Copy one or more files", not "Copies"). If you have the inclination to change this for all commands, please go for it 😄

johnnovak:
> Thanks for the review! I've made those changes in the style that was predominant for consistency, but I agree that the GNU style may be better / simpler, I'll switch to that.

Thanks for taking care of this @ftortoriello. It's a big job, and it needs someone like you with an eye for detail and consistency 🏅

src/shell/shell.cpp
ftortoriello: I chose the second option, as most descriptions are short.
For the few which were two sentences I separated them by semicolons, leaving only one sentence, to avoid changing styles

ftortoriello:
I ended up changing the help messages for all the commands :smile:
I went with "Parameters" instead of "Arguments" because it is more widely used in DOSBox, but I can change it

src/dos/program_loadfix.cpp
```diff
@@ -87,14 +87,14 @@ void LOADFIX::Run(void)
...
+	        "Load a program in the specific memory region and then run it.\n"
 	        "\n"
 	        "Usage:\n"
 	        "  [color=light-green]loadfix[reset] [color=light-cyan]GAME[reset] [color=white][PARAMETERS][reset]\n"
 	        "  [color=light-green]loadfix[reset] [/d] (or [/f])[reset]\n"
 	        "\n"
-	        "Where:\n"
-	        "  [color=light-cyan]GAME[reset] is a game or program to be loaded, optionally with parameters.\n"
+	        "Parameters:\n"
+	        "  [color=light-cyan]GAME[reset]  game or program to be loaded, optionally with parameters\n"
```
johnnovak, line 97: to load

src/dos/program_mixer.cpp
```diff
@@ -795,22 +795,22 @@ void MIXER::Run()
...
-	        "                 volumes of stereo channels separately (e.g. [color=white]10:20[reset], [color=white]150:d6[reset])\n"
-	        "    Stereo mode: [color=white]stereo[reset], or [color=white]reverse[reset] (stereo channels only)\n"
-	        "    Crossfeed:   [color=white]x0[reset] to [color=white]x100[reset], sets crossfeed strength (stereo channels only)\n"
-	        "    Reverb:      [color=white]r0[reset] to [color=white]r100[reset], sets reverb level\n"
-	        "    Chorus:      [color=white]c0[reset] to [color=white]c100[reset], sets chorus level\n"
+	        "                 volumes of stereo channels separately (e.g. [color=white]10:20[reset], [color=white]150:d6[reset]).\n"
+	        "    Stereo mode: [color=white]stereo[reset], or [color=white]reverse[reset] (stereo channels only).\n"
+	        "    Crossfeed:   [color=white]x0[reset] to [color=white]x100[reset], sets crossfeed strength (stereo channels only).\n"
+	        "    Reverb:      [color=white]r0[reset] to [color=white]r100[reset], sets reverb level.\n"
+	        "    Chorus:      [color=white]c0[reset] to [color=white]c100[reset], sets chorus level.\n"
```
johnnovak, line 813: "set" for all three

johnnovak:
Looks very good now @ftortoriello, just a few minor comments and we're done 😄 The other thing, the commit history needs a bit of a cleanup, as per below.

I think you'll need update the English language file again at the end. Better yet, you could rebase your history and only do a single final language file update as the very last commit.

These two comments could also be squashed into earlier relevant commits, but at the very least the wording of the commit message should be changed (see [here](https://cbea.ms/git-commit/), there's a link to this in the PR checklist too):

- Replace "Where:" occurrence I missed 
- Minor message changes I missed before

src/dos/program_loadfix.cpp
ftortoriello: Agreed. I changed it in other messages too.

ftortoriello:
Ok, I squashed the two commits you mentioned and the .lng file updates, leaving only the latest one

johnnovak:
> Ok, I squashed the two commits you mentioned and the .lng file updates, leaving only the latest one

Thanks, so only addressing those two small comments I made earlier and rebasing to main is outstanding, then merge time 😄

johnnovak (approved):
Approving in the meantime, great work! 🎉

ftortoriello:
> Thanks, so only addressing those two small comments I made earlier and rebasing to main is outstanding

Rebased!
You mean the comments about "to be loaded" and "sets"?
I changed that in the last two commits (before the lng file update)

johnnovak:
> > Thanks, so only addressing those two small comments I made earlier and rebasing to main is outstanding
> 
> Rebased! You mean the comments about "to be loaded" and "sets"? I changed that in the last two commits (before the lng file update)

Ah, missed that. CI is all green, so merging! 🚀 Thanks again, this was a _very nice_ cleanup job! ❤️

Issue #3326, opened 2024-01-14
Emulated Joystick default when Joystick=auto


Python-Exoproject:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Using Hardball! (1987) as an example, in ECE if you have no real joystick plugged in and joystick=auto the game will not detect joystick for player control and the mapper will show "disabled" above the joystick controls. If you then set joystick=2axis the game will detect the emulated joystick and joystick 1 will be active in the mapper, as it should.

In Staging 81-RC1 joystick=2axis works as it should, however if you set joystick=auto the game can still see the emulated joystick and offers it for player control.

I think this is a bug as you would only want to opt in to the emulated joystick by setting joystick=2axis\4axis\etc, not have it on all the time unless joystick=disabled

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Set joystick=auto in conf
2. Start Hardball!
3. Look at the controls offered for the two players

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

weirddan455:
Confirmed.  This behavior is also a regression from 0.80.1.  In that previous version, Hardball does not detect a joystick when `joystick = auto` is set and there is no joystick plugged in.

I'll dig into this more in a little bit.  I believe the reason we do this is to allow the user to create a "virtual" joystick by mapping joystick keys to other devices such as a mouse and keyboard (by using the mapper accessed by `dosbox -startmapper`).  Perhaps the best solution is to only present a joystick to DOS if the user has such keys mapped or a real joystick is plugged in.

weirddan455:
Fixed in #3327

Issue #3320, opened 2024-01-13
Joystick detection issue in some LucasArts flight sims


almeath:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.0-alpha

### What Operating System are you using?

Windows 11

### Whats your question and how can we help?

I have been configuring some older LucasArts flight sims to work in DOSBox Staging, which are currently already running fine for me under DOSBox-X. 

I have carefully checked joystick settings and all other config options, but two games in particular seemingly lock up on a message that says "center joystick then click button" which appears after the title screen. The affected games are Battlehawks 1942 and Secret Weapons of the Luftwaffe. 

I have tried pressing all keyboard keys and mouse buttons and nothing will let me advance beyond this screen. Disabling the joystick in the config file also does nothing. I would be happy to play these games with the keyboard, but running them in Staging seems to demand the presence of a joystick.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

almeath:
I have figured out that the 'hidden' joystick setting is now required rather than 'auto' in order to bypass this issue.

johnnovak:
> I have figured out that the 'hidden' joystick setting is now required rather than 'auto' in order to bypass this issue.

We should probably add this info to the wiki.

rderooy:
If the game has joystick support, and it works in other DOSBox variants, then it would appear something is broken. Hiding or disabling the joystick is only a work-around, not a fix.

weirddan455:
I am unable to reproduce this.  I tried both games with every available joystick setting.  Tried with my gamepad plugged in and with it unplugged.  I never see the screen you're talking about.

weirddan455:
Probably this is the same issue as #3326 which I was able to reproduce.

weirddan455:
Should be fixed in #3327 but we can leave this open for now as I was unable to reproduce.

If you could please test against latest main and let us know if the issue is now resolved, that would be great.

johnnovak:
@almeath This should be fixed in the latest DEV build. Could you please confirm?

almeath:
Thanks. I can confirm that in 0.81.0-rc2, this issue is resolved. I can specify joystick=auto and the above mentioned games start up normally.

johnnovak:
[...]

PR #3327, opened 2024-01-14
Disable joystick emulation when "joystick = auto" is set and no joystick is found

weirddan455:
# Description

We support a "virtual" joystick by allowing the user to use `dosbox -startmapper` to map joystick keys to another device if they do not have a physical joystick.  On v0.80.1 ,this feature was broken.  On main, this "virtual" joystick is presented to DOS regardless of whether the user has any joystick keys actually mapped when `joystick = auto` is set (default setting).

This PR changes behavior to not emulate a joystick unless there is a physical joystick attached or the user explicitly enabled a joystick in the config (ex. by setting `joystick = 2axis`).  This matches the behavior of Dosbox 0.74.3.

Ideally, we may want "auto" to be a bit smarter and check if the user has joystick keys mapped but `sdl_mapper.cpp` makes me a bit nauseated every time I look at it so I think this quick fix is fine for v0.81.

My thinking is that mapping joystick keys is more of an advanced feature and those users can edit the config file.  This prevents certain games from being unplayable when a joystick is not present (see related issues).

## Related issues

#3320 (unable to reproduce)
#3326

# Manual testing

Tested with Hardball game on issue #3226

When `joystick = auto` is set, no joystick is detected unless there is one physically connected.

# Checklist

I have:

### DIFF
--- a/contrib/resources/translations/en.lng
+++ b/contrib/resources/translations/en.lng
@@ -1487,6 +1487,7 @@ Override the frame rate code used during video playback:
 :CONFIG_JOYSTICKTYPE
 Type of joystick to emulate:
   auto:      Detect and use any joystick(s), if possible (default).
+             Joystick emulation is disabled if no joystick is found.
   2axis:     Support up to two joysticks, each with 2 axis.
   4axis:     Support the first joystick only, as a 4-axis type.
   4axis_2:   Support the second joystick only, as a 4-axis type.
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1022,6 +1022,7 @@ void DOSBOX_Init()
 	pstring->Set_help(
 	        "Type of joystick to emulate:\n"
 	        "  auto:      Detect and use any joystick(s), if possible (default).\n"
+	        "             Joystick emulation is disabled if no joystick is found.\n"
 	        "  2axis:     Support up to two joysticks, each with 2 axis.\n"
 	        "  4axis:     Support the first joystick only, as a 4-axis type.\n"
 	        "  4axis_2:   Support the second joystick only, as a 4-axis type.\n"
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -2877,11 +2877,16 @@ static void QueryJoysticks()
 	if (SDL_WasInit(SDL_INIT_JOYSTICK) != SDL_INIT_JOYSTICK)
 		SDL_InitSubSystem(SDL_INIT_JOYSTICK);
 
+	const bool wants_auto_config = joytype & (JOY_AUTO | JOY_ONLY_FOR_MAPPING);
+
 	// Record how many joysticks are present and set our desired minimum axis
 	const auto num_joysticks = SDL_NumJoysticks();
 	if (num_joysticks < 0) {
 		LOG_WARNING("MAPPER: SDL_NumJoysticks() failed: %s", SDL_GetError());
 		LOG_WARNING("MAPPER: Skipping further joystick checks");
+		if (wants_auto_config) {
+			joytype = JOY_NONE_FOUND;
+		}
 		return;
 	}
 
@@ -2890,10 +2895,12 @@ static void QueryJoysticks()
 	mapper.sticks.num = static_cast<unsigned int>(num_joysticks);
 	if (num_joysticks == 0) {
 		LOG_MSG("MAPPER: No joysticks found");
+		if (wants_auto_config) {
+			joytype = JOY_NONE_FOUND;
+		}
 		return;
 	}
 
-	const bool wants_auto_config = joytype & (JOY_AUTO | JOY_ONLY_FOR_MAPPING);
 	if (!wants_auto_config)
 		return;
 


johnnovak:
I'm not full across the joystick emulation code @weirddan455, but I have an uncomfortable feeling that you might be undoing previous improvements in this area that were introduced for good reasons by @kcgen. Specifically, I'm referring to one of @kcgen 's comments in one the linked PRs:

> joysticktype = none: should detect and allow mapping of joystick events to keyboard events, however the joystick will be invisible to DOS programs. This is useful to prevent some games from auto-using the joystick or beginning a calibration phase on startup when you'd rather use the keyboard/mouse.

> joysticktype = auto: should behave equal to the existing latest sources.

Could you get yourself up-to-speed on the joystick emulation and all the recent issues and fixes that happened in 2023 and re-think/re-validate your fix in light of that? I just want to avoid the case when someone who doesn't have the full 100% picture undoes earlier changes, that's all 😄 Or @kcgen can chime in perhaps, if he has time/energy for it.

If you're already a joystick expert, then I take it all back—I'm not, it just seems to be that this fix conflicts with earlier improvements.

Relevant recent PRs (there might be more):

https://github.com/dosbox-staging/dosbox-staging/pull/2713

https://github.com/dosbox-staging/dosbox-staging/pull/2516

https://github.com/dosbox-staging/dosbox-staging/pull/1295

https://github.com/dosbox-staging/dosbox-staging/pull/1149

kcgen (approved):
Unable to test here, but looks perfect and the logic makes sense.

I'm not sure if the current conf wording is still OK (maybe this fixed behavior now matches the wording?)

johnnovak:
> Unable to test here, but looks perfect and the logic makes sense.
> 
> I'm not sure if the current conf wording is still OK (maybe this fixed behavior now matches the wording?)

How about the points I brought up? Doesn't this conflict with your earlier work? Like I said, I'm far from being an expert in the area, but it appeared to be conflicting 😄

kcgen:
Sorry @johnnovak , will try to answer:

>> How about the points I brought up? Doesn't this conflict with your earlier work? Like I said, I'm far from being an expert in the area, but it appeared to be conflicting 😄

> `joysticktype = none`: should detect and allow mapping of joystick events to keyboard events, however the joystick will be invisible to DOS programs. This is useful to prevent some games from auto-using the joystick or beginning a calibration phase on startup when you'd rather use the keyboard/mouse.

We don't have `joysticktype = none` anymore, but it's equivalent to `hidden` now, because it doesn't emulate the DOS joystick, however the SDL joystick subsystem is init'd and so allows the user to map the host joystick events over to DOS keyboard events.

So I think that's all still fine here, too.

As for this:

> `joysticktype = auto`: should behave equal to the existing latest sources.

(It's likely I wasn't aware of this regression :smile:)

@weirddan455's fix is exactly what I was thinking when I replied to Python:



So I think your fix, @weirddan455 is perfect, and I think seeing "joysticktype = 2axis" is perfectly reasonable for users mapping host keyboard events over to DOS joystick events.

Thanks for rounding these up, @johnnovak !

weirddan455:
>  I don't think it's great that joystick emulation is only enabled/disabled when a physical controller is plugged in or the joy emulation keys are mapped. These should be separate concerns, and I think they were until this fix.

To be clear, this is only affecting the "auto" setting.   Prior to this PR, auto would always emulate a joystick.  Now it only emulates a joystick when one is plugged in.  If a user wants to emulate a joystick by mapping keyboard keys, they would need to change a config setting to explicitly say which type of joystick they would like.  This is how mainline Dosbox (and apparently most/all of the other forks) work.

>  I should be able to map the joy keys during running the game, so turning the joystick emulation off just because the joy hotkeys are unmapped at startup is an unfortunate conflation of things.

I don't think it's possible right now to change mappings in the middle of a game.  If you mean in the future with your OSD, hopefully this entire mapper will be re-written by then :smile: 

>  joysticktype = none: should detect and allow mapping of joystick events to keyboard events, however the joystick will be invisible to DOS programs. This is useful to prevent some games from auto-using the joystick or beginning a calibration phase on startup when you'd rather use the keyboard/mouse.

This is kind of the inverse of the issue I am solving.  This is called `joystick = hidden` now and it's for when you want to map physical joystick buttons to virtual keyboard buttons (but hide the joystick from DOS).  This mode should not be affected by this change.

>  Could you get yourself up-to-speed on the joystick emulation and all the recent issues and fixes that happened in 2023 and re-think/re-validate your fix in light of that? 

I did go through most of the recent history before I submitted this PR (skimmed through the rest just now).  As far as I can tell, I still believe this is the correct fix.

johnnovak:
Thanks @kcgen & @weirddan455, that explains a lot but I'm still a bit confused. `auto` makes sense now. But I'm confused about `hidden` and `disabled`.

> We don't have `joysticktype = none` anymore, but it's equivalent to `hidden` now, because it (`hidden`) hides the joystick from the DOS side (even if you have one plugged in). The benefit is you can map the host joystick events over to DOS keyboard events.

If `hidden` make the joysticks essentially invisible to DOS, then why do we have the `disabled` setting? How can you "disable" the joystick even more than that? 😅 Just having `auto`, `hidden` renamed to `disabled`, and the concrete joy type settings would make sense to me. What does the current `disabled` mode achieve that `hidden` doesn't? Why is it necessary?

> To be clear, this is only affecting the "auto" setting. Prior to this PR, auto would always emulate a joystick. Now it only emulates a joystick when one is plugged in. If a user wants to emulate a joystick by mapping keyboard keys, they would need to change a config setting to explicitly say which type of joystick they would like. This is how mainline Dosbox (and apparently most/all of the other forks) work.

Would you be able to extend the `joysticktype` descriptions, especially for the `auto` setting, with these additional details? In a much shorter form, of course.

kcgen:
> I don't think it's possible right now to change mappings in the middle of a game.

Yes - mappings are 'live' and immediately available in-game once the mapper closes.

weirddan455:
>  If hidden make the joysticks essentially invisible to DOS, then why do we have the disabled setting? How can you "disable" the joystick even more than that? 😅 Just having auto, hidden renamed to disabled, and the concrete joy type settings would make sense to me. What does the current disabled mode achieve that hidden doesn't? Why is it necessary?

`disabled` means don't even look at the joystick on the host side.  It won't init SDL's joystick subsystem so we won't receive or poll for joystick events.

kcgen:
> disabled means don't even look at the joystick on the host side. It won't init SDL's joystick subsystem so we won't receive or poll for joystick events.
"strict mode", forcing users into keyboard controls no matter what, just like DOS games were meant to be played :laughing: 

More critically some users have really laggy USB controllers (30+ ms-per-poll.. yes, it's brutal) - and when they set `disabled`, all of a sudden their sluggish games pop to life with low latency keyboard controls (it's happened a couple times now).

johnnovak:
> `disabled` means don't even look at the joystick on the host side. It won't init SDL's joystick subsystem so we won't receive or poll for joystick events.

I fully get that from a purely technical perspecive. Maybe I wasn't clear, I'm questioning the usefulness of that _technical_ feature.

Wouldn't just renaming `hidden` to `disabled` and getting rid of `disabled` suffice and simplify things? What's the harm in always initialising the SDL controller stuff?

johnnovak:
> "strict mode", forcing users into keyboard controls no matter what, just like DOS games were meant to be played 😆

Ah, we replied at the same time 😄 

> More critically, some users have really laggy USB controllers (30+ ms-per-poll.. yes, it's brutal) - and when they set `disabled`, all of a sudden their sluggish games pop to life with low latency keyboard controls (it's happened a couple times now).

Ok, that's a bit sad, but looks like it's necessary then 🤷🏻

weirddan455:
I added a sentence to the config help text.  I'm unable to merge with the CI being broken (PVS is throwing an HTTP 404 error and I assume the Mac builds are offline).  Someone with higher permissions than me will need to merge when ready.

johnnovak:
> I added a sentence to the config help text. I'm unable to merge with the CI being broken (PVS is throwing an HTTP 404 error and I assume the Mac builds are offline). Someone with higher permissions than me will need to merge when ready.

Cheers. The PVS Studio issue has been rectified, so please rebase. macOS builds will be broken for a while, we'll need to live with that a little longer.

weirddan455:
Rebased.  PVS is green now.

johnnovak:
> Rebased. PVS is green now.

Thanks man. Merging.

Issue #3323, opened 2024-01-13
Call for community provided-or-funded Apple silicon CI runners


kcgen:
For the last couple years I've provided and maintained a macOS CI runner for the project, however I'm not able to continue in this role.

This is my own fault, as any single point of failure on an open-source project is a big risk, and now it's been realized unfortunately rather abruptly.  I should have worked harder to eliminate this risk when I had ample time and ability to prioritize this.

A couple quick options that I can suggest:

- **Apple macOS m1 community funding**
    
    Currently, GitHub doesn't offer the macOS M1 runner for free, so one option is funding the CI minutes with community donations. There have been 7,686 macOS workflow runs since getting it setup during the first year of the project, averaging about 250 MacOS workflows per month. Each workflow consists of two M1 setup & build jobs (debug and release). Assuming each run needs 3 minutes to setup and build on the VM, that adds up to 1500 minutes/mo. At [0.16c/min](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates), that's 240 USD/mo.

   Of course, some months can be multiples higher during very active development. Perhaps the project could re-tool CI and dramatically reduce the use of the these M1 jobs to save on those costs.

- **Apple macOS m1 community runners**

    Another option is having a new team of macOS apple silicon users run GitHub's runner software. The idea is that with enough redundancy, no single person is responsible - and machines can freely come off and online. These users could be credited with an "Apple CI team" rank in Discord and given specific thanks in the release notes.

- **Local builds only**

  Another option is simply doing without, and the maintenance team builds releases locally. Apple M1 users who want alpha builds can setup a compilation environment.

The main point in all of this is **_avoiding making this one person's responsibility_** and ensuring the solution has redundancy and some scalability.  If the community wants it bad enough, then the community needs to step forward with their spare machine cycles and/or funds to make this happen.

There's certainly other creative options; Just throwing these out to help!

weirddan455:
I'd like to throw out third party cloud hosting as another potential option.  Googling "mac m1 cloud hosting" shows several options available at significantly less cost than what GitHub is charging.  As long as they allow SSH access, we should be able to install GitHub runner software on it.

We would still need funding for that of course, but we may be able to bring the cost down.

johnnovak:
Thanks for running a MacMini at your personal expenses for this long @kcgen ❤️ 

[...]

PR #3332, opened 2024-01-18
Use Meson wrap for SDL2 on macOS

kklobe:
# Description

Here is some further hackery to the macOS runner CI machinery for the 0.81 release.

I've switched to using the Meson wrap for SDL2 on macOS so we can control the target macOS version via `MACOSX_DEPLOYMENT_TARGET`. A core issue resulting from HomeBrew use is that packages are compiled to target the version of the OS running HomeBrew, which is by design; they make it very clear that it is not intended for building apps to distribute.

SDL needs to be built with the older target OS so as to avoid a missing symbol crash, and using the Meson wrap was the quickest way to reliably achieve this. I'm only using the wrap for macOS builds.

There are also some other hacks to better support the new temporary runner configuration. The big change I've made is to completely separate the macOS x86_64 and arm64 runners, so they run independently:

CleanShot 2024-01-18 at 08 35 12@2x

I will document this more fully on the wiki soon.

## Related issues

#3323

# Manual testing

@Burrito78, @johnnovak and myself have run the build artifacts on our respective macOS systems.

# Checklist

I have:

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -21,7 +21,7 @@ jobs:
             needs_deps: true
             build_flags: -Denable_debugger=normal
             brew_path: /usr/local/homebrew
-            max_warnings: 0
+            max_warnings: 80
 
 #          - name: GCC 12
 #            host: macos-13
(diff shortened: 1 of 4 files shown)


johnnovak (approved):
Excellent @kklobe 🏅 Looking forward to the doco 😉

PR #3337, opened 2024-01-19
De-officialise MSVC Windows builds

johnnovak:
# Description

We made an executive decision with @kcgen in December that due to various problems with it, we'll simply stop shipping Windows MSVC builds in our releases going forward. The MSVC builds have already been dropped from the 0.81.0 release candidates.

Ideally, we want to keep the MSVC workflow running in CI for hardening, plus Windows devs might prefer MSCV to MSYS2 for development. But as we're not officially shipping these builds, they should be removed from the public dev builds page on our website. Having them listed there would just confuse users, they'd download the wrong build by mistake, etc.

This will mean we no longer support Windows 7 in our new releases. My understanding is MSYS2 builds need Windows 10, but people could still use our MSVC builds on Windows 7.

Before people bring out the pitchforks, here's an announcement from Steam:

> As of January 1 2024, Steam will officially stop supporting the Windows 7, Windows 8 and Windows 8.1 operating systems. After that date, existing Steam Client installations on these operating systems will no longer receive updates of any kind including security updates. Steam Support will be unable to offer users technical support for issues related to the old operating systems, and Steam will be unable to guarantee continued functionality of Steam on the unsupported operating system versions.

---

@kcgen disabled the MSVC workflow because and shared a screenshot how to re-enable it. As mentioned, we'll need to fix the problem after re-enabling it, now it's broken:

image

# Checklist

I have:

### DIFF
--- a/website/docs/releases/development-builds.md
+++ b/website/docs/releases/development-builds.md
@@ -112,8 +112,7 @@ function set_ci_status(workflow_file, os_name, description) {
 document.addEventListener("DOMContentLoaded", () => {
     set_ci_status("linux.yml", "linux", "Linux x86-64");
     set_ci_status("macos.yml", "macos", "macOS ¹");
-    set_ci_status("windows-msys2.yml", "msys2", "Windows MSYS2 builds ²");
-    set_ci_status("windows-msvc.yml", "windows", "Windows MSVC builds ³");
+    set_ci_status("windows-msys2.yml", "windows", "Windows builds ²");
 });
 
 </script>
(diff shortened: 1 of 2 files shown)


website/docs/releases/development-builds.md
```diff
@@ -155,17 +154,6 @@ document.addEventListener("DOMContentLoaded", () => {
...
-      <img style="margin:auto;margin-left:0.1em;" src="../images/dots.svg">
-    </td>
-    <td id="msys2-build-version">
-      <img style="margin:auto;margin-left:0.1em;" src="../images/dots.svg">
-    </td>
-    <td id="msys2-build-date">
-      <img style="margin:auto;margin-left:0.1em;" src="../images/dots.svg">
-    </td>
-  </tr>
   <tr>
```
kklobe, line 157: I haven't had my coffee yet, but don't we want to keep the "msys2-build-*" items and remove the "windows-build-*" ones?

johnnovak: I've just renamed the `msys2-build` CSS ID to `windows-build` so it's consistent with the `macos-build` and `linux-build` IDs. But I guess `windows-msys2-build` would be better if we intend to keep MSVC around in CI.

FeralChild64 (approved):

PR #3341, opened 2024-01-22
Add script to repackage the Windows ZIP release package

johnnovak:
# Description

As per the title.

# Manual testing

I used this to repackage 0.82.0-RC2.

# Checklist

I have:


scripts/repackage-windows-zip.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [[ $IN_ZIP != $IN_ZIP_PREFIX* ]]; then
+	echo "Input filename must start with '$IN_ZIP_PREFIX'"
+	exit 1
+fi
+
+function strip {
+	local STRING=${1#$"$2"}
+	echo "${STRING%$"$2"}"
+}
```
weirddan455, line 29: I'm probably not the best person to review bash scripts because I don't know what the heck this is doing.  I have an idea based on the context but that syntax is a bit of a yikes.

scripts/repackage-windows-zip.sh
```diff
@@ -0,0 +1,45 @@
...
+
+VERSION=$(strip "$IN_ZIP" $IN_ZIP_PREFIX)
+ROOT_DIR=dosbox-staging-$VERSION
+OUT_ZIP=dosbox-staging-windows-$VERSION
+
+RMDIR_CMD="rm -rf $ROOT_DIR"
```
weirddan455, line 35: I just read about a bug the other day where a script erased an entire drive due to a line like this where the variable (under certain circumstances) could be an empty string.  Probably here that can't happen but `rm -rf $VARIABLE` always gives me a bit of a pause.

weirddan455 (approved):
I'll give you a check mark but just know I'm 90% winging it whenever I have to write a bash script :stuck_out_tongue_winking_eye: 

Up to you if you want to wait to have someone else give this a once over.

FeralChild64 (approved):
Looks sane. I was hoping InfoZIP has some options to do it without using bash script, but it seems there is none.

scripts/repackage-windows-zip.sh
johnnovak: Yeah, it could be evil, I agree. It also gave me a pause when I wrote it, but the alternative of potentially having leftover files in the target dir from a previous run is not ideal either. Like you said, it's safe as `$ROOT_DIR` always starts with `dosbox-staging-` (as safe as deleting stuff with `rm` can get... ;) )

johnnovak:
> I'll give you a check mark but just know I'm 90% winging it whenever I have to write a bash script 😜

Haha, I'm the same. I have this love/hate relationship with bash that yeah, the syntax is terrible and I have to look up things every single time I write a bash script as I'm not doing it daily. But whenever I rewrite a bash script in Python or something I regret it. Nothing makes it as easy to "eval" commands and pipe things as Bash; it's kind of an optimal solution for a large number of problems, even with the warts.

Grounded0:
[ShellCheck – shell script analysis tool](https://www.shellcheck.net/)

johnnovak:
> [ShellCheck – shell script analysis tool](https://www.shellcheck.net/)

It's part of the CI process. It complained about a few things that I needed to fix, so it's doing its job 😎

PR #3349, opened 2024-01-24
On Windows, use CreateProcess instead of execvp for restart function

### LINKED ISSUE #3346: Restart facility is unstable when launched from Windows GUI

0.80.1, 0.81-RC1, 0.81-RC2

Windows 11

_No response_

### Relevant hardware info

AMD Ryzen 9 3900X, NVidia GeForce RTX 3070, 32GB RAM

Using the provided Restart feature (default [Ctrl]+[Alt]+[Home]) results in undesirable behaviour should DOSBox Staging be launched directly from the Windows GUI. It does not seem to make any difference whether it is launched directly via the executable or through a shortcut, or whether this is done using the Explorer or the "Run" command.

On the first attempt it presents the following error in the client:

> Illegal command: dosbox.exe
> Preventing an early 'exit' call from terminating.

An accompanying warning in the console indicates:

> SHELL: Exit blocked because program quit after only 0.5 seconds

Any subsequent attempt to Restart will result in client exiting entirely, with the following error in the console (before it also closes):

> ABORT: Restarting failed

Also worth mentioning is that the inclusion of @ symbols or ECHO OFF in the [autoexec] section of the config file can result in the first Restart no longer showing system text, with ECHO OFF having a 100% chance of this occurring. A caret is present, all user-entered text works correctly and the environment itself appears otherwise unaffected, including text-based software (such as the exit screen on many Shareware titles), but no text-based feedback is presented from the DOS environment. This also appears to suppress the Shell warning.

Note that launching from a console (such as Windows Terminal) does _not_ exhibit this (mis)behaviour - including the missing text feedback - and will happily Restart as many times as you care to try. Execution via a batch file also works as expected, even when invoked via the Windows GUI. Both of these must take the current directory into consideration when relative paths are used in the DOSBox config, however.

Explain how to reproduce
1. Launch DOSBox Staging from the Windows GUI.
2. Trigger the Restart feature ([Ctrl]+[Alt]+[Home]).
3. <optional> Trigger the Restart feature a second time.

_No response_

### Your configuration

```ini
Configuration used is the default, as generated by the respective versions of DOSBox Staging.
```

### Provide a Log

```text
https://gist.github.com/Asterchades/7c9f0c2e7ae57c40c8c4ab494d6aa4fd
```

- [X] Yes, I agree.

weirddan455:
# Description
The implementation of execvp on Windows calls CreateProcess. Calling it directly allows us to create a detached process. This fixes an issue where sometimes Dosbox will exit on restart.

## Related issues
Fixes #3346 - I wrote a more detailed explanation of the problem there.

# Manual testing
Tested Ctrl + Alt + Home key combo to restart Dosbox.

- Tested when started by double-clicking the .exe from the GUI
- Tested when started from MSYS2 console
- Tested when started from a normal Windows console

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4525,6 +4525,17 @@ extern void DEBUG_ShutDown(Section * /*sec*/);
 void MIXER_CloseAudioDevice(void);
 
 void restart_program(std::vector<std::string> & parameters) {
+#ifdef WIN32
+	std::string command_line = {};
+	bool first = true;
+	for (const auto &arg : parameters) {
+		if (!first) {
+			command_line.push_back(' ');
+		}
+		command_line.append(arg);
+		first = false;
+	}
+#else
 	char** newargs = new char* [parameters.size() + 1];
 	// parameter 0 is the executable path
 	// contents of the vector follow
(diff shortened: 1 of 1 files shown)


src/gui/sdlmain.cpp
```diff
@@ -4525,6 +4525,17 @@ extern void DEBUG_ShutDown(Section * /*sec*/);
 void MIXER_CloseAudioDevice(void);
 
 void restart_program(std::vector<std::string> & parameters) {
+#ifdef WIN32
+	std::string command_line = {};
+	bool first = true;
+	for (std::string &arg : parameters) {
```
johnnovak, line 4531: How about `for (const auto &arg : parameters)`?

weirddan455:
I don't know why that Mac build is failing.  It won't even show me an error of what went wrong.  This PR should only affect Windows so I assume it's just being flaky.

src/gui/sdlmain.cpp
johnnovak: The functional programmer in me is crying 😢 

```
command_line = parameters.drop(1).join(" ")
```

src/gui/sdlmain.cpp
```diff
@@ -4533,27 +4544,46 @@ void restart_program(std::vector<std::string> & parameters) {
...
-			//However keep the first argument of the passed argv (newargs) with quotes, as else repeated restarts go wrong.
-			if(execvp(edit.c_str(), newargs) == -1) E_Exit("Restarting failed");
-		}
-#endif
+	// Input structure with a bunch of stuff we probably don't care about.
+	// Zero it out to use defaults save for the size field.
+	STARTUPINFO startup_info = {};
+	startup_info.cb = sizeof(startup_info);
+
+	// Output strcture we also don't care about.
```
johnnovak, line 4564: typo: structure

src/gui/sdlmain.cpp
```diff
@@ -4533,27 +4544,46 @@ void restart_program(std::vector<std::string> & parameters) {
...
+			nullptr, // lpThreadAttributes  - nullptr for default
+			FALSE, // Do not inherit handles
+			DETACHED_PROCESS, // Creation flags. DETACHED_PROCESS fixes https://github.com/dosbox-staging/dosbox-staging/issues/3346
+			nullptr, // Environment variables - nullptr to use parent's enviornment
+			nullptr, // Current Directory - nullptr to use parent's current directory
+			&startup_info,
+			&process_information)) {
+		E_Exit("Restarting failed");
+	}
+	ExitProcess(ERROR_SUCCESS);
```
johnnovak, line 4580: Wouldn't it be better to extract these into vars and put line comments above them? I really dislike inline comments, especially when they're long.

E.g.:

```cpp
// nullptr for default
const LPSECURITY_ATTRIBUTES lpProcessAttributes = nullptr;

File: src/gui/sdlmain.cpp [hunk 4]
```diff
@@ -4533,27 +4544,46 @@ void restart_program(std::vector<std::string> & parameters) {
...
+#else
+	if(execvp(newargs[0], newargs) == -1) {
```
johnnovak, line 4582: Space after `if`

johnnovak (approved):
Nice quick fix @weirddan455 🎖️ Just some style comments.

I guess I could test it on Win10, but that won't prove much. We'll need @kklobe or the guy who raised it to test this on Win11 (once CI is green).

src/gui/sdlmain.cpp
weirddan455: Is that line supposed to be real? It would be nice but won't compile so I assume not.

weirddan455: Changed to auto

src/gui/sdlmain.cpp
weirddan455: Fixed

src/gui/sdlmain.cpp
weirddan455: Done.  Just had to turn on my caps lock and let my cat walk on my keyboard to get some of those Win32 type names correct.

src/gui/sdlmain.cpp
weirddan455: Fixed

src/gui/sdlmain.cpp
johnnovak: > Is that line supposed to be real? It would be nice but won't compile so I assume not.

It was a joke 😆 That's how you do it in nicer languages like Kotlin or Scala, etc.

src/gui/sdlmain.cpp
johnnovak: > Done. Just had to turn on my caps lock and let my cat walk on my keyboard to get some of those Win32 type names correct.

Hungarian notation, man. Well, I'm Hungarian, but those prefixes still read like gibberish to me 😄

PR #3351, opened 2024-01-25
Fix console color after restart

### LINKED ISSUE #3346: Restart facility is unstable when launched from Windows GUI

weirddan455:
# Description

Regression from PR #3349 I was under the incorrect assumption that the Windows console was being created via an `AllocConsole` call here:

https://github.com/dosbox-staging/dosbox-staging/blob/0d3c555cb16028face41c0546326503b0a409229/src/gui/sdlmain.cpp#L4841-L4864

It turns out this is not the case.  Staging is being built as a console app meaning Windows creates a console for us before we even get to main.  This means on first run, the `AllocConsole` call fails because we already have a console.  After restart, we were successfully creating a console here but it was probably not using the correct mode for handling color escape codes (https://learn.microsoft.com/en-us/windows/console/setconsolemode).  There's also the issue that we're creating this console after Loguru initializes so the beginning of the logs are never visible.

We probably want to re-visit the above code in the future but the simple fix is to just use the `CREATE_NEW_CONSOLE` flag instead of `DETACHED_PROCESS`.  This both does not inherit the old console and creates a new one for the child process to use.

## Related issues

Fixes #3346

# Manual testing

Used Ctrl + Alt + Home to restart and verified the console has color afterwards.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4567,8 +4567,10 @@ void restart_program(std::vector<std::string> & parameters) {
 
 	const WINBOOL inherit_handles = FALSE;
 
-	// DETACHED_PROCESS fixes https://github.com/dosbox-staging/dosbox-staging/issues/3346
-	const DWORD creation_flags = DETACHED_PROCESS;
+	// CREATE_NEW_CONSOLE fixes a bug where the parent process exiting kills the child process.
+	// This can manifest when you use the "restart" hotkey action to restart DOSBox.
+	// https://github.com/dosbox-staging/dosbox-staging/issues/3346
+	const DWORD creation_flags = CREATE_NEW_CONSOLE;
 
 	// nullptr to use parent's environment
 	const LPVOID environment_variables = nullptr;


johnnovak (approved):

src/gui/sdlmain.cpp
```diff
@@ -4567,8 +4567,9 @@ void restart_program(std::vector<std::string> & parameters) {
 
 	const WINBOOL inherit_handles = FALSE;
 
-	// DETACHED_PROCESS fixes https://github.com/dosbox-staging/dosbox-staging/issues/3346
-	const DWORD creation_flags = DETACHED_PROCESS;
+	// CREATE_NEW_CONSOLE fixes a bug where the parent process exiting kills the child process.
```
johnnovak, line 4570: Very economical with words, @weirddan455 😄 Can you add this at least to explain the "why" behind it?

```
This can manifest when you use the "restart" hotkey action to restart DOSBox.
```

johnnovak: Thanks man 😄 Just thinking of future maintainers, that's all.

PR #3361, opened 2024-01-30
Basic CMake support

Arcnor:
Hello!

I understand DOSBox Staging uses Meson so this is probably not a welcome PR, but in my specific situation I wanted to work on the project using CLion, and it was very difficult because CLion's Meson support is very basic in its current state (plus trying to compile with Meson on Windows was failing because of `pkg-config`, at least).

When used alongside `vcpkg`, these changes compile cleanly on Windows, although there hasn't been any effort spent towards adding any kind of optimization the "official" (MSVC?) build system might have (adding `zlib-ng` for example should be trivial).

I'm just opening this PR in case somebody searches the project for CMake for similar reasons to mine, so they can check my fork if they want.

Feel free to close unless you want to discuss any specifics, and thanks for all your work on DOSBox Staging!


src/gui/sdlmain.cpp
```diff
@@ -4565,7 +4565,7 @@ void restart_program(std::vector<std::string> & parameters) {
 	// nullptr for default
 	const LPSECURITY_ATTRIBUTES thread_attributes = nullptr;
 
-	const WINBOOL inherit_handles = FALSE;
+	const BOOL inherit_handles = FALSE;
```
weirddan455, line 4568: I just merged a fix for this a couple hours ago so might be a conflict when you rebase.

weirddan455:
The only way Meson currently works in Windows (at least for this project) is using MSYS2.  See here for build instructions:  https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/build-windows.md

We recently removed "official" support for the MSVC build files due to some issues with CI and it being somewhat of a maintenance burden to maintain two different build systems.

If you're just trying to use CLion, maybe give this a read:  https://www.jetbrains.com/help/clion/compilation-database.html

If you use the Meson + MSYS2 method, Meson will generate a `compile_commands.json` file which should give you most of the IDE features.

I personally prefer CMake for my own projects due to better MSVC support and most open source libraries using it (making it easier to include dependencies without depending on third-party Meson wraps).  Meson does have better syntax though.  I wasn't around when the decision was made to switch to Meson over whatever ugly thing we were using beforehand (plain Makefiles maybe?).

I'll let @johnnovak weight in on what he wants to do with this.  It would take a lot more work than what you have here to fully replace Meson.  For starters, vcpkg is only used by the MSVC builds.  For every other platform (including MSYS2 Windows) we either dynamically link our dependencies from system libraries or compile them from source via Meson wraps.

Nice job getting it to build in CMake regardless though :+1:

Arcnor:
> The only way Meson currently works in Windows (at least for this project) is using MSYS2. See here for build instructions: https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/build-windows.md

I'm actually using MSYS2 (well, the "git for windows" packages), but like I said above, there are problems with at least pkg-config, which might be fixed by reinstalling everything, but it's not worth modifying my whole system for this :).
 
> We recently removed "official" support for the MSVC build files due to some issues with CI and it being somewhat of a maintenance burden to maintain two different build systems.

Makes sense!

> If you're just trying to use CLion, maybe give this a read: https://www.jetbrains.com/help/clion/compilation-database.html

Yeah, this was what I was going to try in the first place, but as mentioned couldn't build with Meson. Also, adding/removing files is a bit of a pain, as I'd have to add them to meson first and do the whole dance for each change.

> I'll let @johnnovak weight in on what he wants to do with this. It would take a lot more work than what you have here to fully replace Meson. For starters, vcpkg is only used by the MSVC builds. For every other platform (including MSYS2 Windows) we either dynamically link our dependencies from system libraries or compile them from source via Meson wraps.

Well, I think there is a point in using `vcpkg` everywhere (consistency at least), but as mentioned above the purpose of this was not to replace Meson, that doesn't make sense, just wanted to have it known in case it's useful for somebody else 🙂 .

> Nice job getting it to build in CMake regardless though 👍

Thanks!

johnnovak:
@Arcnor The bigger picture is that Windows is only 33% of our supported platforms. Our Meson + wraps solution works on Windows/MSYS2, macOS, and Linux. vcpkg throws extra maintenance burden into the mix, for not much benefit. Same deal with MSVC.

Same for CMake—the team (before my time) decided to use Meson, so we'll stick with it. There's something in deciding to use something, and sticking with it. There are always personal preferences, and every technology has issues, but the cost of switching to something else usually far outweighs the (often dubious) benefits.

So in this case, you're using CMake due to personal preference, which is completely fine for your personal offshoot projects, of course, but I'm not seeing us switching to anything else from Meson anytime soon.

I'm not an expert in this area, but there seem to be problems with the vcpkg releases of the slirp library we use for networking, so keep an eye out for that.

> I'm just opening this PR in case somebody searches the project for CMake for similar reasons to mine, so they can check my fork if they want.

That's thoughtful of you, thanks for that. I was inclined to say let's merge it into `contrib` as a not officially supported build method, but the CMake approach seems to be invasive (lots of source files spread around the `src` tree). And it would slowly become unusable anyway if it's not actively maintained. So I think I'll politely decline this PR.

> Feel free to close unless you want to discuss any specifics, and thanks for all your work on DOSBox Staging!

Thanks, and you're welcome to contribute in other areas in the future that are not related to switching out the whole build system! 😅

Arcnor:
Thanks @johnnovak , although I'm obliged to once again state that I never opened this with the intention of changing the existing build system, as I stated a few times 😄 .

> I was inclined to say let's merge it into contrib as a not officially supported build method, but the CMake approach seems to be invasive (lots of source files spread around the src tree)

This could be done, I just added every file in the same places as Meson because it simplifies things, but as you mention, it will be come obsolete if nobody maintains it, so no point.

> Thanks, and you're welcome to contribute in other areas in the future that are not related to switching out the whole build system! 😅

No worries, and thank you.

johnnovak:
@Arcnor After all, it seems we're migrating to using CMake on all platforms. Would you still be interested in helping out with this?

Arcnor:
@johnnovak sure, unless you're in some sort of hurry of course, free time for side projects is unfortunately limited like everybody else :)

What's your plan? At risk of repeating myself, this PR was never intended for merging, so there are things here (at least some) that will need to be done differently of course, so maybe you already have a rough idea of how to approach this migration?

johnnovak:
> @johnnovak sure, unless you're in some sort of hurry of course, free time for side projects is unfortunately limited like everybody else :)
> 
> What's your plan? At risk of repeating myself, this PR was never intended for merging, so there are things here (at least some) that will need to be done differently of course, so maybe you already have a rough idea of how to approach this migration?

Basically, getting rid of Meson altogether is the goal and replacing it with using CMake directly. With Meson we depend on the availability of the Meson wraps, or if no such thing exist, we need to roll our own. Which is a pain. Plus Meson in general is not very compatible with more intricate CMake build files (e.g., we ran into such a thing with zlib-ng; Meson cannot do different compiler flags per file which is the ideal way to compile zlib-ng).

There's a guy on our team who'll probably pick and whatever you do and modify/continue where you left off. It's just if you've already started, it would be silly _not_ to use your work as a basis for our stuff, that's all 😄 

We're not really in a hurry, but ideally this should happen before the end of the year.

Arcnor:
Ok, so how will this work technically? Do you want to keep a branch for CMake in parallel until it works 100%, or do you want to merge to `main` bit by bit?

Do you have any preference on the location of the CMakeLists.txt files? In my current PR I added one per subdirectory, which is what I prefer, but there are other solutions (like having those files on a completely separate directory mimicking the source layout, or even have everything on a single CMakeLists.txt, although both of those solutions are less "clean" for me)

I can start polishing what I had here, and the plan could roughly be:
1. Adding all the options you had with Meson (I don't think I had any options on my PR) as well as all the specific compiler flags
2. Making it cross-platform
3. Adding install support

johnnovak:
> Ok, so how will this work technically? Do you want to keep a branch for CMake in parallel until it works 100%, or do you want to merge to `main` bit by bit?

The guiding principle is that we still should be able to run our CI workflows and generate Windows, Linux, and macOS dev builds after every merge to `main`. So it can go in bit by bit, but we need to keep the dev builds going, we can't break that.

> Do you have any preference on the location of the CMakeLists.txt files? In my current PR I added one per subdirectory, which is what I prefer, but there are other solutions (like having those files on a completely separate directory mimicking the source layout, or even have everything on a single CMakeLists.txt, although both of those solutions are less "clean" for me)

Okay, so the catch is, I'm not even a beginner when it comes to CMake 😆 But maybe that work for our favour because at least I can look at things with fresh eyes and knowledge doesn't cloud my vision 😎 

More seriously, I think the per subdir approach is fine. I don't really like the other two options you brought up. 

> I can start polishing what I had here, and the plan could roughly be:
> 
>     1. Adding all the options you had with Meson (I don't think I had any options on my PR) as well as all the specific compiler flags
> 
>     2. Making it cross-platform
> 
>     3. Adding install support

That sounds like an excellent plan, man! 💯 🎖️ 

@shermp @kklobe @weirddan455 , what do you think of all this? Any suggestions / ideas / etc?

weirddan455:
>  Do you want to keep a branch for CMake in parallel until it works 100%, or do you want to merge to main bit by bit?

Whatever works best for you.  As @johnnovak said, we need to keep CI on Meson and keep Meson around for development until CMake is ready.  I don't foresee major problems with having two build systems in the tree though so if you're just committing new CMake files, it would be good to merge so that we can test.  If you need to make source changes, just make sure it doesn't break our existing build system.  We can then keep both around in `main` until we're confident that CMake is good to go then we can drop Meson and switch CI over to CMake.

>  Do you have any preference on the location of the CMakeLists.txt files? In my current PR I added one per subdirectory

That sounds fine to me.  That was also the approach we took with Meson.

Glad you're interesting in working on this @Arcnor  Build system stuff has been a pain point for us recently.

Arcnor:
Thank you both for commenting.

My preference would be small PRs (except the first one, which might be this one, probably a bit bigger than the rest) that get merged with main ASAP, so it doesn't break. And of course, Meson will continue working. I even checked how I did certain things on this PR and for example I read the config file and replace `#mesondefine` with `#cmakedefine` on the fly to avoid changing any files, so this should work fine :).

Given there is some sort of consensus and unless somebody else objects, I'll keep the one file per directory, mirroring `meson.build` in each directory.

I started doing some tests today to rebase my changes on top of `main`, but it seems `gitlab.freedesktop.org` was migrating their Gitlab, so `slirp` couldn't be downloaded :). I'll try again later or tomorrow.

One more question, and although as I mentioned I'll tackle cross-platform issues later, do you want to use `vcpkg` for all platforms to keep it consistent, or do you expect non-Windows platform to provide packages using their own package managers? My own preference is using `vcpkg` everywhere, again for consistency's sake, but you probably have your own preference and history on that matter.

weirddan455:
>  My preference would be small PRs (except the first one, which might be this one, probably a bit bigger than the rest) that get merged with main ASAP, so it doesn't break.

That's probably for the best, I agree.

>  One more question, and although as I mentioned I'll tackle cross-platform issues later, do you want to use vcpkg for all platforms to keep it consistent, or do you expect non-Windows platform to provide packages using their own package managers? My own preference is using vcpkg everywhere, again for consistency's sake, but you probably have your own preference and history on that matter.

That's a bit tricky.  For Windows, I think we would just always use vcpkg.  For Linux, I would like to help out distro maintainers as much as possible and they have policies about dynamic linking with system packages wherever possible.

Ideally, we would be able to build either way.  With Meson, it first checks for installed system libraries and if it doesn't find it, it falls back to using a Meson wrap to build the missing dep from source.  There's also a flag `-Dforce_fallback_for=whatever_dep`.  Once we get fully on vcpkg, we may want that to be the default but also provide a build flag to link with system deps instead.

If that's too much trouble, your initial PR can just use vcpkg only and we can work that out at a later time.

johnnovak:
@Arcnor Thank you very much for dedicating your time to get the build system in a bettet shape, we are a small team and we appreciate your help a lot.

I'm in full agreement with @weirddan455 on the direction:

- CMake file per subdir
- Small PRs merged to main and keep Meson working until we do the 100% switchover
- vcpkg on all three platforms first; that's how we will build on CI and that's the recommended local build method
- Optional fallback to system libs on Linux only as the last step, mainly to help distro maintainers

Looking forward to your enhancements, exciting times! 😎 🥳 🚀

johnnovak:
Thanks @Arcnor. I will leave it open for a day to a give for the other guys to comment. I am going to give you at least contributor privileges after this is merged so we won't need to constantly approve the workflow runs.

FeralChild64 (commented):
I suggest to update BUILD.md file with instruction how to build using CMake.

johnnovak:
> I suggest to update BUILD.md file with instruction how to build using CMake.

I think that should be added when everything is mostly working on all platforms, so around the end.

shermp:
Just a question about speexdsp, is pkgconfig the only way to find it?

CMakeLists.txt
```diff
@@ -0,0 +1,187 @@
...
+)
+add_dependencies(dosbox copy_assets)
+
+find_package(PkgConfig REQUIRED)
+
+find_package(SDL2 CONFIG REQUIRED)
+
+# TODO: 0.80.0 release notes mention "sdl2-image is now a dependency" but I cant find any code using it...
+#find_package(SDL2_image CONFIG REQUIRED)
+#target_link_libraries(dosbox PRIVATE $<IF:$<TARGET_EXISTS:SDL2_image::SDL2_image>,SDL2_image::SDL2_image,SDL2_image::SDL2_image-static>)
```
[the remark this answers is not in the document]
Arcnor: Ok, will do soon.

Arcnor:
> Just a question about speexdsp, is pkgconfig the only way to find it?

Yeah, it comes from vcpkg, but it doesn't provide CMake targets. There are actually many vcpkg packages like that, and it's annoying, but fine :).

src/libs/zmbv/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_library(zmbv STATIC zmbv.cpp)
+
+pkg_check_modules(ZLIB_NG REQUIRED IMPORTED_TARGET zlib-ng)
+
+target_include_directories(zmbv PUBLIC ..)
+# FIXME: "ZLIB_NG_LIBRARIES" points to the wrong library name (z-ng), use this simpler way when the port is fixed
```
Arcnor, line 6: VCPKG has fixed this issue, although I haven't tried it yet, so I'll update this ASAP. Only "issue" is that I need to bring the vcpkg baseline to a very recent commit, just need "approval" from any of you to do so.

shermp: Is the 2024.05.24 vcpkg new enough? I've just submitted #3761 which bumps the baseline to that release.

Arcnor: Probably not, we need a baseline from yesterday.

johnnovak: @Arcnor Are you waiting for us to push the baseline forward to be able to finish this PR?

CMakeLists.txt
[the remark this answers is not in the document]
johnnovak: Are you going to address this @Arcnor ?

dreamer:
@Arcnor I am downloading CLion to understand what problem do you experience with CLion+Meson - let's start listing them and report bugs where they are appropriate.

I don't use this IDE myself, but it claims to [support Meson on all platforms](https://blog.jetbrains.com/clion/2021/01/working-with-meson-in-clion-using-compilation-db/). Maybe only some additional steps are needed to configure the project - in such case we should definitely improve our documentation in [docs](https://github.com/dosbox-staging/dosbox-staging/tree/main/docs).

@weirddan455 wrote:
> I personally prefer CMake for my own projects due to better MSVC support and most open source libraries using it (making it easier to include dependencies without depending on third-party Meson wraps). Meson does have better syntax though.

Same; Meson has [VS support](https://mesonbuild.com/Using-with-Visual-Studio.html), but we haven't even tested it. The way forward is to improve it :) If you can test it and it does not work - let's create a new issue and we will start tracking bugs in other projects.

Meson has not only better syntax; it also has type support (!), test support, instrumentation tools support, good system libs support, *portable* dependencies/wraps, plenty of built-in features (e.g. ccache support) and friendly community (I worked with CMake before - for years - debugging other people's cmake scripts is not fun... especially when everything is string-ified as if we still used Makefiles). The only advantage CMake has is popularity... But Meson is also [pretty popular in open source world](https://mesonbuild.com/Users.html).

> I wasn't around when the decision was made to switch to Meson over whatever ugly thing we were using beforehand (plain Makefiles maybe?).

I did this transition; even worse than plain Makefiles - previous buildsystem was based on Autotools. CMake was being considered and I even developed prototype CMake buildsystem back then to evaluate. Meson was just better.

@johnnovak I am very much opposed to the plan of replacing Meson. Transition to another buildsystem should be done only when there is *overwhelming* benefit of doing so. And merely covering 100% of the same functionality is not enough - changing buildsystem affects all packagers on all Linux distros. But I am not active maintainer ATM, so I don't have the final say. In my opinion - the way forward is to improve interop between Meson and IDEs, wherever we know there is friction.

CMakeLists.txt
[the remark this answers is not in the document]
Arcnor: Sure, sorry about the delay, will do tonight.

Arcnor:
@dreamer thanks for your interest. I unfortunately don't have the time to improve meson myself, besides the effort of the tickets I've already filled with CLion, so I'm not sure it will be worth the effort downloading and testing CLion for this.

Suffice to say, CLion cannot open and build this project successfully, or at least not when I first opened this PR, and that was the reason for me to fill this, but moving to CMake was not motivated by IDE support AFAIK.

In any case, I'm really not interested in politics, so please just let me know if I should continue with this PR or not :)

dreamer:
> @dreamer thanks for your interest. I unfortunately don't have the time to improve meson myself, besides the effort of the tickets I've already filled with CLion, so I'm not sure it will be worth the effort downloading and testing CLion for this.

Oh, of course - I did not expect you to work on improving Meson :) I expect **us** to communicate with Meson project about the issues, and perhaps for myself to improve any open source project that needs fixing. As for CLion - can you share the links to their bugtracker?

> Suffice to say, CLion cannot open and build this project successfully, or at least not when I first opened this PR, and that was the reason for me to fill this, but moving to CMake was not motivated by IDE support AFAIK.

If we cannot be built in CLion then I consider it a bug.

> In any case, I'm really not interested in politics, so please just let me know if I should continue with this PR or not :)

As I mentioned earlier, I am not an active maintainer ATM, so I don't have a final say here. In my opinion when CMake buildsystem will be on-par with features supported by our Meson buildsystem the final result will be very ugly and we will need to deal with new problems that just don't exist in Meson. Of couse you can prove me wrong :)

Arcnor:
> Oh, of course - I did not expect you to work on improving Meson :) I expect **us** to communicate with Meson project about the issues, and perhaps for myself to improve any open source project that needs fixing. As for CLion - can you share the links to their bugtracker?

https://youtrack.jetbrains.com/issues?q=tag:%20Meson

> As I mentioned earlier, I am not an active maintainer ATM, so I don't have a final say here. In my opinion when CMake buildsystem will be on-par with features supported by our Meson buildsystem the final result will be very ugly and we will need to deal with new problems that just don't exist in Meson. Of couse you can prove me wrong :)

Well, nothing has been more true on software matters than "beauty is in the eye of the beholder", so I doubt I can change your point of view there, and I'm also not sure if you need to change it. But yeah, I don't see many issues with matching what Meson is already doing, and for what there might be issues I'm sure it could be solved in a slightly different way, as always with this kind of thing.

Anyway, @johnnovak can you tell me if you guys want to continue with this or not? Will update the PR in a few hours if you are.

johnnovak:
Thank you for stopping by and offering your viewpoint @dreamer. We appreciate that you introduced Meson to the project initially, and it has definitely served it well for a while.

However, in the last 8 months or so since kcgen unfortunately had to withdraw from the project, maintaining the build system became largely @kklobe's and my responsibility (with help from the entire core team, of course). In short, Meson resulted in a lot of complications that we wouldn't have with CMake. We've been considering the pros and cons of switching to CMake, and I was actually against it initially, just like you. However, over time, the majority of the core team reached the consensus that CMake is the way to go. If you are interested in the details, you can search back our conversations on Discord; all the info is there. Believe me, we're not doing this lightly and we have seriously considered the implications.

I certainly understand Meson is close to your heart as you introduced it to the project, so it was your "baby", so to speak 😄 But we've made our decision, and decision is to migrate to CMake as it solves what we're trying to do with minimum fuss. Thank you for accepting our decision.

@Arcnor Yes, please proceed with the PR, thank you 👍🏻

src/libs/zmbv/CMakeLists.txt
johnnovak: @Arcnor If you would be able to do that, that would be great, thanks.

CMakeLists.txt
```diff
@@ -0,0 +1,187 @@
...
+set_property(GLOBAL PROPERTY FILE_CONTENTS "${CONFIG_CONTENTS}")
+
+function(setConfig01 VAR VALUE)
+	get_property(fc GLOBAL PROPERTY FILE_CONTENTS)
+	string(REPLACE "#cmakedefine ${VAR}" "#cmakedefine01 ${VAR}" fc "${fc}")
+	set_property(GLOBAL PROPERTY FILE_CONTENTS "${fc}")
+	set(${VAR} ${VALUE} PARENT_SCOPE)
+endfunction()
+
+if (CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
```
dreamer, line 81: My pretty standard Linux build failed here:

```
CMake Error at CMakeLists.txt:84 (message):
  Unknown processor x86_64
```

dreamer (changes requested):

CMakeLists.txt
Arcnor: Of course, if you read the whole PR you'll notice I said this will *only* work for Windows right now.

johnnovak: @dreamer We agreed with @Arcnor to merge in his work in stages in many small PRs, versus one big PR when everything is working. So yeah, this is the initial Windows only PR.

dreamer (changes requested):
Please add some instructions in `docs/` on how to use another buildsystem. Also, if `CMakeLists.txt` is going to be merged to `main` then you must update `README.md`/`BUILD.md` to indicate the limitations/intended use of this.

Arcnor:
> Please add some instructions in `docs/` on how to use another buildsystem. Also, if `CMakeLists.txt` is going to be merged to `main` then you must update `README.md`/`BUILD.md` to indicate the limitations/intended use of this.

Thanks, but again, this was already discussed above. The decision was to do this after it's actually useable.

johnnovak:
@dreamer The build instructions in BUILD.md, README.md, and the platform-specific build instructions spell out how to build with Meson, which is our current way of building Staging, and continues to be until it's ready to switch over to CMake. When the CMake method is good enough to use on all platforms, we will document how to use it.

People should do what those build instructions say and ignore any other files present in the directory tree that the building guides do not mention.

But, I can also see some people might get confused by the presence of those CMake files, fair point. I will put a short warning at the top of the build related readme files shortly after this PR goes in. I don't want to test @Arcnor 's patience any further 😄 We, the core team are very appreciative of your efforts ❤️ 

Please do the final steps you planned to do on this @Arcnor, then it's merge time 🎉 Let us not overload this PR with discussions any further for a work-in-progress, internal-only feature not ready yet for public consumption

johnnovak (approved):

CMakeLists.txt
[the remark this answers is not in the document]
Arcnor: This has been removed

src/libs/zmbv/CMakeLists.txt
Arcnor: I've updated the baseline to one from yesterday. IT's maybe too recent, you let me know, otherwise I can pick one from 9 days ago.

Arcnor:
Rebased with `main` again, checked that everything compiles (had to add some new file in `libdos`, but everything else is the same), let me know if there is anything missing.

johnnovak:
@Arcnor It seems you've pushed the vcpkg baseline to a commit from yesterday, and the CI runners apparently don't have a new enough vcpkg installed.

In general, we should be careful with pushing vcpkg to the very bleeding edge baseline. Probably a 2-4 weeks old baseline is a better idea, that way we hopefully won't be hit with all the early-adopter problems.

Could you please push the vcpkg baseline a bit back to make the MSVC workflow compile again?

Arcnor:
> @Arcnor It seems you've pushed the vcpkg baseline to a commit from yesterday, and the CI runners apparently don't have a new enough vcpkg installed.

Yeah, I mentioned this on the conversation over the issue. I've pushed again with the very minimum I require for it to work. If this doesn't work then I'll have to bring back my hack until we can move to the proper baseline, which is ~10 days old.

johnnovak:
> > @Arcnor It seems you've pushed the vcpkg baseline to a commit from yesterday, and the CI runners apparently don't have a new enough vcpkg installed.
> 
> Yeah, I mentioned this on the conversation over the issue. I've pushed again with the very minimum I require for it to work. If this doesn't work then I'll have to bring back my hack until we can move to the proper baseline, which is ~10 days old.

Thanks, it's all green now 🚀

Arcnor:
Awesome. As per my plan above, I'll (slowly) start adding the missing options, need to check what those are.

johnnovak:
> Awesome. As per my plan above, I'll (slowly) start adding the missing options, need to check what those are.

Sure, no rush, whatever is convenient for you 😄

PR #3369, opened 2024-02-02
Introduce unittests for the Value class

japsmits:
# Description

Added a unittest for the Value class as implemented in setup.h and setup.cpp.

I currently have some work on this class, so it's good to have a working unittest first. And since it's my first commit to dosbox-staging, it seems like a reasonable place to start.

In the process, I introduced the Overload pattern, which is added to support.h.

# Manual testing

This is an addition to existing unittests, and it passes as is.

# Checklist

I have:

### DIFF
--- a/include/support.h
+++ b/include/support.h
@@ -330,4 +330,23 @@ template <typename T>
 std::pair<std::unique_ptr<T[]>, T*> make_unique_aligned_array(
         const size_t byte_alignment, const size_t req_elems, const T& init_val = {});
 
+// This struct can be used in combination with std::visit and std::variant to
+// define code for each type specified in the variant.
+//
+// Example:
+//
+// std::variant<bool, int> data = true;
+// std::string typename = std::visit(Overload{
+//     [](bool) { return "bool"; },
+//     [](int)  { return "int"; },
+// }, data);
+//
+// typename now contains the value "bool".
+template <typename... Ts>
+struct Overload : Ts... {
+	using Ts::operator()...;
+};
+template <class... Ts>
+Overload(Ts...) -> Overload<Ts...>;
+
 #endif
--- a/include/support.h
+++ b/include/support.h
@@ -47,13 +47,13 @@
 #include "std_filesystem.h"
 
 #ifdef _MSC_VER
-#define strcasecmp(a, b) _stricmp(a, b)
+#define strcasecmp(a, b)     _stricmp(a, b)
 #define strncasecmp(a, b, n) _strnicmp(a, b, n)
 #endif
 
 #ifdef PAGESIZE
 // Some platforms like ppc64 have page sizes of 64K, so uint16_t isn't enough.
-constexpr uint32_t host_pagesize = { PAGESIZE };
+constexpr uint32_t host_pagesize = {PAGESIZE};
 #else
 constexpr uint16_t host_pagesize = 4096;
 #endif
@@ -88,15 +88,17 @@ char drive_letter(uint8_t index);
  *  This function does not attempt to capture exceptions that may
  *  be thrown from std::stod(...)
  */
-template<typename T>
-T to_finite(const std::string& input) {
+template <typename T>
+T to_finite(const std::string& input)
+{
 	// Defensively set NaN from the get-go
-	T result = std::numeric_limits<T>::quiet_NaN();
+	T result          = std::numeric_limits<T>::quiet_NaN();
 	size_t bytes_read = 0;
 	try {
 		const double interim = std::stod(input, &bytes_read);
-		if (!input.empty() && bytes_read == input.size())
+		if (!input.empty() && bytes_read == input.size()) {
 			result = static_cast<T>(interim);
+		}
 	}
 	// Capture expected exceptions stod may throw
 	catch (...) {
@@ -112,25 +114,23 @@ std::string get_basename(const std::string& filename);
 
 // Select the nearest unsigned integer capable of holding the given bits
 template <int n_bits>
-using nearest_uint_t = \
-        std::conditional_t<n_bits <= std::numeric_limits<uint8_t>::digits, uint8_t,
+using nearest_uint_t = std::conditional_t<
+        n_bits <= std::numeric_limits<uint8_t>::digits, uint8_t,
         std::conditional_t<n_bits <= std::numeric_limits<uint16_t>::digits, uint16_t,
-        std::conditional_t<n_bits <= std::numeric_limits<uint32_t>::digits, uint32_t,
-        std::conditional_t<n_bits <= std::numeric_limits<uint64_t>::digits, uint64_t,
-        uintmax_t>>>>;
+                           std::conditional_t<n_bits <= std::numeric_limits<uint32_t>::digits, uint32_t,
+                                              std::conditional_t<n_bits <= std::numeric_limits<uint64_t>::digits,
+                                                                 uint64_t, uintmax_t>>>>;
 
 // Select the next larger signed integer type
 template <typename T>
 using next_int_t = typename std::conditional<
-        sizeof(T) == sizeof(int8_t),
-        int16_t,
+        sizeof(T) == sizeof(int8_t), int16_t,
         typename std::conditional<sizeof(T) == sizeof(int16_t), int32_t, int64_t>::type>::type;
 
 // Select the next large unsigned integer type
 template <typename T>
 using next_uint_t = typename std::conditional<
-        sizeof(T) == sizeof(uint8_t),
-        uint16_t,
+        sizeof(T) == sizeof(uint8_t), uint16_t,
         typename std::conditional<sizeof(T) == sizeof(uint16_t), uint32_t, uint64_t>::type>::type;
 
 template <typename cast_t, typename check_t>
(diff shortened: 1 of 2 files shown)


johnnovak:
First of all, welcome to the project @japsmits. I like your enthusiasm 🎉 

> Added a unittest for the Value class as implemented in setup.h and setup.cpp.

That's great, such unit test additions are always welcome.

> I currently have some work on this class, so it's good to have a working unittest first. And since it's my first commit to dosbox-staging, it seems like a reasonable place to start.

I highly suggest checking with us first before doing larger-scale reworks that affect the whole codebase. Configuration handling in one of these areas. We're planning to switch the entire config handling mechanism out to something better—it is still to be decided to what exactly.

Consequently, we're only doing maintenance type of work to the current config stuff as it will go away, eventually.

> In the process, I introduced the Overload pattern, which is added to support.h.

We're not against using some more "advanced" C++ features when it's warranted (I started using `std::variant` and `std::visit` myself a while ago), but in this case the ratio of extra complexity added by these features versus the actual test code is rather high.

Unless I'm missing something, I think you could write the exact same test without visit, variant, templates and the new Overload stuff. It will be a bit more verbose, but simpler overall, so it would be a net win. But feel free to explain what's the bigger picture here, I might be missing it 😄 

As a final comment, we generally try to do mass-scale reformats/non-functional cleanups as separate commits. So in this case, it would be a lot better to split this into two commits: 1) cleanup, 2) adding the new test.

japsmits:
> First of all, welcome to the project @japsmits. I like your enthusiasm 🎉

And thanks for your time and effort in writing a comment on my first ever PR on Github. 😄 

> 
> > Added a unittest for the Value class as implemented in setup.h and setup.cpp.
> 
> That's great, such unit test additions are always welcome.
> 
> > I currently have some work on this class, so it's good to have a working unittest first. And since it's my first commit to dosbox-staging, it seems like a reasonable place to start.
> 
> I highly suggest checking with us first before doing larger-scale reworks that affect the whole codebase. Configuration handling in one of these areas. We're planning to switch the entire config handling mechanism out to something better—it is still to be decided to what exactly.
> 
> Consequently, we're only doing maintenance type of work to the current config stuff as it will go away, eventually.

Agreed. My next commit will be something on the issue-list.

> 
> > In the process, I introduced the Overload pattern, which is added to support.h.
> 
> We're not against using some more "advanced" C++ features when it's warranted (I started using `std::variant` and `std::visit` myself a while ago), but in this case the ratio of extra complexity added by these features versus the actual test code is rather high.
> 
> Unless I'm missing something, I think you could write the exact same test without visit, variant, templates and the new Overload stuff. It will be a bit more verbose, but simpler overall, so it would be a net win. But feel free to explain what's the bigger picture here, I might be missing it 😄 Also, the new implementation will probably just use subclassing for the different types, not chucking every possible type into a Value class which is a bit questionable...

I would propose to keep the Overload struct in place, since it is mentioned in several places when looking for how-to's with respect to std::visit and std::variant. It might come in handy in the future. And I'm still surprised it's not part of std, but that's a different matter.

Would it be convenient to add something to our code documentation about the utilities in support.h? It already contains several convenient and/or useful things. And if so, what would be a good place?

In this case, I had no deeper underlying reason to do it this way other than that it was fun to write. :wink: I'll break up and simplify the testcase into simpler to read pieces, since I see where you're comming from. If no longer needed, it can be removed altogether.

> 
> As a final comment, we generally try to do mass-scale reformats/non-functional cleanups as separate commits. So in this case, it would be a lot better to split this into two commits: 1) cleanup, 2) adding the new test.

I realised a bit too late where this was comming from, but I'll split up the commit.

Actual processing of this comment will come next week, since we have a birthday-girl this weekend. 🎂

include/support.h
```diff
@@ -330,4 +334,23 @@ template <typename T>
 std::pair<std::unique_ptr<T[]>, T*> make_unique_aligned_array(
         const size_t byte_alignment, const size_t req_elems, const T& init_val = {});
 
+// Allow for overload resolition for use with std::visit in combination with an
```
japsmits, line 337:
```suggestion
// Allow for overload resolution for use with std::visit in combination with an
```

FeralChild64:
Actually, I have a plan to introduce the _overload_ pattern too - for my upcoming `dir` command rework:

https://github.com/dosbox-staging/dosbox-staging/blob/2a3b319e5aa9ac58139339529a0e1c25e5fee4d3/src/shell/shell_cmds.cpp#L53

johnnovak:
> Actually, I have a plan to introduce the _overload_ pattern too - for my upcoming `dir` command rework:
> 
> https://github.com/dosbox-staging/dosbox-staging/blob/2a3b319e5aa9ac58139339529a0e1c25e5fee4d3/src/shell/shell_cmds.cpp#L53

Yeah, that's a more justified use case 😄

include/support.h
```diff
@@ -330,4 +330,23 @@ template <typename T>
...
+//
+// std::variant<bool, int> data = true;
+// std::string typename = std::visit(Overload{
+//     [](bool) { return "bool"; },
+//     [](int)  { return "int"; },
+// }, data);
+//
+// typename now contains the value "bool".
+template <typename... Ts>
+struct Overload : Ts... {
```
johnnovak, line 350: minor: Probably lowercase `overload` would be better?

include/support.h
```diff
@@ -330,4 +330,23 @@ template <typename T>
 std::pair<std::unique_ptr<T[]>, T*> make_unique_aligned_array(
         const size_t byte_alignment, const size_t req_elems, const T& init_val = {});
 
+// Allow for overload resolition for use with std::visit in combination with an
+// std::variant. A list of functors for each type must be specified.
```
johnnovak, line 338: Maybe include the comment from the commit message here? Not everybody knows what a "functor" is, I think, so this is a bit cryptic.

include/support.h
```diff
@@ -330,4 +330,23 @@ template <typename T>
...
+//
+// Example:
+//
+// std::variant<bool, int> data = true;
+// std::string typename = std::visit(Overload{
+//     [](bool) { return "bool"; },
+//     [](int)  { return "int"; },
+// }, data);
+//
+// typename now contains the value "bool".
```
johnnovak, line 348: Nice example 🎉

johnnovak (commented):
Looks good now @japsmits, thanks for making these changes.

Yes, the Overload stuff can stay, apparently @FeralChild64 is planning to put it to good use later.

I've made one small remark about the comments, and the commit messages need a bit of a cleanup. I recommend reading the [How to Write a Git Commit Message](https://cbea.ms/git-commit/) article, we follow those recommendations. But here's in a nutshell what to do:

Your comment commit messages:

```
Introduced the Overload struct
Automatic reformatting
Introduce unittests for the Value class
Automatic reformatting
Typo was fixed
```

After normalising the grammar a bit and squashing the "Typo was fixed" into the first commit, the final result should look something like this:

```
Introduce the Overload struct
Reformat
Introduce unit tests for the Value class
Reformat
```

Thanks @japsmits, we're almost there, this is all just minor stuff 😄

japsmits:
@johnnovak : Thanks again! :smile:

After convincing Git of what I wanted, it looks good now. Could you do one final check? (And for the sake of procedure: who resolves conversations: the initiating author or the one who addresses the point made?)

johnnovak:
> @johnnovak : Thanks again! 😄
> 
> After convincing Git of what I wanted, it looks good now. Could you do one final check? (And for the sake of procedure: who resolves conversations: the initiating author or the one who addresses the point made?)

Thanks @japsmits. In your first few contributions it's probably best to let the reviewer resolve the comments, but we don't have hard rules about it 😄

johnnovak:
Don't worry about the MSVC situation, the CI situation is in a bit of a flux right now.

We'll merge this once the 0.81 final is out, so in a few week's time (we're in a code freeze period; only changes that contribute to the final should go into `main`, generally).

japsmits:
> Don't worry about the MSVC situation, the CI situation is in a bit of a flux right now.
> 
> We'll merge this once the 0.81 final is out, so in a few week's time (we're in a code freeze period; only changes that contribute to the final should go into `main`, generally).

Okay. :smile:. Feel free to merge it whenever it suits.

johnnovak:
> > Don't worry about the MSVC situation, the CI situation is in a bit of a flux right now.
> > We'll merge this once the 0.81 final is out, so in a few week's time (we're in a code freeze period; only changes that contribute to the final should go into `main`, generally).
> 
> Okay. 😄. Feel free to merge it whenever it suits.

We'll need to ask you to rebase when the time comes. We'll ping you! 😄

johnnovak:
@japsmits Can you rebase to main, please? Then it can be merged.

japsmits:
@johnnovak Done! 😄

include/support.h
```diff
@@ -330,4 +330,23 @@ template <typename T>
...
+template <typename... Ts>
+struct Overload : Ts... {
+	using Ts::operator()...;
+};
+template <class... Ts>
+Overload(Ts...) -> Overload<Ts...>;
+
```
weirddan455, line 355: Ok, so you wrote this templated thing but are you actually using it anywhere?  I'm a little confused about this part.

weirddan455:
I had to hit the button so CI will run since it is @japsmits first PR to the project.  Looks like @johnnovak already reviewed it so if he's happy, I think this can be merged.

I was just a little confused because I didn't see a use case for that template but more unit tests are always good.  Thanks for your first contribution :+1:

include/support.h
johnnovak: We agreed it can go in. @FeralChild64 will use it later, and it's handy to have it.

johnnovak:
Thanks for the extra review @weirddan455 , appreciated.

PR #3374, opened 2024-02-04
Fix invalid iterator crash in debug build

weirddan455:
# Description

Regression from https://github.com/dosbox-staging/dosbox-staging/commit/1bba63c6da9ff3dc77c77ff3a8d4e8172121173b although the root problem was around before this commit.

The CurMode iterator was being initalized in a global. This leads to undefined behavior as to when the constructor gets run. On Windows and Linux, this leads to a crash in VGA init code.

The iterator got invalidated here:

https://github.com/dosbox-staging/dosbox-staging/blob/5bec21b908e9aac3524a3e1b90332a27bcd18fbb/src/hardware/vga_s3.cpp#L767-L769

Then the Crash happened on this line:

https://github.com/dosbox-staging/dosbox-staging/blob/5bec21b908e9aac3524a3e1b90332a27bcd18fbb/src/hardware/vga_dac.cpp#L83

The fix is to init the iterator in INT10_Init after the vector has been modified.  This way the iterator stays valid.  It would also crash in a predictable way if someone tries to use it early but this does not happen in testing.

# Manual testing

- Stepped through with debugger to figure out wtf was going on.
- Modified code
- No more crash at startup :)

# Checklist

I have:

### DIFF
--- a/src/ints/int10.cpp
+++ b/src/ints/int10.cpp
@@ -732,6 +732,7 @@ static void SetupTandyBios(void) {
 }
 
 void INT10_Init(Section* /*sec*/) {
+	CurMode = std::prev(ModeList_VGA.end());
 	INT10_SetupPalette();
 	INT10_InitVGA();
 	if (IS_TANDY_ARCH) SetupTandyBios();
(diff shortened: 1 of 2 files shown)


johnnovak (approved):
Good catch, looks good @weirddan455 👍🏻

kklobe (approved):
nice work!

PR #3373, opened 2024-02-04
Pin MSYS2 SDL2 version to 2.26.5

johnnovak:
Hacking the system 😎

PR #3375, opened 2024-02-04
Runtime SDL version checks

johnnovak:
# Description

It's preferable to do the SDL version check at runtime in certain scenarios (when the API is the same between two versions, but the behaviour is different, so we need runtime-version dependent workarounds).

## Related issues

https://github.com/dosbox-staging/dosbox-staging/pull/3373

# Manual testing

Quick general regression testing plus making sure the mouse pointer is centered to the window after uncapturing the mouse:

- [x] Tested on macOS
- [x] Tested on Windows
- [ ] Tested on Linux

On Windows, I downloaded the ZIP installer, confirmed SDL 2.28.5 gets logged, then tested that uncapturing the mouse centers the pointer to the window at 125% DPI scale factor.

Repeated the same steps after replacing `SDL2.dll` with 2.26.5 -- the mouse pointer still got centered at 125% DPI scale factor.

# Checklist

I have:

### DIFF
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -337,4 +337,18 @@ struct SDL_Block {
 
 extern SDL_Block sdl;
 
+constexpr uint32_t sdl_version_to_uint32(const SDL_version version)
+{
+	return (version.major << 16) + (version.minor << 8) + version.patch;
+}
+
+bool is_runtime_sdl_version_at_least(const SDL_version min_version)
+{
+	SDL_version version = {};
+	SDL_GetVersion(&version);
+	const auto curr_version = sdl_version_to_uint32(version);
+
+	return curr_version >= sdl_version_to_uint32(min_version);
+}
+
 #endif
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2302,11 +2302,17 @@ void GFX_CenterMouse()
 	int width  = 0;
 	int height = 0;
 
-#if defined(WIN32) && !SDL_VERSION_ATLEAST(2, 28, 1)
-	const auto canvas_size_px = get_canvas_size_in_pixels(sdl.rendering_backend);
+#if defined(WIN32)
+	if (is_runtime_sdl_version_at_least({2, 28, 1})) {
+		SDL_GetWindowSize(sdl.window, &width, &height);
 
-	width  = canvas_size_px.w;
-	height = canvas_size_px.h;
+	} else {
+		const auto canvas_size_px = get_canvas_size_in_pixels(
+		        sdl.rendering_backend);
+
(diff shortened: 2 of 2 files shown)


src/gui/sdlmain.cpp
```diff
@@ -4870,15 +4876,16 @@ int sdl_main(int argc, char* argv[])
...
+			                "permonitorv2") == SDL_FALSE) {
+				LOG_WARNING("SDL: Failed to set DPI awareness flag");
+			}
+			if (SDL_SetHint(SDL_HINT_WINDOWS_DPI_SCALING, "1") ==
+			    SDL_FALSE) {
+				LOG_WARNING("SDL: Failed to set DPI scaling flag");
+			}
 		}
-#endif
 #endif // WIN32
```
weirddan455, line 4889: The intent of this one is different.  The macro `SDL_HINT_WINDOWS_DPI_AWARENESS` was added in 2.24 so the purpose of the compile time check is to allow compilation with older versions.  A runtime check doesn't make sense because presumably the function call will fail if the runtime SDL was too low and we would want to log that warning to the user.

These things are mostly here to support old versions of Debian Stable and such (whatever the oldest Linux distro we support is that may have an old SDL version).  As this check is inside an `#if defined(WIN32)` though, we can probably just remove the check all together.

Going forward (after this release) we may want to think about removing these once we get a fixed version solution working on Linux (vcpkg, Meson wrap, or equivalent).  These checks would really only benefit developers working on such distros.  The logic being if some distro is shipping Staging to end users and they aren't updating SDL, they probably aren't shipping the latest Staging either :slightly_smiling_face:

johnnovak: Good point @weirddan455, TBH I was a bit unsure about this yesterday, but was also tired because of all the build talk, so this is what happened 😅 

I agree with your suggestions for the future; reverting this to be a compile-time check for now.

weirddan455 (approved):
Looks good now. Thanks.

include/sdlmain.h
```diff
@@ -337,4 +337,18 @@ struct SDL_Block {
 
 extern SDL_Block sdl;
 
+inline uint32_t sdl_version_to_uint32(const SDL_version version)
```
kklobe, line 340: good candidate for constexpr?

`constexpr uint32_t sdl_version_to_uint32(const SDL_version version)`

johnnovak: Yeah, why not 😄

kklobe (approved):
looks good, approved even without the `constexpr` nitpick :grin:

johnnovak:
> looks good, approved even without the `constexpr` nitpick 😁

Thanks man 😎 🤘🏻

PR #3386, opened 2024-02-08
Fix crash on startup when `machine` is set to `hercules`

johnnovak:
# Description

@weirddan455's previous fix (https://github.com/dosbox-staging/dosbox-staging/commit/512efbd9aebc49b280e9903dc9d51b8f781f7e10) caused DOSBox crash right away with `machine = hercules`:

```
Stack trace:
10         0x18cfb10e0 start + 2360
9          0x100dded00 main + 36
8          0x10133926c sdl_main(int, char**) + 5844
7          0x100e8c7c4 Config::Init() const + 116
6          0x100e8ca38 Section::ExecuteInit(bool) + 244
5          0x1012f621c RENDER_Init(Section*) + 640
4          0x101504bec VGA_SetMonochromePalette(MonochromePalette) + 60
3          0x101504d78 VGA_SetHerculesPalette() + 308
2          0x1014f1904 VGA_DAC_CombineColor(unsigned char, unsigned char) + 92
1          0x1014f19a4 vga_dac_send_color(unsigned char, unsigned char) + 144
0          0x18d361a24 _sigtramp + 56
2024-02-08 18:21:32.222 | Signal: SIGSEGV
```

None of the other `machine` modes was affected 🤦🏻 Lesson learned: always test with all `machine` types after any video-related changes 😅 

The root cause of the problem is that we have all sorts of global initialisation going on, plus we have the section initialisers that are called in registration order. That's a bit shit; we should have a cycle-free dependency graph instead and no globals to ensure that every dependency gets created and initialised before the stuff that depends on on it (easier said than done, though, with this spaghetti mess... 🍝) So there's a real risk in fixing global initialisation because depending on where you move the init stuff within the chain of initialisers that "grew organically" and hang together like a house of cards, things might just blow up... A bit sad, really.

I could also "fix" this by moving `RENDER_AddConfigSection(control);` after `secprop->AddInitFunction(&INT10_Init);` in `dosbox.cpp`, but that's way too risky. Who knows what other bugs we might introduce by changing the init order of those sections... So, this more surgical approach seems more appropriate.

# Manual testing

Start with `machine = hercules` in the config — no crash.

Regression tested with all `machine` types (`svga_s3` is enough for VGA). Tested CGA composite as well.

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_dac.cpp
+++ b/src/hardware/vga_dac.cpp
@@ -76,12 +76,6 @@ static void vga_dac_send_color(const uint8_t palette_idx, const uint8_t color_id
 {
 	const auto rgb666 = vga.dac.rgb[color_idx];
 
-	// We might be in the middle of a mode change, so we can't use
-	// VGA_GetCurrentVideoMode() here. That's because the INT 10H mode
-	// change BIOS routine needs to set up the Palette and Color Registers
-	// which will trigger this function.
-	const auto bios_mode_number = CurMode->mode;
-
 	// In the automatic "video mode specific" CRT emulation mode, we want
 	// "true EGA" games (EGA modes with the default EGA palette on VGA) to
 	// use the single-scanline EGA shader. But VGA games that just happen to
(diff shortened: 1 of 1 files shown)


src/hardware/vga_dac.cpp
```diff
@@ -76,12 +76,6 @@ static void vga_dac_send_color(const uint8_t palette_idx, const uint8_t color_id
 {
 	const auto rgb666 = vga.dac.rgb[color_idx];
 
-	// We might be in the middle of a mode change, so we can't use
-	// VGA_GetCurrentVideoMode() here. That's because the INT 10H mode
-	// change BIOS routine needs to set up the Palette and Color Registers
-	// which will trigger this function.
-	const auto bios_mode_number = CurMode->mode;
-
```
johnnovak, line 84: I'm just moving this single line into `if (machine == MCH_VGA` as this is where the crash happens with `machine = hercules` only (because the Hercules palettes are initialised before the INT10H init stuff is called that initialised `CurMode`).

Not a great fix, but a _safe_ fix 😎 And it makes sense to restrict the `CurMode` check to the VGA case.

japsmits: Sounds reasonable.

Should we make an issue for of the underlying problem (initialisation spagetti)? Than at least that problem is documented somewhat.

weirddan455 (approved):
The fix looks good.  Yeah these globals that need initialization are really shit.  Ideally, we would wrap these things in a class and make them impossible to access before they've been initialized.

I'm not saying we need to object orient everything or get rid of all globals but we can't have code reading in invalid state and carrying on as if nothing happened.  We're not writing JavaScript here.

src/hardware/vga_dac.cpp
johnnovak: It's a known issue and I've been planning to create a refactoring issue ticket for this. Probably I'll do it after the release.

johnnovak:
> The fix looks good. Yeah these globals that need initialization are really shit. Ideally, we would wrap these things in a class and make them impossible to access before they've been initialized.
> 
> I'm not saying we need to object orient everything or get rid of all globals but we can't have code reading in invalid state and carrying on as if nothing happened. We're not writing JavaScript here.

Yeah, no disagreements. It will be a lot of work to get rid of global accesses, but we should at least prevent access across module boundaries.

PR #3387, opened 2024-02-09
Fix crash on startup when `machine` is set to `cga_mono`

johnnovak:
# Description

Okay, the `CurMode` iterator issue keeps giving... 😅 This is part 2 after the Hercules fix. I forgot to test `cga_mono` as well, mea culpa...

I'm doing a more extensive refactoring/cleanup/modularisation as part of this fix.

Recommend to review by commit.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/pull/3386

# Manual testing

### `machine = cga_mono`

Tested **Zak McKracken and the Alien Mindbenders (v1)** with the following base config:

```ini
[dosbox]
machine = cga_mono

[render]
monochrome_palette = green
```

Tested that:

- All valid values set from the config: `amber`, `green`, `white`, and `paperwhite`.
- An invalid value will default to `amber` and a warning gets logged.
- Changing `monochrome_palette` at runtime works as expected (as per the previous two tests).
- Changing the palette at runtime with the **Mono CGA Pal** hotkey works.

### `machine = hercules`

Repeated the same tests

### Other tests

Random regression testing with `svga_s3`, `tandy`, `ega`, `cga`, etc.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
[...]

PR #3388, opened 2024-02-09
Re-create CurMode iterator after backing array has been changed

weirddan455:
# Description

I think my #3374 PR was just bad after @johnnovak submitted 2 PRs fixing crashes caused by it.  I've reverted that and the previous workaround.

Rather than trying to pick a place (somewhat arbitrarily in `INT10_Init`) to initialize this thing, leave the global init  just re-create it when it gets invalidated.  This happens exactly once in the code in SVGA_S3 init.

The real problem here was the edge case when the iterator gets invalidated.  So just fix the edge case rather than trying to dive into this init mess so close to the 0.81 release.

## Related issues

#3374
#3386
#3387

# Manual testing

Tested with all machine types to confirm no more crashes.

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_dac.cpp
+++ b/src/hardware/vga_dac.cpp
@@ -76,6 +76,12 @@ static void vga_dac_send_color(const uint8_t palette_idx, const uint8_t color_id
 {
 	const auto rgb666 = vga.dac.rgb[color_idx];
 
+	// We might be in the middle of a mode change, so we can't use
+	// VGA_GetCurrentVideoMode() here. That's because the INT 10H mode
+	// change BIOS routine needs to set up the Palette and Color Registers
+	// which will trigger this function.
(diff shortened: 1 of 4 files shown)


johnnovak:
Thanks @weirddan455. I agree this is for the better and is the least invasive for now. I'm closing my second fix.

I think when we refactor the whole codebase to introduce proper modularisation and init order, that will be a better time to despaghettify all this global init stuff. Especially my last PR seemed a bit arbitrary too; it might be not necessary with better modularisation and init order.

Grounded0:
This would be the final PR before 0.81.0 next week.

johnnovak:
> This would be the final backported PR before 0.81.0 next week.

If we don't find something else, yeah.

johnnovak (approved):
We got there in the end @weirddan455 😅 I appreciate your persistence 👍🏻

Issue #1901, opened 2022-09-04
Sign the DOSBox Staging app for an easier, quicker and less scary first program start


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Mac OS

### Is your feature request related to a problem? Please describe.

Since the DOSBox Staging app isn't currently signed by an Apple Developer account, users that download the app have to jump to a few more hoops to get it running and might be scared by all the warnings that are popping up.

This will only get worse with newer versions of macOS.

This comparison is valid if the user already has allowed apps downloaded from the internet in security settings:
<img width="780" alt="Screenshot 2022-09-04 at 12 44 54" src="https://user-images.githubusercontent.com/61995128/188309689-f0a244ee-0b84-4afa-ab03-53bcbad48288.png">

DOSBox Staging
1. Double click app
2. Click 'OK'
3. Right click app
4. Click 'Open'
5. Click 'Open' again

![Open DOSBox Staging first start](https://user-images.githubusercontent.com/61995128/188309768-3a99c3a8-2de5-4518-91ad-98bb798c9e0f.gif)

Vanilla DOSBox
1. Double click app
2. Click 'Open'

![Open DOSBox vanilla first start](https://user-images.githubusercontent.com/61995128/188309786-7fe71d5b-7bea-4988-b35e-85c61a0746e3.gif)

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Grounded0:
Privilege to sign apps costs 99 USD per year from Apple. If you boost your yearly contribution to 100 USD per year we could earmark your contribution to pay for this fee. Personally i'm opposed that my donations would be used to bankroll 2 Trillion+ USD corporation to mint more money.

Burrito78:
Maybe someone with a developer account would be willing to sign it for us, @kcgen wanted to make a connection with one or two guys i know of.

Burrito78:
@dreamer My donation of 10$/mo would already pay for this, could you please consider signing Staging in your name?

dreamer:
@Burrito78 Yes. I need to learn how to do it anyway.

Burrito78:
#2426 is depending on this.

Burrito78:
If possible, would be nice to have for 0.81 @dreamer!

dreamer:
[...]

PR #3389, opened 2024-02-10
Create basic macOS notarization script

kklobe:
# Description

This is a rough script to take a macOS DMG artifact and repackage it after notarizing it with Apple. It is out-of-band, not intended to be run as part of the CI process since it requires a golden Developer ID signing certificate; it can be run on any of our DMGs after the fact.

# Related issues

#1901 

# Manual testing

I verified the UI elements are in the target disk image:

CleanShot 2024-02-09 at 17 52 29@2x

I uploaded the notarized DMG and re-downloaded it to trigger Gatekeeper, and ran the app to verify:

CleanShot 2024-02-09 at 17 51 52@2x

# Checklist

I have:

### DIFF
--- a/scripts/notarize-macos-dmg.sh
+++ b/scripts/notarize-macos-dmg.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Copyright (C) 2024-2024	The DOSBox Staging Team
+
+# This script notarizes a macOS DMG produced from the CI builds by
+# extracting the DMG, signing the app bundle, using the rest of the files
+# and the newly-signed app bundle to create a new DMG, then signing and 
+# notarizing the entire new DMG. Xcode must be installed.
+
+# Ensure developer identity and keychain profile name are set as environment variables
+# For example:
+# export DEVELOPER_IDENTITY="Developer ID Application: Valued Developer (xxxxxxxxxxx)"
+# export KEYCHAIN_PROFILE="Valued Developer Personal"
+#
+# Apple's notarization documentation: 
+# https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution
+
+if [[ -z "$DEVELOPER_IDENTITY" || -z "$KEYCHAIN_PROFILE" ]]; then
+    echo "Error: DEVELOPER_IDENTITY and KEYCHAIN_PROFILE environment variables must be set."
+    exit 1
+fi
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 INPUT_DMG OUTPUT_DMG"
+    exit 1
+fi
+
+SCRIPT_DIR=$(pwd)
+INPUT_DMG="$1"
+OUTPUT_DMG="$2"
+TEMP_DIR=$(mktemp -d)
+
+ATTACHED_VOLUME_PATH=$(hdiutil attach "$INPUT_DMG" | grep "Volumes" | awk 'BEGIN {FS="\t"}; {print $NF}')
+ATTACHED_VOLUME_NAME=$(basename "$ATTACHED_VOLUME_PATH")
+
+if [ -z "$ATTACHED_VOLUME_NAME" ]; then
+    echo "Error attaching DMG."
+    exit 1
+fi
+
+SRC_DIR="$TEMP_DIR/$ATTACHED_VOLUME_NAME"
+
+cp -a "$ATTACHED_VOLUME_PATH" "$TEMP_DIR"
+
+cd "$TEMP_DIR/$ATTACHED_VOLUME_NAME" || { echo "Failed to change to $TEMP_DIR/$ATTACHED_VOLUME_NAME"; exit 1; }
+
+codesign -s "$DEVELOPER_IDENTITY" -fv "DOSBox Staging.app" --entitlements "${SCRIPT_DIR}/contrib/macos/Entitlements.plist" --options runtime
+
+hdiutil create -volname "$ATTACHED_VOLUME_NAME" -srcfolder "$SRC_DIR" -ov -format UDZO "$OUTPUT_DMG"
+
+codesign -s "$DEVELOPER_IDENTITY" -fv "$OUTPUT_DMG"
+
+xcrun notarytool submit "$OUTPUT_DMG" --keychain-profile "$KEYCHAIN_PROFILE" --wait
+xcrun stapler staple "$OUTPUT_DMG"
+
+hdiutil detach "$ATTACHED_VOLUME_PATH"
+rm -rf "$TEMP_DIR"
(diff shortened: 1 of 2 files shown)


scripts/notarize-macos-dmg.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
johnnovak, line 1: `#!/usr/bin/env bash` would be more robust (but I keep forgetting this for my own scripts 😅)

johnnovak: Although, on macOS we can assume `/bin/bash`...

scripts/notarize-macos-dmg.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Copyright (C) 2024-2024	The DOSBox Staging Team
+
```
johnnovak, line 6: Maybe mention that the script must be run on macOS and no there are no extra prerequisites (all commands used by the script are included in the stock OS).

scripts/notarize-macos-dmg.sh
```diff
@@ -0,0 +1,58 @@
...
+# export DEVELOPER_IDENTITY="Developer ID Application: Valued Developer (xxxxxxxxxxx)"
+# export KEYCHAIN_PROFILE="Valued Developer Personal"
+
+if [[ -z "$DEVELOPER_IDENTITY" || -z "$KEYCHAIN_PROFILE" ]]; then
+    echo "Error: DEVELOPER_IDENTITY and KEYCHAIN_PROFILE environment variables must be set."
+    exit 1
+fi
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <input_dmg_path> <output_directory>"
```
johnnovak, line 23: How about: `Usage: $0 INPUT_DMG OUTPUT_DMG`

scripts/notarize-macos-dmg.sh
```diff
@@ -0,0 +1,58 @@
...
+    exit 1
+fi
+
+SCRIPT_DIR=$(pwd)
+INPUT_DMG="$1"
+OUTPUT_DIR="$2"
```
johnnovak, line 29: Can't we just do `OUTPUT_DMG="$2"` and get rid of the OUTPUT_DIR business? The output is just a single file.

scripts/notarize-macos-dmg.sh
```diff
@@ -0,0 +1,58 @@
...
+cd "$TEMP_DIR/$ATTACHED_VOLUME_NAME" || { echo "Failed to change to $TEMP_DIR/$ATTACHED_VOLUME_NAME"; exit 1; }
+
+codesign -s "$DEVELOPER_IDENTITY" -fv "DOSBox Staging.app" --entitlements "${SCRIPT_DIR}/contrib/macos/Entitlements.plist" --options runtime
+
+hdiutil create -volname "$ATTACHED_VOLUME_NAME" -srcfolder "$SRC_DIR" -ov -format UDZO "$OUTPUT_DMG"
+
+codesign -s "$DEVELOPER_IDENTITY" -fv "$OUTPUT_DMG"
+
+xcrun notarytool submit "$OUTPUT_DMG" --keychain-profile "$KEYCHAIN_PROFILE" --wait
+xcrun stapler staple "$OUTPUT_DMG"
```
johnnovak, line 55: That's right, staple it! 😂

scripts/notarize-macos-dmg.sh
kklobe: Right now it uses the input file name and adds -notarized to it, and the output just specify the target directory, rather than the entire output DMG path - but, could change it to do that instead.

scripts/notarize-macos-dmg.sh
kklobe: I'll mention in the description comment below that Xcode is required

johnnovak (approved):
Great stuff @kklobe 🎉 This will simplify the first launch for the macOS folk! Just a few minor non-essential improvement suggestions.

If we notarise the 0.81 final (and we should), I should also update the instructions on the website (downloads page & release notes), stating that the complicated authorisation dance is only necessary for releases before the 0.81 final.

Just typing this out to increase the chances that I'll remember this 😂

scripts/notarize-macos-dmg.sh
johnnovak: Yeah, I think INFILE OUTFILE is pretty standard.

Grounded0:
https://www.shellcheck.net

Grounded0:
Connected issues and PRs:

- #1901

Grounded0:
Pinging @Burrito78

johnnovak:
> https://www.shellcheck.net

It's part of the CI loop:
https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/verify-bash.sh#L7-L8

If shellcheck fails, it aborts the CI build (I discovered this a while ago when it complained about my new Bash script 😅).

scripts/notarize-macos-dmg.sh
```diff
@@ -48,6 +48,12 @@ cd "$TEMP_DIR/$ATTACHED_VOLUME_NAME" || { echo "Failed to change to $TEMP_DIR/$A
 
 codesign -s "$DEVELOPER_IDENTITY" -fv "DOSBox Staging.app" --entitlements "${SCRIPT_DIR}/contrib/macos/Entitlements.plist" --options runtime
 
+xattr -c "Start DOSBox Staging (logging).command"
+xattr -c "Start DOSBox Staging.command"
+
+codesign -s "$DEVELOPER_IDENTITY" -fv "Start DOSBox Staging (logging).command"
+codesign -s "$DEVELOPER_IDENTITY" -fv "Start DOSBox Staging.command"
```
johnnovak, line 55: Yeah, I don't think this does anything with `.command` files.

kklobe: It does add a signature, but still seems to be triggering Gatekeeper after a download.

Before signing the command files:
```log
codesign --verify -v /Volumes/DOSBox\ Staging/Start\ DOSBox\ Staging.command 
/Volumes/DOSBox Staging/Start DOSBox Staging.command: code object is not signed at all
```
after:
```log
codesign --verify -v /Volumes/DOSBox\ Staging/Start\ DOSBox\ Staging.command
/Volumes/DOSBox Staging/Start DOSBox Staging.command: valid on disk
/Volumes/DOSBox Staging/Start DOSBox Staging.command: satisfies its Designated Requirement
```

I'm guessing that it can't actually scan a script for the user to make sure it's safe, so people will still get that prompt.

johnnovak: Yeah that must be it and it makes sense.

PR #3393, opened 2024-02-11
Don't execute compile workflows on documentation and script changes

johnnovak:
# Description

It makes no sense to rebuild the binaries when _only_ changing documentation-related stuff in a PR or push. So with this change, I'm excluding those build workflows from firing on doco-only changes.

Additionally, I'm splitting the previous "analyse" workflow into "Clang analysis" and "Lint scripts & Markdown", and I'm running the script linter workflow as a prerequisite when deploy the website.

It would be nice if we didn't have to provide the whole list of patterns in each `paths-ignore` block, but unforunately that's not possible currently.

A few people have suggested how this could be improved, but no movement so far:

https://github.com/orgs/community/discussions/18608

https://github.com/orgs/community/discussions/70584

# Manual testing

1. Made a test website related commit to make sure only the `linters` workflow gets triggered.

2. Ran the `deploy-website` workflow off this branch and verified that it runs the `linters` workflow as a dependency:

image

# Checklist

I have:

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,47 @@
+name: Lint scripts & Markdown
+
+on: [push, pull_request, workflow_call]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  run_linters:
+    name: Run scripts & Markdown linters
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          submodules: false
+
+      - name: Update packages
+        run:  sudo apt-get update
+
+      - name: Run shellcheck
+        run:  ./scripts/verify-bash.sh
+
+      - name: Install pylint
+        run: |
+          sudo apt-get install python3-setuptools
+          sudo pip3 install pylint beautifulsoup4 html5lib
+
+      - name: Run pylint
+        run:  ./scripts/verify-python.sh
+
+      - name: Install markdownlint
+        run: |
+          sudo apt-get install ruby-full
+          ruby --version
+          sudo gem install mdl
+
+      - name: Run markdownlint
+        run:  ./scripts/verify-markdown.sh
+
+      - name: Install appstream-util
+        run:  sudo apt-get install appstream-util
+
+      - name: Verify metainfo.xml
+        run:  appstream-util validate-relax --nonet contrib/linux/dosbox-staging.metainfo.xml
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,6 +1,23 @@
 name: Linux builds
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - '.clang-format'
+      - '.mdl-styles'
+      - '*.md'
+      - 'docs/**'
+      - 'licenses/**'
+      - 'website/**'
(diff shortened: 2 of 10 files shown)


.github/workflows/linux.yml
```diff
@@ -1,6 +1,25 @@
...
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - '.clang-format'
+      - '.mdl-styles'
+      - '*.md'
+      - 'docs/**'
+      - 'licenses/**'
+      - 'website/**'
```
FeralChild64, line 11: contrib/resources/translations/** is IMHO, another good candidate.

johnnovak: Cheers, adding.

johnnovak: Ok, I think we must rebuild whenever any file in the bundled resources folder changes.

weirddan455:
You've rebased to the wrong branch.  I see my PCjr commit in there which hasn't been merged yet.

johnnovak:
> You've rebased to the wrong branch. I see my PCjr commit in there which hasn't been merged yet.

Cheers, dunno what happened there...

.github/workflows/linting.yml
```diff
@@ -0,0 +1,47 @@
...
+          sudo gem install mdl
+
+      - name: Run markdownlint
+        run:  ./scripts/verify-markdown.sh
+
+      - name: Install appstream-util
+        run:  sudo apt-get install appstream-util
+
+      - name: Verify metainfo.xml
+        run:  appstream-util validate-relax --nonet contrib/linux/dosbox-staging.metainfo.xml
```
shermp, line 47: Since you are moving this out to it's own workflow anyway, how about simplifying the workflow to install all required packages in a single step?

It's fine if you don't want to do that in this PR, but it was something I just noticed, so I'm mentioning it here.

.github/workflows/linux.yml
```diff
@@ -1,6 +1,25 @@
...
+      - 'scripts/**'
```
shermp, line 12: Are there some scripts here that get run in the workflow? Such as `scripts/create-package.sh` for example. It might be good to not ignore them if that's the case.

(same comment for the other workflows)

.github/workflows/linting.yml
shermp: Although the current setup does allow for "early abort" I suppose.

.github/workflows/linux.yml
johnnovak: Good point. I'll double-check that and/or remove the whole `scripts/**` entry perhaps because then maintaining per-filename exceptions per workflow could become quite fragile and error-prone.

But if it's just a few scripts, maybe I'll keep it anyway. I'll see.

.github/workflows/linting.yml
johnnovak: I slightly prefer the current approach because these steps also install Python and Ruby-specific stuff with `pip` and `gem`, respectively. If something fails, it's at least a bit easier to see where exactly the problem occurred.

Also, I don't think merging this all into a single step would add any meaningful speed benefits (this runs very quickly currently anyway). Leaving as is for now.

.github/workflows/linux.yml
johnnovak: I've just removed `scripts/**` from the ignore list. There's not direct support for ignoring the contents of a directory _except_ a few script files, and even if it was, maintaining that exclusion list is too error prone.

The scripts don't change too often, so I think this is fine. Not rebuilding on doco changes (which are much more frequent) is the main improvement here.

shermp (approved):

PR #3404, opened 2024-02-14
Fix handling multiple game controllers

bluddy:
# Description

With 2 joysticks (throttle + stick), I suddenly encountered a problem whereby the sticks were swapped in the mapper. Trying to map my joystick in the mapper would instead map my throttle and vice versa. After digging in, I found that  SDL2 has a level of indirection that wasn't accounted for in the code. With 1 joystick, you don't notice the difference. With 2 joysticks, the joysticks have a 50-50 chance of either being in the right order (responding and mapping correctly in the keymapper) or being swapped. Accounting for the indirection fixes it.

As an aside, what even is this code? One cpp file? I feel an urgent need to refactor.

# Manual testing

Ran the mapper again and it works correctly. This is on Windows 11.

# Checklist

I have:

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -593,10 +593,9 @@ static void set_joystick_led([[maybe_unused]] SDL_Joystick *joystick,
 
 class CStickBindGroup : public CBindGroup {
 public:
-	CStickBindGroup(int _stick, uint8_t _emustick, bool _dummy = false)
+	CStickBindGroup(int _stick_index, uint8_t _emustick, bool _dummy = false)
 	        : CBindGroup(),
-	          stick(_stick),       // the number of the physical device (SDL
-	                               // numbering)
+	          stick_index(_stick_index), // the number of the device in the system
 	          emustick(_emustick), // the number of the emulated device
 	          is_dummy(_dummy)
 	{
@@ -616,7 +615,21 @@ public:
 		emulated_hats=0;
 		JOYSTICK_Enable(emustick,true);
 
-		sdl_joystick=SDL_JoystickOpen(_stick);
+		// From the SDL doco
+		// (https://wiki.libsdl.org/SDL2/SDL_JoystickOpen):
+
+		// The device_index argument refers to the N'th joystick presently
+		// recognized by SDL on the system. It is NOT the same as the
+		// instance ID used to identify the joystick in future events.
+
+		// Also see: https://wiki.libsdl.org/SDL2/SDL_JoystickInstanceID
+
+		// We refer to the device index as `stick_index`, and to the
+		// instance ID as `stick_id`.
+
+		sdl_joystick = SDL_JoystickOpen(stick_index);
+		stick_id = SDL_JoystickInstanceID(sdl_joystick);
+
 		set_joystick_led(sdl_joystick, on_color);
 		if (sdl_joystick==nullptr) {
 			button_wrap=emulated_buttons;
@@ -645,7 +658,7 @@ public:
 			button_wrap = MAXBUTTON;
 
 		LOG_MSG("MAPPER: Initialised %s with %d axes, %d buttons, and %d hat(s)",
-		        SDL_JoystickNameForIndex(stick), axes, buttons, hats);
+		        SDL_JoystickNameForIndex(stick_index), axes, buttons, hats);
 	}
 
 	~CStickBindGroup() override
@@ -696,7 +709,7 @@ public:
 			const int axis_id = event->jaxis.axis;
 			const auto axis_position = event->jaxis.value;
 
-			if (event->jaxis.which != stick)
+			if (event->jaxis.which != stick_id)
 				return nullptr;
 #if defined(REDUCE_JOYSTICK_POLLING)
 			if (axis_id >= axes)
@@ -711,7 +724,7 @@ public:
 			return CreateAxisBind(axis_id, toggled);
 
 		} else if (event->type == SDL_JOYBUTTONDOWN) {
-			if (event->jbutton.which != stick)
+			if (event->jbutton.which != stick_id)
 				return nullptr;
 #if defined (REDUCE_JOYSTICK_POLLING)
 			return CreateButtonBind(event->jbutton.button%button_wrap);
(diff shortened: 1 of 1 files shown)


src/gui/sdl_mapper.cpp
```diff
@@ -595,8 +595,7 @@ class CStickBindGroup : public CBindGroup {
 public:
 	CStickBindGroup(int _stick, uint8_t _emustick, bool _dummy = false)
 	        : CBindGroup(),
-	          stick(_stick),       // the number of the physical device (SDL
-	                               // numbering)
+	          stick_order(_stick), // the number of the device in the system
```
johnnovak, line 598: Let's call it `stick_index` as the SDL documentation.

src/gui/sdl_mapper.cpp
```diff
@@ -616,7 +615,14 @@ class CStickBindGroup : public CBindGroup {
 		emulated_hats=0;
 		JOYSTICK_Enable(emustick,true);
 
+		// NOTE: the _stick number passed in here is *one* index SDL uses to
+		// find the joystick, but it's not the one that will appear later in events.
+		// (see https://wiki.libsdl.org/SDL2/SDL_JoystickInstanceID)
```
johnnovak, line 620: Nice comments but can you be clearer in your wording and use SDL terminology consistently? Let's not add our terms as well to the mix 😄 

We should only talk about "joystick index (Nth joystick in the system" and "SDL joystick instance ID".

src/gui/sdl_mapper.cpp
```diff
@@ -616,7 +615,14 @@ class CStickBindGroup : public CBindGroup {
...
 		sdl_joystick=SDL_JoystickOpen(_stick);
+
+		// Get actual index SDL will use in the future to refer to this joystick in events
```
johnnovak, line 623: This is not the "index". This is the "instance ID".

src/gui/sdl_mapper.cpp
johnnovak: Also, I suggest refactoring `_stick` to `_stick_index`.

src/gui/sdl_mapper.cpp
```diff
@@ -919,7 +925,10 @@ class CStickBindGroup : public CBindGroup {
 	uint8_t emulated_buttons = 0;
 	int hats = 0;
 	int emulated_hats = 0;
-	int stick;
+	int stick{-1};    // This is the real index of the joystick as appears in
+	                  // SDL events
```
johnnovak, line 929: "This is the SDL instance ID..."

johnnovak:
Thanks @bluddy, very nice improvement and I'd say backport worthy! 🎉 

We'll need to test it a bit, I have I think 4 controllers on my system 😄 

> As an aside, what even is this code? One cpp file? I feel an urgent need to refactor.

It's quite something, isn't it? 😅 We don't want to touch it too much, it's too easy to break it and we'll nuke it anyway and rewrite the mapper 100% from scratch.

johnnovak:
Two more minor things @bluddy:

- We don't do prefixes like this in commit messages: `sdl_mapper: rename and clean up comments`
- Can you go through the checklist again please and address what's missing (hint: assignee, labels)

Thanks! 😄

src/gui/sdl_mapper.cpp
```diff
@@ -925,10 +925,8 @@ class CStickBindGroup : public CBindGroup {
 	uint8_t emulated_buttons = 0;
 	int hats = 0;
 	int emulated_hats = 0;
-	int stick{-1};    // This is the real index of the joystick as appears in
-	                  // SDL events
-	int stick_index{-1}; // This is a different number -- the number of the
-	                     // joystick in the system
+	int stick_id{-1};    // Instance ID of the joystick as it appears in SDL events
+	int stick_index{-1}; // Index of the joystick in the system
```
johnnovak, line 929: Very nice, now it's crystal clear. Thanks for doing this 🏅

bluddy:
> Two more minor things @bluddy:
> 
> - We don't do prefixes like this in commit messages: `sdl_mapper: rename and clean up comments`
> - Can you go through the checklist again please and address what's missing (hint: assignee, labels)
> 
> Thanks! 😄 

I did try to add assignees and labels but it wouldn't let me. I'm nearing bedtime so I'll try again tomorrow.

johnnovak:
> > Two more minor things @bluddy:
> > 
> > * We don't do prefixes like this in commit messages: `sdl_mapper: rename and clean up comments`
> > * Can you go through the checklist again please and address what's missing (hint: assignee, labels)
> > 
> > Thanks! 😄
> 
> I did try to add assignees and labels but it wouldn't let me. I'm nearing bedtime so I'll try again tomorrow.

Ah, probably you don't have the rights then. I've done it, no prob. Thanks for trying 😄

src/gui/sdl_mapper.cpp
johnnovak: Honestly, I think it would be best to simply quote the SDL doco here. I find their wording clearer as you're still talking about "two different indexes", where in reality there is no such thing: there is the _one_ index, the device index, which is used for opening the device, then after that there's the instance ID in SDL events. That's all there is to it.

This device index as seen by the OS and instance ID in events distinction is very clear; talking about two different indexes muddies the waters and makes this appear more complicated than it really is.

So I suggest to just replace your comment with this:

```
From the SDL doco (https://wiki.libsdl.org/SDL2/SDL_JoystickOpen): 

  The device_index argument refers to the N'th joystick presently recognized
  by SDL on the system. It is NOT the same as the instance ID used to identify
  the joystick in future events.

Also see: https://wiki.libsdl.org/SDL2/SDL_JoystickInstanceID
```

johnnovak:
Looks good @bluddy , just some more nitpicking about naming things and the comments 😄 

Plus please fix up the commit messages and get rid of that merge commit by rebasing your history. We don't want to see any merge commits in main.

After that we're good to go. Thanks!

src/gui/sdl_mapper.cpp
bluddy: Done.

src/gui/sdl_mapper.cpp
```diff
@@ -919,7 +930,8 @@ class CStickBindGroup : public CBindGroup {
 	uint8_t emulated_buttons = 0;
 	int hats = 0;
 	int emulated_hats = 0;
-	int stick;
+	int stick_id{-1};    // Instance ID of the joystick as it appears in SDL events
+	int stick_index{-1}; // Index of the joystick in the system
```
johnnovak, line 934: Please put the comments on the previous line. I really dislike long inline comments, and kcgen is the same. We eradicate existing ones on-sight, so please don't add new ones 😄

bluddy:
OK I think I covered everything.

bluddy:
No prob. I appreciate picky comments -- they show adherence to excellence.
The only issue I have is that every project has its own standards, and they're just so different it's virtually impossible to work on multiple projects while remembering all the nuances. At work we have to adhere to the AUTOSAR c++ standards, which are insane in their own way, and then we have our own conventions, which are also slightly different for 2 of our projects due to historical reasons.

johnnovak:
> The only issue I have is that every project has its own standards, and they're just so different it's virtually impossible to work on multiple projects while remembering all the nuances.

I can certainly symphatise 😅 Looking forward to your future code contributions -- it only gets easier from here 🚀

PR #3413, opened 2024-02-17
Use uncompressed CPI files instead of compressed CPX ones

FeralChild64:
# Description

This PR replaces the compressed CPX code page files from the FreeDOS project with uncompressed ones, generated with `upx -d ORIGINAL.CPX` command. Rationale:

1. In order to use the CPX file, DOSBox executes the decompression routines within it's virtual environment - which can be slow if number of cycles is set to a low value
2. The code to handle decompression used to have issues in the past, and is definitely still buggy (that's why EGA18.CPI is already provided in an uncompressed form) - IMHO we should consider purging it in the future, as there is probably no good use case for it, the only CPX files I know come from the FreeDOS projects, and the last time they were updated was well over a decade ago, see https://github.com/FDOS/cpi
3. Currently the _code page -> CPX file_ mapping is hardcoded, which is not too elegant. If the supplied files are uncompressed, we could read them all at startup and build the list of supported code pages dynamically. This would be handy if we decide to add code pages from CPIWIN, CPIKOI, or CPIISO directories.

# Manual testing

1. Start `dosbox --noprimaryconf`
3. Change number of cycles to something low, i. e. execute `cycles 500` command
4. Change the keyboard layout, i.e. `keyb pl`, `keyb ua` - it should work quickly now

# Checklist

I have:


johnnovak (approved):

johnnovak:
Looks straightforward enough @FeralChild64.

PR #3414, opened 2024-02-17
Move IPX server to a background thread

kklobe:
# Description

The IPX server is self-contained, so let's move it off the main thread's TIMER tick loop and put it on its own thread.

# Manual testing

Ran IPX network games of Quake and Rise of the Triad, using both macOS and Windows as servers.

Started and stopped the server several times on both platforms and observed the thread properly create and destroy by observing thread count in Activity Monitor/Task Manager

# Checklist

I have:

### DIFF
--- a/src/hardware/ipxserver.cpp
+++ b/src/hardware/ipxserver.cpp
@@ -31,6 +31,7 @@ static constexpr int UDP_UNICAST = -1; // SDLNet magic number
 
 static IPaddress ipxServerIp;     // IPAddress for server's listening port
 static UDPsocket ipxServerSocket; // Listening server socket
+static SDLNet_SocketSet socket_set = nullptr;
 
 static packetBuffer connBuffer[SOCKETTABLESIZE];
 
@@ -210,7 +211,9 @@ void IPX_StopServer() {
 		ipx_server_thread.join();
 	}
 
+	SDLNet_FreeSocketSet(socket_set);
 	SDLNet_UDP_Close(ipxServerSocket);
+	socket_set = nullptr;
 }
 
 bool IPX_StartServer(uint16_t portnum)
@@ -224,14 +227,35 @@ bool IPX_StartServer(uint16_t portnum)
 			i.connected = false;
 		}
 
-		ipx_server_running = true;
-		ipx_server_thread  = std::thread([]() {
-                        while (ipx_server_running) {
-                                IPX_ServerLoop();
-                                std::this_thread::sleep_for(
-                                        std::chrono::milliseconds(1));
-                        }
-                });
+		if (!socket_set) {
+			socket_set = SDLNet_AllocSocketSet(1);
+			if (!socket_set) {
+				LOG_ERR("IPXServer: %s", SDLNet_GetError());
+				return false;
+			}
+			if (SDLNet_UDP_AddSocket(socket_set, ipxServerSocket) == -1) {
+				LOG_ERR("IPXServer: %s", SDLNet_GetError());
+				return false;
+			}
+		}
+
+		if (!ipx_server_running) {
+			ipx_server_running = true;
+			ipx_server_thread  = std::thread([]() {
+                                while (ipx_server_running) {
+                                        const int num_ready = SDLNet_CheckSockets(
+                                                socket_set, 100);
+                                        if (num_ready == -1) {
+                                                LOG_ERR("IPXServer: %s",
+                                                        SDLNet_GetError());
+                                                continue;
+                                        }
+                                        if (num_ready > 0) {
+                                                IPX_ServerLoop();
+                                        }
+                                }
+                        });
+		}
 		return true;
 	}
 	return false;
--- a/src/hardware/ipxserver.cpp
+++ b/src/hardware/ipxserver.cpp
@@ -20,26 +20,28 @@
 
 #if C_IPX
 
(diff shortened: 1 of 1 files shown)


src/hardware/ipxserver.cpp
```diff
@@ -212,10 +220,18 @@ bool IPX_StartServer(uint16_t portnum)
...
+			i.connected = false;
+		}
 
-		TIMER_AddTickHandler(&IPX_ServerLoop);
+		ipx_server_running = true;
+		ipx_server_thread  = std::thread([]() {
+                        while (ipx_server_running) {
+                                IPX_ServerLoop();
+                                std::this_thread::sleep_for(
+                                        std::chrono::milliseconds(1));
```
weirddan455, line 232: This is usually a red flag when implementing threading.  Worker threads usually want to be waiting on an event rather than sleeping for an arbitrary amount of time.

I had a look `IPX_ServerLoop` and I think it would be best to use `SDLNet_CheckSockets` to implement blocking.  That way you can use a longer time-out but also be woken up right away when data arrives on the socket.

src/hardware/ipxserver.cpp
```diff
@@ -212,10 +220,18 @@ bool IPX_StartServer(uint16_t portnum)
...
 
-		for (uint16_t i = 0; i < SOCKETTABLESIZE; ++i)
-			connBuffer[i].connected = false;
+		for (auto& i : connBuffer) {
+			i.connected = false;
+		}
 
-		TIMER_AddTickHandler(&IPX_ServerLoop);
+		ipx_server_running = true;
+		ipx_server_thread  = std::thread([]() {
```
weirddan455, line 228: This will throw an exception if a caller calls `IPX_StartServer` twice as the thread will be destroyed without having been joined first.

Often I will keep a worker thread running for the duration of the entire application waiting on a start event to avoid the overhead of repeatedly creating/destroying threads.  If `IPX_StartServer` gets called infrequently (I'm guessing it does?), it's fine to spin up a thread on demand but you might want to explicitly call `IPX_StopServer` beforehand.

weirddan455 (commented):
Couple of comments but this is a great idea.  I like to see more stuff moved off the main thread, especially when it's quick and easy like this :+1:

src/hardware/ipxserver.cpp
kklobe: I like this much better, was not aware of that function.

src/hardware/ipxserver.cpp
kklobe: good point - it's pretty safe, since the user explicitly has to do `ipxnet startserver`, so I don't want a thread running if the user doesn't want it, however it's easy to put a quick check for `!ipx_server_running` before spinning it up, just in case.

weirddan455 (approved):
Looks good now.

src/hardware/ipxserver.cpp
```diff
@@ -212,10 +223,39 @@ bool IPX_StartServer(uint16_t portnum)
...
+		}
+
+		if (!socket_set) {
+			socket_set = SDLNet_AllocSocketSet(1);
+			if (!socket_set) {
+				LOG_ERR("IPXServer: %s", SDLNet_GetError());
+				return false;
+			}
+			if (SDLNet_UDP_AddSocket(socket_set, ipxServerSocket) == -1) {
+				LOG_ERR("IPXServer: %s", SDLNet_GetError());
```
johnnovak, line 237: @kklobe minor: We do ALL_UPPER_CASE log prefixes, so `IPXSERVER` would be more consistent.

kklobe: ah, shikes, I don't know how that got in there. Worth a quick add-on PR?

src/hardware/ipxserver.cpp
```diff
@@ -20,26 +20,29 @@
...
 #include "timer.h"
-#include <cstdlib>
-#include <cstring>
 
-constexpr int UDP_UNICAST = -1; // SDLNet magic number
+static constexpr int UDP_UNICAST = -1; // SDLNet magic number
+
+static IPaddress ipxServerIp;     // IPAddress for server's listening port
+static UDPsocket ipxServerSocket; // Listening server socket
+static SDLNet_SocketSet socket_set = nullptr;
```
johnnovak, line 34: @kklobe minor: We try to eradicate inline comments on sight 😎

kklobe: hah! will try to remember that from now on.

src/hardware/ipxserver.cpp
johnnovak: Up to you; if you feel like it, sure why not 😄 Or the next person touching this file can do a cleanup sweep.

I was mainly commenting on these minor issues as an "information dissemination" exercise. 99% of the work here was making it threaded and that's great—I'm just reflecting on the minor superficial stuff 😄 

Some other common things kcgen and I use to correct on sight:

```cpp
all_vars_lower_snake_case (basically any var definition anywhere)
enum class MyEnum { SomeEnumValue, AndAnotherOne }
static void i_am_a_private_function
struct UpperCamel { ... } (no weird _t postfix)
```

kklobe: ok, I'm going to create a `.clang-tidy` that flags these

Issue #3146, opened 2023-11-23
Do not always recreate the window when changing `[sdl]` section settings


johnnovak:
Currently, the window gets recreated and resized when changing pretty much any setting in the `[sdl]` section. E.g., do the following:

1. Start Staging and resize the window so it's a tall rectangle (just to make it obvious when it changes its size automatically).
2. Execute `window_position = 0,0`
3. Note the window got resized to its initial size.
4. Resize the window again to a tall rectangle.
5. Execute `pause_when_inactive = off`.
6. The window size got reset again.

If you do the same in fullscreen, you'll get delays and/or screen flashes instead.

While recreating the window might be necessary in some cases (e.g., when changing the `vsync` setting or the `output` mode), we should aim to preserve its size and position when this is practical.

Then for some settings, such as `pause_when_inactive`, `mute_when_inactive`, `waitonerror`, `priority`, and probably a few others, we should absolutely should leave the window completely alone and neither change its position or size nor recreate it.

Burrito78:
This will probably close https://github.com/dosbox-staging/dosbox-staging/issues/2549

weirddan455:
I have looked at this before.  There have been similar issues raised #2527 #2549 #3075 #3080 that all boil down to limitations with our config system.

Currently, whenever any setting gets changed at runtime, it calls that section's `Destroy()` + `Init()` functions.  This leads to doing too much work (SDL re-initing everything on some simple change), bad coupling (SDL's re-init function is coupled to SDL Mapper), and sometimes just not fitting well with architectural decisions we've made elsewhere (Fluidsynth and MT32 being "children" of the general MIDI module).

One solution could be to register a function pointer with every setting (could be the same function for multiple settings if that makes sense).  If no function is registered, that means the setting is not changeable at runtime and we display an error to the user.  This lets us be more flexible instead of just destroy + init everything.

However, this something I am hesitant to do until after 0.81 because of the pending feature freeze.  It would be a large refactor with high risk to cause regressions or introduce additional bugs.

johnnovak:
Definitely after 0.81.0 @weirddan455, we don't have capacity for this now. Thanks for looking into it and thinking about it, though.

We'll need to discuss this in detail and come up with a plan first if you're contemplating a larger scale refactor, not just addressing the symptoms described in this ticket. There are several moving parts here:

- @kcgen 's module rework he started is related to this
- We are contemplating changing the whole config system too
[...]

PR #3417, opened 2024-02-18
Refactor some init crap

### LINKED ISSUE #3411: timed=true messes up conventional memory

_No response_

Windows 10

_No response_

### Relevant hardware info

Intel i3 9100, 32GB RAM

entering the command timed=true uses 35840 (!) KB of conventional memory, leaving 0 KB free.

Explain how to reproduce
1. Run DOSBox-Staging 0.81
2. timed true
3. mem

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

I refactored a bunch of stuff.  Better description later.  I'm tired and there's still more work on I need to do on this PR.

## Related issues

Fixes #3411

The problem was `BIOS_Destroy` got marked `changeable_at_runtime` but `BIOS_Init` was not.  This means when a config setting changed under the `joystick` section, we deleted the BIOS object and just never created a new one.  The goal of this refactor was to make this type of bug impossible.

# Manual testing

Monkey Island 2 starts.  More testing needed.  I probably broke some stuff.  The window kind of flickers once on start-up.

# Checklist

I have:

### DIFF
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -522,6 +522,9 @@ struct CPUBlock {
 
 extern CPUBlock cpu;
 
+void CPU_Init(Section* sec);
+void CPU_ShutDown(Section* sec);
+
 void CPU_SetFlags(const uint32_t word, uint32_t mask);
 void CPU_SetFlagsd(const uint32_t word);
 void CPU_SetFlagsw(const uint32_t word);
--- a/include/setup.h
+++ b/include/setup.h
@@ -304,23 +304,16 @@ public:
 
 typedef void (*SectionFunction)(Section*);
 
+struct SectionFunctions {
+	SectionFunction init = nullptr;
+	SectionFunction destroy = nullptr;
+	bool execute_on_config_change = false;
+};
+
 class Section {
 private:
-	// Wrapper class around startup and shutdown functions. the variable
-	// changeable_at_runtime indicates it can be called on configuration
-	// changes
-	struct Function_wrapper {
-		SectionFunction function;
-		bool changeable_at_runtime;
 
-		Function_wrapper(const SectionFunction fn, bool ch)
-		        : function(fn),
-		          changeable_at_runtime(ch)
-		{}
-	};
-
-	std::deque<Function_wrapper> init_functions   = {};
-	std::deque<Function_wrapper> destroyfunctions = {};
+	std::vector<SectionFunctions> section_functions = {};
 	std::string sectionname                       = {};
 
 public:
@@ -334,11 +327,7 @@ public:
 	// Children must call executedestroy!
 	virtual ~Section() = default;
 
-	void AddInitFunction(SectionFunction func, bool changeable_at_runtime = false);
-
-	void AddDestroyFunction(SectionFunction func,
-	                        bool changeable_at_runtime = false);
-
+	void AddFunctions(const SectionFunctions *funcs);
 	void ExecuteInit(bool initall = true);
 	void ExecuteDestroy(bool destroyall = true);
 
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -2187,8 +2187,6 @@ OplMode find_oplmode()
 	return opl_mode;
 }
 
-void SBLASTER_ShutDown(Section*);
-
 class SBLASTER final {
 private:
 	/* Data */
@@ -2463,8 +2461,5 @@ void SBLASTER_Init(Section* sec)
 {
 	assert(sec);
 
-	sblaster = std::make_unique<SBLASTER>(sec);
-
-	constexpr auto changeable_at_runtime = true;
-	sec->AddDestroyFunction(&SBLASTER_ShutDown, changeable_at_runtime);
+	sblaster = std::make_unique<SBLASTER>(sec);;
 }
--- a/src/misc/setup.cpp
+++ b/src/misc/setup.cpp
@@ -1128,14 +1128,13 @@ bool Config::WriteConfig(const std_fs::path& path) const
 	return true;
 }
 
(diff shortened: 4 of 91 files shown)


johnnovak:
Won't review this now, but this most certainly cannot go in as a single megacommit.

This needs to be broken up as much as possible. Ideally, one commit per module, so it's easily bisectable.

include/cpu.h
```diff
@@ -522,6 +522,9 @@ struct CPUBlock {
 
 extern CPUBlock cpu;
 
+void CPU_Init(Section* sec);
+void CPU_ShutDown(Section* sec);
```
FeralChild64, line 526: IMHO either `MODULE_ShutDown`, or `MODULE_Destroy` - not both.

FeralChild64:
Right, it's a bit too large. But it's a huge step in a proper direction.

include/cpu.h
johnnovak: Yeah, I don't like Destroy... Should be Init and Shutdown, IMO; those are standard terms. And the word is "shutdown", so no weird ShutDown capitalisation.

johnnovak:
> Right, it's a bit too large. But it's a huge step in a proper direction.

It is, but we need granular bisectability.

src/hardware/sblaster.cpp
```diff
@@ -2463,8 +2461,5 @@ void SBLASTER_Init(Section* sec)
 {
 	assert(sec);
 
-	sblaster = std::make_unique<SBLASTER>(sec);
-
-	constexpr auto changeable_at_runtime = true;
-	sec->AddDestroyFunction(&SBLASTER_ShutDown, changeable_at_runtime);
+	sblaster = std::make_unique<SBLASTER>(sec);;
```
japsmits, line 2464:
```suggestion
	sblaster = std::make_unique<SBLASTER>(sec);
```

src/misc/setup.cpp
```diff
@@ -1209,61 +1208,45 @@ void Config::Init() const
...
-			// If the above function call appended items then we
-			// need to avoid calling them immediately in this
-			// current pass by advancing our index across them. The
-			// setup class will call the added function itself.
-			//
-			const auto num_appended = size_on_exit - size_on_entry;
-			i += num_appended;
-			assert(i < init_functions.size());
-		}
+		const auto size_on_entry = section_functions.size();
```
japsmits, line 1230: Add a [[maybe_unused]], since these variables are unused if not compiled in DEBUG-mode.

japsmits (commented):
W.O.W.!!! 😄

include/cpu.h
weirddan455: Yeah I've left the function names as they were but they should be standardized across the codebase.

include/setup.h
```diff
@@ -304,23 +304,16 @@ class Prop_hex final : public Property {
 
 typedef void (*SectionFunction)(Section*);
 
+struct SectionFunctions {
+	SectionFunction init = nullptr;
+	SectionFunction destroy = nullptr;
+	bool execute_on_config_change = false;
```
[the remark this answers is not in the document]
weirddan455: It's not an enum.  It's using all of these fields.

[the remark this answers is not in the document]
japsmits: I see GitHub is a bit confusing here, since I only selected line 310 for comment, not the whole struct. I meant just replacing the "bool"-type with an enum-type. 

Point is that if I see something like "{DOS_init, DOS_destroy, true}", I have no clue what "true" means until I look at the definition of the struct. If I see for example "{DOS_init, DOS_destroy, OnConfigChange::execute}", I have a clue what the intention is. And we have the opportunity to easily add more options if needed.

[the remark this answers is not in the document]
johnnovak: @japsmits That's a good suggestion—for the future.

The config handling will be extensively refactored/rewritten. Until then, we only fix critical bugs in it, otherwise we're just spinning our wheels making small incremental improvements to something that will be switched out eventually anyway.

Until then, we use this "named flag" approach to avoid in passing in nondescript "true" and "false" literals.

But yeah, ultimately, we'll go with the approach you described in the rewrite. Enums all the way 😎

weirddan455:
>  Also, you should study this abandoned PR of kcgen before embarking on such refactors: https://github.com/dosbox-staging/dosbox-staging/pull/2746

My intention was not to go that deep.  I may just close this.  I was trying to fix one small part of the init/config system (ensuring that both init and destroy are using the same "changeable" setting).

I'm not actually sure this is the correct approach though now that I look back.  I found this comment I made a few months back describing some other problems with the current system I had forgotten about.  https://github.com/dosbox-staging/dosbox-staging/issues/3146#issuecomment-1832096509

We probably need to make some architectural decisions before we do any large scale refactors.  I believe @johnnovak mentioned formulating a plan of some sort for how to move forward.

johnnovak:
> I'm not actually sure this is the correct approach though now that I look back. I found this comment I made a few months back describing some other problems with the current system I had forgotten about. [#3146 (comment)](https://github.com/dosbox-staging/dosbox-staging/issues/3146#issuecomment-1832096509)

Yeah, that comment is a good summary of some of the issues we're facing.
 
> We probably need to make some architectural decisions before we do any large scale refactors. I believe @johnnovak mentioned formulating a plan of some sort for how to move forward.

Yes, I'm planning to create a discussion ticket with my ideas in March, after the patch release is out, then we can collectively decide on the best path forward. There are quite a few considerations, and we want to get the new design reasonably correct this time, so I don't want to rush anything.

weirddan455:
half-measures

johnnovak:
> half-measures

My favourite series ❤️

Issue #461, opened 2020-06-23
Support MMX emulation


kcgen:
Late-stage MSDOS games required or used MMX registers to accelerate their math operations.

Kekko (author of the 3dfx patch), has recently [shared a patch on Vogons](https://www.vogons.org/viewtopic.php?t=74877) adding MMX support:

> It adds Pentium MMX emulation to Dosbox.
> The instructions are for the interpreter; the recompiler can use it too, but it has to fall back to the interpreter core.
> It finally allows to run games which strictly require this instruction set, like Serious Sam (combined with the 3dfx patch).
> Probably the first AAA title from 21st century ever run in Dosbox :-)

The post includes the following attachments.

![ss1](https://user-images.githubusercontent.com/1557255/85446424-47cbcc80-b549-11ea-89d8-06bc93a445a8.png)
![ss2](https://user-images.githubusercontent.com/1557255/85446428-48646300-b549-11ea-886c-88da8cbc0359.png)
![ss3](https://user-images.githubusercontent.com/1557255/85446430-48fcf980-b549-11ea-884f-e9fca94b877d.png)
![ss4](https://user-images.githubusercontent.com/1557255/85446432-48fcf980-b549-11ea-83db-d9f17e9172d8.png)
![ss5](https://user-images.githubusercontent.com/1557255/85446436-49959000-b549-11ea-9280-3c0fb2c47c2c.png)
![ss6](https://user-images.githubusercontent.com/1557255/85446438-49959000-b549-11ea-87c7-2af298648eb1.png)

**The patch:** [:floppy_disk: mmx_20200623.diff.gz](https://github.com/dosbox-staging/dosbox-staging/files/4821368/mmx_20200623.diff.gz) 5.8KB.

Per Kekko's past request regarding the 3dfx patch, we will similarly provide any further improvements via the Vogon's thread linked above.

dreamer:
AFAIK Serious Sam is not a DOS game, so I guess it depends on Windows 95 or later running inside dosbox? Personally I am not interested in running games from this era via dosbox-staging - Wine works sooo much better for this purpose (the game works perfectly in Proton); also, there is Linux-native port in the form of Serious Sam Fusion.

If anyone plans to work on inclusion of this patch (I don't) - I would prefer if we avoided new build-time configurations.

kcgen:
[...]

PR #3426, opened 2024-02-20
Integrate MMX emulation

FeralChild64:
# Description

Integrate DOSBox MMX emulation patch by Kekko, shared on Vogons forum, https://www.vogons.org/viewtopic.php?t=74877. The PR contains additional fixes and merges from DOSBox-X done by @kklobe.

To enable:
- in the config file set `cputype = pentium_mmx`

example

It's possible that presence of the MMX instruction set will improve performance of some Windows 9x games or multimedia software too - SiSoft Sandra 2001 (https://archive.org/details/san710_zip) under Windows 98, with Pentium processor emulated:

Pentium

.. and with Pentium MMX emulated:

Pentium_MMX

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/461

# Manual testing

- run _Extreme Assault_ game, test with _MMX Technology_ enabled/disabled in the _Graphic Detail_ menu
- run _MMXtest_ tool from https://calsci.com/techSupport.html (requires Windows 9x) - should report MMX tests passed and report a speedup - speedup values are not stable, but they always show performance improvement, at least on my system
- get _Seven Heaven_ demo by _Exceed_ from https://files.scene.org/view/parties/2000/mekkasymposium00/in64/h7-final.zip - should run correctly on emulated Pentium MMX and refuse to run on non-MMX CPUs
- get the _Impulse Tracker_ and try MMX driver, for example launch it with `IT.EXE -Sitsb16.mmx` command, open a module, start playing with F5 function key

# Checklist

I have:


kklobe:
I would also like to checkout this library, it seems very well supported and has active development. Looks like you can run any SIMD instruction set, including MMX, on most any popular architecture, so MMX on ARM NEON. 

https://github.com/simd-everywhere/simde

kklobe:
ok @FeralChild64 I think I have the dyn_x86 crash fixed, checkout https://github.com/dosbox-staging/dosbox-staging/tree/kk/fc-mmx-1

FeralChild64:
@kklobe That's great, thanks for the fix! I can confirm Extreme Assault is now fixed :)

src/cpu/core_dyn_x86/dyn_mmx.h
```diff
@@ -0,0 +1,63 @@
...
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#include "mmx.h"
+
+extern uint32_t * lookupRMEAregd[256];
+
+#define LoadMd(off) mem_readd_inline(off)
+#define LoadMq(off) ((uint64_t)((uint64_t)mem_readd_inline(off+4)<<32 | (uint64_t)mem_readd_inline(off)))
```
kklobe, line 25: I added quadword reads and writes awhile back, so this and the write below can become e.g.

```cpp
mem_readq_inline(off)
```

kklobe:
Tested (Extreme Assault) on macOS M3, works great.

CleanShot 2024-02-21 at 14 53 32@2x

kklobe:
I've tested ARM64 via macOS, and 32-bit/64-bit dyn-x86 via Windows, using Extreme Assault, no issues so far.

The Heaven Seven demo isn't working right on dyn-x86, but looks great on dynrec (my M3), so I will start looking into that.

src/cpu/core_dyn_x86/dyn_mmx.h
FeralChild64: Done - in a separate commit, for bisectability.

FeralChild64:
I wonder what can we do about:

`../include/fpu.h:72:24: warning: ISO C++ prohibits anonymous structs [-Wpedantic]`

compiler warning. I would like to keep the existing FPU code intact, if possible.

kklobe:
> I wonder what can we do about:
> 
> `../include/fpu.h:72:24: warning: ISO C++ prohibits anonymous structs [-Wpedantic]`
> 
> compiler warning. I would like to keep the existing FPU code intact, if possible.

We might have to resort to something like

```cpp
struct mmx_m_struct {
	uint32_t m1;
	uint32_t m2;
};

struct FPU_P_Reg {
#if C_MMX
	union {
#endif
		mmx_m_struct m = {};
#if C_MMX

		MMX_reg reg_mmx;
	};
#endif

	uint16_t m3 = 0;

	uint16_t d1 = 0;
	uint32_t d2 = 0;
};
```

that gets rid of the warning for me and isn't too disruptive.

kklobe:
> I wonder what can we do about:
> 
> `../include/fpu.h:72:24: warning: ISO C++ prohibits anonymous structs [-Wpedantic]`
> 
> compiler warning. I would like to keep the existing FPU code intact, if possible.

I tried making a completely separate set of MMX registers which takes care of the warning, and so far both Heaven Seven and Extreme Assault work the same on macOS and Windows (H7 still doesn't work quite right on dyn-x86).

Check it out and let me know what you think. 567296f590e057649e9b84c08c5454a6b78c456a

kklobe:
ok @FeralChild64 I dug into the issues I was seeing and found out it was very straightforward to import the [dosbox-x](https://github.com/joncampbell123/dosbox-x/blob/master/src/cpu/core_dyn_x86/dyn_mmx.h) modifications. Heaven Seven now looks correct:

image

I've pushed to my branch.

FeralChild64:
@kklobe I have integrated the patch, thanks! Separating MMX registers from FPU registers to get rid of compiler warning is, IMHO, a bad idea - this will definitely break if more than one process at the same time tries to use MMX code under Windows, as the task switching won't be aware of the difference from the actual hardware.

------

Putting this back to draft now - this became more stable than I expected, I want to get rid of compiler / static analysis warnings and make it enabled by default. Might take few days, I am very short of time lately.

kklobe:
> @kklobe I have integrated the patch, thanks! Separating MMX registers from FPU registers to get rid of compiler warning is, IMHO, a bad idea - this will definitely break if more than one process at the same time tries to use MMX code under Windows, as the task switching won't be aware of the difference from the actual hardware.

This only separates the virtual registers used by Staging for non-dyn_86 cores - the actual machine registers when used by dyn_x86 are fully synchronized, and outside of it are never touched.

FeralChild64:
I'll try some pointer casting and references. I just want to get the summary of compiler warnings from different workflows first.

src/cpu/core_dyn_x86/dyn_mmx.h
```diff
@@ -0,0 +1,190 @@
...
+{
+	mem_writed(addr, mmxtmp.ud.d0);
+}
+
+static void MMX_LOAD_64(PhysPt addr)
+{
+	mmxtmp.ud.d0 = mem_readd(addr);
+	mmxtmp.ud.d1 = mem_readd(addr + 4);
+}
+
```
kklobe, line 39: Forgot to quadword these in this commit, added to my branch 778cd5eedc5a501ce14a565ce7bc307bcdec257f

FeralChild64: Already included :)

kklobe: I mean this should be:
```cpp
static void MMX_LOAD_64(PhysPt addr)
 {
 	mmxtmp.q = mem_readq(addr);
 }
```

kklobe:
P.S. I did test on Windows x86-32, works fine.

src/cpu/core_dyn_x86/dyn_mmx.h
FeralChild64: Ah, sorry. Cherry-picked just now.

kklobe:
As an experiment, I added simde intrinsics for all the MMX instructions: 5d23dc61a6b8c02aa950d2f246e155735e44475b

kklobe:
Looks like the intrinsics have negligible effect on performance. Here's my MacBook M3 with SIMDE:

```txt
       MMX Speed Test DOS Version 1.1 Fri Feb 23 10:01:28 2024
                 Copyright (C) 1997, Roy Longbottom

  Via Watcom Version 11, M = Standard Macros, E = Extended Macros

   Array Dimensions 512 64 bit words x 2, log file MMXres.txt

Test                        Passes   Secs     MOPS     MB/S    Result

Compare bytes pcmpeqb  M    121904   4.83    12.92    206.8  ff0000ff
Compare bytes pcmpeqb  E    188679   5.05    19.13    306.1  ff0000ff
Multiply add  pmaddwd  M    129292   5.11    12.95    207.3  00000020
Multiply add  pmaddwd  E    188679   5.05    19.13    306.1  00000020
Pack bytes    packuswb M    121904   4.94    12.63    202.2  efffff00
Pack bytes    packuswb E    180722   4.78    19.36    309.7  efffff00
Shift right   psrawi   M    136170   5.05    13.81    110.4  ffffffff
Shift right   psrawi   E    194805   5.11    19.52    156.1  ffffffff
Or 64 bits    por      M    129292   5.00    13.24    211.8  fff0f0ff
Or 64 bits    por      E    182926   4.89    19.15    306.4  fff0f0ff
42 ops plus moves      E     13913   5.11    58.55           01010101
```

without:

```txt
       MMX Speed Test DOS Version 1.1 Fri Feb 23 10:03:26 2024
                 Copyright (C) 1997, Roy Longbottom

  Via Watcom Version 11, M = Standard Macros, E = Extended Macros

   Array Dimensions 512 64 bit words x 2, log file MMXres.txt

Test                        Passes   Secs     MOPS     MB/S    Result

Compare bytes pcmpeqb  M    123076   5.00    12.60    201.6  ff0000ff
Compare bytes pcmpeqb  E    182926   4.94    18.96    303.3  ff0000ff
Multiply add  pmaddwd  M    129292   5.16    12.83    205.3  00000020
Multiply add  pmaddwd  E    182926   5.00    18.73    299.7  00000020
Pack bytes    packuswb M    116363   4.99    11.94    191.0  efffff00
Pack bytes    packuswb E    170454   4.95    17.63    282.1  efffff00
Shift right   psrawi   M    129292   4.78    13.85    110.8  ffffffff
Shift right   psrawi   E    188679   4.89    19.76    158.0  ffffffff
Or 64 bits    por      M    121904   4.72    13.22    211.6  fff0f0ff
Or 64 bits    por      E    182926   4.89    19.15    306.4  fff0f0ff
42 ops plus moves      E     12929   5.05    55.05           01010101
```

So they simplify the logic in our code, but, that's about it.

FeralChild64:
@kklobe Similar results with SiSoft Sandra 2001, on my AMD64 machine - both implementations produced almost the same results.

FeralChild64:
Right.

kklobe:
> [edit[ Sorry, it was with dyn-x86.

Got it, yeah, dyn_x86 JITs the MMX opcodes directly, the intrinsics code path only gets run on the non-dyn_x86 cores.

FeralChild64:
With dynrec the benchmark performance was just horrible - but also not much difference between implementations.

kklobe:
> With dynrec the benchmark performance was just horrible - but also not much difference between implementations.

We can use the SIMDE implementation for simplicity in our code, or just leave it out, since it's not helping performance.

FeralChild64:
It would indeed simplify our code, but - on the other hand - we would need to pull another 3rd party library. Also, SIMDE would be useful if someday someone decides to implement SSE instruction set family.

kklobe:
> It would indeed simplify our code, but - on the other hand - we would need to pull another 3rd party library. Also, SIMDE would be useful if someday someone decides to implement SSE instruction set family.
> 
> I’m really not sure what to do.

I do have the self-contained header just sitting in `src/libs/simde`, so there's no new dependency overhead. We can always add their `x86/sse.h` header as a one-off if we need it.

kklobe:
@FeralChild64 I rebased off the latest here and pushed a new branch for the simde stuff into a single commit, feel free to pull it in if you want. No library deps, simde is all in an amalgamated header in `src/libs/simde`. 384e3e997f475d0fcc1ee0600ea03ba3c965280c

FeralChild64:
OK, I have imported your change - but I have split out the library itself into a separate commit, so that it is easier to properly credit the author.

johnnovak:
If this is a well-maintained header-only library, I think using it is still better as then we can effectively outsource the SIMD stuff to them.

Would it be too much effort to include the version that does not use the library in our history as a previous commit for safekeeping? Then it's there in case we need to revert. The work has already been done so IMO it's worth preserving.

FeralChild64:
@johnnovak Everything is preserved, just revert the two last commits and you have the non-simde implementation.

johnnovak:
> @johnnovak Everything is preserved, just revert the two last commits and you have the non-simde implementation.

Cool, I guess my point was not to drop it in case you want to clean up the history before merging 😄

johnnovak:
Why `pentium_mmx_slow` @FeralChild64, why not just `pentium_mmx`? Kinda confusing as MMXs were the fastest true DOS-era machines 😄

FeralChild64:
> Why `pentium_mmx_slow` @FeralChild64, why not just `pentium_mmx`? Kinda confusing as MMXs were the fastest true DOS-era machines 😄

Because it is basically a `pentium_slow` with the MMX. I have no idea why Pentium and 486 are called _slow_ (if you ask me, I would prefer to get rid of them), and I don't really understand the difference between all these 386's.

johnnovak:
> Because it is basically a `pentium_slow` with the MMX. I have no idea why Pentium and 486 are called _slow_ (if you ask me, I would prefer to get rid of the _slow_ part), and I don't really understand the difference between all these 386's.

How about we call this `pentium_mmx` then and later add aliases for the rest with the slow suffix removed?

johnnovak:
> I'm not sure if our config file handling code supports aliases.

We do the mapping of setting values to enums anyway, so just add the variants without slow then map them to the same enums manually.

Grounded0:
We know this is battle-tested code for being years on DOSBox-X and this will give us a nice coverage of Pentium II feature level so good job people. DOSBox Staging is becoming heckuva Win9x disk image runner as Wine/CrossOver support for early Win32 stuff is becoming crappier.

FeralChild64:
@Grounded0 There is still a long way before DOSBox Staging becomes a good emulator for Windows 98. We have no proper page fault handling, no 3DNow!, no SSE (given the existence of the KernelEx project, SSE2 might be useful in some cases, too), if I’m not mistaken yielding on HLT is not implemented either, no APM, no FAT32 support by our MOUNT command, no VMWare directory sharing, VMWare mouse interface not compatible, no VMWare SVGA, no 3D acceleration passtrough (see < https://github.com/JHRobotics/softgpu> project), no A3D, no EAX.

johnnovak:
> There is still a long way before DOSBox Staging becomes a good emulator for Windows 98.

Plus the raw host CPU speed is just not there to emulate faster than about PII 450 CPUs. Maybe it will never happen on commodity hardware, or at least not in our lifetime.

FeralChild64:
@johnnovak Raw speed of a modern desktop CPU is enough to emulate a fast Windows 98 machine :D

benchmark

Of course this is just a synthetic benchmark, but I can clearly see that installing software (like Adobe Acrobat Reader, which took a lot of time to decompress) is crazy fast compared to what I remember from my old AMD K6-2 450 MHz.

@Torinde Sure, some points are much more important to handle - but we are not a corporation, but a bunch of volunteers, who primarily work on what is interesting to us, and Windows 98 is not the main priority.
Regarding your suggestion to use threads to emulate FPU - modern day CPUs (I mean - Pentium class or newer) are superscalar; it means they are smart enough to determine which adjacent instructions can be executed independently, in parallel, even within a single CPU core. The magic you propose is already being done, at the hardware level, it is being improved since decades, by corporations which have more (any by more I mean orders of magnitude) resources than the team.

johnnovak:
> Raw speed of a modern desktop CPU is enough to emulate a fast Windows 98 machine :D

Looks promising 😄 I was referring to this article:

https://86box.net/2022/03/21/why-not-p3.html

meson_options.txt
```diff
@@ -41,10 +41,10 @@ option(
 )
 
 option(
-    'experimental_mmx',
+    'emulate_mmx',
     type: 'boolean',
-    value: false,
-    description: 'Enable experimental Pentium MMX CPU emulation',
+    value: true,
+    description: 'Enable Pentium MMX CPU emulation',
```
johnnovak, line 47: What's the thinking behind keeping the flag for this, given it's not experimental anymore and this is not an external dependency? Caution in case we need to roll back / disable it?

src/dosbox.cpp
```diff
@@ -666,15 +666,28 @@ void DOSBOX_Init()
...
 	pstring->Set_help(
-	        "CPU type used in emulation ('auto' by default). 'auto' is the fastest choice.");
+	        "CPU type used in emulation ('pentium' by default).\n"
+#endif
+	        "Prefetch variant emulates the prefetch queue and requires the 'normal' core.");
+	pstring->SetDeprecatedWithAlternateValue("386", "386dx");
+	pstring->SetDeprecatedWithAlternateValue("386_slow", "386dx");
+	pstring->SetDeprecatedWithAlternateValue("386_prefetch", "386dx_prefetch");
+	pstring->SetDeprecatedWithAlternateValue("486_slow", "486dx");
+	pstring->SetDeprecatedWithAlternateValue("486_prefetch", "486dx");
```
johnnovak, line 690: Should be `486dx_prefetch`.

src/dosbox.cpp
```diff
@@ -666,15 +666,28 @@ void DOSBOX_Init()
...
+	        "CPU type used in emulation ('pentium_mmx' by default).\n"
+#else
+	pstring = secprop->Add_string("cputype", always, "pentium");
 	pstring->Set_values(cputype_values);
 	pstring->Set_help(
-	        "CPU type used in emulation ('auto' by default). 'auto' is the fastest choice.");
+	        "CPU type used in emulation ('pentium' by default).\n"
+#endif
+	        "Prefetch variant emulates the prefetch queue and requires the 'normal' core.");
+	pstring->SetDeprecatedWithAlternateValue("386", "386dx");
```
johnnovak, line 686: Should be `386dx_prefetch`.

src/dosbox.cpp
johnnovak: Actually, I'm not a fan of these `dx` suffixes; they don't add much value.

I'd like to suggest the following deprecation scheme instead:

```
386 -> 386_prefetch
386_slow -> 386
386_prefetch (no change)

486_slow -> 486
486_prefetch (no change)
``` 

I hope I got them right 😄

src/dosbox.cpp
FeralChild64: According to your suggestion above, I got rid of DX postfix.

src/dosbox.cpp
```diff
@@ -666,15 +666,17 @@ void DOSBOX_Init()
...
 	                  "if available and appropriate.");
 
-	const char* cputype_values[] = { "auto", "386", "386_slow", "486_slow", "pentium_slow",
-#if C_MMX
-	                                  "pentium_mmx",
-#endif
-	                                  "386_prefetch", nullptr};
-	pstring = secprop->Add_string("cputype", always, "auto");
+	const char* cputype_values[] = { "386", "486", "pentium", "pentium_mmx",
+	                                 "386_prefetch", "auto", nullptr};
```
johnnovak, line 670: I think this order would be better:

auto, 386, ..., 486, ..., pentium, ...

src/dosbox.cpp
```diff
@@ -666,15 +666,17 @@ void DOSBOX_Init()
...
+	pstring = secprop->Add_string("cputype", always, "pentium_mmx");
 	pstring->Set_values(cputype_values);
 	pstring->Set_help(
-	        "CPU type used in emulation ('auto' by default). 'auto' is the fastest choice.");
+	        "CPU type used in emulation ('pentium_mmx' by default).\n"
```
johnnovak, line 674: Is it wise to switch from `auto` to `pentium_mmx` if `auto` is the fastest? I'm thinking of backward compatibility.

I don't see any mentions of benchmarking in the comments. Are you sure `pentium_mmx` is just as fast as `auto` for non-MMX programs? I guess we'd need to benchmark first before we change the default CPU type.

Regardless of the chosen default, I'd keep the "'auto' is the fastest choice." text (unless we've proven that statement is no longer true).

Alternatively, leave the default alone, and we'll deal with it separately later.

src/dosbox.cpp
FeralChild64: OK, changed.

src/dosbox.cpp
FeralChild64: OK, changed.

FeralChild64:
@kklobe That's a huge improvement! If it's not too large / risky, I think we should include it in the current PR.

kklobe:
> @kklobe That's a huge improvement! If it's not too large / risky, I think we should include it in the current PR.

Let me know what you think: 052c83a97b548269f09fa4b45341b415b496c8d7

The final benchmarks - notice in particular that last line, that's probably the typical general performance boost. Heaven Seven now runs extremely smooth on my M3.

```log
Test                        Passes   Secs     MOPS     MB/S    Result

Compare bytes pcmpeqb  M    129292   5.21    12.71    203.3  ff0000ff
Compare bytes pcmpeqb  E    188679   5.06    19.09    305.5  ff0000ff
Multiply add  pmaddwd  M    123076   4.89    12.89    206.2  00000020
Multiply add  pmaddwd  E    188679   5.06    19.09    305.5  00000020
Pack bytes    packuswb M    130612   5.22    12.81    205.0  efffff00
Pack bytes    packuswb E    182926   4.78    19.59    313.5  efffff00
Shift right   psrawi   M    132231   4.83    14.02    112.1  ffffffff
Shift right   psrawi   E    187500   4.83    19.88    159.0  ffffffff
Or 64 bits    por      M    123076   4.78    13.18    210.9  fff0f0ff
Or 64 bits    por      E    180722   4.83    19.16    306.5  fff0f0ff
42 ops plus moves      E     13913   5.17    57.87           01010101
```
vs.:
```log
Test                        Passes   Secs     MOPS     MB/S    Result

Compare bytes pcmpeqb  M    544680   5.00    55.78    892.4  ff0000ff
Compare bytes pcmpeqb  E   1807228   4.78   193.58   3097.2  ff0000ff
Multiply add  pmaddwd  M    533333   5.17    52.82    845.1  00000020
Multiply add  pmaddwd  E   1818181   4.78   194.75   3116.0  00000020
Pack bytes    packuswb M    528925   4.94    54.82    877.1  efffff00
Pack bytes    packuswb E   1886792   5.11   189.05   3024.8  efffff00
Shift right   psrawi   M    680851   5.00    69.72    557.8  ffffffff
Shift right   psrawi   E   2272727   4.94   235.55   1884.4  ffffffff
Or 64 bits    por      M    528925   4.84    55.95    895.2  fff0f0ff
Or 64 bits    por      E   1875000   4.94   194.33   3109.3  fff0f0ff
42 ops plus moves      E    181818   5.27   741.90           01010101
```

johnnovak:
> Let me know what you think: [052c83a](https://github.com/dosbox-staging/dosbox-staging/commit/052c83a97b548269f09fa4b45341b415b496c8d7)

Whoa, 5-10x speed improvement? Gives this man a medal, that's what I think! 😎 🤘🏻 🏅 

This MMX stuff came together amazingly quickly and these improvements elevate it to the next level, thanks to you @FeralChild64 and @kklobe! Amazing work!

And of course thanks to the original author too. I will send him a not once this is merged (or if you guys want to do that, that's fine, of course).

johnnovak:
```
Total: 161 issues (out of 109 allowed)
```

This needs a bit of more work it seems 😅

FeralChild64:
First we need to understand why the PVS Studio does not like this code:

Screenshot

johnnovak:
> First we need to understand why the PVS Studio does not like this code. Well, not today - where I live it's very late now.

Every case branch in that file starts with this, so I'm pretty sure it wants it to be lifted out. But that won't be so easy because of all this dirty preprocessor macro hackery. Maybe 

```cpp
if (CPU_ArchitectureType < ArchitectureType::PentiumMmx) goto illegal_opcode;
```

Alternatively, you can disable these V1037 warnings one by one on the lines where they occur with the PVS-Studio special suppression comments:

```cpp
if (CPU_ArchitectureType < ArchitectureType::PentiumMmx) goto illegal_opcode; //-V1037
```

More info:
https://pvs-studio.com/en/docs/manual/0017/

src/dosbox.cpp
```diff
@@ -666,15 +666,18 @@ void DOSBOX_Init()
...
 	                  "if available and appropriate.");
 
-	const char* cputype_values[] = { "auto", "386", "386_slow", "486_slow", "pentium_slow",
-#if C_MMX
-	                                  "pentium_mmx",
-#endif
-	                                  "386_prefetch", nullptr};
+	const char* cputype_values[] = { "auto", "386", "486", "pentium",
+	                                 "pentium_mmx", "386_prefetch",
+	                                 nullptr };
```
johnnovak, line 671: Ok, so let's go with this order and let's expose the old `386` setting a `386_fast`. We don't quite understand what we'd lose by removing it, so let's keep it for now.

```
"auto", "386", "386_fast", "386_prefetch", "486", "pentium", "pentium_mmx"
```

johnnovak:
Heaven Seven works very well with `cycles=max 95% limit 400000`, great job.

kklobe:
Another game for testing: Z.A.R. demo, run `zarmmx.exe`. https://www.dosgamesarchive.com/download/zar

kklobe:
> Heaven Seven works very well with `cycles=max 95% limit 400000`, great job.

You and your `limit`. :grin:

kklobe:
> Heaven Seven works very well with `cycles=max 95% limit 400000`, great job.

It even runs decently at 640x480 (`heaven7d -2`), barely stuttered through it before.

Grounded0:
dosbox_cputype

src/dosbox.cpp
johnnovak: We don't want to overcomplicate it so let's stay with my above suggestion. Maybe we'll mention these details in the manual, but these detail say very little to 99.99% of users.

FeralChild64: OK, changed. I have tried to benchmark "386_fast" against "386", but - strangely - I'm getting same results.

FeralChild64:
@johnnovak OK, I have simply silenced the PVS Studio warnings.

johnnovak (approved):

johnnovak:
Very nice job @FeralChild64 & @kklobe, merge away 🚀

johnnovak:
> thus I think it will be a good next step

Towards what end? MMX makes sense for an extremely small number of DOS titles, but 3DNow!, and anything starting from PII is totally out of scope for this project.

> pentium_mmx, pentium_ii, athlon (P2+3DNow), athlon_xp (P2+3DNow+SSE)

Again, why? Win9x emulation is still out of scope.

Grounded0:
MMX was implemented by both Intel and AMD from 1997 onwards while 3DNow! was not. It makes sense to target something that the software actually uses which means it was implemented by both x86 manufacturers.

johnnovak:
> understand neither is a priority. And sorry for hijacking the thread.

To be clear, I see these as the upper limits for the scope of the project:

- Pentium MMX (but technically we can veer into P2 speed territory with cycles 400-500k)
- VESA, Voodoo 1
- AWE32

Anything after this is not classic DOS. Yeah there are post-2000 or even recent developments, but those are extremely niche and insignificant (in the grand scheme of things).

The no.1 reason for the existence of DOSBox is to be able to play the vast DOS and Win3.x gaming catalogue between 1982 and 1997 or so. Everything else, even demos, are extra. Yeah and some people run old DOS and Win3.x apps, but that's the 1% use case.

Grounded0:
I looked up and AMD K6 series doesn't implement `CMOV` and other P6 instructions so it's unlikely any Windows game prior to 2003 would've required `CMOV` because cutting off AMD K6 series would've been out of question. So even for Win 9x disk image runner it's not essential.

FeralChild64:
Mine :) But the results are not representative for this CPU, as I have curbed down the CPU power consumption in UEFI settings to keep the noise within reasonable limits.

PR #3428, opened 2024-02-21
Fix double-scanning behavior for SVGA/VESA games

### LINKED ISSUE #3416: Wrong scaling in Wing Commander 3

_No response_

Windows 10

_No response_

### Relevant hardware info

_No response_

Wing Commander 3 is only showing part of its display area in SVGA:

image0001

Pressing the Win key fixes it:

image0002

Proceeding to the flight control breaks it again:

image0003

This time alt-tabbing or pressing Win has no effect.

Introduction video is also cut in half and displayed in the bottom part of the screen:

image0004

UNIVBE has no effect on this issue.

Console insists the original display is 640x240 256 colors and gets scaled to 1920x1440 pixels with 1:2 (1:2) pixel aspect ratio (instead of 640x480 pixels).

Explain how to reproduce
1. Install Wing Commander 3 with SVGA options enabled. 
2. Run with WC3.EXE as per usual.

_No response_

### Your configuration

```ini
machine            = svga_s3
aspect             = auto
integer_scaling    = vertical
viewport           = fit
```

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

The branch which sometimes skips VGA re-sizing for double-scan modes does not work at SVGA/VESA resolutions.  Those modes have a "faked" double-scan that gets calculated here:

https://github.com/dosbox-staging/dosbox-staging/blob/0c8e8cec90cdbda8821492e1b3094b58d11746cd/src/hardware/vga_draw.cpp#L2108-L2119

Simply doubling `vga.draw.address_line_total` is not enough.

## Related issues

Fixes #3416

# Manual testing

Confirmed this fixes Wing Commander 3.  That game switches into double-scanning (640X240) while playing videos and then back out again (640X480) when the video finishes playing.  These transitions were not being handled properly before.

Regression Testing:

- Watched **Second Reality** demo.  Looks fine.
- **Catacomb 3D** works
- **Legend of Kyrandia** scrolling bug is not present
- **Threat** works

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_crtc.cpp
+++ b/src/hardware/vga_crtc.cpp
@@ -189,31 +189,14 @@ void vga_write_p3d5(io_port_t, io_val_t value, io_width_t)
 
 		vga.crtc.maximum_scan_line.data = new_val.data;
 
-		if (vga.draw.is_double_scanning) {
-			auto flipped = new_val;
-			flipped.start_vertical_blanking_bit9.flip();
-
-			// Only true if only `start_vertical_blanking_bit9` has
-			// been changed and all other bits are unchanged.
-			if (old_val.data == flipped.data) {
-				VGA_StartResize();
-			}
-
-			vga.draw.address_line_total = new_val.maximum_scan_line + 1;
-
-			if (new_val.is_scan_doubling_enabled) {
-				vga.draw.address_line_total *= 2;
-			}
-		} else {
-			// Start resize if any bit except `line_compare_bit9` has been
-			// changed.
-			if ((old_val.maximum_scan_line != new_val.maximum_scan_line) ||
-			    (old_val.start_vertical_blanking_bit9 !=
-			     new_val.start_vertical_blanking_bit9) ||
-			    (old_val.is_scan_doubling_enabled !=
-			     new_val.is_scan_doubling_enabled)) {
-				VGA_StartResize();
-			}
+		// Start resize if any bit except `line_compare_bit9` has been
+		// changed.
+		if ((old_val.maximum_scan_line != new_val.maximum_scan_line) ||
+		    (old_val.start_vertical_blanking_bit9 !=
+		     new_val.start_vertical_blanking_bit9) ||
+		    (old_val.is_scan_doubling_enabled !=
+		     new_val.is_scan_doubling_enabled)) {
+			VGA_StartResize();
 		}
 	} break;
 


src/hardware/vga_crtc.cpp
```diff
@@ -26,6 +26,12 @@
 #include "pic.h"
 #include "vga.h"
 
+static bool is_svga_or_vesa(VGAModes mode)
+{
+	return mode == M_LIN4 || mode == M_LIN8 || mode == M_LIN15 ||
+		mode == M_LIN16 || mode == M_LIN24 || mode == M_LIN32;
+}
+
```
weirddan455, line 34: The C hacker in me kind of wants to change this to:

```c
return mode & (M_LIN4 | MLIN_8 | M_LIN15 | M_LIN16 | M_LIN24 | M_LIN32);
```
Since this enum is also a bitfield.  At the same, mis-reading what bitwise instructions are doing has been a problem.

src/hardware/vga_crtc.cpp
```diff
@@ -189,21 +195,21 @@ void vga_write_p3d5(io_port_t, io_val_t value, io_width_t)
...
-			// Only true if only `start_vertical_blanking_bit9` has
-			// been changed and all other bits are unchanged.
-			if (old_val.data == flipped.data) {
-				VGA_StartResize();
-			}
-
+		if (vga.draw.is_double_scanning && !is_svga_or_vesa(vga.mode)) {
+			// This is a potential fast path where we can skip VGA resizing.
+			// A rare few games (ex. Wing Commander 3) fiddle with the double-scan bit
+			// in SVGA/VESA modes.
```
johnnovak, line 201: I'd prefer if you forced a resizing as I don't like it if this algorithm becomes too "distributed". We don't want a "fast" path here as the resize logic is fast enough, but we absolutely want as much clarity as we can achieve.

What should happen here is detecting the conditions that require a resize, then modify the big resize function (if needed; maybe triggering a resize here is what's missing).

weirddan455: "Fast path" is probably the wrong wording as it's not really about performance.  My first thought was just to call `VGA_StartResize` all the time but you tried that and it apparently broke Legend of Kyriandia then #3279 restored this behavior.

I guess I'm unclear what you mean if not that.

src/hardware/vga_crtc.cpp
weirddan455: I figured you would say something like that :smile:

src/hardware/vga_crtc.cpp
johnnovak: Okay, just proceed as you think is best then we'll talk at review time 😄

weirddan455: I think I see what you mean now.  I did some regression testing and I think we can get rid of this special double-scan path.

If you avoid all checking, the Legend of Kyrandia bug returns.  However, if you keep only the bottom (general) path, everything seems to work.

I force pushed this to a single commit to remove the top branch.  It is nice when you can fix a bug simply by deleting code :smile:

johnnovak: Yeah, this looks good. I meant only to trigger the resize from here, then updating the VGA state based on the register value change should be done in the big `setup_drawing()` function in `vga_draw.cpp` (see below).

If this simplification works, then great 🎉 I will need to read `setup_drawing()` again in light of this change for correctness, then test with a bunch of special VGA programs that do weird things to the hardware... The good news is the outcome of that will be a VGA regression testing wiki page that we sorely need.

Great job so far @weirddan455 👍🏻 Good to see I'm not the only one hacking away at the VGA code.

```cpp
// Can change the following VGA state:
//
//   vga.draw.address_line_total
//   vga.draw.blocks
//   vga.draw.delay.hdend
//   vga.draw.dos_refresh_hz
//   vga.draw.line_length
//   vga.draw.linear_base
//   vga.draw.linear_mask
//   vga.draw.mode
//   vga.draw.resizing
//   vga.draw.uses_vga_palette
//   vga.draw.vblank_skip
//   vga.draw.vret_triggered
//
ImageInfo setup_drawing()
{
```

weirddan455: `vga.draw.address_line_total` (only VGA variable that previously got set here) already gets re-set in `setup_drawing()` according to the register values (and the code there correctly handles edge cases unlike here).  I don't think anything should need changing there.

It seems this "fast path" logic existed before your improvements to double-scanning and so is incorrect now.  Hopefully there won't be any problems just removing it as that would be the simplest solution.

johnnovak: > It seems this "fast path" logic existed before your improvements to double-scanning and so is incorrect now.

Yeah, and I didn't like it, just was reluctant to touch it 😄 

> vga.draw.address_line_total (only VGA variable that previously got set here) already gets re-set in setup_drawing() according to the register values (and the code there correctly handles edge cases unlike here). I don't think anything should need changing there.

That's my hope. I'll just need to test the crap out of it 😄

johnnovak:
Okay, I've expanded the video emulation tests page with lots of VGA tests based on our recent experience with this thing:
https://github.com/dosbox-staging/dosbox-staging/wiki/Video-emulation-tests#vga

I've done pretty much all of that, specifically:

- Testing all 15/16-bit high colour and 24/32-bit true colour VESA modes in QPV.
- All VGA modes in Deluxe Paint II
- Most of the games I added to the VGA tests section.

johnnovak (approved):
Nice one @weirddan455

weirddan455:
Thanks for the tests and the documentation @johnnovak Hopefully this is the last time we have to touch this function for a while.

PR #3419, opened 2024-02-18
Bump Coverity build from Clang 14 to Clang 15

weirddan455:
# Description

Coverity build has been failing since the PR to use C++20.  Bump the compiler version to match what our Ubuntu 22.04 build is running.

# Manual testing

Only way to test is to merge to main I believe.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [ ] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #3437, opened 2024-02-22
Wd/pvs

weirddan455:
# Description

PVS is failing with the same error that Coverity was.  Apply the same fix of bumping the compiler version.

## Related issues

Previous Coverity PR: https://github.com/dosbox-staging/dosbox-staging/pull/3419

# Manual testing

This PR has a temporary commit to test PVS.

# Checklist

I have:


weirddan455:
Wow OK so I think the actual reason PVS turned red is because it's giving us a warning that our license will expire in 29 days.  I went back and looked at past builds and they were green even though they had this build failure.  It was silently failing with a compile error and just only ran static analysis on about 50 of our files.

There is an temporary commit in this PR I intend to remove before merge but I'll leave it as is for now so reviewers can see the build.

johnnovak:
> Wow OK so I think the actual reason PVS turned red is because it's giving us a warning that our license will expire in 29 days.

I will look into renewing the license. Kcgen explained it to me a while again how to do it, need to dig it up.

kklobe:
I've disabled it manually for now

johnnovak (approved):
Looks fine @weirddan455, I've contacted the PVS-Studio guys about the renewal of our yearly license. Plus it will be under my email address now instead of kcgen's. We'll re-run this once I've received the new key.

johnnovak:
@weirddan455 I've renewed our PVS-Studio license for another year, should be good now. Please rebase to main to get 26263d3032, adjust the warning count if needed then it's good to go.

weirddan455:
Good stuff @johnnovak 

PVS is green now.  We've picked up 2 additional warnings somewhere along the way so that's been increased.  The silent build failure is a bit concerning to me but I'm not sure how to fix it.  The red/green status seems to only come from the Analysis step.  If the build step fails, it proceeds to analysis with whatever intermediate .o files it was able to build.

I'll go ahead and merge this though.

johnnovak:
> Good stuff @johnnovak
> 
> PVS is green now. We've picked up 2 additional warnings somewhere along the way so that's been increased.

That's ok, we'll cut back on the warnings later.

> The silent build failure is a bit concerning to me but I'm not sure how to fix it. The red/green status seems to only come from the Analysis step. If the build step fails, it proceeds to analysis with whatever intermediate .o files it was able to build.

Maybe fixable later, but the PVS workflow runs on pushes anyway so the build workflows would fail too. So in a practical sense hardly a concern (but a bit dodgy, yeah).

Issue #3434, opened 2024-02-22
Show by Majic 12 video regression


johnnovak:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

macOS (Apple Silicon)

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The zoom-rotator part in the intro of Show by Majic 12 has regressed. As this used to be my group in the 90s, I'm rather saddened by this unfortunate turn of events 😢 

## main

With `svga_s3`. This is bad.

<img width="1072" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/4b63535b-bc91-49b8-80b6-3914eefac496">

## 0.80.1

Same results with `svga_s3` and `vgaonly`. The results are perfect.

<img width="1072" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/557caf74-04bd-4b03-b786-b5a4f98a1db5">

### Steps to reproduce the behaviour.

Start the demo (needs GUS) then watch the intro. Bamm!

### Download URL of affected game or software

https://www.pouet.net/prod.php?which=1158

### Your configuration

```ini
Stock
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

weirddan455:
Rather large commit unfortunately.

```
bc924b6f7cb966e5a2f1b0235e0d43a5bde68fd4 is the first bad commit
commit bc924b6f7cb966e5a2f1b0235e0d43a5bde68fd4
Author: John Novak
Date:   Tue Aug 22 22:32:08 2023 +1000

    Implement adaptive CRT emulation shaders
    
    - All shader management code has been moved from 'render.cpp' to
      the ShaderManager.
    
    - The 'default' shader alias to 'interpolation/sharp' has been replaced
      with the 'sharp' alias.
    
    - Three new adaptive (auto-switching) CRT emulation shader modes have
      been added: 'crt-auto', 'crt-auto-machine', and 'crt-auto-arcade'.
      These modes switch dynamically between CRT shader variants depending
      on the current video mode, the graphics adapter in use, and the
      viewport resolution (see the description of the 'glshader' setting for
      further details).

 include/render.h           |  35 +--
 src/gui/meson.build        |   5 +-
 src/gui/render.cpp         | 453 ++++++++++++++------------------------
[...]

PR #3441, opened 2024-02-23
Fix "unorthodox" scanline repeats on VGA

johnnovak:
# Description

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3434

The original DOSBox code contained a special case for handling "unorthodox scan doublings" that we lost during the removal of the `vgaonly` refactors.

Some rare demos set up odd Maximum Scan Line CRTC register values; for example, Show by Majic 12 uses the value 4 during the zoom-rotator part in the intro to set up "scanline quintupling". That's right, every scanline is repeated 4 times, resulting in a total number of 5 scanlines per "logical pixel"!

This change fixes the forced single scan case (the authentic double-scanned VGA emulation path was fine). Effectively, scanline repeating is forced when encountering odd Maximum Scan Line values, even in forced single scan mode.

**Note the bug _only_ manifested with forced single scanning. To force single scanning, you need to use `glshader = sharp`, `glshader = crt-auto-arcade`, or `output = texturenb`.**

This is where the original code performed this special casing (it's not so easy to see as I've refactoring the code substantially since then... How did I find this out? By comparing trace logs between the old and the new VGA code for a few hours...)
https://github.com/dosbox-staging/dosbox-staging/blob/release/0.80.x/src/hardware/vga_draw.cpp#L1803-L1811

### Show by Majic 12, zoom rotator part

image

### Zoomed-in view

Count the scanlines — there are five per each "logical pixel"!

image

# Manual testing

Tested the following:

- Copper demo
- Show by Majic 12
- Second Reality by Future Crew
- Catacomb 3D
- Gods
- The Legend of Kyrandia: Book One
- Ravenloft - Strahd's Possession
- Pinball Fantasies
- Pinball Illusions

With these `glshader` settings:

- `crt-auto` (double scanning)
- `crt-auto-arcade` (forces single scanning)
- `sharp` (forces single scanning)

# Checklist

I have:

### DIFF
--- a/include/vga.h
+++ b/include/vga.h
@@ -291,8 +291,35 @@ struct VgaDraw {
 	double custom_refresh_hz  = RefreshRateDosDefault;
 	VgaRateMode dos_rate_mode = VgaRateMode::Default;
 
-	bool double_scanning_enabled = false;
-	bool pixel_doubling_enabled  = false;
+	// If true, double-scanned VGA modes are allowed to be drawn as
+	// double-scanned. For example, the 13h 320x200 mode is drawn as 640x400
+	// (assuming pixel doubling is also allowed).
+	//
+	// If false, double-scanned VGA modes are forced to be drawn as
+	// single-scanned. In other words, video modes are drawn at their "nominal
+	// height". E.g., the 13h 320x200 mode is drawn as 640x200 (assuming pixel
+	// doubling is allowed). The exception to this are the special custom
+	// VGA modes used in some demos that use odd number of scanline repeats
+	// (e.g., 3 or 5); these are always drawn as scan-tripled, quintupled,
+	// etc. even if this flag is false.
+	//
+	// Single scanning is forced by the arcade shaders to achieve the
+	// single-scanned 15 kHz CRT look for double-scanned VGA modes, or by
+	// shaders that treat pixels as flat adjacent rectangles (e.g., the
+	// "sharp" shader and the "no-bilinear" output modes; the double-scanned
+	// and force single-scanned output is exactly identical in these cases,
+	// but single scanning is more performant which matter on low-powered
+	// devices).
+	bool scan_doubling_allowed   = false;
+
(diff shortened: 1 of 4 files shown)


weirddan455 (approved):
Nice job, another VGA edge case down :heavy_check_mark: 

Not expecting you to do anything else with this PR but I did want to point out these double scan variables are a bit confusing.  We have:

- is_double_scanning (local variable)
- vga.draw.is_double_scanning (global variable with the same name)
- vga.draw.double_scanning_enabled
- is_vga_scan_doubling() (function call)

Are we double scanning?  I don't know, check the state of one of these 4 booleans with almost identical names :shrug:

johnnovak:
> Nice job, another VGA edge case down ✔️

Cheers 😎 

> Not expecting you to do anything else with this PR but I did want to point out these double scan variables are a bit confusing. We have:
> 
>     * is_double_scanning (local variable)
> 
>     * vga.draw.is_double_scanning (global variable with the same name)

We need the temp var because whether we set the one in `vga.draw` which is the final calculated draw state depends on other flags (see below).

>     * vga.draw.double_scanning_enabled

Ok, I should document this in the struct. This means that double scanning is allowed (it is set to false in forced single scan mode).

>     * is_vga_scan_doubling() (function call)

There is a big comment for this one; this checks if we should be double scanning based on the VGA register states.

> Are we double scanning? I don't know, check the state of one of these 4 booleans with almost identical names 🤷

Yeah, unfortunately I don't think this can be much improved without putting a whole novel into each and every name 😄 You just need to be familiar with the code. But I'll add more explanatory comments.

johnnovak:
Okay, @weirddan455, on second thought, I think you were right, so I've clarified the double-scanning logic a bit. I'm hoping things are a lot clearer now. Please see the extra commits.

weirddan455 (approved):
Thanks, that helps clarify things.

Issue #3436, opened 2024-02-22
Impossible Mission II: SIGFPE crash in Tandy mode


dani-jozsef:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Linux x86_64

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The game **Impossible Mission II** (Novotrade / Epyx, 1988) crashes with a Floating Point Exception when loading the first game level on `tandy` machine type.

### Steps to reproduce the behaviour.

1. Start emulation with `--machine tandy` command line parameter (the config below is svga by default)
2. Run the game with `im.com` or `im2_tan.exe`
3. Proceed past the title screen and questions about joystick and saved games
4. Note DOSBOX crashing (with a series of strange video mode switches in the log before)

### Download URL of affected game or software

https://www.abandonwaredos.com/abandonware-game.php?abandonware=Impossible+Mission+2&gid=1873

### Your configuration

```ini
https://gist.github.com/dani-jozsef/b07dfef964e095b7dba8e845c514aaaf
```

### Provide a Log

```text
https://gist.github.com/dani-jozsef/5246a9a481d106f36f07bb72851509ca
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Could you test on 0.80.1 as well @dani-jozsef to determine if this is a long-standing bug or a regression?

dani-jozsef:
This seems to be a regression. The crash doesn't happen on 0.80.1. Here's the run log to the point where the game starts:
https://gist.github.com/dani-jozsef/c7dfaf59ad1350499db4d384af0d41b0

weirddan455:
I'll give it I bisect.

weirddan455:
Removing the FPU emulation tag.  This is a good ole integer divide by zero (which for some reason on Linux translates to SIGFPE but on the host side, not the DOS side).  Seems to be caused by a VGA timing variable not being set correctly.

johnnovak:
> 4\. Note DOSBOX crashing (with a series of strange video mode switches in the log before)

That's "normal"; even DOSBox SVN does that. It changes the window size to the current resolution, so you can see the window shrink and grow.

PR #3445, opened 2024-02-24
Guard against zero-height modes when setting up a new video mode

johnnovak:
# Description

Some games set up weird custom modes temporarily that result in a zero-height video mode (e.g., Impossible Mission II on Tandy). These are probably some tricks that work on hardware, but in software we need to guard against divide-by-zero exceptions downstream when calculating the aspect ratios.

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3436

# Manual testing

Impossible Mission 2 works with `machine = tandy` (use `IM2_TAN.EXE`).

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -25,6 +25,7 @@
 #include <array>
 #include <cmath>
 #include <cstring>
+#include <optional>
 #include <utility>
 
 #include "../gui/render_scalers.h"
@@ -1857,6 +1858,8 @@ static Fraction calc_pixel_aspect_from_dimensions(const uint16_t width,
                                                   const bool double_width,
                                                   const bool double_height)
 {
+	assert(height > 0);
+
 	const auto storage_aspect_ratio =
 	        Fraction(static_cast<int64_t>(width) * (double_width ? 2 : 1),
 	                 static_cast<int64_t>(height) * (double_height ? 2 : 1));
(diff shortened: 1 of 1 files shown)


weirddan455:
Yeah I was just looking into this.  This fixes the crash but there's a visual effect they were going for that doesn't look quite right with this fix applied.  Compare this branch to v0.80.1.

On v0.80.1 when the level starts, the video starts with a thin line in the center and expands out vertically until it fills the entire screen.  On this branch it just looks kind of janky.

weirddan455:
Bisected the visual bug to your big double-scanning rework commit dd8ae1423a9d496f78fdcf39798568aea6bbadca

I just used this patch to work around the crash since it applies cleanly to everything from 0.80.1 to current main:

```
diff --git a/src/hardware/vga_other.cpp b/src/hardware/vga_other.cpp
index 308694ceb..4e473c7c3 100644
--- a/src/hardware/vga_other.cpp
+++ b/src/hardware/vga_other.cpp
@@ -87,7 +87,7 @@ static void write_crtc_data_other(io_port_t, io_val_t value, io_width_t)
                vga.other.vadjust = val;
                break;
        case 0x06:              //Vertical rows
-               if (vga.other.vdend ^ val) VGA_StartResize();
+               if ((val > 0) && (vga.other.vdend ^ val)) VGA_StartResize();
                vga.other.vdend = val;
                break;
        case 0x07:              //Vertical sync position
```

johnnovak:
> I just used this patch to work around the crash since it applies cleanly to everything from 0.80.1 to current main:

Yeah, I actually like yours better. Please raise a PR and then I'll close this. It's better not to even call resize in this case.

> On v0.80.1 when the level starts, the video starts with a thin line in the center and expands out vertically until it fills the entire screen. On this branch it just looks kind of janky.

Indeed. Thanks for bisecting. I'm not quite keen on figuring this out right now, so could you please raise a ticket for it? Arguably, it's a minor thing but we should keep track of it and eventually fix it.

The crash fix, however, needs to go in now and we should backport it.

johnnovak:
Fixed in https://github.com/dosbox-staging/dosbox-staging/pull/3448

PR #3454, opened 2024-02-27
Fix Impossible Mission II transition effect

### LINKED ISSUE #3447: Impossible Mission II: Visual Glitch on Tandy

_No response_

Linux x86_64

Arch Linux

### Relevant hardware info

_No response_

As the first level is loading in, there is an animation where the screen starts out as a narrow band in the center and then expands out vertically in both directions until it fills the screen completely.

This animation looks correct in v0.80.1 and incorrect on main.  This is a regression that I have bisected to commit dd8ae1423a9d496f78fdcf39798568aea6bbadca

Note there is a (mostly) unrelated divide by zero crash that must be worked around if checking out older commits.

Explain how to reproduce
1.  Put `machine = tandy` in your config
2. Run IM.COM or IM2_TAN .EXE (the former seems to auto-detect machine type and load the latter)
3. Skip past intro and "Load save game" screen (just press enter)
4. Observe the animation as the level loads in

https://www.abandonwaredos.com/abandonware-game.php?abandonware=Impossible+Mission+2&gid=1873

### Your configuration

```ini
[dosbox]
machine = tandy
```

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

First commit addresses a regression caused by b68b66f07831c91f197acf0b06f847c28c9a7946.  It causes flickering when the video mode changes quickly during the transition effect.

Second commit addresses a regression caused by https://github.com/dosbox-staging/dosbox-staging/commit/dd8ae1423a9d496f78fdcf39798568aea6bbadca.  Previously, the aspect ratio for Tandy was hardcoded to 5/6.  We have since added some complexity where pixel-doubling and double-scanning need to be taken into account.  However, the width needs to be hard-coded to 200 so this effect renders correctly.

## Related issues

Fixes #3447

# Manual testing

With `machine = tandy` tested the following with both `glshader=crt-auto` (pixel doubling enabled) and `glshader=sharp` (pixel doubling disabled).

- Tested Impossible Mission II (160 x 200)
- Tested King's Quest 4 (320 x 200)
- Tested Space Quest 1 (TANDY 320 x 200 plus CGA text-mode 320x200)

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -2054,6 +2054,12 @@ ImageInfo setup_drawing()
 		return pcjr_or_tga();
 	};
 
+	// All Tandy modes have a height of 200.
+	// Some games (ex. Impossible Mission II) fiddle with vga.other.vdend
+	// The result of this should be rendering a short (by height) image in the horizonal center with black on the top/bottom.
+	// Use this hard-coded value when calculating pixel aspect ratio so this effect looks correct.
+	constexpr uint16_t CgaTandyAspectHeight = 200;
+
 	switch (vga.mode) {
 	case M_LIN4:
 	case M_LIN8:
@@ -2364,7 +2370,7 @@ ImageInfo setup_drawing()
 		render_height     = video_mode.height;
 
 		render_pixel_aspect_ratio = calc_pixel_aspect_from_dimensions(
-		        render_width, render_height, double_width, double_height);
+		        render_width, CgaTandyAspectHeight, double_width, double_height);
 		break;
 
 	case M_TANDY4:
@@ -2402,7 +2408,7 @@ ImageInfo setup_drawing()
 		render_height = video_mode.height;
 
 		render_pixel_aspect_ratio = calc_pixel_aspect_from_dimensions(
-		        render_width, render_height, double_width, double_height);
+		        render_width, CgaTandyAspectHeight, double_width, double_height);
 
 		// TODO this seems like overkill; could be probably simplified a
 		// lot
@@ -2452,7 +2458,7 @@ ImageInfo setup_drawing()
 		render_height = video_mode.height;
 
 		render_pixel_aspect_ratio = calc_pixel_aspect_from_dimensions(
-		        render_width, render_height, double_width, double_height);
+		        render_width, CgaTandyAspectHeight, double_width, double_height);
 
 		VGA_DrawLine = VGA_Draw_1BPP_Line;
 		break;
@@ -2563,7 +2569,7 @@ ImageInfo setup_drawing()
 		render_height = video_mode.height;
 
 		render_pixel_aspect_ratio = calc_pixel_aspect_from_dimensions(
-		        render_width, render_height, double_width, double_height);
+		        render_width, CgaTandyAspectHeight, double_width, double_height);
 
 		VGA_DrawLine = VGA_Draw_CGA4_Composite_Line;
 		break;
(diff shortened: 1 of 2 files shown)


src/hardware/vga_other.cpp
```diff
@@ -92,8 +92,11 @@ static void write_crtc_data_other(io_port_t, io_val_t value, io_width_t)
 	case 0x06:		//Vertical rows
 		// Impossible Mission II sets this to zero briefly
 		// This leads to a divide by zero crash if VGA resize code is run
-		if ((val > 0) && (vga.other.vdend ^ val)) VGA_StartResize();
-		vga.other.vdend = val;
+		if (val > 0 && val != vga.other.vdend) {
+			vga.other.vdend = val;
+			// The default half-frame peroid delay leads to flickering
```
johnnovak, line 97: typo: period

johnnovak: "leads to flickering in the level start effect of Impossible Mission II on Tandy"

src/hardware/vga_draw.cpp
```diff
@@ -2054,6 +2054,13 @@ ImageInfo setup_drawing()
 		return pcjr_or_tga();
 	};
 
+	// As far as I can tell, all Tandy modes have a height of 200.
```
johnnovak, line 2057: Let's be more assertive about this 😄 "All Tandy modes have a height of 200 pixels."

src/hardware/vga_draw.cpp
```diff
@@ -2054,6 +2054,13 @@ ImageInfo setup_drawing()
...
+	// Some games (ex. Impossible Mission II) fiddle with vga.other.vdend
+	// The result of this should be rendering a short (by height) image in the horizonal center
+	// With black on the top/bottom
```
johnnovak, line 2060: minor: join the above two lines into a single sentence, period at the end.

johnnovak:
Very nice @weirddan455 🎉 

We should apply the fix to the composite Tandy modes as well. If you run the game with this config, the problem is back:

```ini
[dosbox]
machine = tandy

[composite]
composite = on
```

The solution is to apply the fix to these paths as well:

```
M_CGA4_COMPOSITE
M_CGA2_COMPOSITE
```

src/hardware/vga_draw.cpp
```diff
@@ -2054,6 +2054,13 @@ ImageInfo setup_drawing()
...
+	// Use this hard-coded value when calculating pixel aspect ratio so this effect looks correct.
+	constexpr uint16_t TandyAspectHeight = 200;
```
johnnovak, line 2062: Given we need to fix the composite paths too, and those composite modes affect CGA too, maybe call this `CgaTandyAspectHeight`? Well, PCjr would be affected too, but then the name would be even sillier...

src/hardware/vga_other.cpp
weirddan455: fixed

src/hardware/vga_draw.cpp
weirddan455: fixed

src/hardware/vga_draw.cpp
weirddan455: fixed

src/hardware/vga_draw.cpp
weirddan455: Yep, good catch.  Applied the change to the composite modes and confirmed it looks correct there as well.

johnnovak (approved):
Nice job @weirddan455 , ship it 😎

PR #3460, opened 2024-03-01
Make PCjr application memory contiguous

### LINKED ISSUE #2851: Space Quest I hangs on PCjr
When starting Space Quest I with the following config (eXoDOS v5 version), the emulator hangs and never recovers. Same behaviour with current `main` and v0.80.1. 

```ini
[dosbox]
machine = pcjr

[autoexec]
c:
sq
```

I'm not even sure if the eXoDOS version I'm using has functional PCjr support or not, but a bug is a bug; we should at least fail gracefully.

The sanitizer build doesn't print out anything interesting, but SVN exits immediately with "Corrupt MCB chain" error, and if I Ctrl-C the debug build of Staging, it looks as if it was stuck in an infinite loop in `DOS_MCB` or something.

```
Loguru caught a signal: SIGINT
Stack trace:
21         0x10c4a108c 25  dyld                                0x000000010c4a108c start + 520
20         0x104df389c main + 36
19         0x106292378 sdl_main(int, char**) + 18036
18         0x10507cc14 Config::StartUp() + 168
17         0x10746ec00 SHELL_Init() + 7152
16         0x107465888 DOS_Shell::Run() + 8104
15         0x10746351c DOS_Shell::RunBatchFile() + 2920
14         0x107461f64 DOS_Shell::ParseLine(char*) + 6488
13         0x1074cdcac DOS_Shell::DoCommand(char*) + 3348
12         0x10754c4ec DOS_Shell::ExecuteProgram(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 4680
11         0x1075502dc run_binary_executable(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 7376
10         0x1053741d8 CALLBACK_RunRealInt(unsigned char) + 412
9          0x104df5fac DOSBOX_RunMachine() + 28
8          0x104df5e94 Normal_Loop() + 596
7          0x10597185c DOS_21Handler() + 48020
6          0x105a49a8c DOS_ResizeMemory(unsigned short, unsigned short*) + 632
5          0x105a4776c DOS_CompressMemory() + 528
4          0x105a46d48 DOS_MCB::GetType() const + 276
3          0x106a1b488 mem_readb(unsigned int) + 24
2          0x106a1b27c mem_readb_inline(unsigned int) + 316
1          0x106a26c0c get_tlb_readhandler(unsigned int) + 208
2023-09-10 13:16:32.793 | Signal: SIGINT
```

weirddan455:
# Description

I spent a lot of time researching this.  This comment was very accurate:  https://github.com/dosbox-staging/dosbox-staging/issues/2851#issuecomment-1712875708

## Failed attempt to relocate video memory

I originally tried what he suggested by relocating the video memory from it's default location at 96KB linear memory to 16KB, similar to what a real PCjr memory manager would do.  This worked fine for some games.  Others like Thexder seemingly have the video memory address hard-coded and displayed garbage.

Additionally, this method requires changing a BIOS INT10 function to change the memory page the video card uses.  This would then need a somewhat hacky workaround to revert that change for stand-alone bootable games (that use `boot a:` to launch outside of DOS).  I am kind of curious how the real PCjr memory managers intercepted this BIOS interrupts but that's out of the scope of this PR.

## Successful work-around

What ended up working best from a bunch of testing is simply reserving the entire block up to 144KB.  This leaves the video memory at its standard location while also leaving a large chunk of high memory for application use.

This isn't exactly what real pcjr memory managers do but it's closer than what we had before.  Previously, we had a small free block under the video memory and then a large free block above it.  On real early DOS that pcjr's shipped with, the high memory (above the video memory ending at 128KB) would be invisible since early DOS required contiguous memory.

## Related issues

Fixes #2851

# Manual testing

Tested the following games and they work fine with this PR:

- Space Quest 1.0x (hangs at startup on main)
- Space Quest 2.2 (hangs the entire emulator on main)
- King's Quest 1986 pcjr DOS version (hangs the entire emulator on main)
- King's Quest 1984 pcjr bootable version (works on main)
- Thexder (works on main)
- Impossible Mission 2 (no pcjr support but a test for CGA compatibility) - works on main

# Checklist

I have:

### DIFF
--- a/include/dos_memory.h
+++ b/include/dos_memory.h
@@ -1,0 +1,27 @@
+/*
+ *  Copyright (C) 2024-2024  The DOSBox Staging Team
+ *  Copyright (C) 2002-2021  The DOSBox Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOS_MEMORY_H
+#define DOS_MEMORY_H
+
+constexpr auto ConventionalMemorySizeKb = 640;
+constexpr auto PcjrVideoMemorySizeKb    = 32;
+constexpr auto PcjrStandardMemorySizeKb = 128;
+
+#endif
--- a/src/dos/dos_memory.cpp
+++ b/src/dos/dos_memory.cpp
@@ -18,7 +18,9 @@
 
 #include "dosbox.h"
 
+#include "control.h"
 #include "dos_inc.h"
+#include "dos_memory.h"
 #include "mem.h"
 #include "support.h"
 
@@ -468,6 +470,12 @@ bool DOS_LinkUMBsToMemChain(uint16_t linkstate) {
 	return true;
 }
 
+static constexpr uint16_t kilobytes_to_segment(size_t address_kb)
+{
+	constexpr auto bytes_per_segment = 16;
+	constexpr auto bytes_per_kilobyte = 1024;
+	return (address_kb * bytes_per_kilobyte) / bytes_per_segment;
+}
 
 void DOS_SetupMemory(void) {
 	/* Let dos claim a few bios interrupts. Makes DOSBox more compatible with 
@@ -504,33 +512,65 @@ void DOS_SetupMemory(void) {
 	mcb_sizes+=17;
 	tempmcb2.SetType(middle_mcb_type);
 
-	DOS_MCB mcb((uint16_t)DOS_MEM_START+mcb_sizes);
-	mcb.SetPSPSeg(MCB_FREE);						//Free
-	mcb.SetType(ending_mcb_type); // Last Block
 	if (machine==MCH_TANDY) {
 		/* memory up to 608k available, the rest (to 640k) is used by
 			the tandy graphics system's variable mapping of 0xb800 */
-		mcb.SetSize(0x9BFF - DOS_MEM_START - mcb_sizes);
+		DOS_MCB free_block((uint16_t)DOS_MEM_START+mcb_sizes);
+		free_block.SetPSPSeg(MCB_FREE);
+		free_block.SetType(ending_mcb_type);
+		free_block.SetSize(0x9BFF - DOS_MEM_START - mcb_sizes);
 	} else if (machine==MCH_PCJR) {
-		/* memory from 128k to 640k is available */
-		mcb_devicedummy.SetPt((uint16_t)0x2000);
-		mcb_devicedummy.SetPSPSeg(MCB_FREE);
-		mcb_devicedummy.SetSize(umb_start_seg - 0x2000);
-		mcb_devicedummy.SetType(ending_mcb_type);
+		const auto pcjr_start = DOS_MEM_START + mcb_sizes;
+		constexpr auto mcb_entry_size = 1;
 
-		/* exclude PCJr graphics region */
-		mcb_devicedummy.SetPt((uint16_t)0x17ff);
-		mcb_devicedummy.SetPSPSeg(MCB_DOS);
-		mcb_devicedummy.SetSize(0x800);
-		mcb_devicedummy.SetType(middle_mcb_type);
+		// PCjr video memory uses 32KB shared RAM
+		constexpr auto video_memory_start =
+			kilobytes_to_segment(PcjrStandardMemorySizeKb - PcjrVideoMemorySizeKb);
 
-		/* memory below 96k */
-		mcb.SetSize(0x1800 - DOS_MEM_START - (2+mcb_sizes));
-		mcb.SetType(middle_mcb_type);
+		const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dos"));
+		assert(section);
+		const std::string pcjr_memory_config = section->Get_string("pcjr_memory_config");
+		if (pcjr_memory_config == "expanded") {
+			// With expanded memory, reserve the lower memory up to video memory.
+			// This makes application memory contiguous in order to prevent crashes.
+			// This is needed to prevent Sierra AGI games from crashing.
+			// Further details:
+			// https://www.atarimagazines.com/compute/issue58/pcjr_memory.html
+
+			// Space Quest version 1.0x is a special case.
+			// It requires an additional 16 KB above the 32KB video memory to be reserved.
+			constexpr auto application_segment =
+				video_memory_start + kilobytes_to_segment(PcjrVideoMemorySizeKb + 16);
+
+			// The size from the MCB entry itself must be subtracted from the total size.
+			const auto reserved_size = application_segment - pcjr_start - mcb_entry_size;
+			constexpr auto application_size = umb_start_seg - application_segment - mcb_entry_size;
+
+			DOS_MCB reserved(pcjr_start);
+			reserved.SetPSPSeg(MCB_DOS);
+			reserved.SetSize(reserved_size);
+			reserved.SetType(middle_mcb_type);
+
+			DOS_MCB free_block(application_segment);
+			free_block.SetPSPSeg(MCB_FREE);
+			free_block.SetSize(application_size);
+			free_block.SetType(ending_mcb_type);
+		} else {
+			// No expanded memory means the lower 96KB is usable for applications
+			assert(pcjr_memory_config == "standard");
+
+			DOS_MCB free_block(pcjr_start);
+			free_block.SetPSPSeg(MCB_FREE);
+			free_block.SetSize(video_memory_start - pcjr_start - mcb_entry_size);
+			free_block.SetType(ending_mcb_type);
+		}
 	} else {
 		/* complete memory up to 640k available */
 		/* last paragraph used to add UMB chain to low-memory MCB chain */
-		mcb.SetSize(0x9FFE - DOS_MEM_START - mcb_sizes);
+		DOS_MCB free_block((uint16_t)DOS_MEM_START+mcb_sizes);
+		free_block.SetPSPSeg(MCB_FREE);
+		free_block.SetType(ending_mcb_type);
+		free_block.SetSize(0x9FFE - DOS_MEM_START - mcb_sizes);
 	}
 
 	dos.firstMCB=DOS_MEM_START;
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1229,6 +1229,17 @@ void DOSBOX_Init()
 	        "tab-separated format, used by SETVER.EXE as a persistent storage\n"
 	        "(empty by default).");
 
+	const char* pcjr_memory_configurations[] = {"expanded", "standard", nullptr};
+	pstring = secprop->Add_string("pcjr_memory_config", only_at_start, "expanded");
+	pstring->Set_values(pcjr_memory_configurations);
+	pstring->Set_help(
+		"PCjr memory layout ('expanded' by default).\n"
+		"  expanded:  640 KB total memory with applications residing above 128 KB.\n"
+		"             Compatible with most games.\n"
+		"  standard:  128 KB total memory with applications residing below 96 KB.\n"
+		"             Required for some older games (e.g., Jumpman, Troll)."
+	);
+
 	// Mscdex
 	secprop->AddInitFunction(&MSCDEX_Init);
 	secprop->AddInitFunction(&DRIVES_Init);
--- a/src/ints/bios.cpp
+++ b/src/ints/bios.cpp
@@ -20,8 +20,10 @@
 #include "bios.h"
 
 #include "callback.h"
+#include "control.h"
 #include "cpu.h"
 #include "dosbox.h"
+#include "dos_memory.h"
 #include "hardware.h"
 #include "inout.h"
 #include "joystick.h"
@@ -1211,14 +1213,27 @@ public:
 		/* INT 12 Memory Size default at 640 kb */
 		callback[2].Install(&INT12_Handler,CB_IRET,"Int 12 Memory");
 		callback[2].Set_RealVec(0x12);
-		if (IS_TANDY_ARCH) {
+		if (machine == MCH_TANDY) {
 			/* reduce reported memory size for the Tandy (32k graphics memory
 			   at the end of the conventional 640k) */
-			if (machine==MCH_TANDY) mem_writew(BIOS_MEMORY_SIZE,624);
-			else mem_writew(BIOS_MEMORY_SIZE,640);
-			mem_writew(BIOS_TRUE_MEMORY_SIZE,640);
-		} else mem_writew(BIOS_MEMORY_SIZE,640);
-		
+			mem_writew(BIOS_MEMORY_SIZE, 624);
+			mem_writew(BIOS_TRUE_MEMORY_SIZE, ConventionalMemorySizeKb);
+		} else if (machine == MCH_PCJR) {
+			const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dos"));
+			assert(section);
+			const std::string pcjr_memory_config = section->Get_string("pcjr_memory_config");
+			if (pcjr_memory_config == "expanded") {
+				mem_writew(BIOS_MEMORY_SIZE, ConventionalMemorySizeKb);
+				mem_writew(BIOS_TRUE_MEMORY_SIZE, ConventionalMemorySizeKb);
+			} else {
+				assert(pcjr_memory_config == "standard");
+				mem_writew(BIOS_MEMORY_SIZE, PcjrStandardMemorySizeKb - PcjrVideoMemorySizeKb);
+				mem_writew(BIOS_TRUE_MEMORY_SIZE, PcjrStandardMemorySizeKb);
+			}
+		} else {
+			mem_writew(BIOS_MEMORY_SIZE, ConventionalMemorySizeKb);
+		}
+
 		/* INT 13 Bios Disk Support */
 		BIOS_SetupDisks();
 


src/dos/dos_memory.cpp
```diff
@@ -504,33 +504,45 @@ void DOS_SetupMemory(void) {
...
+		// PCjr video memory uses shared RAM at linear address 96KB - 128KB
+		// This video memory is mapped in 16KB pages
+		// Sierra AGI games crash if the application memory is not contiguous
+		// Work around that by reserving everything up until the application segment
+		const uint16_t reserved_segment = DOS_MEM_START + mcb_sizes;
+
+		// Segment at which application memory starts.
+		// (0x2400 * 16 bytes per segment) / 1024 bytes per kilobyte = 144KB linear address
+		// Space Quest version 1.0x is a special case that requires an additional 16 KB page to be reserved
+		const uint16_t application_segment = 0x2400;
```
johnnovak, line 524: How about something like this to turn the comment into actual code?

```cpp
constexpr auto bytes_per_segment = 16;
constexpr auto application_segment = 144 /* kilobytes */ * 1024 / bytes_per_segment
```

johnnovak: Does this mean the video mem is usually 32KB in total on the PCjr, from 96 to 128 KB, but because SQ v1.0 requires an extra 16 KB of video memory, that pushed our application segment start offset from 144KB from 128KB? I had to guess this, so I think the comments should be a bit clearer.

weirddan455: Yep, that's basically the situation.  SQ 1.0x is a strange case because the video memory of PCjr cannot extend past 128KB.  As far as I can tell, that's a hardware limitation and the code in `vga_other.cpp` seems to replicate that.

I'm not sure what exactly it's doing with that extra 16KB.  Maybe using it for a backbuffer or something.  I just kept running into what looked like memory corruption bugs until I raised it.  The video output looked fine but there were crashes and disk errors and such.

I'll clarify like you suggested.

johnnovak:
So, this improves the situation for a few games, but breaks compatibility with others that assume an unexpanded PCjr with only 128 KB of RAM. One such example is Jumpman which works fine with `main`:

image

I've found some interesting details on the PCjr implementation and PCjr memory management; start reading from this comment:

https://www.vogons.org/viewtopic.php?p=79029#p79029

It seems to me that no "one size fits all" solution exists, we need to make this configurable.

weirddan455:
>  breaks compatibility with others that assume an unexpanded PCjr with only 128 KB of RAM

Hmm... we weren't really handling that case correctly before either.  We shouldn't be setting MCB entries in high memory if we wanted to emulate that.  Plus we should have the BIOS report 112/96KB (depending on how large we want the video buffer).

>  It seems to me that no "one size fits all" solution exists, we need to make this configurable.

I don't want to make more options than we actually need.  Or at least more than I can actually test and find a use for.  But yeah I guess we do need some level of configuration here.  I'll look at it a bit more later.

johnnovak:
> I don't want to make more options than we actually need.

I agree with that, it would be ideal if we could get away with a single hardcoded setting. There's no point in adding all the PCjr memory configuration options just for the sake of it, no doubt. I'm just not sure if a single fixed setting is feasible, but I guess you'll find out 😄

weirddan455:
I added a configuration option leaving the behavior described at the top as the default.  Everything I tested except for Jumpman and a game called Troll works in this configuration.

You can also set `pcjr_expanded_memory = false` to limit to 128KB and map applications in the lower address space.  Jumpman and Troll work with this set but a lot of games need more than 128KB and fail.

I also removed the magic hex numbers to hopefully make it clear in code where the addresses come from.

src/dos/dos_memory.cpp
```diff
@@ -468,6 +469,14 @@ bool DOS_LinkUMBsToMemChain(uint16_t linkstate) {
 	return true;
 }
 
+static constexpr uint16_t kilobytes_to_segment(size_t address)
+{
+	constexpr auto bytes_per_segment = 16;
+	constexpr auto bytes_per_killobyte = 1024;
```
johnnovak, line 475: typo: kilobyte (one L)

src/dos/dos_memory.cpp
```diff
@@ -468,6 +469,14 @@ bool DOS_LinkUMBsToMemChain(uint16_t linkstate) {
 	return true;
 }
 
+static constexpr uint16_t kilobytes_to_segment(size_t address)
```
johnnovak, line 472: Suggest to rename to `address_kb`.

src/dos/dos_memory.cpp
```diff
@@ -468,6 +469,14 @@ bool DOS_LinkUMBsToMemChain(uint16_t linkstate) {
...
+{
+	constexpr auto bytes_per_segment = 16;
+	constexpr auto bytes_per_killobyte = 1024;
+	assert(address < 64 * bytes_per_killobyte);
+	assert(address % bytes_per_segment == 0);
```
johnnovak, line 477: These two asserts seem to be some leftover from an earlier implementation. They're not needed and don't make any sense.

src/dos/dos_memory.cpp
```diff
@@ -504,33 +513,56 @@ void DOS_SetupMemory(void) {
...
+
+		const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dosbox"));
+		assert(section);
+		if (section->Get_bool("pcjr_expanded_memory")) {
+			// With expanded memory. reserve the lower memory up to video memory
+			// This makes application memory contiguous in order to prevent crashes
+			// This is needed to prevent Sierra AGI games from crashing
+
+			// Space Quest version 1.0x is a special case that requires an additional 16 KB above the 32KB video memory to be reserved.
+			constexpr auto application_segment = video_memory_start + kilobytes_to_segment(48);
```
johnnovak, line 537: Too long lines; just run this whole block you've added through clang-format.

src/dos/dos_memory.cpp
```diff
@@ -504,33 +513,56 @@ void DOS_SetupMemory(void) {
...
+
+		// PCjr video memory uses shared RAM at linear address 96KB - 128KB
+		constexpr auto video_memory_start = kilobytes_to_segment(96);
+
+		const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dosbox"));
+		assert(section);
+		if (section->Get_bool("pcjr_expanded_memory")) {
+			// With expanded memory. reserve the lower memory up to video memory
+			// This makes application memory contiguous in order to prevent crashes
+			// This is needed to prevent Sierra AGI games from crashing
```
johnnovak, line 534: Minor: fix grammar (comma after "with expanded memory", periods at the ends of the sentences).

src/dos/dos_memory.cpp
```diff
@@ -504,33 +513,56 @@ void DOS_SetupMemory(void) {
...
-		mcb_devicedummy.SetPSPSeg(MCB_DOS);
-		mcb_devicedummy.SetSize(0x800);
-		mcb_devicedummy.SetType(middle_mcb_type);
-
-		/* memory below 96k */
-		mcb.SetSize(0x1800 - DOS_MEM_START - (2+mcb_sizes));
-		mcb.SetType(middle_mcb_type);
+		const auto pcjr_start = DOS_MEM_START + mcb_sizes;
+
+		// PCjr video memory uses shared RAM at linear address 96KB - 128KB
```
johnnovak, line 526: Maybe say "PCjr video memory uses shared 32KB of RAM..."

src/dos/dos_memory.cpp
johnnovak, line 537: Minor: `kilobytes_to_segment(32 + 16)` would be more descriptive.

src/dos/dos_memory.cpp
```diff
@@ -504,33 +513,56 @@ void DOS_SetupMemory(void) {
...
+		assert(section);
+		if (section->Get_bool("pcjr_expanded_memory")) {
+			// With expanded memory. reserve the lower memory up to video memory
+			// This makes application memory contiguous in order to prevent crashes
+			// This is needed to prevent Sierra AGI games from crashing
+
+			// Space Quest version 1.0x is a special case that requires an additional 16 KB above the 32KB video memory to be reserved.
+			constexpr auto application_segment = video_memory_start + kilobytes_to_segment(48);
+
+			// 1 must be subtracted from the size to account for the MCB entry itself
```
johnnovak, line 539: How about this:

```cpp
// The size from the MCB entry itself must be subtracted from the total size
constexpr auto mcb_entry_size = 1;
const auto reserved_size = application_segment - pcjr_start - mcb_entry_size;
```

Then rewrite the all `- 1` occurrences similarly.

src/dosbox.cpp
```diff
@@ -520,6 +520,15 @@ void DOSBOX_Init()
...
 
+	pbool = secprop->Add_bool("pcjr_expanded_memory", only_at_start, true);
+	pbool->Set_help(
+		"Use PCjr expanded memory layout\n"
+		"  true:  640KB total memory with applications residing above 128KB.\n"
+		"         Compatible with most games. Required if a game needs more than 128KB.\n"
+		"  false: 128KB total memory with applications residing below 96KB\n"
+		"         Required for some older games which aren't compatible with PCjr memory upgrades"
+	);
+
```
johnnovak, line 531: Nice description 🎉 A couple of suggestions and comments:

- `true` and `false` for toggling options is not very user-friendly for non-coders. So I introduced an enhancement for handling booleans a while ago: `on`, `enabled`, `yes`, `true` or `1` can be used for turning a boolean setting on, and `off`, `disabled`, `no`, `false` or `0` for turning it off.
- Consequently, I've eradicated the use of `true` and `false` from the config descriptions to use `on` and `off` instead.
- We always mention the default value on the first line, so this should be `Use PCjr expanded memory layout (enabled by default).`
- Space before `KB` (so `640KB` should become `640 KB`, etc.)
- I'd mention Jumpan and Troll for the standard memory layout option (just chuck `(e.g., Jumpan, Troll).` at the end).
- Periods at the ends of the bullet points (for consistency with the rest of the config descriptions).
- The description text needs to be wrapped to 79 characters so it looks nicely indented when displayed with `config -h <settingname>`. The last line goes above that limit:

image

---

...but with all said, I'd like to suggest not to go with a simple boolean value, but use an enum instead.

I 100% agree that we should keep the number of PCjr memory layout options to the minimum, but using a boolean paints us into a corner. If we need to add a third option in the future, we'd need to rename and deprecate the setting.

A slightly different approach open to potential future extension would be to call this `pcjr_memory_config` instead with only two options for now: `standard` and `expanded`, with the latter being the default.

src/ints/bios.cpp
```diff
@@ -1211,14 +1212,25 @@ class BIOS final : public Module_base{
 		/* INT 12 Memory Size default at 640 kb */
 		callback[2].Install(&INT12_Handler,CB_IRET,"Int 12 Memory");
 		callback[2].Set_RealVec(0x12);
-		if (IS_TANDY_ARCH) {
+		if (machine == MCH_TANDY) {
```
johnnovak, line 1215: Nice! I too dislike the macro hackery...

src/ints/bios.cpp
```diff
@@ -1211,14 +1212,25 @@ class BIOS final : public Module_base{
...
+			mem_writew(BIOS_MEMORY_SIZE, 624);
+			mem_writew(BIOS_TRUE_MEMORY_SIZE, 640);
+		} else if (machine == MCH_PCJR) {
+			const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dosbox"));
+			assert(section);
+			if (section->Get_bool("pcjr_expanded_memory")) {
+				mem_writew(BIOS_MEMORY_SIZE, 640);
+				mem_writew(BIOS_TRUE_MEMORY_SIZE, 640);
+			} else {
+				mem_writew(BIOS_MEMORY_SIZE, 96);
```
johnnovak, line 1227: Would it make sense to introduce a `dos_memory.h` file and reference these memory size constants in `bios.cpp` and `dos_memory.cpp` instead of hardcoding and duplicating them?

Example for a few such constants (there are more):

### dos_memory.h

```cpp
constexpr auto PcjrVideoMemorySizeKb    = 32;
constexpr auto PcjrStandardMemorySizeKb = 128;
```

### dos_memory.cpp

```cpp
constexpr auto video_memory_start = kilobytes_to_segment(PcjrStandardMemorySizeKb - PcjrVideoMemorySizeKb); // instead of hardcoding 96
```

### bios.cpp

```cpp
mem_writew(BIOS_MEMORY_SIZE, PcjrStandardMemorySizeKb - PcjrVideoMemorySizeKb); // instead of hardcoding 96
mem_writew(BIOS_TRUE_MEMORY_SIZE, PcjrStandardMemorySizeKb); // instead of hardcoding 128
```

johnnovak:
Nice job @weirddan455 🎉 

Another question: is the `[dosbox]` section the best place for this config? I'm just wondering because the memory configuration related `xms`, `ems` and `umb` settings are in the `[dos]` section.

I've tested a bunch of games; these work well with the new expanded memory config:

- Ancient Art of War, The
- Below the Root
- Black Cauldron, The (v2.01)
- Blockout
- Bruce Lee
- California Games
- Gold Rush (v2.01)
- Manhunter 1 - New York
- Maniac Mansion
- Maniac Mansion Enhanced
- Space Quest 1
- Total Eclipse
- Wizard and the Princess, The
- Zak McKracken and the Alien Mindbenders
- Zak McKracken and the Alien Mindbenders Enhance

I've found only a single game that needs the standard memory layout. This is actually a PC Booter game from 1982:

- Adventure in Serenia

src/dos/dos_memory.cpp
johnnovak: I would also put a link to this nice article you found here: 

```
// Further details:
// https://www.atarimagazines.com/compute/issue58/pcjr_memory.html
```

src/dos/dos_memory.cpp
weirddan455: fixed

src/dos/dos_memory.cpp
weirddan455: fixed

src/dos/dos_memory.cpp
weirddan455: Yep, I changed what I was doing here and forgot to remove these.  Gone now.

src/dos/dos_memory.cpp
weirddan455: `scripts/format-commit.sh` butchered this code.  It added 2 extra layers of indention for seemingly no reason.  I trimmed these lines manually.

src/dos/dos_memory.cpp
weirddan455: Fixed

src/dos/dos_memory.cpp
weirddan455: fixed

src/dos/dos_memory.cpp
weirddan455: fixed

src/dos/dos_memory.cpp
weirddan455: fixed

src/dosbox.cpp
weirddan455: Yeah, good call on making it a string.  I wasn't thinking about keeping the config backwards compatible if we need to add another option later.  Changed + fixed up the help text.

src/ints/bios.cpp
weirddan455: `IS_TANDY_ARCH` actually means `machine == MCH_TANDY || machine == MCH_PCJR`.  I did this because I was adding further distinctions between the two but it's maybe badly named.

src/ints/bios.cpp
weirddan455: Added the header with the constants like you suggested.

src/dosbox.cpp
johnnovak: > Yeah, good call on making it a string. I wasn't thinking about keeping the config backwards compatible if we need to add another option later. Changed + fixed up the help text.

No problem. I've been doing the rename/deprecation dance of config settings for a while now, so I'm sensitive to it and prefer to avoid it if possible 😄

include/dos_memory.h
```diff
@@ -0,0 +1,27 @@
+/*
+ *  Copyright (C) 2020-2024  The DOSBox Staging Team
```
johnnovak, line 2: 2024-2024

src/dos/dos_memory.cpp
```diff
@@ -504,33 +512,65 @@ void DOS_SetupMemory(void) {
...
-		mcb.SetSize(0x1800 - DOS_MEM_START - (2+mcb_sizes));
-		mcb.SetType(middle_mcb_type);
+		const auto pcjr_start = DOS_MEM_START + mcb_sizes;
+		constexpr auto mcb_entry_size = 1;
+
+		// PCjr video memory uses 32KB shared RAM
+		constexpr auto video_memory_start =
+			kilobytes_to_segment(PcjrStandardMemorySizeKb - PcjrVideoMemorySizeKb);
+
+		const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dos"));
```
johnnovak, line 530: So you've moved it to `[dos]`, that's better 👍🏻

src/dosbox.cpp
```diff
@@ -1229,6 +1229,17 @@ void DOSBOX_Init()
 	        "tab-separated format, used by SETVER.EXE as a persistent storage\n"
 	        "(empty by default).");
 
+	const char* pcjr_memory_configurations[] = {"expanded", "standard", nullptr};
+	pstring = secprop->Add_string("pcjr_memory_config", only_at_start, "expanded");
+	pstring->Set_values(pcjr_memory_configurations);
+	pstring->Set_help(
+		"PCjr memory layout (expanded by default)\n"
```
johnnovak, line 1236: `('expanded' by default)` (string settings are always single-quoted).

src/dosbox.cpp
```diff
@@ -1229,6 +1229,17 @@ void DOSBOX_Init()
...
+		"  expanded: 640KB total memory with applications residing above 128 KB.\n"
```
johnnovak, line 1237: `640 KB`, two spaces after `expanded:`.

src/dosbox.cpp
```diff
@@ -1229,6 +1229,17 @@ void DOSBOX_Init()
...
+		"            Compatible with most games.\n"
+		"  standard: 128KB total memory with applications residing below 96 KB.\n"
```
johnnovak, line 1239: `128 KB`, two spaces after `standard:`.

src/dosbox.cpp
```diff
@@ -1229,6 +1229,17 @@ void DOSBOX_Init()
...
+		"            Required for some older games (e.g., Jumpan, Troll)."
```
johnnovak, line 1240: typo: Jumpman

src/ints/bios.cpp
```diff
@@ -1211,14 +1213,27 @@ class BIOS final : public Module_base{
...
-			mem_writew(BIOS_TRUE_MEMORY_SIZE,640);
-		} else mem_writew(BIOS_MEMORY_SIZE,640);
-		
+			mem_writew(BIOS_MEMORY_SIZE, 624);
+			mem_writew(BIOS_TRUE_MEMORY_SIZE, ConventionalMemorySizeKb);
+		} else if (machine == MCH_PCJR) {
+			const Section_prop* section = static_cast<Section_prop*>(control->GetSection("dos"));
+			assert(section);
+			const std::string pcjr_memory_config = section->Get_string("pcjr_memory_config");
+			if (pcjr_memory_config == "expanded") {
```
johnnovak, line 1225: It's unfortunate that both `dos_memory.cpp` and `bios.cpp` need to have knowledge about the setting strings.

A better solution would be that `dos_memory.cpp` does the parsing on module init, then converts it to an enum:

```cpp
enum class PcjrMemoryLayout { Standard, Expanded };
```

`bios.cpp` then could just ask for the memory layout with the public `DOS_GetPcjrMemoryLayout()` function.

I started using this more manageable approach in new code (e.g., `render.cpp`, `shader_manager.cpp`, etc.)

johnnovak:
Okay @weirddan455 , this is coming together very nicely, great job 🎉 I've made one last round of minor comments (no more rounds, I promise 😅), then it's good to go!

src/ints/bios.cpp
weirddan455: That doesn't work.  BIOS init runs before DOS init does.  The config section that runs `BIOS_Init` is `[joystick]` for some reason and putting it there doesn't make much sense.

I get that it's nicer to work with enums but the config value is already parsed by the time we get here and in one of two valid states.  Introducing a global that depends on which module gets init'd first has been a major problem for us so I'd rather avoid that.

include/dos_memory.h
weirddan455: fixed

src/dosbox.cpp
weirddan455: fixed

src/dosbox.cpp
weirddan455: fixed

src/dosbox.cpp
weirddan455: fixed

src/dosbox.cpp
weirddan455: fixed

Jkapp76:
I hope it will be possible to change from expanded to standard during emulation. I include BAT files with most of these old games to adjust the settings. I like to restore my default settings after exiting the game. So far we can do this with all the settings.

weirddan455:
> I hope it will be possible to change from expanded to standard during emulation. I include BAT files with most of these old games to adjust the settings. I like to restore my default settings after exiting the game. So far we can do this with all the settings.

Not all settings are configurable at runtime.  Changing the memory layout at runtime is not possible without significant hacks to the DOS memory allocator.  Honestly it might not be feasible at all to do this.  It would be like removing sticks of RAM from your computer while it is running.

johnnovak:
> > I hope it will be possible to change from expanded to standard during emulation. I include BAT files with most of these old games to adjust the settings. I like to restore my default settings after exiting the game. So far we can do this with all the settings.
> 
> Not all settings are configurable at runtime. Changing the memory layout at runtime is not possible without significant hacks to the DOS memory allocator. Honestly it might not be feasible at all to do this. It would be like removing sticks of RAM from your computer while it is running.

Agreed. It's out of scope @Jkapp76. Although we have limited support for reconfiguring the hardware at runtime (the best example is adding/removing/changing audio cards), we don't support that for the graphics adapter and memory settings. It's out of scope, too hard and too complex for marginal benefits. I think in some cases it might be even impossible to do it correctly.

src/ints/bios.cpp
johnnovak: Ok, let's leave it for now. When we refactor the config stuff, such enum conversions should happen before anything else is initialised anyway, so this would be just a stopgap measure until then.

johnnovak (approved):
Very nice work improving our PCjr compatibility @weirddan455 🥇

johnnovak:
Thinking of backporting this too as it improved game compatibility if no objections @weirddan455 ?

weirddan455:
Agreed.  Seems like a good backport as it fixes some popular Sierra games.

Issue #2565, opened 2023-06-01
Changing machine type or mem settings with config command will take effect on the next restart with config -r


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

None

### Is your feature request related to a problem? Please describe.

Changing some config setting isn't possible during runtime.

### Describe the solution you'd like

@johnnovak said

> config -r restarts the emulator, so I thought settings not changeable at runtime could be changed that way, but no. That would be a cool feature, e.g. changing the machine type and the mem settings for example with the config command, and instead of raising an error, a message would be displayed saying these are pending changes that will take effect on the next restart with config -r. Basically, we would need to hold one more extra config layer in memory for the pending settings that would be applied on top of all the loaded configs after the restart.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

PR #3470, opened 2024-03-04
Throw error if user attempts to change an unchangeable setting

### LINKED ISSUE #1027: Setting a non-runtime config item should produce a warning
**Describe the bug**
In the late 90's I did some toying around with various demoscene effect. One of these effects was a copper bar. I made several small executables that displayed one or more horizontal color bars moving in a smooth up/down motion.

I coded these on (I think) a 486 with a Trident VGA card, and I'm certain they worked on that machine on pure DOS.

However, they do not display anything at all on DosBox-staging.

**To Reproduce**
Run the attached executables in DosBox

**Expected behaviour**
some fancy colored bars should appear, and maybe move up or down the screen (depending on the exe file... I don't remember what effect each EXE file does, it's been over 20 years by now...)

**Environment (please complete the following information):**
- Windows 7
- x86_64

**Version of dosbox-staging:**
- dosbox-staging version 0.77.0-alpha-760-gdc029

**Configuration parameters:**
Basically I tried svga_se, vgaonly, svga_paradise and some others. I also tried the `vertical retrace poll debug line` parameter but it didn't show anything

**Additional context**
I'm not sure my code is (technically) correct... but it did work on real hardware so I'm not sure if this qualifies as a bug or not...

The programs are pretty small and straightforward. I will try and see if I can find the source code, but basically it's more or less waiting for bit 3 of 0x3da, writing to 0x3c8, and waiting for bit 1 of 0x3da.
[copper.zip](https://github.com/dosbox-staging/dosbox-staging/files/6528684/copper.zip)

### LINKED ISSUE #3080: Better handling for config settings that can/cannot be changed at runtime

latest main

Linux x86_64

Arch Linux

I was looking at issue #2527 where setting `fullscreen=true` does not work at runtime.  Looking at the code, we have an enum that describes when a particular config setting is allowed to be changed.

```cpp
struct Changeable {
	enum Value { Always, WhenIdle, OnlyAtStart, Deprecated };
};

Pbool = sdl_sec->Add_bool("fullscreen", always, false);

pbool = sdl_sec->Add_bool("pause_when_inactive", on_start, false);
```

Here we can see the usage code of two values.  `fullscreen` is meant to always be changeable while `pause_when_inactive` is meant to not be changeable at runtime.

However, I can't find real enforcement of this nor any documentation to the user.  Ideally, typing in `pause_when_inactive=true` would present the user with an error indicating that this value cannot be changed.  However, as it currently stands, there is no error and the window actually flickers because it is running some re-init code.  We shouldn't be trying to refresh the window when the user wants to change an unchangeable value.

Furthermore, it actually seems to update the config setting to the new value even though the pause functionality is unchanged.

pause

Unchangeable values should report an error when the user tries to change them.  We should not be running init functions when this happens.  We should not be updating the value at all.

It would also be nice if this was documented somewhere.  Perhaps in the config file, we write out when this setting is allowed to be changed.

I don't know but I am open to suggestions.

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

The changeable when value on config settings was not actually being enforced anywhere.  This is a simple change to the `config -set` DOS program to check that value and throw an error if the setting is not meant to be changed at runtime.

## Related issues

Fixes #3080
Fixes #1027

# Manual testing

`pause_when_inactive=true` no longer changes the value or runs destroy + init functions.  It instead simply throws an error stating the setting is not allowed to be changed.

Tested with several other config settings as well.

# Checklist

I have:

### DIFF
--- a/src/misc/programs.cpp
+++ b/src/misc/programs.cpp
@@ -798,7 +798,20 @@ void CONFIG::Run(void)
 			if (strlen(result)) {
 				WriteOut(result);
 			} else {
-				Section* tsec = control->GetSection(pvars[0]);
+				auto* tsec = dynamic_cast<Section_prop *>(control->GetSection(pvars[0]));
+				if (!tsec) {
+					WriteOut(MSG_Get("PROGRAM_CONFIG_PROPERTY_ERROR"), pvars[0].c_str());
+					return;
+				}
+				const auto* property = tsec->Get_prop(pvars[1]);
+				if (!property) {
+					WriteOut(MSG_Get("PROGRAM_CONFIG_PROPERTY_ERROR"), pvars[1].c_str());
+					return;
+				}
+				if (property->GetChange() == Property::Changeable::OnlyAtStart) {
+					WriteOut(MSG_Get("PROGRAM_CONFIG_NOT_CHANGEABLE"), pvars[1].c_str());
+					return;
+				}
 				// Input has been parsed (pvar[0]=section,
 				// [1]=property, [2]=value) now execute
 				std::string value(pvars[2]);
@@ -1002,4 +1015,6 @@ void PROGRAMS_Init(Section* sec)
 	MSG_Add("PROGRAM_EXECUTABLE_MISSING", "Executable file not found: '%s'\n");
 
 	MSG_Add("CONJUNCTION_AND", "and");
+
+	MSG_Add("PROGRAM_CONFIG_NOT_CHANGEABLE", "Property '%s' is not changeable at runtime.");
 }
--- a/src/misc/setup.cpp
+++ b/src/misc/setup.cpp
@@ -867,6 +867,16 @@ Property* Section_prop::Get_prop(int index)
 	return nullptr;
 }
 
+Property* Section_prop::Get_prop(const std::string& propname)
+{
+	for (Property* property : properties) {
+		if (property->propname == propname) {
+			return property;
+		}
+	}
+	return nullptr;
+}
+
 std::string Section_prop::Get_string(const std::string& _propname) const
 {
 	for (const_it tel = properties.begin(); tel != properties.end(); ++tel) {
--- a/include/setup.h
+++ b/include/setup.h
@@ -397,6 +397,7 @@ public:
 	                                      const std::string& sep);
 
 	Property* Get_prop(int index);
+	Property* Get_prop(const std::string& propname);
 
 	int Get_int(const std::string& _propname) const;
 


rderooy:
Just thinking, perhaps if such a case DOSBox Staging should give the option to "reboot" the emulator with the changed setting.

weirddan455:
> Just thinking, perhaps if such a case DOSBox Staging should give the option to "reboot" the emulator with the changed setting.

Yeah there's another open issue #2565 requesting that feature.  I suppose it wouldn't be too hard to implement.  The restart feature literally restarts the entire process so it would just be a matter of launching the new process like `dosbox --set propery=value`.

src/misc/setup.cpp
```diff
@@ -867,6 +867,16 @@ Property* Section_prop::Get_prop(int index)
...
 }
 
+Property* Section_prop::Get_prop(const std::string& propname)
+{
+	for (Property* property : properties) {
+		if (property->propname == propname) {
+			return property;
+		}
+	}
+	return nullptr;
```
japsmits, line 877:
```suggestion
    const auto it = std::ranges::find_if(properties, [propname](const auto* prop) { return prop->propname == propname); };
    return it != properties.end() ? *it : nullptr;
```

src/misc/programs.cpp
```diff
@@ -798,7 +798,20 @@ void CONFIG::Run(void)
 			if (strlen(result)) {
 				WriteOut(result);
 			} else {
-				Section* tsec = control->GetSection(pvars[0]);
+				Section_prop* tsec = dynamic_cast<Section_prop *>(control->GetSection(pvars[0]));
```
japsmits, line 801:
```suggestion
				auto* tsec = dynamic_cast<Section_prop *>(control->GetSection(pvars[0]));
```

src/misc/programs.cpp
```diff
@@ -798,7 +798,20 @@ void CONFIG::Run(void)
...
+				if (!tsec) {
+					WriteOut(MSG_Get("PROGRAM_CONFIG_PROPERTY_ERROR"), pvars[0].c_str());
+					return;
+				}
+				Property* property = tsec->Get_prop(pvars[1]);
```
japsmits, line 806:
```suggestion
				const auto* property = tsec->Get_prop(pvars[1]);
```

src/misc/setup.cpp
weirddan455: I think the loop is much easier to read.

johnnovak (approved):
I like this change, nice job @weirddan455.

weirddan455:
I was just thinking through the details of how setting these on restart would work.  There's a couple ways we could go about it:

1. After the "cannot be changed at runtime" error, prompt the user if they would like to restart.  This may break some people's .bat files that aren't expecting an interactive prompt and also would be only changing one setting at a time.
2. As @johnnovak said in #2565 we could maintain a list of settings to change on next restart, either by `config - r` or by the `ctrl + alt + home` hotkey.  I don't think we would need the duplicate the entire config data structure, just a string with key + value pairs should suffice to append onto the command line arguments of the new process.
3. Just merge this PR as is.

Btw, you already kind of can use `config -r` to achieve this.  That command lets you specify command line arguments.  You can do, for example, `config -r --set machine=tandy`.

johnnovak:
> I was just thinking through the details of how setting these on restart would work. There's a couple ways we could go about it:
> 
>     1. After the "cannot be changed at runtime" error, prompt the user if they would like to restart.  This may break some people's .bat files that aren't expecting an interactive prompt and also would be only changing one setting at a time.
> 
>     2. As @johnnovak said in [Changing machine type or mem settings with config command will take effect on the next restart with config -r #2565](https://github.com/dosbox-staging/dosbox-staging/issues/2565) we could maintain a list of settings to change on next restart, either by `config - r` or by the `ctrl + alt + home` hotkey.  I don't think we would need the duplicate the entire config data structure, just a string with key + value pairs should suffice to append onto the command line arguments of the new process.
> 
>     3. Just merge this PR as is.
> 
> 
> Btw, you already kind of can use `config -r` to achieve this. That command lets you specify command line arguments. You can do, for example, `config -r --set machine=tandy`.

I think merge this PR as is, then option 2 could be done later. That's my idea of the OSD too; e.g., when you change the memory size or the video adapter type, the UI will inform you that setting will only be applied after a restart. But it won't restart the emulator every single time you click (or accidentally click) on such a setting... that would be terrible. The situation is a bit similar with option 1 from the CLI, but less pronounced.

Yeah, so I really dislike option 1.

src/misc/programs.cpp
weirddan455: Sure, changed to use `auto`

weirddan455:
Sounds like a plan @johnnovak 

I'll just merge this as is and we can add that feature in later.

src/misc/setup.cpp
johnnovak: I tend to agree; we don't get brownie points for using STL to the max. Let's optimise for readability.

johnnovak:
I think this is safe to backport too @weirddan455.

PR #3471, opened 2024-03-04
Make the Time Stamp Counter more realistic

FeralChild64:
# Description

This PR contains a more realistic implementation of the _Time Stamp Counter_, found on Pentium and later CPUs, often used to determine the CPU clock. It works as follows:
- if `cycles` is set to `max`, the TSC simulates the clock speed of a real CPU (similarly to previous implementation, but uses different values for _Pentium_ and _Pentium MMX_) - it claims this is a fast CPU of the emulated generation
- if absolute value of `cycles` is given, it tries to estimate the CPU speed in MHz according to tables from the DOSBox-X wiki (https://dosbox-x.com/wiki/Guide%3ACPU-settings-in-DOSBox%E2%80%90X#_cycles)
- cycles settings like `max 75% limit <cycles>` will now use the specified `<cycles>` value for MHz estimation (previous implementation was simulating a default 70 MHz clock)

(note: screenshot below is outdated, the calculations have changed)

Screenshot_Annotated

# Manual testing

- set the `cpu` config option to `pentium` or `pentium_mmx`
- start the HWINFO portable for DOS (https://www.hwinfo.com/download/)
- from the menu select _Info_ -> _Mainboard info_.
- retry with different `cycles` settings

# Checklist

I have:


src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+// Round the value in MHz to one of the values typical for a real CPU
+static double round_cpu_mhz(const double cpu_mhz)
+{
+	const auto iter_upper = std::upper_bound(typical_mhz_values.begin(),
+	                                         typical_mhz_values.end(),
+	                                         cpu_mhz);
+
+	if (iter_upper == typical_mhz_values.end()) {
+		constexpr double resolution = 100.0f;
```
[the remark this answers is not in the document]
johnnovak: Should be just `100.0`.

src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+static const CyclesMhzMap Intel86_Map = {{240, 4.77}};
+
+static const CyclesMhzMap Intel286_Map = {{750, 8}, {1510, 12}, {3300, 25}};
+
+static const CyclesMhzMap Intel386_Map = {{4595, 25}, {6075, 33}};
+
+static const CyclesMhzMap Intel486_Map = {{12019, 33},
+                                          {23880, 66},
+                                          {33445, 100},
+                                          {47810, 133}};
```
johnnovak, line 75: Well, the class is about TSC, which is Pentium only, so the pre-Pentium options need to go.

src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+void CPU_ReadTSC()
+{
+	const auto cpu_mhz = get_estimated_cpu_mhz();
+
+	const auto tsc_precise = PIC_FullIndex() * cpu_mhz * 1000;
+	const auto tsc_rounded = static_cast<int64_t>(std::round(tsc_precise));
+
+	reg_edx = static_cast<uint32_t>(tsc_rounded >> 32);
+	reg_eax = static_cast<uint32_t>(tsc_rounded & 0xffffffff);
```
johnnovak, line 256: Okay, I'll refrain from individual comments as my general thought is this should be vastly simplified.

1. TSC is Pentium only. Adding support for pre-Pentium CPUs muddies the waters and is not useful much, as I'll explain later. It just adds unnecessary complexity, so I suggest to simplify the whole thing.
2. Rounding to the nearest know CPU model is a questionable feature. I highly suggest to get rid of it. We can set any cycles number, so why round anything?  Also, you run Pentium CPUs on all sorts of speeds, depending on BIOS and motherboard support. For example, I'm underclocking my Pentium MMX 166 to 100 MHz because it runs cooler and slower is better for most DOS stuff anyway.
3. Ideally, you would get rid of the table approach and just have a simple formula that translates the cycles value to some TSC frequency, job done. Worst case, you could use a function with a. few linear segments to approximate the TSC for a given cycles value for `pentium` and `pentium_mmx` only. Then this whole file could go away and it will all become a one-liner or a small function somewhere else.

Keep it simple, don't add complexity, unless we really need it—that's my preference 😄

johnnovak:
Nice enhancement @FeralChild64. What's the motivation for this, by the way? If just making the CPU identification a bit more accurate in some benchmark utils, that's fine.

But could this affect anything else than benchmarks? Most importantly, could it break games or demos? Conversely, could it help with certain games or demos?

In any case, it's a nice minor enhancement even if it only improves the output of some benchmarks.

> * the code contains values even for pre-Pentium CPU; if more data is collected, and the team is interested, we might extend `cycles` syntax, so that, for example, `cycles = 66mhz` would be allowed

As I explained, `tsc.cpp` is a bad place for this—pre-Pentium CPUs don't have TSC.

But such an approach won't fly. I've thought a lot about it previously as I really wanted to introduce support for something like `cycles = 386dx40` or `cycles = 486dx2-66`. Where this approach all falls apart is that we literally can't accurate emulate any concrete CPU model with a fixed cycles value. So a simple mapping for `386dx40` gets you in the _ballpark_, but often you'll need to adjust it by +/- 50% (or more in extreme cases) to get the game running optimally. This makes introducing such "symbolic cycles constants" problematic.

What we _could_ do, and we will in time, is to "seed" the cycles value from the OSD. So say in the OSD you have a "CPU type" dropdown, and you select "386 DX/40". That will _seed_ the "cycles" slider with the appropriate number which is shown in the OSD, then you can adjust it as needed. But at the time you're adjusting the _number_. That's the only practical approach that can work and is not confusing, as I see it—in the config, I think it's best to stick with cycles values as numbers forever.

src/cpu/tsc.cpp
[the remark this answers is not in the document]
FeralChild64: Fixed. It was a mistake.

src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+		        while (++iter1 != iter_after) {
+			        iter2 = iter1;
+		        }
+		        return iter2;
+	        };
+
+	// Calculate the coefficient to be used to convert cycles to MHz
+	const auto iter_upper = cycles_mhz_map.upper_bound(cycles);
+	if (iter_upper == cycles_mhz_map.end()) {
+		const auto iter_lower = getIterBefore(cycles_mhz_map.end());
```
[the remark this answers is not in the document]
FeralChild64: OK, changed.

src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+	// Helper lambda to get std::map iterator before the given one
+	auto getIterBefore =
+	        [&cycles_mhz_map](const CyclesMhzMap::const_iterator& iter_after) {
+		        auto iter1 = cycles_mhz_map.begin();
+		        auto iter2 = iter1;
+		        while (++iter1 != iter_after) {
+			        iter2 = iter1;
+		        }
+		        return iter2;
+	        };
```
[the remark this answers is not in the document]
FeralChild64: OK, changed.

src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+	// Calculate the coefficient to be used to convert cycles to MHz
+	const auto iter_upper = cycles_mhz_map.upper_bound(cycles);
+	if (iter_upper == cycles_mhz_map.end()) {
+		const auto iter_lower = getIterBefore(cycles_mhz_map.end());
+		coeff = iter_lower->second / iter_lower->first;
+	} else if (iter_upper == cycles_mhz_map.begin()) {
+		coeff = iter_upper->second / iter_upper->first;
+	} else {
+		const auto iter_lower = getIterBefore(iter_upper);
```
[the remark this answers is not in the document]
FeralChild64: OK, changed.

FeralChild64:
@johnnovak Yes, I wanted to improve the information given by various benchmarks / system info software. Seeing a _Pentium MMX 70 MHz_ in _SiSoft Sandra_ (screenshots in PR https://github.com/dosbox-staging/dosbox-staging/pull/3426), feels odd to me, and the default conversion ratio between cycles and MHz is also way off - `cycles=400000` does not correspond to a 3.5 GHz CPU.

I have added a log in `CPU_ReadTSC` function to check which games might be affected, and found nothing. _Duke Nukem 3D_, _Quake_, _Blood_, _Pył_, _Tomb Raider_, _GTA_, _Z_, _Extreme Assault_, started _Windows 98_ - nothing seems to execute `RDTSC` instruction, except for benchmarking software.

OK, I'll simplify the code if no `cycles = xxx MHz` is planned.

src/cpu/tsc.cpp
FeralChild64: OK, done.

src/cpu/tsc.cpp
```diff
@@ -0,0 +1,257 @@
...
+	if (iter_upper == typical_mhz_values.end()) {
+		constexpr double resolution = 100.0f;
+		return std::round(cpu_mhz / resolution) * resolution;
+	}
+
+	if (iter_upper == typical_mhz_values.begin()) {
+		return typical_mhz_values[0];
+	}
+
+	const auto iter_lower = iter_upper - 1;
```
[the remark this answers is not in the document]
FeralChild64: OK, changed.

src/cpu/tsc.cpp
FeralChild64: @johnnovak I have simplified the code, there is no _round to the nearest real CPU clock_ code anymore, calculation segments were already working in a linear way.

@Torinde Pentium 200 MHz was really hard to get back then - believe me, I have tried to buy one in late 90s :) AFAIK the Pentium MMX over 233 MHz were mobile CPUs.

FeralChild64: Update: I have simplified it even further, there is a single coefficient to convert cycles to MHz for the Pentium, and another one for the Pentium MMX.

johnnovak: Yeah that's what I hoped for when I made my above comment that something like this would be sufficient.

src/cpu/cpu.cpp
```diff
@@ -2137,64 +2137,20 @@ static double get_estimated_cpu_mhz(const int32_t cycles)
...
-		assert(weight_upper >= 0);
-		assert(weight_lower >= 0);
-		assert(weight_upper > 0 || weight_lower > 0);
-
-		const double coeff_lower = iter_lower->second / iter_lower->first;
-		const double coeff_upper = iter_lower->second / iter_upper->first;
-
-		coeff = (coeff_lower * weight_lower + coeff_upper * weight_upper) /
-		        (weight_lower + weight_upper);
+		return cycles / DivisorPentiumMmx;
```
johnnovak, line 2152: Nice outcome @FeralChild64 🎉 There's no code like no code 😎 I love it how you managed to simplify this to the minimum that does the job 🚀

FeralChild64: I would still prefer to have a better cycles<->MHz mapping, but this is probably impossible without running a lot of benchmarks on both DOSBox and vintage hardware - multiple vintage PCs, in fact, as the chipset, RAM speed, cache size, or even CPU core revision can affect the speed.

At least the system info tools now give somewhat more realistic values, and this was my main goal.

johnnovak: Well, like you said, because of these differences no two PCs are exactly the same, so DOSBox just emulates just another "generic" PC, so being super accurate with these speed targets it a little bit a moot point. Sure, emulating some aspects of a real PC better would perhaps help with some speed sensitive games (bus contention, RAM speed, or whatever), but emulating motherboard X with chipset Y with RAM speed Z with a very specific CPU version is most likely out of scope for DOSBox.

"Cycle accurate" PC emulation just makes very little sense when no two people had the same hardware in the back in the way and PCs are inherently "cycle incompatible" with each other to some extent.

src/cpu/tsc.cpp
FeralChild64: Well, one could run few different benchmarks (Norton SysInfo, Quake, RAR, …) on a real CPU running at different frequencies, on a few different motherboards, compare this with DOSBox results, select some curve that looks similar, do the least square approximation to select the best curve parameters, outside of the normal frequency range use some linear extrapolation… doable (in fact, I did something similar at work once, to more-or-less predict how much time the calculations are going to take), but that would require a lot of work and a lot of vintage hardware. Definitely an overkill.

src/cpu/cpu.cpp
```diff
@@ -2131,6 +2132,61 @@ void CPU_ENTER(bool use32,Bitu bytes,Bitu level) {
...
+// Estimate the CPU speed in MHz given the amount of cycles emulated
+static double get_estimated_cpu_mhz(const int32_t cycles)
+{
+	assert(CPU_ArchitectureType >= ArchitectureType::Pentium);
+
+	// These will result Pentium 100 needing around the same cycles as
+	// estimated by the DOSBox X team on their wiki:
+	// https://dosbox-x.com/wiki/Guide%3ACPU-settings-in-DOSBox%E2%80%90X#_cycles
+	// and assumes Pentium MMX has roughly 16% better IPC, as stated here:
+	// https://www.tomshardware.com/reviews/pentium-mmx-live-expectations,19.html
```
johnnovak, line 2144: Nice explanation.

johnnovak (approved):
This looks very good now @FeralChild64, great job simplifying the implementation 🏅

PR #3494, opened 2024-03-10
More PVS warning fixes

weirddan455:
# Description

I've gone through the whole list now.  These are more low hanging fruit.  The rest I didn't feel confident in changing without going deep into the code.  A lot are non-issues that can probably be ignored but I would say leave them for now and one of us can maybe take a look if we happen to be in that part of the code doing something else.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #3497, opened 2024-03-11
Fix PVS-Studio workflow on branched PRs & forks

johnnovak:
# Description

The goal here is to allow the PVS-Studio workflow run in forked PRs. If we don't do that, any number of PVS warnings can slip through, then the next PVS run on `main` would fail. This is a pain.

The complication is that PVS-Studio needs a license key, and that's stored as a GitHub Actions secret. Forked PRs don't have access to secrets by default. We can however give access to the secrets and full read/write access in forked PRs as well with `pull_request_target`, but that checks out the target branch (original base branch) by default due to security reasons.

With this change, we explicitly override that to check out the code from the PR instead. This way, we can run the code from the PR while still having access to the secrets (the PVS-Studio license is stored as a secret).

This is potentially dangerous, as any random contributor could modify any of the referenced scripts in the PVS workflow to dump the secrets into the logs or upload them somewhere. Therefore, we enforce in the repository settings that workflow runs must always be manually approved on all forked PRs from external contributors.

The important bit is **_ALWAYS REVIEW PRs from unknown external contributors_** first and look out for malicious code before approving the workflow runs!

Naturally, we trust team members and people on the collaborator list, otherwise no one would be able to get any work done here... 😅 

I've tightened our security so we **_always_** must approve workflow runs in PRs from external contributors (not just first time contributors):

image

## Related issues

https://github.com/dosbox-staging/dosbox-staging/pull/3494

# Manual testing

TBD

# Checklist

I have:

### DIFF
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -14,12 +14,12 @@ on:
       - 'website/**'
 
   # Same as 'pull_request', except for forked PRs where it uses the last
-  # commit of the workflow from the *target branch*, plus gives read/write
-  # token permissions and access to secrets.
-  # 
+  # commit of the *target branch*, plus gives read/write token permissions and
+  # access to secrets.
+  #
   # Access to secrets is the important bit here; with regular 'pull_request',
   # forked PRs could not access the PVS-Studio license stored as a secret and
-  # this workflow would just fail.
+  # this workflow would fail.
   pull_request_target:
     paths-ignore:
       - '.clang-format'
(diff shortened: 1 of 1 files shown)


weirddan455:
First of all, this change didn't work.  It's still checking out the top of main as you can see here in the last line under "checkout repository":  https://github.com/dosbox-staging/dosbox-staging/actions/runs/8225950855/job/22491720095?pr=3497

Secondly, if we do need to expose secrets, can we limit it to `PVSSTUDIONAME` and `PVSSTUDIOKEY`?  This documentation seems to imply you can limit to only the secrets the workflow actually needs:  https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions

johnnovak:
> First of all, this change didn't work. It's still checking out the top of main as you can see here in the last line under "checkout repository": https://github.com/dosbox-staging/dosbox-staging/actions/runs/8225950855/job/22491720095?pr=3497

Ok, the `ref` stuff needs more work, apparently 😅 

> Secondly, if we do need to expose secrets, can we limit it to `PVSSTUDIONAME` and `PVSSTUDIOKEY`? This documentation seems to imply you can limit to only the secrets the workflow actually needs: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions

Yeah we can do that, but we still need `pull_request_target`.

johnnovak:
Attempt 2, @weirddan455. Can you please test? Here goes nothing! 😎

johnnovak:
I'm giving up... Contrary to the info I found in various places, it seems it's no longer possible retrieve the hash of the last commit in a forked PR when using `pull_request_target`. Maybe they tightened their general security at GitHub or something.

Unless someone has an idea how to achieve this, maybe we'll need to live without running PVS in forked PRs, after all.

shermp:
Is it actually required to set the `ref` in `actions/checkout`?

johnnovak:
> Is it actually required to set the `ref` in `actions/checkout`?
> 
> EDIT: never mind, I read the linked post. Although the "Require approval for all outside collaborators" option should negate some of the issues?

The problem is I cannot check out the actual PR code in the forked PR when using `pull_request_target`. Which makes the whole thing useless for what we're trying to achieve here.

johnnovak:
Reverting to the old behaviour: PVS won't run in forked PRs 🤷🏻 .

PR #3505, opened 2024-03-12
Fix macOS CI workflow

kklobe:
# Description

The macOS workflow was failing due to conflicts between the macos-13 x64 runner's system-installed Python and homebrew. Downgrading to the macos-12 image allows it to build. The macos-14 free runner only supports arm64 right now; x64 requires macos-14-large, which isn't available for free yet. The long-term solution is to move away from homebrew, but this should hold us for now.

This puts us back to Xcode 14.2, so our C++20 support will be a bit limited.

# Manual testing

Downloaded and ran the x64 (via Rosetta) and arm64 variants.

# Checklist

I have:

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         conf:
           - name: Clang
-            host: macos-13
+            host: macos-12
             arch: x86_64
             needs_deps: true
             build_flags: -Denable_debugger=normal
@@ -67,7 +67,7 @@ jobs:
 
       - name: Cache subprojects
         id:   cache-subprojects
-        uses: actions/cache@v4.0.1
+        uses: actions/cache@v4
         with:
           path: subprojects.tar
           key: subprojects-${{ hashFiles('subprojects/*.wrap') }}-1
@@ -112,7 +112,7 @@ jobs:
     strategy:
       matrix:
         runner:
-          - host: macos-13
+          - host: macos-12
             arch: x86_64
             build_flags: -Db_lto=true
             brew_path: /usr/local/homebrew
@@ -136,12 +136,11 @@ jobs:
       - name: Install C++ compiler and libraries
         if:   matrix.runner.needs_deps
         run: |
-          pip3 install meson setuptools
-          brew install librsvg tree libpng ninja opusfile speexdsp
+          brew install librsvg tree libpng ninja opusfile speexdsp meson python-setuptools
 
       - name: Cache subprojects
         id: cache-subprojects
-        uses: actions/cache@v4.0.1
+        uses: actions/cache@v4
         with:
           path: subprojects.tar
           key:  subprojects-${{ hashFiles('subprojects/*.wrap') }}-1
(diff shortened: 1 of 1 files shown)


weirddan455:
I'll defer final judgement to someone who actually uses Mac since I do not.  However, I think there's ways to work around this without downgrading if doing so costs us C++ 20 compatibility.

```
 Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.11
Target /usr/local/bin/2to3-3.11
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.11'

  brew link --overwrite python@3.11

  brew link --overwrite python@3.11 --dry-run
```

Can we not just do `brew link --overwrite python@3.11` somewhere?  We're already using the overwrite trick for openssl here:

https://github.com/dosbox-staging/dosbox-staging/blob/40033e1bc4c0709fe0919966efb2048fe0b4d38c/.github/workflows/macos.yml#L62

There's also various ways of maintaining multiple Python installs on your system.  This project for example:  https://github.com/pyenv/pyenv

Of course, being a CI machine we probably don't care about clobbering system files that will just get re-generated on the next run anyway.

kklobe:
> I'll defer final judgement to someone who actually uses Mac since I do not. However, I think there's ways to work around this without downgrading if doing so costs us C++ 20 compatibility.
> 
> ```
>  Error: The `brew link` step did not complete successfully
> The formula built, but is not symlinked into /usr/local
> Could not symlink bin/2to3-3.11
> Target /usr/local/bin/2to3-3.11
> already exists. You may want to remove it:
>   rm '/usr/local/bin/2to3-3.11'
> 
> To force the link and overwrite all conflicting files:
>   brew link --overwrite python@3.11
> 
> To list all files that would be deleted:
>   brew link --overwrite python@3.11 --dry-run
> ```
> 
> Can we not just do `brew link --overwrite python@3.11` somewhere? We're already using the overwrite trick for openssl here:
> 
> https://github.com/dosbox-staging/dosbox-staging/blob/40033e1bc4c0709fe0919966efb2048fe0b4d38c/.github/workflows/macos.yml#L62
> 
> There's also various ways of maintaining multiple Python installs on your system. This project for example: https://github.com/pyenv/pyenv
> 
> Of course, being a CI machine we probably don't care about clobbering system files that will just get re-generated on the next run anyway.

I started down that road, but it still causes the CI to fail. Basically every single Python-related executable gets in the way. I even tried ripping out the image-installed Python before the meson call with a `rm -rf`, but it just wasn't working.

This gets us up and running for now and gives us some time to figure out a better solution (which is getting rid of homebrew for dependencies!).

johnnovak (approved):

johnnovak:
Thanks for figuring out all these arcane details @kklobe and getting our macOS builds back on track.

The C++20 stuff is a nice to have but hardly essential. Getting the macOS builds going again with the limitations GitHub currently impose on us definitely trump it, so you made the right choice.

If C++20 support comes say one year later, it's no big deal, really.

kklobe:
> Thanks for figuring out all these arcane details @kklobe and getting our macOS builds back on track.
> 
> The C++20 stuff is a nice to have but hardly essential. Getting the macOS builds going again with the limitations GitHub currently imposes on us definitely trump it, so you made the right choice.
> 
> If C++20 support comes say one year later, it's no big deal, really.

Oh, don't misunderstand me - even Xcode 14 has a ton of C++20 stuff. Xcode 15.3 brought `std::format` and some other esoteric additions, but no claim could be made for Staging being held back by not being there right away :grin:

johnnovak:
> Oh, don't misunderstand me - even Xcode 14 has a ton of C++20 stuff. Xcode 15.3 brought `std::format` and some other esoteric additions, but no claim could be made for Staging being held back by not being there right away 😁

Right, I missed that detail. Well, the situation is even better than I thought! Great 😎

Issue #3501, opened 2024-03-12
Time Warriors (1997), Deus (1996), Ishar 3 (1994), & Robinson's Requiem (1994)


Python-Exoproject:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

81.0

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Straight after the 'Press DEL to enter setup' screen goes away the games crash with the below screenshot.

Time Warriors also crashes in 80.1 and Daum. It works in X, 0.74, and ECE

Deus also crashes in 80.1 and Daum. In X it hardlocks. It works in 0.74, and ECE

Edit: Ive added more notes below with further information and added two more games with similar issue (Ishar 3 and Robinsons Requiem)

![TimeWarriors](https://github.com/dosbox-staging/dosbox-staging/assets/67406191/05cea578-2985-4848-80bd-a5887c5ed07b)

### Steps to reproduce the behaviour.

I used the exodos version of the games. Run with staging, issue happens during loading of the game

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Python-Exoproject:
Typical, just as I post it I solve it. Seems the games are sensitive to cycles and the same cycles are not cross compatible between all dosbox forks

Python-Exoproject:
Ive found that the crash only happens during the game boot. You can start the game at 10k and then once youve started gameplay increase the cycles, but there must be someway to make Staging do better than that. Even 0.74 can go up to 55k and boot the game (at 60k 0.74 also crashes with an "Illegal Interrupt" error

Python-Exoproject:
Time Warriors is the same as Deus, however there is an additional problem if I want to move this game over to staging in that the game requires an external tsr called 'cdseek.com' (created by ripsaw I think) to get the CD audio working in game. When testing Staging at 10k I noticed its not working

Python-Exoproject:
[...]

PR #3507, opened 2024-03-13
Add `vga_render_per_scanline` option

johnnovak:
# Description

Partially fixes https://github.com/dosbox-staging/dosbox-staging/issues/3501

This flag fixes the crash-at-startup regressions with the following games:

- Deus (1996)
- Ishar 3 (1994)
- Robinson's Requiem (1994)
- Time Warriors (1997)

This is enabled by default to render the VGA image by scanline, which is the new accurate VGA emulation behaviour.

However, a few games (e.g., Deus, Ishar 3, Robinson's Requiem, Time Travelers) would crash at startup with per-scanline rendering enabled. This is most likely due to some VGA emulation deficiency.

Disabling `vga_render_per_scanline` restores the the legacy DOSBox behaviour of rendering the VGA image in 4 parts instead of per-scanline.

Of course, this is a workaround, but finding the root cause for this in the VGA code could take ages... I'm not up for that job right now. We just want to get a patch out to get the games working again. If someone improves the accuracy of the VGA code later, this flag could become a no-op or we can just deprecate/remove it.

# Manual testing

Tested that the games run fine with the following configs. `vga_render_per_scanline = off` is required for all of them and the `cycles` settings needs to be in a certain range for each game (outside of this range, you'll get the crash again).

Tested on both macOS (ARM) and Windows 10 (x86).

### Deus (1996)

```ini
[dosbox]
vga_render_per_scanline = off

[cpu]
# going higher results in a crash at startup
cycles = 40000

[autoexec]
imgmount d cd/Deus.cue -t iso
d:
start
```

### Ishar 3 (1994)

```ini
[dosbox]
vga_render_per_scanline = off

[cpu]
cycles = 20000

[autoexec]
imgmount d cd/Ishar_3.cue -t iso
d:
start
```

### Robinson's Requiem (1994)

```ini
[dosbox]
vga_render_per_scanline = off

[cpu]
# going higher results in a crash at startup
cycles = 22000

[autoexec]
imgmount d cd/robin.cue -t iso
d:
start
```

### Time Warriors (1997)

```ini
[dosbox]
vga_render_per_scanline = off

[cpu]
# going higher results in a crash at startup
cycles = 60000

[autoexec]
imgmount d cd/Timewar.cue -t iso
d:
start
```

# Checklist

I have:

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -584,6 +584,12 @@ void DOSBOX_Init()
 	pbool = secprop->Add_bool("vga_8dot_font", only_at_start, false);
 	pbool->Set_help("Use 8-pixel-wide fonts on VGA adapters (disabled by default).");
 
+	pbool = secprop->Add_bool("vga_render_per_scanline", only_at_start, true);
+	pbool->Set_help(
+	        "Emulate accurate per-scanline VGA rendering (enabled by default).\n"
+	        "Currently, you need to disable this for a few games, otherwise they will crash\n"
+	        "at startup (e.g., Deus, Ishar 3, Robinson's Requiem, Time Warriors).");
+
 	pbool = secprop->Add_bool("speed_mods", only_at_start, true);
 	pbool->Set_help(
 	        "Permit changes known to improve performance (enabled by default).\n"
--- a/src/hardware/vga_draw.cpp
+++ b/src/hardware/vga_draw.cpp
@@ -1452,7 +1452,20 @@ PixelFormat VGA_ActivateHardwareCursor()
 // A single point to set total drawn lines and update affected delay values
 static void setup_line_drawing_delays(const uint32_t total_lines)
 {
-	vga.draw.parts_total = total_lines;
+	const auto conf    = control->GetSection("dosbox");
+	const auto section = static_cast<Section_prop*>(conf);
+	assert(section);
+
+	if (vga.draw.mode == PART && !section->Get_bool("vga_render_per_scanline")) {
+		// Render the screen in 4 parts; this was the legacy DOSBox behaviour.
+		// A few games needs this (e.g., Deus, Ishar 3, Robinson's Requiem,
+		// Time Travelers) and would crash at startup with per-scanline
+		// rendering enabled. This is most likely due to some VGA emulation
+		// deficiency.
+		vga.draw.parts_total = 4;
+	} else {
+		vga.draw.parts_total = total_lines;
+	}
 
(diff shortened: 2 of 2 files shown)


johnnovak:
Can you please test the above with this build @Python-Exoproject?

Make sure to replace the whole Staging folder in your test setup with the contents of the ZIP file, not just the EXE (e.g., append a `.bak` extension to your existing Staging folder for safekeeping).

https://github.com/dosbox-staging/dosbox-staging/actions/runs/8258929473/artifacts/1320910072

Grounded0:
We should file an issue to the issue tracker that this should be resolved in the longer term and remains an active issue. Getting this removed is more important than new features but it's acceptable solution for now.

johnnovak:
> We should file an issue to the issue tracker that this should be removed in the longer term and remains an active issue. Getting this removed is more important than new features but it's acceptable solution for now.

Well, ideally yeah. But that VGA code is extremely hard to figure out, so until someone does that, we'll need to live with this practical solution which is honestly not too terrible. Does the job, you can play the game.

weirddan455 (approved):
Looks like a simple enough workaround for now.  Nice job tracking this down.  Would not have expected VGA to be the culprit (but then again it always comes back to VGA lately doesn't it? :wink: )

johnnovak:
To be clear, original DOSBox and SVN crashes exactly the same way with `machine=vgaonly`. Probably X too but someone should try.

Python-Exoproject:
I can confirm all the fixes work, but is there any suggestions on getting better stability at higher cycles? Ishar 3 and Robinsons are fine but Time Warriors is a bit slow at 60k for a fighting game (70-75 would be optimal for smooth combat) and for Deus 40k runs the basic resolution version but if you press tab in game to activate HD mode its a crawl unless the cycles are pumped to to 50-60k range

johnnovak:
> I can confirm all the fixes work, but is there any suggestions on getting better stability at higher cycles?

Unfortunately, the only thing I can recommend now is to increase cycles after startup which works.

As mentioned in my fix tickets, the root cause for the issue is complex. None of these games work with `vgaonly` in original DOSBox or SVN, and we made effectively an improved version of `vgaonly` the only VGA setting. So the proper fix that will hopefully address the deficiency of the authentic `vgaonly` VGA emulation that we inherited will likely also fix the cycles-dependent crash issue.

That's the hope, at least. But this exercise is potentially very time consuming, so not patch release material.

johnnovak:
Raised an issue ticket for the proper fix for an adventurous soul 😎 
https://github.com/dosbox-staging/dosbox-staging/issues/3512

Issue #3518, opened 2024-03-18
SimCity 2000 Special Edition (1995) issues


bsmith1702:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

This only seems to happen when running the game through the batch file.  When quitting, the DOS screen spits out a huge wall of nonsense text:

![image0001](https://github.com/dosbox-staging/dosbox-staging/assets/155026848/84b246f7-e4f5-4007-b17c-0a6821d91331)

This is just a small portion of what is generated.  Most of it is just random random characters.  A few files with nonsense filenames are also created:

![screenshot](https://github.com/dosbox-staging/dosbox-staging/assets/155026848/d2d02de7-c735-4894-90b7-8390577df579)

This behavior also happens when trying to play the WillTV videos.  (You click on the WillTV button in the game, the is supposed to quit and run WILLTV.EXE).  I've tried running this in DOSBox-ECE.  It doesn't quite show the same behavior; instead it just hangs.  The version of DOSBox that GOG provides (0.73) seems to work fine.  It is worth noting that the problem seems to be with the batch file.  Running the game or the WillTV player directly with the executables seems to work fine.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Launch the game with "call SC2000.BAT" (see attached)
2. Quit or select "Watch WillTV" from the menu:
![image0002](https://github.com/dosbox-staging/dosbox-staging/assets/155026848/17702460-fae5-4cbf-845d-6bfa0c652f16)

[Batch_file.zip](https://github.com/dosbox-staging/dosbox-staging/files/14630321/Batch_file.zip)

### Download URL of affected game or software

https://www.gog.com/en/game/simcity_2000_special_edition

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
@bsmith1702 Can you also reproduce with 0.80.1? If not, we must treat this as a regression.

amak79:
I can reproduce this issue with 0.81.0 (Gentoo Linux) and the CD-CROM version of SimCity 2000 SE. It does not occur with 0.80.1.

My config file:

```
[cpu]
cycles    = 65000

[autoexec]
MOUNT C "dosbox"
[...]

PR #3528, opened 2024-03-22
Keep file handle instead of entry in FileReader

MeAreJeenius:
# Description

When dealing with dosbox file IO (ie. DOS_{Open,Read,Seek,Close}File), it works with what dosbox calls file entries. These are the IDs which these functions will take in to find out which file to operate on.

Despite these sounding exactly like file handles, which is the assumption I was under when making FileReader, dosbox's internal system of open files uses a different set of IDs. I'm not sure why this is. There are functions to translate between the two.

From what I can tell, SC2K basically does something strange internally and invalidates the file handles somehow. When the entry was being used to keep track of the file, its status was not being updated. This means it was essentially a dangling handle, which is why it was reading out garbage. With these new changes having the handle being used instead, it seems that its corresponding entry is set to 0xff, which is used as the "null" ID, so it simply stops the rest of the batch file from being read, since the next read fails. This should bring the behavior to parity before this regression.

If I had to guess, the other problems mentioned in the linked issue are likely due to some deeper problem in how files are being handled.

## Related issues

Resolves the regression portion of #3518 

# Manual testing

Since the issue can't be generalized to a simple problem, the only thing that can be done is to test with the problematic batch file to ensure that the issue doesn't persist, and to test with a wide range of batch files to ensure that nothing is broken by the changes.

# Checklist

I have:

### DIFF
--- a/src/shell/file_reader.cpp
+++ b/src/shell/file_reader.cpp
@@ -18,69 +18,52 @@
  *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-#include <cassert>
-#include <utility>
-
 #include "file_reader.h"
 
 std::optional<FileReader> FileReader::GetFileReader(const std::string& filename)
 {
+	auto fullname = DOS_Canonicalize(filename.c_str());
+	
 	std::uint16_t handle = {};
-	if (!DOS_OpenFile(filename.c_str(), (DOS_NOT_INHERIT | OPEN_READ), &handle)) {
+	if (!DOS_OpenFile(fullname.c_str(), (DOS_NOT_INHERIT | OPEN_READ), &handle)) {
 		return {};
 	}
-	return FileReader(handle);
+	return FileReader(std::move(fullname));
 }
 
-FileReader::FileReader(const uint16_t file_handle) : handle(file_handle) {}
-
-FileReader::FileReader(FileReader&& other) noexcept
-        : handle(std::exchange(other.handle, {}))
+FileReader::FileReader(std::string filename)
+        : filename(std::move(filename)),
+          cursor(0)
 {}
 
-FileReader& FileReader::operator=(FileReader&& other) noexcept
+std::string FileReader::Read()
 {
-	if (this == &other) {
-		return *this;
+	uint16_t entry = {};
+	if (!DOS_OpenFile(filename.c_str(), (DOS_NOT_INHERIT | OPEN_READ), &entry)) {
+		return "";
+	}
+	DOS_SeekFile(entry, &cursor, DOS_SEEK_SET);
+
+	uint8_t data           = 0;
+	uint16_t bytes_to_read = 1;
+	std::string line       = {};
+
+	while (data != '\n') {
+		bool result = DOS_ReadFile(entry, &data, &bytes_to_read);
+		if (!result || bytes_to_read == 0) {
+			break;
+		}
+		line += static_cast<char>(data);
 	}
 
-	if (handle) {
-		DOS_CloseFile(*handle);
-	}
-	handle = std::exchange(other.handle, {});
-	return *this;
-}
+	cursor = 0;
+	DOS_SeekFile(entry, &cursor, DOS_SEEK_CUR);
+	DOS_CloseFile(entry);
 
-std::optional<uint8_t> FileReader::Read()
-{
-	if (!handle) {
-		return {};
-	}
-
-	std::uint8_t data        = 0;
-	std::uint16_t bytes_read = 1;
-
-	const bool result = DOS_ReadFile(*handle, &data, &bytes_read);
-	if (!result || bytes_read == 0) {
-		return {};
-	}
-
-	return data;
+	return line;
 }
 
 void FileReader::Reset()
 {
-	if (!handle) {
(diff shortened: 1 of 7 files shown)


johnnovak (approved):
Thanks for the detailed explanation and the quick fix @MeAreJeenius, it all makes sense.

I think this is backport-worthy material.

weirddan455:
Alternatively, can we just not keep a DOS file handle around at all?  Instead, in `GetFileReader` we could open the DOS file, read the entire thing into memory, close it, and never deal with the DOS interface again.  The state of the `FileReader` class can just be a `std::vector<uint8_t>` plus a position.  Should work fine baring very large files or something very strange like a self-modifying batch file.

Just an idea though.  What do you guys think?

johnnovak:
Great pickup @weirddan455. I've looked into it and it turns out self-modifying MS-DOS batch files are a thing:

https://stackoverflow.com/a/27698558

The whole thread is worth reading, and multiple commenters confirm that remembering the file offset (not the line number) is the correct behaviour. So it seems to me the original code that got removed got this obscure detail right.

While it's not highly likely we'll run into self-modifying game installer batch files anytime soon, it's better to preserve the authentic behaviour, especially when it's not particularly hard to do so.

FeralChild64:
We have a self-modifying .BAT file - Z:\AUTOEXEC.BAT.
At least it can get modified while being run if KEYB command is executed inside [autoexec] section to switch the code page.

johnnovak:
Good writeup @MeAreJeenius. That hack is terrible, I vote for removal... Given that it doesn't even work as it's supposed to, I think that's a no-brainer. Then if the WINE folk need support for some similar functionality, we can treat that separately and implement it cleanly.

weirddan455:
I just tested running a DOS program through Wine on my Fedora 39 system.  It uses Dosbox Staging 0.80.1 as the system Dosbox implementation.  There is no other fork available in the mainline repos.

This feature is broken here.  It runs with these arguments `arguments: dosbox -userconf -conf /home/daniel/.wine/dosdevices/c:/users/daniel/Temp/cfgd358.tmp`.  Staging goes through what looks like normal init, mounts some Wine paths as drive C and D, and then exits before even showing the window.  I'm not sure what's in that temporary config file because it gets deleted immediately afterwards.

For what it's worth I've been using Wine and its various forks for close to 20 years and I've never used this Dosbox integration.  I suppose the usecase is some Windows 95 program wanting to launch a DOS .exe.

I'm fine with removing this feature.  It sounds like a bad implementation and it doesn't work correctly anyway.

johnnovak:
Great testing @weirddan455. We are all in agreement here @MeAreJeenius, that dirty hack has to go. Thanks for bringing the issue to our attention. 🏅

johnnovak:
Nice one @MeAreJeenius. Are we all happy with this now @FeralChild64 and @weirddan455 ?

FeralChild64:
Yes, dropping this hack is reasonable.

src/shell/file_reader.cpp
```diff
@@ -18,69 +18,52 @@
...
-
-FileReader::FileReader(FileReader&& other) noexcept
-        : handle(std::exchange(other.handle, {}))
+FileReader::FileReader(std::string filename)
+        : filename(std::move(filename)),
+          cursor(0)
 {}
 
-FileReader& FileReader::operator=(FileReader&& other) noexcept
+std::string FileReader::Read()
```
japsmits, line 39: Shouldn't this return an std::optional\<std::string\>? This way, it's easy to distinguish between a failed read (p.e. if the file doesn't open) and an empty line.

And maybe also a way to see what has failed. There are numerous ways I/O can fail, so I think we should be graceful around that subject.

johnnovak: Good point @japsmits.

johnnovak:
> Looks good. I did some quick testing on my end to confirm the SC2000 bug is fixed and extended ASCII in batch files did not regress again.

@weirddan455 Given there have been further changes to the PR, would you be able to quickly re-execute your test cases before I press the merge button? Thanks mate.

weirddan455:
> @weirddan455 Given there have been further changes to the PR, would you be able to quickly re-execute your test cases before I press the merge button? Thanks mate.

Yep, tests are still good.

johnnovak:
Actually, we can't backport this as it was built on top of the major shell refactor stuff. Tried to cherry pick but I'm getting tons of conflicts. But it's ok, the bug it fixes is very obscure.

weirddan455:
@Torinde That feature request is not really related to Wine integration.  On the Wine side, they're expecting to launch our GUI not have us interact with the terminal.

We've already decided both of those features are out of scope for Staging.  Wine integration is one of the rare cases where we've broken compatibility with original DOSBox.  We must have accidentally broke it some time before this PR and then intentionally removed a hack for it in this PR.  It was a rarely used and poorly implemented feature from the start so I don't feel too bad about it.

The render to console stuff is unlikely to ever happen either.  There's a lot of nuance to getting something like that implemented and it doesn't really help with playing games (aside from text mode games I guess but that's pretty niche and those games will still work fine in our GUI).

johnnovak:
Yep fully agree with @weirddan455 on all points.

Issue #3519, opened 2024-03-18
Build fails with LTO


eli-schwartz:
I tried to build with the following *FLAGS to optimize the build: `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

Note the -Werror=* flags are used to help detect cases where the compiler tries to optimize by assuming UB cannot exist in the source code -- if it does exist, ordinarily the code would be miscompiled, and this says to make the miscompilation a fatal error.

I got this error:
```
FAILED: dosbox 
[...]

PR #3539, opened 2024-03-25
Cleanup Linux workflows, add LTO build

FeralChild64:
# Description

Linux workflows cleanup. This PR also adds LTO check to make sure the issue https://github.com/dosbox-staging/dosbox-staging/issues/3519 won't return.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3519

# Manual testing

I have checked that adding a commit which reverts https://github.com/dosbox-staging/dosbox-staging/commit/8c9b65c2bee5f3c84693581b5496ec3f0d45b313 causes the modified workflows to fail as in issue https://github.com/dosbox-staging/dosbox-staging/issues/3519

# Checklist

I have:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -89,7 +89,16 @@ jobs:
       matrix:
         conf:
 
-          - name: GCC 10, Ubuntu 22.04
+          - name: GCC 13, LTO, Ubuntu 22.04
+            os: ubuntu-22.04
+            packages: g++-13
+            build_flags: -Dbuildtype=debug -Db_lto=true --native-file=.github/meson/native-gcc-13.ini
+            compile_flags: -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing
+            max_warnings: 0
+            needs_all_deps: true
+            run_tests: true
+
+          - name: GCC 12, debug, Ubuntu 22.04
             os: ubuntu-22.04
             packages: g++-12
             build_flags: -Dbuildtype=debug --native-file=.github/meson/native-gcc-12.ini
@@ -97,7 +106,7 @@ jobs:
             needs_all_deps: true
             run_tests: true
 
-          - name: Clang 15, Ubuntu 22.04
+          - name: Clang 15, debug, Ubuntu 22.04
             os: ubuntu-22.04
             packages: clang-15
             build_flags: -Dbuildtype=debug --native-file=.github/meson/native-clang-15.ini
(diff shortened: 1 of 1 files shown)


weirddan455:
LTO on debug builds is a bit unusual, though it probably doesn't hurt anything for CI purposes unless they take significantly longer to build.

I would definitely add these extra compile flags to the Release build though, where we already have LTO enabled.

johnnovak:
> LTO on debug builds is a bit unusual

I agree, can we drop LTO from the debug builds @FeralChild64 ? That's not how we build the debug builds locally either, and it's frankly pointless for non-optimised non-release builds.

FeralChild64:
But nearly all the builds are currently debug ones - only the minimum GCC 9 build isn’t. I can try tomorrow, but I’m not sure if GCC 9 (the only one with the release workflow) is advanced enough for LTO.

weirddan455:
>  But nearly all the builds are currently debug ones - only the minimum GCC 9 build isn’t. I can try tomorrow, but I’m not sure if GCC 9 (the only one with the release workflow) is advanced enough for LTO.

It's already doing LTO and has been for a while.

https://github.com/dosbox-staging/dosbox-staging/blob/d221c959e2323414ab7c210297228884913bce0f/.github/workflows/linux.yml#L326-L336

If it doesn't support the new flags, maybe it's time to bump the compiler version.  We just dropped support for Debian 10.  Ubuntu 20.04 and Debian 11 both have GCC 10 at least.

johnnovak:
> But nearly all the builds are currently debug ones - only the minimum GCC 9 build isn’t.

As I see it, all we need is to add the strict LTO validation/check flags to the release build so it fails with errors. I'm not even sure adding LTO + strict validation to the debug builds adds much value as stuff in debug builds might be inlined differently (or not at all), etc.

> If it doesn't support the new flags, maybe it's time to bump the compiler version. We just dropped support for Debian 10. Ubuntu 20.04 and Debian 11 both have GCC 10 at least.

Sounds reasonable to move the minimum requirement forward as time progresses, yeah. Especially in Linux land.

FeralChild64:
I have tried to cleanup the workflows - if I understand this file correctly, they are named wrongly. I bumped GCC 9 to GCC 10, for the LTO check I have added a GCC 13 build (the former GCC 9 was a minimal build, it was not really suitable for testing LTO) - but now some shell unit tests fail.

Unfortunately, I have to drop this attempt. I have already started to many topics, I won't have much time for the development (if t all) in the upcoming days, and digging now into shell unit tests is the last thing I have time for.

Issue #1314, opened 2021-10-05
Migrate away from C-based file and string operations to C++17 constructs


kcgen:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Whats your question and how can we help?

One of Staging's goal is to modernize the code base when it makes sense: that is, when we can reduce its complexity and make it more understandable, without harming performance or functionality. 

This is a call to help assist with this, on a one-by-one basis. 

For example, here's a function that processes a command-line string involving redirection. This can benefit from the use of std::string, std::regex, and C++17 string-slicing:

``` c++

Bitu DOS_Shell::GetRedirection(char *s, char **ifn, char **ofn, bool *append)
{
	char * lr=s;
	char * lw=s;
	char ch;
	Bitu num=0;
	bool quote = false;
	char *temp = nullptr;
	size_t temp_len = 0;

	while ( (ch=*lr++) ) {
		if(quote && ch != '"') {
			*lw++ = ch;
			continue;
		}

		switch (ch) {
		case '"':
			quote = !quote;
			break;
		case '>':
			*append=((*lr)=='>');
			if (*append) lr++;
			lr=ltrim(lr);
			if (*ofn) {
				delete[] * ofn;
				*ofn = nullptr;
			}
			*ofn = lr;
			while (*lr && *lr!=' ' && *lr!='<' && *lr!='|') lr++;
			//if it ends on a : => remove it.
			if((*ofn != lr) && (lr[-1] == ':')) lr[-1] = 0;
			temp_len = static_cast<size_t>(lr - *ofn + 1u);
			temp = new char[temp_len];
			safe_strncpy(temp, *ofn, temp_len);
			*ofn = temp;
			continue;

		case '<':
			if (*ifn) {
				delete[] * ifn;
				*ifn = nullptr;
			}
			lr = ltrim(lr);
			*ifn = lr;

			while (*lr && *lr != ' ' && *lr != '>' && *lr != '|')
				lr++;

			if ((*ifn != lr) && (lr[-1] == ':'))
				lr[-1] = 0;

			assert(lr >= *ifn);
			temp_len = static_cast<size_t>(lr - *ifn + 1u);
			temp = new char[temp_len];
			safe_strncpy(temp, *ifn, temp_len);
			*ifn = temp;
			continue;

		case '|': ch = 0; num++;
		}
		*lw++=ch;
	}
	*lw=0;
	return num;
}
```

[...]

PR #3542, opened 2024-03-26
Avoid the use of std::string_view::data()

japsmits:
# Description

By definition, std::string_view::data() is not guaranteed to return a zero-terminated string. If a zero-terminated string is expected, this can result in undefined behaviour.

Use std::string::c_str() where appropriate to avoid hidden bugs when a std::string is replaced by a std::string_view.

If a c_str is required (mostly used by logging), convert the std::string_view to a std::string.

## Related issues

#1314 

# Manual testing

Unittests still functional.

Started DOSBox and performed some random commands.

# Checklist

I have:

### DIFF
--- a/include/dma.h
+++ b/include/dma.h
@@ -83,7 +83,7 @@ public:
 	// Reserves the channel for the owner. If a subsequent reservation is
 	// made then the previously held reservation callback is run to
 	// cleanup/remove that reserver (see the EvictReserver call below).
-	void ReserveFor(const std::string_view new_owner,
+	void ReserveFor(const std::string& new_owner,
 	                const DMA_ReservationCallback new_cb);
 
 private:
@@ -93,7 +93,7 @@ private:
 	                   uint8_t* const buffer);
 
 	DMA_ReservationCallback reservation_callback = {};
-	std::string_view reservation_owner           = {};
+	std::string reservation_owner                = {};
 };
 
 class DmaController {
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -2735,7 +2735,7 @@ static void MAPPER_SaveBinds() {
 }
 
 static bool load_binds_from_file(const std::string_view mapperfile_path,
-                                 const std::string_view mapperfile_name)
+                                 const std::string& mapperfile_name)
 {
 	// If the filename is empty the user wants defaults
 	if (mapperfile_name == "")
@@ -2767,7 +2767,7 @@ static bool load_binds_from_file(const std::string_view mapperfile_path,
 	// default, the mapperfile is not provided
 	if (!was_loaded && mapperfile_name != MAPPERFILE)
 		LOG_WARNING("MAPPER: Failed loading mapperfile '%s' directly or from resources",
-		            mapperfile_name.data());
+		            mapperfile_name.c_str());
 
 	return was_loaded;
 }
--- a/src/hardware/innovation.cpp
+++ b/src/hardware/innovation.cpp
@@ -33,7 +33,7 @@ void Innovation::Open(const std::string_view model_choice,
                       const std::string_view clock_choice,
                       const int filter_strength_6581,
                       const int filter_strength_8580, const int port_choice,
-                      const std::string_view channel_filter_choice)
+                      const std::string& channel_filter_choice)
 {
 	using namespace std::placeholders;
 
@@ -45,13 +45,11 @@ void Innovation::Open(const std::string_view model_choice,
 		return;
 	}
 
-	std::string_view model_name = "";
 	int filter_strength = 0;
 	auto sid_service    = std::make_unique<reSIDfp::SID>();
 
 	// Setup the model and filter
 	if (model_choice == "8580") {
-		model_name = "8580";
 		sid_service->setChipModel(reSIDfp::MOS8580);
 		filter_strength = filter_strength_8580;
 		if (filter_strength > 0) {
@@ -59,7 +57,6 @@ void Innovation::Open(const std::string_view model_choice,
 			sid_service->setFilter8580Curve(filter_strength / 100.0);
 		}
 	} else {
-		model_name = "6581";
 		sid_service->setChipModel(reSIDfp::MOS6581);
 		filter_strength = filter_strength_6581;
 		if (filter_strength > 0) {
@@ -97,7 +94,7 @@ void Innovation::Open(const std::string_view model_choice,
 
 		if (!filter_choice_has_bool) {
 			LOG_WARNING("INNOVATION: Invalid 'innovation_filter' setting: '%s', using 'off'",
-			            channel_filter_choice.data());
+			            channel_filter_choice.c_str());
 		}
 
 		mixer_channel->SetHighPassFilter(FilterState::Off);
@@ -129,16 +126,18 @@ void Innovation::Open(const std::string_view model_choice,
 	// Ready state-values for rendering
 	last_rendered_ms = 0.0;
 
+	// Variable model_name is only used for logging, so use a const char* here
+	const char* model_name = model_choice == "8580" ? "8580" : "6581";
 	constexpr auto us_per_s = 1'000'000.0;
 	if (filter_strength == 0)
 		LOG_MSG("INNOVATION: Running on port %xh with a SID %s at %0.3f MHz",
 		        base_port,
-		        model_name.data(),
+		        model_name,
 		        chip_clock / us_per_s);
 	else
 		LOG_MSG("INNOVATION: Running on port %xh with a SID %s at %0.3f MHz filtering at %d%%",
 		        base_port,
-		        model_name.data(),
+		        model_name,
 		        chip_clock / us_per_s,
 		        filter_strength);
 
--- a/src/hardware/tandy_sound.cpp
+++ b/src/hardware/tandy_sound.cpp
@@ -101,7 +101,7 @@ public:
 	static constexpr IOConfig io = {0xc4, 7, 1};
 
 	TandyDAC(const ConfigProfile config_profile,
-	         const std::string_view filter_choice);
+	         const std::string& filter_choice);
 	~TandyDAC();
 
 	bool IsEnabled() const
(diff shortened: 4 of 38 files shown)


src/gui/sdl_mapper.cpp
```diff
@@ -2767,7 +2767,7 @@ static bool load_binds_from_file(const std::string_view mapperfile_path,
 	// default, the mapperfile is not provided
 	if (!was_loaded && mapperfile_name != MAPPERFILE)
 		LOG_WARNING("MAPPER: Failed loading mapperfile '%s' directly or from resources",
-		            mapperfile_name.data());
+		            std::string(mapperfile_name).c_str());
```
weirddan455, line 2770: Can you check to see if the function signatures for some of these can go from `std::string_view` to `const std::string &`?  I'm not seeing the benefit of string view if you have to do this ugly thing to get a proper null terminated string out of it.

japsmits: I'll take a look at that. It won't catch all places, but where appropriate I think it's the better method.

weirddan455:
First commit is good.  We should prefer `.c_str()` because that one gaurentees null-termination.

I think we should reconsider some of our usage of `std::string_view` though.  The use case for that seems to be that we're working exclusively with C++ style string manipulation _and_ we want to be doing something like taking a substring of a large string and passing that around without making redundant copies.

For most use cases, I think `const std::string &` or even `const char *` is better, particularly if we need to do with APIs that want C-style strings.

japsmits:
I agree that for most cases const std::string& is prefered in cases where there is no string-manipulation. Also storing a string-reference as a string_view might be somewhat dubious (and potentially unsafe). This change eliminated all string_view-to-string conversions, so I think it's closer to what we actually want.

src/midi/midi_coreaudio.h
```diff
@@ -76,9 +76,10 @@ class MidiHandler_coreaudio final : public MidiHandler {
 	          soundfont(nullptr)
 	{}
 
-	std::string_view GetName() const override
+	const std::string& GetName() const override
 	{
-		return "coreaudio";
+		static const std::string name("coreaudio");
+		return name;
 	}
```
weirddan455, line 83: I don't like this pattern.  Most of the time I would avoid return values being references.  It's kind of annoying because this is where `std::string_view` should be useful except we interface with C APIs in the usage code.

I would just return a `std::string` (doesn't need to be const) and be done with it though.  It's not performance sensitive code plus return value optimization and small string optimization should both apply here anyway (automatically done by a compiler in a release build).

src/misc/programs.cpp
```diff
@@ -237,14 +237,14 @@ void Program::WriteOut(const char* format, ...)
 	truncated_chars_message(size);
 }
 
-void Program::WriteOut(const char* format, const char* arguments)
+void Program::WriteOut(const char* format, const std::string& arguments)
```
weirddan455, line 240: You should really look at the usage code before making this change.  It's probably being called a lot with `WriteOut("blah", str.c_str())` and now it's doing a double conversion from `std::string` to `char *` and back again.

I would just leave this one alone personally.  There's a lot of C style string manipulation in the code.  Unless you're trying to do a full refactor to use C++ style string functions, there's not much point in changing this.

src/hardware/tandy_sound.cpp
```diff
@@ -474,7 +474,7 @@ TandyPSG::TandyPSG(const ConfigProfile config_profile, const bool is_dac_enabled
 	} else if (!channel->TryParseAndSetCustomFilter(filter_choice)) {
 		if (!filter_choice_has_bool) {
 			LOG_WARNING("TANDY: Invalid 'tandy_filter' value: '%s', using 'off'",
-			            filter_choice.data());
+			            std::string(filter_choice).c_str());
```
weirddan455, line 477: Did you forget this one?

include/dma.h
```diff
@@ -93,7 +93,7 @@ class DmaChannel {
 	                   uint8_t* const buffer);
 
 	DMA_ReservationCallback reservation_callback = {};
-	std::string_view reservation_owner           = {};
+	std::string reservation_owner                = {};
```
weirddan455, line 96: Nice one here :+1:   This has been a source of bugs in the past when a `std::string_view` gets stored in some class and it outlives its owner.

weirddan455 (changes requested):
Looking good.  A couple of things I would change but overall this is a nice cleanup.

src/midi/midi_coreaudio.h
japsmits: True and true. I'll fix it.

src/misc/programs.cpp
japsmits: Good point.

include/dma.h
japsmits: I already considered this potentially harmful code. As far as I can tell, we just eliminated the last one.

johnnovak:
Nice work @japsmits and thanks for raising the bar of our usage of C++ 👍🏻 

As a general comment, I 100% agree with @weirddan455 and I've been saying this for a while to everybody here, at work, and that's my guiding principle whenever I write some code.

_Always pick the simplest construct/approach that does the job with the least complications (but not simpler)._

Seems trivial, but devs (including me sometimes) have a tendency to get lost in the woods chasing the "most advanced" or "most elegant" way of doing things instead of the simplest and most problem-free solution.

String view has two benefits, as I see it:

- It acts as a "wildcard" string type when used in function args (you can pass in a `std::string` and a `char *` through it too).
- It can avoid making copies when working with substrings of large strings.
- It can make string manipulations faster.

Point 1 is useful, points 2 and 3 however are not applicable to us. We're dealing with minuscule strings most of the time and we're not performing hundreds of thousand or millions of string manipulations on them in tight loops.

Because of this, we don't benefit from the theorethical performance improvements in a meaningful way, but we have to pay the price of string views if they become pervasive in the codebase:

- The null terminator is not guaranteed, so we need to converr it to a `std::string` anyway before calling c_str`.
- Stroring string views in longer-lived data structures instead of always copying the string is asking for trouble. The pointer pair of a string view can become invalid if the string data it points to changes, the you get obscure bugs or crashes.

If we were dealing with multi-megabyte or gigabyte text files routinely and we'd be manipulating them intensely, string views would be a must. But for the typically less than 100K text files we're dealing with plus actual std::string` object with the typical average sizes of 10-100 bytes the best thing to do is to store `std::string` in all long-term structs everywhere.

We can use string views sparingly in function args and for temporaries within a function if they add code clarity (performance is a non issue), but I'm strongly against letting string views escape this local temporary scope. Just always return `std::string` from functions, so a plain copy,  and be done with it.

src/midi/midi_coreaudio.h
johnnovak: 100% agree @weirddan455.

src/misc/programs.cpp
johnnovak: Exactly @weirddan455. We don't get brownie points for using C++ string, usage matters 😄 if all you expect are string literals, `const char *` args are fine and sometimes even preferable.

japsmits:
@johnnovak : that! 👍 

And if std::string_view is used, use it locally, in an environment in which you are absolutely sure the referenced std::string is still existent and constant.

src/misc/programs.cpp
japsmits: In this case, it was a `const char*` -> `std::string_view` -> `const std::string&` -> `const char*` transition. 😆 In the end, the good old `const char*` proved to be the best solution.

Your point is valid, especially in an environment like this, where const char* is still the norm in various parts of the project. Whether that is _always_ a good thing can be debated, since it lacks some safeties std::string/std:string_view have regarding string length, boundary checking, and such, but we can live with it.

johnnovak:
> @johnnovak : that! 👍
> 
> And if std::string_view is used, use it locally, in an environment in which you are absolutely sure the referenced std::string is still existent and constant. Same goes with (const) char*, of course.

Yes, I'd be fine using string views as internal class or static module vars. I draw the line at them escaping module or class boundaries.

src/hardware/innovation.cpp
```diff
@@ -129,16 +126,18 @@ void Innovation::Open(const std::string_view model_choice,
 	// Ready state-values for rendering
 	last_rendered_ms = 0.0;
 
+	// Variable model_name is only used for logging, so use a const char* here
+	const char* model_name{model_choice == "8580" ? "8580" : "6581"};
```
johnnovak, line 130: Minor: We try not to use the `var_name{...}` initialiser syntax.

src/midi/midi_fluidsynth.h
```diff
@@ -40,7 +40,7 @@ class MidiHandlerFluidsynth final : public MidiHandler {
 	~MidiHandlerFluidsynth() override;
 	void PrintStats();
 
-	std::string_view GetName() const override
+	std::string GetName() const override
```
johnnovak, line 43: Good riddance! Returning `std::string_view` from classes/modules always makes me feel uneasy...

johnnovak (approved):
Very nice cleanup job @japsmits, thanks a lot for doing this! 🎉

src/hardware/innovation.cpp
japsmits: Fixed.

src/midi/midi_fluidsynth.h
japsmits: I didn't explicitly look for these, but they are a "nice" candidate for a future PR.

japsmits: Even better: looks like this was the only instance where std::string_view was returned.

src/misc/programs.cpp
johnnovak: > Your point is valid, especially in an environment like this, where const char* is still the norm in various parts of the project. Whether that is _always_ a good thing can be debated, since it lacks some safeties std::string/std:string_view have regarding string length, boundary checking, and such, but we can live with it.

No doubt @japsmits, it would be nice to migrate everything to `std::string` in the codebase and use a logging library that uses `std::string` natively. But it would be a huge error-prone refactoring job with not too much of an upside. I wouldn't just switch to an entirely different logging library because of this.

So, this is the practical reality we live in now. Plus not sure how often strings are being passed back and forth between the core code and the various C/"minimal C++" libraries we're using.

src/midi/midi_fluidsynth.h
johnnovak: > Even better: looks like this was the only instance where std::string_view was returned.

Great! 🥳  Thanks for confirming @japsmits! 🎖️

johnnovak:
Merging once you've confirmed @weirddan455 that you're happy with the updates.

weirddan455 (approved):
Looks good, merge away when you're ready.

Issue #3544, opened 2024-03-27
Crash when quitting while in fullscreen


johnnovak:
1. Start with `-noprimaryconf`
2. Alt+Enter
3. Quit

Happens on macOS 100% of the time.

<img width="980" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/015ca871-262f-4340-a92d-66de6af48648">

shermp:
FWIW, I couldn't trigger this on Windows, using a freshly compiled `main`.

johnnovak:
> FWIW, I couldn't trigger this on Windows, using a freshly compiled `main`.

That's good, cheers. Our init & shutdown order is platform dependent, which is terrible...

weirddan455:
It triggers on my Linux system reliably as well.  Null pointer de-reference.

```
/usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13/bits/unique_ptr.h:462: pointer std::unique_ptr<Config>::operator->() const [_Tp = Config, _Dp = std::default_delete<Config>]: Assertion 'get() != pointer()' failed.

Thread 1 "dosbox" received signal SIGABRT, Aborted.
0x00007ffff75a9834 in __pthread_kill_implementation () from /lib64/libc.so.6

(gdb) bt
#0  0x00007ffff75a9834 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff75578ee in raise () at /lib64/libc.so.6
#2  0x00007ffff753f8ff in abort () at /lib64/libc.so.6
#3  0x00007ffff78d95b0 in std::__glibcxx_assert_fail(char const*, int, char const*, char const*) () at /lib64/libstdc++.so.6
#4  0x000000000098cc32 in std::unique_ptr<Config, std::default_delete<Config> >::operator-> (this=0x12edc60 <control>)
    at /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13/bits/unique_ptr.h:462
#5  0x0000000000ea722a in setup_line_drawing_delays (total_lines=400) at ../../src/hardware/vga_draw.cpp:1455
#6  0x0000000000ea4d67 in setup_drawing () at ../../src/hardware/vga_draw.cpp:2838
#7  0x0000000000ea7658 in VGA_SetupDrawing () at ../../src/hardware/vga_draw.cpp:2928
[...]

PR #3545, opened 2024-03-27
Fix crash when quitting while in fullscreen mode

johnnovak:
# Description

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3544

This fixes a crash caused by sdlmain's entangled nature and the general ad hoc randomness of the current shutdown order.

Also fixing some recently introduced PVS-Studio warnings.

Exiting fullscreen before quitting is a silly thing to do anyway. Why would that even be needed? No game does that...

# Manual testing

Quitting with Cmd+Q on macOS from fullscreen mode no longer causes a crash.

# Checklist

I have:

### DIFF
--- a/src/dos/program_serial.cpp
+++ b/src/dos/program_serial.cpp
@@ -176,7 +176,8 @@ void SERIAL::Run()
 		}
 		if (serialports[port_index] != nullptr) {
 			serialports[port_index]->serialType = desired_type;
-			serialports[port_index]->commandLineString = commandLineString;
+			serialports[port_index]->commandLineString = std::move(
+			        commandLineString);
 		}
 		delete commandLine;
 		showPort(port_index);
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -2738,8 +2738,9 @@ static bool load_binds_from_file(const std::string_view mapperfile_path,
                                  const std::string& mapperfile_name)
 {
 	// If the filename is empty the user wants defaults
-	if (mapperfile_name == "")
+	if (mapperfile_name.empty()) {
 		return false;
+	}
 
 	auto try_loading = [](const std_fs::path &mapper_path) -> bool {
 		constexpr auto optional = ResourceImportance::Optional;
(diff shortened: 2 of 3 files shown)


src/dos/program_serial.cpp
```diff
@@ -176,7 +176,8 @@ void SERIAL::Run()
 		}
 		if (serialports[port_index] != nullptr) {
 			serialports[port_index]->serialType = desired_type;
-			serialports[port_index]->commandLineString = commandLineString;
+			serialports[port_index]->commandLineString = std::move(
+			        commandLineString);
```
japsmits, line 180: Not sure why this was done, but since commandLineString is going out of scope anyhow, I would suggest:
```suggestion
			serialports[port_index]->commandLineString.swap(commandLineString);
```

johnnovak: PVS-Studio suggested a move or swap. Does it really matter which one we use?

japsmits: TL;DR: never mind. :smile:

This basically came from some talks about abuse of std::move, but that's not the case here.

For std::string, a swap might save an allocation later on if the string is reused, but in this case I don't think there is much of a difference. commandLineString gets deleted anyhow.

Which basically negates my remark about things going out of scope in my original remark... 🤦

weirddan455 (approved):
Nice easy fix.  I tested this fixes the crash on my system as well.  Exiting fullscreen before quit would just cause extra window flickering even if it wasn't crashing anyway.

shermp:
My only concern about not existing full screen before quitting is if the full screen resolution is not the same as the desktop resolution, in such cases, users could be left in the wrong resolution.

weirddan455:
>  My only concern about not existing full screen before quitting is if the full screen resolution is not the same as the desktop resolution, in such cases, users could be left in the wrong resolution.

I believe SDL handles this in the `SDL_DestroyWindow` function (if not there, in one of the other cleanup functions that we call).

johnnovak:
> My only concern about not existing full screen before quitting is if the full screen resolution is not the same as the desktop resolution, in such cases, users could be left in the wrong resolution.

Ok, we should test this. I never use `fullresolution=original` myself.

johnnovak:
Ok, `fullresolution=original` works fine on macOS when quitting while in fullscreen. Will test Windows later. Can someone try on Linux?

PR #3556, opened 2024-04-01
Better build version handling at runtime

johnnovak:
# Description

In my recent attempt to build 0.81.1, in the macOS build the correct **0.81.1** version shows up in most places, but `mapperfile` still got a `0.82.0-alpha` suffix because... reasons 🤷🏻 

image

The log complains about this too:

image

It turns out although we have a `DOSBOX_GetDetailedVersion()` function that gets injected `VCS_TAG` during the build process, and which correctly returns `0.81.1` in this case, there is also the legacy `VERSION` C `#define` which is injected in a different way, and the two have drifted. Blergh.

Yeah, I could figure out the difference and fix `VERSION` for this particular build, but I simply don't wanna. I want to use a _single source of truth_ for querying the version number at runtime and use `DOSBOX_GetDetailedVersion()` consistently everywhere. Which is what this PR does.

We have enough version number duplications in various resource files, the Meson config, etc., sbuto at least only use a single method at the C++ code level.

I will forward port this to `main` manually later.

# Manual testing

Stuff still compiles and the version number printed out in the intro screen is not broken. But it's tricky to test this properly; the true test will be to build a tagged build once this is merged.

# Checklist

I have:

### DIFF
--- a/src/capture/image/png_writer.cpp
+++ b/src/capture/image/png_writer.cpp
@@ -206,11 +206,14 @@ void PngWriter::WritePngInfo(const uint16_t width, const uint16_t height,
 
 	char software_keyword[] = "Software";
 	static_assert(sizeof(software_keyword) < 80, "libpng limit");
-	char software_value[] = CANONICAL_PROJECT_NAME " " VERSION;
+
+	auto software_value = format_string("%s %s",
+	                                    CANONICAL_PROJECT_NAME,
+	                                    DOSBOX_GetDetailedVersion());
 
 	texts[0].compression = PNG_TEXT_COMPRESSION_NONE;
 	texts[0].key         = static_cast<png_charp>(software_keyword);
-	texts[0].text        = static_cast<png_charp>(software_value);
+	texts[0].text        = static_cast<png_charp>(software_value.data());
 	texts[0].text_length = sizeof(software_value);
 
 	auto num_text = 1;
@@ -218,7 +221,7 @@ void PngWriter::WritePngInfo(const uint16_t width, const uint16_t height,
 	char source_keyword[] = "Source";
 	static_assert(sizeof(source_keyword) < 80, "libpng limit");
 
-	const auto source_value = format_string(
+	auto source_value = format_string(
 	        "source resolution: %dx%d; source pixel aspect ratio: %d:%d (1:%1.6f)",
 	        video_mode.width,
 	        video_mode.height,
@@ -228,7 +231,7 @@ void PngWriter::WritePngInfo(const uint16_t width, const uint16_t height,
 
 	texts[1].compression = PNG_TEXT_COMPRESSION_NONE;
 	texts[1].key         = static_cast<png_charp>(source_keyword);
-	texts[1].text        = const_cast<png_charp>(source_value.data());
+	texts[1].text        = static_cast<png_charp>(source_value.data());
 	texts[1].text_length = source_value.size();
 
 	++num_text;
--- a/src/hardware/serialport/softmodem.cpp
+++ b/src/hardware/serialport/softmodem.cpp
@@ -609,8 +609,15 @@ void CSerialModem::DoCommand()
 		}
 		case 'I': // Some strings about firmware
 			switch (ScanNumber(scanbuf)) {
-			case 3: SendLine("DOSBox Emulated Modem Firmware V1.00"); break;
-			case 4: SendLine("Modem compiled for DOSBox version " VERSION); break;
+			case 3:
+				SendLine("DOSBox Emulated Modem Firmware V1.00");
+				break;
+			case 4: {
+				const auto line = format_string(
+				        "Modem compiled for DOSBox version %s",
+				        DOSBOX_GetDetailedVersion());
+				SendLine(line.c_str());
+			} break;
 			}
 			break;
 		case 'E': // Echo on/off
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -112,7 +112,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 109
+          MAX_BUGS: 108
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/include/mapper.h
+++ b/include/mapper.h
@@ -54,6 +54,8 @@ constexpr uint32_t MMOD3 = 0x4;
 
 typedef void(MAPPER_Handler)(bool pressed);
 
+std::string MAPPER_GetDefaultMapperfileName();
+
 /* Associate function handler with a key combination
  *
  * handler     - function to be triggered
--- a/include/mapper.h
+++ b/include/mapper.h
@@ -29,8 +29,6 @@
 
 #include <SDL.h>
 
-#define MAPPERFILE "mapper-sdl2-" VERSION ".map"
-
 constexpr uint32_t MMOD1 = 0x1;
 #define MMOD1_NAME "Ctrl"
 constexpr uint32_t MMOD2 = 0x2;
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -27,15 +27,6 @@
 // Name of project, lower-case without spaces
 #define CANONICAL_PROJECT_NAME "@project_name@"
 
-// Emulator Semantic Version (MAJOR.MINOR.PATCH), incremented as follows:
-//  - MAJOR version when you make incompatible API changes
-//  - MINOR version when you add functionality in a backwards compatible manner
-//  - PATCH version when you make backwards compatible bug fixes
-// Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
-// Ref: https://semver.org/
-
-#define VERSION "@version@"
-
 /* Strings to be returned by virtual drivers, etc.
 */
 
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -1264,6 +1264,10 @@ protected:
 	uint16_t button_state = 0;
 };
 
+std::string MAPPER_GetDefaultMapperfileName()
+{
+	return std::string("mapper-sdl2-") + DOSBOX_GetDetailedVersion() + ".map";
+}
 
 void MAPPER_TriggerEvent(const CEvent *event, const bool deactivation_state) {
 	assert(event);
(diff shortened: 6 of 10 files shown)


japsmits (approved):
Good work!

src/hardware/serialport/softmodem.cpp
```diff
@@ -609,8 +609,15 @@ void CSerialModem::DoCommand()
 		}
 		case 'I': // Some strings about firmware
 			switch (ScanNumber(scanbuf)) {
-			case 3: SendLine("DOSBox Emulated Modem Firmware V1.00"); break;
-			case 4: SendLine("Modem compiled for DOSBox version " VERSION); break;
+			case 3:
+				SendLine("DOSBox Emulated Modem Firmware V1.00");
+				break;
+			case 4: {
+				auto line = format_string(
```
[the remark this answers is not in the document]
johnnovak: Sure.

weirddan455:
This `DOSBOX_GetDetailedVersion()` seems to broken on MSVC, at least for a local build.  It has its own implementation in `src\platform\visualc` and it uses a macro that says "will be overwritten by CI" except that doesn't happen on a local build.  This function is used by the `dosbox.exe --version` argument and it just results in "version git" being reported:

version

Maybe this doesn't matter for 0.81.1 if you're planning on using MSYS2 for the Windows build but something to keep in mind before pushing this to `main`

>  Stuff still complains and the version number printed out in the intro screen is not broken. But it's tricky to test this properly; the true test will be to build a tagged build once this is merged.

Also, can you elaborate on what "stuff still complains" means?

johnnovak:
> Maybe this doesn't matter for 0.81.1 if you're planning on using MSYS2 for the Windows build but something to keep in mind before pushing this to `main`

Thanks @weirddan455, that's good info. Yeah I'll tinker with it, but the main point still stands: a single way to request the version info at runtime is preferable. I'll see what I can come up with for MSVC for local builds, even it means some ifdefs for Windows & MSVC.

> Also, can you elaborate on what "stuff still complains" means?

What, you don't have a mild case of dyxlessiaa??! 😆 
Wanted to write "stuff still compiles"... 😅

johnnovak:
For reference, using pre-build events could do the job for MSVC when compiling locally from the IDE:

https://stackoverflow.com/a/46448040

weirddan455:
I think there is a difference in intent between `VERSION` and `DOSBOX_GetDetailedVersion()`.

`VERSION` is also injected by meson but it just copies the thing we have hardcoded at the top of the meson build file.  I'm not sure why this one is breaking for you on Mac.

https://github.com/dosbox-staging/dosbox-staging/blob/6e18337f81348c55ce96ef442bfa427fc06cf93e/meson.build#L5

https://github.com/dosbox-staging/dosbox-staging/blob/6e18337f81348c55ce96ef442bfa427fc06cf93e/meson.build#L413

`DOSBOX_GetDetailedVersion()` gets the version information from `vcs_tag` which gets it from git in some way.  I guess it ends up getting a "pretty" name if the commit is a tagged release but on dev versions it probably spits out a hash or something.

johnnovak:
@weirddan455 But why would we want to have two mechanisms for getting the current version? That's asking for trouble and I'm always wasting hours syncing all this version crap up when creating the release artifacts... Try it once, and probably you'll come to the same conclusion. It's confusing and fiddly as fuck currently and I hate it 😩 

Ultimately, I want a single source of truth, and the best way seems to be [vcs_tag(https://mesonbuild.com/Reference-manual_functions.html#vcs_tag) which we're using for `DOSBOX_GetDetailedVersion()`. My idea is that the current version should be derived from the git tags as much as possible and all these manual version bumping and redundantly setting the version manually in various places should be eradicated as much as possible.

It's not 100% possible, e.g. when it's encoded into filenames (without making the whole build process even more complex), but we should do our best to simplify the current situation IMO. This is a step towards that direction.

If we can get away with _not_ setting `version` manually in the Meson `project` at all, that would be my preference. I remember experimenting with it and trying to figure out how the different version strings get generated and what's the relationship between them, but it's a confusing mess...

I remember looking up this thread for clarification, but that just got me way more confused...
https://github.com/mesonbuild/meson/issues/688

I find the Meson documentation quite lacking in many areas, and I vaguely remember something funky going on that `project_version()` actually returns the VCS tag, not the manually set `version` string... sometimes... somehow...  etc 🥶

weirddan455:
>  But why would we want to have two mechanisms for getting the current version?

You gave a perfect example of that here with the mapper file.  We probably don't want that to be creating/looking for a different file name every time we pull or push a new commit on our dev builds (assuming `vcs_tag()` works the way I think it does).

I agree a single source would be best but it seems we have different use-cases for the version string.

johnnovak:
> but on dev versions it probably spits out a hash or something.

Well, you can see for yourself in our local builds how `DOSBOX_GetDetailedVersion()` uses `vcs_tag()`.

Our last tag is `v0.82.0-alpha` in main. The `main` branch should always have `v0.x.y-alpha` as the last tag—it's our perpetual alpha for the next release. 

Then `vcs_tag()` appends the number of commits since the tag (351 in this case) and the git hash.

image


> You gave a perfect example of that here with the mapper file. We probably don't want that to be creating/looking for a different file name every time we pull or push a new commit on our dev builds (assuming `vcs_tag()` works the way I think it does).
> 
> I agree a single source would be best but it seems we have different use-cases for the version string.

No disagreements there, but I'd prefer a simple helper function in the code that chops off the postfixes from the full string returned by `vcs_tag()`. `DOSBOX_GetShortVersion` or something, then we can use that instead of `VERSION`.

It would transform `v0.82.0-alpha-351-12345` to `v0.82.0-alpha` or just `v0.82.0` and that's it (I don't think we even need the `alpha` postfix, but we could keep it).

weirddan455:
OK, a helper function would probably be the best way to go then so we can get both short and long version strings from the same source.

johnnovak:
> OK, a helper function would probably be the best way to go then so we can get both short and long version strings from the same source.

Yeah, I'll sort this out later in `main` when I forward-port it. Thanks for bringing the MSVC issue to my attention.

For 0.81.1, can I get a tick on this one please @weirddan455 ?

src/capture/image/png_writer.cpp
```diff
@@ -206,19 +206,22 @@ void PngWriter::WritePngInfo(const uint16_t width, const uint16_t height,
...
-	char software_value[] = CANONICAL_PROJECT_NAME " " VERSION;
+
+	auto software_value = format_string("%s %s",
+	                                    CANONICAL_PROJECT_NAME,
+	                                    DOSBOX_GetDetailedVersion());
 
 	texts[0].compression = PNG_TEXT_COMPRESSION_NONE;
 	texts[0].key         = static_cast<png_charp>(software_keyword);
-	texts[0].text        = static_cast<png_charp>(software_value);
+	texts[0].text        = static_cast<png_charp>(software_value.data());
```
weirddan455, line 216: More of this `.data()` vs `.c_str()` mess where the standard doesn't guarantee the former will be null terminated.  It looks like `format_string` is a new C++ 20 feature that returns `string_view` which is somewhat problematic when we need a C style string.

weirddan455 (approved):
Small concern about `string_view` problems that @japsmits brought up (and cleaned up some instances of) in a recent PR.  Probably it'll be fine in practice but 🤷

src/capture/image/png_writer.cpp
johnnovak: @weirddan455 Citation needed, `std::format` returns `std::string` according to this:
https://en.cppreference.com/w/cpp/utility/format/format

weirddan455: >  Class template std::basic_format_string wraps a [std::basic_string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) that will be used by formatting functions.

https://en.cppreference.com/w/cpp/utility/format/basic_format_string

Maybe that's not the return value.  These templated things are hard to read.  If you get a `std::string` back, you should be able to use `.c_str()` and avoid any ambiguity.

johnnovak: > > Class template std::basic_format_string wraps a [std::basic_string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) that will be used by formatting functions.
> 
> https://en.cppreference.com/w/cpp/utility/format/basic_format_string
> 
> Maybe that's not the return value. These templated things are hard to read. If you get a `std::string` back, you should be able to use `.c_str()` and avoid any ambiguity.

Not in this case because `c_str()` returns a `const char *` but `data()` returns a `char *`. Then PVS-Studio complains about casting the `const`ness away, which I'd need to do to pass it to the C API that wants a `char *`.

When you have a `std::string`, using `data()` is a good solution as that returns a `char *`, without the `const`, and that's what I'm doing here.

weirddan455:
Ok, I looked up the `.data()` string stuff on cppreference.com.  It _used_ to be the case that`std::string.data()` was const and not guaranteed to be null terminated.  Now it's changed so that it's not const and _is_ null terminated.

Except `std::string_view.data()` is still not guaranteed to be null terminated and is const.  C++ standard stuff is so hard to keep track of...  Lots of inconsistencies.  But it looks like the code you wrote here is completely correct.

johnnovak:
> Ok, I looked up the `.data()` string stuff on cppreference.com. It _used_ to be the case that`std::string.data()` was const and not guaranteed to be null terminated. Now it's changed so that it's not const and _is_ null terminated.

Ah, yeah, I was aware of this just did not mention it above.

> Except `std::string_view.data()` is still not guaranteed to be null terminated and is const. C++ standard stuff is so hard to keep track of... Lots of inconsistencies. But it looks like the code you wrote here is completely correct.

Yeah it's weird... But kinda makes sense as string views are just pointer-pairs.

That's C++ for you 🤷🏻 

[My assessment](https://blog.johnnovak.net/2016/03/03/the-quest-for-a-programming-language-that-doesnt-suck-part-1/) of the language from 2016 (to be fair, it had improved a lot since then)



> Contrary to popular belief, C++ was well-known and feared among the ancient Greeks. The hellish abomination was commonly depicted as a fiendish, many-headed water serpent rising from the dark depths of the sea, each head representing a different programming paradigm totally incompatible with the rest of the language. The actual number of fully functional heads varies greatly and depends on the target platform, the version of the C++ standard being depicted, the compiler vendor, the exact shade of the compiler vendor’s CEO’s wife’s niece’s toe polish, and the favourite food of the chief compiler implementor’s pet baby wombat. (Illustration by [Ruth Taylor](http://ruthtaylor.nl/))

PR #3557, opened 2024-04-01
Drop Debian 10

johnnovak:
Backport from `main`.


shermp (approved):

johnnovak:
Merging, I think this is fine, just kklobe is currently on holiday so the Mac runner is probably not running on his laptop.

PR #3565, opened 2024-04-05
Fix not being able to set `oplmode = none` regression

johnnovak:
# Description

In 0.81.0, we no longer could disable OPL with `oplmode = none`. This is a regression compared to 0.80.1. The only way to disable OPL was to set `sbtype = none`, but that disabled the Sound Blaster as well.

This PR fixes that and also cleans up the `oplmode` and `sbtype` handling. Now we're using pure functions for the parsing, things are less tangled, etc.

OPL logging is also a bit improved.

# Manual testing

Tested setting `oplmode = none` and all sorts of `oplmode` and `sbtype` permutations, including `oplmode = auto`.

Confirmed that `oplmode` and `sbtype` always stay in sync with the actual settings in use.

# Checklist

I have:

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -2122,77 +2122,77 @@ static void SBLASTER_CallBack(uint32_t len)
 	}
 }
 
-SB_TYPES find_sbtype()
+static std::optional<SB_TYPES> parse_sbtype_pref(const std::string& pref)
 {
-	const auto sect = static_cast<Section_prop *>(control->GetSection("sblaster"));
-	assert(sect);
+	if (pref == "gb") {
+		return SBT_GB;
 
-	const std::string pref = sect->Get_string("sbtype");
-
-	// Default
-	auto sbtype = SB_TYPES::SBT_NONE;
-
-	// Newest to oldest
-	if (pref == "sb16") {
-		sbtype = SBT_16;
-	} else if (pref == "sbpro2") {
-		sbtype = SBT_PRO2;
-	} else if (pref == "sbpro1") {
-		sbtype = SBT_PRO1;
-	} else if (pref == "sb2") {
-		sbtype = SBT_2;
 	} else if (pref == "sb1") {
-		sbtype = SBT_1;
-	} else if (pref == "gb") {
-		sbtype = SBT_GB;
-	}
-	return sbtype;
-}
+		return SBT_1;
 
-OplMode find_oplmode()
-{
-	const auto sect = static_cast<Section_prop *>(control->GetSection("sblaster"));
-	assert(sect);
+	} else if (pref == "sb2") {
+		return SBT_2;
 
-	const std::string pref = sect->Get_string("oplmode");
+	} else if (pref == "sbpro1") {
+		return SBT_PRO1;
 
-	// Default
-	auto opl_mode = OplMode::None;
+	} else if (pref == "sbpro2") {
+		return SBT_PRO2;
 
-	// Newest to oldest
-	if (pref == "opl3gold") {
-		opl_mode = OplMode::Opl3Gold;
-	} else if (pref == "opl3") {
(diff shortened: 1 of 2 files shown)


src/hardware/sblaster.cpp
```diff
@@ -2254,9 +2249,41 @@ class SBLASTER final {
...
+
+			if (const auto maybe_oplmode = parse_oplmode_pref(pref, sb.type);
+			    maybe_oplmode) {
+				return *maybe_oplmode;
+			} else {
+				LOG_WARNING("OPL: Invalid 'oplmode' setting: '%s', using 'auto'",
+				            pref.c_str());
+
+				// Now it's guaranteed to succeed
+				const auto auto_oplmode = parse_oplmode_pref(pref,
```
japsmits, line 2279: pref is unchanged, so it will actually fail again, unless I'm missing something. Do you intend to set "oplmode" to "auto" in this case, followed by a call to `parse_opl_mode("auto", sb.type)`?

src/hardware/sblaster.cpp
```diff
@@ -2254,9 +2249,41 @@ class SBLASTER final {
...
+		sb.type = [&] {
+			const std::string pref = section->Get_string("sbtype");
+
+			if (const auto maybe_sbtype = parse_sbtype_pref(pref);
+			    maybe_sbtype) {
+				return *maybe_sbtype;
+			} else {
+				LOG_WARNING("OPL: Invalid 'sbtype' setting: '%s', using 'sb16'",
+				            pref.c_str());
+				return SBT_16;
```
japsmits, line 2263: Should we also explicitly set "sbtype" to "sb16" in this case?

src/hardware/sblaster.cpp
```diff
@@ -2122,69 +2122,64 @@ static void SBLASTER_CallBack(uint32_t len)
...
-	} else if (pref == "gb") {
-		sbtype = SBT_GB;
+		return SBT_PRO1;
+
+	} else if (pref == "sbpro2") {
+		return SBT_PRO2;
+
+	} else if (pref == "sb16") {
+		return SBT_16;
 	}
```
japsmits, line 2144: Should we add a call to parse_bool_setting to also support "none" here? Having no Sound/Game Blaster should still be an option.

```
	} else if (const auto maybe_bool = parse_bool_setting(pref); maybe_bool) {
		if (!*maybe_bool) {
			return SBT_NONE;
		}
	}
```

japsmits (changes requested):
Found a few loose ends, but in general I think this is an improvement.

src/hardware/sblaster.cpp
johnnovak: Oh, thanks for catching this. This would have introduced the same regression for `sbtype` that I'm fixing for `oplmode` 🤦🏻 Fixing 😅

src/hardware/sblaster.cpp
japsmits: Ah, that's good to know! 😄

src/hardware/sblaster.cpp
johnnovak: Uh... This actually works because of some really confusing and hacky interactions in `setup.cpp`. In the invalid case setting this path actually never runs... I'll leave tracing through it to you if you're _really_ interested; too complicated to explain 😄 

Anyway, you're absolutely right, if this _did_ work as one would logically hink, this would be a problem. Although this is dead code, fixing it as you suggested because I don't want to really on the hack, at least I want the correct implementation to be there. We will get rid of the legacy config handling some time and rewrite it properly, then this error handling logic will come in handy.

japsmits (approved):

johnnovak:
Thanks for the quick review @japsmits and for catching one very important _new_ regression 😅  (`sbtype = none | off`).

Issue #3570, opened 2024-04-06
Cannot set `output` at runtime


interloper98:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

v0.82.0-alpha-356-gf55f0 (latest `main` build)

### What Operating System are you using?

Windows 11

### Relevant hardware info

AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

In recent releases (0.80 and 0.81), I could set `output` at runtime - for example, to switch from OpenGL to `texturenb` and then back to `opengl`.

However when I try with recent builds from CI (latest `main`), it seems to ignore the `output texture` request.

### Steps to reproduce the behaviour.

Explain how to reproduce

1. Launch with defaults: `dosbox.exe --noprimaryconf --nolocalconf`

    Confirm OPENGL is being used:
    ```
    2024-04-06 12:49:23.367 | OPENGL: Vendor: ATI Technologies Inc.
    2024-04-06 12:49:23.367 | OPENGL: Version: 4.6.0 Compatibility Profile Context 24.2.1.240215
    2024-04-06 12:49:23.367 | OPENGL: GLSL version: 4.60
    2024-04-06 12:49:23.367 | OPENGL: NPOT textures supported
    2024-04-06 12:49:23.367 | RENDER: Using adaptive CRT shader based on the graphics standard of the video mode
    ... <snip> ...
    2024-04-06 12:49:23.594 | RENDER: Auto-switched to shader 'crt/vga-1080p'
    2024-04-06 12:49:23.594 | SDL: Using standard SDI (auto) display refresh rate of 60 Hz
    2024-04-06 12:49:23.611 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz throttled VFR,     scaled to 1067x800 pixels with 1:1.35 (20:27) pixel aspect ratio
    2024-04-06 12:49:23.611 | OPENGL: VSync state: off
    ```

2. In DOSBox Staging's console, enter: `output texture` and press enter.
    - Notice that no visible change has occurred (OpenGL and the CRT auto-shader are still active)
    - Notice that no new log messages have been printed

[...]

PR #3571, opened 2024-04-07
Fix 'output' and other 'sdl' settings not changeable in runtime

FeralChild64:
# Description

Fix regression in handling many `[sdl]` section values in runtime.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3570

# Manual testing

The following commands should work:
`Z:\CONFIG.COM -set output=<value>`
`Z:\CONFIG.COM -set window_titlebar=<value>`

# Checklist

I have:

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -3246,43 +3246,27 @@ static void MAPPER_Destroy(Section *sec) {
 	SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
 }
 
-static bool should_skip_bind_keys(const Section_prop* section)
+static bool should_skip_unchanged_titlebar(const Section_prop* section)
 {
-	// Filter out unneeded calls - we only need to execute MAPPER_BindKeys
-	// when one of the values have changed; we do not want it to be executed
-	// if 'window_titlebar' was set (even to the same value as it had) to
-	// avoid screen flicker
+	// Filter out unneeded calls - do not execute MAPPER_BindKeys if only
+	// window titlebar setting has changed, to avoid flicker
 
-	const std::vector<std::string> values = {
-	        section->Get_string("mapperfile"),
-	        section->Get_string("fullscreen"),
-	        section->Get_string("fullresolution"),
-	        section->Get_string("window_position"),
-	        std::to_string(section->Get_bool("window_decorations")),
-	        std::to_string(section->Get_int("transparency")),
-	        section->Get_string("vsync"),
-	        section->Get_string("presentation_mode"),
-	        section->Get_string("output"),
-	        section->Get_string("priority"),
-	};
+	const std::string curr_titlebar = section->Get_string("window_titlebar");
 
-	static bool first_time = true;
-	static std::vector<std::string> old_values = {};
-
-	if (!first_time && values == old_values) {
+	static std::optional<std::string> prev_titlebar = {};
+	if (prev_titlebar && curr_titlebar != *prev_titlebar) {
+		prev_titlebar = curr_titlebar;
 		return true;
 	}
 
-	first_time = false;
-	old_values = values;
-
+	prev_titlebar = curr_titlebar;
 	return false;
 }
 
 void MAPPER_BindKeys(Section* sec)
 {
 	const auto section = static_cast<const Section_prop*>(sec);
-	if (should_skip_bind_keys(section)) {
+	if (should_skip_unchanged_titlebar(section)) {
 		return;
 	}
 
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -3246,21 +3246,49 @@ static void MAPPER_Destroy(Section *sec) {
 	SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
 }
 
-void MAPPER_BindKeys(Section *sec)
+static bool should_skip_bind_keys(const Section_prop* section)
 {
-	// Get the mapper file set by the user
-	const auto section = static_cast<const Section_prop *>(sec);
-	const auto mapperfile_value = section->Get_string("mapperfile");
-	const auto property = section->Get_path("mapperfile");
-	
-	// Filter out unneeded calls
+	// Filter out unneeded calls - we only need to execute MAPPER_BindKeys
+	// when one of the values have changed; we do not want it to be executed
+	// if 'window_titlebar' was set (even to the same value as it had) to
+	// avoid screen flicker
+
+	const std::vector<std::string> values = {
+	        section->Get_string("mapperfile"),
+	        section->Get_string("fullscreen"),
+	        section->Get_string("fullresolution"),
+	        section->Get_string("window_position"),
+	        std::to_string(section->Get_bool("window_decorations")),
+	        std::to_string(section->Get_int("transparency")),
+	        section->Get_string("vsync"),
+	        section->Get_string("presentation_mode"),
+	        section->Get_string("output"),
+	        section->Get_string("priority"),
+	};
+
 	static bool first_time = true;
-	static std::string old_mapperfile_value = {};
-	if (!first_time && mapperfile_value == old_mapperfile_value) {
+	static std::vector<std::string> old_values = {};
+
+	if (!first_time && values == old_values) {
+		return true;
+	}
+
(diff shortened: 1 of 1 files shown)


src/gui/sdl_mapper.cpp
```diff
@@ -3246,21 +3246,49 @@ static void MAPPER_Destroy(Section *sec) {
...
-	// Get the mapper file set by the user
-	const auto section = static_cast<const Section_prop *>(sec);
-	const auto mapperfile_value = section->Get_string("mapperfile");
-	const auto property = section->Get_path("mapperfile");
-	
-	// Filter out unneeded calls
+	// Filter out unneeded calls - we only need to execute MAPPER_BindKeys
+	// when one of the values have changed; we do not want it to be executed
+	// if 'window_titlebar' was set (even to the same value as it had) to
+	// avoid screen flicker
```
johnnovak, line 3254: Dunno, that's a lot of effort for a little flicker. I'd say let it flicker, or better yet, can't we flip this around and detect if `window_titlebar` is the same of has changed? This seems rather hacky and we'll need to remember to maintain this list when we add/remove `[sdl]` settings.

Well, but then the whole `[sdl]` section handling and the whole init process is hacky as hell... So dunno... I guess you don't want to refactor the whole thing as part of this 😄 I'm just wondering if there's a cleaner solution with less maintenance overhead and room for error when `[sdl]` section settings are added or renamed.

weirddan455: Ideally the config system would check if the value is already set to what the user is trying to change it to and avoid re-loading the module if that is the case.  This should happen for all modules but `[sdl]` is the most obvious offender because it reloads the window and has a visual flicker.

Probably you don't want to dive into that for this regression fix though.

weirddan455: Actually, it looks like the command line utility is responsible for calling the init + destroy functions.  I believe this is the only way currently that settings can be changed at runtime.  You could try adding a check here to see if the values are the same.  I'd be a little worried about regressions though.  The config system is a fickle beast.

https://github.com/dosbox-staging/dosbox-staging/blob/f55f0ae3565c144549001fe65bcf439c315fa510/src/misc/programs.cpp#L843-L855

FeralChild64: For now I have simplified the condition to just check for _window_titlebar_ changes - at the expense of flicker if the value wasn't actually changed.

Right, I don't want to mess with the config system right now, or to start yet another serious rework.

src/gui/sdl_mapper.cpp
```diff
@@ -3246,21 +3246,37 @@ static void MAPPER_Destroy(Section *sec) {
...
-	// Get the mapper file set by the user
-	const auto section = static_cast<const Section_prop *>(sec);
-	const auto mapperfile_value = section->Get_string("mapperfile");
-	const auto property = section->Get_path("mapperfile");
-	
-	// Filter out unneeded calls
+	// Filter out unneeded calls - do not execute MAPPER_BindKeys if only
+	// window titlebar setting has changed, to avoid flicker
+
+	const std::string value = section->Get_string("window_titlebar");
```
japsmits, line 3254: Rename `value` to `current_window_titlebar`?

src/gui/sdl_mapper.cpp
```diff
@@ -3246,21 +3246,37 @@ static void MAPPER_Destroy(Section *sec) {
...
+
 	static bool first_time = true;
-	static std::string old_mapperfile_value = {};
-	if (!first_time && mapperfile_value == old_mapperfile_value) {
-		return;
+	static std::string old_value = {};
```
japsmits, line 3257: Rename `old_value` to `previous_window_titlebar`?

src/gui/sdl_mapper.cpp
```diff
@@ -3246,21 +3246,37 @@ static void MAPPER_Destroy(Section *sec) {
 	SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
 }
 
-void MAPPER_BindKeys(Section *sec)
+static bool should_skip_bind_keys(const Section_prop* section)
```
japsmits, line 3249: Rename to `should_skip_unchanged_window_titlebar`?

japsmits (commented):
Nice one! Just some naming issues, since I had to read this twice to see what it does.

src/gui/sdl_mapper.cpp
johnnovak: Thanks @FeralChild64

src/gui/sdl_mapper.cpp
japsmits, line 3257: Maybe change the type to std::optional<std::string>. Then you can ditch the boolean, and there is no way to accidentally access the string.

japsmits (changes requested):

src/gui/sdl_mapper.cpp
johnnovak: Or `curr_*` and `prev_*`; these are well-know abbreviations to cut down on the length.

johnnovak:
Thanks for the quick fix @FeralChild64 🎉 

## The good

`sdl` settings now can be changed at runtime (e.g., `output texture`, `transparency 50`).

## The bad

Changing `window_titlebar` changes the titlebar text, but the graphics become completely frozen (no further screen updates). The following gets logged:

image

However, only the graphics updates are disabled, Staging itself remains response. Simply entering and exiting the mapper restores the graphics updates, so try doing the same thing after changing the window titlebar that the mapper does on exit?

Only tested on macOS so far.

src/gui/sdl_mapper.cpp
FeralChild64: Renamed to `curr_` / `prev_`.

src/gui/sdl_mapper.cpp
FeralChild64: Good idea, changed.

src/gui/sdl_mapper.cpp
FeralChild64: Renamed to `curr_titlebar` - I don't like excessively long variable names, especially with 80-char long lines.

src/gui/sdl_mapper.cpp
FeralChild64: Renamed to `should_skip_unchanged_titlebar`.

FeralChild64:
> graphics become completely frozen

I can't reproduce this problem on Linux. In fact, it is even hard to believe, all that we do here is skipping the reaction on configuration change event. Unless something else is happening behind the scene, I'll try to find out.

src/gui/sdl_mapper.cpp
```diff
@@ -3246,21 +3246,33 @@ static void MAPPER_Destroy(Section *sec) {
...
-	const auto property = section->Get_path("mapperfile");
-	
-	// Filter out unneeded calls
-	static bool first_time = true;
-	static std::string old_mapperfile_value = {};
-	if (!first_time && mapperfile_value == old_mapperfile_value) {
+	// Filter out unneeded calls - do not execute MAPPER_BindKeys if only
+	// window titlebar setting has changed, to avoid flicker
+
+	const std::string curr_titleebar = section->Get_string("window_titlebar");
```
japsmits, line 3254:
```suggestion
	const std::string curr_titlebar = section->Get_string("window_titlebar");
```
Typo?

japsmits (approved):
I won't leave it hanging on a typo. 👍

johnnovak:
> > graphics become completely frozen
> 
> I can't reproduce this problem on Linux. In fact, it is even hard to believe, all that we do here is skipping the reaction on configuration change event. Unless something else is happening behind the scene, I'll try to find out.

Good news, this revised version works fine! 🎉 

Yeah, the SDL/gfx init stuff is beyond belief, comprehension, and the laws of logic 🥶 

You don't how the amount of bullshit and trace logging I had to go through to get the auto-switching shaders working reliably. Then when I thought was done, I had to fix things up on Windows... Then some people reported edge cases on Linux...

src/gui/sdl_mapper.cpp
FeralChild64: Fixed.

johnnovak (approved):
Works fine on Windows 10 and macOS, nice one @FeralChild64 🎉

Issue #3580, opened 2024-04-09
Spell It Plus! (1989) (Regression)


Python-Exoproject:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

I was testing games in eXoDOS and found that this game has an issue in Staging 81.0 and 81.1 where all the white has gone pink. I then tested 80.1 and found that the game displayed correctly there.

81.1
![image](https://github.com/dosbox-staging/dosbox-staging/assets/67406191/97ab23e5-2002-4e6e-8311-515576a690c7)

80.1
![image](https://github.com/dosbox-staging/dosbox-staging/assets/67406191/e9ef470e-5422-4ac0-821d-181f472002d0)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Load the game in Staging 81.1 or 81.0
2.
3.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

interloper98:
Surely an unlikely coincidence, but that similar shade of pink shows up in both the box cover a d floppy disk labels. Would be very interesting to get this tested on real hardware, if it happens to be a palette tweak. 

![6647650-spell-it-plus-dos-front-cover](https://github.com/dosbox-staging/dosbox-staging/assets/165201780/c866ee43-7de6-4f21-ac6e-6a8296c5e185)
![6647662-spell-it-plus-dos-media](https://github.com/dosbox-staging/dosbox-staging/assets/165201780/032f97e9-4219-4850-a980-05e86ca15f88)
![6646528-spell-it-plus-dos-media](https://github.com/dosbox-staging/dosbox-staging/assets/165201780/78fabf8f-d922-43f7-a900-d0096290b0f7)

johnnovak:
We just turned it into a lady 😎 🤘🏻 

Very interesting and I'm curious about the reason. Testing on hardware to find out how it should look.

johnnovak:
Yeah, this is a legit regression; no need to even test on hardware. Fixing.

First bad commit:

```
commit 7c0a39cd139a3f05f61638b4fc3d482ead9cf581 (HEAD)
Author: kcgen <kcgen@users.noreply.github.com>
Date:   Mon Apr 10 15:47:46 2023 -0700

[...]

PR #3581, opened 2024-04-10
Use legacy palette behaviour on `svga_paradise`

johnnovak:
# Description

Mimic the legacy `machine = vgaonly` palette behaviour when emulating the Paradise SVGA card, which is our closest substitute for it and for emulating a pure VGA card. This fixes the wrong colours appearing in some rare titles (e.g., **Spell It Plus**).

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3580

See code comments for further details and my reasoning for choosing this solution. Basically, detecting "true EGA" modes on VGA with our defaults is _far_ more important than supporting this bottom-of-the-barrel title. But now compatibility with this game is restored by using `svga_paradise`. See this commit's comment for further details: https://github.com/dosbox-staging/dosbox-staging/commit/1fdb907aff43424c024de41a50cc5f08a5ac154c

If it turns out we have similar palette-related regressions in a significant number of non-obscure titles (which seems _not_ to be the case at all so far), I might revisit my "true EGA" mode detection approach in the future. But I spent weeks getting it right, so I'm extremely reluctant to touch it. Therefore, until we have some compelling evidence against my detection method, this solution will do.

## Notes

Now, this might break a few games when using `svga_paradise`... E.g, Drakkhen will show wrong colours. But that game runs fine with `ega` or `svga_s3` 🤷🏻 Ultimately, the Paradise emulation seems incomplete anyway, there's lots of TODOs and comments about missing functionality in `svga_paradise.cpp`, so I'm fine "repurposing" it this way for weird edge cases instead of introducing lots of tweakable VGA related options into the config... Probably the palette emulation is not 100% accurate, but it kinda sorta works OK most of the time... until you start changing things around like I did 😅 

This approach also "segregates" workarounds to revert to the "legacy DOSBox behaviour" into `svga_paradise`, which is nice.

As a related note, I noticed palette-related regressions in some rare demos (e.g. The Good, The Bad & The Ugly) that we introduced after 0.76.0, but these are super hard to track down, and frankly, the vast majority of games don't do such freaky stuff to the VGA hardware like demos... So spending tens or hundreds of hours on extreme edge case regressions like this has a very low payoff. But be my guest to fix it and deal with further regressions resulting from the fix 😏  It's a super boring, fiddly, and thankless job.

# Manual testing

## Regression testing

"True EGA" mode detection continues to work with EGA titles on VGA (e.g., Secret of the Silver Blades, Wasteland, Drakkhen, etc.)

Some games that do weird things to the EGA palette will display wrong colours on `svga_paradise`, but we'll just need to live with that.

## Spell It Plus

Palette is fine when running with `--noprimaryconf --machine svga_paradise`.

image

image

## Regression tests

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_dac.cpp
+++ b/src/hardware/vga_dac.cpp
@@ -305,8 +305,11 @@ void VGA_DAC_CombineColor(const uint8_t palette_idx, const uint8_t color_idx)
 {
 	vga.dac.combine[palette_idx] = color_idx;
 
-	if (vga.mode != M_LIN8) {
-		// Used by copper demo; almost no video card seems to support it
+	// Mimic the legacy `machine = vgaonly` behaviour when emulating the
+	// Paradise card, which is our closest substitue for it. This fixes the
+	// wrong colours appearing in some rare titles (e.g., Spell It Plus).
+	//
+	if (svgaCard != SVGA_ParadisePVGA1A && vga.mode != M_LIN8) {
 		vga_dac_send_color(palette_idx, color_idx);
 	}
 }
--- a/src/ints/int10_modes.cpp
+++ b/src/ints/int10_modes.cpp
@@ -1684,7 +1684,37 @@ att_text16:
 	}
 	IO_Read(mono_mode ? 0x3ba : 0x3da);
 
+	auto setup_palette_registers = [&] {
+		for (uint8_t ct = 0; ct < ATT_REGS; ct++) {
+			IO_Write(0x3c0, ct);
+			IO_Write(0x3c0, att_data[ct]);
+		}
+
+		vga.config.pel_panning = 0;
+
+		// Disable palette access
+		IO_Write(0x3c0, 0x20);
+		IO_Write(0x3c0, 0x00);
+
+		// Reset PEL mask
+		IO_Write(0x3c6, 0xff);
+	};
+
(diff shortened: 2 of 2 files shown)


src/hardware/vga_dac.cpp
```diff
@@ -305,8 +305,11 @@ void VGA_DAC_CombineColor(const uint8_t palette_idx, const uint8_t color_idx)
 {
 	vga.dac.combine[palette_idx] = color_idx;
 
-	if (vga.mode != M_LIN8) {
-		// Used by copper demo; almost no video card seems to support it
+	// Mimic the legacy `machine = vgaonly` behaviour when emulating the
```
johnnovak, line 308: Removed the Copper demo comment as it seems to be outdated. The demo runs fine with `svga_paradise` and I haven't noticed any wrong colours.

weirddan455 (approved):
I'm not going down the VGA rabbit hole to try and understand this right now but I trust you.

johnnovak:
> I'm not going down the VGA rabbit hole to try and understand this right now but I trust you.

I'm flattered @weirddan455  😎 

Actually, I wasn't really happy with this solution after raising the PR yesterday, so I kept thinking about alternative methods.

I've been experimenting with something that leaves Paradise alone, so closing this PR for now.

PR #3579, opened 2024-04-09
Translations differ between platforms

japsmits:
# Description

Ensure all translatable strings are visible to the translator, regardless of system configuration used.

## Related issues

#3562

# Manual testing

Started DOSBox and tried the changed commands.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [ ] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #3591, opened 2024-04-11
Rework physical CDROM interface

### LINKED ISSUE #3548: Remove SDL CD dependency

_No response_

None

_No response_

I plan on doing this myself soon.  Just wanted to make a ticket to document what needs to be done.  The SDL CD library has problems and is unmaintained.  I would like to drop it from the codebase.

1. Write OS specific functions to scan the host for CD drives on startup and store them in a list.  This is the only functionality from SDL CD that Linux depends on.  This SDL CD function is also called once in `program_mount.cpp` and  once in`dos_mscdex.cpp`
2. Write a class implementing `CDROM_Interface` that uses the Win32 API.  This should read digital audio data from the CD drive and push it through the DOSBox mixer.  SDL CD bypasses our mixer.
3. Remove the `CDROM_Interface_SDL` class and remove the library from `src/libs/sdlcd`
4. If someone cares enough, they can write an implementation for Mac.  Apparently SDL CD has been broken on Macs for a long time and no one seems to mind.  Probably most Macs these days don't have CD drives.

_No response_

_No response_

- [X] Yes, I agree.

### LINKED ISSUE #3555: Buffer overflow on MSCDEX GetUPC function

_No response_

None

_No response_

### Relevant hardware info

_No response_

We're not doing correct bounds checking on this function call.  According to this MSCDEX documentation, the UPC code returned should be BCD encoded 4 bits per character: https://makbit.com/articles/mscdex.txt

```

UPC Code

UPCCode   DB   14        ; Control block code
          DB   ?         ; CONTROL and ADR byte
          DB   7 dup (?) ; UPC/EAN code
                         ; (last 4 bits are zero; the low-order nibble of ; byte 7)
          DB   ?         ; Zero
          DB   ?         ; Aframe

This function returns the UPC/EAN (Universal Product Code - BAR coding) for the disc. This
information is stored as a mode-2 (ADR=2) Q-channel entry. The UPC code is 13 successive BCD
digits (4 bits each) followed by 12 bits of zero. The last byte is the continuation of FRAME
in mode-1 though in the lead-in track (TNO=0) this byte is zero. If the CONTROL/ADR byte is
zero or if the 13 digits of UPC code are all zero, then either no catalog number was encoded
on the disc or it was missed by the device driver. If the command is not supported, then the
driver will return an error code of Unknown Command. If the command is supported but the disc
does not have a UPC Code recorded, then the driver will return an error code of Sector not
Found.
```

I'm not sure yet if the caller is matching up to this specification but it's providing an 8 byte buffer for the driver to fill:

https://github.com/dosbox-staging/dosbox-staging/blob/9d9aba304140d1863b80afaf57732cfde3f5fff5/src/dos/dos_mscdex.cpp#L1063-L1071

The Linux driver is filling 14 bytes because it is not converting from ASCII to BCD overflowing the buffer:

https://github.com/dosbox-staging/dosbox-staging/blob/9d9aba304140d1863b80afaf57732cfde3f5fff5/src/dos/cdrom_ioctl_linux.cpp#L67

The cdrom image driver does not do any conversion or bounds checking whatsoever:

https://github.com/dosbox-staging/dosbox-staging/blob/9d9aba304140d1863b80afaf57732cfde3f5fff5/src/dos/cdrom_image.cpp#L546

_No response_

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

- CD Audio from physical CDROMs is threaded now, eliminating latency spikes in the main thread
- No more dependence on the un-maintainted and buggy SDL_CD library
- Windows uses direct Win32 `DeviceIoControl` calls (equivalent to Linux `ioctl`'s)
- Windows CD Audio now goes through our mixer (same as on Linux)
- New class `CDROM_Interface_Physical` where the OS independent audio and threading code goes
- Fix to `GetUPC` function which was also a buffer overflow
- No need to use clunky `-usecd` and `listcd` mount parameters (these have been removed).  The underlying device is detected automatically from the mount point and falls back to "fake CDROM" directory mount if a test ioctl doesn't work.

## Related issues

Fixes #3548
Fixes #3555

# Manual testing

Tested on both Linux and Windows with real CD of Quake and virtual CD (using cdemu and DaemonTools) of Secret of Monkey Island CDROM edition.

# Checklist

I have:


src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -20,54 +20,46 @@
...
 
-	struct cdrom_mcn cdrom_mcn;
-	int ret = ioctl(cdrom_fd, CDROM_GET_MCN, &cdrom_mcn);
+	struct cdrom_mcn cdrom_mcn = {};
 
-	if (ret > 0) {
+	if (ioctl(cdrom_fd, CDROM_GET_MCN, &cdrom_mcn) == 0) {
 		attr = 0;
-		safe_strncpy(upc, (char *)cdrom_mcn.medium_catalog_number, 14);
+		upc = (const char *)cdrom_mcn.medium_catalog_number;
```
japsmits, line 58: Note that this is not perfectly safe (it fails when `medium_catalog_number` is not zero-terminated).

japsmits (changes requested):
Looks sound, nice work!

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -335,14 +343,26 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
 	if (mixer_channel) {
 		return;
 	}
 
-	std::string name = std::string(ChannelName::CdAudio) + "_" +
-	                   std::to_string(device_number);
+	bool found_unique_name = false;
+	std::string name;
+	for (int i = 0; i < 32; ++i) {
+		name = std::string(ChannelName::CdAudio) + "_" + std::to_string(i);
```
johnnovak: `const auto MaxCdAudioChannel = 32`

Plus it would be nice to include a comment explaining that we'll create `CDAUDIO_1`, `CDAUDIO_2` for every mounted CD-ROM image/drive or something.

src/dos/cdrom.h
```diff
@@ -102,8 +98,8 @@ class CDROM_Interface
 {
 public:
```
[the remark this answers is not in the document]
johnnovak: I find `[[nodiscard]]` too noisy, so maybe let's not go there.

src/dos/cdrom.h
```diff
@@ -164,6 +163,10 @@ class CDROM_Interface_SDL : public CDROM_Interface
 		return true;
 	}
 	bool LoadUnloadMedia(bool unload) override;
+	bool HasFullMSCDEXSupport() override
```
johnnovak: `HasFullMscdexSupport`

src/dos/dos_mscdex.cpp
```diff
@@ -299,6 +289,46 @@ int CMscdex::RemoveDrive(uint16_t _drive)
 	return 1;
 }
 
+static CDROM_Interface *create_cdrom_interface(const char *path, uint16_t num_drives)
```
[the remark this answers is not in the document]
johnnovak: In general, yes. Up to @weirddan455 whether he's up to it now. Perhaps he had enough of this PR already 😄

src/dos/drive_iso.cpp
```diff
@@ -192,7 +192,7 @@ int isoDrive::UpdateMscdex(char drive_letter, const char *path, uint8_t &sub_uni
 		CDROM_Interface *cdrom = new CDROM_Interface_Image(sub_unit);
 		char pathCopy[CROSS_LEN];
 		safe_strcpy(pathCopy, path);
-		if (!cdrom->SetDevice(pathCopy, 0)) {
+		if (!cdrom->SetDevice(pathCopy)) {
 			CDROM_Interface_Image::images[sub_unit] = oldCdrom;
 			delete cdrom;
 			return 3;
```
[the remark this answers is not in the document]
johnnovak: +1

src/dos/program_mount.cpp
```diff
@@ -404,8 +393,7 @@ void MOUNT::AddMessages() {
 	        "Mount a directory from the host OS to a drive letter.\n"
 	        "\n"
 	        "Usage:\n"
-	        "  [color=light-green]mount[reset] [color=white]DRIVE[reset] [color=light-cyan]DIRECTORY[reset] [-t TYPE] [-usecd #] [-freesize SIZE] [-label LABEL]\n"
-	        "  [color=light-green]mount[reset] -listcd / -cd (lists all detected CD-ROM drives and their numbers)\n"
```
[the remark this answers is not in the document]
johnnovak: Good point, ideally we'd keep them, but remove the documentation, and issue a warning when they're used.

But if it's a major pain to do that, we might just drop them.

src/dos/program_mount.cpp
```diff
@@ -417,23 +405,20 @@ void MOUNT::AddMessages() {
...
-	        "  - '-usecd ID' gives direct access to a CD-ROM drive.\n"
-	        "    This is needed for CD audio (only supported on Windows and Linux).\n"
-	        "    Run 'mount -cd' to find out the list of valid IDs.\n"
 	        "  - Additional options are described in the manual (README file, chapter 4).\n"
 	        "\n"
 	        "Examples:\n"
 #if defined(WIN32)
 	        "  [color=light-green]mount[reset] [color=white]C[reset] [color=light-cyan]C:\\dosgames[reset]\n"
-	        "  [color=light-green]mount[reset] [color=white]D[reset] [color=light-cyan]D:\\[reset] -t cdrom -usecd 0\n"
+	        "  [color=light-green]mount[reset] [color=white]D[reset] [color=light-cyan]D:\\[reset] -t cdrom\n"
```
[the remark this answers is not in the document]
johnnovak: Yeah, platform-specific paths should be ifdef'd here.

src/dos/program_mount.cpp
```diff
@@ -417,23 +405,20 @@ void MOUNT::AddMessages() {
...
 	        "  [color=light-green]mount[reset] [color=white]C[reset] [color=light-cyan]my_savegame_files[reset] -t overlay\n"
 #elif defined(MACOSX)
 	        "  [color=light-green]mount[reset] [color=white]C[reset] [color=light-cyan]~/dosgames[reset]\n"
-	        "  [color=light-green]mount[reset] [color=white]D[reset] [color=light-cyan]\"/Volumes/Game CD\"[reset] -t cdrom -usecd 0\n"
+	        "  [color=light-green]mount[reset] [color=white]D[reset] [color=light-cyan]\"/Volumes/Game CD\"[reset] -t cdrom\n"
```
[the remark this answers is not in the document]
johnnovak: Yeah, platform-specific paths should be ifdef'd here.

src/dos/program_mount.cpp
```diff
@@ -417,23 +405,20 @@ void MOUNT::AddMessages() {
...
 	        "  [color=light-green]mount[reset] [color=white]C[reset] [color=light-cyan]my_savegame_files[reset] -t overlay\n"
 #else
 	        "  [color=light-green]mount[reset] [color=white]C[reset] [color=light-cyan]~/dosgames[reset]\n"
-	        "  [color=light-green]mount[reset] [color=white]D[reset] [color=light-cyan]\"/media/USERNAME/Game CD\"[reset] -t cdrom -usecd 0\n"
+	        "  [color=light-green]mount[reset] [color=white]D[reset] [color=light-cyan]\"/media/USERNAME/Game CD\"[reset] -t cdrom\n"
```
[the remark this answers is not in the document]
johnnovak: Yeah, platform-specific paths should be ifdef'd here.

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Using something like `safe_strncpy` seems safer, yeah.

src/dos/dos_mscdex.cpp
```diff
@@ -1072,10 +1072,24 @@ static uint16_t MSCDEX_IOCTL_Input(PhysPt buffer,uint8_t drive_unit) {
...
 					uint8_t attr = 0;
-					char upc[8] = {};
-					mscdex->GetUPC(drive_unit,attr,&upc[0]);
+					std::string upc = {};
+					mscdex->GetUPC(drive_unit,attr,upc);
 					mem_writeb(buffer+1,attr);
-					for (int i=0; i<7; i++) mem_writeb(buffer+2+i,upc[i]);
+					size_t upc_index = 0;
+					// Convert from ASCII to BCD encoding
+					for (int i = 0; i < 7; i++) {
```
johnnovak, line 1080: Extract to `ascii_to_bcd` util function in `math_utils.h`, then add some unit tests?

src/dos/dos_mscdex.cpp
```diff
@@ -1062,10 +1072,24 @@ static uint16_t MSCDEX_IOCTL_Input(PhysPt buffer,uint8_t drive_unit) {
...
+							high_nibble = upc[upc_index] - '0';
+						}
+						++upc_index;
+						uint8_t low_nibble = 0;
+						if (upc_index < upc.size()) {
+							low_nibble = upc[upc_index] - '0';
+						}
+						++upc_index;
+						mem_writeb(buffer + 2 + i, (high_nibble << 4) | (low_nibble & 0xF));
+					}
```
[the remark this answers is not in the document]
johnnovak: Yep, as per my above comment.

src/misc/rwqueue.cpp
```diff
@@ -45,21 +45,37 @@ size_t RWQueue<T>::Size()
...
 	has_items.notify_all();
 	has_room.notify_all();
 }
 
+template <typename T>
+void RWQueue<T>::Clear()
+{
+	mutex.lock();
+	queue.clear();
+	mutex.unlock();
```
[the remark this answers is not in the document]
johnnovak: That's preferable, I agree 👍🏻

src/dos/cdrom.cpp
```diff
@@ -261,3 +262,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+		}
+
+		// Copy these to local variables so that we can release the mutex lock.
+		const uint32_t local_sectors_remaining = sectors_remaining;
+		const uint32_t local_sector_start = current_sector;
+		lock.unlock();
+
+		// Slow call to read from the CDROM drive. Avoid holding any locks
+		const std::vector<int16_t> audio_samples = ReadAudio(local_sector_start, local_sectors_remaining);
+		const size_t num_frames = audio_samples.size() / 2;
```
johnnovak, line 119: 🥇 for the correct use of frames vs samples.

Maybe do `constexpr auto NumSamplesPerStereoFrame = 2`.

src/dos/cdrom.cpp
```diff
@@ -261,3 +262,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+		const size_t sectors_read = audio_samples.size() / SAMPLES_PER_REDBOOK_FRAME;
+
+		// Convert from int16_t to float before enqueuing
+		std::vector<AudioFrame> audio_frames;
+		for (size_t i = 0; i < num_frames; ++i) {
+			const int16_t left = host_to_le16(audio_samples[i * 2]);
+			const int16_t right = host_to_le16(audio_samples[(i * 2) + 1]);
```
johnnovak, line 126: Using NumSamplesPerFrame instead of 2 would be more readable.

johnnovak: Nice that you did not forget about the correct endianness handling 💯

src/dos/cdrom.cpp
```diff
@@ -261,3 +262,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+			audio_frames.emplace_back(left, right);
+		}
+
+		queue.BulkEnqueue(audio_frames, audio_frames.size());
+
+		lock.lock();
+		if (mixer_channel && !is_paused && sectors_remaining > 0 && queue.IsRunning()) {
+			mixer_channel->Enable(true);
+		}
+		if (sectors_remaining == local_sectors_remaining && current_sector == local_sector_start) {
```
johnnovak, line 136: Is `>=` a bit safer perhaps?

src/dos/cdrom.cpp
```diff
@@ -261,3 +262,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+
+void CDROM_Interface_Physical::InitAudio()
+{
+	if (mixer_channel) {
+		return;
+	}
+
+	bool found_unique_name = false;
+	std::string name;
+	for (int i = 0; i < 32; ++i) {
```
johnnovak, line 151: Could use the `MaxCdAudioChannels` constant here I asked for a couple of commits earlier 😄

src/dos/cdrom.cpp
```diff
@@ -261,3 +262,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+	}
+
+	std::vector<AudioFrame> audio_frames;
+	const size_t num_frames = std::min(static_cast<size_t>(requested_frames), queue_size);
+	if (num_frames > 0) {
+		queue.BulkDequeue(audio_frames, num_frames);
+	}
+
+	audio_frames.resize(requested_frames);
+	mixer_channel->AddSamples_sfloat(audio_frames.size(), &audio_frames[0][0]);
```
johnnovak, line 193: Nice, floats all the way 😎

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -343,173 +320,40 @@ bool CDROM_Interface_Ioctl::ReadSectorsHost([[maybe_unused]] void *buffer,
...
-	                          std::placeholders::_1);
-	mixer_channel = MIXER_AddChannel(callback,
-	                                 REDBOOK_PCM_FRAMES_PER_SECOND,
-	                                 name.c_str(),
-	                                 {ChannelFeature::Stereo,
-	                                  ChannelFeature::DigitalAudio});
 }
 
-void CDROM_Interface_Ioctl::CdAudioCallback(const uint16_t requested_frames)
+std::vector<int16_t> CDROM_Interface_Ioctl::ReadAudio(uint32_t sector, uint32_t num_frames)
```
johnnovak, line 335: const args?

src/dos/cdrom.h
```diff
@@ -400,62 +352,100 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
 	bool IsOpen() const;
 	bool Open(const char* device_name);
+	std::vector<int16_t> ReadAudio(uint32_t sector, uint32_t num_frames) override;
+
+	int cdrom_fd = -1;
+};
+
+#elif defined(WIN32)
+
+#include <Windows.h>
```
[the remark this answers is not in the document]
johnnovak: Doesn't really matter, but lowercase `W` is used everywhere else.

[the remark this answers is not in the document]
johnnovak: > Windows.h is a tricky beast in my experience. I would move its use to the cpp-file and create specific subclasses (PIMPL-like) for the Win-specific details.

Probably overkill as a bunch of header files are already including `windows.h` and nothing has blow up yet :) Just grep for `windows.h` and see for yourself.

src/dos/cdrom.h
```diff
@@ -400,62 +352,100 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
-	mixer_channel_t mixer_channel         = nullptr;
-	std::vector<int16_t> input_buffer     = {};
-	std::vector<AudioFrame> output_buffer = {};
-	size_t input_buffer_position          = 0;
-	size_t input_buffer_samples           = 0;
-	int current_sector                    = 0;
-	int sectors_remaining                 = 0;
-	bool is_playing                       = false;
-	bool is_paused                        = false;
+	HANDLE cdrom_handle                   = INVALID_HANDLE_VALUE;
```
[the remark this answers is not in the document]
johnnovak: Good tip, but we should only do it if we need to because it's a complication. If it compiles, then this is fine IMO.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#if defined(WIN32)
+
+#include "cdrom.h"
+
+#include <Windows.h>
+#include <Ntddcdrm.h>
```
johnnovak, line 24: minor: lowercase

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+	}
+	if (path[1] != ':') {
+		return 0;
+	}
+	const char drive_letter = toupper(path[0]);
+	if (drive_letter >= 'A' && drive_letter <= 'Z') {
+		return drive_letter;
+	}
+	return 0;
+}
```
johnnovak, line 40: Should go into `support.h`, next to `drive_index` and `drive_letter`.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+	}
+	cdrom_handle = INVALID_HANDLE_VALUE;
+}
+
+bool CDROM_Interface_Win32::Open(const char drive_letter)
+{
+	const std::string device_path = std::string("\\\\.\\") + drive_letter + ':';
+
+	HANDLE device = CreateFileA(device_path.c_str(), GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, OPEN_EXISTING, 0, NULL);
+	if (device == INVALID_HANDLE_VALUE) {
```
johnnovak, line 60: Lots of long lines in this file; just do a clang reformat.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+		return false;
+	}
+
+	// Test to make sure this device is a CDROM drive
+	CDROM_TOC toc = {};
+	if (!DeviceIoControl(device, IOCTL_CDROM_READ_TOC, NULL, 0, &toc, sizeof(toc), NULL, NULL)) {
```
johnnovak, line 66: I have zero idea what the args mean... Consider extracting them to const vars.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+	}
+	if (!data.MediaCatalog.Mcval) {
+		return false;
+	}
+	attr = 0;
+	upc = (const char *)data.MediaCatalog.MediaCatalog;
+	return true;
+}
+
+bool CDROM_Interface_Win32::GetAudioTracks(uint8_t& stTrack, uint8_t& end, TMSF& leadOut)
```
johnnovak, line 110: const

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+		return false;
+	}
+	if (!Open(drive_letter)) {
+		return false;
+	}
+	InitAudio();
+	return true;
+}
+
+bool CDROM_Interface_Win32::GetUPC(unsigned char& attr, std::string& upc)
```
johnnovak, line 91: const

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+	}
+	stTrack = toc.FirstTrack;
+	end     = toc.LastTrack;
+	leadOut.min = toc.TrackData[toc.LastTrack].Address[1];
+	leadOut.sec = toc.TrackData[toc.LastTrack].Address[2];
+	leadOut.fr  = toc.TrackData[toc.LastTrack].Address[3];
+	return true;
+}
+
+bool CDROM_Interface_Win32::GetAudioTrackInfo(uint8_t track, TMSF& start, unsigned char& attr)
```
johnnovak, line 130: const

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+	}
+	start.min = toc.TrackData[index].Address[1];
+	start.sec = toc.TrackData[index].Address[2];
+	start.fr  = toc.TrackData[index].Address[3];
+	attr = (toc.TrackData[index].Control << 4) | toc.TrackData[index].Adr;
+	return true;
+}
+
+bool CDROM_Interface_Win32::GetAudioSub(unsigned char& attr, unsigned char& track,
+                                        unsigned char& index, TMSF& relPos, TMSF& absPos)
```
johnnovak, line 151: `rel_pos`, `abs_pos` etc. as per our conventions. Same comment for all arg/var names in the file.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#if defined(WIN32)
+
+#include "cdrom.h"
+
+#include <Windows.h>
+#include <Ntddcdrm.h>
+
```
johnnovak, line 25: All new code should have `CHECK_NARROWING();` after the include block.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,229 @@
...
+std::vector<int16_t> CDROM_Interface_Win32::ReadAudio(uint32_t sector, uint32_t num_frames)
+{
+	constexpr uint32_t MaximumFramesPerCall = 55;
+	num_frames = std::min(num_frames, MaximumFramesPerCall);
+
+	std::vector<int16_t> audio_frames(num_frames * SAMPLES_PER_REDBOOK_FRAME);
+
+	RAW_READ_INFO read_info = {};
+	read_info.DiskOffset.LowPart = sector * 2048;
```
johnnovak, line 220: What's 2048?

johnnovak:
Very nice and welcome addition, @weirddan455 🥳  I'm done with the review; just a couple of comments. Routing the audio through the mixer properly is so much nicer! 🏅 

Thinking ahead a bit, would you be able to also include a short user-facing description of what we should mention about these enhancements in the release notes? It can be a rough draft, I will then polish it up. E.g. mentioning the mixer channels are named `CDAUDIO_1`, etc, that CD Audio is now routed through the mixer, and so on. All that detail because the alternative is that I'll need to read the PR again come release time, and I want to streamline my process as much as possible 😄 Thanks, man!

src/dos/cdrom_win32.cpp
japsmits: These are output parameters. 😄 But we could return an std::optional<struct> instead of using output parameters. That also ensures the caller looks at the return value.

src/dos/cdrom_win32.cpp
japsmits: These are output parameters. 😄 But we could return an std::optional<struct> instead of using output parameters. That also ensures the caller looks at the return value.

src/dos/cdrom_win32.cpp
japsmits: `track` could be. `start` and `attr` are output parameters.

japsmits:
One thing that popped into my mind: I see code for Windows and Linux, but is MacOSX in the equation somewhere? Since I don't own a Mac, I'm not hindered by any knowledge here. 😉

src/dos/cdrom_win32.cpp
johnnovak: Ah, okay 😄

johnnovak:
> One thing that popped into my mind: I see code for Windows and Linux, but is MacOSX in the equation somewhere? Since I don't own a Mac, I'm not hindered by any knowledge here. 😉

@weirddan455 doesn't have a Mac. Me and @kklobe do, but I personally don't have external CD-ROM drives for it, neither am I interested in adding support as it's quite a niche feature. Not many people have CD-ROM drives anymore. 

So adding macOS support will be a future task for some other guy who's interested enough ;) Generally, we try to achieve feature parity across platforms; this is one of the very few outliers but it's kinda "tolerated" as it's a niche feature.

src/dos/cdrom_ioctl_linux.cpp
japsmits: I would suggest adding the following functions to string_utils.h and use one of these:
```
template<size_t N>
std::string safe_tostring(char (&str)[N]) noexcept
{
	return std::string(str, safe_strlen(str));
}

std::string safe_tostring(const char* str, const std::size_t maxlen) noexcept
{
	return std::string(str, strnlen(str, maxlen));
}

```

src/dos/dos_mscdex.cpp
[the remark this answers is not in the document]
weirddan455: I try to use smart pointers in new code.  The caller for this function is legacy code storing raw pointers in an array though so I don't really see the point.  Someone would need to re-factor that data structure to get an advantage and I think that's out of scope for this PR.

src/dos/program_mount.cpp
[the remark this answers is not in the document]
weirddan455: Just tested `-usecd` and it gets silently ignored if you try to use it on this branch.  Technically it's backwards compatible assuming the auto-detect works correctly (and I'm pretty sure it does).  It always checks to see if a path is a CD drive when you mount with `-t cdrom`

src/dos/program_mount.cpp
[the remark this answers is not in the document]
weirddan455: This was inside an `ifdef` but I see what you did in #3579 and I'll rebase to fix.

src/dos/drive_iso.cpp
```diff
@@ -192,7 +192,7 @@ int isoDrive::UpdateMscdex(char drive_letter, const char *path, uint8_t &sub_uni
 		CDROM_Interface *cdrom = new CDROM_Interface_Image(sub_unit);
 		char pathCopy[CROSS_LEN];
 		safe_strcpy(pathCopy, path);
-		if (!cdrom->SetDevice(pathCopy, 0)) {
+		if (!cdrom->SetDevice(pathCopy)) {
 			CDROM_Interface_Image::images[sub_unit] = oldCdrom;
```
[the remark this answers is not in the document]
weirddan455: I don't know, I didn't touch this file except to remove a parameter from the `SetDevice` function.

src/dos/drive_iso.cpp
```diff
@@ -192,7 +192,7 @@ int isoDrive::UpdateMscdex(char drive_letter, const char *path, uint8_t &sub_uni
 		CDROM_Interface *cdrom = new CDROM_Interface_Image(sub_unit);
```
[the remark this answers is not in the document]
weirddan455: It's a C style array storing raw pointers.  It could probably be refactored but I'm going to say out of scope for this PR (because, again, I barely even looked at this file).

src/dos/drive_iso.cpp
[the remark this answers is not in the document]
weirddan455: It's some ugly error codes that gets later used as an output parameter to a constructor that can fail (yuck).

Yeah I don't like it either but I don't want to go re-factoring the .iso implementation just because I changed a single line here.

src/dos/dos_mscdex.cpp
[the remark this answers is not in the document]
johnnovak: I agree @weirddan455; job for another day. Let's not overload this PR (and you! 😅).

src/dos/cdrom.cpp
```diff
@@ -268,3 +93,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+	auto callback = std::bind(&CDROM_Interface_Physical::CdAudioCallback,
+	                          this,
+	                          std::placeholders::_1);
+	mixer_channel = MIXER_AddChannel(callback,
+	                                 REDBOOK_PCM_FRAMES_PER_SECOND,
+	                                 name.c_str(),
+	                                 {ChannelFeature::Stereo,
+	                                  ChannelFeature::DigitalAudio});
+
+	thread = std::thread(&CDROM_Interface_Physical::CdReaderLoop, this);
```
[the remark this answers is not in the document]
weirddan455: First I'm hearing of this feature:  https://en.cppreference.com/w/cpp/thread/jthread

It looks like a regular thread except it automatically joins on destruction?  There's also something about a `stop_source` which looks like an overly complicated replacement of my `should_quit` boolean :sweat_smile: 

Maybe this is worth using though.  Could you provide an example of how I might use it here?  I'm a bit confused because it says it has an internal `stop_source` but then in their example they're also passing one into the worker function:  https://en.cppreference.com/w/cpp/thread/stop_source

src/dos/cdrom.cpp
```diff
@@ -268,3 +93,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+	         ctrl.vol[0],
+	         ctrl.vol[1],
+	         ctrl.out[0],
+	         ctrl.out[1]);
+#endif
+}
+
+bool CDROM_Interface_Physical::PlayAudioSector(const uint32_t start, uint32_t len)
+{
+	mutex.lock();
```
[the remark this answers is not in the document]
weirddan455: Yeah, I can change these.

src/dos/cdrom.cpp
```diff
@@ -268,3 +93,212 @@ void CDROM_Interface::LagDriveResponse() const
...
+
+	thread = std::thread(&CDROM_Interface_Physical::CdReaderLoop, this);
+}
+
+void CDROM_Interface_Physical::CdAudioCallback(const uint16_t requested_frames)
+{
+	assert(requested_frames > 0);
+
+	const size_t queue_size = queue.Size();
+	if (queue_size == 0 && sectors_remaining == 0) {
```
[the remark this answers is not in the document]
weirddan455: No, that's correct.  It's checking that both the queue is empty _and_ the reader thread isn't going to add any more data to the queue.  In that case, the track is finished playing so we should stop the mixer callback.

weirddan455:
It should be easier after this PR to add CDROM support to additional platforms since I've brought the most complicated bits (mixer callback and threading) into a platform independent class.  SDL CD has Mac support but according to our resident Mac users, it doesn't work.  SDL CD's audio support was also broken on Linux so I added Linux-specific `ioctl` calls to fix it.

If someone wants to PR support for Mac or BSD or whatever platform after this, feel free.  Should be 200-300 lines of code give or take based on what Windows and Linux needed.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Sure, those look like some good helper functions.  Added and used.

johnnovak:
> If someone wants to PR support for Mac or BSD or whatever platform after this, feel free. Should be 200-300 lines of code give or take based on what Windows and Linux needed.

I might take a stab at it if I can connect my USB DVD writer to my MacBook somehow... 😆

weirddan455:
I haven't addressed all code comments yet.  I'll get to more tomorrow.  @johnnovak did you bring MSYS2 CI back?  It looks like I'm failing there:  ` ld.lld: error: undefined symbol: vtable for CDROM_Interface_Win32`

I didn't test that build (tested with MSVC on Windows).  Probably my `if defined(WIN32)` isn't working correctly or something.

src/dos/cdrom.h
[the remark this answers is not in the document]
weirddan455: Changed to lowercase.  Tempeted to change to `#include <wINdoWs.H>`

johnnovak:
> I haven't addressed all code comments yet. I'll get to more tomorrow.

No rush man 👍🏻 

> @johnnovak did you bring MSYS2 CI back? It looks like I'm failing there: ` ld.lld: error: undefined symbol: vtable for CDROM_Interface_Win32`

Ignore the MSYS2 builds on `main` and branches off `main`. I just periodically enable MSYS2 and disable MSVC so I can get the `release/0.81.x` branch building in preparation for the 0.81.1 release. Then I sometimes forget it like that for hours or a day 😄 Resetting now.

> I didn't test that build (tested with MSVC on Windows). Probably my `if defined(WIN32)` isn't working correctly or something.

Yeah, ignore MSYS2; only test on MSVC for `main`.

src/misc/rwqueue.cpp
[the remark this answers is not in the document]
weirddan455: Changed

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom.h
weirddan455: Fixed

src/dos/dos_mscdex.cpp
weirddan455: Those existing functions are not suitable for this use-case.  They take in a decimal rather than a string and they're hard-coded to 4 digits.  I extracted this into a separate function with a unit test though.

johnnovak: Cool, I haven't examined them too deeply, I admit.. :)

src/dos/program_mount.cpp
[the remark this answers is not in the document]
johnnovak: That's certainly good enough for this rarely used niche feature @weirddan455.

src/dos/drive_iso.cpp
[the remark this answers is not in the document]
johnnovak: > Yeah I don't like it either but I don't want to go re-factoring the .iso implementation just because I changed a single line here.

Fair enough.

src/dos/cdrom.cpp
weirddan455: No, that would be wrong.  This is checking to see if the member variables were changed from the main thread while the worker thread had released the mutex lock during reading/enqueuing (ex. the game wants to switch tracks).

This is there to avoid a race condition.  We don't want to modify this variable unless we know it has not changed.

I had a comment explaining this but I must have removed it while I was fixing stuff up.  I'll add one back.

src/dos/cdrom.cpp
weirddan455: Fixed

src/dos/cdrom.cpp
weirddan455: Fixed

src/dos/cdrom.cpp
weirddan455: This is the same code moved from the Linux class to the platform independent class.  I added that constant the other day.

src/dos/cdrom_win32.cpp
weirddan455: This is what the base class `CDROM_Interface` uses.  I would have to refactor everything in Linux, Win32 and CDROM_Image to make it consistent.

src/dos/cdrom_ioctl_linux.cpp
weirddan455: Fixed

src/dos/cdrom_win32.cpp
weirddan455: Fixed

src/dos/cdrom_win32.cpp
weirddan455: Moved

src/dos/cdrom_win32.cpp
weirddan455: Fixed

src/dos/cdrom_win32.cpp
weirddan455: Added

src/dos/cdrom_win32.cpp
weirddan455: Ok, I'll boot into Windows here in a bit and clean these up.

src/dos/cdrom.cpp
johnnovak: Cool, yeah some explanation would help.

src/dos/cdrom_win32.cpp
weirddan455: Extracted these into separate variables.  Hopefully this is easier for you to read... Microsoft does like making function that take in like 10 parameters sometimes.

src/dos/cdrom_win32.cpp
weirddan455: Reformatted

src/dos/cdrom.cpp
weirddan455: Comment added

src/dos/cdrom.cpp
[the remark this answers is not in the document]
weirddan455: Ok, this is the last code comment.  @johnnovak do you have an opinion on this?  I'm just not sure if I understand the benefit to using this feature or how to implement it correctly.

I think maybe it saves a single `thread.join()` call in the destructor, although I believe we still have to stop the `RWQueue` and do other cleanup.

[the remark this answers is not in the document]
japsmits: I'm not against leaving it like this. From what I understood, jthread is easier on edge-cases like exceptions and other threads stopping, but I'm don't think they are a real issue here. Explicit cleanup might be better here, especially if there are still other things to clean up.

We can always keep it in the back of our minds and/or experiment with them later. This comment originated from me toying around with std::thread and getting the remark from automatic code review that I should use std::jthread instead, because it's safer. But then again, if you use std::thread well, it should not be much of an issue.

japsmits (approved):
Looking good! 😄

src/dos/cdrom.cpp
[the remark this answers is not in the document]
johnnovak: @weirddan455 Because we need to do some other manual cleanup/synchronisation anyway, I think it's fine as it is.

Thanks for the tip @japsmits, maybe we'll use jthread next time 😄

src/dos/dos_mscdex.cpp
```diff
@@ -666,10 +667,10 @@ bool CMscdex::GetFileName(uint16_t drive, uint16_t pos, PhysPt data) {
 	return success; 
 }
 
-bool CMscdex::GetUPC(uint8_t subUnit, uint8_t& attr, char* upc)
+bool CMscdex::GetUPC(uint8_t subUnit, uint8_t& attr, std::string& upc)
```
johnnovak, line 670: Minor: GetUpc, sub_unit, const args

src/dos/cdrom.h
```diff
@@ -410,8 +412,37 @@ class CDROM_Interface_Image final : public CDROM_Interface
 	static int           refCount;
 };
 
+class CDROM_Interface_Physical : public CDROM_Interface {
+public:
+	~CDROM_Interface_Physical() override;
+
+	bool GetAudioStatus(bool& playing, bool& pause) override;
+	bool PlayAudioSector(const uint32_t start, uint32_t len) override;
```
johnnovak, line 360: Probably better to remove `const` for primitive types (non-reference & non-pointer) in function declarations and only keep them in function definitions.

src/dos/cdrom.h
```diff
@@ -410,8 +412,37 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	bool PauseAudio(bool resume) override;
+	bool StopAudio() override;
+	void ChannelControl(TCtrl ctrl) override;
+
+protected:
+	void InitAudio();
+
+private:
+	virtual std::vector<int16_t> ReadAudio(const uint32_t sector, const uint32_t frames_requested) = 0;
```
johnnovak, line 369: Probably better to remove `const` for primitive types (non-reference & non-pointer) in function declarations and only keep them in function definitions.

src/dos/cdrom.h
```diff
@@ -410,8 +412,37 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
+	void CdAudioCallback(const uint16_t requested_frames);
+	void CdReaderLoop();
+
+	mixer_channel_t mixer_channel  = {};
+	std::thread thread             = {};
+	std::mutex mutex               = {};
+	std::condition_variable waiter = {};
+	RWQueue<AudioFrame> queue        {220500};
```
johnnovak, line 377: Why this number?

src/misc/support.cpp
```diff
@@ -70,6 +70,22 @@ char drive_letter(uint8_t index)
 	return 'A' + index;
 }
 
+char get_drive_letter_from_path(const char* path)
+{
+	const size_t len = strlen(path);
+	if (len < 2 || len > 3) {
```
johnnovak, line 76: Why `len > 3`? Shouldn't this work for something like `C:\BLAH\STUFF.EXE` too? The name implies it should work for full paths.

src/dos/cdrom.h
```diff
@@ -472,7 +472,41 @@ class CDROM_Interface_Ioctl : public CDROM_Interface_Physical {
...
+	bool GetAudioTrackInfo(uint8_t track, TMSF& start, unsigned char& attr) override;
+	bool GetAudioSub(unsigned char& attr, unsigned char& track,
+	                 unsigned char& index, TMSF& relPos, TMSF& absPos) override;
+	bool GetMediaTrayStatus(bool& mediaPresent, bool& mediaChanged,
+	                        bool& trayOpen) override;
+	bool ReadSectors(PhysPt buffer, const bool raw, const uint32_t sector,
+	                 const uint16_t num) override;
+	bool ReadSectorsHost(void* buffer, bool raw, unsigned long sector,
+	                     unsigned long num) override;
+	bool LoadUnloadMedia(bool unload) override;
```
johnnovak, line 435: Probably best to remove `const` for non-reference and non-pointer arguments from the function _declarations_ (but keep them in the function _definitions_).

src/dos/cdrom.h
```diff
@@ -472,7 +472,41 @@ class CDROM_Interface_Ioctl : public CDROM_Interface_Physical {
...
+	{
+		return true;
+	}
+
+private:
+	std::vector<int16_t> ReadAudio(const uint32_t sector, const uint32_t frames_requested) override;
+	bool IsOpen() const;
+	bool Open(const char drive_letter);
+
+	HANDLE cdrom_handle                   = INVALID_HANDLE_VALUE;
```
johnnovak, line 446: Minor: formatting

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,283 @@
...
+{
+	const std::string device_path = std::string("\\\\.\\") + drive_letter + ':';
+
+	LPCSTR file_name = device_path.c_str();
+	DWORD desired_access = GENERIC_READ;
+	DWORD share_mode = FILE_SHARE_READ | FILE_SHARE_WRITE;
+	LPSECURITY_ATTRIBUTES securiy_attributes = NULL;
+	DWORD creation_disposition = OPEN_EXISTING;
+	DWORD flags = 0;
+	HANDLE template_file = NULL;
```
johnnovak: Nice 👍🏻 Could use a clang format though.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,283 @@
...
+	// Test to make sure this device is a CDROM drive
+	CDROM_TOC toc = {};
+
+	DWORD control_code = IOCTL_CDROM_READ_TOC;
+	LPVOID input_buffer = NULL;
+	DWORD input_buffer_size = 0;
+	LPVOID output_buffer = &toc;
+	DWORD output_buffer_size = sizeof(toc);
+	LPDWORD bytes_returned = NULL;
+	LPOVERLAPPED overlapped = NULL;
```
johnnovak: Nice 👍🏻 Could use a clang format though.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,283 @@
...
+	format.Format = IOCTL_CDROM_MEDIA_CATALOG;
+	SUB_Q_CHANNEL_DATA data        = {};
+
+	DWORD control_code = IOCTL_CDROM_READ_Q_CHANNEL;
+	LPVOID input_buffer = &format;
+	DWORD input_buffer_size = sizeof(format);
+	LPVOID output_buffer = &data;
+	DWORD output_buffer_size = sizeof(data);
+	LPDWORD bytes_returned = NULL;
+	LPOVERLAPPED overlapped = NULL;
```
johnnovak: Nice 👍🏻 Could use a clang format though.

src/dos/cdrom_win32.cpp
```diff
@@ -0,0 +1,283 @@
...
+	}
+	if (toc.LastTrack >= MAXIMUM_NUMBER_TRACKS) {
+		return false;
+	}
+	stTrack = toc.FirstTrack;
+	end     = toc.LastTrack;
+	leadOut.min = toc.TrackData[toc.LastTrack].Address[1];
+	leadOut.sec = toc.TrackData[toc.LastTrack].Address[2];
+	leadOut.fr  = toc.TrackData[toc.LastTrack].Address[3];
+	return true;
```
johnnovak, line 181: Clang format the whole file please.

src/misc/support.cpp
weirddan455: You asked me to extract this to a generic function and it probably shouldn't have been.  It's meant to fail if the user does not specify a root drive in the case of attempting to mount a CDROM drive.

src/dos/cdrom.h
weirddan455: It's 5 seconds of audio.  Expanded to `REDBOOK_PCM_FRAMES_PER_SECOND * 5`

src/dos/dos_mscdex.cpp
weirddan455: Added const

src/dos/cdrom.h
weirddan455: Removed const from all primitive type declarations in this file

src/misc/support.cpp
weirddan455: Or at least it shouldn't have been extracted without changes :smile: 

Moved the `> 3` length check to the calling code.

src/dos/cdrom.h
johnnovak: Sorry @weirddan455, can you drop that commit as per our recent discussion?

src/misc/support.cpp
johnnovak: > Moved the `> 3` length check to the calling code.

I think that's a nice solution to keep the helper generic 👍🏻

src/dos/cdrom.h
johnnovak: Sorry, can you drop the commit as per our recent discussion?

src/dos/cdrom.h
johnnovak: Sorry, can you drop the commit as per our recent discussion?

johnnovak (approved):
Approving @weirddan455, nice work! 🎉 Just a few very minor things, please merge once you're happy with it.

Would you be able to give the brief release notes draft a shot, too?

src/dos/cdrom.cpp
```diff
@@ -268,3 +93,217 @@ void CDROM_Interface::LagDriveResponse() const
...
+void CDROM_Interface_Physical::InitAudio()
+{
+	if (mixer_channel) {
+		return;
+	}
+
+	// Each audio channel must have a unique name.
+	// Append a number after the CDAUDIO name.
+	// This way we don't conflict with CDROM_Interface_Image or multiple drives.
+	constexpr int MaxCdAudioChannel = 32;
```
johnnovak, line 156: Ok, last one 😆 Consider moving this into `channel_names.h` and adding a comment above `CdAudio`, mentioning that we apply `_1`, `_2` etc. suffixes when we're dealing with multiple drives.

japsmits (approved):

src/dos/cdrom.h
weirddan455: Commit dropped

src/dos/cdrom.h
weirddan455: Commit dropped

src/dos/cdrom.h
weirddan455: Commit dropped

src/dos/cdrom.cpp
weirddan455: Done

weirddan455:
### Physical CDROM Enhancements

Physical CDROM devices will now be auto-detected when using the mount command.  No more need to use the additional `-usecd` parameter.  For example:

Windows: `mount d d:\ -t cdrom`
Linux: `mount d /mnt/cdrom -t cdrom`

Additionally, Windows CDROM audio now gets routed through the DOSBox mixer, bringing it to feature parity with Linux.

Lastly, reading audio data from the CDROM is now done in a separate thread which should elimante latency and stuttering issues.

johnnovak:
Great summary @weirddan455 , thanks a lot 👍🏻

Issue #3524, opened 2024-03-21
Gold of the Aztecs, The (1990)


Python-Exoproject:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

81.0

### What Operating System are you using?

None

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Another game with graphics flicker. Its pretty bad for this one, but its mostly during the intro if you let it run and if you slect the map in gameplay.

This time Daum doesnt improve things. Staging with a dos_rate of 140 helps while still being playable, but it takes dos_rate of 280 to fix it and that causes other issues like hiccups in the displaying of the graphics.

It would be interesting to add this to the games to check on hardware @johnnovak as Daum didnt fix it which means it might not be the same issue as #2950 #2949 #3244 #2905

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Load game, watch intro
2.
3.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

MX9000:
@Python-Exoproject doing more tests in the meantime, I found that by setting cycles=20000 the flickering disappears!

Python-Exoproject:
Great job @MX9000. 

I wouldnt think a game from 1990 would need that many cycles though. I have managed success with only 8500.

Looks like this one can be closed @johnnovak

johnnovak:
> Great job @MX9000.
> 
> I wouldnt think a game from 1990 would need that many cycles though. I have managed success with only 8500.
> 
> Looks like this one can be closed @johnnovak

@Python-Exoproject 
It is possible to completely get rid of the flicker with the new `vmem_delay` option.

See the details and the config to fix the game here: https://www.vogons.org/viewtopic.php?t=89650

Issue #3244, opened 2023-12-19
James Bond 007 - The Stealth Affair (1990) (AKA Operation Stealth)


Python-Exoproject:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

81.0a

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

This game has screen flicker when run using EGA or VGA graphics. MCGA graphics do not seem to be affected. There is another Delphine software game (Future Wars) that has the same issue with VGA graphics (and again MCGA seems to be ok).

Of note is that all dosbox forks suffer from this flicker *Except* Daum, so it has some special sauce not yet in any other fork. @kcgen did some initial testing and ruled out the direct3d output and confirmed its something that is part of Daum itself

If this can be resolved there are a number of games that will benefit (@kcgen raised 'Corncob 3D' and 'Hostage' as some) and is probably one of the last things Daum can do better than Staging

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Use eXoDOS copy of James Bond 007 - The Stealth Affair
2. Choose the EGA option from the choice menu
3. Under Daum as per current eXoDOS launch there is no flickering. Under Staging you can see the flickering from the first screen, around the 'OK' button.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
From dosbox.com:
https://www.dosbox.com/comp_list.php?showID=191&letter=F

> Flickering is caused by picking 'VGA' from the installation menu. Pick MCGA and the game will run fine without odd clipping problems or flickering.

Same comment in random Vogons threads.

Tried with DOSBox X, it has the same flicker issue just like you said @Python-Exoproject. Don't keep your hopes up, if neither the original team nor Jon figured this out, I don't think we know where to even start looking 😅 Video emulation is THE most complicated and most underdocumented part of the codebase, bar none... Even with the Daum sources, this is a bit of a finding a needle in a haystack situation.

kcgen:
Excellent reproduction, @johnnovak !

I can definitely see that flickering and judder/ghosting, similar to how the Xeen mouse works. 

This is really nice to see that DOSBox's S3 implementation matches the hardware, right down to these corner case behaviors.

johnnovak:
[...]

Issue #2950, opened 2023-09-21
Investigate playability issues with Corncob Deluxe (1994)


Torinde:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

None

### Context
Currently the eXoDOS project runs this game using [the DOSBox Daum fork](https://github.com/ykhwong/dosbox-svn-daum), which is understood to have specific emulation fixes not currently in upstream DOSBox (0.74.x and SVN) or active forks such as X, Pure, and Staging.

DOSBox Daum had 51 releases that included DOSBox SVN's commits ~r3600 through ~r3900, [spanning 2010 through 2015](http://ykhwong.x-y.net/).

### The goal of this ticket is to answer the following
Is Daum _still_ needed for this game?

* [ ]  Does DOSBox SVN ([download latest](http://www.emucr.com/2023/09/dosbox-svn-r4483.html)) play it OK?
* [ ]  Does DOSBox X ([download latest](https://dosbox-x.com/devel-build.html)) play it OK?
* [ ]  Does DOSBox Staging ([download latest](https://dosbox-staging.github.io/downloads/development-builds/)) play it OK?

If Daum plays it better:

* [ ]  What aspect is better?
* [ ]  Which Daum release was the **_first_** to fix it?
  Use these builds [here](http://ykhwong.x-y.net/subpages/old.php) to narrow it down.

### How you can help
1. Download the above lastest DOSBox SVN, X, and Staging builds.
2. Ensure you have [eXoDOS v5](https://www.retro-exo.com/exodos.html) (or eventually v6)
3. Find this game inside the eXoDOS installed folders
4. Run the game using the DOSBox SVN, X, and Staging builds.
   _If it plays fine (everything looks, sounds, and handles OK) then simply indicate as such. Your task is done!_

If SVN, X, and Staging all have problems, then try it with Daum:

1. Download the [first and last Daum releases](http://ykhwong.x-y.net/subpages/old.php).
    Note: _Daum had the most releases for Microsoft Windows, so having a Windows system (or VM) will help the most to answer he last question above._
2. Does the latest Daum release play it correctly? If so: summarize/describe how it's better than DOSBox SVN, X, and Staging (screenshots, video, audio records all help too).
3. Find the earliest (or oldest) Daum release that fixed it.
   
[...]

Issue #2905, opened 2023-09-16
Investigate playability issues with Hostage Rescue Mission (1989)


kcgen:
### Context

Currently the eXoDOS project runs this game using [the DOSBox Daum fork](https://github.com/ykhwong/dosbox-svn-daum), which is understood to have specific emulation fixes not currently in upstream DOSBox (0.74.x and SVN) or active forks such as X, Pure, and Staging.

DOSBox Daum had 51 releases that included DOSBox SVN's commits ~r3600 through ~r3900, [spanning 2010 through 2015](http://ykhwong.x-y.net/).

### The goal of this ticket is to answer the following

Is Daum _still_ needed for this game?
- [ ] Does DOSBox SVN ([download latest](http://www.emucr.com/2023/09/dosbox-svn-r4483.html)) play it OK?
- [ ] Does DOSBox X ([download latest](https://dosbox-x.com/)) play it OK?
- [ ] Does DOSBox Staging ([download latest](https://dosbox-staging.github.io/downloads/development-builds/)) play it OK?

If Daum plays it better:
- [ ] What aspect is better?
- [ ] Which Daum release was the **_first_** to fix it?
    Use these builds [here](http://ykhwong.x-y.net/subpages/old.php) to narrow it down.

### How you can help

 1. Download the above lastest DOSBox SVN, X, and Staging builds.
 1. Ensure you have [eXoDOS v5](https://www.retro-exo.com/exodos.html) (or eventually v6)
 1. Find this game inside the eXoDOS installed folders (look for: `HostageR/`)
 1. Run the game using the DOSBox SVN, X, and Staging builds.
    _If it plays fine (everything looks, sounds, and handles OK) then simply indicate as such. Your task is done!_

If SVN, X, and Staging all have problems, then try it with Daum:
 1. Download the [first and last Daum releases](http://ykhwong.x-y.net/subpages/old.php).
    Note: _Daum had the most releases for Microsoft Windows, so having a Windows system (or VM) will help the most to answer the last question above._
 1. Does the latest Daum release play it correctly? If so: summarize/describe how it's better than DOSBox SVN, X, and Staging (screenshots, video, audio records all help too).
 1. Find the earliest (or oldest) Daum release that fixed it.
    1. Run the Daum release half way between the known good release (where the game playing correctly) and bad release (where the game doesn't playing correctly) until you've narrowed it down to adjacent releases.
[...]

PR #3597, opened 2024-04-12
Port video memory I/O delay tweak from DOSBox-X

johnnovak:
# Description

Inspired by [this comment](https://github.com/dosbox-staging/dosbox-staging/issues/2905#issuecomment-2049181149) by @MX9000, I attempted to port the `vmemdelay` tweak over from DOSBox-X which went surprisingly easily.

What's even better news, it solves the flickering graphics issues in _all_ problematic games that @Python-Exoproject raised! 🥳 These games don't flicker in Daum, so looks like we've managed to discover Daum's "secret ingredient" 😄 

This makes complete sense as these are early VGA or late EGA games so the developers must have assumed slow video memory I/O speeds and coded everything else around that. These games _do flicker_ on my S3 card in my Pentium MMX, which is not too surprising as the S3 cards are one of the fastest late DOS era SVGA boards.

The original DOSBox-X feature has some auto settings, but I dropped that because it didn't do much for me. It basically defaulted to 450 ns on VGA and EGA, and a bit above 3000 ns on CGA. As you can see from my tests, 3000 ns is the magic number that fixes all these games—go below that and the flicker comes gradually back, go above it, it slows down the emulation too much. So I just introduced an `on` setting that sets this to 3000 ns, which seemed reasonable.

As `vmem_delay` emulates the effects of the slow video memory I/O on the CPU (so it effectively throttles the CPU a bit when video memory I/O is happening), this could "even out" the emulation speed in games that run too fast in certain game segments. I saw someone mention that a `vmem_delay` of around 6000 fixes the speed issues in Wing Commander I, but testing that is an exercise for another day ([more info](https://www.vogons.org/viewtopic.php?t=89650)).

Lastly, as this is a very clean patch and an opt-in feature (it's disabled by default), I think it's worthwhile and completely safe to backport it. It increases compatibility with a fair number of games, and eXoDOS would benefit from it and thus further their adoption of Staging.

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/issues/2950
- https://github.com/dosbox-staging/dosbox-staging/issues/2905
- https://github.com/dosbox-staging/dosbox-staging/issues/3244
- https://github.com/dosbox-staging/dosbox-staging/issues/3524

# Manual testing

Tested valid values, invalid values, `on`, `off`, and confirmed the defaulted value after an error is always synced back to `vmem_delay`.

Tested the following games and confirmed the flickering is completely gone with the below settings. Note that for some games `vmem_delay = on` alone is not enough; you also need to use a low enough `cycles` setting.

## Hostage Rescue Mission

```ini
[dosbox]
vmem_delay = on

[cpu]
cycles = 1500
```

## James Bond 007 - The Stealth Affair

VGA (not MCGA) and EGA are now working without flicker.

```ini
[dosbox]
vmem_delay = on

[cpu]
cycles = 3000
```

## Future Wars (floppy version)

The game needs a slow `cycles = 1000` setting as well to completely get rid of the flicker. This is not a problem as it runs at full speed even around 500 cycles (it was written in assembly, most likely).

```ini
[dosbox]
vmem_delay = on

[cpu]
cycles = 1000
```

## Gold of the Aztecs

```ini
[dosbox]
vmem_delay = on

[cpu]
cycles = 3000
```

## Quest for Glory 2

The vertical scrolling of the entire screen in the intro, which starts when the magic carpet leaves the top of the desert scene, is properly timed now (it is ultra-fast and barely noticeable without `vmem_delay`). 

```ini
[dosbox]
vmem_delay = on

[cpu]
cycles = 2400
```

## Corncob Deluxe

`vmemdelay = on` (= 3000) completely gets rid of the flicker in-game, but the credits scroller in the intro is a bit slow. A value of 2000 might be a good compromise at the expense of some in-game flicker.

```ini
[dosbox]
vmem_delay = on

[cpu]
cycles = 3000
```

# Checklist

I have:

### DIFF
--- a/include/vga.h
+++ b/include/vga.h
@@ -1057,9 +1057,11 @@ struct VgaType {
 
 	// Memory for fast (usually 16-colour) rendering,
 	// always twice as big as vmemsize
-	//
 	uint8_t* fastmem  = {};
 	uint32_t vmemsize = 0;
+
+	// How much delay to add to video memory I/O in nanoseconds
+	uint16_t vmem_delay_ns = 0;
 
 #ifdef VGA_KEEP_CHANGES
 	VgaChanges changes = {};
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -553,11 +553,12 @@ void DOSBOX_Init()
 
 	pstring = secprop->Add_string("dos_rate", when_idle, "default");
 	pstring->Set_help(
-	        "Customize the emulated video mode's frame rate, in Hz:\n"
-	        "  default:  The DOS video mode determines the rate (recommended; default).\n"
+	        "Customize the emulated video mode's frame rate.\n"
+	        "  default:  The DOS video mode determines the rate (default).\n"
 	        "  host:     Match the DOS rate to the host rate (see 'host_rate' setting).\n"
-	        "  <value>:  Sets the rate to an exact value, between 24.000 and 1000.000 (Hz).\n"
-	        "We recommend the 'default' rate; otherwise test and set on a per-game basis.");
+	        "  <value>:  Sets the rate to an exact value in between 24.000 and 1000.000 Hz.\n"
+	        "Note: We recommend the 'default' rate, otherwise test and set on a per-game\n"
+	        "      basis.");
 
 	pstring = secprop->Add_string("vesa_modes", only_at_start, "compatible");
 	pstring->Set_values({"compatible", "all", "halfline"});
@@ -568,9 +569,9 @@ void DOSBOX_Init()
 	        "               (default).\n"
 	        "  halfline:    Supports the low-resolution halfline VESA 2.0 mode used by\n"
 	        "               Extreme Assault. Use only if needed, as it's not S3 compatible.\n"
-	        "  all:         All modes for a given video memory size, however some games\n"
-	        "               may not use them properly (flickering) or may need\n"
-	        "               more system memory to use them.");
+	        "  all:         All modes for a given video memory size, however some games may\n"
+	        "               not use them properly (flickering) or may need more system\n"
+	        "               memory to use them.");
 
 	pbool = secprop->Add_bool("vga_8dot_font", only_at_start, false);
 	pbool->Set_help("Use 8-pixel-wide fonts on VGA adapters (disabled by default).");
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -551,6 +551,18 @@ void DOSBOX_Init()
 	        "the selected video adapter ('auto' by default). See the 'machine' setting for\n"
 	        "the list of valid options per adapter.");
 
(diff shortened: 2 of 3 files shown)


MX9000:
@johnnovak Great! :smiley::thumbsup:
I am very happy to know that I have contributed a little bit to the advancement of DOSBox Staging and eXoDOS projects.
Thank you!

kklobe (approved):
Nice work! Tested this with Future Crew's Panic! and it greatly helps the initial scroller sequence:

```ini
[dosbox]
vmem_delay = 1000

[cpu]
cycles = 15000
```

Clean, simple, straightforward code. Approved!

Python-Exoproject:
I think this will fix Wing Commander 1, but I cant remember how the game ran back in the day to zero the settings in well enough for eXoDOS considering the public profile this game has. Ive been testing in the later missions using the launch command 'wc Origin -k s11 m4`

So far im working with cycles=10000 and vmem_delay=on

Anyone want to try and give feedback, especially if they remember how it played back in the day or have real hardware for compare?

johnnovak:
> Anyone want to try and give feedback, especially if they remember how it played back in the day or have real hardware for compare?

I have never played the game, that's why I wanted to leave it to you 😄 (or anyone who is "qualified" to determine how it should play).

johnnovak:
@Python-Exoproject you might be interested in this.

Found a very informative thread on Vogons that backs up my theory that `vmem_delay = on` would likely help "even out" the wild speed discrepancies in some PCjr games (when you just can't find a single cycles setting that makes all parts of the game run equally well).

Do note that these findings should also hold to CGA games as CGA also had very slow video memory compared to the system memory addressed by the CPU.

Tandy and EGA game might also benefit in addition to the early VGA games this feature was primarily introduced for.


The thread in question and the key information is below (@weirddan455 this might interest you too because there's some good info about the PCjr memory layout in there).

**PCjr emulation (CVS)**
https://www.vogons.org/viewtopic.php?f=32&t=11064

> Remember the first 128K of memory is slower than any other memory because of the VGA access. So using a constant like 500 might average things out enough to be usable, but it's not going to be perfect. Perfect is actually close to impossible without knowing the memory setup .. the machine doesn't use DMA to refresh memory (because it doesn't have DMA onboard), so memory refresh is handled by each memory expansion card. Four 128K expansion cards will have four different refresh cycles, compared to one expansion card that just has 512K on it. (Another fun PCjr oddity, but that really only matters if you are crazy enough to do a cycle accurate simulator.) [(source)](https://www.vogons.org/viewtopic.php?p=79031#p79031)

> Access to the first 128K of memory was a tug-of-war between the CPU and the VGA chip, and the VGA chip always won. Which is why the Jr feels slow compared to a 4.77Mhz PC when running in that first 128K. Does KQ feel too fast because the emulator isn't emulating the shared memory design? [(source)](https://www.vogons.org/viewtopic.php?p=78996#p78996)

PR #3605, opened 2024-04-14
Update Portuguese Translation

altiereslima:


johnnovak:
Thanks @altiereslima. Can you please squash these two commits into a single one? This is fine for `main`, but do note `en.lng` is a little bit behind on `main`.

Are you going to raise another PR for `release/0.81.x`? Up to you, but that's the immediate goal to update the translations in the upcoming 0.81.1 release as I mentioned.

PR #3626, opened 2024-04-20
Config setting defaulting/syncing improvements & fixes

johnnovak:
# Description

The main change here is fixing how the defaulting happens when a config param is set to an invalid value:

- Defaulting all `*_filter` settings to the real defaults when an invalid value is set, not just setting them to `off`.
- Syncing the config setting to the default when this happens (e.g., when you set `sb_filter = asdf`, the `sb_filter` property will be defaulted to `auto` instead of containing the `asdf` value).
- I've applied the same syncing when defaulting happens to all the other settings I could find.

Moreover:

- `windowresolution` can now be changed at runtime.
- Invalid `windowresolution` settings now reset the default window size, not set it to 640x480 pixels (which is passable on non-high DPI displays, but on my Retina MacBook this resulted in a tiny window).

# Manual testing

This seems like a lot of changes, but there are no interdependencies between them, and they're all relatively simple. 

I've tested every single setting with all possible values as I went, so I have a very high confidence that everything works correctly.

Because of this this, I really want to backport this.

Once this is backported, we're ready for 0.81.1 as far as I'm concerned 🥳 

# Checklist

I have:

### DIFF
--- a/src/hardware/input/mouse_config.cpp
+++ b/src/hardware/input/mouse_config.cpp
@@ -160,6 +160,17 @@ static void SetSensitivity(const std::string_view sensitivity_str)
 	const auto& user_default = mouse_predefined.sensitivity_user_default;
 	const auto default_value = coeff * user_default;
 
+	auto set_mouse_sensitivity_setting = [](const int value) {
+		set_section_property_value("mouse",
+		                           "mouse_sensitivity",
+		                           format_str("%d", value));
+	};
+	auto set_mouse_sensitivity_settings = [](const int val_x, const int val_y) {
+		set_section_property_value("mouse",
+		                           "mouse_sensitivity",
+		                           format_str("%d %d", val_x, val_y));
+	};
+
 	// Split input string into values
 	auto values_str = split(sensitivity_str, " \t,;");
 	if (values_str.size() > 2) {
@@ -167,6 +178,8 @@ static void SetSensitivity(const std::string_view sensitivity_str)
 		            user_default);
 		mouse_config.sensitivity_coeff_x = default_value;
 		mouse_config.sensitivity_coeff_y = default_value;
+
+		set_mouse_sensitivity_setting(user_default);
 		return;
 	}
 
@@ -174,6 +187,8 @@ static void SetSensitivity(const std::string_view sensitivity_str)
 	if (values_str.empty()) {
 		mouse_config.sensitivity_coeff_x = default_value;
 		mouse_config.sensitivity_coeff_y = default_value;
+
+		set_mouse_sensitivity_setting(user_default);
 		return;
 	}
 
@@ -183,6 +198,7 @@ static void SetSensitivity(const std::string_view sensitivity_str)
 	bool out_of_range   = false;
 	const int value_min = -mouse_predefined.sensitivity_user_max;
 	const int value_max = mouse_predefined.sensitivity_user_max;
+
 	for (auto& value_str : values_str) {
 		// Remove trailing '%' signs, if present
 		if (value_str.ends_with('%')) {
@@ -192,9 +208,12 @@ static void SetSensitivity(const std::string_view sensitivity_str)
 		const auto value = parse_int(value_str);
 		if (!value) {
 			LOG_WARNING("MOUSE: Invalid 'mouse_sensitivity' setting: '%s', using '%d'",
-			            value_str.c_str(), user_default);
+			            value_str.c_str(),
+			            user_default);
 			mouse_config.sensitivity_coeff_x = default_value;
 			mouse_config.sensitivity_coeff_y = default_value;
+
+			set_mouse_sensitivity_setting(user_default);
 			return;
 		}
 
@@ -213,13 +232,18 @@ static void SetSensitivity(const std::string_view sensitivity_str)
 
 	// Set the actual values
 	assert(values_int.size() == 1 || values_int.size() == 2);
-	const auto value_float_0 = static_cast<float>(values_int[0]);
+	const auto value_float_0         = static_cast<float>(values_int[0]);
 	mouse_config.sensitivity_coeff_x = coeff * value_float_0;
+
 	if (values_int.size() == 2) {
 		const auto value_float_1 = static_cast<float>(values_int[1]);
 		mouse_config.sensitivity_coeff_y = coeff * value_float_1;
+
+		set_mouse_sensitivity_settings(values_int[0], values_int[1]);
 	} else {
 		mouse_config.sensitivity_coeff_y = mouse_config.sensitivity_coeff_x;
+
+		set_mouse_sensitivity_setting(values_int[0]);
 	}
 
 	return;
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -590,26 +590,29 @@ static void configure_opl_filter_for_model(mixer_channel_t opl_channel,
 		config.lpf_cutoff_freq_hz = cutoff_freq_hz;
 	};
 
-	const auto filter_type = determine_filter_type(filter_choice, sb_type);
-
-	if (!filter_type) {
-		if (filter_choice != "off") {
-			LOG_WARNING("%s: Invalid 'opl_filter' setting: '%s', using 'off'",
+	const auto filter_type = [&]() {
+		if (const auto maybe_filter_type = determine_filter_type(filter_choice,
+		                                                         sb_type)) {
+			return *maybe_filter_type;
+		} else {
+			LOG_WARNING("%s: Invalid 'opl_filter' setting: '%s', using 'auto'",
 			            sb_log_prefix(),
 			            filter_choice.c_str());
+
+			set_section_property_value("sblaster", "opl_filter", "auto");
+
+			const auto filter_type = determine_filter_type("auto",
+			                                               sb_type);
+			assert(filter_type);
+			return *filter_type;
 		}
-
-		opl_channel->SetHighPassFilter(FilterState::Off);
-		opl_channel->SetLowPassFilter(FilterState::Off);
-
-		set_section_property_value("sblaster", "opl_filter", "off");
-		return;
-	}
+	}();
 
 	// The filter parameters have been tweaked by analysing real hardware
 	// recordings. The results are virtually indistinguishable from the real
 	// thing by ear only.
-	switch (*filter_type) {
+	switch (filter_type)
+	{
 	case FilterType::None:
 	case FilterType::SB16:
 	case FilterType::Modern: break;
@@ -622,7 +625,7 @@ static void configure_opl_filter_for_model(mixer_channel_t opl_channel,
 	}
 
 	constexpr auto OutputType = "OPL";
-	log_filter_config(ChannelName::Opl, OutputType, *filter_type);
+	log_filter_config(ChannelName::Opl, OutputType, filter_type);
 	set_filter(opl_channel, config);
 }
 
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -507,6 +507,8 @@ static void configure_sb_filter_for_model(mixer_channel_t channel,
 
 		channel->SetHighPassFilter(FilterState::Off);
 		channel->SetLowPassFilter(FilterState::Off);
+
+		set_section_property_value("sblaster", "sb_filter", "off");
 		return;
 	}
 
@@ -599,6 +601,8 @@ static void configure_opl_filter_for_model(mixer_channel_t opl_channel,
 
 		opl_channel->SetHighPassFilter(FilterState::Off);
 		opl_channel->SetLowPassFilter(FilterState::Off);
+
+		set_section_property_value("sblaster", "opl_filter", "off");
 		return;
 	}
 
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -601,10 +601,13 @@ static void configure_opl_filter_for_model(mixer_channel_t opl_channel,
 
 			set_section_property_value("sblaster", "opl_filter", "auto");
 
-			const auto filter_type = determine_filter_type("auto",
-			                                               sb_type);
-			assert(filter_type);
-			return *filter_type;
+			if (const auto filter_type = determine_filter_type("auto", sb_type);
+			    filter_type) {
+				return *filter_type;
+			} else {
+				assert(false);
+				return FilterType::None;
+			}
 		}
 	}();
 
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -498,21 +498,21 @@ static void configure_sb_filter_for_model(mixer_channel_t channel,
 		config.zoh_rate_hz = NativeDacRateHz;
 	};
 
-	const auto filter_type = determine_filter_type(filter_choice, sb_type);
+	const auto filter_type = [&]() {
+		if (const auto maybe_filter_type = determine_filter_type(filter_choice,
+		                                                         sb_type)) {
+			return *maybe_filter_type;
+		} else {
+			LOG_WARNING("%s: Invalid 'sb_filter' setting: '%s', using 'modern'",
+			            sb_log_prefix(),
+			            filter_choice.c_str());
 
-	if (!filter_type) {
-		LOG_WARNING("%s: Invalid 'sb_filter' setting: '%s', using 'off'",
-		            sb_log_prefix(),
-		            filter_choice.c_str());
+			set_section_property_value("sblaster", "sb_filter", "modern");
+			return FilterType::Modern;
+		}
+	}();
 
-		channel->SetHighPassFilter(FilterState::Off);
-		channel->SetLowPassFilter(FilterState::Off);
-
-		set_section_property_value("sblaster", "sb_filter", "off");
-		return;
-	}
-
-	switch (*filter_type) {
+	switch (filter_type) {
 	case FilterType::None: enable_zoh_upsampler(); break;
 
 	case FilterType::SB1:
@@ -547,7 +547,7 @@ static void configure_sb_filter_for_model(mixer_channel_t channel,
 	}
 
 	constexpr auto OutputType = "DAC";
-	log_filter_config(sb_log_prefix(), OutputType, *filter_type);
+	log_filter_config(sb_log_prefix(), OutputType, filter_type);
 	set_filter(channel, config);
 }
 
--- a/include/control.h
+++ b/include/control.h
@@ -125,7 +125,7 @@ public:
 		return sectionlist.end();
 	}
 
-	Section* GetSection(const std::string& section_name) const;
+	Section* GetSection(const std::string_view section_name) const;
 	Section* GetSectionFromProperty(const char* prop) const;
 
 	void OverwriteAutoexec(const std::string& conf, const std::string& line);
--- a/include/setup.h
+++ b/include/setup.h
@@ -406,7 +406,7 @@ public:
 	                                      const std::string& sep);
 
 	Property* Get_prop(int index);
-	Property* Get_prop(const std::string& propname);
+	Property* Get_prop(const std::string_view propname);
 
 	int Get_int(const std::string& _propname) const;
 
--- a/include/setup.h
+++ b/include/setup.h
@@ -45,6 +45,10 @@ std::optional<bool> parse_bool_setting(const std::string_view setting);
 bool has_true(const std::string_view setting);
 bool has_false(const std::string_view setting);
 
+void set_section_property_value(const std::string_view section_name,
+                                const std::string_view property_name,
+                                const std::string_view property_value);
+
 class Hex {
 private:
 	int value;
--- a/src/capture/image/image_capturer.cpp
+++ b/src/capture/image/image_capturer.cpp
@@ -26,6 +26,7 @@
 #include "std_filesystem.h"
 
 #include "checks.h"
+#include "setup.h"
 #include "string_utils.h"
 
 CHECK_NARROWING();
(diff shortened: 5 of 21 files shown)


src/gui/sdlmain.cpp
```diff
@@ -2957,6 +2963,8 @@ static void setup_initial_window_position_from_conf(const std::string& window_po
 		            window_position_val.c_str(),
 		            desktop.w,
 		            desktop.h);
+
+		set_auto_window_position_setting();
```
weirddan455, line 2975: This is again mixing concerns about what the config system should be.  The first use-case was good.  Parse the value with `sscanf` and if the value is garbage, fall back to a default.

Here though, you're checking against the user's runtime environment (the user's desktop size) which could change at any time if the user changes desktop resolution (some people probably run DOSBox on a CRT and wish to do this) or move DOSBox to a different monitor.

Say, for example, DOSBox opens on the wrong monitor.  Oops, resolution is too small so we fall back to default.  User moves it to the correct monitor and changes some other setting which causes the window to be re-created.  Except now it gets re-created at the small default size because we've thrown away their config setting.

IMO we should only change config settings if they are invalid, not if the users's screen is too small or their graphics driver fails.  If we wanted this value to be the "current window resolution", we would be updating this every time the window resizes which we aren't doing.

weirddan455: Sorry, this is window position, not resolution but the same point still applies.

src/gui/sdlmain.cpp
```diff
@@ -2899,27 +2877,57 @@ static SDL_Point window_bounds_from_label(const std::string& pref,
...
 	const int h = ceil_sdivide(desktop.h * percent, 100);
 
 	return {w, h};
 }
 
+static SDL_Point parse_window_resolution_from_conf(const std::string &pref)
+{
+	int w = 0;
+	int h = 0;
+	const bool was_parsed = sscanf(pref.c_str(), "%dx%d", &w, &h) == 2;
```
weirddan455, line 2914: Small gripe but it would be nice if this worked regardless of whether the x is capitalized.  I tested and a capital X does not work here.

src/gui/sdlmain.cpp
```diff
@@ -2899,27 +2877,57 @@ static SDL_Point window_bounds_from_label(const std::string& pref,
...
+
+	const bool is_valid = (w >= FallbackWindowSize.x &&
+	                       h >= FallbackWindowSize.y);
+
+	if (was_parsed && is_valid) {
+		maybe_limit_requested_resolution(w, h, "window");
+		return {w, h};
+	}
```
weirddan455, line 2922: I think we should clamp here if the user specifies a size that is too small rather than falling through to the percent-based parser.  `FallbackWindowSize` should be... well, the fallback in that case :smile:

src/gui/sdlmain.cpp
```diff
@@ -2899,27 +2877,57 @@ static SDL_Point window_bounds_from_label(const std::string& pref,
...
 		} else {
-			LOG_WARNING("DISPLAY: Invalid 'windowresolution' setting: '%s', "
-			            "using 'default'",
-			            pref.c_str());
+			LOG_WARNING(
+			        "DISPLAY: Invalid 'windowresolution' setting: '%s', "
+			        "using 'default'",
+			        pref.c_str());
+
+			set_section_property_value("sdl", "windowresolution", "default");
```
weirddan455, line 2897: Contrary to my previous point about window position, this one looks good to me.  We're only updating the setting if it is invalid.  We're not updating it in `maybe_limit_requested_resolution` which checks runtime environment.

src/hardware/input/mouse_config.cpp
```diff
@@ -160,13 +160,21 @@ static void SetSensitivity(const std::string_view sensitivity_str)
...
+
 	// Split input string into values
 	auto values_str = split(sensitivity_str, " \t,;");
 	if (values_str.size() > 2) {
 		LOG_WARNING("MOUSE: Too many values in 'mouse_sensitivity', using '%d'",
 		            user_default);
 		mouse_config.sensitivity_coeff_x = default_value;
 		mouse_config.sensitivity_coeff_y = default_value;
+
+		set_default_mouse_sensitivity_setting();
```
weirddan455, line 177: GitHub won't let me comment on the correct line but you're missing a reset to default on line 184 just under this.

weirddan455 (changes requested):
Outside of `sdlmain.cpp` this is fine for the most part.  I just always try to be careful in that file.  We've exceeded our maximum foot-gun capacity there :smile:

src/hardware/sblaster.cpp
```diff
@@ -603,10 +603,13 @@ static void configure_opl_filter_for_model(mixer_channel_t opl_channel,
 
 			set_section_property_value("sblaster", "opl_filter", "auto");
 
-			const auto filter_type = determine_filter_type("auto",
-			                                               sb_type);
-			assert(filter_type);
-			return *filter_type;
+			if (const auto filter_type = determine_filter_type("auto", sb_type);
+			    filter_type) {
+				return *filter_type;
```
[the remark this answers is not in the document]
johnnovak: > That's a really weird way to write an if statement. Why not leave the assignment outside of the if?

It's a common pattern for handling optional types in the codebase. Check out `program_mixer.cpp`, I counted 10 usages only there.

This form is a shortcut for this typical use case:

```cpp
const auto maybe_p = parse_float(s);
if (maybe_p) {
    // do something with the result
    const auto p = *maybe_p;
    return p;
} else {
    // handle failure case
}
```

It's just more compact to do it like this, especially if you do it a lot. The other benefit is you won't end up with all these `maybe_*` vars one scope above the `if`.

```cpp
if (const auto p = parse_float(s); p) {
    // do something with the result
    return *p;
} else {
    // handle failure case
}
```

> This is more readable IMO:

Yeah, if you're not used to this pattern, I agree. However, I started using this pattern extensively a while ago, and I think once people are used to it, it's quite nice and natural, so my preference is to leave it for consistency.

src/gui/sdlmain.cpp
```diff
@@ -3205,13 +3210,17 @@ static void set_output(Section* sec, const bool wants_aspect_ratio_correction)
 		if (!set_default_window_mode()) {
 			LOG_WARNING("OPENGL: Could not create OpenGL window, "
 			            "using 'texture' output mode");
+
 			sdl.want_rendering_backend = RenderingBackend::Texture;
+			set_texture_output_setting();
```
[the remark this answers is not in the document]
johnnovak: Good pickup, thanks. I will take `sdl.interpolation_mode` into consideration too here, and set `texture` or `texturenb` accordingly. The log message itself is misleading;, we should log `texture` or `texturenb` when falling back to the texture backend from `opengl` or `openglnb`, respectively.

Only the log message was guiding my change, so thanks for looking into it more deeply 🎉 

> Again, I'm not sure what your intent was here. In this case, the config was already set to a valid value but you're updating it based on a runtime error.

Yes, correct that is what's happening.

One can set a config setting to a valid value, e.g. you can set `cms = off`. That's valid. But if you enable the Game Blaster with `sbtype = gb`, the CMS _must_ be enabled as there is no Game Blaster hardware without CMS. So this case, we log an error and notify the user about the forced setting change.

image

image

Naturally, it would be rather confusing to leave the `cms` setting on `off`, don't you agree?

To use a GUI analogy, in more complex scenarios other GUI elements will be auto-reset for you. E.g., in this case selecting the "Game Blaster" from a dropdown would auto-tick the CMS option. This is pretty standard.

> Is the config value meant to be "this is what the user wants" or "this is what we're actually running"?

Yes, the latter—the currently active setting.

src/gui/sdlmain.cpp
johnnovak: Pasting the code in question here to make sure we're talking about the same thing:

```cpp
        const auto desktop = get_desktop_size();

	const bool is_out_of_bounds = x < 0 || x > desktop.w || y < 0 ||
	                              y > desktop.h;
	if (is_out_of_bounds) {
		LOG_WARNING("DISPLAY: Invalid 'window_position' setting: '%s'. "
		            "Requested position is outside the bounds of the %dx%d "
		            "desktop, using 'auto'.",
		            window_position_val.c_str(),
		            desktop.w,
		            desktop.h);

		set_auto_window_position_setting();
		return;
	}
```

I think you're right, and by changing the setting to `auto`, I've changed the behaviour, like you described.

To be fair, It's a pretty change edge case-y change in behaviour, but a change it is, so thanks for pointing it out 😄 Again, I was just going by the log message ("using `auto`"), without thinking about it too much. It's good to have you to keep me on my toes, hey 😎

Although it most cases I'd say syncing the config settings is desirable, I see your point that in some rare edge cases that's worse, this is being one of them.

Btw, it was me who originally added the `window_position` property, that was actually my first Staging PR ever 😄 My motivation was to use it in conjunction with `window_decorations = off` (my doing too) to set the DOSBox window to occupy exactly the left half of the screen when playing RPGs with my mapper program, [Gridmonger](https://gridmonger.johnnovak.net/). I'm actually doubtful people would rely on that edge case you described, but fair point, I'll remove the syncing in this case.

Check this example out:
https://gridmonger.johnnovak.net/_static/img/frontpage/ss-eob-full.png

src/gui/sdlmain.cpp
johnnovak: I agree, it's an easy and useful addition, so doing it.

src/gui/sdlmain.cpp
```diff
@@ -3143,6 +3143,10 @@ static void set_output(Section* sec, const bool wants_aspect_ratio_correction)
 	GFX_DisengageRendering();
 	// it's the job of everything after this to re-engage it.
 
+	auto set_texture_output_setting = []() {
+		set_section_property_value("sdl", "output", "texture");
+	};
```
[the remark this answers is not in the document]
johnnovak: I know, but the whole thing is a mess. I just wanna do surgical improvements at this point, so this will stay (but enhancing it so it handles `opengl`->`texture` and `openglnb`->`texturenb` fallback correctly).

src/hardware/input/mouse_config.cpp
johnnovak: Thanks, and there's another one here; will address that one too:

```cpp
	if (out_of_range) {
		LOG_WARNING("MOUSE: 'mouse_sensitivity' adjusted to range %+d - %+d",
		            value_min,
		            value_max);
	}
```

src/gui/sdlmain.cpp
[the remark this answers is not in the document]
johnnovak: Also, conflating the interpolation method into the rendering backend's name is terrible; we intended to decouple these years ago, but somehow, it never happened. I thought we had a ticket for it, but nope.

We should have something like this:

```ini
output = texture | opengl
texture_interpolation = bilinear | nearest
```

Forcing nearest for OpenGL is just silly, the shaders should have full control over that, plus I can't see a legit use-case for having "nearest" for OpenGL. I'd much rather introduce a `nearest` shader in addition to `none` (which does bilinear, so we could alias `none` to `bilinear` as well).

johnnovak:
Thanks for the detailed and thoughtful review @weirddan455, you've picked up a few questionable changes and omissions, that's great 🎉 

I've replied to your comments with my explanations and reasons, but to reiterate the config setting handling approach I started migrating towards a while ago:

- The config settings almost always should reflect the currently active values. Whether a setting was reset to its default because the user provided an invalid value, or the setting was forced to a different value than requested at runtime (because certain setting combinations just don't make sense), the syncing must always happen.
- There are a very small number of exceptions to this, as you pointed out: the `window_position` and `window_resolution` settings. For these, it makes sense _not_ to sync the config value to the currently active _valid_ setting. Do note that these two settings are very special edge cases, and the syncing makes perfect sense for all the other settings.
- As I see it, _not_ doing this syncing is very confusing for people who like to change settings at runtime via the CLI, and it would be equally super confusing for the GUI. What is laid out in front of you in a GUI should always reflect whats _currently_ active; in a GUI, controls that are not valid should be disabled (greyed out) for certain setting combos, or the contents of drop downs should be dynamically altered (items removed/added based on the setting of other controls), or in some cases some settings would "snap back" to settings that make sense in certain combinations of settings. The CLI behaviour should reflect this to provide a consistent experience across the CLI and GUI (OSD). In a way, I'm laying the groundwork here for the OSD, but doing this syncing has a lot of merit even when you don't take the GUI/OSD aspects into consideration IMO.

src/gui/sdlmain.cpp
johnnovak: Well, I don't really want to get into the deep of this by changing it too much... I disagree with the window size defaulting in general in sdlmain, so I'd re-think it from scratch.

The options here are for now:

1. Do nothing (that's bad for reasons I explained in the top description)
2. Default to `auto` which does the percentage thing (ok in my book)
3. Try to clip it (dunno how that would change the overall behaviour in surprising ways, to be inconsistent with all the other window-size determination paths; `auto` is at least something we already have)
4.

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
weirddan455: Why `if (const auto p = parse_float(s); p)` over just `if (const auto p = parse_float(s))` though?  What does the extra `; p` do?

src/gui/sdlmain.cpp
weirddan455: Well hold on you're changing behavior here.  This is the calling code:

```cpp
	sdl.use_exact_window_resolution = pref.find('x') != std::string::npos;
	if (sdl.use_exact_window_resolution) {
		coarse_size = parse_window_resolution_from_conf(pref);
	} else {
		coarse_size = window_bounds_from_label(pref);
	}
```

It's already doing the check whether it should be parsing for a fixed resolution or a percentage based value.  Then you're falling back to percent based anyway if something went wrong without updating `sdl.use_exact_window_resolution`

Before your change, `parse_window_resolution_from_conf` returned `FallbackWindowSize` if something went wrong.

weirddan455:
>  The config settings almost always should reflect the currently active values. Whether a setting was reset to its default because the user provided an invalid value, or the setting was forced to a different value than requested at runtime (because certain setting combinations just don't make sense), the syncing must always happen.

My biggest concern with this is that you're making behavior changes to `sdlmain.cpp` and want to put it in a bug fix release.  That code likes to fail in the strangest of places whenever a stiff breeze hits it.  Previously, the config values were read-only by `sdlmain`.

I also don't like having two "sources of truth" for the current state of the program.  We already have global SDL variables that store the current renderer, resolution, etc.  You make a valid point about it being better to present the user the current value from command line so I can get over that I suppose.  There's a lot more I dislike about the current config system anyway.

src/gui/sdlmain.cpp
johnnovak: Yes, and I think my changed behaviour is an improvement, but feel free to disagree 😄 

I mentioned this in the top description:

> - Invalid `windowresolution` settings now reset the default window size, not set it to 640x480 pixels (which is passable on non-high DPI displays, but on my Retina MacBook this resulted in a tiny window).

So on a high-DPI screen, if I start up Staging, I get the default window size. So far so good.

Then I go `windowresolution = x` (you need to enable changing `windowresolution` at runtime for that first in the code). Well, that sets it to `FallbackWindowSize`, which is 640x480. But that's in pixels. So now I'm not getting the "default size" I got at startup, but this tiny postage stamp as you can see below 😄 

To me, that's quite broken—defaulting should be a bit more advanced, and at least take high-DPI into account even if we use a hardcoded fallback window size. But that's a more complicated, so I thought, ok, let's fix it by doing what "auto" does. With my fix, `windowresolution = x` will get you the same window size you get at startup without specifying `windowresolution` explicitly, not the postage stamp...

I can revert this whole thing if you really don't like it as I'm not going into this whole window size defaulting rabbit hole. Just wanted to do a small incremental improvement, that's all 😄 

image

image

johnnovak:
> My biggest concern with this is that you're making behavior changes to `sdlmain.cpp` and want to put it in a bug fix release. That code likes to fail in the strangest of places whenever a stiff breeze hits it. Previously, the config values were read-only by `sdlmain`.

Oh right, I thought these were relatively safe, but that was before you pointed out these various issues. Yeah, I agree it's better if I don't cherry-pick these sdlmain related things over. I think I still want the sdlmain improvements to go into main, then we can perhaps further improve them later on.

I will change the history so the sdlmain related stuff will be on the top, then I don't cherry-pick those. Do you find that acceptable? (but I could split them into a separate PR if you want that for tracking or whatever)

> I also don't like having two "sources of truth" for the current state of the program. We already have global SDL variables that store the current renderer, resolution, etc. You make a valid point about it being better to present the user the current value from command line so I can get over that I suppose. There's a lot more I dislike about the current config system anyway.

Yeah, no disagreements there, I think we could remove some redundancies there, but that's a bigger job like you're saying. I'm just trying to improve the user-observable behaviour here when changing settings at runtime.

I also have a draft for these large-scale refactoring jobs, including the config handling, I'm planning to finish that and post it as a discussion ticket once 0.81.1 is out, so in a few weeks I think.

src/gui/sdlmain.cpp
weirddan455: I think it is an improvement but you've stuck it in the wrong place.  Have a look at the calling code in `setup_window_sizes_from_conf`.  `sdl.use_exact_window_resolution` really shouldn't be a global variable.  It's only used in that one function.

You're not updating that variable though and there's more logic that follows that modifies window size:

```cpp
// Save the coarse bounds in the SDL struct for future sizing events
sdl.desktop.requested_window_bounds = {coarse_size.x, coarse_size.y};

// Refine the coarse resolution and save it in the SDL struct.
auto refined_size = coarse_size;
if (sdl.use_exact_window_resolution) {
	refined_size = clamp_to_minimum_window_dimensions(coarse_size);
} else {
	refined_size = refine_window_size(coarse_size,
	                                  wants_aspect_ratio_correction);
}
```

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
johnnovak: Indeed, I don't even remember why I started doing it like that 😄 But yeah, `if (const auto p = parse_float(s))` seems sufficient. Maybe it was to make one of the static analysers happy or something... 

I'll simplify it and see how it goes.

Another reason why I prefer the definition inside the `if` is that there is a small room for error when you define the value on a separate line (you could use a different variable name in the if by mistake and could not notice).

weirddan455:
>  I will change the history so the sdlmain related stuff will be on the top, then I don't cherry-pick those into release/0.81.x. Do you find that acceptable? (but I could split them into a separate PR if you want that for tracking purposes or whatever)

Yeah, everything I saw in the PR seems very low risk for backport outside of the sdlmain stuff.  And maybe I'm being overly paranoid about sdlmain but changes there have been problematic in the past.  Up to you how you want to structure the PRs.  I'm not familiar with how you handle the backporting process.

johnnovak:
> > I will change the history so the sdlmain related stuff will be on the top, then I don't cherry-pick those into release/0.81.x. Do you find that acceptable? (but I could split them into a separate PR if you want that for tracking purposes or whatever)
> 
> Yeah, everything I saw in this PR seems very low risk for backport outside of the sdlmain stuff. And maybe I'm being overly paranoid about sdlmain but changes there have been problematic in the past.

Nah, I think you're right, and those small enhancements are not worth the risk of breaking something else. Case in point: I thought they were safe and simple changes, then you pointed out a bunch of problems 😄 Not backport material, we're in agreement.

> Up to you how you want to structure the PRs. I'm not familiar with how you handle the backporting process.

My method is to cherry-pick commits into the release branch right after merging the PR into main. That's low overhead and it ensures I don't forget about backporting stuff.

PRs labeled with "backport" are also tagged in the release notes draft now which helps me double-check things come release time:
https://github.com/dosbox-staging/dosbox-staging/releases

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
johnnovak: Ok, CI is not complaining, so simplifying all `if` statements.

src/gui/sdlmain.cpp
johnnovak: > It's already doing the check whether it should be parsing for a fixed resolution or a percentage based value. Then you're falling back to percent based anyway if something went wrong without updating sdl.use_exact_window_resolution

Ok got it, that's not very indeed. I've simplified the logic so the decision whether to use the percentage based or exact dimensions is decided at that top level.

> I think it is an improvement but you've stuck it in the wrong place. Have a look at the calling code in setup_window_sizes_from_conf. sdl.use_exact_window_resolution really shouldn't be a global variable. It's only used in that one function.

Right, turned that into a local var. sdlmain is full with cruft like that 😱

johnnovak:
Ok @weirddan455, I've addressed your comment and I've decided to split out the sdlmain related stuff into a second PR, after all. Addressing your sdlmain related comments properly resulted in more extensive rework, so I think this is for the better.

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
weirddan455: This line still looks the same.  Did you forget to change it here?

weirddan455 (approved):
Looks good now.  I think you forgot to change that if statement I commented on but that's a small stylistic thing.  Merge when you're ready.

johnnovak:
> Looks good now. I think you forgot to change that if statement I commented on but that's a small stylistic thing. Merge when you're ready.

Cheers, I thought I simplified all ifs that handle options, but having a second look.

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
johnnovak: Yeah I changed all of them but missed this one :)

PR #3629, opened 2024-04-22
Fix various crashes when the host sample rate is set to low values

johnnovak:
# Description

Ok, last one to backport @weirddan455 😄 

---

###  Fix crash when filter cutoff frequencies are not below half of the sample rate

The IIR library rightfully throws an exception when you attempt to set a filter's cutoff frequency at or above the Nyquist rate (half the sample rate). So with this change we're always clamping the filter cutoff frequencies just below the Nyquist rate.

The reason why we're clamping and not turning the filter off is because this is an edge case that usually only happens when using 20,050 or lower host rates, and it's a lot easier to just clamp than to disable the filter and handle the state management correctly.

To trigger the crash, set rate `rate = 11025` in the `[mixer]` section, then run `sbtype sbpro` which will attempt to configure an 8 kHz lowpass filter. You can also trigger it by setting a custom filter, e.g., `sb_filter lpf 1 10000`.

### Fix some audio devices crashing when host sample rate is set to 8000 Hz

FluidSynth and MT-32 would crash with the `rate = 8000` setting, and that's a valid rate we otherwise support in all audio devices.

# Manual testing

I started Prince of Persia after each of these test cases to make sure that the audio output actually works.

## Case 1

1. Set `rate = 11025` in the config.
2. Confirm in the logs that Staging runs at 11,025 Hz.
3. Run `sbtype sbpro1`.
4. No crash and the following debug message appears in the logs:
    image

## Case 2

1. Start with `--noprimaryconf`
2. Confirm in the logs that Staging runs at 44.1 or 48 kHz.
3. Run `sb_filter lpf 1 40000`.
5. No crash and something like this appears in the logs:
    image

## Case 3

1. Set `rate = 8000` in the config.
2. Start with `--noprimaryconf`.
3. FluidSynth is active now, no crash.
    image
6. Run `mididevice mt32`.
7. MT-32 is active now, no crash.
    image

# Checklist

I have:

### DIFF
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -1147,7 +1147,7 @@ void MixerChannel::SetHighPassFilter(const FilterState state)
 
 	if (do_highpass_filter) {
 		log_filter_settings(name,
-		                    "Highpass",
+		                    "High-pass",
 		                    state,
 		                    filters.highpass.order,
 		                    filters.highpass.cutoff_freq_hz);
@@ -1160,7 +1160,7 @@ void MixerChannel::SetLowPassFilter(const FilterState state)
 
 	if (do_lowpass_filter) {
 		log_filter_settings(name,
-		                    "Lowpass",
+		                    "Low-pass",
 		                    state,
 		                    filters.lowpass.order,
 		                    filters.lowpass.cutoff_freq_hz);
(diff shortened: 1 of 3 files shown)


src/hardware/mixer.cpp
```diff
@@ -2735,7 +2735,12 @@ static bool init_sdl_sound(Section_prop* section)
 		LOG_WARNING("MIXER: SDL changed the requested sample rate of %d to %d Hz",
 		            mixer.sample_rate_hz.load(),
 		            obtained.freq);
+
 		mixer.sample_rate_hz = check_cast<uint16_t>(obtained.freq);
+		set_section_property_value("mixer",
```
weirddan455, line 2740: I would `assert(false)` on this block.  It should never be allowed to happen based on the flags we pass.  We're telling SDL the only thing we're willing to negotiate is the number of samples, not frequency.

```cpp
int sdl_allow_flags = 0;

if (negotiate) {
	// We allow chunk size changes (if supported), but don't allow
	// frequency changes because we've only seen them go beyond
	// SDL's desired 48 Khz limit
#if defined(SDL_AUDIO_ALLOW_SAMPLES_CHANGE)
	sdl_allow_flags |= SDL_AUDIO_ALLOW_SAMPLES_CHANGE;
#endif
}
```
I would actually re-consider these flags though (but not for backport).  If we tell SDL not to negotiate, it will re-sample our audio behind the screns from what we requested to whatever the underlying hardware and OS will accommodate.  SDL's re-sampler is kind of crap.  My one contribution to SDL was better integration of a third-party resampling library but that's off by default.  I believe Staging uses Speex for resampling which is much better.

weirddan455: This is also contrary to what's documented in the config file help.  It says it will negotiate sample rate but it doesn't.

```cpp
bool_prop = sec_prop.Add_bool("negotiate", only_at_start, default_allow_negotiate);
bool_prop->Set_help(
        "Let the system audio driver negotiate possibly better sample rate and blocksize\n"
        "settings (%s by default).");
```

weirddan455 (approved):
Your changes look good.  I noticed some existing issues with how we're setting up SDL but it's probably out of scope for this PR and probably not something we want to backport.

src/hardware/mixer.cpp
weirddan455: On second thought, you don't have to add the assert.  It's a path we probably want to support in the future.  I believe it should never with the way the SDL flags are now though.

johnnovak: > We're telling SDL the only thing we're willing to negotiate is the number of samples, not frequency.

@weirddan455 Huh, I kinda glossed over that detail. Good to know.

> I would actually re-consider these flags though (but not for backport). If we tell SDL not to negotiate, it will re-sample our audio behind the screns from what we requested to whatever the underlying hardware and OS will accommodate.

That's a bit crap. I'd expect to to simply fail with negotiation off if the driver doesn't support that requested rate, and with negotiation on, it should just find the closest driver-supported rate. 

> SDL's re-sampler is kind of crap. My one contribution to SDL was better integration of a third-party resampling library but that's off by default. My motivation for doing so was the built-in one was seg-faulting in some cases 🙄 I believe Staging uses Speex for resampling which is much better.

I'd really like disallow _any_ SDL-side resampling whatsoever. Is that possible with some flags? I want it to work as per above, and I want to keep all resampling business in our hands.

Yeah, I introduced Speex, it's very good quality, fast enough even on the Raspbery Pi 4, and has low latency, which is very important in emulators.

I found the Speex quality setting of 5 to be _almost_ transparent. I could only hear tiny differences with very intense A/B comparisons on very special material in studio headphones, and even then I wasn't 100% sure it wasn't just placebo or expectation bias...

johnnovak:
Thanks for the quick review @weirddan455 🥇 Plus I learned some nasty details about SDL audio 🥶

PR #3631, opened 2024-04-23
Usage of integer types & disabling sign conversion warnings

johnnovak:
# Preamble

**kcgen** and **dreamer** have done an outstanding job raising our quality standards by adopting static analysers, enabling extra warnings, and addressing many of those warnings in our codebase. These improvements have undoubtedly made the code safer and more pleasant to work with.

But in some cases, they overshot a bit. We should periodically re-examine these past decisions based on actual experience and be honest with ourselves: does _[insert safety measure]_ really help us make the code safer, or does it only cause us a lot of busywork and complication with marginal or zero returns? We want to spend our limited time and energy we can dedicate to the project on things that really make a difference. If the net effect of a safety/quality measure is negative, we should by no means keep doing it!

# The problem

One of my and @kklobe's biggest pain points when it comes to warnings is the "liberal" use of fixed-width types (`uint8_t`, `uint16_t`, etc.) everywhere in the codebase. Performing arithmetic operations on unsigned types is very error-prone, plus there are the [integer promotion rules](https://www.geeksforgeeks.org/integer-promotions-in-c/) of C/C++, so we end up with lots of casting everywhere.

I strongly believe adopting this approach was a mistake and should be undone. Using fixed-width unsigned types instead of plain old signed `int` everywhere forced us to do a lot of busy work for exactly zero benefits.

In fact, the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) suggests using `int` everywhere and only deviating from that if really warranted.

Quoting relevant bits from the [integer types](https://google.github.io/styleguide/cppguide.html#Integer_Types) section (I recommend reading the whole thing):

> We use `int` very often, for integers we know are not going to be too big, e.g., loop counters. **Use plain old `int` for such things.** You should assume that an `int` is at least 32 bits, but don't assume that it has more than 32 bits. If you need a 64-bit integer type, use `int64_t` or `uint64_t`.

> For integers we know can be "big", use `int64_t`.

> **You should not use the unsigned integer types such as `uint32_t`**, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, **do not use unsigned types to say a number will never be negative**. Instead, use assertions for this.

> Unsigned integers are good for representing bitfields and modular arithmetic. [..] The fact that unsigned arithmetic doesn't model the behavior of a simple integer, but is instead defined by the standard to model modular arithmetic (wrapping around on overflow/underflow), means that a significant class of bugs cannot be diagnosed by the compiler. In other cases, the defined behavior impedes optimization.

All very sane advice and would simplify our lives a lot!

To explain it further, we should only use fixed-width unsigned (or sometimes signed) types in these scenarios:

- Modeling registers that store bit-field patterns in emulated hardware (e.g., the VGA registers).
- Dealing with the contents of DOS-side memory directly.
- Handling the contents of the video memory.
- Representing signed or unsigned 8 or 16-bit PCM streams.
- Dealing with packed network packets or binary file formats.

That's pretty much it. For every other use case—just use plain `ints`!

As a concrete example, this signature

```cpp
void configure_filter(const uint8_t order, const uint16_t cutoff_freq_hz)
```

should be rewritten as

```cpp
void configure_filter(const int order, const int cutoff_freq_hz) {
	assert(order > 0 && order <= 16);
	assert(cutoff_freq_hz > 0);
}
```

Do note we can be more specific in our asserts to check _value ranges_. An `uint8_t` still doesn't say anything about the valid upper limit.

# But wait, there's more!

Yeah, it's C++, man, so there is always more, and life is never simple... 😅 

Yet again, one of the C++ committee's decisions bites us in the ass:

> Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake, but it is effectively impossible to fix at this point.

I'm working on some more extensive mixer refactorings, so I converted `mixer.cpp` to only use `int`s.But many of those `int` loop counters and variables are used to index containers. Net result: 68 new sign conversion warnings. That's because of `Wsign-conversion` introduced in the `CHECK_NARROWING()` macro.

Once `Wsign-conversion` is removed, the number of new warnings goes down to _only two_.

I'm strongly in favour of removing `-Wsign-conversion` altogether. 99% of the time, we're dealing with numbers that are nowhere near the upper limit of the signed 32-bit integer range. Then, if you're dealing with code that uses "big numbers," you should be careful and _test_ your code instead of forcing the whole codebase to silence these harmless sign conversion warnings by littering every second line with casts...

We can still use `CHECK_NARROWING()` as the other two warnings, "-Wconversion" and "-Wnarrowing", seem to be far less noisy. However, we should revisit their usage and weigh up the pros and cons if they become a nuisance.

Once we have reached a consensus on this, I will adjust the warning counts in the workflows.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -48,7 +48,7 @@ jobs:
             run_tests: false # the shell_cmds_tests hang on arm64
             build_flags: -Dbuildtype=debug
             brew_path: /opt/homebrew
-            max_warnings: 60
+            max_warnings: 42
 
     steps:
       - name: Checkout repository
--- a/include/checks.h
+++ b/include/checks.h
@@ -1,7 +1,7 @@
 /*
  *  SPDX-License-Identifier: GPL-2.0-or-later
  *
- *  Copyright (C) 2022-2022  The DOSBox Staging Team
+ *  Copyright (C) 2022-2024  The DOSBox Staging Team
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
@@ -57,7 +57,6 @@
 
 #ifdef __GNUC__
 #	define CHECK_NARROWING() \
-		_Pragma("GCC diagnostic warning \"-Wsign-conversion\"") \
 		_Pragma("GCC diagnostic warning \"-Wconversion\"") \
 		_Pragma("GCC diagnostic warning \"-Wnarrowing\"") \
 		END_MACRO


weirddan455:
Aren't your two points kind of contradictory to each other?  You're saying:

1. We shouldn't be using unsigned types so much because it can result in unsafe conversions.
2. We should remove the warnings that are telling us about the problems you intend to solve by point 1.

It seems to me that if you're intending to do large refactors converting unsigned to signed, we should leave the warnings in because they will point us to other variables that should probably be converted to signed.

johnnovak:
> Aren't your two points kind of contradictory to each other? You're saying:
> 
>     1. We shouldn't be using unsigned types so much because it can result in unsafe conversions.
> 
>     2. We should remove the warnings that are telling us about the problems you intend to solve by point 1.
> 
> 
> It seems to me that if you're intending to do large refactors converting unsigned to signed, we should leave the warnings in because they will point us to other variables that should probably be refactored to signed types.

Um, no, read it again :)

I'm saying use just `int` because it's a lot more convenient and those warnings are pretty much useless. That's my primary point.

Yeah, and arithmethics with unsigned types are _also_ unsafe, unless you follow the compilers warnings dutifully and have millions of casting back and fort between signed and unsigned. That will make your code "safe", but at what cost?

So, the TL;DR:

1. Using `int` is a lot easier, that's the main concern, and we _also_ sidestep the pitfalls of the unsigned types.
2. Because `int`s are used for indexing in a lot of places, and the singed to unsigned warnings are pretty much worthless except in very special scenarios, I want them to go as well.

The two are very much related, and if you do 1, you also need 2 for practical reasons.

shermp:
Your talking about ints, but what about much smaller types like byte or short, where sign conversion would be far more likely to bite you in the ass, and it would be really nice to be warned about it if you do it on accident.

weirddan455:
Just to be clear, I agree with your point about preferring signed types (or just plain `int`).  That is the easiest thing to do in most cases.  No arguments with anything you said there.

>  I'm working on some more extensive mixer refactorings, so I converted mixer.cpp to only use ints.But many of those int loop counters and variables are used to index containers. Net result: 68 new sign conversion warnings.

If it's only using `int` why are there a bunch of sign conversions?  There's obviously some unsigned types still lingering about.

>  I'm strongly in favour of removing -Wsign-conversion altogether. 99% of the time, we're dealing with numbers that are nowhere near the upper limit of the signed 32-bit integer range.

This is true but the big unsafely issue is when you're converting in the other direction and accidentally cast a negative number to a very large value in an unsigned type.

>  So then I can use an int to index an array of 1024 elements or something without the compiler bitching about implicitly casting int to size_t (which is unsigned int).

Is the compiler really throwing a warning about using an int as an index?  That's silly if so but I think it's more likely you're comparing or doing arithmetic with an unsigned type.

johnnovak:
> Your talking about ints, but what about much smaller types like byte or short, where sign conversion would be far more likely to bite you in the ass, and it would be really nice to be warned about it if you do it on accident.

I think you're mixing things up. `-Wconversion` and `-Wnarrowing` would be still in place. Wouldn't those take care of these scenarios?

I think you're missing some subtle detail here. Without `-Wsign-conversion` we wouldn't need to deal with all that crap where you can't use an `int` to index a container because `int` is not `size_t` (aka `unsigned int`). That's just annoying and super noisy.

If you try to chunk a larger type into `uint8_` / `int8_t` or `uint16_t` / `int16_t`, the narrowing warnings will warn you. I did not talk about removing those.

shermp:
> Is the compiler really throwing a warning about using an int as an index? That's silly if so but I think it's more likely you're comparing or doing arithmetic with an unsigned type.

It's probably a pattern like this that triggers the warning:
```cpp
for (int i = 0; i < vec.size(); ++i) {
   // do something
}
```

shermp:
> > Your talking about ints, but what about much smaller types like byte or short, where sign conversion would be far more likely to bite you in the ass, and it would be really nice to be warned about it if you do it on accident.
> 
> I think you're mixing things up. `-Wconversion` and `-Wnarrowing` would be still in place. Wouldn't those take care of these scenarios?

I'm talking about when you do something like:

```cpp
char a = typedefed_signed_byte;
```
Could bite you in the ass.

johnnovak:
> > Is the compiler really throwing a warning about using an int as an index? That's silly if so but I think it's more likely you're comparing or doing arithmetic with an unsigned type.
> 
> It's probably a pattern like this that triggers the warning:
> 
> ```c++
> for (int i = 0; i < vec.size(); ++i) {
>    // do something
> }
> ```

Yeah, that's it. See it for yourself @weirddan455:

## 1st commit 

Conversion to ints in the mixer done, `-Wsign-conversion` still active:
https://github.com/dosbox-staging/dosbox-staging/actions/runs/8793044838/job/24130295770

Literally tons of crap like this about indexing containers with ints:

image

Summary:

## 2nd commit

`-Wsign-conversion` nuked:
https://github.com/dosbox-staging/dosbox-staging/actions/runs/8793324202/job/24131011561

Result:

image



image

johnnovak:
> > > Your talking about ints, but what about much smaller types like byte or short, where sign conversion would be far more likely to bite you in the ass, and it would be really nice to be warned about it if you do it on accident.
> > 
> > 
> > I think you're mixing things up. `-Wconversion` and `-Wnarrowing` would be still in place. Wouldn't those take care of these scenarios?
> 
> I'm talking about when you do something like:
> 
> ```c++
> char a = typedefed_signed_byte;
> ```
> 
> Could bite you in the ass.

Yeah could be. I'm not saying a warning in this case would not be useful, but if it's either / or, I'd much rather live with this danger than littering the codebase with casts.

How often do we cast `uint8_t` to `int8_t`? Not very often.

weirddan455:
Ok @johnnovak this is some stupid shit for the compiler to be complaining about.  You changed the index to `auto` from `size_t` but this code is fine.  I can agree that this warning is too noisy and can probably be turned off then.

https://github.com/dosbox-staging/dosbox-staging/blob/56a220f5e214c01e8ab0a4d9d44694d86a0fb25a/src/hardware/mixer.cpp#L1295-L1299

johnnovak:
> Ok @johnnovak this is some stupid shit for the compiler to be complaining about. You changed the index to `auto` from `size_t` but this code is fine. I can agree that this warning is too noisy and can probably be turned off then.
> 
> https://github.com/dosbox-staging/dosbox-staging/blob/56a220f5e214c01e8ab0a4d9d44694d86a0fb25a/src/hardware/mixer.cpp#L1295-L1299

Yeah, man, I was surprised by it too. Then I did some research and yeah... "known issue"... C++... 🥶 

To reiterate, I'm not advocating "turn off all the warnings, lol". I'd like to keep an useful set of narrowing/sign-conversion/etc. warnings, so if you guys have better suggestions how to get only the useful warnings, them I'm all ears for it.

I just wanna get rid of the compiler spamming us and forcing us to do a lot of busywork 😄

shermp:
I think the real problem is containers returning `size_t` for their `size()` method.

shermp:
Actually, since we can do c++ 20 now, does `vec.ssize()` work on all compilers we support? That returns a signed type.

shermp:
> That doesn't really help if it's warning on the `[]` operator.

Is it warning on the index operator? Or just the comparison (`i < vec.size()`)?

weirddan455:
>  Is it warning on the index operator? Or just the comparison (i < vec.size())?

The bracket operator.  Scroll up.  I just quoted a block of code that it was warning on.

johnnovak:
> You've convinced me. Maybe wait and see what the rest of the team thinks.

Definitely, that's why I tagged every core member as reviewer.

johnnovak:
> > That doesn't really help if it's warning on the `[]` operator.
> 
> Is it warning on the index operator? Or just the comparison (`i < vec.size()`)?

Yeah, go through what we wrote again carefully please. But here's a toy example:

```cpp
// clang test.cpp -Wsign-conversion                                                  

#include <array>

int main() {
	std::array<char, 5> buf = {};
	for (auto i = 0; i < 5; ++i) {
		buf[i] = 0;
	}
}
```

image

Same error if you change it to `int i`, and `int` for buffer positions is a very common thing.

shermp:
Oh ffs, standard containers are just broken aren't they. Ok, you've convinced me for now.

shermp (approved):

johnnovak:
I've just done this experiment:

```cpp
// Compile with:
// clang test.cpp -Weverything

#include <array>

int main() {
    int8_t   s8 = 0;
    uint8_t  u8 = 0;

    int16_t  s16 = 0;
    uint16_t u16 = 0;

	s8 = u8;   // sign-conversion
	u8 = s8;   // sign-conversion

	// We could  maybe enable this, although not sure how noisy it is in general
	s8 = s16;  // implicit-int-conversion
	s8 = u16;  // implicit-int-conversion
	u8 = s16;  // implicit-int-conversion
	u8 = u16;  // implicit-int-conversion

	s16 = u16; // sign-conversion
	u16 = s16; // sign-conversion

	s16 = s8;  // no warning
	s16 = u8;  // no warning

	u16 = s8;  // Wsign-conversion
	u16 = u8;
}
```

johnnovak:
An excellent commentary about the whole unsigned indexing fiasco. Pretty much all important people think is was a big mistake.

https://eigen.tuxfamily.narkive.com/ZhXOa82S/signed-or-unsigned-indexing#post4

kklobe:
> Yeah, go through what we wrote again carefully, please. Also, read the Google Style Guide—it explains this in detail and has recommendations (e.g., use iterators instead of indexing to work around the issue).

This is mentioned in passing, but this should be the preferred way of dealing with collections when possible, since it solves even more problems, e.g. someone changes the size of the array and all the loops are then busted. Index-based looping over collections should be the exception.

johnnovak:
> > Yeah, go through what we wrote again carefully, please. Also, read the Google Style Guide—it explains this in detail and has recommendations (e.g., use iterators instead of indexing to work around the issue).
> 
> This is mentioned in passing, but this should be the preferred way of dealing with collections when possible, since it solves even more problems, e.g. someone changes the size of the array and all the loops are then busted. Index-based looping over collections should be the exception.

Yeah, it's good to emphasize this, and I'll try to migrate to iterators in my mixer rework.

kklobe:
I'm in violent agreement :grin:.

My only suggestion is to perhaps include some concrete guidance in `CONTRIBUTING.md` with some of the examples and points made above.

Issue #3620, opened 2024-04-18
CDROM_Interface pointer ownership issue


japsmits:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

Git

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The pointer created by new CDROM_Interface_Image gets stored in two places (CDROM_Interface_Image::images and CMmscdex::cdrom) and can get deleted in both places. This is undefined behaviour.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Code inspection...
2.
3.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

japsmits:
Working on a not-too-invasive solution of this. Basically, one owning store for all CDROM_Interface decedents.

weirddan455:
I'd suggest either adding some (temporary) log statements or setting a debugger breakpoint to confirm a double free happens in practice.  Actual behavior can often be surprising.

Either way, a cleanup of this would be welcome though.

PR #3634, opened 2024-04-25
Fix CDROM_Interface pointer ownership issue

japsmits:
# Description

There is now only one owner of CDROM_Interface pointers.

## Related issues

#3620 

# Manual testing

Tried mounting a CD-ROM or image in several ways, and they still work.

# Checklist

I have:

### DIFF
--- a/src/dos/cdrom.cpp
+++ b/src/dos/cdrom.cpp
@@ -28,6 +28,10 @@
 #include "pic.h"
 #include "string_utils.h"
 
+namespace CDROM {
+std::array<std::unique_ptr<CDROM_Interface>, MaxNumDosDriveLetters> cdroms;
+}
+
 // ******************************************************
 // Fake CDROM
 // ******************************************************
--- a/src/dos/cdrom.h
+++ b/src/dos/cdrom.h
@@ -110,10 +110,12 @@ public:
 	virtual bool PauseAudio         (bool resume) = 0;
 	virtual bool StopAudio          () = 0;
 	virtual void ChannelControl     (TCtrl ctrl) = 0;
+	virtual bool ReadSector         (uint8_t* buffer, const bool raw, const uint32_t sector) = 0;
 	virtual bool ReadSectors        (PhysPt buffer, const bool raw, const uint32_t sector, const uint16_t num) = 0;
 	virtual bool ReadSectorsHost    (void* buffer, bool raw, unsigned long sector, unsigned long num) = 0;
 	virtual bool LoadUnloadMedia    (bool unload) = 0;
 	virtual void InitNewMedia       () {}
+	virtual bool HasDataTrack       () const = 0;
 	virtual bool HasFullMscdexSupport() = 0;
 
 protected:
@@ -156,6 +158,11 @@ public:
 
 	void ChannelControl([[maybe_unused]] TCtrl ctrl) override {}
 
+	bool ReadSector(uint8_t* /*buffer*/, const bool /*raw*/,
+	                const uint32_t /*sector*/) override
+	{
+		return true;
+	}
 	bool ReadSectors(PhysPt /*buffer*/, const bool /*raw*/,
 	                 const uint32_t /*sector*/, const uint16_t /*num*/) override
 	{
@@ -168,6 +175,10 @@ public:
 		return true;
 	}
 	bool LoadUnloadMedia(bool /*unload*/) override
+	{
+		return true;
+	}
+	bool HasDataTrack() const override
 	{
 		return true;
 	}
@@ -303,8 +314,8 @@ public:
 	bool ReadSectorsHost(void* buffer, bool raw, unsigned long sector,
 	                     unsigned long num) override;
 	bool LoadUnloadMedia(bool unload) override;
-	bool ReadSector(uint8_t* buffer, const bool raw, const uint32_t sector);
-	bool HasDataTrack();
+	bool ReadSector(uint8_t* buffer, const bool raw, const uint32_t sector) override;
+	bool HasDataTrack() const override;
 	bool HasFullMscdexSupport() override
 	{
 		return true;
@@ -408,11 +419,13 @@ public:
 	                 unsigned char& index, TMSF& relPos, TMSF& absPos) override;
 	bool GetMediaTrayStatus(bool& mediaPresent, bool& mediaChanged,
 	                        bool& trayOpen) override;
+	bool ReadSector(uint8_t* buffer, const bool raw, const uint32_t sector) override;
 	bool ReadSectors(PhysPt buffer, const bool raw, const uint32_t sector,
 	                 const uint16_t num) override;
 	bool ReadSectorsHost(void* buffer, bool raw, unsigned long sector,
 	                     unsigned long num) override;
 	bool LoadUnloadMedia(bool unload) override;
+	bool HasDataTrack() const override;
 	bool HasFullMscdexSupport() override
 	{
 		return true;
@@ -442,11 +455,13 @@ public:
 	                 unsigned char& index, TMSF& relPos, TMSF& absPos) override;
 	bool GetMediaTrayStatus(bool& mediaPresent, bool& mediaChanged,
 	                        bool& trayOpen) override;
+	bool ReadSector(uint8_t* buffer, const bool raw, const uint32_t sector) override;
 	bool ReadSectors(PhysPt buffer, const bool raw, const uint32_t sector,
 	                 const uint16_t num) override;
 	bool ReadSectorsHost(void* buffer, bool raw, unsigned long sector,
 	                     unsigned long num) override;
 	bool LoadUnloadMedia(bool unload) override;
+	bool HasDataTrack() const override;
 	bool HasFullMscdexSupport() override
 	{
 		return true;
--- a/src/dos/cdrom.h
+++ b/src/dos/cdrom.h
@@ -63,6 +63,9 @@
 #define AUDIO_DECODE_BUFFER_SIZE 16512u
 #define SAMPLES_PER_REDBOOK_FRAME (BYTES_PER_RAW_REDBOOK_FRAME / REDBOOK_BPS)
 
+// Number of letters in the English alphabet (A to Z)
+constexpr auto MaxNumDosDriveLetters = 26;
+
 struct TMSF
 {
 	unsigned char min;
@@ -121,6 +124,10 @@ public:
 protected:
 	void LagDriveResponse() const;
 };
+
+namespace CDROM {
+extern std::array<std::unique_ptr<CDROM_Interface>, MaxNumDosDriveLetters> cdroms;
+}
 
 class CDROM_Interface_Fake final : public CDROM_Interface {
 public:
@@ -292,7 +299,7 @@ public:
 		bool                       mode2      = false;
 	};
 
-	CDROM_Interface_Image(uint8_t sub_unit);
+	CDROM_Interface_Image();
 
 	~CDROM_Interface_Image() override;
 	void InitNewMedia() override {}
@@ -320,7 +327,6 @@ public:
 	{
 		return true;
 	}
-	static CDROM_Interface_Image* images[26];
 
 private:
 	static struct imagePlayer {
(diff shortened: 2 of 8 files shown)


src/dos/cdrom_win32.cpp
```diff
@@ -315,6 +321,37 @@ bool CDROM_Interface_Win32::LoadUnloadMedia(bool unload)
...
+{
+	CDROM_TOC toc = {};
+
+	DWORD control_code       = IOCTL_CDROM_READ_TOC;
+	LPVOID input_buffer      = NULL;
+	DWORD input_buffer_size  = 0;
+	LPVOID output_buffer     = &toc;
+	DWORD output_buffer_size = sizeof(toc);
+	LPDWORD bytes_returned   = NULL;
+	LPOVERLAPPED overlapped  = NULL;
```
johnnovak, line 334: Very nice 🎖️

src/dos/cdrom_win32.cpp
```diff
@@ -315,6 +321,37 @@ bool CDROM_Interface_Win32::LoadUnloadMedia(bool unload)
...
+	                     input_buffer,
+	                     input_buffer_size,
+	                     output_buffer,
+	                     output_buffer_size,
+	                     bytes_returned,
+	                     overlapped)) {
+		return false;
+	}
+
+	for (int i = 0; i < toc.LastTrack; i++) {
```
johnnovak, line 347: minor: we do pre-increment/decrement by default.

johnnovak: `auto i`?

src/dos/cdrom.h
```diff
@@ -121,6 +121,10 @@ class CDROM_Interface
 	void LagDriveResponse() const;
 };
 
+namespace CDROM {
+extern std::array<std::unique_ptr<CDROM_Interface>, 26> cdroms;
```
johnnovak, line 125: What is 26?

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -229,6 +229,11 @@ bool CDROM_Interface_Ioctl::GetMediaTrayStatus(bool& mediaPresent,
 	return true;
 }
 
+bool ReadSector(uint8_t* buffer, const bool raw, const uint32_t sector) override
```
johnnovak: As per our discussion in my recent PR, feel free to start migrating things toward simple `int`s (unless it would be a bigger refactoring job and you're not ready for that right now, which is fine 😄 ).

src/dos/cdrom.cpp
```diff
@@ -28,6 +28,10 @@
 #include "pic.h"
 #include "string_utils.h"
 
+namespace CDROM {
```
johnnovak, line 31: CdRom would be better.

src/dos/cdrom.cpp
```diff
@@ -28,6 +28,10 @@
...
+std::array<std::unique_ptr<CDROM_Interface>, 26> cdroms;
+}
```
johnnovak, line 33: Do we even need this namespace? If `cdroms` is module-internal, we don't need it, then if external modules want access the CDROM_Interface, probably a `CDROM_GetDrive()` or something would be better?

src/dos/cdrom_ioctl_linux.cpp
japsmits: I have a cleanup PR in the works (that's how I found out about the pointer ownership), so I'll add this to that PR.

src/dos/cdrom.cpp
johnnovak: Yeah, it makes sense then and it's ok for now.

src/dos/cdrom_ioctl_linux.cpp
```diff
@@ -357,4 +363,9 @@ std::vector<int16_t> CDROM_Interface_Ioctl::ReadAudio(const uint32_t sector, con
 	return audio_samples;
 }
 
+bool CDROM_Interface_Ioctl::HasDataTrack() const
+{
+	return true; /*TODO*/
```
weirddan455, line 368: This is pretty simple to implement (very similar to what you did on Win32).  I can do it once this merges if you don't have a Linux machine.

src/dos/dos_mscdex.cpp
```diff
@@ -290,44 +279,38 @@ int CMscdex::RemoveDrive(uint16_t _drive)
...
 	if (!path_exists(path)) {
-		return nullptr;
+		return {};
 	}
 
 	if (!is_directory(path)) {
-		cdrom_interface = new CDROM_Interface_Image(num_drives);
-		if (cdrom_interface->SetDevice(path)) {
+		if (auto cdrom_interface = std::make_unique<CDROM_Interface_Image>();
+		                                   cdrom_interface->SetDevice(path)) {
```
weirddan455, line 290: Really not a fan of these multi-statement (with a semi-colon in the middle) if's.  I'm actually not even sure how they work.  Do both statements have to be true (pointer is not null and `SetDevice` returns true)?  Does it only check the last one?

I've seen @johnnovak do this same pattern so if he's OK with it then I'll live.

src/dos/dos_mscdex.cpp
```diff
@@ -400,23 +383,20 @@ int CMscdex::AddDrive(uint16_t _drive, const char* physicalPath, uint8_t& subUni
...
 		}
-		cdrom[0] = _cdrom;
-		CDROM_Interface_Image::images[0] = _cdimg;
-		dinfo[0].drive		= (uint8_t)_drive;
-		dinfo[0].physDrive	= (uint8_t)toupper(physicalPath[0]);
-		subUnit = 0;
+		CDROM::cdroms[0]   = std::move(_cdrom);
+		dinfo[0].drive     = (uint8_t)_drive;
+		dinfo[0].physDrive = (uint8_t)toupper(physicalPath[0]);
+		subUnit            = 0;
```
weirddan455, line 395: Can you double-check the logic here any maybe clean this up a bit?  There's lot of `std::move` both in and out of the array and I'm not confident that this is doing the same thing that the old code was.

It would be better if you could keep the smart pointer as a local variable and then do `std::move` once to move it into the array.

weirddan455:
This is a welcome cleanup @japsmits 

Btw feel free to add me as a reviewer on your PRs, especially if it's code I've recently worked on like this is.  Not a huge deal but I probably would have seen it sooner.

src/dos/dos_mscdex.cpp
johnnovak: I'm ok with them, and they're really simple:

```cpp
if (<initialisation expression>; <evaluation expression>)
```

E.g. this will execute the expression in the `if` block:

```cpp
if (const int a = 5; a > 1) 
```

weirddan455: Makes sense.  You can leave it in if you want then.

johnnovak: Also see here @weirddan455. IMO it's a neat C++17 feature:

> C++17 has extended existing if statement’s syntax. Now it is possible to provide initial condition within if statement itself. This new syntax is called "if statement with initializer". This enhancement simplifies common code patterns and helps users keep scopes tight. Which in turn avoids variable leaking outside the scope.


https://stackoverflow.com/a/56866523
https://www.tutorialspoint.com/cplusplus17-if-statement-with-initializer

weirddan455: Oh it's a relatively new feature.  That's why I'm just not used to seeing it.  It used to be you couldn't even declare a variable inside an if statement or it would be a compile error.

src/dos/dos_mscdex.cpp
japsmits: The loop was already there. The only difference between the old and the new is that we only have to rotate things in two places (CDROM::cdroms and dinfo) instead of three (dinfo, cdrom and CDROM_Interface_Image::images).

All this code performs a push_front in an array. I'll add a comment there and see if I can simplify this. CDROM::cdroms should actually be an std::vector...

src/dos/cdrom_ioctl_linux.cpp
japsmits: No Linux-machine at the moment. I'll see if I can whip up a VM somewhere in the near future, but I would like to invite you to implement this one for now. 😄

japsmits:
Ready for another review. 😄

src/dos/dos_mscdex.cpp
```diff
@@ -400,23 +383,23 @@ int CMscdex::AddDrive(uint16_t _drive, const char* physicalPath, uint8_t& subUni
...
-	if (dinfo[0].drive-1==_drive) {
-		CDROM_Interface *_cdrom = cdrom[numDrives];
-		CDROM_Interface_Image *_cdimg = CDROM_Interface_Image::images[numDrives];
-		for (uint16_t i=GetNumDrives(); i>0; i--) {
-			dinfo[i] = dinfo[i-1];
-			cdrom[i] = cdrom[i-1];
-			CDROM_Interface_Image::images[i] = CDROM_Interface_Image::images[i-1];
+	if (dinfo[0].drive - 1 == _drive) {
+		// Perform a push_front on dinfo and CDROM::cdroms
+		for (auto i = GetNumDrives(); i > 0; i--) {
```
johnnovak, line 388: minor: --i for consistency

src/dos/cdrom.h
```diff
@@ -62,6 +62,8 @@
 #define AUDIO_DECODE_BUFFER_SIZE 16512u
 #define SAMPLES_PER_REDBOOK_FRAME (BYTES_PER_RAW_REDBOOK_FRAME / REDBOOK_BPS)
 
+constexpr auto dos_drives = 26;
```
johnnovak, line 65: Constants are UpperCamelCase, and suggest the following change:

```cpp
// Number of letters in the English alphabet (A to Z)
constexpr auto MaxNumDosDriveLetters = 26;
```

johnnovak: Thanks for changing, maybe add this too above it?

```
// Number of letters in the English alphabet (A to Z)
```

johnnovak:
Okay, I gave it some testing I ran into some issues @japsmits.

## Under a Killing Moon

You can get it from here:
https://the-eye.eu/public/Games/eXo/eXoDOS_v6r2/eXo/eXoDOS/

I'm using this simple local config:

```ini
[autoexec]
imgmount d CD/disk*.cue -t iso
```

The disks get mounted:

image

And I can switch between them with Cmd/Ctrl+F4:

image

But quitting Staging results in a crash:

image

## Transport Tycoon Deluxe

Same issue, I'm just mounting the ISO then exiting Staging, then I get a crash. This game doesn't even use CD Audio tracks; it's just a simple ISO with data only.

## Ultima Underworld

Same, just mounting a simple ISO then exiting gets me a segfault.

---

So it looks like this needs more work and more thorough testing @japsmits.

japsmits:
I'm currently unable to spend time on this because of personal matters consuming most of my free time. I'll continue the work when everything is sorted out (might take some time).

weirddan455:
>  I'm currently unable to spend time on this because of personal matters consuming most of my free time. I'll continue the work when everything is sorted out (might take some time).

No worries.  Take care of yourself.  The project will still be around when/if you have time.

johnnovak:
No problem @japsmits. We're here when you're ready.

japsmits:
And back again. 😄 

I think I found the culprit and added some code to unmount cdroms before shutting down MSCDEX.

johnnovak:
> And back again. 😄
> 
> I think I found the culprit and added some code to unmount cdroms before shutting down MSCDEX.

Welcome back @japsmits 🥳 I will test this later today!

src/dos/cdrom_ioctl_linux.cpp
johnnovak: Let's merge this first, then @weirddan455 can tinker with it (if he wants).

johnnovak (approved):
This works very well now @japsmits, no crash on exit now! Just one very minor comment of mine to address, then it's merge time 🥳

src/dos/cdrom.h
japsmits: Done. 👍

PR #3657, opened 2024-05-06
Allow loadfix to execute batch files

### LINKED ISSUE #3655: Loadfix command does not work on batch files

_No response_

Linux x86_64

_No response_

### Relevant hardware info

_No response_

The loadfix command allows you to specify a program as an argument.  This allocates some memory before starting the program and then frees it when it finishes.  However, this does not work with batch files and it would be nice if it did.  Example:  `loadfix test.bat` outputs `64kb allocated. 64kb freed.` and the batch file never gets executed.

This can be worked around by typing `loadfix` on its own and then manually freeing with `loadfix /d` afterwards.

I have tested on v0.80.1 and this is not a regression.  I assume loadfix just never had this working.

Explain how to reproduce
1. `loadfix anybat.bat`
2.  Memory gets allocated and immediately freed without the batch file being executed.

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

`shell.ExecuteProgram()` has code to detect if the filename is a batch file and, if so, appends it to a `batchfiles` data structure but it doesn't actually execute the batch file.  Make a call to `shell.RunBatchFile()` to run the batch file.  This call does nothing if the filename was an executable.

## Related issues

Fixes #3655

# Manual testing

Tested with a test batch file and with Sim City 2000 batch file and direct executable.

# Checklist

I have:

### DIFF
--- a/src/dos/program_loadfix.cpp
+++ b/src/dos/program_loadfix.cpp
@@ -76,7 +76,10 @@ void LOADFIX::Run(void)
 			}
 			// Use shell to start program
 			DOS_Shell shell;
+			// If it's a batch file, this call places it into an internal data structure.
 			shell.ExecuteProgram(filename, args);
+			// Actually run the batch file. This is a no-op if it's an executable.
+			shell.RunBatchFile();
 			DOS_FreeMemory(segment);
 			WriteOut(MSG_Get("PROGRAM_LOADFIX_DEALLOC"),kb);
 		}


src/dos/program_loadfix.cpp
```diff
@@ -77,6 +77,7 @@ void LOADFIX::Run(void)
 			// Use shell to start program
 			DOS_Shell shell;
 			shell.ExecuteProgram(filename, args);
+			shell.RunBatchFile();
```
johnnovak, line 80: In reality, either ExecuteProgram or RunBatchFile runs, right? Can you put them in an if/else then to clarify how this works in actually reality?

The bare minimum would be an explanatory comment; right now, I'm looking at it and it's weird—execute a program first, then run a batch file? Huh? 😄

weirddan455: They both need to run.  Like I said in the description up top, ExecuteProgram puts the batch file in a data structure and RunBatchFile pulls it off and runs it (it's a no-op if there's no batch file).

johnnovak: Oh, right. Yes you did, indeed. It's so weird that I did not even comprehend it 😄 Yeah, please copy paste that top description into the code then, everybody will be super confused by this...

Next Windows build should be fine.

weirddan455: I added a comment to this a couple days ago.  Was there anything else you wanted changed here?

johnnovak: Thanks, that's all. Merge away 😎

johnnovak (approved):
Nice simple fix @weirddan455 🎉

PR #3663, opened 2024-05-10
Fix broken shell_history_file config option (global object version)

### LINKED ISSUE #3661: Command line history saving can't be disabled

0.82.0-alpha-847d0

Windows 10

_No response_

### Relevant hardware info

_No response_

The command line history saving can't be disabled by setting the `shell_history_file` value to empty like the config file suggests. The file `shell_history.txt` is always created and the history is saved.

You can disable it in version 0.81.0

_No response_

_No response_

### Your configuration

```ini
shell_history_file =
```

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

The issue is that the `global_shell_history` object was being constructed before the config file was parsed.  Previously, it was reading the config and the shell history file during object creation, which was happening at the incorrect time (too early in the init process).

In this PR, I made it more explicit when the shell history gets read and written.  Constructors and destructors have been removed and the object is now a real global instead of the weird reference to `unique_ptr` that was going on before.

## Related issues

Fixes #3661

# Manual testing

- Setting `shell_history_file` to empty disables reading and writing shell history entirely
- Removing the `shell_history_file` line entirely falls back to the default of `shell_history.txt`
- Setting `shell_history_file` to an arbitrary file name works as expected
- If `shell_history_file` is not empty and does not exist, a new file is created

# Checklist

I have:

### DIFF
--- a/include/shell.h
+++ b/include/shell.h
@@ -90,20 +90,16 @@ class ShellHistory {
 public:
 	std::vector<std::string> GetCommands(uint16_t code_page) const;
 	void Append(std::string command, uint16_t code_page);
-
-	ShellHistory();
-	~ShellHistory();
-	ShellHistory(const ShellHistory&)            = delete;
-	ShellHistory& operator=(const ShellHistory&) = delete;
-	ShellHistory(ShellHistory&&)                 = delete;
-	ShellHistory& operator=(ShellHistory&&)      = delete;
+	void SetPathFromConfig();
+	void ReadHistoryFromFile();
+	void WriteHistoryToFile();
 
 private:
-	std::vector<std::string> commands{};
-	std_fs::path path;
+	std::vector<std::string> commands = {};
+	std_fs::path path = {};
 };
 
-extern std::unique_ptr<ShellHistory> global_shell_history;
+extern ShellHistory global_shell_history;
 
 class DOS_Shell : public Program {
 private:
@@ -116,7 +112,6 @@ private:
 
 	friend class AutoexecEditor;
 
-	ShellHistory& history                  = *global_shell_history;
 	std::stack<BatchFile> batchfiles       = {};
 	uint16_t input_handle                  = STDIN;
 	bool call                              = false;
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -42,7 +42,7 @@ callback_number_t call_shellstop = 0;
 /* Larger scope so shell_del autoexec can use it to
  * remove things from the environment */
 DOS_Shell *first_shell = nullptr;
-std::unique_ptr<ShellHistory> global_shell_history;
+ShellHistory global_shell_history = {};
 
 static Bitu shellstop_handler()
 {
@@ -1366,12 +1366,15 @@ void SHELL_Init() {
 	// Load SETVER fake version table from external file
 	SETVER::LoadTableFromFile();
 
+	global_shell_history.SetPathFromConfig();
+	global_shell_history.ReadHistoryFromFile();
+
 	// first_shell is only setup here, so may as well invoke
 	// it's constructor directly
 	first_shell = new DOS_Shell;
 	first_shell->Run();
 	delete first_shell;
 	first_shell = nullptr; // Make clear that it shouldn't be used anymore
-	
-	global_shell_history.reset();
+
+	global_shell_history.WriteHistoryToFile();
 }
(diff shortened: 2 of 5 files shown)


weirddan455:
@MeAreJeenius feel free to review if you like since you wrote the `ShellHistory` class.

include/shell.h
```diff
@@ -90,20 +90,16 @@ class ShellHistory {
...
 	std::vector<std::string> GetCommands(uint16_t code_page) const;
 	void Append(std::string command, uint16_t code_page);
-
-	ShellHistory();
-	~ShellHistory();
-	ShellHistory(const ShellHistory&)            = delete;
-	ShellHistory& operator=(const ShellHistory&) = delete;
-	ShellHistory(ShellHistory&&)                 = delete;
-	ShellHistory& operator=(ShellHistory&&)      = delete;
+	void SetPathFromConfig();
```
johnnovak, line 93: Good riddance. Could mark the class `final` as well (`class ShellHistory final`).

src/shell/shell.cpp
```diff
@@ -42,7 +42,7 @@ callback_number_t call_shellstop = 0;
 /* Larger scope so shell_del autoexec can use it to
  * remove things from the environment */
 DOS_Shell *first_shell = nullptr;
-std::unique_ptr<ShellHistory> global_shell_history;
+ShellHistory global_shell_history = {};
```
johnnovak, line 45: Okay, so no globals like this please going forward. We want to remove this stuff, not add more 😄 

We should not rely on the init sequence of globals and static vars. Having global vars are fine as long as they're initialised manually by the code somewhere, so the init order is clear.

weirddan455: Originally the history was a member variable of the `Shell` class but @MeAreJeenius moved it to this global in his rework.  I believe the reasoning was so that we could save the history of child shells as well.  My original implementation only saved the history from `first_shell`.

In this case, init order shouldn't be a problem.  I've removed the constructors and destructors so no worries there.  It's also always in a valid state.  The data members are just a `std::vector<std::string>` and a `std_fs::path`.  When both of those are default constructed, it just means an empty history and no history file (same as if the user set `shell_history_file` to empty in their config file).

I think this is better than the smart pointer.  Before, it was possible (although probably very unlikely in practice) for a dangling reference to be held by some `Shell` object or to have a nullptr de-reference if a `Shell` happened to be created before `global_shell_history`.

IMO if we need a global like this, make a real global but make it as Plain Old Data as possible so we don't run into the issues you mentioned.

johnnovak: @weirddan455 I get your point, it's just Plain Old Data approach is error prone as you need to be "careful". I mean, for every single global you need to check whether it's safe or not, while if you nuke all globals, then the issue is non-existant.

Similarly to always using copies, always defaulting to immutable structures and mutate only when necessary etc., then things are just thread-safe by default, etc.

But ok, if it's too difficult in this case to remove the global, it's fine.

weirddan455: I get that globals should be used sparingly although nuking all globals from the codebase would be challenging to say the least.

In this case, I just really can't think of a better approach.  The current code uses a `std::unique_ptr` but then bypasses the safety of that by de-referencing the smart pointer and storing an un-safe reference as a member variable.  The global is certainly safer than that.

A `std::shared_ptr` would be a much better thing to use but then it's tricky of when to construct the singleton object.  You could do something like this:

```cpp
static std::shared_ptr<ShellHistory> maybe_create_history()
{
    static std::shared_ptr<ShellHistory> global_history = {};
    if (!global_history)
    {
        global_history = std::make_shared();
    }
    return global_history;
}

class Shell
{
    std::shared_ptr<ShellHistory> global_history = {};

    Shell::Shell()
    {
        global_history = maybe_create_history();
    }
}
```

But then that's just a global with extra steps really.

If you have a better idea, feel free to share but I honestly can't think of one.

johnnovak: > In this case, I just really can't think of a better approach. The current code uses a `std::unique_ptr` but then bypasses the safety of that by de-referencing the smart pointer and storing an un-safe reference as a member variable. The global is certainly safer than that.

Yeah, okay, let's not overdo it 😄 We should not choose solutions based on ideology, agreed, rather than to solve the problem at hand with minimal fuss.

I'll look into weak pointers at some points as I have no idea what that is 😄

weirddan455: Weak pointer is a non-owning reference to a shared pointer.  I just put in a 2nd PR using weak pointer and keeping RAII and you can let me know which one you like better.  See #3664

weirddan455:
Merged #3664 instead to stick with RAII

PR #3664, opened 2024-05-10
Fix broken shell_history_file config option (std::weak_ptr version)

### LINKED ISSUE #3661: Command line history saving can't be disabled

weirddan455:
# Description

This PR solves the same problem as #3663 but is more minimal with changes.  It retains RAII (reading the file on construction and writing on destruction) but uses `std::shared_ptr` and `std::weak_ptr` for proper smart point usage.  `global_shell_history` now gets created during construction of the first `DOS_Shell` object and destroyed after the destruction of the last `DOS_Shell object`.

`std::weak_ptr` is a non-owning reference to a `std::shared_ptr`.  weak_ptr's `.lock()` function atomically checks if the reference is still valid and, if so, returns a `std::shared_ptr` object.  We use this as our global object so that it does not hold a reference count and keep `global_shell_history` alive until some undetermined point when global objects get destroyed.  Instead, it will get destroyed when the last `std::shared_ptr` goes out of scope (which is when the last `DOS_Shell` object gets destroyed).

## Related issues

Fixes #3661

Previous PR using a global object #3663

# Manual testing

Functionally the same as the previous PR.  The same tests all work.

- Setting `shell_history_file` to empty disables reading and writing shell history entirely
- Removing the `shell_history_file` line entirely falls back to the default of `shell_history.txt`
- Setting `shell_history_file` to an arbitrary file name works as expected
- If `shell_history_file` is not empty and does not exist, a new file is created

# Checklist

I have:

### DIFF
--- a/include/shell.h
+++ b/include/shell.h
@@ -103,8 +103,6 @@ private:
 	std_fs::path path;
 };
 
-extern std::unique_ptr<ShellHistory> global_shell_history;
-
 class DOS_Shell : public Program {
 private:
 	void PrintHelpForCommands(MoreOutputStrings& output, HELP_Filter req_filter);
@@ -116,7 +114,7 @@ private:
 
 	friend class AutoexecEditor;
 
-	ShellHistory& history                  = *global_shell_history;
+	std::shared_ptr<ShellHistory> history  = {};
 	std::stack<BatchFile> batchfiles       = {};
 	uint16_t input_handle                  = STDIN;
 	bool call                              = false;
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -1109,7 +1109,6 @@ void DOSBOX_Init()
 	pstring = secprop->Add_path("shell_history_file",
 	                            only_at_start,
 	                            "shell_history.txt");
-	global_shell_history = std::make_unique<ShellHistory>();
 
 	pstring->Set_help(
 	        "File containing persistent command line history ('shell_history.txt'\n"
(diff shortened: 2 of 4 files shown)


weirddan455:
Downside of this approach... we're creating 2 temporary shells during init from calls like `PROGRAMS_MakeFile("COMMAND.COM", SHELL_ProgramCreate);`

Everything works but it's not quite as intuitive when things get called.   Here's some debug logs I took.  If you create subshell (run `command` from the first DOS shell), it uses the existing history as expected.

weakptr

johnnovak (approved):
I think I like this version better @weirddan455. Approving! 👍🏻

weirddan455:
OK @johnnovak it ended up looking cleaner than I thought after I wrote it.  The only downside is we're reading and writing from the history for no real reason with those temporary shells but it's a couple hundred KB of disk I/O at the most so probably not a big deal.

I just pushed a small improvement to move the weak pointer `global_shell_history` to be a static local variable inside the `DOS_Shell` constructor since that's the only place it is used now.

johnnovak:
> these get constructed and then destructed before `first_shell`) but it's a couple hundred KB of disk I/O at the most so probably not a big deal.

Yeah, honestly I don't care. Let's optimise for code clarity when the performance doesn't matter.

Issue #3627, opened 2024-04-21
Decouple texture interpolation method from the `output` backend names


johnnovak:
Conflating the interpolation method into the rendering backend's name is rather unfortunate, sloppy, and ultimately, incorrect. We've been intending to decouple these years ago, but somehow, it never happened. 

Now we have these:

- `opengl` – OpenGL + bilinear interpolation
- `openglnb` – OpenGL + nearest-neighbour interpolation
- `texture` – SDL's abstract "texture" backend + bilinear interpolation
- `texturenb` – SDL's abstract "texture" backend + nearest-neighbour interpolation

What I think we should do instead:

1. `output` should only specify the backend: `opengl` or `texture`.

2. The OpenGL backend is shader-based; you always have a shader active, even if you don't specify one with `glshader`. That's just how modern OpenGL works. The default shader is the `none` shader which uses bilinear interpolation.

    If we really wanted to provide a nearest-neighbour OpenGL mode, we should introduce a new `nearest` shader, and possibly alias `none` to `bilinear` as well. Then you could do this:

    ```ini
    [sdl] 
    output = opengl

    [render]
    # for bilinear:
    glshader = none 
    # or
    glshader = bilinear
    
    # for nearest-neighbour:
    glshader = nearest
    ```

3. The texture backend we use as a fallback for systems that are not OpenGL-capable (I would think there are not many out there, but that's a different topic), it makes sense to add a `texture_interpolation` setting with the `bilinear` and `nearest` options. Note we already have a `texture_renderer` setting, so the name of this new setting would be consistent with that.

This is all a rather low-risk change and could go in before the move to SDL3.

johnnovak:
`openglnb` has been removed [here](https://github.com/dosbox-staging/dosbox-staging/pull/3672) for good. I will add a pragma so shaders can request "no bilinear" interpolation to support the XBR upscalers (see [here](https://github.com/dosbox-staging/dosbox-staging/issues/3770), plus I'll add a `nearest.glsl` shader too.

Which leaves us with `texture` and `texturenb`, which is fine, on second thought. Introducing a new `texture_interpolation` setting with the `bilinear` and `nearest` options would be overdoing it IMO for the texture renderer fallback option, so I consider this ticket to be completed.

PR #3672, opened 2024-05-13
Kill `openglnb` with fire 🔥🔥🔥🔥🔥

johnnovak:
# Description

Welcome to the second episode of our popular ["kill X with fire" series](https://github.com/dosbox-staging/dosbox-staging/issues/2448)! 😎 In this installment, the ghastly "non-bilinear" OpenGL mode (`output = openglnb`) goes the way of the Dodo bird! 🐦🥳 🎊 *

### TL;DR for the impatient

- The `openglnb` output option is a relic from the past, born out of necessity in the early 2000s, and carried forward by not keeping up with the times and adopting more advanced image processing techniques in DOSBox land. There is simply zero reason for its continued existence, given we have the far superior `sharp` shader and `integer_scaling` options.

- The `glshader = sharp` + `output = opengl`  combo is superior to `openglnb` as it strikes a balance between _pixel-evenness_ and _sharpness_, which is super important with aspect-ratio correction enabled.

- The `openglnb` (nearest-neighbour, no bilinear) option, however, which is the most primitive form of upscaling an image, prioritises _sharpness_ and totally ignores pixel-evenness, which results in ghastly looking interference artifacts (moire patterns). This is easy to notice on checkerboard dither patterns and on text. It's less easy to notice on 256-colour VGA graphics with lots of smooth gradients, but, of course, the issue is always there.

Thanks to  @Python-Exoproject for drawing my attention to the issue (I almost never use `openglnb` these days, except for testing stuff).

_* No endangered species have been harmed during this exercise._ 😎 

## Related issues

This makes https://github.com/dosbox-staging/dosbox-staging/issues/3627 redundant (we'll just leave `texture` and `texturenb` as is).

## Summary

- `openglnb` performs nearest-neighbour image interpolation; this has no place in quality image processing in 2024 (and it was a very bad and obsolete technique even 20-25 years ago).
- Most shaders yield identical output with both `openglnb` (nearest-neighbour interpolation) and `opengl` (bilinear interpolation). E.g., the new CRT shaders don't care which output mode you choose; they will yield 100% identical output with both (because they do their own interpolation in the shader code).
- The `sharp` shader, however, which is our "sharp crisp pixels" option, takes advantage of the hardware accelerated bilinear interpolation of modern GPUs. Thereforce, if you force-disable bilinear texture interpolation by setting `output = openglnb`, you effectively negate the advantages of the `sharp` shader and get plain old (and bad) nearest-neighbour interpolation. This is bad. Very bad.
- Another issue is that the `crt-auto` shaders revert to the `sharp` shader when there isn't enough resolution to do proper CRT shading. Therefore, the CRT shaders appear to be "fine" with `openglnb`--until you start using small window sizes and/or high-resolution DOS video modes, in which case you'll effectively get nearest-neighbour.
- Arguably, the `openglnb` output should have been removed the moment Staging introduced the `sharp` shader years ago.
- `openglnb` has no performance or compatibility advantages over `opengl` whatsoever.
- The `texturenb` output mode stays as that is our only option for the fallback SDL texture renderer to achieve "sharp" pixels (at the cost of pixel evenness, as we'll see below).

## Historical context

- Although pixel shaders were introduced into common GPUs in 2001, and games like Morrowind used pixel shaders to very good effect, the adoption in emulation circles was slow due to whatever reasons. OpenGL drivers in the early to mid 2000s were not that great either, so maybe that played a role in that too.
- In any case, when DOSBox added its `opengl` output mode, that upscaled the framebuffer via bilinear interpolation, resulting in a blurry image most people absolutely hate and for good reasons. So, the guys thought, okay, let's add an option then to turn off bilinear interpolation, and `openglnb` (nb = no bilinear) was born. In this mode, the hardware-accelerated bilinear interpolation of the GPU is turned off, so you get "sharp" pixels with no fuzzy interpolation artifacts.
- With 2x, 3x, 4x, etc. integer scaling `openglnb` gives perfect results _if_the emulated video mode has square pixels (1:1 pixel aspect ratio). For example, 640x480 is such a mode. With `openglnb` and 2x scaling, you get 1280x960 with perfectly sharp pixels. So far so good.
- But 320x200, 640x350, 720x400, and many other video modes feature non-square pixels. In all these screen modes, the nearest-neighbour interpolation performed by the `openglnb` mode will result in quite ghastly looking, uneven scaling artifacts. Still, this was seen as the lesser evil as the other alternative DOSBox provided was bilinear interpolation via `opengl` which blurred the image.
- Probably a sad side-effect of this situation was that people kinda preferred to turn off aspect ratio correction, because with that off, you were always dealing with square pixels. So `openglnb` with 2x or 3x scaling and `aspect = off` still yielded perfectly sharp pixels for 320x200 content -- but in the _wrong aspect ratio_, so the graphics appeared ~20% squashed vertically.
- Fast forward to recent times... Because DOSBox development had essentially stopped around ~2010, even though other DOSBox forks started adding shader support, somehow using `openglnb` as the "best" option for sharp pixels became common knowledge.
- Not sure when, probably around 2010-2015 or so, the "sharp bilinear" upscaling technique has started becoming popular (e.g., ScummVM added support for it even in its 100% software-rendered mode quite early on). This is what our `sharp` shader does by taking advantage of GPU hardware-acceleration, so it's very fast even on low-end hardware like the Raspberry Pi or old computers from around 2010-2012.
- The `sharp`  shader is superior to `openglnb` as it strikes a balance between _pixel -evenness_ and _sharpness_. The nearest-neighbour option, however, which is the most primitive form of upscaling an image, prioritises _sharpness_ and totally ignores pixel-evenness, which results in ghastly looking interference artifacts (moire patterns). This is easy to notice on checkerboard dither patterns and on text. It's less easy to notice on VGA graphics with lots of smooth gradients, but of course the issue is always there.
- All in all, the `openglnb` option is a relic from the past, born out of necessity, and carried forward by not keeping up with the times and adopting more advanced image processing techniques in DOSBox land. There is simply no reason for its continued existence, given we have the superior `sharp` shader.
- For the SDL texture output, we need to keep the `texturenb` option. Note `texture` and `texturenb` arebasically for people who by some bad luck happen to be on hardware that does not have functioning OpenGL drivers. This is extremely rare nowadays, and it was extremely rare 10-15 years ago... Nevertheless, these people can experience bad nearest-neighbour interpolation with the `texturenb` output mode as their "crisp pixel" option, just like in the early days of DOSBox 😎 (Because shaders are not supported in the texture output modes).

Yes, I am passionate about bad image scaling 😅 Looking at nearest-neighbour interpolated output gets me in a really bad mood, no kidding 🥶 

## Example images

All images were captured straight from DOSBox on a 1080p monitor. On 4k screens, the ugly nearest-neighbour scaling artifacts are much less noticeable. Still, that's not a reason for keeping `openglnb`.

For each game, the first image is `openglnb` output (bad), the second `opengl` + the `sharp` shader (good).

---

**This is super important:** You _must_ view all example images at 100% magnification (with 1:1 pixel mapping), otherwise your browser or image viewer _will_ introduce extra unwanted scaling artifacts, rendering the whole comparison completely pointless!

One good way to do that is to open the images in separate browser tabs, ensure they're displayed at 100% magnification, then switch back and forth between the image-pairs with Ctrl+Tab / Ctrl+Shift+Tab to appreciate the differences.

**Double or triple check you're indeed the images at 100% magnification.** If you hover over the image in the new brower tab and cursor turns into a magnifying glass, you need to press the left mouse button to zoom the image to 100% zoom! You need to do this _every time_ you open a new image in a new tab!

---

### Wonderland

Let's start with a super obvious example to train our eyes a little 😄 This is a 800x600 game scaled to fullscreen with a non-integer scaling ratio. Note how the text is extremely hard to read in the `openglnb` output, and there are extra [moire patterns](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern) forming in the checkerboard background pattern due to extreme aliasing.

image

image

### DOS prompt

Note how uneven the lines making up the letters are on the `openglnb` image.

image

image

### Leisure Suit Larry 3 (1)

Again, the unevenness is very easy to see on the `openglnb` output and it's very jarring. Note this is a 320x200 game!

image

image

### Leisure Suit Larry 3 (2)

Note there is mild to moderate moire interference (depending on your sensitivity 😄) on the checkerboard dither pattern in the sky. All EGA games featuring checkerboard dither are highly problematic with `openglnb`.

Note it is impossible _not_ to have moire with 320x200 games with aspect ratio correction on as the pixels are not squares. Then if you add non-integer image scaling to the mix, the results can get really bad.

image

image

# Manual testing

- Tested that `openglnb` is still accepted but a warning is logged and the output is switched to `opengl`.
- Tested that all shaders still work as before.

# Checklist

I have:

### DIFF
--- a/contrib/resources/glshaders/interpolation/catmull-rom.glsl
+++ b/contrib/resources/glshaders/interpolation/catmull-rom.glsl
@@ -101,26 +101,30 @@ COMPAT_VARYING vec2 vTexCoord;
 
 void main()
 {
-	// We're going to sample a a 4x4 grid of texels surrounding the target UV coordinate. We'll do this by rounding
-	// down the sample location to get the exact center of our "starting" texel. The starting texel will be at
-	// location [1, 1] in the grid, where [0, 0] is the top left corner.
+	// We're going to sample a a 4x4 grid of texels surrounding the target
+	// UV coordinate. We'll do this by rounding down the sample location to
+	// get the exact center of our "starting" texel. The starting texel will
+	// be at location [1, 1] in the grid, where [0, 0] is the top left corner.
 	vec2 samplePos = vTexCoord * rubyTextureSize;
 	vec2 texCoord1 = floor(samplePos - 0.5) + 0.5;
 
-	// Compute the fractional offset from our starting texel to our original sample location, which we'll
-	// feed into the Catmull-Rom spline function to get our filter weights.
+	// Compute the fractional offset from our starting texel to our original
+	// sample location, which we'll feed into the Catmull-Rom spline
+	// function to get our filter weights.
 	vec2 f = samplePos - texCoord1;
 
-	// Compute the Catmull-Rom weights using the fractional offset that we calculated earlier.
-	// These equations are pre-expanded based on our knowledge of where the texels will be located,
-	// which lets us avoid having to evaluate a piece-wise function.
+	// Compute the Catmull-Rom weights using the fractional offset that we
+	// calculated earlier. These equations are pre-expanded based on our
+	// knowledge of where the texels will be located, which lets us avoid
+	// having to evaluate a piece-wise function.
 	vec2 w0 = f * (-0.5 + f * (1.0 - 0.5 * f));
 	vec2 w1 = 1.0 + f * f * (-2.5 + 1.5 * f);
 	vec2 w2 = f * (0.5 + f * (2.0 - 1.5 * f));
 	vec2 w3 = f * f * (-0.5 + 0.5 * f);
 
-	// Work out weighting factors and sampling offsets that will let us use bilinear filtering to
-	// simultaneously evaluate the middle 2 samples from the 4x4 grid.
+	// Work out weighting factors and sampling offsets that will let us use
+	// bilinear filtering to simultaneously evaluate the middle 2 samples
+	// from the 4x4 grid.
 	vec2 w12 = w1 + w2;
 	vec2 offset12 = w2 / (w1 + w2);
 
--- a/contrib/resources/glshaders/interpolation/catmull-rom.glsl
+++ b/contrib/resources/glshaders/interpolation/catmull-rom.glsl
@@ -97,59 +97,6 @@ precision mediump float;
 uniform vec2 rubyTextureSize;
 uniform sampler2D rubyTexture;
 
-/*
-	The following code allows the shader to override any texture filtering
-	configured in DOSBox. if 'output' is set to 'opengl', bilinear filtering
-	will be enabled and OPENGLNB will not be defined, if 'output' is set to
-	'openglnb', nearest neighbour filtering will be enabled and OPENGLNB will
-	be defined.
-
-	If you wish to use the default filtering method that is currently enabled
-	in DOSBox, use COMPAT_TEXTURE to lookup a texel from the input texture.
-
-	If you wish to force nearest-neighbor interpolation use NN_TEXTURE.
-
-	If you wish to force bilinear interpolation use BL_TEXTURE.
-
-	If DOSBox is configured to use the filtering method that is being forced,
-	the default hardware implementation will be used, otherwise the custom
-	implementations below will be used instead.
-
-	These custom implemenations rely on the `rubyTextureSize` uniform variable.
-	The code could calculate the texture size from the sampler using the
-	textureSize() GLSL function, but this would require a minimum of GLSL
-	version 130, which may prevent the shader from working on older systems.
-*/
-
-#if defined(OPENGLNB)
-#define NN_TEXTURE COMPAT_TEXTURE
-#define BL_TEXTURE blTexture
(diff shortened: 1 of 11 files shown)


weirddan455:
I'm fine with killing this.  Sharp looks a lot better from those pictures.  I found a couple more remnants you may want to delete as well:

https://github.com/dosbox-staging/dosbox-staging/blob/ffea16b2a60dbcb545654246e1d67d7af9eb55aa/src/gui/sdlmain.cpp#L1530-L1531

https://github.com/dosbox-staging/dosbox-staging/blob/ffea16b2a60dbcb545654246e1d67d7af9eb55aa/src/gui/sdlmain.cpp#L2007-L2012

johnnovak:
@weirddan455 Addressed your comment plus I've done some further cleanup in separate commits, so you might wanna take a look at those. I've retested everything again, can't find any problems.

weirddan455 (approved):
Looks good

PR #3674, opened 2024-05-14
Make GCC 11 / Ubuntu 22.04 the minimum build target

johnnovak:
# Description

As we discussed today, let's move the minimum GCC version forward to 11 on Linux. That means bumping the GCC version of the "minimal build" target from 9 to 11. Staying on 9 just makes it a lot harder to use the nice new C++20 features. Supporting the last 2 Ubuntu LTS releases seems like a good approach; 3 is perhaps pushing it a little bit.

People who desire to compile Staging on legacy systems are welcome to check out an earlier release tag; those will still build with GCC 9, 10, etc.

Some notes:

- The ARMv7, ppc64 and aarch64 targets seem to use some custom Docker image created by kcgen if I'm not mistaken to cross-compile Staging for those platforms. I haven't touched those, neither am I going to. Feel free to tinker with them or maintain them if they're important for you, otherwise they will be disabled / removed the first time they cause a nuissance.
- We should eventually add a new 24.04 target; Linux guys, please step forward 😎 

# Relates issues

The below PR can only go in after this because of some C++20 features that are unsupported on GCC 9:

- https://github.com/dosbox-staging/dosbox-staging/pull/3650

# Manual testing

None.

# Checklist

I have:

### DIFF
--- a/.github/meson/native-clang-12.ini
+++ b/.github/meson/native-clang-12.ini
@@ -1,3 +1,0 @@
-[binaries]
-c = ['ccache', 'clang-12']
-cpp = ['ccache', 'clang++-12']
--- a/.github/meson/native-clang-14.ini
+++ b/.github/meson/native-clang-14.ini
@@ -1,3 +1,0 @@
-[binaries]
-c = ['ccache', 'clang-14']
-cpp = ['ccache', 'clang++-14']
--- a/.github/meson/native-gcc-10.ini
+++ b/.github/meson/native-gcc-10.ini
@@ -1,3 +1,0 @@
-[binaries]
-c = ['ccache', 'gcc-10']
-cpp = ['ccache', 'g++-10']
--- a/.github/meson/native-gcc-7.ini
+++ b/.github/meson/native-gcc-7.ini
@@ -1,3 +1,0 @@
-[binaries]
-c = ['ccache', 'gcc-7']
-cpp = ['ccache', 'g++-7']
--- a/.github/meson/native-gcc-9.ini
+++ b/.github/meson/native-gcc-9.ini
@@ -1,3 +1,0 @@
-[binaries]
-c = ['ccache', 'gcc-9']
-cpp = ['ccache', 'g++-9']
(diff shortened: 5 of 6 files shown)


FeralChild64 (approved):
Looks OK. GCC 11 is, IMHO, a reasonable minimum.

shermp:
I'm the creator of the Linux Docker images. It's easy enough to bump the debian versions as required to maintain compatibility. I think they're useful to keep (especially the ARM ones) to ensure we aren't introducing new code that stops those archs from compiling.

johnnovak:
> I'm the creator of the Linux Docker images. It's easy enough to bump the debian versions as required to maintain compatibility. I think they're useful to keep (especially the ARM ones) to ensure we aren't introducing new code that stops those archs from compilingI.

Thanks for volunteering 👍🏻 Yes my preference if to keep them, especially the ARM ones, like you said. I just did not want to fall into the Docker image generation rabbit hole, I've never done that before.

Can you please raise a separate PR for this and create some succinct documentation about upgrading/maintaining those Docker cross-compile images? That would be great because of the bus factor, etc. 😏

shermp:
> Can you please raise a separate PR for this and create some succinct documentation about upgrading/maintaining those Docker cross-compile images? That would be great because of the bus factor, etc. 😏

I was thinking this morning about how to improve (and simplify?) the docker image creation and tagging. I'm thinking of switching to scheme of something like `debian-cross:<deb_vers>-<arch>-<checksum of dockerfile and apt sources>`, and maybe (manual?) dosbox tagged versions for posterity.

johnnovak:
> I was thinking this morning about how to improve (and simplify?) the docker image creation and tagging. 

The build system needs a lot of simplifications, we all agree on that, so yeah, go for it.

> I'm thinking of switching to scheme of something like `debian-cross:<deb_vers>-<arch>-<checksum of dockerfile and apt sources>`, 

That sounds like an improvement.

> and maybe (manual?) dosbox tagged versions for posterity.

You mean archieve snapshots of these Docker images whenever a tag is pushed? That is a good idea. Just be aware it will need to handle tag deletions & recreations as well (I "push forward" the release tag a lot when preparing a release).

Alternatively, if you don't want to automate this, the snapshotting could be a final one-time manual process done right after a public release.

PR #3676, opened 2024-05-16
fix msys2 compilation error with cdrom_win32.cpp

strikersix23:
in msys2 it would through undefined when declaring IOCTL_CDROM_LOAD_MEDIA so i change it to IOCTL_DISK_LOAD_MEDIA this fixed compilation and i have had no issues testing the build that it produced.
I have ran all TESTs and asan build there has been no breaking issues on mys2

### DIFF
--- a/src/dos/cdrom_win32.cpp
+++ b/src/dos/cdrom_win32.cpp
@@ -298,7 +298,7 @@ bool CDROM_Interface_Win32::LoadUnloadMedia(bool unload)
 		return false;
 	}
 	DWORD control_code       = unload ? IOCTL_STORAGE_EJECT_MEDIA
-	                                  : IOCTL_CDROM_LOAD_MEDIA;
+	                                  : IOCTL_DISK_LOAD_MEDIA;
 	LPVOID input_buffer      = NULL;
 	DWORD input_buffer_size  = 0;
 	LPVOID output_buffer     = NULL;
--- a/src/dos/cdrom_win32.cpp
+++ b/src/dos/cdrom_win32.cpp
@@ -298,7 +298,7 @@ bool CDROM_Interface_Win32::LoadUnloadMedia(bool unload)
 		return false;
 	}
 	DWORD control_code       = unload ? IOCTL_STORAGE_EJECT_MEDIA
-	                                  : IOCTL_DISK_LOAD_MEDIA;
+	                                  : IOCTL_STORAGE_LOAD_MEDIA;
 	LPVOID input_buffer      = NULL;
 	DWORD input_buffer_size  = 0;
 	LPVOID output_buffer     = NULL;


weirddan455:
Let me test on Windows before I give the OK on this.  I can't find Microsoft documentation on `IOCTL_DISK_LOAD_MEDIA`.  If it's just an alias to `IOCTL_CDROM_LOAD_MEDIA` it's probably fine.

johnnovak:
@strikersix23 Small MSYS2 compatibility fixes like this are fine, but do note we've removed official support for it. MSVC is our only supported build platform for Windows, so no bigger compatibility fixes than a few lines please and no warranties 😄

weirddan455:
It's not just an alias, it's a different ioctl.  I think `IOCTL_STORAGE_LOAD_MEDIA` is the better one to use here.  I used `IOCTL_STORAGE_EJECT_MEDIA` for the eject case and I'm not sure why I did do the matching `IOCTL_STORAGE_LOAD_MEDIA` for the load case.   From a comment on the MSVC headers:

```
// The following device control codes are common for all class drivers.  The
// functions codes defined here must match all of the other class drivers.
//
// Warning: these codes will be replaced in the future with the IOCTL_STORAGE
// codes included below
```

And link to documentation:  https://learn.microsoft.com/en-us/windows/win32/api/winioctl/ni-winioctl-ioctl_storage_load_media

Note I wasn't ever able to test this feature on my computer.  It's a desktop but it has a laptop style CDROM that lacks a motor to pull the tray back in (it can be ejected programmatically but must be manually loaded back).  I doubt many DOS games use this feature either way though.

So just change `DISK` to `STORAGE` in your PR and I'll approve it.  If you want, you can also do an easy fixup of some unused variable warnings in that same file (apparently MSVC didn't pick up on those).  In the functions I've stubbed out, you can just put `[[maybe_unsed]]` in the function declaration like for example:

```cpp
bool CDROM_Interface_Win32::ReadSectors([[maybe_unused]] PhysPt buffer, [[maybe_unused]] const bool raw,
                                        [[maybe_unused]] const uint32_t sector,  [[maybe_unused]] const uint16_t num)
```

This is optional though, I'll approve it either way once you change the ioctl and I appreciate the fix 😄 

I'd like to keep MSYS2 working as long as it's not too much hassle.  I sometimes prefer it over MSVC when I'm on Windows, although it has its downsides.  It was proving to be too much trouble for CI because of the rolling release nature of its packages as @johnnovak pointed out.  It can be pretty nice for local dev work though.

johnnovak:
> I'd like to keep MSYS2 working as long as it's not too much hassle. I sometimes prefer it over MSVC when I'm on Windows, although it has its downsides. It was proving to be too much trouble for CI because of the rolling release nature of its packages as @johnnovak pointed out. It can be pretty nice for local dev work though.

Fair point, and I'm not against it _as long as_ it doesn't complicate anything and people won't start littering the code with MSYS2 preprocessor macros. That's a hard no.

Then, as I said, it will inevitably diverge because of the rolling nature of MSYS2. SDL bugs/regressions come and go between minor releases, so we might not have workarounds in the code for some new SDL library regressions, etc., then weird stuff might happen at rutime (it happened before; some of their mouse related regressions caused a crash when resizing the window or something).

weirddan455:
>  Fair point, and I'm not against it as long as it doesn't complicate anything and people won't start littering the code with MSYS2 preprocessor macros. That's a hard no.

Yeah, I don't think we've ever had to do that.  This is an edge case where MSYS2 has removed some deprecated `#define`'s from their header files but we probably should have been using the non-deprecated ioctls regardless.  I agree though, I dislike having `#ifdef`'s everywhere.  Luckily, in Staging they're limited to a few platform-specific areas.

>  Then, as I said, it will inevitably diverge because of the rolling nature of MSYS2. SDL bugs/regressions come and go between minor releases, so we might not have workarounds in the code for some new SDL library regressions, etc., then weird stuff might happen at rutime (it happened before; some of their mouse related regressions caused a crash when resizing the window or something).

That's no different than compiling on Arch Linux for example.  I believe we have a pinned SDL on MSVC with vcpkg and a Meson wrap on Mac OS X but on Linux, it will use whatever the distro shipped.

Note that I do agree that this is not ideal.  This is one of the reasons I switched my system away from Arch eventually settling on Debian so I don't get major version bumps.  That and I didn't want to upgrade to KDE 6 immediately after it shipped.  I definitely prefer having something non-rolling for CI and binary releases.

strikersix23:
As I prefer the msys2 builds over msvc, but I do understand it can be difficult to mantain. I can try my best to make sure nothing is broken, but I cant guarantee id be able to maintain it, I'm surprisingly garbage at coding. But this was an exception case for me. @weirddan455. I agree with you the code I used was just a quick patch to fix an error and convoluted code makes it harder

johnnovak:
> I believe we have a pinned SDL on MSVC with vcpkg and a Meson wrap on Mac OS X but on Linux, it will use whatever the distro shipped.

Yes, that's the current situation, and the plan is to move to CMake + vcpkg on all three platforms to make things uniform.

strikersix23:
vcpkg does support the msys2 target as well. Has msvc caught up with the msys2 packages? I heard it was lacking with libslirp and a few other

johnnovak:
> vcpkg does support the msys2 target as well. Has msvc caught up with the msys2 packages? I heard it was lacking with libslirp and a few other

All I can say libslirp works and it has all the packages we currently need.

strikersix23:
It actually sounds good though I'm used to unix build commands and meson doesn't like the comand prompt that well when it comes to path handling and package search for me

johnnovak:
> It actually sounds good though I'm used to unix build commands and ninja + meson doesn't like the comand prompt that well when it comes to path handling and package search for me

Do note you'll still be able to just build from the command line. No one will force you to use the MSVC GUI 😄  You can compile MSVC projects from the CLI—that's how our CI build job does it too 😄

PR #3678, opened 2024-05-17
Create new cross compiler docker images and adjust workflows

shermp:
# Description

The previous cross compiler docker images were somewhat complex and prone to causing issues. This is an attempt to remedy those issues.

The previous version used two docker images, a "base" image, then the final image. I have simplified this to a single Dockerfile, so one image per arch.

The docker images are now tagged using the following format: `<debian_vers>-<arch>-<dockerfile_apt_hash>`. This means that the dosbox version is no longer used, which has recently caused issues after 0.82 dev was started. Instead, the contents of the Dockerfile and apt package list(s) are hashed and used.

The images used to be build in a separate workflow. This could lead to a situation where the linux workflow might depend on a workflow that hadn't yet completed, causing the linux workflow to abort. Now, images are build if required directly in the linux workflow. A utility script has been created to aid performing the above file hashing process, as well as checking if an image tag exists in the container registry.

To aid reusablilty, the docker image build steps are in a new composite action, which is currently used by the linux workflow. In the future, we can use it in a separate workflow to create images on demand or a schedule if required.

Oh, I've also knocked out ccache in the linux and msys2 workflows.

# Manual testing

I kept trying until green ticks everywhere :)

But also, the docker images were build locally using `podman build --build-arg DEB_VERS=11 --build-arg DKPG_ARCH=arm64 -t debian-cross:local`

# Checklist

I have:


shermp:
Also, feel free to suggest alterations to the linux build matrix, I feel we could use some changes there, just not sure what.

scripts/docker-cross-tag.sh
```diff
@@ -0,0 +1,55 @@
...
+  hash-files         Compute a hash of specified files for use in a docker image tag
+                     docker image tag. Command takes one or more file paths as options.
+"
+}
+
+tag_exists_in_ghcr() {
+    local image="$1"
+    local tag="$2"
+
+    # jq and curl are required
```
johnnovak, line 25: Maybe move this comment to the top into a "prerequisites" section?

scripts/docker-cross-tag.sh
```diff
@@ -0,0 +1,55 @@
...
+# Copyright (C) 2024-2024  The DOSBox Staging Team
+
+usage() {
+    printf "%s\n" "\
+Usage: $0 FUNC OPTIONS
+
+Utility script to assist with docker image tags
+
+FUNC must be one of:
+  tag-exists         Test if image tag exists in GHCR.
```
johnnovak, line 13: It would be better to spell it out:

```
Test if image tag exists in the GitHub Container Registry (GHCR).
``

That would give people unfamiliar with GHCR at least a trail to follow.

File: scripts/docker-cross-tag.sh [hunk 3]
```diff
@@ -0,0 +1,55 @@
...
+        printf "%s\n" "jq not found!" && exit 1
+    fi
+    if ! command -v curl &> /dev/null ; then
+        printf "%s\n" "curl not found!" && exit 1
+    fi
+
+    # Get an anonymous token from Github to make further API calls.
+    # No need for accounts as the images are public.
+    token=$(curl -s https://ghcr.io/token\?scope="repository:dosbox-staging/${image}:pull" | jq -r .token)
+    # Get the list of tags from the registry, and attempts to find provided tag
```
johnnovak, line 36: typo: attempt (lose the s)

scripts/docker-cross-tag.sh
```diff
@@ -0,0 +1,55 @@
...
+    # in the list using jq. Prints 'true' if tag found, or 'false' otherwise
+    curl -s -H "Authorization: Bearer $token" "https://ghcr.io/v2/dosbox-staging/${image}/tags/list" \
+        | jq --arg T "$tag" '.tags | any(.==$T)'
+}
+
+files_hash() {
+    # Hash the hash of the contents of the provided list of files.
```
johnnovak, line 43: I was surprised that you're hashing twice, so googled it and found this:

https://crypto.stackexchange.com/questions/50017/why-hashing-twice

So this is some kind of a mitigation against "length extension attacks", whatever that means, to improve security.

But we're not doing anything security related here, we just need an unique hash, so maybe use a single hash to remove confusion?

.github/composite-actions/build-debian-docker/action.yml
```diff
@@ -0,0 +1,59 @@
+name: "Build Debian Cross Docker"
```
johnnovak, line 1: Nitpicky: "Build Debian Docker image for cross-compiling"

.github/composite-actions/build-debian-docker/action.yml
```diff
@@ -0,0 +1,59 @@
...
+    # Log into the Github container registry using
+    # the provided credentials passed in the workflow.
+    - name: Log in to the Container registry
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ inputs.user }}
+        password: ${{ inputs.pass }}
+
+    - name: Set up Docker Buildx
```
johnnovak, line 33: typo: Buildx

packages/Dockerfile.debian
```diff
@@ -1,36 +1,64 @@
...
+    && apt-get -y install --no-install-recommends \
         meson \
         crossbuild-essential-${DPKG_ARCH} \
     && /usr/share/meson/debcrossgen --arch ${DPKG_ARCH} -o ${MESON_CROSS_FILE} \
     && apt-get -y remove meson \
     && apt-get -y autoremove \
     && rm -rf /var/lib/apt/lists/*
 
-RUN apt-get update &&  apt-get -y install --no-install-recommends \
+# Install Dosbox-staging dependencies using the appropriate architecture
```
johnnovak, line 46: minor: DOSBox Staging

.github/workflows/linux.yml
```diff
@@ -24,29 +24,17 @@ concurrency:
...
-  CCACHE_DIR:           "/dev/shm/.ccache"
-  CCACHE_DIRECT:        "true"
-  CCACHE_MAXSIZE:       "128M"
-  CCACHE_SLOPPINESS:    "pch_defines,time_macros"
-  CCACHE_UMASK:         "002"
-
-  CLAMDB_DIR:      "/var/lib/clamav"
-
+  # Docker commmand to run if performing cross builds.defaults:
+  # The command remnoves the container on exit, runs the
```
johnnovak, line 28: typo: removes

.github/workflows/windows-msys2.yml
```diff
@@ -249,22 +224,6 @@ jobs:
...
-          echo "yesterday=$(date --date=yesterday -I)" >> $GITHUB_OUTPUT
-
-      - uses:  actions/cache@v4.0.1
-        id:    cache-ccache
-        with:
-          path: ${{ steps.prep-ccache.outputs.dir }}
-          key:  ccache-msys2-${{matrix.conf.sys}}-release-${{ steps.prep-ccache.outputs.today }}-3
-          restore-keys: |
-            ccache-msys2-${{matrix.conf.sys}}-release-${{ steps.prep-ccache.outputs.yesterday }}-3
-
```
johnnovak, line 267: Just commenting because you've touched the MSYS2 workflow. You might want to add the new version handling to this as well:

```
      - name: Inject version string
        run: |
          set -x
          VERSION=$(./scripts/get-version.sh version-and-hash)
          echo "VERSION=$VERSION" >> $GITHUB_ENV

```

scripts/docker-cross-tag.sh
shermp: Nah, nothing security related. It just seemed the easiest way to get a single hash of multiple files. When you run sha256sum on multiple files, it spits out a hash per file. The other way would be to concatenate the files before hashing. This seemed simpler to me but I dunno.

johnnovak (approved):
Great work @shermp and a very nice and succinct summary of what has been changed 🎉

Most of if it went above my head, but at least I was able to make a few comments on the typos 😎  Definitely looks cleaner than before, and if it works, it works! 🥳

scripts/docker-cross-tag.sh
shermp: Ok, I've decided to just `cat` the files and pass it to `sha256sum`. Will make the intention clearer I think.

scripts/docker-cross-tag.sh
shermp: Done. I've moved all prerequisite checking to a function as well.

scripts/docker-cross-tag.sh
shermp: Also lost a `cut` that wasn't needed.

.github/workflows/linux.yml
shermp: I reworded this comment slightly as well. It was mentioning `HOME` directory, instead of environment.

shermp:
Thanks for the review @johnnovak 

I think I've resolved your review comments. I reworked the utility script a bit, so you may wish to review the changes.

scripts/docker-cross-tag.sh
```diff
@@ -0,0 +1,63 @@
...
+
+usage() {
+    printf "%s\n" "\
+Usage: $0 FUNC OPTIONS
+
+Utility script to assist with docker image tags
+
+FUNC must be one of:
+  tag-exists         Test if image tag exists in the GitHub Container Registry (GHCR).
+                     Takes two options: IMAGE_NAME, TAG.
```
johnnovak, line 19: One last thing, maybe remove the comma from after IMAGE_NAME? I find it a bit misleading.

johnnovak:
> Thanks for the review @johnnovak
> 
> I think I've resolved your review comments. I reworked the utility script a bit, so you may wish to review the changes.

Cheers, looks very good now from my side 👍🏻 One last very minor comment, then maybe wait a bit until our Linux guys also have a chance to take a look.

packages/Dockerfile.debian
```diff
@@ -1,36 +1,64 @@
...
+# Enable cross compile architecture. Abort if not set correctly
+RUN case "$DPKG_ARCH" in \
+        arm64|armhf|ppc64el) dpkg --add-architecture $DPKG_ARCH ;; \
+        *)  echo "DPKG_ARCH not set!" && exit 1 ;; \
+    esac
 
-ENV MESON_CROSS_FILE=/usr/local/share/meson/cross/dosbox-cross
+ENV DEBIAN_FRONTEND=noninteractive
+# Install common base package
+RUN apt-get update && apt-get -y install --no-install-recommends \
```
weirddan455, line 17: Do we want to also do an `apt-get upgrade` here following the `apt-get update`?  Update only downloads the latest list of packages.  Upgrade actually upgrades the packages.

packages/Dockerfile.debian
```diff
@@ -1,36 +1,64 @@
...
+RUN echo "Debian version is ${DEB_VERS}"
 
 COPY ./debian-${DEB_VERS}-docker-apt.txt ${DOSBOX_APT_LIST}
 
-RUN mkdir -p $(dirname ${MESON_CROSS_FILE})
-RUN apt-get update && apt-get -y install --no-install-recommends \
+# Install cross toolchain and generate meson 
+# cross file using debcrossgen
+RUN  mkdir -p $(dirname ${MESON_CROSS_FILE}) \
+    && apt-get update \
```
weirddan455, line 37: This is running `apt-get update` again after we've already done it once above.  Does this need to be here?

packages/Dockerfile.debian
```diff
@@ -1,36 +1,64 @@
...
         meson \
         crossbuild-essential-${DPKG_ARCH} \
     && /usr/share/meson/debcrossgen --arch ${DPKG_ARCH} -o ${MESON_CROSS_FILE} \
     && apt-get -y remove meson \
     && apt-get -y autoremove \
     && rm -rf /var/lib/apt/lists/*
 
-RUN apt-get update &&  apt-get -y install --no-install-recommends \
+# Install DOSBox Staging dependencies using the appropriate architecture
+RUN apt-get update \
```
weirddan455, line 47: Now we're updating a third time?

weirddan455 (commented):
Note that I'm not all that familiar with Docker but this stood out as potentially wrong to me.

Would it be better to do a single `apt-get update && apt-get upgrade` at the start and then only do `apt-get install`'s afterwards?

packages/Dockerfile.debian
shermp: The base Debian image this builds on is already updated regularly. I see no reason to upgrade packages again, which just increases the final image size.

packages/Dockerfile.debian
shermp: Notice the `rm -rf /var/lib/apt/lists/*` at the end of the command. This is standard practice when creating a docker image to minimize layer size, as each layer is immutable.

On the other hand, the debian package list probably isn't *that* large, but still.

packages/Dockerfile.debian
shermp: See previous comment.

weirddan455 (approved):
Thanks for the explanation @shermp   Makes sense to me the way you've got it then.  Appreciate you maintaining this CI stuff :smile:

johnnovak (approved):
Thanks for simplifying our CI stuff @shermp and keeping the quality of build related scripts high 🎉 Merging!

Issue #366, opened 2020-05-10
SVGA S3 Trio emulation is faulty for games using VESA LFB (was: Duke Nukem 3D title screen flickers)


bmunger:
I have started testing against some of my DOS games and found this bug.  Seems to only show up at the title screen text after the 3D Realms logo.  Parts of the upper character flicker.  This does not happen in upstream Dosbox.  Also tested with opengl/openglb.  This is on a 64 bit Linux build (Ubuntu 18.04).

dreamer:
@bmunger when you're saying "upstream Dosbox" do you mean 0.74-3 or SVN?

Does the problem happen when you try different `machine` setting, like:
```
machine = vesa_nolfb
```
?

Also, does it happen when you start with `fullscreen = true`?

*(I don't have DN3D set up for testing anywhere near me, so was not able to reproduce the problem yet)*

bmunger:
@dreamer I run the Dosbox SVN newest commit and compile it myself.

I can confirm that the issue goes away with vesa_nolfb

Setting fullscreen = true does not change behavior

Something to do with machine = svga_s3 apparently.  It's consistent though.  I also run Nvidia binary drivers.

dreamer:
I had a hunch about `vesa_nolfb` because I observed a similar regression in SVN when 64-bit dynarec was implemented (in my case "flickering" regression was in TOPBENCH) - relevant code was not touched at all, but faster emulation speed triggered some kind of timing issue in `svga_s3`. We had some more changes, that *might* result in faster emulation speed compared to SVN, so maybe it had negative effect on this timing problem in svga_s3 as well…

@bmunger please add info about what version of dosbox-staging are you testing exactly and what SVN revision you are testing against - it will help us with bisecting the issue.

If you'll have any problems with playing while using `vesa_nolfb`, then report them as well :)

This issue stays **open** until we'll figure out exactly why this regression happened.

bmunger:
@dreamer 
No problem.
dosbox-staging: commit 8a3a9bfc7565ad2c1567827bdd1143bb84501f9e
dosbox-svn: 4339

I just updated dosbox-svn and it seems to also have this issue now.  Going to attempt to figure out where it broke.  I had updated dosbox-svn fairly recently. as well.

dreamer:
[...]

PR #3681, opened 2024-05-19
Introduce a better way to set the CPU cycles with new defaults

johnnovak:
Me: Let's change the default cycles settings. Sounds like a relaxing two-hour job! 

_...
...
(time passes)
..._

_~Two~ Three weeks and 4-5 rewrites later..._

# Release notes

## Simplified CPU cycles settings

The `cycles` setting has been a source of a lot of confusion, so we replaced it with a much simpler way to configure the emulated CPU speed (believe it or not, there are over 20 valid `cycles` value permutations!)

The new streamlined CPU cycles settings are as follows:

- `cpu_cycles` — Global or real mode only cycles setting.
- `cpu_cycles_protected` — Optional protected mode only cycles setting.
- `cpu_throttle` — Whether the CPU emulation should be dynamically throttled if your host CPU cannot keep up.

Please refer to the settings' descriptions to learn more (e.g., run `config -h cpu_cycles | more` from the DOS prompt).

Here are a few concrete examples:

### Set fixed speed globally

This will roughly emulate the speed of the venerable i486DX2-66 in both real and protected mode. `cpu_cycles_protected = auto` means `cpu_cycles` controls the emulated CPU speed globally, both for real and protected mode.

```ini
[cpu]
cpu_cycles = 25000
cpu_cycles_protected = auto
```

###  Different real and protected mode speeds with throttling

20k cycles for real mode programs, and 400k for protected mode. Throttling is enabled, so if your host machine can't keep up, the actual number of emulated CPU cycles per millisecond will be dynamically lowered.

```ini
[cpu]
cpu_cycles = 20000
cpu_cycles_protected = 400000
cpu_throttle = on
```

### Max speed

This will make the CPU emulation run as fast as your computer can handle it. Pretty useful when compiling programs, rendering 3D images, or playing late '90s 3D games in high SVGA/VESA resolutions (as long as they can run glitch-free on very fast CPUs).

```ini
[cpu]
cpu_cycles = max
cpu_cycles_protected = max
```

> **Notes**
> The legacy `cycles` setting is still available for backward compatibility with existing configs. When `cycles` is set, the new CPU settings are simply ignored. Do note, however, that we will remove support for `cycles` in the future; please start migrating your configs to the new settings now.

## Revised CPU cycles defaults

> **TL;DR**
> The default emulated CPU speed (cycles) setting for protected mode programs has been changed from "max" to 60,000 cycles.  The real mode default of 3000 cycles is unchanged. Read on to get the details! 

> **Note**
> When "legacy cycles mode" is enabled by setting `cycles` to a valid value, the old defaults are applied (3000 for real mode, and "max cycles" for protected mode). 

DOSBox traditionally defaulted to 3000 cycles for real mode programs, and "max cycles" (as fast as your host CPU can go) for protected mode programs. 3000 cycles for real mode is a good middle-of-the-road setting that gets most games running (even if it not optimally). The "max cycles" setting, on the other hand, is quite problematic for a couple of reasons:

- A significant number of games crash, misbehave, or manifest subtle bugs when the emulated CPU is "too fast". In the early days of DOSBox, `cycles = max` was not a problem because the much slower host CPUs from 10-20 years ago acted as a "natural throttle" on the effective emulation speed. However, many of those old speed-sensitive games that used to work with `cycles = max` are now broken on today's fast CPUs.

- The `max` setting is not transferable across setups as it depends on the host CPU's speed. Games configured for `cycles = max` might work fine on _your_ particular machine, but they could break on your friend's faster computer, or on yours a few years later when you get your long-awaited CPU upgrade. `max N%` cycles settings (e.g., `max 70%`) are plagued by the same problem.
  
- `cycles = max` often causes audio glitches and drop-outs in games. It's best to use the _lowest fixed cycles value_ that runs the game at an acceptable speed (read more tips about that [here](https://dosbox-staging.github.io/getting-started/beneath-a-steel-sky/#adjusting-the-emulated-cpu-speed)).

- Some DOS programs can _really_ freak out when the emulated CPU is being dynamically changed at runtime, which is what `max` tends to do.

All in all, `max` for protected mode is not a good default in 2024, so we've changed it to fixed 60,000 cycles, which is roughly equivalent to slow Pentium speeds. As protected mode games started appearing in the early to mid-90s, this should run the vast majority of them fairly optimally out-of-the-box.

---

# Technical notes

- I haven't changed the core cycles handling logic in the CPU emulation core at all. All that I've changed is the external interface to the cycles config (the settings).

- When setting `cycles` explicitly, the new cycles config settings will be blanked and ignored. Everything will work as before, no changes whatsoever, and I was careful not to touch the old code at all (barring some simple refactor jobs).

- To achieve the above, I needed to add the new `DeprecatedAndAllowed` config type. I could not just use `Deprecated` because that either removes support for the deprecated setting wholesale (so you can no longer set it at all), or it "redirects" the setting to a new config property. `DeprecatedAndAllowed` is handy for more elaborate deprecation scenarios where you still want to handle the old config for backward compatibility—we'll need it in the new config system as well (so it's not just a hack).

- Once you've set the `cycles` setting to something, there's no going back to "modern mode" in the current emulator session—you need to restart DOSBox to be able to set the "modern" settings again.

- The legacy `cycles` setting did not allow setting separate fixed cycles values for real and protected mode. You could set something like `auto 12000 max limit 50000`, which gave you fixed 12000 in real mode and "max" in protected mode, but throttled to 50k cycles max. Now you can set separate real and protected values, and enable/disable throttling globally for both which is more flexible and a clear operational model.

- This has some real-world significance as I've run into demos that would run fine with, say, `cycles = 120000`, but would consistently crash with `cycles = auto 3000 max limit 120000`. Clearly, the "fixed cycles" and "fixed cycles with throttling" code paths are not exactly the same (but figuring out the root cause for that is out of scope).

- In "modern mode", you can also set "max cycles" (`max` setting) for protected mode only, or for both real and protected mods. However, support for the "_gimme N% of the max performance of a single host CPU core"_ (`cycles = max N%`) has been dropped. It doesn't make sense in a world where even the bottom-of-the-barrel CPUs released in the last 15 years have at least 4 physical cores, and most desktop CPUs have 8 to 24 or more. Stuff like `cycles = max 50%`  would encourage people "scaling down" the "max" setting to get the game running on _their machine_, and the resulting settings would not be transferable across computers. There's no place for "half-max", "3/4 max", and bullshit like that... There are enough CPU cores in the system for other jobs, even if the CPU emulation pegs a single core.

I've removed the localisation of the cycles display from the title bar as it simplified things, and IMO, that's taking localisation too far. Everybody should understand what "cycles" is; the config settings have "cycles" in their names, and so on. But @FeralChild64 is welcome to reinstate it in a subsequent PR if he thinks this is important.

- In "legacy mode", the cycles display in the title bar is as before. In "modern mode", only these variations are possible:
  - `1234 cycles/ms`
  - `1234 cycles/ms (throttled)`
  - `max cycles/ms`

- The valid fixed cycles range is clamped to 50 to 2,000,000 in "modern mode". Note there is a hard 2,000,000 upper limit to the auto-cycles and "max cycles" stuff, so effectively I've introduced the lower limit of 50.

# Manual testing

Shoehorning the modern settings took quite a few attempts to make everything work correctly. In particular, the logic for switching between real and protected mode is quite "delicate", so it's important to test these transitions thoroughly to ensure no change in behaviour.

I've conducted all these tests on both macOS (ARM, Apple M1) and my PC (x86_64, Ryzen 7900).

As you can see, I've tested the living shit out of it 😎

> **Important**
> - Always start Staging with `--noprimaryconf`.
> - Use the _release build_ for the tests, then repeat with the debug build to make sure no asserts are tripped.
> - Optional, but recommended — create a custom release build for testing:
> 	- In `dosbox.cpp`, find `increase_ticks()` and change `#if 0` to `#if 1` inside the function, then create a release > build. This will log the momentary cycles values in auto-cycles mode.

Use the included `#cpu-tests.zip` test pack for the tests:

[#cpu-tests.zip](https://github.com/dosbox-staging/dosbox-staging/files/15387474/cpu-tests.zip)

Contents:

- `TESTS\KOPIO.EXE` — Late '90s protected mode demo. Needs high cycles values to run smoothly, so the lower FPS gives a good indication that lower cycles values are in effect.

- `TESTS\TUBE.COM` — Real mode 256-byte intro (yes!) that needs high cycles values to run smoothly (around ~80k to get stable FPS without frame drops).

- `TESTS\FT2.EXE` — **Fast Tracker II**. It uses protected mode, so it's useful to get quickly in and out of pmode.

- `TESTS\SC.BAT` — Helper batch file to print the current moder `cpu_*` settings to the console.

- `NO\*.*` — **No by NoooN** (demo). Uses protected mode, good general test case.

- `STARS\*.*` — **Stars by NoooN** (demo). Uses protected mode, good general test case, and it does some weird switching between real and pmode at startup with highlighted a few regressions with my modern implementation (the whole process crashed after a few seconds of starting the demo).

### Basic tests

- Start with `--noprimaryconf` and the included `dosbox.conf` (note all cycles settings are commented out; don't touch them yet).
- Confirm there are no cycles related warning messages in the logs with `CPU:` prefix.
- Run `SC.BAT` and confirm the default values:
	- `cpu_cycles` — 3000
	- `cpu_cycles_protected` — 60,000
	- `cpu_throttle` — false
- Run `FT2`. Now we're in pmode, so confirm the title bar shows `60000 cycles/ms`.
- Exit to DOS (**Esc**, then **Enter**).
- We're back to real mode, confirm the title bar shows 3000 cycles.

### Fixed settings

- Start with `--noprimaryconf` and the included `dosbox.conf`.
- Set `cpu_cycles 1234` and confirm the setting in the title bar.
- Set `cpu_cycles_protected 65432` (no change in the title bar, we're in real mode).
- Start `FT2`, confirm 65432 cycles in the title bar.
- Exit, confirm 1234 cycles has been restored.
- Press **Cmd/Ctrl+F12** twice to increase the real mode cycles; the title bar should show 1493 cycles.
- Confirm the `cpu_cycles` setting is kept in sync (just run `cpu_cycles`, or `SC.BAT`).
- Start `FT2`, press **Cmd/Ctrl+F12** three times so the title bar shows 37866 cycles.
- Exit, confirm the title bar shows 1493 cycles, then run SC and confirm the `cpu_*` settings are synced:
	- `cpu_cycles` — 1493
	- `cpu_cycles_protected` — 37,866
	- `cpu_throttle` — false
- Start `FT2` one more time to double-check 37866 is the new pmode setting.
- Exit and confirm we're back to 1493 cycles in real mode.

### Throttled — general

- Start with `--noprimaryconf` and the included `dosbox.conf`.
- Run `cpu_throttle on`, confirm the title bar shows "3000 cycles (throttled)".
- Repeat the "Fixed settings" tests and make sure throttled mode stays on.
- At the end, turn off throttling with `cpu_throttle off`. Confirm that the last set settings have been retained in both real and protected mode.

### Throttling — real mode

- Start with `--noprimaryconf` and the included `dosbox.conf`.
- Set `cpu_throttle on`, then `cpu_cycles 1000000` (1 million). Confirm in the logs that the momentary cycles value however aroud 300k-400k actual cycles.
- Start `TUBE`. Confirm the title bar shows 1000000 cycles, the animation is smooth with no frame drops, and the momentary cycles value is around 100-140k in the logs.
- Exit with `Esc`, then turn off throttled mode (`cpu_throttle off`). Confirm the title bar still shows 10000000. Run `SC`, confirm `cpu_cycles` is still 10000000.
- Turn throttling on again, set `cpu_cycles 30000`.
- Start `TUBE`, confirm the momentary cycles stays at `30000`.
- Don't exit yet! 😎 

### Throttling — protected mode

- Start `KOPIO`. Confirm we're in 60,000 cycles throttled mode (that's the protected mode default). Note the frame rate is quite choppy.
- Exit back to real mode with `Esc`, confirm we're in 1,000,000 cycles throttled mode.
- Set `cpu_cycles_protected 200000` then start `KOPIO`. Note the frame rate is now 100% smooth.
- **Esc**, confirm the correct real mode setting, then start `KOPIO` again.
- Use **Cmd/Ctrl+F11** to decrease the cycles to under 100k, note the animation became choppy.
- Increase the cycles with **Cmd/Ctrl+F12** until the animation becomes smooth again, they keep increasing until about 1 million cycles. Confirm in the logs that the momentary cycles values is around 400-500k because of the throttling. Note down the cycles number in the titlebar.
- Exit with **Esc**, the restart `KOPIO`. Confirm the previous protected cycles setting was restored.
- Check with `SC.BAT` that the config values are correct and reflect the cycles values set by the hotkeys.

### Max cycles — global

- Start with `--noprimaryconf` and the included `dosbox.conf`.
- Set `cpu_cycles max`. Confirm the title bar shows "max cycles/ms" and the logs will show the momentary cycles value (around 350k on my M1 Mac).
- Also confirmt the following warning appeared in the logs:
    - `CPU: Invalid 'cpu_cycles_protected' setting: '60000'; fixed values are not allowed if 'cpu_cycles' is 'max', using 'auto'`
- Run `SC.BAT`, and confirm the config setting values:
    - `cpu_cycles` — `max`
    - `cpu_cycles_protected` — `auto`
    - `cpu_throttle` — `false` (has no effect in "max cycles" mode)
- Start `TUBE` (real mode). Confirm the title bar still shows "max cycles/ms", the animation is smooth with no frame drops, and the momentary cycles value is around 100-140k in the logs. 
- Confirm the increase/decrease cycles hotkeys have no effect.
- Exit with **Esc**, then start `KOPIO` (protected mode). Confirm the same things as per above.
- Set `cpu_throttle on` and confirm "(throttled)" doesn't show up in the title bar after the cycles setting in real mode (so you should see "max cycles/ms" only).

### Max cycles — protected mode only

- Start with `--noprimaryconf` and the included `dosbox.conf`.
- Set `cpu_cycles_protected max` and `cpu_cycles 20000`. Confirm the title bar shows 20,000 cycles.
- Start `TUBE` and confirm it's really laggy at 20k cycles.
- Modify the current cycles value with the hotkeys, then exit with **Esc**.
- Confirm that the real mode cycles value is retained.
- Start `KOPIO` (pmode). The animation should be perfectly smooth at "max cycles", confirm "max cycles" shows up in the title bar, etc.
- Exit with **Esc**, confirm the previously set real mode cycles setting is restored.
- Set `cpu_throttle on` and confirm "(throttled)" doesn't show up in the title bar after the cycles setting in protected mode (so you should see "max cycles/ms" only).

### Cycles limits

- Keep pressing the decrease cycles hotkey and confirm you can't go below 50.
- Set `cpu_throttle on` and `cpu_cycles 1000000` (1 million). Keep pressing the increase cycles hotkey and confirm you can't go above 2,000,000.
- Try setting `cpu_cycles` and `cpu_cycles_throttled` values outside of the valid 50 to 2,000,000 range. Confirm a warning is logged and the value is clamped to the valid range. Also confirm with `SC.BAT` the config settings are kept in sync with the actual clamped values.

## "Legacy mode" tests (`cycles` setting)

TODO

# Checklist

I have:


johnnovak:
@Torinde The reason is original DOSBox introduced that upper limit. As you've mentioned, we're nowhere near reaching that limit anyway on current hardware. E.g., running demanding late 90s demos with `max` cycles on dynarec still only yields around 400-500k cycles on my M1 Mac.

So yeah, very easy to change, but it's pointless, so leaving at 2 mill. Future maintainers in 2050 can bump it up... unless Moore's Law has truly died for good 😄

johnnovak:
> Fair. I just thought a less-arbitrary number would be preferable.

Those "projected values" are pretty arbitrary too, and they're well past the "true DOS era". The popularity of DOS as a platform is basically zero in 2000, and even with that we're being generous. It's "relevancy" as a gaming platform stopped around 1998 or so.

include/dosbox.h
```diff
@@ -74,6 +74,14 @@ void DOSBOX_Init(void);
 
 void DOSBOX_SetMachineTypeFromConfig(Section_prop* section);
 
+int64_t DOSBOX_GetTicksDone();
+void DOSBOX_SetTicksDone(const int64_t ticks_done);
+void DOSBOX_SetTicksScheduled(const int64_t ticks_scheduled);
+
+class Config;
+using ConfigPtr = std::unique_ptr<Config>;
+extern ConfigPtr control;
```
weirddan455, line 83: I moved this to control.h in https://github.com/dosbox-staging/dosbox-staging/commit/a53addf59fcbd4013a1d52169356d545cb068132 to avoid the forward declaration of the Config class.

Recommend removing this and  doing`#include "control.h"` wherever you need this.

src/dosbox.cpp
```diff
@@ -359,8 +367,8 @@ static void increase_ticks()
 			}
 		}
 
-		if (new_cycle_max < CPU_CYCLES_LOWER_LIMIT) {
-			new_cycle_max = CPU_CYCLES_LOWER_LIMIT;
+		if (new_cycle_max < auto_cpu_cycles_min) {
+			new_cycle_max = auto_cpu_cycles_min;
 		}
```
weirddan455: Is there a better name for these variables?  I would expect the max to always be bigger than the min.

src/cpu/cpu.cpp
```diff
@@ -2635,18 +2637,16 @@ void CPU_ShutDown([[maybe_unused]] Section* sec)
 #elif (C_DYNREC)
 	CPU_Core_Dynrec_Cache_Close();
 #endif
-	delete test;
```
weirddan455, line 2501: Why are we no longer destroying the class here?  I think you would reset the unique_pr here unless you want it get destroyed at some unspecified point when global objects get cleaned up.

src/cpu/cpu.cpp
```diff
@@ -2113,7 +2118,7 @@ bool CPU_CPUID(void) {
 	return true;
 }
 
-static Bits HLT_Decode(void)
+static Bits htl_decode(void)
```
weirddan455: Typo.  Should be hlt not htl.

src/cpu/cpu.cpp
```diff
@@ -2133,7 +2138,7 @@ void CPU_HLT(Bitu oldeip)
 	cpu.hlt.cs          = SegValue(cs);
 	cpu.hlt.eip         = reg_eip;
 	cpu.hlt.old_decoder = cpudecoder;
-	cpudecoder          = &HLT_Decode;
+	cpudecoder          = &htl_decode;
```
weirddan455, line 2257: Same typo here.

include/setup.h
```diff
@@ -129,7 +129,13 @@ class Value {
...
 	struct Changeable {
-		enum Value { Always, WhenIdle, OnlyAtStart, Deprecated };
+		enum Value {
+			Always,
+			WhenIdle,
+			OnlyAtStart,
+			Deprecated,
+			DeprecatedButAllowed
+		};
 	};
```
weirddan455, line 139: Why is deprecated in the same enum as changeable?  It seems to me like we're talking about two completely independent things here.  Is the setting changeable at runtime and is the setting deprecated.

src/dosbox.cpp
```diff
@@ -812,34 +812,92 @@ void DOSBOX_Init()
...
-	        "('auto' by default). Setting this value too high may result in sound drop-outs\n"
-	        "and lags.\n"
-	        "  auto:            Try to guess what a game needs. It usually works, but can\n"
-	        "                   fail with certain games.\n"
-	        "  fixed <number>:  Set a fixed number of cycles. This is what you usually\n"
-	        "                   need if 'auto' fails (e.g. 'fixed 4000').\n"
-	        "  max:             Allocate as much cycles as your computer is able to handle.");
+	        "The 'cycles' setting is deprecated but still accepted; please use the\n"
+	        "'cpu_cycles', 'cpu_cycles_protected' and 'cpu_throttle' settings instead as\n"
+	        "support will be removed in the future.");
```
weirddan455: This text does not appear in the config file after building this branch and deleting my old config.  Only the new cycles settings are shown.  That's probably your intent but no one is going to see this message.

src/cpu/cpu.cpp
```diff
@@ -131,6 +158,83 @@ void CPU_Core_Dynrec_Cache_Close();
...
+			CPU_CycleMax        = old_cycle_max;
+		} else {
+			set_modern_cycles_config(CpuMode::Real);
+		}
+
+		GFX_NotifyCyclesChanged();
+	}
+#if (C_DYNAMIC_X86) || (C_DYNREC)
+	if (auto_determine_mode.auto_core) {
+		cpudecoder = &CPU_Core_Normal_Run;
```
weirddan455, line 233: This always seemed odd to me (I know this is pre-existing behavior)  Why do we switch cores to normal in real mode?  Does dynrec have some issue with real mode specifically or is it more of "we don't need extra performance so let's use the simpler (possibly more accurate?) core"?  If the latter, would it not make more sense to switch to normal core depending on the number of cycles we are running?

Maybe this change is out of scope for this PR as it would probably require further testing but this behavior surprised me when I ran into it.  I was expecting auto to use the dynamic core all the time if it's available.

At the very least, it would be nice to document this.  Current help text on the config file is:

> #                 core: CPU core used in emulation ('auto' by default). 'auto' will switch to dynamic
> #                       if available and appropriate.

Maybe change that to "auto will switch to dynamic in protected mode if available" or something.

src/cpu/cpu.cpp
```diff
@@ -1619,44 +1723,54 @@ void CPU_SET_CRX(Bitu cr, Bitu value)
...
+					old_cycle_max       = CPU_CycleMax;
+
+					GFX_NotifyCyclesChanged();
 
-				GFX_NotifyCyclesChanged();
+					if (!displayed_max_cycles_warning) {
+						LOG_WARNING(
+								"CPU: Switched to max cycles for protected mode program. "
+								"Try setting a fixed cycles value if you're getting audio "
+								"drop-outs or the programs runs too fast.");
```
weirddan455, line 1746: Do we want to also display this warning when using modern cycles setting?

src/cpu/cpu.cpp
```diff
@@ -2242,39 +2371,115 @@ static void cpu_increase_cycles(bool pressed)
...
-			LOG_MSG("CPU: fixed %d cycles. If you need more than 20000, "
-			        "try core=dynamic in DOSBox's options.",
+		if (CPU_CycleMax > 20000) {
+			LOG_MSG("CPU: Fixed %d cycles If you need more than 20000, "
+			        "try setting 'core = dynamic' for increased performance.",
 			        CPU_CycleMax);
 		} else {
-			LOG_MSG("CPU: fixed %d cycles.", CPU_CycleMax);
+			LOG_MSG("CPU: Fixed %d cycles", CPU_CycleMax);
 		}
```
weirddan455: Do we want to also display these messages when using modern settings?

src/cpu/cpu.cpp
```diff
@@ -2639,14 +3051,52 @@ class Cpu final {
 	{
 		Section_prop* secprop = static_cast<Section_prop*>(sec);
 
+		{
+			PropMultiVal *p = secprop->GetMultiVal("cycles");
+			std::string type = p->GetSection()->Get_string("type");
+			std::string str;
```
weirddan455: Unused variable `str`?

src/cpu/cpu.cpp
```diff
@@ -3097,13 +3098,6 @@ class Cpu final {
 	{
 		Section_prop* secprop = static_cast<Section_prop*>(sec);
 
-		{
-			PropMultiVal *p = secprop->GetMultiVal("cycles");
-			std::string type = p->GetSection()->Get_string("type");
-			std::string str;
-			CommandLine cmd("", p->GetSection()->Get_string("parameters"));
-		}
-
```
weirddan455: Ok, I guess this entire block was removed in a later commit so you can ignore my unused variable comment.  Not entirely sure what you were doing here.

weirddan455:
I've gone through the entire thing now.  Mostly minor comments.  This is a good improvement.  Definitely prefer the 3 simple config options to the old cycles config that tried to do too much in a single setting and had weird syntax.

interloper98:
My vote would be to have limiting (or throttling, I guess?) enabled by default, especially for protected mode, because it's almost always beneficial. 

When I need to push cycles to 300k or 500k or whatever for high res FPS games, I usually use `limit` otherwise without limiting, I can easily get stutters in the more intense parts of the game.

My understanding is that this is actually how DOSBox currently operates in protected mode: `auto` or `max` already limits, and I think that's a good thing.

include/dosbox.h
johnnovak: Good tip, done.

src/cpu/cpu.cpp
johnnovak: Good pickup, and you're right, I forgot about this. Calling `.reset()` on the unique_ptr on destroy.

include/setup.h
johnnovak: Yes.

😄 

We'll address these in the config system refactor, just shoehorning in the user-facing functionality we need in the least invasive way for now.

src/cpu/cpu.cpp
```diff
@@ -2283,30 +2488,65 @@ static void cpu_decrease_cycles(bool pressed)
 		}
 
 		if (CPU_CyclePercUsed <= 70) {
-			LOG_MSG("CPU: max %d percent. If the game runs too fast, "
-			        "try a fixed cycles amount in DOSBox's options.",
+			LOG_MSG("CPU: max %d percent cycles. If the game runs too fast, "
+			        "try setting a fixed cycles value.",
 			        CPU_CyclePercUsed);
```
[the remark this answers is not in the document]
johnnovak: Yeah, nothing to do here, so moving on 😄

src/cpu/cpu.cpp
johnnovak: Just some trace logging thing I was doing, but it's harmless so I can't be bothered to properly remove it from the commits...

johnnovak:
> My vote would be to have limiting (or throttling, I guess?) enabled by default, especially for protected mode, because it's almost always beneficial.
> 
> When I need to push cycles to 300k or 500k or whatever for high res FPS games, I usually use `limit` otherwise without limiting, I can easily get stutters in the more intense parts of the game.
> 
> My understanding is that this is actually how DOSBox currently operates in protected mode: `auto` or `max` already limits, and I think that's a good thing.

I've been on the fence with this, whether to "throttle or not to throttle" by default. Now I think _not_ to throttle by default is the best option, taking everything into account.

My reasoning:

- At least a few demos crash with throttling enabled, even though my machine is fast enough to hit the target cycles value (e.g., `cycles = 120000` works but `cycles = max limit 120000` consistently crashes every single time).
- Throttling enabled by default instills the wrong usage patterns in people (i.e., thinking to set cycles to some very high setting, because throttling will care of it anyway; but this effectively simulates `cycles = max` and would result in sound stutters and lags and general misbehaviour if the effective cycle count is too high in speed-sensitive games).
- Finding the lowest fixed cycles setting per game that achieves optimal behaviour is the best possible way to set up games. We want to gently nudge people towards this ideal way of configuring games. Throttling enabled by default would muddy the waters, and we're back to "soft max" mode, effectively.
- Games where you need obscenely high cycles values like 300-500k are only a very small percentage of all games that use protected mode. Basically, you pretty much only need that for 3D games that use 640x480 or higher SVGA or VESA modes, so basically only a handful of very late 90s shooters and flight sims. Again, that's a modest portion of all protected mode games, and a minuscule portion of the whole DOS gaming catalogue spanning from 1981 to 2000, comprising of 7000k+ titles.
- Related to the previous point, our aim is not to give people the best way to play late 90s FPS games at max FPS out of the box. Our goal is to provide a _generally most useful_ setting for the _widest range_ of games and programs, while also instilling good usage patterns. Running Blood at 1280x1024 at max FPS with stock settings is simply not a consideration 😎

But, the good news is, my release notes draft goes into the changes in detail, so setting `cpu_throttle = yes` in the primary config is just a few keystrokes away for all those late 90s SVGA gamers 😎

src/dosbox.cpp
johnnovak: Yes, that's 100% intentional. The message will be seen in the logs if you set `cycles = <whatever>` or when you run `config -h cycles`, but it's hidden when writing the config file. I see this as the path forward for legacy settings we must still accept because of backward compatibility, but we effectively want to sweep them under the carpet and only reluctantly tell the user about their existence 😆 That's what the new `DeprecatedButAllowed` config setting type does.

image

src/gui/titlebar.cpp
```diff
@@ -463,12 +441,9 @@ void GFX_NotifyProgramName(const std::string& segment_name,
...
 }
 
-void GFX_NotifyCyclesChanged(const int32_t cycles)
+void GFX_NotifyCyclesChanged()
 {
-	if (cycles >= 0 && state.num_cycles != cycles) {
-		state.num_cycles = cycles;
-		GFX_RefreshTitle();
-	}
+	GFX_RefreshTitle();
```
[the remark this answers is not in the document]
johnnovak: Good point. I want to solve this the most generic way possible, so I'll always store the previous title bar string, and only update the title bar if the new string differs.

interloper98:
Thanks for the detailed explanation, @johnnovak ! Yes; I think throttling would let people fool themselves by cranking up very high cycles while dosbox silently reducing cycles to avoid maxing out their system. 

Totally agree that non-theottled will help people find the correct (and minimum) cycles for a given game.

src/cpu/cpu.cpp
johnnovak: > Does dynrec have some issue with real mode specifically or is it more of "we don't need extra performance so let's use the simpler (possibly more accurate?) core"? 

The dynarec core can sacrifice timing accuracy for speed. Self-modifying code can also fail on the dynamic core and run fine on the normal core.

> If the latter, would it not make more sense to switch to normal core depending on the number of cycles we are running?

Because of the above, I don't think we want to switch the core depending on the cycles value in use.

> At the very least, it would be nice to document this. Current help text on the config file is:

That's a great suggestion, and I started documenting the `cputype` and `core` settings, but that will be a separate follow-up PR. Let's not overload this one any further 😄

src/cpu/cpu.cpp
johnnovak: Yeah, good idea.

src/cpu/cpu.cpp
johnnovak: Yeah good idea.

include/dosbox.h
```diff
@@ -79,8 +79,6 @@ void DOSBOX_SetTicksDone(const int64_t ticks_done);
 void DOSBOX_SetTicksScheduled(const int64_t ticks_scheduled);
 
 class Config;
```
weirddan455, line 81: You don't need this forward `class Config;` declaration either.  It was on there to declare the pointer.

MasterO2:
> Thanks for the detailed explanation, @johnnovak ! Yes; I think throttling would let people fool themselves by cranking up very high cycles while dosbox silently reducing cycles to avoid maxing out their system.
> 
> Totally agree that non-throttled will help people find the correct (and minimum) cycles for a given game.

I can just picture someone doing the opposite with the CPU non-throttled and then putting in a Staging Github thread something like this:

`I set my CPU cycle count to 2000000. Why does Staging no longer run?`

src/gui/titlebar.cpp
```diff
@@ -324,10 +324,17 @@ static void maybe_add_recording_pause_mark(std::string& title_str)
...
+
+	auto new_title_str = state.title_no_tags;
+
+	maybe_add_muted_mark(new_title_str);
+	maybe_add_recording_pause_mark(new_title_str);
+
+	if (new_title_str != last_title_str) {
+		SDL_SetWindowTitle(sdl.window, new_title_str.c_str());
+		last_title_str = new_title_str;
+	}
```
weirddan455, line 337: This is probably fine.  I was thinking to wrap the entire `GFX_RefreshTitle()` function like it was before but now that I look at it more, the function looks safe even if it gets called a lot.  The animation code is just stopping/starting a timer when it needs to.

By "a lot", I was worried this would get called every tick if the CPU throttles and changes cycles but you said that is not the case so probably nothing to fear.

weirddan455 (approved):
This looks good to me.  I think you might still be missing some log messages when using the modern cycles settings.  You resolved my comment but I don't see the change.

Anyway, that's a minor thing and you fixed/answered all my other comments.

johnnovak:
> I can just picture someone doing the opposite with the CPU non-throttled and then putting in a Staging Github thread something like this:
> 
> `I set my CPU cycle count to 2000000. Why does Staging no longer run?`

Sure, but that's true for every backward compatibility breaking improvement we do to achieve more consistent and more reasonable defaults. Solution:

1. Reply with link to relevant release notes / manual section.
2. Apply "not a bug" label and close with "won't do". Done 😎

johnnovak:
> This looks good to me. I think you might still be missing some log messages when using the modern cycles settings. You resolved my comment but I don't see the change.
> 
> Anyway, that's a minor thing and you fixed/answered all my other comments.

Thanks for the detailed review @weirddan455, these were some very good comments there 🎖️ 

I haven't forgotten about those logging improvements; they're half implemented in my local branch, but I need to test them more thoroughly. I resolve comments as soon as I start working on them, but I also started working on the better `core` description in a separate branch, then called it a day.

Usually I ping the reviewer in a comment once I've pushed all my PR actions, but I can try to resolve only after that if that's confusing. I agree pushing 80% of the changes can be confusing 😄 

Who resolves comments and when is a neverending topic of debate at work too 😄

src/gui/titlebar.cpp
johnnovak: Yeah I thought it was safe without this too, but this precautionary diffing is just good practice.

I couldn't wrap it like before as the title bar update gets triggered by the cycles changed notification. Well, I could restrict the diffing to the human readable cycles text only and not call `GFX_RefreshTile()` if no change, but I like this generalised solution better to diff the whole title bar text.

weirddan455:
>  I set my CPU cycle count to 2000000. Why does Staging no longer run?

>  Sure, but that's true for every backward compatibility breaking improvement we do to achieve more consistent and more reasonable defaults.

That's not even really deviating from current behavior.  If you do `cycles = 2000000` on v0.81.1, throttling would be off.  You have to use some weird syntax like `cycles = auto 3000 max limit 120000` to get throttling (and I'm still not sure what exactly that does, I just copied it from the PR description).  The syntax was confusing and I assume most people weren't using it.

I was torn on whether throttling should be off or on by default as well but your reasoning makes sense, plus the fact that it was defacto off by default anyway.

>  I haven't forgotten about those logging improvements; they're half implemented in my local branch, but I need to test them more thoroughly. I resolve comments as soon as I start working on them, but I also started working on the better core description in a separate branch, then called it a day.

No worries.  I wasn't sure if you changed it and I just missed it but it's a minor thing either way.

include/dosbox.h
johnnovak: Cheers, removed.

johnnovak:
Alright, I'm done with this good thing, and I'm glad it's over 😅 Thanks for the review again @weirddan455 and for the useful comments! Hitting the big green button 😎 Here goes nothing! 🤞🏻

johnnovak:
Another interesting anecdote why `cycles = max` is evil:

https://github.com/dosbox-staging/dosbox-staging/issues/366#issuecomment-903362197

PR #3687, opened 2024-05-21
Implement File Locking

### LINKED ISSUE #1489: File locking support for better game compatibility (SHARE.EXE)

_No response_

_No response_

macOS 11.6.2/x86-64

When installing some of the more demanding Windows games this pops up:

Screenshot 2022-09-07 at 17 28 27

This is due to lack of file locking support in DOSBox Staging which affects how both SHARE and VSHARE.386 do file locking.

At least the following games are known to suffer from SHARE.EXE not loaded issue:

- Comix Zone (1995)
- Garfield Caught in the Act (1996)
- Microsoft Flight Simulator 5 - ATC Workshop (1995)
- Monopoly (1995)
- Ultimate Yahtzee (1996)

I would have previously opposed adding file locking support as a too complex to implement. We've since moved to C++17 which has filesystem library that makes this feature easier to implement and more robust to maintain. The best way to implement is by accessing host OS directory mount directly.

The file locking logic can be pretty easily summed up by a snippet of code from DOSBox-X that has this implemented:

dos_files.cpp:890:

```
	} else {
		//Test if file exists, but opened in read-write mode (and writeprotected)
		if((((flags&3) != OPEN_READ) || (enable_share_exe && !strncmp(Drives[drive]->GetInfo(),"local directory ",16))) && Drives[drive]->FileExists(fullname))
			DOS_SetError(DOSERR_ACCESS_DENIED);
		else {
			if(!PathExists(name)) DOS_SetError(DOSERR_PATH_NOT_FOUND); 
			else DOS_SetError(DOSERR_FILE_NOT_FOUND);
		}
		return false;
	}
```

Adding this feature with the upcoming DOS 38h support will make this a general purpose emulator for DOS and Win16 environments. This makes it easier to use for a basic user because running multiple emulators for the same environment can be avoided. This will also attract more users.

There is a method of using fake SHARE but such solutions may lead to data corruption.

The question of is this in the scope of the project arises in this case. My evaluation is that it is based on the Vision page:

https://dosbox-staging.github.io/about/

Screenshot 2022-01-02 at 15 11 23

By utilising C++17 filesystem in implementing i think it passes all the conditions to be a viable, implementable feature. Can be implemented after 0.79 release cycle if there's no resources.

Sample code for C++17 filesystem file locking:

https://stackoverflow.com/a/56348495

```
#include <sys/stat.h>

bool IsLocked(std::string& thePath)
{
    bool isLocked = false;

    struct stat buf;
    int retval = stat(thePath.c_str(), &buf);     // NOTE: retval could be an error, like ENOENT (file doesn't exist)
    if (0 != (buf.st_flags & UF_IMMUTABLE)) {
        isLocked = true;
         //is immutable
    }

    return isLocked;
}
```

- [X] Yes, I agree.

weirddan455:
# Description

I've implemented locking the entire file with sharing bits in `DOS_OpenFile` and `DOS_CreateFile`.

There is another (completely independent from sharing bits) method that lock regions of a file.  This is done through INT 21 AH = 0x5c.  This requires performing a check on `DOS_ReadFile` and `DOS_WriteFile` to check if the region a process is attempting to read/write to has been locked by another process.

Many of the games I've tested require Win32s (who's installer does a `SHARE.EXE` check).  This check failed for me on both DOSBox-X and using `FAKESHR.COM` so we are doing better than them now :smile: 

## Related issues

Fixes #1489

# Manual testing

- 3-D Ultra Pinball
- Comix Zone
- Microsoft Flight Simulator 5 - ATC Workshop
- AFL Finals Fever
- Mind Grid - Install errors but probably not related to this PR (also producible on main see #3703)
- Microsoft Word 6.1 for Windows

# Checklist

I have:

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -345,7 +345,7 @@ public:
 	DOS_Drive();
 	virtual ~DOS_Drive() = default;
 
-	virtual bool FileOpen(DOS_File** file, char* name, uint32_t flags) = 0;
+	virtual bool FileOpen(DOS_File** file, char* name, uint8_t flags) = 0;
 	virtual bool FileCreate(DOS_File** file, char* name,
 	                        FatAttributeFlags attributes) = 0;
 	virtual bool FileUnlink(char* _name)=0;
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -129,7 +129,7 @@ public:
 
 	void SetDrive(uint8_t drv) { hdrive=drv;}
 	uint8_t GetDrive(void) { return hdrive;}
-	uint32_t flags   = 0;
+	uint8_t flags    = 0;
 	uint16_t time    = 0;
 	uint16_t date    = 0;
 	FatAttributeFlags attr = {};
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -98,6 +98,11 @@ struct DosFilename {
 	std::string ext  = {};
 };
 
+struct FileRegionLock {
+	uint32_t pos = 0;
+	uint32_t len = 0;
+};
+
 class DOS_File {
 public:
 	DOS_File() = default;
@@ -137,6 +142,7 @@ public:
 	bool open        = false;
 	std::string name = {};
 	bool newtime     = false;
+	std::vector<FileRegionLock> region_locks = {};
 	/* Some Device Specific Stuff */
 private:
 	uint8_t hdrive = 0xff;
--- a/src/dos/dos_files.cpp
+++ b/src/dos/dos_files.cpp
@@ -180,6 +180,35 @@ static bool file_is_locked(const char *file_name, const uint8_t drive, const uin
 	return false;
 }
 
+static bool regions_overlap(const uint32_t pos1, const uint32_t len1, const uint32_t pos2, const uint32_t len2)
+{
+	return !((pos1 >= pos2 + len2) || (pos1 + len1 <= pos2));
+}
+
+static bool region_is_locked(const int file_handle, const uint32_t pos, const uint32_t len)
+{
+	for (int i = 0; i < DOS_FILES; ++i) {
+		// Ignore locks held by the current file handle,
+		// Need to check other handles pointing to the same file.
+		if (i != file_handle && Files[i]) {
+			const auto drive_match =
+				(Files[i]->GetDrive() == Files[file_handle]->GetDrive());
+
+			const auto filename_match =
+				(Files[i]->IsName(Files[file_handle]->GetName()));
+
+			if (drive_match && filename_match) {
+				for (const auto &lock : Files[i]->region_locks) {
+					if (regions_overlap(pos, len, lock.pos, lock.len)) {
+						return true;
+					}
+				}
+			}
+		}
+	}
+	return false;
+}
+
 uint8_t DOS_GetDefaultDrive(void) {
 //	return DOS_SDA(DOS_SDA_SEG,DOS_SDA_OFS).GetDrive();
 	uint8_t d = DOS_SDA(DOS_SDA_SEG,DOS_SDA_OFS).GetDrive();
@@ -598,6 +627,16 @@ bool DOS_ReadFile(uint16_t entry,uint8_t * data,uint16_t * amount,bool fcb) {
 	}
 */
 	uint16_t toread=*amount;
+
+	// Get current position
+	uint32_t pos = 0;
+	Files[handle]->Seek(&pos, SEEK_CUR);
+
+	if (region_is_locked(handle, pos, toread)) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
+		return false;
+	}
+
 	bool ret=Files[handle]->Read(data,&toread);
 	*amount=toread;
 	return ret;
@@ -620,6 +659,16 @@ bool DOS_WriteFile(uint16_t entry,uint8_t * data,uint16_t * amount,bool fcb) {
 	}
 */
 	uint16_t towrite=*amount;
+
+	// Get current position
+	uint32_t pos = 0;
+	Files[handle]->Seek(&pos, SEEK_CUR);
+
+	if (region_is_locked(handle, pos, towrite)) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
+		return false;
+	}
+
 	bool ret=Files[handle]->Write(data,&towrite);
 	*amount=towrite;
 	return ret;
@@ -1720,4 +1769,47 @@ void DOS_SetupFiles()
 
 	Drives.at(z_drive_index) = DriveManager::RegisterFilesystemImage(
 	        z_drive_index, std::make_unique<Virtual_Drive>());
+}
+
+bool DOS_LockFile(const uint16_t entry, const uint32_t pos, const uint32_t len)
+{
+	const auto handle = RealHandle(entry);
+	if (handle >= DOS_FILES) {
+		DOS_SetError(DOSERR_INVALID_HANDLE);
+		return false;
+	}
+	if (!Files[handle]) {
+		DOS_SetError(DOSERR_INVALID_HANDLE);
+		return false;
+	}
+	if (region_is_locked(handle, pos, len)) {
+		DOS_SetError(DOSERR_LOCK_VIOLATION);
+		return false;
+	}
+	FileRegionLock lock = {};
+	lock.pos = pos;
+	lock.len = len;
+	Files[handle]->region_locks.push_back(lock);
+	return true;
+}
+
+bool DOS_UnlockFile(const uint16_t entry, const uint32_t pos, const uint32_t len)
+{
+	const auto handle = RealHandle(entry);
+	if (handle >= DOS_FILES) {
+		DOS_SetError(DOSERR_INVALID_HANDLE);
+		return false;
+	}
+	if (!Files[handle]) {
+		DOS_SetError(DOSERR_INVALID_HANDLE);
+		return false;
+	}
+	for (auto it = Files[handle]->region_locks.begin(); it != Files[handle]->region_locks.end(); ++it) {
+		if (it->pos == pos && it->len == len) {
+			Files[handle]->region_locks.erase(it);
+			return true;
+		}
+	}
+	DOS_SetError(DOSERR_LOCK_VIOLATION);
+	return false;
 }
--- a/src/dos/dos_files.cpp
+++ b/src/dos/dos_files.cpp
@@ -50,6 +50,136 @@ DOS_File* Files[DOS_FILES] = {};
 // the drive manager class.
 std::array<DOS_Drive*, DOS_DRIVES> Drives = {};
 
+enum class FileSharingMode
+{
+	Compatibility,
+	DenyReadWrite,
+	DenyWrite,
+	DenyRead,
+	DenyNone
+};
+
+// https://stanislavs.org/helppc/int_21-3d.html
+union FileFlagBits
+{
+	uint8_t data = 0;
+	bit_view<0, 3> access_mode;
+	bit_view<4, 3> sharing_mode;
+};
+
+struct FileAccessMode
+{
+	bool read = false;
+	bool write = false;
+};
+
+struct FileOpenFlags
+{
+	FileSharingMode sharing = FileSharingMode::Compatibility;
+	FileAccessMode access = {};
+};
+
+static FileOpenFlags parse_file_flags(const uint8_t flags)
+{
+	FileOpenFlags ret = {};
+
+	const FileFlagBits flag_bits = {flags};
+
+	switch (flag_bits.access_mode) {
+		case OPEN_READ:
+		case OPEN_READ_NO_MOD:
+			ret.access.read = true;
+			ret.access.write = false;
+			break;
+		case OPEN_WRITE:
+			ret.access.read = false;
+			ret.access.write = true;
+			break;
+		case OPEN_READWRITE:
+			ret.access.read = true;
+			ret.access.write = true;
+			break;
+		default:
+			// Assume read/write access for the purpose of file locking checks
+			// This should throw an error when the file actually tries to open
+			ret.access.read = true;
+			ret.access.write = true;
+	}
+
+	switch (flag_bits.sharing_mode) {
+		case 0b000:
+			ret.sharing = FileSharingMode::Compatibility;
+			break;
+		case 0b001:
+			ret.sharing = FileSharingMode::DenyReadWrite;
+			break;
+		case 0b010:
+			ret.sharing = FileSharingMode::DenyWrite;
+			break;
+		case 0b011:
+			ret.sharing = FileSharingMode::DenyRead;
+			break;
+		case 0b100:
+			ret.sharing = FileSharingMode::DenyNone;
+			break;
+		default:
+			// If we somehow get something invalid, assume compatibiliy mode.
+			ret.sharing = FileSharingMode::Compatibility;
+	}
+
+	return ret;
+}
+
+static bool single_directional_sharing_check(const FileOpenFlags& new_file, const FileOpenFlags& existing_file)
+{
+	switch (new_file.sharing) {
+		case FileSharingMode::Compatibility:
+			return existing_file.sharing == FileSharingMode::Compatibility;
+		case FileSharingMode::DenyReadWrite:
+			return false;
+		case FileSharingMode::DenyWrite:
+			if (existing_file.sharing == FileSharingMode::Compatibility) {
+				return false;
+			} else {
+				return !existing_file.access.write;
+			}
+		case FileSharingMode::DenyRead:
+			if (existing_file.sharing == FileSharingMode::Compatibility) {
+				return false;
+			} else {
+				return !existing_file.access.read;
+			}
+		case FileSharingMode::DenyNone:
+			return existing_file.sharing != FileSharingMode::Compatibility;
+	}
+	assertm(false, "Invalid enum value");
+	return false;
+}
+
+static bool file_modes_are_compatible(const FileOpenFlags& new_file, const FileOpenFlags& existing_file)
+{
+	// Needs a two-way check
+	// Run this function twice with arguments reversed the second time
+	return single_directional_sharing_check(new_file, existing_file)
+		&& single_directional_sharing_check(existing_file, new_file);
+}
+
+static bool file_is_locked(const char *file_name, const uint8_t drive, const uint8_t flags)
+{
+	const FileOpenFlags new_file = parse_file_flags(flags);
+
+	for (int i = 0; i < DOS_FILES; ++i) {
+		if (Files[i] && Files[i]->GetDrive() == drive && Files[i]->IsName(file_name)) {
+			const FileOpenFlags existing_file = parse_file_flags(Files[i]->flags);
+			if (!file_modes_are_compatible(new_file, existing_file)) {
+				return true;
+			}
+		}
+	}
+
+	return false;
+}
+
 uint8_t DOS_GetDefaultDrive(void) {
 //	return DOS_SDA(DOS_SDA_SEG,DOS_SDA_OFS).GetDrive();
 	uint8_t d = DOS_SDA(DOS_SDA_SEG,DOS_SDA_OFS).GetDrive();
@@ -578,6 +708,11 @@ bool DOS_CreateFile(const char* name, FatAttributeFlags attributes,
 		return false;
 	}
 
+	if (file_is_locked(fullname, drive, OPEN_READWRITE)) {
+		DOS_SetError(DOSERR_ACCESS_DENIED);
+		return false;
+	}
+
 	/* We have a position in the main table now find one in the psp table */
 	DOS_PSP psp(dos.psp());
 	*entry = fcb?handle:psp.FindFreeFileEntry();
@@ -652,6 +787,10 @@ bool DOS_OpenFile(const char* name, uint8_t flags, uint16_t* entry, bool fcb)
 	if (device) {
 		Files[handle]=new DOS_Device(*Devices[devnum]);
 	} else {
+		if (file_is_locked(fullname, drive, flags)) {
+			DOS_SetError(DOSERR_ACCESS_DENIED);
+			return false;
+		}
 		const auto old_errorcode = dos.errorcode;
 		dos.errorcode = 0;
 		exists = Drives.at(drive)->FileOpen(&Files[handle], fullname, flags);
--- a/src/dos/dos_misc.cpp
+++ b/src/dos/dos_misc.cpp
@@ -63,6 +63,10 @@ static Bitu INT2A_Handler(void) {
 
 static bool DOS_MultiplexFunctions(void) {
 	switch (reg_ax) {
+	case 0x1000:
+		// Report that SHARE.EXE is installed
+		reg_al = 0xff;
+		return true;
 	case 0x1216:	/* GET ADDRESS OF SYSTEM FILE TABLE ENTRY */
 		// reg_bx is a system file table entry, should coincide with
 		// the file handle so just use that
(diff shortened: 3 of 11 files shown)


src/dos/dos_files.cpp
```diff
@@ -603,6 +603,129 @@ bool DOS_CreateFile(const char* name, FatAttributeFlags attributes,
...
+			break;
+		// Assume read/write access for the purpose of file locking checks
+		// This should throw an error when the file actually tries to open
+		default:
+			ret.access.read = true;
+			ret.access.write = true;
+	}
+
+	// Extract bits 4-6
+	const uint8_t sharing_bits = (flags & 0b1110000) >> 4;
```
johnnovak, line 656: This is a perfect candidate for `bit_view`.

src/dos/dos_files.cpp
```diff
@@ -603,6 +603,135 @@ bool DOS_CreateFile(const char* name, FatAttributeFlags attributes,
...
+	DenyWrite,
+	DenyRead,
+	DenyNone
+};
+
+union FileFlagBits
+{
+	uint8_t data = 0;
+	bit_view<0, 3> access_mode;
+	bit_view<4, 3> sharing_mode;
```
johnnovak, line 619: Well, this still leaves you with bit patterns when you interpret these values 😄 

What I meant:

- Put a link to this page above the union: https://stanislavs.org/helppc/int_21-3d.html
- Add bit-level mappings, naming each bit descriptively, just like I started doing in the VGA header files.

That would make the structure self-describing instead of relying on comments and/or magic bit pattern constants (or hex, which can be even worse).

weirddan455: There's not a meaning for a single bit though.  These are both 3 bit patterns, as you can see from that link you just sent.  There's not like a read bit and a write bit.

johnnovak: Mea culpa. I did not read it in detail and I'm still half asleep 😅 Ok then, just putting the link there would suffice 😄

FeralChild64 (commented):
It would be nice if the code could log (`LOG_WARNING`) unsupported sharing modes if the application tries to use them.

src/dos/dos_files.cpp
weirddan455: Added the link

weirddan455:
Fixed a missing region lock check in `DOS_LockFile()`.  You can't lock a region if some other process has already locked it.

Also, not sure if we are handling child process inheritance of file handles correctly.  Documentation for the full file lock (as set by sharing bits in `DOS_OpenFile()` says that child processes should inherit this sharing if the inherit bit is set.

file-lock

But documentation for region lock says it should **not** be inherited by child processes.  File handle duplication should be handled correctly.  That function simply inserts a new reference to an existing `DOS_File` in the caller's PSP segment.  Child processes I am unclear on.  Need to test and see if something breaks.

region-lock

weirddan455:
I've done some testing and updating the top comment with games tested.  Some of the games on #1489 I could not find proper images for but I tested what I could find.

It looks like we are respecting the inherit bit, at least for full file locking.  Partial file locking I am still unclear on if the locking process creates a child process but I cannot find a fail case in a game so I think we can cross that bridge when we come to it.

This is done and ready for review.

src/dos/dos_files.cpp
```diff
@@ -180,6 +180,28 @@ static bool file_is_locked(const char *file_name, const uint8_t drive, const uin
...
+		// Ignore locks held by the current file handle,
+		// Need to check other handles pointing to the same file.
+		if (i != file_handle && Files[i] && Files[i]->GetDrive() == Files[file_handle]->GetDrive()
+				&& Files[i]->IsName(Files[file_handle]->GetName())) {
+			for (const auto &lock : Files[i]->region_locks) {
+				if (regions_overlap(pos, len, lock.pos, lock.len)) {
+					return true;
+				}
+			}
+		}
```
johnnovak, line 200: Readability improvement suggestion:

```cpp
    if (i != file_handle && Files[i]) {
        const auto drive_match =
            (Files[i]->GetDrive() == Files[file_handle]->GetDrive());

        const auto filename_match =
            (Files[i]->IsName(Files[file_handle]->GetName()));

        if (drive_match && filename_match) {
            for (const auto &lock : Files[i]->region_locks) {
                if (regions_overlap(pos, len, lock.pos, lock.len)) {
                    return true;
                }
            }
        }
    }

```

src/dos/dos_files.cpp
```diff
@@ -1721,3 +1763,46 @@ void DOS_SetupFiles()
 	Drives.at(z_drive_index) = DriveManager::RegisterFilesystemImage(
 	        z_drive_index, std::make_unique<Virtual_Drive>());
 }
+
+bool DOS_LockFile(uint16_t entry, uint32_t pos, uint32_t len)
```
johnnovak, line 1767: Could add const

src/dos/dos_files.cpp
```diff
@@ -1721,3 +1763,46 @@ void DOS_SetupFiles()
...
+	FileRegionLock lock = {};
+	lock.pos = pos;
+	lock.len = len;
+	Files[handle]->region_locks.push_back(lock);
+	return true;
+}
+
+bool DOS_UnlockFile(uint16_t entry, uint32_t pos, uint32_t len)
+{
+	const uint32_t handle = RealHandle(entry);
```
johnnovak, line 1791: auto?

src/dos/dos_files.cpp
```diff
@@ -50,6 +50,136 @@ DOS_File* Files[DOS_FILES] = {};
...
+			ret.access.read = false;
+			ret.access.write = true;
+			break;
+		case OPEN_READWRITE:
+			ret.access.read = true;
+			ret.access.write = true;
+			break;
+		// Assume read/write access for the purpose of file locking checks
+		// This should throw an error when the file actually tries to open
+		default:
```
johnnovak, line 104: Comment should go inside the `default` branch & fix indentation.

src/dos/dos_files.cpp
```diff
@@ -50,6 +50,136 @@ DOS_File* Files[DOS_FILES] = {};
...
+			ret.sharing = FileSharingMode::DenyWrite;
+			break;
+		case 0b011:
+			ret.sharing = FileSharingMode::DenyRead;
+			break;
+		case 0b100:
+			ret.sharing = FileSharingMode::DenyNone;
+			break;
+		// If we somehow get something invalid, assume compatibiliy mode.
+		default:
```
johnnovak, line 126: Comment should go inside the default branch & fix indentation.

src/dos/dos_files.cpp
```diff
@@ -50,6 +50,136 @@ DOS_File* Files[DOS_FILES] = {};
...
+		case FileSharingMode::DenyRead:
+			if (existing_file.sharing == FileSharingMode::Compatibility) {
+				return false;
+			} else {
+				return !existing_file.access.read;
+			}
+		case FileSharingMode::DenyNone:
+			return existing_file.sharing != FileSharingMode::Compatibility;
+	}
+	// GCC 12 throws a compile error if we don't have a return statement here
```
johnnovak, line 155: Remove comment; we do this everywhere already, we don't need to comment on it (plus I believe PVS-Studio would flag this anyway).

src/dos/dos_files.cpp
```diff
@@ -50,6 +50,136 @@ DOS_File* Files[DOS_FILES] = {};
...
+	assertm(false, "Invalid enum value");
+	return false;
+}
+
+static bool file_modes_are_compatible(const FileOpenFlags& new_file, const FileOpenFlags& existing_file)
+{
+	// Needs a two-way check
+	// Run this function twice with arguments reversed the second time
+	return single_directional_sharing_check(new_file, existing_file) && single_directional_sharing_check(existing_file, new_file);
```
johnnovak, line 164: Formatting (line too long)

include/dos_system.h
```diff
@@ -345,7 +345,7 @@ class DOS_Drive {
 	DOS_Drive();
 	virtual ~DOS_Drive() = default;
 
-	virtual bool FileOpen(DOS_File** file, char* name, uint32_t flags) = 0;
+	virtual bool FileOpen(DOS_File** file, char* name, uint8_t flags) = 0;
```
johnnovak, line 354: const while you're at it

src/dos/drive_overlay.cpp
```diff
@@ -216,7 +216,7 @@ class OverlayFile final : public localFile {
 	}
 
 	bool Write(uint8_t * data,uint16_t * size) override {
-		uint32_t f = flags&0xf;
+		uint8_t f = flags&0xf;
```
johnnovak, line 219: minor: reformat

src/dos/dos_misc.cpp
```diff
@@ -63,6 +63,10 @@ static Bitu INT2A_Handler(void) {
 
 static bool DOS_MultiplexFunctions(void) {
 	switch (reg_ax) {
+	case 0x1000:
```
johnnovak, line 66: Maybe document what `0x1000` stands for?

johnnovak (approved):
Still looking good @weirddan455, just a few minor style-related comments.

Grounded0:
Thanks for finally implementing this.

Grounded0:
You can expand your testing software with something like Microsoft Word 6.1 for Windows:

https://winworldpc.com/product/microsoft-word/6x

johnnovak:
> You can expand your tested software with something like Microsoft Word 6.1 for Windows:
> 
> https://winworldpc.com/product/microsoft-word/6x

Does it need some locking specific functionality, do you know @Grounded0 ?

Grounded0:
@johnnovak 

Yes.

weirddan455:
@Grounded0 did you get that error on this branch?  If not, I can test it real quick.

Grounded0:
No this is 0.81.1.

weirddan455:
Well, this isn't a good sign.  Word's setup.exe crashes with what looks like a divide by zero.  I think this is unrelated to this PR as it's crashing in VGA code.

```
Loguru caught a signal: SIGFPE
Stack trace:
24      0x56553e48cf81 _start + 33
23      0x7f0524846305 __libc_start_main + 133
22      0x7f052484624a /lib/x86_64-linux-gnu/libc.so.6(+0x2724a) [0x7f052484624a]
21      0x56553e48d072 main + 34
20      0x56553e86077f sdl_main(int, char**) + 7039
19      0x56553e4c1996 Config::StartUp() + 22
18      0x56553eadae4e SHELL_Init() + 3870
17      0x56553ead8d6b DOS_Shell::Run() + 1819
16      0x56553ead7b5c DOS_Shell::ParseLine(char*) + 2124
15      0x56553eae813d DOS_Shell::DoCommand(char*) + 445
14      0x56553eb01433 DOS_Shell::ExecuteProgram(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) + 1267
13      0x56553eb025ab build/debug/dosbox(+0xc7f5ab) [0x56553eb025ab]
12      0x56553e5be1c4 CALLBACK_RunRealInt(unsigned char) + 84
11      0x56553e48d913 DOSBOX_RunMachine() + 19
10      0x56553e48d863 build/debug/dosbox(+0x60a863) [0x56553e48d863]
9       0x56553ea5fd6a build/debug/dosbox(+0xbdcd6a) [0x56553ea5fd6a]
8       0x56553ea6a275 INT10_SetVideoMode(unsigned short) + 2437
7       0x56553ea6d8c5 build/debug/dosbox(+0xbea8c5) [0x56553ea6d8c5]
6       0x56553e91d5e1 IO_WriteB(unsigned short, unsigned char) + 401
5       0x56553e48d913 DOSBOX_RunMachine() + 19
4       0x56553e48d7fd build/debug/dosbox(+0x60a7fd) [0x56553e48d7fd]
3       0x56553e983338 PIC_RunQueue() + 232
2       0x56553e9d5f13 VGA_SetupDrawing(unsigned int) + 83
1       0x56553e9d3464 setup_drawing() + 8516
0       0x7f052485b050 /lib/x86_64-linux-gnu/libc.so.6(+0x3c050) [0x7f052485b050]
2024-05-25 21:34:37.404 | Signal: SIGFPE
Floating point exception (core dumped)
```

weirddan455:
Except that's not constantly reproducible.  Now I get this error.  I might have to revert that SFT commit I made in an earlier PR.

image0008

Grounded0:
I'll test it myself on a test build this Sunday.

Can you provide me a download link to macOS artefact for this branch?

johnnovak:
@weirddan455 @Grounded0 How about we use a "standardised" Win3.x base install for all our testing? Our wiki has a lot of helpful hints on setting up Win3.x, but it's a bit sprawling. I'm thinking of doing a base Win3.x install myself, and create a more streamlined "recommended Win3.x setup" wiki page.

Then probably I'll create a few "snapshots" ZIP files:

- Base install only
- S3 and SB16 drivers only
- Additional updates (WinG, etc.)

How does that sound? Ultimately, we need to support a _single_ S3 driver, and a _single_ SB16 driver and get them work 100% reliably, then all games and apps should just work in theory as they only talk to the hardware through the Windows APIs.

Of course, we should pick non-buggy drivers for this 😄

weirddan455:
Oh, I downloaded Word for DOS.  So the SFT commit I made has a regression where it breaks DOS based programs from working inside Windows 3.1 (displaying the general protection fault error box inside Windows).

The VGA crash is a separate issue (reproducible on `main`) that is not consistently reproducible but happens maybe 1 in 4 times you try to launch a DOS program from inside Windows 3.1.

I'll make/update tickets for those.

Grounded0:
Microsoft Word 6.1 for Windows (3.5-1.44mb):

https://winworldpc.com/download/15c2b403-c2b8-c3a9-c5b8-11c3a5c28f13

johnnovak:
> happens maybe 1 in 4 times you try to launch a DOS program from inside Windows 3.1.

In general, launching DOS programs in Win3.x is the last of our concerns. "Nice to the have", I guess, but should be right at the bottom of our priority list. Just use DOS 🤷🏻

weirddan455:
>  In general, launching DOS programs in Win3.x is the last of our concerns. "Nice to the have", I guess, but should be right at the bottom of our priority list. Just use DOS 🤷🏻

True.  Everything continues to work in DOS.  It also breaks Window's MS-DOS prompt, which is the easiest way to reproduce the regression.

In any case, Word for Windows does work on this branch

image0009

MasterO2:
I thought Windows support was out of scope for Staging?

Grounded0:
Official position seems to be DOS and Windows 3.x (including Win16 and Win32s).

weirddan455:
>  I thought Windows support was out of scope for Staging? At least that's what Staging's web site lists as a non-goal.

It's Windows 9x specifically that's listed as a non-goal.  We're supporting Windows 3.1 stuff.

Grounded0:
File locking is also required by some DOS application software so this also benefits DOS.

src/dos/dos_misc.cpp
weirddan455: I put a comment directly under that line.  This is an interrupt function and all it does is basically return "true" if `SHARE.EXE` is installed.  I don't know that it even has a documented name.  This is the way all interrupt functions look in our codebase.

include/dos_system.h
weirddan455: Some of the `FileOpen` functions modify the arguments.

src/dos/dos_files.cpp
weirddan455: Yeah that looks nicer.  Changed.

src/dos/dos_files.cpp
weirddan455: const'd

src/dos/dos_files.cpp
weirddan455: auto'd

src/dos/dos_files.cpp
weirddan455: Fixed

src/dos/dos_files.cpp
weirddan455: Fixed

src/dos/dos_files.cpp
weirddan455: Removed

src/dos/dos_files.cpp
weirddan455: Formatted

src/dos/drive_overlay.cpp
weirddan455: Formatted

weirddan455:
>  File locking is also required by some DOS application software so this also benefits DOS.

I haven't found any yet but `SHARE.EXE` has been a part of MS-DOS itself since at least version 4.0 so this is definitely in scope for our DOS implementation.

The big distinction between Windows 3.1 and Windows 9x is that Windows 3.1 installs and runs on top of our internal DOS implementation whereas Windows 9x wants to install its own version of DOS on top of bare metal (or bare hardware emulation).

This means things like the `mount` command won't work and maybe even `localDrive` won't work.  I haven't tried it myself but I think you would have to create a FAT image of the Windows 9x install and then use the `boot`` command to boot into it (as far as Staging is concerned `boot` is meant for bootable floppy images of games).

It's not so user friendly to do all that and would require a lot of work to make it so (something like DOSBox-X's GUI).

weirddan455:
@johnnovak A Windows install would be helpful.  I was running into problems with the SB16 drivers myself.

johnnovak:
> I thought Windows support was out of scope for Staging? At least that's what Staging's web site lists as a non-goal.

Read the about page carefully, man 😄 Everythig you need to know is there 😄 
https://www.dosbox-staging.org/about/

Grounded0:
@sduensin 

What was the name of the DOS software that required file locking support (SHARE)?

johnnovak:
> True. Everything continues to work in DOS. It also breaks Window's MS-DOS prompt, which is the easiest way to reproduce the regression.

Yeah, and the guiding principle to everything we do in DOSBox should be this: does this allow to run the DOS and Win 3.x catalogue in an authentic way? DOS prompt in Win 3.x is an extreme niche; making it work brings no value in light of this goal.

Similarly, it's completely beside the point to get _all_ the possible driver working in Win 3.x for all the hardware we support. We just need a _single_ supported driver combo to run all the games optimally, job done. This allows us to spend our energies in a smart way on what truly matters: get the existing DOS/Win3.x catalogue from 1981-2000 working optimally. Running the same game with 10123 Win3.x driver combos that result in the exact same end-user experience is a non-goal and a waste of time and energy.

Just talking in genericsl I'm not saying we're doing this type of busy work now 😄 

> The big distinction between Windows 3.1 and Windows 9x is that Windows 3.1 installs and runs on top of our internal DOS implementation whereas Windows 9x wants to install its own version of DOS on top of bare metal (or bare hardware emulation).

Win 95 and 98 is a huge rabbit hole, and we have enough to in pure DOS/Win3.x land. Hence, we _really_ don't wanna go there, IMO. Keep the scope realistic and all.

Win 9.x kinda sorted "worked" (works) in all DOSBox version by fluke and random coincidence, but to make it work "properly", there's lots of additional work to do. Imagine dealing with all Win9.x game compatibility tickets then... Just no, we don't even wanna go there IMO. We are quite under-resourced just for the DOS / Win3.x stuff, we don't wanna take on something absolutely huge as Win 9.x (to do it properly, not just half-assedly).

Grounded0:
SHARE.EXE has been present in MS-DOS at least since MS-DOS 3.1 (search for SHARE.EXE):

http://www.ctyme.com/intr/int-21.htm

weirddan455:
>  We are quite under-resourced just for the DOS / Win3.x stuff, we don't wanna take on something absolutely huge as Win 9.x (to do it properly, not just half-assedly).

I agree 100%.  I'm not even sure DOSBox is the right approach at all for Win9x.  Maybe DOSBox-X does a decent job, haven't tried it but all the DOS stuff we implement becomes useless and then you're also running an entirely different class of hardware.  A lot of games need Pentium 2/3 class CPUs and much stronger GPUs than Voodoo.

johnnovak:
> I agree 100%. I'm not even sure DOSBox is the right approach at all for Win9x. Maybe DOSBox-X does a decent job, haven't tried it but all the DOS stuff we implement becomes useless and then you're also running an entirely different class of hardware. A lot of games need Pentium 2/3 class CPUs and much stronger GPUs than Voodoo.

Absolutely, even the emulated hardware requirements are significantly different, that's an excellent point. There is overlap, but I'd say Win 9x emulation should be a distinct and separately project from DOSBox; it's really its own thing. Huge respect to DOSBox-X for trying to pull it off, and all the best wishes in the world & good luck, but this is not for us.

> A Windows install would be helpful. I was running into problems with the SB16 drivers myself.

I'll do that over the coming days, ideally sometime next week. The planned outcome is a new succinct wiki page with installation steps, and a couple of ZIP archives on archive.org if someone just wants to quickly grab it and not go through the steps.

weirddan455:
>  I'll do that over the coming days, ideally sometime next week.

Sounds good.  Anyway, I've addressed your comments so I'll merge soon if no objections.

johnnovak:
> > I'll do that over the coming days, ideally sometime next week.
> 
> Sounds good. Anyway, I've addressed your comments so I'll merge soon if no objections.

Thanks! Merge away, man 😎

interloper98:
This use mentions that their accounting software's multi-user SQL back-end doesn't work under original dosbox, which might point to the lack of share support:

https://www.vogons.org/viewtopic.php?f=7&t=37333&p=328571&hilit=needs%20need%20share%20exe#p328571

My hunch is that it's Peachtree Accounting, because they were massive in the early 2000's -- even their 2004 release of Peachtree Complete Accounting 11 was solely DOS based!

I could only find v 8.0 on archive: https://archive.org/details/peachtree-complete-accounting-8.0-2001, but I haven't tried it w/ the latest CI builds yet.

interloper98:
8rn448

weirddan455:
The DOSBox Staging team will not be held liable for any and all accounting errors.

Actually I think we've had to make a statement like that before.  I vaguely remember it came out that someone was using DOSBox for some engineering software and our floating-point calculations are not always 100% accurate on all platforms compared to old x86 hardware that uses 80-bit precision.

So yeah, we won't be held liable for any collapsed bridges either.

johnnovak:
> The DOSBox Staging team will not be held liable for any and all accounting errors.
> 
> Actually I think we've had to make a statement like that before. I vaguely remember it came out that someone was using DOSBox for some engineering software and our floating-point calculations are not always 100% accurate on all platforms compared to old x86 hardware that uses 80-bit precision.
> 
> So yeah, we won't be held liable for any collapsed bridges either.

[I've already taken care of this.](https://www.dosbox-staging.org/about/#disclaimer)

johnnovak:
@Python-Exoproject Just pinging you to show off the rapid progress we're making on the Win 3.x emulation front 😎

interloper98:
Peachtree Complete Accounting 9.0 runs perfectly.

The startup batch file (`PEACH.BAT`) first runs a tool call `SHARETST.EXE` and from its return code decides if it should run launch the local/stand-alone `PT.EXE` or the network executable `NETPEACH.EXE`.

I suspect the `SHARETST.EXE` can detect if the files are sitting on a network share or are local. (The old way of running this software in a multi-tenant environment is that you have a Novell or NetBUI server hosting the sole copy of the software, and lots of accountants client machines mounting this shared directory. Lots of users can then be working simultaneously, and Peachtree takes care of the serializing/locking/unlocking access behind the scenes).
 
The main Staging CI build and DOSBox-X can run both executables:


https://github.com/dosbox-staging/dosbox-staging/assets/165201780/b21a1b98-1452-489b-9665-c9388a4652b3


https://github.com/dosbox-staging/dosbox-staging/assets/165201780/a72a9b27-9288-4b92-996c-b3d60064dcd5

But older 81.1 can't run them. Not even `FAKESHAR` is enough to fool it:


https://github.com/dosbox-staging/dosbox-staging/assets/165201780/ac503403-186d-417d-8880-e3da72b54b72

Unfortunately I'm not too sure how to get Microsoft's NetBUI client working (and same w/ Novel's 2.7 client + NE2000 drivers). So I was able to test a real DOS-based network share. 

@Torinde - do you know how to share and mount a DOS directory across two DOSBox instances (either Ethernet or IPX, using either MS or Novel?)

PR #3692, opened 2024-05-23
Update homepage URL 

johnnovak:
# Description

In our desperate quest to be taken seriously, we've changed our homepage URL to https://www.dosbox-staging.org/ 😎

I used the following command, then patched up a few things manually:

```
 git grep -l "https://dosbox-staging.github.io" |  xargs sed -i "s#https://dosbox-staging.github.io#https://www.dosbox-staging.org#g"
```

Not sure about the `dosbox-staging.metainfo.xml`. I'm guessing we don't wanna change the `<id>` there, right? Linux folks? @weirddan455 @FeralChild64 @shermp 

```xml
<component type="desktop">
  <id>io.github.dosbox-staging</id>
  <metadata_license>CC0-1.0</metadata_license>
  <project_license>GPL-2.0-or-later</project_license>
  <name>DOSBox Staging</name>
  <summary>DOS/x86 emulator with advanced features</summary>
```

# Manual testing

- Publish dev site, test that the links are not broken.

- Test that the intro screen still looks fine (had to insert two extra spaces after the URL so the border is still aligned correctly; I've made the same change in all language files).

image

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [ ] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
[...]

Issue #3686, opened 2024-05-20
SIGSEV if the code is compiled with PIE


runlevel5:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.81.1

### What Operating System are you using?

Linux ppc64le

### If Other OS, please describe

Fedora 40

### Relevant hardware info

IBM POWER9

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Reported in Fedora at https://bugzilla.redhat.com/show_bug.cgi?id=2276363

TL;DR; the Fedora's default `LDFLAGS` which has `-specs=/usr/lib/rpm/redhat/redhat-hardened-ld` option that would makes the app prone to `SIGSEV` error whenever running any DOS game

The content of `/usr/lib/rpm/redhat/redhat-hardened-ld` spec file is:

```
*self_spec:
+ %{!static:%{!shared:%{!r:-pie}}}
```

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Set `LDFLAGS='-specs=/usr/lib/rpm/redhat/redhat-hardened-ld'` then ` /usr/bin/meson setup --buildtype=plain  --wrap-mode=nodownload --auto-features=enabled . build && cd build && meson compile`  
2. Run `./dosbox`
3. Run any game and would get SIGSEV:

```
Thread 1 "dosbox" received signal SIGSEGV, Segmentation fault.
0x0000000100121518 in CPU_READ_CRX (cr=<optimized out>, retvalue=<error reading variable: Cannot access memory at address 0x1bfa2b0>) at ../src/cpu/cpu.cpp:1670
1670		retvalue=CPU_GET_CRX(cr);
(gdb) bt
#0  0x0000000100121518 in CPU_READ_CRX (cr=<optimized out>, retvalue=<error reading variable: Cannot access memory at address 0x1bfa2b0>) at ../src/cpu/cpu.cpp:1670
#1  0x00007fff9d814c6c in ??? ()
#2  0x00000001001503f4 in CPU_Core_Dynrec_Run () at ../src/cpu/core_dynrec.cpp:278
#3  0x00000001000a66b4 in Normal_Loop () at ../src/dosbox.cpp:146
#4  0x0000000100118194 in DOSBOX_RunMachine () at ../src/dosbox.cpp:322
[...]

PR #3697, opened 2024-05-24
Remove Fedora Linux related build instructions

johnnovak:
# Description

As per @oturpe's suggestion (see [here](https://github.com/dosbox-staging/dosbox-staging/pull/3692#issuecomment-2127918370
)), we're removing the Fedora Linux related contribs scripts as they're outdated and no one actively keeps them up-to-date.

I'm also removing all Fedora-related build info from the website in separate commits; let me know if you think these should stay. @FeralChild64 @weirddan455 @Grounded0 I can restore it if you think it's still needed.

# Checklist

I have:


weirddan455:
I don't think we should be removing Fedora build instructions.  We might get someone who wants to contribute on Fedora and wants to build from source.  This is different than Fedora _packaging._  I have no problem with removing the `contrib/fedora` files.

On the website, I think we should leave the dependency instructions as well.  A user might be on Fedora 39 which is still maintained for another 6 months or so and wants to install v0.81.1.  Fedora 39 still packages v0.80.1 in the repos.  You could put a note near that section that Fedora is in the official repos if we want to promote that.  We do mention it further down on the page but it's not the first thing the user will see.

Grounded0:
People should be pointed out to official packages as a priority.

johnnovak:
> I don't think we should be removing Fedora build instructions. We might get a Fedora user who wants to contribute and needs to to build from source to do so. This is different than Fedora _packaging._ I have no problem with removing the `contrib/fedora` files.
> 
> On the website, I think we should leave the dependency instructions as well. A user might be on Fedora 39 which is still maintained for another 6 months or so and wants to install v0.81.1. Fedora 39 still packages v0.80.1 in the repos. 

Good points @weirddan455, yeah I wasn't sure. I was thinking Fedora is some niche distro no one uses these days. This makes sense, I'll keep the build instructions as you suggested.

> You could put a note near that section that Staging is in the official repos of Fedora if we want to promote that. We do mention it further down on the page but it's not the first thing the user will see.

Yeah, I agree, I will come up with something.

weirddan455:
>  I was thinking Fedora is some niche distro no one uses these days.

It's hard to find accurate usage statistics for distro usage but Fedora is one of the big ones.  It's probably 2nd to Ubuntu based distros if I had to guess.

johnnovak:
Updated. Leaving the Linux page as-is as I think we should point people to our tested binary releases. We have no control over how the distro maintainers repackage our stuff; those are "secondary" artifacts and might contains bugs and all sorts of weirdnesses (e.g., mismatching SDL2 lib versions).

weirddan455:
Mismatching SDL2?  If anything, our binary release would be more prone to that as that's a dynamically linked dependency.  Distributions will compile with the same SDL2 version that their users will actually be using.

Changing build flags is more of a concern here (see #3686 although that bug only affects PPC which we don't even ship a binary release for).

johnnovak:
> Mismatching SDL2? If anything, our binary release would be more prone to that as that's a dynamically linked dependency. Distributions will compile with the same SDL2 version that their users will actually be using.
> 
> Changing build flags is more of a concern here (see #3686 although that bug only affects PPC which we don't even ship a binary release for).

Yeah, I take it back. I always forget we're not shipping our own SDL2 lib on Linux. I just can't get over that "random library version" delivery model, apparently 😆

weirddan455 (approved):
In any case, removing `contrib/fedora` is fine as it's unmaintained and the Fedora downstream isn't using it.

Issue #3700, opened 2024-05-24
ES1688 emulation missing ESS Mixer registers


wbcbz7:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Hey there!
I've been recently checking out DOSBox Staging, and noticed that ESS AudioDrive emulation does not support extended ESS mixer registers, in particular Extended FM Volume register (index 0x36). 
For example, koolnESS by The Furnace Posse uses mixer registers to fade out audio during song changing or exit, and seems that this effect is missing in DOSBox Staging (reference capture from real ES1869 with fadeout here: https://youtu.be/i_4dg8xUaBA?si=cMV6Tn63Y4rMLMHs&t=190)

Thanks in advance! :)

### Download URL of affected game or software

https://www.pouet.net/prod.php?which=96920

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Indeed, thanks for reporting @wbcbz7! DOSBox-X has the 0x36 command implemented, so I missed that when I ported the code over. Enjoy Staging 😎

johnnovak:
> I've been recently checking out DOSBox Staging, and noticed that ESS AudioDrive emulation does not support extended ESS mixer registers

@wbcbz7 I've fixed this in the linked ticket. Are you aware of any other programs that exercise the ESS specific mixer registers?

wbcbz7:
I've just checked the latest dev build, and seems like fadeout is almost there, but it sounds like volume curve being too steep; also volume adjustment (using -/+ keys) doesn't work properly (volume drops by ~12dB and locks at that level). Looking in mixer-related commit, I've noticed there is a [typo](https://github.com/dosbox-staging/dosbox-staging/commit/c52451f2647c4be79baaa23fbccdd71a2ea49e61#diff-a7f7250de582823c8296278848d0e29b8a4dbf4a885ef4832e3f8546d270600aR2591) in `ctmixer_read()` function - it uses register index 0x34 instead of 0x36 for extended FM volume read. As volume control in musicdisk reads current volume from the mixer, this can be the issue for now. Nice progress, anyway :)

[...]

PR #3702, opened 2024-05-25
Add support for ESS mixer registers

johnnovak:
# Description

This PR ports over the ESS-specific mixer regusters from DOSBox-X to read/write the mixer volume levels.

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3700

# Manual testing

Run **koolnESS** referenced in the above ticket, and confirm the music fades out when switching between tunes.

# Checklist

I have:

### DIFF
--- a/include/math_utils.h
+++ b/include/math_utils.h
@@ -252,14 +252,14 @@ constexpr uint32_t clamp_to_uint32(const T val)
 	return static_cast<uint32_t>(std::clamp(val, min_val, max_val));
 }
 
-constexpr uint8_t read_low_nibble(const uint8_t byte)
+constexpr uint8_t low_nibble(const uint8_t byte)
 {
-	return static_cast<uint8_t>(byte & 0x0f);
+	return byte & 0x0f;
 }
 
-constexpr uint8_t read_high_nibble(const uint8_t byte)
+constexpr uint8_t high_nibble(const uint8_t byte)
 {
-	return static_cast<uint8_t>(byte >> 4);
+	return (byte & 0xf0) >> 4;
 }
 
 inline float decibel_to_gain(const float decibel)
--- a/src/hardware/input/mouseif_dos_driver.cpp
+++ b/src/hardware/input/mouseif_dos_driver.cpp
@@ -1871,27 +1871,26 @@ static void prepare_driver_info()
 
 	const std::string str_copyright = DOSBOX_COPYRIGHT;
 
-	auto read_low_nibble_str = [&](const uint8_t byte) {
-		return std::to_string(static_cast<int>(read_low_nibble(byte)));
+	auto low_nibble_str = [&](const uint8_t byte) {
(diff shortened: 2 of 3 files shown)


include/math_utils.h
```diff
@@ -252,14 +252,14 @@ constexpr uint32_t clamp_to_uint32(const T val)
...
-	return static_cast<uint8_t>(byte & 0x0f);
+	return byte & 0x0f;
 }
 
-constexpr uint8_t read_high_nibble(const uint8_t byte)
+constexpr int high_nibble(const int byte)
 {
-	return static_cast<uint8_t>(byte >> 4);
+	return (byte & 0xf0) >> 4;
 }
```
weirddan455, line 263: I get your reasoning on moving stuff to int but I don't like it for functions doing bitwise operations.  As an example, you can run into problems because C will do an arithmetic right shift on signed integers when you usually will want a logical right shift.  It works out to the same thing here but you had to put in a bitwise AND to counteract that.

Also, I don't think the `static_cast` was required in the original code.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -1871,27 +1871,26 @@ static void prepare_driver_info()
...
 
-	auto read_low_nibble_str = [&](const uint8_t byte) {
-		return std::to_string(static_cast<int>(read_low_nibble(byte)));
+	auto low_nibble_str = [&](const int byte) {
+		return std::to_string(low_nibble(byte));
 	};
-	auto read_high_nibble_str = [&](const uint8_t byte) {
-		return std::to_string(static_cast<int>(read_high_nibble(byte)));
+	auto high_nibble_str = [&](const int byte) {
+		return std::to_string(high_nibble(byte));
```
weirddan455, line 1878: Maybe this was your motivation to change this?  I don't think `static_cast` was needed in the original code though.  `uint8_t` to `int` is always safe so should be fine as an implicit cast.

include/math_utils.h
johnnovak: I was on the fence with this one. I'm happy to restore it to accept `uint8_t`s only, I just thought this might more convenient. But probably we'll always dealing with `uint8_t` at the call site anyway when messing around with nibbles.

Technically, I don't think the signed bit would cause problems as I'm doing the AND before the shift, but moot point 😄

src/hardware/input/mouseif_dos_driver.cpp
johnnovak: Yeah, I think that was it.

johnnovak:
Thanks for the review @weirddan455, I've addressed your comment.

weirddan455 (approved):
Ok, the rest looks fine although I'm not familiar with the soundblaster code.

johnnovak:
> Ok, the rest looks fine although I'm not familiar with the soundblaster code.

Thanks man! 😎

PR #3708, opened 2024-05-29
Improve the `core` and `cputype` setting descriptions

johnnovak:
# Description

Also, revert `core` to `normal` if needed for `cputype = 386_prefetch`. Previously, the `core = dynamic` and `cputype = 386_prefetch` combo resulted in a hard exit, which is not exactly great behaviour...

It took me quite a bit of research to find out all these details, and I'm quite confident I got everything right 😎 

### References

- [DOSBox-X Wiki — Differences in emulated CPU types](https://github.com/joncampbell123/dosbox-x/wiki/Guide:CPU-settings-in-DOSBox%E2%80%90X#differences-in-emulated-cpu-types)
- [VOGONS — CPU type options](https://www.vogons.org/viewtopic.php?p=186286#p186286)
- [VOGONS — Games that requires specific cputype settings](https://www.vogons.org/viewtopic.php?p=161679#p161679)
- [VOGONS — `cputype` feature matrix](https://www.vogons.org/viewtopic.php?p=905713#p905713)

The feature matrix is pretty handy, so here it is for posterity:

image

# Manual testing

image

image

# Checklist

I have:

### DIFF
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -3240,21 +3240,59 @@ void init_cpu_dosbox_settings(Section_prop& secprop)
 	pstring->Set_values({
 		"auto",
 #if C_DYNAMIC_X86 || C_DYNREC
-		        "dynamic",
+		"dynamic",
 #endif
-		        "normal", "simple"
+		"normal", "simple"
 	});
+
 	pstring->Set_help(
-	        "CPU core used in emulation ('auto' by default). 'auto' will switch to dynamic\n"
-	        "if available and appropriate.");
+	        "Type of CPU emulation core to use ('auto' by default).\n"
+	        "  auto:     'normal' core for real mode programs, 'dynamic' core for protected\n"
+	        "            mode programs (default). Most programs will run correctly with this\n"
+	        "            setting.\n"
+	        "  normal:   The DOS program is interpreted instruction by instruction. This\n"
+	        "            yields the most accurate timings, but puts 3-5 times more load on\n"
+	        "            the host CPU compared to the 'dynamic' core. Therefore, it's\n"
+	        "            generally only recommended for real mode programs that don't need\n"
+	        "            a fast emulated CPU or are timing-sensitive. The 'normal' core is\n"
+	        "            also necessary for programs that self-modify their code.\n"
+	        "  simple:   The 'normal' core optimised for old real mode programs; it might\n"
+	        "            give you slightly better compatibility with older games. Auto-\n"
+	        "            switches to the 'normal' core in protected mode.\n"
+	        "  dynamic:  The instructions of the DOS program are translated to host CPU\n"
+	        "            instructions in blocks and are then executed directly. This puts\n"
+	        "            3-5 times less load on the host CPU compared to the 'normal' core,\n"
+	        "            but the timings might be less accurate. The 'dynamic' core is a\n"
+	        "            necessity for demanding DOS programs (e.g., 3D SVGA games).\n"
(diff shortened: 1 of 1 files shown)


src/cpu/cpu.cpp
```diff
@@ -3078,21 +3078,26 @@ class Cpu final {
...
-				cpudecoder = &CPU_Core_Prefetch_Run;
-
-				CPU_PrefetchQueueSize = 16;
-
-			} else if (cpu_core == "auto") {
+			if (cpu_core == "auto") {
 				cpudecoder = &CPU_Core_Prefetch_Run;
 
 				CPU_PrefetchQueueSize         = 16;
 				auto_determine_mode.auto_core = false;
```
weirddan455, line 3085: If CPU core is auto, then auto_core is false.  Is that intentional?

src/cpu/cpu.cpp
```diff
@@ -3104,19 +3109,26 @@ class Cpu final {
...
 				cpudecoder = &CPU_Core_Prefetch_Run;
 
-				CPU_PrefetchQueueSize = 32;
-
-			} else if (cpu_core == "auto") {
-				cpudecoder = &CPU_Core_Prefetch_Run;
-
-				CPU_PrefetchQueueSize           = 32;
+				CPU_PrefetchQueueSize         = 32;
 				auto_determine_mode.auto_core = false;
```
weirddan455, line 3116: Same logic here.  Is this is because you're explicitly setting `cpudecoder`?  I'm not going to read through all the logic here.  Just wanted to point it out in case this is a bug.

src/cpu/cpu.cpp
```diff
@@ -3240,21 +3240,59 @@ void init_cpu_dosbox_settings(Section_prop& secprop)
...
+	        "            setting.\n"
+	        "  normal:   The DOS program is interpreted instruction by instruction. This\n"
+	        "            yields the most accurate timings, but puts 3-5 times more load on\n"
+	        "            the host CPU compared to the 'dynamic' core. Therefore, it's\n"
+	        "            generally only recommended for real mode programs that don't need\n"
+	        "            a fast emulated CPU or are timing-sensitive. The 'normal' core is\n"
+	        "            also necessary for programs that self-modify their code.\n"
+	        "  simple:   The 'normal' core optimised for old real mode programs; it might\n"
+	        "            give you slightly better compatibility with older games. Auto-\n"
+	        "            switches to the 'normal' core in protected mode.\n"
```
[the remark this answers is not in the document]
weirddan455: Oh I see what you're saying after re-reading it.  If you set `simple`, it will change to `normal` when you enter protected mode.  I thought you were talking about the `auto` setting.

Nevermind this string is fine :smile:

weirddan455 (approved):
Looks good.  Much more clear what the different settings do now.

src/cpu/cpu.cpp
johnnovak: Yeah, intentional. I've just flipped this around:

image

I agree with your point that it's confusing. Well, the whole CPU code is confusing, and although I tried to improve it with better namings and stuff, it's still incredibly tangled and I'm very reluctant to change the logic substantially.

src/cpu/cpu.cpp
johnnovak: As above, and thanks for flagging it. It's best to double-check such suspicious looking things, no doubt 😄

johnnovak:
> Looks good. Much more clear what the different settings do now.

Thanks man 😎

PR #3711, opened 2024-05-29
Implement read-only mounting + make Y drive read-only by default

### LINKED ISSUE #2602: The `Y` drive should be (ideally) read-only
I've just deleted all commands in `Y:\DOS`. That's not great.

Ideally, that should be a read-only drive to prevent such mishaps from happening. This would actually necessitate adding a "read-only" mount option to the `MOUNT` command and the auto-mounting mechanism, but I assume this can't be too hard given CD-ROM images are read-only too.

image

Read-only files are handled well in programs, e.g. this is what happens when I try to delete the built-in `LOADROM.COM` command in DOS Navigator:

image

weirddan455:
# Description

I mostly stuck to my plan here:  https://github.com/dosbox-staging/dosbox-staging/issues/2602#issuecomment-2134255705

1. Implement a `readonly` flag for `localDrive`.  This already exists for `fatDrive`.  If `readonly` is set, throw a `DOSERR_ACCESS_DENIED` on file writes, creation, and deletions.
2. Implement a `-ro` flag for the `mount` command.  This already exists for `imgmount` but not for `mount`.
3. Add a `readonly` config setting to `[drive]`.  This will simply pass the `-ro` flag to `mount`.
4. Add `readonly = true` to `contrib/resources/drives/y.conf` so it is read-only by default.  Users can of course change this if they want.

The first point just took a lot more commits than I originally thought :smile: 

2, 3, and 4, were very simple.  A single small commit each (last 3 commits of this PR).

The `DOS_Drive` classes could really use some refactoring in some more places even but I did a fair amount of cleanup here.  Ideally, more of the logic would be handled in the generic `DOS_` functions in `dos_files.cpp` regarding what DOS errors to throw and such.

## Related issues

Fixes #2602

# Manual testing

Removing, deleting and creating files is not allowed when `-ro` is set by the mount command, as expected.

The Y drive gets mounted read-only as expected.  May need to do a clean build so it copies the new `y.conf` to the build directory.

Just to see what would happen, I mounted my C drive read-only and tested Doom, Space Quest 1, Space Quest 3.  These games work, only throwing an error when attempting to save the game.  Tyrian does not launch, and this gets logged to console.

```
2024-05-29 16:38:21.983 | FILESYSTEM: protected from modification: palette.dat
2024-05-29 16:38:21.998 | FILESYSTEM: protected from modification: tyrian.cfg
```

It is expected that some games won't like it if they can't write to the hard drive.

# Checklist

I have:


weirddan455:
I guess I've angered the static analyzers somehow so I'll need to clean that up.

include/drives.h
```diff
@@ -124,6 +124,7 @@ class localDrive : public DOS_Drive {
 
 private:
 	bool IsFirstEncounter(const std::string& filename);
+	bool FileIsReadOnly(char* name);
```
johnnovak, line 127: const char* ?

src/dos/drive_local.cpp
```diff
@@ -46,11 +46,22 @@
 #include "cross.h"
 #include "inout.h"
 
+bool localDrive::FileIsReadOnly(char* name)
+{
+	if (isReadOnly()) {
```
johnnovak, line 51: Can we not carry forward old conventions? So `IsReadOnly()` perhaps?

src/dos/drive_local.cpp
```diff
@@ -572,14 +550,14 @@ bool localDrive::AllocationInfo(uint16_t* _bytes_sector, uint8_t* _sectors_clust
 
 bool localDrive::FileExists(const char* name)
 {
-	char newname[CROSS_LEN];
-	safe_strcpy(newname, basedir);
-	safe_strcat(newname, name);
-	CROSS_FILENAME(newname);
-	dirCache.ExpandNameAndNormaliseCase(newname);
+	const std::string system_filename = GetSystemFilename(name);
```
johnnovak, line 536: I was wondering about this `GetSystemFilename` function myself. Does it map the passed-in name to the name on the host filesystem? If so, could we rename it to `MapDosToHostFilename()`? 

It would be nice to use "DOS filename" and "host filename" consistently and "system file" has unfortunate connotations and is rather unclear. First I thought this function deals with some special system files and it confused the hell out of me 😄

include/drives.h
```diff
@@ -87,10 +87,10 @@ class localDrive : public DOS_Drive {
 	localDrive(const char* startdir, uint16_t _bytes_sector,
 	           uint8_t _sectors_cluster, uint16_t _total_clusters,
 	           uint16_t _free_clusters, uint8_t _mediaid,
-	           bool _always_open_ro_files = false);
+	           bool _readonly, bool _always_open_ro_files = false);
 	bool FileOpen(DOS_File** file, char* name, uint8_t flags) override;
-	virtual FILE* GetSystemFilePtr(const char* const name, const char* const type);
-	virtual bool GetSystemFilename(char* sysName, const char* const dosName);
+	FILE* GetSystemFilePtr(const char* const name, const char* const type);
+	std::string GetSystemFilename(const char* const dosName);
```
johnnovak, line 93: I'm not saying to refactor the whole file, but it would beg good to fix the naming conventions when we touch a function. So `dos_name` in this case, etc.

johnnovak (approved):
Very nice work @weirddan455, quite a heroic effort. I'm glad to see your improving this super messy code, and I admire your patience 😎 Just a few naming related suggestions; it would be nice to improve the weird function names in the API at least that are really not intuitive at all (to me, at least).

include/drives.h
weirddan455: It was calling into another function that wasn't using const.  I pushed a commit to just `const` all the `char *` arguments in these classes.  Luckily it just compiled so no one was modifying the strings in place.

src/dos/drive_local.cpp
weirddan455: Renamed.  Also renamed `isRemovable` and `isRemote` for consistency.

src/dos/drive_local.cpp
weirddan455: That's a good point.  Renamed as you suggested and also renamed `GetSystemFilePtr` to `GetHostFilePtr`

include/drives.h
weirddan455: Renamed

Issue #1099, opened 2021-07-02
"CTRL-ALT-F8	Start/Stop the recording of raw MIDI commands." is missing


LowLevelMahn:
https://www.dosbox.com/wiki/Special_Keys
dosbox\docs\dosbox.1
...
.IP CTRL\-ALT\-F8
Start/Stop the recording of raw MIDI commands.
...

dosbox-staging\docs\dosbox.1
missing that line

and it also does not seem to work in staging

kcgen:
This was deliberately changed under https://github.com/dosbox-staging/dosbox-staging/issues/585

@dreamer - is there a different hotkey we can use for MIDI recording?

@HW-Scanno has also expressed a need for this, and has actually produced some MIDI remixes using this feature (after manually mapping the feature).

We looked in Ctrl+F8, and it's surprisingly open on all OSes - perhaps we can use that?

LowLevelMahn:
FYI why i need that feature: i try to record differences between original and reversed game sound driver

LowLevelMahn:
would be also nice to have a command line parameter that captures the whole dosbox session

kcgen:
We've got three different audio stream types.
With the current approach, each stream needs a separate hotkey.

What if "Record Audio" (Ctrl + F6) did that holistically? 
 - PCM samples
 - OPL instructions, if present
 - MIDI sysex, if present

With this approach, I'd only have to remember: _"Ctrl+F6 records my audio"._  and I'm done.
Likewise, a single command-line argument `--record-audio` could do the same.

@LowLevelMahn , @HW-Scanno -- what do you guys think?

LowLevelMahn:
and do all streams end in the same file or different?

kcgen:
> and do all streams end in the same file or different? i would say different - same as now but started with a single hotkey
> is that much more pressure for the CPU? (needs to be at least buffered or else the game will have laggy sound)

The plan would be have this mode write three seperate files. The nice thing about audio capture is that it happens in "game time", not "wall time", within the game loop, and is not affected by host-side performance.  

This mean that even if a host is busy or has competing workloads and DOSBox cannot keep up (use experience slow frame-rates and audio drop-outs), the recording side will will be unaffected by these anomalies because it writes data reliably every pass through the game loop.. however long that takes.

LowLevelMahn:
and for the combined sound - if the game is using multiple outputs parallel (is that even possible?) one can use an extern recording tool

[...]

PR #3714, opened 2024-05-31
Map 'Capture MIDI action' to Ctrl+Alt+F6 (Cmd+Opt+F6) by default

johnnovak:
# Description

What the title says—pretty uncontroversial, I hope. I've been on the fence about adding this mapping by default, but I see no harm in it, so doing it because it came up a few times lately.

Addresses https://github.com/dosbox-staging/dosbox-staging/issues/1099

# Manual testing

The capture MIDI hotkey is mapped by default to Ctrl+Alt+F6 (Cmd+Opt+F6) and it does its job 😄 

# Checklist

I have:

### DIFF
--- a/src/capture/capture.cpp
+++ b/src/capture/capture.cpp
@@ -617,8 +617,8 @@ static void init_key_mappings()
 	                  "Rec. Audio");
 
 	MAPPER_AddHandler(handle_capture_midi_event,
-	                  SDL_SCANCODE_UNKNOWN,
-	                  0,
+	                  SDL_SCANCODE_F6,
+	                  PRIMARY_MOD | MMOD2,
 	                  "caprawmidi",
 	                  "Rec. MIDI");
 


shermp (approved):
Seems to work fine here on Win 11.

weirddan455:
Just FYI this default hotkey is not usable on Linux.  Ctrl+Alt+F* are used to switch TTYs.  It'll switch from the graphical desktop to a text console.

johnnovak:
> Just FYI this default hotkey is not usable on Linux. Ctrl+Alt+F* are used to switch TTYs. It'll switch from the graphical desktop to a text console.

Huh, that is still a thing in graphical Linux environments in 2025? 😆 LOL. I get it it's useful for pure terminal users in text-mode only mode... but just open a new terminal window in a graphical environment?

I guess I could make it Alt+F6. How about that?

johnnovak:
> Maybe I misunderstand something, but is it really needed to separate the different captures - audio, MIDI, etc.? The audio recording hotkey can be used to simultaneously capture also the MIDI action (if any; if there isn't - no file is created) - user gets two files and discards the one they don't need (or keeps both).
> 
> Long ago maybe that was a performance problem, but now there are many threads, etc.?

That "capture everything audio related, then let the user sort out and delete files" is a very heavy handed approach. Why should I let DOSBox create a several hundred megabyte WAV file if all I want to capture is some MIDI stream which is several orders of magnitude smaller?

Besides, creating junk files you did not ask for is very bad user experience.

I see only drawbacks in this idea, and no benefits.

Issue #3594, opened 2024-04-12
Consider using the 16-bit integer renderer of libmt32emu


johnnovak:
Curently, we're using the float renderer which has less quantisation noise, but it's inauthentic. Such noises are part of the charm and character of the original hardware. Moreover, the _less authentic_ float renderer is several times _less_ performant than the authentic 16-bit integer renderer.

Ultimately, we should only keep using the float renderer if the quantisation noise is really annoying. This needs a bit of testing.

More info:

https://www.vogons.org/viewtopic.php?p=1251914#p1251914

https://www.vogons.org/viewtopic.php?p=587185#p587185

https://www.vogons.org/viewtopic.php?f=40&t=51306

PR #3729, opened 2024-06-04
Various MT-32 improvements & fix Bumpy's Arcade Fantasy intro tune

johnnovak:
# Description

## Summary

This changes a couple of things in our libmt32 config:

- Turn off MIDI delay emulation (see reasoning below). This fixes the intro tune in **Bumpy's Arcade Fantasy**, and probably other game startup scenarios. Thanks to @madbrain76 for raising this 🎉 
- Use 16-bit integer rendering mode which is more accurate to the real hardware _and_ a lot faster (I couldn't hear any difference, but the performance matters). This resolves https://github.com/dosbox-staging/dosbox-staging/issues/3594.
- Use "nice" partials mode (this skips the emulation of a quirk that results in weird-sounding notes under some conditions; it's hard to imagine this is a desirable thing to emulate)

## Details

Enabling delay emulation could result in missed MIDI events if the MT-32 receives a large number of events in quick bursts, causing its internal ring buffers to overflow. This can be heard as missed notes and wrong sounding instruments (e.g., in the intro tune of Bumpy's Arcade Fantasy).

Emulating MIDI protocol induced micro-delays are neither musically significant nor desirable. Most importantly, these odd up to 1-3 ms delays on some events are not noticeable. Losing MIDI events that potentially set up the correct sounds when the game starts is a much more important concern.

While such transfer bursts are most likely an MPU-401 intelligent mode emulation bug in DOSBox, disabling the delay mode emulation in libmt32 effectively fixes the problem and makes the resulting music sound closer to the composer's intentions.

# Related discussion

- https://github.com/munt/munt/issues/119

# Manual testing

- Confirmed **Bumpy's Arcade Fantasy** plays _more correctly_ now, and less notes are cut off from the start of the intro tune. Also, the instrument sounds are now close to the Amiga soundtrack:
https://www.youtube.com/watch?v=hju8oJdzvo8

   @Python-Exoproject ☝🏻 This might be relevant to your interests 😎 

- Tested a few other MT-32 games (Eric the Unready, Space Quest 3, Spellcasting 101). All sound fine.

# Checklist

I have:

### DIFF
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -60,10 +60,11 @@ constexpr auto AnalogMode = MT32Emu::AnalogOutputMode_ACCURATE;
 // within range) Higher quality than the real devices
 constexpr auto DacEmulationMode = MT32Emu::DACInputMode_NICE;
 
-// Analog rendering types: int16_t, FLOAT
-// Use float samples in the renderer and simplified wave generator model.
-// Maximum output quality and minimum noise.
-constexpr auto RenderingType = MT32Emu::RendererType_FLOAT;
+// Analog rendering types: BIT16S, FLOAT
+// Use 16-bit signed samples in the renderer and the accurate wave generator
+// model based on logarithmic fixed-point computations and LUTs. Maximum
+// emulation accuracy and speed (it's a lot faster than the FLOAT renderer).
+constexpr auto RenderingType = MT32Emu::RendererType_BIT16S;
 
 // Sample rate conversion quality: FASTEST, FAST, GOOD, BEST
 constexpr auto ResamplingQuality = MT32Emu::SamplerateConversionQuality_BEST;
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -82,9 +82,9 @@ constexpr bool UseNiceRamp = true;
 // possible on the real hardwre).
 constexpr bool UseNicePanning = true;
 
-// Prefer not forcing always in-phase partial mixing (this is more authentic
-// sounding).
-constexpr bool UseNicePartialMixing = false;
+// Force always in-phase partial mixing to avoid the sometimes out-of-phase
+// partials of the original units which can result in phasing artifacts.
+constexpr bool UseNicePartialMixing = true;
 
 // Do not attempt to emulate delays introduced by the slow MIDI transfer
 // protocol.
--- a/src/midi/midi_mt32.cpp
+++ b/src/midi/midi_mt32.cpp
@@ -51,7 +51,14 @@
 // ----------------
 
 // Analogue circuit modes: DIGITAL_ONLY, COARSE, ACCURATE, OVERSAMPLED
+//
+// Accurate mode achieves finer emulation of LPF circuit. Output signal is
+// upsampled to 48 kHz to allow emulation of audible mirror spectra above 16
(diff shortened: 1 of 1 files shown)


weirddan455 (approved):
These all look like reasonable changes to me.  Nice one :+1: 

I assume this also will resolve #3594

johnnovak:
> These all look like reasonable changes to me. Nice one 👍

Cheers @weirddan455.

> I assume this also will resolve #3594

Yes it does, thanks for reminding me. Added reference to the top description.

madbrain76:
Thanks for working on this. I downloaded the build from https://productionresultssa8.blob.core.windows.net/actions-results/0b426d30-4dd0-4122-87a3-21eb885064e2/workflow-job-run-5322bc1b-3da0-5fd0-3647-1cd42d79cd88/artifacts/bafe317abd4f0547330c77f942b6a580cec9e66b5c815196f78608e5df2d814b.zip?rscd=attachment%3B+filename%3D%22dosbox-staging-windows-x64-0.82.0-alpha-0979a.zip%22&se=2024-06-05T00%3A19%3A32Z&sig=93nfRSY%2FcN4iXQmgiNL6Wm%2F8lgLYemwtwgin%2F8rXq4E%3D&ske=2024-06-05T08%3A10%3A35Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-06-04T20%3A10%3A35Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2023-11-03&sp=r&spr=https&sr=b&st=2024-06-05T00%3A09%3A27Z&sv=2023-11-03 

I'm not seeing any improvement with Bumpy or Baby Jo. Does this build have these MT-32 fixes ? I looked at history a bit and it seems like it should have them, but I may be confused.

johnnovak:
@madbrain76 Not sure where you get that link from, please download our latest dev builds from here:

https://www.dosbox-staging.org/releases/development-builds/

The current dev build contains this fix.

madbrain76:
@johnnovak 
That's the page I went to. Then I clicked on "Windows" in the middle of the page (not the top, which goes to stable release).
That wasn't an actual download link. It led me to to https://github.com/dosbox-staging/dosbox-staging/actions/runs/9375202623 . I then downloaded the artifact at the bottom. Is there another place I should have looked at ?

johnnovak:
> @johnnovak That's the page I went to. Then I clicked on "Windows" in the middle of the page (not the top, which goes to stable release). That wasn't an actual download link. It led me to to https://github.com/dosbox-staging/dosbox-staging/actions/runs/9375202623 . I then downloaded the artifact at the bottom. Is there another place I should have looked at ?

Ah right, yes that was correct then.

madbrain76:
@johnnovak , so I switched to the internal Munt since I was using the external one. The startup of Bumpy is still cut off, though. I believe in the very same spot, from what I can see. Unless I applied the patch incorrectly (I don't think I did) then I'm really not seeing any change for Bumpy.

johnnovak:
> @johnnovak , so I switched to the internal Munt since I was using the external one. The startup of Bumpy is still cut off, though. I believe in the very same spot, from what I can see. Unless I applied the patch incorrectly (I don't think I did) then I'm really not seeing any change for Bumpy.

The intro music uses very different instruments with the fix, should be obvious to spot. There is still a small cut note at the start, but much less noticeable.

Issue #3670, opened 2024-05-12
Implement `MODE` Command


MasterO2:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 10

### Is your feature request related to a problem? Please describe.

Dosbox-Staging cannot natively adjust the number of rows and columns at the command prompt:  

https://www.reddit.com/r/dosbox/comments/1cl1rlo/can_the_dos_terminal_size_be_adjusted_in_terms_of/

### Describe the solution you'd like

Dosbox-Staging should implement its own `MODE` command to address the issue.

### Describe alternatives you've considered

**Alternative 1:**

On Dosbox-Staging's Discord channel, user `interloper` proposed the following:

> Looking at my default install, mode.com isn't part of the builtin or the addon Y:\DOS programs.   I tested the vidmode.zip here: https://ftp.sunet.se/mirror/archive/ftp.sunet.se/pub/simtelnet/msdos/screen/
I just dropped the 25.COM, 43.COM, and 50.COM files into the resource\drives\y\dos in my install area, and now these tools are in my PATH and i can switch between the modes.  Description of all those utilities in index.html - https://ftp.sunet.se/mirror/archive/ftp.sunet.se/pub/simtelnet/msdos/screen/00_index.txt

Those `.COM` files are a passable workaround, but `MODE` would address this issue more cleanly.

**Alternative II:**

Dosbox-X has ttf support which allows adjusting of font sizes as well as the settings below in the `[ttf]` section of `dosbox-x.conf`:

`lins = 0`
`cols = 0`

**Alternative III:**

Dosbox-X has its own `MODE` command.

### Add any other context or screenshots about the feature request here.

`MODE` command from **Dosbox-X**:

![command_000](https://github.com/dosbox-staging/dosbox-staging/assets/43938622/62c170f6-1841-42f4-8994-7ca630f5fe1c)

_No response_

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
@FeralChild64 I'm asking your opinion on this because the original MS-DOS `MODE` command was a bit of Swiss Army Knife of everything and it had support for changing codepages and COM & LPT port stuff as well. Because you've added lots of COM port and localisation related functionality and config settings, my understanding is using the `MODE` command for these things might be redundant or perhaps even undesired? 

From here:
https://home.csulb.edu/~murdock/mode.html

```
[...]

PR #3741, opened 2024-06-06
Add MODE command & various text mode improvements

johnnovak:
# Description

Implements https://github.com/dosbox-staging/dosbox-staging/issues/3670.

I ended up on a wild goose chase with this one again... Took a grand tour in the bowels of the VGA code one more time, and I've done some cleanups & minor refactors along the way. After a few rewrites, the final solution ended up being quite simple, but I think it's worthwhile to keep the cleanups anyway. The real changes are in the last few commits.

I ended up adding custom BIOS modes for the various non-standard text modes the MODE command supports (just on VGA, though). This has several benefits:

- I could fine-tune the timings to make these modes fill the screen completely.
- It turns out this is the most compatible way, as the "traditional way" of setting up a standard BIOS mode first and then tweaking it further results in some weirdness (e.g., the tweaks can, under some circumstances, "bleed into" subsequent standard INT 10h mode changes. This might be an emulation bug, but I'm really hesitant to change the VGA text mode stuff significantly).
- You can use these modes in programs that let you set custom BIOS modes. DOS Navigator 1.51 is a prime example—you can configure DN to set any valid BIOS text mode on startup _and_ restore it whenever an executable launched from DN exits. Like a pro 😎

These custom modes are safe additions as programs can't list the available BIOS modes pre-VESA, so they wouldn't even know they exist.

Interestingly, 132-column modes on the Tseng and Paradise adapters seem to be at least partially broken—they can be set all right, but the DOS prompt will behave rather erratically (e.g., disappearing cursor, black screen, etc.) I verified this going back to 0.79.1; it's not a recent regression, and even tested it in SVN. Yep, it's broken in SVN too, so it wasn't our doing.  So I've simply excluded the 132-column modes from the Tseng and Paradise.

I've also tweaked the timings of some of the existing modes a little bit so the image fills the screen better, and made the 132-column modes uniformly "widescreen" (that's a nice use-case of the 132-column modes on modern screens).

Also fixed a bug where the 8x14 font was set instead of the 9x14 variant on VGA for BIOS modes that use 14-line fonts. According to the "Programmer's Guide to the EGA, VGA and Super VGA Cards (3rd ed.)", VGA cards prefer the 9-dot fonts in all modes if they're available. Well, I guess I could test it on my S3 cards, but I'm too lazy to do that 😄 In any case, 9x14 looks better, and I'm 90% sure this is authentic.

8-dot fonts can still be forced on VGA with the `vga_8dot_font` setting (@FeralChild64 can be happy as he seemed to like 8-dot fonts). I tried adding support for selecting 8 vs 9-dot fonts, but it turned out to be a bad idea as it's hard to do correctly, so I've dropped the feature.

Help text after setting `mode 80x30`:

image

# Manual testing

Tested:

- All modes on all supported graphics adapters, including `vga_8dot_font = on` on VGA adapters
- All error scenarios on different graphics adapters
- The typematic rate settings

# Checklist

I have:


src/dos/program_mode.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+// clang-format off
+static std::set<std::string> all_valid_video_modes = {
+ 	 "40x25",
+	 "80x25",  "80x28",  "80x30",  "80x34",  "80x43",  "80x50",  "80x60",
+	"132x25", "132x28", "132x30", "132x34", "132x43", "132x50", "132x60"
+};
+// clang-format on
+
+static std::set<std::string> valid_video_modes_hercules = {"40x25"};
```
rderooy, line 116: 80x25 instead?

johnnovak: Good pickup, yes 😄

MasterO2:
One small suggestion, @johnnovak:  Specify in the `MODE` explanation page that the default display mode is 80x43.

johnnovak:
> One small suggestion, @johnnovak: Specify in the `MODE` explanation page that the default display mode is 80x43.

80x25 is the default mode. I'll consider mentioning it, but probably I won't.

src/dos/program_mode.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <string>
+
+class MODE final : public Program {
+public:
+	MODE()
+	{
+		AddMessages();
+		help_detail = {HELP_Filter::All,
+		               HELP_Category::File,
```
FeralChild64, line 35: `HELP_Category::File` -> `HELP_Category::Misc`

You can see the result in `help /all` command output, the parameter decides upon section the command is listed in.

src/dos/program_mode.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+	IO_Write(0x3c4, ClockingModeRegisterIndex);
+	auto clocking = ClockingModeRegister{IO_Read(0x3c5)};
+
+	// This selects between 8 and 9-dot fonts, which also switches the
+	// horizontal resolution between 640 and 720 pixels in 80-column modes,
+	// and 1056 and 1188 in 132-column modes.
+	clocking.is_eight_dot_mode = (width == FontWidth::_8dot);
+
+	IO_Write(0x3c4, ClockingModeRegisterIndex);
```
FeralChild64, line 84: We have constants for I/O addresses in `inout.h`; it would be nice to continue using this convention.

FeralChild64:
Just two remarks. I've played a little bit with the command - it mostly works, but a lot of DOS software cannot handle the extended text modes, sadly. There are, however, problems with `8dot` and `9dot` commands.
1. If `vga_8dot_font` is enabled in config file, switching between 8/9 dot modes does not work.
2. 8 dot font set by the `vga_8dot_font` option is different than the one set by the `8dot` parameter - see the screenshots below, look at _mo_ letters in _mode_ word. IMHO it's the `vga_8dot_font` option that did it right.

johnnovak:
> Just two remarks. I've played a little bit with the command - it mostly works, but a lot of DOS software cannot handle the extended text modes, sadly.

Yeah, that's expected. And practically all graphical programs reset the 80x25 mode on exit anyway. Maybe I could hook into the exit process to set my own mode, but not sure it's worth it.

>     1. If `vga_8dot_font` is enabled in config file, switching between 8/9 dot modes does not work.

That's expected. That flag emulates old VGA boards that did not have the 9dot fonts. Effectivy it forces 8dot mode. But I could do better and raise a warning or error if it's set.

>     2. 8 dot font set by the `vga_8dot_font` option is different than the one set by the `8dot` parameter - see the screenshots below, look at _mo_ letters in _mode_ word. IMHO it's the `vga_8dot_font` option that did it right.

I will look into this, I haven't noticed 😄

FeralChild64:
I have also noticed that the text cursor blinks much more slowly in modes like 132x30 than in 80x25.

johnnovak:
> I have also noticed that the text cursor blinks much more slowly in modes like 132x30 than in 80x25.

Indeed, you have a good eye. I've never noticed.

That's because the cursor blink rate is apparently tied to the refresh rate of the video mode.

80x30 -> 60 Hz
80x25 -> 70 Hz
80x43 -> 81 Hz

132x25 -> 37 Hz
etc...

The non-VESA 132 Hz modes have sub-40 Hz refresh rates. I suspect this is most likely an emulation inaccuracy as most non-multisync VGA monitors couldn't sync below 50 Hz. In any case, I'm not gonna delve deeper... 😅 

What I could do, though, is to use the VESA 132-column modes instead as those are uniformly around 69-70 Hz with the more flexible VESA timings. Then I could convert my extra 132-column modes to VESA modes.

johnnovak:
> Just a curiosity of mine, not related to the code here - to me the DOS prompt cursor (the dash that blinks) doesn't look like any of the ASCII characters. Am I wrong - is it an ASCII character? Or is it custom-drawn somehow?

Google is a wonderful thing 😎  This is what my guess was before looking it up, btw 😏 

https://retrocomputing.stackexchange.com/a/5604

src/ints/int10_modes.cpp
```diff
@@ -1949,51 +2055,67 @@ bool INT10_SetVideoMode(uint16_t mode)
...
-		IO_Write(crtc_base,0x58);IO_Write(crtc_base+1,0x3);		//Enable 8 mb of linear addressing
 
-		IO_Write(crtc_base,0x38);IO_Write(crtc_base+1,0x48);	//Register lock 1
-		IO_Write(crtc_base,0x39);IO_Write(crtc_base+1,0xa5);	//Register lock 2
+		IO_Write(crtc_base,0x3a);
+		IO_Write(crtc_base+1,reg_3a);
+
+		IO_Write(crtc_base,0x31);
+		// Enable banked memory and 256k+ access
+		IO_Write(crtc_base+1,reg_31);
```
japsmits, line 2093: Where does the comment belong? Should it be something like this:
```suggestion
		// Enable banked memory and 256k+ access
		IO_Write(crtc_base,0x31);
		IO_Write(crtc_base+1,reg_31);
```
If so, this comment goes for quite a few other places as well.

src/ints/int10_modes.cpp
```diff
@@ -1647,84 +1706,112 @@ bool INT10_SetVideoMode(uint16_t mode)
...
 	case M_TANDY16:
 		// TODO: TANDY_16 seems like an oversight here, as
 		//       this function is supposed to deal with
 		//       MCH_EGA and MCH_VGA only.
-		att_data[0x10]=0x01;		//Color Graphics
-		for (uint8_t ct=0;ct<16;ct++) att_data[ct]=ct;
+
+		// Color Graphics
+		att_data[0x10] = 0x01;
+		for (uint8_t ct = 0; ct < 16; ct++) {
```
japsmits, line 1737: Strictly speaking, it would have to become the following:
```suggestion
		for (uint8_t ct = 0; ct < 16; ++ct) {
```
But since this is obviously a reformat, this can wait for some other time. 😄

src/ints/int10_memory.cpp
```diff
@@ -22,112 +22,179 @@
...
+
+	uint16_t base = real_readw(BIOSMEM_SEG, BIOSMEM_CRTC_ADDRESS);
+
+	bool mono = (base == VGAREG_MDA_CRTC_ADDRESS);
+
+	// Put video adapter in planar mode
+	IO_Write(0x3c4, 0x02);
+
+	// select plane 2 for writing
+	IO_Write(0x3c5, 0x04);
```
japsmits, line 63: Formatting here is borked 🐔 . The comments are misplaced. I would expect something like:
```
// Put video adapter in planar mode

// select plane 2 for writing
IO_Write(0x3c4, 0x02);
IO_Write(0x3c5, 0x04);

// odd/even off in SEQ
IO_Write(0x3c4, 0x04);
IO_Write(0x3c5, 0x07);

...
```

src/ints/int10_memory.cpp
```diff
@@ -22,112 +22,179 @@
...
+    /* e */ 0x00,  // Change to add new functions
+    /* f */ 0x00   // reserved
 };
 
-static uint16_t map_offset[8]={
-	0x0000,0x4000,0x8000,0xc000,
-	0x2000,0x6000,0xa000,0xe000
+static std::vector<uint16_t> map_offset = {
+	0x0000, 0x4000, 0x8000, 0xc000,
+	0x2000, 0x6000, 0xa000, 0xe000
```
japsmits, line 44: `static const std::array<uint16_t, 8>` instead of a vector? Or is this supposed to be expandable?

src/ints/int10_memory.cpp
```diff
@@ -22,112 +22,179 @@
...
- /* 8 */ 0x04,  // total number of character blocks available in text modes
- /* 9 */ 0x02,  // maximum number of active character blocks in text modes
- /* a */ 0xff,  // Misc Flags Everthing supported 
- /* b */ 0x0e,  // Support for Display combination, intensity/blinking and video state saving/restoring
- /* c */ 0x00,  // reserved
- /* d */ 0x00,  // reserved
- /* e */ 0x00,  // Change to add new functions
- /* f */ 0x00   // reserved
+// clang-format off
+static uint8_t static_functionality[0x10] = {
```
japsmits, line 26:
```suggestion
static const std::array<uint8_t, 0x10> static_functionality = {
```

src/ints/int10.cpp
```diff
@@ -632,34 +797,45 @@ static Bitu INT10_Handler(void) {
...
 		}
 		break;
+
 	case 0xff:
-		if (!warned_ff) LOG(LOG_INT10,LOG_NORMAL)("INT10:FF:Weird NC call");
-		warned_ff=true;
+		if (!warned_ff) {
+			LOG(LOG_INT10, LOG_NORMAL)("INT10: FF:Weird NC call");
+		}
+		warned_ff = true;
```
japsmits, line 836:
```suggestion
		if (!warned_ff) {
			LOG(LOG_INT10, LOG_NORMAL)("INT10: FF:Weird NC call");
			warned_ff = true;
		}
```

src/ints/int10_memory.cpp
```diff
@@ -22,112 +22,179 @@
...
+	// select plane 0 for reading
+	IO_Write(0x3cf, 0x00);
+	IO_Write(0x3ce, 0x05);
+
+	// write mode 0, odd/even on in GFX
+	IO_Write(0x3cf, 0x10);
+	IO_Write(0x3ce, 0x06);
+
+	// Bx000-BxFFF, odd/even on
+	IO_Write(0x3cf, mono ? 0x0a : 0x0e);
```
japsmits, line 118: 👨‍🍳🐔 
```suggestion
	// Return to normal text mode

	// select planes 0&1 for writing
	IO_Write(0x3c4, 0x02);
	IO_Write(0x3c5, 0x03);

	// odd/even on in SEQ
	IO_Write(0x3c4, 0x04);
	IO_Write(0x3c5, 0x03);

	// select plane 0 for reading
	IO_Write(0x3ce, 0x04);
	IO_Write(0x3cf, 0x00);

	// write mode 0, odd/even on in GFX
	IO_Write(0x3ce, 0x05);
	IO_Write(0x3cf, 0x10);

	// Bx000-BxFFF, odd/even on
	IO_Write(0x3ce, 0x06);
	IO_Write(0x3cf, mono ? 0x0a : 0x0e);
```

src/dos/program_mode.cpp
```diff
@@ -0,0 +1,293 @@
...
+		MoreOutputStrings output(*this);
+		output.AddString(MSG_Get("PROGRAM_MODE_HELP_LONG"));
+		output.Display();
+		return;
+	}
+
+	auto is_set_display_mode_command = [](const std::string& command) {
+		// Set display mode commands start with digits (e.g., 80x43)
+		constexpr auto MinLength = sizeof("80x43") - 1;
+		return isdigit(command[0]) && command.size() >= MinLength;
```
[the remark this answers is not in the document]
johnnovak: I'm doing that check in the function where I handle the video mode command.

src/ints/int10_modes.cpp
johnnovak: I'm not gonna spend more time on these reactors now, just making small gradual improvements for now.

src/ints/int10.cpp
johnnovak: Okay, but bear in mind I'm just doing minimal cleanups/reformats as parts of this change. I intentionally did _not_ want to even begin to understand that code too much; this was just cosmetic cleanup.

src/ints/int10_memory.cpp
johnnovak: Cheers, yeah I split the lines first, then moved the comments around, that was my mistake 😄

src/ints/int10_memory.cpp
johnnovak: Doing the `static const` part, but leaving it as a vector. I dislike having to specify the size; that's error-prone. `std::vector` is just an array if you never resize it, so personally I'd just default to `std::vector` everywhere.

src/ints/int10_memory.cpp
johnnovak: Adding `const`, though.

src/ints/int10_modes.cpp
japsmits: Understandable, but then it might be preferable to revert these changes completely. The placing of the comments does not match the intention of the previous code, and if this is committed as is, the association between code and comment is lost.

src/ints/int10_modes.cpp
```diff
@@ -1949,51 +2056,67 @@ bool INT10_SetVideoMode(uint16_t mode)
...
+		// Enable 8 mb of linear addressing
+		IO_Write(crtc_base+1,0x3);
+
+		IO_Write(crtc_base,0x38);
+		// Register lock 1
+		IO_Write(crtc_base+1,0x48);
+
+		IO_Write(crtc_base,0x39);
+		// Register lock 2
+		IO_Write(crtc_base+1,0xa5);
```
japsmits, line 2106: I'll give you a hand. 😄 
```suggestion
		IO_Write(crtc_base,0x3a);
		IO_Write(crtc_base+1,reg_3a);

		// Enable banked memory and 256k+ access
		IO_Write(crtc_base,0x31);
		IO_Write(crtc_base+1,reg_31);

		// Enable 8 mb of linear addressing
		IO_Write(crtc_base,0x58);
		IO_Write(crtc_base+1,0x3);

		// Register lock 1
		IO_Write(crtc_base,0x38);
		IO_Write(crtc_base+1,0x48);

		// Register lock 2
		IO_Write(crtc_base,0x39);
		IO_Write(crtc_base+1,0xa5);
```

src/ints/int10_modes.cpp
johnnovak: Yeah, I agree with fixing the comments. I thought you were referring to other cosmetic changes. Thanks for catching these.

src/ints/int10_modes.cpp
johnnovak: Thanks @japsmits 😄

johnnovak:
Alright, @FeralChild64, I've changed things so I'm using the 132-column VESA modes now, so you got your 70 Hz cursor blink rate back 😎 

Plus a bunch of other cleanups, refactors, etc... the usual stuff.

@japsmits I believe I've addressed all your comments too.

FeralChild64 (approved):
I have no more remarks. Played a little with the new command, didn't find any problems now.

rderooy:
Would it make sense from a compatibility perspective with the MS-DOS mode command, to add support for the `co40`, `co80`, and `mono` mode options? Perhaps just as simple aliases for the 25x40 and 25x80 modes (assuming the emulated video adapter supports it)?

interloper98:
Look at all those modes!

```c++
...
        { "40x25", 0x001}, { "80x25", 0x003}, { "80x28", 0x070},
        { "80x30", 0x071}, { "80x34", 0x072}, { "80x43", 0x073},
        { "80x50", 0x074}, { "80x60", 0x043}, {"132x25", 0x109},
        {"132x28", 0x230}, {"132x30", 0x231}, {"132x34", 0x232},
        {"132x43", 0x10a}, {"132x50", 0x10b}, {"132x60", 0x10c}
...
        { "40x25", 0x01}, { "80x25", 0x03}, { "80x28", 0x70},
        { "80x30", 0x71}, { "80x34", 0x72}, { "80x43", 0x73},
        { "80x50", 0x74}
...
```

Amazing, @johnnovak! Blows away the external tool I've been using.

out

johnnovak:
> Would it make sense from a compatibility perspective with the MS-DOS mode command, to add support for the `co40`, `co80`, and `mono` mode options? Perhaps just as simple aliases for the 25x40 and 25x80 modes (assuming the emulated video adapter supports it)?

@rderooy Yeah I can add the `co40` and `co80` aliases easily. Not sure adding a new `mono` mode is worth it, but I can alias it to 80x25 so at least the command succeeds.

I'll merge this then I'll do that in a separate PR.

johnnovak:
Thanks @FeralChild64 and @interloper98. Hopefully these new text modes won't leave people wanting for anything else 😄

Issue #3691, opened 2024-05-23
Keyboard input delay/locking/queuing in Gods (Bitmap Brothers)


FatBarry:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 11

### Relevant hardware info

Gigabyte AORUS 17 XE4 (Core™ i7-12700H, RTX 3070 Ti)

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The issue has only been identified when playing Gods (Bitmap Brothers). No issues identified with other games played so far.

As mentioned in Discord: https://discord.com/channels/514567252864008206/626653390830698516/1242998746153877534

Details:
- Keyboard inputs appear to be delayed and/or queued up (multiple actions despite pressing key once) and sometimes locked (had one game where "left" was locked).
- The issue doesn't always occur immediately - sometimes it can take 30 seconds or so before occurring.
- The issue didn't appear to be present in DOSBox Staging 0.80.1, DOSBox 0.74-3 or DOXBox-X 2024.03.01
- Issue occurs in both fullscreen and windows
- Changing output, glshader does not seem to affect issue.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Using default DOSBox Staging 0.81.1 config

### Your configuration

```ini
https://gist.github.com/FatBarry/5eb96740971a3e075de1cb86cb66ca3c
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

codengine:
Can confirm that this happens since 0.81.1 (probably 0.81.0 as well) and in latest dev build, it does not occur in 0.80.1.
For me the easiest way to reproduce the issue is just to walk to the right in the first room and alternating between walking left and right. I can also confirm that if you hold an arrow key pressed there is a delay until the next action is invoked.

codengine:
By the way if one gets stuck in the intro, just use LOADFIX. It doesn't happen all the time for some strange reason.

FeralChild64:
If I see correctly, this game used to work correctly in 0.81.0 - but on the current `main` there is a problem like described in the ticket.

interloper98:
Maybe your improvements @FeralChild64 are OK, because Gods apparently doesn't handle fast typematic keystrokes on real hardware:

https://www.vogons.org/viewtopic.php?t=76407

[...]

PR #3748, opened 2024-06-08
Callback code generation rework

FeralChild64:
# Description

I intend to change the code of one callback, so that it behaves more like the real BIOS. Unfortunately, the current code generator is rather hard to tamper with, one needs to constantly watch the offset from the start of the code and adapt it accordingly for every instruction after the inserted/deleted code.

This rework is intended to make the generator code a little bit easier to maintain. There is a potential for regressions (despite I have self-reviewed the code twice) - but it should be easy to withdraw specific parts of the rework to find out the root cause.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3691 (rework related to the issue; PR does not fix the issue)

# Manual testing

Tried to run CtMouse 2.0, CtMouse 2.1, DOS Navigator, several games (Warcraft II, Settlers II, Colonization, etc.) - no problems found.

# Checklist

I have:

### DIFF
--- a/src/cpu/callback.cpp
+++ b/src/cpu/callback.cpp
@@ -158,19 +158,18 @@ const char* CALLBACK_GetDescription(callback_number_t cb_num)
 static uint8_t callback_setup_extra(const callback_number_t callback_number,
                                     const uint8_t callback_type,
                                     const PhysPt start_address,
-                                    const bool should_use_callback = true)
+                                    const bool use_callback = true)
 {
 	if (callback_number >= CB_MAX) {
-		E_Exit("Unknown callback number 0x%04x", callback_number);
+		LOG_ERR("CPU: Unknown callback number 0x%04x", callback_number);
 		return 0;
 	}
 
-	// TODO: temporary, until rework is complete
-	auto physAddress = start_address;
-	auto cb_num      = callback_number;
-	auto use_cb      = should_use_callback;
-
 	auto current_address = start_address;
+
+	auto advance = [&](const uint8_t bytes) {
+		current_address += bytes;
+	};
 
 	auto add_instruction_1 = [&](const uint8_t byte) {
 		phys_writeb(current_address, byte);
@@ -180,6 +179,24 @@ static uint8_t callback_setup_extra(const callback_number_t callback_number,
 	auto add_instruction_2 = [&](const uint8_t byte1, const uint8_t byte2) {
 		add_instruction_1(byte1);
 		add_instruction_1(byte2);
+	};
+
+	auto add_instruction_3 = [&](const uint8_t byte1,
+	                             const uint8_t byte2,
+	                             const uint8_t byte3) {
+		add_instruction_1(byte1);
+		add_instruction_1(byte2);
+		add_instruction_1(byte3);
+	};
+
+	auto add_instruction_4 = [&](const uint8_t byte1,
+	                             const uint8_t byte2,
+	                             const uint8_t byte3,
+	                             const uint8_t byte4) {
+		add_instruction_1(byte1);
+		add_instruction_1(byte2);
+		add_instruction_1(byte3);
+		add_instruction_1(byte4);
 	};
 
 	// Effectively 4-byte long - reflect it in the lambda name, to make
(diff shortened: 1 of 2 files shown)


johnnovak:
@FeralChild64 It's WIP I know, but you seem to be mixing huge non-functional reformats with actual code changes.

Could you separate the two? Ideally we'd have many separate commits for easy bisecting. Pure reformat commits can be big, that doesn't matter, but please don't mix in functional changes and refactors. Code refactors and changes/additions should be as many little self-contained changes as possible. The more, the better, so we can pinpoint what exactly introduced a regression.

Thanks 😄

src/cpu/callback.cpp
```diff
@@ -154,29 +155,54 @@ const char* CALLBACK_GetDescription(callback_number_t cb_num)
...
 
-callback_number_t CALLBACK_SetupExtra(callback_number_t cb_num, Bitu type, PhysPt physAddress, bool use_cb = true)
+static uint8_t callback_setup_extra(const callback_number_t callback_number,
+                                    const uint8_t callback_type,
+                                    const PhysPt start_address,
+                                    const bool should_use_callback = true)
 {
-	if (cb_num >= CB_MAX)
+	if (callback_number >= CB_MAX) {
+		E_Exit("Unknown callback number 0x%04x", callback_number);
```
johnnovak, line 164: Would it be better to just log an error and return without hard exiting? I really dislike hard exits.

src/cpu/callback.cpp
```diff
@@ -543,17 +569,19 @@ callback_number_t CALLBACK_SetupExtra(callback_number_t cb_num, Bitu type, PhysP
 			phys_writeb(physAddress+0x1B,(uint8_t)0xC3);	//A RETN Instruction
 		return (use_cb?32:27);
 	default:
-		E_Exit("CALLBACK:Setup:Illegal type %" sBitfs(u),type);
+		E_Exit("Unknown callback type 0x%04x", callback_type);
```
johnnovak, line 525: Would it be better to just log an error and return without hard exiting? I really dislike hard exits.

src/cpu/callback.cpp
```diff
@@ -154,29 +155,54 @@ const char* CALLBACK_GetDescription(callback_number_t cb_num)
 	return CallBack_Description[cb_num].c_str();
 }
 
-callback_number_t CALLBACK_SetupExtra(callback_number_t cb_num, Bitu type, PhysPt physAddress, bool use_cb = true)
+static uint8_t callback_setup_extra(const callback_number_t callback_number,
+                                    const uint8_t callback_type,
+                                    const PhysPt start_address,
+                                    const bool should_use_callback = true)
```
johnnovak, line 161: minor: just `use_callback`?

src/cpu/callback.cpp
```diff
@@ -154,29 +155,54 @@ const char* CALLBACK_GetDescription(callback_number_t cb_num)
...
 {
-	if (cb_num >= CB_MAX)
+	if (callback_number >= CB_MAX) {
+		E_Exit("Unknown callback number 0x%04x", callback_number);
 		return 0;
-	switch (type) {
+	}
+
+	// TODO: temporary, until rework is complete
+	auto physAddress = start_address;
```
johnnovak: `phys_address`

src/cpu/callback.cpp
```diff
@@ -195,379 +212,315 @@ static uint8_t callback_setup_extra(const callback_number_t callback_number,
...
-			phys_writeb(physAddress+0x01,(uint8_t)0x38);	//Extra Callback instruction
-			phys_writew(physAddress + 0x02, cb_num);        // The immediate word
-			physAddress += 4;
+		if (should_use_callback) {
+			add_native_call_4(callback_number);
 		}
-		phys_writeb(physAddress+0x00,(uint8_t)0xCA);		//A RETF 8 Instruction
-		phys_writew(physAddress+0x01,(uint16_t)0x0008);
-		return (use_cb?7:3);
+		add_instruction_3(0xca, 0x08, 0x00);       // retf 8
```
johnnovak, line 227: Oh man, manually returning and maintaining all these values... Why do people do such things to themselves... 😕

src/cpu/callback.cpp
```diff
@@ -195,379 +212,315 @@ static uint8_t callback_setup_extra(const callback_number_t callback_number,
...
+		if (should_use_callback) {
+			add_native_call_4(callback_number);
 		}
-		phys_writeb(physAddress+0x01,(uint8_t)0xCF);		//An IRET Instruction
-		for (uint8_t i = 0; i <= 0x0b; ++i)
-			phys_writeb(physAddress + 0x02 + i, 0x90);
-		phys_writew(physAddress + 0x0e, (uint16_t)0xedeb); // jmp callback
-		return (use_cb ? 0x10 : 0x0c);
+		add_instruction_1(0xcf);                   // iret
+		for (uint8_t i = 0; i <= 0x0b; ++i) {
```
johnnovak, line 393: minor `auto i`

johnnovak (approved):
Looks like a major maintainability improvement @FeralChild64 , very nice 🥇 I have only reviewed the generic principle behind this (which is good), and spot checked a few of the conversions—all look good. I guess if there's a typo somewhere, we'll find it out sooner or later, and it should be easy to verify the correctness if there's a regression.

src/cpu/callback.cpp
FeralChild64: I can't understand this neither. True, they didn't have lambdas back then - but passing a reference to current address variable would have improved this, too.

src/cpu/callback.cpp
FeralChild64: OK, changed.

src/cpu/callback.cpp
FeralChild64: OK, changed.

src/cpu/callback.cpp
FeralChild64: OK, changed - no more `E_Exit` in the reworked function.

src/cpu/callback.cpp
FeralChild64: OK, changed - no more `E_Exit` in the reworked function.

PR #3754, opened 2024-06-10
Config setting improvements and `CONFIG` command fixes

johnnovak:
# Description

## Fixes

- `CONFIG -h viewport` displayed no description because the `%` signs inside the help description confused `WriteOut()` which has printf-like behaviour. The change in behaviour was most likely unintentionally introduced by @japsmits config-related improvements (to be fair, this would be hard to spot as it only affects setting descriptions that include `%` signs in a very specific way).
- Therefore, `%` signs are now encoded as `%%` in all config descriptions, and the config writer handles them appropriately.
- Fixed `CONFIG -h SEC PROP` resulting in a crash if _PROP_ is non-existent (regression introduced in eee8e2c8050d4)
- `CONFIG -h SEC PROP` printed two error messages if both _SEC_ and _PROP_ are non-existent—now it only prints a single error.

## Cosmetic changes

I've shuffled the order of a few config settings around; the relevant commit messages should make everything self-explanatory.

I _almost_ reordered some sections too, but that's risky as the init functions are called in registration order. So changing the section order is just the ticket to potentially introduce subtle bugs here and there, or outright crashes when changing certain settings at runtime.

Reordering properties within sections, on the other hand, is safe.

I've also added newlines before and after the output of the `CONFIG -h setting_name` command:

image

# Manual testing

Started up Staging after each change, then ran a few games at the the end.

# Checklist

I have:


weirddan455 (approved):
Looks like a solid improvement :+1:

src/dosbox.cpp
```diff
@@ -1331,8 +1331,8 @@ void DOSBOX_Init()
 	control->AddSection_line("autoexec", &AUTOEXEC_Init);
 
 	MSG_Add("AUTOEXEC_CONFIGFILE_HELP",
-	        "Lines in this section will be run at startup.\n"
-	        "You can put your MOUNT lines here.\n");
+	        "Every line in this section is executed at startup as DOS commands.\n"
```
FeralChild64, line 1334: Maybe:

_Every line in this section is executed at startup as a DOS command._

or

_All the lines in this section are executed at startup as DOS commands._

src/gui/sdlmain.cpp
```diff
@@ -4213,14 +4213,65 @@ static void config_add_sdl()
...
+	const std::string default_output = "opengl";
+#else
+	const std::string default_output = "texture";
+#endif
+	pstring = sdl_sec->Add_string("output", always, default_output.c_str());
+
+	pstring->SetOptionHelp("opengl_default",
+	                       "Rendering backend to use for graphics output ('opengl' by default).\n"
+	                       "Only the 'opengl' backend has shader support and is thus the preferred option.\n"
+						   "The 'texture' backend is only provided as a last resort fallback for buggy or\n"
```
FeralChild64, line 4226: Indentation is wrong here.

src/dosbox.cpp
johnnovak: Nice catch, I'll take the first 😄

johnnovak:
@FeralChild64 @weirddan455 I've discovered some regressions when addressing the PR comments, so I've added a couple of more commits and I've extended the top description. Ready for another round of review, gents 😎

src/gui/sdlmain.cpp
```diff
@@ -4306,10 +4306,10 @@ static void config_add_sdl()
 	        "Set the transparency of the DOSBox Staging screen (0 by default).\n"
 	        "From 0 (no transparency) to 90 (high transparency).");
 
-	pstring = sdl_sec->Add_path("max_resolution", deprecated, "");
+	pstring = sdl_sec->Add_string("max_resolution", deprecated, "");
```
weirddan455, line 4309: This and `viewport_resolution` appear to be unused.  I would expect `deprecated` to do something.  Otherwise why not just remove it?  Are we missing a call to `SetDeprecatedWithAlternateValue` here?

The replacement config value `viewport` also has this incorrect path type by the way:

https://github.com/dosbox-staging/dosbox-staging/blob/3157a8c7f7604430e7e97b8715cbb22a0c84ff9f/src/gui/render.cpp#L1064

src/gui/render.cpp
```diff
@@ -1108,12 +1108,12 @@ static void init_render_settings(Section_prop& secprop)
 	        "                   (e.g., 960x720). The specified size must not be larger than\n"
 	        "                   the desktop. If it's larger than the window size, it's\n"
 	        "                   scaled to fit within the window.\n"
-	        "  N%:              Similar to 'WxH' but the size is specified as a percentage\n"
+	        "  N%%:              Similar to 'WxH' but the size is specified as a percentage\n"
 	        "                   of the desktop size.\n"
-	        "  relative H% V%:  The viewport is set to a 4:3 aspect ratio rectangle fit into\n"
+	        "  relative H%% V%%:  The viewport is set to a 4:3 aspect ratio rectangle fit into\n"
```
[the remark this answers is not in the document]
FeralChild64: Will this work with `blocksize`, `prebuffer`, and similar values, having `%s` in the config option help string?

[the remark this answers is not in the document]
weirddan455: Oh, maybe it's more complicated than I thought.  `GetHelp()` also parses the % signs.

https://github.com/dosbox-staging/dosbox-staging/blob/3157a8c7f7604430e7e97b8715cbb22a0c84ff9f/src/misc/setup.cpp#L304-L315

My suggested change is still better than what we have.  It avoids parsing the string a 2nd time for no reason.  But it might not solve the problem if it's happening inside `GetHelp()`.

If that's the case then yeah I guess we need the double %% sign.

src/gui/sdlmain.cpp
johnnovak: Try `config -h max_resolution`, or put `max_resolution` in your config, and you'll see why they need to stay.

We want to inform people about deprecated settings in their configs and give helpful tips on how to migrate to the new settings.

johnnovak: Also changing the type of `viewport` to string, good pickup 🎉

src/gui/render.cpp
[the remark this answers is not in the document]
johnnovak: Yeah we need the `%%` signs. I've added some explanatory comments to `GetHelp()`, plus I've ditched `replace_all` in favour of `format_str`. Keeping `replace_all`, though, as it's a useful string util function to have.

johnnovak:
Okay @weirddan455 & @FeralChild64, I've addressed the comments again and yet again made some further minor improvements in additional commits if you wanna take another look.

weirddan455 (approved):
Looks good

FeralChild64 (approved):
Looks OK, I have no more remarks.

johnnovak:
Thanks @FeralChild64 & @weirddan455 for the review. Another big one down 🎉

Issue #1446, opened 2021-12-18
Gravis Ultrasound can't coexist with Sound Blaster 16


FeralChild64:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

dosbox (dosbox-staging), version 0.78.0-alpha-1091-g91d3d200+

### What Operating System are you using?

Linux x86_64

### If Other OS, please describe

Gentoo

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

It seems it is not possible to use Gravis Ultrasound in DOSBox when Sound Blaster is enabled in the configuration too. In such case GUS initialization fails (screenshot) and games are unable to play MIDI files using GUS patches (non-midi digital sound usually still works).

I didn't have any similar problems with similar configuration on PCem, so I assume it is valid and cards should be able to work together.

![screenshot](https://user-images.githubusercontent.com/48332137/146653516-dbf2ee1f-b5d3-4bc6-a05e-a7f51a05aa5c.png)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Enable Gravis Ultrasound in config, but not the Sound Blaster
2. Install GUS drivers 4.11 (latest version I know of) for DOS (ULTRINIT.EXE is version 2.28)
3. Enable Sound Blaster 16 (tried some other SB models, probably all of them trigger the problem), restart DOSBox
4. Execute **C:\ULTRASND\ULTRINIT.EXE -dj**
5. Result: GUS initialization fails (see the screenshot). Setting the BLASTER variable to the value defined by GUS software (for it's Sound Blaster emulation) does not help. Initialization succeeds again if Sound Blaster is disabled.

### Your configuration

```ini
[cpu]
core      = dynamic
cputype   = pentium_slow
cycles    = 10000

[sblaster]
sbtype  = sb16
sbbase  = 220
irq     = 7
dma     = 1
hdma    = 5
sbmixer = true
oplmode = auto
oplemu  = default

[gus]
gus      = true
gusbase  = 240
gusirq   = 5
gusdma   = 3
ultradir = C:\ULTRASND

[midi]
mididevice = none

[innovation]
sidmodel   = none

[serial]
serial1    = disabled
serial2    = disabled
serial3    = disabled
serial4    = disabled

[ethernet]
ne2000     = false

[autoexec]
@MOUNT C ~/.config/dosbox/drive_test
[...]

PR #3755, opened 2024-06-10
Fix the GUS reset register's startup behavior

interloper98:
# Description

Fixes a regression in Soltys (1995) where the GUS was not starting up after the game auto-detects the GUS.

The fix involves always processing the three reset register bits (instead of using if/else logic), and starting the GUS running once the run bit is positive.

**_Edit:_** The change also goes a bit further and uses a `bit_view` to define the GUS Reset Register and uses it directly on read and write (which turns out to simplify the code quite a bit). Thanks for this tip, @johnnovak.

## Related issues

Caught by the  eXoDOS team @parricc, @SmilingSpectre, and @Python-Exoproject reported in https://github.com/exoscoriae/eXoDOS/issues/5213

# Manual testing

When [soltys.zip](https://github.com/user-attachments/files/15757169/soltys.zip) starts, you can test GUS:
- full auto-detect with 'Automatic'
- partial auto-detect (select 'GUS', then click all 'Auto' settings) 
- full manual (select 'GUS' followed by 240h base, IRQ 5, and DMA 3)

You can also select GUS + MIDI provided you select '330h' for the MIDI port.

Here are those results:

GUS full auto-detect:

https://github.com/dosbox-staging/dosbox-staging/assets/165201780/dfd1d4dc-44c0-4149-b42f-0f75ed895d99

GUS + MIDI (manual settings):

https://github.com/dosbox-staging/dosbox-staging/assets/165201780/26bfe388-3c74-4170-8eb9-d27f9be0faf7

Also tested GUS games like One Must Fall, Star Control 2, Blood, Descent, Duke 3D, Hocus Pocus, Rise of the Triad, lots of Pinball games, and Jazz Jackrabbit. Tested some well known GUS  demos like 1st Infection, Second Reality, and Unreal by Future Crew.

If there are other tests, let me know!

## Related documentation

gus-reset-register

https://www.infania.net/misc1/GUS/SOURCE%20CODE/UltraSound%20Lowlevel%20ToolKit%20v2.22%20(21%20December%201994).pdf

Attached incase source goes defunct:

[UltraSound Lowlevel ToolKit v2.22 (21 December 1994).pdf](https://github.com/user-attachments/files/15777966/UltraSound.Lowlevel.ToolKit.v2.22.21.December.1994.pdf)

# Checklist

I have:

### DIFF
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -254,7 +254,6 @@ private:
 
 	void ActivateVoices(uint8_t requested_voices);
 	void AudioCallback(uint16_t requested_frames);
-	void BeginPlayback() noexcept;
 	void CheckIrq();
 	void CheckVoiceIrq();
 	uint32_t GetDmaOffset() noexcept;
@@ -276,8 +275,10 @@ private:
 
 	const std::vector<AudioFrame>& RenderFrames(const int num_requested_frames);
 
+	void StartRunning() noexcept;
+	void StopAndReset() noexcept;
+
 	void RenderUpToNow();
-	void StopPlayback();
 	void UpdateDmaAddress(uint8_t new_address);
 	void UpdateWaveMsw(int32_t& addr) const noexcept;
 	void UpdateWaveLsw(int32_t& addr) const noexcept;
@@ -1088,7 +1089,7 @@ void Gus::PopulatePanScalars() noexcept
 	}
 }
 
-void Gus::BeginPlayback() noexcept
+void Gus::StartRunning() noexcept
 {
 	assert(reset_register.is_running);
 
@@ -1322,7 +1323,7 @@ void Gus::RegisterIoHandlers()
 	write_handlers.at(8).Install(0x20b + port_base, write_to, io_width_t::byte);
 }
 
-void Gus::StopPlayback()
+void Gus::StopAndReset() noexcept
 {
 	assert(!reset_register.is_running);
 
@@ -1566,7 +1567,7 @@ void Gus::WriteToRegister()
 		return;
 	case 0x4c: // Reset register
 		reset_register.data = static_cast<uint8_t>(register_data >> 8);
-		reset_register.is_running ? BeginPlayback() : StopPlayback();
+		reset_register.is_running ? StartRunning() : StopAndReset();
 		return;
 	default:
 		break;
@@ -1650,7 +1651,7 @@ void Gus::WriteToRegister()
 Gus::~Gus()
 {
 	LOG_MSG("GUS: Shutting down");
-	StopPlayback();
+	StopAndReset();
 
 	// Prevent discovery of the GUS via the environment
 	ClearEnvironment();
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -273,7 +273,6 @@ private:
 	uint16_t ReadFromPort(io_port_t port, io_width_t width);
 
 	void RegisterIoHandlers();
-	void Reset(uint8_t state);
 
 	const std::vector<AudioFrame>& RenderFrames(const int num_requested_frames);
 
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -30,7 +30,7 @@
 #include <vector>
 
 #include "autoexec.h"
-#include "bitops.h"
+#include "bit_view.h"
 #include "channel_names.h"
 #include "control.h"
 #include "dma.h"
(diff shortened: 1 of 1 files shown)


src/hardware/gus.cpp
```diff
@@ -1563,16 +1563,30 @@ void Gus::WriteToRegister()
...
-		} else if (active_voices) {
-			BeginPlayback();
 		}
+
+		// Bit 1 - Enable DAC output. DAC's will not run unless its set.
+		dac_enabled = bit::is(reset_register, bit::literals::b1);
+
+		// Bit 2 - Master IRQ enable. This bit MUST be set to get ANY of
+		//         the GF1-generated IRQs (wavetable, volume, etc).
+		irq_enabled = bit::is(reset_register, bit::literals::b2);
```
johnnovak, line 1589: Very nice @interloper98! Bonus points for including the GUS SDK reference 🎉 

For _even more bonus points_ 😆, consider using our quite nice and self-documenting `bit_view` template for the Reset Register. Check out `vga.h` for usage examples. Entirely optional, though—this can be merged as-is if you don't feel like spending more time on this.

src/hardware/gus.cpp
```diff
@@ -1563,16 +1563,30 @@ void Gus::WriteToRegister()
...
+		if (bit::cleared(reset_register, bit::literals::b0)) {
 			StopPlayback();
-		} else if (state == 1) {
+		} else {
 			PrepareForPlayback();
-		} else if (active_voices) {
-			BeginPlayback();
 		}
+
+		// Bit 1 - Enable DAC output. DAC's will not run unless its set.
```
johnnovak, line 1584: typo: "DAC", "it's set"

src/hardware/gus.cpp
interloper98: I'm in it for the bonus points, so following those notes :-) Thanks for the tips @johnnovak

johnnovak: > Looking at the code now, the reset register is just the tip of the iceberg that can benefit from it compared to lots of larger registers in the GUS.

Thanks for doing this, and yeah, that's right; `bit_view` is great for making all that low-level register-banging code self-documenting.

If you wanna convert everything in GUS to `bit_view` in a follow-up PR, that would be a most welcome change 😉 😉

src/hardware/gus.cpp
```diff
@@ -194,6 +194,25 @@ class Voice {
 static void GUS_TimerEvent(uint32_t t);
 static void GUS_DMA_Event(uint32_t val);
 
+// Reset Register (4Ch)
```
johnnovak, line 197: Minor typo in the commit message: "This replaces three standard alone variable"

Standalone?

src/hardware/gus.cpp
```diff
@@ -1552,30 +1566,8 @@ void Gus::WriteToRegister()
...
-		}
-
-		// Bit 1 - Enable DAC output. DAC's will not run unless its set.
-		dac_enabled = bit::is(reset_register, bit::literals::b1);
-
-		// Bit 2 - Master IRQ enable. This bit MUST be set to get ANY of
-		//         the GF1-generated IRQs (wavetable, volume, etc).
-		irq_enabled = bit::is(reset_register, bit::literals::b2);
-	}
+		reset_register.data = static_cast<uint8_t>(register_data >> 8);
```
johnnovak, line 1569: I think you can simply just do this, and that's preferable:

`reset_register = {register_data >> 8};`

That's how I'm doing it in `void SurroundProcessor::ControlWrite(const uint8_t val)` in `adlib_gold.cpp`.

src/hardware/gus.cpp
interloper98: oops! fixing.

src/hardware/gus.cpp
interloper98: Thanks; I might keep poking around! (probably far too much to bite off; I got lucky with this one and it worked out OK!)

src/hardware/gus.cpp
johnnovak: Thanks for trying anyway 😄

johnnovak (approved):
Thanks @interloper98, very nice work and a 🥇 for filling out the PR template properly 😎

interloper98:
> Thanks @interloper98, very nice work and a 🥇 for filling out the PR template properly 😎

Thanks for the review and feedback, @johnnovak!

johnnovak:
> @interloper98, maybe you'll want to check also
> 
>     * [Gravis Ultrasound can't coexist with Sound Blaster 16 #1446](https://github.com/dosbox-staging/dosbox-staging/issues/1446)

Very low priority as you can just turn the GUS/SB on/off at runtime. Probably your energy would be better spent on more pressing issues 😄

Issue #3595, opened 2024-04-12
Consider using the Speex resampler consistenly for every audio device


johnnovak:
1. Some audio devices use the resampler from the reSIDfp library. This creates an unfortunate dependency on reSID in these unrelated devices, plus we're already using the Speex resampler for the Sound Blaster, and there's generic resampling support in the mixer for any mixer channel.

    <img width="1190" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/1c172863-8889-414f-afc5-64480b4dd1a2">

2. The Speex resampler is probably more performant than all these other resamplers bundled with the libraries.

3. `TwoPassSincResampler` gives you an assertion failure when you set a target sample rate between about 16010 and 16050 Hz. With 16000 Hz or lower, you get no assertion but no sound either. Speex doesn't have such problems, and this conflicts with the variable sample rate enhancement. Affected sound devices:
    - Game Blaster
    - PS/1 Audio
    - Tandy 3-Voice

5. ~~Consider rendering OPL and MT-32 output at their original rates too, then use Speex to resample to the host sample rate. Rendering at the host sample rate instead of the device's native sample rate might introduce timing discrepancies and these bundled resamplers might be lower quality and not as performant as Speex.~~ **_This has been done for the MT-32 and OPL; the rest is outstanding._**

weirddan455:
I'll continue our discussion from #3629 here.  We probably want to add the flag `SDL_AUDIO_ALLOW_FREQUENCY_CHANGE` when we open the SDL audio device.

```cpp
int sdl_allow_flags = 0;

if (negotiate) {
	// We allow chunk size changes (if supported), but don't allow
	// frequency changes because we've only seen them go beyond
	// SDL's desired 48 Khz limit
#if defined(SDL_AUDIO_ALLOW_SAMPLES_CHANGE)
	sdl_allow_flags |= SDL_AUDIO_ALLOW_SAMPLES_CHANGE;
#endif
```

>  I'd really like disallow any SDL-side resampling whatsoever. Is that possible with some flags? I want it to work as per above, and I want to keep all resampling business in our hands.

I believe this is enough to avoid resampling for our use-case but I'd really have to check the SDL source code to be sure.  If we ever use the SDL audio streams, it's a bit different but I'd probably recommend staying away from those and continue using the lower-level callback like we currently are.

[...]

PR #1299, opened 2021-09-29
Fix audio issues introduced with push API

kklobe:
After extensive testing and feedback, the push API is being reverted to the old pull/callback API. The push API ended up just being a game of shifting the latency around to different parts of the mixer pipeline with no clear benefit and many issues.

There have been some minor improvements retained, namely the `negotiate` config option that allows the system audio driver to modify the `rate` and `blocksize` options for optimal performance.

There were also many thread-safety issues related variables in the`mixer` struct being accessed from the callback function, which is called from a separate thread managed by SDL, and the other mixer code.

PR #3756, opened 2024-06-10
Further mixer fixes and improvements mega PR #2

johnnovak:
# Description

- Fix mixer time stretching regression (see first commit for details). It's relatively hard to trigger, but a regression nonetheless.

- The LERP upsampler can now also downsample if required. This resampling mode was introduced specifically for the default `sb_filter = modern` mode, which emulates the old DOSBox linear interpolation "resampler."

- Previously, the upsampler did nothing if the channel rate exceeded the host rate. This resulted in lower-pitched audio, e.g., when a program uses the maximum native 45,454 Hz Sound Blaster rate and the host rate is 44,100 Hz. Or if you set a 22,050 Hz host rate and the SB rate is above that. Now that's fixed; in these scenarios, now we use the Speex resampler to do proper downsampling.

- Because of an oversight, all audio channels defaulted to the LERP upsampler. This generally worked out fine as long as your host rate was 44.1 kHz or 48 kHz. For example, the max GUS rate is 44.1 kHz, so if your host rate is 44.1 kHz, you could never hit the lower-pitched audio bug... unless you set a lower than 44.1k host rate, then you could 🥶 Anyway, the LERP upsampler is now fixed, but we also default to the Speex resampler on all channels as the default resampling method (the individual audio devices can override this, of course).

- Host sample rates can now be set continuously from 8,000 to 96,000 Hz. What's the bloody point of that, one might wonder? Well, many DOS devices have non-standard sample rates, e.g. the MT-32 is fixed 32,000 Hz, the OPL 49,716 Hz, the Sound Blaster up to 45,454 Hz, etc. By setting custom sample rates, you can render WAV files at an audio device's native rate without resampling. This might be useful for debugging (e.g., for _yours truly_). Sure, going up to 96 kHz is rather pointless, but FluidSynth supports it, so why not, plus we need an upper limit anyway. 96 kHz is as good as any.

- I've got rid of the `TwoPassSincResampler` in Game Blaster, PS/1 Audio, and Tandy 3-Voice, and I'm using our standard Speex resampler instead. This partially addresses https://github.com/dosbox-staging/dosbox-staging/issues/3595. See the commit messages for the details.

- The `nosound` option now works better because it renders the audio and progresses the "audio frames rendered" counters. Previously, programs that somehow timed themselves to these counters counters were completely stalled in "no sound" mode (e.g., FastTracker II). Additionally, now you can capture the audio to a WAV file even in "no sound" mode.

- `blocksize` can now be set to non-power-of-two values too. It seems SDL negotiates such buffer sizes on certain audio cards.

# Manual testing

Executed all the [audio emulation test cases](https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-emulation-tests), and the following additional tests:

- Start DOSBox normally without specifying `rate` in the config and confirm in the logs that 48 kHz host sample rate is being used. Run `PT.BAT` and `CM.BAT` to play the same music in PT4DOS (SB Pro) and CapaMod (GUS), respectively, to familiarise yourself how the music should sound when played back at normal speed.

- Set `rate = 22050` and confirm the music plays at normal speed in both programs, not at half-speed, like `main`. 

- Set `nosound = on` in the config and confirm that:
  - Both programs show the playback animations, and are not stalled like `main`.
  - Capturing the audio records the music at the correct speed.

- Confirm arbitrary rates can be set both (e.g., `rate = 33333`) both with `negotiate` enabled or disabled. Confirm audio capturing works at these arbitraty rates (e.g., load the WAV into Audacity and check the sample rate).

- Confirm `stype = sb16` works with:
  - `sb_filter = auto` and 8000, 22050, 44100, 48000, and 96000 rates.
  - `sb_filter = modern` and the same rates.
  - `sb_filter = off` and the same rates.

- Repeat the tests with `sbtype = sbpro2`.

- Confirm 8000 Hz and 96000 Hz sample rates work, and that reverb and chorus work fine at these rates too (the audio will sound muffled on 8000 Hz; that's normal).

- Confirm the below audio devices work fine at 8 Khz and 96 Khz (these devices render directly at the mixer rate without resampling at the mixer level).
  - CD Audio, Covox, Disney, Game Blaster, GUS, Innovation, PC Speaker (both modes), PS1 Audio, ReelMagic, Tandy

- Confirm the mixer is not reinitialised when changing any of these settings at runtime:
  - `chorus`, `compressor`, `crossfeed`, `reverb`

- Change the above settings at runtime and confirm the sound output is as expected.

# Checklist

I have:

### DIFF
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -41,8 +41,6 @@
 
 #include <Iir.h>
 
-typedef void (*MIXER_MixHandler)(uint8_t* sampdate, int len);
-
 // The mixer callback can accept a static function or a member function
 // using a std::bind. The callback typically requests enough frames to
 // fill one millisecond with of audio. For an audio channel running at
@@ -242,28 +240,28 @@ public:
 	float GetChorusLevel() const;
 
 	template <class Type, bool stereo, bool signeddata, bool nativeorder>
-	void AddSamples(const int len, const Type* data);
+	void AddSamples(const int num_frames, const Type* data);
 
-	void AddSamples_m8(const int len, const uint8_t* data);
-	void AddSamples_s8(const int len, const uint8_t* data);
-	void AddSamples_m8s(const int len, const int8_t* data);
-	void AddSamples_s8s(const int len, const int8_t* data);
-	void AddSamples_m16(const int len, const int16_t* data);
-	void AddSamples_s16(const int len, const int16_t* data);
-	void AddSamples_m16u(const int len, const uint16_t* data);
-	void AddSamples_s16u(const int len, const uint16_t* data);
-	void AddSamples_m32(const int len, const int32_t* data);
-	void AddSamples_s32(const int len, const int32_t* data);
-	void AddSamples_mfloat(const int len, const float* data);
-	void AddSamples_sfloat(const int len, const float* data);
-	void AddSamples_m16_nonnative(const int len, const int16_t* data);
-	void AddSamples_s16_nonnative(const int len, const int16_t* data);
-	void AddSamples_m16u_nonnative(const int len, const uint16_t* data);
-	void AddSamples_s16u_nonnative(const int len, const uint16_t* data);
-	void AddSamples_m32_nonnative(const int len, const int32_t* data);
-	void AddSamples_s32_nonnative(const int len, const int32_t* data);
+	void AddSamples_m8(const int num_frames, const uint8_t* data);
+	void AddSamples_s8(const int num_frames, const uint8_t* data);
+	void AddSamples_m8s(const int num_frames, const int8_t* data);
+	void AddSamples_s8s(const int num_frames, const int8_t* data);
+	void AddSamples_m16(const int num_frames, const int16_t* data);
+	void AddSamples_s16(const int num_frames, const int16_t* data);
+	void AddSamples_m16u(const int num_frames, const uint16_t* data);
+	void AddSamples_s16u(const int num_frames, const uint16_t* data);
+	void AddSamples_m32(const int num_frames, const int32_t* data);
+	void AddSamples_s32(const int num_frames, const int32_t* data);
+	void AddSamples_mfloat(const int num_frames, const float* data);
+	void AddSamples_sfloat(const int num_frames, const float* data);
+	void AddSamples_m16_nonnative(const int num_frames, const int16_t* data);
+	void AddSamples_s16_nonnative(const int num_frames, const int16_t* data);
+	void AddSamples_m16u_nonnative(const int num_frames, const uint16_t* data);
+	void AddSamples_s16u_nonnative(const int num_frames, const uint16_t* data);
+	void AddSamples_m32_nonnative(const int num_frames, const int32_t* data);
+	void AddSamples_s32_nonnative(const int num_frames, const int32_t* data);
 
-	void AddStretched(const int len, int16_t* data);
+	void AddStretched(const int num_frames, int16_t* data);
 
 	void FillUp();
 	void Enable(const bool should_enable);
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -359,7 +359,7 @@ private:
 	bool last_samples_were_stereo  = false;
 	bool last_samples_were_silence = true;
 
-	ResampleMethod resample_method = {};
+	ResampleMethod resample_method = ResampleMethod::Resample;
 	bool do_resample               = false;
 	bool do_zoh_upsample           = false;
 
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -250,16 +250,12 @@ public:
 	void AddSamples_s16(const int num_frames, const int16_t* data);
 	void AddSamples_m16u(const int num_frames, const uint16_t* data);
 	void AddSamples_s16u(const int num_frames, const uint16_t* data);
-	void AddSamples_m32(const int num_frames, const int32_t* data);
-	void AddSamples_s32(const int num_frames, const int32_t* data);
 	void AddSamples_mfloat(const int num_frames, const float* data);
 	void AddSamples_sfloat(const int num_frames, const float* data);
 	void AddSamples_m16_nonnative(const int num_frames, const int16_t* data);
 	void AddSamples_s16_nonnative(const int num_frames, const int16_t* data);
 	void AddSamples_m16u_nonnative(const int num_frames, const uint16_t* data);
 	void AddSamples_s16u_nonnative(const int num_frames, const uint16_t* data);
-	void AddSamples_m32_nonnative(const int num_frames, const int32_t* data);
-	void AddSamples_s32_nonnative(const int num_frames, const int32_t* data);
 
 	void AddStretched(const int num_frames, int16_t* data);
 
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -137,10 +137,11 @@ struct MixerChannelSettings {
 };
 
 enum class ResampleMethod {
-	// Use simple linear interpolation to resample from the channel sample
-	// rate to the mixer rate. This is the legacy behaviour, and it acts as a
-	// sort of a low-pass filter.
-	LinearInterpolation,
+	// If the channel sample rate is higher than the mixer sample rate,
+	// we'll do proper downsampling via Speex (e.g., when the mixer rate is
+	// 44,100 Hz but the Sound Blaster is running at its 45,454 Hz maximum
+	// rate, or the OPL channel at its native 49,716 Hz rate).
+	LerpUpsampleOrResample,
 
 	// Upsample from the channel sample rate to the zero-order-hold target
 	// frequency first (this is basically the "nearest-neighbour" equivalent
@@ -284,8 +285,8 @@ private:
 	AudioFrame ConvertNextFrame(const Type* data, const int pos);
 
 	template <class Type, bool stereo, bool signeddata, bool nativeorder>
-	void ConvertSamples(const Type* data, const int frames,
-	                    std::vector<float>& out);
+	void ConvertSamplesAndMaybeZohUpsample(const Type* data, const int frames,
+	                                       std::vector<float>& out);
 
 	void ConfigureResampler();
 	void ClearResampler();
@@ -360,8 +361,10 @@ private:
 	bool last_samples_were_silence = true;
 
 	ResampleMethod resample_method = ResampleMethod::Resample;
-	bool do_resample               = false;
-	bool do_zoh_upsample           = false;
+
+	bool do_lerp_upsample = false;
+	bool do_zoh_upsample  = false;
+	bool do_resample      = false;
 
 	struct {
 		float pos             = 0.0f;
--- a/src/hardware/disney.h
+++ b/src/hardware/disney.h
@@ -45,8 +45,8 @@ private:
 	void WriteControl(const io_port_t, const io_val_t value, const io_width_t);
 
 	// The DSS is an LPT DAC with a 16-level FIFO running at 7kHz
-	static constexpr auto dss_7khz_rate_hz = 7000;
-	static constexpr uint8_t max_fifo_size = 16;
+	static constexpr auto DisneySampleRateHz = 7000;
+	static constexpr auto MaxFifoSize        = 16;
 
 	// Managed objects
 	std::queue<uint8_t> fifo = {};
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -647,8 +647,8 @@ void MIXER_DeregisterChannel(MixerChannelPtr& channel_to_remove)
 
 	MIXER_LockAudioDevice();
 
-	auto it = mixer.channels.begin();
-	while (it != mixer.channels.end()) {
+	auto it = mixer.channels.cbegin();
+	while (it != mixer.channels.cend()) {
 		const auto [name, channel] = *it;
 
 		if (channel.get() == channel_to_remove.get()) {
@@ -1998,11 +1998,11 @@ void MixerChannel::AddSamples(const int num_frames, const Type* data)
 	if (do_lerp_upsample) {
 		auto& s = lerp_upsampler;
 
-		auto in_pos = mixer.temp_buf.begin();
+		auto in_pos = mixer.temp_buf.cbegin();
 		auto& out   = mixer.out_buf;
 		out.resize(0);
 
-		while (in_pos != mixer.temp_buf.end()) {
+		while (in_pos != mixer.temp_buf.cend()) {
 			AudioFrame curr_frame = {*in_pos, *(in_pos + 1)};
 
 			assert(s.pos >= 0.0f && s.pos <= 1.0f);
@@ -2074,9 +2074,9 @@ void MixerChannel::AddSamples(const int num_frames, const Type* data)
 	auto aux_reverb_pos   = mixer.aux_reverb.begin() + pos_offset;
 	auto aux_chorus_pos   = mixer.aux_chorus.begin() + pos_offset;
 
-	auto out_buf_pos = mixer.out_buf.begin();
+	auto out_buf_pos = mixer.out_buf.cbegin();
 
-	while (out_buf_pos != mixer.out_buf.end()) {
+	while (out_buf_pos != mixer.out_buf.cend()) {
 		AudioFrame frame = {*out_buf_pos++, *out_buf_pos++};
 
 		if (filters.highpass.state == FilterState::On) {
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -865,9 +865,7 @@ void MixerChannel::Enable(const bool should_enable)
 		prev_frame = {0.0f, 0.0f};
 		next_frame = {0.0f, 0.0f};
 
-		if (do_lerp_upsample || do_zoh_upsample || do_resample) {
-			ClearResampler();
-		}
+		ClearResampler();
 	}
 
 	is_enabled = should_enable;
@@ -1002,10 +1000,16 @@ void MixerChannel::ConfigureResampler()
 	}
 }
 
-// Clear the resampler and prime its input queue with zeros
+// Clear the resampler and prime its input queue with zeroes
 void MixerChannel::ClearResampler()
 {
-	auto init_resampler = [&] {
+	if (do_lerp_upsample) {
+		InitLerpUpsamplerState();
+	}
+	if (do_zoh_upsample) {
+		InitZohUpsamplerState();
+	}
+	if (do_resample) {
 		assert(speex_resampler.state);
 		speex_resampler_reset_mem(speex_resampler.state);
 		speex_resampler_skip_zeros(speex_resampler.state);
@@ -1015,32 +1019,6 @@ void MixerChannel::ClearResampler()
 		          name.c_str(),
 		          speex_resampler_get_input_latency(speex_resampler.state));
 #endif
-	};
-
-	switch (resample_method) {
-	case ResampleMethod::LerpUpsampleOrResample:
-		if (do_lerp_upsample) {
-			InitLerpUpsamplerState();
-		}
-		if (do_resample) {
-			init_resampler();
-		}
-		break;
-
-	case ResampleMethod::ZeroOrderHoldAndResample:
-		if (do_zoh_upsample) {
-			InitZohUpsamplerState();
-		}
-		if (do_resample) {
-			init_resampler();
-		}
-		break;
-
-	case ResampleMethod::Resample:
-		if (do_resample) {
-			init_resampler();
-		}
-		break;
 	}
 }
 
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -2158,7 +2158,6 @@ void MixerChannel::AddStretched(const int len, int16_t* data)
 		switch (resample_method) {
 		case ResampleMethod::LerpUpsampleOrResample:
 		case ResampleMethod::Resample:
-			assert(pos >= 0.0f && pos <= 1.0f);
 			out_sample = lerp(prev_sample, curr_sample, pos);
 			break;
 
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -3009,11 +3009,10 @@ static void init_mixer_dosbox_settings(Section_prop& sec_prop)
 	constexpr bool DefaultAllowNegotiate = true;
 #endif
 
-	constexpr auto Always      = Property::Changeable::Always;
 	constexpr auto WhenIdle    = Property::Changeable::WhenIdle;
 	constexpr auto OnlyAtStart = Property::Changeable::OnlyAtStart;
 
-	auto bool_prop = sec_prop.Add_bool("nosound", Always, false);
+	auto bool_prop = sec_prop.Add_bool("nosound", OnlyAtStart, false);
 	assert(bool_prop);
 	bool_prop->Set_help(
 	        "Enable silent mode (disabled by default).\n"
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -57,6 +57,8 @@ CHECK_NARROWING();
 
 // #define DEBUG_MIXER
 
+constexpr auto DefaultSampleRateHz = 48000;
+
 // Over how many milliseconds will we permit a signal to grow from
 // zero up to peak amplitude? (recommended 10 to 20ms)
 constexpr auto EnvelopeMaxExpansionOverMs = 15;
@@ -129,7 +131,7 @@ struct ReverbSettings {
 };
 
 struct ChorusSettings {
-	ChorusEngine chorus_engine = ChorusEngine(48000);
+	ChorusEngine chorus_engine = ChorusEngine(DefaultSampleRateHz);
 
 	ChorusPreset preset            = ChorusPreset::None;
 	float synthesizer_send_level   = 0.0f;
@@ -159,8 +161,8 @@ struct MixerSettings {
 	RingBuffer<AudioFrame, MixerBufferByteSize> aux_reverb = {};
 	RingBuffer<AudioFrame, MixerBufferByteSize> aux_chorus = {};
 
-	std::vector<float> resample_temp = {};
-	std::vector<float> resample_out  = {};
+	std::vector<float> temp_buf = {};
+	std::vector<float> out_buf  = {};
 
 	AudioFrame master_volume = {1.0f, 1.0f};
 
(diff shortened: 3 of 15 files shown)


johnnovak:
Read for review, but going through all the manual testing steps will take me probably a week. I'm intending to put together a sound test pack & wiki page as part of that exercise.

johnnovak:
Thanks @vladkosi, I'll investigate, maybe I accidentally broke the filter selection

Yes, the upper version is how `modern` should behave, with lots of aliasing. The lower version should only be obtainable with `sbfilter = auto` that emulated the SB16 brickwall filtering authentically.

johnnovak:
That was valuable feedback @vladkosi, thanks for the early testing. I'll need to adjust the new logic because the LERP path was a bit broken.

johnnovak:
> This should be `sdl_allow_flags |= SDL_AUDIO_ALLOW_FREQUENCY_CHANGE` so that both flags are set.

Thanks again, pushed up a fix now (plus to the linear resampling stuff, but I'm yet to test it in detail).

> The "invalid blocksize setting" warning is also a bit misleading. It probably shouldn't be there when SDL changes the blocksize.

`blocksize` just accepts a list of power of two enum values, so this is a generic warning message issued by the config system.  There are two levels of corrections here:

- The config system reverting to the default `blocksize` if you provide an invalid value (e.g., not a power of two value, or less than 64, or greater than 8192)
- The SDL negotiation mechanism giving you a different blocksize (e.g., you asked for 64, but the driver only supports 256 or larger buffers)

> ```
> MIXER: SDL changed the requested blocksize of 1024 to 441 frames
> CONFIG: Invalid 'blocksize' setting: '441', using '1024'
> MIXER: Negotiated 2-channel 44100 Hz audio with a buffer size of 441 sample frames
> ```

Interesting, I can't trigger the first `SDL changed the requested blocksize of 1024 to 441 frames` error, not even without the fix. And I don't see how you could end up with this error message 🤔 

Anyway, now things should work as intended.

johnnovak:
Very interesting @vladkosi. Well, you get 44.1 kHz sample rate, so the 441 sample frame buffer size is clearly related to that.

That's... super weird. I'm not aware of audio drivers using non power of two buffer size. What OS are you on, and what sound card are you using? If you're using on-board audio, what motherboard do you have?

johnnovak:
@vladkosi And it seems audio playback is fine for you with the super weird 441 and 480 buffer sizes?

Even the SDL doco states buffers should have power of two sizes, and yeah, obviously hardware buffers are power of two...
https://wiki.libsdl.org/SDL2/SDL_OpenAudioDevice

> desired->samples is the desired size of the audio buffer, in sample frames (with stereo output, two samples--left and right--would make a single sample frame). **This number should be a power of two**, and may be adjusted by the audio driver to a value more suitable for the hardware. **Good values seem to range between 512 and 8096 inclusive**, depending on the application and CPU speed. 

Although, 8096 is not a power of two number... another LOL 🤦🏻

johnnovak:
> > What OS are you on, and what sound card are you using?
> 
> Windows 10 and Topping USB DAC. But I get the same 441 value on my old ThinkPad T440s with integrated audio.
> 
> > Also, can you set a power of two buffer size with the latest `main` build with `negotiate = off`?
> 
> Yes, with too low value of 128 the audio is glitching out, with 8192 it's delayed. So it's working.
> 
> > I'm syncing back the negotiated 441 value to the blocksize config setting so it reflects actual reality but 441 is not part of allowed enum values, therefore a warning is logged.
> 
> I see, so the mixer is actually using 441 and not the reverted value as I wrongly assumed in my last comment.
> 
> > And it seems audio playback is fine for you with the super weird 441 and 480 buffer sizes?
> 
> Yes, the audio sounds fine.

Thanks for confirming. Well, I think it's an SDL bug, but to be honest I'm not willing to explore this further... The good thing is you can get the audio working without issues, just you get some weirdness in the logs. If more people start reporting this, I might have another look, but the high-level goal is working audio, and we're achieving that goal.

johnnovak:
> I agree, if it works it works. The only potential issue I can see is that you can't change the blocksize anymore, it's always set to the SDL negotiated value.

Fair point, and therefore I think I'll make another change so `negotiate` controls both the sample rate and buffer size negotiation. So you're in full manual mode with `negotiate = off` which is frankly a better model.

johnnovak:
@vladkosi Okay, so now `negotiate = on` negotiates both the block size and the sample rate. With off, we should be in full manual mode. If you would be able to test it with a few setting permutations and report back, that would be appreciated.

johnnovak:
> Everything seems to be working.
> 
> With `negotiate = off`, it behaves exactly like the latest `main`. Low blocksize setting stutters, high adds delay.

Great to hear, thanks for confirming 👍🏻 

> With `negotiate = on`, I always get the 441 value and the audio sounds fine. But I'm no longer getting the warning that SDL changed the size, this condition is always false: [dd514cf#diff-d63eec157cb639fc1c09f2ec383f7626052f59a82c4df058087a7a6c006c8e6bR2834](https://github.com/dosbox-staging/dosbox-staging/commit/dd514cf4ae833948fb8cd2f1c4a5708798e857b3#diff-d63eec157cb639fc1c09f2ec383f7626052f59a82c4df058087a7a6c006c8e6bR2834) `mixer.blocksize` has been set to the value of `obtained_blocksize` on line 2816.

Ah, that's an oversight. Pushed a fix for it. I just can't test `negotiate` easily as my machine accepts pretty much any sample rate and buffer setting (macOS). Will try on my Windows box a bit later too.

src/hardware/mixer.cpp
```diff
@@ -3042,8 +3053,11 @@ static void init_mixer_dosbox_settings(Section_prop& sec_prop)
 
 	bool_prop = sec_prop.Add_bool("negotiate", OnlyAtStart, DefaultAllowNegotiate);
 	bool_prop->Set_help(
-	        "Let the system audio driver negotiate possibly better sample rate and blocksize\n"
-	        "settings (%s by default).");
+	        "Negotiate a possibly better sample rate setting that does not require resampling\n"
+	        "the emulator's sound output, and an optimal block size (%s by default).\n"
+	        "Disable it if you want to force a specific block size and sample rate (e.g., to\n"
+	        "capture audio at an emulated audio device's native rate), but that might incur\n"
+	        "additional resampling at the host OS audio level.");
```
weirddan455, line 3092: "additional resampling at the OS level" isn't quite accurate.  If you turn sample rate negotiate off and SDL cannot send the requested sample rate to the OS, it will do resampling inside SDL.

IMO this should be avoided at all cost because SDL's resampler is quite primitive.  I've looked at that code when it seg faulted on me.  That bug has since been fixed but I still would avoid using it.

I don't see a benefit to ever have sample rate negotiation turned off.  We have a Speex re sampler than can handle any arbitrate sample rate and it does a better job than SDL so just use that.

The worst case happens when a user requests some weird sample rate that their hardware/OS doesn't support.  We resample to that weird rate, hand it to SDL, and it just re-samples it a 2nd time before passing it to the OS.

weirddan455:
>  Very interesting @vladkosi. Well, you get 44.1 kHz sample rate, so the 441 sample frame buffer size is clearly related to that.

That means it has a buffer size of 1ms worth of audio.  That isn't too unusual.  IIRC Pipewire on Linux does the same thing when you use its API directly (been a while since I played with that so my memory is a bit fuzzy).  Pipewire and Pulseaudio are also very flexible on what they will accept though so it might not have that same behavior when going through SDL.  I would have to test to be sure.

>  Well, I think it's an SDL bug

I don't think this is really a bug.  SDL is doing what we asked which is negotiate with the OS and give us back the value that works best on the currently running OS/hardware.  I don't know what's so special about a power of 2 buffer size anyway.

src/hardware/mixer.cpp
johnnovak: Okay, you're right @weirddan455. For some reason, it did not occur to me I can do that if the user set `rate` and the SDL-negotiated sample frame rate mismatch... 🤦🏻 That's some tunnel vision, eh? 😆

I kinda assumed the current SDL resampler is not as shit as it used to be, but yeah, it's best to take the resampling matter 100% out of SDL's hands, and it's easy to do so with Speex.

Thanks man! 😎 🤘🏻

johnnovak:
> That means it has a buffer size of 1ms worth of audio.

Yeah, I also concluded it's a buffer to hold 1ms worth of audio

> That isn't too unusual. IIRC Pipewire on Linux does the same thing when you use its API directly (been a while since I played with that so my memory is a bit fuzzy). Pipewire and Pulseaudio are also very flexible on what they will accept [...]

Good to know, and that was my guess that the host OS/audio subsystem/driver/whatever is lenient on the buffer sizes, at least on the _software side_.

> > Well, I think it's an SDL bug
> 
> I don't think this is really a bug. SDL is doing what we asked which is negotiate with the OS and give us back the value that works best on the currently running OS/hardware. I don't know what's so special about a power of 2 buffer size anyway.

Well, I'm thinking of hardware and my experience with pro audio devices that I happen to use. Those use [ASIO drivers](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output) on Windows. To quote from Wikipedia:

> ASIO bypasses the normal audio path from a user application through layers of intermediary Windows operating system software so that an application connects directly to the [sound card](https://en.wikipedia.org/wiki/Sound_card) hardware. Each layer that is bypassed means a reduction in latency (the delay between an application sending audio information and it being reproduced by the sound card, or input signals from the sound card being available to the application). In this way, ASIO offers a relatively simple way of accessing multiple audio inputs and outputs independently.

So basically you get direct access in your audio app to the hardware buffers, more or less. And ASIO buffer sizes are power-of-two, common settings are 32, 64, 128, 256, and 1024. I don't think 2048 is even a thing because the max buffer size maps to the max size of the ring buffer at the hardware level that holds the sample data for the DAC.

So, the TL;DR is that there must be a power-of-two ring buffer in the hardware _somewhere_ in pretty much all audio devices, pro or not, because that's how audio hardware that has a DAC works 😄 So non power of two block sizes accepted by APIs that sit above the hardware are software abstractions. Whatever Pipewire and co does, they must internally interface with a power of two buffer somewhere at the lowest level, then offer a convenient API abstraction over that and do additional interfacing with the hardware buffer inside Pipewire.

In short, that surprised me, because I thought all audio APIs expose the raw hardware buffer sizes. But clearly not.

weirddan455:
@johnnovak Interesting.  I didn't know how the hardware buffers work.

>  In short, that surprised me, because I thought all audio APIs expose the raw hardware buffer sizes. But clearly not.

Definitely not.  I'm not sure how Windows/Mac work but on Linux there's a whole sound server (Pipewire or Pulseaudio) sitting in userspace that sits on top of the kernel-level ALSA API (and who knows what abstractions the kernel provides).  Then SDL goes on top of all that.  Each layer could potentially provide its own buffer in software.

I just tested hard-coding random blocksizes, bypassing the config system.  Even with negotiate on, SDL just accepts whatever I give it.  A blocksize of 300 sounds fine.  A blocksize of 200 sounds very bad but is still accepted.  I'm using PulseAudio currently.

I might have mis-spoke about Pipewire wanting 1ms.  Where I'm remembering that number from is our own internal mixer callbacks (where CDROM audio sends its audio data for example).  We always request audio in 1ms chunks from emulated audio devices.

I'll test on Windows here in a bit because I'm curious.

johnnovak:
> @johnnovak Interesting. I didn't know how the hardware buffers work.

Well, digital is binary, so it "prefers" power of two sizes in, well, everything 😄 Think of RAM chip sizes, etc.

Then a ring buffer implementation becomes simple with `2^N` buffer sizes as you just do an AND operation with an `2^N - 1` bit flag to do the wraparound. Surely you could support non power of two sizes in hardware too, but why bother? It would add to the cost if you need more gates, etc. So it's rather pointless, `2^N` buffer sizes are just fine.

src/hardware/mixer.cpp
```diff
@@ -2973,9 +2973,12 @@ static void init_mixer_dosbox_settings(Section_prop& sec_prop)
...
 	assert(int_prop);
-	int_prop->Set_values(
-	        {"8000", "11025", "16000", "22050", "32000", "44100", "48000"});
-	int_prop->Set_help("Mixer sample rate in Hz (%s by default).");
+	int_prop->SetMinMax(8000, 96000);
+	int_prop->Set_help(
+	        "Sample rate of DOSBox's internal audio mixer in Hz (%s by default).\n"
+	        "Valid range is 8000 to 96000 Hz. With the default 'negotiate = on' setting,\n"
+	        "the actual selected sample rate depends on what the host OS and audio driver\n"
+	        "support. Use 'negotiate = off' to force arbitrary mixer sample rates.");
```
weirddan455, line 3067: As we discussed on Discord, I think it would be best to always allow SDL to negotiate the sample rate.  In practice, the setting does not matter on any recent OS.

I just tested disabling PulseAudio on my Linux system running Staging directly through ALSA.  Even then, ALSA has a plugin that resamples any arbitrary sample rate you give it to 48,000hz by default and SDL uses that rather than mixing itself.

I would recommend changing this help text to something like:

```
"Sample rate of DOSBox's internal audio mixer in Hz (%s by default).\n"
"Valid range is 8000 to 96000 Hz\n"
"The vast majority of consumer-grade sound hardware uses a native rate of 48,0000Hz.\n"
"Recommend leaving this as is unless you have good reason.\n"
"The OS is likely to resample anything else to 48,0000Hz anyway.\n"
```

This is the best we can do really.  Detecting native sample rate would be difficult because modern sound systems like PulseAudio allow the user to switch between sound devices at runtime (ex. You can switch between headphones and HDMI out while Staging is running and we're none the wiser.)

weirddan455:
Discussed on Discord but I did test on Windows and was able to confirm the same blocksize behavior that @vladkosi mentioned.  With `negotiate = on`, SDL seems to always request a blocksize equal to `samplerate / 1000`.  This value could differ depending on hardware as SDL also takes into account the hardware's device period, which it gets from Windows through a call to `IAudioClient_GetDevicePeriod`

Presumably, this is a sane value.  If we set `negotiate = off`, SDL will probably just maintain its own internal buffer and give it to Windows at the blocksize it wants anyway.  However, we do have `negotiate = off` set as the Windows default and `negotiate = on` as the default for other platforms.  This may have been a problem for some Windows users historically.  I am just unsure.

src/hardware/mixer.cpp
```diff
@@ -2767,77 +2767,83 @@ void MIXER_CloseAudioDevice()
...
-	spec.samples  = check_cast<uint16_t>(mixer.blocksize);
+	desired.channels = NumStereoChannels;
+	desired.format   = AUDIO_S16SYS;
+	desired.freq     = requested_sample_rate_hz;
+	desired.samples  = check_cast<uint16_t>(requested_blocksize_in_frames);
+
+	desired.callback = mixer_callback;
+	desired.userdata = nullptr;
 
 	int sdl_allow_flags = 0;
```
weirddan455, line 2776: Recommend setting `int sdl_allow_flags = SDL_AUDIO_ALLOW_FREQUENCY_CHANGE` here.

src/hardware/mixer.cpp
```diff
@@ -2767,77 +2767,83 @@ void MIXER_CloseAudioDevice()
...
-		sdl_allow_flags |= SDL_AUDIO_ALLOW_SAMPLES_CHANGE;
-#endif
+	if (allow_negotiate) {
+		// Allow negotiating the sample rate and the audio buffer size
+		// (hopefully to match the audio hardware's native rate to
+		// prevent resampling the final output, and an ideal blocksize
+		// that achieves stutter-free playback at a low latency).
+		//
+		sdl_allow_flags = SDL_AUDIO_ALLOW_SAMPLES_CHANGE |
+		                  SDL_AUDIO_ALLOW_FREQUENCY_CHANGE;
```
weirddan455, line 2785: And this becomes `sdl_allow_flags |= SDL_AUDIO_ALLOW_SAMPLES_CHANGE` like it was before this commit.

src/hardware/mixer.cpp
```diff
@@ -2767,77 +2767,83 @@ void MIXER_CloseAudioDevice()
...
-	if (obtained.channels != spec.channels) {
-		E_Exit("SDL gave us %d-channel output but we require %d channels",
-		       obtained.channels,
-		       spec.channels);
-	}
+	// Opening SDL audio device succeeded
+	const auto obtained_sample_rate_hz = obtained.freq;
+	const auto obtained_blocksize      = obtained.samples;
+
+	mixer.sample_rate_hz = obtained_sample_rate_hz;
```
weirddan455, line 2816: You're doing the correct thing here by setting the mixer to the obtained sample rate (the vast majority of the time, this will also be the same as the requested sample rate).

src/hardware/mixer.cpp
```diff
@@ -2767,77 +2767,83 @@ void MIXER_CloseAudioDevice()
...
-		LOG_WARNING("MIXER: SDL changed the requested blocksize of %d to %d frames",
-		            mixer.blocksize,
-		            obtained_blocksize);
+	// Did SDL negotiate a different blocksize?
+	if (obtained_blocksize != requested_blocksize_in_frames) {
+		LOG_WARNING(
+		        "MIXER: SDL negotiated the requested blocksize of "
+		        "%d to %d frames",
+		        requested_blocksize_in_frames,
+		        obtained_blocksize);
```
weirddan455, line 2840: This is likely to happen on most Windows systems if negotiate is turned on (but it is turned off on Windows by default).  Up to you if you want to leave this warning in or maybe demote it to a normal log message.

src/hardware/mixer.cpp
johnnovak: Thanks for the extra testing and looking into the Linux situation @weirddan455. I agree with you, and I came to similar conclusions. The revised description looks good too, I like it. Will make these changes today.

src/hardware/disney.h
```diff
@@ -45,8 +45,8 @@ class Disney final : public LptDac {
 	void WriteControl(const io_port_t, const io_val_t value, const io_width_t);
 
 	// The DSS is an LPT DAC with a 16-level FIFO running at 7kHz
-	static constexpr auto dss_7khz_rate_hz = 7000;
-	static constexpr uint8_t max_fifo_size = 16;
+	static constexpr auto SampleRateHz  = 7000;
```
weirddan455, line 48: Make this `DisneySampleRate` or something?  This could clash with another similar looking constant and get confusing.

weirddan455 (changes requested):
Looks good.  A few minor naming comments plus the SDL negotiate changes we discussed.

include/mixer.h
```diff
@@ -360,8 +361,10 @@ class MixerChannel {
 	bool last_samples_were_silence = true;
 
 	ResampleMethod resample_method = ResampleMethod::Resample;
-	bool do_resample               = false;
-	bool do_zoh_upsample           = false;
+
+	bool do_lerp_upsample = false;
```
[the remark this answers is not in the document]
johnnovak: It's used in multiple places.

src/hardware/mixer.cpp
```diff
@@ -1921,92 +1994,95 @@ void MixerChannel::AddSamples(const int frames, const Type* data)
...
-			auto& out   = mixer.resample_out;
-			out.resize(0);
+	if (do_lerp_upsample) {
+		auto& s = lerp_upsampler;
 
-			while (in_pos != mixer.resample_temp.end()) {
-				AudioFrame curr_frame = {*in_pos, *(in_pos + 1)};
+		auto in_pos = mixer.temp_buf.begin();
+		auto& out   = mixer.out_buf;
+		out.resize(0);
```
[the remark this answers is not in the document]
johnnovak: Well, `clear()` would literally clear the contents, which is pointless. `resize()` just sets the length, then `emplace_back` will replace the contents, and enlarge the vector if needed.

I'm not big on micro-optimising stuff, but this is audio code, and it's such a simple optimisation, so why not do it. But I'll put a comment above the `resize(0)` invocations.

include/mixer.h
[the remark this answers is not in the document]
weirddan455: Oh, this is a header file.  For some reason, I thought this was a local variable inside the body of a function.  Nevermind :smile:

src/hardware/mixer.cpp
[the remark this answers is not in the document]
weirddan455: They do exactly the same thing according to what's in `stl_vector.h` on my system, only `resize()` has an extra conditional (is the new value bigger or smaller?).  If anything, `clear()` should be faster.  If there were non-trivial objects in the vector, they would need to be destroyed in both cases (and can probably be optimized either way to just set the length if not).

```cpp
      void
      resize(size_type __new_size)
      {
	if (__new_size > size())
	  _M_default_append(__new_size - size());
	else if (__new_size < size())
	  _M_erase_at_end(this->_M_impl._M_start + __new_size);
      }
```

```cpp
      void
      clear() _GLIBCXX_NOEXCEPT
      { _M_erase_at_end(this->_M_impl._M_start); }
```

But I mean it's a very minor thing so I don't have a problem if you want to leave it.

src/hardware/mixer.cpp
```diff
@@ -1921,92 +1994,95 @@ void MixerChannel::AddSamples(const int frames, const Type* data)
...
 
-			auto in_pos = mixer.resample_temp.begin();
-			auto& out   = mixer.resample_out;
-			out.resize(0);
+	if (do_lerp_upsample) {
+		auto& s = lerp_upsampler;
 
-			while (in_pos != mixer.resample_temp.end()) {
-				AudioFrame curr_frame = {*in_pos, *(in_pos + 1)};
+		auto in_pos = mixer.temp_buf.begin();
```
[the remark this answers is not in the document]
johnnovak: Good point. You can't use `const` though, you need to use `cbegin/cend`. Changing where appropriate.

My own `RingBuffer` does not support constant iterators though, and adding support is a pain. Not ready to jump into that rabbit hole right now, but adding a TODO comment 😄

src/hardware/mixer.cpp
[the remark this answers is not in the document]
johnnovak: TBH, I would not _really_ trust one particular implementation. They might do the same thing, the might do very different things. But `resize()` has the potential to be a lot more optimal, but yeah, lazy implementations might just implement one function to handle both. 

I'm going by these descriptions, and I assume they adhere to the standard:

image

image

So clear can _potentially_ do a lot of extra work, including clearing. The best case scenario for `resize()` is that it just changes the size internal var of the container and that's it.

Then things get murky, and maybe STL does specialisations for primitive types, etc. Anyway, I prefer resize, but it's really minor, yeah 😆 

Ref:
- https://en.cppreference.com/w/cpp/container/vector/resize
- https://en.cppreference.com/w/cpp/container/vector/clear

src/hardware/mixer.cpp
```diff
@@ -1921,92 +1994,95 @@ void MixerChannel::AddSamples(const int frames, const Type* data)
...
-		case ResampleMethod::LinearInterpolation: {
-			auto& s = lerp_upsampler;
+	ConvertSamplesAndMaybeZohUpsample<Type, stereo, signeddata, nativeorder>(
+	        data, frames, convert_dest_buf);
 
-			auto in_pos = mixer.resample_temp.begin();
-			auto& out   = mixer.resample_out;
-			out.resize(0);
+	if (do_lerp_upsample) {
+		auto& s = lerp_upsampler;
```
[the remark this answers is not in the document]
johnnovak: ...done on purpose, to remove noise 😎  Sometimes I use single-letter aliases like this in small loops/functions to reduce clutter (replace `s` with the full name and tell me if you like it better 😆 ).

src/hardware/mixer.cpp
johnnovak: Yep, as discussed.

src/hardware/mixer.cpp
johnnovak: Yeah, normal message is better, cheers.

johnnovak: Actually, maybe do not auto-default to negotiate the buffer size on Windows either and see what happens. I don't think setting arbitrary buffer sizes can fail either starting from Windows 10.

src/hardware/disney.h
johnnovak: That's better. I had an uneasy feeling when I did this... but I repressed it 😎 🤘🏻

src/hardware/mixer.cpp
weirddan455: Negotiate is off by default on Windows already.  It's on by default on other platforms.  It's been like that for a long time.

kklobe: Yep, when I originally added the negotiate logic, it defaulted to on across the board. Issues came up on some Windows systems, which is when kcgen added the Win-specific defaults.

src/hardware/mixer.cpp
johnnovak: Actually, as discussed, this is not needed.

src/hardware/mixer.cpp
johnnovak: @kklobe Makes sense, but do you remember the specific use case? I could be some dude had problems on Win XP which had a very different audio API, much less lenient.

I'm inclined to "ask for forgiveness later"; turn `blocksize` negotiate off by default and see what happens. I assume noboby will run into an issue on Windows 10+.

weirddan455: @johnnovak You keep saying you're going to turn negotiate off by default when that's already the case on Windows.

johnnovak: @weirddan455 Indeed 😩 I find these OS specific details confusing as fuck. Okay, I'm getting fed up with all this, I'll just leave the defaults and `negotiate` alone for now 🤷🏻 It will just negotiate the block size, that's it. Maybe change it later if we need to.

kklobe: @johnnovak that was all done by kcgen; I did a quick search through the closed issues and didn't see anything specific. My original intent (https://github.com/dosbox-staging/dosbox-staging/pull/1299) was to wipe out OS-specific configs and let SDL interact with the audio driver for "ideal" sample rate and blocksize, but I suppose there are drivers out there that just didn't give back the correct settings for our callback performance requirements.

weirddan455: Yeah just leave the blocksize stuff alone for now IMO.  I'm planning to take a look at the mixer to see if I can offload most of the work to a separate thread and hopefully make this a non-issue.  We can re-evaluate then.

Ideally, we would be able to accept any reasonable blocksize but let's leave it alone until we fully understand the problem.

johnnovak: Sounds like a plan @weirddan455 👍🏻

johnnovak:
> Looks good. A few minor naming comments plus the SDL negotiate changes we discussed.

Thanks for the thorough review, man, appreciated. 🏅 

I believe I've addressed all the comments. I'll leave this open for a day or two more and I'll do some more testing and expand the [audio tests wiki page](https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-emulation-tests) in the meantime... until I run out of steam 😆 This grew a bit too big again in the end...

weirddan455 (approved):
Nice work.  I don't have any more comments.  Merge whenever you're ready.

johnnovak:
I've really run out of steam with the testing, so here goes nothing! 😎 But I think it's pretty solid now.

PR #3765, opened 2024-06-12
Use native file I/O in local and overlay drive

weirddan455:
# Description

This PR replaces the use of C library I/O (`fopen()` and friends) with native file I/O.  Windows uses Win32 API calls and everyone else uses POSIX.

I was also able to remove some hacky `ftell()` + `fseek()` nonsense that should not be needed since there's no library buffering.  The OS itself will still buffer but that should be fine.  All we care about is that a write followed by a read will return the new data.

I have some more work on the file I/O planned to better handle timestamps (which will depend on native file I/O) but I thought it would be best to put that in a separate PR and get this in first.

# Manual testing

Tested on Windows (both MSYS2/GCC and MSVC/Clang) and Linux.  Crystal Caves and Mortal Kombat saves work which had been problematic on Windows in the past.

Other games with save state tested:  Secret of Monkey Island 2, Quake, Doom, Space Quest 1

# Checklist

I have:

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -27,6 +27,7 @@
 
 #include "bit_view.h"
 #include "cross.h"
+#include "fs_utils.h"
 #include "mem.h"
 #include "support.h"
 
@@ -189,7 +190,7 @@ private:
 
 class localFile : public DOS_File {
 public:
-	localFile(const char* name, const std_fs::path& path, FILE* handle,
+	localFile(const char* name, const std_fs::path& path, const NativeFileHandle handle,
 	          const char* basedir, bool _read_only_medium);
 	localFile(const localFile&)            = delete; // prevent copying
 	localFile& operator=(const localFile&) = delete; // prevent assignment
@@ -199,7 +200,6 @@ public:
 	bool Close() override;
 	uint16_t GetInformation() override;
 	bool UpdateDateTimeFromHost() override;
-	void Flush();
 	bool IsOnReadOnlyMedium() const override { return read_only_medium; }
 	const char* GetBaseDir() const
 	{
@@ -209,21 +209,13 @@ public:
 	{
 		return path;
 	}
-	FILE* fhandle = nullptr; // todo handle this properly
+	NativeFileHandle file_handle = InvalidNativeFileHandle;
 private:
 	const std_fs::path path = {};
 	const char* basedir     = nullptr;
-	long stream_pos         = 0;
-
-	bool ftell_and_check();
-	void fseek_and_check(int whence);
-	bool fseek_to_and_check(long pos, int whence);
 
 	const bool read_only_medium = false;
 	bool set_archive_on_close   = false;
-
-	enum class LastAction : uint8_t { None, Read, Write };
-	LastAction last_action = LastAction::None;
 };
 
 /* The following variable can be lowered to free up some memory.
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -190,8 +190,9 @@ private:
 
 class localFile : public DOS_File {
 public:
-	localFile(const char* name, const std_fs::path& path, const NativeFileHandle handle,
-	          const char* basedir, bool _read_only_medium);
+	localFile(const char* name, const std_fs::path& path,
+	          const NativeFileHandle handle, const char* basedir,
+	          bool _read_only_medium);
 	localFile(const localFile&)            = delete; // prevent copying
 	localFile& operator=(const localFile&) = delete; // prevent assignment
 	bool Read(uint8_t* data, uint16_t* size) override;
@@ -210,6 +211,7 @@ public:
 		return path;
 	}
 	NativeFileHandle file_handle = InvalidNativeFileHandle;
+
 private:
 	const std_fs::path path = {};
 	const char* basedir     = nullptr;
--- a/include/fs_utils.h
+++ b/include/fs_utils.h
@@ -46,21 +46,14 @@
 
 constexpr int64_t NativeSeekFailed = -1;
 
-struct NativeIoResult
-{
+struct NativeIoResult {
 	int64_t num_bytes = 0;
-	bool error = false;
+	bool error        = false;
 };
 
-enum class NativeSeek
-{
-	Set,
-	Current,
-	End
-};
+enum class NativeSeek { Set, Current, End };
 
-struct DosDateTime
-{
+struct DosDateTime {
 	uint16_t date = 0;
 	uint16_t time = 0;
 };
@@ -207,17 +200,20 @@ uint16_t local_drive_get_attributes(const std_fs::path& path,
 uint16_t local_drive_set_attributes(const std_fs::path& path,
                                     const FatAttributeFlags attributes);
 
-
 // Native file I/O wrappers.
-// Currently only used by local drive and overlay drive but suitable for use elsewhere.
+// Currently only used by local drive and overlay drive but suitable for use
+// elsewhere.
 NativeFileHandle open_native_file(const std_fs::path& path, const bool write_access);
 NativeFileHandle create_native_file(const std_fs::path& path,
                                     const std::optional<FatAttributeFlags> attributes);
 
-NativeIoResult read_native_file(const NativeFileHandle handle, uint8_t *buffer, const int64_t num_bytes_requested);
-NativeIoResult write_native_file(const NativeFileHandle handle, const uint8_t *buffer, const int64_t num_bytes_requested);
+NativeIoResult read_native_file(const NativeFileHandle handle, uint8_t* buffer,
+                                const int64_t num_bytes_requested);
+NativeIoResult write_native_file(const NativeFileHandle handle, const uint8_t* buffer,
+                                 const int64_t num_bytes_requested);
 
-int64_t seek_native_file(const NativeFileHandle handle, const int64_t offset, const NativeSeek type);
+int64_t seek_native_file(const NativeFileHandle handle, const int64_t offset,
+                         const NativeSeek type);
 
 // Returns offset from beginning of file
 int64_t get_native_file_position(const NativeFileHandle handle);
--- a/include/fs_utils.h
+++ b/include/fs_utils.h
@@ -32,6 +32,39 @@
 
 #include "std_filesystem.h"
 
+#if defined(WIN32)
+	#include <windows.h>
+
+	using NativeFileHandle = HANDLE;
+	// Cannot be constexpr due to Win32 macro
+	#define InvalidNativeFileHandle INVALID_HANDLE_VALUE
+
+#else // Linux, macOS
+	using NativeFileHandle = int;
+	constexpr NativeFileHandle InvalidNativeFileHandle = -1;
+#endif
+
+constexpr int64_t NativeSeekFailed = -1;
+
+struct NativeIoResult
+{
+	int64_t num_bytes = 0;
+	bool error = false;
+};
+
+enum class NativeSeek
+{
+	Set,
+	Current,
+	End
+};
+
+struct DosDateTime
+{
+	uint16_t date = 0;
+	uint16_t time = 0;
+};
+
 // return the lines from the given text file or an empty optional
 std::optional<std::vector<std::string>> get_lines(const std_fs::path &text_file);
 
@@ -168,12 +201,32 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept;
 
 union FatAttributeFlags; // forward declaration
 
-FILE* local_drive_create_file(const std_fs::path& path,
-                              const FatAttributeFlags attributes);
 uint16_t local_drive_create_dir(const std_fs::path& path);
 uint16_t local_drive_get_attributes(const std_fs::path& path,
                                     FatAttributeFlags& attributes);
 uint16_t local_drive_set_attributes(const std_fs::path& path,
                                     const FatAttributeFlags attributes);
+
+
+// Native file I/O wrappers.
+// Currently only used by local drive and overlay drive but suitable for use elsewhere.
+NativeFileHandle open_native_file(const std_fs::path& path, const bool write_access);
+NativeFileHandle create_native_file(const std_fs::path& path,
+                                    const std::optional<FatAttributeFlags> attributes);
+
+NativeIoResult read_native_file(const NativeFileHandle handle, uint8_t *buffer, const int64_t num_bytes_requested);
+NativeIoResult write_native_file(const NativeFileHandle handle, const uint8_t *buffer, const int64_t num_bytes_requested);
+
+int64_t seek_native_file(const NativeFileHandle handle, const int64_t offset, const NativeSeek type);
+
+// Returns offset from beginning of file
+int64_t get_native_file_position(const NativeFileHandle handle);
+
+void close_native_file(const NativeFileHandle handle);
+
+// Sets the file size to be equal to the current file position
+bool truncate_native_file(const NativeFileHandle handle);
+
+DosDateTime get_dos_file_time(const NativeFileHandle handle);
 
 #endif
--- a/src/dos/drive_local.cpp
+++ b/src/dos/drive_local.cpp
@@ -82,9 +82,9 @@ bool localDrive::FileCreate(DOS_File** file, const char* name, FatAttributeFlags
 	const bool file_exists = FileExists(expanded_name);
 
 	attributes.archive = true;
-	FILE* file_pointer = local_drive_create_file(expanded_name, attributes);
+	NativeFileHandle file_handle = create_native_file(expanded_name, attributes);
 
-	if (!file_pointer) {
+	if (file_handle == InvalidNativeFileHandle) {
 		LOG_MSG("Warning: file creation failed: %s", expanded_name);
 		DOS_SetError(DOSERR_ACCESS_DENIED);
 		return false;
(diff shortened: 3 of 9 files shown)


include/dos_system.h
```diff
@@ -189,7 +190,7 @@ class DOS_Device : public DOS_File {
 
 class localFile : public DOS_File {
 public:
-	localFile(const char* name, const std_fs::path& path, FILE* handle,
+	localFile(const char* name, const std_fs::path& path, NativeFileHandle handle,
```
johnnovak, line 193: `const NativeFileHandle`?

include/fs_utils.h
```diff
@@ -32,6 +32,37 @@
...
 
+#if defined(WIN32)
+#include <windows.h>
+using NativeFileHandle = HANDLE;
+// Cannot be constexpr due to Win32 macro
+#define InvalidNativeFileHandle INVALID_HANDLE_VALUE
+#else
+using NativeFileHandle = int;
+constexpr NativeFileHandle InvalidNativeFileHandle = -1;
+#endif
```
johnnovak, line 43: Maybe format it like this to make it look marginally nicer:

```cpp
#if defined(WIN32)
	#include <windows.h>

	using NativeFileHandle = HANDLE;
	// Cannot be constexpr due to Win32 macro
	#define InvalidNativeFileHandle INVALID_HANDLE_VALUE

#else // Linux, macOS
	using NativeFileHandle = int;
	constexpr NativeFileHandle InvalidNativeFileHandle = -1;
#endif
```

include/fs_utils.h
```diff
@@ -32,6 +32,37 @@
...
+	Set,
+	Current,
+	End
+};
+
+struct DosDateTime
+{
+	uint16_t date = 0;
+	uint16_t time = 0;
+};
```
johnnovak, line 64: Just run the formatter on these. Then you'll get stuff like this:

```cpp
enum class NativeFileSeek { Set, Current, End };
```

include/fs_utils.h
```diff
@@ -168,12 +199,31 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept;
...
+NativeFileHandle local_drive_create_file(const std_fs::path& path,
+                                         const FatAttributeFlags attributes);
 uint16_t local_drive_create_dir(const std_fs::path& path);
 uint16_t local_drive_get_attributes(const std_fs::path& path,
                                     FatAttributeFlags& attributes);
 uint16_t local_drive_set_attributes(const std_fs::path& path,
                                     const FatAttributeFlags attributes);
 
+// Windows ReadFile and WriteFile take in a DWORD (32 bit unsigned) for number of bytes
+// We'll make that our limit until someone can give a good reason why they need to read/write more than 4GB (unlikely)
```
johnnovak, line 213: Indeed 😎

include/fs_utils.h
```diff
@@ -168,12 +199,31 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept;
...
+NativeFileReturn local_drive_read_file(const NativeFileHandle handle, uint8_t *buffer, const uint32_t bytes_requested);
+NativeFileReturn local_drive_write_file(const NativeFileHandle handle, const uint8_t *buffer, const uint32_t bytes_requested);
```
johnnovak, line 215: I guess we could even just use `int` for all these byte sizes because of the FAT16 limits if it's easier overall.

include/fs_utils.h
```diff
@@ -168,12 +199,31 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept;
...
+
+int32_t local_drive_seek_file(const NativeFileHandle handle, const int32_t offset, const NativeFileSeek type);
```
johnnovak, line 217: Just `int offset`? Or `int64`?

include/fs_utils.h
```diff
@@ -168,12 +199,31 @@ std::deque<std_fs::path> get_xdg_data_dirs() noexcept;
...
+
+// Returns offset from beginning of file
+int32_t local_drive_get_file_position(const NativeFileHandle handle);
```
johnnovak, line 220: `int64`?

include/fs_utils.h
```diff
@@ -32,6 +32,37 @@
...
+#else
+using NativeFileHandle = int;
+constexpr NativeFileHandle InvalidNativeFileHandle = -1;
+#endif
+
+constexpr int32_t NativeFileSeekFailed = -1;
+
+struct NativeFileReturn
+{
+	uint32_t bytes = 0;
```
johnnovak, line 49: `int64`?

src/misc/fs_utils_posix.cpp
```diff
@@ -354,4 +357,104 @@ uint16_t local_drive_set_attributes(const std_fs::path& path,
 	return status ? DOSERR_NONE : DOSERR_ACCESS_DENIED;
 }
 
+// POSIX does not guarantee to read or write all bytes requested at once.
+// On Linux, it will read the entire chunk in one call (assuming it's a normal file).
+// On Windows, it does not always do so (although we're not using these functions on Windows).
+// On Mac and the various BSDs, who knows. So be safe and do a loop.
```
johnnovak, line 363: Nice explanation 👍🏻 And yeah, this is the correct way to read a file on all platforms. Even Linux might change things up tomorrow.

src/misc/fs_utils_posix.cpp
```diff
@@ -354,4 +357,104 @@ uint16_t local_drive_set_attributes(const std_fs::path& path,
...
+NativeFileReturn local_drive_read_file(const NativeFileHandle handle, uint8_t *buffer, const uint32_t bytes_requested)
+{
+	NativeFileReturn ret = {};
+	ret.bytes = 0;
+	ret.error = false;
+	while (ret.bytes < bytes_requested) {
+		const auto bytes = read(handle, buffer + ret.bytes, bytes_requested - ret.bytes);
```
johnnovak, line 370: `num_bytes_read`?

src/misc/fs_utils_posix.cpp
```diff
@@ -354,4 +357,104 @@ uint16_t local_drive_set_attributes(const std_fs::path& path,
 	return status ? DOSERR_NONE : DOSERR_ACCESS_DENIED;
 }
 
+// POSIX does not guarantee to read or write all bytes requested at once.
+// On Linux, it will read the entire chunk in one call (assuming it's a normal file).
+// On Windows, it does not always do so (although we're not using these functions on Windows).
+// On Mac and the various BSDs, who knows. So be safe and do a loop.
+NativeFileReturn local_drive_read_file(const NativeFileHandle handle, uint8_t *buffer, const uint32_t bytes_requested)
```
johnnovak, line 364: `num_bytes_requested`?

src/misc/fs_utils_posix.cpp
```diff
@@ -354,4 +357,104 @@ uint16_t local_drive_set_attributes(const std_fs::path& path,
...
+		if (bytes <= 0) {
+			ret.error = bytes < 0;
+			break;
+		}
+		ret.bytes += bytes;
+	}
+	return ret;
+}
+
+NativeFileReturn local_drive_write_file(const NativeFileHandle handle, const uint8_t *buffer, const uint32_t bytes_requested)
```
johnnovak, line 380: `num_bytes_requested`?

src/misc/fs_utils_posix.cpp
```diff
@@ -354,4 +357,104 @@ uint16_t local_drive_set_attributes(const std_fs::path& path,
...
+{
+	NativeFileReturn ret = {};
+	ret.bytes = 0;
+	ret.error = false;
+	while (ret.bytes < bytes_requested) {
+		const auto bytes = write(handle, buffer + ret.bytes, bytes_requested - ret.bytes);
```
johnnovak, line 386: `num_bytes_written`?

src/misc/fs_utils_posix.cpp
```diff
@@ -354,4 +357,104 @@ uint16_t local_drive_set_attributes(const std_fs::path& path,
...
+		}
+		ret.bytes += bytes;
+	}
+	return ret;
+}
+
+int32_t local_drive_seek_file(const NativeFileHandle handle, const int32_t offset, const NativeFileSeek type)
+{
+	int posix_seek_type = SEEK_SET;
+	switch(type) {
```
johnnovak, line 399: Space after `switch`.

johnnovak: Well, just reformat the whole file; formatting is off.

include/fs_utils.h
```diff
@@ -32,6 +32,37 @@
...
+// Cannot be constexpr due to Win32 macro
+#define InvalidNativeFileHandle INVALID_HANDLE_VALUE
+#else
+using NativeFileHandle = int;
+constexpr NativeFileHandle InvalidNativeFileHandle = -1;
+#endif
+
+constexpr int32_t NativeFileSeekFailed = -1;
+
+struct NativeFileReturn
```
johnnovak, line 47: `NativeFileIoResult` would be better.

include/fs_utils.h
johnnovak: `num_bytes` would be better.

include/fs_utils.h
```diff
@@ -32,6 +32,37 @@
...
+{
+	uint32_t bytes = 0;
+	bool error = false;
+};
+
+enum class NativeFileSeek
```
johnnovak, line 53: `NativeSeek`?

shermp:
Is there a case for making these file i/o wrappers more generic so that other parts of the code could use them later. I'm talking about seek, read, write etc.

Other code that could use this would be the disk image code, in which case large file support would definitely be appreciated.

src/misc/fs_utils_posix.cpp
weirddan455: Linux is pretty strict about not breaking userspace so I doubt this behavior will change.  I suppose someone could mount their DOSBox drive on a networked filesystem or something though and then I'm not sure what the behavior is but the loop should work everywhere.

include/fs_utils.h
weirddan455: I suppose I can go up to `int64_t`.  I would need to write some additional code to break up large (greater than 4GB) reads/writes on Windows into multiple calls but that'll be easy enough.

weirddan455:
> Is there a case for making these file i/o wrappers more generic so that other parts of the code could use them later. I'm talking about seek, read, write etc.
> 
> Other code that could use this would be the disk image code, in which case large file support would definitely be appreciated.

They're already fairly generic and should be fine to use elsewhere.  Some of the calls (especially create and open) may need additional flags depending on the use-case but I think we can cross that bridge when we come to it.

Maybe I should rename the functions to `create_native_file` `read_native_file`, etc.  I'll bump the size types up to `int64_t` as @johnnovak suggested as well.

include/fs_utils.h
FeralChild64: @johnnovak And the maximum file size for FAT32 is still 4GB.

FeralChild64 (commented):
Looks OK. I have no remarks.

So far tried some Windows 3.1 software, Colonization game. and MkS_Vir virus scanner, didn't find any problems/

include/dos_system.h
weirddan455: fixed

include/fs_utils.h
weirddan455: fixed

include/fs_utils.h
weirddan455: formatted

include/fs_utils.h
weirddan455: Removed this comment.  It can handle large files now.

include/fs_utils.h
weirddan455: Changed to `int64_t`

include/fs_utils.h
weirddan455: Changed to `int64_t`

include/fs_utils.h
weirddan455: Changed to `int64_t`

include/fs_utils.h
weirddan455: Changed to `int64_t` and renamed

src/misc/fs_utils_posix.cpp
weirddan455: renamed

src/misc/fs_utils_posix.cpp
weirddan455: renamed

src/misc/fs_utils_posix.cpp
weirddan455: renamed

src/misc/fs_utils_posix.cpp
weirddan455: renamed

src/misc/fs_utils_posix.cpp
weirddan455: formatted

include/fs_utils.h
weirddan455: renamed

include/fs_utils.h
weirddan455: renamed

weirddan455:
Thanks for testing @FeralChild64 

I pushed changes addressing the comments above.  Mostly re-naming things and fixing formatting.  I also changed to `int64_t` everywhere in these native I/O functions.

They should be suitable for use where-ever else in the code they might be needed.  I renamed them as I said above to `open_native_file` and such (removing the `local_drive` prefix I had before).  I also made the fat attributes on `create_native_file` a `std::optional` as that one doesn't make sense if we want to create a normal file on the host.

src/dos/drive_local.cpp
```diff
@@ -697,65 +622,54 @@ bool localFile::Write(uint8_t *data, uint16_t *size)
...
 	// The inbound position is actually an int32_t being passed through a
 	// uint32_t* pointer (pos_addr), so reinterpret the underlying memory as
 	// such to prevent rollover into the unsigned range.
-	const auto pos = *reinterpret_cast<int32_t *>(pos_addr);
-	if (!fseek_to_and_check(pos, seektype)) {
+	const auto requested_pos = *reinterpret_cast<int32_t*>(pos_addr);
+
+	auto returned_pos = seek_native_file(file_handle, requested_pos, seek_type);
+
+	// TODO: Test Black Throne and see if this hack is still needed
```
johnnovak, line 652: Could you test this before merging?

johnnovak (approved):
This is solid step in the right direction @weirddan455 🎉 Looks very good now, left one last comment about testing Blackthorne if you have the capacity. Great work!

src/dos/drive_local.cpp
weirddan455: Tested.  It worked with this hack in and failed to launch with it removed.  I pushed a new commit with more accurate/less hacky behavior and tested this works on Linux and Windows.

I was side-eyeing that `reinterpret_cast` and it turned out to be incorrect for `SEEK_SET`.  The game seeks to 2777296400 (over 2GB) into the file.  Windows and POSIX both allow this.  It doesn't actually resize the file on seek but if you then write to the file, it will fill with zeros up to that size.

Anyway, we just need to interpret the value as unsigned if using `SEEK_SET`.  I put some references to MS-DOS Encyclopedia in the comments.

The error handling and set to 0 isn't needed by Blackthorne but may be needed by other DOS games doing weird things.  Apparently the DOS Seek interrupt never fails assuming you give it a valid handle and seek type.

weirddan455:
I'll let you take a look at my last commit if you like @johnnovak addressing the Blackthrone issue.

weirddan455:
Added a `LOG_WARNING` if the file seek fails.  I wasn't quite sure what to do in this case because apparently the real DOS interrupt never fails.  The most likely case is a game trying to seek before the beginning of the file so I'm just setting the file position to 0.

Blackthorn never hits this case.  Using unsigned integer lets it succeed.  I wanted this warning to be visible to end users rather than a debug log because it seems to be rare and I want users to put this in a bug report if my assumption proves to be incorrect.

src/dos/drive_local.cpp
johnnovak: Thanks for the detailed explanation @weirddan455. This seems to have been quite a rabbit hole, but I'm glad you got to the bottom of it 🎉

src/dos/drive_local.cpp
```diff
@@ -642,33 +642,45 @@ bool localFile::Seek(uint32_t *pos_addr, uint32_t type)
...
 
-	// The inbound position is actually an int32_t being passed through a
-	// uint32_t* pointer (pos_addr), so before we save the seeked position
-	// back into it we first ensure the current long stream_pos (which is a
-	// signed 64-bit on some platforms + OSes) can fit within the int32_t
-	// range before assigning it.
-	*reinterpret_cast<int32_t*>(pos_addr) = check_cast<int32_t>(returned_pos);
+	// The returned value is always positive.
+	// It can exceed 32-bit signed max (ex. Blackthorne)
+	*pos_addr = check_cast<uint32_t>(returned_pos);
```
johnnovak, line 683: Very nice coments, @weirddan455, that's medal-worthy 🎖️

One minor thing, if you'd include the book page numbers next to the citation, that would be helpful. Entirely optional, though, if you're sick of this PR already 😆

johnnovak (approved):
Outstanding work @weirddan455, I'm glad you've managed to remove the Blackthorne hack as well 🚀 I agree with your last note about logging a warning, that's the best we can do.

src/dos/drive_local.cpp
weirddan455: Sure, added the page numbers.

src/dos/drive_local.cpp
weirddan455: It didn't actually take me too long.  Luckily it was an easily reproducible error and fail to launch on Blackthorne.  Then I just added a log statement and saw it was being cast to a large negative seek value, where it should have always been positive.

src/dos/drive_local.cpp
johnnovak: Noice 😎

PR #3775, opened 2024-06-18
Import missing S3 Trio64 XGA compare and fill routines from DOSBox-X

### LINKED ISSUE #2767: S3 Trio64 not emulated well enough for Windows 3.1x Trio64 drivers

current GIT main (0.81.0-alpha) - e410259b43c3bf3fd61eb42d390d43007e5e1c81

Linux x86_64

_No response_

### Relevant hardware info

_No response_

Since the configuration file claims Staging emulates the S3 Trio64 card, we should IMHO have this issue open so that the problem is not forgotten.

There is at least one Windows 3.1x game - Critical Mass by Sean O'Connir - which exposes a Trio64 emulation problem if used with S3 Trio 32/64 drivers v1.70.04:

S3_Trio64_1 70 04

Observe the Alpha Squadron panel window on the right side of the screen - it's misses the ships, they are not drawn. 

Here is how the game should look like - screenshot taken using S3 Vision964 drivers v1.41B5:

S3_964_1 41B5

1. Install Windows 3.11 for Workgroups
2. Install S3 Trio64 driver v1.70.04, resolution 1024x768 64K SF
3. Start Crtitical Mass game demo, press OK, then BEGIN MISSION
4. Observe the Alpha Squadron window on the right side of the screen

https://archive.org/details/CriticalMass_1020

### Your configuration

```ini
[dosbox]
machine = svga_s3
vmemsize = 4
```

### Provide a Log

```text
Extracted errors from the log (full log gets really long really quickly):

2023-08-17 20:00:00.825 | XGA: Wrote to port 8104 with 3d700de, IO width=4
2023-08-17 20:00:00.825 | XGA: Wrote to port 810c with 3df00e8, IO width=4
2023-08-17 20:00:00.825 | XGA: Unhandled draw command 3
2023-08-17 20:00:00.825 | XGA: Wrote to port 8104 with 3d400e6, IO width=4
2023-08-17 20:00:00.825 | XGA: Wrote to port 810c with 3da00e6, IO width=4
2023-08-17 20:00:00.825 | XGA: Unhandled draw command 3
2023-08-17 20:00:00.825 | XGA: Wrote to port 810c with 3da00ec, IO width=4
2023-08-17 20:00:00.825 | XGA: Unhandled draw command 3
2023-08-17 20:00:00.825 | XGA: Wrote to port 8104 with 3d700df, IO width=4
2023-08-17 20:00:00.825 | XGA: Wrote to port 810c with 3d900e3, IO width=4
2023-08-17 20:00:00.825 | XGA: Unhandled draw command 3
```

- [X] Yes, I agree.

interloper98:
# Description

Imports the missing S3 Trio64 XGA compare and fill routines from DOSBox-X. 

Thanks for the through implementations, @joncampbell123, as well as the detailed comments in the code :+1: 

Followed the wiki on how to import changes:

https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages-for-patches-authored-by-someone-else

## Related issues

Fixes #2767

# Manual testing

Tested the Trio64 v1.42B22 and [v1.70.04](https://github.com/user-attachments/files/15878266/w3117004.zip) drivers with Microsoft Office, Critical Mass, and Wheel of Fortune.

Tested the 'S3ID' tool:

s3id-trio64

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_s3.cpp
+++ b/src/hardware/vga_s3.cpp
@@ -386,8 +386,12 @@ uint8_t SVGA_S3_ReadCRTC(io_port_t reg, io_width_t)
 	case 0x2e:	/* New Chip ID  (low byte of PCI device ID) */
 		return 0x11;	// Trio64
 	case 0x2f:	/* Revision */
-		return 0x00;	// Trio64 (exact value?)
-//		return 0x44;	// Trio64 V+
+		// The documention leaves this unspecified and says "This will
+		// change with each stepping". Therefore we max this out under
+		// the assumption that newer steppings fix(ed) issues in prior
+		// steppings, and we want to avoid drivers witholding features
+		// based on stepping bugs.
+		return 0xff;
 	case 0x30:	/* CR30 Chip ID/REV register */
 		return 0xe1;	// Trio+ dual byte
 	case 0x31:	/* CR31 Memory Configuration */
--- a/src/hardware/vga_s3.cpp
+++ b/src/hardware/vga_s3.cpp
@@ -335,6 +335,9 @@ void SVGA_S3_WriteCRTC(io_port_t reg, io_val_t value, io_width_t)
 				(3d4h index 18h). Bit 8 is in 3d4h index 7 bit 4 and bit 9 in 3d4h
 				index 9 bit 6.
 		*/
+	case 0x63: // Extended Control Register CR63
+		vga.s3.reg_63 = val;
+		break;
 	case 0x67:	/* Extended Miscellaneous Control 2 */
 		/*
 			0	VCLK PHS. VCLK Phase With Respect to DCLK. If clear VLKC is inverted
@@ -452,6 +455,8 @@ uint8_t SVGA_S3_ReadCRTC(io_port_t reg, io_width_t)
 		return (vga.s3.la_window >> 8);
 	case 0x5a:	/* Linear Address Window Position Low */
 		return (vga.s3.la_window & 0xff);
+	case 0x63: // Extended Control Register CR63
+		return vga.s3.reg_63;
 	case 0x5D:	/* Extended Horizontal Overflow */
 		return vga.s3.ex_hor_overflow;
 	case 0x5e:	/* Extended Vertical Overflow */
--- a/src/hardware/vga_xga.cpp
+++ b/src/hardware/vga_xga.cpp
@@ -898,9 +898,6 @@ void XGA_BlitRect(Bitu val) {
 					LOG_MSG("XGA: DrawPattern: Shouldn't be able to get here!");
 					break;
 			}
-
-			bool doit = true;
-
 			// For more information, see the "S3 Vision864 Graphics
 			// Accelerator" datasheet
 			//
@@ -912,21 +909,19 @@ void XGA_BlitRect(Bitu val) {
 			// Register (COLOR_CMP)" which this code holds as
 			// xga.color_compare.
 
-			if (xga.control1 & 0x100) {
-				// COLOR_CMP enabled. control corresponds to XGA
-				// register BEE8h.
+			// Always update if we're not comparing (COLOR_CMP is
+			// bit 8). Otherwise, either update if the SRC_NE bit is
+			// set with a matching colour or vice-versa (SRC_NE not
+			// set with non-matching colour).
 
-				// control1 bit 7 is SRC_NE. If clear, don't
-				// update if source value == COLOR_CMP. If set,
-				// don't update if source value != COLOR_CMP
-				doit = !!(((srcval == colorcmpdata) ? 0 : 1) ^
-				          ((xga.control1 >> 7u) & 1u));
-			}
+			using namespace bit::literals;
 
-			if (doit) {
-				Bitu destval = GetMixResult(mixmode, srcval, dstdata);
-				//LOG_MSG("XGA: DrawPattern: Mixmode: %x Mixselect: %x", mixmode, mixselect);
+			if (bit::cleared(xga.control1, b8) ||
+			    bit::is(xga.control1, b7) == (srcval == colorcmpdata)) {
 
+				const auto destval = GetMixResult(mixmode, srcval, dstdata);
+
+				// LOG_MSG("XGA: DrawPattern: Mixmode: %x Mixselect: %x", mixmode, mixselect);
 				XGA_DrawPoint((Bitu)tarx, (Bitu)tary, destval);
 			}
 
--- a/src/hardware/vga_xga.cpp
+++ b/src/hardware/vga_xga.cpp
@@ -187,17 +187,14 @@ void XGA_DrawPoint(Bitu x, Bitu y, Bitu c) {
 
 }
 
-Bitu XGA_PointMask() {
-	switch(XGA_COLOR_MODE) {
-		case M_LIN8:
-			return 0xFFul;
-		case M_LIN15:
-		case M_LIN16:
-			return 0xFFFFul;
-		case M_LIN32:
-			return 0xFFFFFFFFul;
-		default:
-			break;
+static uint32_t get_point_mask()
+{
+	switch (XGA_COLOR_MODE) {
+	case M_LIN8: return UINT8_MAX;
+	case M_LIN15:
+	case M_LIN16: return UINT16_MAX;
+	case M_LIN32: return UINT32_MAX;
+	default: break;
 	}
 	return 0;
 }
@@ -840,7 +837,7 @@ void XGA_BlitRect(Bitu val) {
 	if(((val >> 5) & 0x01) != 0) dx = 1;
 	if(((val >> 7) & 0x01) != 0) dy = 1;
 
-	colorcmpdata = xga.color_compare & XGA_PointMask();
+	colorcmpdata = xga.color_compare & get_point_mask();
 
 	Bitu mixselect = (xga.pix_cntl >> 6) & 0x3;
 	uint32_t mixmode = 0x67; /* Source is bitmap data, mix mode is src */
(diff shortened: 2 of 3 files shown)


src/hardware/vga_xga.cpp
```diff
@@ -882,10 +902,25 @@ void XGA_BlitRect(Bitu val) {
 					break;
 			}
 
-			const Bitu destval = GetMixResult(mixmode, srcval, dstdata);
-			//LOG_MSG("XGA: DrawPattern: Mixmode: %x Mixselect: %x", mixmode, mixselect);
+			bool doit = true;
```
johnnovak: Can we find a better name? 😆

src/hardware/vga_xga.cpp
```diff
@@ -882,10 +902,25 @@ void XGA_BlitRect(Bitu val) {
...
+
+			/* For more information, see the "S3 Vision864 Graphics Accelerator" datasheet
+			 * [http://hackipedia.org/browse.cgi/Computer/Platform/PC%2c%20IBM%20compatible/Video/VGA/SVGA/S3%20Graphics%2c%20Ltd/S3%20Vision864%20Graphics%20Accelerator%20(1994-10).pdf]
+			 * Page 203 for "Multifunction Control Miscellaneous Register (MULT_MISC)" which this code holds as xga.control1, and
+			 * Page 198 for "Color Compare Register (COLOR_CMP)" which this code holds as xga.color_compare. */
+			if (xga.control1 & 0x100) { /* COLOR_CMP enabled. control1 corresponds to XGA register BEE8h */
+				/* control1 bit 7 is SRC_NE.
+				 * If clear, don't update if source value == COLOR_CMP.
+				 * If set, don't update if source value != COLOR_CMP */
+				doit = !!(((srcval == colorcmpdata)?0:1)^((xga.control1>>7u)&1u));
```
johnnovak: Minor: we do `//` comments only in new code.

src/hardware/vga_xga.cpp
```diff
@@ -912,20 +909,24 @@ void XGA_BlitRect(Bitu val) {
...
+				using namespace bit::literals;
+				// Always update if we're not comparing (COLOR_CMP is bit 8)
+				if (bit::cleared(xga.control1, b8)) {
+					return true;
+				}
+				// If the SRC_NE bit (bit 7) is set ...
+				const auto src_ne_enabled = bit::is(xga.control1, b7);
+				const auto colours_match = (srcval == colorcmpdata);
+				// Then update on a match otherwise update if they differ.
+				return src_ne_enabled == colours_match;
```
johnnovak, line 928: Nice, the original was madness 😅

src/hardware/vga_xga.cpp
interloper98: Check the next commit that refactor it (I also wasn't a fan of `doit` :laughing: )

src/hardware/vga_s3.cpp
```diff
@@ -386,8 +389,12 @@ uint8_t SVGA_S3_ReadCRTC(io_port_t reg, io_width_t)
...
 		return 0x11;	// Trio64
 	case 0x2f:	/* Revision */
-		return 0x00;	// Trio64 (exact value?)
-//		return 0x44;	// Trio64 V+
+		// The documention leaves this unspecified and says "This will
+		// change with each stepping". Therefore we max this out under
+		// the assumption that newer steppings fix(ed) issues in prior
+		// steppings, and we want to avoid drivers witholding features
+		// based on stepping bugs.
+		return 0xff;
```
johnnovak, line 397: This is the only slightly risky thing in the whole PR, but we'll just need to test a few more titles and see how it goes.

src/hardware/vga_xga.cpp
interloper98: It took me a while to puzzle my way out of that one! Jenga for the mind :-)

interloper98:
> Very nice @interloper98, it wasn't too much code in the end. It looks like the changes are pretty much contained to the XGA code, making this a rather safe enhancement.
> 
> We should test it with the older drivers and a couple of other Windows 3.x games over a few days, though.

Sounds like a safe plan; I tried to compartmentalise the changes across commits, so in the worst case, we can isolate any problems pretty easily.

So far it's running great in win 3.1, and I'm not getting any more of those XGA-not-implemented console warnings.

johnnovak:
> Sounds like a safe plan; I tried to compartmentalise the changes across commits, so in the worst case, we can isolate any problems pretty easily.

Yeah, I've noticed! Very nice, exemplary work @interloper98 🎉

src/hardware/vga_xga.cpp
```diff
@@ -873,19 +894,48 @@ void XGA_BlitRect(Bitu val) {
...
+			};
+			if (should_mix_and_draw()) {
```
johnnovak, line 930: This works, but a slightly improved pattern for creating a const and initialising it with a lambda would be this. I started using this where appropriate.

```cpp
const auto stuff = [&] {
  ... calculate stuff
}();
```

src/hardware/vga_xga.cpp
johnnovak: Ah yeah, I reviewed it commit by commit and somehow missed the refactoring step.

src/hardware/vga_xga.cpp
joncampbell123: > This works, but a slightly improved pattern for creating a const and initialising it with a lambda would be this. I started using this where appropriate.
> 
> ```c++
> const auto stuff = [&] {
>   ... calculate stuff
> }(); // this calls it immediately
> ```

As far as I know a C++ lambda is a somewhat elaborate function pointer. Are you willing to make a function pointer call per pixel? Depending on architecture that might be a bit of a performance issue.

FeralChild64:
That was quick!

BTW, with your patch the Trio64 drivers [v1.70.04](https://github.com/user-attachments/files/15878266/w3117004.zip) seem to correctly handle the hacky `vmemsize = 8` setting. The previously recommended S3 Vision drivers (or the v1.42B22 Trio64 drivers) often display black square instead of mouse cursor with this setting.

src/hardware/vga_xga.cpp
interloper98: @johnnovak @joncampbell123 -- good points ; I'll aim to drop the lamda.

interloper98: Ok, simplified without the lamba and also avoided using temporary variables.

interloper98:
@Grounded0, thanks for the tips regarding WinG. Got that installed and tested. Here I'm running 800x600 256c at fixed 6500 cycles:

https://github.com/dosbox-staging/dosbox-staging/assets/165201780/1c8a8082-24c6-4626-b2d5-ba04396c4622

src/hardware/vga_xga.cpp
johnnovak: @joncampbell123 Yes, in the very general sense, if you pass lambas around, they become function pointers.

However, they can be trivially inlined in simple cases like this. More recent C++ features don't have to carry the legacy of C ("act as a high level assembler"), and thus have a lot more freedom to perform powerful optimisations.

This is a typical case when I use the lamba to express intent: run this simple code and assign the result to a const. I expect all compilers to inline the lambda. I'm 99% sure theydo, but I will check in godbolt later and report back.

Related:
https://stackoverflow.com/a/13722515

johnnovak:
> That was quick!
> 
> BTW, with your patch the Trio64 drivers [v1.70.04](https://github.com/user-attachments/files/15878266/w3117004.zip) seem to correctly handle the hacky `vmemsize = 8` setting. The previously recommended S3 Vision drivers (or the v1.42B22 Trio64 drivers) often display black square instead of mouse cursor with this setting.

According to this, only Trio 3D and ViRGE boards could handle 8 MB RAM:
http://www.dosdays.co.uk/topics/Manufacturers/s3.php

I assume that's why you call it hacky? I guess some VESA programs work fine on an 8 MB Trio 64, though. Especially DOS stuff. People sometimes soldered chips on top of each other, and if the hardware could address it, the software would work 😏

FeralChild64:
Precisely. I suppose S3 used same source tree (or at least shared components) for building drivers for different chipsets, so when they released the 8 MB capable ViRGE, they updated shared driver parts to handle that much VRAM.

If I remember correctly, the Build engine games can use the extra VRAM for tripple-buffering in high resolutions, that’s why the 8 MB option was added.

johnnovak:
> Precisely. I suppose S3 used same source tree (or at least shared components) for building drivers for different chipsets, so when they released the 8 MB capable ViRGE, they updated shared driver parts to handle that much VRAM.
> 
> If I remember correctly, the Build engine games can use the extra VRAM for tripple-buffering in high resolutions, that’s why the 8 MB option was added.

Seems like a worthwhile addition then, agreed.

FeralChild64:
There is still a risk that something less obvious is broken, though - we’ll see how it behaves in the long run. Anyway, my old idea (postponed to some less busy time) to merge the 2D ViRGE support now has even lower priority. If this driver works with 8 MB VRAM, I see no point - at least not for Windows 3.x.

rderooy:
In my testing with DOSBox-X, the official S3 Trio64(V+) Windows drivers will not offer additional resolutions or higher colour depths when you set VRAM to 8MB. The only I found to use more than 4MB was with VESA.

I'm guessing that either the drivers themselves are limited as to how much memory they support, depending on card type, or something is missing in the emulation that makes larger amounts of VRAM only work with VESA.

FeralChild64:
Windows 3.1x is limited to 4 MB VRAM by it's architecture, drivers won't use more.

johnnovak:
Well, and reality check, guys: what are you even planning to do in 1600x1200 in 24-bit true colour in Win 3.x? 🤣 That's like buying a Ferrari to take your kids to school in inner city peak traffic... 😎 🚗 Possibility != actual usefulnesd.

99% of Win3.x games require 640x480 with 256 colours, then a much smaller number benefits from 1024x768... if you have good eyesight, and a big monitor. Typically strategy games that use 256 colours. I don't think any high/true colour games need more than 640x480 or maybe 800x600.

Grounded0:
Colonization for Windows has a nice big viewport at 1600x1200@16-bit colour. 😁

src/hardware/vga_xga.cpp
johnnovak: Well, I don't have time to prove this is fine @interloper98, and it's a minor thing, so I won't let this hold up the PR process. However, I would bet good money on it it incurs zero performance penalty, so feel free to restore the lambda. Up to you.

johnnovak (approved):
Nice work @interloper98, feel free to merge when you're ready.

johnnovak:
> Colonization for Windows has a nice big viewport at 1600x1200@16-bit colour. 😁

Will keep that in mind for the time when I upgrade my glasses to something like this.

image

src/hardware/vga_xga.cpp
interloper98: It's even more straightforward in the current form with fewer intermediate variables, so not much to gain going back to the lamda.

johnnovak: Sounds good 😎

Issue #3504, opened 2024-03-12
Silmarils games with missing intro tone


Python-Exoproject:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

81.0

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

During the intro logo for the listed Silmaril games there is a tone played (OPL channel I think) that doesnt happen when using Staging 81.0

While the single tone itself isnt a big deal, the question is raised, if this tone isnt playing what else might not be playing due to whatever is causing the issue.

Its also odd in that Staging is the only fork that seems to have this issue, 0.74, ECE, Daum, and X all play the sound.

Of the other forks when comparing and investigating, X plays the tone best as there is some pitch change to the note, the other forks just play one constant sound.

The games are:

Transantartica (1993)
Ishar 2 (1993)
Ishar 1 (1992)
Bunny Bricks (1993)
Storm Master (1992)
Boston Bomb Club (1991)
Metal Mutant (1991)
Crystals of Arborea (1990)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Start game
2. Watch intro and listen for tone
3. Compare to X to hear the tone that should be there

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
An interesting problem and we should get to the bottom of it. Thanks for reporting @Python-Exoproject 🖖🏻

johnnovak:
I'm interested in investigating this, so I've assigned it to myself. However, if anyone else wants to go ahead with it, don't let that hold yourself back. I will probably get to this only in a few months' time.

johnnovak:
FYI, the _author_ (!) of Silmarils' DOS sound driver has made a very insightful comment on a related discussion ticket:
https://github.com/dosbox-staging/dosbox-staging/discussions/3467

Hopefully we'll get to the bottom of this soon with his help! 🚀

johnnovak:
Just merged the fix for this @Python-Exoproject. It will be available in the next dev build in the next 30 mins or so.

Python-Exoproject:
@interloper98 As per our discord discussion there was a bit more nuance to this that still needs fixing in Staging. 

[...]

PR #3785, opened 2024-06-21
Add carry, sign, zero, and parity flags to integer division instructions

interloper98:
# Description

The existing integer division instructions (unsigned `IDIV*()` and signed `DIV*()`)  already set the overflow flag, however they weren't setting the other relevant flags (carry, sign, zero, and parity) or resetting the other flags back to their undefined state.

Presumably the majority of DOS applications and games out there just grab the resulting quotient and don't care about the flag bits.

Silmarils games use an unsigned 16-bit division and its resulting flags to calculate an attenuation value passed to OPL channel 1 across roughly 100 video frames. Every frame, the calculation (typically) produces a larger attenuation value until it reaches `0x3f`, which is the OPL chip's max channel attenuation (or silence).

**Without the division flags**

The game only sets the initial attenuation (0% or max volume) and final attenuation value (100% or silent) leaving the 100-frame loop ineffective - so the tone blasts at max volume for the roughly 1.5 seconds.

```
7.211 | OPL: Attenuating volume of channel 1 by 0.00%

1.5 seconds elapse with channel 1 at full volume without the attenuation loop

8.623 | OPL: Attenuating volume of channel 1 by 100.00%
```

https://github.com/dosbox-staging/dosbox-staging/assets/165201780/ac758a6d-ec7d-4207-8552-cf7ca5c89ff6

---

**With the division flags**

The game calculates and applies increasing attenuation:

```
6.167 | OPL: Attenuating volume of channel 1 by 1.59%
6.181 | OPL: Attenuating volume of channel 1 by 1.59%
6.196 | OPL: Attenuating volume of channel 1 by 3.17%
6.209 | OPL: Attenuating volume of channel 1 by 4.76%
6.223 | OPL: Attenuating volume of channel 1 by 4.76%
6.238 | OPL: Attenuating volume of channel 1 by 6.35%
6.252 | OPL: Attenuating volume of channel 1 by 6.35%
6.267 | OPL: Attenuating volume of channel 1 by 7.94%
6.281 | OPL: Attenuating volume of channel 1 by 9.52%
6.295 | OPL: Attenuating volume of channel 1 by 9.52%
6.310 | OPL: Attenuating volume of channel 1 by 11.11%
6.323 | OPL: Attenuating volume of channel 1 by 12.70%
6.338 | OPL: Attenuating volume of channel 1 by 12.70%
6.352 | OPL: Attenuating volume of channel 1 by 14.29%
6.366 | OPL: Attenuating volume of channel 1 by 15.87%
6.380 | OPL: Attenuating volume of channel 1 by 15.87%
6.394 | OPL: Attenuating volume of channel 1 by 17.46%
6.409 | OPL: Attenuating volume of channel 1 by 17.46%
6.423 | OPL: Attenuating volume of channel 1 by 19.05%
6.437 | OPL: Attenuating volume of channel 1 by 20.63%
6.453 | OPL: Attenuating volume of channel 1 by 20.63%
6.466 | OPL: Attenuating volume of channel 1 by 22.22%
6.480 | OPL: Attenuating volume of channel 1 by 23.81%
6.495 | OPL: Attenuating volume of channel 1 by 23.81%
6.510 | OPL: Attenuating volume of channel 1 by 25.40%
6.523 | OPL: Attenuating volume of channel 1 by 26.98%
6.538 | OPL: Attenuating volume of channel 1 by 26.98%
6.551 | OPL: Attenuating volume of channel 1 by 28.57%
6.567 | OPL: Attenuating volume of channel 1 by 28.57%
6.580 | OPL: Attenuating volume of channel 1 by 30.16%
6.595 | OPL: Attenuating volume of channel 1 by 31.75%
6.608 | OPL: Attenuating volume of channel 1 by 31.75%
6.623 | OPL: Attenuating volume of channel 1 by 33.33%
6.638 | OPL: Attenuating volume of channel 1 by 34.92%
6.651 | OPL: Attenuating volume of channel 1 by 34.92%
6.666 | OPL: Attenuating volume of channel 1 by 36.51%
6.681 | OPL: Attenuating volume of channel 1 by 36.51%
6.694 | OPL: Attenuating volume of channel 1 by 38.10%
6.708 | OPL: Attenuating volume of channel 1 by 39.68%
6.724 | OPL: Attenuating volume of channel 1 by 39.68%
6.737 | OPL: Attenuating volume of channel 1 by 41.27%
6.752 | OPL: Attenuating volume of channel 1 by 42.86%
6.765 | OPL: Attenuating volume of channel 1 by 42.86%
6.780 | OPL: Attenuating volume of channel 1 by 44.44%
6.794 | OPL: Attenuating volume of channel 1 by 46.03%
6.808 | OPL: Attenuating volume of channel 1 by 46.03%
6.822 | OPL: Attenuating volume of channel 1 by 47.62%
6.837 | OPL: Attenuating volume of channel 1 by 47.62%
6.851 | OPL: Attenuating volume of channel 1 by 49.21%
6.866 | OPL: Attenuating volume of channel 1 by 50.79%
6.880 | OPL: Attenuating volume of channel 1 by 50.79%
6.895 | OPL: Attenuating volume of channel 1 by 52.38%
6.909 | OPL: Attenuating volume of channel 1 by 53.97%
6.922 | OPL: Attenuating volume of channel 1 by 53.97%
6.937 | OPL: Attenuating volume of channel 1 by 55.56%
6.951 | OPL: Attenuating volume of channel 1 by 57.14%
6.965 | OPL: Attenuating volume of channel 1 by 57.14%
6.980 | OPL: Attenuating volume of channel 1 by 58.73%
6.995 | OPL: Attenuating volume of channel 1 by 58.73%
7.009 | OPL: Attenuating volume of channel 1 by 60.32%
7.022 | OPL: Attenuating volume of channel 1 by 61.90%
7.037 | OPL: Attenuating volume of channel 1 by 61.90%
7.052 | OPL: Attenuating volume of channel 1 by 63.49%
7.065 | OPL: Attenuating volume of channel 1 by 65.08%
7.079 | OPL: Attenuating volume of channel 1 by 65.08%
7.094 | OPL: Attenuating volume of channel 1 by 66.67%
7.108 | OPL: Attenuating volume of channel 1 by 68.25%
7.122 | OPL: Attenuating volume of channel 1 by 68.25%
7.137 | OPL: Attenuating volume of channel 1 by 69.84%
7.151 | OPL: Attenuating volume of channel 1 by 69.84%
7.165 | OPL: Attenuating volume of channel 1 by 71.43%
7.179 | OPL: Attenuating volume of channel 1 by 73.02%
7.194 | OPL: Attenuating volume of channel 1 by 73.02%
7.208 | OPL: Attenuating volume of channel 1 by 74.60%
7.222 | OPL: Attenuating volume of channel 1 by 76.19%
7.236 | OPL: Attenuating volume of channel 1 by 76.19%
7.251 | OPL: Attenuating volume of channel 1 by 77.78%
7.265 | OPL: Attenuating volume of channel 1 by 79.37%
7.279 | OPL: Attenuating volume of channel 1 by 79.37%
7.293 | OPL: Attenuating volume of channel 1 by 80.95%
7.308 | OPL: Attenuating volume of channel 1 by 80.95%
7.322 | OPL: Attenuating volume of channel 1 by 82.54%
7.337 | OPL: Attenuating volume of channel 1 by 84.13%
7.350 | OPL: Attenuating volume of channel 1 by 84.13%
7.366 | OPL: Attenuating volume of channel 1 by 85.71%
7.379 | OPL: Attenuating volume of channel 1 by 87.30%
7.394 | OPL: Attenuating volume of channel 1 by 87.30%
7.407 | OPL: Attenuating volume of channel 1 by 88.89%
7.423 | OPL: Attenuating volume of channel 1 by 88.89%
7.437 | OPL: Attenuating volume of channel 1 by 90.48%
7.455 | OPL: Attenuating volume of channel 1 by 92.06%
7.465 | OPL: Attenuating volume of channel 1 by 92.06%
7.479 | OPL: Attenuating volume of channel 1 by 93.65%
7.493 | OPL: Attenuating volume of channel 1 by 95.24%
7.508 | OPL: Attenuating volume of channel 1 by 95.24%
7.522 | OPL: Attenuating volume of channel 1 by 96.83%
7.536 | OPL: Attenuating volume of channel 1 by 98.41%
7.551 | OPL: Attenuating volume of channel 1 by 98.41%
7.564 | OPL: Attenuating volume of channel 1 by 100.00%
7.579 | OPL: Attenuating volume of channel 1 by 100.00%
7.579 | OPL: Attenuating volume of channel 1 by 100.00%
```

https://github.com/dosbox-staging/dosbox-staging/assets/165201780/01d58c44-20b8-460e-9236-4bad029adbfb

To inspect the OPL channel attenuation, save this to `log_opl_channel_attenuation.patch` then `git apply log_opl_channel_attenuation.patch` in the root of DOSBox Staging source tree:

```diff
diff --git a/src/hardware/opl.cpp b/src/hardware/opl.cpp
index b640ff79b..328a96dea 100644
--- a/src/hardware/opl.cpp
+++ b/src/hardware/opl.cpp
@@ -315,6 +315,13 @@ void Opl::WriteReg(const io_port_t selected_reg, const uint8_t val)
 
 
        } else { // OPL
+               if (selected_reg >= 0x40 && selected_reg <= (0x40 + 18)) {
+                       const auto opl_channel = 1 + selected_reg - 0x40;
+                       const auto attenuation_percent = (val * 100.0) / 0x3f;
+                       LOG_MSG("OPL: Attenuating volume of channel %d by %.2f%%",
+                               opl_channel,
+                               attenuation_percent);
+               }
                OPL3_WriteRegBuffered(&opl.chip, selected_reg, val);
                if (selected_reg == 0x105) {
                        opl.newm = selected_reg & 0x01;
```

## Related issues

Finishing touches on https://github.com/dosbox-staging/dosbox-staging/issues/3504

# Manual testing

Tested plenty of real mode, protected mode, and various Windows 3.1 programs and did not notice any problems or differences.

I suppose this could impact performance a bit, but I didn't see any (despite doing more work on the host). 

# Checklist

I have:

### DIFF
--- a/include/regs.h
+++ b/include/regs.h
@@ -23,6 +23,8 @@
 #include "mem.h"
 #endif
 
+#include <cassert>
+
 // x86 CPU FLAGS register bit positions
 // - Intel iAXP 286 Programmer's Reference Manual
 // - Intel 64 and IA-32 Architectures Software Developer's Manual. Vol.1
@@ -133,11 +135,76 @@ constexpr auto BL_INDEX = 0;
 
 #endif
 
+// The CpuTestFlags struct allows CPU test flags to be applied en masse. First
+// construct it with the set of flags you will be adjusting followed by
+// assigning the actual flag state to the verbosely-named members. For example:
+//
+// CpuTestFlags my_flags(FLAG_CF | FLAG_OF);
+//   my_flags.has_carry = (did the operation need to carry results?);
+//   my_flags.has_overflow = (did the operation overflow?);
+//
+// Then apply the flags to the CPU registers:
+//   cpu_regs.ApplyTestFlags(my_flags);
+//
+struct CpuTestFlags {
+	constexpr CpuTestFlags() = delete;
+	constexpr CpuTestFlags(const uint32_t clear_mask);
+	constexpr uint32_t GetClearMask() const;
+	constexpr uint32_t GetSetMask() const;
+
+	bool has_carry        = false;
+	bool has_odd_parity   = false;
+	bool has_auxiliary    = false;
+	bool is_zero          = false;
+	bool is_sign_negative = false;
+	bool has_overflow     = false;
+
+private:
+	const uint32_t clear_mask = 0;
+};
+
+constexpr CpuTestFlags::CpuTestFlags(const uint32_t _clear_mask)
+        : clear_mask(_clear_mask)
+{
+	assert((clear_mask & ~FMASK_TEST) == 0 &&
+	       "Attempting to clear more than the test flags");
+}
+
+constexpr uint32_t CpuTestFlags::GetClearMask() const
+{
+	return clear_mask;
+}
+
+constexpr uint32_t CpuTestFlags::GetSetMask() const
+{
+	// clang-format off
+	const uint32_t set_mask = 
+		(has_carry        ? FLAG_CF : 0) |
+		(has_odd_parity   ? FLAG_PF : 0) |
+		(has_auxiliary    ? FLAG_AF : 0) |
+		(is_zero          ? FLAG_ZF : 0) |
+		(is_sign_negative ? FLAG_SF : 0) |
+		(has_overflow     ? FLAG_OF : 0);
+	// clang-format on
+
+	assert((set_mask & ~clear_mask) == 0 &&
+	       "Attempting to set flags that aren't cleared");
+	return set_mask;
+}
+
 struct CPU_Regs {
 	GenReg32 regs[8] = {};
-	GenReg32 ip = {};
-	uint32_t flags = 0;
+	GenReg32 ip      = {};
+	uint32_t flags   = 0;
+	constexpr void ApplyTestFlags(const CpuTestFlags& requested_flags);
 };
+
+constexpr void CPU_Regs::ApplyTestFlags(const CpuTestFlags& requested_flags)
+{
+	// First clear then set the requested flags
+	flags &= ~requested_flags.GetClearMask();
+	flags |= requested_flags.GetSetMask();
+}
 
 extern Segments Segs;
 extern CPU_Regs cpu_regs;
--- a/include/regs.h
+++ b/include/regs.h
@@ -23,26 +23,67 @@
 #include "mem.h"
 #endif
 
-#define FLAG_CF		0x00000001
-#define FLAG_PF		0x00000004
-#define FLAG_AF		0x00000010
-#define FLAG_ZF		0x00000040
-#define FLAG_SF		0x00000080
-#define FLAG_OF		0x00000800
+// x86 CPU FLAGS register bit positions
+// - Intel iAXP 286 Programmer's Reference Manual
+// - Intel 64 and IA-32 Architectures Software Developer's Manual. Vol.1
 
-#define FLAG_TF		0x00000100
-#define FLAG_IF		0x00000200
-#define FLAG_DF		0x00000400
+// Carry Flag (bit 0), 1 indicates an arithmetic carry or borrow has been
+// generated out of the most significant arithmetic logic unit (ALU) bit
+// position.
+constexpr uint32_t FLAG_CF = 1 << 0;
 
-#define FLAG_IOPL	0x00003000
-#define FLAG_NT		0x00004000
-#define FLAG_VM		0x00020000
-#define FLAG_AC		0x00040000
-#define FLAG_ID		0x00200000
+// Parity Flag (bit 2) indicates whether the modulo 2 sum of the low-order eight
+// bits of the result is even (PF=O) or odd (PF=1).
+constexpr uint32_t FLAG_PF = 1 << 2;
 
-#define FMASK_TEST		(FLAG_CF | FLAG_PF | FLAG_AF | FLAG_ZF | FLAG_SF | FLAG_OF)
-#define FMASK_NORMAL	(FMASK_TEST | FLAG_DF | FLAG_TF | FLAG_IF )	
-#define FMASK_ALL		(FMASK_NORMAL | FLAG_IOPL | FLAG_NT)
+// Auxiliary Carry Flag (bit 4), 1 indicates a carry from the lower nibble or a
+// for the lower nibble in BCD (Binary-coded Decimal) operations.
+constexpr uint32_t FLAG_AF = 1 << 4;
+
+// Zero Flag (bit 6), 1 indicates the result is zero.
+constexpr uint32_t FLAG_ZF = 1 << 6;
+
+// Sign Flag (bit 7), 1 indicates the result is negative.
+constexpr uint32_t FLAG_SF = 1 << 7;
+
+// Overflow Flag (bit 11), 1 indicates that the operation has overflowed; the
+// complete result was too large to be stored in the resulting register.
+constexpr uint32_t FLAG_OF = 1 << 11;
+
+// Trap flag (bit 8) 1 indicates that the processor is in single-step mode
+// (debugging).
+constexpr uint32_t FLAG_TF = 1 << 8;
+
+// I/O level flag (bit 9), 1 indicates that interrupts are enabled.
+constexpr uint32_t FLAG_IF = 1 << 9;
+
+// direction flag (bit 10), 1 indicates the direction is down. The meaning of
+// 'down' is in context to the instruction.
+constexpr uint32_t FLAG_DF = 1 << 10;
+
+// I/O privilege level flags (bits 12 and 13). 286+ only. This is all ones on
+// 8086 and 186.
+constexpr uint32_t FLAG_IOPL = (1 << 12) | (1 << 13);
+
+// Nested task flag (bit 14), 286+ only. This is always 1 on 8086 and 186.
+constexpr uint32_t FLAG_NT = 1 << 14;
+
+// Virtual 8086 mode flag (bit 17), 386+ only.
+constexpr uint32_t FLAG_VM = 1 << 17;
+
+// Alignment Check (bit 18), 486+-only.
+constexpr uint32_t FLAG_AC = 1 << 18;
+
+// CPUID instruction availability (bit 21), Pentium+-only.
+constexpr uint32_t FLAG_ID = 1 << 21;
+
+// Flag groups:
+constexpr uint32_t FMASK_TEST = FLAG_CF | FLAG_PF | FLAG_AF | FLAG_ZF |
+                                FLAG_SF | FLAG_OF;
+
+constexpr uint32_t FMASK_NORMAL = FMASK_TEST | FLAG_DF | FLAG_TF | FLAG_IF;
+
+constexpr uint32_t FMASK_ALL = FMASK_NORMAL | FLAG_IOPL | FLAG_NT;
 
 #define SETFLAGBIT(TYPE,TEST) if (TEST) reg_flags|=FLAG_ ## TYPE; else reg_flags&=~FLAG_ ## TYPE
 
--- a/src/cpu/flags.cpp
+++ b/src/cpu/flags.cpp
@@ -1054,7 +1054,7 @@ void set_cpu_test_flags_for_division(const T quotient) noexcept
 	SETFLAGBIT(OF, false);
 }
 
-// Explicit instantiations (replaces 12 instances)
+// Explicit instantiations (replaces 18 instances)
 template void set_cpu_test_flags_for_division<uint8_t>(const uint8_t) noexcept;
 template void set_cpu_test_flags_for_division<uint16_t>(const uint16_t) noexcept;
 template void set_cpu_test_flags_for_division<uint32_t>(const uint32_t) noexcept;
--- a/src/cpu/flags.cpp
+++ b/src/cpu/flags.cpp
@@ -1051,6 +1051,8 @@ void FillFlagsNoCFOF(void) {
 template <typename T>
 void set_cpu_test_flags_for_division(const T quotient) noexcept
 {
+	// Set the instruction type to division
+	lflags.type = t_DIV;
 
 	// Create a new test flags object with the flags we'll be settings
 	CpuTestFlags div_test_flags(FLAG_OF);
--- a/src/cpu/flags.cpp
+++ b/src/cpu/flags.cpp
@@ -1051,7 +1051,13 @@ void FillFlagsNoCFOF(void) {
 template <typename T>
 void set_cpu_test_flags_for_division(const T quotient) noexcept
 {
-	SETFLAGBIT(OF, false);
+
+	// Create a new test flags object with the flags we'll be settings
+	CpuTestFlags div_test_flags(FLAG_OF);
+
+	div_test_flags.has_overflow = false;
+
+	cpu_regs.ApplyTestFlags(div_test_flags);
 }
 
 // Explicit instantiations (replaces 18 instances)
--- a/src/cpu/flags.cpp
+++ b/src/cpu/flags.cpp
@@ -1046,7 +1046,13 @@ void FillFlagsNoCFOF(void) {
 	}
 	lflags.type=t_UNKNOWN;
 }
-   
+
+template <typename T>
+constexpr bool is_negative(const T value)
+{
+	return std::is_signed_v<T> && value < 0;
+}
+
 // Set CPU test flags for all IDIV and DIV quotients
 template <typename T>
 void set_cpu_test_flags_for_division(const T quotient) noexcept
@@ -1055,7 +1061,7 @@ void set_cpu_test_flags_for_division(const T quotient) noexcept
 	lflags.type = t_DIV;
 
 	// Create a new test flags object with the flags we'll be settings
-	CpuTestFlags div_test_flags(FLAG_CF | FLAG_OF);
+	CpuTestFlags div_test_flags(FLAG_CF | FLAG_OF | FLAG_SF);
 
 	// After performing IDIV, Intel's micro-operation clears the carry and
 	// overflow flags. Curiously, the 8086 documentation declares that the
@@ -1066,8 +1072,9 @@ void set_cpu_test_flags_for_division(const T quotient) noexcept
 	//
 	//_https://www.righto.com/2023/04/reverse-engineering-8086-divide-microcode.html
 	//
-	div_test_flags.has_carry    = false;
-	div_test_flags.has_overflow = false;
+	div_test_flags.has_carry        = false;
+	div_test_flags.has_overflow     = false;
+	div_test_flags.is_sign_negative = is_negative(quotient);
 
 	cpu_regs.ApplyTestFlags(div_test_flags);
 }
--- a/src/cpu/flags.cpp
+++ b/src/cpu/flags.cpp
@@ -1046,6 +1046,22 @@ void FillFlagsNoCFOF(void) {
 	}
 	lflags.type=t_UNKNOWN;
 }
+   
+// Set CPU test flags for all IDIV and DIV quotients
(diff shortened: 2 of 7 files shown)


weirddan455:
>  The existing integer division instructions (unsigned IDIV*() and signed DIV*()) already set the overflow flag, however they weren't setting the other relevant flags (carry, sign, zero, and parity) or resetting the other flags back to their undefined state.

I think this is incorrect, at least regarding the overflow flag.  A divide overflow is supposed to raise an exception.

Reference:  https://www.felixcloutier.com/x86/div

>  Overflow is indicated with the #DE (divide error) exception rather than with the CF flag.

Regarding flags:

>  The CF, OF, SF, ZF, AF, and PF flags are undefined.

I ran into this while writing some assembly testing my DOS file changes.  Here's a simple test case.  It can be assembled by `nasm hang.asm -o hang.com`  This program causes Staging to hang (not a full emulator hang but the shell becomes unresponsive).

```
org 100h

section .text
mov ax, 0ffffh
mov dx, 0ffffh
mov cx, 10
div cx

xor ax, ax
int 21h
```

In MS-DOS 6.22 the shell recognizes the overflow, displays an error, and resumes.  I was originally thinking this might just be a shell feature we don't have emulated but now I'm thinking this might be an inaccuracy with the CPU emulator.

overflow

interloper98:
@weirddan455 , yes - the current code will actually generate an exception on division overflow. 

If it makes it past the exception check, then there's no overflow so clears the OF flag (this is what the original code did too).

weirddan455:
>  yes - the existing/current code already generates an exception on division overflow (and that hasn't changed here).

Ok, I hadn't looked at the code yet.  I thought the issue I described might be related, but maybe it's not.  It looks like you've done your homework based on the comments in this PR.

>  Curiously, the 8086 documentation declares that the status flags are undefined following IDIV, but the microcode explicitly clears the carry and overflow flags.

So this game is depending on undefined CPU behavior, presumably.  Interesting.  Nice job finding this.

I really haven't worked on Dosbox's CPU emulator before so I'm maybe not the best person to review this but I'll give it a more detailed look later.

interloper98:
Thanks @weirddan455 !
Any additional sanity checks and validation would be a huge help.

Grounded0:
Here's what we use for performance profiling put together by myself:

https://anttipeltola.eu/misc/files/DOSBENCH.ZIP

interloper98:
Adjusted the commits to sequence changes one flag at a time.

The first commit starts at functional-equivalence (just moved the same code into a function), then adds more flag handling commit-by-commit so we can isolate any problems to the exact flag (if this causes problems down the road).

Thanks @Grounded0. I'll run these before-and-after.

Grounded0:
`DOSBENCH.ZIP` now updated with MMX Speed Test.

src/cpu/instructions.h
```diff
@@ -631,6 +631,23 @@
...
+template <typename T>
+constexpr void set_division_flags(const T quotient) noexcept
+{
+	// After performing IDIV, Intel's micro-operation clears the carry and
+	// overflow flags. Curiously, the 8086 documentation declares that the
+	// status flags are undefined following IDIV, but the microcode
+	// explicitly clears the carry and overflow flags. Signed DIV re-uses
+	// the above IDIV routine, so this is true for both signed and unsigned
+	// division [3].
+	SETFLAGBIT(OF, false);
```
johnnovak, line 651: Excellent explanatory comment 🏅

src/cpu/instructions.h
```diff
@@ -656,6 +656,10 @@ constexpr void set_division_flags(const T quotient) noexcept
 
 	// Zero flag (ZF) indicates a zero result when set [1], [2]
 	SETFLAGBIT(ZF, quotient == 0);
+
+	// Parity flag (PF) indicates whether the modulo 2 sum of the low-order
+	// eight bits of the operation is even (PF=O) or odd (PF=1) [1]
+	SETFLAGBIT(PF, parity_lookup[quotient & 0xff] != 0);
```
johnnovak, line 662: That's a helpful explanation for the next guy 👍🏻

johnnovak:
That's some standout work there @interloper98; an excellent demonstration of troubleshooting skills and the issue description is very well written 🚀 Quality work 💯 🧐 

I don't expect this to cause regressions, emulator-side performance problems, or compatibility issues because:

- You based your improvements on the actual reversed microcode, not on Intel's (likely incorrect) published documentation.
- Division is the absolutely slowest integer operation, so performance-conscious code avoided it like the plague. I expect division to be used rather sparingly in demos and games. Connsequently, I don't think this could cause measurable performance impact on the CPU emulation.
- As you've pointed out, 99.99% of DOS programs just don't care about these flags other than the overflow flag after a division. That makes sense to me. So we could even randomise those other flags and 99.99% of DOS programs would continue to work 😎 

I will give this a good testing later!

interloper98:
Three runs each:

| Benchmark | menu key  | result units | main  | PR  |
|---|---|---|---|---|
| 3D Bench for fast PC | 2 | FPS | 368.9, 366.1, 368.36  | 369.7, 367.8, 369.7  |
| Chris's 640x480 | 4 | FPS | 332.0, 334.8, 324.8| 338.6, 335.2, 331.8|
| PC Player 640x480 | 6 | FPS | 282.3, 281.7, 282.6 |282.9, 281.3, 282.5 |
| Doom fast PC | b| realticks | 274, 278, 275| 276, 277, 279 |
| Quake 640x480 | e| FPS | 144.9, 144.7, 146.0| 145.9, 146.9, 145.8 |
| PerfDOS | p | FPS | 760 to 763 | 760 to 763 |
| Linkpack optimised | q | Mflops| 94.55, 94.14, 94.56 |95.10, 94.86, 94.19 |

Pentium MMX (w) lots of metrics, so took screenshots. Only one run each (results looked identical).

main:



PR:

2024-06-21 17_46_23-DOSBox Status Window (Admin)

interloper98:
Thanks for the comments @johnnovak; 

> Division is the absolutely slowest integer operation, so performance-conscious code avoided it like the plague. I expect division to be used rather sparingly in demos and games.

The benchmarks results show that 100%! Cool; at least we're not weighing "correctness vs speed", in this case.

johnnovak:
> Thanks for the comments @johnnovak;
> 
> > Division is the absolutely slowest integer operation, so performance-conscious code avoided it like the plague. I expect division to be used rather sparingly in demos and games.
> 
> The benchmarks results show that 100%! Cool; at least we're not weighing "correctness vs speed", in this case.

Good to see my theory backed up by hard data. Thanks for those measurements @interloper98 💯

weirddan455:
The code looks good.  I think this should be a safe change, especially since you've researched how the real 8086 did it.

However, I'm pretty sure this will only affect the normal core.  I'm assuming the game you tested on is a real mode game and that's why this change worked.  The default core setting is `auto` which uses the normal core for real mode games and the dynamic core for protected mode games.

Those benchmarks you did are probably protected mode (I know for sure Doom and Quake are).  So they look identical because they are identical.  Not that I'm expecting setting a few flag bits to drastically affect performance though.

Have you tested this game with with `core = dynamic` to see if that is affected by the bug?  If so, you may also want to apply this fix to that core.  This doesn't have to be in this PR if you don't want to.  This fix is fine to go in as-is IMO.

For dyn_x86 see `src/cpu/core_dyn_x86/helpers.h`  -  This file contains only the division functions.  This is the default dynamic core used by x86 processors that is supposed to be highly optimized for the architecture.  I assume there is less translation that needs to happen since presumably some instructions can be near 1:1 map.

For the normal dynrec core see `src/cpu/core_dynrec/operators.h`  - Search for `dynrec_div_word`.  This is the core used by other architectures.  It can also be used on x86 but must be enabled at compile time by the Meson flag `-Ddynamic_core=dynrec` if you want to use it instead of dyn_x86.

interloper98:
Thanks for catching these holes, @weirddan455!
Ideally I can reuse the same code for all; will get these filled so it's working right on all the cores.

interloper98:
Thanks for the pointers into the dynrec and dynamic code paths, @weirddan455. Got those paths added now (wound up re-working the PR to use one function for all of them; wanted to avoid code duplication).

Tested each build type separately (`-Ddynamic_core=dyn-x86`, `-Ddynamic_core=dynrec`, `-Ddynamic_core=none`).

Re-checked performance a couple runs for each tests and the numbers are all in the same range as reported. (also checked normal an dnyrec cores and they're still on part w/ `main`).

Think its ready for review and any more testing the team can throw at it.

weirddan455:
Excellent work @interloper98 

I wasn't sure if you were going to be able to share code between the different cores but you figured it out.  Nice job cleaning up those flag constants and adding comments as well.

Again, this all looks fine to me on a surface level.  We probably do want to test on a non-x86 arch.  If you haven't already, @johnnovak said he was going to test on his ARM based Mac.  If everything looks good there, I'm happy.

>  Re-checked performance and the numbers are all in the same range as the earlier post. Also checked the normal an dnyrec cores, and they're still matching main. Like @johnnovak mentioned, I guess integer division isn't used heavily enough for this to hurt.

Yeah, wasn't expecting any major performance impact.  We're already paying the cost in these DIV instructions of calling back into C++ code.  The dynrec cores can't simply generate a DIV instruction on the host (like they do for other instructions) because DIV might raise an exception (divide by 0 or overflow) and that would crash the emulator.  So just setting a few extra variables while we're already in there isn't going hurt much.

interloper98:
`core = dynamic` is confirmed good on my pi 4 box (non-x86).

Here's the meson setup results, if that helps:

```
dosbox-staging 0.82.0-alpha

  Build Summary
    Build type            : release
    SIMD instruction sets : native, NEON
    Host page size (bytes): 4096

    zlib-ng provider      : built-in
    zlib provider         : built-in (zlib-ng)
    PNG provider          : built-in
    SDL_net 2.0 support   : true
    slirp support         : true
    FluidSynth support    : true
    mt32emu support       : true
    ManyMouse support     : True
    ALSA support          : true
    Byte order            : little-endian
    Per-page W^X          : True
    CPU dynamic core      : generic (dynrec)
    CPU core target arch  : ARMV8LE

  Subprojects
    iir                   : YES
    libpng                : YES 1 warnings
    mt32emu               : YES 1 warnings
    zlib-ng               : YES

Found ninja-1.11.1 at /usr/bin/ninja
```

and more info:

```
pi@raspberrypi:~/src/dosbox-staging $ uname -a                                                                                                      
Linux raspberrypi 6.6.31+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.31-1+rpt1 (2024-05-29) aarch64 GNU/Linux
```

include/regs.h
```diff
@@ -133,12 +133,45 @@ constexpr auto BL_INDEX = 0;
...
 
+struct CpuTestFlags {
+	bool HasCarry       = false;
+	bool IsOddParity    = false;
+	bool HasAuxiliary   = false;
+	bool IsZero         = false;
+	bool IsSignNegative = false;
+	bool HasOverflow    = false;
+	constexpr uint32_t AsValue() const;
+};
```
johnnovak, line 144: We use `lower_camel_case` for struct members.

src/cpu/flags.cpp
```diff
@@ -1051,7 +1051,11 @@ void FillFlagsNoCFOF(void) {
 template <typename T>
 void set_cpu_test_flags_for_division(const T quotient) noexcept
 {
-	SETFLAGBIT(OF, false);
+	const CpuTestFlags cpu_test_flags = {
+	        .HasOverflow = false,
+	};
```
johnnovak, line 1072: Won't this generate a warning that not all members have been initialised? If so, maybe don't use the designated initialiser like this.

src/cpu/flags.cpp
```diff
@@ -1053,8 +1053,18 @@ void set_cpu_test_flags_for_division(const T quotient) noexcept
...
+	// status flags are undefined following IDIV, but the microcode
+	// explicitly clears the carry and overflow flags. Signed DIV re-uses
+	// the above IDIV routine, so this is true for both signed and unsigned
+	// division.
+	//
+	//_https://www.righto.com/2023/04/reverse-engineering-8086-divide-microcode.html
+
 	const CpuTestFlags cpu_test_flags = {
+	        .HasCarry    = false,
 	        .HasOverflow = false,
```
johnnovak: Same comment; I think where we use designated initialisers, we can't to provide all values otherwise we'll get a warning.

src/cpu/flags.cpp
```diff
@@ -1064,8 +1064,9 @@ void set_cpu_test_flags_for_division(const T quotient) noexcept
 	//_https://www.righto.com/2023/04/reverse-engineering-8086-divide-microcode.html
 
 	const CpuTestFlags cpu_test_flags = {
-	        .HasCarry    = false,
-	        .HasOverflow = false,
+	        .HasCarry       = false,
+	        .IsSignNegative = std::is_signed_v<T> && quotient < 0,
+	        .HasOverflow    = false,
```
johnnovak, line 1071: Same comment 😄

include/regs.h
interloper98: thanks; fixed.

src/cpu/flags.cpp
interloper98: (I wasn't seeing warnings on my side or any new ones in CI) but I wound up reworking how this CpuTestFlags object works.

Previously all of the test flags were cleared (which was fine for the DIV instructions), but other instructions only clear and set a subset. So the new implementation lets you define the set of flags you'll be adjusting at construction.

src/cpu/flags.cpp
interloper98: Replaced with a slightly different approach.

src/cpu/flags.cpp
interloper98: Replaced with a slightly different approach.

interloper98:
Thanks for the review and testing @johnnovak :+1: 
Finished updates, did another round of testing, and rebased against main.
Ready for review.

include/regs.h
```diff
@@ -92,12 +135,77 @@ constexpr auto BL_INDEX = 0;
...
+// CpuTestFlags my_flags(FLAG_CF | FLAG_OF);
+//   my_flags.has_carry = (did the operation need to carry results?);
+//   my_flags.has_overflow = (did the operation overflow?);
+//
+// Then apply the flags to the CPU registers:
+//   cpu_regs.ApplyTestFlags(my_flags);
+//
+struct CpuTestFlags {
+	constexpr CpuTestFlags() = delete;
+	constexpr CpuTestFlags(const uint32_t _clear_mask);
```
johnnovak, line 151: Minor: you could call this `clear_mask` in the declaration.

src/cpu/flags.h
```diff
@@ -0,0 +1,36 @@
...
+// flags are handled. For an example of this pattern, see the division helper
+// functions in core_dyn_x86/helpers.h and the use of set_skipflags in
+// core_dyn_x86/decoder.h.
+
+
+// Set CPU test flags for all IDIV and DIV quotients
+template <typename T>
+void set_cpu_test_flags_for_division(const T quotient) noexcept;
+
+#endif
```
johnnovak, line 36: minor: newline at the end

johnnovak (approved):
Looks good @interloper98 👍🏻 I think we've tested this enough at this point, so probably it's best to get it merged then fix any potential issues later (but I don't expect any). Great work!

src/cpu/flags.h
interloper98: Added

include/regs.h
interloper98: Nice improvement to the API 👍
Fixed.

PR #3786, opened 2024-06-22
The great DOS_File/DOS_Drive Refactor (plus some timestamp improvements)

### LINKED ISSUE #620: New files have incorrect timestamp after changing date in dosbox-staging session
**Describe the bug**

Files created inside dosbox-staging DOS session have timestamp of OS, not timestamp of DOS session.

**To Reproduce**

    C:\>date 02-02-2011
    C:\>echo x > datefak
    C:\>dir datefak
    ...
    DATEFAK                       3 25-09-2020 12:56
    ...

**Expected behaviour**

DATEFAK should have timestamp indicating 2011, not 2020.

**Additional context**
This is long-standing issue (not a regression), somewhat related to lack of handling of DOS function 5701h (Set file time/date). After #227 will be merged in we'll have no excuse not to support this usecase as well :) That PR implements most of needed functionality to fix *this* issue.

weirddan455:
# Description

Several changes here.  Probably recommend reviewing by commit because I've got broken down decently well.

General overview:

- DOS_Files array moved from raw pointer to `unique_ptr`
- DOS_Drive moved from "owning" vectors of `unique_ptr` with lots of raw pointers everywhere to `shared_ptr`.
- Improvements to timestamp handling to better match MS-DOS behavior

## Related issues

Fixes #620

# Manual testing

Wrote test programs in x86 assembly and ran on MS-DOS 6.22 to confirm behavior of various combinations of DOS interrupts to see when timestamps get updated.  Opened multiple handles to the same file. Used the get date interrupt to verify when the timestamp gets updated after write, truncate, create, set date interrupt, etc.

Tested a few games with save files to verify they still work:  Crystal Caves, Doom, Secret of Monkey Island 2

Tested the case on #620 to verify that we now set the timestamp equal to Dosbox's BIOS rather than host time.

# Checklist

I have:

### DIFF
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -196,6 +196,21 @@ constexpr uint16_t DOS_PackDate(const struct tm &datetime) noexcept
 	                    static_cast<uint16_t>(datetime.tm_mday));
 }
 
+constexpr struct tm DOS_UnpackDateTime(const uint16_t date, const uint16_t time)
+{
+	struct tm ret = {};
+	ret.tm_sec = (time & 0x1f) * 2;
+	ret.tm_min = (time >> 5) & 0x3f;
+	ret.tm_hour = (time >> 11) & 0x1f;
+	ret.tm_mday = date & 0x1f;
+	ret.tm_mon = ((date >> 5) & 0x0f) - 1;
+	ret.tm_year = (date >> 9) + 1980 - 1900;
+	// have the C run-time library code compute whether standard
+	// time or daylight saving time is in effect.
+	ret.tm_isdst = -1;
+	return ret;
+}
+
 /* Routines for Drive Class */
 bool DOS_OpenFile(const char* name, uint8_t flags, uint16_t* entry, bool fcb = false);
 bool DOS_OpenFileExtended(const char* name, uint16_t flags,
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -119,7 +119,7 @@ constexpr int FakeSftEntries = 16;
 /* internal Dos Tables */
 
 extern std::unique_ptr<DOS_File> Files[DOS_FILES];
-extern std::array<DOS_Drive*, DOS_DRIVES> Drives;
+extern std::array<std::shared_ptr<DOS_Drive>, DOS_DRIVES> Drives;
 extern DOS_Device * Devices[DOS_DEVICES];
 
 extern uint8_t dos_copybuf[0x10000];
--- a/include/dos_inc.h
+++ b/include/dos_inc.h
@@ -118,7 +118,7 @@ constexpr int FakeSftEntries = 16;
 
 /* internal Dos Tables */
 
-extern DOS_File * Files[DOS_FILES];
+extern std::unique_ptr<DOS_File> Files[DOS_FILES];
 extern std::array<DOS_Drive*, DOS_DRIVES> Drives;
 extern DOS_Device * Devices[DOS_DEVICES];
 
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -124,7 +124,7 @@ public:
 	virtual bool	Read(uint8_t * data,uint16_t * size)=0;
 	virtual bool	Write(uint8_t * data,uint16_t * size)=0;
 	virtual bool	Seek(uint32_t * pos,uint32_t type)=0;
-	virtual bool	Close()=0;
+	virtual void	Close()=0;
 	virtual uint16_t	GetInformation(void)=0;
 	virtual bool IsOnReadOnlyMedium() const = 0;
 
@@ -168,7 +168,7 @@ public:
 	bool Read(uint8_t* data, uint16_t* size) override;
 	bool Write(uint8_t* data, uint16_t* size) override;
 	bool Seek(uint32_t* pos, uint32_t type) override;
-	bool Close() override;
+	void Close() override;
 	uint16_t GetInformation(void) override;
 	bool IsOnReadOnlyMedium() const override { return false; }
 	virtual bool ReadFromControlChannel(PhysPt bufptr, uint16_t size,
@@ -195,7 +195,7 @@ public:
 	bool Read(uint8_t* data, uint16_t* size) override;
 	bool Write(uint8_t* data, uint16_t* size) override;
 	bool Seek(uint32_t* pos, uint32_t type) override;
-	bool Close() override;
+	void Close() override;
 	uint16_t GetInformation() override;
 	bool UpdateDateTimeFromHost() override;
 	bool IsOnReadOnlyMedium() const override { return read_only_medium; }
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -190,38 +190,6 @@ private:
 	Bitu devnum;
 };
 
-class localFile : public DOS_File {
-public:
-	localFile(const char* name, const std_fs::path& path,
-	          const NativeFileHandle handle, const char* basedir,
-	          bool _read_only_medium);
-	localFile(const localFile&)            = delete; // prevent copying
-	localFile& operator=(const localFile&) = delete; // prevent assignment
-	bool Read(uint8_t* data, uint16_t* size) override;
-	bool Write(uint8_t* data, uint16_t* size) override;
-	bool Seek(uint32_t* pos, uint32_t type) override;
-	void Close() override;
-	uint16_t GetInformation() override;
-	bool IsOnReadOnlyMedium() const override { return read_only_medium; }
-	const char* GetBaseDir() const
-	{
-		return basedir;
-	}
-	std_fs::path GetPath() const
-	{
-		return path;
-	}
-	NativeFileHandle file_handle = InvalidNativeFileHandle;
-
-private:
-	void MaybeFlushTime();
-	const std_fs::path path = {};
-	const char* basedir     = nullptr;
-
-	const bool read_only_medium = false;
-	bool set_archive_on_close   = false;
-};
-
 /* The following variable can be lowered to free up some memory.
  * The negative side effect: The stored searches will be turned over faster.
  * Should not have impact on systems with few directory entries. */
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -317,26 +317,27 @@ public:
 	                                           uint8_t flags) = 0;
 	virtual std::unique_ptr<DOS_File> FileCreate(const char* name,
 	                                             FatAttributeFlags attributes) = 0;
-	virtual bool FileUnlink(const char* _name) = 0;
-	virtual bool RemoveDir(const char * _dir)=0;
-	virtual bool MakeDir(const char * _dir)=0;
-	virtual bool TestDir(const char * _dir)=0;
-	virtual bool FindFirst(const char * _dir,DOS_DTA & dta,bool fcb_findfirst=false)=0;
-	virtual bool FindNext(DOS_DTA & dta)=0;
+	virtual bool FileUnlink(const char* name)                           = 0;
+	virtual bool RemoveDir(const char* dir)                             = 0;
+	virtual bool MakeDir(const char* dir)                               = 0;
+	virtual bool TestDir(const char* dir)                               = 0;
+	virtual bool FindFirst(const char* dir, DOS_DTA& dta,
+	                       bool fcb_findfirst = false)                  = 0;
+	virtual bool FindNext(DOS_DTA& dta)                                 = 0;
 	virtual bool GetFileAttr(const char* name, FatAttributeFlags* attr) = 0;
 	virtual bool SetFileAttr(const char* name, const FatAttributeFlags attr) = 0;
 	virtual bool Rename(const char* oldname, const char* newname) = 0;
 	virtual bool AllocationInfo(uint16_t* _bytes_sector, uint8_t* _sectors_cluster,
 	                            uint16_t* _total_clusters,
 	                            uint16_t* _free_clusters) = 0;
-	virtual bool FileExists(const char* name)=0;
-	virtual uint8_t GetMediaByte(void)=0;
-	virtual void SetDir(const char *path);
+	virtual bool FileExists(const char* name)                     = 0;
+	virtual uint8_t GetMediaByte(void)                            = 0;
+	virtual void SetDir(const char* path);
 	virtual void EmptyCache() { dirCache.EmptyCache(); }
 	virtual bool IsReadOnly() const = 0;
-	virtual bool IsRemote(void)=0;
-	virtual bool IsRemovable(void)=0;
-	virtual Bits UnMount(void)=0;
+	virtual bool IsRemote(void)     = 0;
+	virtual bool IsRemovable(void)  = 0;
+	virtual Bits UnMount(void)      = 0;
 
 	DosDriveType GetType() const
 	{
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -130,7 +130,6 @@ public:
 
 	virtual void AddRef() { refCtr++; }
 	virtual Bits RemoveRef() { return --refCtr; }
-	virtual bool UpdateDateTimeFromHost() { return true; }
 
 	void SetDrive(uint8_t drv) { hdrive=drv;}
 	uint8_t GetDrive(void) { return hdrive;}
@@ -197,7 +196,6 @@ public:
 	bool Seek(uint32_t* pos, uint32_t type) override;
 	void Close() override;
 	uint16_t GetInformation() override;
-	bool UpdateDateTimeFromHost() override;
 	bool IsOnReadOnlyMedium() const override { return read_only_medium; }
 	const char* GetBaseDir() const
 	{
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -128,7 +128,6 @@ public:
 	virtual uint16_t	GetInformation(void)=0;
 	virtual bool IsOnReadOnlyMedium() const = 0;
 
-	virtual bool IsOpen() { return open; }
 	virtual void AddRef() { refCtr++; }
 	virtual Bits RemoveRef() { return --refCtr; }
 	virtual bool UpdateDateTimeFromHost() { return true; }
@@ -140,7 +139,6 @@ public:
 	uint16_t date    = 0;
 	FatAttributeFlags attr = {};
 	Bits refCtr      = 0;
-	bool open        = false;
 	std::string name = {};
 	bool newtime     = false;
 	std::vector<FileRegionLock> region_locks = {};
@@ -157,14 +155,13 @@ public:
 		: DOS_File(orig),
 		  devnum(orig.devnum)
 	{
-		open = true;
+
 	}
 
 	DOS_Device &operator=(const DOS_Device &orig)
 	{
 		DOS_File::operator=(orig);
 		devnum = orig.devnum;
-		open = true;
 		return *this;
 	}
 
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -343,10 +343,11 @@ public:
 	DOS_Drive();
 	virtual ~DOS_Drive() = default;
 
-	virtual bool FileOpen(DOS_File** file, const char* name, uint8_t flags) = 0;
-	virtual bool FileCreate(DOS_File** file, const char* name,
-	                        FatAttributeFlags attributes) = 0;
-	virtual bool FileUnlink(const char* _name)=0;
+	virtual std::unique_ptr<DOS_File> FileOpen(const char* name,
+	                                           uint8_t flags) = 0;
+	virtual std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                             FatAttributeFlags attributes) = 0;
+	virtual bool FileUnlink(const char* _name) = 0;
 	virtual bool RemoveDir(const char * _dir)=0;
 	virtual bool MakeDir(const char * _dir)=0;
 	virtual bool TestDir(const char * _dir)=0;
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -104,6 +104,12 @@ struct FileRegionLock {
 	uint32_t len = 0;
 };
 
+enum class FlushTimeOnClose {
+	NoUpdate,
+	ManuallySet,
+	CurrentTime,
+};
+
 class DOS_File {
 public:
 	DOS_File() = default;
@@ -139,7 +145,7 @@ public:
 	FatAttributeFlags attr = {};
 	Bits refCtr      = 0;
 	std::string name = {};
-	bool newtime     = false;
+	FlushTimeOnClose flush_time_on_close = FlushTimeOnClose::NoUpdate;
 	std::vector<FileRegionLock> region_locks = {};
 	/* Some Device Specific Stuff */
 private:
@@ -208,6 +214,7 @@ public:
 	NativeFileHandle file_handle = InvalidNativeFileHandle;
 
 private:
+	void MaybeFlushTime();
 	const std_fs::path path = {};
 	const char* basedir     = nullptr;
 
--- a/include/drives.h
+++ b/include/drives.h
@@ -40,18 +40,18 @@ class imageDisk; // forward declare
 
 class DriveManager {
 public:
-	using filesystem_images_t = std::vector<std::unique_ptr<DOS_Drive>>;
+	using filesystem_images_t = std::vector<std::shared_ptr<DOS_Drive>>;
 	struct DriveInfo {
 		filesystem_images_t disks = {};
 		uint16_t current_disk     = 0;
 	};
 	using drive_infos_t = std::array<DriveInfo, DOS_DRIVES>;
 
-	static std::vector<DOS_Drive*> AppendFilesystemImages(
-	        const int drive, filesystem_images_t& filesystem_images);
+	static void AppendFilesystemImages(const int drive,
+	                                   const filesystem_images_t& filesystem_images);
 
-	static DOS_Drive* RegisterFilesystemImage(
-	        const int drive, std::unique_ptr<DOS_Drive>&& filesystem_image);
+	static void RegisterFilesystemImage(const int drive,
+	                                    std::shared_ptr<DOS_Drive> filesystem_image);
 
 	static void InitializeDrive(int drive);
 	static int UnmountDrive(int drive);
--- a/include/drives.h
+++ b/include/drives.h
@@ -88,11 +88,11 @@ public:
 	           uint8_t _sectors_cluster, uint16_t _total_clusters,
 	           uint16_t _free_clusters, uint8_t _mediaid,
 	           bool _readonly, bool _always_open_ro_files = false);
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
 	FILE* GetHostFilePtr(const char* const name, const char* const type);
 	std::string MapDosToHostFilename(const char* const dos_name);
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
 	bool FileUnlink(const char* name) override;
 	bool RemoveDir(const char* dir) override;
 	bool MakeDir(const char* dir) override;
@@ -202,9 +202,9 @@ public:
 	         bool roflag);
 	fatDrive(const fatDrive&)            = delete; // prevent copying
 	fatDrive& operator=(const fatDrive&) = delete; // prevent assignment
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
 	bool FileUnlink(const char* name) override;
 	bool RemoveDir(const char* dir) override;
 	bool MakeDir(const char* dir) override;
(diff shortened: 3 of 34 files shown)


include/dos_system.h
```diff
@@ -343,10 +343,11 @@ class DOS_Drive {
 	DOS_Drive();
 	virtual ~DOS_Drive() = default;
 
-	virtual bool FileOpen(DOS_File** file, const char* name, uint8_t flags) = 0;
-	virtual bool FileCreate(DOS_File** file, const char* name,
-	                        FatAttributeFlags attributes) = 0;
-	virtual bool FileUnlink(const char* _name)=0;
+	virtual std::unique_ptr<DOS_File> FileOpen(const char* name,
+	                                           uint8_t flags) = 0;
```
johnnovak, line 317: `const uint8_t flags` ?

include/dos_system.h
```diff
@@ -343,10 +343,11 @@ class DOS_Drive {
...
+	virtual std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                             FatAttributeFlags attributes) = 0;
```
johnnovak, line 319: `const  FatAttributeFlags attributes` ?

include/dos_system.h
```diff
@@ -343,10 +343,11 @@ class DOS_Drive {
...
+	virtual bool FileUnlink(const char* _name) = 0;
```
johnnovak, line 320: minor: You can use `name` here, and only use `_name` in the implementation.

include/drives.h
```diff
@@ -88,11 +88,11 @@ class localDrive : public DOS_Drive {
 	           uint8_t _sectors_cluster, uint16_t _total_clusters,
 	           uint16_t _free_clusters, uint8_t _mediaid,
 	           bool _readonly, bool _always_open_ro_files = false);
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
```
johnnovak, line 77: const?

include/drives.h
```diff
@@ -88,11 +88,11 @@ class localDrive : public DOS_Drive {
...
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
```
johnnovak, line 79: const?

include/drives.h
```diff
@@ -202,9 +202,9 @@ class fatDrive final : public DOS_Drive {
 	         bool roflag);
 	fatDrive(const fatDrive&)            = delete; // prevent copying
 	fatDrive& operator=(const fatDrive&) = delete; // prevent assignment
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
```
johnnovak, line 197: const?

include/drives.h
```diff
@@ -277,9 +277,9 @@ class cdromDrive final : public localDrive
 	           uint16_t _bytes_sector, uint8_t _sectors_cluster,
 	           uint16_t _total_clusters, uint16_t _free_clusters,
 	           uint8_t _mediaid, int& error);
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
```
johnnovak, line 272: const?

include/drives.h
```diff
@@ -378,9 +378,9 @@ class isoDrive final : public DOS_Drive {
 	isoDrive(char driveLetter, const char* device_name, uint8_t mediaid,
 	         int& error);
 	~isoDrive() override;
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
```
johnnovak, line 374: const?

include/drives.h
```diff
@@ -450,9 +450,9 @@ using vfile_block_t = std::shared_ptr<VFILE_Block>;
 class Virtual_Drive final : public DOS_Drive {
 public:
 	Virtual_Drive();
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
```
johnnovak, line 446: const?

include/drives.h
```diff
@@ -495,9 +495,9 @@ class Overlay_Drive final : public localDrive {
 	              uint8_t _mediaid,
 	              uint8_t &error);
 
-	bool FileOpen(DOS_File** file, const char* name, uint8_t flags) override;
-	bool FileCreate(DOS_File** file, const char* name,
-	                FatAttributeFlags attributes) override;
+	std::unique_ptr<DOS_File> FileOpen(const char* name, uint8_t flags) override;
+	std::unique_ptr<DOS_File> FileCreate(const char* name,
+	                                     FatAttributeFlags attributes) override;
```
johnnovak, line 493: const?

src/dos/drive_iso.cpp
```diff
@@ -211,32 +211,35 @@ void isoDrive::Activate(void) {
...
-		(*file)->flags = flags;
+	if (!lookup(&de, name) && !IS_DIR(FLAGS1)) {
+		return nullptr;
 	}
-	return success;
+
+	FileStat_Block file_stat;
+	file_stat.size = DATA_LENGTH(de);
+	file_stat.attr = FatAttributeFlags::ReadOnly;
+	file_stat.date = DOS_PackDate(1900 + de.dateYear, de.dateMonth, de.dateDay);
```
johnnovak, line 226: Maybe extract 1900 to `DosEpochYear` or something?

src/dos/drive_local.cpp
```diff
@@ -172,17 +174,19 @@ bool localDrive::FileOpen(DOS_File **file, const char *name, uint8_t flags)
...
 	if (fallback_to_readonly) {
 		MaybeLogFilesystemProtection(host_filename);
 	}
 
-	*file = new localFile(name, host_filename, file_handle, basedir, IsReadOnly());
-	(*file)->flags = flags;  // for the inheritance flag and maybe check for others.
+	auto file = std::make_unique<localFile>(
+	        name, host_filename, file_handle, basedir, IsReadOnly());
+	file->flags = flags; // for the inheritance flag and maybe check for
+	                     // others.
```
johnnovak: minor: comment on line above while you're at it 😄

src/dos/drive_overlay.cpp
```diff
@@ -490,74 +480,87 @@ bool Overlay_Drive::FileOpen(DOS_File** file, const char* name, uint8_t flags)
 	CROSS_FILENAME(newname);
 
 	NativeFileHandle file_handle = open_native_file(newname, write_access);
-	bool fileopened = false;
 	if (file_handle != InvalidNativeFileHandle) {
-		if (logoverlay) LOG_MSG("overlay file opened %s",newname);
-		*file = new localFile(name, newname, file_handle, overlaydir, IsReadOnly());
+		if (logoverlay) {
+			LOG_MSG("overlay file opened %s", newname);
```
johnnovak, line 495: Prefix with `FS: ` log prefix?

```
FS: Overlay file '%s' opened.
```

src/dos/drive_overlay.cpp
```diff
@@ -490,74 +480,87 @@ bool Overlay_Drive::FileOpen(DOS_File** file, const char* name, uint8_t flags)
...
 
-	if (fileopened) {
-		if (logoverlay) LOG_MSG("file opened %s",name);
-		//Convert file to OverlayFile
-		OverlayFile* f = ccc(*file);
-		f->flags = flags; //ccc copies the flags of the localfile, which were not correct in this case
-		f->overlay_active = overlayed; //No need to switch if already in overlayed.
-		*file = f;
+	if (logoverlay) {
+		LOG_MSG("file opened %s", name);
```
johnnovak, line 536: Clean up log message as per above?

src/dos/drive_local.cpp
```diff
@@ -735,10 +736,7 @@ localFile::localFile(const char* _name, const std_fs::path& path,
 
 bool localFile::UpdateDateTimeFromHost()
 {
-	if (!open) {
-		return false;
-	}
-
+	assert(file_handle != InvalidNativeFileHandle);
```
johnnovak: Nice asserts! 👍🏻

include/dos_system.h
```diff
@@ -168,7 +168,7 @@ class DOS_Device : public DOS_File {
 	bool Read(uint8_t* data, uint16_t* size) override;
 	bool Write(uint8_t* data, uint16_t* size) override;
 	bool Seek(uint32_t* pos, uint32_t type) override;
-	bool Close() override;
+	void Close() override;
```
johnnovak, line 176: I agree. Nobody cares about whether `Close()` actually succeeded.

src/ints/bios_disk.cpp
```diff
@@ -598,14 +598,10 @@ void BIOS_SetupDisks(void) {
...
-		DriveManager::CloseNumberedImage(image_ptr);
-		image_ptr = nullptr;
-	}
+	imageDiskList.fill(nullptr);
 
 	// Clear any raw disk images
 	diskSwap.fill(nullptr);
-	DriveManager::CloseRawFddImages();
 
 	diskparm0 = CALLBACK_Allocate();
```
johnnovak, line 606: `This thing was handing out raw pointers like candy.`

LOL 😆

src/dos/program_mount.cpp
```diff
@@ -285,15 +284,14 @@ void MOUNT::Run(void) {
...
-			        mediaid,
-			        error);
+			newdrive  = std::make_shared<cdromDrive>(drive,
+			                                         temp_line.c_str(),
+			                                         sizes[0],
+			                                         int8_tize,
+			                                         sizes[2],
+			                                         0,
+			                                         mediaid,
+			                                         error);
```
johnnovak, line 294: Consider extracting the args to well-named constants; I have no idea what all this crap means 😆

include/dos_inc.h
```diff
@@ -196,6 +196,21 @@ constexpr uint16_t DOS_PackDate(const struct tm &datetime) noexcept
...
 
+constexpr struct tm DOS_UnpackDateTime(const uint16_t date, const uint16_t time)
+{
+	struct tm ret = {};
+	ret.tm_sec = (time & 0x1f) * 2;
+	ret.tm_min = (time >> 5) & 0x3f;
+	ret.tm_hour = (time >> 11) & 0x1f;
+	ret.tm_mday = date & 0x1f;
+	ret.tm_mon = ((date >> 5) & 0x0f) - 1;
+	ret.tm_year = (date >> 9) + 1980 - 1900;
```
johnnovak, line 207: Jesus... Existing code, I know... 🥶

src/misc/fs_utils_posix.cpp
```diff
@@ -457,4 +457,29 @@ DosDateTime get_dos_file_time(const NativeFileHandle handle)
...
+	// Last access time
+	// We don't really care about this but apparently it must be updated as well
+	unix_times[0].tv_sec = unix_seconds;
+	unix_times[0].tv_nsec = 0;
+
+	// Last modification time
+	unix_times[1].tv_sec = unix_seconds;
+	unix_times[1].tv_nsec = 0;
+
+	// I like hating on Win32 API but at least they have better function names...
```
johnnovak, line 480: 💯 

The Win32 API is quite nice.

src/misc/fs_utils_posix.cpp
```diff
@@ -457,4 +457,29 @@ DosDateTime get_dos_file_time(const NativeFileHandle handle)
...
+	// F U too Mr. Timens
```
johnnovak, line 481: Didn't quite get the joke... 😏

src/dos/dos_files.cpp
```diff
@@ -696,6 +696,15 @@ bool DOS_WriteFile(uint16_t entry,uint8_t * data,uint16_t * amount,bool fcb) {
 
 	bool ret=Files[handle]->Write(data,&towrite);
 	*amount=towrite;
+	if (ret) {
+		// When a write happens, mark that the time should be flushed on close.
+		// The updated time value is not the time of the write but the time of the close.
+		// Writes also do not update the local date/time fields.
+		// Local date/time fields (as returned by a call to DOS_GetFileDate())
+		// are set on file open and only get changed by a call to DOS_SetFileDate()
+		// This matches the behavior as tested on MS-DOS 6.22
```
johnnovak, line 705: That's some serious original research there 😎 🤘🏻

include/dos_system.h
```diff
@@ -190,38 +190,6 @@ class DOS_Device : public DOS_File {
...
-
-private:
-	void MaybeFlushTime();
-	const std_fs::path path = {};
-	const char* basedir     = nullptr;
-
-	const bool read_only_medium = false;
-	bool set_archive_on_close   = false;
-};
-
```
johnnovak, line 222: > I was running into circular header dependencies for what I wanted to do next.

Oh yeah, header files suck hard in C/C++. And the inability of two files to export/import symbols between themselves without a third intermediary.

src/dos/drive_local.cpp
```diff
@@ -90,11 +90,26 @@ std::unique_ptr<DOS_File> localDrive::FileCreate(const char* name,
...
 	}
 
-	// Make the 16 bit device information
-	auto file = std::make_unique<localFile>(
-	        name, expanded_name, file_handle, basedir, IsReadOnly());
+	DosDateTime dos_time = {};
+	dos_time.date        = DOS_GetBiosDatePacked();
+	dos_time.time        = DOS_GetBiosTimePacked();
+
+	// The timestamp cache is for emulating DOS behavior, not performance.
```
johnnovak, line 97: Good to clarify that upfront 💯

src/dos/drive_local.cpp
```diff
@@ -90,11 +90,26 @@ std::unique_ptr<DOS_File> localDrive::FileCreate(const char* name,
...
+	// On some hosts, the timestamp will be updated with current host time
+	// on write. DOS instead writes the timestamp on close. It also uses
+	// internal BIOS time which can differ from host time. Ex. The "date"
+	// command can set it to any arbitary date.
```
johnnovak, line 101: Excellent documentation of the DOS behaviour 🏅

src/dos/drive_local.cpp
```diff
@@ -176,10 +191,28 @@ std::unique_ptr<DOS_File> localDrive::FileOpen(const char* name, uint8_t flags)
...
-	file->flags = flags; // for the inheritance flag and maybe check for
-	                     // others.
+	DosDateTime dos_time   = {};
+	const auto cache_entry = timestamp_cache.find(host_filename);
+	if (cache_entry == timestamp_cache.end()) {
+		// The first time a file gets opened, read the timestamp from
+		// the host and add it to the cache. This gets "locked in" until
+		// the file is closed. This handles the case where a single file
+		// gets opened with multiple handles. The 2nd handle should not
+		// see updated timestamps from writes to the 1st handle.
```
johnnovak, line 201: You've thought of everything 💯

src/dos/drive_local.cpp
```diff
@@ -653,13 +688,36 @@ bool localFile::Seek(uint32_t *pos_addr, uint32_t type)
...
+	// lifetime. All we're doing is updating the cache which has no effect
+	// if there are no other references to the drive.
+	const auto drive_ptr = local_drive.lock();
+	if (drive_ptr) {
+		auto& cache = drive_ptr->timestamp_cache;
+		auto entry  = cache.find(path);
+		if (entry != cache.end()) {
+			// Only update the cache if it already contains
+			// the entry. This handles an edge case where
+			// the file has been unlinked while it is open.
```
johnnovak, line 718: 💯

src/dos/drive_overlay.cpp
```diff
@@ -487,9 +491,23 @@ std::unique_ptr<DOS_File> Overlay_Drive::FileOpen(const char* name, uint8_t flag
...
+		}
+
+		auto file            = std::make_unique<OverlayFile>(name,
+                                                          newname,
+                                                          file_handle,
+                                                          overlaydir,
+                                                          IsReadOnly(),
+                                                          weak_from_this(),
+                                                          dos_time,
+                                                          flags);
```
johnnovak, line 514: formatting

src/dos/drive_overlay.cpp
```diff
@@ -540,10 +558,20 @@ std::unique_ptr<DOS_File> Overlay_Drive::FileCreate(const char* name,
...
+	timestamp_cache[path] = dos_time;
+
+	auto file            = std::make_unique<OverlayFile>(name,
+                                                  path,
+                                                  file_handle,
+                                                  overlaydir,
+                                                  IsReadOnly(),
+                                                  weak_from_this(),
+                                                  dos_time,
+                                                  OPEN_READWRITE);
```
johnnovak, line 577: formatting

src/dos/drive_overlay.cpp
```diff
@@ -540,10 +558,20 @@ std::unique_ptr<DOS_File> Overlay_Drive::FileCreate(const char* name,
 		return nullptr;
 	}
 
-	auto file = std::make_unique<OverlayFile>(
-	        name, path, file_handle, overlaydir, IsReadOnly());
+	DosDateTime dos_time  = {};
+	dos_time.date         = DOS_GetBiosDatePacked();
+	dos_time.time         = DOS_GetBiosTimePacked();
```
johnnovak, line 567: Tip: You can use designated initialisers now that we got rid of old GCC versions:

```cpp
DosDateTime dos_time = {
    .date = DOS_GetBiosDatePacked(),
    .time = DOS_GetBiosTimePacked()
};
```

Also see 1b21f2f823f253e5d

src/dos/drive_overlay.cpp
```diff
@@ -1230,6 +1260,10 @@ bool Overlay_Drive::Rename(const char * oldname, const char * newname) {
 
 		result = !ec; // success if no error-code
 
+		if (result == true) {
```
johnnovak, line 1267: `if (result)` or better yet, also refactor `result` to `success`.

src/dos/drive_local.cpp
```diff
@@ -777,6 +777,20 @@ localFile::localFile(const char* _name, const std_fs::path& path,
 	SetName(_name);
 }
 
+localFile::~localFile()
+{
+	// Make sure we close the host file handle and flush the timestamps.
+	// This can happen if the user closes Dosbox while a game is running.
```
johnnovak, line 783: picky: DosBox

johnnovak (approved):
Great work @weirddan455, I very much agree with all your decisions. Nice commit messages and explanatory comments too! 🥇 👍🏻 

I've only some superficial comments about style and minor things; the overall approach is solid 👍🏻 I guess we'll see how things go with these changes over time, so feel free to merge after considering my comments.

src/misc/fs_utils_posix.cpp
weirddan455: Please fill out this struct with no less than 10 fields describing how much you like the Win32 API.

src/misc/fs_utils_posix.cpp
weirddan455: It was just the first thing that popped into my head when I was trying to figure out how they came up with the function name `futimens()`

It's probably actually something like "File descriptor Update Time in Nanoseconds" but come on...

weirddan455:
Thanks for the fast review @johnnovak

I've got to sort out the CI issues before merge.  I think they're fairly straight-forward and I'll get that plus your comments addressed in the next couple days or so.

src/misc/fs_utils_posix.cpp
johnnovak: Got it now 😆

src/misc/fs_utils_posix.cpp
johnnovak: To their credit, the struct approach is superior for C vs doing a function call like this:

```c
do_something_maybe(true, false, 0, 0, 4443, NULL, false, 3);
```

The structs are the poor man's keyword args for C & C++ 😆

src/dos/drive_iso.cpp
weirddan455: DOS uses 1980 as epoch, at least in the packed format.  1900 comes from C maybe?  https://cplusplus.com/reference/ctime/tm/

I'm not actually sure what it means it this context because it's coming from some internal ISO direntry struct.

src/dos/drive_local.cpp
weirddan455: I already removed this comment in a later commit.

src/dos/program_mount.cpp
weirddan455: I don't either, honestly.  Mostly all I did in this file was change from `unique_ptr` to `shared_ptr`.  I'm gonna save that for another day because I would probably have to refactor that entire large function to figure out what the heck is actually going on.

include/dos_inc.h
weirddan455: Yeah... I did review it before I copied it over and it looks correct to me though.

include/dos_system.h
weirddan455: It gets modified inside the functions.  Apparently I can lie in and state that it's const inside the header file and it still compiles with no errors/warnings interestingly enough.  Only the `const` in the actual function definition matters.

include/dos_system.h
weirddan455: Also gets modified inside some of the functions.

weirddan455: Also, not sure why we care if this argument is const.  It's pass by value so it doesn't matter to the caller.

src/dos/drive_overlay.cpp
weirddan455: That's nifty.  I used that syntax in one other place but I didn't know you could put function calls in there.  It compiled though.

src/dos/drive_overlay.cpp
weirddan455: Fixed

src/dos/drive_overlay.cpp
weirddan455: Fixed

src/dos/drive_overlay.cpp
weirddan455: You can blame clang-format for that.  It's not perfect.  Fixed it manually though.

src/dos/drive_overlay.cpp
weirddan455: fixed

src/dos/drive_overlay.cpp
weirddan455: fixed

src/dos/drive_local.cpp
weirddan455: fixed

include/dos_system.h
dreamer: Yup, that's how C/C++ works. Passing by value creates a copy, so it does not matter if function definition uses it as a const value or not. This also means that we can always make any parameter passed by value a const without modifying signature of the function (which is a good thing). One of those obscure C/C++ corners…

src/dos/drive_local.cpp
```diff
@@ -774,23 +778,22 @@ cdromDrive::cdromDrive(const char _driveLetter,
...
+		return nullptr;
 	}
-	bool success = localDrive::FileOpen(file, name, flags);
-	return success;
+	return localDrive::FileOpen(name, flags);
 }
 
-bool cdromDrive::FileCreate(DOS_File** /*file*/, const char* /*name*/,
-                            FatAttributeFlags /*attributes*/)
+std::unique_ptr<DOS_File> cdromDrive::FileCreate(const char* /*name*/,
```
dreamer, line 837: How about `[[maybe_unused]]` instead of commenting-out param?

src/dos/dos_files.cpp
```diff
@@ -1422,15 +1422,6 @@ bool DOS_FCBOpen(uint16_t seg, uint16_t offset)
 	if (!DOS_MakeName(shortname, fullname, &drive))
 		return false;
 
-	/* Check, if file is already opened */
```
dreamer, line 1425: Damn, this is a nice find that we can remove this. Great find and thank you for writing good commit message.

include/dos_system.h
weirddan455: @dreamer Yeah we had some discussion on that a while back.  I knew it worked the other way (const in the definition, non-const in the declaration).  I was a bit surprised by this though, considering it's also a virtual function.  Makes sense though.

I'm in favor of leaving these value types as-is then considering it doesn't even enforce that the argument is const in the implementation and seemingly makes no difference at all.

dreamer (commented):
Sorry for partial review, I was able to go through first few commits only so far.

Have you tried any testing with overlay drive engaged? I have weird feeling that we might find some regression there after all these refactors (hopefully not).

Can you upload your assembly test programs as well? Perhaps we can add them as tests (in as separate PR).

include/dos_system.h
johnnovak: Yeah, if you don't wanna refactor this whole mess further, that's fine @weirddan455 

But in general, to recap what we agreed on previously:

1. We use const both in function definitions and declarations for consistency. Having it in one place is a bit too clever, confusing, and not worth it, agreed.
2. We use const for local variables where applicable. The same reasoning that applies to making local vars const also applied to incoming function arguments, so we make them const whenever possible in new code.
3. When refactoring old code, yeah, fixing up function definitions and declarations it's up to the pain tolerance of the person doing the refactoring 😆

include/dos_system.h
johnnovak: As explained above. It's not from the point of the view of the caller.

It's from the point of view from the _function_. You'd make a local var const when applicable, even it's an int or a bool, no? The same reasoning goes for the incoming function args, otherwise they become "free to mutate for all" vars for no good reason 😄 

But, I get it it's can be hassle to refactor old code because you need to introduce temporaries then inside the function. I've been doing that, personally, but you don't have to do it now if you don't feel like it. But ultimately we should for consistency at some point IMO.

src/dos/drive_overlay.cpp
johnnovak: > That's nifty. I used that syntax in one other place but I didn't know you could put function calls in there. It compiled though.

Sure you can. It's just syntactic sugar for what you've done previously, plus the really good thing about it is that it raises a warning if any of the struct members are left uninitialised.

This is super handy when adding new struct members down the road. With the old-school way, you'd get no warning about the newly added member not being initialised.

src/dos/program_mount.cpp
johnnovak: Sure, it's no worse than before. That's something 😆

src/dos/drive_iso.cpp
johnnovak: Yeah, okay, maybe job for another day... 😅

src/dos/drive_overlay.cpp
weirddan455: Yeah, I like it.  I pushed it and it passed CI so I'm happy.  I think you could potentially run into problems if the order the functions get called in matters (in this case it doesn't).  Ex: If the order of the struct members differs from the order of your initialization list, I believe it will use the former.  Edge case though.  Most times we'll be better off using this.

weirddan455:
>  Have you tried any testing with overlay drive engaged? I have weird feeling that we might find some regression there after all these refactors (hopefully not).

No, I've never used that feature.  Honestly the code in there in there was a mess before I started this and I would expect bugs.  I'll give it a test later and report back though.

>  Can you upload your assembly test programs as well? Perhaps we can add them as tests (in as separate PR).

They're not really in a useful state for tests right now.  They literally just print out 1s and 0s to the screen as binary representation of the timestamps.  I was just using them to observe behavior on MS-DOS.

It would be nice to have some tests for this but how would you propose implementing that?  As far as I know, we have no testing infrastructure for running DOS programs and I don't think we have an assembler as part of our build toolchain either.

johnnovak:
> > Have you tried any testing with overlay drive engaged? I have weird feeling that we might find some regression there after all these refactors (hopefully not).
> 
> No, I've never used that feature. Honestly the code in there in there was a mess before I started this and I would expect bugs. I'll give it a test later and report back though.
> 

Yeah, we concluded with @weirddan455 that the overlay feature is not something we trust 😆 The code is full of TODOs about unimplemented features and comments that do not inspire confidence 😅 There's also a documented list in a Vogons post about stuff that the overlay feature doesn't handle.

So, in short, I guess we don't wanna make it worse, but it's a kinda/sorta broken feature that's good enough for the GOG cloud save support, but I'd not use it myself for anything outside of that, personally.

johnnovak:
> It would be nice to have some tests for this but how would you propose implementing that? As far as I know, we have no testing infrastructure for running DOS programs and I don't think we have an assembler as part of our build toolchain either.

I think we could chuck these asm sources into the dev section of our wiki in a ZIP with some comments, and that's it. I'm gonna refactor the dev wiki section at some point to have a proper "dev testing" master page that points to the video, audio, etc. subpages I've recently created.

src/dos/drive_local.cpp
weirddan455: Sure.  Changed.

weirddan455:
All the review comment that have come in so far have been addressed.

I tested overlay drive with save file from Monkey Island 2 and Crystal Caves and it works as expected.  Overwriting an existing save leaves the file in `localDrive` untouched and creates a new save in the overlay folder.  Creating a new save only puts it in the overlay.  Loading saves from both locations works.

@dreamer or anyone else feel free to leave more comments if you'd like.  I'll leave this up for another day or two.  If not, it's ready to merge I think.

src/dos/drive_local.cpp
```diff
@@ -876,7 +876,7 @@ bool cdromDrive::GetFileAttr(const char* name, FatAttributeFlags* attr)
 	return result;
 }
 
-bool cdromDrive::FindFirst(const char* _dir, DOS_DTA& dta, bool /*fcb_findfirst*/)
+bool cdromDrive::FindFirst(const char* _dir, DOS_DTA& dta, [[maybe_unused]]bool fcb_findfirst)
```
johnnovak, line 879: Nice 👍🏻

src/dos/drive_overlay.cpp
johnnovak: >  I think you could potentially run into problems if the order the functions get called in matters (in this case it doesn't).

Sure, if the function you're calling inline are not pure functions, anything is possible. That's a bit of broader problem though with non-pure functions.

PR #3791, opened 2024-06-23
Fix improper handling of some characters in shell history

FeralChild64:
# Description

DRAFT for now, because I'm still testing this code.

I have discovered a small shell history related bug:
- STEP: Press <LEFT CONTROL>+a, <ENTER>
- RESULT: A smiley face is printed and illegal command error (OK)
- RESULT: A log is printed, _UNICODE: No fallback mapping for code point 0x263a_ (NOT OK)
- STEP: Press <CURSOR UP> to bring back the last command
- RESULT: A question mark is brought back instead of the smiley face (not OK)

Screenshot

Solution: the Unicode engine needs to be precisely told what kind of DOS string do we have - with control codes, with screen codes, or with no special characters like that at all.
The Unicode extension implemented here might be helpful for the future use cases too - for UTF8 host file names, or for the upcoming (hopefully someday) EDIT.COM replacement.

# Manual testing

- try to reproduce the problem according to steps within the PR description; make sure that `shell_history_file` config option is not empty check that after restarting DOSBox the shell history is preserved, including the smiley
- check that `tree` command works without graphic glitches
- execute `keyb br 3848` command and check that the `tree` command still works, this time with fallback artwork
- check that `intro` command still works and does not produce graphical glitches
- press `CTRL+F6`, pay attention to the animated recording mark on the title bar and task bar (if your OS has one...) - there should be no regression here
- set `language = pl` config option, there should be no graphical glitches in the default tittle bar (but a couple of Polish national characters)
- set `language = pl` config option, type execute `more z:\autoexec.bat` - there should be no graphical glitches
- with language set to Polish (as above), execute `keyb pl` and `more z:\autoexec.bat` - there should be no graphical glitches, the first line of the file shall end with _ń_ letter (accented n)
- execute 'config -wcd' command with non-empty `[autoexec]` section; config file should be written correctly, including the `[autoexec]` section
- execute `config -axadd FOO`, and `config -axtype` afterwards - content of the updated `[autoexec]` section should be correctly printed out
- display help of any command, for example execute `dir /?`

# Checklist

I have:

### DIFF
--- a/include/string_utils.h
+++ b/include/string_utils.h
@@ -296,19 +296,36 @@ uint16_t get_utf8_code_page();
 // Specifies what to do if the DOS code page does not contain character
 // representing given Unicode grapheme
 enum class UnicodeFallback {
+
 	// Convert all unknown graphemes to 0 - to be used in code like TREE
 	// command implementation, which has it's own specialized ASCII fallback
 	// drawing code
 	Null,
+
 	// Try to provide reasonable fallback using all the characters available
 	// in target DOS code page; use for features like clipboard content
 	// exchange with host system
 	Simple,
+
 	// Do not use certain DOS code page characters in order to draw boxes
 	// (tables) which are consistent; for example, if code page contains
 	// character '╠', but not '╣', both will be replaced with a fallback
 	// character ('║' for example)
 	Box
+};
+
+// Specifies how to interpret characters 0x00-0x1f and 0x7f in DOS strings
+enum class DosStringConvertMode {
+
+	// String contains control codes (new line, tabulation, delete, etc.)
+	WithControlCodes,
+
+	// String does not have any codes characters, consider all characters
+	// as screen codes
+	ScreenCodesOnly,
+
+	// String should not contain characters mentioned above
+	NoSpecialCharacters
 };
 
 // Convert the UTF-8 string to the format intended for display inside emulated
@@ -317,12 +334,15 @@ enum class UnicodeFallback {
 // Return value 'false' means there were problems with string decoding or
 // rendering, but the overall output should still be sane.
 bool utf8_to_dos(const std::string& in_str, std::string& out_str,
+                 const DosStringConvertMode convert_mode,
                  const UnicodeFallback fallback);
 bool utf8_to_dos(const std::string& in_str, std::string& out_str,
+                 const DosStringConvertMode convert_mode,
                  const UnicodeFallback fallback, const uint16_t code_page);
-void dos_to_utf8(const std::string& in_str, std::string& out_str);
 void dos_to_utf8(const std::string& in_str, std::string& out_str,
-                 const uint16_t code_page);
+                 const DosStringConvertMode convert_mode);
+void dos_to_utf8(const std::string& in_str, std::string& out_str,
+                 const DosStringConvertMode convert_mode, const uint16_t code_page);
 
 // Convert DOS code page string to lower/upper case; converters are aware of all
 // the national characters. Functions without 'code_page' parameter use current
--- a/src/misc/unicode.cpp
+++ b/src/misc/unicode.cpp
@@ -43,7 +43,7 @@ CHECK_NARROWING();
 using box_drawing_set_t = std::array<uint16_t, 40>;
 
 // List of box drawing characters, ordered as in code page 437
-static constexpr box_drawing_set_t box_drawing_set_regular = {
+static constexpr box_drawing_set_t BoxDrawingSetRegular = {
         0x2502, // 0xb3 - '│', BOX DRAWINGS LIGHT VERTICAL
         0x2524, // 0xb4 - '┤', BOX DRAWINGS LIGHT VERTICAL AND LEFT
         0x2561, // 0xb5 - '╡', BOX DRAWINGS VERTICAL SINGLE AND LEFT DOUBLE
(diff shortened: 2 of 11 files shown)


include/string_utils.h
```diff
@@ -311,18 +311,31 @@ enum class UnicodeFallback {
...
 
+// Specifies how to interpret characters 0x00-0x1f and 0x7f in DOS strings
+enum class DosStringType {
+	// String contains control codes (new line, tabulation, delete, etc.)
+	WithControlCodes,
+	// String does not have any codes characters, consider all characters
+	// as screen codes
+	ScreenCodesOnly,
+	// String should not contain characters mentioned above
+	NoSpecialCharacters
```
johnnovak, line 322: Could you put new lines between the enums to aid readability?

include/string_utils.h
```diff
@@ -311,18 +311,31 @@ enum class UnicodeFallback {
 	Box
 };
 
+// Specifies how to interpret characters 0x00-0x1f and 0x7f in DOS strings
+enum class DosStringType {
```
johnnovak, line 315: Technically, it's a bit incorrect to call it like this. You're controlling the conversion logic, the DOS string is a DOS string, just a bunch of bytes. Maybe think of a better name? `DosStringConvertMode`?

include/string_utils.h
```diff
@@ -311,18 +311,31 @@ enum class UnicodeFallback {
...
-                 const UnicodeFallback fallback);
+                 const DosStringType string_type, const UnicodeFallback fallback);
 bool utf8_to_dos(const std::string& in_str, std::string& out_str,
+                 const DosStringType string_type,
                  const UnicodeFallback fallback, const uint16_t code_page);
-void dos_to_utf8(const std::string& in_str, std::string& out_str);
 void dos_to_utf8(const std::string& in_str, std::string& out_str,
-                 const uint16_t code_page);
+                 const DosStringType string_type);
+void dos_to_utf8(const std::string& in_str, std::string& out_str,
```
johnnovak, line 337: Instead of the old school "out var" style, you could never mutate passed in string args, but return `std::optional<std::string>` instead. This would improve the API.

src/misc/unicode.cpp
```diff
@@ -496,6 +496,98 @@ bool Grapheme::operator<(const Grapheme& other) const
 	return false;
 }
 
+// ***************************************************************************
+// Helpers for control codes handing
+// ***************************************************************************
+
+// Constants to detect whether DOS character is a control code
+constexpr uint8_t control_code_delete    = 0x7f;
+constexpr uint8_t control_code_threshold = 0x20;
```
johnnovak, line 505: UpperCamelConstants

Applied to all `constexpr` in this file.

src/misc/unicode.cpp
```diff
@@ -843,31 +958,16 @@ static bool wide_to_dos(const std::vector<uint16_t>& str_in, std::string& str_ou
...
-	        0x25ba, 0x25c4, 0x2195, 0x203c, // 10-13
-	        0x00b6, 0x00a7, 0x25ac, 0x21a8, // 14-17
-	        0x2191, 0x2193, 0x2192, 0x2190, // 18-1b
-	        0x221f, 0x2194, 0x25b2, 0x25bc, // 1c-1f
-	};
-
-	constexpr uint16_t codepoint_7f = 0x2302;
-
+static void dos_to_wide(const std::string& str_in, std::vector<uint16_t>& str_out,
+                        const DosStringType string_type, const uint16_t code_page)
```
johnnovak, line 962: Let's not use out args if we can avoid it. And it's trivial to avoid it in this case: just return `std::string`.

johnnovak (approved):
Looks like a nice improvement @FeralChild64 👍🏻

include/string_utils.h
FeralChild64: OK, no problem.

include/string_utils.h
FeralChild64: OK, changed.

src/misc/unicode.cpp
FeralChild64: OK, changed.

include/string_utils.h
FeralChild64: False does not mean the conversion totally failed, just that there were some problems - but the string should be usable. `std::optional` is not suitable here.

src/misc/unicode.cpp
FeralChild64: Some functions from this family returns true/false status, which is being used in locale/translation support. I prefer to keep the API consistent for now. Since I am in the process of reworking locale support, I'll keep this in mind.

johnnovak: Sounds good.

include/string_utils.h
johnnovak: @FeralChild64 Ok, that's a bit unexpected. Is this behaviour documented in the header files,?

FeralChild64: @johnnovak Yes, it's documented.

BTW, after some more checking it seems that API rework is easier than I thought. I'm preparing a follow-up PR, with some Unicode refactoring, https://github.com/dosbox-staging/dosbox-staging/pull/3797

johnnovak: Great, thanks @FeralChild64 👍🏻

PR #3164, opened 2023-12-01
Country support improvements

FeralChild64:
# Description

- cleanup - moved all the lengthy/boring locale tables to `dos_locale.cpp`; file contains a little business code, but not much
- country settings now influence currency symbol/format and list separators
- added possibility to select between historic and modern locale
- added `--list-countries` command line switch
- changed default code page for Bulgaria, as requested by a native speaker @Torinde

Many thanks to @rderooy for his work on extracting locale data from various DOS'es.

#### Limitations / out of scope of this PR / not bugs

- there is no `--list-keyboard-layouts` or `--list-codepages`; I agree they could be useful, but there is not enough time to do this for 0.80.1, besides the values are documented in `contrib/resources/freedos-*` directories; the list of supported countries was not documented anywhere
- I agree that locale synchronization with host can be improved in many ways, but again - no time for this
- there are going to be slight differences between DOSBox and MS-DOS 6.22 regarding currency; the reason is that MS-DOS often does not care whether the current code page supports national characters needed for the currency (for example, country 48 and standard code page gives a borked currency symbol results in MS-DOS), while I do care for this; not a bug, but a feature :D
- when no country is set, MS-DOS 6.22 sets the US locale but with a different date separator; I believe this is a bug in MS-DOS, which is not replicated
- MS-DOS / PC-DOS / DR-DOS uses the same country code for _Australia_ and _International English_; I have decided to use 0 for _International English_ in our configuration file, but the DOS API will return the same country ID as for _Australia_, for possibly better compatibility with guest software (DOS seems to never return country code 0)
- collation and uppercase tables are not generated yet, defaults are used; especially collation table will require quite some work to be generated well
- our internal `dir` command always uses 4-digit year (in reality it should print 2-digit year in case for Historic locale) - I intend to fix this with an upcoming command rework (won't be ready for 0.81.0)

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3128

# Manual testing

[CTEST.C.gz](https://github.com/dosbox-staging/dosbox-staging/files/13536348/CTEST.C.gz)
[CTEST.EXE.gz](https://github.com/dosbox-staging/dosbox-staging/files/13536349/CTEST.EXE.gz)

[...]

PR #3797, opened 2024-06-26
Unicode API refactoring

FeralChild64:
# Description

- Unicode routines now have their own header file - `string_utils.h`, IMHO, grew too crowded
- DOS<->Unicode conversion API was reworked; there is no `bool` return value anymore (it's usage in current code was replaced by searching the output string for `\0` characters)
- some minor internal changes

# Manual testing

- checked that `intro` command still works, even with `language = pl` setting
- checked that `tree` command still works, even after `keyb br 3848` command (in this case it should use fallback ASCII artwork)
- using country test program available in other PR (https://github.com/dosbox-staging/dosbox-staging/pull/3164) I have checked that for `country = 48` (Poland) currency symbol is `zł` with code page 668 (execute _keyb pl 668_ command) and `zl` (no stroke on `l`) for code page 437;
- using the same tool I have checked that for `country = 49` and `locale_period = modern`  currency symbol is `EUR` with code page 437 (`keyb us 437` command) and `€` with code page 858 (`keyb us 858` command)

# Checklist

I have:


include/unicode.h
```diff
@@ -0,0 +1,90 @@
...
+uint16_t get_utf8_code_page();
+
+// Specifies what to do if the DOS code page does not contain character
+// representing given Unicode grapheme
+enum class UnicodeFallback {
+
+	// Convert all unknown graphemes to 0 - to be used in code like TREE
+	// command implementation, which has it's own specialized ASCII fallback
+	// drawing code
+	Null,
```
weirddan455, line 38: It took me a while to figure out what was going on with the searching for `\0` in the returned string.  As far as I can tell, in all use-cases that use this mode, the returned string is never used if it failed to convert.  Instead, they're checking if the returned string has null terminators in the middle of it and, if so, treats it as an error condition.

I would suggest instead returning a `std::optional<std::string>` or just an empty string to indicate failure.  The other two modes appear to never fail but instead provide a fallback so you could also remove this mode from the enum and put it into a separate function that can fail if that makes more sense.

Example: `std::optional<std::string> utf8_to_dos_or_fail` and leave the other two modes simply returning `std::string`.

I'll leave that up to you to decide if that makes sense or not.

src/dos/dos_locale.cpp
```diff
@@ -2616,34 +2617,23 @@ static void refresh_currency_format(const LocaleInfoEntry &source)
...
-		                 candidate,
-		                 DosStringConvertMode::NoSpecialCharacters,
-		                 UnicodeFallback::Null)) {
-			continue;
-		}
-		if (candidate.length() > MaxCurrencySymbolLength) {
+		// TODO[C++23]: Use 'contains()' instead of 'find()'
+		if (candidate.find('\0') != std::string::npos ||
+		    candidate.length() > MaxCurrencySymbolLength) {
 			continue;
```
weirddan455, line 2629: Usage case: discards the string if `Null` fallback fails.  Also looks confusing.  Why are we searching for a null termination character in the middle of a C++  `std::string`?

src/dos/program_tree.cpp
```diff
@@ -191,9 +190,9 @@ void TREE::PreRender()
...
 		// draw the tree, use standard 7-bit ASCII characters
 
-		std::string tmp_str;
-		to_dos("─├│└", tmp_str);
-		if (tmp_str.find('\0') != std::string::npos) {
+		const auto tmp = to_dos("─├│└");
+		// TODO[C++23]: Use 'contains()' instead of 'find()'
+		if (tmp.find('\0') != std::string::npos) {
 			use_ascii_fallback = true;
 		}
```
weirddan455, line 197: Usage case: Also discarding the returned string.

include/unicode.h
FeralChild64: This was done with one future use case in mind: the filesystem, namely Unicode file names (UTF-16 on Windows, UTF-8 elsewhere). I thought in case the function returns some `\0`'s (usually when the current code page does not have a suitable character) the filesystem emulation could still handle it somehow, like the host `FOOϪBAR.TXT` file could be presented as `FOOBAR~1.TXT` or `FOO_BA~1.TXT`to the guest software, in a similar way long file names are shown.

So maybe I should keep the `Null`, but add additional mode, for example called `EmptyString`? Right now you are probably the most experienced with filesystem code, so I'm asking you for an advice :)

weirddan455: Hmm... maybe that would be useful for the filesystem but it's hard to tell in advance what the best approach is until someone implements it.

The DOS to host file mapping is done in `drive_cache.cpp` which I have not touched.  I took a look at it now and I think it will need a fair bit of refactoring to properly handle Unicode.  Not sure how well it deals with non-ASCII characters in filenames at the moment but I'm going to assume not very well.

Everything is done using C style `char` arrays and pointers and heavy use of `strcpy` `strlen` and friends.  It's getting the host file names with `opendir` and `readdir` which may give back something utf-8 encoded but may blow up on Windows (don't think these functions can ever give back utf-16 encoding).

>  So maybe I should keep the Null, but add additional mode, for example called EmptyString? 

Yeah, that's probably fine.  Maybe in the future the drive cache can make use of a "token" null terminator and replace it with something sensible but we can cross that bridge when we come to it.

johnnovak: In general, adding code we don't need right now but to cater for a future scenario is not a good idea It's not easy to predict the future 😄 Unless you're quite sure you'll need something... but even that can often change.

So I'm inclined to say remove it if we don't need it right now.

include/unicode.h
```diff
@@ -0,0 +1,90 @@
...
+	// as screen codes
+	ScreenCodesOnly,
+
+	// String should not contain characters mentioned above
+	NoSpecialCharacters
+};
+
+// Convert the UTF-8 string to the format intended for display inside emulated
+// environment, or vice-versa. Code page '0' means a pure 7-bit ASCII. Functions
+// without 'code_page' parameters uses current DOS code page.
```
johnnovak, line 68: typo: use

include/unicode.h
```diff
@@ -0,0 +1,90 @@
...
+                        const UnicodeFallback fallback);
+std::string utf8_to_dos(const std::string& in_str,
+                        const DosStringConvertMode convert_mode,
+                        const UnicodeFallback fallback,
+                        const uint16_t code_page);
+std::string dos_to_utf8(const std::string& in_str,
+                        const DosStringConvertMode convert_mode);
+std::string dos_to_utf8(const std::string& in_str,
+                        const DosStringConvertMode convert_mode,
+                        const uint16_t code_page);
```
johnnovak, line 80: minor: Please insert new lines between these to break up the block of text. It's quite unreadable as it is now.

include/unicode.h
```diff
@@ -0,0 +1,90 @@
...
+
+// Convert DOS code page string to lower/upper case; converters are aware of all
+// the national characters. Functions without 'code_page' parameter use current
+// DOS code page.
+void lowercase_dos(std::string& in_str);
+void lowercase_dos(std::string& in_str, const uint16_t code_page);
+void uppercase_dos(std::string& in_str);
+void uppercase_dos(std::string& in_str, const uint16_t code_page);
```
johnnovak, line 88: For consistency these should also return a string instead of mutating it in place?

src/gui/titlebar.cpp
```diff
@@ -430,6 +430,7 @@ void GFX_NotifyAudioMutedStatus(const bool is_muted)
 void GFX_NotifyProgramName(const std::string& segment_name,
                            const std::string& canonical_name)
 {
+	constexpr auto convert_mode  = DosStringConvertMode::ScreenCodesOnly;
```
johnnovak, line 433: ConvertMode

include/unicode.h
```diff
@@ -0,0 +1,90 @@
...
+
+#endif
```
johnnovak, line 90: I don't see the value of the `in_` prefix in the name `in_str`. Surely, all `const` function arguments are "inputs" 😄

src/misc/unicode.cpp
```diff
@@ -594,7 +596,7 @@ static std::optional<uint8_t> grapheme_to_screen_code(const Grapheme& grapheme,
 // Conversion routines
 // ***************************************************************************
 
-static bool utf8_to_wide(const std::string& str_in, std::vector<uint16_t>& str_out)
+static wide_string utf8_to_wide(const std::string& str_in)
```
johnnovak, line 599: Just `str` instead of `str_in` (all `const` args are inputs, that's obvious).

src/misc/unicode.cpp
```diff
@@ -685,24 +699,25 @@ static bool utf8_to_wide(const std::string& str_in, std::vector<uint16_t>& str_o
...
 
 		str_out.push_back(static_cast<uint16_t>(code_point));
 	}
 
-	return status;
+	return str_out;
 }
 
-static void wide_to_utf8(const std::vector<uint16_t>& str_in, std::string& str_out)
+static std::string wide_to_utf8(const wide_string& str_in)
```
johnnovak, line 718: Just `str` instead of `str_in` (all `const` args are inputs, that's obvious).

src/misc/unicode.cpp
```diff
@@ -754,12 +770,12 @@ static void warn_code_page(const uint16_t code_page)
 	LOG_WARNING("UNICODE: Requested unknown code page %d", code_page);
 }
 
-static bool wide_to_dos(const std::vector<uint16_t>& str_in, std::string& str_out,
-                        const DosStringConvertMode convert_mode,
-                        const UnicodeFallback fallback, const uint16_t code_page)
+static std::string wide_to_dos(const wide_string& str_in,
+                               const DosStringConvertMode convert_mode,
+                               const UnicodeFallback fallback,
+                               const uint16_t code_page)
```
johnnovak, line 776: Just `str` instead of `str_in` (all `const` args are inputs, that's obvious).

src/misc/unicode.cpp
```diff
@@ -957,14 +972,14 @@ static bool wide_to_dos(const std::vector<uint16_t>& str_in, std::string& str_ou
...
-	return status;
+	return str_out;
 }
 
-static void dos_to_wide(const std::string& str_in, std::vector<uint16_t>& str_out,
-                        const DosStringConvertMode convert_mode,
-                        const uint16_t code_page)
+static wide_string dos_to_wide(const std::string& str_in,
+                               const DosStringConvertMode convert_mode,
+                               const uint16_t code_page)
```
johnnovak, line 980: `str_in` -> `str`

src/misc/unicode.cpp
```diff
@@ -2107,62 +2124,59 @@ uint16_t get_utf8_code_page()
...
 }
 
-static bool utf8_to_dos_common(const std::string& in_str, std::string& out_str,
-                               const DosStringConvertMode convert_mode,
-                               const UnicodeFallback fallback,
-                               const uint16_t code_page)
+static std::string utf8_to_dos_common(const std::string& in_str,
+                                      const DosStringConvertMode convert_mode,
+                                      const UnicodeFallback fallback,
+                                      const uint16_t code_page)
```
johnnovak, line 2130: `str_in` -> `str`

johnnovak (approved):
Looks good @FeralChild64 , just some minor comments.

include/unicode.h
FeralChild64: OK, changed.

include/unicode.h
FeralChild64: OK, changed everywhere.

src/misc/unicode.cpp
FeralChild64: OK, changed everywhere.

src/misc/unicode.cpp
FeralChild64: OK, changed everywhere.

src/misc/unicode.cpp
FeralChild64: OK, changed everywhere.

src/misc/unicode.cpp
FeralChild64: OK, changed everywhere.

src/misc/unicode.cpp
FeralChild64: OK, changed everywhere.

src/gui/titlebar.cpp
FeralChild64: OK, changed.

include/unicode.h
FeralChild64: OK, changed.

include/unicode.h
FeralChild64: Fixed.

include/unicode.h
FeralChild64: OK, I have changed the fallback mode. It'll be easy to provide the _Null_ one if needed in the future.

src/dos/dos_locale.cpp
FeralChild64: As above, changed the fallback mode.

src/dos/program_tree.cpp
FeralChild64: As above, changed the fallback mode.

PR #3801, opened 2024-07-01
Add DOS Files to shutdown step to avoid a crash when shutting down

interloper98:
# Description

When closing Windows 3.11 with Ctrl+F9, my debug build on RetroPie would crash with:

```
11      0x62fedb2b4753 std::unique_ptr<DOS_File, std::default_delete<DOS_File> >::~unique_ptr() + 67
10      0x62fedb2b613c std::default_delete<DOS_File>::operator()(DOS_File*) const + 44
9       0x62fedb3545c9 localFile::~localFile() + 25
8       0x62fedb354560 localFile::~localFile() + 64
7       0x62fedb354337 localFile::Close() + 295
6       0x62fedb35406b localFile::MaybeFlushTime() + 203
5       0x62fedb297a32 DOS_GetBiosTimePacked() + 18
4       0x62fedb55a383 unsigned int mem_readd<(MemOpMode)0>(unsigned int) + 19
```

What's happening is that there's a global array holding the open DOS file objects:

```
std::unique_ptr<DOS_File> Files[DOS_FILES] = {};
```

Adding some tracing, it looks like DOSBox shutdowns each of the 'Init'd subsystems in reverse order that they were setup. 
 - One of them is the the DOS subsystem (in dos.cpp), which clears the DOS drives, devices, and memory layout.
 - The overall DOSBox shutdown sequence keeps going and clears our the physical layers like the RAM.

However, that global array holding the DOS files is still active. Those DOS File objects are finally destructed just before the DOSBox Staging process ends -- but by then the DOS subystem is long gone so those DOS File destructors aren't able to access things like 'DOS_GetBiosTimePacked' (which require reading into DOS memory space). 

So this PR just adds the DOS files to that DOS shutdown phase, so they're not left orphaned.

## Related issues

None.

# Manual testing

Tested controlled and force shutdowns (Ctrl+F9) with Windows 3.11 and various games; and didn't hit any crashes.

# Checklist

I have:

### DIFF
--- a/src/dos/dos_files.cpp
+++ b/src/dos/dos_files.cpp
@@ -44,7 +44,7 @@
 #define FCB_ERR_EOF     3
 #define FCB_ERR_WRITE   1
 
-std::unique_ptr<DOS_File> Files[DOS_FILES] = {};
+std::array<std::unique_ptr<DOS_File>, DOS_FILES> Files = {};
 
 std::array<std::shared_ptr<DOS_Drive>, DOS_DRIVES> Drives = {};
 
--- a/src/dos/dos_files.cpp
+++ b/src/dos/dos_files.cpp
@@ -1800,12 +1800,7 @@ bool DOS_SetFileDate(uint16_t entry, uint16_t ntime, uint16_t ndate)
 
 void DOS_SetupFiles()
 {
-	/* Setup the File Handles */
-	for (uint8_t i = 0; i < DOS_FILES; ++i)
-		Files[i] = nullptr;
-	/* Setup the Virtual Disk System */
-	for (uint8_t i = 0; i < DOS_DRIVES; ++i)
-		Drives[i] = nullptr;
+	DOS_ClearDrivesAndFiles();
 
 	const auto z_drive_index = drive_index('Z');
 
@@ -1855,4 +1850,17 @@ bool DOS_UnlockFile(const uint16_t entry, const uint32_t pos, const uint32_t len
 	}
 	DOS_SetError(DOSERR_LOCK_VIOLATION);
 	return false;
+}
+
+void DOS_ClearDrivesAndFiles()
+{
+	// Clear all the DOS files. This calls the files' derived destructors,
+	// which might be LocalFiles, OverlayFile, etc.
+	for (auto& f : Files) {
+		f = nullptr;
+	}
+
+	// Clear the shared drive pointers. The actual objects are managed by
+	// the drive manager class.
+	Drives.fill(nullptr);
 }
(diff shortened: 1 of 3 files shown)


src/dos/dos_files.cpp
```diff
@@ -1856,3 +1851,16 @@ bool DOS_UnlockFile(const uint16_t entry, const uint32_t pos, const uint32_t len
...
 }
+
+void DOS_ClearDrivesAndFiles()
+{
+	// Clear all the DOS files. This calls the files' derived destructors,
+	// which might be LocalFiles, OverlayFile, etc.
+	for (auto& f : Files) {
+		f.reset();
+	}
+
```
interloper98, line 1862: @weirddan455, I wanted to use the `.fill(nullptr)` like you did for the Drives array, but I hit some issues with the `operator =` being ambiguous for the DOS_File class, so had to switch to this looping approach.

weirddan455: I'm getting the same error.  I think it's because the `DOS_Drive` array contains `shared_ptr` while the `DOS_File` array contains `unique_ptr`.  `.fill()` doesn't have an overload for r-value references and `unique_ptr` can only be moved, never copied.

Although if you put `f = nullptr` in the loop it works fine.  Just weird C++ things :roll_eyes:  `.fill()` must be constructing a temporary and then attempting to do a copy, which fails since that operator is deleted for `unique_ptr`.

weirddan455 (approved):
Looks good.  I missed this in testing because it only happens when a program explicitly requests the date/time to be changed with the interrupt that calls `DOS_SetFileDate`.  And then you have to quit DOSBox while the file is still open to trigger the bug.

Nice find.  Good to get more things explicitly cleaned up instead of leaving it to whenever C++ decides to run global destructors.

src/dos/dos_files.cpp
weirddan455: I guess it makes sense in a way because you can't move a single object to 256 different places so it has to copy.

interloper98: Thanks @weirddan455, that makes sense. 

Replaced the reset() with the = nullptr; thanks!

interloper98:
> it only happens when a program explicitly requests the date/time to be changed with the interrupt that calls DOS_SetFileDate.

Nice - in that case, very happy to have stumbled on this corner case!

> Good to get more things explicitly cleaned up instead of leaving it to whenever C++ decides to run global destructors.

Right on. Yes; looks like that global `Files` array has been in the code a long time without having a cleanup.

I don't think there's much risky stuff here - so going to merge it. Thanks for the review @weirddan455 !

src/dos/dos_files.cpp
weirddan455: `f.reset()` is probably more idiomatic for smart pointers.  I was just pointing out assigning nullptr in a loop works because I thought it was interesting.  Doesn't really matter either way though, it does the same thing.

interloper98: Good point; I probably should have stuck with it, but figured pairing it in style w/ the Drives was a better fit. Agree though it's not a big deal though :-), but will remember for next time.

Issue #3677, opened 2024-05-16
500 USD bounty for DOSBox Staging official Debian package


Grounded0:
Modified: Wed 10 Jul 2024 21:58:13 UTC

Deploy DOSBox Staging into the official Debian package repository with a one-year commitment to being the package maintainer and collect your 500 USD. I will not raise this any further.

I can pay via SWIFT, the cryptos available in Coinbase, or PayPal.

Resolves a long-standing ticket opened in 2020:

- #664

### Links

[Debian ITP for DOSBox Staging](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=973822)
[Debian ITS for DOSBox Staging](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1036708)

### Connected issues and PRs

- #418
- #664
- #3806
- #3809

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

okias:
I recommend specifying the outcome more precisely.

Do you want to get this into Debian or just external repository.
 - Maintained by who? For how long? etc. etc.
 - Should the person become the maintainer of the package in Debian? etc. etc.

Grounded0:
DOSBox Staging 0.81.1 into Debian repository. Maintain through 2024.

kkremitzki:
As a Debian Developer with only a superficial awareness of DOSBox, I took a look at the package and its open bugs. I can see that Debian would be amenable to replacing the DOSBox package with one of its maintained forks to address longstanding issues (moreso than keeping the package and adding one or more forks.) However, a primary question would be why one fork should be chosen over another. If there were a brief explanation of this available, that would address the main non-technical obstacle to getting this packaged for Debian.

runlevel5:
Packaging would be easy part. The hard part is to get it accepted into Debian. Who's going to be the maintainer moving forward? I highly recommend you bring this up with the Debian community instead.

Castor215:
This was mentioned on the mailing lists but I'm bringing it up here, too. What are the binary blobs in `contrib/resources`? I'm talking about files that end in .com, .sys, .exe and .cpx/cpi. These seem to be already-compiled files and would be a violation of the DFSG (specifically guideline 2 "Inclusion of source code"). As it is, dosbox-staging would need to go in the `contrib` or `non-free` sections. Can dosbox-staging function without these files?

FeralChild64:
[...]

PR #3809, opened 2024-07-04
Add Debian package build scripts and workflow

interloper98:
# Description

Adds `debian/*` records and a `build-debian-package.sh` script to build the `.deb` package. This can be run locally by users, and it's also called by a new workflow.

Regarding the workflow:
 - It's separate and not tacked onto the Linux one because the Debian tools require a clean repo (the Debian tools diff the source archive against the repo).
 - It uses Ubuntu 22.04 (instead of Docker) because it was very straight forward. 
 - I've matrix'd it by architecture, so it's ready to go when we get arm64 Ubuntu worker machines.

The artifact is a 3 MB zip file that contains the `.deb` file.  Users on any Debian-based system can run `apt install ./dosbox-staging-<version>.deb` and it will install all the dependencies as well as the package.

You can see an example here:

https://github.com/dosbox-staging/dosbox-staging/actions/runs/9799340051

## Related issues

Makes progress on the Debian tickets #3806 and #3677:
 - It's a small win for Debian users because this package leverages all available Debian libraries while giving them a full-featured DOSBox Staging binary.
 - Once installed, DOSBox Staging is part of the user's dpkg inventory. It has hooks into dependencies and can be queried, just like any other .deb package.
 - The only thing missing is it's not part of the official Debian repo; only the Debian maintainers can decide if they want to use any of this.

# Manual testing

Tested the builder and `.deb` package on Debian 12 running on a Raspberry Pi.

# Checklist

I have:


.github/workflows/debian-package.yml
```diff
@@ -0,0 +1,77 @@
...
+
+  pull_request:
+#    branches: [ main ]
+    paths-ignore:
+      - '.clang-format'
+      - '.mdl-styles'
+      - '*.md'
+      - 'docs/**'
+      - 'licenses/**'
+      - 'website/**'
```
interloper98, line 22: Thinking we probably don't need to generate .deb files for every branch, and instead only produce debs from `main` branch commits.

if the team agrees I'll comment-in these two branch lines in before merge.

debian/.gitignore
```diff
@@ -0,0 +1,8 @@
+.debhelper/
+debhelper-build-stamp
+dosbox-staging.debhelper.log
+dosbox-staging.substvars
+dosbox-staging/
+files
+changelog
+
```
johnnovak, line 8: Can't we move `debian` to `contrib/linux/debian`? I don't really like it at the root level.

.github/workflows/debian-package.yml
johnnovak: Oh, main is definitely more than enough 😄 Maybe only tagged releases would be enough, so not on every commit?

interloper98: Yes; maybe not every commit 😄 , but this would build a deb at least for the PR merges to main, so more like "every feature". 
(There's a lot of active debian PPA packages that track upstream maik branches, so this feel about right and matches how the website let's people download zips of the main)

johnnovak: If it runs quickly and it's cheap, why not, yeah. If it's long-running and heavy, we'll need to discuss this further. After all, it's a relatively niche thing, so we don't want it to cause problems and slowdowns.

debian/.gitignore
interloper98: Yes! I've now moved it now, and the debian package script simply copies it to root and blows it away as soon as it's done using a bash trap. It's working well and now that pollution is swept firmly under the rug 🧹 👍

johnnovak: Nice one @interloper98 👍🏻

.github/workflows/debian-package.yml
interloper98: > If it runs quickly and it's cheap, why not, yeah.

(:+1: - it completes faster than all the workflows; it's only a single build)

Updated to only run after PR's are merged to `main`, so basically the Debian-equivalent for how Mac users getting a .dmg and Windows users getting a setup.exe that tracks the `main` branch.

Debian-based distros are (I think) the most popular, so this hopefully will be a good option for those who don't have a build environment.

johnnovak: > Updated to only run after PR's are merged to main, so basically the Debian-equivalent for how Mac users getting a .dmg and Windows users getting a setup.exe that tracks the main branch.

Yeah, seem reasonable.

> Debian-based distros are (I think) the most popular, so this hopefully will be a good option for those who don't have a build environment.

Cool, just pointing out we've been publishing Linux dev build binaries for quite a while now:
https://github.com/dosbox-staging/dosbox-staging/actions/runs/9771784818

So grabbing simply the binaries has always been an option regardless of what Linux distro you're on, and continues to be an option.

johnnovak (approved):
Thanks for this @interloper98, high quality work as usual. 🎖️ 

Hopefully some of the Debian guys will make Staging part of their official package selection with this in place. The ball is certainly in their cour now; we can't possibly help even more than this 😏

.github/workflows/debian-package.yml
interloper98: > So grabbing simply the binaries has always been an option regardless of what Linux distro 

Understood; zip and deb have the same binary inside.

(The only benefit here is just a bit more integration for Debian-based users:  their package manager will download and install all the dependencies and `dosbox` is added to their path; these are minor but some people like those freebies :laughing: )

interloper98:
Thanks for the review, @johnnovak! 
Ok, will merge.

PR #3812, opened 2024-07-07
Eliminate excessive repeat heap calls during normal runtime

interloper98:
# Description

This PR uses an object pool to manage the invalidation maps used in the dynamic core's cache blocks. There can be thousands of cache blocks in flight, and their invalidation maps can come and go in rapid succession.

This PR switches to using `try_emplace` in the IO port byte handlers. The 'try_emplace' versus 'emplace' avoids creating a temporary object on the heap before emplacing.  `try_emplace` was added in C++17 is usually an improvement -- https://stackoverflow.com/a/46047547  (in this case, I'm suggesting we use it here because the IO ports can be hit tens of thousands of times a second).

I tested with optimised release builds to be sure the compiler has done all of its smart temporary value elimination and code elision.

On the raspberry pi, I used 'heaptrack /path/to/dosbox' to monitor heap allocations.

Before the PR, 60 second run of Quake timedemo 320x200:
 - Heap allocations: 700,000
 - Temporary heap allocation: 550,000

After the PR, 60 second run of Quake timedemo 320x200:
 - Heap allocations: 180,000
 - Temporary heap allocation: 14,000

I looked into the remainder, and most of these happen at startup and shutdown - which is fine, however there's still some burned at runtime and those are caused by the raspberry pi's video driver.  As for SDL and OpenGL -> these libraries aren't making any heap call during the runtime phase, which is great news. 

## Related issues

None.

# Manual testing

Tested various games, applications, and Windows 3.11 for workgroups.

For those on Linux, you can check on your system by installing 'heaptrack' and 'heaptrack-gui'.  I used optimized builds for comparisons.

Then launch with: `heaptrack /path/to/dosbox`

When DOSBox Staging quits, heaptrack will automatically open the results in its GUI.

Performance might have ticked up by a frame or two on the Pi. I was getting max results slightly better than I've been getting before with `main`.

Comparison of allocations (top=before, bottom=after):

allocations

Comparison of temporary allocations (top=before, bottom=after):

temporary_allocations

# Checklist

I have:

### DIFF
--- a/include/object_pool.h
+++ b/include/object_pool.h
@@ -1,0 +1,109 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2024-2024  The DOSBox Staging Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#ifndef DOSBOX_OBJECT_POOL_H
+#define DOSBOX_OBJECT_POOL_H
+
+#include "dosbox.h"
+
+// This class is a generic object re-use pool. Consider using it if your program
+// performs tens of thousands (or more) large object heap allocations as part of
+// its normal and ongoing runtime loop. Avoiding excessive heap allocations can
+// save CPU time and reduce memory fragmentation.
+//
+// Ongoing runtime loop:
+//     auto my_obj_ptr = new HugeType(arg1, arg2);
+//     .. do a bit of work ..
+//     delete my_obj_ptr;
+//
+// If you have this pattern, then the ObjectPool is a good candidate:
+//
+//  Instantiate a long-lived pool for your HugeType:
+//     ObjectPool<HugeType> pool = {};
+//
+//  Ongoing runtime loop:
+//     auto my_obj_ptr = pool.Acquire(arg1, arg2);
+//     .. do a bit of work ..
+//     pool.Release(my_obj_ptr);
+
+#include <cassert>
+#include <memory>
+
+// Clang 15 for macOS 12 and 13 is missing C++17's <memory_resource>, however
+// Clang 15 for macOS 14 has it. When the build system eventually uses macOS 14
+// for all Apple builds then the '#else' portions can be removed. In the
+// meantime we simply fallback to 'new' and 'delete' for this deficient compiler.
+#ifdef HAVE_MEMORY_RESOURCE
+	#include <memory_resource>
+#endif
+
+template <typename T>
+class ObjectPool {
+public:
+	ObjectPool() = default;
+	~ObjectPool() = default;
+
+	// Acquire an object: allocate and construct it with the given arguments.
+	template <typename... Args>
+	constexpr T* Acquire(Args&&... args)
+	{
+#ifdef HAVE_MEMORY_RESOURCE
+		auto ptr = ObjectTraits::allocate(allocator, OneItem);
+		ObjectTraits::construct(allocator, ptr, std::forward<Args>(args)...);
+#else
+		auto ptr = new (std::nothrow) T(std::forward<Args>(args)...);
+#endif
+		assert(ptr);
+		return ptr;
+	}
+
+	// Release the object: destruct it and deallocate its memory.
+	constexpr void Release(T* ptr)
+	{
+#ifdef HAVE_MEMORY_RESOURCE
+		assert(ptr);
+		ObjectTraits::destroy(allocator, ptr);
+		ObjectTraits::deallocate(allocator, ptr, OneItem);
+#else
+        delete ptr;
+#endif
+	}
+
+#ifdef HAVE_MEMORY_RESOURCE
+private:
+	using ObjectAllocator = std::pmr::polymorphic_allocator<T>;
+	using ObjectTraits = std::allocator_traits<ObjectAllocator>;
+
+	constexpr static auto OneItem = 1;
+
+	// We use C++'s off-the-shelf pool resource as the backing memory for the
+	// object's allocator. The allocator is used in the above ObjectTraits::
+	// calls, so when we allocate() and deallocate() an object the memory comes
+	// from the pool (and not the heap).  More details:
+	// - https://en.cppreference.com/w/cpp/memory/allocator_traits
+	// - https://en.cppreference.com/w/cpp/memory/polymorphic_allocator
+	//
+	std::pmr::unsynchronized_pool_resource pool = {};
+	ObjectAllocator allocator{&pool};
+
+#endif
+};
+
+#endif
--- a/src/cpu/dyn_cache.h
+++ b/src/cpu/dyn_cache.h
@@ -19,12 +19,14 @@
  *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
+#include <array>
 #include <cassert>
 #include <cerrno>
 #include <new>
 #include <type_traits>
 
 #include "mem_unaligned.h"
+#include "object_pool.h"
 #include "paging.h"
 #include "types.h"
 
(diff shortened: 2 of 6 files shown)


include/object_pool.h
```diff
@@ -0,0 +1,101 @@
...
+//
+//  Ongoing runtime loop:
+//     auto my_obj_ptr = pool.Acquire(arg1, arg2);
+//     .. do a bit of work ..
+//     pool.Release(my_obj_ptr);
+
+// Note: this is less about performance and more about fitting a well known
+// use-case (frequent and ongoing heap allocation and destruction of a large
+// object type) with a well known solution (object pool).  The before-and-after
+// benefit will be reduced call outs to the heap and less heap fragmentation.
```
johnnovak, line 48: Not to nitpick much, but in practice, this is always done due to performance reasons and to avoid memory fragmentation 😄

johnnovak: Very nice explanatory comments, btw 🎖️

include/object_pool.h
```diff
@@ -0,0 +1,101 @@
...
+
+#include <cassert>
+#include <memory>
+
+// Clang 15 for macOS 12 and 13 is missing C++17's <memory_resource>, however
+// Clang 15 for macOS 14 has it. When the build system eventually uses macOS 14
+// for all Apple builds then the '#else' portions can be removed. In the
+// meantime, we simply fallback to new and delete for this defficient compiler.
```
johnnovak, line 56: typo: deficient

include/object_pool.h
```diff
@@ -0,0 +1,101 @@
...
+#else
+        delete ptr;
+#endif
+	}
+
+#ifdef HAVE_MEMORY_RESOURCE
+private:
+	std::pmr::unsynchronized_pool_resource pool = {};
+	std::pmr::polymorphic_allocator<T> allocator{&pool};
+	using AllocTraits = std::allocator_traits<std::pmr::polymorphic_allocator<T>>;
```
johnnovak, line 97: I have no idea what you're doing here, but if it works, it works 😅 Looking forward to having macOS 14 as the baseline, then we just simply use the stdlib.

src/cpu/dyn_cache.h
```diff
@@ -154,6 +156,32 @@ static uint8_t* cache_code_link_blocks = {};
 static std::vector<CacheBlock> cache_blocks(CACHE_BLOCKS);
 static CacheBlock link_blocks[2] = {}; // default linking (specially marked)
 
+// Use an object pool to manage the invalidation maps
+class InvalidationMapPool {
+private:
+	static constexpr size_t map_size = 4096;
```
johnnovak, line 162: MapSize

johnnovak:
Very nice @interloper98 and that's some cool tips about heaptrack! 🎉 

> Performance might have ticked up by a frame or two on the Pi.

I was kinda expecting a bigger increase, but then that's good news—it means Staging is quite efficient overall already 😎 

> The 'try_emplace' versus 'emplace' avoids creating a temporary object on the heap before emplacing. 

How about `emplace_back`? My understanding was that constructs at the place of destination too, thus avoiding the use of a temporary object. We're using that extensively in the codebase already (but there's still a fair bit of regular `emplace()`). 

I looked at the definition of [try_emplace](https://en.cppreference.com/w/cpp/container/map/try_emplace), but as the saying goes, life is too short for C++ 😅  You seem to be a bit of a C++ expert, and I really don't want to understand all that C++ complexity, I just need the rule of thumb: shall we always use `try_emplace` from now on, or `emplace_back`? Plain `emplace` is best to be avoided, that much is clear.

johnnovak (approved):
Great stuff! It's the usual high-quality work one would expect from @interloper98 at this point. You're spoiling us! 😆

include/object_pool.h
interloper98: > in practice, this is always done due to performance reasons and to avoid memory fragmentation

lol; good point :-) Will update the wording :+1: 

When I saw the numbers, I was thinking that 4 KB maps re-allocated 10's of thousands of times per second was quite a bit churn (over 4 GB ever 10s or so!), even if we're not leaking them.

I guess even the Raspberry Pi hardware and Linux kernel are pretty efficient at handling that, but figured we might as well avoid that if we can.

weirddan455:
>  How about emplace_back? My understanding was that constructs at the place of destination too, thus avoiding the use of a temporary object. We're using that extensively in the codebase already (but there's still a fair bit of regular emplace()).

These are `std::unordered_map` so `emplace_back()` isn't a valid method.  There's no "back" so to speak of.  `emplace_back()` is for `std::vector` and such and is just shorthand for `emplace(blah, my_vector.back())` because `emplace()` takes in a position argument and `emplace_back()` just means put it at the end.

`try_emplace()` seems to be only implemented for hash map type containers solving some inefficiency of the key/value pair you would pass to `empalce()`.  First I'm hearing of it though.

include/object_pool.h
weirddan455: I understand that this has something to do with an allocator but @interloper98 knows more about this than we do :laughing: 

I've never used these things either.

johnnovak:
> > How about emplace_back? My understanding was that constructs at the place of destination too, thus avoiding the use of a temporary object. We're using that extensively in the codebase already (but there's still a fair bit of regular emplace()).
> 
> These are `std::unordered_map` so `emplace_back()` isn't a valid method. There's no "back" so to speak of. `emplace_back()` is for `std::vector` and such and is just shorthand for `emplace(blah, my_vector.back())` because `emplace()` takes in a position argument and `emplace_back()` just means put it at the end.
> 
> `try_emplace()` seems to be only implemented for hash map type containers solving some inefficiency of the key/value pair you would pass to `emplace()`. First I'm hearing of it though.

Heh, cool, thanks for the explanation. I will keep that in mind, but I guess it only makes a difference for performance-intensive code or with _really_ expensive to create objects.

include/object_pool.h
interloper98: The polymorphic memory stuff (the PMR namespace) is template heavy and quite rough on the eyes :laughing: ; but the pieces only fit together in certain ways (otherwise it fails to compile). 

Mostly it lets you customise how a type handles memory when it's allocated, constructed, destructed, and deallocated. The `allocator{&pool}` hooks up the pool resource to the allocator (which is where the re-use comes into play).

All memory 'resources' need to provide standard API hooks for the allocator to use it (otherwise it won't compile). And in our case, we're just using C++'s own pool resource, but I suppose some people write their own resource pools (who knows.. maybe allocating memory from storage or across networks or crazy stuff like that).

All of the above so far only covers allocating and de-allocate bulk memory from a pool for our template type - but to be object oriented, we also need to call the constructor and destructor routines on the re-used chunk of memory. 

With that, our Acquired object will have the old contents. This is where  the `AllocTraits` comes into play. It lets us run the template type's constructor and destructor against the reused block of memory (including passing constructor options to it).

That's about it! Thankfully all the hard work is done under the hood by the C++ libraries, and we're gluing the pieces together! :jigsaw: 

I'll add a couple references to the documentation for these in the code, just for interested readers.

interloper98:
I'm running on the Pi, and saw maybe a 7x reduction overall (allocation + temp allocs, taken together), when I ran Quake 320x200 timedemo.  @weirddan455, curious what you see on your x86 linux system if you want to give it a try :eyes: :slightly_smiling_face:

include/object_pool.h
johnnovak: Thanks 😅 Formidable stuff 😆

weirddan455:
I'm getting a 25x reduction in total allocations and an FPS improvement in Quake from 177FPS to 184FPS (tested at max cycles in a release build).  Very nice @interloper98 

Before:
```
heaptrack stats:
        allocations:            4944497
        leaked allocations:     3571
        temporary allocations:  4771093
```

After:
```
heaptrack stats:
        allocations:            199213
        leaked allocations:     3568
        temporary allocations:  19639
```

johnnovak:
> I'm getting a 25x reduction in total allocations and an FPS improvement in Quake from 177FPS to 184FPS (tested at max cycles in a release build). Very nice @interloper98

That's a quite healthy 5% bump, definitely significant! 🎉 Less memory alloc/dealloc can also help with getting rid of micro-spikes on the main thread that might cause audio glitches (just theorising; I don't have concrete proof).

include/object_pool.h
interloper98: Updated the wording to mention the performance fragmentation benefits :+1:

include/object_pool.h
interloper98: (Added some comments about how the pool and allocator interact, and pointed to the ref docs for more details)

src/cpu/dyn_cache.h
interloper98: Fixed; thanks!

interloper98:
Thanks for sharing those results, @weirddan455. 

> Less memory alloc/dealloc can also help with getting rid of micro-spikes on the main thread that might cause audio glitches

:100: I sure hope.. I'm all for anything that can smooth things out; audio glitches and stutters are still an ongoing problem.

interloper98:
Took a final pass through and also confirmed the sanitizer builds is good on the pi as well. Thanks for the reviews and tests. Merging.

PR #3813, opened 2024-07-09
Fix Ring Buffer PrevIndex() wrapping behavior for non power of 2 size

weirddan455:
# Description
This is currently a compile time constant that is hard-coded to off so there is no visible bug but might as well make this correct.

Wrapping behavior was wrong.  This would result in a crash the next time index would be used as it would wrap to the max value of `size_t`.

# Manual testing
Compiled and tested by setting `#define POWER_OF_TWO_ARRAY_SIZE 0`.  The only place I found that the decrement operator is called is in the ring buffer unit test.  On `main` with power of 2 turned off, the unit test fails with the following:

```
[ RUN      ] RingBuffer.iterator_sub_wraparound
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/array:208: reference std::array<int, 16>::operator[](size_type) [_Tp = int, _Nm = 16]: Assertion '__n < this->size()' failed.
Aborted (core dumped)
```

With this PR, all tests pass.

# Checklist

I have:

### DIFF
--- a/include/ring_buffer.h
+++ b/include/ring_buffer.h
@@ -167,9 +167,10 @@ public:
 			--index;
 			index &= IndexMask;
 #else
-			--index;
-			if (index == array->size()) {
-				index = 0;
+			if (index == 0) {
+				index = array->size() - 1;
+			} else {
+				--index;
 			}
 #endif
 		}


johnnovak (approved):
Nice one @weirddan455. The original code used non-power-of-two sizes, so one of the first things I changed was adding power of two support, then I never looked back.

interloper98:
Nice catch @weirddan455 combined with those thorough unit tests @johnnovak 🤘

Issue #2869, opened 2023-09-15
sbtype: SBPro2, sb_filter doesn't work in DOOM


James-F2:
0.81.0-alpha-1676-gfed6c628d
Windows 10

### A clear and concise description of what the bug is.

DOOM starts with sb_filter disabled.

SETTING:
[sblaster]
sbtype = sbpro2
sb_filter = auto

When I exit the game it also shows "SB: Lowpass filter enabled (12 dB/oct at 3200 Hz)", IIRC (I may not), the real SBPro2 card DISABLED the filter when exiting Doom, not enabled it.
This needs to be confirmed with a real DOS machine and a real SBPro2 but I think the SBPro filter register is flipped in DOSBox.

Also, I have attached a dos tool called sbp-set.exe that can change the SBPro mixer settings, where the DNFI command set the output filter.
Strangely /DNFI:**OFF** enables the filter in DOSBox.... it should disable it.

[SBP-SET.zip](https://github.com/dosbox-staging/dosbox-staging/files/12617416/SBP-SET.zip)

PR #3815, opened 2024-07-11
Fix programmatic toggling of the Sound Blaster Pro 2 output filter

johnnovak:
# Description

Previously, toggling the analog output filter programmatically on the Sound Blaster Pro 2 (`sbtype = sbpro2`) was flipped due to a coding error.

From the official "Sound Blaster Series Hardware Programming Guide" by Creative Labs ([download link](https://pdos.csail.mit.edu/6.828/2018/readings/hardware/SoundBlaster.pdf)), page 67:

image

So the stereo bit was handled correctly (1=stereo, 0=mono), but the filter bit handling was inverted (0 means **ON**, and 1 means **OFF**!).

Nobody noticed this because sound card output filtering wasn't implemented in any DOSBox fork for a long time. DOSBox-X implemented output filtering before I did my own implementation from scratch, but the X code handles the filter bit incorrectly, too (@joncampbell123, you too might wanna fix this).

For people who think PCem/86box are more accurate than Staging: PCem doesn't handle the filter bit at all, but the 86box folks [got it right](https://github.com/86Box/86Box/blob/master/src/sound/snd_sb.c#L876). So that's a +1 for 86box 😎 

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/2869

# Manual testing

The analog output filter can only be programmatically toggled on the Sound Blaster Pro 2. It cannot be toggled on any other model, including the SB Pro 1.

So it's important to use the following config for testing:

```ini
[sblaster]
sbtype = sbpro2
sb_filter = auto
```

Use Creative's `SBP-SET.EXE` utility that came with the SB Pro 2 cards to toggle the filter:
[SBP-SET.zip](https://github.com/user-attachments/files/16173073/SBP-SET.zip)

## Test 1

1. Start Staging (use the above config).
2. Make sure the filter is enabled for startup (look for `SB: Low-pass filter enabled (12 dB/oct at 3200 Hz)` in the logs).
3. Disable the filter with `SBP-SET /dnfi:off`. You'll see the following log entry when using a debug build:

   
    image

4. Enable the filter with `SBP-SET /dnfi:on`. This should be logged:

    image

## Test 2

1. Disable the filter as per above.
2. Start DOOM (eXoDOS version) and confirm the game enables the filter:

    image

3. Exit DOOM and confirm the game disables the filter on exit:

    image

# Checklist

I have:

### DIFF
--- a/include/bitops.h
+++ b/include/bitops.h
@@ -255,7 +255,7 @@ constexpr bool any(const T reg, const unsigned int bits)
 	return reg & static_cast<T>(bits);
 }
 
-// Check if the indicated bits is cleisd (not set)
+// Check if the indicated bits is cleared (not set)
 template <typename T>
 constexpr bool cleared(const T reg, const unsigned int bits)
 {
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -29,6 +29,7 @@
 #include "autoexec.h"
 #include "bios.h"
 #include "bit_view.h"
+#include "bitops.h"
 #include "channel_names.h"
 #include "control.h"
 #include "dma.h"
@@ -2279,6 +2280,8 @@ static uint8_t read_ess_volume(const uint8_t v[2])
 
 static void ctmixer_write(const uint8_t val)
 {
+	using namespace bit::literals;
+
 	switch (sb.mixer.index) {
 	case 0x00: // Reset
 		ctmixer_reset();
@@ -2327,10 +2330,13 @@ static void ctmixer_write(const uint8_t val)
 
 	case 0x0e: {
 		// Output/Stereo Select
-		sb.mixer.stereo_enabled = (val & 0x02) > 0;
+		sb.mixer.stereo_enabled = bit::is(val, b1);
 
 		const auto last_filter_enabled = sb.mixer.filter_enabled;
-		sb.mixer.filter_enabled        = (val & 0x20) > 0;
+
+		// This is not a mistake; clearing bit 5 enables the filter as per
+		// the official Creative documentation.
+		sb.mixer.filter_enabled = bit::cleared(val, b5);
 
 		if (sb.type == SbType::SBPro2) {
 			// Toggling the filter programmatically is only possible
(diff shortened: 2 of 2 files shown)


interloper98 (approved):
Looks good! Helpful comment about how Creative only let the SB Pro 2 toggle the filter. :+1:

johnnovak:
Thanks guys, merging.

joncampbell123:
Actually, despite the name and assignment of sb.mixer.filtered, the code in DOSBox-X does correctly comment that the bit enables a filter *bypass* When I developed that code I used a real Sound Blaster Pro card to compare emulation against to verify that bit's behavior.

I also tested against a Sound Blaster 2.0 and 1.5 card which from my tests have no lowpass filter at all. Everything below 6KHz has that lovely metallic sound to it.

johnnovak:
> Actually, despite the name and assignment of sb.mixer.filtered in the DOSBox-X code, that code does correctly comment that the bit enables a filter _bypass_. When I developed that code I used a real Sound Blaster Pro card to compare emulation against to verify that bit's behavior.
> 
> I also tested against a Sound Blaster 2.0 and 1.5 card which from my tests have no lowpass filter at all. Everything below 6KHz has that lovely metallic sound to it.
> 
> I also have the same documentation in my collection which mentions that: http://hackipedia.org/browse.cgi/Computer/Platform/PC%2c%20IBM%20compatible/Sound%20and%20Music/Creative%20Labs/Sound%20Blaster/Sound%20Blaster%20Series%20Developer%20Kit%2c%20Second%20Edition%2epdf

Indeed, the emulation-side boolean in the X code mimics the flipped SB logic. That threw me off.

```cpp
if (sb.mixer.filtered/*setting the bit means to bypass the lowpass filter*/)
```

Personally, I find that confusing as most people would think this is a bug... unless they find the above comment further down in the code that actually sets the DSP filter.

joncampbell123:
What happened is I initially thought that setting the bit turned on the filter, and *then* I discovered the opposite meaning of the bit and coded that into the filter handling, instead of fixing the assignment.

PR #3823, opened 2024-07-15
Build Windows package for the ARM64 platform

interloper98:
# Description

Noticed someone in eXoDOS with a Snapdragon X Elite system asking about ARM64 Windows support, so figured we should build packages for these users, too.

arm64-artifacts

---

Although most of the content in this PR involves adding ARM64 equivalents to the Visual Studio project targets, other notable points are:
 - Tracy doesn't support or compile on Windows ARM* platforms, so this had to be excluded in `vcpkg.json` just for this OS+platform combination (otherwise it errors out early in the build process).
 - The x86 Windows CI machines can't run the as-built ARM64 unit test executables, so I've left out the debug build job and only added release builds. If/when GiTHub adds real ARM64 Windows CI machines, then it will make sense to add this. In the meantime, looks like macOS CI builds and run ARM64 unit tests, so that's at least some coverage.

# Manual testing

I don't have one of these machines, but will ask that eXoDOS user to try it :-)

# Checklist

I have:


johnnovak:
Good stuff @interloper98, and this paves the way for ARM support on Windows for the future, so it's a welcome enhancement. The change doesn't seem to significantly complicate the CI workflow or anything else, and it doesn't seem to add much to the total CI run time, so that's good too.

However, I have quite firm opinions on our official builds; those need to be tested thoroughly by core team members. So, until at least one of us gets hold of an ARM machine for testing the Windows ARM builds, this can only be an experimental deliverable. We might mention it on the Windows downloads page further down, but as an "experimental no warranties, no support" build for the adventurous.

In the interim, people on ARM Windows boxes can use the built-in x86/x86 emulation of Windows to run Staging:

> Windows 10 enables existing unmodified x86 apps to run on Arm devices. Windows 11 adds the ability to run unmodified x64 Windows apps on Arm devices! This ability to run x86 & x64 apps on Arm devices gives end-users confidence that the majority of their existing apps & tools will run well even on new Arm-powered devices.

That's more than acceptable; I don't imagin Steam, GOG, and similar online gaming marketplaces offering ARM binaries until x86/x64 is the dominant Windows platform, so we'll just follow then when the change happens. We certainly won't be paving the way 😏

interloper98:
Yes, best effort is all would imagine for anything that the team can't test on (or reproduce).

The dragon X elite user in exodos reported poor stuttery performance with the x86 builds, but it cleared up once he switched to the dosbox-x arm64 build.

I think there's a lot of existing overlap. For example, it's using the exact same dynrec core as you get on Apple silicon and the Raspberry Pi (so the core should be functionally identical).

Same with all the non-SDL dependencies; if they have arm64 vs x86 code branches, then those arm paths are already being used on macos and Linux arm systems. 

The leaves the UI specific code branches like in SDLmain and SDL libs in general - which could have some different code paths based on CPU architecture (although I'm not seeing any in the project itself; not sure about SDL). 

I would hope Microsoft's done a good job at making their API behave as close as possible to their x86 implementation as possible.. but you never know. 

Worst case, if a bug can't be reproduced then the user will have to fallback to the x86 build.

johnnovak:
Yeah, I have similar views on all that and good points. It's quite convenient we already have a very well tested dynamic recompiler for ARM—this ensures the longevity of the project.

As you said, in theory, it should be smooth sailing and the ARM builds should JustWork(tm), and you'd expect the vcpkg provided deps to be well-tested and stable. But, the difference between practice and theory is that, in theory, there is no difference 😆 

Curious to hear how the build fares on the eXoDOS user's ARM box!

johnnovak (approved):
Change looks fine to me, great job @interloper98 and thanks for taking the initiative on this 🎉 

Maybe leave it open for a few days to give the other team member a chance to comment.

interloper98:
I'll pass along the results when the user tries the build; until then the PR can hang out ☺️

weirddan455:
Our codebase is quite portable.  We've had people build it some niche platforms like PowerPC/Linux and some nice contributors have fixed bugs there even though no one on the core team has the hardware.

ARM/Windows is going to be more mainstream than that so I think it's good to provide binaries.  An "experimental" label does seem appropriate though.  I'm fine with getting it out there for testing.  We can fix what bugs we can if they're reported.  We might be limited in what we can do without hardware but maybe some new contributors will come along and help us out :smile: 

Nice job getting this to build and run through CI @interloper98   I don't have any comments on the code itself.  It's mostly build system changes which all look fine to me at a glance.

johnnovak:
Probably time to merge this @interloper98 ?

interloper98:
> Probably time to merge this @interloper98 ?

I think so! The person who asked for a native build hasn't replied in 5 days now (I was hoping they'd confirm the build at least runs and doesn't instant-crash).

Oh well; we'll know eventually when someone tries it :construction_worker_man:

PR #3826, opened 2024-07-18
Change ppc64 to use correct `POWERPC` target

Grounded0:
# Description

There's three different Power ISA platforms: powerpc (PowerPC 32-bit), ppc64 (PowerPC 64-bit), and ppc64le  (PowerPC 64-bit little-endian).

Changing target from incorrect `PPC64LE` to correct `POWERPC` for ppc64 platform.

## Related issues

- #3145

# Manual testing

No hardware to test on.

# Checklist

I have:


classilla:
Shouldn't break anything, but I still haven't heard if the JIT even works properly on big-endian PowerPC64.

johnnovak:
> Shouldn't break anything, but I still haven't heard if the JIT even works properly on big-endian PowerPC64.

Good point. On the other hand, there is no hope in hell for the LE build to even start up on BE PPC, am I right? So with the correct LE build people could at least use the non-dynarec cores.

Also, ignore the Windows breakages; those were me building 0.81.2 and forgetting to switch back the Windows workflows.

Grounded0:
PowerPC 64-bit Kernel Internals

https://lwn.net/2001/features/OLS/pdf/pdf/ppc64.pdf

> IBM’s 64-bit implementations of the PowerPC processor architecture provide full 32-bit execution compatibility with no performance loss, a feature that allows the Linux/PPC64 to provide excellent support of existing Linux/PPC32 applications.

Grounded0:
> Good point. On the other hand, there is no hope in hell for the LE build to even start up on BE PPC, am I right? So with the correct LE build people could at least use the non-dynarec cores.

Yes.

johnnovak:
> > Good point. On the other hand, there is no hope in hell for the LE build to even start up on BE PPC, am I right? So with the correct LE build people could at least use the non-dynarec cores.
> 
> Yes. With the correct BE build.

Ah, yes, fixed in my original comment.

Grounded0:
@casey-ac plays Theme Hospital with Power Mac G5 so I would say dynarec definitely works:

- #3144

src/cpu/meson.build
```diff
@@ -23,7 +23,7 @@ core_selection = [
     [ ['aarch64'],                ['auto', 'dynrec'],  'C_DYNREC',      'ARMV8LE', 1,         1 ], # ARMv8+ (64-bit)
     [ ['arm'],                    ['auto', 'dynrec'],  'C_DYNREC',      'ARMV7LE', 1,         1 ], # ARMv7+
     [ ['armv6'],                  ['auto', 'dynrec'],  'C_DYNREC',      'ARMV4LE', 0,         0 ], # ARMv6 or older
-    [ ['ppc64',   'powerpc64'],   ['auto', 'dynrec'],  'C_DYNREC',      'PPC64LE', 1,         1 ], # 64 bit PPC processors
+    [ ['ppc64',   'powerpc64'],   ['auto', 'dynrec'],  'C_DYNREC',      'POWERPC', 1,         1 ], # 64 bit PPC processors
```
runlevel5, line 26: Can't we duplicate POWERPC to `PPC64` or `PPC64BE` to separate the 2 targets?

johnnovak: I think if you want more substantial changes, you'll need to do the work yourself as a PPC user 😄

classilla: Actually, it just dawned on me what's going on. This is in fact specifying the _32-bit_ PowerPC JIT, which is inherently big-endian. But 32-bit instructions will run fine on a 970MP. This is probably the best option for now since I'm still pretty sure the 64-bit JIT isn't necessarily endian-safe, barring any of the weird edge cases we've smoked out of the 64-bit Power dynrec.

johnnovak (approved):
Looks fine to me. Any other substantial PPC-related changes need to be a community effort as the core team has neither the means nor the interest to test Staging on PPC.

Grounded0:
@johnnovak 

Since I'm a maintainer is this "Rebase and merge" option safe or am I going to blow up something?

johnnovak:
> @johnnovak
> 
> Since I'm a maintainer is this "Rebase and merge" button safe to press or am I going to blow up something?

I've just pressed rebase. Let it run the CI workflows again, then let's merge once all green (feel to press the merge button).

Issue #610, opened 2020-09-12
Linux only issue - Win 3.x game "Critical Path" throws "You must install Critical Path. Please run setup .exe"


Tkcmd:
I noticed this issue with the game "Critical Path".
Its a Windows 3.X "multimedia" game written in VB2. If you can't find it try the Win3xO archive.
While this game works perfect with the Windows build of dosbox-staging it doesn't work when using a Linux build.
When trying to launch the game with dosbox-staging under Linux a message box with "You must install Critical Path. Please run setup .exe".
I cant pinpoint the exact issue but I would suspect that there is a difference between the Linux and Windows build regarding how files or the CD image is handled.

dreamer:
@Tkcmd can you share .conf file? How CD image is mounted?

Tkcmd:
Nothing special in the config file, attatched it to the post.
Just using imgmount d -t iso to mount the CD.
[dosbox-staging-git.conf.txt](https://github.com/dosbox-staging/dosbox-staging/files/5214034/dosbox-staging-git.conf.txt)

kcgen:
@Tkcmd , it's possible the autoexec portion in your conf is missing some common statements typically needed to bring up windows:

``` ini
[autoexec]
mount c /home/user/Dos/Critical
imgmount d home/user/Dos/Crticial/cd/critical.bin -t iso
C:
cd windows
win
```

For example, normally the Windows directory is added to PATH, and normally the TEMP variable is set pointing to a writable directory.  

I'm not able to currently access my system, so these are just from memory. Maybe other folks will give a more thorough list.

Tkcmd:
The DOS "environment" has nothing to do with this bug.
The same equivalent autoexec lines work on a windows build of Dosbox make the game work.
As stated, its something with the Linux build.
I'm just not able to pinpoint what causes this difference in behavior between the Linux/Windows build of Dosbox.

kcgen:
Ahh indeed; thanks for reminding me of that distinction @Tkcmd.

Tkcmd:
I've been digging a bit deeper into the issue.
After trying to de-compile the games main executable with an old VB2 de-compiler it looks the error is thrown if one of the following things occur.
If can not read the proper data from the "critpath.ini" file which is located in the c:\windows directory
Or it can not read data from the file "CDDRIVE\lang.mec"
This file access is done by the windows kernel function "GetPrivateProfileString"
[...]

PR #3837, opened 2024-07-25
Minor FPU cleanup

kklobe:
# Description

I wasn't able to fix the Critical Path issue (#610) yet, but I was reminded of a few things in the FPU code that annoyed me. This is a largely non-functional cleanup, but there a few small optimizations and one fix. 

I extended the single precision rounding fix to all rounding modes, instead of just `ROUND_Chop`. This is really only noticeable when logging the output, as I didn't notice any visible differences during e.g. Quake gameplay or any of the demos that heavily use single precision mode, but for consistency it should be done for all modes.

The minor optimizations are to the log-related functions.

The cleanup is mostly `std::` prefixing, some more de-Bitu-ing, and the bigger change of using `std::optional` for the array of the 64-bit FILD/FIST memcpy trick, thus eliminating the corresponding array of `bool` used to track the memcpy state.

# Manual testing

MCPDIAG gives the same result:

CleanShot 2024-07-24 at 21 47 27@2x

 as `main`:

CleanShot 2024-07-24 at 21 48 05@2x

The Quake screen size option slider still works:

CleanShot 2024-07-24 at 21 49 27@2x

I also tested:

- Sunflower
- Carmageddon
- moralhardcandy
- Heaven 7
- Quake (QTD for performance)
- multikolor
- toontown
- Z.A.R. demo (MMX mode)
- matrix (MMX demo https://www.pouet.net/prod.php?which=1100)

# Checklist

I have:

### DIFF
--- a/include/fpu.h
+++ b/include/fpu.h
@@ -20,9 +20,11 @@
 #ifndef DOSBOX_FPU_H
 #define DOSBOX_FPU_H
 
+#include <optional>
+
 #ifndef DOSBOX_DOSBOX_H
-//So the right config.h gets included for C_DEBUG
-#include "dosbox.h"
+	// So the right config.h gets included for C_DEBUG
+	#include "dosbox.h"
 #endif
 
 #ifndef DOSBOX_MEM_H
@@ -89,8 +91,8 @@ enum FPU_Round : uint8_t {
 struct FPU_rec {
 	FPU_Reg regs[9]         = {};
 #if !C_FPU_X86
-	int64_t regs_memcpy[9]  = {}; // for FILD/FIST 64-bit memcpy fix
-	bool use_regs_memcpy[9] = {};
+    // for FILD/FIST 64-bit memcpy fix
+	std::optional<int64_t> regs_memcpy[9] = {};
 #endif
 	FPU_P_Reg p_regs[9]     = {};
 	MMX_reg mmx_regs[8]     = {};
--- a/src/fpu/fpu.cpp
+++ b/src/fpu/fpu.cpp
@@ -78,7 +78,7 @@ void FPU_GetPRegsTo(uint8_t dyn_regs[8][10])
 			*/
 
 static void EATREE(Bitu _rm){
-	const uint8_t group=(_rm >> 3) & 7;
+	const uint8_t group = (_rm >> 3) & 7;
 	switch(group){
 		case 0x00:	/* FADD */
 			FPU_FADD_EA(TOP);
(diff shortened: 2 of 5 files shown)


src/fpu/fpu.cpp
```diff
@@ -117,8 +117,8 @@ void FPU_ESC0_EA(Bitu rm,PhysPt addr) {
 }
 
 void FPU_ESC0_Normal(Bitu rm) {
-	Bitu group=(rm >> 3) & 7;
-	Bitu sub=(rm & 7);
+	const uint8_t group=(rm >> 3) & 7;
+    const uint8_t sub=(rm & 7);
```
johnnovak: indentation

src/fpu/fpu.cpp
```diff
@@ -152,8 +152,8 @@ void FPU_ESC0_Normal(Bitu rm) {
 
 void FPU_ESC1_EA(Bitu rm,PhysPt addr) {
 // floats
-	Bitu group=(rm >> 3) & 7;
-	Bitu sub=(rm & 7);
+    const uint8_t group=(rm >> 3) & 7;
+    const uint8_t sub=(rm & 7);
```
johnnovak: indentation (just reformat the whole file)

src/fpu/fpu.cpp
```diff
@@ -429,8 +429,8 @@ void FPU_ESC4_EA(Bitu rm,PhysPt addr) {
 
 void FPU_ESC4_Normal(Bitu rm) {
 	/* LOOKS LIKE number 6 without popping */
-	Bitu group=(rm >> 3) & 7;
-	Bitu sub=(rm & 7);
+    const uint8_t group=(rm >> 3) & 7;
+    const uint8_t sub=(rm & 7);
```
johnnovak: formatting (spaces around the `=`)

src/fpu/fpu_instructions.h
```diff
@@ -607,8 +607,8 @@ static void FPU_FLDENV(PhysPt addr){
 
 static void FPU_FSAVE(PhysPt addr){
 	FPU_FSTENV(addr);
-	Bitu start = (cpu.code.big?28:14);
-	for(Bitu i = 0;i < 8;i++){
+    PhysPt start = (cpu.code.big?28:14);
```
johnnovak: indentation

include/fpu.h
```diff
@@ -89,8 +91,8 @@ enum FPU_Round : uint8_t {
 struct FPU_rec {
 	FPU_Reg regs[9]         = {};
 #if !C_FPU_X86
-	int64_t regs_memcpy[9]  = {}; // for FILD/FIST 64-bit memcpy fix
-	bool use_regs_memcpy[9] = {};
+	std::optional<int64_t> regs_memcpy[9] = {}; // for FILD/FIST 64-bit
+	                                            // memcpy fix
```
johnnovak, line 95: Minor: can you put the comment above the line while you're at it? 😄

src/fpu/fpu.cpp
kklobe: yeah, seems like the script doesn't quite do the right thing all the time.

johnnovak (approved):
Very nice @kklobe and extra points for the thorough testing 🚀

src/fpu/fpu.cpp
kklobe: oh, wait, looks like you were looking at an outdated commit, nevermind

PR #3838, opened 2024-07-25
Improve redirection handling in the DOS shell

### LINKED ISSUE #3829: X-Wing B-Wing Add-On installation corrupts files

_No response_

macOS (Apple Silicon)

_No response_

### Relevant hardware info

_No response_

When using latest release and DEV versions of Staging, installing the B-Wing Add-On to a fresh X-Wing installation will result in corrupt game files.

The following DOSBox versions are affected (X) or not affected (OK).

DOSBox Staging 0.78.1: OK
DOSBox Staging 0.79.1: OK
DOSBox Staging 0.80.1: OK
DOSBox Staging 0.81.1: X
DOSBox Staging Main: X

DOSBox-X 2024.03.01: OK
DOSBox 0.74 SVN: OK

Floppy disk images from Digitoxins page on Archive.org:
https://archive.org/download/20220917_20220917_0541/Star%20Wars%20-%20X-Wing%20%281993%29%20%28v1.0%29%20%281.44M%29%20%5B%21%5D.zip
https://archive.org/download/20220917_20220917_0541/Star%20Wars%20-%20X-Wing%20-%20B-Wing%20%281993%29%20%28v1.2%29%20%281.44M%29%20%5B%21%5D.zip

Explain how to reproduce

First install X-Wing from above floppy images to default folder using default settings.
Then install B-Wing into the same folder (default) using default settings.

Launch X-Wing using "XWING".

Find copy protection codes here:
https://archive.org/details/x-wing-handbuch

Select the TOPACE4 pilot and enter the main hall, then go top right to the Tour Desk.

If your installation is corrupt, you will see this screen:
Bildschirmfoto 2024-07-19 um 18 30 22

If your installation is OK, you will see this screen:
Bildschirmfoto 2024-07-19 um 11 46 02

Installed files that are OK:
[XWING + BWING DOSBox 0.74.zip](https://github.com/user-attachments/files/16314621/XWING.%2B.BWING.DOSBox.0.74.zip)

Installed files that are corrupt:
[Uploading XWING + BWING Staging corrupt.zip…]()

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

interloper98:
# Description

This fixes and improves the redirection handling in the DOS Shell.

- Adds many more unit tests based on real MS-DOS behavior.
    
- Now generates syntax errors when illegal combinations of redirection characters are used. For example, previously things like: `echo || dir`, `echo |< > hi`, and so on were processed but real MS-DOS genrates a syntax error and returns `ERRORLEVEL` `0`. Now we match this behavior.
    
- Fixes whitespace handling around output targets to match real MS-DOS. (See unit tests for details).
    
- Fixes multiple target handling to match real MS-DOS. For example, `echo hi>1>2>3` should only write to the file 3 but previously we couldn't handle this.
    
- Replaces all of the previous C-string code with a loop that iterates over regular expression matches.
    
- Refactors the function's previous style of modifying its arguments with a return struct of values.

## Related issues

Fixes #3829

# Manual testing

Tested X-Wing + B-Wing installation and compared countless variations of shell redirection handling against real MS-DOS.

Also tested (and ran unit tests) w/ sanitiser builds.

# Checklist

I have:

### DIFF
--- a/src/shell/shell.cpp
+++ b/src/shell/shell.cpp
@@ -23,8 +23,8 @@
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
-#include <list>
 #include <memory>
+#include <regex>
 
 #include "../dos/program_more_output.h"
 #include "../dos/program_setver.h"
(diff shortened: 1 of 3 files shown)


src/shell/shell.cpp
```diff
@@ -54,7 +54,7 @@ std::unique_ptr<Program> SHELL_ProgramCreate() {
 	return ProgramCreate<DOS_Shell>();
 }
 
-DOS_Shell::DOS_Shell()
+DOS_Shell::DOS_Shell() : history(std::make_shared<ShellHistory>())
```
weirddan455, line 57: I'm not sure what the purpose of this is.  The `ShellHistory` object gets created a few lines down from this in the constructor.  With this change, you're creating a redundant object that gets discarded.

We want exactly one `ShellHistory` object so the code is using a `static weak_ptr` to achieve this.  If the `weak_ptr` is empty, that means this shell is either the first shell or all active shells have been destroyed, so we create a new one and store it in the static variable.  Otherwise, we use the existing one stored.

src/shell/shell.cpp
```diff
@@ -61,76 +61,130 @@ DOS_Shell::DOS_Shell() : history(std::make_shared<ShellHistory>())
 	               HELP_Category::Misc,
 	               HELP_CmdType::Program,
 	               "COMMAND"};
-
-	static std::weak_ptr<ShellHistory> global_shell_history = {};
-	history = global_shell_history.lock();
-	if (!history) {
-		history = std::make_shared<ShellHistory>();
-		global_shell_history = history;
-	}
```
weirddan455, line 70: This is the block I was talking about in my previous comment.  Deleting this breaks the shell history for nested shells.  Test by typing `COMMAND` in the DOS prompt to launch a nested shell.  Type `EXIT` to quit from the nested shell.

I believe what you will observe is that when the nested shell exits and gets destroyed, it will over-write the contents of the shell history text file.  When you quit the root shell, it will over-write that file again and the history of commands you typed in while in the nested shell will be lost.

`ShellHistory` always writes to the same text file which is why I had all `Shell` instances share the same `ShellHistory` object by that static variable.

interloper98: Whoops; definitely didn't mean to delete this chunk! I probably tried something early on and it got dragged in with the 'commit -a' 🤦‍♂️ ; thanks for flagging this @weirddan455 - will get this put back.

interloper98: Ok, all back to normal - thanks @weirddan455.

src/shell/shell.cpp
interloper98: (back to original; thanks for catching this @weirddan455.)

src/shell/shell.cpp
```diff
@@ -70,67 +70,129 @@ DOS_Shell::DOS_Shell()
...
+// returns the results as a struct. The results include the redirection targets
+// as well as the processed command line with the targets stripped off.
+//
+// Note that real MS-DOS is quite nuanced in its whitespace handling:
+// - 'echo 1>out.txt' produces a 3-byte file: '1CRLF'
+// - 'echo 1 > out.txt' produces a 4-byte file: '1 CRLF'
+// - 'echo 1 >out.txt ' produces a 5-byte file: '1  CRLF'
+// This behavior is replicated here; see the shell_redirection_tests for more
+// examples.
+//
```
johnnovak, line 83: Excellent comment documenting the quirky MS-DOS behaviour @interloper98 🏅

src/shell/shell.cpp
```diff
@@ -70,67 +70,129 @@ DOS_Shell::DOS_Shell()
...
+std::optional<DOS_Shell::RedirectionResults> DOS_Shell::GetRedirection(
+        const std::string_view line)
```
johnnovak, line 85: Nice use of `optional`, I approve 😄

src/shell/shell.cpp
```diff
@@ -70,67 +70,129 @@ DOS_Shell::DOS_Shell()
...
+	//    ("[^"]*"\s*)  # Group 1: double-quoted text
+	//    |             # -or-
+	//    (             # Group 2: the redirection tokens, either
+	//        [<>|]     # Single character tokens
+	//        |         # -or-
+	//        >>|<<     # Double character tokens
+	//    )
+	//    \s*           # Whitespace after the redirection token
+	//    ([^<>| ]+)    # Group 3: target file or program name
+	//    (\s*)         # Group 4: the target's tail whitespace
```
johnnovak, line 104: Give this man an award, please 💯 Yeah, regexes need to be documented like that, otherwise you're in write-only mode quickly 😏

tests/shell_redirection_tests.cpp
```diff
@@ -39,152 +39,455 @@
...
+TEST(DOS_Shell_GetRedirection, TwoOutputsFirstBothTrailing)
+{
+	constexpr auto line = "echo test>out1.txt >out2.txt ";
+
+	const auto expected_results = RedirectionResults{.processed_line = "echo test  ",
+	                                                 .out_file = "out2.txt"};
+
+	// verified in MS-DOS 6.22
+	test_redirection(line, expected_results);
+}
```
johnnovak, line 493: That's some very thorough test coverage and nice use of the designated initialisers 🚀

johnnovak (approved):
Top-notch work @interloper98, looks good to me! Bonus points for the excellent test coverage 🎖️

src/shell/shell.cpp
interloper98: I was surprised by it! At least the redirection results and unit tests handle this whitespace correctly.

That said, our downstream shell and command-execution code that use the results still strips that whitespace in various places (and trying to fix that opened a bit can of worms.. so I stepped back :laughing: ), so it's a bit pedantic right now :slightly_smiling_face: .. but one more step in the right direction.

src/shell/shell.cpp
interloper98: It was really useful to support the syntax error scenario after seeing real MS-DOS doesn't process any of the content and just bails out on the command entirely.

src/shell/shell.cpp
interloper98: > otherwise you're in write-only mode quickly

That's for sure!

The comment is still not as 'DRY' as I'd like.. I tried hard to document the actual statement on the fly.. like:

```c++
regex(
  R("...") // Group 1 comment
  R("...") // Group 2 comment
  ... 
);
```

But that approach changed the behaviour, so I had to go with the single line.

interloper98:
Thanks for the review @johnnovak! Will merge now :-)

Issue #3836, opened 2024-07-24
Need user testing for USB CDROM Audio on Windows (potential regression)


weirddan455:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

codengine on Discord has reported a regression with CDROM audio on Windows.  I merged a change to the development branch which removes the SDL CDROM dependency and instead uses the Win32 API directly (specifically it uses `DeviceIoControl` calls to read audio data from CDROM and send it to the mixer).  Our latest release (v0.81.2) still uses SDL CDROM and does not contain this change.

I would like to fix this regression for the next release but I need some testing data as I don't have a USB CDROM on hand.  If you are willing, please comment with the following information.

1. Does CDROM audio play on a USB CDROM drive?
2. Is it a regression?  (Broken on a development build and working on v0.81.2)
3. What version of Windows are you using?
4. What game or application did you test with?

If you are a developer, you can do one better and add some log statements to the Win32 CDROM file and let me know what call is failing.  I would start with the `Open` function to see if the drive is able to be opened.  Relevant code is located here:  https://github.com/dosbox-staging/dosbox-staging/blob/main/src/dos/cdrom_win32.cpp

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Mount the CDROM by `mount d d:\ -t cdrom`.  The first letter being the DOS drive and the 2nd being the Windows drive.
2. Launch a game/application which uses CDROM audio.
3. Test on a development build and on v0.81.2.
4. Post your results here as a comment.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

FeralChild64:
I definitely have the USB optical drive, just need to find it in my closet - I hope it still works. I should be able to do some testing during the weekend (not sure if I find the time today).

FeralChild64:
OK, found time for some small test:

- **Windows 11**
- **Alone in The Dark 3** (pressed cover CD from the Polish magazine _CD-Action_, 30th November 1998)
- **DOS Navigator OSP 6.4.0** (main menu: Utilities -> CD Player)

-----

### DOSBox Staging 0.81.2

I have used the _-usecd_ parameter to point the emulator to proper drive.

- **Pioneer BDR-2209** (internal, SATA)
[...]

PR #3840, opened 2024-07-26
Fix CD audio support on physical USB drives on Windows

FeralChild64:
# Description

There is a problem with CD audio support under Windows if using a physical USB CD drive (SATA drives work fine), the CD audio does not play at all. Cause:
- IOCTL used to read audio tracks only accepts requests up to certain number of frames per call, value depending on the drive interface used
- previous implementation used a constant value which was thought to be good for any CD drive, but was found to be too high for USB drives

This PR lowers the value. Some auto-detection routines were discussed, but they were considered too complex. Example code attached below, in case anyone needs it in the future:

[old_solution.tar.gz](https://github.com/user-attachments/files/16402733/old_solution.tar.gz)

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3836

# Manual testing

I have used Windows 11 and different CD drives (change is irrelevant for non-Windows platforms).

Insert _Alone in the Dark 3_ game CD, mounted it using command `mount D X: -t cdrom`, try to play the game a little, try to play audio tracks using _DOS Navigator OSP 6.4.0_ (from the main menu  select: _Utilities_ -> _CD Player_).

Make sure to test both cases:
- audio track plays till the end and only then another one starts playing
- audio track is interrupted and another one starts playing

# Checklist

I have:

### DIFF
--- a/src/dos/cdrom_win32.cpp
+++ b/src/dos/cdrom_win32.cpp
@@ -356,7 +356,11 @@ bool CDROM_Interface_Win32::HasDataTrack() const
 std::vector<int16_t> CDROM_Interface_Win32::ReadAudio(const uint32_t sector,
                                                       const uint32_t frames_requested)
 {
-	constexpr uint32_t MaximumFramesPerCall = 55;
+	// According to testing done so far:
+	// - 55 is the maximum for SerialATA drives
+	// - 27 is the maximum for USB drives
+	// Higher values makes the IOCTL_CDROM_RAW_READ fail.
+	constexpr uint32_t MaximumFramesPerCall = 27;
 	const uint32_t num_frames = std::min(frames_requested, MaximumFramesPerCall);
 
 	std::vector<int16_t> audio_frames(num_frames * SAMPLES_PER_REDBOOK_FRAME);


src/dos/cdrom_win32.cpp
```diff
@@ -356,34 +359,92 @@ bool CDROM_Interface_Win32::HasDataTrack() const
...
+			    	if (!already_warned) {
+					LOG_ERR("CDROM_IOCTL: Valid number of frames changed");
+					already_warned = true;
+				}
+			}
+		} else {
+			// Some other error, audio frames should contain silence
+			return audio_frames;
+		}
+	}
```
johnnovak, line 447: Are there any drawbacks to setting this to a know-good low-enough value, or 27, or day even going lower to 20?

I guess I prefer the simplest solution that does the job 😄

FeralChild64: This was one of my proposals, but...

Problem 1: We have no idea what the safe minimum is, we only have test results from my system, no clear documentation, no insight how the driver stack is organized, no OS source code access. I can fully imagine the maximum being even lower on some more exotic USB controllers, ARM laptops, etc.

Problem 2: I have almost 0 experience with the desktop Windows programming, but on Windows CE I have seen WinAPI calls sporadically taking a lot of time to return, without any obvious reasons. So I would prefer to limit the amount of calls if we can.

@weirddan455 What do you think? If you think it's safe, I'm perfectly fine with dropping this solution and just using lower frame limit (maybe with some LOG_ERR to immediately draw attention to the proper place if the IOCTL fails).

interloper98: One thought is that we should try to avoid large and infrequent "bursty" reads and instead try for more frequent and smaller "smooth" reads, because the former will cause fast drives to spin up.

Listening to music with a drive running at 1x is a lot more pleasant than a drive running at 16x or faster.

The original CDROM drives only spun at a cool and quiet 1x (~150 KB/s). To keep up with realtime, we only have to read 75 CDDA frames per second or about one every every 13 ms.

With @weirddan455's threading and queuing, maybe we should try for just 1 frame per read as this is probably the closest read profile to encourage the drive to relax down to it's lowest rotation speed. (this is just a hunch.. unfortunately I don't have hardware to try it out).

FeralChild64: @interloper98 This is indeed a valid point to consider.

I have tested both my USB drives, but I could hear no noise pattern change when reading just a single frame. With both devices the drive spins up and slows down repeatedly. Probably the only think I have changed was to empty the internal OS buffers in multiple calls vs in one go.

As for the internal drive, I can't tell, I have a rather power hungry CPU (which means the fans make quite some noise), and the drive is not one of the dirt-cheap ones, so the disc doesn't vibrate.

interloper98: > With both devices the drive spins up and slows down repeatedly. Probably the only think I have changed was to empty the internal OS buffers in multiple calls vs in one go.

Nice tests @FeralChild64 ! Yes, that yo-yo'ing is painful but it sounds like number of frames isn't going to get that under control on it's own. 

I had one of those 'Mad Dog 70x' drives in the early 2000s and it was my mission to find applications to lock that thing to it's slowest rate of something like 8x or 12x.. because yo-yo'ing from 70x to idle on every game read was madness! I guess that's why it was called the 'Mad Dog'.  :man_facepalming: 

> Manipulating the CD-ROM speed has to be done using special IOCTL’s - on Windows these are IOCTL_CDROM_GET_PERFORMANCE / IOCTL_CDROM_SET_SPEED, on Linux these are CDROM_SELECT_SPEED

Excellent. Some logic could be: set the drive speed to the slowest if it's an audio-only CD (no data track), otherwise (data or mixed-mode) set the speed to something reasonable for the 90s DOS era, like 8x or 12x.

Grounded0: 8x would be good since its still CAV:

drive

FeralChild64: https://github.com/dosbox-staging/dosbox-staging/issues/3842 - now let's back to the main discussion here.

interloper98: Based on your results comparing multi-frame vs single frame, my vote would be for a frame size of 1 given we're dealing with audio (which is the lowest possible CD rate) and given the thread + queue is keeping our outbound buffer full.

Even if we're draining an intermediate buffer (driver or OS), fetching one at a time gives the buffer a change to refill from the hard at various levels (say, 50% full).

weirddan455: If we can find a "known safe value" that works for all drives, that would be the simplest solution.  I've come up empty so far trying to research that though.

@FeralChild64 's solution seems reasonable, though IMO it is a bit noisy having all this binary search logic inside the `ReadAudio()` function.

Possible alternative:  Can we do a test to determine max frames a drive can read inside `Open()` by reading the first audio track?  I have a hunch that the Win32 call will fail immediately if the `SectorCount` value is out of range.  If that's the case, we can do a linear test starting at some maximum (55 or higher if you think some drives can handle it) and decrementing by 1 if that doesn't work.  Take the first value that succeeds and store that as the maximum to use in `ReadAudio()`.

This would simplify the code quite a bit and we only have to test once when the drive initially gets mounted.  Assuming the test doesn't take too long to execute, I think it would be viable.

FeralChild64: We would need to read the TOC and determine a suitable audio track for testing. Maybe it would be easier to do this within the first ReadAudio call?
At this point we at least know where the first audio track starts.

interloper98: Anything in the sub-second range of buffering is reasonable (4 KB to ~128 KB ), because the hardware can hand that back to us with just a couple revolutions of the media, and these small chunks are unlikely to cause big latency stalls on the drive.

(So I'll modify my vote for anything that gives us those :-)

(The flip side would be searching for the maximum buffer with a cap: _"Hey, my BD drive support 256 MB reads!"_, which is the high-latency dilemma the network industry got itself into with runaway [bufferbloat](https://en.wikipedia.org/wiki/Bufferbloat). The concern is when events in the game change and you need to seeking to another chunk of audio or switching tracks - refilling several MB of buffer hurts latency).

weirddan455: >  We would need to read the TOC and determine a suitable audio track for testing. Maybe it would be easier to do this within the first ReadAudio call?

I already am reading the TOC inside `Open()` as kind of an arbitrary ioctl to test whether the requested drive is actually a CDROM drive.  It would probably be fine to test in the first `ReadAudio()` call as well but we don't have TOC information stored there either.  Just a sector number which could theoretically be at the end of the disk with a read request less than our maximum.

FeralChild64: OK, indeed. Is there a reason why isn’t the TOC cached?

weirddan455: >  OK, indeed. Is there a reason why isn’t the TOC cached?

Simplicity I guess.  We don't need the TOC for anything internal.  We just read it on demand when the DOS application requests track information (which is fairly infrequently I believe).  For the frequent `ReadAudio()` calls, all we need is a sector number and a length.

FeralChild64: But the routine to convert the values from TOC to start track sector is somewhere else, right? The `CDROM_Interface_Win32` class does not seem to have this knowledge (which is good, since this code should be shared across the platforms).

I would prefer not to do any refactoring here.

weirddan455: >  But the routine to convert the values from TOC to start track sector is somewhere else, right?

All of that information is in the TOC.  See `GetAudioTracks()` and `GetAudioTrackInfo()` for how it is extracted.  The Win32 call does return track information in Minute/Second/Frame format rather than sector.  There's some code in the MSCDEX layer to convert that though (on the DOS side, programs can request in either format).

https://github.com/dosbox-staging/dosbox-staging/blob/489f43010fba7c8ac1d11a1f7981c9da99fe989e/src/dos/dos_mscdex.cpp#L490-L500

There's a chance you can get away without fussing with that though.  All we care about for the test is what `SectorCount` values the CDROM driver will accept.  If the `IOCTL_CDROM_RAW_READ` call will succeed reading from a data track, we can just test reading from the first sector on the disc.

johnnovak: Ok, so how about we aim for simplicity and hardcode the value to 1 and see how that goes? It cannot get safer than that and reading at redbook audio rates is rather undemanding. I can also try to test it with USB drives, I think I have two of them somewhere. Then we reconsider this if people report issues with CD Audio .

Yeah I always used a CD speed throttler utility in Windows because I hated the hoover sound when playing games on 32x-40x drives 😆

FeralChild64: Exactly. @weirddan455, what you are proposing, far exceeds the effort I am willing to spend here. I’m not digging into MSCDEX (I have already too many complex or time consuming topics started), I am not doing any non-trivial refactoring here, and I am definitely not willing to take the risk to read some random sectors using IOCTL which is intended by Microsoft to read the CDDA tracks only.

I can either do something similar to this PR, or just reduce the max number of sectors read. Anything more = someone else’s job :)

weirddan455: It was just a suggestion.  You don't need to dig into MSCDEX or refactor anything.  The conversion from MSF to Sector was all I was pointing out with that MSCDEX function.  I could write a simple `TestMaxSector()` function and it wouldn't be that much code.  I don't mind doing it if you want to hand this off to me.

>  Ok, so how about we aim for simplicity and hardcode the value to 1 and see how that goes?

Or just do this.  I'm slightly concerned if slower drives can keep up.  I did some timing on this back in my original single-threaded implementation.  A value of 1 is a lot more inefficient than, say 20.  More total time is spent inside the IOCTL calls (presumably due to overhead of making 20x more system calls).  Going from 55 down to 20 is a lot less noticeable.

>  I agree with @interloper98 on bursty reads being potentially problematic.

Bursty reads were problematic in the single-threaded implementation.  They would stall the main thread for too long so I had to lower the sector count from the maximum.  It's not really a problem with the way this threaded reader is implemented.  All we really care about is throughput which is why I raised the count back to maximum.  If the drive can keep up reading audio in real-time, everyone's happy.

weirddan455: tl;dr on my last comment:  I'm fine with lowering the value to 1 and re-evaluating later if someone complains.

interloper98: Threaded queue, low-latency backlog, and drive RPM reduction (in @FeralChild64's other ticket) is going to be the deluxe way to play DOS CD games :rocket:

weirddan455: Latency isn't really a concern with CDROM audio in general.  The `RWQueue` uses a large 5 second buffer just because I can and it helps with intermittent spikes in read time.  That gets filled with a bursty read when a track starts playing (and it still will with reading a single frame at a time, it'll just take more syscalls to do it).

The mixer doesn't wait until the buffer is fully filled to start playing though.  It'll start as soon as the queue has a minimum number of frames (usually 1-10ms worth of audio).  On track change, the queue gets immediately cleared so it doesn't wait to fully drain either.

johnnovak: > Threaded queue, low-latency backlog, and drive RPM reduction (in @FeralChild64's other ticket) is going to be the deluxe way to play DOS CD games 💿 🚀

Yeah, lot of effort for something that's well on the way out (physical CD-ROM drives), but why not if it doesn't complicate the code too much 😆

Amaroq-Clearwater: Optical drives. _My precious…_

FeralChild64: Reduced the value to a safe one, dropped the guessing code.

PR #3854, opened 2024-08-03
Centralize app_id handling (continuation)

FeralChild64:
# Description

This is the continuation of PR https://github.com/dosbox-staging/dosbox-staging/pull/3817 by [ReillyBrogan](https://github.com/ReillyBrogan). Changes done by me:
- handled review remarks which were still open
- updated some copyright strings
- improved comment formatting in the `dosbox.h` file

## Original description:

On modern Linux it is preferred to use the fully qualified application ID for naming the desktop, icon, and metainfo files. Add a meson option to capture this variable and set it to `org.dosbox-staging.DOSBoxStaging` which better matches that the homepage is dosbox-staging.org.

Then ensure that we're setting the X11 `resourceclass` and the Wayland app_id to this value. Those values need to match the name of the .desktop file in order for compositors/window managers to correctly associate windows to the appropriate .desktop file (see [here](https://github.com/zed-industries/zed/issues/12707#issuecomment-2168742761) for an explanation I previously wrote on how that works). Now that we're ensuring that both X11 and Wayland should have correct window associations we can drop `StartupWMClass` from the desktop file as it is now unnecessary.

Also, since SDL3 changed how that is done I added some #if statements so that that's ready whenever dosbox-staging is ported to SDL3.

# Manual testing

- compiled the project under _Kubuntu_ running inside _VirtualBox_ (I didn't want to potentially mess with my main system), executed `meson install`, checked that the menu system correctly recognizes icons
- using command `xprop -spy` I've checked the `WM_CLASS` looks correctly

(please consider I'm not a Linux packaging specialist, so my testing was limited)

# Checklist

I have:


contrib/linux/install-icons.sh
```diff
@@ -25,14 +25,16 @@ usage() {
...
+    cp desktop/org.dosbox-staging.dosbox-staging.desktop $DEST_PATH/applications/org.dosbox-staging.dosbox-staging.desktop
 
     ## Replace executable location in the .desktop file
-    sed -i 's\Exec=dosbox\Exec='$(realpath dosbox)'\g' $DEST_PATH/applications/dosbox-staging.desktop
+    sed -i 's\Exec=dosbox\Exec='$(realpath dosbox)'\g' $DEST_PATH/applications/org.dosbox-staging.dosbox-staging.desktop
 }
 
 remove_icons() {
     find $DEST_PATH/icons -name dosbox-staging* -delete
+    rm $DEST_PATH/applications/org.dosbox-staging.dosbox-staging.desktop
```
johnnovak, line 36: Maybe extract `org.dosbox-staging.dosbox-staging.desktop` to a var?

johnnovak (approved):
Looks fine to me @FeralChild64

contrib/linux/install-icons.sh
```diff
@@ -25,14 +25,16 @@ usage() {
...
-    cp desktop/dosbox-staging.desktop $DEST_PATH/applications
+    cp desktop/org.dosbox-staging.dosbox-staging.desktop $DEST_PATH/applications/org.dosbox-staging.dosbox-staging.desktop
 
     ## Replace executable location in the .desktop file
-    sed -i 's\Exec=dosbox\Exec='$(realpath dosbox)'\g' $DEST_PATH/applications/dosbox-staging.desktop
+    sed -i 's\Exec=dosbox\Exec='$(realpath dosbox)'\g' $DEST_PATH/applications/org.dosbox-staging.dosbox-staging.desktop
 }
 
 remove_icons() {
     find $DEST_PATH/icons -name dosbox-staging* -delete
```
weirddan455, line 35: Same comment I made on the original PR.  This no longer deletes icons since the names have changed.

johnnovak: Fair point.

contrib/linux/install-icons.sh
FeralChild64: OK, changed.

contrib/linux/install-icons.sh
FeralChild64: I think I misunderstood the comment - to fix the problem I have added lines:

```bash
    ## Also remove legacy files, if present
    rm $DEST_PATH/applications/dosbox-staging.desktop
```

to the `remove_icons()` function. AFAIK this `sed` command does not remove eny files, it just updates the content of the installed `.desktop` file. Shall I remove legacy files here, too? I don't really know what are the exactl use cases of this script, unfortunately.

weirddan455: The script is meant to be run by end users that download the tarball from our website.  It copies the icon files and .desktop file to either `~/.local/share/` or, with the system flag `/usr/share/applications`.  This installs a shortcut and is also needed by Wayland to show the Staging icon in the task bar (Wayland doesn't have a way to programmatically set this unlike X11 so it looks for a .desktop file that matches the executable).

These files are re-named in this PR so the remove function will not work.  You've got the .desktop part handled correctly but not the .png files.

FeralChild64: OK, now I understand. Fixed.

weirddan455 (approved):
This looks good to me now as far as I can tell.

interloper98 (approved):
It's too bad we can't use a common list + for loop to grind through all the resolutions, but it seems every OS and desktop cares about a slight different set! Oh well.
 
No concerns from me; looks simple and maintainable. If it helps more Linux users get on board, then that's great.

johnnovak (approved):
Thanks @FeralChild64 @interloper98 and @weirddan455 , merging.

yselkowitz:
The problem with this name is that it is not "legal" in some uses, particularly flatpaks:

https://github.com/flatpak/flatpak-builder/issues/37
https://specifications.freedesktop.org/desktop-entry-spec/latest/file-naming.html
https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

FeralChild64:
@yselkowitz OK, thanks for the info, I’ll look into this.

PR #3864, opened 2024-08-09
Handle SDL2 like the other dependencies

interloper98:
# Description

Previously when requesting static linking (`-Ddefault_library=static`) and using the as-versioned deps (`-Dwrap_mode=forcefallback`), only macOS actually got SDL versioned and statically linked, while non-macOS systems were left with whatever version the system provided.

This PR corrects this exception and lets SDL be compiled statically and as-versioned when requested (just like other deps).

I stumbled on this when @weirddan455 mentioned:

> This wrap only works on Macs so I have no way of testing this.

(I also couldn't try it on my Debian-based raspberry pi system.)

Besides this consistency correction, this PR also lightens the pantload of unused features in SDL, which produces a quicker starting executable. For example, the round trip start-to-exit time (`dosbox -c exit`) was cut into a quarter (0.4s verus 1.6s) on my system.

The statically linked SDL still performs runtime linking to detect host drivers and interfaces both on Linux and macOS, for example:

```
sdl2| Run-time dependency gl found: YES 1.2
sdl2| Run-time dependency glesv2 found: YES 3.2
sdl2| Run-time dependency egl found: YES 1.5
sdl2| Run-time dependency glx found: YES 1.4
sdl2| Run-time dependency alsa found: YES 1.2.6.1
sdl2| Run-time dependency libpulse-simple found: YES 15.99.1
sdl2| Run-time dependency jack found: YES 0.125.0rc1
sdl2| Run-time dependency libpipewire-0.3 found: YES 0.3.48
sdl2| Run-time dependency vulkan found: YES 1.3.204
sdl2| Run-time dependency x11 found: YES 1.7.5
sdl2| Run-time dependency xext found: YES 1.3.4
sdl2| Run-time dependency xi found: YES 1.8
sdl2| Run-time dependency xkbcommon found: YES 1.4.0
sdl2| Run-time dependency xrandr found: YES 1.5.2
sdl2| Run-time dependency xfixes found: YES 6.0.0
sdl2| Run-time dependency xcursor found: YES 1.2.0
sdl2| Dependency threads found: YES unknown (cached)
sdl2| Run-time dependency dbus-1 found: YES 1.12.20
sdl2| Run-time dependency ibus-1.0 found: YES 1.5.26
sdl2| Run-time dependency wayland-client found: YES 1.20.0
sdl2| Run-time dependency wayland-cursor found: YES 1.20.0
sdl2| Run-time dependency wayland-egl found: YES 18.1.0
sdl2| Run-time dependency wayland-protocols found: YES 1.25
sdl2| Run-time dependency wayland-scanner found: YES 1.20.0
sdl2| Run-time dependency libdecor-0 found: YES 0.1.0
```

And on macOS:

```
Run-time dependency appleframeworks found: YES (AudioToolbox, AudioUnit,
 Carbon, Cocoa, CoreAudio, CoreFoundation, CoreGraphics, CoreServices,
 ForceFeedback, Foundation, IOKit, QuartzCore)
```

Even if there are unforseen differences (or bugs) caused this the static SDL versus repo-provided SDL, the good news is that this is all opt-in: the user needs to ask for this when compiling. The default `meson setup` will continue using the system-provided lib, and repo maintainers are already forcing the use of all system-provided shared libs. 

# Manual testing

Testing video drivers (x11 and wayland), sound drivers (pulse and pipewire), joystick, keyboard, mouse, and texture and opengl output modes.

# Checklist

I have:


interloper98:
For those on Linux and macOS, try it with:

```shell
meson setup -Ddefault_library=static -Dwrap_mode=forcefallback build/all-static-deps

meson compile -C build/all-static-deps
```

Then benchmark round-trip startup: `time dosbox -c exit` against a `main` release  build.

kklobe:
I can never keep it straight - aren't Linux builds supposed to always use system dependencies through a package manager instead of vendoring the source code?

kklobe:
@Burrito78 can you make sure this build runs on your macOS 10.15 setup?

interloper98:
> I can never keep it straight - aren't Linux builds supposed to always use system dependencies through a package manager instead of vendoring the source code?

Yes; that's the package maintainers goals - avoid static builds, avoid version pinning, and always use repo deps.

But from what I see here, we provided self-contained binaries using with static linking (and building the specific versions of deps that are pinned in place through wraps) on macOS and Linux, and vcpkg on Windows (well, and ship the .dlls).

For example, the Linux and macOS releases from CI have built-in FluidSynth, MT-32, IIR, and so on.  Even the SDL are built and linked in the macOS binary, however SDL wasn't being hadnled the same for the Linux binary. This PR fixes that, so the Linux binary from CI is just like the macOS CI -- totally self-contained.

weirddan455:
The static build works for me.  We had a small discussion about this on Discord where I tried hacking on `meson.build` to get this to work.  Turns out the reason Wayland didn't work for me is I was missing the `wayland-protocols` package which is a compile-time dep for SDL.  Installed that and now everything seems to work.

I don't see a measurable difference in start-up time on my system vs a dynamic build.  It's all in the ballpark of about 0.5s for me.

Dynamic SDL build with wrap fallback doesn't work for me though.  Compile with `meson setup --force-fallback-for=sdl2 build/release`.  SDL compiles successfully but I still get linked to my system SDL.  It's supposed to compile a `libsdl.so` in the build directory and dynamically link to that.  This method works for other libraries like MT-32 and Fluidsynth.

interloper98:
> OK gotcha. The only reason I did that special handling for macOS was to force compiling from source to enable targeting old versions of the OS for the 0.81 release (homebrew versions of SDL are targeted at the latest OS version); my intent at the time was to pretty much leave every other OS alone, so this is a welcome improvement.

:+1: Makes sense; minimal impact :-)

interloper98:
> I don't see a measurable difference in start-up time on my system vs a dynamic build. It's all in the ballpark of about 0.5s for me.

That's snappy (and good to know)! 

I'm on Raspbian that uses Debian's repos - and the shared SDL lib take up to 4 seconds to initialise the video drivers. DOSBox Staging appears to hang for that time. :laughing: Now with the local build it pops to life really quickly.

> Dynamic SDL build with wrap fallback doesn't work for me though. Compile with `meson setup --force-fallback-for=sdl2 build/release`. SDL compiles successfully but I still get linked to my system SDL. It's supposed to compile a libsdl.so in the build directory and dynamically link to that. This method works for other libraries like MT-32 and Fluidsynth.

Thanks - will look into this!

weirddan455:
>  I'm on Raspbian that uses Debian's repos - and the shared SDL lib take up to 4 seconds to initialise the video drivers.

I'm on plain old Debian on an x86 system and have never seen a hang that long.  Maybe something with the Pi's video drivers? :man_shrugging:

interloper98:
> I'm on plain old Debian on an x86 system and have never seen a hang that long. Maybe something with the Pi's video drivers? 🤷‍♂️

very likely; it's pretty slow overall, too.

> SDL compiles successfully but I still get linked to my system SDL. It's supposed to compile a libsdl.so in the build directory and dynamically link to that. This method works for other libraries like MT-32 and Fluidsynth.

Here's what I get from `meson setup --force-fallback-for=sdl2 build/release`, **`ldd build/release/dosbox`**:

```text
libSDL2_net-2.0.so.0 => /lib/aarch64-linux-gnu/libSDL2_net-2.0.so.0
libSDL2-2.0.so.0 => /usr/src/dosbox-staging/build/release/subprojects/SDL2-2.30.6/libSDL2-2.0.so.0
```

In my case when I launch `build/release/dosbox`, I'm actually getting the `2.30.6` versions built from wraps.  So it's behaving as it should, at least for me.

Maybe `rm -rf build/release` and try from scratch, and then do an `ldd build/release/dosbox`? (Hopefully it looks similar to mine above).

weirddan455:
`meson setup --force-fallback-for=sdl2,sdl2_net build/release` - This worked.  Odd thing is though, this command statically linked SDL.  I see no references to SDL in `ldd` and I see it output `build/release/subprojects/SDL2-2.30.6/libSDL2-2.0.a`.

This is good enough for my purposes though.  My use-case would be testing things with a newer SDL version than my 2.26.5 system lib from Debian.   I just wanted to point out it's odd that you get a dyn lib while I get a static lib.

>  at runtime you'll have both your compiled SDL2 lib and the host SDL2 lib active (via SDL_net linkage)

Isn't this a problem?  I'm pretty sure I've run into run-time crashes when you get duplicate symbols like this.  Usually from glib because that library is a pain.

johnnovak:
@weirddan455 @interloper98 Just to clarify, do you guys want this Linux-only change to be part of 0.82.0? I'm fine with it as it doesn't affect Windows and macOS.

The startup time and memory consumption difference is rather uninteresting, IMO. The big thing is SDL version pinning, so we can provide tested and approved binaries on Linux too. At the very least, that should give distro maintainers a benchmark so they know with certainty if some weird issue is an SDL bug/regression or not.

johnnovak (approved):
Looking good @interloper98, very nice explanatory comments in the comits! 🎉

weirddan455:
>  Just to clarify, do you guys want this Linux-only change to be part of 0.82.0?

Probably a good idea.  This gives us probably as close as we can get to a full static build.  The only things we're dynamically linking now are:

- libGL - Probably needs to stay dynamic linked as this depends on video drivers.
- ALSA - Only used for MIDI output I think but also kind of a core system lib on Linux.
- libm, libc, lidstdc++, libgcc - Standard C++ library stuff that everyone should have installed.

johnnovak:
> > Just to clarify, do you guys want this Linux-only change to be part of 0.82.0?
> 
> Probably a good idea. This gives us probably as close as we can get to a full static build. The only things we're dynamically linking now are:
> 
>     * libGL - Probably needs to stay dynamic linked as this depends on video drivers.
> 
>     * ALSA - Only used for MIDI output I think but also kind of a core system lib on Linux.
> 
>     * libm, libc, lidstdc++, libgcc - Standard C++ library stuff that everyone should have installed.

Yeah, these can be considered essential OS-level services; statically linking these would be indeed quite silly. I'd never propose that 😆 

I don't consider SDL as an essential service, it's just a wrapper over platform-specific APIs to aid development. It's just another non-system-level library like FluidSynth, libmt32, etc. (Yeah, I know the "standard" Linux mentality is to treat every package and library as "part of the OS". Let's just say I disagree with that notion on practical grounds 😆 )

interloper98:
> meson setup --force-fallback-for=sdl2,sdl2_net build/release - This worked. Odd thing is though, this command statically linked SDL. I see no references to SDL in ldd and I see it output build/release/subprojects/SDL2-2.30.6/libSDL2-2.0.a and no .so file.

Very interesting. Perhaps one difference between our systems is I've uninstalled all the SDL system `-dev` packages (because I don't need them anymore :slightly_smiling_face:); and I actually get two shared libs from Meson.

> This is good enough for my purposes though. My use-case would be testing things with a newer SDL version than my 2.26.5 system lib from Debian without having to compile all deps. 

Same here; I can reinstall the SDL repo libs and use them or switch to our pinned version.

> I just wanted to point out it's odd that you get a dyn lib while I get a static lib.

Thanks - and good to know. Just thinking out loud.. maybe Meson is detecting some conditions (beyond our control) and deciding to link it statically for you. I'm not really sure. The only adjustment we have in the dep is `static: <bool>`, and I've added the same criteria as all the other deps.

> at runtime you'll have both your compiled SDL2 lib and the host SDL2 lib active (via SDL_net linkage)

> Isn't this a problem? I'm pretty sure I've run into run-time crashes when you get duplicate symbols like this. Usually from glib because that library is a pain.

I'm able to link and run without crashes any warnings about duplicate symbols. At least at the meson-level, we're not telling meson to use two SDL libs to satisfy our top-level SDL needs, which I can imagine would indeed cause a duplicate symbol explosion (instead, in meson.build `sdl2_dep` is only assigned one provider). So there's probably some safety there (but I'm just guessing!). 

> > Just to clarify, do you guys want this Linux-only change to be part of 0.82.0?

> Probably a good idea. This gives us probably as close as we can get to a full static build.

I also agree because it pins SDL to the version the team has tested with for a given release. From a project perspective, it's nice knowing that SDL's underlying  functionality won't be shifting beneath the binary as the months go by.

weirddan455:
>  Yeah, I know the "standard" Linux mentality is to treat every package and library as "part of the OS".

That entirely depends on your perspective.  For our purposes where we ship a single Linux binary that we want to run everywhere, static linking is the way to go.  If you're targeting one specific distro and want to integrate with its package manger, dynamic linking makes sense.

johnnovak:
@interloper98 Not sure if there's anything outstanding, please feel free to merge when you guys are happy with it.

kklobe:
> @interloper98 Not sure if there's anything outstanding, please feel free to merge when you guys are happy with it.
> 
> I think I should mention in the release notes that the Linux builds now statically link SDL too.

Please wait for @Burrito78 to confirm that it runs on his older macOS setup.

johnnovak:
> > @interloper98 Not sure if there's anything outstanding, please feel free to merge when you guys are happy with it.
> > I think I should mention in the release notes that the Linux builds now statically link SDL too.
> 
> Please wait for @Burrito78 to confirm that it runs on his older macOS setup.

Sounds like a good idea 👍🏻

Burrito78:
Will test it later today!

Burrito78:
Tested build
https://github.com/dosbox-staging/dosbox-staging/actions/runs/10310317879

Result: Works as expected!

Screen Shot 2024-08-09 at 10 13 46 pm

```
Last login: Fri Aug  9 22:14:19 on ttys000
/Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox ; exit;
alex@Alexanders-Mac-mini ~ % /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox ; exit;
date       time         | 
2024-08-09 22:15:08.658 | arguments: /Applications/DOSBox\ Staging.app/Contents/MacOS/dosbox
2024-08-09 22:15:08.658 | Current dir: /Users/alex
2024-08-09 22:15:08.658 | stderr verbosity: 0
2024-08-09 22:15:08.658 | -----------------------------------
2024-08-09 22:15:08.659 | dosbox-staging version 0.82.0-alpha (a07ee)
2024-08-09 22:15:08.659 | ---
2024-08-09 22:15:08.659 | LOG: Loguru version 2.1.0 initialised
2024-08-09 22:15:08.662 | CONFIG: Created primary config file '/Users/alex/Library/Preferences/DOSBox/dosbox-staging.conf'
2024-08-09 22:15:08.664 | CONFIG: Loaded primary config file '/Users/alex/Library/Preferences/DOSBox/dosbox-staging.conf'
2024-08-09 22:15:08.664 | LANG: Using internal English language messages
2024-08-09 22:15:08.778 | SDL: version 2.30.6 initialised (cocoa video and coreaudio audio)
2024-08-09 22:15:08.778 | DISPLAY: Using 1184x888 window size in windowed mode on display-0
2024-08-09 22:15:09.179 | OPENGL: Vendor: Intel Inc.
2024-08-09 22:15:09.179 | OPENGL: Version: 2.1 INTEL-14.7.28
2024-08-09 22:15:09.179 | OPENGL: GLSL version: 1.20
2024-08-09 22:15:09.179 | OPENGL: NPOT textures supported
2024-08-09 22:15:09.179 | RENDER: Using adaptive CRT shader based on the graphics standard of the video mode
2024-08-09 22:15:09.220 | RENDER: Auto-switched to shader 'crt/hercules'
2024-08-09 22:15:09.220 | VGA: Pixel doubling enabled
2024-08-09 22:15:09.220 | VIDEO: Using the DOS video mode's frame rate
2024-08-09 22:15:09.235 | MEMORY: Using 4096 DOS memory pages (16 MB) at address: 0x7fb60c000000
2024-08-09 22:15:09.235 | VIDEO: Initialised S3 Trio64 VESA 2.0 compatible with 4 MB of FP DRAM supporting 92 modes
2024-08-09 22:15:09.243 | VGA: Double scanning VGA video modes enabled
2024-08-09 22:15:09.243 | VOODOO: Initialized with 4 MB of RAM, multithreading, and no bilinear filtering
2024-08-09 22:15:09.244 | CAPTURE: Image capturer started
2024-08-09 22:15:09.285 | MIXER: Initialised stereo 48000 Hz audio with 512 sample frame buffer
2024-08-09 22:15:09.285 | MIXER: Master compressor enabled
2024-08-09 22:15:10.495 | MIDI: Opened device: coremidi
2024-08-09 22:15:10.495 | MPU-401: Running in intelligent mode on ports 330h and 331h
2024-08-09 22:15:10.495 | OPL: Operating at 49716 Hz and downsampling to the output rate
2024-08-09 22:15:10.495 | OPL: Running OPL3 on ports 220h and 388h
2024-08-09 22:15:10.495 | OPL: Sound Blaster 16 OPL output filter enabled
2024-08-09 22:15:10.496 | DMA: Initialised primary controller
2024-08-09 22:15:10.496 | DMA: Initialised secondary controller
2024-08-09 22:15:10.496 | SB: Operating at 22050 Hz and upsampling to the output rate
2024-08-09 22:15:10.496 | SB16: Modern DAC output filter enabled
2024-08-09 22:15:10.496 | SB16: Setting 'BLASTER' environment variable to 'A220 I7 D1 H5 T6'
2024-08-09 22:15:10.496 | SB16: Running on port 220h, IRQ 7, DMA 1, and high DMA 5
2024-08-09 22:15:10.498 | PCSPEAKER: Operating at 32000 Hz and upsampling to the output rate
2024-08-09 22:15:10.498 | PCSPEAKER: Initialised impulse model
2024-08-09 22:15:10.498 | PCSPEAKER: High-pass filter enabled (18 dB/oct at 120 Hz)
2024-08-09 22:15:10.498 | PCSPEAKER: Low-pass filter enabled (18 dB/oct at 4300 Hz)
2024-08-09 22:15:10.498 | MOUSE: Will be captured after the first left or right button click
2024-08-09 22:15:10.498 | MOUSE: Middle button will capture/release the mouse (clicks not sent to the game/program)
2024-08-09 22:15:10.498 | MOUSE (PS/2): Standard, 3 buttons
2024-08-09 22:15:10.519 | DOS: Loaded modern locale for country 1, 'United States'
2024-08-09 22:15:10.519 | SLIRP: Slirp version: 4.7.0
2024-08-09 22:15:10.519 | SLIRP: Successfully initialized
2024-08-09 22:15:10.519 | NE2000: Initialised on port 300h and IRQ 3
2024-08-09 22:15:10.520 | CONFIG: Loaded auto-mounted drive config file '/Applications/DOSBox Staging.app/Contents/MacOS/../Resources/drives/y.conf'
2024-08-09 22:15:10.521 | MAPPER: No joysticks found
2024-08-09 22:15:10.522 | MAPPER: Loaded default key bindings
2024-08-09 22:15:10.533 | RENDER: Auto-switched to shader 'crt/vga-1080p'
2024-08-09 22:15:10.533 | SDL: Using standard SDI (auto) display refresh rate of 60 Hz
2024-08-09 22:15:10.551 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz throttled VFR, scaled to 1067x800 pixels with 1:1.35 (20:27) pixel aspect ratio
2024-08-09 22:15:10.551 | OPENGL: Disabled vsync
2024-08-09 22:15:10.658 | MOUNT: Path '/Applications/DOSBox Staging.app/Contents/Resources/drives/y' found
```

interloper98:
> OK, tried to test this but the macOS build failed, see here: https://github.com/dosbox-staging/dosbox-staging/actions/runs/10311361167
> 
> Can someone link me a build to check out please?

Thanks for checking this, @Burrito78.

At the top of this PR, click the "Checks (37)" tab.
The click the "MacOS builds" link on the left (don't expand it down.. just click the link itself). 

Then scroll to the very bottom and the packages will be there.

interloper98:
Thank you @Burrito78.
OK - I think we're clear to merge!

PR #3869, opened 2024-08-10
DYN_LOG: Fix formatting type mismatch

Ringdingcoder:
# Description

This is the complaint from the compiler:

```
In file included from ../../src/cpu/core_dyn_x86.cpp:257: ../../src/cpu/core_dyn_x86/decoder.h: In function ‘CacheBlock* CreateCacheBlock(CodePageHandler*, PhysPt, Bitu)’: ../../src/cpu/core_dyn_x86/decoder.h:2427:69: warning: format ‘%X’ expects argument of type ‘unsigned int’, but argument 2 has type ‘Bitu’ {aka ‘long unsigned int’} [-Wformat=]
 2427 |                                 LOG_MSG("Unhandled dual opcode 0F%02X",dual_code);
      |                                                                  ~~~^  ~~~~~~~~~
      |                                                                     |  |
      |                                                                     |  Bitu {aka long unsigned int}
      |                                                                     unsigned int
      |                                                                  %02lX
```

# Checklist

I have:

### DIFF
--- a/src/cpu/core_dyn_x86/decoder.h
+++ b/src/cpu/core_dyn_x86/decoder.h
@@ -2424,7 +2424,7 @@ restart_prefix:
 
 			default:
 #if DYN_LOG
-				LOG_MSG("Unhandled dual opcode 0F%02X",dual_code);
+				LOG_MSG("Unhandled dual opcode 0F%02" PRIXPTR, dual_code);
 #endif
 				goto illegalopcode;
 			}
@@ -3006,7 +3006,7 @@ restart_prefix:
 			break;
 		default:
 #if DYN_LOG
-//			LOG_MSG("Dynamic unhandled opcode %X",opcode);
+//			LOG_MSG("Dynamic unhandled opcode %" PRIXPTR, opcode);
 #endif
 			goto illegalopcode;
 		}
--- a/src/cpu/core_dynrec/decoder.h
+++ b/src/cpu/core_dynrec/decoder.h
@@ -621,7 +621,7 @@ static CacheBlock *CreateCacheBlock(CodePageHandler *codepage, PhysPt start, Bit
 				        break;
 			        default:
 #if DYN_LOG
-//					LOG_MSG("Unhandled dual opcode 0F%02X",dual_code);
+//					LOG_MSG("Unhandled dual opcode 0F%02" PRIXPTR, dual_code);
 #endif
 				goto illegalopcode;
 			}
@@ -973,7 +973,7 @@ static CacheBlock *CreateCacheBlock(CodePageHandler *codepage, PhysPt start, Bit
 
 		default:
 #if DYN_LOG
-//			LOG_MSG("Dynrec unhandled opcode %X",opcode);
+//			LOG_MSG("Dynrec unhandled opcode %" PRIXPTR, opcode);
 #endif
 			goto illegalopcode;
 		}


FeralChild64:
Could you give some more info - which compiler, which version, which OS, which platform (x86, ARM, ...)?

Ringdingcoder:
I forgot to mention that I enabled the `DYN_LOG` define near the top of the file. Plain Fedora 39 x86_64, gcc 20240522 (Red Hat 13.3.1-1)

interloper98 (approved):
Looks ok. size_t can be 32bit or 64bit depending on the host, and priptr and prixptr (or %zu and %zb) are the printf specifiers to get this dynamic sizing right regardless of arch type.

weirddan455 (approved):
I hate wrangling these `printf` format specifiers.  Usually I'll just do something like `LOG_MSG("Blah: %lld, (long long) foo);` but this is a pointer type that can be different on 32 or 64 bit machines so that doesn't really work.

I'm just not sure how portable those type macros are but if it passes CI, it's portable enough so looks like a good change to me :smile:

Thanks for the contribution @Ringdingcoder

johnnovak:
Dunno... IMO it's a bit pointless to mess around with these debug/trace log messages meant for local dev only.

De-uint'ing the CPU code would be a much better change, or getting rid of the other warnings we get in the normal non-debug-only code, and then lowering the `max_warnings` counts.

weirddan455:
>  Dunno... IMO it's a bit pointless to mess around with these debug/trace log messages meant for local dev only.

Oh, yeah these shouldn't be compiled in a normal build unless you've manually modified the source to uncomment a `// 
#define DYN_LOG 1`

>  De-uint'ing the CPU code would be a much better change, or getting rid of the other warnings we get in the normal non-debug-only code, and then lowering the `max_warnings` counts.

Also true.  `Bitu` is defined as `using Bitu = uintptr_t;` but these specific variables are really just integers, not pointers so it's really the wrong type to be using.

This still seems like a fine change to me in the interrum though.  It doesn't hurt anything and I don't know if @Ringdingcoder wants to go about re-factoring our CPU code :sweat_smile:

johnnovak:
Yeah thanks @Ringdingcoder, we can get this in. Don't want to be negative, just wanted to clear up the direction we're heading towards 😄

weirddan455:
>  Also regarding 32-bit/64-bit, we've made an executive decision with @kklobe earlier this year to drop 32-bit support. We stopped providing the Windows 32-bit packages, we only support Win 10 64-bit as minimum, macOS has been 64-bit only for a while now, and so on.

Not willing to die on this hill but 32 bit is still the default for Raspberry Pi OS:  https://www.raspberrypi.com/software/operating-systems/

johnnovak:
> > Also regarding 32-bit/64-bit, we've made an executive decision with @kklobe earlier this year to drop 32-bit support. We stopped providing the Windows 32-bit packages, we only support Win 10 64-bit as minimum, macOS has been 64-bit only for a while now, and so on.
> 
> Not willing to die on this hill but 32 bit is still the default for Raspberry Pi OS: https://www.raspberrypi.com/software/operating-systems/

Good to know, but they provide 64-bit images and I'm sure slowly everybody will just migrate to 64-bit on the Pi as well. Only the very old RPi 1 and 2 are 32-bit only.

Also, we don't officially support the Pi, and think of it this way: at least we can play our small part in dragging people over to 64-bit 😎

Ringdingcoder:
> This still seems like a fine change to me in the interrum though. It doesn't hurt anything and I don't know if @Ringdingcoder wants to go about re-factoring our CPU code 😅

I don’t :wink:

Although I’m better at cleaning up other people’s stuff than my own…

There are copies of the same in dynrec, and it should probably be synchronized. I amended these two lines just now.

Issue #1789, opened 2022-07-12
Make the Sound Blaster code consistent with other audio devices


kcgen:
- [ ] Refactor it into a class
- [ ] Switch to the `MaybeRenderFrame()` callback approach
- [ ] Generate floating-point frames

When all audio devices support this, the frame handling logic can be generalized in the mixer instead of existing in each audio device.

johnnovak:
> Maybe related:
> 
>     * [[post-0.82.0] Standardize the Sound Blaster against other audio devices #3875](https://github.com/dosbox-staging/dosbox-staging/pull/3875)

@Torinde, please explain it to me, if you don't know if it's related or not, why did you make this comment? Boredom?

Grounded0:
I'm quitting all issue tracker duties so its good we still have someone sleuthing these old tickets since no one else does.

Looks like its close enough in its intent to potentially close this ticket when its merged.

interloper98:
> Looks like its close enough in its intent to potentially close this ticket when its merged.

My PR definitely gets the third item.

For the second item, the PR always renders per-frame at exactly the playback 
 rate (DMA and DAC); so I think it hit that point, just without the 'Maybe' aspect.

As for the first point (turning SB into a a class): I definitely saw this too - but I haven't touched it in my PR. Right now there's a global SB struct and a separate config class; it would be nice to combine all these and use member functions instead of static functions. Maybe I'll take that on later.

Thanks for connecting these points @Torinde. I'll mention this in the PR wording.

johnnovak:
> I'm quitting all issue tracker duties so its good we still have someone sleuthing these old tickets since no one else does.

Except for me and @Burrito78 a couole of times a year 😉 

> Right now there's a global SB struct and a separate config class; it would be nice to combine all these and use member functions instead of static functions. Maybe I'll take that on later.

Yeah, could be done, but we don't really need individual tickets for these refactor jobs IMO. This was a reminder for kcgen I think.

Then we have a ticket for the float conversion of all audio devices, plus yeah there's @interloper98's PR so I think this one can be closed.

> I was fairly certain it's related (c'mon, just by reading the titles it seems related). And by looking at the PR to me it seemed some things from here are accomplished in the PR. But I'm not confident to claim all that as fact, so I just wanted to notify you guys, so that you can.

Ok, thanks 😄 You just meddle a lot in dev related tickets for a non-developer, so I'm a bit wary when I see your comments these days 😆

[...]

Issue #3486, opened 2024-03-08
Audio popping in Entity game


Aardvark1982:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 10

### Relevant hardware info

i9-9900KS, RTX 2080

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Every time this game plays sound effect in DOSBox, it is accompanied by intolerable pop.
https://youtu.be/Ns5hg1GUuDY

In real hardware, and also in PCem with SBPro1 emulation, the audio only pops once, when the sample is played for the first time.
https://youtu.be/dEE2QmNIVlY

This game had CD release which supported ReelMagic videos, so it would be nice if constant popping could be fixed in DOSBox staging.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. 
2.
3.

### Download URL of affected game or software

https://archive.org/details/entity_202006

### Provide a Log

```text
https://gist.github.com/Aardvark1982/cd3a09afcd324628547bea260c157d42
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
> Every time this game plays sound effect in DOSBox, it is accompanied by intolerable pop.

I kinda like, it actually 😎 It gives a nice attack to the laser shot sound.

> In real hardware, and also in PCem with SBPro1 emulation, the audio only pops once, when the sample is played for the first time.

Well, PCem doesn't prove much... Are you saying you have tested this on real hardware, now? Not just relying some distant memory from the past? Can post a hardware recording?

Usually these pops happen because the sample contains a DC offset, and there's not _too much_ we can do about that, at least not easily, and not without degrading the sound quality for everything else.

Kinda curious of the real hardware recording if you can provide one, but probably the answer will be you'll have to live with it. Some games are just like this, some people did not clean up their samples properly... I think last time a similar issue came up was with Star Control 2, and we gave the same answer (those samples also have huge DC-offsets).

Aardvark1982:
> Well, PCem doesn't prove much... Are you saying you have tested this on real hardware, and recently? Not just relying on some distant memory from the past, right? Can you post a hardware recording too?
> 
[...]

PR #3875, opened 2024-08-14
Standardize the Sound Blaster against other audio devices

interloper98:
# Description

The Sound Blaster differed significantly from other audio devices in that:

 - It was relying on the Mixer and the AudioVector class to generate its output stream of `AudioFrame`s

 - It had Sound Blaster-specific (and exclusive) features implemented in the Mixer's source instead of in the Sound Blaster source; two of these were the DAC audio stretching and having its own queue inside the mixer.

 - It had its own channel-disabling routine, while (from what I can see) most of the other devices let the Mixer Channel's "sleep" feature manage this automatically (the work is then on the device side to decide when to wakeup). 

 - It had a lot of mechanisms to queue incoming data and then operate on it in chunks of data (the stretching feature was a good example), which added complexity to the code. As far as I can see, all the other audio devices use a 'RenderFrame' approach that have per-frame granularity.
 
 - The SB's DMA types are now converted inside the SB code at the point where the DMA array `maybe_silence`d, which now returns a standardised `vector<AudioFrame>`.

All of the above are addressed, plus a couple other minor adjustments:
- using a pool in the `rwqueue` (to avoid allocating from the heap per-frame),
- directly dequeuing into SDL's C-array,

The running heap allocations are reduced into an eighth of what they were; and for non-MIDI devices their allocations are flat-lined to zero.

Here's a comparison of Space Quest 3 using the MT-32:

- `dosbox sq3.bat & sleep 5 && heaptrack -p $(pidof dosbox)`

Before (about 2000 heap allocs per second):

sq3-mt32-heaptrack-main

After (about 250 heap allocs per second):

sq3-mt32-heaptrack-pr

# Related issues

Solved point 2 and 3 in ticket https://github.com/dosbox-staging/dosbox-staging/issues/1789

After this, I might try to address the first point.

# Manual testing

Tested all the audio devices and games as I did in @weirddan455's threaded mixer PR. Also ran thread and address sanitizer builds.

# Checklist

I have:


include/rwqueue.h
```diff
@@ -39,14 +39,27 @@
...
 #include <mutex>
 #include <optional>
 #include <vector>
 
 template <typename T>
 class RWQueue {
 private:
-	std::deque<T> queue{}; // faster than: vector, queue, and list
+#ifdef HAVE_MEMORY_RESOURCE
+	std::pmr::unsynchronized_pool_resource pool = {};
```
johnnovak, line 55: We'll trust you on this 😅 I don't think any of us are familiar with this, we're not exactly C++ gurus 😆

interloper98: Good news is that the rwqueue_tests.cpp appear to be very thorough and exercise the pooled deque as the rwqueue's internal container. (No need for trusted gurus here :genie_man:!) :laughing:

src/hardware/mixer.cpp
```diff
@@ -2527,22 +2527,26 @@ static void SDLCALL mixer_callback([[maybe_unused]] void* userdata,
...
-	if (num_frames > 0) {
-		std::vector<AudioFrame> audio_frames = {};
-		mixer.final_output.BulkDequeue(audio_frames, num_frames);
-		memcpy(stream, audio_frames.data(), audio_frames.size() * sizeof(AudioFrame));
-	}
+	const auto frames_to_dequeue = std::min(mixer.final_output.Size(),
+	                                        frames_requested);
+
+	const auto frames_received = mixer.final_output.BulkDequeue(frame_stream,
+	                                                            frames_to_dequeue);
```
weirddan455, line 2454: See comment on the `BulkDequeue` assert.  It's good if we can not special case the zero frames case.

src/hardware/sblaster.cpp
```diff
@@ -309,7 +311,8 @@ static int e2_incr_table[4][9] = {
 };
 
 static int frames_added_this_tick = 0;
-RWQueue<std::unique_ptr<AudioVector>> soundblaster_mixer_queue{128};
+
+RWQueue<AudioFrame> soundblaster_mixer_queue{128};
```
weirddan455, line 315: The size needs to be increased.  Previously, each element was about 1ms of audio (so 128ms total buffer).  Now this is just 128 frames (less than 3ms at a typical sample rate).

Ideally this can re-use the generic `MIXER_PullFromQueueCallback` which re-sizes it dynamically based on frames per block and whether we're currently in fast-forward mode.

src/hardware/sblaster.cpp
```diff
@@ -2849,23 +2945,29 @@ bool SB_GetAddress(uint16_t &sbaddr, uint8_t &sbirq, uint8_t &sbdma)
 	return (sbaddr != 0 && sbirq != 0 && sbdma != 0);
 }
 
-static void sblaster_mixer_callback([[maybe_unused]]const int requested_frames)
+static void sblaster_mixer_callback(const int requested_frames)
```
weirddan455, line 2948: So yeah, I don't think this function does anything special now.  I wrote this to handle the `AudioVector` class but now that it's gone, you should be able to delete this entire thing and replace the callback with `MIXER_PullFromQueueCallback`

src/hardware/mixer.cpp
```diff
@@ -2036,7 +2036,7 @@ bool MixerChannel::WakeUp()
 template <class Type, bool stereo, bool signeddata, bool nativeorder>
 void MixerChannel::AddSamples(const int num_frames, const Type* data)
 {
-	assert(num_frames >= 0);
+	assert(num_frames > 0);
```
weirddan455, line 2039: I'm pretty sure I ran into this assert triggering with 0 frames at some point.  It may have been while I was doing something dumb before I got my PR into the final state though so I guess we can leave it and see if it triggers again.

weirddan455 (changes requested):
Nice job getting the heap allocations down.  Found a few bugs.  Also recommend testing Alone in the Dark configured with "Sound Blaster (no DMA)" as one of the few games that uses the SoundBlaster DAC to make sure that works correctly with your changes.

src/hardware/sblaster.cpp
```diff
@@ -903,10 +943,54 @@ static uint32_t read_dma_16bit(const uint32_t bytes_to_read, const uint32_t i =
...
+		const auto index_a  = std::max(0, static_cast<int>(dac_index));
+		const auto sample_a = static_cast<float>(dac_samples[index_a]);
+		const auto fraction = dac_index - static_cast<float>(index_a);
+
+		if (is_zero_hold || fraction == 0.0f) {
+			return sample_a;
+		} else {
+			const auto index_b  = std::min(index_a + 1, dac_end_index);
+			const auto sample_b = static_cast<float>(dac_samples[index_b]);
+			return lerp(sample_a, sample_b, fraction);
```
weirddan455, line 978: Also, just realized.  This new function deviates in behavior from `main`.  The old code called this function instead of a call to `AddSamples` and did this simple resampling method in-place.  This new code does both.  So it will end up re-sampling a 2nd time (possibly going through Speex) and probably won't sound quite right.

src/misc/rwqueue.cpp
```diff
@@ -244,24 +257,30 @@ std::optional<T> RWQueue<T>::Dequeue()
 }
 
 template <typename T>
-bool RWQueue<T>::BulkDequeue(std::vector<T>& into_target, const size_t num_requested)
+size_t RWQueue<T>::BulkDequeue(std::vector<T>& into_target, const size_t num_requested)
 {
-	constexpr size_t min_items = 1;
-	assert(num_requested >= min_items);
```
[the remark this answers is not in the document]
johnnovak: IMO, it's good not to have to special-case zero-size operations in such containers. I don't like it when something blows up when I pass it an empty container, etc.

src/hardware/mixer.cpp
johnnovak: Yeah, so general comment, it's extremely hard to 100% guarantee that audio devices will _never ever_ try to publish zero frames. And I don't really see why would that be a problem, either, it if happens occasionally.

One school of thought would be to make sure publishing zero frames works, but keep the `num_frames > 0` debug assert. I'm fine with that, it might draw attention to some bigger issues. Maybe. But I really want everything to handle zero-length data correctly in release builds.

johnnovak:
I will review this in more detail later @interloper98, but reading @weirddan455's comments I've concluded that this is just too risky. 

I don't think it's a good idea to merge this is 5 days before the release date (20 Aug). We've tested @weirddan455's reworks extensively, I'm confident in that, but this PR just changes too much and there are many things that can go wrong.

src/hardware/sblaster.cpp
interloper98: Thanks for all the help with this @weirddan455; I've dropped the `sblaster_mixer_callback` call and switch to the `MIXER_PullFromQueueCallback` pattern and it's working nicely!
I especially like how this function adjusts the queue size on the fly depending on if we're fastwording or not! :1st_place_medal:

interloper98: This is transition is a separate commit (and acts as a nice "step" in the evolution; both the prior and later compile and run)

src/misc/rwqueue.cpp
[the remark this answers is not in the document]
interloper98: Yes, it handles a zero-request size without issue; but it does need to lock and unlock, so I've left the assert at least to catch no op spinners :wheel:.

We might as well cut off the no-op spinning as close to the source as possible, because (at least on my system), hundreds of locks and unlocks per millisecond can chew up a lot of CPU time :laughing: 

(I'm looking at Tyrian's startup phase where the mixer thread runs at 100% CPU my system for a couple seconds.. then settles down to roughly matching the main thread! That game hits it very hard :laughing: )

src/hardware/sblaster.cpp
interloper98: > Ideally this can re-use the generic `MIXER_PullFromQueueCallback` which re-sizes it dynamically based on frames per block

Moved the SB to use MIXER_PullFromQueueCallback, so this global `soundblaster_mixer_queue` is now gone like you suggested.

src/hardware/sblaster.cpp
```diff
@@ -903,10 +943,54 @@ static uint32_t read_dma_16bit(const uint32_t bytes_to_read, const uint32_t i =
 	return check_cast<uint32_t>(bytes_read);
 }
 
-static void enqueue_frames(std::unique_ptr<AudioVector> frames)
+static void enqueue_frames(std::vector<AudioFrame>& frames)
 {
-	frames_added_this_tick += frames->num_frames;
-	soundblaster_mixer_queue.NonblockingEnqueue(std::move(frames));
+	soundblaster_mixer_queue.NonblockingBulkEnqueue(frames);
+	frames_added_this_tick += check_cast<int>(frames.size());
```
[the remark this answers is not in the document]
interloper98: Fixed; thanks!

src/hardware/sblaster.cpp
```diff
@@ -457,6 +460,12 @@ static void set_filter(MixerChannelPtr channel, const FilterConfig& config)
 	}
 
 	channel->SetResampleMethod(config.resample_method);
+
+	// The SB's DAC samples are stretched locally before being sent to the
+	// mixer. The stretching is done using the sample resampling method as
+	// the channel, so we keep track of this with the other DAC settings.
+	//
+	sb.dac.resample_method = config.resample_method;
```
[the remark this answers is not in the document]
interloper98: Added getter!

src/hardware/sblaster.cpp
```diff
@@ -2910,11 +3011,10 @@ static void generate_frames(const int frames_requested)
...
 	if (sb.dac.used > 0) {
-		enqueue_frames(std::make_unique<AudioVectorStretched>(sb.dac.used, sb.dac.data));
-		// Reverse the addition done by enqueue_frames()
-		// This works everywhere but here because of stretched frames
-		frames_added_this_tick -= sb.dac.used;
-		frames_added_this_tick += frames_requested;
+		enqueue_frames(stretch_dac_to_frames(sb.dac.data,
+		                                     sb.dac.used,
+		                                     frames_requested,
+		                                     sb.dac.resample_method));
```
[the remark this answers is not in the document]
interloper98: `->GetResampleMethod()` is now in place.

src/hardware/sblaster.cpp
interloper98: The original stretch code bypassed the sound blaster's own channel and was injecting samples directly into the final mix buffer, and needed to do it's own scaling and channel mapping. This is unlike any of the audio devices - even the LPT DACs, Tandy DAC, and PS1 DACs don't have this "direct feed" into the main mixer.

The new code is entirely self-contained inside the Sound Blaster. It resamples the DAC up to whatever the Sound Blaster's "hardware" output rate is set at (which may or may not be the SDL's mixer rate). This is entirely software controlled; games can set the SB to run at 22 KHz or 45 KHz.. and then they can still feed it DAC samples at some other rate (greater or less than).

So this new code goes from the DAC rate to the SB's own rate using the desired resampling method.  Typically the DAC rate is something like 8 KHz and the SB's running at 22 KHz. If ZOH is used, then you get the crunchy sound from 8 KHz -> 22 KHz, which I think, is 'proper'.

And from there, the SB's channel output ("crunchy 22 KHz") will be sampled again using ZOH up to whatever the user's mixer is running at.. typically 48 KHz.

interloper98:
> I don't think it's a good idea to merge this in 5 days before the RC release date (20 Aug). We've tested @weirddan455's rework extensively, I'm confident in that, but this PR just changes too much and there are many things that can go wrong.

No worries on timing! post 0.82 is fine and give us more time to test it and maybe get more refinements in.

johnnovak:
> Nice job getting the heap allocations down. Found a few bugs. Also recommend testing Alone in the Dark configured with "Sound Blaster (no DMA)" as one of the few games that uses the SoundBlaster DAC to make sure that works correctly with your changes.

Yep, this is important. It's described here:
https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-emulation-tests

Executing as many tests as you can on this page is a good idea @interloper98 — that's how I will ultimately tests this before it gets merged.

src/hardware/mixer.cpp
interloper98: > One school of thought would be to make sure publishing zero frames works, but keep the num_frames > 0 debug asserts.

Yes; those are the couple adjustments I've made in the bulk dequeue; so it's safe in release builds but we can catch scenarios where non-blocking calls may be spinning hard with zero requests (because there's no back pressure to throttle them as they're non-blocking). At least we can be made aware and decide how to handle it.

src/misc/rwqueue.cpp
```diff
@@ -51,19 +51,16 @@ size_t RWQueue<T>::Size()
...
 
 template <typename T>
 void RWQueue<T>::Stop()
 {
-	std::unique_lock<std::mutex> lock(mutex);
 	if (!is_running) {
 		return;
 	}
 	is_running = false;
-	lock.unlock();
```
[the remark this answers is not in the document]
interloper98: Fixed; thanks @weirddan455!

src/hardware/sblaster.cpp
weirddan455: I agree with all those points in theory.  What's more important is testing and making sure it works well in practice.  SoundBlaster DAC gave me a bit of trouble at first.

src/hardware/mixer.cpp
interloper98: Fixed

interloper98:
@johnnovak - thanks for all those audio test cases. I've been using them during the refactor and it's passing all of them. 

@weirddan455, the new push addresses all comments and ready for more review and testing (all post 0.82; so no rush!)

The Sound Blaster is now passing some other things, too, like:

https://github.com/user-attachments/assets/20ef14df-5ef1-4b4d-bac6-060a1cafde2a

Which previously sounded like:

https://github.com/user-attachments/assets/f124c861-d43e-41ef-8600-54ccf1fd1b53

Here's another one we're passing:

https://github.com/user-attachments/assets/50619989-afd5-49f7-884a-7076f3007688

In Crystal Dream around the 3-minute mark you can hear some little 'warbly' hitches, however it does this on real hardware too (also around 3 min): https://youtu.be/LNXERInw69g?t=204

include/rwqueue.h
```diff
@@ -39,14 +39,27 @@
...
+
+	// The pooled deque recycles objects from the pool. It only allocates
+	// when it needs to grow beyond its prior peak size.
+	std::pmr::deque<T> queue = {};
+#else
+	std::deque<T> queue = {}; // faster than: vector, queue, and list
```
johnnovak, line 59: Minor: No end of line comments please unless absolutely necessary (or in case blocks for port descriptions).

include/rwqueue.h
```diff
@@ -134,37 +132,52 @@ class RWQueue {
...
-	// function clear()s the vector such that it's in a defined state on
-	// return (and can be reused). The method potentially blocks until there
-	// is enough capacity in the queue for the new items.
+	// Items are std::move'd out of the source vector and also erased so the
+	// source vector is returned in a defined state. The method potentially
+	// blocks until there is enough capacity in the queue for the source items.
 
 	// If queuing has stopped prior to bulk enqueing, then this will
-	// immediately return false and no items will be queued.
+	// immediately return with 0 and no items will be queued.
```
johnnovak, line 140: Prefer simple language: "If the queue is stopped, ..."

include/rwqueue.h
interloper98: Will fix that (it was an existing line of code; I only added the PMR chunk).

include/rwqueue.h
```diff
@@ -134,37 +132,52 @@ class RWQueue {
...
+	// until the requested number of items have been dequeued. Normally all
+	// of the requested items are dequeued, however:
+	//
+	// - If queuing was stopped prior to bulk dequeueing then this
+	//   immediately returns with a value of 0 and no items dequeued.
 	//
-	// The vector is always sized to match the number of items returned.
-	bool BulkDequeue(std::vector<T>& into_target, const size_t num_requested);
+	// - If queuing was stopped in the middle of bulk dequeueing then this
+	//   immediately returns with a value indicating the subset dequeued.
```
johnnovak, line 171: Is there any value in providing API guarantees for "stop queue while performing an action" behaviour? I don't think so as enqueing normally you only stop the queue on shutdown or subsystem reinit' right?

Would it simplify the code, the contract, and the doco to just stipulate at the top of the class in a comment to ensure no enque/dequeue should be happening during shutting down if you care about well-specified, deterministic behaviour? I think it's reasonable to push the responsibility of waiting before shutdown until all enque/dequeue ops have finished to the client, otherwise undefined behaviour is fine.

I'd wager we don't even care about proper queue shutdown that during reinit, right? And surely not during the final shutdown on emulator exit.

src/misc/rwqueue.cpp
```diff
@@ -244,24 +264,34 @@ std::optional<T> RWQueue<T>::Dequeue()
...
 
 	while (num_remaining > 0) {
 		std::unique_lock<std::mutex> lock(mutex);
 
-		const auto num_items = std::max(min_items,
-		                                std::min(num_remaining,
-		                                         queue.size()));
+		constexpr size_t MinItems = 1;
+		const auto num_items      = std::max(MinItems,
+                                                std::min(num_remaining,
```
johnnovak, line 293: Use clamp from math utils in all such cases.

include/mixer.h
```diff
@@ -56,6 +56,7 @@ static constexpr int MixerBufferMask     = MixerBufferByteSize - 1;
 
 // TODO This seems like is a general-purpose lookup, consider moving it
 extern int16_t lut_u8to16[UINT8_MAX + 1];
+static constexpr int16_t* lut_s8to16 = lut_u8to16 + 128;
```
johnnovak, line 59: Nice; you're doing the exact refactorings I had in mind but ran out of steam to do 😆

include/mixer.h
```diff
@@ -518,30 +520,21 @@ inline void MIXER_PullFromQueueCallback(const int frames_requested, DeviceType*
...
-			if (std::is_same<AudioType, AudioFrame>::value) {
-				// AudioFrame has the same memory layout as 2x floats
-				device->channel->template AddSamples<float, stereo, signeddata, nativeorder>(frames_recieved, reinterpret_cast<float*>(to_mix.data()));
-			} else {
-				device->channel->template AddSamples<AudioType, stereo, signeddata, nativeorder>(frames_recieved, to_mix.data());
-			}
-			#ifndef __clang__
-			#pragma GCC diagnostic pop
-			#endif
+	static std::vector<AudioType> to_mix = {};
```
johnnovak, line 519: 💯 So glad to see the pragma junk gone 😆 🥳

src/hardware/mixer.cpp
```diff
@@ -2508,7 +2508,9 @@ static void capture_callback()
 
 	// We can't block waiting on the mixer thread
 	// Some mixer channels block waiting on the main thread and this would deadlock
-	std::vector<int16_t> frames = {};
+	static std::vector<int16_t> frames = {};
+	frames.clear();
```
johnnovak, line 2433: 💯 We should use static buffers more where applicable, yeah.

src/hardware/mixer.cpp
```diff
@@ -1132,6 +1132,12 @@ float MixerChannel::GetFramesPerBlock() const
 	return static_cast<float>(mixer.blocksize) * stretch_factor;
 }
 
+double MixerChannel::GetMsPerFrame() const
+{
+	return MillisInSecond / sample_rate_hz;
+	// The double return value is used for PIC timing (which uses doubles)
```
johnnovak, line 1140: Minor: can you move the comment to the preceding line please

include/rwqueue.h
weirddan455: I would just get rid of that comment honestly.

"faster than vector" - Needs benchmark to really prove that.

"faster than queue" - Misleading.  std::queue is only an interface that uses std::deque as the implementation by default.

"faster than list" - Probably true.  Friends don't let friends use linked lists.

include/rwqueue.h
weirddan455: The queues do get stopped at runtime inside the `MIXER_LockMixerThread()` function.  Otherwise you get deadlocks.  Most common situation this happens is when changing mixer settings.

It doesn't simplify anything to change this.  Logic looks like this (in pseudo code):

```cpp
BulkDequeue()
{
    while(queue_is_running && output.size() < requested_size) {
        cond_var.wait_until(has_more_data || queue_is_stopped);
        output_data.enqueue(audio_data);
    }
}
```

When the queue gets stopped, it returns immediately instead of waiting for more data.  The main point of stopping the queue is to make the consumer stop waiting for more data so that another thread can acquire the global mixer mutex.

src/hardware/mixer.cpp
```diff
@@ -1983,8 +1983,11 @@ AudioFrame MixerChannel::Sleeper::MaybeFadeOrListen(const AudioFrame& frame)
 	if (!had_signal) {
 		// Otherwise, we simply passively listen for signal
 		constexpr auto SilenceThreshold = 1.0f;
-		had_signal = fabsf(frame.left)  > SilenceThreshold ||
-		             fabsf(frame.right) > SilenceThreshold;
+
+		had_signal = fabsf(frame.left - last_frame.left) > SilenceThreshold ||
+		             fabsf(frame.right - last_frame.right) > SilenceThreshold;
+
```
johnnovak, line 1991: Very nice commit message 👍🏻 

Basically, you've extended this so it treats DC as silence. That's fine, but I will introduce a per-channel averaging DC remover which will obviate the need for this (but it will do no harm).

`SilenceThreshold` reads a bit weird in this altered context; you might wanna change it.

> This also means that these channels will no longer pollute the downstream filters (which was theoretically possible).

Not sure what you means by "pollute". The per-channel LPF/HPF filters don't care about DC, and the master 20 Hz HPF will remove DC completely from the final output. Because we're using floats, DC doesn't matter in the sense that it cannot cause clipping, then we remove it anyway with the master HPF.

The per-channel averaging DC remover will be needed however to remove  "zipper noise" in certain interesting DC-infested scenarios. See #3486 

Anyway, you can leave this in if you want for now.

src/hardware/sblaster.cpp
```diff
@@ -3117,308 +3154,282 @@ static bool is_cms_enabled(const SbType sbtype)
...
+void SBLASTER::SetupEnvironment()
+{
+	// Ensure our port and addresses will fit in our format widths.
+	// The config selection controls their actual values, so this is
+	// a maximum-limit.
+	assert(sb.hw.base < 0xfff);
+	assert(sb.hw.irq <= 12);
+	assert(sb.hw.dma8 < 10);
+
+	char blaster_env_val[] = "AHHH II DD HH TT";
```
johnnovak, line 3166: That's what I say when I'm trying to order my tenth beer 😎 🍻

johnnovak (approved):
Top notch 10/10 work @interloper98! 🎖️ I very much appreciate the detailed commit messages, exemplary stuff! 🚀 

Getting rid of the stretching code and replacing it with proper sampling is great; that's how this should've been implemented from the get go.

Only minor comments, this can go in first thing when the 0.83.x cycle starts :party

include/rwqueue.h
interloper98: snipped!

include/rwqueue.h
interloper98: fixed:

```c++
// If the queue is stopped then the function unblocks and returns the
// quantity enqueued (which can be less than the number requested).
```

include/rwqueue.h
interloper98: Yes; even if the bookkeeping granularity was made coarser - and we only traked bulk actions at the "chunk" level (and ignored individual item quantities), the "issue" is that the while loop is already performing enmasse `std::move`s of swaths of items (with a single algorithm call), in largest chunk size possible.

I also checked if we could move the notification outside the bulk while loops - but that's also impossible because it can lead to deadlocks (which makes sense: if we're enqueuing or dequeuing more items in bulk than the  queue's size.. those notifications after each chunk will "pump" the other end of the queue to relieve the back or front pressure).

src/hardware/mixer.cpp
```diff
@@ -2526,7 +2526,7 @@ static void capture_callback()
 static void SDLCALL mixer_callback([[maybe_unused]] void* userdata,
                                    Uint8* stream, int bytes_requested)
 {
-	assert(bytes_requested >= 0);
+	assert(bytes_requested > 0);
```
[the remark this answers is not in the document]
interloper98: yup; these helped catch my own bugs in this PR (one of the loops was spinning hard and the mixer's CPU usage was at 100% :sweat_smile: )

src/hardware/mixer.cpp
interloper98: Thanks! Fixed.

src/hardware/sblaster.cpp
interloper98: :rofl:

src/hardware/mixer.cpp
interloper98: > Basically, you've extended this so it treats DC as silence.

Oh yes. I'll rename the constant to use better terminology (maybe `DeltaThreshold` or `ChangeThreshold`.

> I will introduce a per-channel averaging DC remover which will obviate the need for this (but it will do no harm).

Excellent; right - if a constant DC shift in a signal is removed, so all signals ultimately oscillate around the 0-line, then inspecting absolute values (instead of difference) will again work fine for sleeping.

I noticed with the DAC, some games an demos leave the sound blaster running in DAC mode (which is fine), however the new approach of "sampling the register" means we just keep reading the last written reg value over and over - so it leads to a constant DC until the game or demo decides to write new stuff.  This at least turns off the channel after half a second or so.

> Not sure what you means by "pollute". The per-channel LPF/HPF filters don't care about DC, and the master 20 Hz HPF will remove DC completely from the final output. Because we're using floats, DC doesn't matter in the sense that it cannot cause clipping, then we remove it anyway with the master HPF.

Even better; thanks for that info. Yes - even with this change, I didn't hear any differences, but I wanted to let the channel sleep because it relaxes the CPU usage.

interloper98: Fixed.

src/misc/rwqueue.cpp
interloper98: applied the clamps

the_silence_of_the_clamps

include/rwqueue.h
johnnovak: Cool, I was just wondering 😄

src/hardware/sblaster.cpp
```diff
@@ -2816,8 +2903,52 @@ static uint8_t read_sb(const io_port_t port, const io_width_t)
...
+	}
+
+	// This callback is fired at a rate of once every channel frame, so we
+	// decrement the positive balance because we've consumed one frame
+	// simply through the passage of time.
+	if (queued_frames_balance) {
+		--queued_frames_balance;
+	}
+
+	PIC_AddEvent(sblaster_pic_callback, sblaster->channel->GetMsPerFrame());
```
weirddan455, line 2927: Not sure if I'm a fan of this.  Maybe I could be convinced if you explained your rational.  This is adding a lot of events doing this for every single audio frame.  The PIC queue is implemented as a linked list so it's not efficient for large amounts of things.

I'm also not seeing what the advantage is over using a loop like I had before.  From my understanding, the PIC queue fires every 1ms after the CPU emulation is done.  So these events will either happen immediately following this or on the next tick.  It should work the same as a loop but with extra steps and "PIC abuse" as @johnnovak likes to call it.

src/hardware/sblaster.cpp
```diff
@@ -2850,125 +2981,44 @@ bool SB_GetAddress(uint16_t &sbaddr, uint8_t &sbirq, uint8_t &sbdma)
...
-		// 5 seconds is safe to avoid stuttering. Be careful if modifying this value as we're about to clear pending audio.
-
-		// This is different from the "sb.speaker_enable = false" state
-		// If the speaker is off, this function never gets called
-		// This is speaker on but playing silence
-		soundblaster_mixer_queue.Stop();
-		soundblaster_mixer_queue.Clear();
-		sb.chan->Enable(false);
-	}
-	break;
```
weirddan455, line 2904: Did you move this silence detection elsewhere or just remove it?  It's kind of important for games that primarily use MIDI sound and infrequently play DMA sound effects.  While the SoundBlaster channel is running, it forces the entire mixer into "synchronous mode" where it must wait on the main thread to generate sound.  Allowing the SoundBlaster channel to stop lets the mixer go async with glitch free MIDI.

weirddan455 (commented):
Gave this another look and have some concerns.  I don't know what's going on with CI.  It's showing a lot of skipped tests for some reason.  I build it locally and it works.  Tested Alone in the Dark for SoundBlaster DAC mode and Doom for DMA.

johnnovak:
> The Sound Blaster is now passing some other things, too, like:

@interloper98 
Would you be able to add these test cases to the audio test wiki page? If you could add Crystal Dreams to the test packs as well and re-upload it, that would be great too.

Documenting these special games and demos that do weird things to the audio hardware  is super valuable for regression testing.

Also, I will try to review this whole big PR over the next week or so.

src/hardware/sblaster.cpp
johnnovak: General comment: simpler (and I daresay, more primitive) solutions are better that get the job done, and no more PIC abuse please if it can be avoided 😆

src/hardware/sblaster.cpp
```diff
@@ -921,10 +953,14 @@ static uint32_t read_dma_16bit(const uint32_t bytes_to_read, const uint32_t i =
...
+// This keeps track of the balance of queued frames. It's a balance because
+// it's decremented through the passage of time (every per-frame tick).
+size_t queued_frames_balance = 0;
+
+void enqueue_frames(std::vector<AudioFrame>& frames)
 {
-	frames_added_this_tick += frames->num_frames;
-	soundblaster_mixer_queue.NonblockingEnqueue(std::move(frames));
+	assert(sblaster);
+	queued_frames_balance += sblaster->output_queue.NonblockingBulkEnqueue(frames);
```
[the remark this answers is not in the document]
johnnovak: Yep, sounds like a bug.

src/hardware/sblaster.cpp
```diff
@@ -1209,26 +1227,9 @@ static void flush_remainig_dma_transfer()
...
-	//   Sound Blaster Series Hardware Programming Guide,
-	//   41h Set digitized sound output sampling rate, DSP Commands 6-15
-	//   https://pdos.csail.mit.edu/6.828/2018/readings/hardware/SoundBlaster.pdf
-	//
-	constexpr int MinRateHz = 5000;
-
-	const auto rate_hz = std::clamp(requested_rate_hz, MinRateHz, NativeDacRateHz);
+	const auto rate_hz = std::clamp(MinPlaybackRateHz,
+	                                requested_rate_hz,
+	                                NativeDacRateHz);
```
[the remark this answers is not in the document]
johnnovak: Oh yeah, this would screw up the samplerate in some cases.

src/hardware/sblaster.cpp
johnnovak: Yep, reverting to glitch-free MIDI is super important. Put that SB DMA to sleep and get it out of the way when we can 😆

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
interloper98: Thanks; fixed in the push.

src/hardware/sblaster.cpp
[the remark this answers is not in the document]
interloper98: Didn't have much time lately; but now getting spooled back up on this. Thanks for the reviews.

In the per-frame generating cases (the DAC as well as demos and games that setup a single frame DMA transfer count), this count is unnecessary because the PIC event callback is run at exactly the SB's playback rate. 

(Will refer back to this per-frame callback in the next comment!)

src/misc/rwqueue.cpp
```diff
@@ -244,24 +264,34 @@ std::optional<T> RWQueue<T>::Dequeue()
...
-
-	if (into_target.size() != num_requested) {
+	if (into_target.size() < num_requested) {
 		into_target.resize(num_requested);
 	}
 
-	auto target_start  = into_target.begin();
+	const auto num_dequeued = BulkDequeue(into_target.data(), num_requested);
+
+	// cap off the target vector to match the number that were dequeued
```
johnnovak, line 280: were -> was

src/misc/rwqueue.cpp
```diff
@@ -244,24 +264,34 @@ std::optional<T> RWQueue<T>::Dequeue()
...
 	while (num_remaining > 0) {
 		std::unique_lock<std::mutex> lock(mutex);
 
-		const auto num_items = std::max(min_items,
-		                                std::min(num_remaining,
-		                                         queue.size()));
+		constexpr size_t MinItems = 1;
+		const auto num_items      = std::max(MinItems,
+                                                std::min(num_remaining,
+                                                         queue.size()));
```
johnnovak: Hmm, I thought you changed this to use `clamp` from `math_utils`?

src/hardware/sblaster.cpp
interloper98: The way DMA and the SB work together is that the card requests bytes from the DMA controller, sample by sample across time, that the card decodes and then plays.
 
The SB doesn't have a buffer to pull in multiple samples at a time or buffer block locally.. instead the SB literally treats the DMA memory as it's own, and it just reads from it at the playback rate. That's how CDreams works... it continually overwrites the DMA memory (instead of setting up big chunky transfers). 

So that's how the current PR gets this and other demos right, but also works fine for quake and other DMA games.

But your right @weirddan455 - running the PIC callback at a per-frame timing is more expensive than running the callback once every tick; (I could measure between a 2% and 5% fps loss in quake in the per-frame scenario)

So I'm adding a per-tick approach (N frames per 1 tick), exactly like you've done in the GUS and other cards. This mode will be used when the DMA transfer uses a larger multi-frame count.

src/misc/rwqueue.cpp
```diff
@@ -102,6 +102,13 @@ bool RWQueue<T>::IsEmpty()
 	return queue.empty();
 }
 
+template <typename T>
+bool RWQueue<T>::IsFull()
+{
+	std::lock_guard<std::mutex> lock(mutex);
+	return (queue.size() >= capacity);
+}
+
```
johnnovak, line 111: I'm confused. Why do we need `IsFull()` at all? All we need is to use `!IsEmpty()` in the client code.

src/hardware/sblaster.cpp
interloper98: Oh, it was a comment in the commit that removed it:

> 5. The local 5 second sleep was another Sound Blaster-specific feature
   that the other devices weren't using.  It's now replaced with the
   mixer's 'Sleep' feature.

include/mixer.h
```diff
@@ -518,30 +520,21 @@ inline void MIXER_PullFromQueueCallback(const int frames_requested, DeviceType*
...
+
+	const auto frames_received = check_cast<int>(
+	        device->output_queue.BulkDequeue(to_mix, frames_requested));
+
+	if (frames_received > 0) {
+		if constexpr (std::is_same_v<AudioType, AudioFrame>) {
+			device->channel->AddAudioFrames(to_mix);
+		} else {
+			device->channel->template AddSamples<AudioType, stereo, signeddata, nativeorder>(
+			        frames_received, to_mix.data());
```
johnnovak, line 529: OMG, that's some C++ weirdness I haven't seen before (the "inline" `constexpr` and `template` thing). I don't even wanna know what this does... 😅 Luckily, it will all go away with the move to using floats everywhere 😆

include/mixer.h
```diff
@@ -182,6 +182,7 @@ class MixerChannel {
 	int GetSampleRate() const;
 	float GetFramesPerTick() const;
 	float GetFramesPerBlock() const;
+	double GetMsPerFrame() const;
```
johnnovak, line 185: Suggest `GetMillisPerFrame` or `GetMillisecondsPerFrame`.

src/hardware/mixer.cpp
```diff
@@ -1140,6 +1140,13 @@ float MixerChannel::GetFramesPerBlock() const
 	return static_cast<float>(mixer.blocksize) * stretch_factor;
 }
 
+double MixerChannel::GetMsPerFrame() const
+{
+	// Note: the double return value is used for PIC timing (which uses doubles)
```
johnnovak: 👍🏻

src/misc/rwqueue.cpp
interloper98: They're not the same though (`!Is empty()` means you might have 1 up to the max)

src/misc/rwqueue.cpp
johnnovak: Ok... I was reviewing commit by commit. Found the change in a later commit 😄

interloper98: Had me worried I lost some commits or something! Glad it's addressed!

src/hardware/sblaster.h
```diff
@@ -0,0 +1,58 @@
...
+
+	OplMode oplmode = OplMode::None;
+
+	bool cms = false;
+
+	void SetupEnvironment();
+	void ClearEnvironment();
+};
+
+extern std::unique_ptr<SBLASTER> sblaster;
```
johnnovak, line 56: I know you're following existing patterns here, but I wouldn't expose this. I'd do this instead in `mixer.cpp`:

```cpp
void MIXER_UnlockMixerThread()
{
	PCSPEAKER_NotifyUnlockMixer();
	TANDY_NotifyUnlockMixer();
	...
	SBLASTER_NotifyUnlockMixer();
	REELMAGIC_NotifyUnlockMixer();

	mixer.mutex.unlock();
}

void MIXER_LockMixerThread()
{
	PCSPEAKER_NotifyLockMixer();
	TANDY_NotifyLockMixer();
	...
	SBLASTER_NotifyLockMixer();
	REELMAGIC_NotifyLockMixer();

	mixer.mutex.lock();
}
```

I've already started using this notify pattern in `render.cpp`; it's a super useful and powerful concept to reduce coupling between components. Looser is almost always better when it comes to coupling. @FeralChild64 also adopted it for the mouse and title bar stuff.

It would be great if you could spend a bit of time refactoring all the other audio devices as shown in my above example. These are just simple mechanical changes, but if you're busy, could you at least do the Sound Blaster and put a TODO there?

src/hardware/sblaster.cpp
```diff
@@ -886,22 +887,53 @@ static std::array<uint8_t, 2> decode_adpcm_4bit(const uint8_t data)
...
 
-	static std::vector<T> quiet_buffer = {};
-	constexpr auto Silent = Mixer_GetSilentDOSSample<T>();
+	static std::vector<AudioFrame> frames = {};
 
-	if (quiet_buffer.size() < num_samples) {
-		quiet_buffer.resize(num_samples, Silent);
+	const auto is_mono = (frame_type == FrameType::Mono);
+
+	const size_t num_frames = num_samples / (is_mono ? 1 : 2);
```
johnnovak, line 903: minor: `num_frames_for_frame_type` helper function instead?

src/hardware/sblaster.cpp
johnnovak: Ok, that's fine, then I guess adding a secondary emulator-only PIC to handle extra internal emulator-only duties that the real PIC was never intended to do is a job for another day.

src/hardware/sblaster.cpp
```diff
@@ -125,6 +121,22 @@ enum class DmaMode {
...
+	AudioFrame RenderFrame();
+
+private:
+	double last_write_ms   = {};
+	int rate_change_tally  = {};
+	int high_water_rate_hz = MinPlaybackRateHz;
+};
+
+// TODO: Combine the SBLASTER class, 'sb' global struct, and all the static
+// functions into a single unified class with member data and member functions.
```
johnnovak, line 138: General comment: there is _a lot_ happening in this commit! In your commit message you're listing 5 individual changes. I know it's a hassle, but would it be possible to at least move the mechanical code-shuffling out of this huge commit for easier bisecting? If this commit breaks something, well, good luck hunting down the offending line  a few months later... 😆

src/misc/rwqueue.cpp
johnnovak: Ok... brain malfunction, carry on 😆

src/hardware/sblaster.h
interloper98: Definitely; will add that. Thanks for this tip 👍

src/hardware/sblaster.cpp
interloper98: A lot of it was coupled; but will try to move out as much as possible into separate commits.

src/misc/rwqueue.cpp
interloper98: Np!

include/mixer.h
interloper98: > it will all go away with the move to using floats everywhere

🏅 That'll be so nice to have one code path instead of all that type handling.

src/hardware/sblaster.cpp
weirddan455: >  The SB doesn't have a buffer to pull in multiple samples; instead it treats the DMA memory as its own and reads from it at the playback rate.
>
>  (Crystal Dreams continually overwrites the DMA memory sample by sample; if the SB read in chunks, instead of setting up big chunky transfers).

Correct, and that's the same problem I ran into with GUS on #3993  I haven't researched it heavily yet but my theory is most games are going to treat DMA sound as a ring buffer and write ahead of where the SB is reading from (but also making sure not to write so far ahead the sound overlaps the read position).

Performance is a concern but also the ability to debug becomes much harder if you have to figure out how many events are sitting in the PIC queue.

If I can figure out a good way to get SoundBlaster DMA working on its own thread, that would better model how it works on real hardware.  Again, haven't done a lot of research yet but I think I'll find on real hardware, the SoundBlaster doesn't use the PIC at all for this (instead the SoundBlaster reads asynchronously from memory while the CPU running the game writes to it).

interloper98:
Almost done the separation of features to commits.. currently done:
1. SB to class header
4. Direct DAC handling (no more stretching in the mixer)
5. Mixer notifier adjustment (move to API calls per audio device)
6. Use the mixer's pull pattern in the SB

Still left:
 - Conditionally use the per-frame callback timing for DMA
 - Switch to the mixer's 'Sleep' feature

(Haven't addressed some comments yet.. will get to them :sweat_smile: )

johnnovak:
> Almost done the separation of features to commits.. currently done:
> 
>     1. SB to class header
> 
>     2. Direct DAC handling (no more stretching in the mixer)
> 
>     3. Mixer notifier adjustment (move to API calls per audio device)
> 
>     4. Use the mixer's pull pattern in the SB
> 
> 
> Still left:
> 
>     * Conditionally use the per-frame callback timing for DMA
> 
>     * Switch to the mixer's 'Sleep' feature
> 
> 
> (Haven't addressed some comments yet.. will get to them 😅 )

Excellent progress @interloper98 🎉 

Let me know when it's ready for the final review & testing (and no rush 😎 ).

src/hardware/sblaster.h
interloper98: Added.

src/hardware/sblaster.cpp
interloper98: Added; moved it to a `constexpr`.

include/mixer.h
interloper98: renamed

src/hardware/sblaster.cpp
interloper98: > Performance is a concern but also the ability to debug becomes much harder if you have to figure out how many events are sitting in the PIC queue.

The good news is there aren't a pile per-frame events in the PIC queue's linked list all sequenced out in time (so it's actually no different than the per-tick callback scenario).

When a callback is added to `TIMER_AddTickHandler(callback_function)`, the PIC system runs that function every tick, forever more, until it's explicitly removed from the handler.

Where when you use `PIC_AddEvent(callback_function, upcoming_event_millisecond)` that callback is only run one time (at the scheduled upcoming time), and then it's dropped from the PIC linked list.

So on both the per-frame and per-tick callback scenarios, there's only ever one upcoming callback in the PIC system to generate the next chunk of frame(s).

> If I can figure out a good way to get SoundBlaster DMA working on its own thread, that would better model how it works on real hardware. Again, haven't done a lot of research yet but I think I'll find on real hardware, the SoundBlaster doesn't use the PIC at all for this (instead the SoundBlaster reads asynchronously from memory while the CPU running the game writes to it).

Yup, that's exactly how DMA works. There's not PIC system involved; the device that uses the DMA memory makes hardware calls directly across the ISA bus and reads those memory locations without any concern for the CPU runtime state. 

When the game sets up the DMA transfer, is tells the DMA controller the main memory address it's writing data into, followed by how many elements its writing ("elements" being a generic quantity based on if the DMA channel is 16-bit or 8-bit). 

Ideally we could then copy that DMA transfer "object" into a thread safe space where the sound blaster could read from it fully async from the main thread, and paced out in time at the playback rate.

interloper98:
Ok! Ready for review.

interloper98:
Another on that we get right now, [X-mas 92 by Cascada](https://www.pouet.net/prod.php?which=4174) 

https://github.com/user-attachments/assets/31704242-8d19-4631-8056-2f3f5032cf94

src/hardware/sblaster.cpp
```diff
@@ -2858,21 +2865,25 @@ bool SB_GetAddress(uint16_t &sbaddr, uint8_t &sbirq, uint8_t &sbdma)
 
 static void sblaster_mixer_callback([[maybe_unused]]const int requested_frames)
 {
+	assert(sblaster && sblaster->channel);
```
johnnovak: Minor: I prefer separate asserts for such scenarios, but it's borderline 😄

src/hardware/sblaster.cpp
```diff
@@ -2932,7 +2943,7 @@ static void generate_frames(const int frames_requested)
 	{
 		// No-op if channel is already running
 		soundblaster_mixer_queue.Start();
-		sb.chan->Enable(true);
+					sblaster->channel->Enable(true);
```
johnnovak: Just run clang-format on this commit.

src/hardware/sblaster.h
```diff
@@ -0,0 +1,50 @@
...
+
+private:
+	IO_ReadHandleObject read_handlers[0x10]   = {};
+	IO_WriteHandleObject write_handlers[0x10] = {};
+
+	static constexpr auto BlasterEnvVar = "BLASTER";
+
+	OplMode oplmode = OplMode::None;
+
+	bool cms = false;
```
johnnovak, line 51: `is_cms_enabled` or `cms_enabled`?

src/hardware/sblaster.cpp
```diff
@@ -1230,6 +1247,46 @@ static void flush_remainig_dma_transfer()
...
+
+void CallbackTiming::SetPerTick()
+{
+	if (timing != TimingType::PerTick) {
+		Stop();
+		frames_added_this_tick = 0;
+		TIMER_AddTickHandler(per_tick_callback);
+		timing = TimingType::PerTick;
+	}
+}
```
johnnovak, line 1331: I'm not really seeing it, but does this handle fractional "samples per tick" values? In general terms, that would be needed I think. But if it's just about shuffling code around, and the existing code only dealt with truncated integers, maybe it's fine for now.

src/hardware/sblaster.cpp
```diff
@@ -126,6 +125,21 @@ enum class DmaMode {
...
 enum class EssType { None, Es1688 };
 
+class Dac {
+public:
+	std::optional<int> MeasureDacRateHz();
+	AudioFrame RenderFrame();
+
+private:
+	// 1% change threshold
+	static constexpr int ChangeTallyThreshold = 10;
```
johnnovak, line 136: So ` RateChangeThresholdPercent` is clear, but what about this? Is this about the rate change tally... or something else?

src/hardware/mixer.cpp
```diff
@@ -295,45 +289,30 @@ bool MIXER_FastForwardModeEnabled()
...
-	}
-	if (tandy_dac) {
-		tandy_dac->output_queue.Stop();
-	}
-	if (ps1_dac) {
-		ps1_dac->output_queue.Stop();
-	}
-	if (lpt_dac) {
-		lpt_dac->output_queue.Stop();
-	}
```
johnnovak, line 309: So we don't need similar changes for the LPT DAC devices?

src/hardware/sblaster.cpp
```diff
@@ -922,22 +922,62 @@ static std::array<uint8_t, 2> decode_adpcm_4bit(const uint8_t data)
...
+		return lut_s8to16[sample];
+	} else if constexpr (std::is_same_v<T, uint16_t>) {
+		return static_cast<int16_t>(le16_to_host(sample) -
+		                            Mixer_GetSilentDOSSample<T>());
+	} else if constexpr (std::is_same_v<T, int16_t>) {
+		return static_cast<int16_t>(
+		        le16_to_host(static_cast<uint16_t>(sample)));
+	}
+	static_assert(sizeof(T) <= 2, "Conversion is for 8 & 16-bit ints only!");
+	return 0.0f;
```
johnnovak, line 948: Sweet Jesus on a stick! 🥶 😆

src/hardware/sblaster.cpp
```diff
@@ -1343,6 +1345,13 @@ void SBLASTER::SetChannelRateHz(const int requested_rate_hz)
 	 }
 }
 
+// Wakeup the queue and channel to resume processing and playback
```
johnnovak, line 1353: minor: "Wake up"

src/hardware/sblaster.cpp
```diff
@@ -3016,50 +3022,16 @@ bool SB_GetAddress(uint16_t &sbaddr, uint8_t &sbirq, uint8_t &sbdma)
 
 static void generate_frames(const int frames_requested)
 {
-	assert(sblaster && sblaster->channel);
-
-	static std::vector<AudioFrame> empty_frames = {};
-	static int ticks_of_silence = 0;
+	assert(sblaster && sblaster->channel && sblaster->output_queue.IsRunning());
```
johnnovak, line 3030: IMO this assert is doing too much.

src/hardware/sblaster.h
```diff
@@ -32,6 +32,10 @@ class SBLASTER final {
 
 	void SetChannelRateHz(const int requested_rate_hz);
 
+	// Returns true of the channel was sleeping and was awoken.
+	// Returns false if the channel was already awake.
+	bool MaybeWakeup();
```
johnnovak, line 37: It's "wake up", so this should be called `MaybeWakeUp()`.

src/hardware/sblaster.cpp
```diff
@@ -819,7 +819,12 @@ static void dsp_dma_callback(const DmaChannel* chan, const DmaEvent event)
 			// ready for playback. This is when we set the callback running to
 			// play the data.
 			sblaster->MaybeWakeup();
-			callback_timing.SetPerTick();
+
+			constexpr auto PerTickCountThreshold = 4;
```
johnnovak, line 823: Any reason for picking the number 4? Could you explain in a comment?

johnnovak:
Heroic effort @interloper98, done with the review (...again; I think it's my 3rd pass over the last few months 😆)

src/hardware/sblaster.cpp
interloper98: Yup; added a comment and used a bit better name:

```c++
// If the DMA transfer is setup with a base count of fewer than
// three elements (which is four bytes given one 16-bit stereo frame
// held in an 8-bit DMA channel), then we know the software intends
// to overwrite the DMA content on the fly instead of pre-generating
// large chunks of DMA audio. In these cases we prefer the
// fine-grained per-frame callback. (The minus one is because DMA
// counts are in addition to one; so a base count of zero is one
// element).

constexpr auto MaxSingleFrameBaseCount = sizeof(int16_t) * 2 - 1;

(chan->base_count <= MaxSingleFrameBaseCount)
        ? callback_type.SetPerFrame()
        : callback_type.SetPerTick();
```

src/hardware/sblaster.h
interloper98: Fixed; thanks.

src/hardware/sblaster.cpp
interloper98: Yes; split them all.

src/hardware/sblaster.cpp
interloper98: Fixed; thanks.

src/hardware/sblaster.cpp
interloper98: :rofl: I've tried to make them a tad nicer:

```c++
// Convert sample to float based on type
template <typename T>
static constexpr float to_float(T sample)
{
	if constexpr (std::is_same_v<T, uint8_t>) {
		return lut_u8to16[sample];
	}
	if constexpr (std::is_same_v<T, int8_t>) {
		return lut_s8to16[sample];
	}
	if constexpr (std::is_same_v<T, uint16_t>) {
		return static_cast<int16_t>(le16_to_host(sample) -
		                            Mixer_GetSilentDOSSample<T>());
	}
	if constexpr (std::is_same_v<T, int16_t>) {
		return static_cast<int16_t>(
		        le16_to_host(static_cast<uint16_t>(sample)));
	}
	// compile-time checks to prevent the template being misused
	static_assert(std::is_integral_v<T>, "Conversion is only for integers");
	static_assert(sizeof(T) <= 2, "Conversion is only for 8 & 16-bit ints");
	return 0.0f;
```

src/hardware/mixer.cpp
interloper98: I don't think so; from my tests, they're working fine as-is.

src/hardware/sblaster.cpp
interloper98: Thanks; it wasn't that clear. 
Renamed slightly and added more comments.

```c++
public:
	// When the DAC is in use, we run the Sound Blaster at exactly the rate
	// the DAC is being written to and generate frame by frame. To support
	// this, we need to measure the rate the DAC is being written to.
	std::optional<int> MeasureDacRateHz();
	AudioFrame RenderFrame();

private:
	// We use two criteria to monitor and decide when the rate's changed:
	// percent difference (versus current) and when the new rate persists
	// across a sequential count. These two thresholds (1% change confirmed
	// across 10 sequential changes) were selected based on inspecting
	// actual games and demos (Alone in the Dark, Overload demo, EMF demo,
	// Cronolog demo, Chickens demo). These get the timing right without
	// whip-sawing or chasing.
	//
	static constexpr float PercentDifferenceThreshold = 0.01f;
	static constexpr int SequentialChangesThreshold   = 10;
```

src/hardware/sblaster.cpp
interloper98: Yup; I've renamed this slightly to be clearer that we're just managing to two callback types (that differ on timing), as opposed to the timing itself.

I've added more comments for each callback (and noting here the handling of fractional frames that @weirddan455 added in the per-tick handler).

```c++
// This callback is run once per emulator tick (every 1ms), so it generates a
// batch of frames covering each 1ms time period. For example, if the Sound
// Blater's running at 8 Hz, then that's 8 frames per call. Many rates aren't
// evenly divisible by 1000 (For example, 22050 Hz is 22.05 frames/millisecond),
// so this function keeps track of exact fractional frames and uses rounding to
// ensure partial frames are accounted for and generated across N calls.
static void per_tick_callback()
{

// This callback is run exactly once per frame, so it only generates a single
// frame each call. This allows the emulator (and game) to run between each
// frame, which is more accurate because changes to the Sound Blaster are
// reflected in the next frame.
//
// This approach is more costly to emulate so we use it only when conditions
// demand it, and prefer the leaner per-tick approach otherwise. However,
// if/when the Sound Blaster is moved off the main emulator loop then we can
// (ideally) use this all of the time.
static void per_frame_callback(uint32_t)
```

src/hardware/sblaster.h
interloper98: Turns out there's already a static function called `is_cms_enabled()`, and it's only used in one place during init (and shutdown is a no-op if CMS wasn't running), so this extra state bool was useless and dropped it.

src/hardware/sblaster.cpp
interloper98: Formatted; thanks for flagging it!

src/hardware/sblaster.cpp
interloper98: Definitely agree; more obvious when an assert trips. 
Switched them all.

interloper98:
> Heroic effort @interloper98, done with the review (...again; I think it's my 3rd pass over the last few months 😆)

Thanks for all the reviews; it's in a whole lot better shape.
Next up for SB. Hopefully full thread lift out with @weirddan455's magic touch.

src/hardware/sblaster.cpp
johnnovak: Awesome!

src/hardware/sblaster.cpp
johnnovak: Thanks, that's better 😄

src/hardware/sblaster.cpp
johnnovak: Great stuff, man! 🚀

src/hardware/sblaster.cpp
johnnovak: Very nicely explained, I really appreciate your effort to clearly document these things for the next guy 🎖️ 💯

johnnovak:
> Thanks for all the reviews; it's in a whole lot better shape!

Great stuff, and thanks for addressing my comments so thoroughly. Stand-out work, man, and you've added some excellent explanatory comments 💯 🥇 

I will give this a good last round of testing over the coming days, then it's merge time.

johnnovak (approved):
All my tests have passed @interloper98, great work! 🎖️ 💯 

I've added [Takeover by Epical](https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-emulation-tests#takeover-by-epical-sb2) to the SB tests, nice find!

Hit that big merge button when ready, man 😎 🤘🏻

interloper98:
It's been a epic grind; thanks for the reviews @johnnovak and @weirddan455 !

interloper98:
@johnnovak , @weirddan455 - it's showing "merging is blocked" pending requested changed by weirddan455. Double checked all the feedback and everything's resolved/closed. 

No rush; if there were some things I haven't covered can you flag them @weirddan455 ? I'm just not seeing them in the comments.

johnnovak:
I'm pretty sure it's about this comment of @weirddan455:

https://github.com/dosbox-staging/dosbox-staging/pull/3875#pullrequestreview-2238608593

interloper98:
> I'm pretty sure it's about this comment of @weirddan455:
> 
> [#3875 (review)](https://github.com/dosbox-staging/dosbox-staging/pull/3875#pullrequestreview-2238608593)

Yup; that comment is three months old (left Aug 15). Alone in the Dark is working fine and (I think :sweat_smile: ) I've covered any issues mentioned further in the comments. Will hold for @weirddan455.

src/hardware/sblaster.cpp
```diff
@@ -829,7 +829,20 @@ static void dsp_dma_callback(const DmaChannel* chan, const DmaEvent event)
...
+			// to overwrite the DMA content on the fly instead of pre-generating
+			// large chunks of DMA audio. In these cases we prefer the
+			// fine-grained per-frame callback. (The minus one is because DMA
+			// counts are in addition to one; so a base count of zero is one
+			// element).
+			constexpr auto MaxSingleFrameBaseCount = sizeof(int16_t) * 2 - 1;
+
+			(chan->base_count <= MaxSingleFrameBaseCount)
+			        ? callback_type.SetPerFrame()
+			        : callback_type.SetPerTick();
```
weirddan455, line 845: I still don''t understand what the benefit of these per frame callbacks are.  Theoretically it should end up doing exactly the same thing since these callbacks can only run at most once per 1ms.

I'm fine leaving it for now though.  You've got both methods in there so it should be an easy swap if this runs into any problems in the future.

weirddan455 (approved):
Didn't realize hitting "Request changes" would block a merge even after @johnnovak hit Approve.  Anyway, this is good to go as far as I'm concerned.  It's been open for long enough already :smile: 

Nice job with this @interloper98 :+1:

src/hardware/sblaster.cpp
interloper98: It looks that way on the surface, but each PIC event is scheduled to run at some fractional amount of milliseconds in the future. 

The fractional value is a double, and the event is run at exactly at that ratio of cycle position within the allotted total cycles.

So for example, if you schedule a pic event with a value of 0.5ms, and if the system is running at fixed 3000 cycles/ms, then the event will trigger at exactly the 1500-cycle mark. So you can get events happening inside the 1ms cycle tick.

If you scheduled it for 0.1, it would trigger at the 300-cycle mark, followed by 2700 more cpu cycles after the event before the emulator's full 1ms round is done. 

So to play audio from a memory address that Crystal Dreams is writting to at 16 KHz, we need the SB reading from DMA using PIC events that are each scheduled  1/16000s in the future.

Let's say dosbox is running Crystal Dreams at a cycle count of 10000 cycles/ms. Our per frame events will fire after every 625 cycles elapses, which is exactly 16 times within the 1ms emulator round (where it's doing 10k cycles/ms). 

If you print the PIC full index you can see that fractional time count by. That's how I worked backward to land on this approach. 

This differs from the timer callback system that only gives you 1ms granularity, and those callbacks are only run after the entire set of cycles is complete.

johnnovak: Nice explanation @interloper98 👍🏻

weirddan455: Got it.  I was under the impression that each 1ms tick always finished in its entirety before any PIC events could be run.  I take it you've looked into this more than I have though :smile:

PR #3878, opened 2024-08-17
It translation

farsil:
# Description

Further improvements to the Italian translation file.


farsil:
Sorry, wrong target. I do have some minor improvments scheduled but they are going to go through a different process.

PR #3879, opened 2024-08-18
Improve website

johnnovak:
# Description

Mostly just a bunch of website-related improvements. See the individual commits for details. This is planned to go live when we release 0.82.0-RC in two days, so you can suggest improvements/changes until then.

# Manual testing

I've deployed the changes to the preview website:
https://www.dosbox-staging.org/preview/dev/

# Checklist

I have:

### DIFF
--- a/docs/build-windows.md
+++ b/docs/build-windows.md
@@ -5,8 +5,10 @@ Windows builds can be created using:
 - Visual Studio 2022 IDE suite with Clang/LLVM compiler, and vcpkg to provide
   dependencies. This is the fully-supported toolchain used to create release
   builds.
-- The Clang or GCC compilers using the Meson buildsystem running within the
-  MSYS2 environment to provide dependencies.
+
+- The Clang or GCC compilers using the Meson build system running within the
+  MSYS2 environment to provide dependencies (deprecated, support will be probably
+  removed in the future).
 
 > **Note**
 >
@@ -19,6 +21,7 @@ Windows builds can be created using:
 
 1. Install Visual Studio Community 2022: <https://visualstudio.microsoft.com/vs/community/>.
     - Select "C++ Clang tools for Windows" in the Visual Studio installer
+
 2. Install/configure vcpkg.
     - Recent versions of Visual Studio already include it[^1], but it needs to
       be initialized. Open a Visual Studio Developer Command Prompt and run the
@@ -44,7 +47,12 @@ you build a binary optimized for gaming.
 2. Select a **Release** build type in Visual Studio, and run the build.
 
 
-## Build using MSYS2
+## Build using MSYS2 (deprecated)
+
+> **Note**
+>
+> MSYS2 builds are not actively maintained or tested by the development 
+> team. We will probably remove this section in the future.
 
 1. Install MSYS2: <https://www.msys2.org/wiki/MSYS2-installation/>
 
--- a/scripts/repackage-windows-zip.sh
+++ b/scripts/repackage-windows-zip.sh
@@ -11,12 +11,12 @@
 set -e
 
 if [ $# -ne 1 ]; then
-	echo "Usage: repackage-windows-zip.sh MSYS2_ZIP_NAME"
+	echo "Usage: repackage-windows-zip.sh CI_ZIP_NAME"
 	exit 1
 fi
 
 IN_ZIP=$1
-IN_ZIP_PREFIX=dosbox-staging-windows-msys2-x86_64-
+IN_ZIP_PREFIX=dosbox-staging-windows-x64-
 
 if [[ $IN_ZIP != $IN_ZIP_PREFIX* ]]; then
 	echo "Input filename must start with '$IN_ZIP_PREFIX'"
--- a/website/docs/about/index.md
+++ b/website/docs/about/index.md
@@ -41,7 +41,6 @@ The key features for developers are summarised [here](https://github.com/dosbox-
 - Deliver a consistent **cross-platform experience**.
 - Leverage **upstream** and **community developments** in DOSBox.
 - Encourage **new contributors** by removing barriers to entry.
-- Fix, clean up, and integrate notable **community-developed patches**.
 - Prioritise **code quality** to minimise technical
   debt and ease maintenance. This generally means following the
   [Staging Coding Style Guide](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#coding-style)
@@ -51,7 +50,7 @@ The key features for developers are summarised [here](https://github.com/dosbox-
 ## Non-goals
 
 - **Support old, end-of-life operating systems** (e.g., Windows 7 or older, OS/2,
-  or Mac OS X 10.5 or older) and limited CPU/memory hardware, which are
+  Mac OS X 10.5 or older) and limited CPU/memory hardware, which are
   constraints the original [DOSBox](https://www.dosbox.com/) continues to
   support.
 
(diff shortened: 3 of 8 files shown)


weirddan455 (approved):
This all looks good to me.  Thanks for keeping this stuff up to date :+1:

docs/build-windows.md
```diff
@@ -5,8 +5,10 @@ Windows builds can be created using:
 - Visual Studio 2022 IDE suite with Clang/LLVM compiler, and vcpkg to provide
   dependencies. This is the fully-supported toolchain used to create release
   builds.
-- The Clang or GCC compilers using the Meson buildsystem running within the
-  MSYS2 environment to provide dependencies.
+
+- The Clang or GCC compilers using the Meson build system running within the
+  MSYS2 environment to provide dependencies (deprecated, support will be probably
+  removed in the future).
```
interloper98, line 11: One interpretation of **_'support being removed'_** is that "the supported thing" will no longer be possible from the user's perspective. 

On the dev side, that means removing the tangible stuff that was supporting the thing: be it workflows, build system code, or code-base features.

For example, if Microsoft says support for WIN32 binaries is being removed from Visual Studio 2026, the implication is it won't be able to compile WIN32 binaries.

But when I check our repo:
 - There are no MSYS2-specific workflows in CI
 - There is no MSYS2-specific code in Meson
 - There is no MSYS2-specific code in the source files

So I don't think we'll be actively deleting stuff to eliminate the ability to compile under MSYS2.

Instead, we probably mean that **_MSYS2 is not actively supported_**, the key being that we're not "removing" stuff. 

Well - I suppose the build notes themselves might be deleted, however the underlying ability is still possible.  So worst case, we could move the MSYS notes into a `build-msys2.md` file and mark it "not actively supported/tested; maintainers welcome!".

(The key here is that there's no MSYS2 stuff to maintain, the only fixes would be to the notes themselve or generic code-base fixes, because the code base has zero MSYS2 branches and would instead follow the same path as Visual Studio, which also uses Clang).

docs/build-windows.md
```diff
@@ -44,7 +47,12 @@ you build a binary optimized for gaming.
 2. Select a **Release** build type in Visual Studio, and run the build.
 
 
-## Build using MSYS2
+## Build using MSYS2 (deprecated)
+
+> **Note**
+>
+> MSYS2 builds are deprecated; we will probably remove support for this in the
+> future.
```
interloper98, line 55: See comment above.

Suggest stating:

```
MSYS2 builds are not actively maintained or tested by the development team.

We will probably relocate the MSYS2 notes to a separate markdown file and mention that these builds are currently unmaintained.
```

interloper98: Or if we really want to delete the notes:

```
MSYS2 builds are not actively maintained or tested by the development 
team. We will probably removes these notes in the future.
```

docs/build-windows.md
weirddan455: >  The key here is that there's no MSYS2 stuff to maintain

Yes and no.  There's no specific MSYS2 stuff because it uses the same build system as Linux but with Windows headers.  In practice, it requires maintenance or stuff breaks.  Off the top of my head:

- It's rolling release and more bleeding edge than Arch Linux even.  Dependency and toolchain updates have broken the build multiple times with no changes on our side.
- The Windows header file definitions are slightly different than MSVC.  I've seen this cause compilation failure and the need to change for example a constant that was missing a `#define` in MSYS2.
- There was a weird bug where GCC had a different sleep implementation than MSVC that led to severe performance degradation.  Only affected GCC on MSYS2.  This was fixed but still these oddball things are possible.

>  Instead, we probably mean that MSYS2 is not actively supported, the key being that we're not "removing" stuff.

All that said, yeah maybe that wording makes more sense.  I don't think there's anything we can remove.  The whole point of MSYS2 is the ability to compile arbitrary POSIX programs.

For what it's worth, I do sometimes compile on MSYS2 whenever I'm on Windows but that's not very often so any bugs or build breaks are likely to linger due to lack of testing.

docs/build-windows.md
johnnovak: Yeah, I'll take this version, thanks. 

We'll just nuke these instructions at some point. I guess I could have done it now, but did not want to do it without @shermp agreeing to it (he created the now disabled MSYS2 workflow).

docs/build-windows.md
johnnovak: Yeah, as @weirddan455 said, probably it will be difficult to compile anything but main due to the rolling nature of MSYS2. Well, Linux has the same issues, sure, but we're solving that with vcpkg that pins dependecies for anything but truly system level libs.

Plus yeah, MSYS2 had other non-trivial problems too as @weirddan455 mentioned. All in all, we should not provide build instructions untested and unmaintained by the team because they will be most likely be broken, and that's wasting people's time. We also don't want take up the burden maintaining multiple build systems on Windows, so the logical conclusion is to remove the MSYS2 stuff altogether.

I can adjust this comment as per your below suggestion, the when we get a confirmation from @shermp whether he wants to singlehandedly maintain MSYS2 going forward, we'll decide its fate. But my current understanding is no one wants to keep MSYS2 around.

johnnovak: Also, personally, I'd remove all the other probably stale build instructions too if nobody tested they work in the last 12 months (all the weird Linux variants).

johnnovak: > * There are no MSYS2-specific workflows in CI

But we do:

https://github.com/dosbox-staging/dosbox-staging/blob/main/.github/workflows/windows-msys2.yml

Historical context: The MSYS2 workflow was probably the no.1 catalyst for revamping our whole build system and migrate to vcpkg because it caused so much pain to prop it up with weird hacks over time as it just kept breaking... 😅

johnnovak:
> * Legacy Windows - Win8/10 **32-bit** - last version 0.80.1
> 
>     * About - "Fix, clean up, and integrate notable community-developed patches." seems superfluous:
>       image

Good tips @Torinde. Adding a Win 8/10 section and removing that redundant bullet point.

docs/build-windows.md
interloper98: Good to know; thanks for this background info @weirddan455 and @johnnovak !

interloper98: Perfect use of the wiki; maybe users can share how to build under Steam or WSL.

interloper98 (approved):

docs/build-windows.md
johnnovak: Yeah, I appreciate without the historical context this removal might look like an arbitrary decision @interloper98, but we have our reasons 😄 The future end goals is vcpkg on all platforms + CMake instead of Meson.

I also get where you're coming from regarding "negative affirmative" statements such as "support for X is removed"—it looks weird out of context, no doubt. The high-level end-goal here is just provide "positive affirmative" instructions: this is how to build the software on all three platforms, this is what the CI workflows and the core team does, so this is 100% supported "official" build method. And then stop there 😄

johnnovak:
Thanks for the review guys, merging.

docs/build-windows.md
interloper98: We'll regarding only going with positive affirmations 👍

*"you didn't specifically say that HPUX 4.3 wasn't supported, therefore I demand support! :neckbeard:"* 😂

Definitely could be numberous if we had to list them all!

Linux_Distribution_Timeline

johnnovak: Wow, what a chart from hell @interloper98 !  😅 😈  Also, today I learned about the neckbeard emoji 🤣

PR #3885, opened 2024-08-21
Remove special case for joystick axis 2 and 5 and add detection for gamepad trigger buttons

### LINKED ISSUE #3883: Controller Right Stick Left\Right Axis Not Working (Regression)

_No response_

Windows 11

_No response_

### Relevant hardware info

_No response_

I am using a PS5 DualSense controller. 

In Staging 75.2 I can load DOOM go into the gameplay, load the dosbox mapper up, and map the keyboard left arrow to right stick left and keyboard right arrow to right stick right and it works no problem.

In 82 RC I can map to the controller for right stick left\right but in game it does nothing

I checked version between 75.2 and 82RC and found that 76 works, but from 77 it does not. 

One item I noted that changed in 77 was that prior to 77 the information in the mapper under "BIND" when I mapped to the right stick left\right axis was "axis2-" for left and "axis2+" for right. As of 77 this changed and both directions now state "axis2+".

Explain how to reproduce
1. Add joysticktype=4axis to conf
2. Plug in a DualSense PS5 controller
3. Load any game, I used DOOM
4. Get into gameplay and load the dosbox mapper up
5. Map the keyboard left arrow to the right stick left direction and keyboard right arrow to the right stick right direction
6. Test if does anything in game

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

Regression from 5c800f9677d1ec32a4d715f0dd13ac0d7ca1f1ce for the PS5 DualSense controller, although this commit does fix a bug with controllers with triggers mapped to axes 2 and 5.

These trigger buttons are actually analogue triggers with a resting position of around -32000

Previous code had a special case for these hard-coded to axis 2 and 5 Some controllers, such as the PS5 DualSense have these axes mapped to the analogue stick

Instead, probe the axes on init to get their resting position

## Related issues

Fixes #3883

# Release notes

Fix right analogue stick bindings in the mapper UI for the PS5 DualSense controller.

# Manual testing

Tested with my Xbox One S controller.  Behavior is the same as on main with axis 2 and 5 being detected as trigger buttons.

# Checklist

I have:


interloper98 (approved):
Really nice fix and thorough comment :rocket:

src/gui/sdl_mapper.cpp
```diff
@@ -642,6 +642,15 @@ class CStickBindGroup : public CBindGroup {
 			LOG_MSG("SDL: Can't detect axes; %s", SDL_GetError());
 		axes = clamp(sdl_axes, 0, MAXAXIS);
 
+		// Axis position is reported as a signed 16 bit integer
+		// On a typical gamepad, analogue sticks have a resting position of 0
+		// Trigger buttons have a resting position close to -32000
+		// Probe the axes and if we have one with a very low value, we're probably dealing with an analogue trigger
+		// Hope the user isn't mashing their controller while we're initalizing :)
```
johnnovak, line 649: Yeah, this is as good as it gets until do the proper [GameControllerDB](https://github.com/mdqinc/SDL_GameControllerDB) integration.

Nice comments!

johnnovak:
@Python-Exoproject Could you please confirm that this fixes your problem with your PS5 DualShock controller?

To download the Windows build for this PR (maybe you're already know all this, but in case you don't, here are the instructions because this is very useful to confirm a fix before it gets merged):

1. Go to the top of this page and click on the **Checks** tab

image

2. Click on **Windows MSVC builds** in the left pane:

image

3. Scroll down to the bottom of the page and download the x64 portable ZIP artifact (the middle one):

image

johnnovak (approved):
Very nice fix @weirddan455, and extra bonus points for including the release notes 🎉 

Let's wait until @Python-Exoproject confirms this fixes his problem, then merge time 🥳

weirddan455:
>  The right bottom trigger (axis 4) is picked up as axis4- when pulled. Its acting like its continuously being pulled though which creates a couple of issues for mapping as its the main trigger button.

Hmm... sounds like it's not detecting the PS5 trigger buttons correctly then.  Trigger buttons should always be mapped at the positive axis to avoid that issue.  I'm not sure how that particular controller behaves though.

I wrote this very simple SDL program to test my Xbox trigger buttons to see what the range is.  You can compile with `gcc test.c -o test -lSDL2`.  If you're on Windows it's a bit more of a pain.  I can make you a Windows .exe later if you need.

```c
#include <SDL2/SDL.h>

int main(void)
{
    SDL_Init(SDL_INIT_JOYSTICK);
    SDL_JoystickOpen(0);
    SDL_Event event;
    while (SDL_WaitEvent(&event)) {
        if (event.type == SDL_QUIT) {
            return 0;
        }
        if (event.type == SDL_JOYAXISMOTION) {
            printf("%hhu: %hd\n", event.jaxis.axis, event.jaxis.value);
        }
    }
}
```
Just press the trigger slowly and let me know what it logs to console.  Mine starts at around -32,000 at the top and then bottoms out at positive 32,000 with 0 being around the center.  I wrote the trigger detection with that assumption in mind.

Python-Exoproject:
Happy to test but yeah im on windows and have no idea about compiling anything, not my wheelhouse

weirddan455:
Ok @Python-Exoproject here's a compiled version of the code.  It's a console program that just writes out the current values of the different axes as you move them.  Just move your trigger buttons slowly and let me know what the minimum and maximum values are and how they correspond to how far you have the trigger button pressed.

If you don't see anything, you may need to launch the .exe from a command prompt but on my system it brought one up just by double-clicking it so you may not need to.

Interestingly, I also stumbled on an SDL bug while testing this where on Windows I had to pass `SDL_INIT_VIDEO` or SDL would not report all of the axes (but still reported some of them).  Included the updated code in the .zip as well.

[joy.zip](https://github.com/user-attachments/files/16705532/joy.zip)

Python-Exoproject:
The right trigger starts at -32768 and when fully pulled end at 32767. 

If it matters, I do note that the axis labels seem reversed compared to what Staging was reporting

Python-Exoproject:
> Ok, so that's the same as my Xbox controller. I'm not sure why it would get bound as a negative axis then. Just to confirm, did you redo your bindings after downloading the build from this PR?

Yes, each time I delete the previous .map file and start fresh

> Also, is your left trigger working?

Same issue as the right trigger. I have tested on two controllers with the same results as below for both

> Reversed how? What are your axis labels exactly?

Ok, heres the full list. 

Im using Staging 75.2 for the labels as per Staging since in 82 it just says "DualSense Wireless Controller" for all binds with no description of the button or axis used. Either trigger acts as if always being pulled outside of mapper once mapped in Staging 75.2 and 82 causing the bound action to be "pressed" continuously and when actually pulled the bound action will stop while held.

| Staging mapper  | Axis |
| ------------- | ------------- |
| Left Stick left | 0-  |
| Left Stick right  | 0+  |
| Left Stick up | 1-  |
| Left Stick down  | 1+  |
| Right Stick left | 2-  |
| Right Stick right  | 2+  |
| Right Stick up | 5-  |
| Right Stick down  | 5+  |
| Left main trigger | 3- (if pulled hard and quick it reads as button 6 occasionally)  |
| Right main trigger  | 4- (if pulled hard and quick it reads as button 7 occasionally) |


| Joy app  | Axis |
| ------------- | ------------- |
| Left Stick left | 1  |
| Left Stick right  | 1  |
| Left Stick up | 0  |
| Left Stick down  | 0  |
| Right Stick left | 3  |
| Right Stick right  | 3  |
| Right Stick up | 2  |
| Right Stick down  | 2  |
| Left main trigger | 4  |
| Right main trigger  | 5  |

Python-Exoproject:
I just realised that I can read the mapper file directly for the axis so here is the values as per the Staging 82 mapper file:

| Staging 82 mapper  | Axis |
| ------------- | ------------- |
| Left Stick left | 0-  |
| Left Stick right  | 0+  |
| Left Stick up | 1-  |
| Left Stick down  | 1+  |
| Right Stick left | 2-  |
| Right Stick right  | 2+  |
| Right Stick up | 3-  |
| Right Stick down  | 3+  |
| Left main trigger | 4-  |
| Right main trigger  | 5- |

Also, the fix of changing the "0" to a "1" for the trigger mappings works as expected as a temporary fix

Grounded0:
Seems like the green button can be pressed on this one.

Python-Exoproject:
Well the right stick was fixed but I believe @weirddan455 was looking into why the triggers were being detected as a negative axis when they should be a positive axis in order to avoid continual fire. 

This is important as when a mapper is loaded that uses one of the triggers mapped to something like `ctrl` (which is the shoot button in many FPS) and it hasnt manually been fixed it blocks any keyboard input during the eXoDOS screens that ask things like what version of a game to launch or what sound device. This block happens as dosbox thinks `ctrl` is being held down and so when a user types `n` to answer no to a question dosbox thinks they are pressing `ctrl + n` and doesnt proceed.

weirddan455:
Don't merge yet.  I think it might be something specific to Windows and I need to do some more testing.

>  here is the values as per the Staging 82 mapper file

Glad you posted that follow-up.  I was really confused as to why you were getting completely different axis numbers from my test program but it's probably due to the 0.75.2 build you tested with using an old SDL version.  At least current version maps to the same axis numbers :sweat_smile:

I may have to come up with a better method of checking for trigger buttons.  It's possible I can use the Gamepad API just in the init part.  The newer Gamepad API still uses the lower-level Joystick API under the hood.

It looks like SDL has a function to get a Gamepad handle from a Joystick handle so I might be able to use that strictly for detection without refactoring a bunch of stuff.

I'll just need to work on this a bit more under Windows when I have some time.

weirddan455:
I was able to reproduce the problem on Windows.  The probing for resting values on init does not work in Windows.  It seems to report all of the axis values as 0 until it receives an input event.

I re-wrote the trigger detection logic to use `SDL_GameControllerMappingForDeviceIndex()` which takes in a joystick index and returns a gamepad mapping in the form of a string.  I parse that string and now we know where the trigger buttons are.

I tested that this works on both Windows and Linux.  On Windows my Xbox controller actually has the same axis numbers as @Python-Exoproject reported on his PS5 controller.  On Linux it is different.

I think this is good to go now since I was able to reproduce everything on my machine but I'll let @Python-Exoproject test if he wants to.

src/gui/sdl_mapper.cpp
```diff
@@ -921,6 +925,49 @@ class CStickBindGroup : public CBindGroup {
...
+		char* sdl_mapping = SDL_GameControllerMappingForDeviceIndex(stick_index);
+		if (!sdl_mapping) {
+			return;
+		}
+		SetTriggerButtonFor("lefttrigger:", sdl_mapping);
+		SetTriggerButtonFor("righttrigger:", sdl_mapping);
+		SDL_free(sdl_mapping);
+	}
+
+	bool is_trigger[MAXAXIS] = {};
```
interloper98, line 969: As long as I've used DOSBox, I've never been able to get consistent mappings across controller brands: if you change controller brands then you have to re-do your mappings from the ground up.

I understand this is why eXoDOS can't ship joystick mappings.

If your approach here can also cover the buttons and axis, then we'd have 100% normalization, and finally users would have portable and sharable mappings independent of the controller make and model.

(I realize there's a plan to re-write the controller, but looking at this code - it's something that could be used in 0.83 and used in the re-write, too)

interloper98 (approved):

src/gui/sdl_mapper.cpp
weirddan455: This doesn't actually address that problem.  It only marks what axis are trigger buttons because they need a special case to handle their input.

To get real controller normalization we would need to use SDL's gamepad API everywhere (this is what provides the mappings that I use here) which would mean both a refactor of the mapper code and breaking everyone's existing mapper files.

interloper98: Thanks for that explanation; (darn; was hoping there was an easy silver bullet in here! I guess the move to the game controller DB is still the one true solution)

johnnovak: Let's not overdo it @interloper98. This will all be nuked. We only do surgical fixes until this is rebuilt, and this PR already goes beyond that (but it's a welcome change).

In the new mapper, we'll have proper SDL ControllerDB support. Let's not invent alternative home-grown mechanisms until then, that's my view.

interloper98: Yeah; I was seeing a solution when there wasn't one. 

1. Load DB
2. Assign axis from DB
3. Assign buttons from DB
4. Assign triggers from DB

(In about as many lines of code, too). But turns out it's not as simple or possible; Here's to dreaming! 🤷‍♂️ 😆

johnnovak:
Does the release notes section need to be updated @weirddan455 ?

src/gui/sdl_mapper.cpp
```diff
@@ -921,6 +925,49 @@ class CStickBindGroup : public CBindGroup {
...
+		if (!substring) {
+			return;
+		}
+		substring += strlen(button_name);
+		if (*substring != 'a') {
+			// Not an axis so no need to do anything
+			return;
+		}
+		++substring;
+		if (!(*substring >= '0' && *substring <= '9')) {
```
johnnovak, line 943: `isdigit()`

weirddan455: Changed

johnnovak (approved):
Very nice work @weirddan455 🎉 Just update the release notes please as discussed.

johnnovak:
Excellent release notes @weirddan455; it tells the users what they need to know in a clear, simple language. 🎖️

PR #3893, opened 2024-08-23
Ensure logs are UTF-8 on Windows, improve deprecated config warning log

FeralChild64:
# Description

Test case:
- set `language = pl` in the config file
- run DOSBox
- execute `keyb pl`
- execute `config -set cycles=max`

Result on Linux and possibly other systems:

Log-Before

Result after the change:

Log-After

# Manual testing

- tried to reproduce the problem from the description
- executed DOSBox with `--list-countries` switch, with language set to `pl` in the configuration file

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -76,6 +76,30 @@
 #include "vga.h"
 #include "video.h"
 
+static void switch_console_to_utf8()
+{
+#if WIN32
+	constexpr uint16_t CodePageUtf8 = 65001;
+	if (!sdl.original_code_page) {
+		sdl.original_code_page = GetConsoleOutputCP();
+		// Don't do anything if we couldn't get the original code page
+		if (sdl.original_code_page) {
+			SetConsoleOutputCP(CodePageUtf8);
+		}
+	}
+#endif
+}
+
+static void restore_console_encoding()
+{
+#if WIN32
+	if (sdl.original_code_page) {
+		SetConsoleOutputCP(sdl.original_code_page);
+		sdl.original_code_page = 0;
+	}
+#endif
+}
+
 #if C_OPENGL
 //Define to report opengl errors
 //#define DB_OPENGL_ERROR
@@ -288,6 +312,7 @@ static void QuitSDL()
 		SDL_Quit();
 #endif
 	}
+	restore_console_encoding();
 }
 
 // Globals for keyboard initialisation
@@ -4644,7 +4669,7 @@ static void list_glshaders()
 {
 #if C_OPENGL
 	for (const auto& line : RENDER_GenerateShaderInventoryMessage()) {
-		printf_utf8("%s\n", line.c_str());
+		printf("%s\n", line.c_str());
 	}
 #else
 	fprintf(stderr,
@@ -4656,7 +4681,7 @@ static void list_glshaders()
 static void list_countries()
 {
 	const auto message_utf8 = DOS_GenerateListCountriesMessage();
-	printf_utf8("%s\n", message_utf8.c_str());
+	printf("%s\n", message_utf8.c_str());
 }
 
 static int print_primary_config_location()
@@ -4737,6 +4762,11 @@ static void set_wm_class()
 
 int sdl_main(int argc, char* argv[])
 {
+	// Ensure we perform SDL cleanup and restore console settings
+	atexit(QuitSDL);
+
+	switch_console_to_utf8();
+
 	CommandLine command_line(argc, argv);
 	control = std::make_unique<Config>(&command_line);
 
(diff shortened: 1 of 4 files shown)


shermp:
Just a thought. Instead of modifying Loguru, why not just set the console output to "UTF-8" in sdlmain before initializing loguru?

FeralChild64:
@shermp Actually, decided to try with your proposal - PR set to draft until I do some tests. Original fix available on https://github.com/dosbox-staging/dosbox-staging/tree/fc/utf8-logs-1-old branch for now.

shermp:
I'm pretty sure I've seen a giant try / catch in sdlmain that could be used to hopefully ensure the console output is reset.

johnnovak:
> @shermp I was considering this, but that would highly increase the risk that in case of crash we'll leave the console in the altered state. But if it is OK for the team, I can do exactly as you propose.

Yeah, that's perfectly fine IMO.

src/gui/sdlmain.cpp
```diff
@@ -76,6 +76,30 @@
 #include "vga.h"
 #include "video.h"
 
+static void switch_console_to_utf8()
+{
+#if WIN32
+	constexpr uint16_t CodePageUtf8 = 65001;
+	if (!sdl.original_code_page) {
+		sdl.original_code_page = GetConsoleOutputCP();
+		// Don't do anything if we couldn't get original code page
```
johnnovak, line 85: "the original code page"

src/gui/sdlmain.cpp
```diff
@@ -4737,6 +4762,11 @@ static void set_wm_class()
 
 int sdl_main(int argc, char* argv[])
 {
+	// Ensure we cleanup the SDL and restore console settings
```
johnnovak, line 4765: "we perform SDL cleanup" or something

src/misc/setup.cpp
```diff
@@ -1035,8 +1035,9 @@ bool Section_prop::HandleInputline(const std::string& line)
 		}
 
 		if (p->IsDeprecated()) {
-			LOG_WARNING("CONFIG: Deprecated option '%s'", name.c_str());
-			LOG_WARNING("CONFIG: %s", p->GetHelp().c_str());
+			LOG_WARNING("CONFIG: Deprecated option '%s'\n\n%s\n",
+			            name.c_str(),
+			            p->GetHelpUtf8().c_str());
```
johnnovak, line 1040: Hmm, so this would make deprecation messages appear like this in the logs:

image

Instead of this:

image

I'm on the fence about the new multiline format. What do you guys think?

src/gui/sdlmain.cpp
FeralChild64: Fixed.

src/gui/sdlmain.cpp
FeralChild64: Fixed.

src/misc/setup.cpp
FeralChild64: It already used to be multiline, in fact - just strangely formatted :)

johnnovak: Well, previously we had multiple single-line log statements. Now we have a single multi-line log statement — that's the difference.

interloper98: I also like the new format.
The gap makes the warning stand out for easier reading.
(Plus.. warnings should be addressed anyway, so once it's solved then the log I back to line-by-line)

weirddan455 (approved):
This all looks fine as far as I'm concerned.  I don't have a strong opinion on the multi-line logs.  Whatever you guys decide, I'm fine with :smile:

interloper98 (approved):

src/misc/setup.cpp
johnnovak: Yeah, okay, let's keep the new multiline format then.

I was thinking of the future when these notifications would surface on the user interface, but I'll deal with that problem then.

Arguably, having to look at the logs for warnings and informational messages is a terrible, terrible user experience — all these warnings should be GUI notifications. While the logs can be useful for debugging for power users, regular users should never _have_ to look at the logs like ever in their lives, and should be able to use Staging effectively.

johnnovak (approved):
Nice work @FeralChild64, merge away 🚀

src/misc/setup.cpp
FeralChild64: Well, once the OSD is available, a lot more will have to change/improve, not only this little non-critical warning :)

johnnovak: > Well, once the OSD is available, a lot more will have to change/improve, not only this little non-critical warning :)

Yeah, no doubt 😆 Hence I'm like, yeah, do whatever for now 😆

Issue #3891, opened 2024-08-23
Mouse non-functional in 3dfx version of DESCENT 2


x73rmin8r:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

84963 dev build and RC1 0.82.0

### What Operating System are you using?

Windows 11

### Relevant hardware info

r5 5600x, 7900xt 

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

There is no mouse input when D2VOODOO.EXE is run, either movement or clicks (the game itself runs with 3d acceleration as expected though. keyboard works). Mouse works properly when normal DESCENT2.EXE is run during the same instance of DOSBOX. As far as I can tell the two executables use the same settings configuration.

I looked around for general mouse issues involving D2VOODOO on real hardware but wasn't able to find anything. It's possible that I've just got something set up wrong, but I haven't seen mouse issues with any other games I've tried. I don't think I have any other 3dfx capable DOS games to test.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Descent 2 CDs are version 1.0, GOG release is 1.1
2. patch Descent 2 to version 1.2 (https://archive.org/details/D2PATCH)
3. copy 3dfx patch files into game directory (https://archive.org/details/D2VOODOO)
4. run D2VOODOO.EXE, verify no mouse input

### Download URL of affected game or software

https://archive.org/details/descent-2.iso

### Your configuration

```ini
https://gist.github.com/x73rmin8r/9dc9d454a74c7f15cf8c9915771286a7.js
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Grounded0:
Doesn't happen with Carmageddon while in 3dfx mode so probably game specific if its our bug. Tested with 0.82.0 RC1.

FeralChild64:
Does it work if you start CtMouse driver before running the game (this could be a possible workaround)?
Another option you can try to workaround the problem is to use the original Microsoft mouse driver - but  it requires `dos_mouse_driver = false` in the DOSBox config file.

Grounded0:
First step is to grab a copy from eXoDOS and see does it work. We've spent so much time chasing ghosts here when the game was just set up poorly.

FeralChild64:
@Grounded0 I don't use eXoDOS or similar packages, I prefer to install everything by myself.

[...]

PR #3896, opened 2024-08-24
  [post-0.82.0] Fix mouse support in Descent II with a Voodoo patch

FeralChild64:
# Description

A one-line fix + comments.

_Descent II_ version 1.2 with a _3Dfx Voodoo_ patch (the official one) is very picky regarding mouse driver. From my tests it only works with driver by Microsoft/IBM, version 7 or later - it doesn't work with _A4Tech_, _Genius_, _DR DOS_, _Mouse Systems_, _CuteMouse_, or _VBADOS_ drivers.

I've checked that with _DOSBox Staging_ the game the game just calls INT 33 function 0x00 (reset the driver and check it's capabilities), then function 0x20, afterwards there are no further mouse driver calls.

The game source code (https://github.com/OpenSourcedGames/Descent-2/blob/master/SOURCE/BIOS/MOUSE.C#L287) reveals:

```c
	// Enable mouse driver
	memset( &inregs, 0, sizeof(inregs) );
	inregs.w.ax = 0x0020;
	int386(0x33, &inregs, &outregs);
	if (outregs.w.ax != 0xffff )
		return 0;
```

despite all the documentation I have claims that function 0x20 returns nothing. I have checked with a test program (attached below) - and indeed, compatible mouse drivers return `AX=0xffff.

[TESTDESC.tar.gz](https://github.com/user-attachments/files/16735776/TESTDESC.tar.gz)

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3891

# Release notes

Fixed a problem with mouse not working in _Descent II_ version 1.2 with the official _3Dfx Voodoo_ patch.

# Manual testing

Tried to reproduce the PR, played a few games which are using a mouse.

# Checklist

I have:

### DIFF
--- a/src/hardware/input/mouseif_dos_driver.cpp
+++ b/src/hardware/input/mouseif_dos_driver.cpp
@@ -1620,12 +1620,20 @@ static Bitu int33_handler()
 		state.hidden    = 1;
 		// According to Ralf Brown Interrupt List it returns 0x20 if
 		// success,  but CuteMouse source code claims the code for
-		// success if 0x1f. Both agree that 0xffff means failure.
+		// success is 0x1f. Both agree that 0xffff means failure.
 		// Since reg_ax is 0x1f here, no need to change anything.
+		// [FeralChild64] My results:
+		// - MS driver 6.24 always returns 0xffff
+		// - MS driver 8.20 returns 0xffff if 'state.enabled == false'
+		// - 3rd party drivers I tested (A4Tech 8.04a, Genius 9.20,
+		//   Mouse Systems 8.00, DR-DOS driver 1.1) never return anything
 		break;
 	case 0x20: // MS MOUSE v6.0+ - enable mouse driver
 		state.enabled = true;
 		state.hidden  = state.oldhidden;
+		// At least Descent II with Voodoo patch expects this. Checked
+		// that MS driver alters AX this way starting from version 7.
+		reg_ax = 0xffff;
 		break;
 	case 0x22: // MS MOUSE v6.0+ - set language for messages
 		// 00h = English, 01h = French, 02h = Dutch, 03h = German, 04h =


FeralChild64:
@joncampbell123, @javispedro - you might be interested in this.

interloper98 (approved):
Great find @FeralChild64, and very helpful documentation 👍

Looking back at very early commits: 9ac71b742055546caae85753e4c267b5ff, they had int33 ax 0x21 doing a mouse 'software reset', which returned 0xffff in ax but also set bx to the number of buttons! Interesting:-)

I wonder if the MS driver is doing those, too?

Grounded0:
It would be good if someone digs up the commit that changed it and the reason why it was changed.

FeralChild64:
20 years have passed since this commit…

I only fear that this 0xFFFF actually means something, or that I just broke some forgotten late 80’s software. In such case we’ll need to introduce a new config option for this behavior.

Grounded0:
Or write a new game issues entry for Descent 2 3dfx.

javispedro:
Well, MS doesn't document the return value of int33/ax=0x20, but RBIL does document it to return "ax=0x20" on success and "ax=ffff" on failure. 

My unverified guess is that this game has a bug. 0x20 is supposed to re-enable the mouse interrupts but it doesn't call 0x19 which is required to disable it in the first place. So I guess MSMOUSE will just return failure / -1 on this case? Which happens to make the game happy...

VBADOS doesn't support 0x19/0x20 at all :)

FeralChild64:
OK, with Microsoft driver the 0x1f indeed sometimes returns -1 (if I start the test tool from the DOS Navigator), and sometimes doesn't change anything. Back to draft, the behavior seems to be more complicated than I thought.

javispedro:
Anyway, for me on msmouse 8 ax=0x20 always seems to return -1, even when it actually works and restores the handlers.... so this patch is good enough as an approximation.

FeralChild64:
Things starts getting interesting:
- Microsoft driver version 6.02 - `0x1f' never returns
- Microsoft driver version 6.24 - `0x1f` always returns -1, `0x20` never returns anything
- Microsoft driver version 8.20 - `0x1f` returns -1 if already disabled, `0x20` always returns -1
- 3rd party drivers (Mouse Systems, Genius, DR-DOS, A4Tech)  never return anything

So the behavior should be a correct approximation indeed, but I want to give some more comments.

interloper98:
> I only fear that this 0xFFFF actually means something, or that I just broke some forgotten late 80’s software. In such case we’ll need to introduce a new config option for this behavior

We'll have to take that risk 😌 and see what happens. If there are older games impacted, then the msmouse driver version could be available to get the 6.x behavior versus newer 8.x behavior, and even `= auto` would try to provide the most compatible behavior (real-mode apps get 6.x and protected-mode gets 8.x).

Grounded0:
I would like to see commit that first changed 2003 behaviour checked in by Harekiet and why it was done. The facts probably lie there.

javispedro:
I think that at no point has dosbox returned -1 for ax=0x20. Neither has cutemouse, fwiw. 
What is quoted above is returning -1 for ax=0x21, which is normal behavior.

interloper98:
> I think that at no point has dosbox returned -1 for ax=0x20. Neither has cutemouse, fwiw. What is quoted above is returning -1 for ax=0x21, which is normal behavior.

My fault in mentioned `0x21` above - I saw the series of 0x21, 0x22, 0x23, 0x24, and 0x25 and thought 0x21 was simply not implemented (https://github.com/dosbox-staging/dosbox-staging/pull/3896/files) ; I see it's now near the top and the cases aren't fully in sort-order. (please ignore my comment above @Grounded0!)

johnnovak:
> > I only fear that this 0xFFFF actually means something, or that I just broke some forgotten late 80’s software. In such case we’ll need to introduce a new config option for this behavior
> 
> Agree; and we'll have to take that risk 😌 and see what happens.
> 
> I guess if older games are impacted, the msmouse driver version could become a conf option (6.x behavior versus newer 8.x behavior), with `auto` providing the most compatible behavior (real-mode apps get 6.x and protected-mode gets 8.x).. or something like that.

We should tread carefully. The problem is a change like this could break hundreds or thousands of games in eXoDOS.

The "see what breaks" approach is not practical as no one will go and see what breaks out of those 7k games... 😅 Existing users could be super pissed that the mouse stopped working in their favourite games, and if we can only say "but the 3dfx version of Descent II now works without having to load a specific mouse driver", well that's a hard sell 😆 

Our default mouse driver is a "generalised" driver that has proven to work with most games over the last 20 years. We can make changes to it when the issues are clear cut and the benefits outweigh the risks, of course, but I don't think that is the case here.

Introducing such a risk for a single game is not worth it IMO, especially when there is a clear solution. I wouldn't even call it a workaround; this late DOS era game was probay only tested with that late 90s Microsoft mouse driver, so it needs that as our default mouse driver seems to be conflicting with it slightly.

Also, we don't introduce _more_ config options to get a single game running if there are other workarounds.

To be clear:

- If at least 10 games could benefit from this, (and they aren't just bottom of the barrel Spanish shareware football manager games that 5 people played ever 😏), a config option could be considered.
- If it was 100% clear this change is safe and would improve _overall_ mouse compatibility for the _entire_ DOS gaming catalog, I'm all for it.
- As it stands now, with the current info I have, I'm heavily in favour of _not_ merging this change. Just document it on the wiki if this is a single outlier.

So yeah, please do a cost/benefit analysis in such ambiguous cases when fixing one game could break thousands of other games, and think of the testing effort. eXoDOS is our biggest user; we don't want to break shit randomly with exploratory changes/fixes. In such cases, saying "I've tested a few other games" is not good enough for such exploratory changes.

We don't wanna play a game of whack a mole for months...

Grounded0:
I'm going to write Game issues entry on this now.

Grounded0:
- https://github.com/dosbox-staging/dosbox-staging/wiki/Game-issues#descent-1-and-2

interloper98:
Just providing some thoughts and comments @johnnovak ; I understand the concern is that we're not sure if Microsoft's 6.x driver return value is more or less compatible than Microsoft's 8.20 return value; and that testing all the mouse'able games in eXoDOS would need an army of testers for months.

---

But here's are my thoughts and some investigation.

This PR feels very similar to what I did with the Silmarils `ADLIB.COM` driver PR, where the result was upgrading DOSBox Staging to match the behaviour/state of the official commercial TSR/driver.

This one upgrades Stating to match Microsoft's v8.20 return value for this specific call. It turns out the v8.20 driver is a compatibility beast among hardware DOS users:

- https://www.vogons.org/viewtopic.php?p=654156#p654156 - _"If you have free UMB, then use Microsoft mouse 8.20, it is more compatible, and especially if you use Windows 3.1x."_

- https://www.vogons.org/viewtopic.php?p=654177#p654177 - _"Carrier Command doesn't like Cute Mouse. It wants a Microsoft driver specifically."_

- https://www.vogons.org/viewtopic.php?p=1004383#p1004383 - _"I run CuteMouse 2.0/2.1 and Microsoft Mouse 8.20 DOS drivers. I can't reproduce the problem"_

- https://www.vogons.org/viewtopic.php?p=774809#p774809 - _"IBM/Microsoft driver ver.8.20 - success!"_ (and reports many other drivers failed)

- https://www.vogons.org/viewtopic.php?p=1100091#p1100091 - lists specific driver versions that work with Day of the Tentacle _"CuteMouse 1.9.1 (later versions lock up my PC immediately), Microsoft MOUSE.COM 6.02 (the oldest MS driver I have), Microsoft MOUSE.COM 8.20 (the newest MS driver I have)"_

- https://www.vogons.org/viewtopic.php?p=661306#p661306 - Monkey Island talkie issues and problems with mouse drivers; user mentions: _"At least in my case the answer seems to be to use the Microsoft Mouse Driver (8.20)"_

- https://www.vogons.org/viewtopic.php?p=1261535#p1261535 - _"Try MSCDEX. That shsucdx is nice but is really not 100% compatible and sometimes gives trouble. The same thing is with Microsoft mouse 8.20 versus cutemouse."_

- https://www.vogons.org/viewtopic.php?p=1183216#p1183216 - discussion about drivers in a mouse pack _"I prefer 8.2 for compatibility"_

- Blue Byte Software [Settlers Gold documentation](https://web.archive.org/web/20070222103617/http://www.bluebyte.net/eng/products/settlers2gold/tips_techfaq.htm), Discussion about broken mouse behaviour: _"Please make sure you have the latest version of your mouse driver. In most cases this will solve the problem. We recommend Microsoft®'s mouse.exe version 8.20. (Later versions produced by Microsoft® should also work correctly.)"_

---

> the default mouse driver is a "generalised" driver that has proven to work with most games over the last 20 years. 

Staging's overhauled mouse driver that @FeralChild64 added is even better because it fixes some long standing stubborn games that the pre-DOSbox-Staging mouse implementation did not. Here are some: [[1](https://github.com/exoscoriae/eXoDOS/issues/5151)], [[2](https://github.com/exoscoriae/eXoDOS/issues/4989)], [[4](https://github.com/exoscoriae/eXoDOS/issues/693)], [[5](https://github.com/exoscoriae/eXoDOS/issues/4191)]. [[6](https://github.com/exoscoriae/eXoDOS/issues/5265)]

Comments on Reddit also mention that Staging, specifically, has near hardware-grade mouse handling:

reddit-dos-gaming-86box-July2024

---

> We don't wanna play a game of whack a mole for months

I would say that if a single mole needs whacking, then it would justify a conf setting, something like `ms_mouse_version = 6 / 8 / (auto?)`

Or if we do a thought exercise and invert the scenario :bulb: :sweat_smile: 

1. Let's assume the baseline code happened to implemented the v8.20 return value, and we're equality confident that it generally very compatible and we're just as risk adverse.
2. @FeralChild64 submits a PR switching the mouse behavior to implement v6.x empty return value, to fix some unknown game.  
3. But now here the twist! An army of testers appears and they start testing eXoDOS games. Within a couple days, one of them declares this PR is bad because it breaks Descent II mouse handling.
4. What would we decide to do? My guess is we'd agree that we need a conf setting to let users switch between these two major Microsoft version return values.

(Not sure if that helps or not; but I love the mouse work and it's only getting better!)

Grounded0:
With current implementation we're only aware of Descent II with 3dfx Voodoo update not working. With v7.04+ behaviour we haven't got data. I think this probably should be merged once 0.82.0 is released so it can start gathering data from users of development builds. Roll back if its shite.

FeralChild64:
Just to clarify some doubts:

- this PR makes the `0x20` call behave in a way compatible with Microsoft mouse driver 7.04 and all the later revisions I have tested; driver 7.04 was released in... May 1990 :)
- not that many PC games were released before that date (compare the numbers: [80s on dosgames.com](https://www.dosgames.com/listing.php?sortby=nameasc&sterms2=&cat=all&tag=&license=all&year=80s&filesize=all&developer=all&publisher=all) and [90s on dosgames.com](https://www.dosgames.com/listing.php?sortby=nameasc&sterms2=&cat=all&tag=&license=all&year=90s&filesize=all&developer=all&publisher=all))
- AFAIK not that many of the games released back then supported the mouse, this further limits the risk
- the `0x20` call didn't exist before the driver version 6, this limits the risk a bit more
- in case some game is incompatible with our mouse driver, we already have a workaround proven many times: `dos_mouse_driver = false`, load any compatible mouse driver by hand

_Descent II_ with a Voodoo patch looks to me like a much more important use case to me.

Deactivating our mouse driver in a runtime would be a nice quality of life improvement, but it's not that trivial to implement, it would require me to deal with DOSBox areas I have no experience with. Might happen one day, but so far this is low priority to me.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -1620,12 +1620,20 @@ static Bitu int33_handler()
...
+		// - MS driver 8.20 returns 0xffff if 'state.enabled == false'
+		// - 3rd party drivers I tested (A4Tech 8.04a, Genius 9.20,
+		//   Mouse Systems 8.00, DR-DOS driver 1.1) never return anything
 		break;
 	case 0x20: // MS MOUSE v6.0+ - enable mouse driver
 		state.enabled = true;
 		state.hidden  = state.oldhidden;
+		// At least Descent II with Voodoo patch expects this. Checked
+		// that MS driver alters AX this way starting from version 7.
+		reg_ax = 0xffff;
```
FeralChild64, line 1636: Note to self: to limit the risk of the change, we should do so in a protected mode only.

Grounded0:
If integrated mouse driver was able to be deactivated one could have all the 7.04+ compatibility and load Microsoft Mouse driver 6.26 for the few potential outliers since those aren't input lag dependent like Descent II indeed is. 🙂

FeralChild64:
@interloper98 

> with auto providing the most compatible behavior (real-mode apps get 6.x and protected-mode gets 8.x).. or something like that

Unfortunately, protected mode games temporarily switch to the real mode for the short duration of interacting with the mouse driver - so this can't be done this way easily. Looking at the graphics mode might not work either, because the game might initialize the mouse driver before initializing the graphics mode.

Any kind of 'auto' mode would need to be rather hacky.

Grounded0:
@LowLevelMahn 

Since you know how to write TSRs could you take a look into writing a TSR that fakes Microsoft Mouse driver 7.04 or newer behaviour to make this game run. Thanks.

rderooy:
I think setting this post 0.82.0 as default, no config option, and regardless of CPU mode is a good way forward.

That will give plenty of time for testing in dev before the next release.

You don't want to freeze dosbox-staging, like vanilla dosbox effectively has been, right? Because with such a risk-aversion strategy, it sure sounds like it.

Grounded0:
Work froze on DOSBox because the original devs had enough of it by 2012 and weren't willing to hand off control for new people to take over the project like we've done. That resulted in a 50+ dead DOSBox forks and massive amount of duplicated and wasted effort.

FeralChild64:
Well, Larry 2, 3, and 5 are not affected.

And Descent II is a frequent visitor of various "best games" lists, so definitely not a niche - https://gamesnostalgia.com/browse-games/DOS/1996.

johnnovak:
> Well, Larry 2, 3, and 5, are not affected.
> 
> And Descent II is a frequent visitor of various "best games" lists, so definitely not a niche - https://gamesnostalgia.com/browse-games/DOS/1996.

That's good news. Yeah, Descent II in itself is somewhat common, just not the 3dfx version  

Anyway, I'm not against this if it can be proven at least major 80s games with mouse support work fine and we don't need extra config settings.

johnnovak:
> You don't want to freeze dosbox-staging, like vanilla dosbox effectively has been, right? Because with such a risk-aversion strategy, it sure sounds like it.

You don't seem to get what I wrote. When a fix is obviously tge correct thing to do, by apl means do it. When you apply an experimental fix for a single game that could break other stuff, that's very questionable.

What's the end goal here? To play all DOS games with minimum fuss and as few regressions as possible 

This is not some alpha project where we can break shit all tge time after every second update.

johnnovak:
> So not everyone will be happy with such workaround for fast paced games.

Can you quantify the latency? I think it would 1-2 ms max, which is a non-issue.

interloper98:
Tested the PR on all of the following - all are comfirmed working with mouse control.  I'm trying to hit all of the major game dev companies:

Loom, Monkey Island 1, Star Wars X-Wing (Lucas Arts)
Martian Memorandum, Links 386 (Access)
Elder Scrolls II (Bethesda)
Warcraft II (Blizzard)
Heores of Might and Magic (New World Computing)
Falcon Gold (Spectrum Holobyte)
Red Alert (Westwood Studios)
Stellar 7, Aces of the Deep (Dynamix)
Duke Nukem 3D (3D Realms)
Heretic (Raven, id Software)
Sim City 2000 (Maxis)
Tyrian (Eclise)
Ultima 6 (Origin)
X-COM UFO (Microprose)
Buzz Aldrin's Space (Interplay)
Eric the Unready (Legend)
Space Quest 3 (Sierra)
Where in the World is Carmen Sandiego (Broderbund)

Others:
Allegro SDK/platform
DOS Navigator
Windows 3.11 (Microsoft)

FeralChild64:
@interloper98 Actually, we should focus on games from 1986 (Microsoft mouse driver 6.0 release year) to, like, 1991. Later titles should be compatible, or the customers would have complained they don’t work with then current mouse drivers.

@johnnovak Right, probably some 1-2 milliseconds. For me this is absolutely unnoticeable, but apparently some hardcore gamers can feel the difference.

interloper98:
> @interloper98 Actually, we should focus on games from 1986 (Microsoft mouse driver 6.0 release year) to, like, 1991.

Ok! Here's what I've tested; so far 100% success rate with this PR:

Maniac Mansion (1987)
King's Quest IV (1988)
Populous (1989)
SimCity (1989)
Space Quest III (1989)
Commander Keen (1990)
King's Quest V (1990)
Loom (1990)
The Secret of Monkey Island (1990)
Wing Commander (1990)
Civilization (1991)
Eye of the Beholder (1991)
Lemmings (1991)

johnnovak:
Thanks guys, this explanation and the test results put my concerns to rest.

I'm happy this to go in the 0.83.x cycle @FeralChild64.

Again, great work @FeralChild64, and don't take this as a negative criticism, just a note for the future, and I'm not singling you out. The PR contains this in the test section:

> Tried to reproduce the PR, played a few games which are using a mouse.

Yeah, so that's not overly confidence inspiring for such a sweeping change. If instead it contained your recent test results and the explanation about tge 6.0 driver affecting the 86-91 timeframe, we could have spared most of this discussion.

There's no arguing with facts 😎 

We should all carry out extensive testing _before_ taking a PR out of draft mode in order not to waste everyone's time. In my own PRs, half the time I need to go back to the drawing board after some extensive testing.

--- 

On a related note, would you like to start an input testing page in the dev section of our wiki, @FeralChild64? I already have a video and audio test cases page so an input testing page that covers the keyboard, mouse, and game controllers is begging to be created there 😏  As our input handling guru, you're the best person for the job 😎 

You could start small, just paste in the above list of games you tested fror a comprehensive general mouse regression testing suite, then we can expand the page with special cases and such further (e.g., listing games that just have glitchy and semi-broken DOS drivers, and thus we're emulating the glitches, e.g., Ultima Underworld).

These test pages formalise our testing method. Nobody could possibly keep all these details in their heads, and even a person inexperienced in a particular area should be able to execute most of these tests (e.g., during our RC periods to ensure no regressions). That would go a long way towards achieving consistent quality across releases vs just doing ad hoc manual testing.

Then future PRs would become even simpler: one could just state "carried our all mouse regression tests from the input test wiki page." I started doing this for video and audio related changes.

johnnovak:
> > @interloper98 Actually, we should focus on games from 1986 (Microsoft mouse driver 6.0 release year) to, like, 1991.
> 
> Ok! Here's what I've tested; so far 100% success rate with this PR:
> 
> Maniac Mansion (1987) King's Quest IV (1988) Populous (1989) SimCity (1989) Space Quest III (1989) Commander Keen (1990) King's Quest V (1990) Loom (1990) The Secret of Monkey Island (1990) Wing Commander (1990) Civilization (1991) Eye of the Beholder (1991) Lemmings (1991)

What a prestigious list of games! 😍 This covers pretty much my favourite DOS era and includes many of my favourite games ❤️

johnnovak (approved):

interloper98:
> What a prestigious list of games! 😍 This covers pretty much my favourite DOS era and includes many of my favourite games ❤️

That's for sure! I had a couple sequels to try (WC2, SQ4, etc) but at that point, I had to stop testing and **_start playing_**; they are just too damn good! :smile:

Grounded0:
There's apparently Voodoo patch released in 1998 for Descent II:

https://www.mobygames.com/game/694/descent-ii/patches/

https://www.patches-scrolls.de/patch/1212/7/22825

Could you guys check does patch release 1.06b work with our current mouse driver?

FeralChild64:
@johnnovak Additional testing and investigation happened because of suggestion that this 0xFFFF return value might actually be some kind of error code.

Yes, I think I can prepare such a Wiki page - this will have to wait a little, though, I won't be having much time in the next 2 weeks or so.

FeralChild64:
@Grounded0 This unofficial patch actually fixes the mouse problem we have here, too. Nice find! Changed the description to make it clear it fixes the problem with the _official_ patch.

johnnovak:
> @johnnovak Additional testing and investigation happened because of suggestion that this 0xFFFF return value might actually be some kind of error code.

Yeah, I guess I'm saying, we should all develop better intuition for it what change is problematic. Literally a few minutes after my initial excitement for the fix I was like, oh shit, this could break other games 😄

Grounded0:
> @Grounded0 This unofficial patch actually fixes the mouse problem we have here, too. Nice find! Changed the description to make it clear it fixes the problem with the official patch.

This means I'm opposed to this merge. I'll enter this into Game issues wiki and we can call this off?

Grounded0:
Updated Game issues wiki with the final fix:

https://github.com/dosbox-staging/dosbox-staging/wiki/Game-issues#descent-1-and-2

johnnovak:
> > @Grounded0 This unofficial patch actually fixes the mouse problem we have here, too. Nice find! Changed the description to make it clear it fixes the problem with the official patch.
> 
> This means I'm opposed to this merge. I'll enter this into Game issues wiki as a solution to close the issue and other people here can then debate is it good enough for that.

Seconded. Although this appears safe so far based on the recent test results, it's still questionable whether such a general change is a good idea just for the sake of a single game.

I'd say stash this PR away and if we find more games over time that require this specific behaviour of this specific version of tge MS mouse driver, we will reconsider.

Again: risk vs reward, not broken don't fux it, etc 😄

FeralChild64:
@johnnovak

> it's still questionable whether such a general change is a good idea just for the sake of a single game.

You have added a new sound card for a sake of a single game, but you don't want a mouse driver fix :) A config option to specify the mouse driver version we try to be compatible with is a perfectly fine solution to me.

> Again: risk vs reward, not broken don't fix it, etc

The official Voodoo release of a 1996 game (definitely not my favorite one, but iconic to many) does not work correctly with our emulated mouse driver, while it works with the original Microsoft mouse drivers. To me it's broken, and the fix is easy.

johnnovak:
> You have added a new sound card for a sake of a single game, but you don't want a mouse driver fix :) 

Apples vs oranges comparison. You can add _new_ sound card emulations like no tomorrow; that won't break compatibility with any existing game.

If someone tried to change the Sound Blaster DAC emulation, on the other hand, for the sake of a single game, I'd also advise extreme caution. That is a comparable change.

> A config option to specify the mouse driver version we try to be compatible with is a perfectly fine solution to me.

Yeah, but we try not to introduce new config settings for the sake of a single game. Also, how would anyone know to try setting a different mouse driver version via the config? That's a bit obscure. To me, that's not much better than finding a patch for the problematic game.

> The official Voodoo release of a 1996 game (definitely not my favorite one, but iconic to many) does not work correctly with our emulated mouse driver, while it works with the original Microsoft mouse drivers. To me it's broken, and the fix is easy.

But the non Voodoo version is fine, and there is a patch for the Voodoo version which I assume people would want anyway.

So, we're back to my original problem: adding another mouse option or a potentially breaking change (but safe looking so far, ok) for a single game.

Look, I don't want to die on this hill, but I don't see this change making our default mouse emulation _better_. It would certainly make it slightly _different_ though... for a single game. So, for me, given a patch exists, it's a very hard sell. Same deal with the config option -- just why? Would anything else benefit from this? Probably not, so I just don't find it acceptable. I don't like adding config settings "just in case".

Anyway, probably we'll need to vote or flip a coin at some point 😆

Grounded0:
My understanding is that the official v1.2 release is so buggy that 3dfx had to patch it themselves in 1998 so that the game could be actually be played through as there's crashes triggered by game events.

johnnovak:
> My understanding is that the official v1.2 release is so buggy that 3dfx had to patch it themselves in 1998 so that the game could be actually be played through as there's crashes triggered by game events.

I'm not familiar with the game, but if it can be proven that's the case, that makes improving out of the box compatibility with v1.2 totally pointless and settles the issue.

We don't care about improving the situation for broken game version, especially if that involves complication like this.

Grounded0:
Another example of a trash release is Terminator SkyNET that requires unofficial patch applied on top of the official patch to actually be able to finish the game. True Bethesda quality right there.

Grounded0:
Like I earlier posted here if someone has the energy to dig on this they should find out why the original commit by Harekiet in 2003 that behaves like MS Mouse driver 7.04+ was changed. There probably lies clues about it breaking mouse init sequence in some shitty Spanish football management game released in 1991.

There's a lot to dislike about the DOSBox development team but they had the basics nailed down pretty good by 2012 due to user feedback.

interloper98:
Despite Microsoft's extreme pursuit of backward compatibility, there are no doubt small changes as these versions evolved over the years.

| MS Mouse Version | Release year(s) | MS-DOS range coverage  |
| --- | --- | --- |
| MS Mouse 2.0x | 1983 & 1984 | MS-DOS 2.0 to 3.0 ('83 to '84) |
| MS Mouse 4.0x | 1985 & 1986 | MS-DOS 3.10 to 3.20 ('84 to '86) |
| MS Mouse 5.0x | 1986 | MS-DOS 3.20 to 3.21 ('86 to '87) |
| MS Mouse 6.0x | 1986 | MS-DOS 3.20 to 4.0 ('86 to '88) |
| MS Mouse 7.04 | 1990 | MS-DOS 4.0 to 5.0 ('88 to '91) |
| MS Mouse 8.20 | 1992 | MS-DOS 5.0 to 6.22 ('91 to '94) |

The original DOSBox project selected '5.0' to be the default and looks like Staging continues to use MS-DOS ver 5.0 as the default.  I suspect this is a compatibility sweet spot.

I would suggest that the `'ver = '` setting be not /just/ the superficial two digits but we actually put impactful MS-DOS-version-specific behaviour behind those numbers.

For example, if a game needs some old DOS 3.x behavior - be it memory layout, MCBs, `COMMAND.COM` parsing differences, and so on.. then the `ver` setting is what users will expect to "switch DOSBox to behave like DOS 3.3".

So in this case - I would think it's reasonable to do the same for mouse behaviour.

Grounded0:
DOS version mode flip is the best way to keep everyone happy. Please get it done. MS-DOS 5.0 gets old behaviour while MS-DOS 6.x gets the new behaviour.

johnnovak:
Great idea @interloper98! DOSBox's primary aim is to run games on a typical period accurate PC config appropriate for a given game with minimum fuss. In order to keep the number of config options down, we could bundle really low level minor differences into the DOS version setting, like you said.

I support this idea, with a small difference as @Grounded0 mentioned: make the current behaviour tied to the default DOS version (or less), and the PR behaviour to highet versions in order not to upset the out-of-the-box game compatibility.

- VER <= 5.0 -- current main behaviour
- VER > 5.0 -- PR behaviour

Either this, or drop the change altogether.

We have spent enough time debating this, so @FeralChild64 could you please get it done (or drop the PR) then we move on with our lives? 😆

Grounded0:
I actually did what I requested everyone else here to do and used this fancy tool I paid for when I was working in tech called Sublime Merge to dump history on `file:src/hardware/input/mouseif_dos_driver.cpp`:

Screenshot 2024-08-28 at 11 25 23

Looks like Feralchild64 rewrote the whole section and missed this because it was not documented in papers that were used as reference.

Grounded0:
Rewrote the solution in Game Issues wiki to point people to the 3dfx patch because Descent II v1.2 is inherently unstable to the point of being unplayable.

johnnovak:
I don't think this should been merged. The DOS version dependent behaviour should implemented as discussed.

Look at the current SVN code; it does not set AX to `ffffh`:
https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/src/ints/mouse.cpp#l947

@Grounded0 Can you please refrain from merging PRs in the future? It's best if merging is left to the author. Thank you.

Grounded0:
Will not do this again. Someone should do this in a follow-up PR. At least there's good documentation included now whats happening in that section and it can be fixed with this:

- #https://github.com/dosbox-staging/dosbox-staging/pull/3896#issuecomment-2314367358

javispedro:
> Someone should track from the SVN or our own git which goes back to the first commit when original commit by Harekiet in 2003 was changed to current form. There's likely been a game compatibility issue that has caused that change so DOS version mouse behaviour flip will be necessary to support every game.

I do not understand what you mean -- there was no behavior change in 2003.  I can't see any Dosbox version that ever returned -1 for function 0x20.

FWIW, my experience is more with applications, rather than games, but I am yet to see _any_ application ever calling functions 0x1F/0x20. The only usecase I can see for these functions is an application that wants to configure the mouse in a special way that may confuse the mouse driver up to the point of a crash -- but again I never saw this. This is why everyone gets away with almost nop implementations for these funcs, like vbados or dosbox itself...

This game's use of function 0x20 is obviously a bug, that caused it to be incompatible with everything except one version of one mouse driver, and which was fixed by the next patch of the game because they'd likely got one too many compatibility complains. Likely no other game ever calls function 0x20 much less notice its return value...

FeralChild64:
This definitely shouldn't have been merged, I'll prepare a revert PR in a moment. Yes, I can implement different mouse driver behavior depending on the simulated DOS version as @interloper98 suggested (with DOS 5 using the current behavior), that's not a problem.

I think I'll also extend the `MOUSECTL` tool to allow changing the simulated driver version, and definitely to print it out - we'll see. This, unfortunately, won't happen in the next 2 weeks - I'm a little busy now, besides I'm testing the RC, and slowly cooking another PR to fix some locale issues. And I need to think trough the user interface side :)

johnnovak:
@javispedro Thank you so much for chiming in with your valuable experience on the matter ❤️ 

This settles the issue and the solution is crystal clear: we don't want this change.

@interloper98 Let's keep your nice `VER` idea in mind for the next time something similar comes up.

@FeralChild64 Could you please revert this in a follow-up PR? I can't as I'm on mobile for the next three weeks.

FeralChild64:
@javispedro 

```This game's use of function 0x20 is obviously a bug, that caused it to be incompatible with everything except one version of one mouse driver, and which was fixed by the next patch of the game because they'd likely got one too many compatibility complains```

Could you be more specific? Which patch and which driver version do you have in mind? I have tried the patch https://archive.org/details/D2VOODOO (I am not aware about any further official patches - I bet there are countless unofficial ones nowadays, since the game source code was released in 1998), and the following Microsoft mouse drivers:

- 9.01 - works
- 9.00 - works
- 8.20a - works
- 7.04 - works (driver file is dated May 1990)
- 6.24 - does not work
- 6.10 - does not work
- 6.02 - does not work

I'm sorry if I'm start getting ill-tempered, but this gets really frustrating. Just a moment ago there was an agreement how the fix should look like, I already told the issue reporter that I intend to do a fix, and now I suddenly again hear "NO" based on claims that contradict my test results :/

interloper98:
@johnnovak , @FeralChild64 :

The 6.x driver was from the mid 80s. 

Anyone running MS DOS 6.x in the 90s would have definitely been using MS Mouse 7.04 at the oldest and more likely 8.20. 

They wouldn't be using drivers from the last decade. 

With Descent II as our proof, this is a perfect example where, if a user sets `ver = 6.22` they _**should**_ get this This one-line adjustment to give them the MS Mouse 7.04/8.2/9/10.x return code behavior.

johnnovak:
@FeralChild64 @interloper98 Personally, I'd drop the while thing, but you seem to super invested in this, so do the `VER` thing then if you think this is important. If nothing uses this, then it doesn't matter, but ok, do it...

I'll stop posting because this is already the most disputed single line change in the history of Staging 😆 Just' don't change the out-of-box behaviour please if you can avoid it.

johnnovak:
Last comment 😆 

> Anyone running MS DOS 6.x in the 90s would have definitely been using MS Mouse 7.04 at the oldest and more likely 8.20.

Anyone? Definitely? 😄 I don't think this statement is true.

Only if they had brought a brand new Microsoft mouse. How about non-MS mice? There is Logitech, IBM, Genius, etc. I never had a MS mouse in my life.

Look how many non-MS mice exist with DOS drivers, and I'm sure this driver collection is incomplete:

https://www.vogonsdrivers.com/index.php?catid=29&menustate=26,0

A _portion_ of PC users were on this MS mouse version in the 90s, that's correct. _Many_ used other brands, or perhaphs an old MS mouse with old drivers, or they used alternste drivers with their whatever mouse to maximise base memory (very common).

javispedro:
> Which patch and which driver version do you have in mind? I have tried the patch https://archive.org/details/D2VOODOO (I am not aware about any further official patches - I bet there are countless unofficial ones nowadays, since the game source code was released in 1998)

Fair enough, I guess there is a distinction between official and unofficial patches. Anyway, I suppose it was fixed in the unofficial patches for a reason. Many people would have used non-MS mouse drivers at the very least. 

My biggest point was that no program/game will care much about these functions. So I would either do the change unconditionally or document the game as requiring the unofficial patch. In any case, my guess about games is as good as yours.

johnnovak:
> So I would either do the change unconditionally or document the game as requiring the unofficial patch. In any case, my guess about games is as good as yours.

We've done the latter, and I consider this issue closed; we've wasted enough cycles on this.

PR #3908, opened 2024-08-30
Fix ZMBV CompareBlock function returning a negative value for 24/32bpp types

weirddan455:
# Description

Regression from 6892f70c8502d9ef790059756a7b33d2faa7f880
Above commit changed a variable from int to auto
This resulted in a logical right shift instead of the expected arithmatic right shift

This bug caused the motion vector to always be (0, 0) resulting in more inefficient compression

Found this bug while I was writing a ZMBV decoder for a video player I'm working on.

The logic is kind of hard to follow but form what I can tell `CompareBlock()` is supposed to return a positive value indicating how closely the two blocks match with 0 being (mostly?) identical and a large number meaning very different.

The bug made this function always return a negative number when in 24 or 32 bit per pixel mode.  It's a templated function so it happened when `P` is `uint32_t`.  This made the logic in `AddXorFrame()` think it always found a good match on the first try so it didn't try other motion vector possibilities.

This doesn't result in any visual glitches in the output video file so it went un-noticed for a long time.  Only observable affect is a slightly larger file size.

# Release notes

Fixed a bug with video capture that resulted in slightly larger file sizes.

# Manual testing

Recorded about 9 seconds of Crystal Caves from level 1 start to the hole in the ground.  Stubbed out audio capture as that was the bulk of the file size otherwise as audio is uncompressed.

Result was 1.49MB on this branch and 1.62MB on main

For a more scientific test, I would need to rip out the encoder and compress the exact same video but this was as close as I could get it without doing a bunch of extra work.

# Checklist

I have:

### DIFF
--- a/src/libs/zmbv/zmbv.cpp
+++ b/src/libs/zmbv/zmbv.cpp
@@ -181,19 +181,18 @@ int VideoCodec::PossibleBlock(const int vx, const int vy, const FrameBlock & blo
 template <class P>
 int VideoCodec::CompareBlock(const int vx, const int vy, const FrameBlock & block)
 {
-	int ret = 0;
+	int diff_count = 0;
 	P *pold = reinterpret_cast<P *>(oldframe) + block.start + (vy * pitch) + vx;
 	P *pnew = reinterpret_cast<P *>(newframe) + block.start;
-	;
+
 	for (auto y = 0; y < block.dy; y++) {
 		for (auto x = 0; x < block.dx; x++) {
-			const auto test = 0 - ((pold[x] - pnew[x]) & 0x00ffffff);
-			ret -= check_cast<int>(test >> 31);
+			diff_count += ((pold[x] ^ pnew[x]) & 0x00ffffff) != 0;
 		}
 		pold += pitch;
 		pnew += pitch;
 	}
-	return ret;
+	return diff_count;
 }
 
 template <class P>


src/libs/zmbv/zmbv.cpp
```diff
@@ -184,11 +184,11 @@ int VideoCodec::CompareBlock(const int vx, const int vy, const FrameBlock & bloc
...
 	P *pnew = reinterpret_cast<P *>(newframe) + block.start;
-	;
+
 	for (auto y = 0; y < block.dy; y++) {
 		for (auto x = 0; x < block.dx; x++) {
-			const auto test = 0 - ((pold[x] - pnew[x]) & 0x00ffffff);
-			ret -= check_cast<int>(test >> 31);
+			const int test = 0 - ((pold[x] - pnew[x]) & 0x00ffffff);
+			ret -= (test >> 31);
 		}
```
interloper98, line 192: That's pretty subtle! Great find, @weirddan455.

Maybe the `ret = ` could be renamed to describe the number of pixels found to be different (if that's what we're counting?)

Also, can the "zero minus" and the `ret -=` be flipped over to positives, so it reads more naturally:

```c++
    int diff_count = 0;
    ...    
    for ( ...
        for ( ...
            diff_count += ((pold[x] ^ pnew[x]) & 0x00ffffff) != 0;
        }
    }

    return diff_count;
```

(if not, maybe a comment is needed to describe why we need to subtract negatives)

interloper98:
> Found this bug while I was writing a ZMBV decoder for a video player I'm working on.

:rocket: ; right on! ZMBV has impressed me with how space-efficient it can be for game content; lots of great ideas went into it.

src/libs/zmbv/zmbv.cpp
weirddan455: Oh, yeah I guess it is just masking out the alpha channel (which is un-unsed) and then counting the number of pixels that differ.

I was wondering what that `test >> 31` meant but it's just checking the sign bit and incrementing the return value in the most obtuse way possible :sweat_smile: 

I'll stick your code in there and run a test to see if the values are identical.  If they are, yours is a lot more readable so I'll go with that.

interloper98: > it's just checking the sign bit and incrementing the return value in the most obtuse way possible

:100: I guess some of these elite coding patterns were for performance reasons (or something!) :hurtrealbad:

weirddan455: Ok, yours wins.  Added a log statement that fired if the values differed from the existing code and it was the same in both EGA (16bpp) and VGA (32bpp).

The only thing I can figure is the original author was trying to avoid a branch that the `!= 0` might cause but I took a quick look at the assembly out of curiosity and couldn't tell much difference.  The compiler was pretty smart and inlined and auto-vectorized a lot of this code with SIMD.

Changed to match your suggestion for readability :smile:

interloper98: > The compiler was pretty smart and inlined and auto-vectorized a lot of this code with SIMD.

Nice!

interloper98 (approved):
Looking forward to even smaller .avi's after this!

src/libs/zmbv/zmbv.cpp
johnnovak: > > it's just checking the sign bit and incrementing the return value in the most obtuse way possible
> 
> 💯 . I guess some of these elite coding patterns were for performance reasons (or something!) :hurtrealbad:

Standard 90's game /demo coder fare -- nothing unusual to see, moving on 😎 

Plus yeah, optimising C compilers were in their infancy 20-30 years ago compared to what we have now. One way to make C code from the 80/90s fly with almost zero effort is to simply recompile it with clang/gcc instead or Watcom/Turbo C/etc. (if you are able to find your way around the DOS era nemory model bullshit in the modern compiler).

weirddan455:
@johnnovak are you good with merging this?

johnnovak:
> @johnnovak are you good with merging this?

Yeah, 0.82.0 final material.

PR #3926, opened 2024-09-12
Provide application name to SDL

FeralChild64:
# Description

When testing on Linux with:
- SDL 2.30.7 using PipeWire audio driver
- KDE Plasma 6.1 with KDE Frameworks 6.5

I have noticed the volume applet displays DOSBox as a mysterious _SDL Application_ (you might need to execute DOSBox using `SDL_AUDIODRIVER=pipewire dosbox` command to see the issue - depending on your OS configuration):

Screenshot-Before

This PR sets the name to _DOSBox Staging_:

Screenshot-After

Unfortunately, changing the application name later on does not change the displayed name (only the first SDL hint is used) - so that adapting the name based on `window_titlebar` config option is not feasible, at least for now.

# Release notes

On _Linux_ systems using _PipeWire_ sound daemon, tools like `pw-top`, `pavucontrol`, or KDE volume applet now show the proper application name, _DOSBox Staging_, instead of the generic _SDL Application_.

# Manual testing

Execute DOSBox Staging with `SDL_AUDIODRIVER=pipewire dosbox` command, check the application name displayed in the KDE volume applet.

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4941,6 +4941,17 @@ int sdl_main(int argc, char* argv[])
 		        SDL_GetCurrentVideoDriver(),
 		        SDL_GetCurrentAudioDriver());
 
+#if defined SDL_HINT_APP_NAME
+		// For KDE 6 volume applet and PipeWire audio driver; further
+		// SetHint calls have no effect in the GUI, only the first
+		// advertised name is used
+		SDL_SetHint(SDL_HINT_APP_NAME, DOSBOX_NAME);
+#endif
+#if defined SDL_HINT_AUDIO_DEVICE_STREAM_NAME
+		// Useful for 'pw-top' and possibly other PipeWire CLI tools
+		SDL_SetHint(SDL_HINT_AUDIO_DEVICE_STREAM_NAME, DOSBOX_NAME);
+#endif
+
 		for (auto line : arguments->set) {
 			trim(line);
 


interloper98 (approved):
Very nice! Visible from the CLI tools, too (confirmed on raspberry pi OS):

**`pw-cli ls Node`**

```
	id 48, type PipeWire:Interface:Node/3
 		object.serial = "40370"
 		factory.id = "7"
 		client.id = "74"
 		application.name = "DOSBox Staging"
 		node.description = "Audio Stream"
 		node.name = "Audio Stream"
 		media.class = "Stream/Output/Audio"
 		media.type = "Audio"
 		media.category = "Playback"
 		media.role = "Game"
```

johnnovak (approved):
Looks good @FeralChild64

FeralChild64:
@johnnovak The risk seems to be very small here - do you think it would be feasible to be included in 0.82.0?

interloper98:
> @interloper98 There also exists `SDL_HINT_AUDIO_DEVICE_STREAM_NAME`, but I'm not sure if any GUI displays the value.

The stream name looks even more useful for CLI tools.
It shows up in `pw-top` (without stream name, there's no way to know what application owns the stream), and it also appears as the `node.name` and `node.description` in `pw-cli`.

Will post comparison screenshots shortly.

weirddan455:
I think there's something we can do with the .desktop file to make this even better and show the icon.  I'm on KDE 5.27 with PulseAudio and this is what I see on `main`.

pulse

However, the weird thing is when I install the Debian package for plain DOSBox, I see this when I launch Staging.  It must be pulling from DOSBox's .desktop file.  I can't quite figure out what the difference is from ours that makes it work correctly.  The .desktop file is also used for the Wayland driver to set the icon and that one correctly shows the Staging icon, even with the original DOSBox package installed.

pulse-db

FeralChild64:
@interloper98 Added `SDL_HINT_AUDIO_DEVICE_STREAM_NAME` support. I wasn't aware of the `pw-top` utility, thank you!

interloper98:
> @interloper98 Added `SDL_HINT_AUDIO_DEVICE_STREAM_NAME` support. I wasn't aware of the `pw-top` utility, thank you!

:beers:

Yes - super handy to see all your audio streams, the format they're using, and especially the latency!

FeralChild64:
> The stream name is where some terse details could be included (similar to our GUI title bar), like DOS executable, mixer rate, prebuffer, and block size.

@interloper98 I think we can consider this when 0.82.0 is out. It seems the GUI tool you have shown and the `pw-top` use different 'strategies' for application/stream display...

weirddan455:
Figured it out, kind of.  Renaming `org.dosbox-staging.dosbox-staging.desktop` to `dosbox.desktop` fixes it.  This is without setting the hint.  I'm not sure why it doesn't like the org naming schema when the KDE apps themselves also use it.  There may be a string somewhere (maybe another SDL hint?) we need to set to `org.dosbox-staging.dosbox-staging`.

pulse-staging

interloper98:
I especially like that we're able to give this info to SDL, without us having to delve into OS-specific APIs. Maybe these details are now being fed to macos and windows audio handlers, too.

FeralChild64:
@interloper98 I did one more test with `pavucontrol` (the tool you have shown), `pw-top`, and `pw-cli ls Node` - neither of these can handle application name / audio stream name change; this must be a limitation of the PipeWire server. Which means, the names will have to be static for now - no executable name, no custom title, nothing.

FeralChild64:
@weirddan455 https://wiki.libsdl.org/SDL3/SDL_HINT_APP_ID - unfortunately, only available since SDL 3.0.

weirddan455 (approved):
I looked into it more and this is the best we can do until SDL 3.  On the PulseAudio backend, SDL sets the name to `NULL` if the hint is not set.  This causes PulseAudio itself to try and auto-detect the name and icon, which seems to look for a .desktop file matching the binary name.

However, on the PipeWire backend, SDL falls back to the name `"SDL Application"` which matches what you were seeing on your system @FeralChild64 

Best to get the name correct on all platforms and not worry about the icon for now.  In SDL 3, there is a hint [SDL_HINT_AUDIO_DEVICE_APP_ICON_NAME](https://wiki.libsdl.org/SDL3/SDL_HINT_AUDIO_DEVICE_APP_ICON_NAME) that does what the name describes so that's what we should do in the future.

As a side-note, I tested Staging using the SDL3 -> SDL2 compatibility layer and it works :smile:

johnnovak:
> Maybe these details are now being fed to macos and windows audio handlers, too.

The program name"DOSBox Staging" is appearing correctly in the taskbar, Windows mixer, etc. because we've been providing the metadata on non-Linux platforms for ages (.rc file on Windows, .plist file on macOS). I think only Linux needs the extra help.

johnnovak:
> @johnnovak The risk seems to be very small here - do you think it would be feasible to be included in 0.82.0?

Yeah let's include it.

PR #3928, opened 2024-09-13
Fix data race in the mapper when using `autotype`

interloper98:
# Description

When testing older TSAN builds, I had trouble automating some intro steps with `AUTOTYPE` because TSAN threw data race warnings from the mapper and PIC system.  For example:

```
WARNING: ThreadSanitizer: data race (pid=1810861)
  Write of size 8 at 0x00000319d118 by main thread:
    #0 PIC_RunQueue() ../../src/hardware/pic.cpp:560
    #1 Normal_Loop ../../src/dosbox.cpp:168
    #2 DOSBOX_RunMachine() ../../src/dosbox.cpp:450
    #3 CALLBACK_RunRealInt(unsigned char) ../../src/cpu/callback.cpp:117
    #4 device_CON::Read(unsigned char*, unsigned short*) ../../src/dos/dev_con.h:89
    ...

  Previous write of size 8 at 0x00000319d118 by thread T18:
    #0 AddEntry ../../src/hardware/pic.cpp:442
    #1 PIC_AddEvent(void (*)(unsigned int), double, unsigned int) ../../src/hardware/pic.cpp:484
    #2 restart_delay_timer ../../src/hardware/input/intel8042.cpp:452
    #3 maybe_transfer_buffer ../../src/hardware/input/intel8042.cpp:480
    ...
```

This PR fixes those warnings by refactoring the 'Typist' class in the mapper.

## Related issues

Here's a full log of all TSAN issues hit when using `AUTOTYPE`:

[autotype-tsan-multiple.txt.gz](https://github.com/user-attachments/files/16989177/autotype-tsan-multiple.txt.gz)

# Manual testing

Tested normal builds, TSAN builds, with normal and missing/bad AUTOTYPE buttons. It's working as expected with the same wait and pause timings as before.

# Checklist

I have:

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -24,18 +24,15 @@
 #include <algorithm>
 #include <cassert>
 #include <cctype>
-#include <chrono>
 #include <cstdarg>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <atomic>
 #include <list>
-#include <thread>
+#include <queue>
 #include <vector>
 
 #include <SDL.h>
-#include <SDL_thread.h>
 
 #include "control.h"
 #include "joystick.h"
@@ -107,6 +104,7 @@ static std::vector<std::unique_ptr<CButton>> buttons;
 static std::vector<CBindGroup *> bindgroups;
 static std::vector<CHandlerEvent *> handlergroup;
 static std::list<CBind *> all_binds;
+static std::queue<std::string> auto_type_queue = {};
 
 typedef std::list<CBind *> CBindList;
 typedef std::list<CBind *>::iterator CBindList_it;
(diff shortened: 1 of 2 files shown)


weirddan455 (approved):
I was not aware that this autotype feature existed.  It looks like it basically just does delayed input.

This is a much saner approach than threading.  It's the reason why we have PIC events.  Nice job cleaning that mess up.

I don't have any comments on this.  It all looks fine to me at a glance.  We should let @johnnovak decide if this goes in 0.82.0 before merge.  It looks like a niche self-contained feature to me so it's probably fine but could go either way I guess.

johnnovak:
> I was not aware that this autotype feature existed. It looks like it basically just does delayed input.
> 
> This is a much saner approach than threading. It's the reason why we have PIC events. Nice job cleaning that mess up.
> 
> I don't have any comments on this. It all looks fine to me at a glance. We should let @johnnovak decide if this goes in 0.82.0 before merge. It looks like a niche self-contained feature to me so it's probably fine but could go either way I guess.

Yeah, it can go into 0.82, agreed. It's a nicer approach than threads for sure.

FeralChild64 (approved):
Looks OK. I use AUTOTYPE in several of my game launching scripts (like Civilization or Eye of the Beholder), and they continue working properly.

interloper98:
Thanks for the reviews. OK, might as well merge :-)

interloper98:
> I was not aware that this autotype feature existed. It looks like it basically just does delayed input.

Yup! I recently learned about it through some recentl eXoDOS tickets: https://github.com/exoscoriae/eXoDOS/issues/5119 and https://github.com/exoscoriae/eXoDOS/issues/5045.

PR #3929, opened 2024-09-13
Fix CI to create Debian package on `main` push

interloper98:
# Description

The Debian packager workflow runs on branches at the time they're merged to `main`, however we originally wanted it to run on `main` (_after_ branches are merged).  This fixes that criteria.

(If we want to include it for 0.82, then we can attach the CI .deb to the release binaries, too).

# Manual testing

None, because the adjustment is only to the CI workflow.

# Checklist

I have:

### DIFF
--- a/.github/workflows/debian-package.yml
+++ b/.github/workflows/debian-package.yml
@@ -1,9 +1,9 @@
 name: Debian package build
 
 on:
-  pull_request:
-    branches: [ main ]
-    types: [ closed ]
+  push:
+    branches:
+      - main
     paths-ignore:
       - '.clang-format'
       - '.mdl-styles'


johnnovak:
> If we want to include it for 0.82, then we can attach the CI .deb to the release binaries, too

Not sure we really want that? Isn't life simpler if only officially release the tar.gz for Linux? Less things to test, and the tar.gz is universal, while the .deb is kinda distro-specific, isn't it?

But I'll you Linux guys it decide, just I'm leaning towards simplicity and less maintenance overhead.

johnnovak (approved):
Looking good.

interloper98:
> Not sure we really want that @interloper98? 

It's just the same full-featured CI build, but inside .deb packaging, a bit like our `setup.exe` for Windows and `.dmg` for macOS (when we could offer up plain `.zip`s for those, too.. well, I guess we do for Windows, which indeed is similar to the tarball).

Like the Setup.exe, there are some hand-holding benefits to the .deb, which are that the executable is put in their PATH (and they can installer newer .debs and it's remove/replace the prior files). It also installs the man pages, etc. Just a bit more integrated than a bare zip.

> Less things to test, and the tar.gz is universal, while the .deb is kinda distro-specific, isn't it?

I think Debian is the biggest distro including derivatives, so it's as close to "maintain stream Linux" as we can get. 

My take is that that gap between the `.deb` and tar.gz is pretty benign, from the team support side (I've tested both on x86 and arm when I created that script.. the binaries are built with the same flags, the same compiler, the same libc, etc) but certainly if the .deb causes us any grief we can toss it down to "unsupported" or even scrap it entirely.

What I really hope is that it puts latent pressure on Debian itself to start packaging DOSBox Staging, especially because our .deb includes **_all_** the dependencies (including build-in `libmt32emu` "How dare they!"), which certainly might cause some hand wringing and pearl clutching at Debian HQ :smile:.

PR #3930, opened 2024-09-13
Follow-up fix for Debian package on `main` push

interloper98:
# Description

Preface -- sorry for the churn guys!

After the prior merge, the job still didn't run:

debian-pkg-still-not-running

Turns out there was one remaining condition blocking it from running, which I've now snipped out. I don't see any other blocks, so hopefully this is it!

# Manual testing

(Not tested.. unfortunately main-only-CI tests only trigger after merging to `main`).

# Checklist

I have:

### DIFF
--- a/.github/workflows/debian-package.yml
+++ b/.github/workflows/debian-package.yml
@@ -18,7 +18,6 @@ concurrency:
 
 jobs:
   build:
-    if: ${{ github.event.pull_request.merged }}
     name: Release package ${{ matrix.arch }}
     runs-on: ubuntu-22.04
     strategy:


interloper98:
Wanted to mention regarding the `.deb`, I'm 100% OK with whatever approach the team wants, come release time.
 - Not gotta offer it! (NP, users can find it CI jobs)
 - We'll offer it conditionally (_"Supported by the Linux-using contingent"_)
 - Put it on my shoulders :laughing:  (_"Supported only by @interloper98"_)

johnnovak:
> Wanted to mention regarding the `.deb`, I'm 100% OK with whatever approach the team wants, come release time.
> 
>     * Don't offer it (OK; knowledgeable users can dig it out from CI)
> 
>     * Offer it conditionally: _"Supported on a best effort basis if bugs are reproduceable."_
> 
>     * Put it on my shoulders 😆  (_"Supported only by @interloper98"_)

I don't know how important or useful providing a .deb package is for Linux users given we already provide the .tar.gz. I use stuff like Neovim on WSL; I just download their .tar.gz, extract it to `~/bin/`, and Bob's your uncle 😎 

They also provide AppImages but no Debian packages:

https://github.com/neovim/neovim/releases

I don't remember if our .deb artifact is essentially the statically linked artifact, just in .deb format, or it depends on system libs of a specific Debian version.

If the former, I see it as redundant.

If the latter, yeah, maybe useful but only for Debian (and maybe Ubuntu) users? (Probably this is the case because AFAIK Debian folks get medium strength full-body convulsions if someone mentions the term "static linking" 🤣)

But if we officially provide it, it _must_ be tested _and_ supported. If you want to take that on @interloper98 and you're planning to stick around, I see no harm. I see _some_ risk of people asking why their favourite MyLittlePonyLinux distro is not supported, why just Debian, so this could open the floodgates, but those questions can always be swiftly dealt with (we can always just say: "because" 😆 )

johnnovak (approved):

interloper98:
> it must be tested and supported. If you want to take that on

I'll test the .deb before the release and before we post it to the available packages for download.

> I don't remember if our .deb artifact is essentially the statically linked artifact, just in .deb format

I'm making the .deb as compliant as possible and that means letting Debian provide as many deps as possible (installing the .deb automatically does that for the user).

Every non-compliance is a hard-checked violation that's listed, and one of those is using dependencies already available in Debian.

The .deb process inspects the binary and detects when you've statically linked a lib that Debian has in its repo (yes, I tried :smile: and it flagged violations for libz, png, fluidsynth, and so on). Violations aren't fatal.. we could plow ahead with them, but I wanted to make the .deb as acceptable as possible.

Users can inspect this metadata before installing the package.

One of the biggest security risks on Linux systems is running old versions of unpatched and known-exploited software, so a conservative user just has to make sure they keep their system up-to-date and never install packages with these static linked violations (which can become stale). So I see the benefit from that angle.

johnnovak:
> > it must be tested and supported. If you want to take that on
> 
> I'll test the .deb before the release and before we post it to the available packages for download.
> 
> > I don't remember if our .deb artifact is essentially the statically linked artifact, just in .deb format
> 
> I'm making the .deb as compliant as possible and that means letting Debian provide as many deps as possible (installing the .deb automatically pulls them in).
> 
> Every non-compliance is a hard-checked violation that's listed by Lintian, and one of those violations is statically linking dependencies that already available in Debian.
> 
> The .deb process inspects the binary and detects when you've statically linked a lib that Debian has in its repo (yes, I tried 😄 and it flagged violations for libz, png, fluidsynth, and so on). Violations aren't fatal.. we could plow ahead with them, but I wanted to make the .deb as acceptable as possible.
> 
> Users can inspect this metadata before installing the package.
> 
> One of the biggest security risks on Linux systems is running old versions of unpatched and known-exploited software, so a conservative user just has to make sure they keep their system up-to-date and never install packages with these static linked violations (which can become stale). So I see the benefit from that angle.

Sounds good man 👍🏻

PR #3935, opened 2024-09-17
Use an optimal number of threads in Voodoo 

interloper98:
# Description

This PR replaces the `"voodoo_multithreading"` conf setting with `"voodoo_threads"`, which will use up to seven threads based on available CPU cores. This is enabled by default, however the user can either disable it or set a custom number of threads.

The sweet spot of seven threads was based on @PoloniumRain's benchmarks and recommendation:  https://github.com/schellingb/dosbox-pure/issues/300#issuecomment-2144050898.

## Related issues

None.

# Release notes

The `"voodoo_multithreading"` conf setting has been replaced with `"voodoo_threads"`, which uses up to seven threads based on available CPU cores. This is enabled by default, however the user can either disable it or set a custom number of threads.

Thanks to @PoloniumRain's from the DOSBox Pure project for [his benchmark results](https://github.com/schellingb/dosbox-pure/issues/300#issuecomment-2144050898).   

# Manual testing

I'm seeing the same performance on the Pi (4 core), which is expected.
On my Ryzen (8 cores) in win11, I'm seeing modest speed up in Descent 2 and Quake 2 for DOS.

Testing valid and invalid values in the new `"voodoo_threads"` conf setting.

# Checklist

I have:

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -834,8 +834,19 @@ void DOSBOX_Init()
 	        "   4: 2 MB for the FBI and one TMU with 2 MB (default).\n"
 	        "  12: 4 MB for the FBI and two TMUs, each with 4 MB.");
 
-	pbool = secprop->Add_bool("voodoo_multithreading", only_at_start, true);
-	pbool->Set_help("Use threads to improve 3dfx Voodoo performance (enabled by default).");
+	// Deprecate the boolean Voodoo multithreading setting
+	pbool = secprop->Add_bool("voodoo_multithreading", deprecated, false);
+	pbool->Set_help("Use 'voodoo_threads = auto / <value>'.");
+
+	pstring = secprop->Add_string("voodoo_threads", only_at_start, "auto");
+	pstring->Set_help(
+	        "Use threads to improved 3dfx Voodoo performance:\n"
+	        "  auto:     Use up to 7 threads based on available CPU cores (default).\n"
+	        "            This is in addition to the permanent main thread.\n"
+	        "  <value>:  Set a specific number of additional threads, for example: 12\n"
+	        "Note: Tests show that frame rates increase up to 7 threads after which\n"
+	        "      they level off or decrease, in general.\n"
+	        "      Threads are only created and active when 3dfx Voodoo is in use.");
 
 	pbool = secprop->Add_bool("voodoo_bilinear_filtering", only_at_start, false);
 	pbool->Set_help(
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -271,8 +271,6 @@ enum VoodooModel
 	VOODOO_2,
 };
 
-enum { TRIANGLE_THREADS = 3, TRIANGLE_WORKERS = TRIANGLE_THREADS + 1 };
-
 /* maximum number of TMUs */
 #define MAX_TMU					2
 
@@ -700,14 +698,16 @@ using rgb_union = voodoo_reg;
 
 /* note that this structure is an even 64 bytes long */
 struct stats_block {
-	int32_t pixels_in;          // pixels in statistic
-	int32_t pixels_out;         // pixels out statistic
-	int32_t chroma_fail;        // chroma test fail statistic
-	int32_t zfunc_fail;         // z function test fail statistic
-	int32_t afunc_fail;         // alpha function test fail statistic
+	int32_t pixels_in   = 0;
+	int32_t pixels_out  = 0;
+	int32_t chroma_fail = 0;
+	int32_t zfunc_fail  = 0;
+	int32_t afunc_fail  = 0;
 	// int32_t clip_fail;       // clipping fail statistic
 	// int32_t stipple_count;   // stipple statistic
-	int32_t filler[64 / 4 - 5]; // pad this structure to 64 bytes
+
+	// pad this structure to 64 bytes
+	int32_t filler[64 / 4 - 5] = {};
 };
 static_assert(sizeof(stats_block) == 64);
 
@@ -907,19 +907,75 @@ struct draw_state {
 
 struct triangle_worker
 {
-	std::atomic_bool threads_active;
-	bool use_threads, disable_bilinear_filter;
-	uint16_t *drawbuf;
-	poly_vertex v1, v2, v3;
-	int32_t v1y, v3y, totalpix;
-	std::array<std::thread, TRIANGLE_THREADS> threads;
-	std::array<Semaphore, TRIANGLE_THREADS> sembegin;
-	Semaphore semdone;
-	int done_count;
+	triangle_worker(const int num_threads_)
+	        : num_threads(num_threads_),
+	          num_workers(num_threads + 1),
+	          threads(num_threads),
+	          sembegin(num_threads)
+	{
+		assert(num_threads > 0);
+		assert(num_workers > num_threads);
+		assert(!threads.empty());
+		assert(!sembegin.empty());
+	}
+
+	triangle_worker()                                  = delete;
+	triangle_worker(const triangle_worker&)            = delete;
+	triangle_worker& operator=(const triangle_worker&) = delete;
+
+	void NotifySemaphores()
+	{
+		std::for_each(sembegin.begin(), sembegin.end(), [](auto& sem) {
+			sem.notify();
+		});
+	}
+
+	void WaitForSemaphores()
+	{
+		std::for_each(sembegin.begin(), sembegin.end(), [&](auto&) {
+			semdone.wait();
+		});
+	}
+
+	const int num_threads = 0;
+	const int num_workers = 0;
+
+	bool disable_bilinear_filter = {};
+
+	std::atomic_bool threads_active = {};
+
+	uint16_t* drawbuf = {};
+
+	poly_vertex v1 = {};
+	poly_vertex v2 = {};
+	poly_vertex v3 = {};
+
+	int32_t v1y      = 0;
+	int32_t v3y      = 0;
+	int32_t totalpix = 0;
+
+	std::vector<std::thread> threads = {};
+	std::vector<Semaphore> sembegin  = {};
+
+	Semaphore semdone = {};
+
+	int done_count = 0;
 };
 
 struct voodoo_state
 {
+	voodoo_state(const int num_threads)
+	        : tworker(num_threads),
+	          thread_stats(tworker.num_workers)
+	{
+		assert(num_threads > 0);
+		assert(!thread_stats.empty());
+	}
+
+	voodoo_state()                               = delete;
+	voodoo_state(const voodoo_state&)            = delete;
+	voodoo_state& operator=(const voodoo_state&) = delete;
+
 	uint8_t chipmask = {}; /* mask for which chips are available */
 
 	voodoo_reg reg[0x400]    = {}; /* raw registers */
@@ -941,9 +997,6 @@ struct voodoo_state
 	                                                    rasterizers */
 #endif
 
-	stats_block thread_stats[TRIANGLE_WORKERS] = {}; /* per-thread
-	                                                    statistics */
-
 	bool send_config   = {};
 	bool clock_enabled = {};
 	bool output_on     = {};
@@ -957,8 +1010,9 @@ struct voodoo_state
 	const char *const *	regnames;				/* register names array */
 #endif
 
-	draw_state draw         = {};
-	triangle_worker tworker = {};
+	draw_state draw = {};
+	triangle_worker tworker;
+	std::vector<stats_block> thread_stats = {};
 };
 
 #ifdef C_ENABLE_VOODOO_OPENGL
@@ -3020,7 +3074,7 @@ iterated W    = 18.32 [48 bits]
 
 static voodoo_state* v = nullptr;
 static auto vtype = VOODOO_1;
-static auto voodoo_multithreading     = true;
+
 static auto voodoo_bilinear_filtering = false;
 
 #define LOG_VOODOO LOG_PCI
@@ -3695,7 +3749,8 @@ static void init_fbi(fbi_state* f, int fbmem)
 
 	f->sverts = 0;
 
-	memset(&f->lfb_stats, 0, sizeof(f->lfb_stats));
+	f->lfb_stats = {};
+
 	memset(&f->fogblend, 0, sizeof(f->fogblend));
 	memset(&f->fogdelta, 0, sizeof(f->fogdelta));
 }
@@ -4308,7 +4363,7 @@ static void update_statistics(voodoo_state *vs, bool accumulate)
 			accumulate_statistics(vs, &thread_stat);
 		}
 	}
-	memset(vs->thread_stats, 0, sizeof(vs->thread_stats));
+	std::fill(vs->thread_stats.begin(), vs->thread_stats.end(), stats_block());
 
 	/* accumulate/reset statistics from the LFB */
 	auto& fbi = vs->fbi;
@@ -4316,14 +4371,15 @@ static void update_statistics(voodoo_state *vs, bool accumulate)
 	if (accumulate) {
 		accumulate_statistics(vs, &fbi.lfb_stats);
 	}
-	memset(&fbi.lfb_stats, 0, sizeof(fbi.lfb_stats));
+	fbi.lfb_stats = {};
 }
 
 /***************************************************************************
     COMMAND HANDLERS
 ***************************************************************************/
 
-static void triangle_worker_work(triangle_worker& tworker, int32_t worktstart, int32_t worktend)
+static void triangle_worker_work(const triangle_worker& tworker,
+                                 const int32_t work_start, const int32_t work_end)
 {
 	/* determine the number of TMUs involved */
 	uint32_t tmus     = 0;
@@ -4359,8 +4415,16 @@ static void triangle_worker_work(triangle_worker& tworker, int32_t worktstart, i
 
 	stats_block my_stats = {};
 
-	const int32_t from = tworker.totalpix * worktstart / TRIANGLE_WORKERS;
-	const int32_t to   = tworker.totalpix * worktend / TRIANGLE_WORKERS;
+	// The number of workers represents the total work, while the start and
+	// end represent a fraction (up to 100%) of the total total.
+	assert(work_end > 0 && tworker.num_workers >= work_end);
+
+	// The following suppresses div-by-0 false positive reported in Clang
+	// analysis. This is confirmed fixed in Clang v18.
+	const auto num_workers = tworker.num_workers ? tworker.num_workers : 1;
+
+	const int32_t from = tworker.totalpix * work_start / num_workers;
+	const int32_t to   = tworker.totalpix * work_end / num_workers;
 
 	for (int32_t curscan = tworker.v1y, scanend = tworker.v3y, sumpix = 0, lastsum = 0;
 	     curscan != scanend && lastsum < to;
@@ -4403,7 +4467,7 @@ static void triangle_worker_work(triangle_worker& tworker, int32_t worktstart, i
 
 		raster_generic(v, tmus, texmode0, texmode1, tworker.drawbuf, curscan, &extent, my_stats);
 	}
-	sum_statistics(&v->thread_stats[worktstart], &my_stats);
+	sum_statistics(&v->thread_stats[work_start], &my_stats);
 }
 
 static int triangle_worker_thread_func(int32_t p)
@@ -4425,13 +4489,10 @@ static void triangle_worker_shutdown(triangle_worker& tworker)
 		return;
 	}
 	tworker.threads_active = false;
-	for (size_t i = 0; i != TRIANGLE_THREADS; i++) {
-		tworker.sembegin[i].notify();
-	}
 
-	for (size_t i = 0; i != TRIANGLE_THREADS; i++) {
-		tworker.semdone.wait();
-	}
+	tworker.NotifySemaphores();
+
+	tworker.WaitForSemaphores();
 
 	for (auto& thread : tworker.threads) {
 		if (thread.joinable()) {
@@ -4442,11 +4503,10 @@ static void triangle_worker_shutdown(triangle_worker& tworker)
 
 static void triangle_worker_run(triangle_worker& tworker)
 {
-	if (!tworker.use_threads)
-	{
+	if (!tworker.num_threads) {
 		// do not use threaded calculation
 		tworker.totalpix = 0xFFFFFFF;
-		triangle_worker_work(tworker, 0, TRIANGLE_WORKERS);
+		triangle_worker_work(tworker, 0, tworker.num_workers);
 		return;
 	}
 
@@ -4485,7 +4545,7 @@ static void triangle_worker_run(triangle_worker& tworker)
 	// Don't wake up threads for just a few pixels
 	if (tworker.totalpix <= 200)
 	{
-		triangle_worker_work(tworker, 0, TRIANGLE_WORKERS);
+		triangle_worker_work(tworker, 0, tworker.num_workers);
 		return;
 	}
 
@@ -4501,13 +4561,11 @@ static void triangle_worker_run(triangle_worker& tworker)
 			++worker_id;
 		}
 	}
-	for (auto& begin_semaphore : tworker.sembegin) {
-		begin_semaphore.notify();
-	}
-	triangle_worker_work(tworker, TRIANGLE_THREADS, TRIANGLE_WORKERS);
-	for (size_t i = 0; i != TRIANGLE_THREADS; i++) {
-		tworker.semdone.wait();
-	}
+	tworker.NotifySemaphores();
+
+	triangle_worker_work(tworker, tworker.num_threads, tworker.num_workers);
+
+	tworker.WaitForSemaphores();
 }
 
 /*-------------------------------------------------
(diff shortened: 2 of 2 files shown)


weirddan455:
Don't have time to fully review this now (I'll take a deeper look at the code later) but I did test with Descent 2 on my 8 core Ryzen 7 1700.  I think there's maybe a 5-10 FPS increase but it's hard to tell.  I was testing with the game's build in FPS counter (just type "frametime") and the framerate jumps around with swings from 40FPS up to over 100FPS.  It would be nice if there was a benchmark that showed minimum + average FPS so I could tell definitively.

Interesting the guy on that DOSBox Pure thread reported that it performance decreases with more than 7 threads on his 24 core Threadripper.  Usually render workloads can scale better than that but I guess if the per-thread workload is too small you just lose out from synchronization.

interloper98:
@weirddan455 , try cranking up Descent 2's cycle count to 180k and run it throttled. I think this moves the bottleneck from Descent's "core" to the Voodoo output side.

At 180k cycles, facing forward the rates vary quite a bit, but the upper end is definitely higher. With `main` it never breaks 300 fps, but I'm seeing low to mid 300's on the branch w/ 7 threads reported.
When I turn behind and look at the door frame, `main` is mid 300s but the branch is high 400s to low 500s:

`main`
image0001-rendered
image0002-rendered

branch:
image0003-rendered
image0004-rendered

Not that these frame rates make a difference; it looks just as smooth at 60 fps :sweat_smile: but at least it's showing an improvement with more cores.

I wasn't able to capture video. Capturing slows `main` and the branch down into the 40 FPS range.

weirddan455:
>  try cranking up Descent 2's cycle count to 180k and run it throttled.

I was testing with max but I tried again with those settings and it's about the same.  You're getting a lot higher FPS than me.  Even I stare at a wall, it maxes out at 120FPS on my system.  Not sure what my bottleneck is :man_shrugging:

interloper98:
Darn. I'm using a release build and no other tweaks. This system is running a Ryzen 7840u, it's quite a little rocket and runs circles around the pi! But the pi feels more fun and challenging.

FeralChild64:
So far I haven’t found any DOS-based 3Dfx (Glide) benchmark.

I’m going to try GLQuake on Windows 98 - it seems to be working even on the Voodoo 1, https://m.youtube.com/watch?v=G3aM2jmK29U

johnnovak:
> I'm running Ghost Spectre, which is a super stripped down Win11

Uh oh... I hope you don't use that user modded OS of questionable origin for anything sensitive. Like buying stuff online with your credit card, internet banking, etc. Better safe than sorry, it can contain any number of backdoors.

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7046,45 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+{
+	constexpr auto MinNumThreads = 1;
+	constexpr auto MaxNumThreads = 7;
+
+	static auto num_threads = std::clamp(get_cpu_core_count() - 1,
+	                                     MinNumThreads,
+	                                     MaxNumThreads);
+
+	return num_threads;
+}
```
johnnovak, line 7086: Any reason for the high levels of mistrust of these CPU core count query function? 🤣 

I expect `SDL_GetCPUCount` to be battle-tested and always work correctly. Besides, you clamp to 1 to 7 anyway, so this validation and fallback mechanism is pointless.

I strongly suggest to drop this function and use `SDL_GetCPUCount` directly wherever you need the core count.

johnnovak:
Nice one @interloper98 🎉 The CPU core detection could be a lot simpler, but otherwise looks good. Nice cleanup job too 👍🏻

src/hardware/voodoo.cpp
interloper98: I agree; now simplified and added a couple comments in the function.

interloper98:
> > I'm running Ghost Spectre, which is a super stripped down Win11
> 
> Uh oh... I hope you don't use that user modded OS of questionable origin for anything sensitive. Like buying stuff online with your credit card, internet banking, etc. Better safe than sorry; it can contain any number of backdoors.

Lol yes.. it's an untrusted system on the guest LAN :-) (Another option I've tried are the Tron scripts on reddit, which are decent).

src/hardware/voodoo.cpp
johnnovak: Yeah, I highly doubt they'd ever actually fail in practice 😉

Grounded0:
Nice performance improvements on high-end machines. Number of threads can be added later after this is merged mainly for power management reasons as this will eat through my battery real fast while not providing any tangible benefit since its drawing at over 70 FPS. Gaming laptops are super popular these days.

johnnovak:
> Nice performance improvements on high-end machines. Manual control for the number of threads can be added later after this is merged mainly for power management reasons as this will eat through my battery real fast while not providing any tangible benefit since its drawing at over 70 FPS. Gaming laptops are super popular these days.

Not if you don't play any Voodoo game.

Grounded0:
Some high core count CPUs may also benefit from manually restricting this number but we shall see how this performs once it hits `main`.

Grounded0:
> Not if you don't play any Voodoo game.

EF2000 happens to be one of my top 5 games on DOS so that's not a luxury I have.

FeralChild64:
I g e up. I have no idea how to install Voodoo drivers on Windows 98 under DOSBox Staging.

PoloniumRain:
I'd help you guys test this out if i wasn't away from home at the moment. But here's some ways how i did my DOSBox Pure benchmarks... Maybe it will be of some help...

[QDOS](http://dk.toastednet.org/QDOS/) and especially [Q2DOS](http://dk.toastednet.org/Q2DOS/) are excellent. They're 3dfx DOS backports of Quake 1 and 2.

To run the built-in benchmark for QDOS:
1. Press the tilde key `~` or backtick `` ` `` key (it depends of your keyboard region layout) to bring up the console.
2. Type `timedemo demo1` and press Enter to run the benchmark on the first map (or `demo2` or `demo3` for other maps).

For Q2DOS:
1. Bring up the console the same way same as above.
2. Type `timedemo 1` and press Enter, followed by `map demo1.dm2` and press Enter (or `demo2.dm2` for map 2).
3. Close the console to start the benchmark.
4. There's also an option to show FPS in Options > Extended Options.

Some 3dfx DOS games with built-in frame rate counters:

- Shadow Warrior:
Press T, type `swloc` and press Enter.
- Blood:
Press T, type `rate` and press Enter.
- Battle Arena Toshinden:
The pause menu has a frame rate counter (press Escape during gameplay), but game is capped to 60fps.
- And of course Descent II by just typing `frametime`.

I also used the DOSBox Pure built-in frame rate counter. I don't know if Staging has anything like this.

Some Windows games that work with the Voodoo 1 and have built-in benchmarks:

- Forsaken:
Go to the main menu > Load Game > Play Demo. 
- Expendable:
Go to the games folder and create a shortcut for `go.exe`.
Right click on the shortcut > Properties > and in the Target box, add `-timedemo 1` to the end. 
Then launch the game with  this shortcut. 
Results are saved to a text file in the games folder.
- Deathkarz:
After installation it will add a `Performance Test` shortcut to the Windows Start menu.
- Unreal Tournament:
Menu screen > Tools > enable `TimeDemo Statistics`.
Press the tilde key `~` or backtick `` ` `` key to bring up the console. 
Type `timedemo 1` and press Enter, then type `cityintro` and press Enter.
Then press Escape to exit the menu screen.
- Quake III:
Press the tilde key `~` or backtick `` ` `` key to bring up the console. 
Type `timedemo 1` and press Enter.
Main menu > Demos > Play.
After it's finished, bring up the console again to show the average FPS.

>I'm running Ghost Spectre

I actually ran all my benchmarks on this lol. BTW i've used it for well over a year and i've run full system scans with multiple different AV software (at least 6)  and monitored network traffic for a long time.  I haven't found anything dodgy or had my password/credit card info stolen (so far!)...

Grounded0:
This addresses @Kappa971 wish for more performance so this will close it. Everything else on that wish list will not be implemented since its Windows-specific and Glide wrappers are notoriously buggy.

weirddan455:
> Nice performance improvements on high-end machines. Manual control for the number of threads can be added later after this is merged mainly for power management reasons as this will eat through my battery real fast while not providing any tangible benefit since its drawing at over 70 FPS. Gaming on laptops is super popular these days.

This might be a worthwhile config option.  When I tested Decent 2, this PR more than doubled my CPU usage from about 12% to 28% while giving minimal performance gains.  Not a big deal on a desktop but on a battery powered laptop, that's a real concern.  Also, you could let users crank up to 16 threads if they want since it was reported that helped with Decent 2 at least.

I'm going to test on Windows here in a bit to see if performance is any better there.  I suspect I'm being bottle-necked somewhere in the emulation core rather than Voodoo as I was getting audio stutters even on 60k cycles if I left it unthrottled (although this didn't happen with the software renderer so maybe not...)  @interloper98 's CPU is probably faster than mine but I wouldn't think by a factor of 4x so I'm curious what's going on there.

@PoloniumRain Thanks for the detailed bench-marking steps.  I probably won't fuss with the Win98 stuff as that's out of scope for Staging but I'll give those 3dfx Quake ports a test.

interloper98:
Should we keep the multithread conf setting?

It can default to yes with the new auto-thread-count approach, but could also allow manual control over the number of threads, if there's enough user interest or value added.

Most of the work in this PR was refactoring out the hard-coded global thread count, so it's no much work to make it user-set now.

interloper98:
>  @interloper98 's CPU is probably faster than mine but I wouldn't think by a factor of 4x so I'm curious what's going on there.

I want to figure this out too. There's no way AMD's uplifted IPC by 4x across a couple generations; more like 25%, I would guess, especially given both our systems are capped out w/ 7 threads.

Whatever benefit is happening there I want others to get it too. I'll try @PoloniumRain's QDOS benchmarks and keep poking around.

johnnovak:
> Should we keep the multithread conf setting?
> 
> It can default to yes with the new auto-thread-count approach, but could also allow manual control over the number of threads, if there's enough user interest or value added.
> 
> Most of the work in this PR was refactoring out the hard-coded global thread count, so it's not much work to make it user-set now.

Yeah, if there are tangible benefits to allowing specifying the thread count, I'm not against it. I was against adding it by rote without measuring things first, but seems like it's worthwhile for power consumption optimisation purposes.

interloper98:
Will add the conf adjustment. Thanks for the feedback on that @Grounded0 and go-ahead @johnnovak.

weirddan455:
>  Should we keep the multithread conf setting?

IMO replace the old config options with a user-specified number of threads or `auto` to use the detected core count clamped to 7.  Maybe include a warning that performance decreases in some games with high thread count (greater than 7).

There is also a question of backwards compatibility.  @johnnovak did some work on the config system to introduce deprecated config options (big use case being the CPU cycles).  You could do the same thing here and make `voodoo_multithreading = true` translate to `voodoo_threads = auto` and `voodoo_multithreading = false` translate to either 1 or 0 threads (whichever is equivalent).

I'm not sure we really need the backwards compatibility for this setting as it seems to be strictly a performance optimization and shouldn't break anyone's games but thought I would mention it.

interloper98:
Thanks @weirddan455. `voodoo_threads = .. ` is better name given we're letting the user set a quantity. Will do!

weirddan455:
I tested in Windows and I'm getting 2x higher FPS in Windows than in Linux (both in main and in this branch).  This branch also seems to give around 20-40 FPS increase in Windows so nice job there :+1:

I'll need to profile it to figure out what's going on in Linux.  There's obviously some platform specific bottleneck going on.  I tried downloading the [latest Clang 18 binary](https://github.com/llvm/llvm-project/releases/tag/llvmorg-18.1.8) (built for Ubuntu 18.04 but it just worked after setting my CC and CXX environment variables).  No real difference in performance there sadly.

shermp:
Just a tip to _really_ stress test Descent II 3DFX performance. free the guide bot and then run into it to produce sparks. The framerate should plummet.

johnnovak:
> Thanks @weirddan455. `voodoo_threads = .. ` is better name given we're letting the user set a quantity. Will do!

Sounds good and thanks for mentioning backwards compatibility @weirddan455. I like your idea and it's easy enough to; it's a good exercise for @interloper98 to learn about the deprecation mechanism.

johnnovak:
> @interloper98, please let the user number be able to go high... so that users like @PoloniumRain can test on their Threadrippers :)
> 
> I see no harm in leaving the hypothetical max at 256 or even more (I think Win10 is limited to 1280).

That's silly. At ridiculous numbers the performance benefits will turn become negative due to synchronisation. Capping it at 16 is more than enough.

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7046,37 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+	if (auto count = std::thread::hardware_concurrency(); count > 0) {
+		return static_cast<int>(count);
+	}
+	LOG_WARNING("VOODOO: Failed getting the number of host processors");
+
+	// SDL uses similar API calls with some variations, however it will lie
+	// and return 1 even if it fails. So we only use it as a fallback and
+	// accept whatever it gives us.
+
+	return SDL_GetCPUCount();
```
johnnovak, line 7063: Better, but why not just use `SDL_GetCPUCount()` and be done with it? 😆 The C/C++ standard is sometimes a bit silly because it needs to support all sorts of weirdo hardware... I'd bet large sums of money `SDL_GetCPUCount()` is spot on all the time on the much narrower platforms supported by SDL.

weirddan455:
I found the Linux bottleneck.  It was in this Sempahore wait code.  I don't really have a good solution yet but I uploaded a branch that replaces the mutex/conditional variable wait with an atomic compare and swap busy wait and it's 4x faster now and I can see a real performance improvement from the extra threads.

Again, not a great solution because it maxes out all the cores but if anyone wants to test/benchmark for themselves here it is:  https://github.com/dosbox-staging/dosbox-staging/tree/wd/voodoo-busy-wait

The real problem is probably that we're waiting more than we should and Windows must be faster at waking the threads up from a conditional variable.

src/hardware/voodoo.cpp
interloper98: Yeah.. I've added a log of the number of threads detected, so combined with the manual setting, worst case for a mis-detect is the user had to punch in their desired number of threads (and that's going to pretty rare!)

weirddan455:
QDOS kept throwing an error for me but I got Q2DOS to work.

https://archive.org/details/quake-2-play

1. Copy vooodoo/sst1/glide3x.dxe to the root directory.
2. Run q2.exe
3. Go into options -> video, and select the 3dfx renderer

Tested with stock settings in Linux at 640X480 resolution.  Unfortunately, this game does not seem to scale well but the busy wait gives a boost.

3 threads - 7.9 fps
3 threads busy wait - 12.7 fps
7 threads - 7.8 fps
7 threads busy wait - 12.6 fps
14 threads busy wait - 11.3 fps

Descent 2 starting area (after the door closes for more FPS):
3 threads - 50 fps
3 threads busy wait - 100 fps
7 threads - 75 fps
7 threads busy wait - 200 fps
14 threads - 80 fps
14 threads busy wait - 150 fps

Note that I'm on an 8 core CPU so busy waiting on 14 threads is only utilizing hyperthreads.  Busy waiting when hyperthreading is involved is a bad idea because a core doing actual work can get slowed down by its hyperthread sibling that's busy waiting.  Descent 2 may actually scale up quite nicely on a CPU with lots of cores.

I'll edit this with Windows results in a bit.

interloper98:
@weirddan455 , thanks for those steps to get Quake 2 working.

Took a recording of my Win11 desktop, AMD settings, and `timedemo 1` using this branch followed by your branch with the busy wait, both with 7 threads. I think my speed up is > 10%, but I'm not sure where to inspect the FPS value.

Here's the recording (quality is pretty rough to fit it all in 10 MB):

https://github.com/user-attachments/assets/48d140a0-877d-46de-bf80-542cbc873a39

weirddan455:
>  I'm not sure where to inspect the FPS value

You didn't input the 2nd command.

```
timedemo 1
map demo1.dm2
```

Then open the console after the demo has finishing playing and the result should be in there.

interloper98:
> Then open the console after the demo has finishing playing and the result should be in there.

Thanks! That did it, and busy wait is rocking:
 - 7 threads mutex/conditional: 50.5 fps
 - 7 threads busy wait: 63.9 fps

It's pushed Descent 2 further, too:

https://github.com/user-attachments/assets/707d6759-952c-44fe-b4bd-0ebe30e7ca8c

johnnovak:
I have a hunch using a "wait for multiple objects" call would fix the Linux situation.

Windows has such an API call, I know that for sure.

Related:
https://stackoverflow.com/questions/7780749/waiting-on-multiple-semaphores-without-busy-waiting-c-c-linux

johnnovak:
> Linux has `futex_waitv` since kernel 5.16 (Linux specific though, not POSIX). It was written by a Valve contractor to improve performance for Windows games that use `WaitForMultipleObjects` so it serves the same purpose.

Good stuff 👍🏻

interloper98:
@weirddan455 , Linux (catchyos, kernel 6.11) pulls ahead of win11 (ghostspectre) with your busywait branch :sunglasses: 

image0011-rendered

interloper98:
New conf setting `voodoo_threads = yes / no / <value>` is ready to go.

@Torinde, if you've got an Epyc or ThreadRipper.. crank up that thread count, let out the guide bot, and let the sparks fly! :sparkle: :sparkler: :sparkles:

src/dosbox.cpp
```diff
@@ -834,8 +834,19 @@ void DOSBOX_Init()
...
+	// Deprecate the boolean Voodoo multithreading setting
+	pbool = secprop->Add_bool("voodoo_multithreading", deprecated, false);
+	pbool->Set_help("Use 'voodoo_threads = yes / no / <value>'.");
+
+	pstring = secprop->Add_string("voodoo_threads", only_at_start, "yes");
+	pstring->Set_help(
+	        "Use threads to improved 3dfx Voodoo performance ('yes' by default).\n"
+	        "  yes:      Use up to 7 threads based on available CPU cores (default).\n"
+	        "  no:       Disable threads: low frame rates with only one core used.\n"
+	        "  <value>:  Set a specific number of threads.\n"
```
weirddan455, line 846: IMO `auto` is more descriptive than `yes` and `no` can be removed since it's the same as a value of 0.  If you want, you can put a note that these are _additional_ worker threads (the main thread always does Voodoo work as well) so a good value would be core count minus 1.

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7054,40 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
 }
 
+static int get_num_threads()
+{
+	auto get_auto_threads = []() {
+		// Sweet spot found to be seven threads based on benchmarks:
+		// https://github.com/schellingb/dosbox-pure/issues/300#issuecomment-2144050898
+		constexpr auto MinThreads = 1;
+		constexpr auto MaxThreads = 7;
+		return std::clamp(SDL_GetCPUCount() - 1, MinThreads, MaxThreads);
```
weirddan455, line 7064: Then here set `MinThreads` to 0 to remove the special case of `no`.  This also has the benefit of a single core CPU not spawning any additional worker threads which would probably only hurt performance.

weirddan455 (approved):
Small suggestion about the config option but otherwise this all looks good to me.  Nice job :+1:

src/dosbox.cpp
johnnovak: Seconded, should be `auto` and the number option.

interloper98: Great suggestions @weirddan455

src/hardware/voodoo.cpp
interloper98: Nice! Added.

src/dosbox.cpp
interloper98: > I would also consider `none/off` options - just like in `pcspeaker`, `oplmode`, `window_titlebar`, etc.

Will add that @FeralChild64; makes sense, and good to be consistent. Thanks.

interloper98:
Ok guys, ready for the next stage: @weirddan455 strapping rockets on those semaphore locks :rocket:

weirddan455:
> Ok guys, ready for the next stage: @weirddan455 strapping rockets on those semaphore locks 🚀

Yeah, I think merging this first is best and then I'll toy with the locks to see if I can find something performant that's better than busy waiting.

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7054,54 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+		// Sweet spot found to be seven threads based on benchmarks:
+		// https://github.com/schellingb/dosbox-pure/issues/300#issuecomment-2144050898
+		constexpr auto MinThreads = 0;
+		constexpr auto MaxThreads = 7;
+		return std::clamp(SDL_GetCPUCount() - 1, MinThreads, MaxThreads);
+	};
+
+	auto parse_positive_base10_int = [](const std::string& str) {
+		constexpr auto Base10 = 10;
+		const auto maybe_int  = parse_int(str, Base10);
```
johnnovak, line 7073: This is noisy, just do `parse_int`; base10 is the default (`const int base = 10`).

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7054,54 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+	};
+
+	// Get the user's "[voodoo] voodoo_threads" setting
+	constexpr auto DefaultSetting = "auto";
+	const auto sec = dynamic_cast<Section_prop*>(control->GetSection("voodoo"));
+	const auto user_setting = sec ? sec->Get_string("voodoo_threads")
+	                              : DefaultSetting;
+
+	// Did the user provide the default or a standardized positive setting?
+	if (user_setting == DefaultSetting || has_true(user_setting)) {
```
johnnovak, line 7084: We allow `auto`, or a number, so `has_true` needs to be ditched. Allowing yes/no, enabled/disabled, etc. confuses the matters for no good reason, so please drop it.

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7054,54 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+		return get_auto_threads();
+	}
+	// Did the user provide a custom value?
+	if (const auto maybe_int = parse_positive_base10_int(user_setting)) {
```
johnnovak, line 7088: `parse_positive_int`

src/hardware/voodoo.cpp
```diff
@@ -6996,6 +7054,54 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+		return *maybe_int;
+	}
+	// Did the user provide a standardized negative setting?
+	if (has_false(user_setting)) {
+		return 0;
+	}
```
johnnovak, line 7094: Drop this please, as per the above explanation.

src/hardware/voodoo.cpp
johnnovak: It would be slightly more descriptive if you renamed `DefaultSetting` to `AutoSetting`.

src/dosbox.cpp
```diff
@@ -834,8 +834,19 @@ void DOSBOX_Init()
 	        "   4: 2 MB for the FBI and one TMU with 2 MB (default).\n"
 	        "  12: 4 MB for the FBI and two TMUs, each with 4 MB.");
 
-	pbool = secprop->Add_bool("voodoo_multithreading", only_at_start, true);
-	pbool->Set_help("Use threads to improve 3dfx Voodoo performance (enabled by default).");
+	// Deprecate the boolean Voodoo multithreading setting
+	pbool = secprop->Add_bool("voodoo_multithreading", deprecated, false);
+	pbool->Set_help("Use 'voodoo_threads = auto / <value>'.");
```
johnnovak, line 839: Not bad, but for consistency I'd suggest to use simply this:

``Renamed to 'voodoo_threads'``

Then we don't need to duplicate information in the deprecation text (one less thing to get out of sync).

image

src/dosbox.cpp
```diff
@@ -834,8 +834,19 @@ void DOSBOX_Init()
...
+
+	pstring = secprop->Add_string("voodoo_threads", only_at_start, "auto");
+	pstring->Set_help(
+	        "Use threads to improved 3dfx Voodoo performance:\n"
```
johnnovak, line 843: typo: improve (drop the "d")

src/dosbox.cpp
```diff
@@ -834,8 +834,19 @@ void DOSBOX_Init()
...
+	        "  auto:     Use up to 7 threads based on available CPU cores (default).\n"
+	        "            This is in addition to the permanent main thread.\n"
```
johnnovak, line 845: Drop the "This is in addition to the permanent main thread." part (doesn't add value, confusing, and probably not even true?)

src/dosbox.cpp
```diff
@@ -834,8 +834,19 @@ void DOSBOX_Init()
...
+	        "  <value>:  Set a specific number of additional threads, for example: 12\n"
```
johnnovak, line 846: Just "Set a specific number of threads." (we don't provide examples for trivial settings such as numbers).

src/dosbox.cpp
```diff
@@ -834,8 +834,19 @@ void DOSBOX_Init()
...
+	        "Note: Tests show that frame rates increase up to 7 threads after which\n"
+	        "      they level off or decrease, in general.\n"
```
johnnovak, line 848: Minor: Start the next sentence on this line (we don't do paragraphs like this).

src/hardware/voodoo.cpp
```diff
@@ -7805,8 +7913,11 @@ void VOODOO_Init(Section* sec)
...
 
 	// Log the startup
-	LOG_MSG("VOODOO: Initialized with %s MB of RAM, %smultithreading, and %sbilinear filtering",
+	const auto num_threads = get_num_threads();
+
+	LOG_MSG("VOODOO: Initialized with %s MB of RAM, %d %s, and %sbilinear filtering",
 	        memsize_pref.c_str(),
-	        (voodoo_multithreading ? "" : "no "),
+	        num_threads,
+	        num_threads == 1 ? "thread" : "threads",
```
johnnovak, line 7921: Nice touch 😎

johnnovak:
@interloper98 Nice job man! 🎉 

I hadn't had a chance to finish my review of your recent changes before you merged, so I left a few additional comments now. Could you please address these in a follow-up PR?

src/hardware/voodoo.cpp
interloper98: Thanks, added.

src/hardware/voodoo.cpp
interloper98: @FeralChild64 suggested that 'none' be allowed to match other conf settings, but most (all?) numeric settings don't allow 'none'; for example: `[mixer] rate = none` and `[sblaster] irq = none` aren't allowed..  So OK, that makes sense.
Fixed both.

src/dosbox.cpp
interloper98: Ok!

src/dosbox.cpp
interloper98: thanks.

src/dosbox.cpp
interloper98: Ok, dropped. @weirddan455 suggested this because the original wording might imply that at least 1 thread is needed for Voodoo. _"If I set 0, then Voodoo won't even work"_; so this was an attempt to clarify that Voodoo always runs, even with zero threads.

src/dosbox.cpp
interloper98: Dropped in the suggested wording and limits. Thanks!

src/dosbox.cpp
interloper98: dropped!

src/hardware/voodoo.cpp
interloper98: thanks :-)
OK.. sending the PR. Thanks for the wrap-up comments.

interloper98:
Thanks for the detailed feedback @johnnovak; sending these last round of fixes in a follow up PR.

PR #3939, opened 2024-09-20
Address additional feedback on Voodoo thread count PR #3935

interloper98:
# Description

(Per title)

# Manual testing

### Conf block:

```ini
#            voodoo_threads: Use threads to improve 3dfx Voodoo performance:
#                              auto:     Use up to 8 threads based on available CPU cores (default).
#                              <value>:  Set a specific number of threads between 1 and 16.
#                            Note: Tests show that frame rates increase up to 8 threads after which
#                                  they level off or decrease, in general.
```

### Results with valid values:

1. **`voodoo_threads = auto`** (on 4 CPU core rpi)

    `VOODOO: Initialized with 4 MB of RAM, 4 threads, and no bilinear filtering`

1. **`voodoo_threads = 1`**

    `VOODOO: Initialized with 4 MB of RAM, 1 threads, and no bilinear filtering`

1. **`voodoo_threads = 16`**

    `VOODOO: Initialized with 4 MB of RAM, 16 threads, and no bilinear filtering`

### Results with invalid values:

1. **`voodoo_threads = yes`** (on 4 CPU core rpi)

    `VOODOO: Invalid 'voodoo_threads' setting: 'yes', using 'auto'`
    `VOODOO: Initialized with 4 MB of RAM, 3 threads, and no bilinear filtering`

1. **`voodoo_threads = no`** (on 4 CPU core rpi)

    `VOODOO: Invalid 'voodoo_threads' setting: 'no', using 'auto'`
    `VOODOO: Initialized with 4 MB of RAM, 3 threads, and no bilinear filtering`

1. **`voodoo_threads = -1`**
    `CONFIG: Invalid 'voodoo_threads' setting: '17'. Value outside of the valid range 1-16, using '1'`
    `VOODOO: Initialized with 4 MB of RAM, 1 threads, and no bilinear filtering`

1. **`voodoo_threads = 128`**

    `CONFIG: Invalid 'voodoo_threads' setting: '128'. Value outside of the valid range 1-16, using '16'`
    `VOODOO: Initialized with 4 MB of RAM, 16 threads, and no bilinear filtering`

# Checklist

I have:

### DIFF
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -7054,52 +7054,49 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
 	return 0xffffffff;
 }
 
-// Get the number of additional threads to use for Voodoo work based on the
-// user's conf setting. By default we use up to 7 threads however the user can
-// customize this.
+// Get the number of total threads to use for Voodoo work based on the user's
+// conf setting. By default we use up to 8 threads (which includes the main
+// thread) however the user can customize this.
 
-static int get_num_threads()
+static int get_num_total_threads()
 {
-	auto get_auto_threads = []() {
-		// Sweet spot found to be seven threads based on benchmarks:
-		// https://github.com/schellingb/dosbox-pure/issues/300#issuecomment-2144050898
-		constexpr auto MinThreads = 0;
-		constexpr auto MaxThreads = 7;
-		return std::clamp(SDL_GetCPUCount() - 1, MinThreads, MaxThreads);
-	};
+	constexpr auto MinThreads     = 1;
+	constexpr auto MaxAutoThreads = 8;
+	constexpr auto MaxThreads     = 16;
 
-	auto parse_positive_base10_int = [](const std::string& str) {
-		constexpr auto Base10 = 10;
-		const auto maybe_int  = parse_int(str, Base10);
-		return (maybe_int && *maybe_int >= 0) ? maybe_int : std::nullopt;
-	};
+	constexpr auto SectionName = "voodoo";
+	constexpr auto SettingName = "voodoo_threads";
+	constexpr auto AutoSetting = "auto";
 
-	// Get the user's "[voodoo] voodoo_threads" setting
-	constexpr auto DefaultSetting = "auto";
-	const auto sec = dynamic_cast<Section_prop*>(control->GetSection("voodoo"));
-	const auto user_setting = sec ? sec->Get_string("voodoo_threads")
-	                              : DefaultSetting;
+	const auto sec = dynamic_cast<Section_prop*>(control->GetSection(SectionName));
+	const auto user_setting = sec ? sec->Get_string(SettingName) : AutoSetting;
 
-	// Did the user provide the default or a standardized positive setting?
-	if (user_setting == DefaultSetting || has_true(user_setting)) {
-		return get_auto_threads();
+	if (const auto maybe_int = parse_int(user_setting)) {
+		const auto valid_int = std::clamp(*maybe_int, MinThreads, MaxThreads);
+
+		// Use a property to test and warn if the value's outside the range
+		constexpr auto always_changeable = Property::Changeable::Always;
+		auto range_property = Prop_int(SettingName, always_changeable, valid_int);
+		range_property.SetMinMax(MinThreads, MaxThreads);
+
+		if (!range_property.IsValidValue(*maybe_int)) {
+			set_section_property_value(SectionName,
+			                           SettingName,
+			                           std::to_string(valid_int));
+		}
+		return valid_int;
 	}
-	// Did the user provide a custom value?
-	if (const auto maybe_int = parse_positive_base10_int(user_setting)) {
-		return *maybe_int;
-	}
-	// Did the user provide a standardized negative setting?
-	if (has_false(user_setting)) {
-		return 0;
-	}
-	// The user provided an invalid setting.
-	LOG_WARNING("VOODOO: Invalid 'voodoo_threads' setting: '%s', using '%s'",
-	            user_setting.c_str(),
-	            DefaultSetting);
 
-	set_section_property_value("voodoo", "voodoo_threads", DefaultSetting);
+	if (user_setting != AutoSetting) {
+		LOG_WARNING("VOODOO: Invalid '%s' setting: '%s', using '%s'",
+		            SettingName,
+		            user_setting.c_str(),
+		            AutoSetting);
 
-	return get_auto_threads();
+		set_section_property_value(SectionName, SettingName, AutoSetting);
+	}
+
+	return std::clamp(SDL_GetCPUCount(), MinThreads, MaxAutoThreads);
 }
 
 /***************************************************************************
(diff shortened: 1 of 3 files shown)


johnnovak:
Thank you for doing this @interloper98 🎉 

Looks perfect, just one last thing, and I'm not trying to test your patience, trust me 😅 

I was wondering if it would be better if we'd count the always present "base emulation thread" into the total count threads. So, effectively change the range from 0-16 to 1-16 (should be trivial) and update the config setting's description to this:

```
#            voodoo_threads: Number of threads to use for emulating the 3dfx Voodoo:
#                              auto:     Use up to 8 threads based on available CPU cores (default).
#                              <value>:  Set a specific number of threads between 1 and 16.
#                            Note: Tests show that frame rates increase up to 8 threads after which
#                                  they level off or decrease, in general.
```

I guess I'd prefer this because the valid `voodoo_threads = 0` settings feels a bit strange. With this slight adjustment, the definition of `voodoo_threads` would be nice and high level and actually reflect reality: just the total number of threads used for emulating the Voodoo, which is always at least 1 thread.

The fact that we always have our "base emulation thread" and optionally some extra worker threads is an implementation detail; it's none of the end user's business. This way we'd fully get rid of the "necessity" to call these "extra" or "additional" threads (but then, I have no qualms about not calling these "extra" threads even if leave the PR as is; we should curb our programmer's OCD tendencies a bit and try to keep regular non-OCD users in mind 😆  Right @Torinde ? 😉 )

src/hardware/voodoo.cpp
```diff
@@ -7060,46 +7060,43 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
-		return 0;
+	                              : AutoSetting;
+
+	if (const auto maybe_int = parse_int(user_setting)) {
+		const auto valid_int = std::clamp(*maybe_int, MinThreads, MaxThreads);
+
+		if (*maybe_int != valid_int) {
+			LOG_WARNING(
+			        "VOODOO: Invalid custom 'voodoo_threads' quantity: '%s', "
+			        "Valid values range from '%d' to '%d'. Using '%d'",
```
johnnovak, line 7078: Minor suggestion: maybe we can follow the wording of the generic out-of-range error message of pure integer conf settings.

image

interloper98:
> So, it's counting not only the physical CPU cores, but also the logical ones...

@Torinde, can you test the Q2DOS benchmark on your 2 core / 4 thread laptop and compare `voodoo_threads = auto` versus `voodoo_threads = 1` results?

This is really important because most Steam gamers have fewer than eight physical cores: https://store.steampowered.com/hwsurvey/cpus

My guess is that letting hyper threads participate in the work shouldn't hurt performance. It's like painting 100 boards on a fence: you can finish the fence faster with a couple slow helpers. This is the same here because there's a lot more triangles than cores/hyperthreads, per frame.

johnnovak:
My understanding is hyper threads (logical cores) also result in performance improvements, just not as big as with real cores (otherwise why would they exist if they don't result in performance improvements, right?)

So while going from 4 to 8 threads on a trivially parallelisable task (e.g. raytracing) would yield a 2x improvement, on 4 physical cores adding 4 virtual cores would result in a ~25-30% improvement. Which means there it makes no sense to not allocate work to virtual cores / hyperthreads as well (again, why would they exist at all then if they are "not to be used"?). The performance improvement is there, just more modest.

Grounded0:
Hyperthreading was implemented because the front end couldn't keep the core execution resources fully utilised without dividing tasks to two separate threads for the front end logic:

Sandy Bridge (client) - Microarchitectures - Intel:

https://en.wikichip.org/wiki/intel/microarchitectures/sandy_bridge_(client)

src/hardware/voodoo.cpp
interloper98: Will do!

interloper98:
@johnnovak , I'll try to think of a way to start there thread count at 1.

The biggest problem right now is that the bulk of the threading code uses the idea of num_workers and num_threads, where the threads can be 0 and workers is always threads+1, to account for the existing non-threaded worker, which always exists. 

Currently the user space conf settings matches the code implementation.

If we go to "minimum 1 thread", then we need a minus - 1 to reconcile these two halves.

johnnovak:
> I think John's idea is not to change anything in the code except the mapping between .conf value and "threads"

Exactly.



> But I agree that if there is hyperthreading, then apparently the architects of that CPU made such tradeoffs that made hyperthreading useful for it. In some special case software it may happen that performance worsens, but generally should be fine to use as many threads as there are logical cores.

Yes, that's my experience when running ray tracing jobs in Blender on a 4790k. Haven tried my Ryzen 7900 with Blender yet, but I expect similar improvements when fully utilising all 24 threads. I think it's best to think of the hyper threads as threads/cores with lower performance in most practical contexts -- still massively useful in most CPU bound jobs (rendering, video encoding, audio work, etc.).

Grounded0:
> (e.g. 1 thread is sufficient to consume the whole power budget of 5W fanless, which is not a lot)

What kind of TDP rating does the CPU have?

I got 45 W TDP on a laptop and thats enough. No performance without power consumption at least on the x86.

interloper98:
Ok; ready for review.

Was able to use a local `Prop_int` to do the range check and error logging; a lot DRYer now.

Also caught a small inconsistency in the log wording in setup.cpp, which is adjusted in a separate commit.

src/hardware/voodoo.cpp
```diff
@@ -7054,52 +7054,49 @@ static constexpr uint32_t voodoo_r(const uint32_t addr)
...
+	const auto sec = dynamic_cast<Section_prop*>(control->GetSection(SectionName));
+	const auto user_setting = sec ? sec->Get_string(SettingName) : AutoSetting;
+
+	if (const auto maybe_int = parse_int(user_setting)) {
+		const auto valid_int = std::clamp(*maybe_int, MinThreads, MaxThreads);
+
+		// Use a property to test and warn if the value's outside the range
+		constexpr auto always_changeable = Property::Changeable::Always;
+		auto range_property = Prop_int(SettingName, always_changeable, valid_int);
+		range_property.SetMinMax(MinThreads, MaxThreads);
```
johnnovak, line 7080: Nice use case of exploiting "config setting polymorphism" 😆 👍🏻

src/hardware/voodoo.cpp
```diff
@@ -7112,9 +7109,10 @@ static int get_num_threads()
 static void voodoo_init() {
 	assert(!v);
 
-	const auto num_threads = get_num_threads();
+	// Deduct 1 because the main thread is always preset
```
johnnovak, line 7112: typo: present

johnnovak (approved):
This looks very good now @interloper98 & thanks for bearing with all my comments 👍🏻 😄

src/hardware/voodoo.cpp
interloper98: Fix inbound; thanks!

src/hardware/voodoo.cpp
interloper98: It's slightly clunky, but gets the job done. (clucky is better than duped codes :laughing: )

interloper98:
> This looks very good now @interloper98 & thanks for bearing with all my comments 👍🏻 😄

Gladly; that thread count function is a whole lot better all round now, too!

Grounded0:
2017 15" MacBook Pro with thread load 100 % and using `svga_s3`:

Screenshot 2024-09-21 at 23 32 18

interloper98:
@Torinde, @Grounded0 : thanks for running those benchmarking comparing core counts given the different power constraints. The results were enough to go ahead with the PR (I just had meant to thank you sooner but forgot prior to merging the PR 😌).

PR #3942, opened 2024-09-20
Fix byte calculation overflow using pages per MB

### LINKED ISSUE #3940: Crash when setting high `memsize` values on macOS
Setting `memsize` to high values (~2500 and above) results in weird recursive exception throwing and an eventual crash, apparently only on macOS:

image

Reproduce with the following config:

```ini
[dosbox]
memsize = 2500
```

This is a regression; values up to 3072 were working fine on 0.81.2.

Well, and very it's low priority because even server-grade Socket 5 and 7 motherboards only support up to 512 or 1024 MB of memory. More than 64 MB for DOS is rather pointless (that's the limitation of MS-DOS 6.22 via HIMEM.SYS).

I was happy with 64 MB of RAM in my first Celeron 350 Win95/98 box I bought in 1999, so... 😄 It was even fine for Windows 2000.

interloper98:
# Description

The calculation from megabytes to pages results in a large peak byte value causing an overflow of the `int`.

This uses a pages-per-megabyte variable to let the calculations read more naturally and also vastly reduces the peak value used when converting megabytes to pages.

## Related issues

Fixes #3940 

# Manual testing

Tested with large memory sizes up to 3000 MB, including sanitizer builds.

# Checklist

I have:

### DIFF
--- a/src/hardware/memory.cpp
+++ b/src/hardware/memory.cpp
@@ -28,10 +28,12 @@
 #include "setup.h"
 #include "support.h"
 
-constexpr auto megabyte = 1024 * 1024;
+constexpr auto Megabyte = 1024 * 1024;
+
+constexpr auto PagesPerMegabyte = Megabyte / dos_pagesize;
 
 constexpr auto MinMegabytes = static_cast<uint16_t>(1);
-constexpr auto MaxMegabytes = static_cast<uint16_t>(PciMemoryBase / megabyte);
+constexpr auto MaxMegabytes = static_cast<uint16_t>(PciMemoryBase / Megabyte);
 
 constexpr auto SafeMegabytesDos   = 31;
 constexpr auto SafeMegabytesWin95 = 480;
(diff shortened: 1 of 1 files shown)


src/hardware/memory.cpp
```diff
@@ -30,6 +30,8 @@
 
 constexpr auto megabyte = 1024 * 1024;
 
+constexpr auto PagesPerMegabyte = megabyte / dos_pagesize;
```
johnnovak, line 33: While you're at it: `Megabyte`

src/hardware/memory.cpp
```diff
@@ -158,7 +160,8 @@ PageHandler * MEM_GetPageHandler(Bitu phys_page) {
 		return memory.lfb.handler;
 	}
 
-	constexpr uint32_t pages_in_16mb = {0x01000000u / dos_pagesize};
+	constexpr uint32_t pages_in_16mb = 16 * PagesPerMegabyte;
```
johnnovak, line 163: `PagesIn16Mb`

johnnovak (approved):
This fixes the problem @interloper98! 🎉 Thanks for jumping on this so quickly 🏃🏻‍♂️ 🎖️

Issue #3917, opened 2024-09-06
Corrupted installation screen in Druid: Daemons of the Mind with latest 0.82.0 release candidate


KainXVIII:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.82.0-rc1 (9d55e)

### What Operating System are you using?

Windows 10

### Relevant hardware info

i5-3570, GTX 970

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Druids - Daemons of the Mind has corrupted installation program screen in dosbox 0.82.0-rc1 (9d55e) (all is fine in version 0.81.2)
rc1
![image0002](https://github.com/user-attachments/assets/7ac7fdd0-30e5-47c5-b0b5-d4986afe816f)

0.81.2
![image0001](https://github.com/user-attachments/assets/fd8a4b8b-0f37-4f46-b863-6df9dc27df4b)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Run install.bat from game cd.
2.
3.

### Your configuration

```ini
default config
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Thanks @KainXVIII

interloper98:
Bistected to be3d059469a867623026e6f180a779c547493a64

johnnovak:
> Bistected to [be3d059](https://github.com/dosbox-staging/dosbox-staging/commit/be3d059469a867623026e6f180a779c547493a64)

Loading the font based on the clocking mode register was a reasonable guess on my part. I still think that's generally the correct thing to do, but probably the font switching criteria is a bit more nuanced. Hopefully the VGA programming book holds all the answers 🤞🏻

PR #3943, opened 2024-09-21
Fix `INT10_ReloadFont()` regression on VGA

johnnovak:
# Description

This fixes a regression in the "Druid: Daemons of the Mind" installer where the whitespaces of the custom VGA font appear corrupted (introduced in https://github.com/dosbox-staging/dosbox-staging/commit/be3d059469a867623026e6f180a779c547493a64).

`INT10_ReloadFont()` should trigger the loading of alternate characters for 9-dot VGA fonts only, while `INT10_LoadFont()` should never do this. This has regressed during a recent cleanup & refactoring job.

This change also implements this conditional alternate charset loading mechanism more cleanly. Previously, the `font_block` argument was used as a "side channel" for signalling this criteria via a hack (by using a magic value of 0x80 which is not a valid value in the real BIOS call).

The last commit is the actual fix, the previous ones are  cleanup & refactoring jobs that don't change existing functionality (famous last words 🤞🏻 😆).

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3917

# Release notes

N/A

# Manual testing

- Confirmed the installer of ""Druid: Daemons of the Mind" appears correctly (also with `vga_8dot_font = on`):

image

- Confirmed the following MODE commands yield 9-dot fonts with the default config (720 pixel wide modes), and 8-dot fonts with `vga_8dot_fonts = on` (640 pixel wide modes). (The remaining custom modes all use the 8x8 font, so the screen widths are always 640 for 80-column modes, and 1056 for 132-column modes.)
  - `MODE 80x25`
  - `MODE 80x28`
  - `MODE 80x30`
  - `MODE 40x25` (this is half-width, so 360/320 pixels)

- Executed a good portion of the [VGA text mode tests](https://github.com/dosbox-staging/dosbox-staging/wiki/Video-emulation-tests#text-modes).

# Checklist

I have:

### DIFF
--- a/src/ints/int10_memory.cpp
+++ b/src/ints/int10_memory.cpp
@@ -369,45 +369,58 @@ void INT10_SetupRomMemory(void) {
 	}
 }
 
-void INT10_ReloadRomFonts(void) {
+void INT10_ReloadRomFonts(void)
+{
 	// 16x8 font
-	PhysPt font16pt=RealToPhysical(int10.rom.font_16);
-	for (Bitu i=0;i<256*16;i++) {
-		phys_writeb(font16pt+i,int10_font_16[i]);
+	PhysPt font16pt = RealToPhysical(int10.rom.font_16);
+	for (Bitu i = 0; i < 256 * 16; i++) {
+		phys_writeb(font16pt + i, int10_font_16[i]);
 	}
-	phys_writeb(RealToPhysical(int10.rom.font_16_alternate),0x1d);
+	phys_writeb(RealToPhysical(int10.rom.font_16_alternate), 0x1d);
+
 	// 14x8 font
-	PhysPt font14pt=RealToPhysical(int10.rom.font_14);
-	for (Bitu i=0;i<256*14;i++) {
-		phys_writeb(font14pt+i,int10_font_14[i]);
+	PhysPt font14pt = RealToPhysical(int10.rom.font_14);
+	for (Bitu i = 0; i < 256 * 14; i++) {
+		phys_writeb(font14pt + i, int10_font_14[i]);
 	}
-	phys_writeb(RealToPhysical(int10.rom.font_14_alternate),0x1d);
+	phys_writeb(RealToPhysical(int10.rom.font_14_alternate), 0x1d);
+
 	// 8x8 fonts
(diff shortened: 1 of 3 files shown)


weirddan455 (approved):
Nice fix.  This looks fine to me.

src/ints/int10_memory.cpp
```diff
@@ -149,49 +150,41 @@ void INT10_LoadFont(const PhysPt _font, const bool reload, const int count,
...
-		               Offset,
-		               Map,
-		               Height);
-		break;
+		const auto font_data      = RealToPhysical(int10.rom.font_16);
+		constexpr auto CharHeight = 16;
+
+		INT10_LoadFont(font_data, Reload, NumChars, FirstChar, FontBlock, CharHeight);
+	} break;
 	}
```
interloper98: No complaints here; but noticed the switch-case only really controlling one variable, so allows for something like this:

```c++
void INT10_ReloadFont()
{
	constexpr auto Reload    = false;
	constexpr auto NumChars  = 256;
	constexpr auto FirstChar = 0;
	constexpr auto FontBlock = 0;

	const auto char_height = CurMode->cheight;

	const auto font_data = RealToPhysical(
		char_height == 8  ? int10.rom.font_8_first :
		char_height == 14 ? int10.rom.font_14 :
		                    int10.rom.font_16
  );

	INT10_LoadFont(font_data, Reload, NumChars, FirstChar, FontBlock, char_height);
}
```

I find it a bit easier to read and quickly see the variables in-play, and the ternary worm reads like a switch/case.  

Edit - I see one bool added in the last commit, so might be doable. Just wanted to toss it as an option :-)

johnnovak: That's a good idea @interloper98, I'll condense the switch/case block as suggested.

interloper98 (approved):
Nice fix @johnnovak and very helpful documentation added about how these settings work.

johnnovak:
Thanks  for the review @interloper98 & @weirddan455 👍🏻

src/ints/int10_memory.cpp
johnnovak: Done.

PR #3950, opened 2024-09-23
Update debian cross docker images and CI to debian 12

shermp:
# Description

There has been recent chat about GCC 10 lacking support of some C++20 features. Since the only CI actions using GCC 10 are the debian cross compile actions, I decided to upgrade from Debian 11 "Bullseye" to Debian 12 "Bookworm". Debian 12 includes GCC 12, so should have decent C++20 support.

Since the main goal of the cross compilers is to sanity check ARM builds, and one of the main ARM platforms Raspberry Pi OS is now based on Bookworm, it makes sense to upgrade the images.

# Manual testing

The docker images were built locally before creating the remote branch.

# Checklist

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [ ] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] provided the release notes draft (for significant user-facing changes).
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #3952, opened 2024-09-23
Use atomic operations and busy wait for Voodoo threads

### LINKED ISSUE #3947: Voodoo performance bottleneck in Semaphore locks

_No response_

Linux x86_64

Debian 12

### Relevant hardware info

Ryzen 7 1700, Radeon RX 560

I described the problem in comments on the PR #3935 and uploaded a branch as a proof of concept.

https://github.com/dosbox-staging/dosbox-staging/tree/wd/voodoo-busy-wait

Performance of busy waiting is often 2x or more FPS than with these Semaphore locks.  This problem is more pronounced on Linux where it can see up to a 4x increase but it affects Windows as well.

So far, I have not found a better solution than busy waiting that achieves the same performance gain.  I've tried the following different approaches:

1. Using POSIX semaphores, removing our custom class.
2. Busy waiting on just the main thread.
3. Using a single condition variable shared among the worker threads and `notify_all()` to wake them all up with a single call.

None of these had a significant performance gain.  We're simply waiting too often.  I measured that we're sometimes waking each thread up 10 times per ms (in other words, trying to sleep the threads for 100 microseconds).

With calls this frequent, busy waiting starts to make sense unless someone (not me) wants to really dive into this Voodoo code and do some better batching/queuing.

Type "frametime" in Descent 2

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

- Fix an assertion failure when `voodoo_threads = 1`
- Removes the Semaphore class and replaces it with atomic operations.  You can call it busy waiting but lock-free programming sounds better.  There's no more locks in the Voodoo code after this 😄
- No more assigning chunks to specific threads.  There is now an atomic chunk index.  Whoever gets there first does the work.  This can help if the OS scheduler sends a thread out to lunch for a while.
- `voodoo_threads = auto` is now based on physical cores rather than logical cores.  I see a performance decrease when number of threads exceeds physical cores.  Since the threads are spinning all the time, hyperthreading hurts more than it helps.

## Related issues

Fixes #3947

# Release notes

Performance increase for 3dfx Voodoo emulation

# Manual testing

Benchmark numbers are about the same as I previously reported here:  https://github.com/dosbox-staging/dosbox-staging/pull/3935#issuecomment-2357320164

Tested with `taskset -c 0-7 dosbox` to simulate a quad core with hyperthreading by setting process affinity.  On my system, the hyperthread siblings are sequential so logical cores 0 and 1 share physical core ID 0 and so on.  This can vary based on CPU so check `/proc/cpuinfo` if you're on Linux.

If someone with an actual 2 or 4 core CPU with hyperthreading wants to test and see if more threads helps or hurts performance, that would be useful information.

# Checklist

I have:


src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
...
+
+#include <sys/sysctl.h>
+
+int get_num_physical_cpus()
+{
+	int num_cpus = 0;
+	size_t size = sizeof(num_cpus);
+	sysctlbyname("hw.physicalcpu", &num_cpus, &size, nullptr, 0);
+	return std::max(num_cpus, 1);
+}
```
weirddan455, line 755: This Mac path is untested by the way.  Apple wins with this API being by far the simplest so hopefully it just works.

johnnovak: @weirddan455 
With this code I'm getting the following on my M1 MacBook:

image

So it reports 10, whereas it should only report 8 I think.

I can get the number of performance cores (8) by querying `hw.perflevel0.physicalcpu`. See here:
https://forums.developer.apple.com/forums/thread/692671

image

weirddan455: Oh I didn't think about E-cores.  They're still physical cores so I assume the API is working as expected.  Unsure if we want to be spinning up extra threads for E cores or not.  I'm pretty sure Windows and Linux count E cores as physical cores on the newer Intel CPUs as well so we'll have to make the same decision for x86.  I don't have one of those CPUs to benchmark on though.

I guess the M1 chips don't have a hyper-threading equivalent?  That was what this was trying to remove from the count.

johnnovak: > I guess the M1 chips don't have a hyper-threading equivalent? That was what this was trying to remove from the count.

Apparently they don't:
https://www.quora.com/Why-doesnt-Apple-use-Hyper-Threaded-CPUs-instead-of-multiple-cores-like-other-companies-do

Cool, then your original code is working as expected 👍🏻

weirddan455: My hunch says we should probably exclude E cores but without hardware I'm just guessing really.

It won't actually make a difference on your laptop or the high end Intel CPUs since we're clamping to a maximum of 8 threads.  It would mostly be these Core i5s that have 6 P-cores and 4 or 8 E-cores:  https://en.wikipedia.org/wiki/Raptor_Lake#List_of_14th_generation_Raptor_Lake_processors

I'm also not sure if the OS prefers to schedule these high-usage threads on E-cores or the logical (hyperthreaded sibling) P-cores.

johnnovak: Yeah, I'm fine to leave it as-is for now, then complicate it later once we have data 😆

src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
...
+{
+	// Size of the buffer as input, gets modified on function call to size of return data
+	DWORD buffer_size = 10 * 1024 * 1024;
+	uint8_t* cpu_info = (uint8_t *)malloc(buffer_size);
+	GetLogicalProcessorInformationEx(RelationProcessorCore,
+	                                 (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX)cpu_info,
+	                                 &buffer_size);
+
+	// Return value is a variably sized struct and requires type-punning to read
+	// Whoever made this API was probably on substances that are illegal in many countries
```
johnnovak, line 771: Well, one would at least hope so 🤣

src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
...
+				int core_id = atoi(core_id_str);
+				if (core_id >= 0 && core_id < MaxCpus && !core_ids[core_id]) {
+					core_ids[core_id] = true;
+					++num_cores;
+				}
+			}
+			close(fd);
+		}
+	}
+	return std::max(num_cores, 1);
```
johnnovak, line 742: minor: some newlines & clang format would improve the readability slightly

src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
...
+}
+
+#elif defined (WIN32)
+
+#include <Windows.h>
+
+int get_num_physical_cpus()
+{
+	// Size of the buffer as input, gets modified on function call to size of return data
+	DWORD buffer_size = 10 * 1024 * 1024;
```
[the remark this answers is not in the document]
weirddan455: Nah I overdid it :laughing:

On my system it was less than 100 bytes per CPU core.  Can probably drop it down to 100KB and it'll be fine even on a Threadripper.

src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
 	return lang;
 }
 
+#if defined (__linux__)
+
```
interloper98, line 719: Suggest using these defines:

```c++
// Defined if compiling for OS from BSD family
#mesondefine BSD

// Defined if compiling for Linux (non-Android)
#mesondefine LINUX

// Defined if compiling for macOS
#mesondefine MACOSX

// Defined if compiling for Windows (any option)
#mesondefine WIN32
```

Looks like they're used in other code, like:

```
src/gui/sdlmain.cpp:#if defined(BSD) || defined(LINUX)
```

and

```
src/hardware/serialport/libserial.cpp:#if defined (LINUX) || defined (MACOSX) || defined (BSD)
```

(and others)

src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
...
+		PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX info =
+		        (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX)(cpu_info +
+		                                                   bytes_read);
+		bytes_read += info->Size;
+		++num_entries;
+	}
+	free(cpu_info);
+
+	// There should be one entry per core so all we care about is the number of entries we got back
+	return std::max(num_entries, 1);
```
interloper98, line 784: 🧿🌈🍄☮🪴 substance-induced API, indeed! Wow.

I went back to the `GetLogicalProcessorInformation` (win 2003+), versus the newer `GetLogicalProcessorInformationEx` (win 7+), and it looks like the prior non-Ex API is slightly more sane; probably when they were younger and smarter :grimacing: 

It could be used something like this:

```c++

DWORD total_bytes = 0;
if (!GetLogicalProcessorInformation(nullptr, &total_bytes)) {
    return 1;
}

const auto num_infos = total_bytes / sizeof(SYSTEM_LOGICAL_PROCESSOR_INFORMATION);

std::vector<SYSTEM_LOGICAL_PROCESSOR_INFORMATION> infos(num_infos);
    
if (!GetLogicalProcessorInformation(infos.data(), &total_bytes)) {
    return 1;
}

int num_hw_cores = 0;
for(const auto& info : infos) {
    if (info.Relationship == RelationProcessorCore) {
        ++num_hw_cores;
    }
}

return std::max(num_hw_cores, 1);
```

interloper98 (approved):
Just some very minor comments; this is a win across the board. Awesome, @weirddan455.

Q2DOS `timedemo 1`, `map demo1.dm2`:
- Pi4 main: 4.5 fps, branch: 5.4 fps
  usage/core main: ~25%, branch: 100%
 
- Ryzen linux: main: 45.7 fps, branch: 63.4 fps
  usage/core: main 20%, branch: 100%

src/misc/cross.cpp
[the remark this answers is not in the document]
johnnovak: You could use your `MaxCpus` constant here too.

src/misc/cross.cpp
johnnovak: We're moving away from Meson to CMake, so I don't think that's a good idea.

src/misc/cross.cpp
johnnovak: > I went back to the `GetLogicalProcessorInformation` (win 2003+), versus the newer `GetLogicalProcessorInformationEx` (win 7+), and it looks like the prior non-Ex API is slightly more sane; probably when they were younger and smarter 😬

Those psychedelics take a heavy tool if you keep mixing them for decades. Just look at the seminal final works of Philip K. Dick (I love him, by the way, but man, half of those ideas must be the result of some _heavy duty_ substance abuse, I have zero doubts about that... 😎)

src/misc/cross.cpp
interloper98: :man_shrugging: it's just consistent with the existing code base.  If we're planning to rename these with CMake indentifiers (whatever those might be), then the conversion still is done consistently with a mass search-and-replace.

johnnovak: Nah, you're right @interloper98. I haven't actually checked the sources. Yeah, let's go ahead with this, then we mass replace it when (if) needed.

src/misc/cross.cpp
interloper98: Sounds like I should read some of his stuff :smile: (At least Philip wasn't designing OS APIs! :laughing:)

johnnovak: Well, the film industry has been profiting from adapting his books to movies for quite some time:
https://en.wikipedia.org/wiki/List_of_adaptations_of_works_by_Philip_K._Dick

The mini-series Electric Dreams was also quite good:
https://en.wikipedia.org/wiki/Electric_Dreams_(2017_TV_series)

I quite enjoyed the novels Ubik, VALIS, The Three Stigmata of Palmer Eldritch, Martian Timeslip, Flow My Tears the Policeman Said, and The Transmigration of Timothy Archer.

interloper98: @johnnovak, thanks for the tips to get me going! I'm a newbie to his work; looking forward to it 🤘 🍻

FeralChild64:
I tested QDOS (Quake 1 3dfx), `timedemo demo1`, on my 13th generation Intel Core CPU:

- When I let the CPU reach normal idle temperature before testing, performance on my setup went from around 62 FPS (main branch) to around 88 FPS (this PR). Nice!
- When I run the test like 15 times one after another, to let the CPU heat up and thermal throttling to kick in, it was like 23 FPS (main) vs 30 FPS (this PR). Also nice!

To be honest, I'm a bit shocked. I wasn't aware the thermal throttling reduces the desktop CPU performance so much. With my next PC I'll definitely get a larger cooler :)

src/misc/cross.cpp
```diff
@@ -715,6 +715,88 @@ std::string cfstr_to_string(CFStringRef source)
...
 }
 
+#if defined (__linux__)
+
+#include <fcntl.h>
+
+int get_num_physical_cpus()
+{
+	constexpr int MaxCpus = 128;
+	bool core_ids[MaxCpus] = {};
```
FeralChild64, line 725: How about using `std::set<int> all_core_ids` instead of `core_ids[MaxCpus]` and `int num_cores`? Then:

```c++
...
if (core_id >= 0) {
    all_core_ids.insert(core_id);
}
...
```
and at the end `all_core_ids.size()` contains the number of cores.

interloper98:
> my 13th generation Intel Core CPU:
> 
> * When I let the CPU reach normal idle temperature before testing, performance on my setup went from around 62 FPS (main branch) to around 88 FPS (this PR). Nice!
> * When I run the test like 15 times one after another, to let the CPU heat up and thermal throttling to kick in, it was like 23 FPS (main) vs 30 FPS (this PR). Also nice!
> 
> To be honest, I'm a bit shocked. I wasn't aware the thermal throttling reduces the desktop CPU performance so much.

Holy crap; a 62% reduction - that's a quite a hair cut!

The tax man cometh, and he's wearing an **_Intel Inside ®_** hat.

FeralChild64:
Intel Heatside :D I could probably lessen the effect by allowing the fans to spin faster, but I want my PC to run reasonably quiet.

johnnovak:
> Intel Heatside :D I could probably lessen the effect by allowing the fans to spin faster, but I want my PC to run reasonably quiet.

Ryzen 7900 + big ass Noctua cooler with two low RPM 12 cm Noctua fans in a "silent" Antec case = no discernable noise here at max loads (I hate fan noise like the plague, so I've been building silent PCs for the last two decades).

kklobe:
200% -> 800% CPU usage from main to this on my M3. Good thing I don't play any voodoo games.

johnnovak:
> 200% -> 800% CPU usage from main to this on my M3. Good thing I don't play any voodoo games.

Good thing you can always mitigate that by limiting the thread count now 😉

kklobe:
> > 200% -> 800% CPU usage from main to this on my M3. Good thing I don't play any voodoo games.
> 
> Good thing you can always mitigate that by limiting the thread count now 😉

Great, I can lose all perf gains of the extra threads to go back to the original CPU usage! 😁

johnnovak:
> > > 200% -> 800% CPU usage from main to this on my M3. Good thing I don't play any voodoo games.
> > 
> > 
> > Good thing you can always mitigate that by limiting the thread count now 😉
> 
> Great, I can lose all perf gains of the extra threads to go back to the original CPU usage! 😁

More performance results in more stress on the CPU—am I missing anything here? 😆 Law of conservation of energy, etc... 😉

kklobe:
> > > > 200% -> 800% CPU usage from main to this on my M3. Good thing I don't play any voodoo games.
> > > 
> > > 
> > > Good thing you can always mitigate that by limiting the thread count now 😉
> > 
> > 
> > Great, I can lose all perf gains of the extra threads to go back to the original CPU usage! 😁
> 
> More performance results in more work and thus more stress on the CPU—am I missing anything here? 😆 Law of conservation of energy, etc... 😉

Frame rate increase GREATLY out of proportion to power increase 😁

weirddan455:
>  Great, I can lose all perf gains of the extra threads to go back to the original CPU usage! 😁

>  Frame rate increase GREATLY out of proportion to power increase 😁

Would you like to actually benchmark that @kklobe ?  On my system, this branch performs nearly as well with 4 threads as `main` does with 8 threads.

As far as I know, no one has benchmarked this on Mac or ARM so you're really just speculating.

src/misc/cross.cpp
weirddan455: The array is a lot more efficient, not that matters too much for this probing code that runs once or twice.  You can think of it as a "perfect hash map" in that the values are already normalized to a reasonable range and you don't actually have to hash anything.  Also `std::unordered_set` is supposed to be more performant than `std::set` unless you really need the ordering for some reason.

kklobe:
> > Great, I can lose all perf gains of the extra threads to go back to the original CPU usage! 😁
> 
> > Frame rate increase GREATLY out of proportion to power increase 😁
> 
> Would you like to actually benchmark that @kklobe ? On my system, this branch performs nearly as well with 4 threads as `main` does with 8 threads.
> 
> As far as I know, no one has benchmarked this on Mac or ARM (besides @interloper98 testing on his Raspberry Pi) so you're really just speculating.

Oh. Sorry I wasn't clear, this is on my M3 MacBook Pro, testing Blood.

johnnovak:
> Frame rate increase GREATLY out of proportion to power increase 😁

Yeah, and I'm not even arguing, really, because I don't care much about Voodoo games myself either 🤷🏻 😆 But maybe we could rethink our default thread count values if the power consumption is so excessive in general as you're saying.

Anyway, I'm happy to defer this to our local Voodoo enthusiasts and review PRs from the background with the occasional comments 😆

kklobe:
> Anyway, I'm happy to defer this to our local Voodoo enthusiasts and review PRs from the background with the occasional comments 😆

Ditto! It certainly does smooth things out a bit in Blood, so, nice improvement, but just nice to know that I'll likely want to drop the threads down from "auto" to a more palatable CPU consumption for my tastes.

src/misc/cross.cpp
johnnovak: Yeah, I prefer the array. Using a set is IMO overdoing it.

johnnovak:
> Not for this PR, but do you think it'll be beneficial if main thread isn't burdened by any Voodoo emulation - all 3dfx to go to other threads? Like already done for capture, CD audio, etc.
> 
> That will allow the main thread (which handles CPU emulation) to have more power budget (not sharing a core with 3dfx emulation), thus be faster, thus being able to feed better the 3dfx threads on the other cores.

Not sure if that's possible if the main Voodoo thread needs to go in lockstep with the CPU emulation and other stuff. But yeah, in theory it would be good, but probably no one wants to do it unless it's trivial.

kklobe:
I tried a quick experiment (kk/voodoo-stdbarrier-1) with `std::barrier` which looks promising - about 80-85% of the speed benefit of the busywait with less than 1/2 the CPU utilization, but it's not supported prior macOS 11 (we still support 10.15 on the x86 build), or on the Debian 11 PPC build. So, something to keep in mind for the future when we deprecate the older OSes.

weirddan455:
> Not for this PR, but do you think it'll be beneficial if main thread isn't burdened by any Voodoo emulation - all 3dfx to go to other threads? Like already done for capture, CD audio, etc.
> 
> That will allow the main thread (which handles CPU emulation) to have more power budget (not sharing a core with 3dfx emulation), thus be faster, thus being able to feed better the 3dfx threads on the other cores.

Capture is still done on the main thread for now.  Voodoo differs from other threaded tasks like CD audio and MT32 in that it's still done synchronously.  As the code is architected now, the main thread must wait until the worker threads have completed their work before it can do anything else so it may as well help out to get that work done faster (it quite literally has nothing better to do).

Moving Voodoo to a more asynchronous model would probably be better in theory but would require a lot more work.  I'm not quite sure the best way to go about implementing that and I have some work planned that is higher priority for me so I'm not looking at this in the near future but it is something that can be improved upon.

>  I tried a quick experiment (kk/voodoo-stdbarrier-1) with std::barrier which looks promising - about 80-85% of the speed benefit of the busywait with less than 1/2 the CPU utilization

I just tested that branch.  Unfortunately, I do not have the same results.  On my system, barrier is about the same performance as `main` (maybe a small bit faster).  In Descent 2, I'm looking at about 60fps on your branch and 200fps with busy wait.  It looks like `std::barrier` is using much the same method of waiting as a conditional variable (lots of syscalls to `futex_wait`)

perf

>  I'm firmly of the opinion we, the core team, should have zero consideration for PPC—it's a dead platform.

Big endian PPC that was used on old Macs is dead.  Little endian PPC is still around and is supported on Debian 12 and Fedora 40.  It is a very niche platform so I'm not against dropping the runner from CI if it becomes problematic.  I think in this case though, it's a problem with Debian 11 and nothing to do with PPC itself.  #3950 updates the PPC runner from Debian 11 to Debian 12 and likely makes this a non-issue.

kklobe:
> I just tested that branch.  Unfortunately, I do not have the same results.  On my system, barrier is about the same performance as `main` (maybe a small bit faster).  In Descent 2, I'm looking at about 60fps on your branch and 200fps with busy wait.  It looks like `std::barrier` is using much the same method of waiting as a conditional variable (lots of syscalls to `futex_wait`)
> 

Bummer. Thanks for giving it a spin.

src/misc/cross.cpp
weirddan455: I'll leave this as is for now.  I'm not seeing an easy way to identify E cores without comparing frequencies or cache information.  I don't feel confident in such code working without me being able to test on real hardware.  This can always be adjusted later.

src/hardware/voodoo.cpp
```diff
@@ -4467,15 +4450,22 @@ static void triangle_worker_work(const triangle_worker& tworker,
...
 {
-	triangle_worker& tworker = v->tworker;
-	for (const int32_t tnum = p; tworker.threads_active;) {
-		tworker.sembegin[tnum].wait();
-		if (tworker.threads_active) {
-			triangle_worker_work(tworker, tnum, tnum + 1);
+	int i;
+	while ((i = tworker.worker_index.load()) < tworker.num_workers) {
+		if (tworker.worker_index.compare_exchange_weak(i, i + 1)) {
+			triangle_worker_work(tworker, i, i + 1);
```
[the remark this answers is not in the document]
weirddan455: Changed

src/misc/cross.cpp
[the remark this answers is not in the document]
weirddan455: Shrunk to 100KB.  `MaxCpus` isn't really helpful without knowing an exact size of the struct per CPU (and the size might even vary from system to system).  This will probably be enough for 1000 cores or so though.

src/misc/cross.cpp
weirddan455: Fixed

src/misc/cross.cpp
weirddan455: Sounds good.  I was already using the correct ones for Mac and Windows.  Changed the Linux macro as you noted.

src/misc/cross.cpp
weirddan455: I think I'll stick to the Ex version.  It works and it's more flexible which we may end up needing if we want to identify E cores in the future.

Good example though.  I added a missing error check thanks to that.

interloper98:
@kklobe , on the Pi4 barriers get close to busy wait: 5.2 fps versus 5.5 fps. 

On my Ryzen, barriers are roughly on par with `main`, which is fine for me because I have plenty of FPS already.

> about 80-85% of the speed benefit of the busywait with less than 1/2 the CPU utilization,

With barriers and `main`, my cores are between 10% to 20% used, and the die runs at around ~58 °C and fan RPM is essentially idle.

With busy wait, even when I limit `voodoo_threads = 4` (half my physical cores),  those four cores are running at 100% usage, 5.1 GHz with the die hitting ~80+ °C steady state, and the fan RPM is working harder to keep temps under control.

I think that's the dilemma here: if users limit the thread count, then FPS will be no better than barriers, but quite a bit higher CPU and power usage.

I'll be very curious if @FeralChild64 can re-test his extended run (after 15 minutes?) and see if barriers can deliver more fps than busy wait.

The flip side is this PR rocks for serious Voodoo gaming: If I want the smoothest FPS in Quake 2, then I'll bite the bullet and give it all 8 physical cores. Light 'em up, baby! :fire: 

I'm not sure if it would be easy to offer this as a conf option; it would be really nice to flip between barriers and busy wait, depending on the game and my mood. :grin:

FeralChild64:
I have tested the barrier implementation, and it gave me similar results to the _main_ branch - same experience @weirddan455 had.

I really wouldn’t spend too much time trying to find the best values. My air cooling is clearly not up to the task here, someone with water cooling will definitely have different experiences. Different scheduling options in Linux kernels might give you different results (that’s why I have withdrawn my GameMode patch after testing it with this PR, it was worsening the performance). The AMD X3D CPU might behave differently than a regular one. Intel P-cores / E-cores might give different results than Apple Silicon cores. 

There is just too many variables to test everything. I would vote to just merge this PR once all the review remarks are addressed - as this is a clear improvement - and leave possible future enhancements after 0.82.0 is released.

johnnovak:
> There is just too many variables to test everything. I would vote to just merge this PR once all the review remarks are addressed - as this is a clear improvement - and leave possible future enhancements after 0.82.0 is released.

+1 from me on that one.

weirddan455:
Well I've addressed all the review comments so far.  Let me know if anyone has anything else.

@johnnovak Do you want this to go in 0.82.0?  I think it pairs nicely with @interloper98 's previous PR that went in to give a nice boost to Voodoo performance but if you want to hold off I'm fine with that too.

johnnovak:
> Well I've addressed all the review comments so far. Let me know if anyone has anything else.
> 
> @johnnovak Do you want this to go in 0.82.0? I think it pairs nicely with @interloper98 's previous PR that went in to give a nice boost to Voodoo performance but if you want to hold off I'm fine with that too.

I think it's fine to include it in 0.82.0. You guys have collectively tested the hell out of this 😉

FeralChild64 (approved):

johnnovak (approved):
Really nice improvement @weirddan455. Light up those cores 😎 🔥 🤘🏻

kklobe:
My only real argument was that "auto" commandeering up to 8 threads is a bit aggressive as a default. On my M3, with 12 cores, 8 threads actually gives worse performance than 4, while eating up significantly more battery. Yes, there are many variables, but even with 4 on this branch I get a very nice improvement over main (and on my Windows machine), which would probably be a great default for most users; benchmark wankery to get 300fps vs 220 could always hand-tune 😁

weirddan455:
>  On my M3, with 12 cores, 8 threads actually gives worse performance than 4, while eating up significantly more battery.

Are any of those 12 cores E cores or "virtual" hyperthreaded cores?  You're the only one that's tested with that CPU.  I could make the change @johnnovak found of querying for only P cores using `hw.perflevel0.physicalcpu` for Macs (and leave Linux and Windows as-is since they don't have an easy API for that that I know of).

Or if you have another suggestion, let me know.  Everyone else has reported better performance going from 4 to 8 threads but it's not too surprising that you get different results using a completely different OS and CPU architecture.

johnnovak:
> benchmark wankery to get 300fps vs 220 could always hand-tune 😁

Yeah, so with vsync disabled at the emulation level, we're burning CPU usage for no good reason. If I'm correct about this, this is going on:

- vsync is disabled in-game in the Voodoo settings (I assume every game must have a vsync on/off setting somewhere)
- We're maxing out those 8 threads to render at 200+ FPS is we're able
- But at the _VGA emulation level_ we're in a 640x480 @ 60 Hz mode! So we're just _sampling_ that 200 FPS stream at 60 Hz, effectively.
- Well, then at the _host level_ the physical monitor might be running at 50 Hz. So we're doing another round of sampling the 60 FPS _emulated VGA output_ at 50 Hz (unlikely, most people run their monitors at 60 Hz+, just for illustrative purposes).

I've noticed the vertical sync handling paths are commented out in `voodoo.cpp`. So I assume the original code _did_ support vsync enabled at the Voodoo level, but someone got rid of that for... reasons? 😆 

With vsync enabled at the Voodoo level, no matter how many cores we have dedicated to the Voodoo, we'd never go above 60 Hz in 640x480 @ 60 Hz mode, which is where we want to be IMO. The "oversized" worker pool of 8 threads would guarantee a stable 60 FPS 99% of the time (hopefully 🤞🏻), but we'd never go into silly 100-200+ FPS territory when the host monitor is running at 60 Hz anyway...

(Of course, vsync disabled at the Voodoo level and a VRR monitor is a valid use case. But many people are still on non-VRR, fixed refresh rate monitors).

weirddan455:
>  Can anyone test some games with vsync enabled/disabled at the game config level? Do games offer these options?

Descent 2 and Quake 2 do not have vsync options that I could find.

>  Or maybe the Voodoo is hard-vsynced by default? I don't know! 

I don't know either.  At least from the in-game fps counters, our emulation does not appear to be utilizing vsync.  I'm not sure what real hardware does.  Probably we're getting much higher fps than a real Voodoo 1/2 ever could would be my guess :laughing:

I never owned one of the early 3dfx cards.  I went from a Riva 128 on a Pentium 2 to a Voodoo 4 on a Pentium 3.  Both running Windows 98.  I never had the OG experience of Voodoo 1 on DOS.  The DOS computers my family had before that didn't have any fancy 3D accelerator cards.

>  We're maxing out those 8 threads to render at 200+ FPS (if we're able).

The way the Voodoo emulation code is architected as-of now, it will max out the worker cores no matter what.  Ideal solution would be to split the work into larger chunks but that would require a deeper understanding of the Voodoo emulation code than I have :smile: 

The waiting bottleneck happens because these chunks happen every 100 microseconds or so.  This means a lot of syscalls trying to wait for a very small amount of time.

 >  I've noticed the vertical sync handling paths are commented out in voodoo.cpp. So I assume the original code did support vsync at the Voodoo level, but someone got rid of that for... reasons?

OK, I haven't even gone that deep in the Voodoo code.  Maybe we could re-enable that and get some power efficiency back.

@kklobe I'm going to go ahead and merge this.  There's clearly room for improvement still with respect to v-sync or larger chunk batching so the threads can sleep efficiently.  You could also do something smarter with the `auto` setting.  If Macs are getting worse performance with a higher thread count, maybe lower that for the M3 hardware.  You could even do something like detect if the system is currently running on battery and, if so, lower the thread count.

Anyway, IMO, this is a clear performance gain for the average user but I'm not saying this is perfect by any means.  Future PRs are very welcome :smile:

PR #3960, opened 2024-09-26
Let GUS use its documented IRQ, DMA, and base addresses

### LINKED ISSUE #3957: Fasttracker II crashes with `gusdma = 7`

_No response_

macOS (Apple Silicon)

_No response_

### Relevant hardware info

_No response_

Fast Tracker II crashes consistently as soon as Ultrasound is chosen as output device. If it was already chosen (because the files came from a real system or were used in a different emulator) it will crash on launch.
I tried Fast Tracker II version 2.08 (latest official) as well as the unofficial updates by 8bitbubsy. They all act in the same way. Of course I checked the same files in another emulator (DOSBox-X) and they are an exact copy of my real system, where it also works as expected.

Because I was just kindly instructed to use Fast Tracker II to play XM files by one of the developers, I wonder if this crash is specific to the Apple Silicon or macOS build.

Explain how to reproduce
1. Download the latest official version of Fast Tracker II [here](https://files.scene.org/view/resources/music/trackers/ft208.zip), or the latest unofficial version [here](https://16-bits.org/etc/FT213.ZIP)
2. Start the program (FT2.EXE)
3. It is possible the program is configured for Sound Blaster or "None" out of the box. Click on the Config button and chose Ultrasound.
4. It crashes.

_No response_

### Your configuration

```ini
https://gist.github.com/digistorm79/3841c9fe01b44a97a6b9f17d61691df3
```

### Provide a Log

```text
https://gist.github.com/digistorm79/5fcd805839b6cf88db634aab1a714031
```

- [X] Yes, I agree.

interloper98:
# Description

Previously some IRQ's weren't available, like IRQ 2 and 15, and they weren't defined in the order that was expected via the IO port 2xB address selector.

(more detail in the commit message and comments in the code).

## Related issues

Fixes #3957

# Release notes

The Gravis Ultrasound's configuration options have been changed to reflect the actual options available on the GUS card:
 - Base addresses now range from 210 through 260,
 - IRQs 2 and 15 have been added, and
 - DMA 0 has been dropped.

This also fixes a crash in Fast Tracker II when GUS is configured to use IRQ 7.

# Manual testing

Tested mod players (capamod, FT 208, FT 213, impulse tracker), 15+ games, and some that ultramid.

# Checklist

I have:

### DIFF
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -216,6 +216,34 @@ union ResetRegister {
 	bit_view<2, 1> are_irqs_enabled;
 };
 
+// Mix Control Register (2X0), section 2.13, page 28, of the UltraSound Software
+// Development Kit (SDK) Version 2.22
+union MixControlRegister {
+	// Default state: lines disabled, latches enabled
+	uint8_t data = 0b0000'1011;
+
+	bit_view<0, 1> line_in_disabled;
+	bit_view<1, 1> line_out_disabled;
+	bit_view<2, 1> microphone_enabled;
+	bit_view<3, 1> latches_enabled;
+	bit_view<4, 1> channel1_irq_combined_with_channel2;
+	bit_view<5, 1> midi_loopback_enabled;
+	bit_view<6, 1> irq_control_selected;
+	// bit 7 reserved
+};
+
+// IRQ and DMA Control Select Register (2XB), section 2.14-2.15, page 28-30, of
+// the UltraSound Software Development Kit (SDK) Version 2.22
+union AddressSelectRegister {
+	uint8_t data = 0;
+	bit_view<0, 3> channel1_selector;
+	bit_view<3, 3> channel2_selector;
+	// Note: If the channels are sharing, then channel 2's IRQ selector must
+	// be set to 0 and bit 6 must be enabled.
+	bit_view<6, 1> channel2_combined_with_channel1;
+	// bit 7 reserved
+};
+
 // The Gravis UltraSound GF1 DSP (classic)
 // This class:
 //   - Registers, receives, and responds to port address inputs, which are used
@@ -283,7 +311,8 @@ private:
 	void StopAndReset() noexcept;
 
 	void RenderUpToNow();
-	void UpdateDmaAddress(uint8_t new_address);
+	void UpdatePlaybackDmaAddress(const uint8_t new_address);
+	void UpdateRecordingDmaAddress(const uint8_t new_address);
 	void UpdateWaveMsw(int32_t& addr) const noexcept;
 	void UpdateWaveLsw(int32_t& addr) const noexcept;
 	void WriteToPort(io_port_t port, io_val_t value, io_width_t width);
@@ -329,7 +358,6 @@ private:
 	uint8_t selected_register = 0;
 
 	// Control states
-	uint8_t mix_ctrl    = 0x0b; // latches enabled, LINEs disabled
 	uint8_t sample_ctrl = 0;
 	uint8_t timer_ctrl  = 0;
 
@@ -348,6 +376,7 @@ private:
 	uint8_t irq_status = 0;
 
 	ResetRegister reset_register = {};
+	MixControlRegister mix_control_register = {};
 
 	bool irq_previously_interrupted = false;
 	bool should_change_irq_dma      = false;
@@ -695,7 +724,8 @@ Gus::Gus(const io_port_t port_pref, const uint8_t dma_pref, const uint8_t irq_pr
 
 	ms_per_render = MillisInSecond / audio_channel->GetSampleRate();
 
-	UpdateDmaAddress(dma_pref);
+	UpdatePlaybackDmaAddress(dma_pref);
+	UpdateRecordingDmaAddress(dma_pref);
 
 	// Populate the volume, pan, and auto-exec arrays
 	PopulateVolScalars();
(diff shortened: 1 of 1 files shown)


interloper98:
digistorm79, feel free to give this branch a try.

src/hardware/gus.cpp
```diff
@@ -1771,12 +1812,12 @@ void init_gus_dosbox_settings(Section_prop& secprop)
...
 	assert(int_prop);
-	int_prop->Set_values({"3", "5", "7", "9", "10", "11", "12"});
+	int_prop->Set_values({"2", "3", "5", "7", "11", "12", "15"});
 	int_prop->Set_help("The IRQ number of the Gravis UltraSound (5 by default).");
 
 	int_prop = secprop.Add_int("gusdma", when_idle, 3);
 	assert(int_prop);
-	int_prop->Set_values({"0", "1", "3", "5", "6", "7"});
+	int_prop->Set_values({"1", "3", "5", "6", "7"});
 	int_prop->Set_help("The DMA channel of the Gravis UltraSound (3 by default).");
```
johnnovak, line 1821: There must be some magical link between you and I @interloper98 because I was considering doing the same enhancements... 🥳 🧙🏻‍♂️🔮 🪄 😂   

..._OR_ you've also just seen the list of available options in the sound setup utility of the demo **Inside by CNCD** 😉 

Would fixing up `gusbase` be worthwhile too? Our current values:

image

Values from the sound setup utility:

image

The IRQ and DMA settings are now in-line with the utility:

image

image

johnnovak:
> Note that we use IRQ2 in GUS's public interface (conf, environment, and the IO port 2xB address selector lookup
tables) because that's what the documentation describes and more critically, it's what games and applications expect and
use via the IO port.

I'm confused by this comment. The default for `gusirq` is 5 in Staging, and the factory default is 11.

Base address 220, IRQ 11, and DMA 1 are the stock GUS settings, but DOSBox has historically been altering them so the GUS can coexist with the SB without resource conflicts.

Could you clarify this IRQ 2 business please?

src/hardware/gus.cpp
```diff
@@ -621,12 +617,34 @@ void Voice::WriteWaveRate(uint16_t val) noexcept
...
 
+// We use IRQ2 in GUS's public API (conf, environment, and the IO port 2xB
+// address selector lookup tables) because that's what the documentation
+// describes and more critically, it's what games and applications expect and
+// use via the IO port.
+//
+// However we convert IRQ2 to IRQ9 internally because that's how real hardware
+// worked (IBM reserved IRQ2 for cascading to the second controller where it
+// becomes IRQ9), so we translate IRQ2 to 9 and vice-versa on this API
+// boundaries. This is also what DOSbox expects: it uses IRQ9 instead of IRQ2.
```
johnnovak, line 628: Very nice explanation! 🎖️

src/hardware/gus.cpp
```diff
@@ -1443,26 +1468,42 @@ void Gus::WriteToPort(io_port_t port, io_val_t value, io_width_t width)
...
-		if (mix_ctrl & 0x40) {
-			// IRQ configuration, only use low bits for irq 1
-			const auto i        = val & 7;
-			const auto& address = irq_addresses.at(i);
-			if (address) {
-				irq1 = address;
-			}
+		if (should_change_irq_dma) {
+
+			//  The write to 2XB MUST occur as the NEXT IOW or else
```
johnnovak, line 1473: Maybe spell out "I/O write".

src/hardware/gus.cpp
```diff
@@ -1443,26 +1468,42 @@ void Gus::WriteToPort(io_port_t port, io_val_t value, io_width_t width)
...
+			const auto selector = static_cast<uint8_t>(val & 0b111);
+
+			// When Bit 6 (Control Register Select) is set to a 1,
+			// the next IO write to 2XB will be to the IRQ control
+			// latches. When this is set to a 0, the next IO write
+			// to 2XB will be to the DMA channel latches. UltraSound
+			// Software Development Kit (SDK), Section 2.13.
+			//
+			const auto is_irq_selected = bit::is(mix_ctrl,
+			                                     bit::literals::b6);
```
johnnovak, line 1492: Next level: `bit_view` register definitions 😉 😉 😆

johnnovak (approved):
Tried a few wild DMA & IRQ settings and couldn't get FT2 to crash, so this looks good 👍🏻 Nice work @interloper98, merge when you're ready 🎉

src/hardware/gus.cpp
interloper98: Ahh yes; even better. 
I see that `mix_ctrl` isn't used that much, so this will fit fine with this PR (without treading into wide-scale refactor territory).

interloper98: `bit_view`'d the mix control and address selector regs :+1: that's a lot more self documenting now :-)

interloper98:
> remove the last paragraph from the commit message altogether as it's ambiguous 

Removed - thanks for that suggestion.

src/hardware/gus.cpp
interloper98: Updated the base addresses.

src/hardware/gus.cpp
johnnovak: > `bit_view`'d the mix control and address selector regs 👍 that's a lot more self documenting now :-)

Noice 😎

johnnovak:
Very nice @interloper98 , looks perfect now 👍🏻

interloper98:
> Also, could you add the FT2 crash fix to the release notes draft please?

(Added 👍)

johnnovak:
@interloper98  Just noticed that the "no DMA" (DMA 0) option was dropped from the conf setting:

image

Not too sure if there is any point in setting DMA 0, so I'm not hung up about this at all, just I thought I'd mention.

interloper98:
> Just noticed that the "no DMA" (DMA 0) option was dropped from the conf setting

> Not too sure if there is any point in setting DMA 0

I tried `gusdma = 0` with a couple programs: Star Control 2 blows up moments after activating the GUS channels in  `PIC_RunQueue()` and Fast Tracker II crashes in the GUS with it attempt to `PerformDmaTransfer()`.

So it seems to be an invalid setting. At least from the SDK, applications know that "DMA 0" actually disables DMA on the card (and a DMA-less GUS can't move samples in or out of the card).

I guess on real hardware, setting DMA 0 "parks" the card without it consuming any DMA channels (but what happens if you still try to use it?). In DOSBox, we can just leave `gus = off` to avoid resource conflicts.

So from the conf perspective, we only allow nominal/active config values, similar to program screenshoted:

image


But why do the DMA and IRQ arrays have a starting 0? 

```c++
constexpr std::array<uint8_t, 8> IrqAddresses = {0, 2, 5, 3, 7, 11, 12, 15};

constexpr std::array<uint8_t, 6> DmaAddresses = {0, 1, 3, 5, 6, 7};
```

They are there to let the channel selector's index into them 1-to-1 per the docs:

out

So for example, if the user sets `ULTRASND` with DMA 5, then the program or game will set the selector to 3 and pass it to 2xB, which will hit `DmaAddresses[3] = 5`.

Without the 0, the selector indexes would be off by one (same with the IRQ selector indexes).

Using DMA 0 (and IRQ 0, too) is prevented both at the conf level and in the selector code - we ignore the requests for index 0 because they should be off limits.

johnnovak:
Thanks for the explanation @interloper98, it makes perfect sense, and I agree with your conclusion -- we should cull config values only useful on hardware but not in Staging. Cheers 🍻 😎

PR #3964, opened 2024-09-30
Translation (format string) fixes across several languages

FeralChild64:
# Description

The work-in-progress string sanitizer tool I am developing for 0.83.0 has found several problems with outdated format strings across many translations.

This PR fixes most of the problems (including all that could have caused a crash) - with the exception of Dutch (nl) translation, which is handled in a separate PR, https://github.com/dosbox-staging/dosbox-staging/pull/3967

# Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3965

# Manual testing

- Run code from branch `https://github.com/dosbox-staging/dosbox-staging/tree/fc/messages-1` using updated `*.lng` files from this PR, using various translations (`language = ...` config file options) - there are still some problems detected by the sanitizer code, but they won't cause a crash.
- executed `config -h integer_scaling` command when using different translations
- using Polish (pl) and  French (fr) translations configured `machine = cga` and `startup_verbosity = high`, checked that startup banner looks sane

# Checklist

I have:


contrib/resources/translations/fr.lng
```diff
@@ -1609,7 +1609,7 @@ Il n'est possible d'utiliser SUBST que sur les lecteurs locaux.
 :SHELL_STARTUP_CGA
 ║ DOSBox supporte le mode composite CGA.                                     ║
 ║ Utilisez [31mF12[37m pour régler la sortiecomposite sur ON, OFF, ou AUTO (défaut). ║
-║ [31m(%s+)F11[37m change la teinte; [31m%s+%s+F11[37m sélectionne type de modèle CGA.%s   ║
+║ [color=light-red]F10[color=white] selects the CGA settings to change and [color=light-red](%s+)F11[color=white] changes it.           ║
```
weirddan455, line 1612: The french translation is saying something about a Ctrl+Alt+F11 hotkey.  That's what the extra strings are for: "Ctrl" and "Alt" plus an empty string that appears unused even here.

I can't confirm what Ctrl+Alt+F11 is supposed to do.  On Linux, it flips me to an empty virtual console :laughing: 

I'd recommend removing the extra unused string parameter from the caller.

FeralChild64: OK, removed - looks like you are right.

weirddan455 (approved):
I don't speak these languages to review the translations themselves but looks fine to me otherwise.

interloper98 (approved):
Same comment at @weirddan455.
Catching string format problems are critically important; I remember printf in general will blow up instantly at runtime if there are mismatches.

johnnovak:
> Same comment at @weirddan455. Catching string format problems are critically important; I remember printf in general will blow up instantly at runtime if there are quantity and type size mismatches.

Yeah, and before writing your tool @FeralChild64, I'd highly suggest to look into ready-to-use, battle-tested open-source localisation solutions. Surely, yhis must be a solved problem in the C++ world, and perhaphs the extra validation tooling already exists -- we might just need to reach our hands for it.

At least it would be nice if you conducted a research first and posted a brief overview of the localisation library landscape, then you could make a better judgement if writing a custom tool is a worthwhile investment of your time.

It might be; after all I did write my own GUI library, but it's good to check for existing tools first (apologies if you've already done all this).

Otherwise, `std::vformat` seems to be a huge improvement in safety and placeholder options compared to the old C functions:

https://en.cppreference.com/w/cpp/utility/format/vformat

johnnovak (approved):
I trust you've tested this—I'm definitely not going to test the localisations myself, neither do I speak any of these languages 😆 Nice fix @FeralChild64 🚀

FeralChild64:
Among these languages, I only speak Polish - for the remaining languages I only did either trivial changes, or changes which were necessary to prevent possible crashes.

Regarding existing localization tools - I am only aware about _gettext_ and clones, but it is not string-ID based, so introducing it would require significant effort to introduce into DOSBox (including manual conversion of all the translated strings). Plus, it does not support `printf` or `std::format` format strings - it has it’s own syntax.

Another problem is that we have a custom markup syntax for creating ANSI escape sequences, plus we support like 100 different code pages, precisely matching the FreeDOS ones - in a way mostly transparent to the code which uses the translated strings.
No off-the-shelf tool is going to provide these, we would need to create like additional layer between our code and _gettext_, which would be compatible with _gettext_ code scanner - which means even more effort.

Writing the sanitizer is the least effort solution, believe me - actually, the code is already almost finished, save for some cleanups.

What we might do is to try to switch the format of our translation files to be compatible with _gettext_ compatible GUI tools - but this is a task for the future, right now I want to harden the code against translation-originated crashes (and detect badly outdated translations) and prepare it for runtime translation switching - including fallback to English if the code page does not support Cyrillic/Greek/etc. alphabet, required by the selected translation.

johnnovak:
> Writing the sanitizer is the least effort solution, believe me - actually, the code is already almost finished, save for some cleanups.

Sure, and you're the expert here @FeralChild64—I was just wondering 😄 Thanks for the detailed explanation.

You must have put a lot of research this into already, and I wasn't really aware our support for different code pages complicates things, but yeah it makes sense. Seems like incremental enhancements to our existing localisation layer is the correct approach then, so good luck then 😆 

You might still wanna look into `std:format`, `std::vformat` and friends as they are more safer than what we use now, AFAIK (you can't make them crash with mismatching/invalid format strings and arguments, or at least it much harder to make them misbehave, that's my understanding). But those would be trivial changes.

interloper98:
The `#include <format>` is fast, too (using `format` but `format_to` using iterators is even faster!)

[benchmark](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0645r10.html#Benchmarks) compiled with clang 18 on my Pi4:

```
pi@raspberrypi:~/src/string-format $ ./clang-main 
2024-10-01T15:28:26-07:00
Running ./clang-main
Run on (4 X 2200 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 48 KiB (x4)
  L2 Unified 1024 KiB (x1)
Load Average: 0.46, 0.25, 0.10
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** Library was built as DEBUG. Timings may be affected.
--------------------------------------------------------
Benchmark              Time             CPU   Iterations
--------------------------------------------------------
sprintf          1817830 ns      1817830 ns          322
ostringstream    7138442 ns      7137618 ns           94
to_string         360076 ns       360012 ns         1941
format            711595 ns       711545 ns          977
format_to         339423 ns       339284 ns         2070
```

With gcc 12:

```
pi@raspberrypi:~/src/string-format $ ./gcc-main   
2024-10-01T15:28:59-07:00
Running ./gcc-main
Run on (4 X 2200 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 48 KiB (x4)
  L2 Unified 1024 KiB (x1)
Load Average: 0.30, 0.24, 0.10
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** Library was built as DEBUG. Timings may be affected.
--------------------------------------------------------
Benchmark              Time             CPU   Iterations
--------------------------------------------------------
sprintf          1679732 ns      1679566 ns          368
ostringstream    7172792 ns      7160811 ns           95
to_string         359940 ns       359916 ns         1948
format            607906 ns       607877 ns         1124
format_to         276677 ns       276648 ns         2529
```

`#include <format>` is available on gcc 12 (maybe even gcc 11?), LLVM's clang 14, and Visual Studio 2019 v16.10+. macOS's doesn't support `<format>` until clang/xcode 15.3, and looks like our CI machines are a couple version behind.

So we'd need some fallback replacement like I used for the pooled memory stuff (`std::pmr`), because the CI macOS machines also didn't support it.

Maybe a couple years away - or we could wrap the `fmt` lib for macOS until CI is up to speed.

PR #3968, opened 2024-10-01
Auto switch shader when starting a Voodoo game

weirddan455:
# Description

I noticed when starting a Voodoo game, the display was not filling the window.  Resizing the window fixed the issue.  You can see it at about the 1:00 mark on this video @interloper98 posted:  https://github.com/dosbox-staging/dosbox-staging/pull/3935#issuecomment-2357425531

It turns out the problem was the shader was not switching when initially starting a Voodoo game (but it does on window resize).  Added a function call to `RENDER_MaybeAutoSwitchShader` to resolve this.  This code matches what `VGA_SetupDrawing` does for non-Voodoo games.

# Release notes

Fixed a bug where the wrong shader could be loaded in Voodoo games.

# Manual testing

Start a Voodoo game and observe that the shader now correctly auto switches.

# Checklist

I have:

### DIFF
--- a/src/hardware/voodoo.cpp
+++ b/src/hardware/voodoo.cpp
@@ -7496,6 +7496,9 @@ static void Voodoo_UpdateScreen()
 			const auto frames_per_second = static_cast<float>(
 			        1000.0 / v->draw.frame_period_ms);
 
+			constexpr auto reinit_render = false;
+			RENDER_MaybeAutoSwitchShader(GFX_GetCanvasSizeInPixels(), video_mode, reinit_render);
+
 			RENDER_SetSize(image_info, frames_per_second);
 		}
 


interloper98 (approved):
Nice quick fix, @weirddan455 !  

Re-watching that video does have me wondering why such a small adjustment to the window size popped the texture size up edge to edge. 

Gets my vote for 0.82, if it can make it in!

johnnovak:
> Gets my vote for 0.82, if it can make it in!

Definitely! Nice fix @weirddan455 🎉 Yeah that auto shader switching code is a bit tricky 😄

johnnovak (approved):

weirddan455:
>  Re-watching that video does have me wondering why such a small adjustment to the window size popped the texture size up edge to edge.

It's because it was using whatever shader the command prompt was using.  On my system, it was the VGA CRT 1080p shader.  The black bars are from integer scaling.  When the window is that small, the auto shader switcher uses the sharp shader which doesn't need integer scaling.  I also tested it switches back to the CRT shader once you make the window larger or go full screen.

The auto shader switcher is supposed to re-evaluate shader choice whenever the window size or the DOS resolution changes.  The Voodoo code wasn't doing the latter which is what this PR fixes.

PR #3973, opened 2024-10-04
Fix ARM64 Windows builds

shermp:
# Description

The ARM64 Windows builds have been failing in CI recently.

After a brief investigation, I discovered that MSBuild was passing `--triplet=ARM64-windows` to VCPKG. A [recent change](https://github.com/microsoft/vcpkg-tool/pull/1474) in VCPKG made passing invalid triplets in this manner an error. 

Valid triplets consist of lowercase alphanumeric strings separated by '-'. The uppercase `ARM64` in the triplet is therefore not valid and treated as an error. The prior behavior seems to have implicitly corrected the triplet.

Fortunately, it is possible to explicitly set the triplet in the MSBuild project. Therefore, I set it to `arm64-windows` for all configurations using the `ARM64` platform.

Requesting a review from @interloper98 as the person who implemented Windows ARM64 support. If you know of a better way to handle this, please chime in!

# Manual testing

Changes were tested locally before pushing to a branch to test in CI.

# Checklist

I have:

### DIFF
--- a/tests/vs/tests.vcxproj
+++ b/tests/vs/tests.vcxproj
@@ -185,6 +185,12 @@
     <VcpkgEnableManifest>true</VcpkgEnableManifest>
     <VcpkgAutoLink>false</VcpkgAutoLink>
   </PropertyGroup>
+  <PropertyGroup Label="Vcpkg" Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
+    <VcpkgTriplet>arm64-windows</VcpkgTriplet>
+  </PropertyGroup>
+  <PropertyGroup Label="Vcpkg" Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
+    <VcpkgTriplet>arm64-windows</VcpkgTriplet>
+  </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <WarningLevel>Level3</WarningLevel>
--- a/src/libs/zmbv/zmbv.vcxproj
+++ b/src/libs/zmbv/zmbv.vcxproj
@@ -179,9 +179,13 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'" Label="Vcpkg">
     <VcpkgConfiguration>Debug</VcpkgConfiguration>
+    <VcpkgTriplet>arm64-windows</VcpkgTriplet>
   </PropertyGroup>
   <PropertyGroup Label="Vcpkg" Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <VcpkgConfiguration>Debug</VcpkgConfiguration>
+  </PropertyGroup>
+  <PropertyGroup Label="Vcpkg" Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
+    <VcpkgTriplet>arm64-windows</VcpkgTriplet>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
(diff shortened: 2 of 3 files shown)


tests/vs/tests.vcxproj
```diff
@@ -185,6 +185,12 @@
     <VcpkgEnableManifest>true</VcpkgEnableManifest>
     <VcpkgAutoLink>false</VcpkgAutoLink>
   </PropertyGroup>
+  <PropertyGroup Label="Vcpkg" Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
+    <VcpkgTriplet>arm64-windows</VcpkgTriplet>
+  </PropertyGroup>
```
interloper98, line 190: For the debug `PropertyGroup`, might also need:
```xml
<VcpkgConfiguration>Debug</VcpkgConfiguration>
```

interloper98 (approved):
Maybe this is a temporary bug, because the Win32 platform doesn't need a lower-cased `win32-windows` triplet explicitly specified.

At the time, Visual Studio simply generated those `PropertyGroup` as they were without the triplets specified.
 
Looks good though - nice fix, @shermp.

(Just one comment; might be a missing debug specifier in one of the groups).

tests/vs/tests.vcxproj
shermp: `<VcpkgConfiguration>` isn't set anywhere in the test project. It is set in the GUI though, so I dunno what's going on there. I don't think I'll make any changes here.

shermp:
> Nice fix, @shermp!
> 
> Maybe this is a temporary bug, because the Win32 platform doesn't need a lower-cased `win32-windows` triplet explicitly specified.
> 
> At the time when I added the ARM64 target, Visual Studio simply generated those `PropertyGroup`s as they were without the triplets specified.
> 
> (Just one comment; might be a missing debug specifier in one of the groups).

If I try `-p:Platform=Win32` it doesn't even work, fails with `The specified solution configuration "Release|Win32" is invalid.` `x86 works though, therefore the correct triplet is set.

I think this is a bug/oversight in MSBuild/VS. VCPKG was lenient before, so it was missed.

tests/vs/tests.vcxproj
interloper98: I just had noticed that in all of the **debug** property groups (across all the project files), that we explicitly specify the **debug** vcpkg configuration.

On the other hand, for the **release** property groups (across all files), we don't include the vcpkg configuration (perhaps because **release** is default?).

So this comment here just catches the only place where we're not specifying the debug configuration inside the debug property group. (So.. we're probably getting release libs; I dunno if we case though! Maybe not a big deal).

shermp: I dunno what's going on, as I said, it's set in the GUI.

image

interloper98: Ahh; thanks - I didn't make that connection!
I guess that setting is just somewhere else at a higher level in the XML 😌

weirddan455:
Why are our build tools being updated behind our back and breaking stuff?  That was the reason to switch from MSYS2 to MSVC in the first place.  There's probably still less breakage with MSVC but I'm not of fan of this.  Ideally, depandabot would put in a PR to update a version string in the runner files and then we can easily track when an update causes problems.

Not sure if there's anything we can do about this or not.  Maybe GitHub just updates their runners whenever they feel like it.

shermp:
Yeah, Github update their runners weekly I believe.

Honestly, I'm more annoyed at VCPKG than anything else. They either never tested this with ARM64 builds, or we had something wrong setup in our projects.

johnnovak:
> Why are our build tools being updated behind our back and breaking stuff? That was the reason to switch from MSYS2 to MSVC in the first place. There's probably still less breakage with MSVC but I'm not of fan of this.

Me neither, but mistakes happen 😉 Vcpkg is just an objectively better conceptual approach (you can pin version baselines, hey!), and I think this is the first such problem we've had with it.

Compared to the hoops we had to jump through to pin version with MSYS2, plus all the breakages, it's like a dream, man 😎 

Plus we're planning to use it on all three platforms (vs maintaining platform-specific toolsets, which triples the maintenance work and the required knoiwledge).

So I'm personally not annoyed; this is an easy oversight to make, but yeah, it should've been caught by testing.

PR #3979, opened 2024-10-08
Close modem server socket before dialing

NicknineTheEagle:
# Description

This is a small fix where modem server socket listening for incoming calls is now closed before modem dials a number. Without this, you can dial yourself (something not physically possible with real phones) which results in really weird shit in games. If dialing fails then `EnterIdleState` will re-open the server socket.

# Release notes

Fixed a modem emulation oversight where the user could dial themselves.

# Manual testing

Tested with Telemate by trying to dial my own Dosbox instance. Connection properly fails since Dosbox is no longer listening for incoming connections.

# Checklist

I have:

### DIFF
--- a/src/hardware/serialport/softmodem.cpp
+++ b/src/hardware/serialport/softmodem.cpp
@@ -310,7 +310,7 @@ bool CSerialModem::Dial(const char * host) {
 	clientsocket.reset(NETClientSocket::NETClientFactory(socketType,
 	                                                     destination, port));
 	if (!clientsocket->isopen) {
-		clientsocket.reset(nullptr);
+		clientsocket.reset();
 		LOG_MSG("SERIAL: Port %" PRIu8 " failed to connect.", GetPortNumber());
 		SendRes(ResNOCARRIER);
 		EnterIdleState();
@@ -378,7 +378,7 @@ void CSerialModem::Reset(){
 	plusinc = 0;
 	oldDTRstate = getDTR();
 	dtrmode = 2;
-	clientsocket.reset(nullptr);
+	clientsocket.reset();
 
 	memset(&reg,0,sizeof(reg));
 	reg[MREG_AUTOANSWER_COUNT] = 0;  // no autoanswer
@@ -401,8 +401,8 @@ void CSerialModem::EnterIdleState(){
 	ringing = false;
 	dtrofftimer = -1;
 	warmup_remain_ticks = 0;
-	clientsocket.reset(nullptr);
-	waitingclientsocket.reset(nullptr);
+	clientsocket.reset();
+	waitingclientsocket.reset();
 
 	// get rid of everything
 	if (serversocket) {
@@ -412,7 +412,7 @@ void CSerialModem::EnterIdleState(){
 		}
 	}
 	if (listenport) {
-		serversocket.reset(nullptr);
+		serversocket.reset();
 		serversocket.reset(NETServerSocket::NETServerFactory(socketType,
 		                                                     listenport));
 		if (!serversocket->isopen) {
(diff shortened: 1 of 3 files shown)


src/hardware/serialport/softmodem.cpp
```diff
@@ -288,6 +288,18 @@ void CSerialModem::SendRes(const ResTypes response) {
...
+	waitingclientsocket.reset(nullptr);
+
+	if (serversocket) {
+		waitingclientsocket.reset(serversocket->Accept());
+		while (waitingclientsocket) {
+			waitingclientsocket.reset(serversocket->Accept());
+		}
+
+		serversocket.reset(nullptr);
+	}
```
interloper98, line 301: The function `CSerialModem::EnterIdleState()` contains the same code that discards all the server connections:

```c++
waitingclientsocket.reset(nullptr);
if (serversocket) {
	waitingclientsocket.reset(serversocket->Accept());
	while (waitingclientsocket) {
		waitingclientsocket.reset(serversocket->Accept());
	}
	...
```

Suggest de-duping to something like:

```c++
void CSerialModem::CloseServerSocket() {
    waitingclientsocket.reset();

    if (serversocket) {
        while (auto accepted = serversocket->Accept()) {
            delete accepted;
        }
        serversocket.reset();
    }
}
```

johnnovak: Seem reasonable to reuse that code/logic and not copy/paste it into two places, yeah.

NicknineTheEagle:
```
..\src\hardware\serialport\misc_util.cpp(111,25): warning : call to pure virtual member function 'Accept' has undefined behavior; overrides of 'Accept' in subclasses are not available in the destructor of 'NETServerSocket' [-Wcall-to-pure-virtual-from-ctor-dtor]
```
Seem that this won't be that simple.

johnnovak: Yeah, like @interloper98 said.

src/hardware/serialport/softmodem.cpp
```diff
@@ -288,6 +288,13 @@ void CSerialModem::SendRes(const ResTypes response) {
 }
 
 bool CSerialModem::Dial(const char * host) {
+	// Close the server socket before dialing
+	waitingclientsocket.reset(nullptr);
+	if (serversocket) {
+		serversocket->Close();
+		serversocket.reset(nullptr);
+	}
```
interloper98, line 296: Looking good :+1: 

One last suggestion (just style) from me is to use `.reset()` instead of with `nullptr`, because that's how all the other sources do it (I did a `git grep '.reset('` to show me all the instances)

johnnovak: Good point, after that it's good to go.

johnnovak:
@NicknineTheEagle @interloper98 I assume it's safe to include this in 0.82.0, right? @NicknineTheEagle I assume you have tested multiplayer with this change, and no regressions, right?

NicknineTheEagle:
> @NicknineTheEagle @interloper98 I assume it's safe to include this in 0.82.0, right?
> 
> @NicknineTheEagle I assume you have tested multiplayer with this change, and no regressions, right? Just we don't want to break multiplayer in eXoDOS.

I just did a quick test with Doom and Quake and haven't noticed connection issues.

src/hardware/serialport/softmodem.cpp
NicknineTheEagle: Every `reset` instance in that file does `reset(nullptr)`. Unless you want me to do refactoring in this file, I think it's best to leave it as is.

interloper98:
> I assume it's safe to include this in 0.82.0, right?

I think so too; should be no harm in actively closing queued client connections as a courtesy to those clients, where as before they would be left dangling to timeout.

src/hardware/serialport/softmodem.cpp
interloper98: Yup - I would say mass replace 'em if we're changing that code anyway (can be a separate commit` "Make socket resets consistent"` or something like that). 

@johnnovak's always suggested these type of consistency improvements in my PRs, might as well do it. It's a thumbs up from me :smile:

interloper98 (approved):

src/hardware/serialport/softmodem.cpp
johnnovak: @NicknineTheEagle Yeah, can you please mass-replace `reset(nullptr)` with `reset()` in a separate commit, thanks. They only occur in this file, nowhere else:

image

johnnovak (approved):
Great work @NicknineTheEagle 🚀

PR #3981, opened 2024-10-10
Sort SoundFonts in `MIXER /LISTMIDI` output & fix current SF2 matching

johnnovak:
# Description

This PR changes the following in the `MIXER /LISTMIDI` command's output _only_ (the resolution of the SoundFont file hasn't been changed at all):

- The list of SoundFonts is now sorted
- The current SoundFont is not moved to the top of the list anymore
- The highlighting of the current SoundFont file is performed in a case-insensitive manner on Windows and macOS (as they have case-preserving filesystems, in contrast to the strictly case-sensitive Linux filesystems that everybody seems to be still using)
 
# Release notes

The list of available SoundFont files in the `MIXER /LISTMIDI` command's output is now sorted. The highlighting of the current SoundFont file is done in a case-insensitive manner on Windows and macOS.

# Manual testing

Tested on macOS; should be applicable to Windows, too. Haven't tested on Linux but there should be no change in behaviour there.

## Current `main` behaviour

Problems:

- The SoundFont names are in a random unsorted order
- The highlighted item always appears at the top of the list (for no good reason)
- `creative_1mgm.sf2` (what `soundfont` is set to) and `Creative_1mgm.sf2`(the actual name on the filesystem) both appear in the list

image

## PR behaviour

- The SoundFonts are not alphabetically sorted by name
- The current SoundFont is highlighted in the A-Z sorted list (its position is not altered)
- `Creative_1mgm.sf2` is highlighted (despite we set `soundfont` to `creative_1mgm.sf2`, all lowercase)

image

# Checklist

I have:

### DIFF
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -241,7 +241,7 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 	Close();
 
 	FluidSynthSettingsPtr fluid_settings(new_fluid_settings(),
-	                                    delete_fluid_settings);
+	                                     delete_fluid_settings);
 	if (!fluid_settings) {
 		LOG_WARNING("FSYNTH: new_fluid_settings failed");
 		return false;
@@ -259,12 +259,10 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 	const auto sample_rate_hz = MIXER_GetSampleRate();
 	ms_per_audio_frame        = MillisInSecond / sample_rate_hz;
 
-	fluid_settings_setnum(fluid_settings.get(),
-	                      "synth.sample-rate",
-	                      sample_rate_hz);
+	fluid_settings_setnum(fluid_settings.get(), "synth.sample-rate", sample_rate_hz);
 
 	FluidSynthPtr fluid_synth(new_fluid_synth(fluid_settings.get()),
-	                         delete_fluid_synth);
+	                          delete_fluid_synth);
 	if (!fluid_synth) {
 		LOG_WARNING("FSYNTH: Failed to create the FluidSynth synthesizer.");
 		return false;
@@ -287,9 +285,10 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 	}
 
 	if (scale_by_percent < 1 || scale_by_percent > 800) {
-		LOG_WARNING("FSYNTH: Invalid volume scaling percentage: %d; "
-		            "must be between 1 and 800, defaulting to 100%%",
-		            scale_by_percent);
+		LOG_WARNING(
+		        "FSYNTH: Invalid volume scaling percentage: %d; "
+		        "must be between 1 and 800, defaulting to 100%%",
+		        scale_by_percent);
 		scale_by_percent = 100;
 	}
 	fluid_synth_set_gain(fluid_synth.get(),
@@ -321,13 +320,14 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 		// convert the string to a double
 		const auto val = atof(str_val.c_str());
 		if (val < min_val || val > max_val) {
-			LOG_WARNING("FSYNTH: Invalid %s setting (%s), needs to be between "
-			            "%.2f and %.2f: using default (%.2f)",
-			            name,
-			            str_val.c_str(),
-			            min_val,
-			            max_val,
-			            def_val);
+			LOG_WARNING(
+			        "FSYNTH: Invalid %s setting (%s), needs to be between "
+			        "%.2f and %.2f: using default (%.2f)",
+			        name,
+			        str_val.c_str(),
+			        min_val,
+			        max_val,
+			        def_val);
 			return def_val;
 		}
 		return val;
@@ -378,15 +378,17 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 			if (chorus[4] == "triangle") {
 				chorus_mod_wave = fluid_chorus_mod::FLUID_CHORUS_MOD_TRIANGLE;
 			} else if (chorus[4] != "sine") { // default is sine
-				LOG_WARNING("FSYNTH: Invalid chorus modulation wave type ('%s'), "
-				            "needs to be 'sine' or 'triangle'",
-				            chorus[4].c_str());
+				LOG_WARNING(
+				        "FSYNTH: Invalid chorus modulation wave type ('%s'), "
+				        "needs to be 'sine' or 'triangle'",
+				        chorus[4].c_str());
 			}
 
 		} else {
-			LOG_WARNING("FSYNTH: Invalid number of custom chorus settings (%d), "
-			            "should be five",
-			            static_cast<int>(chorus.size()));
+			LOG_WARNING(
+			        "FSYNTH: Invalid number of custom chorus settings (%d), "
+			        "should be five",
+			        static_cast<int>(chorus.size()));
 		}
 	}
 	// API accept an integer voice-count
@@ -426,9 +428,10 @@ bool MidiHandlerFluidsynth::Open([[maybe_unused]] const char* conf)
 			reverb_level = validate_setting(
 			        "reverb level", reverb[3], reverb_level, 0.0, 1.0);
 		} else {
-			LOG_WARNING("FSYNTH: Invalid number of custom reverb settings (%d), "
-			            "should be four",
-			            static_cast<int>(reverb.size()));
+			LOG_WARNING(
+			        "FSYNTH: Invalid number of custom reverb settings (%d), "
+			        "should be four",
+			        static_cast<int>(reverb.size()));
 		}
 	}
 
@@ -586,8 +589,9 @@ void MidiHandlerFluidsynth::Close()
 	MIXER_LockMixerThread();
 
 	if (had_underruns) {
-		LOG_WARNING("FSYNTH: Fix underruns by lowering CPU load, increasing "
-		            "your conf's prebuffer, or using a simpler SoundFont");
+		LOG_WARNING(
+		        "FSYNTH: Fix underruns by lowering CPU load, increasing "
+		        "your conf's prebuffer, or using a simpler SoundFont");
 		had_underruns = false;
 	}
 
@@ -884,9 +888,9 @@ MIDI_RC MidiHandlerFluidsynth::ListAll(Program* caller)
 
 		if (do_highlight) {
 			const auto output = format_str("%s* %s%s\n",
-			                                  green,
-			                                  line.c_str(),
-			                                  reset);
+			                               green,
+			                               line.c_str(),
+			                               reset);
 
 			caller->WriteOut(convert_ansi_markup(output).c_str());
 		} else {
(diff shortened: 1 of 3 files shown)


include/string_utils.h
```diff
@@ -217,6 +217,8 @@ inline bool is_empty(const char* str) noexcept
 // case-insensitive comparisons
 bool ciequals(const char a, const char b);
 
+bool case_insensitive_equals(const std::string& a, const std::string& b);
+
 // case-insensitive comparison for combinations of
 // const char *, const std::string&, and const string_view
 template <typename T1, typename T2>
```
interloper98, line 224: Maybe the new function isn't needed? Noticed the comment below is for insensitive string equality `iequal`, and used throughout the code base.

johnnovak: Thanks, I overlooked that function somehow 😄 

I ended up with a much better solution using [std::filesystem::canonical](https://en.cppreference.com/w/cpp/filesystem/canonical) (and [std::filesystem::equivalent](https://en.cppreference.com/w/cpp/filesystem/equivalent) is another option). These use the actual underlying filesystem for name resolution and comparisons. Changes incoming soon!

interloper98: Looks great; didn't know about these filesystem "equivalency" calls.

johnnovak: Yeah, they're great! Pushed new version.

src/midi/midi_fluidsynth.cpp
```diff
@@ -22,25 +22,25 @@
...
-#include "pic.h"
-#include "programs.h"
-#include "string_utils.h"
-#include "support.h"
+	#include <bitset>
+	#include <cassert>
+	#include <deque>
+	#include <numeric>
+	#include <string>
+	#include <tuple>
```
weirddan455, line 30: I would remove these indents.  `#include`'s are never indented in the rest of our codebase and I don't think we want to be indenting the entire file because of a top level `#if C_FLUIDSYNTH` that applies to this entire file.

src/midi/midi_fluidsynth.cpp
```diff
@@ -143,7 +143,7 @@ std::tuple<std::string, int> parse_soundfont_pref(const std::string& line)
 	return {filename, percent};
 }
 
-#if defined(WIN32)
+	#if defined(WIN32)
```
weirddan455, line 146: More weird indentation... Not going to comment on every instance of this.

src/midi/midi_fluidsynth.cpp
```diff
@@ -207,7 +207,17 @@ static std_fs::path find_sf_file(const std::string& name)
...
+				// `soundfont = FluidR3_GM.sf2` and `soundfont =
+				// fluidr3_gm.sf2` refer to the same file on case-preserving
+				// filesystems on Windows and macOS.
+				//
+				// `std_fs::canonical` returns the absolute path and matches
+				// its casing to that of the actual physical file. This
+				// prevents certain subtle bugs downstream when we use this
+				// path in comparisons.
+				//
+				return std_fs::canonical(sf).c_str();
```
weirddan455, line 220: I'm not sure that `std_fs::cannoical` does anything with regards to casing.  At least https://en.cppreference.com/w/cpp/filesystem/canonical has no mention of it.

src/midi/midi_fluidsynth.cpp
```diff
@@ -612,7 +637,10 @@ void MidiHandlerFluidsynth::Close()
 	// Reset the members
 	synth.reset();
 	settings.reset();
-	selected_font.clear();
+
+	if (current_sf2_path) {
+		current_sf2_path = {};
+	}
```
weirddan455, line 643: Minor:  Do you really need the if statement?  It always ends up cleared regardless.

src/midi/midi_fluidsynth.cpp
```diff
@@ -869,28 +897,24 @@ std::string format_sf2_line(size_t width, const std_fs::path& sf2_path)
...
 	auto write_line = [&](const std_fs::path& sf2_path) {
-		constexpr auto green = "[color=light-green]";
-		constexpr auto reset = "[reset]";
-
 		const auto line = format_sf2_line(term_width - 2, sf2_path);
-		const bool do_highlight = is_open &&
-		                          (selected_font == sf2_path.string());
+
+		const auto path_matches = current_sf2_path &&
+		                          (*current_sf2_path == sf2_path.string());
```
weirddan455, line 906: Can you not do direct compares of two paths?  Why do you need the `.string()`?

src/midi/midi_fluidsynth.cpp
johnnovak: Yeah, the standard does not mention it, but it's de-facto behaviour. I learned about it because many sources recommended it online.

It's similar to resizing a vector to a smaller size—the standard doesn't really say there will be no realloc involved, but virtually all sane standard libraries won't realloc when shrinking the vector, just update its size internal var.

As we're not aiming to run Staging on all sorts of esoteric C++ compilers and platforms that might implement `canonical` differently, I think it's ok to rely on such de-facto Clang/gcc behaviour.

src/midi/midi_fluidsynth.cpp
johnnovak: Yeah, this is clang-reformat, and it annoys me a bit too. I enabled indenting nested preprocessor directives a while ago:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#indentppdirectives

That works well for stuff like this:

```
#if FOO
  #if BAR
    #include <foo>
  #endif
#endif
```

But not in other cases, apparently. It's not smart enough to DoTheRightThing(tm), so I might disable this feature as I rely on mass-formatting files.

src/midi/midi_fluidsynth.cpp
johnnovak: Might be a leftover from getting an earlier revision compile on MSVC. I vaguely remember I had to do `p1.string() == p2.string()` otherwise MSVC complained about it... But this works, so I'm really unsure. I'll remove `string()` and see what happens 🤷🏻

src/midi/midi_fluidsynth.cpp
johnnovak: Disabled the indenting of preprocessor directives & reformatted.

src/midi/midi_fluidsynth.cpp
johnnovak: Mass fixed it all 👍🏻

src/midi/midi_fluidsynth.cpp
johnnovak: Yeah, works 🤷🏻

src/midi/midi_fluidsynth.cpp
johnnovak: Yeah, cosmetic, as shown on the "current main behaviour" screenshot above. I'm pretty sure it works on Windows too, but I'll confirm.

src/midi/midi_fluidsynth.cpp
johnnovak: Sorry, resolved by mistake then forgot about it. Yeah, the `if` is not needed, removed.

AnonTet:
That sort seems to only work with the default .\soundfonts path.
I for one, have a soundfonts folder and point soundfont = path\soundfond.sf2 so I don't need to copy the .sf2 files over to every dosbox installation.
Would be a nice to have if we have a sfdir much like romdir for mt32 roms.

johnnovak:
> That sort seems to only work with the default .\soundfonts path. I for one, have a soundfonts folder and point soundfont = path\soundfond.sf2 so I don't need to copy the .sf2 files over to every dosbox installation. Would be a nice to have if we have a sfdir much like romdir for mt32 roms.

Why would you need to copy anything? Just set `soundfont = path/to/blah.sf2` in all your DOSBox installations. You can use absolute paths too.

AnonTet:
> > That sort seems to only work with the default .\soundfonts path. I for one, have a soundfonts folder and point soundfont = path\soundfond.sf2 so I don't need to copy the .sf2 files over to every dosbox installation. Would be a nice to have if we have a sfdir much like romdir for mt32 roms.
> 
> Why would you need to copy anything? Just set `soundfont = path/to/blah.sf2` in all your DOSBox installations. You can use absolute paths too.

Which is why I said nice to have and that's what I'm doing.
However "mixer /listmidi" command does not recognize any file that is not under ./soundfonts so it shows not soundfonts whatsoever and would be confusing if we didn't have to configure those ourselves though we can't make sure we didn't forget to uncomment the soundfont we want to test before launching a game if we're not looking at the logs.

Since it recognize those in .soundfonts but not ones with absolute paths, if we could set an alternate path, I'd assume it would recognize those as well and showed them and the predefined one.

interloper98:
@AnonTet , if I'm understanding: you'd like the SF2 scanner to (additionally!) include the `soundfont = some/directory/` (with the file chopped off), specified in the active configuration?

(Sounds reasonable given the user might have a bucket of their favorite SF2 files in their own directory for use that goes beyond DOSBox. And the preceding path in the conf lets us discover this directory)

AnonTet:
> @AnonTet , if I'm understanding: you'd like the SF2 scanner to (additionally!) include the `soundfont = some/directory/` (with the file chopped off), specified in the active configuration?
> 
> (Sounds reasonable given the user might have a bucket of their favorite SF2 files in their own directory for use that goes beyond DOSBox. And the preceding path in the conf lets us discover this directory)

This. Thanks.

johnnovak:
> @AnonTet , if I'm understanding: you'd like the SF2 scanner to (additionally!) include the `soundfont = some/directory/` (with the file chopped off), specified in the active configuration?
> 
> (Sounds reasonable given the user might have a bucket of their favorite SF2 files in their own directory for use that goes beyond DOSBox. And the preceding path in the conf lets us discover this directory)

A bit questionable bahaviour, or at least inconsistent. Why don't you rather use symlinking the SF2 directories @AnonTet ? That's a cleaner solutions for such special custom setups, you can do that of NTFS too.

AnonTet:
Never though of symlinks. It works and "mixer /listmidi" shows all the files under soundfonts. 

I still think the most elegant solution would be "soundfont_dir = path/to/soundfonts" parameter IF it is easy to implement.

interloper98:
I agree @johnnovak; _"`soundfont =` is the name or path to your desired soundfont.. but wait, there's more! We'll list peer SF2's found alongside your soundfont in the `mixer /listmidi` command"_. 

That's a bit overloaded and magic compared to a `soundfont_dir = ` setting, which is more explicit.  If left default, soundfonts are looked up from the standardised  locations (for the given platform). When set, soundfonts are read from the custom directory.

johnnovak:
> I agree @johnnovak; _"`soundfont =` is the name or path to your desired soundfont.. but wait, there's more! We'll list peer SF2's found alongside your soundfont in the `mixer /listmidi` command"_.
> 
> That's a bit overloaded and magic compared to a `soundfont_dir = ` setting, which is more explicit. If left default, soundfonts are looked up from the standardised locations (for the given platform). When set, soundfonts are read from the custom directory.

Yeah, absolutely. That's what I'll do in a follow up PR. I'm not too much of a fan of hidden magical behaviour 😆

src/midi/midi_fluidsynth.cpp
johnnovak: Confirmed, works on Windows too.

johnnovak:
Can I please get a tick @weirddan455 ?

weirddan455 (approved):
Yep, looks good

PR #3983, opened 2024-10-11
Discard incoming modem traffic from waiting clients

NicknineTheEagle:
# Description

Another small fix for modem. This makes it so any incoming traffic from the client waiting for the call to be accepted is read and then discarded. Same goes for the connected client while the modem is in command mode. This is something you want to do because otherwise incoming TCP packets pile up in the receive queue without being acked which is bad - TCP protocol with throttle connection thinking that the program isn't fast enough to process incoming packets and the client will repeatedly re-transmit packets. I assume UDP would have similar problems.
As a bonus, if the dialing modem hangs up, the server will now actually stop ringing and go back to idle state whereas previously, it would ring indefinitely until the modem was reset.

# Manual testing

Tested dialing with Telemate and hanging up before the call is answered. The server properly closes the waiting client socket. Tested Doom, no regressions.

# Checklist

I have:

### DIFF
--- a/src/hardware/serialport/softmodem.cpp
+++ b/src/hardware/serialport/softmodem.cpp
@@ -991,6 +991,7 @@ void CSerialModem::Timer2()
 			EnterIdleState();
 		}
 	}
+
 	// Handle incoming to the serial port
 	if (!commandmode && clientsocket && rqueue->left()) {
 		size_t usesize = rqueue->left() >= 16 ? 16 : rqueue->left();
@@ -1008,8 +1009,30 @@ void CSerialModem::Timer2()
 		}
 	}
 
+	// Discard any incoming traffic while in command mode
+	if (commandmode && clientsocket) {
+		// Read buffer size doesn't really matter, we just need big
+		// enough buffer to discard data decently quickly
+		size_t usesize = 80;
+		if (!clientsocket->ReceiveArray(tmpbuf, usesize)) {
+			SendRes(ResNOCARRIER);
+			LOG_INFO("SERIAL: No carrier on receive");
+			EnterIdleState();
+		}
+	}
+
+	// Discard any incoming traffic from the waiting client
+	if (waitingclientsocket) {
+		// Read buffer size doesn't really matter, we just need big
+		// enough buffer to discard data decently quickly
+		size_t usesize = 80;
+		if (!waitingclientsocket->ReceiveArray(tmpbuf, usesize)) {
+			EnterIdleState();
+		}
+	}
+
 	// Tick down warmup timer
-	if (clientsocket && warmup_remain_ticks) {
+	if (connected && warmup_remain_ticks) {
 		// Drop all incoming and outgoing traffic for a short period after
 		// answering a call. This is to simulate real modem behavior where
 		// the first packet is usually bad (extra data in the buffer from


src/hardware/serialport/softmodem.cpp
```diff
@@ -1008,8 +1009,26 @@ void CSerialModem::Timer2()
 		}
 	}
 
+	// Discard any incoming traffic while in command mode
+	if (commandmode && clientsocket) {
+		size_t usesize = 80;
```
johnnovak, line 1014: I've done some digging, and looks like that's the max number of bytes to be received in bytes. The name should reflect that (e.g. `constexpr size_t MaxReceiveBytes = 80` or something).

Also, why 80 and not any other number? I think a comment is in order to explain the magic number.

src/hardware/serialport/softmodem.cpp
```diff
@@ -1008,8 +1009,26 @@ void CSerialModem::Timer2()
...
+		if (!clientsocket->ReceiveArray(tmpbuf, usesize)) {
+			SendRes(ResNOCARRIER);
+			LOG_INFO("SERIAL: No carrier on receive");
+			EnterIdleState();
+		}
+	}
+
+	// Discard any incoming traffic from the waiting client
+	if (waitingclientsocket) {
+		size_t usesize = 80;
```
johnnovak, line 1024: Could this `80` constant be extracted into the outer scope then instead of copy/pasting it? I assume 80 is some important magic number here, and the code would work differently with receiving less or more than 80 bytes max? (I'm really unsure, so I'm asking 😄 )

johnnovak:
Nice one @NicknineTheEagle, just some comments about clarifying a few things.

Would it make sense to call this change out in the release notes? If so, could you write a release notes entry for it?

src/hardware/serialport/softmodem.cpp
NicknineTheEagle: I just wanted to pick some big number to make sure we can discard data as quickly as possible.

NicknineTheEagle:
> Nice one @NicknineTheEagle, just some comments about clarifying a few things.
> 
> Would it make sense to call this change out in the release notes? If so, could you write a release notes entry for it? (If it's too insignificant, that's fine too, but I assume this fixes some annoyances when playing networked games that the users must have experienced?)

I've deliberately omitted release notes. This is just an internal optimization, it's not really a user facing change and average Joe doesn't know anything about packets and receive queues. I think there's no reason to mention this.

src/hardware/serialport/softmodem.cpp
johnnovak: Put this comment in there then:

"The read size doesn't really matter; we just want to discard data quickly enough".

johnnovak:
> > Nice one @NicknineTheEagle, just some comments about clarifying a few things.
> > Would it make sense to call this change out in the release notes? If so, could you write a release notes entry for it? (If it's too insignificant, that's fine too, but I assume this fixes some annoyances when playing networked games that the users must have experienced?)
> 
> I've deliberately omitted release notes. This is just an internal optimization, it's not really a user facing change and average Joe doesn't know anything about packets and receive queues. I think there's no reason to mention this.

Fair point, thanks. I just ask because I have zero experience with networked DOS games 😄

src/hardware/serialport/softmodem.cpp
```diff
@@ -1008,8 +1009,28 @@ void CSerialModem::Timer2()
...
```
johnnovak, line 1026: We need that comment here too as we're doing the same thing.

johnnovak (approved):
Thanks @NicknineTheEagle — another nice modem improvement 🚀

Issue #3991, opened 2024-10-16
Pointer casts in src/cpu/core_dynrec/risc_ppc.h lose precision


oturpe:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Linux ppc64le

### If Other OS, please describe

Fedora

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

I am trying to build Fedora package for 0.82.0 version. It builds successfully for other architectures, but for ppc64le, I see many errors of the following kind:

```
In file included from ../src/cpu/core_dynrec.cpp:176:
../src/cpu/core_dynrec/risc_ppc.h: In function ‘HostReg gen_addr(int32_t&, HostReg)’:
../src/cpu/core_dynrec/risc_ppc.h:170:22: error: cast from ‘Segments*’ to ‘int32_t’ {aka ‘int’} loses precision [-fpermissive]
  170 |         off = addr - (int32_t)&Segs;
      |                      ^~~~~~~~~~~~~~
../src/cpu/core_dynrec/risc_ppc.h:177:22: error: cast from ‘CPU_Regs*’ to ‘int32_t’ {aka ‘int’} loses precision [-fpermissive]
  177 |         off = addr - (int32_t)&cpu_regs;
      |                      ^~~~~~~~~~~~~~~~~~
```

Probably, Fedora's compiler flags also have a role here, in making that an error rather than just a warning. So one path forward could be to adjust the flags during dosbox-staging build. But I report this issue here, so you can take a look and decide if you want to do something about this.

### Steps to reproduce the behaviour.

From [build log](https://kojipkgs.fedoraproject.org//work/tasks/5958/124895958/build.log):

```
CFLAGS='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -mcpu=power8 -mtune=power8 -fasynchronous-unwind-tables -fstack-clash-protection '
export CFLAGS
[...]

PR #3996, opened 2024-10-19
Fix platform specific parameters for ppc64le

oturpe:
# Description

Previously, build script assumed that for little-endian PowerPC, Meson would report `ppc64le` as `cpu_family`. Actually, Meson returns just `ppc64` and endianness needs to be checked separately. Fix the Meson build script to do that.

Without this change, compiling on `ppc64le`  included 32-bit PowerPC code from `risc_ppc.h`, causing many errors about invalid casts from pointer to 32-bit integer types.

This issue was revealed by building DOSBox Staging 0.82.0 for Fedora in Fedora build environment. Builds for other architectures (x86_64, aarch64 and s390x) succeeded, but ppc64le failed due to this problem.

To help test the change, the issues in previously broken _Platform targets_ workflow are fixed.

## Related issues

#3991

# Manual testing

After applying this patch, Fedora build succeeded for all  architectures (x86_64, aarch64, ppc64le and s390x). The one for which I have hardware for, x86_64, also ran successfully.

I tested that _Platform builds_ workflow, with workflow fixes from this pull request otherwise succeeds, but ppc64le build fails. After including also the build fix from this pull request, aslo _Platform builds_ succeeds.

# Checklist

I have:

### DIFF
--- a/src/cpu/meson.build
+++ b/src/cpu/meson.build
@@ -14,32 +14,34 @@ conf_data.set10('C_PER_PAGE_W_OR_X', per_page_w_or_x_pref.auto() or per_page_w_o
 
 # Platform-specific
 core_selection = [
-    # cpu_family                  selected_core        dynrec_define    target     unaligned  per-page
-    #                                                                              mem        W^X
-    [ ['x86_64'],                 ['auto', 'dyn-x86'], 'C_DYNAMIC_X86', 'X86_64',  1,         1 ],
-    [ ['x86'],                    ['auto', 'dyn-x86'], 'C_DYNAMIC_X86', 'X86',     1,         1 ],
-    [ ['x86_64'],                 ['dynrec'],          'C_DYNREC',      'X86_64',  1,         1 ],
-    [ ['x86'],                    ['dynrec'],          'C_DYNREC',      'X86',     1,         1 ],
-    [ ['aarch64'],                ['auto', 'dynrec'],  'C_DYNREC',      'ARMV8LE', 1,         1 ], # ARMv8+ (64-bit)
-    [ ['arm'],                    ['auto', 'dynrec'],  'C_DYNREC',      'ARMV7LE', 1,         1 ], # ARMv7+
-    [ ['armv6'],                  ['auto', 'dynrec'],  'C_DYNREC',      'ARMV4LE', 0,         0 ], # ARMv6 or older
-    [ ['ppc64',   'powerpc64'],   ['auto', 'dynrec'],  'C_DYNREC',      'POWERPC', 1,         1 ], # 64 bit PPC processors (big-endian), running 32-bit big-endian code
-    [ ['ppc64le', 'powerpc64le'], ['auto', 'dynrec'],  'C_DYNREC',      'PPC64LE', 1,         1 ], # 64 bit PPC processors (little-endian)
-    [ ['ppc',     'powerpc'],     ['auto', 'dynrec'],  'C_DYNREC',      'POWERPC', 1,         0 ], # 32 bit PPC processors (big-endian)
-    [ ['mips'],                   ['auto', 'dynrec'],  'C_DYNREC',      'MIPSEL',  0,         0 ], # 32 bit MIPS processor
+    # cpu_family              endian      selected_core        dynrec_define    target     unaligned  per-page
+    #                                                                                      mem        W^X
+    [ ['x86_64'],             ['little'], ['auto', 'dyn-x86'], 'C_DYNAMIC_X86', 'X86_64',  1,         1 ],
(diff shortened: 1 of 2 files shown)


oturpe:
I cannot find a way to add labels my self, so that is the reason that box is unchecked. Other unchecked boxes are not applicable to this change.

oturpe:
I would like to test the result different arches, but unfortunately I do not have the hardware.

johnnovak:
Related: we have a "Platform builds" action which includes a `ppc64le` target, but looks like it's broken. It was kcgen's pet project, so I'm not saying anyone from the core team _should_ maintain it, but if someone wants to, here's the opportunity.

Personally, I don't care about it as we don't officially support these platforms, and we want to reduce the number of things we maintain, not increase it. So I'm inclined to kill it (delete the workflow, so it's gone for good), but I thought I'd ask first.
 
https://github.com/dosbox-staging/dosbox-staging/actions/runs/11329477154

johnnovak:
> I cannot find a way to add labels my self, so that is the reason that box is unchecked. Other unchecked boxes are not applicable to this change.

Yeah, you can't do that as an external contributor. Fixed it.

> I would like to test the result different arches, but unfortunately I do not have the hardware.

I'm not trying to give you extra homework 😆, but if you _really_ want to do this, the platform builds action I mentioned in my previous comment is a way without having access to the real hardware:

https://github.com/dosbox-staging/dosbox-staging/blob/main/.github/workflows/platforms.yml#L35-L49

johnnovak (approved):
Looks all good to me @oturpe, thanks! 🎉

runlevel5:
Look good to me 👍

src/cpu/meson.build
```diff
@@ -14,32 +14,34 @@ conf_data.set10('C_PER_PAGE_W_OR_X', per_page_w_or_x_pref.auto() or per_page_w_o
...
+    # cpu_family              endian      selected_core        dynrec_define    target     unaligned  per-page
+    #                                                                                      mem        W^X
+    [ ['x86_64'],             ['little'], ['auto', 'dyn-x86'], 'C_DYNAMIC_X86', 'X86_64',  1,         1 ],
+    [ ['x86'],                ['little'], ['auto', 'dyn-x86'], 'C_DYNAMIC_X86', 'X86',     1,         1 ],
+    [ ['x86_64'],             ['little'], ['dynrec'],          'C_DYNREC',      'X86_64',  1,         1 ],
+    [ ['x86'],                ['little'], ['dynrec'],          'C_DYNREC',      'X86',     1,         1 ],
+    [ ['aarch64'],            ['little'], ['auto', 'dynrec'],  'C_DYNREC',      'ARMV8LE', 1,         1 ], # ARMv8+ (64-bit)
+    [ ['arm'],                ['little'], ['auto', 'dynrec'],  'C_DYNREC',      'ARMV7LE', 1,         1 ], # ARMv7+
+    [ ['armv6'],              ['little'], ['auto', 'dynrec'],  'C_DYNREC',      'ARMV4LE', 0,         0 ], # ARMv6 or older
+    [ ['ppc64', 'powerpc64'], ['big'],    ['auto', 'dynrec'],  'C_DYNREC',      'POWERPC', 1,         1 ], # 64 bit PPC processors (big-endian), running 32-bit big-endian code
```
runlevel5, line 26:
```suggestion
    [ ['ppc64', 'powerpc64'], ['big'],    ['auto', 'dynrec'],  'C_DYNREC',      'POWERPC', 1,         1 ], # 64 bit PPC processors (big-endian), running 64-bit big-endian code
```

otto-liljalaakso-nt: I also found that comment surprising. But, my attempt here is to avoid changing anything for any other row that the 64-bit little-endian PowerPC one, so I refrained from making any other changes than adding the `endian` column.

I can add a comment change if it is certain that the old comment was wrong. However, I suspect it is actually correct. The failure that requires this fix is that ppc64le arch selected the highlighted row, and build failed because the its code path contains pointer conversions to 32-bit integers. To me, that sounds like "running 32-bit code" is actually accurate. (Talking as somebody who knows very little about PowerPC, and has just general understanding of ISAs.)

johnnovak:
Great stuff @oturpe, thanks for doing all this 🎉 The MSVC failure is unrelated, so I'm merging this.

johnnovak (approved):

interloper98:
Right on; I think those platform builds have been failing for as long as I've been on the project! I wonder if it's needed for 0.82, as well? (I'm guessing fedora and other distributions can benefit from this)

johnnovak:
> Right on; I think those platform builds have been failing for as long as I've been on the project! I wonder if it's needed for 0.82, as well? (I'm guessing fedora and other distributions can benefit from this)

Well, most of it is CI stuff, but we might make it available under the `0.82.0` tag, why not 😄

PR #3997, opened 2024-10-19
Fix file time handling on Windows

NicknineTheEagle:
# Description

This fixes getting/setting file modify time on Windows. GetFileTime and SetFileTime operate with UTC timestamps whereas DOS stores local time - and since FileTimeToDosDateTime and DosDateTimeToFileTime do not do any time zone conversion, the current implementations of get_dos_file_time of set_dos_file_time output incorrect time.

I've added conversion between UTC time and local time to fix this.

Note that I did not check POSIX implementation since I don't have the necessary environment. Someone will need to test it to make sure it doesn't have the same problem.

# Release notes

None since this is a change to in-dev code.

# Manual testing

Tested installing Prince of Persia from floppy, timestamps are correctly copied over.

Tested saving in Doom, timestamp is correctly set to current computer time.

# Checklist

I have:

### DIFF
--- a/src/misc/fs_utils_win32.cpp
+++ b/src/misc/fs_utils_win32.cpp
@@ -226,21 +226,34 @@ DosDateTime get_dos_file_time(const NativeFileHandle handle)
 	ret.time        = 1;
 	ret.date        = 1;
 
-	FILETIME last_write_time = {};
-	if (!GetFileTime(handle, nullptr, nullptr, &last_write_time)) {
+	FILETIME write_time = {};
+	if (!GetFileTime(handle, nullptr, nullptr, &write_time)) {
 		return ret;
 	}
 
-	FileTimeToDosDateTime(&last_write_time, &ret.date, &ret.time);
+	FILETIME local_write_time = {};
+	if (!FileTimeToLocalFileTime(&write_time, &local_write_time)) {
+		return ret;
+	}
+
+	FileTimeToDosDateTime(&local_write_time, &ret.date, &ret.time);
 	return ret;
 }
 
-void set_dos_file_time(const NativeFileHandle handle, const uint16_t date, const uint16_t time)
+void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
+                       const uint16_t time)
 {
-	FILETIME last_write_time = {};
-	if (DosDateTimeToFileTime(date, time, &last_write_time)) {
-		SetFileTime(handle, nullptr, nullptr, &last_write_time);
+	FILETIME local_write_time = {};
+	if (!DosDateTimeToFileTime(date, time, &local_write_time)) {
+		return;
 	}
+
+	FILETIME write_time = {};
+	if (!LocalFileTimeToFileTime(&local_write_time, &write_time)) {
+		return;
+	}
+
+	SetFileTime(handle, nullptr, nullptr, &write_time);
 }
 
 #endif


FeralChild64:
> I did not check POSIX implementation since I don't have the necessary environment. Someone will need to test it to make sure it doesn't have the same problem

Just checked, looks like the timestamps on Linux are correct (I'm currently using GMT+2 time zone).

vs/dosbox.vcxproj
```diff
@@ -276,7 +276,13 @@ IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0</Command>
...
     <PostBuildEvent>
       <Command>%systemroot%\System32\robocopy ..\contrib\resources "$(OutDir)resources" /mir /xf "meson.build" "*.sh"
-IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0</Command>
+IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0
+rd /s /q D:\Emulators\Dosbox\resources
+del D:\Emulators\Dosbox\*.exe
+del D:\Emulators\Dosbox\*.dll
+copy $(TargetDir)$(TargetName)$(TargetExt) D:\Emulators\Dosbox
+copy $(TargetDir)*.dll D:\Emulators\Dosbox
+xcopy $(TargetDir)resources D:\Emulators\Dosbox\resources /E /I</Command>
```
weirddan455, line 285: What is all this?  I assume you didn't mean to commit this so please remove it.  If it's intentional, put this in a separate commit and explain the purpose.

weirddan455 (commented):
Good catch.  I tested in Windows to confirm the bug and your fix.  I also tested in Linux and it's already doing the conversions on both read and write so no changes needed to the POSIX layer.

vs/dosbox.vcxproj
NicknineTheEagle: Oh, FFS, not again.

weirddan455 (approved):
Looks good.  Thanks for the fix.

weirddan455:
I just want to ping @johnnovak before merge since we're getting close to release.  IMO this should go in 0.82.0.  This was a regression from 0.81.2.

johnnovak:
> I just want to ping @johnnovak before merge since we're getting close to release. IMO this should go in 0.82.0. This was a regression from 0.81.2.

Yes, we'll include this in 0.820. Nice catch @NicknineTheEagle and thanks for the fix 🎉 Can you please fill out the release notes section then?

NicknineTheEagle:
> > I just want to ping @johnnovak before merge since we're getting close to release. IMO this should go in 0.82.0. This was a regression from 0.81.2.
> 
> Yes, we'll include this in 0.820. Nice catch @NicknineTheEagle and thanks for the fix 🎉 Can you please fill out the release notes section then?

As I already stated, there are no release notes to make since this fixes a regression from 0.81.2.

johnnovak:
> > > I just want to ping @johnnovak before merge since we're getting close to release. IMO this should go in 0.82.0. This was a regression from 0.81.2.
> > 
> > 
> > Yes, we'll include this in 0.820. Nice catch @NicknineTheEagle and thanks for the fix 🎉 Can you please fill out the release notes section then?
> 
> As I already stated, there are no release notes to make since this fixes a regression from 0.81.2 where file timestamps worked correctly.

Yeah, okay, now I got it. Just a tip, maybe be a bit more clear next time, so this would be better:

> None, regressions since 0.81.2

instead of 

> None since this is a change to in-dev code.

PR #4000, opened 2024-10-21
Improve platform builds workflow

oturpe:
# Description

_Platform builds_ workflow was broken for ppc64le and s390x architectures. This was because run-on-arch-action used by that workflow had reduced the set of available (os, arch) pairs. Migrate to supported pairs, and do two general improvements:

* Allow workflow dispatch for this workflow, so it can be run manually in addition to running periodically
* Fix version on run-on-arch-action, so that in the future, breaking changes there do not automatically break the workflow

## Related issues

Problems testing #3991 motivated this effort.

# Manual testing

Ran the workflow in my fork. Got s390x to pass and ppc64le to fail in the way that #3991 fixes. All ARM builds are failing for some reason related to zlib-ng Meson wrap. It looks like a genuine build problem when using that Meson wrap, so out of scope for this pull request where the focus is in the workflow itself.

# Checklist

I have:

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -54,7 +54,7 @@ jobs:
         with:
           submodules: false
 
-      - uses: uraimo/run-on-arch-action@master
+      - uses: uraimo/run-on-arch-action@v2.8.1
         name: Build artifact
         id: build
         with:
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -2,6 +2,7 @@ name: Platform builds
 
 on:
   schedule: [cron: '0 14 * * *']
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -42,12 +42,12 @@ jobs:
           - name: ARM64 (Ubuntu 20.04)
             arch: aarch64
             distro: ubuntu20.04
-          - name: s390x (Fedora)
+          - name: s390x (Debian "bookworm")
             arch: s390x
-            distro: fedora_latest
-          - name: ppc64le (Fedora)
+            distro: bookworm
+          - name: ppc64le (Debian "bookworm")
             arch: ppc64le
-            distro: fedora_latest
+            distro: bookworm
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


johnnovak:
@oturpe Can't you just run the action in your forked repo? You wouldn't have privileges to run actions manually in our repo anyway.

oturpe:
> Can't you just run the action in your forked repo? You wouldn't have privileges to run actions manually in our repo anyway.

That was basically what I was trying to here. The pull request was intended for my fork, but accidentally ended up here.

Since I did some fixes for the _Platform builds_ workflow, let me change this pull request to propose those fixes. I have updated title and description, and am now actually proposing these for merge.

oturpe:
Here is an example from my fork about how _Platform builds_ now looks like: https://github.com/oturpe/dosbox-staging/actions/runs/11462012162/job/31892411919

johnnovak:
> That was basically what I was trying to do here. The pull request was intended for my fork, but accidentally ended up here.

@oturpe Thanks, makes sense. This is a good incremental improvement so happy to merge as-is.

I saw you ran into problems with zlib-ng and Meson. Note that our future goal is to move away from Meson entirely and use CMake directly. That would sidestep the whole problem, but I don't have any timeline for this. Hopefully it will happen in 2025.

johnnovak (approved):

FeralChild64:
@Torinde

1. Probably no team member has any RISC-V hardware, so we would have no idea whether the build works at all.
2. So far the RISC-V CPU performance was disappointing - much slower than Raspberry Pi.
3. There is no dynamic recompiler for RISC-V, which means DOSBox Staging would be VERY slow.

oturpe:
In the technical sense, adding RISC-V should be simple. _Platform builds_ uses _run-on-arch-action_, and its [Supported Platforms](https://github.com/uraimo/run-on-arch-action?tab=readme-ov-file#supported-platforms) table lists `riscv64`. So, you would only need to add one more item to the [platform builds matrix](https://github.com/dosbox-staging/dosbox-staging/blob/66f60749c311116596493c3e7e202b83912b831f/.github/workflows/platforms.yml#L34) using that arch. I will leave the decision if it is a good idea or not to the core team.

johnnovak:
> Is it possible to add there RISC-V (RV64I) as well?

Of course! You add it in a PR, test it, then maintain it 😆 All possible 😎

PR #4002, opened 2024-10-21
Cache MT32 ROM records

### LINKED ISSUE #3994: MIXER /LISTMIDI freezes for ~30 seconds

dosbox-staging version 0.82.0-alpha (690e5)

Windows 10

_No response_

### Relevant hardware info

_No response_

running MIXER /LISTMIDI freezes the sytem for ~30 second before listing the configuration

Downloaded latest 82 alpha (0.82.0-alpha (690e5) running with default settings (using empty dosbox-staging.conf in the main folder and letting dosbox fill it) with mt32-roms folder 

I see you are refactoring the MIDI section in the 2 pull requests for post 0.82 so I though I'd mention this.

Explain how to reproduce
1. download alpha
2. add empty dosbox-staging.conf in the root
3. add mt32-roms folder in the root with some roms (Valid ones, I checked)
4. start dosbox
5. run MIXER /LISTMIDI

_No response_

### Your configuration

```ini
default dosbox-staging.conf
```

### Provide a Log

```text
no entries seen in the log
```

- [X] Yes, I agree.

interloper98:
# Description

This PR adds caching to the MT32 ROM lookup routine to eliminate duplicate reads caused by `mixer /listmidi` or changing the MT32 model at runtime.

It's light-weight: only metadata is cached (`filesystem::path`s and ROM ID strings are stored, so we're looking at one KB or less for all the ROMs).

## Related issues

Fixes #3994 

# Release notes

MT32 ROM scanning is now faster.

# Manual testing

Tested multiple source directories, some empty and some populated with good and bad ROM files.

Added logging to make sure that directories aren't scanned twice (even bad directories aren't re-entered), and both good and bad ROMs are also only scanned once.

Tested a directory holding 10,000 ROMs held in a local RAM drive, and the PR now processes `mixer /listmidi` in the same millisecond, where as `main` takes several seconds to chew through it.

# Checklist

I have:

### DIFF
--- a/src/midi/midi_lasynth_model.cpp
+++ b/src/midi/midi_lasynth_model.cpp
@@ -23,7 +23,8 @@
 #if C_MT32EMU
 
 #include <cassert>
-#include <unordered_set>
+#include <map>
+#include <set>
 
 #include "fs_utils.h"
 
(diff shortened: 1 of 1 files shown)


src/midi/midi_lasynth_model.cpp
```diff
@@ -50,48 +51,72 @@ LASynthModel::LASynthModel(const std::string &rom_name,
...
 	}
 
-	std::error_code ec;
+	// Named types to make the directory cache self-documenting
+	using RomId     = std::string;
+	using RomPath   = std_fs::path;
+	using DirPath   = std_fs::path;
+	using RomsCache = std::map<RomId, RomPath>;
+	using DirsCache = std::map<DirPath, RomsCache>;
+	using JunkFiles = std::set<RomPath>;
```
johnnovak, line 70: IMO, the path aliases just obscure the intent, so I'd suggest getting rid of them.

I also have a slight preference for `UnknownFiles` vs `JunkFiles`, but we're splitting hairs here 😆 

```cpp
using RomId        = std::string;
using RomsCache    = std::map<RomId, std_fs::path>;
using DirsCache    = std::map<std_fs::path, RomsCache>;
using UnknownFiles = std::set<std_fs::path>;
```

src/midi/midi_lasynth_model.cpp
```diff
@@ -50,48 +51,72 @@ LASynthModel::LASynthModel(const std::string &rom_name,
...
+
+	static DirsCache dirs_cache = {};
+	static JunkFiles junk_files = {};
```
johnnovak, line 73: Just `static std::set<std_fs::path> unknown_files = {}`?

interloper98: Yup; at first I created  std::map<std_fs::path, std::map<std::string, std_fs::path> > and thought "nope, this looks like a maze" so just aliased each key value.. but dropping the `std_fs::path`s is a better balance and is still self-documenting enough. 

Fixed; thanks @johnnovak!

interloper98:
Suggest this be considered 0.82 because the current behavior is pretty excessive. Every file in an mt32 romdir is read and checksumed for all 27 supported ROM files. 

So if you have 15 ROM files, this generates 405 full file reads and checksums every time you change or list the roms. 

This PR knocks that example down to 15 reads and checksums, and not any more.

src/midi/midi_lasynth_model.cpp
johnnovak: Yeah, some type aliasing can help, but I'm finding too much just obscures the intent. I generally steer away from aliasing simple types directly, but `RomId = std::string` helps here for sure.

I find quite often you need to turn these types into simple structs anyway if you need method-like behaviour on them. E.g., check out my `struct ImageInfo` thing in `video.h`.

johnnovak:
> Suggest this be considered for 0.82 because the current behavior is pretty excessive. Every file in an mt32 romdir is read and checksumed for all 27 supported ROM files.
> 
> So if you have 15 ROM files, this generates 405 full file reads and checksums every time you change or list the roms.
> 
> This PR knocks that example down to 15 reads and checksums, and not any more.

I was wondering about that, and good call, I agree—0.82.0 material! 🎉 Thanks for doing this @interloper98 🥳

johnnovak (approved):

src/midi/midi_lasynth_model.cpp
interloper98: Thanks for the tip; will check it out next time 👍

PR #4012, opened 2024-10-27
[post 0.82] Update and always use the vcpkg baseline in CI

### LINKED ISSUE #4010: vcpkg & glib impasse
According to the vcpkg team, our glib 2.78.4 isn't compatible with the newer version of vcpkg's embedded python 3.12.

Our options are rolling back to an older vcpkg that uses python 3.11 or using the newer glib 2.80 (which is compatible with the latest vcpkg tools)

See https://github.com/microsoft/vcpkg/issues/41791#event-14908038274

I grepped for `2.78.4` to find where this glib version is defined, but I'm not getting any hits; how is this stuff defined?

Hunch is there's near zero risk with either approach, but I don't know how to progress it. Need help 📯

interloper98:
# Description

Vcpkg in CI has been failing to build with glib errors for a while now.

This PR bumps the vcpkg baseline just far enough such that we're still using the currently pinned glib 2.78.4, without going to 2.80.x.

The other more important long-term change is that the project baseline will be checked out on the CI host (the vcpkg installation itself is a git repo). 

This means that the we will "hold vcpkg back in time" to always match that of the baseline. For example, the baseline uses python 3.11 however if we went with the latest vcpkg baseline then it would use python 3.12.

Previous we weren't doing this checkout step, so the host vcpkg was using newer and newer tooling, which eventually clashed with our older desired packages.

## Related issues

Fixes #4010 

# Release notes

None.

# Manual testing

None; just a CI fix.

# Checklist

I have:

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -45,6 +45,16 @@ jobs:
         with:
           submodules: false
 
+      - name: Checkout vcpkg baseline
+        shell: pwsh
+        run: |
+          $baseline = (Get-Content vcpkg.json | ConvertFrom-Json).'builtin-baseline'
+          cd $env:VCPKG_INSTALLATION_ROOT
+          git fetch
+          rm vcpkg.exe
+          git -c advice.detachedHead=false checkout $baseline
+          bootstrap-vcpkg.bat -disableMetrics
+
       - name: Export GitHub Actions cache environment variables
         uses: actions/github-script@v7
         with:
@@ -111,6 +121,16 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: false
+
+      - name: Checkout vcpkg baseline
+        shell: pwsh
+        run: |
+          $baseline = (Get-Content vcpkg.json | ConvertFrom-Json).'builtin-baseline'
+          cd $env:VCPKG_INSTALLATION_ROOT
+          git fetch
+          rm vcpkg.exe
+          git -c advice.detachedHead=false checkout $baseline
+          bootstrap-vcpkg.bat -disableMetrics
 
       - name: Export GitHub Actions cache environment variables
         uses: actions/github-script@v7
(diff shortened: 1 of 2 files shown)


interloper98:
Not sure if this is needed for 0.82.

johnnovak:
Thanks for doing this @interloper98. Pinning vcpkg itself is not a bad idea, although the problem here is the horrible Glib package that relies on the Python version used by vcpkg or the system Python version. That is just asking for trouble.

johnnovak:
> Not sure if this is needed for 0.82.

No, I'm releasing it in 5 mins.

shermp:
One concern I have is that the `vcpkg.exe` executable may or may not be suitable for the checked out repository. The executable is downloaded when running the `bootstrap-vcpkg.bat` script.

Would you be interested in investigating https://github.com/lukka/run-vcpkg instead? That may be a better way to go.

johnnovak (approved):
Thanks @interloper98 , looks alright to me 🎉

interloper98:
@shermp  - thanks for that suggestion. I've added the bootstrap fetch to it, and confirmed that it's pulling the baseline's release executable.

So for this PR, the baseline (just prior to the python 3.12 switch) pulls the Sep-30 `vcpkg.exe` release.

```
vcpkg package management program version 2024-09-30-ab8988503c7cffabfd440b243a383c0a352a023d
```

> Would you be interested in investigating https://github.com/lukka/run-vcpkg instead? That may be a better way to go.

The checkout and bootstrap are basic mechanics, and they seems to be working fine. The nice thing is anyone can do these steps on their own `c:\vcpkg` too, where as the `lukka/run-vcpkg` action relies on GitHub's CI machinery. 

Also, I sure hope vcpkg can get this version correct.. it would be very strange for Microsoft's own versioning repository system to ultimately rely on some random dudes scripts to change the version! (no offence, lukka!) That's just musing though.. there's good money in augmenting corpo bloatware, so we might visit lukka one day :laughing:

Issue #1864, opened 2022-08-17
Pasteboard support for DOSBox Staging


Grounded0:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Mac OS

### Is your feature request related to a problem? Please describe.

Having a copy-paste support from DOSBox Staging to the client OS and back would be a nice feature and make life a little bit easier.

### Describe the solution you'd like

Having a support for X11 pasteboard, Windows clipboard and macOS pasteboard and syncronising them to DOSBox Staging. Windows 3.x has APIs for them while in DOS you would probably have to use a character injection scheme like autotyping for pasting.

### Describe alternatives you've considered

Functionality cannot be achieved without client OS hooks in place.

### Add any other context or screenshots about the feature request here.

SheepShaver is SDL2 based and already does this perfectly:
<img width="1392" alt="Screenshot 2022-08-17 at 4 25 47" src="https://user-images.githubusercontent.com/48192273/185014314-aaf74bd1-436d-4a4b-b99c-489f6c05c85d.png">

DOSBox-X also does this but cannot comment on the implementation as haven't tested it.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Torinde:
Would be good if this also handles DOS codepage to Host codepage (Unicode) conversion - similar to #2422

Grounded0:
Now we got codepage conversion to UTF-8 and back engine too:

- #2422

FeralChild64:
Starting from today, my work-in-progress `EDIT.COM` replacement (https://github.com/dosbox-staging/dosbox-staging/tree/fc/edit-1, issue https://github.com/dosbox-staging/dosbox-staging/issues/2463) can paste text from host clipboard. Remarks:
- the editor itself is still FAR from being finished
- for now only pasting (CTRL+V) works, as text selection is not implemented yet
- Windows build system is not updated yet - it will only build on Linux or macOS
- the clipboard support code is implemented as a separate API, not limited to text-GUI applications - in case someone someday wants to implement i. e. hack to provide Windows clipboard integration (but I am not going to be this person :D)

Torinde:
There is some work in DOSEMU2 on DOSbox-X compatible `CLIP$`: dosemu2/dosemu2/issues/1179

Grounded0:
[...]

PR #4036, opened 2024-11-01
Initial clipboard sharing support

FeralChild64:
# Description

Initial clipboard sharing support - see _Release notes_ for details. If secure mode is enabled, clipboard functions in _detached_ mode - appears to work, but there is no exchange with the host clipboard.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/1864

# Release notes

- Implemented command `CLIP.EXE` (originally available on the _Windows 98 Resource Kit_ CD, now part of modern _Microsoft Windows_ - [link](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/clip)) to allow clipboard handling from within BAT scripts or command line.
- Pressing _Ctrl+V_ in DOSBox command line now pastes the first line of the clipboard's text content.

# Manual testing

- using your favorite (host side) ASCII editor, put something into system clipboard; from within DOSBox execute command `clip` or press Ctrl+V to test pasting
- execute command `dir | clip` or `clip < Z:\AUTOEXEC.BAT` within DOSBox - check that the result can be now pasted on the host side, or on the guest side (see previous point)

Don't be afraid to try national characters, or to use `keyb` command to switch code pages while the text is in the clipboard - example (RightAlt + letter produces a relevant Polish national character):

```BAT
keyb pl 668
echo ąęść | clip
clip
keyb pl 667
clip
keyb us 437
clip
keyb pl 667
clip
keyb pl 668
clip
```

# Checklist

I have:


johnnovak:
Cool feature @FeralChild64 👍🏻 

> If secure mode is enabled, clipboard functions in detached mode - appears to work, but there is no exchange with the host clipboard.

Dunno, isn't this overdoing the secure mode thing a bit? What "security" would disabling the clipboard bring—I'm unsure, I can't see any benefits.

My understanding is the whole secure mode thing was a way to prevent mounting parts of the filesystem on public-facing installations, such as when Staging runs on archive.org.

So maybe just drop this?

FeralChild64:
> Dunno, isn't this overdoing the secure mode thing a bit? What "security" would disabling the clipboard bring—I'm unsure, I can't see any benefits.
> 
> My understanding is the whole secure mode thing was a way to prevent mounting parts of the filesystem on public-facing installations, such as when Staging runs on archive.org.
> 
> So maybe just drop this?

The idea was to prevent a "data leak" if something which should not be shown to untrusted users is accidentally left in the host OS clipboard. This is actually just few lines of code. AFAIK DOSBox-X does something similar.

What does the rest of the team think about it?

interloper98:
> The idea was to prevent a "data leak" if something which should not be shown to untrusted users is accidentally left in the host OS clipboard. This is actually just few lines of code. AFAIK DOSBox-X does something similar.
> 
> What does the rest of the team think about it?

Sounds like DOSBox's "secure mode" is for users who don't fully trust their software environment. Not unreasonable given MS-DOS lead to Windows 95, which spawned the lucrative Malware / Spyware industry. 

_"Hey, why is my text editor plugin reading my personal directories and uploading hundreds of MB to a foreign IP address?"_ :skull:

So if there's already a secure mode conf setting, we might as well block clipboard content harvesting in either direction. I've seen bank employees using something like DOS on their Windows 11 PCs. Not sure if it's DOSBox.. but who knows!

johnnovak:
Yeah, I guess it doesn't hurt, after all.

> Not unreasonable given MS-DOS lead to Windows 95, which spawned the lucrative Malware / Spyware industry.

I'm quite certain Linux, macOS, and all other alternative OSes mostly enjoy security by obscurity in this regard 😏

include/clipboard.h
```diff
@@ -0,0 +1,37 @@
+/*
+ *  SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ *  Copyright (C) 2023-2024  The DOSBox Staging Team
```
johnnovak, line 4: 2024-2024

include/unicode.h
```diff
@@ -92,4 +92,7 @@ std::string lowercase_dos(const std::string& str, const uint16_t code_page);
 std::string uppercase_dos(const std::string& str);
 std::string uppercase_dos(const std::string& str, const uint16_t code_page);
 
+// Check if two code pages are the same
+bool is_code_page_duplicate(const uint16_t code_page1, const uint16_t code_page2);
```
johnnovak, line 96: `is_code_page_equal` ?

src/dos/program_clip.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "checks.h"
+#include "clipboard.h"
+#include "dos_inc.h"
+#include "math_utils.h"
+#include "program_more_output.h"
+
+CHECK_NARROWING();
+
+// Maximum allowed size of data to copy to host clipboard
+constexpr uint32_t MaxFileSize = 16 * 1024 * 1024;
```
johnnovak, line 32: `int64_t` would be better. Avoid unsigned types unless dealing with hardware registers and binary protocols and such.

src/dos/program_clip.cpp
```diff
@@ -0,0 +1,124 @@
...
+	MSG_Add("PROGRAM_CLIP_HELP_LONG",
+	        "Copy/paste text to/from clipboard.\n"
+	        "\n"
+	        "Usage:\n"
+	        "  [color=light-cyan]COMMAND[reset] | [color=light-green]clip[reset]\n"
+	        "  [color=light-green]clip[reset] < [color=light-cyan]FILE[reset]\n"
+	        "  [color=light-green]clip[reset]\n"
+	        "\n"
+	        "Notes:\n"
+	        "  - If no input is provided, the command prints out the clipboard content.\n"
```
johnnovak, line 114: "clipboard's content"

src/dos/program_clip.cpp
```diff
@@ -0,0 +1,124 @@
...
+	        "  - This command is only for handling text data, not binary data.\n"
+	        "\n"
+	        "Examples:\n"
+	        "  [color=light-cyan]dir[reset] | [color=light-green]clip[reset]               ; copies the list of files to the clipboard\n"
+	        "  [color=light-green]clip[reset] < [color=light-cyan]Z:\\AUTOEXEC.BAT[reset]   ; copies the file to the clipboard\n"
+	        "  [color=light-green]clip[reset]                     ; displays the clipboard content\n");
```
johnnovak, line 120: "clipboard's content"

src/dos/program_clip.cpp
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Copy text to the clipboard
+	CLIPBOARD_CopyText(input_data);
+}
+
+void CLIP::AddMessages()
+{
+	MSG_Add("PROGRAM_CLIP_HELP_LONG",
+	        "Copy/paste text to/from clipboard.\n"
```
johnnovak, line 106: This is a bit of a weird sentence. You could explain this better in 2-3 lines. You can take a few hints of the original CLIP.EXE command's help:

https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/clip
https://ss64.com/nt/clip.html

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "unicode.h"
+
+#include <SDL.h>
+
+CHECK_NARROWING();
+
+static struct Clipboard {
+	// Content - text
+	uint16_t text_code_page = 0;
+	// Since DOS -> UTF-8 -> DOS string conversion won't necessary produce
```
johnnovak, line 36: necessarily

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+	// the same output as original DOS encoded text (some code pages has
```
johnnovak, line 37: as the original
DOS-encoded

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+	// multiple blank/unused characters, etc.), in case the text comes from
+	// the DOS side we keep both DOS and UTF-8 encoded strings - this way
```
johnnovak, line 39: the DOS side, (comma missing)

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+	// the DOS can get precisely the same content as it has sent to the
```
johnnovak, line 40: DOS -> DOSBox

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+} clipboard;
+
+// ***************************************************************************
+// Host clipboard synchronization
+// ***************************************************************************
+
+static bool text_equals(const std::string& str_1, const std::string& str_2)
+{
+	// String compare, insensitive to end of the line encoding; needed
+	// because we convert the end of line endings to host native before
```
johnnovak, line 66: "end of line endings" -> "end-of-line markers" or something?
host native -> host-native

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+	// copying the text to the host clipboard
```
johnnovak, line 67: to -> into

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+	};
+
+	while ((index_1 < str_1.size()) && (index_2 < str_2.size())) {
+		if (get_character(str_1, index_1) != get_character(str_2, index_2)) {
+			return false;
+		}
+	}
+
+	return (index_1 == str_1.size()) && (index_2 == str_2.size());
+}
```
johnnovak, line 92: We should have unit tests for this function.

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+static void maybe_fetch_text_from_host()
+{
+	if (control->SecureMode()) {
+		// We won't need the host content for anything more
+		clipboard.host_text_utf8.clear();
+		// Do not fetch anything from host clipboard in secure mode
```
johnnovak, line 99: from the

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+		return;
+	}
+
+	bool host_content_changed = false;
```
johnnovak, line 103: `has_host_content_changed`?

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+		// intact for proper synchronization with host OS clipboard
+	}
+
+} clipboard;
+
+// ***************************************************************************
+// Host clipboard synchronization
+// ***************************************************************************
+
+static bool text_equals(const std::string& str_1, const std::string& str_2)
```
johnnovak, line 63: `is_text_equal`?

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+static void maybe_push_text_to_host()
+{
+	if (control->SecureMode()) {
+		// We won't need the host content for anything more
+		clipboard.host_text_utf8.clear();
+		// Do not push anything to host clipboard in secure mode
+		return;
+	}
```
johnnovak, line 133: push -> copy?

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+	// Convert end-of-the-line encoding to host native
```
johnnovak, line 135: "end-of-line markers"

src/gui/clipboard.cpp
```diff
@@ -0,0 +1,220 @@
...
+		} else if (character == '\n') {
+			converted += newline;
+			if (index + 1 < clipboard.text_utf8.size() &&
+			    clipboard.text_utf8[index + 1] == '\r') {
+				++index;
+			}
+		} else {
+			converted.push_back(character);
+		}
+	}
```
johnnovak, line 161: It would be better to extract the conversion part into the function and add some unit tests. Also make the line ending a parameter of the function so you can add tests for both line ending styles.

johnnovak:
Nice work @FeralChild64. A couple of observations:

- What happens in programs that use Ctrl+V for something? I'm thinking of DOS games and apps in particular, but also Windows 3.x. Would this interfere with the Windows 3.x clipboard.
- Following from the previous point, this should be a mappable action. Users might want to map this to something else if they don't want it to interfere with the Windows 3.x clipboard, perhaps? Or unmap it altogether if they don't care about it as it might interfere with some games. But mapping it to Ctrl-V by default would be still fine; it should be fine with most games.
- Some parts of the code do character-level string mangling and conversions. These are trivially testable in unit tests after some minor code restructurings, so let's do that!

include/clipboard.h
FeralChild64: Changed - it is actually based on the code I wrote the last year, that's why it was 2023-2024.

include/unicode.h
FeralChild64: Changed.

src/dos/program_clip.cpp
FeralChild64: Changed.

src/dos/program_clip.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/dos/program_clip.cpp
FeralChild64: Exactly - this is being compared against value from DOS API which is `uint32_t`. I don't want the compiler to tell me `warning: comparison between signed and unsigned` or anything like this.

src/dos/program_clip.cpp
FeralChild64: I can't use more than 1 line. In fact, I can't even use the full line, as the description has to fit in the `help /all` command output:

Help

FeralChild64: `Copy text to the clipboard or retrieve the clipboard content.` is probably the best that still fits - changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

src/gui/clipboard.cpp
FeralChild64: Changed.

FeralChild64:
>     * What happens in programs that use Ctrl+V for something? I'm thinking of DOS games and apps in particular, but also Windows 3.x. Would this interfere with the Windows 3.x clipboard?

Exactly the same would happen as before my change. The _Ctrl+V_ is not implemented as a global shortcut, handling is limited to our `COMMAND.COM` implementation, it won't interfere with anything - just like handling the _Tab_ (for file name autocomplete) does not interfere with tabulation in text editors.

>     * Some parts of the code do character-level string mangling and conversions. These are trivially testable in unit tests after some minor code restructurings, so let's do that!

Yes, I'll do the unit tests.

FeralChild64:
@Torinde There is currently no clipboard synchronization with Windows 3.1x (or Windows 9x) , no AUTOTYPE-like universal pasting, etc. Just what is stated - the `clip` command and _Ctrl+V_ in the command line (not even in the command line visible from the _DOS Navigator_ or _Norton Commander_), nothing more.

Some improvements are likely to appear in the future - I stated _Initial_ for a reason - but they need deeper investigation.
For example, both _VirtualBox_ and _VMware_ provide an API for clipboard synchronization with the host.
Or, there exists some API allowing DOS applications to interact with Windows clipboard, _DOSBox-X_ implements it - but, if I implement it, it will have to be provided to compatible applications only; we don't want games or installers erroneously thinking they are running under Windows.

src/gui/clipboard.cpp
FeralChild64: Done, unit test created.

src/gui/clipboard.cpp
FeralChild64: Done, unit test created.

src/dos/program_clip.cpp
johnnovak: Yeah, that's a lot better, thanks. Ok, wasn't aware this was the short description 😄

johnnovak:
Thanks for the explanation @FeralChild64, that answered my questions. I kinda glossed over the "command line only" part in the release notes. Probably I'll put a warning there in the final notes that this is for the DOS prompt only and doesn't even work in NC, VC or DN.

tests/string_utils_tests.cpp
```diff
@@ -519,4 +519,131 @@ TEST(ReplaceAll, Valid)
...
+	EXPECT_FALSE(is_text_equal(s2_win32, s1_mixed));
+	EXPECT_FALSE(is_text_equal(s3_posix, s1_mixed));
+	EXPECT_FALSE(is_text_equal(s3_win32, s1_mixed));
+
+	// Compare strings without newline characters
+	EXPECT_TRUE(is_text_equal("FooBar", "FooBar"));
+	EXPECT_FALSE(is_text_equal("FooBar", "BarFoo"));
+	EXPECT_FALSE(is_text_equal("FooBar", "FooBarBaz"));
+	EXPECT_FALSE(is_text_equal("FooBarBaz", "FooBar"));
+}
```
johnnovak, line 647: Very nice test coverage, man! 🥇 💯

johnnovak (approved):
Looks very good now @FeralChild64, thanks for addressing all the comments.

src/dos/program_clip.cpp
FeralChild64: OK, changed.

PR #4040, opened 2024-11-02
Simplify `mididevice` and `midiconfig` settings

johnnovak:
# Description

- After removing support for OSS MIDI on Linux, we now only have one MIDI interface per platform to send MIDI data out of Staging. Therefore, it makes zero sense to keep the `win32`, `alsa`, and `coremidi` settings for `mididevice`; these have been replaced by the single platform-agnostic `port` setting. This makes the MIDI config a little bit more portable (although the corresponding `mididevice` setting to select a MIDI port explicitly could be platform-specific, still).

- `coreaudio` on macOS stays as is because it uses the built-in SoundFont player of macOS via the Core Audio API. Technically, it uses a FluidSynth-like SF2 player built into the OS, so it's different than the `port` setting which just sends the MIDI data out, without having any knowledge of its final destination, from the emulator's point of view.

- We had a weird super confusing feature where you could set `mididevice = auto`, then select FluidSynth or MT-32 with `midiconfig = fluidsynth` and `midiconfig = mt32`, respectively. This is a pretty silly non-feature—why would you want to do that? Just set `mididevice = mt32` or `fluidsynth` and be done with it... It only muddies the waters for no good reason. Well, the good news is it never worked 😎 , or at least going back to 0.79.0... Probably some legacy cruft, so the only thing I had to do was to remove it from the `midiconfig`'s description 🤷🏻 Thanks to @Torinde for discovering this 🎉 

- The `mididevice` and `midiconfig` descriptions have been extensively revised; everybody should get platform-specific instructions now (like, why should a Linux user be shown Windows-specific info, or vice versa?)

   Here are the new macOS descriptions:

image

image

## Related issues

https://github.com/dosbox-staging/dosbox-staging/discussions/4031

# Release notes

### MIDI configuration changes

- The `auto` option for the `mididevice` setting has been renamed to `port`. This makes a lot more sense as this option is for routing the MIDI output to a MIDI port of the host operating system's standard MIDI interface.

- The OS-specific `win32` (Windows), `coremidi` (macOS), and `alsa` (Linux) settings for `mididevice` have been deprecated. They are still accepted with a warning and are mapped to `mididevice = port`.

# Manual testing

## General

- `mididevice = port` is the default and behaves like `auto` previously
- Deprecation warnings & defaulting works:
    
image

- Defaulting to the `mididevice = port` if the requested MIDI device cannot be opened works:

    
image

    
image

## Windows

- Tested a few games by sending MIDI data to SCVA 
- Tested a few MT-32 games just to be sure
- `mididevice = win32` defaults to `port` with a warning, which then uses the `win32` driver

image

## macOS

- Core Audio works:

image

image

- Core MIDI works:

image

image

image

image

# Checklist

I have:

### DIFF
--- a/src/midi/midi.cpp
+++ b/src/midi/midi.cpp
@@ -93,7 +93,7 @@ static std::unique_ptr<MidiDevice> create_device(const std::string& name,
 {
 	using namespace MidiDeviceName;
 
-	// Internal MIDI synths
+	// Built-in MIDI synths
 #if C_FLUIDSYNTH
 	if (name == FluidSynth) {
 		return std::make_unique<MidiDeviceFluidSynth>();
@@ -128,29 +128,6 @@ static std::unique_ptr<MidiDevice> create_device(const std::string& name,
 #endif
 
 	// Device not found
-	return {};
-}
-
-static std::vector<const char*> auto_device_candidates = {MidiDeviceName::Alsa,
-                                                          MidiDeviceName::CoreAudio,
-                                                          MidiDeviceName::CoreMidi,
-                                                          MidiDeviceName::Win32};
-
-static std::unique_ptr<MidiDevice> try_create_auto_device(const std::string& config)
-{
-	for (const auto device_name : auto_device_candidates) {
-		try {
-			if (auto device = create_device(device_name, config); device) {
-				return device;
-			}
-			// nullptr means the device is not supported on
-			// the current platform; try the next one
-
-		} catch (const std::runtime_error& ex) {
-			// error opening device; try the next one
-			continue;
-		}
-	}
 	return {};
 }
 
@@ -639,8 +616,11 @@ static std::string get_mididevice_setting()
 	return get_midi_section()->Get_string("mididevice");
 }
 
+static auto MidiDevicePortPref    = "port";
+static auto DefaultMidiDevicePref = MidiDevicePortPref;
+
 // We'll adapt the RtMidi library, eventually, so hold off any substantial
-// rewrites on the MIDI stuff until then to unnecessary work.
+// rewrites of the MIDI stuff until then to avoid unnecessary work.
 class MIDI final {
 public:
 	MIDI()
@@ -669,10 +649,21 @@ public:
 
 		trim(midiconfig_prefs);
 
-		if (device_pref == "auto") {
-			// Use the first working device
-			midi.device = try_create_auto_device(midiconfig_prefs);
-
+		if (device_pref == MidiDevicePortPref) {
+			// Use system-level MIDI interface of the host OS
+#if C_COREMIDI
+			// macOS
+			midi.device = create_device(MidiDeviceName::CoreMidi,
+			                            midiconfig_prefs);
+#elif defined(WIN32)
+			// Windows
+			midi.device = create_device(MidiDeviceName::Win32,
+			                            midiconfig_prefs);
+#elif C_ALSA
+			// Linux
+			midi.device = create_device(MidiDeviceName::Alsa,
+			                            midiconfig_prefs);
+#endif
 		} else {
 			midi.device = create_device(device_pref, midiconfig_prefs);
 		}
@@ -766,10 +757,9 @@ static void midi_init([[maybe_unused]] Section* sec)
 
 		} catch (const std::runtime_error& ex) {
 			const auto mididevice_pref = get_mididevice_setting();
-			if (mididevice_pref == "auto") {
+			if (mididevice_pref == MidiDevicePortPref) {
 				LOG_WARNING(
 				        "MIDI: Error opening device '%s'; "
-				        "MIDI auto-discovery failed, "
 				        "using 'mididevice = none' and disabling MIDI output",
 				        mididevice_pref.c_str());
 
@@ -782,11 +772,14 @@ static void midi_init([[maybe_unused]] Section* sec)
 			} else {
 				// If 'mididevice' was set to a concrete value
 				// and the device could not be initialiased,
-				// we'll try 'auto' as a fallback.
-				LOG_WARNING("MIDI: Error opening device '%s'; using 'auto'",
-				            mididevice_pref.c_str());
+				// we'll try 'port' as a fallback.
+				LOG_WARNING("MIDI: Error opening device '%s'; using '%s'",
+				            mididevice_pref.c_str(),
+				            DefaultMidiDevicePref);
 
-				set_section_property_value("midi", "mididevice", "auto");
+				set_section_property_value("midi",
+				                           "mididevice",
+				                           MidiDevicePortPref);
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


johnnovak:
@interloper98 @weirddan455 It would be great if you guys could regression test this on Linux with external Munt or something.

FeralChild64:
Did some test on Linux:
- installed _Timidity++_, started the daemon with `timidity -iA` command
- `config -set mididevice=port` - result:
```
2024-11-04 00:12:29.778 | MIDI:ALSA: Connected to MIDI port 128:0 - TiMidity
2024-11-04 00:12:29.778 | MIDI: Opened device 'alsa'
```
- started _DOOM_ configured to General Midi, tried to play MIDI file using the _GSMIDI_ player - both worked correctly

johnnovak:
> Did some test on Linux:
> 
>     * installed _Timidity++_, started the daemon with `timidity -iA` command
> 
>     * `config -set mididevice=port` - result:
> 
> 
> ```
> 2024-11-04 00:12:29.778 | MIDI:ALSA: Connected to MIDI port 128:0 - TiMidity
> 2024-11-04 00:12:29.778 | MIDI: Opened device 'alsa'
> ```
> 
>     * started _DOOM_ configured to General Midi, tried to play MIDI file using the _GSMIDI_ player - both worked correctly
> 
> 
> [edit] The `midiconfig` setting also seems to work as expected.

Great! Thank you so much for testing the Linux behaviour @FeralChild64 ❤️ 💯

AnonTet:
Just tried this on windows v0.83.0-alpha (9bfb10) and got this:
2024-11-04 12:00:01.493 | **CONFIG: Invalid 'mididevice' setting: 'port', using 'auto'**
2024-11-04 12:00:01.498 | MPU-401: Running in intelligent mode on ports 330h and 331h
2024-11-04 12:00:01.506 | MIDI: Using delayed SysEx processing
2024-11-04 12:00:01.506 | MIDI:WIN32: Selected output device CoolSoft MIDIMapper
2024-11-04 12:00:01.564 | MIDI: Opened device 'win32'

src/midi/midi.cpp
```diff
@@ -801,97 +794,95 @@ void init_midi_dosbox_settings(Section_prop& secprop)
...
+	        "coreaudio",
+	        "  - When using 'mididevice = coreaudio', this setting specifies the SoundFont\n"
+	        "    to use. You must use the absolute path of the SoundFont file.");
 
-	str_prop->SetOptionHelp("alsa",
-	                        "  - When using ALSA, use the Linux command 'aconnect -l' to list all open\n"
-	                        "    MIDI ports and select one (e.g. 'midiconfig = 14:0' for sequencer\n"
+	str_prop->SetOptionHelp("linux",
+	                        "  - When using 'mididevice = port', use the Linux command 'aconnect -l' to list\n"
+	                        "    all open MIDI ports and select one (e.g., 'midiconfig = 14:0' for sequencer\n"
```
weirddan455, line 856: I would suggest leaving this config option named ALSA.  Changing this breaks backwards compatibility with old config files and is less clear as to the interface we're using.  There are other MIDI interfaces on Linux such as JACK.  It's also inconsistent with what we're doing on Mac (the config option is named "coreaudio" not "macosx".)

weirddan455:
>  Just tried this on windows v0.83.0-alpha (9bfb10) and got this:

@AnonTet  Are you sure you're running the build from this PR?  That's not the commit number from this PR and I don't even see that commit in the tree at all for that matter so I don't know where you got that build from.

weirddan455:
>  I don't want to mislead you but I think it was from this link: https://github.com/dosbox-staging/dosbox-staging/actions/runs/11657645430

That build is from a completely different branch.  It doesn't include the changes mentioned in this PR.

AnonTet:
> > I don't want to mislead you but I think it was from this link: https://github.com/dosbox-staging/dosbox-staging/actions/runs/11657645430
> 
> That build is from a completely different branch. It doesn't include the changes mentioned in this PR.

Again, my bad. I was under the impression they were cumulative. I got it now, sorry for the confusion.

src/midi/midi.cpp
johnnovak: Reflecting to your concrete points first @weirddan455:

> I would suggest leaving this config option named ALSA. Changing this breaks backwards compatibility with old config files and is less clear as to the interface we're using.

No, it does not break backwards compatibility. If you set `mididevice = alsa`, that would revert to `mididevice = port` via the deprecation mechanism. Also, you can set it to `mididevice = blah` too, which would revert to `port`. Try it and see! 😄 

> There are other MIDI interfaces on Linux such as JACK.

AFAIK, JACK is built on top of ALSA. Virtually zero normal users would use JACK as it's a complicated beast. I used it one (on macOS, though), and yeah it was pretty horrible. It seems it's universally hated:

https://www.reddit.com/r/linuxaudio/comments/gmjl9s/jack_and_its_remarkable_ability_to_kill_creativity/
https://linuxmusicians.com/viewtopic.php?t=19258
https://linuxmusicians.com/viewtopic.php?t=24344
https://linuxmusicians.com/viewtopic.php?t=25557

To me, this means ALSA is the de-facto MIDI interface on Linux, and it's extremely unlikely we'll ever add JACK support, especially because they're not mutually exclusive. ALSA would always work for JACK users.

> It's also inconsistent with what we're doing on Mac (the config option is named "coreaudio" not "macosx".)

No, read this section more carefully. `linux`, `windows_or_macos`, `mt32` here are _not_ actual setting names but just some symbolic constants you pass into `SetEnabledOptions`. This is a recent feature that enables us conditionally include just some of these OS-specific help texts with define macros, _but_ they will all show up in the translations because they're all added with `SetOptionHelp`. 

---

Okay, so my general comments. Please read my original description carefully and check out the screenshots that show the settings' help because I have a feeling you haven't fully understood what I'm doing here.

Like I said, we have a one-to-one mapping of MIDI interfaces and operating systems now. As you've pointed out, JACK could break this, at least at the first glance, but I described above, I don't think that's the case.

I'm not dying to get the new `port` setting in, but I thought it's neat and better than the vague `auto` setting.

If you think having `alsa` and `jack` (in the future) for Linux is important, we could rename `port` back to `auto`, then `auto` would mean `alsa` on Linux, or you could set `alsa` or `jack` explicitly. But the `win32` and `coremidi` options would go away; there's no other way for system-level MIDI on Windows and macOS so `auto` could always mean only one thing.

I just don't really like `auto` when what it really means is "use the system-level MIDI interface".

weirddan455: Oh, I missed that "linux" here was just an internal name.  I built it and checked how the config looks and it looks good.  Agree about auto being confusing.  We can always change internal naming later if/when needed.

weirddan455 (approved):
Looks good

johnnovak:
> Looks good

Thanks for the review, man 😎

AnonTet:
Tested the main branch and it works on windows 11 latest patch without any glitch as far as I can tell.
Even when using midiloop or midimapper to load vsti plugins it doesn't seem to affect anything.

PR #4055, opened 2024-11-12
Let unchangeable settings be queued and used on restart

### LINKED ISSUE #2565: Changing machine type or mem settings with config command will take effect on the next restart with config -r

interloper98:
# Description

This PR lets unchangeable values hold a (new) queued value that's used on restart (`CONFIG -r` or via hotkey), without having to write a temporary config file.

## Related issues

Fixes #2565 (thanks for the tip @FeralChild64 )

# Release notes

DOSBox Staging will now queue changes for unchangeable settings such as `machine` and `memsize`, and use them when restarting with `CONFIG -r` or via hotkey. 

For example, to change the machine type to CGA and use 1 MB of RAM:

```shell
Z:\> machine cga
Z:\> memsize 1
z:\> config -r
```

# Manual testing

Confirmed queued changes are properly written to conf files:
- `CONFIG -wcd`
- `CONFIG -wc`
- `CONFIG -wc custom.conf`

Confirmed queued changes are used on restart.

Confirmed that multiple restarts and changes work:
- Changing the same properly doesn't conflict (ie: machine cga, then ega, then pcjr)
- Changes made during a prior restart are correctly held in subsequent restart. For example, setting `memsize 1` persists across multiple machine type restarts.

# Checklist

I have:

### DIFF
--- a/include/setup.h
+++ b/include/setup.h
@@ -430,6 +430,16 @@ public:
 	Property* Get_prop(int index);
 	Property* Get_prop(const std::string_view propname);
 
+	const_it begin() const
+	{
+		return properties.begin();
+	}
+
+	const_it end() const
+	{
+		return properties.end();
+	}
+
 	int Get_int(const std::string& _propname) const;
 
 	std::string Get_string(const std::string& _propname) const;
--- a/include/setup.h
+++ b/include/setup.h
@@ -173,6 +173,10 @@ public:
 		return default_value;
 	}
 
+	void SetQueueableValue(std::string&& value);
+
+	const std::optional<std::string>& GetQueuedValue() const;
+
 	bool IsRestrictedValue() const
 	{
 		return !valid_values.empty();
@@ -213,6 +217,7 @@ protected:
 	std::vector<Value> valid_values                        = {};
 	std::vector<std::string> enabled_options               = {};
 	std::map<Value, Value> deprecated_and_alternate_values = {};
+	std::optional<std::string> queueable_value             = {};
 	bool is_positive_bool_valid                            = false;
 	bool is_negative_bool_valid                            = false;
 
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -652,7 +652,7 @@ void DOSBOX_Init()
 	secprop->AddInitFunction(&PAGING_Init);
 	secprop->AddInitFunction(&MEM_Init);
 
-	pint = secprop->Add_int("memsize", when_idle, 16);
+	pint = secprop->Add_int("memsize", only_at_start, 16);
 	pint->SetMinMax(MEM_GetMinMegabytes(), MEM_GetMaxMegabytes());
 	pint->Set_help(
 	        "Amount of memory of the emulated machine has in MB (16 by default).\n"
--- a/src/misc/programs.cpp
+++ b/src/misc/programs.cpp
@@ -906,15 +906,10 @@ void CONFIG::Run(void)
 					return;
 				}
 
-				const auto* property = tsec->Get_prop(pvars[1]);
+				auto* property = tsec->Get_prop(pvars[1]);
 
 				if (!property) {
 					WriteOut(MSG_Get("PROGRAM_CONFIG_PROPERTY_ERROR"), pvars[1].c_str());
-					return;
-				}
-
-				if (property->GetChange() == Property::Changeable::OnlyAtStart) {
-					WriteOut(MSG_Get("PROGRAM_CONFIG_NOT_CHANGEABLE"), pvars[1].c_str());
 					return;
 				}
 
@@ -935,6 +930,13 @@ void CONFIG::Run(void)
 
 				if (value.empty()) {
 					WriteOut(MSG_Get("PROGRAM_CONFIG_SET_SYNTAX"));
+					return;
+				}
+
+				// If the value can't be set at runtime then queue it and let the user know
+				if (property->GetChange() == Property::Changeable::OnlyAtStart) {
+					WriteOut(MSG_Get("PROGRAM_CONFIG_NOT_CHANGEABLE"), pvars[1].c_str());
+					property->SetQueueableValue(std::move(value));
 					return;
 				}
 
--- a/src/misc/programs.cpp
+++ b/src/misc/programs.cpp
@@ -215,7 +215,7 @@ void Program::ChangeToLongCmd()
 	}
 
 	// Clear so it gets even more save
-	full_arguments.assign(""); 
+	full_arguments.assign("");
 }
 
 bool Program::SuppressWriteOut(const char* format)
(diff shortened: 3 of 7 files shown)


src/dosbox.cpp
```diff
@@ -652,7 +652,7 @@ void DOSBOX_Init()
 	secprop->AddInitFunction(&PAGING_Init);
```
johnnovak, line 652: Typo in the commit message (first line, "inidicate")

src/misc/setup.cpp
```diff
@@ -747,6 +747,17 @@ std::vector<Value> Property::GetDeprecatedValues() const
...
+void Property::SetQueueableValue(std::string&& value)
+{
+	assert(!value.empty());
+	queueable_value = std::move(value);
+}
+
+const std::optional<std::string>& Property::GetQueuedValue() const
+{
+	return queueable_value;
+}
```
johnnovak, line 759: Nice, I'm planning to mark these "queued" changes in the GUI as well when I get to it.

src/misc/programs.cpp
```diff
@@ -1145,7 +1146,8 @@ void PROGRAMS_Init(Section* sec)
 	MSG_Add("CONJUNCTION_AND", "and");
 
 	MSG_Add("PROGRAM_CONFIG_NOT_CHANGEABLE",
-	        "Setting '%s' is not changeable at runtime.\n");
+	        "The '%s' setting can't be changed at runtime, but will be written to a\n"
```
johnnovak, line 1149: "written to the"

johnnovak: Don't you wanna also mention that these queued changes will take effect when restarting with `CONFIG -r` or via the reset hotkey?

src/misc/setup.cpp
```diff
@@ -1573,6 +1573,55 @@ bool Config::ParseConfigFile(const std::string& type,
...
+					++it;
+				}
+			}
+
+			// Add the new arguments with the queued value
+			args.emplace_back("--set");
+			args.emplace_back(sec_prop + *queued_value);
+		}
+	}
+}
```
johnnovak, line 1623: Right... I think it would be better to have each config setting have a "slot" for its "future queued value" instead of having this separate vector. I'm just thinking of the GUI; I'd like to be able to call `HasQueuedSetting()` at a per-property level.

But many changes might be necessary for the GUI stuff, so this is fine for now. Even if you reimplemented it like I explained, I might need to thoroughly refactor the entire config subsystem... So probably this is fine until then.

johnnovak:
Nice release notes, man 🎖️ 

I will remove the "as mentioned in the README file" because that's severely outdated and incomplete anyway. I will delete it the moment we'll have the documentation up on the website 😄

johnnovak:
@interloper98 Actually, the more I'm thinking about this I think we should just use the queued values for the next restart.

The conf properties should always reflect the current state, and the config writet should always write thw current config, period.

This would work consistently with the GUI too.

src/misc/setup.cpp
interloper98: Yup, this PR adds this slot; the config property class now has a std::optional queueable_value private member, with Set and Get function.

This snippet you've flagged here modifies the to-be-used CLI arguments with the queued values. (And the CLI args are just a string.. so that's why it's appending). 

The restart action sets up the new process + argv, launches the process, and then the current dosbox process dies.

 So the only way to convey information (the new conf settings) to the new PID is via argv (or some temp file, which is much uglier 😅 )

interloper98:
Sounds reasonable @johnnovak ; I'll knock out using the queued values when writing the confs, and only use them for the restart.

src/misc/setup.cpp
johnnovak: Right, I kinda missed that part 😄

FeralChild64:
Isn't it this ticket? https://github.com/dosbox-staging/dosbox-staging/issues/2565

src/dosbox.cpp
interloper98: Thanks, fixed.

src/misc/programs.cpp
FeralChild64: I suggest to reformat it to something like:

_The '%s' setting can't be changed at runtime._`[newline]`
_Additional description. Lorem ipsum dolor sit amet, consectetur adipiscing elit._

This way settings with longer names, like `vga_render_per_scanline`, won't cause a "line break" in an unexpected place.

interloper98:
> Isn't it this ticket? #2565

Thanks! That's definitely a 100% match (but first I've seen it :slightly_smiling_face: ; might have been this one when I learned about config -r https://github.com/dosbox-staging/dosbox-staging/issues/2565)

src/misc/programs.cpp
interloper98: Thanks, yes - added.

interloper98:
Quick recording of it in action:

https://github.com/user-attachments/assets/db9de03a-0b64-4ba9-9ff0-67bffa61c5a6

include/setup.h
```diff
@@ -173,6 +173,10 @@ class Property {
 		return default_value;
 	}
 
+	void SetQueueableValue(std::string&& value);
+
+	const std::optional<std::string>& GetQueuedValue() const;
+
```
johnnovak, line 179: I think you'll need to adjust the commit message and remove references to writing the "future state" of the config.

src/misc/programs.cpp
```diff
@@ -1147,7 +1147,8 @@ void PROGRAMS_Init(Section* sec)
 	MSG_Add("CONJUNCTION_AND", "and");
 
 	MSG_Add("PROGRAM_CONFIG_NOT_CHANGEABLE",
-	        "Setting '%s' is not changeable at runtime.\n");
+	        "[color=yellow]The '%s' setting can't be changed at runtime.[reset]\n"
+	        "However, it will be applied on restart (CONFIG -r) or via hotkey.\n");
```
johnnovak, line 1151: Should be this, no?

```
However, it will be applied on restart (CONFIG -r or via hotkey).
```

src/misc/setup.cpp
johnnovak: Ok... It has finally dawned on me why this thing is needed. We don't restart "internally", but we literally quit and restart the process. Got it 😄

src/misc/programs.cpp
interloper98: fixed

include/setup.h
interloper98: Good catch. Removed!

johnnovak (approved):
Looking good @interloper98, thanks for addressing all my comments 🚀

interloper98:
Thanks for the reviews @johnnovak and @FeralChild64. Merging!

johnnovak:
@interloper98 Do you think it's backport-worthy (and safe)? I think so.

interloper98:
I think so, too. It only touches existing functionality at the point when the restart happens, and it's pretty limited (and safe) in what it does.

PR #3951, opened 2024-09-23
Rework locale settings and autodetection

FeralChild64:
# Description

For the user-visible changes, see the Release Notes draft - in this section I'll only describe technical details.

The PR is, unfortunately, huge - but I can't split it into many parts, this change is all-or-nothing. There were numerous 'smart' attempts to synchronize locale with host (like check the configured country and set everything else based on it), but they were all too simplified - that's why I decided to go the _let's rewrite it_ route.

Fortunately, quite a lot of lines of code are due to various host-to-guest conversion tables, metadata, or the country information data moved to a separate file in order not to clutter the file with the actual business logic.

And we do need these conversion tables - you can't just blindly attempt to convert the _ISO 3166-1_ country/territory code (where, for example, _tt_ means _Trinidad and Tobago_) to the _FreeDOS_ keyboard layout (where _tt_ means _Tatar_); it might work well for _France_, _Germany_, or _Poland_, but the world does not end here :)

All the detected/loaded locale settings are logged, with some information about values from the host OS which were used for detection. Example:

```
2024-09-28 20:36:43.878 | DOS: Country detected from 'LC_TELEPHONE=pl_PL.UTF-8'
2024-09-28 20:36:43.878 | DOS: Loading native locale for country 48 - 'Poland'
2024-09-28 20:36:43.878 | DOS: Using numeric/time/date/currency format detected from host OS
2024-09-28 20:36:43.878 | DOS: Keyboard layout and code page detected from 'KDE: pl'
2024-09-28 20:36:43.878 | DOS: Loaded code page 668
2024-09-28 20:36:43.878 | DOS: Loaded keyboard layout 'pl' - 'Polish (programmers, QWERTY/phonetic)'
```

#### Limitations:
- Issues https://github.com/dosbox-staging/dosbox-staging/issues/1925 and https://github.com/dosbox-staging/dosbox-staging/issues/3233 are still not fixed; to fix loading keyboard layouts, I would probably need to rewrite the code for loading KEYBOARD.SYS and friends; out of scope, this PR is already huge.
- To properly fix point 2. from https://github.com/dosbox-staging/dosbox-staging/issues/3498, we would need to seriously modify `messages.cpp`, as the current design does not allow reverting back to English - out of scope of this PR.
- There is no keyboard layout autodetection for _macOS_ yet; the system seems to store keyboard layout settings in the `com.apple.HIToolbox.plist`, I am not yet sure how to read it from the C++
[...]

PR #4069, opened 2024-11-24
Keyboard layout / code page autodetection tweaks + macOS support

FeralChild64:
# Description

# Release notes

(no need to update release notes, description from https://github.com/dosbox-staging/dosbox-staging/pull/3951 should be enough)

# Manual testing

# Checklist

I have:


src/misc/host_locale_macos.cpp
```diff
@@ -179,6 +405,198 @@ static std::string get_language_file(std::string& log_info)
...
+	}
+	const auto array_ref = static_cast<CFArrayRef>(array_ptr);
+
+	// Loop over all the input sources
+	const auto sources_count = CFArrayGetCount(array_ref);
+	for (auto index = 0; index < sources_count; ++index) {
+		// NOTE: this code does not recognize logographic (Chinese,
+		// Korean, Japanese, etc.) writing systems nor transliteration
+		// input sources; their configuration is much more complex,
+		// plus we do not support these alphabets nevertheless.
```
johnnovak, line 547: "Anyway" instead of "nevertheless"

src/misc/host_locale_macos.cpp
```diff
@@ -179,6 +405,198 @@ static std::string get_language_file(std::string& log_info)
...
+			continue;
+		}
+		const auto layout_id_ref  = static_cast<CFNumberRef>(layout_id_ptr);
+		if (CFNumberIsFloatType(layout_id_ref) ||
+		    CFNumberGetByteSize(layout_id_ref) > MaxIntSize) {
+		    	continue;
+		}
+		const auto layout_name_ref = static_cast<CFStringRef>(layout_name_ptr);
+
+		// Convert keybaord layout name and identifier to C++ types
```
johnnovak, line 588: Typo: keyboard

src/misc/host_locale_macos.cpp
FeralChild64: OK, changed.

src/misc/host_locale_macos.cpp
```diff
@@ -30,127 +30,353 @@
...
-*/
 
+// ***************************************************************************
+// Detection data
+// ***************************************************************************
+
+// Mapping from Macintosh to DOS keyboard layout. Collected using:
+// 'System Settings' -> 'Keyboard' -> 'Text Input' -> 'Input Sources' settings
+// on macOS 'Sequoia' 15.1.1.
+static const std::map<int64_t, KeyboardLayoutMaybeCodepage> MacToDosKeyboard = {
```
interloper98, line 41: Super minor: suggest `std::unordered_map` for static constant/fixed maps. (write once, read many) because look ups are basically array-like `O(1)`, versus the ordered `std::map`, which does lookups in `O(log N)` where it has to search for the key in a tree using `<` operators.

interloper98: Suggest protecting the formatting of these big static objects with `// clang-format off` and ``// clang-format on` before and after.

src/misc/host_locale_macos.cpp
```diff
@@ -179,6 +405,198 @@ static std::string get_language_file(std::string& log_info)
...
+		const auto layout_name = to_string(layout_name_ref);
+		int64_t layout_id = 0;
+		CFNumberGetValue(layout_id_ref, kCFNumberSInt64Type, &layout_id);
+
+		apple_layouts.emplace_back(layout_id, layout_name);
+	}
+
+	CFRelease(plist_ref);
+	return get_layouts_maybe_codepages(apple_layouts, log_info);
+}
```
interloper98, line 598: Apple has created a hideous disaster with their C/C++ Core Foundation APIs. It has to be deliberate given their otherwise minimal/elegant aesthetic; probably to maximise harm to developers using cross-platform C++ and STL, with the escape hatch to their Swift language. :facepalm:

src/misc/host_locale_posix.cpp
```diff
@@ -200,6 +200,8 @@ static const std::map<std::string, KeyboardLayoutMaybeCodepage> X11ToDosKeyboard
 	{ "cz:qwerty-mac",              { "cz489" }      },
 	{ "cz:qwerty_bksl",             { "cz489" }      },
 	{ "cz:winkeys-qwerty",          { "cz489" }      },
+	{ "sk:qwerty",                  { "cz489" }      }, // Slovakia
```
interloper98, line 203: std::unordered for this one too

src/misc/host_locale_win32.cpp
```diff
@@ -108,6 +108,7 @@ static const std::map<std::string, KeyboardLayoutMaybeCodepage> WinToDosKeyboard
 	// Czech (programmers, QWERTY)
 	{ "00010405", { "cz489" }      }, // Czech (QWERTY)
 	{ "00020405", { "cz489" }      }, // Czech Programmers
+	{ "0001041b", { "cz489" }      }, // Slovak (QWERTY)
```
interloper98, line 111: sdd::unordered

src/misc/host_locale_macos.cpp
FeralChild64: The `// clang-format off` was actually there, but I have removed it by mistake (`on` is still present at the end) - fixed.

OK, changed the `map` type - although it probably won't make much difference here (this is not a performance critical code at all).

src/misc/host_locale_posix.cpp
FeralChild64: Changed.

src/misc/host_locale_win32.cpp
FeralChild64: Changed.

src/misc/host_locale_macos.cpp
FeralChild64: It's, IMHO, not that bad - it's probably just meant to be used in some kind of wrapper classes rather than directly.

At least the API is consistent, doesn't force me to use UTF-16, or anything like this.

johnnovak: @interloper98 Huh? CoreFoundation is well designed and well documented, and it's not really meant to be used directly. C/C++ programmers using it directly for _entire programs_ kinda deserve it, yeah...

All the "ugliness" is due to the limitations of C.

Even C++ is generally a horrific choice for GUI programming. The sensible choice is exactly what Apple has been proposing for decades: only use C for really low system level interfacing, C++ for the really performance intensive, algorithm heavy stuff, then the rest should be done in a pleasant high level language, preferably dynamic, such as ObjC or Swift.

Game coders have been doing the same things for decades (C++ engine controlled by Lua or C# code), many the Adobe products do the same (e.g., Lightroom UI is in Lua), and even non-sucky FOSS projects such as Blender (rendering backend is C++, UI logic is Python).

johnnovak: Yeah, because as i said, it's not meant to be used directly. You don't need "absolute control over memory allocations" for GUI apps. Maybe some low-level embedded UI-lesz software in your car, yes, but not your desktop app. Then again, really mission critical software doesn't normally use dynamic memory management because you cannot reason about memory fragmentation without a GC. Full manual memory management can be _less_ safe than a good GC because if your malloc in your space shuttle control software gives you an error due to memory fragmentation, it's game ovet...

The CF ref counting function are meant to be used by rhe Swift and ObjC stdlibs and SDKs and that's it.

johnnovak: I don't think such things would be hard for people capable designing entire OSes and ecosystems. They did not do it becaus first-class support for writing modern macOS apps in C and C++ was never a top priority or even design consideration.

It's not a question of capability; I'm capable of doing many, many things that I choose not to do 😄 Same goes for every individual and company.

CF is a bridging library. You can read about it's interesting histor here, but it was build as a bridge to aid Cocoa development in ObjC.

https://www.reddit.com/r/swift/comments/40eswd/swift_standard_library_versus_core_foundation/

Let's just conclude this by saying that handling dynamic maps in C is a nightmare, no matter what you do 😄 Same for strings and any data structure really that's not a static one-dimensional array of primitive types 😆

weirddan455: I don't think Apple has ever done C++ APIs.  If you look at the SDL source code, they write their Mac platform code in Objective-C and I assume they have a Swift API as well.

This just looks like a thing you would use to write bindings for another language.  Basically every language has a C FFI so you always need something like this to bridge languages.  I've seen this kind of thing with C++ libraries as well.  Exposing higher level types to C always looks ugly but it's really the only way to make an API that can be portable to other languages.

johnnovak: Nice fine, but I'd just use ObjC like SDL. Introducing 3rd party deps like that is never a good idea if you only need it for a few use cases. I'd rather live with 1000 lines of ugly direct CF code than another layer of abstraction, honestly, and it's questionable how long they're gonna support this, etc.

FeralChild64: For now I can think of only one other direct use of macOS API - physical CD (including audio tracks) support. I intend to investigate this, I have already found some open source code using IOKit for CD access.

But I don’t intend to use/create any generic bindings. Maybe the `CFString` to `std::string` conversion routine will go to `cross.h`, but that’s really it.

PR #4077, opened 2024-11-28
Add non-blocking disk access delays for IO operations

### LINKED ISSUE #3512: Cycles and VGA rendering accuracy dependent crash in Silmarils games
The following Silmarils games crash at startup above a certain cycles setting:

- Deus (1996)
- Ishar 3 (1994)
- Robinson's Requiem (1994)
- Time Warriors (1997)

This is how the crash looks like:

image

The cycles threshold where the crash occurs varies per game and is also dependent on the type of VGA emulation. In SVN and 0.80.1, using the more accurate per-scanline VGA emulation with `machine = vgaonly` drastically lowers the threshold.

These are the rough cycles threshold values for **Deus** on DOSBox SVN & Staging 0.80.1:

#### `machine = svga_s3`
  - `cycles = 42000` -> runs fine
  - `cycles = 45000` -> crash

#### `machine = vgaonly`
  - `cycles = 4000` -> runs fine
  - `cycles = 5000` -> crash

The `vga_render_per_scanline = off` workaround in https://github.com/dosbox-staging/dosbox-staging/pull/3507 effectively reintroduces the previous non-`vgaonly`behaviour where a single frame of VGA output is rendered not per scanline but in four parts. This special flag effectively restores the 0.80.1 `machine = svga_s3` behaviour for these special games.

The proper fix would be understanding the root cause of this cycles and VGA rendering accuracy dependent crash and make the emulation more robust so the games don't crash with per-scanline rendering. Maybe our per-scanline rendering code gets confused and it overwrites some non-video memory memory location with garbage or zeroes, or the game gets triggered by some non-authentically emulated VGA register value combination at some point and goes off the rails... Probably debugging the game's code is the only way to get to the bottom of this.

It seems the same (or very similar) startup was code was shared between these games. So fixing the problem for one affected game would almost certainly fix all the others as well.

interloper98:
# Description

This adds realistic delays to `DOS_File` derived drive classes (virtual, local, ISO, FAT, device, and CDROM) read, write, seek, and close file operations.

It fixes DOS games that either directly measure IO call timing (and then get divide by zero errors using the results), or have logic that's sensitive to nearly instant IO call timing.

It uses a disk access delay guard object to keep delay functionality in one place. The guard aspect means the work is performed when it goes out of scope. This makes it easy to accumulate 'work' in large functions that have many exit points, without having to add code at each exit point.

Game benchmarks (FPS rates) are unaffected, however drive benchmarks are now much more realistic:

Hercules, early 80s:

hercules-machine-IO-speeds

CGA/Tandy/PCjr, mid 80s:

pcjr-cga-tandy-machine-IO-speeds

EGA, mid to late 80s:

ega-machine-IO-speeds

VGA, very late 80s and on:

vga-machine-IO-speeds

There's a conf option:

```ini
[dosbox]
disk_access_delay = bool (default false)
```

When disabled:

media-delay-off-IO-speeds

## Related issues

Fixes #3512 **Silmarils games newer than 1994**, Deus (1996), Ishar 3 (1994), Robinson's Requiem (1994), and Time Warriors (1997)

Fixes **Chasm The Rift** audio track looping (replaces feature from #2622)

Fixes **Star Trek - A Final Unity (1995)** CDROM speed check [1](https://www.vogons.org/viewtopic.php?t=43375), [2](https://www.vogons.org/viewtopic.php?t=3763), [3](https://www.vogons.org/viewtopic.php?t=71955), [4](https://www.vogons.org/viewtopic.php?t=27182), [5](https://adventuregamers.com/archive/forums/hint-requests-technical-problems/16492-pissed-off-about-final-unity.html)

Fixes **Emergency Room (1995)** CDROM speed check [1](https://www.vogons.org/viewtopic.php?t=17726)

# Release notes

Mount drives can now have their access times realistically delayed. This fixes division by zero crashes in Deus (1996), Ishar 3 (1994), Robinson's Requiem (1994), and Time Warriors (1997) and fixes the CDROM speed checks in the Star Trek - A Final Unity (1995) and Emergency Room (1995) installers.

This can be enabled using the `disk_access_delay` setting either at runtime or in the conf file under the `[dosbox]` section.

# Manual testing

Tested many games from various eras (CGA, EGA, VGA),  tested drives mounted from local directoriy (`mount d cdrom/ -t iso`) and images (`imgmount ... `), also tested CDDA mounts from image and files. Tested benchmarks (FPS is unaffected and load times are still very fast), also tested Windows 3.1.

# Checklist

I have:

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -147,7 +147,59 @@ public:
 	std::string name = {};
 	FlushTimeOnClose flush_time_on_close = FlushTimeOnClose::NoUpdate;
 	std::vector<FileRegionLock> region_locks = {};
-	/* Some Device Specific Stuff */
+
+	// The DiskAccessDelayGuard simulates realistic disk operation timing to fix
+	// games that expect specific IO delays. Creates artificial delays for file
+	// operations like Read, Write, Seek, and Close.
+	//
+	// Usage: Create an object in DOS_File IO methods and use Add*() calls to
+	//        queue delays. Delays are performed when the object is destroyed.
+	//
+	// Fixes timing issues in games like:
+	// - Chasm the Rift (1997): CDDA audio track fails to loop 
+	// - Emergency Room (1995): CDROM speed check during install
+	// - Deus (1996), Ishar 3 (1994), Robinson's Requiem (1994), and Time
+	//   Warriors (1997): division by 0 interrupt handler crash.
+	//
+	class DiskAccessDelayGuard {
+	public:
+		constexpr DiskAccessDelayGuard() = default;
+		~DiskAccessDelayGuard();
+
+		constexpr void AddBytesRead(const int bytes = 1) noexcept
+		{
+			bytes_read += bytes;
+		}
+		constexpr void AddBytesWritten(const int bytes = 1) noexcept
+		{
+			bytes_written += bytes;
+		}
+		constexpr void AddSeek() noexcept
+		{
+			++num_seeks;
+		}
+		constexpr void AddVerySlowSeek() noexcept
+		{
+			num_seeks += 50;
+		}
+		constexpr void AddClose() noexcept
+		{
+			++num_closes;
+		}
+		static constexpr void SetDelayPercent(
+		        const std::optional<float> percent) noexcept;
+
+	private:
+		constexpr int GetNumOperations() const noexcept;
+
+		inline static std::optional<float> delay_percent = {};
+
+		int bytes_read    = 0;
+		int bytes_written = 0;
+		int num_seeks     = 0;
+		int num_closes    = 0;
+	};
+
 private:
 	uint8_t hdrive = 0xff;
 };
--- a/src/dos/dos_devices.cpp
+++ b/src/dos/dos_devices.cpp
@@ -129,6 +129,8 @@ bool DOS_ExtDevice::WriteToControlChannel(PhysPt bufptr, uint16_t size, uint16_t
 
 bool DOS_ExtDevice::Read(uint8_t *data, uint16_t *size)
 {
+	DiskAccessDelayGuard disk_access_delay = {};
+
 	PhysPt bufptr = (dos.dcp << 4) | 32;
 	for (uint16_t no = 0; no < *size; no++) {
 		// INPUT
@@ -140,12 +142,16 @@ bool DOS_ExtDevice::Read(uint8_t *data, uint16_t *size)
 			}
 			*data++ = mem_readb(bufptr);
 		}
+
+		disk_access_delay.AddBytesRead();
 	}
 	return true;
 }
 
 bool DOS_ExtDevice::Write(uint8_t *data, uint16_t *size)
 {
+	DiskAccessDelayGuard disk_access_delay = {};
+
 	PhysPt bufptr = (dos.dcp << 4) | 32;
 	for (uint16_t no = 0; no < *size; no++) {
 		mem_writeb(bufptr, *data);
@@ -158,6 +164,8 @@ bool DOS_ExtDevice::Write(uint8_t *data, uint16_t *size)
 			}
 		}
 		data++;
+
+		disk_access_delay.AddBytesWritten();
 	}
 	return true;
 }
--- a/src/dos/dos_devices.cpp
+++ b/src/dos/dos_devices.cpp
@@ -30,6 +30,7 @@
 #include "support.h"
 #include "drives.h"
 #include "dev_con.h"
+#include "math_utils.h"
 
 DOS_Device * Devices[DOS_DEVICES];
 
@@ -373,6 +374,76 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
 	hdrive=orig.hdrive;
 	name = orig.name;
 	return *this;
+}
+
+void DOS_SetDiskAccessDelay(const bool use_disk_access_delay)
+{
+	auto get_delay_percent = []() -> std::optional<float> {
+		// the older the machine, the slower their disks performed
+		// relative to to the baseline (VGA machines). For example,
+		// drives in the early 80s (Hercules era) were about 3x slower
+		// than early 90s drives. CGA/Tandy/PCjr drives were about 50%
+		// slower, and EGA drives were about 25% slower. These are just
+		// rough differences; of course, floppy drives didn't see the
+		// same advances as hard disk drives and CD-ROM drives, but for
+		// games and software, these ballparks are more than sufficient.
+		switch (machine) {
+		case MCH_HERC: return 3.0f;
+		case MCH_CGA:
+		case MCH_TANDY:
+		case MCH_PCJR: return 1.5f;
+		case MCH_EGA: return 1.25f;
+		case MCH_VGA:
+		default: return 1.0f;
+		}
+	};
+
+	DOS_File::DiskAccessDelayGuard::SetDelayPercent(
+	        use_disk_access_delay ? get_delay_percent() : std::nullopt);
+
+	LOG_MSG("DOS: %s",
+	        use_disk_access_delay ? "Using realistic disk access delays"
+	                              : "Realistic disk access delays disabled");
+}
+
+DOS_File::DiskAccessDelayGuard::~DiskAccessDelayGuard()
+{
+	if (!delay_percent) {
+		return;
+	}
+	auto scaled_ops = iroundf(GetNumOperations() * *delay_percent);
+	while (scaled_ops--) {
+		CALLBACK_Idle();
+	}
+}
+
+constexpr void DOS_File::DiskAccessDelayGuard::SetDelayPercent(
+        const std::optional<float> percent) noexcept
+{
+	delay_percent = percent;
+}
+
+constexpr int DOS_File::DiskAccessDelayGuard::GetNumOperations() const noexcept
+{
+	// The Idle() callback used to inject the IO delay is too coarse to account
+	// for byte-level IO timing, so we use a quantity of N-to-1 for each
+	// delayable 'Operation' defined in the following constants. These produce
+	// disk rates roughly on par with early 90s HDDs: roughly 4 MB/s reads,
+	// slight faster writes of around 5 MB/s (assuming some HDD buffering), and
+	// several hundred seeks IOPS/s (this is a bit generous, but there's little
+	// benefit in making it even slower).
+	constexpr auto ReadBytesPerOp  = 128;
+	constexpr auto WriteBytesPerOp = 192;
+	constexpr auto SeekMultiplier  = 20;
+
+	// clang-format off
+
+	return (bytes_read / ReadBytesPerOp)
+	     + (bytes_written / WriteBytesPerOp)
+	     + (num_seeks * SeekMultiplier)
+	     + num_closes;
+
+	// clang-format on
 }
 
 uint8_t DOS_FindDevice(const char* name)
--- a/src/dos/drive_fat.cpp
+++ b/src/dos/drive_fat.cpp
@@ -99,6 +99,8 @@ fatFile::fatFile(const char* /*name*/, uint32_t startCluster, uint32_t fileLen,
 }
 
 bool fatFile::Read(uint8_t * data, uint16_t *size) {
+	DiskAccessDelayGuard disk_access_delay = {};
+
 	// check if file opened in write-only mode
 	if ((this->flags & 0xf) == OPEN_WRITE) {
 		DOS_SetError(DOSERR_ACCESS_DENIED);
@@ -114,6 +116,7 @@ bool fatFile::Read(uint8_t * data, uint16_t *size) {
 		currentSector = myDrive->getAbsoluteSectFromBytePos(firstCluster, seekpos);
 		if(currentSector == 0) {
 			/* EOC reached before EOF */
+			disk_access_delay.AddBytesRead();
 			*size = 0;
 			loadedSector = false;
 			return true;
@@ -131,6 +134,8 @@ bool fatFile::Read(uint8_t * data, uint16_t *size) {
 			return true; 
 		}
 		data[sizecount++] = sectorBuffer[curSectOff++];
+		disk_access_delay.AddBytesRead();
+
 		seekpos++;
 		if(curSectOff >= myDrive->getSectorSize()) {
 			currentSector = myDrive->getAbsoluteSectFromBytePos(firstCluster, seekpos);
@@ -153,6 +158,8 @@ bool fatFile::Read(uint8_t * data, uint16_t *size) {
 }
 
 bool fatFile::Write(uint8_t * data, uint16_t *size) {
+	DiskAccessDelayGuard disk_access_delay = {};
+
 	// check if file opened in read-only mode
 	if ((this->flags & 0xf) == OPEN_READ || myDrive->IsReadOnly()) {
 		DOS_SetError(DOSERR_ACCESS_DENIED);
@@ -168,6 +175,7 @@ bool fatFile::Write(uint8_t * data, uint16_t *size) {
 
 	if (seekpos < filelength && *size == 0) {
 		/* Truncate file to current position */
+		disk_access_delay.AddBytesWritten();
 		if (firstCluster != 0)
 			myDrive->deleteClustChain(firstCluster, seekpos);
 		if (seekpos == 0)
@@ -224,6 +232,8 @@ bool fatFile::Write(uint8_t * data, uint16_t *size) {
 			filelength = seekpos+1;
 		}
 		sectorBuffer[curSectOff++] = data[sizecount++];
+		disk_access_delay.AddBytesWritten();
+
 		seekpos++;
 		if(curSectOff >= myDrive->getSectorSize()) {
 			if(loadedSector) myDrive->writeSector(currentSector, sectorBuffer);
@@ -256,24 +266,31 @@ finalizeWrite:
 }
 
 bool fatFile::Seek(uint32_t *pos, uint32_t type) {
-	int32_t seekto=0;
-	
+	int32_t requested_pos = 0;
+
 	switch(type) {
-		case DOS_SEEK_SET:
-			seekto = (int32_t)*pos;
-			break;
+		case DOS_SEEK_SET: requested_pos = (int32_t)*pos; break;
 		case DOS_SEEK_CUR:
-			/* Is this relative seek signed? */
-			seekto = (int32_t)*pos + (int32_t)seekpos;
+			// Is this relative seek signed?
+			requested_pos = (int32_t)*pos + (int32_t)seekpos;
 			break;
 		case DOS_SEEK_END:
-			seekto = (int32_t)filelength + (int32_t)*pos;
+			requested_pos = (int32_t)filelength + (int32_t)*pos;
 			break;
 	}
 //	LOG_MSG("Seek to %d with type %d (absolute value %d)", *pos, type, seekto);
 
-	if(seekto<0) seekto = 0;
-	seekpos = (uint32_t)seekto;
+	if (requested_pos < 0) {
+		requested_pos = 0;
+	}
+
+	if (seekpos != static_cast<uint32_t>(requested_pos)) {
+		seekpos = static_cast<uint32_t>(requested_pos);
+
+		DiskAccessDelayGuard disk_access_delay = {};
+		disk_access_delay.AddSeek();
+	}
+
 	currentSector = myDrive->getAbsoluteSectFromBytePos(firstCluster, seekpos);
 	if (currentSector == 0) {
 		/* not within file size, thus no sector is available */
@@ -309,6 +326,9 @@ void fatFile::Close()
 		if (loadedSector) {
 			myDrive->writeSector(currentSector, sectorBuffer);
 		}
+
+		DiskAccessDelayGuard disk_access_delay = {};
+		disk_access_delay.AddClose();
 	}
 
 	set_archive_on_close = false;
--- a/src/dos/drive_iso.cpp
+++ b/src/dos/drive_iso.cpp
@@ -68,8 +68,11 @@ isoFile::isoFile(std::shared_ptr<isoDrive> iso_drive, const char *name, FileStat
 }
 
 bool isoFile::Read(uint8_t *data, uint16_t *size) {
-	if (filePos + *size > fileEnd)
+	DiskAccessDelayGuard disk_access_delay = {};
+
+	if (filePos + *size > fileEnd) {
 		*size = (uint16_t)(fileEnd - filePos);
+	}
 
 	uint16_t nowSize = 0;
 	uint32_t sector = filePos / ISO_FRAMESIZE;
@@ -106,6 +109,9 @@ bool isoFile::Read(uint8_t *data, uint16_t *size) {
 	}
 	*size = nowSize;
 	filePos += *size;
+
+	disk_access_delay.AddBytesRead(*size);
+
 	return true;
 }
 
@@ -114,21 +120,25 @@ bool isoFile::Write(uint8_t* /*data*/, uint16_t* /*size*/) {
 }
 
 bool isoFile::Seek(uint32_t *pos, uint32_t type) {
+	uint32_t requested_pos = {};
+
 	switch (type) {
-		case DOS_SEEK_SET:
-			filePos = fileBegin + *pos;
-			break;
-		case DOS_SEEK_CUR:
-			filePos += *pos;
-			break;
-		case DOS_SEEK_END:
-			filePos = fileEnd + *pos;
-			break;
-		default:
-			return false;
+	case DOS_SEEK_SET: requested_pos = fileBegin + *pos; break;
+	case DOS_SEEK_CUR: requested_pos = filePos + *pos; break;
+	case DOS_SEEK_END: requested_pos = fileEnd + *pos; break;
+	default: return false;
 	}
-	if (filePos > fileEnd || filePos < fileBegin)
+
+	if (filePos != requested_pos) {
+		filePos = requested_pos;
+
+		DiskAccessDelayGuard disk_access_delay = {};
+		disk_access_delay.AddSeek();
+	}
+
+	if (filePos > fileEnd || filePos < fileBegin) {
 		filePos = fileEnd;
+	}
 
 	*pos = filePos - fileBegin;
 	return true;
--- a/src/dos/drive_virtual.cpp
+++ b/src/dos/drive_virtual.cpp
@@ -347,6 +347,8 @@ Virtual_File::Virtual_File(const vfile_data_t& in_data)
 
 bool Virtual_File::Read(uint8_t* data, uint16_t* bytes_requested)
 {
+	DiskAccessDelayGuard disk_access_delay = {};
+
 	if (file_pos > file_data->size()) {
 		// File has been read beyond the end and is in an invalid state,
 		// so inform the caller with a negative rvalue.
@@ -363,6 +365,9 @@ bool Virtual_File::Read(uint8_t* data, uint16_t* bytes_requested)
 
 	file_pos += bytes_to_read;
 	*bytes_requested = check_cast<uint16_t>(bytes_to_read);
+
+	disk_access_delay.AddBytesRead(*bytes_requested);
+
 	return true;
 }
 
@@ -374,35 +379,47 @@ bool Virtual_File::Write(uint8_t* /*data*/, uint16_t* /*size*/)
 
 bool Virtual_File::Seek(uint32_t* new_pos, uint32_t type)
 {
+	uint32_t requested_pos = {};
+
 	switch (type) {
 	case DOS_SEEK_SET:
 		if (*new_pos <= file_data->size()) {
-			file_pos = *new_pos;
+			requested_pos = *new_pos;
 		} else {
 			return false;
 		}
 		break;
 	case DOS_SEEK_CUR:
 		if ((*new_pos + file_pos) <= file_data->size()) {
-			file_pos = *new_pos + file_pos;
+			requested_pos = *new_pos + file_pos;
 		} else {
 			return false;
 		}
 		break;
 	case DOS_SEEK_END:
 		if (*new_pos <= file_data->size()) {
-			file_pos = file_data->size() - *new_pos;
+			requested_pos = file_data->size() - *new_pos;
 		} else {
 			return false;
 		}
 		break;
 	}
+
+	if (file_pos != requested_pos) {
+		file_pos = requested_pos;
+
+		DiskAccessDelayGuard disk_access_delay = {};
+		disk_access_delay.AddSeek();
+	}
+
 	*new_pos = file_pos;
 	return true;
 }
 
 void Virtual_File::Close()
 {
+	DiskAccessDelayGuard disk_access_delay = {};
+	disk_access_delay.AddClose();
 }
 
 uint16_t Virtual_File::GetInformation() {
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -561,6 +561,8 @@ static void DOSBOX_RealInit(Section* sec)
 	}
 
 	VGA_SetRatePreference(section->Get_string("dos_rate"));
+
+	DOS_SetDiskAccessDelay(section->Get_bool("disk_access_delay"));
 }
 
 // Returns decimal seconds of elapsed uptime.
@@ -704,6 +706,17 @@ void DOSBOX_Init()
 	        "            500 to 5000 ns is the most useful range.\n"
 	        "Note: Only set this on a per-game basis when necessary as it slows down\n"
 	        "      the whole emulator.");
+
+	pbool = secprop->Add_bool("disk_access_delay", when_idle, false);
+	pbool->Set_help(
+	        "Emulate disk access delays ('off' by default).\n"
+	        "on:   Emulate disk access delays for floppy drives, hard drives, and CD-ROM\n"
+	        "      drives (default). This is preferable in cases were games crash or\n"
+	        "      misbehave if the disk access times are unrealistically fast (e.g.,\n"
+	        "      Deus, Ishar 3, Robinson's Requiem, Time Warriors, Chasm - The Rift,\n"
+	        "      Start Trek - A Final Unity, Emergency Room, etc.)\n"
+	        "off:  Disable disk access delay emulation. The only benefit to this is\n"
+	        "      this slightly faster install and load time.");
 
 	pstring = secprop->Add_string("dos_rate", when_idle, "default");
 	pstring->Set_help(
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -707,9 +707,9 @@ void DOSBOX_Init()
 	        "Note: Only set this on a per-game basis when necessary as it slows down\n"
 	        "      the whole emulator.");
 
-	pbool = secprop->Add_bool("disk_access_delay", when_idle, false);
+	pbool = secprop->Add_bool("disk_access_delay", when_idle, true);
 	pbool->Set_help(
-	        "Emulate disk access delays ('off' by default).\n"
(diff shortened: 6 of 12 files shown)


src/dos/drive_virtual.cpp
```diff
@@ -374,29 +374,36 @@ bool Virtual_File::Write(uint8_t* /*data*/, uint16_t* /*size*/)
...
+			requested_pos = file_data->size() - *new_pos;
 		} else {
 			return false;
 		}
 		break;
 	}
+
+	if (file_pos != requested_pos) {
+		file_pos = requested_pos;
+	}
```
weirddan455: Remove the if satement.  Assuming these are normal integers and there's no weird copy constructor behavior, it doesn't accomplish anything to compare them before setting them equal.

src/dos/drive_iso.cpp
```diff
@@ -114,19 +114,20 @@ bool isoFile::Write(uint8_t* /*data*/, uint16_t* /*size*/) {
...
-			return false;
+	case DOS_SEEK_SET: requested_pos = fileBegin + *pos; break;
+	case DOS_SEEK_CUR: requested_pos = filePos + *pos; break;
+	case DOS_SEEK_END: requested_pos = fileEnd + *pos; break;
+	default: return false;
 	}
+
+	if (filePos != requested_pos) {
+		filePos = requested_pos;
+	}
```
weirddan455: Same comment as before.  Here I don't even see what adding a new `requested_pos` variable does.  Behavior looks to me to be exactly the same as just setting `filePos` directly like it was before.

src/dos/drive_fat.cpp
```diff
@@ -256,25 +256,30 @@ bool fatFile::Write(uint8_t * data, uint16_t *size) {
...
 
-	if(seekto<0) seekto = 0;
-	seekpos = (uint32_t)seekto;
+	if (requested_pos < 0) {
+		requested_pos = 0;
+	}
+
+	if (seekpos != static_cast<uint32_t>(requested_pos)) {
+		seekpos = static_cast<uint32_t>(requested_pos);
+	}
```
weirddan455: Same comment about the if statement.  This code does use `requested_pos` to check for wrapping so makes sense to have the variable.

src/dos/drive_virtual.cpp
johnnovak: I tend to agree; I don't see why this is necessary. Old code was fine.

include/dos_system.h
```diff
@@ -147,7 +147,54 @@ class DOS_File {
...
+		}
+		static constexpr void SetDelayPercent(const int scalar) noexcept;
+
+	private:
+		constexpr uint16_t GetNumOperations() const noexcept;
+		inline static int delay_percentage{100};
+		uint16_t bytes_read{0};
+		uint16_t bytes_written{0};
+		uint16_t num_seeks{0};
+		uint16_t num_closes{0};
```
johnnovak, line 195: I guess you're using `uint16_t` due to FAT limitations and/or because the DOS I/O API already uses `uint16_t` everywhere so you don't need to cast?

I'm 50/50 on that. Using signed `int64_t` is preferred for such accumulators, so maybe consider changing it?

weirddan455 (commented):
I'm not necessarily against adding this feature but I have a couple of concerns:

IMO I would prefer to have this off by default.  It only solves very niche issues at the expensive of increasing install and load times.  I think Dosbox-X has a similar feature and is also off by default (I tried to look it up in their conf file but it's very large and I couldn't find it so you may want to confirm this if you're curious).

This code also doesn't appear to take into account the host system's natural delay in I/O.  The file we're reading could be on a network drive or physical CDROM (CDROM drive inherits from `localDrive` and shares that same code for reading normal files).

Regarding the issues you linked:

There's no evidence that #3512 is directly linked to I/O being too fast.  Maybe it is but it's equally likely the root cause is a timing issue elsewhere.  Were you able to reproduce this and confirm the fix?  IIRC I wasn't ever able to get a reliable repro.

#2622 (the `LagDriveResponse()` function) was solving a very specific problem where the game would have a bug if the CD Audio position did not advance from the last time it was checked.  It only lagged audio calls, not *all* I/O and I vaguely remember even that caused a regression somewhere and the function had to be re-tweaked later (maybe I'm mis-remembering, would be good to check past issues on this though).

The one concrete issue that this solves is CDROM speed checks in game installers.  I have run into this in one of the Wing Commander games but every one I have seen can be skipped even if it fails.

src/dosbox.cpp
```diff
@@ -797,6 +799,13 @@ void DOSBOX_Init()
 	        "you're using a copy-on-write or network-based filesystem, this setting avoids\n"
 	        "triggering write operations for these write-protected files.");
 
+	pbool = secprop->Add_bool("media_delay", when_idle, true);
+	pbool->Set_help(
+	        "Emulates media delay for mounted HDD, CDROM, and floppy drives.\n"
+	        "This is useful for games that have problems if media is too fast,\n"
+	        "      such as Deus (1996), Ishar 3 (1994), Robinson's Requiem (1994),\n"
+	        "      and Time Warriors (1997).");
```
johnnovak, line 807: How about renaming this to `disk_access_delay` and using the below description?

```
Emulate disk access delays ('on' by default).
  on:   Emulate disk access delays for floppy drives, hard drives, and CD-ROM
        drives (default). This is preferable as many DOS games crash or
        misbehave if the disk access times are unrealistically fast (e.g.,
        Deus, Ishar 3, Robinson's Requiem, Time Warriors, Chasm - The Rift,
        Start Trek - A Final Unity, Emergency Room, etc.)
  off:  Disable disk access delay emulation. There is generally no benefit to
        this as disk access is fast enough even with the delay emulation
        enabled.
```

---

"Media" is such an overloaded term, and it's really isn't clear what it refers to. Especially in a "multimedia" app like DOSBox where "media" could mean audio or video as well.

The term "disk access delay" is still short and user-friendly and 100% unambigous.

include/dos_system.h
```diff
@@ -147,7 +147,54 @@ class DOS_File {
...
+	// are sensitive to IO calls that return too quickly. Examples are Deus
+	// (1996), Ishar 3 (1994), Robinson's Requiem (1994), and Time Warriors
+	// (1997). They (otherwise) report an iterrupt #00 handler error, which
+	// is triggered on a divide by zero.
+	//
+	// Create an object in DOS_File derived class's Read, Write, Seek, and
+	// Close functions. Add..() respective quantities. When the function
+	// goes out of scope, the cost for all added quantities will be applied.
+	//
+	class MediaDelayGuard {
```
johnnovak, line 162: As I mentioned in my comment about the config setting's description further down, I would prefer `DiskAccessDelayGuard` and the term "disk access delay" in general. "Media" is very ambiguous (it could mean audio or video as well) and there's not much point in choosing a more ambiguous term when we have something much narrower that fits the bill, which is "DiskAccessDelay".

include/dos_system.h
```diff
@@ -147,7 +147,54 @@ class DOS_File {
...
+		}
+		constexpr void AddClose() noexcept
+		{
+			++num_closes;
+		}
+		static constexpr void SetDelayPercent(const int scalar) noexcept;
+
+	private:
+		constexpr uint16_t GetNumOperations() const noexcept;
+		inline static int delay_percentage{100};
```
johnnovak, line 191: Would float/double be better?

include/dos_system.h
```diff
@@ -147,7 +147,54 @@ class DOS_File {
...
+			num_seeks += 50;
+		}
+		constexpr void AddClose() noexcept
+		{
+			++num_closes;
+		}
+		static constexpr void SetDelayPercent(const int scalar) noexcept;
+
+	private:
+		constexpr uint16_t GetNumOperations() const noexcept;
```
johnnovak, line 190: Just use `int`. Specific-sized integers and especially unsigned types should only ever be used for binary protocols and register-level hackery in our codebase.

I know there was a "push" to start using fixed-size types religiously _everywhere_ in the codebase... but that was a mistake.

More info about it here:
https://google.github.io/styleguide/cppguide.html#Integer_Types

include/dos_system.h
johnnovak: After some further though, these should all be just `int`s, especially the count vars.

include/dos_system.h
```diff
@@ -147,7 +147,54 @@ class DOS_File {
...
+	// Create an object in DOS_File derived class's Read, Write, Seek, and
+	// Close functions. Add..() respective quantities. When the function
+	// goes out of scope, the cost for all added quantities will be applied.
+	//
+	class MediaDelayGuard {
+	public:
+		constexpr MediaDelayGuard() = default;
+		~MediaDelayGuard();
+
+		constexpr void AddBytesRead(const uint16_t bytes = 1) noexcept
```
johnnovak, line 167: `int num_bytes = 1`

include/dos_system.h
```diff
@@ -147,7 +147,54 @@ class DOS_File {
...
+		{
+			bytes_read += bytes;
+		}
+		constexpr void AddBytesWritten(const uint16_t bytes = 1) noexcept
```
johnnovak, line 171: `int num_bytes = 1`

include/dos_system.h
johnnovak: Also, can you get rid of the weird initialiser syntax? There's no benefit to using it here. Just do the least surprising thing, please:

`int bytes_read = 0;`

src/dos/dos_devices.cpp
```diff
@@ -375,6 +375,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
 	return *this;
 }
 
+void DOS_SetMediaDelay(const bool use_media_delay)
+{
+	auto get_delay_percent = []() {
+		LOG_MSG("machine = %d", machine);
```
johnnovak, line 389: Remove debug log message?

src/dos/dos_devices.cpp
```diff
@@ -375,6 +375,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+		case MCH_VGA:
+		default: return 100;
+		}
+	};
+
+	DOS_File::MediaDelayGuard::SetDelayPercent(
+	        use_media_delay ? get_delay_percent() : 0);
+
+	LOG_MSG("DOS: Media will %s delays",
+	        use_media_delay ? "use realistic" : "run without");
```
johnnovak, line 405: We don't really use future tense in the logs; we state things:

image

Can you reword please to "Using realistic disk access delays" and "Realistic disk access delays disabled" or something?

src/dos/dos_devices.cpp
```diff
@@ -375,6 +375,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+}
+
+DOS_File::MediaDelayGuard::~MediaDelayGuard()
+{
+	if (delay_percentage == 0) {
+		return;
+	}
+	auto scaled_ops = (GetNumOperations() * delay_percentage) / 100;
```
johnnovak, line 413: Dunno, I'd just use float calculations here. Maybe it doesn't matter now, but you can get into some weird trunaction situation with integer arithmetic, so there's not much point to it.

src/dos/dos_devices.cpp
```diff
@@ -375,6 +375,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+{
+	assert(percentage >= 0);
+	delay_percentage = percentage;
+}
+
+constexpr uint16_t DOS_File::MediaDelayGuard::GetNumOperations() const noexcept
+{
+	constexpr uint16_t ReadBytesPerOp  = 128;
+	constexpr uint16_t WriteBytesPerOp = 192;
+	constexpr uint16_t SeekMultiplier  = 20;
```
johnnovak, line 429: `int`s all the way :)

src/dos/dos_devices.cpp
```diff
@@ -375,6 +375,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+	if (delay_percentage == 0) {
+		return;
+	}
+	auto scaled_ops = (GetNumOperations() * delay_percentage) / 100;
+	while (scaled_ops--) {
+		CALLBACK_Idle();
+	}
+}
+
+constexpr void DOS_File::MediaDelayGuard::SetDelayPercent(const int percentage) noexcept
```
johnnovak, line 419: I think `const int percent` is the correct word here, not "percentage".

src/dos/dos_devices.cpp
```diff
@@ -375,6 +375,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+{
+	assert(percentage >= 0);
+	delay_percentage = percentage;
```
johnnovak, line 422: I think `delay_percent` is the correct name.

include/dos_system.h
johnnovak: How about this for the last paragraph:

```
Usage:

- Create `DiskAccessDelayGuard` object in `DOS_File` derived class's Read,
  Write, Seek, and Close functions.
- Call the `Add*()` methods to add the respective quantities.
- When the function (and thus the guard object) go out of scope, the cost for
  all added quantities will be applied.
```

include/dos_system.h
johnnovak: Also, no weird initialisers please when we can avoid them. 

```
static int delay_percentage = 100;
```

Also, why the inline? Have benchmarks identified a bottleneck here? I'd say let the compiler decide what's best unless we have a very good reason to override what it's doing.

src/dos/dos_devices.cpp
johnnovak: Some explanation is in order here. What is scaled by this percent value? First I thought 100% means "no delay". But that's clearly not the case as when the delay is disabled, you set this to 0%.

Would you care to put some explanatory comments here on what 0%, 100%, and any other percent value mean?

src/dos/dos_devices.cpp
johnnovak: Would you care to elaborate a bit in a comment how you ended up with these magic constants?

src/dos/dos_devices.cpp
```diff
@@ -375,6 +383,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+	auto get_delay_percent = []() {
+		LOG_MSG("machine = %d", machine);
+		switch (machine) {
+		case MCH_HERC: return 300;
+		case MCH_CGA:
+		case MCH_TANDY:
+		case MCH_PCJR: return 150;
+		case MCH_EGA: return 125;
+		case MCH_VGA:
+		default: return 100;
```
johnnovak, line 397: Would you be able to elaborate a bit in extra comments on why you picked these magic constants?

johnnovak:
> I'm not necessarily against adding this feature but I have a couple of concerns:
> 
> IMO I would prefer to have this off by default. It only solves very niche issues at the expensive of increasing install and load times. I think Dosbox-X has a similar feature and is also off by default (I tried to look it up in their conf file but it's very large and I couldn't find it so you may want to confirm this if you're curious).

I'm of two minds regarding this. In principle, emulating I/O delays is _more_ authentic, therefore it should improve emulation accuracy.

But then, DOSBox is full of assumptions and features built on top of previous shortcuts and simplification. In other words, we don't really know what problems the delay emulation might introduce.

Given that, I think I'm slightly preferring to leave this off by default, after all, following the "not broken, don't fix it" approach. It's best to make these changes "opt-in", after all, that don't improve the experience of _every_ single program or game (like VGA double scanning was... which was super difficult to get it right after many many regression fixes...)

> This code also doesn't appear to take into account the host system's natural delay in I/O. The file we're reading could be on a network drive or physical CDROM (CDROM drive inherits from `localDrive` and shares that same code for reading normal files) and therefore is "slow enough" already.

Interesting; if the resulting delay is native I/O delay dependent, then that throws another variable into the mix... Oh well. Ideally, these emulated I/O speeds would be host-agonistic. But then I guess that's only possible if the host I/O is always a lot faster than DOS-era I/O rates... which is always true? But yeah, maybe network shares introduce their own problems.

> Regarding the issues you linked:
> 
> There's no evidence that #3512 is directly linked to I/O being too fast. Maybe it is but it's equally likely the root cause is a timing issue elsewhere. Were you able to reproduce this and confirm the fix? IIRC I wasn't ever able to get a reliable repro.

First I thought it directly linked, but you're right; there's actually no evidence. Might as well some regression introduced by the changed VGA timings, no?

> #2622 (the `LagDriveResponse()` function) was solving a very specific problem where the game would have a bug if the CD Audio position did not advance from the last time it was checked. It only lagged audio calls, not _all_ I/O and I vaguely remember even that caused a regression somewhere and the function had to be re-tweaked later (maybe I'm mis-remembering, would be good to check past issues on this though).

I vaguely remember that issue too 😄 We gotta be careful not to undo previous fixes with this fix, yeah 😅 

> The one concrete issue that this solves is CDROM speed checks in game installers. I have run into this in one of the Wing Commander games but every one I have seen can be skipped even if it fails.

Another reason for making this opt-in, then.

include/dos_system.h
johnnovak: Sounds reasonable, thanks for the extra info.

src/dos/drive_virtual.cpp
weirddan455: Yep sorry I missed that.  If statement makes sense then.

Now that I think about it though, should we really be adding a delay to Seek at all?  It doesn't necessarily move the drive head.  MS-DOS Encyclopedia has the interrupt documented with the name "Move File Pointer" rather than "Seek" because really all it needs to do is update an in-memory internal variable to point to the next byte it wants to read/write.  It shouldn't need to actually access the disk until the next read or write.

johnnovak: You're right, most likely it works like the C seek call; it only updates the in-memory file pointer which will then be take in to consideration on the next read/write call.

weirddan455:
>  throttling disk speeds has a positive side effect of improving emulation performance

There's no potential for performance improvement here.  That quote is from 86box and can only be true when you're running a real DOS kernel under emulation.  DOSBox has its own implementation of the DOS kernel that runs at full speed on the host CPU so we're never "wasting" emulation cycles on I/O.

Sidenote:  There is low-level IDE code in DOSBox that lets you boot into real MS-DOS where something like what you said can maybe apply but that's not how most people use it and this PR doesn't touch the IDE code.

johnnovak:
Thanks @interloper98. So, like I said, I'm of two minds on this.

On one hand, I would _like_ to make this on by default because the problem is how would people know what hit them and what obscure conf setting they need to enable to get over the issue?

On the other hand, I'm wary of regressions and general slowdowns in I/O heavy games/installers.

So I'd say let's examine the issue a little more.

> > It only solves very niche issues at the expensive of increasing install and load times.
> 
> Yup; most games and programs shouldn't be affected (but some are; at least we know of a handful.. how many more are more subtly affected? I really don't know).

Do we know how much slower the install times are? In a way, disabling the setting just to speed up installs is more logical than having to _guess_ that enabling "disk access delays" is the fix if a game crashes at startup. Basically no one would be able to guess that, but "guessing" that disabling "disk access delays" would speed up installers is not even a guess, that's obvious.

So if the install times for typical games are not absolutely terrible, I don't see the slowdown as a showstopper. I mean, right now install times are lightning fast. So what kind of slowdowns are we talking about? Factor of 10? I could live with 20-30 second install times instead of 2-3 seconds _by default_, which people can speed up by disabling this setting.

> > The file we're reading could be on a network drive or physical CDROM (CDROM drive inherits from `localDrive` and shares that same code for reading normal files) and therefore is "slow enough" already.
> 
> Even the slowest host-side IO calls are hidden from the emulation's understanding of the passage of time (because they block), and we don't time or thread those calls.

Okay, I think I understand what you're saying. Basically, you're saying host OS file access delays have zero effect on the disk access delays "seen" by the emulated code. If so, that's good and we can move on.

> > Regarding the issues you linked:
> > There's no evidence that #3512 is directly linked to I/O being too fast.
> 
> It's definitely related to fake/instant IO.
> 
> These Silmarils games run through a text Q&A, then you launch the game from the CLI menu. They set up their 256c VGA mode and then starts doing a bunch of back-to-back seeks and reads. The program is stable up to that point, however if those reads take too little time then the game's division by zero INT0 handler throws that error. There's not much more going on (in relation to DOS calls) during those IO reads - they just need to be spaced out to realistic rates.
> 
> Robinson's Requiem tolerates its game files being on any kind of disk, and I tested all of the mount types (ISO from bin/cue, ISO from `.iso`, virtual mount `mount d cdrom/ -t cdrom`, and local files `mount c game`); so was able to witness the same exception happening from all mountable disk types - and that's why this PR touches all of them.
 
That's all good and well and seems plausible. However, how do you explain that my `vga_render_per_scanline ` workaround which can also fix these Silmaris games (but not as well as your fix, no doubt)? I'm not trying to be difficult here, just want to understand how anything VGA emulation related could have any effect on the program getting or not getting a divide by zero error when measuring I/O read speeds. Did you investigate and/or understand this part of the puzzle as well? If we get a convincing explanation for this, that would increase the likelihood of enabling this by default, and I'd love to remove my workaround as well.

> > Maybe it is but it's equally likely the root cause is a timing issue elsewhere. Were you able to reproduce this and confirm the fix? IIRC I wasn't ever able to get a reliable repro.

So @weirddan455, for the record, these Silmarils crashes on "too high" cycles values are 100% reproducible by me and many others, including the eXo team. I introduced my workaround for the eXo guys to make this games usable at all with the latest Staging version.

FeralChild64:
> We could try putting the IO calls on a thread while running the disk idler's callbacks on the main thread to hide as much host IO time as possible, but this would probably need a long-lived IO threaded object that we can quickly lock/unlock, because there are tons of programs and games that do single byte IO (read(1), seek(1), read(1), seek(1) tens of thousands of times a second.. when I started down this PR even batch files were running slowly 😆 )

IMHO we need a read-ahead buffer:
- guest side requests data we have in a buffer => we return buffered data, doing `CALLBACK_Idle()` from time to time, as in the current PR
- guest side requests data we don't have in a buffer yet => we wake up the worker, `CALLBACK_Idle()`, we check if the worked has finished (`std::atomic<bool>`?), if not - `CALLBACK_Idle()` again until it finishes

This could also minimize audio stuttering due to slow/unreliable I/O, like when running games directly from a pendrive, SD card (Raspberry Pi!), network share, etc.

Just please don't apply the read-ahead buffer to drive `Z:`; the `Z:\AUTOEXEC.BAT` can get regenerated in the background if it executes a command to change the code page.

I don't suggest to implement it now, this is rather an idea for the `2.0` version of your feature.

weirddan455:
>  IMHO we need a read-ahead buffer

Did MS-DOS use an I/O buffer?  We probably want to match that behavior if we want to get the delays as accurate as possible.  If it did (and I'm guessing it might if games were able to read 1 byte at a time) then we could use one as well that's a good idea.

I would suggest instead of threading, we look at asynchronous I/O APIs.  Ex: io_uring on Linux: https://man7.org/linux/man-pages/man7/io_uring.7.html

I've read Windows has something similar.  No idea about Mac though.

>  I don't suggest to implement it now, this is rather an idea for the 2.0 version of your feature.

Agreed.  I merged a change some months back to use native OS APIs for I/O so we already have some abstraction here making the change to async a bit easier to refactor in the future.

src/dos/dos_devices.cpp
```diff
@@ -375,6 +383,61 @@ DOS_File &DOS_File::operator=(const DOS_File &orig)
...
+	constexpr uint16_t ReadBytesPerOp  = 128;
+	constexpr uint16_t WriteBytesPerOp = 192;
+	constexpr uint16_t SeekMultiplier  = 20;
+
+	// clang-format off
+
+	return (bytes_read / ReadBytesPerOp)
+	     + (bytes_written / WriteBytesPerOp)
+	     + (num_seeks * SeekMultiplier)
+	     + num_closes;
```
weirddan455, line 436: So this is the problem I was trying to describe with the Seek delay.  You're adding a delay when a game seeks within the same file (presumably to account for the drive head moving) but you have no such delay (that I can tell) when a game reads from one file and then immediately afterwards reads from a different file which could be on the other side of the drive.

Seeking within the same file shouldn't cause much of a delay in the first place.  Barring fragmentation, the bytes should be contiguous.  The drive head delay really comes in when you're reading from multiple files spread across the disk.

If we really wanted to be accurate we would need to keep track of what the last file we read or wrote to is.  I wasn't suggesting that for this PR though.  I was just suggesting to remove the seek delay entirely as, IMO, it doesn't make sense the way it's implemented here.  Let me know if I'm missing something though.

FeralChild64:
DOS could have a read-ahead buffer if a `SMARTDRV.EXE` (or a 3rd party disk caching software, for example from the _Norton Utilities_) was used.
But my proposal is a read-ahead buffer invisible to the guest side - just to work around the problem of slowly working semaphores.

weirddan455:
> DOS could have a read-ahead buffer if a `SMARTDRV.EXE` (or a 3rd party disk caching software, for example from the _Norton Utilities_) was used. But my proposal is a read-ahead buffer invisible to the guest side - just to work around the problem of slowly working semaphores.

Oh right, that makes sense.  We would be adding an artificial delay anyway so we just want to minimize host-side latency.  Yeah, that's a good idea then.

weirddan455:
>  I think there are still two holes in this implementation: physical CD-ROM mounts and when using IDE storage booted from real MS-DOS.

Physical CD-ROM should be covered by what you've already done.  There's a `cdromDrive` class in `drive_local.cpp` but it inherits from `localDrive` for the `Read()` and `Seek()` methods.  It mostly works the same as a local directory mount.  The small bit of unique code is to ensure it's read-only and interface with MSCDEX in a couple places.

As for real MS-DOS boot, that works at the block level rather than the file level.  I haven't worked in that part of the code but maybe look at the `imageDisk` class in `bios_disk.cpp` and there's also some code in `ide.cpp` that may be relevant.  I'd probably consider that out of scope for this PR if I were you but it's your call.  It's likely to need a different approach than what you've got here to work properly.

johnnovak:
Someon clarify this please because I think you guys are conflating two separate issues here. @interloper98 wrote this:

> Even the slowest host-side IO calls are **hidden** from the emulation's understanding of the passage of time (because they block), and we don't time or thread those calls.

Then:

> We could try putting the IO calls on a thread while running the disk idler's callbacks on the main thread to **hide** as much host IO time as possible, 

I think I'm slight confused about the word hidden in both paragraphs 😄 But reducing host OS induced delays has nothing to with the emulated disk acces delays, right? It would just be preferable to reduce overall emulator stalls due to blocking I/O.

Just wanted a confirmation on this because it muddies the waters a lot to talk about it in this PR.

shermp:
My reading is that this PR is additive in total time - basically the emulation delay time + existing host time, wheras a threaded/async approach would allow "hiding" the host access time within the emulation delay.

johnnovak:
Thanks for confirming what I thought @interloper98. Now the idea to "convert" the measured host induced I/O delays into "disk access delay" makes more sense, but ultimately probably not worth it as it could complicate things too much.

Making host I/O non-blocking at the whole emulator level has value, though. Just probably it's easiest and best to keep it separate from the disk access delay emulation.

johnnovak:
> I benchmarked starting up Quake going straight into a time demo, and both startup in roughly 1.8 seconds with or without the delays.

So ~30 seconds with delay emulation vs under 1 second with the delay.

That's perfectly acceptable, I don't see a problem. People installing a single game surely can wait up to a minute. Then people installing tens or hundreds of games from CD-ROM in bulk need to disable the access delays explicitly.

Again, it's a trade-off: the setting is not discoverable when a disk access delay related issue hits you; you wouldn't know what caused the crash. But it's very much discoverable is you're bothered by slow emulated disk access delays.

I also have an alternative idea: introduce an `auto` setting that dynamically disables disk access delay emulation if the current executable is called `SETUP.EXE` or `INSTALL.EXE` (and maybe we can add a few more white-listed names). Yes, some installers will complain about the CD-ROM being "too fast", but again, then you figure it out rather easily to enable disk access delay emulation.

My guiding principle is to hit a balance between minimise surprises, making things discoverable, and maximise the number of titles working correctly out of the box.

FeralChild64:
Or maybe do a little bit less delays with the emulated CD drive? For example, to make it behave more like a 16x drive?

johnnovak:
> but I kind of don't like these; they feel like sneaky tweaks

I personally don't mind realistic installation times at all. If anything, I welcome them—sub-one-second install times just make it obvious to me I'm inside the emulator. It undoes the finely crafted illusion created by the CRT shaders, man 😎 It was just a mitigation idea if the rest of team break out the pitchforks... 😆 

> @FeralChild64 , I'm a bit reluctant to tighten down the delays much more.. there's a point where it causes problems with Chasm the Rift looping. Maybe I could find some constants that more generous on reads while more punitive for seeks. But it's pretty tedious testing that game.

Yeah, let's leave them as they are. We want these benefits, after all.

> Should I default it to on so we can test it during the 0.83 alpha phase, and then make the call come release time? (if it's a pain in the butt and everyone's shutting it off, then yeah, default it to off).

> Ok - added a commit that enables it by default; so if it's a problem we can just revert it.

Yes & good 😄

So on the VGA drawing "entanglement" is to be understood a bit better. VGA rendering per-scanline obviously takes a lot less time than rendering the whole screen in quarter chunks which I believe is what my workaround brings back from the dead 😄 What makes me feel uncomfortable about this is the VGA delays _potentially_ contributing to the I/O delays... or the two being entagled... or... something 😅

johnnovak:
@interloper98 Okay thank, but all your results are expected so far. This is the interesting scenario that you haven't measured: what does your logging say with `vga_render_per_scanline = on` and `disk_access_delay = off`? I expect delays being introduced solely by `vga_render_per_scanline = on` to the I/O reads, otherwise I can't see how can the issue be fixed _either_ by `vga_render_per_scanline = on` _or_  `disk_access_delay = on`.

Could you post those results as well? If I'm right, the VGA code is intertwined with the I/O code in some way.

johnnovak:
Thanks @interloper98, those real MS-DOS test results are interesting. Before going further with the analysis, let's just summarise the important action points:

- Should disk access delays be enabled by default? **YES**
- Is the PR ready for merging? **YES**
- Does this fix make the `vga_render_per_scanline` unnecessary? **YES**

Thank you for your work on this; I'm happy I can retire my workaround for those four Silmarils titles.

So, from a practical sense, we're done, but I don't think we've understand the real root cause here (further analysis below the line 😆)

---

Taking your disk access delay stuff entirely out of the equation, the game crashes on 0.80.1 (and earlier), 0.74 and SVN with `vgaonly` and `cycles = 3400` or higher (3300 is fine, no crash!):

```ini
[dosbox]
machine = vgaonly

[cpu]
cycles = 3400

[autoexec]
mount c drives/c
imgmount d cd/robin.cue -t iso
d:
start
```

With the default `svga_s3` machine type, which does no double scanning in these DOSBox versions, you can turn cycles up to 22,000 without triggering a crash. Set it to 23k, it's crash city again.

No disk access delay emulation in sight.

---

Moving on to 0.81.0 and later and `main`. With my `vga_render_per_scanline` workaround, I'm effectively restoring the `vgaonly` quarter-screen rendering, so the same cycles behaviour applies as with `svga_s3` on  0.80.1 (and earlier), 0.74 and SVN:


```ini
[dosbox]
vga_render_per_scanline = off

[cpu]
cycles = 22000 
```

No disk access delay emulation in sight.

---

So, to summarise the results as it applies to Staging (only to emulated DOS, so no MS-DOS):

- **0.80.1 (and earlier), default config** — crash from 23,000 cycles 
- **0.80.1 (and earlier),** `machine = vgaonly` — crash from 3400 cycles
- **0.81.0 (and later, incl. main)** —  crash from 3400 cycles
- **0.81.0 (and later, incl. main)**, `vga_render_per_scanline = off` — crash from 23,000 cycles

Observations so far:

- Disk access delay is _not_ emulated in these cases—this is before your PR! 😄 
- The division by zero issue at startup is triggered above a specific CPU cycles threshold.
- The threshold is dependent (!) on whether the VGA emulation draws each scanline one by one (so the draw line function is called 200 times per frame via our PIC timer hack), or by quarter frame (so it runs 4 times per frame via our PIC timer hack).

Now with your PR with everything at default (so with disk access delay emulation enabled) and only changing the `cycles `setting:

- Up until 70,000 cycles the game runs perfectly. No crash at startup and there is sound in the intro.
- From 71,000 cycles there is no crash, but we lose the sound.
- With `cycles = max`, the crash is back on my M1 MacBook.
- I can also trigger the crash with CPU throttling enabled from about 10k cycles (tried with the debug build):
   ```ini
   [cpu]
   cpu_cycles = 10000
   cpu_cycles_protected = auto
   cpu_throttle = on
   ```

So we _cannot say_ the crashes are 100% CD-ROM delay emulation dependent. It seems an interesting contributing factor, and the disk access delay emulation seems to _cancel out_ the VGA drawing method dependent cycles sensitivity in these four games... but we gotta stop here as everything else is conjecture!

What we can say is that something about the specific way of doing VGA rendering and the CD-ROM speed emulation are contributing factors to the div by 0 issue at startup—and that's the end of it. These two things might influence a _third_ thing the game is measuring, directly or indirectly, for all we know.

I'm just want to make it crystal clear that we have _not_ understood the true root cause for the crashes; we're massaging things. Which is fine, and it's a happy conincidence in a way that emulating more realistic CD-ROM speeds fixes/cancels out/mitigates this mystery issue (and it does so better than what we had before with the quarter screen rendering, no doubt about that!)

So, well done, @interloper98 🎉 Hopefully it won't raise from the dead again after some random future improvement that changes up the timings again 😆

johnnovak:
> For this PR, I did more testing with this and think that using dosbox-x's approach like @weirddan455 suggested will likely provide the best results.

You mean make it opt-in? What about the discoverability angle I brought up?

johnnovak:
> > > For this PR, I did more testing with this and think that using dosbox-x's approach like @weirddan455 suggested will likely provide the best results.
> > 
> > 
> > You mean make it opt-in? What about the discoverability angle I brought up?
> 
> Oh, absolutely default-enabled because that provides IO timings that developers experienced back then (and might have those assumptions or timings measured in the code).

Yep, I agree.

> But as far as implementation, DOSBox X has the delay happening inside the DOS interruption handler, and also has eliminated the fake cycle adjustment that currently exists: both of these approaches produce better test results than this current state of this PR.

Makes sense, that sounds like a better approach, yeah. But if you don't do the fake cycles adjustment, will we still get the exact same improvements for the affected Silmarils games? I guess you'll try and see. I just have a hunch it was the fake CPU speed slowdown that fixed them.

interloper98:
Closing this because I haven't been able to work on it for some weeks now, and don't foresee being able to for some months. (and since this time, I've not been able to join chats or help in reviewing PRs beyond some quick comments).

Thanks to everyone who reviewed and tested it. 

 If someone wants to progress it in the meantime, feel free to run with it 🤘 🙂

johnnovak:
> Closing this because I haven't been able to work on it for some weeks now, and don't foresee being able to for some months. (and since this time, I've not been able to join chats or help in reviewing PRs beyond some quick comments).
> 
> Thanks to everyone who reviewed and tested it.
> 
> If someone wants to progress it in the meantime, feel free to run with it 🤘 🙂

@interloper98 
It's a pity because it sounded like you were close with your X-like approach. Isn't that a "nice to have" enhancement only, though? Would it make sense to merge this as-is as it brings immediate improvements to the users, then you can rework it months later at your leisure?

Unless you've found something fundamentally wrong with the whole approach in this PR, this would be preferable IMO. My understanding is the improved approach you started working on would let people use ridiculously high cycles settings, like over 1 million or max. But in practice, this fix already does its job when the cycles setting is kept within reasonable (and actually useful) ranges, doesn't it?

interloper98:
I appreciate the confidence @johnnovak, but I don't think it's good in current form. It's too spread out and doesn't produce as good of test results compared to the X implementation.

johnnovak:
> I appreciate the confidence @johnnovak, but I don't think it's good in current form. It's too spread out and doesn't produce as good of test results compared to the X implementation.

Well, then I hope you'll be able to pick it up next year at some point once you'll have more time 😄

PR #4098, opened 2024-12-09
Fix debugger keypresses not being registered on Windows

### LINKED ISSUE #4089: Dosbox Debugger does not react on key presses

_No response_

Windows 10

_No response_

### Relevant hardware info

_No response_

(already discussed on discord but having a real issue is better then some comments somewhere)

the builtin debugger of dosbox does not react to keypresses - that worked in former releases (and also under linux)

minor: also the [x] close icon doesn't close the debugger-window

on discord: interloper98 found the relevant commit: 

https://github.com/dosbox-staging/dosbox-staging/commit/d225c9a619fc69de19a68da473dcfa244862582d

that seems to break the keys in the debugger

1. build Dosbox with config.h `#define C_DEBUG 1`
2. start dosbox
3. in console `debug attrib.com` then the debug windows comes up
4. press F5(run) or F10(single step) - nothing happens

_No response_

### Your configuration

_No response_

### Provide a Log

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

Regression from either d225c9a619fc69de19a68da473dcfa244862582d or 0b1f510683b36d46440a781f8e151b7f942a9f23 Latter commit would not build so I can't confirm

One of those locale changes changed the behavior of the toupper() function Wrote my own ascii_to_upper() function to match expectations in the debugger code

## Related issues

Fixes #4089

# Release notes

Don't think this was ever in a release

# Manual testing

- [x] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/include/support.h
+++ b/include/support.h
@@ -65,6 +65,26 @@ constexpr uint16_t dos_pagesize = 4096;
 // is in-range of a char and returns it as such.
 char int_to_char(int val);
 
+// A change in one of these 2 commits changed the behavior of toupper() on Windows with MSVC
+// d225c9a619fc69de19a68da473dcfa244862582d - Probably this giant commit
+// 0b1f510683b36d46440a781f8e151b7f942a9f23 - This one won't compile so I can't rule it out
+
+// I found some documentation that states toupper() can change behavior based on locale
+// However, the debugger depends on this specific behavior
+// It passes in values larger than an 8 bit char can hold and needs those to stay intact
+
+// Not even sure if this is defined behavior (might be one of those janky C functions that casts int to char)
+// There are several other instances of toupper() throughout the code that should be looked at
+// This is a workaround for the specific case of the debugger for now
+// We may choose to remove this code later or use it elsewhere where only ascii letters should be considered
+constexpr int ascii_to_upper(int c)
+{
+	if (c >= 'a' && c <= 'z') {
+		return c - 'a' + 'A';
+	}
+	return c;
+}
+
 constexpr bool char_is_negative([[maybe_unused]] char c)
 {
 #if (CHAR_MIN < 0) // char is signed
--- a/src/debug/debug.cpp
+++ b/src/debug/debug.cpp
@@ -1079,7 +1079,7 @@ bool ChangeRegister(char* str)
 bool ParseCommand(char* str) {
 	char* found = str;
 	for(char* idx = found;*idx != 0; idx++)
-		*idx = toupper(*idx);
+		*idx = ascii_to_upper(*idx);
 
 	found = trim(found);
 	std::string s_found(found);
@@ -1743,7 +1743,7 @@ uint32_t DEBUG_CheckKeys(void) {
 			break;
 		}
 #endif
-		switch (toupper(key)) {
+		switch (ascii_to_upper(key)) {
 		case 27:	// escape (a bit slow): Clears line. and processes alt commands.
 			key=getch();
 			if(key < 0) { //Purely escape Clear line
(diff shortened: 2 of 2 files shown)


include/support.h
```diff
@@ -65,6 +65,27 @@ constexpr uint16_t dos_pagesize = 4096;
 // is in-range of a char and returns it as such.
 char int_to_char(int val);
 
+// A change in one of these 2 commits changed the behavior of toupper() on Windows with MSVC
```
FeralChild64, line 68: According to Microsoft, the `toupper` should only be called for ASCII characters, otherwise the results might be unexpected: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/toupper-toupper-towupper-toupper-l-towupper-l?view=msvc-170#remarks

It seems when I started asking the standard C++ library about various locale formats, the Microsoft implementation (and only this one) changed the `toupper` behavior.

The comment needs to be updated, and we need to check all the other `toupper` calls in DOSBox Staging.

include/support.h
```diff
@@ -65,6 +65,27 @@ constexpr uint16_t dos_pagesize = 4096;
...
+// It passes in values larger than an 8 bit char can hold and needs those to stay intact
+
+// Not even sure if this is defined behavior (might be one of those janky C functions that casts int to char)
+// There are several other instances of toupper() throughout the code that should be looked at
+// This is a workaround for the specific case of the debugger for now
+// We may choose to remove this code later or use it elsewhere where only ascii letters should be considered
+constexpr int ascii_to_upper(int c)
+{
+	if (c >= 'a' && c <= 'z') {
+		constexpr int UpperDiff = 'a' - 'A';
```
FeralChild64, line 83: I think this `constexpr` is an overkill. `return c - 'a' + 'A'` should be enough.

weirddan455: Yeah, that's probably better.  I was going to write `return c - 32;` at first but thought better of it.

johnnovak: Yeah, it must be one of those old dirty C functions that cheerfully assume ASCII, no doubt 😆

include/support.h
johnnovak: Honestly, when weird shit like this surfaces I wish DOSBox was written in Java 😎 

No, really. Our build issues would go away and all the platform specific bullshit.

I'm like I'm transported back to the 60s when I see messy dodgy stuff like this... 😆 These are solved problems since Java 1.0 in the 90s.

weirddan455: >  I wish DOSBox was written in Java

stinky-clint-eastwood

johnnovak (approved):
Looks fine to me @weirddan455 , thanks for fixing this.

LowLevelMahn:
works for me - also Alt+Pause for jumping into the debugger is working

include/support.h
weirddan455: I looked at where `getch()` gets they key code from.  It's a PDCurses function that goes through a couple levels of parsing but ultimately the data comes from SDL events.  I don't think this should be affected by the locale changes so this is correct for the debugger.

I think it's best to merge this for now to fix the debugger problem.  I wrote that lengthy comment so we could go back and examine other uses of `toupper()` at a later date.

weirddan455: I'll also echo what @FeralChild64 said in that this usecase of `toupper()` was probably a bug all along that just accidentally worked until the locale changed.  Not everything it passes is even a character (ex. the F5 key which is above the 255 that an 8 bit int can even hold).

If this turns out to be an edge case only applicable to the debugger, we can consider moving this function to `debug.cpp` and not worry about it elsewhere.

johnnovak:
I'm confused. Did we break Alt+Pause and/or the debugger handling again on `main`?

**Alt+Pause** doesn't bring up the debugger in this recent build off `main`:
https://github.com/dosbox-staging/dosbox-staging/actions/runs/12591377678

All the while **Alt+Pause** _does_ bring up the debugger in 0.82.0. I'm testing on Win 10 Pro.

weirddan455:
Discussed on Discord but just putting here for reference:  The mapper on a non-debugger build does not set the debugger hotkey.  Instead it sets Alt + Pause to "Pause emulator".  Solution is to either delete the old mapper file or use the mapper to set the Debugger hotkey from a debugger build.

I think the issue in #4140 is unrelated to this.  Mapper still works fine for me in `main`.

PR #4099, opened 2024-12-09
Don't assert on empty log message

weirddan455:
# Description

I'm getting this assert fail + crash in a debug build since the recent locale changes.  I don't think this should be an assert.  I'm using KDE but my `.config/kxkbrc` file does not contain a `[Layout]` section so it fails to detect.

Probably should change the log message but I don't know what it does exactly if this fails :laughing: 

# Manual testing

~~Tested on Linux.  This is Linux only code.~~

Scratch that.  I think this is actually cross-platform code but all I'm doing is removing an assert.  The `GetHostKeyboardLayouts()` function right above the assert calls into platform-specific code.

# Checklist

I have:

### DIFF
--- a/src/dos/dos_locale.cpp
+++ b/src/dos/dos_locale.cpp
@@ -1404,9 +1404,12 @@ static void load_keyboard_layout()
 
 	if (using_detected) {
 		const auto& host_locale = GetHostKeyboardLayouts();
-		assert(!host_locale.log_info.empty());
-		LOG_MSG("LOCALE: Keyboard layout and code page detected from '%s'",
-		        host_locale.log_info.c_str());
+		if (host_locale.log_info.empty()) {
+			LOG_MSG("LOCALE: Unable to detect keyboard layout. Falling back to default or some shit.");
+		} else {
+			LOG_MSG("LOCALE: Keyboard layout and code page detected from '%s'",
+					host_locale.log_info.c_str());
+		}
 	}
 
 	// Apply the code page


johnnovak (approved):
Looks alright to me.

src/dos/dos_locale.cpp
```diff
@@ -1404,9 +1404,12 @@ static void load_keyboard_layout()
 
 	if (using_detected) {
 		const auto& host_locale = GetHostKeyboardLayouts();
-		assert(!host_locale.log_info.empty());
-		LOG_MSG("LOCALE: Keyboard layout and code page detected from '%s'",
-		        host_locale.log_info.c_str());
+		if (host_locale.log_info.empty()) {
```
FeralChild64, line 1407: I would prefer to keep the assert - we want the host-specific code to fill-in the `log_info`, so that the log contains the information telling why the selected keyboard layout was even considered in the first place.

We should probably move the `LOG_MSG` to more appropriate place - please wait for me with this, I should have time to look into this today evening, or latest tomorrow.

FeralChild64: I have prepared the fix preserving the assertion - https://github.com/dosbox-staging/dosbox-staging/tree/fc/kb-autodetect-fixes-1 - just need some more testing.

I also came to conclusion that preferring CPI files font over ROM font for auto-detected keyboard layouts was probably not a good idea. I have changed this, too.

FeralChild64:
@weirddan455 I'll close this PR - the fix was just merged, https://github.com/dosbox-staging/dosbox-staging/pull/4103

Issue #2992, opened 2023-10-12
Default volume levels are too high, we should think about using realistic headrooms


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

None

### Is your feature request related to a problem? Please describe.

> Yeah @rderooy as James-F has pointed it out, our default volume levels are too high. I set the master to 40 in my global config, so at some point we should think about using realistic headrooms.

https://github.com/dosbox-staging/dosbox-staging/issues/2876#issuecomment-1728490219

Related:
https://github.com/dosbox-staging/dosbox-staging/issues/2467

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Grounded0:
`PCSPEAKER` at 100 feels like someone is trying to rip your soul off when an error beep played.

GranMinigun:
I wonder why people keep setting output volumes per-app instead of setting it once globally...

johnnovak:
> I wonder why people keep setting output volumes per-app instead of setting it once globally...

I set the MASTER volume to 40 in my primary config too, but the audio interfaces should default to realistic headrooms still. Especially the SB and the GUS, the two most common options.

johnnovak:
@Burrito78 
The more I think about it, the more I think we don't wanna dick around with the relative volume balances of different audio devices.

People already have configs there with fine-tuned balances between MIDI and digital sound, or MIDI and PC Speaker, or any two (or three) combinations of devices. Changing the volume levels of any single device would upset this balance for not much benefit.

We could never achieve a "theoretically perfect" balance, such thing does not exist. All Sound Blaster models and clones have different volume levels, for a start. People just needed to tweak things per game; that is why in-game volume controls were invented for music and speech/effects.

Also, we do not emulate per-device headrooms. We just generate the audio stream as float values, do the mixing in float-space, then only convert to integers at the very last step before sending the output to the operating system. Well, even that might go away and we could sent a float stream to SDL, it's possible.

Per-device headroom emulation is unnecessary, anyway; I doubt any game exploited the headroom of any given soundcard in meaningful and "creative" ways (e.g., to intentionally overdrive the output to get some sort of distortion for free).  And with a float pipeline, headrooms become meaningless as float samples have a virtually infinite headroom and dynamic range.

What I am considering doing is to simply just set the default master volume to 50, and that's it.

[...]

PR #4105, opened 2024-12-14
Lower default master volume

johnnovak:
# Description

I've been using DOSBox with 50% master volume (`mixer master 50` in my primary config) for years now as the default volume levels are just too hot. The default levels can easily result in clipping or overworking the compressor, which results in audible pumping.

Other users have come to the similar conclusions that it's best to reduce the master volume to the 40-60 range as a starting point (see [here](https://github.com/dosbox-staging/dosbox-staging/issues/2992#issuecomment-2480921229)).

Reducing the master volume is preferable to changing the volume balance of individual sound devices. First of all, we can never get it right as all Sound Blaster models and their clones have different volume levels and OPL / DAC balances, and even more importantly, we don't want to upset the fine-tuned channel balances people have achieved in their configs. See my detailed analysis [in this comment](https://github.com/dosbox-staging/dosbox-staging/issues/2992#issuecomment-2480907604)

Also related, emulating the actual headroom of the audio devices is out of scope as emulating running out of headroom is a non-desirable characteristic (similarly to hiss, crosstalk, and hearing the hard drive motor's activity in the Sound Blaster's output 😆). We use a float mixing path now with infinite per-device headroom, so the only thing we need to get right is not overloading the host audio interface at the final float to 16-bit integer conversion of the summed stereo master channel.

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/issues/2992

# Release notes

The default master channel volume has been reduced to 50% (-6 dB) in the mixer to avoid clipping.

You can use the `mixer master 100` DOS command to restore the previous behaviour, but it's advised to only do that for games that are too quiet on a case by case basis, and only raise the volume of the channels that needs a bit of a boost (e.g., `mixer sb 200` to make the Sound Blaster digital output twice as loud).

# Manual testing

image

- [ ] Windows
- [x] macOS (trivial change)
- [x] Linux

# Checklist

I have:

### DIFF
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -233,12 +233,17 @@ public:
 	void SetResampleMethod(const ResampleMethod method);
 	void SetZeroOrderHoldUpsamplerTargetRate(const int target_rate_hz);
 
+	// The crossfeed strength is a perceptually linear scale from 0.0
+	// to 1.0. A value of 0.0 means no crossfeed, and 1.0 means the stereo
+	// signal is collapsed into mono.
 	void SetCrossfeedStrength(const float strength);
 	float GetCrossfeedStrength() const;
 
+	// The reverb level is a perceptually linear scale from 0.0 to 1.0.
 	void SetReverbLevel(const float level);
 	float GetReverbLevel() const;
 
+	// The chorus level is a perceptually linear scale from 0.0 to 1.0
 	void SetChorusLevel(const float level);
 	float GetChorusLevel() const;
 
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -261,7 +261,7 @@ public:
 	bool WakeUp();
 
 	std::vector<AudioFrame> audio_frames = {};
-	std::recursive_mutex mutex = {};
+	std::recursive_mutex mutex           = {};
 
 	std::atomic<bool> is_enabled = false;
 
@@ -270,7 +270,6 @@ public:
 		float send_gain = 0.0f;
 	} reverb            = {};
 	bool do_reverb_send = false;
-
 
 	struct {
 		float level     = 0.0f;
@@ -427,7 +426,7 @@ private:
 			int order          = 0;
 			int cutoff_freq_hz = 0;
 		} lowpass = {};
-	} filters               = {};
+	} filters = {};
 
 	struct {
 		float strength  = 0.0f;
@@ -480,14 +479,15 @@ void MIXER_SetReverbPreset(const ReverbPreset new_preset);
 ChorusPreset MIXER_GetChorusPreset();
 void MIXER_SetChorusPreset(const ChorusPreset new_preset);
 
-// Generic callback used for audio devices which generate audio on the main thread
-// These devices produce audio on the main thread and consume on the mixer thread
-// This callback is the consumer part
+// Generic callback used for audio devices which generate audio on the main
+// thread These devices produce audio on the main thread and consume on the
+// mixer thread This callback is the consumer part
 template <class DeviceType, class AudioType, bool stereo, bool signeddata, bool nativeorder>
 inline void MIXER_PullFromQueueCallback(const int frames_requested, DeviceType* device)
 {
-	// Currently only handles mono sound (output_queue is a primitive type and frames == samples)
-	// Special case for AudioType == AudioFrame (stereo floating-point sound)
+	// Currently only handles mono sound (output_queue is a primitive type
+	// and frames == samples) Special case for AudioType == AudioFrame
+	// (stereo floating-point sound)
 	static_assert((!stereo) || std::is_same_v<AudioType, AudioFrame>);
 
 	// AudioFrame type is always stereo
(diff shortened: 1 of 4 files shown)


FeralChild64:
I use master level of 80 dy default. At least on my system going down to 40 gives a huge volume difference between the DOSBox and other apps.

include/mixer.h
```diff
@@ -480,14 +484,15 @@ void MIXER_SetReverbPreset(const ReverbPreset new_preset);
 ChorusPreset MIXER_GetChorusPreset();
 void MIXER_SetChorusPreset(const ChorusPreset new_preset);
 
-// Generic callback used for audio devices which generate audio on the main thread
-// These devices produce audio on the main thread and consume on the mixer thread
-// This callback is the consumer part
+// Generic callback used for audio devices which generate audio on the main
```
FeralChild64, line 487: The beautifier went too far here. I suggest to manually adjust line breaks, or introduce full stops at the end of sentence.

include/mixer.h
```diff
@@ -480,14 +484,15 @@ void MIXER_SetReverbPreset(const ReverbPreset new_preset);
...
+// thread These devices produce audio on the main thread and consume on the
+// mixer thread This callback is the consumer part
 template <class DeviceType, class AudioType, bool stereo, bool signeddata, bool nativeorder>
 inline void MIXER_PullFromQueueCallback(const int frames_requested, DeviceType* device)
 {
-	// Currently only handles mono sound (output_queue is a primitive type and frames == samples)
-	// Special case for AudioType == AudioFrame (stereo floating-point sound)
+	// Currently only handles mono sound (output_queue is a primitive type
+	// and frames == samples) Special case for AudioType == AudioFrame
```
FeralChild64, line 494: Again, the comment should be formatted manually.

include/mixer.h
```diff
@@ -496,22 +501,29 @@ inline void MIXER_PullFromQueueCallback(const int frames_requested, DeviceType*
...
+		                                  MaxExpectedFastForwardFactor));
 	} else {
-		// Normal case, resize the queue to ensure we don't have high latency
-		// Resize is a fast operation, only setting a variable for max capacity
-		// It does not drop frames or append zeros to the end of the underlying data structure
-
-		// Size to 2x blocksize. The mixer callback will request 1x blocksize.
-		// This provides a good size to avoid over-runs and stalls.
-		device->output_queue.Resize(iceil(device->channel->GetFramesPerBlock() * 2.0f));
+		// Normal case, resize the queue to ensure we don't have high
```
FeralChild64, line 517: Again, full stops or manual formatting is needed here.

src/hardware/mixer.cpp
```diff
@@ -280,24 +285,19 @@ bool MIXER_FastForwardModeEnabled()
 }
 
 // The queues listed here are for audio devices that run on the main thread
-// The mixer thread can be waiting on the main thread to produce audio in these queues
-// We need to stop them before aquiring a mutex lock to avoid a deadlock
-// These are called infrequently when global mixer state is changed (mostly on device init/destroy and in the MIXER command line program)
-// Individual channels also have a mutex which can be safely aquired without stopping these queues
+// The mixer thread can be waiting on the main thread to produce audio in these
```
FeralChild64, line 288: Again, full stops or manual formatting is needed here.

src/hardware/mixer.cpp
```diff
@@ -2448,9 +2482,11 @@ static void SDLCALL mixer_callback([[maybe_unused]] void* userdata,
 
 	const auto frames_requested = check_cast<size_t>(bytes_requested /
 	                                                 BytesPerAudioFrame);
-	// Mac OSX has been observed to be problematic if we ever block inside SDL's callback
-	// This ensures that we do not block waiting for more audio
-	// In the queue has run dry, we write what we have available and the rest of the request is silence
+
+	// Mac OSX has been observed to be problematic if we ever block inside
```
FeralChild64, line 2486: Again, full stops or manual formatting is needed here.

src/hardware/mixer.cpp
```diff
@@ -2472,61 +2508,88 @@ static void mixer_thread_loop()
...
-		const double expected_time = (static_cast<double>(mixer.blocksize) / static_cast<double>(mixer.sample_rate_hz)) * 1000.0;
+		const double expected_time = (static_cast<double>(mixer.blocksize) /
+		                              static_cast<double>(mixer.sample_rate_hz)) *
+		                             1000.0;
 		last_mixed = now;
 
 		// "Underflow" is not a concern since moving to a threaded mixer
-		// If the CPU is running slower than real-time, the audio drivers will naturally slow down the audio
-		// Therefore, we can always request at least a blocksize worth of audio
+		// If the CPU is running slower than real-time, the audio
```
FeralChild64, line 2519: Again, full stops or manual formatting is needed here.

src/hardware/mixer.cpp
```diff
@@ -2472,61 +2508,88 @@ static void mixer_thread_loop()
...
 			mixer.output_buffer.clear();
 			mixer.output_buffer.resize(mixer.blocksize);
+
 			mixer.final_output.BulkEnqueue(mixer.output_buffer);
 			continue;
 		}
 
-		// Only true if we were in fast-forward mode at the time we calculated frames_requested
-		// That variable could have changed by now but we need to always squash down to a blocksize of audio
+		// Only true if we were in fast-forward mode at the time we
```
FeralChild64, line 2563: Again, full stops or manual formatting is needed here.

src/hardware/mixer.cpp
johnnovak: Cheers, I fixed up a few but did not read all comments. @weirddan455 I'd suggest to get into the habit of using punctuation in comments 😄 These were all yours from the mixer refactoring. Very helpful comments, so please keep commenting; just a minor suggestion to help with reformatting 😄

johnnovak:
> I use master level of 80 dy default. At least on my system going down to 40 gives a huge volume difference between the DOSBox and other apps. Maybe 60 would be a better default after all?
> 
> Otherwise, some minor formatting-related comments.

Yeah 80 is a bit on the high side to get the benefits. I was hesitating between 40 and 50, you're saying 60, so we can go with 50 😆

The main thing is to cut the default volumes roughly in half to avoid clipping and overloading the compressor too early when we get close to the 0dB ceiling.

johnnovak:
> I use master level of 80 dy default. At least on my system going down to 40 gives a huge volume difference between the DOSBox and other apps. Maybe 60 would be a better default after all?
> 
> Otherwise, some minor formatting-related comments.

Thanks, fixed up the comments and went with 50 (-6 dB) in the end.

johnnovak:
@Python-Exoproject You might want to put `mixer master 100` in your config if you want to keep the same volume levels across all forks. Personally, I'd recommend using `mixer master 50` to bring down the volume of the other forks to Staging.

FeralChild64 (approved):

PR #4106, opened 2024-12-14
Fix crash if opening the `soundfont_dir` directory failed

johnnovak:
# Description

Setting `soundfont_dir` to a directory that cannot be opened and then executing `mixer /listmidi` or `mididevice fluidsynth` resulted in a crash. Nasty! Somehow this evaded my attention when I introduced the setting...

This PR fixes that.

## Related issues

N/A

# Release notes

N/A (`soundfont_dir` was introduced in the current dev cycles)

# Manual testing

Let me tell you in pictures 😎 

image

- [x] Windows
- [x] macOS
- [x] Linux (I assume it works too, but can someone verify, please?)

# Checklist

I have:

### DIFF
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -197,7 +197,7 @@ static std::vector<std_fs::path> get_data_dirs()
 		// `find_sf_file()` as well).
 		if (path_exists(sf_dir)) {
 			std::error_code err = {};
-			const auto canonical_path = std_fs::canonical(sf_dir, err);
+			const auto canonical_path = std_fs::canonical(sf_dir, err); //-V821
 			if (!err) {
 				dirs.insert(dirs.begin(), canonical_path);
 			}
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -68,8 +68,8 @@ static void init_fluidsynth_dosbox_settings(Section_prop& secprop)
 	        "the standard system locations.");
 
 	constexpr auto DefaultVolume = 100;
-	constexpr auto MinVolume = 1;
-	constexpr auto MaxVolume = 800;
+	constexpr auto MinVolume     = 1;
+	constexpr auto MaxVolume     = 800;
 
 	auto int_prop = secprop.Add_int("soundfont_volume", WhenIdle, DefaultVolume);
 	int_prop->SetMinMax(MinVolume, MaxVolume);
@@ -463,9 +463,7 @@ static void setup_reverb(fluid_synth_t* synth)
 // Current API calls as of 2.2
 #if FLUIDSYNTH_VERSION_MINOR >= 2
 	fluid_synth_reverb_on(synth, FxGroup, reverb_enabled);
-	fluid_synth_set_reverb_group_roomsize(synth,
-	                                      FxGroup,
-	                                      reverb_room_size);
+	fluid_synth_set_reverb_group_roomsize(synth, FxGroup, reverb_room_size);
 
 	fluid_synth_set_reverb_group_damp(synth, FxGroup, reverb_damping);
 	fluid_synth_set_reverb_group_width(synth, FxGroup, reverb_width);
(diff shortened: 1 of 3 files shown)


src/midi/midi_fluidsynth.cpp
```diff
@@ -195,7 +195,16 @@ static std::vector<std_fs::path> get_data_dirs()
 		// of the actual path on Linux & Windows, so we need to
 		// normalise that to avoid some subtle bugs downstream (see
 		// `find_sf_file()` as well).
-		dirs.insert(dirs.begin(), std_fs::canonical(sf_dir));
+		if (path_exists(sf_dir)) {
+			dirs.insert(dirs.begin(), std_fs::canonical(sf_dir));
```
weirddan455, line 199: I would consider using the form of `canonical` that returns an error code and never throws an exception:  https://en.cppreference.com/w/cpp/filesystem/canonical

Up to you.  I'll go ahead and approve either way.  Your way fixes the crash but I think it's more robust to prefer functions that don't throw exceptions as we really don't handle them properly.

weirddan455 (approved):
I can confirm the crash on Linux and that this fixes it.  As for my gripe about the (long standing) bug we have with exceptions:

```
2024-12-14 04:15:35.505 | EXCEPTION: Standard library exception: filesystem error: cannot make canonical path: No such file or directory [asdf]
2024-12-14 04:15:35.539 | atexit
2024-12-14 04:15:35.539 | MPU-401: Shutting down
2024-12-14 04:15:35.539 | OPL: Shutting down OPL3
2024-12-14 04:15:35.539 | SB16: Shutting down
2024-12-14 04:15:35.539 | PCSPEAKER: Shutting down impulse model
terminate called without an active exception

corrupted size vs. prev_size in fastbins

^C
Loguru caught a signal: SIGINT

Loguru caught a signal: SIGTERM

Loguru caught a signal: SIGTERM
Killed
```

This is what I get on `main`.  Some error about heap corruption and then it hangs indefinitely ignoring both `SIGINT` and `SIGTERM`.  I have to go into my system monitor and give it a `SIGKILL` to exit the process.

johnnovak:
Thanks for testing this on Linux @weirddan455 🎖️ 

> This is what I get on `main`. Some error about heap corruption and then it hangs indefinitely ignoring both `SIGINT` and `SIGTERM`. I have to go into my system monitor and give it a `SIGKILL` to exit the process.

I believe I fixed the MIDI subsystem shutdown problem here:
https://github.com/dosbox-staging/dosbox-staging/pull/4090/commits/7ae1f7e00bd371f4f3e0705c8777af027ee07c50

weirddan455:
> I believe I fixed the MIDI subsystem shutdown problem here in my Sound Canvas PR: https://github.com/dosbox-staging/dosbox-staging/commit/7ae1f7e00bd371f4f3e0705c8777af027ee07c50

Different bug.  I get this hang (with uncaught exceptions) on that PR as well.  I get it even with `mididevice = none`.

My theory is that the root cause is the same though.  When an exception gets thrown, it doesn't run our `Destroy()` functions.  It's just cleaning up global objects in some nondeterministic order.  We probably have underlying memory corruption bugs lurking in the codebase that get triggered when this happens.  Adding the `MIDI_Destroy()` function is a band-aid (although a pretty good band-aid because it fixes the normal exit path).

src/midi/midi_fluidsynth.cpp
johnnovak: Good point, forgot about that. Changed, plus also fixed up `cdrom_ioctl_linux.cpp`.

johnnovak:
> > I believe I fixed the MIDI subsystem shutdown problem here in my Sound Canvas PR: [7ae1f7e](https://github.com/dosbox-staging/dosbox-staging/commit/7ae1f7e00bd371f4f3e0705c8777af027ee07c50)
> 
> Different bug. I get this hang (with uncaught exceptions) on that PR as well. I get it even with `mididevice = none`.
> 
> My theory is that the root cause is the same though. When an exception gets thrown, it doesn't run our `Destroy()` functions. It's just cleaning up global objects in some nondeterministic order. We probably have underlying memory corruption bugs lurking in the codebase that get triggered when this happens. Adding the `MIDI_Destroy()` function is a band-aid (although a pretty good band-aid because it fixes the normal exit path).

Yeah it's not easy for handle these cases in a cleaner without rewriting the whole config system, unfortunately. If you still can trigger some deterministic shutdown problems after this and the Sound Canvas changes go in, please raise a ticket. Sadly they're also somewhat compiler and OS specific.

Issue #4108, opened 2024-12-17
Website shows macOS 10.15 as supported, but it isn't anymore


Torinde:
> MacBook Pro 11.1 -- so I take it that's the problem; there's now a big enough gap between github CI machines and your system that the runtime libraries aren't compatible.

So if you're running that [2014 MacBook model](https://www.londontechandservices.com/product/apple-macbookpro111-a1502-emc-2875-intel-i5-4278u-2-60ghz-8gb-250gb-ssd/), the most recent macOS version you can run on it is Big Sur (11.x). We only support the last three major versions in Staging for every supported OS (and even that is a best case scenario, it might be less than the last three if that's overly difficult), so that is macOS 13.x currently.

11.x is two full revisions behind that.

https://eshop.macsales.com/guides/Mac_OS_X_Compatibility

> @johnnovak, are there any static build tricks vcpkg can do on MacOS to fix this? Anyone with a mac know how to build a 100% static binary?

I'm not aware of any such tricks and have low interest in figuring it out. I don't think it's possible without some tricks, and if it requires compiling on an older macOS version than what GitHub provides, it's not going to happen.

If I was doing this commercially, yeah I'd have 10-15 old laptops laying around to test stuff and perhaps to compile the plugins on the oldest possible OS revisions to ensure compatibility. But this is a hobby, so I'm uninterested in sinking time into this.

However, if someone is super into this and figures out a way to make it happen on GitHub CI without much complication, I'll merge the PR. Note that currently I'm building the universal Mac binaries locally on my machine; it's surely doable on CI too, but I haven't bothered making my `build-macos.sh` script and the `run-vcpkg` actions work together.

PS: I think that error happens in Staging, not in my CLAP plugin, btw. Anyway, the exact same reasoning applies. We'll not use a self-hosted CI machine like previously for this—too much effort.

@jowtron Sorry to disappoint, but there would be too many difficulties involved in supporting legacy OSes. macOS 11.x is no longer supported by Apple since 30 November 2013. It's a bit unrealistic to expect us to support it after Apple's own commercial support has ended.

_Originally posted by @johnnovak in https://github.com/dosbox-staging/dosbox-staging/issues/4090#issuecomment-2547522166_
            
So it seems macOS 13 is the minimum supported by Staging releases, but the website [still mentions 10.15](https://www.dosbox-staging.org/releases/macos/).

[...]

PR #4111, opened 2024-12-18
Revert object pool stuff

johnnovak:
# Description

This reverts the object pool stuff introduced by @interloper98 that increases the macOS requirements from 11.x to 14.x.

Sorry man, it's just not worth it. I know you were able to "prove" this improves things marginally, but those were synthetic benchmarks that might never matter under real-world usage. However, increasing the minimum system requirements by two major versions _very much_ affects users.

Full disclosure: I agreed to the inclusion of this complication reluctantly because I never really liked it. Just didn't want to curb your enthusiasm 😄 But it's problematic, so I think it has to go.

So yeah, probably you could make it work without using newer C++ features. But again, I never thought adding this complication was warranted in the first place (I was just like "yeah, whatever 🤷🏻 😄"); it doesn't solve any real-world problem we're experiencing.

## Related issues

See discussion here:
- https://github.com/dosbox-staging/dosbox-staging/issues/4108

# Release notes

N/A (only affects 0.83.x)

# Manual testing

TODO

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


kklobe:
I agree with this whole-heartedly. I've been stymied a couple of times trying to add new C++20 features, but they caused issues with various platforms and compilers, so I understand the enthusiasm to do this. But in this case, it's just not worth it; as @johnnovak mentions, this doesn't appear to be solving any real problems.

weirddan455:
To play devil's advocate, we could just turn this feature off for Mac since it works perfectly fine on Windows and Linux.

I don't understand what this code is doing though so I'm not right person to give an informed decision.  I know the concept of an object pool but this goes over my head.  Maybe give @interloper98 a chance to weigh in on what the benefits of keeping this are?

interloper98:
No big deal. Just ran the heap profiler and saw allocations were in the tens-of-thousands per second (one of them being memory used by the dynamic core) and it was an easy fix.
Profiling dosbox SVN showed it has near zero ongoing heap allocations,  so I was trying to get back to baseline.
Approving.

interloper98 (approved):

interloper98:
I think what should happen is just more careful use of static vectors; let them allocate up to the edge of what the software uses and then  reusing that memory space through the rest of the runtime period.

johnnovak:
> I think what should happen is just more careful use of static vectors; let them allocate up to the edge of what the software uses and then reuse that memory space through the rest of the runtime period.

Yeah I agree with that. Preallocating even a few megabytes for vectors like that is nothing these days. Doesn't even matter much for mobile phones, lol 😄 

Memory pools can make a difference, but we should always measure total app performance—the fact that 10k allocations happen per second and it's "unpooled" might or might not matter.

This is a good write-up about it (explaing why it _might_ make difference, but why it usually doesn't):
https://stackoverflow.com/a/67629914

johnnovak:
> To play devil's advocate, we could just turn this feature off for Mac since it works perfectly fine on Windows and Linux.

The overarching line of thought here is not adding any complexity where it doesn't make a difference. So why would you do that if it doesn't make a difference? The previous code is simpler. Simpler == better, all other things being equal (or _mostly_ equal, even).

Issue #4093, opened 2024-12-07
keyboard_layout=auto does not work on Fedora 40, it works on Ubuntu 24.10


LowLevelMahn:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

git

### What Operating System are you using?

Linux x86_64

### If Other OS, please describe

Fedora 40, Ubuntu 24.10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Dosbox does not correctly detect my german-keyboard with english Fedora 40
but my english Ubuntu 24.10 + german keyboard works

fresh Linux installes, fresh dosbox release build, fresh config + keyboard_layout=auto 

i think the mru-sources and XDG_* env-vars are not handled correctly in my case

**in Fedora 40 (English and german layout installed: german is active)**
  `echo $XDG_CURRENT_DESKTOP` ==> `GNOME`
  `echo $XDG_SESSION_DESKTOP` ==> `gnome`
  `gsettings get org.gnome.desktop.input-sources sources` ==> `[('xkb', 'us'), ('xkb', 'de+nodeadkeys')]`
  `gsettings get org.gnome.desktop.input-sources mru-sources` ==> `('xkb', 'de+nodeadkeys'), ('xkb', 'us')]`

the first of mru-sources seems to be the current one - follows my switching of layout

**in Ubnutu 24.10 (only german layout installed)** , **SUSE Tumbleweed behaves the same as Ubuntu with gsettings but the envs are like in Fedora with GNOME and gnome** 
  `$XDG_CURRENT_DESKTOP` ==> `ubuntu:GNOME`
  `$XDG_SESSION_DESKTOP` ==> `ubuntu`
  `gsettings get org.gnome.desktop.input-sources sources` ==> `[('xkb', 'de')]`
  `gsettings get org.gnome.desktop.input-sources mru-sources` ==> `@a(ss) []`
  
`@a(ss) []` means empty list (use only first entry of sources)

when i add English as layout to Ubuntu i get the following
  `gsettings get org.gnome.desktop.input-sources sources` ==> `[('xkb', 'de'), ('xkb', 'us')]`
  `gsettings get org.gnome.desktop.input-sources mru-sources` ==> `[('xkb', 'de')]`
  
  changing to english layout

[...]

PR #4116, opened 2024-12-20
Consider keyboard layout list random on GNOME

FeralChild64:
# Description

At least some GNOME variants start with the last used keyboard layout - therefore it is quite possible the user did not reorder the keyboard layouts according to his preference. Thus, consider the order random for the purpose of autodetection.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4093

# Release notes

(not needed; new keyboard layout autodetection is not yet present in any release)

# Manual testing

Set `keyboard_layout = auto`. Now:

- on KDE, DOSBox should, in general, honor the order of keyboard layouts; if English is present before German, English sohuld be selected, if German is present before English, german should be selected
-on GNOME, DOSBox should not care about order when deciding whether English or German layout should be selected; it should now always select German, regardless how the layouts are ordered in the configuration tool

- [ ] Windows
- [ ] macOS
- [x] Linux (only the Linux-specific code was changed)

# Checklist

I have:

### DIFF
--- a/src/misc/host_locale_posix.cpp
+++ b/src/misc/host_locale_posix.cpp
@@ -922,7 +922,7 @@ static std::pair<std::string, std::string> split_posix_locale(const std::string&
 }
 
 static std::optional<DosCountry> get_dos_country(const std::string& category,
-                                                 std::string& log_info)
+                                                 std::string& out_log_info)
 {
 	const std::vector<std::string> Variables = {LcAll, category, VariableLang};
 
@@ -930,7 +930,7 @@ static std::optional<DosCountry> get_dos_country(const std::string& category,
 	if (value.empty()) {
 		return {};
 	}
-	log_info = variable + "=" + value;
+	out_log_info = variable + "=" + value;
 
 	const auto [language, teritory] = split_posix_locale(value);
 	if (is_language_generic(language)) {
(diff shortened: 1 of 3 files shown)


src/misc/host_locale_posix.cpp
```diff
@@ -1247,7 +1247,7 @@ static std::vector<std::string> get_keyboard_layouts_tty()
 #endif // __linux__
 
 static std::vector<KeyboardLayoutMaybeCodepage> get_layouts_maybe_codepages_desktop(
-        std::string& log_info)
+        bool& is_layout_list_sorted, std::string& log_info)
```
johnnovak, line 1250: Consider using a struct for returning the values instead of a combination of an actual return value and out args. I always use this pattern these days in every language; no nore "out args" for me 😏 

```c
struct LayoutList {
   std::vector<KeyboardLayoutMaybeCodepage> layouts;
   std::string is_sorted;
   std::string log_info
}
```

But at the very least, prefix the out args with `out_`.

johnnovak (approved):
One code-organisation related comment, otherwise look alright and I'm not a Linux user 🤷🏻 😄

src/misc/host_locale_posix.cpp
FeralChild64: For now just renamed the variables.

I'll need to do some more significant changes to the host-specific files, to support language preference lists (not just the GUI language), I'm going to experiment with the structures then. I'm just afraid I'll end up with a lot of structures used only once - bu we'll see.

Issue #2619, opened 2023-06-19
Map high-ASCII chars in DOS filenames to UTF-8 on the host


johnnovak:
FAT16 allows all "high ASCII" characters (128-255) to be used in 8.3 filenames under MS-DOS ([reference](https://en.wikipedia.org/wiki/8.3_filename#Directory_table)).

That's all the funky symbols in the lower half of this chart showing the standard default CP437 charset:

<img width="484" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/44b3275b-a740-4f70-a045-2de4a73a64cd">

But who would be mad enough to do such a thing? All the cool demoscene kids from the 90s, brother! 😎  They thought that's a s000per kewl l337 h4xxx0r thing to do! 🤘🏻 

**cUz dAt $h!T iZ Th3rE 2 B UzEd, br0!!!!!!!111**

Case in point, I ran into such things in authentically preserved, mint condition ZIP files delivered to me straight from the 90s! As you can see, I was able to browse the contents of the ZIP file using DOS Navigators built-in archive viewer, but the unpacking failed because DOSBox was not able to map the funky characters to host-side filenames:

<img width="971" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/24b08c77-7141-46e8-a208-68188cb9dc96">

There's some complaints about the failed conversion in the logs:

<img width="710" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/040a03d5-c4a7-46b0-a9b5-77fb5fdef1d7">

Another example:

<img width="923" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/47fda826-4df8-457e-9c64-f81ded0f0c05">

<img width="746" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/45f571ad-8cca-4670-bc2c-983c2e63127f">

Both ZIP files are attached at the end of the ticket.

Just sharing this comical value: the built-in macOS unpack feature in Finder failed to unpack `IMPHOB6.ZIP` with an "invalid format" error. I had better luck with `MYSTIC~1.ZIP`, but I ended up with some Asian lookin characters. Wut? Clearly, there aren't many ex-demosceners at Apple... Americans!

<img width="499" alt="image" src="https://github.com/dosbox-staging/dosbox-staging/assets/698770/34189b6f-9ad5-4d70-97fa-0d965145fca9">

---

[...]

PR #4119, opened 2024-12-22
Unicode engine improvements

FeralChild64:
# Description

This PR brings several small Unicode changes:

1. Added a specialized routines to UTF-8 <-> 437 conversion for the future Unicode support in the filesystem. The 437 -> UTF-8 one is just a syntactic sugar, but UTF-8 -> 437 is simplified and does not handle combining marks - so, for example, for character 0x94 it requires precisely `U+00F6`; `U+006F` followed by `U+0308` (which, from the host OS filesystem perspective differs from `U+00F6`) won't be accepted.

2. There was a minor mistake in code page 1118 definition, this is now fixed (but the character was similar).

3. The engine used to have problems if DOS code point was a mere combining mark (not a regular character + combining mark), it did not even accept such definition.
I have worked this around by silently replacing such definition with SPACE + combining mark, so that now I could add the list of combining marks for all the supported Unicode blocks.
Knowing that some code point is a combining mark is important for the engine, as it strips the combining marks when converting UTF-8 text to DOS if the original grapheme can't be represented in the given code page - i. e. the original UTF-8 grapheme `l̃` can be represented as letter `l` after conversion to DOS.
Currently this only affects one code for which we do provide a screen font (864, Arabic) - but there are few more cases like that in code pages where we do not bundle screen fonts yet.

4. After some testing of the result of point 3. and further investigation, I have added an entry to override the official Unicode mapping, so that we now use an isolated form of the _Arabic Shadda_ instead of the combining mark (probably the character wasn't available in Unicode when the original mapping was created).

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/2619 (tor the extended Unicode API)

# Release notes

(we can mention that our Unicode engine has seen some minor tweaks and improvements, but I wouldn't go into the details)

# Manual testing

1. To test the _Arabic Shadda_ character available at code point `0xf1` of code page `864`, execute inside the DOSBox:
`keyb us 864`
`echo <with ALT pressed type 241 on the numeric keypad> | clip`
and paste the character to the host text editor.

2. To test that all code pages can still be set, execute inside the DOSBox the following BAT script and observe the logs.
[TESTCP.BAT.TXT](https://github.com/user-attachments/files/18223328/TESTCP.BAT.TXT)
Currently it prints out warnings regarding code pages _58335_ and _60258_, but these are not related to this change - the _58335_ really has the same characters at points `0x81` and `0x97` (you can set by executing `keyb us 58335` and check with some ASCII table software, for example from _DOS Navigator_ menu select _Utilities_ -> _ASCII Table_), and the _60258_ is indeed strange - just try to type uppercase letter `I` with it :D

3. Unfortunately, Unicode engine does not function under unit tests (one more thing on my personal TODO list of issues to fix), so I tested it by temporarily adding some test code to one of the commands:

```C++
#include "unicode.h"

	...

	const std::string test1 = std::string("\n\n") +
		fs_utf8_to_dos_437("ÇüéâäàåçêëèïîìÄÅ") + "\n" +
		fs_utf8_to_dos_437("ÉæÆôöòûùÿÖÜ¢£¥₧ƒ") + "\n" +
		fs_utf8_to_dos_437("áíóúñÑªº¿⌐¬½¼¡«»") + "\n" +
		fs_utf8_to_dos_437("░▒▓│┤╡╢╖╕╣║╗╝╜╛┐") + "\n" +
	        fs_utf8_to_dos_437("└┴┬├─┼╞╟╚╔╩╦╠═╬╧") + "\n" +
	        fs_utf8_to_dos_437("╨╤╥╙╘╒╓╫╪┘┌█▄▌▐▀") + "\n" +
	        fs_utf8_to_dos_437("αßΓπΣσµτΦΘΩδ∞φε∩") + "\n" +
	        fs_utf8_to_dos_437("≡±≥≤⌠⌡÷≈°∙·√ⁿ²■ ") + "\n" +
	        "\n";
	WriteOut(test1.c_str());

	const std::string test2 = std::string("\n") +
	        fs_utf8_to_dos_437("abc def ghi jkl mno pqr stu vwx yz ABC 123 ''") + "\n" +
	        fs_utf8_to_dos_437("This line should not be visible due to '\n'") + "\n" +
	        fs_utf8_to_dos_437("This line should not be visible due to 'Ł'") + "\n" +
	        fs_utf8_to_dos_437("This line should not be visible due to 'l̃'") + "\n" +
	        "\n";
	WriteOut(test2.c_str());
```

- [ ] Windows
- [ ] macOS
- [x] Linux (trivial changes)

# Checklist

I have:


include/unicode.h
```diff
@@ -82,6 +82,12 @@ std::string dos_to_utf8(const std::string& str,
                         const DosStringConvertMode convert_mode,
                         const uint16_t code_page);
 
+// Specialized routines for converting between code page 437 and UTF-8 without
+// any combining marks - to be used in filesystem emulation
+
+std::string fs_utf8_to_dos_437(const std::string& str);
+std::string dos_437_to_fs_utf8(const std::string& str);
+
```
johnnovak, line 90: These are welcome additions, cheers 🚀

src/misc/unicode.cpp
```diff
@@ -2185,8 +2219,51 @@ std::string dos_to_utf8(const std::string& str,
...
+	}
+	assert(per_code_page_mappings.contains(CodePage));
+
+	const auto mapping = per_code_page_mappings.at(CodePage).dos_to_grapheme_normalized;
+
+	std::string str_out = {};
+	for (const auto code_point : utf8_to_wide(str)) {
+		if (is_control_code(code_point)) {
+			// Not valid for file/directory name
+			return {};
```
johnnovak, line 2238: Nice you thought of this 💯

johnnovak (approved):
Looks OK as far as I can tell, but I'm not a Unicode expert. Point 1 seems useful, then the Arabic stuff is out of scope for Staging, but I guess the related Unicode engine fixes/enhancements are useful.

FeralChild64:
Some support for Arabic is there - you can set the keyboard layout and code page and type/read texts. But that's it - if right-to-left scripts require some special handling, it is not implemented.

PR #4122, opened 2024-12-27
Enable hotplugging joysticks

arifogel:
# Description

* fixes #946
* remove REDUCE_JOYSTICK_POLLING hack, which prevented detection of joystick hotplug events
* re-initialize mapper when receiving following events:
  * SDL_CONTROLLERDEVICEADDED
  * SDL_CONTROLLERDEVICEREMOVED
  * SDL_JOYDEVICEADDED
  * SDL_JOYDEVICEREMOVED

## Related issues

* #946
* #4124

# Release notes

* Add option to enable hotplugging of joystick devices after program start (enabled by default)

# Manual testing

The change has been manually tested on:

- [x] Windows
- [x] macOS
- [X] Linux

I tested on Ubuntu 24.04 with XBox 360 Wireless Controller (ID 045e:0719 Microsoft Corp. Xbox 360 Wireless Adapter).
I ran DOSBox with and without controller turned on.  I verified that controller still functioned in game in each case after power cycling the controller (which causes device to disconnect and reconnect).

Built using meson on Clion.

Subsequently, I tested on Windows using VS 2022 community edition.

Afterwards, I tested on macOS Ventura 13.7.1. I built using Homebrew. Notably, I needed to `brew install glib`, which was not listed in the list of brew dependencies in the instructions. On macOS the xbox 360 wireless receiver does not work, so instead I tested with:
```
Xbox Wireless Controller:
  Vendor ID:	0x045E
  Product ID:	0x0B13
```

# Checklist

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [x] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [x] my change has been manually tested on Windows, macOS, and Linux.
- [x] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
[...]

PR #4079, opened 2024-11-29
Ensure physical host buttons only perform one bound event

interloper98:
# Description

This PR makes sure that physical host buttons only trigger one logical event. 

For example, if you map the 'F10' key to the screenshot action, then this PR will knock out the 'F10' key's prior logical event of passing the 'F10' keystroke to the DOS program. So now the F10 key would only perform the screenshot action and the DOS program won't get an F10 keypress.

An amber warning in the mapper UI now informs the user when a physical event's prior logical event is dropped (see screenshot below).

This "displacement" also happens when event handlers are programmatically bound to physical buttons, like how the CGA composite code binds the F10, F11, and F12 physical keys to composite actions:

```
MAPPER: Host button 'F10' now performs 'hand_select' instead of 'key_f10'
MAPPER: Host button 'F11' now performs 'hand_incval' instead of 'key_f11'
MAPPER: Host button 'F12' now performs 'hand_cgacomp' instead of 'key_f12'
```

This fixes the problem in Carmen San Diego when switching to composite mode using the F12 also locked the game when it was simultaneously processing the F12 keystroke in the same moment as the video mode switch. 

## Related issues

Fixes #4051.

# Release notes

When using the mapper, only one logical event can be bound to a given physical host button.  For example, if you map the 'F10' key to the screenshot action, then the F10 key will now perform the screenshot action and the DOS program won't get an F10 keypress.

An amber warning in the mapper UI will inform you when a prior logical event is dropped:

remap-in-ui

When booting with `machine = cga`, this change prevents the 1986 edition of Where in the World is Carmen San Diego from locking up when switching to composite video using the F12 key.

# Manual testing

- Tested mapping multiple (unique) physical host buttons to a single logical events (works).
- Tested mapping the same physical host button to multiple logical events (no longer allowed; prior events are dropped).
- Tested and confirmed that with a default configuration and mapping, Carmen San Diego now can switch to composite mode using F12.

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [x] commented on the particularly hard-to-understand areas of my code.
[...]

Issue #946, opened 2021-03-16
Controller hot-plug support


maxigaz:
**Is your feature request related to a problem? Please describe.**
DOSBox Staging detects my DualShock 4 and Wireless XBox 360 Controller only if they are switched on before DOSBox Staging is launched. If any of them is switched off and on again, DOSBox Staging has to be restarted as well so that they can be used again.

**Describe the solution you'd like**
Automatically detect any game controller if it is connected via USB or Bluetooth.

**Describe alternatives you've considered**
Retroarch with DOSBox Pure (and I believe all the other Libretro cores) support this feature.

**Additional context**
From what I remember, the current behaviour is also present in vanilla DOSBox. I've only tested these different versions under Manjaro Linux x86-64.

The change would be useful in case I turn off any of the gamepads to save battery during a break or the XBox Controller switches itself off automatically (which happens after being idle for 15 minutes).

Burrito78:
@nemo93 Do you have an idea how this could be supported?

weirddan455:
I think this will have to wait until we refactor the mapper.  I took a look at this thinking it would be easy to add but `sdl_mapper.cpp` is kind of a disaster.

The best way to support hotplugging (IMO) is to listen for the events `SDL_JOYDEVICEADDED` and `SDL_JOYDEVICEREMOVED`.  However, there's some hack to reduce the number of times the joystick gets polled that calls `SDL_JoystickEventState(SDL_DISABLE);` which tells SDL not to send us joystick events and instead we poll manually at a set interval.  Apparently this is needed for Macs to reduce CPU usage according to a code comment.

OK, second best option is to call `SDL_NumJoysticks()` during the polling interval and see if the number of joysticks changed.  Then call `MAPPER_BindKeys()` to re-init (same as we do if the user changes their mapperfile at runtime).  I tried that and it doesn't work either.  The bindgroups don't get re-created out when that happens because `MAPPER_Destroy` doesn't get called so it's not actually a full destroy + init.

I'm afraid if I go any deeper, I'm likely to cause regressions.  This mapper is storing the same raw pointers in multiple different `std::vectors` and also in some C style arrays with manual indexing.  So making sense of all this while trying to avoid double free, use after free, and memory leaks is going to be challenging.

Best to nuke this file from orbit and re-write it from scratch which I believe @johnnovak is planning to do at some point.  Adding hotplug support is not that difficult but making sense of this mapper code is.

Grounded0:
[...]

PR #4124, opened 2024-12-28
Immediately save user mapper changes to disk

arifogel:
# Description

Immediately save user mapper changes to disk
    
* remove "Save" button in mapper UI
* immediately save any completed user-initiated mapping change to disk

## Related issues
Related to #4122 (see discussion of mapper reloads)
Related to #4126 (had to be resolved first)

# Release notes

Bind changes made in the mapper UI are now immediately saved to disk; the "Save" button has been removed from the mapper UI.

# Manual testing

- [X] Windows
- [X] macOS
- [X] Linux

### Linux testing
Built using meson on CLion on Ubuntu 24.04.
For each of the following actions, I observed log entries stating that binds were saved to disk, and following a restart of DOSBox observed that the changes had persisted:
1. Deleting a bind
1. Adding a guest joystick event bind to a host joystick button
1. Adding a guest joystick event bind to a host joystick axis
1. Adding a guest joystick event bind to a host joystick hat
1. Adding a guest keyboard key event bind to a host keyboard button
1. Adding a guest keyboard key event bind to a host joystick button
1. Adding a guest keyboard key event bind to a host joystick axis
1. Adding a guest keyboard key event bind to a host joystick hat
1. Adding a guest mouse button event bind to a host keyboard key
1. Adding a guest mouse button event bind to a host joystick button
1. Adding a guest mouse button event bind to a host joystick axis
1. Adding a guest mouse button event bind to a host joystick hat

(as of [ec1ca5f](https://github.com/dosbox-staging/dosbox-staging/commit/ec1ca5fa95fa5cf32fcee00a06ba1136a85867bd))
For the regression fix, I tested by mapping keys / joystick axes / joystick buttons - that were already mapped to one event - to another event. In each case, I got a message that the existing input was now mapped to the new event from the old event. Prior to the fix, even just clicking on a joystick event in the mapper caused a crash.

### Windows testing
Built using VS 2022 Community Edition on Windows 11.
For each of the following actions, I observed log entries stating that binds were saved to disk, and following a restart of DOSBox observed that the changes had persisted:
1. Deleting a bind
1. Adding a guest joystick event bind to a host joystick axis
1. Adding a guest keyboard key event bind to a host keyboard button

### macOS testing
Built on macOS Ventura 13.7.1 using homebrew instructions, with:
```
meson setup build
meson compile -C build
```
For each of the following actions, I observed log entries stating that binds were saved to disk, and following a restart of DOSBox observed that the changes had persisted:
1. Deleting a bind
1. Adding a guest joystick event bind to a host joystick axis
1. Adding a guest keyboard key event bind to a host keyboard button

# Checklist

I have:

- [X] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [X] my change has been manually tested on Windows, macOS, and Linux.

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -63,10 +63,7 @@ constexpr Rgb888 color_white(255, 255, 255);
 constexpr Rgb888 color_red(255, 0, 0);
 constexpr Rgb888 color_green(0, 255, 0);
 
-enum BB_Types {
-	BB_Next,BB_Add,BB_Del,
-	BB_Save,BB_Exit
-};
+enum BB_Types { BB_Next, BB_Add, BB_Del, BB_Exit };
 
 enum BC_Types {
 	BC_Mod1,BC_Mod2,BC_Mod3,
@@ -111,6 +108,8 @@ typedef std::list<CBind *>::iterator CBindList_it;
 typedef std::vector<CBindGroup *>::iterator CBindGroup_it;
 
 static CBindList holdlist;
+
+static bool suppress_save_mapper_file_message = false;
 
 class CEvent {
 public:
@@ -1628,8 +1627,12 @@ public:
 				if (mapper.abindit == mapper.aevent->bindlist.end())
 					mapper.abindit=mapper.aevent->bindlist.begin();
 			}
-			if (mapper.abindit!=mapper.aevent->bindlist.end()) SetActiveBind(*(mapper.abindit));
-			else SetActiveBind(nullptr);
+			if (mapper.abindit != mapper.aevent->bindlist.end()) {
+				SetActiveBind(*(mapper.abindit));
+			} else {
+				SetActiveBind(nullptr);
+			}
+			MAPPER_SaveBinds();
 			break;
 		case BB_Next:
 			if (mapper.abindit != mapper.aevent->bindlist.end())
@@ -1637,9 +1640,6 @@ public:
 			if (mapper.abindit == mapper.aevent->bindlist.end())
 				mapper.abindit = mapper.aevent->bindlist.begin();
 			SetActiveBind(*(mapper.abindit));
-			break;
-		case BB_Save:
-			MAPPER_SaveBinds();
 			break;
 		case BB_Exit:
 			mapper.exit=true;
@@ -1693,15 +1693,19 @@ public:
 		switch (type) {
 		case BC_Mod1:
 			mapper.abind->mods^=BMOD_Mod1;
+			MAPPER_SaveBinds();
 			break;
 		case BC_Mod2:
 			mapper.abind->mods^=BMOD_Mod2;
+			MAPPER_SaveBinds();
 			break;
 		case BC_Mod3:
 			mapper.abind->mods^=BMOD_Mod3;
+			MAPPER_SaveBinds();
 			break;
 		case BC_Hold:
 			mapper.abind->flags^=BFLG_Hold;
+			MAPPER_SaveBinds();
 			break;
 		}
 		mapper.redraw=true;
@@ -2469,7 +2473,6 @@ static void CreateLayout() {
 	bind_but.del = new CBindButton(250, 400, 100, 20, "Remove bind", BB_Del);
 	bind_but.next = new CBindButton(250, 420, 100, 20, "Next bind", BB_Next);
 
-	bind_but.save=new CBindButton(400,450,50,20,"Save",BB_Save);
 	bind_but.exit=new CBindButton(450,450,50,20,"Exit",BB_Exit);
 
 	bind_but.bind_title->Change("Bind Title");
@@ -2734,8 +2737,11 @@ static void MAPPER_SaveBinds() {
 		fprintf(savefile,"\n");
 	}
 	fclose(savefile);
-	change_action_text("Mapper file saved.", color_white);
-	LOG_MSG("MAPPER: Wrote key bindings to %s", filename);
+	change_action_text("Mapping updated.", color_white);
+	if (!suppress_save_mapper_file_message) {
+		suppress_save_mapper_file_message = true;
+		LOG_INFO("MAPPER: Wrote input bindings to %s", filename);
+	}
 }
 
 static bool load_binds_from_file(const std::string_view mapperfile_path,
@@ -2751,6 +2757,8 @@ static bool load_binds_from_file(const std::string_view mapperfile_path,
 		auto lines = get_resource_lines(mapper_path, optional);
 		if (lines.empty())
 			return false;
+
+		suppress_save_mapper_file_message = true;
 
 		ClearAllBinds();
 		for (auto &line : lines)
@@ -2869,6 +2877,7 @@ void BIND_MappingEvents() {
 				mapper.aevent->AddBind(newbind);
 				SetActiveEvent(mapper.aevent);
 				mapper.addbind=false;
+				MAPPER_SaveBinds();
 				break;
 			}
 		}


johnnovak:
Great stuff @arifogel 🎉 The changes look very straightforward. We'll test it on Windows and macOS as well, then if all seems well you can rebase your other PR on top of this, as you mentioned 👍🏻

arifogel:
Oh I was also going to test it on Windows and Mac in a few minutes, but also useful to have someone else try it.

arifogel:
I completed abbreviated testing on macOS and Windows, described in edits to top comment.

johnnovak:
> I completed abbreviated testing on macOS and Windows, described in edits to top comment.

Great, thanks for that 👍🏻

johnnovak:
> To avoid saving to disk (or worse, network share) after each binding - is it possible instead to combine Save and Exit functions, e.g. Exit button actually does "Save, then Exit"?

No, and saving to disk is fine. Storing the mapping files on a network share is a pretty silly thing to do. We don't deal with silly 😎

src/gui/sdl_mapper.cpp
```diff
@@ -2101,7 +2096,12 @@ static void SetActiveEvent(CEvent * event) {
 			SetActiveBind(*(mapper.abindit));
 
 			const auto new_bind = dynamic_cast<CKeyBind*>(*mapper.abindit);
-			drop_other_bound_events(new_bind);
+			// TODO: version of drop_other_bound_events that works
+			// for joystick binds, for which above line assigns
+			// nullptr to newbind and causes assertion failure
+			if (new_bind) {
+				drop_other_bound_events(new_bind);
+			}
```
weirddan455, line 2104: I confirmed this fixes #4126 but I'm a little confused on what this `TODO` means.  This seems to work fine as-is with joystick binds.

I would also suggest adding "Regression from f33cc25e57b7d1b8835f1efae7e43efa4ea7691e" to the commit message for future reference since you went through the trouble of bisecting it.  I also confirmed this was the commit that introduced the crash.

src/gui/sdl_mapper.cpp
```diff
@@ -1646,6 +1643,7 @@ class CBindButton final : public CClickableTextButton {
 			}
 			if (mapper.abindit!=mapper.aevent->bindlist.end()) SetActiveBind(*(mapper.abindit));
 			else SetActiveBind(nullptr);
```
weirddan455, line 1645: Suggest fixing up the formatting on these 2 lines.  This is our formatting style for if/else statements (for new code, we still have legacy code that doesn't follow our style)  I know you didn't write this but you added a line after this single-line else so it would be nice to clean this up for readability.

```cpp
if (mapper.abindit != mapper.aevent->bindlist.end()) {
    SetActiveBind(*(mapper.abindit));
} else {
    SetActiveBind(nullptr);
}

**Review by weirddan455 — COMMENTED** [review]
Nice job with this.  I tested and it seems to work fine on my system.  Just a couple of minor comments.

**Comment by weirddan455** [comment]
> > To avoid saving to disk (or worse, network share) after each binding - is it possible instead to combine Save and Exit functions, e.g. Exit button actually does "Save, then Exit"?
> 
> Not worth it, and saving to disk is fine. Storing the mapping files on a network share is a pretty silly thing to do. We don't deal with silly 😎

It should be fine even with network shares.  The mapper file is only ~3KB for me and we're still only saving when we add a binding.  Shouldn't much delay as long as you're network is faster than 56kbps dial-up (if so it might take a second or so :laughing: )

File: src/gui/sdl_mapper.cpp (hunk 1 above)
Comment (arifogel) [reply src/gui/sdl_mapper.cpp]: I updated the first commit to just implement the TODO bits. The problem was that only bindings of host keys were being deduplicated, not bindings of host joystick axes / joystick buttons / joystick hats. The crash was due to a dynamic cast to `CKeyBind*` of something that was a pointer to a different subclass of `CBind` (which returns `nullptr`), then passing that result to the `drop_other_bound_events` function which immediately asserts its argument is not nullptr.

With the updated commit, dedupe is now implemented for all subtypes of CBind, and the downcast is removed.

File: src/gui/sdl_mapper.cpp (hunk 2 above)
Comment (johnnovak) [reply src/gui/sdl_mapper.cpp:1682]: Another formatting tip @arifogel, we use Clang Format, there is a `.clang-format` config at the root of the repo. All popular editors and IDEs can be configured or have plugin supportfor Clang Format. I personally use [vim-clang-format](https://github.com/rhysd/vim-clang-format).

Comment (arifogel) [reply src/gui/sdl_mapper.cpp]: done, just rebased again

**Comment by johnnovak** [comment]
> > > To avoid saving to disk (or worse, network share) after each binding - is it possible instead to combine Save and Exit functions, e.g. Exit button actually does "Save, then Exit"?
> > 
> > 
> > Not worth it, and saving to disk is fine. Storing the mapping files on a network share is a pretty silly thing to do. We don't deal with silly 😎
> 
> It should be fine even with network shares. The mapper file is only ~3KB for me and we're still only saving when we add a binding. Shouldn't have much delay as long as your network is faster than 56kbps dial-up (if so it might take a second or so 😆 )

Yeah 😆 What irked me about this comment is that people should know not to put the data files of performance intensive / realtime apps on shared drives or slow media like USB sticks. We don't want to "optimise" for silly situations like that. Staging could be hampered in a lot of ways if you start spreading its files around on random slow drives... but let's not add specific code paths for crazy 😏

File: src/gui/sdl_mapper.cpp (hunk 2 above)
Comment (arifogel) [reply src/gui/sdl_mapper.cpp]: @johnnovak I am running `./scripts/format-commit.sh` on each commit, but AFAICT it only touches changed lines. I can set my IDE to clang-format automatically, but I would still only want it to hit changed lines so my PRs are legible. So it would not have been caught because:

> I know you didn't write this but you added a line after this single-line else so it would be nice to clean this up for readability.

Comment (weirddan455) [reply src/gui/sdl_mapper.cpp]: That script will usually look a couple lines above and below changed lines and format them as well but it's not perfect.  Sometimes it will even completely butcher formatting with certain patterns so it's worth re-checking manually.  Most of the time it works alright though.

File: src/gui/sdl_mapper.cpp (hunk 1 above)
Comment (arifogel) [reply src/gui/sdl_mapper.cpp]: I updated the linux testing procedure to include how I tested the implementation of the TODO bits.

File: src/gui/sdl_mapper.cpp (hunk 2 above)
Comment (johnnovak) [reply src/gui/sdl_mapper.cpp]: Makes sense @arifogel. Yeah, sometimes it needs a little manual help. It's handy to have it configured for your IDE too so you can reformat only the selected lines. I use that all the time.

File: src/gui/sdl_mapper.cpp [hunk 3]
```diff
@@ -2052,27 +2091,25 @@ static void drop_other_bound_events(const T* new_bind)
...
+		if constexpr (std::is_same_v<T, CBind>) {
 			if (!other_bind || new_bind->event == other_bind->event) {
 				return false;
 			}
 		}
 
 		if (new_bind->HasSameBinding(other_bind)) {
 			const auto msg = format_str(
-			        "Host button '%s' now performs '%s' instead of '%s'",
+			        "Host input '%s' now performs '%s' instead of '%s'",
```
johnnovak, line 2101: 👍🏻

johnnovak:
I've tested it and I've noticed a couple of problems:

- Modifying the mod keys and the "hold" flag for a binding does not trigger a save -- this is an omission and must be fixed.
- Logging saving the config every time we bind/unbind something is way too chatty. So I propose to do this:
  - If Staging loaded a mapper file at startup, never log saving the mapper file. Something like this appears in the logs at startup if a mapper file was loaded: `MAPPER: Loaded 158 key bindings from 'D:\Emulators\DOS\DOSBox-Staging\mapper-sdl2-0.83.0-alpha.map'`
  - If no mapper file was loaded at startup, this appears in the logs: `MAPPER: Loaded default key bindings` (e.g., when you start Staging in portable mode for the first time). In this case it would be nice to log the creation of the new mapper file _once_ on the first save, but not after save logging should happen after that (a simple static boolean flag would do the job).

I've noticed other deal breaking bugs/regressions too introduced by https://github.com/dosbox-staging/dosbox-staging/commit/f33cc25e57b7d1b8835f1efae7e43efa4ea7691e, so let's not build stuff on top of it until those regression are resolved. 

I will raise a separate issue ticket for it shortly.

src/gui/sdl_mapper.cpp
```diff
@@ -404,14 +412,13 @@ class CKeyBind final : public CBind {
...
-
-	const char* GetEventName() const
+	bool HasSameBinding(const CBind* other) const override
 	{
-		return event ? event->GetName() : "";
+		const auto other_typed = dynamic_cast<const CKeyBind*>(other);
+		if (!other_typed) {
+			return false;
+		}
+		return other_typed && key == other_typed->key && mods == other->mods;
```
weirddan455, line 418: You're checking whether `other_typed` is null twice.  You could remove the if statement and it'll work exactly the same.  Alternatively, you could say you're leaving the separate if statement for readability and remove the `other_typed &&` from the return statement.

src/gui/sdl_mapper.cpp
```diff
@@ -526,6 +533,16 @@ class CJAxisBind final : public CBind {
...
 	}
 
+	bool HasSameBinding(const CBind* other) const override
+	{
+		const auto other_typed = dynamic_cast<const CJAxisBind*>(other);
+		if (!other_typed) {
+			return false;
+		}
+		return other_typed && axis == other_typed->axis &&
+		       positive == other_typed->positive && mods == other->mods;
```
weirddan455, line 540: Same comment here regarding `other_typed` null check.

src/gui/sdl_mapper.cpp
```diff
@@ -555,6 +572,16 @@ class CJButtonBind final : public CBind {
...
 	}
 
+	bool HasSameBinding(const CBind* other) const override
+	{
+		const auto other_typed = dynamic_cast<const CJButtonBind*>(other);
+		if (!other_typed) {
+			return false;
+		}
+		return other_typed && button == other_typed->button &&
+		       mods == other->mods;
```
weirddan455, line 579: And again

src/gui/sdl_mapper.cpp
```diff
@@ -602,6 +629,16 @@ class CJHatBind final : public CBind {
...
 	}
 
+	bool HasSameBinding(const CBind* other) const override
+	{
+		const auto other_typed = dynamic_cast<const CJHatBind*>(other);
+		if (!other_typed) {
+			return false;
+		}
+		return other_typed && hat == other_typed->hat &&
+		       dir == other_typed->dir && mods == other->mods;
```
weirddan455, line 636: And one more time

src/gui/sdl_mapper.cpp
```diff
@@ -1873,9 +1910,11 @@ class CHandlerEvent final : public CTriggeredEvent {
 
 	void Active(bool yesno) override { (*handler)(yesno); }
 
-	bool HasSameBinding(const CKeyBind* other) const
+	bool HasSameBinding(const CBind* other) const
 	{
-		return other && defkey == other->key && defmod == other->mods;
+		auto other_keybind = dynamic_cast<const CKeyBind*>(other);
+		return other_keybind && defkey == other_keybind->key &&
+		       defmod == other->mods;
```
weirddan455, line 1915: Except here you don't have the separate if statement.  Definitely prefer consistency one way or the other.

src/gui/sdl_mapper.cpp
```diff
@@ -2099,10 +2136,7 @@ static void SetActiveEvent(CEvent * event) {
 		mapper.abindit=event->bindlist.begin();
 		if (mapper.abindit!=event->bindlist.end()) {
 			SetActiveBind(*(mapper.abindit));
-
-			const auto new_bind = dynamic_cast<CKeyBind*>(*mapper.abindit);
-			drop_other_bound_events(new_bind);
-
+			drop_other_bound_events(*mapper.abindit);
 		} else SetActiveBind(nullptr);
```
weirddan455, line 2138: Reformat this else branch like I said before,

src/gui/sdl_mapper.cpp
```diff
@@ -2099,10 +2136,7 @@ static void SetActiveEvent(CEvent * event) {
 		mapper.abindit=event->bindlist.begin();
 		if (mapper.abindit!=event->bindlist.end()) {
 			SetActiveBind(*(mapper.abindit));
-
-			const auto new_bind = dynamic_cast<CKeyBind*>(*mapper.abindit);
-			drop_other_bound_events(new_bind);
-
+			drop_other_bound_events(*mapper.abindit);
```
weirddan455, line 2137: Minor but I really consistency:  Remove the parenthesis in that top line.  Member access happens before the de-reference so it's not needed.  Should just be `SetActiveBind(*mapper.adindit);`

I only just now noticed it because you added that last line that didn't have the extra parenthesis.

weirddan455 (commented):
What does all this do that your first patch did not?  I easily understood the first patch:  It was passing nullptr into `drop_other_bound_events()` when the event was from a joystick and crashing.  You fixed it by adding a null check.

I'm not sure what all this is doing and I'm just now reading the `drop_other_bound_events()` function which is templated and also hard to understand what exactly it does :laughing: 

If this has some tangible benefit, then sure we can go this route.  Otherwise, I'd prefer to stick with the easy 3 line patch you had before.

Not sure if you're aware but this entire `sdl_mapper.cpp` is the worst code I've seen in the entire project.  It's already been scheduled to be killed with fire/nuked from orbit and re-built from scratch whenever @johnnovak has the time (or someone else wants to step up but it's a large job).  There's not too much point in trying to refactor this code and as you can see last time we touched it, we broke it which is easy to do with this fragile code :laughing:

The crash on "Add Bind" regression needs to be fixed.  I just prefer to do these things with minimal changes, especially when dealing with this file.  There's other places in the codebase where refactoring makes more sense and would be welcomed but this mapper code is very flawed in a lot of ways.

weirddan455:
Sorry @arifogel I just now saw your earlier comment:

>  The problem was that only bindings of host keys were being deduplicated, not bindings of host joystick axes / joystick buttons / joystick hats. 

So this new version of your patch stops the joystick axes from being mapped to 2 different keyboard keys.  I tested that and it works.  That makes it worthwhile.  You can mostly ignore my top comment then.  I just never miss an opportunity to complain about the mapper when I get a change I guess :laughing:

johnnovak:
> Not sure if you're aware but this entire `sdl_mapper.cpp` is the worst code I've seen in the entire project. It's already been scheduled to be killed with fire/nuked from orbit and re-built from scratch whenever @johnnovak has the time (or someone else wants to step up but it's a large job). There's not too much point in trying to refactor this code and as you can see last time we touched it, we broke it which is easy to do with this fragile code 😆

Indeed, it's terrible and unsalvagable. We've made a decision that we'll only make surgical regression fixes to it and _very minimal_ enhancements if they're just a few lines of easy to understand code.

Look, I'd hate to shoot this down @arifogel as it's your first contribution, I'm just worried this is going to spiral out of control. Like the hotredeploy losing the mappings, which necessitates saving after each mapping change, which necessitates further refinements because now it's buggy and a bit rough around the edges (the logging in particular). We're in "here be dragons territory" 🐉 🔥 💥 😆 

What worries me most is this earlier regression that I found during testing this:

- https://github.com/dosbox-staging/dosbox-staging/issues/4128

I _really_ don't want to dick around with the mapper, so if someone will fix the regressions introduced by @interloper98 's PR (https://github.com/dosbox-staging/dosbox-staging/pull/4079), that's fine and welcome, but if not, I'm inclined to revert it.

Then all this stuff seems to be built on that shaky foundation (at least partially), so again, I don't wanna dragged into dealing with the mapper. Happy to test it sporadically, but let us keep our perspective: surgical fixes and minor enhancements are fine, but for not-really-critical enhancements we need to tread with caution and press the big red "Abort" button if things start spiraling out of control.

I'd hate to curb your enthusiasm @arifogel, and you're in a bit of a deep end here, toching the mapper code in your first PR 😄 Just wanted to make you aware of the difficulties and my perspective, and at the end of the day neither hotpluggin nor @interloper98 's enhancements are essential.

weirddan455:
Yeah, I agree 100% with @johnnovak 

>  I've noticed other deal breaking bugs/regressions too introduced by https://github.com/dosbox-staging/dosbox-staging/commit/f33cc25e57b7d1b8835f1efae7e43efa4ea7691e, so let's not build stuff on top of it until those regression are resolved.

This needs to be fixed first, yeah.  The good news is the first commit in this PR solves part of the problem.  Up to you if you want to try to fix the rest of those bugs.

@interloper98 just commented on #4128 saying please revert it.  I was leaning that way myself.

@arifogel I _really_ don't want to scare off new contributors.  We're already kind of short of developers and can always use the help.  That's why I was warning you about the mapper.  It's probably the single worst place I'd want a new contributor to start.  If I thought the entire codebase was written like _that_ I would have run for the hills and never contributed anything.

I had actually looked into adding hotplug support last year sometime.  I just went back and read #946 and remembered I left a warning there as well :laughing: 

I'm not going to stand in your way if you want to do this but yeah it's tricky.

johnnovak:
> @arifogel I _really_ don't want to scare off new contributors. We're already kind of short of developers and can always use the help. That's why I was warning you about the mapper. It's probably the single worst place I'd want a new contributor to start. If I thought the entire codebase was written like _that_ I would have run for the hills and never contributed anything.

Same here; I'm _really_ grateful for contributions and want to encourage people to work on Staging. Just you started digging into one of the dirtiest, nastiest areas of the codebase... 😄Think of it as us extending a helping hand, not letting you sink into the quicksand (and at the same saving ourselves as well; it's no fun if we all go down together 😆).

arifogel:
From my perspective, I've implemented an enhancement that I am able to use locally, so that's good. I realize though that as a contribution, this is nontrivial stuff that will 100% be thrown away eventually even if it is accepted now. On the other hand, as a software engineer I am used to large planned migrations simply never occurring, so I would still prefer that these PRs be accepted to provide a UX improvement in the interim.

Given the above, here is what I would suggest :

1. Revert the breaking change (sounds like there is already consensus this is going to happen). Unmapping previous mappings of the same input is not necessarily even a desirable change, as it prevents some use cases. I will drop the forward patch (first commit in this PR) for uniquifying joystick input mappings.
2. I can fix up the primary commit of this PR per the review comments and issues unearthed in testing. Without the first commit, this should still be a relatively small PR. It's not exactly surgical, but it's still IMO a low-cost straightforward enhancement to UX.
3. I just a few minutes ago pushed a fix to the hotplugging PR that removes the crashes. It's not trivial, but I am fairly confident it's ready to go once this PR is merged.

johnnovak:
> From my perspective, I've implemented an enhancement that I am able to use locally, so that's good. I realize though that as a contribution, this is nontrivial stuff that will 100% be thrown away eventually even if it is accepted now. On the other hand, as a software engineer I am used to large planned migrations simply never occurring, so I would still prefer that these PRs be accepted to provide a UX improvement in the interim.
> 
> Given the above, here is what I would suggest :
> 
>     1. Revert the breaking change (sounds like there is already consensus this is going to happen). Unmapping previous mappings of the same input is not necessarily even a desirable change, as it prevents some use cases. I will drop the forward patch (first commit in this PR) for uniquifying joystick input mappings.
> 
>     2. I can fix up the primary commit of this PR per the review comments and issues unearthed in testing. Without the first commit, this should still be a relatively small PR. It's not exactly surgical, but it's still IMO a low-cost straightforward enhancement to UX.
> 
>     3. I just a few minutes ago pushed a fix to the hotplugging PR that removes the crashes. It's not trivial, but I am fairly confident it's ready to go once this PR is merged.

Sounds like a plan. Yeah with the regression and related issues out of the mix, hopefully the hotpluggin alone will be a simple and clean change.

arifogel:
> * Modifying the mod keys and the "hold" flag for a binding does not trigger a save -- this is an omission and must be fixed.
> * Logging saving the config every time we bind/unbind something is way too chatty. So I propose to do this:
>   
>   * If Staging loaded a mapper file at startup, never log saving the mapper file. Something like this appears in the logs at startup if a mapper file was loaded: `MAPPER: Loaded 158 key bindings from 'D:\Emulators\DOS\DOSBox-Staging\mapper-sdl2-0.83.0-alpha.map'`
>   * If no mapper file was loaded at startup, this appears in the logs: `MAPPER: Loaded default key bindings` (e.g., when you start Staging in portable mode for the first time). In this case it would be nice to log the creation of the new mapper file _once_ on the first save, but not after save logging should happen after that (a simple static boolean flag would do the job).

Fixed in e52bca7f047f7ce19bbf044f1d06e3a22d931eb7

arifogel:
Better diff: https://github.com/arifogel/dosbox-staging/compare/35f12fdfc3bc926105941e6e7bc2a57f4a9b27ff..e52bca7f047f7ce19bbf044f1d06e3a22d931eb7

weirddan455:
Sounds like a good plan to me @arifogel 

This PR looks good to me now.  I resolved all my comments about the first commit that was removed.  I think we should wait until @johnnovak merges the revert in #4129 and then this PR is good to go from my perspective.

arifogel:
> I'm not sure, but I think the merge of the two buttons into "(Save, then) Exit" would've solved both of those.

This would not have been acceptable in the case the user sets some binds, then turns on a joystick (with hot swapping) before leaving the mapper UI. In that case, the binds set prior to the joystick being connected would be lost.

arifogel:
Rebased after #4129 and retested on Linux.

weirddan455 (approved):
I hit approve for CI.  Since you're a first time contributor, it doesn't start automatically without approval.  I think this is good to merge unless @johnnovak has any concerns.

johnnovak:
> I hit approve for CI. Since you're a first time contributor, it doesn't start automatically without approval. I think this is good to merge unless @johnnovak has any concerns.

I'm gonna do my last round of testing shortly.

johnnovak (approved):
Just tested it and works very well now! 🎉 Thanks for bearing with us and addressing all our comments @arifogel. A very nice first time contribution, congrats! 🥇

Issue #4118, opened 2024-12-21
Volume label doesn't support any special characters such as underscore


distantvale:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When mounting and using the -label option with a specific volume label, special characters are removed before the label is set.
This includes characters such as underscore which should be supported.
Also see this -
https://jeffpar.github.io/kbarchive/kb/026/Q26295/

### Steps to reproduce the behaviour.

Explain how to reproduce
1. mount any drive with "-label A_B" option
2. navigate to the drive and run the "dir" command
3. see that the volume label is actually set as "AB" (underscore omitted)

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
All yours @NicknineTheEagle

NicknineTheEagle:
https://github.com/dosbox-staging/dosbox-staging/blob/e5c8d3a24e2834fb2898dde1517e982d025c7934/src/dos/drives.cpp#L28-L45

Well, there's the problem, To_Label function used by DIR command calls strip_punctuation which removes all characters that std::ispunct returns true for. This issue is actually acknowledged in the comment and it also points out that this issue is kind of a mess.

johnnovak:
> https://github.com/dosbox-staging/dosbox-staging/blob/e5c8d3a24e2834fb2898dde1517e982d025c7934/src/dos/drives.cpp#L28-L45
> 
> Well, there's the problem, To_Label function used by DIR command calls strip_punctuation which removes all characters that std::ispunct returns true for. This issue is actually acknowledged in the comment and it also points out that label formatting is kind of a mess.

Cool, please PR the change then (you said you wanted to work on this).

johnnovak:
@NicknineTheEagle @distantvale Can you guys extend the top description with an actual concrete where this caused a problem? You couldn't mount a game installer CD or something? Which game?

distantvale:
@johnnovak I don't recall a specific test for CD but I did see old original CDs with underscore in their label and I just added a link to one of them in the top description.

Dustman0192:
[...]

PR #4134, opened 2024-12-30
Fix special character filtering in volume labels

NicknineTheEagle:
# Description

This adjusts the code that fixes up volume label for functions like DIR command so it correctly filters out characters not allowed for volume labels on DOS. Previously, it was too strict and filtered out characters like underscores which are perfectly legal.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4118.

# Release notes

Fixed some characters being omitted from volume labels.

# Manual testing

Mounted a directory as a drive with `-label A_B` option, checked the label with DIR command, the label shows up correctly. Mounted Quake and Need for Speed: Special Edition disc images, no regressions noticed.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/drives.cpp
+++ b/src/dos/drives.cpp
@@ -25,12 +25,16 @@
 
 extern char sfn[DOS_NAMELENGTH_ASCII];
 
+constexpr bool is_special_character(const char c)
+{
+	constexpr auto special_characters = std::string_view("\"+=,;:<>[]|?*");
+	return special_characters.find(c) != std::string_view::npos;
+}
+
 // TODO Right now label formatting seems to be a bit of mess, with various
 // places in code setting/expecting different format, so simple GetLabel() on
 // a drive object might not yield an expected result. Not sure how to sort it
 // out, but it will require some attention to detail.
-// Also: this function is too strict - it removes all punctuation when *some*
-// punctuation is acceptable in drive labels (e.g. '_' or '-').
 //
 std::string To_Label(const char* name) {
 	// Reformat the name per the DOS label specification:
@@ -38,7 +42,15 @@ std::string To_Label(const char* name) {
 	// - Internal spaces allowed but no: tabs ? / \ | . , ; : + = [ ] < > " '
 	std::string label(name);
 	trim(label); // strip front-and-back white-space
-	strip_punctuation(label); // strip all punctuation
+	label.erase(std::remove_if(label.begin(),
+	                           label.end(),
+	                           [](unsigned char c) {
+		                           return c == '\t' || c == '/' ||
+		                                  c == '\\' || c == '.' ||
+		                                  c == '\'' ||
+		                                  is_special_character(c);
+	                           }),
+	            label.end());
 	label.resize(11); // collapse remainder to (at-most) 11 chars
 	upcase(label);
 	return label;
@@ -73,12 +85,6 @@ void Set_Label(const char* const input, char* const output, bool cdrom)
 	//Remove trailing dot. except when on cdrom and filename is exactly 8 (9 including the dot) letters. MSCDEX feature/bug (fifa96 cdrom detection)
 	if((labelPos > 0) && (output[labelPos-1] == '.') && !(cdrom && labelPos ==9))
 		output[labelPos-1] = 0;
-}
-
-constexpr bool is_special_character(const char c)
-{
-    constexpr auto special_characters = std::string_view("\"+=,;:<>[]|?*");
-    return special_characters.find(c) != std::string_view::npos;
 }
 
 /* Generate 8.3 names from LFNs, with tilde usage (from ~1 to ~9999). */


vs/dosbox.vcxproj
```diff
@@ -276,7 +276,13 @@ IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0</Command>
     </ResourceCompile>
     <PostBuildEvent>
       <Command>%systemroot%\System32\robocopy ..\contrib\resources "$(OutDir)resources" /mir /xf "meson.build" "*.sh"
-IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0</Command>
+IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0
+rd /s /q D:\Emulators\Dosbox\resources
```
FeralChild64, line 280: Are you sure the changes to this file should be a part of the commit?

NicknineTheEagle: Oh, for god's sake, this just keeps happening. I should really come up with a better solution for copying the compiled binaries.

johnnovak:
Can you tell us what the volume label of **Quake and Need for Speed: Special Edition** is?

src/dos/drives.cpp
```diff
@@ -25,20 +25,32 @@
 
 extern char sfn[DOS_NAMELENGTH_ASCII];
 
+constexpr bool is_special_character(const char c)
+{
+	constexpr auto special_characters = std::string_view("\"+=,;:<>[]|?*");
```
johnnovak, line 30: minor: `constexpr` names are always UpperCamelCase.

src/dos/drives.cpp
```diff
@@ -25,20 +25,32 @@
...
-	strip_punctuation(label); // strip all punctuation
+	label.erase(std::remove_if(label.begin(),
+	                           label.end(),
+	                           [](unsigned char c) {
+		                           return c == '\t' || c == '/' ||
+		                                  c == '\\' || c == '.' ||
+		                                  c == '\'' ||
+		                                  is_special_character(c);
+	                           }),
+	            label.end());
```
johnnovak, line 53: According to this, period characters should be allowed, no?

https://ftp.zx.net.nz/pub/archive/ftp.microsoft.com/MISC/KB/en-us/26/295.HTM

Also, this comment seems to be slightly wrong:

```
Internal spaces allowed but no: tabs ? / \ | . , ; : + = [ ] < > " '
```

The single-quote char is allowed, so it should be removed from the end.

I'd suggest putting this link here for reference:
https://ftp.zx.net.nz/pub/archive/ftp.microsoft.com/MISC/KB/en-us/26/295.HTM

NicknineTheEagle: > According to this, period characters should be allowed, no?

Then the comment is wrong about the period, too.

NicknineTheEagle: Something important that this article mentions is that control characters are not allowed. I'll need to add that.

johnnovak:
> > Can you tell us what the volume label of **Quake and Need for Speed: Special Edition** is?
> 
> QUAKE101 and NFSSE.

Okay... So you're saying you haven't tested this with a CD image that actually has the special chars in its volume name? Can you please do so then and post the results?

src/dos/drives.cpp
johnnovak: Cool, let's make sure everything is in sync (both the code and comments) according to that Microsoft knowledge base article. For extra points you cross-check with real MS-DOS 6.22 running in a VM.

NicknineTheEagle:
> > > Can you tell us what the volume label of **Quake and Need for Speed: Special Edition** is?
> > 
> > 
> > QUAKE101 and NFSSE.
> 
> Okay... So you're saying you haven't tested this with a CD image that actually has the special chars in its volume name? Can you please do so then and post the results?

I've tested those two to check for possible regressions. I've used reproduction method from the original issue to test underscores. But if it's important, I can get that game @distantvale mentioned in Discord chat.

src/dos/drives.cpp
```diff
@@ -25,20 +25,32 @@
...
-// punctuation is acceptable in drive labels (e.g. '_' or '-').
 //
 std::string To_Label(const char* name) {
 	// Reformat the name per the DOS label specification:
 	// - Upper-case, up to 11 ASCII characters
 	// - Internal spaces allowed but no: tabs ? / \ | . , ; : + = [ ] < > " '
 	std::string label(name);
 	trim(label); // strip front-and-back white-space
-	strip_punctuation(label); // strip all punctuation
+	label.erase(std::remove_if(label.begin(),
```
weirddan455, line 45: Do you need this outer `label.erase()`?  Doesn't `std::remove_if()` handle everything?  From the docs here:  https://en.cppreference.com/w/cpp/algorithm/remove

What I'm reading is at best the `label.erase()` is a no-op because `std::remove_if()` returns an end iterator.  At worst, it might be undefined behavior because the compiler could evaluate `label.end()` before `std::remove_if()` and then the end iterator becomes invalidated.

johnnovak:
@Torinde Thanks for testing, the real DOS behaviour is a royal mess. 😱 

We certainly don't want to implement all that nonsense. Total bug-compatibility with MS-DOS is not really our goal 😅, only where it really matters (rare). If you could just paste here the actual list of special chars accepted by the LABEL command, we'll use those and move on.

If it's proven later any of this craziness matters, we might add gradual support on a per-game basis, but now a combination of special chars said to be allowed by the knowledge base article plus the few extras you discovered will suffice.

In reality, we did all this to allow the underscores 😆

johnnovak:
> I've used reproduction method from the original issue to test underscores. But if it's important, I can get that game @distantvale mentioned in Discord chat.

@NicknineTheEagle 

That's fair. I think it's important to test the real-world scenario why we're doing this: mounting an actual ISO that has underscores in the label. So yes please, if you could test with that ISO it would be great, and let's document the title of the game here.

Then we can mention this game in the game compatibility release notes section.

NicknineTheEagle:
> I tested MS-DOS 6.22 in Virtualbox. The highlights:
> 
>     * high ASCII - some get replaced by regular latin letters, others get capitalized
> 
>     * URL lists twice `/` and `:` and doesn't list some characters, which testing shows as invalid
> 
>     * it seems that if you run the test too many times eventually you get "Cannot execute LABEL.EXE"
> 
>     * most invalid/ignored characters are actually shown by dir/label if an IMG file has them (e.g. modified by Hex editor or some DOS tool different from LABEL). LABEL can't modify anymore the label if some (most?) of those characters are already present.
> 
> 
> Test results below for:
> 
>     * `LABEL A: < 65CR` (where 65CR is a 2 byte file containing character 65 and a CR)
> 
>     * `LABEL A: < A65ACR` (where A65ACR is a 4 byte file containing A, character 65, A and a CR) - some symbols are not acceptable as single character, but are acceptable in the middle. I haven't tested if they are acceptable as first character in a longer string.
> 
>     * hexedited IMG
> 
> 
> image

Ok so illegal characters are in fact shown by DIR command, they're just not accepted by LABEL command. That changes things.

johnnovak:
> > I tested MS-DOS 6.22 in Virtualbox. The highlights:
> > ```
> > * high ASCII - some get replaced by regular latin letters, others get capitalized
> > 
> > * URL lists twice `/` and `:` and doesn't list some characters, which testing shows as invalid
> > 
> > * it seems that if you run the test too many times eventually you get "Cannot execute LABEL.EXE"
> > 
> > * most invalid/ignored characters are actually shown by dir/label if an IMG file has them (e.g. modified by Hex editor or some DOS tool different from LABEL). LABEL can't modify anymore the label if some (most?) of those characters are already present.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Test results below for:
> > ```
> > * `LABEL A: < 65CR` (where 65CR is a 2 byte file containing character 65 and a CR)
> > 
> > * `LABEL A: < A65ACR` (where A65ACR is a 4 byte file containing A, character 65, A and a CR) - some symbols are not acceptable as single character, but are acceptable in the middle. I haven't tested if they are acceptable as first character in a longer string.
> > 
> > * hexedited IMG
> > ```
> 
> Ok, so illegal characters are in fact shown by DIR command, they're just not accepted by LABEL command. That changes things.

Like I said a few comments above, I don't think we need to overcomplicate things this much. We need to emulate just enough of MS-DOS so we can run games and mount ISO images for installing stuff. We don't need to go overboard with 100% bug-compatibility and support insane edge cases like these.

FeralChild64:
@Torinde Keep it mind, that what the LABEL command accepts almost certainly depends on what COUNTRY.SYS knows about current code page (FUCASE, FCHAR, etc. - see http://www.delorie.com/djgpp/doc/rbinter/it/22/26.html for the format description), might depend whether NLSFUNC.EXE is loaded, etc.

We don't have all such things implemented yet, and they won't be implemented anytime soon - a proper implementation will take quite some time, it won't happen anytime soon. For now the best solution is to just be tolerant about volume labels and file names.

johnnovak:
> We don't have all such things implemented yet, and they won't be implemented anytime soon - a proper implementation will take quite some time, it won't happen anytime soon. For now the best solution is to just be tolerant about volume labels and file names.

And let me tell it for the 5th time: we just need enough to run games, the end 😄 

@Torinde this is too much, we don't need this level of detail... I don't think anyone will go through what you posted.

johnnovak:
@Torinde, I just asked a simple question: which characters are _not_ allowed in addition to the ones listed in the knowledge base article. As I see if, you haven't identified any.

Also, I don't have the faintest clue why you're testing with PC DOS 7.10, DR-DOS 7.03, FreeDOS and all that crap... We only care about MS-DOS.

---

Ok, @NicknineTheEagle, please just proceed as we originnaly planned. Just disallow the chars from the knowledge base article and the control chars below code 32. Regular ASCII lower case letters should be uppercased, and that's it.

We can always complicate this later if those weird edge cases matter, but I'm 99.9999% sure they won't. Let's not overdo things for no good reason...

johnnovak:
> > which characters are not allowed in addition to the ones listed in the knowledge base article.
> 
> That's listed in my 3rd bullet [above](https://github.com/dosbox-staging/dosbox-staging/pull/4134#issuecomment-2566932986) "symbols ... not allowed for entry", specifically `^&*()+=[]\|;:",<.>/?` - if you really need a split per KB:
> 
>  * listed in the KB:  `*+=[]\|:",<>/?`
> 
>  * not listed in the KB: `^&();.`

Cheers 

> Of course you can skip the rest of the nuances - actually **you can just allow ALL symbols** - if somebody on purpose uses a Hex editor to make a funky label it's for him to deal with incompatibilities, if any. And except 7-9 all are even at least visible in DIR.

We don't care about replicating MS-DOS misbehaviour and crashes.

> > why you're testing with PC DOS 7.10, DR-DOS 7.03, FreeDOS
> 
> [#2539 (comment)](https://github.com/dosbox-staging/dosbox-staging/issues/2539#issuecomment-1613515986)

Read it again more carefully. FeralChild talks about diverging from MS-DOS messages if we can provide better user-facing error messages and sych, and mentions other DOS versions he used had have slightly different _messages_. Then he says MS-DOS _compatibility_ is _all_ what we care about. Only that, nothing else.

Re-read our about page and project goals carefully every time you post something (_before_ it, actually). Also, no information overload, unnecessary details, and far-out ideas... We asked you to stop doing these before many times, and now you're reverting to these habits again.

johnnovak:
> Are there Github private messages somewhere? For back-and-forths like below
> 
> > > > why you're testing with PC DOS 7.10, DR-DOS 7.03, FreeDOS
> > > 
> > > 
> > > [#2539 (comment)](https://github.com/dosbox-staging/dosbox-staging/issues/2539#issuecomment-1613515986)
> > 
> > 
> > Read it again more carefully. FeralChild talks about diverging from MS-DOS messages
> 
> "As for command ... features ... I am trying to implement various extensions, from: ..."
> 
> Since the KB differs from reality I wanted to check if non-MS versions adhere to it or not. I just checked MS-DOS 7.10 and LABEL there rejects `^&();.` as well.
> 
> The KB has two symbols listed twice `/:/:`, which obviously is a mistake and probably they ought to write `/:;.`, which would make the LABEL forbidden symbols list the same as the [list for FAT16 filenames](https://aeb.win.tue.nl/linux/fs/fat/fat-1.html#ss1.4).
> 
> So the real extra forbidden symbols are `^&()` - misbehavior in a sense :)
> 
> > > Of course you can skip the rest of the nuances - actually **you can just allow ALL symbols** - if somebody on purpose uses a Hex editor to make a funky label it's for him to deal with incompatibilities, if any. And except 7-9 all are even at least visible in DIR.
> > 
> > 
> > We don't care about replicating MS-DOS misbehaviour and crashes.
> 
> More akin non-documented behavior - DIR shows almost all symbols (which is useful), LABEL rejects some control codes in a slightly awkward manner with only one case where you need to restart.
> 
> Isn't it normal to expect some nuances like what to do with DEL/127 and leading space or deleted file symbol?

It's not if it's just curiosity and it doesn't affect anything (running games and apps).

NicknineTheEagle:
> Are you going to finish this @NicknineTheEagle? We tend to close unfinished PRs after 2 weeks of inactivity.

Yeah, just didn't have a chance to work on this these past two weeks.

johnnovak:
> > Are you going to finish this @NicknineTheEagle? We tend to close unfinished PRs after 2 weeks of inactivity.
> 
> Yeah, just didn't have a chance to work on this these past two weeks.

3 more weeks have passed, please re-open this one or a new PR when you're ready to continue

Issue #4123, opened 2024-12-27
Abuse now won't load, with error message "unable to load cache file name c:\lcache00.tmp"


Gryzor1363:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.82

### What Operating System are you using?

Windows 7

### Relevant hardware info

FX-8350 processor, GTX 1070 graphics card, 16 Gb of RAM.

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Upon game launch, error appears as described in title.

Games still works using the same configuration as usual and confirmed as working on mainline Dosbox 0.74, and also Dosbox Staging 0.80 and 0.81. 

Also tried lowering CPU to 3000 Mips, RAM to 16, type to either 486 or Pentium, tweaked EMS/XMS/UMB, all point to the same error specifically in 0.82.

I have tried granting Dosbox full permissions and I/O and run it as administrator, set _**allow_write_protected_files = true**_ within the config, switched the game folder and its parent to read/write mode as well, as per recommendation on very old threads pertaining to a similar issue in the past with older versions of Dosbox, tried deleting the temporary file in question, to no avail. 

Please see output details from game in attachment. 

![image](https://github.com/user-attachments/assets/f0a4d254-38b0-4750-abea-0c9718b878cf)

### Steps to reproduce the behavior.

Explain how to reproduce
1. Launch game
2.Wait until loading hits a wall with the error message as described.

### Download URL of affected game or software

https://www.myabandonware.com/game/abuse-1ne

### Your configuration

```ini
https://gist.github.com/Gryzor1363/1c944cebcde5f077252f77972743589b
```

### Code of Conduct & Contributing Guidelines

- [x] Yes, I agree.

Gryzor1363:
Thanks for reacting @weirddan455 . I cannot reproduce on Staging 0.80, 0.81 or mainline 0.74-3 either, all pointing to the same exact game folder and file locations on my computer, and none of them even need running as admin, so it's definitely not user or permissions-related on OS-level.
But considering your 0.82 binary works fine, I'm starting to think something related to Win 7 that I'm using, might still be the culprit... I'll try with the same exe and conf on another computer ASAP and come back with results.

weirddan455:
The file creation failed error narrows the problem down.  It's failing here in the code:

[...]

Issue #3262, opened 2023-12-24
Crystal Caves, Episode 1: After Restoring a Saved Game, If You Attempt to Save Again, Game Displays `File Error` Message


MasterO2:
**Windows 10, 64-bit**
https://github.com/dosbox-staging/dosbox-staging/releases/download/v0.81.0-rc1/dosbox-staging-windows-msys2-x86_64-v0.81.0-rc1.zip

Crystal Caves: Episode 1, `CC1.EXE`, has a save game bug where if you restore a saved game after saving it in slot 0, and then try to save again it in slot 0, Crystal Caves displays an error message `FILE ERROR`.  It **does not** do that in Staging 0.80.1.  The error occurred on the 0.81 RC Windows, 64-bit zip file release.

The game folder **is not** read-only.  This problem also **did not** occur at all in Dosbox-Staging 0.80-1, Windows 10, 64-bit.  

Yes, the mounted `dosgames` folder path has a space in it which is surrounded by single quotation marks.

Could it possibly have something to do with https://github.com/dosbox-staging/dosbox-staging/issues/2965?

```
dosbox.exe -noprimaryconf
date       time         |
2023-12-24 17:10:58.892 | arguments: dosbox.exe -noprimaryconf
2023-12-24 17:10:58.892 | Current dir: D:\stuff\dian war\emus\dosbox-staging-windows-msys2-x86_64-v0.81.0-rc1
2023-12-24 17:10:58.892 | stderr verbosity: 0
2023-12-24 17:10:58.892 | -----------------------------------
2023-12-24 17:10:58.893 | dosbox-staging version 0.81.0-rc1
2023-12-24 17:10:58.893 | ---
2023-12-24 17:10:58.893 | LOG: Loguru version 2.1.0 initialised
2023-12-24 17:10:58.893 | CONFIG: Using portable configuration layout in 'D:\stuff\dian war\emus\dosbox-staging-windows-msys2-x86_64-v0.81.0-rc1'
2023-12-24 17:10:58.894 | LANG: Using internal English language messages
2023-12-24 17:10:58.910 | SDL: version 2.28.5 initialised (windows video and wasapi audio)
2023-12-24 17:10:58.910 | DISPLAY: Initialised 1067x800 windowed mode using bilinear scaling on display-0
2023-12-24 17:10:59.162 | OPENGL: Vendor: NVIDIA Corporation
[...]

PR #4142, opened 2025-01-04
Do not open or create files on Windows with FILE_SHARE_DELETE

weirddan455:
# Description
Fixes the game Abuse failure to launch

It attempts to delete a file while it is holding open a file handle to it
Removing this flag causes the delete to fail but it re-creates the file anyway and successfully launches

This may not be the best fix.  It causes different behavior than POSIX where the delete will always succeed.  I can do some more verification of real DOS behavior later.

I think this is what the old C library `fopen` calls were doing differently.  Again, I'd like to confirm behavior on older versions of Staging.

Setting this to Draft PR until I verify the above.  It's just late and I'm not doing this tonight.

## Related issues

#4123

# Release notes

Fixes Abuse error on launch on Windows 7 and on Windows 10+ network drives.

# Manual testing

Tested that Crystal Caves regression #3262 did not come back.  This game depends on the read and write share flags but not delete.

Tested that Abuse launches from a network drive on Windows 10

- [x] Windows

This is a Windows specific change.

# Checklist

I have:

### DIFF
--- a/src/misc/fs_utils_win32.cpp
+++ b/src/misc/fs_utils_win32.cpp
@@ -95,7 +95,7 @@ NativeFileHandle open_native_file(const std_fs::path& path, const bool write_acc
 
 	return CreateFileW(path.c_str(),
 	                   access,
-	                   FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+	                   FILE_SHARE_READ | FILE_SHARE_WRITE,
 	                   nullptr,
 	                   OPEN_EXISTING,
 	                   FILE_ATTRIBUTE_NORMAL,
@@ -111,7 +111,7 @@ NativeFileHandle create_native_file(const std_fs::path& path,
 
 	return CreateFileW(path.c_str(),
 	                   GENERIC_READ | GENERIC_WRITE,
-	                   FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+	                   FILE_SHARE_READ | FILE_SHARE_WRITE,
 	                   nullptr,
 	                   CREATE_ALWAYS,
 	                   win32_attributes,


src/misc/fs_utils_win32.cpp
```diff
@@ -95,7 +95,7 @@ NativeFileHandle open_native_file(const std_fs::path& path, const bool write_acc
 
 	return CreateFileW(path.c_str(),
 	                   access,
-	                   FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+	                   FILE_SHARE_READ | FILE_SHARE_WRITE,
```
johnnovak, line 98: Would be a perfectly acceptable and clean fix if it worked reliably. Great find @weirddan455, let's see how the testing goes 👍🏻

weirddan455:
Testing shows this is not how real DOS works.  This PR would have `DOS_UnlinkFile()` throw an error if there are open file handles (on Windows).

Real DOS (tested on MS-DOS 6.22 in a VM) seems to work more like Unix.  The delete succeeds even with open handles, open after delete fails (showing at least the file entry is deleted immediately) but reading from the old handle succeeds and returns the old content.

I'm not sure if the old data will remain accessible when we start creating and writing to new files.  DOS may end up reading from whatever happens to be on the disk where the old file used to be, which could be overwritten.  Anyway, that part *probably* isn't crucial to emulate.

My quick and dirty test program written in assembly if interested:  https://gist.github.com/weirddan455/5cae02fc3d6f932fe55025138ed2f47c

Good news is on POSIX systems, we seem to be pretty accurate.  I may need to think of a better Windows implementation though.

johnnovak:
> Real DOS (tested on MS-DOS 6.22 in a VM) seems to work more like Unix. The delete succeeds even with open handles, open after delete fails (showing at least the file entry is deleted immediately) but reading from the old handle after delete succeeds and returns the old content.

Just skimmed this, but this roughly corresponds with my theory on why you can still read the file:
https://www.datarecoverytools.co.uk/2009/12/07/file-deletion-in-fat16/

So the file pointer still points to those DATA region clusters for open file handles, so you can keep reading the file, no problem (until the deleted file marked as "deleted" is "garbage collected", or overwritten in other words).

It also makes sense why the open operation doesn't work after delete -- the file is marked as deleted, but that's just the directory entry.
 
> I'm not sure if the old data will remain accessible when we start creating and writing to new files. DOS may end up reading from whatever happens to be on the disk where the old file used to be, which could be overwritten. Anyway, that part _probably_ isn't crucial to emulate.

Definitely not crucial, but it follows from the above that it will remain readable until its data is overwritten. Then once it's overwritten, we're in undefined behaviour land and a crash is bound to happen... IMO it's perfectly sufficient to emulate the "file remains readable after delete forever" until we hold on the same file handle (and yeah, there might other weird bugs if the file handle itself it recycled by DOS... but again, we don't wanna go there).

> My quick and dirty test program written in assembly if interested: https://gist.github.com/weirddan455/5cae02fc3d6f932fe55025138ed2f47c

asm, that's the way 😎🤘🏻 

> Good news is on POSIX systems, we seem to be pretty accurate. I may need to think of a better Windows implementation though.

Yeah, would be nice, agreed.

weirddan455:
Thanks for the insight @johnnovak 

I'll close this PR and try for a new approach that more accurately emulates DOS behavior.  This PR can cause files that should be deleted to error out and not be deleted.  Likely to cause more regressions than it solves.

PR #4145, opened 2025-01-05
Revert Hotplug Support #4122,#4138,#4143

arifogel:
# Description

* Reverts commits introduced in #4122 - as well as its follow-on commits from #4138 and #4143- due to various regressions

## Related issues
* Revert of #4122, #4138, #4143
* Reopens #946

# Manual testing

Omitted manual testing because this is a full revert to a known state without any intervening conflicts

The change has been manually tested on:

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

- [X] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [X] performed a self-review of my code.
- [X] commented on the particularly hard-to-understand areas of my code.
- [X] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [ ] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [ ] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] my change has been manually tested on Windows, macOS, and Linux.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] provided the release notes draft (for significant user-facing changes).
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #4143, opened 2025-01-04
Remove reinterpret_cast for accessing SDL Events

weirddan455:
# Description

SDL_Event is a tagged union. We can just access the relevant member rather than casting.
This fixes a PVS warning.

# Manual testing

Tested that hotplug still works in Quake.

The change has been manually tested on:

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

_Please tick the items as you have addressed them. Don't remove items; leave the ones that are not applicable unchecked._

I have:

- [x] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [x] performed a self-review of my code.
- [ ] commented on the particularly hard-to-understand areas of my code.
- [x] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [x] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [ ] my change has been manually tested on Windows, macOS, and Linux.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] provided the release notes draft (for significant user-facing changes).
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #4138, opened 2025-01-03
Suppress redundant mapper initialization at startup

arifogel:
# Description

* Suppress handling of joystick connection events that occur during joystick subsystem init, which occurs in the middle of first mapper init, which is already aware of such joysticks.

## Related issues

Follows #4122, addressing regression noted in [this comment](https://github.com/dosbox-staging/dosbox-staging/pull/4122#issuecomment-2568469531)

# Manual testing

For each of Windows, macOS, Linux:
* disconnected at startup, then connected later
* connected at startup

In each case, the mapper was only initialized once at startup. Only in the `disconnected at startup case` was a connection event reported in the log.

The change has been manually tested on:

- [X] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

- [X] followed the project's [contributing guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/dosbox-staging/dosbox-staging/blob/master/CODE_OF_CONDUCT.md).
- [X] performed a self-review of my code.
- [X] commented on the particularly hard-to-understand areas of my code.
- [X] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [x] applied the appropriate labels (bug, enhancement, refactoring, documentation, etc.)
- [X] [checked](https://github.com/dosbox-staging/dosbox-staging/blob/main/scripts/compile_commits.sh) that all my commits can be built.
- [X] my change has been manually tested on Windows, macOS, and Linux.
- [ ] confirmed that my code does not cause performance regressions (e.g., by running the Quake benchmark).
- [ ] added unit tests where applicable to prove the correctness of my code and to avoid future regressions.
- [ ] provided the release notes draft (for significant user-facing changes).
- [ ] made corresponding changes to the documentation or the website according to the [documentation guidelines](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md).
- [ ] [locally verified](https://github.com/dosbox-staging/dosbox-staging/blob/main/DOCUMENTATION.md#previewing-documentation-changes-locally) my website or documentation changes.

PR #4146, opened 2025-01-05
Enable hotplugging joysticks (2nd attempt)

### LINKED ISSUE #946: Controller hot-plug support

arifogel:
# Description

* fixes #946
* partially remove REDUCE_JOYSTICK_POLLING hack, which prevented detection of joystick hotplug events:
  * delete inactive code paths assuming REDUCE_JOYSTICK_POLLING not defined
  * preserve code guarded by `REDUCE_JOYSTICK_POLLING` define, except:
    * remove calls to SDL_JoystickEventState (always active now)
* re-initialize mapper when receiving following events:
  * SDL_CONTROLLERDEVICEADDED
  * SDL_CONTROLLERDEVICEREMOVED
  * SDL_JOYDEVICEADDED
  * SDL_JOYDEVICEREMOVED

## Related issues

* Fixes #946
* Re-attempts hotplug support introduced in reverted PRs: #4122, #4138,#4143 (reverted by #4145)

# Release notes
* Added support for hot-plugging of game controllers after program start

# Manual testing

### Linux/Windows/macOS testing

In addition to testing performed for #4122 (inlined below), also tested for *absence* of following regressions introduced by #4122:
* spurious emulated button presses triggered by unmapped host joystick buttons
* non-functional autofire

Specifically, tested with this PR and also compared with main @ a02719911688d4005c772bee86123bf5839bf86f using [JOYCHECK.EXE](https://www.vogons.org/download/file.php?id=7643) as follows:
* with attached mapper configuration: [[mapper-sdl2-0.83.0-alpha.map.zip](https://github.com/user-attachments/files/18309401/mapper-sdl2-0.83.0-alpha.map.zip)] for all runs
* with attached configuration [[dosbox-staging.conf.zip](https://github.com/user-attachments/files/18309418/dosbox-staging.conf.zip)] except as follows:
  * with `autofire=true`, with and without `timed=false`, pressed and held all buttons mapped to virtual joystick buttons
    * with `timed=true`:
      * reported repeat rate on main and PR averaged around 23Hz with minimal variance (Linux/Windows/macOS)
(Linux / Windows)  / ~**4Hz typically, occasionally increasing to up to 8Hz (macOS)**~ 18-40 averaging around 30Hz (macOS with macOS polling hack removed)
    * with `timed=false`:
      * reported repeat rate on main and PR averaged around 23Hz with minimal variance (Linux/Windows/macOS)
(Linux) / 40-50Hz (Windows) /   ~**9Hz typically, occasionally increasing to up to 17Hz (macOS)**~ 28-44Hz (macOS with mac polling hack removed)
      * with `autofire=false`, with and without `timed=false`
        * tested all sticks and buttons which appeared to work correctly
        * pressing unmapped buttons on the host joystick did not result in emulated button presses (spurious button presses occurred in original hotplug work)
      * with `joysticktype=fcs`
        * set joystick type to FC in joycheck, and observed no difference in functionality between main and PR (Windows)
        * **FCS hat did not function properly in main nor in PR on macOS**
      * with `joysticktype=ch`
        * set joystick type to CH in joycheck, and observed no difference in functionality between main and PR (Windows/macOS)
* Additional autofire testing:
  * One must fall 2097 repeated punching and kicking worked fine with corresponding buttons held (needed timed=false for joystick axes to function correctly some but not all of the time, in both main and PR)
  * Commander Keen 1, repeated jumping and repeated pogo toggling worked fine both when buttons held separately and when held while also moving using sticks (did not function correctly in original hotplug work)
* Tested that buttonwrap functions correctly in OMF 2097  in PR (Windows/macOS)
* Repeated previous testing:
   > I tested on Ubuntu 24.04 with XBox 360 Wireless Controller (ID 045e:0719 Microsoft Corp. Xbox 360 Wireless Adapter).
   > I ran DOSBox with and without controller turned on. I verified that controller still functioned in game in each case after power cycling the controller (which causes device to disconnect and reconnect).
* Benchmarking
  * Ran Quake benchmark with no joystick, with joystick improperly calibrated, with joystick properly calibrated, with joystick properly calibrated and left stick held at bottom right following calibration, with each test repeated 3 times
    * There was no noticeable difference between main and PR in frame rate distribution for any of these tests

- [X] Windows
- [X] macOS
- [X] Linux

# Checklist

I have:

- [X] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/CONTRIBUTING.md#commit-messages).
- [X] my change has been manually tested on Windows, macOS, and Linux.

### DIFF
--- a/include/mapper.h
+++ b/include/mapper.h
@@ -87,4 +87,17 @@ void MAPPER_CheckEvent(SDL_Event *event);
 // for good measure.
 constexpr int MaxBindNameLength = 100;
 
+/**
+ * Handle a game controller being connected or disconnected by reinitializing
+ * binds. Caller is responsible for any user interface updates, e.g. updating
+ * the mapper UI buttons to reflect joystick binds being added / removed.
+ *
+ * @param event A pointer to an SDL_JoyDeviceEvent whose type must be one of:
+ * - SDL_CONTROLLERDEVICEADDED
+ * - SDL_CONTROLLERDEVICEREMOVED
+ * - SDL_JOYDEVICEADDED
+ * - SDL_JOYDEVICEREMOVED
+ */
+void MAPPER_HandleJoyDeviceEvent(SDL_JoyDeviceEvent* event);
+
 #endif
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -111,6 +111,9 @@ static CBindList holdlist;
 
 static bool suppress_save_mapper_file_message = false;
 
+// In SDL ticks (msec)
+static auto joystick_subsystem_init_timestamp = std::numeric_limits<uint32_t>::max();
+
 class CEvent {
 public:
 	CEvent(const char *ev_entry) : bindlist{}
@@ -732,10 +735,8 @@ public:
 
 			if (event->jaxis.which != stick_id)
 				return nullptr;
-#if defined(REDUCE_JOYSTICK_POLLING)
 			if (axis_id >= axes)
 				return nullptr;
-#endif
 			if (abs(axis_position) < 25000)
 				return nullptr;
 
@@ -747,11 +748,7 @@ public:
 		} else if (event->type == SDL_JOYBUTTONDOWN) {
 			if (event->jbutton.which != stick_id)
 				return nullptr;
-#if defined (REDUCE_JOYSTICK_POLLING)
-			return CreateButtonBind(event->jbutton.button%button_wrap);
-#else
-			return CreateButtonBind(event->jbutton.button);
-#endif
+			return CreateButtonBind(event->jbutton.button % button_wrap);
 		} else if (event->type==SDL_JOYHATMOTION) {
 			if (event->jhat.which != stick_id) return nullptr;
 			if (event->jhat.value == 0) return nullptr;
@@ -760,32 +757,9 @@ public:
 		} else return nullptr;
 	}
 
-	bool CheckEvent(SDL_Event * event) override {
-		SDL_JoyAxisEvent * jaxis = nullptr;
-		SDL_JoyButtonEvent *jbutton = nullptr;
-
-		switch(event->type) {
-			case SDL_JOYAXISMOTION:
-				jaxis = &event->jaxis;
-				if(jaxis->which == stick_id) {
-					if(jaxis->axis == 0)
-						JOYSTICK_Move_X(emustick, jaxis->value);
-					else if (jaxis->axis == 1)
-						JOYSTICK_Move_Y(emustick, jaxis->value);
-				}
-				break;
-			case SDL_JOYBUTTONDOWN:
-			case SDL_JOYBUTTONUP:
-				jbutton = &event->jbutton;
-				bool state;
-				state = jbutton->type == SDL_JOYBUTTONDOWN;
-				const auto but = check_cast<uint8_t>(jbutton->button % emulated_buttons);
-				if (jbutton->which == stick_id) {
-					JOYSTICK_Button(emustick, but, state);
-				}
-				break;
-			}
-			return false;
+	bool CheckEvent([[maybe_unused]] SDL_Event* event) override
+	{
+		return false;
 	}
 
 	virtual void UpdateJoystick() {
@@ -1026,34 +1000,6 @@ public:
 		JOYSTICK_Enable(1, true);
 	}
 
-	bool CheckEvent(SDL_Event * event) override {
-		SDL_JoyAxisEvent * jaxis = nullptr;
-		SDL_JoyButtonEvent *jbutton = nullptr;
-
-		switch(event->type) {
-			case SDL_JOYAXISMOTION:
-				jaxis = &event->jaxis;
-				if(jaxis->which == stick_id && jaxis->axis < 4) {
-					if(jaxis->axis & 1)
-						JOYSTICK_Move_Y(jaxis->axis >> 1 & 1, jaxis->value);
-					else
-						JOYSTICK_Move_X(jaxis->axis >> 1 & 1, jaxis->value);
-		        }
-		        break;
-			case SDL_JOYBUTTONDOWN:
-			case SDL_JOYBUTTONUP:
-				jbutton = &event->jbutton;
-				bool state;
-				state = jbutton->type == SDL_JOYBUTTONDOWN;
-				const auto but = check_cast<uint8_t>(jbutton->button % emulated_buttons);
-				if (jbutton->which == stick_id) {
-					JOYSTICK_Button((but >> 1), (but & 1), state);
-				}
-				break;
-		}
-		return false;
-	}
-
 	void UpdateJoystick() override {
 		/* query SDL joystick and activate bindings */
 		ActivateJoystickBoundEvents();
@@ -1088,42 +1034,6 @@ public:
 		if (button_wrapping_enabled) button_wrap=emulated_buttons;
 		JOYSTICK_Enable(1,true);
 		JOYSTICK_Move_Y(1, INT16_MAX);
-	}
-
-	bool CheckEvent(SDL_Event * event) override {
-		SDL_JoyAxisEvent * jaxis = nullptr;
-		SDL_JoyButtonEvent * jbutton = nullptr;
-		SDL_JoyHatEvent *jhat = nullptr;
-
-		switch(event->type) {
-			case SDL_JOYAXISMOTION:
-				jaxis = &event->jaxis;
-				if(jaxis->which == stick_id) {
-					if(jaxis->axis == 0)
-						JOYSTICK_Move_X(0, jaxis->value);
-					else if (jaxis->axis == 1)
-						JOYSTICK_Move_Y(0, jaxis->value);
-					else if (jaxis->axis == 2)
-						JOYSTICK_Move_X(1, jaxis->value);
-				}
-				break;
-			case SDL_JOYHATMOTION:
-				jhat = &event->jhat;
-				if (jhat->which == stick_id)
-					DecodeHatPosition(jhat->value);
-				break;
-			case SDL_JOYBUTTONDOWN:
-			case SDL_JOYBUTTONUP:
-				jbutton = &event->jbutton;
-			bool state;
-			state=jbutton->type==SDL_JOYBUTTONDOWN;
-				const auto but = check_cast<uint8_t>(jbutton->button % emulated_buttons);
-				if (jbutton->which == stick_id) {
-					JOYSTICK_Button((but >> 1), (but & 1), state);
-				}
-				break;
-		}
-		return false;
 	}
 
 	void UpdateJoystick() override {
@@ -1214,66 +1124,6 @@ public:
 		emulated_hats=1;
 		if (button_wrapping_enabled) button_wrap=emulated_buttons;
 		JOYSTICK_Enable(1,true);
-	}
-
-	bool CheckEvent(SDL_Event * event) override {
-		SDL_JoyAxisEvent * jaxis = nullptr;
-		SDL_JoyButtonEvent * jbutton = nullptr;
-		SDL_JoyHatEvent * jhat = nullptr;
-		Bitu but = 0;
-		static const unsigned button_magic[6] = {
-		        0x02, 0x04, 0x10, 0x100, 0x20, 0x200};
-		static const unsigned hat_magic[2][5] = {
-		        {0x8888, 0x8000, 0x800, 0x80, 0x08},
-		        {0x5440, 0x4000, 0x400, 0x40, 0x1000}};
-		switch(event->type) {
-			case SDL_JOYAXISMOTION:
-				jaxis = &event->jaxis;
-				if(jaxis->which == stick_id && jaxis->axis < 4) {
-					if(jaxis->axis & 1)
-						JOYSTICK_Move_Y(jaxis->axis >> 1 & 1, jaxis->value);
-					else
-						JOYSTICK_Move_X(jaxis->axis >> 1 & 1, jaxis->value);
-				}
-				break;
-			case SDL_JOYHATMOTION:
-				jhat = &event->jhat;
-				if (jhat->which == stick_id && jhat->hat < 2) {
-					if (jhat->value == SDL_HAT_CENTERED)
-						button_state &= ~hat_magic[jhat->hat][0];
-					if (jhat->value & SDL_HAT_UP)
-						button_state|=hat_magic[jhat->hat][1];
-					if(jhat->value & SDL_HAT_RIGHT)
-						button_state|=hat_magic[jhat->hat][2];
-					if(jhat->value & SDL_HAT_DOWN)
-						button_state|=hat_magic[jhat->hat][3];
-					if(jhat->value & SDL_HAT_LEFT)
-						button_state|=hat_magic[jhat->hat][4];
-				}
-				break;
-			case SDL_JOYBUTTONDOWN:
-				jbutton = &event->jbutton;
-				but = jbutton->button % emulated_buttons;
-				if (jbutton->which == stick_id)
-					button_state|=button_magic[but];
-				break;
-			case SDL_JOYBUTTONUP:
-				jbutton = &event->jbutton;
-				but = jbutton->button % emulated_buttons;
-				if (jbutton->which == stick_id)
-					button_state&=~button_magic[but];
-				break;
-		}
-
-		unsigned i;
-		uint16_t j;
-		j=button_state;
-		for(i=0;i<16;i++) if (j & 1) break; else j>>=1;
-		JOYSTICK_Button(0,0,i&1);
-		JOYSTICK_Button(0,1,(i>>1)&1);
-		JOYSTICK_Button(1,0,(i>>2)&1);
-		JOYSTICK_Button(1,1,(i>>3)&1);
-		return false;
 	}
 
 	void UpdateJoystick() override {
@@ -2637,8 +2487,8 @@ static void ClearAllBinds()
 	}
 }
 
-static void CreateDefaultBinds() {
-	ClearAllBinds();
+static void CreateDefaultBinds()
+{
 	char buffer[512];
 	Bitu i=0;
 	while (DefaultKeys[i].eventend) {
@@ -2760,7 +2610,6 @@ static bool load_binds_from_file(const std::string_view mapperfile_path,
 
 		suppress_save_mapper_file_message = true;
 
-		ClearAllBinds();
 		for (auto &line : lines)
 			CreateStringBind(line.data());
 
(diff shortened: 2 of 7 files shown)


include/mapper.h
```diff
@@ -87,4 +87,17 @@ void MAPPER_CheckEvent(SDL_Event *event);
 // for good measure.
 constexpr int MaxBindNameLength = 100;
 
+/**
+ * Handle a game controller being connected or disconnected by reiniitializing
```
johnnovak, line 91: typo: reinitializing

arifogel: fixed

johnnovak:
Thanks @arifogel, the changes look ok. Are you planning to complete the macOS and Windows testing as well?

arifogel:
> Thanks @arifogel, the changes look ok. Are you planning to complete the macOS and Windows testing as well?

Yeah, I can do that later today.

arifogel:
Rebased, tested in Windows, tested CH and FCS joystick types.

arifogel:
Tested on macOS. Significant autofire repeat rate reduction on macOS for PR as compared to main. Needs investigation.

arifogel:
[Additionally removing the macos-specific polling hack](https://github.com/dosbox-staging/dosbox-staging/compare/2c8799532cfa7b4bafa672cb5a8ad54f5684871b..c3cc7de1accf8de1f9304f5bbd7a085e0207f448) brings autofire repeat rate up to linux/macos rates. I reran benchmarks with this hack removed, and confirmed there is no obvious performance drop.

FeralChild64:
Played a little with my pad on Linux (wireless), Windows 11 (wireless), and macOS 15 (wired) - hot-plugging works, autofire works, I haven't noticed anything suspicious.

Let's wait till more team members do some tests - I'm not even a joystick/pad player :)

src/gui/sdlmain.cpp
```diff
@@ -3815,29 +3805,11 @@ static void handle_user_event(const SDL_Event& event)
...
-	static auto last_check_joystick = GetTicks();
-	auto current_check_joystick = GetTicks();
-	if (GetTicksDiff(current_check_joystick, last_check_joystick) > 20) {
-		last_check_joystick = current_check_joystick;
-		if (MAPPER_IsUsingJoysticks()) SDL_JoystickUpdate();
+	if (MAPPER_IsUsingJoysticks()) {
+		SDL_JoystickUpdate();
 		MAPPER_UpdateJoysticks();
 	}
-#endif
```
weirddan455, line 3840: I took a look at SDL source code and I understand the reasoning behind `REDUCE_JOYSTICK_POLLING` more now.

SDL puts a "sentinel event" inside of its event queue which tells it when it needs to run `SDL_PumpEvents()`.  It runs `SDL_PumpEvents()` inside of `SDL_PollEvent()` but not on every call.  I uses the sentinel to decide when it needs to run.  One of things it does inside of `SDL_PumpEvents()` is this:

```c
/* Check for joystick state change */
if (SDL_update_joysticks) {
    SDL_JoystickUpdate();
}
```

Prior to this PR, `SDL_update_joysticks == false` most of the time due to `SDL_JoystickEventState(SDL_DISABLE)` calls which you have now removed.  This made `SDL_JoystickUpdate()` mandatory to call explicitly since SDL won't do it for us when we have events disabled.

Now that you've re-enabled events, it shouldn't be strictly necessary to call `SDL_JoystickUpdate()` at all.  Furthermore, if the comments here are to be believed, we are calling this function at a much greater frequency than SDL would.  Behavior of `SDL_JoystickUpdate()` varies based on OS and even the type of joystick in some cases but I did see `ioctl()` calls in there for Linux and polling of HID devices on Mac.  So it actually is polling hardware here (or at the very least the driver).

I know you said you tested for performance but I'm still a bit worried this will cause regressions somewhere.  Especially on Mac as it has been noted to be problematic but also the problem may only appear on certain types of joysticks.  It's doing low level driver pooling so maybe it ends up being more expensive on Bluetooth and you didn't notice on your 360 pad.

My suggestion would be to remove `SDL_JoystickUpdate()` entirely and let SDL poll the joystick at its own interval inside `SDL_PollEvents()`.  As I noted on your last PR, you may also want to stick `MAPPER_UpdateJoysticks()` _after_ `SDL_PollEvents()` has run so we get the latest state.

You could also leave this `GetTicks()` stuff as-is if that's problematic to keep polling behavior (mostly) the same as it is on `main`.

If you do decide to change any of this code, at the very least, I would strongly suggest putting this into a 2nd commit.  It's not strictly needed to change the frequency of polling to make hotplug work and if we run into regressions here, it would make it easier to bisect if this was in a separate small commit.

src/gui/sdl_mapper.cpp
```diff
@@ -760,32 +757,11 @@ class CStickBindGroup : public CBindGroup {
 		} else return nullptr;
 	}
 
-	bool CheckEvent(SDL_Event * event) override {
```
weirddan455, line 763: Did you confirm that the mapper GUI still works after removing these `CheckEvent()` functions.  I believe when we spoke on Discord, you had confirmed the one place the mapper was using SDL events over polling was inside the mapper GUI.

weirddan455 (commented):
Sorry for the delay in my review.  I didn't want to keep you hanging.  Just been busy with other stuff.

I have a bit of a concern with changes you've made to joystick polling in our main event loop.  Wrote a lengthy comment below.

I did test this briefly on Linux and it seemed to work fine, hotplug included.

src/gui/sdlmain.cpp
arifogel: > My suggestion would be to remove SDL_JoystickUpdate() entirely and let SDL poll the joystick at its own interval inside SDL_PollEvents(). As I noted on your last PR, you may also want to stick MAPPER_UpdateJoysticks() after SDL_PollEvents() has run so we get the latest state.

Moving `MAPPER_UpdateJoysticks()` inside the loop at all breaks autofire.

arifogel: > Moving MAPPER_UpdateJoysticks() inside the loop at all breaks autofire.

And I will check, but I believe since we need joystick updates outside the loop that we need `SDL_JoystickUpdate` outside the loop as well to preserve autofire.

src/gui/sdl_mapper.cpp
arifogel: Yes mapper GUI still works. These functions were not used for mapper GUI.

src/gui/sdlmain.cpp
arifogel: Autofire seems to function fine with `SDL_JoystickUpdate()` call removed in Linux

arifogel: Let me test on macOS with sdl update call removed, but without removing the macos polling hack.

arifogel: On macOS with sdl_update_call removed and macos polling hack left in, autofire rate is heavily degraded.

arifogel: Note that on macOS I am testing with XBox Series X wireless controller (bluetooth)

weirddan455: >  Moving MAPPER_UpdateJoysticks() inside the loop at all breaks autofire.

I meant after the poll loop, not inside it.

arifogel: > I know you said you tested for performance but I'm still a bit worried this will cause regressions somewhere. Especially on Mac as it has been noted to be problematic but also the problem may only appear on certain types of joysticks. It's doing low level driver pooling so maybe it ends up being more expensive on Bluetooth and you didn't notice on your 360 pad.

I did a few more benchmarks on macOS, and I need to correct the results I posted above. There is a consistent performance drop with this PR, with or without the removal of the call to `SDL_UpdateJoystick()`, likely because it is being called internally more often. The drop in one test is from 152 average fps to 146 average fps. It's small (~4%), but it's consistently faster on main than on this PR.

arifogel: > I meant after the poll loop, not inside it.

Moving this makes the joystick feel different, but I cannot objectively state how. It's not obviously better or worse.. it might make no difference perceptually. Happy to move to end, but not sure how much better that is than just manually polling and calling update as we were before. It might be redundant, but that is not the source of the performance drop best I can tell. I will check, but I believe the drop in frame rate is just due to running the event loop more often by removing the macos "polling" hack

arifogel: Confirmed that frame drop is due to just running the rest of the function more often OTHER than joystick stuff.
If I restore the macos polling hack which early exits `GFX_Events`, then move the joystick stuff before the early exit, the performance is back to `main` level and autofire is not degraded.

arifogel: Latest commit restores the macos crap but moves joystick polling before it. Note I haven't put joystick update at after polling event loop yet, because I didn't want to fiddle with macos preprocessor stuff.

johnnovak: > My suggestion would be to remove SDL_JoystickUpdate() entirely and let SDL poll the joystick at its own interval inside SDL_PollEvents(). As I noted on your last PR, you may also want to stick MAPPER_UpdateJoysticks() after SDL_PollEvents() has run so we get the latest state.

Agree with @weirddan455 here. Either leave everything as is, OR move to the "SDL managed" model where we rely on events and don't do any polling. Don't mix the two please, anything can happen like @weirddan455 explained—we just don't know what happens on all sorts of weird setups when you get both events and also poll.

> The drop in one test is from 152 average fps to 146 average fps. It's small (~4%), but it's consistently faster on main than on this PR.

Yeah, let's not fixate on such small differences—it does not matter. The conceptual integrity of the implementation is what matters.

> Moving this makes the joystick feel different, but I cannot objectively state how. It's not obviously better or worse.. it might make no difference perceptually. Happy to move to end, but not sure how much better that is than just manually polling and calling update as we were before. It might be redundant, but that is not the source of the performance drop best I can tell. I will check, but I believe the drop in frame rate is just due to running the event loop more often by removing the macos "polling" hack

As above. Don't dwell on the 4% perf. difference, it does not matter. Let's remove any hacks and that's it.

weirddan455: >  Agree with @weirddan455 here. Either leave everything as is, OR move to the "SDL managed" model where we rely on events and don't do any polling. Don't mix the two please, anything can happen like @weirddan455 explained—we just don't know what happens on all sorts of weird setups when you get both events and also poll.

That's not what I was saying.  Ideally, I think we should use events only but he tried that and it didn't work out so this PR is using polling like it is in `main`.  Also, events must be turned on in order for hotplug to work.  I believe that's the only thing using events now.

The part I was worried about is removing the timing checks to limit how often `SDL_UpdateJoystick()` gets called.  I haven't checked myself how often we call `GFX_Events()` but from the comments it's thousands of times per second so it very well could bite us on a hardware/OS configuration we haven't tested yet.

arifogel: Let me summarize options and findings here, because this discussion is getting to be a bit of a mess:
|macOS hack|MAPPER_UpdateJoysticks position|Autofire performance|Quake FPS|
|---|--|--|--|
|present|before macOS hack early exit (beginning of function)|faster than main|unaffected|
|absent|before OR after event loop|faster than main|-4% (macOS, possibly others)|
|present|after hack, before OR after event loop|faster than main Windows/Linux, but only 4Hz macOS|unaffected|
|*|in event loop|completely broken (all OSes)|no data|

arifogel: * So the code right now in [f4fa842](https://github.com/arifogel/dosbox-staging/commit/f4fa842a92bad01f5adb8f01ee78675ab215d417) implements Option 1.
* Sounds like @johnnovak is advocating for option 2
* Sounds like @FeralChild64 is advocating for option 2, but specifically after event loop - or some version of option 1 that is still after event loop. I imagine the latter would be very dirty.
* If no one cares about the 4%, option 2 is certainly cleaner and appears to work just fine
* Best I can tell we can remove manual call to `SDL_UpdateJoystick` without any perceptual difference in response.

weirddan455: What do mean by "autofire performance"?  Are you talking about how quickly the button gets automatically pressed or are you talking about measurable compute time spent inside mapper functions?  The former I would not qualify as "performance" and could be considered a regression if firing faster is undesirable behavior.

Also, forget about `MAPPER_UpdateJoysticks()` position for now.  That was a minor gripe to put it after the loop.  Really the only difference should be reducing of a couple ms of input lag but don't worry about that for now.  Not even that big of a deal, it can stay where it is if it's problematic.

My main point is that `SDL_JoystickUpdate()` is getting called much more frequently than it is in `main` and you haven't given a clear reason for that change.

Any changes that you want to make to `GFX_Events()` (besides removing the `#ifdef` as that won't change behavior) need to be in a separate commit with a clear explanation of what you're changing and why.  It's not needed for hotplug and it could introduce regressions.

arifogel: > What do mean by "autofire performance"?

Repeat rate as reported by joycheck when autofire is enabled and all buttons are held down. In the case of dropping from 24Hz to 4Hz on macOS, this is obviously a regression. In all other cases tested with hotplug enabled, the repeat rate has a higher variance and average than main. The higher rate and/or variance could both be considered regressions in a sense as well.

> My main point is that SDL_JoystickUpdate() is getting called much more frequently than it is in main and you haven't given a clear reason for that change.

I think there has been some confusion as well as disagreement on what parts of the `REDUCE_JOYSTICK_POLLING` hack need to be removed. More to the point, we can certainly leave in the original rate limiting code that was guarded by that `#ifdef`. Restoring that code results in hotplug working and autofire firing at main rates.

See [latest commit](https://github.com/dosbox-staging/dosbox-staging/commit/620e56d0ba2330f840703186f37504234ea4ea27)

As far as the explicit call to `SDL_JoystickUpdate()`, that can be removed with no noticeable difference in behavior or performance AFAICT. I left it in in the above commit, which is less of a departure from `main`. Regardless though, it is likely that HW is being polled more often than in that rate limited block, because joystick events are enabled for hotplug to work.

arifogel: Tested https://github.com/dosbox-staging/dosbox-staging/commit/620e56d0ba2330f840703186f37504234ea4ea27 above on macOS:
* autofire works at main rate (~22Hz)
* Quake benchmark FPS same as main

johnnovak: Ok, so to reiterate it again @arifogel, and to take a step back, the goal here was to do the bare minimum to get hotplugging work and not to touch anything else. Is it possible to just do that? Just the bare minimum? That would be the best, then we can move in with our lives...

src/gui/sdlmain.cpp
```diff
@@ -3827,17 +3817,16 @@ bool GFX_Events()
...
-		last_check_joystick = current_check_joystick;
-		if (MAPPER_IsUsingJoysticks()) SDL_JoystickUpdate();
-		MAPPER_UpdateJoysticks();
+	if (MAPPER_IsUsingJoysticks()) {
+		auto current_check_joystick = GetTicks();
+		if (GetTicksDiff(current_check_joystick, last_check_joystick) > 20) {
+			last_check_joystick = current_check_joystick;
+			SDL_JoystickUpdate();
+			MAPPER_UpdateJoysticks();
+		}
```
weirddan455, line 3827: Commenting on your latest changes now.  You've reverted most of your changes to this function but again we have a divergence from `main` with no explanation as to why.

Previous code always ran `MAPPER_UpdateJoysticks()` when the duration exceeded 20ms.  You've now put this entire block behind `MAPPER_IsUsingJoysticks()`.  It makes logical sense that we shouldn't be updating joysticks if we're not using any but the mapper is a pile of spaghetti and this could very well break something.

Again, I'm just going to insist that any significant changes to this function be in a separate commit with an explanation.

src/gui/sdl_mapper.cpp
```diff
@@ -760,32 +757,11 @@ class CStickBindGroup : public CBindGroup {
...
-				}
-				break;
-			}
-			return false;
+	bool CheckEvent(SDL_Event* event) override
+	{
+		// suppress unused warning
+		(void)event;
+		return false;
 	}
```
weirddan455, line 765: `[[maybe_unused]] SDL_Event *event` in the function signature is a cleaner way of suppressing this warning.

arifogel: fixed

src/gui/sdlmain.cpp
arifogel: Reverted the rest except the `#if defined REDUCE_JOYSTICK_POLLING`

src/gui/sdlmain.cpp
weirddan455: >  In the case of dropping from 24Hz to 4Hz on macOS, this is obviously a regression.

Why did this happen?  You really need to understand the why before you go trying to fix regressions.

>  I think there has been some confusion as well as disagreement on what parts of the REDUCE_JOYSTICK_POLLING hack need to be removed.

The problem is you haven't really removed the hack at all.  With `REDUCE_JOYSTICK_POLLING` undefined, the mapper is supposed to function using only SDL events, not polling.  That code path is broken though, so you've removed it.  I'm good with removing broken/dead code but don't try to remove half of the working code path.

>  More to the point, we can certainly leave in the original rate limiting code that was guarded by that #ifdef. Restoring that code results in hotplug working and autofire firing at main rates.

I've seen your latest commit.  Does this mean the 24hz to 4hz regression on MacOS was caused by your changes to `GFX_Events()` which you've now reverted?  Everything is good on all OS's now?

>  the goal here was to do the bare minimum to get hotplugging work and not to touch anything else. Is it possible to just do that? Just the bare minimum? That would be the best, then we can move in with our lives...

Yes, please.

>  If hotplugging needs us to get the SDL controller/joystick events that were disabled previously (because we were doing poling ourselves), and now the events AND the polling can have unintended side effects, I might just abort the whole thing.

He's reverted some of his changes and I think there's less of a chance at regression now.  `SDL_UpdateJoystick()` will now be called periodically inside `SDL_PollEvents()` but I don't think that would be a huge problem.  I was mainly taking issue with having an explicit `SDL_UpdateJoystick()` with no rate limiting.

>  Overall, I'd say the changes are too risky for something as trivial and unimportant as controller hot plugging. @FeralChild64 @weirddan455 are we in agreement?

I think it's better now after the latest push,  Commented a minor concern below.  It's your call but I'm fine with merging this if we can remove changes to our main event loop in `GFX_Events()`.

arifogel: @johnnovak I believe I've pared back everything not required for hotplugging. But given your reservations, I'm going to put this down unless and until the team arrives at a consensus that there is a version of this work that would be accepted.

johnnovak: >I think it's better now after the latest push, Commented a minor concern below. It's your call but I'm fine with merging this if we can remove changes to our main event loop in GFX_Events() and the autofire regressions have been fixed.

I'm fine with that, just I stand by my earlier comment that I don't wanna go deeply into this mapper bullshit. We'll merge this, then if we'll get regression reports and this change is the culprit, we'll revert and that's the end of it.

arifogel: > I've seen your latest commit. Does this mean the 24hz to 4hz regression on MacOS was caused by your changes to GFX_Events() which you've now reverted? Everything is good on all OS's now?

Current commit - with no functional changes to `GFX_Events()` - is indistinguishable from main AFAICT wrt autofire and FPS on macOS. I can retest this version on other OSes if there is a consensus to move forward.

arifogel: Retested on Linux. Same FPS and autofire rate as main

johnnovak: >  I can retest this version on other OSes if there is a consensus to move forward.

Yes please, then if @weirddan455 is happy, we'll test the merge button and wait for user feedback.

arifogel: Retested on Windows. Same FPS and autofire rate as main.

Also retested OMF 2097 with and without autofire. Appears to be working properly.

Updated top comment in line with latest results, i.e. removed blurbs about autofire rate divergence from main.

FeralChild64: > Sounds like @FeralChild64 is advocating for option 2

@arifogel I'm not advocating towards anything - I merely did some tests on the hardware I have. I have no knowledge about the joystick emulation or mapper inner working, I use my pad rarely (so not much gamer experience here), and I don't even know how to use the mapper.

> @FeralChild64 @weirddan455 are we in agreement?

@johnnovak I would leave the decision to @weirddan455.

arifogel: Sorry that was a typo @FeralChild64, I meant @weirddan455 was advocating. Regardless, that part of the discussion has been mooted by later comments.

weirddan455 (approved):
I think this is good now.  It's fairly minimal and contained to mapper code now.  @johnnovak @FeralChild64 Do you have any objections to merge?

johnnovak:
> I think this is good now. It's fairly minimal and contained to mapper code now. @johnnovak @FeralChild64 Do you have any objections to merge?

I'm good, merging.

@arifogel Thanks for your perseverance on this and not giving up halfway through (I would've gotten close to it if I were you 😅).

It was a good lesson/reminder for us all to _not_ touch the mapper code unless we absolutely have to 😆 Let's hope this won't cause regressions down the line and we can keep your nice hotplugging improvement for 0.83.0! 🚀

arifogel:
Thanks everyone for donating their time to shepherd this PR!

PR #4149, opened 2025-01-08
On Windows, create a temp file when deleting a DOS file with handles still open

weirddan455:
# Description

See comments on #4142 for context

Regression first appeared in eeb5374bbe29cf9c298cc816b1cd85a3eca90ca6 with the move to Win32 API I/O.  `FLAG_SHARE_DELETE` broke an old (less accurate than this) hack to workaround this behavior.

b8f9015b577f79efe3d7f74d3b51352def9b97a9 - Later commit where I removed the old hack.  `FLAG_SHARE_DELETE` made the delete succeed with open handles so it never got executed to close existing handles after the switch to Win32 API I/O.

This more accurately emulates how real DOS handles files.  This PR handles an edge case when a DOS game deletes a file with handles still open and then tries to create a new file with the same name.  This only affects Windows.  POSIX systems handle this correctly by default.

My solution is to create a temp file using the Windows API to find a location.  The Visual Studio build places this in `AppData\Local\Temp`.  MSYS2 places it in `C:\msys64\tmp` but I think that's because I was launching it from the MSYS terminal with whatever enviornment variables it sets.  Both builds work correctly however.

When a file is deleted, it scans the open DOS handles to see if there is a local (native/host) file open at the same place we're trying to delete from.  If so, copy the contents to a temp file and replace the native file handle with the temp file.  At the end, we can delete the temp file and the handles remain valid (to avoid cluttering the user's temp folder on multiple launches).

This happens with the DOS layer being none the wiser that the file got moved on the host.

## Related issues

#4123

# Release notes

Fixed an issue where the game Abuse failed to launch on Windows 7 and on newer versions of Windows when using a network drive.

# Manual testing

- Abuse works
- Tested Visual Studio and MSYS2 builds
- Crystal Caves saves work
- Doom saves work
- Read after delete works - Tested with https://gist.github.com/weirddan455/5cae02fc3d6f932fe55025138ed2f47c

- [x] Windows
- [ ] macOS
- [ ] Linux

This change only affects Windows.

# Checklist

I have:

### DIFF
--- a/src/dos/drive_local.cpp
+++ b/src/dos/drive_local.cpp
@@ -232,6 +232,79 @@ std::string localDrive::MapDosToHostFilename(const char* const dos_name)
 	return dirCache.GetExpandNameAndNormaliseCase(newname);
 }
 
+void localFile::ReplaceNativeFileHandle(const NativeFileHandle new_handle, const std_fs::path& new_path)
+{
+	close_native_file(file_handle);
+	file_handle = new_handle;
+	path = new_path;
+}
+
+[[maybe_unused]] static void copy_file_contents(const NativeFileHandle source_handle, const std_fs::path& dest_path)
+{
+	const auto dest_handle = open_native_file(dest_path, true);
+	if (dest_handle == InvalidNativeFileHandle) {
+		LOG_WARNING("FS: Failed to open file '%s'", dest_path.string().c_str());
+		return;
+	}
+	seek_native_file(source_handle, 0, NativeSeek::Set);
+	uint8_t buffer[4096] = {};
+	while (true) {
+		const auto read_result = read_native_file(source_handle, buffer, sizeof(buffer));
+		if (read_result.num_bytes < 1) {
+			break;
+		}
+		write_native_file(dest_handle, buffer, read_result.num_bytes);
+	}
+	close_native_file(dest_handle);
+}
+
+// Windows specific hack. This handles the case where a file gets unlinked with file handles still open.
+// FILE_SHARE_DELETE gets mostly the correct behavior here except it prevents a new file with the same name from being created.
+// See issue with the game Abuse here: https://github.com/dosbox-staging/dosbox-staging/issues/4123
+// POSIX gives us the behavior we want by default so nothing needs to be done on non-Windows OSs.
+#if defined(WIN32)
+static void move_open_handles_to_temp(const std_fs::path& path)
+{
+	std_fs::path temp_filename = {};
+	for (auto& file : Files) {
+		auto local_file = dynamic_cast<localFile*>(file.get());
+		if (!local_file || local_file->GetPath() != path) {
+			continue;
+		}
+		auto file_pos = get_native_file_position(local_file->file_handle);
+		if (file_pos == NativeSeekFailed) {
+			LOG_WARNING("FS: Failed to get current file pos. Setting to zero and hoping for the best :)");
+			file_pos = 0;
+		}
+		if (temp_filename.empty()) {
+			temp_filename = create_temp_file();
+			if (temp_filename.empty()) {
+				LOG_WARNING("FS: Failed to create temp file");
+				return;
+			}
+			copy_file_contents(local_file->file_handle, temp_filename);
+		}
+		const uint8_t permissions = local_file->flags & 0xf;
+		bool write_access = false;
+		if (permissions == OPEN_WRITE || permissions == OPEN_READWRITE) {
+			write_access = true;
+		}
+		const auto temp_handle = open_native_file(temp_filename, write_access);
+		if (temp_handle == InvalidNativeFileHandle) {
+			LOG_WARNING("FS: Failed to open temp file '%s'", temp_filename.string().c_str());
+			continue;
+		}
+		seek_native_file(temp_handle, file_pos, NativeSeek::Set);
+		local_file->ReplaceNativeFileHandle(temp_handle, temp_filename);
+	}
+	// After all file handles have been moved, we can delete the temp file
+	// Since we open files with FILE_SHARE_DELETE this will succeed and the handles are still accessible
+	if (!temp_filename.empty() && !DeleteFileW(temp_filename.c_str())) {
+		LOG_WARNING("FS: Failed to delete temp file '%s'", temp_filename.string().c_str());
+	}
+}
+#endif
+
 // Attempt to delete the file name from our local drive mount
 bool localDrive::FileUnlink(const char* name)
 {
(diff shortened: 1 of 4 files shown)


weirddan455:
Note to myself:  The `path` variable needs to be changed as well to point to the temp path.  Otherwise, there could be bugs if the same file gets deleted and created multiple times.

I think it's just a one-line change but I just booted back into Linux and then thought about this.  I'll make this change later so hold off on merge until then but it's otherwise ready for review.

weirddan455:
Above is fixed and re-tested.  Not quite a one-liner because I added a small member method to `localFile` but easy enough.

johnnovak:
I'm not exactly in love with this temp file solution, it seems kinda clunky. Is the idea that we'll only be dealing with small files here?

Have you entertained "virtualising"/decoupling the DOS-side operation from what we're _actually_ doing on the host system like I explained in my possible implementation idea [in this comment](https://github.com/dosbox-staging/dosbox-staging/issues/4123#issuecomment-2570054692). Is there anything wrong with that solution? IMO, it wouldn't complicate the code that much, it would be a lot cleaner, and it wouldn't involve temp files. Moreover, the emulation of the DOS-side delete/reopen would be implemented in code, in an OS-agnostic manner. With this solution (apart from the clunkiness) we'd be relying on OS/filesystem specific behaviour that kinda works on Linux/macOS by fluke.

> Again, we could complicate the native to DOS file mapping code to flag that when this happens, we close all open native handles for that file first and proceed with the physical delete, then mark all open DOS handles with a flag. If that flag is set, all subsequent operations on the flagged and still-open DOS handles would be simply no-ops. That would also work, no?
> 
> All I'm saying, nobody says there should be a strict 1:1 mapping between DOS and native file operations; we've been able to get away with that so far, but there's no "rule" it must be so.

weirddan455:
There's problems with that approach.  From your initial idea:

>  we close all open native handles for that file first and proceed with the physical delete, then mark all open DOS handles with a flag. If that flag is set, all subsequent operations on the flagged and still-open DOS handles would be simply no-ops

That solves the bug Abuse and would let us create a new file with the same name _but_ it wouldn't let programs read data after a delete.  We'd have to flag it and return an error on read or return all zeros or something (not what real DOS does).

>  So effectively we'd be doing "refcounting" (kinda) on the file handles; the delete won't actually happen until there are still "open references" on the handle, we just mark the file to be deleted. Then when the refcount is zero, we actually delete it, but not before that, so it remains readable.

This would not solve the Abuse bug.  The problem there is not read after delete.  The delete must happen immediately so that a new file with the same name can be created.

I thought about this for a long time and could not figure out a way to solve both of these problems without copying the old data somewhere.  Could be an in-memory copy if you really don't want the temp files but that has its own problems and additional complexity.

>  I'm not looking forward to this scenario happening with an ISO file

Well, it wouldn't happen with an ISO file you intend to mount with DOSBox because that's not part of the DOS filesystem.  It could happen with large files potentially but it's also an edge case that most DOS programs will never hit.  It only makes the temp file if a program holds open file handles and then deletes the file with them open.

weirddan455:
>  We could then delete these files on exit, just to make sure they're gone (as an exit hook kind of thing), or even delete them "on sight" whenever we mount something (but that would involve finding files in a directory, so maybe not the best idea).

I actually have a better solution for delete in this PR already.  I just delete the temp file immediately after opening the extra handles to it.  `FILE_SHARE_DELETE` (which we're already using) allows us to keep file handles open to deleted files so we don't have to go back and clean up after the fact.

>  Would it be possible to just move the file, though (instead of deleting it), and keep it within the same host folder?

I thought about that as well.  What stopped me was the fact that the temp files would be visible to DOS.  I didn't have the immediate delete in mind at the time though.  With that, they wouldn't be visible to DOS _but_ there's a very edge case where a DOS program wants to create a file with the same name as our temp file and it errors out.

Deleting a file while it's open makes kind of a "ghost file".  It's not actually visible on the drive but if you try to create a file with the same name, Windows throws an error.  It also throws an error if you try to open it after the delete.

Anyway it's still an option.  We could rename it to some random name and then delete it.  I would just have to write a temp file name chooser that spits out some unique golbitty gook that hopefully no real program will try to use :smile:

johnnovak:
> I actually have a better solution for delete in this PR already. I just delete the temp file immediately after opening the extra handles to it. FILE_SHARE_DELETE (which we're already using) allows us to keep file handles open to deleted files so we don't have to go back and clean up after the fact.

Nice! I like this solution 💯 🎖️ 

> Anyway it's still an option. We could rename it to some random name and then delete it. I would just have to write a temp file name chooser that spits out some unique golbitty gook that hopefully no real program will try to use 😄

We have a winner there, let's do that! How about just using `std::tmpnam`?

https://en.cppreference.com/w/cpp/io/c/tmpnam

src/dos/drive_local.cpp
```diff
@@ -232,6 +232,79 @@ std::string localDrive::MapDosToHostFilename(const char* const dos_name)
...
+{
+	std_fs::path temp_filename = {};
+	for (auto& file : Files) {
+		auto local_file = dynamic_cast<localFile*>(file.get());
+		if (!local_file || local_file->GetPath() != path) {
+			continue;
+		}
+		auto file_pos = get_native_file_position(local_file->file_handle);
+		if (file_pos == NativeSeekFailed) {
+			LOG_WARNING("FS: Failed to get current file pos. Setting to zero and hoping for the best :)");
```
johnnovak, line 276: Let's keep the log messages a bit more formal 😄 "Setting to zero" will suffice, and it's "file position" 😏

src/misc/fs_utils_win32.cpp
```diff
@@ -256,4 +256,28 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+	// Max path length plus null terminator
+	constexpr DWORD BUFFER_SIZE = MAX_PATH + 1;
+	wchar_t temp_path[BUFFER_SIZE] = {};
+	const DWORD path_len = GetTempPathW(BUFFER_SIZE, temp_path);
+	if (path_len == 0) {
+		return {};
+	}
+	assert(path_len <= BUFFER_SIZE);
+
+	// Prefix only uses first 3 letters, abreviate DosBox Staging
```
johnnovak, line 270: Typo: abbreviate

weirddan455:
Most of the code will go away if I'm able to get move/rename working.  I think I won't need to move file handles in that case.  I just need to set aside some time to test things.  Setting this to draft for now.

weirddan455:
Closing this.  I'll be going with the approach on #4152 instead.

PR #4152, opened 2025-01-11
Rename files on Windows before delete

weirddan455:
# Description

Second attempt at #4149

Same basic principle of creating a temp file but now we create it in the same directory as the existing file to avoid copying potentially large files to a separate drive/partition.

It ended up being quite a bit cleaner on the code side as well as I was able to encapsulate all the Windows specific jank inside `fs_utils_win32.cpp` and there's no more need to move file handles.  Existing file handles remain valid after a rename/move.

## Related issues

#4123

# Release notes

Fixed an issue where the game Abuse failed to launch on Windows 7 and on newer versions of Windows when using a network drive.

# Manual testing

- Abuse Works
- Crystal Caves saves works
- Read after delete works - Tested with https://gist.github.com/weirddan455/5cae02fc3d6f932fe55025138ed2f47c

- [x] Windows
- [ ] macOS
- [x] Linux

~~Will test Linux later.~~  Tested above on Linux as well.  Only change there was to use `unlink()` over `remove()` but `remove()` was a very thin wrapper over `unlink()` so I don't expect any regressions.

# Checklist

I have:

### DIFF
--- a/src/misc/fs_utils_win32.cpp
+++ b/src/misc/fs_utils_win32.cpp
@@ -258,46 +258,58 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
 
 // Includes a Windows specific hack.
 // We move the file to be deleted to a temp file before delete.
-// This allows a file with the same name to be created even if there are open file handles.
-// See bug report about the game Abuse: https://github.com/dosbox-staging/dosbox-staging/issues/4123
+// This allows a file with the same name to be created even if there are open
+// file handles. See bug report about the game Abuse:
+// https://github.com/dosbox-staging/dosbox-staging/issues/4123
 bool delete_native_file(const std_fs::path& path)
 {
 	// Prefix of the temp file. Only uses 3 characters.
 	// $ as convention to designate temp file followed by DB for DosBox.
-	const wchar_t *prefix = L"$DB";
+	const wchar_t* prefix = L"$DB";
 
 	// Zero for UniqueNumber uses system time.
 	// Zero also means it creates the file and ensures it is unique.
 	constexpr UINT UniqueNumber = 0;
 
-	// Maximum return size of GetTempFileNameW() is MAX_PATH plus a null terminator.
+	// Maximum return size of GetTempFileNameW() is MAX_PATH plus a null
+	// terminator.
 	wchar_t temp_file[MAX_PATH + 1] = {};
 
-	// Create a temporary file inside the same directory as the file to be deleted.
-	// Returns unique identifier on success and 0 on failure.
-	if (GetTempFileNameW(path.parent_path().c_str(), prefix, UniqueNumber, temp_file) == 0) {
-		LOG_ERR("FS: Failed to create temp file. Deleting '%s' directly.", path.string().c_str());
-		// We failed to create a temp file but we should still try to delete the original file.
+	// Create a temporary file inside the same directory as the file to be
+	// deleted. Returns unique identifier on success and 0 on failure.
+	if (GetTempFileNameW(path.parent_path().c_str(), prefix, UniqueNumber, temp_file) ==
+	    0) {
+		LOG_ERR("FS: Failed to create temp file. Deleting '%s' directly.",
+		        path.string().c_str());
+		// We failed to create a temp file but we should still try to
+		// delete the original file.
 		return DeleteFileW(path.c_str());
 	}
 
-	// Above function creates the temp file so we replace it here with the flag MOVEFILE_REPLACE_EXISTING.
+	// Above function creates the temp file so we replace it here with the
+	// flag MOVEFILE_REPLACE_EXISTING.
 	if (MoveFileExW(path.c_str(), temp_file, MOVEFILE_REPLACE_EXISTING)) {
 		// We can immediately delete the temporary file.
 		// Any open handles can be still be read from or written to.
 		// A new file can be created with the original file name.
 		if (!DeleteFileW(temp_file)) {
-			LOG_ERR("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
-			// Failed to delete the temp file but the original file was moved.
-			// We should still return success so that the filesystem removes the original file from the cache.
+			LOG_ERR("FS: Failed to delete temporary file: '%s'",
+			        std_fs::path(temp_file).string().c_str());
+			// Failed to delete the temp file but the original file
+			// was moved. We should still return success so that the
+			// filesystem removes the original file from the cache.
 		}
 		return true;
 	}
 
-	// We failed to move the file. We need to delete both the temp file and the original file.
-	LOG_ERR("FS: Failed to move '%s' to temp file '%s' before delete.", path.string().c_str(), std_fs::path(temp_file).string().c_str());
+	// We failed to move the file. We need to delete both the temp file and
+	// the original file.
+	LOG_ERR("FS: Failed to move '%s' to temp file '%s' before delete.",
+	        path.string().c_str(),
+	        std_fs::path(temp_file).string().c_str());
 	if (!DeleteFileW(temp_file)) {
-		LOG_ERR("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
+		LOG_ERR("FS: Failed to delete temporary file: '%s'",
+		        std_fs::path(temp_file).string().c_str());
 	}
 	return DeleteFileW(path.c_str());
 }
--- a/src/misc/fs_utils_win32.cpp
+++ b/src/misc/fs_utils_win32.cpp
@@ -256,4 +256,9 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
 	SetFileTime(handle, nullptr, nullptr, &write_time);
 }
 
+bool delete_native_file(const std_fs::path& path)
+{
+	return remove(path.string().c_str()) == 0;
+}
+
 #endif
--- a/src/misc/fs_utils_win32.cpp
+++ b/src/misc/fs_utils_win32.cpp
@@ -256,9 +256,50 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
 	SetFileTime(handle, nullptr, nullptr, &write_time);
 }
 
+// Includes a Windows specific hack.
+// We move the file to be deleted to a temp file before delete.
+// This allows a file with the same name to be created even if there are open file handles.
+// See bug report about the game Abuse: https://github.com/dosbox-staging/dosbox-staging/issues/4123
 bool delete_native_file(const std_fs::path& path)
 {
-	return remove(path.string().c_str()) == 0;
+	// Prefix of the temp file. Only uses 3 characters.
+	// $ as convention to designate temp file followed by DB for DosBox.
+	const wchar_t *prefix = L"$DB";
+
+	// Zero for UniqueNumber uses system time.
+	// Zero also means it creates the file and ensures it is unique.
+	constexpr UINT UniqueNumber = 0;
+
+	// Maximum return size of GetTempFileNameW() is MAX_PATH plus a null terminator.
+	wchar_t temp_file[MAX_PATH + 1] = {};
+
+	// Create a temporary file inside the same directory as the file to be deleted.
+	// Returns unique identifier on success and 0 on failure.
+	if (GetTempFileNameW(path.parent_path().c_str(), prefix, UniqueNumber, temp_file) == 0) {
+		LOG_ERR("FS: Failed to create temp file. Deleting '%s' directly.", path.string().c_str());
+		// We failed to create a temp file but we should still try to delete the original file.
+		return DeleteFileW(path.c_str());
+	}
+
+	// Above function creates the temp file so we replace it here with the flag MOVEFILE_REPLACE_EXISTING.
+	if (MoveFileExW(path.c_str(), temp_file, MOVEFILE_REPLACE_EXISTING)) {
+		// We can immediately delete the temporary file.
+		// Any open handles can be still be read from or written to.
+		// A new file can be created with the original file name.
+		if (!DeleteFileW(temp_file)) {
+			LOG_ERR("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
+			// Failed to delete the temp file but the original file was moved.
+			// We should still return success so that the filesystem removes the original file from the cache.
+		}
+		return true;
+	}
+
+	// We failed to move the file. We need to delete both the temp file and the original file.
+	LOG_ERR("FS: Failed to move '%s' to temp file '%s' before delete.", path.string().c_str(), std_fs::path(temp_file).string().c_str());
+	if (!DeleteFileW(temp_file)) {
+		LOG_ERR("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
+	}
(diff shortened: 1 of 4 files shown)


src/misc/fs_utils_win32.cpp
```diff
@@ -256,4 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+		}
+		return true;
+	}
+
+	// We failed to move the file. We need to delete both the temp file and the original file.
+	LOG_WARNING("FS: Failed to move '%s' to temp file '%s' before delete.", path.string().c_str(), std_fs::path(temp_file).string().c_str());
+	if (!DeleteFileW(temp_file)) {
+		LOG_WARNING("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
+	}
+	return DeleteFileW(path.c_str());
```
shermp, line 301: How does this all work with `path` and the `W` functions?

AFAIK, whether `path.c_str()` is `char*` or `wchar_t*` depends on how `path` is constructed is it not? Or is the compiler/library smart enough to implicitly convert from `char` to `wchar`?

weirddan455: It's always `wchar_t*` on Windows and `char*` on POSIX.  The return type has to be known at compile time so you can't have 2 different ways of backing a `std::path` on the same OS.  On Windows when a `std::path` gets constructed, it gets implicitly converted from `char` to `wchar` if needed.

https://en.cppreference.com/w/cpp/filesystem/path/native

>  The native string representation of the pathname, using native syntax, native character type, and native character encoding. This string is suitable for use with OS APIs.

shermp: Oh, I see. I did a bit more reading on cppreference, and I think are indeed correct. I _thought_ that `path::value_type` would depend on the value passed into the path constructor (it is a template after all), but it _seems_ `path::value_type` is just `wchar_t` on Windows. Definitely not exactly obvious.

Anyway, my bad. Was my misunderstanding of the docs.

weirddan455: I think if `value_type` was templated, you would have to explicitly state the type when you pass it around to functions and such.  See `std::vector` for example where you have to do ex: `std::vector<float> &foo` in the function definition.

It also spells it out here under "Member Types":

>  value_type:  character type used by the native encoding of the filesystem: char on POSIX, wchar_t on Windows

More proof if you're not convinced.  This is from Linux C++ headers since I'm not in Windows at the moment.  It's just a using statement that gets `#ifdef`'d depending on OS:

```cpp
#ifdef _GLIBCXX_FILESYSTEM_IS_WINDOWS
    using value_type = wchar_t;
    static constexpr value_type preferred_separator = L'\\';
#else
# ifdef _GLIBCXX_DOXYGEN
    /// Windows uses wchar_t for path::value_type, POSIX uses char.
    using value_type = __os_dependent__;
# else
    using value_type =  char;
# endif
```

shermp: Yeah, is was me misunderstanding the docs. Carry on.

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
 	SetFileTime(handle, nullptr, nullptr, &write_time);
 }
 
+// Includes a Windows specific hack
+// We move the file to be deleted to a temp file before delete
+// This allows a file with the same name to be created even if there are open file handles
+// See bug report about the game Abuse: https://github.com/dosbox-staging/dosbox-staging/issues/4123
```
johnnovak, line 262: Please add periods at the end of each sentence. I'm not (just) being pedantic 😄; if someone reformats this later into a single paragraphs, it will be hard to see where the sentences end. This happened to me previously when reformatting stuff.

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+
+	// Zero for unique_number uses system time
+	// Zero also means it creates the file and ensures it is unique
+	constexpr UINT unique_number = 0;
+
+	// Buffer size is Win32 defined MAX_PATH plus a null terminator
+	wchar_t temp_file[MAX_PATH + 1] = {};
+
+	// Create a temporary file inside the same directory as the file to be deleted
+	// Returns unique identifier on success and 0 on failure
```
johnnovak, line 276: Please add periods.

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+
+	// Above function creates the temp file so we replace it here with the flag MOVEFILE_REPLACE_EXISTING
+	if (MoveFileExW(path.c_str(), temp_file, MOVEFILE_REPLACE_EXISTING)) {
+		// We can immediately delete the temporary file.
+		// Any open handles can be still be read from or written to.
+		// A new file can be created with the original file name.
+		if (!DeleteFileW(temp_file)) {
+			LOG_WARNING("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
+			// Failed to delete the temp file but the original file was moved
+			// We should still return success so that the filesystem removes the original file from the cache
```
johnnovak, line 291: Please add periods.

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
 {
-	return remove(path.string().c_str()) == 0;
+	// Prefix only uses first 3 letters, abbreviate DosBox Staging
+	const wchar_t *prefix = L"DBS";
+
+	// Zero for unique_number uses system time
+	// Zero also means it creates the file and ensures it is unique
+	constexpr UINT unique_number = 0;
+
+	// Buffer size is Win32 defined MAX_PATH plus a null terminator
```
johnnovak, line 272: "is in Win32  defined as" or something?

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+	wchar_t temp_file[MAX_PATH + 1] = {};
+
+	// Create a temporary file inside the same directory as the file to be deleted
+	// Returns unique identifier on success and 0 on failure
+	if (GetTempFileNameW(path.parent_path().c_str(), prefix, unique_number, temp_file) == 0) {
+		LOG_WARNING("FS: Failed to create temp file. Deleting '%s' directly.", path.string().c_str());
```
johnnovak, line 278: Should rather be a `LOG_ERR`, no?

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+		// We failed to create a temp file but we should still try to delete the original file
+		return DeleteFileW(path.c_str());
+	}
+
+	// Above function creates the temp file so we replace it here with the flag MOVEFILE_REPLACE_EXISTING
+	if (MoveFileExW(path.c_str(), temp_file, MOVEFILE_REPLACE_EXISTING)) {
+		// We can immediately delete the temporary file.
+		// Any open handles can be still be read from or written to.
+		// A new file can be created with the original file name.
```
johnnovak, line 287: Please add periods.

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+		if (!DeleteFileW(temp_file)) {
+			LOG_WARNING("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
```
johnnovak, line 289: `LOG_ERR`?

johnnovak (approved):
Looks very good @weirddan455, much cleaner than the previous attempts 👍🏻 Thanks for your perseverance on this! 🚀 

I think your nice asm test case is very much worth capturing. I've created a new "DOS filesystem tests" page, would you be able to add your asm source to it? Just as text now will suffice, then we'll rethink the approach later if the page grows too big. 

https://github.com/dosbox-staging/dosbox-staging/wiki/DOS-filesystem-tests

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
 
+// Includes a Windows specific hack
+// We move the file to be deleted to a temp file before delete
+// This allows a file with the same name to be created even if there are open file handles
+// See bug report about the game Abuse: https://github.com/dosbox-staging/dosbox-staging/issues/4123
 bool delete_native_file(const std_fs::path& path)
 {
-	return remove(path.string().c_str()) == 0;
+	// Prefix only uses first 3 letters, abbreviate DosBox Staging
+	const wchar_t *prefix = L"DBS";
```
FeralChild64, line 266: How about something more descriptive? Like "DBS-deleted-"?

src/misc/fs_utils_win32.cpp
```diff
@@ -256,4 +256,49 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
+		if (!DeleteFileW(temp_file)) {
+			LOG_WARNING("FS: Failed to delete temporary file: '%s'", std_fs::path(temp_file).string().c_str());
+			// Failed to delete the temp file but the original file was moved
+			// We should still return success so that the filesystem removes the original file from the cache
+		}
+		return true;
+	}
+
+	// We failed to move the file. We need to delete both the temp file and the original file.
+	LOG_WARNING("FS: Failed to move '%s' to temp file '%s' before delete.", path.string().c_str(), std_fs::path(temp_file).string().c_str());
```
FeralChild64, line 297: Please use the` format-commit.sh` script. This line is well over 80 characters long.

FeralChild64:
Just some minor comments from me.

BTW, have you tested deleting two separate directories in one go, where these directories contain files of the same names?

src/misc/fs_utils_win32.cpp
weirddan455: From the Windows documentation, it will only use the first three characters:  https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettempfilenamew

>  The null-terminated prefix string. The function uses up to the first three characters of this string as the prefix of the file name. This string must consist of characters in the OEM-defined character set.

It's not really user visible anyway as the file gets deleted immediately after the move.

weirddan455:
>  BTW, have you tested deleting two separate directories in one go, where these directories contain files of the same names?

I don't know what you mean here.  This function is only used for deleting files, not directories.

FeralChild64:
Then it’s OK - I was afraid that this might disrupt directory tree deletion by, for example, Norton Commander, or the DELTREE command.

src/misc/fs_utils_win32.cpp
weirddan455: Ran the script.  I don't love how it shortened some of the lines but I'm not picky about it.  The line you commented is better now but it also likes to chop my comments short.

src/misc/fs_utils_win32.cpp
weirddan455: Yeah, you mentioned that before.  Sorry I forgot.  Added periods everywhere.

src/misc/fs_utils_win32.cpp
weirddan455: I think I was trying to say `MAX_PATH` is a Windows specific macro.  I just re-wrote the comment to be clearer and removed the Win32 part.

src/misc/fs_utils_win32.cpp
weirddan455: Sure, that makes sense.  Changed to `LOG_ERR`

src/misc/fs_utils_win32.cpp
weirddan455: Changed here and the 2 other warnings at the bottom.

src/misc/fs_utils_win32.cpp
johnnovak: Yeah I think it's fine. You could make it even uglier and more obvious this is some temp file by using a `$DB` prefix or something, but honestly, I don't think the name matters much.

weirddan455: It gives a .tmp extension as well to hopefully make it obvious it's a temp file.  It looks like DBS47BC.tmp where the last 4 digits are hex derived from system time.

johnnovak: Yeah, maybe do the `$DB` thing. That makes it even more obvious it's a temp file.

src/misc/fs_utils_win32.cpp
johnnovak: > Yeah, you mentioned that before. Sorry I forgot. Added periods everywhere.

Thanks man 🙇🏻‍♂️  😄

johnnovak:
Just a reminder @weirddan455; I think you're nice test case is worth capturing:

> I think your nice asm test case is very much worth capturing. I've created a new "DOS filesystem tests" page, would you be able to add your asm source to it? Just as text now will suffice, then we'll rethink the approach later if the page grows too big.
> 
> https://github.com/dosbox-staging/dosbox-staging/wiki/DOS-filesystem-tests

FeralChild64 (approved):
Played with file operations a little on Windows, I didn't notice any regressions.

src/misc/fs_utils_win32.cpp
weirddan455: Changed to `$DB`

weirddan455:
Updated the wiki here with my test case: https://github.com/dosbox-staging/dosbox-staging/wiki/DOS-filesystem-tests

>  Played with file operations a little on Windows, I didn't notice any regressions.

Thanks for testing @FeralChild64 

I think this is ready for merge now unless anyone has anything else.

weirddan455:
>  Also, this is backport-worthy material, wouldn't you agree?

Yeah, it's probably good to backport since it fixes a regression.

src/misc/fs_utils_win32.cpp
```diff
@@ -256,9 +256,50 @@ void set_dos_file_time(const NativeFileHandle handle, const uint16_t date,
...
 bool delete_native_file(const std_fs::path& path)
 {
-	return remove(path.string().c_str()) == 0;
+	// Prefix of the temp file. Only uses 3 characters.
+	// $ as convention to designate temp file followed by DB for DosBox.
+	const wchar_t *prefix = L"$DB";
+
+	// Zero for unique_number uses system time.
+	// Zero also means it creates the file and ensures it is unique.
+	constexpr UINT unique_number = 0;
```
johnnovak, line 272: minor: UniqueNumber for constexpr

johnnovak:
> Updated the wiki here with my test case: https://github.com/dosbox-staging/dosbox-staging/wiki/DOS-filesystem-tests

Thanks for that, it's looking very good 💯  

> I think this is ready for merge now unless anyone has anything else.

Just one more extremely petty comment 😅, but otherwise feel free to merge!

src/misc/fs_utils_win32.cpp
weirddan455: Fixed

Issue #4004, opened 2024-10-23
Rework CPI (screen font) file loading 


FeralChild64:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

None

### Is your feature request related to a problem? Please describe.

The CPI (code pages - DOS screen fonts) file loading needs a rework:
- the legacy C code should be rewritten in a modern C++; current code is hard to read/analyze
- CPX file format supports should be 🔥 killed with fire 🔥 - it's known to be buggy and no one (including me) seems to be interested in fixing it; if the user tries to load the CPX file, he should be instructed to decompress it using the freely available UPX tool, thus effectively converting it to a CPI format
- if feasible (read: not too much work), the CPI loading code should handle DR-DOS and Windows NT formats - some documentation is available

### Add any other context or screenshots about the feature request here.

CPI format documentation:
- https://www.seasip.info/DOS/CPI/cpi.html
- http://kbd-project.org/docs/font-formats/font-formats-3.html

Also attached as PDFs, in case the pages disappear:
[CPI format description by John Elliott.pdf](https://github.com/user-attachments/files/17497738/CPI.format.description.by.John.Elliott.pdf)
[CPI format description from KBD project.pdf](https://github.com/user-attachments/files/17497750/CPI.format.description.from.KBD.project.pdf)

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Another worthy entry of the kill with fire series 😎🤘🏻🔥

PR #7, opened 2019-10-01
Modernize MSVC build and add it to CI

dreamer:
See changes in INSTALL file for usage instructions.

[vcpkg](https://github.com/microsoft/vcpkg) solves a lot of headaches around developing open source project using Visual Studio. It's almost as good Linux package managers ;)

I tested it on Windows 10 (using VS Community Edition, which is free-to-use) and on CI, which runs Windows Server 2019 (and VS Enterprise Edition).

New solution and project files were needed, because VS does not recognize old format any more - it automatically converted files to the new format. Some tiny adjustments were needed to remove linker warnings (only 1 remains, and I think it might be false-positive).

The build reports only 17 warnings (using `/W3`) - I tried changing it to report more, but `/Wall` resulted in flood of trivial warnings and even`/W4` reported thousands of unhelpful ones; warnings generated by `/W3` seem meaningful.

One unfortunate problem not handled by vcpkg is include path of SDL libraries, normally they should be `#include <SDL.h>`, but vcpkg does not add `SDL` subdirectory to the includes list, so in VS/MSBuild they need to be `#include <SDL/SDL.h>` - this problem would not exist if DOSBox used CMake, but I can't address this ATM. For now, I included fake sdl headers in `src/platform/visualc/` redirecting to correct paths - those are used only by MSVC builds, so rest of the code doesn't need to be touched.

PR #5, opened 2019-09-26
Enforce limit on issues found in static analysis

dreamer:
Implements new script (count-bugs.py) for peeking inside clang static
analyzer's report and print just a summary.

If number of detected bugs goes beyond the limit, script will return
with error code 1, thus failing the CI run.  The upper limit is set to
113, which is current result of static analysis in our CI environment
(local run is likely to indicate different number); upper limit will
be updated in time, as issues get fixed or new compiler (detecting more
bugs) will be introduced.

This commit includes also slight modifictaions to count-warnings.py
script, to keep the both scripts outputting in similar format.

PR #4154, opened 2025-01-11
New CPI (DOS screen font) loader

FeralChild64:
# Description

What should I say - the CPI loading was rewritten from scratch. Also, when testing various CPI files I have discovered one missing file with Unicode definitions (most likely I have created it but forgot to add) - this is now fixed. For user visible changes see the proposed release notes.

FreeDOS CPX support was killed with fire 🔥🔥🔥. Rationale:
- it wasn't working properly nevertheless (could not load `EGA18.CPX` from the FreeDOS package)
- it is relatively easy for the user to convert the files to CPI format - just get UPX, execute `upx -d <filename>`, done
- to use the highly compressed CPX file we either need to execute the x86 code inside (not the safest method, if you ask me), or decompress it by ourselves, most likely by including half of the UPX in our code
- the format is slow to decompress - unless you are emulating a very fast machine
- all the FreeDOS CPX files are (in the uncompressed format) now included in the resources

BTW, if you want to see 0.82.0 crashing on loading CPI files, do - for example - one of the following:
- `keyb us 437 EGA.ICE` - where `EGA.ICE` comes from MS-DOS 6.0
- `keyb us 437 EGA.CPI` - where `EGA.CPI` comes from Windows 2000
- `keyb ua 866 DISPLAY.CPI` - where `DISPLAY.CPI` comes from PTS-DOS 2000

Each of these crashes have slightly different origin - but the common flaw is that _DOSBox_ blindly trusts the offsets inside the file, without checking if everything is within a valid memory range.
 
#### Limitations

1. Few warning messages displayed in the logs (list of code pages in the custom CPI file,  information about not all the font sizes being available) would be better displayed in the `keyb` command output, but it would require pumping more information through the keyboard layout support code; I don't want to do this now - this PR is already quite large and keyboard layout handling should also be improved/bugfixed.

2. It is possible the code won't be able to read some unusual/exotic/incompatible CPI files - a web page with the best format description (https://www.seasip.info/DOS/CPI/cpi.html) claims:

> At least one pathological CPI file is known to exist where values above 64k are stored as segment:offset rather than a 32-bit pointer (eg: 0x1000abcd rather than 0x0001abcd)

I don't have this file - I'm not sure if it is going to work (after consulting with John Elliott I implemented offset adaptation strategy which has a chance to work), but at least it should not crash the emulator.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4004

# Release notes

Fixed the CPI (DOS screen font) loader, so that it does not crash anymore on some CPI files.
CPI files can now be loaded from mounted disk images (not only from mounted directories).
Added support for _DR-DOS_ and _Windows NT_ CPI file formats. Most _MS-DOS_ style CPI files over 64 KB in size should work now.
Removed direct support for the _FreeDOS_ CPX files; they can still be used if they are uncompressed first with the 3rd party _UPX_ tool.
Some problematic code pages (like 60258, which replaces a regular ASCII character uppercase I with a national variant) are now patched when loading.

# Manual testing

To test loading screen fonts from various CPI files, you can use these test scripts:

[TESTBATS.tar.gz](https://github.com/user-attachments/files/18389723/TESTBATS.tar.gz)

Files used for testing:
- AST DOS 5.0 - `EGA-EE.CPI`, `EGA-NW.CPI`, `EGA-SE.CPI`, `LCD.CPI`
- DR DOS 7.x - `EGA.CPI`
- MS-DOS 6.0 - `EGA.CPI`, `EGA.ICE`
- MS-DOS 6.22 - `EGA.CPI`, `EGA2.CPI`, `EGA3.CPI`, `ISO.CPI`, `LCD.CPI`
- PC-DOS 2000 - `865.CPI`, `912.CPI`, `915.CPI`, `EGA.CPI`, `EGAX.CPI`, `ISO.CPI`
- PTS-DOS 2000 - `DISPLAY.CPI`
- Windows 2000 - `EGA.CPI`
- Dominus - `DOM437.CPI`, `DOM850.CPI`; get them from https://github.com/studio8502/Dominus
- John Elliot (these are DRDOS format fonts) - `GSANS/GSANS.CPI`, `GSANS/GEM.CPI`, `TSANS/TSANS.CPI`, `TSANS/GEM.CPI`, `PCW/PCW.CPI`, `PCW/GEM.CPI`; get them from http://www.seasip.info/DOS/CPI/
- Kostis - `ISOLATIN.CPI`,` ISOOTHER.CPI` get them from http://www.kostis.net/freeware/index-en.htm

If you just want good test coverage:
- test at least one DR-DOS format font (you can recognize them by `DRFONT` string starting from the 2nd byte of the file)
- test at least one Windows NT format font (string is `FONT.NT`)
- test at least one MS-DOS format font (string is `FONT`, without any suffixes)
- test `EGA.ICE` from MS-DOS 6.0 and `DISPLAY.CPI` from PTS-DOS (for PTS-DOS check code page 866); these are unusual - both are MS-DOS format, both are over 64 KB size, but they use different offset encodings inside the file

Some additional tests:
- try to load code page not existing in the file, like `keyb us 456 <filename` - error message should be displayed and the log output should list all the code pages available in the file
- try to load any code page from the FreeDOS UPX file (for example from here: https://gitlab.com/FreeDOS/base/cpidos/-/tree/master/CPI) - error message should be displayed telling what to do with CPX files
- try to load any screen font from the printer CPI file, for example `keyb us 437 5202.CPI`, where `5202.CPI` comes from MS-DOS 6.22 - an error message should state that this is a printer CPI file
- try to run the BAT scripts from the archive above with `language = pl` settings - accented national characters should turn into plain Latin one if the code page does not support them
- to test different size fonts: command `mode 80x25` select screen mode with 8x16 fonts, `mode 80x34` - with 8x14 fonts, `mode 80x43` - with 8x8 fonts

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:


johnnovak:
Related (and a rather minor, relatively unimportant thing), it occurred to me a while ago it would be cool if we had more authethic machine and display adapter brand specific default firmware fonts. E.g. this CPI packs contains the Tandy fonts:

https://archive.org/details/cpi_dos_font_collections

Then here we have an insane amount of original fimware font dumps:

https://int10h.org/oldschool-pc-fonts/fontlist/

E.g., this one is pretty cool:

https://int10h.org/oldschool-pc-fonts/fontlist/font?toshibasat_9x16

But maybe your CPI loader is not the answer but we should be "patching" the currently hardcoded firmware fonts instead. Anyway, just something I'm vaguely interested in 😄

FeralChild64:
Thanks for the archive.org link - I'm really curious who made this set and which tool was used. Regarding the file structure, these CPIs are just horrible, a lot of non-essential struct fields have improper values. And they too can be easily used to bring the existing main branch CPI loader down -` keyb de 850 <filename>` should work for most of them :D

Our CPI loader (contrary to the real DOS ones) actually replaces the ROM font, so it is a partial answer. What we need is to make it work also with Hercules and CGA (writing to `int10.rom.font_16`, etc. does not work if one of these cards is selected).

Regarding this firmware ROM dump set (if I see correctly, these are not straight dumps, plus many national characters are added - nice!) - using modern day bitmap fonts or TTF fonts is on my radar (that's why I was asking about the TTF engine you intend to use for the GUI), but I definitely want to make the original DOS format files work correctly first.

Another solution might be a `TTF output`, similarly to what DOSBox-X can do - and either rendering fonts to the native resolution, or to the resolution optimal for your CRT shaders to process. And it should be smart enough to fallback to normal rendering if the application or a game tries to load it's own font. Well, distant future if you ask me.

FeralChild64:
Updated my *.BAT file set to allow testing the CPIs from Archive.org.

johnnovak:
> Updated my *.BAT file set to allow testing the CPIs from Archive.org.

Nice 😄 Yeah TTF output like in X would be kinda cool and no too hard I think. If I end up using blend2d, that does its own font rendering internally, so you could use that to generate your glyph atlasses. Alternatively, you could use stb_truetype and fontstash from the NanoVG project, they're but ver small and have no further deps.

FeralChild64:
I'm actually reluctant to use the _stb_truetype_ for user-supplied fonts - the header starts with a disclaimer:

```C
//    NO SECURITY GUARANTEE -- DO NOT USE THIS ON UNTRUSTED FONT FILES
//
// This library does no range checking of the offsets found in the file,
// meaning an attacker can use it to read arbitrary memory.
```

Even ignoring the security part -  I would prefer not to see issues like _this favorite font of mine works with everything, just not with DOSBox Staging_.

BTW, there was actually some effort to port the TTF output from X (https://github.com/dosbox-staging/dosbox-staging/tree/ww/ttf-output, https://github.com/dosbox-staging/dosbox-staging/tree/ww/ttf-output-1), but eventually it wasn't merged.

johnnovak:
> I'm actually reluctant to use the _stb_truetype_ for user-supplied fonts - the header starts with a disclaimer:
> 
> ```c
> //    NO SECURITY GUARANTEE -- DO NOT USE THIS ON UNTRUSTED FONT FILES
> //
> // This library does no range checking of the offsets found in the file,
> // meaning an attacker can use it to read arbitrary memory.
> ```
> 
> Even ignoring the security part - I would prefer not to see issues like _this favorite font of mine works with everything, just not with DOSBox Staging_.
> 
> BTW, there used to be some effort to port the TTF output from X (https://github.com/dosbox-staging/dosbox-staging/tree/ww/ttf-output, https://github.com/dosbox-staging/dosbox-staging/tree/ww/ttf-output-1), but eventually it wasn't merged.

Yeah, with vcpkg we can easily use FreeType, true. The big selling point of stb_truetype is easy integration (it's header only and small). It makes a lot of sense for projects like NanoVG where you have control over the font file (one bundled and hardcoded UI font, done), but probably not the best for general TTF support. I'm quite sure it doesn't do ligatures and any complex stuff, just simple glyphs.

It's perfect for static linking in projects like my Gridmonger thingy that only uses Roboto and a custom icon font I created, so you don't need to bring in the huuuuge and difficult to compile "full fat" TTF engines like FreeType.

FeralChild64:
@johnnovak I have looked into the old code and I intend to try to write the TTF text renderer now. This is not a rocket science - and it might be a game-changer for some users.

Fortunately, @kcgen did the work of integrating the _FreeType_ into our build system, I was able to reuse it after some small adaptations (I think...).

The _renderer.cpp_ interface has changed, but looks reasonably easy to use. We don't need the quick&dirty Unicode handling of the old code, we have our own engine now. Plus, I would like different config settings - just two options, the TTF font file (or any other font format supported by FreeType) and a rendering mode. I also want an auto-fallback - if the user loads a custom CPI file (or a bundled one with a code page the selected font can't fully handle), or if the application redefined the character set, the TTF renderer should move aside ASAP.

So I'm afraid not much will be reused from the old implementation :D

johnnovak:
@FeralChild64 Yeah, it can be simple _if_ you keep it simple, just try to stay out of the woods in the initial implementation 😉 

E.g., check out the "How do I render a line of text in FreeType?" section here:
https://freetype.org/freetype2/docs/ft2faq.html

I'd only do Latin alphabet (plus variants) and other "single-glyph" languages first (Russian, Greek, etc.) and leave the Asian stuff for much later. Ideally, those should be done by a native speaker who understands what they're doing.

FeralChild64:
I'm not doing any DBCS, or right-to-left writing support.

Regarding Latin/Cyrillic/Greek/Armenian/Cherokee/Georgian - there is not much difference between them from the implementation point of view, it's just a matter of providing appropriate font. Arabic and Hebrew might work if one doesn't expect any special right-to-left support.

Yes, I already know the official FreeType documentation.

FeralChild64:
Unfortunately, the solution won't be that simple - it seems one has to be very careful when using FreeType, do micro-adjustments to counter rounding errors, etc. Otherwise, you can easily end up with TTF display with 1-pixel gaps between table drawing characters, like here:

Screenshot-gaps

My very initial font rendering code also produces this effect (depending on the concrete font and target resolution combination), I have to carefully work it around.

weirddan455:
The kerning looks off in that picture as well, or maybe kerning is not implemented at all.

https://freetype.org/freetype2/docs/glyphs/glyphs-4.html

>  Note that OpenType fonts (OTF) provide two distinct mechanisms for kerning, using the ‘kern’ and ‘GPOS’ tables, respectively, which are part of the OTF files. Older fonts only contain the former, while recent fonts contain both tables or even ‘GPOS’ data only. FreeType only supports kerning via the (rather simple) ‘kern’ table. For the interpretation of kerning data in the (highly sophisticated) ‘GPOS’ table you need a higher-level library like [ICU](http://icu-project.org/) or [HarfBuzz](http://harfbuzz.org/) since it can be context dependent (this is, the kerning may vary depending on the position within a text string, for example).

For what it's worth, stb_truetype supports GPOS.  I don't know if it does a better job than HarfBuzz at that.  FreeType + HarfBuzz is kind of the more "professional" stack.

Also, not sure how this is supposed to look in real DOS with a monospace terminal font.  Maybe kerning is not needed there but we may want it if you're planning on using this font loader for OSD or something in the future.

Just food for thought.  I've briefly used both FreeType and stb_truetype in some projects years back.  Haven't looked at all at Dosbox's font rendering code though.

FeralChild64:
TTF output for _DOSBox_ is basically a "terminal emulation", so kerning has no use here.

A font shaping engine (like _Harfbuzz_ or _Pango_) would be useful for rendering code pages which need combining marks (Lithuanian and many  _FreeDOS_-specific ones)  - but I don't want to introduce excessive amount of dependencies. And I would prefer not to use the _ICU_ for anything - it's huge and it constantly breaks the ABI.

johnnovak:
> Sidenote: HarfBuzz is another Glib infested library so maybe we don't want to use that 😆 

> Unfortunately, the solution won't be that simple - it seems one has to be very careful when using FreeType, do micro-adjustments to counter rounding errors, etc. Otherwise, you can easily end up with TTF display with 1-pixel gaps between table drawing characters,

Terminal emulators must be doing something that right, as you don't see the gap with box drawing chars in most, no? Having said that, it's font dependent; some fonts have gaps, some don't in the same terminal program 🤷🏻 So I wouldn't fuss too much over it, probably they'll need to use the right fonts.

> A font shaping engine (like _Harfbuzz_ or _Pango_) would be useful for rendering code pages which need combining marks (Lithuanian and many _FreeDOS_-specific ones) - but I don't want to introduce excessive amount of dependencies. And I would prefer not to use the _ICU_ for anything - it's huge and it constantly breaks the ABI.

Good call, no more glib projects in Staging, please 😆 FreeType is fine, and like you said I don't think you need kerning because you're rendering a font atlas into fixed-width cells, pretty much. I don't even think fixed-width TTF/OTF fonts have a kerning table (it wouldn't make a lot of sense).

FeralChild64:
Terminal emulators have to workaround such problems too, and their solutions are not always perfect either:

https://github.com/kovidgoyal/kitty/issues/16
https://github.com/alacritty/alacritty/issues/5809

johnnovak:
> Terminal emulators have to workaround such problems too, and their solutions are not always perfect either:
> 
> [kovidgoyal/kitty#16](https://github.com/kovidgoyal/kitty/issues/16) [alacritty/alacritty#5809](https://github.com/alacritty/alacritty/issues/5809)

Yeah, and perfection is the enemy of the good 😉

Doing some hacks/workarounds to "nudge" box chars to appear fine is completely fine in my book—that's the reality we need to deal with with those 42123 fonts out there. No one's gonna redesign their fonts just for some terminal emulators.

At the end of the day, some fonts will work without the workarounds, some will need the workarounds, and some will be resistant to them, and that's fine.

include/dos_inc.h
```diff
@@ -894,6 +894,15 @@ class DOS_SDA final : public MemStruct {
 };
 extern DOS_InfoBlock dos_infoblock;
 
+enum class ScreenFontOrigin {
```
johnnovak, line 897: Maybe `ScreenFontType`?

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+#include "support.h"
+#include "unicode.h"
+
+#include <algorithm>
+#include <array>
+#include <fstream>
+#include <optional>
+#include <set>
+#include <unordered_map>
+#include <vector>
```
johnnovak, line 40: I think we tend to order imports a bit differently. E.g. see this example:
https://github.com/dosbox-staging/dosbox-staging/blob/main/src/gui/shader_manager.cpp#L21-L39

So stdlib imports come first, 3rd party and Staging specific imports next.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+
+CHECK_NARROWING();
```
johnnovak, line 42: 👍🏻

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	// Used by MS-DOS and most clones
+	MsDos,
+	// Used by Windows NT, slight variation of MS-DOS format
+	WinNt,
+	// Used by DR-DOS, encodes screen fonts in much more efficient way
+	DrDos
+};
+
+// Supported file size limit, just for some extra safety.
+// - Almost all the CPI files in the wild are below 64 KB size (MS DOS does not
```
johnnovak, line 77: typo: MS-DOS

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+//   support anything bigger)
+// - 'DISPLAY.CPI' file from PTS-DOS 2000 is below 88 KB
+// - 'EGA.CPI' file from Windows 2000 (English) is slightly below 128 KB
+constexpr uint32_t MaxFileSize = 1024 * 1024;
```
johnnovak, line 81: `MaxFileSizeBytes` would be better (always include the unit).

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+
+constexpr uint8_t IdMsDosWinNt = 0xff;
+constexpr uint8_t IdDrDos      = 0x7f;
+
+static const std::string SignatureMsDos = "FONT   ";
+static const std::string SignatureWinNt = "FONT.NT";
+static const std::string SignatureDrDos = "DRFONT ";
```
johnnovak, line 88: Consider using namespaces instead of prefixes. Check out my Sound Canvas code where I started to use namespaces. IMO, it's way better than old C-style prefixing from the 80s...

```cpp
namespace Signature {
   MsDos = "FONT    ";
   ...
}```

File: src/dos/dos_code_page.cpp [hunk 7]
```diff
@@ -0,0 +1,2223 @@
...
+
+static const std::string SignatureUpx = "UPX!";
+
+static const std::string DeviceEga = "EGA";
+static const std::string DeviceLcd = "LCD";
+static const std::set<std::string> KnownPrinterDevices =
+        {"4201", "4208", "5202", "1050", "EPS", "PPDS"};
```
johnnovak, line 95: Namespaces?

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+
+	std::vector<uint8_t> font_8x16 = {};
+	std::vector<uint8_t> font_8x14 = {};
+	std::vector<uint8_t> font_8x8  = {};
+};
+
+using ScreenFonts = std::unordered_map<uint16_t, ScreenFont>;
+
+// ***************************************************************************
+// Persistend data storage
```
johnnovak, line 124: typo: Persistent

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+struct FontFileHeader {
+	static constexpr auto StructName = "FontFileHeader";
+
+	uint32_t struct_offset = 0;
+
+	uint8_t id_byte       = 0;
+	std::string id_string = {};
+	uint16_t num_pointers = 0;
+	uint8_t pointer_type  = 0;
+	uint32_t fih_offset   = 0;
```
johnnovak, line 142: What's `fih`? Probably an explanatory comment is in order. Better yet, consider adding explanatory comments to all struct members.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+
+// DR-DOS format specific structure
+struct FontFileExtendedHeader {
+	static constexpr auto StructName = "FontFileExtendedHeader";
+
+	uint32_t struct_offset = 0;
+
+	uint8_t fonts_per_codepage       = 0;
+	std::vector<uint8_t> cell_size   = {};
+	std::vector<uint32_t> dfd_offset = {};
```
johnnovak, line 153: Explanatory comments? Or at least external link to a document that describes the struct?

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	static constexpr auto StructName = "CodePageEntryHeader";
+
+	uint32_t struct_offset = 0;
+
+	uint16_t cpeh_size        = 0;
+	uint32_t next_cpeh_offset = 0;
+	uint16_t device_type      = 0;
+	std::string device_name   = {};
+	uint16_t code_page        = 0;
+	uint32_t cpih_offset      = 0;
```
johnnovak, line 174: Comments or links to reference documents, etc. (I'd prefer comments)

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	static constexpr auto StructName = "ScreenFontHeader";
+
+	uint32_t struct_offset = 0;
+
+	uint8_t height     = 0;
+	uint8_t width      = 0;
+	uint8_t y_aspect   = 0;
+	uint8_t x_aspect   = 0;
+	uint16_t num_chars = 0;
+};
```
johnnovak, line 197: For all structs: comments or links to reference documents, etc. (I'd prefer comments)

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+
+#if 0
```
johnnovak, line 199: Is this for debugging? Do it via a `DEBUG_DOS_CODE_PAGE` define or something instead of this.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	size_t GetContentSize() const;
+	uint32_t GetCurrentOffset() const;
+	void SetCurrentOffset(const uint32_t new_offset);
+	void SkipBytes(const uint32_t num_bytes);
+
+	bool HasUpxSignature() const;
+
+	bool IsCopyrightString() const;
+
+	std::vector<uint8_t> ReadBlob(const uint32_t length);
```
johnnovak, line 245: `num_bytes` is better than the vague `length`. Prefer `int` or `int64`; only use unsigned integer types of specific widths when dealing with binary formats or register data.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	std::optional<FontFileExtendedHeader> ReadFontFileExtendedHeader();
+	std::optional<FontInfoHeader> ReadFontInfoHeader();
+	std::optional<CodePageEntryHeader> ReadCodePageEntryHeader();
+	std::optional<CodePageInfoHeader> ReadCodePageInfoHeader();
+	std::optional<ScreenFontHeader> ReadScreenFontHeader();
+	std::optional<CharacterIndexTable> ReadCharacterIndexTable(const uint32_t num_chars);
+
+	virtual void WarnDataOutOfBounds(const uint32_t offset_start,
+	                                 const uint32_t offset_end,
+	                                 const std::string& struct_name = {}) = 0;
```
johnnovak, line 257: `int` or `int64` params it at all possible.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	virtual void WarnStructSizeInvalid(const uint32_t offset,
+	                                   const std::string& struct_name,
+	                                   const uint32_t size) = 0;
```
johnnovak, line 260: `int` or `int64` params it at all possible.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	// Strip leading device name spaces
+	while (result.device_name.ends_with(" ")) {
+		result.device_name.pop_back();
+	}
+
+	// According to https://www.seasip.info/DOS/CPI/cpi.html:
+	// - the structure is normally 28 bytes
+	// - many tools simply ignore the size
+	// - sometimes (in old files) the structure is 26 bytes due to offset
+	//   being stored on 2 bytes
```
johnnovak, line 489: The comment is good, but extract the 26 and 28 magic numbers to constants with sensible names.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+
+// ***************************************************************************
+// CPI file parser - declaration
+// ***************************************************************************
+
+struct ParserResult {
+	// A general return code; 'false' if the file can't be parsed at all
+	bool status_ok = true;
+
+	// If the reson of parsing failure was an unsupported UPX compression
```
johnnovak, line 570: typo: reason

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	static bool IsScreenCodePage(const CodePageEntryHeader& header);
+
+	void StoreFontType(const CodePageEntryHeader& header, ParserResult& result);
+
+	// Concrete screen font extraction
+
+	std::optional<ScreenFont> GetScreenFontFormat1(const uint16_t num_fonts,
+	                                               const uint16_t total_size);
+	std::optional<ScreenFont> GetScreenFontFormat2(const uint16_t num_fonts,
+	                                               const uint16_t total_size);
```
johnnovak, line 660: Can't just use `int`s?

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	                                const std::string& struct_name,
+	                                const uint8_t width);
+	void WarnScreenFontDuplicateSize(const uint32_t struct_offset,
+	                                 const std::string& struct_name,
+	                                 const uint8_t height);
+	void WarnScreenFontWithoutSizes(const uint32_t struct_offset,
+	                                const std::string& struct_name);
+	void WarnScreenFontWithoutUsefulSizes(const uint32_t font_offset);
+	void WarnScreenFontSizeWithoutCharacters(const uint32_t struct_offset,
+	                                         const std::string& struct_name);
```
johnnovak, line 749: Blank lines between methods, please. This just looks like a huge block of text, hard to parse where one ends and the other starts.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+}
+
+void CpiParser::WarnUnknownScreenFontDevice(const uint32_t struct_offset,
+                                            const std::string& struct_name,
+                                            const std::string& device_name)
+{
+	const auto message = format_str("unknown screen font device '%s'",
+	                                device_name.c_str());
+	PrintWarning(struct_offset, struct_name, message);
+}
```
johnnovak, line 1007: Do you need all these methods? Can't you just construct the errors inline? IMO some minor duplication would be preferable to having so many different warning/error logging methods.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+			// At least one CPI file ('4208.CPI' from MS-DOS 6.22)
+			// is known to fail this check just before the last
+			// font (the offset is 0).
+			// Since this is a printer font (not useful right now),
+			// there was no in-depth investigation.
+			break;
+		}
+		SetCurrentOffset(next_offset);
+		++idx;
+	}
```
johnnovak, line 1591: Can you make this method less huge? Break it up into logical sub-methods?

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+		}
+		code_pages_str += std::to_string(item);
+	}
+	LOG_WARNING("LOCALE: The file '%s' does not contain code page %d; it contains %s",
+	            file_name.c_str(),
+	            code_page,
+	            code_pages_str.c_str());
+
+	return KeyboardLayoutResult::NoCodePageInCpiFile;
+}
```
johnnovak, line 1942: Another huge method, a good candidate for splitting things up into more logical chunks.

src/dos/dos_code_page.h
```diff
@@ -0,0 +1,39 @@
...
+bool DOS_CanLoadScreenFonts();
+
+// Loads and sets the screen font from the DOS file or (if no file name given)
+// from one of the bundled CPI files.
+// Only to be called from within the keyboard layout handling code!
+KeyboardLayoutResult DOS_LoadScreenFont(const uint16_t code_page,
+                                        const std::string& file_name = {});
+
+// Restores the standard ROM screen font.
+// Only to be called from within the keyboard layout handling code!
```
johnnovak, line 36: Nice module-level interface doco 🎖️

src/dos/program_keyb.cpp
```diff
@@ -362,7 +376,7 @@ void KEYB::AddMessages()
 	        "Parameters:\n"
 	        "  [color=light-cyan]LAYOUT[reset]    keyboard layout code\n"
 	        "  [color=white]CODEPAGE[reset]  code page number, e.g. [color=white]437[reset] or [color=white]850[reset]\n"
-	        "  [color=white]CPIFILE[reset]   screen font file, in CPI or CPX format\n"
+	        "  [color=white]CPIFILE[reset]   screen font file, in a CPI format\n"
```
johnnovak, line 367: typo: "in CPI format"

src/dos/program_keyb.cpp
```diff
@@ -371,6 +385,9 @@ void KEYB::AddMessages()
 	        "    and code page.\n"
 	        "  - The [color=white]CPIFILE[reset], if specified, must contain the screen font for the given\n"
 	        "    [color=white]CODEPAGE[reset].\n"
+	        "  - The MS-DOS, DR-DOS, and Windows NT formats of the CPI files are supported\n"
```
johnnovak, line 376: Drop the starting "The" from the start of the sentence.

src/dos/dos_code_page.cpp
```diff
@@ -44,6 +45,17 @@ CHECK_NARROWING();
 // - https://www.seasip.info/DOS/CPI/cpi.html
 // - http://kbd-project.org/docs/font-formats/font-formats-3.html
 
+extern void DOS_UpdateCurrentProgramName();
+
+static void notify_code_page_changed()
+{
+	// Re-create various information to match new code page
```
johnnovak, line 52: minor type: it's usually one word, "recreate".

src/dos/dos_code_page.cpp
```diff
@@ -2046,6 +2050,91 @@ static std::optional<uint16_t> find_ega_font_storage_index(const uint16_t code_p
...
+static void patch_font_dotted_i(ScreenFont& font)
+{
+	// Incomplete fonts should not be allowed in storage
+	assert(font.font_8x16.size() == ScreenFont::FullSize_8x16);
+	assert(font.font_8x14.size() == ScreenFont::FullSize_8x14);
+	assert(font.font_8x8.size()  == ScreenFont::FullSize_8x8);
+
+	// Codes of characters to swap
+	constexpr uint8_t CodePoint1 = 0x49;
+	constexpr uint8_t CodePoint2 = 0xf2;
```
johnnovak, line 2062: Maybe explain in a comment which characters are these.

src/dos/dos_code_page.cpp
FeralChild64: Fixed.

src/dos/dos_code_page.cpp
FeralChild64: Fixed.

src/dos/dos_code_page.cpp
FeralChild64: OK, changed.

src/dos/dos_code_page.cpp
FeralChild64: Fixed.

include/dos_inc.h
FeralChild64: OK, changed.

src/dos/dos_code_page.cpp
FeralChild64: OK, added comments.

src/dos/program_keyb.cpp
FeralChild64: OK, changed.

src/dos/program_keyb.cpp
FeralChild64: OK, changed.

src/dos/dos_code_page.cpp
FeralChild64: Done.

src/dos/dos_code_page.cpp
FeralChild64: OK, I've put all these data into namespaces.

src/dos/dos_code_page.cpp
FeralChild64: Fixed.

src/dos/dos_code_page.cpp
FeralChild64: Ok, changed.

src/dos/dos_code_page.cpp
FeralChild64: Added comments, everywhere.

src/dos/dos_code_page.cpp
FeralChild64: Added comments, everywhere. Links to the format description can be found at the top of the file.

src/dos/dos_code_page.cpp
FeralChild64: Added comments, everywhere. Links to the format description can be found at the top of the file.

src/dos/dos_code_page.cpp
FeralChild64: Added comments, everywhere. Links to the format description can be found at the top of the file.

src/dos/dos_code_page.cpp
FeralChild64: No, this is just information about printer code page definition, in case someone in the future needs to read the printer data (currently almost completely ignored) and the documentation is no longer available.

Added a comment to prevent confusion.

johnnovak: > Added a comment to prevent confusion.

Yeah, thanks. We should always do that because disabled code without comments like that just confuses people.

src/dos/dos_code_page.cpp
```diff
@@ -0,0 +1,2223 @@
...
+	                                   const std::string& struct_name,
+	                                   const uint32_t size) = 0;
+
+private:
+	bool IsOffsetBeyondContent() const;
+
+	uint8_t InternalReadByte();
+	uint16_t InternalReadShort();
+	uint32_t InternalReadLong();
+	std::string InternalReadString(const uint32_t length);
```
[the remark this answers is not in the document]
johnnovak: Yeah, I was sure, but it just seemed like noise to me.

src/dos/dos_code_page.cpp
FeralChild64: OK, done.

src/dos/dos_code_page.cpp
FeralChild64: OK, changed to generate part of the error strings inline (struct name and offset are still formed in one certain place to ensure consistency).

Exception: `WarnDataOutOfBounds`, as it is used in many places.

src/dos/dos_code_page.cpp
FeralChild64: Extracted reading the file via MS-DOS API to external function; this largely reduced the function size.

src/dos/dos_code_page.cpp
FeralChild64: OK, extracted some parts to separate methods.

src/dos/dos_code_page.cpp
FeralChild64: Parameter renamed.

Regarding the type - we are dealing with a binary input here, almost all the unsigned integers of specific width are taken from the binary data format and the remaining ones are there to prevent expressions like `value1 += value2` from producing compiler warnings and to allow `std::min(value1, value2)` expressions to compile.

src/dos/dos_code_page.cpp
FeralChild64: Offsets in the CPI files are unsigned 32 bit integers (in rare cases these are pairs of 16 bit _segment:offset_ values), I would prefer to use the same type as in the file format description.

src/dos/dos_code_page.cpp
FeralChild64: Offsets in the CPI files are unsigned 32 bit integers (in rare cases these are pairs of 16 bit _segment:offset_ values), I would prefer to use the same type as in the file format description.

src/dos/dos_code_page.cpp
FeralChild64: These values are taken straight from the binary data format, I would prefer to keep the original integer width.

src/dos/dos_code_page.cpp
johnnovak: Sure, I think this qualifies for the dealing with binary formsts case.

src/dos/dos_code_page.cpp
johnnovak: Yeah that makes sense. We're dealing with a binary format so that's ok, I just wasn't 100% sure.

FeralChild64:
Regarding TTF output support - here is an uncorrected test pattern rendered using the `LiberationMono-Regular.ttf` font (code page 437):

Raw

And here is a corrected image (code page 991 this time):
- table drawing characters (rows #6 and #7) now properly connect to each other, without gaps  between rows caused by rounding errors and anti-aliasing on borders; similar treatment was applied to block characters in line #7
- custom ligature rendering code was able to produce a `zł` symbol (this is the PLN currency symbol, nowadays used as two letters, decades ago also commonly used in the form of a ligature ) in row #5 - despite it is not provided by the font and not even present in the Unicode standard
- GUI elements in row #1 are aspect-corrected (not that problematic in 80x25 text mode, but switching to 80x50 text mode would produce awfully squashed symbols) and centered (centering is not a problem with this particular font, but with some fonts it is)
- inverse circle characters in row #1 (U+25D8, U+25D9) are now more DOS-like - I have simply used a neighbor glyph and reversed the colors

Corrected

It's still not perfect - in some cases the borders of the shade blocks (line #6) are darker than they should be due to anti-aliasing; for now I don't have a solution for this

johnnovak:
> Regarding TTF output support - here is an uncorrected test pattern rendered using the `LiberationMono-Regular.ttf` font (code page 437):
> 
> Raw
> 
> And here is a corrected image (code page 991 this time):
> 
>     * table drawing characters (rows [Update package repositories before build #6](https://github.com/dosbox-staging/dosbox-staging/pull/6) and [Modernize MSVC build and add it to CI #7](https://github.com/dosbox-staging/dosbox-staging/pull/7)) now properly connect to each other, without gaps  between rows caused by rounding errors and anti-aliasing on borders; similar treatment was applied to block characters in line [Modernize MSVC build and add it to CI #7](https://github.com/dosbox-staging/dosbox-staging/pull/7)
> 
>     * custom ligature rendering code was able to produce a `zł` symbol (this is the PLN currency symbol, nowadays used as two letters, decades ago also commonly used in the form of a ligature ) in row [Enforce limit on issues found in static analysis #5](https://github.com/dosbox-staging/dosbox-staging/pull/5) - despite it is not provided by the font and not even present in the Unicode standard
> 
>     * GUI elements in row [Create initial Workflows setup #1](https://github.com/dosbox-staging/dosbox-staging/pull/1) are aspect-corrected (not that problematic in 80x25 text mode, but switching to 80x50 text mode would produce awfully squashed symbols) and centered (centering is not a problem with this particular font, but with some fonts it is)
> 
>     * inverse circle characters in row [Create initial Workflows setup #1](https://github.com/dosbox-staging/dosbox-staging/pull/1) (U+25D8, U+25D9) are now more DOS-like - I have simply used a neighbor glyph and reversed the colors
> 
> 
> Corrected
> 
> It's still not perfect - in some cases the borders of the shade blocks (line #6) are darker than they should be due to anti-aliasing; for now I don't have a solution for this

Is that the 1:1 unscaled output via FreeType? Antialiasing looks a bit funky... Kinda soft, blurry, just weird. Wonder how it looks like via stb_truetype, or can you try some other FT render params to make it look better?

FeralChild64:
Shader works on top of what I render with FreeType, scaling it from 640x400 to 4:3 and stretching it. I don't have this part handled yet.

johnnovak:
> Shader works on top of what I render with FreeType, scaling it and stretching from 640x400 to some 4:3 resolution. I don't have this part handled yet.

That explains it. I was getting worried 😆

johnnovak (approved):
I'm approving this as we went thorough a few rounds of code reviews already. I assume you've tested it in detail @FeralChild64, plus I don't have much inclination to test i18n stuff like this in detail anyway 😄 (usually I only test that English has not regressed).

So feel free to merge, or ask some our test members for extra testing if you need more confidence on the changes.

Issue #4049, opened 2024-11-09
List of games that use DMA for audio and start glitching above ~50k cycles


johnnovak:
Some games that use DMA for audio exhibit severe glitching above about 50k cycles in 0.82.0. Many 3dfx games are similarly affected.

This is the result of the recent mixer improvements of @weirddan455, which does indeed reduce audio glitches and dropouts in the vast majority of titles, but DMA access can cause problems. So it's a sort of a regression in specific circumstances for a few games; it's a "two steps forward, one step back" kind of situation.

This ticket is just a collection of these problematic titles. The success criteria of the upcoming audio-related DMA handling improvements will be that these games run glitch-free, just as they did in 0.81.2 prior the the grand mixer refactor.

- **Archimedean Dynasty** (3dfx version; 80k cycles was fine in 0.81.1, but glitching starts above 60k in 0.82.0)
- **Blood** (GUS and SB with DMA above ~50k cycles for high-res SVGA modes, more details [here](https://github.com/dosbox-staging/dosbox-staging/issues/3993) )
- **Nethack — Falcon's Eye**

johnnovak:
@Python-Exoproject Please feel free to add to the list in the comments; that would help the testing effort.

weirddan455:
The Blood issue was fixed.  That was due to GUS being moved to the mixer thread which caused issues with DMA so I moved it back.

SoundBlaster is still on the main thread as it's always been.

As for the 3dfx game, are you sure it's due to the the mixer changes and not the recent Voodoo changes that also went into 0.82.0?

weirddan455:
Also, we should check if #3875 helps (or hurts) any of these.  I'm hesitant to make any SoundBlaster changes until that goes in as it'll cause merge conflicts.

johnnovak:
> The Blood issue was fixed. That was due to GUS being moved to the mixer thread which caused issues with DMA so I moved it back.

Great! I assume we'll need to retest it though if you're eventually moving the GUS DMA off the mixer thread again, so it makes sense to take note of it.

> SoundBlaster is still on the main thread as it's always been.

Yep 👍🏻 

> As for the 3dfx game, are you sure it's due to the the mixer changes and not the recent Voodoo changes that also went into 0.82.0?

Nope; I haven't tested anything, I'm just collecting the issues here based on the feedback I'm getting. That's a good point, it might be related to the increased threading for the Voodoo emulation.

[...]

PR #4165, opened 2025-01-31
Make SoundBlaster minimally blocking on other mixer channels

weirddan455:
# Description

The main goal of this PR is to stop OPL and MIDI from stuttering while SoundBlaster DAC/DMA are playing.  I also did away with the per-frame callbacks (which was mostly used by DAC but also sometimes DMA if the writes are small).

# Release notes

Less audio stuttering in games using SoundBlaster

# Manual testing

- Alone in the Dark (only SoundBlaster DAC game I know of)
- Tyrian
- Doom
- Duke3D

Tyrian had stuttering in the main menu and while pressing Esc to access the in-game menu.

Duke3D had even worse stuttering at the 3DRealms logo at game start.

Both of these are reproducible all of the time in a debug build and also in release build with high cycles.  It now "slows down" the audio rather than producing a crackling/stuttery mess.

I'll do some more testing later but if you know a good test-case let me know.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -2986,11 +2986,10 @@ static void generate_frames(const int frames_requested)
 	switch (sb.mode) {
 	case DspMode::None:
 	case DspMode::DmaPause:
-	case DspMode::DmaMasked: {
-		static std::vector<AudioFrame> empty_frames = {};
-		empty_frames.resize(frames_requested);
-		enqueue_frames(empty_frames);
-	} break;
+	case DspMode::DmaMasked:
+		sblaster->output_queue.Stop();
+		frames_added_this_tick += frames_requested;
+		break;
 
 	case DspMode::Dac:
 		// Frames get added synchronously inside dsp_do_command
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -3046,6 +3046,46 @@ static void per_tick_callback()
 	frames_added_this_tick -= total_frames;
 }
 
+void SBLASTER::AudioCallback(const int frames_requested)
+{
+	if (MIXER_FastForwardModeEnabled()) {
+		// Special case, normally only hit when using the fast-forward
+		// hotkey (Alt + F12). We need a very large buffer to compensate
+		// or it results in static.
+
+		// Mostly arbitrary but it works well in testing.
+		// The queue just needs to be large enough to hold the large
+		// frame requests we get in fast-forward mode. This value can be
+		// tweaked without much consequence if it ever becomes
+		// problematic.
+		constexpr float MaxExpectedFastForwardFactor = 100.0f;
+		output_queue.Resize(iceil(channel->GetFramesPerBlock() *
+		                                  MaxExpectedFastForwardFactor));
+	} else {
+		// Normal case, resize the queue to ensure we don't have high
+		// latency. Resize is a fast operation, only setting a variable
+		// for max capacity It does not drop frames or append zeros to
+		// the end of the underlying data structure.
+
+		// Size to 2x blocksize. The mixer callback will request 1x
+		// blocksize. This provides a good size to avoid over-runs and
+		// stalls.
+		output_queue.Resize(
+		        iceil(channel->GetFramesPerBlock() * 2.0f));
+	}
+
+	const auto frames_received = check_cast<int>(
+	        output_queue.BulkDequeue(mixer_buffer, frames_requested));
+
+	if (frames_received > 0) {
+			channel->AddAudioFrames(mixer_buffer);
+	}
+	// Fill any shortfall with silence
+	if (frames_received < frames_requested) {
+		channel->AddSilence();
+	}
+}
+
 static SbType determine_sb_type(const std::string& pref)
 {
 	if (pref == "gb") {
@@ -3349,14 +3389,10 @@ SBLASTER::SBLASTER(Section* conf)
 		channel_features.insert(ChannelFeature::Stereo);
 	}
 
-	constexpr bool Stereo      = true;
-	constexpr bool SignedData  = true;
-	constexpr bool NativeOrder = true;
-
 	const auto callback = std::bind(
-	        MIXER_PullFromQueueCallback<SBLASTER, AudioFrame, Stereo, SignedData, NativeOrder>,
-	        std::placeholders::_1,
-	        this);
+	        &SBLASTER::AudioCallback,
+	        this,
+	        std::placeholders::_1);
 
 	channel = MIXER_AddChannel(callback,
 	                           DefaultPlaybackRateHz,
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -283,23 +283,6 @@ static SbInfo sb = {};
 
 std::unique_ptr<SBLASTER> sblaster = {};
 
-class CallbackType {
-public:
-	void SetNone();
-	void SetPerTick();
-	void SetPerFrame();
-	~CallbackType()
-	{
-		SetNone();
-	}
-
-private:
-	enum class TimingType { None, PerTick, PerFrame };
-	TimingType timing_type = TimingType::None;
-};
-
-static CallbackType callback_type = {};
-
 // clang-format off
 
 // Number of bytes in input for commands (sb/sbpro)
@@ -825,20 +808,6 @@ static void dsp_dma_callback(const DmaChannel* chan, const DmaEvent event)
 			// ready for playback. This is when we set the callback running to
 			// play the data.
 			sblaster->MaybeWakeUp();
-
-			// If the DMA transfer is setup with a base count of fewer than
-			// three elements (which is four bytes given one 16-bit stereo frame
-			// held in an 8-bit DMA channel), then we know the software intends
-			// to overwrite the DMA content on the fly instead of pre-generating
-			// large chunks of DMA audio. In these cases we prefer the
-			// fine-grained per-frame callback. (The minus one is because DMA
-			// counts are in addition to one; so a base count of zero is one
-			// element).
-			constexpr auto MaxSingleFrameBaseCount = sizeof(int16_t) * 2 - 1;
-
-			(chan->base_count <= MaxSingleFrameBaseCount)
-			        ? callback_type.SetPerFrame()
-			        : callback_type.SetPerTick();
 		}
 		break;
 	default: assert(false); break;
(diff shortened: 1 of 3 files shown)


johnnovak:
@weirddan455 This can be considered a regression fix as well because your otherwise stellar mixer improvements introduced audio glitches on high cycles counts in certain games that were fine before.

I started documenting these games here. Hopefully your enhancement will fix these regressions as well:

- https://github.com/dosbox-staging/dosbox-staging/issues/4049

src/hardware/sblaster.cpp
```diff
@@ -1320,56 +1289,6 @@ static void flush_remaining_dma_transfer()
...
-	if (timing_type != TimingType::PerFrame) {
-
-		SetNone();
-
-		add_next_frame_callback();
-
-		timing_type = TimingType::PerFrame;
-	}
-}
-
```
johnnovak, line 1376: Oh man, I love when these emulator-specific complexities go away 💯 The closer the emulation code is to what's actually happening on the hardware, the better 🚀 You usually end up with complex emulator code when you fudge things.

src/hardware/sblaster.cpp
```diff
@@ -1843,15 +1762,27 @@ static void dsp_do_command()
...
+		// Stopping the queue here makes the mixer dequeue nonblocking.
+		// Direct DAC mode has no set sample rate and it varies quite frequently.
+		// Without this call, I was getting stutters in OPL music in Alone in the Dark.
+
+		// If there are reports of glitchy DAC sound, we may want to revisit this.
+		// The previous approaches of per-frame callbacks and audio stretching solved this by matching to the mixer's sample rate.
+		// The advantage here is that we guarantee to never block OPL or MIDI music channels.
+
+		// It's a bit of a trade-off but with my sample size of 1 game, this sounds fine.
+		// Direct DAC on the SoundBlaster is very rarely used.
```
johnnovak, line 1784: Nice one, and direct DAC games are pretty rare anyway. Maybe some obscure SB 1.0 era games use it that not many people care about, dunno... Great comment 👍🏻

src/hardware/sblaster.cpp
```diff
@@ -3124,29 +3047,6 @@ static void per_tick_callback()
...
-
-	if (!sblaster->channel->is_enabled) {
-		callback_type.SetNone();
-		return;
-	}
-
-	generate_frames(1);
-	add_next_frame_callback();
-}
-
```
johnnovak, line 3166: Good riddance 🥇

src/hardware/sblaster.cpp
```diff
@@ -2986,11 +2986,10 @@ static void generate_frames(const int frames_requested)
...
 	case DspMode::DmaPause:
-	case DspMode::DmaMasked: {
-		static std::vector<AudioFrame> empty_frames = {};
-		empty_frames.resize(frames_requested);
-		enqueue_frames(empty_frames);
-	} break;
+	case DspMode::DmaMasked:
+		sblaster->output_queue.Stop();
+		frames_added_this_tick += frames_requested;
+		break;
```
johnnovak, line 2992: Hmm okay, I guess make sure fast-forwarding still works OK.

src/hardware/sblaster.cpp
```diff
@@ -3046,6 +3046,46 @@ static void per_tick_callback()
...
+	if (MIXER_FastForwardModeEnabled()) {
+		// Special case, normally only hit when using the fast-forward
+		// hotkey (Alt + F12). We need a very large buffer to compensate
+		// or it results in static.
+
+		// Mostly arbitrary but it works well in testing.
+		// The queue just needs to be large enough to hold the large
+		// frame requests we get in fast-forward mode. This value can be
+		// tweaked without much consequence if it ever becomes
+		// problematic.
```
johnnovak, line 3063: 💯 comment for the next guy maintaining this

src/hardware/sblaster.cpp
```diff
@@ -3046,6 +3046,46 @@ static void per_tick_callback()
...
+		constexpr float MaxExpectedFastForwardFactor = 100.0f;
+		output_queue.Resize(iceil(channel->GetFramesPerBlock() *
+		                                  MaxExpectedFastForwardFactor));
+	} else {
+		// Normal case, resize the queue to ensure we don't have high
+		// latency. Resize is a fast operation, only setting a variable
+		// for max capacity It does not drop frames or append zeros to
```
johnnovak, line 3070: typo: period after capacity

src/hardware/sblaster.cpp
weirddan455: This entire function was mostly copy + pasted from the templated function in `mixer.h` btw.  I might figure a way to share common code once this PR is stabilized.  In the mean time, it's easier to make small changes with this function extracted.

src/hardware/sblaster.cpp
weirddan455: I did test fast-forward and it works :smile:

src/hardware/sblaster.cpp
johnnovak: I was about to say, it would be good to move this into some helper function and have the big explanatory comments in one place.

src/hardware/sblaster.cpp
```diff
@@ -829,20 +808,6 @@ static void dsp_dma_callback(const DmaChannel* chan, const DmaEvent event)
...
-			// three elements (which is four bytes given one 16-bit stereo frame
-			// held in an 8-bit DMA channel), then we know the software intends
-			// to overwrite the DMA content on the fly instead of pre-generating
-			// large chunks of DMA audio. In these cases we prefer the
-			// fine-grained per-frame callback. (The minus one is because DMA
-			// counts are in addition to one; so a base count of zero is one
-			// element).
-			constexpr auto MaxSingleFrameBaseCount = sizeof(int16_t) * 2 - 1;
-
-			(chan->base_count <= MaxSingleFrameBaseCount)
```
interloper98, line 843: Just saw this PR and noticed this is being snipped out. 

If we want to collapse the two approaches (per-frame vs per-tick) into one, then the key hinges on the base count size.

The base count is how large of a buffer the game is setting up, basically it's DMA chunk size, or it could also be thought of as the game's own audio latency. 

You can invert the size to get timing: DMA-base-count / (SB-channels * SB-freq)

For example, a base count of 90 when the SB16 is set to stereo and 45 KHz results in a 1ms DMA buffer. 

In this example, DOSBox's 1ms tick size would work out perfectly, as it revisits the SB every 1ms tick through the emulator.

Base counts that are even bigger also work fine with the 1ms tick. 

But some games and demos might use a smaller base count, like Triton's Crystal Dream.

I haven't checked this branch yet, but my guess is we have a regression with this demo. 

I remember testing lots of other demos that (now) work on the main branch, but that might also regress with this PR. I'll check them and upload a zip of them later today. 

My point here isn't that I want to save the two-prong approach (good riddance, like @johnnovak said), but I just want to be extra careful that we keep the SB accurate -- and the key (for DMA) is to use the base count to determine the revisit frequency. 

As for DAC, I suspect 1ms revisit like you have is fine. 

So this would only be for DMA.

interloper98:
@weirddan455 ,

Here's a round up of the test suite  I used on the per-frame PR: 

[soundblaster-submillisecond-dma-test-suite.zip](https://github.com/user-attachments/files/18627549/soundblaster-submillisecond-dma-test-suite.zip)

There are 50 demos that setup a smaller DMA size; I've tagged their directories with the 'dma' suffix.

They all use autotype, so only need to launch dosbox inside the directory and let it answer the audio card questions.

The main branch currently passes all of them.

src/hardware/sblaster.cpp
johnnovak: > I remember testing lots of other demos that (now) work on the main branch, but that might also regress with this PR. I'll check them and upload a zip of them later today.

Thanks for that, that will help a lot in testing this change 👍🏻 

> In this example, DOSBox's 1ms tick size would work out perfectly, as it revisits the SB every 1ms tick through the emulator.

I kinda get what you're trying to say, but I doubt this is true. There is zero guarantee our 1ms tick at the emulator level would "line up" perfectly with that supposedly 1ms long DOS-program-size DMA buffer. And even if it did, we should not rely on such assumptions.

> My point here isn't that I want to save the two-prong approach (good riddance, like @johnnovak said), but I just want to be extra careful that we keep the SB accurate -- and the key (for DMA) is to use the base count to determine the revisit frequency.

Understood and thanks for chiming in; that demo pack will be very valuable to detect any regressions with this simplified approach. We'll see! 😎

johnnovak:
> There are 50 demos that setup a smaller DMA size; I've tagged their directories with the 'dma' suffix.

Good stuff, thanks a lot for these! 🥳 I'll upload them our sound tests wiki page 🚀

src/hardware/sblaster.cpp
interloper98: Yup, the emulator's overall 1ms tick approach is too coarse; I should have used a more generous quantity that goes well beyond 1ms. 

(but using the full index + callback has very accurate timing, enough to get down to single sample-level timing).

The one-audio-frame at a time approach is how the hardware works; the card requests frame-by-frame bytes from the DMA controller paced out at the playback rate, and that code path universally works; but there's a bit more overhead in pumping all the queues with one frame instead of 1ms-worth of frames (so that's why I left the two approaches in my last SB PR).

I'm confident @weirddan455 will figure it out, and land on the best of both worlds :-)

johnnovak: > but there's a bit more overhead in pumping all the queues with one frame instead of 1ms-worth of frames (so that's why I left the two approaches in my last SB PR).

Yeah, so I'm all for getting closer to hardware-like behaviour in the code, as I said that above. Performance is a non-issue IMO if it's fast enough on current(ish) hardware. IMO, optimisations like this should only be done if there is an insurmountable barrier to getting the hardware-like emulation working. Case in point, the 16-bit colour restriction of the VGA code that we removed 1.5 years ago. Probably that made things more efficient in the early 2000s, and surely we've added some "overhead" by doing things the proper way by using 24-bit colours, but "optimising" for 5% less CPU on current hardware at the expense of accurate emulation is a hard sell.

weirddan455: Thanks a lot for the explanation @interloper98  I was hoping to get your input on this.  I was asking about you on Discord but figured you were away/busy.

I had some incorrect assumptions about how this per-frame callback worked when I was reviewing your PR up until I wrote this and took some measurements.  I was under the assumption that the CPU emulation core always ran in 1ms ticks and that there was only a single PIC/Event queue.

The truth is, it runs at a much smaller granularity.  The TickHandlers run every 1ms but the Event queue you were using for per-frame can be scheduled at much smaller intervals and uses an entirely separate queue.  It makes a lot more sense why you went for this approach now.

My main problem with the per-frame callbacks is that it makes it harder to synchronize with the mixer.  My approach in this PR is to simply read more frames from DMA when the audio queue is running low.

I'll test with that suite you provided and see if there's regressions.  If so, it seems the core problem is that we need to read DMA more frequently.  I could probably fine a place in our main loop to do these callbacks in that case.  It may also be needed for other audio devices.

weirddan455:
This causes a regression in Triton's Crystal Dream demo as @interloper98 suspected.  I tested that the sound was broken in v0.82.0 (before the per-frame callbacks were introduced), is working in `main` and then regressed in my commit to remove per-frame.

I'm going to close this and rethink my approach.  I have an idea of what needs to happen to reduce SoundBlaster blocking without causing regressions but I need to work on it and I'll just open a new PR when that's ready.

johnnovak:
> I'm going to close this and rethink my approach. I have an idea of what needs to happen to reduce SoundBlaster blocking without causing regressions but I need to work on it and I'll just open a new PR when that's ready.

That's how it goes, it was a nice try 😄 I'm hoping you'll be able to keep some of those noce cleanup PRs.

Thatnk for the test pack again @interloper98; stuff like that is super valuable for avoiding regressions 💯

Issue #3993, opened 2024-10-18
Blood heavy sound crackling when using GUS


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.82.0 latest "final" RC

### What Operating System are you using?

macOS (Apple Silicon)

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Blood has strong crackling right from the start in current 0.82.0 "final" RC that doesn't happen in 0.81.2.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Use the attached repro pack to launch the game.
2.
3.

### Download URL of affected game or software

https://www.dropbox.com/scl/fi/449kykbe1yalem0zcr90b/Blood-0.82.0-GUS-repro-pack.7z?rlkey=x0nvrvpe10ih8ddi1ksuou3ey&st=q4zyli4x&dl=0

### Your configuration

```ini
Default for 0.82.0, except gus = true.
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

interloper98:
Bisect is landing on d75e1057f5c6f7854b7ab1d9b3874cc8a6b8afce ; pinging  @weirddan455.

At least it's reproducible right away in corporate banner. A couple recordings (I'm using `mixer gus 200` because the FMV volume is a bit soft).

Before:

https://github.com/user-attachments/assets/93538323-9bb0-43ae-9679-e83997f22fe4


After: 

https://github.com/user-attachments/assets/f59044fc-0d54-4515-9736-feaa8c009e8d

johnnovak:
@Burrito78 Set `cpu_throttle = on`. Does that fix it?

weirddan455:
>  Set `cpu_throttle = on`. Does that fix it?

I was able to reproduce it.  That doesn't help.  I'm not seeing any under/overflows in the mixer so I think the bug is in the GUS code.  It's doing some IRQ and DMA stuff inside the callback.  I'm going to try moving it to the main thread just as a test to see if it's a race condition that for some reason didn't affect other games I tested.

weirddan455:
Moving GUS to the main thread does resolve the issue.  I've a got a local branch now that I could clean up and put a PR in.  It's kind of unfortunate if we have to go that route but it's better than having regressions I guess.

[...]

Issue #3887, opened 2024-08-21
Jane's ATF with NATO Fighters expansion (1996) 0.82.0 pacing issue


Grounded0:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.82.0 RC1

### What Operating System are you using?

macOS (Intel)

### If Other OS, please describe

macOS 12.6.5

### Relevant hardware info

2017 15" MacBook Pro with Radeon Pro 560

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Jane's ATF with NATO Fighters expansion triggers a weird audio glitch on the intro screen and can be listened in the menus. Regression from 0.81.2. Combat simulation seems to run glitch free probably because it puts an enormous workload on the main thread as these are among the most demanding DOS games ever released:

<img width="1392" alt="Screenshot 2024-08-21 at 12 15 11" src="https://github.com/user-attachments/assets/6d67d2c7-daed-4425-b7f4-b6bd1c3f9b20">

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Put the emulation to balls to the wall setting like thread load 100 %
2. Run Jane's ATF with NATO Fighters expansion
3. Hear the audio glitching on the menus before entering combat simulation

### Your configuration

```ini
https://github.com/user-attachments/files/16689305/dosbox-staging.conf.zip
```
```
ajp@milla:~/Misc/dosbox/c_drive/JANES/ATF $ cat RUN.BAT 
imgmount -u d
imgmount d y:\atf\atf.iso y:\atf\nato.img -t cdrom
config -set "cycles=auto 12000 105%%"
af.exe %1 %2 %3
config -set "cycles=auto 12000 50%%"
if errorlevel = 255 goto exit
if errorlevel = 1 goto TNTerror
goto exit

:TNTerror
echo 

echo 

echo You do not have enough conventional memory to run Advanced Tactical Fighters!
echo Please refer to your Install Guide on making a boot disk that will free
echo a total of 420K of memory.
echo 

goto exit

:exit
ajp@milla:~/Misc/dosbox/c_drive/JANES/ATF $ 
```

### Provide a Log

```text
ajp@milla:~ $ dosbox
date       time         | 
2024-08-21 12:07:22.575 | arguments: dosbox
2024-08-21 12:07:22.575 | Current dir: /Users/ajp
[...]

PR #4175, opened 2025-02-04
Increase the number of audio frames generated by SoundBlaster when the mixer is underflowing

weirddan455:
# Description

Second attempt at #4165 but this one is much more minimalistic in number of changes.

I'm keeping the per-frame callbacks.  Triton's Crystal Dream demo is an extreme example of why we need them.  That demo sets up DMA to read a single 2 byte frame from the same memory location repeatedly.  The demo code is modifying that 2 byte memory location on-demand and we have to read at a 44 microsecond interval to keep up.

`PIC_AddEvent()` is the cleanest way to achieve this.  It modifies the `CPU_Cycles` variable to ensure that we return from the CPU core in time to run the next event.

# Release notes

Reduce audio stuttering in games using SoundBlaster's DMA and Direct DAC modes.

# Manual testing

Still need to do more testing but this is what I've done so far.

- Tyrian (fixes stuttering in main menu and Esc in-game menu)
- Duke3D (fixes stuttering on the 3DRealms screen)
- Quake
- Doom
- Archimedean Dynasty (fixes stuttering at 80k cycles with 3dfx patch)
- Blood

Demo's from @interloper98 's test suite

- Crystal Dreams
- Buttman (lol)
- Overlord (best test-case of DAC I've seen as it outputs constant audio)

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -3080,11 +3080,14 @@ static void generate_frames(const int frames_requested)
 	} break;
 
 	case DspMode::Dac:
-		// DAC mode must render one frame at a time because the DOS
+		// DAC mode typically renders one frame at a time because the DOS
 		// program will be writing to the DAC register at the playback
 		// rate.
-		assert(frames_requested == 1);
-		sblaster->output_queue.NonblockingEnqueue(sb.dac.RenderFrame());
+		// In a mixer underflow situation, we render the current frame multiple times.
+		for (int i = 0; i < frames_requested; ++i) {
+			sblaster->output_queue.NonblockingEnqueue(sb.dac.RenderFrame());
+		}
+		frames_added_this_tick += frames_requested;
 		break;
 
 	case DspMode::Dma:
@@ -3130,7 +3133,9 @@ static void per_tick_callback()
 
 	static float frame_counter = 0.0f;
 
-	frame_counter += sblaster->channel->GetFramesPerTick();
+	frame_counter += std::max(
+		static_cast<float>(sblaster->frames_needed.exchange(0, std::memory_order_acq_rel)),
+		sblaster->channel->GetFramesPerTick());
 	const int total_frames = ifloor(frame_counter);
 	frame_counter -= static_cast<float>(total_frames);
 
@@ -3160,8 +3165,29 @@ static void per_frame_callback(uint32_t)
 		return;
 	}
 
-	generate_frames(1);
+	int mixer_needs = std::max(
+		sblaster->frames_needed.exchange(0, std::memory_order_acq_rel),
+		1);
+
+	// Frames added this tick is only useful when we're in an underflow situation with the mixer.
+	// generate_frames() may not give us everything we need in a single call.
+	// We're not concerned about over-filling while in this mode so just zero it out.
+	frames_added_this_tick = 0;
+	while (frames_added_this_tick < mixer_needs) {
+		generate_frames(mixer_needs - frames_added_this_tick);
+	}
+
 	add_next_frame_callback();
+}
+
+void SBLASTER::MixerCallback(const int frames_requested)
+{
+	constexpr bool Stereo      = true;
+	constexpr bool SignedData  = true;
+	constexpr bool NativeOrder = true;
+
+	frames_needed.store(std::max(frames_requested - check_cast<int>(output_queue.Size()), 0), std::memory_order_release);
+	MIXER_PullFromQueueCallback<SBLASTER, AudioFrame, Stereo, SignedData, NativeOrder>(frames_requested, this);
 }
 
 static SbType determine_sb_type(const std::string& pref)
@@ -3467,14 +3493,10 @@ SBLASTER::SBLASTER(Section* conf)
 		channel_features.insert(ChannelFeature::Stereo);
 	}
 
-	constexpr bool Stereo      = true;
-	constexpr bool SignedData  = true;
-	constexpr bool NativeOrder = true;
-
 	const auto callback = std::bind(
-	        MIXER_PullFromQueueCallback<SBLASTER, AudioFrame, Stereo, SignedData, NativeOrder>,
-	        std::placeholders::_1,
-	        this);
+	        &SBLASTER::MixerCallback,
+	        this,
+	        std::placeholders::_1);
 
 	channel = MIXER_AddChannel(callback,
 	                           DefaultPlaybackRateHz,
--- a/src/hardware/sblaster.h
+++ b/src/hardware/sblaster.h
@@ -39,6 +39,7 @@ public:
 	// Public members used by MIXER_PullFromQueueCallback
 	RWQueue<AudioFrame> output_queue{1};
 	MixerChannelPtr channel = nullptr;
+	std::atomic<int> frames_needed = 0;
 
 private:
 	IO_ReadHandleObject read_handlers[0x10]   = {};
@@ -50,6 +51,7 @@ private:
 
 	void SetupEnvironment();
 	void ClearEnvironment();
+	void MixerCallback(const int frames_requested);
 };
 
 #endif


johnnovak:
Could you test these games as well from https://github.com/dosbox-staging/dosbox-staging/issues/4049?

- **Archimedean Dynasty** (3dfx version; 80k cycles was fine in 0.81.1, but glitching starts above 60k in 0.82.0)
- **Blood** (GUS and SB with DMA above ~50k cycles for high-res SVGA modes, more details https://github.com/dosbox-staging/dosbox-staging/issues/3993 
- **Nethack — Falcon's Eye**
- https://github.com/dosbox-staging/dosbox-staging/issues/3887

The idea is that this fix would address those regressions as well, correct?

weirddan455:
>  Archimedean Dynasty (3dfx version; 80k cycles was fine in 0.81.1, but glitching starts above 60k in 0.82.0)

Tested this.  It's not a regression for me.  It stutters just as badly in v0.81.1 as it does in v0.82.0.  Make sure you update your config file when switching versions.  v0.82.0 also introduced the new CPU cycles settings.  If you try to use this config on v0.81.2, it falls back to `cycles = max` which doesn't stutter due to CPU throttling.

In any case, this PR actually does fix the stuttering for the most part.  I still hear some small stutters on screen transitions but I hear those same stutters on v0.81.2 with `cycles = max`.  At `cycles = 80000` it's much better though.  There's no more constant crackling like there is in `main` and below.

I wasn't sure if this PR would help with pure DMA stutters.  The main goal is to unblock the mixer so that OPL wouldn't stutter.  This game stutters entirely with DMA sound and it does help so it seems to be a win all around.

weirddan455:
>  Blood (GUS and SB with DMA above 50k cycles for high-res SVGA modes, more details https://github.com/dosbox-staging/dosbox-staging/issues/3993 (you mentioned the Blood issue is fixed in the other ticket as fixed).

I'm not sure why you noted that bug for GUS _and_ SB.  The bug I fixed was only for GUS (and I may re-visit that next, I moved GUS to the main thread to fix Blood which is not ideal).

Regardless, I tested it and cannot reproduce any stuttering with SoundBlaster on `main` or in this PR.  I went all the way up to 300k cycles at 800x600 and it plays fine.

weirddan455:
>  Nethack — Falcon's Eye

This is an interesting one.  I had to go up to 300k cycles to reproduce the stutters on `main`.  This PR "fixes" the stutters but it's really just a workaround.  It slows the OPL music down, while v0.81.2 played at full speed with no stutters at 300k cycles.

I bisected this to b97cbafec4772be769fb9a4aa25074aba64a9ad4 which was in the same PR as my mixer re-work but this commit was before I threaded the mixer.  I was fixing some stupid thread sanitizer warnings but ordered this commit before I did the threading for easier bisecting in case of regressions.

I think this is really just a performance regression.  By default C++ will use [std::memory_order_seq_cst](http://en.cppreference.com/w/cpp/atomic/memory_order) which is the strictest (and slowest).  I'll need to take a closer look at the usage here but I think we can probably lower these to relaxed ordering.  IIRC, the mixer only needs to read PIC timings (never writes) and it's mostly only used for fast-forward mode.

johnnovak:
> I'm not sure why you noted that bug for GUS and SB. The bug I fixed was only for GUS (and I may re-visit that next, I moved GUS to the main thread to fix Blood which is not ideal).

General confusion I guess plus I was in a hurry 😆 

> I wasn't sure if this PR would help with pure DMA stutters. The main goal is to unblock the mixer so that OPL wouldn't stutter. 

Yeah you did say that in the original PR, I just thought the cause for both the OPL + DMA stutters and the sound stutter regressions at high cycles settings have the same root cause—your mixer enhancements, so this would address both. But I get it if you wanna focus on only one issue at a time, and there is overlap between the two so fixing one probably improves the other as you have noticed in your testing.

Nice progress, I will also give this a good testing, probably on the weekend when I have time.

> I'll tackle this in a separate PR though. It's not strictly related to SoundBlaster.

Sounds like plan 👍🏻

interloper98:
Im not able to do an in-depth review, but if those fine-grain DAC test cases are running good, then that's a good sign.

My only worry would be that some SDL (or host drivers) might make SDL do big chunky reads (say, 64 KB at a time, then it sits dormant for 10ms and plays the buffer out to the driver), in which case we could get long runs of duplicate FAC samples, instead of "metering them out" at the games rate.

Kind of like toilet paper rationing during COVID - some families demanded filling the back of their Suburban, but the shop owner assured them that only 2 rolls per family per day was sufficient. 

So SDL might /want/ to fill it's big buffer, but dosbox will assuredly give if frames just-in-time :-)

Anyways.. enough analogies! If you've added logging an saw that the normal scenario (non FFWD) still drip feeds one DAC sample at time (instead of big runs of dupeS), then it sounds like things are working good.

weirddan455:
>  My only worry would be that some SDL (or host drivers) might make SDL do big chunky reads (say, 64 KB at a time, then it sits dormant for 10ms and plays the buffer out to the driver)

Yeah, that's typical.  SDL usually wants chunks of 10-15ms worth of audio as a blocksize.

>  in which case we could get long runs of duplicate DAC samples, instead of "metering them out" at the game's rate (which happens to also match the SB channel rate).

This never happens though.  This PR can only increase the number of samples read.  It never decreases or makes it go dormant.  It still reads a single sample in the per-frame callback at the same rate.  If the buffer stays filled, that's all it ever does.  If the buffer starts to run dry, it reads extra samples to fill the deficit and then goes back to single sample reads.

interloper98:
>  This PR can only increase the number of samples read. It never decreases or makes it go dormant. It still reads a single sample in the per-frame callback at the same rate. If the buffer stays filled, that's all it ever does. If the buffer starts to run dry, it reads extra samples to fill the deficit and then goes back to single sample reads.

👌 Perfect @weirddan455 ; then it's a go from my limited review.

johnnovak:
> > My only worry would be that some SDL (or host drivers) might make SDL do big chunky reads (say, 64 KB at a time, then it sits dormant for 10ms and plays the buffer out to the driver)
> 
> Yeah, that's typical. SDL usually wants chunks of 10-15ms worth of audio as a blocksize.

Huh, what? 64K audio buffers? No such thing exists.

Audio buffer sizes might grew to 4096-8192 samples in _extreme scenarios_. But normal buffer sizes are in the 256-1024 sample range. Even the SDL doco says so:

https://wiki.libsdl.org/SDL2/SDL_OpenAudioDevice

I don't think we should worry or care about such hypothetical buffer size, they just don't happen in real life. If you think about it, those buffers in the hardware cost money, nobody puts such huge buffers in hardware, and besides, it's pointless, it would result in crazy long latency times.

> desired->samples is the desired size of the audio buffer, in sample frames (with stereo output, two samples--left and right--would make a single sample frame). This number should be a power of two, and may be adjusted by the audio driver to a value more suitable for the hardware. Good values seem to range between 512 and 4096 inclusive, depending on the application and CPU speed. Smaller values reduce latency, but can lead to underflow if the application is doing heavy processing and cannot fill the audio buffer in time. Note that the number of sample frames is directly related to time by the following formula: ms = (sampleframes*1000)/freq

weirddan455:
>  Huh, what? 64K audio buffers? No such thing exists.
> 
>  Audio buffer sizes might grow to 4096-8192 samples, but only in extreme scenarios. Normal buffer sizes are in the 256-1024 sample range. Even the SDL doco says so:

8192 * 8 bytes per frame (4 byte floats * 2 channels) gives you a 64KB buffer.  At 48,000hz sample rate, that's 170ms of latency.  That's the maximum blocksize we support.  Default is 512.

Regardless, it doesn't matter what the host blocksize is.  We still fill at a constant rate.  This PR just provides a way to compromise if the buffer starts to underflow.

>  Second comment: does this pattern also apply to the other DACs like Tandy, PS1audio, and the LPT devices?

Yeah, it might.  I would just like to focus on one audio device at a time and do testing on it to try and prevent regressions.

>  (Or are those non issues because the games are much less demanding and at low cycles there no issues with the core blocking for too long?)

It's not really an issue of how demanding the game is.  Any spikes in core emulator performance can cause audio stutters, even at low cycles.  It's more likely to happen at high cycles though.  Either way, the goal is to build mitigations to prevent a full mixer stall when these spikes happen.

interloper98:
Yup; I wanted to be sure the generation side is paced per-frame (unique frames), and that SDL's more coarse per-blocksize requests, like @weirddan455 mentioned, are firewalled at the mixer level. Sounds like this all in order.

johnnovak:
> > Huh, what? 64K audio buffers? No such thing exists.
> > Audio buffer sizes might grow to 4096-8192 samples, but only in extreme scenarios. Normal buffer sizes are in the 256-1024 sample range. Even the SDL doco says so:
> 
> 8192 blocksize * 8 bytes per frame (4 byte floats * 2 channels) gives you a 64KB buffer. At 48,000hz sample rate, that's 170ms of latency. That's the maximum blocksize we support. Default is 512.

Yeah, okay... I guess thinking in kilobytes is just a weird way to think about audio latencies. It's the number of stereo sample frames that matters. 8192 sample is an ultra-extremely large buffer size. Actually useful range stops at about 2048 samples. I'd be surprised if any card or integrated audio from the last 15-20 years really _needed_ more than 2048, but we support it... OK 😄 

> Regardless, it doesn't matter what the host blocksize is. We still fill at a constant rate. This PR just provides a way to compromise if the buffer starts to underflow.

Yeah, that's good 👍🏻 

> > Second comment: does this pattern also apply to the other DACs like Tandy, PS1audio, and the LPT devices?
> 
> Yeah, it might. I would just like to focus on one audio device at a time and do testing on it to try and prevent regressions.

Sure thing.

> > (Or are those non issues because the games are much less demanding and at low cycles there no issues with the core blocking for too long?)
> 
> It's not really an issue of how demanding the game is. Any spikes in core emulator performance can cause audio stutters, even at low cycles. It's more likely to happen at high cycles though. Either way, the goal is to build mitigations to prevent a full mixer stall when these spikes happen.

That's exactly it; the audio path needs to be reasonably "spike tolerant".

johnnovak:
Tested these games so far, this is working great @weirddan455! 🥳 No glitches when the OPL/MIDI and the SB are playing at the same time. I tried spamming Alt+Enter in all games, couldn't trigger dropouts that way either.

- **Azrael's Tear** (Sound Canvas + SB speech & sfx)
- **Dark Seed II** (Win 3.x, MIDI via Windows OPL Synth + SB speech & sfx)
- **Dreamweb** (SB sound)
- **Elvira** (OPL + SB sfx)
- **Indiana Jones and the Fate of Atlantis** (CD version, MT-32 + SB speech & sfx)
- **Rome: Pathway to Power** (MT-32 + SB sfx)
- **Shivers** (Win 3.x, SB16 sound)
- **System Shock** (CD version, Sound Canvas + SB speech + sfx)
- **X-COM: UFO Defense** (OPL + SB sfx)

johnnovak:
@weirddan455 I'm thinking this looks solid enough for now so maybe merge it? I was planning to do more testing on the weekend, but you also raised https://github.com/dosbox-staging/dosbox-staging/pull/4181, so I'd like to cut the testing effort in half and test that branch rebased to master once this is merged.

How does that sound? Or do you you think it would be risky?

johnnovak (approved):
Very nice work; this basically eliminated all remaining glitches and dropouts from the audio emulation for me! 🎉 🥳

weirddan455:
> @weirddan455 I'm thinking this looks solid enough for now so maybe merge it? I was planning to do more testing on the weekend, but you also raised #4181, so I'd like to cut the testing effort in half and test that branch rebased to master once this is merged.
> 
> How does that sound? Or do you you think it would be risky?

Yeah, that makes sense.  I was planning on testing some more of @interloper98 's test suite but it would be easier to test both PRs at once.

The only thing is the Nethack issue is a bit easier to hear the difference in without this PR.  This PR makes the OPL track slow down rather than stutter (because the mixer doesn't block waiting on SoundBlaster which is running but playing silence on the main menu).  Stuttering is more obvious to my ears when there's a problem.

Just something to keep in mind though.  I'm fine with merging and if I run into an issue in testing I can bisect between the 2 PRs.

I assume you tested on Windows based on the "MIDI via Windows OPL Synth" comment.  So I can check that off the list?  Do you still have a Mac to test on (can wait if you want to merge and test the 2nd PR together with this)?

johnnovak:
> Yeah, that makes sense. I was planning on testing some more of @interloper98 's test suite but it would be easier to test both PRs at once.

Yeah let's do that then (test both at once).

> The only thing is the Nethack issue is a bit easier to hear the difference in without this PR. This PR makes the OPL track slow down rather than stutter (because the mixer doesn't block waiting on SoundBlaster which is running but playing silence on the main menu). Stuttering is more obvious to my ears when there's a problem.

Yeah I kinda trust you on the Nethack fix. I guess I just wanna test the final results of both PRs applied form and end-user's perspective (do I hear any audio issues? yes or no).

> I assume you tested on Windows based on the "MIDI via Windows OPL Synth" comment. So I can check that off the list? Do you still have a Mac to test on (can wait if you want to merge and test the 2nd PR together with this)?

I can only test on Windows currently, so yes, you can tick off the Windows testing part.

"MIDI via Windows OPL Synth" meant the Windows 3.x OPL synth. So in Windows 3.x you can route General MIDI either to a MIDI output port (this way we can use FluidSynth or Sound Canvas, or any MIDI device external to Staging), or you can route it to the internal OPL synthesiser of Windows 3.x which emulated General MIDI instruments. I used the latter in my testing.

As you can see on the below screenshot:

- **SB16 All/Bas(ic)/Ext(ended) MIDI** -- output to a real MIDI port (emulated MIDI port in our case)
- **SB16 All/Bas(ic)/Ext(ended) Synth** -- route MIDI to the Win3.x OPL synth that emulated General MIDI

image

weirddan455:
>  "MIDI via Windows OPL Synth" meant the Windows 3.x OPL synth. So in Windows 3.x you can route General MIDI either to a MIDI output port (this way we can use FluidSynth or Sound Canvas, or any MIDI device external to Staging), or you can route it to the internal OPL synthesiser of Windows 3.x which emulates General MIDI instruments. I used the latter in my testing.

Oh, that's cool.  I was not aware that existed.  I don't really play any Windows 3.x games.

>  Yeah let's do that then (test both at once).

Ok, I'll go ahead and merge this and rebase the other PR.

PR #4181, opened 2025-02-07
Remove std::atomic from performance critical global variables

### LINKED ISSUE #4176: Performance regression from use of std::atomic

Latest main (regression from v0.82.0)

Linux x86_64

Debian 12

### Relevant hardware info

_No response_

I've bisected a performance issue to https://github.com/dosbox-staging/dosbox-staging/commit/b97cbafec4772be769fb9a4aa25074aba64a9ad4 with the game "Nethack — Falcon's Eye" when using 300k cycles which causes audio stutters not present before.

Root cause is the use of `std::atomic` inside a performance critical function `PIC_RunQueue()`.  I can measure that this functions runs at least every 5 microseconds when using high cycle counts though I suspect it's running at sub-microsecond intervals because sticking a call to `GetTicksUs()` slows things down to a crawl.  It's really too fast to measure 😄 

I honestly just stuck `std::atomic` on these global variables to shut up thread sanitizer warnings.  See #3848 where the very first comment I received was thread sanitizer complaining about this.

Why is `std::atomic` so slow?  Well it defaults to [std::memory_order_seq_cst](http://en.cppreference.com/w/cpp/atomic/memory_order).  Unfortunately, changing this to relaxed ordering (which is all we should need for this particular use-case) leads to so gnarly-looking code.  Take this line:

```cpp
CPU_CycleLeft += CPU_Cycles;
```

These are both atomic now.  To make everything relaxed ordering and to remove the atomic add (We don't need it, the main thread is the only one that ever writes to these variables.  The mixer thread only reads.) you have to do this:

```cpp
CPU_CycleLeft.store(CPU_CycleLeft.load(std::memory_order_relaxed) + CPU_Cycles.load(std::memory_order_relaxed), std::memory_order_relaxed);
```

A lot less readable but I tested that doing this everywhere does solve the performance issue.

I think a better solution is going to be removing the `std::atomic` entirely from these global variables and committing these timing changes to a separate `std::atomic` (or mutex locked) variable that lives in `mixer.cpp`.  Maybe make that happen once per ms or something (much better than 5-10 atomic operations per microsecond).

I'm going to be reviewing all of the other `std::atomic` usages I've made simply to shut up thread sanitizer and probably reverting them.

1. Set cycle count to 300k.  Some have reported stuttering happening at 100k or lower.
2. Launch Nethack — Falcon's Eye
3. Audio stutters

_No response_

### Your configuration

```ini

```

### Provide a log

```text

```

- [x] Yes, I agree.

weirddan455:
# Description

First few commits revert some mutexes and atomics that were only used by GUS.  GUS has since been moved to the main thread so these are no longer needed.

GUS was raising some IRQs and that's probably part of the reason it didn't work out so well threaded.  These are unlikely to return.  If I find a way to (partially) thread GUS, it will still need a main thread callback to get the timings right for Blood.  IRQ stuff should stay on the main thread.  It just never works out well with timing and trying to guarantee thread safety is also a headache.

I then implemented `PIC_AtomicIndex()`.  This gets updated by the main thread inside `PIC_RunQueue` which runs at least once per millisecond.  In practice, it runs much more frequently and at 300k cycles it's somewhere close to every microsecond.

This is precise enough for our needs in the mixer and audio callbacks.  It's only used to update `last_rendered_ms` variables and for some special fast-forward code.  It won't work globally though.  VGA had some regressions when I tried so I'm not going there and added a warning comment.

Finally, remove the `std::atomics` from the CPU cycles variables.  These get updated at least once every cycle inside CPU core emulation.  In hindsight, not a good idea to add that many atomic operations to performance critical sections.

It wasn't even correct before anyway.  Calculating `PIC_FullIndex()` was using multiple atomic operations on 3 different variables.  We never guaranteed atomicity of the entire thing.  Now we do.

## Related issues

Fixes #4176

# Release notes

Fix audio stuttering at high cycle count in Nethack — Falcon's Eye

# Manual testing

- Nethack — Falcon's Eye now runs stutter free on my system up to 300k cycles.  This is independent of my changes on #4175  I based and tested this from `main` without the other PR applied.
- Could not measure any significant change in FPS from a Quake timedemo.  If anyone knows of a good DOS benchmark, I'd be curious to see if we get improvements.

Regression testing for affected audio devices:

- OPL - Tyrian
- MT-32 - Eric the Unready, Space Quest 3
- Fluidsynth - Doom, Duke3D
- SoundCanvas - Doom, Duke3D
- IMFC - Police Quest 2
- GameBlaster - Silpheed
- Innovation - Red Storm Rising
- PS1 Audio - Sierra's Christmas Card
- Tandy - Thexder

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/include/pic.h
+++ b/include/pic.h
@@ -29,7 +29,7 @@
 typedef void(PIC_EOIHandler)();
 typedef void (*PIC_EventHandler)(uint32_t val);
 
-extern std::atomic<uint32_t> PIC_IRQCheck;
+extern uint32_t PIC_IRQCheck;
 
 // Elapsed milliseconds since starting DOSBox
 // Holds ~4.2 B milliseconds or ~48 days before rolling over
--- a/include/pic.h
+++ b/include/pic.h
@@ -35,6 +35,8 @@ extern uint32_t PIC_IRQCheck;
 // Holds ~4.2 B milliseconds or ~48 days before rolling over
 extern std::atomic<uint32_t> PIC_Ticks;
 
+extern std::atomic<double> atomic_pic_index;
+
 // The number of cycles not done yet (ND)
 static inline int32_t PIC_TickIndexND()
 {
@@ -58,6 +60,20 @@ static inline int32_t PIC_MakeCycles(double amount)
 static inline double PIC_FullIndex()
 {
 	return static_cast<double>(PIC_Ticks) + PIC_TickIndex();
+}
+
+// Thread safe version of PIC_FullIndex()
+// Callers on the main thread should prefer PIC_FullIndex() as it is more precise.
+// I attempted to change this everywhere and had regressions from VGA code for example.
+// It should be good enough for audio though.
+static inline double PIC_AtomicIndex()
+{
+	return atomic_pic_index.load(std::memory_order_acquire);
+}
+
+static inline void PIC_UpdateAtomicIndex()
+{
+	atomic_pic_index.store(PIC_FullIndex(), std::memory_order_release);
 }
 
 void PIC_ActivateIRQ(uint8_t irq);
--- a/include/pic.h
+++ b/include/pic.h
@@ -33,7 +33,7 @@ extern uint32_t PIC_IRQCheck;
 
 // Elapsed milliseconds since starting DOSBox
 // Holds ~4.2 B milliseconds or ~48 days before rolling over
-extern std::atomic<uint32_t> PIC_Ticks;
+extern uint32_t PIC_Ticks;
 
 extern std::atomic<double> atomic_pic_index;
 
--- a/include/cpu.h
+++ b/include/cpu.h
@@ -54,18 +54,13 @@ enum class ArchitectureType {
 	Mixed           = 0xff,
 };
 
-// PIC_FullIndex() is used by the mixer thread for determining timing
-// Variables used in this calculation must be atomic or Clang's thread sanitizer complains
-// This includes CPU_Cycles, CPU_CycleLeft, and CPU_CycleMax
-// The rest can be left as normal variables
-
 // Current cycles values
-extern std::atomic<int> CPU_Cycles;
-extern std::atomic<int> CPU_CycleLeft;
+extern int CPU_Cycles;
+extern int CPU_CycleLeft;
 
 // Cycles settings for both "legacy" and "modern" modes
 extern bool CPU_CycleAutoAdjust;
-extern std::atomic<int> CPU_CycleMax;
+extern int CPU_CycleMax;
 extern int CPU_CyclePercUsed;
 extern int CPU_CycleLimit;
 
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -55,13 +55,13 @@ CPUBlock cpu      = {};
 Segments Segs     = {};
 
 // Current cycles values
-std::atomic<int> CPU_Cycles    = 0;
-std::atomic<int> CPU_CycleLeft = CpuCyclesRealModeDefault;
+int CPU_Cycles    = 0;
+int CPU_CycleLeft = CpuCyclesRealModeDefault;
 
 // Cycles settings for both "legacy" and "modern" modes
-std::atomic<int> CPU_CycleMax = CpuCyclesRealModeDefault;
-int CPU_CyclePercUsed         = 100;
-int CPU_CycleLimit            = -1;
+int CPU_CycleMax      = CpuCyclesRealModeDefault;
+int CPU_CyclePercUsed = 100;
+int CPU_CycleLimit    = -1;
 
 static int old_cycle_max       = CpuCyclesRealModeDefault;
 static bool legacy_cycles_mode = false;
@@ -2416,7 +2416,7 @@ static void cpu_increase_cycles_legacy()
 		CPU_Cycles    = 0;
 
 		if (!maybe_display_switch_to_dynamic_core_warning(CPU_CycleMax)) {
-			LOG_MSG("CPU: Fixed %d cycles", CPU_CycleMax.load());
+			LOG_MSG("CPU: Fixed %d cycles", CPU_CycleMax);
 		}
 	}
 }
@@ -2538,7 +2538,7 @@ static void cpu_decrease_cycles_legacy()
 		CPU_CycleLeft = 0;
 		CPU_Cycles    = 0;
 
-		LOG_MSG("CPU: Fixed %d cycles.", CPU_CycleMax.load());
+		LOG_MSG("CPU: Fixed %d cycles.", CPU_CycleMax);
 	}
 }
 
@@ -2612,7 +2612,7 @@ std::string CPU_GetCyclesConfigAsString()
 				s += format_str("max %d%%", CPU_CyclePercUsed);
 			}
 		} else {
-			s += format_str("%d", CPU_CycleMax.load());
+			s += format_str("%d", CPU_CycleMax);
 		}
 
 		s += " ";
@@ -2919,9 +2919,8 @@ public:
 	void ConfigureCyclesLegacy(Section_prop* secprop)
 	{
 		// Sets the value if the string in within the min and max values
-		// Output value is either int or std::atomic<int>
-		auto set_if_in_range = []<typename T>(const std::string& str,
-		                          T& value,
+		auto set_if_in_range = [](const std::string& str,
+		                          int& value,
 		                          const int min_value = 1,
 		                          const int max_value = 0) {
 			std::istringstream stream(str);
--- a/src/dos/cdrom.cpp
+++ b/src/dos/cdrom.cpp
@@ -86,7 +86,7 @@ void CDROM_Interface::LagDriveResponse() const
 
 	// Handle tick-rollover
 	static decltype(PIC_Ticks) prev_ticks = 0;
-	prev_ticks = std::min(PIC_Ticks.load(), prev_ticks.load());
+	prev_ticks = std::min(PIC_Ticks, prev_ticks);
 
 	// Ensure results a monotonically increasing
 	auto since_last_response_ms = [=]() { return PIC_Ticks - prev_ticks; };
@@ -95,7 +95,7 @@ void CDROM_Interface::LagDriveResponse() const
 		CALLBACK_Idle();
 	}
 
-	prev_ticks = PIC_Ticks.load();
+	prev_ticks = PIC_Ticks;
 }
 
 void CDROM_Interface_Physical::CdReaderLoop()
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -297,7 +297,7 @@ static void increase_ticks()
 		auto ratio = (ticks.scheduled * (CPU_CyclePercUsed * 1024 / 100)) /
 		             ticks.done;
 
-		auto new_cycle_max = CPU_CycleMax.load();
+		auto new_cycle_max = CPU_CycleMax;
 
 		auto cproc = static_cast<int64_t>(CPU_CycleMax) *
 		             static_cast<int64_t>(ticks.scheduled);
@@ -417,7 +417,7 @@ static void increase_ticks()
 		// ticks.added > 15 but ticks.scheduled < 5, lower the cycles
 		// but do not reset the scheduled/done ticks to take them into
 		// account during the next auto cycle adjustment.
-		CPU_CycleMax = CPU_CycleMax / 3;
+		CPU_CycleMax /= 3;
 
 		if (CPU_CycleMax < auto_cpu_cycles_min) {
 			CPU_CycleMax = auto_cpu_cycles_min;
@@ -462,7 +462,7 @@ static void DOSBOX_UnlockSpeed( bool pressed ) {
 		if (CPU_CycleAutoAdjust) {
 			autoadjust = true;
 			CPU_CycleAutoAdjust = false;
-			CPU_CycleMax = CPU_CycleMax / 3;
+			CPU_CycleMax /= 3;
 			if (CPU_CycleMax<1000) CPU_CycleMax=1000;
 		}
 	} else {
--- a/src/gui/render.cpp
+++ b/src/gui/render.cpp
@@ -257,8 +257,6 @@ static void halt_render()
 	render.active   = false;
 }
 
-extern uint32_t PIC_Ticks;
-
 void RENDER_EndUpdate(bool abort)
 {
 	if (!render.updating) {
--- a/src/hardware/gameblaster.cpp
+++ b/src/hardware/gameblaster.cpp
@@ -250,7 +250,7 @@ void GameBlaster::AudioCallback(const int requested_frames)
 		channel->AddSamples_sfloat(1, &frame[0]);
 		--frames_remaining;
 	}
-	last_rendered_ms = PIC_FullIndex();
+	last_rendered_ms = PIC_AtomicIndex();
 }
 
 void GameBlaster::WriteToDetectionPort(io_port_t port, io_val_t value, io_width_t)
--- a/src/hardware/imfc.cpp
+++ b/src/hardware/imfc.cpp
@@ -4995,7 +4995,7 @@ AudioFrame ym2151_device::RenderFrame()
 
 void ym2151_device::RenderUpToNow()
 {
-	const auto now = PIC_FullIndex();
+	const auto now = PIC_AtomicIndex();
 	// Keep rendering until we're current
 	while (last_rendered_ms < now) {
 		last_rendered_ms += ms_per_render;
@@ -5032,7 +5032,7 @@ void ym2151_device::sound_stream_update(const int requested_frames)
 		audio_channel->AddSamples_sfloat(1, &frame[0]);
 		--frames_remaining;
 	}
-	last_rendered_ms = PIC_FullIndex();
+	last_rendered_ms = PIC_AtomicIndex();
 }
 
 // clang-format off
--- a/src/hardware/innovation.cpp
+++ b/src/hardware/innovation.cpp
@@ -256,7 +256,7 @@ void Innovation::AudioCallback(const int requested_frames)
 		}
 		--frames_remaining;
 	}
-	last_rendered_ms = PIC_FullIndex();
+	last_rendered_ms = PIC_AtomicIndex();
 }
 
 Innovation innovation;
--- a/src/hardware/input/mouse_manymouse.cpp
+++ b/src/hardware/input/mouse_manymouse.cpp
@@ -249,7 +249,7 @@ bool ManyMouseGlue::ProbeForMapping(uint8_t &physical_device_idx)
 
 	// Wait a little to speedup screen update
 	constexpr uint32_t ticks_threshold = 50; // time to wait idle in PIC ticks
-	const auto pic_ticks_start = PIC_Ticks.load();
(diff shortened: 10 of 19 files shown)


johnnovak:
Added backport label (should be in 0.82.1 I think).

include/pic.h
```diff
@@ -60,6 +62,20 @@ static inline double PIC_FullIndex()
 	return static_cast<double>(PIC_Ticks) + PIC_TickIndex();
 }
 
+// Thread safe version of PIC_FullIndex()
+// Callers on the main thread should prefer PIC_FullIndex() as it is more precise.
+// I attempted to change this everywhere and had regressions from VGA code for example.
+// It should be good enough for audio though.
```
johnnovak, line 68: Nice explanatory comment 💯

johnnovak:
(Sorry, fat fingered the close button 😊)

johnnovak:
Okay, @weirddan455, here's the first round of results. I will continue to test this over the next week, so please leave it open. Results so far:

# The good

(Re-)tested the following so far on Windows 10 and my MOTU Ultralite mk3 interface. All programs have been tested with CPU throttling disabled.

## Games

- **Azrael's Tear** (Sound Canvas music + SB speech & sfx; zero dropouts even at 600k cycles, only the MIDI music starts to slow down somewhat)
- **Dark Seed II** (Win 3.x, MIDI music via the Windows 3.x OPL Synth + SB speech & sfx)
- **Dark Seed** (CD version, OPL music + SB speech)
- **Dark Seed** (Floppy version, OPL music only)
- **DreamWeb** (CD version, SB sound; zero issues at the maximum 2 million (!) cycles setting)
- **Elvira** (OPL + SB sfx)
- **Indiana Jones and the Fate of Atlantis** (MT-32 music + SB speech & sfx)
- **Rome: Pathway to Power** (MT-32 music + SB sfx)
- **Shivers** (Win 3.x, SB sound)
- **Space Quest III** (MT-32 music & sfx)
- **Space Quest V** (Sound Canvas music + SB speech & sfx)
- **Prince of Persia** (OPL music + SB sfx)

## Demos

The GUS is overall a bit more prone to glitches when switching between windowed and fullscreen. The curious thing is such glitched have been almost completely eliminated for the SB DAC.

- Inside by CNCD (GUS)
- No! by Nooon (GUS)
- Zilog by Sunflower (GUS)

# The slightly less good 😂 

With 0.82.0 and earlier, I can run Pinball Dreams with `blocksize = 128` and `prebuffer = 10` with _almost_ no glitches (there's a very sporadic, faint dropout now and then), and without glitches with `blocksize = 256` and `prebuffer = 10`.

On `main`, these settings results in _extreme_ audio distortion artifacts. The minimum usable setting is `512 / 20`, and of course our default `1024 / 25` settings works well too, so no change there.

But the recent mixer changes (prior to this PR, actually) made the mixer a _lot_ more sensitive to the blocksize and prebuffer params. So that's a bit of a letdown because in action games like Pinball Drams 128 & 10 vs 1024/512 & 20 makes a rather noticeable difference. Just start the game, launch the ball, then use the paddles and notice the delay between the paddles moving on the screen and the sound they make.

So this is a bit of a regression. Look, _personally_ I'm very happy with your recent mixer improvements and I don't care too much about action games. It's also true that most people just leave the sound settings on the default settings. But for people on better sound cards who have tweaked the buffer sizes for minimum latency this could be a problem. So maybe there are some easy tweaks to gain back the ability to use smaller buffers?

weirddan455:
>  The GUS is overall a bit more prone to glitches when switching between windowed and fullscreen. The curious thing is such glitches have been almost completely eliminated for the SB DAC.

Yes, because GUS is running on the main thread ever since I fixed that Blood regression.  I haven't really touched it since except the small bug fix in f93f520 and I very much doubt that would break anything.  I don't even think I rebased this PR to include that commit.

I plan on looking into what I can do to improve GUS but unless you've noticed actual regression here, I don't think this is relevant to this PR.

>  But the recent mixer changes in main (prior to this PR, actually) made the mixer a lot more sensitive to the blocksize and prebuffer params, apparently.

So you're saying this regressed in the SoundBlaster PR?  And Pinball Dreams uses SoundBlaster DMA/DAC?  I'll give it a test later.

johnnovak:
> I plan on looking into what I can do to improve GUS but unless you've noticed actual regressions here, I don't think this is relevant to this PR.

Fair point, I think it's fine. I just noted my observations, but makes perfect sense, yeah.
 
> > But the recent mixer changes in main (prior to this PR, actually) made the mixer a lot more sensitive to the blocksize and prebuffer params, apparently.
> 
> So you're saying this regressed in the SoundBlaster PR? And Pinball Dreams uses SoundBlaster DMA/DAC? I'll give it a test later.

Yes. Configure Pinball Dreams for SoundBlaster Pro, and "highest sound quality" in the `SETSOUND` util. Then press space to skip the intro and you'll hear it in the intro music right away with `blocksize = 128/256` and `prebuffer = 10`. Compare it to 0.82.0 -- these settings don't result in those heavy dropouts on my system.

Then F1 in the main menu to launch a board, F1 again to start the game, then hold down the down arrow cursor key then release it to launch the ball. Press the left/right shift keys to operate the paddles, and note the sound lag when the ball is in the lower half of the screen so you see the paddles.

weirddan455:
Regarding Pinball Dreams (I tested on Pinball Dreams Deluxe CDROM version, don't know if that makes a difference), I can confirm sound distortion with `blocksize = 256` and `prebuffer = 10` but it's not a regression for me.

There's a constant static that I hear and it sounds exactly the same on v0.81.2, v0.82.0, `main` (with and without my SoundBlaster PR), and this PR.

johnnovak:
> Regarding Pinball Dreams (I tested on Pinball Dreams Deluxe CDROM version, don't know if that makes a difference), I can confirm sound distortion with `blocksize = 256` and `prebuffer = 10` but it's not a regression for me.
> 
> There's a constant static that I hear and it sounds exactly the same on v0.81.2, v0.82.0, `main` (with and without my SoundBlaster PR), and this PR.

I'm using the floppy version of Pinball Dreams, but it seems the CD version is same as the floppy version (I'm on the train now, can't test).

Okay, could be that my sound interface has lower usable settings than yours. In which case this would be a bit hard for you to troubleshoot.

I've tested with my MOTU Ultralite mk3 extrnal FIreWire interface (_almost_ pro level(ish)). SDL autopicked the WASAPI drivers for me.

I will retest in the evening with motherboard audio (Realtek, so pretty common) and different SDL audio backends and will post my results.

weirddan455:
Now that you mentioned SDL drivers, it sparked my memory that the low block size static is actually just an issue with PulseAudio (or least SDL's PulseAudio drivers).  With `SDL_AUDIODRIVER=alsa` I can play this game at `blocksize = 128` `prebuffer = 10` with no audio issues.

Either way, I can't reproduce your specific issue.  I'll test on Windows later to see if I can reproduce it there because maybe it's OS/driver specific.

In other news, I found a rare memory corruption regression in 274cd6c50a3f8443cab2e653db3cfc94607f0c31

I'm only able to reproduce it by completely turning off PulseAudio and running pure ALSA.  Simply doing `SDL_AUDIODRIVER=alsa` goes through a PulseAudio compatibility module and does not reproduce this.

It turns out the `read_dma_8bit()` and `read_dma_16bit()` functions in `sblaster.cpp` read into a fixed size buffer with no bounds checking.  Fun!  In rare cases, we overflow the buffer, corrupt variables, and segfault.  I'll need to add some bounds checks to these before I do anything else...

weirddan455:
Tested Pinball Dreams on Windows.  It's a stuttery mess going back to v0.81.2 at `blocksize = 128`.  At `blocksize = 256` I can hear a regression in 274cd6c50a3f8443cab2e653db3cfc94607f0c31  There's some stutters before but that commit made it noticeably worse.

It's not related to the memory corruption bug.  I cherry picked my fix there to verify it didn't help.

I'll look more into it later.

johnnovak:
> Tested Pinball Dreams on Windows. It's a stuttery mess going back to v0.81.2 at `blocksize = 128`. At `blocksize = 256` I can hear a regression in [274cd6c](https://github.com/dosbox-staging/dosbox-staging/commit/274cd6c50a3f8443cab2e653db3cfc94607f0c31) There's some stutters before but that commit made it noticeably worse.
> 
> It's not related to the memory corruption bug. I cherry picked my fix there to verify it didn't help.
> 
> I'll look more into it later.

Great, then I don't need to do further testing to prove it 🎉😄 Thanks for looking into it.

johnnovak:
> I'll look more into it later.

@weirddan455 
I guess it would make sense to address that in this PR then we can test everything at once? I shall hold off further testing in that case.

johnnovak:
It's great you've done a deep-dive into this @weirddan455 🎉 Indeed, with `negotiate = on` the problem is effectively gone. Also, I might have fooled myself that I'm getting better video-to-audio latencies by setting 128 ms buffer sizes... I've done the comparison again, and it must have been placebo...

I looked around a bit in the SDL issue tracker are there are some weird tickets about buffer sizes, unexpected audio callback frequencies, etc. I think all these problems are caused by the the way SDL handles audio: it has its own mixer, so effectively it does a second level of buffering on top of the audio subsystem buffering. It must need this for more advanced use cases, e.g., when you use multiple SDL-level audio streams, potentially at different sample rates and in different formats, then SDL mixes all that for you and hands it off to the audio subsystem. But we would be better of without all that by simply just feeding the audio subsysytem's buffers directly in the callbacks. Maybe SDL3 will have an option for that,  to bypass the SDL mixer and thus the resulting "double buffering" entirely.

Anyway, it's not something we can do much about now. I've experimented with DirectSound as well (`set SDL_AUDIODRIVER=directsound` in the command line before starting DOSBox), but I got worse results. I'm thinking probably enabling `negotiate` by default on Windows too is the best solution for now.

In the future, If SDL3 won't improve things and we'll be still bothered by this, another option is to integrate [RtAudio](https://github.com/thestk/rtaudio) and not use any SDL audio features at all. Probably that would mean saying goodbye to the SDL-assisted (kinda) hardware CD audio support, but RtAudio is just much better overall and supports ASIO on Windows for ultra-low latencies (could be useful for musician types like me who are also into games 😎). Duckstation uses it, and probably many other emulators and serious audio software (e.g., VCV Rack).

Okay, so this means I'll continue testing in the coming days.

weirddan455:
>  Maybe SDL3 will have an option for that, to bypass the SDL mixer and thus the resulting "double buffering" entirely.

The best you can do is let it negotiate everything.  As soon as SDL finds a forced setting (`negotiate = off` for sample rate, blocksize, number of channels, format type, etc), it creates a stream internally to work out the difference for you (may involve resampling or what have you).

SDL3 is kind of worse in that regard.  It forces everything to be a stream.  I haven't looked at the internals though to see if it has a fast-path.

>  another option is to integrate [RtAudio](https://github.com/thestk/rtaudio) and not use any SDL audio features at all. Probably that would mean saying goodbye to the SDL-assisted (kinda) hardware CD audio support, but RtAudio is just much better overall and supports ASIO on Windows for ultra-low latencies

I ripped out the SDL CD stuff sometime last year.  It should be pretty straight-forward to switch over to another audio library.  Looks like RtAudio doesn't support PipeWire on the Linux side (SDL does).  PipeWire does have a PulseAudio compatibility mode but PipeWire is supposed to be the more modern, low latency sound server.

Worth looking into in any case.  For what it's worth PulseAudio and PipeWire are both pretty easy to program for directly if it comes down to it.  They're high-ish level sound servers.  Much easier than ALSA or DirectSound anyway :laughing:

johnnovak:
> The best you can do is let it negotiate everything. As soon as SDL finds a forced setting (`negotiate = off` for sample rate, blocksize, number of channels, format type, etc), it creates a stream internally to work out the difference for you (may involve resampling or what have you).

Yeah, let's just enable negotiate for Windows too in a subsequent PR.

> SDL3 is kind of worse in that regard. It forces everything to be a stream. I haven't looked at the internals though to see if it has a fast-path.

Oh man... 🤦🏻 This makes RtAudio look more attractive by the minute 😆 

> Worth looking into in any case. For what it's worth PulseAudio and PipeWire are both pretty easy to program for directly if it comes down to it. They're high-ish level sound servers. Much easier than ALSA or DirectSound anyway 😆

Yeah, if/when I make the switch I'll probably ask you to help out with that on Linux.

johnnovak:
@weirddan455 I'm continuing with the testing today. It's going well, overall, but I found a bug that I wanted to report quickly.

I can trigger an SB DMA related crash by doing the following (also happens on `main`, but not in 0.82.0).

1. Start **Gabriel Knight: Sins of the Fathers** (CD-ROM version) configured for Sound Blaster 16 digital sound (MIDI doesn't matter).
2. Press the fast-forward hotkey when the Sierra logo appears and keep pressing it.
3. You'll most likely get an assertion failure and an "infinite loop" type of crash in release builds (the process becomes unresponsive and never recovers). This is what the debug build prints out:

```
2025-02-15 12:10:24.262 | Fast Forward ON
2025-02-15 12:10:25.700 | Fast Forward OFF
2025-02-15 12:10:26.365 | Fast Forward ON
Assertion failed: bytes_read <= DmaBufSize * sizeof(sb.dma.buf.b16[0]), file ..\src\hardware\sblaster.cpp, line 1026
```

I can trigger it 8 times out of 10 with the below settings (it's a bit harder with the default 60k cycles setting on my box):

```ini
[sblaster]
sbtype = sb16

[cpu]
cpu_cycles = 25000
cpu_cycles_protected = auto

[autoexec]
c:
cd GK1
SIERRA
```

weirddan455:
>  I can trigger an SB DMA related crash by doing the following

@johnnovak I'm pretty sure that's fixed by #4188  You approved it and then left more comments which I addressed.  I just saw you resolved those comments so I can merge that and rebase this branch if that's good with you.

johnnovak:
> > I can trigger an SB DMA related crash by doing the following
> 
> @johnnovak I'm pretty sure that's fixed by #4188 You approved it and then left more comments which I addressed. I just saw you resolved those comments so I can merge that and rebase this branch if that's good with you.

Yeah, I didn't quite make the connection (you mentioned you could only trigger it on Linux under specific scenarios). Sure, that sounds like a plan, please incorporate it into this PR so we can focus on testing a single branch.

weirddan455:
> Yeah, I didn't quite make the connection (you mentioned you could only trigger it on Linux under specific scenarios).
>
> Sure, that sounds like a plan. Please incorporate it into this PR so we can focus on testing a single branch.

Yeah, it wasn't quite a niche of a bug as I originally thought.  That was the only way I was able to reproduce it but I'm not surprised it can happen in other cases.

I never actually hit the assert you did because if the DMA controller is in 16-bit mode, the assert only triggers if you've overflowed by 2x or more of the buffer size.  That's fixed in my PR.

>  You'll most likely get an assertion failure and an "infinite loop" type of crash in release builds (the process becomes unresponsive and never recovers).

It's undefined behavior so who knows, that could happen,  In my case what happened in a release build is that the overflow clobbered the global `sb.dma.remain_size` variable setting it to 6 million or something.  Then it segfaulted in the next iteration when it used that as an index to write to whatever memory was 6MB ahead of where the buffer is.

Anyway, fix is merged now and I've rebased this on top of `main` with the fix in place.

johnnovak:
Alright, here are my current results. This is looking very good @weirddan455, all the below games were running perfectly fine without any glitches whatsoever.

Some minor general observations (not really problems, though): `vsync = on` results in the audio being paused for about half a second on windowed/fullscreen switches (Alt+Enter). With vsync off (default), you can spam Alt+Enter like a madman and the audio keeps playing without glitches or pauses.

## FluidSynth

- **Gabriel Knight 1: Sins of the Fathers** (FluidSynth music + SB16 speech & sfx)
- **Star Wars: Dark Forces (demo)** (FluidSynth music + SB16 speech & sfx)


## Game Blaster / Creative Music System (CMS)

- **Times of Lore** (CMS music)
- **Windwalker** (CMS music + PC Speaker sfx)


## GUS

Fullscreen/windowed switches cause audio pauses, but this is expected because the GUS is not threaded as you mentioned (these pauses are completely gone on all other audio devices).

- **Dark Sun: Shattered Lands** (GUS)
- **Epic Pinball** (GUS)
- **Pinball Dreams** (GUS)
- **Start Control 2** (GUS)
- **Crusader: No Regret** (GUS)


## IMFC

- **IMFCDISK** (IMFC music disk, [link](https://www.pouet.net/prod.php?which=81106); works but trips an assert which is probably unrelated, see https://github.com/dosbox-staging/dosbox-staging/issues/4200 )


## Innovation

- **Windwalker** (Innovation music + PC Speaker sfx)


## MT-32

- **Budokan: The Martial Spirit** (MT-32 only)
- **Future Wars** (CD-ROM) (CD Audio music + MT-32 sfx)
- **Future Wars** (Floppy) (MT-32 music & sfx)
- **Indiana Jones and the Fate of Atlantis** (MT-32 music + SB speech & sfx)
- **Rome: Pathway to Power** (MT-32 music + SB sfx)
- **Space Quest III** (MT-32 music & sfx)


## OPL

- **Cybersphere Plus** (OPL3 music & sfx)
- **Dark Seed** (Floppy version, OPL music only)
- **Elvira (OPL + SB sfx)**
- **Prince of Persia** (OPL music + SB sfx)


## PS/1 Audio

- **Laura Bow 1: Colonel's Bequest**


## Sound Blaster

- **Dark Seed** (CD version, OPL music + SB speech)
- **Dark Seed II** (Win 3.x, MIDI music via the Windows 3.x OPL Synth + SB speech & sfx)
- **DreamWeb** (CD version, SB sound; zero issues at the maximum 2 million (!) cycles setting)
- **Gods** (OPL + SB title music, OPL-only in-game)
- **Prehistorik 2** (SB music & sfx, same clicky mess as before but it's the same on hardware)
- **Shivers** (Win 3.x, SB sound)


## Sound Canvas

- **Azrael's Tear** (Sound Canvas music + SB speech & sfx; zero dropouts even at 600k cycles, only the MIDI music starts to slow down somewhat)
- **Discworld** (Sound Canvas + SB speech & sfx)
- **Gabriel Knight 1: Sins of the Fathers** (Sound Canvas music + SB16 speech & sfx)
- **Space Quest V** (Sound Canvas music + SB speech & sfx)
- **Star Wars: Dark Forces (demo)** (Sound Canvas music + SB16 speech & sfx)


## Tandy

- **Manhunter 1: New York**
- **Manian Mansion Enhanced**
- **Ski or Die**
- **Zak McKracken and the Alien Mindbenders Enhanced**

johnnovak:
@weirddan455 Have you executed the https://github.com/dosbox-staging/dosbox-staging/wiki/Audio-emulation-tests ? It's looking pretty good so far.

weirddan455:
I just re-based this on top of main again and ran through all of @interloper98 's test suite.  That was mostly more testing I meant to do for 274cd6c50a3f8443cab2e653db3cfc94607f0c31 which is already merged.  I found a bug with fast-forward but it's not a recent regression.  See #4208  Other than that, all of the demos worked fine.

I think these changes to the PIC atomics are lower risk than the SoundBlaster change that was already merged.  IMO this is ready for merge.  I appreciate all the testing you did on this @johnnovak

johnnovak (approved):
Absolutely stellar work @weirddan455 🎖️ Special thanks for the very thorough testing and addressing various issues promptly as we encountered them. This looks very good from my side, I'll let you press the big green button! :shipit:😎 🤘🏻

PR #4185, opened 2025-02-08
Fix command line parser bug where an extension-less file prevents execution

### LINKED ISSUE #4179: Overkill fails to launch due to having files "OVERKILL" and "OVERKILL.EXE" in the same directory

v0.82.0

Linux x86_64

Debian 12 KDE

### Relevant hardware info

_No response_

I'm only reporting this as it was mentioned on Discord.  I haven't tested this myself yet.

Root problem appears to be that if we have say `FOO` and `FOO.EXE` in the same directory, then typing `FOO` prints an error message and does not launch `FOO.EXE`.  Explicitly typing `FOO.EXE` is reported to still work.

It is reported that DOSBox-X does not have this behavior.  @NicknineTheEagle reported that he tested on real MS-DOS 6.22 and it also does not have this behavior.

We probably have a bug in our command line parser somewhere.

1. Install Overkill
2. CD into the installed directory
3. Run `OVERKILL` and get an error (must run `OVERKILL.EXE` to actually launch the game).

_No response_

### Your configuration

```ini

```

### Provide a log

```text

```

- [x] Yes, I agree.

weirddan455:
# Description

Regression from 1ebea7216206f79e37a9e5712c824199c39f51b4 - Passed wrong pointer into execute_binary_file()
From the same PR ffd952cc6c0c7a28d7d1cfe8835610ff41c67f8a - Code detecting extension-less files was removed

If a directory contains both DOOM and DOOM.EXE and a user runs the command DOOM, this was previously thowing an error.

We should try appending (in order) .COM, .EXE, and .BAT if and only if the command contains no extension.

## Related issues

Fixes #4179

# Release notes

Fixed a bug where an extension-less file with the same name as an executable incorrectly threw an "Invalid command" error when attempting to launch the executable from the command line.

# Manual testing

- Create empty file `DOOM` in the same directory as `DOOM.EXE`.
- Type `DOOM` in the command line.
- Previously this threw invalid command error.  Now it doesn't.

Also confirmed Overkill fails to launch on `main` and works after this PR.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/shell/shell_misc.cpp
+++ b/src/shell/shell_misc.cpp
@@ -498,7 +498,7 @@ bool DOS_Shell::ExecuteProgram(std::string_view name, std::string_view args)
 		return true;
 	}
 
-	const auto fullname                                = Which(name);
+	const auto fullname                                = ResolvePath(name);
 	constexpr decltype(fullname.size()) extension_size = 4;
 
 	if (fullname.empty() || fullname.size() <= extension_size) {
@@ -536,7 +536,7 @@ bool DOS_Shell::ExecuteProgram(std::string_view name, std::string_view args)
 	return false;
 }
 
-std::string DOS_Shell::Which(const std::string_view name) const
+std::string DOS_Shell::ResolvePath(const std::string_view name) const
 {
 	static constexpr auto extensions = {"", ".COM", ".EXE", ".BAT"};
 
--- a/src/shell/shell_misc.cpp
+++ b/src/shell/shell_misc.cpp
@@ -486,6 +486,23 @@ void CommandPrompt::SetCursor(const std::string::size_type index)
 	                   position_zero.page);
 }
 
+// Only for use in ExecuteProgram()
+// Not suitable for generic use as it only handles 3 letter file extensions.
+// It also doesn't look at space padding, which never happens here.
+// Space padding matters when dealing with internal DOS data strctures which we are not doing.
+static std::string_view get_executable_extension(const std::string_view filename)
+{
+	constexpr size_t ExtensionSize = 4;
+	if (filename.size() <= ExtensionSize) {
+		return "";
+	}
+	const auto dot_position = filename.size() - ExtensionSize;
+	if (filename[dot_position] != '.') {
+		return "";
+	}
+	return filename.substr(dot_position);
+}
+
 bool DOS_Shell::ExecuteProgram(std::string_view name, std::string_view args)
 {
 	if (name.size() > 1 && (std::isalpha(name[0]) != 0) &&
@@ -498,14 +515,8 @@ bool DOS_Shell::ExecuteProgram(std::string_view name, std::string_view args)
 		return true;
 	}
 
-	const auto fullname                                = ResolvePath(name);
-	constexpr decltype(fullname.size()) extension_size = 4;
-
-	if (fullname.empty() || fullname.size() <= extension_size) {
-		return false;
-	}
-
-	auto extension = fullname.substr(fullname.size() - extension_size);
+	const auto fullname  = ResolvePath(name);
+	const auto extension = get_executable_extension(fullname);
 
 	if (iequals(extension, ".BAT")) {
 		const auto current_echo = batchfiles.empty()
@@ -538,7 +549,7 @@ bool DOS_Shell::ExecuteProgram(std::string_view name, std::string_view args)
 
 std::string DOS_Shell::ResolvePath(const std::string_view name) const
 {
-	static constexpr auto extensions = {"", ".COM", ".EXE", ".BAT"};
+	static constexpr auto Extensions = {".COM", ".EXE", ".BAT"};
 
 	std::vector<std::string> prefixes = {""};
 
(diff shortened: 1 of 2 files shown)


src/shell/shell_misc.cpp
```diff
@@ -486,6 +486,19 @@ void CommandPrompt::SetCursor(const std::string::size_type index)
...
+{
+	constexpr size_t ExtensionSize = 4;
+	if (filename.size() <= ExtensionSize) {
+		return "";
+	}
+	const auto dot_position = filename.size() - ExtensionSize;
+	if (filename[dot_position] != '.') {
+		return "";
+	}
+	return filename.substr(dot_position);
```
johnnovak, line 499: So if there's a file called `A.B` that's stored internally as `A       .B  ` (space-padded to 8.3 chars)?
Or maybe it's `A.B  ` (so only the extension is space-padded to 3 chars)?

If so, and you're 100% sure about that padding is always performed, maybe put a comment there explaining this assumption.

But I'd guess I'd prefer if this code was a bit more generic and resilient to change.

johnnovak: Lol, I still have no idea how the 8.3 filenames are stored internally (space padded or not). I assume it's an inconsistent mess and all variations can occur 😂 

Found this, so there's at least a function that can space-pad filenames to the 8.3 format. But no clue about the contents of this `std::string_view filename` function arg; that can be anything.

https://github.com/dosbox-staging/dosbox-staging/blob/aacca00a25d72cbd02991651f99cb76b76d6d109/src/dos/dos_files.cpp#L1343-L1346

I wish all such functions used `DosFilename`:
https://github.com/dosbox-staging/dosbox-staging/blob/aacca00a25d72cbd02991651f99cb76b76d6d109/include/dos_system.h#L97-L100

And I also wish it was defined like this to remove all ambiguity:

```cpp
struct DosFilename {
    // space-padded
    char name[8] = {};

    // space-padded
    char ext[3]  = {};
};
```

Then we could just have a few simple functions that operate on `DosFilename` structs (equality check, ordering, etc.)

weirddan455: DOS files can have spaces? O.o

We don't care about 1-2 char extension sizes here.  This function is only called by `ExecuteProgram()` and if it's anything other than `.BAT` `.COM` and `.EXE`, it's supposed to fail.

johnnovak: > DOS files can have spaces? O.o
> 
> We don't care about 1-2 char extension sizes here. This function is only called by `ExecuteProgram()` and if it's anything other than `.BAT` `.COM` and `.EXE`, it's supposed to fail.

In that case the `get_extension()` function name is lying and it's misleading. I strongly suggest to refactor it then to `has_executable_extension()` and simply return a bool as this is way too misleading.

weirddan455: >  I think extension-less filenames are also allowed but would need to test.

Yes, they are.  That's the reason this PR exists 😆 

>  But yeah, at the FAT data level I'm pretty sure everything is space-padded to 8.3.

I'm pretty sure this function is not dealing with names at that level.  We really only care about FAT for FAT images and that's a whole separate area of the code (`drive_fat.cpp` or something).

>  In that case the get_extension() function name is lying and it's misleading. I strongly suggest to refactor it then to has_executable_extension() and simply return a bool as this is way too misleading.

Yeah, I can rename it.  This doesn't need to be some super generic thing.  If we had one of those, I would have used it.  It's called in exactly the 2 places you see in this PR.

weirddan455: I renamed the function and added a comment.

I looked at the calling code for `ExecuteProgram()`.  The filename strings are coming from the command line or from a batch file.  Spaces aren't valid as the line has already been trimmed and separated into arguments before we get here.  We're also not dealing with internal DOS structures at this point.

johnnovak: > Yeah, I can rename it. This doesn't need to be some super generic thing. If we had one of those, I would have used it. It's called in exactly the 2 places you see in this PR.

Thanks, that's fine. But don't we have a generic `get_extension` function somewhere in the codebase already? Can't we just use that and check if the returned extension is an executable extension?

weirddan455: I searched and did not find one.  The closest I saw was this:

https://github.com/dosbox-staging/dosbox-staging/blob/33a83fb77173a590b6a89be35bc22cc6bb867806/src/dos/dos_classes.cpp#L425-L433

But that's not suitable because

1. It's a member method of the `DOS_DTA` class which I'm not using here.
2. `ExecuteProgram()` is handling shell arguments.  Above `GetExtension()` function would parse `.\OVERKILL' as having an extension of '\OVERKILL' and we do the wrong thing.

johnnovak:
Just ran into this regression with **Ski or Die**.  The main executable is called `SKI.EXE` but there is also a `SKI` data file without extension in the same directory.

src/shell/shell_misc.cpp
johnnovak: Ok this is fine then for now, let's not overdo it 😄

johnnovak (approved):
Nice fix @weirddan455, ship it! :shipit:

johnnovak:
@Python-Exoproject 

This is a regression where at least **Overkill** and **Ski or Die** couldn't be started in Staging. Will be fixed in 0.82.1.

PR #4186, opened 2025-02-08
Initial Linux support in CMake files

FeralChild64:
# Description

Quick and dirty _Linux_ support in the _CMake_ files. No _Microsoft vcpkg_ required - uses _pkg-config_, like every sane piece of software does on _Linux_. Also enabled color output for the _Clang_ and _GCC_ compilers and added initial _Linux_ _CMake_ usage documentation.

As with other platforms - no _ManyMouse_ and unit tests support yet.

## Related issues

# Manual testing

To compile on Linux, type:

```bash
rm -rf ./build
cmake --preset=release-linux
cmake --build --preset=release-linux
```

or

```bash
rm -rf ./build
cmake --preset=debug-linux
cmake --build --preset=debug-linux
```

I have tried to compile on the Macintosh, but I ran into problems - the `cmake --preset=release-macos` command resulted in:

```
-- Running vcpkg install - done
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:5 (project):
  No CMAKE_C_COMPILER could be found.

CMake Error at CMakeLists.txt:5 (project):
  No CMAKE_CXX_COMPILER could be found.
```

But this happens for me on the _main_ branch, too - so this is not a regression of my PR.

I am not sure how to setup everything on _Windows_.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -34,16 +34,10 @@ CCACHE_SLOPPINESS="pch_defines,time_macros"
 > **Note**
 >
 > CMake support is currently an experimental internal-only, work-in-progress
-> feature; it's not ready for public consumption yet, but if you want to experiment:
+> feature; it's not ready for public consumption yet. If you want to experiment,
+> check the [chapter](#experimental-cmake-support) at the end of the document.
 > 
 > **PLEASE DO NOT SUBMIT ANY BUGS OR HELP REQUESTS!**
-> 
-> You'll need vcpkg installed and your VCPKG_ROOT environment variable set to vcpkg's 
-> installed location. There are presets via CMakePresets.json for convenience. For the
-> generic `release` and `debug` presets, you'll need Ninja build installed. You should
-> then be able to run e.g. `cmake --preset=debug && cmake --build --preset=debug` to 
-> produce a debug build in the `build/debug` directory. vcpkg will automatically install
-> dependencies during the build; there is no need to explicitly run `vcpkg install`.
 
 ## OS-specific instructions
 
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
 Run the sanitizer binary as you normally would, then exit and look for
 sanitizer messages in the log output.  If none exist, then your program
 is running clean.
+
+## Experimental CMake support
+
+> **Note**
+>
+> This is experimental, internal-only, work-in-progress feature.
+> 
+> **PLEASE DO NOT SUBMIT ANY BUGS OR HELP REQUESTS!**
+
+### Building on Linux, using system dependencies
+
+1. Install the necessary build tools:
+
+- GCC or Clang compiler
+- GIT
+- CMake
+- pkg-config
+
+1. Install the dependencies, development packages are needed too:
+
+- SDL 2.x
+- SDL2_net
+- IIR
+- OpusFile
+- MT32Emu
+- FluidSynth
+- ALSA
+- libpng
+- OpenGL headers
+
+1. Clone DOSBox Staging:
+
+```bash
+git clone https://github.com/dosbox-staging/dosbox-staging.git
+```
+
+1. Configure the sources and build DOSBox Staging:
+
+```bash
+cd dosbox-staging
+cmake --preset=release-linux
+cmake --build --preset=release-linux
+```
+
+1. You can now launch DOSBox with the command:
+
+```bash
+./build/release-linux/dosbox
+```
+
+### Building on Linux, using `vcpkg` to fetch dependencies at compile time
+
+1. Install the necessary build tools:
+
+- for Ubuntu:
+
+```bash
+sudo apt-get install git build-essential pkg-config cmake curl ninja-build \
+             autoconf bison libtool libgl1-mesa-dev libsdl2-dev
+```
+
+1. Install the `vcpkg` tool:
+
+- clone the `vcpkg` tool into your home directory
+
+```bash
+cd ~
+git clone https://github.com/microsoft/vcpkg.git
+```
+
+- bootstrap the `vcpkg` tool:
+
+```bash
+cd vcpkg
+./bootstrap-vcpkg.sh -disableMetrics
+```
+
+- set the vcpkg path in your compile shell (it is recommended to add the command
+  to your shell startup script, usually `~/.bashrc`)
+
+```bash
+export VCPKG_ROOT=$HOME/vcpkg
+```
+
+1. Clone DOSBox Staging into your home directory:
+
+```bash
+cd ~
+git clone https://github.com/dosbox-staging/dosbox-staging.git
+```
+
+1. Configure the sources and build DOSBox Staging:
+
+```bash
+cd dosbox-staging
+cmake --preset=release-linux-vcpkg
+cmake --build --preset=release-linux-vcpkg
+```
+
+1. You can now launch DOSBox with the command:
+
+```bash
+./build/release-linux/dosbox
+```
--- a/src/capture/CMakeLists.txt
+++ b/src/capture/CMakeLists.txt
@@ -11,11 +11,11 @@ add_library(libcapture STATIC
 		image/png_writer.cpp
 )
 
-find_package(PNG CONFIG REQUIRED)
+find_package(PNG REQUIRED)
+target_link_libraries(libcapture PRIVATE PNG::PNG)
 
 target_link_libraries(libcapture PRIVATE
 		loguru
 		zmbv
-		PNG::PNG
 		$<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.26)
+cmake_minimum_required(VERSION 3.22)
 
 set(VCPKG_USE_HOST_TOOLS ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,6 @@ add_subdirectory(src)
 target_link_libraries(dosbox PRIVATE
 		$<TARGET_NAME_IF_EXISTS:SDL2::SDL2main>
 		$<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
-
 		libcapture
 		libmisc
 		libcpu
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,10 @@ else()
 	message(NOTICE "Unknown compiler")
 endif()
 
+# Enable color output for certain compilers
+if (DOSBOX_COMPILER_GCC OR DOSBOX_COMPILER_CLANG)
+	add_compile_options("-fdiagnostics-color=always")
+endif()
 
 find_package(Git)
 if(Git_FOUND)
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -9,6 +9,7 @@
       "binaryDir": "${sourceDir}/build/debug",
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Debug",
         "RESOURCE_COPY_PATH": "/../resources"
       }
@@ -21,8 +22,59 @@
       "binaryDir": "${sourceDir}/build/release",
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Release",
         "RESOURCE_COPY_PATH": "/../resources"
+      }
+    },
+    {
+      "name": "debug-linux",
+      "displayName": "Debug (Linux host CPU), using system libs",
+      "description": "Configure for Debug build",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/debug-linux",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "RESOURCE_COPY_PATH": "/resources",
+        "USE_SYSTEM_LIBS": "ON"
+      }
+    },
+    {
+      "name": "release-linux",
+      "displayName": "Release (Linux host CPU), using system libs",
+      "description": "Configure for Release build",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/release-linux",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "RESOURCE_COPY_PATH": "/resources",
+        "USE_SYSTEM_LIBS": "ON"
+      }
+    },
+    {
+      "name": "debug-linux-vcpkg",
+      "displayName": "Debug (Linux host CPU), using vcpkg",
+      "description": "Configure for Debug build",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/debug-linux",
+      "cacheVariables": {
+        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
+        "CMAKE_BUILD_TYPE": "Debug",
+        "RESOURCE_COPY_PATH": "/resources"
+      }
+    },
+    {
+      "name": "release-linux-vcpkg",
+      "displayName": "Release (Linux host CPU), using vcpkg",
+      "description": "Configure for Release build",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/release-linux",
+      "cacheVariables": {
+        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
+        "CMAKE_BUILD_TYPE": "Release",
+        "RESOURCE_COPY_PATH": "/resources"
       }
     },
     {
@@ -33,6 +85,7 @@
       "binaryDir": "${sourceDir}/build/debug-macos",
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Debug",
         "RESOURCE_COPY_PATH": "/Resources"
       }
@@ -45,6 +98,7 @@
       "binaryDir": "${sourceDir}/build/release-macos",
       "cacheVariables": {
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Release",
         "RESOURCE_COPY_PATH": "/Resources"
       }
@@ -58,6 +112,7 @@
       "cacheVariables": {
         "CMAKE_OSX_ARCHITECTURES": "arm64",
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Release",
         "RESOURCE_COPY_PATH": "/Resources"
       }
@@ -71,6 +126,7 @@
       "cacheVariables": {
         "CMAKE_OSX_ARCHITECTURES": "x86_64",
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Release",
         "RESOURCE_COPY_PATH": "/Resources"
       }
@@ -84,6 +140,7 @@
       "cacheVariables": {
         "CMAKE_GENERATOR_TOOLSET": "ClangCL",      
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Debug",
         "RESOURCE_COPY_PATH": "/Resources"
       }
@@ -97,6 +154,7 @@
       "cacheVariables": {
         "CMAKE_GENERATOR_TOOLSET": "ClangCL",      
         "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG" : "ON",
         "CMAKE_BUILD_TYPE": "Release",
         "RESOURCE_COPY_PATH": "/Resources"
       }
@@ -110,6 +168,26 @@
     {
       "name": "release",
       "configurePreset": "release"
+    },
+    {
+      "name": "debug-linux",
+      "configurePreset" : "debug-linux",
+      "configuration": "Debug"
+    },
+    {
+      "name": "release-linux",
+      "configurePreset" : "release-linux",
+      "configuration": "Release"
+    },
+    {
+      "name": "debug-linux-vcpkg",
+      "configurePreset" : "debug-linux",
+      "configuration": "Debug"
+    },
+    {
+      "name": "release-linux-vcpkg",
+      "configurePreset" : "release-linux",
+      "configuration": "Release"
     },
     {
       "name": "debug-macos",
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -38,32 +38,6 @@
       }
     },
     {
-      "name": "release-windows",
-      "displayName": "Release (Windows host CPU)",
-      "description": "Configure for Release build",
-      "generator": "Visual Studio 17 2022",
-      "binaryDir": "${sourceDir}/build/release-windows",
-      "cacheVariables": {
-        "CMAKE_GENERATOR_TOOLSET": "ClangCL",      
-        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
-        "CMAKE_BUILD_TYPE": "Release",
-        "RESOURCE_COPY_PATH": "/Resources"
-      }
-    },
-    {
-      "name": "debug-windows",
-      "displayName": "Debug (Windows host CPU)",
-      "description": "Configure for Debug build",
-      "generator": "Visual Studio 17 2022",
-      "binaryDir": "${sourceDir}/build/debug-windows",
-      "cacheVariables": {
-        "CMAKE_GENERATOR_TOOLSET": "ClangCL",      
-        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
-        "CMAKE_BUILD_TYPE": "Debug",
-        "RESOURCE_COPY_PATH": "/Resources"
-      }
-    },
-    {
       "name": "release-macos",
       "displayName": "Release (macOS host CPU)",
       "description": "Configure for Release build",
@@ -100,6 +74,32 @@
         "CMAKE_BUILD_TYPE": "Release",
         "RESOURCE_COPY_PATH": "/Resources"
       }
+    },
+    {
+      "name": "debug-windows",
+      "displayName": "Debug (Windows host CPU)",
+      "description": "Configure for Debug build",
+      "generator": "Visual Studio 17 2022",
+      "binaryDir": "${sourceDir}/build/debug-windows",
+      "cacheVariables": {
+        "CMAKE_GENERATOR_TOOLSET": "ClangCL",      
+        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_BUILD_TYPE": "Debug",
+        "RESOURCE_COPY_PATH": "/Resources"
+      }
+    },
+    {
+      "name": "release-windows",
+      "displayName": "Release (Windows host CPU)",
+      "description": "Configure for Release build",
+      "generator": "Visual Studio 17 2022",
+      "binaryDir": "${sourceDir}/build/release-windows",
+      "cacheVariables": {
+        "CMAKE_GENERATOR_TOOLSET": "ClangCL",      
+        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
+        "CMAKE_BUILD_TYPE": "Release",
+        "RESOURCE_COPY_PATH": "/Resources"
+      }
     }
   ],
   "buildPresets": [
@@ -110,16 +110,6 @@
     {
       "name": "release",
       "configurePreset": "release"
-    },
-    {
-      "name": "debug-windows",
-      "configurePreset" : "debug-windows",
-      "configuration": "Debug"
-    },
-    {
-      "name": "release-windows",
-      "configurePreset" : "release-windows",
-      "configuration": "Release"
     },
     {
       "name": "debug-macos",
@@ -139,6 +129,16 @@
     {
       "name": "release-macos-x86_64",
       "configurePreset" : "release-macos-x86_64",
+      "configuration": "Release"
+    },
+    {
+      "name": "debug-windows",
+      "configurePreset" : "debug-windows",
+      "configuration": "Debug"
+    },
+    {
+      "name": "release-windows",
+      "configurePreset" : "release-windows",
       "configuration": "Release"
     }
   ]
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -8,10 +8,12 @@ add_library(libgui STATIC
 		titlebar.cpp
 )
 
+
 find_package(OpenGL REQUIRED)
+target_link_libraries(libgui PRIVATE OpenGL::GL)
+
 
 target_link_libraries(libgui PRIVATE
 		$<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
-		OpenGL::GL
 		libmisc
 )
--- a/src/hardware/CMakeLists.txt
+++ b/src/hardware/CMakeLists.txt
@@ -79,8 +79,11 @@ add_library(libhardware STATIC
 
 pkg_check_modules(SPEEXDSP REQUIRED IMPORTED_TARGET speexdsp)
 
-find_package(iir CONFIG REQUIRED)
-find_package(SDL2_net CONFIG REQUIRED)
+find_package(iir REQUIRED)
+target_link_libraries(libhardware PRIVATE $<IF:$<TARGET_EXISTS:iir::iir>,iir::iir,iir::iir_static>)
+
+find_package(SDL2_net REQUIRED)
+target_link_libraries(libhardware PRIVATE $<IF:$<TARGET_EXISTS:SDL2_net::SDL2_net>,SDL2_net::SDL2_net,SDL2_net::SDL2_net-static>)
 
 target_link_libraries(libhardware PRIVATE
 		libesfmu
@@ -92,6 +95,4 @@ target_link_libraries(libhardware PRIVATE
 		libgui
 		libmidi
 		libshell
-		PkgConfig::SPEEXDSP
-		$<IF:$<TARGET_EXISTS:iir::iir>,iir::iir,iir::iir_static>
-		$<IF:$<TARGET_EXISTS:SDL2_net::SDL2_net>,SDL2_net::SDL2_net,SDL2_net::SDL2_net-static>)
+		PkgConfig::SPEEXDSP)
--- a/src/libs/decoders/CMakeLists.txt
+++ b/src/libs/decoders/CMakeLists.txt
@@ -10,9 +10,17 @@ add_library(libdecoders STATIC
 
 target_include_directories(libdecoders PUBLIC ..)
 
-find_package(OpusFile CONFIG REQUIRED)
+if (USE_SYSTEM_LIBS)
+	pkg_check_modules(OPUSFILE REQUIRED IMPORTED_TARGET opusfile)
+	target_include_directories(libdecoders PRIVATE ${OPUSFILE_INCLUDE_DIRS})
+	target_compile_options(libdecoders PRIVATE ${OPUSFILE_CFLAGS_OTHER})
+	target_link_libraries(libdecoders PRIVATE ${OPUSFILE_LIBRARIES})
+else()
+	find_package(OpusFile REQUIRED)
+	target_link_libraries(libdecoders PRIVATE OpusFile::opusfile)
+endif()
+
 target_link_libraries(libdecoders
 		PRIVATE
-		OpusFile::opusfile
 		$<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
 )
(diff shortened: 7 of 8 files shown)


src/capture/CMakeLists.txt
```diff
@@ -11,11 +11,18 @@ add_library(libcapture STATIC
...
-find_package(PNG CONFIG REQUIRED)
+if (PKG_CONFIG_FOUND)
+	pkg_check_modules(PNG REQUIRED IMPORTED_TARGET libpng)
+	target_include_directories(libcapture PRIVATE ${PNG_INCLUDE_DIRS})
+	target_compile_options(libcapture PRIVATE ${PNG_CFLAGS_OTHER})
+	target_link_libraries(libcapture PRIVATE ${PNG_LIBRARIES})
+else()
+	find_package(PNG CONFIG REQUIRED)
+	target_link_libraries(libcapture PRIVATE PNG::PNG)
+endif()
```
weirddan455, line 22: This shouldn't be needed for the majority of our dependencies.  CMake ships with a bunch of Find*.cmake files to make `find_package()` work for system deps.  In Debian, the libpng one is here for example `/usr/share/cmake-3.25/Modules/FindPNG.cmake`

I had to change this line from `find_package(PNG CONFIG REQUIRED)` to `find_package(PNG REQUIRED)`.  More info on the `CONFIG` parameter here:  https://cmake.org/cmake/help/latest/command/find_package.html#full-signature

@kklobe must have added `CONFIG` to his PR after I tested an early version of it.  If I'm reading the docs right, I think it effectively disables searching system libs so we may want to (conditionally?) remove that flag.

FeralChild64: I could use _CMake_ to find _libpng_, _OpenGL_, _SDL2_net_, and _ALSA_. Unfortunately, the flag has to stay - _iir_, _OpusFile_, _MT32Emu_, and _FluidSynth_ can't be found this way.

kklobe: I added `CONFIG` because that forces `find_package` to use the *Config.cmake or -config.cmake file that comes with the vcpkg dep instead of the generic one included with CMake, or finding the system-installed version, so I figured any package that we are installing with vcpkg should be configured with `CONFIG` so we are sure to pull it from vcpkg specifically.

weirddan455:
First problem I run into CMake is my version is too old.  I modified the first line of `CMakeLists.txt` changing from 3.26 to 3.25.  It looks like we have Ubuntu 22.04 runners in CI and they're likely running something even older.  I'd recommend dropping the required version.  It doesn't look like we actually need 3.26 for anything because at least mine builds fine on `main`.

I can't easily test the system dep build on Debian as it doesn't package all the needed dependencies.  IIR is the first error I run into and MT-32 is also required since that hasn't been added as an optional flag yet (it's marked TODO in CMakeLists.txt).

I'd need to spin up a Fedora VM to properly test this.  You're on Gentoo right @FeralChild64 ?  I guess they package everything? 

I'm probably just going to be using vcpkg for my personal dev builds going forward.  This PR caused a regression of a linker error for MT-32 and Fluidsynth for a vcpkg build.  I just tested that it builds fine on `main`.

I build by setting `VCPKG_ROOT` enviornment variable and then doing `cmake --preset=release ..` from inside my build dir.

```
[324/324] Linking CXX executable dosbox
FAILED: dosbox 
: && /usr/bin/c++ -O3 -DNDEBUG  CMakeFiles/dosbox.dir/src/main.cpp.o CMakeFiles/dosbox.dir/src/dosbox.cpp.o -o dosbox  vcpkg_installed/x64-linux/lib/libSDL2main.a  vcpkg_installed/x64-linux/lib/libSDL2.a  src/capture/liblibcapture.a  src/misc/liblibmisc.a  src/cpu/liblibcpu.a  src/dos/liblibdos.a  src/fpu/liblibfpu.a  vcpkg_installed/x64-linux/lib/libpng16.a  vcpkg_installed/x64-linux/lib/libz.a  src/libs/zmbv/libzmbv.a  vcpkg_installed/x64-linux/lib/libz-ng.a  src/hardware/liblibhardware.a  src/midi/liblibmidi.a  src/hardware/liblibhardware.a  src/midi/liblibmidi.a  vcpkg_installed/x64-linux/lib/libSDL2_net.a  src/libs/ESFMu/liblibesfmu.a  src/libs/residfp/liblibresidfp.a  src/libs/YM7128B_emu/liblibym7128bemu.a  src/libs/tal-chorus/liblibtalchorus.a  src/libs/nuked/liblibnuked.a  src/ints/liblibints.a  src/gui/liblibgui.a  src/misc/liblibmisc.a  src/libs/whereami/liblibwhereami.a  /usr/lib/x86_64-linux-gnu/libGLX.so  /usr/lib/x86_64-linux-gnu/libOpenGL.so  src/shell/liblibshell.a  vcpkg_installed/x64-linux/lib/libspeexdsp.a  -lmt32emu  -lfluidsynth  -lgthread-2.0  -lglib-2.0  -lpcre2-8  -lasound  -lpthread  vcpkg_installed/x64-linux/lib/libasound.a  -ldl  -lrt  src/audio/liblibaudio.a  src/libs/loguru/libloguru.a  src/libs/decoders/liblibdecoders.a  vcpkg_installed/x64-linux/lib/libSDL2.a  -pthread  -lopusfile  -logg  -lopus  -lm && :
/usr/bin/ld: cannot find -lmt32emu: No such file or directory
/usr/bin/ld: cannot find -lfluidsynth: No such file or directory
collect2: error: ld returned 1 exit status
```

johnnovak:
@weirddan455 @FeralChild64 Thanks to both of you for jumping on the CMake stuff, much appreciated.

To reiterate our high-level final state where we're going with all this:

- All CI builds will use vcpkg with pinned dep versions (no system libs) for the Windows, macOS, and Linux builds. That's our official and recommended build method where we 100% guarantee the correctness of our software. This "benchmark" build will eve be useful for 3rd repackage (e.g., Linux distro maintainers) as they can test their modified builds against it for correctness.-
 
- Adding _optional support_ for using system deps on macOS and Linux is a welcome addition. Some people might prefer it to using vcpkg and it might help distro packagers (although my understanding is they usually write their own make files from scratch anyway).

Just a heads up, using system deps will not be possible with all libraries. E.g., Blend2D (and AsmJIT which it depends on) are available via vcpkg, but not via package managers on Linux and macOS. So there needs to be a sort of a _third_ option for package resolution:

- If the optional "use system deps" flag is on (which can be thought of as "don't use vcpkg" 😆), then CMake will need to fetch the sources of the dependency via the build-in `FetchContent` call, build it, then statically link against it. More info in this [well-written article](https://www.foonathan.net/2022/06/cmake-fetchcontent/) on the subject.

My preference would be create the vcpkg Linux CI workflow first, and make sure it's working. That's our benchmark. Then you guys can add these extra system deps features, and the CI workflow will guard against regressions of the standard vcpkg path (like the one that apparently just hit @weirddan455).

weirddan455:
>  it might help distro packagers (although my understanding is they usually write their own build scripts from scratch anyway).

Not exactly.  They have their own scripts but they're mostly used for packaging.  Their scripts usually just call into upstream's build system as-is and they'll make small patches where they need to.

FeralChild64:
@weirddan455 Could you check if it compiles for you if you omit my last commit?

[Ubuntu package search](https://packages.ubuntu.com) claims they are providing 3.22.1 with _Ubuntu_ 22.04, so I'll decrease the required version - we'll see how it goes. I'll setup few VMs and try, but I'm rather short of time today.

Yes, I am using _Gentoo_ and it provides all our dependencies. The last time I tried I had no problems compiling _DOSBox Staging_ on _Ubuntu_, too.

@johnnovak I'm fine with the CI builds using whatever works, I can test with _vcpkg_ on a VM if necessary (as an additional step when messing with the build scripts) - but I am not using _vcpkg_ on my main _Linux_ system, period, this is non-negotiable.

_Gentoo_ already provides the [AsmJIT](https://packages.gentoo.org/packages/dev-libs/asmjit), there is an unofficial ebuild for [Blend2D](https://data.gpo.zugaina.org/guru/media-libs/blend2d/) already; I suppose the only reason it is not in the main tree is because nothing requires it yet.

You underestimate the _Linux_ distros :D

johnnovak:
> @johnnovak I'm fine with the CI builds using whatever works, I can test with _vcpkg_ on a VM if necessary (as an additional step when messing with the build scripts) - but I am not using _vcpkg_ on my main _Linux_ system, period, this is non-negotiable.

Cool, and I'm sure nobody in the team would object to you running vcpkg in a VM😏 If you make useful contributions to the build and CI scripts, you could run it on your Commodore 64 for all we care 😆Well, it runs in a VM too on GitHub CI, so really, who cares 🤷🏻 

> _Gentoo_ already provides the [AsmJIT](https://packages.gentoo.org/packages/dev-libs/asmjit), there is an unofficial ebuild for [Blend2D](https://data.gpo.zugaina.org/guru/media-libs/blend2d/) already; I suppose the only reason it is not in the main tree is because nothing requires it yet.

That's cool, but probably you'll need to maintain that source fetching CMake thing for a few libraries if you want to support your use system deps option for a wider range of distros. But anyway, it will be up to; I think you'll be the main user of the "use system deps" feature 😏

weirddan455:
>  I can test with vcpkg on a VM if necessary (as an additional step when messing with the build scripts) - but I am not using vcpkg on my main Linux system, period, this is non-negotiable.

I'm not going to try to convince you to install software you don't want to but I will point out that vcpkg is self-contained.  The official install instructions have you clone the repo from GitHub and run a bootstrap script.  The entire thing lives in a single folder in your /home directory, plus some caching in ~/.cache.

It doesn't pollute your /lib or anything.  You never even run it as root.

johnnovak:
> > I can test with vcpkg on a VM if necessary (as an additional step when messing with the build scripts) - but I am not using vcpkg on my main Linux system, period, this is non-negotiable.
> 
> I'm not going to try to convince you to install software you don't want to but I will point out that vcpkg is self-contained. The official install instructions have you clone the repo from GitHub and run a bootstrap script. The entire thing lives in a single folder in your /home directory, plus some caching in ~/.cache.
> 
> It doesn't pollute your /lib or anything. You never even run it as root.

Exactly. All the vcpkg stuff is in a single dir and in no way inteferes with the system. That's the point; it's completely isolated and self-contained.

weirddan455:
>  @weirddan455 Could you check if it compiles for you if you omit my last commit?

That doesn't work but removing the line `set(USE_SYSTEM_LIBS "${DOSBOX_PLATFORM_LINUX}")` (with all of your commits included) fixes the problem.

A possible solution could be making that variable a cache entry and setting/unsetting it in the presets so that it doesn't get set when using the vcpkg one.  Right now it's getting set for all Linux builds and even `cmake  -DUSE_SYSTEM_LIBS=false` doesn't work because it's not a cache variable.

https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry

FeralChild64:
One directory for _vcpkg_, one directory for it's cache, who knows how much disk space wasted, duplicated packages in the system, duplicated OS mechanism (and without any good reason), longer compilation time (think about the Raspberry Pi), more complex offline compilation, plus lower safety (yet another way to sideload packages = yet another attack vector).

So - why bother, if one can easily use the system libraries and get rid of all these problems? The _Linux_ community has already chosen not to use _vcpkg_ and I fully support this stance.

> I think you'll be the main user of the "use system deps" feature 😏

Plus the Linux distro devs, plus (indirectly) probably most of our _Linux_ users. I don't know what the others do - but I don't go to the _Mozilla_ website to get the _Firefox_ binaries, to the _OpenTTD_ website to get their binaries, to the _Darktable_ website to get their binaries, etc. I don't have time for all this and my OS is smart enough to handle this for me.

----

I have crated separate Linux presets for system libraries and _vcpkg_. The latter one is not yet tested, I need to setup a VM, preferably with the old LTS Ubuntu - not today.

kklobe:
I think we might be able to remove the USE_SYSTEM_LIBS manual machinery by removing the CONFIG option from all the `find_package` calls. In that case, it should search for modules instead of config files, and if not using `vcpkg`, it should find everything via system deps. For the presets that use vcpkg, we should then be able to force config file usage via `CMAKE_FIND_PACKAGE_PREFER_CONFIG`

kklobe:
> One directory for _vcpkg_, one directory for it's cache, who knows how much disk space wasted, duplicated packages in the system, duplicated OS mechanism (and without any good reason), longer compilation time (think about the Raspberry Pi), more complex offline compilation, plus lower safety (yet another way to sideload packages = yet another attack vector).

Aside from the majority of these being non-issues in reality ("I don't like it" is included as a non-issue), as long as we have strong Linux dev support to troubleshoot issues like "it's not working on this distro that has a 2-year old version of SDL or Fluidsynth", I don't have any objection to workarounds as long as it doesn't interfere with and clutter up the core CMake logic. The **entire point of this whole exercise** is to have a consistent, repeatable build across multiple platforms with consistent dependency versions, so someone on some platform isn't complaining about their joystick not working because the dev made the code change with SDL 2.30.1 and the user is linking against 2.28.3 or whatever.

weirddan455:
>  So - why bother, if one can easily use the system libraries and get rid of all these problems?

Well, mostly because it doesn't work for me :laughing:  I was having to side-load packages from Meson wraps anyway as not all distros package all of our dependencies.

>  The Linux community has already chosen not to use vcpkg and I fully support this stance.

That's because vcpkg is not a Linux package manager.  It's for developers and no different in concept than Cargo, NuGet, Pip, npm, etc.  Plenty of Linux developers use these tools even though they're not suitable for distro packaging.

>  Plus the Linux distro devs, plus (indirectly) probably most of our Linux users. I don't know what the others do - but I don't go to the Mozilla website to get the Firefox binaries, to the OpenTTD website to get their binaries, to the Darktable website to get their binaries, etc. I don't have time for all this and my OS is smart enough to handle this for me.

But yeah, I agree with you there.  IMO it is important to support system deps for distro packagers.  I use what's in my distro's repo wherever possible for programs I just want to use and not develop.

>  "it's not working on this distro that has a 2-year old version of SDL or Fluidsynth"

:100: The Fluidsynth version that Debian packages has a bug that makes the SC55 soundfont I was using sound off so that was another dep I was having to build from Meson wrap anyway.

johnnovak:
> That's because vcpkg is not a Linux package manager. It's for developers and no different in concept than Cargo, NuGet, Pip, npm, etc. Plenty of Linux developers use these tools even though they're not suitable for distro packaging.

100% this ☝🏻 It's a **build tool**; nobody claimed it was anything else. Adding support for using system deps is _orthogonal_ to using it (if those system deps are available, as @weirddan455 pointed out).

FeralChild64:
>  think we might be able to remove the USE_SYSTEM_LIBS manual machinery by removing the CONFIG option from all the find_package calls. In that case, it should search for modules instead of config files, and if not using vcpkg, it should find everything via system deps. For the presets that use vcpkg, we should then be able to force config file usage via CMAKE_FIND_PACKAGE_PREFER_CONFIG

Good idea! For some reason this does not work for the `OpusFile` library.

> The entire point of this whole exercise is to have a consistent, repeatable build across multiple platforms with consistent dependency versions

Well, I want to have a build system consistent with common practices on the host OS'es we support.

Consistent dependency versions are not that useful on _Linux_. Most end users won't be using the dependency versions selected by us anyway, they will be using versions selected by the distro developers - and these will change even during the lifetime of the given _DOSBox Staging_ release. Even if you try to pin the specific versions, the distros will simply patch your build system to use their system libraries - in part to be able to quickly patch vulnerabilities in dependencies, in part to decrease the build time by an order of magnitude.

Sure, we can have `vcpkg` for the official binaries - but these are not the binaries that are going to be used by the _Linux_ general public.

------

I see we have 3 major use cases:
1. Official release, statically linked build - `vcpkg` is to be used for (almost) everything.
2. The distro provides all the dependencies, or the build system user is willing to add missing dependencies in a distro-friendly way by himself - no `vcpkg` should be required.
3. Not all dependencies are available, the user is fine with `vcpkg` - a mixed approach should be used.

kklobe:
> Well, I want to have a build system consistent with common practices on the host OS'es we support.

The goal here is a build system that's minimal work to maintain across the three major platforms in terms of what us, the dev team, has the time, resources, and inclination to support. vcpkg isn't "common practice" on macOS, it's homebrew. @johnnovak likes macports. That's great, and we may even use those ourselves for day-to-day work, but any testing of release versions should be using the supported system. 

> 
> Consistent dependency versions are not that useful on _Linux_. Most end users won't be using the dependency versions selected by us anyway, they will be using versions selected by the distro developers - and these will change even during the lifetime of the given _DOSBox Staging_ release. Even if you try to pin the specific versions, the distros will simply patch your build system to use their system libraries - in part to be able to quickly patch vulnerabilities in dependencies, in part to decrease the build time by an order of magnitude.

Perfectly fine if someone wants to do that, but any compatibility issues due to differing dependency versions will have zero support - no trouble tickets, no issues, zip. @johnnovak let me know if that's too strongly worded.
> 
> Sure, we can have `vcpkg` for the official binaries - but these are not the binaries that are going to be used by the _Linux_ general public.

Also perfectly fine. Again, if an official vcpkg binary works fine, but someone pulling in whatever version of a local system dep runs into an issue, zero support.

> 
> I see we have 3 major use cases:
> 
> 1. Official release, statically linked build - `vcpkg` is to be used for (almost) everything.
> 2. The distro provides all the dependencies, or the build system user is willing to add missing dependencies in a distro-friendly way by himself - no `vcpkg` should be required.
> 3. Not all dependencies are available, the user is fine with `vcpkg` - a mixed approach should be used.

I agree with these, and I think the point John and I are trying to push is "supported". To belabor the point, you can deviate from the "official" build system and process, but you are 100% on your own, no PRs to make little tweaks for "this edge case on Debian 10 using an AMD GPU with Wayland because it has SDL 2.0.18 and there's a bug."

johnnovak:
This has all been a productive discussion so far, thanks to everyone for the input and for remaining objective. I'll start with some interesting remarks about the broader Linux situation on the emulator front, then I'll get to the summary and the way forward for us.

---

> Well, I want to have a build system consistent with common practices on the host OS'es we support.

I appreciate that, but I'd say that falls into the "personal preference" category; the situation is not so clear-cut (more details on that further below). To quote @kklobe's excellent summary, this is what we want:

> The goal here is a build system that's minimal work to maintain across the three major platforms in terms of what we, the dev team, have the time, resources, and inclination to support. vcpkg isn't "common practice" on macOS, it's homebrew. @johnnovak likes macports. That's great, and we may even use those ourselves for day-to-day work, but any testing of release versions should be using the supported system.

For the record, neither @kklobe or myself are particularly in love with vcpkg---we view it as a tool to get the job done with minimum fuss and complications, and that's it. We might all personally prefer a different tool, but vcpkg is objectively the single best well-supported and widely used **C++ build tool** available to achieve our goals across our three platforms.

---

> Consistent dependency versions are not that useful on Linux.
> Sure, we can have vcpkg for the official binaries - but these are not the binaries that are going to be used by the Linux general public.

Based on the research I've just done, and I'm not sure about that statement. I'm trying to be objective about this and base our discussion on real-world data points, so I just typed in "best linux emulators" into Google and this page came up on the top:

https://retrododo.com/best-emulators-for-linux/

Then I went and checked out the homepage of the "top 4" emulators the article recommended to see how they deliver their Linux versions:

1. [RetroArch](https://www.retroarch.com/?page=platforms) — Steam, itch.io, Snapcraft, Flatpak, AppImage
2. [Dolphin](https://dolphin-emu.org/download/?ref=btn) – Flatpak
3. [RPCS3](https://rpcs3.net/download) — AppImage, Flatpak
4. [PCSX2](https://pcsx2.net/downloads) — AppImage, Flatpak

These are all high-profile emulator projects. After this, I've checked out a few more prominent emulator projects I'm personally familiar with:

- [Duckstation](https://github.com/stenzek/duckstation) — AppImage, Flatpak
- [FS-UAE](https://fs-uae.net/download#linux) — distro-agnostic x86_64 tar.gz
- [Flycast](https://github.com/flyinghead/flycast) – Flatpak
- [Redream](https://redream.io/download) – distro-agnostic x64 tar.gz

So we have eight state-of-the-art, best-in-class, cream-of-the-crop emulators projects maintained by different teams here, and all deliver their official builds via distro-agonistic app images or binaries archives.  I'm pretty sure their reasons are similar to ours: reduce maintenance burden, ensure build consistency across platforms, and ensure correctness across various Linux distros. I've searched for them on https://pkgs.org/ and distro-level support is rather spotty, lagging behind, or non-existent.

I get what you're saying about using package managers on Linux as a standard mechanism for installing basic infrastructure, system libraries, command line tools, and even some GUI applications, but bypassing distro-level package managers appears to be standard practice for emulator projects on Linux. Users of emulator projects must therefore be used to this, and even expecting this (as it has benefits).

---

To be crystal clear about what we're going to do, here's a summary of the state of things so far.

1. **CMake + vcpkg is the build method for the official binaries on all platforms.** We're all in agreement on this, which is great, so moving on to the next point.

2. **Supported builds**. Clearly, we only "officially" support our official builds, and that's the end of it. Quoting @kklobe again:

> Perfectly fine if someone wants to do that, but any compatibility issues due to differing dependency versions will have zero support - no trouble tickets, no issues, zip. @johnnovak let me know if that's too strongly worded.
> you can deviate from the "official" build system and process, but you are 100% on your own, no PRs to make little tweaks for "this edge case on Debian 10 using an AMD GPU with Wayland because it has SDL 2.0.18 and there's a bug."

"Official support" is a bit of a weird thing to say in a hobby project like this; we're not paid employees. But what it means is that we're not making promises to fix any issues found in our non-official builds. Of course, members in our core team are free to go above and beyond this and offer support on a case-by-case basis at their own discretion—no one will stop them, it's their choice. But nobody can _expect_ we're gonna do this, that's the key point.

Example: if @FeralChild64  is on extended leave and someone complains about bugs in some Debian 10 distro-installed Staging version, I will ask them to re-test using our official Linux binaries. If the bug is not present in our official Linux build, or if they don't want to run the official binaries, I will simply close the issue ticket with "out of scope" and that's the end of it.

3. **Support for using system deps**. 

This also counts as an extra thing on top of our standard CMake + vcpkg build method. People might chose to use it because of personal preference, or to help distro maintainers to do their thing. Both are fine, but to be crystal clear about this: **neither me nor** @kklobe **are going to maintain these extras**, and they **cannot** introduce too much extra complexity into our standard CMake + vcpkg path. So small changes to make this happen are okay, but refactoring/rewriting/restructuring the whole CMake stuff is **definitely not**. If the solution is overly complex, it must go into `contrib/linux` or something as an alternative build method, completely removed from the `src` tree.

From this follows that neither me nor @kklobe particularly care about how people interested in maintaining the "use system deps" path will go about dealing with packages not available via the distro-level package managers. Using a mixed vcpkg + use system deps method is fine, grabbing the sources via `FetchContent` is fine too (that was just an idea of mine). We don't care, but it must **not** overly complicate the main CMake + vcpkg path. If it results in complications, into `contrib/linux` it goes.

From this, it should be also clear that maintaining the system deps option is **entirely** the responsibility of those who want this to be supported in the first place.

---

I know it was long, so thanks for reading until the end. Onwards and upwards! 🚀 😎

shermp:
Looking at the above list of emulators, perhaps we should adopt Flatpak as our official Linux compilation target? It gives us a known distro agnostic platform to target, and each version of the Freedesktop.org runtime is supported for 2 years.

johnnovak:
> Looking at the above list of emulators, perhaps we should adopt Flatpak as our official Linux compilation target? It gives us a known distro agnostic platform to target, and each version of the Freedesktop.org runtime is supported for 2 years.

I'm not against it, but one step at a time. Once we're on CMake + vcpkg and we've delivered a major release and a few patch releases using that method, we can think about next steps. But yeah, good point 👍🏻

weirddan455:
That's a pretty accurate summary of the Linux situation @johnnovak  Good write-up.

>  We also won't integrate the sources of these libraries into the source tree just because then things are easier for the system libs path; that's not an option.

I'm a bit confused about this point though.  We're already doing this for several libraries (everything in `src/libs`).  What are you saying our policy is for vendoring libraries like this?

>  Looking at the above list of emulators, perhaps we should adopt Flatpak as our official Linux compilation target? It gives us a known distro agnostic platform to target, and each version of the Freedesktop.org runtime is supported for 2 years.

Are you volunteering to maintain this?  I don't really like Flatpak personally but I wouldn't be opposed to shipping it either.  The issue with Flatpak is that you have to setup a dev environment to link against their runtime libraries.  We'd probably also need to add it to CI to automate the packaging process.  Lastly, the whole thing is Linux specific.

The advantage of vcpkg is that it works the same way, with the same library versions across all of our platforms.  We're pretty short on number of Linux developers to try and maintain two different build paths.  The system deps I'm more on board with because it's not "official".

johnnovak:
> I'm a bit confused about this point though. We're already doing this for several libraries (everything in `src/libs`). What are you saying our policy is for vendoring libraries like this?

I'm glad you've asked; I've been contemplating the same thing today 😄 

The vendored libraries in `src/libs` exist because most of those libs are rather niche, so no system deps or Meson ports exist for them. Vendoring them was the easiest and most practical thing to do when they were introduced (and I did introduce a few).

The new policy is very simple: only vendor something if it's not available through vcpkg. Moreover, we should gradually de-vendor what we can and use the vcpkg packages instead.

I quickly went through `src/libs` and this is my assessment. The following could be de-vendored as they are available via vcpkg:

- clap
- decoders/drlibs
- decoders/libflac
- decoders/libvorbis
- decoders/opus
- decoders/stb (for stb_vorbis)
- decoders/xxhash
- ghc
- enet
- loguru
- PDCurses
- simde

These must stay vendored as they're very niche and don't have vcpkg packages:

- ESFMu
- manymouse
- mverb
- nuked
- residfp
- tal-chorus
- whereami
- YM7128B_emu
- zmbv

rderooy:
There is already a dosbox-staging flatpak image, it is just not automatically built right now. I manually update the yaml manifest whenever a new version is released, which starts the build process.

Also regarding how few emulators are packaged by upstream distros, has a lot to do with emulators often requiring something that cannot be shipped by said distros, such as BIOS ROMs. And distros normally want something that at least works at some level without users having to get those pieces someplace.

FeralChild64:
@kklobe

> The goal here is a build system that's minimal work to maintain across the three major platforms in terms of what us, the dev team, has the time, resources, and inclination to support.

Exactly. I have no inclination to maintain a working _vcpkg_ install somewhere in my home directory.

> Perfectly fine if someone wants to do that, but any compatibility issues due to differing dependency versions will have zero support - no trouble tickets, no issues, zip.

It's your choice. If someone is using my software on a supported OS, in a way which is normal on this OS, I am not refusing to support this.

@johnnovak

> I appreciate that, but I'd say that falls into the "personal preference" category

It's not - automagically downloading dependencies during compilation is considered a security risk by probably every major Linux distro.

> It seems to me if the "Linux general public" wants to use these emulators, pretty much the official AppImages/Flatpaks and the distro-agnostic binaries are the way to go! Maybe devs can build their own, but that's not really "generic public" anymore 😏

It seems to you, but you have almost no real life experience with _Linux_. _Flatpaks_ do have serious issues. The GUI theming and many other settings do not work for the flatpack'ed apps, they can't access - for example - `/usr/share/soundfonts` to load your favorite soundfonts you set as a default for every piece of MIDI software you have, accessing mounted disks (for example, when you keep your photos/music/DOS games on a separate partition) is problematic, and so on. Sure, they are a reasonable way to provide distro-agnostic binaries (a step forward from just an archive to unpack), the format gives some safety to the end user (a big advantage) - but that's it.
Corporations behind _ChromeBooks_, _SteamDecks_, etc., probably love them, but they are rather controversial among the Linux community.

> must not overly complicate the main CMake + vcpkg path

Look at the PR. Do you think this even has a potential to overcomplicate the build scripts?

BTW, keep in mind that the whole effort (and even a bounty) **to get DOSBox Staging into Debian has failed**. The project I used to lead (technically I am still the leader there) **was taken into Debian without me even knowing**, despite needing a very exotic toolchain and having a rather bizarre _Makefile_. That's a fact.

------

Back to the topic - I have a problem using the _CMake+Ninja_ on _Ubuntu_ (both 22.04 and 24.04) - for both `vcpkg` and `non-vcpkg` presets:

```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
```

johnnovak:
> Look at the PR. Do you think this even has a potential to overcomplicate the build scripts?

In it's current form it's ok. Then it can go in many different ways depending on how ypu intend to solve the problem of dependencies that aren't available via system package managers.

I wrote the above to set expectations on what the project goals are, how will we build stuff, what is in scope, what do we support, etc. Will probably turn it into a wiki page and refer people to it if any future dispute comes up.

kklobe:
> Back to the topic - I have a problem using the _CMake+Ninja_ on _Ubuntu_ (both 22.04 and 24.04) - for both `vcpkg` and `non-vcpkg` presets:
> 
> ```
> CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
> CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
> CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
> -- Configuring incomplete, errors occurred!
> ```

What is the error text further up? I got these errors testing on WSL2 with both of those Ubuntu versions as well, and it always said I was missing something, like `pkg-config`, or `bison`, or `g++`, etc.:

```bash
CMake Error at scripts/cmake/vcpkg_find_acquire_program.cmake:166 (message):
  Could not find bison.  Please install it via your package manager:

      sudo apt-get install bison
Call Stack (most recent call first):
  buildtrees/versioning_/versions/gettext/1c1122632dae7ab9078655ee52e41c415ee4cfb4/portfile.cmake:43 (vcpkg_find_acquire_program)
  scripts/ports.cmake:196 (include)


error: building gettext:x64-linux failed with: BUILD_FAILED
See https://learn.microsoft.com/vcpkg/troubleshoot/build-failures?WT.mc_id=vcpkg_inproduct_cli for more information.
Elapsed time to handle gettext:x64-linux: 2.8 s
/source/src/vcpkg/commands.build.cpp(1723): unreachable code was reached
-- Running vcpkg install - failed
CMake Error at /home/kklobe/vcpkg/scripts/buildsystems/vcpkg.cmake:904 (message):
  vcpkg install failed.  See logs for more information:
  /home/kklobe/src/dosbox-staging/build/release/vcpkg-manifest-install.log
Call Stack (most recent call first):
  /usr/share/cmake-3.28/Modules/CMakeDetermineSystem.cmake:170 (include)
  CMakeLists.txt:5 (project)


CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
```

hubby4:
I managed to compile this on Raspberry Pi OS for my Pi 5. Cmake gave a bunch of warnings but there were no errors during the build. Unfortunately, I get a segfault error each time I try to run a game from a mounted CD-ROM image.

It's possible that I did something wrong on my end, but it's difficult to find out what it is if cmake isn't spitting out errors.

I'll stick with meson for now.

FeralChild64:
@kklobe Indeed, I haven't notice it. Developing late in the evening... I had to install quite a few packages:

```
pkg-config
bison
autoconf
libtool
libgl-dev
libopusfile-dev
```

But now I am stuck with _opusfile_:

```
CMake Error at /home/roman/test-vcpkg/vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package):
  By not providing "Findopusfile.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "opusfile",
  but CMake did not find one.

  Could not find a package configuration file provided by "opusfile" with any
  of the following names:

    opusfileConfig.cmake
    opusfile-config.cmake

  Add the installation prefix of "opusfile" to CMAKE_PREFIX_PATH or set
  "opusfile_DIR" to a directory containing one of the above files.  If
  "opusfile" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  src/libs/decoders/CMakeLists.txt:19 (find_package)
```

(on _Ubuntu_ I am testing the _release-linux-vcpkg_ preset; I can't compile it without _vcpkg_, as it Ubuntu does not provide _libiir_, or maybe I don't know where to find it)

FeralChild64:
@johnnovak Handling outdated dependencies is not our problem - it's the distro maintainer's job to provide the minimal required version.
On the rolling release distro (like _Gentoo_) you are more likely to end up with _DOSBox Staging_ running with dependencies newer than the one which were available during our release.

If a dependency is not available in _Gentoo_, for my development purposes I'll simply add it by writing a custom quick&dirty ebuild - they are really easy to create, _Gentoo_ is very tinker-friendly.
The target solution (adding the dependency to the distro) is a distro maintainer's responsibility (if they want to package _DOSBox Staging_, of course).

weirddan455:
>  But now I am stuck with opusfile:

You broke the casing in this PR.  This fixes it:

```
--- a/src/libs/decoders/CMakeLists.txt
+++ b/src/libs/decoders/CMakeLists.txt
@@ -16,7 +16,7 @@ if (USE_SYSTEM_LIBS)
        target_compile_options(libdecoders PRIVATE ${OPUSFILE_CFLAGS_OTHER})
        target_link_libraries(libdecoders PRIVATE ${OPUSFILE_LIBRARIES})
 else()
-       find_package(opusfile REQUIRED)
+       find_package(OpusFile REQUIRED)
        target_link_libraries(libdecoders PRIVATE OpusFile::opusfile)
 endif()

```

>  I can't compile it without vcpkg, as it Ubuntu does not provide libiir, or maybe I don't know where to find it

I'm in the same boat on Debian.  It's probably not in Ubuntu either as they largely pull from Debian's repos.

@hubby4 I can't reproduce the segfault on CDROM images.  You'd need to put in a seperate ticket with reproduction instructions for us to look into it.  Otherwise, stick with Meson.  CMake should still be considered experimental.

eli-schwartz:
> > It's not - automagically downloading dependencies during compilation is considered a security risk by probably every major Linux distro.
>
> Java, Scala, Kotlin, Python, Rust, Go, Nim, NodeJS, etc. (the list could go on) devs on Linux want to have a word with you 😆

Devs of any variety tend to have a high tolerance for insecurity, which is why they keep installing malware because a blog article suggested to do `curl https://coolsoftware.com/install | sudo bash`. This is particularly notable in some ecosystems such as NodeJS.

Python represents a weird spot in this discussion because the entire python ecosystem effectively requires you install software including dependencies the same way whether you do it with a tool that downloads dependencies at install time, or package it via a linux distro. 

Java is uhhhhh not something you want to bring up if you're trying to defend your point of view, sorry. :D

golang and rust both have checksummed lockfiles, which decreases the risk of downloading dependencies during compilation -- because you can vet those and guarantee that they are what they claim to be. golang recently made the news anyway: https://www.theregister.com/2025/02/04/golang_supply_chain_attack/

Linux distros are relied on by a lot more than any one given program, so there's a lot of scope for security failures to cause real damage. And with a robust legacy of caring a lot about security and authenticity of the software being shipped, distros also tend to be significantly more trusted by users to simply "install anything you want from the distro repository, it's sure to be safe" -- and that necessitates taking said responsibility seriously. So yes, Linux distros practice basic security hygiene a lot more than "devs on Linux".

...

On the topic of unpredictable and inconsistent dependencies, I will never understand why projects are terrified of declaring minimum supported requirements and actually enforcing those via `>=X.Y` scoped dependency lookups in their build system, and yet somehow think the solution is to only use vendored dependencies that are newer. If you require SDL 2.30.1 but the user tries to use a stone-age distro which only ships 2.0.18, then just... error out when configuring the build, tell them they need to somehow provide 2.30.1. ***how*** they provide 2.30.1 is irrelevant (running vcpkg install before configuring the project is "just another source of libs", looking up the system copy if it is suitable is inherently eminently reasonable).

Same thing really for totally unpackaged dependencies. It's *convenient* for some developer-only tooling to provide a way to rapidly install an obscure library not yet packaged in distros, but anyone who wants to package the software will naturally package the dependencies too. Nobody will say "gosh, it was impossible to package dosbox-staging because it had dependencies which aren't packaged, but now that you download the dependency at compile time I can package it".

They will just package the dependency...

Being overly reliant on "how" precisely, you get a given dependency, is... a code smell. Sorry. :) (But it is a very *cmake* kind of code smell, to be sure, considering the immense pain of performing an action such as "locate dependency XYZ" which requires you to know whether it is a) a subproject, b) provided by FindFoo.cmake, c) provided by fooConfig.cmake, d) provided by pkg_check_modules, e) provided by `foo-config --cflags --libs`.)

FeralChild64:
@weirddan455 Thanks. This, plus a fix to build presets, and I was able to build on Ubuntu using `vcpkg`. I intend to write a build instruction, assuming that plain Ubuntu (stock installation) is used as a starting point.

If you agree, I'll create a tracking issue with a checklist of all the problems which has to be solved before we declare the _CMake_ support as production-ready - so we won't forget about anything.

FeralChild64:
@weirddan455, @johnnovak, @kklobe Ready for re-review.

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+- CMake
+- pkg-config
+
+1. Install the dependencies, development packages are needed too:
+
+- SDL 2.x
+- SDL2_net
+- IIR
+- OpusFile
+- MT32Emu
```
johnnovak, line 376: Typo: lowercase L

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+- FLuidSynth
+- ALSA
+- libpng
+- OpenGL headers
+
+1. Clone the DOSBox Staging:
```
johnnovak, line 382: Typo: lose "the"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+
+```bash
+git clone https://github.com/dosbox-staging/dosbox-staging.git
+```
+
+1. Configure the sources and build the DOSBox:
```
johnnovak, line 388: "and build DOSBox Staging"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+
+```bash
+cd dosbox-staging
+cmake --preset=release-linux
+cmake --build --preset=release-linux
+```
+
+1. You can now launch DOSBox with command:
```
johnnovak, line 396: "the command"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+```
+
+- set the vcpkg path in your compile shell (it is recommended to add the command
+  to your shell startup script, usually `~/.bashrc`)
+
+```bash
+export VCPKG_ROOT=$HOME/vcpkg
+```
+
+1. Clone the DOSBox Staging into your home directory:
```
johnnovak, line 436: "Clone DOSBox Staging"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+
+```bash
+cd ~
+git clone https://github.com/dosbox-staging/dosbox-staging.git
+```
+
+1. Configure the sources and build the DOSBox:
```
johnnovak, line 443: "build DOSBox Staging"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+
+```bash
+cd dosbox-staging
+cmake --preset=release-linux-vcpkg
+cmake --build --preset=release-linux-vcpkg
+```
+
+1. You can now launch DOSBox with command:
```
johnnovak, line 451: "the command"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+
+> **Note**
+>
+> This is experimental, internal-only, work-in-progress feature.
+> 
+> **PLEASE DO NOT SUBMIT ANY BUGS OR HELP REQUESTS!**
+
+### Building on Linux, using system dependencies
+
+1. Install necessary build tools:
```
johnnovak, line 363: "the necessary"

BUILD.md
johnnovak, line 382: Typo: lose "the"

BUILD.md
```diff
@@ -355,3 +349,107 @@ will leave your normal `build/` files untouched.
...
+- pkg-config
+
+1. Install the dependencies, development packages are needed too:
+
+- SDL 2.x
+- SDL2_net
+- IIR
+- OpusFile
+- MT32Emu
+- FLuidSynth
```
johnnovak, line 377: Typo: lowercase L

johnnovak (approved):
Changes looks fine to me, clean and simple 👍🏻 Well done, @FeralChild64 🎉 Just a few minor comments about typos/grammar in the doco.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

BUILD.md
FeralChild64: Fixed.

PR #4187, opened 2025-02-09
Document Microsoft Digital Media Pro Keyboard scancodes

FeralChild64:
# Description

I briefly got my hands on a retro keyboard (photos below) - I've used the keyboard testing tool from the _DosLib_ package (a _DOSBox-X_ sideproject) to capture the scancodes of all the extra keys and added them to _DOSBox Staging_.

These scancodes are not currently mapped to the SDL keyboard kodes - we are not emulating multimedia keyboards yet, as these are only useful starting from _Windows 9x_. But I would like to keep the codes documented - 10 years from now it might be much harder to get this information.

IMG_1547

IMG_1548

# Release notes

(none - no user visible changes)

# Manual testing

Checked that the keyboard input still works - the change only touches constants which are currently not used for anything.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/include/keyboard.h
+++ b/include/keyboard.h
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2022-2023  The DOSBox Staging Team
+ *  Copyright (C) 2022-2025  The DOSBox Staging Team
  *  Copyright (C) 2002-2021  The DOSBox Team
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


weirddan455:
I'm confused.  So this PR has no functional changes and we also currently have no plans to make these keys functional?  In general, I don't really like using active code for documentation purposes because then someone comes along later to work on keyboard code and says "Why are all these keys not working?  Must be a bug."  Or someone needs to do a refactor and we have to shuffle a bunch of dead code around.

On the other hand, it looks like there's already a bunch of these media keys already in the code so this PR doesn't change the situation much.  Are these existing media keys also non-functional?

FeralChild64:
No plans for the near future - but once we have a GUI, a new mapper, and _Windows 9x_ becomes a higher priority goal (I believe this will happen someday), we will need this knowledge to emulate the multimedia keys.

OK, I'll put this data into the comment and remove the unneeded enum values, this is not a problem.

johnnovak:
I'm with @weirddan455 on this; we should not use the code as a repository for random stuff we might need one day. That day might never come 😆 

Scan codes for all sorts of keyboards like the Microsoft multimedia keyboards are documented online:
https://www.scs.stanford.edu/10wi-cs140/pintos/specs/kbd/scancodes-5.html

And of course all these scancodes are listed in the Microsoft documentation: 
https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes

Moreover, you can always refer to other emulators that support Win95 and later for a list of these codes (PCem, 86box, etc.). Information like this won't be "lost".

I'd actually like to in the _opposite_ direction; remove all references of scancodes not used by DOS, and only re-add them if the project officially supports Win98 (and that's a big "if").

> On the other hand, it looks like there's already a bunch of these media keys already in the code so this PR doesn't change the situation much. Are these existing media keys also non-functional?

Yeah, and I remember when @FeralChild64 introduced those scancodes a few years back, I had mixed feelings about the whole thing (same reasoning: the code should not be a repository of random knowledge we don't currently need; there are other ways to store that information outside of the codebase).

FeralChild64:
The hardware/input directory already contains a `README.md` file. Would it be OK if I just move all the extra scancode information there?

@johnnovak

The _Microsoft_ page you have linked contains Windows internal identifiers, these are no use for us - PS/2 hardware scancodes are different, they also differ from the USB keyboard scancodes.

The _Stanford_ page contains a lot of scancodes for different keyboard, but it does not even tell whether these are for scancode set 1 or scancode set 2 (set 3 is probably not widely used) - _Microsoft_ keyboard codes are clearly for scancode set 1, while the set 2 codes are missing.

The last time I checked, `86Box` did not support multimedia keyboards at all.

Before suggest this - Linux kernel source code is probably also not a 100% reliable source of information; I have tried to implement a dual-wheel mouse support based on the information found there, but I definitely missed something, I couldn't get the _A4Tech_ drivers for _Windows 3.1x_ to talk with my emulation code.

johnnovak:
> The hardware/input directory already contains a `README.md` file. Would it be OK if I just move all the extra scancode information there?

Moving all unused special keys from the code into that `README.md` is fine, just please make it clear in the doco that this is only some information for the future.

FeralChild64:
@weirddan455, @johnnovak Ready for re-review.

johnnovak (approved):
Thanks for moving these out, I'm fine with this 👍🏻

PR #4188, opened 2025-02-10
Add bounds checking to SoundBlaster's read_dma_16bit() and read_dma_8bit()

weirddan455:
# Description

Regression from 274cd6c50a3f8443cab2e653db3cfc94607f0c31
That commit allowed for larger DMA reads.
I had assumed the DMA code did bounds checking but it did not.

This caused a buffer overflow and segfault in some cases.
I was only able to reproduce the crash by disabling PulseAudio and using pure ALSA.

# Release notes

Regression was never in a release.

# Manual testing

Crash occurred in Pinball Dreams Delux, when sound starts playing.  If not at the main menu, then as soon as you entered a level.

I was only able to reproduce it by using pure ALSA with these commands:

`systemctl --user mask pulseaudio.socket` (to prevent applications from automatically re-lauching it)
`systemctl --user stop pulseaudio`

With `blocksize = 128` it only crashed about 1 in 10 launches.  With `blocksize = 512` (our default), it crashed nearly 100% of the time.

I launched it about 20-30 times after this PR to confirm it's fixed.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -1011,21 +1011,69 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
 	return frames;
 }
 
-static uint32_t read_dma_8bit(const uint32_t bytes_to_read, const uint32_t i = 0)
+static uint32_t read_dma_8bit(const uint32_t bytes_to_read, const uint32_t buffer_index = 0)
 {
-	const auto bytes_read = sb.dma.chan->Read(bytes_to_read, sb.dma.buf.b8 + i);
-	assert(bytes_read <= DmaBufSize * sizeof(sb.dma.buf.b8[0]));
+	static_assert(sizeof(sb.dma.buf.b8[0]) == 1);
+
+	if (buffer_index >= DmaBufSize) {
+		// Should never happen as the code is currently written.
+		// Calling code has buffer_index either 0 or 1 to handle a dangling sample from the last read.
+		// This is to solve an edge case for stereo sound when the DMA buffer has an odd number of samples.
+		assertm(false, "SBLASTER: Read requested out of bounds of sb.dma.buf.b8");
+		return 0;
+	}
+
+	// Assert that the DMA controller is configured for 8-bit reads,
+	// so we don't have to do the whole song and dance we do for 16-bit.
+	// We should always be dealing with bytes here.
+	assertm(!sb.dma.chan->is_16bit, "SBLASTER: read_dma_8bit() called but DMA controller is configured for 8-bit reads");
+
+	const uint32_t bytes_available = DmaBufSize - buffer_index;
+	const uint32_t clamped_bytes = std::min(bytes_to_read, bytes_available);
+	const auto bytes_read = sb.dma.chan->Read(clamped_bytes, sb.dma.buf.b8 + buffer_index);
+	assert(bytes_read <= clamped_bytes);
 
 	return check_cast<uint32_t>(bytes_read);
 }
 
-static uint32_t read_dma_16bit(const uint32_t bytes_to_read, const uint32_t i = 0)
+static uint32_t read_dma_16bit(const uint32_t words_to_read, const uint32_t buffer_index = 0)
 {
-	const auto unsigned_buf = reinterpret_cast<uint8_t*>(sb.dma.buf.b16 + i);
-	const auto bytes_read = sb.dma.chan->Read(bytes_to_read, unsigned_buf);
-	assert(bytes_read <= DmaBufSize * sizeof(sb.dma.buf.b16[0]));
+	static_assert(sizeof(sb.dma.buf.b16[0]) == 2);
 
-	return check_cast<uint32_t>(bytes_read);
+	if (buffer_index >= DmaBufSize) {
+		// Should never happen as the code is currently written.
+		// Calling code has buffer_index either 0 or 1 to handle a dangling sample from the last read.
+		// This is to solve an edge case for stereo sound when the DMA buffer has an odd number of samples.
+		assertm(false, "SBLASTER: Read requested out of bounds of sb.dma.buf.b16");
+		return 0;
+	}
+
+	// In dma.cpp, if sb.dma.chan->is_16bit is set, we're dealing with 16-bit words.
+	// Otherwise, we're dealing with 8-bit words (bytes).
+	// Calling code handles this case and conditionally divides by two.
+	uint32_t bytes_requested = words_to_read;
+	if (sb.dma.chan->is_16bit) {
+		bytes_requested *= 2;
+		// Calling code uses these internal SoundBlaster variables.
+		// sb.dma.chan->is_16bit is the source of truth in terms of memory safety.
+		// Assert that these two always match up.
+		assertm(sb.dma.mode == DmaMode::Pcm16Bit, "SoundBlaster expected 8-bit read but DMA controller is 16-bit");
+	} else {
+		assertm(sb.dma.mode == DmaMode::Pcm16BitAliased, "SoundBlaster expected 16-bit read but DMA controller is 8-bit");
+	}
+
+	// Clamp words to read so we don't overflow our buffer.
+	const uint32_t bytes_available = (DmaBufSize - buffer_index) * 2;
+	uint32_t clamped_words = std::min(bytes_requested, bytes_available);
+	if (sb.dma.chan->is_16bit) {
+		clamped_words /= 2;
+	}
+
+	const auto unsigned_buf = reinterpret_cast<uint8_t*>(sb.dma.buf.b16 + buffer_index);
+	const auto words_read = sb.dma.chan->Read(clamped_words, unsigned_buf);
+	assert(words_read <= clamped_words);
+
+	return check_cast<uint32_t>(words_read);
 }
 
 static void enqueue_frames(std::vector<AudioFrame>& frames)


src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
 
 static uint32_t read_dma_8bit(const uint32_t bytes_to_read, const uint32_t i = 0)
 {
-	const auto bytes_read = sb.dma.chan->Read(bytes_to_read, sb.dma.buf.b8 + i);
-	assert(bytes_read <= DmaBufSize * sizeof(sb.dma.buf.b8[0]));
+	static_assert(sizeof(sb.dma.buf.b8[0]) == 1);
+
+	if (i >= DmaBufSize) {
+		// Should never happen as the code is current written.
```
johnnovak, line 1019: typo: currently

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+		// Calling code has i either 0 or 1 to handle a dangling sample from the last read.
```
johnnovak, line 1020: Could we rename the "i" arg to something more meaningful?

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+		assertm(false, "SBLASTER: Read requested out of bounds of sb.dma.buf.b8");
+		return 0;
+	}
+
+	// Assert that the DMA controller is configured for 8 bit reads.
```
johnnovak, line 1025: typo: probably comma at the end of line?
Also, "8-bit" and "16-bit".

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+	// so we don't have to do the whole song and dance we do for 16 bit.
+	// We should always be dealing with bytes here.
+	assertm(!sb.dma.chan->is_16bit, "SBLASTER: read_dma_8bit() called but DMA controller is configured for 8 bit reads");
```
johnnovak, line 1028: 8-bit

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+
+	const uint32_t bytes_available = DmaBufSize - i;
+	const uint32_t clamped_bytes = std::min(bytes_to_read, bytes_available);
+	const auto bytes_read = sb.dma.chan->Read(clamped_bytes, sb.dma.buf.b8 + i);
```
johnnovak, line 1032: C int promotion... oh man 🥶

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
 	return check_cast<uint32_t>(bytes_read);
 }
 
-static uint32_t read_dma_16bit(const uint32_t bytes_to_read, const uint32_t i = 0)
+static uint32_t read_dma_16bit(const uint32_t words_to_read, const uint32_t i = 0)
 {
+	static_assert(sizeof(sb.dma.buf.b16[0]) == 2);
+
+	if (i >= DmaBufSize) {
+		// Should never happen as the code is current written.
```
johnnovak, line 1043: Use a better arg name than "i"?

johnnovak, line 1043: typo: currently

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+		// Calling code has i either 0 or 1 to handle a dangling sample from the last read.
+		assertm(false, "SBLASTER: Read requested out of bounds of sb.dma.buf.b16");
+		return 0;
+	}
+
+	// In dma.cpp, if sb.dma.chan->is_16bit is set, we're dealing with 16-bit words.
+	// Otherwise, we're dealing with 8 bit words (bytes).
```
johnnovak, line 1050: typo: 8-bit

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+	// Calling code handles this case and conditionally divides by two.
+	uint32_t bytes_requested = words_to_read;
+	if (sb.dma.chan->is_16bit) {
+		bytes_requested *= 2;
+		// Calling code uses these internal SoundBlaster variables.
+		// sb.dma.chan->is_16bit is the source of truth in terms of memory safety.
+		// Assert that these two always match up.
+		assertm(sb.dma.mode == DmaMode::Pcm16Bit, "SoundBlaster expected 8 bit read but DMA controller is 16 bit");
+	} else {
+		assertm(sb.dma.mode == DmaMode::Pcm16BitAliased, "SoundBlaster expected 16 bit read but DMA controller is 8 bit");
```
johnnovak, line 1060: Very nice assertion failure messages 🎉 
typo: 8-bit, 16-bit

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+	}
+
+	// Clamp words to read so we don't overflow our buffer.
+	const uint32_t bytes_available = (DmaBufSize - i) * 2;
```
johnnovak, line 1064: Yeah, so if "i" had a meaningful name, we'd actually understand what's going on here 😆

src/hardware/sblaster.cpp
```diff
@@ -1013,19 +1013,65 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
+	uint32_t clamped_words = std::min(bytes_requested, bytes_available);
+	if (sb.dma.chan->is_16bit) {
+		clamped_words /= 2;
```
johnnovak, line 1067: Assertion opportunity: check that `clamped_words` is even?

johnnovak (approved):
Good thing you ran into this by chance and fixed it @weirddan455 🎉 🏅  Just a few minor comments, otherwise looks all good 👍🏻

src/hardware/sblaster.cpp
weirddan455: Let me re-visit this.  It's not a problem if there's an odd number of bytes available in the buffer as long as we're doing mono sound.  We just round down the number of 16-bit words to read.

I think the calling code has a bug for stereo though where it could play silence in one channel for a frame if this happens.  And then also potentially invert the channels the next time it runs due to this weird "dangling sample" edge case.

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: They're all `uint32_t`.  I don't think there's any promotion happening here?

src/hardware/sblaster.cpp
weirddan455: fixed

weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
weirddan455: fixed

src/hardware/sblaster.cpp
johnnovak: Yeah, I think you're right. If one of the operands is unsigned, then unsigned "wins" (for mixed signed/unsigned arithmetic). But this is pure unsigned, so that doesn't even come into the picture.

This is a good description of the main algo:

https://learn.microsoft.com/en-us/cpp/c-language/usual-arithmetic-conversions?view=msvc-170

src/hardware/sblaster.cpp
johnnovak: > The "dangling sample" code handles exactly this case. I verified it always enqueues whole frames into the mixer. If DMA returns an odd number of samples, it moves that odd sample to the front of the buffer so it gets used on the next call.

Is there a comment explaining this where the dangling sample is added? If not, could you add a short version of the above explanatory comment there? It would be super useful for the next guy doing a deep dive into the mixer code 🤿 🌊 

>  This seems unlikely to ever happen to me but who knows there's probably a weird demo out there that does it 😆

I'd bet good money on that! 😂

src/hardware/sblaster.cpp
```diff
@@ -1011,21 +1011,67 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
-static uint32_t read_dma_8bit(const uint32_t bytes_to_read, const uint32_t i = 0)
+static uint32_t read_dma_8bit(const uint32_t bytes_to_read, const uint32_t buffer_index = 0)
 {
-	const auto bytes_read = sb.dma.chan->Read(bytes_to_read, sb.dma.buf.b8 + i);
-	assert(bytes_read <= DmaBufSize * sizeof(sb.dma.buf.b8[0]));
+	static_assert(sizeof(sb.dma.buf.b8[0]) == 1);
+
+	if (buffer_index >= DmaBufSize) {
+		// Should never happen as the code is currently written.
+		// Calling code has buffer_index either 0 or 1 to handle a dangling sample from the last read.
```
johnnovak, line 1020: It would be also useful to explain here briefly the problem that "dangling sample" solves (padding stereo sample streams).

src/hardware/sblaster.cpp
weirddan455: Yeah there's a comment already.  I just didn't see it at first:  https://github.com/dosbox-staging/dosbox-staging/blob/33a83fb77173a590b6a89be35bc22cc6bb867806/src/hardware/sblaster.cpp#L1120-L1138

src/hardware/sblaster.cpp
weirddan455: done

src/hardware/sblaster.cpp
```diff
@@ -1011,21 +1011,67 @@ static std::vector<AudioFrame>& maybe_silence(const T* samples,
...
 {
-	const auto unsigned_buf = reinterpret_cast<uint8_t*>(sb.dma.buf.b16 + i);
-	const auto bytes_read = sb.dma.chan->Read(bytes_to_read, unsigned_buf);
-	assert(bytes_read <= DmaBufSize * sizeof(sb.dma.buf.b16[0]));
+	static_assert(sizeof(sb.dma.buf.b16[0]) == 2);
 
-	return check_cast<uint32_t>(bytes_read);
+	if (buffer_index >= DmaBufSize) {
+		// Should never happen as the code is currently written.
+		// Calling code has buffer_index either 0 or 1 to handle a dangling sample from the last read.
```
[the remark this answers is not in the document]
weirddan455: done

PR #4189, opened 2025-02-10
Add WIN32 compile definition to CMake

weirddan455:
# Description

I got Staging to build without the Visual Studio generator or msbuild at all.  I installed native Windows versions of CMake, Ninja, and LLVM and it works just like in Linux from the command line.  I'm using the same `release` template that uses Ninja.

Only small problem I ran into was some compile errors that `WIN32` is not defined.  I think the Windows headers have it defined as `__WIN32__` or something.  We were manually defining it in Meson.

I think that `set(WIN32, TRUE)` was trying to do this but that only sets a variable for use in CMake itself and I didn't see any other usage of `WIN32` in the build scripts.

# Manual testing

It builds!  It runs Doom!

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ include(CheckSymbolExists)
 
 if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
 	set(DOSBOX_PLATFORM_WINDOWS ON)
-	set(WIN32 ON)
+	add_compile_definitions(WIN32)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
 	set(DOSBOX_PLATFORM_MACOS ON)
 	set(MACOSX ON)


kklobe:
I think that will change it from getting set in the `config.h` file as `#define WIN32 1` to just a compiler-level definition. Those are used for the DOSBox code rather than trying to replace those compiler-level __WIN32__ definitions. There are some places where the source code does an if test, e.g. `program_intro.cpp`:

```cpp
void INTRO::DisplayMount(void) {
    /* Basic mounting has a version for each operating system.
        * This is done this way so both messages appear in the language file*/
    WriteOut(MSG_Get("PROGRAM_INTRO_MOUNT_START"));
#if (WIN32)
    WriteOut(MSG_Get("PROGRAM_INTRO_MOUNT_WINDOWS"));
#else
    WriteOut(MSG_Get("PROGRAM_INTRO_MOUNT_OTHER"));
#endif
    WriteOut(MSG_Get("PROGRAM_INTRO_MOUNT_END"));
}
```

If we do make the change, we should probably do it for all the platforms and make sure any references in the code get changed to #if defined() instead.

```cmake
if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
	set(DOSBOX_PLATFORM_WINDOWS ON)
	set(WIN32 ON)
elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
	set(DOSBOX_PLATFORM_MACOS ON)
	set(MACOSX ON)
elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
	set(DOSBOX_PLATFORM_LINUX ON)
	set(LINUX ON)
else()
	message(FATAL_ERROR "Unknown system ${CMAKE_SYSTEM_NAME}")
endif()
```

weirddan455:
>  I think that will change it from getting set in the config.h file as #define WIN32 1 to just a compiler-level definition.
>
>  EDIT: nevermind. It appears that the generated config.h ends up with the same WIN32 define for both main and the branch.

Ok, I forgot to look in `config.h.in.cmake`.  It's doing this:

```
#ifndef WIN32
#cmakedefine WIN32
#endif
```

Where `cmakedefine` does use the CMake `set(WIN32 TRUE)` var to conditionally define for the C pre-processor.

>  There are some places where the source code does an if test

And I think those were the only ones that were failing for me, now that you mention it.  I think what's happening is Visual Studio does a `#define WIN32 1` or something and this config.h does a `#define WIN32` if the former is not already defined.

So a better fix could be changing `#cmakedefine WIN32` to `#cmakedefine WIN32 1` maybe?  Or changing all the `#if WIN32` to `#ifdef WIN32`?  It always trips me up that those 2 are not identical.

kklobe:
> So a better fix could be changing `#cmakedefine WIN32` to `#cmakedefine WIN32 1` maybe? Or changing all the `#if WIN32` to `#ifdef WIN32`? It always trips me up that those 2 are not identical.

Let's leave it as a raw define - that's how it's done in the Windows SDK, in `minwindef.h` (which gets pulled in via `windows.h`):

```cpp
#ifndef WIN32
#define WIN32
#endif
```

And then, yeah, let's get convert any `#if (WIN32)` to `#ifdef` - I did this for `#if (MACOSX)` on my initial PR as well.

kklobe:
After thinking about this for a bit, I think this is a case of "I like to use this non-standard toolchain but I have to tweak the mainline CMake config". This is something that's a good fit for `CMakeUserPresets.json`, where you can add a macro/compiler define or whatever if said toolchain doesn't support it. I specifically added `CMakeUserPresets.json` to the `.gitignore`, and that seems like a good answer to using personal preference toolchains.

@johnnovak thoughts on this?

weirddan455:
Closing this.  See #4191

johnnovak:
> After thinking about this for a bit, I think this is a case of "I like to use this non-standard toolchain but I have to tweak the mainline CMake config". This is something that's a good fit for `CMakeUserPresets.json`, where you can add a macro/compiler define or whatever if said toolchain doesn't support it. I specifically added `CMakeUserPresets.json` to the `.gitignore`, and that seems like a good answer to using personal preference toolchains.
> 
> @johnnovak thoughts on this?

@kklobe @weirddan455 
I was gonna say, easy with these customisations, otherwise we'll end up with 5+ different build methods in no time again 😅 But @weirddan455 was able to solve it unobtrusively in his other PR, so all is well if it ends well 😎

PR #4191, opened 2025-02-10
Replace #if WIN32 with #ifdef WIN32

weirddan455:
# Description

Fixes a compile error when building with CMake + Ninja

My hunch I mentioned in #4189 was right.  Visual Studio inserts a preprocessor option to define `WIN32`.  It would be like compiling with `clang -DWIN32` which according to the Clang docs will define it to 1. "Define <macro> to <value> (or 1 if <value> omitted)"

Ninja does not insert this flag so it falls back to our `config.h` which defines it to nothing.

Most of these have been converted already.  After fixing these last 3 instances, it compiles successfully.

# Manual testing

Built with CMake + Ninja

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/program_intro.cpp
+++ b/src/dos/program_intro.cpp
@@ -57,7 +57,7 @@ void INTRO::DisplayMount(void) {
     /* Basic mounting has a version for each operating system.
         * This is done this way so both messages appear in the language file*/
     WriteOut(MSG_Get("PROGRAM_INTRO_MOUNT_START"));
-#if (WIN32)
+#ifdef WIN32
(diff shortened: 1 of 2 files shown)


johnnovak (approved):
Nice one @weirddan455 🎉 

As I said in my [other comment](https://github.com/dosbox-staging/dosbox-staging/pull/4189#issuecomment-2649561027), we should be wary of adding 5+ new alternative CMake build methods 😅, but this is a clean way to address the problem and an improvement anyway. Well done! 🎉 

(Ignore the PVS warning; CDN issue.)

PR #4195, opened 2025-02-14
Mute audio when the mapper is active

johnnovak:
# Description

Simple fix to address the regression where the audio is not muted while the mapper is open. The problem occurs with audio devices that keep rendering independently of the main thread in their own threads (e.g., MT-32, OPL, etc.)

# Release notes

Fix regression where the audio is not muted while the mapper is open.

# Manual testing

Tested with **Space Quest III** (MT-32) and **Prince of Persia** (OPL). Also tested with `mute_when_inactive = on` and various pause & manual mute permutations before entering the mixer, and made sure mute state handling always works OK.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -2982,15 +2982,22 @@ void MAPPER_LosingFocus() {
 
 void MAPPER_RunEvent(uint32_t /*val*/)
 {
-	KEYBOARD_ClrBuffer();           // Clear buffer
-	GFX_LosingFocus();		//Release any keys pressed (buffer gets filled again).
+	// Clear buffer
+	KEYBOARD_ClrBuffer();           
+	
+	// Release any keys pressed (buffer gets filled again).
+	GFX_LosingFocus();		
+
 	MAPPER_DisplayUI();
 }
 
 void MAPPER_Run(bool pressed) {
-	if (pressed)
+	if (pressed) {
 		return;
-	PIC_AddEvent(MAPPER_RunEvent,0);	//In case mapper deletes the key object that ran it
+	}
+
+	// In case mapper deletes the key object that ran it
+	PIC_AddEvent(MAPPER_RunEvent,0);	
 }
 
 SDL_Surface* SDL_SetVideoMode_Wrap(int width,int height,int bpp,uint32_t flags);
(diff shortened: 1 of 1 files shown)


weirddan455:
>  The 100% solution would be to not just mute the mixer but completely stop it (e.g., introduce MIXER_Pause() and MIXER_Unpause() or something)

You can replace `MIXER_Mute()` with `MIXER_LockMixerThread()` and `MIXER_Unmute()` with `MIXER_UnlockMixerThread()` if you want to achieve this.  That's what we do for the pause hotkey.

johnnovak:
> > The 100% solution would be to not just mute the mixer but completely stop it (e.g., introduce MIXER_Pause() and MIXER_Unpause() or something)
> 
> You can replace `MIXER_Mute()` with `MIXER_LockMixerThread()` and `MIXER_Unmute()` with `MIXER_UnlockMixerThread()` if you want to achieve this. That's what we do for the pause hotkey.

Cheers, that's works very well and now it's perfect! Changed it.

src/gui/sdl_mapper.cpp
```diff
@@ -2982,15 +2983,24 @@ void MAPPER_LosingFocus() {
...
 }
 
 void MAPPER_Run(bool pressed) {
-	if (pressed)
+	if (pressed) {
 		return;
-	PIC_AddEvent(MAPPER_RunEvent,0);	//In case mapper deletes the key object that ran it
+	}
+
+	MIXER_LockMixerThread();
```
weirddan455, line 3000: I don't like the lock and unlock not being in the same function.  Unlock being inside `MAPPER_DisplayUI()`.

There's actually undefined behavior here as well.  If you start Dosbox with `--startmapper`, it calls `MAPPER_DisplayUI()` directly.  In that case, we unlock a mutex while we're not holding it.

I'd recommend either putting the lock at the top of `MAPPER_DisplayUI()` or putting both the lock and unlock at the call site of `MAPPER_RunEvent`.  The latter doesn't stop the mixer for the `--startmapper` case but I don't think we would ever be playing sound there anyway at first launch.

johnnovak: Good catch. I was quite tired when I did this yesterday night.

I put them into `MAPPER_DisplayUI()` because that's where the mouse & gfx disengage/engage calls are, so at least they're in one place. It's a royal mess, only the full rewrite will make this file better...

johnnovak:
> Looks good. 

Cheers!

> There's a lot of `E_Exit()` in that `MAPPER_DisplayUI()` function and I'm not sure what happens when you call that while holding the mixer mutex. It might deadlock or throw an exception that the thread wasn't joined.
> Not really a dealbreaker though since they're fatal errors anyway.

Yeah, that code is shit anyway and if there's a fatal error and we exit the OS will clean up everything. I'd say a non-issue.

Issue #4201, opened 2025-02-15
Sound Blaster assertion failure when fast-forwarding


johnnovak:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

main

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

Well, I'm trying to do something else entirely, but I keep running into these weird mixer bugs. Must be my magic touch 😂 

This time I managed to get Azrael's Tear shit itself when using the fast-forward feature with Sound Canvas music and SB16 sound configured. 

![Image](https://github.com/user-attachments/assets/78c03520-3def-4acc-8cf8-90f0bb6a41c6)

### Steps to reproduce the behaviour

1. Start the game (use the below config).
2. Once you're in the game, decrease cycles to about 1000-1500 via hotkeys.
3. Hold fast-forward until the crash happens (takes about 10 seconds of fast-forwarding). In a release build the process hangs and you have to force-quit it.

### Related issue

Don't wanna create a separate issue for this...

Do as instructed above, but only hold fast-forward briefly for 1 or 2 seconds. Note that the background sound will start glitching periodically and it never recovers (you can increase cycles back to 60k and wait a minute; nope, won't recover).

### Your configuration

```ini
[midi]
mididevice = soundcanvas

[autoexec]
imgmount d "cd\azrael.cue" -t cdrom
c:
azrael
```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
Fixed https://github.com/dosbox-staging/dosbox-staging/pull/4202

PR #4202, opened 2025-02-15
Don't let the mixer request frames for SoundBlaster in fast-forward mode

weirddan455:
# Description

Fast forward mode can cause the mixer to get very far behind.
In extreme cases, it can cause SoundBlaster code to integer overflow and get stuck in an infinte loop.
Avoid that by not setting frames_needed and just letting the mixer block.

## Related issues

#4201

I was not able to reproduce the assert failure.  I was getting the process hang though.  There was an integer overflow here that ended up setting `len` to 0.  It then gets stuck in an infinite loop because we call `generate_frames()` in a loop until it coughs up the frames we need.

https://github.com/dosbox-staging/dosbox-staging/blob/7fc816b50c25ccf5a2081fb34fc455d89f205a6e/src/hardware/sblaster.cpp#L3149-L3150

The background MIDI/SoundCanvas music still has the stuttering problem where it doesn't recover from fast-forward.  That's a separate issue and probably not a regression from anything I've done recently.

# Manual testing

Followed repro steps on #4201

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/sblaster.cpp
+++ b/src/hardware/sblaster.cpp
@@ -3234,7 +3234,12 @@ void SBLASTER::MixerCallback(const int frames_requested)
 	constexpr bool SignedData  = true;
 	constexpr bool NativeOrder = true;
 
-	frames_needed.store(std::max(frames_requested - check_cast<int>(output_queue.Size()), 0), std::memory_order_release);
+	// Fast forward mode can cause the mixer to get very far behind.
+	// In extreme cases, it can cause SoundBlaster code to integer overflow and get stuck in an infinte loop.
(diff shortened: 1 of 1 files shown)


johnnovak (approved):
Thanks for the quick fix @weirddan455 🥇 The crash is completely gone now.

In my test all sorts of comical things happen until both the SB and the SoundCanvas recover and "get back on track" over the next 30 seconds or so, but they _do_ recover and that's the key point here. I'm planning to address the fast-forward issue with the threaded MIDI devices like Sound Canvas later today.

Issue #4197, opened 2025-02-15
Middle-button click events to release the mouse are passed to the DOS program


johnnovak:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

main, 0.82.0

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

When you click the middle mouse button to release the captured mouse pointer, the middle-click event is also passed to the DOS game. 

This seems wrong; if `mouse_middle_release` is enabled, the middle-click event should not be sent to the DOS application.

`mouse_middle_release` is enabled by default, which is good as very few games support 3-button mice (e.g., Stonekeep and Wolfenstein 3D, see [related thread](https://www.vogons.org/viewtopic.php?t=80300)).

The similar DOSBox-X feature seems to work correctly according to [this thread](https://www.reddit.com/r/dosbox/comments/10wyz7c/can_i_disable_middle_click_from_capturing_and/) (I haven't even tried).

### Extra info

I've double checked with some GLFW example program that prints out all input events that my mouse isn't configured to also send a left button click event or something. The below screenshot is the result of pressing and releasing all three buttons on my mouse once:

![Image](https://github.com/user-attachments/assets/f307c7f7-4299-4770-95ec-26cd61c6cc95)

### Steps to reproduce the behaviour

1. Start **Gabriel Knight: Sins of the Fathers** (CD-ROM version) with Sound Blaster speech configured.
2. Press **Play** in the main menu and wait until the bookstore scene starts.
3. Note that whenever you press the middle mouse button to uncapture the mouse, it causes the dialog to skip to the next line (the left mouse button does the same thing).

### Your configuration

```ini
Default config
```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

FeralChild64:
I can't reproduce this, at least on my Linux box. In fact, this should not happen - the code was explicitly written to prevent such behavior:

```C++
void MOUSE_EventButton(const MouseButtonId button_id, const bool pressed)
{
	(...)

	if (pressed) {
		// Handle mouse capture by button click
		if (state.should_capture_on_click) {
			state.capture_was_requested = true;
			MOUSE_UpdateGFX();
			return;
		}

[...]

PR #4206, opened 2025-02-20
Improve some mouse setting descriptions

johnnovak:
# Description

This improves the descriptions of some mouse related settings. The main improvements are:

- Explain why the built-in driver is the preferred choice
- Spell out that `ps2_mouse_model` and `com_mouse_model` don't affect the built-in driver
- Spell out that you need to load a real DOS mouse driver manually if you want to use the serial mouse (or a PS/2 mouse with the built-in mouse driver disabled)

It was rather confusing to use these mouse-related settings without these extra details (see the discussion between @FeralChild64 and myself [here](https://github.com/dosbox-staging/dosbox-staging/issues/4197)).

One of the important points here is that we must spell out when a feature is only usable if the user _manually_ loads a driver -- this tripped me up big time!

Most of our features don't require the manual loading of a driver like on real DOS. E.g., Sound Blaster cards need some drivers or at least some mixer program to be loaded that initialises the card. But not so in DOSBox; you just set `sbtype`. Same deal for many other sound devices. Similarly, the various VESA extensions work without the user having to load a driver; you just set `vesa_mode`. 

One could argue that I should have known, yeah... but as all this mouse stuff is not in my head all the time like in @FeralChild64 's, it was not obvious to me what I needed to do. And if it's not obvious to me, it's rather likely that a not very technical user would be at least as confused as me. Then these nice features will remain unused, and we don't want that.

Also, `dos_mouse_driver` is not the greatest name, but I'm not changing that now. Maybe `emulated_mouse_driver` or `builtin_mouse_driver` would be better. I guess all DOS mouse drivers (real or emulated/built-in) qualify for the `dos_mouse_driver` name, so it doesn't say much 😏 

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/issues/4197

# Release notes

Clarify the descriptions of some mouse related settings.

# Manual testing

image

image

image

image

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/input/mouse_config.cpp
+++ b/src/hardware/input/mouse_config.cpp
@@ -383,10 +383,21 @@ static void config_init(Section_prop& secprop)
 	prop_bool = secprop.Add_bool("dos_mouse_driver", only_at_start, true);
 	assert(prop_bool);
 	prop_bool->Set_help(
-	        "Enable built-in DOS mouse driver (enabled by default).\n"
-	        "Notes:\n"
-	        "  - Disable if you intend to use original MOUSE.COM driver in emulated DOS.\n"
-	        "  - When guest OS is booted, built-in driver gets disabled automatically.");
+	        "Enable the built-in mouse driver (enabled by default). This results in the\n"
+	        "lowest possible latency and the smoothest mouse movement, so only disable it\n"
+	        "and load a real DOS mouse driver if it's really necessary (e.g., if a game is\n"
+	        "not compatible with the built-in driver).\n"
+	        "  on:   Enable the built-in mouse driver. `ps2_mouse_model` and\n"
+	        "        `com_mouse_model` have no effect on the built-in driver.\n"
+	        "  off:  Disable the built-in mouse driver (if you don't want mouse support or\n"
+	        "        you want to load a real DOS mouse driver). To use a real DOS driver\n"
+	        "        (e.g., MOUSE.COM or CTMOUSE.EXE), configure the mouse type with\n"
+	        "        `ps2_mouse_model` or `com_mouse_model`, then load the driver.\n"
+	        "        A real DOS driver might increase compatibility with some programs,\n"
+	        "        but will introduce more input latency.\n"
+	        "Note: The built-in driver is auto-disabled if you boot into real MS-DOS or\n"
+	        "      Windows 9x under DOSBox. Under Windows 3.x, the driver is not disabled,\n"
+	        "      but the Windows 3.x mouse driver takes over.");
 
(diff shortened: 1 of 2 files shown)


src/hardware/input/mouse_config.cpp
```diff
@@ -383,10 +383,20 @@ static void config_init(Section_prop &secprop)
...
+	        "not compatible with the built-in driver).\n"
+	        "  on:   Enable the built-in mouse driver. `ps2_mouse_model` and\n"
+	        "        `com_mouse_model` have no effect on the built-in driver.\n"
+	        "  off:  Disable the built-in mouse driver (if you don't want mouse support or\n"
+	        "        you want to load a real DOS mouse driver). To use a real DOS driver\n"
+	        "        (e.g., MOUSE.COM or CTMOUSE.EXE), configure the mouse type with\n"
+	        "        `ps2_mouse_model` or `com_mouse_model`, then load the driver.\n"
+	        "        A real DOS driver might increase compatibility with some programs,\n"
+	        "        but will introduce more input latency.\n"
+	        "Note: The built-in driver is auto-disabled if you start Windows 3.x or\n"
```
FeralChild64, line 398: The text in the `Note` is not true:
- The built-in driver gets disabled when you boot the real DOS (or Windows 9x) - but this is only due to DOSBox limitations.
- The driver doesn't get disabled when you start _Windows 3.x_; _Windows 3.x_ just ignores it, which is not good, it leads to issues with DOS programs run in the window. I would like to fix it one day, but this definitely won't be easy.

FeralChild64:
I have just one remark.

src/hardware/input/mouse_config.cpp
johnnovak: Thanks, fixed.

> The driver doesn't get disabled when you start Windows 3.x; Windows 3.x just ignores it, which is not good, it leads to issues with DOS programs run in the window. I would like to fix it one day, but this definitely won't be easy.

What would be the use-case for getting that working in the context of our project goals, which is running DOS and Win 3.x games?

I'm not against fixing such things if it's easy of even medium-hard, but if it involves extensive rewrites that have the potential of introducing mouse regressions with games, then that's a hard sell.

We should always keep the project goals in mind: running games, then maybe some demos, and then applications (but usually they just work). We should do any change that endangers the primary goal of running games extremely carefully (endangers as in causing regressions that we'll need to fix over many months or years).

johnnovak:
> I have just one remark.

Thanks for the review, I've addressed that and I've updated the first screenshot. Ready for the final review.

FeralChild64 (approved):
Now it's OK.

Issue #4212, opened 2025-02-27
GUS audio quality regression in Star Control 2 music


marsej:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Relevant hardware info

4790k, GTX 760, SSL12 audio interface

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

Hi-hats of the drums sound harsh starting at 9 seconds into the recording. (sc2 shipyard dosbox-staging-windows-x64-0.83.0-alpha-b39cf default cfg dma5 mixer100.wav)

(they sound harsh also at default mixer 50 setting but since the older versions used mixer 100 I added that change to autoexec for latest version)

### Steps to reproduce the behaviour

Compare (in captures.zip attached)
sc2 shipyard dosbox-staging-windows-x64-0.83.0-alpha-b39cf default cfg dma5 mixer100.wav
sc2 shipyard dosbox-staging-v0.82.0-rc1 bad.wav
sc2 shipyard dosbox-staging-v0.81.2 ok.wav

The hi-hats started to sound harsh between 0.81.2 and 0.82.0-rc1.

To repro:

1. Copy content of Star_Control_I_and_II_Collection_DOS_1995_for_Windows_10.iso\VirtualCD\STARCON.ISO to eg. SC2CD folder on HDD, then run SC2CD\STARCON2.COM
2. Unpacking the sc2saves.zip to the same folder starcon2.com is in:.
3. Press enter on the question
4. Load shipyard save game, then go to shipyard in the game and press ctrl+f6 to record audio and compare.
[captures.zip](https://github.com/user-attachments/files/19000778/captures.zip)
[sc2saves.zip](https://github.com/user-attachments/files/19000812/sc2saves.zip)

### Download URL of affected game or software

https://archive.org/details/starcontroliandiicollectiondos1995

### Your configuration

```ini
Default except:
gus        = true
gusbase    = 240
gusirq     = 5
gusdma     = 5  
gus_filter = off
ultradir   = C:\ULTRASND

[autoexec]
# Each line in this section is executed at startup as a DOS command.
mixer master 100
```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

marsej:
[...]

PR #4219, opened 2025-03-06
 Fix GUS resampling regression & emulate the GUS' output more authentically

johnnovak:
# Description

Previously, the output of the Gravis UltraSound was rendered at the GUS' internal mixer rate which is dependent on the number of active voices, then this output was LERP-upsampled to the host mixer rate, so typically to 44.1 or 48 kHz.

Typical GUS mixer rate examples:

- **Star Control II:** 28 active voices, 22050 Hz mixer sample rate

- **DOOM I & II:** 28 active voices, 22050 Hz mixer sample rate (with the maximum 8 sfx channels configured)

- **Pinball Fantasies:** 16 active voices, 38587 Hz mixer sample rate

This upsampling basically had a weird smoothing/low-pass filtering effect on the output which was not too far off from the GUS Classic sound _by fluke_.

My intention was to only keep the LERP upsampling for the `sb_filter = modern` mode, so nothing else would use it. But I made a mistake: I left the default resampling method at the LERP upsampler, so some audio devices like the GUS kept using it instead of the proper Speex resampler. I fixed the wrong default in this commit, and this is what introduced the "regression" (subtle change of behaviour is a better name for it): 76f6bd0744861d

So this changed the resampling method from LERP upsampling to Speex for the GUS, and that removed the high-end-smoothing effect of the LERP upsampler (if you use linear interpolation as a "poor man's upsampling method", that acts as wonky and slightly crappy lowpass filter). This made things sound subtly different in the high frequencies (subjectively, the top end became a bit harsher).

This PR introduces a better approach that gets us much closer to the actual GUS Classic sound:

- On real hardware, the GUS always outputs a 44.1 kHz sample stream to the DAC; the channel count dependent mixer rate degradation is achieved via dropping samples in the GUS mixer.

- Our GUS emulation renders at the mixer rate, but with ZOH upsampling to 44.1 kHz we can approximate this sample dropping and fixed 44.1 kHz output rate behaviour almost perfectly. We then resample this with Speex to the host rate which is usually 48 kHz.

- But this adds a lot of high-end, so we need to faithfully emulate the analog low-pass filter stage of the GUS as well. On real hardware, some gentle low-pass filtering starts from about 7-9 kHz; at 10kHz the drop is about -3 dB, and at 20kHz around -9 dB. A 6 dB/oct LPF from 8 kHz gets us very close to this sound (this has been tweaked by comparing the filtered output to real hardware GUS Classic recordings).

---

I'm also enabling the GUS by default in this PR. Why? Because the GUS is cool! 😎 It's configured not to conflict with the SB, should this shouldn't cause any problems with existing configs. @kklobe rejoice 🥳 

CPU usage is a non-issue as all audio channels are put to sleep after a few seconds of inactivity. So you'll only "pay for it" when something actually uses the GUS.

@Python-Exoproject You might wanna play it extra safe and add `gus = false` to your global config to disable the GUS by default.

## Reference recordings

- [Pinball Fantasies (GUS Classic)](https://www.youtube.com/watch?v=SyMMxKJN_cg)
- [Doom II (GUS Classic)](https://www.youtube.com/watch?v=yMZi-QPWnwU)
- [Star Control 2 (GUS ACE)](https://nerdlypleasures.blogspot.com/2016/10/reasons-for-owning-gravis-ultrasound.html)

## Other references

- [Gravis UltraSound frequency response](https://www.epanorama.net/documents/soundcardtest/gus_classic.html#)

From the article:

gusout

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4212

# Release notes

### Gravis UltraSound changes

- The Gravis UltraSound is now enabled by default.

- Fix subtle Gravis UltraSound regression where the high-frequency content of the GUS's output was a little harsher than before. This was mostly only noticeable in music with lots of hi-hats and similar high-pitched sounds.

- Additionally, the sound of the Gravis UltraSound Classic and MAX models are emulated more faithfully now. This includes the emulation of the analog low-pass output filter which is enabled by default.

# Manual testing

Tested the following games and compared the output to real hardware recordings:

- Doom I
- Doom II
- Pinball Fantasies
- Star Control 2

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/gus.cpp
+++ b/src/hardware/gus.cpp
@@ -388,21 +388,45 @@ Gus::Gus(const io_port_t port_pref, const uint8_t dma_pref, const uint8_t irq_pr
 
 	assert(channel);
 
+	// We render at the GUS' internal mixer rate, then ZOH upsample to
+	// the native 44.1 kHz GUS rate. This emulates the behaviour of the
+	// real GF1 chip which always outputs a 44.1 kHz sample stream to
+	// the DAC, but starts dropping samples in the internal mixer above
+	// 14 active voices due to bandwidth limitations. Technically, we
+	// could emulate this exact behaviour, but in practice it would 
+	// make little to no difference compared to our current method.
+	//
+	channel->SetResampleMethod(ResampleMethod::ZeroOrderHoldAndResample);
+	channel->SetZeroOrderHoldUpsamplerTargetRate(GusOutputSampleRate);
+
 	// GUS is prone to accumulating beyond the 16-bit range so we scale back
 	// by RMS.
 	constexpr auto rms_squared = static_cast<float>(M_SQRT1_2);
 	channel->Set0dbScalar(rms_squared);
 
-	if (!channel->TryParseAndSetCustomFilter(filter_prefs)) {
-		if (filter_prefs != "off") {
-			LOG_WARNING("GUS: Invalid 'gus_filter' setting: '%s', using 'off'",
-			            filter_prefs.c_str());
+	// The filter parameters have been tweaked by analysing real hardware
+	// recordings of the GUS Classic (GF1 chip).
+	//
+	auto enable_filter = [&]() {
+		constexpr auto Order        = 1;
+		constexpr auto CutoffFreqHz = 8000;
+
+		channel->ConfigureLowPassFilter(Order, CutoffFreqHz);
+		channel->SetLowPassFilter(FilterState::On);
+	};
+
+	if (const auto maybe_bool = parse_bool_setting(filter_prefs)) {
+		if (*maybe_bool) {
+			enable_filter();
+		} else {
+			channel->SetLowPassFilter(FilterState::Off);
 		}
+	} else if (!channel->TryParseAndSetCustomFilter(filter_prefs)) {
+		LOG_WARNING("GUS: Invalid 'gus_filter' setting: '%s', using 'on'",
+		            filter_prefs.c_str());
 
-		channel->SetHighPassFilter(FilterState::Off);
-		channel->SetLowPassFilter(FilterState::Off);
-
-		set_section_property_value("gus", "gus_filter", "off");
+		set_section_property_value("gus", "gus_filter", "on");
+		enable_filter();
 	}
 
 	ms_per_render = MillisInSecond / channel->GetSampleRate();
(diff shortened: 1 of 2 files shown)


weirddan455:
Nice job figuring this out. Do you have a specific commit for when this regression happened? I like to document that when possible and it might help me understand more. These upsampling and filtering options go a bit over my head (there's a lot of different combinations we use).

>  The Gravis UltraSound is now enabled by default.

My main concern with that is that for the majority of games, GUS doesn't work without placing drivers in `C:\ULTRASND`. I'm worried some games will auto-detect GUS and then leave users wondering why they have no sound in the new version where it previously worked with SoundBlaster or something.

IMO enabled by default features should be for things that work without additional steps. If the user has to manually install drivers, they should also be able to change a config setting.

johnnovak:
> Nice job figuring this out. Do you have a specific commit for when this regression happened? I like to document that when possible and it might help me understand more. These upsampling and filtering options go a bit over my head (there's a lot of different combinations we use).

Thanks, I've updated the top comment with that info, plus provided a better explanation of the sequence of events. I've also updated the commit message with the same info.

> My main concern with that is that for the majority of games, GUS doesn't work without placing drivers in C:\ULTRASND. I'm worried some games will auto-detect GUS and then leave users wondering why they have no sound in the new version where it previously worked with SoundBlaster or something.

I'll reflect to this later today. Good that you're thinking about these things, so it's great that you brought it up. Just I have some further-reaching ideas on solving this and the question of ROMs in general.

src/hardware/gus.cpp
```diff
@@ -1571,10 +1606,11 @@ void init_gus_dosbox_settings(Section_prop& secprop)
 	int_prop->Set_values({"1", "3", "5", "6", "7"});
 	int_prop->Set_help("The DMA channel of the Gravis UltraSound (3 by default).");
 
-	auto* str_prop = secprop.Add_string("gus_filter", when_idle, "off");
+	auto* str_prop = secprop.Add_string("gus_filter", when_idle, "on");
 	assert(str_prop);
 	str_prop->Set_help(
 	        "Filter for the Gravis UltraSound audio output:\n"
+	        "  on:        Filter the output (default).\n"
 	        "  off:       Don't filter the output (default).\n"
```
weirddan455, line 1614: Remove the `(default)` from "off"

weirddan455 (approved):
Thanks for the clarification. The code looks fine. One minor comment about the config text.

It's your call ultimately what you want to do with the GUS default. Just saying we may get some complaints if people's sound stops working due to missing drivers.

src/hardware/gus.cpp
johnnovak: Cheers, fixed.

johnnovak:
> Thanks for the clarification. The code looks fine. One minor comment about the config text.

Thanks man, fixed.
 
> It's your call ultimately what you want to do with the GUS default. Just saying we may get some complaints if people's sound stops working due to missing drivers.

Reverted that for now. I'm intending to experiment with including a minimal GUS folder in our Z drive (so with most EXE files stripped out, just the bare minimium for the 99% use cases).

PR #4220, opened 2025-03-06
Use std::atomic's wait over busy wait in Voodoo

weirddan455:
# Description

Please read the commit messages for more information on individual tweaks I've done.

I was never really happy with the busy waiting I did in #3952  It did increase performance but has several downsides:

- Multitasking performance suffers
- Occasional audio glitches due to the busy wait threads sometimes starving the mixer thread of resources
- Some computers will ramp up fan noise to an uncomfortable level
- The busy wait can cause thermal throttling (@FeralChild64 mentioned this)
- Threads stay maxed out after exiting from a Voodoo game

https://en.cppreference.com/w/cpp/atomic/atomic/wait is the solution I came up with. It lets us keep most of the performance due to still being lock-free while reducing CPU usage from 100% to 50-60% on my system.

FPS I measured in Descent II at the starting area:
Main: 180 FPS
This PR: 150 FPS
Main with commit 95e19509065933add80b33b162cf5ef5fe6e5bc4 reverted (old Semaphore based locking algorithim) - 70 FPS

We're down about 20%-ish performance but we're not melting the CPU anymore. And we still get over 2x improvement over the old code with locks.

# Release notes

Mac OSX minimum version increased from 10.15 to 11.0

TODO: Update https://www.dosbox-staging.org/releases/macos/ with new OSX requirement before the next release.

Reduce CPU usage of Voodoo emulation.

`voodoo_threads = auto` now sets threads equal to the number of logical (hyperthreaded) cores rather than just physical cores. The maximum auto will set has been raised from 8 to 16.

You can now manually set `voodoo_threads` to a max of 128 (up from 16). We don't know how well it will perform at high thread count but if you have a Threadripper please let us know if it scales.

# Manual testing

- Lots of performance testing in Descent II on Linux.
- Ran Quake 2 benchmark
- Tested Descent II on Windows

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:


weirddan455:
```
 /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/atomic:1600:62: note: 'wait' has been explicitly marked unavailable here
    _LIBCPP_AVAILABILITY_SYNC _LIBCPP_INLINE_VISIBILITY void wait(_Tp __v, memory_order __m = memory_order_seq_cst) const _NOEXCEPT
```

I hate Macs...

I was kind of afraid of this when I saw atomic wait was a C++ 20 feature.

I can stub the waits + notifies out on Macs so they just go back to busy waiting.  It's really not that much code to do a manual atomic wait either, assuming OSX has an OS level primitive for it.

weirddan455:
Turns out atomic wait works on Mac but it needs OSX 11.0 or later. I got CI to pass by just bumping the minimum deployment version in the GitHub workflow files.

@kklobe @johnnovak Do you have any objections to raising our OS requirement? ARM was already set to 11.0 because it looks like that's the first version that supported ARM. x86 was set to 10.15. 11.0 came out nearly 5 years ago and is EOL. Looks like 13.0 is the newest that's still maintained.

I could still stub out the atomic waits for x86 Macs with an `#ifdef` (or maybe manually implement it with OS primitives but I don't have a Mac to test on) if there's strong objections to bumping the OS requirement. 

On one hand, it kind of sucks to do this just for Voodoo but my thoughts are we're probably going to want to bump the version at some point anyway if we want other C++ 20 features.

johnnovak:
Thanks for taking a deep-dive into the macOS situation @weirddan455 despite your dislike for Macs 😆 

I think we should bump our minimum supported version to 11.0 Big Sur. Although we reverted a previous bump to 14.0 done by @interloper98 in the past, that change did not bring actual improvements outside of synthetic tests, and from memory, it caused some other issues. 

This is different: we're only increasing the minimum requirement to 11.0, and we are introducing tangible user-observable benefits. Reducing CPU usage by 50% while dropping the FPS by ~20% and stll keeping it in the 100-150 FPS range in demanding games is a huge win! 🥳 

For perspective, Big Sur was released in 2020 and it's the latest macOS version that can be installed on Intel Macs. I did an SSD and Big Sur upgrade on my mom's 2014 MacMini last year, and it flies. So in reality, we still support more than 10 year old Macs, at least for people willing to upgrade to Big Sur! Well, and those who don't wanna upgrade, we need to draw the line somewhere. Mac users who cannot or don't want to upgrade are still welcome to use 0.82.x which allows them to play the vast majority of the DOS gaming catalogue on their aging hardware.

@kklobe @Burrito78 Thoughts?

kklobe:
No major objection, I was just keeping the minimum version low for @Burrito78 :grin:

johnnovak:
I've tested this with **Lands of Lore: Guardians of Destiny** and the difference is staggering!

Using 0.82.0 my CPU goes into "hoover mode" right after starting the game (Ryzen 9 7900) and Windows reports ~50% CPU usage.

With this change, I'm getting ~20% CPU utilisation _with Sound Canvas emulation enabled (!)_, and hoover mode isn't activated 😆 

I can't notice any discernible FPS drop when playing the game either.

That's some solid improvement, @weirddan455 , great work! 🥳 

## 0.82.0

image

## PR

With Sound Canvas emulation enabled.

image

src/hardware/voodoo.cpp
```diff
@@ -4450,6 +4450,31 @@ static void triangle_worker_work(const triangle_worker& tworker,
 	sum_statistics(&v->thread_stats[work_start], &my_stats);
 }
 
+// NOTE (weirddan455): In case anyone wants to optimize this further on ARM:
+//
+// I was conservative with setting memory order on these atomic variables.
+// I've set all loads to acquire, stores to release, and load+modify+store to acq_rel.
+// x86 gets these symantics essentially for free due to it being a stronly ordered platform.
```
johnnovak, line 4457: typo: semantics

src/hardware/voodoo.cpp
```diff
@@ -4450,6 +4450,31 @@ static void triangle_worker_work(const triangle_worker& tworker,
...
+// On x86, if you ask for relaxed ordering, you'll get aquire/release anyway baring compiler re-ordering.
```
johnnovak, line 4458: typo: barring

src/hardware/voodoo.cpp
```diff
@@ -4450,6 +4450,31 @@ static void triangle_worker_work(const triangle_worker& tworker,
...
+//
+// cppreference for memory order. This one describes things more thoroughly and formally but is harder to understand:
+// https://en.cppreference.com/w/cpp/atomic/memory_order
+//
+// We shouldn't ever need Sequentially Consistent memory ordering for this use-case.
+// That's for edge cases like some lockless multiple producer multiple consumer queues.
+//
+// Loads should be either acquire or relaxed.
+// Stores should be either release or relaxed.
+// Fetch+Modify+Store operations (like fetch_add) can be acq_rel, acquire, release, or relaxed.
```
johnnovak, line 4477: That some excellent help for the next guy wanting to optimise this further on ARM. 💯 Well done, man 🎖️

src/dosbox.cpp
```diff
@@ -843,9 +843,12 @@ void DOSBOX_Init()
...
 	        "  auto:     Use up to 16 threads based on available CPU cores (default).\n"
-	        "  <value>:  Set a specific number of threads between 1 and 16.\n"
-	        "Note: Tests show that frame rates increase up to 16 threads after which\n"
-	        "      they level off or decrease, in general.");
+	        "  <value>:  Set a specific number of threads between 1 and 128.\n"
+	        "Note: Setting this to a higher value than the number of logical CPUs\n"
+	        "      your hardware supports is very likely to harm performance.\n"
+	        "      This has been measured to scale well up to 8-16 threads.\n"
+	        "      It's not been tested on a many-core CPU since performance enhancments.\n"
+	        "      If you have a Threadripper or something, please let us know how it goes.");
```
johnnovak, line 851: I'm not against upping this to 128 max thread, just would like to suggest trimming the note section a bit and wrapping the text to 80 columns. Something like this:

```cpp
        "Note: Setting this to a higher value than the number of logical CPUs\n"
	"      your hardware supports is very likely to harm performance.\n"
	"      This has been measured to scale well up to 8-16 threads.\n"
	"      It's not been tested on a many-core CPU since performance enhancments.\n"
	"      If you have a Threadripper or something, please let us know how it goes.");
```

src/misc/cross.cpp
```diff
@@ -644,95 +644,6 @@ std::string cfstr_to_string(CFStringRef source)
...
-int get_num_physical_cpus()
-{
-	// Some other OS, fall back to SDL
-	// This will actually be logical CPUs but oh well
-	return SDL_GetCPUCount();
-}
-
-#endif
-
 // ***************************************************************************
```
johnnovak, line 647: Oh yeah 😎 Good riddance! 🎉 🔥

src/hardware/voodoo.cpp
```diff
@@ -908,7 +908,7 @@ struct triangle_worker
 {
 	triangle_worker(const int num_threads_)
 	        : num_threads(num_threads_),
-	          num_work_units(num_threads + 1),
+	          num_work_units((num_threads + 1) * 4),
```
johnnovak, line 911: Very nice explanatory commit message 💯

src/hardware/voodoo.cpp
```diff
@@ -4536,9 +4549,13 @@ static void triangle_worker_run(triangle_worker& tworker)
 		return;
 	}
 
-	if (!tworker.threads_active)
+	// The main thread is the only one who sets threads_active (here and in shutdown) so there is no race condition.
+	// In the future, if this changes, this will need to be an atomic compare_exchange.
+	// For now, this is better because 99% of the time threads_active == true.
+	// We only spin up the threads once and a load is much faster than a compare_exchange.
```
johnnovak, line 4580: Nice comments! 🥇

johnnovak (approved):
Great work @weirddan455, this gets my seal of approval 🦭 😎 

Big thanks for lowering the carbon footprint of those Voodoo retro gamers among us! 😆🤘🏻 And I finally can play Lands of Lore 2 on the Voodoo without turning my PC into a hoover 🥳

kklobe:
Can confirm the awesome improvement on MacBook M3 Pro, ~400% utilization in Blood vs. ~200%, stellar work!

src/dosbox.cpp
FeralChild64: I have a Core i9 13900k, tested using unofficial Quake 3Dfx patch - performance with 16 threads is only slightly better than with 8 threads.

FeralChild64 (approved):
I did some testing - a little bit better performance than with previous solution, fan does not spin up that loud, almost no thermal throttling is triggered now :)

src/dosbox.cpp
johnnovak: Is per core utilisation lower on your box on 16 threads vs 8 threads? I think that's another consideration.

src/hardware/voodoo.cpp
weirddan455: fixed

src/hardware/voodoo.cpp
weirddan455: fixed

src/hardware/voodoo.cpp
weirddan455: Added comment

src/dosbox.cpp
FeralChild64: @johnnovak The System Monitor shows more-or-less twice the CPU usage with 16 threads used than with 8 threads used.

@weirddan455 Around 15-16 threads give the best performance. Afterwards it drops - probably because Hyper Threading gets into play, this way some Voodoo emulation thread might steal performance from the CPU emulation thread, even if it runs on a separate logical core. I'm still using kernel 6.6 (but I won't be for long), I haven't tried to set affinity.

To be honest, I don't think such testing will lead us anywhere. Hyper Threading, E-Cores/P-Cores, 3D cache (for some or all the chiplets), OS, scheduler, CPU cooler performance, maybe even the game we are running - all this can influence the results. If the user really wants the max performance, he'll have to do the testing by himself.

weirddan455: >  To be honest, I don't think such testing will lead us anywhere. Hyper Threading, E-Cores/P-Cores, 3D cache (for some or all the chiplets), OS, scheduler, CPU cooler performance, maybe even the game we are running - all this can influence the results. If the user really wants the max performance, he'll have to do the testing by himself.

I agree. This was just my own curiosity. There's too many variables to try to tweak things like CPU affinity in DOSBox code. Best we can do is set reasonable defaults for number of threads and users can tweak these things themselves.

>  Around 15-16 threads give the best performance.

Also good to know. `voodoo_threads = auto` is now set to use a max of 16 threads. Now we know that number is good for both 8 core AMDs and the new Intels with 8 P-Cores. Sounds like a good default to me.

kklobe: > Is per core utilisation lower on your box on 16 threads vs 8 threads? I think that's another consideration.

As is the number of people who can actually benefit from it:

https://store.steampowered.com/hwsurvey/cpus/

75% with 8 or fewer.

weirddan455: >  75% with 8 or fewer

That's survey is measuring physical cores though. 16 threads is to take advantage of 8 core CPUs with hyperthreading.

`voodoo_threads = auto` should do the right thing for most people. Ex: It'll use 8 threads on a quad-core CPU that has hyperthreading and 4 threads on a quad-core that does not have hyperthreading.

kklobe: ah, so that's n threads **including** the main thread, just saw the n - 1. 👍🏻

johnnovak: I'm happy with this as-is, seems like a good default to me.

weirddan455:
Did some more testing and I think this is good to merge now. Tested Quake 2 and performance is about the same as busy-wait. Tested in Descent 2 in Windows and I'm getting close to the same numbers I am in Linux.

I updated the Release Notes section with a TODO to bump the Mac OSX requirement on the website. I assume the time to do that would be whenever you open a dev page for the next major release.

johnnovak:
> I updated the Release Notes section with a TODO to bump the Mac OSX requirement on the website. I assume the time to do that would be whenever you open a dev page for the next major release.

Thanks for that 👍🏻 

Could you mention the ~50% reduced CPU usage in the release notes as well? That's the major benefit of this improvement, after all.

weirddan455:
> Could you mention the ~50% reduced CPU usage in the release notes as well? That's the major benefit of this improvement, after all.

Done

Issue #4180, opened 2025-02-07
Nethack — Falcon's Eye shows incorrect cycle count in title bar


weirddan455:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

1d47c5509f1b7fd12fa1c49179c92cb08ad225d1 (lastest main at time of writing)

### What operating system are you using?

Linux x86_64

### If other OS or Linux, please describe

Debian 12 KDE

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

Nethack — Falcon's Eye shows incorrect cycle count in the titlebar after unfocusing and refocusing the window.  I took a look at where the string gets generated here:

https://github.com/dosbox-staging/dosbox-staging/blob/1d47c5509f1b7fd12fa1c49179c92cb08ad225d1/src/cpu/cpu.cpp#L2637-L2648

It takes the "real mode" branch after a window re-focus.  `cpu.pmode` somehow got unset after the game launched.  I'm working on the audio stuttering for this game and from that I can deduce the game is still running at 300k cycles (my configured protected mode cycles) due to stuttering.  If it were to actually be running at 100k (my configured real mode cycles), it would not stutter.

### Steps to reproduce the behaviour

Explain in detail how to reproduce the bug. Assume no familiarity with the software.
1.  Launch Nethack — Falcon's Eye
2.  The titlebar shows the configured cycle count for protected mode.
3.  Unfocus the window (click on a terminal window or something), then refocus the window.
4.  The titlebar now shows configured cycle count for real mode.  I can observe this to be incorrect.

### Your configuration

```ini
[cpu]
core = dynamic
cpu_cycles = 100000
cpu_cycles_protected = 300000
```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
@weirddan455 Oops. Nice find. I'll take a look at some point; all that cycles setting code was rather messy, as usual, so I cleaned it up and decomplected it a bit. Might have made some mistake somewhere.

weirddan455:
>  ...or the game toggles between real and protected mode, and we don't update the titlebar on that change, but the defocus/focus window actions force a refresh. It's strange because the pmode flag should be the ultimate truth.

[...]

PR #4221, opened 2025-03-07
Display the correct real or protected mode cycles value in the title bar

johnnovak:
# Description

This makes displaying the correct real or protected mode cycles value in the title bar 100% reliable and in sync with what's actually happening in the CPU emulation.

As it turns out, the `cpu.pmode` flag is being rapidly and continuously flip-flopping in many protected mode games and demos (e.g., in Nethack and DOOM). Not too sure why that's the case; maybe it happens when the program calls real-mode interrupts from pmode, or maybe it's some DOSBox-special quirk. But the flow-on effect is that looking at `cpu.pmode` is not a reliable way to update the cycles indicator in the title bar.

There is more info about dealing with real mode interrupts in pmode [here](https://www.delorie.com/djgpp/doc/ug/interrupts/inthandlers2.html), but frankly, I'm not going down that rabbit hole because the whole memory handling fiasco on pre-Windows 95 PCs is disgusting, and the good news is, I don't even need to do that to fix this 😎 

What we know for certain:

- Once a program switches to pmode, it stays there for good until the program exits back to DOS.
- The pmode / real mode switching in my revised `cpu_cycles` / `cpu_cycles_protected` model works correctly (if it wasn't, we'd be swamped with issue tickets by now).
- `cpu.pmode` is not a reliable way to detect whether the currently running _program_ uses pmode or not (regardless of the momentary state of the CPU emulation).

So we just need another flag that gets flipped when the program enters pmode for the first time, then reset it when it exits back to DOS. Which is exactly how my revised pmode/real mode cycles switching works.

Phew, this was an easy one, after all 🥳 

## Related issues

Fixed https://github.com/dosbox-staging/dosbox-staging/issues/4180

# Release notes

Fixed regression where sometimes the real mode cycles value was incorrectly displayed in the DOSBox Staging window's title bar in protected mode programs.

# Manual testing

- Nethack repro case from https://github.com/dosbox-staging/dosbox-staging/issues/4180 works fine.

- Also tested DOOM and Azrael's Tear with different real and protected mode cycles settings, then Alt-Tabbed in and out of the Staging window a couple of times (this forces a title bar refresh). Confirmed we always show the correct protected mode cycles value.

- Exited and restarted a few games and made sure the real / pmode switching is reflected correctly in the title bar.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -254,6 +254,8 @@ static void set_modern_cycles_config(const CpuMode mode)
 	}
 }
 
+static bool is_protected_mode_program = false;
+
 void CPU_RestoreRealModeCyclesConfig()
 {
 	if (cpu.pmode || (!last_auto_determine_mode.auto_core &&
@@ -274,6 +276,7 @@ void CPU_RestoreRealModeCyclesConfig()
 			set_modern_cycles_config(CpuMode::Real);
 		}
 
+		is_protected_mode_program = false;
 		GFX_NotifyCyclesChanged();
 	}
 #if C_DYNAMIC_X86 || C_DYNREC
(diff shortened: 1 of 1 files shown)


weirddan455 (approved):
Nice and simple fix. Looks good to me :+1:

PR #4222, opened 2025-03-08
Dynamically load fluidsynth library at runtime

shermp:
# Description
Fluidsynth depends on glib. Compiling Glib on windows using VCPKG takes an eternity. This PR switches from linking fluidsynth at compile time to loading the fluidsynth library at runtime, falling back to host MIDI output if the library cannot be loaded.

Our official releases can include pre-compiled fluidsynth binaries for a seamless transition. Exactly how this will work is TBD.

The fluidsynth library is only loaded if fluidsynth is enabled in the config.

# Manual testing

- [x] Windows
- [x] macOS
- [x] Linux

Dynamic loading has been tested by core team members on all supported OS's.

# Checklist

I have:

### DIFF
--- a/src/libs/include/README.md
+++ b/src/libs/include/README.md
@@ -1,0 +1,24 @@
+# Runtime loaded library headers
+
+This directory contains the headers for the libraries loaded at runtime using 
+`dlopen` or `LoadLibrary`. The directory has been added to the include paths, 
+and headers can be included as they normally would.
+
+The headers for the following libraries are included:
+
+- Fluidsynth
+
+Ideally, headers should come from the minimum required version of the library, 
+and only updated when new features of the API are used.
+
+Notes about each library are provided in the next sections.
+
+## Fluidsynth
+
+Headers are sourced from the automated GitHub archive from the 2.2.3 release. 
+The contents of the `include` directory have been copied into this directory.
+
+The original `fluidsynth.cmake` file has been renamed to `fluidsynth.h` and 
+the conditional logic for `BUILD_SHARED_LIBS` was removed and replaced with a 
+bare `#define FLUIDSYNTH_API`. Additionally, `#include "fluidsynth/version.h"` 
+has been commented out.
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -104,6 +104,31 @@ enum fluid_interp
 };
 
 /**
+ * FluidSynth log levels.
+ */
+enum fluid_log_level
+{
+    FLUID_PANIC,   /**< The synth can't function correctly any more */
+    FLUID_ERR,     /**< Serious error occurred */
+    FLUID_WARN,    /**< Warning */
+    FLUID_INFO,    /**< Verbose informational messages */
+    FLUID_DBG,     /**< Debugging messages */
+    LAST_LOG_LEVEL /**< @internal This symbol is not part of the public API and ABI
+                     stability guarantee and may change at any time! */
+};
+
+/* This typedef comes from FluidSynth */
+
+/**
+ * Log function handler callback type used by fluid_set_log_function().
+ *
+ * @param level Log level (#fluid_log_level)
+ * @param message Log message text
+ * @param data User data pointer supplied to fluid_set_log_function().
+ */
+typedef void (*fluid_log_function_t)(int level, const char *message, void *data);
+
+/**
  * FluidSynth dynamic library handle
  */
 static dynlib_handle fsynth_lib = {};
@@ -120,7 +145,9 @@ void (*fluid_version)(int *major, int *minor, int *micro) = nullptr;
 fluid_settings_t* (*new_fluid_settings)(void)                       = nullptr;
 fluid_synth_t*       (*new_fluid_synth)(fluid_settings_t *settings) = nullptr;
 
-int  (*fluid_settings_setnum)(fluid_settings_t *settings, const char *name, double val) = nullptr;
+fluid_log_function_t (*fluid_set_log_function)(int level, fluid_log_function_t fun, void *data)      = nullptr;
+
+int  (*fluid_settings_setnum)(fluid_settings_t *settings, const char *name, double val)              = nullptr;
 
 int               (*fluid_synth_chorus_on)(fluid_synth_t *synth, int fx_group, int on)               = nullptr;
 int     (*fluid_synth_set_chorus_group_nr)(fluid_synth_t *synth, int fx_group, int nr)               = nullptr;
@@ -192,6 +219,7 @@ static DynLibResult load_fsynth_dynlib()
 			return DynLibResult::LibOpenErr;
 		}
 		DB_GET_FSYNTH_SYM(fluid_version);
+        DB_GET_FSYNTH_SYM(fluid_set_log_function);
 
 		DB_GET_FSYNTH_SYM(new_fluid_settings);
 		DB_GET_FSYNTH_SYM(new_fluid_synth);
@@ -226,6 +254,11 @@ static DynLibResult load_fsynth_dynlib()
 		DB_GET_FSYNTH_SYM(fluid_synth_pitch_bend);
 		DB_GET_FSYNTH_SYM(fluid_synth_sysex);
 		DB_GET_FSYNTH_SYM(fluid_synth_write_float);
+
+        /* Keep ERR and PANIC logging only */
+        for (auto level : {FLUID_DBG, FLUID_INFO, FLUID_WARN}) {
+            fluid_set_log_function(level, nullptr, nullptr);
+        }
 	}
 	return DynLibResult::Success;
 }
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -458,13 +458,12 @@ static std_fs::path find_sf_file(const std::string& sf_name)
 				// comparisons.
 				std::error_code err = {};
 				const auto canonical_path =
-				        std_fs::canonical(sf, err).c_str();
+				        std_fs::canonical(sf, err);
 
 				if (err) {
 					return {};
-				} else {
-					return canonical_path;
 				}
+				return canonical_path;
 			}
 		}
 	}
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -133,7 +133,7 @@ typedef void (*fluid_log_function_t)(int level, const char *message, void *data)
  */
 static dynlib_handle fsynth_lib = {};
 
-/* The following function pointers will be set to their corresponding symbols in the FluidSynth library */
+// The following function pointers will be set to their corresponding symbols in the FluidSynth library
 
 /**
  * A 'X-Macro' to generate a list of function pointers to symbols in the 
@@ -213,7 +213,7 @@ static DynLibResult load_fsynth_dynlib(std::string& err_str)
 		}
 		FSYNTH_FUNC_LIST(FSYNTH_FUNC_GET_SYM)
 
-        /* Keep ERR and PANIC logging only */
+        // Keep ERR and PANIC logging only
         for (auto level : {FLUID_DBG, FLUID_INFO, FLUID_WARN}) {
             fluid_set_log_function(level, nullptr, nullptr);
         }
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -71,7 +71,7 @@ struct FSynthVersion
 constexpr FSynthVersion min_fsynth_version = {2, 2, 3};
 constexpr FSynthVersion max_fsynth_version_exclusive = {3, 0, 0};
 
-namespace fsynth 
+namespace FluidSynth 
 {
 /**
  * FluidSynth dynamic library handle
@@ -128,16 +128,16 @@ static dynlib_handle fsynth_lib = {};
 
 FSYNTH_FUNC_LIST(FSYNTH_FUNC_DECLARE)
 
-} // namespace fsynth
+} // namespace FluidSynth
 
 /**
  * A filthy macro to resolve fluidsynth symbols, and return from the 
  * calling function below on error.
  */
 #define FSYNTH_FUNC_GET_SYM(ret_type, name, sig) \
-	fsynth::name = (decltype(fsynth::name))dynlib_get_symbol(fsynth::fsynth_lib, #name); \
-	if (!fsynth::name) { \
-		dynlib_close(fsynth::fsynth_lib); \
+	FluidSynth::name = (decltype(FluidSynth::name))dynlib_get_symbol(FluidSynth::fsynth_lib, #name); \
+	if (!FluidSynth::name) { \
+		dynlib_close(FluidSynth::fsynth_lib); \
 		err_str = "FSYNTH: Failed to get symbol: '" #ret_type " " #name #sig "'"; \
 		return DynLibResult::ResolveSymErr; \
 	}
@@ -149,9 +149,9 @@ FSYNTH_FUNC_LIST(FSYNTH_FUNC_DECLARE)
  */
 static DynLibResult load_fsynth_dynlib(std::string& err_str)
 {
-	if (!fsynth::fsynth_lib) {
-		fsynth::fsynth_lib = dynlib_open(fsynth_dynlib_file);
-		if (!fsynth::fsynth_lib) {
+	if (!FluidSynth::fsynth_lib) {
+		FluidSynth::fsynth_lib = dynlib_open(fsynth_dynlib_file);
+		if (!FluidSynth::fsynth_lib) {
 			err_str = "FSYNTH: Failed to load FluidSynth library";
 			return DynLibResult::LibOpenErr;
 		}
@@ -159,7 +159,7 @@ static DynLibResult load_fsynth_dynlib(std::string& err_str)
 
         // Keep ERR and PANIC logging only
         for (auto level : {FLUID_DBG, FLUID_INFO, FLUID_WARN}) {
-            fsynth::fluid_set_log_function(level, nullptr, nullptr);
+            FluidSynth::fluid_set_log_function(level, nullptr, nullptr);
         }
 	}
 	return DynLibResult::Success;
@@ -494,13 +494,13 @@ static void setup_chorus(fluid_synth_t* synth, const std_fs::path& sf_path)
 	constexpr int FxGroup = -1;
 
 // Current API calls as of 2.2
-	fsynth::fluid_synth_chorus_on(synth, FxGroup, chorus_enabled);
-	fsynth::fluid_synth_set_chorus_group_nr(synth, FxGroup, chorus_voice_count);
-	fsynth::fluid_synth_set_chorus_group_level(synth, FxGroup, chorus_level);
-	fsynth::fluid_synth_set_chorus_group_speed(synth, FxGroup, chorus_speed);
-	fsynth::fluid_synth_set_chorus_group_depth(synth, FxGroup, chorus_depth);
+	FluidSynth::fluid_synth_chorus_on(synth, FxGroup, chorus_enabled);
+	FluidSynth::fluid_synth_set_chorus_group_nr(synth, FxGroup, chorus_voice_count);
+	FluidSynth::fluid_synth_set_chorus_group_level(synth, FxGroup, chorus_level);
+	FluidSynth::fluid_synth_set_chorus_group_speed(synth, FxGroup, chorus_speed);
+	FluidSynth::fluid_synth_set_chorus_group_depth(synth, FxGroup, chorus_depth);
 
-	fsynth::fluid_synth_set_chorus_group_type(synth,
+	FluidSynth::fluid_synth_set_chorus_group_type(synth,
 	                                  FxGroup,
 	                                  static_cast<int>(chorus_mod_wave));
 
@@ -567,12 +567,12 @@ static void setup_reverb(fluid_synth_t* synth)
 	constexpr int FxGroup = -1;
 
 // Current API calls as of 2.2
-	fsynth::fluid_synth_reverb_on(synth, FxGroup, reverb_enabled);
-	fsynth::fluid_synth_set_reverb_group_roomsize(synth, FxGroup, reverb_room_size);
+	FluidSynth::fluid_synth_reverb_on(synth, FxGroup, reverb_enabled);
+	FluidSynth::fluid_synth_set_reverb_group_roomsize(synth, FxGroup, reverb_room_size);
 
-	fsynth::fluid_synth_set_reverb_group_damp(synth, FxGroup, reverb_damping);
-	fsynth::fluid_synth_set_reverb_group_width(synth, FxGroup, reverb_width);
-	fsynth::fluid_synth_set_reverb_group_level(synth, FxGroup, reverb_level);
+	FluidSynth::fluid_synth_set_reverb_group_damp(synth, FxGroup, reverb_damping);
+	FluidSynth::fluid_synth_set_reverb_group_width(synth, FxGroup, reverb_width);
+	FluidSynth::fluid_synth_set_reverb_group_level(synth, FxGroup, reverb_level);
 
 
 	if (reverb_enabled) {
@@ -601,7 +601,7 @@ MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 	}
 
 	FSynthVersion vers = {};
-	fsynth::fluid_version(&vers.major, &vers.minor, &vers.micro);
+	FluidSynth::fluid_version(&vers.major, &vers.minor, &vers.micro);
 	if (vers < min_fsynth_version || vers >= max_fsynth_version_exclusive) {
 		const auto msg = "FSYNTH: FluidSynth version must be at least 2.2.3 and less than 3.0.0";
 		LOG_ERR("%s. Version loaded is %d.%d.%d", msg, vers.major, vers.minor, vers.micro);
@@ -610,8 +610,8 @@ MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 		LOG_MSG("FSYNTH: Successfully loaded FluidSynth %d.%d.%d", vers.major, vers.minor, vers.micro);
 	}
 	
-	FluidSynthSettingsPtr fluid_settings(fsynth::new_fluid_settings(),
-	                                     fsynth::delete_fluid_settings);
+	FluidSynthSettingsPtr fluid_settings(FluidSynth::new_fluid_settings(),
+	                                     FluidSynth::delete_fluid_settings);
 	if (!fluid_settings) {
 		const auto msg = "FSYNTH: Failed to initialise the FluidSynth settings";
 		LOG_ERR("%s", msg);
@@ -629,12 +629,12 @@ MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 	const auto sample_rate_hz = MIXER_GetSampleRate();
 	ms_per_audio_frame        = MillisInSecond / sample_rate_hz;
 
-	fsynth::fluid_settings_setnum(fluid_settings.get(),
+	FluidSynth::fluid_settings_setnum(fluid_settings.get(),
 	                              "synth.sample-rate",
 	                              sample_rate_hz);
 
-	FluidSynthPtr fluid_synth(fsynth::new_fluid_synth(fluid_settings.get()),
-	                          fsynth::delete_fluid_synth);
+	FluidSynthPtr fluid_synth(FluidSynth::new_fluid_synth(fluid_settings.get()),
+	                          FluidSynth::delete_fluid_synth);
 	if (!fluid_synth) {
 		const auto msg = "FSYNTH: Failed to create the FluidSynth synthesizer";
 		LOG_ERR("%s", msg);
@@ -645,14 +645,14 @@ MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 	const auto sf_name = section->Get_string("soundfont");
 	const auto sf_path = find_sf_file(sf_name);
 
-	if (!sf_path.empty() && fsynth::fluid_synth_sfcount(fluid_synth.get()) == 0) {
+	if (!sf_path.empty() && FluidSynth::fluid_synth_sfcount(fluid_synth.get()) == 0) {
 		constexpr auto ResetPresets = true;
-		fsynth::fluid_synth_sfload(fluid_synth.get(),
+		FluidSynth::fluid_synth_sfload(fluid_synth.get(),
 								   sf_path.string().c_str(),
 								   ResetPresets);
 	}
 
-	if (fsynth::fluid_synth_sfcount(fluid_synth.get()) == 0) {
+	if (FluidSynth::fluid_synth_sfcount(fluid_synth.get()) == 0) {
 		const auto msg = format_str("FSYNTH: Error loading SoundFont '%s'",
 		                            sf_name.c_str());
 
@@ -661,7 +661,7 @@ MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 	}
 
 	auto sf_volume_percent = section->Get_int("soundfont_volume");
-	fsynth::fluid_synth_set_gain(fluid_synth.get(),
+	FluidSynth::fluid_synth_set_gain(fluid_synth.get(),
 								 static_cast<float>(sf_volume_percent) / 100.0f);
 
 	// Let the user know that the SoundFont was loaded
@@ -677,7 +677,7 @@ MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 	constexpr int FxGroup = -1;
 
 	// Use a 7th-order (highest) polynomial to generate MIDI channel waveforms
-	fsynth::fluid_synth_set_interp_method(fluid_synth.get(),
+	FluidSynth::fluid_synth_set_interp_method(fluid_synth.get(),
 	                                      FxGroup,
 	                                      FLUID_INTERP_HIGHEST);
 
@@ -845,9 +845,9 @@ void MidiDeviceFluidSynth::ApplyChannelMessage(const std::vector<uint8_t>& msg)
 
 	// clang-format off
 	switch (status) {
-	case MidiStatus::NoteOff:         fsynth::fluid_synth_noteoff(     synth.get(), channel, msg[1]);         break;
-	case MidiStatus::NoteOn:          fsynth::fluid_synth_noteon(      synth.get(), channel, msg[1], msg[2]); break;
-	case MidiStatus::PolyKeyPressure: fsynth::fluid_synth_key_pressure(synth.get(), channel, msg[1], msg[2]); break;
+	case MidiStatus::NoteOff:         FluidSynth::fluid_synth_noteoff(     synth.get(), channel, msg[1]);         break;
+	case MidiStatus::NoteOn:          FluidSynth::fluid_synth_noteon(      synth.get(), channel, msg[1], msg[2]); break;
+	case MidiStatus::PolyKeyPressure: FluidSynth::fluid_synth_key_pressure(synth.get(), channel, msg[1], msg[2]); break;
 
 	case MidiStatus::ControlChange: {
 		const auto controller = msg[1];
@@ -892,13 +892,13 @@ void MidiDeviceFluidSynth::ApplyChannelMessage(const std::vector<uint8_t>& msg)
 			// used rarely and usually only to add some subtle flair to the
 			// start of the notes in synth-oriented soundtracks.
 		} else {
-			fsynth::fluid_synth_cc(synth.get(), channel, controller, value);
+			FluidSynth::fluid_synth_cc(synth.get(), channel, controller, value);
 		}
 	} break;
 
-	case MidiStatus::ProgramChange:   fsynth::fluid_synth_program_change(  synth.get(), channel, msg[1]);                 break;
-	case MidiStatus::ChannelPressure: fsynth::fluid_synth_channel_pressure(synth.get(), channel, msg[1]);                 break;
-	case MidiStatus::PitchBend:       fsynth::fluid_synth_pitch_bend(      synth.get(), channel, msg[1] + (msg[2] << 7)); break;
+	case MidiStatus::ProgramChange:   FluidSynth::fluid_synth_program_change(  synth.get(), channel, msg[1]);                 break;
+	case MidiStatus::ChannelPressure: FluidSynth::fluid_synth_channel_pressure(synth.get(), channel, msg[1]);                 break;
+	case MidiStatus::PitchBend:       FluidSynth::fluid_synth_pitch_bend(      synth.get(), channel, msg[1] + (msg[2] << 7)); break;
 	default: log_unknown_midi_message(msg); break;
 	}
 	// clang-format on
@@ -910,7 +910,7 @@ void MidiDeviceFluidSynth::ApplySysExMessage(const std::vector<uint8_t>& msg)
 	const char* data = reinterpret_cast<const char*>(msg.data());
 	const auto n     = static_cast<int>(msg.size());
 
-	fsynth::fluid_synth_sysex(synth.get(), data, n, nullptr, nullptr, nullptr, false);
+	FluidSynth::fluid_synth_sysex(synth.get(), data, n, nullptr, nullptr, nullptr, false);
 }
 
 // The callback operates at the audio frame-level, steadily adding samples to
@@ -957,7 +957,7 @@ void MidiDeviceFluidSynth::RenderAudioFramesToFifo(const int num_audio_frames)
 		audio_frames.resize(num_audio_frames);
 	}
 
-	fsynth::fluid_synth_write_float(synth.get(),
+	FluidSynth::fluid_synth_write_float(synth.get(),
 									num_audio_frames,
 									&audio_frames[0][0],
 									0,
--- a/src/midi/midi_fluidsynth.cpp
+++ b/src/midi/midi_fluidsynth.cpp
@@ -20,10 +20,9 @@
 
 #include "midi_fluidsynth.h"
 
-#if C_FLUIDSYNTH
-
 #include <bitset>
 #include <cassert>
+#include <compare>
 #include <numeric>
 #include <string>
 #include <tuple>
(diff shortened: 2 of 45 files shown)


src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +42,149 @@
...
+int  (*fluid_synth_set_interp_method)(fluid_synth_t *synth, int chan, int interp_method) = nullptr;
+int  (*fluid_synth_noteoff)(fluid_synth_t *synth, int chan, int key) = nullptr;
+int  (*fluid_synth_noteon)(fluid_synth_t *synth, int chan, int key, int vel) = nullptr;
+int  (*fluid_synth_key_pressure)(fluid_synth_t *synth, int chan, int key, int val) = nullptr;
+int  (*fluid_synth_cc)(fluid_synth_t *synth, int chan, int ctrl, int val) = nullptr;
+int  (*fluid_synth_program_change)(fluid_synth_t *synth, int chan, int program) = nullptr;
+int  (*fluid_synth_channel_pressure)(fluid_synth_t *synth, int chan, int val) = nullptr;
+int  (*fluid_synth_pitch_bend)(fluid_synth_t *synth, int chan, int val) = nullptr;
+int  (*fluid_synth_sysex)(fluid_synth_t *synth, const char *data, int len,
+							char *response, int *response_len, int *handled, int dryrun) = nullptr;
```
johnnovak, line 125: minor: indentation

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +42,149 @@
...
+int  (*fluid_synth_write_float)(fluid_synth_t *synth, int len,
+								void *lout, int loff, int lincr,
+								void *rout, int roff, int rincr) = nullptr;
```
johnnovak, line 128: minor: indentation

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +42,149 @@
...
+
+/**
+ * A filthy macro to make resolving library symbols a little less verbose, 
+ * while still including robust error checking.
```
johnnovak, line 132: 💯

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +42,149 @@
 
 constexpr auto SoundFontExtension = ".sf2";
 
+/**
+ * Platform specific extension for shared libraries
+ */
+#if defined(WIN32)
+#define DB_FLUID_LIB_EXT ".dll"
+#else
+#define DB_FLUID_LIB_EXT ".so"
```
johnnovak, line 51: Not entirely sure whether it's `.so` or `.dylib` on the Mac in this case. Can't test on a Mac now, so maybe @kklobe will be able to confirm.

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +42,149 @@
...
+    FLUID_INTERP_DEFAULT = FLUID_INTERP_4THORDER, /**< Default interpolation method */
+    FLUID_INTERP_HIGHEST = FLUID_INTERP_7THORDER, /**< Highest interpolation method */
+};
+
+/**
+ * Fluidsynth dynamic library handle
+ */
+static dynlib_handle fluid_lib = {};
+
+/* The following function pointers will be set to their corresponding symbol in the fluidsynth library */
```
johnnovak, line 88: typo: symbols (plural)
nitpicking: FluidSynth

src/midi/midi_fluidsynth.cpp
```diff
@@ -386,17 +526,6 @@ static void setup_chorus(fluid_synth_t* synth, const std_fs::path& sf_path)
...
-#else
-	fluid_synth_set_chorus_on(synth, chorus_enabled);
-	fluid_synth_set_chorus(synth,
-	                       chorus_voice_count,
-	                       chorus_level,
-	                       chorus_speed,
-	                       chorus_depth,
-	                       chorus_mod_wave);
-#endif
-
```
johnnovak, line 399: It's not the end of the world, but it would be nicer to do these all version check removals in a separate commit (and good job removing them 💯 ).

src/midi/midi_fluidsynth.cpp
```diff
@@ -491,6 +610,34 @@ static void setup_reverb(fluid_synth_t* synth)
 
 MidiDeviceFluidSynth::MidiDeviceFluidSynth()
 {
+	DynLibResult res = load_dynlib_fluid();
+	switch (res) {
+		case DynLibResult::Success:
+			break;
+		case DynLibResult::LibOpenErr: {
+			const auto sym_msg = "FSYNTH: Failed to load fluidsynth library";
```
johnnovak, line 618: Maybe `FluidSynth library`? (capitalisation)

src/midi/midi_fluidsynth.cpp
```diff
@@ -491,6 +610,34 @@ static void setup_reverb(fluid_synth_t* synth)
...
+			LOG_ERR("%s", sym_msg);
+			throw std::runtime_error(sym_msg);
+			break;
+		}
+		case DynLibResult::ResolveSymErr: {
+			const auto sym_msg = "FSYNTH: Failed to resolve one or more fluidsynth symbols";
```
johnnovak, line 624: Maybe `FluidSynth library`? (capitalisation)

src/midi/midi_fluidsynth.cpp
```diff
@@ -491,6 +610,34 @@ static void setup_reverb(fluid_synth_t* synth)
...
+			LOG_ERR("%s", sym_msg);
+			throw std::runtime_error(sym_msg);
+			break;
+		}
+	}
+
+	int v_maj, v_min, v_micro;
+	fluid_version(&v_maj, &v_min, &v_micro);
+	if (v_maj != 2 || v_min < 2) {
+		const auto msg = "FSYNTH: Fluidsynth version must be at least 2.2.0";
```
johnnovak, line 634: minor: `FluidSynth` (capitalisation)

src/midi/midi_fluidsynth.cpp
```diff
@@ -491,6 +610,34 @@ static void setup_reverb(fluid_synth_t* synth)
...
+		}
+		case DynLibResult::ResolveSymErr: {
+			const auto sym_msg = "FSYNTH: Failed to resolve one or more fluidsynth symbols";
+			LOG_ERR("%s", sym_msg);
+			throw std::runtime_error(sym_msg);
+			break;
+		}
+	}
+
+	int v_maj, v_min, v_micro;
```
johnnovak, line 631: minor: maybe spell out "major" and "minor" too

src/midi/midi_fluidsynth.cpp
```diff
@@ -491,6 +610,34 @@ static void setup_reverb(fluid_synth_t* synth)
...
+	fluid_version(&v_maj, &v_min, &v_micro);
+	if (v_maj != 2 || v_min < 2) {
+		const auto msg = "FSYNTH: Fluidsynth version must be at least 2.2.0";
+		LOG_ERR("%s. Version loaded is %d.%d.%d", msg, v_maj, v_min, v_micro);
+		throw std::runtime_error(msg);
+	} else {
+		LOG_MSG("FSYNTH: Successfully loaded fluidsynth %d.%d.%d", v_maj, v_min, v_micro);
```
johnnovak, line 638: `FluidSynth`

vcpkg.json
```diff
@@ -4,7 +4,6 @@
   "description": "DOSBox Staging is a modern continuation of DOSBox with advanced features and current development practices.",
   "version": "0.83.0-alpha",
   "dependencies": [
-    "fluidsynth",
```
johnnovak, line 7: Oh yeah 😎 🥇

src/midi/midi_fluidsynth.cpp
shermp: I was thinking "but then I have to implement the symbols, just to remove them".

But duh, I can just remove them before introducing the dynamic loading system.

johnnovak:
@shermp I've tested this and it works beautifully. Great job! 🎉 🎖️  Falling back to the default MIDI device also works fine if `mididevice = fluidsynth` is requested and the DLL cannot be loaded.

Just one thing, I think it would be better if we disabled FluidSynth-level logging through its logging API:

https://www.fluidsynth.org/api/group__logging.html

By default, the DLL will log a lot of junk to stdout:

image

shermp:
@johnnovak thanks for testing.

I haven't seen that debug output myself, I wonder what causes it?

johnnovak:
> @johnnovak thanks for testing.

My pleausure 😎 

> I haven't seen that debug output myself, I wonder what causes it?
> Unless you're using a dll compiled in debug mode?

That's it yeah, FluidSynth DLL generate by compiling the Visual Studio project in debug mode.

I've also found a re-init issue; I've updated my above comment. Should be easy to fix, though.

shermp:
> > @johnnovak thanks for testing.
> 
> My pleasure 😎
> 
> > I haven't seen that debug output myself, I wonder what causes it?
> > Unless you're using a dll compiled in debug mode?
> 
> That's it yeah, FluidSynth DLL generated by compiling the Visual Studio project in debug mode.

Does this occur on a debug build of `main` by any chance? If it does, maybe out of scope?

src/midi/midi_fluidsynth.cpp
johnnovak: > But duh, I can just remove them before introducing the dynamic loading system.

Yeah, that's exactly what I thought, let's do that 😄

johnnovak:
> Does this occur on a debug build of `main` by any chance? If it does, maybe out of scope?

It's just a FluidSynth DLL compiled in debug mode. Probably the vcpkg package sets up the default log level in a particular way for debug-mode FluidSynth DLLs.

But it's one API call do disable FluidSynth logging altogether, no matter what, so why not just do that single call? It's pretty easy to do.

shermp:
> > Does this occur on a debug build of `main` by any chance? If it does, maybe out of scope?
> 
> It's just a FluidSynth DLL compiled in debug mode. Probably the vcpkg package sets up the default log level in a particular way for debug-mode FluidSynth DLLs.
> 
> But it's one API call do disable FluidSynth logging altogether, no matter what, so why not just do that single call? It's pretty easy to do.

Err, unless I'm missing something, I don't see a way of disabling logging altogether, except by providing a logging function that does nothing. Am I missing something? Or, I could just set the log level to PANIC I suppose.

johnnovak:
> Err, unless I'm missing something, I don't see a way of disabling logging altogether, except by providing a logging function that does nothing. Am I missing something? Or, I could just set the log level to PANIC I suppose.

Lol, we're in race condition mode with my concurrent edits 😄 See my updated comment (and you're right, yeah).

shermp:
Oh...

> A client application can install a new log function to handle the messages differently. In the following example, the application sets a callback function to display [FLUID_PANIC](https://www.fluidsynth.org/api/group__logging.html#ggaa341232c15addd9c336131984840ed9ea62d9ddd89d0a853606ce38d195daa99a) messages in a dialog, and ignores all other messages by setting the log function to NULL:

shermp:
I'm thinking it might be worth keeping ERROR and PANIC log levels.

johnnovak:
> I'm thinking it might be worth keeping ERROR and PANIC log levels.

We can do that see how it goes.

weirddan455:
I'm getting a segfault on start-up. It looks like a symbol collision problem. I was able to fix it by setting the function pointers to `static`.

```
Thread 1 "dosbox" received signal SIGSEGV, Segmentation fault.
0x0000555558e4bf38 in fluid_synth_reverb_on ()
(gdb) bt
#0  0x0000555558e4bf38 in fluid_synth_reverb_on ()
#1  0x00007fffe195e050 in new_fluid_synth () at /usr/lib/x86_64-linux-gnu/libfluidsynth.so.3
#2  0x00005555564b890f in MidiDeviceFluidSynth::MidiDeviceFluidSynth() (this=0x555558fcaae0)
    at ../../src/midi/midi_fluidsynth.cpp:663
#3  0x00005555564b0b48 in std::make_unique<MidiDeviceFluidSynth>() () at /usr/include/c++/12/bits/unique_ptr.h:1065
#4  0x00005555564ac7ef in create_device(std::string const&, std::string const&) (name="fluidsynth", config="")
    at ../../src/midi/midi.cpp:103
#5  0x00005555564b0667 in MIDI::MIDI() (this=0x555559386a00) at ../../src/midi/midi.cpp:671
#6  0x00005555564b12fe in std::make_unique<MIDI>() () at /usr/include/c++/12/bits/unique_ptr.h:1065
#7  0x00005555564ad9f5 in midi_init(Section*) (sec=0x555558ee22f0) at ../../src/midi/midi.cpp:774
#8  0x0000555555e0d81d in Section::ExecuteInit(bool) (this=0x555558ee22f0, init_all=true) at ../../src/misc/setup.cpp:1374
#9  0x0000555555e0d658 in Config::Init() const (this=0x555558eace80) at ../../src/misc/setup.cpp:1346
#10 0x0000555556297e88 in sdl_main(int, char**) (argc=1, argv=0x7fffffffdda8) at ../../src/gui/sdlmain.cpp:5072
#11 0x0000555555ddcb89 in main(int, char**) (argc=1, argv=0x7fffffffdda8) at ../../src/main.cpp:30
```
The crash happens inside of libfluidsynth when it calls into one of its own functions that you've made a global function pointer here. Note the memory address of `fluid_synth_reverb_on()`

```
(gdb) p fluid_synth_reverb_on
$1 = (int (*)(fluid_synth_t *, int, int)) 0x7fffe195ac90 <fluid_synth_reverb_on>
(gdb) p &fluid_synth_reverb_on
$2 = (int (**)(fluid_synth_t *, int, int)) 0x555558e4bf38 <fluid_synth_reverb_on>
```
It's jumping execution to the address of the function pointer rather than the actual function that it's pointing to. Inside of libfluidsynth, it's expecting `fluid_synth_reverb_on` to be a function but you've re-exported that symbol to be a function pointer instead.

Took me a while to figure that one out. Linker problems are never fun to debug and their behavior can vary between OS's. `static` works as it tells the linker not to export the symbol (it remains local to the compilation unit).

>  Exeuting mididevice fluidsynth from the shell twice results in a crash.

I don't have this problem but it may be the same root cause manifesting differently in Windows.

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +42,149 @@
...
+
+/**
+ * Load the fluidsynth library and resolve all required symbols.
+ * 
+ * If the library is already loaded, does nothing.
+ */
+static DynLibResult load_dynlib_fluid()
+{
+	if (!fluid_lib) {
+		fluid_lib = dynlib_open("libfluidsynth-3" DB_FLUID_LIB_EXT);
```
weirddan455, line 145: The name of the library file on my system is `libfluidsynth.so.3` which I think is going to be the most common on Linux. This is also the name I get when building fluidsynth from source. It's actually a symlink to `libfluidsynth.so.3.1.4` but the minor version will differ between distros.

After changing this and fixing the segfault issue, everything seems to work fine.

shermp:
Thanks for testing @weirddan455 

I had a suspicion that symbol collisions might happen. Thanks for confirming it is indeed an issue.

My original implementation had all the function pointers in the `MidiDeviceFluidSynth` class. Maybe a namespace will do the trick.

I'm trying to avoid having to modify every call to fluidsynth functions if I can. Makes it easily reversible.

src/midi/midi_fluidsynth.cpp
shermp: I may have gone a bit overboard with formatting these now...

src/midi/midi_fluidsynth.cpp
shermp: I may have gone a bit overboard with formatting these now...

src/midi/midi_fluidsynth.cpp
shermp: I'm now using a version struct.

src/midi/midi_fluidsynth.cpp
shermp: I checked Debian and Fedora, and they're both using `libfluidsynth.so.3`, so switched to that.

shermp:
Ok, I think I've addressed the first wave of comments.

I decided to use the namespace approach to avoid symbol clashes. That's the entire purpose of namespaces after all.

Debug, warning, info logging from fluidsynth is now suppressed.

Switching `mididevice` doesn't crash for me with the current changes.

weirddan455:
>  I decided to use the namespace approach to avoid symbol clashes. That's the entire purpose of namespaces after all.

Nice, I just tested and everything works now. I wasn't sure namespaces would work at the linker level but I guess it mangles the symbol names so it's a good solution.

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +43,191 @@
...
+int         (*fluid_synth_pitch_bend)(fluid_synth_t *synth, int chan, int val)                       = nullptr;
+
+int        (*fluid_synth_sysex)(fluid_synth_t *synth, 
+                                const char *data, 
+                                int len,
+                                char *response, 
+                                int *response_len, 
+                                int *handled, 
+                                int dryrun) = nullptr;
+int  (*fluid_synth_write_float)(fluid_synth_t *synth,
```
johnnovak, line 182: Needs blank line above this.

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +43,191 @@
...
+                                int *handled, 
+                                int dryrun) = nullptr;
+int  (*fluid_synth_write_float)(fluid_synth_t *synth, 
+                                int len,
+								void *lout, 
+                                int loff, 
+                                int lincr,
+								void *rout, 
+                                int roff, 
+                                int rincr) = nullptr;
```
johnnovak, line 189: Yeah, all that formatting looks good in Visual Studio, but it's all over the place here because you used a mixture of tabs and spaces. Can you replace all tabs with spaces in this block with your editor? Then it's perfect 😄

shermp: Oh FFS. 

I think it's partly because I'm editing in VS code and compiling in VS 2022.

shermp:
Well that was a nightmare trying to fix whitespace issues. Apologies for the churn.

johnnovak:
This is shaping up nicely @shermp 🎉 We're almost there, just two more things:

## Improve the doco

Could you make the following improvements to the `mididevice` setting's description? It's pretty much copy & paste ready 😄 

```cpp
	str_prop->SetOptionHelp(MidiDeviceName::SoundCanvas,
	                        "  soundcanvas:  The internal Roland SC-55 synthesiser (requires a CLAP audio\n"
	                        "                plugin that implements the Sound Canvas to be available;\n"
	                        "                see the [soundcanvas] section).");

	str_prop->SetOptionHelp(MidiDeviceName::FluidSynth,
	                        "  fluidsynth:   The internal FluidSynth MIDI synthesizer (SoundFont player)\n"
	                        "                (requires the FluidSynth dynamic-link library to be available;\n"
	                        "                see the [fluidsynth] section).");
```

## Fix debug build

There is something wrong when I'm using a debug build of Staging. Everything is the same between the below two runs, I'm only using a Release x64 Visual Studio build first, then a Debug build.

### Release build

I'm executing `midedevice fluidsynth` from the DOS prompt after startup:

image

### Debug build

With the debug build, the DLL gets loaded and things are going fine for a while, but the `fluid_synth_sfload()` call throws an exception. I logged that in my local build, the message just says "unknow error", then I did not investigate further. But something is wrong here.

image

shermp:
To be fair, I don't think mixing release and debug libraries is the best of ideas, but I'll see if I can reproduce. 

Are you just copying the debug fluidsynth library into a release dosbox directory without any of the debug glib and friends libraries?

johnnovak:
> To be fair, I don't think mixing release and debug libraries is the best of ideas, but I'll see if I can reproduce.
> 
> Are you just copying the debug fluidsynth library into a release dosbox directory without any of the debug glib and friends libraries?

Yeah, so this is not about "mixed mode" usage; I just want the thing to work with debug build at all. Yesterday it worked, but after your refactorings from today it no longer does.

I'm pretty sure I checked out `main`, build the debug version which made `libfluidsynth-3.dll` appear in the `vs\x64\Debug` folder, then I rebuild using your branch and used that DLL.

But I noticed `libfluidsynth-3.dll` has the same size in both the debug and release VS builds. So not exactly sure what's going on in here.

shermp:
Ok, so something definitely up with debug builds. Will investigate.

shermp:
The issues with debug builds in Windows also occurs on `main`. Looks like some potential path encoding shenanigans going on.

shermp:
Found the culprit, this little beauty here:

````cpp
std::error_code err = {};
const auto canonical_path =
	std_fs::canonical(sf, err).c_str();

if (err) {
	return {};
} else {
	return canonical_path;
}
```

Note that the function returns a `std_fs::path`.

So, the code gets a canonical path, turns it into a "native" path string, stores that to a variable. Then on function return, it's implicitly turned back into a path again.

The fix is to not call `c_str()` on the canonical path.

shermp:
Fix implemented. Feel free to cherry-pick it to `main` if you like.

I think this was a lifetime issue. `std_fs::canonical(sf, err).c_str()` returns a pointer to a temporary string, and because the variable it was being saved to was declared with `const auto`, it remained a pointer, rather than a new string being constructed.

Debug builds probably clear or randomizes memory upon the end of (stdlib) object lifetime, but probably not in release builds, which is probably why it "worked" in release builds.

johnnovak:
> Fix implemented. Feel free to cherry-pick it to `main` if you like.
> 
> I think this was a lifetime issue. `std_fs::canonical(sf, err).c_str()` returns a pointer to a temporary string, and because the variable it was being saved to was declared with `const auto`, it remained a pointer, rather than a new string being constructed.
> 
> Debug builds probably clear or randomizes memory upon the end of (stdlib) object lifetime, but probably not in release builds, which is probably why it "worked" in release builds.

Thanks for figuring this out 🎉 I think it's fine if it goes in as part of this PR.

weirddan455:
Yeah, that's a dangling pointer. @johnnovak it's also a recent regression from one of yours bdf615d099d6bea84d15949e72978c4f334c7180

If you've included that commit as v0.82.1, I would cherry-pick @shermp 's fix here as well.

I don't have this crash on a debug build in Linux but the "fun" part about undefined behavior is that sometimes it accidentally works and sometimes it stops working like when @shermp refactored a completely different part of this file :laughing:

johnnovak:
> Yeah, that's a dangling pointer. @johnnovak it's also a recent regression from one of yours [bdf615d](https://github.com/dosbox-staging/dosbox-staging/commit/bdf615d099d6bea84d15949e72978c4f334c7180)
> 
> If you've included that commit as v0.82.1, I would cherry-pick @shermp 's fix here as well.

Cheers, I've just checked and stuff isn't backported (it's not possible because it came after my grand MIDI device refactor after I shoehorned in the Sound Canvas).
https://github.com/dosbox-staging/dosbox-staging/pull/4106

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +43,224 @@
...
+static DynLibResult load_fsynth_dynlib()
+{
+	using namespace fsynth;
+	if (!fsynth_lib) {
+		fsynth_lib = dynlib_open(fsynth_dynlib_file);
+		if (!fsynth_lib) {
+			return DynLibResult::LibOpenErr;
+		}
+		DB_GET_FSYNTH_SYM(fluid_version);
+        DB_GET_FSYNTH_SYM(fluid_set_log_function);
```
johnnovak, line 220: Indentation

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +43,224 @@
...
+		DB_GET_FSYNTH_SYM(fluid_synth_program_change);
+		DB_GET_FSYNTH_SYM(fluid_synth_channel_pressure);
+		DB_GET_FSYNTH_SYM(fluid_synth_pitch_bend);
+		DB_GET_FSYNTH_SYM(fluid_synth_sysex);
+		DB_GET_FSYNTH_SYM(fluid_synth_write_float);
+
+        /* Keep ERR and PANIC logging only */
+        for (auto level : {FLUID_DBG, FLUID_INFO, FLUID_WARN}) {
+            fluid_set_log_function(level, nullptr, nullptr);
+        }
```
johnnovak, line 259: Indentation

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +43,224 @@
...
+		DB_GET_FSYNTH_SYM(fluid_synth_noteon);
+		DB_GET_FSYNTH_SYM(fluid_synth_key_pressure);
+		DB_GET_FSYNTH_SYM(fluid_synth_cc);
+		DB_GET_FSYNTH_SYM(fluid_synth_program_change);
+		DB_GET_FSYNTH_SYM(fluid_synth_channel_pressure);
+		DB_GET_FSYNTH_SYM(fluid_synth_pitch_bend);
+		DB_GET_FSYNTH_SYM(fluid_synth_sysex);
+		DB_GET_FSYNTH_SYM(fluid_synth_write_float);
+
+        /* Keep ERR and PANIC logging only */
```
johnnovak, line 256: minor: line comment?

src/midi/midi_fluidsynth.cpp
```diff
@@ -44,6 +43,224 @@
...
+ * @param data User data pointer supplied to fluid_set_log_function().
+ */
+typedef void (*fluid_log_function_t)(int level, const char *message, void *data);
+
+/**
+ * FluidSynth dynamic library handle
+ */
+static dynlib_handle fsynth_lib = {};
+
+/* The following function pointers will be set to their corresponding symbols in the FluidSynth library */
```
johnnovak, line 134: minor: line comment

johnnovak:
@kklobe Would you be able to test this on your Mac? Haven't bought my new Mac Mini yet.

FeralChild64:
Checked current version on my Linux system (both debug and release builds) - works correctly.

FeralChild64:
... but on my macOS I can't setup the _FluidSynth_ even on the main branch code.

```ini
soundfont        = FluidR3_GM.sf2
soundfont_dir    = 
```

results in (on the main branch):

```
fluidsynth: error: fluid_sfloader_load(): Failed to open '����': File does not exist.
fluidsynth: error: Unable to open file '����'
fluidsynth: error: Failed to load SoundFont "����"
2025-03-10 20:35:27.778 | FSYNTH: Error loading SoundFont ''
2025-03-10 20:35:27.778 | MIDI: Error opening device 'fluidsynth'; using 'port'
```

shermp:
> ... but on my macOS I can't setup the _FluidSynth_ even on the main branch code.
> 
> ```ini
> soundfont        = FluidR3_GM.sf2
> soundfont_dir    = 
> ```
> 
> results in (on the main branch):
> 
> ```
> fluidsynth: error: fluid_sfloader_load(): Failed to open '����': File does not exist.
> fluidsynth: error: Unable to open file '����'
> fluidsynth: error: Failed to load SoundFont "����"
> 2025-03-10 20:35:27.778 | FSYNTH: Error loading SoundFont ''
> 2025-03-10 20:35:27.778 | MIDI: Error opening device 'fluidsynth'; using 'port'
> ```

This is likely to be the same bug found with debug builds on Windows.

@FeralChild64 can you please tell me what the library name (including extension) is on Mac OS? It's likely I have the name wrong in this PR.

johnnovak:
> ... but on my macOS I can't setup the _FluidSynth_ even on the main branch code.

Yeah, as @shermp explained, can you test with the release build as well?

FeralChild64:
Tried the official 0.82.0 release build (a `dmg` file) - _FluidSynth_ works there. Same config file with the current main:

```
fluidsynth: error: fluid_sfloader_load(): Failed to open '�R�oY': File does not exist.
fluidsynth: error: Unable to open file '�R�oY'
fluidsynth: error: Failed to load SoundFont "�R�oY"
2025-03-11 22:20:13.474 | FSYNTH: Error loading SoundFont 'FluidR3_GM.sf2'
2025-03-11 22:20:13.474 | MIDI: Error opening device 'fluidsynth'; using 'port'
```

(previous log was wrong, it seems I did not save the config file after changes - probably I've pressed _Ctrl+S_ instead of _Super+S_ - I mistake keyboard shortcuts frequently and I have no idea how to set the _macOS_ to use _Windows_-style shortcuts)

The build created from this PR can't use the _FluidSynth_ for me either, error is:

```
FSYNTH: Failed to load FluidSynth library
MIDI: Error opening device 'fluidsynth'; using 'port'
```

Tried the `brew install fluidsynth` command, but it did not help.

Frankly, I have no idea where to expect the _FluidSynth_ dynamic library on _macOS_, and how it could be named, or even if this `brew` command should install the dynamic linking library at all. I always thought _Macintosh_ was using `FooBar.Library` pattern for dynamic libraries (just like the _Amiga OS_), but it seems this is not the case. Searching for _*fluidsynth*_ (case insensitive), using the _Midnight Commander_, did not find anything in `/Library` or `~/Library`.

kklobe:
I needed to add the library name for macOS, e.g.:

```cpp
#if defined(WIN32)
constexpr const char fsynth_dynlib_file[] = "libfluidsynth-3.dll";
#elif defined(__APPLE__)
constexpr const char fsynth_dynlib_file[] = "libfluidsynth.3.dylib";
#else
constexpr const char fsynth_dynlib_file[] = "libfluidsynth.so.3";
#endif
```

For homebrew, the file is located in /opt/homebrew/lib, and that seems to need to be in the launch directory. I set an alias for the release and debug binaries, so just copying the dylib to the binary directory doesn't work; I need to copy the dylib to each directory from which I wish to launch, so I would suggest retrieving the path of the launched dosbox and trying to load it from there.

kklobe:
P.S. recent macOS versions restrict setting DYLD_LIBRARY_PATH in the environment, so unfortunately that doesn't work to "help" `dlopen()` find it.

johnnovak:
So, just to make sure we're all on the same page, we're doing this to avoid building FluidSynth in every build because it slows down things massively (because it pulls in glib and related GNUisms on Windows, and probably to some extent on macOS too as that's a BSD). It's not about supporting system-provided FluidSynth on Windows and macOS (Linux is different).

On **macOS** and **Windows,** we expect the FluidSynth library to be present in the same directory as the Staging executable. We'll put the library binaries into our official packages via CI. For local dev, it's pretty easy to copy the library file into the build dir and make sure rebuilding doesn't clean/delete it.

On **Linux**, the library file will be either loaded from the directory where the Staging executable resides, or if not present `/lib` or `/usr/lib` will be used (the full lookup order is described [here](https://man7.org/linux/man-pages/man3/dlopen.3.html)). So this caters for our official package scenario where we put the `fluidsynth-3.so` into the .tar.gz package file, and in distro-provided packages the library will be dynamically loaded from standard system locations.

On all OSes, if the library cannot be resolved, we fall back to using the system-level MIDI device. So that's a pretty consistent behaviour across platforms, with some concessions on the Linux side for the system-level FluidSynth scenario.

johnnovak:
> (previous log was wrong, it seems I did not save the config file after changes - probably I've pressed _Ctrl+S_ instead of _Super+S_ - I mistake keyboard shortcuts frequently and I have no idea how to set the _macOS_ to use _Windows_-style shortcuts)

@FeralChild64 
System Preferences, enter "modifier keys" in the search box, click "Customise modifier keys" then change it here:

image

As you can see, I've swapped my Caps Lock and Control keys. You can remap pretty much any mod key to any other here.

kklobe:
> On **macOS** and **Windows,** we expect the FluidSynth library to be present in the same directory as the Staging executable. We'll put the library binaries into our official packages via CI. For local dev, it's pretty easy to copy the library file into the build dir and make sure rebuilding doesn't clean/delete it. Sure, you can play around with `DYLD_LIBRARY_PATH`, symlinks, or whatnot, but what we really care about is this standard "library in the exe folder" path.
 
ok, after killing and restarting my terminal, it is now finding it properly - scratch my previous complaint (but the code still needs to be updated for the correct macOS dylib name).

johnnovak:
> Frankly, I have no idea where to expect the _FluidSynth_ dynamic library on _macOS_, and how it could be named, or even if this `brew` command should install the dynamic linking library at all. I always thought _Macintosh_ was using `FooBar.Library` pattern for dynamic libraries (just like the _Amiga OS_), but it seems this is not the case. Searching for __fluidsynth__ (case insensitive) using the _Midnight Commander_, did not find anything in `/Library` or `~/Library`.

It does for system libs. But brew deps are "their own little universe", so to speak, for the most part. Hence they live in ` /opt/homebrew/lib`. Those `.Library` libs are truly system-level, but if any 3rd party package manager interfered with the "truly system level" deps, all hell would break loose.

Technically, you can even install brew and MacPorts version of the same libs that you have at the system level, they you have three different variants of each (but I don't recommend this 😆). In practical terms, you use brew only for CLI stuff. So all your brew-installed CLI stuff will use the brew deps, and that's it (same with MacPorts). But yeah, even some brew stuff can use _fundamental_ dynamic system libs that Apple ships with the OS (all the properietary Apple frameworks for UI, audio, graphics, etc.)

shermp:
I've added the MacOS library name. Thanks for letting me know what it's called @kklobe

FeralChild64:
@johnnovak OK, good to know - but modifier keys is one piece of the puzzle, we have all these _Alt+Tab_ (here I think some app is needed), _Alt+F4_, _F10_, etc. shortcuts which won't work on the _Mac_.

On _Linux_ with _KDE_ changing the shortcut scheme is easy; on _Macintosh_ or _Windows_, not really.

Screenshot

johnnovak:
> @johnnovak OK, good to know - but modifier keys is one piece of the puzzle, we have all these _Alt+Tab_ (here I think some app is needed), _Alt+F4_, _F10_, etc. shortcuts which won't work on the _Mac_.
> 
> On _Linux_ with _KDE_ changing the shortcut scheme is easy; on _Macintosh_ or _Windows_, not really.
> 
> Screenshot

Oh, you'll never get that level of identical shortcuts on the Mac 😆 And the UI paradigms are slightly different too, so IMO it's pointless. E.g. you use Cmd + ` (tilde) to cycle between multiple windows of the same app, which is a Mac paradigm that has no Windows equivalent.

You just need to get used to the different paradigm and shortcuts, and ingrain it in your "muscle memory". You're far better off that way that trying to turn macOS into Windows. I don't think it takes more than a couple of weeks of using macOS. I switch between Windows and macOS all the time, and the different shortcuts are never a problem for me. It's just like learning anything new; give it some time.

IMO, longer term you're far better of if you have the mindset to appreciate macOS for its unique qualities.

FeralChild64:
It always takes me a quite a while to get used when I switch between the GUI paradigms, so I want consistency among the OSes I use. It's not a problem for me to switch to _macOS_ shortcuts (in fact, I consider them superior from the ergonomic standpoint), I can even get used to something as crazy as _Hyprland_ - but changing the environment is painful for me then.

Unfortunately, I can't escape from _Windows_, I have to use it at work, usually 40 hours a week.

johnnovak:
@shermp So looks like the first phase of this is done, which is loading the library at runtime.

The next phase would be building the artifacts and making them available for download, then the Staging github project can download those artifacts in the CI workflows and inject them into the packages.

Do you want me to set up a new project for that, e.g., `dosbox-staging-data` or something, then you can perhaps give it a shot building the FluidSynth dynamic library on all three platforms via vcpkg in a simple CI job?

weirddan455:
The simplest thing to do would be to grab a pre-compiled binary but it looks like they only make binaries for Windows and Android.

I see two options then:

1. Fork the FluidSynth repo at a stable release and setup a CI job like you were saying. I'm not sure how vcpkg would fit in here though. It wouldn't really be a dependency anymore and we would probably remove it from both Meson wrap and vcpkg in Staging. We would just be building a standalone .dll in a separate repo and it would probably be easier to use FluidSynth's native build system (CMake).

2. Don't make a separate repo. Add FluidSynth as a submodule in our build system so that it builds a dynamic library but does not link to it and then optionally copies it where it needs to go for Staging to find it at runtime. Put it behind a build flag that's off by default and Staging's CI turns that flag on. It would be more tightly integrated that way but it would take CI longer to build (though probably no worse than it is now).

shermp:
> The simplest thing to do would be to grab a pre-compiled binary but it looks like they only make binaries for Windows and Android.

Also, their binaries look to be compiled using GCC. I'm wary of combining libraries built with different compilers and standard libraries.

FeralChild64:
@johnnovak I don’t think I’m the right person to be asked about the approach - I always use system-supplied libraries, unless I have a very goot reason not to do so.

But in general I prefer using GIT submodules over forking the project.

johnnovak:
> > The simplest thing to do would be to grab a pre-compiled binary but it looks like they only make binaries for Windows and Android.
> 
> Also, their binaries look to be compiled using GCC. I'm wary of combining libraries built with different compilers and standard libraries.

Yeah, we won't try to do that, that's asking for trouble. We'll build our own, 100%.

> Fork the FluidSynth repo at a stable release and setup a CI job like you were saying. 

But I did not say that 😆

> I'm not sure how vcpkg would fit in here though.

In a separate repo, we just run `vcpkg install fluidsynth` to build FS for the three OSes we support in an on-demand CI job. The job uploads the generated artifacts which then we can download munually and create a release manually for it in that other repo. Or maybe even that can be automated; but that's just detail.

Then our DOSBox Staging CI loop just downloads those binary artifacts and injects them into our ZIP packages.

It's simple and it works. We'd need to do this manual build & release step in the other repo on FS upgrades only.

Why vcpkg? Because you can only build FS using MSYS2 on Windows with the glib crap installed, and the official Visual Studio install steps also start with "install vcpkg so you can install glib". Check it out here:

https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake#building-with-visual-studio-on-windows

So we're just going around in circles; MSYS2 is a requirement (`vcpkg install glib` pulls in MSYS2 as we know).

So for that reason the submodule option would accomplish nothing.

weirddan455:
I didn't fully think through the glib mess. Your plan seems like the best way to go then @johnnovak 

An issue I just thought of is that we can have conflicting glib symbols with SLIRP. I recall that being an issue in the past when we had MSYS2 static builds in CI. When you have SLRIP being linked statically and FluidSynth being linked dynamically and both depend on glib, it sometimes goes boom depending on the OS and versions.

Maybe we get lucky and it's not a problem but it needs testing. I've only tested this branch with system libraries. We may need to replace SLRIP with the de-glibified version that someone mentioned in Discord (I think it was @kklobe )

johnnovak:
> I didn't fully think through the glib mess. Your plan seems like the best way to go then @johnnovak
> 
> An issue I just thought of is that we can have conflicting glib symbols with SLIRP. I recall that being an issue in the past when we had MSYS2 static builds in CI. When you have SLRIP being linked statically and FluidSynth being linked dynamically and both depend on glib, it sometimes goes boom depending on the OS and versions.

Yeah, that's a good point. I think the easiest thing is to generate the slirp binaries too off the same vcpkg baseline, and job done. Then someone will only need to "dynlibify" our usage of slirp the same way as @shermp did for FS.

> Maybe we get lucky and it's not a problem but it needs testing. I've only tested this branch with system libraries. We may need to replace SLRIP with the de-glibified version that someone mentioned in Discord (I think it was @kklobe )

Also an option, but that did not seem to be actively maintained, and we don't wanna take that on if it can be avoided. So just using vcpkg for slirp too seems to be the most attractive option to me that gets the job done with the lowest effort.

weirddan455:
> I think the easiest thing is to generate the slirp binaries too off the same vcpkg baseline, and job done.

We might need to dynamically link glib as well. We don't need the manual `dlopen`s since we don't use it directly but it may need to be a separate .dll that FluidSynth and SLRIP both share. Not 100% sure on that but we'll know after testing it.

johnnovak:
> > I think the easiest thing is to generate the slirp binaries too off the same vcpkg baseline, and job done.
> 
> We might need to dynamically link glib as well. We don't need the manual `dlopen`s since we don't use it directly but it may need to be a separate .dll that FluidSynth and SLRIP both share. Not 100% sure on that but we'll know after testing it.

Yeah I tested this, but I can't remember whether I needed to copy `glib-2.0-0.dll` over too along with `libfluidsynth-3.dll` built by vcpkg. If both FS and SLIRP built by vcpkg have glib as a transitive dynamic dependency, we just need to copy that too, so it's easy. Then if they're statically link against it, it's even easier. We'll see, but it's easy either way.

shermp:
At long last I've finally figured out the (infamous?) `X` macro. I've switched to using one of those for the symbol declaration/resolving.

With the previous code, it was possible to define a function pointer, call it, but never actually resolve the symbol. Now that can't happen.

johnnovak:
@shermp Actually, this has just occurred to me now: can't we just copy the FluidSynth header files only into the project verbatim to get the compilation to succeed? Is that not possible with C++ compilers? If we don't need the .dll/.so/.dynlib to be present at link time, this would be the simplest solution, no?

weirddan455:
> can't we just copy the FluidSynth header files only into the project verbatim to get the compilation to succeed?

It would work for struct definitions and such. The functions would need to be re-defined to function pointers though or we'll get linker errors when we stop linking to it at build time.

johnnovak:
> > can't we just copy the FluidSynth header files only into the project verbatim to get the compilation to succeed?
> 
> It would work for struct definitions and such. The functions would need to be re-defined to function pointers though or we'll get linker errors when we stop linking to it at build time.

Okay, makes sense. Wasn't sure it worked like that. IMO, the way dynamic linking is implemented in C/C++ kinda sucks—it's only "half-dynamic". To be "fully dynamic", you'd need to be able to compile your dynamically linked stuff _without_ having access to the dynamic library binaries at compile/linking time. Then all the function pointer resolution would happen at runtime when you load the app, not at the time of linking.

Anyway, it is what it is 🤷🏻

shermp:
Actually, it is possible to do this on Linux at least, I think you can instruct the GCC linker to ignore unresolved symbols at compile time. I don't know about other platforms though.

weirddan455:
Linkers are complicated. The symbol resolution does happen at runtime. On Linux, the compiled binary has a table for dynamically linked symbols that calls into `ld.so` (runtime linker) and the table gets patched at runtime to replace that one time call with a memory address inside the library.

One reason it needs the library .so at build time is to fetch versioning information. That way if the ABI changes and the user has an old version, it can throw an error rather than doing some undefined behavior.

On Linux, everything is inside the .so file. On Windows, there's separate .dll for runtime and .lib for build time linker information but I don't really know the details of what exactly it needs at build time.

It's not really a fault of C++. That's just the way the OS and ELF and EXE file formats work.

johnnovak:
> It's not really a fault of C++. That's just the way the OS and ELF and EXE file formats work.

Yeah, the more I think about it, the more I arrive to the usual "everything is terrible" conclusion, lol 😆

johnnovak:
I've started the experimentation here, the Windows path is basically 95% working.
https://github.com/dosbox-staging/dosbox-staging-ext

I've also figured out which dynamic libraries need to be injected. That's for both FluidSynth and SLIRP, so @shermp if you could do the SLIRP conversion too, that would be great 😎 Then when that's done that I'll do the rest of the CI tomfoolery.

So these are the libs we'll need to pull in in a ZIP:

```
glib-2.0-0.dll
iconv-2.dll
intl-8.dll
libfluidsynth-3.dll
pcre2-8.dll
slirp-0.dll
```

I've determined those by doing a build on Windows with FluidSynth and SLIRP enabled, then one without them, the diffing te resulting set of DLLs.

### FluidSynth and SLIRP enabled

```
glib-2.0-0.dll
iconv-2.dll
iir.dll
intl-8.dll
libfluidsynth-3.dll
libpng16.dll
libspeexdsp.dll
mt32emu-2.dll
ogg.dll
opus.dll
pcre2-8.dll
SDL2.dll
SDL2_net.dll
slirp-0.dll
zlib-ng2.dll
```

### FluidSynth and SLIRP disabled

```
iir.dll
libpng16.dll
libspeexdsp.dll
mt32emu-2.dll
ogg.dll
opus.dll
SDL2.dll
SDL2_net.dll
zlib-ng2.dll
zlib1.dll
```

johnnovak:
Thanks for the tips @PatTheMav. We're "outsourcing" all this library building stuff to vcpkg—for good reason, it's a rabbit hole to get things right on three different platforms.

These tips might come in handy, but our situation is a lot simpler as vcpkg lets us sidestep all these problems; it's the vcpkg package maintainer's job to get all this right. I'm just moving the exact same vcpkg steps to build FluidSynth and SLIRP into a separate repo, so we'll get the exact same dynamic libraries as before. Like for like, so there's no room for things to be different than in our current build pipeline which just works fine.

You can see it here, I got the Windows path kinda/sort working yesterday. But I'll probably also upload the debug variants of the libraries, just in case:

https://github.com/dosbox-staging/dosbox-staging-ext/commit/fc3a089bdb1455b0148913d71b14943ce91c2984#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R72-R123

PatTheMav:
Yeah you benefit from `vcpkg` finally being able to handle that - iirc proper support for RPATH handling was only finished in mid-2024, which was years after we started down our path (we also wanted more control to ensure better compatibility with Xcode).

johnnovak:
> Yeah you benefit from `vcpkg` finally being able to handle that - iirc proper support for RPATH handling was only finished in mid-2024, which was years after we started down our path (we also wanted more control to ensure better compatibility with Xcode).

Yeah we'd like Xcode compatibility for the reasons you outlined, but this would be a show-stopper:

> OBS cannot be launched as a "bare binary" anymore - which would be Linux-style - but can only exist as an application bundle

Then again, with vcpkg and CMake we can actually treat macOS "just like another Linux" for the most part, which is nice (easier, and does the job well enough).

johnnovak:
Alright gents, I've created a [new repo](https://github.com/dosbox-staging/dosbox-staging-ext) where we'll build these external deps. Check out the README.

I've managed to get things to work on all three platforms. I've tested Windows so far.

- @kklobe can you please test this branch on macOS with the deps downloaded from the below URL?
- @weirddan455 @FeralChild64 can you please test this branch on Linux with the deps downloaded from the below URL?

https://github.com/dosbox-staging/dosbox-staging-ext/actions/runs/14153196881

If all is good, I'll make the necessary changes to our CI workflow, then @shermp can cherry-pick into this branch and we can merge it. Then the next stop is SLIRP! 🚀

weirddan455:
There's a reference to fluidsynth that needs to be removed from Meson: `grep fluidsynth meson_options.txt`

weirddan455:
> can you please test this branch on Linux with the Linux deps downloaded from the below URL?

First of all, I had to hard-code a path to the library. Linux doesn't search the current directory for libraries like Windows does. We'll need to decide what folder we want these to live in and then code some logic in this PR to prepend that path. If we don't find it there, we can fall back to using the system lib.

Secondly, there's a regression from the Meson wrap. It's also in the `main` CMake build so I filed a separate issue #4264

Lastly, it's interesting how Linux handles `dlload`. I can delete all the extra deps like glib and it'll fall back to using system deps. If SLIRP is in the mix, it uses my system glib regardless because I guess it can tell that the symbols are already loaded. If I compile without SLIRP, it prefers to use the provided glib over my system one.

`libasound.so.3` is never used. Fluidsynth is linking to it but it always uses the system library because we're also linking to `libasound` directly for ALSA MIDI. It would be nice to remove that as a FluidSynth dependency. The Meson wrap doesn't link to it. I think we trimmed down some of the options.

If vcpkg has a way to tweak compile options, I'd recommend whatever feature from FluidSynth that depends on ALSA. We don't need it.

PatTheMav:
Relative path functionality is also available on Linux - you can use an `RPATH` entry like `$ORIGIN/` or `$ORIGIN/../lib` which tells the dynamic library loader where else to look for the required library apart from the default library search paths.

Using `readelf --dynamic <binary>` you can inspect if a `RUNPATH` is encoded within the binary. Though some distributions discourage using this feature for system packages as they'd effectively allow "library injection" and it's expected that any package relies on distribution-provided dependencies only for dynamic linking.

weirddan455:
@PatTheMav Yeah, and that's exactly what the `libfluidsynth` is doing in the binary @johnnovak built:

` 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]`

So the dependencies of `libfluidsynth` use relative paths explaning the behavior I described above. But the DOSBox Staging binary does not have a `RUNPATH` set so we'll either need to set that with build flags or just have some extra runtime logic to search the right directory.

johnnovak:
> First of all, I had to hard-code a path to the library. Linux doesn't search the current directory for libraries like Windows does. We'll need to decide what folder we want these to live in and then code some logic in this PR to prepend that path.

Wouldn't just prepending the Linux library name string with `./` in the code solve it? Or we can use `./lib/` on Linux only. Or are relative paths forbidden too?

johnnovak:
Okay, so the best solution seems to set the `RPATH` thing to `$ORIGIN` which will resolve dynamic libraries from the directory the executable was started from as first priority. Passing `-Wl,-rpath='$ORIGIN'` to the linker when building the Staging executable will do the trick.

Here's to do it with CMake:

```cmake
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath='$ORIGIN'")
```

(Source: https://serverfault.com/a/926072)

PatTheMav:
> Here's to do it with CMake:
> 
> ```cmake
> set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath='$ORIGIN'")
> ```
> 
> (Source: https://serverfault.com/a/926072)

There's a few issues with this particular suggestion:

* Linker flags should not hardcode something like `-Wl`, as those are linker-dependent. Instead `LINKER:-rpath='$ORIGIN'` should be used to ensure the right vendor-specific prefix is used (otherwise this would only succeed with GCC but not with clang)
* Any flag (whether it's compiler flags or linker flags) should be set as target properties via the appropriate `target_compile_options`/`target_link_options` functions so that CMake itself can merge them with other (possibly global) compile/link options
* In this specific case just setting the `INSTALL_RPATH` target property would achieve the desired result without having to futz around with linker command lines.

johnnovak:
Thanks a lot for the suggestions @PatTheMav. As you can see, we're not exactly CMake experts here 😅 "Learning on the job", as the saying goes 😏  

> otherwise this would only succeed with GCC but not with clang)

Yeah, we need both Clang and GCC support.

>     * Any flag (whether it's compiler flags or linker flags) should be set as target properties via the appropriate `target_compile_options`/`target_link_options` functions so that CMake itself can merge them with other (possibly global) compile/link options

Good tips, cheers 👍🏻 

>     * In this specific case just setting the `INSTALL_RPATH` target property would achieve the desired result without having to futz around with linker command lines.

Looks like we have a winner here—the easier, the better, agreed 💯

PatTheMav:
> Probably it doesn't get simpler & cleaner than this:
> 
> ```cmake
> set(CMAKE_INSTALL_RPATH $ORIGIN)
> ```

Yeah that sets a global default variable - many of those prefixed with `CMAKE_` are then set as properties on targets without the prefix, which in turn can be changed individually by changing the target property.

Note that the `RPATH` setting on the binary is only changed when `cmake --install` is used to install the target. You can check the `cmake_install.cmake` file and see that it will contain fix-up code for the respective platform.

There is a boolean target property `BUILD_WITH_INSTALL_RPATH` that can make CMake build the binary directly with the desired RPATH, though depending on your setup it might break the ability to run/debug the project from inside the build tree.

---

I've created this for OBS Studio's plugin developers as an "entry point" to learn about modern CMake, maybe you find some of it helpful: https://github.com/obsproject/obs-plugintemplate/wiki/CMake-Build-System-Guide

Craig Scott also published great help articles (he's a regular in KitWare's bug tracker): https://crascit.com

johnnovak:
> There is a boolean target property `BUILD_WITH_INSTALL_RPATH` that can make CMake build the binary directly with the desired RPATH, though depending on your setup it might break the ability to run/debug the project from inside the build tree.

Yeah that seems to be the best option then; I don't wanna `install`. Thanks for the CMake guide links, I'll check them out some time 👍🏻

johnnovak:
@shermp Can you please rebase? That should "unstuck" those those macOS Clang CI jobs.

shermp:
> @shermp Can you please rebase? That should "unstuck" those those macOS Clang CI jobs.

Done

johnnovak:
@shermp Just wanna understand & summarise what's outstanding to get this over the line. Let me know if there's anything else:

- Set the RPATH thing for the Linux build of DOSBox Staging only so we can load dynamic libraries from the current directory just by passing the library name to `dlopen()` (see [this comment](https://github.com/dosbox-staging/dosbox-staging/pull/4222#issuecomment-2764829514)). But maybe specifying a path with a directory component is enough (so `lib/blah.so` instead of just `blah.so`).
- Create a release in the https://github.com/dosbox-staging/dosbox-staging-ext project with the FS dynamic libs, then download that release ZIP file in our CI workflows and inject the contents into the final release packages.

Is there anything else? My preference is to do FluidSynth end-to-end in the PR, merge, then do SLIRP in a follow-up PR.

shermp:
So, a part of me is wondering if this is all worth the headache and extra complication. I know I've moaned (frequently!) about building `glib` with vcpkg but really it's not the end of the world. This is basically trading occasional slow build speed on Windows MSVC versus the complications of cross-platform runtime dynamic linking and packaging, and I'm not sure the tradeoff is worth it.

johnnovak:
> So, a part of me is wondering if this is all worth the headache and extra complication. I know I've moaned (frequently!) about building `glib` with vcpkg but really it's not the end of the world. This is basically trading occasional slow build speed on Windows MSVC versus the complications of cross-platform runtime dynamic linking and packaging, and I'm not sure the tradeoff is worth it.

Yeah, it's worth it. And we're almost there, so it's a bit strange to stop just before the finish line after we've done all this work...

One of the big problems is bisecting on Windows; you potentially need to wait for 20-30 minutes if the vcpkg baseline has changed.

Also, I'm fine to take over and finish it if you don't feel like it. Just wanted input on what's outstandin from your POV.

shermp:
If vcpkg wasn't so sensitive to changes, I'd love if we could just package up the `vcpkg_installed` dir and call it a day...

johnnovak:
> If vcpkg wasn't so sensitive to changes, I'd love if we could just package up the `vcpkg_installed` dir and call it a day...

Ok, so back to my question, anything else outstanding from your POV?

The RPATH thing is a one-liner and the CI injection like 5-10 lines, to put things into perspective. Very minimal amount of work is left to do. Unless you're aware of something else?

src/midi/midi_fluidsynth.cpp
shermp: No longer relevant.

src/midi/midi_fluidsynth.cpp
shermp: What's wrong with this?

johnnovak: Seemed like the indentation was off, but probably I was confused. Probably I did not see the full context, just 2-3 preceding liens.

shermp:
@johnnovak have you reviewed this since I switched to using `X` style macros?

johnnovak:
> @johnnovak have you reviewed this since I switched to using `X` style macros?

Just had a look now. It seems to be a safer method than the previous one. I'm a pragmatic person, so if a `#define` gets us better maintainability, I'm fine with that. Thanks for doing this improvement 👍🏻

shermp:
> > @johnnovak have you reviewed this since I switched to using `X` style macros?
> 
> Just had a look now. It seems to be a safer method than the previous one. I'm a pragmatic person, so if a `#define` gets us better maintainability, I'm fine with that. Thanks for doing this improvement 👍🏻

Thanks for taking a look. Yeah, the #define approach while a bit dirty seems safer because you can't forget to resolve function symbols you've declared.

johnnovak:
> > > @johnnovak have you reviewed this since I switched to using `X` style macros?
> > 
> > 
> > Just had a look now. It seems to be a safer method than the previous one. I'm a pragmatic person, so if a `#define` gets us better maintainability, I'm fine with that. Thanks for doing this improvement 👍🏻
> 
> Thanks for taking a look. Yeah, the #define approach while a bit dirty seems safer because you can't forget to resolve function symbols you've declared.

Yeah all I care about is:

- Simplicity
- Maintainability
- End result

I have no interest in ideology or collecting imaginary brownie points by adhering to "standards" 😆 

You've done the right thing and have picked the best overall solution for the job 💯

johnnovak:
@shermp I've done an experiment to disable SLIRP and FluidSynth altogether in the CI builds, and now I can answer your "is this all worth it?" question with a resounding **YES**, now backed up by actual numbers.

## Windows

- With FS and SLRIP, the Windows CI builds took consistently 50 to 70 minutes to complete. That's pretty long!
- With FS and SLIRP completely removed, we're in the 10 to 20 minutes range! Huge improvement!!! 🥳 

[This one](https://github.com/dosbox-staging/dosbox-staging/actions/runs/14632879197) took 13 minutes and the previous one 16 minutes built off my experimental [jn/no-glib](https://github.com/dosbox-staging/dosbox-staging/tree/jn/no-glib) branch. Check out earlier runs from other branches; they're all in the 50 to 70 minute range 🥶 

## macOS

The macOS CMake build situation was't that bad, but it was still in the 15-30 minute range. Usually, it was below the 20-minute mark.

With FS and SLIRP disabled, my [last run](https://github.com/dosbox-staging/dosbox-staging/actions/runs/14633169445) took 9 minutes and the [previous run](https://github.com/dosbox-staging/dosbox-staging/actions/runs/14632879202) was 7 minutes. That's a roughly 50% improvement! 🥳 

## Linux

The Linux situation is inconclusive, it's a wash. The [last two CMake builds](https://github.com/dosbox-staging/dosbox-staging/actions/runs/14633169435) with SLIRP & FS disabled were 11 minutes, prior to that we have numbers from 10 to 20 minutes. So seems like there's no change.

--

So by removing FluidSynth and SLIRP from our regular CI builds, all platforms are in the 10 to 20 minute range, just like we had before with Meson.

shermp:
To be fair, we do cache VCPKG built binaries on the Windows CI, and should do the same for the others. That works pretty well for the most part.

johnnovak:
> To be fair, we do cache VCPKG built binaries on the Windows CI, and should do the same for the others. That works pretty well for the most part.

Doesn't seem to work, though? Maybe it caches something, but it doesn't seem to help much — check my numbers!

johnnovak:
@shermp Could you please rebase this to `main` to get the new CMake Linux workflow, and the do the change to include the header files as per our Discord discussion?

As discussed, including the headers is the easier option for SLIRP, so we might as well do it for FluidSynth as well to make our approach to side-loading dynamic libraries consistent.

In the meantime, I'm working out the RPATH setting for the Linux build and injecting the binaries from the -ext project into the final packages. I'll just do those on a separate branch, then you'll be able to cherry-pick them onto this branch.

Just a few puzzle pieces are left, then let's get this merged this week! 🚀

src/midi/midi_fluidsynth.h
```diff
@@ -23,19 +23,25 @@
...
 #include <optional>
 #include <thread>
 #include <vector>
 
+#include "dynlib.h"
 #include "mixer.h"
 #include "rwqueue.h"
 #include "std_filesystem.h"
 
+namespace fsynth {
```
johnnovak, line 38: @shermp Can you please change this `namespace FluidSynth`? We use UpperCamelCase for namespaces (e.g. `namespace SoundCanvas`).

src/libs/include/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Fluidsynth
+
+Headers are sourced from the automated GitHub archive from the 2.2.3 release. 
+The contents of the `include` directory have been copied into this directory.
+
+The original `fluidsynth.cmake` file has been renamed to `fluidsynth.h` and 
+the conditional logic for `BUILD_SHARED_LIBS` was removed and replaced with a 
+bare `#define FLUIDSYNTH_API`. Additionally, `#include "fluidsynth/version.h"` 
+has been commented out.
```
johnnovak, line 24: Very nice doco @shermp 🎖️ 🚀 💯

johnnovak:
@shermp The external dynamic library injection is 100% working now; it has been tested end-to-end on Windows, macOS, and Linux.

Please cherry pick my four commits from my `jn/fluidsynth-dynamic-load-3` branch. Make sure to rebase your branch to lastest `main` first before doing the cherry-picking. 

After that, can you please address my minor naming related comment I made today on your PR, then let's merge this! 🎉

src/midi/midi_fluidsynth.h
shermp: Done.

shermp:
> @shermp The external dynamic library injection is 100% working now in my branch based on this PR; it has been tested end-to-end on Windows, macOS, and Linux 🥳
> 
> Please rebase your branch to lastest `main` first, then cherry pick my four commits from my `jn/fluidsynth-dynamic-load-3` into this PR.
> 
> After that, can you please address my minor naming related comment I made today on your PR, then let's get this in! 🎉

Rebased, cherry-picked and namespace changed to `FluidSynth`

I've taken the PR out of draft now.

src/libs/include/README.md
shermp: Thanks

johnnovak (approved):
Nice job @shermp, we got there in the end 🥳 I'm happy with the current state of this, so unless we want to give it another round of testing, hit the big green button 😎 (just remove "[WIP]" from the title first).

shermp:
Title changed. I don't have permission to merge though, so someone else will have to do that.

PR #4223, opened 2025-03-08
Improve default crossfeed setting behaviour

johnnovak:
# Description

Setting `crossfeed = on` previously enabled crossfeed on all active stereo channels with the default strength of 40. In retrospect, this wasn't such a great behaviour as it also enabled crossfeed for the `SB`, `GUS`, `MT32`, etc. channels which is rarely wanted.

This bit me in the ass when I was testing audio stuff in the last few days and haven't even realised for a while...

I only wanted to enable crossfeed on the OPL channel. Okay, so let's do this then in the global config:

```ini
[mixer]
crossfeed = on

[autoexec]
mixer x0 OPL x40
```

So `crossfeed = on` enabled crossfeed on all stereo channels with the default strength of 40, then `mixer x0` sets it to 0 on all active channels, then `OPL x40` sets it 40 on the OPL channel only. Well, this is already pretty convoluted and not very user friendly, but wait, it get worse...

If I launch a game with a local config that mounts a CD with `IMGMOUNT`, the CD mixer channel will became active _after_ my `mixer x0` command in the global config (remember, all global mixer commands only set stuff for the currently active channels, which is reasonable). So what happens is that when my `CD` channel pops into existence, it gets a crossfeed of 40 😐

IMO, this is all rather confusing and not that great, so I'm simplifying the crossfeed presets a bit.

In practice, 9 times out of 10 you only wanna set crossfeed for the OPL and CMS channels which typically feature hard-panned sounds. So I'm changing the behaviour of the `crossfeed` setting to do only apply to those two channels. Then people can add crossfeed to other mixer channels manually to their liking, but they should need that very rarely 
(e.g., a few GUS games that use hard-panned sounds, such as Pinball Dreams).

I've also enhanced the `chorus` and `reverb` setting descriptions a bit.

# Release notes

Enabling crossfeed (e.g., by setting `crossfeed = on` in the config) now only applies the crossfeed effect to the **OPL** and **CMS** channels. These are the two devices that typically feature hard-panned sounds. Previously, crossfeed presets applied to all stereo channels, but this was rarely wanted.  You can apply crossfeed to other channels as well with the `MIXER` command if you want (e.g., `MIXER GUS x50`).

# Manual testing

With `crossfeed = on` in the global config and mounting some ISOs in the local config, only the OPL channel has crossfeed enabled:

image

Crossfeed is only applied to Dual OPL2 and OPL3, but not to OPL2:

image

The **CMS** channel has crossfeed applied to it too when enabled:

image

### Setting descriptions

image

image

image

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -3063,7 +3063,11 @@ static void init_mixer_dosbox_settings(Section_prop& sec_prop)
 	        "           features lots of white noise).\n"
 	        "  normal:  Normal chorus that works well with a wide variety of games.\n"
 	        "  strong:  An obvious and upfront chorus effect.\n"
-	        "Note: You can fine-tune each channel's chorus level using the MIXER.");
+	        "\n"
+	        "Notes:\n"
+	        "  - The presets apply the chorus effect to the synth channels only (except\n"
+	        "    for synths with built-in chorus; e.g. the Roland MT-32).\n"
+	        "  - Use the MIXER command to fine-tune the chorus levels per channel.");
 	string_prop->Set_values({"off", "on", "light", "normal", "strong"});
 
 	MAPPER_AddHandler(handle_toggle_mute, SDL_SCANCODE_F8, PRIMARY_MOD, "mute", "Mute");
--- a/src/hardware/mixer.cpp
+++ b/src/hardware/mixer.cpp
@@ -347,7 +347,11 @@ static void set_global_crossfeed(const MixerChannelPtr& channel)
 {
 	assert(channel);
 
-	if (!mixer.do_crossfeed || !channel->HasFeature(ChannelFeature::Stereo)) {
+	const auto apply_crossfeed = (channel->GetName() == ChannelName::Opl &&
+	                              channel->HasFeature(ChannelFeature::Stereo)) ||
+	                             (channel->GetName() == ChannelName::Cms);
+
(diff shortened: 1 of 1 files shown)


src/hardware/mixer.cpp
```diff
@@ -347,7 +347,10 @@ static void set_global_crossfeed(const MixerChannelPtr& channel)
 {
 	assert(channel);
 
-	if (!mixer.do_crossfeed || !channel->HasFeature(ChannelFeature::Stereo)) {
+	const auto apply_crossfeed = (channel->GetName() == ChannelName::Opl) ||
+	                             (channel->GetName() == ChannelName::Cms);
+
+	if (!mixer.do_crossfeed || !apply_crossfeed) {
```
weirddan455, line 353: I think you may need to leave the Stereo check in. OPL2 is mono:

https://github.com/dosbox-staging/dosbox-staging/blob/f65827490fac96577addb4afeb5734ad4dc9fdbe/src/hardware/opl.cpp#L841-L845

johnnovak: Ah yeah, forgot about that.

johnnovak: Fixed. Also updated the top description with a test case.

johnnovak:
Ready for final review @weirddan455

weirddan455 (approved):
Looks good

PR #4228, opened 2025-03-12
Allow graceful failure if OpenGL context could not be created

### LINKED ISSUE #4215: Allow graceful failure if OpenGL context could not be created

b39cfc00a2a40040cb65828148aab39068ef6f6c

Linux x86_64

Debian 12

### Relevant hardware info

_No response_

It was brought up on Discord where someone was trying to launch DOSBox as headless using `SDL_VIDEODRIVER=dummy` and getting a fatal error that OpenGL could not be created.

Workaround is to both compile with `-Duse_opengl=false` and set `output = texture` and `texture_renderer = software`.

It would be good to make the `-Duse_opengl=false` compile flag not necessary.  It looks like the only reason we're creating the OpenGL context is to try to set SRGB.  Code block in question is below:

https://github.com/dosbox-staging/dosbox-staging/blob/b39cfc00a2a40040cb65828148aab39068ef6f6c/src/gui/sdlmain.cpp#L1313-L1339

Theoretically this problem is not limited to headless mode.  It could happen if there's no suitable OpenGL drivers or hardware.  If the user is trying to use texture mode, we should ideally fail gracefully (set SRGB to "not supported" and move on).

1. Set the config as mentioned below to not use OpenGL.
2. Launch with `SDL_VIDEODRIVER=dummy dosbox`
3. DOSBox fails to launch with an error about OpenGL.

_No response_

### Your configuration

```ini
[sdl]
output = texture
texture_renderer = software
```

### Provide a log

```text

```

- [x] Yes, I agree.

weirddan455:
# Description

These should be pretty safe changes. I removed some code leftover from an old workaround (removed in 3d7852b71e0d2874057162025bd477f8403b4f32) that was lingering around being useless and added a fallback to create a non-OpenGL window if we're using the Texture backend.

See commit messages for more details.

## Related issues

Fixes #4215

# Release notes

Fixes a crash when using "output = texture" with non-functional OpenGL drivers or hardware.
SDL_VIDEODRIVER=dummy now works if you want to use DOSBox heedlessly for some reason.

# Manual testing

- SDL_VIDEODRIVER=dummy works even if `output = opengl`. There was existing code to fallback to `output = texture` that was broken due to the temporary OpenGL context.
- `output = opengl` and `output = texture` continue to work as before with normal drivers.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1313,11 +1313,10 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 #if C_OPENGL
 		if (rendering_backend == RenderingBackend::OpenGl) {
 			flags |= SDL_WINDOW_OPENGL;
-		}
-
-		if (SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, 1)) {
-			LOG_ERR("OPENGL: Failed requesting an sRGB framebuffer: %s",
-			        SDL_GetError());
+			if (SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, 1)) {
+				LOG_ERR("OPENGL: Failed requesting an sRGB framebuffer: %s",
+						SDL_GetError());
+			}
 		}
 #endif
 		if (!sdl.desktop.window.show_decorations) {
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1315,23 +1315,6 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 			flags |= SDL_WINDOW_OPENGL;
 		}
 
-		// We need a context to query the vendor string.
-		const auto temp_window = SDL_CreateWindow(
-		        "", 0, 0, 200, 200, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
-		if (temp_window == nullptr) {
-			LOG_ERR("SDL: Failed to create temporary window: %s",
-			        SDL_GetError());
-			return nullptr;
-		}
-		const auto temp_context = SDL_GL_CreateContext(temp_window);
-		if (temp_context == nullptr) {
-			LOG_ERR("OPENGL: Failed to create temporary context: %s",
-			        SDL_GetError());
-			return nullptr;
-		}
-
-		SDL_GL_DeleteContext(temp_context);
-		SDL_DestroyWindow(temp_window);
 		if (SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, 1)) {
 			LOG_ERR("OPENGL: Failed requesting an sRGB framebuffer: %s",
 			        SDL_GetError());
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1330,6 +1330,14 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 
 		assert(sdl.window == nullptr); // enusre we don't leak
 		sdl.window = SDL_CreateWindow("", pos.x, pos.y, width, height, flags);
+		if (!sdl.window && rendering_backend == RenderingBackend::Texture && (flags & SDL_WINDOW_OPENGL)) {
+			// opengl_driver_crash_workaround() call above conditionally sets SDL_WINDOW_OPENGL.
+			// It sometimes gets this wrong (ex. SDL_VIDEODRIVER=dummy).
+			// This can only be determined reliably by trying SDL_CreateWindow().
+			// If we failed to create the window, try again without it.
+			flags &= ~SDL_WINDOW_OPENGL;
+			sdl.window = SDL_CreateWindow("", pos.x, pos.y, width, height, flags);
+		}
 		if (!sdl.window) {
 			LOG_ERR("SDL: Failed to create window: %s", SDL_GetError());
 			return nullptr;


src/gui/sdlmain.cpp
```diff
@@ -1315,23 +1315,6 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
...
-		}
-		const auto temp_context = SDL_GL_CreateContext(temp_window);
-		if (temp_context == nullptr) {
-			LOG_ERR("OPENGL: Failed to create temporary context: %s",
-			        SDL_GetError());
-			return nullptr;
-		}
-
-		SDL_GL_DeleteContext(temp_context);
-		SDL_DestroyWindow(temp_window);
```
johnnovak, line 1334: Yeah AFAIK this is the "standard" solution to query the vendor string if you want to initialise your real context differently depending on the vendor.

Maybe refactor into a function called `get_opengl_vendor`? Could save some dicking around later if we need it. But then we might just migrate to the new SDL3 shader thing and not use OpenGL directly anymore.

I'm 50/50, I'll leave it to you.

src/gui/sdlmain.cpp
```diff
@@ -1330,6 +1330,14 @@ static SDL_Window* SetWindowMode(const RenderingBackend rendering_backend,
 
 		assert(sdl.window == nullptr); // enusre we don't leak
 		sdl.window = SDL_CreateWindow("", pos.x, pos.y, width, height, flags);
+		if (!sdl.window && rendering_backend == RenderingBackend::Texture && (flags & SDL_WINDOW_OPENGL)) {
+			// opengl_driver_crash_workaround() call above conditionally sets SDL_WINDOW_OPENGL.
+			// It sometimes gets this wrong (ex. SDL_VIDEODRIVER=dummy).
+			// This can only be determined reliably by trying SDL_CreateWindow().
+			// If we failed to create the window, try again without it.
+			flags &= ~SDL_WINDOW_OPENGL;
```
johnnovak, line 1338: I'm uneasy about adding such hacks. Did this random behaviour happen to you during testing your PR? If so, which OS and SDL versions? It would be best to put that info into the comment.

Not sure if we're on latest SDL2 either. If not, have you tried with the latest?

Lastly, do the SDL guys know about this? If you found and SDL issue tracker ticket, we should put a link in the comment. If not, I highly recommend to report it.

src/gui/sdlmain.cpp
weirddan455: I don't understand your comment. I'm removing this vendor querying code because we're not using it. We were previously checking for "Intel" because at some point Intel drivers had problems with SRGB. That was removed in 3d7852b71e0d2874057162025bd477f8403b4f32 because I guess newer drivers don't have that problem.

src/gui/sdlmain.cpp
weirddan455: This is "undoing" the hack in `opengl_driver_crash_workaround()`. That's a DOSBox functions that checks if the default render driver is OpenGL and, if so, adds the OpenGL flag to window creation.

Comment states:
```
// This is a hack to prevent SDL2 from re-creating window internally. Prevents
// crashes on Windows and Linux, and prevents initial window from being visibly
// destroyed (for window managers that show animations while creating window,
// e.g. Gnome 3).
```
I have experienced the window flicker before when working with SDL but not the crashes. When SDL creates a Renderer (for Texture backend), it needs the window to be set with the appropriate flag. If it's not, it will destroy and re-create the window. `opengl_driver_crash_workaround()` tries to avoid that by conditionally setting the OpenGL flag.

IMO the cleaner way to solve the window flicker is to create the window as hidden and then un-hide it after the Renderer has been created. That might not fix the crashes but I don't know what the conditions have to be to re-produce that.

For this PR, I didn't want to touch that hack that's been in the code for 5 years (introduced in e50fc3135ffe959f67537e1fdecee27010df2fb9).

This code I added only triggers if window creation failed. It's reproducible with `SDL_VIDEODRIVER=dummy`.

johnnovak: Thanks for the explanation. You know what, maybe it's not worth dwelling on this too much. Realistically, we will upgrade to SDL2 in a 2-3 years anyway, and if this hack was good for 5 years, it will be good for a few more 😆 

Just learned we could use the new SDL3 GPU API with compiled shader binaries right now via SDL_shadercross.
https://github.com/libsdl-org/SDL_shadercross

So we don't even need to wait to the SDL shader language to materialise, we can just insert SDL_shadercross into our pipeline and ship with compiled shaders technically right now. That makes SDL3 adoption very appealing (e.g. to get Vulkan support for free, and then all direct OpenGL calls will 100% go away).

src/gui/sdlmain.cpp
johnnovak: Yeah, I was thinking we'd maybe need to check vendor strings again in 2 months... for reasons 😄 But maybe don't leave code there "just in case", agreed. I get it that currently it's unused code.

weirddan455: Oh, I see what you're saying now. The code I removed doesn't even have the part that gets the vendor string. That part was removed earlier. All that's left is creating and destroying a temp window and context. It's not even a fully functional block if we wanted to refactor :laughing:

src/gui/sdlmain.cpp
weirddan455: Well the hack only does anything for Texture backend (where SDL might internally use OpenGL). When we're using OpenGL directly, we always set the `SDL_WINDOW_OPENGL` flag for obvious reasons.

> Just learned we could use the new SDL3 GPU API with compiled shader binaries right now via SDL_shadercross.
> https://github.com/libsdl-org/SDL_shadercross

We'd need to change our shader format. We're currently using GLSL and that tool states "It takes SPIRV or HLSL as the source and outputs DXBC, DXIL, SPIRV, MSL, or HLSL."

>  That makes SDL3 adoption very appealing (e.g. to get Vulkan support for free, and then all direct OpenGL calls will 100% go away).

Downside is that SDL3's GPU API does not support OpenGL so we'd be cutting support for older hardware. Damn near every PC has support for OpenGL. Less so for Vulkan.

src/gui/sdlmain.cpp
johnnovak: Right, I just skimmed it. In that case, remove it with _extreme prejudice_ 😎 🤘🏻

src/gui/sdlmain.cpp
johnnovak: > We'd need to change our shader format. We're currently using GLSL and that tool states "It takes SPIRV or HLSL as the source and outputs DXBC, DXIL, SPIRV, MSL, or HLSL."

Yeah, but it's trivial to change from one format to the other at the source level for most simple stuff like ours. Not a real issue.

> Downside is that SDL3's GPU API does not support OpenGL so we'd be cutting support for older hardware. Damn near every PC has support for OpenGL. Less so for Vulkan.

Ah, fuck, you're right. Forgot about that. Well, I guess we want to use the SDL3 GPU path eventually, but that means we'll use that _AND_ our home-made OpenGL path as well for GLSL support... until we wanna carry that around. Then at some point we just pull the plug and bye bye OpenGL.

johnnovak:
> @johnnovak Was there any changes you wanted me to make here? We kind of got sidetracked talking about SDL 3.

Yeah 😄 It's good to go, resolved all the discussions just now.

weirddan455:
@johnnovak Can you hit approve then? :smile:

johnnovak:
> @johnnovak Can you hit approve then? 😄

_- Could you tell me the time?
- Yes. (and walks away)_

Approved. Just my morning caffeine levels are dangerously low 😆

johnnovak (approved):
:shipit:

Issue #3008, opened 2023-10-15
OPL3 hanging note issue with multiple games


Grounded0:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

macOS (Intel)

### Relevant hardware info

2017 15" MacBook Pro

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

California Games II consistently triggers OPL3 hanging note issue. Finally a solid repro case?

Other test cases:

- **Beneath a Steel Sky** intro (quite loud constant note)
- **Doom E2M2** ([real hardware recording](https://youtu.be/B1WP-LPw1Bk?feature=shared&t=1654))
- **Passport to Adventure -- Monkey Island demo** (several notes are left hanging forever at a very low volume, sounds like a chord)
- **Frederik Pohl's Gateway**,  **Gateway II: Homeworld** -- hanging note in the background after exiting to DOS
- **Wizardry 6 -- Bane of the Cosmic Forge** (when OPL audio is selected)
- **Tetris Classic (1992)** -- Weird buzzing noise after each levels intro music has finished.
- **The 1st Degree (Win 3.1, 1995)** -- MIDI Setup triggers this when selecting either test song and pressing Play button to test the MIDI output (via Win 3.x OPL MIDI driver).

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Run California Games II
2. Quit the game
3. Observe OPL3 hanging note issue

### Your configuration

```ini
[sblaster]

sbtype              = sb16
sbbase              = 220
irq                 = 7
dma                 = 1
hdma                = 5
sbmixer             = true
sbwarmup            = 100
oplmode             = auto
sb_filter           = modern
sb_filter_always_on = true
opl_filter          = off
cms_filter          = on
```

### Provide a Log

```text
date       time         | 
2023-10-15 14:33:04.015 | arguments: dosbox
2023-10-15 14:33:04.015 | Current dir: /Users/ajp/Library/Preferences/DOSBox
2023-10-15 14:33:04.015 | stderr verbosity: 0
2023-10-15 14:33:04.015 | -----------------------------------
2023-10-15 14:33:04.015 | dosbox-staging version 0.80.1
2023-10-15 14:33:04.015 | ---
2023-10-15 14:33:04.015 | LOG: Loguru version 2.1.0 initialized
[...]

PR #4230, opened 2025-03-13
Add denoiser to filter out residual OPL and Sound Canvas mk1 noise

johnnovak:
# Description

## OPL hanging note issue

I'm happy to report that finally I have the definitive answer to the mysterious "OPL hanging note" issue (see https://github.com/dosbox-staging/dosbox-staging/issues/3008; paging @Grounded0 and @0mnicydle 😆). First we suspected this might be Staging or NukedOPL3 bug, but nukeykt confirmed this is actually the result of accurate low-level hardware emulation (see my NukedOPL3 [issue ticket](https://github.com/nukeykt/Nuked-OPL3/issues/35)). I couldn't really test this on hardware because the noise is just below the relatively high noise floor of my AWE32 board, but it's quite audible in good quality headphones in Staging where there is no hiss to mask the OPL noise.

Here's nukeykt's explanation of the issue:

> OPL chips use bitwise inversion to negate operator output for negative part of sine, so 0 becomes -1, 1 -> -2, etc, so small oscillation between 0 and -1 is expected even when envelope generator is muted.

All these residual oscillations between 0 and -1 absolute sample values can add up to a periodic noise signal in the [-9, 0] range on the OPL2, and [-18, 0] on the OPL3 even when all channels are muted. And indeed, this residual noise in my in-game Gods OPL audio capture is in the [-8, 0] range (that's the raw NukedOPL3 16-bit output, so with the master highpass filter disabled and no extra volume scalings):

gods

This is how it looks zoomed-in. The 9 operators are oscillating at slightly different frequencies, so they add up to a periodic waveform that sounds like a buzzing chord if you raise the volume in quiet passages. The sound of the "chord" changes depending on the state of the oscillators and you don't hear it in all games -- it depends on how the game writes to the OPL registers to set up the sounds.

opl-noise-pattern

Here's nukeykt's real-hardware recording that also showcases the issue (video starts from a part where the noise is audible):

**Doom E2M2** ([real hardware recording](https://youtu.be/B1WP-LPw1Bk?feature=shared&t=1654))

## Noise gate

The solution is to use a noise gate that is finely tuned to only remove this noise and leave the useful signal alone. Noise gates are standard audio processors; they attenuate the signal below a given threshold, and let it pass through "unharmed" above that threshold. The opening/closing of the gate can be tweaked with the attack and release controls so it's not abrupt but more gradual.

## Sample recordings

Here are the recordings of Gods (OPL) and DOOM (Sound Canvas) that illustrate the residual noise. Gods is a very good test case for the noise gate as the in-game sound effects are very quiet, barely above the residual noise level.

**doom-noise.wav** is just the decay part of the sound amplified so you can hear the SC-55 artifacts better.

🔴🔴🔴 _**Warning: loud audio!**_ 🔴🔴🔴
[recordings.zip](https://github.com/user-attachments/files/19228411/recordings.zip)

## Sound Canvas mk1 noise

Turns out the original Sound Canvas ("mk1") is also plagued by a noise issue where the decaying notes have a weird distorted metallic ring to them, and after the sound has finished playing, quite often there is some kind of noise between the notes where there should be silence. This noise comes and goes; sometimes it's on both channels, sometimes on only one, sometimes it's gone for a while. This is also accurate hardware behaviour that NukedSC55 faithfully emulates; I did notice the noise when I recorded those 50+ SC-55 game soundtracks from my hardware SC-55 two years ago.

The good news is we can use the noise gate to filter out the SC-55 noise as well, we only need to use slightly different settings. The first picture shows the spectral distribution of the noise (there's a DC element as well that the gate filters out too). The second diagram shows the original waveform on the left side (it should decay into silence, but it doesn't), and on the right side (yellow background) you can see the same waveform with the noise removed by the noise gate.

sc-55-spectrum

sc-55-noise

## Related issues

Addresses https://github.com/dosbox-staging/dosbox-staging/issues/3008

# Release notes

A new `denoiser` setting has been added (enabled by default) that removes residual noise from the OPL synth's and the Roland Sound Canvas' output. As these devices are very accurately emulated, it's unavoidable to also emulate some of their undesirable quirks, namely some extra residual noise between sounds where there should be silence. This can be heard in some games on good-quality headphones if you have sensitive ears. The denoiser (technically, a noise gate) has been finely tuned so it doesn't degrade the useful signal in any way.

The noise issue only affects the original "mk1" Sound Canvas models (in all games). The OPL noise issue is only triggered by some games. Here are a few examples:

- **Beneath a Steel Sky** — audible constant noise during the intro
- **Doom** — chord-like noise at the start of the E2M2 level music
- **Frederik Pohl's Gateway, Gateway II: Homeworld** — hanging note after exiting to DOS
- **Gods** — audible noise in-game between the quiet sound effects
- **Tetris Classic (1992)**  — buzzing noise after each level's intro music has finished playing
- **Wizardry 6: Bane of the Cosmic Forge** — hanging notes after sound effects

# Manual testing

image

### OPL

- Tested all OPL games from https://github.com/dosbox-staging/dosbox-staging/issues/3008 with the denoiser disabled to check the noise is audible, then confirmed it is gone with the denoiser enabled.
- Tested Doom E2M2 music configured for the OPL (start new game, then enter `idclev22` to jump to that level).
- Tested toggling the denoiser from the DOS prompt then restarting the game to make sure the change took effect.

### Sound Canvas

- With the denoiser disabled, start DOOM configured for Sound Canvas music and start a new game.
- Go to the options menu and increase the music volume to the maximum briefly, then quickly decrease it to zero.
- Note the strange noise at the end of the decay, and the constant hiss after the sound has completely decayed (might need to turn up the volume to hear it).
- Repeat with the denoise enabled; now the weird noise at the end of the decay is minimised and there should be no hiss after the decay, just complete silence.
- Tested toggling the denoiser from the DOS prompt then restarting the game to make sure the change took effect.
- Confirmed the denoiser does its job on all mk1 models.
- Confirmed there is sound with the mk2 models with the denoiser enabled.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


weirddan455:
Your first commit 85d579a167acc70cd0d0e8df67fc8bda8e49df51 doesn't compile. The top of your branch does but I recommend fixing this because it's annoying to hit something that doesn't build when you're doing a bisect.

```
In file included from ../../src/audio/compressor.cpp:29:
../../include/mixer.h:44:10: fatal error: Iir.h: No such file or directory
   44 | #include <Iir.h>
```

src/hardware/mixer.cpp
```diff
@@ -1293,6 +1298,38 @@ void MixerChannel::SetLowPassFilter(const FilterState state)
 	}
 }
 
+void MixerChannel::ConfigureNoiseGate(const float threshold_db, const float attack_time_ms,
+                                      const float release_time_ms)
+{
+	assert(attack_time_ms > 0.0f);
```
weirddan455, line 1304: You might want to add a mutex here, especially if you plan to extend support for this.

It's currently fine because this only gets called from OPL and SoundCanvas constructors which take the global `MIXER_LockMixerThread()`.

If you ever want to add support for changing this by the `MIXER` command line program (or any other time at runtime) it'll need a lock.

src/hardware/mixer.cpp
```diff
@@ -1293,6 +1298,38 @@ void MixerChannel::SetLowPassFilter(const FilterState state)
...
+	noise_gate.attack_time_ms  = attack_time_ms;
+	noise_gate.release_time_ms = release_time_ms;
+
+	InitNoiseGate();
+}
+
+void MixerChannel::EnableNoiseGate(const bool enabled)
+{
+	LOG_MSG("%s: Noise gate %s", name.c_str(), (enabled ? "enabled" : "disabled"));
+	do_noise_gate = enabled;
```
weirddan455, line 1317: Same comment here. It's currently fine. Since this is just a bool, you can make it `std::atomic` instead of using a mutex to allow for future runtime changes.

src/hardware/mixer.cpp
```diff
@@ -3126,6 +3138,17 @@ static void init_mixer_dosbox_settings(Section_prop& sec_prop)
...
 	string_prop->Set_values({"off", "on", "light", "normal", "strong"});
 
+	bool_prop = sec_prop.Add_bool("denoiser", WhenIdle, DefaultOn);
+	bool_prop->Set_help(
+	        "Remove low-level residual noise from the output of the OPL synth and the Roland\n"
+	        "Sound Canvas. The emulation of these devices is accurate to the original\n"
+	        "hardware units, which includes the emulation of a very low-level semi-random\n"
+	        "noise. Although this is authentic, most people would find it slightly annoying.\n"
+	        "  off:  Disable the denoiser.\n"
+	        "  on:   Enable the denoiser on the OPL and SOUNDCANVAS channels (default).\n"
```
weirddan455, line 3148: Minor and not entirely unique to this config option: You've documented this as "on/off" but when you generate a new config it defaults to "true". true/false still works as expected though.

johnnovak:
> Your first commit [85d579a](https://github.com/dosbox-staging/dosbox-staging/commit/85d579a167acc70cd0d0e8df67fc8bda8e49df51) doesn't compile. The top of your branch does but I recommend fixing this because it's annoying to hit something that doesn't build when you're doing a bisect.

Cheers, and yes, I agree. I did check that every commit compiles at some point, but must have broken it when fuxing up CI failures.

weirddan455 (commented):
I've looked through it all now. Code looks mostly fine. A few (maybe overly paranoid) comments about thread safety. In non-performance sensitive areas I'd rather have a lock that I don't need than have to debug race conditions later though.

src/hardware/mixer.cpp
```diff
@@ -1319,38 +1326,56 @@ static int clamp_filter_cutoff_freq([[maybe_unused]] const std::string& channel_
...
-
 	const auto cutoff_freq_hz = clamp_filter_cutoff_freq(name, _cutoff_freq_hz);
 
-	assert(order > 0 && order <= MaxFilterOrder);
-	for (auto& f : filters.highpass.hpf) {
-		f.setup(order, mixer.sample_rate_hz, cutoff_freq_hz);
-	}
-
 	filters.highpass.order          = order;
 	filters.highpass.cutoff_freq_hz = cutoff_freq_hz;
```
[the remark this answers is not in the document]
johnnovak: Yeah, good point. I believe when I removed it this I changed things so it could be only called from a channel and never from the mixer. But then I kept changing things and forgot to put it back.

I agree, it's best to make everything safe. That's the point of thread safety 😄

src/audio/noise_gate.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+	threshold_value = std::exp2(threshold_db / 6.0f);
+
+	const auto sample_rate_hz = static_cast<float>(_sample_rate_hz);
+
+	attack_coeff = 1.0f /
+	               std::pow(10.0f, 1000.0f / (attack_time_ms * sample_rate_hz));
+
+	release_coeff = 1.0f /
+	                std::pow(10.0f, 1000.0f / (release_time_ms * sample_rate_hz));
```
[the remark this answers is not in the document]
johnnovak: Haha, actually I have no idea how it works 😆 I just adapted to code of that open source gate effect; mostly straightforward stuff, but this part is a bit of a mystery...

src/hardware/mixer.cpp
johnnovak: I think a good rule would be to guard all public mixer and channel methods by mutexes, no? Mutexes like these are extremely fast and I'm pretty sure they're no-ops most of the time.

What do you think? I think that would be 100% safe and enforcable. Considering every single pupuethodb lic method individually requires thinking, and thinking about threading is error-prone.

src/hardware/mixer.cpp
johnnovak: Yeah I changed all descriptions to on/off from true/false a while ago. It's more user friendly. I'll look into writing the defaults as on and off instead.

Btw, 1, enabled, on, true are all valid "positive" options, then 0, none, off, disabled, and false are the "negative" variants.

src/hardware/mixer.cpp
weirddan455: Actually a mutex would probably be faster since this function is rarely called. That way the mixer can read this variable non-atomically since it acquires the mutex first anyway. Very unlikely to cause a measurable difference either way though.

src/hardware/mixer.cpp
weirddan455: That's probably a good rule. I was throwing around mutexes pretty liberally when people started complaining about thread sanitizer warnings. Especially for things like changing config settings that get called very rarely, there's not much of a downside. The mixer mutexes are also recursive so you don't have to think about it too much.

>  I'm pretty sure they're no-ops most of the time

Not necessarily no-ops but a good mutex implementation will do an atomic compare and swap to see if it can acquire it straight away. Failing that, they'll often do a small busy-wait loop to see if it becomes quickly available before making the expensive/blocking syscall to wait.

johnnovak: Yeah, I think this is good advice:
https://stackoverflow.com/a/66506828

Quoting the relevant bits but maybe read the whole thing 😄 

> Mutexes are considered "heavy" because they are often believed to result in a syscall, i.e. a round-trip to the kernel. A trip to the kernel takes on the order of 1,000+ CPU cycles due to context switching between privileged and unprivileged code.

> In many OSes these days mutexes are optimized to not go to the kernel until a contention occurs. [...] However, once there is a contention, there will be a trip to the kernel.

> In practice, however, a mutex will provide adequate balance between performance and convenience. So I wouldn't worry about it unless you are counting nanoseconds (as in HFT or real-time applications).

Some audio guys advocate the use of complicated spinlocks and atomics and don't wanna go near mutexes:

https://timur.audio/using-locks-in-real-time-audio-processing-safely

Well, there's not a single real-world benchmark or measurement in that article, so I view it with suspicion. In either case, I think using mutexes with rarely called methods is a no brainer.

weirddan455: > Some audio guys advocate the use of complicated spinlocks and atomics and don't wanna go near mutexes:

There's definitely some truth to that. You generally can't use mutexes when you're talking about actual real-time guarantees. Of course you also need a real-time kernel to make those guarantees.

It also matters for low latency but for what it's worth PulseAudio doesn't work reliably at the 128 sample block-size that article is talking about. I sometimes get stutters if I try to go that low and I tested that on a small test program I wrote as well outside of DOSBox that actually used a lock-free queue. PipeWire and JACK are supposed to be better but it's mostly professional audio software that cares about latencies that low.

> Well, there's not a single real-world benchmark or measurement in that article, so I view it with suspicion.

You can look at our Voodoo code for a real-world example of where locks slowed us down :smile:

Anyway, for the purposes of our mixer, these locks are fine. The one place I might consider a lock-free queue is inside SDL's callback. That runs on the actual audio thread and we really don't want to be blocking there. There was a bug in Mac OSX I ran into when writing the threaded mixer where sound would cut out and never come back if we blocked for too long (it happened if you paused so a very long time of blocking).

The DOSBox mixer thread is not as sensitive to blocking. It's doing a lot of expensive stuff anyway like resampling and generating OPL audio. The important part is that it doesn't block SDL's callback thread from running while it does that. SDL can only get blocked from the single `RWQueue BulkEnqueue()` it does at the end.

johnnovak: Oh yeah, I get it that there are faster thread synchronisation techniques than locks & mutexes; I guess what I'm wondering at what point does the difference matter. Audio rates are quite slow compared to what you can do on a CPU when performing calculations at full speed, so my guess is some DSP guys are overplaying the drawbacks of locks.

> You can look at our Voodoo code for a real-world example of where locks slowed us down 

Voodoo going at full speed vs calling a mutex even at every audio frame, so 48 thousand times a second, I don't think they're comparable scnearios. Pure computational code can be called hundreds of thoursand of millions of times a second, so yeah, in those cases the difference can be significant.

Yeah good points on using queues instead of blocking and trying to minimise or fully remove the contention arising from accessing the queue from multiple threads simultaenously.

As you said, the nice thing about async queue-based communication like that is that it removes the whole concept of blocking from the equation. I guess all these public methods that can be called both from the channel and the mixer could just put a command into a queue, then the mixer would process those commands when it's safe to do so. So yeah, that would be better, but at the same a pretty big overkill, so I think we can just keep using mutexes for these rarely called methods.

weirddan455: > Audio rates are quite slow compared to what you can do on a CPU when performing calculations at full speed, so my guess is some DSP guys are overplaying the drawbacks of locks.

You're conflating latency with throughput here. Your audio thread shouldn't be doing much more than a `memcpy()` especially if it's real-time safe like the article you linked was talking about. It's low throughput but it needs to be low latency when you're dealing with small blocksizes. That's the tradeoff real-time makes in general is exchanging throughput for low latency.

> Most of the time they use locks in response to some UI interaction, and those happen at extremely slow rates compared to the CPU, and infrequently.

No, you wouldn't ever block your audio thread for UI events. Like I said before, the audio thread should really only be reading from a queue.

In DOSBox, the SDL callback is the real audio thread and it basically is just "lock + memcpy + unlock". You can think of it as the sound card saying "hey I need some audio right now please" and it gets very grumpy if you make it wait. The lock is not ideal but it's fine as long as you're not doing low latency or real-time.

The mixer thread (which is what we're talking about here) is allowed to be slow and block sometimes. We exchange some latency for this because our mixer has its own buffer on top SDL's buffer but it gives us well... a buffer to insulate us because we're ideally generating audio at least a blocksize or two in advance of when SDL will need it.

weirddan455: Just to be clear, SDL's callback runs from its own thread (what I'm calling the "real" audio thread, separate from the mixer thread) and the only lock there is inside the `RWQueue` that SDL reads from and the mixer thread outputs to.

All the explicit mutex locks in `mixer.cpp` are between the mixer thread and the main thread and they don't matter to the same degree as locking a "real" audio thread.

johnnovak: Yeah, I get it it's about not blocking for long, but my point was, if the mutex doesn't take long (like in the microsecond range, which is normal), and you call it infrequently between buffer refills (very sporadically, and only a couple of times now and then), that can only possibly adds a few microseconds of latency, which most likely doesn't matter.

E.g. I'm getting ~2.5-3.0 ms latency when using 64 sample buffers @ 48kHz in REAPER with my MOTU interface:

image

So I doubt adding a random 0 to 20 microsecond noise on top of that would change anything, but the devil is always in the details, of course 😄 And once we're blocked even for just 1 ms, we're screwed, no doubt about that 😆

weirddan455: > And once we're blocked even for just 1 ms, we're screwed, no doubt about that 😆

Well, yeah that's the problem. The article you linked was specifically talking about realtime in an OS context which means you must run in some deterministic time. Blocking for an undermined amount of time isn't allowed at all. I've heard some people say syscalls in general aren't allowed but I think that depends on the OS and syscall.

If you were to take a lock, you can have priority problems. Linux has a special PI futex to deal with that but I don't think you can get that from a plain old `std::mutex`.

> Linux supports priority-inheritance (PI) futexes in order to
> handle priority-inversion problems that can be encountered with
> normal futex locks.  Priority inversion is the problem that occurs
> when a high-priority task is blocked waiting to acquire a lock
> held by a low-priority task, while tasks at an intermediate
> priority continuously preempt the low-priority task from the CPU.
> Consequently, the low-priority task makes no progress toward
> releasing the lock, and the high-priority task remains blocked.

https://man7.org/linux/man-pages/man2/futex.2.html

Mutexes are generally fast but "generally" doesn't fly in a realtime context.

johnnovak: > Linux has a special PI futex to deal with that but I don't think you can get that from a plain old std::mutex.

Yeah, I guess we'd need to check out the sources (if we need to). Some people claim `std::mutex` maps to futexes in current implementations. But then you're at the mercy of the compiler I guess, so I see why people prefer to be explicit about it for really critical stuff. Fortunately, that's not our use case in the mixer 😄

weirddan455: On Linux, `std::mutex` is just a thin wrapper over `pthread_*` calls. Those in turn do use futex but PI futex is a special type of futex you have to explicitly request. You can do it with these flags on the pthread level: https://linux.die.net/man/3/pthread_mutexattr_setprotocol

I don't see any way to get it from `std::mutex` though. Anyway, realtime is a whole can of worms and that solution is POSIX specific, maybe even Linux specific.

include/setup.h
```diff
@@ -557,9 +557,12 @@ class Module_base {
 
 bool config_file_is_valid(const std_fs::path& path);
 
-// Helper functions for retrieving common configuration sections
+// Helper functions for retrieving configuration sections
```
epifaniooo, line 560: auto-value-parcel

johnnovak: Wat?

src/hardware/mixer.cpp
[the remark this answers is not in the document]
johnnovak: Mutex restored.

src/hardware/mixer.cpp
johnnovak: Fixed in https://github.com/dosbox-staging/dosbox-staging/pull/4248

src/hardware/mixer.cpp
johnnovak: I've added the lock guard to all public MixerChannel methods, as discussed.

src/hardware/mixer.cpp
```diff
@@ -1319,38 +1326,56 @@ static int clamp_filter_cutoff_freq([[maybe_unused]] const std::string& channel_
...
-		f.setup(order, mixer.sample_rate_hz, cutoff_freq_hz);
-	}
+	assert(_cutoff_freq_hz > 0);
+
+	const auto cutoff_freq_hz = clamp_filter_cutoff_freq(name, _cutoff_freq_hz);
 
 	filters.lowpass.order          = order;
 	filters.lowpass.cutoff_freq_hz = cutoff_freq_hz;
+
+	InitLowPassFilter();
```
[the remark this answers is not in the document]
johnnovak: Added lock guard to all public MixerChannel methods as discussed.

johnnovak:
Thanks for the review @weirddan455, I've addressed your comment.s

weirddan455:
You still have that compile error in the first commit.

johnnovak:
> You still have that compile error in the first commit.

Hmm, I don't think I have it on Windows. Care to post the error you're getting on Linux?

weirddan455:
Same error I posted last time:

```
In file included from ../../src/audio/compressor.cpp:29:
../../include/mixer.h:44:10: fatal error: Iir.h: No such file or directory
   44 | #include <Iir.h>
```

You could try doing the CMake build in Windows to reproduce it instead of the VS solution files. I get the same thing in both CMake and Meson.

shermp:
It looks like a file case issue?

eg: `#include <Iir.h>` should be `#include <iir.h>`?

weirddan455:
No, the casing is correct. It's in `subprojects/iir1-1.9.3/Iir.h` in Meson and pulled in by vcpkg in CMake.

johnnovak:
> Same error I posted last time:

My bad, sorry. I think I know what it is.

johnnovak:
@weirddan455 I believe I managed to fix up the first commit.

include/mixer.h
```diff
@@ -177,20 +177,20 @@ class MixerChannel {
...
 
-	bool HasFeature(ChannelFeature feature) const;
-	std::set<ChannelFeature> GetFeatures() const;
-	const std::string& GetName() const;
-	int GetSampleRate() const;
-	float GetFramesPerTick() const;
-	float GetFramesPerBlock() const;
-	double GetMillisPerFrame() const;
+	bool HasFeature(ChannelFeature feature);
+	std::set<ChannelFeature> GetFeatures();
```
weirddan455, line 181: I thought you liked the use of `const`. I think `const` for member methods can be used by the compiler for optimization letting it know the object isn't mutable inside the function. I believe it translates the implicit `this` pointer into `const this`.

src/hardware/mixer.cpp
```diff
@@ -801,25 +806,31 @@ void MixerChannel::UpdateCombinedVolume()
 	combined_volume_gain = user_volume_gain * app_volume_gain * db0_volume_gain;
 }
 
-const AudioFrame MixerChannel::GetUserVolume() const
+const AudioFrame MixerChannel::GetUserVolume()
 {
+	std::lock_guard lock(mutex);
```
weirddan455, line 811: Unless taking the mutex lock requires non-const, which would make sense.

weirddan455 (approved):
Looks good. Ran `scripts/compile_commits.sh` and everything builds now.

src/hardware/mixer.cpp
johnnovak: As per above.

Issue #4238, opened 2025-03-15
Options menu not appearing in-game in Jane's AH-64D Longbow (vanilla, non-GOLD edition)


jordigo:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 11

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Relevant hardware info

AMD Ryzen 5600X, Radeon RX 6700XT

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

Game version is 1.07: While inside a mission, you should be able to open the options menu (that allows for graphics selection between VGA and SVGA as well as joystick selection and calibration, among others) but it won't appear. It will open in the main menu, though. Tried with other DOSBOX forks (ECE) and works correctly.

If updating to 1.08, the game fails to launch any mission reporting an "internal error" (the game itself, not DOSBOX).

### Steps to reproduce the behaviour

Explain in detail how to reproduce the bug. Assume no familiarity with the software.
1. Launch game
2. Go to "instant action" (control tower on top of screen)
3. Once game has loaded, press "ALT + O" to (unsuccessfully) open options menu

### Download URL of affected game or software

https://archive.org/details/longbow_202411

### Your configuration

```ini
common conf:
https://gist.github.com/jordigo/63cc92959945e839a64b798a01fd664a#file-gistfile1-txt

Specific game conf:
https://gist.github.com/jordigo/1d5733e3add06af82736a5815e1e2afa#file-gistfile1-txt
```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
@jordigo Does this work correctly in earlier Staging versions?

jordigo:
Just checked on 0.80.1 and 0.81.0 both work as intended, options menu opens without problem.

When changing to 0.82.0, same game installation used for testing with previous versions won't open menu when in game or refuse to open encountering "internal error" (if updated from 1.07 to 1.08).

johnnovak:
Thanks for the addtitional testing @jordigo, that's very good info 👍🏻 

Sounds like an input handling regression, doesn't it @FeralChild64 ?

Not sure about the internal error, though.

jordigo:
No problem, happy to test anything else you need.

[...]

PR #4245, opened 2025-03-21
VESA 2.0 mode list regression fix

johnnovak:
# Description

## Problem description

Starting from the VESA 2.0 standard, applications were required to query the list of available VESA modes instead of relying of card-specific mode numbers which was a common pre-VESA-2.0 practice. In fact, VESA 2.0 does not specify any fixed video modes, only _hints_ at it that manufacturers _may_ implement some common VESA 1.2 modes, but no guarantees whatsoever. Basically, everything is up to the graphics card/chipset implementors, and applications should not assume anything.

This is all good and well, but some questionably coded games still did not handle VESA 2.0 mode querying quite correctly, despite the standard (including the famous Quake). The coders made certain assumptions on the maximum number of entries the mode list can contain, and this can cause programs to misbehave or outright crash if the list is "too large" (e.g., by triggering buffer overflows, in which case pretty much anything can happen).

I added three extra VESA 2.0 text modes as part of my `MODE` command implementations for some custom 132-column modes, and while this was not a problem during my spot testing of a few VESA-enabled programs, it turns out  **Jane's Combat Simulations AH-64D Longbow** (non-Gold edition) gets confused by the presence of these modes (see the related [issue ticket](https://github.com/dosbox-staging/dosbox-staging/issues/4238) for details). Interestingly, the Gold version is not affected--presumably, the developers realised the error of their ways and fixed their VESA mode querying routine in the later Gold re-release...

## Solution

The solution is to not make these special 132-column VESA 2.0 text modes available when `vesa_modes = compatible` is set, which is our default. That's the safest thing to do because who knows what else might be broken due to this slightly longer mode list. People wishing to use these modes will need to set `vesa_modes = all`, but that can trigger the "mode list too large" problem with certain games.

I think this is the best compromise as these special text modes are only nice-to-haves and hardly essential. Being compatible with as many games as possible by default is our primary goal, so something had to give. Plus it's best to avoid introducing regressions when something non-essential like this is introduced.

## Alternatives considered

None of the below options were to my liking, I'm just listing them for completeness:

1. Simply document on the Wiki that the game runs fine with with `vmemsize` set to 2 MB (the default is 4 MB for the S3 Trio64). This reduces the number of available VESA modes, so the game bug is not triggered.

2. Convert these special 132-column VESA modes into non-VESA text modes. I tried that briefly, but ran into issues (I got 40-column modes instead), and I'm not interested in spending too much time investigating this. Probably VESA is needed for some reason to support 132-column modes--why exactly, I'm not sure.

3. Introduce various mode list culling options [like DOSBox X](https://dosbox-x.com/wiki/Guide%3AVideo-card-support-in-DOSBox%E2%80%90X#_vesa_modelist_cap).  For example, we could disallow resolutions larger than 1280x1024 or something if a special config option is set. But that doesn't help with out-of-the-box compatibility and I'm a bit averse to introducing new config settings for these workarounds. There is a large number of these VESA-culling config options in X, but I'd say let's not go there.

4. Remove these three DOSBox-special 132-column modes. Nah, I wanna keep them! 😎 

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4238

The `MODE` command and the new 132-column VESA modes were introduced in this PR of mine: https://github.com/dosbox-staging/dosbox-staging/pull/3741

# Release notes

- Fixed **Jane's Combat Simulations AH-64D Longbow** (original non-Gold version) regression where the in-game options menu could not be opened.

- The 132x28, 132x30, and 132x34 DOSBox-special VESA text modes are now only available if `vesa_modes = all` is set.

# Manual testing

- Confirmed the game now works with the stock settings.

- Confirmed setting the special 132-column modes with the `MODE` command works with `vesa_modes = all`, and in `compatible` mode an error is displayed.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/vga_s3.cpp
+++ b/src/hardware/vga_s3.cpp
@@ -768,7 +768,8 @@ void filter_compatible_s3_vesa_modes()
 		}
 
 		// Allow common standard VESA modes, except 320x200 hi-color
-		// modes that were rarely properly supported until the late 90s.
+		// modes that were rarely properly supported until the late 90s,
+		// and the DOSBox-specific widescreen modes.
 		constexpr auto s3_vesa_modes_start = 0x150;
 
 		if (m.mode < s3_vesa_modes_start) {
@@ -776,8 +777,20 @@ void filter_compatible_s3_vesa_modes()
 			constexpr auto _320x200_16bit = 0x10e;
 			constexpr auto _320x200_32bit = 0x10f;
 
-			return !(m.mode == _320x200_15bit || m.mode == _320x200_16bit ||
-			         m.mode == _320x200_32bit);
+			// Additional DOSBox-specific widescreen modes
+			constexpr auto _848x480_8bit  = 0x222;
+			constexpr auto _848x480_15bit = 0x223;
+			constexpr auto _848x480_16bit = 0x224;
+			constexpr auto _848x480_32bit = 0x225;
+
+			return !contains(std::vector({_320x200_15bit,
+			                              _320x200_16bit,
+			                              _320x200_32bit,
+			                              _848x480_8bit,
+			                              _848x480_15bit,
+			                              _848x480_16bit,
+			                              _848x480_32bit}),
+			                 m.mode);
 		}
 
 		// Selectively allow S3-specific VESA modes.
--- a/src/hardware/vga_s3.cpp
+++ b/src/hardware/vga_s3.cpp
@@ -751,9 +751,14 @@ void filter_compatible_s3_vesa_modes()
 	}
 
 	auto mode_allowed = [&](const VideoModeBlock& m) {
-		// Allow all text modes
+		// Only allow standard text modes
 		if (m.type == M_TEXT) {
-			return true;
+			constexpr auto _132x28 = 0x230;
+			constexpr auto _132x30 = 0x231;
+			constexpr auto _132x34 = 0x232;
+
+			return !contains(std::vector({_132x28, _132x30, _132x34}),
+			                 m.mode);
 		}
 
 		// Allow all non-VESA modes (standard VGA modes, and CGA and EGA
--- a/src/dos/program_mode.cpp
+++ b/src/dos/program_mode.cpp
@@ -166,10 +166,15 @@ void MODE::HandleSetDisplayMode(const std::string& _mode_str)
 
 				const auto& mode = it->second;
 
-				if (mode < MinVesaBiosModeNumber) {
-					INT10_SetVideoMode(mode);
+				if (VESA_IsVesaMode(mode)) {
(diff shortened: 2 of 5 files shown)


johnnovak:
@jordigo If interested, you can test the game with the CI build of this branch.

jordigo:
@johnnovak I don't have the required environment for compiling so I've downloaded the available ZIP ([dosbox-staging-windows-x64-0.83.0-alpha-19aca) from https://github.com/dosbox-staging/dosbox-staging/actions/runs/13988180088 and it works beautifully. Thank you very much!

src/hardware/vga_s3.cpp
```diff
@@ -751,9 +751,18 @@ void filter_compatible_s3_vesa_modes()
...
+			constexpr auto _132x28 = 0x230;
+			constexpr auto _132x30 = 0x231;
+			constexpr auto _132x34 = 0x232;
+
+			const std::vector disallowed_modes = {_132x28, _132x30, _132x34};
+
+			const auto is_allowed = (std::ranges::find(disallowed_modes,
+			                                           m.mode) ==
+			                         disallowed_modes.end());
+			return is_allowed;
```
shermp, line 765: We have a `contains()` convenience template function in `support.h` that could be used?

It could become as compact as something like:

```cpp
return !contains(std::array{_132x28, _132x30, _132x34}, m.mode);
```

src/hardware/vga_s3.cpp
```diff
@@ -772,16 +772,33 @@ void filter_compatible_s3_vesa_modes()
...
+			                                      _320x200_32bit,
+			                                      _848x480_8bit,
+			                                      _848x480_15bit,
+			                                      _848x480_16bit,
+			                                      _848x480_32bit};
+
+			const auto is_allowed = (std::ranges::find(disallowed_modes,
+			                                           m.mode) ==
+			                         disallowed_modes.end());
+			return is_allowed;
```
shermp, line 801: `contains()` from `support.h`?

src/hardware/vga_s3.cpp
johnnovak: Ah, cheers, forgot about that. Yeah I don't like it either how verbose this is.

johnnovak:
Thanks for the review @shermp. I've addressed your comments.

shermp (approved):
Haven't tested, but looks ok. `contains()` is definitely easier to read.

johnnovak:
@jordigo Confirming this also fixes the exit to DOS problem when opening the options screen in-game in 1.08F.

jordigo:
@johnnovak Indeed, sorry I forgot to mention I tested it with the 1.08F updated version of the game. Thanks for the effort and fix!

PR #4247, opened 2025-03-22
Jn/setting help shortcut

johnnovak:
# Description

We can already use shortcuts for getting/setting the values of config settings:

```
sbtype 
sbtype sb16
sbtype = sb16
```

But we still need to use the tedious `config -h sbtype` form to display a setting's help. This PR fixes that; now can simply just say `sbtype /?` which is quite natural.

# Release notes

- Now you can use the `SETTINGNAME /?` shortcut to get a config setting's help. For example, you can use `sbtype /?` instead of the tedious to type `config -h sbtype` command.

# Manual testing

image

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


src/shell/shell_cmds.cpp
```diff
@@ -132,11 +132,12 @@ bool lookup_shell_cmd(std::string name, SHELL_Cmd &shell_cmd)
 	return true;
 }
 
-bool DOS_Shell::ExecuteConfigChange(const char* const cmd_in, const char* const line)
+bool DOS_Shell::ExecuteConfigChange(const char* const cmd_in, const std::string args)
```
weirddan455: Minor but you could take in `std::string_view` to avoid a copy. It's the more direct C++ equivalent to `char *`.

src/shell/shell_cmds.cpp
```diff
@@ -132,35 +132,48 @@ bool lookup_shell_cmd(std::string name, SHELL_Cmd &shell_cmd)
...
 
-bool DOS_Shell::ExecuteConfigChange(const char* const cmd_in, const std::string args)
+bool DOS_Shell::ExecuteConfigChange(const char* const property_name, const std::string _args)
 {
 	assert(control);
 	const auto section_dosbox = static_cast<Section_prop*>(
 	        control->GetSection("dosbox"));
 
+	auto args = _args;
+	trim(args);
```
weirddan455, line 142: Or since you're taking `std::string` already you can avoid this 2nd copy and just mutate it in place.

src/shell/shell_cmds.cpp
```diff
@@ -132,35 +132,48 @@ bool lookup_shell_cmd(std::string name, SHELL_Cmd &shell_cmd)
...
-	char newcom[1024];
-	safe_sprintf(newcom, "z:\\config -set %s %s%s", test->GetName(), cmd_in,
-	             args.c_str());
-	DoCommand(newcom);
+	if (args == "/?") {
+		const auto command = format_str("z:\\config -h %s %s",
+		                                section->GetName(),
+		                                property_name);
+
+		DoCommand(const_cast<char*>(command.c_str())); //-V2018
```
weirddan455, line 167: This is UB since `DoCommand()` modifies the string. From a comment in my `basic_string.h` header: "This is a handle to internal data.  Do not modify or dire things may happen".

`.data()` apparently has an overload that allows mutation but I'm not sure how to use it: https://en.cppreference.com/w/cpp/string/basic_string/data

Another comment from the `.data()` method:

```cpp
      /**
       *  @brief  Return const pointer to contents.
       *
       *  This is a pointer to internal data.  It is undefined to modify
       *  the contents through the returned pointer. To get a pointer that
       *  allows modifying the contents use @c &str[0] instead,
       *  (or in C++17 the non-const @c str.data() overload).
      */
```

src/shell/shell_cmds.cpp
```diff
@@ -132,35 +132,48 @@ bool lookup_shell_cmd(std::string name, SHELL_Cmd &shell_cmd)
...
+		return true;
+	}
+
+	const auto command = format_str("z:\\config -set %s %s %s",
+	                                section->GetName(),
+	                                property_name,
+	                                args.c_str());
 
+	DoCommand(const_cast<char *>(command.c_str())); //-V2018
```
weirddan455, line 176: Same comment here

weirddan455 (changes requested):
Minor comments about un-needed copying. Not a performance sensitive part of the code so it's whatever.

The UB I mentioned is more concerning.

src/shell/shell_cmds.cpp
johnnovak: Yeah I'd rather not mutate input args. That should be only done as a special optimisation step for performance sensitive code.

src/shell/shell_cmds.cpp
johnnovak: Yeah I'll try that.

src/shell/shell_cmds.cpp
johnnovak: Yeah the whole shell command execution stuff is a horrific mess of low-level C style string manipulation with mutation everywhere 🥶

Problem is you almost need to refactor everything to use `std::string` in a normal way. I'll see what I can do with minimal further refactors. Probably `DoCommand` would need to take a string view or `std::string` and _not_ mutate it in place... but it's used in so many places; that would be a big refactoring job.

The whole thing is a fucking mess. A much better would have been tokenise the command line into a vector of tokens, pass that arpund, then perhaphs maintain an index where we are at with processing the list of tokens, etc. instead of this direct string pointer manipulation hackery.

I hate C string handling.

src/shell/shell_cmds.cpp
johnnovak: Actually, I can't because I'm calling `c_str()` on it later. The big showstopper with calling `data()` on `std::string_view` is that the terminating zero is not guaranteed (which kinda makes sense, but it's bad for C interop as you need to make a copy anyway to be 100% safe).

src/shell/shell_cmds.cpp
johnnovak: Reverted to use `safe_snprintf` and C strings. Converting to `std::string` would require a much deeper refactoring job so we don't just pass a `char *` around that everybody mutates... 🥶

src/shell/shell_cmds.cpp
johnnovak: As per above.

johnnovak:
> The UB I mentioned is more concerning.

Alright, I've addressed that, cheers.

MasterO2:
That's an immensely helpful feature, @johnnovak. Thanks for adding it!

PR #4248, opened 2025-03-22
Change display names of bool conf settings to 'on' and 'off'

johnnovak:
# Description

The primary audience for DOSBox is gamers, not programmers, so we should avoid using terms that only a programmer would understand.

This change completely eradicates the use of `true` and `false` for toggle config settings (that we like to call "booleans" that nobody understands who isn't a mathematician or a programmer 😄). Now the displayable names for those settings are `on` and `off`.

From now on, if you write a new config with the `config -wc blah.conf` command, the config file will contain `on` and `off` for the toggle setting types.

Note we have already been accepting `false`, `off`, `disabled`, `0` and `none` for negative values, and `true`, `on`, `enabled`, and `1` for positive values, so this is a display-only change.

# Release notes

Probably it's not worth mentioning.

# Manual testing

image

image

image

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -3383,7 +3383,7 @@ void init_cpu_dosbox_settings(Section_prop& secprop)
 	        "cannot keep up (%s by default).\n"
 	        "Only affects fixed cycles settings. When enabled, the number of cycles per\n"
 	        "millisecond can vary; this might cause issues in some DOS programs.",
-	        (CpuThrottleDefault ? "enabled" : "disabled")));
+	        (CpuThrottleDefault ? "'on'" : "'off'")));
 
 	auto pint = secprop.Add_int("cycleup", Always, DefaultCpuCycleUp);
 	pint->SetMinMax(CpuCycleStepMin, CpuCycleStepMax);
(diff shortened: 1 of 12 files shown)


weirddan455 (approved):
Looks good. My main gripe was the inconsistency between the help text and the generated option. I tested and it looks like everything correctly updated.

There still is a bit of inconsistency though. For example `waitonerror` help text says "enabled by default", `mute_when_inactive` says "disabled by default". There's several instances of those. Maybe change those to "on/off by default" while you're at it?

johnnovak:
> Looks good. My main gripe was the inconsistency between the help text and the generated option. I tested and it looks like everything correctly updated.
> 
> There still is a bit of inconsistency though. For example `waitonerror` help text says "enabled by default", `mute_when_inactive` says "disabled by default". There's several instances of those. Maybe change those to "on/off by default" while you're at it?

Yeah I could now say "('on' by default)" instead of "(enabled by default)". I guess that is clearer. Just previously it was inconsistent, i.e., the help talked about 'on' and 'off', but the written config contained 'true' and 'false', so I kind of "resolved" that conflict by talking about enabled or disabled defaults 🤷🏻.

johnnovak:
Confirmed that we have `('on' by default)` and `('off' by default)` everywhere when writing a new conf file.

PR #4250, opened 2025-03-23
Enable bilinear filtering for the 3dfx Voodoo by default

johnnovak:
# Description

My understanding is we only disabled bilinear filtering for the Voodoo because of performance reasons, but I can't see any significant CPU usage difference between bilinear on and off in my recent tests when using the stock settings. Probably @weirddan455's nice threading improvements brought the "bilinear on" performance on par with the "bilinear off" path! 🎉 

For context, bilinear filtering was a big deal back in the day. It was impossible to do such smoothing via purely CPU-rendered texture mapping, so blocky pixelated rendering was considered yesterday's news--everybody wanted the bilinear-filtered wax figurine look of the Voodoo in 3D games 😆 That was _the shit_ in the late 90s in DOS circles among 3D enthusiasts! 😎 

So yeah, I think we should emulate the Voodoo as accurately as possible by default.

Some sample screenies with bilinear and no bilinear:

lol2_1-filter

lol2_1-no-filter

lol2_2-filter

lol2_2-no-filter

# Release notes

- Bilinear filtering is now enabled by default on the 3dfx Voodoo to emulate the Voodoo look more accurately.

# Manual testing

I've also improved the config descriptions a little bit:

image

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -824,14 +824,20 @@ void DOSBOX_Init()
 	secprop = control->AddSection_prop("voodoo", &VOODOO_Init);
 
 	pbool = secprop->Add_bool("voodoo", when_idle, true);
-	pbool->Set_help("Enable 3dfx Voodoo emulation ('on' by default).");
+	pbool->Set_help(
+	        "Enable 3dfx Voodoo emulation ('on' by default). This is authentic low-level\n"
+	        "emulation of the Voodoo card without any OpenGL passthrough, so it requires a\n"
+	        "powerful CPU. Most games need the DOS Glide driver called 'GLIDE2X.OVL' to be\n"
+	        "in the path for 3dfx mode to work. Many games include their own Glide driver\n"
+	        "variants, but for some you need to provide a suitable 'GLIDE2X.OVL' version.\n"
+	        "A small number of games integrate the Glide driver into their code, so they\n"
+	        "don't need 'GLIDE2X.OVL'.");
 
 	pstring = secprop->Add_string("voodoo_memsize", only_at_start, "4");
 	pstring->Set_values({"4", "12"});
 	pstring->Set_help(
-	        "Set the amount of video memory for 3dfx Voodoo graphics, either 4 or 12 MB.\n"
-	        "The memory is used by the Frame Buffer Interface (FBI) and Texture Mapping Unit\n"
-	        "(TMU) as follows:\n"
+	        "Set the amount of video memory for 3dfx Voodoo graphics. The memory is used by\n"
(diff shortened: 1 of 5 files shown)


weirddan455 (approved):
This is fine by me. I tested it in Descent II and it definitely looks blurrier but that's also how I remember early 3D games looking (although I never had an early Voodoo).

As for performance, I can't tell a measurable difference between bilinear on and off. However, an x86 build is using SSE2 intrinsics to help out.

https://github.com/dosbox-staging/dosbox-staging/blob/main/src/hardware/voodoo.cpp#L221-L245

I did a find + replace to `#if 0` to disable SSE2 and then I noticed a slight drop from ~145 FPS to ~135 FPS when using bilinear filtering (bilinear turned off still maintained 145 FPS with this build).

So ARM CPUs might have a slight performance impact from this setting since they don't have SSE2 and there is no NEON path. You might want to leave the note about performance in the help text for Raspberry Pi users. ARM Mac users might notice a slight dropoff but I'm guessing Apple CPUs will generally be fast enough that it's not a deal-breaker.

johnnovak:
Thanks for pointing out the SSE2 optimisation @weirddan455. Yeah, I could only test it on Windows.

> So ARM CPUs might have a slight performance impact from this setting since they don't have SSE2 and there is no NEON path. You might want to leave the note about performance in the help text for Raspberry Pi users. 

That's a good idea, I'll put that note back.

> ARM Mac users might notice a slight dropoff but I'm guessing Apple CPUs will generally be fast enough that it's not a deal-breaker.

Yeah, exactly my thoughts; non-issue on Apple silicon Macs.

kklobe:
@johnnovak just pushed a branch that uses simde for SSE2 in the filtering to make it cross-platform, check it out if interested. https://github.com/dosbox-staging/dosbox-staging/tree/kk/voodoo-simde-sse2

johnnovak:
> @johnnovak just pushed a branch that uses simde for SSE2 in the filtering to make it cross-platform, check it out if interested. https://github.com/dosbox-staging/dosbox-staging/tree/kk/voodoo-simde-sse2

Nice one! I'll check it on Windows!

johnnovak:
@kklobe I've tested [kk/voodoo-simde-sse2](https://github.com/dosbox-staging/dosbox-staging/tree/kk/voodoo-simde-sse2) on Windows; it works fine and I'm getting the same performance, which is what we expected. I'd say raise a PR for it, the SSE2 compatibility mode might improve things on slower non-x86 platforms like the RPi.

johnnovak:
@weirddan455 Restored the performance related help text plus added an extra refactoring commit at the end so you might wanna take a second look.

weirddan455 (approved):
Looks good

Issue #4264, opened 2025-03-30
vcpkg provided FluidSynth sounds incorrect compared to Meson wrap


weirddan455:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Linux x86_64

### If other OS or Linux, please describe

Debian 12

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

It sounds like it's underwater or something. I also get the bad sound with my system (Debian provided) FluidSynth. Only the Meson wrap sounds correct.

Bad - System 2.3.1
Good - Meson wrap 2.3.3
Bad - vcpkg 2.4.2

[fluidsynth-recordings.zip](https://github.com/user-attachments/files/19526623/fluidsynth-recordings.zip)

Soundfont I'm using:

[SC-55.SoundFont.v1.2b.sf2.zip](https://github.com/user-attachments/files/19526626/SC-55.SoundFont.v1.2b.sf2.zip)

### Steps to reproduce the behaviour

Explain in detail how to reproduce the bug. Assume no familiarity with the software.
1. Use Sound font linked above with FluidSynth
2. Play first level of Doom using General MIDI
3. Test with meson using `meson setup --force-fallback-for=fluidsynth` - Sound correct
4. Test with CMake using vcpkg - Unlock hidden Doom underwater level

### Your configuration

```ini

```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

weirddan455:
I found the difference by doing some FluidSynth builds out of tree straight from upstream. In our `meson.build`, we're setting `enable-floats=true`. There's a CMake setting with the same name upstream. Description is "enable type float instead of double for DSP samples".

We probably want to set this in vcpkg if possible. At least this SC55 soundfont I'm using seems to depend on it.

johnnovak:
That's a FluidSynth bug. They had single precision float related bugs in their DSP code before, so looks like there's more:

https://github.com/FluidSynth/fluidsynth/issues/1331

Can you report that on the FluidSynth issue tracker as well? I will look into setting 32-bit float mode at build time in the meantime (I think it helps with performance on low-end hardware somewhat).

weirddan455:
[...]

Issue #4260, opened 2025-03-27
Failure to build due to git show issue with commit `117512d8` (Windows - Visual Studio)


Zvezda-Zake-Kuerten:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### What is your question and how can we help?

I've been trying to build Dosbox Staging for a couple of hours now, and always end up with the same error:
while checking out baseline from commit '117512d870c0f260f4914b6b4bb006b9054b06eb', failed to `git show` versions/baseline.json. This may be fixed by fetching commits with `git fetch`.

This is happening with dosbox, tets and zmbv.

It is my first time building Dosbox Staging, and I believe I'm doing something wrong, and would really appreciate your help in figuring out the issue.

I've cloned the git through Visual Studio itself;
I can confirm that all dependencies are installed;
I followed the info in the readme;
I'm using Release x64 to try and build it with Visual Studio;

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

weirddan455:
I believe that's a vcpkg error. `117512d870c0f260f4914b6b4bb006b9054b06eb` is the baseline from `vcpkg.json`.

Make sure you follow the instructions here. Particularly step 2 concerning vcpkg: https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/build-windows.md

Zvezda-Zake-Kuerten:
> I believe that's a vcpkg error. `117512d870c0f260f4914b6b4bb006b9054b06eb` is the baseline from `vcpkg.json`.
> 
> Make sure you follow the instructions here. Particularly step 2 concerning vcpkg: https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/build-windows.md

Awesome! I'll triple check that I'm doing everything right on this one.
I do recall getting some issues with installing vcpkg a while back. I'll see if a clean install solves the issue.

Thank you so much!

johnnovak:
@Zvezda-Zake-Kuerten Have you succeeded? Can we close this?

weirddan455:
I just ran into this error myself while trying to build #4270

[...]

PR #4270, opened 2025-04-04
Bump vcpkg baseline

kklobe:
# Description

libffi was not building with CMake 4.0 on macOS, so I bumped the vcpkg baseline to the latest. Might as well get everything updated.

# Manual testing

Ran the usual test cases:

- Sunflower
- QTD
- Blood (Voodoo mode)
- moralhardcandy
- Unreal (Future Crew)

- [ ] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/scripts/vcpkg-baseline-diff.sh
+++ b/scripts/vcpkg-baseline-diff.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+if [ "$#" -ne 2 ]; then
+  echo "Usage: $0 <old-baseline> <new-baseline>"
+  exit 1
+fi
+old="$1"
+new="$2"
+patterns=$(jq -r '.dependencies[] | if type=="string" then . else .name end' vcpkg.json | sed 's/.*/^[[:space:]]*-[[:space:]]*&([[:space:]]|$)/')
+vcpkg portsdiff "$old" "$new" | grep -Ef <(echo "$patterns")
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -24,5 +24,5 @@
       "host": true
     }
   ],
-  "builtin-baseline": "117512d870c0f260f4914b6b4bb006b9054b06eb"
+  "builtin-baseline": "9b75e789ece3f942159b8500584e35aafe3979ff"
 }


weirddan455 (approved):
Yeah, I had noticed a warning from libffi on Linux and was considering doing a bump. It was saying `FindPackage(libffi)` is deprecated or something along those lines.

I'm sure you're aware but just for reference, this will bump all of our dependencies to the latest versions. It might be good to document which of our direct dependencies are getting upstream version updates in a commit message but I'm not sure if there's an easy way to do that with vcpkg or not.

kklobe:
> I'm sure you're aware but just for reference, this will bump all of our dependencies to the latest versions. It might be good to document which of our direct dependencies are getting upstream version updates in a commit message but I'm not sure if there's an easy way to do that with vcpkg or not.

Yeah, that's why I mentioned "Might as well get everything updated". I'm poking around at a script that could print the version changes based on the baselines.

weirddan455:
> I'm poking around at a script that could print the version changes based on the baselines.

Awesome. If it turns out to be tricky, we can go ahead and merge this as-is. Just thought it might help us keep track of our dependencies better in case we run into regressions.

johnnovak:
> > I'm poking around at a script that could print the version changes based on the baselines.
> 
> Awesome. If it turns out to be tricky, we can go ahead and merge this as-is. Just thought it might help us keep track of our dependencies better in case we run into regressions.

That's a very good idea, @weirddan455 👍🏻 

The package manager of NodeJS, `npm`, has a feature to write a `package-lock.json` file that pins all deps to a fixed version. You can generate this version lock file for already installed deps in a project with `npm i --package-lock-only`.

I was wondering if vcpkg had a similar feature to generate a new `vcpkg.json` manifest file based on the current one but with all package versions pinned, but I could not find it. So looks like it doesn't.

Apparently, you can only generate the whole dependency tree for all packages in manifest mode with the below workaround. Maybe this helps, @kklobe 

https://github.com/microsoft/vcpkg/issues/36520#issuecomment-1922680837

weirddan455:
Tested that this builds on Linux. Also did a quick test of Doom using FluidSynth. That still works and is parked at 2.3.5. Tyrian also works.

I did have to manually go into my vcpkg root folder and do a `git pull` or it throws some pretty cryptic errors like someone else ran into in #4260

Also I'm still getting the warning from libffi. I wouldn't worry too much about that though. From the callstack, it's a dependency of Glib (and like everything involving Glib it's janky). We're gonna yank Glib out of our builds soon enough.

```
CMake Warning at build/release/vcpkg_installed/x64-linux/share/libffi/libffiConfig.cmake:2 (message):
  find_package(libffi) is deprecated.

  libffi can be imported via CMake FindPkgConfig module:

  

      find_package(PkgConfig)
      pkg_check_modules(LIBFFI REQUIRED IMPORTED_TARGET libffi)
      target_link_libraries(main PRIVATE PkgConfig::LIBFFI)

  

  vcpkg provides proprietary CMake targets:

  

      find_package(unofficial-libffi CONFIG REQUIRED)
      target_link_libraries(main PRIVATE unofficial::libffi::libffi)

Call Stack (most recent call first):
  /home/daniel/vcpkg/scripts/buildsystems/vcpkg.cmake:893 (_find_package)
  build/release/vcpkg_installed/x64-linux/share/fluidsynth/FindGLib2.cmake:214 (find_package)
  /home/daniel/vcpkg/scripts/buildsystems/vcpkg.cmake:893 (_find_package)
  /usr/share/cmake-3.25/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  build/release/vcpkg_installed/x64-linux/share/fluidsynth/FluidSynthConfig.cmake:75 (find_dependency)
  /home/daniel/vcpkg/scripts/buildsystems/vcpkg.cmake:893 (_find_package)
  src/midi/CMakeLists.txt:16 (find_package)
```

kklobe:
> Apparently, you can only generate the whole dependency tree for all packages in manifest mode with the below workaround. Maybe this helps, @kklobe

We could manually pin versions like we did with SDL2 previously (https://github.com/dosbox-staging/dosbox-staging/commit/1c87432b53da676db304fb3a48b152776deb0304#diff-dbbb1b147126744a5eee012901d2b9a29cc478be03677f67825b9b2794f5d283R25), but let's not defeat the intent of the baseline commit strategy until we absolutely have to.

kklobe:
> Awesome. If it turns out to be tricky, we can go ahead and merge this as-is. Just thought it might help us keep track of our dependencies better in case we run into regressions.

Ok, I've added the version upgrades in the commit message, generated with a hacky little script.

johnnovak:
> > Apparently, you can only generate the whole dependency tree for all packages in manifest mode with the below workaround. Maybe this helps, @kklobe
> 
> We could manually pin versions like we did with SDL2 previously ([1c87432#diff-dbbb1b147126744a5eee012901d2b9a29cc478be03677f67825b9b2794f5d283R25](https://github.com/dosbox-staging/dosbox-staging/commit/1c87432b53da676db304fb3a48b152776deb0304#diff-dbbb1b147126744a5eee012901d2b9a29cc478be03677f67825b9b2794f5d283R25)), but let's not defeat the intent of the baseline commit strategy until we absolutely have to.

Yeah, I prefer the explicit locking of package versions in NodeJS/npm via lock files (didn't imagine I would cite NodeJS as an example for anything 😆), that's the 100% solution. Then I upgrade what I want, when I want.

But I agree, if vcpkg pushes people gently towards the "upgrade whole baseline" approach, and I don't think we should go out of our way to work against that.

Putting the currently versions into the commit message is a nice in-between solution I think. I think we should also put the current vcpkg version in the README and update it there, so people can just see the current status of the deps at a glance. Don't worry about that part, I'll do that.

> Ok, I've added the version upgrades in the commit message, generated with a hacky little script.

Can you please include that script in your PR? Just put it into `scripts`. Then I'll put some words around the baseline upgrade process into the README, and also create a table with the current dep versions. Updating that will be part of the upgrade process. Cheers!

kklobe:
> Yeah, I prefer the explicit locking of package versions in NodeJS/npm via lock files (didn't imagine I would cite NodeJS as an example for anything 😆), that's the 100% solution. Then I upgrade what I want, when I want.

Knock yourself out! :grin: (https://learn.microsoft.com/en-us/vcpkg/consume/lock-package-versions?tabs=inspect-bash#2---add-version-constraints-using-a-baseline)

> Can you please include that script in this PR as well? Just put it into `scripts/`. Then I'll put some words around the baseline upgrade process in the README, and also create a table with the current dep versions. Updating the table will be part of the vcpkg baseline/version upgrade process.

Done. Linux friends, please test!

johnnovak:
> Knock yourself out! 😁 (https://learn.microsoft.com/en-us/vcpkg/consume/lock-package-versions?tabs=inspect-bash#2---add-version-constraints-using-a-baseline)

Yeah, because bumping the baseline is explicit, on a second thought it fine to bump that, test, see what breaks, and only pin something to an earlier version if needed.

What I really dislike is "always get the latest" on every build, that's a horrible model. You get random dependencies, and you're forced to deal with complications resulting from a forced upgrade in the most inconvenient moments 😅 (talking from actual experiences at work when trying to get our some important production patch...)

I'm building now this on Windows then I'll report back with my test results!

weirddan455:
> Done. Linux friends, please test!

The script works for me and shows exactly what I was hoping for. Nice job. I tested the build with the new baseline the other day as well so all good there as far as I can tell.

Only comment I have is that FluidSynth isn't actually getting updated since we're currently using an overlay to pin it to 2.3.5. Probably not worth automating and it may still be worth documenting for baseline bumps. Just wanted to point it out to avoid confusion.

kklobe:
> Only comment I have is that FluidSynth isn't actually getting updated since we're currently using an overlay to pin it to 2.3.5. Probably not worth automating and it may still be worth documenting for baseline bumps. Just wanted to point it out to avoid confusion.

What is the benefit of using an overlay instead of pinning it with `"overrides":`  in the main `vcpkg.json`? (https://learn.microsoft.com/en-us/vcpkg/consume/lock-package-versions?tabs=inspect-bash#5---force-a-specific-version)

weirddan455:
> What is the benefit of using an overlay instead of pinning it with "overrides": in the main vcpkg.json?

The overlay lets you change compile flags, which I'm doing with FluidSynth.

> Or will this all go away once we start including our dynamic binaries?

Yes, or at least it will get offloaded to the dosbox-staging-ext repo. We're tracking the issue in #4264

johnnovak:
Tested Discworld and Azrael's Tear on Windows with SoundCanvas and FluidSynth; looking good @kklobe 👍🏻

johnnovak (approved):

Issue #4269, opened 2025-04-04
Astral Blur^TBL stuck at first scene with GUS


BackSpace54:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 11

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

Hi,
[Astral Blur](https://www.pouet.net/prod.php?which=1) from TBL is stuck at first scene if GUS soundcard is set on.
With GUS = false the demo runs from start to end but no sound.
Could you check this issue?
Thanks.

### Steps to reproduce the behaviour

1. Get the demo, unzip it (conf a.)
2. Use a clean install and empty conf file of DosBox Staging 0.82.1 (conf a.)
3. Set GUS = true (if not) and set file_locking = false (conf a.)
4. Mount the directory where demo was unzipped (conf a.)
5. run astral.exe (conf a.)
6. demo show only the first scene (conf a.)
7. restart from step 2. but set GUS = false (conf b.)
8. demo runs from start to end without sound (conf b.)

### Download URL of affected game or software

https://files.scene.org/get/demos/groups/tbl/pc/astral.zip

### Your configuration

```ini
a. Empty conf file + GUS = true and file_locking = false
b. Empty conf file + GUS = false and file_locking = false
```

### Provide a log

```text
Nothing relevant
log conf a. GUS: Activated 16 voices at 38587 Hz
log conf b. SB16: Resetting DSP
```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

weirddan455:
Works on 0.81.2 so it's a regression since then. Commit 7d806a86d6f346418e57805608e2fcd885dcbbe8 also only shows the first scene. Looks like I didn't fully test this demo. I only made sure it got passed the file system init error (whoops).

This bug looks to be a separate regression from the file locking stuff. Probably in some of the GUS or mixer changes we made. I'll do a bisect in a bit.

weirddan455:
First bad commit is bdba6032523b363a4bd207f5229e860b783a1e09 from @interloper98 

FYI I had to `git cherry-pick 57fed293aa41a86553e9297a14a226f9f22ce4bf` as I was bisecting to get passed the file system init error but that commit cleanly merges here.

[...]

PR #4271, opened 2025-04-04
Fix GUS Mix Control Register not getting reset to the correct value

weirddan455:
# Description
Regression from bdba6032523b363a4bd207f5229e860b783a1e09

After the above refactor, the register was getting reset to 0 which is not its default value.
Added a constant and make sure we reset to that.

## Related issues

Fixes Astral Blur demo https://github.com/dosbox-staging/dosbox-staging/issues/4269

# Release notes

Fix GUS regression that caused Astral Blur demo to infinitely loop the first scene

# Manual testing

Astral Blur demo works now.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/gus.h
+++ b/src/hardware/gus.h
@@ -217,11 +217,13 @@ union ResetRegister {
 	bit_view<2, 1> are_irqs_enabled;
 };
 
+// Default state: lines disabled, latches enabled
+constexpr uint8_t MixControlRegisterDefaultState = 0b0000'1011;
+
 // Mix Control Register (2X0), section 2.13, page 28, of the UltraSound Software
 // Development Kit (SDK) Version 2.22
 union MixControlRegister {
-	// Default state: lines disabled, latches enabled
-	uint8_t data = 0b0000'1011;
+	uint8_t data = MixControlRegisterDefaultState;
 
 	bit_view<0, 1> line_in_disabled;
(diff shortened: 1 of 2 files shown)


src/hardware/gus.h
```diff
@@ -217,11 +217,13 @@ union ResetRegister {
 	bit_view<2, 1> are_irqs_enabled;
 };
 
+// Default state: lines disabled, latches enabled
+constexpr uint8_t MixControlRegisterDefaultState = 0b0000'1011;
```
johnnovak, line 221: Looks good. It's a pity we can't initialise this via setting the named fields in the bit_view. I guess as a workaround we could just turn this into a mutable member var the initialise in the constructor. Not quite sure that's better, so probably this is fine.

johnnovak (approved):
Nice quick ninja bisecting & fix @weirddan455 🥷🥇 

Btw, I'm not tagging these fixes with backport now because our next release should be 0.83.0, ideally before September (I'm going away for a while then). It would be good to get NukedSC55 out there, so I'm not even thinking anything major for 0.83.0 apart from that, just finish in-flight work.

interloper98 (approved):
Great find @weirddan455! Thanks for fixing this mistake on my part 👍

PR #4274, opened 2025-04-09
Fix emulator becoming unresponsive if the user exits the app while inside the reboot handler

### LINKED ISSUE #4232: Hangs when attempting to close Staging while rebooting

_No response_

Windows 11

_No response_

### Relevant hardware info

_No response_

Some games lead to a reboot process ("Rebooting in 3.. 2.. 1.. " counter).
If the DOSBox window is closed during that time, the app simply hangs indefinitely and stops responding to any input.

1. See attached zip file (Uridium)
2. Run the game using PLAY.BAT
3. After the Mindscape logo, you'll see the "Rebooting..." message
4. Close the Staging window using the mouse
5. Staging will be stuck

This has been observed in different games as well, so it seems to be consistent during the aformentioned "rebooting" process.

_No response_

### Your configuration

```ini

```

### Provide a log

```text

```

- [x] Yes, I agree.

weirddan455:
# Description

It was getting stuck inside an idle loop that's meant to delay a three second countdown message. We just need to bail out early if shutdown is requested. We've had to add similar logic in other places (grep for "shutdown_requested").

## Related issues

Fixes #4232

# Release notes

Fix emulator becoming unresponsive if the user exits the app while inside the reboot handler

# Manual testing

Tested reboot functionality with the URIDIUM.ZIP from #4232

The game doesn't actually run like this though. It looks like it's a booter disk and the game wants to reboot the computer if it's inside DOS so that it can boot from floppy. You'd have to get a .img of the floppy and do a `boot URIDIUM.img` to play this game with DOSBox.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -109,7 +109,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 84
+          MAX_BUGS: 137
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/src/ints/bios.cpp
+++ b/src/ints/bios.cpp
@@ -1072,6 +1072,11 @@ static Bitu reboot_handler()
 		const auto start = PIC_FullIndex();
 		while ((PIC_FullIndex() - start) < delay_ms) {
 			CALLBACK_Idle();
+			// Bail out if the user closes the window.
+			// Otherwise we get stuck in an infinite loop.
+			if (shutdown_requested) {
+				return CBRET_NONE;
+			}
 		}
 	}
 


weirddan455:
PVS updated since the last time it ran on main. The new version is reporting 50+ more warnings than the old version. None of them are coming from this PR. Most of the new warnings are in `sdl_mapper.cpp`.

johnnovak:
> PVS updated since the last time it ran on main. The new version is reporting 50+ more warnings than the old version. None of them are coming from this PR. Most of the new warnings are in `sdl_mapper.cpp`.

Short term, you can just increase the warning count here if these are indeed new warnings:
https://github.com/dosbox-staging/dosbox-staging/blob/main/.github/workflows/pvs-studio.yml#L112

But if these have new distinct identifiers, probably better to add them to the list of disabled warnings:
https://github.com/dosbox-staging/dosbox-staging/blob/main/.github/workflows/pvs-studio.yml#L74

Either solutions would be fine for this PR, I'm like 50/50. Maybe increasing the warning count is slightly better if we care about this new class of errors/warnings—if not, disabling them is the way to go.

Note we also have `.pvs-suppress` here that is passed in as a suppress-file to the PVS analyser in the CI workflow, but not too sure how that was generated:
https://github.com/dosbox-staging/dosbox-staging/blob/main/.pvs-suppress

PVS has a base-lining feature, but I think it's better to do the bare minimum here to keep the pipeline green, then we can re-think our approach to PVS separately when we're on CMake in all our CI builds. That should come soon.

johnnovak (approved):
Change looks good, thanks for fixing this. Could you take care of the PVS warnings too, please?

weirddan455:
I'm just going to raise the max warning count. The warning is for a possible memory leak which seems useful in our sane code paths. The mapper is doing a very suspicious pattern like this in `AddKeyButtonEvent`

```cpp
CKeyEvent * event=new CKeyEvent(buf,key);
new CEventButton(x,y,dx,dy,title,event);
return event;
```

It's warning about the caller not using the return value. Three levels of inheritance into the constructor, it's storing the `this` pointer in a data structure so it's probably not truly leaked but we should avoid this pattern in new code :smile:

johnnovak:
> Three levels of inheritance into the constructor, it's storing the `this` pointer in a data structure so it's probably not truly leaked but we should avoid this pattern in new code 😄

Yeah, it seems it builds a tree structure, but the style of implementation is rather questionable 😆 

Thanks for fixing the warning count, good to go from my side.

johnnovak:
Adding the backport label because we'll release 0.82.2 to fix the macOS min OS version regression, after all.

Issue #4207, opened 2025-02-21
Add 2-button mouse support for the built-in mouse driver


johnnovak:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Is your feature request related to a problem? Please describe.

Currently, the built-in emulated mouse driver can only emulate a 3-button mouse.

As it turns out, some games that do not support 3-button mice misbehave when receiving middle mouse button events (e.g., **Eye of the Beholder** does nothing on a middle-click, but treats middle-click + drag as a left click). Some other games treat the middle button as a left click (e.g., **Gabriel Knight**, **Space Quest III**).

Both behaviours can be pretty annoying, so we should add an 2-button mouse emulation option to the built-in mouse driver.

3-button mice only became mainstream after Microsoft put wheel mice on the map with the release of their [IntelliMouse](https://en.wikipedia.org/wiki/IntelliMouse) product in 1996.

Behold the original IntelliMouse:

![Image](https://github.com/user-attachments/assets/bae9479a-2249-4492-99a7-1a588fbccce5)

This was the first widely-adopted three-button scroll mouse in all computing history. Prior to that, mice with more than two buttons were pretty much only used by professionals on the PC (e.g., graphics designers), but they were rather uncommon. They were a lot more common on very expensive UNIX workstations where the middle button usually acted as the paste text function, but no in home/personal computing circles.

From this follows 3-button mouse with scroll wheels only became popular at the _very_ tail end of the "true DOS gaming era" (which ended in 1997, or in 1998 with a lot of goodwill). It's pretty important that we offer at least an option to emulate the mouse common mouse used for DOS gaming, which is the 2-button mouse.  I personally used a 2-button mouse until 1999, which was the year when I switched from DOS to Windows 9x. Only then did I get a 3-button mouse with a scroll wheel.

Quick 2-minute montage from eBay images when searching for "ps2 mouse":

![Image](https://github.com/user-attachments/assets/fafcba28-34a6-45b5-8ad9-9ff63bf59c0a)

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
@FeralChild64 Assigned this to you as I assume you want to do this, and the rest of us aren't too enthusiastic about touching the input handling code anyway 😏

MrMadguy64:
[...]

PR #4284, opened 2025-04-17
Improve DOS builtin mouse driver configurability

FeralChild64:
# Description

Fixed/improved builtin DOS mouse driver configurability.

Rationale behind making serial port options case-insensitive:

1. Help in case of `SERIAL.COM` syntax error:

2. `SERIAL.COM` command help:

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4207
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/3891

# Release notes

Naming scheme of the built-in DOS mouse driver options got changed to be less confusing:
- `dos_mouse_driver` has been renamed to `builtin_dos_mouse_driver`
- `dos_mouse_immediate` became an `immediate` parameter of `builtin_dos_mouse_driver_options`
It is now possible to select the mouse model simulated by the built-in mouse driver - can be useful for games which misinterpret the middle mouse button presses, like _Eye of the Beholder_.
It is now possible to fine-tune mouse driver behavior, for the _Descent II_ official 3dfx patch.
Serial port config options are now case-insensitive; examples from `SERIAL.COM` help now work correctly.
Fixed `SERIAL.COM` help description of serial mouse config options.

# Manual testing

1. Check that _Descent II Vertigo_ with the official 3dfx patch can now be played with a mouse if `builtin_dos_mouse_driver_options = modern` is set (see https://github.com/dosbox-staging/dosbox-staging/issues/3891)
2. Check that `GTEST.COM` utility from the _Genius DOS mouse driver_ 9.20 package shows 2-button mouse if `builtin_dos_mouse_driver_model = 2button` (it should show 3 button mouse otherwise).
3. Check that if `builtin_dos_mouse_driver_model = 2button`, the _Eye of the Beholder_ game does not react to middle mouse button (see https://github.com/dosbox-staging/dosbox-staging/issues/4207).
4. Using _MPXPlay v1.66_ check that mouse wheel is supported (can be used to scroll directory tree) if `builtin_dos_mouse_driver_model = wheel` (`2button` or `3button` options should result in mouse wheel being ignored)
5. Execute command `SERIAL.COM 1 MOUSE MODEL:MSM` (try different case variants) - _CuteMouse v2.1_ driver can't be started with `ctmouse /S1` command (complains about hardware not found), but can be started with `ctmouse /S1 /M` (detects Mouse Systems mouse)
6. Execute command `SERIAL.COM 1 MOUSE MODEL:MSM` (try different case variants) - _CuteMouse v2.1_ driver can be started with _ctmouse /S1_ command, detects a wheel mouse

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/input/mouse_config.cpp
+++ b/src/hardware/input/mouse_config.cpp
@@ -84,6 +84,8 @@ static const std::vector<uint16_t> list_rates = {
         // issues.
 };
 
+constexpr auto DefaultBuiltinDosMouseDriverOptions = "";
+
 const std::vector<uint16_t>& MouseConfig::GetValidMinRateList()
 {
 	return list_rates;
@@ -179,6 +181,9 @@ static void set_serial_mouse_model(const std::string_view model_str)
 
 static void set_dos_driver_options(const std::string_view options_str)
 {
+	const std::string OptionImmediate = "immediate";
+	const std::string OptionModern    = "modern";
+
 	auto& last_options_str = mouse_config.dos_driver_last_options_str;
 	if (last_options_str == options_str) {
 		return;
@@ -188,39 +193,49 @@ static void set_dos_driver_options(const std::string_view options_str)
 	mouse_config.dos_driver_immediate = false;
 	mouse_config.dos_driver_modern    = false;
 
-	bool config_needs_sync = false;
-
 	for (const auto& token : split(options_str, " ,")) {
-		if (token == "immediate") {
+		if (token == OptionImmediate) {
 			mouse_config.dos_driver_immediate = true;
-		} else if (token == "modern") {
+		} else if (token == OptionModern) {
 			mouse_config.dos_driver_modern = true;
 		} else {
 			LOG_WARNING(
 			        "MOUSE: Invalid 'builtin_dos_mouse_driver_options' "
-			        "parameter: '%s', ignoring",
+			        "parameter: '%s', using defaults",
 			        token.c_str());
-			config_needs_sync = true;
+
+			set_section_property_value("mouse",
+			                           "builtin_dos_mouse_driver_options",
+			                           DefaultBuiltinDosMouseDriverOptions);
+
+			set_dos_driver_options(DefaultBuiltinDosMouseDriverOptions);
+			return;
 		}
 	}
 
-	if (config_needs_sync) {
-		// Config syntax was wrong, create valid setting
-		last_options_str = {};
-		if (mouse_config.dos_driver_immediate) {
-			last_options_str = "immediate";
-		}
-		if (mouse_config.dos_driver_modern) {
-			if (!last_options_str.empty()) {
-				last_options_str += " ";
-			}
-			last_options_str += "modern";
-		}
-
-		set_section_property_value("mouse",
-		                           "builtin_dos_mouse_driver_options",
-		                           last_options_str);
+	// Log new config options
+	static std::string last_logged_str = "";
+	if (options_str == last_logged_str) {
+		return;
 	}
+
+	std::string log_options = {};
+	if (mouse_config.dos_driver_immediate) {
+		log_options = OptionImmediate;
+	}
+	if (mouse_config.dos_driver_modern) {
+		if (!log_options.empty()) {
+			log_options += ", ";
+		}
+		log_options += OptionModern;
+	}
+
+	if (log_options.empty()) {
+		log_options = "none";
+	}
+
+	LOG_INFO("MOUSE (DOS): Driver options: %s", log_options.c_str());
+	last_logged_str = options_str;
 }
 
 static void set_sensitivity(const std::string_view sensitivity_str)
@@ -493,7 +508,9 @@ static void config_init(Section_prop& secprop)
 	        "            No DOS game uses the mouse wheel, only a handful of DOS applications\n"
 	        "            and Windows 3.x with special third-party drivers.");
 
-	prop_str = secprop.Add_string("builtin_dos_mouse_driver_options", always, "");
+	prop_str = secprop.Add_string("builtin_dos_mouse_driver_options",
+	                              always,
+	                              DefaultBuiltinDosMouseDriverOptions);
 	assert(prop_str);
 	prop_str->Set_help(
 	        "Additional built-in DOS mouse driver settings as a list of space or comma\n"
--- a/src/hardware/input/mouse_config.cpp
+++ b/src/hardware/input/mouse_config.cpp
@@ -94,25 +94,25 @@ const std::vector<uint16_t>& MouseConfig::GetValidMinRateList()
 bool MouseConfig::ParseComModel(const std::string_view model_str,
                                 MouseModelCOM& model, bool& auto_msm)
 {
-	if (model_str == ModelCom2Button) {
+	if (iequals(model_str, ModelCom2Button)) {
 		model    = MouseModelCOM::Microsoft;
 		auto_msm = false;
-	} else if (model_str == ModelCom3Button) {
+	} else if (iequals(model_str, ModelCom3Button)) {
 		model    = MouseModelCOM::Logitech;
 		auto_msm = false;
-	} else if (model_str == ModelComWheel) {
+	} else if (iequals(model_str, ModelComWheel)) {
 		model    = MouseModelCOM::Wheel;
 		auto_msm = false;
-	} else if (model_str == ModelComMsm) {
+	} else if (iequals(model_str, ModelComMsm)) {
 		model    = MouseModelCOM::MouseSystems;
 		auto_msm = false;
-	} else if (model_str == ModelCom2ButtonMsm) {
+	} else if (iequals(model_str, ModelCom2ButtonMsm)) {
 		model    = MouseModelCOM::Microsoft;
 		auto_msm = true;
-	} else if (model_str == ModelCom3ButtonMsm) {
+	} else if (iequals(model_str, ModelCom3ButtonMsm)) {
 		model    = MouseModelCOM::Logitech;
 		auto_msm = true;
-	} else if (model_str == ModelComWheelMsm) {
+	} else if (iequals(model_str, ModelComWheelMsm)) {
 		model    = MouseModelCOM::Wheel;
 		auto_msm = true;
 	} else {
@@ -124,13 +124,13 @@ bool MouseConfig::ParseComModel(const std::string_view model_str,
 
 static void set_capture_type(const std::string_view capture_str)
 {
-	if (capture_str == CaptureTypeSeamless) {
+	if (iequals(capture_str, CaptureTypeSeamless)) {
 		mouse_config.capture = MouseCapture::Seamless;
-	} else if (capture_str == CaptureTypeOnClick) {
+	} else if (iequals(capture_str, CaptureTypeOnClick)) {
 		mouse_config.capture = MouseCapture::OnClick;
-	} else if (capture_str == CaptureTypeOnStart) {
+	} else if (iequals(capture_str, CaptureTypeOnStart)) {
 		mouse_config.capture = MouseCapture::OnStart;
-	} else if (capture_str == CaptureTypeNoMouse) {
+	} else if (iequals(capture_str, CaptureTypeNoMouse)) {
 		mouse_config.capture = MouseCapture::NoMouse;
 	} else {
 		assertm(false, "Invalid mouse capture value");
@@ -141,11 +141,11 @@ static void set_dos_driver_model(const std::string_view model_str)
 {
 	auto new_model = mouse_config.model_dos;
 
-	if (model_str == ModelDos2Button) {
+	if (iequals(model_str, ModelDos2Button)) {
 		new_model = MouseModelDos::TwoButton;
-	} else if (model_str == ModelDos3Button) {
+	} else if (iequals(model_str, ModelDos3Button)) {
 		new_model = MouseModelDos::ThreeButton;
-	} else if (model_str == ModelDosWheel) {
+	} else if (iequals(model_str, ModelDosWheel)) {
 		new_model = MouseModelDos::Wheel;
 	} else {
 		assertm(false, "Invalid DOS driver mouse model value");
@@ -159,13 +159,13 @@ static void set_dos_driver_model(const std::string_view model_str)
 
 static void set_ps2_mouse_model(const std::string_view model_str)
 {
-	if (model_str == ModelPs2Standard) {
+	if (iequals(model_str, ModelPs2Standard)) {
 		mouse_config.model_ps2 = MouseModelPS2::Standard;
-	} else if (model_str == ModelPs2Intellimouse) {
+	} else if (iequals(model_str, ModelPs2Intellimouse)) {
 		mouse_config.model_ps2 = MouseModelPS2::IntelliMouse;
-	} else if (model_str == ModelPs2Explorer) {
+	} else if (iequals(model_str, ModelPs2Explorer)) {
 		mouse_config.model_ps2 = MouseModelPS2::Explorer;
-	} else if (model_str == ModelPs2NoMouse) {
+	} else if (iequals(model_str, ModelPs2NoMouse)) {
 		mouse_config.model_ps2 = MouseModelPS2::NoMouse;
 	} else {
 		assertm(false, "Invalid PS/2 mouse model value");
--- a/src/hardware/input/mouse_config.cpp
+++ b/src/hardware/input/mouse_config.cpp
@@ -94,26 +94,28 @@ const std::vector<uint16_t>& MouseConfig::GetValidMinRateList()
 bool MouseConfig::ParseComModel(const std::string_view model_str,
                                 MouseModelCOM& model, bool& auto_msm)
 {
+	using enum MouseModelCOM;
+
 	if (iequals(model_str, ModelCom2Button)) {
-		model    = MouseModelCOM::Microsoft;
+		model    = Microsoft;
 		auto_msm = false;
 	} else if (iequals(model_str, ModelCom3Button)) {
-		model    = MouseModelCOM::Logitech;
+		model    = Logitech;
 		auto_msm = false;
 	} else if (iequals(model_str, ModelComWheel)) {
-		model    = MouseModelCOM::Wheel;
+		model    = Wheel;
 		auto_msm = false;
 	} else if (iequals(model_str, ModelComMsm)) {
-		model    = MouseModelCOM::MouseSystems;
+		model    = MouseSystems;
 		auto_msm = false;
 	} else if (iequals(model_str, ModelCom2ButtonMsm)) {
-		model    = MouseModelCOM::Microsoft;
+		model    = Microsoft;
 		auto_msm = true;
 	} else if (iequals(model_str, ModelCom3ButtonMsm)) {
-		model    = MouseModelCOM::Logitech;
+		model    = Logitech;
 		auto_msm = true;
 	} else if (iequals(model_str, ModelComWheelMsm)) {
-		model    = MouseModelCOM::Wheel;
+		model    = Wheel;
 		auto_msm = true;
 	} else {
 		return false;
@@ -124,14 +126,16 @@ bool MouseConfig::ParseComModel(const std::string_view model_str,
 
 static void set_capture_type(const std::string_view capture_str)
 {
+	using enum MouseCapture;
+
 	if (iequals(capture_str, CaptureTypeSeamless)) {
-		mouse_config.capture = MouseCapture::Seamless;
+		mouse_config.capture = Seamless;
 	} else if (iequals(capture_str, CaptureTypeOnClick)) {
-		mouse_config.capture = MouseCapture::OnClick;
+		mouse_config.capture = OnClick;
 	} else if (iequals(capture_str, CaptureTypeOnStart)) {
-		mouse_config.capture = MouseCapture::OnStart;
+		mouse_config.capture = OnStart;
 	} else if (iequals(capture_str, CaptureTypeNoMouse)) {
-		mouse_config.capture = MouseCapture::NoMouse;
+		mouse_config.capture = NoMouse;
 	} else {
 		assertm(false, "Invalid mouse capture value");
 	}
@@ -139,14 +143,16 @@ static void set_capture_type(const std::string_view capture_str)
 
 static void set_dos_driver_model(const std::string_view model_str)
 {
+	using enum MouseModelDos;
+
 	auto new_model = mouse_config.model_dos;
 
 	if (iequals(model_str, ModelDos2Button)) {
-		new_model = MouseModelDos::TwoButton;
+		new_model = TwoButton;
 	} else if (iequals(model_str, ModelDos3Button)) {
-		new_model = MouseModelDos::ThreeButton;
+		new_model = ThreeButton;
 	} else if (iequals(model_str, ModelDosWheel)) {
-		new_model = MouseModelDos::Wheel;
+		new_model = Wheel;
 	} else {
 		assertm(false, "Invalid DOS driver mouse model value");
 	}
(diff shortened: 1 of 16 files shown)


src/hardware/input/mouse_config.cpp
```diff
@@ -42,6 +42,10 @@ constexpr auto capture_type_onclick_str  = "onclick";
 constexpr auto capture_type_onstart_str  = "onstart";
 constexpr auto capture_type_nomouse_str  = "nomouse";
 
+constexpr auto model_dos_2button_str = "2button";
+constexpr auto model_dos_3button_str = "3button";
+constexpr auto model_dos_wheel_str   = "wheel";
+
```
johnnovak: Constexpr is UpperCamelCase by convention, and another minor comment, I'm 50/50 on the `_str` prefix. Dangerously close to Hungarian notation, and we don't put `_str` after every string constant.

src/hardware/input/mouse_config.cpp
```diff
@@ -131,7 +135,27 @@ static void SetCaptureType(const std::string_view capture_str)
...
+	auto new_model = mouse_config.model_dos;
+
+	if (model_str == model_dos_2button_str) {
+		new_model = MouseModelDos::TwoButton;
+	} else if (model_str == model_dos_3button_str) {
+		new_model = MouseModelDos::ThreeButton;
+	} else if (model_str == model_dos_wheel_str) {
+		new_model = MouseModelDos::Wheel;
+	} else {
+		assert(false);
```
johnnovak, line 149: Suggest to use `assertm`, e.g.:

```ini
assertm(false, "Invalid mouse model value");
```

src/hardware/input/mouse_config.cpp
```diff
@@ -146,14 +170,60 @@ static void SetPs2Model(const std::string_view model_str)
...
+		} else {
+			LOG_WARNING(
+			        "MOUSE: Invalid 'builtin_dos_mouse_driver_options' "
+			        "parameter: '%s', ignoring",
+			        token.c_str());
+			config_needs_sync = true;
+		}
+	}
+
+	if (config_needs_sync) {
```
johnnovak, line 207: Hmm, bit of an overkill, isn't it? Generally, config settings just revert to the default if an invalid setting provided.

So you could just do this:

```cpp
secprop.Add_string("builtin_dos_mouse_driver_options", always, DefaultBuiltinDosMouseDriverOptions);
```

Then you can ditch the whole `config_needs_sync` thing and simply do this after `LOG_WARNING`:
```cpp
		set_section_property_value("mouse",
		                           "builtin_dos_mouse_driver_options",
		                           DefaultBuiltinDosMouseDriverOptions);
```

src/hardware/input/mouse_config.cpp
```diff
@@ -399,18 +472,41 @@ static void config_init(Section_prop& secprop)
...
-	        "('off' by default). May improve gameplay, especially in fast-paced games\n"
-	        "(arcade, FPS, etc.), as for some games it effectively boosts the mouse\n"
-	        "sampling rate to 1000 Hz, without increasing interrupt overhead.\n"
-	        "Might cause compatibility issues. List of known incompatible games:\n"
-	        "  - Ultima Underworld: The Stygian Abyss\n"
-	        "  - Ultima Underworld II: Labyrinth of Worlds\n"
-	        "Please report it if you find another incompatible game so we can update this\n"
-	        "list.");
+	prop_bool = secprop.Add_bool("dos_mouse_driver", deprecated, true);
+	prop_bool->Set_help("Renamed to 'builtin_dos_mouse_driver'.");
```
johnnovak, line 476: Nice, that's much better; now the ambiguity is gone 💯

src/hardware/input/mouse_config.cpp
```diff
@@ -399,18 +472,41 @@ static void config_init(Section_prop& secprop)
...
+	                              always,
+	                              model_dos_wheel_str);
+	assert(prop_str);
+	prop_str->Set_values(
+	        {model_dos_2button_str, model_dos_3button_str, model_dos_wheel_str});
+	prop_str->Set_help(
+	        "Set the mouse model to be simulated by the built-in DOS mouse driver ('wheel'\n"
+	        "by default).\n"
+	        "  2button:  2 buttons, use for some old games which malfunction if the middle\n"
+	        "            button gets pressed.\n"
```
johnnovak, line 488: Hmm, I don't think a game being old is the differentiator here, plus "old" is a very vague term when the project is about emulating games from the 80s and 90s. All are old! 😆 

Most DOS games don't even utilise the right button. So probably 70% use the left button only, 28% the left+right buttons, and maybe 2% all 3 buttons. So out of ~7k DOS games, probably about 100-150 utilise the middle button, that sounds about right. Probably a lot less.

So I'd suggest wording this differently because mentioning "old games" doesn't really help the user. Plus Eye of the Beholder is from 1990 which misbehaves when the middle button is pressed, and I wouldn't call EoB an "old" game even in a strictly DOS gaming context. I reserve the "old" term for CGA and PCjr games, but it's somewhat subjective 😆 

So how about this:

```
- 2 buttons, the safest option for most games. The majority of DOS games assume a 2-button mouse and might misbehave if the third button is pressed.

- 3 buttons, only supported by a very few DOS games. In general, only use this if the game is known to support a 3-button mouse.

- 3 buttons + wheel, supports the CuteMouse WheelAPI version 1.0. No DOS game uses the mouse wheel, only a handful of DOS applications and Windows 3.x with special third-party drivers.
```

I also think we should really change the default from `wheel` to `2button` as that's the best overall option that results in the least unwanted side effects in most games. The `3button` option should be opt-in on a per game basis.

src/hardware/input/mouse_config.cpp
```diff
@@ -399,18 +472,41 @@ static void config_init(Section_prop& secprop)
...
+	        "  3button:  3 buttons, only supported by very few games.\n"
+	        "  wheel:    3 buttons + wheel, supports the CuteMouse WheelAPI version 1.0.");
+
+	prop_str = secprop.Add_string("builtin_dos_mouse_driver_options", always, "");
+	assert(prop_str);
+	prop_str->Set_help(
+	        "Additional built-in DOS mouse driver settings, space or comma separated\n"
+	        "(unset by default).\n"
```
johnnovak, line 496: How about this?

Additional built-in DOS mouse driver settings as a list of space or comma separated options.

src/hardware/input/mouse_config.cpp
```diff
@@ -399,18 +472,41 @@ static void config_init(Section_prop& secprop)
...
+	        "  immediate:  Update mouse movement counters immediately, without waiting for\n"
+	        "              interrupt. May improve mouse movement reaction time in fast-paced\n"
```
johnnovak, line 498: Isn't this simpler? "May improve mouse latency in..."

"but causes issues" should be "but might cause issues"

src/hardware/input/mouse_config.cpp
```diff
@@ -399,18 +472,41 @@ static void config_init(Section_prop& secprop)
...
+	        "              games (arcade, FPS, etc.), but causes issues in some titles.\n"
+	        "              List of known incompatible games:\n"
+	        "                - Ultima Underworld: The Stygian Abyss\n"
+	        "                - Ultima Underworld II: Labyrinth of Worlds\n"
+	        "              Please report it if you find another incompatible game so we can\n"
+	        "              update this list.\n"
```
johnnovak, line 504: `Please report other incompatible games so we can update this list.`

src/hardware/input/mouse_config.cpp
```diff
@@ -399,18 +472,41 @@ static void config_init(Section_prop& secprop)
...
+	        "  modern:     Enable modern (v7.0+) mouse driver behavior. Needed at least by\n"
+	        "              Descent II with the official Voodoo patch.");
```
johnnovak, line 506: I see you've shoehorned this in 😆 I was on the fence with this because it's a bit of a stretch to include a config option for a single game revision that's not even playable, lol, but perhaps it's useful for other very late DOS era games. We don't wanna our config to look like DOSBox-X's where you need to be a hardware expert to configure anything 😅 

Maybe this would be a bit more descriptive:

If provided, v7.0+ Microsoft mouse driver behaviour is emulated, otherwise the v6.0 and earlier behaviour (the two are slightly incompatible). Only Descent II with the official Voodoo patch has been found to require the v7.0+ behaviour so far.

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -60,7 +60,7 @@ static constexpr uint8_t  cursor_size_x  = 16;
 static constexpr uint8_t  cursor_size_y  = 16;
 static constexpr uint16_t cursor_size_xy = cursor_size_x * cursor_size_y;
```
johnnovak, line 61: While you're at it, maybe UpperCamelCase every constexpr in the file?

src/hardware/input/mouseif_dos_driver.cpp
```diff
@@ -60,7 +60,7 @@ static constexpr uint8_t  cursor_size_x  = 16;
...
 
-static constexpr uint8_t num_buttons = 3;
+static constexpr uint8_t MaxButtons = 3;
 
 enum class MouseCursor : uint8_t { Software = 0, Hardware = 1, Text = 2 };
```
johnnovak, line 65: If you need it to be `uint8_t` for binary "enum compatibility" with DOS drivers / interuppts, it's fine, otherwise just remove the width specifier if it's 100% native code.

src/hardware/input/mouse_config.cpp
```diff
@@ -530,15 +530,15 @@ static void config_init(Section_prop& secprop)
...
-	                      model_com_2button_msm_str,
-	                      model_com_3button_msm_str,
-	                      model_com_wheel_msm_str});
+	prop_str->Set_values({ModelCom2Button,
+	                      ModelCom3Button,
+	                      ModelComWheel,
+	                      ModelComMsm,
+	                      ModelCom2ButtonMsm,
+	                      ModelCom3ButtonMsm,
+	                      ModelComWheelMsm});
```
johnnovak, line 541: 🎖️ for doing the UpperCamelCase conversion 🎉

johnnovak:
@FeralChild64 Rebasing to master should fix the stuck macOS Clang CI jobs.

src/hardware/input/mouse_config.cpp
FeralChild64: Actually, I have already renamed these in the next commit - but OK, I have moved the cleanup here.

johnnovak: Yeah, I generally review commit by commit and I usually go back to delete my redundant commens,, but forgot about this one.

src/hardware/input/mouse_config.cpp
FeralChild64: Changed, also in other places in the file.

src/hardware/input/mouse_config.cpp
FeralChild64: Changed.

src/hardware/input/mouse_config.cpp
FeralChild64: Changed.

src/hardware/input/mouse_config.cpp
FeralChild64: Changed.

src/hardware/input/mouseif_dos_driver.cpp
FeralChild64: Changed to `auto` - this particular value is not returned to the guest side.

src/hardware/input/mouse_config.cpp
FeralChild64: True, a 2 button mouse is a better default - changed. Also changed the description to something very similar to your proposal.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, changed the description.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, simplified handling.

src/shell/command_line.cpp
```diff
@@ -161,6 +161,23 @@ bool CommandLine::FindStringBegin(const char* const begin, std::string& value,
...
+	size_t len = strlen(begin);
+	for (cmd_it it = cmds.begin(); it != cmds.end(); ++it) {
+		if (ciequals_n(begin, (*it).c_str(), len)) {
+			value = ((*it).c_str() + len);
+			if (remove) {
+				cmds.erase(it);
+			}
+			return true;
+		}
+	}
```
johnnovak, line 177: I'm not a fan of introducing yet another C-style string comparison function. We should try hard not to do this and convert C strings to `std::string` instead.

Just convert the snippet between the begin/end markers to a `std::string`, then use `iequals` on it from `string_utils.h`. `ciequals_n` here is a `micro-optimisation.` It does not matter one bit if we create a few temporariy `std::strings` when performing this find.

johnnovak: Actually, I'm pretty sure you can use string view with `iequals` because it's templated. So you won't even need to create temporary copies (not that it would matter).

src/hardware/input/mouse_config.cpp
```diff
@@ -198,28 +198,15 @@ static void set_dos_driver_options(const std::string_view options_str)
...
-			last_options_str += "modern";
-		}
+			set_section_property_value("mouse",
+			                           "builtin_dos_mouse_driver_options",
+			                           DefaultBuiltinDosMouseDriverOptions);
 
-		set_section_property_value("mouse",
-		                           "builtin_dos_mouse_driver_options",
-		                           last_options_str);
+			set_dos_driver_options(DefaultBuiltinDosMouseDriverOptions);
```
johnnovak, line 208: It would be nice if we logged the current settings when the users sets these options.

johnnovak:
I've tested this on Windows and it's working very well! Well done, @FeralChild64! 🥳  Just some minor further comments about the logging, but the functionality is spot-on! 🚀 

### Test results

- Tested **Eye of the Beholder** and **Gabriel Knight 1** with the below settings. All good, the middle button and buttons 4 & 5 are ignored. 🍏 Buttons 3, 4, & 5 uncapture the mouse if `mouse_middle_release` is enabled 🍏

  ```ini
  [mouse]
  builtin_dos_mouse_driver_options = immediate
  builtin_dos_mouse_driver_model = 2button
  ```

- Tested **King's Quest V** with:
  - `3button` + `mouse_middle_release = off` - the middle button and buttons 4 & 5 all perform the in-game middle button function 🍏 
  - `3button` + `mouse_middle_release = off` - buttons 4 & 5 perform the in-game middle button function, and the physical middle button uncaptures the mouse; very nice! 🏅🍏 
  - `2button` -- the middle button and buttons 4 & 5 are ignored in-game, and they uncapture the mouse if `mouse_middle_release` is enabled 🍏 

### Logging issues

This is what gets logged at startup when using the built-in mouse driver. This is confusing, the log statement only mentions PS/2, and there's no mentioning of the built-in driver being active. 

```
2025-04-26 18:20:26.776 | MOUSE: Will be captured after the first left or right button click
2025-04-26 18:20:26.776 | MOUSE: Middle button will capture/release the mouse (clicks not sent to the game/program)
2025-04-26 18:20:26.776 | MOUSE (PS/2): Standard, 3 buttons
```

I guess something like this instead of the last line would be a lot less ambiguous:

```
MOUSE: Using the built-in mouse driver (PS/2, standard, 3 buttons)
```

Also, regardless of whether we set `2button`, `3button`, or `wheel`, the log always says `MOUSE (PS/2): Standard, 3 buttons` which is clearly wrong.

Related to this, and I mentioned this in an individual comments as well, we should log the current state of the mouse whenever reconfiguring the mouse.

### Release notes

The release notes draft should at least mention the setting names that were deprecated and the new names that replaced them. The idea is that the user should get a reasonable amount of help to migrate to the new settings. So in general, the minimum we should do is give them the old and new names, and the command to look up further online help.

A good example that has the right level of detail; not too much, not too little IMO:
https://www.dosbox-staging.org/releases/release-notes/0.82.0/#cputype-normalisation

If you don't wanna spend too much time on writing the notes, that's okay -- it's a bit of an art to write these 😆 Just list the settings please so I won't need to read the PR when I flesh out the notes at release time

src/shell/command_line.cpp
FeralChild64: Reworked to use `iequals` based solution.

src/hardware/input/mouse_config.cpp
FeralChild64: OK, added a log.

FeralChild64:
Updated the release notes proposal. Regarding the logs - a log:

```MOUSE: Using the built-in mouse driver (PS/2, standard, 3 buttons)```

would be very wrong:

- `MOUSE (PS/2):` is a log prefix related to the PS/2 port mouse
- `MOUSE (COM1):` is a log prefix related to the COM1 (serial) port mouse
- `MOUSE (DOS):` is a log prefix related to the built-in DOS mouse driver

These 3 are almost totally independent. I have harmonized the model/protocol logs, default options will now print:

```
MOUSE (DOS): 2 buttons
MOUSE (PS/2): 3 buttons
```

**Comment by johnnovak** [comment]
Thanks for updating the release notes with more detail @FeralChild64 

You explained the PS/2 protocol vs emulated mouse driver differences [in this comment of yours](
https://github.com/dosbox-staging/dosbox-staging/issues/4197#issuecomment-2666961514), but that's far too much low level info for regular users. You can change the PS/2 protocol type logging to a DEBUG statement if it's only for dev work and troubleshooting, as you said -- it's confusing for the user to say 2 button in one line, then 3 button right in the next line, so let's not do that in the regular logs.

Problem is, I look at this and I'm scratching my head:

```
MOUSE (DOS): 2 buttons
MOUSE (PS/2): 3 buttons
```
 I have configured a 2-button mouse, so this information is confusing as hell. Let's be real, 99% of our users will never emulate multiple mice, serial mice, all those weirdness, etc. So can we tweak these messages so they're crystal clear for this 99% use case? Then you can log whatever in the complex scenarios, but can we make the common 99% single mouse path logging as clear and unambiguous as possible? I feel we're not even close to that goal yet 😆 

For example:

```
MOUSE (DOS): Using built-in 2-button DOS mouse driver in immediate mode
```

That's a clear and unambigious statement.

You could log this with variations depending on the options and configured mouse model, of course. If the user changes any built-in mouse setting, we log this line again.

If we disable the built-in mouse driver, we can simply omit this.

If the user configures all the complex serial nd multi-mice scenarios, you can go crazy with the logging (as long as it's sensible).

We should make the 99% path straightforward and easy. Then might make the 1% possible, but not at the detriment of the 99% path. I'm sure you current logging is "technically correct", but that does not help usability.

Thanks 😃

**Comment by johnnovak** [comment]
> I'll make the logs more descriptive. I'll also try to make them appear for the first time only when something starts interacting with the DOS driver, PS/2 AUX port, or the PS/2 mouse BIOS functions (serial mouse emulation already works this way) - this way DOS games won't trigger the PS/2 mouse logs, hopefully this will be less confusing.

I think that's a good idea if it's feasible. The high level goal is remove confusion; then how exactly that is accomplished, you're the best person to decide that as you know more about input handling than the rest of us combined 😆 
 
> If you think it is a good idea, we can also cherry-pick [2edf866](https://github.com/dosbox-staging/dosbox-staging/commit/2edf8667412e68202144913a1d417eebd55fa3d1) from one of my experimental branches - this way, under normal circumstances, the PS/2 mouse logs will appear somewhere near the _Starting Microsoft Windows_ log, hopefully lessening the confusion.

I like that idea. Anything that makes things more obvious for the end-user.

> But for maintainability I would prefer to keep the DOS driver parameters as a separate log - right now we have just two parameters, in the future we might need more, for example if I ever decide to emulate the _Logitech CyberMouse 3D_ (https://bitsavers.org/pdf/logitech/cyberman/CyberMan_3D_SWIFT_Supplement_Ver_1.0_upd_199401.pdf) or anything crazy like this :D

Yeah that's fine, of course. Just make things easy to understand.

Thanks for the patience on this 😄

**Comment by FeralChild64** [comment]
Done - but I've decided to keep the DOS mouse driver log during the startup; Windows 3.11 triggers the DOS mouse driver when it starts, so we would have both DOS and PS/2 mouse log at almost the same time.

**Comment by johnnovak** [comment]
@FeralChild64 

The logging is very clear now, and we get log statements when changing the mouse driver model as well. Detecting Windows 3.x and logging it is also a nice touch, great job 💯 

One last very minor improvement suggestion:

```
MOUSE (DOS): Built-in driver is simulating a '2 buttons' model
```

That's alright, but `simulating a 2-button model` is a bit better English.

Approving, nice job! 🎉

**Review by johnnovak — APPROVED** [review]

**Comment by FeralChild64** [comment]
@johnnovak PVS Studio got broken:

```
Run-time dependency iir found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency iir
Building fallback subproject with default_library=shared

meson.build:672:13: ERROR: Unhandled python exception
```

Interestingly, trying to bring-up the https://github.com/dosbox-staging/dosbox-staging/pull/4288, I've also run into the `iir` problem.

johnnovak:
> @johnnovak Adapted the logs.
> 
> But now the PVS Studio go broken:
> 
> ```
> Run-time dependency iir found: NO (tried pkgconfig and cmake)
> Looking for a fallback subproject for the dependency iir
> Building fallback subproject with default_library=shared
> 
> meson.build:672:13: ERROR: Unhandled python exception
> ```
> 
> Interestingly, when trying to bring-up the #4288, I've also run into the `iir` problem.

I fixed PVS in https://github.com/dosbox-staging/dosbox-staging/pull/4299. Maybe let's merge that first, then rebase your PRs.

johnnovak:
@FeralChild64 As you can see, we're also experimenting with SonarQube with @kklobe. You can ignore those warnings for now, we literally enabled it like an hour ago.

johnnovak:
Nice job addressing the high severity SonsrQube issues @FeralChild64 🎉 

Are you planning to address the rest of them too? I skimmed through them and I think they're not essential but are good suggestions that would improve code quality (except for the TODO related one).

I first thought the `std::byte` suggestion was overkill, but I looked into it a bit and it seems it's a great language feature worth experimenting with. If it's not cumbersome to use, it might be a better alternative to our current use of uints for dealing with binary protocol / register data.

https://runebook.dev/en/articles/cpp/types/byte

FeralChild64:
Quite possibly. I haven't investigated all of them yet.

FeralChild64:
Unfortunately, I'll have to withdraw the `uint8_t` -> `std::byte` - this has triggered a massive amount of other changes requested, mostly related to union and `bit_view` mechanism, many with A severity, blocking the merge.

I can't do a 2000 LOC rework of legacy code to satisfy the static analysis.

FeralChild64:
I've fixed all the non-blocking warnings which seemed to me to be reasonable to fix. Few are left.

Medium priority:
- [Value stored to 'new_model' during its initialization is never read](https://sonarcloud.io/project/issues?pullRequest=4284&open=AZaJMoNsxiGm4L9OvHC1&id=dosbox-staging_dosbox-staging) - I don't agree here; this is a fail-safe mechanism, on non-debug build the assertion in the function is ignored
- [Use "std::byte" for byte-oriented data manipulation](https://sonarcloud.io/project/issues?pullRequest=4284&open=AZagnh2_gd8FZtumm9lp&id=dosbox-staging_dosbox-staging) - changing the unions to use `uint8_t` instead of `std::byte` triggers a huge chain of compilation problems and static analysis warnings to resolve; this should be a separate rework

Low priority:
- [Use the init-statement to declare "idx" inside the if statement.](https://sonarcloud.io/project/issues?pullRequest=4284&open=AZaJMoNBxiGm4L9OvHCy&id=dosbox-staging_dosbox-staging), [Use the init-statement to declare "idx" inside the if statement](https://sonarcloud.io/project/issues?pullRequest=4284&open=AZaJMoNBxiGm4L9OvHCz&id=dosbox-staging_dosbox-staging) - I don't agree here; the code would be harder to understand.
- [Replace this const reference to "std::string" by a "std::string_view"](https://sonarcloud.io/project/issues?pullRequest=4284&open=AZacHv_-IIvN3Cj6iZBB&id=dosbox-staging_dosbox-staging)  - I prefer not to do this now; the whole file should be reworked to get rid of the C-ism and to accept `std::string_view` everywhere; this should be a separate rework

johnnovak:
> I've fixed all the non-blocking warnings which seemed to me to be reasonable to fix. Few are left.

Thanks for addressing these @FeralChild64, great job! 🎉  I've had a second more detailed look now and I agree with your assessments. We should treat the non-critical warnings as guidelines and do what's sensible and practical for us, yeah.

`std::byte` is a nifty thing, but not sure how practical it is in practice when half the codebase uses `uin8_t` already. I will play around with it in new code, but maybe it's more hassle than worth in actual practice, despite the "theoretical" benefits.

PR #4302, opened 2025-05-02
Fix edge case where CDROM mixer hangs at the end of a track

### LINKED ISSUE #4296: Tomb Raider 1: sound disappears when CD audio track ends

_No response_

Linux x86_64

Fedora Kinoite (KDE), dosbox-staging from flathub

### Relevant hardware info

Intel 6th gen with iGPU, likely irrelevant

When playing Tomb Raider 1 (pulled from GOG release) on 0.82.1, sound disappears after a CD audio track ends. The sound reappears when exiting to menu, but further playing will make the sound disappear again.

I'm using 3dfx patch, but it doesn't seem to be different without it.

I have tried disabling sbmixer or mess around with buffer sizes without any difference. Same goes with cycles, albeit I haven't gone under 60k for playability reasons.

I tried a downgrade 0.81.0 (https://github.com/flathub/io.github.dosbox-staging/commit/00b2655f3206747b8e69a13705c422bfdb6e1511) and that actually helped. Downgrade just to 0.82.0 (https://github.com/flathub/io.github.dosbox-staging/commit/a6644490a56670ed5abff9b3b2a1ea5750bed6d4) didn't help. I hope to not stick to 0.81 because of way inferior voodoo experience there though :)

I can't really tell if it is specific to TR1 or not, but I have also played Magic Carpet Plus (The same SB16 config for sound, albeit with MIDI instead of TR1's CD audio for music) on the same setup extensively without any sound issues so at least it's not everywhere.

1. Start dosbox-staging
2. mount the gog disk image (mixed CUE with music in mp3)
3. run the game
4. start the first level and wait a little over 2 minutes (the length of the CD audio track played at that point).
5. the whole sound is gone. Pulling pistols and shooting makes no sound at all. 
Loading later save has identical behaviour. Haven't tried past the first level with this though.

https://www.gog.com/en/game/tomb_raider_123

### Your configuration

```ini
Generated default except for setting midi device to fluidsynth instead of auto and setting GeneralUser soundfont.
```

### Provide a log

```text
Nothing in log seems relevant - no warnings or errors
```

- [x] Yes, I agree.

weirddan455:
# Description

When the track got to the end, it was doing some back-and-forth logic going between time and sector number to figure out where to seek next.

It can hang if either the conversion is slightly off or the MP3 decoder returns slightly fewer frames that it was expecting.

Even on MP3 files where it didn't hang, it sometimes did 10+ seeks back-to-back each seeking to the last second of the file.

New code skips all that conversion business and just goes to the start of the next track :)

## Related issues

Fixes #4296

# Release notes

Fix audio hanging after about 2 minutes of play in the GOG version of Tomb Raider

# Manual testing

Tested both MP3 and .bin audio of Tomb Raider. Tested both with and without the 2 second pregap.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/cdrom_image.cpp
+++ b/src/dos/cdrom_image.cpp
@@ -531,6 +531,7 @@ CDROM_Interface_Image::~CDROM_Interface_Image()
 
 bool CDROM_Interface_Image::SetDevice(const char* path)
 {
+	std::lock_guard lock(player.mutex);
 	const bool result = LoadCueSheet(path) || LoadIsoFile(path);
 	if (!result) {
 		// print error message on dosbox console
@@ -701,41 +702,20 @@ bool CDROM_Interface_Image::GetMediaTrayStatus(bool& mediaPresent, bool& mediaCh
 	return true;
 }
 
-bool CDROM_Interface_Image::PlayAudioSector(uint32_t start, uint32_t len)
+bool CDROM_Interface_Image::PlayAudioTrack(const Track& track, const uint32_t sector_offset)
 {
-	// Find the track that holds the requested sector
-	track_const_iter track = GetTrack(start);
-	std::shared_ptr<TrackFile> track_file;
-	if (track != tracks.end())
-		track_file = track->file;
-
-	// Guard: sanity check the request beyond what GetTrack already checks
-	if (len == 0 || track == tracks.end() || !track_file ||
-	    track->attr == 0x40 || !player.channel) {
+	const auto track_file = track.file;
+	if (!track_file || track.attr == 0x40 || !player.channel) {
 		StopAudio();
-#ifdef DEBUG
-		LOG_MSG("CDROM: PlayAudioSector => sanity check failed");
-#endif
 		return false;
 	}
-	// If the request falls into the pregap, which is prior to the track's
-	// actual start but not so earlier that it falls into the prior track's
-	// audio, then we simply skip the pre-gap (beacuse we can't negatively
-	// seek into the track) and instead start playback at the actual track
-	// start.
-	if (start < track->start) {
-		len -= (track->start - start);
-		start = track->start;
-	}
 
-	// Calculate the requested byte offset from the sector offset
-	const auto sector_offset = start - track->start;
-	const auto byte_offset = track->skip + sector_offset * track->sectorSize;
+	const auto byte_offset = track.skip + sector_offset * track.sectorSize;
 
 	// Guard: Bail if our track could not be seeked
 	if (!track_file->seek(byte_offset)) {
 		LOG_MSG("CDROM: Track %d failed to seek to byte %u, so cancelling playback",
-		        track->number, byte_offset);
+		        track.number, byte_offset);
 		StopAudio();
 		return false;
 	}
@@ -750,10 +730,9 @@ bool CDROM_Interface_Image::PlayAudioSector(uint32_t start, uint32_t len)
 	// Update our player with properties about this playback sequence
 	player.cd = this;
 	player.trackFile = track_file;
-	player.startSector = start;
-	player.totalRedbookFrames = len;
 	player.isPlaying = true;
 	player.isPaused = false;
+	currentTrackIndex = track.number - 1;
 
 	// Assign the mixer function associated with this track's content type
 	if (track_file->getEndian() == AUDIO_S16SYS) {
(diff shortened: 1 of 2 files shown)


src/dos/cdrom_image.cpp
```diff
@@ -766,15 +745,53 @@ bool CDROM_Interface_Image::PlayAudioSector(uint32_t start, uint32_t len)
...
+	if (len == 0 || track == tracks.end()) {
+		StopAudio();
+#ifdef DEBUG
+		LOG_MSG("CDROM: PlayAudioSector => sanity check failed");
+#endif
+		return false;
+	}
+	// If the request falls into the pregap, which is prior to the track's
+	// actual start but not so earlier that it falls into the prior track's
+	// audio, then we simply skip the pre-gap (beacuse we can't negatively
```
johnnovak, line 771: typo: because

src/dos/cdrom_image.cpp
```diff
@@ -1001,13 +1015,28 @@ bool CDROM_Interface_Image::ReadSectorsHost(void *buffer, bool raw, unsigned lon
...
+void CDROM_Interface_Image::PlayNextAudioTrack()
+{
+	if (currentTrackNumber >= tracks.size()) {
+		StopAudio();
+		return;
+	}
+
+	// track number is index + 1 so current track number == next track index
+	const auto &next_track = tracks[currentTrackNumber];
+	assert(next_track.number == currentTrackNumber + 1);
```
johnnovak, line 1027: Hmm, why don't we store `currentTrackIndex` instead internally, then things would become more obvious, perhaps?

johnnovak:
@weirddan455 Rebasing to master should fix the PVS issue I think.

Does this only affect the MP3 encoded audio track path? If I understand it correctly, seeking is not affected in general, just the case when we move from one track to the next? Maybe it would be worthwhile to test a few more games before merging? 

I've tried to look up in the history why this complicated seeking was implemented, but haven't found much. Maybe this one is related, but can't be certain:

https://github.com/dosbox-staging/dosbox-staging/commit/eaeb001

weirddan455:
> Does this only affect the MP3 encoded audio track path?

MP3 is the only way I've been able to reproduce the bug but it could theoretically affect everything. It likely depends on decoder implementation details. Ex: We seek to "end - N seconds" or less. If the decoder gives us 0 audio frames, we get stuck in an infinite loop. The real bug is what I fixed in the higher level code to eliminate that as a possibility.

> If I understand it correctly, seeking is not affected in general, just the case when we move from one track to the next?

Yes. The only way for a DOS game to seek is through `PlayAudioSector()`. That remains unchanged save for extracting some shared code out to a new function. We just don't want to call that function on track change or we can end up seeking back into the same track the decoder just returned 0 frames on.

> I've tried to look up in the history why this complicated seeking was implemented, but haven't found much. Maybe this one is related, but can't be certain

My assumption is that whoever wrote it wanted to avoid code duplication with `PlayAudioSector()` since that takes care of resetting some player variables.

src/dos/cdrom_image.cpp
weirddan455: Internally where? I don't know what you mean by that.

johnnovak:
> > Does this only affect the MP3 encoded audio track path?
> 
> MP3 is the only way I've been able to reproduce the bug but it could theoretically affect everything. It likely depends on decoder implementation details. Ex: We seek to "end - N seconds" or less. If the decoder gives us 0 audio frames, we get stuck in an infinite loop. The real bug is what I fixed in the higher level code to eliminate that as a possibility.
> 
> > If I understand it correctly, seeking is not affected in general, just the case when we move from one track to the next?
> 
> Yes. The only way for a DOS game to seek is through `PlayAudioSector()`. That remains unchanged save for extracting some shared code out to a new function. We just don't want to call that function on track change or we can end up seeking back into the same track the decoder just returned 0 frames on.
> 
> > I've tried to look up in the history why this complicated seeking was implemented, but haven't found much. Maybe this one is related, but can't be certain
> 
> My assumption is that whoever wrote it wanted to avoid code duplication with `PlayAudioSector()` since that takes care of resetting some player variables.

Cheers, it seems rather safe then. I was just thinking whether is it worth backporting this to 0.82.2 or is it too risky.

src/dos/cdrom_image.cpp
weirddan455: Oh, number to index. That's the part I didn't see. Yeah I can do that.

src/dos/cdrom_image.cpp
weirddan455: fixed

src/dos/cdrom_image.cpp
weirddan455: done

johnnovak:
@Python-Exoproject I'm not too sure if you use MP3 audio tracks for any of your games. I've spot checked about 10 and all seem to be using CUE/BIN with a single BIN file, which is a solid choice 👍🏻 Tagging you anyway so you're aware of the potential issue.

Issue #4094, opened 2024-12-07
Small localisation related logging issues


LowLevelMahn:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

git

### What Operating System are you using?

None

### If Other OS, please describe

Windows,Linux

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Warnings/Message in log output - happens only under windows

`2024-12-07 21:42:43.340 | LOCALE: Message 'PROGRAM_MOUNT_READONLY' not found `
comes up because AddCommonMountMessages() gets exited in the first line, so all common mount message are skiped at that time 

`2024-12-07 21:55:35.188 | UNICODE: Problem decoding UTF8 string, position 0`
comes up because the routine utf8_to_wide just logs that string - the condition seems wrong or not correct 

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
@FeralChild64 Perhaphs you can confirm if this is still an issue? Can't keep track with the localisation changes :laughing:

LowLevelMahn:
i'll do the test!

LowLevelMahn:
fresh git + fresh vcpkg + DEBUG + Windows 10/x64 build + freshly created config (by dosbox itself)

**FIXED**: the `LOCALE: Message 'PROGRAM_MOUNT_READONLY' not found` does not occure

**NOT FIXED**: the `UNICODE: Problem decoding UTF8 string, position 0` warning still occures two times in the startup log

```
date       time         |
2025-03-18 07:26:37.654 | arguments: D:\\temp\\dos_stuff\\dosbox_dev\\dosbox-staging_clean\\vs\\x64\\Debug\\dosbox.exe
2025-03-18 07:26:37.654 | Current dir: D:\temp\dos_stuff\dosbox_dev\dosbox-staging_clean\vs
2025-03-18 07:26:37.654 | stderr verbosity: 0
2025-03-18 07:26:37.654 | -----------------------------------
2025-03-18 07:26:37.654 | dosbox-staging version 0.83.0-alpha (git)
2025-03-18 07:26:37.654 | ---
2025-03-18 07:26:37.655 | LOG: Loguru version 2.1.0 initialised
[...]

PR #4309, opened 2025-05-06
C++ library locale autodetection fixes

FeralChild64:
# Description

This PR fixes some problems related to locale autodetection using the C++ standard library:
- **Windows**: disabled currency format detection due to test string returned by the C++ standard library not being UTF-8
- **Linux**: only perform currency format detection if the monetary locale is UTF-8
- **macOS**: do not use the C++ standard library for locale detection, as it returns dummy values; use _CoreFoundation_ functions instead

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4094

# Release notes

(not needed - bugfix for a feature which wasn't released yet)

# Manual testing

- make sure `locale_period = native` in the config file
- start DOSBox, check that no `UNICODE: Problem decoding UTF8 string` is printed out
- execute DOSBox  `dir` command, check that time/date format and file size format follows the host OS settings
- repeat with different locale settings (on some OSes or desktop environments you might need to re-login for the new settings to be detected by DOSBox)

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/misc/host_locale.cpp
+++ b/src/misc/host_locale.cpp
@@ -417,8 +417,14 @@ void StdLibLocale::GetDateFormat(const std::locale& locale)
 	}
 }
 
+#if !defined(WIN32)
 void StdLibLocale::DetectCurrencyFormat(const std::locale& locale)
 {
+	// We can only get a suitable data if monetary locale format is UTF-8
+	if (!IsMonetaryUtf8(locale)) {
+		return;
+	}
+
 	// Retrieve currency code
 	constexpr bool International = true;
 	const auto& format_code =
@@ -479,6 +485,7 @@ void StdLibLocale::DetectCurrencyFormat(const std::locale& locale)
 	currency_code_format = detect_format(money_code_example, currency_code);
 	currency_utf8_format = detect_format(money_code_example, currency_utf8);
 }
+#endif
 
 void StdLibLocale::DetectTimeDateFormat(const std::locale& locale)
 {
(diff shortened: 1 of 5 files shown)


FeralChild64:
Do I understand correctly, the _Sonar Qube_ complains that by changes to `host_locale_win32.cpp` are too similar to changes to `host_locale_macos.cpp`?

johnnovak:
> Do I understand correctly, the _Sonar Qube_ complains that by changes to `host_locale_win32.cpp` are too similar to changes to `host_locale_macos.cpp`?

Removing _every duplication of code_ results in a big mess in my experience; some judicious duplication is actually better and results in more maintainable code.

I see this as "soft persuasion" because hard-failing a build if more than 3% (!) of the new code is a "duplicate" is just bullshit. That's not a useful rule, and there's way to ignore those warnings on the web UI like you can with issues. You can create new "quality gate" profiles... but only in the paid plan. We don't wanna pay $35 per month for this privilege 😆

Unless @kklobe who has Sonar experience has a solution, I'm inclined to give the core team rights to force merge in instances like this. I think Sonar is worth keeping because most of the issues it flagged were genuinely useful, but we wanna get over the "no duplicate code" requirement without paying the monthly fee.

So yeah, force merge privileges is one way to go and I trust the team not to abuse this.

Another way would be to make the SonarQube task a "soft check", so a failed Sonar job wouldn't result in a failed build. I just don't see how it integrated into our workflow... we don't have a workflow file for it, it just did it via the GitHub API after I pointed SonarQube to our repo. Bit of a black box...

johnnovak (approved):
Looks fine at the code level.

Issue #4196, opened 2025-02-14
Migrate the build system to CMake


FeralChild64:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Other

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Is your feature request related to a problem? Please describe.

List of tasks to be completed before the CMake support can be declared as production-ready. Feel free to add more:

[ ] custom build should show GIT hash on the startup banner and in the VER command output
[ ] ManyMouse support, port detection code for macOS
[ ] FluidSynth usage should be optional
[ ] MT32Emu usage should be optional
[ ] Tracy usage should be optional
[ ] networking should be optional, depending on SDL2_Net
[ ] proper compiler detection in the main CMakeList.txt
[ ] enable certain warnings-as-errors in the main CMakeList.txt
[ ] handle compiler warnings only showing with the CMake build
[ ] add Debian VCPKG build documentation
[ ] add Fedora VCPKG build documentation
[ ] add OpenSuSE VCPKG build documentation
[ ] add macOS VCPKG build documentation
[ ] add Windows VCPKG build documentation
[ ] create CI jobs for CMake builds

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
> * Platform builds - please add RISC-V build to the Linux workflow, if you're going to disable Platform builds
> 
>     * Config heavy - what is the background story on this? I see it helped in some situations:
>       
>       * "Config Heavy then revealed a new missing symbol" [Add lib iphlpapi for static SDL builds under Windows #494](https://github.com/dosbox-staging/dosbox-staging/pull/494)
>       * "Config Heavy gets credit for catching this one" [Ensure SDL2_net preceeds SDL2 in the link-order #456](https://github.com/dosbox-staging/dosbox-staging/pull/456)

No to both.

We don't officially support RISC-V.

The actual release package builds will reveal what needs revealing. We don't need to "reveal" weird stuff that nobody uses.

weirddan455:
CMake bug I just found: Delete `dosbox-staging.conf` and launch DOSBox to regenerate it. This shows in the log

```
2025-05-13 01:48:08.166 | CONFIG: No help available for 'ne2000'.
2025-05-13 01:48:08.166 | CONFIG: No help available for 'nicbase'.
[...]

PR #4316, opened 2025-05-10
Fix Git hash injection in CI and local builds

johnnovak:
# Description

The Git build hash was not injected into our CI builds (I've checked the macOS build, but I'm pretty sure Linux and Windows was also affected). This fixes the issue. 

Also addresses the "custom local build should show GIT hash on the startup banner and in the VER command output" item of our [CMake migration ticket](https://github.com/dosbox-staging/dosbox-staging/issues/4196) "for free"

This also paves the way for replacing MSBuild with CMake in MSVC as now the Git hash injection is shell-agnostic; it only need Git to be available via CMake which is a given in the MVSC + CMake integration.

## Related issues

# Release notes

N/A

# Manual testing

I've tested it on macOS and Windows, see the screenshots below. Can you Linux people please test the Linux local and CI builds? @weirddan455 or @FeralChild64 ?

- [x] Windows
- [x] macOS
- [x] Linux

## macOS local build

image

## macOS CI build

image

## Windows CI build

This will only work with local builds when we migrate to building via CMake in MSVC too.

image

# Checklist

I have:


FeralChild64 (approved):

FeralChild64:
It's probably better not to execute `git add .`, it will mess up with the repo. How about checking the output of

```git status --porcelain```

command, as suggested here: https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git?

johnnovak:
@FeralChild64 Good point. I posted that right before going to sleep so I didn't quite think it through 😄 The "porcelain" command seems to be better 👍🏻

shermp:
I think this approach will only update if something triggers cmake to reconfigure - it might not change for every commit. Fine for CI though which is doing a fresh configure every time.

johnnovak:
> I think this approach will only update if something triggers cmake to reconfigure - it might not change for every commit. Fine for CI though which is doing a fresh configure every time.

Indeed. I've wasted no less than 4 hours trying to inject the Git hash at build time as @FeralChild64 suggested, but I couldn't get it to work. I'm going to merge this as is (plus a reformat commit), then if anyone has the appetite for playing around with this, be my guest. The last two commits on this branch is my best attempt:

https://github.com/dosbox-staging/dosbox-staging/compare/main...jn/cmake-version-2

But I don't think this can be made to work reliable anyway. E.g., imagine you've just made some changes that you haven't checked in yet, and you've built the project. That will show `HASH-dirty`, in principle. Then you make no changes to the source code whatsoever, only check in the changed files and a bit later build again. Well, that won't trigger a rebuild, so instead of seeing `NEXT_HASH` you'll still see `HASH`.

That's just confusing, and IMO this would add too much complication for little benefit. A half-baked solution is only confusing, and I think only half-baked solutions are our options. C/C++ and Git integration via various command-line tools is just not very tight.

So, I think only injecting the hash at configure time is fine. Merging after the build turns green again.

PR #4328, opened 2025-05-15
Spring cleaning, part 1 🌷 🧹 🗑️ 🔥 

johnnovak:
# Description

The `contrib` and `scripts` folders haven't been in a good shape for a while now; somehow they became a dumping ground for various scripts that I'm not even sure anybody has been using for anything lately. I certainly haven't 😄 

So now I'm deleting everything that I personally never used and is not references anywhere. If something you think is worth keeping, let me know.

I've also started organising the scripts in a more logical manner, i.e. grouping them by function in subdirectories. This is just the first sweep, there's a lot more to do. I'm gonna address that in subsequent PRs (e.g., moving stuff fromt `contrib` to `scripts`, and making some order in `contrib` too because it's a mess).

# Checklist

I have:


FeralChild64 (commented):
I don't know any of these removed scripts either. If no one protest, remove them.

weirddan455 (approved):
I've never used any of these scripts.

I haven't heard of AutoFDO before either. It sounds like a cool concept but probably more trouble than it's worth for DOSBox.

I'm not sure why the move to `scripts/tools`. `compile_commits.sh` and `format-commit.sh` are the only 2 scripts I ever really use and they're stand-alone for me, not part of some larger tool. I suppose I don't care too much except for having to type 5 extra characters when I want to run them.

johnnovak:
> I've never used any of these scripts.

👍🏻 

> I haven't heard of AutoFDO before either. It sounds like a cool concept but probably more trouble than it's worth for DOSBox.

Exactly my thoughts. Seems like an experimental thing with dubious utility and that README I removed is super detailed, it kinda implies this is something we do and is useful. 

I'm not against experimentation, of course, but what ends up in the `main` branch should be solid and stuff we actually use. Experiments and random test programs can be stored in private branches or even another repo in the DOSBox Staging org.

> I'm not sure why the move to `scripts/tools`. `compile_commits.sh` and `format-commit.sh` are the only 2 scripts I ever really use and they're stand-alone for me, not part of some larger tool. I suppose I don't care too much except for having to type 5 extra characters when I want to run them.

Yeah, I've been wondering about the same thing myself. This is the state where I ended up in my second part: 

image

One one hand, I like the logical grouping, on the other hand, I'm wondering if it's too much. Maybe I could flatten it to something like this:

```
scripts
    build
        (contents of ci and linting)
    packaging
        (remove linux/macos/windows subfolders)
    svn-import
        (to keep it separate)
    (contents of tools goes under scripts)
```

Thoughts? I'm leaning towards flattening it like this, after all.

FeralChild64:
I like the layout from the screenshot, and I think some categorization of the scripts is a good idea.

johnnovak:
Alright, I'm keeping the directory structure in `scripts` for now. We might revisit this later, but at least now every script will be sorted into logical categories.

PR #4330, opened 2025-05-16
Dynamically load Slirp library at runtime

shermp:
# Description

Part 2 in the quest to remove glib as a build dependency.

Since the whole feature is a copy-paste job from #4222 let's just copy the description as well:

> slirp depends on glib. Compiling Glib on windows using VCPKG takes an eternity. This PR switches from linking slirp at compile time to loading the slirp library at runtime, falling back to host MIDI output if the library cannot be loaded.
>
> Our official releases can include pre-compiled slirp binaries for a seamless transition. Exactly how this will work is TBD.
>
> The slirp library is only loaded if slirp is enabled in the config.

Hopefully I've scrubbed the evidence from the code...

# Manual testing

- [x] Windows
- [x] macOS
- [x] Linux

Smoke tested on Windows 11 using the following testing pack provided by @kcgen a long time ago.

[ethernet-pack.zip](https://github.com/user-attachments/files/20242415/ethernet-pack.zip)

Simply extract the pack, and start dosbox from the ethernet-pack/ethernet directory. It will run a couple of tests, concluding with a ping to google.

Note, the ping to google currently fails, both in this PR, and `main`. It seems to be a recent regression, as it successfully pings on 0.82.1.

# Checklist

I have:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -180,6 +180,11 @@ jobs:
       - name: Dump workspace contents
         run: find $RUNNER_WORKSPACE
 
+      - name: Verify executable RPATH
+        run: |
+          set -x
+          objdump -x build/release-linux/dosbox | grep \$ORIGIN/lib
+
       - name: Package
         run: |
           echo "*** dosbox transitive deps:"
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -80,10 +80,19 @@ jobs:
           cmake --preset release-macos-${{ matrix.conf.arch }}
           cmake --build --preset release-macos-${{ matrix.conf.arch }} 2>&1 | tee build.log
 
+      - name: Dump workspace contents
+        run: find $RUNNER_WORKSPACE
+
       - name: Summarize warnings
         env:
           MAX_WARNINGS: ${{ matrix.conf.max_warnings }}
         run:  python3 ./scripts/count-warnings.py -lf build.log
+
+      - name: Verify executable RPATH
+        run: |
+          set -x
+          otool -l build/release-macos-${{ matrix.conf.arch }}/Release/dosbox \
+            | grep "path @executable_path/lib"
 
       - name: Upload binary
         uses: actions/upload-artifact@v4
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -129,7 +129,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 170
+          MAX_BUGS: 176
         run: |
           echo "Full report is included in build Artifacts"
           echo
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,19 +195,19 @@ configure_file(
 
 ### end of config.h ###
 
-add_executable(dosbox src/main.cpp src/dosbox.cpp)
-target_include_directories(
-  dosbox PUBLIC include ${CMAKE_CURRENT_BINARY_DIR}/include
-)
-
 # Allow dynamically loading externalised vcpkg dependencies at runtime
-# from the $EXE_DIR/lib directory
+# from the $EXE_DIR/lib directory (must be set before 'add_executable')
 if (APPLE)
   set(CMAKE_INSTALL_RPATH "@executable_path/lib")
 else()
   set(CMAKE_INSTALL_RPATH "$ORIGIN/lib")
 endif()
 set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
+
+add_executable(dosbox src/main.cpp src/dosbox.cpp)
+target_include_directories(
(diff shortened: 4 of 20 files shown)


shermp:
The PVS failure appears to be related to the NE2000 code that was (re)enabled in this PR.

shermp:
I don't think this PR is going to pass sonarqubecloud. Too many issues from the included header, and it doesn't like the callbacks in `ethernet_slirp.cpp`.

johnnovak:
> I don't think this PR is going to pass sonarqubecloud. Too many issues from the included header, and it doesn't like the callbacks in `ethernet_slirp.cpp`.

The warnings have a "time-to-live" of 30 days, I believe. Unfortunately, we cannot exclude files/directories in the free version, and the paid one is $20-30 a month.

shermp:
Ok, the ping might be a windows only thing, and probably not much of an issue.

At the suggestion of @weirddan455 I tried `telnet telehack.com`, and it works fine. Also tested on Fedora 41 with it's system slirp.

johnnovak (approved):
Looks fine to me. Great work @shermp, thanks a lot for doing this! 🏅 💯 

I will test this on macOS, but probably only tomorrow.

johnnovak:
@shermp Okay, I've tested this on macOS and ran into a bunch of problems. I've fixed all of them, Can you please cherry-pick my commits from https://github.com/dosbox-staging/dosbox-staging/compare/jn/slirp-dynamic-load-2 ?

Specifically:

- I fucked up RPATH handling by moving the RPATH setting stuff around in the CMake config. So the the dynamic library loading wasn't working (if it worked for anybody, it must have loaded system libs...)
- The Linux and macOS Slirp names were wrong. Again, if you managed to test this successfully on Linux, it must have loaded a system lib.
- As `ne2000` is now an optional feature, we should disable it by default, just like FluidSynth is not enabled by default.
- I've cleaned up the logging a bit.

Let me know if you vehemently disagree about any of this 😆 

I'm still yet to test the actual networking stuff on macOS, but now the library can be loaded at runtime by setting `ne2000 = on`. Setting `mididevice = fluidsynth` works again too.

johnnovak:
# macOS test 🍏 

## Positive test

Using my `jn/slirp-dynamic-load-2` branch that builds upon this branch (macOS & Linux dynamic loading from `$EXE_DIR/lib` is broken in this branch)

1. Downloaded the libs manually from the ext project and put them in `lib`:

image


2. Bypassed Gatekeeper on my build directory:

```
xattr -r -d com.apple.quarantine build
```

3. Using the above Ethernet test pack, Slirp gets initialised when `ne2000 = true` is set.

image

4. Test passes. Ping works on macOS out-of-the-box just fine:

image

## Negative test

1. Renamed libslirp to something else.
2. Started with the default config then set `ne2000 on` from the CLI.
3. Error in the logs as expected and `ne2000` is auto-disabled:

image

shermp:
@johnnovak I checked out your branch on Fedora, and it fails to find `libslirp.so`. I checked Debian, and it also uses `libslirp.so.0`, same as Fedora.

So my inclination is that vcpkg has got this one wrong. Looking at the other libraries in the linux dosbox-staging-ext package, they have numbered suffixes, so I think the libslirp port has an issue with it's name.

I would suggest we just rename the library to `libslirp.so.0` in the dosbox-staging-ext archive, and maybe open an issue in vcpkg to change it upstream.

johnnovak:
> @johnnovak I checked out your branch on Fedora, and it fails to find `libslirp.so`. I checked Debian, and it also uses `libslirp.so.0`, same as Fedora.
> 
> So my inclination is that vcpkg has got this one wrong. Looking at the other libraries in the linux dosbox-staging-ext package, they have numbered suffixes, so I think the libslirp port has an issue with it's name.
> 
> I would suggest we just rename the library to `libslirp.so.0` in the dosbox-staging-ext archive, and maybe open an issue in vcpkg to change it upstream.

Sure, that makes sense if we want be able to fall back to system libraries 👍🏻 I didn't consider that.

Your plan sounds good, I'll rename to lib to `ibslirp.so.0` before the packaging step in the ext project in the CI job. Then I'll raise a ticket for the guys to fix the discrepancy.

For the macOS library name, we don't care; we assume our ext libraries are always injected into the application bundle that we ship.

shermp:
Ok @johnnovak I've cherry-picked from your branch, dropped the linux so name change, and rebased onto `main`.

johnnovak:
@shermp Updated the ext libs, should be good now:

image

Re-running the failed jobs now 🤞🏻

johnnovak:
Raised a change request for the vcpkg port:
https://github.com/microsoft/vcpkg/issues/45632

johnnovak:
Raised the PVS-Studio warnings from 170 to 176, will do another test on macOS and Windows once it's green, then merge time.

johnnovak:
macOS & Windows smoke tests are good, I'm pressing the button! 😎

Issue #4306, opened 2025-05-03
Add full keyboard capture on macOS, like in VirtualBox


checktext00:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

0.83.0-alpha-2b866

### What operating system are you using?

macOS (Intel)

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Is your feature request related to a problem? Please describe.

Could you add a full keyboard capture like in VirtualBox? On macOS, the Ctrl+arrow keys activate the macOS Mission Control feature or the macOS Spaces switching feature. A lot of Windows games use Ctrl+arrow keys, so this is sometimes annoying.

See [my comment](https://github.com/joncampbell123/dosbox-x/issues/4911#issuecomment-2848534856) on a same request for DOSBox-X

### Describe the solution you'd like

full keyboard capture

### Add any other context or screenshots about the feature request here

see the linked DOSBox-X comment

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
I don't think it's possible on macOS without dirty hacks.

Grounded0:
add full keyboard capture, like in VirtualBox => Add full keyboard capture, like in VirtualBox

Fixes my OCD.

johnnovak:
Okay, this SDL2 call might do the trick in a platform-agnostic manner:

https://wiki.libsdl.org/SDL2/SDL_SetWindowKeyboardGrab

If this works, we just need to expose a `keyboard_capture` conf setting and job done.

Grounded0:
Reminds me of how I resolved "who gets Command-Tab (application switcher)" keypress with SheepShaver. Thankfully Mac OS 8.6 allows application switcher key to be remapped so it is now Option-Tab. 👍

checktext00:
> Okay, this SDL2 call might do the trick in a platform-agnostic manner:
> 
> https://wiki.libsdl.org/SDL2/SDL_SetWindowKeyboardGrab
> 
> If this works, we just need to expose a `keyboard_capture` conf setting and job done.

Thanks, I'll be sure to test a build and also report back to DOSBox-X if they can implement it

FeralChild64:
Out of curiosity, I've created a test branch for this, https://github.com/dosbox-staging/dosbox-staging/tree/fc/keyboard-grab-1, with `keyboard_capture` option in `[sdl]` section.

[...]

Issue #2527, opened 2023-05-24
'config -set fullscreen=true' doesn't work during runtime


Burrito78:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

DEV

### What Operating System are you using?

Mac OS

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When running the above command in windowed mode, nothing happens.
When running the command fullscreen, sometimes it brings the window out of fullscreen (maximised).

Same in 0.80.1 and 0.81.0-775.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. 
2.
3.

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

Burrito78:
Works in DOSBox-X.

johnnovak:
Yes.

😄 

I noticed it too but couldn't be bothered to raise a ticket... The OSD will discover a _lot_ more of these issues anyway. It's a matter of time...

Burrito78:
@GranMinigun Can you comment on this? Thanks!

Burrito78:
> Because `sdl.desktop.fullscreen` is set only on launch (`GUI_StartUp()`), not during window regeneration.

Would it be "trivial" to change and are there any downsides?

GranMinigun:
It is trivial to change. The downside is that when you touch any other setting within that section, the fullscreen flag will be read from the set option instead of being carried over from the old window. That's just how the current configuration system works.

Burrito78:
@johnnovak @GranMinigun 

So i guess best to prevent its use during runtime with a proper error message and when the GUI hits we have to make this adjustable during runtime anyway? Sounds good?

johnnovak:
> he fullscreen flag will be read from the set option instead of being carried over from the old window.

It's also trivial to set the flag on Alt-Enter, no?

weirddan455:
I looked into this.  There's a handler function when you press Alt+Enter called `SwitchFullScreen` that does exactly what it says.  The window does not get destroyed and re-created, it simply changes to fullscreen.  Config settings do not get read or modified here either.  It just toggles the fullscreen flag.

[...]

PR #4338, opened 2025-05-22
Add keyboard grabbing support

FeralChild64:
# Description

Adds a `keyboard_capture` option to call https://wiki.libsdl.org/SDL2/SDL_SetWindowKeyboardGrab - to grab more keyboard shortcuts, so they trigger DOSBox or game action instead of host OS action.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4306
Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4350

# Release notes

If `keyboard_capture` option is set, emulator now takes over more keyboard shortcuts, like ALT+TAB or some platform specific - CTRL+ESC (normally activates the start menu on Windows), CTRL+F4 (normally switches to virtual desktop #4 on KDE), CTRL+LeftArrow (shows the desktop on macOS). They are now passed to the running game or they trigger DOSBox actions.

# Manual testing

Test results with `keyboard_capture` enabled so far:
- _Linux_ (_KDE_, _Wayland_) - without keyboard grab the F12 launches _Yakuake_ (a _Quake_-style console), with keyboard grab it does not and the key is passed to the guest software; also CTRL+F4 does not switch to virtual desktop #4 anymore, but switches the mounted CD image
- _macOS_ - in fullscreen mode CTRL+LeftArrow is not taken over by the _macOS_
- _Windows_ -at least CTRL+ESC keyboard shortcut is grabbed

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ DOSBox Staging has the following library dependencies:
 | [libpng](http://www.libpng.org/pub/png/libpng.html)      | PNG encoding of screen captures                 | libpng             | 1.6.46#0        | yes :green_circle:  |
 | [Munt](https://github.com/munt/munt)                     | Roland MT-32 and CM-32L emulation               | libmt32emu         | 2.7.1#0         | yes :green_circle:  |
 | [Opus File](https://opus-codec.org/)                     | CD Audio playback for Opus-encoded audio tracks | opusfile           | 0.12+20221121#1 | **no** :red_circle: |
-| [SDL 2.0](https://github.com/libsdl-org/SDL)             | OS-agnostic API for video, audio, and eventing  | sdl2               | 2.32.4#0        | **no** :red_circle: |
+| [SDL 2.0](https://github.com/libsdl-org/SDL)             | OS-agnostic API for video, audio, and eventing  | sdl2               | 2.32.6#0        | **no** :red_circle: |
 | [SDL_net 2.0](https://github.com/libsdl-org/SDL_net)     | Network API for emulated serial and IPX         | sdl2-net           | 2.2.0#3         | yes :green_circle:  |
 | [slirp](https://gitlab.freedesktop.org/slirp)            | TCP/IP library for Ethernet emulation           | libslirp           | 4.9.0#0         | yes :green_circle:  |
 | [SpeexDSP](https://github.com/xiph/speexdsp)             | Audio resampling                                | speexdsp           | 1.2.1#1         | **no** :red_circle: |
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -1,7 +1,7 @@
 /*
  *  SPDX-License-Identifier: GPL-2.0-or-later
  *
- *  Copyright (C) 2022-2024  The DOSBox Staging Team
+ *  Copyright (C) 2022-2025  The DOSBox Staging Team
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
@@ -293,6 +293,8 @@ struct SDL_Block {
 
 	bool mute_when_inactive  = false;
 	bool pause_when_inactive = false;
+
+	bool keyboard_capture = false;
 
 	SDL_Rect draw_rect_px     = {};
 	SDL_Window* window        = nullptr;
--- a/include/sdlmain.h
+++ b/include/sdlmain.h
@@ -294,8 +294,6 @@ struct SDL_Block {
 	bool mute_when_inactive  = false;
 	bool pause_when_inactive = false;
 
-	bool keyboard_capture = false;
-
 	SDL_Rect draw_rect_px     = {};
 	SDL_Window* window        = nullptr;
 	SDL_Renderer* renderer    = nullptr;
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3400,9 +3400,11 @@ static void apply_active_settings()
 	// At least on some platforms grabbing the keyboard has to be repeated
 	// each time we regain focus
 	if (sdl.window) {
+		const auto capture_keyboard = get_sdl_section()->Get_bool(
+		        "keyboard_capture");
+
 		SDL_SetWindowKeyboardGrab(sdl.window,
-	        	                  sdl.keyboard_capture ? SDL_TRUE :
-	        	                                         SDL_FALSE);
+		                          capture_keyboard ? SDL_TRUE : SDL_FALSE);
 	}
 }
 
@@ -3473,8 +3475,6 @@ static void read_gui_config(Section* sec)
 	sdl.pause_when_inactive = section->Get_bool("pause_when_inactive");
 
 	sdl.mute_when_inactive = (!sdl.pause_when_inactive) && section->Get_bool("mute_when_inactive");
-
-	sdl.keyboard_capture = section->Get_bool("keyboard_capture");
 
 	// Assume focus on startup
 	apply_active_settings();
@@ -4504,7 +4504,7 @@ static void config_add_sdl()
 	pbool = sdl_sec->Add_bool("pause_when_inactive", on_start, false);
 	pbool->Set_help("Pause emulation when the window is inactive ('off' by default).");
 
-	pbool = sdl_sec->Add_bool("keyboard_capture", on_start, false);
+	pbool = sdl_sec->Add_bool("keyboard_capture", always, false);
 	pbool->Set_help("Takes over more host OS keyboard shortcuts ('off' by default).");
 
 	pstring = sdl_sec->Add_path("mapperfile", always, MAPPERFILE);
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1,7 +1,7 @@
 /*
  *  SPDX-License-Identifier: GPL-2.0-or-later
  *
- *  Copyright (C) 2020-2024  The DOSBox Staging Team
+ *  Copyright (C) 2020-2025  The DOSBox Staging Team
  *  Copyright (C) 2002-2021  The DOSBox Team
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -3396,6 +3396,14 @@ static void apply_active_settings()
 	if (sdl.mute_when_inactive && !MIXER_IsManuallyMuted()) {
 		MIXER_Unmute();
 	}
+
+	// At least on some platforms grabbing the keyboard has to be repeated
+	// each time we regain focus
+	if (sdl.window) {
+		SDL_SetWindowKeyboardGrab(sdl.window,
+	        	                  sdl.keyboard_capture ? SDL_TRUE :
+	        	                                         SDL_FALSE);
+	}
 }
 
 static void ApplyInactiveSettings()
@@ -3465,6 +3473,8 @@ static void read_gui_config(Section* sec)
 	sdl.pause_when_inactive = section->Get_bool("pause_when_inactive");
 
 	sdl.mute_when_inactive = (!sdl.pause_when_inactive) && section->Get_bool("mute_when_inactive");
+
+	sdl.keyboard_capture = section->Get_bool("keyboard_capture");
 
 	// Assume focus on startup
 	apply_active_settings();
@@ -4494,6 +4504,9 @@ static void config_add_sdl()
 	pbool = sdl_sec->Add_bool("pause_when_inactive", on_start, false);
 	pbool->Set_help("Pause emulation when the window is inactive ('off' by default).");
 
+	pbool = sdl_sec->Add_bool("keyboard_capture", on_start, false);
+	pbool->Set_help("Takes over more host OS keyboard shortcuts ('off' by default).");
+
 	pstring = sdl_sec->Add_path("mapperfile", always, MAPPERFILE);
 	pstring->Set_help(
 	        "Path to the mapper file ('mapper-sdl2-XYZ.map' by default, where XYZ is the\n"
@@ -4938,6 +4951,10 @@ int sdl_main(int argc, char* argv[])
 		// Seamless mouse integration feels more 'seamless' if mouse
 		// clicks on out-of-focus window are passed to the guest
 		SDL_SetHint(SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH, "1");
+
+		// We have a keyboard shortcut to exit the fullscreen mode,
+		// so we don't necessary need the ALT+TAB shortcut
+		SDL_SetHint(SDL_HINT_ALLOW_ALT_TAB_WHILE_GRABBED, "0");
 
 		if (const auto err = check_kmsdrm_setting(); err != 0) {
 			return err;
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4505,7 +4505,11 @@ static void config_add_sdl()
 	pbool->Set_help("Pause emulation when the window is inactive ('off' by default).");
 
 	pbool = sdl_sec->Add_bool("keyboard_capture", always, false);
-	pbool->Set_help("Takes over more host OS keyboard shortcuts ('off' by default).");
+	pbool->Set_help(
+	        "Capture system keyboard shortcuts ('off' by default).\n"
+	        "When enabled, most system shortcuts such as Alt+Tab are captured and sent to\n"
+	        "DOSBox Staging. This is useful for Windows 3.1x and some DOS programs with\n"
+	        "unchangeable keyboard shortcuts that conflict with system shortcuts.");
 
 	pstring = sdl_sec->Add_path("mapperfile", always, MAPPERFILE);
 	pstring->Set_help(
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -1,0 +1,5 @@
+{
+  "overlay-ports": [
+    "./vcpkg-ports"
+  ]
+}
--- a/vcpkg-ports/sdl2/alsa-dep-fix.patch
+++ b/vcpkg-ports/sdl2/alsa-dep-fix.patch
@@ -1,0 +1,14 @@
+diff --git a/SDL2Config.cmake.in b/SDL2Config.cmake.in
+index cc8bcf26d..ead829767 100644
+--- a/SDL2Config.cmake.in
++++ b/SDL2Config.cmake.in
+@@ -35,7 +35,8 @@ include("${CMAKE_CURRENT_LIST_DIR}/sdlfind.cmake")
+ 
+ set(SDL_ALSA @SDL_ALSA@)
+ set(SDL_ALSA_SHARED @SDL_ALSA_SHARED@)
+-if(SDL_ALSA AND NOT SDL_ALSA_SHARED AND TARGET SDL2::SDL2-static)
++if(SDL_ALSA)
++  set(CMAKE_REQUIRE_FIND_PACKAGE_ALSA 1)
+   sdlFindALSA()
+ endif()
+ unset(SDL_ALSA)
--- a/vcpkg-ports/sdl2/cxx-linkage-pkgconfig.diff
+++ b/vcpkg-ports/sdl2/cxx-linkage-pkgconfig.diff
@@ -1,0 +1,24 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 2a91824..a8e9de4 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -3162,6 +3162,19 @@ set(SDL_STATIC_LIBS ${SDL_LIBS} ${EXTRA_LDFLAGS} ${_EXTRA_LIBS})
+ list(REMOVE_DUPLICATES SDL_STATIC_LIBS)
+ listtostr(SDL_STATIC_LIBS _SDL_STATIC_LIBS)
+ set(SDL_STATIC_LIBS ${_SDL_STATIC_LIBS})
++if("${SOURCE_FILES};" MATCHES "[.]cpp;")
++  set(FAKE_CXX_LINKAGE "")
++  foreach(lib IN LISTS CMAKE_CXX_IMPLICIT_LINK_LIBRARIES)
++      if(lib IN_LIST CMAKE_C_IMPLICIT_LINK_LIBRARIES)
++          continue()
++      elseif(EXISTS "${lib}")
++          string(APPEND FAKE_CXX_LINKAGE " ${CMAKE_LINK_LIBRARY_FILE_FLAG}${lib}")
++      else()
++          string(APPEND FAKE_CXX_LINKAGE " ${CMAKE_LINK_LIBRARY_FLAG}${lib}")
++      endif()
++  endforeach()
++  string(APPEND SDL_STATIC_LIBS "${FAKE_CXX_LINKAGE}")
++endif()
+ listtostr(SDL_LIBS _SDL_LIBS)
+ set(SDL_LIBS ${_SDL_LIBS})
+ listtostr(SDL_CFLAGS _SDL_CFLAGS "")
--- a/vcpkg-ports/sdl2/deps.patch
+++ b/vcpkg-ports/sdl2/deps.patch
@@ -1,0 +1,13 @@
+diff --git a/cmake/sdlchecks.cmake b/cmake/sdlchecks.cmake
+index 65a98efbe..2f99f28f1 100644
+--- a/cmake/sdlchecks.cmake
++++ b/cmake/sdlchecks.cmake
+@@ -352,7 +352,7 @@ endmacro()
+ # - HAVE_SDL_LOADSO opt
+ macro(CheckLibSampleRate)
+   if(SDL_LIBSAMPLERATE)
+-    find_package(SampleRate QUIET)
++    find_package(SampleRate CONFIG REQUIRED)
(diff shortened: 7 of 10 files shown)


es20490446e:
Tested to work 👍

weirddan455:
This immediately crashes for me with an assert fail:

```
dosbox: /home/daniel/code/dosbox-staging/src/gui/sdlmain.cpp:3401: void apply_active_settings(): Assertion `sdl.window' failed
```

FeralChild64:
OK, forgot to test it with the debug build - setting to DRAFT for now, I need to adapt the code and retest on all the OSes.

FeralChild64:
Ok, fixed, retested. I've now found out, that on Windows the CTRL+ESC is affected by the new settings, it is now grabbed by DOSBox.

weirddan455:
I can confirm this does "overwrite" KWin shortcuts but I see some bugs.

> The most critical shortcuts, like ALT+TAB, should continue to work as before, or be replaced by SDL with own implementation.

This breaks Alt+Tab for me while in windowed mode. I don't believe SDL is capable of replacing this. It's the window manager's job. Tested on both X11 and Wayland KDE.

A couple other bugs that are only reproducible in X11. Note that this is still the default for SDL even if you're using a Wayland compositor. You must set `SDL_VIDEODRIVER=wayland` for these bugs to not occur:

- The window becomes unfocused as soon as the mouse cursor leaves the window border. This is noticeable visually at the bottom task bar where the blue "focused" highlight is removed. It then refocuses as soon as the mouse cursor re-enters the window.
- `Ctrl+F4` shortcut no longer works unless the mouse cursor is inside the window. This is true even if there is no conflicting KWin shortcut. This is a regression from `main`.

rderooy:
If you are going to properly capture the keyboard, it should include ALT-TAB. ALT-TAB is used by Win 3.1 and Win9x, so is necessary. Other emulators capture that key combination, as they should. It may be inconvenient at times, but the same could be argued for any other host keyboard combination.

FeralChild64:
@weirddan455 Regarding  the ALT+TAB - I must have misread the documentation, it says _SDL will continue to handle Alt+Tab when the window is full-screen to ensure the user is not trapped in your application_ - so this is expected behavior.

And I can't reproduce the remaining ones, neither with KDE/X11 nor with KDE/Wayland, setting` SDL_VIDEODRIVER=wayland` does not change this. I am running kwin version 6.3.5 and SDL version 2.32.8.

@rderooy You're right - I'll prepare a version which should pass ALT+TAB to the guest side, always.

FeralChild64:
OK, pushed the updated code - if `keyboard_capture = on`, ALT+TAB is now passed to the guest side, always.
For now tested on Linux (KDE, Wayland) only - I'll do more testing, but not today, it's getting late.

FeralChild64:
Since at least @weirddan455 is experiencing unwanted behavior with this setting, I've set it to `off` by default.

weirddan455:
> Ctrl+F4 shortcut no longer works unless the mouse cursor is inside the window. This is true even if there is no conflicting KWin shortcut. This is a regression from main.

I tested again and this one is only reproducible on XWayland. KDE on Wayland with `SDL_VIDEODRIVER=x11`. I have Plasma 5.27.5 and SDL from vcpkg 2.32.4.

> OK, pushed the updated code - if keyboard_capture = on, ALT+TAB is now passed to the guest side, always.

And yeah, now ALT+TAB is broken everywhere :smile: 

> Since at least @weirddan455 is experiencing unwanted behavior with this setting, I've set it to off by default.

Just the ALT+TAB behavior is enough to make me want this off by default. I get that it's useful for Win 3.1 stuff but the normal use case for DOSBox is games. I like to Alt+Tab out of my fullscreen games sometimes and no other game I play hijacks Alt+Tab.

FeralChild64:
True, the behavior is a little bit... controversial. But we can only do what SDL has an API for.

I'll wait a day or two before merging to give others a chance to test.

johnnovak:
@FeralChild64 I've just tested this on macOS, and it doesn't do jackshit, to put it succinctly 😄 

I agree that a full keyboard capture feature common in VMs would be a welcome addition. I appreciate your effort of exploring the SDL-provided functionality in this area, but it's pretty clear that the pure SDL-based solution is woefully inadequate for our needs.

We're trying hard not to introduce any platform-specific features, and currently, this only seems to work relatively well on Linux. I haven't tested on Windows yet, but given it does virtually nothing on macOS, that disqualifies this PR from being merged in its current form.

Ironically, this work started in response to the below ticket which asked for adding full input capture on macOS 😆 
- https://github.com/dosbox-staging/dosbox-staging/issues/4306

> But we can only do what SDL has an API for.

Well, we can do anything 😎 It seems the only way to solve this properly is to write our own OS-specific keyboard input capture mechanism, then we can feed those events into the event loop with [SDL_PushEvent](https://wiki.libsdl.org/SDL2/SDL_PushEvent) (from memory, that's one standard way to bypass such SDL limitations). There are existing open source solutions that show how to do this, e.g., check out the platform-specific keyboard input capture layer in this Python library:

https://github.com/SerpentAI/sneakysnek/tree/master/sneakysnek/recorders

Surely this is a solved problem in other emulators and VMs, so it would be worthwhile to check out how PCem, 86box, VirtualBox, etc. do this and simply copy their battle-tested approach.

The integration of our custom solution with SDL's event handling might be tricky, but these are the cards we're dealt. Alternatively, we might submit patches to SDL if this is important to us because the SDL folks seem to have a low-interest in turning full keyboard capture into a robust cross-platform feature (somewhat understandably as games are their focus, not VMs). But that might also mean we'd need to migrate to SDL3 sooner (if they don't accept substantial SDL2 patches).

So, I'm suggesting to explore more robust, custom solutions (if you feel the inclination). I don't think this PR is ready to be merged in its current form.

johnnovak (changes requested):
Not ready for prime time, as per my detailed comment.

FeralChild64:
@johnnovak It works on Windows according to my test, looks like only the _Macintosh_ is problematic. The topic seems to be tricky, it's probably best to be investigated by someone more familiar with the _macOS_ than me.

I see SDL is using a private, undocumented API on this - namely the `CGSSetGlobalHotKeyOperatingMode` call. But it's guarded by `#ifdef SDL_MAC_NO_SANDBOX` and I don't know whether it is defined in the SDL build we are using or not. And I'm definitely not the person who is going to investigate this, try to convince the _HomeBrew_ / _MacPorts_ / _vcpkg_ / whoever to change the build options, etc. - tried once to convince the _HomeBrew_ team not to ship the _Meson_ release which was known to be broken on the _Mac_, never more.

I have also found some information about `CGDisplayCapture(CGMainDisplayID())` / `CGDisplayRelease(CGMainDisplayID())` calls, but they probably only work in fullscreen.

I intend to experiment a little with the _Mac_ - but I'll definitely need some help to assess the possible unwanted side effects, I'm too unfamiliar with this environment. Setting this back to DRAFT for now.

johnnovak:
@FeralChild64 That's fine, we can put it on the backburner. I might have a look at some point myself and experiment with it, but I have far more important things in the pipeline and I'm time-poor 😄 Also, it wouldn't be not terrible to defer this to the full mapper rework in the future, post-SDL3-migration.

FeralChild64:
@FeralChild64 It seems the SDL build which is being taken for the compilation indeed omits the `CGSSetGlobalHotKeyOperatingMode` call - once I've added it, the `CTRL+Arrow` is no longer taken over by the macOS.

I still have build system issues to solve:
 - _Meson_ does not link the binary with the `CoreGraphics`, which leads to linking failure due to missing symbols
 - _CMake_ does not define the `C_COREGRAPHICS` in the code, which leads to the extra _macOS_ support I've just added not being compiled in
 
 [edit] The new GUI and mapper might not arrive even in 5 years, so I prefer not to wait for it. And the keyboard grabbing code in SDL3 is the same as in SDL2.

FeralChild64:
_Meson_ build fixed. Now the _CMake_...

FeralChild64:
Build systems fully adapted now. @johnnovak, please retest now.

weirddan455:
Forgive me for going off topic but since SDL 3 was mentioned, I plan on re-opening the migration PR shortly after the next Staging release. We're already 90% of the way to ditching Meson completely which I need to happen but I also don't want to overload the next release which has had quite a lot of changes already.

FeralChild64:
@weirddan455 Fully agree - such huge&risky changes are best to be done early in the release development cycle.

es20490446e:
I'm time rich.

The trick: end what has been started now.

Otherwise all the invested time is wasted.

johnnovak:
_(Sorry, didn't want to close  😄)_

Alright, keyboard capture works very well on macOS using the latest build 🥳  Great job persevering with this @FeralChild64 🎉

With `capture_keyboard` enabled, system-level shortcuts such as F11, Ctrl-UpArrow, Ctrl-DownArrow, Cmd+Space are completely disabled 🍏  (but unfortunately Cmd+Tab as well 😐).

The situation is the same on Windows, apparently all system-level shorcuts are disabled except for Ctrl+Alt+Del 🍏 (but Alt+Tab too, unfortunately 😐).

So this changes things substantially. Given that cross-platform support is now good, this can go in. I quite like this because using DOS Navigator shortcuts conflict a lot with the default macOS shortcuts, so I had to disable those wholesale for all apps.

I think we can improve the implementation a bit before merging:

1. I see no reason for not allowing changing `keyboard_capture` at runtime. As far as I can tell, SDL supports toggling it at any time just like capturing the mouse.

2. We don't need any custom code to enable the feature, we just need a vcpkg port overlay and add `-DSDL_MAC_NO_SANDBOX=ON` [here](https://github.com/microsoft/vcpkg/blob/master/ports/sdl2/portfile.cmake#L46-L53) (getting the vcpkg port maintainer to add a feature toggle for this another matter, let's treat that separately). @weirddan455 has figured out how to add vcpkg port overlays cleanly to a CMake project, just check out the [ext repo](https://github.com/dosbox-staging/dosbox-staging-ext/tree/main) and @weirddan455's relevant commits. Probably worthwhile to read about [port overlays](https://learn.microsoft.com/en-us/vcpkg/concepts/overlay-ports) in the manual too to familiarise yourself with this handy feature. ***

3. I'm with @weirddan455, losing Alt-Tabbing is super inconvenient and capturing Alt-Tab as well is a very niche use case (Win3.1 games that use multiple windows, or using Win3.1 in non-gaming contexts). I've spent 15 mins reading the SDL code and probably we'd need SDL-level support to make this happen; we can't make an exception for Alt-Tab easily on our side. But, as a stopgap measure, I think we can substantially improve the situation to make this feature more usable:

  - Option 1: Simply only make keyboard capturing active when the mouse is captured. So if you want to Alt-Tab, you can quickly middle-click to uncapture the mouse _and_ the keyboard first. I think this is the most intuitive behaviour and what 99.999% people want (I don't dare to write 100%, I'm sure someone will complain about this... 😆). The only minor complication is the `mouse_capture = nomouse` case, but in that case the middle click and Ctrl+F10 could still be used to uncapture the keyboard only. So entirely doable and not hard to do, just the messages and setting descriptions need some further thought & updating.
  
  - Option 2: Introduce `sync_mouse_and_keyboard_capture` which does basically the above thing, but it's opt-in. IMO, that's a stupid setting name, and unnecessary complication, and we want to avoid config-bloat. Capturing "all input" (if keyboard capture is enabled as well) is a good simple model, I don't see a good reason for keyboard and mouse capturing to work separately (this actually confused me a lot during testing; I just want the middle-button to release the keyboard as well... and actually Alt-Tab to release the mouse too, but that's a topic for another ticket 😏).
  
Thoughts, guys? I prefer option 1 by a large margin.

---

*** `SDL_MAC_NO_SANDBOX ` is disabled by default because it's a no-go if you want your SDL-based game to be published in the App Store (understandable, intercepting system-level keyboard events is kind of a security issue). But the good news is, we have no aspirations to get Staging into the App Store, and even if we wanted to, we couldn't because of Apple's stance on dynamic JIT recompliers (although [the situation might change](https://9to5mac.com/2025/05/01/jit-enabler-lands-on-app-store-likely-unlocking-wii-and-switch-emulation-on-ios/)).

> On OS X have you tried building SDL2 with SDL_MAC_NO_SANDBOX defined? Doing so enables better mouse grabbing code, at the cost of being unable to use the library in a program published in the Mac App Store apparently.
([source](https://github.com/libsdl-org/SDL/issues/1116))

es20490446e:
Do you think of any relevant case where Alt+Tab is used within a Dos application?

rderooy:
Apparently, according to Wikipedia, Alt-TAB was introduced with Windows 1.0.
https://en.wikipedia.org/wiki/Alt-Tab

I don't know what its early competitors like VisiOn, TopView, GEM, DESQview, DeskMate and GEOS used. I did experiment with some back in the day, but don't recall any details ;-)

FeralChild64:
@johnnovak 

> 1. I see no reason for not allowing changing keyboard_capture at runtime. As far as I can tell, SDL supports toggling it at any time just like capturing the mouse.

SDL might, but `sdl_main.cpp` does not. I see `read_gui_config` is executed just once, on purpose - so it's not enough to just change the `on_start` parameter to `always`, some more rework would be needed.

Even if I read the setting every time, directly in `read_config`, it's not clear to me when I should apply the settings, and when not to do it - I don't know this code too well, and I'm afraid of the risks.

There are more settings there which should probably be changeable at runtime, like `mute_when_inactive` and `pause_when_inactive` - but this, IMHO, should be a separate PR.

> 3. (...) losing Alt-Tabbing is super inconvenient (...) I've spent 15 mins reading the SDL code and probably we'd need SDL-level support to make this happen

Oh, it's not that complicated at all - this is what they do to fake the ALT+TAB when fullscreen:

```C
    if ((... a lot of conditions here ...) &&
        SDL_GetHintBoolean(SDL_HINT_ALLOW_ALT_TAB_WHILE_GRABBED, true)) {
        /* We will temporarily forfeit our grab by minimizing our window,
           allowing the user to escape the application */
        SDL_MinimizeWindow(keyboard->focus);
    }
```

They don't tell the host OS _hey, do the application switching now_ -  they simply minimize the window; at this point the host OS has to decide what window to switch to, but the user is pressing ALT+TAB...

We could implement it by ourselves too - a side effect is going to be that when pressing ALT+TAB in windowed mode the emulator will get minimized. But I would prefer to experiment with it in a separate PR, especially that I'm not familiar with the mapper and keyboard shortcuts handling at all.

> Option 1: Simply only make keyboard capturing active when the mouse is captured

And what about games like DOOM? Probably most players do not use here the mouse at all, DOOM uses arrows for moving and CTRL to fire - and CTLR+LeftArrow on MacOS...

And what about seamless mouse integration? Should we capture the keyboard in such case or not? And what about the fullscreen mode? IMHO there is far too many possible use cases to try to be "smart".

> Option 2: Introduce `sync_mouse_and_keyboard_capture` which does basically the above thing, but it's opt-in.

This looks really scary and overcomplicated to me.

FeralChild64:
> 2. We don't need any custom code to enable the feature, we just need a vcpkg port overlay and

Well, I'm neither familiar with _CMake_ nor with _vcpkg_, and from just briefly looking into the links I still have no idea what precisely to do. I'll have a deeper look when I have more time.

johnnovak:
Ok, I forgot that the `[sdl]` section is a horrible mess. Okay, maybe leave setting this at runtime alone for now (we should really rewrite that whole damn `sdlmain.cpp` cleanly from the ground up; it a waste of time to try to salvage it, it's just garbage...)

> Oh, it's not that complicated at all - this is what they do to fake the ALT+TAB when fullscreen:

> We could implement it by ourselves too - a side effect is going to be that when pressing ALT+TAB in windowed mode the emulator will get minimized. But I would prefer to experiment with it in a separate PR, especially that I'm not familiar with the mapper and keyboard shortcuts handling at all.

Okay, let's experiment with that in a follow-up PR. But if it involves force-minimising the window when Alt-Tabbing in windowed mode, that's not that great.

> And what about games like DOOM? Probably most players do not use here the mouse at all, DOOM uses arrows for moving and CTRL to fire - and CTLR+LeftArrow on MacOS...
> 
> And what about seamless mouse integration? Should we capture the keyboard in such case or not? And what about the fullscreen mode? IMHO there is far too many possible use cases to try to be "smart".

Yeah, I was just trying to think of something that would work, but agreed, it wouldn't be general enough (it would work for simple mouse-driven game use-cases, though). The real solution would be to just leave Alt+Tab (and Cmd+Tab) alone by default, and have an extra option to capture that too. WinUAE basically takes over all system shortcuts (well, most), but leaves Alt+Tab always alone in both windowed and fullscreen mode, which is what pretty much everybody wants.

We could have `keyboard_capture = limited` and `full` or something, where `full` includes Alt+Tab and `limited` doesn't. Just ideas. But out of scope for this PR then.

> Well, I'm neither familiar with CMake nor with vcpkg, and from just briefly looking into the links I still have no idea what precisely to do. I'll have a deeper look when I have more time

Overlay ports are basically us taking a copy of the port files from the vcpkg repo and applying whatever modifications we want to them. It's a simple concept. So we just modify the portfile to add that define and that's it. I'm sure you can figure it out in 10 minutes by examinig the ext repo (you don't even need to read the doco), but I can help with that next week sometime. I suggest to keep this open until then because there's no point in introducing extra code for this when it's a simple SDL compile time option.

Yeah, and don't worry about Meson, that will go away anyway by the end of the year (as in deleted from the repo).

FeralChild64:
> Overlay ports are basically us taking a copy of the port files from the vcpkg repo and applying whatever modifications we want to them

This doesn't look good, if you ask me. I've found this tutorial https://learn.microsoft.com/en-us/vcpkg/consume/install-locally-modified-package?pivots=shell-powershell - but it looks like a maintenance nightmare to me.

Is it possible that `vcpkg` changes something upstream, our patch stops working (or our cloned+modified recipe stops being compatible with the rest of the system), and from now on no one is going to be able to compile the source we ship anymore? And what if we want to bisect some problem?

>  I can help with that next week sometime. I suggest to keep this open until then

OK, I'll wait.

johnnovak:
> Is it possible that `vcpkg` changes something upstream, our patch stops working (or our cloned+modified recipe stops being compatible with the rest of the system), and from now on no one is going to be able to compile the source we ship anymore? And what if we want to bisect some problem?

Nah, man. We're in complete control, what you describe is not possible, it cannot break. I don't think you quite understand it how it works.

It's a one-line patch, literally. Upgrading to the next version is a 5 minute job, copy the port files from the next SDL port version, then apply the same one line patch and commit the overlay port into our repo.

Bisecting 100% works, of course, it cannot possibly break (only if the SDL repo gets completely deleted, but then all vpkg SDL ports would break too...)

Yeah, I'm planning to ask for support for this upstream, then we can remove the overlay port stuff, but until then this is the best solution. I'll just do it next week and then you can drop your second commit and cherry-pick my change into this branch.

es20490446e:
My sensation is that Alt+Tab shall be always for the host OS by default, except if stated otherwise in the configuration.

Maybe don't do anything more complicated than that, and see if it leads to issues.

Once seeing the emerging issue in context it would be easier to decide what would be the appropriate action to take.

FeralChild64:
> My sensation is that Alt+Tab shall be always for the host OS by default

We don't know any way to grab everything except ALT+TAB - we can either grab all or none :(

es20490446e:
Missing `Alt+Tab` is just too crucial.

Having two modes, grabbed and ungrabbed may procreate complexity a lot.

What about just changing those `F#`shortcuts to something that is quite hard to have conflicts, yet easier to remember?

Using `AltGr`, as VirtualBox does to avoid conflicts, may be handy here.

Let me make a couple of tests and I will let you know.

johnnovak:
> What about just changing those `F#`shortcuts to something that is quite hard to have conflicts, yet easier to remember?

We're not changing the default shortcuts at this point @es20490446e, as I've told you before. Customise them if you don't like them and/or if they conflict with your specific setup.

es20490446e:
After a few tests I'm convinced that the only thing that is needed is to simplify the shortcuts, to only having these:

- scrshot "printscreen + mod2"
- fullscr "enter + mod2"
- restart "home + mod2"
- pause "pause + mod2"
- cycledown "down + mod2"
- cycleup "up + mod2"
- hand_swapimg "tab + mod1"

Everything else is just a very specialized functionality, or better done using an external application.

[mapper.map.txt](https://github.com/user-attachments/files/20647458/mapper.map.txt)

johnnovak:
> After a few tests I'm convinced that the only thing that is needed is to simplify the shortcuts, to only having these:
> 
>     * scrshot "printscreen + mod2"
> 
>     * fullscr "enter + mod2"
> 
>     * restart "home + mod2"
> 
>     * pause "pause + mod2"
> 
>     * cycledown "down + mod2"
> 
>     * cycleup "up + mod2"
> 
>     * hand_swapimg "tab + mod1"
> 
> 
> Everything else is just a very specialized functionality, or better done using an external application.
> 
> [mapper.map.txt](https://github.com/user-attachments/files/20647458/mapper.map.txt)

Out of scope for now, will happen with the mapper rework...

johnnovak:
@FeralChild64 

Alright, I'm done, you just need to drop your macOS workaround commit and cherry-pick the commits from [this branch](https://github.com/dosbox-staging/dosbox-staging/compare/main...refs/heads/jn/keyboard-grab-port-overlay) and this is good to go.

This is what I've added:

- Use vcpkg overlay to enable keyboard capture on macOS
- Allow `keyboard_capture` to be changed at runtime (it was easy to do it cleanly and safely)
- Improve the setting's description

FeralChild64:
Done.

I would not implement retrieving the config value each time we gain focus, but I guess I spent too much time developing embedded software :)

johnnovak:
> I would not implement retrieving the config value each time we gain focus, but I guess I spent too much time developing embedded software :)

_Maybe_ it matters it you run it on a Z80 or a 6502, yeah 😄 

It's utterly pointless to "optimise" such paths that get executed sporadically, and getting a bool config value is not slow at all. "Slow" is realtive to call frequency and actual time spent... Execute that once per pixel => slow. Once per windowing event => doesn't even register in your profiler 😎

es20490446e:
Nice 👍

weirddan455:
> Ok, I forgot that the [sdl] section is a horrible mess. Okay, maybe leave setting this at runtime alone for now (we should really rewrite that whole damn sdlmain.cpp cleanly from the ground up; it a waste of time to try to salvage it, it's just garbage...)

This is more of a weakness of our config system as a whole. Whenever a setting gets changed at runtime, it always destroys and re-inits the entire module (`[sdl]` in this case which has been the most problematic because that means re-creating the window at a bare minimum).

There's been a couple tickets on this such as #2527 and #3146

In this case, all that needs to happen when `keyboard_capture` is changed is to change a single boolean variable but instead we re-create the window and do a whole bunch of other crap we don't need to do including running `MAPPER_BindKeys()` because this mapper is more coupled to `sdlmain.cpp` than I would like.

> I would not implement retrieving the config value each time we gain focus, but I guess I spent too much time developing embedded software :)

It does a linear string compare search. Out of curiosity I benchmarked it and it takes less than 10 microseconds in a debug build. It does happen twice for some reason but the 2nd one drops down to 1 or 2 microseconds (probably due to being hot in the CPU cache?).

> It's utterly pointless to "optimise" such paths that get executed sporadically, and getting a bool config value is not slow at all. "Slow" is realtive to call frequency and actual time spent... Execute that once per pixel => slow. Once per windowing event => doesn't even register in your profiler 😎

Agreed. This is rarely executed. Only happens when the window was previously unfocused and the user clicks into it. And we're measuring in microseconds.

We could cache it by reading from the config inside `GFX_RegenerateWindow()` which is run every time an `[sdl]` config option is changed but I really wouldn't bother based on the times I just measured.

The more problematic thing I wanted to point out is that we're flickering the window for a bunch of settings we don't need to be.

johnnovak:
Yeah, good thing you measured it, now we know the actual impact 🚀 I'm not saying it wouldn't be better to store it as a bool flag, of course; it's just too hard now because as you've pointed out, the whole config section gets reinited if any "changeable" setting in it changes. So this is a good pragmatic solution until we improve the config handling.

Improving the config system will take some substantial rework, so I'm deferring that until we absolutely can't ignore it. The OSD work will necessitate a significant rework in this area, so my idea is to let that "drive" the changes in the future. Right know we have some idea what needs improving (e.g., per-setting reinit vs per-section, at least conceptually), but the OSD features will put the most "stress" on the current config system to expose its limitations and shortcomings.

es20490446e:
Try the smallest refactoring.

If it sticks refactor more.

If you defer it, it will never get done.

johnnovak:
> Try the smallest refactoring.
> 
> If it sticks refactor more.
> 
> If you defer it, it will never get done.

I've been doing this for 20+ years @es20490446e 😏  I know what I'm doing and where I want to arrive, eventually 😏 Generic "advice" is not very helpful.

checktext00:
How does this work? There's no mention of any `keyboard_capture` in the default dosbox-staging.conf for DOSBox Staging 0.82.2 or in the [release notes](https://www.dosbox-staging.org/releases/release-notes/0.82.2/)

johnnovak:
> How does this work? There's no mention of any `keyboard_capture` in the default dosbox-staging.conf for DOSBox Staging 0.82.2 or in the [release notes](https://www.dosbox-staging.org/releases/release-notes/0.82.2/)

Yeah, because it's for 0.83.0 😄 It will be mentioned in the 0.83.0 release notes. But you can use the feature today if you download the latest dev build.

PR #4344, opened 2025-05-24
Add ManyMouse support to CMake scripts

FeralChild64:
# Description

Updated _CMake_ scripts for _ManyMouse_ support. Should not interfere with the work-in-progress spring cleanup.

_macOS_ remark: Imported some _CMake_ support scripts from the _OGRE_ project - to check whether the OS provides appropriate interfaces (carryover behavior from from old _Meson_ scripts).

_Linux_ remark: The _XInput_ interface is disabled by default:
- distros use _evdev_ since a long time
- _ManyMouse_ driver for _evdev_ works better, thus it is already the default one with 0.82
- more and more distros are now dropping the _X11_ 

## Related issues

# Manual testing

- grab the tarball from CI or compile the project using CMake scripts (like `cmake --preset=release-linux`, then `cmake --build --preset=release-linux`)
- if on Windows, set `mouse_raw_input = false`
- start DOSBox, type `mousectl -all` - it should print out all the USB mice found
- execute `mousectl dos -map` command, press mouse button, check with any software that the mapped mouse works; it should not react on non-mapped mice

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,15 @@ if(NOT HAS_UNISTD)
   )
 endif()
 
+# File-descriptor manipulation routines, such as FD_ZERO, are used
+# by Enet, slirp, and ManyMouse's X11 interface. Unfortunately these
+# routines aren't universally available.
+if(DOSBOX_PLATFORM_WINDOWS)
+  check_symbol_exists(FD_ZERO  "winsock2.h"    HAVE_FD_ZERO)
+else()
+  check_symbol_exists(FD_ZERO  "sys/select.h"  HAVE_FD_ZERO)
+endif()
+
 ### config.h shenanigans
 
 set(TEST_CODE_BUILTIN_AVAILABLE "
@@ -151,9 +160,43 @@ if(OPT_DEBUG)
   endif()
 endif()
 
+# ManyMouse - optional XInput support
+option(OPT_MANYMOUSE
+       "Use ManyMouse library for single-computer multiplayer gaming in The Settlers I/II"
+       ON
+)
+
+# ManyMouse - optional XInput support
+option(OPT_XINPUT
+       "Let ManyMouse use the X Input protocol"
+       OFF
+)
+
 # ManyMouse
-set(C_MANYMOUSE OFF)    # TODO
-set(SUPPORT_XINPUT2 OFF)    # TODO
+if (OPT_MANYMOUSE)
+  if (NOT HAVE_FD_ZERO)
+    message(WARNING "ManyMouse requires FD_ZERO support")
+  elseif(DOSBOX_PLATFORM_MACOS)
+    include (${CMAKE_CURRENT_SOURCE_DIR}/src/libs/manymouse/cmake/FindIOKit.cmake)
+    if (NOT ${IOKit_FOUND})
+      message(WARNING "ManyMouse requires IOKit")
+    else()
+      set(C_MANYMOUSE ON)
+    endif()
+  else()
+    set(C_MANYMOUSE ON)
+  endif()
+
+  if (C_MANYMOUSE AND OPT_XINPUT)
+    check_include_file("X11/extensions/XInput2.h" HAVE_XINPUT)
+    include (${CMAKE_ROOT}/Modules/FindX11.cmake)
+    if ((NOT ${X11_Xinput_FOUND}) OR (NOT HAVE_XINPUT))
+      message(WARNING "XInput not found.")
+    else()
+      set(SUPPORT_XINPUT2 ON)
+    endif()
+  endif()
+endif()
 
 # macOS
 set(C_COREAUDIO      "${DOSBOX_PLATFORM_MACOS}")
--- a/.github/workflows/pvs-studio.yml
+++ b/.github/workflows/pvs-studio.yml
@@ -121,7 +121,7 @@ jobs:
 
       - name: Summarize report
         env:
-          MAX_BUGS: 176
+          MAX_BUGS: 177
         run: |
           echo "Full report is included in build Artifacts"
           echo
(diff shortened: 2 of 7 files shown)


scripts/cmake/FindIOKit.cmake
```diff
@@ -14,7 +14,7 @@
 #  IOKit_INCLUDE_DIRS - the IOKit include directories 
 #  IOKit_LIBRARIES - link these to use IOKit
 
-include(FindPkgMacros)
+include(${CMAKE_CURRENT_SOURCE_DIR}/scripts/cmake/FindPkgMacros.cmake)
```
johnnovak, line 17: These are not scripts, so let's not put them into `scripts`.

A common approach is to create ` cmake` folder in the project root for global CMake modules.

Example:
https://github.com/stenzek/duckstation

Or when the modules are not global, just put them in subfolder of the parts of the codebase they apply to:

Example:
https://github.com/pabloariasal/modern-cmake-sample/tree/master/libjsonutils/cmake

So in this case, I think the best place for them would be `src/libs/manymouse/cmake`.

CMakeLists.txt
```diff
@@ -152,8 +161,26 @@ if(OPT_DEBUG)
...
+       OFF
+)
+check_include_file("X11/extensions/XInput2.h" HAS_XINPUT)
+if (C_MANYMOUSE AND HAS_XINPUT AND OPT_XINPUT)
+  include (${CMAKE_ROOT}/Modules/FindX11.cmake)
+  if (NOT ${X11_Xinput_FOUND})
+    message(FATAL_ERROR "XInput not found.")
+  endif()
+  set(SUPPORT_XINPUT2 ON)
+endif()
```
johnnovak, line 183: All this stuff is only related to the specifics of building ManyMouse support in `src/libs/manymouse`, right? Therefore, it would be better to move most of it into the `CMakeLists.txt` there. 

All these decisions result in failing the build or setting `SUPPORT_XINPUT2 `, but that is only checked in `src/libs/manymouse/x11_xinput2.c`. So yeah, you don't need `SUPPORT_XINPUT2` in the global scope either, just when building ManyMouse.

CMakeLists.txt
```diff
@@ -238,6 +265,7 @@ endif()
 target_link_libraries(dosbox PRIVATE
   $<TARGET_NAME_IF_EXISTS:SDL2::SDL2main>
   $<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
+  $<IF:$<BOOL:${SUPPORT_XINPUT2}>,${X11_Xinput_LIB},>
```
johnnovak, line 268: Yeah, so no need for `SUPPORT_XINPUT2` to be defined here. You can do `target_link_libraries` in `src/libs/manymouse/CMakeLists.txt`, many other submodules in the project do it similarly. It's better to do these in the individual subfolder/submodules rather than in the global CMakeLists.txt.

CMakeLists.txt
```diff
@@ -152,8 +161,26 @@ if(OPT_DEBUG)
...
 
 # ManyMouse
-set(C_MANYMOUSE OFF)    # TODO
-set(SUPPORT_XINPUT2 OFF)    # TODO
+if (DOSBOX_PLATFORM_MACOS)
+  include (${CMAKE_CURRENT_SOURCE_DIR}/scripts/cmake/FindIOKit.cmake)
+  set(C_MANYMOUSE (${HAVE_FD_ZERO} AND ${IOKit_FOUND}))
+else()
+  set(C_MANYMOUSE "${HAVE_FD_ZERO}")
+endif()
```
johnnovak, line 169: So it will be always on? Don't we want to do `OPT_MANYMOUSE`, so turn it into an option?

scripts/cmake/FindIOKit.cmake
FeralChild64: OK, moved to `src/libs/manymouse/cmake`.

CMakeLists.txt
```diff
@@ -151,9 +160,43 @@ if(OPT_DEBUG)
   endif()
 endif()
 
+# ManyMouse - optional XInput support
+option(OPT_MANYMOUSE
+       "Use ManyMouse library for single-seat multiplayer gaming in The Settlers I/II"
```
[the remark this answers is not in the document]
weirddan455: Haha yeah I think hotseat is the term.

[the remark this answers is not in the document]
FeralChild64: OK, changed to `single-computer`.

CMakeLists.txt
FeralChild64: But how to I propagate the `C_MANYMOUSE` value to `src/hardware/input/` then?

johnnovak: Yeah, I wasn't aware of that 😄 It's best to keep it this way then if `C_MANYMOUSE` is used in two different branches of the source tree.

You could use forced cache variables as global variables (see [here](https://cliutils.gitlab.io/modern-cmake/chapters/basics/variables.html#cache-variables)), but let's leave setting all `C_*` options in the top level CMakeLists.txt then.

Could you at least do the `target_link_libraries ... $<IF:$<BOOL:${SUPPORT_XINPUT2}>,${X11_Xinput_LIB},>` stuff in `src/libs/manymouse/CMakeLists.txt`? That would be better and could be done without complications, I think.

FeralChild64: Done, target link library moved.

FeralChild64:
@johnnovak Main already fails with 178 PVS Studio warnings instead of allowed 176.

I had to raise allowed warnings to 177 (there is one warning against `src/hardware/input/mouse_manymouse.cpp` on the _main_, which only occurs if stub code is used instead of _ManyMouse_).

johnnovak:
> Main already fails with 178 PVS Studio warnings instead of allowed 176.

Yeah, I bumped it to 178 in my cleanup PR too. Maybe it was coming from the hard disk noise PR (PVS doesn't run on PRs from forks).

johnnovak (approved):
Looking good @FeralChild64, nice work 👍🏻

PR #4345, opened 2025-05-25
Replace non-ASCII copyright symbol in winres.rc

weirddan455:
# Description

My Clang + Ninja build on WIndows started to fail after 3b20fb811024a090ee1ec84df1e8d9f148c14208 now that we're compiling `winres.rc`. It gives the following error: 

```
llvm-rc: Error in VERSIONINFO statement (ID 1):
Non-ASCII 8-bit codepoint can't be interpreted in the current codepage
```

I researched the problem and found that libcurl ran into this same issue and this was their solution. I tried the prefix with "L" thing but it didn't work.

https://github.com/curl/curl/issues/7765

# Manual testing

Only affects Windows. It builds for me now.

# Checklist

I have:

### DIFF
--- a/src/winres.rc
+++ b/src/winres.rc
@@ -18,12 +18,12 @@ BEGIN
     BEGIN
         BLOCK "040904b0"
         BEGIN
-            VALUE "Comments", "© 2025 The DOSBox Staging Team, published under GNU GPL"
+            VALUE "Comments", "(C) 2025 The DOSBox Staging Team, published under GNU GPL"
             VALUE "CompanyName", "DOSBox Staging"
             VALUE "FileDescription", "DOSBox Staging DOS Emulator"
             VALUE "FileVersion", "0, 83, 0, 0"
             VALUE "InternalName", "dosbox-staging"
-            VALUE "LegalCopyright", "Copyright © 2025 The DOSBox Staging Team"
+            VALUE "LegalCopyright", "Copyright (C) 2025 The DOSBox Staging Team"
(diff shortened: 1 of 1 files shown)


shermp:
The other option might be to start the rc file with `#pragma code_page(65001)` to ensure it is interpreted as UTF-8 (and make sure the file is also saved as UTF-8.

See https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569

johnnovak:
> The other option might be to start the rc file with `#pragma code_page(65001)` to ensure it is interpreted as UTF-8 (and make sure the file is also saved as UTF-8.
> 
> See https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569

It's a very minor thing, so I'm happy either way. I'm leaving it do @weirddan455.

weirddan455:
> The other option might be to start the rc file with `#pragma code_page(65001)` to ensure it is interpreted as UTF-8 (and make sure the file is also saved as UTF-8.
> 
> See https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569

Tried it and it doesn't work. Seems that's another issue with llvm-rc: https://github.com/llvm/llvm-project/issues/63426

I don't know how to pass that CLI argument since we're not manually invoking llvm-rc, CMake is doing it automatically.

If had more extensive use of non-ascii characters I might dig into this more but in this case I'm in favor of doing the simple dumb thing and replacing with `(C)`. If @shermp or someone else wants to play with this more in a later PR, feel free.

PR #4349, opened 2025-05-29
Simplify 'for' help message handling

FeralChild64:
# Description

The `for` command help needs to display a percent (`%`) sign several time in it's output and the current help string contains the percent sign verbatim (no `%%` to print just one character by the prinf-like functions).

This is the only command needing this character, and handling the string without proper percent escaping needs quite some extra code - this PR adds the escaping and simplifies the implementation.

# Manual testing

For every language containing the `SHELL_CMD_FOR_HELP_LONG` string (en, br, de, es, it, nl, pl) check the `for` command help, for example:

```
config -set language=pl
help for
```

There should be no `SHELL_CMD_FOR_HELP_LONG` related warnings in the log, the `help for` command should print out the localized help message - each of them should contain a single `%%` (two percent signs) sequence.

- [ ] Windows
- [ ] macOS
- [x] Linux (the change is trivial)

# Checklist

I have:

### DIFF
--- a/include/messages.h
+++ b/include/messages.h
@@ -29,7 +29,6 @@ void MSG_LoadMessages();
 
 // Add message (in UTF-8 encoding) to the translation system
 void MSG_Add(const std::string& name, const std::string& message);
-void MSG_AddNoFormatString(const std::string& name, const std::string& message);
 
 // Get message text, adapted to the current code page,
 // with markup tags converted to ANSI escape codes
--- a/contrib/resources/translations/br.lng
+++ b/contrib/resources/translations/br.lng
@@ -4416,18 +4416,18 @@ Executa um comando especificado para cada string em um conjunto.
 .
 :SHELL_CMD_FOR_HELP_LONG
 Uso:
-  [color=light-green]for[reset] [color=white]%VAR[reset] [color=light-cyan]in[reset] [color=white](CONJUNTO)[reset] [color=light-cyan]do[reset] [color=white]COMANDO[reset]
+  [color=light-green]for[reset] [color=white]%%VAR[reset] [color=light-cyan]in[reset] [color=white](CONJUNTO)[reset] [color=light-cyan]do[reset] [color=white]COMANDO[reset]
 
 Parâmetros:
-  [color=white]%VAR[reset]          caractere único representando uma variável, prefixado por um '%'
+  [color=white]%%VAR[reset]          caractere único representando uma variável, prefixado por um '%%'
   [color=light-cyan]in[reset]            palavra-chave insensível a maiúsculas e minúsculas
-  [color=white](CONJUNTO)[reset]    conjunto de strings para substituir instâncias de [color=white]%VAR[reset]
+  [color=white](CONJUNTO)[reset]    conjunto de strings para substituir instâncias de [color=white]%%VAR[reset]
                 em [color=white]COMANDO[reset]
   [color=light-cyan]do[reset]            palavra-chave insensível a maiúsculas e minúsculas
   [color=white]COMANDO[reset]       comando a repetir para cada string em [color=white](CONJUNTO)[reset]
 
 Notas:
-  - Em arquivos em lote, [color=white]%VAR[reset] deve ser escrito como [color=white]%%VAR[reset] (dois sinais de
+  - Em arquivos em lote, [color=white]%%VAR[reset] deve ser escrito como [color=white]%%%%VAR[reset] (dois sinais de
     percentagem) em vez disso.
   - As strings em [color=white](CONJUNTO)[reset] podem ser separadas por qualquer separador
     DOS válido.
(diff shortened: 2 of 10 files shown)


include/messages.h
```diff
@@ -29,7 +29,6 @@ void MSG_LoadMessages();
 
 // Add message (in UTF-8 encoding) to the translation system
 void MSG_Add(const std::string& name, const std::string& message);
-void MSG_AddNoFormatString(const std::string& name, const std::string& message);
```
johnnovak, line 32: Good riddance 👍🏻 I'm working in this area on something and I was planning suggest to you to re-think this (or remove it myself), but looks like you have it under control 💯 Noice! 😎

johnnovak: Not really huge, no. After much deliberation I'm planning to go with this:

- `MSG_GetRaw` disappears in favour of `MSG_GetEnglishRaw` and `MSG_GetTranslatedRaw`.
- `MSG_GetForHost` will be renamed to `MSG_GetTranslatedRaw`
- `MSG_GetEnglishRaw` will be a new function

Returning `std::string` will be a welcome change, that doesn't affect me so we can work on this in parallel. Let me know if you have objections to the plan, IMO it would make for a more consistent API, and I need the `MSG_GetEnglishRaw` functionality in my upcoming PR.

This is the current state of things:

```cpp
/*
 * Add English message to the translation system in UTF-8.
 * The message may contais ANSI markup tags (e.g., [white]CONFIG[reset]).
 */
void MSG_Add(const std::string& name, const std::string& message);

/*
 * Get translated message preprocessed for output to the DOS console.
 *
 * On success, this returns the translated message, or the English message if
 * no translation exists for the current language, adapted to the currently
 * active DOS code page with markup tags converted to ANSI escape codes.
 *
 * Returns " MESSAGE NOT FOUND! " or " MESSAGE NOT VALID! " if there was an
 * error.
 *
 * Use `CONSOLE_Write()` to output the message to the DOS console.
 */
const char* MSG_Get(const std::string& name);

/*
 * Get the original English message in UTF-8.
 *
 * On success, this returns the exact same byte-sequence previously added with
 * `MSG_Add()` without any DOS code page or ANSI markup tags conversions
 * applied.
 *
 * Returns " MESSAGE NOT FOUND! " or " MESSAGE NOT VALID! " if there was an
 * error.
 */
const char* MSG_GetEnglishRaw(const std::string& name);

/*
 * Get the translated message in UTF-8.
 *
 * On success, this returns the translated message, or English message if no
 * translation exists for the current language, without any DOS code page or
 * ANSI markup tags conversions applied. It effectively returns the translated
 * message as-is from the UTF-8 language file.
 *
 * Returns " MESSAGE NOT FOUND! " or " MESSAGE NOT VALID! " if there was an
 * error.
 */
const char* MSG_GetTranslatedRaw(const std::string& name);
```

johnnovak (approved):
Nice cleanup job! 🎖️  Just one very minor comment.

include/messages.h
johnnovak: Actually, if you're happy with the above, how about I merge mine first in a couple of days then you do your `std::string` refactor? Just because I'm simplifying `messages.cpp` too a little bit in a few places (nothing major) so it would be easier for the both of us if you waited with your rework a bit.

FeralChild64: Sure, I'll wait. I just want to make it done before 0.83.0 - returning `char *` is not exactly safe as the lifecycle of the pointed data is not clear to the caller.

This API cleanup is perfectly fine for me.

FeralChild64:
@johnnovak The _Meson_ build system is no longer working for me on `main`. This is bad - we need to be able to run unit tests or produce sanitizer builds, and the CMake build system does not support these yet!

johnnovak:
> @johnnovak The _Meson_ build system is no longer working for me on `main`. This is bad - we need to be able to run unit tests or produce sanitizer builds, and the CMake build system does not support these yet!
> 
> [edit] Fixed on main, rebased.

Cheers, it's good you fixed it. It broke for me a while ago on macOS, that's when I decided I won't spend time on Meson personally but rather get CMake in a good shape instead.

The three things that are missing are unit test support, sanitizer builds, and the options for conditionally toggling features at compile time. 

I got the sanitizer builds going on macOS at least, just didn't raise a PR yet. Options are rather easy too. If you (or anybody else) could help with the unit test, that would be appreciated.

johnnovak:
> > The three things that are missing are unit test support, sanitizer builds, and the options for conditionally toggling features at compile time.
> 
> Don't forget the install target for cmake. Once we remove meson the arch AUR package will stop working, and having a working cmake install will make it trivial for them to make the necessary changes to their package.

That's a nice to have, an optional extra, not something we need for our official Linux packages. 

How Linux distros repackage our stuff is not our concern, but the Linux guys in our core team might add some extras to help them at their discretion. I'll leave it to them to decide and implement that as I don't use Linux as my primary OS.

shermp:
To be fair, if we get the `install` target right, we can likely get rid of the `create-package.sh` script.

johnnovak:
> To be fair, if we get the `install` target right, we can likely get rid of the `create-package.sh` script.

For Linux, maybe. We still need the macOS DMG creation and Windows ZIP and installer creation. This "install" concept is very specific to Linux and only when you build from source, then install the artifacts to your local machine.

I like the packager script concept better because the job of that is to create distributable packages for the three OSes we support.

So I'm like leave the packager script alone, but someone could add a Linux-only install command to the CMake script. Maybe the packager script could even call that to remove some redundancy.

johnnovak:
> EDIT: It's on my list of things to look at if no one else gets to it first.

Sure, but don't forget about macOS and Windows.

dividebysandwich:
@johnnovak I am not talking about repackaging, just about installing from source. I am a fan of sane defaults, and "make install" and the whole prefix thing has been standard regardless of distro. And the old meson build supported this as well. That's all I meant.
The arch PKGBUILD file just makes use of that, it sets the prefix and runs meson install. The fact that arch uses this to then build a package that actually gets installed is an arch speciality and we shouldn't care about this of course.

FeralChild64:
I suggest to move the CMake discussion here: https://github.com/dosbox-staging/dosbox-staging/issues/4196

PR #4351, opened 2025-05-30
[WIP] Introduce notifications subsystem

johnnovak:
# Description

## Related issues

# Release notes

# Manual testing

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


weirddan455:
Oh this could be a nice feature. I'm thinking this could be expanded to use for example `libnotify` on Linux to show toast notifications. Those are pretty easy to code and don't require linking against the large GUI frameworks.

Doesn't need to be done in this PR of course but I may look into it once you're done.

include/messages.h
```diff
@@ -24,25 +24,60 @@
...
+ * Load translated messages according to the configuration.
+ */
 void MSG_LoadMessages();
 
-// Add message (in UTF-8 encoding) to the translation system
+/*
+ * Add English message to the translation system in UTF-8.
+ * The message may contais ANSI markup tags (e.g., [white]CONFIG[reset]).
+ */
 void MSG_Add(const std::string& name, const std::string& message);
```
FeralChild64, line 19: `name` -> `message_key`

src/misc/setup.cpp
```diff
@@ -178,7 +179,7 @@ bool Value::SetBool(const std::string& in)
 	auto in_lowercase = in;
 	lowcase(in_lowercase);
 	const auto parsed = parse_bool_setting(in_lowercase);
-	_bool = parsed ? *parsed : false;
+	_bool             = parsed ? *parsed : false;
```
FeralChild64: IMHO this is indented too far.

src/misc/setup.cpp
```diff
@@ -1752,8 +1783,12 @@ Verbosity Config::GetStartupVerbosity() const
 		             : Verbosity::High;
 	}
 
-	LOG_WARNING("SETUP: Invalid 'startup_verbosity' setting: '%s', using 'high'",
-	            user_choice.c_str());
+	NOTIFY_DisplayWarning(Notification::Source::Console,
+	                      "SETUP",
```
FeralChild64: Why "SETUP" and not "CONFIG"?

FeralChild64:
@johnnovak Something strange is happening with the GitHub review, it behaves now in a way I don't really understand. Do you see my comments?

image

johnnovak:
@FeralChild64 Yeah I got individual notifications for each commit you commented on. I can only see most of your comments if I view the individual commits. In the PR view, I can only see 3 comments.

That's fine, I'll find them, so don't try to make the same comments again in a different way.

Maybe they're changing how commenting on individual commits are handled in big PRs.

src/misc/setup.cpp
johnnovak: I just used the existing prefix from the log statement, but CONFIG would be better for these, agreed.

src/misc/setup.cpp
johnnovak: Yeah, I'll do another pass and insert blank lines to get rid of these.

src/misc/setup.cpp
johnnovak: Changed all to "CONFIG".

include/notifications.h
```diff
@@ -174,7 +174,7 @@ void display_message([[maybe_unused]] const Level level, const Source source,
 	const auto is_console_initialised = first_shell;
 
 	if (source == Source::Console && is_console_initialised) {
-		const auto format = MSG_Get(message_key);
+		const auto& format = MSG_Get(message_key);
```
FeralChild64, line 177: `const auto& format` - why do we need a reference here? Maybe just write:

```C++
const auto str = format_str(MSG_Get(message_key), args...);
```

include/messages.h
```diff
@@ -46,7 +46,7 @@ void MSG_Add(const std::string& name, const std::string& message);
  *
  * Use `CONSOLE_Write()` to output the message to the DOS console.
  */
-const char* MSG_Get(const std::string& message_key);
+const std::string& MSG_Get(const std::string& message_key);
```
FeralChild64: Why do we want to return a reference to `std::string` instead of `std::string`?

The current code is not 100% safe - in the event of translation or code page change the returned pointer gets invalidated. We need to return `std::string` to make the code safe, not `std::string&`. Yes, there are going to be some extra memory allocations, but this is not a performance critical code at all.

johnnovak: The reason is [const-correctness](https://isocpp.org/wiki/faq/const-correctness). The number 1 goal is clear contracts and restricting the scope of mutations because code that can potentially mutate anything is extremely hard to reason about and maintain. Performance is merely a side effect.

The same for use of consts in function args; primary goal is clean, maintable code, and we get better performance "for free" by being const-correct and using references.

I don't see how having dangling pointers to const string references is any worse than to non-const strings. The signature of `c_str()` is this

```
const char* c_str() const;
```
You cannot modify the data returned by `c_str()`, unless you do explicit unsafe casting, but both the compiler and the linters will give you a very hard time.

I don't get it where you think the bug can happen. If you don't hold on to the std::string in the downstream code but hold on to the invalidated pointer returned by c_str and use it after the non-const std::string goes out of scope, you'll be getting random segfaults potentially. If that's the case, that needs to be fixed, and that has nothing to do with this change.

I don't see how using const string references changes any of that. Same level of safety when using c_str on it (or the lack thereof, depending on downstream code).

johnnovak: A good perspective on why making everything that can be const as early as possible const is important:

https://forum.osdev.org/viewtopic.php?t=32603

johnnovak: @FeralChild64 Nevermind... it was early in the morning. I just remembered you can't really have immutable strings in C++ easily in all scenarios. Locally `const std::string` is good, but when returning stuff, probably returning a copy with `std::string` is what you want in 90% of the cases.

Forgot that in `const std::string&` it's actually the _reference_ that is const, not the string itself 🥶 Anyway, happy to change it to simply return copies.

weirddan455: > The number 1 goal is clear contracts and restricting the scope of mutations

Then you should just return by value always unless you need to save performance. There's a worse bug that can happen and with a very unclear contract by returning a reference. Take this function for example (which was like this before this PR):

https://github.com/dosbox-staging/dosbox-staging/blob/ccc8309414420879996c7506051c721603646c08/src/misc/messages.cpp#L125-L142

```cpp
auto first_raw = msg.Get();
// Change the code page
auto second_raw = msg.Get();
LOG(first_raw);
```

That's undefined behavior that you wouldn't necessarily expect. The second call can free the string returned by the first call which is arguably worse than any violation of const correctness.

The main problem here is lifetimes and C++ has no way to reason about them so a reference is almost as unsafe as a raw pointer.

johnnovak: > That's undefined behavior that you wouldn't necessarily expect. The second call can free the string returned by the first call which is arguably worse than any violation of const correctness.

I read it 5 times already, but i'm not seeing it 😅 I believe you, though 😆 

> The main problem here is lifetimes and C++ has no way to reason about them so a reference is almost as unsafe as a raw pointer.

Yeah, and I take that all back 😄 Just I'm using other languages at work with much better immutability guarantees, so that's my standard. But yeah, in C++ it's a bit of a shitshow... I still stand by what I wrote above—conceptually, but it's really hard to make that happen in C++ with return values, agreed.

Returning `std::string` it is then 99.99% of the time. We're not dealing with megabytes/gigabytes of strings, so who cares about a few copies and extra mallocs under the hood.

All the smart pointer stuff is rather nice though (when you can use them).

weirddan455: > I read it 5 times already, but i'm not seeing it 😅 I believe you, though 😆

`message_dos_ansi` is the backing string. The `Get()` function both replaces `message_dos_ansi` (freeing the old string) and returns a reference to it. So it's very easy to have a dangling reference.

> All the smart pointer stuff is rather nice though (when you can use them).

You can do `std::shared_ptr<std::string>>` although that's a bit crap because it can have a double indirection. Probably a copy-on-write String library would be the way to go if we run into performance issues.

johnnovak: > Probably a copy-on-write String library would be the way to go if we run into performance issues.

Yeah, that will never happen I think because `std::string`s are not used in the core hardware emulation layers where performance matters, just for stuff like config handling. They are also used in some of our DOS implementation, I think, but copying a few filenames now and then is nothing.

Even in the 90s even the demo people started writing file I/O, config handling, etc. stuff in C because performance is just a non-issue; you're not even calling that code once per frame. Those things can be many many orders of magnitude slower than graphics routines and such.

FeralChild64:
I'll probably look at it again tomorrow - unfortunately, I had to do a crazy amount of overhours at work, not much time left for the hobby.

FeralChild64:
I cannot compile the current branch (GCC 14.3.0):

```
../src/midi/midi_fluidsynth.cpp:447:9: error: narrowing conversion of ‘3.0e+0’ from ‘double’ to ‘int’ [-Wnarrowing]
  447 |         3.0, // voice count
      |         ^~~
../src/midi/midi_fluidsynth.cpp: In function ‘std::optional<ChorusParameters> parse_custom_chorus_params(const std::string&)’:
../src/midi/midi_fluidsynth.cpp:524:46: warning: narrowing conversion of ‘(double)level’ from ‘double’ to ‘float’ [-Wnarrowing]
  524 |         return ChorusParameters{voice_count, level, speed, depth, mod_wave};
      |                                              ^~~~~
../src/midi/midi_fluidsynth.cpp:524:53: warning: narrowing conversion of ‘(double)speed’ from ‘double’ to ‘float’ [-Wnarrowing]
  524 |         return ChorusParameters{voice_count, level, speed, depth, mod_wave};
      |                                                     ^~~~~
../src/midi/midi_fluidsynth.cpp:524:60: warning: narrowing conversion of ‘(double)depth’ from ‘double’ to ‘float’ [-Wnarrowing]
  524 |         return ChorusParameters{voice_count, level, speed, depth, mod_wave};
      |                                                            ^~~~~
../src/midi/midi_fluidsynth.cpp: At global scope:
../src/midi/midi_fluidsynth.cpp:612:13: warning: ‘void setup_reverb(fluid_synth_t*)’ defined but not used [-Wunused-function]
  612 | static void setup_reverb(fluid_synth_t* synth)
      |             ^~~~~~~~~~~~
```

johnnovak:
@FeralChild64 Pull again and and make sure to force-reset to origin. I force pushed various updates to earlier commits. All commits compile here and CI is fine too (those failed builds are just warnings count issues)

I've tested everything except for the generic config messages, that's still outstanding. Plus I'm in the middle of reworking the chorus & reverb handling in FluidSynth (easier for me to bundle it up with this change; I've generalised some error messages, and the existing implementation was buggy & a bit messy).

So except for FluidSynth and the generic errors you can test everything else.

image

FeralChild64:
I've already pulled the latest code. BTW, there are compiler warnings even on the `main` branch now; I'm not sure whether they are due to recent code changes or due to me updating the GCC compiler from 14.2.x to 14.3.0:

```
../src/gui/sdlmain.cpp: In function ‘void handle_cli_set_commands(const std::vector<std::__cxx11::basic_string<char> >&)’:
../src/gui/sdlmain.cpp:5271:44: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::__cxx11::basic_string<char> >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
 5271 |                         for (auto i = 3; i < pvars.size(); i++) {
      |                                          ~~^~~~~~~~~~~~~~
../src/gui/sdlmain.cpp: At global scope:
../src/gui/sdlmain.cpp:950:13: warning: ‘void exit_fullscreen()’ defined but not used [-Wunused-function]
  950 | static void exit_fullscreen()
```

johnnovak:
> I'm not sure whether they are due to recent code changes or due to me updating the GCC compiler from 14.2.x to 14.3.0:

That could be it. CI uses GCC 12, and the CI builds are fine.

johnnovak:
@FeralChild64 Just pushed up a fix for those warnings you posted. Probably they were flagged by PVS or Sonar too, but I just finished this before dinner so didn't check 😆

FeralChild64:
```
../src/misc/messages_stubs.cpp:28:13: error: ambiguating new declaration of ‘const char* MSG_Get(const std::string&)’
   28 | const char* MSG_Get(const std::string&)
      |             ^~~~~~~
In file included from ../src/misc/messages_stubs.cpp:24:
../include/messages.h:49:13: note: old declaration ‘std::string MSG_Get(const std::string&)’
   49 | std::string MSG_Get(const std::string& message_key);
      |             ^~~~~~~
../src/misc/messages_stubs.cpp:33:13: error: ambiguating new declaration of ‘const char* MSG_GetEnglishRaw(const std::string&)’
   33 | const char* MSG_GetEnglishRaw(const std::string&)
      |             ^~~~~~~~~~~~~~~~~
../include/messages.h:60:13: note: old declaration ‘std::string MSG_GetEnglishRaw(const std::string&)’
   60 | std::string MSG_GetEnglishRaw(const std::string& message_key);
      |             ^~~~~~~~~~~~~~~~~
../src/misc/messages_stubs.cpp:38:13: error: ambiguating new declaration of ‘const char* MSG_GetTranslatedRaw(const std::string&)’
   38 | const char* MSG_GetTranslatedRaw(const std::string&)
      |             ^~~~~~~~~~~~~~~~~~~~
../include/messages.h:72:13: note: old declaration ‘std::string MSG_GetTranslatedRaw(const std::string&)’
   72 | std::string MSG_GetTranslatedRaw(const std::string& message_key);
      |   
```

johnnovak:
> ```
> ../src/misc/messages_stubs.cpp:28:13: error: ambiguating new declaration of ‘const char* MSG_Get(const std::string&)’
>    28 | const char* MSG_Get(const std::string&)
>       |             ^~~~~~~
> In file included from ../src/misc/messages_stubs.cpp:24:
> ../include/messages.h:49:13: note: old declaration ‘std::string MSG_Get(const std::string&)’
>    49 | std::string MSG_Get(const std::string& message_key);
>       |             ^~~~~~~
> ../src/misc/messages_stubs.cpp:33:13: error: ambiguating new declaration of ‘const char* MSG_GetEnglishRaw(const std::string&)’
>    33 | const char* MSG_GetEnglishRaw(const std::string&)
>       |             ^~~~~~~~~~~~~~~~~
> ../include/messages.h:60:13: note: old declaration ‘std::string MSG_GetEnglishRaw(const std::string&)’
>    60 | std::string MSG_GetEnglishRaw(const std::string& message_key);
>       |             ^~~~~~~~~~~~~~~~~
> ../src/misc/messages_stubs.cpp:38:13: error: ambiguating new declaration of ‘const char* MSG_GetTranslatedRaw(const std::string&)’
>    38 | const char* MSG_GetTranslatedRaw(const std::string&)
>       |             ^~~~~~~~~~~~~~~~~~~~
> ../include/messages.h:72:13: note: old declaration ‘std::string MSG_GetTranslatedRaw(const std::string&)’
>    72 | std::string MSG_GetTranslatedRaw(const std::string& message_key);
>       |   
> ```

Well, I'm not running Meson anymore, so fixing up the unit tests will have to wait until someone adds unit test support for CMake. Or if you volunteer to fix them up (after this PR goes in) if you're still using Meson 😄

FeralChild64:
I'm sorry, but in such case I can't help you right now. Due to personal life issues I won't have time to work on DOSBox Staging for a few days - and I certainly won't have time to fix the build regressions caused by someone else's unfinished PRs.

Please don't break unit tests and please don't break Meson build system until CMake build system is ready. CMake build system is hard to use for me, as it spams the output with countless compiler warnings, so I can't easily see whether everything is OK with my changes. Plus, given the nature of the `const char *` -> `std::string` type transition, it would be great to test your PR using the instrumented build.

[edit] Sorry if it sounds harsh - I intend to look at the CMake build scripts once again one day, but I definitely don't have time for such things at this moment.

include/notifications.h
```diff
@@ -0,0 +1,252 @@
...
+void NOTIFY_DisplayInfoMessage(const Source source, const std::string_view topic,
+                               const std::string& message_key,
+                               const Args&... args) noexcept
+{
+	display_message(Level::Info, source, topic, message_key, args...);
+}
+
+/*
+ * Display a informational message in response to the user changing a
+ * contiuously adjustable settin.
```
weirddan455: Typo: setting

weirddan455: Also continuously

src/shell/shell.cpp
```diff
@@ -39,10 +39,16 @@
 #include "timer.h"
 
 callback_number_t call_shellstop = 0;
-/* Larger scope so shell_del autoexec can use it to
- * remove things from the environment */
+
+// Larger scope so shell_del autoexec can use it to
+// remove things from the environment
 DOS_Shell *first_shell = nullptr;
```
weirddan455: Can this be made static now?

johnnovak: Yeah, I'll do that.

johnnovak:
> I'm sorry, but in such case I can't help you right now. Due to personal life issues I won't have time to work on DOSBox Staging for a few days - and I certainly won't have time to fix the build regressions caused by someone else's unfinished PRs.

No prob, and no rush. Working on Staging should be an enjoyable free-time activity 💯 

> Please don't break unit tests and please don't break Meson build system until CMake build system is ready.

Good point, I've asked the guys for help to add CMake unit test and sanitizer build support as well. I've done enough CMake stuff for a while myself 😄 

> CMake build system is hard to use for me, as it spams the output with countless compiler warnings, so I can't easily see whether everything is OK with my changes.

Yeah, the output is a bit noisy, I agree. It would be good to somehow make it a bit leaner, e.g., by not printing out the full compiler invocation commands which are super long. Maybe there's some Ninja option to trim down the output.

> Plus, given the nature of the `const char *` -> `std::string` type transition, it would be great to test your PR using the instrumented build.

That's a fair point.

> [edit] Sorry if it sounds harsh - I intend to look at the CMake build scripts once again one day, but I definitely don't have time for such things at this moment.

It didn't sound harsh, appreciate the feedback 👍🏻

johnnovak:
Back to draft until someone adds unit test and sanitizer build support to CMake.

FeralChild64:
@johnnovak I've taken a look at the warnings - I've fixed most of them (https://github.com/dosbox-staging/dosbox-staging/pull/4392), but not all.

The problem is that our Meson build system sets a lot of defines, like `HAVE_CLOCK_GETTIME`, which are used in several places in the code. This mechanism is not ported to CMake yet - so the CMake produced builds are most likely not compiled correctly, might miss some minor features, use obsolete APIs, etc. This definitely has to be handled before we migrate.

weirddan455:
> The problem is that our Meson build system sets a lot of defines, like HAVE_CLOCK_GETTIME, which are used in several places in the code. This mechanism is not ported to CMake yet - so the CMake produced builds are most likely not compiled correctly, might miss some minor features, use obsolete APIs, etc. This definitely has to be handled before we migrate.

There is a CMake equivalent to handle these defines. `HAVE_CLOCK_GETTIME` was an outlier that was forgotten. #4393 fixes this. If you've noticed any others that are missing, please let me know.

FeralChild64:
@weirddan455 At first glance: `HAVE_SETPRIORITY`, `PAGESIZE`, `HAVE_SYS_TYPES_H` (and other header checks).

We need to review them all, from start to end - if one was forgotten, there might be more. Maybe some are not needed and we can get rid of them.

johnnovak:
Thanks guys, that's great 👍🏻

FeralChild64:
@johnnovak To fix the compilation of `messages_stubs.cpp`, just put the following definitions instead of the current ones:

```c++
void MSG_Add(const std::string&, const std::string&) {}

std::string MSG_Get(const std::string&)
{
	return {};
}

std::string MSG_GetEnglishRaw(const std::string&)
{
	return {};
}

std::string MSG_GetTranslatedRaw(const std::string&)
{
	return {};
}

bool MSG_Exists(const std::string&)
{
	return true;
}

bool MSG_WriteToFile(const std::string&)
{
	return true;
}

void MSG_LoadMessages() {}
```

johnnovak:
This is considered tested & done from my end. I've tested all the messages I've converted to notifications, made a few adjustments to the message texts as I went, etc.

I guess we can either wait until someone adds unit testing and sanitizer support, _or_ @FeralChild64 or someone on Linux could perhaps do a quick test with the sanitizer build then we can merge it before that.

johnnovak:
@FeralChild64 All unit tests are passing now and I've tested it with the sanitiser as well, changing languages at runtime between English, German, Polish and French. No issues.

Can I get a tick please, then let's get this in?

To recap, I've done what you requested and I'm just returning `std::string` everywhere.

FeralChild64 (approved):

FeralChild64:
OK, so ready to merge.

PR #4360, opened 2025-06-02
Fix AUTOEXEC.BAT being deleted on disk swap

### LINKED ISSUE #4358: Swapping disks prevents further execution of [autoexec] section

_No response_

Linux x86_64

_No response_

### Relevant hardware info

_No response_

Using Ctrl+F4 to swap disk images causes DOSBox to "forget" any remaining `[autoexec]` commands or commands given with the `-c` command line argument. See reproduction section below.

Put the following into your [autoexec] section in dosbox-staging.conf. The images can be anything, either floppy or cdrom. The game can be anything as well (just using doom as an example):

```
[autoexec]
imgmount a /path/to/disk01.img /path/to/disk02.img -t floppy
mount c /path/to/dosroot
c:
cd doom_se
doom
exit
```

While your game is running, press Ctrl+F4 to swap images then exit the game, The act of swapping images causes DOSBox to forget about any commands remaining in `[autoexec]`. In this case, `exit` does not get run but this is reproducible for any command. If you do not Ctrl+F4, `exit` gets run as expected once Doom exits.

_No response_

### Your configuration

```ini

```

### Provide a log

```text

```

- [x] Yes, I agree.

weirddan455:
# Description

Root cause regression from f466410f49e7c7cae31c743878258723b35859a7
Further regression in 09722228235850c56c0e36b8f90c47dbfdd4dae5 - This caused the [autoexec] section to stop execution on disk swap

Virtual_Drive::EmptyCache() deletes and re-creates all files in Z:\ autoexec.cpp relied on an internal boolean to decide whether to update or re-create AUTOEXEC.bat This boolean is invalided by clearing the cache

Fix by attemping the update first and if the file does not exist, create a new one

Also removed some std::move shenaigans that weren't implemented properly to begin with. I'm pretty sure the receiving function needs to take a && for std::move to do any good.

## Related issues

Fixes #4358

# Release notes

Swapping disks with Ctrl+F4 no longer breaks AUTOEXEC.BAT

# Manual testing

Reproduction steps in #4358

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/include/dos_system.h
+++ b/include/dos_system.h
@@ -417,9 +417,9 @@ void VFILE_Register(const char *name,
                     const uint8_t *data,
                     const uint32_t size,
                     const char *dir = "");
-void VFILE_Register(const char* name, const std::vector<uint8_t>& blob,
+void VFILE_Register(const char* name, const std::vector<uint8_t>& data,
                     const char* dir = "");
-void VFILE_Update(const char* name, std::vector<uint8_t> blob, const char* dir = "");
+bool VFILE_Update(const char* name, const std::vector<uint8_t> &data, const char* dir = "");
 void VFILE_Remove(const char* name, const char* dir = "");
 
 #endif
--- a/src/dos/drive_virtual.cpp
+++ b/src/dos/drive_virtual.cpp
@@ -206,18 +206,21 @@ void VFILE_Register(const char *name,
 	first_file = new_file;
 }
 
-void VFILE_Register(const char *name, const std::vector<uint8_t> &blob, const char *dir)
+void VFILE_Register(const char *name, const std::vector<uint8_t> &data, const char *dir)
 {
-	VFILE_Register(name, blob.data(), check_cast<uint32_t>(blob.size()), dir);
+	VFILE_Register(name, data.data(), check_cast<uint32_t>(data.size()), dir);
 }
 
-void VFILE_Update(const char* name, std::vector<uint8_t> blob, const char* dir)
+bool VFILE_Update(const char* name, const std::vector<uint8_t> &data, const char* dir)
 {
 	auto vfile = find_vfile_by_name_and_dir(name, dir);
 
 	if (vfile) {
-		vfile->data = std::make_shared<std::vector<uint8_t>>(std::move(blob));
+		vfile->data = std::make_shared<std::vector<uint8_t>>(data);
+		return true;
 	}
+
+	return false;
 }
 
 void VFILE_Remove(const char* name, const char* dir)
--- a/src/shell/autoexec.cpp
+++ b/src/shell/autoexec.cpp
@@ -229,13 +229,11 @@ static void create_autoexec_bat_dos(const std::string& input_utf8,
 	                                             autoexec_bat_dos.end());
 
 	// Register/refresh Z:\AUTOEXEC.BAT file
-	if (is_vfile_registered) {
-		VFILE_Update(AutoexecFileName.c_str(), std::move(autoexec_bat_bin));
-	} else {
-		VFILE_Register(AutoexecFileName.c_str(),
-		               std::move(autoexec_bat_bin));
-		is_vfile_registered = true;
+	if (!VFILE_Update(AutoexecFileName.c_str(), autoexec_bat_bin)) {
+		VFILE_Register(AutoexecFileName.c_str(), autoexec_bat_bin);
 	}
+
+	is_vfile_registered = true;
 
 	// Store current code page for caching purposes
 	vfile_code_page = code_page;


johnnovak:
> Also removed some std::move shenaigans that weren't implemented properly to begin with. I'm pretty sure the receiving function needs to take a && for std::move to do any good.

That was a contributor called MeeAreJenius (lol 😂) who was "vehemently" arguing for using more advanced features like `std::move` and some other stuff. The problem is, `std::move` adds considerable mental overhead and like everything in C++, there are arcane and non-intuitive rules around its usage. My view is only use it when it makes a significant difference (e.g., moving vs copying megabytes of data), not for something as trivial as string manipulation in the filesystem code... But even then, I'd try hard to use plain old references instead.

I still don't fully understand whether his use of `std::move` is 100% correct or not, but what is certain, we're still fixing regressions introduced by his cleanup years later (there were at least 4-5 similar regressions so far).

include/dos_system.h
```diff
@@ -419,7 +419,7 @@ void VFILE_Register(const char *name,
                     const char *dir = "");
 void VFILE_Register(const char* name, const std::vector<uint8_t>& blob,
                     const char* dir = "");
-void VFILE_Update(const char* name, std::vector<uint8_t> blob, const char* dir = "");
+bool VFILE_Update(const char* name, const std::vector<uint8_t> &blob, const char* dir = "");
```
johnnovak: Can you rename `blob` to something actually meaningful while you're at it? Please 😅 

All these names carry the exact same amount of information (zero):

- blob
- data
- theThing
- asdf
- bytes
- ...

Although "data" is kinda crap, at least rename it to "data" as this ends up in the "data" member of some struct?

johnnovak (approved):
Change looks okay, thanks for fixing it. It's pretty much impossible to reason about these things by just looking at the code alone, so approving if it works now 😅

weirddan455:
> > Also removed some std::move shenaigans that weren't implemented properly to begin with. I'm pretty sure the receiving function needs to take a && for std::move to do any good.
> 
> That was a contributor called MeeAreJenius (lol 😂) who was "vehemently" arguing for using more advanced features like `std::move` and some other stuff. The problem is, `std::move` adds considerable mental overhead and like everything in C++, there are arcane and non-intuitive rules around its usage. My view is only use it when it makes a significant difference (e.g., moving vs copying megabytes of data), not for something as trivial as string manipulation in the filesystem code... But even then, I'd try hard to use plain old references instead.
> 
> I still don't fully understand whether his use of `std::move` is 100% correct or not, but what is certain, we're still fixing regressions introduced by his cleanup years later (there were at least 4-5 similar regressions so far).

It was @FeralChild64 that wrote these instances of `std::move`.

The short of it is `std::move` is just a cast from, for example, `std::vector` to `std::vector &&`. The optimization magic happens on the receiving end. Ex: `std::string a = std::move(b);` is only able to optimize out the copy because the standard library provides an operator overload that takes an R-value reference `&&`.

In this case, we're calling one of our own functions that was actually pass by value for the `blob` parameter. The `std::move` cast doesn't do what we want because it has to construct a new `std::vector` anyway.

johnnovak:
> It was @FeralChild64 that wrote these instances of `std::move`.

Okay, MeAreJeenius is off the hook then in this case. Maybe 😛 

> In this case, we're calling one of our own functions that was actually pass by value for the `blob` parameter. The `std::move` cast doesn't do what we want because it has to construct a new `std::vector` anyway.

Makes sense, thanks for the explanation.

include/dos_system.h
weirddan455: Renamed to data

FeralChild64:
[edit] OK, I was too quick with my concerns - although the comment claims we shouldn't run `create_autoexec_bat_utf8` more than once (and it is going to be run now with every disk swap), it uses static strings, so it won't change the file length. Yet, I think we could simplify the whole implementation.

My proposal:
- `AUTOEXEC_RegisterFile` and `AUTOEXEC_NotifyNewCodePage` should get unified into one function, I'm going to call it `AUTOEXEC_Regenerate` here (but feel free to invent a better name)
- `Add_VFiles` should call `AUTOEXEC_Regenerate` always, the `add_autoexec` argument shouldn't be needed
- `is_vfile_registered` variable should disappear

The `AUTOEXEC_Regenerate` should do as follows:
- if `shutdown_requested` is set - we are shutting down, do nothing, quit
- if `autoexec_module` (pointer to the `AutoExecModule` object) is still null - we are to early during the startup, we don't have all the information to generate the `AUTOEXEC.BAT` yet, quit
- if the UTF-8 variant of the `AUTOEXEC.BAT` hasn't been created yet - create it by calling `create_autoexec_bat_utf8` (this should be done only once during the emulator lifecycle)
- call `create_autoexec_bat_dos` to generate `AUTOEXEC.BAT` translated to the current code page, update the virtual file if it exists, create it otherwise (just like with your change)

The optimization:

```C++
	const auto code_page = get_utf8_code_page();
	if (code_page == vfile_code_page) {
		return;
	}
```

is not crucial - feel free to leave it out if you find it problematic.

FeralChild64:
@weirddan455 Since I was the one who originally wrote the code, it was easier for me to do the rework - I did in on top of your fix: https://github.com/dosbox-staging/dosbox-staging/tree/fc/autoexec-tmp-1

So maybe just merge the changes and I'll submit my cleanup PR afterwards.

FeralChild64 (approved):

weirddan455:
@FeralChild64 I was just now working through your comment.  I agree with your proposed changes but I'll let you take it from here since you're more familiar with this code.

johnnovak:
@weirddan455 Nice work 🎉 Could you be a bit more specific in the release notes, though?

PR #4368, opened 2025-06-09
Make restart_dosbox() go through the normal shutdown routine

### LINKED ISSUE #4318: Command history is not preserved when restarting with `CONFIG -R`

_No response_

Windows 10

_No response_

### Relevant hardware info

_No response_

1. Change some setting that requires a restart, e.g. `windowresolution 1200x700`
2. `CONFIG -R`
3. Press the up arrow and note `windowresolution 1200x700` is not in the command history.

Basically, the command history is not preserved when `CONFIG -R` is issued like it is on exit.

_No response_

_No response_

### Your configuration

```ini

```

### Provide a log

```text

```

- [x] Yes, I agree.

weirddan455:
# Description

Restarting either by `config -r` or by Ctrl+Alt+Home hotkey now goes through the normal shutdown procedure. A child process gets created that waits for the parent process to terminate before continuing.

What caused me to look into this was the command line history not being saved on restart but I was concerned about other clean-up routines not being run (ex. some of the file system meta-data doesn't get written until close).

This ensures that restart is the same as a normal shutdown. It also delays init of the new process until the old process is fully terminated.

## Related issues

Fixes #4318

# Release notes

Restarting DOSBox now correctly saves the command line history.

# Manual testing

Command line history is now saved after restart on Windows and Linux. Log output confirms the parent process gets fully shutdown before the child process starts init.

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -29,6 +29,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <signal.h>
 #include <sys/types.h>
 #include <tuple>
 #include <unistd.h>
@@ -39,7 +40,6 @@
 
 #ifdef WIN32
 #include <process.h>
-#include <signal.h>
 #include <windows.h>
 #endif
 
@@ -4773,6 +4773,36 @@ int sdl_main(int argc, char* argv[])
 	control = std::make_unique<Config>(&command_line);
 
 	const auto arguments = &control->arguments;
+
+	if (arguments->wait_pid) {
+#ifdef WIN32
+		// Synchronize permission is all we need for WaitForSingleObject()
+		constexpr DWORD DesiredAccess = SYNCHRONIZE;
+		constexpr BOOL InheritHandles = FALSE;
+		HANDLE process                = OpenProcess(DesiredAccess,
+                                             InheritHandles,
+                                             *arguments->wait_pid);
+		if (process) {
+			// Waits for the process to terminate.
+			// If we failed to open it, it's probably already dead.
+			constexpr DWORD Timeout = INFINITE;
+			WaitForSingleObject(process, Timeout);
+			CloseHandle(process);
+		}
+#else
+		for (;;) {
+			// Signal of 0 does not actually send a signal.
+			// It only checks for existance and permissions of the PID.
+			constexpr int Signal = 0;
+			int ret = kill(*arguments->wait_pid, Signal);
+			// ESRCH means PID does not exist.
+			if (ret == -1 && errno == ESRCH) {
+				break;
+			}
+			Delay(50);
+		}
+#endif
+	}
 
 	switch_console_to_utf8();
 
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -4570,11 +4570,38 @@ static int edit_primary_config()
 extern void DEBUG_ShutDown(Section * /*sec*/);
 #endif
 
-void restart_dosbox(std::vector<std::string> &parameters) {
+static void remove_waitpid(std::vector<std::string> &parameters)
+{
+	auto it = parameters.begin();
+	while (it != parameters.end()) {
+		if (*it == "-waitpid" || *it == "--waitpid") {
+			it = parameters.erase(it);
+			if (it != parameters.end()) {
+				auto &pid = *it;
+				if (!pid.empty() && std::isdigit(pid[0])) {
+					// The integer value should be the next element following "--waitpid".
+					// If we found an integer, remove it as well.
+					it = parameters.erase(it);
+				}
+			}
+		} else {
+			++it;
+		}
+	}
+}
+
+void restart_dosbox(std::vector<std::string> &parameters)
+{
 
 	control->ApplyQueuedValuesToCli(parameters);
 
+	// Remove any existing --waitpid parameters.
+	// This can happen with multiple restarts.
+	remove_waitpid(parameters);
+
 #ifdef WIN32
+	parameters.emplace_back("--waitpid");
+	parameters.emplace_back(std::to_string(GetCurrentProcessId()));
 	std::string command_line = {};
 	bool first = true;
 	for (const auto &arg : parameters) {
@@ -4585,6 +4612,8 @@ void restart_dosbox(std::vector<std::string> &parameters) {
 		first = false;
 	}
 #else
+	parameters.emplace_back("--waitpid");
+	parameters.emplace_back(std::to_string(getpid()));
 	char** newargs = new char* [parameters.size() + 1];
 	// parameter 0 is the executable path
 	// contents of the vector follow
@@ -4595,9 +4624,7 @@ void restart_dosbox(std::vector<std::string> &parameters) {
 	newargs[parameters.size()] = nullptr;
 #endif // WIN32
 
-	MIXER_CloseAudioDevice();
-	Delay(50);
-	QuitSDL();
+	GFX_RequestExit(true);
 
 #if C_DEBUG
 	// shutdown curses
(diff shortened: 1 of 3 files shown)


johnnovak:
> It's hipster Unix and they're too cool to have a `/proc` apparently.

Erm, excuse me? 😂 

https://www.opengroup.org/openbrand/register/brand3710.htm

It's _more_ Unix than Linux, mate 😎🤘🏻 Well, at least it's certified (whatever that means, lol 😂)

And some actual info to make my comment not only participation-award worthy:
https://superuser.com/a/631696

Happy to test stuff if you find a viable macOS solution.

UPDATE: Can't you just use `waitpid()`? 

https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/waitpid.2.html

weirddan455:
> UPDATE: Can't you just use waitpid()?

`waitpid()` only works when you're the parent waiting on a child. We're doing the reverse (child waiting on the parent).

If I try it the other way, it segfaults. SDL/X11/PulseAudio don't like it when a new child process tries to take over for the parent so the child process has to be the one that calls `execvp()` and the parent has to clean up after itself.

I'm still going to look into this more later.

johnnovak:
@weirddan455 We could always use the filesystem on macOS: generate a "pid" file in `/tmp`, delete the file on exit, then poll that file in the child process until it gets deleted.

weirddan455:
> @weirddan455 We could always use the filesystem on macOS: generate a "pid" file in `/tmp`, delete the file on exit, then poll that file in the child process until it gets deleted.

But when is "on exit"? The last bit of our code that runs from what I can tell is `atexit(QuitSDL);` but then global destructors still run after that.

If I had a defined exit point in our code, I would just defer `execvp()` until then rather than making a temp file but it's more complicated than that.

johnnovak:
@weirddan455 Magic AI generated answer how to do this on a BSD (macOS is a BSD, effectively). @kklobe is now proud of me, I'm sure 😛 

Maybe it's all humbug, but hey, took me 5 seconds 😆 

Btw, I was thinking even before looking this up to programatically query the parent PID from the child process, then you don't need to pass it along. Maybe just detect at runtime if the current Staging process has a parent process. I'm sure there is a Windows API call for getting the parent PID as well.


image

image

image

weirddan455:
I'll give AI half-credit. `kill()` with a signal of 0 is pretty clever and works on Linux as well.

> Btw, I was thinking even before looking this up to programatically query the parent PID from the child process, then you don't need to pass it along. Maybe just detect at runtime if the current Staging process has a parent process. I'm sure there is a Windows API call for getting the parent PID as well.

This doesn't really work. At least on Linux, a process always has a parent. If the parent DOSBox process has terminated already, it gets re-parented but not always to PID 1. On my system, `systemd --user` became the parent with a PID of over 1000. We would have to query the parent process for more information and decide whether it's DOSBox or a system process and that seems error-prone and OS specific.

A way I could avoid sending the PID is to do the `kill()` + wait loop after `fork()` and before `execvp()` but I don't think there's a way to do this on Windows since there's no real `fork()` alternative. We're using `CreateProcess()`.

johnnovak:
> This doesn't really work. At least on Linux, a process always has a parent.

Ah, that's right. I'm sure it's the same on all OSes, actually. I'm just not a systems programmer so it didn't really occur to me when I wrote that 😄 

> A way I could avoid sending the PID is to do the kill() + wait loop after fork() and before execvp() but I don't think there's a way to do this on Windows since there's no real fork() alternative. We're using CreateProcess().

Nah, just pass it then. That works and it's 100% reliable.

weirddan455:
Ready for review. Updated toplevel comment, Added and tested Windows codepath. Mac will take the same codepath as Linux but I need a Mac user to test.

src/gui/sdlmain.cpp
```diff
@@ -4774,6 +4774,33 @@ int sdl_main(int argc, char* argv[])
...
+			// Waits for the process to terminate.
+			// If we failed to open it, it's probably already dead.
+			WaitForSingleObject(process, INFINITE);
+			CloseHandle(process);
+		}
+		#else
+		while (1) {
+			// Signal of 0 does not actually send a signal.
+			// It only checks for existance and permissions of the PID.
+			constexpr int SIGNAL = 0;
```
johnnovak: Signal

src/gui/sdlmain.cpp
```diff
@@ -4774,6 +4774,33 @@ int sdl_main(int argc, char* argv[])
...
+		constexpr BOOL InheritHandles = FALSE;
+		HANDLE process = OpenProcess(DesiredAccess, InheritHandles, *arguments->wait_pid);
+		if (process) {
+			// Waits for the process to terminate.
+			// If we failed to open it, it's probably already dead.
+			WaitForSingleObject(process, INFINITE);
+			CloseHandle(process);
+		}
+		#else
+		while (1) {
```
johnnovak: No love for `for (;;)`? 😎

src/gui/sdlmain.cpp
```diff
@@ -4774,6 +4774,33 @@ int sdl_main(int argc, char* argv[])
...
+			// It only checks for existance and permissions of the PID.
+			constexpr int SIGNAL = 0;
+			int ret = kill(*arguments->wait_pid, 0);
+			// ESRCH means PID does not exist.
+			if (ret == -1 && errno == ESRCH) {
+				break;
+			}
+			Delay(50);
+		}
+		#endif
```
johnnovak: Nice explanatory comments 🎖️ 

Just do a Clang reformat on this block, please; that will pull the ifdefs to the start of the lines.

src/gui/sdlmain.cpp
```diff
@@ -4570,11 +4570,38 @@ static int edit_primary_config()
...
+					// The integer value should be the next element following "--waitpid".
+					// If we found an integer, remove it as well.
+					it = parameters.erase(it);
+				}
+			}
+		} else {
+			++it;
+		}
+	}
+}
```
johnnovak, line 4591: I think we should use `CommandLine::FindRemoveIntArgument()` instead.

Ah, or maybe we can't use that here...

src/gui/sdlmain.cpp
```diff
@@ -4570,11 +4570,38 @@ static int edit_primary_config()
...
+
+void restart_dosbox(std::vector<std::string> &parameters)
+{
 
 	control->ApplyQueuedValuesToCli(parameters);
 
+	// Remove any existing --waitpid paramaters.
```
johnnovak: typo: parameters

src/gui/sdlmain.cpp
```diff
@@ -4646,13 +4673,24 @@ void restart_dosbox(std::vector<std::string> &parameters) {
...
+	int ret = fork();
+	switch (ret) {
+		case -1:
+			LOG_ERR("Restart failed: fork failed: %s", strerror(errno));
+			break;
+		case 0:
+			// Newly created child process immediately executes a new instance of DOSBox.
+			// It is not safe for a child process in a multi-threaded program to do much else.
+			if (execvp(newargs[0], newargs) == -1) {
+				E_Exit("Restart failed: execvp failed: %s", strerror(errno));
```
johnnovak, line 4688: Any reason for the hard exit? Wouldn't `LOG_ERR` be better?

src/gui/sdlmain.cpp
weirddan455: I had looked at that before I wrote this. It's tied to the CommandLine class and doesn't do exactly what I need anyway.

src/gui/sdlmain.cpp
weirddan455: This is in the child process. It crashes if you try to do a normal exit. The parent process will do a normal exit regardless if we've gotten this far. If the child fails `execvp()`, we have to kill it.

src/gui/sdlmain.cpp
```diff
@@ -4570,11 +4570,38 @@ static int edit_primary_config()
...
-void restart_dosbox(std::vector<std::string> &parameters) {
+static void remove_waitpid(std::vector<std::string> &parameters)
+{
+	auto it = parameters.begin();
+	while (it != parameters.end()) {
+		if (*it == "-waitpid" || *it == "--waitpid") {
+			it = parameters.erase(it);
+			if (it != parameters.end()) {
+				auto &pid = *it;
+				if (!pid.empty() && pid[0] >= '0' && pid[0] <= '9') {
```
FeralChild64: `std::isdigit(pid[0])`?

src/gui/sdlmain.cpp
weirddan455: fixed

src/gui/sdlmain.cpp
weirddan455: I don't really have a preference. I changed it for you :smile:

src/gui/sdlmain.cpp
weirddan455: Formatted

src/gui/sdlmain.cpp
weirddan455: fixed

src/gui/sdlmain.cpp
weirddan455: Sure. Changed.

src/gui/sdlmain.cpp
johnnovak: Fair enough.

src/gui/sdlmain.cpp
johnnovak: Haha, I'm just used to it 😆

src/gui/sdlmain.cpp
johnnovak: Yeah ok, that makes sense.

johnnovak:
Is there anything outstanding on this @weirddan455 ? I've confirmed it works on macOS as expected.

weirddan455:
> Is there anything outstanding on this @weirddan455 ? I've confirmed it works on macOS as expected.

Nope, this just slipped my mind. Merged.

Issue #4374, opened 2025-06-10
Disknoises audio callback is running even when the feature is disabled


weirddan455:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Linux x86_64

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Are you able to trigger the issue with our official DOSBox Staging release?

- [x] I am running an official Staging release, not a release installed by my operating system's package manager, or a version repackaged by an external team.

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

`DiskNoises::AudioCallback()` is constantly running even when the feature is disabled. Found this by accident when it segfaulted on me. See #4373 although it's not directly related to this.

Don't have time to dig more into this at the moment. I'm not sure if it has any negative user impact but it's generally best practice to disable mixer channels if we're not using them to conserve resources.

### Steps to reproduce the behaviour

Explain in detail how to reproduce the bug. Assume no familiarity with the software.
1. Set `hard_disk_noise = off` and  `floppy_disk_noise = off`
2. Put log statement inside `DiskNoises::AudioCallback()` and observe it constantly running.

### Your configuration

```ini
[disknoise]
hard_disk_noise   = off
floppy_disk_noise = off
```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

weirddan455:
Tagging @dividebysandwich if you want to have a look. If not, I can probably take care of this later but I have other things I'm working on.

johnnovak:
> I'm not sure if it has any negative user impact but it's generally best practice to disable mixer channels if we're not using them to conserve resources.

I agree with that. Should be zero user and performance impact but we should disable channels when they're not in use.

Jkapp76:
Are there any plans to include activity lights with the disk sounds?

PR #4376, opened 2025-06-11
Only register disk noise callback if required

dividebysandwich:
# Description

This PR prevents registering the audio callback when no disk noise is enabled in the configuration

## Related issues

#4374 

# Manual testing

Tested locally. Added log output inside the callback, disabled both FDD and HDD noise in config. Also tested with only HDD and only FDD, as well as both enabled, to verify no errors on shutdown.

- [ ] Windows
- [ ] macOS
- [X] Linux

# Checklist

I have:

- [X] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/CONTRIBUTING.md#commit-messages).

### DIFF
--- a/src/hardware/disk_noise.cpp
+++ b/src/hardware/disk_noise.cpp
@@ -50,14 +50,18 @@ DiskNoises::DiskNoises(const bool enable_floppy_disk_noise,
                        const std::string& floppy_spin,
                        const std::vector<std::string>& floppy_seek_samples)
 {
+	if (!enable_floppy_disk_noise && !enable_hard_disk_noise) {
+		return;
+	}
+
 	MIXER_LockMixerThread();
 	const auto mixer_callback = std::bind(&DiskNoises::AudioCallback,
 	                                      this,
 	                                      std::placeholders::_1);
-	mix_channel               = MIXER_AddChannel(mixer_callback,
-                                       DiskNoiseSampleRateInHz,
-                                       ChannelName::DiskNoise,
-	                                             {ChannelFeature::Stereo});
+	mix_channel = MIXER_AddChannel(mixer_callback,
+	                               DiskNoiseSampleRateInHz,
+	                               ChannelName::DiskNoise,
+	                               {ChannelFeature::Stereo});
 	mix_channel->Enable(true);
 	float vol_gain = percentage_to_gain(100);
 	mix_channel->SetAppVolume({vol_gain, vol_gain});
@@ -111,13 +115,8 @@ DiskNoises::~DiskNoises()
 {
 	MIXER_LockMixerThread();
 
-	if (floppy_noise) {
-		floppy_noise.reset();
-	}
-
-	if (hdd_noise) {
-		hdd_noise.reset();
-	}
+	floppy_noise.reset();
+	hdd_noise.reset();
 	active_devices.clear();
 
 	if (active_devices.empty() && mix_channel) {


src/hardware/disk_noise.cpp
```diff
@@ -50,34 +50,36 @@ DiskNoises::DiskNoises(const bool enable_floppy_disk_noise,
...
-
-	floppy_noise = std::make_shared<DiskNoiseDevice>(DiskType::Floppy,
-	                                                 enable_floppy_disk_noise,
-	                                                 floppy_spin_up,
-	                                                 floppy_spin,
-	                                                 floppy_seek_samples,
-	                                                 false);
-	active_devices.emplace_back(floppy_noise);
-	MIXER_UnlockMixerThread();
+	if (enable_floppy_disk_noise || enable_hard_disk_noise) {
```
weirddan455: I'd prefer flattening this like:

```cpp
if (!enable_floppy_disk_noise && !enable_hard_disk_noise) {
    return;
}
```

Better yet, can we bail out even earlier inside `disknoise_init()`? That way we leave `disk_noises` as null instead of having this weird half-constructed object.

src/hardware/disk_noise.cpp
```diff
@@ -109,24 +111,26 @@ void DiskNoises::AudioCallback(const int num_frames_requested)
 
 DiskNoises::~DiskNoises()
 {
-	MIXER_LockMixerThread();
+	if (floppy_noise || hdd_noise) {
```
weirddan455: There's not really a point in this check. We should always try destroying the members. If they don't exist, it doesn't cause a problem. These extra if statements just add noise (no pun intended :laughing: )

src/hardware/disk_noise.cpp
```diff
@@ -109,24 +111,26 @@ void DiskNoises::AudioCallback(const int num_frames_requested)
...
+		MIXER_LockMixerThread();
 
-	if (floppy_noise) {
-		floppy_noise.reset();
-	}
+		if (floppy_noise) {
+			floppy_noise.reset();
+		}
```
weirddan455: Same here. It doesn't matter if this is null or not. We're going to set it to null either way (and it's a no-op if it is already null).

src/hardware/disk_noise.cpp
```diff
@@ -109,24 +111,26 @@ void DiskNoises::AudioCallback(const int num_frames_requested)
...
 
-	if (hdd_noise) {
-		hdd_noise.reset();
-	}
-	active_devices.clear();
+		if (hdd_noise) {
+			hdd_noise.reset();
+		}
```
weirddan455: This if statement can go as well.

weirddan455 (commented):
My main concern with touching this code is whether it will break the I/O throttling you implemented. I don't see anything to suggest it will but if you can test that to be sure, that would be good.

Also go ahead and rebase this PR on `main` so that CI will pass. I pushed some fixes yesterday.

Thanks for taking care of this so quickly :smile:

src/hardware/disk_noise.cpp
johnnovak: I agree, we try to use "flat" structures when exiting early from a function vs nesting.

src/hardware/disk_noise.cpp
johnnovak: Yeah, unnecessary if guards just add confusion.

src/hardware/disk_noise.cpp
dividebysandwich: That was my java lizard-brain at work here ;)

weirddan455:
Can you squash those two commits into one? We try to avoid "fixup" commits since they don't really have any value. You can do `git rebase -i main` and mark the 2nd commit as "fixup" to squash it together with the first.

CI for Mac is also failing with "Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run" but I can't see how that would possibly be related to your changes.

Hopefully it'll fix itself once you push again. If not I'd say merge anyway and see what happens on `main`. CI has been real irritating lately with how often it fails for reasons completely unrelated to the PR it fails on.

johnnovak:
> CI for Mac is also failing with "Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run" but I can't see how that would possibly be related to your changes.

Yeah, usually the second run is fine.

> CI has been real irritating lately with how often it fails for reasons completely unrelated to the PR it fails on.

"AI agents" overloading public services... maybe?

weirddan455 (approved):
I think this is good now. @johnnovak do you have any objections to merging this?

johnnovak:
> I think this is good now. @johnnovak do you have any objections to merging this?

Looks good, merged! 😎

Issue #3552, opened 2024-03-29
Mount improvements megaticket


johnnovak:
@weirddan455's [recent ticket](https://github.com/dosbox-staging/dosbox-staging/issues/3548) reminded me of some long-standing enhancement ideas I've had with kcgen for the mount commands. Also, it turns out we have a lot of mount-related tickets in our issue tracker.

These are all somewhat related, and there is some overlap between them, so ideally, a single person would take charge of these mounting improvements. This does not mean he would solve them all, but at least triage them and come up with a plan, etc.

The most important things to address IMO are these at a high level:

- **Support automounting CD-ROM images** (e.g., just chuck a bunch of images into `drives/d` in your game folder, or `drives/e`, `drives/f` etc. for games that support multiple CD-ROm drives).
- **Infer the `-t` argument whenever it can be done with a reasonably high degree of confidence.** For `.bin`, `.cue` and `.iso` files this is trivial. For floppy images, they have fixed sizes so we can just simply look at the file size. Probably these are the most important anyway.
- **Logging & help text improvements**. Fairly trivial, see the tickets.

All mount related tickets:

- https://github.com/dosbox-staging/dosbox-staging/issues/3548
- https://github.com/dosbox-staging/dosbox-staging/issues/2509
- https://github.com/dosbox-staging/dosbox-staging/issues/3527
- https://github.com/dosbox-staging/dosbox-staging/issues/3305
- https://github.com/dosbox-staging/dosbox-staging/issues/2811
- https://github.com/dosbox-staging/dosbox-staging/issues/2602
- https://github.com/dosbox-staging/dosbox-staging/issues/3450
- https://github.com/dosbox-staging/dosbox-staging/issues/3030
- https://github.com/dosbox-staging/dosbox-staging/issues/1265
- https://github.com/dosbox-staging/dosbox-staging/issues/92

Jkapp76:
I don't understand why making the Y drive read-only is better.

johnnovak:
> I don't understand why making the Y drive read-only is better.

Because the Y drive holds commands that are part of DOSBox Staging. So it's preferable if you can't delete them.

Jkapp76:
I'm a rule-breaker I guess.

johnnovak:
> I'm a rule-breaker I guess.

Technically, you can do it, but why bother when you can create something like a U: drive (that's what I do, for "utilities"), then just put it into the path. Then you don't muck around with our built-in Y drive.

[...]

PR #4379, opened 2025-06-14
Automount CD images from drives/[x] folders

farsil:
# Description

This PR allows CD images in supported formats to be detected at start in `drives/[x]` folders and auto-mounted accordingly. Multiple images are supported and mounted in the order `directory_iterator` returns files. The mechanism can be overridden by explicitly forcing a `type` in the `x.conf` file that is not `cdrom`. 

I implemented the override because it was easy to do so, but an argument could be made to axe it completely due to using the `type` setting in a way that wasn't originally intended. I've decided to leave it so that it can be discussed.

## Related issues

[I remember there was some mega-ticket concerning filesystem mounts, but I can't seem to find it]

# Release notes

No additional configuration settings are necessary, but if we decide to include the override, `type` will have a slightly different meaning when image files are present, and that will have to be documented.

# Manual testing

I tested it in debug mode and release mode by dropping a couple of bin/cue disks in the `drives/d` folder. They were detected and mounted without issues.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:


src/shell/autoexec.cpp
```diff
@@ -260,6 +260,14 @@ class AutoExecModule final : public Module_base {
...
 	void AddMessages();
+
+	// Specify a 'Drive' config object with allowed key and value types
+	static std::unique_ptr<Config> SpecifyDriveConf();
+
+	// Parse a 'Drive' config file and return object with allowed key and
+	// value types
+	static std::tuple<std::string, std::string, std::string, bool> ParseDriveConf(
+		std::string drive_letter, const std_fs::path& conf_path);
 };
```
johnnovak: Nice cleanup job. We don't tend to do static class methods, though. In this case, I think it would be better to just keep that as static private functions outside of the class, so just a straight move.

src/shell/autoexec.cpp
```diff
@@ -619,6 +628,54 @@ std::string AutoExecModule::BuildAutoMountFolderCommand(
 	return command;
 }
 
+std::vector<std_fs::path> AutoExecModule::GetMountableImagesPaths(const std_fs::path& drive_path)
```
johnnovak: Maybe `GetMountableCdImagesPaths()`? There are also floppy images (and HDD images, for that matter).

Better yet, maybe you could generalise it to pass in the list of allowed extensions, then we could use it later for auto-mounting floppy images too?

src/shell/autoexec.cpp
```diff
@@ -619,6 +628,54 @@ std::string AutoExecModule::BuildAutoMountFolderCommand(
...
+{
+	std::vector<std_fs::path> image_paths;
```
johnnovak: Init with `{}`.

src/shell/autoexec.cpp
```diff
@@ -619,6 +628,54 @@ std::string AutoExecModule::BuildAutoMountFolderCommand(
...
+
+	for (auto const& entry : std_fs::directory_iterator(drive_path)) {
+		if (entry.exists() && entry.is_regular_file() ) {
+			auto const& path = entry.path();
+			std::string ext = path.extension();
+			std::ranges::transform(ext, ext.begin(), ::tolower);
```
johnnovak: Use `lowcase` from `string_utils.h`.

src/shell/autoexec.cpp
```diff
@@ -619,6 +628,54 @@ std::string AutoExecModule::BuildAutoMountFolderCommand(
...
+			if (ext == ".iso" || ext == ".cue") {
+				image_paths.push_back(path);
+			}
+		}
+	}
+
+	return image_paths;
+}
+
+std::string AutoExecModule::BuildAutoMountImagesCommand(
```
johnnovak: `BuildAutoMountCdImagesCommand` perhaps? We can refactor later if we decide to generalise this.

src/shell/autoexec.cpp
```diff
@@ -636,7 +636,7 @@ std::vector<std_fs::path> AutoExecModule::GetMountableImagesPaths(const std_fs::
 		if (entry.exists() && entry.is_regular_file() ) {
 			auto const& path = entry.path();
 			std::string ext = path.extension();
-			std::ranges::transform(ext, ext.begin(), ::tolower);
+			lowcase(ext);
```
johnnovak, line 629: Right, so you've discovered it... I'm commenting commit by commit, so ignore my earlier comment please 😄

src/shell/autoexec.cpp
```diff
@@ -693,9 +693,10 @@ void AutoExecModule::AutoMountDetectedDrive(const std::string& dir_letter)
 	// See if there are mountable images in the path
 	const auto image_paths = GetMountableImagesPaths(drive_path);
 
-	// Setting the drive type to 'dir' will force dosbox to mount the path
-	// as a directory, even if there are mountable images in the path
-	if (image_paths.empty() || settings->type == "dir") {
+	// Explicitly setting the drive type to anything but 'cdrom' will force
+	// dosbox to mount the path as a directory, even if there are mountable
+	// images in the path
+	if (image_paths.empty() || (!settings->type.empty() && settings->type != "cdrom")) {
```
johnnovak: Agree with the commit description, we don't want to complicate things for niche usages like that. Good call!

johnnovak:
Here's the megaticket you couldn't find:

- https://github.com/dosbox-staging/dosbox-staging/issues/3552

I've done a quick first pass, this is some high quality work, mate! 🎉 Very good job adopting our coding standards virtually perfectly in your first PR, congrats! 🥳 

Will give a good testing when it comes out of draft, but things are looking very promising so far!

src/shell/autoexec.cpp
```diff
@@ -507,9 +540,152 @@ void AutoExecModule::AddLine(const Placement placement, const std::string& line)
...
+	}
+
+	return command;
+}
+
+std::vector<std_fs::path> AutoExecModule::GetMountableImagesPaths(const std_fs::path& drive_path)
+{
+	std::vector<std_fs::path> image_paths;
+
+	for (const auto& entry : std_fs::directory_iterator(drive_path)) {
```
FeralChild64, line 625: The directory content order is unspecified, right?
Wouldn't it be better to sort it, so that `FooBarGame_CD1.iso` is going to be mounted as the 1st image, `FooBarGame_CD2.iso` as the 2nd, etc.?

src/shell/autoexec.cpp
```diff
@@ -260,6 +260,39 @@ class AutoExecModule final : public Module_base {
...
+	// Specify a 'Drive' config object with allowed key and value types
+	static std::unique_ptr<Config> SpecifyDriveConf();
+
+	// Parse a 'Drive' config file and return object with allowed key and
+	// value types
+	static std::optional<AutoMountSettings> ParseDriveConf(const std_fs::path& conf_path);
+
+	// Build a command to mount a drive from a folder, based on the
+	// provided settings
+	static std::string BuildAutoMountFolderCommand(
```
FeralChild64: Style: I would prefer `AutoMountDirectoryCommand` name (same for other identifiers with _folder_ in them). All the APIs we use (`POSIX`, `std::fs`, etc.) use a _directory_ term (not the _folder_ term), same with almost every single piece of currently supported guest software.

FeralChild64:
@farsil I believe you meant this ticket: https://github.com/dosbox-staging/dosbox-staging/issues/3552

src/shell/autoexec.cpp
johnnovak: Good point, please only use "directory" in the code.

src/shell/autoexec.cpp
johnnovak: Yeah, I'm pretty sure you'll need to sort it (check out how the FluidSynth SF2 resolution works, that must be A-Z sorted).

src/shell/autoexec.cpp
farsil: I don't mind doing that, I actually thought about doing this later (maybe in another PR) if we ever wanted to support other types of images. I guess the answer is yes :grin:

src/shell/autoexec.cpp
farsil: I've thought about it, and I decided against it because not every game uses that convention. For example, the game Command & Conquer: Red Alert does not have a CD1 and a CD2, but an Allied Disc and a Soviet Disc.

I assume the majority of games though still use CD1 CD2 etc so it could make sense to sort by default and let the users deal with the exceptions if you prefer that approach.

src/shell/autoexec.cpp
farsil: Yeah, to be fair, as I mentioned in the PR description, I challenge the usefulness of the override at all, but since it was so simple I thought there was no harm in supporting this scenario.

src/shell/autoexec.cpp
farsil: Do we use varargs, or should we use other mechanisms like initializer lists? I've not seen either of them in the codebase?

The idea would be to have something like this:

```
get_files_by_ext(const std_fs::path& path, const std::initializer_list<std::string_view>& exts);

...

get_files_by_ext(path, { ".cue", ".iso" });
```

src/shell/autoexec.cpp
```diff
@@ -274,21 +283,21 @@ AutoExecModule::AutoExecModule(Section* configuration)
...
-			AutoMountDrive({letter});
+			AutoMountDetectedDrive({letter});
 		}
 	}
 
 	// Check -securemode switch to disable mount/imgmount/boot after
 	// running autoexec.bat
-	const auto cmdline = control->cmdline; // short-lived copy
-	const auto arguments = &control->arguments;
+	const auto cmdline               = control->cmdline; // short-lived copy
```
farsil, line 292: Ah, my bad this must have been my IDE. If you want I can revert these formatting changes to unrelated parts of the codebase.

src/shell/autoexec.cpp
farsil: I've used initializer lists for now to test if the idea worked, please tell me if you prefer a different approach.

src/shell/autoexec.cpp
johnnovak: @farsil 
From the documentation:
https://en.cppreference.com/w/cpp/filesystem/directory_iterator.html

> The iteration order is unspecified, except that each directory entry is visited only once.

Unspecified means that the mount order will be completely random. Might be different on different runs, on somebody else's machine when you move the game over. Or your own machine when you simply move the game dir to a different place and the underlying filesystem representation changes.

We want features to be deterministic, and relying on unspecified C/C++ behaviour is always a bad idea when different compilers and operating systems are involved.

The IMGMOUNT command supports wildcard mounts and it uses natural sort order. So that's the way to go for consistency, and that's the best way to handle this anyway.

See here:
https://github.com/dosbox-staging/dosbox-staging/blob/57556eefeafd4eff92e0ac4efa9d4aee9d54360d/src/dos/program_imgmount.cpp#L118

People can always prepend their filenames with numbers if they're dealing with different base names and want to ensure a specific order. We just need to document this.

src/shell/autoexec.cpp
johnnovak: We don't shy away from using certain language features, but there has to be a good reason for adopting a new approach. Generally, the simplest and least surprising way to do something is the best unless the different and new approach has significant benefits.

We don't use initialiser lists like this, so it's not the least surprising approach. Neither is it the simplest; it involves using a verbose language construct..

Can you please just make the function accept a `const std::vector<std::string>>&`, then define these extensions as a constant vector in the implementation cpp file and just pass that in?

src/shell/autoexec.cpp
```diff
@@ -507,9 +516,163 @@ void AutoExecModule::AddLine(const Placement placement, const std::string& line)
...
+	auto conf = std::make_unique<Config>();
+
+	// Define the [drive] section
+	constexpr auto changeable_at_runtime = false;
+	const auto prop = conf->AddSection_prop("drive", nullptr, changeable_at_runtime);
+
+	// Define the allowed keys and types
+	constexpr auto on_startup = Property::Changeable::OnlyAtStart;
+	prop->Add_string("type", on_startup, "")
+	        ->Set_values({"dir", "floppy", "cdrom", "overlay"});
```
johnnovak, line 532: Clang reformat please

src/shell/autoexec.cpp
```diff
@@ -507,9 +516,163 @@ void AutoExecModule::AddLine(const Placement placement, const std::string& line)
...
+			lowcase(ext);
+			if (std::ranges::find(extensions, ext) != extensions.end()) {
+				paths.push_back(path);
+			}
+		}
+	}
+
+	std::ranges::sort(paths, [](const std_fs::path& a, const std_fs::path& b) {
+		return a.filename() < b.filename();
+	});
```
johnnovak, line 638: As mentioned above, use `natural_compare` here, just like the `IMGMOUNT` command.

johnnovak:
Tested on macOS and it's working very well @farsil. I've tested mounting multiple drives as well, so `drives/d` and drives/e` with one CD image in each folder.

I think it's worth keeping the override. It might be handy, and consistency beats special casing, so it's better IMO that you can specify the override both for HDD and CD-ROM mounts.

Not entirely sure why the Windows build is failing... I tried re-running it about 3 times, but it always failed the same way. I created a new test branch off current `main`, in that one the Windows build succeeds 🤷🏻 

https://github.com/dosbox-staging/dosbox-staging/actions/runs/15657841748

So dunno, maybe rebase to `main` and see what happens?

src/shell/autoexec.cpp
farsil: That's absolutely fair, will do

src/shell/autoexec.cpp
farsil: I think this is already the result of `clang-format` conventions

johnnovak: Just tried it, and indeed. A bit weird, but whatever then 😆 🤷🏻

farsil:
Rebased, but still no dice :/ It seems that cmake is successful but something returns a non-zero error code that makes it fail.

image

weirddan455:
Why are we running the build command like this: `cmake --build --preset release-windows >> build.log 2>&1`

Makes it impossible to see the error because it's just going to some temp file on a GitHub server we can't access.

weirddan455:
#4382 should make the errors visible (I hope, we'll see once CI finishes)

weirddan455:
@farsil Change to CI is merged. Rebase to main again and you should be able to see why the Windows build is failing.

src/shell/autoexec.cpp
```diff
@@ -507,9 +516,160 @@ void AutoExecModule::AddLine(const Placement placement, const std::string& line)
...
+	const auto prop = conf->AddSection_prop("drive", nullptr, changeable_at_runtime);
+
+	// Define the allowed keys and types
+	constexpr auto on_startup = Property::Changeable::OnlyAtStart;
+	prop->Add_string("type", on_startup, "")
+	        ->Set_values({"dir", "floppy", "cdrom", "overlay"});
+	prop->Add_string("label", on_startup, "");
+	prop->Add_string("path", on_startup, "");
+	prop->Add_string("override_drive", on_startup, "");
+	prop->Add_bool("verbose", on_startup, true);
```
farsil: @johnnovak I've noticed that, when the conf file exists, the default value of the verbose flag is `true` while if the conf file does not exist, verbose is assumed `false`. What do you think of changing this default to `false` as well, so that both cases are consistent? Is there a reason it was designed this way?

johnnovak: I think it's simply a coding mistake. Good catch! 🎉 

The defaults were defined this way in the code you refactored:

```cpp
	// Default return values
	constexpr auto default_args = "";
	constexpr auto default_path = "";
	constexpr auto default_verbosity = false;
```

So yeah, the default should be false.

Just an idea: your new `AutoMountSettings` struct already contains the correct defaults, so instead of duplicating them as literals, you could do this for all relevant settings:

```cpp

const AutoMountSettings defaults = {};
...
prop->Add_bool("verbose", on_startup, defaults.verbosity);
```

johnnovak:
@farsil Here's some help for resolving the Windows-only error:

On Windows, we need to do `path.string()` instead of just `path` when we want to treat it as a `std::string`. Looks like it gets implicitly converted on non-Windows platforms or something. Just do `git grep "path\.string()"` and you'll see we use`path.string()` everywhere.

It's one of those murky areas of the C++ standard 🤷🏻 

@shermp tried to explain it to me once... wide-characters... Unicode... UTF-8... something something 😅 

So this will compile:

```cpp
LOG_ERR("AUTOMOUNT: %s: setting 'override_drive = %s' is invalid",
		        conf_path.string().c_str(),
		        override_drive.c_str());
```

weirddan455:
`std::path` uses platform native encoding which is UTF-16 on Windows and UTF-8 everywhere else. The only time it's really problematic is when you call `.c_str()` on it because then you get a `w_char *` on Windows rather than the `char *` most C functions expect.

So yeah, like @johnnovak said, workaround is to do an intermediate conversion to `std::string` like `path.string().c_str()`.

shermp:
> `std::path` uses platform native encoding which is UTF-16 on Windows and UTF-8 everywhere else. The only time it's really problematic is when you call `.c_str()` on it because then you get a `w_char *` on Windows rather than the `char *` most C functions expect.
> 
> So yeah, like @johnnovak said, workaround is to do an intermediate conversion to `std::string` like `path.string().c_str()`.

I think it's even worse than that, `std::path` can be either UTF-16 OR "ANSI" (8 bit) depending on how the path was originally constructed.

farsil:
What's worse is that the build errored on lines 627 and 635 of autoexec.cpp (you can see that in lines 608 and 616 of the log file [here](https://github.com/dosbox-staging/dosbox-staging/actions/runs/15675187820/job/44153743647)) , which means it really doesn't like implicit path to string conversion. The line where I convert to `c_str()` merely produced a warning. At any rate, I made every conversion explicit now so hopefully it should work.

e: It does :tada:

src/shell/autoexec.cpp
farsil: The only annoying part of this is that `Add_string` wants a C-style string so I have to `.c_str()` everything:

```
	prop->Add_string("type", on_startup, defaults.type.c_str())
	        ->Set_values({"dir", "floppy", "cdrom", "overlay"});
	prop->Add_string("label", on_startup, defaults.label.c_str());
	prop->Add_string("path", on_startup, defaults.path.c_str());
	prop->Add_string("override_drive", on_startup, defaults.override_drive.c_str());
	prop->Add_bool("verbose", on_startup, defaults.verbose);
	prop->Add_bool("readonly", on_startup, defaults.readonly);
```

I'd still go for it since it's good to have a single source of truth for defaults

farsil:
About the override, we may want to think if we like to "extend" the current behavior to the other future, possible cases. If we decide to go that way:

Floppy images will be overridden with type equals to "dir", "overlay" or "cdrom".
HDD images (if we will ever care) will be overridden with type equals to "floppy", "overlay" or "cdrom".

An alternative would be to introduce other configuration settings or more types, although I have the feeling John doesn't like the idea too much :grinning:

johnnovak:
> About the override, we may want to think if we like to "extend" the current behavior to the other future, possible cases. If we decide to go that way:
> 
> Floppy images will be overridden with type equals to "dir", "overlay" or "cdrom". HDD images (if we will ever care) will be overridden with type equals to "floppy", "overlay" or "cdrom".
> 
> An alternative would be to introduce other configuration settings or more types, although I have the feeling John doesn't like the idea too much 😀

Yeah setting `type` explicitly will take care of all the use cases when you want to override the automatic behaviour. Kinda defeats the point of the automation a little bit, though (it's a hassle to create a drive config), but the option is there.

johnnovak (approved):
Nice work @farsil 🎉 Thanks for addressing all our comments promptly 🙏🏻 

If there's nothing outstanding on this, I'm happy to merge it.

farsil:
All good for me, let's merge this :)

dux119-cmd:
This has been super handy! Thank you!

Is there any chance for floppy supports such as  `drives/a` and `/drives/b` when full of `.ima` or `.img` files?

I sometimes install Windows 3.1 software, and they were often released on floppies (sometimes lots of floppies!) Being able to drag an application's `.img` files into `drives/a` would make it super easy to try out, and avoid me having to adjust my conf or point imgmount to a different path to find the application images.

I think some early DOS games also used floppies for save games, so this would be really handy too.

Issue #3971, opened 2024-10-03
Salvage parts of GranMinigun's unfinished SDL & shader related rework


johnnovak:
As per title. It doesn't seem likely he'll come back anytime soon and pick these up again (judging by his public GitHub activity, at least), and there are some good improvements there. I'm intending to do this at some point, but if anyone else feels the urge, be my guest 😎 

(But cc'ing you anyway, man, just in case 😄  @GranMinigun)

Needs careful manual application of the changes with some slight further rework, perhaps.

These are the commit in question:

- https://github.com/dosbox-staging/dosbox-staging/pull/2450/commits/30d31f243be2a886388812bf5d49cd5a52f37321
- https://github.com/dosbox-staging/dosbox-staging/pull/2450/commits/c9f5de28b4a1f34c97fecd34bb4fe01f10b05616
- https://github.com/dosbox-staging/dosbox-staging/pull/2997/commits/fe713c0da8882b86dd3e60e77a946130f99b0693

Make sure to read the PR descriptions too for context (e.g., dropping the ancient fixed pipeline OpenGL stuff is definitely a good idea and should happen):

- https://github.com/dosbox-staging/dosbox-staging/pull/2997
- https://github.com/dosbox-staging/dosbox-staging/pull/2450

Grounded0:
![image](https://github.com/user-attachments/assets/3ffcc57f-038e-49dd-94dc-8ab8deff1b29)

Grounded0:
Sir GranMinigun has been notified on IRC.

weirddan455:
![sdlmain](https://github.com/user-attachments/assets/c4acf1cb-f247-4e0b-8608-d7babf28272c)

weirddan455:
On a more serious note, https://github.com/dosbox-staging/dosbox-staging/commit/30d31f243be2a886388812bf5d49cd5a52f37321 is something I noticed myself and mentioned on Discord.  I agree with removing those fixed function OpenGL calls.  I wasn't aware he had already started work on doing so.  I think that code rarely, if ever, gets run on any system.  It's used as a fallback in case the shaders fail to compile.  That should only happen if we have a bug in the shader code (if so, we fix the bug) or the user has a problem with their OpenGL drivers (if so, `output = texture` is a better fallback).

[...]

PR #4384, opened 2025-06-16
SDL main cleanup, part 1

johnnovak:
# Description

No functional changes (if I haven't screwed anything up), just a lift-and-shift job, basically.

Definitely review commit by commit.

I have grand plans for sdlmain, such as adding OSD support, a multi-pass shader pipeline, collapse all the CRT shaders into a single shader that will be parameterised at runtime, add brightness/contrast/saturation/white balance/etc. monitor controls, and more! I'm intending to add all that soon(ish) (definitely this year, and for 0.83.0), but I need to rein the chaos in a bit first.

Parts 2 & 3 will incorporate parts of GranMinigun's SDL & shader rework/cleanup (see https://github.com/dosbox-staging/dosbox-staging/issues/3971) and moving the OpenGL stuff into its own file.

# Manual testing

- Start up a few games
- Test various the command line params still work
- Test restarting works
- Test pause/unpause and the pause when inactive mode still works

- [ ] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -2748,7 +2748,7 @@ void GFX_Start() {
 	sdl.active=true;
 }
 
-void GFX_ObtainDisplayDimensions() {
+static void get_display_dimensions() {
 	SDL_Rect displayDimensions;
 	SDL_GetDisplayBounds(sdl.display_number, &displayDimensions);
 	sdl.desktop.full.width = displayDimensions.w;
@@ -2764,7 +2764,7 @@ void GFX_ObtainDisplayDimensions() {
 void GFX_UpdateDisplayDimensions(int width, int height)
 {
 	if (sdl.desktop.full.display_res && sdl.desktop.fullscreen) {
-		/* Note: We should not use GFX_ObtainDisplayDimensions
+		/* Note: We should not use get_display_dimensions()
 		(SDL_GetDisplayBounds) on Android after a screen rotation:
 		The older values from application startup are returned. */
 		sdl.desktop.full.width = width;
@@ -3531,7 +3531,7 @@ static void read_gui_config(Section* sec)
 
 	sdl.desktop.full.display_res = sdl.desktop.full.fixed && (!sdl.desktop.full.width || !sdl.desktop.full.height);
 	if (sdl.desktop.full.display_res) {
-		GFX_ObtainDisplayDimensions();
+		get_display_dimensions();
 	}
 
 	set_output(section, is_aspect_ratio_correction_enabled());
@@ -3654,7 +3654,7 @@ static void handle_video_resize(int width, int height)
 	/* Maybe a screen rotation has just occurred, so we simply resize.
 	There may be a different cause for a forced resized, though.    */
 	if (sdl.desktop.full.display_res && sdl.desktop.fullscreen) {
-		/* Note: We should not use GFX_ObtainDisplayDimensions
+		/* Note: We should not use get_display_dimensions()
 		(SDL_GetDisplayBounds) on Android after a screen rotation:
 		The older values from application startup are returned. */
 		sdl.desktop.full.width  = width;
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -266,7 +266,7 @@ static const char* safe_gl_get_string(const GLenum requested_name,
 	// the result points to a static string but can be null
 	const auto result = glGetString(requested_name);
 
-	// the default, howeever, needs to be valid
+	// the default, however, needs to be valid
 	assert(default_result);
 
 	return result ? reinterpret_cast<const char*>(result) : default_result;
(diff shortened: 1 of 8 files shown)


src/gui/sdlmain.cpp
```diff
@@ -101,67 +101,89 @@ static void restore_console_encoding()
...
-typedef void (APIENTRYP PFNGLGETSHADERINFOLOGPROC) (GLuint shader, GLsizei bufSize, GLsizei *length, GLchar *infoLog);
-typedef GLint (APIENTRYP PFNGLGETUNIFORMLOCATIONPROC) (GLuint program, const GLchar *name);
-typedef void (APIENTRYP PFNGLLINKPROGRAMPROC) (GLuint program);
-//Change to NP, as Khronos changes include guard :(
-typedef void (APIENTRYP PFNGLSHADERSOURCEPROC_NP) (GLuint shader, GLsizei count, const GLchar **string, const GLint *length);
-typedef void (APIENTRYP PFNGLUNIFORM2FPROC) (GLint location, GLfloat v0, GLfloat v1);
-typedef void (APIENTRYP PFNGLUNIFORM1IPROC) (GLint location, GLint v0);
-typedef void (APIENTRYP PFNGLUSEPROGRAMPROC) (GLuint program);
-typedef void (APIENTRYP PFNGLVERTEXATTRIBPOINTERPROC) (GLuint index, GLint size, GLenum type, GLboolean normalized, GLsizei stride, const GLvoid *pointer);
+typedef void(APIENTRYP PFNGLATTACHSHADERPROC)(GLuint program, GLuint shader);
```
FeralChild64, line 119: Would it be feasible to start moving OpenGL stuff to a separate file? I've always felt `sdlmain.cpp` does way too much.

johnnovak: Yeah, I was thinking the same. Most likely I'll do it in my next PR. I'll also get rid of our home-grown OpenGL proc loader and I'll use [glad](https://github.com/Dav1dde/glad) instead (in header-only mode).

src/gui/sdlmain.cpp
```diff
@@ -4761,6 +4761,37 @@ static void init_logger(const CommandLineArguments* args, int argc, char* argv[]
 	loguru::init(argc, argv);
 }
 
+static void maybe_write_primary_config(const CommandLineArguments* args)
```
FeralChild64: Unless I have a good reason, I prefer to pass a reference instead of a pointer - it's safer.

src/gui/sdlmain.cpp
```diff
@@ -4792,6 +4792,52 @@ static void maybe_write_primary_config(const CommandLineArguments* args)
 	}
 }
 
+static std::optional<int> maybe_handle_command_line_output_only_actions(
+        const CommandLineArguments* args, char* argv[])
```
FeralChild64: Again, I prefer a reference instead if pointer. And we don't need the whole `char* argv[]` inside - passing `argv[0]` (program name) should be enough.

src/gui/sdlmain.cpp
```diff
@@ -259,6 +259,109 @@ static void get_opengl_proc_addresses()
...
 }
 
+// A safe wrapper around that returns the default result on failure
+static const char* safe_gl_get_string(const GLenum requested_name,
+                                      const char* default_result = "")
+{
+	// the result points to a static string but can be null
+	const auto result = glGetString(requested_name);
+
+	// the default, howeever, needs to be valid
```
FeralChild64: Spelling mistake: howeever

src/gui/sdlmain.cpp
```diff
@@ -259,6 +259,109 @@ static void get_opengl_proc_addresses()
...
+
+	if (is_shader_compiled) {
+		return shader;
+	} else {
+		glDeleteShader(shader);
+		return 0;
+	}
+}
+
+static bool LoadGLShaders(const std::string& source, GLuint* vertex, GLuint* fragment)
```
FeralChild64: Is it possible to change _vertex_ and _fragment_ into to references?

FeralChild64:
Just a couple remarks from my side.

I have a question: what's the rationale behind removing the SDL 2.0.18 high-DPI workaround? Is it confirmed this is not needed anymore?

src/gui/sdlmain.cpp
johnnovak: Good point, done.

src/gui/sdlmain.cpp
johnnovak: Good points, done both.

src/gui/sdlmain.cpp
johnnovak: Good point, done.

johnnovak:
> I have a question: what's the rationale behind removing the SDL 2.0.18 high-DPI workaround? Is it confirmed this is not needed anymore?

Oops, my bad. That's still needed, I misinterpreted that version check.

Fixed that up and removed some more version checks in new commits (but kept the code intact).

johnnovak:
Thanks for the review and Linux testing @FeralChild64 🏅

PR #4390, opened 2025-06-21
SDL main cleanup, part 2 — OpenGL modernisation (some of it)

johnnovak:
# Description

The SDLpain saga continues; now we're putting the OpenGL stuff into a somewhat better shape, whether it wants it or not:

- _"We target OpenGL 2.1+ currently, so shaders support is always available. Removal of the deprecated fixed-function pipeline also simplifies future adoption of the modern OpenGL 3.x+ Core profile."_ (@GranMinigun's words; this is a continuation of salvaging his [abandoned work](https://github.com/dosbox-staging/dosbox-staging/issues/3971))
- Remove NPOT texture support—same reason, every GPU supports NPOT textures for at least the last 15 years now.
- Better error reporting when shader compilation failed.
- Use the [Glad 2](https://github.com/Dav1dde/glad) OpenGL loader instead of our home-grown solution

I'm doing things incrementally, in small steps that I can individually test and reason about. Going straight to @GranMinigun's final result would be super error prone (I've tried and I just can't, plus the code has diverged too much since his work). I can only do it this way. Plus this gives us bisectability to pinpoint where something went wrong. Probably there will be 50+ further commits across 3-4 PRs or more until we arrive at more or less the final state (OpenGL and SDL texture backends living in their own files, and sdlmain communicating them via an abstract `RendererBackend` interface). Just don't wanna raise huge PRs and I need some pause between these 😅 

@weirddan455 @shermp @FeralChild64 could you please do a quick check on Linux that I haven't broken anything.

As a reminder, I'm doing all this because I wanna add multi-pass shader support, better shader management, and lots of other goodies. But paying back tech debt (which is substantial) needs to happens first—I don't wanna add to the mess, this code is almost completely unmaintainable at this point.

I'm crediting our Russian when I'm taking a more substantial snippet of his work.

Next up:

- Simplifying the `presentation_mode` and `vsync` settings; that will make further refactorings a lot easier.

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/issues/3971

# Release notes

N/A

# Manual testing

- Start with texture and opengl output.
- Change output at runtime.
- Enter/exit fullscreen.
- Make sure shaders work & auto-shader switching works.

- [ ] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:


johnnovak:
@Torinde Added the OpenGL 2.1+ requirements here (now it's an explicit requirement in the init code):
https://github.com/dosbox-staging/dosbox-staging/pull/4390/commits/7a14434c8a6da69486af1daac5faf3dd8301942d

FeralChild64:
> could you please do a quick check on Linux that I haven't broken anything.

the compilation is broken:

```
../src/gui/sdlmain.cpp:48:10: fatal error: glad/gl.h: No such file or directory
   48 | #include "glad/gl.h"
      |          ^~~~~~~~~~~
compilation terminated.
```

Torinde:
Is it straightforward to fallback to texture output automatically if opengl is unavailable? (with a LOG warning, of course)

johnnovak:
> > could you please do a quick check on Linux that I haven't broken anything.
> 
> the compilation is broken:
> 
> ```
> ../src/gui/sdlmain.cpp:48:10: fatal error: glad/gl.h: No such file or directory
>    48 | #include "glad/gl.h"
>       |          ^~~~~~~~~~~
> compilation terminated.
> ```

I've added new files, so delete the CMake build folder and configure the build from scratch.

Can't be broken if it builds on CI 😆

johnnovak:
> Is it straightforward to fallback to texture output automatically if opengl is unavailable? (with a LOG warning, of course)

We've always done that. The problem is buggy/broken drivers that report success but then only display a black screen.

FeralChild64:
@johnnovak 

> Can't be broken if it builds on CI 

image

weirddan455:
Ha, that broken CI job is falling back to Meson. It has some autobuild script that tries CMake first but without vcpkg. That fails because there is no iir package in Ubuntu so then it finds the Meson files and runs that.

I'm guessing @FeralChild64 is using Meson as well. This made several changes to CMake but no changes to Meson so I'm not surprised it doesn't compile. At the very least `src/libs/glad/include/` needs to be added to the include path.

This builds with CMake for me and didn't immediately blow up on me when I tried a few games. I am getting a new warning though: `OPENGL: Failed getting double buffering status: OpenGL error: 00000502`

src/gui/sdlmain.cpp
```diff
@@ -340,26 +219,156 @@ static GLuint build_shader_gl(GLenum type, const std::string& source)
...
 		}
-		glDeleteShader(vertex_out);
 	}
-	return false;
+
+	if (!is_program_linked) {
+		glDeleteProgram(shader_program);
+		return 0;
+	}
+
```
weirddan455, line 279: Traced the new warning I'm getting `OPENGL: Failed getting double buffering status: OpenGL error: 00000502` back to here. First bad commit is 1d751f521216c9130ce2c8bd46f8f401095e5526

Fix is after this block (we know the program is linked) add `glUseProgram(shader_program);`. This is the way it was before refactoring this into a separate function. After the refactor, you wait to `glUseProgram()` until the function returns but I believe several of the `gl*` functions following this depend on the new program being in use.

johnnovak: Cheers, that's good feedback. Yeah, I'm trying to slowly untangle all this mess and not do too much in any single commit. Will fix that today.

johnnovak:
> Ha, that broken CI job is falling back to Meson. It has some autobuild script that tries CMake first but without vcpkg. That fails because there is no iir package in Ubuntu so then it finds the Meson files and runs that.

It won't have a chance to fall back for much longer 😆 

> I'm guessing @FeralChild64 is using Meson as well. This made several changes to CMake but no changes to Meson so I'm not surprised it doesn't compile. At the very least `src/libs/glad/include/` needs to be added to the include path.

Yeah, I can do that now, that's easy enough. But the days of Meson are numbered 😆

src/gui/sdlmain.cpp
johnnovak: Fixed.

johnnovak:
@FeralChild64 Made a feeble attempt for adding Meson support. At least `meson setup` completes without errors now, but I cannot build with Meson even on `master`, so fixing any potential Meson problems is an exercise for the reader... 😛

weirddan455:
> @FeralChild64 Made a feeble attempt for adding Meson support. At least `meson setup` completes without errors now, but I cannot build with Meson even on `master`, so fixing any potential Meson problems is an exercise for the reader... 😛

You almost got it. It needs this one extra parameter and then it compiles and works:

```
diff --git a/src/libs/glad/meson.build b/src/libs/glad/meson.build
index bffdb8a35..d0d3fbb8d 100644
--- a/src/libs/glad/meson.build
+++ b/src/libs/glad/meson.build
@@ -1,4 +1,4 @@
-libglad = static_library('glad', ['src/gl.c'])
+libglad = static_library('glad', ['src/gl.c'], include_directories: 'include')
 
 libglad_dep = declare_dependency(
   link_with: libglad,
```

src/gui/sdlmain.cpp
```diff
@@ -340,26 +219,158 @@ static GLuint build_shader_gl(GLenum type, const std::string& source)
...
+	glUseProgram(sdl.opengl.program_object);
```
weirddan455: This needs to be `glUseProgram(shader_program)`. The global has not been updated yet.

johnnovak: Huh, yeah. I was really sleepy when I did it 😆

johnnovak:
> You almost got it. It needs this one extra parameter and then it compiles and works:

Cheers, I've fixed that up 🎉 Teamwork 😂

FeralChild64:
> could you please do a quick check on Linux that I haven't broken anything.

Quickly checked, no breakages for me.

src/gui/sdlmain.cpp
```diff
@@ -1526,18 +1526,38 @@ static SDL_Window* set_window_mode(const RenderingBackend rendering_backend,
...
 				LOG_ERR("OPENGL: Failed requesting an sRGB framebuffer: %s",
 				        SDL_GetError());
 			}
+
+			// Explicitly request an OpenGL 2.1 compatibility context
+			SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+			SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 1);
+
+			SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK,
+								SDL_GL_CONTEXT_PROFILE_COMPATIBILITY);
```
weirddan455: This might break Raspberry Pi. I believe some Raspberry Pi models only support OpenGL ES so I'm not sure what effect forcing desktop OpenGL will have.

src/gui/sdlmain.cpp
```diff
@@ -1569,8 +1569,11 @@ static SDL_Window* set_window_mode(const RenderingBackend rendering_backend,
 
 		const auto pos = get_initial_window_position_or_default(default_val);
 
-		assert(sdl.window == nullptr); // enusre we don't leak
-		sdl.window = SDL_CreateWindow("", pos.x, pos.y, width, height, flags);
+		// ensure we don't leak
+		assert(sdl.window == nullptr);
+
+		sdl.window = SDL_CreateWindow(
+		        "DOSBox Staging", pos.x, pos.y, width, height, flags);
```
weirddan455: The title should get overwritten by a later call to `GFX_RefreshTitle()`. It probably doesn't hurt to set it to something here as well. I'd recommend using the constant `DOSBOX_NAME` we have defined in `dosbox.h`.

src/gui/sdlmain.cpp
```diff
@@ -1582,8 +1585,10 @@ static SDL_Window* set_window_mode(const RenderingBackend rendering_backend,
 			// window, try again without it.
 			flags &= ~SDL_WINDOW_OPENGL;
 
-			sdl.window = SDL_CreateWindow("", pos.x, pos.y, width, height, flags);
+			sdl.window = SDL_CreateWindow(
+			        "DOSBox Staging", pos.x, pos.y, width, height, flags);
```
weirddan455: `DOSBOX_NAME` here as well.

src/gui/sdlmain.cpp
johnnovak: I think it's fine:

> Neither the RPi 3 nor RPi4 will give you an OpenGL 3 context because AFAIK neither has full hardware support for it, the maximum for OpenGL is 2.1 and for OpenGLES is 3.2 (though Mesa 19.3.2 only supports up to 3.1).
([source](https://forums.raspberrypi.com/viewtopic.php?t=294077))

> Currently Pi 4 is limited to OpenGL 2.1
([source](https://raspberrypi.stackexchange.com/questions/119154/opengl-3-3-on-raspberry-pi-4-8gb))

> I’ve managed to get it working with OpenGL 2.1.
> I had to add compatibility fixes for my shaders, to support #version 120
([source](https://discourse.glfw.org/t/glfw-on-rpi-with-opengl/1215/3))

(☝🏻 we do have `#version 120` in all our shaders already)

---

Technically, RPi4 and up supports OpenGL ES 3.1 and Vulkan 1.0 ([source](https://www.raspberrypi.com/products/raspberry-pi-4-model-b/specifications/)), but asking for OpenGL 2.1 should work.

As per the last reference above, OpenGL 2.1 works on RPi3 even. I honestly don't care what happens on RPi models earlier than that 😆 I don't think they can even run Staging, probably they're way too weak.

I have an RPi4 I specifically bought for Staging testing, I should set it up and try it, but I'm 99% sure this is fine.

Ideal state in the future: we'll have an OpenGL 3.3+ core profile backend, and maybe an OpenGL ES 3.1 backend. The differences should be fairly minimal.

Only targeting OpenGL ES 3.1 is a bad idea because it's not "normal" to use that on a desktop. Driver support might be buggy/spotty/reduced performance... or fine, who knows. OpenGL ES is not commonly used in desktop apps, it's mostly for mobile.

weirddan455: Ok, fair enough. I have a Pi 3 I've been meaning to test Staging on someday. Even that one is pretty weak. We can re-visit this later if it's problematic.

johnnovak: Yeah, I'm a bit in the "ask for forgiveness later" camp with this, but of course I don't want to outright break Pi support if I can easily avoid it, yeah. Explicit OpenGL ES support is a different kettle of fish. It's not clear to me what we'd _actually_ gain by asking for a 3.3+ core profile given our use of OpenGL is so minimal. Probably nothing, so we can just stay on 2.1 indefinitely.

Even if technically drivers in 3.3+ core profile mode could be "more efficient" (how do you measure that, right?), our usage is so minimal that I fail to see why it would matter at all.

johnnovak:
> I've reviewed everything and it looks good. Confirmed Meson builds and the OpenGL warning I was getting is gone. Just a couple minor issues I pointed out but you can merge when ready.

Thanks for the testing, man. I'll address those, do one last sanity check on Windows, and then merge time!

src/gui/sdlmain.cpp
weirddan455: Whatever is most compatible in practice I guess is what we should use. OpenGL 3.3 Core removes a bunch of old stuff from the spec like the fixed-function pipeline (which you've removed from our code in this PR so now it's possible for us to request it).

In theory drivers could remove support for fixed-function and such and *only* work with the newer Core profiles. For now if 2.1 works everywhere, we can just keep using it.

Torinde: Then the .conf description should be "OpenGL 2.1 drivers" instead of "OpenGL 2.1**+** drivers"

johnnovak: Well, "OpenGL 2.1 capable" is probably the best wording.

johnnovak: Final version:

> The 'texture' backend is a last resort fallback for graphics cards without functional OpenGL 2.1 support (this is extremely rare).

src/gui/sdlmain.cpp
johnnovak: Yeah it doesn't matter too much, just setting it to something else than an empty string seemed marginally better.

PR #4396, opened 2025-06-26
Add several missing checks to the CMake build system + other fixes

FeralChild64:
# Description

Added missing checks to set various `HAVE_*` defines during the compilation.
Removed `C_HAS_BUILTIN_EXPECT` - it is not used anywhere.
Small cleanups.

## Related issues

# Manual testing

Started few apps (Norton Commander, DOS Navigator), Windows 3.11, games (DOOM, Warcraft), played a little.

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:


CMakeLists.txt
```diff
@@ -21,12 +21,12 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
 # `include/checks.h` for details.
 if(MSVC)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4244")
+  add_compile_options("/wd4244")
   
   # Disable cl/clangcl warnings about insecure C functions
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-conversion -Wno-narrowing")
+  add_compile_options("-Wno-conversion" "-Wno-narrowing")
```
johnnovak, line 29: Nice, this is definitely the better way to do this 👍🏻

CMakeLists.txt
```diff
@@ -208,26 +223,32 @@ set(C_CORESERVICES   "${DOSBOX_PLATFORM_MACOS}")
...
 check_symbol_exists(
   pthread_jit_write_protect_np "pthread.h" HAVE_PTHREAD_WRITE_PROTECT_NP
 )
 check_symbol_exists(
   sys_icache_invalidate "libkern/OSCacheControl.h" HAVE_SYS_ICACHE_INVALIDATE
 )
 
-set(_USE_MATH_DEFINES ON)
+# TODO: Detect PAGESIZE using contrib/pagesize/*_.pagesize.c
+
```
johnnovak: I'm looking at the [commit notes](https://github.com/dosbox-staging/dosbox-staging/commit/7e20f6e401956a7a308f1b3462294d7ac9fa5db8) of pagesize.c and I'm scratching my head. The CPU page size is 4k on x86 and ARM. According to the comment, we'd get the wrong page size when cross-compiling for Alpha and Itanium processors... which we don't do, neither do we support running Staging on expensive server hardware, lol 😆 

Then what if we just use 4k on these platforms as well? What's the harm? Probably nothing as this seems to be some performance optimisation heuristics to align memory block sizes with the page size or something. Whether that results in anything beneficial, who knows...

I'd say we just kill this with Meson.

Torinde: x86, ARM and RISC-V _support_ 4K, but they also support other sizes, so maybe that code aims to align with whatever is enabled when running?

Of the currently relevant platforms only LoongArch default is 16K, but I assume that's not a priority for Staging.

johnnovak: 4k is standard on common platforms. The reason is unclear why we need to be super precise about this. It's an optimisation thing that might yield zero benefit, so I'd say kill it.

Search for `host_pagesize`, you'll find some ppc64le specific workarounds for 64k page sizes... We don't support test ppc64le, we don't test it, etc. so this whole thing is pointless.

johnnovak (approved):
Very nice work @FeralChild64, great stuff 💯 Just a few minor comments, feel free to merge when you're ready.

CMakeLists.txt
FeralChild64: Right, just the optimization, but for the performance-critical code.

I've hardcoded the value to 4K, but added a runtime check to warn if different size is detected - we don't know what size is going to be a typical value 10 years from now.

FeralChild64:
@johnnovak I didn't rebase yet, as the `main` branch has Meson compilation broken (not only for me, the CI has the same problem) - I've tried to fix itfor half an hour, but I honestly don't know what's wrong; the `include_directories` directive does not seem to work as expected.

Screenshot-Error

FeralChild64:
@johnnovak Surprise - it seems the macOS uses page size of 16K :D Setting it back to DRAFT for now, I need to investigate this, do some benchmarks, etc.

CMakeLists.txt
FeralChild64: @johnnovak Surprise - it seems the macOS uses page size of 16K :D It seems like this is a typical value for ARM64. And Linux can use different sizes, depending on the platform and Kernel configuration.

[edit] Tested on macOS using SiSoft Sandra 2001 CPU benchmark and Quake timedemo, there is hardly any difference. I have hardcoded the values to 16K (ARM 64-bit) or 4K (rest of the world). Not much sense to spend too much time on it.

weirddan455:
The fix to `src/libs/glad/meson.build` got reverted in @johnnovak 's PR between the time I tested/approved it and when it got merged. It needs this extra parameter: https://github.com/dosbox-staging/dosbox-staging/pull/4390#issuecomment-3005021158

FeralChild64:
@weirddan455 Thanks - added the fix, inserted it before my 1st commit.

johnnovak:
> The fix to `src/libs/glad/meson.build` got reverted in @johnnovak 's PR between the time I tested/approved it and when it got merged. It needs this extra parameter: [#4390 (comment)](https://github.com/dosbox-staging/dosbox-staging/pull/4390#issuecomment-3005021158)

My bad, I worked on that PR on two different machines and overwrote my earlier fix with a force push. Sorry guys.

CMakeLists.txt
johnnovak: It's great you looked into it a bit more deeply @FeralChild64 🎉 Yeah these optimisation attempts need to be backed up by real world profiler metrics, otherwise it becomes just "cargo culting" to keep them alive forever and they just add complexity.

Hardcoding to 16k on macOS ARM is the appropriate solution, so good call. 👍🏻

PR #4403, opened 2025-06-30
Make mixer master gain volume atomic

weirddan455:
# Description

Fixes a thread sanitizer warning.

# Manual testing

I was able to confirm the sanitizer warning and fix with Meson so it's not dependent on the CMake PR.

Built with `meson setup -Db_sanitize=thread`. Running the command `mixer master 75` at runtime produced the data race warning. This is now fixed.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:


FeralChild64:
I confirm the problem is now fixed - but some CI jobs are failing now.

weirddan455:
> I confirm the problem is now fixed - but some CI jobs are failing now.

Yeah, I was just looking into that. We need to be linking with `libatomic` on Linux. Meson has some auto-detection code for this but it's wrong. We've gotten away without it so far because so far we've only been using `std::atomic` with primitive types and the compilers have been inlining everything.

Here I'm using `std:atomic<AudioFrame>`. GCC 12 and Clang 16 are in-lining but Clang 14 is not and is failing to build. It also fails on Meson because the auto-detection is wrong.

I think we just need to always link against `libatomic` because if we were to use `std::atomic` on some more complex type, GCC and newer Clangs could still generate calls into it.

johnnovak:
@weirddan455 I like our usage of `std::atomic` for primitive types, that's lightweight and understandable. However, I have mixed feelings about using it with non-primitive types like this. I kinda agree with this comment; probably using `std::mutex` is a better solution for non-primitive types. The moment you need to perform anything more complex than a load/store (so multiple operations), you need a mutex anyway.

https://stackoverflow.com/a/33501669

As the commenter on this comment mentions, using atomics like this might be better suited for lock-free algorithms, but I really don't want us to be messing around with that stuff (hard to get right and wrap your head around it, and we don't need that level of extreme performance I'd say).

weirddan455:
I got CI to pass. We should have been linking to `libatomic` all along for Linux to be at feature parity with Windows and Mac. It's part of the GCC runtime library.

> However, I have mixed feelings about using it with non-primitive types like this. I kinda agree with this comment; probably using std::mutex is a better solution for non-primitive types. The moment you need to perform anything more complex than a load/store (so multiple operations), you need a mutex anyway.

I usually like to stick with primitive types but this is a reasonable exception IMO. I'm never planning on doing anything more than a load/store. It's not part of some lock-free algorithm and it's not managing synchronization of other data.

It's also just 2 32-bit floats. The compiler optimizes it to a single 64-bit read/write. I was considering bit-packing them into a single 64-bit variable myself but this was easier and more readable.

johnnovak:
> I got CI to pass. We should have been linking to `libatomic` all along for Linux to be at feature parity with Windows and Mac. It's part of the GCC runtime library.

Fair enough, seems like a good idea to do that.

> I usually like to stick with primitive types but this is a reasonable exception IMO. I'm never planning on doing anything more than a load/store. It's not part of some lock-free algorithm and it's not managing synchronization of other data.

Yeah, I was thinking it can be borderline alright for cases like this when there's just a struct with a few primitive members in it. 

> It's also just 2 32-bit floats. The compiler optimizes it to a single 64-bit read/write. I was considering bit-packing them into a single 64-bit variable myself but this was easier and more readable.

That's fine until we restrict ourselves to simple cases, like this one, then use mutexes for anything more complex. Just reading that `std::atomic` can use any copyable type gave me a pause 😄

src/hardware/mixer.cpp
```diff
@@ -152,7 +152,7 @@ struct MixerSettings {
 
 	// The default master gain is -6dB (50% volume) to minimise the change
 	// for clipping.
-	AudioFrame master_gain = {Minus6db, Minus6db};
```
johnnovak, line 155: While you're at it, can you fix my typo please? "minimise the chance". Cheers 😄

johnnovak (approved):
Nice job @weirddan455 🎉

src/hardware/mixer.cpp
weirddan455: fixed

Issue #4157, opened 2025-01-17
Render to console


unxed:
### Are you using the latest Dosbox-Staging Version?

- [x] I have checked releases and am using the latest release.

### What Operating System are you using?

Linux x86_64

### Is your feature request related to a problem? Please describe.

Can not interact to text mode DOS apps from Linux console.

### Describe the solution you'd like

With modern advanced terminals, like [kitty](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), it is possible to have full information about keyboard events in console apps. It is also possible to color any screen cell in any TrueColor color. May it be possible now to make DOSBox render text EGA/VGA/... modes to console as modern Wine versions do for text mode apps?

### Code of Conduct & Contributing Guidelines

- [x] Yes, I agree.

FeralChild64:
Possible? Probably yes, if I remember correctly _DOSEmu_ could do something like this. But - IMHO - such level of host OS integration is rather out of scope of _DOSBox Staging_.

johnnovak:
Way out of scope and conflicts with our [project goals](https://www.dosbox-staging.org/about/#goals).

Issue #1450, opened 2021-12-20
Terminal mode for DOS development (ANSI passthrough or curses text mode)


promovicz:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Is your feature request related to a problem? Please describe.

My main use case for dosbox is for development. For that I currently use a dosbox window, but it would be extremely convenient if dosbox would support terminal text mode. An application that uses only console/ANSI output could potentially run without curses by simple passthrough and be part of the terminal experience like any other POSIX command.

### Describe the solution you'd like

I imagine it would be possible to specify a DOS command on the dosbox commandline:

```
Usage: dosbox [command...]
```

A simple implementation could use curses to display any of the already supported character cell text modes. Smart screen management could be used to make the "DOS subwindow" appear in the right place in the terminal for the output to appear contigouos.

An alternative and more advanced implementation could pass through console/ANSI output so that DOS application output can be printed on the terminal as if it were POSIX output. Such an implementation could also engage graphics mode when required, opening an SDL window for the VGA while showing either a blank screen or text mode contents on the terminal.

For simple development use it might be sufficient to have only an ANSI pass-through mode, but it is likely that many tools will break under such circumstances.

```
$ dosbox dir
Volume in drive C is C_DRIVE     
 Directory of  c:\bc31\*.*

.            <DIR>       1-23-18   2:07 
..           <DIR>      12-20-21  23:33 
BGI          <DIR>       1-22-18   8:15 
BIN          <DIR>       2-02-18   8:19 
CLASSLIB     <DIR>       1-22-18   8:16 
CRTL         <DIR>       1-22-18   8:15 
DOC          <DIR>       1-22-18   8:17 
EXAMPLES     <DIR>       1-22-18   8:19 
INCLUDE      <DIR>       1-23-18  20:27 
LIB          <DIR>       1-23-18   2:07 
TVISION      <DIR>       1-22-18   8:19 
bc31p1.mrk           0   1-23-18   2:07 
bc32p2.mrk           0   1-23-18   2:07 
filelist.doc    121614   1-22-18   8:17 
readme           56343   1-22-18   8:18 
readme.com        4217   1-22-18   8:18 
        182,174 bytes in 5 files and 11 dirs    260,096 bytes allocated
    262,111,744 bytes free
```

### Describe alternatives you've considered

None. This is just an idea I am trying to share.

### Code of Conduct & Contributing Guidelines

[...]

PR #4414, opened 2025-07-07
Cleanup some code related to machine types

FeralChild64:
# Description

This is a cleanup only PR:

- variables `machine` and `svgaCard` are now `enum class`
- replaced macros to check for machine type with inline functions
- cleaned up code to put string signature into SVGA BIOS ROM 
- reduced number of allowed macOS warnings

Motivation: I intend to look once again into possibility of porting DOSBox-X S3 improvement (most likely with S3 model selection), but I want to do some cleanup first.

# Manual testing

- checked that Windows 3.11 still works with S3 964 1.41B5 drivers
- checked that my Windows 98 install still boots in SVGA resolution and does not detect any new hardware
- checked that DOSBox started with `--noprimaryconf --set machine=hercules` starts in Hercules mode (also checked other machine types)

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/include/dosbox.h
+++ b/include/dosbox.h
@@ -70,8 +70,10 @@ enum class MachineType {
 	// PC XT with Hercules
 	Hercules,
 
-	// PC XT with CGA
-	Cga,
+	// PC XT with CGA and monochrome monitor
+	CgaMono,
+	// PC XT with CGA and color monitor
+	CgaColor,
 	// IBM PCjr
 	Pcjr,
 	// Tandy 1000
@@ -98,8 +100,6 @@ enum class SvgaType {
 
 extern MachineType machine;
 extern SvgaType    svga_type;
-
-extern bool mono_cga;
 
 inline bool is_machine_svga() {
 	return svga_type != SvgaType::None;
@@ -129,8 +129,17 @@ inline bool is_machine_pcjr_or_tandy() {
 	return is_machine_pcjr() || is_machine_tandy();
 }
 
-inline bool is_machine_cga() {
-	return machine == MachineType::Cga;
+inline bool is_machine_cga_mono() {
+	return machine == MachineType::CgaMono;
+}
+
+inline bool is_machine_cga_color() {
+	return machine == MachineType::CgaColor;
+}
+
+inline bool is_machine_cga()
+{
+	return is_machine_cga_mono() || is_machine_cga_color();
 }
 
 inline bool is_machine_cga_or_better() {
--- a/include/dosbox.h
+++ b/include/dosbox.h
@@ -63,37 +63,85 @@ int64_t DOSBOX_GetTicksDone();
 void DOSBOX_SetTicksDone(const int64_t ticks_done);
 void DOSBOX_SetTicksScheduled(const int64_t ticks_scheduled);
 
-enum SVGACards {
-	SVGA_None,
-	SVGA_S3Trio,
-	SVGA_TsengET4K,
-	SVGA_TsengET3K,
-	SVGA_ParadisePVGA1A
-}; 
+enum class MachineType {
+	// Value not set yet
+	None,
 
-extern SVGACards svgaCard;
-extern bool mono_cga;
+	// PC XT with Hercules
+	Hercules,
 
-enum MachineType {
-	// In rough age-order: Hercules is the oldest and VGA is the newest
-	// (Tandy started out as a clone of the PCjr, so PCjr came first)
-	MCH_INVALID = 0,
-	MCH_HERC    = 1 << 0,
-	MCH_CGA     = 1 << 1,
-	MCH_TANDY   = 1 << 2,
-	MCH_PCJR    = 1 << 3,
-	MCH_EGA     = 1 << 4,
-	MCH_VGA     = 1 << 5,
+	// PC XT with CGA
+	Cga,
+	// IBM PCjr
+	Pcjr,
+	// Tandy 1000
+	Tandy,
+
+	// PC AT with EGA
+	Ega,
+	// PC AT with VGA or SVGA
+	Vga
 };
 
-extern MachineType machine;
+enum class SvgaType {
+	// Non-SVGA card
+	None,
+	// S3 Graphics series (emulated card is mostly Trio64 compatible)
+	S3,
+	// Tseng Labs ET3000
+	TsengEt3k,
+	// Tseng Labs ET4000
+	TsengEt4k,
+	// Paradise Systems PVGA1A
+	Paradise
+}; 
 
-inline bool is_machine(const int type) {
-	return machine & type;
+extern MachineType machine;
+extern SvgaType    svga_type;
+
+extern bool mono_cga;
+
+inline bool is_machine_svga() {
+	return svga_type != SvgaType::None;
 }
-#define IS_TANDY_ARCH ((machine==MCH_TANDY) || (machine==MCH_PCJR))
-#define IS_EGAVGA_ARCH ((machine==MCH_EGA) || (machine==MCH_VGA))
-#define IS_VGA_ARCH (machine==MCH_VGA)
+
+inline bool is_machine_vga_or_better() {
+	return machine == MachineType::Vga;
+}
+
+inline bool is_machine_ega() {
+	return machine == MachineType::Ega;	
+}
+
+inline bool is_machine_ega_or_better() {
+	return is_machine_ega() || is_machine_vga_or_better();
+}
+
+inline bool is_machine_tandy() {
+	return machine == MachineType::Tandy;
+}
+
+inline bool is_machine_pcjr() {
+	return machine == MachineType::Pcjr;
+}
+
+inline bool is_machine_pcjr_or_tandy() {
+	return is_machine_pcjr() || is_machine_tandy();
+}
+
+inline bool is_machine_cga() {
+	return machine == MachineType::Cga;
+}
+
+inline bool is_machine_cga_or_better() {
+	return is_machine_cga() ||
+               is_machine_pcjr_or_tandy() ||
+               is_machine_ega_or_better();
+}
+
+inline bool is_machine_hercules() {
+	return machine == MachineType::Hercules;
+}
 
 #ifndef DOSBOX_LOGGING_H
 #include "logging.h"
--- a/src/dosbox.cpp
+++ b/src/dosbox.cpp
@@ -135,8 +135,6 @@ void DOSBOX_SetTicksScheduled(const int64_t ticks_scheduled)
 	ticks.scheduled = ticks_scheduled;
 }
 
-bool mono_cga = false;
-
 void Null_Init([[maybe_unused]] Section *sec) {
 	// do nothing
 }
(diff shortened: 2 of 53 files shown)


include/dosbox.h
```diff
@@ -63,37 +63,34 @@ int64_t DOSBOX_GetTicksDone();
...
+	Ega,
+	Vga
 };
 
+enum class SvgaType {
+	None, // non-SVGA card
+	S3,
+	TsengET4K,
+	TsengET3K,
+	ParadisePVGA1A
```
johnnovak: How about just `Tseng3k`, `Tseng4k` and `Paradise`? Especially the name of the Paradise card seems a bit too detailed for an enum. We can specify the full name and other details about these cards in the enum comments. That would be handy too (e.g., get a summarized info from dosdays.co.uk).

At the very least should be `SvgaType::TsengEt4k` etc for consistent capitalisation. 

Also, why not `S3Trio`? We are emulating the S3 Trio series:

http://www.dosdays.co.uk/topics/Manufacturers/s3_trio.php

There's also the S3 Vision range that we're not emulating.

src/hardware/vga_draw.cpp
```diff
@@ -1250,9 +1252,13 @@ static void VGA_VerticalTimer(uint32_t /*val*/)
 	case M_TANDY_TEXT:
 	case M_CGA_TEXT_COMPOSITE:
 	case M_HERC_TEXT:
-		if (machine==MCH_HERC) vga.draw.linear_mask = 0xfff; // 1 page
-		else if (IS_EGAVGA_ARCH) vga.draw.linear_mask = 0x7fff; // 8 pages
-		else vga.draw.linear_mask = 0x3fff; // CGA, Tandy 4 pages
+		if (machine == MachineType::Hercules) {
+			vga.draw.linear_mask = 0xfff; // 1 page
+		} else if (IS_EGAVGA_ARCH) {
```
johnnovak: Forgot about this one.

include/dosbox.h
```diff
@@ -63,37 +63,34 @@ int64_t DOSBOX_GetTicksDone();
...
-	MCH_CGA     = 1 << 1,
-	MCH_TANDY   = 1 << 2,
-	MCH_PCJR    = 1 << 3,
-	MCH_EGA     = 1 << 4,
-	MCH_VGA     = 1 << 5,
+	Invalid,
+	Hercules,
+	Cga,
+	Tandy,
+	PcJr,
```
johnnovak: I'd just go `Pcjr`. It's stylized as `PCjr` originally 😄 I'm using this form already in `GraphicsStandard::Pcjr`.

src/ints/bios.cpp
```diff
@@ -955,7 +956,7 @@ static Bitu INT15_Handler(void) {
 		LOG(LOG_BIOS,LOG_ERROR)("INT15:Unknown call %4X",reg_ax);
 		reg_ah=0x86;
 		CALLBACK_SCF(true);
-		if ((IS_EGAVGA_ARCH) || (machine==MCH_CGA)) {
+		if ((IS_EGAVGA_ARCH) || (machine == MachineType::Cga)) {
```
johnnovak: Missed one.

src/hardware/voodoo.cpp
```diff
@@ -7887,8 +7887,8 @@ static void voodoo_init(Section* sec)
 	auto* section = dynamic_cast<Section_prop*>(sec);
 
 	// Only activate on SVGA machines and when requested
-	if (machine != MCH_VGA || svgaCard == SVGA_None || !section ||
-	    !section->Get_bool("voodoo")) {
+	if (svga_type == SvgaType::None ||
```
johnnovak: Are you sure this is safe and equivalent to the previous check?

src/ints/bios_disk.cpp
```diff
@@ -336,17 +336,22 @@ static Bitu INT13_DiskHandler(void) {
 			 */
 			if (any_images && driveInactive(drivenum)) {
 				/* driveInactive sets carry flag if the specified drive is not available */
-				if ((machine==MCH_CGA) || (machine==MCH_PCJR)) {
+				if ((machine == MachineType::Cga) ||
+				    (machine == MachineType::PcJr)) {
```
johnnovak: Inner parens are redundant.

src/ints/int10_char.cpp
```diff
@@ -326,20 +328,22 @@ void INT10_SetActivePage(uint8_t page) {
 
 void INT10_SetCursorShape(uint8_t first,uint8_t last) {
 	real_writew(BIOSMEM_SEG,BIOSMEM_CURSOR_TYPE,last|(first<<8));
-	if (machine==MCH_CGA || IS_TANDY_ARCH) goto dowrite;
+	if (machine == MachineType::Cga || IS_TANDY_ARCH) goto dowrite;
```
johnnovak: Missed one.

include/dosbox.h
```diff
@@ -63,37 +63,34 @@ int64_t DOSBOX_GetTicksDone();
...
-	SVGA_ParadisePVGA1A
-}; 
-
-extern SVGACards svgaCard;
-extern bool mono_cga;
-
-enum MachineType {
+enum class MachineType {
 	// In rough age-order: Hercules is the oldest and VGA is the newest
 	// (Tandy started out as a clone of the PCjr, so PCjr came first)
```
Torinde: Per this comment - PCjr should be listed before Tandy, but in the code list afterwards the order is reversed?

johnnovak: Yeah, we can say "In order of release" and reshuffle the order now. I suggest the following:

- Invalid
- Cga
- Hercules
- Pcjr
- Tandy
- Ega
- Vga

include/dosbox.h
FeralChild64: `TsengET3K `-> `TsengEt3k` and `TsengET4K` -> `TsengEt4k` changed. I don't want to drop `Et`, as there are some internal variables/functions in the Tseng implementation which starts from chip name.

> Also, why not `S3Trio`

Because in the future I hope to introduce a separate config option and variable to specify the S3 chipset. In fact, this is why I am doing this rework in a separate PR, I want to reduce the size of the main PR.

BTW, `INT10_SetupRomMemory` function currently sets the `S3 86C764` name in the ROM :D

include/dosbox.h
FeralChild64: OK, changed.

src/ints/bios.cpp
FeralChild64: I didn't forget - I haven't finished the rework yet.

src/ints/bios_disk.cpp
FeralChild64: OK, fixed.

src/ints/int10_char.cpp
FeralChild64: I didn't forget - I haven't finished the rework yet.

include/dosbox.h
FeralChild64: I have sorted hem based on the hardware capabilities and compatibility:

Hercules (alone in the group - we are not emulating Hercules Plus or anything similar)
CGA < PCjr < Tandy
CGA < EGA < VGA

src/hardware/voodoo.cpp
FeralChild64: Yes, see how the variables are set in `dosbox.cpp`. The goal here is to only activate Voodoo if SVGA is selected.

BTW, I didn't do any self-review yet (I intend to do this some other day, definitely when I'm not tired), so I'm not 100% sure there are no mistakes.

src/hardware/vga_draw.cpp
FeralChild64: I didn't handle this macro yet, the rework is not finished.

include/dosbox.h
johnnovak: Yeah, that's actually better than the "strict" by release year ordering.

include/dosbox.h
```diff
@@ -63,37 +63,84 @@ int64_t DOSBOX_GetTicksDone();
...
+}
+
+inline bool is_machine_cga() {
+	return machine == MachineType::Cga;
+}
+
+inline bool is_machine_cga_or_better() {
+	return is_machine_cga() ||
+               is_machine_pcjr_or_tandy() ||
+               is_machine_ega_or_better();
```
johnnovak, line 148: I was wondering why you didn't just use the `>=` operator and was thinking about performance, but probably a non-issue and the compiler will optimise these into a `>=` check anyway.

src/dosbox.cpp
```diff
@@ -45,8 +45,9 @@
 #include "video.h"
 
 bool shutdown_requested = false;
-MachineType machine;
-SVGACards svgaCard;
+
+MachineType machine   = MachineType::Invalid;
```
johnnovak: How about `MachineType::None` for consistency? Better yet, get rid of it in a separate commit (the init code needs it to be "none" for a while, so it might be tricky).

src/dosbox.cpp
```diff
@@ -467,40 +468,49 @@ void DOSBOX_SetMachineTypeFromConfig(Section_prop* section)
...
-		machine = MCH_HERC;
-	} else if (mtype == "ega") {
-		machine = MCH_EGA;
-	} else if (mtype == "svga_s3") {
-		svgaCard = SVGA_S3Trio;
-	} else if (mtype == "vesa_nolfb") {
-		svgaCard = SVGA_S3Trio;
+
+	if (machine_str == "cga") {
+		machine = MachineType::Cga;
```
johnnovak: `MachineType::CgaMono` and dump the flag? It's a nasty hack.

src/dosbox.cpp
Torinde: I hope someday `MachineType::None` does [console redirection](https://en.wikipedia.org/wiki/Conventional_memory#Additional_memory):

> It was also possible to use console redirection (either by specifying an alternative console device like AUX: when initially invoking COMMAND.COM or by using CTTY later on) to direct output to and receive input from a dumb terminal or another computer running a terminal emulator. Assuming the System BIOS still permitted the machine to boot (which is often the case at least with BIOSes for embedded PCs), the video card in a so called headless computer could then be removed completely, and the system could provide a total of 960 KB of continuous DOS memory for programs to load.

Basically a "no videocard" option - output going to whichever is the most straightforward to implement:
- DOSBox window
- AUX/COM for nullmodem to localhost Telnet or similar

(Omitting more complex options like [this](https://github.com/dosbox-staging/dosbox-staging/issues/1450) and [that](https://github.com/dosbox-staging/dosbox-staging/issues/4157))

johnnovak: @Torinde you're getting into scope creep territory again 😆 Read again what our mission goal is (hint: games).

This will not happen.

Please re-read our about page periodically 😄 

This is not an experimental DOS project or a PCem-like project. This is about games. Please make such comments in other projects that are interested in experimental features like this...

include/dosbox.h
FeralChild64: Yes - plus I'm concerned that someone might try to reorder the enum class once again.

src/dosbox.cpp
FeralChild64: Done. Also renamed `MachineType::Cga` to `MachineType::CgaColor`, otherwise `is_machine_cga()` name was a little ambiguous.

src/dosbox.cpp
FeralChild64: Changed `Init` to `None`.

Getting rid of it (if we don't want to use `std::optional` or any separate variable) indeed might be tricky, as `shader_manager.cpp` uses this value and I'm not sure what the init flow looks like (and which use cases are possible) - I'm probably not the right person for this job :)

johnnovak: Thanks, I might do it later when I rework the shader loading in the coming months.

Issue #3206, opened 2023-12-14
`fullresolution = original` is broken


vol-2:
# See distilled repro steps here: https://github.com/dosbox-staging/dosbox-staging/issues/3206#issuecomment-1890187685

### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.1 and nightly build 12-11-23

### What Operating System are you using?

None

### Relevant hardware info

Intel i7, GTX 980

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When I have the output set to "original," Staging does not always output the resolution 1:1. When I am playing a game that is 320x200, it properly displays it at 640x400 (line doubled), but above that the results are usually wrong and inconsistent at best. It will often downscale the image. It scales 320x200 to 800x500 and 640x480 to 533x400.

With the nightly build, it no longer displays 640x400 properly either, it now shifts the whole screen over to the right about an inch, and the image is not beyond the borders of my monitor because I can use my screen shifting buttons to drag it left but the image is not there.

If I try to display at 800x600, it will sometimes just crash when I exit the game back to Staging.

I have since last testing (with Nick over on Discord) that I can alt-return in and out of the game and it will sometimes suddenly start to display correctly. This only works if I do not have the Taskbar on my desktop as that will be present on top of the game image if it is. Luckily, I have a second monitor for DOSBox (a 1600x1200 CRT), and I don't need the Taskbar open on that one.

So far, I have only tried the alt-return in and out with the nightly build, not with 0.80.1.

I asked ToastyX over on the Custom Resolution Utility forum if he had any insight as to why SDL and Staging might be doing this, and if there was anything I could do to fix it and he said:

"The details of how the resolutions are defined only matter to the graphics driver. The only thing that might matter is if DOSBox Staging is looking at the preferred/native resolution to determine how to scale and you have the first detailed resolution set to 640x400, it might be trying to scale to that but messing up somewhere. This definitely sounds like some bug with DOSBox Staging."

This happens if I set 800x600 to the first (or only) resolution in my detailed resolution list, so I don't think that's it. 🤷‍♂️

It doesn't seem to matter if I set things to texture or opengl. Aspect is set to off, shader is none.

I tried Blood and Quake as examples, both behaved the exact same way.

[...]

PR #4415, opened 2025-07-08
Add forced borderless fullscreen mode & general window config settings rework

johnnovak:
# Description

Graphics card drivers might opt-out of fullscreen optimisations on Windows when they decide the performance loss would be too great compared to exclusive fullscreen. This means you'll get exclusive fullscreen even if you asked for borderless fullscreen at the SDL API level (`SDL_WINDOW_FULLSCREEN_DESKTOP` flag).

This is exactly what happens on my Windows 10 / Nvidia RTX 3060 setup when I use 120 Hz refresh rate at 4k. With 4k@60 Hz, borderless mode is working; 1080p@120 Hz is fine too, but 4k@120 Hz is deemed too ambitious by the GPU gods 😆, so I'm always getting exclusive fullscreen! This is a pain during DOSBox dev because I cannot freely Alt-Tab anymore, and when playing undemanding adventure games and RPGs where the performance drop simply doesn't matter.

The SDL devs have [explicitly stated](https://github.com/libsdl-org/SDL/issues/12791) they won't add workarounds for forcing borderless mode, so we'll need to take matters into our own hands.

There are a few different ways to go about this, one is in that ticket, another one [here](https://stackoverflow.com/questions/79614593/forcing-non-exclusive-borderless-fullscreen-with-sdl3-opengl-on-windows-11) (which doesn't work, btw; I've tried). In the end, I opted for only using the SDL API to mimic borderless mode, so no low-level messing around with Windows APIs. Basically, creating a window 1px wider than the desktop resolution does the trick, as the trigger for borderless fullscreen is having a window with exactly the same dimensions as the desktop size and with disabled window decorations. Then it's in the driver's hands: it might keep the window and exclude it from the compositor for more direct access, or it might enter the old legacy fullscreen mode.

Borderless fullscreen optimisation is a Windows-only thing, making `forced-borderless` a Windows-only option. macOS doesn't have the equivalent of borderless fullscreen, so it doesn't make any sense there. I imagine Linux users have a lot more fine-grained control over the driver/OS behaviour, so they can force "always borderless" elsewhere.

Other changes:

- renamed `windowresolution` to `window_size` while I was it
- `fullresolution = desktop` has become `fullscreen_mode = standard`
- `fullresolution = WxH` has been dropped. I don't think there's any good use case for that. We still support `fullscreen_mode = original` for using Staging with a CRT monitor, but it's a bit janky (a concession for the guy who raised [this ticket](https://github.com/dosbox-staging/dosbox-staging/issues/3206); Phil also made a [video](https://www.youtube.com/watch?v=pQ-DWEY3E7M) about it, he uses `original` with a bunch of custom modes)
- the usual SDL cleanups

On a related note, I can only get perfectly smooths 2D scrolling in games like Pinball Dreams in exclusive fullscreen mode. That's 100% smooth. In forced borderless fullscreen, the scrolling is only about 85-90% smooth. Still very good, but I can notice a _tiny_ jitter that manifests more as a slight blur rather than actual jerky motion.

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/3206

# Release notes

- The `windowresolution` setting has been renamed to `window_size`.
- The `fullresolution` setting has been renamed to `fullscreen_mode`, and `fullresolution = desktop` is now available as `fullscreen_mode = standard`. The option to explicitly set screen resolutions in `WxH` format has been removed as it's not useful neither on modern hardware nor when using DOSBox Staging with a real CRT monitor.
- A new `forced-borderless` fullscreen mode has been introduced to force borderless operation on Windows systems where the graphics card driver opts out of fullscreen optimisations, resulting in exclusive fullscreen which makes Alt+Tabbing very cumbersome. See the setting's help for further info (run `fullscreen_mode /?` from the DOS prompt).

# Manual testing

- Tested `forced-borderless` results in borderless fullscreen that's visually indistinguishable from "true" borderless mode (quick Alt-Tabbing works, etc). You can only meaningfully test this on Windows systems that opt-out of fullscreen optimisations when using certain driver settings (e.g., Windows 10, Nvidia RTX 3060, 120 Hz over DisplayPort to a 4k screen in my case).
- Tested that maximising the window then toggling forced borderless mode works correctly.
- Tested toggling "show desktop" with Win+D works correctly when in forced borderless mode.
- Confirmed the `forced-borderless` option doesn't show up on macOS and it cannot be set.
- Confirmed the `original` option works as before on Windows and macOS.
- Confirmed `window_position`, `window_size`, and `window_decorations` still work.
- Confirmed the mapper still works and resizing the window with the mapper active works.

## New `fullscreen_mode` description

image

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/sdl_mapper.cpp
+++ b/src/gui/sdl_mapper.cpp
@@ -1880,7 +1880,6 @@ static void SetActiveEvent(CEvent * event) {
 }
 
 extern SDL_Window* GFX_GetWindow();
-extern void GFX_UpdateDisplayDimensions(int width, int height);
 
 static void DrawButtons() {
 	SDL_SetRenderDrawColor(mapper.renderer,
@@ -2760,8 +2759,6 @@ void BIND_MappingEvents()
 			 */
 			if ((event.window.event == SDL_WINDOWEVENT_RESIZED) ||
 			    (event.window.event == SDL_WINDOWEVENT_RESTORED)) {
-				GFX_UpdateDisplayDimensions(event.window.data1,
-				                            event.window.data2);
 				SDL_RenderSetLogicalSize(mapper.renderer, 640, 480);
 				mapper.redraw = true;
 			}
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3452,6 +3452,55 @@ static void restart_hotkey_handler([[maybe_unused]] bool pressed)
 	DOSBOX_Restart();
 }
 
+static void set_fullscreen_mode()
+{
+	const std::string fullscreen_mode_pref = get_sdl_section()->Get_string(
+	        "fullscreen_mode");
+
+	auto set_desktop_mode = [&] {
+		sdl.desktop.full.fixed  = true;
+		sdl.desktop.full.width  = 0;
+		sdl.desktop.full.height = 0;
+	};
+
+	if (fullscreen_mode_pref == "desktop") {
+		set_desktop_mode();
+
+	} else if (fullscreen_mode_pref == "original") {
+		sdl.desktop.full.fixed  = false;
+		sdl.desktop.full.width  = 0;
+		sdl.desktop.full.height = 0;
+
+	} else {
+		const auto parts = split_with_empties(fullscreen_mode_pref, 'x');
+		if (parts.size() == 2) {
+			const auto maybe_width  = parse_int(parts[0]);
+			const auto maybe_height = parse_int(parts[1]);
+
+			if (maybe_width && maybe_height) {
+				sdl.desktop.full.fixed  = true;
+				sdl.desktop.full.width  = *maybe_width;
+				sdl.desktop.full.height = *maybe_height;
+
+				maybe_limit_requested_resolution(
+				        sdl.desktop.full.width,
+				        sdl.desktop.full.height,
+				        "fullscreen");
+
+				// Success
+				return;
+			}
+		}
+
+		// Failure
+		LOG_WARNING("DISPLAY: Invalid 'fullscreen_mode' setting: '%s'; using 'desktop'",
+		            fullscreen_mode_pref.c_str());
+
+		set_desktop_mode();
+		set_section_property_value("sdl", "fullscreen_mode", "desktop");
+	}
+}
+
 static void read_gui_config(Section* sec)
 {
 	sec->AddDestroyFunction(&shutdown_gui);
@@ -3477,39 +3526,7 @@ static void read_gui_config(Section* sec)
 	// Assume focus on startup
 	apply_active_settings();
 
-	std::string fullresolution = section->Get_string("fullresolution");
-
-	sdl.desktop.full.fixed  = false;
-	sdl.desktop.full.width  = 0;
-	sdl.desktop.full.height = 0;
-
-	if (!fullresolution.empty()) {
-		lowcase(fullresolution); // so x and X are allowed
-		                         //
-		if (fullresolution != "original") {
-			sdl.desktop.full.fixed = true;
-
-			if (fullresolution != "desktop") { // desktop uses 0x0,
-				                           // below sets a
-				                           // custom WxH
-				std::vector<std::string> dimensions =
-				        split_with_empties(fullresolution, 'x');
-
-				if (dimensions.size() == 2) {
-					sdl.desktop.full.width =
-					        parse_int(dimensions[0]).value_or(0);
-
-					sdl.desktop.full.height =
-					        parse_int(dimensions[1]).value_or(0);
-
-					maybe_limit_requested_resolution(
-					        sdl.desktop.full.width,
-					        sdl.desktop.full.height,
-					        "fullscreen");
-				}
-			}
-		}
-	}
+	set_fullscreen_mode();
 
 	const std::string host_rate_pref = section->Get_string("host_rate");
 	if (host_rate_pref == "auto") {
@@ -4466,7 +4483,10 @@ static void init_sdl_config_section()
 	auto pbool = sdl_sec->Add_bool("fullscreen", always, false);
 	pbool->Set_help("Start in fullscreen mode ('off' by default).");
 
-	pstring = sdl_sec->Add_string("fullresolution", always, "desktop");
+	pstring = sdl_sec->Add_string("fullresolution", deprecated, "");
+	pstring->Set_help("Please use 'fullscreen_mode' instead.");
+
+	pstring = sdl_sec->Add_string("fullscreen_mode", always, "desktop");
 	pstring->Set_help(
 	        "What resolution to use for fullscreen: 'original', 'desktop'\n"
 	        "or a fixed size, e.g. 1024x768 ('desktop' by default).");
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -3191,22 +3191,30 @@ static void save_window_size(const int w, const int h)
 //     not take into account scaling or aspect correction.
 //  - 'sdl.desktop.window', with the refined size.
 //
-static void setup_window_sizes_from_conf(const char* windowresolution_val,
-                                         const bool wants_aspect_ratio_correction)
+static void setup_window_sizes_from_conf(const bool wants_aspect_ratio_correction)
 {
+
+	const auto window_size_pref = [&]() {
+		const auto legacy_pref = get_sdl_section()->Get_string("windowresolution");
+		if (!legacy_pref.empty()) {
+			set_section_property_value("sdl", "windowresolution", "");
+			set_section_property_value("sdl", "window_size", legacy_pref);
+		}
+		return get_sdl_section()->Get_string("window_size");
+	}();
+
 	// Get the coarse resolution from the users setting, and adjust
 	// refined scaling mode if an exact resolution is desired.
-	const std::string pref = windowresolution_val;
 	SDL_Point coarse_size  = FallbackWindowSize;
 
-	sdl.use_exact_window_resolution = pref.find('x') != std::string::npos;
+	sdl.use_exact_window_resolution = window_size_pref.find('x') != std::string::npos;
 
 	if (sdl.use_exact_window_resolution) {
-		coarse_size = parse_window_resolution_from_conf(pref);
+		coarse_size = parse_window_resolution_from_conf(window_size_pref);
 	} else {
 		const auto desktop = get_desktop_size();
 
-		coarse_size = window_bounds_from_label(pref, desktop);
+		coarse_size = window_bounds_from_label(window_size_pref, desktop);
 	}
 
 	// Save the coarse bounds in the SDL struct for future sizing events
@@ -3372,8 +3380,7 @@ static void set_output(Section* sec, const bool wants_aspect_ratio_correction)
 	setup_initial_window_position_from_conf(
 	        section->Get_string("window_position"));
 
-	setup_window_sizes_from_conf(section->Get_string("window_size").c_str(),
-	                             wants_aspect_ratio_correction);
+	setup_window_sizes_from_conf(wants_aspect_ratio_correction);
 
 #if C_OPENGL
 	if (sdl.want_rendering_backend == RenderingBackend::OpenGl) {
@@ -4436,7 +4443,8 @@ static void init_sdl_config_section()
 
 	constexpr auto always     = Property::Changeable::Always;
 	constexpr auto deprecated = Property::Changeable::Deprecated;
-	constexpr auto on_start   = Property::Changeable::OnlyAtStart;
+	constexpr auto deprecated_but_allowed = Property::Changeable::DeprecatedButAllowed;
+	constexpr auto on_start = Property::Changeable::OnlyAtStart;
 
 #if C_OPENGL
 	const std::string default_output = "opengl";
@@ -4533,8 +4541,10 @@ static void init_sdl_config_section()
 #endif
 	                     "original"});
 
-	pstring = sdl_sec->Add_string("windowresolution", deprecated, "");
-	pstring->Set_help("Renamed to 'window_size'.");
+	pstring = sdl_sec->Add_string("windowresolution", deprecated_but_allowed, "");
+	pstring->Set_help(
+	        "The 'windowresolution' setting is deprecated but still accepted;\n"
+	        "please use 'window_size' instead.");
 
 	pstring = sdl_sec->Add_string("window_size", on_start, "default");
 	pstring->Set_help(
--- a/src/gui/sdlmain.cpp
+++ b/src/gui/sdlmain.cpp
@@ -1646,25 +1646,25 @@ static SDL_Window* set_window_mode(const RenderingBackend rendering_backend,
 	 * although it has its own issues.
 	 */
 	if (fullscreen) {
-		SDL_DisplayMode displayMode;
-		SDL_GetWindowDisplayMode(sdl.window, &displayMode);
+		SDL_DisplayMode display_mode;
+		SDL_GetWindowDisplayMode(sdl.window, &display_mode);
 
-		displayMode.w = width;
-		displayMode.h = height;
+		display_mode.w = width;
+		display_mode.h = height;
 
 		// TODO pixels or logical unit?
-		if (SDL_SetWindowDisplayMode(sdl.window, &displayMode) != 0) {
+		if (SDL_SetWindowDisplayMode(sdl.window, &display_mode) != 0) {
 			LOG_WARNING("SDL: Failed setting fullscreen mode to %dx%d at %d Hz",
-			            displayMode.w,
-			            displayMode.h,
-			            displayMode.refresh_rate);
+			            display_mode.w,
+			            display_mode.h,
+			            display_mode.refresh_rate);
 		}
 		SDL_SetWindowFullscreen(sdl.window,
-		                        sdl.desktop.fullscreen.display_res
+		                        sdl.desktop.fullscreen.mode == FullscreenMode::Standard
 		                                ? enum_val(SDL_WINDOW_FULLSCREEN_DESKTOP)
 		                                : enum_val(SDL_WINDOW_FULLSCREEN));
 	} else {
-		// we're switching down from fullscreen, so let SDL use the
+		// We're switching down from fullscreen, so let SDL use the
 		// previously-set window size
 		if (sdl.desktop.switching_fullscreen) {
 			SDL_SetWindowFullscreen(sdl.window, 0);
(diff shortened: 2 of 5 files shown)


johnnovak:
I'd appreciate some basic Linux testing @weirddan455 , @FeralChild64 , or whoever has time.

src/gui/sdlmain.cpp
```diff
@@ -3703,16 +3737,6 @@ void GFX_RegenerateWindow(Section* sec)
 
 static void handle_video_resize(int width, int height)
```
FeralChild64: I'm getting a compilation warning here:

```
../src/gui/sdlmain.cpp: In function ‘void handle_video_resize(int, int)’:
../src/gui/sdlmain.cpp:3738:48: warning: unused parameter ‘height’ [-Wunused-parameter]
```

src/gui/sdl_mapper.cpp
```diff
@@ -2760,8 +2760,6 @@ void BIND_MappingEvents()
 			 */
 			if ((event.window.event == SDL_WINDOWEVENT_RESIZED) ||
 			    (event.window.event == SDL_WINDOWEVENT_RESTORED)) {
-				GFX_UpdateDisplayDimensions(event.window.data1,
```
FeralChild64, line 2763: Are we sure this is Android-only? What about convertible laptops?

FeralChild64:
Tried on my Linux system (both with X11 and with Wayland) - I see no difference between `fullscreen_mode = exclusive` and `fullscreen_mode = borderless`. Quake `timedemo demo1` benchmark gives me the same results.

weirddan455:
> The fullresolution setting has been renamed to fullscreen_mode, and fullresolution = original is now available as fullscreen_mode = exclusive

You missed the purpose of `fullresolution = original` because these do different things. In `main` with `fullscreen = original`, it tries to match the resolution to what DOS is running at. On my machine, it starts at 800x600 at the DOS prompt at then switches to 640x480 when I launch a game. I don't think my monitor supports 320x200 so this is as close as it can get.

On this branch, `fullscreen_mode = exclusive` just runs at 1920x1080 all the time but it switches the refresh rate for some reason from my desktop 144hz to 85hz so it takes my monitor a good 5 seconds to switch modes when I alt-tab. It's pretty useless in this state. In `main` I can see a usecase for CRT monitors but this PR breaks that.

src/gui/sdl_mapper.cpp
johnnovak: Reworked it a bit so it only affects the special Android workaround for the mapper. I'm reluctant to keep workaround's that we cannot test and are unsure about. I think it's best to get rid of them and only reintroduce them if people report problems if we have no way to test them.

johnnovak:
> You missed the purpose of `fullresolution = original` because these do different things. In `main` with `fullscreen = original`, it tries to match the resolution to what DOS is running at. On my machine, it starts at 800x600 at the DOS prompt at then switches to 640x480 when I launch a game. 

You're right, I thought people used `fullresolution = 640x480` or something, but then watched [Phil's video on the subject](https://www.youtube.com/results?search_query=phil%27s%20computer%20labs%20crt%20dosbox), and turns out they use `original` with some custom screen modes that DOSBox will then just pick up automatically.

> I don't think my monitor supports 320x200 so this is as close as it can get.

All sub-350 line VGA modes are double scanned. So VGA cards effectively _emulate_ 320x200; it's 640x400 at the analog signal level. The analog signal output by the VGA card is _100% identical at the electrical level_ for mode 13h 320x200 and 640x400 (if you 2x upscale the same 320x200 image). There is zero difference from the monitor's perspective.

This was probably a cost-cutting measure to only emulate earlier 320x200 CGA/EGA modes as fixed 31 kHz monitors are just cheaper to manufacture and nobody cared about "15 kHz monitor scanlines" in 1987 that you can observe on CGA & EGA monitors 😄 "Hardware 2x upscaling" 320x200 content appeared sharper, and sharper was better. All their cared about was sharpness at IBM because PCs were meant for business applications first and foremost, people staring at spreadsheets and text all day.

So yeah, when you use Staging like this with a true CRT monitor, we already do the double scanning for VGA modes correctly (thanks to _yours truly_ 😎). So people only need to set up a custom 640x400 @ 70 Hz resolution to run most VGA games that use the standard 320x200 13h mode.

> On this branch, `fullscreen_mode = exclusive` just runs at 1920x1080 all the time but it switches the refresh rate for some reason from my desktop 144hz to 85hz so it takes my monitor a good 5 seconds to switch modes when I alt-tab. It's pretty useless in this state. In `main` I can see a usecase for CRT monitors but this PR breaks that.

Yeah, so I've removed support for that. I thought it might be handy to provide that as well, but maybe let's revisit that when there's an actual use case for it. Moreover, although exclusive mode "works" on macOS, it triggers a _different_ fullscreen mode than the "standard" macOS fullscreen with all the OS-level shortcuts disabled. Wtf?! There's an SDL3 ticket about this to improve the situation, but let's just put all that aside while we're still on SDL2... 😅 

Exclusive fullscreen doesn't make much sense on Linux either, at least not on X11 that basically always uses "borderless fullscreen" if I'm correct. I've read with Wayland there is a way to force Windows-style "exclusive" fullscreen where the whole windowing system is effectively bypassed, but let's not complicate things for no good reason.

So, in the end, I've made the following changes based on your feedback:

- `desktop` has been renamed to `standard` (as `borderless` was a bad name; there's "just fullscreen" on X11, and macOS doesn't have the "borderless" concept at all; it's "just fullscreen" there too)
- `original` has been restored
- The `WxH` option is gone. It's just a complication as nobody really wants to use non-native resolutions on modern flat screens, and apparently `original` is the ticket for real CRT monitors, so I see no use case for it.

Thanks for the testing and the feedback ❤️

src/gui/sdlmain.cpp
johnnovak: I think it's fine now.

weirddan455:
> All sub-350 line VGA modes are double scanned. So VGA cards effectively emulate 320x200; it's 640x400 at the analog signal level. The analog signal output by the VGA card is 100% identical at the electrical level for mode 13h 320x200 and 640x400 (if you 2x upscale the same 320x200 image). There is zero difference from the monitor's perspective.

I'm getting 640x480, not 640x400. Again, not sure if my monitor supports 640x400. 640x480 was a more typical resolution from the Windows 95 era onward.


> So, in the end, I've made the following changes based on your feedback:
> 
> - `desktop` has been renamed to `standard` (as `borderless` was a bad name; there's "just fullscreen" on X11, and macOS doesn't have the "borderless" concept at all; it's "just fullscreen" there too)

There's a distinction. For one, alt-tabbing a borderless fullscreen window shows the fullscreen application in the background while you can have other windows overlayed on top of it. Exclusive fullscreen will minimize on alt-tab. Exclusive fullscreen can also change the resolution and refresh rate while borderless fullscreen always operates at whatever the desktop uses.

Example using borderless:

borderless

Borderless is also a common term. Ex: This is Dota 2's video settings which has a native Linux port.

dota-fullscreen

> - `original` has been restored
> - The `WxH` option is gone. It's just a complication as nobody really wants to use non-native resolutions on modern flat screens, and apparently `original` is the ticket for real CRT monitors, so I see no use case for it.
> 
> Thanks for the testing and the feedback ❤️ 

That does address my major concern with this. I'll try to give it a proper code review later today.

johnnovak:
>There's a distinction. For one, alt-tabbing a borderless fullscreen window shows the fullscreen application in the background while you can have other windows overlayed on top of it. Exclusive fullscreen will minimize on alt-tab. Exclusive fullscreen can also change the resolution and refresh rate while borderless fullscreen always operates at whatever the desktop uses.

@weirddan455 That's cool, so it looks like modern Windows and Linux have more or less the same borderless and exclusive concepts. On Windows I think it's "more exclusive", though; entering/exiting exclusive mode usually incurs that 2-3 second delay that you also noticed.

But macOS throws a spanner in the works. To summarise:

- `SDL_WINDOW_FULLSCREEN_DESKTOP` gives us borderless mode on Windows and Linux which I'm calling now "standard". Alt+Tabbing leaves the Staging windows in the background, so other windows can appear on top of it. But this mode on macOS results in the regular macOS fullscreen mode: you enter fullscreen via a short animation, then Cmd+Tabbing fades the Staging window out and it then it disappears. You can't have other windows on top of it in this mode!

- `SDL_WINDOW_FULLSCREEN` results in exclusive mode both on Windows and Linux, apparently (although it behaves slightly differently on Linux as you described). But this mode on macOS is kinda weird: there is _no_ enter/exit animation, and _all_ OS-level shortcuts are disabled, including controlling the volume, Cmd+Tabbing, everything. You can only exit from it by pressing the fullscreen toggle hotkey again or quitting the emulator. It effectively has forced full keyboard capture that I'm not even able to disable (I've tried). This is not a very useful mode as implemented in SDL2.

Now, SDL3 has improved the macOS `SDL_WINDOW_FULLSCREEN` behaviour, see [this ticket](https://github.com/libsdl-org/SDL/issues/7776). The change makes this fullscreen more usable on macOS, but the concepts are completely different than on Windows and Linux: this _is_ exclusive fullscreen on macOS in the sense that you are able to set a different refresh rate and resolution, but it appears instantenaouly versus `SDL_WINDOW_FULLSCREEN_DESKTOP` which has the short animation... and in _neither_ mode can you display other windows on top of the Staging window. So hence I'm saying, macOS simply doesn't have the same "borderless fullscreen" concept that Linux and Windows have where you can seamlessly Alt-Tab to other windows and they appear on top of you borderless fullscreen window. This mode of fullscreen operation just doesn't exist in macOS land (but maybe you can "emulate" it, somehow... but I'm not going there, no interest from my side).

Phew... So hope it makes sense why I'm trying to get away from the `borderless` name. Having `borderless` and `exclusive` (if needed) setting options would make sense on Linux and Windows as the concepts are well-known and SDL supports them well. On macOS, however, `borderless` is a syntax error. 

So I introduced `standard` for `SDL_WINDOW_FULLSCREEN_DESKTOP` operation. But then that's a bit weird for Windows and Linux users.

I guess I don't wanna have completely different option names per platform... Maybe just call the default fullscreen mode `auto`. So just rename `standard` to `auto`. Then, over time, we might introduce platform-specific options, but `auto` would be the "most common" option on each individual platform (borderless on Linux & Windows, and "just fullscreen" on macOS). We're kinda doing the same thing with `mididevice = auto`.

Any better suggestions?

PS: I thought this was going to be a simple 2-hour job, but there are always complications... 😅 Here we go, two days later we're still debating names... 😆

Grounded0:
Comment from bleeding-heart macOS user.

I've noticed that there's two full screen modes:

1. Clicking the green button in the window title bar to enter the window into full screen. In this mode you can still interact with the window title bar, Mac menu bar and the Dock which is bad since I'm looking for 100 % mouse capture for games naturally.

2. Entering full screen through hotkey assigned in mapper. This mode hits the sweet spot since it has 100 % mouse capture but still allows you to jump between your virtual desktops and other full screen apps through trackpad gestures and key shortcuts.

The most major complaint macOS users had was how badly SDL1 handled this in DOSBox by throwing us into exclusive mode with weird resolutions that looked terrible on old non-HiDPI displays. Window resizing also wasn't possible back then like it has been since 0.75.0:

dosbox-staging v0.75.0 showcase - resizable window

https://www.youtube.com/watch?v=9J2U-qFxE1E

Screenshot 2025-07-10 at 1 42 14

Grounded0:
SheepShaver has disabled the title bar activated full screen mode:

Screenshot 2025-07-10 at 2 24 35

src/gui/sdlmain.cpp
```diff
@@ -4473,9 +4473,12 @@ static void init_sdl_config_section()
 	        "What resolution to use for fullscreen: 'original', 'desktop'\n"
 	        "or a fixed size, e.g. 1024x768 ('desktop' by default).");
 
-	pstring = sdl_sec->Add_string("windowresolution", on_start, "default");
+	pstring = sdl_sec->Add_string("windowresolution", deprecated, "");
+	pstring->Set_help("Renamed to 'window_size'.");
```
weirddan455: I'm not sure if deprecated is working as expected. You've used this previously for the CPU cycles rework where the old setting continues to work. I tested here and `windowresolution` does not work so it's really more obsoleted than deprecated.

src/gui/sdlmain.cpp
```diff
@@ -4466,7 +4483,10 @@ static void init_sdl_config_section()
 	auto pbool = sdl_sec->Add_bool("fullscreen", always, false);
 	pbool->Set_help("Start in fullscreen mode ('off' by default).");
 
-	pstring = sdl_sec->Add_string("fullresolution", always, "desktop");
+	pstring = sdl_sec->Add_string("fullresolution", deprecated, "");
+	pstring->Set_help("Please use 'fullscreen_mode' instead.");
```
weirddan455: Same comment here regarding deprecation.

johnnovak:
@Grounded0 I think I'm getting your second option even when clicking on the green button in the window's title. In other words, clicking the green button and using Cmd+Enter results in the same outcome for me.

> Entering full screen through hotkey assigned in mapper. This mode hits the sweet spot since it has 100 % mouse capture but still allows you to jump between your virtual desktops and other full screen apps through trackpad gestures and key shortcuts.

Are you aware we've added the `keyboard_capture` option recently? That will disable all these OS-level shortcuts but Cmd+Tab will continue to work.

See here: https://github.com/dosbox-staging/dosbox-staging/pull/4338

Grounded0:
> @Grounded0 I think I'm getting your second option even when clicking on the green button in the window's title. In other words, clicking the green button and using Cmd+Enter results in the same outcome for me.

Interesting. Maybe they have tweaked how that button behaves with later macOS releases. I'm on 12.6.5. Here's what I'm getting when pressing the green button:

Screenshot 2025-07-10 at 3 51 24

> Are you aware we've added the keyboard_capture option recently? That will disable all these OS-level shortcuts but Cmd+Tab will continue to work.

Yup. This is a great addition for most people and probably the last macOS annoyance that's now fixed.

johnnovak:
> Maybe they have tweaked how that button behaves with later macOS releases. I'm on 12.6.5.

I'm on 15.5 Sequoia. So older macOS versions add another dimension to this problem... 🤯 💥 All in all, we can tune the behaviour only for the latest OS versions because that's what we're running in the team, then whatever happens on earlier versions will be the behaviour... We can make small tweaks based on user feedback for earlier macOS versions, though, if it's worthwhile and not too hacky or complicated.

weirddan455 (approved):
The code looks fine to me. I just question how settings deprecation is supposed to work.

> Here we go, two days later we're still debating names

I don't want to split hairs over naming. What you have is fine IMO and makes sense now that you've explained the MacOS situation.

src/gui/sdlmain.cpp
johnnovak: @weirddan455 
Yeah, things got complicated with `cycles`. Supporting both `cycles` and `cpu_*` was rather special and it took me a while to get it right.

But it turns out this was much simpler. See my last few commits, and this is how it behaves in action when setting `fullresolution = desktop`:

image

src/gui/sdlmain.cpp
johnnovak: I've added added support for both.

johnnovak:
> The code looks fine to me. I just question how settings deprecation is supposed to work.

Added support for old settings in two more commits, it was rather easy.
 
> I don't want to split hairs over naming. What you have is fine IMO and makes sense now that you've explained the MacOS situation.

Cheers, I'll just leave the `standard` option name then. We can always rename it later, but this will do for a good while. Thanks for the thorough review 👍🏻

johnnovak:
@vol-2 Let me know if `fullscreen_mode = original` works as intended with your CRT.

vol-2:
> @vol-2 Let me know if `fullscreen_mode = original` works as intended with your CRT.

ok. Assuming that we are talking about the 0.82.2 release that was posted 3 weeks ago:

I am testing on 3 separate systems, 2 with Windows 10 and 1 with Windows 11. Each one uses a different GPU and CRT. All seem to behave exactly the same as far as I can tell. I'm using texturenb and direct3d11

In terms of the image, things are basically the same as they were before. When a program loads, the position on the screen and the scale of the image is loaded incorrectly, but I can "correct" it by alt-returning out and back into full screen. 

It certainly isn't more broken. However, I believe that it performs a little slower in Texture modes than it use to. I will have to do a Quake benchmark to be sure, but it feels like things bog down a little now occasionally, whereas I never noticed that before. 

Opengl seems faster, but Opengl always applies the bilinear filtering and, at 320x200, the bilinear filtering makes everything look like oatmeal and is unplayable. The sharp option does nothing I can tell. I believe the openglnb did this back when it existed, although I feel like I had less success with it in terms of scaling compatibility.

That did get me wondering about [the shaders that require (or use to require?)](https://github.com/dosbox-staging/dosbox-shaders/blob/master/README.md) openglnb though. I wonder if those can still be used by people now, or did they remake them for the new staging?

Going back and using texturenb with the opengl render option seems to produce more consistent and faster results. It appears that simply opening the viewport and starting a game doesn't always require exiting and entering fullscreen anymore. If I open DOSBox and launch Quake directly in 800x600, it seems to open to the correct resolution just fine (it hasn't done that before). However, when  I exit Quake, the Viewport is distorted, and requires exiting and re-entering fullscreen to fix it.
I also still need to do this if I change resolutions. For example, if I open the Quake video settings and change the current resolution from 800x600 to 320x200, it will not change the viewport size, but it also doesn't display a tiny 320x200 image in the center like I would expect. Instead, it displays a scaled up portion of a 320x200 image without altering the viewport at all. Exiting and re-entering fullscreen fixes the image as it usually does.

If I start a program that is 320x200 by default, it doesn't start correctly, it appears off to the left of the screen. Again, I can fix the position by exiting and re-entering fullscreen.
PXL_20250711_152631781

When I exit the program though, the Viewport appears to the right of the screen.


PXL_20250711_153126692



Messing around with Windows settings on my own in a variety of other situations, I have found that properly displayed resolutions on a CRT are mostly effected by the refresh rate. If I have trouble getting the CRT to display a resolution, it's always because Windows switched the screen resolution but not the refresh rate. Going into the advanced display settings an manually switching the refresh rate will fix it for me 100% of the time. I'm not sure if this is something that matters to what you're working with (SDL I believe), but I though it might be worth mentioning.

Please let me know if this all makes sense or if I need to clarify anything. Also if there's anything you want me to try.

johnnovak:
> It certainly isn't more broken.

Thanks @vol-2, that's all I wanted to confirm.

> However, I believe that it performs a little slower in Texture modes than it use to. I will have to do a Quake benchmark to be sure, but it feels like things bog down a little now occasionally, whereas I never noticed that before.

This change has nothing to do with performance. There might be other factors at play, but if you believe something is wrong and can 100% repro it, please raise a new ticket. However, I haven't noticed any performance regressions in our latest releases using the Quake benchmark, so I think it's likely you just confused yourself with something or it's placebo, etc.

> That did get me wondering about [the shaders that require (or use to require?)](https://github.com/dosbox-staging/dosbox-shaders/blob/master/README.md) openglnb though. I wonder if those can still be used by people now, or did they remake them for the new staging?

A lot of information on that page is simply wrong, including the requirement for `openglnb`. That's just not true. AFAIK the guy stopped updating those shaders. Just use our bundled shaders, they're the best.

---

For the rest of your observations, like I said, we're not spending any effort on "perfecting" `fullscreen_mode = original`. It is what it is, a concession for CRT users out there, with bugs and all 😄 Again, I just wanted to make sure it's not twice as bad as before, but you confirmed it's not, so we'll leave it at that.

gobalius:
Wow! Fullscreen, borderless. ALT+TAB wet dream!
Kudos!

johnnovak:
> Wow! Fullscreen, borderless. ALT+TAB wet dream! Kudos!

Enjoy 😎

PR #4418, opened 2025-07-10
Allow 3.5x and 4.5x integer scaling for the adaptive CRT shaders

johnnovak:
# Description

Also disable integer scaling above 5.0x scaling. These enhancements reduce the pillar & letterboxing of the adaptive CRT shaders in fullscreen, especially on 1440p and 4k screens.

I've tested this extensively with all sorts of content and this "relaxed" integer scaling results in no artifacts once we're above 3 rendered lines per emulated scanline (so there's no visible moire).

These are the improvements in fullscreen at 4k (2160p), 1440p, and 1080p, but of course windowed mode benefits too (the jumps are smaller between 3.0x and 5.0x when resizing the window), especially on 1440p or higher. Note 320x200 VGA is covered by the 400-line case due to double scanning. Improvements are marked with 🟢.

### 4k

- 2160 / 200 = 10.8 → now fills the screen vertically 🟢 (was 2000px high before)
- 2160 / 350 = 6.14 → now fills the screen vertically 🟢 (was 2100px high before)
- 2160 / 400 = 5.4  → now fills the screen vertically 🟢 (was 2000px high before)
- 2160 / 480 = 4.5  → now fills the screen vertically 🟢 (was 1920px high before)

### 1440p

- 1440 / 200 = 7.2  → now fills the screen vertically 🟢 (was 1400px high before)
- 1440 / 350 = 4.11 → 1400px high image (same as before)
- 1440 / 400 = 3.6  → 3.5 scaling, 1400px high image 🟢 (was 1200px before) 
- 1440 / 480 = 3.0  → fills the screen (same as before) 

### 1080p

- 1080 / 200 = 5.4  → now fills the screen vertically 🟢 (was 1000px high before)
- no change for other screen modes on 1080p

### Further changes

- Fix display properties logging when changing the window size.
- Log display properties when switching between fullscreen/windowed in forced borderless mode.
- Improve a bunch of display-related config descriptions.
- Remove the outdated conf settings in the unit tests config file (we should really be relying on the defaults as much as possible).

# Release notes

Integer scaling for the adaptive CRT shaders (the `integer_scaling = auto` option) has been refined to reduce pillar and letterboxing. The improvements are especially noticeable on 4k and 1440p monitors: common video modes now completely fill the screen vertically on 4k monitors in fullscreen (no more letterboxing), and 320x200 VGA modes result in minimal letterboxing on 1440p screens. All CGA and EGA modes on 1080p now have no letterboxing in fullscreen either.

At the technical level, adaptive CRT shaders can now use 3.5x and 4.5x vertical scaling factors, too, and integer scaling is disabled for them above 5.0x scaling. These relaxations of the integer scaling rules result in no visual artifacts.

# Manual testing

- Tested edge cases with 3.5x / 4.5x scaling and no integer scaling above 5.0x scaling, such as 100% white screens, black text on 100% white screens, and large areas checkerboard dither patterns (e.g., the default Windows 3.1x background image and The Magnetic Scroll Collection which has a 100% black & white checkerboard background pattern).
- Confirmed the CRT shaders still look good at all scaling factors 😎 
- Tested the integer scaling logic for regressions.
- Verified the updated config descriptions have the correct wrapping & indentation.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


shermp:
Seems to look ok on my 1440p monitor. I don't see anything obviously wrong. But I don't really know what I'm looking for so...

johnnovak:
> Seems to look ok on my 1440p monitor. I don't see anything obviously wrong. But I don't really know what I'm looking for so...

If you wanna see the moire/interference artifacts resulting from non-integer scaling factors when the scanlines are "dense", disable integer scaling and try resizing the window manually. The artifacts are easiest to spot on checkerboard dither patterns at 640x480 or higher.

Examples (these are kinda worst-case scenarios). Make sure to view images at 100% magnification, otherwise the browser will add its own scaling artifacts. Save the images and view them with some image viewer app. Browsers are kinda unpreditable how they scale images...

First - notice the periodic circular colour fringing artifacts both in the horizontal and vertical directions.
Second - notice the vertical periodic wave patterns.

The thing is, most people could disable integer scaling even on 1440p and most games would look fine. But not all, and this aims to be a conservative setting that doesn't allow any artifacts to show up in 99% of the use cases.

image

image

src/gui/render.cpp
```diff
@@ -976,6 +976,79 @@ DosBox::Rect RENDER_CalcRestrictedViewportSizeInPixels(const DosBox::Rect& canva
...
+		case IntegerScalingMode::Off: return draw_size_fit_px;
+
+		case IntegerScalingMode::Auto: {
+			const auto shader_info =
+			        ShaderManager::GetInstance().GetCurrentShaderInfo();
+
+			if (!shader_info.name.empty() && shader_info.is_adaptive) {
+				return calc_vert_integer_scaling_dims_in_pixels();
+			}
+			return draw_size_fit_px;
```
weirddan455: There was a check for OpenGL here that got removed when you moved this function. Both a compile time `#if OPENGL_C` and a runtime check of `sdl.rendering_backend`. Is this intentional and did you check it for correctness when using the `texture` backend?

weirddan455 (approved):
Looks good. Just the one comment where something stood out as different in a refactor commit. I tested texture mode and it seems to work though (integer scaling is disabled when set to auto which I believe is desired behavior since there is no shader).

FeralChild64:
Great idea! Just tested, I really like the new behavior.

src/gui/render.cpp
johnnovak: Yeah, I kinda thought about this and it works because `shader_info.name` is initialised with a blank string in texture mode. But maybe it will misbehave is you start in OpenGL mode then switch to texture. I'll test this, plus it would be better to make the output mode check explicit, no doubt.

johnnovak: Fixed by adding an `GFX_GetRenderingBackend() != RenderingBackend::OpenGl`. I'm already using that in other places in `render.cpp` but forgot about it.

PR #4421, opened 2025-07-12
Reintroduce vcpkg binary caching

shermp:
# Description

We used to cache vcpkg binaries, first using the Github packages NuGet repository, then with the integrated GH actions cache in vcpkg. However when the integrated solution stopped working (it was removed by vcpkg), we have been rebuilding our dependencies for every push/pull request.

This PR adds back caching using the default `files` mechanism, and `actions/cache`.

The cache key I've chosen contains:

- OS/workflow
- CPU architecture
- GH runner ImageOS
- GH runner ImageVersion
- A hash of our vcpkg manifest and overlay files

I believe this combination should be robust, and not get stale. In general, the cache should be invalidated approximately once a week as GH update their runners. Using the runner version in the cache key is important, as vcpkg includes things such as exact compiler versions, cmake version, powershell versions as part of its ABI hash.

The result is even faster CI times. Due to the inconsistent nature of the GH runners, an exact speedup cannot be specified, but it appears to be in the order of 3-8 minutes when restoring the cache.

# Manual testing

- [x] Windows
- [x] macOS
- [x] Linux

This is a CI change, All actions were green before opening the PR.

# Checklist

I have:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,14 +24,14 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  VCPKG_ROOT: ${{ github.workspace }}/vcpkg
+
 jobs:
   build_linux:
     name:    ${{ matrix.conf.name }} ${{ matrix.arch }}
     runs-on: ${{ matrix.conf.os }}${{ matrix.arch == 'arm64' && '-arm' || '' }}
     if:      github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-    env:
-      VCPKG_ROOT: ${{ github.workspace }}/vcpkg
-
     strategy:
       matrix:
 #       TODO enable arm64 builds
@@ -92,6 +92,13 @@ jobs:
           cd $VCPKG_ROOT
           ./bootstrap-vcpkg.sh
 
+      - name: Set up vcpkg cache
+        uses: ./.github/actions/cache-vcpkg
+        with:
+          name_os: linux
+          arch: "${{ matrix.arch }}"
+          compiler: "${{ matrix.conf.cc }}"
+
       - name: Log environment
         run: ./scripts/ci/log-env.sh
 
@@ -124,9 +131,6 @@ jobs:
     name:    Release build
     runs-on: ubuntu-22.04
     if:      github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-    env:
-      VCPKG_ROOT: ${{ github.workspace }}/vcpkg
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -149,6 +153,13 @@ jobs:
         run: |
           cd $VCPKG_ROOT
           ./bootstrap-vcpkg.sh
+
+      - name: Set up vcpkg cache
+        uses: ./.github/actions/cache-vcpkg
+        with:
+          name_os: linux
+          arch: "x64"
+          compiler: "gcc"
 
       - name: Log environment
         run: ./scripts/ci/log-env.sh
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,9 +24,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-env:
-  VCPKG_ROOT: ${{ github.workspace }}/vcpkg
-
 jobs:
   build_linux:
     name:    ${{ matrix.conf.name }} ${{ matrix.arch }}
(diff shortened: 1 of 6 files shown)


johnnovak (approved):
A welcome addition, @shermp, and it seems to help the build times. 🚀  I've restarted the jobs in this PR to make sure I'm looking at the cached results, and these are the timings:

Windows ~11 min
macOS ~7 min
Linux ~6 min
PVS ~9min

So the average runtimes seem to be less than 10 minutes, which is quite nice! I remember we had 40+ minute Windows build times about 2 months ago, so great improvements thanks the contributions of the whole team! 🥳 

One improvement suggestion if you feel like looking into this. Although we use some duplication in the workflows already, I think this addition would be a very good fit for reusable composite actions. Here's a short tutorial about it, it's rather easy to do, then the three copy-pasted caching pre-steps could live in a `cache-vcpkg` composite action that we can just call.

https://docs.github.com/en/actions/tutorials/creating-a-composite-action

.github/workflows/linux.yml
```diff
@@ -92,6 +92,13 @@ jobs:
           cd $VCPKG_ROOT
           ./bootstrap-vcpkg.sh
 
+      - name: Setup vcpkg cache
```
johnnovak: minor: "Set up" is the verb form

shermp:
@johnnovak that was a good idea. I had actually used composite actions back when I had implemented the Docker cross compile stuff, but I had forgotten about it!

As an aside, does anyone remember why we are explicitly checking out the vcpkg baseline commit in the Windows workflow? Is it still needed? If not, that can save us a bit more time too.

johnnovak:
> @johnnovak that was a good idea. I had actually used composite actions back when I had implemented the Docker cross compile stuff, but I had forgotten about it!

I wasn't aware of it, just thought it would be nice to extract these copy-pasted steps into a common workflow, and that's how I found out about it.

> As an aside, does anyone remember why we are explicitly checking out the vcpkg baseline commit in the Windows workflow? Is it still needed? If not, that can save us a bit more time too.

Probably that's a remnant of kklobe's early Window vcpkg workflow that we forgot to remove. It should not be needed.

.github/workflows/windows.yml
```diff
@@ -64,6 +64,12 @@ jobs:
           git -c advice.detachedHead=false checkout $baseline
           bootstrap-vcpkg.bat -disableMetrics
 
+      - name: Set up vcpkg cache
+        uses: ./.github/actions/cache-vcpkg
```
johnnovak: minor: Why the `./` prefix?

Ok, the example in the tutorial has it too:

```
uses: ./.github/actions/hello-world-composite-action
```

It just looks a bit weird, I was wondering if it's really needed.

shermp: That's how it's written in the help page you referenced.

shermp:
Sigh... removing the Windows checkout alters the vcpkg ABI hash, which makes the cache ineffective right now. The options are:
- include the current vcpkg commit as part of the cache key
- add a suffix to manually bump the cache

Thoughts?

johnnovak:
> Sigh... removing the Windows checkout alters the vcpkg ABI hash, which makes the cache ineffective right now. The options are:
> 
>     * include the current vcpkg commit as part of the cache key
> 
>     * add a suffix to manually bump the cache
> 
> 
> Thoughts?

Adding the commit to the cache key is a good idea IMO, that should do the trick.

If if doesn't, we can restore the "Checkout vcpkg baseline" step (with a comment stating why we need it).

I'm not sure what you mean by "suffix to manually bump the cache".

shermp:
> I'm not sure what you mean but "suffix to manually bump the cache".

The old cache actions (such as ccache) used to have numbered suffixes that could be manually bumped to force fresh caches.

But anyway, the vcpkg commit short hash is now included in the cache key.

shermp:
Actually, I think it would be a good idea to check out a stable commit anyway, because vcpkg can update several times a day, and the linux/macos jobs just clone the vcpkg repo.

johnnovak:
> Actually, I think it would be a good idea to check out a stable commit anyway, because vcpkg can update several times a day, and the linux/macos jobs just clone the vcpkg repo.

Yeah, I think I like that approach better. I think it's better if we just checked out our baseline explicitly (vs checking out the latest, then using our baseline). If that creates problems, that should be rare, and we don't wanna lag behing like 6+ months anyway. But lagging behind 3-4 weeks or a bit more is beneficial when things are in flux with newer vcpkg versions and newer ports. I like us dictating the pace of upgrades vs being pushed on us at inconvenient times.

shermp:
Ok, I settled on a middle ground. It turns out all the runner images have vcpkg preinstalled. So I've updated the actions to just use the preinstalled version.

johnnovak:
Looks very good now @shermp, I think it's ready for merging. I don't think there's anything controversial there that would warrant further review.

shermp:
Thanks @johnnovak 

The "long" vcpkg build times were starting to annoy me, hence the PR. I've been meaning to do this for a while.

PR #4425, opened 2025-07-14
Inject Nuked-SC55-CLAP plugin into the release packages

johnnovak:
# Description

As per title, plus some cleanup.

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/pull/4090

# Release notes

Release notes are in https://github.com/dosbox-staging/dosbox-staging/pull/4090

# Manual testing

- [x] Windows (both portable & installer)
- [x] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -72,6 +72,9 @@ jobs:
         with:
           submodules: false
 
+      - name: Set common vars
+        uses: ./.github/actions/set-common-vars
+
       - name: Set up CMake
         uses: lukka/get-cmake@6b3e96a9bc9976b8b546346fdd102effedae0ca8 # v4.0.3
 
@@ -90,12 +93,6 @@ jobs:
 
       - name: Log environment
         run: ./scripts/ci/log-env.sh
-
-      - name: Inject version string
-        run: |
-          set -x
-          VERSION=$(./scripts/ci/get-version.sh version-and-hash)
-          echo "VERSION=$VERSION" >> $GITHUB_ENV
 
       - name: Build
         run: |
@@ -126,6 +123,9 @@ jobs:
         with:
           submodules: false
 
+      - name: Set common vars
+        uses: ./.github/actions/set-common-vars
+
       - name: Set up CMake
         uses: lukka/get-cmake@6b3e96a9bc9976b8b546346fdd102effedae0ca8 # v4.0.3
 
@@ -144,12 +144,6 @@ jobs:
 
       - name: Log environment
         run: ./scripts/ci/log-env.sh
-
-      - name: Inject version string
-        run: |
-          set -x
-          VERSION=$(./scripts/ci/get-version.sh version-and-hash)
-          echo "VERSION=$VERSION" >> $GITHUB_ENV
 
       - name: Build release
         run: |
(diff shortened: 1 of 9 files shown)


johnnovak:
@weirddan455 @FeralChild64 Would you be able to test the Linux release package?

.github/workflows/linux.yml
```diff
@@ -190,6 +190,20 @@ jobs:
           cp vcpkg-deps/release/* $LIB_DIR
 
 
+      - name:  Inject Nuked-SC55-CLAP plugin
+        run: |
+          set -x
+          NUKED_SC55_CLAP_VERSION=v0.7.0
```
FeralChild64: Would it be feasible to define the version somewhere else, so that we don't have to change it in 3 separate places?

FeralChild64:
Did a quick test on my Linux system - SC55 works as soon as you copy the rom images into appropriate `resources/plugins` subdirectory.

.github/workflows/linux.yml
johnnovak: Yeah, I was considering that, good idea. I'll move setting this and the external vcpkg deps version into a common action.

weirddan455:
All the CI jobs are failing at the moment so there's no output binary for me to test.

> Did a quick test on my Linux system - SC55 works as soon as you copy the rom images into appropriate resources/plugins subdirectory.

If you're going to be including all of the needed SC55 files, should we put it in the resources folder alongside shaders and such to avoid the need for this manual step?

FeralChild64:
But the ROM files are the CLAP plugin resources - they are DOSBox Staging resources only indirectly.

johnnovak:
> If you're going to be including all of the needed SC55 files, should we put it in the resources folder alongside shaders and such to avoid the need for this manual step?

We don't wanna include these copyrighted files in the release packages stored on the GitHub servers. Providing the ROM files will be a manual step, similarly to the MT-32 files.

I'll fix the CI job a bit later, started doing some improvements as @FeralChild64 suggested.

.github/workflows/linux.yml
johnnovak: Done.

johnnovak:
Okay @weirddan455, I've stopped fooling around, you can test it now.

liPillON:
> All the CI jobs are failing at the moment so there's no output binary for me to test.
> should we put it in the resources folder alongside shaders and such to avoid the need for this manual step?

I've been lurking for a while, this comment gives me the opportunity to jump in.

Would it be possible to add a dosbox setting to specify the path to the folder where the clap plugin has been deployed?

johnnovak:
> Would it be possible to add a dosbox setting to specify the path to the folder where the clap plugin has been deployed?

No, it has to be in its designated location, and it will be bundled with the release as you can see. So why would you even want it to live somewhere else? It's very small, you can have a few copies on your machine 😄

weirddan455 (approved):
> We don't wanna include these copyrighted files in the release packages stored on the GitHub servers. Providing the ROM files will be a manual step, similarly to the MT-32 files.

Got it so you're only including the .clap file and not the ROM files. I've tested it and it works. The one thing that confused me at first is that it failed when I have both the .clap file and the ROM files inside `~/.config/dosbox/plugins`.

It seems that `resources/plugins` takes priority over `.config` and the ROM files must be located in the same folder as the .clap file. I had to copy the ROM files for it to work. This is probably fine as long as we document it so people can get up and running easily.

This behavior does make `.config/dosbox/plugins/` a bit pointless for normal users as it will never be searched as long as we're bundling this .clap file. It's still needed for our local dev builds though.

johnnovak:
> Got it so you're only including the .clap file and not the ROM files. I've tested it and it works. The one thing that confused me at first is that it failed when I have both the .clap file and the ROM files inside `~/.config/dosbox/plugins`.

All good, thanks for the testing 👍🏻 

> It seems that `resources/plugins` takes priority over `.config` and the ROM files must be located in the same folder as the .clap file. I had to copy the ROM files for it to work. This is probably fine as long as we document it so people can get up and running easily.

Yeah that's the key part, and logically it makes sense to me: everything works fine as long as the CLAP plugin and its ROMs are in the same folder. Yep, it's already documented in the release notes here: https://github.com/dosbox-staging/dosbox-staging/pull/4090

> This behavior does make `.config/dosbox/plugins/` a bit pointless for normal users as it will never be searched as long as we're bundling this .clap file. It's still needed for our local dev builds though.

Kinda, yeah, just it's easier to rely on the existing config subdir lookup mechanism.

PR #4435, opened 2025-07-22
Build system improvements, including Linux CMake install rules

FeralChild64:
# Description

The following build system improvements were implemented:
- Moved the Unicode data license to the standard location.
- Added Linux install rules to the CMake build system. Rationale: they are needed by the distro developers to create packages.
- As a bonus, our packages created by CI does not contain `meson.build` or `.gitattributes` files in the resources directories anymore.

Example usage of the install rules:

```bash
cmake --preset=release-linux
cmake --build --preset=release-linux
cmake --install build/release-linux --prefix "/home/foobar/dosbox-staging"
```

## Related issues

# Manual testing

1. (Linux, Windows) Downloaded packages created by CI, confirmed that resources and licenses are still there.

2. (Linux) Uninstalled the distro DOSBox Staging, compiled the code and installed it to a given directory (as described in the description). Checked that DOSBox Staging can be started and that keyboard layout can be changed (`keyb de`, `keyb fr`, etc. commands). Also checked that directory structure (icons, menu entry file, etc.) is the same as for the software installed by the Linux package manager.

I was unable to check whether the macOS build artifacts are correct - I don't know how to create the proper Macintosh package and check it's content - but when I messed up the paths when working on the PR, I could observe build failures as a result; now it builds without errors.

- [x] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/scripts/packaging/create-package.sh
+++ b/scripts/packaging/create-package.sh
@@ -53,43 +53,28 @@ install_doc()
     # Install common documentation files
     case $platform in
         linux)
-            install_file docs/README.template      "${pkg_dir}/README"
-            install_file LICENSE                   "${pkg_dir}/LICENSE"
-            install_file docs/dosbox.1             "${pkg_dir}/man/dosbox.1"
-            install_file licenses/BSD-2-Clause.txt "${pkg_dir}/doc/licenses/BSD-2-Clause.txt"
-            install_file licenses/BSD-3-Clause.txt "${pkg_dir}/doc/licenses/BSD-3-Clause.txt"
-            install_file licenses/DEBUG.COM.txt    "${pkg_dir}/doc/licenses/DEBUG.COM.txt"
-            install_file licenses/GPL-2.0.txt      "${pkg_dir}/doc/licenses/GPL-2.0.txt"
-            install_file licenses/LGPL-2.1.txt     "${pkg_dir}/doc/licenses/LGPL-2.1.txt"
-            install_file licenses/MIT.txt          "${pkg_dir}/doc/licenses/MIT.txt"
-            install_file licenses/UNICODE.txt      "${pkg_dir}/doc/licenses/UNICODE.txt"
-            install_file licenses/Zlib.txt         "${pkg_dir}/doc/licenses/Zlib.txt"
+            for filename in ${build_dir}/doc/licenses/*; do
+            	install_file "${filename}" "${pkg_dir}/doc/licenses/$(basename ${filename})"
+            done
+            install_file "${build_dir}/doc/LICENSE" "${pkg_dir}/LICENSE"
+            install_file docs/dosbox.1              "${pkg_dir}/man/dosbox.1"
+            install_file docs/README.template       "${pkg_dir}/README"
             readme_tmpl="${pkg_dir}/README"
             ;;
         macos)
-            install_file docs/README.template      "${macos_content_dir}/SharedSupport/README"
-            install_file LICENSE                   "${macos_content_dir}/SharedSupport/LICENSE"
-            install_file licenses/BSD-2-Clause.txt "${macos_content_dir}/doc/licenses/BSD-2-Clause.txt"
-            install_file licenses/BSD-3-Clause.txt "${macos_content_dir}/doc/licenses/BSD-3-Clause.txt"
-            install_file licenses/DEBUG.COM.txt    "${macos_content_dir}/doc/licenses/DEBUG.COM.txt"
-            install_file licenses/GPL-2.0.txt      "${macos_content_dir}/doc/licenses/GPL-2.0.txt"
-            install_file licenses/LGPL-2.1.txt     "${macos_content_dir}/doc/licenses/LGPL-2.1.txt"
-            install_file licenses/MIT.txt          "${macos_content_dir}/doc/licenses/MIT.txt"
-            install_file licenses/UNICODE.txt      "${macos_content_dir}/doc/licenses/UNICODE.txt"
-            install_file licenses/Zlib.txt         "${macos_content_dir}/doc/licenses/Zlib.txt"
+            for filename in ${build_dir}/licenses/*; do
+            	install_file "${filename}" "${macos_content_dir}/doc/licenses/$(basename ${filename})"
+            done
+            install_file "${build_dir}/LICENSE" "${macos_content_dir}/SharedSupport/LICENSE"
+            install_file docs/README.template       "${macos_content_dir}/SharedSupport/README"
             readme_tmpl="${macos_content_dir}/SharedSupport/README"
             ;;
         windows)
-            install_file docs/README.template      "${pkg_dir}/README.txt"
-            install_file LICENSE                   "${pkg_dir}/LICENSE.txt"
-            install_file licenses/BSD-2-Clause.txt "${pkg_dir}/doc/licenses/BSD-2-Clause.txt"
-            install_file licenses/BSD-3-Clause.txt "${pkg_dir}/doc/licenses/BSD-3-Clause.txt"
-            install_file licenses/DEBUG.COM.txt    "${pkg_dir}/doc/licenses/DEBUG.COM.txt"
-            install_file licenses/GPL-2.0.txt      "${pkg_dir}/doc/licenses/GPL-2.0.txt"
-            install_file licenses/LGPL-2.1.txt     "${pkg_dir}/doc/licenses/LGPL-2.1.txt"
-            install_file licenses/MIT.txt          "${pkg_dir}/doc/licenses/MIT.txt"
-            install_file licenses/UNICODE.txt      "${pkg_dir}/doc/licenses/UNICODE.txt"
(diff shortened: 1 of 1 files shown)


shermp:
I made a start of this at [sp/cmake-install-1](https://github.com/dosbox-staging/dosbox-staging/tree/sp/cmake-install-1).

Since you have decided to take this on, please feel free to steal anything useful from that branch.

FeralChild64:
@shermp It seems I am using a different approach, so probably not much can be taken directly. But we'll see - I didn't knew about the `GNUInstallDirs` module, for example.

FeralChild64:
The change is mostly done (@shermp, thanks for your branch - there were a few things there I have missed).

What I don't like is that the list of licenses to bundle is stored 3 times in  `create-packages.sh` and now 1 time more in the main `CMakeLists.txt` file - this is not a safe approach, I intend to fix this before I mark the PR as ready for review.

shermp:
Glad you found it useful!

As an aside for those following along, on most modern linux distros, if you install to `~/.local`, then desktop integration should "just work (tm)".

contrib/resources/drives/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(FILES
+
+    y/dos/debug.com
+    y/dos/deltree.com
+    y/dos/xcopy.exe
+    y.conf
+)
```
johnnovak, line 7: Why not just glob `y/*` too here? We shouldn't put anything in to the `y` folder we don't wanna include.

contrib/resources/mapping/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(FILES
+
+    ASCII.TXT
+    CAPITAL_SMALL.TXT
+    DECOMPOSITION.TXT
+    MAIN.TXT
+)
+
+list(TRANSFORM FILES PREPEND "translations/")
+set(RESOURCE_FILES_TRANSLATIONS ${FILES} PARENT_SCOPE)
```
johnnovak: My problem with this approach is that none of this globbing should be necessary once we nuke all `meson.build` files from the repo. The idea with `contrib/resources` (which I'm gonna move to `$REPO_ROOT/resources` in the future) is that you can `cp -R resources/*` into the release package (minus perhaps a single `CMakeList.txt` in `resources` root). If that doesn't hold true, stuff needs to be moved out of `resources`, simple as that.

Right now, there's a bunch of utility scripts and other stuff in `resources/translations/` that doesn't belong there.

I'd prefer to keep things simple, so I think it's best to remove all this globbing and precise file matching in each resources subdir, and just copy the `resource` wholesale. We can live with the leftover `meson.build` files for a few months until we remove Meson, they did not cause any actual problems in the past years, it's a minor annoyance at best. So let's build this once for the future I'd say :smile:

What you _can_ do now is move all the files from `contrib/resources/translations` that doesn't belong to the release package to `contrib/translation-scripts` or something. We will find a better name for it later, probably rename it back to `contrib/translations` once I separated `contrib` and `resources`.

contrib/resources/translations/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+    br.lng
+    de.lng
+    es.lng
+    fr.lng
+    it.lng
+    nl.lng
+    pl.lng
+    ru.lng
+)
```
johnnovak: Yeah, so all the rest of the crap can go into `contrib/translation-scripts`.

CMakeLists.txt
```diff
@@ -334,13 +334,76 @@ target_include_directories(
...
 )
 
-# TODO This happens on every execution, better to have a target with
-# dependencies instead
-add_custom_target(copy_assets ALL
-  COMMAND ${CMAKE_COMMAND} -E copy_directory
-    "${CMAKE_CURRENT_SOURCE_DIR}/contrib/resources"
-    "${CMAKE_CURRENT_BINARY_DIR}${RESOURCE_COPY_PATH}"
+# Add contrib directory to make the list of resource files available
+add_subdirectory(contrib)
```
johnnovak: Yeah, so just copy the whole `contrib/resources/*` here wholesale as per my earlier comment. What-you-see-is-what-you-get in terms of what will appear in the final release package is a clear to model to follow with the `resources` dir.

CMakeLists.txt
```diff
@@ -334,13 +334,76 @@ target_include_directories(
...
+endforeach()
+
+# Add a copy command for the main license file
+set(SOURCE      "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
+set(DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/doc/LICENSE")
+add_custom_command(
+    OUTPUT  ${ASSET_DESTINATION_FILES}
+    DEPENDS "${SOURCE}"
+    COMMAND ${CMAKE_COMMAND} -E copy "${SOURCE}" "${DESTINATION}"
+    APPEND
```
johnnovak: That's a lot of extra code dealing with only packing in the main CMakeList.txt. I would much prefer to separate this out into another CMake script and just call it from here. Just like with functions in a regular programming language, that would also clarify what the inputs are to the packaging function instead of just using variables from from the top-level namespace.

contrib/resources/translations/CMakeLists.txt
FeralChild64: We are also omitting `en.lng` here. I would like to get rid of the scripts in the future and just use the gettext format, but we are not there yet.

contrib/resources/mapping/CMakeLists.txt
FeralChild64: There are definitely more files there, which we'll have to take care for - I would prefer to keep the code page and keyboard layout documentation from the FreeDOS project.

OK, I'll rework it to take the whole resources directory tree (maybe I'll find some easy way to filter out the `meson.build` files) - but cleaning up the tree is probably going to be a separate PR.

shermp: From the [docs](https://cmake.org/cmake/help/v3.25/command/install.html#installing-directories) it looks like you can include or exclude patterns for installation at least.

CMakeLists.txt
FeralChild64: Done.

contrib/resources/translations/CMakeLists.txt
FeralChild64: I am now copying out the whole directory, ignoring `meson.build` and `.git(...)` files.

contrib/resources/mapping/CMakeLists.txt
FeralChild64: I am now copying out the whole directory, ignoring `meson.build` and `.git(...)` files.

CMakeLists.txt
FeralChild64: I am now copying out the whole directory, ignoring `meson.build` and `.git(...)` files.

FeralChild64:
Back to draft - I have implemented the changes, but I need to fully test them now, possible do some further cleanup too.

johnnovak:
Thanks for the rework @FeralChild64.

> There are definitely more files there, which we'll have to take care for - I would prefer to keep the code page and keyboard layout documentation from the FreeDOS project.

If you want to keep them, those can go into a separate subdir, outside of `resources`. I'd say a subdir in `docs` is the place for doco stuff, and a subdir in `licenses` for license stuff.

In reality, it's rare that stuff we bundle in `resources` comes with extra files that we wanna keep. It's best to keep resources WYSIWYG.

> OK, I'll rework it to take the whole resources directory tree (maybe I'll find some easy way to filter out the meson.build files) 

As above, I think it's worth optimising for keeping `resources` "WYSIWYG". That's my main consideration, then whatever else follows from that design choice, so be it. So just plain of `cp -R resources/*` into the release packages, done, then any extra files must be stored in other places.

> - but cleaning up the tree is probably going to be a separate PR.

That's perfectly fine 👍🏻

scripts/cmake/add_copy_assets.cmake
```diff
@@ -0,0 +1,69 @@
+
+function(add_copy_assets)
+
```
johnnovak: As per the project structure recommended in the Moder CMake guide, it would be better to put these files into `cmake` at the root level.

https://cliutils.gitlab.io/modern-cmake/chapters/basics/structure.html

`scripts` should only contain actual scripts and tools we can execute from the CLI.

FeralChild64: OK, changed.

FeralChild64:
> So just plain of `cp -R resources/*` into the release packages

Well, it is not that easy - I want the CMake to know the copy command dependencies, so that `cmake --build --preset=(...)` command after updating the translation files will copy the updated translations. It's working now, after my changes - not perfectly (CMake does not notice if a new file is added to the resources), but this is a CMake limitation.

johnnovak (approved):
Nice work 👍🏻

PR #4437, opened 2025-07-26
Improvements to the resource loader

### LINKED ISSUE #4434: CLAP plugin not found when Dosbox is launched from another working directory

_No response_

Linux x86_64

Debian 12

### Relevant hardware info

_No response_

Regression from #4432

`resources/plugins/` is found relative to where the dosbox binary is located. I believe it uses the `whereami` library to do this. 

`plugins/` is found by looking at the current working directory instead. This is likely to cause issues with scripts and desktop shortcuts that don't set the working directory.

Launch Dosbox from a terminal in another directory the full path. Ex: if dosbox is located in `/home/bob/dosbox-dir/dosbox`.

```
cd /home/bob
dosbox-dir/dosbox
```

The Soundcanvas plugin will then fail to be found.

_No response_

### Your configuration

```ini
[midi]
mididevice = soundcanvas
```

### Provide a log

```text
2025-07-20 20:08:02.875 | SOUNDCANVAS: Failed to load 'auto' Sound Canvas model
2025-07-20 20:08:02.875 | MIDI: Error opening device 'soundcanvas'; using 'port'
```

- [x] Yes, I agree.

weirddan455:
# Description

Found and fixed several bugs with the resource loader. View commit by commit as I wrote the problems in the commit messages.

CLAP plugins now searches executable root / plugins rather than the resources directory.

Need some help with Mac. CI is still throwing Nuked.clap  in Resources directory and I have no idea what the folder structure for a Mac app is.

## Related issues

Fixes #4434

# Manual testing

Tested CLAP plugin loads

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckCXXSourceCompiles)
 include(CheckSymbolExists)
+include(GNUInstallDirs)
 
 if(WIN32)
   set(DOSBOX_PLATFORM_WINDOWS ON)
@@ -297,7 +298,7 @@ check_symbol_exists(
   sys_icache_invalidate "libkern/OSCacheControl.h" HAVE_SYS_ICACHE_INVALIDATE
 )
 
-set(CUSTOM_DATADIR " ")
+set(CUSTOM_DATADIR "${CMAKE_INSTALL_FULL_DATADIR}")
 
 set(project_name "${PROJECT_NAME}")
 set(version      "${PROJECT_VERSION}")
--- a/include/support.h
+++ b/include/support.h
@@ -256,6 +256,11 @@ int64_t stdio_num_sectors(FILE* f);
 const std_fs::path& get_executable_path();
 std_fs::path get_resource_path(const std_fs::path& name);
 std_fs::path get_resource_path(const std_fs::path& subdir, const std_fs::path& name);
+const std::vector<std_fs::path>& get_resource_parent_paths();
+
+std::vector<std_fs::path> get_directory_entries(
+        const std_fs::path& dir, const std::string_view files_ext,
+        const bool only_regular_files);
 
 std::map<std_fs::path, std::vector<std_fs::path>> get_files_in_resource(
         const std_fs::path& res_name, const std::string_view files_ext,
--- a/include/support.h
+++ b/include/support.h
@@ -16,7 +16,6 @@
 #include <cstring>
 #include <functional>
 #include <limits>
-#include <map>
 #include <memory>
 #include <stdexcept>
 #include <string>
@@ -261,10 +260,6 @@ std::vector<std_fs::path> get_plugin_paths();
 
 std::vector<std_fs::path> get_directory_entries(
         const std_fs::path& dir, const std::string_view files_ext,
-        const bool only_regular_files);
-
-std::map<std_fs::path, std::vector<std_fs::path>> get_files_in_resource(
-        const std_fs::path& res_name, const std::string_view files_ext,
         const bool only_regular_files);
 
 enum class ResourceImportance { Mandatory, Optional };
--- a/include/support.h
+++ b/include/support.h
@@ -257,6 +257,7 @@ const std_fs::path& get_executable_path();
 std_fs::path get_resource_path(const std_fs::path& name);
 std_fs::path get_resource_path(const std_fs::path& subdir, const std_fs::path& name);
 const std::vector<std_fs::path>& get_resource_parent_paths();
+std::vector<std_fs::path> get_plugin_paths();
 
 std::vector<std_fs::path> get_directory_entries(
         const std_fs::path& dir, const std::string_view files_ext,
--- a/src/config.h.in.cmake
+++ b/src/config.h.in.cmake
@@ -200,10 +200,10 @@
 // Modern MSVC provides POSIX-like routines, so prefer that over built-in
 #cmakedefine HAVE_STRNLEN
 
-// Holds the "--datadir" specified during project setup. This can
+// Holds the CMAKE_INSTALL_DATADIR specified during project setup. This can
 // be used as a fallback if the user hasn't populated their
-// XDG_DATA_HOME or XDG_DATA_DIRS to include the --datadir.
-#cmakedefine CUSTOM_DATADIR
+// XDG_DATA_HOME or XDG_DATA_DIRS.
+#cmakedefine CUSTOM_DATADIR "@CUSTOM_DATADIR@"
 
 
 #endif
--- a/src/misc/support.cpp
+++ b/src/misc/support.cpp
@@ -274,6 +274,23 @@ const std_fs::path& get_executable_path()
 	return exe_path;
 }
 
+static void maybe_add_path(const std_fs::path& path, std::vector<std_fs::path>& paths)
+{
+	std::error_code ec = {};
+	if (!std_fs::is_directory(path, ec)) {
+		return;
+	}
+	// Filter out duplicates by first canonicalizing the path
+	// and then checking if it already exists in the paths vector.
+	// Ex: /usr/share/dosbox-staging and get_executable_path() /../share can point to the same location
+	const auto canonical_path = std_fs::canonical(path, ec);
+	if (ec) {
+		LOG_ERR("RESOURCE: Failed to canonicalize path '%s': %s", path.string().c_str(), ec.message().c_str());
+	} else if (std::find(paths.begin(), paths.end(), canonical_path) == paths.end()) {
+		paths.emplace_back(std::move(canonical_path));
+	}
+}
+
 static const std::vector<std_fs::path>& GetResourceParentPaths()
 {
 	static std::vector<std_fs::path> paths = {};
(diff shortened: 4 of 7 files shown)


CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckCXXSourceCompiles)
 include(CheckSymbolExists)
+include(GNUInstallDirs)
```
johnnovak, line 42: Just saw "GNU", so I'm wondering if this has any flow-on effects on @FeralChild64 WIP CMake install work?

https://github.com/dosbox-staging/dosbox-staging/pull/4435

src/config.h.in.cmake
```diff
@@ -200,10 +200,10 @@
 // Modern MSVC provides POSIX-like routines, so prefer that over built-in
 #cmakedefine HAVE_STRNLEN
 
-// Holds the "--datadir" specified during project setup. This can
+// Holds the CMAKE_INSTALL_DATADIR specified during project setup. This can
 // be used as a fallback if the user hasn't populated their
-// XDG_DATA_HOME or XDG_DATA_DIRS to include the --datadir.
-#cmakedefine CUSTOM_DATADIR
+// XDG_DATA_HOME or XDG_DATA_DIRS.
+#cmakedefine CUSTOM_DATADIR "@CUSTOM_DATADIR@"
```
johnnovak, line 206: Same question for @FeralChild64.

src/misc/support.cpp
```diff
@@ -284,7 +284,15 @@ static const std::vector<std_fs::path>& GetResourceParentPaths()
 	auto add_if_exists = [&](const std_fs::path& p) {
 		std::error_code ec = {};
 		if (std_fs::is_directory(p, ec)) {
-			paths.emplace_back(p);
+			// Filter out duplicates by first canonicalizing the path
+			// and then checking if it already exists in the paths vector.
+			// Ex: /usr/share/dosbox-staging and get_executable_path() /../share can point to the same location
+			const auto canonical_path = std_fs::canonical(p, ec);
+			if (ec) {
+				LOG_ERR("RESOURCE: Failed to canonicalize path '%s': %s", p.string().c_str(), ec.message().c_str());
```
johnnovak: Nice one 🎖️

src/misc/support.cpp
```diff
@@ -342,6 +342,35 @@ const std::vector<std_fs::path>& get_resource_parent_paths()
...
+std::vector<std_fs::path> get_plugin_paths()
+{
+	// Intentionally not using the static cache as this function only gets called once.
+	std::vector<std_fs::path> paths = {};
+
+	// Current working directory
+	maybe_add_path(std_fs::path(PluginsDir), paths);
+
+	maybe_add_path(get_executable_path() / PluginsDir, paths);
+	maybe_add_path(get_executable_path() / ".." / PluginsDir, paths);
```
johnnovak, line 357: What's the use-case for `$EXE_PATH/../plugins`?

For macOS, I think the best place for the plugins is the standard `PlugIns` folder inside the application bundle.

> - `PlugIns`
> Contains loadable bundles that extend the basic features of your application. You use this directory to include code modules that must be loaded into your application’s process space in order to be used. You would not use this directory to store standalone executables.

([source](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1), Table 2-5  Subdirectories of the Contents directory)

So you'll need an extra special entry here for macOS, something like this, then I'll help testing it:

```cpp
#if defined(MACOSX)
    maybe_add_Path(get_executable_path() / ".." / "PlugIns");
#endif
```

So yeah, inside the app bundle it's `PlugIns`, then outside of it `plugins`; this will be a macOS-special thing 🤷🏻

src/misc/support.cpp
```diff
@@ -342,6 +342,35 @@ const std::vector<std_fs::path>& get_resource_parent_paths()
...
+
+	#if !defined(WIN32) && !defined(MACOSX)
+	maybe_add_path(get_xdg_data_home() / DOSBOX_PROJECT_NAME / PluginsDir, paths);
+
+	for (const auto& data_dir : get_xdg_data_dirs()) {
+		maybe_add_path(data_dir / DOSBOX_PROJECT_NAME / PluginsDir, paths);
+	}
+	#endif
+
+	maybe_add_path(get_executable_path() / "../share" / DOSBOX_PROJECT_NAME / PluginsDir, paths);
```
johnnovak, line 368: Should this belong to the `#if !defined(WIN32) && !defined(MACOSX)` block as it's a Linux-special thing?

include/support.h
```diff
@@ -263,10 +262,6 @@ std::vector<std_fs::path> get_directory_entries(
         const std_fs::path& dir, const std::string_view files_ext,
         const bool only_regular_files);
 
-std::map<std_fs::path, std::vector<std_fs::path>> get_files_in_resource(
-        const std_fs::path& res_name, const std::string_view files_ext,
-        const bool only_regular_files);
-
```
johnnovak, line 263: To further clarify your commit message, does insertion order mean the user cannot override bundled resources? If that's the case, that's very good. E.g., they cannot override the `interpolation/sharp` shader by providing `glshader/interpolation/sharp.glsl` in their config dir.

Also, could you explain this lookup behaviour somewhere, probably in a comment in the header file?

johnnovak:
Thanks for this @weirddan455 , looks great 💯 Just a few minor comments. 

> Need some help with Mac. CI is still throwing Nuked.clap in Resources directory and I have no idea what the folder structure for a Mac app is.

Yeah, upon further consideration, the `PlugIns` folder inside the app bundle is the correct place for it, so the CI script needs a slight tweak. See the comment I made above about this. Happy to help testing it on macOS once you got it to pass CI.

src/misc/support.cpp
weirddan455: > What's the use-case for $EXE_PATH/../plugins

I copied it from `get_resource_parent_paths()`. For the `resources` directory it's needed for local builds. Ex: Binary goes in `build/debug/`. Resources folder goes in `build/resources`. I don't know if we really need it for plugins. It would solve the Mac case if I just `#ifdef` the `PluginsDir` constant though.

I've never see anyone capitalize it like `PlugIns` before. Macs...  :roll_eyes:

src/misc/support.cpp
weirddan455: Again, copied that from existing code. I think MSYS/Cygwin on Windows may have been the reason to leave it turned on everywhere (although I don't know why you would use those outside of dev work). Does Mac have a `/usr/share` folder at all?

src/misc/support.cpp
johnnovak: Well, technically "plugins" will work as the filesystem is case-preserving but not case sensitive, now that I think of it, so I think no action is needed, after all.

image

include/support.h
weirddan455: > To further clarify your commit message, does insertion order mean the user cannot override bundled resources? If that's the case, that's very good. E.g., they cannot override the interpolation/sharp shader by providing glshader/interpolation/sharp.glsl in their config dir.

Yes, but this doesn't change behavior for shaders. It's always been like that. The shader code was only using this flawed function for logging. CLAP was using this for actual lookup and would have been a problem.

> Also, could you explain this lookup behaviour somewhere, probably in a comment in the header file?

There's lots of comments in the `get_resource_parent_paths()` function explaining the priorities.

src/misc/support.cpp
johnnovak: It does, but that's only used for "system level" tools, Unixy command line tools, but not by desktop applications with a GUI. Generally, desktop apps live in a different world, and they're roughly similar to FlatPaks. All the `/usr`, `/bin` etc. stuff are really for system stuff and Unix-like CLI tools, really.

So no, `/usr/share` lookups are not needed on macOS as we're a desktop app with a GUI, not a low-level system service or a CLI-only dev/system tool.

image

include/support.h
johnnovak: > There's lots of comments in the get_resource_parent_paths() function explaining the priorities.

Indeed, just it's not in once place, but that will do the job for now. Ideally, this would be in the header file above the function definition.

CMakeLists.txt
FeralChild64: Yes - I'm including this module too :)

src/config.h.in.cmake
FeralChild64: Interference between our patches is possible, but I rather suspect a positive impact :)

johnnovak:
@weirddan455 
Tested on macOS using the CI generated artifact. Manually moving `plugins` from `Resources` inside the app bundle to `PlugIns` at the same level as `Resources` works as expected because of the case-preserving nature of the filesystem.

image

So we just need a small change to the macOS CI workflow, to the "Inject Nuked-SC55-CLAP plugin" step, specifically. This is what we have now at the end of it:

```bash
PLUGINS_DIR="dist/DOSBox Staging.app/Contents/Resources/plugins"
mkdir "$PLUGINS_DIR"
cp -R nuked-sc55-clap/* "$PLUGINS_DIR"
```

I think changing the first line this this will do the trick, so please try that then I'll retest:

```bash
PLUGINS_DIR="dist/DOSBox Staging.app/Contents/PlugIns"
```

Changes look good otherwise, great job 👍🏻

src/misc/support.cpp
johnnovak: So I can confirm this works. Can you put this extra comment above `maybe_add_path(get_executable_path() / ".." / PluginsDir, paths);`, please?

```
// This will also resolve `$APP_BUNDLE/Contents/PlugIns` on macOS,
// as the filesystem is case-preserving.
```

weirddan455: Added comment

src/misc/support.cpp
weirddan455: Good to know. I added that line to the `#if` block and for resources as well. I think it really only makes sense on non-hipster Unix's like Linux and BSD.

weirddan455:
Fixed Mac CI and addressed the other comments. If you could test the latest build on Mac again that would be great :smile:

johnnovak:
> Fixed Mac CI and addressed the other comments. If you could test the latest build on Mac again that would be great 😄

Cheers, will do later today.

johnnovak:
@weirddan455 Things work as expected on macOS after I put the ROM files into `PlugIns`. Approving PR, nice work 🏅 

image

Issue #4385, opened 2025-06-17
Audio dropouts when switching video modes on Windows only (regression)


johnnovak:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

main

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Are you able to trigger the issue with our official DOSBox Staging release?

- [x] I am running an official Staging release, not a release installed by my operating system's package manager, or a version repackaged by an external team.

### Relevant hardware info

Ryzen 7900, RTX 3060, Realtek on-board audio

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

I can reproduce this 100% of the time on Windows, but not on macOS. So it seems to be a Windows specific thing (probably how SDL interfaces with the OS audio APIs, and how our code changes upset that a bit).

The issue started appearing after the grand mixer refactor in 0.82.0, so paging @weirddan455. In earlier version I'm getting absolutely zero audio glitches on emulated video mode changes while the GUS is playing back the music in the demo, but from 0.82.0 onwards, there are very noticeable audio dropouts on each screen mode change.

Not exactly a high priority issue because most games don't switch screen modes all the time like this, but still a regression. The intro of **Pinball Illusions** also switches modes like crazy, so most likely it's also affected (haven't tested).

So yeah, I'm suspecting this is an unwanted side-effect of the otherwise very nice mixer improvements that made MIDI-based games so much more enjoyable to play.

### Steps to reproduce the behaviour

1. Start the attached repro pack (Show by Majic 12) with `dosbox --noprimaryconf`
2. Windowed mode is fine, make sure you can see the log window as well as the demo runs.
3. Watch the first minute of the demo until the white dotted circles appear after the big SHOW logo.
4. Note that each time the program switches to a different video mode (about 5 times until the big logo), there are very noticeable audio dropouts.

### Download URL of affected game or software

Repro pack: [show.zip](https://github.com/user-attachments/files/20771334/show.zip)

### Your configuration

I'm using the `sharp` shader to rule out any shader recompilation induced extra delays caused by the auto-CRT shaders.

```ini
[sdl]
presentation_mode = cfr
vsync = off
# same results with the texture backend
# output = texture

[render]
glshader = sharp

[gus]
gus = on

[autoexec]
c:
cd show
show
```

[...]

Issue #4361, opened 2025-06-04
Simplify frame presentation and potentially improve VRR frame pacing


johnnovak:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

## Overview

Frame presentation as it stands currently is way too complex, especially the `auto` mode. The guy who wrote it left the project (kcgen), and no one really dares to touch it because of the complexity of the implementation.

Moreover, later incarnations of the frame presentation code introduced some regressions in `auto` mode (see https://github.com/dosbox-staging/dosbox-staging/issues/4341).

The goal is to keep the good parts of the existing code, do away with the questionable/complicated/flaky parts, and vastly simplify the implementation.

## Context

kcgen did some genuinely useful improvements around better (more even) frame-pacing when the emulated DOS mode and the host monitor have different fixed refresh rates (e.g. 60 Hz monitor and a 70 Hz DOS game, or vice versa), plus he added proper VRR support. But he didn't stop there; he started overloading these features with questionable quality-of-life improvements that didn't pan out that well. I think the intention was good but he did not really consider the complexity involved in all this and it turns out we can't really do these automatic decisions reliably anyway (but hindsight is always 20/20, to be fair).

For example, he also tried to "optimise" presenting frames, so if a DOS game only updates the screen say at 10 Hz, only present the frames at 10 Hz, regardless of the fact that the game uses say the 320x200 70 Hz VGA mode. That's nice in principle, but it complicates things a lot, especially if VRR monitors are involved that can't just handle dropping down to 10 FPS as they usually have a VRR range that starts just a bit below 50 Hz, or maybe from around 28-35 Hz on better models, but certainly not as low as 10 Hz. He also attempted to add this weird "auto-adjusting auto vsync" feature where the vsync criteria is re-evaluated after every DOS mode and host-side windowed/fullscreen change, and decisions made by the "vsync auto adjuster" are fed into the "presentation mode auto adjuster" 😅

[...]

Issue #4336, opened 2025-05-20
Heavy audio stutter in 70hz games if vsync is enabled and running in windowed


mckellenjaundice122:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Are you able to trigger the issue with our official DOSBox Staging release?

- [x] I am running an official Staging release, not a release installed by my operating system's package manager, or a version repackaged by an external team.

### Relevant hardware info

i5-7200U @ 2.50 GHz, 8 GB Ram, Intel HD 630, 60hz/1080p Monitor

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

In games or simply whenever the refresh rate changes to 70hz, the audio performance takes a huge nose dive and starts to stutter heavily regardless of cycle or mixer settings. I have tried nearly every permutation of these settings that i could think of, and there doesn't seem to be any way to keep vsync on and also have a smooth experience in windowed mode.

Another important thing to note, is that this exact issue only started happening as of 0.81.0, and has not happened before.

The way vsync worked before, is that if it was enabled, it would only work in fullscreen mode, and as far as i know, this is intended behavior since it seems trying to force vsync in windowed wasn't implemented, and maybe something changed, because i can tell, that with vsync enabled now, the image is smooth both in windowed and fullscreen in games that otherwise are very jerky such as pinball games. In fullscreen, everything seems fine and audio returns working fine as it did back in 0.80.1 and prior.

I have tested this using a default configuration file by changing only the vsync option and nothing else, all other settings left to normal, other than that, i've also tested to see if it was the CRT shader causing this, but setting it to sharp also doesn't seem to change anything, like mentioned previously, increasing the block size and prebuffer does not help, neither does increasing or decreasing the cycles.

From what i can tell, this might be because vsync is being applied to windowed mode as well, which might be a new feature since it's been a while since i updated, but on low end hardware such as mine, this completely makes vsync unusable in windowed mode, and there doesn't seem to be a way to enable or disable vsync on the fly, as changing it from the command line such as before starting a game doesn't seem to make any effect. Perhaps adding an additional setting for enabling vsync only in fullscreen rather than both windowed and fullscreen might help in order to reflect vsync behavior before 0.81.0.

[...]

PR #4438, opened 2025-07-27
[WIP] Frame presentation overhaul

johnnovak:
# Description

## Related issues

# Release notes

# Manual testing

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


johnnovak:
@kklobe Nice try, but that auto-review was like a bad version of Sonar 😆 The summary was half-passable, but clearly it generated it from my commit messages. It caught one typo, that was its only useful contribution, so thanks, expensive spellchecker! 🤣

johnnovak:
@0mnicydle If you're still around, you might want to update your high-refresh gaming page for 0.83.0, given the VFR mode will go away. I think you can achieve similar results using `dos_rate` to get into those obscenely high refresh rate territories 😆 

https://github.com/dosbox-staging/dosbox-staging/wiki/High-refresh-rate-gaming

0mnicydle:
@johnnovak If I understand correctly, I think you might as well nuke the HFR guide at this point as it mainly documents an interesting interaction between VFR and 1000Hz dosrate whereby you essentially force immediate (same tick) screen updates in accordance with video memory changes, meaning a game's natural frame rate is presented to a VRR monitor. Using CFR and 1000Hz dosrate will result in frames exceeding max refresh of VRR monitors, thus nullifying the benefits of VRR (syncing monitor refresh to incoming frame rate).

So for an example of what the functional difference is, in Dark Forces which actually wants to run at ~73fps, at 1000Hz dosrate with VFR and 144Hz VRR the monitor syncs to 73Hz and thus frame pacing is native/natural and appears consistently smooth. Without VFR you might think if you simply set dosrate to 73Hz or some multiple it will result in no functional difference, however the issue is now you have an emulated display frequency sitting between the game and the physical monitor running at a specific frequency that regardless of value is not synchronized, and thus will drift in and out of sync in terms of frame pacing, leading to appearance of rolling stutter storms if an attempt at manual matching is made. This is further complicated by the fact that the game frame rate will most likely fluctuate; in the 1000Hz+VFR+VRR scenario, the monitor will always transparently maintain native frame pacing making this a non-issue in terms of smoothness.

So yeah, let me know if I'm missing something, but I think the specific benefit the guide documents will no longer exist.

johnnovak:
@0mnicydle Thanks, I also thought the guide will be no longer relevant.

However, your Dark Forces explanation doesn't make any sense to me. The game uses the standard 70.087 Hz 320x200 VGA mode. It cannot "want" to run at 73 Hz! Running it on hardware or DOSBox Staging on my 48-120 Hz VRR monitor yields perfectly smooth frame pacing, of course. In VRR mode, frames get presented at even 70 Hz intervals that's all there is to it.

The game engine is coded to present frames vsynced to the standard 70.087 Hz VGA rate, the end.

> however the issue is now you have an emulated display frequency sitting between the game and the physical monitor running at a specific frequency that regardless of value is not synchronized, and thus will drift in and out of sync in terms of frame pacing, leading to appearance of rolling stutter storms

If you play it vsynced on a fixed 60 Hz refresh rate monitor, yes. Not on a VRR monitor. Any weirdo refresh rate will work perfectly fine as long as it's in the VRR range. And on a 48-120 Hz monitor, that will be true for any rate. I tried with demos that do smooth scrolling in custom resolutions with custom refresh rates; it's all perfect.

> This is further complicated by the fact that the game frame rate will most likely fluctuate; in the 1000Hz+VFR+VRR scenario, the monitor will always transparently maintain native frame pacing making this a non-issue in terms of smoothness.

You are confusing at least two things here:

- The refresh rate of the DOS mode. This is _constant_ for any given mode.
- The _actual_ frame rate the game is capable of producing. In games that use hardcoded vsync in the game engine, this is either the DOS rate or 1/2, 1/3, 1/4, etc. of the DOS rate.

Now, VFR confused the concepts because it presented _changed_ frames only. So in the DF case, you'd get a steady stream of frames at 70.087 Hz if your cycles setting was high enough and there's constant movement on the screen. If the emulated cycles setting was not high enough, you could drop into 1/2, 1/3, etc. present rate, just like in modern games. You could _never_ exceed the 70.087 Hz present rate, though, as the game vsyncs _internally_, at the emulated VGA card level! I don't know where are you getting the 73 Hz figure from.

Also, emulated VGA card level vsync is _not_ the same as host level vsync Staging might or might not be doing.

Your 1000 Hz trick makes sense for Build engine games and similar late 90s 3D games like Quake, though. These games can run at any rate with in-game vsync disabled on high cycles settings. But I question the usefulness of that hack as now you can just set a high enough `dos_rate` and call it a day. I doubt there would much of a difference above 120 Hz or 240 Hz. The actual frame rate of the game might fluctuate between 70 and 150 FPS or whatever, and there will be tearing at the VGA emulation level (this was there with the 1000 Hz VFR tick too) but that doesn't matter at 120 Hz or higher in 3D games, you can't see it.

Then with your 1000 Hz VFR trick, a lot of VRR monitor would misbehave and flicker with the wildly fluctuating present rates, which is bad. Now the present rate is always steady, effectively CFR at the DOS video mode or host rate.

Hope that clears things up.

0mnicydle:
@johnnovak So my wild speculation based on observed behavior and a little uneducated perusal of [Theforceengine's](https://github.com/luciusDXL/TheForceEngine) reverse engineering leads me to believe that Dark Forces uses a fixed timer rate that physics calculations are based on, specifically 145.6521Hz (Programmable Interval Timer/8192?), and additionally composes frames to a system memory buffer which is swapped to vid mem only when a complete frame is finished. Observed behavior is that if the game is unrestricted by (standard) dos_rate, the frame rate will cap out at ~73fps, with each frame containing an equal amount of temporal information. If the game is restricted by (standard) dos_rate, once per second ~39ms of temporal information will be dropped resulting in visible stutter. With VFR and 1000Hz dos_rate, the game itself drives the physical display refresh rate as if it were a modern game using adaptive sync. At 1000Hz with CFR, you will get essentially smooth appearance, however since the incoming frames are now no longer synced to the real monitor, the complete frames the game carefully constructed may appear to have some slight tearing.

As far as the flickering you mentioned, well that's entirely down to inferior implementations of VRR tech and doesn't happen on my monitor, but in any case I'm not attempting to convince you to preserve the behavior; it's pretty limited in utility.

johnnovak:
> Observed behavior is that if the game is unrestricted by (standard) dos_rate, the frame rate will cap out at ~73fps,

I don't see how that's possible. The game doesn't tear on hardware, so it must write to the VGA framebuffer during the vblank period. In other words, on real hardware, it "presents" frames at 70.087 Hz if it can (the refresh rate of VGA mode 13h it uses). If it can't, it will do that on the next vblank period, which cuts the FPS in half, so 35.0435 Hz, and so on.

Without even going near reverse engineering the game's engine, just consider this unchangeable fact of DOS games that don't tear when there's fast movement. I don't see how 73 Hz can be even observed in DOSBox in VRR mode. It seems to be a physical impossibility to me, unless you go and rewrite the main loop of the game to not wait for vblank anymore.

All DOS games that sync to vblank roughly do this in the main loop:

- Update tick counter
- Read user input
- Update world state based on user input (or autonomous enemy movement, scripts, etc.)
- Render frame into an in-memory buffer (that's the "render to backbuffer" in modern parlance)
- Wait until vblank starts ("wait until vsync")
- Quickly copy the in-memory buffer to the VGA memory (that's the "present" call)
- Start all over again

That's the rough idea. You simply can't go above the emulated video mode's refresh rate. If you can, the game wasn't waiting for vblank in the first place. But if it wasn't, it must tear like crazy in 3D games. As Dark Forces does not tear at all, it must sync to vblank—just like any other DOS game that does frequent screen updates and doesn't tear.

Whatever PIC refresh rate they use to update the world state is immaterial. They could update the world state at 500 Hz for all we care and it wouldn't matter. Frames would still be rendered at the vblank rate. Probably they set the world state update "callback rate" a bit higher than the actual refresh rate of the screen mode for whatever technical reasons. But they must use real-time to update the world state anyway, so the actual "callback rate" number doesn't matter from a frame presentation perspective. They could have chosen a 80 Hz or 100 Hz "callback rate" for the world update (as long as it's higher than the actual FPS rate), but 73 Hz was probably good enough .

> If the game is restricted by (standard) dos_rate, once per second ~27ms of temporal information will be dropped resulting in visible stutter. 

> With VFR and 1000Hz dos_rate, the game itself drives the physical display refresh rate as if it were a modern game using adaptive sync. 

As above, can't happen with Dark Forces. Can happen with Quake and Build engine games, though (with in-game vsync off).

If you can "see" this on Dark Forces, either some other factors are at play, or it's placebo, sorry 😎 The only way you can possibly achieve this by setting `dos_rate = 73`.

Why don't you go and try the build from this branch? Just play some Dark Forces on your VRR monitor with the default settings (`presentation_mode = dos-rate` and `vsync = off`). To me, the results are perfectly smooth.

> As far as the flickering you mentioned, well that's entirely down to inferior implementations of VRR tech and doesn't happen on my monitor,

That's cool, but many VRR monitors are badly affected by flicker. Until monitor makers fix their essentially broken implementations, the best bet is to supply a steady rate of frames. And that's the most logical and practical model in emulators anyway; DOSBox was quite the outlier with this VFR mode relic from the past (maybe it made some sense in the early 2000s).

> but in any case I'm not attempting to convince you to preserve the behavior; it's pretty limited in utility.

Yeah, I'm not really arguing either, more like educating people on the subtleties of how frame presentation works—not just for your benefit, but for anybody reviewing this PR and reading this page later. I feel I must be super clear and precise when describing what's happening because it's easy to get confused by this whole topic 😎 

Plus I wanted to give you a chance to update your high-refresh rate page to the new system I'm introducing (e.g., how to use `dos_rate` to achieve high refresh rates in Build engine games and Quake). Up to you, really, and I can just simply remove that page too, as you said, that's fine too.

0mnicydle:
@johnnovak The issue as I see it in DF is a mismatch between the physics rate and the standard VGA rate, since the physics are based on a fraction unrelated to the standard VGA timing (though somewhat close). For the full spectrum of possible temporal information to be conveyed to the user, ~73fps are required. At 70Hz, 70fps worth of those fixed fractions are used, and then 3 are skipped.

Stand in place and hold the arrow key or whatever your binding to turn left or right and just hold it... at 70Hz you should see a weird little hitch in motion once a second. With 1000Hz VFR delivering 73fps, no such jump is observed. Whatever the reality about what should or should not be possible, this is the observed behavior on the main branch.

I will try the branch tonight and let you know if the behavior doesn't match my expectations I'm outlining here.

johnnovak:
> @johnnovak The issue as I see it in DF is a mismatch between the physics rate and the standard VGA rate, since the physics are based on a fraction unrelated to the standard VGA timing (though somewhat close). For the full spectrum of possible temporal information to be conveyed to the user, ~73fps are required. At 70Hz, 70fps worth of those fixed fractions are used, and then 3 are skipped.
> 
> Stand in place and hold the arrow key or whatever your binding to turn left or right and just hold it... at 70Hz you should see a weird little hitch in motion once a second.

I believe you, but then some other factors must be at play.

> With 1000Hz VFR delivering 73fps, no such jump is observed. Whatever the reality about what should or should not be possible, this is the observed behavior on the main branch.

How can you know that 73 frames are output every second with the 1000 Hz hack, not 70? You might as well hit some weird "perfect storm" of things that makes things appear smoother on your particular OS, GPU driver, driver settings, monitor, etc. plus the DOSBox hacks. That's possible, but going above the refresh rate of the emulated DOS mode isn't at the emualtion level 🤷🏻 (unless you hit some weird bugs too with these combinations of settings too, but fairly unlikely 😂)

I've also been doing the "turn around test" during my frame presentation rework, and Dark Forces is one of the best games for that test as it presents at a steady 70 Hz rate. I'm getting perfectly smooth results with:

- VRR monitor in VRR mode, default settings using this branch, and fullscreen (both on Windows and macOS). Fullscreen is important; there's always some judder in windowed mode. Also disable all DXGI swapchain customisations; just use stock driver defaults (messing around with the swapchain stuff can also mess up the timings; the driver knows best, let it do its job 😎)
- Still this branch, but running the game with `dos_rate = 71` or `dos_rate = 70` even, then using a fixed 71 or 70 Hz custom mode (via Nvidia Control Panel).

> I will try the branch tonight and let you know if the behavior doesn't match my expectations I'm outlining here.

Thanks for that. Note that in 0.82.2 `presentation_mode = cfr` and `vsync = on` was kinda broken. You might have seen judder because of the broken implementation there, yeah. I'm getting music slowdowns and slowed-down framepacing with that setting on different machines and operating systems, so it's definitely buggy.

---

Also, I was mildly curious, so I looked into briefly into this mystery 73 Hz thing. In short, there's no such thing, as I expected 😆 

From the TFE source code, the internal "tick rate" (world "update rate") of the game is 145.6521 Hz:

```
	// The actual timer frequency used by the DOS code.
	#define TIMER_FREQ 145.6521
```
https://github.com/luciusDXL/TheForceEngine/blob/5a0465b85b03d69c94351ef21f9d2591bdcbde77/TheForceEngine/TFE_DarkForces/time.h#L18

The guy who did the reversing talks about this in [this thread](https://www.doomworld.com/forum/topic/114497-the-force-engine-jedi-engine-portreplacement/?page=4).

So I still have no idea where you got the 73 Hz idea from 😆 But doesn't matter, just test this branch with the stock settings, GPU driver settings reverted to the defaults, and your monitor in VRR mode then and report back, please 😄 Cheers!

vladkosi:
While testing this PR, I've noticed a difference in the sharp shader. I assume this wasn't intended, to me it looks worse.

Latest development build (0.83.0-alpha-029f2):
image0001-rendered

This PR (0.83.0-alpha-5545c):
image0002-rendered

johnnovak:
> While testing this PR, I've noticed a difference in the sharp shader. I assume this wasn't intended, to me it looks worse.

It wasn't intended, it looks broken, indeed. I'm not aware of this, so can you please raise a separate issue ticket with exact repro details? Thanks.

Please include save games too so I can quickly repro it.

vladkosi:
The regression is in this PR, latest development build looks fine.

johnnovak:
> The regression is in this PR, latest development build looks fine.

Okay, can you please give me details, please?

- What's this game (Warcraft 2?)
- Your config, logs, and your monitor resolution
- A save game so I can easily repro it

Thanks. I need all this because I haven't changed the shaders, so it's a bit of a mystery how this PR could have caused it. So, details please 😄

johnnovak:
@vladkosi Nevermind, I've found the bug. I was dicking around with the sRGB framebuffer setting stuff (forgot about that), and lost a line there. Thanks for reporting, no action is needed on your part. Will fix later today, thanks man 😎

0mnicydle:
So testing with this branch/default conf I observed my expected behavior: Regular hitch in constant movement turning in place in DF. VRR was engaged and working at 70Hz, and API frame pacing was as good as ever observed, however in-game motion fluidity was uneven. Switching to 1000Hz in-game motion fluidity was perfect, though minor tearing was observed as incoming frame rate was almost 1000fps and thus VRR was not engaged.

> How can you know that 73 frames are output every second with the 1000 Hz hack, not 70?

Either of two ways: 1) enable the refresh rate overlay on your VRR monitor, or 2) use performance monitoring software that monitors the graphics API presentation rate (RTSS in my case, see attached). These methods both confirm staging main runs DF at a maximum of 73fps/Hz with VFR at 1000Hz dos_rate.

> So I still have no idea where you got the 73 Hz idea from

Observed behavior. 73Hz is also the approx. In reality the max frame rate appears to be exactly 145.6521/2. When it comes to game physics there are 3 main ways to handle it as far as i know: 1) frame rate dependent, which in DOS with vsync would be refresh rate limited (which is why many DOS games run faster at higher dosrates), 2) according to internal timer, and 3) according to internal timer with dynamically generated sub ticks as needed (interpolation).

What I assume is happening in the case of DF is it uses an internal timer, and the game will not generate more than 73 unique fps because it won't have a materially new frame to deliver if it doesn't have a new physics tick to make one with. If the game has the resources available to fully saturate the refresh rate, and that refresh rate is under ~73Hz, in this case 70Hz, the 70 frames it delivers will still each contain the same temporal information as they would at 73Hz because the physics fractions are fixed, and will thus result in perceived stutter since the time budget is consumed before the full set of fractions can be displayed. Now whether under the hood the game is declining to interact with video memory until it has a new frame ready in system memory to copy over, or if it really sends the same frame over and over until a new physics interval is available and dosbox VFR was simply performing an evaluation between frames and declining to present if no changes are detected I couldn't say without more digging, but since you just worked on this you might have an idea.
dosbox_2025_07_30_21_53_33_996

johnnovak:
@0mnicydle Yeah okay, what you're describing there is plausible if we force the DOS rate to a non-existent 1000 Hz rate (non-existent on real VGA cards), the DOS game presents in _some_ vblank periods, "quantised" to 1000 Hz, then the old VFR mode only presents host side when an actual frame change happened. Well, that's at least plausible, but still a lot of "if"s 😄 Just your initial theory-making muddied things a lot—I find it's best to stick to facts 😄 

Speaking of facts, the error with this theory is that with the 1000 Hz refresh rate things are quantised to strict 1 Hz intervals. Whereas 70.087 Hz, or 145.6521/2 = 72.92605 would fall between two such quantised 1 Hz "buckets". So that could introduce its own unevenness into the mix. In theory, because I don't think this all matters in practice 🤷🏻 

I've just tested 0.82.2 with `presentation_mode = vfr` and `dos_rate = 1000`. I'm not noticing more smoothness in the turn around test, and the OSD of my monitor fluctuates between 65-78 Hz. Generally, those OSD readouts are unreliable, but whatever. With this branch, I can't notice a significant degradation or in fact any difference in motion smoothness when turning compared to 0.82.2 with the 1000 Hz hack, and the monitor's OSD reading fluctuates in a similar range randomly.

So from my POV, it could be placebo, or you could legitimately see something there (I have to take your word for it as I can't reproduce this), or maybe my setup doesn't allow the difference to be seen. Or my eyes. Or my nervous system 😆 It's all inconclusive from my POV. But as the current branch with the defaults is certainly smooth enough in Dark Forces to my eyes, I'm going to leave it at that 😄 In any case, even if there is there a difference, it must be minuscule, and wouldn't warrant bringing back the VFR mode (I know you didn't ask for it, just thinking out loud 😄). Plus even if the slight occasional "bump" is there in the evenness, that's how the game looks on a real CRT when running on real hardware.

Deleting the high-refresh wiki page, and thanks for the testing and the chat 👍🏻

0mnicydle:
> Speaking of facts, the error with this theory is that with the 1000 Hz refresh rate things are quantised to 1 Hz intervals/ Whereas 70.087 Hz, or 145.6521/2 = 72.92605 would fall between two such quantised 1 Hz "buckets". So that could introduce it's own unevenness. In theory, because I don't think this all matters in practice.

Absotootles, however ~1ms is well within even my extremely low tolerance for frame pacing variance 🤣

johnnovak:
> > Speaking of facts, the error with this theory is that with the 1000 Hz refresh rate things are quantised to 1 Hz intervals/ Whereas 70.087 Hz, or 145.6521/2 = 72.92605 would fall between two such quantised 1 Hz "buckets". So that could introduce it's own unevenness. In theory, because I don't think this all matters in practice.
> 
> Absotootles, however ~1ms is well within even my extremely low tolerance for frame pacing variance 🤣

Yeah, I'm very sensitive to smooth 2D scrolling and quite allergic to it if there are slight bumps in that every now and then. I grew up with CRTs with the C64 and the Amiga where smooth scrolling was a "staple" in games. So maybe my brain is wired to be able to easily see if 2D scrolling is not 100% smooth.

In 3D games, frame pacing must be really out of whack for me to see anything wrong. I also find it a lot easier to see it in modern games that run at 60 FPS and sometimes drop to 30. I don't think I can notice fluctuations in a 60+ FPS game if the variance is in the plus minus 5-10 FPS range, but I believe there are people out there with a supercharged nervous system who can 😎 I'm 46, for the record 😄 

PS: Some people complain about "24fps judder" in movies, especially on camera pans. I don't know what they're talking about, I've never noticed anything wrong 😂 So probably I'm not sensitive to 3D judder too much 🤷🏻

FeralChild64:
Just compiled this code to do some quick test on Linux - and I have noticed that each time the guest screen mode changes DOSBox switches from fullscreen to windowed mode.

johnnovak:
> Just compiled this code to do some quick test on Linux - and I have noticed that each time the guest screen mode changes DOSBox switches from fullscreen to windowed mode.

Could you bisect which commit caused this? It's fine on Windows and macOS, of course. The problem is SDL event handling is not _truly_ cross platform...

Are you sure `main` is not affected? I might have caused this with the fullscreen rework in https://github.com/dosbox-staging/dosbox-staging/pull/4415, you just might not have noticed yet.

johnnovak:
@vladkosi The sharp shader regression has been fixed.

FeralChild64:
> Could you bisect which commit caused this? It's fine on Windows and macOS, of course. The problem is SDL event handling is not _truly_ cross platform...

```
7763fa9aeeca594594dc9fa6f8c9a5a8ee5c5d14 is the first bad commit
commit 7763fa9aeeca594594dc9fa6f8c9a5a8ee5c5d14
Author: John Novak <john@johnnovak.net>
Date:   Mon Jul 14 09:00:38 2025 +1000

    Fix not allowing to change `[sdl]` section settings after init
```

weirddan455:
> Just compiled this code to do some quick test on Linux - and I have noticed that each time the guest screen mode changes DOSBox switches from fullscreen to windowed mode.

I can't reproduce this. I tested with KDE in X11, XWayland, and native Wayland and fullscreen sticks for me.

weirddan455:
Why are you removing adaptive v-sync as an option? It's probably a better solution to the problem you described than toggling it on/off based on fullscreen.

> It has been reported that on some on-board graphics adapters enabling
> vsync in windowed mode can cause frame presentation stalls, even though
> compositor-level vsync is already enabled in Windows.

> Adaptive vsync enables v-blank synchronisation when the frame rate is higher than the sync rate, but disables synchronisation when the frame rate drops below the sync rate. Disabling the synchronisation on low frame rates prevents the common problem where the frame rate syncs to a integer fraction of the screen's refresh rate in a complex scene. 

https://www.khronos.org/opengl/wiki/Swap_Interval#Adaptive_Vsync

johnnovak:
> I can't reproduce this. I tested with KDE in X11, XWayland, and native Wayland and fullscreen sticks for me.

Thanks for testing that @weirddan455. 

@FeralChild64 That commit from your bisect exercise undoes loading the [sdl] config settings only once at startup. You did that in 95055de for reasons unclear to me, but that's wrong because it will stop the user from modifying any [sdl] section setting at runtime after startup.

I don't know what desktop environment you're using, but I suspect it's something niche. In general, I'd say supporting and testing on that niche environment should be your job. But I'd say hold off with that as I'm going to refactor our very hacky and messy fullscreen switching anyway in my ongoing SDL integration cleanup work (based on GranMinigun's previous unfinished clean up work).

Until then, you can re-apply your workaround for your personal builds if this bothers you after this PR goes in.

> Why are you removing adaptive v-sync as an option? It's probably a better solution to the problem you described than toggling it on/off based on fullscreen.

As per the quoted description, it was invented to prevent the FPS to drop to half the refresh rate in modern 3D games with vsync enabled (e.g., fluctuating between 60 FPS and 30 FPS whenever the FPS drops a bit below 60). Some might prefer tearing at say 40-55 FPS instead of dropping to 30 FPS with no tearing. But in an emulator context, games that use a refresh rate a bit below 60 Hz would start tearing all the time.

Check out my [frame presentation test page](https://github.com/dosbox-staging/dosbox-staging/wiki/Video-tests-%E2%80%94-Presentation); quite a few games use the 59.713 Hz custom VGA modes and similar non-standard refresh rates. And those are 2D games that do smooth scrolling, so tearing would be very distracting. It would be quite confusing to users why some games are tearing while some others aren't in fullscreen mode (e.g., 70 Hz VGA games) with adaptive sync enabled.

So even if it worked perfectly, it's a half-solution at best. The 100% solution is to either enable regular vsync or not, regardless of the refresh mode of the emulated video mode.

The other problem is I've read reports claiming that the implementation is flaky, especially on Nvidia cards. Windows users reported dropped frames and strange uneven pacing and various hiccups with adaptive vsync enabled in windowed mode. So all in all, it's not worth the trouble for a half-solution when we have a reliable solution IMO. It's another OpenGL-special thing with half-assed support, which is not too surprising as 99% of games target DirectX and that only 🤷🏻 

We might add support back if someone can clearly articulate an actual scenario where adaptive vsync is superior to the vsync methods we're currently providing, but I'm not seeing any.

FeralChild64:
> I don't know what desktop environment you're using, but I suspect it's something niche.

I could reproduce the problem on KDE (Plasma version 6.4.3, Applications version 25.4.3 - both are the latest ones), both running on X11 and Wayland and on Hyprland (this one is niche, Wayland-only; version 0.49 - I can't upgrade to 0.50 easily, because it's written in C++26 and I haven't switched to GCC 15.1 yet).

The only really niche part of my system is the GPU - Intel Arc A770 (a so-called Alchemist).

> But I'd say hold off with that as I'm going to refactor our very hacky and messy fullscreen switching anyway in my ongoing SDL integration cleanup work (based on GranMinigun's previous unfinished clean up work).

OK, I'll wait - in the worst case scenario I'll look at it by myself.

FeralChild64:
@johnnovak I have mislead you, sorry - the problem happens when I change the window title; just type something like:

```
config -set window_titlebar='program=path'
```

and it switches from fullscreen to windowed mode. I am running almost every single game via a dedicated BAT file, which adjusts the CPU speed, audio settings, etc, and also changes the window titlebar - that's the reason.

[EDIT] Changing other settings from `[sdl]` section can have similar effect, try `config -set keyboard_capture=off` - same result.

johnnovak:
@FeralChild64 Thanks for the extra info, that's helpful. I'll try to reproduce it 

> @johnnovak I have mislead you, sorry - the problem happens when I change the window title; just type something like:
> 
> ```
> config -set window_titlebar='program=path'
> ```
> 
> and it switches from fullscreen to windowed mode. I am running almost every single game via a dedicated BAT file, which adjusts the CPU speed, audio settings, etc, and also changes the window titlebar - that's the reason.
> 
> [EDIT] Changing other settings from `[sdl]` section can have similar effect, try `config -set keyboard_capture=off` - same result.

Thanks for the extra info, now I can reproduce the issue on Windows too. DOS screen mode changes in a running program don't exit fullscreen, but if I start Staging windowed, manually switch to fullscreen, then set any `[sdl]` section setting, it always reverts back to windowed mode.

You know what, I'll spend a bit of time on this tomorrow. It's probably something simple related to our internal fullscreen state handling. First I thought it's some SDL event handling discrepancy that only affects your specific Linux setup, but now it's clear the problem manifests on every OS as it's about our internal fullscreen flag tracking that gets our of sync with the actual fullscreen state.

I think when I tested setting vsync on / off from the DOS prompt, I only tested in windowed mode so I did not notice this.

johnnovak:
@FeralChild64 Alright, fixed in https://github.com/dosbox-staging/dosbox-staging/pull/4438/commits/70df967c2f883efb9d4ba7566f38bba05a53922b, it was a trivial thing.

FeralChild64:
I confirm the problem is now fixed. Played a little bit on my Linux system (144Hz VFR monitor), I didn't encounter any regressions. Looked through the code, I haven't found anything suspicious.

Since this is rather large PR, let's keep it open for a few days to allow more testing.

johnnovak:
> I confirm the problem is now fixed. Played a little bit on my Linux system (144Hz VFR monitor), I didn't encounter any regressions. Looked through the code, I haven't found anything suspicious.
> 
> Since this is rather large PR, let's keep it open for a few days to allow more testing.

Thanks for testing. The vsync'ed operation is what needs to most testing as that's the biggest change. If you could put your monitor/desktop into 60 Hz mode and could test the vsync'ed operation from the presentation mode tests wiki page (link in the top description), that would be great.

No need to test all the programs I listed, just testing Pinball Dreams and Dark Forces would be enough.

weirddan455:
I tested Pinball Dreams. As far as smoothness of the intro goes, `dos_rate = host` is the only setting that matters. It's smooth regardless of what `vsync` or `presentation_mode` is set to. Setting `dos_rate` to default or 70 results in jerky scrolling.

However, `dos_rate = host`  has the same problem as you mentioned on `main` when using `presentation_mode = vfr` where it changes the speed of the game. It's a little hard to tell at 60hz but I think it is slower. At 144hz host refresh rate, you can easily tell that the game runs twice as fast though (still has smooth scrolling but runs way too fast).

johnnovak:
Tagging @FeralChild64 as well; you might find these Linux testing notes and further explanations useful.

> I tested Pinball Dreams. As far as smoothness of the intro goes, `dos_rate = host` is the only setting that matters. It's smooth regardless of what `vsync` or `presentation_mode` is set to. 

That's expected. You'll only ever get perfect smooth scrolling either on a VRR display with `vsync = off`, or when running your display at a fixed refresh rate with `dos_rate = host` and `vsync = on`. But in the latter case, you'll need to run your display at the refresh rate of the emulated DOS video mode, which is 70 Hz for most VGA games, otherwise the game will run at the wrong speed as you've noticed.

> Setting `dos_rate` to default or 70 results in jerky scrolling.

Yes, because your display is probably running at its native 144 Hz refresh rate. It's expected and fine.

> However, `dos_rate = host` has the same problem as you mentioned on `main` when using `presentation_mode = vfr` where it changes the speed of the game. It's a little hard to tell at 60hz but I think it is slower. At 144hz host refresh rate, you can easily tell that the game runs twice as fast though (still has smooth scrolling but runs way too fast).

It's not really a problem as running the emulated VGA card at custom rates (other than the refresh rates used by actual hardware for a given video mode) is a hack. Please refer to `dos_rate /?` for more details. This hack works with some games that are "less sensitive" to running at custom VGA refresh rates (e.g. late 90s 3D games) and in some earlier games that do smooth 2D scrolling, but with those you might get speed issues. How much the speed issues matter depends on the game and individual sensitivity to speedups/slowdowns. For some, running a game at slightly slower speed but no tearing is preferable to tearing, which can be quite insufferable.

Games fall roughly into these categories:

- **Neither the gameplay nor the audio is tied to vblank** — These are the Build engine 3D games from the late 90s, for instance. You can set `dos_rate = host` to get perfect vsync at your host rate of 144 Hz or whatever.
- **The audio is not tied to vblank, but the gameplay/screen updates are** — These are most 2D games that feature smooth scrolling. You'll get slowdowns or speedups with `dos_rate = host` and `vsync = on`, but tearing will be 100% eliminated. Not very useful at 144 Hz, as you've noticed, and not ideal to run 70 Hz games at ~85% speed (60/70) either on a fixed 60 Hz refresh rate monitor (although that's game dependent and some people are willing to make this sacrifice to get 100% smooth scrolling). One nice trick is to create a custom fixed 70.000 Hz refresh rate on your monitor if you can, then use `dos_rate = host` with `vsync = on` (or just `vsync = on` in the current branch; it's the same). The game will run a _tiny bit_ slower (the standard VGA refresh rate is 70.086 Hz), but that's unnoticeable and you'll get rid of the periodic small jumps that you'd get because of the 70.000 vs 70.086 Hz mismatch with vsync on and using the DOS rate.
- **Both the audio and gameplay/screen updates are tied to vblank** — Dark Forces is such a game. Both the audio and the gameplay/visuals will speed up/slow down when using non-standard refresh rates with `dos_rate`.

Then some games might outright break if you attempt to run them at non-standard refresh rates with `dos_rate`. Same way as some games start exhibiting bugs if you run them at "too fast" cycles settings. Again, `dos_rate` is just a hack, and it's only really useful on fixed refresh monitors. With VRR monitors, there's not much reason to ever use it. VRR monitors just simply sync to our frame presentation rate.

In an ideal world where all monitors are VRR, we could drop `dos_rate` completely.

---

To recap, this is the important test to do on your fixed refresh rate monitor:

- Set your monitor/desktop to 60 Hz refresh rate.
- Run Pinball Dreams with `vsync = on`, both in windowed and fullscreen.
- Scrolling won't be smooth anymore, but the gameplay won't slow down. This is the expected and correct behaviour on this branch (we run the game at 70 Hz internally and render frames at 70 Hz; we just _sample_ the rendered frames at the host rate of 60 Hz and present those sampled frames at 60 Hz).
- You might wanna repeat the test with 144 Hz refresh rate as well. You'll still get judder (non-smooth scrolling), but with `vsync = on` the speed of the gameplay and the audio should not change at all.

So this mode of operation is all wrong on `main`. The game gets throttled to 60 Hz, so there is no way to play games without tearing _but at the original speed_ on a 60 Hz monitor.

On this branch, you have the following options on a 60 Hz fixed refresh rate display:

- Run the game at the correct speed with no tearing, but with judder instead of smooth scrolling (`vsync = on`)
- Run the game at 85% speed with no tearing and 100% perfect smooth scrolling (`vsync = on` and `dos_rate = host`). Note the music will still play at 100% speed in Pinball Dreams.

On main, you can only do the second option. Accomplishing the first option is impossible.

Some extra testing notes:

- Make sure to test in fullscreen as well with the compositor is bypassed (you'll always get vsync "for free" with compositor-level triple-buffering). You can easily test whether compositor-level is happening with Pinball Dreams or Dark Forces (the "turn around test"). With `vsync = off` and compositor level vsync disabled, you'll notice a lot of tearing.
- It would be ideal on both X11 and Wayland and whatever else is available on your system. Just some common permutations, we can't possibly support every weirdo Linux setup.

Hope this helped, and I appreciate some of this info might be confusing if you're not super into frame presentation, running games at correct speed, getting zero tearing, perfect smooth scrolling, etc. 😄

VRR monitors are the "holy grail" for emulators because then the monitor can sync to any weird refresh rate with no tearing. The MAME people and the WinUAE guy has been saying the same to people complaining about vsync quirks for a good 10 years now: just buy a VRR monitor, you'll never look back 😆 But I think it's still important to give people a few options who are on fixed refresh displays, and yeah, most of those options are various levels of compromises.

weirddan455:
> It's not really a problem as running the emulated VGA card at custom rates (other than the refresh rates used by actual hardware for a given video mode) is a hack. Please refer to dos_rate /? for more details.

Ok, that makes sense. That's not a new setting either so no problem with this PR. I had just never used it before.

> On this branch, you have the following options on a 60 Hz fixed refresh rate display:
>
> - Run the game at the correct speed with no tearing, but with judder instead of smooth scrolling (vsync = on)
> - Run the game at 85% speed with no tearing and 100% perfect smooth scrolling (vsync = on and dos_rate = host). Note the music will still play at 100% speed in Pinball Dreams.
>
> On main, you can only do the second option. Accomplishing the first option is impossible.

I just tested again at 60hz. I did before as well but I wasn't sure what to look for so thanks for clarifying. On `main` with `presentation_mode = vfr` and `vsync = on`, it slows the game down as you noticed. It also slows the audio down which is different from what you saw.

This branch has the same behavior with Pinball Dreams (as far as I can tell) as `presentation_mode = cfr`. In both `main` and this branch, I can turn vsync on or off and it runs at the correct speed but with jerky scrolling. I don't notice tearing with vsync off but I'm not that sensitive to it. Dark Forces may be a better test for tearing.

So that's all fine and dandy and I'm in agreement with simplifying things and removing VFR. I'm just not sure what improvement I should be looking for or if I'm checking for regressions.

> Hope this helped, and I appreciate some of this info might be confusing if you're not super into frame presentation, running games at correct speed, getting zero tearing, perfect smooth scrolling, etc. 😄

Yeah, I've never looked into this stuff before and it's kind of a lot. At least the new setting are a bit simpler. I'll test Dark Forces here in a bit and report back.

weirddan455:
> Run the game at the correct speed with no tearing, but with judder instead of smooth scrolling (vsync = on)

I forgot to mention. You said this was impossible on `main` but I can accomplish this with `presentation_mode = cfr` or `presentation_mode = auto` and `vsync = on`. `presentation_mode = auto` runs at something called "throttled VFR" (which is different than regular VFR) according to the logs. Maybe this is different on Windows or with your specific monitor resolution.

```
2025-08-05 00:03:38.003 | DISPLAY: VGA 320x200 256-colour graphics mode 13h at 70.086 Hz throttled VFR, scaled to 1440x1080 pixels with 1:1.2 (5:6) pixel aspect ratio
```

I'm probably missing something here but at least I can confirm your PR does away with all the weird modes and does the right thing.

weirddan455:
Tested Dark Forces at 60hz:

> Use 60k cycles and set vsync = on which result in VFR mode being picked.

Again, this game sets to "throttled VFR" for me which does not have any of the problems you described. I have to set `presentation_mode = vfr` to reproduce these.

> There will be a lot of crackling in the PCM playback when the logo appears in the intro.

Difficult for me to reproduce. I heard one or two pops but this only happens maybe 50% of the time. The other 50% it sounds perfectly fine. It's also fine on this branch.

> Now hold the left or right cursor keys to keep turning around. Note how the music plays slower and the screen updates are completely smooth.

This is reproducible only with `presentation_mode = vfr`. On this branch, there's no problem with vsync on or off.

> Throttling a 70 Hz VGA game to 60 Hz also creates weird latency and "delayed input processing" issues. If you keep playing, you'll notice movement doesn't always stop when you release the movement cursor keys; sometimes you keep going for a bit longer for a fraction of a second after you release the keys.

This game just has really floaty controls. You keep some of your forward momentum after releasing the move key. I believe you when you say there's latency issues but it's a bit hard for me to say for sure.

> With presentation_mode = cfr and vsync = on, the crackling in the intro

Can't reproduce this at all. CFR sounds fine to me 100% of the time.

Based on your comments, it sounds like `presentation_mode = auto` was giving you regular VFR rather than the throttled VFR I get. Regular VFR does seem to have lots of issues so I can see why this was bothering you. I wouldn't have ever noticed with the default settings as throttled VFR does not have any of these bugs.

johnnovak:
Thanks for the thorough testing and reports @weirddan455 and confirming the new system works as expected.

I've just retested the two games on Windows with a fixed 60 Hz refresh rate to double check things. It seems you're getting different (better) results on Linux with the previous system that what I'm getting on Windows and macOS. The good news is these OS-specific variations in behaviour seem to be gone (hopefully completely) in my reworked system 🥳 It's good to see my new way of doing things is more OS-agnostic! 🎉 

I think the payback for all this effort is:

- Removing (or vastly reducing) the OS-specific variations in behaviour
- A simpler model -- 99% of users should either use the stock settings (people on VRR screen, people only using windowed mode, or who just don't care about tearing), or `vsync = on` (people on 60 Hz displays who care about tearing).
- Much simpler code with all the auto-stuff removed. I have no idea how most of those heuristics worked! 😅 
- VFR was just a weird thing that is now completely eradicated
- Improved vsync'ed performance that I could measure on both macOS and Windows with Quake bench
- Generally simplifying/streamlining the presentation related config settings

---

Here are the results of my re-tests, just to document things:


> I just tested again at 60hz. I did before as well but I wasn't sure what to look for so thanks for clarifying. On `main` with `presentation_mode = vfr` and `vsync = on`, it slows the game down as you noticed. It also slows the audio down which is different from what you saw.

> I forgot to mention. You said this was impossible on main but I can accomplish this with presentation_mode = cfr or presentation_mode = auto and vsync = on. presentation_mode = auto runs at something called "throttled VFR" (which is different than regular VFR) according to the logs. 

I've just tested Pinball Dreams again on Windows with my VRR monitor set to fixed 60 Hz (G-SYNC off).

- 0.82.2, stock config + `vsync = on` -- Throttled VFR gets picked, just like for you. Scrolling is juddery as expected but otherwise gameplay has the normal speed and so does the music playback. Very similar to this branch with `vsync = on`, but I think the frame pacing is a bit smoother on my branch. Could be placebo, though 😆I think there are skipped frames on 0.82.2 and some "breathing" behaviour (periodic uneven frame pacing).  In any case, the results are close enough for most people.
- 0.82.2, explicit `presentation_mode = cfr` & `vsync = on` -- All hell breaks loose. Game is running slower at 60 Hz, controls are weird and laggy, and there are constant severe audio dropouts (which makes the music appear to play slower, yeah). The scrolling appears to be smooth in-game, but not during the intro. Audio dropouts get more severe when the screen is being updated. There's also some weird "acceleration" going on every now and then, it's as if the framerate was "pulsating" and had sudden periodic surges when it's "rushing ahead" for a split second, then back to normal.
- 0.82.2, explicit `presentation_mode = vfr` & `vsync = on`  -- Very similar to the above CFR case, all sorts of weirdness is happening.
- 0.82.2, same CFR & VFR tests but with `vsync = off` -- Interestingly, most of the issues are still present, including the severe audio glitching. I would've expected disabling vsync to help with these issues, but nope.

> This branch has the same behavior with Pinball Dreams (as far as I can tell) as presentation_mode = cfr. In both main and this branch, I can turn vsync on or off and it runs at the correct speed but with jerky scrolling. I don't notice tearing with vsync off but I'm not that sensitive to it. 

As per above, only the stock settings are usable or stock + `vsync = on` in 0.82.2 on my Windows 10 system. All the other `presentation_mode` and `vsync` combinations result in a very bad experience, rendering the emulator virtually useless.

---

Dark Forces results on fixed 60 Hz on Windows 10:

- 0.82.2, stock settings + `vsync on` --- Throttled VFR gets picked, and things work fine, just like on this branch with `vsync on`. Game plays at the correct speed, no tearing, no audio glitches or slowdowns.
- 0.82.2, `presentation_mode cfr`, `vsync off` --- All is well, just with tearing.
- 0.82.2, `presentation_mode cfr`, `vsync on` --- Game is slowed down by about 15% (locked to 60 Hz), terrible audio glitching (but no tearing).
- 0.82.2, `presentation_mode vfr` -- The `vsync on` and `vsync off` rest results are same as with the CFR mode. Weird, but that's how it is 🤷🏻 

So same deal, only stock or stock + `vsync on` are usable.

> Based on your comments, it sounds like presentation_mode = auto was giving you regular VFR rather than the throttled VFR I get. Regular VFR does seem to have lots of issues so I can see why this was bothering you.

Nope, I'm getting throttled VFR too. Just any other explicitly set `presentation_mode` is completely useless on my system when I'm using a fixed 60 Hz refresh rate.

---

The other thing is that when my monitor is in VRR mode and 120 Hz desktop refresh rate, then it accepts input in the 48-120 Hz VRR range. But it's really best to feed VRR monitor with a steady frame rate, and in VFR mode we can get big FPS jumps because of the VFR behaviour in programs that don't update the screen constantly. So that's another negative with the old system. That cannot be fixed; I don't think you can detect VRR mode at all without querying the GPU drivers directly, so our best bet is to just present rates at a steady rate at the DOS rate by default, which is what the new system always does.

src/gui/sdlmain.cpp
```diff
@@ -2261,6 +2246,41 @@ std::optional<uint8_t> init_gl_renderer(const uint8_t flags, const int render_wi
...
+								? "1"
+								: "0";
+
+	if (SDL_SetHint(SDL_HINT_RENDER_VSYNC, hint_str) == SDL_FALSE) {
+		LOG_WARNING("SDL: Failed setting vsync mode to %s (%s): %s",
+					to_string(mode),
+					hint_str,
+					SDL_GetError());
+		return;
+	}
```
weirddan455: This is wrong. I'm pretty sure setting a hint only works if you do so before creating the window and/or renderer. We're doing it after the fact. It's better to use `SDL_RenderSetVSync()` here.

weirddan455 (approved):
Reviewed the code. Only one error stood out to me. I'm not that familiar with our presentation code but I'm happy the complicated parts I _really_ didn't understand have been removed. Very nice job simplifying things.

src/gui/sdlmain.cpp
johnnovak: I think you're right; I've just moved this code around, but most likely it was called before creating the window.

I'll change it and test vsync with texture output (haven't done too much texture testing).

johnnovak: Thanks, that's fixed. I've realised I need to dick around for a bit more with the `texture` output as I haven't tested that too much. It's almost good now, just need to test on Windows again, etc.

FeralChild64:
After doing various tests I think I've found a problem - if I switch my monitor refresh rate to 60 Hz and use the following settings:

```
vsync               = on
presentation_mode   = dos-rate
```

applications/games do not start - tried DOOM, DOS Navigator, LaunchBox for MS-DOS, Terminal Velocity, and others. Some display a couple of normally looking messages and then freeze, some don't display anything.
Enabling/disabling the adaptive sync makes no difference. Problem happens both in a window and if using a fullscreen. Problem happens with both `SDL_VIDEODRIVER=wayland` and  `SDL_VIDEODRIVER=x11` environment variable.

Setting `vsynf = off` or `presentation_mode = host-rate` makes the problem disappear. Setting the monitor refresh rate to 95 Hz or 144 Hz also fixes the problem.

Tested using KDE Plasma 6.4.4 running on Wayland. I was also able to reproduce the problem using X11.

johnnovak:
Thanks for the extra testing @FeralChild64 👍🏻 

I wasn't able to reproduce this either on macOS or Windows 10 using `presentation_mode = dos-rate` and `vsync = on`. Tried both windowed and fullscreen at 48-120 Hz VRR, fixed 120 Hz, and 60 Hz. This confirms OS and driver specific differences and quirks do exist, so it's always good to retain the ability for full manual control, especially for completely disabling vsync as that is what can cause these issues on some setups.

In my testing, the games ran at 60 Hz (e.g., Pinball Dreams, a 70 Hz VGA game, ran in slow motion), so the main loop was throttled because of blocking vsync waits. Because of that the `presentation_mode = dos-rate` + `vsync = on` combo is conceptually just wrong (even if it kinda/sorta works, like on my systems); in vsync'd mode only `host-rate` makes sense. I'm inclined to simply disallow this combo and do not allow enabling `vsync` with `dos-rate`.

Why not kill `presentation_mode` altogether? Well, I'd love to make presentation mode an internal-only thing, controlled by `vsync` alone. But as per the release notes draft, the `presentation_mode = host-rate` + `vsync = off` combo might be still needed on those "vsync locked" setups that do a blocking wait on present, no matter what. So I think that's worth keeping, just in case.

Here's a table to recap the various combinations. The only change I'm gonna do is to disallow the 🔴 option. And again, 99% of users don't need to understand any of this; people on fixed refresh displays might want to set `vsync = on`, and that's it. For everybody else, the stock settings will JustWork(tm) (except for the unfortunate souls plagued with the issue marked with🔴) 

| `presentation_mode` | `vsync` | Comments |
| ---- | ---- | ----|
| `auto` | `off` | 🟢 will use `dos-rate` (useful for VRR displays or if the user doesn't care about vsync) | 
| `auto` | `on` | 🟢 will use `host-rate` (useful for fixed refresh displays) |
| `dos-rate` | `off` | 🟢 same as `auto` + `vsync off` |
| `dos-rate` | `on` | 🔴 conceptually wrong and causes issues, therefore `vsync` is disallowed in this case |
| `host-rate` | `off` | 🟡 niche option: might be only useful on fixed 50/60 Hz setups that enforce vsync and always block on present, no matter what |
| `host-rate` | `on` | 🟢 same as `auto` + `vsync on` |

0mnicydle:
Actually dos-rate with vsync on may still have a valid use case here... I don't know how much of this applies to Freesync as I personally have only ever been satisfied with true NVIDIA G-Sync (display has G-Sync module) when it comes to VRR tech, but G-Sync + Vsync on is actually the intended way for G-Sync to be used for 2 reasons, only one of which potentially applies here: 1) it enforces vsync when frame rate would exceed VRR range to prevent tearing (which in the case of DOSBox with forced dos-rate presentation is undesirable when max refresh rate is below dos-rate due to excessive blocking behavior), and more pertinently 2) it enforces vsync when below the max refresh rate to always prevent tearing. At first vsync may seem unnecessary when using VRR as the display adapts to the incoming rate already, but the technical reality is some very minor tearing is actually still possible with VRR even when incoming rate is below max refresh if vsync is not enabled due to the not absolutely instantaneous response time of the adaptation mechanisms; it's incredibly fast, but not fast enough to absolutely guarantee no tearing ever; I've actually seen it happen with my own eyes. Enabling vsync will ensure tearing will absolutely not happen under any circumstances at the expense of some microseconds (I think it actually be as much as ~1ms) of additional latency, as opposed to several milliseconds when using vsync on a FRR display (or at the max refresh rate of a VRR display).

So someone might want to override the auto host-rate when using vsync in order to get guaranteed tear free experience, but with the accurate emulation rate in order to get ideal frame pacing (which host-rate will not provide). Now to be fair the issue of VRR not always being enough to prevent tearing even when within VRR range is usually associated with sudden and large fluctuations of incoming frame rate, which shouldn't really be happening in DOSBox using dos-rate. When incoming rates are consistent VRR alone will likely be sufficient to ensure no tearing, but it is however _normal_ practice among the well informed to enable vsync even when the frame rate is expected to stay below max refresh, and VRR users using other emulators may naturally expect this functionality.

All this being said, I think anyone who really wants to could still defeat any logic meant to prevent this config by having vsync off in the conf and enforcing it via driver.

[Please see "G-SYNC + V-SYNC On" under "Within G-SYNC Range" here for corroboration.](https://blurbusters.com/gsync/gsync101-input-lag-tests-and-settings/2/)

johnnovak:
@0mnicydle VRR + vsync is mainly only useful with modern games when the present rate might go over the VRR range. So with a 48-120 Hz VRR range if the present rate is above 120 Hz, VRR might misbehave (flicker, delays, who knows).

With DOS games that's not possible. All VESA modes games use stay below 100 Hz, and non-VESA modes are between 50 and 70 Hz, roughly. The are literally no VRR monitors in existence with an upper VRR limit lower than 100 Hz.

So that's not an issue, doesn't apply to us. We don't need vsync for VRR operation.

When people force custom refresh rates with `dos_rate`, they just need to subtract 3 Hz or so from their upper VRR range limit "to be sure". So for example `dos_rate = 117` for a 48-120 Hz, monitor. Or set it much lower; dancing just at the edge of the VRR limit with these handful of late 90s shooters is a personal choice 😎 

> but it is however normal practice among the well informed to enable vsync even when the frame rate is expected to stay below max refresh, and VRR users using other emulators may naturally expect this functionality.

Nah, not for emulators. For modern games, yes. Emulators are different, and usually emulator authors warn people not to dick around with framecappers and such tomfoolery when using their software 😆 

Using frame cappers with VRR monitors is a somewhat standard practice, again, for modern 3D games (like I don't care about more than 60 Hz, so I might as well just framecap to 60 Hz on my VRR monitor and that's it).

PS: Your posts remind me of this saying:  "I would have written a shorter letter, but did not have the time." Can you please try to be on point and more succinct in the future? 😄 Thank you.

0mnicydle:
Emulators don't need external frame limiters because they run at fixed rate already. And it is normal btw to use vsync when running emulators that run below max refresh on the VRR monitor. It's in fact usually recommended as that is the only way to ensure tearing will not happen under any circumstances.

[See here for just one example.](https://www.libretro.com/index.php/category/g-sync/)

johnnovak:
> Emulators don't need external frame limiters because they run at fixed rate already. And it is normal btw to use vsync when running emulators that run below max refresh on the VRR monitor. It's in fact usually recommended as that is the only way to ensure tearing will not happen under any circumstances.
> 
> [See here for just one example.](https://www.libretro.com/index.php/category/g-sync/)

That's something RetroArch/libretro specific. They don't just say "enable vsync", they tell users to enable a _bunch of settings_, which includes setting vsync in their preferences. Probably something very specific to their software.

Whether RA under the hood _literally_ requests vsynced operation from the graphics driver as a result of this combination of RA settings is anyone's guess.

But I don't know, and literally don't care how RA does things 😄 

image

With other emulators I'm using, no vsync enabled anywhere (driver settings, emulators, etc.) yields perfectly smooth scrolling. That's VICE and WinUAE in my personal experience, and Toni of WinUAE tells everyone to just disabled vsync on VRR monitors. It's not needed.

That's the whole point—on VRR monitors, the present rate dictates the "sync pulse".

> It's in fact usually recommended as that is the only way to ensure tearing will not happen under any circumstances.

Patently false according to my experience, and also doesn't make much sense if you understand how VRR works.

0mnicydle:
> Whether RA under the hood literally requests vsynced operation from the graphics driver as a result of this combination of RA settings is anyone's guess.

No need to guess, it still works without it and most people may never experience any difference. It's not a matter of internal logic, it's just a matter of ideal config.

But in any case if you don't want to believe tearing can happen with VRR even within the VRR range I'm not gonna argue. I will however recommend you read the whole blur busters article I linked earlier if you only skimmed it. The importance of the part I drew attention to might not be obvious without the additional context of what circumstances exactly are being described.

johnnovak:
> The importance of the part I drew attention to might not be obvious without the additional context of what circumstances exactly are being described.

I did read that a while ago, and now I read it again.

That article is about G-SYNC + vsync on supposedly doing a better job at smoothing out _wild_ framerate fluctuations compared to just G-SYNC with no vsync.

Quoting:

> G-SYNC + V-SYNC “Off” disables the G-SYNC module’s ability **to compensate for sudden frametime variances**, meaning, instead of aligning the next frame scan to the next scanout (the process that physically draws each frame, pixel by pixel, left to right, top to bottom on-screen), G-SYNC + V-SYNC “Off” will opt to start the next frame scan in the current scanout instead. **This results in simultaneous delivery of more than one frame in a single scanout (tearing).**

Yeah, and that makes some sense if true.

So... emulators... Emulators produce a steady stream of frames as you just said a few posts above (when you correctly mentioned no framecappers are needed with typical emulators).

So why would we need to enable vsync if we produce a steady stream of frames anyway? 😄 "Just in case" is not a good reason.

Supposing that article from 2017 was even true in the first place and still relevant 8 years later, in 2025. Currently, there are virtually zero monitors available with a hardware G-SYNC module for purchase. All current models are either FreeSync or "G-SYNC lite" implemented in software.

In any case, the article might be still true for modern games with large FPS fluctuations, but I fail to see how it's relevant in an emulator context when the FPS is fixed. Again, zero issues on my literally not even G-SYNC certified Dell monitor with a bunch of home computer emulators. I never use vsync on, I always get 100% smooth 2D scrolling 🤷🏻 😆

0mnicydle:
Yes and I mentioned that earlier (consistent frame rates not presenting as much difficulty). The point is, it doesn't hurt to leave it enabled, it can only potentially help, and is the "ideal" config in the sense you can rest assured tearing will never happen and it can just become standard procedure for some users.

Also it's true G-Sync monitors are not as common as they used to be, but they are still being made, and may become a little more common again soon as the module is required for ULMB 2. Personally I really only care about true G-Sync and will be sad if and when the day comes it's truly dead. The Freesync monitors I've used are all very sloppy with the frame pacing, in my experience. They just kind of just get close and fluctuate in weird ways, like hover in the general vicinity or close to some multiple of the incoming rate.

johnnovak:
> The Freesync monitors I've used are all very sloppy with the frame pacing, in my experience. They just kind of just get close and fluctuate in weird ways, like hover in the general vicinity or close to some multiple of the incoming rate.

Yeah, my DELL 2725QE is "VRR as per specified in HDMI2.1" certified, so not even FreeSync I guess 😆 I needed to click all the "are you sure to enable a non-G-SYNC-certified monitor for VRR" buttons in the Nvidia Control panel, but then I'm getting perfectly smooth, CRT-like smooth scrolling, and I have a bunch of CRTs in the house for reference. I'm quite sensitive to it when watching Amiga and C64 demos, it must be perfect... and it is, luckily! 😎 

Just sharing my experience, I believe you had bad ones with other FreeSync monitors 😄

johnnovak:
In any case, I've pushed the change that disallows `dos-rate` + `vsync on`. I'm not against removing that constraint later if there are better justifications than "gamer bros tend to do it like that" 😆 

I'm all for reducing the chance for our users to shoot themselves in the foot, and I think this restriction achieves that with probably zero disadvantages (people needing to learn they don't need to set `vsync on` if they use a VRR monitor with Staging is _not_ a disadvantage 😏).

0mnicydle:
> I'm not against removing that constraint later if there are better justifications than "gamer bros tend to do it like that" 😆

That is needlessly reductive. The reason it's the preferred config is because that's literally how the tech was designed to be used from the beginning. I've been using G-Sync since day one, and remember a time when you weren't even allowed to disable vsync when using G-Sync; the settings were tied together in the control panel and vsync was always enforced by the driver when G-Sync was enabled. Disabling vsync results in less than full functionality of G-Sync. It might mostly work in some cases like an emulator with almost perfectly stable frame rate, but it is technically out of spec. Eventually they [released a driver](http://us.download.nvidia.com/Windows/353.06/353.06-win8-win7-winvista-desktop-release-notes.pdf) that allowed users to keep vsync off because people just refused to believe how the tech was designed to work (imagine that), and there was much weeping and gnashing of teeth from "gamer bros" who didn't understand the specifics and thought vsync always means more input lag, not understanding that under conditions not applicable to our discussion, a fully uncapped frame rate even measuring in the hundreds could actually have more input lag (due to cpu run ahead) than a properly configured game running below max refresh with G-Sync and vsync active.

Anyway, if you want to keep this as-is that's fine, and I'm not trying to convince you to reverse course, but there is no need to be dismissive about this. It is [very normal, with good reason,](https://www.pcgamingwiki.com/wiki/Glossary:Variable_refresh_rate_(VRR)#Recommended_optimal_settings) to enable vsync with G-Sync even when using fixed rates below max refresh and within the VRR range. And regarding the stability of frame rates in DOSBox and emulators in general, while typically it is very solid, frame time spikes happen to everyone despite best intentions. At least on Windows... I can't speak to Linux, but Windows seems incapable of running anything perfectly at all times. When unavoidable frame time spikes occur, my preference is to not compound it with potential tearing. Just my 2c.

Grounded0:
Since this got merged, these could be closed as complete, thus reducing our doom level (200 is doomed) on the issue tracker:

- #4361 
- #4336 
- #4385

johnnovak:
> Since this got merged, these could be closed as complete,

Ah, thanks man! I've been meaning to, but then forgot.
 
Done.

> thus reducing our doom level (200 is doomed) on the issue tracker:

Indeed, we hit 200, and that opens the dimensional gates for these guys 😅 

image

I very much prefer to keep it under 200. Under 100 would be even better 😄

PR #4458, opened 2025-08-09
Dev/4118 fix volume label issue

Dustman0192:
# Description

Removed call to std::ispunct, replaced with more specific punctuation char checking.

## Related issues

# Release notes

Mounting volumes with labels containing _ and - should now work properly.

# Manual testing

1. Performed manual testing to reproduce issue.
2. Manually tested fix.
3. Added unit tests.

1. Mounted a drive B with label A_B.
4. Changed to drive B.
5. Issued DIR command.
6. Verified label was shown as A_B.

Linux Mint 22.1, Kernel 6.8.0-71-generic.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:


src/misc/string_utils.cpp
```diff
@@ -263,12 +263,14 @@ std::string strip_word(std::string& line)
...
 }
 
-void strip_punctuation(std::string &str)
+void strip_punctuation(std::string& str)
 {
-	str.erase(std::remove_if(str.begin(),
-	                         str.end(),
-	                         [](unsigned char c) { return std::ispunct(c); }),
-	          str.end());
+	auto punct_chars = {
```
johnnovak: minor: `special_chars` would be better ("punct" isn't an obvious abbreviation, and some of these chars such as `[`, `<`, `\`, etc. are not punctuation)

src/misc/string_utils.cpp
```diff
@@ -263,12 +263,14 @@ std::string strip_word(std::string& line)
...
+	        '\t', '?', '/', '\\', '|', '.', ',', ';', ':', '+', '=', '[', ']', '<', '>', '"', '\''};
```
johnnovak: minor: a new line here would improve readability

src/misc/string_utils.cpp
```diff
@@ -263,12 +263,14 @@ std::string strip_word(std::string& line)
...
+	auto is_punct = [punct_chars](unsigned char c) {
+		return std::count(punct_chars.begin(), punct_chars.end(), c) > 0;
```
johnnovak: `std::count` is not the best fit here; although it works, it muddies the intent a bit.

[std::find](https://en.cppreference.com/w/cpp/algorithm/find.html) is the perfect solution, can you please use that?

src/misc/string_utils.cpp
johnnovak: As per the [this reference](https://hwiegman.home.xs4all.nl/msdos/26295.htm), the asterisk (`*`) character is also not allowed. Could you add that?

It also mentions this:

> In addition, control codes, such as CTRL+A, CTRL+D, and so forth, are not allowed.

So I'd suggest adding a condition to your lambda function to return false if the char code is below 32. See the ASCII chart here:

https://cs.smu.ca/~porter/csc/ref/ascii.html

johnnovak:
That's a nice first-time contribution @phoenixno624, thanks for that 🎉 

Overall, this looks good. I've left a few comments, and just pointing out a few more minor things:

- The text in italics in the issue template is just a placeholder, so feel free to just replace it with your descriptions.
- It would be good to list the original ticket and the abandoned PR in the related issues section:
  - https://github.com/dosbox-staging/dosbox-staging/issues/4118#issuecomment-3172256862
  - https://github.com/dosbox-staging/dosbox-staging/pull/4134
- "Fix special character filtering in volume labels" would be slightly better title 😄 
- We don't put periods at the end of commit messages

I'd also like to suggest this for the release notes. The notes are for end users; we need to explain to them how they benefit from a given fix or feature.

> Previously, special characters, including underscores, were removed from the volume labels of mounted CD images. This caused problems with CD-ROM games that had underscores in their CD volume labels (e.g., `STAR_CRUSAD`) and were checking for the label name in their installers or when starting the game. This has been fixed—underscores and other allowed special characters are not stripped from the volume labels anymore.

I always remind people to do the release notes prior to merging because otherwise I'll need to do detective work on each ticket 6-12 months later at release time, and it can be pretty time consuming for 30-50 PRs 😄

Dustman0192:
@johnnovak Thanks for the feedback! I will integrate it and update the PR when I get a chance (probably tomorrow.)

johnnovak:
> @johnnovak Thanks for the feedback! I will integrate it and update the PR when I get a chance (probably tomorrow.)

Thank you, and there's no rush 😄

src/misc/string_utils.cpp
shermp: Or `contains()` in `support.h`

johnnovak: > Or `contains()` in `support.h`

Forgot about that, that's the winner, yeah 😄

Torinde:
Result of extensive testing of MS-DOS LABEL/DIR behavior is [here](https://github.com/dosbox-staging/dosbox-staging/pull/4134#issuecomment-2567090393):
> you can just allow ALL symbols ... except 7-9 all are even at least visible in DIR.

Dustman0192:
> Result of extensive testing of MS-DOS LABEL/DIR behavior is [here](https://github.com/dosbox-staging/dosbox-staging/pull/4134#issuecomment-2567090393):
> 
> > you can just allow ALL symbols ... except 7-9 all are even at least visible in DIR.

OK, so we should just completely remove special character stripping from drive labels, then? I can do that.

Does anyone disagree?

johnnovak:
> > Result of extensive testing of MS-DOS LABEL/DIR behavior is [here](https://github.com/dosbox-staging/dosbox-staging/pull/4134#issuecomment-2567090393):
> > > you can just allow ALL symbols ... except 7-9 all are even at least visible in DIR.
> 
> OK, so we should just completely remove special character stripping from drive labels, then? I can do that.
> 
> Does anyone disagree?

Yeah, that's actually the best and I vaguely remember we agreed in the end to keep it relaxed. We can start disallowing characters when needed, but probably the LABEL command does the validation and anything goes at the filesystem level.

Basically don't filter, yeah. That will simplify the PR a lot 😄

Dustman0192:
I have amended the PR with the requested changes. There's a typo in some of the commit messages ("strip_puncuation" *groan* ). I also removed the strip_punctuation function altogether because, from what I could tell, it wasn't used anywhere else. I can always add it back in its original state if that was in error, though.

src/dos/drives.cpp
```diff
@@ -21,10 +21,8 @@ extern char sfn[DOS_NAMELENGTH_ASCII];
 std::string To_Label(const char* name) {
 	// Reformat the name per the DOS label specification:
 	// - Upper-case, up to 11 ASCII characters
-	// - Internal spaces allowed but no: tabs ? / \ | . , ; : + = [ ] < > " '
 	std::string label(name);
 	trim(label); // strip front-and-back white-space
-	strip_punctuation(label); // strip all punctuation
 	label.resize(11); // collapse remainder to (at-most) 11 chars
```
johnnovak, line 27: Can you put a comment somewhere here about special chars? Something like this:

> The LABEL command disallows the use of a range special characters when updating the volune labels, but other commands and DOS itself doesn't seem to perform this validation, so we are intentionally lenient here.

Dustman0192: Done!

Torinde:
Lowercase letters are also "allowed" (by DIR, FAT, etc.), so there's no need for conversion to upper case.

Dustman0192:
> > // Also: this function is too strict - it removes all punctuation when _some_
> > // punctuation is acceptable in drive labels (e.g. '_' or '-').
> 
> That's no longer applicable.

I've removed the unnecessary comments and the call to uppercase the label (accidentally misnamed the commit...)

src/misc/string_utils.cpp
Dustman0192: Removed the function altogether.

src/misc/string_utils.cpp
Dustman0192: Removed the function altogether.

src/dos/drives.cpp
Dustman0192: Added the comment.

Dustman0192:
It also appears to be uppercasing the drive label on mount if it's not a CDROM drive due to line 48 in this function here:
https://github.com/dosbox-staging/dosbox-staging/blob/38d4d49938b9565c235403387fbb12b5dce4ebd6/src/dos/drives.cpp#L33-L62

Torinde:
> uppercasing the drive label on mount if it's not a CDROM drive

Reading the code and the comment above - to me it seems the reverse - if it IS a CDROM, then it's also uppercased?
> //another mscdex quirk. Label is not always uppercase. (Daggerfall)

And there is a game mentioned - Daggerfall? May be interesting to check if it works before/after the PR.

My testing was with FDD/HDD images, not with CD, so I would advise not to change anything related to CD-ROM labels for now.

johnnovak:
> My testing was with FDD/HDD images, not with CD, so I would advise not to change anything related to CD-ROM labels for now.

Not so fast! The original issue ticket is clearly about CD-ROM labels 😄

https://github.com/dosbox-staging/dosbox-staging/issues/4118

> Also see an example of an old CD game that has underscore in its label -
> http://redump.org/disc/47060/

We'll just need to do what's sensible (be relaxed with the pruning of special chars), and re-test with CD images (like the one referenced above) and a few floppy images.

Dustman0192:
> > uppercasing the drive label on mount if it's not a CDROM drive
> 
> Reading the code and the comment above - to me it seems the reverse - if it IS a CDROM, then it's also uppercased?
> 
> > //another mscdex quirk. Label is not always uppercase. (Daggerfall)
> 
> And there is a game mentioned - Daggerfall? May be interesting to check if it works before/after the PR.
> 
> My testing was with FDD/HDD images, not with CD, so I would advise not to change anything related to CD-ROM labels for now.

As I understood it, the ternary statement

`output[labelPos] = (cdrom?input[vnamePos]:toupper(input[vnamePos]));`

expands to

`if (cdrom) {
   output[labelPos] = input[vnamePos];
}
else {
   output[labelPos] = toupper(input[vnamePos]);
}`

Though all this seems to affect is the uppercasing of the label when it's _set_ during drive mount. In effect it may have made the uppercasing we were doing on label _get_ a moot operation in some cases because the label was already uppercased.

johnnovak:
## Test 1

Okay, I've tested this:

1. Download "Start Crusader (USA)" Redump image (referenced in the [original issue ticket](https://github.com/dosbox-staging/dosbox-staging/issues/4118), get the image [from here](https://myrient.erista.me/files/Redump/IBM%20-%20PC%20compatible/)).

2. Mount it using this branch and issue the `DIR` command on the D drive. Note the label is `STAR_CRU.SA`

image

3. Repeat the test with `main` — the label is `STARCRUSAD`:

image

4. The `VOL` command doesn't even work with the D drive (same behaviour on `main` and even in 0.82.2, so this is a pre-existing issue):

image

---

So something is wrong here. Although we don't drop the underscore, the name became is `STAR_CRU.SA`. It should be `STAR_CRUSAD`.

I've verified this by converting the CUE/BIN to ISO and then mounting it in macOS:

image

image

---


## Test 2

Similar deal with mounting a folder and setting a label with `-label`.

### This branch

image

### main

image

Dustman0192:
@johnnovak Will dig into this when I can! It will be a nice break from writing the text wrapper function (who knew text wrapping could be so complicated??? LOL)

johnnovak:
> @johnnovak Will dig into this when I can! It will be a nice break from writing the text wrapper function (who knew text wrapping could be so complicated??? LOL)

Yeah, not entirely sure how this can happen... but it does 😆 Some of the old code is quite messy and makes certain assumptions which are apparently no longer true after this innocent looking change.

Dustman0192:
I'm going to have to take some time to better understand the Set_Label function because that's what's adding that . character. It's not explained why, though.

Dustman0192:
After carefully reading your test results, it appears the VOL command is formatting the label correctly whereas DIR is not. So is it safe to say we should convert DIR to use whatever label formatting VOL is using? I find it odd that they're different. They both report volume label, why is it not the same??

Dustman0192:
Spent some time going down the drive labeling rabbit hole and finally found this code:
https://github.com/dosbox-staging/dosbox-staging/blob/a96011c5c70da56c79d7b9eb9a8c0053b2d93b65/src/dos/dos_ioctl.cpp#L178C1-L202C11

So the . character is a character that's inserted by SetLabel to indicate there's an extension. If there's an extension, the above code allows an extra 3 characters in the VOL label name. If there is no extension, it will truncate the drive label to 8 characters.

With this in mind, and also with the discussion in mind that we just want to allow users to label their drives with whatever they so desire, we can just alter SetLabel to make a copy of the input string and return it. We can then alter the VOL label code to allow for 11 characters in drive labels not having an extension. Thoughts?

I'm still VERY much learning how this all works, so I may be unaware of all of the dragons lurking in the dark. So by all means, if my suggested fix is going to cause problems, let me know :)

johnnovak:
Yeah, so that call you referenced is this DOS interrupt call:

- INT 21 - DOS Function Dispatcher — https://www.stanislavs.org/helppc/int_21.html
  - INT 21,44 - I/O Control for Devices (IOCTL) — https://www.stanislavs.org/helppc/int_21-44.html 
    - INT 21,44,D / IOCTL,D - Generic I/O for Block Devices — https://www.stanislavs.org/helppc/int_21-44-d.html

Ultimately we pass "minor function code" 066h in CL, but that's not even described here, lol 😆 

image

A rabbit hole, indeed! 😆 

To do this properly and not cause regressions:

1. We'd want to return the same data at the referenced DOS interrupt function level. Maybe that will fix everything up.
2. If not, we'll still need to massage the output of commands we re-implemented from scratch (`DIR`, `VOL`, etc.)

This is fiddly work, someone would need to write a small assembly or C program and compile it under DOS that calls this DOS interrupt function, and check how it behaves with these CD volume labels that contain special chars, etc. Then we'd need to do exactly the same in our DOS interrupt implementation.

This could be done in a VM (e.g., VirtualBox), using a real MS-DOS 6.22 installation.

But... I'm still not sure what actual real-world problem this would fix. We'd need _at least one_ test case where a CD-based game wouldn't install of run when special chars such as underscores are dropped from the label.

@Dustman0192 I fully understand if this would be too much for you as your first ticket (and I agree, it would be 😆). Happy to flesh out the original ticket with this info later and put it back to the backburner.

Man, I did not anticipate this becoming so complicated 😂

Dustman0192:
@johnnovak Normally I don't shy away from a challenge but, yes, give my current knowledge of the system I think this item may be a bit too much for me at the moment, hah! So go ahead and take it off my plate and put it on the back burner if you'd like :)

johnnovak:
> @johnnovak Normally I don't shy away from a challenge but, yes, give my current knowledge of the system I think this item may be a bit too much for me at the moment, hah! So go ahead and take it off my plate and put it on the back burner if you'd like :)

Yeah, that's completely understandable 😄 I will remove the "good first issue" label too 😄

johnnovak:
Updated #4118 with our findings and closing this PR. This will need further research, as per the update.

PR #4459, opened 2025-08-10
Update CONTRIBUTING.md

johnnovak:
Pretty non-controversial I hope. This came up when a new contributor started using SDL's thread support, and I've realised we're not discussing this in our contributor guide, so how would anyone know they shouldn't do this.

You can view the rendered output here:

https://github.com/dosbox-staging/dosbox-staging/blob/c967ae5920be1ad28046156e08f5c04819105fb3/docs/CONTRIBUTING.md

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -94,6 +94,27 @@ design. To clarify:
   new code. All new code must use the C++ standard library instead of C
   arrays, C filesystem functions, and OS-specific concurrency APIs if possible.
 
+- Before starting to use platform-specific, OS-level APIs, check if the C++
+  standard library or SDL provides a cross-platform implementation already.
+  They probably do. If that's not the case, look into using battle-tested
+  libraries instead of adding lots of platform-specific custom code.
+
+- In general, try to use well-established libraries for common tasks. Don't
+  reinvent the wheel unnecessarily.
+
+- Our use of SDL is generally restricted to the texture renderer fallback
+  option, and interfacing with OS-specific APIs that aren't covered by the C++
+  standard. So that means window management, OpenGL context creation, input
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -94,6 +94,26 @@ design. To clarify:
...
+  libraries instead of adding lots of platform-specific custom code.
+
+- In general, try to use well-established libraries for common tasks. Don't
+  reinvent the wheel unnecessarily.
+
+- We only use SDL for the texture renderer fallback option, and interfacing
+  with OS-specific APIs that aren't covered by the C++ standard. So that means
+  window management, OpenGL context creation, input handling, and interfacing
+  with OS-level audio and MIDI APIs. We don't use SDL's thread, timer, and
+  filesystem APIs, neither any other utilities that are available in the C++
```
weirddan455: It's good advise in general but we do use SDL's timer for the titlebar animation and we are using SDL threads in IMFC. I'm not sure that timer really has a C++ alternative. We should definitely prefer C++ standard functions for thread and filesystem though.

johnnovak: Yeah ok, I'll reword that a bit 😄

FeralChild64: One could implement a timer by creating std::thread and using std::this_thread::sleep_for. I have originally chosen SDL threads because the titlebar code was bound to SDL nevertheless.

Also our clipboard support uses SDL - simply because the C++ standard does not provide the clipboard access.

FeralChild64 (approved):
Looks OK to me.

docs/CONTRIBUTING.md
johnnovak: Softened the wording a little to discourage the use of SDL for generic tasks but leave a little bit of wiggle room there 😄

Issue #3769, opened 2024-06-15
Improve `CONFIG -h` output


johnnovak:
Improve `CONFIG -h` output so the list of possible values is indented.

`config -h machine` output (no other settings are affected):

![image](https://github.com/dosbox-staging/dosbox-staging/assets/698770/b93a5e53-0094-459b-a8cb-fe45180dda3e)

Dustman0192:
So I fiddled with this one a bit too, but the only solution I could come up with was to pad the values with a \n and then 18 spaces: 

<img width="842" height="359" alt="Image" src="https://github.com/user-attachments/assets/9fef2b5d-cfc3-4113-9481-08be5babb0a5" />

I think that solution is rather inelegant. But unless you guys know of a better way, it at least does the trick:

<img width="842" height="359" alt="Image" src="https://github.com/user-attachments/assets/f7534ae7-44b2-40c5-b658-2518a3ebfea4" />

Let me know and I'll spin up a PR and get this fix going.

johnnovak:
@Dustman0192 Sorry, that won't do it 😄 The proper solution is to implement a mini text wrapping engine that adapts the output to the screen width. That's usually 80 chars, but it can be 40 or 132 too.

Dustman0192:
@johnnovak Got it! OK, still learning the ropes of things :) This one will give me something to chew on for a while!

johnnovak:
Yeah so if you look at the screenshot above and compare it with the string in the code, you'll see it's wrapped manually with newline characters, and indented manually with spaces at the start of each line. So it would be better if descriptions like `cga_mono` and `svga_paradise` would be a _single line_ in the string. In other words, only the paragraphs and bullet point items would be separated by newline chars, then the text would be wrapped intelligently to the available width.

I've expanded the scope of this ticket (check the top description now 😄).

Dustman0192:
I accidentally posted a duplicate screenshot above. Here is the screenshot I meant to post:

<img width="1867" height="308" alt="Image" src="https://github.com/user-attachments/assets/5750ca41-763b-4ec9-b821-63a9e1106c3f" />

Is that the kind of wrapping you're looking for? Or would you prefer to have each of the "Possible values" on their own lines? I can do either or and actually did start thinking about a small wrapping engine that could wrap text within a given character length limit (say we have a max 80 char line length, then values that cause the length to go over that limit would be wrapped to the next line.)

Dustman0192:
[...]

PR #4473, opened 2025-08-15
Dev/3769 implement text wrapper

Dustman0192:
# Description
Added initial text wrapping function in string_utils.cpp to wrap text on spaces on hyphens by default. Also added some integration code to programs.cpp and more_output.h.

## Related issues
#3769 

# Release notes
Previously, certain commands such as `config -h machine` would render a list of possible values which wrap at improper breakpoints (such as in the middle of a word.) This fix is meant to improve text wrapping so it breaks in a more expected way (on spaces or hyphens) as well as neatly align with the text above it.

# Manual testing
PR is more in a "show and tell" state to make it easier to get assistance. Little testing has been done thus far other than just trying to get it to work and improve on the base solution.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist
I have:


Dustman0192:
I don't expect this PR to be merged right now. I mostly made it so we could move the discussion here and actively work on improving the base solution (which still needs work.)

src/misc/string_utils.cpp
```diff
@@ -438,3 +440,43 @@ bool is_text_equal(const std::string& str_1, const std::string& str_2)
 
 	return (index_1 == str_1.size()) && (index_2 == str_2.size());
 }
+
+std::string wrap_text(const std::string& str,
+                      const std::uint16_t max_line_length,
```
johnnovak: Just `const int`. Please read our usage of integers [here](https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/CONTRIBUTING.md#code-style) (point 9).

src/misc/string_utils.cpp
```diff
@@ -438,3 +440,43 @@ bool is_text_equal(const std::string& str_1, const std::string& str_2)
...
+                      const std::vector<char>& additional_wrap_chars,
+                      const std::uint16_t indent_length)
+{
+	std::ostringstream out_stream;
```
johnnovak: As per the guide 😄 

> But do not use C++ iostreams at all.

...unless we have to. That's not the case here, just please just do `std:string out = {}` then  `out.append(...)`, etc.

src/misc/string_utils.cpp
```diff
@@ -438,3 +440,43 @@ bool is_text_equal(const std::string& str_1, const std::string& str_2)
...
+	std::uint16_t line_len = 0;
+
+	const auto is_wrap_char = [additional_wrap_chars](const char c) {
```
johnnovak: `is_wrap_char = [&](const char c)` is fine and less noisy.

src/misc/string_utils.cpp
```diff
@@ -438,3 +440,43 @@ bool is_text_equal(const std::string& str_1, const std::string& str_2)
...
+		return c == ' ' || std::ranges::find(additional_wrap_chars, c)
+		       != additional_wrap_chars.end();
```
johnnovak: Please use `contains()` from `utils/support.h` (works with STL containers).

src/dos/programs/more_output.h
```diff
@@ -175,6 +175,19 @@ class MoreOutputStrings final : public MoreOutputBase {
...
+	                   const Args&... args)
+	{
+		const std::string formatted_string =
+			format_str(format, args...);
+		const std::uint16_t max_columns = GetMaxColumns();
+		const std::string wrapped_text  =
+			wrap_text(formatted_string,
+			          max_columns);
+		input_strings += wrapped_text;
+	}
```
johnnovak, line 191: How about `AddStringWrapped`? But we can debate naming at the very end 😄

shermp:
@johnnovak before getting into the weeds of a full review, perhaps we should discuss the high level API first?

I have some thoughts that I want to put down after work, but the TLDR will be, ignore indentation in the text wrapping function, and return a vector of strings. I will go into detail on why I think these are a good idea a bit later when I have time.

johnnovak:
This looks good so far @Dustman0192 🎉 

Okay, so let's talk about the acceptance criteria for this feature. The help text for the `window_titlebar` setting is a very good example to get started (the string is defined in `titlebar.cpp`). You can just execute `window_titlebar /?` inside Staging to see the help text, or you can execute `config -wc test.conf` to write the config as `test.conf`, then inspect the file. I've done the latter:

image

So right now this is wrapper to 80 columns manually in `titlebar.cpp` and the translation files. What we want is to wrapping become 100% automatic while keeping the indentation for the setting value descriptions.

What I'm proposing is that the the wrapper should wrap the multi-line input string line by line. So, conceptually, every input line would represent a "paragraph" to be wrapped. The automatic wrapping rules would be simple:

- Take the next line from the multi-line input string.
- If the line starts with a space character, figure out the indentation:
  - the indentation is the number of space chars before the first non-space character in the simplest case
  - but if the line contains a setting, e.g. `version=<value>:  ` (followed by two spaces), then the setting name is added to the indent length
- Wrap the input line using the auto-determined indentation
- Back to the start and onto the next line (paragraph)

Then the the message text will need to be modified so only the end of paragraph new lines remain:

image

I highly suggest to turn this into a unit test case.

Once this is done, we can talk about the handling of ANSI control sequences, but that's level 2 😆 

Of course, if you have other ideas, let us know and we can discuss.

johnnovak:
> @johnnovak before getting into the weeds of a full review, perhaps we should discuss the high level API first?
> 
> I have some thoughts that I want to put down after work, but the TLDR will be, ignore indentation in the text wrapping function, and return a vector of strings. I will go into detail on why I think these are a good idea a bit later when I have time.

Sure, I'm curious of your ideas. Please read my proposal above as well 😄

shermp:
Ok, first @Dustman0192 thanks for having a go at this!

I am going to somewhat disagree with @johnnovak on this. Not so much with the final outcome of automated, nicely formatted help/config output, but how we get there.

So, to start with, I think text wrapping should be just that. Wrap text to a certain line length, taking into account ANSI formatting. Stuff like indentation should be handled at a higher level, either at the call site, or with a dedicated layout function/class. The above example for `window_titlebar` illustrates my point, that's not just a bit of indenting, that's a table nested 3-deep.

This is where returning a vector of (padded?) lines really helps. It allows the caller to control how they actually want to layout the text, or do other stuff with it.

For example, when DOSBox Staging starts, it prints a nice welcome message in a blue box with a border. If the text wrapper returns a vector of lines, this allows you to do stuff like:

```cpp
// Write the top border
// Real code would use ASCII border characters
WriteOut("=======================================");
for (auto& line : lines) {
  WriteOut("||%s||", line);
}
// Write bottom border
WriteOut("=======================================");
```
(Also add ansi formatting for background colour etc).

With lines of wrapped text, one can then use those lines to build up a more complex layout. But these should be separate concerns IMHO.

Looking at the above example provided by @johnnovak we just about need a nestable two column table layout class...

johnnovak:
@shermp 

Right now, there's a single message ID for the `window_titlebar` message, a single string for it in the code, then N number of translated string for the N language files.

How would your proposed solution work end-to-end with the messaging and translation system while keeping this model?

shermp:
You would break the gigantic big help string into its different components with their own message id which can then laid out as necessary. Arguably this is better anyway, because you can reorder components without losing sync with translations.

Also, I'm wary about using heuristics (been there, done that in the ebook world), inevitably you find you have to start special-casing stuff.

Honestly, I'm willing to have a crack at this layout challenge if there was working text wrapping available. Maybe I might try a standalone prototype this weekend.

johnnovak:
> You would break the gigantic big help string into its different components with their own message id which can then laid out as necessary. 

That would complicate things and would result in 5x as many text messages. I think this should work like a very much simplified version of Markdown.

> Arguably this is better anyway, because you can reorder components without losing sync with translations.

IMO, that's the wrong thing to optimise for 😄 Such reordering happens rarely, but dealing with one vs 5-10 messages happens every time when working with translations and it's a hassle.

> Also, I'm wary about using heuristics (been there, done that in the ebook world), inevitably you find you have to start special-casing stuff.

Well, as much heuristics as Markdown has, just a lot less and a lot simpler 😄 

> Honestly, I'm willing to have a crack at this layout challenge if there was working text wrapping available. Maybe I might try a standalone prototype this weekend.

Sure, but I really think we should keep the single string per message approach with embedded markdown like syntax. I'm interested in you coming up with a better Markdown-like syntax instead of what I proposed (literally use the number of leading spaces). But do realise we don't wanna support a _complete_ layout/formatting solution, just what's enough for our purposes. Just look at the existing text messages and extrapolate the minimum set of features from that.

shermp:
_Looks at all the calls for SetHelp..._

Ok you're right, I think we do need a solution to stick with the already written help messages.

src/misc/string_utils.cpp
Dustman0192: Done!

src/misc/string_utils.cpp
Dustman0192: Done!

src/misc/string_utils.cpp
Dustman0192: Still learning C++. Was not aware we could just use & to capture by reference all variables in the current scope :) Done!

src/misc/string_utils.cpp
Dustman0192: Done!

src/dos/programs/more_output.h
Dustman0192: Done! That is a better name.

Dustman0192:
> This looks good so far @Dustman0192 🎉
> 
> Okay, so let's talk about the acceptance criteria for this feature. The help text for the `window_titlebar` setting is a very good example to get started (the string is defined in `titlebar.cpp`). You can just execute `window_titlebar /?` inside Staging to see the help text, or you can execute `config -wc test.conf` to write the config as `test.conf`, then inspect the file. I've done the latter: (**remaining text redacted for brevity**)

Thanks! Though to continue presenting problems, issue is the input we're trying to format is a dynamically generated string. It starts out here as a format string:
https://github.com/dosbox-staging/dosbox-staging/blob/3382ddb39b9ce807f5551d4156d2eb02ade6a50d/src/misc/programs.cpp#L1006-L1007

And the possible values start out here:
https://github.com/dosbox-staging/dosbox-staging/blob/3382ddb39b9ce807f5551d4156d2eb02ade6a50d/src/dosbox.cpp#L630-L642

The possible values are joined from a vector into a string delimited by a comma and a space, then inserted into the aforementioned format string. So it's all dynamically generated and there are no newlines or anything in there. It's basically just one long string with no newlines to indicate where a line would end.

This is the result for possible values _without_ any text wrapping. It's what ultimately gets printed to the screen:

`[37;1mPossible values:[0m  hercules, cga_mono, cga, pcjr, tandy, ega, svga_s3, svga_et3000, svga_et4000, svga_paradise, vesa_nolfb, vesa_oldvbe`

Those <?> characters are ESC characters.

Dustman0192:
I'm wondering if we could just look for the first index of %s and that would be our indent length? Though the special characters in that string still present a problem...

It appears those may be escape sequences? So perhaps we can skip them. Do escape sequences always start with a 0x1b and end with an m? If so I may be able to write up something to disregard them (if we don't have something that does that already.)

Dustman0192:
I've added a new function called "get_label_width_in_cols". This relies on the concept of a "label" which is whatever is in the left display column of an output string. It attempts to determine the width, in character columns, of that display column by ignoring ANSI escape sequences and then stopping on the first occurrence of a % character indicating a potential C-style formatter. It ain't pretty but it seems to work OK... lol

src/misc/string_utils.cpp
johnnovak: Yeah, it's pretty handy 😄

johnnovak:
@Dustman0192 

I appreciate all the intricacies of the config system can be a lot to take in and there's a learning curve to all the DOSBox Staging specific text output handing, so let me try to give you a high level overview first:

- The strings can be added to the message system with `AddString` (as you've seen), or string literals can be printed to the DOS console directly with just `WriteOut` (search for `WriteOut` to see).
- The strings can have ANSI escapes in them via some special markup (e.g., `[color=light-green]imgmount[reset]`, search for `[color=` in the codebase to see further examples).
- When these strings are printed to the console, we need to call `convert_ansi_markup` to turn the ANSI markup into escape sequences, then the resulting string can be passed to `WriteOut` or `CONSOLE_Write`.
- Different subsystems treat the same string message differently (e.g., when a message is printed to the console, the ANSI sequences will result in coloured output, but when the same message is written to the logs or the configuration file,  the ANSI sequences are stripped).

> It appears those may be escape sequences? So perhaps we can skip them. Do escape sequences always start with a 0x1b and end with an m? If so I may be able to write up something to disregard them (if we don't have something that does that already.)

So I've just answered your question; you're operating on the string that contains ANSI sequences and went through the `convert_ansi_markup` transform.

Hopefully you see the full picture better now, and my earlier comments about having a single generic wrapping solution that caters for all these use case makes more sense. For now, I think it's a safe choice to assume the string can contain the ANSI escape sequences but not our ANSI markup tags.

To keep things simpler for you, let's just focus on getting the `config -h <setting_name>` output right in the DOS console, then I will help with integrating the wrapping solution in all the other areas of the codebase 😄 But that will come in another PR, let's keep this small and focused 😄 

The ANSI sequence skipper seems okay, but you'll need to separate that out and integrate it into the text wrapping function somehow as well. Escape sequences can happen anywhere, e.g., some words might be surrounded by ANSI escapes in the middle of the line to change the color of a few words.

> I'm wondering if we could just look for the first index of %s and that would be our indent length? Though the special characters in that string still present a problem...

Okay, so I missed something when I reviewed your code previously. We should not do the wrapping when adding the strings with `AddString`. Instead, wrapping should happen at the _output_ only, so inside `WriteOut`, potentially. It gets a bit complicated, and I don't want to overload you with even more details, so for now just insert your wrapping logic into `WriteOut`, so everything that passes through it gets wrapped. You can call `INT10_GetTextColumns()` to get the width of the DOS console—that will be your wrap length.

Again, anyone reading this, this might not be the final solution, but I can't talk about all concerns to @Dustman0192 now because then this would be five times as long 😆 One step at a time, then we can refine 😎 

---

Next, about the wrapping rules. We definitely want to do the single string approach, so can you please read my description of the [proposed algorithm](https://github.com/dosbox-staging/dosbox-staging/pull/4473#issuecomment-3190502831) from this comment again and let me know if something's unclear? For now, let's just proceed with what I suggested because that's solid and will work in all use cases.

So, to recap:

- If a line starts with at least one space, then a "word" and a semicolon, we'll wrap `Description text...`, indented.
  - pattern: `  some_config_param: Description text...`
- If a line doesn't start with space, but contains some leading text followed by a colon and two spaces (`:  `), we format the remaining text indented. This will be out `Possible values:  hercules, cga_mono...` case, for instance.
- If neither criteria is true, we just wrap normally.

Then for this input string

```
Possible values:  hercules, cga_mono, cga, pcjr, tandy, ega, svga_s3, svga_et3000, svga_et4000, svga_paradise, vesa_nolfb, vesa_oldvbe
Deprecated values:  vgaonly
```

We'll generate this output, taking advantage of the second rule:

```
Possible values:  hercules, cga_mono, cga, pcjr, tandy, ega, svga_s3,
				  svga_et3000, svga_et4000, svga_paradise, vesa_nolfb,
				  vesa_oldvbe
Deprecated values:  vgaonly
```

---

@Dustman0192 I know this can be a lot to take in, so please read this a few times 😄 I believe all the info you need is here and in my previous comment. If something is still unclear, maybe let's chat on Discord sometime in real-time!

Dustman0192:
@johnnovak Thanks for taking the time to explain these things! I'll read it a few times, mull it over, think about it, and try to internalize it this weekend.

So far working on dosbox has been really fun. It's the perfect blend of old school coding with having to emulate DOS behavior and new school tools and concepts.

It's a lot to take in and learn, for sure, but that will come with time and exposure to the code base, just like any decent sized and sophisticated project. Something I've had to do many times in my career. Dig dig dig and ask a lot of questions! :)

johnnovak:
> @johnnovak Thanks for taking the time to explain these things! I'll read it a few times, mull it over, think about it, and try to internalize it this weekend.

No problem 😄 

> So far working on dosbox has been really fun. It's the perfect blend of old school coding with having to emulate DOS behavior and new school tools and concepts.

Great to hear you're having fun. There's surely lots of things to learn here 😄 

> It's a lot to take in and learn, for sure, but that will come with time and exposure to the code base, just like any decent sized and sophisticated project. Something I've had to do many times in my career. Dig dig dig and ask a lot of questions! :)

Yeah, absolutely, these things take time. I'm just trying to narrow down the scope for you a bit so you can focus on one thing without getting frustrated by trying to do everything at once. Getting familiar with the intricacies of the config and message system can take months and even years (it was more like "years" for my case 😄).

johnnovak:
Closing as @Dustman0192 mentioned to me on Discord he won't have time to work on this in the foreseeable future.

PR #4475, opened 2025-08-18
ci: Fix getting the current version with `get-version.sh`

johnnovak:
# Description

I deleted the obsolete `include/version.h` last week, but forgot to update the `get-version.sh` script that still attempted to fetch the current version from this file.

This had the unfortunate consequence of the version being dropped from the release package file names which in turn broke the dev downloads page.

For instance:

image

This has been broken for almost a week now and I'm kinda disappointed no one in our Discorded has reported it yet...

Wake up, people! ⏰ 😆 

# Manual testing

## Case 1

`CMakeLists.txt`

image

**Script output:**

image

## Case 2

`CMakeLists.txt`

image

**Script output:**

image

---

CI's looking good:

image

CI generated artifact looking good:

image

- [ ] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/scripts/ci/get-version.sh
+++ b/scripts/ci/get-version.sh
@@ -30,15 +30,20 @@ fi
 ROOT=$(git rev-parse --show-toplevel)
 
 VERSION=$(
-    grep "#define\\s*DOSBOX_VERSION\\s*\"" "$ROOT/include/version.h" \
-        | cut -d"\"" -f2
+    grep "project(dosbox-staging" -A 2 "$ROOT/CMakeLists.txt" \
+        | grep "VERSION" | sed "s/\\s*VERSION\\s*//g" | cut -d"\"" -f2
+)
+
+SUFFIX=$(
+    grep "set(DOSBOX_VERSION" "$ROOT/CMakeLists.txt" | cut -d" " -f2 \
+        | sed "s/\${PROJECT_VERSION}//g" | sed "s/)//g"
 )
 
 GIT_HASH=$(git rev-parse --short=5 HEAD)
 
 case $1 in
-    version)          echo "$VERSION" ;;
+    version)          echo "$VERSION$SUFFIX" ;;
     hash)             echo "$GIT_HASH" ;;
-    version-and-hash) echo "$VERSION-$GIT_HASH" ;;
+    version-and-hash) echo "$VERSION$SUFFIX-$GIT_HASH" ;;
     *)                usage; exit 1 ;;
 esac


weirddan455:
Ran the script locally:

```
$ scripts/ci/get-version.sh version
0.83.0
```

Shouldn't it say `0.83.0-alpha` instead?

johnnovak:
> Ran the script locally:
> 
> ```
> $ scripts/ci/get-version.sh version
> 0.83.0
> ```
> 
> Shouldn't it say `0.83.0-alpha` instead?

No, because the version is defined in a single place in `CMakeList.txt`, and that version format doesn't support postfixes like `-alpha`:

image

I did look into this, but the solutions to support that via CMake are overly complex, and I'd say for very marginal benefits. So we've dropped the `-alpha` postfixed entirely—we just need to know what our latest stable release is, then the alpha comes after that.

PS: @weirddan455 I guess we could bring back `include/version.h` to bring back the `alpha` postfix, but I prefer to handle all the config templating stuff in CMake. They might add support for version suffixes later, then we could do it via CMake directly: https://gitlab.kitware.com/cmake/cmake/-/issues/16716

PS2: Now that I think about it, maybe we'll need the suffixes for the release candidates? 🤔 Maybe I'll just bring back `include/version.h`, then we'll need to manually bump that in sync with the other version number definitions (we either use pure CMake features, or we put in the source code; we can't call `get-version.sh` on Windows).

PS3: Lol, I have a memory like a sieve 😆 Plus reading comprehension check failed, too 😂 Look at this:

image

So I'll just need to extend `get-version.sh` to read the suffix from `DOSBOX_VERSION` a few lines below 😆 Making the change today.

johnnovak:
@weirddan455 
Fixed now. Thanks for flagging the issue. I've updated the manual testing section.

scripts/ci/get-version.sh
```diff
@@ -30,15 +30,20 @@ fi
...
 
 VERSION=$(
-    grep "#define\\s*DOSBOX_VERSION\\s*\"" "$ROOT/include/version.h" \
-        | cut -d"\"" -f2
+    grep "project(dosbox-staging" -A 2 "$ROOT/CMakeLists.txt" \
+        | grep "VERSION" | sed "s/\\s*VERSION\\s*//g" | cut -d"\"" -f2
+)
+
+SUFFIX=$(
+    grep "set(DOSBOX_VERSION" CMakeLists.txt | cut -d" " -f2 \
```
weirddan455: "$ROOT/CMakeLists.txt"

johnnovak: Yeah, fixing.

FeralChild64 (approved):

weirddan455 (approved):
Looks good now

PR #4486, opened 2025-08-25
Add seek-only mode for disk noise emulation, fix hdd noises missing i…

dividebysandwich:
…f no io-delay, add placeholder fdd spin files.

# Description

Add "seek-only" mode for disk noises. 
Fixed a bug that caused hdd noises to go silent when copying to a floppy while io-speed set to maximum
Added fdd spin noise placeholders. 

# Release notes

The disk noise emulation now supports a "seek-only" mode in addition to "on" (both spin and seek noises) and "off" (no disk noises), for both floppy and hard disk drives.

# Manual testing

Manually tested copying between HDD/HDD, HDD/FDD, FDD/FDD with io speed set all the valid values.
Bug can be tested by setting io speed to maximum (the default) and copying from the harddisk to a floppy disk. If successful both hdd and fdd noises should be heard.

- [X] Windows
- [ ] macOS
- [X] Linux

I have:

### DIFF
--- a/src/dos/dos.cpp
+++ b/src/dos/dos.cpp
@@ -164,7 +164,8 @@ void DOS_RegisterIoCallback(std::function<void()> callback, DiskType disk_type)
 	}
 }
 
-// Unregister a given callback reference
+// TODO: Unregister should work on particular callbacks instead of all of a
+// given disk type
 void DOS_UnregisterIoCallback(DiskType disk_type)
 {
 	switch (disk_type) {


src/audio/disk_noise.cpp
```diff
@@ -491,15 +497,31 @@ static void disknoise_destroy([[maybe_unused]] Section* sec)
...
+{
+	if (has_false(mode)) {
+		return DiskNoiseMode::Off;
+	} else if (mode == "seek-only") {
+		return DiskNoiseMode::SeekOnly;
+	} else if (mode == "on") {
+		return DiskNoiseMode::On;
+	}
+	LOG_WARNING("DISKNOISE: Unknown disk noise mode '%s', defaulting to 'off'",
+	            mode.c_str());
```
johnnovak: It's better to set the list of possible enum values, then let the config system handle invalid values and defaulting (see my other comments about this a bit further down).

Then I tend to do it like this (I like to leave the disabled state last):

```cpp
    if (has_true(mode)) {
		return DiskNoiseMode::On;
	} else if (mode == "seek-only") {
		return DiskNoiseMode::SeekOnly;
	} else {
        assert(mode == "off");
		return DiskNoiseMode::Off;
	}
```

src/audio/disk_noise.cpp
```diff
@@ -532,17 +554,23 @@ static void init_disknoise_dosbox_settings(SectionProp& secprop)
...
-	bool_prop->SetHelp(
+	auto* str_prop = secprop.AddString("hard_disk_noise", OnlyAtStart, "off");
+	str_prop->SetHelp(
 	        "Enable emulated hard disk noises ('off' by default).\n"
 	        "Plays spinning disk and seek noise sounds when enabled. It's recommended to\n"
-	        "set 'hard_disk_speed' to lower than 'maximum' for an authentic experience.");
+	        "set 'hard_disk_speed' to lower than 'maximum' for an authentic experience.\n"
+	        "  off:           No hard disk noises (default).\n"
+	        "  seek-only:     Play hard disk seek noises only, no spin noises.\n"
+	        "  on:            Play both hard disk seek and spin noises.");
```
johnnovak, line 567: You didn't set the list of enum values; that's why you needed to handle the invalid values manually 😄 

So add this please:

```cpp
	str_prop->SetValues({"off", "seek-only", "on"});
```

src/audio/disk_noise.cpp
```diff
@@ -532,17 +554,23 @@ static void init_disknoise_dosbox_settings(SectionProp& secprop)
...
-	bool_prop->SetHelp(
+	str_prop = secprop.AddString("floppy_disk_noise", OnlyAtStart, "off");
+	str_prop->SetHelp(
 	        "Enable emulated floppy disk noises ('off' by default).\n"
 	        "Plays spinning disk and seek noise sounds when enabled. It's recommended to\n"
-	        "set 'floppy_disk_speed' to lower than 'maximum' for an authentic experience.");
+	        "set 'floppy_disk_speed' to lower than 'maximum' for an authentic experience.\n"
+	        "  off:           No floppy disk noises (default).\n"
+	        "  seek-only:     Play floppy disk seek noises only, no spin noises.\n"
+	        "  on:            Play both floppy disk seek and spin noises.");
```
johnnovak, line 577: As per my above comment; set the allowed enum values, please.

src/audio/disk_noise.cpp
```diff
@@ -532,17 +554,23 @@ static void init_disknoise_dosbox_settings(SectionProp& secprop)
 {
 	constexpr auto OnlyAtStart = Property::Changeable::OnlyAtStart;
 
-	auto* bool_prop = secprop.AddBool("hard_disk_noise", OnlyAtStart, false);
-	bool_prop->SetHelp(
+	auto* str_prop = secprop.AddString("hard_disk_noise", OnlyAtStart, "off");
```
johnnovak: Instead of using `OnlyAtStart`, can you change these settings to `Always`? That will cause the entire disk noise module to be destructed then re-inited when the setting is changed at runtime, which is fine (we don't need to go overboard with optimisations like not reloading the samples; destruct & reinit is fast enough in this case, so it's pointless).

Also, in the future, users will have the option to toggle disk noise sounds from the GUI (OSD).

dividebysandwich: Roger that. I'll need to test the hell out of this then though, I feel like this has a good chance of breaking if it can be enabled and disabled at runtime.

johnnovak: Yeah, it shouldn't be too hard. Probably it will just work fine, the code looks alright and the shutdown works reliably already.

dividebysandwich:
@johnnovak FYI enabling disk noises at runtime crashes. Looking into this now.

johnnovak:
> @johnnovak FYI enabling disk noises at runtime crashes. Looking into this now.

Yeah, it's good to sort these things out as it leads to more robust code, and it will be nice to be able to toggle this from the GUI without having to restart. Good luck 😄

dividebysandwich:
> > @johnnovak FYI enabling disk noises at runtime crashes. Looking into this now.
> 
> Yeah, it's good to sort these things out as it leads to more robust code, and it will be nice to be able to toggle this from the GUI without having to restart. Good luck 😄

Aaand done :)

dividebysandwich:
I'd need someone to help test on a Macintosh. IDK, one of these apple things :)

FeralChild64:
Just tested on macOS - whatever I do, I can't hear the floppy disk noise. Example config tried:

```ini
[dosbox]
hard_disk_speed             = fast
floppy_disk_speed           = medium

[disknoise]
hard_disk_noise   = off
floppy_disk_noise = on
```

Mounted floppy disk image using command:

```bat
imgmount A floppy.img -t floppy
```

Got logs:

```
IMGMOUNT: Path 'floppy.img' not found, maybe it's a DOS path
IMGMOUNT: Path 'DOS\TMP\FLOPPY.IMG' found on virtual drive C:
FAT: Mounted ../../../DOSBox/drive_c/DOS/TMP/FLOPPY.IMG as FAT12 volume with 2847 clusters
DRIVE: Unable to find A drive's volume label
```

And accessing the floppy (copying files to/from the floppy) does not produce any sound. Plus, it's very fast - like several orders of magnitude faster than a real floppy.

[edit] Same problem on Linux. The HDD part of the PR seems to be working, there are no more missing file warning logs, and the seek-only sound is great!

FeralChild64:
OK, now I understand - your disk improvements (speed, noise) only works for directories mounted as disks, not for the disk images. If this is an intentional limitation, it should be - IMHO - documented in the config description.

Otherwise - I confirm it works on macOS.

johnnovak:
> OK, now I understand - your disk improvements (speed, noise) only works for directories mounted as disks, not for the disk images. If this is an intentional limitation, it should be - IMHO - documented in the config description.

Right, I wasn't aware of that. @dividebysandwich, it would be really nice if you could enhance the disk noise emulation to work with floppy images as well.

> Otherwise - I confirm it works on macOS.

Thanks @FeralChild64, merging.

dividebysandwich:
> OK, now I understand - your disk improvements (speed, noise) only works for directories mounted as disks, not for the disk images. If this is an intentional limitation, it should be - IMHO - documented in the config description.

Aye, disk images use completely separate code and it'll be a bit more work for that. Definitely something for a separate PR.

PR #4487, opened 2025-08-26
CDROM Audio: Do not continue to the next track if less than 2 seconds remains

### LINKED ISSUE #4445: Alone in the Dark 2 music skipping

Windows 10

In Alone in the Dark 2, after playing the first music track of the area that you're in, instead of looping or switching to a different track, the game will play the next track for about a second, skip it, then correctly loop or play a different track.
Works correctly/can't reproduce in DOSBox-X and in DOSBox packaged with GOG release.

Noticed this appearing in the log during track switching:
Image

Just stand in one spot and wait until the track currently playing ends or play normally and you'll notice something funky with the music tracks changing.

- [x] Yes, I agree.

weirddan455:
# Description

I tested Alone in the Dark 2 with OG DOSBox 0.74-3. It has the same behavior Staging does pre #4302 (audio track cuts off early). After #4302 the audio goes on for too long, playing a couple seconds of the next track.

DOSBox-X plays the audio tracks correctly but their code has significantly diverged from ours. Search for `CDROM_Interface_Image::CDAudioCallBack` in https://github.com/joncampbell123/dosbox-x/blob/master/src/dos/cdrom_image.cpp

I would rather do this simple change than try to understand the triple nested loops they've got going on over there :laughing: 

This is kind of a hack but also a simple fix. If we have less than 5 seconds remaining, don't go on to the next track.

## Related issues

Fixes #4445

# Release notes

Fix CDROM audio tracks not playing for the correct length in Alone in the Dark 2

# Manual testing

Tested the "Shorty Leg" book in the inventory of Alone in the Dark 2

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/cdrom_image.cpp
+++ b/src/dos/cdrom_image.cpp
@@ -1044,8 +1044,23 @@ void CDROM_Interface_Image::CDAudioCallback(const int desired_track_frames)
 		// This particular CDDA track has come to an end, but the
 		// program has requested we continue playing for a longer
 		// period. So keep going!
-		player.cd->PlayNextAudioTrack();
-		return;
+		const auto frames_remaining = (player.playedTrackFrames <
+		                               player.totalTrackFrames)
+		                                    ? (player.totalTrackFrames -
+		                                       player.playedTrackFrames)
+		                                    : 0;
+		const auto num_frames_in_two_seconds = track_file->getRate() * 2;
+		if (frames_remaining < num_frames_in_two_seconds) {
+			// Less than 2 seconds remain of requested playback.
+			// Stop the audio rather than playing the next track for
+			// only a few seconds. Fixes Alone in the Dark 2:
+			// https://github.com/dosbox-staging/dosbox-staging/issues/4445
+			player.cd->StopAudio();
+			return;
+		} else {
+			player.cd->PlayNextAudioTrack();
+			return;
+		}
 	}
 
 	// Use the stereo or mono and native or nonnative AddSamples call


src/dos/cdrom_image.cpp
```diff
@@ -1044,8 +1044,17 @@ void CDROM_Interface_Image::CDAudioCallback(const int desired_track_frames)
...
 		// program has requested we continue playing for a longer
 		// period. So keep going!
-		player.cd->PlayNextAudioTrack();
-		return;
+		const auto frames_remaining = player.playedTrackFrames < player.totalTrackFrames ? player.totalTrackFrames - player.playedTrackFrames : 0;
+		if (frames_remaining < track_file->getRate() * 5) {
+			// Less than 5 seconds remain of requested playback.
+			// Stop the audio rather than playing the next track for only a few seconds.
+			// Fixes Alone in the Dark 2: https://github.com/dosbox-staging/dosbox-staging/issues/4445
+			player.cd->StopAudio();
```
johnnovak, line 1058: Couple of questions.

1. With this hack, do we get 1:1 parity with what happens in X?
2. Don't we cause any glitches by doing this hack when the track loops around?
3. First I thought this can happen when there is a discrepancy between the length of the audio track specified in the CUE sheet and the actual length of the MP3/OGG/FLAC encoded track. But this happens with a CUE/BIN image, so the length of the track in the CUE sheet and the actual length of the PCM data in the BIN file must be in perfect sync? Or is this particular CUE file buggy/invalid? I just want to understand where this discrepancy is coming from.
4. In any case, isn't 5 seconds too generous?
5. Have you tested this hack with other CD based games? We should build up a regression test suite for CD Audio games on the Wiki, otherwise we can end up play whack-a-mole with these hacks. So I'd appreciate if you'd start such a page.

johnnovak:
> I would rather do this simple change than try to understand the triple nested loops they've got going on over there 😆

Probably a good call 😅 

I'd just like to understand more about the hack/fix (see my other comment), and I really think we should build up a regression test suite for CD Audio games in the dev wiki. Initially, it could consist of the games related to the CD Audio fixes you had done over the past the year or so.

src/dos/cdrom_image.cpp
weirddan455: > With this hack, do we get 1:1 parity with what happens in X?

Don't know. Their code is hard to understand and currently has a bug with Alone in the Dark 2 where everything sounds completely garbled.

> Don't we cause any glitches by doing this hack when the track loops around?

Can't happen. The only way a DOS game can initiate CDROM audio playback is `PlayAudioSector(sector, length)`. There's no "auto-loop" feature in MSCDEX. If a game wants to loop a single track, they have to issue a 2nd call once the track has finished playing. And in that case, they would never want to go to the next track anyway.

> First I thought this can happen when there is a discrepancy between the length of the audio track specified in the CUE sheet and the actual length of the MP3/OGG/FLAC encoded track. But this happens with a CUE/BIN image, so the length of the track in the CUE sheet and the actual length of the PCM data in the BIN file must be in perfect sync? Or is this particular CUE file buggy/invalid? I just want to understand where this discrepancy is coming from.

It's dependent on the specific image. Bug only occurs on the one that's split up into multiple small .bin files. It's possible it's an inaccurate rip but I think we should compensate for it if it's easy enough.

> In any case, isn't 5 seconds too generous?

Possibly. It's arbitrary. 2 seconds seems to work as well for this game. Could lower it to 2 or 3 seconds.

> Have you tested this hack with other CD based games? We should build up a regression test suite for CD Audio games on the Wiki, otherwise we can end up play whack-a-mole with these hacks. So I'd appreciate if you'd start such a page.

These things are hard to reproduce. I have a total of 2 test cases now: Tomb Raider and Alone in the Dark 2 and both require a specific image in order to reproduce. 

The Tomb Raider bug was a result of complicated seeking logic that jumped back and forth indefinitely and hung the entire mixer thread. I've removed all that and simplified things so that should never happen again.

This should also not cause any regression for the vast majority of games that just use CDROM to play background music. This game is an outlier in that it has 60+ small tracks that it uses for voice-overs and such.

johnnovak: Thanks for the explanations @weirddan455. Approving, and probably we don't really need a CD Audio tests page, after all. 

Probably 2 seconds would suffice for the check and you could do this to make things a bit more descriptive:

```cpp
const auto num_frames_in_two_seconds = track_file->getRate() * 2;
```

johnnovak (approved):
Nice job @weirddan455 🎉

src/dos/cdrom_image.cpp
```diff
@@ -1044,8 +1044,18 @@ void CDROM_Interface_Image::CDAudioCallback(const int desired_track_frames)
 		// This particular CDDA track has come to an end, but the
 		// program has requested we continue playing for a longer
 		// period. So keep going!
-		player.cd->PlayNextAudioTrack();
-		return;
+		const auto frames_remaining = player.playedTrackFrames < player.totalTrackFrames ? player.totalTrackFrames - player.playedTrackFrames : 0;
```
johnnovak: @weirddan455 
picky: Please put parenthesis around these, then Clang reformat. Will end up looking like this or something.

```cpp
const auto frames_remaining = (player.playedTrackFrames < player.totalTrackFrames)
                                  ? (player.totalTrackFrames - player.playedTrackFrames) : 0;
```

johnnovak: Better yet:

```cpp
const auto frames_remaining = std::max(player.totalTrackFrames - player.playedTrackFrames, 0);
```

weirddan455: 2nd one won't work. It's unsigned types so it'd wrap around and get something close to u32::max.

johnnovak: Yeah, this is why we should never use unsigned unless there's a very good reason (dealing with binary protocols or emulated registers, and that's it). `int` or `int64_t` is what people need 99% of the time.

src/dos/cdrom_image.cpp
weirddan455: Changed to 2 seconds

src/dos/cdrom_image.cpp
weirddan455: Yeah, added parens

PR #4490, opened 2025-08-28
Fix video output freezing 35 minutes, 47 seconds, 483 milliseconds, and 648 microseconds after startup

johnnovak:
# Description

Just had to use this ominous title! 😎 🤘🏻 

Thanks to @FeralChild64 for spotting it to @vladkosi for additional testing 🎉 

I did consider showing a pop-up dialog after 34 minutes to buy the time-unlocked DOSBox Staging Ultimate version for the low price of $9.99, but ended up doing this fix instead 😂 

## Related issues

- Fixes https://github.com/dosbox-staging/dosbox-staging/discussions/4489#discussioncomment-14237922

# Manual testing

Started Gods on macOS let it run for a while — the animations were still playing after a few hours.

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


src/gui/sdlmain.h
```diff
@@ -271,12 +271,12 @@ struct SDL_Block {
 		PresentationMode windowed_mode =   {};
 		PresentationMode fullscreen_mode = {};
 
-		int frame_time_us           = 0;
-		int early_present_window_us = 0;
-		int last_present_time_us    = 0;
+		int frame_time_us            = 0;
+		int early_present_window_us  = 0;
+		int64_t last_present_time_us = 0;
```
johnnovak, line 276: Actual fix.

Grounded0:
😁

src/gui/sdlmain.h
FeralChild64: That's why I prefer `int64_t` / `int32_t` / whatever instead of `int`.

src/misc/video.h
```diff
@@ -139,8 +139,8 @@ struct VideoMode {
 	// 640x400 at the signal level. Similarly, all 200-line CGA and EGA
 	// modes are effectively emulated on VGA adapters and are output width
 	// and height doubled.
-	uint16_t width  = 0;
-	uint16_t height = 0;
+	int width  = 0;
```
FeralChild64, line 142: Why do we want to change this?
1. It makes it necessary to use casting.
2. It adds compiler warnings if compiling using Meson:
```
[181/323] Compiling C++ object src/hardware/libhardware.a.p/reelmagic_video_mixer.cpp.o
../src/hardware/reelmagic/video_mixer.cpp: In function ‘void setup_video_mixer(bool)’:
../src/hardware/reelmagic/video_mixer.cpp:530:34: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  530 |         if ((_vgaImageInfo.width == _mpegPictureWidth) &&
      |              ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../src/hardware/reelmagic/video_mixer.cpp:531:35: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  531 |             (_vgaImageInfo.height == _mpegPictureHeight)) {
      |              ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../src/hardware/reelmagic/video_mixer.cpp:537:41: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  537 |         } else if ((_vgaImageInfo.width == (_mpegPictureWidth * 2)) &&
      |                     ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
../src/hardware/reelmagic/video_mixer.cpp:538:42: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  538 |                    (_vgaImageInfo.height == ((_mpegPictureHeight * 2)))) {
      |                     ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/hardware/reelmagic/video_mixer.cpp:544:41: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  544 |         } else if ((_vgaImageInfo.width == _mpegPictureWidth) &&
      |                     ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../src/hardware/reelmagic/video_mixer.cpp:552:41: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  552 |         } else if ((_vgaImageInfo.width == (_mpegPictureWidth * 2)) &&
```

For some reason the CMake build does not show these warnings - we should probably review the warning flags.

weirddan455: I'm looking into the warnings. Just adding `-Wall` to the compile flags is enough to show this warning which IMO should be the bare minimum of warnings we want.

In Meson, we're setting `warning_level=3` which supposedly sets `-Wall -Wextra -Wpedantic`: https://mesonbuild.com/Builtin-options.html

CMake doesn't seem to have an equivalent so we'd just need to add the compile flags directly but `-Wpedantic` shows some extra warnings that don't appear in Meson.

I'll try to get a PR up soon to bring CMake warnings to parity.

src/gui/sdlmain.h
weirddan455: Should `frame_time_us` and `early_present_window_us` also be `int64_t`? I guess they would normally be small numbers but maybe it's better to be safe and use 64-bit whenever we're representing microseconds.

src/misc/video.h
johnnovak: > Why do we want to change this?
> It makes it necessary to use casting.

We decided a while ago to adopt the "just use `int`" style as much as possible, except for MS-DOS and BIOS internal structures that use specific widths, and for hardware registers, of course.

So anything else should be `int` or rarely `int64_t`, like in this case.

This makes life a lot easier when you use these values in calculations due to C/C++ int promotion rules. We've been all through this before, and I've made adjustments to our coding standards and referenced the Google C++ Style Guide which makes the same recommendations (just use `int`, then `int64_t` if you really need to, and specific types and unsigned types just when dealing with binary protocols, but nothing else).

kcgen was pushing the use of "right-sized" int types, but it was not a good idea in retrospect. Like I said, it causes unnecessary complications in calculations because of integer promotion, and I don't see any upsides. At the end of the day, those size restrictions are too loose (e.g., how does it help that width & height are 16-bit? Valid values are in the 200-2000 range or so; might as well just use a generic `int`).

Again, all in line with many C++ recommendations. Just use `int` 90% of the time 🤷🏻 We're dealing with small numbers 90% of the time.

I suspect `_mpegPictureWidth` could be `int` too, and in that case we wouldn't get these warnings.

@weirddan455 I'm not against improving the warnings, in general, but wouldn't adding `CHECK_NARROWING()` to `video_mixer.cpp` cause these warnings to be raised?

We did the `CHECK_NARROWING()` thing so we can convert the files over gradually, instead of turning up the warnings to eleven and living with hundreds or thousands of warnings being raise.

I don't have time now, but look into that first. If `CHECK_NARROWING()` does it, no action is needed. We should be a bit more mindful of adding `CHECK_NARROWING()` and fixing the resulting warnings whenever we touch a file going forward.

src/gui/sdlmain.h
johnnovak: They're small numbers, it's not needed. We should just use `int64_t` on a case by case basis, and adding `CHECK_NARROWING()` clearly flagged there is a narrowing problem.

We should not widen types and start using explicit integer types "just to be sure"; the solution is adopting `CHECK_NARROWING()` across the board, and when every file has it, it can be dropped and the compiler flags would be used globally.

johnnovak:
@weirddan455 @FeralChild64 Let's merge this first because it affects the dev build; it makes it basically unusable past 35 mins.

Then we can mull over the warnings situation in the coming weeks, that's not something we need to decide immediately.

src/misc/video.h
weirddan455: > @weirddan455 I'm not against improving the warnings, in general, but wouldn't adding CHECK_NARROWING() to video_mixer.cpp cause these warnings to be raised?

It does not. I just tested and `video_mixer.cpp` compiles without warnings if I add `CHECK_NARROWING()` Those are different warnings and I believe are only for assignment. This warning is for comparison of differing signs and is enabled globally in Meson. We're also not seeing unused function warnings (and there is one in `sdlmain.cpp`).

weirddan455: I would like it if there were even stricter warnings we could roll out incrementally similar to what we're doing in `CHECK_NARROWING()`. In Rust, any operation involving different types (like adding a u32 to u64) is a compile error and you must explicitly cast if that's what you want. I don't know if C++ can be quite that strict (because it does implicit casting all over the place) but it really does help prevent most of these issues.

johnnovak: Thanks for testing it. I agree with what you wrote above; stricter warnings would be better, but we need to roll them out incrementally otherwise it becomes impractical and/or too overwhelming.

PR #4492, opened 2025-08-29
Add and fix additional warnings to the CMake build

weirddan455:
# Description

In Meson, we had set `warning_level=3` which according to Meson docs sets `-Wall -Wextra -Wpedantic`. CMake does not have an equivalent that I could find so we just need to set these compiler flags directly.

`-Wpedantic` shows several warnings that are not shown in Meson so either the Meson docs are incorrect or we are supressing these warnings elsewhere. The warnings are mostly regarding 128-bit SIMD types. Example:

```
/home/daniel/code/dosbox-staging/src/libs/simde/x86/mmx.h:7679:30: warning: ISO C++ does not support ‘__int128’ for ‘simde_poly128’ [-Wpedantic]
 7679 | #  define SIMDE_POLY128_TYPE __int128
```

For now, I'm only setting `-Wall` and `-Wextra`.

Regarding the warnings in `video_mixer.cpp`, the internal `_mpegPictureWidth` and `_mpegPictureHeight` are used in bit-shifting and pointer arithmetic so it's a bit risky to change their types. I just added some `check_cast`'s to resolve the warnings.

# Manual testing

It compiled and I played Doom for a bit. Pretty low risk of regression I think.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ if(MSVC)
   # Disable cl/clangcl warnings about insecure C functions
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
-  add_compile_options("-Wno-conversion" "-Wno-narrowing")
+  add_compile_options("-Wall" "-Wextra" "-Wno-unknown-pragmas" "-Wno-conversion" "-Wno-narrowing")
 endif()
 
 option(OPT_DEBUGGER "Enable debugger" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,8 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # CHECK_NARROWING() macro to opt-in these checks on a per-file basis. See
 # `src/util/checks.h` for details.
 if(MSVC)
-  add_compile_options("/wd4244" "/MP")
+  # /W4 displays level 1, level 2, and level 3 warnings, and all level 4 (informational) warnings that aren't off by default.
+  add_compile_options("/wd4244" "/MP" "/W4")
 
   # Disable cl/clangcl warnings about insecure C functions
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,12 @@ else()
   add_compile_options("-Wall" "-Wextra" "-Wno-unknown-pragmas" "-Wno-conversion" "-Wno-narrowing")
 endif()
 
+# TODO: These are sprintf deprecated warnings.
+# We probably do want to remove sprintf usage at some point.
+if(APPLE)
+  add_compile_options("-Wno-deprecated-declarations")
+endif()
+
 option(OPT_DEBUGGER "Enable debugger" OFF)
 option(OPT_HEAVY_DEBUGGER "Enable heavy debugger" OFF)
 if(OPT_HEAVY_DEBUGGER)
--- a/src/libs/glad/CMakeLists.txt
+++ b/src/libs/glad/CMakeLists.txt
@@ -1,3 +1,7 @@
+if(APPLE)
+  add_compile_options("-Wno-cast-function-type-mismatch")
+endif()
+
 add_library(libglad STATIC
   src/gl.c
 )
--- a/src/utils/bit_view.h
+++ b/src/utils/bit_view.h
@@ -88,6 +88,27 @@ big-endian.
 #include "utils/bitops.h"
 #include "misc/support.h"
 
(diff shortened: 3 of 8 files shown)


weirddan455:
There's some additional warnings that need to be cleaned up in different compilers. I only tested with GCC 12 initially.

johnnovak:
> There's some additional warnings that need to be cleaned up in different compilers. I only tested with GCC 12 initially.

Looks good so far!

weirddan455:
Finally got all the CI checks to pass. It's ready for review now.

CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # CHECK_NARROWING() macro to opt-in these checks on a per-file basis. See
 # `src/util/checks.h` for details.
 if(MSVC)
-  add_compile_options("/wd4244" "/MP")
+  add_compile_options("/wd4244" "/MP" "/W4")
```
johnnovak, line 25: Maybe add a succinct version of the meaning of W4 from here in a comment?

https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170

CMakeLists.txt
```diff
@@ -29,6 +29,12 @@ else()
   add_compile_options("-Wall" "-Wextra" "-Wno-unknown-pragmas" "-Wno-conversion" "-Wno-narrowing")
 endif()
 
+# TODO: These are sprintf deprecated warnings.
+# We probably do want to remove sprintf usage at some point.
```
johnnovak, line 34: Great comment 💯

src/libs/glad/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if(APPLE)
+  add_compile_options("-Wno-cast-function-type-mismatch")
```
johnnovak, line 2: Maybe we could consider submitting a GLAD patch upstream to get rid of the warning at the source level, but that would be more involved.

src/utils/bit_view.h
```diff
@@ -88,6 +88,27 @@ big-endian.
...
+// Clang does not have -Wmaybe-uninitialized so it needs -Wunknown-warning-option
+// to supress a warning about unknown warnings.
+
+// We can't just remove the pragma clang stuff because from testing,
+// Clang accepts pragma GCC but GCC does not accept pragma clang.
+
+// Also, GCC does not have -Wunknown-warning-option,
+// so if we *only* have pragma GCC, we get warned about
+// an unkown warning that's supposed to supress unkown warnings.
+// Sheesh...
```
johnnovak, line 100: 😅

tests/string_utils_tests.cpp
```diff
@@ -183,12 +183,6 @@ TEST(SafeStrlen, EmptyString)
 	EXPECT_EQ(0, safe_strlen(buffer));
 }
 
-TEST(SafeStrlen, Uninitialized)
-{
-	char buffer[4];
-	EXPECT_GT(sizeof(buffer), safe_strlen(buffer));
-}
```
johnnovak, line 190: I agree with the removal.

Typo in the commit message: inlining

johnnovak (approved):
Great work @weirddan455 

One more thing before merging: could you scale back the max allowed macOS warnings from 9 to 0?

src/libs/glad/CMakeLists.txt
weirddan455: GLAD is generated code. If you look at their GitHub it's a bunch of Python files that generate the C++ code: https://github.com/Dav1dde/glad

They also don't ship any build system components so this CMake file is all ours. I'm not going to fuss with this upstream personally especially because the warning only shows on Mac so I can't test locally.

CMakeLists.txt
weirddan455: Copied the first sentence from that link as a comment :smile:

tests/string_utils_tests.cpp
weirddan455: Fixed commit typo.

Also just realized this test is even worse than I first thought. It fails if there happens to be a zero byte in that uninitialized stack memory :sweat_smile:

src/libs/glad/CMakeLists.txt
johnnovak: Fair enough. Could be fixed at the generator level, that's what I'm saying, but I understand if you don't feel the in inclination 😄 (I did submit patches in the past, and I also submitted a Nim codegen backend)

tests/string_utils_tests.cpp
weirddan455: Or wait... no it doesn't. It's just expecting the `strlen` to be less than the array size. Still not a very useful test. I'm just now having my coffee :coffee:

PR #4451, opened 2025-08-08
Major project organisation improvements

johnnovak:
# Description

The approach of handling include files we inherited from the original DOSBox project [has been annoying me for a while now](https://github.com/dosbox-staging/dosbox-staging/issues/2757). The `include` folder contains a mishmash of header-only utilities, public APIs, and then some not-so-public-APIs. Because it's basically one big bag of "things", there is no namespacing. What will you exactly get when you include "mem.h", "rgb.h", or "regs.h"? It's anyone's guess. Sure, you can get used to it, but that can be said for any less than ideal thing...

...except I couldn't get used to it, so I took it upon myself to finally rectify the situation 😎 

I've done some research on what's the current best-practice of structuring C++ projects, and many people seem to prefer the following project layout:

- Only have an `include` dir separate from the `src` dir for libraries, in which case `include` contains the public API. Clearly, we're not a library.
- For applications like DOSBox Staging, a popular solution is to only have `src` and put the header files next to the implementation files.
- The most common way for referencing include files in such a layout is to simply add  `src` to the include search path and then `#include` the full path under `src`. Confusing and clunky parent-relative paths are never used (e.g., rewrite stuff like `../../audio/mixer.h` to `audio/mixer.h`).
- If the directory hierarchy is logically structured, this will result in "natural namespacing" of the includes (e.g., `hardware/audio/sblaster.h`, `dos/memory.h`, or `util/rgb.h`).

I've spent a whole day doing this for the _whole source tree_ 😆 First I thought I'd do it only partially then farm it out to you guys, but then went "screw it, let's get the whole thing done". It was strangely therapeutical, I quite enjoyed it 😎🤘🏻 Yeah, I'm the type who likes grinding in RPGs 😂 ⚔️ 🧙🏻 (I used some automation, of course, before you ask...)

The results speak for themselves! `mixer.cpp` before/after:

image

But what about private headers? This approach can make it hard to differentiate between "module public" and "module private" headers.

That's a valid point, and a common solution people use is to put module private headers into a `private` or `internal` subdir inside the module directory. I've chosen `private` and I've converted the `src/hardware/input` module to this format as demonstration:

image

We don't have to do this for every small module, but I think it clarifies the intent for bigger modules and results in better overall design. I encourage people to do this conversion when they work on some module in the future.

---

I've also done a few further improvements:

- Moved audio stuff that doesn't emulate actual physical hardware (e.g., mixer, compressor, noise gate, etc.) into a new module in `src/audio` 
[...]

PR #4513, opened 2025-09-17
Organise files (part 2)

johnnovak:
# Description

Follow up to https://github.com/dosbox-staging/dosbox-staging/pull/4451. I ran out of steam there and missed a few improvement opportunities—I'm doing those now. Some files ended up in the wrong place, so that's now corrected as well.

Other than that, I've broken apart `video.h` which was a bit of a kitchen sink and have moved its contents to more appropriate places. I've also made most of the `gui` definitions private.

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/pull/4451

# Manual testing

Stuff compiles, Staging starts up.

- [ ] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:


FeralChild64:
> Move virtualbox.cpp|h & vmware.cpp|h to src/hardware/input

This is not a good location for those files. Currently they indeed only serve the purpose of emulating the VirtualBox/VMWare mouse interfaces, but their intended role is more generic:
- they tap into I/O ports, PCI subsystem, etc.
- they implement interface identification, handshaking with guest side drivers, interface version reporting, they inform about hypervisor capabilities, etc.
- they decode functionality-specific guest side requests and pass them to appropriate subsystem; currently this is mouse emulation only, but it can be anything else in the future - host directory sharing (should be useful for Windows 9x support), clipboard sharing, maybe even a virtual GPU emulation

IMHO they should either stay where they are or go to a new dedicated directory, like `src/hardware/virtual_machines`.

The “input“ (or rather the mouse part) is implemented in the `mouseif_virtual_machines.cpp` file, which is already placed in the `src/hardware/input` directory.

johnnovak:
> IMHO they should either stay where they are or go to a new dedicated directory, like `src/hardware/virtual_machines`.

Sure, I'll move them back to `hardware`. Thanks for the input, wasn't sure if they really belong to `hardware/input`.

On a second thought, not too sure if `hardware` is the best place for them as in my view only emulations of actual pieces of hardware should be there, but these are a bit special so it's ok for now.

FeralChild64:
Does not link anymore if compiling using meson on Linux:

```
c++  -o dosbox dosbox.p/src_main.cpp.o dosbox.p/src_dosbox.cpp.o dosbox.p/src_libs_ghc_fs_std_impl.cpp.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -Wl,-Map,dosbox.map -Wl,--Map,dosbox.map -Wl,--gc-sections -rdynamic -Wl,--start-group src/misc/libmisc.a src/libs/loguru/libloguru.a src/libs/whereami/libwhereami.a src/audio/libaudio.a src/capture/libcapture.a src/libs/zmbv/libzmbv.a src/config/libconfig.a src/cpu/libcpu.a src/dos/libdos.a src/libs/decoders/libdecoders.a src/fpu/libfpu.a src/gui/libgui.a src/libs/glad/libglad.a src/hardware/libhardware.a src/libs/ESFMu/libesfmu.a src/libs/nuked/libnuked.a src/libs/residfp/libresidfp.a src/libs/YM7128B_emu/libym7128bemu.a src/ints/libints.a src/midi/libmidi.a src/shell/libshell.a src/utils/libutils.a src/libs/tal-chorus/libtalchorus.a src/libs/manymouse/libmanymouse.a -latomic -lstdc++fs /usr/lib64/libSDL2.so -pthread /usr/lib64/libiir.so /usr/lib64/libslirp.so /usr/lib64/libglib-2.0.so -lstdc++fs -ldl /usr/lib64/libspeexdsp.so /usr/lib64/libpng16.so /usr/lib64/libz-ng.so /usr/lib64/libSDL2_net.so /usr/lib64/libopusfile.so /usr/lib64/libGL.so /usr/lib64/libasound.so /usr/lib64/libmt32emu.so /usr/lib64/libXi.so -Wl,--end-group
/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: src/hardware/libhardware.a.p/network_ne2000.cpp.o: in function `NE2K::NE2K(Section*)':
ne2000.cpp:(.text._ZN4NE2KC2EP7Section[_ZN4NE2KC5EP7Section]+0x5c3): undefined reference to `ETHERNET_OpenConnection(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
collect2: error: ld returned 1 exit status
```

CMake build seems to be functioning correctly.

[edit] Maybe it's time to drop the Meson build system?

johnnovak:
> Does not link anymore if compiling using meson on Linux:

Fixing now in my Ubuntu VM.

> [edit] Maybe it's time to drop the Meson build system?

I wanted to drop it 2-3 months ago, but @weirddan455 argued that it would be useful to keep it for a while longer so we can compare things between the Meson and the CMake builds if we run into some strange bugs.

I think that's a good idea, so dropping Meson after 0.83.0 final is out is probably the best time.

johnnovak:
@FeralChild64 Now all commits are building for me with Meson on my Ubuntu VM.

weirddan455:
> I wanted to drop it 2-3 months ago, but @weirddan455 argued that it would be useful to keep it for a while longer so we can compared things between the Meson and the CMake builds if we run into some strange bugs.
>
> I think that's a good idea, so dropping Meson after 0.83.0 final is out is probably the best time.

IIRC at the time we didn't have ManyMouse and networking libraries working in CMake. Now we're at parity with Meson or at least very close to it.

Waiting until 0.83.0 is probably the safest option but we're closing to being able to kill it with fire now :fire:

johnnovak:
> Waiting until 0.83.0 is probably the safest option but we're closing to being able to kill it with fire now 🔥

Yeah, I'm intentionally not doing anything OSD related until 0.83.0 final is out because I'm intending to use [Blend2D](https://blend2d.com/) for the rendering that doesn't have a Meson port.

FeralChild64 (approved):
OK, compiles with Meson now.

PR #4516, opened 2025-09-19
Improve handling config setting updates & module re-initialisations

johnnovak:
# Description

## Background & motivation

Handling config setting updates at runtime has been a long-standing pain-point of Staging development. The problem with the current implementation is that when a setting is changed after startup (via the `config` command),  the config system destroys then reinitialises the entire config section the setting belongs to. Although this kinda sorta works, it's often a very heavy handed approach. Finer-grained setting change handling is often a much more desirable solution. 

For example, if you change any FluidSynth parameter like the chorus or reverb, currently the entire FluidSynth instance will get destroyed the reinitialised, and the (potentially large) SoundFont will then be reloaded from disk. With fine-grained control, we could only set the new chorus or reverb parameters on the running FluidSynth instance without reinitialising it and reloading the SoundFont.

Fine-grained changes that do just what's necessary and not more in response setting changes are a must for the upcoming OSD work (e.g., reinitialising the entire SoundBlaster or GUS module while a game is running is not an option when we're only changing the SB or GUS output low-pass filter settings from the OSD/GUI).

## Details of the current implementation

The current config system allows registering an init and shutdown handler for each config section. The init handler runs at startup, and the shutdown handler when exiting Staging. In the simplest case, each section's shutdown handler is executed and then the module is reinitalised via the init handler whenever a setting from the section is changed. Even in this simple case, detecting which config section has been actually changed involves storing the previous state of all settings from the config section manually and possibly introducing an `is_first_init_invocation` static flag as well. This quickly becomes rather messy but it's doable (I've done such things myself in a couple of places, like in the render section to make the auto-CRT shaders work more smoothly without reiniting the entire rendering system on shader changes).

In more complicated cases, however, this approach becomes ultra messy. The problem is that both the init and shutdown handlers take an extra flag called `changeable_at_runtime` which controls whether the handler should be executed only at application startup and exit, or in response to runtime setting changes, too. Then some modules control the lifecycle of dependent child modules as well(e.g., `[cpu]` and `[dos]`), and the `changeable_at_runtime` flag can (and does) vary per sub-module and handler as well, which further increases complexity and the cognitive load.

In such cases, it's super hard or even virtually impossible to come up with a manually-maintained fine-grained config setting change handling mechanism that does all the more granular reinit tasks correctly across all child modules. The "dynamic nature" of the config system (it uses callbacks for the init and shutdown handlers) actively works against us in such scenarios; often we can't just rely on the static behaviour imposed by the config system, but we need to do conditional reinits programmatically instead (depending on which setting has been changed). Currently, this is only possible with a large number of special state variables and flags which is a nightmare to write, test, debug, reason about and maitain. The worst of all, this mechanism would need to be reimplemented in every single module manually.

But wait, there's more! Init handlers are "forever" once registered, but destroy handlers are "consumed" on execution and need to be re-registered if you want them to fire again the next time. This hack/kludge was introduced to support setting changes at runtime to some extent (it's a bit complicated to explain why that's so, but this was clear once I understood the current maze of init/destroy handlers a bit better). All in all, the original devs did the "easy" thing vs the "correct" thing (a more suitable solution for the task at hand), and this invariable results in unmaintable and messy code.

Lastly, some of the init handlers are registered in seemingly unrelated modules just to force a specific init order (due to inter-module dependencies). This adds another layer of complexity and confusion to the whole init & shutdown process, and hides dependencies because these handler registrations are spread over the whole codebase.

## New implementation

In the new implementation, the init and shutdown (destroy) handlers have been _completely_ eradicated. They're gone 🎉  Modules are now explicitly calling the `MODULE_Init()` and `MODULE_Destroy()` functions of their child modules explicitly in their own init and destroy functions. This enables fine-grained programmatic control of partial child module re-initialisations on config setting changes, and makes the whole init/destroy process a lot more explicit and clearly laid out in front of you in one place.

There's a top-level `DOSBOX_InitModules()` / `DOSBOX_DestroyModules()` pair of functions, too, to init and destruct the entire dependency tree at startup and exit (it's called from `sdlmain.cpp`).

To handle setting updates at runtime, each config section can register a setting update callback via `AddUpdateHandler()`. This callback gets executed when a setting is changed at runtime (e.g., via the config command, or from the future OSD). A reference to the section and the setting's name are passed in as arguments to the callback invocation, allowing precise fine-grained control over setting changes.

(As a sidenote, choosing the _least powerful_ technique for solving a particular programming problem is almost invariably the most maintainable one. In our case, our modules are known at compile time, so the dynamic config system that uses callbacks doesn't add any value just complications and inflexibility. As there is no dynamic behaviour involved, the init and destroy processes are better handled by a _less powerful_ and less complicated technique, which is good old explicit function calls. On the other hand, introducing the setting update callback mechanism is appropriate because setting updates can happen at any random time—this must be a runtime mechanism.)

## Improved setting update handlers

I've improved the setting update handling in the following files (made it more granular not requiring a full reinit for at least some of the settings). It's a work in progress, there's room for further improvement in most of these update handlers. Check out the `notify_*_setting_updated` function for details.

- dosbox.cpp
- fluidsynth.cpp
- gus.cpp
- mixer.cpp
- mouse_config.cpp
- render.cpp

# Manual testing

The good news is that although the PR is rather large, the changes only involve module initialisation and destruction, not the actual workings of the modules. Generally, if it starts up and log something, the change was successful! This makes testing things relatively easy because the following can happen if the init/reinit process goes wrong:

- Crash, hang, or tripping some debug assert at startup
- Crash, hang, or tripping some debug assert at exit
- Crash, hang, or tripping some debug assert when changing settings at runtime
- In rare cases, changing a setting might seems fine, but it becomes a no-op if things went wrong in the refactored code, or Staging might subtly misbehave after the setting change in the absolute worst scenario (very unlikely, though).

I've tested a large number of games that exercise various features (ReelMagic, MT-32, Sound Canvas, various machine types, Voodoo, etc.) and I've tested changing all config settings at runtime after I've tackled a given config section or module. As far I can tell, everything works just like before.

I've also done some minor fixes to a few existing config setting change related problems—see the commit messages for details.

This is pretty hard to review meaningfully, so if you guys could do some external testing in the areas you're familiar with, that would be appreciated. For example, I'm not too familiar with the networking and serial port stuff, neither the localisation and the debugger, so extra testing in those areas would be welcome (I've done the basics, though).

- [x] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:


src/config/setup.h
```diff
@@ -331,27 +332,16 @@ class PropHex final : public Property {
 
 #define NO_SUCH_PROPERTY "PROP_NOT_EXIST"
 
-typedef void (*SectionFunction)(Section*);
+using SectionUpdateHandler =
+        std::function<void(SectionProp*, const std::string& prop_name)>;
```
FeralChild64: How about changing the API to use a reference (`SectionProp&`) instead of a pointer?

johnnovak: I think I can do that easily in a new commit. Good idea.

johnnovak: Okay, I've done it, but it took like an hour 😅 Not hard, but really time consuming... Thanks for the homework! 😂

src/cpu/cpu.cpp
```diff
@@ -3198,10 +3204,26 @@ class Cpu final {
...
+static std::unique_ptr<Cpu> cpu_instance = {};
+
+void CPU_Init()
+{
+	auto section = get_section("cpu");
+
+	cpu_instance = std::make_unique<Cpu>(section);
+
+	DMA_Init();
+	VGA_Init();
```
FeralChild64: Is there any good reason why DMA, VGA, KEYBOARD and PCI lifecycle is handled inside the CPU module?

johnnovak: They get reinited when the CPU module gets reinited. I kept all the non-trivial dependencies like this one; I only moved module inits to the top level when it was trivial to reason about it.

Discovering why the lifecycles of those modules are tied to the CPU module is out of scope for this PR.

FeralChild64:
Tested on Linux:
- `config -axadd foobar` & `config -wcd` - works, new `[autoexec]` line is added
- changing titlebar at runtime - works
- `config -set mouse_capture=(...)` - works
- using `serial` command to change device connected to COM1 (mouse, dummy), or to change the emulated serial mouse parameters - works
- `config -set serial1=(...)` - changing to dummy or mouse works as expected 

Unfortunately, DOSBox Staging now has problems starting some games, for example:
- Warcraft (CD version) - black screen
- Settlers II Gold Edition (CD version) - black screen
- Grand Theft Auto - black screen
Also screen any emulated mode change now produces a lot more "flickering".

Something is definitely broken regarding screen mode changes. This is how Warcraft starts now (I'm using custom BAT scripts to adjust cycles, emulated sound hardware, etc.) resulting in black screen only - it re-initializes the S3 card emulation during the startup (see the `VIDEO:` logs - this is definitely wrong!):

```
2025-10-01 18:06:29.232 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:06:29.304 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:06:29.403 | MAPPER: Joystick subsystem disabled
2025-10-01 18:06:29.404 | MAPPER: Loaded default key bindings
2025-10-01 18:06:29.404 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:06:29.434 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:06:29.487 | DMA: Shutting down primary controller
2025-10-01 18:06:29.487 | DMA: Shutting down SB16 on 8-bit DMA channel 1
2025-10-01 18:06:29.487 | OPL: Shutting down OPL3
2025-10-01 18:06:29.488 | SB16: Shutting down
2025-10-01 18:06:29.488 | DMA: Shutting down GUS on 8-bit DMA channel 3
2025-10-01 18:06:29.488 | GUS: Shutting down
2025-10-01 18:06:29.488 | DMA: Shutting down secondary controller
2025-10-01 18:06:29.488 | VIDEO: Initialised S3 Trio64 VESA 2.0 compatible with 4 MB of FP DRAM supporting 89 modes
2025-10-01 18:06:29.496 | VIDEO: Initialised S3 Trio64 VESA 2.0 compatible with 4 MB of FP DRAM supporting 89 modes
2025-10-01 18:06:29.501 | OPL: Operating at 49716 Hz and downsampling to the output rate
2025-10-01 18:06:29.502 | OPL: Noise gate enabled
2025-10-01 18:06:29.502 | OPL: Running OPL3 on ports 220h and 388h
2025-10-01 18:06:29.502 | OPL: Sound Blaster 16 OPL output filter enabled
2025-10-01 18:06:29.502 | DMA: Initialised primary controller
2025-10-01 18:06:29.502 | DMA: Initialised secondary controller
2025-10-01 18:06:29.502 | SB: Operating at 22050 Hz and upsampling to the output rate
2025-10-01 18:06:29.502 | SB16: Sound Blaster 16 DAC output filter enabled
2025-10-01 18:06:29.502 | SB16: Setting 'BLASTER' environment variable to 'A220 I7 D1 H5 T6'
2025-10-01 18:06:29.502 | SB16: Running on port 220h, IRQ 7, DMA 1, and high DMA 5
2025-10-01 18:06:29.502 | OPL: Shutting down OPL3
2025-10-01 18:06:29.502 | SB16: Shutting down
2025-10-01 18:06:29.503 | OPL: Operating at 49716 Hz and downsampling to the output rate
2025-10-01 18:06:29.503 | OPL: Noise gate enabled
2025-10-01 18:06:29.503 | OPL: Running OPL3 on ports 220h and 388h
2025-10-01 18:06:29.503 | OPL: Sound Blaster 16 OPL output filter enabled
2025-10-01 18:06:29.503 | SB: Operating at 22050 Hz and upsampling to the output rate
2025-10-01 18:06:29.503 | SB16: Sound Blaster 16 DAC output filter enabled
2025-10-01 18:06:29.503 | SB16: Setting 'BLASTER' environment variable to 'A220 I7 D1 H5 T6'
2025-10-01 18:06:29.503 | SB16: Running on port 220h, IRQ 7, DMA 1, and high DMA 5
2025-10-01 18:06:29.504 | MPU-401: Shutting down
2025-10-01 18:06:29.504 | MPU-401: Running in intelligent mode on ports 330h and 331h
2025-10-01 18:06:29.504 | FSYNTH: Shutting down
2025-10-01 18:06:29.509 | FSYNTH: Successfully loaded FluidSynth 2.4.3
2025-10-01 18:06:29.606 | FSYNTH: Using SoundFont '/usr/share/sounds/sf2/default.sf2'
2025-10-01 18:06:29.606 | FSYNTH: Chorus enabled with 3 voices at level 1.20, 0.30 Hz speed, 8.00 depth, and sine-wave modulation
2025-10-01 18:06:29.606 | FSYNTH: Reverb enabled with a 0.61 room size, 0.23 damping, 0.76 width, and level 0.56
2025-10-01 18:06:29.607 | FSYNTH: Operating at 48000 Hz without resampling
2025-10-01 18:06:29.608 | MIDI: Opened device 'fluidsynth'
2025-10-01 18:06:29.608 | PCSPEAKER: Shutting down discrete model
2025-10-01 18:06:29.608 | PCSPEAKER: Operating at 48000 Hz without resampling
2025-10-01 18:06:29.608 | PCSPEAKER: Initialised discrete model
2025-10-01 18:06:29.608 | PCSPEAKER: High-pass filter enabled (18 dB/oct at 120 Hz)
2025-10-01 18:06:29.608 | PCSPEAKER: Low-pass filter enabled (12 dB/oct at 4800 Hz)
2025-10-01 18:06:29.610 | IMGMOUNT: Path '..\-VIRTCDS\WARCRAFT\W1\CD01.CUE' not found, maybe it's a DOS path
2025-10-01 18:06:29.610 | IMGMOUNT: Path 'GAMES\-VIRTCDS\WARCRAFT\W1\CD01.CUE' found on virtual drive C:
2025-10-01 18:06:29.610 | CDAUDIO: Operating at 48000 Hz without resampling
2025-10-01 18:06:29.610 | CDROM: Loaded CD01_T02.FLA [44100 Hz, 2-channel, 0.8 minutes]
2025-10-01 18:06:29.611 | CDROM: Loaded CD01_T03.FLA [44100 Hz, 2-channel, 4.2 minutes]
2025-10-01 18:06:29.611 | CDROM: Loaded CD01_T04.FLA [44100 Hz, 2-channel, 3.8 minutes]
2025-10-01 18:06:29.611 | CDROM: Loaded CD01_T05.FLA [44100 Hz, 2-channel, 3.7 minutes]
2025-10-01 18:06:29.611 | CDROM: Loaded CD01_T06.FLA [44100 Hz, 2-channel, 3.8 minutes]
2025-10-01 18:06:29.611 | CDROM: Loaded CD01_T07.FLA [44100 Hz, 2-channel, 3.6 minutes]
2025-10-01 18:06:29.611 | CDROM: Loaded CD01_T08.FLA [44100 Hz, 2-channel, 4.0 minutes]
2025-10-01 18:06:29.620 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:06:29.673 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:06:29.710 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:06:29.757 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:06:29.807 | DMA: Shutting down primary controller
2025-10-01 18:06:29.807 | DMA: Shutting down SB16 on 8-bit DMA channel 1
2025-10-01 18:06:29.807 | OPL: Shutting down OPL3
2025-10-01 18:06:29.807 | SB16: Shutting down
2025-10-01 18:06:29.807 | DMA: Shutting down secondary controller
2025-10-01 18:06:29.807 | VIDEO: Initialised S3 Trio64 VESA 2.0 compatible with 4 MB of FP DRAM supporting 89 modes
```

And this is how it looks like on the main branch (here the game starts successfully):

```
2025-10-01 18:09:27.503 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:09:27.508 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:09:27.508 | MAPPER: Joystick subsystem disabled
2025-10-01 18:09:27.508 | MAPPER: Loaded default key bindings
2025-10-01 18:09:27.509 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:09:27.546 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:09:27.627 | OPL: Shutting down OPL3
2025-10-01 18:09:27.627 | SB16: Shutting down
2025-10-01 18:09:27.627 | OPL: Operating at 49716 Hz and downsampling to the output rate
2025-10-01 18:09:27.627 | OPL: Noise gate enabled
2025-10-01 18:09:27.627 | OPL: Running OPL3 on ports 220h and 388h
2025-10-01 18:09:27.627 | OPL: Sound Blaster 16 OPL output filter enabled
2025-10-01 18:09:27.628 | SB: Operating at 22050 Hz and upsampling to the output rate
2025-10-01 18:09:27.628 | SB16: Sound Blaster 16 DAC output filter enabled
2025-10-01 18:09:27.628 | SB16: Setting 'BLASTER' environment variable to 'A220 I7 D1 H5 T6'
2025-10-01 18:09:27.628 | SB16: Running on port 220h, IRQ 7, DMA 1, and high DMA 5
2025-10-01 18:09:27.628 | OPL: Shutting down OPL3
2025-10-01 18:09:27.628 | SB16: Shutting down
2025-10-01 18:09:27.628 | OPL: Operating at 49716 Hz and downsampling to the output rate
2025-10-01 18:09:27.628 | OPL: Noise gate enabled
2025-10-01 18:09:27.628 | OPL: Running OPL3 on ports 220h and 388h
2025-10-01 18:09:27.628 | OPL: Sound Blaster 16 OPL output filter enabled
2025-10-01 18:09:27.629 | SB: Operating at 22050 Hz and upsampling to the output rate
2025-10-01 18:09:27.629 | SB16: Sound Blaster 16 DAC output filter enabled
2025-10-01 18:09:27.629 | SB16: Setting 'BLASTER' environment variable to 'A220 I7 D1 H5 T6'
2025-10-01 18:09:27.629 | SB16: Running on port 220h, IRQ 7, DMA 1, and high DMA 5
2025-10-01 18:09:27.629 | MPU-401: Shutting down
2025-10-01 18:09:27.629 | FSYNTH: Shutting down
2025-10-01 18:09:27.635 | MPU-401: Running in intelligent mode on ports 330h and 331h
2025-10-01 18:09:27.635 | FSYNTH: Successfully loaded FluidSynth 2.4.3
2025-10-01 18:09:27.734 | FSYNTH: Using SoundFont '/usr/share/sounds/sf2/default.sf2'
2025-10-01 18:09:27.734 | FSYNTH: Chorus enabled with 3 voices at level 1.20, 0.30 Hz speed, 8.00 depth, and sine-wave modulation
2025-10-01 18:09:27.736 | FSYNTH: Operating at 48000 Hz without resampling
2025-10-01 18:09:27.736 | MIDI: Opened device 'fluidsynth'
2025-10-01 18:09:27.737 | PCSPEAKER: Shutting down discrete model
2025-10-01 18:09:27.737 | PCSPEAKER: Operating at 48000 Hz without resampling
2025-10-01 18:09:27.737 | PCSPEAKER: Initialised discrete model
2025-10-01 18:09:27.737 | PCSPEAKER: High-pass filter enabled (18 dB/oct at 120 Hz)
2025-10-01 18:09:27.737 | PCSPEAKER: Low-pass filter enabled (12 dB/oct at 4800 Hz)
2025-10-01 18:09:27.739 | IMGMOUNT: Path '..\-VIRTCDS\WARCRAFT\W1\CD01.CUE' not found, maybe it's a DOS path
2025-10-01 18:09:27.739 | IMGMOUNT: Path 'GAMES\-VIRTCDS\WARCRAFT\W1\CD01.CUE' found on virtual drive C:
2025-10-01 18:09:27.739 | CDAUDIO: Operating at 48000 Hz without resampling
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T02.FLA [44100 Hz, 2-channel, 0.8 minutes]
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T03.FLA [44100 Hz, 2-channel, 4.2 minutes]
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T04.FLA [44100 Hz, 2-channel, 3.8 minutes]
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T05.FLA [44100 Hz, 2-channel, 3.7 minutes]
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T06.FLA [44100 Hz, 2-channel, 3.8 minutes]
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T07.FLA [44100 Hz, 2-channel, 3.6 minutes]
2025-10-01 18:09:27.739 | CDROM: Loaded CD01_T08.FLA [44100 Hz, 2-channel, 4.0 minutes]
2025-10-01 18:09:27.740 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:09:27.749 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:09:27.750 | DISPLAY: Using 2400x1800 window size in windowed mode on display-0
2025-10-01 18:09:27.781 | OPENGL: Version: 4.6, GLSL version: 4.60, vendor: Intel
2025-10-01 18:09:29.501 | GUS: Activated 32 voices at 19293 Hz
2025-10-01 18:09:29.502 | GUS: Activated 28 voices at 22050 Hz
2025-10-01 18:09:29.675 | DISPLAY: VGA 320x200 256-colour graphics mode 13h at 70.086 Hz, scaled to 2400x1800 pixels with 1:1.2 (5:6) pixel aspect ratio
2025-10-01 18:09:29.675 | DISPLAY: Presenting at DOS rate without vsync
```

FeralChild64:
Another problem (not sure if the same or unrelated) - I'm running a dynamic CPU core and when executing `config -set core=normal` DOSBox Staging crashes now:

```
2025-10-01 19:54:42.169 | VIDEO: Initialised S3 Trio64 VESA 2.0 compatible with 4 MB of FP DRAM supporting 89 modes

Stack trace:
17      0x55b3242488d5 _start + 37
16      0x7f13fb4504fb __libc_start_main + 139
15      0x7f13fb45043b /usr/lib64/libc.so.6(+0x2743b) [0x7f13fb45043b]
14      0x55b324414989 sdl_main(int, char**) + 2809
13      0x55b32461d806 SHELL_InitAndRun() + 30182
12      0x55b324615d1a DOS_Shell::Run() + 1834
11      0x55b324624c7d DOS_Shell::ShowPrompt() + 301
10      0x55b32458c0b6 CONSOLE_Write(std::basic_string_view<char, std::char_traits<char> >) + 86
9       0x55b324255f20 DOS_WriteFile(unsigned short, unsigned char*, unsigned short*, bool) + 160
8       0x55b324307fa6 device_CON::Write(unsigned char*, unsigned short*) + 102
7       0x55b324546d20 INT10_TeletypeOutputAttrViaInterrupt(unsigned char, unsigned char, bool) + 528
6       0x55b3242e7816 CALLBACK_RunRealInt(unsigned char) + 70
5       0x55b32424915e DOSBOX_RunMachine() + 14
4       0x55b324248af3 ./build/release-linux/dosbox(+0x1bbaf3) [0x55b324248af3]
3       0x55b3245438ae ./build/release-linux/dosbox(+0x4b68ae) [0x55b3245438ae]
2       0x55b324547391 void write_char<(CallPlacement)0>(unsigned char, unsigned char, unsigned char, unsigned short, bool) + 289
1       0x55b32454666f WriteChar(unsigned short, unsigned short, unsigned char, unsigned char, unsigned char, bool) + 719
0       0x55b32452d09f mem_writeb(unsigned int, unsigned char) + 31
2025-10-01 19:54:42.174 | Signal: SIGSEGV

```

johnnovak:
@FeralChild64 Thanks for the testing, that's great info and exactly the type of feedback I was looking for 🎉 Looking into it. I'll ask you to send me repro cases in ZIP files if I can't trigger the problems.

> This is how Warcraft starts now (I'm using custom BAT scripts to adjust cycles, emulated sound hardware, etc.) resulting in black screen only - it re-initializes the S3 card emulation during the startup (see the VIDEO: logs - this is definitely wrong!):

I know you do these batch files that reconfigure the machine at runtime (I don't), so these type of batch files are very good test cases.

Could you attach the following in a comment in a ZIP:

- Your primary config
- Any local config
- All batch files you use to launch the games

Cheers 😄

johnnovak:
@FeralChild64 Okay, I've found the root cause: it was the DMA, keyboard, VGA, and PCI bus modules being reinitialised whenever any `[cpu]` section setting was changed. You asked about that stuff in a comment, so you have a good eye for detail! 😆 

It turns out those modules were tied to the CPU module to ensure the correct init/shutdown order (they must be initialised _after_ the CPU module, and the reverse on shutdown). But they were not re-initialised on `[cpu]` setting changes after startup, so that was my mistake. I've fixed that up; now you can change `core` and similar CPU settings just fine without any hangs or crashes 🥳 

Please re-test the problematic games using your elaborate batch file setup and let me know how it goes. In theory, this should fix all your issues.

FeralChild64:
Looks like all the problems I have reported earlier got fixed, but now I can see more problems:

1. There is now much more flicker when switching screen modes, for example when starting the game - this is very visible in fullscreen mode, but I think this happens in windowed mode too.
2. Set `mouse_capture = seamless` in the config file, switch DOSBox Staging to fullscreen, start Windows 3.1 using the following BAT file (of course you'll need to adapt the paths):

```bat
@ECHO OFF
C:
CD C:\WIN31PL
Z:\CONFIG.COM -set "window_titlebar= program='Microsoft Windows 3.11 (Polish)' cycles=on"
WIN.COM
```

and, after Windows starts, both the host OS and Windows 3.11 mouse cursors are visible (this is bad - only Windows 3.11 cursor should be visible). The `Z:\CONFIG.COM -set "window_titlebar=(...)` command is crucial to reproduce the problem.

FeralChild64:
BTW - I am starting each and every game using a BAT file containing the command like:

```BAT
Z:\CONFIG.COM -set "window_titlebar= program='FooBarGame' cycles=on"
```

just before staring the game executable - this makes the difference in flickering much more visible.

FeralChild64:
One remark - my intention was that the command to change the window titlebar should not cause screen flickering AT ALL. I did not fully succeed - on 0.82.2 there is no flickering if the new title differs from the old one (which is, fortunately, the most common case), but it does flicker if you set the same `window_titlebar` parameter value twice in a row (sadly - old config system limitation).

On the current `main` it flickers every time, no exception. Using code from this PR, the flickering is even more visible. Something goes really wrong here - we are regressing.

johnnovak:
> Looks like all the problems I have reported earlier got fixed

Awesome, that's great to hear! 🥳 🎉 

> my intention was that the command to change the window titlebar should not cause screen flickering AT ALL. I did not fully succeed - on 0.82.2 there is no flickering if the new title differs from the old one (which is, fortunately, the most common case), but it does flicker if you set the same `window_titlebar` parameter value twice in a row (sadly - old config system limitation).
> 
> On the current `main` it flickers every time, no exception. Using code from this PR, the flickering is even more visible. Something goes really wrong here - we are regressing.

I could repro this, and it was a very simple fix, thanks to the new config update callback mechanism: https://github.com/dosbox-staging/dosbox-staging/pull/4516/commits/9fe825e62e76b1748d396c97ef42af12ee1979b5

With the fix in place, I'm happy to report I'm getting zero flicker whenever changing `window_titlebar` in either windowed or fullscreen mode on both Windows and macOS. Hopefully, you'll get the same results on your Linux setup (we're only calling `SDL_SetWindowTitle()` now when the titlebar text changes).

> Set mouse_capture = seamless in the config file, switch DOSBox Staging to fullscreen, start Windows 3.1 using the following BAT file (of course you'll need to adapt the paths) [...] and, after Windows starts, both the host OS and Windows 3.11 mouse cursors are visible (this is bad - only Windows 3.11 cursor should be visible). The Z:\CONFIG.COM -set "window_titlebar=(...) command is crucial to reproduce the problem.

I cannot reproduce this with the current version of the PR. When using the VMware driver in Windows 3.11, seamless mouse operation works perfectly for me and I'm not getting the double cursor either on Windows or macOS. 

Can you please retest? If you're still having the problem, please make a minimal repro ZIP and verify you can repro it on Windows and macOS as well, then send a link to me on Discord and I'll have a second look.

I'm using the following config:

```ini
[cpu]
cpu_cycles_protected = 25000

[sblaster]
sbtype = sb16

[mouse]
mouse_capture = seamless

[autoexec]
mount C "C"
imgmount D "CD\shiv102.iso" -t iso
set PATH=%PATH%;C:\WINDOWS;
set TEMP=C:\WINDOWS\TEMP
C:
cd windows
Z:\CONFIG.COM -set "window_titlebar= program='Microsoft Windows 3.11 (Polish)' cycles=on"
win
exit
```

---

In the meantime, I've tested about 25-30 more games, a mix of my own one "single config per game" setups, and many eXoDOS titles that reconfigure various DOSBox settings at runtime via a `run.bat` batch file (e.g., to run the game with SoundBlaster, GUS, MT-32, etc. sound, similar to your setup). I haven't found any new issues! 🎉 

Given the magnitude of the change, I think we're doing pretty well! 😎  I was certainly expecting a lot more regressions! 😆 I think we're in the last 3-5% territory, and thanks to your thorough testing, we'll reach 100% in no time! 🎉

Let me know how it goes with the current state of this branch.

johnnovak:
Thanks for the review and testing @FeralChild64. 🏅 

I've marked the first and last commit of the work with `feature/config-setting-update-start` and `feature/config-setting-update-end`, respectively. This will help us narrow down the offending commit if we come across any config system related regressions in the coming months.

Issue #4519, opened 2025-09-23
Fix pegging a CPU core with `cpu_cycles = max` at the DOS prompt


andras63:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

0.82.2 release

### What operating system are you using?

Windows 11

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Is your feature request related to a problem? Please describe.

I'm using DOSBox for maintaining old DOS software (DOS base machine control software).
That ancient build environment is entire DOS based, working fine in staging, the resulting builds are run on native DOS (freedos nowadays) embedded 32 bit X86 SOCs.
All project files, reposititories, etc, reside on a linux development server with a samba share mounted in dosbox. While working on them on the linux server, dosbox is sitting in the background in the command prompt waiting for being switched to foreground, so that I can type make... This happens at 100% cpu, even when switched to background.
This blows the power management of may workstation crazy with the normally silent cooling fan going to huricane mode. During real activity - compiling with old dos compiler - it shuts down again.

### Describe the solution you'd like

Apparently, while waiting in the command prompt, the CON read function device_CON::Read() is implemented via CALLBACK_RunRealInt(0x16). CALLBACK_RunRealInt() just waits for input via interrupt in the main DOSBOX_RunMachine(), which is a **hotpolling loop**. That loop drives up one cpu core on the host windows to full 4.2 GHz heating up the CPU.

It would be really save engergy and nerves if some kind of event waiting could be added to the hotpolling loop, at least if the window is not in foreground. Some kind of 'WaitSDLEvent()' sleeping if the windows is displayed and Int16 is active.

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

andras63:
A possible solution to implement this could be using SDL_WaitEventTimeout() instead of SDL_PollEvent() in DOSBOX_PollAndHandleEvents().
This requires SDL2.0 and timeout needs to be synced with the joystick ticks, but that shouldn't be a problem - at least worth a try.

Unfortunately I don't have a working compile environment to compile dosbox myself, but when I have time to bring my clang cross compiler up to a working state I'll give it a try.

johnnovak:
[...]

PR #4520, opened 2025-09-23
[WIP] [for testing] Idle on HLT instruction if `cpu_cycles=max`

FeralChild64:
# Description

(only briefly tested on Linux so far - needs much more testing)

## Related issues

https://github.com/dosbox-staging/dosbox-staging/issues/4519

# Release notes

DOSBox Staging now goes idle for a brief moment if `cpu_cycles=max` and the HLT instruction is executed by the emulated CPU. This way software like:
- DOSidle (for DOS)
- WQGHLT (for Windows 3.1x)
- AmnHLT (for Windows 9x)
- Rain (for Windows 9x)

now has a chance to do their intended work, lowering the host CPU temperature, fan noise, and power consumption.

# Manual testing

- [ ] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/cpu/cpu.cpp
+++ b/src/cpu/cpu.cpp
@@ -6,8 +6,9 @@
 
 #include <cassert>
 #include <cstddef>
-#include <sstream>
 #include <memory>
+#include <sstream>
+#include <thread>
 
 #include "config/config.h"
 #include "config/setup.h"
@@ -50,6 +51,8 @@ int CPU_CycleLimit    = -1;
 
 static int old_cycle_max       = CpuCyclesRealModeDefault;
 static bool legacy_cycles_mode = false;
+
+static bool should_idle_on_hlt = false;
 
 // Cycles settings for 'modern mode' only
 static struct {
@@ -2267,6 +2270,9 @@ static Bits hlt_decode()
 	} else {
 		CPU_IODelayRemoved += CPU_Cycles;
 		CPU_Cycles = 0;
+		if (should_idle_on_hlt) {
+			std::this_thread::yield();
+		}
 	}
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


johnnovak:
I did not quite understand how this can possibly work, then found this in the [documentation of DOSidle](https://maribu.home.xs4all.nl/zeurkous/download/mirror/dosidle/DOSIDLE.TXT) which explained things:

```
-Hlt:
Selects cooling method: HLT instructions.
This switch causes DOSidle to execute HLT instructions every time it detects
that the system is idle. Since most processors have a "Suspend on HLT"
feature, this instruction places them in suspend mode, thus saving energy and
achieving very good cooling.
This switch is enabled by default.
```

I'm a bit wary of yielding threads on Windows. We'll need to performance test this on Windows as the OS scheduler might work very differently than on Linux and macOS.

Better yet, maybe we could call `YieldProcessor()` on Windows instead which doesn't give the thread back to the OS scheduler.

Alternatively, and I think this is the best solution, we could call the `_mm_pause()` (x64) and `__yield()` (arm) intrinsincs directly on all platforms (`YieldProcessor()` seems to be an alias for these). These put the CPU into a "pause" state for a few tens of nanoseconds (it's supposed to be used in busy loops to ease the load), so we might need to execute them a few hundred times in a loop, perhaps, to get the power saving benefits.

- YieldProcessor
  https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-yieldprocessor

- "use _mm_pause in poll loop to conserve some energy" ([source](https://community.intel.com/t5/Intel-Moderncode-for-Parallel/Low-latency-application-for-sparse-load/m-p/877405/highlight/true#M3204))

FeralChild64:
Yes, this is going to need quite some testing.

FeralChild64:
Unfortunately, this change does not seem to bring much improvement - I did a test on Linux with DOSBox Staging pinned to the first CPU core, with `cpu_cycles=max`:

```bash
taskset 0x01 <dosbox command>
```

- without DOSIdle running - 99%-100% core utilization
- DOSIdle running, on `main` - 24% core utilization
- DOSIdle running, with my patch - 22% core utilization

This attempt causes more risk than it's worth - abandoning.

PR #4521, opened 2025-09-24
Improve the internal Unicode engine

FeralChild64:
# Description

Switch to C++20 types where applicable
Change the internals to work on 32-bit values
Add support for UTF-16 encoding
Add unit tests for UTF-8 and UTF-16 encoding/decoding

# Manual testing

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:


tests/unicode_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+	// 3-byte encoding (range #2)
+	test_utf8(0x00e000, "\xee\x80\x80");
+	test_utf8(0x00f123, "\xef\x84\xa3");
+	test_utf8(0x00ffff, "\xef\xbf\xbf");
+
+	// 4-byte encoding
+	test_utf8(0x010000, "\xf0\x90\x80\x80");
+	test_utf8(0x031234, "\xf0\xb1\x88\xb4");
+	test_utf8(0x10ffff, "\xf4\x8f\xbf\xbf");
```
johnnovak, line 56: It would be better to have separate tests for each heading.

src/misc/unicode.cpp
```diff
@@ -1120,24 +957,41 @@ static bool get_tokens(const std::string& line, std::vector<std::string>& tokens
...
 }
 
-static bool get_hex_16bit(const std::string& token, uint16_t& out)
+static bool get_hex_32bit(const std::string& token, char32_t& out)
 {
+	if (!token.starts_with("0x")) {
+		return false;
+	}
+
+	if (token.size() < 6 || token.size() > 10 || token.size() % 2 != 0) {
```
johnnovak: Maybe explain the meaning of these magic length constants? Extract them to constants with meaningful names?

src/misc/unicode_encodings.cpp
```diff
@@ -0,0 +1,457 @@
...
+// For UTF-16 encoding explanation see here:
+// - https://en.wikipedia.org/wiki/UTF-16#Description
+
+static char16_t maybe_reverse_endianness(const char16_t value, const bool reverse)
+{
+	if (!reverse) {
+		return value;
+	}
+
+	const auto result = (value / 0x100) + (value % 0x100) * 0x100;
```
johnnovak: Explain what this is doing? Looks super weird 😅 Can't you do this with explicit byte swapping in a buffer instead of maths? That would be easier to understand, and the compiler will optimise it anyway.

src/misc/unicode_encodings.cpp
```diff
@@ -0,0 +1,457 @@
...
+		if (!previous) {
+			// We don't have the first surrogate
+			warn_decode_problem(i);
+			str_out.push_back(static_cast<char32_t>(UnknownCharacter));
+			continue;
+		}
+
+		const auto result = (*previous - Utf16::LowSurrogateBase) * 0x400 +
+		                    (token - Utf16::HighSurrogateBase) +
+		                    Utf16::EncodeThresholdSurrogate;
```
johnnovak, line 352: What's that `* 0x400` thing? 😅 Isn't there a more direct way to express this? This seems like micro-optimisation, and most likely the compiler will do a more than good enough job at such things.

src/misc/unicode_encodings.cpp
```diff
@@ -0,0 +1,457 @@
...
+			// Just copy the value
+			str_out.push_back(static_cast<char16_t>(code_point));
+			continue;
+		}
+
+		// Encode using two surrogates
+		const auto tmp = code_point - Utf16::EncodeThresholdSurrogate;
+
+		const auto high = Utf16::LowSurrogateBase + (tmp / 0x400);
+		const auto low  = Utf16::HighSurrogateBase + (tmp % 0x400);
```
johnnovak: You know what I'm about to say... 😆 Can we _not_ do bit twiddling if there's a way to avoid it, please 😄 At least introduce some inlined helper function if you really wanna do it this way.

src/misc/unicode_encodings.cpp
weirddan455: That took me a minute too. I would have caught on quicker if you at least used the bit shift operator instead of multiplies and divides.

johnnovak: > That took me a minute too. I would have caught on quicker if you at least used the bit shift operator instead of multiplies and divides.

Yeah with shifts it would be borderline acceptable, but better to introduce some helpers. Actually, can't you use the BE/LE conversion functions from `byteorder.h`?

FeralChild64: OK, changed the code to use function from `byteorder.h` - I wasn't aware of it.

BTW, after something like 8 semesters of algebra during my university studies, I often forget that bit shifts are more intuitive to 99% of developers than division and modulo :D

src/misc/unicode.cpp
FeralChild64: Added explanation, simplified this a bit.

src/misc/unicode_encodings.cpp
FeralChild64: Wikipedia tells to do so (link in the comment above the function):

```
To encode U+10437 (𐐷) to UTF-16:

    Subtract 0x10000 from the code point, leaving 0x0437.
    For the high surrogate, shift right by 10 (divide by 0x400), then add 0xD800, resulting in 0x0001 + 0xD800 = 0xD801.
    For the low surrogate, take the low 10 bits (remainder of dividing by 0x400), then add 0xDC00, resulting in 0x0037 + 0xDC00 = 0xDC37.

To decode U+10437 (𐐷) from UTF-16:

    Take the high surrogate (0xD801) and subtract 0xD800, then multiply by 0x400, resulting in 0x0001 × 0x400 = 0x0400.
    Take the low surrogate (0xDC37) and subtract 0xDC00, resulting in 0x37.
    Add these two results together (0x0437), and finally add 0x10000 to get the final code point, 0x10437.
```

Defined 0x400 as `SurrogateRangeSize`, hopefully this is more clear now.

src/misc/unicode_encodings.cpp
FeralChild64: As above, this was implemented according to description from Wikipedia. Added a constexpr value for 0x400.

tests/unicode_tests.cpp
FeralChild64: OK, made the tests much more granular.

Issue #2681, opened 2023-07-13
Unicode in paths are not properly being recognized.


jpetriejr:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

0.80.1 release

### What Operating System are you using?

Windows 11

### If Other OS, please describe

Windows 10, Probably Windows 7 as well.

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

![DOSBox-log](https://github.com/dosbox-staging/dosbox-staging/assets/25973833/be6e41d1-76b9-4eee-a5e7-0d7bbc78c0a4)

If a path used to load a custom configuration file has Unicode the characters are not properly recognized and DOSBox will fail to load properly.  The attached image shows the log wherein the Windows username contains a Unicode character and is not properly recognized in the passed in config file path.  The username is: **Jaś**

You'll also see that the default paths in the AppData director cannot be created because it attempts to convert the username to ASCII as **Jas** which will not work (_obviously_).

This came to my attention because users of my software (Classic Gaming Arena) reported that DOSBox was not starting for them and after some investigation this is what I discovered.  I did testing of this from a Windows 11 VM in order to replicate the issue with the same setup as the user who reported the issue where their Windows username contained Unicode.

For reference, the actual path being passed in is: C:\Users\Jaś\.cgaclient\users\1000\cga-2559.conf

### Steps to reproduce the behaviour.

1.  Either create a Windows user with Unicode in their username or simply pass a custom config file into DOSBox on the command line with a path with Unicode in it.
2. Create a shortcut to test with.  I did this to make sure it wasn't my software causing the issue and it turned out to just be DOSBox.

If you need anything else from me, please let me know.

Thanks!

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

johnnovak:
Thanks for the report @jpetriejr

shermp:
A couple of potential areas to look at:

- DOSBox is not receiving Unicode arguments via `main`. A quick look at `main.cpp` indicates this could be the case. AFAIK, `main()` on Windows provides arguments in the local code page, which *may* (but probably isn't) `utf8`. Windows provides a `wmain()` entrypoint for unicode (utf-16le) arguments
[...]

Issue #3450, opened 2024-02-24
Crash when trying to mount a non existent ISO file on autoexec


AndywinXp:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### What Operating System are you using?

Windows 10

### Relevant hardware info

AMD Ryzen 7 5800H, NVIDIA GeForce RTX 3070 Laptop GPU

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

If an `IMGMOUNT` statement pointing to a non existent ISO file is available on the config file, DOSBox Staging crashes at start-up. The application does not crash if the `IMGMOUNT` statement is performed in the DOS command line.

[dosbox-staging.conf.txt](https://github.com/dosbox-staging/dosbox-staging/files/14392620/dosbox-staging.conf.txt)

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Edit the `dosbox-staging.conf` config file to include an IMGMOUNT instruction pointing to a non existent file (e.g. `IMGMOUNT E "G:\My-Beautiful-File.iso" -t iso`)
2. Run DOSBox Staging
3. Crash! ` libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in __wide_to_char: Illegal byte sequence `

### Your configuration

```ini
(config file attached)
```

### Provide a Log

```text
`libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in __wide_to_char: Illegal byte sequence`
```

### Code of Conduct & Contributing Guidelines

- [X] Yes, I agree.

shermp:
I have tried, but I cannot reproduce this on Windows 11, with either my own testing, or the provided config file. I'm using 0.81.

johnnovak:
@AndywinXp Maybe some character encoding issue is at play. Can you reproduce the crash on ypur system again, then upload that exact config that caused the crash?

AndywinXp:
Sure thing. I just reproduced the crash with the following config file.
[dosbox-staging.zip](https://github.com/dosbox-staging/dosbox-staging/files/14406260/dosbox-staging.zip)

I don't know if it helps, but since it might be a character encoding issue, my system language is Italian.

weirddan455:
[...]

PR #4526, opened 2025-09-28
Windows activeCodePage and resource file build improvements

shermp:
# Description

This PR has a couple of minor tidy-ups regarding the Windows resource file, and uses the (relatively) new functionality to set the active code page to UTF-8 using an unpackaged application manifest.

- I've created a `windows` subfolder in `src/platform`, and moved the `unistd.h` and `winres.rc` there
- The RC file is now a template configured by cmake `configure_file`. The primary purpose is to use the cmake VERSION variable, and therefore should be one less thing to change during releases
- An unpackaged application manifest has been added to set the `activeCodePage` property to `UTF-8`. This is supported in Windows 10 (1903 and later) and 11. Older versions of Windows will behave as they do now.

With setting the `activeCodePage` property to `UTF-8`, the Win32 API treats narrow, or "ANSI" strings as UTF-8 instead of the current local code page. This is especially useful for file operations that include characters outside of the users code page. It affects dosbox command line arguments such as `--conf` and `--working-dir`, config and resource paths stored in the users home directory ("foreign" usernames now work), and paths stored in the config file (except autoexec).

This is the guide I used to do this: https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

## Related issues

This should fix https://github.com/dosbox-staging/dosbox-staging/issues/2681

# Release notes

On Windows 10 (1903 or later) and 11, DOSBox Staging now supports unicode file and directory paths in the following situations:
- At the command line with options such as `--conf` and `--working-dir`
- User directory (`C:\users\<username>\appdata\Roaming`) where <username> contains characters outside of your local language
- Folder/file paths set in config files (but NOT autoexec)

Older versions of Windows will continue to be restricted in what characters can be used in paths.

# Manual testing

Extract the attached ZIP file on Windows and do the following
[テストフォルダ.zip](https://github.com/user-attachments/files/22580088/default.zip)

1. Add a copy of 4GMGSMT.SF2 soundfont to the `testテストフォルダ` folder OR modify the `dosbox.conf` file to add our own soundfont in that directory
2. Run dosbox from the `テストフォルダ` (with the local `dosbox.conf`
3. Dosbox should launch, with the soundfont loaded, the localdrive folder mounted as C: and dosbox running at 33333 cycles
4. For bonus points, create a new local user account with non-local characters in the username and run dosbox. It should create and use a config in the users appdata directory

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:


johnnovak:
Nice work @shermp and great find 🎉 

Have you tested the generated installers to confirm the app icon still shows up fine and the version number is correct when right-clicking the executable?

A minor comment: could you prefix the first two commit messages with `build: `?

Also, would this change fix these two tickets?

- https://github.com/dosbox-staging/dosbox-staging/issues/2681
- https://github.com/dosbox-staging/dosbox-staging/issues/3450

shermp:
I haven't checked the installers, but locally I've checked the built dosbox.exe for its properties and icon. Changing the VERSION and rebuilding changes the exe properties.

This should indeed fix #2681 and was the issue I tried to find. I don't know about the other.

johnnovak:
> I haven't checked the installers, but locally I've checked the built dosbox.exe for its properties and icon. Changing the VERSION and rebuilding changes the exe properties.

Could you please do that when you change the RC file?

> This should indeed fix #2681 and was the issue I tried to find. I don't know about the other.

Great! Could you please update the related tickets section up to. Thanks! 😄

shermp:
> Could you please do that when you change the RC file?

Done.
 
> Great! Could you please update the related tickets section up to. Thanks! 😄

Also done.

shermp:
I would like a non english speaking person to test this if possible. @FeralChild64 as our resident encoding and translation guru, are you able to smoke test this on Windows?

johnnovak (approved):
Very nice work @shermp

shermp:
Also, here's the log line when running as a user with Japanese characters in the username:

`CONFIG: Loaded primary config file 'C:\Users\テストAccount\AppData\Local\DOSBox\dosbox-staging.conf'`

FeralChild64:
~~One remark - it's possible that after these changes the `switch_console_to_utf8` and `restore_console_encoding` from `sdlmain.cpp` won't be needed anymore.~~

[edit] Just tested - unfortunately, we still need this :/

Torinde:
Related:
- https://github.com/dosbox-staging/dosbox-staging/issues/2619

johnnovak:
> Related:
> 
>     * [Map high-ASCII chars in DOS filenames to UTF-8 on the host #2619](https://github.com/dosbox-staging/dosbox-staging/issues/2619)

Nope, it's not related, not really. That's a different issue.

Torinde:
My results:
- prior to this PR - even adding only a `Я` to the Staging folder results in it closing during startup.
- with this PR I can run Staging from folder named `Чч!@№$%€§()_+ЯВЕРТЪУИОПШЩЮявертъуиопшщюасдфгхйкл;'АСДФГХЙКЛЗѝЦЖБНМзьцжбнм,.ЬЍ^#–„“ІV`

FeralChild64:
@shermp Can we merge it, or are we still waiting for something?

shermp:
Fine by me. Someone will have to do so though as I don't have permission.

PR #4535, opened 2025-10-04
Use emulated HLT to lower the host CPU usage in some cases

FeralChild64:
# Description

This PR implements two changes to help lowering the CPU usage:

1. Our BIOS (interrupt 16h, functions 0x00 and 0x10) now executes the emulated HLT instruction if it couldn't get the keystroke, similarly to the `DOSIdle` tool. This lowers the host CPU usage in our DOS command prompt and possibly some other utilities which seat idle most of the time.
I couldn't find any game triggering the code I have modified, so I assume this should be 99% safe. The remaining 1% are power users wanting to use the real `DOSIdle`, `FDAPM`, and similar tools (possibly combining them with JEMM386 or that like) - these folks can disable the new behavior using the `cpu_idle` setting.

2. The change from point 1 won't bring any benefit from running the command prompt under Windows 3.1x, as it provides own interrupt 16h routines and hides our implementation.
For this use case I did a change in our `CON:` device implementation; if running under Windows 3.x, the `Read` routine first waits in the loop until any character is ready in the keyboard buffer - but in each and every iteration it releases the time slice back to the emulated Windows. This way the Windows 3.1x gets more CPU time for any other purposes. If the user happens to be running the WQGHLT driver, it will execute the HLT instructions, giving host CPU power usage benefits similar to point 1.

Unfortunately, not all the features I have planned could have been implemented:
- it seems we can't easily detect the Windows 3.1x being idle - most likely we would need to implement the APM
- I couldn't get any interrupt 2Fh idle notifications from protected mode games - I'm guessing we would need to provide our own DPMI implementation

[edit] WQGHLT is surprisingly hard to find nowadays - I'm attaching it here, it's GPL 2 licensed:

[WQGHLT.ver.1.00.English.zip](https://github.com/user-attachments/files/22703825/WQGHLT.ver.1.00.English.zip)

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4519

# Release notes

DOSBox Staging uses less host CPU power if it's shell is idle (if you want to use othe DOS idling software, you can disable the new behavior using the `cpu_idle` option to prevent conflicts) - for `cpu_cycles = max`, on a fast desktop CPU, this can reduce core usage from 100% to around 25%.
Additionally, our shell is now more multitasking friendly, so for Windows 3.1x DOS window software like WQGHLT will work better.

# Manual testing

For testing you need a tool to measure CPU usage (I was using `btop` on Linux). To make things easier, you can pin DOSBox to one CPU core (for example start it with `taskset 0x1 dosbox` command on Linux).

1. Set `cpu_cycles = max`, start DOSBox - observe the CPU usage change after commands:

```bat
config -set cpu_idle=off
config -set cpu_idle=on
```

2. Set `cpu_cycles = max`, start Windows 3.1x with WQGHLT driver installed, double click _MS-DOS Prompt_ icon in the _Main_ group - DOSBox should use much less than 100% of the assigned CPU core (without this PR, it goes to 100%)

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/dev_con.h
+++ b/src/dos/dev_con.h
@@ -9,6 +9,7 @@
 
 #include <cstring>
 
+#include "dos/dos_windows.h"
 #include "ints/bios.h"
 #include "ints/int10.h"
 #include "utils/ascii.h"
@@ -79,9 +80,26 @@ bool device_CON::Read(uint8_t* data, uint16_t* size)
 		readcache = 0;
 	}
 	while (*size > count) {
-		reg_ah = is_machine_ega_or_better() ? 0x10 : 0x0;
+		// If running under guest Windows, check if any key is ready;
+		// until then keep releasing our time slice - this allows the
+		// guest Windows to do more processing while we are waiting.
+		// Do not do this if Windows is not running - that would negate
+		// the benefit of running the HLT instruction by our BIOS.
+		if (WINDOWS_IsStarted()) {
+			while (true) {
+				reg_ah = is_machine_ega_or_better() ? 0x11 : 0x1;
+				CALLBACK_RunRealInt(0x16);
+				if (reg_ax != 0) {
+					break;
+				}
+				WINDOWS_ReleaseTimeSlice();
+			}
+		}
 
+		// Read the key from the buffer
+		reg_ah = is_machine_ega_or_better() ? 0x10 : 0x0;
 		CALLBACK_RunRealInt(0x16);
+
 		switch (reg_al) {
 		case Ascii::CarriageReturn:
 			data[count++] = Ascii::CarriageReturn;


src/dos/dev_con.h
```diff
@@ -79,9 +80,26 @@ bool device_CON::Read(uint8_t* data, uint16_t* size)
 		readcache = 0;
 	}
 	while (*size > count) {
-		reg_ah = is_machine_ega_or_better() ? 0x10 : 0x0;
+		// If running under guest Windows, check if any key is ready;
+		// until then keep releasing our time slice - this allows the
+		// guest Windows to do more processing while we are waiting.
+		// Do not do this if Windows is not running - that would negate
+		// the benefit of running the HLT instruction by our BIOS.
+		if (WINDOWS_IsStarted()) {
```
johnnovak, line 88: Just checking, this is unrelated to the HLT stuff, right? So we don't need `&& CPU_ShouldHltOnIdle()` check in the condition?

src/cpu/cpu.cpp
```diff
@@ -3385,8 +3389,20 @@ void init_cpu_config_settings(SectionProp& secprop)
 	                   "Values lower than 100 are treated as a percentage decrease.",
 	                   DefaultCpuCycleDown));
 
+	pbool = secprop.AddBool("cpu_idle", Always, true);
+	pbool->SetHelp(
+	        "Emulate the HLT instruction if DOSBox is idle ('on' by default). Reduces the CPU\n"
+	        "usage when running the shell or certain applications.\n"
+	        "You might need to disable it to use other DOS idling software.");
```
johnnovak: You start with the specifics in the sentence, so how about this instead:

```
Reduce the CPU usage in the DOS shell and in some applications when DOSBox is
idle ('on' by default). This is done by emulating the HLT CPU instruction, so
it might interfere with other power management tools such as DOSidle and FDAPM
when enabled. Try disabling this setting if a game is misbehaving.
```

I'm a bit unsure about the last sentence, maybe leave that off.

FeralChild64: OK, changed - left off the last sequence.

src/dos/dev_con.h
FeralChild64: Right, it's not really related (it's not doing a HLT, it only yields the CPU time slice back to Windows when not needed), and I don't think we need a config option for this; running DOS games under Windows 3.1x is already a really bad idea.

johnnovak: > running DOS games under Windows 3.1x is already a really bad idea.

I certainly agree with that 😄

FeralChild64:
> One more thing, could you perhaps quantify the average CPU usage savings in the DOS shell when idle in the release notes draft?

I could only give information about a fast CPU, which I have. Average CPU might see worse results, I don't have a way to test it.

PR #4538, opened 2025-10-08
Implement MDS/MDF CDROM images

### LINKED ISSUE #3805: Support MDF/MDS CDROM image files

_No response_

None

_No response_

These are formats used by Alcohol 120%.  Would be nice to support them.  MDS is a small file that does the same job as a CUE file, only it's encoded in a binary format rather than text.  MDF is the BIN equivalent.  If we're lucky, all we would need to do is write some parsing code for the small MDS file to read track information and it can use the existing code for BIN/CUE.

Write an MDS parser and see if it works.

Maybe there's existing code from another project we can use.

_No response_

- [X] Yes, I agree.

weirddan455:
# Description

Implements MDS/MDF file types for CDROM images. Chance of regression for CUE/ISO types is very low. If the image does not contain the MDS signature as the first 17 bytes, it falls through to the CUE and ISO handlers as it does currently.

Users won't need to pass any additional flags. Just `imgmount game.mds -t cdrom` as they normally would with CUE or ISO types.

I used CDEMU code as reference but did not directly copy any of their code: https://github.com/cdemu/cdemu/blob/master/libmirage/images/image-mds/parser.c

There's a lot of cases we're not handling but most are not relevant for DOS games. Ex: DVDs and copy protection. There are a few DOS copy protected games but implementing that will need further changes to the CDROM code as a whole. This is a stand-alone PR with very minimal changes to existing code.

## Related issues

Fixes #3805

# Release notes

Add support for MDS/MDF cdrom images

# Manual testing

Tested with Star Wars Dark Forces which someone ripped with Alcohol 120% and contains subchannel data.

Tested with Alone in the Dark 2. Converted myself from BIN/CUE to MDS/MDF with Alcohol 120%. Contains 62 small audio tracks and is a good test that the code is seeking correctly. See previous issue #4445 I confirmed that bug did not re-surface here.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/cdrom.h
+++ b/src/dos/cdrom.h
@@ -351,6 +351,8 @@ private:
 	void PlayNextAudioTrack();
 	bool PlayAudioTrack(const Track& track, const uint32_t sector_offset);
 
+	bool LoadMdsFile(const char *filename);
+
 	// Private functions for cue sheet processing
 	bool  LoadCueSheet(const char *cuefile);
 	bool  GetRealFileName(std::string& filename, std::string& pathname);
--- a/src/dos/cdrom_image.cpp
+++ b/src/dos/cdrom_image.cpp
@@ -5,6 +5,7 @@
 // #define DEBUG 1
 
 #include "cdrom.h"
+#include "cdrom_mds.h"
 
 #include <cassert>
 #include <cctype>
@@ -515,7 +516,7 @@ CDROM_Interface_Image::~CDROM_Interface_Image()
 bool CDROM_Interface_Image::SetDevice(const char* path)
 {
 	std::lock_guard lock(player.mutex);
-	const bool result = LoadCueSheet(path) || LoadIsoFile(path);
+	const bool result = LoadMdsFile(path) || LoadCueSheet(path) || LoadIsoFile(path);
 	if (!result) {
 		// print error message on dosbox console
 		char buf[MAX_LINE_LENGTH];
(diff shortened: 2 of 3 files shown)


src/dos/cdrom.h
```diff
@@ -272,14 +272,15 @@ class CDROM_Interface_Image final : public CDROM_Interface
...
-		bool                       mode2       = false;
+		std::shared_ptr<TrackFile> file            = nullptr;
+		uint32_t                   start           = 0;
+		uint32_t                   length          = 0;
+		uint32_t                   skip            = 0;
+		uint16_t                   sector_size     = 0;
+		uint16_t                   subchannel_size = 0;
+		uint8_t                    number          = 0;
+		uint8_t                    attr            = 0;
+		bool                       mode2           = false;
```
johnnovak, line 283: Is "attr" and "mode2" the real name in the specs or something?

src/dos/cdrom_image.cpp
```diff
@@ -1174,6 +1174,235 @@ static std::string dirname(char* file)
...
+	}
+	if (tracks.empty()) {
+		LOG_ERR("CDROM: Failed to find any tracks");
+		return false;
+	}
+	Track lead_out = {};
+	lead_out.start = tracks.back().start + tracks.back().length;
+	tracks.push_back(lead_out);
+	return true;
+}
```
johnnovak, line 1425: This method is way too big. Can you break it up into 2-3 or more smaller methods per logical function/task? A trivial candidate is the stuff happening in the for loop, but even that is too big.

src/misc/fs_utils.cpp
```diff
@@ -169,3 +169,59 @@ int64_t get_native_file_position(const NativeFileHandle handle)
...
+	}
+	uint8_t ret = 0;
+	stream.read(reinterpret_cast<char *>(&ret), sizeof(uint8_t));
+	if (stream.fail()) {
+		return {};
+	}
+	return ret;
+}
+
+std::optional<uint16_t> read_native_u16_at(std::ifstream &stream, std::ifstream::pos_type pos)
```
johnnovak: Confused about the name `read_native_u16_at`. Shouldn't this be called `read_le16_at()`? That would mean it treats the input as little-endian, and of course the output is native byte order (that's implied).

If you want to be really explicit, then `read_le16_as_native_u16_at`.

Same comment for all the rest of the functions in this file.

src/dos/cdrom.h
weirddan455: Yeah, they're both from the CDROM spec.

attr is obtained on real CDROM drives from an ioctl by merging two fields together like `attr = (entry.cdte_ctrl << 4) | entry.cdte_adr;` in the Linux path. In DOS, its a single field. For images, all we handle is 0 for audio and 0x40 for data.

The mode is basically whether there are error correction bytes at the end of the sector. Mode 1 has error correction and Mode 2 does not (it can store extra data there instead): https://en.wikipedia.org/wiki/CD-ROM#Sector_structure

src/dos/cdrom_image.cpp
```diff
@@ -1174,6 +1174,235 @@ static std::string dirname(char* file)
...
+	constexpr char MdsSignature[] = {'M', 'E', 'D', 'I', 'A', ' ', 'D', 'E', 'S', 'C', 'R', 'I', 'P', 'T', 'O', 'R', 1};
+	char sig[sizeof(MdsSignature)] = {};
+	file.read(sig, sizeof(MdsSignature));
+	if (file.fail()) {
+		return false;
+	}
+	if (memcmp(sig, MdsSignature, sizeof(MdsSignature))) {
+		// Not as MDS file. Fall through to CUE/ISO handlers.
+		return false;
+	}
```
shermp: How about using `std::array<uint8_t, 17>` here for `MdsSignature` and `sig`? Would be no need to use `sizeof` or `memcmp` then.

shermp:
I can't say I'm a fan of the magic number offsets, at the very least these should be named constants.

Personally, for sections that have a known size/layout, I would consider reading into a packed struct, byteswap fields if necessary then use the struct to do the validation steps. It's then easier to figure out what's going on, and you're less likely to have offset oopsies (been there, done that).

src/dos/cdrom_image.cpp
FeralChild64: Or simply `const std::string`?

shermp: As written, a string won't work because of the trailing `1`, although as I mentioned below, personally I think the version should be validated as part of a wider "header" validation.

FeralChild64: `const std::string MdsSignature = "MEDIA DESCRIPTOR\1";` should work, I believe. Or - even better - check the 1 separately.

weirddan455:
> I can't say I'm a fan of the magic number offsets, at the very least these should be named constants.
> 
> Personally, for sections that have a known size/layout, I would consider reading into a packed struct, byteswap fields if necessary then use the struct to do the validation steps. It's then easier to figure out what's going on, and you're less likely to have offset oopsies (been there, done that).

These are the structs where I calculated the offsets from: https://github.com/cdemu/cdemu/blob/master/libmirage/images/image-mds/image-mds.h

The reason I didn't use them is they're filled with a bunch of "dummy" fields (look at `MDS_Header` for example, there's 4 of them) and also fields that are irrelevant for our purposes.

I guess I could use them if it makes the code more readable. I was on the fence about it. I could basically copy what they've done, after changing the Glib types to standard `uint*_t` types.

shermp:
Personally, I don't have a problem with dummy fields. It lets the reader know "hey, there's some unknown stuff here", and maybe someone will be inspired to do further reverse engineering of the format in the future.

johnnovak:
> Personally, I don't have a problem with dummy fields. It lets the reader know "hey, there's some unknown stuff here", and maybe someone will be inspired to do further reverse engineering of the format in the future.

That's a good point, and I agree.

src/dos/cdrom_image.cpp
weirddan455: `std::array` doesn't really help because there's no == overload between that and a C-style array. I'm reading the signature as part of the header struct now and we're using C-style arrays in most of our structs by convention.

`std::string` doesn't work as it's not null terminated.

src/misc/fs_utils.cpp
weirddan455: These functions are all gone now. I'm reading the data in as structs instead of individual integers.

src/dos/cdrom_image.cpp
johnnovak: Yeah this is fine.

src/dos/cdrom_image.cpp
johnnovak: Picking this PR up again, this is my last outstanding comment. It would be good to break this big function into 2-3 smaller functions if it's possible.

src/dos/cdrom_mds.h
```diff
@@ -0,0 +1,206 @@
...
+	uint16_t dummy1[2];
+	uint16_t bca_len;
+	uint32_t dummy2[2];
+	uint32_t bca_data_offset;
+	uint32_t dummy3[6];
+	uint32_t disc_structures_offset;
+	uint32_t dummy4[3];
+	uint32_t session_block_offset;
+	uint32_t dpm_blocks_offset;
+};
```
shermp, line 34: I would suggest either packing the struct (using `#pragma pack()` or similar), or `static_assert` the size of the struct to avoid any unexpected surprises. I tend to go for the `static_assert` these days if I am reasonably sure the compiler will not add extra padding, although the former is probably the more correct way to do it.

Same comment for the other structs as well.

shermp:
@weirddan455 thank you for the struct approach. It's much clearer to read now.

src/dos/cdrom_mds.h
johnnovak: `#pragma pack(push, 1)` is on line 13, and `#pragma pack(pop)` on line 204.

I think it would be better to surround just the structs with the pragmas for clarity, even if that means more pragmas.

`static_assert`ing the size as well is not a bad idea.

shermp: Oops, missed that, thanks. I'm used to see it surrounding individual structs rather than globally for an entire header.

johnnovak: Yeah me too, I think we should use that style.

src/dos/cdrom_image.cpp
weirddan455: Extracted some stuff into functions

src/dos/cdrom_mds.h
weirddan455: Added `static_assert` for the sizes.

I disagree with doing push and pop for every struct. It greatly increases the likely-hood of forgetting a single `pop` and then the `pack 1` gets spilled out into the .cpp file that included this header and wrecks havok.

I made this a new header file specifically for Mds related structs so that the entire thing can be pragma packed.

The one that is intended to be used on a per-struct basis is something like `__attribute__((packed))` but that is non-standard.

src/dos/cdrom_image.cpp
```diff
@@ -1174,6 +1175,255 @@ static std::string dirname(char* file)
...
+	if (mds_header->session_block_offset == 0) {
+		LOG_ERR("CDROM: Invalid MDS file");
+		return {};
+	}
+
+	return mds_header;
+}
+
+static std::optional<MdsSessionBlock> read_and_validate_mds_session_block(std::ifstream &file, std::ifstream::pos_type pos)
```
johnnovak: second arg const?

src/dos/cdrom_mds.h
johnnovak: Yeah, okay, I think that's fine then. I kinda missed this file only contains packed structs. Maybe put an extra comment at the top of the file, explaining with the packing removed the code would break?

src/dos/cdrom_image.cpp
weirddan455: fixed

src/dos/cdrom_mds.h
weirddan455: Done

PR #4544, opened 2025-10-13
Warn about MOUSE.COM options which are not going to be implemented

FeralChild64:
# Description

`MOUSE.COM` now prints out more descriptive error messages for several MS-DOS switches, which are most likely not going to be implemented.

# Release notes

(not needed - `MOUSE.COM` is a new tool, release notes already given in other PRs).

# Manual testing

Try to run `MOUSE.COM` with switch like `/R1`, `/S50`, or `/le` - a descriptive error message should be displayed, telling the user what to do instead.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/dos/programs/mouse.cpp
+++ b/src/dos/programs/mouse.cpp
@@ -8,6 +8,7 @@
 #include "dos/dos.h"
 #include "dos/dos_windows.h"
 #include "hardware/input/mouse.h"
+#include "misc/notifications.h"
 #include "more_output.h"
 #include "utils/checks.h"
 
@@ -29,12 +30,6 @@ void MOUSE::Run()
 	// TODO: Add support for mouse driver unloading: 'off' (as in Microsoft
 	// driver) or /U switch (drivers from DR-DOS, Mouse Systems, CtMouse).
 
-	// TODO: Implement options to change mouse sensitivity: /Vn (vertical),
-	// /Hn (horizontal), /Sn (both), Microsoft accepts values 5-100.
-
-	// TODO: Implement option /Rn to change interrupt rate; n can be
-	// 1 (30Hz), 2 (50Hz), 3 (100Hz), or 4 (200Hz).
-
 	// TODO: Find out what /Mn option of the Microsoft driver exactly does.
 
 	// TODO: Implement missing mouse driver functions and the relevant
@@ -43,10 +38,7 @@ void MOUSE::Run()
 	// cursor delay (/Nn, values 0-255), ballistic curve selection (/Pn,
 	// values 1-4), double speed threshold setting (/Dn).
 
-	// TODO: Implementing original options to change the driver language
-	// value: /LD (German), /LE (Spanish), /LF (French), /LI (Italian),
-	// /LK (Korean), /LJ (Japanese), /LNL (Dutch), /LP (Portuguese),
-	// /LS (Swedish), /LSF (Finnish)
+	RemoveUnsupportedOptions();
 
 	// The quiet mode should not inhibit error messages - checked with
 	// Microsoft Mouse Driver v9.01
@@ -55,25 +47,6 @@ void MOUSE::Run()
 	// Microsoft option to load the driver, currently the only supported
 	// action
 	cmd->FindExistRemoveAll("on");
-
-	// Check if user requested to proble for mouse port
-	bool warn_port_selection = cmd->FindExistRemoveAll("/f");
-	// Check if user requested PS/2 (/z), bus (/b), inport (/i1, /i2), or
-	// serial (/c1 - /c4) mouse port
-	for (const auto option : { "/z", "/b", "/i1", "/i2",
-	                           "/c1", "/c2", "/c3", "/c4" }) {
-		if (cmd->FindExistRemoveAll(option)) {
-			warn_port_selection = true;
-		}		
-	}
-	if (warn_port_selection) {
-		WriteOut(MSG_Get("PROGRAM_MOUSE_PORT_SELECTION"));	
-	}
-
-	// This option disables hardware mouse cursor on some cards
-	if (cmd->FindExistRemoveAll("/y")) {
-		WriteOut(MSG_Get("PROGRAM_MOUSE_HARDWARE_CURSOR"));
-	}
 
 	const bool has_option_low_memory = cmd->FindExistRemoveAll("/e");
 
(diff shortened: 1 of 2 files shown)


FeralChild64:
BTW. I intend to eventually get rid of `Z:\MOUSECTL.COM` and move all it's functionality to `Z:\MOUSE.COM` - but probably only once we migrate to `SDL 3` and I rework the dual-mouse gaming to use `SDL 3` API instead of `ManyMouse`.

I expect this migration will enforce some changes to the mouse mapping, and I don't really want to change the mapping behavior each time we release a new DOSBox Staging major version.

src/dos/programs/mouse.cpp
```diff
@@ -115,6 +88,80 @@ void MOUSE::Run()
...
 }
 
+void MOUSE::RemoveUnsupportedOptions()
+{
+	// Due to the nature of DOSBox host mouse driver, these options are
+	// probably never going to be supported
+
+	// Mouse hardware port selection: probe (/f), PS/2 (/z), bus (/b),
+	// inport (/i1, /i2), or serial (/c1 - /c4).
+	// Not feasible to be implemented, we are a virtual (host) mouse driver.
```
johnnovak, line 98: Good call, and I agree 👍🏻

src/dos/programs/mouse.cpp
```diff
@@ -115,6 +88,80 @@ void MOUSE::Run()
...
+	if (cmd->FindExistRemoveAll(
+	            "/f", "/z", "/b", "/i1", "/i2", "/c1", "/c2", "/c3", "/c4")) {
+		NOTIFY_DisplayWarning(Notification::Source::Console,
+		                      "DOS",
+		                      "PROGRAM_MOUSE_PORT_SELECTION");
+	}
+
+	// Disabls hardware mouse cursor on some cards, we are not emulating
```
johnnovak: Typo: Disable

src/dos/programs/mouse.cpp
```diff
@@ -115,6 +88,80 @@ void MOUSE::Run()
...
+	if (cmd->FindExistRemoveAll(
+	            "/LD", "/LE", "/LF", "/LI", "/LK", "/LJ",
+	            "/LNL", "/LP", "/LS", "/LSF")) {
+		NOTIFY_DisplayWarning(Notification::Source::Console,
+		                      "DOS",
+		                      "PROGRAM_MOUSE_LANGUAGE");
+	}
+
+	// Switches below are not implemented, because our driver accepts wider
+	// range of values; thus they are skipped, mainly to avoid confision
```
johnnovak: Typo: confusion

src/dos/programs/mouse.cpp
```diff
@@ -63,29 +69,100 @@ void MOUSE::Run()
...
-	if (MOUSEDOS_IsDriverStarted()) {
+	// Check if we have a mouse driver running
+	const auto is_builtin_driver_started  = MOUSEDOS_IsDriverStarted();
+	const auto is_3rdparty_driver_started = !is_builtin_driver_started &&
+                                                IsAnyMouseDriverStarted();
+
+	const auto is_driver_started = is_builtin_driver_started ||
+	                               is_3rdparty_driver_started;
+
+	// Whether we have something to do after the driver is startup
```
johnnovak: "driver has been started"

johnnovak:
@FeralChild64 Great stuff, man! 🎉 I'm very glad to see moving more obscure settings/options from the config into these utilities. Just summarising my general thoughts about the direction where this is going:

- I agree with the eventual removal of `MOUSETCTL`; moving all that runtime config functionality into `MOUSE.COM`. is a good idea 👍🏻 Doing it after the SDL 3 migration sounds about right, yeah.
- How about keeping the most useful, most often used mouse config options in the config (e.g., immediate mode), while obscure and very rarely used stuff can indeed go to `MOUSE.COM` (such as `modern`). Kinda like the 80/20 rule: 80% of the stuff people generally use for gaming stays in the config, then niche and obscure options should belong to `MOUSE.COM`. This approach would have a few important benefits:
  - It would be better if immediate mouse mode was not buried in the 20% "minority options" in `MOUSE.COM` as it's a pretty important option and it makes a big improvement to mouse smoothness in most games.
  - Probably a good rule of thumb would be that most people should _not_ use `MOUSE.COM` at all _unless_ they have some specific niche needs. Or, in other words, if it's a non-niche option, and we would want to set/change from the OSD, it should stay in the config.
  - It's better to have concrete config options for the major, most often used features for per-game configs and the OSD. That will make it a lot easier to save/load per-game configs from the OSD. We could always inject those config options into the `[autoexec]` section, but that's a pain to deal with (especially parsing it back). (Yeah, there's the MIXER command, but that's special because we have so many mixer options multiplied by the number of sound devices.)
- This would also be a good rule (and I think you're doing that already): a mouse option should either exist in the config, _or_ in `MOUSE.COM`, but not in both (except in special cases like multi-mouse setups, where the config would set the global settings, and `MOUSE.COM` would allow per-mouse fine-tunings and overrides, perhaps).

Thanks for considering these thoughts 😄

FeralChild64:
> How about keeping the most useful, most often used mouse config options in the config (e.g., immediate mode), while obscure and very rarely used stuff can indeed go to MOUSE.COM (such as modern).

Both enabling `modern` and disabling `immediate` is rarely needed. And - in fact - till now we know no game which misbehaves with `modern` enabled.

> It's better to have concrete config options for the major, most often used features for per-game configs and the OSD. That will make it a lot easier to save/load per-game configs from the OSD

In such case neither `modern` nor `immediate` should be hidden (not a problem, I can simply drop the 2nd commit), and we should consider exposing the interrupt rate setting for the builtin mouse driver, too. Global interrupt rate setting IMHO makes no sense - different mouse interfaces used to have different capabilities in this regard.

johnnovak:
> Both enabling `modern` and disabling `immediate` is rarely needed 

From memory, I had smoother results in some games on Windows, but that was with my fixed 60 Hz refresh monitor. Maybe at 120 Hz VRR it feels different, or maybe it has nothing to do with it, or maybe all I felt was placebo 😄 Hard to tell 😆 

> And - in fact - till now we know no game which misbehaves with `modern` enabled.
 
Great, that's good info 🎉 

> In such case neither `modern` nor `immediate` should be hidden (not a problem, I can simply drop the 2nd commit), 

Yeah I think that's for the best for now.

src/dos/programs/mouse.cpp
FeralChild64: Fixed.

src/dos/programs/mouse.cpp
FeralChild64: Fixed.

src/dos/programs/mouse.cpp
FeralChild64: This part of the PR has been dropped.

FeralChild64:
OK - for now I have only left the "informational" part of the PR, no change to the config system.

johnnovak:
> OK - for now I have only left the "informational" part of the PR, no change to the config system.

Cool, can you update the PR title, too?

FeralChild64:
Title updated - sorry, I have totally forgotten.

PR #4550, opened 2025-10-17
Reinstate FluidSynth as an internal dependency

shermp:
# Description

The latest version of fluidsynth (2.5.0) no longer depends on Glib. Without the Glib dependency, fluidsynth compiles quickly, so it makes sense to bring it back as a compile time dependency, which simplifies dosbox-staging fluidsynth codebase.

Fluidsynth 2.5.0 has been added to vcpkg. For now, to avoid bumping the vcpkg baseline, I have added the port to our vcpkg overlay directory. My preference is to only bump the baseline to tagged vcpkg releases.

Once this is merged, the dosbox-staging-ext project can drop building fluidsynth, and the `-x '*/libfluidsynth*'` workaround can be dropped in the CI workflows.

## Related issues

This essentially reverts most of the changes made in #4222

# Manual testing

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -927,6 +927,29 @@ if get_option('use_opengl')
 endif
 conf_data.set10('C_OPENGL', opengl_dep.found())
 
+# FluidSynth (depends on glib)
+fluid_dep = optional_dep
+if get_option('use_fluidsynth')
+    fluid_dep = dependency(
+        'fluidsynth',
+        version: ['>= 2.2.3', '< 3'],
+        modules: ['FluidSynth::libfluidsynth'],
+        static: ('fluidsynth' in static_libs_list or prefers_static_libs),
+        not_found_message: msg.format('use_fluidsynth'),
+        default_options: [
+            'default_library=static',
+            'try-static-deps=true',
+            'enable-floats=true',
+            'openmp=disabled',
+            'enable-threads=false',
+            'tests=' + wants_tests.to_string(),
+            'warning_level=0',
+        ],
+        include_type: 'system',
+    )
+endif
+summary('FluidSynth support', fluid_dep.found())
+
 # mt32emu
 mt32emu_dep = optional_dep
 if get_option('use_mt32emu')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -13,6 +13,13 @@ option(
 )
 
 option(
+    'use_fluidsynth',
+    type: 'boolean',
+    value: true,
+    description: 'Enable built-in MIDI support via FluidSynth',
+)
+
+option(
     'use_mt32emu',
     type: 'boolean',
     value: true,
--- a/src/midi/meson.build
+++ b/src/midi/meson.build
@@ -18,6 +18,7 @@ libmidi = static_library(
         alsa_dep,
         coreaudio_dep,
         coremidi_dep,
+        fluid_dep,
         ghc_dep,
         libiir_dep,
         libloguru_dep,


johnnovak:
Nice work @shermp, thanks for your tireless effots on this 🎉🥇 Looks we're going full circle, but for the better now that glib is out of the picture 🗑️😆 

> For now, to avoid bumping the vcpkg baseline, I have added the port to our vcpkg overlay directory. My preference is to only bump the baseline to tagged vcpkg releases.

Why is that so? Apart from using a human-readable tag, what other benefits do the tagged releases have?

shermp:
i think vcpkg releases also coincide with vcpkg the tool generally. Also the VS integrated vcpkg is more likely to use tagged releases.

But really, I'm just not a fan of requiring the latest main branch of vcpkg, because it might not even be available in CI yet, nor is it available for people using VS.

The overlay port works for now (it's identical to upstream), and I'd rather carry that for a few weeks/months until we decide it's time to bump the baseline.

johnnovak:
> The overlay port works for now (it's identical to upstream), and I'd rather carry that for a few weeks/months until we decide it's time to bump the baseline.

Okay, I think that's the key point here, so I'm fine with the overlay port for now.

My approach has been to wait a few weeks for things to "stabilise" when bumping the baseline, so stay a few weeks behind the "bleeding edge". Test on CI, then if it works, it works. Not too sure if using the tags would accomplish the same result because they can't possibly re-test all those 2701 packages with all the transitive deps, etc. In other words, I kinda doubt that the tagged releases are significantly more stable than any random hash.

Also, if we 100% restrict ourselves to tagged releases and shit breaks when upgrading to a new tag, we can only 1) roll back to the previous tag, 2) wait another month or two for the new tagged release. Or use a more recent hash anyway to get the fix... but then we're not using tags anymore 😄 

---

One more thing @shermp  before merging. Could you please update the README.md?

https://github.com/dosbox-staging/dosbox-staging?tab=readme-ov-file#dependencies

- Please put FluidSynth back into the deps table.
- Then update the "Dynamically loaded dependencies" section.

Thanks 😄

shermp:
Yeah, I guess I'm not too attached to the tagged vcpkg release, although it is nice to refer to the `<some date> release` rather than `random commit`, but it doesn't really matter I suppose.

I've updated the readme as requested. I've labeled fluidsynth as a required dependency now, because we don't have an option for it in our CMakeLists.txt

johnnovak:
> Yeah, I guess I'm not too attached to the tagged vcpkg release, although it is nice to refer to the `<some date> release` rather than `random commit`, but it doesn't really matter I suppose.

I think it's worth trying; it's marginally nicer and a bit safer as you said. The proof is in the pudding 😎 We can always abandon that approach easily and revert to using hashes.

> I've updated the readme as requested. I've labeled fluidsynth as a required dependency now, because we don't have an option for it in our CMakeLists.txt

Thanks, that's perfect 🎉 Approving!

shermp:
@johnnovak now that https://github.com/dosbox-staging/dosbox-staging-ext/pull/5 has been merged, if you can make a new release there, I can use that release in this PR, keeps it all nice and tidy then.

johnnovak:
> @johnnovak now that [dosbox-staging/dosbox-staging-ext#5](https://github.com/dosbox-staging/dosbox-staging-ext/pull/5) has been merged, if you can make a new release there, I can use that release in this PR, keeps it all nice and tidy then.

Thanks man, I've just done it:

https://github.com/dosbox-staging/dosbox-staging-ext/releases/tag/v0.83.0-2

shermp:
I've made the change. I assume FluidSynth is statically linked on our MacOS and Linux builds because there's no libfluidsynth dynamic library in the archives, but there is one in the Windows archive. If someone can smoke test this on Linux and MacOS, it would be much appreciated.

Also, @johnnovak as our resident MIDI expert, you might want to test that FluidSynth 2.5.0 doesn't introduce subtle regressions that I'm too inexperienced to miss. My testing amounted to playing a game (Descent 2) with MIDI to ensure it still plays.

johnnovak:
> I assume FluidSynth is statically linked on our MacOS and Linux builds because there's no libfluidsynth dynamic library in the archives, but there is one in the Windows archive.

Yeah, we statically link the libraries on macOS and Linux. It would be nice if we'd do that on Windows too, but I have low motivation to tinker around with that 😄 

> If someone can smoke test this on Linux and MacOS, it would be much appreciated.

I've just done that on macOS, all good 👍🏻 

> Also, @johnnovak as our resident MIDI expert, you might want to test that FluidSynth 2.5.0 doesn't introduce subtle regressions that I'm too inexperienced to miss. My testing amounted to playing a game (Descent 2) with MIDI to ensure it still plays.

Azrael's Tear makes sound on macOS with GeneralUser GS, so ship it :shipit: 🚢 😎 I will spend some more time with FluidSynth before 0.83.0, there are some FS-related tickets open in the issue tracker assigned to me. Until then, we're good 😎

weirddan455:
Tested building with system libs on Linux and Fluidsynth works.

FYI there's a low-priority TODO to make these deps optional in CMake. Not just FluidSynth but things like Munt and SDL2_Net as well. Currently they're all hard dependencies but they were optional in Meson and still listed as optional here: https://github.com/dosbox-staging/dosbox-staging?tab=readme-ov-file#dependencies

Doesn't need to be done in this PR (or even in the next release as it doesn't affect end-users).

shermp:
> Tested building with system libs on Linux and Fluidsynth works.
> 
> FYI there's a low-priority TODO to make these deps optional in CMake. Not just FluidSynth but things like Munt and SDL2_Net as well. Currently they're all hard dependencies but they were optional in Meson and still listed as optional here: https://github.com/dosbox-staging/dosbox-staging?tab=readme-ov-file#dependencies
> 
> Doesn't need to be done in this PR (or even in the next release as it doesn't affect end-users).

This PR changes the readme to say that FluidSynth is required. I honestly don't see any reason to make it optional to be honest.

johnnovak:
@shermp 

New NukedSC55 CLAP release: https://github.com/johnnovak/Nuked-SC55-CLAP/releases/tag/v0.9.0

> FYI there's a low-priority TODO to make these deps optional in CMake. Not just FluidSynth but things like Munt and SDL2_Net as well. Currently they're all hard dependencies but they were optional in Meson and still listed as optional here: https://github.com/dosbox-staging/dosbox-staging?tab=readme-ov-file#dependencies

@weirddan455 
We could make those optional, but I'm with @shermp:

>  honestly don't see any reason to make it optional to be honest.

Yeah I made OpenGL optional by rote, but then I started wondering what's the point. I don't see the use-case for building crippled versions of Staging when all the libraries we depend on are free.

I can only see a case for optionally not compiling in MT-32 support for those license-obsessed Linux freedom fighter enthusiasts, but for the rest... nothing "controversial" about any of the other libs (FluidSynth just plays MIDI, which is an open standard).

So we might wanna rethink carrying toggling features off at compile forward.

weirddan455:
> Yeah I made OpenGL optional by rote, but then I started wondering what's the point. I don't see the use-case for building crippled versions of Staging when all the libraries we depend on are free.

Saves on compile times and some distros might not have our optional deps in their repos. Arch Linux doesn't have Munt for example so I had to get it off the AUR. It's not a licensing issue as far as I know (Arch doesn't have a problem packing proprietary software like Discord even). It's just a niche package and none of their devs volunteered to maintain it.

It's also a pretty easy thing to do and most of the work is already done since it works in Meson.

@FeralChild64 do you have an opinion on this? I know a lot of Gentoo guys love their compile flags.

johnnovak:
> It's also a pretty easy thing to do and most of the work is already done since it works in Meson.
> 
> @FeralChild64 do you have an opinion on this? I know a lot of Gentoo guys love their compile flags.

We can do it if there's interest, it's easy. Personally, I see no point in compiling without OpenGL support, for instance, but whatever, it's easy enough to keep the same optional Meson compile flags 😄

FeralChild64:
> > Yeah I made OpenGL optional by rote, but then I started wondering what's the point. I don't see the use-case for building crippled versions of Staging when all the libraries we depend on are free.
> 
> Saves on compile times and some distros might not have our optional deps in their repos. Arch Linux doesn't have Munt for example so I had to get it off the AUR. It's not a licensing issue as far as I know (Arch doesn't have a problem packing proprietary software like Discord even). It's just a niche package and none of their devs volunteered to maintain it.
> 
> It's also a pretty easy thing to do and most of the work is already done since it works in Meson.
> 
> @FeralChild64 do you have an opinion on this? I know a lot of Gentoo guys love their compile flags.

_Munt_ is indeed a niche, so we can ship it - but _FluidSynth_ or _SDL2_Net_ should, IMHO, definitely be provided by the distros.

FeralChild64:
@shermp Your changes work fine for me when I compile with `CMake`, but the `Meson` build is now broken for me - I have a lot of linking errors, like:

`/usr/lib/gcc/x86_64-pc-linux-gnu/15/../../../../x86_64-pc-linux-gnu/bin/ld: fluidsynth.cpp:(.text._ZN20MidiDeviceFluidSynth15SetChorusParamsERK16ChorusParameters+0x2f): undefined reference to `fluid_synth_set_chorus_group_level'`

johnnovak:
> @shermp Your changes work fine for me when I compile with `CMake`, but the `Meson` build is now broken for me - I have a lot of linking errors, like:
> 
> `/usr/lib/gcc/x86_64-pc-linux-gnu/15/../../../../x86_64-pc-linux-gnu/bin/ld: fluidsynth.cpp:(.text._ZN20MidiDeviceFluidSynth15SetChorusParamsERK16ChorusParameters+0x2f): undefined reference to `fluid_synth_set_chorus_group_level'`

@shermp Yeah, you'll simply just need to revert the Meson build files to the state before we externalised FluidSynth.

shermp:
Ok, Meson support back, can you please test @FeralChild64 ?

FeralChild64:
Still not working. The command:

```
meson setup -Dbuildtype=release -Ddefault_library=shared build
```

fails with error:

```
meson.build:932:3: ERROR: Tried to access unknown option 'use_fluidsynth'.
```

shermp:
> use_fluidsynth

Let's try again. I've added back the option, although setting it to false will break the build, as I have not reinstated the `C_FLUIDSYNTH` defines.

shermp:
Is this all good to merge? Someone else will have to do so.

johnnovak:
> Is this all good to merge? Someone else will have to do so.

Yep, thanks for doing this. Merging.

Issue #4506, opened 2025-09-09
Implement in-memory OpenGL shader cache


johnnovak:
OpenGL shaders should be compiled only once in the same DOSBox Staging session. On subsequent use, we should switch to the compiled shader from the in-memory shader cache instead of loading it from disk and recompiling it.

This will reduce the occasional audio glitch when switching shaders due to video mode changes (we can block the main thread for up to 20-30 ms when we reload and compile shaders).

### Depends on

- https://github.com/dosbox-staging/dosbox-staging/issues/4505

johnnovak:
Done in [Implement in-memory shader caching in the OpenGL renderer #4551](https://github.com/dosbox-staging/dosbox-staging/pull/4551)

PR #4551, opened 2025-10-18
Implement in-memory shader caching in the OpenGL renderer

johnnovak:
# Description

This reimplements GranMinigun's shader caching improvement in a more comprehensive manner in the new renderer architecture (see https://github.com/dosbox-staging/dosbox-staging/issues/3971).

It also fixes some non-critical but annoying issues, such as excessive flicker in some situations when changing shaders, and a "dummy" `crt/hercules` shader being initialised before switching to the real CGA, EGA, VGA, etc. shader variant (appropriate for the machine and CRT auto-shader settings) when showing the mode 03h starting screen.

In general, this substantially decreases the likelihood for audio dropouts when switching shaders (as we're no longer potentially blocking the main thread for 5-30 ms!), but we're already in a pretty good place with most audio devices thanks to @weirddan455's mixer improvements.

This paves the way for implementing a shader preset system in the next PR and merging all those 24 CRT shader variants into a single CRT shader with 24 presets.

## Architectural changes

The big change is that shader switching is not driven by ShaderManager anymore but by the render backend. ShaderManager has less "autonomy" now and is simpler; it just executes "commands" issued by the render backend. Its duties now only include loading shader sources, parsing shader metadata, handling the CRT auto-switching logic, and symbolic shader name aliases.

All this makes the interaction between the SDL GUI, the renderer backend, and the VGA emulation a bit less convoluted.

The entry point for auto-switching is now the render backend which "drives" the whole process (i.e., the OpenGL renderer only at the time being). Shader caching is implemented in the backend—it has to be, as shaders are backend-specific.

# Related issues

-  Implements https://github.com/dosbox-staging/dosbox-staging/issues/4506
 - https://github.com/dosbox-staging/dosbox-staging/issues/3971

# Manual testing

- Shader auto-switching works when resizing the window and toggling fullscreen mode with the `crt-auto` shaders.
- Executed the whole `crt-auto` test suite using [QPV](https://github.com/dosbox-staging/dosbox-staging/wiki/Video-tests-%E2%80%94-CRT-shaders#crt-auto-tests-using-qpv) and [DPaint II](https://github.com/dosbox-staging/dosbox-staging/wiki/Video-tests-%E2%80%94-CRT-shaders#crt-auto-tests-using-deluxe-paint-iie). 
- Setting aliased shader names works (`none`, `sharp`, `bilinear`, etc.)
- Refreshing the current shader with Alt/Cmd+F2 works; changes made to the shader file and the double scanning and sRGB pragmas take effect after the refresh.
- Reloading the current shader (Ctrl/Cmd+F2) leaves the currently active shader intact if the shader cannot be compiled, and an error is logged.
 Reloading the current shader is a no-op in texture output mode.
- Setting a non-existent shader with `glshader` leaves the current shader intact, and an error is logged.
- Setting shaders with `glshader` is a no-op in texture output mode.
- Confirmed the shader cache is used correctly by enabling `#define DEBUG_OPENGL` and inspecting the log output when shaders get auto-switched, manually loaded, or reloaded.
- Confirmed there's no excessive logging when auto-switching shaders and toggling fullscreen.

- [ ] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/render/opengl_renderer.cpp
+++ b/src/gui/render/opengl_renderer.cpp
@@ -24,6 +24,8 @@
 
 CHECK_NARROWING();
 
+// #define DEBUG_OPENGL
+
 // Define to report OpenGL errors
 // #define DEBUG_OPENGL_ERROR
 
@@ -694,6 +696,16 @@ std::optional<OpenGlRenderer::Shader> OpenGlRenderer::GetOrLoadAndCacheShader(
 		assert(shader_info.name == shader_name);
 
 		shader_cache[shader_info.name] = {*maybe_shader_program, shader_info};
+
+#ifdef DEBUG_OPENGL
+		LOG_DEBUG("OPENGL: Built and cached shader '%s'",
+		          shader_name.c_str());
+#endif
+
+	} else {
+#ifdef DEBUG_OPENGL
+		LOG_DEBUG("OPENGL: Using cached shader '%s'", shader_name.c_str());
+#endif
 	}
 
 	return shader_cache[shader_name];


src/gui/render/opengl_renderer.cpp
```diff
@@ -605,21 +606,94 @@ GLuint OpenGlRenderer::BuildShaderProgram(const std::string& source)
...
+		if (!maybe_shader_program) {
+			return {};
+		}
+		assert(shader_info.name == shader_name);
+
+		shader_cache[shader_info.name] = {*maybe_shader_program, shader_info};
+	}
+
+	return shader_cache[shader_name];
 }
```
shermp: Look, I know you're fond of `auto`, but I personally think this is getting a bit too much. As a reviewer, I have absolutely no idea what these types are supposed to be, as I'm usually reviewing on the web interface without an IDE telling me what the types are.

Without knowing the types, it's harder to spot potential issues with the use of said types.

johnnovak: Thanks for taking a look, but this is what we have been doing in the whole codebase in all new code for years now 😄 Never has been a problem for me; you just gotta check out what the functions being called do.

I've been using type inference in all my professional work for like 8+ years now (Scala, Kotlin, and now Java), it's just something to get used to. To do non-surface level reviews, people need to understand what the functions being called are doing anyway, so I see that as a feature 😄  In the grand scheme of things, less noise in the code and being "forced" to used meaningful var names is better.

Ok, let's go through this then:

- `source` is well, the shader source code—pretty self explanatory.
- `shader_info` is a `ShaderInfo` struct. Again, gotta be familiar with those to review this meaningfully 😄
- `maybe_shader_program` — it's obvious if you're familiar with OpenGL and our style of using the "maybe" prefix for optionals.

So I don't think `auto` is a problem in this snippet, but familiarity with the rest of the code, our style, and a bit of OpenGL is necessary, agreed.

shermp: You're also less likely to have use-after-free bugs in Java land... I've had to fix a bug in staging due to an `auto` refactor and use-after-free. (I think it was the loading soundfont bug)

I'm not against the use of auto (I use it myself), I'm just not a big fan of using it for everything.

Sorry, it's just something about this function that triggered me tonight.

shermp:
Apart from my complaints about `auto`, I think a shader cache is a great idea. It always seemed a bit silly to recompile shaders every time one switched to a different one.

johnnovak:
> Apart from my complaints about `auto`, I think a shader cache is a great idea. It always seemed a bit silly to recompile shaders every time one switched to a different one.

Thanks for the review @shermp. Yeah, it was just something that I did quickly 2 years ago to get the CRT auto-shaders out with the release. I guess it took so long to clean it up because it was not that terrible in practice, but of course, it's better to cache them 😄

src/gui/render/opengl_renderer.cpp
weirddan455: I hated `auto` at first too until I learned Rust where type inference is used everywhere by default.

IMO it's almost always a good idea to use `auto` to store the result of a function like he's doing here. I don't know how you got a use-after-free bug from `auto` but I can tell you, you can easily get an implicit cast where you don't want one if you **don't** use auto, especially after a refactor.

What I don't like is `auto i = 0;` In C++ this is always equivalent to `int i = 0;` so just use that. Rust looks ahead in the code and will make it a `u8` if you use it for arithmetic with another `u8`. C++ only looks after the equal sign in the single statement so `auto` is pretty pointless in this case and it's better to make the type explicit.

weirddan455:
> In general, this substantially decreases the likelihood for audio dropouts when switching shaders (as we're no longer potentially blocking the main thread for 5-30 ms!)

Just curious, did you do any performance measurements here? IIRC last time we discussed this, we were thinking this mostly only affects Macs because Windows and Linux should cache shaders at the driver level.

johnnovak:
> > In general, this substantially decreases the likelihood for audio dropouts when switching shaders (as we're no longer potentially blocking the main thread for 5-30 ms!)
> 
> Just curious, did you do any performance measurements here? IIRC last time we discussed this, we were thinking this mostly only affects Macs because Windows and Linux should cache shaders at the driver level.

Nope, but I don't need to measure it. We always loaded the shaders from disk, previously, did all the parsing, then handing the source over to the driver to compile it. I highly doubt that drivers implement some cache where the full shader source (the input) is the cache key 😄 

Update: Okay, I've done some reading. Apparently, it hashes the AST of the shader source and that's the cache key. Nifty. So whitespaces changes won't even matter. That helps with recompiling the same shaders over and over at startup, but what we're doing here is the most efficient way for the auto shader switching, and it really wasn't hard to do.

johnnovak:
@FeralChild64 @weirddan455 @shermp Can any of you smoke test it on Linux, please? Then merge time if no other comments, I guess.

johnnovak:
> Tested briefly on Linux, no problems found. No remarks regarding the source code.

Thanks for the testing and the review.

Issue #4553, opened 2025-10-18
Add support for Mac "document packages" (i.e. folder.dosbox) to open in the app.


eduo:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

macOS (Apple Silicon)

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Is your feature request related to a problem? Please describe.

Not a "problem" per se, but an inconvenience that could be solved easily: The inability **in macOS** to have clickable icons for specific set-ups (a folder with its own conf file and files, for example). Shell scripts work, but are cumbersome to create and maintain for users.

### Describe the solution you'd like

MacOS has a functionality similar to "bundles" called "[Document packages](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/DocumentPackages/DocumentPackages.html#//apple_ref/doc/uid/10000123i-CH106-SW1)" that allows a folder with a registered extension to open in an app and the app to interpret its contents. 

The folder with such an extension behaves like a file when double-clicked, and makes processing of complex documents in macOS easier.

This change would require for the app to register the document package extension in the mac app and when such a document package is opened when app would assume the document as the working directory and if there is a dosbox.conf file inside (or any other pre-defined name) then it would be loaded up.

This is currently the behaviour of iDOS and Boxer and is very useful.

### Describe alternatives you've considered

It used to be possible to place shell scripts in the same directory to make it easier for users to launch, but with Tahoe opening these became a nightmare.

Similarly, it used to be possible to put the files inside an .app folder and have a launch script named exactly like the parent folder, but same as above, in Tahoe this is almost impossible to do.

In Windows it's possible to drop files over the app to replicate some of the behaviour, but it's not fully possible.

Other tools that allowed to make clickable versions of game folders have stopped working in Tahoe due to security controls that don't allow apps to make other apps.

A Document Package was probably always the best option, but I can't see in the code or issues that it's been requested.

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
We don't have such a feature, but it's not a bad idea. However, at some point we'll have a cross-platform built-in GUI with a game browser, so I'm not too sure I wanna go down this path for macOS only, personally.

I'm still on Sequoia, so can't test this.

[...]

PR #4557, opened 2025-10-20
Handle .dosbox document packages on macOS (via SDL_DROPFILE)

eduo:
# Description

When a user double-clicks a .dosbox document package in Finder, macOS launches the application but delivers the document path via an SDL_DROPFILE event rather than through command-line arguments.

Event polling in GFX_Init() added immediately after SDL_Init() to catch these events. When a .dosbox package is detected, it's scanned for .conf files in the first two directory levels, then a new process is forked with the package expanded into --working-dir and --conf arguments. The parent process exits cleanly, allowing the child to run with standard command-line arguments.

The fork/exec is implemented directly rather than using the restart mechanism since DOSBOX_SetRestartFunction() has not been called yet at this point in initialization.

Two helper functions in platform/macos/package_handler check if a path is a .dosbox package and to scan packages for configuration files, converting all paths to absolute form to work regardless of the current working directory.

This completes .dosbox document package support on macOS, enabling users to organize games in package folders and launch them by double-clicking in Finder. This also makes it trivial to add folder dropping onto the DOSBox-Staging icon and have the app treat that folder as if it was a path passed to the command line (not included in this change, but .dosbox folder dropping is already supported)

New: src/platform/macos/package_handler.cpp and .h to handle processing the package.
Modified: meson.build to add the files above (I'm not familiar enough with Meson to know if this is strictly necessary)
Modified: extras/macos/info.plist.template to include the entries for the document package.
Modified: src/gui/sdl_gui.cpp handles the event and passes it for parsing relaunches if necessary.

## Related issues

- https://github.com/dosbox-staging/dosbox-staging/issues/4553

# Release notes

- Added support for .dosbox document packages in MacOS. Add ".dosbox" to a folder and it can be double-clicked. Folder will become the working directory and conf files within the package's first two nested levels will be read in alphabetical order.

# Manual testing

Tested .dosbox packages of various forms. Autoexec mount commands relative to .dosbox folder work. Dropping .dosbox folder onto icon opens them correctly.

CLI behaviour is not changed but "open folder.dosbox" will behave properly in MacOSX by interpreting the document package.

To test, add .dosbox to any folder. If it has .conf files, they would be interpreted upon opening. Otherwise, just the working dir is defined. No automounting should happen.

Tested in Tahoe, Big Sur, Sonoma.

All changes are made specific to MacOS via #IFDEF but tests have been made only on MacOS.

- [ ] Windows
- [X] macOS
- [ ] Linux

# Checklist

I have:

- [X] split my work into well-defined, bisectable commits, and I [named my commits well](https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/CONTRIBUTING.md#commit-messages).

### DIFF
--- a/src/hardware/audio/soundblaster.cpp
+++ b/src/hardware/audio/soundblaster.cpp
@@ -1956,11 +1956,13 @@ static void dsp_do_command()
 		if (sblaster->MaybeWakeUp()) {
 			// If we're waking up, then the DAC hasn't been running
 			// (or maybe wasn't running at all), so start with a
-			// fresh DAC state and ensure we're using per-frame
-			// callback timing.
+			// fresh DAC state.
 			sb.dac = {};
-			callback_type.SetPerFrame();
 		}
+
+		// Ensure we're using per-frame callback timing because DAC samples
+		// are sent one after another with sub-millisecond timing.
+		callback_type.SetPerFrame();
 
 		if (const auto dac_rate_hz = sb.dac.MeasureDacRateHz(); dac_rate_hz) {
 			sblaster->SetChannelRateHz(*dac_rate_hz);


src/gui/sdl_gui.cpp
```diff
@@ -1766,6 +1770,81 @@ void GFX_Init()
 	        SDL_GetCurrentVideoDriver(),
 	        SDL_GetCurrentAudioDriver());
 
+#ifdef MACOSX
+	// On macOS, check for document drop events (double-click to open .dosbox packages)
+	// When a user double-clicks a .dosbox package in Finder, macOS delivers the
+	// file path via SDL_DROPFILE event rather than command-line arguments.
+	// We check for this immediately after SDL_Init but before the restart function
+	// is registered, so we implement the restart directly here.
```
weirddan455: @johnnovak Why did you make `DOSBOX_Restart()` use a function pointer? There's only one restart function `dosbox_restart` and we set the function pointer exactly one time in init and never change it.

It would be better to use the existing restart function here. You should able to move the `DOSBOX_SetRestartFunction()` to before `GFX_Init()` or better yet do away this with this function pointer nonsense all together.

weirddan455:
You need to add the new files you created to the CMake build system. That's going to soon replace Meson.

johnnovak:
Very nice work @eduo 🎉 This is a clean and unobtrusive change, so we'll take it, of course 😎 I will test it today and will do the review.

But first of all, you'll need to add CMake support; without that it won't even build on CI. We just keep Meson around until 0.83.0 is out to help troubleshoot potential build system related subtle regressions, but it's deprecated.

src/gui/sdl_gui.cpp
johnnovak: Because of C++ compilation unit problems. Try removing the function pointer stuff and call it directly to see it for yourself 😄 From memory, that's because of our unit testing approach. One fn pointer is better than potentially reworking all that stuff, and I like the main / sdl_gui separation.

I don't think there's any other solution here than using a function pointer (while keeping main and sdl_gui separate), but you're welcome to try 😄 

> It would be better to use the existing restart function here.

I think we can just init the restart handler a bit earlier here, I'll play around with that later.

weirddan455: Ok, there's a linker error when building the tests if you do a forward declaration. Can we not just move the code for the restart function to `dosbox.cpp` though? I did that and it compiled and can PR it.

weirddan455: #4558

weirddan455: @eduo That's fixed and merged now. If you rebase this PR on `main` you can call `DOSBOX_Restart()` from here now. No more registering function pointers.

Grounded0:
✅

eduo:
> Very nice work @eduo 🎉 This is a clean and unobtrusive change, so we'll take it, of course 😎 I will test it today and will do the review.
> 
> But first of all, you'll need to add CMake support; without that it won't even build on CI. We just keep Meson around until 0.83.0 is out to help troubleshoot potential build system related subtle regressions, but it's deprecated.

Understood. I couldn't originally make CMake work but didn't mind because I was able to make Meson work. 

For CMake was thinking to make the changes following whatever is the current pattern since it should be simple enough (just add two files) but I realized the CMake build currently doesn't have a macOS platform subdirectory (unlike Windows which has src/platform/windows/CMakeLists.txt).

Is the expectation that we have a new src/platform/macos/CMakeLists.txt with target_sources(libdosboxcommon PRIVATE package_handler.cpp) where I add these files to the build? (like in Windows where GUI-specific sources include platform-specific code)

If so I would also need to modify src/CMakeLists.txt to add the new subdirectory which would mirror the existing Windows platform directory structure but I'm hesitant to as it seems a major decision for such a small change.

The alternative I guess is changing the main CMakeLists.txt with one conditional in target_sources but that feels... untidy?

To be fair, the latter is what I did for Meson. I added the .cpp file to dosbox_sources when on macOS in lines 1237-1240.

EDIT: Tidied the comment

johnnovak:
@eduo 

`src/platform/visualc/unistd.h` is a bit of an anomaly, let's not add to that.

We already have a pattern for macOS-only stuff, so please follow that. There's this section in the top-level CMakeList.txt:

```
# macOS
set(C_COREAUDIO      "${DOSBOX_PLATFORM_MACOS}")
set(C_COREMIDI       "${DOSBOX_PLATFORM_MACOS}")
set(C_COREFOUNDATION "${DOSBOX_PLATFORM_MACOS}")
set(C_CORESERVICES   "${DOSBOX_PLATFORM_MACOS}")
```

So pick one, `C_COREAUDIO` will do, and follow the bouncing ball 😎  Check out `src/midi/midi_coreaudio.cpp` and `src/midi/midi_coreaudio.h`, etc., and do the same thing—should be pretty straightforward. 

Thanks for that 😄 I'll test this once you've added CMake support and it builds on CI. I lost the ability to compile with Meson on my Mac Mini 6+ months ago 😆 (I use a Linux VM to periodically check whether the Meson build still works 😆 )

eduo:
> @eduo
> 
> We already have a pattern for macOS-only stuff, so please follow that. There's this section in the top-level CMakeList.txt:
> 
> ```
> # macOS
> set(C_COREAUDIO      "${DOSBOX_PLATFORM_MACOS}")
> set(C_COREMIDI       "${DOSBOX_PLATFORM_MACOS}")
> set(C_COREFOUNDATION "${DOSBOX_PLATFORM_MACOS}")
> set(C_CORESERVICES   "${DOSBOX_PLATFORM_MACOS}")
> ```
> 
> So pick one, `C_COREAUDIO` will do, and do the same thing. Check out `src/midi/midi_coreaudio.cpp` and `src/midi/midi_coreaudio.h`, etc, should be straightforward.
> 
> Thanks 😄 I'll test this once you're done with this, I lost the ability to compile with Meson on my Mac Mini 6+ months ago 😆 (I use a Linux VM to periodically check whether the Meson build still works 😆 )

Oh! I missed this. How dumb :D I went straight for the hatchet. I'm sorry!

johnnovak:
> Oh! I missed this. How dumb :D I went straight for the hatchet. I'm sorry!

No problem 😄 It's a big and complex codebase.

extras/macos/Info.plist.template
```diff
@@ -52,5 +52,47 @@
...
+			<array>
+				<string>dosbox</string>
+			</array>
+			<key>CFBundleTypeName</key>
+			<string>DOSBox Game Package</string>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>LSItemContentTypes</key>
+			<array>
+				<string>io.github.dosbox-staging.dosbox-game</string>
```
johnnovak: Namespace should be `org.dosbox-staging`

And `dosbox-package`, so `org.dosbox-staging.dosbox-package`.

extras/macos/Info.plist.template
```diff
@@ -52,5 +52,47 @@
...
+		<dict>
+			<key>UTTypeConformsTo</key>
+			<array>
+				<string>com.apple.package</string>
+				<string>public.directory</string>
+			</array>
+			<key>UTTypeDescription</key>
+			<string>DOSBox Game Package</string>
+			<key>UTTypeIdentifier</key>
+			<string>io.github.dosbox-staging.dosbox-game</string>
```
johnnovak: Namespace should be `org.dosbox-staging`, so `org.dosbox-staging.dosbox-package`.

extras/macos/Info.plist.template
```diff
@@ -52,5 +52,47 @@
...
+	
+	<key>CFBundleDocumentTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeExtensions</key>
+			<array>
+				<string>dosbox</string>
+			</array>
+			<key>CFBundleTypeName</key>
+			<string>DOSBox Game Package</string>
```
johnnovak: "DOSBox Staging Package" perhaps? People use Staging for running demos and apps too (although gaming is our first priority, indeed 😄).

src/gui/sdl_gui.cpp
```diff
@@ -1766,6 +1770,81 @@ void GFX_Init()
...
+	SDL_Event event;
+	while (SDL_PollEvent(&event)) {
```
johnnovak: Can we start polling immediately after startup? Isn't it possible for a race condition to happen here, don't we need to sleep "a bit" in the 50-200ms range perhaps before polling? Just wondering, I don't know 😄

src/gui/sdl_gui.cpp
```diff
@@ -1766,6 +1770,81 @@ void GFX_Init()
...
+		if (event.type == SDL_DROPFILE && event.drop.file != nullptr) {
+			const std::string dropped_file_path = event.drop.file;
+			SDL_free(event.drop.file);
+
+			LOG_MSG("MACOS: Received dropped file via SDL_DROPFILE: '%s'",
+			        dropped_file_path.c_str());
+
+			// Check if it's a .dosbox package
+			if (macos::IsDosBoxPackage(dropped_file_path)) {
+				LOG_MSG("MACOS: Detected .dosbox package, scanning for configuration files");
```
johnnovak: Use the `CONFIG: ` log prefix please.

src/gui/sdl_gui.cpp
```diff
@@ -1766,6 +1770,81 @@ void GFX_Init()
...
+
+				// Scan the package for .conf files and get absolute paths
+				auto package = macos::ScanDosBoxPackage(dropped_file_path);
+
+				if (!package.is_valid) {
+					LOG_ERR("MACOS: Failed to scan .dosbox package");
```
johnnovak: Use the `CONFIG: ` log prefix please.

Maybe just use LOG_WARN here; we reserve LOG_ERR for _really_ serious stuff, bad user input is always just a warning and this can be considered as such.

src/gui/sdl_gui.cpp
```diff
@@ -1766,6 +1770,81 @@ void GFX_Init()
...
+					continue;
+				}
+
+				LOG_MSG("MACOS: Found %zu configuration file(s)", package.conf_files.size());
```
johnnovak: Use the `CONFIG: ` log prefix please for all the remaining log statements 😄

src/gui/sdl_gui.cpp
```diff
@@ -1766,6 +1770,81 @@ void GFX_Init()
...
+				} else if (pid > 0) {
+					// Parent process: exit cleanly to let child run
+					LOG_MSG("MACOS: Forked child process (PID %d), exiting parent", pid);
+					exit(0);
+				} else {
+					// Fork failed
+					LOG_ERR("MACOS: Fork failed: %s", strerror(errno));
+				}
+			} else {
+				LOG_MSG("MACOS: Dropped file is not a .dosbox package, ignoring");
```
johnnovak: LOG_WARN ?

src/platform/macos/package_handler.cpp
```diff
@@ -0,0 +1,106 @@
+// SPDX-FileCopyrightText:  2025 The DOSBox Staging Team
```
johnnovak: Please use 2025-2025. We always use a range; it makes the year updater script a bit simpler.

src/platform/macos/package_handler.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <filesystem>
+
+#include "utils/fs_utils.h"
+#include "utils/string_utils.h"
+
+namespace macos {
+
+namespace fs = std::filesystem;
+
+bool IsDosBoxPackage(const std::string& path)
```
johnnovak: Let's not use a namespace here.

`is_macos_dosbox_package()` and `scan_macos_dosbox_package()` will do.

src/platform/macos/package_handler.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include <algorithm>
+#include <filesystem>
+
+#include "utils/fs_utils.h"
+#include "utils/string_utils.h"
+
+namespace macos {
+
+namespace fs = std::filesystem;
```
johnnovak: We don't do this; please use the fully qualified names everywhere in the file.

src/platform/macos/package_handler.cpp
```diff
@@ -0,0 +1,106 @@
...
+			const auto extension = file_path.extension().string();
+
+			if (iequals(extension, ".conf")) {
+				result.conf_files.push_back(file_path.string());
+			}
+		}
+	}
+
+	// Sort conf files alphabetically for consistent ordering
+	std::sort(result.conf_files.begin(), result.conf_files.end());
```
johnnovak: Probably better to use `natural_compare` from `utils/string_utils.h`?

src/platform/macos/package_handler.cpp
```diff
@@ -0,0 +1,106 @@
...
+		const auto file_path = entry.path();
+		const auto extension = file_path.extension().string();
+
+		if (iequals(extension, ".conf")) {
+			result.conf_files.push_back(file_path.string());
+		}
+	}
+
+	// Scan second level (subdirectories)
+	for (const auto& level1_entry : fs::directory_iterator(pkg_path, ec)) {
```
johnnovak: I'm not too sure supporting two levels is a good idea. It just complicates the feature, and I'm not seeing the use cases.

Could you illustrate through a few example why you think we should do this?

src/platform/macos/package_handler.h
```diff
@@ -0,0 +1,29 @@
...
+#define DOSBOX_PLATFORM_MACOS_PACKAGE_HANDLER_H
+
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace macos {
+
+// Represents the expanded form of a .dosbox package
+struct DosBoxPackage {
```
johnnovak: `MacOsDosBoxPackage`?

We use namspaces in more recent code (e.g., for CLAP plugin support), but I don't really like doing it for platform-specific stuff.

extras/macos/Info.plist.template
```diff
@@ -52,5 +52,47 @@
 	<key>x86_64</key>
 	<string>10.10.0</string>
 	</dict>
+	
+	<key>CFBundleDocumentTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeExtensions</key>
+			<array>
+				<string>dosbox</string>
```
johnnovak: `dosbox-package` perhaps? But I'm not sure of the extension naming conventions, maybe just `dosbox` is enough.

src/gui/sdl_gui.cpp
eduo: I did indeed wait a bit before, but with all my tests it didn't seem to be necessary so I removed it. It may be safer to do it, as it's not expensive.

extras/macos/Info.plist.template
eduo: Makes sense. Even if the extension is eventually shared, common practice is to have application-specific descriptions for common formats anyway.

extras/macos/Info.plist.template
```diff
@@ -52,5 +52,47 @@
...
+	<key>UTExportedTypeDeclarations</key>
+	<array>
+		<dict>
+			<key>UTTypeConformsTo</key>
+			<array>
+				<string>com.apple.package</string>
+				<string>public.directory</string>
+			</array>
+			<key>UTTypeDescription</key>
+			<string>DOSBox Game Package</string>
```
johnnovak: DOSBox Staging Package?

extras/macos/Info.plist.template
eduo: My mistake. I am so used to launching in the terminal using open -nb io.github.dosbox-staging that I didn't think twice about it.

src/gui/sdl_gui.cpp
eduo: Good catch. I used this to isolate in the console and forgot to change it back.

src/platform/macos/package_handler.cpp
eduo: I added two levels for a specific scenario I've seen often: eXoDOS. 

Since all their conf files refer to ./exodos/gamename but the conf itself is inside the gamename folder, games originally set-up for it tend to be "Full Game Name (year)/eXoDOS/ShortName" as paths break otherwise. 

Supporting the second level seemed the easiest solution for those cases.

Alternatively, the "spec" for .dosbox packages can be "conf files must be in the root directory"

src/gui/sdl_gui.cpp
johnnovak: I vaguely remember I had to do that in one of my glfw apps on macOS that handle some window system events at startup, so yeah, probably it's worthwhile to add a bit of delay back.

src/platform/macos/package_handler.cpp
johnnovak: Yeah, I'm not entirely sure I follow. I think in practice people will just need to massage the files a bit because the way eXoDOS is set up is a complicated maze of relative paths (but it works for them, hey, with their specific constraints). Supporting their very specific directory layout is way out of scope for us, and it would be a bit too much.

For example, using the full eXoDOS collection (maybe you were using some older revision?):

image

So in this case this is what I'd do:

1. Create `Azrael's Tear.dosbox` folder
2. Copy the CD images from `x:\eXoDOS\eXo\eXoDOS\AZRAEL\cd` into `Azrael's Tear.dosbox/drives/d`
3. Copy the rest of the contents of `x:\eXoDOS\eXo\eXoDOS\AZRAEL\` into `Azrael's Tear.dosbox/drives/c`
4. Copy `dosbox.conf` from `x:\eXoDOS\eXo\eXoDOS\!dos\AZRAEL\` into `Azreal's Tear.dosbox`
5. Massage `dosbox.conf` a bit 😆 (for a start, remove `@mount c .\eXoDOS\` and `imgmount d .\eXoDOS\AZRAEL\cd\azreal.cue -t cdrom`)

So let's just support the simple case where we only look for `dosbox.conf` in the document package root.

extras/macos/Info.plist.template
Grounded0: `.dosbox` seems good.

extras/macos/Info.plist.template
Grounded0: "DOSBox Staging Package" seems good.

extras/macos/Info.plist.template
Grounded0: "DOSBox Staging Package" seems good.

src/platform/macos/package_handler.cpp
eduo: I will change it then to the package root and to only process "dosbox.conf" rather than multiple conf in alphabetical order.

eduo: Cool. This approach works because even if you want to do vanilla dosbox package, it's still possible and clean by having the dosbox.conf in the package root and tweaking the paths and everything would still be self-contained. I was probably overcomplicating because the most common use I've seen is not from standard exodos installs but from manual ones and it may have just been some opinionated conversion mechanism.

src/platform/macos/package_handler.cpp
eduo: Will remove as we'll only process dosbox.conf

eduo:
With the changes to dosbox_start and the decision to do a single dosbox.conf in root, the change ended up being pretty small.

I hope it's also properly done, now.

weirddan455:
You didn't rebase correctly. You've somehow included my commit as part of this PR. My commit is already in main so it needs to be removed and then rebase the rest on top of main.

eduo:
> You didn't rebase correctly. You've somehow included my commit as part of this PR. My commit is already in main so it needs to be removed and then rebase the rest on top of main.

I'm sorry. I've force-pushed the proper commit now. Only my changes show up.

src/gui/sdl_gui.cpp
```diff
@@ -1784,6 +1786,60 @@ void GFX_Init()
...
+
+				// Use the standard restart mechanism to launch with expanded package
+				LOG_MSG("CONFIG: Restarting with expanded package arguments");
+				DOSBOX_Restart(new_params);
+			} else {
+				LOG_WARNING("CONFIG: Dropped file is not a .dosbox package, ignoring");
+			}
+		}
+	}
+#endif
```
johnnovak: Can you extract this whole thing into a static function please? `handle_macos_dosbox_package()` will do.

src/platform/macos/package_handler.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef DOSBOX_PLATFORM_MACOS_PACKAGE_HANDLER_H
+#define DOSBOX_PLATFORM_MACOS_PACKAGE_HANDLER_H
+
+#include <optional>
+#include <string>
+#include <vector>
+
+// Represents the expanded form of a .dosbox package
+struct dosbox_macos_package {
```
johnnovak: We use UpperCamelCase for struct names in new code (but this will go away anyway, as per my later comments).

src/gui/sdl_gui.cpp
```diff
@@ -1784,6 +1786,60 @@ void GFX_Init()
...
+
+				// Add --working-dir to set the game directory
+				new_params.push_back("--working-dir");
+				new_params.push_back(package.package_path);
+
+				// Add --conf for each configuration file found
+				for (const auto& conf_file : package.conf_files) {
+					new_params.push_back("--conf");
+					new_params.push_back(conf_file);
+				}
```
johnnovak: Okay, so let's simplify things: 

- We're only checking for `dosbox.conf` at the root level in the package now.
- So we don't need to provide the `--conf` settings, we just need to set `--working-dir` and that's it. `dosbox.conf` will be then picked up automatically, no need for the `--conf` arg.
- We only need `is_macos_dosbox_package()` now because things are a lot simpler (`scan_macos_dosbox_package()` can be merged into `is_macos_dosbox_package()`).
- The `dosbox_macos_package` will fall away too; we just need to pass the `std_fs::path` to the package around.

src/platform/macos/package_handler.h
```diff
@@ -0,0 +1,25 @@
...
+	std::string package_path = {};           // Full path to .dosbox package
+	std::vector<std::string> conf_files = {}; // Paths to .conf files found
+	bool is_valid = false;
+};
+
+// Check if a path is a .dosbox document package
+bool is_macos_dosbox_package(const std::string& path);
+
+// Scan a .dosbox package for .conf files (first 2 directory levels)
```
johnnovak: > (first 2 directory levels)

No longer needed, but as previously noted, we only really need `is_macos_dosbox_package()` now anyway.

johnnovak: Actually, I think this whole `src/platform/macos/package_handler.h` is not needed. The complication is not really warranted for a single `is_macos_dosbox_package()` function. Let's just put that into `sdl_gui.cpp` and call it a day.

I mentioned before `src/platform/windows/visualc` is just an outlier. For everything else things just to in the regular source tree and are ifdef'd in and out per platform. We don't want to pull platform-specific things into their own subtree in the source tree.

johnnovak:
@eduo Left further comments. After the significant simplifications we've done here, everything becomes a lot simpler, so could you please follow up on all that and remove the remnants of the more complex initial solution?

Can you please also merge your three commits into one and adjust the commit message to reflect the final solution?

After that, I think we're done 😄 Thanks for the patience, this will be a great enhancement for all the macOS folks out there! 🥳 🎉

eduo:
> @eduo Left further comments. After the significant simplifications we've done here, everything becomes a lot simpler, so could you please follow up on all that and remove the remnants of the more complex initial solution?
> 
> Can you please also merge your three commits into one and adjust the commit message to reflect the final solution?
> 
> After that, I think we're done 😄 Thanks for the patience, this will be a great enhancement for all the macOS folks out there! 🥳 🎉

Sure. I will do it all this afternoon.

eduo:
I have pushed the latest change here, but it may make more sense to do a separate PR with just the changes? The final size of the full change is barely over a hundred lines and half of it is boilerplate XML.

src/gui/sdl_gui.cpp
```diff
@@ -1753,6 +1754,62 @@ static void add_default_sdl_section_mapper_bindings()
...
+	std_fs::path pkg_path = std_fs::absolute(dropped_file_path, ec);
+	if (ec) {
+		LOG_WARNING("CONFIG: Failed to convert package path to absolute");
+		return;
+	}
+
+	// Build new startup parameters with expanded package
+	auto new_params = control->startup_params;
+
+	// Add --working-dir to set the game directory
```
johnnovak: Minor: package directory

src/gui/sdl_gui.cpp
```diff
@@ -1784,6 +1841,20 @@ void GFX_Init()
 	        SDL_GetCurrentVideoDriver(),
 	        SDL_GetCurrentAudioDriver());
 
+#ifdef MACOSX
+	// Check for .dosbox document packages dropped from Finder
+	// (double-click to open or drag-and-drop onto the app icon)
+	SDL_Delay(100);  // Brief delay to allow OS to queue the event
```
johnnovak: Minor: we try not to use end of line comments like this

johnnovak:
> I have pushed the latest change here, but it may make more sense to do a separate PR with just the changes? The final size of the full change is barely over a hundred lines and half of it is boilerplate XML.

Thanks, it's looking great now. I will test it shortly. I left two further very minor comments about comments 😄 

I don't see why we would want to raise a new PR? We always discuss things and address comments in the original PR. In some cases we have over a 100 comments in big PRs 😅 

Btw, it's great the final solution is fairly minimal 🎉

eduo:
> > I have pushed the latest change here, but it may make more sense to do a separate PR with just the changes? The final size of the full change is barely over a hundred lines and half of it is boilerplate XML.
> 
> Thanks, it's looking great now. I will test it shortly. I left two further very minor comments about comments 😄
> 
> I don't see why we would want to raise a new PR? We always discuss things and address comments in the original PR. In some cases we have over a 100 comments in big PRs 😅
> 
> Btw, it's great the final solution is fairly minimal 🎉

Thanks for your patience. I'm a long-time coder but while I've been using git for a while I always did so for my personal use. In the job we use other, more esoteric versioning platforms. I'm trying not to screw up the PR nor take too much time from you guys.

weirddan455:
> Thanks for your patience. I'm a long-time coder but while I've been using git for a while I always did so for my personal use. In the job we use other, more esoteric versioning platforms. I'm trying not to screw up the PR nor take too much time from you guys.

Not to sound impatient but you did somehow get a bunch of commits from `main` stuck into this PR again.

I do appreciate the contribution though :smile:

eduo:
> > Thanks for your patience. I'm a long-time coder but while I've been using git for a while I always did so for my personal use. In the job we use other, more esoteric versioning platforms. I'm trying not to screw up the PR nor take too much time from you guys.
> 
> Not to sound impatient but you did somehow get a bunch of commits from `main` stuck into this PR again.
> 
> I do appreciate the contribution though 😄

I realized as soon as it happened. And I had just written a comment about trying not to screw it up.

eduo:
I think this is it.

johnnovak:
> I think this is it.

Sorry, it's still not good. Some earlier changes from `main` showed up somehow, and remnants of your 2-level lookup stuff returned in a few places. Just have a look in the unified diff when clicking the "Files changed" tab up top and you'll see.

> Thanks for your patience. I'm a long-time coder but while I've been using git for a while I always did so for my personal use. In the job we use other, more esoteric versioning platforms. I'm trying not to screw up the PR nor take too much time from you guys.

We thank you for your perseverance, and the work you've done here is very good! We welcome you to continue contributing to Staging, we need more experienced devs in the project. The issue tracker is full of work to do; maybe you'll find something you're interested in working on in the future.

So yeah, I could see you're not new to coding at all, but Git and the Git workflows can take some time to get used to, no doubt. Talking from personal experience! 😆 I don't want to overload you and force you to become a Git expert in the duration of a single PR 😅, so yeah, it's fine to raise a new clean PR this time if that's easier for you.

I've shared some tips for getting up to speed with our Git workflow below, but that's more for the future if you've had enough of Git this time around 😄 

---

We use the "rebase workflow"; the idea is that your changes from the feature branch (PR branch) should be cleanly applied on top of `main` without a merge commit. So in `main` we have a linear history—no merge commits!

In principle, what you should do here is rebase the commits in your branch into a single commit that can be applied on top of `main` after "merging" (we still call it a merge, and GitHub has a "merge" button, but the merging will just result in more commits added to `main` in the linear history).

These tutorials are good introductions to the rebase workflow and using the `git rebase` command to massage your history in PR branches:

https://www.atlassian.com/git/tutorials/merging-vs-rebasing
https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase

eduo:
Tried one last time. Didn't make it. Will do a new PR. I'm awfully sorry. It doesn't help the timezone differences mean every time I do this there's a dozen new commits :D

johnnovak:
> Tried one last time. Didn't make it. Will do a new PR. I'm awfully sorry. It doesn't help the timezone differences mean every time I do this there's a dozen new commits :D

It's all good, we're not going anywhere 😄

weirddan455:
FYI you don’t have to always rebase when there’s new commits to main. GitHub will handle that automatically unless there’s merge conflicts.

PR #4564, opened 2025-10-22
Adjust VMware interface behavior to match VMware Workstation Player

FeralChild64:
# Description

It seems there exists a "demo" version of the VMware - VMware Workstation Player. It's discontinued, but at least the last version can still be installed on Windows 11.

I have written some test software to check whether my implementation of the VMware mouse interface behaves mostly the same way as the original hypervisor - but I have found some important differences:
- return code `0xffffffff` is used to signal an unimplemented command (but not an unimplemented mouse subcommand!)
- the number of bytes in a buffer (returned via Intel 8042 I/O port) is at most 16-bit; the upper 16-bits look like some kind of command acknowledgement or status code

I have adapted my implementation to match the VMware behavior - this should lower the risk of updated driver not working anymore under DOSBox Staging.

# Manual testing

Using the test tools below checked that the results are similar to the VMware Workstation Player results. Some differences are allowed:
- I believe the values after PS/2 interface disable command (0x000000f5) are garbage
- PS/2 interface events data definitely is going to be different, it's impossible to precisely repeat the mouse position

[test_tools.tar.gz](https://github.com/user-attachments/files/23059213/test_tools.tar.gz)

Results1

Results2

Regression testing:
- checked that the official Windows 9x VMware mouse driver still works
- checked that the unofficial Windows 3.11 VMware mouse driver (https://github.com/NattyNarwhal/vmwmouse) still works
- checked that the unofficial DOS VMware mouse driver (https://git.javispedro.com/cgit/vbados.git/about/) still works (set `builtin_dos_mouse_driver = off` and `virtualbox_mouse = off` in the config file to use the driver with the VMware interface)

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/hardware/vmware.cpp
+++ b/src/hardware/vmware.cpp
@@ -35,6 +35,14 @@ static std::string program_segment_name = {};
 // Queued data, waiting to be fectehd by the guest side driver
 static std::vector<uint32_t> abs_pointer_queue = {};
 
+enum class QueueContent {
+	None,
+	CommandResponse,
+	MouseStatus,
+};
+
+static QueueContent abs_pointer_queue_content = QueueContent::None;
+
 // ***************************************************************************
 // Various common constants and type definitions
 // ***************************************************************************
@@ -48,6 +56,13 @@ static std::set<std::string> SegmentBlackList = {
 
 // Magic number for all VMware calls
 static constexpr uint32_t VmWareMagic = 0x564d5868u;
+
+// Not 100% sure if the meaning is correct, values were deducted by observing
+// the VMware Workstation Player behavior
+static constexpr uint32_t VmWareUnimplemented = 0xffffffffu;
+static constexpr uint32_t VmWareAcknowledged  = 0x01u;
+
+static constexpr size_t VmWareMouseStatusSize = 4;
 
 // The exact meaning of the version ID below is unknown - so far we know that:
 // - Linux kernel requires precisely this particular version ID, otherwise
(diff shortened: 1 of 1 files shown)


shermp:
VMWare workstation player was discontinued because Workstation Pro is now free. Of course, you have to create an account with Broadcom to download it might not be worth the bother.

johnnovak:
Yeah, so taking a step back, all this virtualiser mouse protocol business is only for supporting seamless mouse mode in Win3.x and maybe Win9x down the road in Windowed mode, right?

That's a hell of a lot of code for one niche feature 😆

So I'm ok with whatever, but are you sure you wanna make adjustments based on reverse engineering discountinued and unmaintained products? How can you be sure this wasn't just a bug in VMware Player that they fixed later?

> I have adapted my implementation to match the VMware behavior - this should lower the risk of updated driver not working anymore under DOSBox Staging.

I don't get it. You made these adjustments for the discontinued version, no? Why not only make sure your stuff is conformant to the behaviour of the latest officially supported VMware release? Then if there's a bug in the 3rd party driver, get the maintainer of the driver to fix it?

FeralChild64:
I have used the latest version easily available to me - if I see correctly, _VMware Workstation Player_ was discontinued in 2024, so it's not terribly outdated yet.

It seems the registration is not enough to get the new version mentioned by @shermp - according to [Broadcom](https://knowledge.broadcom.com/external/article/368734/download-desktop-hypervisor-workstation.html):

> After registering a basic user profile on the Broadcom Support Portal,  your profile will need to be screened and approved by our Internal Export Compliance team before you are allowed to download the free tier of available products (The profile screening  and approval process  can take  several weeks to complete and is completely dependent on  the number of profiles that are in  queue for screening,).

And a bug can as well appear in the latest version, there is no rule - if the VMware hypervisor behavior is more-or-less sane and the official drivers continue working properly, then without some official interface documentation we won't be able to decide whether the observed behavior is a bug or not.

johnnovak (approved):
Looks alright

PR #4570, opened 2025-10-24
Clean up config handling code

johnnovak:
# Description

Just a minor follow up to my recent refactoring jobs.

The most significant thing is removing the ability to override config settings via shell env variables. I honestly don't see why keeping that would be a good idea when we have `--set` already. 

# Manual testing

Start up a few games (the changes are trivial).

- [ ] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -623,9 +623,9 @@ int main(int argc, char* argv[])
 		// executable dir).
 		//
 		// TODO Consider forcing portable mode in secure mode (this
-		// could be accomplished by passing a flag to InitConfigDir);.
+		// could be accomplished by passing a flag to init_config_dir);.
 		//
-		InitConfigDir();
+		init_config_dir();
 
 		// Register essential DOS messages needed by some command line
 		// switches and during startup or reboot.
--- a/src/misc/cross.cpp
+++ b/src/misc/cross.cpp
@@ -183,7 +183,7 @@ static std_fs::path get_or_create_config_dir()
 
 static std_fs::path cached_config_dir = {};
 
-void InitConfigDir()
+void init_config_dir()
 {
 	if (cached_config_dir.empty()) {
 		// Check if a portable layout exists
--- a/src/misc/cross.h
+++ b/src/misc/cross.h
@@ -92,7 +92,7 @@ constexpr auto localtime_r = ::localtime_r;
 // Create or determine the location of the config directory (e.g., in portable
 // mode, the config directory is the executable dir). Must be called before
 // calling get_config_dir().
-void InitConfigDir();
+void init_config_dir();
 
 std_fs::path get_config_dir();
 std::string get_primary_config_name();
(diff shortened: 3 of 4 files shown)


weirddan455:
You mean removing the shorthand where you can type for example `cpu_cycles=6000` at the command prompt? This is the main way that I've been adjusting settings at runtime and I thought you were the one that added this feature in the first place. Removing this is likely to break people's scripts and `[autoexec]`.

johnnovak:
> You mean removing the shorthand where you can type for example `cpu_cycles=6000` at the command prompt? This is the main way that I've been adjusting settings at runtime and I thought you were the one that added this feature in the first place. Removing this is likely to break people's scripts and `[autoexec]`.

No way! I'm crazy but not stupid 😎 

Check the relevant commit message and the diff, please.

https://github.com/dosbox-staging/dosbox-staging/pull/4570/commits/6b2e2e91a7094434e9ad1cb65c0934b0ec49ccd1

weirddan455:
I did read that commit message but I must have misunderstood. Now I'm looking at the code and is this looking at environment variables on the host? If so, I had no idea we had this feature in the first place so I don't have concerns about removing it.

johnnovak:
> Now I'm looking at the code and is this looking at environment variables on the host?

Yeah, so you could do `export DOSBOX_CPU_CPU_CYCLES=6000` to override `cpu_cycles` in the `[cpu]` cycles from the shell 😐 

> If so, I had no idea we had this feature in the first place so I don't have concerns about removing it.

Me neither 🤷🏻 Seemed like an easy decision to lose it 😆

src/shell/command_line.cpp
```diff
@@ -350,12 +350,7 @@ uint16_t CommandLine::Get_arglength()
 		total_length += cmd.size() + 1;
 	}
 
-	if (total_length > UINT16_MAX) {
-		LOG_WARNING("CONFIG: Command line length too long, truncating");
-		total_length = UINT16_MAX;
-	}
-
-	return static_cast<uint16_t>(total_length);
+	return total_length;
```
shermp, line 353: Was there any reason why total_length was limited to `UINT16_MAX` previously?

It might be good to explicitly cast the return type from `size_t` to `int`.

johnnovak: There wasn't.

Yeah, I don't care about `size_t` (maybe just to make some warnings go away in some cases). If someone will provide more than 4,294,967,295 command line arguments and make this code "misbehave", they deserve the fruits of their efforts 😆 Job well done! 😂 

Jokes aside, there's a check somewhere downstream that limits the range to 100. `int` is fine here (like it's fine for 99.999% of general-purpose code dealing with numbers).

shermp: Yeah, I was mostly worried about any compiler warnings for implicit narrowing or sign conversion on return.

johnnovak: I've just checked PVS-Studio and there are new warnings for this, so marking as resolved. I think that's because the entire `uint16_t` range fits within the signed `int32_t` range.

johnnovak:
@weirddan455 @shermp We're good? Tick please?

johnnovak:
> > The most significant thing is removing the ability to override config settings via shell env variables.
> 
> Shouldn't we put this into release notes? Otherwise, the change looks OK to me.

Obscure feature probably no one knows about, so I guess not?

PR #4572, opened 2025-10-25
Add `soundcanvas_rom_dir` config setting and environment variable helpers

### LINKED ISSUE #4440: Add `soundcanvas_rom_dir` config setting
Add `soundcanvas_rom_dir` config setting that in turn sets the `SOUNDCANVAS_ROM_DIR` env var to the chosen path.

The SoundCanvas CLAP plugin we happen to load might use this env var as a hint to help locate the ROM files.

Or it might not.

Who knows, right? 😏 Life is one big mystery 😎

shermp:
# Description

Adds a `soundcanvas_rom_dir` setting to the `[soundcanvas]` config section. Dosbox uses this (or the default) to set the `SOUNDCANVAS_ROM_PATH` environment variable which may be used by a sound canvas plugin.

To do this, I refactored the rom directory listing from the mt32 code into common functions that can be used by any code that needs to find roms.

I also added cross platform functions to get and set environment variables, and moved most of our uses of getenv/setenv to use these.

Finally, I removed support for resolving `~username` style home directory resolving. I've personally never seen it in the wild, although if there is strong objection from the linux users here, I can drop that commit.

## Related issues

Closes #4440 

# Release notes

You can set `soundcanvas_rom_dir` in the `[soundcanvas]` config section to hint to a sound canvas plugin where to find ROM files.

# Manual testing

- [x] Windows
- [ ] macOS
- [x] Linux

Tested placing ROM set in the following situations:
- No config option set: ROM's loaded from `<config_dir>/soundcanvas-roms` if exists
- Absolute directory set: ROM's loaded from specified directory if exists
- Relative directory set: ROM's loaded from working directory if exists

# Checklist

I have:

### DIFF
--- a/src/midi/soundcanvas.cpp
+++ b/src/midi/soundcanvas.cpp
@@ -774,6 +774,7 @@ void SOUNDCANVAS_ListDevices(MidiDeviceSoundCanvas* device, Program* caller)
 	constexpr auto Indent      = "  ";
 
 	if (!available_models_initialised) {
+		available_models.clear();
 		for (auto m : all_models) {
 			if (const auto p = try_load_plugin(*m); p.plugin) {
 				available_models.insert(m);
@@ -928,6 +929,10 @@ static void register_soundcanvas_text_messages()
 static void notify_soundcanvas_setting_updated([[maybe_unused]] SectionProp& section,
                                                const std::string& prop_name)
 {
+	if (prop_name == "soundcanvas_rom_dir") {
+		available_models_initialised = false;
+	}
+
 	const auto device = dynamic_cast<MidiDeviceSoundCanvas*>(
 	        MIDI_GetCurrentDevice());
 


src/utils/env_utils.cpp
```diff
@@ -0,0 +1,54 @@
+// SPDX-FileCopyrightText:  2022-2025 The DOSBox Staging Team
+// SPDX-License-Identifier: GPL-2.0-or-later
```
johnnovak, line 2: You put the header file in `src/utils` but the source file is `src/misc`. Both should be in `src/utils`.

shermp: I was copying how `fs_utils` was handled. But sure, I can move it.

johnnovak: Yeah, then fs_utils need to be fixed up too; belongs to utils.

src/midi/mt32.cpp
```diff
@@ -31,9 +31,14 @@
 #include "utils/fs_utils.h"
 #include "utils/math_utils.h"
 #include "utils/string_utils.h"
+#include "utils/rom_dir.h"
 
 // #define DEBUG_MT32
 
+// MT-32 ROM locations
+constexpr auto mt32_platform_rom_dir_name = "mt32-rom-data";
+constexpr auto mt32_platform_rom_dir_name_macos = "MT32-Roms";
```
johnnovak: Sorry man, I'm afraid you didn't get the generalisation quite right here 😄 

We want to let these paths evolve independently of each other per sound device and platform, so we should not couple them like this. This overgeneralisation just adds complexity and unnecessary coupling which is the no.1 enemy of evolving software. Getting rid of _every single_ "duplication" is not a goal. Only things that "must change together, 100% of the time" are good candidates for generalisation.

Please just leave `mt32.cpp` alone and copy the whole `get_platform_rom_dirs()` and `get_rom_dirs()` thing with the ifdefs over to `soundcanvas.cpp` and adjust the paths accordingly, that's the best way to go. 

If you want to re-use something, the Linux variant of `get_platform_rom_dirs()` could be extracted and generalised to `get_linux_platform_data_dirs()`, then you'd call that from `mt32.cpp` and `soundcanvas.cpp` to construct the lookup path list.

Probably that's a good idea because the XDG lookup thing should be reusable and consistent whenever we look up "resource data" paths on Linux. But for Windows and macOS, the rules are simple so we'll just keep using a list of paths ifdef'd per platform—that's the simplest thing to do (in fact, the `/usr/*` paths should be removed from the macOS lookups; nobody puts MT-32 ROMs there 😆)

Good job spotting and removing this, btw—it's no longer needed because we're dealing with proper `std_fs::path`s.

```cpp
	if (selected_romdir.back() != '/' && selected_romdir.back() != '\\') {
		selected_romdir += CROSS_FILESPLIT;
	}
```

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +284,60 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
...
+			}
+			LOG_MSG("SOUNRCANVAS: Setting SOUNDCANVAS_ROM_DIR env variable to '%s'", canonical_rom_dir.string().c_str());
+			set_env_var("SOUNDCANVAS_ROM_DIR", canonical_rom_dir.string().c_str());
+			rom_dir_found = true;
+			break;
+		}
+	}
+	if (!rom_dir_found) {
+		LOG_MSG("SOUNRCANVAS: %s", "ROM dir not found, using plugin defaults");
+	}
```
johnnovak: Sorry again, let's not do this.

We don't want to "leak knowledge" about the CLAP plugin's internals into the Staging codebase. Just set the env var, init the CLAP plugin, and report the error in the Staging logs if initialising the plugin failed. That's all we need to do here!

We don't wanna detect anything, we're just giving a hint to the CLAP plugin via the env var. We don't wanna encode knowledge here how the CLAP plugin stuctures its data files, what the extensions of those files are, and so on. The CLAP plugin should be a black box to Staging.

src/midi/soundcanvas.cpp
```diff
@@ -816,6 +870,15 @@ static void init_soundcanvas_config_settings(SectionProp& sec_prop)
...
 	        "  <version>:  Use the exact specified model version (e.g., 'sc55_121').");
+	
+	str_prop = sec_prop.AddString("soundcanvas_rom_dir", when_idle, "");
+	str_prop->SetHelp(
+	        "The directory containing the Roland Sound Canvas ROMs (unset by default).\n"
+	        "The directory can be absolute or relative, or leave it unset to use the\n"
+	        "'soundcanvas-roms' directory in your DOSBox configuration directory. Other common\n"
+	        "system locations will be checked as well.\n"
+	        "Notes:\n"
+	        "  - The romset for each Sound Canvas model should be in its own directory");
```
johnnovak: Please remove the notes section; that's plugin-specific detail.

johnnovak:
Nice work @shermp, thanks for doing this.

I've left some comments about using the appropriate abstractions when generalising things and a few further simplifications that we need to do, but the rest looks good.

src/utils/env_utils.cpp
shermp: fs_utils and rom_dirs moved to the utils folder

johnnovak: Thanks man 😎 The mistake about `fs_utils` was mine when I moved files around like crazy 😆

shermp: Sorry! I moved `env_utils`. I left `fs_utils` alone, best to clean that up in another PR in my mind

src/midi/mt32.cpp
shermp: I originally did just copy and modify the MT-32 lookup functions into the sound canvas code. 

I thought it would be a good idea to unify it though, because it offers consistency in ROM directory lookup behavior across the code base.

But if you want them to be completely separate, I can go back to that.

src/midi/soundcanvas.cpp
shermp: Are you suggesting I get rid of the `filetype_at_depth()` stuff? I'm fine with that, I was of two minds about keeping that anyway.

src/utils/env_utils.cpp
johnnovak: That's fine, I'll do it later 😄

src/midi/mt32.cpp
johnnovak: > But if you want them to be completely separate, I can go back to that.

Yes please, thanks 😄 

But extracting `get_linux_platform_data_dirs()` would be worthwhile I think.

src/midi/soundcanvas.cpp
johnnovak: Yes please, that can go 👍🏻

src/midi/mt32.cpp
shermp: > But extracting `get_linux_platform_data_dirs()` would be worthwhile I think.

I'm not sure you can extract that with the way the rom dirs are constructed with different leaf names.

At best you could have a function that returns a list of xdg locations and then the caller has to construct a new list figuring out what lead directory to append. I'm not sure if it's worth the effort.

johnnovak: Yeah, I kept staring at this code for a good 10 minutes, looking for meaningful re-use opportunities, but you have probably been staring at it for longer, so I'm okay with that conclusion 😄 

Let's keep these fully separate then, and use similar(ish) lookup orders by convention per class/device/etc. We don't have hundreds of these path related config vars, just a low single digit number, and that's not gonna change. So yeah, let's not overdo it.

src/midi/soundcanvas.cpp
shermp: Actually, just trying to think how this would work. I still need to know which directory to fall back to.

Maybe I'll just choose the first non-empty directory.

src/midi/mt32.cpp
shermp: Alright, done. I made the minor `std_fs::path` refactor to the `get_rom_dirs()` mt32 (and then sound canvas) function.

src/midi/soundcanvas.cpp
shermp: Ok, the code just now tries to use the first non-empty directory.

src/midi/soundcanvas.cpp
shermp: Done.

shermp:
Oh, you renamed `GetConfigDir`... oops.

src/midi/soundcanvas.cpp
johnnovak: That's for the best, yeah.

src/midi/mt32.cpp
```diff
@@ -466,17 +466,14 @@ static std::deque<std_fs::path> get_rom_dirs()
 	auto rom_dirs = get_platform_rom_dirs();
 
 	// Get the user's configured ROM directory; otherwise use 'mt32-roms'
-	std::string selected_romdir = get_mt32_section()->GetString("romdir");
+	std_fs::path selected_romdir = get_mt32_section()->GetString("romdir");
```
johnnovak, line 469: Nice one 👍🏻

src/midi/mt32.cpp
```diff
@@ -466,17 +466,14 @@ static std::deque<std_fs::path> get_rom_dirs()
...
 
 	if (selected_romdir.empty()) { // already trimmed
 		selected_romdir = DefaultMt32RomsDir;
 	}
-	if (selected_romdir.back() != '/' && selected_romdir.back() != '\\') {
-		selected_romdir += CROSS_FILESPLIT;
-	}
```
johnnovak, line 476: 👍🏻

src/dosbox.h
```diff
@@ -167,11 +167,12 @@ inline bool is_machine_hercules() {
 #include "misc/logging.h"
 #endif // the logging system.
 
-constexpr auto DefaultMt32RomsDir   = "mt32-roms";
-constexpr auto DefaultSoundfontsDir = "soundfonts";
-constexpr auto GlShadersDir         = "glshaders";
-constexpr auto DiskNoiseDir         = "disknoises";
-constexpr auto PluginsDir           = "plugins";
+constexpr auto DefaultMt32RomsDir   	 = "mt32-roms";
+constexpr auto DefaultSoundcanvasRomsDir = "soundcanvas-roms";
```
johnnovak: minor: SoundCanvas (uppercase C because it's "Roland Sound Canvas")

johnnovak: Plus existing naming, but it should be SoundFonts (if you feel like it 😆). Very minor...

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +280,90 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
 	}
 }
 
+#if defined(WIN32)
+
+static std::deque<std_fs::path> get_platform_rom_dirs()
+{
+	return {
+	        get_config_dir() / DefaultSoundcanvasRomsDir,
+	        "C:\\soundcanvas-rom-data\\",
```
johnnovak: Yeah, let's lose `"C:\\soundcanvas-rom-data\\"`. I'm intending to remove the `C:\\...` path from the MT-32 too one day. Nobody stores emulator ROM files in the root folder of their system drive... 😆 It's just some questionable legacy cruft.

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +280,90 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
...
+
+#elif defined(MACOSX)
+
+static std::deque<std_fs::path> get_platform_rom_dirs()
+{
+	return {
+	        get_config_dir() / DefaultSoundcanvasRomsDir,
+	        resolve_home("~/Library/Audio/Sounds/SoundCanvas-Roms/"),
+	        "/usr/local/share/soundcanvas-rom-data/",
+	        "/usr/share/soundcanvas-rom-data/",
```
johnnovak: Yeah, let's lose the `/usr/` paths too. I'll clean up the MT-32 paths in a follow up PR myself, enough is enough 😆

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +280,90 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
...
+	// variable for the ROM directory if possible. Note, on POSIX systems 
+	// getenv/setenv are NOT thread safe
+	std::error_code ec;
+	// Search for a directory that exists and is not empty
+	for (auto rom_dir : get_rom_dirs()) {
+		if (std_fs::is_directory(rom_dir, ec)) {
+			if (!std_fs::is_empty(rom_dir, ec) && !ec) {
+				const auto canonical_rom_dir = std_fs::canonical(rom_dir, ec);
+				if (ec) continue;
+				LOG_MSG("SOUNRCANVAS: Setting SOUNDCANVAS_ROM_DIR env variable to '%s'", canonical_rom_dir.string().c_str());
```
johnnovak: typo: SOUN**D**CANVAS

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +280,90 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
...
+	rom_dirs.emplace_front(resolve_home(selected_romdir.string()));
+	return rom_dirs;
+}
+
 MidiDeviceSoundCanvas::MidiDeviceSoundCanvas()
 {
 	using namespace SoundCanvas;
 
+	// Before attempting to load any plugins, we must set an environment
+	// variable for the ROM directory if possible. Note, on POSIX systems
```
johnnovak: Maybe soften the wording a bit: 

> Before attempting to load any plugins, we set the ROM directory location hint via the SOUNDCANVAS_ROM_DIR.

Minor thing, just I kept the coupling to the absolute minimum between Staging and the plugin 😄

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +280,90 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
...
+	// Before attempting to load any plugins, we must set an environment
+	// variable for the ROM directory if possible. Note, on POSIX systems 
+	// getenv/setenv are NOT thread safe
+	std::error_code ec;
+	// Search for a directory that exists and is not empty
+	for (auto rom_dir : get_rom_dirs()) {
+		if (std_fs::is_directory(rom_dir, ec)) {
+			if (!std_fs::is_empty(rom_dir, ec) && !ec) {
+				const auto canonical_rom_dir = std_fs::canonical(rom_dir, ec);
+				if (ec) continue;
```
johnnovak: No one-liner conditionals, please. Run if through Clang-format, or change it manually to:

```cpp
if (ec) {
    continue;
}
```

src/dosbox.h
shermp: Actually, here's a question. Why are we sticking these in `dosbox.h`?

shermp: Oh, it's for directory creation in `main.cpp`

shermp: Ok renamed to `DefaultSoundCanvasRomsDir`

src/midi/soundcanvas.cpp
shermp: Gone.

src/midi/soundcanvas.cpp
shermp: Gone.

shermp:
Ok, review comments resolved.

I've also added creation of the sound canvas rom directory to `maybe_create_resource_directories()`, as well as a lambda refactor of that function.

johnnovak:
This looks very good now @shermp, thanks making the adjustments so quickly. I'll test it a bit later on macOS today then merge time 🚀

src/midi/soundcanvas.cpp
```diff
@@ -280,10 +280,88 @@ static void setup_filter(MixerChannelPtr& channel, const bool filter_enabled)
...
+				const auto canonical_rom_dir = std_fs::canonical(rom_dir, ec);
+				if (ec) {
+					continue;
+				}
+				LOG_MSG("SOUNDCANVAS: Setting SOUNDCANVAS_ROM_DIR env variable to '%s'", canonical_rom_dir.string().c_str());
+				set_env_var("SOUNDCANVAS_ROM_DIR", canonical_rom_dir.string().c_str());
+				break;
+			}
+		}
+	}
```
weirddan455: This isn't very robust. MT-32 actually iterates over every ROM dir rather than stopping at the first non-empty directory.

I tested this with my existing ROM location of `~/.config/dosbox/plugins/Nuked-SC55-Resources` and a simple `touch ~/.config/dosbox/soundcanvas-roms/a.txt` broke my audio. If Windows drops a `thumbs.db` or whatever hidden file it likes to do, people's audio will break.

I would say at the very least, the CLAP plugin should always check `Nuked-SC55-Resources` if it fails to find ROMs in the environment variable location. A better solution would be to handle the actual iteration on the plugin side and maybe we pass a list of directories over in an environment variable.

johnnovak: > I would say at the very least, the CLAP plugin should always check Nuked-SC55-Resources if it fails to find ROMs in the environment variable location. A better solution would be to handle the actual iteration on the plugin side and maybe we pass a list of directories over in an environment variable.


Good points @weirddan455 (and I forgot to test this), but why don't we just simply iterate over all potential resource paths in priority order, try to init the plugin using that path, and if we could init at least one model, we stop there.

That would work. We don't wanna support weirdo cases where you put one half of ROMs in one dir, the other half in another... Neither do we need to complicate things by passing a list of directories via the env var, that's not necessary.

weirddan455: That could work too if it’s not too slow to re-init the plugin multiple times.

weirddan455: Also it might be a bit noisy on the logs.

johnnovak: Yeah, let's try. I don't care that much about that noise, tbh. But we can try to clean it up a bit and only report the failed attempts in a single line then the success in detail or something.

shermp: I had a more robust solution but was told to get rid of it because it was too specific...

Personally, I prefer the environment list solution, and beef up the CLAP plugin to handle that case. If I do this, are you guys ok if we merge this as-is, then enhance it with a list solution in a follow-up PR?

shermp: Although if you're ok to wait a few more days, I can add the list support to the CLAP plugin then implement that here.

johnnovak: Why don't you try Staging side iteration through the dirs first and if it's not slow, we're done? In practice, I don't think you'll gain anything by complicating things on the plugin side.

Let's try the simplest thing first 😄

shermp: So, I went ahead and implemented the list support to the CLAP plugin. It was a fairly minor change, PR forthcoming. Implementing the support in Staging was equally straightforward.

johnnovak:
@shermp Released a new version of the CLAP plugin:

https://github.com/johnnovak/Nuked-SC55-CLAP/releases/tag/v0.10.0

Can you please bump the plugin version in the CI workflows, then we can test with the updated plugin before doing the final approval.

src/midi/soundcanvas.cpp
shermp: Ok, this has been implemented.

johnnovak:
@shermp I've just tested this on macOS, and looks like it needs some further work. Please see my observations below.

---

## Test 1

ROMs in `~/Library/Preferences/DOSBox/soundcanvas-roms`

image



- `mididevice soundcanvas` works and the path is passed via the env var as expected 🍏 

image

- But `mixer /listmidi` doesn't list the available plugins although the CLAP plugins were found  — that's a regression ☠️ 

image

image



## Test 2

- Rename `soundcanvas-roms` to `xsoundcanvas-roms` in the DOSBox prefs folder

image

- If I pass in the absolute path at startup, then then things work as expected 🍏 

image

image

image


## Test 3

- Changing the `soundcanvas_rom_dir` setting only seems to work for the _first time_. In the following example, I set the path to some bogus value which doesn't work. But then I'm setting the correct absolute path, and still no workie. If I switch the order and set the correct absolute path first, then it's fine (I'll spare you the details).

image

In theory, this should work because in `notify_soundcanvas_setting_updated` we should do a `MIDI_Init()` when changing `soundcanvas_rom_dir`, which is the read in the `MidiDeviceSoundCanvas` constructor. Well, you'll need to debug what's going on here 😆 

# Other issues

- Some lines are too long in the help text.
- Why don't we default this to `soundcanvas-roms`? That would be the most logical thing, no? It seems a bit silly to tell in the help text that "empty string means `soundcanvas-rom`", no? (And yes, probably the other path/dir config settings need to be revisited as well, but one step at a time 😄)

image

shermp:
I've done some testing, run out of time to check everything in your comment but at least on Windows, the following works:

**mixer /listmidi**

image

**Setting soundcanvas_rom_dir**

This seems to work for me, I can set a dummy value (and it falls back to the default location), or an absolute path several times, and it loads every time.

I'll try and test on Linux when I get home this evening.

shermp:
Ok, I can reproduce at least some of these issues if `soundcanvas` is not set as the `mididevice` in the config (I've been testing with a local conf file that has this set).

johnnovak:
> Ok, I can reproduce at least some of these issues if `soundcanvas` is not set as the `mididevice` in the config (I've been testing with a local conf file that has this set).

Yeah please fix as much as you can repro, then I can look into the macOS situation. Maybe that's some macOS specific quirk you cannot test.

shermp:
Alright, a bit more testing:

I cannot reproduce the `mixer /listmidi` when soundcanvas loaded issue on Windows or Linux.

As for the `soundcanvas_rom_dir` issue, I think I can reproduce this on Linux, although it seems to work on Windows.

johnnovak:
> Alright, a bit more testing:
> 
> I cannot reproduce the `mixer /listmidi` when soundcanvas loaded issue on Windows or Linux.
> 
> As for the `soundcanvas_rom_dir` issue, I think I can reproduce this on Linux, although it seems to work on Windows.

Yeah, do your best please and make sure all is flawless on Windows & Linux when changing `soundcanvas_rom_dir` and then setting `mididevice soundcanvas` at runtime repeatedly, multiple times, then I'll debug it on macOS if there are still issues. Then I'll create a macOS fixup commit for you to cherry-pick into this PR.

shermp:
So I think I've diagnosed what the problem is, not sure if it's fixable or not. Basically, the plugin is not reading the updated environment variable (it's getting the variable first set). Will have to read up a bit more on `setenv/getenv`.

shermp:
> So I think I've diagnosed what the problem is, not sure if it's fixable or not. Basically, the plugin is not reading the updated environment variable (it's getting the variable first set). Will have to read up a bit more on `setenv/getenv`.

Oops, I forgot about the third "overwrite" parameter, which I set a default value of `0`.

johnnovak:
> > So I think I've diagnosed what the problem is, not sure if it's fixable or not. Basically, the plugin is not reading the updated environment variable (it's getting the variable first set). Will have to read up a bit more on `setenv/getenv`.
> 
> Oops, I forgot about the third "overwrite" parameter, which I set a default value of `0`.

That would explain the behaviour, yep.

shermp:
Alright, fixing that little oversight seems to have got things working properly on linux:

image

image

image

The reason this was an issue on Linux and MacOS but not Windows is Windows `SetEnvironmentVariableA()` function doesn't have an overwrite parameter.

shermp:
I was thinking about this comment @johnnovak 
> Why don't we default this to soundcanvas-roms? That would be the most logical thing, no? It seems a bit silly to tell in the help text that "empty string means soundcanvas-rom", no? (And yes, probably the other path/dir config settings need to be revisited as well, but one step at a time 😄)

I don't think this will work. In the case of the soundcanvas anyway, a relative path in this config option will try and find a folder relative to the **current working directory**. Therefore `soundcanvas-roms` would try and find `<current_dir>/soundcanvas-roms` rather than searching the users config location.

shermp:
I've been trying runtime changing of rom dirs, and enabling and disabling soundcanvas support at runtime on both Linux and Windows, and it seems to be working Ok.

Can you please test on MacOS again @johnnovak?

EDIT: I also fixed the help line length.

johnnovak:
> I've been trying runtime changing of rom dirs, and enabling and disabling soundcanvas support at runtime on both Linux and Windows, and it seems to be working Ok.
> 
> Can you please test on MacOS again @johnnovak?
> 
> EDIT: I also fixed the help line length.

Thanks @shermp. I will retest later today.

johnnovak:
@shermp The env var handling works fine now, good job 👍🏻 

I've fixed the `/LISTMIDI` problem, just incorporate this commit into this PR and then it's good to go: https://github.com/dosbox-staging/dosbox-staging/commit/1c110e8be258e4cefd89ce435775ea16d8b046cd

So `/LISTMIDI` worked _after_ setting the SoundCanvas as the current MIDI device with `mididevice soundcanvas` at least once after startup, but it must work _before_ doing that, too (e.g. when no MIDI device is active, or when some other MIDI device is in use). The solution is to simply set the env var in `SOUNDCANVAS_ListDevices()` as well.

shermp:
> @shermp The env var handling works fine now, good job 👍🏻
> 
> I've fixed the `/LISTMIDI` problem, just incorporate this commit into this PR and then it's good to go: [1c110e8](https://github.com/dosbox-staging/dosbox-staging/commit/1c110e8be258e4cefd89ce435775ea16d8b046cd)
> 
> So `/LISTMIDI` worked _after_ setting the SoundCanvas as the current MIDI device with `mididevice soundcanvas` at least once after startup, but it must work _before_ doing that, too (e.g. when no MIDI device is active, or when some other MIDI device is in use). The solution is to simply set the env var in `SOUNDCANVAS_ListDevices()` as well.

This fix sort of works, but if you call `mixer /listmidi` before setting a correct `soundcanvas_rom_dir`, it will never work, even if setting the correct `soundcanvas_rom_dir` after.

eg the following will always print "No available models" assuming a romset cannot initially be found:

```
Z:> mixer /listmidi
Z:> soundcanvas_rom_dir <path/to/correct/dir>
Z:> mixer /listmidi
Z:> mididevice soundcanvas
Z:> mixer /listmidi
```

shermp:
I think I've found the problem, `SOUNDCANVAS_ListDevices()` will only try and init the plugin once, because of the `available_models_initialised` variable.

johnnovak:
> I think I've found the problem, `SOUNDCANVAS_ListDevices()` will only try and init the plugin once, because of the `available_models_initialised` variable.

Okay, so you'll need to reset that flag to invalidate the cache in the config setting update handler when `soundcanvas_rom_dir` is changed. I missed that part 😄

shermp:
Ok, think I've fixed it. Changing the `soundcanvas_rom_dir` option now clears the set of `available_models` and repopulates it.

johnnovak (approved):
I've done another quick round of `/LISTMIDI` testing on macOS and all seem to be fine. Nice work @shermp 👍🏻 Let me know if you're done, then I'll merge it.

shermp:
Ship it!

PR #4587, opened 2025-11-06
Always use SDL_PIXELFORMAT_ARGB8888 for the SDL texture renderer

weirddan455:
# Description

We were previously just using the first supported format reported by the renderer. This doesn't always work though. I found that if we get BGRA our colors get messed up.

Most GPUs should support this format.
If not, SDL will take a slow path and the code should still work.

This is also the format the OpenGL renderer always uses.

The `GFX_CAN_RANDOM` check didn't make any sense. We're drawing to a software surface so we always have random access to the framebuffer, same as we do in the OpenGL renderer.

## Related issues

https://github.com/libsdl-org/SDL/issues/14403

I initially reported this as an SDL regression as I found this bug by testing with the latest `main` build of SDL. It turns out the only relevant thing they changed was re-ordering the formats returned by `SDL_GetRendererInfo()` and we just weren't handling BGRA correctly.

# Release notes

Probably nothing user observable unless they're running a git build of SDL.

# Manual testing

Colors look correct in both the git and release builds of SDL.

- [ ] Windows
- [ ] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/render/opengl_renderer.cpp
+++ b/src/gui/render/opengl_renderer.cpp
@@ -161,6 +161,8 @@ bool OpenGlRenderer::InitRenderer()
 	         safe_gl_get_string(GL_SHADING_LANGUAGE_VERSION, "unknown"),
 	         safe_gl_get_string(GL_VENDOR, "unknown"));
 
+	// TODO: GFX_CAN_32 | GFX_CAN_RANDOM is always set for both SDL and OpenGL backends.
+	// Opportunity to remove dead code from render.cpp.
 	gfx_flags = GFX_CAN_32 | GFX_CAN_RANDOM;
 
 	return true;
--- a/src/gui/render/sdl_renderer.cpp
+++ b/src/gui/render/sdl_renderer.cpp
@@ -13,6 +13,8 @@
 
 CHECK_NARROWING();
 
+static constexpr uint32_t SdlPixelFormat = SDL_PIXELFORMAT_ARGB8888;
+
 SdlRenderer::SdlRenderer(const int x, const int y, const int width,
                          const int height, const uint32_t sdl_window_flags,
                          const std::string& render_driver,
@@ -118,24 +120,10 @@ bool SdlRenderer::InitRenderer(const std::string& render_driver)
 	}
 	LOG_MSG("SDL: Using '%s' SDL render driver", info.name);
 
-	pixel_format = SDL_AllocFormat(info.texture_formats[0]);
-	if (!pixel_format) {
-		LOG_ERR("SDL: Error allocating pixel format: %s", SDL_GetError());
-		SDL_DestroyRenderer(renderer);
-		return false;
-	}
-
-	switch (SDL_BITSPERPIXEL(info.texture_formats[0])) {
-	case 8: gfx_flags = GFX_CAN_8; break;
-	case 15: gfx_flags = GFX_CAN_15; break;
-	case 16: gfx_flags = GFX_CAN_16; break;
-	case 24: // SDL_BYTESPERPIXEL is probably 4, though
-	case 32: gfx_flags = GFX_CAN_32; break;
-	}
-
-	if (!(info.flags & SDL_RENDERER_ACCELERATED)) {
-		gfx_flags |= GFX_CAN_RANDOM;
-	}
+	// TODO: GFX_CAN_32 | GFX_CAN_RANDOM is always set for both SDL and OpenGL backends.
+	// Opportunity to remove dead code from render.cpp.
+	static_assert(SDL_PIXELTYPE(SdlPixelFormat) == SDL_PIXELTYPE_PACKED32);
+	gfx_flags = GFX_CAN_32 | GFX_CAN_RANDOM;
 
 	if (SDL_SetRenderDrawColor(renderer, 0, 0, 0, SDL_ALPHA_OPAQUE) < 0) {
 		LOG_ERR("SDL: Error setting render clear color: %s", SDL_GetError());
@@ -146,10 +134,6 @@ bool SdlRenderer::InitRenderer(const std::string& render_driver)
 
 SdlRenderer::~SdlRenderer()
 {
-	if (pixel_format) {
-		SDL_FreeFormat(pixel_format);
-		pixel_format = {};
-	}
 	if (renderer) {
 		// Frees associated textures automatically.
 		SDL_DestroyRenderer(renderer);
@@ -210,20 +194,12 @@ void SdlRenderer::UpdateViewport(const DosBox::Rect draw_rect_px)
 
 bool SdlRenderer::UpdateRenderSize(const int render_width_px, const int render_height_px)
 {
-	// Retrieve available texture formats.
-	SDL_RendererInfo renderer_info;
-	if (SDL_GetRendererInfo(renderer, &renderer_info) < 0) {
-		LOG_ERR("SDL: Error retrieving SDL renderer info: %s",
-		        SDL_GetError());
-		return false;
-	}
-
 	if (texture) {
 		SDL_DestroyTexture(texture);
 	}
 
 	texture = SDL_CreateTexture(renderer,
-	                            renderer_info.texture_formats[0],
+	                            SdlPixelFormat,
 	                            SDL_TEXTUREACCESS_STREAMING,
 	                            render_width_px,
 	                            render_height_px);
@@ -267,13 +243,13 @@ bool SdlRenderer::UpdateRenderSize(const int render_width_px, const int render_h
 	                                               render_width_px,
 	                                               render_height_px,
 	                                               BitDepth,
-	                                               renderer_info.texture_formats[0]);
+	                                               SdlPixelFormat);
 
 	last_framebuf = SDL_CreateRGBSurfaceWithFormat(Flags,
 	                                               render_width_px,
 	                                               render_height_px,
 	                                               BitDepth,
-	                                               renderer_info.texture_formats[0]);
+	                                               SdlPixelFormat);
 
 	if (!curr_framebuf || !last_framebuf) {
 		SDL_DestroyTexture(texture);
@@ -451,7 +427,6 @@ RenderedImage SdlRenderer::ReadPixelsPostShader(const DosBox::Rect output_rect_p
 uint32_t SdlRenderer::MakePixel(const uint8_t red, const uint8_t green,
                                 const uint8_t blue)
 {
-	assert(pixel_format);
-
-	return SDL_MapRGB(pixel_format, red, green, blue);
+	static_assert(SdlPixelFormat == SDL_PIXELFORMAT_ARGB8888);
+	return ((blue << 0) | (green << 8) | (red << 16)) | (255 << 24);
 }
--- a/src/gui/render/sdl_renderer.h
+++ b/src/gui/render/sdl_renderer.h
@@ -79,7 +79,6 @@ private:
 	// True if the last framebuffer has been updated since the last present
 	bool last_framebuf_dirty = false;
 
-	SDL_PixelFormat* pixel_format = {};
 	SDL_Texture* texture          = {};
 
 	TextureFilterMode texture_filter_mode = TextureFilterMode::Bilinear;


weirddan455:
This SonarCloud shit is just annoying. It's failing because I added TODO comments... It's also been failing on `main` for months now.

EDIT: I guess the real reason it's failing is because it wants me to cast constants to `std::byte` before doing any sort of bitwise operation. Idk... we don't do that anywhere else in the codebase.

johnnovak:
> This SonarCloud shit is just annoying. It's failing because I added TODO comments... It's also been failing on `main` for months now.

Yeah, ignore that.

> EDIT: I guess the real reason it's failing is because it wants me to cast constants to `std::byte` before doing any sort of bitwise operation. Idk... we don't do that anywhere else in the codebase.

Ignore that too 😄 

Sometimes it raises good suggestions and improvement opportunities that PVS doesn't, but it's a "soft check". We sometimes look at the results in the PRs and take some of the advice when it makes sense. But yeah, a good 30-50% of the things it flags is just bullshit.

In the paid version at work you can dismiss/close warnings, so you can make the check go green. You can also finetune the warnings, but not in the free version we're using. But hey, it's free 😄

johnnovak:
Nice one @weirddan455. I believe this has the potential to break the texture renderer on big-endian platforms, see the description of ARGB8888 here:

https://wiki.libsdl.org/SDL3/SDL_PixelFormat

However, I'm completely fine with removing BE support. More than fine, I'm actually advocating for it. I remember you and I spent significant time documenting our pixel format usages on BE and LE platforms. The problem is, only one person in the team was able to verify our changes/assumptions (kcgen), as only he had a BE PPC box. We can break BE support any moment and we wouldn't know! Sure, there have been two guys over the years sporadically submitting BE/ppc64 patches, but keeping high-maintenance support for an essentially dead platform would be a bit of a crazy choice for us. We simply don't have the hardware to test BE-correctness, only a handful of people are on BE machines today, and BE on the desktop won't make a comeback. It's dead.

I'm of the same opinion as the Darktable guy. Just assume little-endian in all our codebase and that's it. Also, I'm also not buying the argument that supporting BE would esult in more robust code. It will just result in more busywork and code paths we cannot test.

https://github.com/darktable-org/darktable/issues/14739#issuecomment-2263563755

So if you see further simplification opportunities regarding removing BE support, go for it!

weirddan455:
It depends on how the rendering code writes the pixels out. `SDL_PIXELFORMAT_ARGB8888` is the endianess agnostic "byte array" one. If the rendering code is writing pixels out 1 byte at a time (or as an array of `uint8_t`), that's the correct one. If it's writing it out as a `uint32_t` we need `SDL_PIXELFORMAT_BGRA32` instead.

But yeah the two are equal on little endian. There's a bit in the SDL 2 header that says (trimmed):

````
#if SDL_BYTEORDER == SDL_BIG_ENDIAN
    SDL_PIXELFORMAT_BGRA32 = SDL_PIXELFORMAT_BGRA8888
#else
    SDL_PIXELFORMAT_BGRA32 = SDL_PIXELFORMAT_ARGB8888
````

I was getting a headache looking at `render.cpp` though and couldn't determine which one is correct to use. I would probably only feel confident if I had a big endian machine to test on.

**Comment by weirddan455** [comment]
Also I chose `SDL_PIXELFORMAT_ARGB8888` because it is the first one returned by `SDL_GetRendererInfo()` so it keeps behavior the same. If this patch doesn't work on big endian, it was very likely already broken on `main`.

**Comment by johnnovak** [comment]
> Also I chose `SDL_PIXELFORMAT_ARGB8888` because it is the first one returned by `SDL_GetRendererInfo()` so it keeps behavior the same. If this patch doesn't work on big endian, it was very likely already broken on `main`.

All good by me. As I said, let's just kill BE support. We require a LE host CPU, end of story.

File: src/gui/render/opengl_renderer.cpp [hunk 1]
```diff
@@ -161,6 +161,8 @@ bool OpenGlRenderer::InitRenderer()
 	         safe_gl_get_string(GL_SHADING_LANGUAGE_VERSION, "unknown"),
 	         safe_gl_get_string(GL_VENDOR, "unknown"));
 
+	// TODO: GFX_CAN_32 | GFX_CAN_RANDOM is always set for both SDL and OpenGL backends.
+	// Possibily to remove dead code from render.cpp.
```
johnnovak: Is there an opportunity to lose `gfx_flags` then? Maybe do it in this PR?

src/gui/render/sdl_renderer.cpp
```diff
@@ -118,24 +120,10 @@ bool SdlRenderer::InitRenderer(const std::string& render_driver)
...
-	case 32: gfx_flags = GFX_CAN_32; break;
-	}
-
-	if (!(info.flags & SDL_RENDERER_ACCELERATED)) {
-		gfx_flags |= GFX_CAN_RANDOM;
-	}
+	// TODO: GFX_CAN_32 | GFX_CAN_RANDOM is always set for both SDL and OpenGL backends.
+	// Possibily to remove dead code from render.cpp.
+	static_assert(SDL_PIXELTYPE(SdlPixelFormat) == SDL_PIXELTYPE_PACKED32);
+	gfx_flags = GFX_CAN_32 | GFX_CAN_RANDOM;
```
johnnovak, line 126: Same comment.

src/gui/render/opengl_renderer.cpp
johnnovak: typo: Possibility? Possible?

weirddan455: Possibility is what I typo'd but maybe opportunity is a better word.

Yes, but not for this PR. grep for `gfx_fags` in `render.cpp`. It's used in 20 places in code I am very unfamiliar with. There's potential to hack and slash a lot of code in there if we can safely say we don't care about non 32-bit framebuffers on the host side.

johnnovak: That's fine, let's leave it for later. I'm thinking now we're only dealing with 32-bit BGRA data, so all the rest could be removed. I'll play around with that sometime (and make sure they're not used for image and video capturing somehow, but I don't think so as those just get the raw bytebuffer of the rendered frame from the VGA emulation).

Ideal outcome: we're only dealing with 32-bit BGRA everywhere, and we assume LE host CPU. Then lots of crap can fall away, indeed.

johnnovak (approved):
Merge when ready @weirddan455 👍🏻

Grounded0:
I'd say we'll definitely hear about it on the issue tracker if we blow up big endian support. I've got a spare PowerBook G4 around, but I'm not sacrificing it to run Linux.

johnnovak:
> I'd say we'll definitely hear about it on the issue tracker if we blow up big endian support. I've got a spare PowerBook G4 around, but I'm not sacrificing it to run Linux.

It's up to us decide what we can and want to support. 

We support the last 3 major macOS releases. None of them support PPC. Even Intel is on the way out slowly.

We provide x86 Linux binaries and yeah ARM is supported (but we don't provide a binary yet). PPC & BE was always kcgen's pet project and probably a total of 5 people used it regularly. So come on, man 😆 PowerBook G4 doesn't run the last 3 macOS releases last time I checked... PPC is long dead...

weirddan455:
Big endian is pretty much dead. It's even more of a niche than this Linux KMSDRM thing I was working on. The major Linux distros have all dropped support for the old PowerPC Macs. IIRC kcgen was testing on NetBSD or something.

That said my MDS/MDF PR I merged last week did have some byte swapping code for big endian. That one I'm pretty sure is correct because it's self-contained reading from a little endian filetype. The pixel/video stuff is hard to prove correctness without actually testing on hardware that (mostly) no one uses anymore.

johnnovak:
> That said my MDS/MDF PR I merged last week did have some byte swapping code for big endian. That one I'm pretty sure is correct because it's self-contained reading from a little endian filetype.

We should continue to use byte swapping functions when we're reading BE formats (e.g., network packets). But that's different because we can easily test the correctness of such code on our LE hardware. 

> The pixel/video stuff is hard to prove correctness without actually testing on hardware that (mostly) no one uses anymore.

100%. And it would add a significant amount of overhead to properly support both BE and LE even if we had the hardware to test it.

And anyway, we support the last 2-3 major macOS, Windows and Linux releases (except when an OS is long discontinued, like Windows 7) and hardware not older than 10-15 years (but 15 years is pushing it).

> The last Macs to use a PowerPC processor were the Power Mac G5, discontinued in August 2006, and the iMac G5 (late 2005), which was the final PowerPC-based iMac. Apple's transition to Intel processors began in 2006.

Grounded0:
I was just saying that if big endian support would break down at some point, we would hear about it. Since nobody has reported it being broken, we can assume it's still working. We definitely don't have to support something we don't even have the hardware for to test against.

johnnovak:
> I was just saying that if big endian support would break down at some point, we would hear about it. Since nobody has reported it being broken, we can assume it's still working. We definitely don't have to support something we don't even have the hardware for to test against.

That's fair. I misread as if you were arguing to keep supporting it indefinitely, apologies.

Torinde:
The latest setup that supports BE (excluding mainframes, I think) is [Raptor](https://www.raptorcs.com/content/TLSDS3/intro.html) POWER9 with [Gentoo, Adélie, custom Linux builds](https://catfox.life/2018/11/03/clearing-confusion-regarding-modern-powerpc-endianness/) or [FreeBSD](https://www.freebsd.org/platforms/) (although with issues [1](https://forums.raptorcs.com/index.php?topic=379.0), [2](https://wiki.freebsd.org/powerpc/Raptor/TalosII)), but it seems more and more distributions drop BE support and POWER9 EOL is in January 2026 for [IBM systems](https://www.ibm.com/docs/en/announcements/services-withdrawal-discontinuance-service-select-power-systems-products-replacements-available).

@runlevel5, do you think Power10/11 will appear in a Raptor-like systems? Will Raptor POWER9 systems remain available for purchase after Jan 2026? Do you use yours in BE mode?

weirddan455:
@Torinde That computer comes pre-installed with little endian Debian.

> To get you started as quickly as possible, the Special Developer System comes pre-loaded with a Debian Buster ppc64el installation

ppc64el is little endian because someone thought it would be cute to byte swap “le”

While that CPU may be able to boot in big endian mode, I think it’s done very little in practice. IIRC ARM is the same way with a big endian mode but it’s almost never done.

runlevel5:
> @runlevel5, do you think Power10/11 will appear in a Raptor-like systems? Will Raptor POWER9 systems remain available for purchase after Jan 2026? Do you use yours in BE mode?

P11 is very likely. AFAIK I do not see any issue why Raptor would cease the sale of the POWER9-based workstation. And yes I do run VMs in BE. I agree that BE will get less popular these days as many distributions have ceased support for BE variant.

johnnovak:
Yeah, I think it's reasonable to say we only support LE from now on. x86 and arm are "officially" supported, and that's it. But that's really for the dynarec only; the rest of the emulation is CPU architecture-agnostic so you can use it with the normal core on any LE platform if you find a way to compile it yourself (not 100% true because you'll need to be able to compile our deps too... but in theory it's true 😆).

We can leave those PPC BE, PPC64 LE, and MIPS32 LE dynarec backends alone, they don't really do any harm, but they're at the mercy of contributors propping them up from time to time because we do zero testing for them, and we won't do CI builds for them either.

Given we'll make no special effort to be BE compatible anymore, it's only a matter of time we'll break BE compatibility somehow, so probably `risc_ppc.h` should be removed.

Grounded0:
What I'm personally aware of is this boutique Linux distro:

http://fienixppc.blogspot.com

If you want to stick with Power ISA, then buy POWER8+ hardware that's capable of `ppc64le`. 🙂

> Given we'll make no special effort to be BE compatible anymore, it's only a matter of time we'll break BE compatibility somehow, so probably risc_ppc.h should be removed.

Let's leave everything in place for now. Maybe we get lucky and get someone who uses this hardware fixing any endianness issues for us once we eventually break it.

johnnovak:
> What I'm personally aware of is this boutique Linux distro:

image

image

Checks out 😂 Party on 😎

Grounded0:
They lost their repository in February 2025:

https://fienixppc-news.blogspot.com

johnnovak:
> They lost their repository in February 2025:
> 
> https://fienixppc-news.blogspot.com

And apparently their motivation too 😂 

Reminds me of the Amiga "revivalist" jokers. All these nu-Amiga hardware for the price of a new PC, no software, and the user base of 3 people 😆 (all devs, of course)

https://www.amigaos.net/hardware/36/pegasos-ii

https://www.morphos-team.net/

etc.

Grounded0:
This veers to shitposting, but my measure of a healthy and viable platform is:

"Is there a Mastodon client available for this platform?"

Both DOS and Classic Mac OS check out, so I'm happy with both. 🙂

johnnovak:
> This veers to shitposting, but my measure of a healthy and viable platform is:
> 
> "Is there a Mastodon client available for this platform?"
> 
> Both DOS and Classic Mac OS check out, so I'm happy with both. 🙂

I can think of a few:

- SID player
- Some Rogue clone
- Tetris
- C64 & ZX Speccy emulator
- etc.

Grounded0:
The colours seem to be correct on big endian:

- https://github.com/dosbox-staging/dosbox-staging/pull/4592#issuecomment-3518534765

Issue #4369, opened 2025-06-09
xbr-lv3 OpenGL shader compilation warning


Achaean:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Linux x86_64

### If other OS or Linux, please describe

Debian stable based, XFCE.

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Are you able to trigger the issue with our official DOSBox Staging release?

- [x] I am running an official Staging release, not a release installed by my operating system's package manager, or a version repackaged by an external team.

### Relevant hardware info

Quad core AMD Athlon, Nvidia GT 710.

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

From console's output (after activating xbr-lv3, instead of the default):
```
2025-06-09 14:56:08.289 | OPENGL: Program info log:
Fragment info
-------------
0(210) : warning C7050: "pix1" might be used before being initialized
0(211) : warning C7050: "blend1" might be used before being initialized
0(210) : warning C7050: "pix2" might be used before being initialized
0(211) : warning C7050: "blend2" might be used before being initialized
```

### Steps to reproduce the behaviour

Explain in detail how to reproduce the bug. Assume no familiarity with the software.
1. Activate scaler/xbr-lv3. (at config file: glshader = scaler/xbr-lv3)
2. Launch DOSBox.

[CONSOLE OUTPUT.zip](https://github.com/user-attachments/files/20654357/CONSOLE.OUTPUT.zip)

[dosbox-staging.conf.zip](https://github.com/user-attachments/files/20654360/dosbox-staging.conf.zip)

### Your configuration

```ini

```

### Provide a log

```text

```

### Code of conduct and contributing guidelines

- [x] Yes, I agree.

johnnovak:
But the output looks fine?

Achaean:
1) At the uploaded console's output  -> lines 53-59:
```
2025-06-09 14:56:08.289 | OPENGL: Program info log:
Fragment info
-------------
0(210) : warning C7050: "pix1" might be used before being initialized
0(211) : warning C7050: "blend1" might be used before being initialized
0(210) : warning C7050: "pix2" might be used before being initialized
0(211) : warning C7050: "blend2" might be used before being initialized
```
2) Actually, everything seems to be fine here. It don't seems to be a bug.
Just a warning.
[...]

Issue #4507, opened 2025-09-09
Collapse CRT shader variants into a single shader


johnnovak:
Currently, we have many CRT shader variants for the various machine types and screen resolutions. Most of these shaders use the same shader under the hood (Hyllian's CRT shader) with different parameters set via define macros at shader compile time.

This was just a stopgap measure to get the auto CRT feature out quickly, but this solution is hard to maintain. We need to collapse all these CRT shader variants into a single shader and set the parameters via uniforms at runtime. This way, we can switch between different CRT shader variants per frame (if we need to) with zero overhead.

### Depends on

- https://github.com/dosbox-staging/dosbox-staging/issues/4505

johnnovak:
Implemented in https://github.com/dosbox-staging/dosbox-staging/pull/4590

PR #4590, opened 2025-11-07
Add shader preset support

johnnovak:
# Description

As per the title, this PR introduces shader preset support. Specifically:

- The 20+ copy-pasted CRT shader variations that only differed in shader parameters have been condensed to a single `crt/crt-hyllian` shader (so back to where we were before introducing the adaptive CRT shaders, more or less).
- Switching between the shader variations is now accomplished via simply updating the shader parameters via uniforms. We can do this per-frame basis without any performance penalty.
- Shader parameters are organised into presets which are simple INI files (see the new `shader-presets` folder in resources).
- I've introduced the simpleini header-only INI parser/writer library to the project because our home-grown INI loader is a bit of a mess. This will come in handy later to support the OSD, e.g., it can preserve comments and the order of entries which is important when modifying existing hand-written config files. We will completely replace the current config loader/writer with simpleini later on, of course, but that's out of scope for this PR.
- `glshader` has been renamed to `shader`, but `glshader` is still accepted with a warning.
- We can now specify shader presets in the `shader = SHADER_NAME:PRESET_NAME` format, but I'm reluctant to advertise that just yet (would need to test how things work if `shader-presets` is also present in the user folder and I've had enough of this for a while 😅). This is an internal-only feature for now.
- Some follow-up cleanup of the stuff @weirddan455 started in the previous PR.

## Related issues

- Implements https://github.com/dosbox-staging/dosbox-staging/issues/4507
- Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4369

# Release notes

- The `glshader` setting in the `[render]` section has been renamed to `shader` (`glshader` is still accepted with a warning).
- Automatic shader switching with the adaptive CRT shaders active is now 100% smooth on screen mode changes (no more occasional minor audio hiccups).

# Manual testing

Tested / confirmed that:

- Shader & preset caches are used when triggering preset changes with the auto-CRT meta shader (by looking at the logs with `DEBUG_OPENGL` defined).
- All auto-CRT shaders work (`crt-auto`, `crt-auto-machine`, `crt-auto-arcade`, `crt-auto-arcade-sharp`).
- Watched **Show by Majic 12**; that demo switches between custom screen modes like crazy.
- Shader & preset reloading works (Cmd/Ctrl+F2).
- All shaders can be set with `shader <shader-name>` at runtime.
- Manual preset switching works (e.g. `shader crt/crt-hyllian:vga-4k`).
- Specifying the optional `.glsl` extension in the shader name works (e.g. `shader crt/crt-hyllian.glsl:vga-4k`).
- Specifying a non-existent preset name will load the default preset with a warning (e.g., `shader sharp:asdf`).
- Fallback to `crt-auto` & `sharp` works if shader compilation fails when switching to another shader or reloading the current shader.
- Various render setting combos still work with the shaders (e.g., `aspect stretch` and `integer_scaling off` stretches the image to the window as you keep resizing it, and shader auto-switching works).
- Executed the auto-CRT test cases:
  - https://github.com/dosbox-staging/dosbox-staging/wiki/Video-tests-%E2%80%94-CRT-shaders#crt-auto-tests-using-qpv
  - https://github.com/dosbox-staging/dosbox-staging/wiki/Video-tests-%E2%80%94-CRT-shaders#crt-auto-tests-using-deluxe-paint-iie
- `output = texture` and `texturenb` still work

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/render/opengl_renderer.cpp
+++ b/src/gui/render/opengl_renderer.cpp
@@ -70,7 +70,8 @@ OpenGlRenderer::OpenGlRenderer(const int x, const int y, const int width,
 {
 	window = CreateSdlWindow(x, y, width, height, sdl_window_flags);
 	if (!window) {
-		const auto msg = format_str("OPENGL: Error creating window: %s", SDL_GetError());
+		const auto msg = format_str("OPENGL: Error creating window: %s",
+		                            SDL_GetError());
 		LOG_ERR("%s", msg.c_str());
 		throw std::runtime_error(msg);
 	}
@@ -117,7 +118,7 @@ SDL_Window* OpenGlRenderer::CreateSdlWindow(const int x, const int y,
 	auto flags = sdl_window_flags;
 	flags |= SDL_WINDOW_OPENGL;
 
-	SDL_Window *window = SDL_CreateWindow(DOSBOX_NAME, x, y, width, height, flags);
+	SDL_Window* window = SDL_CreateWindow(DOSBOX_NAME, x, y, width, height, flags);
 	if (!window) {
 		// Try again without sRGB. This has been a problem with KMSDRM.
 		SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, 0);
@@ -656,14 +657,16 @@ std::optional<GLuint> OpenGlRenderer::BuildShaderProgram(const std::string& shad
 	return shader_program;
 }
 
-bool OpenGlRenderer::SetShader(const std::string& shader_descriptor)
+OpenGlRenderer::SetShaderResult OpenGlRenderer::SetShader(const std::string& shader_descriptor)
 {
 	return SetShaderInternal(shader_descriptor);
 }
 
-bool OpenGlRenderer::SetShaderInternal(const std::string& shader_descriptor,
-                                       const bool force_reload)
+OpenGlRenderer::SetShaderResult OpenGlRenderer::SetShaderInternal(
+        const std::string& shader_descriptor, const bool force_reload)
 {
+	using enum OpenGlRenderer::SetShaderResult;
+
 	auto& shader_manager = ShaderManager::GetInstance();
 
 	const auto curr_descriptor = force_reload
@@ -675,11 +678,20 @@ bool OpenGlRenderer::SetShaderInternal(const std::string& shader_descriptor,
 	const auto new_descriptor = shader_manager.GetCurrentShaderDescriptor();
 
 	if (!MaybeSwitchShaderAndPreset(curr_descriptor, new_descriptor)) {
-		return false;
+		return ShaderError;
+	}
+
+	if (!new_descriptor.preset_name.empty() &&
+	    current_shader_descriptor.preset_name.empty()) {
+
+		current_shader_descriptor_string = current_shader_descriptor.shader_name;
+
+		// We could set the shader but not the preset.
+		return PresetError;
 	}
 
 	current_shader_descriptor_string = shader_descriptor;
-	return true;
+	return Ok;
 }
 
 bool OpenGlRenderer::ForceReloadCurrentShader()
@@ -695,7 +707,8 @@ bool OpenGlRenderer::ForceReloadCurrentShader()
 	shader_preset_cache.erase(descriptor.ToString());
 
 	constexpr auto ForceReload = true;
-	return SetShaderInternal(current_shader_descriptor_string, ForceReload);
+	return (SetShaderInternal(current_shader_descriptor_string, ForceReload) ==
+	        OpenGlRenderer::SetShaderResult::Ok);
 }
 
 bool OpenGlRenderer::MaybeAutoSwitchShader(const DosBox::Rect canvas_size_px,
@@ -725,8 +738,6 @@ bool OpenGlRenderer::MaybeSwitchShaderAndPreset(const ShaderDescriptor& curr_des
 			return false;
 		}
 	}
-
-	const auto& shader_info = shader_cache[new_descriptor.shader_name].info;
 
 	if (changed_shader ||
 	    (curr_descriptor.preset_name != new_descriptor.preset_name)) {
@@ -767,21 +778,20 @@ void OpenGlRenderer::SwitchShaderPresetOrSetDefault(const ShaderDescriptor& desc
 		current_shader_preset = default_preset;
 	};
 
+	current_shader_descriptor = descriptor;
+
 	if (descriptor.preset_name.empty()) {
 		set_default_preset();
 
 	} else {
-		const auto maybe_preset = GetOrLoadAndCacheShaderPreset(descriptor);
+		if (const auto maybe_preset = GetOrLoadAndCacheShaderPreset(descriptor);
+		    maybe_preset) {
 
-		if (maybe_preset) {
 			current_shader_preset = *maybe_preset;
-		} else {
-			LOG_WARNING(
-			        "OPENGL: Error loading shader preset '%s'; "
-			        "using default preset",
-			        descriptor.ToString().c_str());
 
+		} else {
 			set_default_preset();
+			current_shader_descriptor.preset_name.clear();
 		}
 	}
 }
--- a/src/gui/render/opengl_renderer.h
+++ b/src/gui/render/opengl_renderer.h
@@ -40,7 +40,7 @@ public:
 	bool UpdateRenderSize(const int new_render_width_px,
 	                      const int new_render_height_px) override;
 
-	bool SetShader(const std::string& shader_descriptor) override;
+	SetShaderResult SetShader(const std::string& shader_descriptor) override;
 
 	bool MaybeAutoSwitchShader(const DosBox::Rect canvas_size_px,
 	                           const VideoMode& video_mode) override;
@@ -82,8 +82,8 @@ private:
 
 	bool InitRenderer();
 
-	bool SetShaderInternal(const std::string& shader_descriptor,
-	                       const bool force_reload = false);
+	SetShaderResult SetShaderInternal(const std::string& shader_descriptor,
+	                                  const bool force_reload = false);
 
 	bool MaybeSwitchShaderAndPreset(const ShaderDescriptor& curr_descriptor,
 	                                const ShaderDescriptor& new_descriptor);
@@ -155,6 +155,8 @@ private:
 
 	Shader current_shader              = {};
 	ShaderPreset current_shader_preset = {};
+
+	ShaderDescriptor current_shader_descriptor = {};
 
 	// Current shader descriptor string as set by the user (e.g., if the
 	// user set `crt-auto`, this will stay `crt-auto`; it won't be synced to
--- a/src/gui/render/render.cpp
+++ b/src/gui/render/render.cpp
@@ -575,12 +575,25 @@ void RENDER_AddMessages()
 
 static bool set_shader(const std::string& descriptor)
 {
-	if (!GFX_GetRenderer()->SetShader(descriptor)) {
-		return false;
-	}
+	using enum RenderBackend::SetShaderResult;
 
-	set_scan_and_pixel_doubling();
-	return true;
+	switch (GFX_GetRenderer()->SetShader(descriptor)) {
+	case ShaderError: return false;
+
+	case PresetError:
+		NOTIFY_DisplayWarning(Notification::Source::Console,
+		                      "RENDER",
+		                      "RENDER_DEFAULT_SHADER_PRESET_FALLBACK",
+		                      descriptor.c_str(),
+		                      ShaderName::Sharp);
+
+		set_scan_and_pixel_doubling();
+		return true;
+
+	case Ok: set_scan_and_pixel_doubling(); return true;
+
+	default: assertm(false, "Invalid SetShaderResult value"); return false;
+	}
 }
 
 static void set_fallback_shader_or_exit()
@@ -1513,12 +1526,16 @@ static void notify_render_setting_updated(SectionProp& section,
 static void register_render_text_messages()
 {
 	MSG_Add("RENDER_SHADER_RENAMED",
-	        "Built-in shader [color=white]'%s'[reset] has been renamed to "
-	        "[color=white]'%s'[reset]; using [color=white]'%s'[reset]");
+	        "Built-in shader [color=white]'%s'[reset] has been renamed to [color=white]'%s'[reset];\n"
+	        "using [color=white]'%s'[reset]");
 
 	MSG_Add("RENDER_SHADER_FALLBACK",
-	        "Error setting shader [color=white]'%s'[reset], "
+	        "Error setting shader [color=white]'%s'[reset],\n"
 	        "falling back to [color=white]'%s'[reset]");
+
+	MSG_Add("RENDER_DEFAULT_SHADER_PRESET_FALLBACK",
+	        "Error setting shader preset [color=white]'%s'[reset],\n"
+	        "falling back to default preset");
 }
 
 void RENDER_AddConfigSection(const ConfigPtr& conf)
--- a/src/gui/render/render_backend.h
+++ b/src/gui/render/render_backend.h
@@ -48,7 +48,9 @@ public:
 	//
 	// Similarly, `sharp` is mapped to `interpolation/sharp.glsl`, etc.
 	//
-	virtual bool SetShader(const std::string& symbolic_name) = 0;
+	enum class SetShaderResult { Ok, ShaderError, PresetError };
+
+	virtual SetShaderResult SetShader(const std::string& symbolic_name) = 0;
 
 	// Can be a no-op if the backend does't support shaders.
 	virtual bool MaybeAutoSwitchShader(const DosBox::Rect canvas_size_px,
--- a/src/gui/render/sdl_renderer.cpp
+++ b/src/gui/render/sdl_renderer.cpp
@@ -250,13 +250,14 @@ bool SdlRenderer::UpdateRenderSize(const int render_width_px, const int render_h
 	return true;
 }
 
-bool SdlRenderer::SetShader([[maybe_unused]] const std::string& shader_name)
+SdlRenderer::SetShaderResult SdlRenderer::SetShader(
+        [[maybe_unused]] const std::string& shader_name)
 {
 	// no shader support; always report success
 	//
 	// If we didn't, the rendering backend agnostic fallback mechanism would
 	// fail and we'd hard exit.
-	return true;
+	return SetShaderResult::Ok;
 }
 
 bool SdlRenderer::MaybeAutoSwitchShader([[maybe_unused]] const DosBox::Rect canvas_size_px,
--- a/src/gui/render/sdl_renderer.h
+++ b/src/gui/render/sdl_renderer.h
@@ -33,7 +33,7 @@ public:
 	bool UpdateRenderSize(const int new_render_width_px,
 	                      const int new_render_height_px) override;
 
-	bool SetShader([[maybe_unused]] const std::string& shader_name) override;
+	SetShaderResult SetShader([[maybe_unused]] const std::string& shader_name) override;
 
 	bool MaybeAutoSwitchShader([[maybe_unused]] const DosBox::Rect canvas_size_px,
 	                           [[maybe_unused]] const VideoMode& video_mode) override;


shermp:
I'll admit I haven't looked at this in depth yet, but I have a question @johnnovak 

Is the intention to allow the end user to modify the presets? If not, why the INI files?

johnnovak:
> I'll admit I haven't looked at this in depth yet, but I have a question @johnnovak
> 
> Is the intention to allow the end user to modify the presets? If not, why the INI files?

- Eventually, we could allow them to create their own presets, yes.
- It's cleaner than hardcoding a bunch of values (I prefer using resource files, personally).
- Most importantly, it allows me (or any shader developer/tweaker) to tweak settings at runtime by modifying the ini file, then hitting the "reload shader" hotkey. You simply cannot develop shader settings/presets if you need to recompile the exe, that's a horrific workflow. We need instant feedback.
- People who like to tweak things can even tweak the auto-CRT presets if they feel so inclined.

All in all, the INI loading is like 20 lines—it's a small piece of the solution.

I said in the description above that this is an internal-only feature. I think I'll make a few more enhancements in a few weeks' time so presets are available to the end user too (it would make sense with some of our upscaler shaders, for instance; I've added preset support to them).

shermp:
Yeah, developer tweaking at runtime is a very good reason. Carry on!

src/gui/CMakeLists.txt
```diff
@@ -14,3 +14,5 @@ target_sources(libdosboxcommon PRIVATE
 target_link_libraries(libdosboxcommon PRIVATE
   OpenGL::GL
   $<IF:$<BOOL:${C_OPENGL}>,libglad,>)
+
+target_link_libraries(libdosboxcommon PRIVATE simpleini)
```
weirddan455, line 18: I'm pretty sure you can remove all of these CMake changes you did. It's a header only library. There's no compilation unit for CMake to compile and link.

weirddan455:
Minor bug. You added an error to DOS console on invalid shader that's not present in main. That's a nice touch but it makes a bit of a mess of the welcome banner if you set something invalid in the .conf file.

shader

src/gui/CMakeLists.txt
johnnovak: Yeah, probably. @kklobe added these for the simde stuff, so I followed the same pattern.

I guess I can remove it for both simde and this; I'm not too sure if there's any drawback to removing it (probably isn't).

johnnovak:
> Minor bug. You added an error to DOS console on invalid shader that's not present in main. That's a nice touch but it makes a bit of a mess of the welcome banner if you set something invalid in the .conf file.
> shader

Nice find. The fix is to ensure these errors are printed after the banner. We definitely want notifications to the console in these cases, so it's an init order problem.

src/gui/render/opengl_renderer.cpp
```diff
@@ -386,7 +388,6 @@ void OpenGlRenderer::PresentFrame()
 {
 	glClear(GL_COLOR_BUFFER_BIT);
 
-	frame_count++;
```
weirddan455, line 389: Explanation for this change? Just want to make sure this isn't a mistake that snuk in as part of a large commit.

src/gui/render/opengl_renderer.cpp
```diff
@@ -403,37 +404,59 @@ void OpenGlRenderer::PresentFrame()
...
 	            static_cast<GLfloat>(draw_rect_px.w),
 	            static_cast<GLfloat>(draw_rect_px.h));
 
-	glUniform1i(uniform.frame_count, frame_count);
+	glUniform1i(uniforms.frame_count, frame_count);
+
+	for (const auto& [uniform_name, value] : current_shader_preset.params) {
+		if (uniforms.params.contains(uniform_name)) {
+			const auto location = uniforms.params[uniform_name];
+			glUniform1f(location, value);
```
weirddan455: You can do this to avoid the 2nd hash lookup you're doing here:

```cpp
if (auto location = uniforms.params.find(uniform_name); location != uniforms.params.end()) {
    glUniform1f(location->second, value);
} else {
    LOG_WARNING("You've been warned");
}
```
Also, aren't these preset uniforms constants? This function gets called at least once every frame so it's a bit wasteful if so.

src/gui/render/opengl_renderer.cpp
```diff
@@ -627,33 +651,46 @@ std::optional<GLuint> OpenGlRenderer::BuildShaderProgram(const std::string& sour
...
-	return UpdateRenderSize(render_width_px, render_height_px);
+	current_shader_descriptor_string = shader_descriptor;
+	return true;
 }
 
 bool OpenGlRenderer::ForceReloadCurrentShader()
 {
+	assert(shader_cache.contains(current_shader.info.name));
+
+	const auto shader = shader_cache[current_shader.info.name];
```
weirddan455: I believe this is doing a deep copy of a fairly hefty type that has a `std::unordered_map` in it. Should probably take a reference here.

src/gui/render/opengl_renderer.cpp
```diff
@@ -664,18 +701,35 @@ bool OpenGlRenderer::MaybeAutoSwitchShader(const DosBox::Rect canvas_size_px,
...
+	const auto changed_shader = (curr_descriptor.shader_name !=
+	                             new_descriptor.shader_name);
+	if (changed_shader) {
+		if (!SwitchShader(new_descriptor.shader_name)) {
+			return false;
+		}
 	}
 
-	return SwitchShader(new_shader_name);
+	const auto shader_info = shader_cache[new_descriptor.shader_name].info;
```
weirddan455: Another deep copy of a hash map (this variable is not even used).

src/gui/render/opengl_renderer.cpp
```diff
@@ -688,30 +742,100 @@ bool OpenGlRenderer::SwitchShader(const std::string& shader_name)
...
+
+	auto set_default_preset = [&]() {
+#ifdef DEBUG_OPENGL
+		LOG_DEBUG("OPENGL: Using default shader preset",
+		          descriptor.ToString().c_str());
+#endif
+		assert(shader_cache.contains(descriptor.shader_name));
+		auto default_preset = shader_cache[descriptor.shader_name].info.default_preset;
+
+		current_shader_preset = default_preset;
```
weirddan455: Two deep copies of hash maps.

src/gui/render/opengl_renderer.cpp
```diff
@@ -688,30 +742,100 @@ bool OpenGlRenderer::SwitchShader(const std::string& shader_name)
...
+	}
+}
+
+std::optional<ShaderPreset> OpenGlRenderer::GetOrLoadAndCacheShaderPreset(
+        const ShaderDescriptor& descriptor)
+{
+	assert(!descriptor.shader_name.empty());
+
+	assert(shader_cache.contains(descriptor.shader_name));
+	auto default_preset = shader_cache[descriptor.shader_name].info.default_preset;
```
weirddan455: Deep copy

src/gui/render/render.cpp
```diff
@@ -542,13 +542,28 @@ bool RENDER_MaybeAutoSwitchShader(const DosBox::Rect canvas_size_px,
                                   const VideoMode& video_mode,
                                   const bool reinit_render)
 {
+	const auto curr_preset = GFX_GetRenderer()->GetCurrentShaderPreset(); //-V821
+
 	const auto shader_changed = GFX_GetRenderer()->MaybeAutoSwitchShader(
 	        canvas_size_px, video_mode);
 
+	const auto new_preset = GFX_GetRenderer()->GetCurrentShaderPreset(); //-V821
+
```
weirddan455: Two more deep copies of these presets. `new_preset` should at least be inside the `if (shader_changed)` block.

weirddan455:
Another minor bug. Enter a valid shader with an invalid preset. Ex: `shader = crt/crt-hyllian:blah`. It will fallback to its default preset but if you type `shader` again to view what it is set to, it will return the invalid `crt/crt-hyllian:blah`.

src/gui/render/opengl_renderer.cpp
johnnovak: They're constants but sending them every frame is fine. A few method calls per frame is nothing.

The driver most likely puts these into a contiguous memory area and sends them to the GPU per frame.

Sending them "on demand" only would be an unnecessary microptimisation. We should only do that type of caching if it has measurable benefits. Sending them always is the easiest thing to do.

So don't forget most OpenGL calls just build up command buffers/data structures, then the driver sends them over in batches on demand / when necessary. In actual reality the driver has a small memory-mapped area to hold the uniforms, and the shaders read the values every single frame anyway whether we modify them or not (they _can_ change per frame, that's the crucial bit). We happen to always update them per frame in this memory area because the cost is insignifant and it makes the code straightfward. In fact, I was thinking of not even caching the uniform locations at all but simplifying this further at the code level and calling `glGetUniformLocation() & glUniform()` per frame for each uniform.

Simpler code == better code (when performance is not a concern).

Also, if there's no measurable difference, it's not wasteful.

src/gui/render/opengl_renderer.cpp
johnnovak: Good point, reference is better. Although, copying a few hundred bytes is nothing or even few Ks now and then.

src/gui/render/opengl_renderer.cpp
johnnovak: Yeah, once we took it from the map and assigned it to `current_shader_preset`, I want a copy, 100%.

I don't want to "micro-optimise" holding references to few hundred byte memory blocks, it's totally pointless... If we're dealing with megabytes, then it would make sense. Copy is the correct thing here.

I can use `auto&` on the temporary but that's about it. Compiler should be able to optimise the copy out, I write code for readability first.

src/gui/render/render.cpp
johnnovak: Doesn't matter. Take the size of the structures and the frequency of the calls into account. A copy is preferable almost every single time in such cases.

src/gui/render/opengl_renderer.cpp
johnnovak: By design, it's fine. Well, I guess I could make it a ref then copy on the next line. But man, this is low level shit the compiler should be able to figure out, and even if it copies, it does not matter.

Let's not micro-optimise for nothing 😄 

I could make it `auto&` in this case easily, though.

johnnovak:
> Another minor bug. Enter a valid shader with an invalid preset. Ex: `shader = crt/crt-hyllian:blah`. It will fallback to its default preset but if you type `shader` again to view what it is set to, it will return the invalid `crt/crt-hyllian:blah`.

Yeah, that's harder to fix, so I figured I'll leave it like that. Very edge casey territory. Maybe I'll fix it one day. I might give it another go but from memory it was too complicated to handle that edge case.

src/main.cpp
```diff
@@ -510,8 +510,14 @@ static void maybe_create_resource_directories()
...
-	const auto glshaders_dir = get_config_dir() / GlShadersDir;
-	try_create_resource_dir(glshaders_dir);
+	const auto shaders_dir = get_config_dir() / ShadersDir;
+	try_create_resource_dir(shaders_dir);
+
+	if (create_dir(shaders_dir, 0700, OK_IF_EXISTS) != 0) {
+		LOG_WARNING("CONFIG: Can't create directory '%s': %s",
+		            shaders_dir.string().c_str(),
+		            safe_strerror(errno).c_str());
+	}
```
weirddan455: This doesn't need to be here. This is a duplicate of the `try_create_resource_dir()` lambda. Maybe you copy+pasted it by mistake.

shermp: The lambda is very recent, I only just created it in the `soundcanvas_rom_dir` PR. So probably a leftover from starting this work earlier than that.

src/gui/render/opengl_renderer.cpp
johnnovak: It's incremented in `PrepareFrame()` so we were double-incrementing it in `PresentFrame()` as well.

src/gui/render/opengl_renderer.cpp
johnnovak: Using `auto&` on the temporary now.

src/gui/render/render.cpp
johnnovak: Moved it inside the `if`.

src/gui/render/opengl_renderer.cpp
johnnovak: Done.

src/main.cpp
johnnovak: Ah yeah, rebase error or something. Fixed.

src/gui/render/opengl_renderer.cpp
johnnovak: So this hash table is always only gonna contain less than 20 entries, usually less than 10, so this is heavy micro-optimisation territory, but I took your suggestion this time 😆

src/gui/CMakeLists.txt
johnnovak: Okay, using an interface library seems to be the recommended way for header-only libs, so leaving this as-is (and for simde as well).

https://stackoverflow.com/a/60604878

johnnovak:
@weirddan455 Alright, I've addressed the default preset fallback issue and the banner plus a few minor things. The former was surprisingly hard, and we're simply not printing any notifications to the console before we enter the command line interpreter loop (the banner is printed before that).

src/shell/shell.cpp
```diff
@@ -477,17 +479,28 @@ void DOS_Shell::Run()
...
 			RunBatchFile();
 		} else {
-			if (echo) ShowPrompt();
+			if (echo) {
+				ShowPrompt();
+			}
 			InputCommand(input_line);
 			ParseLine(input_line);
 		}
 	}
```
weirddan455, line 496: This fixes the welcome banner bug I mentioned. It's probably out of scope for this PR but I just thought of a concern regarding the notification system as whole.

What if a notification happens while a user is playing a game? Can writing to the console cause graphical glitches in game? Particularly if it's a text based game?

Here we're never setting `is_shell_running = false` (technically this is correct since the shell is still running in the background). We may need a way to detect if the user is actually sitting at the command prompt or has a game/application running though.

johnnovak: It's good you're thinking about these things 😄 

It is not possible to change config settings while a program is running, so for now we're fine.

When the OSD is introduced, most invalid settings or setting combinations will be blocked by the UI, but there's still room for runtime error. I think I should adjust my description in `notifications.h` to account for that:

https://github.com/dosbox-staging/dosbox-staging/blob/main/src/misc/notifications.h#L102

So yeah, if let's say the user tries to load a malformed SoundFont via the OSD, nothing should be printed to the console (well, we might be in a graphics mode anyway), but they should be informed about the error via an OSD notification somehow.

Also related, we'll need to lock down a lot of settings in the OSD when a program is running (e.g., we must not allow reconfiguring sound devices like changing `sbtype` or disabling the GUS when not in the DOS prompt; these settings must be greyed out).

weirddan455: > It is not possible to change config settings while a program is running, so for now we're fine.

Yes but things like MT-32 messages use the notification system correct? I can't remember what all you migrated over to the new notification system but I imagine we might also want to use this for things like screenshots, CPU cycle changes via hotkeys, shader changes, etc.

src/gui/render/opengl_renderer.cpp
```diff
@@ -673,18 +718,35 @@ bool OpenGlRenderer::MaybeAutoSwitchShader(const DosBox::Rect canvas_size_px,
...
+{
+	const auto changed_shader = (curr_descriptor.shader_name !=
+	                             new_descriptor.shader_name);
+	if (changed_shader) {
+		if (!SwitchShader(new_descriptor.shader_name)) {
+			return false;
+		}
+	}
+
+	const auto& shader_info = shader_cache[new_descriptor.shader_name].info;
```
weirddan455: Remove this line. It's an unused variable. Not sure why none of our CI checks caught this.

src/shell/shell.cpp
johnnovak: > > It is not possible to change config settings while a program is running, so for now we're fine.
> 
> Yes but things like MT-32 messages use the notification system correct?

Not yet:

https://github.com/dosbox-staging/dosbox-staging/blob/main/src/midi/mt32.cpp#L575-L588

> I can't remember what all you migrated over to the new notification system

I've only migrated some of the user-initiated warnings triggered by console commands, nothing else.

> but I imagine we might also want to use this for things like screenshots, CPU cycle changes via hotkeys, shader changes, etc.

Yep, that's the plan. I've thought a lot about this (months 😅) and I have identified all the possible use cases. The notification system currently only handles the `Console` type from here:

https://github.com/dosbox-staging/dosbox-staging/blob/main/src/misc/notifications.h#L124-L136

With the introduction of the OSD, I'll add support for the `Event`, `Gui`, and `Hotkey` types. Here are few mockup images I created to illustrate how these OSD notifications would look like:

https://github.com/dosbox-staging/dosbox-staging/discussions/1953

I'm intending to add support for OSD notifications in 0.84.0 next year (so notifications when you press a hotkey, MT-32 messages, etc.) Then next step will be the full GUI.

I recommend reading the entire `notifications.h`; I'm going into great detail there in the comments on how the whole thing should hang together.

src/gui/render/opengl_renderer.cpp
johnnovak: Done.

johnnovak:
All green now @weirddan455.

weirddan455:
I'm giving it a final review now.

src/gui/render/opengl_renderer.cpp
```diff
@@ -636,33 +657,58 @@ std::optional<GLuint> OpenGlRenderer::BuildShaderProgram(const std::string& sour
...
-	if (!SwitchShader(new_actual_shader_name)) {
-		return false;
+	if (!new_descriptor.preset_name.empty() &&
+	    current_shader_descriptor.preset_name.empty()) {
+
+		current_shader_descriptor_string = current_shader_descriptor.shader_name;
+
+		// We could set the shader but not the preset.
+		return PresetError;
 	}
```
weirddan455, line 691: > Another minor bug. Enter a valid shader with an invalid preset. Ex: shader = crt/crt-hyllian:blah. It will fallback to its default preset but if you type shader again to view what it is set to, it will return the invalid crt/crt-hyllian:blah.

I think this is the code you added to fix this bug. It doesn't always work though. Ex: `shader = crt-auto:blah`. The preset gets discarded by ShaderManager and never ends up in the new descriptor.

src/gui/render/opengl_renderer.h
```diff
@@ -124,14 +136,36 @@ class OpenGlRenderer : public RenderBackend {
...
+	ShaderDescriptor current_shader_descriptor = {};
+
+	// Current shader descriptor string as set by the user (e.g., if the
+	// user set `crt-auto`, this will stay `crt-auto`; it won't be synced to
+	// the actual shader & preset combo in use, such as
+	// `crt/crt-hyllian:vga-4k`).
+	//
+	// Might contain the .glsl file extension if set by the user.
+	//
+	std::string current_shader_descriptor_string = {};
```
weirddan455, line 168: There's a lot of similarly named variables here that got me confused a few times as I was reading through.

In most cases, you use the descriptor from `ShaderManager::GetCurrentShaderDescriptor()` but then you also store a copy here.

weirddan455 (approved):
I'll go ahead and approve this. Just the one minor bug left that I noticed but it's hardly a show stopper.

Also, not new to this PR but I found the role of `ShaderManager` confusing at first. It has functions like `MaybeAutoSwitchShader()` which is also an identically named function in `OpenGlRenderer` so it makes you think "who is actually switching the shaders here?" Well, `OpenGlRenderer` is the one actually doing the work.

I also don't really like this pattern:

```cpp
shader_manager.NotifyShaderChanged(shader_descriptor, GlslExtension);
const auto new_descriptor = shader_manager.GetCurrentShaderDescriptor();
```
Really all this is doing is resolving the "auto" shaders. I would consider having a function like `ShaderDescriptor ShaderManager::ResolveAutoShaders(std::string &description, std::string &extension);` that returns you the resolved descriptor.

Anyway, that stuff is out of scope for this PR. Just some feedback as I was reading through these classes I hadn't spent much time in before.

src/gui/render/opengl_renderer.h
johnnovak: Yeah, it gets complicated because we need sometimes to differentiate between the user specified descriptor (e.g., `shader = crt-auto`), and the actual resolved descriptor (`crt-auto` can map to any of the presets, e.g., `crt-hyllian:vga-4k` or `crt-hyllian:ega-1080p`, etc.)

I guess I could rename a few of these to "resolved", I'll have a quick look.

Ultimately, it's not the simplest code, it will only makes 100% sense if you spend some time working with it (as in making changes to the code, trace logging what's really going on with the flow, etc.)

johnnovak:
> I'll go ahead and approve this. Just the one minor bug left that I noticed but it's hardly a show stopper.

Thanks for the review, man 👍🏻 

> Also, not new to this PR but I found the role of `ShaderManager` confusing at first. It has functions like `MaybeAutoSwitchShader()` which is also an identically named function in `OpenGlRenderer` so it makes you think "who is actually switching the shaders here?" Well, `OpenGlRenderer` is the one actually doing the work.

`MaybeAutoSwitchShader()` is public in `OpenGlRenderer` and an implementation detail (private method) in `ShaderManager` -- an important distinction. But I can see the word "switch" has different meaning in the two classes which might be confusing at first.

> I also don't really like this pattern:
> 
> ```c++
> shader_manager.NotifyShaderChanged(shader_descriptor, GlslExtension);
> const auto new_descriptor = shader_manager.GetCurrentShaderDescriptor();
> ```
> Really all this is doing is resolving the "auto" shaders. I would consider having a function like `ShaderDescriptor ShaderManager::ResolveAutoShaders(std::string &description, std::string &extension);` that returns you the resolved descriptor.

There's a lot more going on, not just resolution. `ShaderManager` maintains its own state about the current shader as well, so there's some state-based logging on shader switches, etc. We could push that code around elsewhere, but it would be just a different type of complex (or ditch logging, sure, that would simplify it, but I want to keep the logs). Have you read this comment? I wrote that to clarify things a bit.

https://github.com/dosbox-staging/dosbox-staging/blob/main/src/gui/private/shader_manager.h#L111-L125

It's a bit complex how things hang together, I'll give you that, but this is my 3rd or 4th almost total rewrite over the past few weeks. It was a lot worse and more tangled before 😆 I'm sure equivalent functionality could be done in 10 other different ways, but I'm afraid we'd just be pushing code around, and the overall complexity won't be significantly less.

> Anyway, that stuff is out of scope for this PR. Just some feedback as I was reading through these classes I hadn't spent much time in before.

Yeah, that's good feedback. I'm afraid some things are just complicated. Maybe over time I'll have an "a-ha" moment and will manage to simplify it further, but currently, I'm not seeing it.

src/gui/render/opengl_renderer.cpp
johnnovak: Lol, you just can't always win 😆 I'll probably rethink the whole descriptor handling in the future because I can only think of bandaid solutions to this special case. I'd just leave it as-is for now.

Torinde:
> This fixes the welcome banner bug I mentioned.

Is the warning still issued after the banner? Or it's only recorded in the log?

johnnovak:
> > This fixes the welcome banner bug I mentioned.
> 
> Is the warning still issued after the banner? Or it's only recorded in the log?

Only in the log. You can try it out and see 😄

Issue #4514, opened 2025-09-17
Modernise our use of OpenGL


johnnovak:
Currently, our OpenGL backend doesn't request a specific OpenGL profile version but uses archaic fixed pipeline calls plus shaders. This is kinda sorta compatibility mode, but not explicitly requested, so we're at the mercy of the driver. Apparently, OpenGL drivers are quite lenient so this works. If it didn't, lots of messily coded old OpenGL apps would not work, but there's no good reason to rely on such driver-level compatibility hacks. The current fixed pipeline method would only matter if we were targeting 2005 or earlier GPUs, which we clearly aren't due to our minimum OS requirements.

So the plan is to modernise or OpenGL backend and target OpenGL core profile 3.3 / OpenGL ES 2.0. Adopting modern OpenGL will give us more explicit control, potentially resulting in performance improvements and less driver-dependent quirky behaviour. 

macOS supports up to core profile 4.1, and Raspberry Pi 4 or later OpenGL ES 3.1 which corresponds to OpenGL core profile 4.3 (those versions introduced compute shaders).

Basically all GPUs on Windows released in the last 10-12 years support core profile 4.6, so macOS is the "bottleneck". That's in quotes because our usage of OpenGL is extremely minimal, so OpenGL core profile 3.3 / OpenGL ES 2.0+ will be sufficient for our needs in practice. But technically we could use OpenGL core profile 4.1 features if we wanted to.

Besides, if one really needs 4.6 functionality, one's better off with DirectX 12 or Vulkan. But that's AAA game territory, not us.

### Depends on

- https://github.com/dosbox-staging/dosbox-staging/issues/

weirddan455:
The fixed function pipeline calls should be all gone now. I believe this was from one of your PRs where you merged GranMinigun's improvements. Commit 78b5bf9429e55c4cf6b30e475a68f68a3b9707ae

We also are explicitly requesting OpenGL 2.1 compatibility mode. I'm not against bumping the version but I'm not convinced it will change much unless you intend to use some of the newer functionality.

https://github.com/dosbox-staging/dosbox-staging/blob/c0d65a44fbd615ccdd9ec971f34ac90da4825f6f/src/gui/sdlmain.cpp#L1166-L1171

johnnovak:
[...]

Issue #926, opened 2021-03-04
OpenGLES support for shaders


SupervisedThinking:
**Describe the solution you'd like**
Since DOSBox always nags about `CONFIG: Unknown option glshader` on my SBCs... how about adopting those patches in a somewhat similar way to add an OpenGLES renderer which allows to use shaders on ARM builds which mostly lack support of the OpenGL API.

https://www.raspberrypi.org/forums/viewtopic.php?t=110957
https://www.vogons.org/viewtopic.php?f=41&t=48303#p500777

The average GPU of SBCs is fast enough to render such stuff & scanlines would add a nice retro touch.

dreamer:
@SupervisedThinking these patches predate SDL2 and OpenGL glshader support, so they won't apply to our code :(

When we started work on importing SDL2 patch, it had OpenGLES bits for Android inside, but IIRC it was broken and unfinished - I removed those OpenGLES bits to speed up SDL2 migration back in the day (SDL2 texture with opengles gives equivalent level of support as those old bits).

OpenGLES will need to be reimplemented (perhaps some code from that patch can be salvaged, but I'm not sure how much - maybe 50%).

SupervisedThinking:
>edit There's https://github.com/aqualung99/dosbox-0.74-ES though - it claims to be SDL2 based, so that's promising, maybe we can salvage work from there.

That's the reason why I've pointed you there & looks like the changes are somewhat "recent":
https://github.com/aqualung99/dosbox-0.74-ES/commits/master?after=fb071dd58b47df3d244462e8fe65f3ef731f3837+34&branch=master

GranMinigun:
Slang and GL ES context are separate things, but yeah, both are on my roadmap.

johnnovak:
Now my view is that we'll target a subset of the intersection of OpenGL ES 3.1 and OpenGL core profile 4.1 in the upcoming reworked OpenGL renderer, so removing the "won't do" and "out of scope" labels to avoid confusion.

Grounded0:
@johnnovak 

What's the minimum OpenGL feature level required to run upcoming 0.83.0? Is it still OpenGL 2.1+ or has that been bumped up?

johnnovak:
> [@johnnovak](https://github.com/johnnovak)
> 
> What's the minimum OpenGL feature level required to run upcoming 0.83.0? Is it still OpenGL 2.1+ or has that been bumped up?

[...]

PR #4591, opened 2025-11-11
Use OpenGL 3.3 Core Profile in the OpenGL render backend

johnnovak:
# Description

Currently, our OpenGL backend uses the OpenGL 2.1 compatibility profile. That kinda works, but it's archaic and definitely not the best option in 2025. I have plans to add a multipass shader pipeline and all sorts of render backend improvements, so coding for an archaic version of OpenGL from the early 2000s is just silly.

First of all, most good modern OpenGL learning resources teach Core Profile 3.3 and upwards (where "modern" means from the past 10-15 years). Secondly, coding for a "legacy but still tolerated" OpenGL version is just asking for trouble. Drivers might do funky things in compatibility mode, and the code has a tendency to become more complicated than necessary (e.g., several mandatory core profile 3.3 features were implemented as optional extensions in earlier non-core versions). Core Profile 3.3 and upwards is also more explicit than earlier versions, meaning we're less at the mercy of the driver to "do the correct thing". When using core profiles, there's less guesswork involved on the driver's side.

As for OpenGL version support across our official platforms, macOS added full Core Profile 4.1 support in 2011. On the Windows & Linux front, GPUs released in the last 10-12 years all support Core Profile 4.6. So the minimum Core Profile 3.3 requirement is rather modest, and it's 100% met on our supported OSes (last three major OS revisions, in general). TL;DR virtually all GPUs released from 2010 onwards that are not complete garbage support Core Profile 3.3 or better (generally at least 4.1).

Now, this change might break the OpenGL backend on the Raspberry Pi boards, but that's **_not a problem_** because:

- The Pi is for all intents and purposes an embedded device with direct support for OpenGL ES 3.1 only. We target desktop computers, _not_ mobile phones and embedded devices. The Pi driver seems to allow requesting a non-ES compatibility context, but as far as I know Core Profile 3.3 is not supported directly (however, there might be some emulation layers available. or maybe setting `MESA_GL_VERSION_OVERRIDE=3.3` helps—to be investigated by Pi enthusiasts).

- Even if we targeted OpenGL ES 3.1 directly, the GPU of the Pi is extremely weak, equivalent to a desktop GPU from the early to mid 2000s _at best_. It simply cannot run our auto CRT shaders at adequate FPS, and it barely works with our dead simple `sharp` shader at 1080p (based on user feedback). We don't want to design our render backend around an embedded device with an extremely weak and dated GPU—we target desktop systems and modern(ish) GPUs.

- The Pi was popular 8-10 years ago in emulator circles when it cost around $35, but the interest has apparently fizzled out recently as the Raspberry company "pivoted" to primarily target industrial customers. The price of the more recent Pi boards also went up significantly, so today you can buy a much more powerful laptop or second-hand mini-PC for emulator usage for the same amount of money. To be blunt about it, I personally refuse to spend tens or hundreds of hours on a separate OpenGL ES backend or attempt to shoehorn both Core Profile and ES support into the same backend just because some people are unwilling to buy a cheap second-hand laptop or mini-PC for around 100 bucks...

- However, this doesn't mean existing Pi users cannot use our software going forward. They can keep using the SDL texture backends, which aren't going anywhere (`output = texture` or `texturenb`).

- Same comment for any old device that doesn't a GPU with Core Profile 3.3 support (Note that modern apps flat-out require Core Profile support and if your GPU is too old, well, that's too bad. Get some better hardware. For example, latest Blender requires Core Profile 4.1 or higher).
- Same comment for any GPU that has garbage OpenGL drivers.

- More specifically, the Broadcom VideoCoreVII GPU in the Pi 5 is capable of **~77 GFLOPS.** That's about GeForce 6800 GS equivalent performance -- yep, that's a GPU from 2006 😆
  - GeForce GT 240 from 2009, **260 GFLOPS**.
  - GeForce GTX 750 Ti from 2014 (my previous card that was "mid" even in 2015 when I bought it), **1300 GFLOPS**.
  - GeForce 10 series (2016-2017), **600 to 12,000 (!) GFLOPS**.
  - Then the more recent RTX card are measured in low single digit to **100 TFLOPS** (that's **terra**flops, yikes! 😎 🤘🏻)
  - Best for last (the garbage-tier on-board Intel GPUs everyone hates):
    - Intel HD 3000 from 2011, **125 GFLOPS**
    - Intel HD 4000 from 2012, **269 GFLOPS**
  
- So when I'm saying the Pi 5's GPU is piss weak even compared to ancient pre-2010 standards, I think I'm right on the money 😂 People have created a bunch of shaders that run at acceptable framerates even on the Pi (e.g, `crt-potato` 🤣). These are very simple and low quality and thus pointless to include for desktop use. In theory, someone could add an OpenGL ES backend to Staging and port those shaders. What I'm trying to drive home, it's rather pointless to even try to run our current set of built-in shaders on the Pi.

- But instead of fixating on OpenGL ES and the Pi, it would be much more fruitful to focus our efforts on adding Vulkan support and backend-agnostic shader support (perhaps via [SDL_shadercross](https://github.com/libsdl-org/SDL_shadercross)). More recent Pi boards have [Vulkan 1.3 support](https://www.vulkan.org/news/auto-22842-c5c77c2da002ef993800251296a06bf3), so probably they would benefit from the Vulkan backend as well (but of course the sever performance limitations would still apply, but at the very least we could support simplistic shaders without having to maintain an OpenGL ES backend).

## Related issues

- Implements https://github.com/dosbox-staging/dosbox-staging/issues/4514
- https://github.com/dosbox-staging/dosbox-staging/issues/926 (rejected)

# Release notes

- Starting from this release, an OpenGL 3.3 Core Profile compatible GPU is required for the OpenGL backend. If you're on a Raspberry Pi or a device that doesn't have functional OpenGL drivers, please use the SDL texture renderer (`output = texture` or `texturenb` for the "non-blurry" version), or an earlier DOSBox Staging version.

# Manual testing

- Test that all shaders work on macOS and Windows.
- Ran the Quake bench on Windows; no difference between `main` and this branch (~108 FPS). Repeated the test with `crt-auto`, `sharp`, and `interpolation/catmull-rom`.

- [x] Windows
- [x] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/resources/shaders/interpolation/catmull-rom.glsl
+++ b/resources/shaders/interpolation/catmull-rom.glsl
@@ -18,8 +18,6 @@
  *	more details
  */
 
-#pragma use_srgb_texture
-#pragma use_srgb_framebuffer
 #pragma force_single_scan
 #pragma force_no_pixel_doubling
 
@@ -47,6 +45,35 @@ out vec4 FragColor;
 
 uniform vec2 rubyTextureSize;
 uniform sampler2D rubyTexture;
+
+#define GAMMA             2.2
+#define GAMMA_IN(color)   pow(color, vec4(GAMMA))
+#define GAMMA_OUT(color)  pow(color, vec4(1.0 / GAMMA))
+
+vec4 texture_linear(in sampler2D sampler, in vec2 uv)
+{
+	// subtract 0.5 here and add it again after the floor to centre the texel
+	vec2 texCoord = uv * rubyTextureSize - vec2(0.5);
+
+	vec2 s0t0 = floor(texCoord) + vec2(0.5);
+	vec2 s0t1 = s0t0 + vec2(0.0, 1.0);
+	vec2 s1t0 = s0t0 + vec2(1.0, 0.0);
+	vec2 s1t1 = s0t0 + vec2(1.0);
+
+	vec2 invTexSize = 1.0 / rubyTextureSize;
+
+	vec4 c_s0t0 = GAMMA_IN(texture(sampler, s0t0 * invTexSize));
+	vec4 c_s0t1 = GAMMA_IN(texture(sampler, s0t1 * invTexSize));
+	vec4 c_s1t0 = GAMMA_IN(texture(sampler, s1t0 * invTexSize));
+	vec4 c_s1t1 = GAMMA_IN(texture(sampler, s1t1 * invTexSize));
+
+	vec2 weight = fract(texCoord);
+
+	vec4 c0 = c_s0t0 + (c_s1t0 - c_s0t0) * weight.x;
+	vec4 c1 = c_s0t1 + (c_s1t1 - c_s0t1) * weight.x;
+
+	return (c0 + (c1 - c0) * weight.y);
+}
 
 void main()
 {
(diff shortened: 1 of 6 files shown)


src/libs/glad/README.md
```diff
@@ -1,20 +1,11 @@
...
-- Enable function pointer aliasing
-- Extensions:
-  - GL_ARB_framebuffer_sRGB
-  - GL_EXT_bgra
-
-Use this permalink to re-generate the loader with different parameters (e.g.,
-to add extensions):
-
-https://gen.glad.sh/#generator=c&api=gl%3D2.1&profile=gl%3Dcompatibility%2Cgles1%3Dcommon&extensions=GL_ARB_framebuffer_sRGB%2CGL_EXT_bgra&options=ALIAS%2CALIAS
+parameters.
```
FeralChild64: > with the following parameters.

This is unclear - the list of parameters is not here.

johnnovak: Indeed. Fixed.

FeralChild64:
I have noticed the `catmull-rom` shader now produces a very dark image.

- I am switching the shader at runtime - my default is `crt-auto-machine`, but my bat file to start Windows 3.x executes the command:
`Z:\CONFIG.COM -set "shader=interpolation/catmull-rom"`
- tested under Linux, with the Intel Arc A770 drivers

johnnovak:
> I have noticed the `catmull-rom` shader now produces a very dark image.
> 
>   * I am switching the shader at runtime - my default is `crt-auto-machine`, but my bat file to start Windows 3.x executes the command:
>       `Z:\CONFIG.COM -set "shader=interpolation/catmull-rom"`
> 
>   * tested under Linux, with the Intel Arc A770 drivers

Ah, forgot the gamma encode step at the end (didn't test, ooops 😊). Fixed, thanks for spotting it.

arrowgent:
it is true that on rpi4 & rpi5 the use of
`MESA_GL_VERSION_OVERRIDE=3.3 command` forces compatibility even if not supported
some of it may actually be using software emulation on the CPU instead of pure GPU hardware
its hard to tell without some trace debugger in place.

i take no threat of comparing the low power and somewhat incompatible hardware choices by the raspberry pi boards as neglect.
some limitation comes from mesa drivers (linux) v3d/v3dv that cannot reproduce hardware calls on the closed source propriterary broadcom firmware.  some blame also goes to broadcom for not supporting or releasing information to do that. i am uncertain if the broadcom gpu is even capable of higher OGL/VK compatibility within the hardware itself

i also have an arm64 snapdragon 7c using the freedreno (ogl) and turnip (vk) mesa drivers which have much better compatibility, albeit **similar performance to the later pi broadcom** hardware
^ for that note, the new steam framework will be using an arm snapdragon 8 model using the same mesa drivers.  anyone actually playing arm64 dosbox in VR must be insane however.

the number of times ive tested GLES for various opengl3d software and found the performance results underwhelming.  unless expressly required i try not to use GLES.  i dont know if its a driver limitation or the implementation GLES vs OGL


even attempting to maintain "fluid" support for underperforming hardware or outdated becomes fairly moot.  progress must be made as the older systems and less used hardware becomes less commonly used for the newest software

as it sits, dosbox-staging technically doesnt even support arm64 (or armhf) in any amount, stated by yourself that only official releases (win, macos, linux `x86_amd64`) have any kind of support
even if technically such "mobile" arm devices can compile and run staging.
where i could compile staging on my pinephone with debian running arm64 mali with mesa panfrost/panvk driver, what would be the point on such a device?  for the fun of it im sure.


just to say, good work changing how staging calls OGL and requests a compatibility version.
and thoroughly noting the performance metrics of your baseline compatible support

Issue #4593, opened 2025-11-11
Mesa reporting OpenGL errors


weirddan455:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### What operating system are you using?

Linux x86_64

### If other OS or Linux, please describe

Arch Linux

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Are you able to trigger the issue with our official DOSBox Staging release?

- [x] I am running an official Staging release, not a release installed by my operating system's package manager, or a version repackaged by an external team.

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

I found that Mesa is reporting OpenGL errors. To reproduce, set this environment variable like so `export MESA_DEBUG=1` and then run Staging. This tells Mesa to print any problems to the terminal. Will not on Mac and Windows.

The first one happens during init:

```
2025-11-11 15:16:30.800 | RENDER: Using adaptive CRT shader based on the graphics standard of the video mode
2025-11-11 15:16:30.800 | RENDER: Auto-switched to shader 'crt/vga-1080p'
Mesa: error: GL_INVALID_OPERATION in glUniform(program not linked)
2025-11-11 15:16:30.809 | VGA: Double scanning VGA video modes enabled
2025-11-11 15:16:30.809 | VGA: Pixel doubling enabled
2025-11-11 15:16:30.812 | DISPLAY: VGA 720x400 16-colour text mode 03h at 70.087 Hz, scaled to 1067x800 pixels with 1:1.35 (20:27) pixel aspect ratio
```

The second one happened after launching Doom with the resolution change:

```
2025-11-11 15:16:45.723 | RENDER: Auto-switched to shader 'crt/vga-1080p-fake-double-scan'
2025-11-11 15:16:45.726 | VGA: Forcing single scanning of double-scanned VGA video modes
Mesa: error: GL_INVALID_OPERATION in glUniform(location=3)
2025-11-11 15:16:45.728 | DISPLAY: VGA 320x200 256-colour graphics mode 13h at 70.086 Hz, scaled to 1067x800 pixels with 1:1.2 (5:6) pixel aspect ratio
```

### Steps to reproduce the behaviour

Explain in detail how to reproduce the bug. Assume no familiarity with the software.
1. `export MESA_DEBUG=1`
2. First error happens during init.
3. Second error happens after launching Doom.

### Your configuration

[...]

PR #4594, opened 2025-11-12
Fix OpenGL uniform errors/warnings when using a debug context

johnnovak:
# Description

These are harmless errors/warnings raised when the uniform we're trying to set does not exist in the compiled shader, but it's easy to do things properly, so I figured why not.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4593

# Manual testing

Trivial change, Staging starts up on Windows and the shaders work.

- [x] Windows
- [ ] macOS
- [ ] Linux

# Checklist

I have:

### DIFF
--- a/src/gui/render/opengl_renderer.cpp
+++ b/src/gui/render/opengl_renderer.cpp
@@ -417,23 +417,38 @@ void OpenGlRenderer::GetUniformLocations()
 
 	uniform.frame_count = glGetUniformLocation(current_shader.program_object,
 	                                           "rubyFrameCount");
+
+	uniform.input_texture = glGetUniformLocation(current_shader.program_object,
+	                                             "rubyTexture");
 }
 
 void OpenGlRenderer::UpdateUniforms() const
 {
-	glUniform2f(uniform.texture_size,
-	            static_cast<GLfloat>(render_width_px),
-	            static_cast<GLfloat>(render_height_px));
+	if (uniform.texture_size > -1) {
+		glUniform2f(uniform.texture_size,
+		            static_cast<GLfloat>(render_width_px),
+		            static_cast<GLfloat>(render_height_px));
+	}
 
-	glUniform2f(uniform.input_size,
-	            static_cast<GLfloat>(render_width_px),
-	            static_cast<GLfloat>(render_height_px));
+	if (uniform.input_size > -1) {
+		glUniform2f(uniform.input_size,
+		            static_cast<GLfloat>(render_width_px),
+		            static_cast<GLfloat>(render_height_px));
+	}
 
-	glUniform2f(uniform.output_size,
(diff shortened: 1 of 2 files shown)


weirddan455:
This isn't the cause of the error. It's here:

https://github.com/dosbox-staging/dosbox-staging/blob/4d8652275e3d19852faadb1347287cc0f80c6720/src/gui/render/opengl_renderer.cpp#L621-L625

You need to call `glUseProgram(shader_program);` at some point before calling `glUniform1i();`

johnnovak:
I'll check with the GLAD debug context. Interesting that what I fixed don't show up as warnings for you, a few on the standard uniforms are not actually used in all shaders, so glGetUniforLocation returns -1 for them (not found).

weirddan455:
> I'll check with the GLAD debug context. Interesting that what I fixed don't show up as warnings for you

I don't know what you mean by "GLAD debug context". I just did `export MESA_DEBUG=1` and the errors get printed to the console but of course that's Linux specific for the most part.

> a few on the standard uniforms are not actually used in all shaders, so glGetUniforLocation returns -1 for them (not found).

I noticed that too so I'm sure the change you made is fine. OpenGL probably just ignores a -1 location anyway.

The error above I pointed out is more likely to be problematic. It can't set a uniform without using the program first. The first shader it loads fails due to "program not linked". With subsequent shaders it's worse because it can stomp on a uniform from the previous shader.

johnnovak:
> I don't know what you mean by "GLAD debug context". I just did `export MESA_DEBUG=1` and the errors get printed to the console but of course that's Linux specific for the most part.

When you generate the GLAD OpenGL wrappers/loader [here](https://gen.glad.sh/), you can tick the "debug" checkbug in the options section further down the page. This will generate a loader that instruments every single OpenGL call to check for errors and print them to the console.

In my WIP Core Profile PR I've added two version of GLAD: one for the release builds (no debug instrumentation), and one for debug builds (with debug instrumentation).

https://github.com/dosbox-staging/dosbox-staging/pull/4591/commits/e6de674b98feb858ab536f0e1ec4983e547bf38d

 
> > a few on the standard uniforms are not actually used in all shaders, so glGetUniforLocation returns -1 for them (not found).
> 
> I noticed that too so I'm sure the change you made is fine. OpenGL probably just ignores a -1 location anyway.

Yeah, it's more of a warning. 

> The error above I pointed out is more likely to be problematic. It can't set a uniform without using the program first. The first shader it loads fails due to "program not linked". With subsequent shaders it's worse because it can stomp on a uniform from the previous shader.

I agree, and I will look into it today. I just missed that one, was tired yesterday when I authored this PR.

johnnovak:
@weirddan455 I think it's good now.

Issue #4579, opened 2025-10-30
Hungarian keyboard layout detection and default Hungarian layout issues (0.83.0-alpha)


RevSnowfox:
### Are you using the latest DOSBox Staging version?

- [x] I have checked and I'm using the latest release.

### Different version than latest?

0.83.0-alpha

### What operating system are you using?

Windows 10

### Are you running DOSBox Staging on a native OS?

- [x] I am running Staging on a native OS

### Are you able to trigger the issue with our official DOSBox Staging release?

- [x] I am running an official Staging release, not a release installed by my operating system's package manager, or a version repackaged by an external team.

### Have you checked that no other similar issue already exists?

- [x] I have searched and have not found similar issues.

### A clear and concise description of what the bug is

DOSBox Staging 0.83.0-alpha has some issues with keyboard layout detection, specifically detecting that I have a Hungarian physical keyboard (ISO-style), and OS input set to Hungarian layout as active. I have some other input layouts enabled, but I rarely switch to them.

0.83.0-alpha defaults to the `uk` layout with codepage 437, the default IBM ASCII-US codepage. I feel like the expected behavior by default should be to match the active OS keyboard layout when Staging starts up.

0.82.2 however defaulted to the `hu` layout with codepage 3845. This also has its issues, but it was a better match.

- the `hu` layout expects a 101-key ANSI-style keyboard and it's QWERTY-based
- the `hu208` layout expects a 102-key ISO-style keyboard and it's QWERTZ-based
- codepage 852 (aka MS-DOS Latin-2) is the official IBM codepage for the Hungarian language introduced with MS-DOS
- codepage 3845 (aka CWI-2) is an unofficial encoding which was created to be legible even if decoded as plain old CP437

[...]

Issue #3233, opened 2023-12-18
KEYB.COM and log discrepancies


Torinde:
### Are you using the latest Dosbox-Staging Version?

- [X] I have checked releases and am using the latest release.

### Different version than latest?

dosbox-staging-windows-msys2-x86_64-v0.81.0-alpha-2226-g2dff7

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [X] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Testing latest build for the issues I mentioned at https://github.com/dosbox-staging/dosbox-staging/issues/3128#issuecomment-1825396175

1. LOG always says codepage is 3021, even if changed by KEYB.COM to 872 (and it's actually 872/855 - Cyrillic letters appears on the place of box characters that were 3021)
![image](https://github.com/dosbox-staging/dosbox-staging/assets/97228894/72844127-58c3-4d54-81f6-c579954242f1)
![image](https://github.com/dosbox-staging/dosbox-staging/assets/97228894/047e0ddf-556e-4603-ae01-d150d94cccf3)

2. KEYB bg442 complains about "file bg442 not found" (see above screenshot) and LOG has a message "DOS: The keyboard layout's country code 'bg442' does not have a corresponding country", but `KEYB bg442 855` reports success (although LOG gives the same warning) - but with codepage 872 (while LOG as always says 3021... so **user requests 855, KEYB states it provides 872, LOG states 3021 is used**, actually  is used 855 or 872 (I haven't checked with Euro sign which of the two)
![image](https://github.com/dosbox-staging/dosbox-staging/assets/97228894/520c512c-9fb4-4abb-9972-7aa80120d22c)
![image](https://github.com/dosbox-staging/dosbox-staging/assets/97228894/7f6d565d-2339-4466-8b44-7fbcce789f57)

3. On changes of codepage (but not keyboard layout) with KEYB - LOG doesn't get any message (expected was "LAYOUT: Loaded codepage XXX for detected language XX"), but maybe that's because of the "always 3021" effect (thus from this wrong perspective there was no change from certain command, thus nothing was logged?)

4. bg241 layout seems to have flipped shift/non-shift for the number keys - but since I don't recognize it at all (what's the origin/purpose of this layout?) I can't claim for sure - maybe the author wanted it like that:
[...]

PR #4597, opened 2025-11-15
Locale detection improvements

FeralChild64:
# Description

Main change: the keyboard layout autodetection is now influenced by the language settings:
- it tries to match the keyboard layout to the translation language (if non-English)
- for OSes where the user can't specify own keyboard layout priority, it then tries to match the keyboard layout to the GUI language (if non-English)

As before, the auto-detected keyboard layout is only selected from the ones configured in the host OS.

Additional changes in this PR:
- default code page for Hungarian keyboard layouts changes to 852
- Spanish/Portugal/Latin American keyboard layouts now have higher priority than English; I just feel this is the right choice, as these layouts are mostly standard QWERTY and other western-European layouts (like Italian) already have their priorities higher than English
- cleaned up DOS country code detection, now uses predefined string constants instead of hardcoded strings

Note: I have written unit tests for the new/refactored code, but I had to disable some of the checks, as checks involving `std::optional` do not seem to work correctly with the old GCC 12 we are using.

## Related issues

Fixes https://github.com/dosbox-staging/dosbox-staging/issues/4579

# Release notes

Keyboard layout autodetection is now influenced by the translation or host GUI language.

# Manual testing

Make sure `keyboard_layout = auto` in the config file.

Linux:
- Launch the emulator with `LANGUAGE="fr_FR" dosbox` / `LANGUAGE="de_DE" dosbox` command (if `language = auto` in the config file) or `dosbox -lang fr` / `dosbox -lang de` - if the French/German keyboard layout is configured in the host OS, it should be selected.

Windows, macOS:
- Only checked that keyboard layout detection still works (in my case, Polish layout is selected) - full testing would require changing the OS system language.

- [x] Windows
- [x] macOS
- [x] Linux

# Checklist

I have:

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -14,6 +14,7 @@ add_executable(dosbox_tests
     fraction_tests.cpp
     fs_utils_tests.cpp
     int10_modes_tests.cpp
+    language_territory_tests.cpp
     math_utils_tests.cpp
     mixer_tests.cpp
     port_containers_tests.cpp
--- a/tests/language_territory_tests.cpp
+++ b/tests/language_territory_tests.cpp
@@ -1,0 +1,280 @@
+// SPDX-FileCopyrightText:  2025-2025 The DOSBox Staging Team
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "misc/host_locale.h"
+#include "misc/support.h"
+
+#include <gtest/gtest.h>
+
+namespace {
+
+TEST(LanguageTerritory, Empty)
+{
+	const auto empty = LanguageTerritory("");
+
+	EXPECT_TRUE(empty.IsEmpty());
+
+	EXPECT_FALSE(empty.IsGeneric());
+	EXPECT_FALSE(empty.IsEnglish());
+	EXPECT_FALSE(empty.GetDosCountryCode());
+
+	EXPECT_TRUE(empty.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(empty.GetLanguageFiles().empty());
+}
+
+TEST(LanguageTerritory, Unknown)
+{
+	const auto unknown = LanguageTerritory("foo-BAR");
+
+	EXPECT_FALSE(unknown.IsEmpty());
+
+	EXPECT_FALSE(unknown.IsGeneric());
+	EXPECT_FALSE(unknown.IsEnglish());
+	EXPECT_FALSE(unknown.GetDosCountryCode());
+
+	EXPECT_TRUE(unknown.GetMatchingKeyboardLayouts().empty());
+
+	EXPECT_EQ(unknown.GetLanguageFiles(),
+	          std::vector<std::string>({"foo_BAR", "foo"}));
+}
+
+TEST(LanguageTerritory, InvalidFormat)
+{
+	const auto invalid_1 = LanguageTerritory("-");
+	const auto invalid_2 = LanguageTerritory(".FooBar");
+	const auto invalid_3 = LanguageTerritory("@FooBar");
+
+	EXPECT_TRUE(invalid_1.IsEmpty());
+	EXPECT_TRUE(invalid_2.IsEmpty());
+	EXPECT_TRUE(invalid_3.IsEmpty());
+
+	EXPECT_FALSE(invalid_1.IsGeneric());
+	EXPECT_FALSE(invalid_2.IsGeneric());
+	EXPECT_FALSE(invalid_3.IsGeneric());
+
+	EXPECT_FALSE(invalid_1.IsEnglish());
+	EXPECT_FALSE(invalid_2.IsEnglish());
+	EXPECT_FALSE(invalid_3.IsEnglish());
+
+	EXPECT_FALSE(invalid_1.GetDosCountryCode());
+	EXPECT_FALSE(invalid_2.GetDosCountryCode());
+	EXPECT_FALSE(invalid_3.GetDosCountryCode());
+
+	EXPECT_TRUE(invalid_1.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(invalid_2.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(invalid_3.GetMatchingKeyboardLayouts().empty());
+
+	EXPECT_TRUE(invalid_1.GetLanguageFiles().empty());
+	EXPECT_TRUE(invalid_2.GetLanguageFiles().empty());
+	EXPECT_TRUE(invalid_3.GetLanguageFiles().empty());
+}
+
+TEST(LanguageTerritory, Generic)
+{
+	const auto generic_1 = LanguageTerritory("c");
+	const auto generic_2 = LanguageTerritory("C");
+	const auto generic_3 = LanguageTerritory("posix");
+	const auto generic_4 = LanguageTerritory("POSIX");
+	const auto generic_5 = LanguageTerritory("PoSiX");
+
+	EXPECT_FALSE(generic_1.IsEmpty());
+	EXPECT_FALSE(generic_2.IsEmpty());
+	EXPECT_FALSE(generic_3.IsEmpty());
+	EXPECT_FALSE(generic_4.IsEmpty());
+	EXPECT_FALSE(generic_5.IsEmpty());
+
+	EXPECT_TRUE(generic_1.IsGeneric());
+	EXPECT_TRUE(generic_2.IsGeneric());
+	EXPECT_TRUE(generic_3.IsGeneric());
+	EXPECT_TRUE(generic_4.IsGeneric());
+	EXPECT_TRUE(generic_5.IsGeneric());
+
+	EXPECT_FALSE(generic_1.IsEnglish());
+	EXPECT_FALSE(generic_2.IsEnglish());
+	EXPECT_FALSE(generic_3.IsEnglish());
+	EXPECT_FALSE(generic_4.IsEnglish());
+	EXPECT_FALSE(generic_5.IsEnglish());
+
+	EXPECT_FALSE(generic_1.GetDosCountryCode());
+	EXPECT_FALSE(generic_2.GetDosCountryCode());
+	EXPECT_FALSE(generic_3.GetDosCountryCode());
+	EXPECT_FALSE(generic_4.GetDosCountryCode());
+	EXPECT_FALSE(generic_5.GetDosCountryCode());
+
+	EXPECT_TRUE(generic_1.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(generic_2.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(generic_3.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(generic_4.GetMatchingKeyboardLayouts().empty());
+	EXPECT_TRUE(generic_5.GetMatchingKeyboardLayouts().empty());
+
+	EXPECT_EQ(generic_1.GetLanguageFiles(), std::vector<std::string>({"en"}));
+	EXPECT_EQ(generic_2.GetLanguageFiles(), std::vector<std::string>({"en"}));
+	EXPECT_EQ(generic_3.GetLanguageFiles(), std::vector<std::string>({"en"}));
+	EXPECT_EQ(generic_4.GetLanguageFiles(), std::vector<std::string>({"en"}));
+	EXPECT_EQ(generic_5.GetLanguageFiles(), std::vector<std::string>({"en"}));
+}
+
+TEST(LanguageTerritory, English)
+{
+	const auto en_1  = LanguageTerritory("en");
+	const auto en_2  = LanguageTerritory("EN");
+	const auto en_3  = LanguageTerritory("eN");
+	const auto en_US = LanguageTerritory("en_US");
+	const auto en_GB = LanguageTerritory("en-GB");
+
+	EXPECT_FALSE(en_1.IsEmpty());
+	EXPECT_FALSE(en_2.IsEmpty());
+	EXPECT_FALSE(en_3.IsEmpty());
+	EXPECT_FALSE(en_US.IsEmpty());
+	EXPECT_FALSE(en_GB.IsEmpty());
+
+	EXPECT_FALSE(en_1.IsGeneric());
+	EXPECT_FALSE(en_2.IsGeneric());
+	EXPECT_FALSE(en_3.IsGeneric());
+	EXPECT_FALSE(en_US.IsGeneric());
+	EXPECT_FALSE(en_GB.IsGeneric());
+
+	EXPECT_TRUE(en_1.IsEnglish());
+	EXPECT_TRUE(en_2.IsEnglish());
+	EXPECT_TRUE(en_3.IsEnglish());
+	EXPECT_TRUE(en_US.IsEnglish());
+	EXPECT_TRUE(en_GB.IsEnglish());
+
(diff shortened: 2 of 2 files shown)


Torinde:
I can't find artifact for this yet (maybe due to [WIP]), so sharing my thoughts without testing:
- bg103 "Phonetic" should be the default layout for Bulgaria (preference for bg103 is clear, although there is sizeable number of people using the bg/bg442 "BDS", especially those with typewriter experience)
- regardless which of the bg layouts is selected - keyboard entry by default should remain in Latin/QWERTY state, so that user doesn't get Cyrillic letters at the DOS prompt (those can be manually invoked when user wants them).

FeralChild64:
@Torinde, OK, `bg103` is now the first one.

All the keyboard layouts start with the 1st plane when loaded - in the FreeDOS keyboard layouts we supply the 1st plane is always Latin (but not always QWERTY - can be QWERTZ or AZERTY in some cases).

johnnovak:
Just one note I said on Discord already:

- Auto-detecting the language and keyboard based on OS settings is all fine and well... on the surface, at least.
- The fact is that most DOS software are English and assume CP437. So special chars would be displayed incorrectly in such software out-of-the-box for people on non-English OSes (potentially).
- I think games reading the scancodes directly from port 60h is a much lesser problem becase all they care about is the scancodes.

Isn't potentially breaking the majority of English DOS text-mode software that assume CP437 a problem with this stuff? Text mode MOD players, ANSI editors, etc. — all use graphical chars and non-CP437 might break them.

People running _actual_ DOS software that assume their auto-detected codepage is extremely minor, niche of a niche of a niche... I think the semi-random non-CP437 code page people get on non-English OSes would break far more stuff.

So we're talking about trading potentially being wrong 99% of the time for being correct in 1% of the niche use cases... unless I'm missing something obvious 😆 

I think what you _really_ ned the non-English CP for is to display the localised setting help text, right? That's not a bad objective, indeed. Can we achieve that by a better mechanism while keeping the actual DOS codepage CP437? So changing the DOS codepage would always be an explicit user action for that 1% niche non-English software.

As I see it, we write the help text to the console using our C++ code, so not via emulated BIOS interrupts. So we could do whatever. We could always load two charsets: one for the actual codepage configured by the user (CP437 unless they explicitly change it), and another charset for the locale setting just for being able to display the help text correctly.

I know it's a complication, but maybe worth considering it 😄 I appreciate you like to tinker with the localisation stuff, just this is a shitload of code already, so it's a hard sell for having all this automation to actually _break_ lots of common DOS software 😆 So adding 20-30% more complexity on top of all this complex stuff isn't a big deal I think, and the behaviour would be _way_ better (no more weird questions about "why does my old text mode MOD player look broken on my French OS" with stock settings...).

rderooy:
Why not allow the integrated help text to be displayed in UTF-8, why yet another limited codepage, that may or may-not have the correct character?

Torinde:
Using different codepage (or just UTF-8) for localized messages is interesting, yes.

As for the auto-detection vs CP437 - my suggestion:
- .conf setting "auto" uses the auto-detection
- default is US/English/CP437/etc., so user has to put `auto` or his country in .conf when they want that, while the majority just gets CP437.

Additionally it'll help if there is:
- [ability to change the codepages on-the-fly](https://github.com/dosbox-staging/dosbox-staging/issues/3233#issuecomment-1867827184) (via key combo and later OSD), e.g. between 437 and country codepage (which I know isn't planed currently)
  - Illustrating why it's needed for the user to be able to switch to/from 437 - side-by-side - 437 vs 3021, including both Latin and Cyrillic text:
image

johnnovak:
Yeah, so high-level goal:

- Run as many DOS games and also apps correctly out-of-the-box

That's it. Whatever we need to do to make that happen. Localisation is secondary.

Automatically changing the codepage from the most common CP437 based on host-OS settings just so that we can display localised help text while breaking many CP437-assuming DOS programs is a bad idea.

Note I'm not against localisation, I'm against diverging from the most compatible setting out-of-the-box in a semi-random way (based on host-OS settings).

FeralChild64:
> People running actual DOS software that assume their auto-detected codepage is extremely minor, niche of a niche of a niche...

Norton Commander, Windows 3.11 setup... here is how the official Polish versions look on the code page 437:

Screenshot2

Screenshot3

> I think what you really ned the non-English CP for is to display the localised setting help text, right? That's not a bad objective, indeed. Can we achieve that by a better mechanism while keeping the actual DOS codepage CP437? So changing the DOS codepage would always be an explicit user action for that 1% niche non-English software.

Our localized help and error/status messages can be displayed using any code page (unless a Cyrillic, Greek, or other non-Latin writing system is used by the translation) - but DOS software/games are problematic. French version of Another World was already mentioned on the Discord, here is another example - a README file supplied with Earth 2140 (RTS originally developed by a Polish studio):

Screenshot1

If you are a non-English user of DOSBox (any fork) who wants to do more than running pre-configured games, there is a high chance you'll need to learn about switching the code pages.

----

Besides, this PR is not much about code pages, more about keyboard layout autodetection. Switch the keyboard layout to French (`keyb fr`) and try to work in the shell for a while, display some file using the `type` command, navigate the directories, etc. - this is exactly how French or German users feel like when they are getting the US keyboard layout. And not all the FreeDOS keyboard layouts can be used with code page 437 (I have a plan to change this - but this is quite some work).

As for the fast switching of code pages and keyboard layouts - even if I implement it, it won't be obvious for the user how to change it. We might implement this once we have the GUI.

FeralChild64:
macOS build job seems to be broken:

>  Error: The template is not valid. /Users/runner/work/dosbox-staging/dosbox-staging/./.github/actions/setup-vcpkg/action.yml (Line: 45, Col: 12): hashFiles('vcpkg.json, vcpkg-configuration.json, vcpkg-ports/**') failed. Fail to hash files under directory '/Users/runner/work/dosbox-staging/dosbox-staging'

Torinde:
> not all the FreeDOS keyboard layouts can be used with code page 437 (I have a plan to change this - but this is quite some work).

If that will be useful - I can make tables of correspondence key-character for some of them - after you decide what format those should be in. Also, I assume you can ask each translator to prepare such for his languages.

FeralChild64:
@Torinde I am not going to ask the translators to prepare keyboard layouts, it's not their role. For now I don't know much about the keyboard layout file formats, this is something I'll have to investigate this to fix several bugs. But this won't happen for 0.83.0.

Adapting the keyboard layouts to different code pages should be possible to be done automatically.

johnnovak:
> If you are a non-English user of DOSBox (any fork) who wants to do more than running pre-configured games, there is a high chance you'll need to learn about switching the code pages.

Yeah, but my whole point is they're more likely run English software. There's way more English software, non-English stuff is 5% of all software at best.

Also, people who used non-English software must have learned about codepages and all that stuff.

In retrospect, you could have done a simplified master setting that controls all localisation stuff, then no need for all the complicated auto-detection stuff. It's tons of code, the user could just do `language = french` once in the primary config and done.

FeralChild64:
But almost all of the English stuff (especially games) works fine with code pages like 850.

> In retrospect, you could have done a simplified master setting that controls all localisation stuff, then no need for all the complicated auto-detection stuff. It's tons of code, the user could just do `language = french` once in the primary config and done.

Not really. Which French keyboard layout should be loaded in such case? The regular one? International? Swiss? Belgian? Canadian? They differ.

If you ask me, the keyboard layout autodetection shouldn't have been implemented in the first place - we can keep improving it till some point, but the implementation will never guess 100% correctly. But the functionality is already there and removing it now could upset many folks, especially from the "QWERTZ" or "AZERTY" areas.

Language (translation), country code, time&date format, numeric format - these can be all autodetected reliably, it's only the keyboard layout + code page which is problematic.

johnnovak:
> Not really. Which French keyboard layout should be loaded in such case? The regular one? International? Swiss? Belgian? Canadian? They differ.

It's always the French. Troublemakers 😂 

> If you ask me, the keyboard layout autodetection shouldn't have been implemented in the first place - we can keep improving it till some point, but the implementation will never guess 100% correctly. But the functionality is already there and removing it now could upset many folks, especially from the "QWERTZ" or "AZERTY" areas.

It's never too late to delete code 😏 Like I did with complicated VFR/CFR stuff and all the uber-complex (and flaky) autodetection nonsense.

Deleting one's own code is also a good practice for non-attachement 😆 (if you're interested in Buddhist philosophy 😆 )

My learning from this is that auto-detecting "environmental" stuff is often fraught with danger. In my upcoming PR that will add wide color gamut support I'm not even attempting it -- reasonable defaults, then people can manually set it up for their use case. Too much variations and weird setups, millions of ways to get it wrong, and platform-specific APIs. Just nope.

But ok, if you wanna keep it it's fine. But please try to hit the 90% sweet spot with it, and call it a day. "Best effort", then let people use manual override in the 10% scenario.

> Language (translation), country code, time&date format, numeric format - these can be all autodetected reliably, it's only the keyboard layout + code page which is problematic.

Yeah, that's reasonable 👍🏻 I hope you have learned from it, autodetection can be very hard and costly to support.

Tbh, I don't see what's wrong with a fully manual setup like in every other emulator, but whatever. It's your time 😆

FeralChild64:
> But please try to hit the 90% sweet spot with it, and call it a day.

That's mostly what I want to achieve. I hope we'll receive some more feedback after RC1 is released.

> I hope you have learned from it, autodetection can be very hard and costly to support.

I knew it from the beginning - it wasn't me who initially implemented the keyboard layout autodetection. It just happened that I am the one who is maintaining it now.

src/misc/iso_locale_codes.cpp
```diff
@@ -0,0 +1,697 @@
...
+const std::string Kosraean         = "kos";
+const std::string Niuean           = "niu";
+const std::string Palauan          = "pau";
+const std::string Pohnpeian        = "pon";
+const std::string Sonsoralese      = "sov";
+const std::string Tobian           = "tox";
+const std::string Tokelauan        = "tkl";
+const std::string Tuvaluan         = "tvl";
+const std::string Ulithian         = "uli";
+const std::string Yapese           = "yap";
```
johnnovak, line 694: Man, are we serious about adding these languages nobody ever heard of? 😆 Oceania languages, Africa languages, Russian Federation minority languages... really? 🤣

I guess you did it for completeness, but quite pointless to add support for all the minority languages DOS wasn't even released for...

DOS supported about 21 langues according to this article:

https://learn.microsoft.com/en-us/previous-versions/tn-archive/cc722862(v=technet.10)?redirectedfrom=MSDN#XSLTsection136121120120

src/misc/private/iso_locale_codes.h
```diff
@@ -0,0 +1,721 @@
...
+
+// Three-letter codes - Oceania languages
+extern const std::string Chuukese;   // a.k.a Chuuk
+extern const std::string Gilbertese; // a.k.a. Kiribati
+extern const std::string Hawaiian;
+extern const std::string Kosraean;
+extern const std::string Niuean;
+extern const std::string Palauan;   // also covers Angaur
+extern const std::string Pohnpeian; // a.k.a. Ponapean
+extern const std::string Sonsoralese;
```
johnnovak, line 712: Seriously, let's say user has OS set to Sonsoralese. What will even happen then? Does DOS even have a codepage for it? What's the point?

johnnovak (approved):
Code looks good, so approving.

Personally, I think you're really making life difficult for yourself with the keyboard auto-detection and adding all those languages with less than a few thousand total speakers 🤣 

Probably a better approach would be to only support the major European languages and the English variants, just like MS-DOS did, so about 21 languages plus regional variations. Anything else => unsupported, default to English. But as long as you're maintaining this, I'm fine with it 😄  

And again:

> I knew it from the beginning - it wasn't me who initially implemented the keyboard layout autodetection. It just happened that I am the one who is maintaining it now.

If you wanna kill it, I won't hold you back 😏

src/misc/iso_locale_codes.cpp
FeralChild64: But we are using FreeDOS CPI files and keyboard layouts, which support a LOT more languages - see for example here, https://gitlab.com/FreeDOS/base/keyb_lay/-/blob/master/DOC/KEYB/LAYOUTS/UK-INTL.TXT

I have just mapped the ISO 639 codes to what the FreeDOS documentation states. And since the last real (non-trivial) modification of the FreeDOS CPI/SYS locale files happened like 15 years ago, I don't think this will need much maintenance effort.

src/misc/private/iso_locale_codes.h
FeralChild64: image

But it will only possibly affect the keyboard layout autodetection.

FeralChild64:
> If you wanna kill it, I won't hold you back 😏

Everything depends on the users feedback.

If, after all my effort, there is still going to be a lot of complaints about keyboard layout auto-detection wrongly guessing the layout, and there will be no way to easily fix them, I'll probably remove the auto-detection code.

src/misc/iso_locale_codes.cpp
johnnovak: Ok, forgot about FreeDOS. That kinda stretches our goal of running DOS software from the pre-2000 DOzs era, but okay 😄 I guess the keyword is it's not hard to add complete support.

johnnovak:
> If, after all my effort, there is still going to be a lot of complaints about keyboard layout auto-detection wrongly guessing the layout, and there will be no way to easily fix them, I'll probably remove the auto-detection code.

Sounds like a good plan and a wise decision.

Torinde:
> @Torinde, OK, `bg103` is now the first one.

@FeralChild64, with current main I get bg/bg442 (national BDS) and not bg103 (phonetic) - with empty .conf file:
image

Am I doing something wrong?

Then I can switch to bg103 manually, of course:
image

FeralChild64:
@Torinde Your Windows has a `Bulgarian (Typewriter)` keyboard layout set, right? The mapping of the host Windows layout to guest looks like below:

```c++
	// Bulgarian (QWERTY/national)
	{ "00030402", { "bg" }         }, // Bulgarian
	{ "00010402", { "bg" }         }, // Bulgarian (Latin)
	{ "00000402", { "bg" }         }, // Bulgarian (Typewriter)
	// Bulgarian (QWERTY/phonetic)
	{ "00040402", { "bg103" }      }, // Bulgarian (Phonetic Traditional)
	{ "00020402", { "bg103" }      }, // Bulgarian (Phonetic)
```

That's why `bg` was selected, not the phonetic version.

Torinde:
No, I don't have Typewriter - I have Phonetic Traditional.
But in addition to stock also I have some custom made layouts including this:
> LOCALENAME	"bg-BG"
> LOCALEID	"00000402"
> ...
> DESCRIPTIONS
> 0409	Bulgarian (Phonetic Customized)
> LANGUAGENAMES
> 0409	Bulgarian (Bulgaria)

It seems the LOCALEID is wrong - I should've written 00040402?

I should test on a computer only with stock Phonetic or with the custom layout fixed with 00040402...

Still, since I have _also_ Phonetic Traditional - shouldn't that take priority for Staging? E.g. if bg103 and bg are both present, use bg103?

In the log I see only 402 and 409, no 40402.... maybe again some problem due to the custom layouts?

FeralChild64:
I have taken the value `00000402` from https://kbdlayout.info/kbdbu. `00000409` is English layout, see here https://kbdlayout.info/kbdus.

If the IDs are not valid anymore due to custom keyboard layouts in your Windows, then there is nothing I can do about it.

Torinde:
I removed the custom keyboard layout and Staging properly went to bg103 by itself. Thank you for your efforts @FeralChild64!

As for the layout numbers - yours are correct. My custom layout has to be fixed to 40402, that has nothing to do with Staging, true.

